Issue #327, opened 2026-08-03
Layer streaming: the VRAM pre-flight over-predicts (LOGITS_BYTES_PER_ELEMENT is stack-dependent, not preference-specific)


MakazhanAlpamys:
v0.72.4 budgets DPO / ORPO / SimPO at `2 x batch_size` rows, because their forward concatenates chosen and rejected into one tensor. That is correct and it never under-predicts — but it charges those rows at `ForCausalLMLoss`'s **measured 14 bytes/element**, and TRL's preference losses do not pay that.

## The gap, measured

TRL reduces logits to per-token log-probs (`selective_log_softmax`) instead of holding a full-vocabulary fp32 upcast. Measured on the dev box: DPO's entire above-resident cost was **51.76 MB**, where the 14 B/elt charge for the same shape is **~458 MB**.

Consequence on a 4 GB card (Llama-3.2-1B, vocab 128 256, NF4, 3.32 GB free):

```
seq  512  SFT 1.65 GB allowed  |  DPO 2.63 GB allowed
seq  768  SFT 2.14 GB allowed  |  DPO 3.60 GB REFUSED
seq 1024  SFT 2.63 GB allowed  |  DPO 4.58 GB REFUSED
```

So DPO is refused from `max_length: 768` upward on configurations that would probably run.

## Why no constant was published in v0.72.4

Two measurement grids were attempted and **both were invalid**, so nothing was fitted:

1. With short rows, `max_length` never bound the effective sequence — DPO measured an identical 63.43 MB at seq 128, 256 **and** 512. An independent variable that does not move measures nothing.
2. With long rows, the batch shapes showed TRL was not truncating to `max_length` either (`(1, 2401)` at every setting).

Shipping the conservative bound was the right call because under-predicting is the strictly worse failure: on Windows it is not an exception but a silent WDDM spill.

## Fix path

Get real control over the effective sequence length first (that is the actual blocker), then repeat v0.72.3's methodology: vary seq and rows at a realistic vocab, subtract the measured resident baseline, solve for the per-element slope, and use a per-loss constant.

## Acceptance criteria

- [ ] a grid where the effective sequence length demonstrably varies with the setting
- [ ] a fitted per-element cost for the preference path, with worst-case error stated
- [ ] the estimator still **never** under-predicts across the grid
- [ ] the seq-768 DPO configuration above is no longer refused

MakazhanAlpamys:
## Measured on real models — and the cause is not the preference path

[...]

PR #386, opened 2026-08-13
feat(stream): let training.stream_vram_override replace the measured free VRAM

AmirF194:
## Root cause

`decide_stream_fit` refuses any layer-streaming step it predicts will not fit, with no way through except lowering `training.batch_size` or `data.max_length` (`trainer/stream_setup.py:_stream_budget_lines`). That refusal is deliberate for the common case: on Windows an under-prediction does not raise, it silently spills to host memory. But the free-VRAM figure it checks against comes from `torch.cuda.mem_get_info()`, a device-level driver query, and that query cannot see a per-process cap: `torch.cuda.set_per_process_memory_fraction`, a MIG slice, or a card another process is also using. On that hardware the pre-flight sees the whole device's free VRAM regardless of what the process can actually use, so there was no way to make the pre-flight simulate the real constraint either.

## Fix

`training.stream_vram_override` (bytes) replaces the measured free-VRAM figure the pre-flight checks against, in either direction:

- raised past a documented over-prediction, to let a known-safe config through;
- lowered below the real free VRAM, to enforce a cap the driver itself cannot report (a capped Colab/Kaggle T4, a MIG slice).

Rejected at config load when set while `stream_layers` is false, mirroring the existing `stream_source`/`stream_buffers` footgun gate.

**One note on shape, since the issue changed under me while it was open.** The original acceptance criteria describe a `bool`/`force` flag that lets an over-budget run proceed. The follow-up comment on this issue narrows and extends that: a boolean cannot express "assume only 4 GB is free" for the capped-notebook case, so it asks for a numeric value that *replaces* the measured figure instead. I built the follow-up comment's shape, since it is the one the notebook use case actually needs and it still satisfies the original ask (an over-prediction can be raised past). One consequence: the override does not unconditionally let a run "proceed" the way the original bool would have. It changes the yardstick `decide_stream_fit` measures against, and that comparison still refuses if the predicted peak exceeds it, which is what makes "override set below real free VRAM refuses a fitting config" a real test rather than a tautology.

## Verification

- New tests fail on `main` (`ImportError`, `resolve_available_vram_bytes` does not exist there) and pass on this branch: `resolve_available_vram_bytes` unit tests (no override uses the measured figure; override replaces a larger or smaller measured figure either way), the acceptance test the follow-up comment describes verbatim (override set below real free VRAM turns an otherwise-fitting config into a refusal), and schema tests (default `None`, accepts an explicit byte count, rejects `bool`/negative, rejects being set while `stream_layers` is false).
[...]

PR #384, opened 2026-08-11
fix(stream): canonicalise named_parameters() for streamed models

AmirF194:
## Root cause

v0.72.1 made `StreamedDecoderLayer.state_dict()` canonical for serialisation, delegating at the wrapper's own prefix so a saved adapter loads into any normal model. `named_parameters()`/`named_modules()` were left untouched: every wrapped decoder layer's parameters still carry an extra `.inner.` segment (`...layers.0.inner.self_attn.q_proj...`) that a resident (non-streamed) model of the same checkpoint does not have.

That gap first looked harmless because nothing shipped compared the two by name, until the #331 repair gate at 32B did exactly that for gradients and read `grads exact 0/0` as success (`benchmarks/gate-h100-validation.md` L2627-2643): the streamed and resident key sets did not intersect, so `n_exact == n_compared` was vacuously true on nothing compared.

## The invariant

A name-keyed comparison between a streamed and a non-streamed model of the same checkpoint must be over parameter names that actually mean the same weight, and a comparison helper must fail loudly on an empty intersection rather than report it as a pass.

## Fix

Ships fix shape (B) from the issue (the cheaper one, preserving the "gates stay valid" property without re-running the v0.72.0 bit-exactness suites):

- `canonical_named_parameters(model)`: strips `.inner.` the same way `state_dict()` already does.
- `assert_canonical_parameters_intersect(model_a, model_b)`: computes both canonical name sets and raises `ValueError` if the intersection is empty, returning the shared set on success.

Fix shape (A), making `named_parameters()` itself canonical, would also require re-running the v0.72.0/.2/.3 bit-exactness suites on real hardware. This sandbox has no GPU, so I could not do that safely; (B) touches neither the forward path nor `state_dict()`, so those gates stay valid unexercised, the same way they did for the v0.72.1 fix itself.

## Verification

- New regression tests (`tests/test_issue369_canonical_named_parameters.py`) build a real streamed DPO wrapper and a resident model from the same tiny checkpoint (CPU, `python:3.12-slim`, `pip install -e ".[dev]"`). A control test proves the bug: the two models' raw LoRA parameter names share nothing (reproducing the gate's `0/0`). The other tests fail on `main` (`ImportError`, the functions do not exist yet) and pass on this branch: canonical names strip `.inner.`, the streamed and resident canonical sets are equal, and `assert_canonical_parameters_intersect` both raises on a disjoint pair and returns the shared names on an intersecting one.
[...]

PR #390, opened 2026-08-14
fix(stream): wire calibrated_logits_bytes_per_element into the VRAM pre-flight

### LINKED ISSUE #348: Layer streaming: calibrated_logits_bytes_per_element is dead code — wire it into estimate_stream_peak_vram
`src/soup_cli/utils/layer_stream.py` ships `calibrated_logits_bytes_per_element()`
(and the `measure_logits_loss_bytes_per_element()` probe underneath it), documented
as the guard against a future stack growing a fourth fp32 buffer in the loss path —
which would be under-budgeted by 12.5% today with nothing to catch it.

**It has no caller.** `grep -rn "calibrated_logits_bytes_per_element" src/` matches
only its own definition and the docstrings that reference it.
`trainer/stream_setup.py:455` calls `estimate_stream_peak_vram(...)` without
`logits_bytes_per_element=`, so the parameter is always `None` and the shipped
constant is always used. There is no CLI flag or env var that reaches it either.
The function is exercised only by `tests/test_issue327_logits_estimate.py`.

So the calibration is **an API, not a live path**. The hole it was written to close
is still open.

## Proposed fix

Thread it through: `_stream_budget_lines` calls `calibrated_logits_bytes_per_element()`
and forwards the result as `logits_bytes_per_element=` to `estimate_stream_peak_vram`.

The existing semantics already make this safe — the value is floored at the shipped
constant inside `estimate_logits_bytes`, so it can only *raise* the budget, never
lower it. A stack that measures below 12 is a probe that could not see the retention,
not a stack that does not pay it.

Two things to decide when implementing:

- **Cost.** The probe allocates `14 * vocab * max(tokens)` transiently — 96 MiB at
  the defaults — and needs CUDA. It returns `None` (falling back to the constant)
  rather than raising when it cannot run, which is the right shape, but it should
  probably be skipped when the fit decision cannot depend on it, the way
  `choose_tier` takes a callable so the ~9 s disk probe is only paid when it matters.
- **Reporting.** If the calibration raises the budget above the constant, the
  pre-flight panel should say so, with both numbers. A budget that silently differs
  from the documented constant is worse than no calibration.

## Acceptance criteria

- [ ] `_stream_budget_lines` forwards a calibrated value.
- [ ] Test asserting the pre-flight *uses* it: a stub probe returning a high value
      raises the prediction, and one returning a low value does not lower it.
- [ ] Test that a `None` probe (no CUDA) leaves the prediction byte-identical to
      today's.
- [ ] Panel line shows the calibrated value when it differs from the constant.

AmirF194:
## Root cause

`calibrated_logits_bytes_per_element()` exists to raise the VRAM pre-flight's budget when this stack's loss path measures a heavier retention than the shipped constant assumes: the guard against a future stack silently under-budgeting by 12.5% with nothing to catch it. It had no caller. `_stream_budget_lines` calls `estimate_stream_peak_vram()` without `logits_bytes_per_element=`, so the parameter was always `None` and the calibration ran only inside `tests/test_issue327_logits_estimate.py`.

## Fix

`_stream_budget_lines` now calls `calibrated_logits_bytes_per_element()` once and forwards the result to both `estimate_stream_peak_vram()` (the actual budget) and `estimate_logits_bytes()` (the panel's `logits` figure), so the two stay consistent. The value is floored at the shipped constant inside `estimate_logits_bytes`, so this can only raise the prediction, never lower it. When the calibration measures above the constant, the panel prints an extra line naming both numbers.

## Verification

- `tests/test_issue348_calibrated_logits_wiring.py` (new): a stubbed high calibration raises the predicted peak, a stubbed calibration below the constant does not lower it, a `None` probe (no CUDA) leaves the prediction byte-identical to today, and the panel adds a line only when the calibration diverges from the constant. Confirmed each assertion fails on `main` and passes on this branch.
- `tests/test_issue327_logits_estimate.py` (the existing calibration-arithmetic suite) passes unchanged.
- Full suite (`pytest tests/`, Docker, `pip install -e .[dev]`) on Python 3.10 and 3.12: 186 passed, 18 skipped (CUDA-only rows this sandbox has no GPU for). `ruff check` and `mypy` clean on the changed file.
- Not verified: the actual GPU probe path (`measure_logits_loss_bytes_per_element` with a real CUDA device) and the Windows/macOS legs of the CI matrix, since this sandbox is Linux CPU-only. The wiring is covered by stubbing the calibration function; the probe's own arithmetic was already pinned by #327's tests.

Fixes #348

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **Layer streaming's VRAM pre-flight now actually calls its own calibration hook (#348).**
+  `calibrated_logits_bytes_per_element()` exists to raise the budget when a stack's loss
+  path measures a heavier retention than the shipped constant assumes, guarding against a
+  future stack silently under-budgeting by 12.5% with nothing to catch it. `_stream_budget_lines`
+  called `estimate_stream_peak_vram()` without `logits_bytes_per_element=`, so the parameter
+  was always `None` and the calibration never ran outside its own test. It is now forwarded to
+  both the budget and the panel's `logits` figure; the value can only raise the prediction
+  (floored at the shipped constant), and the panel prints an extra line naming both numbers
+  when the calibration measures above it.
 - **MLX backend now actually dispatches to the MLX trainer for `task: sft`.** Previously `backend: mlx` silently fell through to the transformers `SFTTrainerWrapper`, training on MPS/CUDA instead of MLX. The trainer was also rewritten for mlx-lm >= 0.31 (`create_dataset` + `CacheDataset`, `TrainingCallback`), with `model.freeze()` before LoRA — without it the saved "adapter" was a full fine-tune (172 tensors vs 24 LoRA tensors on a 1.2B model) — and an `adapter_config.json` is written so the output dir loads directly with `mlx_lm.load(..., adapter_path=dir)`. (#362)
 - **`mlx-lm` floor raised to >= 0.31.3** (the version the MLX SFT path is built against).
 - **`training.seed` reached the SFT wrapper and nothing else (#353).** #341 added the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,13 @@ reproducing 70+ versions of notes.
   was always `None` and the calibration never ran outside its own test. It is now forwarded to
   both the budget and the panel's `logits` figure; the value can only raise the prediction
   (floored at the shipped constant), and the panel prints an extra line naming both numbers
-  when the calibration measures above it.
+  when the calibration measures above it. This makes the probe unconditional rather than
+  opt-in (see #327 below, whose wording is updated to match): every streamed run now pays
+  one transient `14 * vocab_size * max(tokens)` allocation (96 MiB at the defaults) plus two
+  `torch.cuda.synchronize()` calls before the fit decision is taken. On today's measured
+  stacks this is a no-op in effect (`measured` is 12.0 with zero spread, `max(14, 14)` is
+  14, no extra line prints), so the cost buys nothing yet, which is the point of a guard
+  against a stack that hasn't shipped.
 - **MLX backend now actually dispatches to the MLX trainer for `task: sft`.** Previously `backend: mlx` silently fell through to the transformers `SFTTrainerWrapper`, training on MPS/CUDA instead of MLX. The trainer was also rewritten for mlx-lm >= 0.31 (`create_dataset` + `CacheDataset`, `TrainingCallback`), with `model.freeze()` before LoRA — without it the saved "adapter" was a full fine-tune (172 tensors vs 24 LoRA tensors on a 1.2B model) — and an `adapter_config.json` is written so the output dir loads directly with `mlx_lm.load(..., adapter_path=dir)`. (#362)
 - **`mlx-lm` floor raised to >= 0.31.3** (the version the MLX SFT path is built against).
 - **`training.seed` reached the SFT wrapper and nothing else (#353).** #341 added the
@@ -335,10 +341,12 @@ exist, and repairs four backends.
 - **`LOGITS_BYTES_PER_ELEMENT` is split into two independently measured terms** (#327).
   The 14 is 12 + 2, measured stage by stage on an H100 with zero spread across three
   repeats, and only the 2 differs between stacks. It is deliberately **not lowered** —
-  see Known Limitations. What is new is an opt-in, upward-only calibration
+  see Known Limitations. What is new is an upward-only calibration
   (`max(14, measured + 2)`), which closes a real unguarded hole: a future stack that
   grew a fourth fp32 buffer would be under-budgeted by 12.5% today with nothing to catch
   it. Default behaviour is byte-identical and the pre-flight path takes no new CUDA.
+  (Originally landed as an opt-in probe with no caller; #348 above wires it into the
+  pre-flight itself, so it now runs on every streamed run rather than sitting inert.)
 
 ### Fixed — eval, ship and export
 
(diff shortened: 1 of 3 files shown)


MakazhanAlpamys (changes requested):
Wiring is right, the floor is in the correct place, and forwarding one calibrated value to both the budget and the panel so they cannot disagree is the detail I would have asked for. CI is 11/11. Two things before merge, and the first is the same shape I have flagged twice now — which is worth saying plainly, because it is not carelessness, it is a specific blind spot worth naming once so it stops recurring.

## Blocking: the suite passes with the calibration disconnected from the budget

I removed the forwarding to `estimate_stream_peak_vram` and left it going to `estimate_logits_bytes` — i.e. the panel shows a calibrated logits figure while the budget that actually refuses a run keeps the shipped constant. That is precisely the defect #348 exists to close:

```
$ # estimate_stream_peak_vram(...)  <- logits_bytes_per_element removed
5 passed in 0.33s
```

Nothing failed. The reason is in the assertion:

```python
baseline = _budget_lines()[0]
...
raised = _budget_lines()[0]
assert baseline != raised
```

`_budget_lines()[0]` is one string carrying **both** numbers:

```
  peak VRAM    ~X.XX GB at batch 1 x seq 4 (logits Y.YY GB)
```

so a change in the `logits` half alone satisfies `baseline != raised`. The test pins *"the line changed"*, and the PR description claims *"a stubbed high calibration raises the predicted peak"* — those are different statements, and only the second is #348. The display figure is cosmetic; the budget is the thing that refuses a run.

The other two cannot cover it either: `test_a_below_constant_calibration_does_not_lower_the_peak` compares two equal outcomes, and `test_a_none_probe_leaves_the_peak_byte_identical_to_today` runs at `calibrated == 14 == LOGITS_BYTES_PER_ELEMENT`, where forwarding and not forwarding produce identical bytes by construction.

Assert on the peak alone, so the two halves are separable:

```python
def _peak_gb(line: str) -> str:
    return line.split("~", 1)[1].split(" GB", 1)[0]

assert _peak_gb(baseline) != _peak_gb(raised)
```

My mutation then fails, and — worth adding as a second assertion — the panel's `logits` figure moving *with* it becomes a separate claim rather than the one carrying the test.

**The pattern, said once.** #384: a positive test satisfied by `embed_tokens`/`norm`/`lm_head`, which never carried `.inner.`. #386: `override=0` pinned at config load but not at the resolver. Here: a peak assertion satisfied by the logits display. Each time the code is correct and the assertion is answered by something *adjacent* to the claim. The mutation test is what separates them — you ran it yourself on #386 and found the real value coming back, which is exactly the habit. Worth applying to the assertion as well as to the code: after writing it, ask what else could make it green.

## Second, and it needs your opinion rather than a fix: this makes an opt-in thing unconditional

`calibrated_logits_bytes_per_element()` now runs on **every** streamed run. #327's recorded decision was that the constant stays at 14 and only an **opt-in**, upward-only calibration was added. #348 asks for the wiring, so this is arguably the completion of that work rather than a reversal — but it is a change to a decision that was made deliberately, and it should be made out loud rather than as a side effect.

The concrete cost, from the probe's own docstring: one transient allocation of `14 * vocab_size * max(tokens)` — **96 MiB at the defaults** — plus two `torch.cuda.synchronize()` calls, on the card that is short of VRAM, before the fit decision has been taken. It also calls `torch.cuda.reset_peak_memory_stats()`, so anything downstream reading `max_memory_allocated` now measures from the probe rather than from process start.

Today it is a no-op in effect: `measured` is 12.0 with zero spread on this stack, `+2` retention gives exactly 14, `max(14, 14)` is 14, and the extra panel line correctly does not print. So the cost buys nothing until the stack changes — which is the entire point of a guard, and I am not arguing against it.

I would land it as written and take the cost. But say which you intend, so the next reader of #327's decision finds the reason: **(a)** always-on, cost accepted, `#327`'s "opt-in" wording updated in the same PR; or **(b)** gated behind a config field, default off, with the guard reachable for anyone who wants it. If (a), one line in the CHANGELOG entry naming the 96 MiB is enough.

---

The test change is the only thing I need. Answer (a)/(b) in a sentence and I will merge on your next push.

AmirF194:
Fixed. Split the peak GB out of the line and asserted on it alone (`_peak_gb`), then added a second test asserting the logits figure separately so each half is its own claim. Ran the mutation you described (drop `logits_bytes_per_element=` from the `estimate_stream_peak_vram` call only) in a clean container: the new peak assertion fails on it, the old one did not.

Answer to the second point: (a), land as written, cost accepted. Updated #327's entry to drop the "opt-in" wording and named the 96 MiB / two-sync cost in #348's entry.

Pushed to the same branch.

MakazhanAlpamys:
Both closed, and the second one better than I asked.

**The peak assertion is now real.** Ran the mutation again — calibration reaching the panel but not the budget:

```
FAILED ...::TestCalibrationIsForwardedToTheBudget::test_a_high_calibration_raises_the_predicted_peak
1 failed, 5 passed
```

Exactly the right test, and the five others unaffected. Splitting `_peak_gb` out and giving the logits figure its own test is the shape that stops the two halves covering for each other, and you checked the old assertion did *not* fail on the mutation — which is the half people skip.

**You answered (a) and then did the part that makes it a decision rather than a side effect:** dropping "opt-in" from #327's entry and naming the 96 MiB + two-sync cost in #348's, with the honest note that it is a no-op in effect today. Someone reading #327 in six months now finds a reason instead of a contradiction. I asked for one sentence; the CHANGELOG edit is worth more than the sentence.

CI 11/11. Merging.

---

Three PRs in four days, and the through-line is worth naming since it is the reason these went fast: each time the code was right first try, and what needed another round was an *assertion* that something adjacent could satisfy. You closed that gap yourself this time, before I could — you ran the mutation, checked the old test did not catch it, and only then wrote the new one. That is the whole technique.

#348 closes with this. If you want more, `good first issue` and `help wanted` are [labelled in the tracker](https://github.com/MakazhanAlpamys/Soup/labels/help%20wanted); nothing is reserved and there is no expectation either way.

AmirF194:
Thanks. Running the mutation before writing the new assertion is the part I want to keep doing on every test I write from now on, not just when a review points it out.

Issue #297, opened 2026-07-04
soup mcp serve: live train/export execution (v1 mutating tools are plan-only)


MakazhanAlpamys:
## Context

`soup mcp serve` (v0.71.28) exposes two mutating tools — `train_start` and `export` — gated behind `--allow-mutating`. In v1 they are **plan-only**: even with the flag, they validate the config/format and render the exact `soup train` / `soup export` command that *would* run, but never execute it. This was a deliberate safety choice (an agent-initiated GPU job needs a human-in-the-loop design), not an oversight.

## Ask

Allow `train_start` / `export` to actually execute when explicitly opted in, with a human-in-the-loop safeguard.

## Design considerations (needs a short design before code)

- **Opt-in surface:** a stronger flag than `--allow-mutating` (e.g. `--allow-execute`) so plan-only stays the default even for `--allow-mutating`.
- **Human-in-the-loop:** an agent should not silently kick off a multi-hour GPU job. Options: MCP elicitation/confirmation, a dry-run-then-confirm handshake, or a hard per-session cap.
- **Streaming:** training progress would need to stream back over MCP (ties into #296's HTTP transport for anything beyond fire-and-forget).
- **Isolation:** run as a subprocess (never in the server process); reuse the existing `soup train` / `soup export` entry points and their guards.

## Acceptance criteria

- [ ] A design note in the issue agreed before implementation.
- [ ] Live execution is strictly opt-in and off by default even with `--allow-mutating`.
- [ ] Runs as an isolated subprocess; progress is observable.
- [ ] Documented security posture + known risks.

CODING-DARSH:
One design question regarding the human-in-the-loop requirement.
The issue mentions several possible safeguards (MCP elicitation/confirmation, a dry-run → confirm handshake, or a per-session cap), but doesn't specify which direction is preferred.
From a server design perspective, should the confirmation be:
1. **Client-enforced**, where `--allow-execute` enables execution and the MCP client (Claude Code, Cursor, etc.) is responsible for obtaining user confirmation before calling the execution path, or
2. **Server-enforced**, where the MCP server itself requires an additional confirmation step (e.g. a plan → confirm handshake) before spawning the subprocess?
Both seem compatible with the rest of the design, but they lead to different implementations, so I wanted to clarify the intended direction.

MakazhanAlpamys:
[...]

PR #391, opened 2026-08-14
feat(mcp): add allow-execute gate

CODING-DARSH:
## What does this PR do?

Implements the first slice of the agreed MCP execution design by adding the --allow-execute gate to soup mcp serve.

Adds --allow-execute as a stronger opt-in than --allow-mutating.
--allow-execute implies --allow-mutating.
Threads allow_execute through run_stdio_server and build_registry.
Adds the _refuse_execute gate for the future execution path.
Preserves train_start and export as plan-only tools; no actual execution is introduced.
Adds tests for flag behavior, plumbing, and execution refusal.
Documents the new flag and its current plan-only behavior.

The confirmation-token handshake, subprocess execution, elicitation, streaming, and other execution behavior remain intentionally deferred to the subsequent implementation slice.

## Type of change

- [ ] Bug fix
- [✓ ] New feature
- [ ] Refactor / cleanup
- [✓ ] Documentation
- [ ✓] Tests

## Checklist

- [✓ ] `ruff check src/soup_cli/ tests/` passes
- [✓ ] `pytest tests/ -v` passes
Partially addresses #297


MakazhanAlpamys (changes requested):
Splitting the gate from the execution path was the right call, and the two-slice shape is what made a same-day review of the second one possible at all. 11/11 green, `tests/test_v07128.py` 91 passed / 1 skipped locally.

One thing to change, and one to note.

## Change before merge: `allow_execute` should default to `False`

`build_registry(*, allow_mutating: bool, allow_execute: bool)` makes the new flag a **required** keyword on a module `.claude/CLAUDE.md` documents as a public surface. The tell is in your own diff: ten call sites in `tests/test_v07128.py` changed for no reason other than to satisfy an argument they do not care about, and any out-of-tree caller breaks the same way with no deprecation.

`allow_execute: bool = False` costs nothing here — the flag is strictly additive, and its default is exactly the old behaviour. Same for `_mutating_specs`.

Please apply it on **#393's branch too**, since that one is stacked on this and will otherwise reintroduce it.

## Noted, not blocking: `_refuse_execute` has no caller in `src/`

Only `tests/test_v07128.py:529` reaches it, by calling the factory directly. So as this PR stands it ships a handler no client can trigger and a user-facing string — *"Execution tools are not implemented in this version"* — that nobody can ever see.

Normally I would push back harder, because this project spent #390 fixing exactly that shape: `calibrated_logits_bytes_per_element()` had no caller, was exercised only by its own test, and the guard it existed to provide therefore did not exist. The difference here is that #393 is already open and wires it immediately, so it is dead for a day rather than for releases. That is fine — but worth saying out loud, because "reserved for the next slice" and "dead code" look identical in a diff, and only the existence of #393 distinguishes them.

Same for `allow_execute` reaching `_mutating_specs` without being read in its body: harmless as a signature that the next commit fills in, invisible if #393 stalls.

## Correct, and worth naming

- **`allow_mutating = allow_mutating or allow_execute` is applied in both `commands/mcp.py` and `build_registry`.** Belt and braces on a security gate is the right instinct; a caller that reaches `build_registry` directly and skips the CLI still gets the implication.
- **The mode string distinguishes all three states** (`read-only` / `plan-only` / `plan-only; execution disabled`). The one line an operator reads at startup now says which of the two opt-ins is on, which matters precisely because one of them will eventually mean something much stronger.
- **The help text says execution is not implemented in this version** rather than describing the eventual behaviour. That is the honest tense and it is easy to get wrong.

One line, then I merge — and #393's review is written and follows right after.

CODING-DARSH:
Applied the requested backwards-compatibility fix. allow_execute now defaults to False in both build_registry() and _mutating_specs(), preserving the existing behavior for callers that don't provide the new flag.

Verified with tests/test_v07128.py — 82 passed, 10 skipped; Ruff passes and git diff --check is clean.

I’ll apply the same default on the stacked execution branch as requested. Thanks for catching this.

MakazhanAlpamys (approved):
Default applied in both `build_registry` and `_mutating_specs`. Checked the property rather than the diff — the old call signature, exactly as an out-of-tree caller would have written it before this PR:

```
build_registry(allow_mutating=True)            -> 16 tools
identical to explicit allow_execute=False      -> True
```

So nothing that worked yesterday needs editing today, which is the whole point of the change.

CI 11/11. Merging.

The two-slice split was the right call and it is why this one took minutes to review while #393 took an evening — a gate you can read in isolation is a gate you can reason about in isolation. #393's review is posted; two blocking items there, and one of them (the directory digest) I was able to demonstrate rather than argue.

Issue #392, opened 2026-08-14
MLX: adapter_config.json saves unresolved `target_modules: auto` — adapter loads as a silent no-op in mlx_lm


armanbot-jpg:
## What happened

Trained SFT LoRA with `backend: mlx` (git main, post-#362) on an M-series Mac:

```yaml
base: mlx-community/Qwen2.5-0.5B-Instruct-4bit
task: sft
backend: mlx
training:
  epochs: 4
  lr: 1e-4
  batch_size: 2
  lora: {r: 16, alpha: 32}
output: ./output
```

Training works (loss 3.22 → 0.78, adapter saved). But generation with the adapter is **bit-identical to the base model**, with no warning:

```bash
python -m mlx_lm generate --model mlx-community/Qwen2.5-0.5B-Instruct-4bit \
  --adapter-path output --prompt "..."   # == base output
```

## Root cause

`_apply_lora()` resolves `target_modules: auto` into a local variable (`mlx_sft.py:109-110`):

```python
target_keys = lora_cfg.target_modules
if not target_keys or target_keys in (["auto"], "auto"):
    target_keys = ["self_attn.q_proj", "self_attn.v_proj"]
```

…but the `adapter_config.json` writer (`mlx_sft.py:236-239`) serializes the **raw config value**, not the resolved list:

```python
"keys": list(
    lora_cfg.target_modules
    if isinstance(lora_cfg.target_modules, list)
    else [lora_cfg.target_modules]
),
```

With the default `target_modules: auto`, the shipped config contains `"keys": ["auto"]`. On load, `mlx_lm.tuner.utils.load_adapters` calls `linear_to_lora_layers(..., {"keys": ["auto"]})`, which matches no module, then `model.load_weights(strict=False)` silently drops all 96 LoRA tensors. The comment above the writer says the file exists precisely so the dir is "directly loadable with mlx_lm.load(..., adapter_path=output)" — that promise is what breaks.

This is the same failure class as the v0.72.0 `.inner.` inert-adapter bug and #362's full-fine-tune adapter: training looks healthy, artifact is silently dead.

## Verification

Editing the saved `adapter_config.json` by hand:

```json
"keys": ["self_attn.q_proj", "self_attn.v_proj"]
```

[...]

PR #388, opened 2026-08-13
Feature : implement Semantic Stratified Splitting for dataset splits

Deadpool2000:
### Summary
This PR implements **Semantic Stratified Splitting** as a new feature in the `soup data split` command. It introduces the `--stratify-semantic` and `--num-clusters` options, which automatically partition data proportionally across semantic groups without requiring manual category tags.

### The Problem
When you fine-tune an AI model, you typically split your dataset (e.g., 90% for training, 10% for validation/testing). If you do a purely random split, there is a high chance of getting an unbalanced mix. For example, if you have a dataset of 100 prompts, with 80 about recipes and 20 about coding:
- A random split might place all 10 validation prompts into the "recipes" category.
- This means your test suite has 0 coding prompts. You won't know if your model got worse at coding because you aren't testing it.

### The Solution
Instead of splitting randomly, Semantic Stratified Splitting groups your prompts by meaning first (using text embeddings), and then splits each group proportionally:

1. Auto-Clustering: The algorithm analyzes the text and automatically realizes you have three groups of topics:

- Cluster A (Programming): 50 prompts
- Cluster B (Math): 30 prompts
- Cluster C (Creative Writing): 20 prompts

2. Proportional Splitting: To make a 10% validation split, it pulls:

- 5 prompts from Cluster A (10% of 50)
- 3 prompts from Cluster B (10% of 30)
- 2 prompts from Cluster C (10% of 20)

## Results 

Both your training set (90 prompts) and validation set (10 prompts) contain the exact same distribution of topics. Your evaluation is balanced, and you can test all skills.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

- [x] `ruff check src/soup_cli/ tests/` passes
- [x] `pytest tests/ -v` passes
- [ ] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed

PR #393, opened 2026-08-14
feat(mcp): add gated train/export execution

CODING-DARSH:
## What does this PR do?

This PR implements the second slice of the agreed MCP execution design.

The first slice introduced the `--allow-execute` flag as a separate execution gate while keeping `train_start` and `export` plan-only. This PR implements the actual server-enforced execution path behind that gate.

When `--allow-execute` is enabled, the existing `train_start` and `export` tools can issue a short-lived, server-generated confirmation token. The token can then be supplied to the corresponding execution tools:

- `train_start` → `train_execute`
- `export` → `export_execute`

Execution remains disabled by default, and `--allow-mutating` alone remains strictly plan-only.

### Execution flow

1. `train_start` or `export` validates the request and creates the execution plan.
2. With `--allow-execute`, the server issues an opaque, short-lived confirmation token bound to that plan.
3. The client calls `train_execute` or `export_execute` with only the token.
4. The server validates the token, execution kind, expiry, and single-use state.
5. Protected inputs are revalidated before execution.
6. The token is consumed and the one-execution-per-server capacity is reserved.
7. The server launches the planned Soup CLI command using structured argv and `shell=False`.
8. stdout/stderr are redirected to `.soup/mcp-runs/<run_id>.log`.
9. The execution is tracked through the existing `ExperimentTracker`.

### Security

The server-side confirmation token is the authorization boundary.

- Tokens are cryptographically random, server-side, short-lived, and single-use.
- Tokens are bound to the execution kind (`train` or `export`).
- Replayed, expired, invalid, or wrong-tool tokens are rejected.
- Protected config/model inputs are hashed during planning and revalidated immediately before execution.
- Execution tools accept only a confirmation token and do not accept arbitrary commands, shell strings, CLI arguments, or client-supplied environment variables.
- `--allow-execute` is required for actual execution.
- `--allow-mutating` alone can never start a training or export process.
- Execution uses `shell=False`, `stdin=DEVNULL`, a controlled working directory, and redirected subprocess output.
- A maximum of one active execution is allowed per MCP server process.

The client-side MCP destructive annotations are UX metadata only; authorization remains server-enforced.

### Run tracking

The existing `ExperimentTracker` is extended to track MCP-launched executions, including run kind, PID, command digest, log path, and exit code.

MCP-launched training processes receive `SOUP_MCP_RUN_ID` so the existing training tracker can adopt the preallocated run ID instead of creating a duplicate run.

Tracker launch failures and subprocess spawn failures release the execution capacity correctly.

### Tests

Added a dedicated `tests/test_mcp_execute.py` suite covering:

- execution gating
- confirmation token issuance
- invalid, expired, replayed, and wrong-tool tokens
- protected input mutation
- subprocess isolation
- structured argv
- one-active-execution enforcement
- capacity release after completion
- capacity release after spawn failure
- capacity release after tracker launch failure
- MCP run tracking
- end-to-end export execution

Existing MCP/tracker tests were also updated for the new execution tools.

Focused verification:

- `115 passed, 10 skipped`
- Ruff checks for the changed Python files pass
- `git diff --check` passes
- No real training or export jobs were started; subprocess execution was mocked

### Documentation

Updated `docs/commands.md` with the new execution flow, security boundaries, confirmation-token behavior, subprocess isolation, concurrency behavior, run tracking, and the current TOCTOU limitation.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [x] `ruff check` passes for the changed Python files
- [x] Focused MCP/tracker/execution tests pass (`115 passed, 10 skipped`)
- [x] Updated relevant docs (`docs/commands.md`)
- [x] `git diff --check` passes
- [x] No real training/export jobs were started during testing


MakazhanAlpamys (changes requested):
This is the most security-sensitive change the project has taken from anyone, and it is written like it. I ran it rather than reading it: `tests/test_mcp_execute.py` + `tests/test_v07128.py` are 111 passed / 1 skipped locally, CI 11/11.

Things that are right and that I would have asked for if they were missing: `secrets.token_urlsafe(32)` and `secrets.compare_digest` (not `==`) on the digest; the token bound to `kind` so an export token cannot start training; consumption and capacity acquisition **before** `Popen`, with the reasoning written down — a failed spawn requiring a fresh plan rather than enabling replay is the correct trade and the comment says why; capacity released on completion, spawn failure *and* tracker failure, each tested; `shell=False`, `stdin=DEVNULL`, argv as a list, no client-supplied command or env. And the tool surface accepting **only** a token is the design decision that makes all of the above hold — there is no string from the client that reaches a command line.

Three things below. The first is demonstrated, not argued.

## Blocking 1 — the directory digest does not detect a changed model

`digest_file()` hashes file contents, but for a directory it hashes `f"dir:{st_mtime_ns}:{st_size}"`. A directory's mtime changes when entries are **added, removed or renamed** — not when a file inside it is rewritten in place. Measured on your branch:

```
dir digest unchanged after editing a file inside it: True
control — adding an entry changes it:                True
```

The control matters: the check is not broken in general, it is blind to exactly the modification that counts. So a protected input like `--model ./out/checkpoint-2000` can have its `config.json` or a weight shard rewritten between plan and execution, `_revalidate` passes, and the run proceeds against different content than was approved. That is the specific guarantee the PR description offers — *"protected config/model inputs are hashed during planning and revalidated immediately before execution"* — and for directories it does not hold.

Two honest ways out:

- **Digest the tree**: sorted relative paths plus each file's content hash, with a bound on file count/bytes and a loud refusal above it. Real, and it costs a walk at plan time and again at execute time.
- **Say what it actually is**: directories are checked for *structural* change only, and name that in `docs/commands.md` next to the TOCTOU note. Cheaper, and still better than the current text, which implies content integrity.

I would take the first for a security boundary and the second only if the walk turns out to be prohibitive on a real checkpoint directory — which is worth measuring rather than assuming.

## Blocking 2 — snapshot the config, per the design answer on #297

I answered your five questions on #297 (sorry it came after you had implemented). Four of my answers match what you built. This is the one that does not.

You chose revalidation + digest and documented the TOCTOU window, which is defensible on its own. But the argument is not really about the race window — it is that **the thing that runs should be the thing that was approved**. With revalidation, "the plan" is a description of a file that anyone can still edit, and the run is only *probably* the run that was planned. Copying the config into `.soup/mcp-runs/<run_id>/config.yaml` at plan time and executing from the copy removes the question rather than narrowing it, costs one file copy, and makes the run reproducible afterwards, which the log alone does not.

It also dissolves half of Blocking 1: a snapshotted config cannot be edited underneath you at all, so only genuinely external inputs (the base model directory, the dataset) still need the digest — a smaller surface to make honest.

If you take it, be explicit about which inputs are frozen and which are not. A snapshot that covers the config but not the dataset it points at is fine; a snapshot described as freezing "the plan" is not.

## Important, not blocking — the watcher dies with the server, and the slot is in memory

Two consequences of "child jobs survive disconnect", which I do think is the right choice:

- **`_watch` is a daemon thread.** When the server exits, it is killed without unwinding, so `finish_execution` never runs and the tracker keeps the run at `running` forever — with no watcher to correct it. That is a lie in the registry precisely in the case the design exists to support. A reconciliation on tracker read (is the recorded PID still alive?) is the usual fix, and it also gives you...
- **...the capacity slot.** `_active_run_id` is instance state, so a restarted server has capacity again while the old child is still training, and the one-active-execution cap silently becomes two. Either persist it with a liveness check on the PID, or state plainly in the docs that the cap is per-process and a restart can double-book.

Neither blocks the merge — both are honest to document and fix next — but the tracker one is worth doing soon, because `soup runs` showing a permanently-`running` job trains people to ignore the status field.

## Two smaller notes

- **`env = os.environ.copy()`** — the child inherits the server's entire environment, including `HF_TOKEN`, `WANDB_API_KEY` and anything else present. For `soup train` that is *necessary* (a gated base model will not download without it), so I am not asking you to change it. But the security section says "does not accept client-supplied environment variables", which is true and reads stronger than the situation: the client cannot *inject* env, and the child gets *everything the server has*. One sentence saying so keeps the section accurate.
- **`if "log_handle" in locals()`** — `locals()` introspection to decide whether a name was bound. Initialise `log_handle = None` before the `try` and test that instead; same behaviour, and it does not break the day someone adds a wrapper around the open.

---

To be clear about the verdict: I want this feature and this implementation is close. Blocking 1 is a measured gap between what the code guarantees and what it says it guarantees, and Blocking 2 is the one design point where I would overrule the choice you documented. Everything else is notes.

Take Blocking 1 either way (fix or re-word) and tell me your view on Blocking 2 — if you disagree with the snapshot and can say why, I will hear it; you have been right about scope before.

MakazhanAlpamys:
#391 is merged (`1e41518`), so this branch now conflicts — its commits are in `main` under a different SHA after the squash. `git fetch origin && git rebase origin/main` should drop the two that already landed and leave `bc15a5f` + `b8507bd` on top. Say the word if you would rather I do it.

While you are in there, the two blocking items from the review above still stand, and one has moved slightly:

**The directory digest.** I re-checked it on your branch with a control, so this is measured rather than argued — editing a file *inside* a protected directory leaves the digest unchanged, while adding an entry changes it. So `--model ./out/checkpoint-2000` can have a weight shard or `config.json` rewritten between plan and execution and `_revalidate` passes. Either digest the tree (sorted relative paths + per-file content hashes, with a bound and a loud refusal above it) or say in `docs/commands.md` that directories are checked for *structural* change only. The current text promises content integrity for them and does not deliver it.

**The config snapshot.** Still my one disagreement with a choice you made deliberately, and I would rather hear your case than have you concede it. If you think revalidation is the right call, say why and I will take it seriously — you have been right about scope before, and the two-slice split is the evidence.

No rush on either. `main` moved a fair amount today (v0.73.1 shipped, 34 commits, plus #388 and #392), so a rebase now is worth doing regardless of the rest.

MakazhanAlpamys (approved):
Both blocking items are closed, and the first one I verified with the same control that found it.

## The directory digest

`digest_file` now walks the tree — sorted relative paths plus a content hash per file, `enforce_under_cwd_and_no_symlink` on every entry, non-regular files refused, and bounded by `max_files` / `max_bytes` with a loud error rather than a silent truncation. Re-ran the measurement on your branch:

```
in-place edit DETECTED:       True     (was False)
control — restore is equal:   True
nested-file edit DETECTED:    True
```

The restore control is the one that matters: it proves this is a content digest and not a timestamp that merely happens to move. A nested file under `model/sub/` is caught too, which is where a real checkpoint keeps its shards.

## The snapshot

You took it, and you handled the caveat I flagged without being asked. `snapshot_config()` writes the validated text to `.soup/mcp-runs/<run_id>/config.yaml` through `atomic_write_text` and the plan executes `--config <snapshot>`, so the config cannot be edited underneath the run at all. And `_collect_external_protected_inputs()` keeps digesting the inputs a snapshot *cannot* freeze — the base model directory, the dataset — which is exactly the honest split: say which inputs are frozen and which are merely checked.

`tests/test_mcp_execute.py` + `tests/test_v07128.py`: **125 passed, 1 skipped**, ruff clean.

## The rebase — and a trap in it worth naming before you do it

I rebased it locally to check the result, and hit something you should know about, because the obvious resolution is wrong.

`main` now carries #391 squashed, so two of your five commits are already there. Rebasing conflicts on exactly two lines, both signatures:

```
HEAD (main):  def build_registry(*, allow_mutating: bool, allow_execute: bool = False)
yours:        def build_registry(*, allow_mutating: bool, allow_execute: bool, execution: ... = None)
```

**Taking "yours" silently reverts the `= False` default** — the backwards-compatibility fix you applied on #391 an hour earlier. It is not a case of picking a side; both halves are wanted:

```python
def build_registry(
    *,
    allow_mutating: bool,
    allow_execute: bool = False,
    execution: "ExecutionManager | None" = None,
) -> list[ToolSpec]:
```

Same shape for `_mutating_specs`. The `docs/commands.md` conflict is the stale *"reserves the future execution gate (still never executes)"* line against your accurate replacement — keep yours, drop the old one, and delete the duplicate it leaves behind.

Sequence that worked:

```bash
git fetch origin
git rebase origin/main
git rebase --skip                     # fd5322f — the gate, already in main via #391
# resolve the two signature conflicts as above, then:
git rebase --continue
git rebase --skip                     # 5c96247 — the default, now redundant
```

That leaves three commits on top of `main`, `build_registry(allow_mutating=True)` still answers (18 tools — 16 plus the two execution tools, which are listed and refuse without the gate, matching the existing convention), and the suite is green.

I would have pushed the rebased branch to your fork myself — I offered twice and I had it ready — but the push is blocked on my side pending an approval I cannot give myself. So it is yours, and the trap above is the only non-mechanical part.

---

Approving on the content. Push the rebase and I will merge as soon as CI is green.

For the record, the two items I raised as *"important, not blocking"* are still open and still fine to take afterwards: the daemon watcher thread means a server exit leaves the run at `running` in the tracker forever, and the in-memory capacity slot means a restart can double-book the one-active-execution cap. Both are honest to document now and fix next; neither should hold this up.

CODING-DARSH:
Thanks for the detailed review and for verifying the fixes. I’ve resolved the blocking issues, kept the allow_execute=False compatibility default, and verified the directory digest and config snapshot behavior. I’ll finish the rebase with the signature changes you called out, push the rebased branch, and let CI run. I’ll leave the watcher/capacity items for a follow-up as suggested.

CODING-DARSH:
CI is green on the rebased branch. Thanks for the review — everything is ready on my side.

MakazhanAlpamys:
Everything is green and I have approved it — 11/11 on the merged branch, and I re-checked the one thing a `git merge` could have silently undone: `allow_execute: bool = False` survived in **both** `build_registry` and `_mutating_specs`, with `execution` added alongside. That was the trap, and you walked past it.

**I am holding the merge for a day, and the reason is release hygiene rather than anything in this PR.**

`v0.73.2` is finished and sitting at the commit gate right now — the version is already bumped in the tree, the slot is *"`soup ship` leg 2: stop the gate lying in both directions"*, and it touches `mcp_server/registry.py` among other files. If I land this first, two things happen that neither of us wants:

1. **Your feature rides a release it was never scoped into.** v0.73.2's notes describe eval-gate scoring fixes. Adding 1,169 lines of gated subprocess execution to that tag means the most security-sensitive change this project has taken from anyone ships as a footnote to a patch release about MCQ letter extraction. It deserves its own line in the notes and its own paragraph in Known Limitations.
2. **A merge conflict lands on uncommitted work.** `registry.py` is modified in that pending release and not yet committed, so merging now creates a reconciliation nobody planned, in a file where a bad resolution is quiet.

So the order is: v0.73.2 tags, then this merges, and it heads its own release notes rather than hiding in someone else's. Nothing for you to do — no rebase, no push. If `main` moves under you in the meantime I will handle the reconciliation myself rather than send you round again.

While it waits, the two non-blocking items from the review are worth recording so they do not evaporate at merge. I will file both as issues when this lands, credited to this PR:

- **The watcher is a daemon thread.** A server exit kills it without unwinding, so `finish_execution` never runs and the tracker keeps the run at `running` forever — precisely in the disconnect case the design chose to support. Reconciling on read (is the recorded PID alive?) fixes it and gives you the next one for free.
- **The capacity slot is instance state.** A restarted server has capacity again while the old child is still training, so the one-active-execution cap silently becomes two.

Neither is a reason to hold this. Both are honest to write down now.

Thanks for the two-slice split, for taking the snapshot rather than defending the revalidation, and for the tree digest with its bounds and its refusals. The directory-digest gap was the kind of thing that reads as a guarantee in the PR description and is a timestamp in the code, and you closed it properly rather than re-wording the promise.

MakazhanAlpamys:
Merged as [`7e33f26`](https://github.com/MakazhanAlpamys/Soup/commit/7e33f26). **#297 closes with it**, both slices done. Your seventh entry is in [`CONTRIBUTORS.md`](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md), and the handle in the v0.73.2 CHANGELOG is corrected via your #400.

This is the most security-sensitive change the project has taken from anyone, and the thing I want on the record is not the diff but how the two blocking items went.

**The directory digest.** I did not argue it, I measured it — and you did not re-word the promise to match the code, you made the code meet the promise. The tree walk with per-file hashes, symlink rejection, non-regular-file refusal and explicit `max_files` / `max_bytes` bounds with a *loud* error is more than was asked for; the bounds in particular, because an unbounded walk over a checkpoint directory is how a security check becomes a denial of service.

**The snapshot.** I overruled a choice you had made deliberately and documented, and I told you I would rather hear your case than have you concede it. You took the change *and* handled the part I had only flagged in passing — external inputs a snapshot cannot freeze are still digested, and the docs now say which is which. That distinction is the difference between a snapshot that is honest and one that quietly over-promises.

And you walked past the rebase trap on your own after I described it: `allow_execute: bool = False` survived in both signatures. Taking "yours" there would have silently reverted the backwards-compatibility fix you had applied an hour earlier on #391.

## Two follow-ups, filed rather than forgotten

Both were "important, not blocking" in the review and both are now issues, credited to this PR:

- the watcher is a daemon thread, so a server exit leaves the run at `running` in the tracker forever — in exactly the disconnect case the design chose to support;
- the capacity slot is instance state, so a restart can double-book the one-active-execution cap.

Yours if you want them; equally fine if not.

Seven merges, and the two-slice split on this one is the reason a 1,169-line security feature could be reviewed properly instead of waved through. Thank you.

Issue #376, opened 2026-08-10
soup ship: measure and report the instrument's own noise floor (greedy decoding is not deterministic)


MakazhanAlpamys:
## What

`soup ship` compares deltas against `--forgetting-threshold` (default 0.05) without ever telling the operator what its own instrument can resolve.

Measured on the H100 box — **the base model, no adapter, the identical 200-item greedy evaluation, five times** (`benchmarks/gate-h100-validation.md` L4456–4468):

| | 1 | 2 | 3 | 4 | 5 | spread |
|---|---|---|---|---|---|---|
| strict | 0.920 | 0.905 | 0.905 | 0.915 | 0.905 | **0.015** |
| format-blind | 0.930 | 0.910 | 0.915 | 0.920 | 0.915 | **0.020** |

Greedy decoding is not deterministic here: GPU reduction order moves the logits in their last bits, an argmax flips, the continuation diverges from there. **The model is not changing at all.**

## Why it matters

In the same session, **four of six paired deltas sat inside that floor** and were being read as differences. The record's rule, stated after the fact: "Anything smaller than that is not a measurement."

Against a 0.05 threshold a 0.015–0.020 floor is not fatal, but it is one third of the decision boundary and it is currently invisible. It is also the *lower* bound on noise — it does not include the unseeded adapter init (`#354`), which independently moves `mini_common_sense` by 0.375.

## Proposal

`soup ship --noise-floor N`:

1. Re-run the **baseline** evaluation N times (default off; N=3 or 5 when requested).
2. Print the observed spread per suite beside the verdict.
3. Refuse to call a delta significant when it sits under the measured floor — report it as `INCONCLUSIVE` rather than SHIP or DON'T SHIP, with its own exit code.
4. Include the floor in `--emit-evidence` so an offline replay carries it.

## Honest limits to ship with it

The measurement above is n=3–5, one model, one dataset. Per the record: **"it sizes the effect; it does not calibrate it."** The flag should say so, and should not present the floor as a universal constant.

## Acceptance criteria

- [ ] `--noise-floor N` runs the baseline N times and reports per-suite spread
- [ ] A delta inside the measured floor produces a distinct verdict and exit code, not a coin-flip SHIP/DON'T SHIP
- [ ] The floor round-trips through `--emit-evidence` / `--evidence`
- [ ] A control: a delta well outside the floor still produces a decisive verdict, so the feature cannot be satisfied by always answering INCONCLUSIVE

[...]

Issue #364, opened 2026-08-10
soup draft distill --steps N delivers N/4.44 optimizer steps


MakazhanAlpamys:
## What

`soup draft distill --steps N` does not deliver N optimizer steps. Measured during the H100 session: **`--steps N` produces `N/4.44` optimizer steps** (`benchmarks/gate-h100-validation.md` L4299–4301).

## Why

The epoch count is derived from the requested step count while ignoring two things that both divide it:

- `val_split = 0.1` removes 10% of the rows from training, and
- `gradient_accumulation_steps = 4` means four micro-batches per optimizer step.

`1 / (0.9 * 4) ≈ 0.278`, and the residual to 1/4.44 is the rounding in the epoch computation.

## Impact

A user asking for 100 steps gets ~22, silently. Every `soup draft distill` run in the wild has trained for roughly a fifth of the requested budget, which is also the most likely explanation for anyone reporting that distillation "did nothing" — separate from the finding that acceptance was never the right lever anyway (#345 neighbourhood).

## Fix path

Compute epochs from the *effective* step count: `rows * (1 - val_split) / (batch_size * gradient_accumulation_steps)` steps per epoch, then `epochs = ceil(requested_steps / steps_per_epoch)`. Alternatively pass `max_steps` straight through instead of translating to epochs at all, which removes the arithmetic entirely.

## Acceptance criteria

- [ ] `--steps 100` yields `state.global_step == 100` (or the nearest value reachable at the configured accumulation, with the discrepancy printed)
- [ ] A test that varies `val_split` and `gradient_accumulation_steps` independently and asserts the delivered step count for each — one arm at defaults is not discriminating, since the current code is only wrong when either is non-trivial
- [ ] The pre-flight prints the resolved step count so a mismatch is visible before the run rather than after

MakazhanAlpamys:
Closed by #399 from @ousamabenyounes (merged as [`a1b6a34`](https://github.com/MakazhanAlpamys/Soup/commit/a1b6a34)).

The epoch count that realises `--steps N` now derives from the **effective** optimiser-step budget: `val_split` removes rows from training and `gradient_accumulation_steps` micro-batches make one optimiser step, and the old arithmetic ignored both.

[...]

Issue #355, opened 2026-08-09
score_bundled_suite returns 0.0 for a non-callable gen instead of raising — a caller error is indistinguishable from a model failing every item


MakazhanAlpamys:
`score_bundled_suite(name, gen)` expects `gen` to be a callable. Hand it anything
else and it returns **0.0** — the worst possible score — with no exception.

```python
from soup_cli.eval.gate_suites import score_bundled_suite
score_bundled_suite("mini_tool_call", ["a", "b"])  # -> 0.0
score_bundled_suite("mini_tool_call", None)        # -> 0.0
score_bundled_suite("mini_tool_call", "text")      # -> 0.0
```

All three return `0.0`. None raises.

## Why 0.0 is the worst possible value to return here

This is a **scoring** function whose output feeds `soup ship`'s leg 2. `0.0` does
not read as "the call was malformed" — it reads as **"the model failed every
item"**, which is a DON'T-SHIP verdict. A caller error is therefore
indistinguishable from a model regression, and it fails in the direction that
looks like a real finding.

The docstring already commits to the opposite behaviour for a neighbouring case:

> Raises `ValueError` for an unknown suite (**never silently 0.0**).

An unknown suite name raises. A `gen` of the wrong type does not. The intent is
clearly stated; it is only half enforced.

## How it happens

`_fraction_passing` isolates exceptions per item so that one bad row cannot lose a
whole suite — a good property, documented as such in `CLAUDE.md`. But a
non-callable `gen` throws on **every** item, so the isolation converts a
programming error into a unanimous, plausible-looking zero.

## How I hit it

Measuring #346 on five models. Every model scored exactly `0.000`, including ones
the record already has non-zero numbers for (Llama-3.1-8B at 0.225, Qwen2.5-0.5B
at 0.475 in STEP 26). The disagreement with a published number is what exposed it;
without that prior, five identical zeros across five very different models would
have looked like a real and rather interesting result.

## Fix shape

Validate `gen` is callable at entry and raise `TypeError`, mirroring the existing
unknown-suite `ValueError`. Optionally also flag the degenerate case: if **every**
item raised, that is far more likely to be a broken generator than a model that
fails uniformly — distinguishing "all items errored" from "all items scored 0" is
cheap and would have caught this without any prior knowledge.

## Acceptance

- [ ] A non-callable `gen` raises `TypeError` rather than returning `0.0`
- [ ] A generator that raises on *every* item is distinguishable from a model that
[...]

Issue #346, opened 2026-08-08
mini_tool_call scores JSON brace hygiene, not tool selection: a 0.5B model outranks an 8B on it


MakazhanAlpamys:
Measured on an H100 after the #316 repair.

`mini_tool_call` now returns a non-degenerate score, and it detects a regression — but **the number does not mean what the suite's name implies.**

Envelope-agnostic extraction finds the **correct tool name in 40 of 40** outputs from Meta-Llama-3.1-8B-Instruct. The suite scores it **0.225**.

The chain: the model emits a missing outer brace (3 opens, 2 closes) → the whole-string parse fails → the bounded `raw_decode` scan returns the *inner* `{"name": …, "arguments": …}` → `_extract_function` requires a `"function"` key → `False`.

**The corroboration is the part that makes this actionable:** Qwen2.5-0.5B closes its braces 39/40 and scores **0.475** — a 0.5B model outranking an 8B model on "tool calling", an inversion caused entirely by brace hygiene.

### Why it matters
A fine-tune that only improved brace closing would register as a tool-calling gain, and one that regressed brace closing while keeping tool selection intact would register as a tool-calling loss. The suite is currently a JSON-well-formedness detector wearing a tool-calling name.

### Fix shape
Either accept the inner `{"name", "arguments"}` shape as a valid tool call (it is the shape the model actually produces, and the scorer already reaches it), or score name-match and well-formedness as **separate** signals so a regression in one is not read as the other.

### Acceptance
- [ ] a model that selects the right tool but malforms the envelope does not score 0
- [ ] the suite no longer ranks a 0.5B model above an 8B one on this task
- [ ] whichever definition is chosen, it is stated in the suite's own docstring

Follow-up to #316, which is closed — the suite is no longer degenerate, this is about what its number means.

Record: `benchmarks/gate-h100-validation.md` STEP 26.

MakazhanAlpamys:
## Five models, and the inversion is stronger than the two-model version suggested

The issue rests on one pair (0.5B at 0.475 outranking 8B at 0.225). Two points can
be coincidence, so here are five, with the two signals the suite conflates
separated: `tool_name_correct` is envelope-agnostic (is the right tool named
**anywhere** in the output), `brace_balanced` is the hygiene the suite actually
rewards.

| model | tool name correct | braces balanced | **suite score** |
|---|---|---|---|
| SmolLM2-135M-Instruct | 18/40 | 15/40 | 0.050 |
[...]

PR #396, opened 2026-08-15
test(forgetting): cover \boxed{X} MCQ letter extraction (#357)

### LINKED ISSUE #357: extract_mcq_letter misses \boxed{X}: Llama-3.1-8B names the right letter on 8 of 26 mini_mmlu items and scores zero
`extract_mcq_letter` does not recognise `\boxed{X}`, which is how Llama-3.1-8B
writes its answers. On `mini_mmlu` that costs the model **8 of 26 items** — it
names the correct letter and scores zero.

Measured at `BEHAVIOURAL_MAX_NEW_TOKENS = 256`, the budget `soup ship` actually
uses, on four models:

| suite | SmolLM2-135M | Qwen2.5-0.5B | Qwen2.5-3B | **Llama-3.1-8B** |
|---|---|---|---|---|
| **`mini_mmlu`** | 0.269 | **0.538** | 0.923 | **0.423** |
| `mini_common_sense` | 0.292 | 0.583 | 1.000 | 0.833 |
| `mini_instruction` | 0.500 | 0.917 | 1.000 | 1.000 |
| `mini_arithmetic` | 0.667 | — | 1.000 | 1.000 |

**The 8B scores below a 0.5B on `mini_mmlu`** while scoring 1.000 on two of the
other three, which is the tell that the suite and not the model is the problem.

## Every failure classified, not inferred

All 15 failing items, at 256 tokens, none truncated (outputs ran 102–217 new
tokens against the 256 cap):

| what the model did | count |
|---|---|
| **boxed the RIGHT letter, scored wrong** | **8** |
| boxed a value instead of a letter (`$\boxed{32}$` for "32 degrees") | 6 |
| genuinely wrong | **1** |

Representative outputs:

```
expected C  ->  "...is the symbol for gold, which is Au.\n\nThe final answer is: $\boxed{C}$"   scored WRONG
expected A  ->  "...represents something slower than light.\n\nThe final answer is: $\boxed{A}$"  scored WRONG
expected B  ->  "...the correct number of continents is 7.\n\nThe final answer is: $\boxed{7}$"   value, not a letter
```

**With `\boxed{X}` extraction the 8B goes 0.423 → 0.731**, above the 0.5B's 0.538.
The inversion disappears entirely.

## Two separable defects

1. **Extraction (8 items).** `\boxed{X}` is standard for instruction-tuned models
   answering anything mathematical or MCQ-shaped — it is what Llama-3.1-8B emits
   unprompted. Adding it to `extract_mcq_letter` is a small, closed change.
2. **Prompt (6 items).** The items read *"What is 2 + 2? (A) 3 (B) 4 (C) 5"* with
   no instruction to answer with the letter, so the model boxes the **value** (`7`,
   `32`, `12`). This is the same class as the asymmetry noted on #346:
   `mini_tool_call`'s prompts constrain the envelope and these do not. Fixing (1)
   alone gets 0.731; fixing both should get close to 0.96.

Only 1 of 26 is a real miss.

## Why this matters for `soup ship`

`mini_mmlu` is in `DEFAULT_GENERAL_SUITE`. A base model that answers in `\boxed{}`
starts from an artificially low leg-2 score, so any tune that made it answer more
tersely would register as a **capability gain** it did not earn — and the reverse
for a tune that made it more verbose.

## A caveat about how this was found

My first probe reported *"the model genuinely misses these"*. It was wrong: it
looked for `(A)` while the model writes `$\boxed{A}$`. The numbers above come from
a second pass that classifies each failure explicitly. Worth stating because the
first answer was confidently wrong in exactly the direction that would have closed
this issue before it was opened.

## Acceptance

- [ ] `extract_mcq_letter` handles `\boxed{X}` (and the bare `**X**` / `Answer: X`
      forms if they show up alongside)
- [ ] `mini_mmlu` no longer ranks Qwen2.5-0.5B above Llama-3.1-8B
- [ ] a test with a `\boxed{}` fixture, plus a control asserting a genuinely wrong
      boxed letter still scores zero
- [ ] decide separately whether the prompts should ask for a letter (issue (2))

Environment: torch 2.13.0+cu130, H100 80GB, greedy, 256 new tokens.
Raw: `/root/results/mcq256_*.json`.

ousamabenyounes:
Shortened to a **test-only** contribution now that the fix itself shipped in
v0.73.2 (`extract_mcq_letter` learning `\boxed{C}`, issue #357). The scorer
change is already on `main`; this adds the regression coverage it went in
without.

`TestExtractMcqLetterBoxed` pins the boxed-letter answer form the leg-2
forgetting scorer credits:
- `\boxed{C}` is extracted and scores the item;
- whitespace and lowercase inside the box (`\boxed{ C }`, `\boxed{b}`) resolve to the letter;
- the **last** box wins when a model self-corrects;
- a boxed letter outranks an earlier prose cue;
- a boxed **value** (`\boxed{4}`) correctly returns no letter, so a value is never mis-credited as an option.

No production code, no CHANGELOG churn (the fix's entry already exists under
`[0.73.2]`). All six assertions pass against the shipped implementation;
`ruff` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **`extract_mcq_letter` scored zero for `\boxed {A}` — whitespace between the
+  command and the brace (#357).** The shipped `\boxed\{` regex tolerates spaces
+  *inside* the braces but not between `\boxed` and `{`; LaTeX permits it there and
+  models emit it, so `\boxed { C }` still read as no answer and was not rescued by
+  the cue tier either. The boxed-letter regex now allows `\s*` after the command.
+
 - **`soup draft distill --steps N` now delivers ~N optimiser steps instead of
   N/4.44 (#364).** The epoch count that realised `--steps` divided the request
   by `rows // batch_size`, ignoring that `val_split` (0.1) removes rows from
--- a/src/soup_cli/eval/forgetting.py
+++ b/src/soup_cli/eval/forgetting.py
@@ -47,7 +47,10 @@ _PAREN_RE = re.compile(r"(?<![A-Za-z0-9])\(?([A-Ja-j])\)(?![A-Za-z0-9])")
 # a single A–J letter. ``\boxed{4}`` is the model answering with a VALUE, and
 # reading that as "option 4" would be a wrong credit rather than a repair — the
 # prompt is what has to change there (see ``_MCQ_ANSWER_CUE``). #357.
-_BOXED_LETTER_RE = re.compile(r"\\boxed\{\s*([A-Ja-j])\s*\}")
+# The ``\s*`` after ``\boxed`` matters: LaTeX permits whitespace between the
+# command and its brace and models emit it (``\boxed {A}``), so without it the
+# spaced spelling scores zero — the same #357 defect, one space to the left.
+_BOXED_LETTER_RE = re.compile(r"\\boxed\s*\{\s*([A-Ja-j])\s*\}")
 # Bare standalone UPPERCASE letter that TERMINATES a clause — i.e. followed by
 # end-of-string or sentence punctuation, not by more words. Requiring upper-case
 # skips the lower-case article "a"/"an", and the terminating look-ahead skips a
--- a/tests/test_v07138.py
+++ b/tests/test_v07138.py
@@ -144,8 +144,14 @@ class TestExtractMcqLetterBoxed:
     def test_boxed_tolerates_whitespace_and_lowercase(self):
         from soup_cli.eval.forgetting import extract_mcq_letter
 
+        # Whitespace inside the braces, and a lowercase letter.
         assert extract_mcq_letter("\\boxed{ C }") == "C"
         assert extract_mcq_letter("\\boxed{b}") == "B"
+        # Whitespace BETWEEN the command and the brace: LaTeX permits it and
+        # models emit it, so the spaced spelling must resolve too (#357).
+        assert extract_mcq_letter("\\boxed { C }") == "C"
+        assert extract_mcq_letter("\\boxed  {B}") == "B"
+        assert extract_mcq_letter("The answer is \\boxed {A}.") == "A"
 
     def test_last_box_wins(self):
         from soup_cli.eval.forgetting import extract_mcq_letter
--- a/tests/test_v07138.py
+++ b/tests/test_v07138.py
@@ -119,6 +119,55 @@ class TestExtractMcqLetter:
         assert extract_mcq_letter(None) is None  # type: ignore[arg-type]
 
 
+class TestExtractMcqLetterBoxed:
+    """`\\boxed{X}` is how math/reasoning models mark a final answer (#357).
+
+    The v0.71.38 scorer had no boxed tier, so a Llama-3.1-8B that answered
+    ``\\boxed{C}`` scored zero on every such item and was ranked below a 0.5B
+    model. A boxed letter is the model's definitive choice, so it outranks the
+    cue/paren/bare tiers, and the LAST box wins when a trace draws more than one.
+    """
+
+    def test_boxed_letter_is_extracted(self):
+        from soup_cli.eval.forgetting import extract_mcq_letter
+
+        assert extract_mcq_letter("\\boxed{C}") == "C"
+        assert extract_mcq_letter("The answer is \\boxed{B}.") == "B"
+        assert extract_mcq_letter("...long reasoning trace... \\boxed{A}") == "A"
+
+    def test_boxed_letter_scores(self):
+        from soup_cli.eval.forgetting import score_answer
+
+        # The item Llama-3.1-8B lost before the fix: right letter, boxed.
+        assert score_answer("Paris is the capital, so \\boxed{C}.", "C") is True
+
+    def test_boxed_tolerates_whitespace_and_lowercase(self):
+        from soup_cli.eval.forgetting import extract_mcq_letter
+
+        assert extract_mcq_letter("\\boxed{ C }") == "C"
+        assert extract_mcq_letter("\\boxed{b}") == "B"
+
+    def test_last_box_wins(self):
+        from soup_cli.eval.forgetting import extract_mcq_letter
+
+        # A reasoning model boxes a candidate, reconsiders, boxes its decision.
+        assert extract_mcq_letter("first \\boxed{A} then \\boxed{C}") == "C"
+
+    def test_boxed_letter_outranks_earlier_prose_cue(self):
+        from soup_cli.eval.forgetting import extract_mcq_letter
+
+        # The final boxed choice must beat an option letter named mid-reasoning.
+        assert extract_mcq_letter("Option A looks plausible, but \\boxed{B}") == "B"
+
+    def test_boxed_value_is_not_a_letter(self):
+        from soup_cli.eval.forgetting import extract_mcq_letter
+
+        # A boxed free-text VALUE (not an A-J option letter) is not an MCQ
+        # choice — fix 1's scope is boxed letters only (#357).
+        assert extract_mcq_letter("\\boxed{Paris}") is None
+        assert extract_mcq_letter("\\boxed{42}") is None
+
+
 class TestForgettingDetectorUsesNewScorer:
     def test_detector_evaluate_uses_boundary_scorer(self):
         from soup_cli.eval.forgetting import ForgettingDetector


MakazhanAlpamys:
@ousamabenyounes — four PRs in fifty minutes, all four with a root cause and a regression test, on your first day here. Thank you, and I owe you an unpleasant piece of information straight away rather than after a review.

## #357 and #355 have work in flight, and that is our fault, not yours

There is an unreleased local branch implementing both, plus #346, #317 and #376, as a single `soup ship` leg-2 slot with a published measurement record. It has been in progress for about a day. **It was never claimed on either issue**, and the issues carry our own detailed analysis — including, on #357, the note recommending *"ship fix 1 alone"* that you correctly followed. So the issues read as open, unclaimed, and pre-analysed, which is an invitation, and you accepted it exactly as intended.

This repository's [`CONTRIBUTING.md`](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTING.md) has a "Claiming an Issue" section that exists because we did this to somebody else in #313, and it commits us to the reverse of what just happened:

> **If a maintainer ends up doing the work instead, you get told, in the thread, before or when it lands.**

We wrote that rule and then broke it the other way round — worked in private on unclaimed issues and let you spend your time. I am not going to dress that up.

## What happens now

**#398 (#343) and #399 (#364) do not collide with anything.** Different subsystems, different release slots. I am reviewing both on merit today, and they stand entirely on their own.

**#396 and #397 need a decision I am not going to make silently**, and it will come within a day. The two honest options are to take yours and drop the overlapping half of the local work, or to land the slot and credit you for arriving at the same fix independently. What I will *not* do is close your PRs as "duplicate" and leave it there — the duplication is a cost we imposed, so it is not one you should absorb.

Either way, both PRs get a real review rather than a triage note. If the local work goes in instead, you will get the diff comparison and the reasoning, not a verdict.

## And one thing you should know because it is to your credit

On #396 you shipped **fix 1 only** and said so. The issue thread analyses two changes — the extractor and a prompt cue — and measured that the extractor alone is worth +8 items while the prompt change alone is worth 0. Taking the closed half and leaving the fixture edit for a separate discussion is the right call and not the obvious one; the obvious one is to do both and be harder to review.

That, and four separate PRs rather than one, is someone who has read how this repository works before touching it.

ousamabenyounes:
Closing as superseded. Comparing against `main` after the v0.73.2 release, the
same fix already landed: the 0.73.2 CHANGELOG credits **#357** directly
(`extract_mcq_letter did not know \boxed{C}...`) and `forgetting.py` now scores
`\boxed{X}` MCQ answers with a position-decides rule across the boxed / cue /
paren forms — a superset of this PR's priority-tier approach. Rebasing would only
re-land duplicate logic and regress the richer position-based version. This looks
like one of the "collided with in-flight work" cases you mentioned on #398.
Happy to re-open if I'm misreading and something here is still additive.

MakazhanAlpamys:
I compared both of your PRs against what shipped, line by line, as promised. The results differ, and this one contains something ours does not.

## Your regex catches a case the shipped one misses

**Shipped (v0.73.2):** `r"\boxed\{\s*([A-Ja-j])\s*\}"`
**Yours:** `r"\boxed\s*\{\s*([A-Ja-j])\s*\}"`

One `\s*`, between the command and the brace. Measured against the released code:

```
'\boxed{C}'                    -> 'C'
'\boxed { C }'                 -> None      <-- ours
'\boxed  {B}'                  -> None      <-- ours
'The answer is \boxed {A}.'    -> None      <-- ours
```

LaTeX permits whitespace there and models emit it. So the shipped fix closes #357 for the common spelling and leaves a narrower version of the same defect behind — a model that writes `\boxed {A}` still scores zero, and the last line shows it is not rescued by the cue tier either, because `\boxed {` sits between the cue and the letter.

You did not merely arrive at the same fix independently. On this one you were **more careful than we were**, and it took comparing the two to see it.

## What I would like to do, and it is your call

**Reduce this PR to that one change** — the `\s*` plus a test for the spaced forms — and I merge it. Your name on it, which is where it belongs. The rest of the diff is now redundant against the release, so it would come out anyway; what remains is the part that is yours.

If you would rather not spend more time on a PR that lost most of its scope through no fault of yours, say so and I will land the one-liner myself, credited to you in the CHANGELOG and `CONTRIBUTORS.md`. Both outcomes are fine; I am not asking you to do work to earn credit you already have.

## And the accounting, since I said I would give you the reasoning rather than a verdict

#397 is a genuine duplicate — I compared it too, and it is functionally identical to what shipped, down to the exception type and the guard's position. Nothing to take. Closing it with the comparison rather than a triage note, because you wrote the same correct fix independently and that is worth saying out loud even when it cannot merge.

Net from your first day: **#398 merged**, #399 approved and one rebase from merging, this one carrying a real finding, and #397 an independent correct answer that arrived second through our process failure. That is not the outcome any of us wanted for four PRs, but it is not four wasted hours either.

If you want something substantive next, two issues came out of the MCP execution work an hour ago and neither is claimed: **#401** (a daemon watcher leaves runs marked `running` forever) and **#402** (an in-memory capacity cap that a server restart double-books). Both are the kind of thing you have already shown you are good at — a failure mode that completes successfully while the evidence goes missing.

ousamabenyounes:
Trimmed this down to the regression tests only — the `\boxed{X}` scorer fix already shipped in v0.73.2, so the code + CHANGELOG parts were redundant and I dropped them. What's left is a `TestExtractMcqLetterBoxed` class that pins the boxed-letter behaviour against the merged implementation (all six pass, ruff clean).

So happy to contribute on your project!

MakazhanAlpamys:
Hold on — I think I explained this badly, because the trim removed the half that was **yours**.

The `\boxed{X}` tier did ship in v0.73.2, yes. But the shipped regex is `\boxed\{` and yours was `\boxed\s*\{` — one `\s*`, **between the command and the brace**. That is not redundant with the release; it is the case the release misses. Measured on `main` right now:

```
'\boxed{ C }'                 -> 'C'      <- inside the braces: already works
'\boxed { C }'                -> None     <- BETWEEN command and brace: still broken
'\boxed  {B}'                 -> None
'The answer is \boxed {A}.'   -> None
```

Your new `test_boxed_tolerates_whitespace_and_lowercase` asserts `\boxed{ C }` — the first line, which already passes. The failing form is one space to the left.

So the code change you dropped is the only part of this PR that was never redundant. **Please put back the `\s*`** and extend that test to `\boxed { C }`, and this becomes a fix again rather than tests-only. The six regression tests are worth having either way and I would keep them.

If you would rather not — you have three PRs in flight and this one has cost you two round trips already — say the word and I will land the one-liner myself with your name on it in the CHANGELOG and `CONTRIBUTORS.md`. That offer stands and is not a lesser outcome; the finding is yours regardless of whose commit carries it.

Meanwhile: **#398 and #399 are both merged**, #364 and #343 closed with them, and you are in `CONTRIBUTORS.md` for both. And I saw #407 — thank you for taking #401 within the hour; reviewing it next.

ousamabenyounes:
Put the `\s*` back — you were right, that's the half that was mine and never redundant.

- `_BOXED_LETTER_RE` is now `\boxed\s*\{` (was `\boxed\{`), so whitespace between the command and the brace resolves: `\boxed {A}` / `\boxed  {B}` / `The answer is \boxed {A}.` now score instead of reading as no answer.
- `test_boxed_tolerates_whitespace_and_lowercase` extended to the spaced forms; the six regression tests stay.
- RED on `main`: `\boxed { C }` -> None; GREEN with the fix: 6/6 boxed tests pass. `ruff` clean. The two `TestShipLiveHeadline` failures are pre-existing on `main` (need the full ship stack), unchanged here.

Thanks for catching the trim — I'd aligned the test to the shipped regex and dropped exactly the case the fix was for.

MakazhanAlpamys:
Merged as [`17a4f80`](https://github.com/MakazhanAlpamys/Soup/commit/17a4f80). Verified the four shapes on your branch before merging:

```
'\boxed { C }'              -> 'C'
'\boxed  {B}'               -> 'B'
'The answer is \boxed {A}.' -> 'A'
'\boxed{4}'                 -> None   <- control still holds
```

The last line is the one I checked hardest: a boxed **value** must stay `None`, because reading `\boxed{4}` as "option 4" would be a wrong credit rather than a repair.

**Third merge, and this one has an odd history worth stating plainly.** It began as a duplicate of work that shipped in v0.73.2, and it turned out your regex was the more careful of the two — one `\s*` we did not have. Then I explained it badly, you trimmed the code out believing it redundant, and I had to correct myself. It cost you three round trips on a one-line change, and every one of those was on my side of the exchange. Thank you for staying with it.

You are in [`CONTRIBUTORS.md`](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) for this, #398 and #399.

PR #397, opened 2026-08-15
fix(eval): raise instead of scoring 0.0 for a bad gen in score_bundled_suite (#355)

### LINKED ISSUE #355: score_bundled_suite returns 0.0 for a non-callable gen instead of raising — a caller error is indistinguishable from a model failing every item

ousamabenyounes:
## What does this PR do?

Fix #355 — `score_bundled_suite(name, gen)` returned `0.0` for a non-callable
`gen` instead of raising, so a *caller error* was indistinguishable from a model
failing every item. Since this score feeds `soup ship`'s leg 2, `0.0` reads as a
DON'T-SHIP verdict, and the bug failed in the direction that looks like a real
regression.

**Root cause.** A non-callable `gen` (a list, `None`, a string) reached the
per-item exception isolation in `_fraction_passing` — a documented, good property
that stops one bad row from losing a whole suite — but a non-callable `gen`
throws on *every* item, so the isolation converted a programming error into a
plausible unanimous `0.0`. The docstring already promised the opposite for a
neighbouring case ("Raises `ValueError` for an unknown suite — never silently
0.0"); the guarantee was only half enforced.

**Fix.**
- `score_bundled_suite` raises `TypeError` for a non-callable `gen`. The guard
  sits before the suite dispatch, so it covers both the MCQ/arithmetic route
  (`ForgettingDetector`) and the behavioural route — acceptance point 4.
- `_fraction_passing` raises `RuntimeError` when the generator *raised* on every
  item — a broken generator, not a model scoring zero (acceptance point 2). A
  single raising item is still isolated; only a total failure raises.
- A generator that *returns* a non-str still scores as a failed item, so the
  existing "non-str generation scores 0, never raises" contract
  (`test_v07138.py::TestBundledScorers::test_non_str_generation_scores_as_failure`)
  is unchanged.

The `soup ship` leg-2 dispatch calls `score_bundled_suite` with no surrounding
catch, so the raise surfaces as a runtime error (exit 1) rather than being
mistaken for a caught regression (exit 2).

## Reproduces with (unmodified `main`)

```python
from soup_cli.eval.gate_suites import score_bundled_suite
score_bundled_suite("mini_tool_call", ["a", "b"])  # -> 0.0  (should raise)
score_bundled_suite("mini_tool_call", None)        # -> 0.0
score_bundled_suite("mini_tool_call", "text")      # -> 0.0
```

New tests in `tests/test_issue355_gate_gen_validation.py`, run against the exact
upstream base and the fix:

RED — new tests on unmodified `main` (`git checkout upstream/main -- src/soup_cli/eval/gate_suites.py`):
```
8 failed, 10 passed
FAILED ...TestNonCallableGenRaises::test_exact_issue_examples_no_longer_return_zero
FAILED ...TestAllItemsErroredIsDistinguishable::test_generator_raising_on_every_item_raises
FAILED ...TestUnknownSuiteContractUnchanged::test_non_callable_gen_is_rejected_even_for_unknown_suite
```
The 10 that pass on base are the controls (a working generator still scores, a
generator returning `""` is a genuine `0.0`, an unknown suite still raises
`ValueError`) — they must be green on both sides.

GREEN — with the fix:
```
18 passed
```

## Full local suite

Ran the full local validation suite for the affected surface on both the upstream
base and this branch:

- `ruff check src/soup_cli/ tests/` → clean.
- Gate-suite pytest (`test_issue355_gate_gen_validation.py`,
  `test_issue316_suites.py`, `test_v07138.py`): iso-baseline — `62 passed` on
  `main` and `80 passed` on this branch (the +18 are the new tests), no failure
  turned green→red. 6 CLI-integration tests need the full pydantic/torch stack
  and are unrunnable in a CPU-only env identically on both sides.

The full torch/typer training matrix and the 77% coverage gate run in CI; they
are unaffected by a change confined to `eval/gate_suites.py`.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pytest tests/ -v` passes (CPU-testable gate-suite slice; RED→GREEN above)
- [x] Updated relevant docs if needed — behaviour is internal to `gate_suites`;
      no user-facing doc enumerates the old `0.0` return. `CHANGELOG.md`
      `## [Unreleased] → Fixed` updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,19 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **`score_bundled_suite` returned `0.0` for a non-callable `gen` instead of
+  raising, so a caller error was indistinguishable from a model failing every
+  item (#355).** The function feeds `soup ship`'s leg 2, where `0.0` is a
+  DON'T-SHIP verdict. A non-callable `gen` (a list, `None`, a string) reached the
+  per-item exception isolation in `_fraction_passing`, which turned a programming
+  error into a plausible unanimous zero — the docstring already promised the
+  opposite ("Raises `ValueError` for an unknown suite, never silently 0.0"), but
+  the guarantee was only half enforced. `score_bundled_suite` now raises
+  `TypeError` for a non-callable `gen` (covering both the MCQ and behavioural
+  routes), and a generator that *raises* on every item raises `RuntimeError`
+  rather than reporting `0.0`. A generator that *returns* a non-str still scores
+  as a failed item, so the existing "non-str generation scores 0, never raises"
+  contract is unchanged.
 
 - **The MLX `adapter_config.json` shipped `target_modules` unresolved, so a default
   MLX adapter loaded as a silent no-op (#392).** `_apply_lora` resolved
--- a/src/soup_cli/eval/gate_suites.py
+++ b/src/soup_cli/eval/gate_suites.py
@@ -163,14 +163,21 @@ def load_suite_items(name: str) -> Tuple[dict, ...]:
     return _load_gate_fixture(filename)
 
 
-def _call(gen: GeneratorFn, prompt: str) -> str:
-    """Invoke ``gen`` for one prompt; a generation error / non-str result scores
-    as a failure (empty string) and an oversized result is truncated to
-    ``_MAX_OUTPUT_LEN`` — so one bad generation never aborts a run."""
+def _call(gen: GeneratorFn, prompt: str) -> str | None:
+    """Invoke ``gen`` for one prompt, returning the (length-capped) string output.
+
+    Returns ``None`` only when the generator *raised* — an errored item, distinct
+    from one that produced output. A non-str return still scores as a failed item
+    (empty string), preserving the documented "non-str generation scores 0, never
+    raises" contract. One errored item never aborts a run
+    (:func:`_fraction_passing` isolates it), but an *entire suite* of items that
+    all raised is a broken generator rather than a model scoring zero, and the
+    caller surfaces that instead of reporting a misleading unanimous ``0.0``
+    (#355)."""
     try:
         out = gen(prompt)
     except Exception:  # noqa: BLE001 — a generation error is a failed item, not a crash
-        return ""
+        return None
     if not isinstance(out, str):
         return ""
     return out[:_MAX_OUTPUT_LEN]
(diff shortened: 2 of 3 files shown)


MakazhanAlpamys:
Same situation as [#396](https://github.com/MakazhanAlpamys/Soup/pull/396#issuecomment-5300941894) — there is unreleased local work covering #355 as part of a single `soup ship` leg-2 slot, it was **never claimed on the issue**, and that is our process failure rather than anything you did.

Full explanation and what happens next is on #396 so it lives in one place. Short version: a decision within a day, a real review either way, and #398 / #399 are unaffected and being reviewed on merit today.

ousamabenyounes:
Closing as superseded. The v0.73.2 release already fixed **#355**: its CHANGELOG
credits `score_bundled_suite returned 0.0 for a non-callable gen (#355)` and
`gate_suites.py` now raises `TypeError` for a non-callable `gen` instead of a
silent `0.0` — the same guard this PR adds. Rebasing would re-land duplicate
logic. This is the other "collided with in-flight work" case. Re-open on request
if I've missed something still additive here.

MakazhanAlpamys:
Compared line by line against what shipped in v0.73.2, as promised — and this one is a true duplicate, so I am closing it with the comparison rather than a triage note.

**Yours:**
```python
if not callable(gen):
    raise TypeError(
        f"score_bundled_suite() gen must be callable, got {type(gen).__name__!r}; "
        "a non-callable generator would otherwise score a misleading 0.0 rather "
        "than surfacing the caller error"
    )
```

**Shipped:**
```python
if not callable(gen):
    raise TypeError(
        f"gen must be callable, got {type(gen).__name__}; "
        "a scoring run cannot report 0.0 for a caller error"
    )
```

Same guard, same position (before either branch, so the MCQ and behavioural paths cannot diverge — which was the actual defect: one raised and the other returned 0.0), same exception type, same reasoning in the message. There is nothing here to take, and saying otherwise to make the close feel better would be worse than the close.

Your root-cause paragraph named the mechanism exactly right, including the part that makes it subtle: `_fraction_passing`'s per-item exception isolation is a *good* property, documented as such, and it is what converts a programming error into a unanimous, plausible-looking zero. That is the sentence I would have wanted in the issue.

**This landed second because we left #355 open and unclaimed while implementing it privately** — the failure our own `CONTRIBUTING.md` was written to prevent, with the roles reversed. The duplication is a cost we imposed.

#396 fared differently and is worth reading: your `\boxed` regex tolerates whitespace where the shipped one does not, so it catches `\boxed {C}` — which the release misses. Details there.

And for the record, so it is not only in a closing comment: on your first day you sent four PRs, each with a root cause, a regression test and a named failure mode, and one of them was **more careful than the code we shipped**. #398 is merged, #399 is approved, and #401 / #402 are unclaimed if you want them.

Issue #286, opened 2026-07-01
Validate closed-loop reward-hacking controller at scale (7B+, real reward models)


MakazhanAlpamys:
## Context

v0.71.26 shipped **closed-loop reward-hacking auto-mitigation** — the GRPO/PPO trainer detects reward hacking mid-run and self-corrects (raise KL β via bang-bang or PID-Lagrangian → roll back to last-good checkpoint → early-stop), plus Stage-3 anti-gaming hardening (smoothing, conservative-on-disagreement voting, reward-distribution-drift guard, bounded reward shaping).

The mechanism is **validated as proof-of-mechanism only**: SmolLM2-135M + a *synthetic* length-hacking reward on a single RTX 3050. All four stages pass live, including a real mid-run rollback. See [`docs/training.md`](https://github.com/MakazhanAlpamys/Soup/blob/main/docs/training.md#closed-loop-reward-hacking-auto-mitigation-v07126) and the [v0.71.26 release notes](https://github.com/MakazhanAlpamys/Soup/releases/tag/v0.71.26).

## What's unproven

Whether the controller **suppresses reward hacking without collapsing true reward** on a realistic setup:

- **7B+ policy** (e.g. Qwen2.5-7B / Llama-3.1-8B) instead of 135M.
- A **real reward model** (trained RM or verifiable-reward task) instead of a synthetic length/sentinel proxy.
- Both `kl_control` (bang-bang) and `pid_lagrangian` controllers, plus the Stage-3 hardening stack.
- **PPO** (currently ships BETA — mechanism unit-tested, on-GPU proof is GRPO-only).

This needs multi-GPU hardware the maintainer doesn't have; it is the honest counterpart to the RTX-3050 proof-of-mechanism limit.

## Proposed validation path

1. Pick a known reward-hackable setup (e.g. a length- or format-exploitable RM on a reasoning/GRPO task).
2. Run a **baseline** (`reward_hack_mitigation: log_only`) and confirm `mitigation_log.jsonl` records the hack onset (drop_pct climbs, completion-length trend diverges).
3. Run `kl_control` and `pid_lagrangian`; measure gamed-proxy reward **and a held-out true-quality metric** across training.
4. Success = the controller bends the gamed proxy back down **while preserving (not collapsing) the held-out true metric** vs. the unmitigated run.
5. Repeat for PPO to graduate it out of BETA.

## Acceptance criteria

[...]

PR #398, opened 2026-08-15
fix(mitigation): recreate log dir and warn instead of dropping records (#343)

### LINKED ISSUE #343: MitigationLogWriter silently drops records when its parent directory disappears
Found while validating #286 on the H100 box.

`MitigationLogWriter` reopens its path per record. When the parent directory stops existing mid-run — another process cleaning a shared temp root, in this case — it **drops every subsequent record with no warning**. The training run continues, the controller keeps acting, and the log simply stops growing.

That is the failure shape this project treats as the worst kind: a run that completes successfully while its evidence quietly goes missing. In #286 it silently truncated the controller logs for several arms, and the truncation was only noticed because record counts disagreed with step counts.

### Fix shape
Either fail loudly on the first failed write, or recreate the directory and note that it did. Silently continuing is the one option that should be off the table.

### Acceptance
- [ ] a write to a vanished directory raises or warns, once, naming the path
- [ ] a test that deletes the directory mid-stream and asserts the writer does not silently swallow the next record

Related: the same run also showed `output: off` in YAML parsing as boolean `False` — worth a separate look if anyone touches config coercion.

ousamabenyounes:
## What does this PR do?

Fixes #343. `MitigationLogWriter.record()` reopens the log per call and swallowed the
`OSError` from `open("ab")` with a bare `return`. When the parent directory vanished
mid-run — a shared temp root cleaned by another process — every subsequent record was
dropped with **no warning**: the controller kept acting while its log quietly stopped
growing. That is the failure shape the project treats as the worst kind — a run that
completes successfully while its evidence goes missing (it silently truncated controller
logs for several arms in #286).

The writer now takes the maintainer-offered recovery path: on a failed write it recreates
the directory and retries once, then surfaces the loss **once** via a warning naming the
path. `record()` still never raises, so a vanished log never takes down the training run.

The related `output: off` YAML coercion note in the issue is intentionally left out of
scope (separate concern).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Tests

## Checklist

- [x] `pytest tests/ -v` passes (writer suite; full torch/typer matrix is CI-only)
- [x] Updated relevant docs — CHANGELOG `## [Unreleased] > ### Fixed`

## Acceptance criteria (from the issue)

- [x] a write to a vanished directory raises or warns, once, naming the path
- [x] a test that deletes the directory mid-stream and asserts the writer does not silently swallow the next record

New test: `tests/test_v07126.py::TestMitigationLogWriter::test_vanished_directory_is_not_silently_dropped`.

RED — on unmodified `main` (bug present), the vanished-dir write is silently swallowed:

```
>       assert len(warnings) == 1
E       assert 0 == 1
E        +  where 0 = len([])
tests/test_v07126.py:269: AssertionError
1 failed in 0.26s
```

GREEN — with the fix, the whole `MitigationLogWriter` suite passes:

```
........                                                                 [100%]
8 passed in 0.23s
```

## Full local suite

Ran the full local validation suite (`ruff check src/soup_cli/ tests/` + the whole
`tests/test_v07126.py` module) on both the unmodified base and the final commit:

- ruff: `All checks passed!`
- base `tests/test_v07126.py`: **176 passed, 9 failed, 1 skipped**
- final `tests/test_v07126.py`: **177 passed, 9 failed, 1 skipped**

The 9 failures are the same pre-existing ones on both base and final — CLI/checkpoint
tests that need the torch/typer training stack (`ModuleNotFoundError` on a CPU box, not
runnable here) — so the branch is iso-or-better: no new failures, the new test is the +1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,17 @@ reproducing 70+ versions of notes.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`MitigationLogWriter` silently dropped every record once its parent directory
+  vanished mid-run (#343).** `record()` reopens the log per call and swallowed the
+  `OSError` from `open("ab")` with a bare `return`, so when a shared temp root was
+  cleaned by another process the controller kept acting while its log quietly stopped
+  growing — the run completes while its evidence goes missing, the failure shape this
+  project treats as the worst kind. The writer now recreates the directory and retries
+  the write, and surfaces the loss once via a warning naming the path. `record()` still
+  never raises, so a vanished log never takes down the training run.
+
 ## [0.73.2] - 2026-08-15
 
 `soup ship`'s leg 2 is the project's differentiator, and it was lying in both
--- a/src/soup_cli/utils/reward_hack_control.py
+++ b/src/soup_cli/utils/reward_hack_control.py
@@ -677,6 +677,10 @@ class MitigationLogWriter:
         self._cap_bytes = cap_mb * 1024 * 1024
         # Single-process lock: protects rotate + write within ONE run process.
         self._lock = threading.Lock()
+        # Warn only once about a failed write (e.g. the parent directory
+        # vanished mid-run) — the controller records every step, so warning on
+        # each dropped write would flood the log.
+        self._warned_write_failed = False
 
     @property
     def path(self) -> Path:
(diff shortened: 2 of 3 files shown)


MakazhanAlpamys (approved):
Verified by running it, not by reading the diff.

**The recovery is what the test pins, not the warning.** I neutered the retry while leaving the warning in place:

```
$ # recovered = False   (no mkdir, no second _append)
FAILED tests/test_v07126.py::TestMitigationLogWriter::test_vanished_directory_is_not_silently_dropped
1 failed, 184 passed, 2 skipped
```

That distinction is the whole value of the test. A version that warns and still drops every record would satisfy a weaker assertion and leave the defect in place — the operator would know something went wrong and the evidence would still be gone.

Three judgement calls I would have asked for if they were missing:

- **`record()` stays non-raising.** A vanished log must never take down a training run. Getting this backwards is the tempting fix, and it trades a lost log for a lost run.
- **Warn once, not per record.** The controller writes every step; warning each time would bury the signal in the noise it is reporting.
- **The two messages say different things** — *"recreated; earlier records may have been lost"* versus *"could not be recovered; subsequent records may be lost"*. Past tense against future tense, and an operator needs exactly that difference to know whether the log they are holding is trustworthy from here on.

`tests/test_v07126.py` 185 passed / 2 skipped locally. CI had not run because your PRs sat in `action_required` — first contribution here — which I have now approved; that stops after the first merge.

## One process note, not about this PR

`main` is currently one commit behind an unreleased local slot (v0.73.2) that is paused at its commit gate. I am holding merges — this one, #399, and a fifth PR from another contributor — until that tags, so none of them rides a release whose notes describe something else. #343 is slotted for v0.73.6 in the roadmap, so it gets its own line rather than a footnote.

Approved on content. Nothing needed from you.

MakazhanAlpamys:
Both approved and CI-green, and the hold is over — **v0.73.2 tagged and published** an hour ago, and #393 has merged behind it.

That release is also what put a conflict on this branch: it rewrote `CHANGELOG.md` substantially. **`CHANGELOG.md` is the only conflicting file** — I checked both of your PRs, and neither has a code conflict.

```bash
# keep both entries; your `[Unreleased]` block goes above the new `## [0.73.2]` heading
```

Push and I merge on sight — no re-review, the approval stands.

I would have rebased it for you (I offered the same to another contributor this week), but pushing to a fork is gated on my side and I cannot self-approve it. Given the conflict is one text block and you have been turning things around in minutes, asking is faster than waiting for that.

Worth saying while you wait: this is your first day here and all four PRs arrived with a root cause, a regression test and a named failure mode. The two that collided with in-flight work did so because **we** left the issues unclaimed — that is still being resolved and it is not a reflection on the work.

PR #399, opened 2026-08-15
fix(draft): derive distill epochs from the effective step budget (#364)

### LINKED ISSUE #364: soup draft distill --steps N delivers N/4.44 optimizer steps

ousamabenyounes:
## Summary

Fix #364

`soup draft distill --steps N` delivered only `~N/4.44` optimiser steps. The epoch
count that realises `--steps` divided the request by `rows // batch_size`, ignoring
that `val_split` (0.1) removes rows from training and `gradient_accumulation_steps`
(4) micro-batches make one optimiser step. Both divide the budget, so every distill
run trained for roughly a fifth of the requested steps, silently.

- Epochs are now derived from the **effective** steps-per-epoch:
  `rows * (1 - val_split) / (batch_size * gradient_accumulation_steps)`, then
  `epochs = ceil(requested_steps / steps_per_epoch)`.
- The emitted distill config now **pins** the run shape it assumed (`val_split`,
  `gradient_accumulation_steps`) so the arithmetic and the trainer cannot drift.
- There is no `max_steps` knob in the trainer (per `shrink.py::_build_heal_config_yaml`),
  so `--steps` is realised through the epoch count; rounding up meets or overshoots
  the request by less than one epoch.
- The pre-flight prints the resolved step count, so a mismatch is visible before the run.

New tests in `tests/test_v07133.py::TestDistillStepBudget` (the module that already
covers `soup draft distill`).

RED — on unmodified `main`, `_build_distill_config_yaml(steps=100, data_rows=100)`
emits `epochs: 1`, which delivers `floor(100*0.9)//(1*4) = 22` optimiser steps:

```
    assert delivered >= requested
E   assert 22 >= 100
tests/test_v07133.py:1902: AssertionError
1 failed
```

GREEN — with the fix, the same request emits `epochs: 5` (≈110 delivered), and the
suite passes:

```
tests/test_v07133.py::TestDistillStepBudget .......                       [100%]
tests/test_v07133.py::TestDraftDistillCli .....................
36 passed
```

The parametrised arm varies `val_split` and `gradient_accumulation_steps`
independently (one arm at defaults is not discriminating). `ruff check src/soup_cli/ tests/`
passes clean on both base and the fix.

## Files changed

| File | Change |
|------|--------|
| `src/soup_cli/commands/draft.py` | effective-step epoch derivation + pinned run shape + pre-flight print |
| `tests/test_v07133.py` | `TestDistillStepBudget` — delivered-steps RED→GREEN + independent shape variation |
| `CHANGELOG.md` | `### Fixed` entry (#364) |

## Full local suite

The change is torch-free (pure epoch arithmetic in `commands/draft.py`), so the two
CI jobs that can catch it were replayed locally on both the `main` baseline and the
fix commit — iso-or-better, no new failures:

- `ruff check src/soup_cli/ tests/` — passes clean (blocking CI job).
- `pytest tests/test_v07133.py` distill/config classes (`TestDistillStepBudget`,
  `TestDraftDistillCli`, `TestDraftCliPlumbing`, `TestDraftNoTopLevelTorch`) — 36 passed.

The full `torch` pytest matrix (3.10/3.11/3.12) and the 77% coverage gate run in CI;
this change adds tests and touches no torch code path, so it cannot lower either.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,18 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **`soup draft distill --steps N` now delivers ~N optimiser steps instead of
+  N/4.44 (#364).** The epoch count that realised `--steps` divided the request
+  by `rows // batch_size`, ignoring that `val_split` (0.1) removes rows from
+  training and `gradient_accumulation_steps` (4) micro-batches make one
+  optimiser step — both divide the budget, so every distill run trained for
+  roughly a fifth of the requested steps, silently. Epochs are now derived from
+  the effective steps-per-epoch, the emitted config pins the run shape
+  (`val_split`, `gradient_accumulation_steps`) so the arithmetic and the trainer
+  cannot drift, and the pre-flight prints the resolved step count.
+
+### Fixed
+
 - **`MitigationLogWriter` silently dropped every record once its parent directory
   vanished mid-run (#343).** `record()` reopens the log per call and swallowed the
   `OSError` from `open("ab")` with a bare `return`, so when a shared temp root was
--- a/src/soup_cli/commands/draft.py
+++ b/src/soup_cli/commands/draft.py
@@ -64,6 +64,14 @@ _DISTILL_TIMEOUT_SECONDS = 24 * 60 * 60
 _DISTILL_BATCH_SIZE = 1
 _DISTILL_MAX_LENGTH = 1024
 _MAX_DISTILL_EPOCHS = 100
+# #364 — the epoch count that realises ``--steps N`` is derived from the
+# EFFECTIVE optimiser-step budget, which depends on the run shape: ``val_split``
+# removes rows from training and ``gradient_accumulation_steps`` micro-batches
+# make one optimiser step, so both divide the naive ``rows // batch``. Pin that
+# shape here (equal to the schema defaults) AND emit it into the config, so the
+# builder's arithmetic and the trainer's behaviour cannot drift apart.
+_DISTILL_VAL_SPLIT = 0.1   # DataConfig.val_split default
+_DISTILL_GRAD_ACCUM = 4    # TrainingConfig.gradient_accumulation_steps default
 # How much of a failed subprocess's output to surface (mirrors shrink.py).
 _SUBPROCESS_ERROR_TAIL_CHARS = 800
 # The distill trainer writes a LoRA adapter (never dense base weights), so it
@@ -180,6 +188,43 @@ def _load_pair_member(
 # ---------------------------------------------------------------------------
 # distill
 # ---------------------------------------------------------------------------
+def _distill_steps_per_epoch(
+    data_rows: int, *, val_split: float, batch_size: int, grad_accum: int
+) -> int:
+    """Optimiser steps one epoch actually delivers for the distill run shape.
+
+    ``val_split`` removes rows from training and ``grad_accum`` micro-batches
+    make one optimiser step, so both divide the naive ``rows // batch_size``
+    the epoch count used to assume (#364).
+    """
+    train_rows = math.floor(data_rows * (1.0 - val_split))
+    return max(1, train_rows // (batch_size * grad_accum))
+
+
+def _distill_epochs_for_steps(
+    steps: int,
+    data_rows: int,
+    *,
+    val_split: float,
+    batch_size: int,
+    grad_accum: int,
+) -> int:
+    """Epochs whose delivered optimiser steps land nearest to ``steps``.
+
+    There is no ``max_steps`` knob in the trainer (see
+    ``commands/shrink.py::_build_heal_config_yaml``), so ``--steps`` is realised
+    through the epoch count. Rounding up means the request is met or overshot by
+    less than one epoch, never the ~1/4.44 undershoot of the old arithmetic.
+    """
+    per_epoch = _distill_steps_per_epoch(
+        data_rows,
+        val_split=val_split,
+        batch_size=batch_size,
+        grad_accum=grad_accum,
+    )
+    return max(1, math.ceil(steps / per_epoch))
+
+
 def _build_distill_config_yaml(
     *,
     draft_base: str,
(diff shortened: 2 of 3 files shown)


MakazhanAlpamys (approved):
The arithmetic is right and the tests pin it. Ran the mutation rather than reasoning about it — restored the old naive `rows // batch_size`:

```
FAILED tests/test_v07133.py::TestDistillStepBudget::test_steps_are_delivered_not_quartered
1 failed, 115 passed, 1 skipped
```

Exactly one test notices, and it is the one named for the defect. `tests/test_v07133.py` 116 passed / 1 skipped otherwise.

**The part I want to name is emitting the shape into the config rather than only assuming it.** Two constants equal to the schema defaults would have been the obvious fix and would have been correct *today*; writing `val_split` and `gradient_accumulation_steps` into the generated YAML makes the builder's arithmetic and the trainer's behaviour the same fact rather than two facts that agree for now. I checked both against the live schema and they match — but that check is a snapshot, and your emission is what makes it not matter.

`test_delivered_steps_track_request_across_shapes` parametrised over `val_split` × `accum` is the right generalisation: the bug was one specific ratio (~1/4.44), and a test pinned to that ratio would pass for a fix that is wrong at every other shape.

Also correct, and easy to get wrong in the other direction: **rounding up**. `--steps` has no `max_steps` knob to land on exactly, so it is realised through epochs; overshooting by less than one epoch is the right side to miss on, and you said so in the docstring with the reason.

## One process note, not about this PR

`main` is one commit behind an unreleased local slot (v0.73.2) paused at its commit gate. I am holding merges — this, #398, and a fifth PR from another contributor — until that tags, so nothing rides a release whose notes describe something else. #364 sits in the v0.73.4 `soup draft` slot in the roadmap, so it gets its own line.

CI had not run because your PRs were in `action_required` — first contribution here. Approved now; that stops after the first merge.

Approved on content. Nothing needed from you.

MakazhanAlpamys:
Both approved and CI-green, and the hold is over — **v0.73.2 tagged and published** an hour ago, and #393 has merged behind it.

That release is also what put a conflict on this branch: it rewrote `CHANGELOG.md` substantially. **`CHANGELOG.md` is the only conflicting file** — I checked both of your PRs, and neither has a code conflict.

```bash
# keep both entries; your `[Unreleased]` block goes above the new `## [0.73.2]` heading
```

Push and I merge on sight — no re-review, the approval stands.

I would have rebased it for you (I offered the same to another contributor this week), but pushing to a fork is gated on my side and I cannot self-approve it. Given the conflict is one text block and you have been turning things around in minutes, asking is faster than waiting for that.

Worth saying while you wait: this is your first day here and all four PRs arrived with a root cause, a regression test and a named failure mode. The two that collided with in-flight work did so because **we** left the issues unclaimed — that is still being resolved and it is not a reflection on the work.

MakazhanAlpamys:
**#398 is merged** ([`9ad98b9`](https://github.com/MakazhanAlpamys/Soup/commit/9ad98b9)) — your first merge here. Its CHANGELOG entry is what now conflicts with this one, so this needs one more `git fetch origin && git rebase origin/main`. `CHANGELOG.md` only, again; no code conflict.

Cascading CHANGELOG conflicts between two PRs from the same author landing minutes apart are an artifact of how this repo keeps its notes, not something you did — and I merged them in the order that made this one, not the other, pay for it. Sorry for the second round trip.

Approval stands; push and I merge on sight.

ousamabenyounes:
Rebased onto latest `main` — kept the `[Unreleased]` #364 entry above `## [0.73.2]`, alongside #343 which landed with #398. `CHANGELOG.md` only, no code change; `ruff` clean. Ready to merge.

MakazhanAlpamys:
Merged as [`a1b6a34`](https://github.com/MakazhanAlpamys/Soup/commit/a1b6a34) — your second, and #364 closes with it. Both entries are in [`CONTRIBUTORS.md`](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md). Sorry again for the two rebases; both conflicts were `CHANGELOG.md` collateral from things landing around you, not anything in your work.

The detail worth naming is the one I flagged in review and want on the record now that it is in `main`: **emitting the run shape into the generated config instead of only assuming it.** Two constants matching the schema defaults would have been correct today and would have rotted the first time a default moved. Writing `val_split` and `gradient_accumulation_steps` into the YAML makes the builder's arithmetic and the trainer's behaviour the same fact rather than two facts that agree for now.

And parametrising the test over `val_split` × `accum` rather than pinning the ~1/4.44 ratio: a test fixed to that number passes for a fix that is wrong at every other shape, which is exactly how a repaired bug comes back.

Issue #402, opened 2026-08-15
MCP execution: the one-active-execution cap is in-memory, so a server restart can double-book it


MakazhanAlpamys:
Split out of #393 (@CODING-DARSH), raised as important-but-not-blocking, and a sibling of #401.

`ExecutionManager._active_run_id` is instance state. #297 settled that child jobs survive a client disconnect, so a training run outlives the server that started it — and a **restarted** server starts with `_active_run_id = None`, sees free capacity, and will launch a second execution while the first is still training.

The cap is documented as one active execution per MCP server *process*, which is literally true and is not what an operator reads it as. What they read is a guarantee about the machine, and on the machine two trainings can now compete for the same GPU.

## Fix path

Persist the slot and gate it on liveness rather than on process memory: record the active `run_id` where a new server can find it, and treat it as free only when the recorded PID is gone. That is the same liveness check #401 needs for the tracker, so the two are one piece of work and should probably land together.

If persistence is judged not worth it, the honest alternative is to say so in `docs/commands.md` — that the cap is per-process and a restart can double-book — rather than leaving the stronger reading in place.

## Acceptance criteria

- [ ] a restarted server does not launch a second execution while a previous child is still alive
- [ ] a stale record whose PID is gone does NOT permanently block execution — the failure mode of an over-eager fix is a server that refuses everything forever with no error a user can act on
- [ ] whichever way it is decided, `docs/commands.md` states the actual scope of the cap

MakazhanAlpamys:
Closed by #408 — thanks to @ousamabenyounes.

Gating the cap on a **persisted** run whose pid is still alive is the right shape: it survives a restart, which was the actual complaint, while a stale record whose process is gone never wedges execution permanently. Both halves matter and the PR tested both.

## Recorded here for the next person

This and #401 collided on `main` shortly after both merged — see my note in #401 for the full account. Short version: each PR was green against a `main` that did not carry the other, and the interaction turned this PR's own `_pid_is_alive` guard into dead code, because #401's reconcile-on-read rewrites the dead row before this code inspects `status`.

Measured: deleting that guard passed all 41 MCP tests. It is now pinned by a test that neutralises reconcile so this cap's own check is the only thing that can reject a dead pid, plus a control proving the check still accepts a live one.

Repaired in `7bc3f83`; nothing needed from you.

PR #407, opened 2026-08-15
fix(mcp): reconcile orphaned 'running' runs on read (#401)

### LINKED ISSUE #401: MCP execution: the watcher is a daemon thread, so a server exit leaves the run 'running' forever
Split out of #393 (@CODING-DARSH), where it was raised as important-but-not-blocking.

`ExecutionManager._watch` runs as a `daemon=True` thread. When the MCP server process exits, daemon threads are killed without unwinding, so `finish_execution()` never runs and `ExperimentTracker` keeps the run at `running` — forever, with no watcher left to correct it.

**This is worst in exactly the case the design chose to support.** #297 settled that child jobs should SURVIVE a client disconnect, because a training run is hours and an MCP client disconnects for routine reasons (the editor restarts, the laptop sleeps). That decision is right, and it is only safe because the run is discoverable afterwards through the tracker — which is the thing this defect makes unreliable.

The concrete cost is not a wrong row in a database. It is that `soup runs` showing a permanently-`running` job trains an operator to ignore the status field, and the status field is the only thing standing between them and a second concurrent training run.

## Fix path

Reconcile on read rather than trying to make the watcher survive: when the tracker reports a run as `running`, check whether the recorded PID is still alive and, if not, mark it terminated with an unknown exit code. That is a smaller change than process supervision and it also solves #(the capacity-slot issue filed alongside this), since the same liveness check is what a restarted server needs.

## Acceptance criteria

- [ ] a run whose process is gone is not reported as `running` by `soup runs` / `runs show`
- [ ] the distinction between 'completed with exit code N' and 'the watcher never saw it finish' survives into the record — an unknown outcome must not be written as success
- [ ] a test that kills the watcher without letting it unwind, then asserts the reconciliation, with a CONTROL that a genuinely-running process is still reported as running

ousamabenyounes:
## Summary

Fix #401

`ExecutionManager._watch` runs as a `daemon=True` thread. When the MCP server
process exits, the watcher is killed without unwinding, so `finish_execution`
never runs and `ExperimentTracker` keeps the run at `running` forever — with no
watcher left to correct it. A permanently-`running` row in `soup runs` trains
the operator to ignore the status field, which is the only thing guarding
against a second concurrent training run.

Per the issue's suggested path, this **reconciles on read** rather than trying
to keep the watcher alive:

- `ExperimentTracker.list_runs` / `get_run` now check a `running` row's recorded
  pid. If the process is gone, the row is rewritten to `terminated` with an
  unknown (`None`) exit code via `finish_execution`, whose existing guard
  (`status NOT IN ('completed','failed')`) keeps a richer terminal status intact
  and makes the rewrite idempotent.
- An **unknown outcome is never written as success** — `terminated` + `exit_code
  = None` is distinct from `completed`.
- Only MCP-spawned runs carry a pid; a run without one is left untouched (its
  liveness cannot be checked), so direct training runs are unaffected.
- `soup runs` / `runs show` no longer hardcode `running` in the status
  fallback — they render the real status, so a reconciled row shows
  `terminated`, not `running`.

The same liveness check is what a restarted server needs for the sibling
capacity-slot issue.

## Acceptance criteria

- [x] a run whose process is gone is not reported as `running` by `soup runs` / `runs show`
- [x] `completed with exit code N` stays distinct from `the watcher never saw it finish` — an unknown outcome is written as `terminated` / `exit_code=None`, never success
- [x] a test kills the watcher (real process, exited, pid dead) without unwinding and asserts the reconciliation, with a CONTROL that a genuinely-running process (this test's own pid) is still reported `running`

New test: `tests/test_issue401_orphaned_run_reconcile.py`.

RED — on unmodified `main`, a dead-process run is still reported `running`:

```
RED: dead-pid run status = running exit_code = None
AssertionError: BUG #401 present: dead-process run still 'running'
```

GREEN — with the fix:

```
tests/test_issue401_orphaned_run_reconcile.py ....                      [100%]
4 passed in 0.24s
```

## No regression

```
tests/test_tracker.py tests/test_runs.py tests/test_mcp_execute.py tests/test_issue401_orphaned_run_reconcile.py
63 passed in 4.06s
```
`ruff check` clean on the changed files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,22 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **A run whose watcher died was reported `running` forever (#401).**
+  `ExecutionManager._watch` runs as a `daemon=True` thread, so when the MCP
+  server process exits it is killed without unwinding and `finish_execution`
+  never runs — `ExperimentTracker` kept the run at `running` with no watcher
+  left to correct it, training the operator to ignore the one status field that
+  guards against a second concurrent run. The tracker now reconciles on read:
+  when a `running` row carries a pid whose process is gone, `list_runs` /
+  `get_run` rewrite it to `terminated` with an unknown (`None`) exit code — an
+  unknown outcome is never recorded as success, and the richer
+  `completed`/`failed` terminal statuses are left untouched. `soup runs` no
+  longer hardcodes `running` for a non-running row.
+
+### Fixed
+
+### Fixed
+
 - **`kl_control` rewrote the trainer's β/kl_coef on every step, including a `hold`, so a
   non-acting run was numerically identical to `log_only` (#371).** `_run_bang_bang` called
   `_apply_coefficient` unconditionally; on a `hold` the controller writes back the value
--- a/src/soup_cli/commands/runs.py
+++ b/src/soup_cli/commands/runs.py
@@ -95,7 +95,10 @@ def list_runs(
         elif status == "failed":
             status_str = "[red]failed[/]"
         else:
-            status_str = "[yellow]running[/]"
+            # Show the real status (e.g. 'terminated' reconciled from a dead
+            # watcher, issue #401) — never hardcode 'running' for a row that is
+            # no longer running.
+            status_str = f"[yellow]{status}[/]"
 
         # Format loss
         loss_str = ""
--- a/src/soup_cli/experiment/tracker.py
+++ b/src/soup_cli/experiment/tracker.py
@@ -10,6 +10,7 @@ Usage:
 from __future__ import annotations
 
 import json
+import os
 import secrets
 import sqlite3
 from datetime import datetime
@@ -17,6 +18,42 @@ from pathlib import Path
 from typing import Optional
 
 from soup_cli.utils.constants import EXPERIMENTS_DB, SOUP_DIR
+
+# Run status values this module reconciles. A watcher that never unwound (its
+# daemon thread was killed when the MCP server exited) leaves the run at
+# _STATUS_RUNNING forever. Reconcile-on-read rewrites such a row to
+# _STATUS_TERMINATED with an unknown (None) exit code so a lost outcome is never
+# mistaken for success. See issue #401.
+_STATUS_RUNNING = "running"
+_STATUS_TERMINATED = "terminated"
+
+
+def _process_is_alive(pid: int) -> bool:
+    """Return True if a process with this PID currently exists.
+
+    Signal 0 runs the OS existence/permission check without delivering a
+    signal. A PID owned by another user raises PermissionError and still counts
+    as alive; only a missing process (ProcessLookupError / ESRCH) is dead.
+
+    The ``except OSError: return False`` branch is not an optional tidy-up: on
+    Windows a dead PID surfaces as ``OSError`` (winerror 87), never
+    ProcessLookupError, so on Windows that branch is the *primary* path every
+    reconciliation arrives through — do not narrow it away.
+
+    Known limitation — PID reuse: a dead run whose PID has since been recycled
+    by an unrelated process reads as alive forever, so reconciliation never
+    fires for it. That is inherent to PID-liveness (a start-time comparison
+    would be needed to close it) and is out of scope for issue #401.
+    """
+    try:
+        os.kill(pid, 0)
+    except ProcessLookupError:
+        return False
+    except PermissionError:
+        return True
+    except OSError:
+        return False
+    return True
 
 _SCHEMA_SQL = """
 CREATE TABLE IF NOT EXISTS runs (
@@ -331,13 +368,35 @@ class ExperimentTracker:
         conn.execute("UPDATE runs SET status = 'failed' WHERE run_id = ?", (run_id,))
         conn.commit()
 
+    def _reconcile_orphaned_run(self, run: dict) -> dict:
+        """Rewrite a stale 'running' row whose process is gone (issue #401).
+
+        Only MCP-spawned runs carry a pid; a run recorded without one is left
+        untouched because its liveness cannot be checked here. A dead pid is
+        persisted as _STATUS_TERMINATED with exit_code None (unknown) through
+        finish_execution, whose guard keeps a richer 'completed'/'failed'
+        terminal status intact and makes the rewrite idempotent.
+        """
+        pid = run.get("pid")
+        if (
+            run.get("status") == _STATUS_RUNNING
+            and pid is not None
+            and not _process_is_alive(pid)
+        ):
+            self.finish_execution(
+                run["run_id"], status=_STATUS_TERMINATED, exit_code=None
+            )
+            run["status"] = _STATUS_TERMINATED
+            run["exit_code"] = None
+        return run
+
     def list_runs(self, limit: int = 50) -> list[dict]:
         """Return list of runs ordered by created_at desc."""
         conn = self._get_conn()
         rows = conn.execute(
             "SELECT * FROM runs ORDER BY created_at DESC, rowid DESC LIMIT ?", (limit,)
         ).fetchall()
-        return [dict(row) for row in rows]
+        return [self._reconcile_orphaned_run(dict(row)) for row in rows]
 
     def get_run(self, run_id: str) -> Optional[dict]:
         """Get full details of a single run. Supports prefix matching."""
@@ -365,7 +424,7 @@ class ExperimentTracker:
             elif len(rows) > 1:
                 return None  # ambiguous prefix
 
-        return dict(row) if row else None
+        return self._reconcile_orphaned_run(dict(row)) if row else None
 
     def get_metrics(self, run_id: str) -> list[dict]:
         """Get all metric rows for a run, ordered by step."""
--- a/src/soup_cli/experiment/tracker.py
+++ b/src/soup_cli/experiment/tracker.py
@@ -13,6 +13,7 @@ import json
 import os
 import secrets
 import sqlite3
+import sys
 from datetime import datetime
 from pathlib import Path
 from typing import Optional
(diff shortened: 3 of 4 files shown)


MakazhanAlpamys (approved):
Taken within the hour of the issue being filed, and it does the thing the issue asked for rather than the thing that looks like it: reconcile on read, not keep the watcher alive.

`tests/test_issue401_orphaned_run_reconcile.py` 4 passed locally.

## The line that looks defensive is load-bearing, and I measured it

```python
except OSError:
    return False
```

The docstring frames this as a fallback so *"a corrupt record reconciles instead of sticking on 'running' forever"*. On **Windows it is the primary path**, not the fallback. `os.kill` there is not the POSIX signal-0 existence check — it goes through `OpenProcess`, and a dead PID raises `OSError` with `winerror=87`, never `ProcessLookupError`:

```
never-existed pid            -> OSError winerror=87
exited child, handle dropped -> OSError winerror=87
orphaned grandchild          -> OSError winerror=87
```

So on the maintainer's own platform, every reconciliation arrives through that branch. I confirmed it is protected — mutating it to `return True` fails `test_dead_watcher_run_is_reconciled_not_reported_running` — but the comment above it invites exactly the "tidy-up" that would silently disable the fix on Windows while every POSIX test stayed green. **Please say so in the docstring**: that on Windows a dead PID surfaces as `OSError`, so this branch is not optional.

Two hypotheses I tested and discarded, recorded because they are the reason I looked:

- *"`os.kill(pid, 0)` terminates the process on Windows"* — CPython's Windows `os.kill` routes non-console signals to `TerminateProcess`, so this looked likely. **It does not**: a spawned child survives it, twice measured. Withdrawn.
- *"the fix never fires on Windows"* — a just-exited child **does** read as alive while its spawner still holds an open handle. But that is not the #401 shape: there the MCP server is *gone*, nothing holds a handle, and the PID then reports `OSError` correctly. Withdrawn too.

Both were wrong, and the controls are what killed them. I mention it only so the docstring note above does not read as speculation.

## One known limitation worth naming rather than fixing

**PID reuse.** A dead run whose PID has been recycled by an unrelated process reads as alive forever, and reconciliation never fires for it. That is inherent to PID-liveness and out of scope here — the fix is a recorded process start-time compared against the run's own — but it belongs in the docstring or `docs/`, because a reader will otherwise take "reconciled on read" as unconditional.

Right calls elsewhere: `PermissionError` → alive (a PID owned by another user exists); only MCP-spawned runs carry a pid, and rows without one are left untouched rather than guessed at; and routing the rewrite through `finish_execution`, whose `status NOT IN ('completed','failed')` guard keeps a richer terminal status intact and makes the rewrite idempotent — that last one is what stops a reconciliation from overwriting a real exit code on a second read.

---

Approving. The docstring note is the only thing I would like added, and it is one sentence — push it and I merge, or say you would rather I add it on merge and I will.

Fourth PR, three merged or approved, and this one was filed as an issue an hour before you took it. Thank you.

ousamabenyounes:
Added both docstring notes on `_process_is_alive` and rebased onto latest `main` (CHANGELOG only):

- the `except OSError: return False` branch is documented as **not optional** — on Windows a dead PID surfaces as `OSError` (winerror 87), never `ProcessLookupError`, so that branch is the primary path every reconciliation arrives through there.
- the PID-reuse limitation is named: a dead run whose PID has been recycled reads as alive forever, so reconciliation never fires for it — inherent to PID-liveness, out of scope here.

Test still green, `ruff` clean. Thank you for measuring the Windows path and the controls that killed the two hypotheses — the start-time comparison is the right way to close PID-reuse if it ever becomes worth it.

MakazhanAlpamys (changes requested):
Stop — do not merge this yet. **CI is red on all three Windows cells**, and the cause is not the docstring or the test. `_process_is_alive` does not check liveness on Windows at all.

## `os.kill(pid, 0)` on Windows is not an existence check — it fires Ctrl+C

```
>>> signal.CTRL_C_EVENT
0
```

CPython's Windows `os.kill` branches on the signal *before* doing anything else: if `sig` is `CTRL_C_EVENT` or `CTRL_BREAK_EVENT` it calls `GenerateConsoleCtrlEvent(sig, pid)` and returns. **`CTRL_C_EVENT` is `0`.** So `os.kill(pid, 0)` on Windows sends a console Ctrl+C to a process *group* — it never reaches the `OpenProcess` path, never raises `ProcessLookupError`, and tells you nothing about whether the PID exists.

The POSIX contract the docstring describes ("signal 0 runs the OS existence/permission check without delivering a signal") is true on POSIX and false on Windows, in the most inconvenient possible way: the value chosen to mean *"do not deliver a signal"* is the value Windows uses to mean *"deliver Ctrl+C"*.

## The CI failure matches that mechanism exactly

```
windows-latest 3.10  FAIL   2348 passed, 41 skipped ... KeyboardInterrupt (subprocess.py:1506)
windows-latest 3.11  FAIL   2348 passed, 41 skipped ... KeyboardInterrupt
windows-latest 3.12  FAIL   2348 passed, 41 skipped ... KeyboardInterrupt (subprocess.py:1590)
ubuntu × 3, macos × 3, lint, type-check   all PASS
```

Three different Python versions stopping at **exactly 2348 passed** is not a flake or a timeout — it is deterministic. And the interrupt is a `KeyboardInterrupt` raised *inside the test runner*, in `subprocess.wait()`, which is what a console Ctrl+C delivered to the runner's own process group looks like. Your code sent it.

I could not reproduce the interrupt in a local Git Bash shell — `os.kill(pid, 0)` returned normally there and the child survived — but that is a console-attachment difference, not a contradiction: Git Bash is not a Windows console session that receives the event. The CI cells are, three times over, at the same instruction. **The nine-cell matrix is the instrument here, and it is unambiguous.**

## What this means for the fix

The reconciliation is right and the design is right. The liveness primitive is not portable, and neither is the correct-dead path I praised in the earlier review — my `winerror=87` measurements came from a *different* call shape and I now think they were reached by luck of the shell, not by the code path CI takes. I am withdrawing that half of my previous review; it pointed you at a docstring when the function underneath was wrong.

A portable check needs a real Windows branch, roughly:

```python
if sys.platform == "win32":
    # OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION) + GetExitCodeProcess,
    # alive iff STILL_ACTIVE (259). A failed open with ERROR_INVALID_PARAMETER
    # means the pid does not exist; ERROR_ACCESS_DENIED means it does.
else:
    os.kill(pid, 0)   # POSIX: genuine existence check
```

Whatever shape you take, the acceptance test is the one CI just ran: **the full matrix green, Windows included.** A targeted suite cannot see this — the local `tests/test_issue401_orphaned_run_reconcile.py` was 4/4 green on my Windows box while the real defect was live.

## Two things worth saying

**This is a good catch by your own PR, not against it.** Nothing in the repo checks process liveness today, so nobody had hit this. It surfaced because you wrote the first code that does — and it surfaced in CI, before merge, which is where it should.

**And it strengthens the docstring note rather than replacing it.** The point stands that a POSIX-shaped assumption about signals silently means something else on Windows; it just turns out to bite one line earlier than either of us thought.

Sorry for approving before the matrix finished — that was mine, and it is exactly the case the "green locally ≠ green on Linux/Windows" rule in this repo exists for. Re-review as soon as you push; #408 is unaffected in design but shares the primitive, so it will want the same fix.

ousamabenyounes:
You're right and I was wrong to accept the winerror-87 story — thank you for re-measuring. `os.kill(pid, 0)` on Windows is `CTRL_C_EVENT` (0), so it routes to `GenerateConsoleCtrlEvent` and fires a console Ctrl+C at the group instead of checking existence, which is exactly the `KeyboardInterrupt` the three Windows cells hit.

Fixed with a real Windows branch: `_windows_process_is_alive` opens a `PROCESS_QUERY_LIMITED_INFORMATION` handle and reads `GetExitCodeProcess` — alive iff `STILL_ACTIVE` (259); a failed open is dead unless it's only `ERROR_ACCESS_DENIED` (exists, no rights). `restype`/`argtypes` are set so the HANDLE isn't truncated on Win64. POSIX keeps signal 0. Rebased onto `main` (CHANGELOG). POSIX tests green locally; the Windows branch will prove itself on your CI cells.

MakazhanAlpamys:
Merged as [`2de76e0`](https://github.com/MakazhanAlpamys/Soup/commit/2de76e0), and #408 behind it as [`986a482`](https://github.com/MakazhanAlpamys/Soup/commit/986a482). #401 and #402 both closed. Entries in [`CONTRIBUTORS.md`](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md).

**The Windows fix is verified, not assumed.** `_process_is_alive` dispatches to the `OpenProcess` + `GetExitCodeProcess` branch *before* the POSIX path, so `os.kill` is unreachable on win32 — confirmed by monkeypatching `os.kill` to raise and watching it never fire. Probed both directions with the project's own function: `os.getpid()` → alive, `999999` → dead, and a real child → alive while running, dead after it exits. That last one is the discriminating probe, because `OpenProcess` still succeeds on a just-exited child; a naive `if handle: return True` would answer alive. Yours answers dead.

CI: three Windows cells at **17603 passed** against the previous run's 2348-then-`KeyboardInterrupt`.

**One residual, filed as #424 rather than held against this.** `STILL_ACTIVE` is 259, which is also a legal exit code, so a process exiting *with* 259 reads as alive forever. Measured, with a control (exit 7 → correctly dead). In #401 that means one run never reconciles; in **#402 it means the capacity gate is never released and the server refuses every subsequent execution** — the failure your own PR body called out as the one an over-eager fix must not cause. Low probability, unbounded consequence. `WaitForSingleObject(handle, 0)` disambiguates it.

#424 also carries two things worth doing while you are in there: the ~50 lines are duplicated verbatim across `tracker.py` and `execution.py` and their docstrings have **already** diverged, and no test names the platform split — so a revert to `os.kill(pid, 0)` would be caught only by the whole Windows matrix dying at 2348 with the same confusing signature as last time.

Seven merges today. The one I would single out is not a fix at all — it is that when the Windows matrix went red you did not paper over it, you found the actual mechanism and wrote a real platform branch. Plenty of people would have added a skip.

PR #408, opened 2026-08-15
fix(mcp): gate the execution cap on a live persisted run (#402)

### LINKED ISSUE #402: MCP execution: the one-active-execution cap is in-memory, so a server restart can double-book it

ousamabenyounes:
## Summary

Fix #402

`ExecutionManager._active_run_id` is instance state. Because a child job
survives a client disconnect (#297), a training run outlives the server that
started it — and a **restarted** server starts with `_active_run_id = None`,
sees free capacity, and would launch a second execution while the first is still
training on the same GPU. The documented "one active execution per server
*process*" reads to an operator as a guarantee about the machine.

Per the issue's suggested path, the slot is now gated on **liveness of a
persisted run** rather than on process memory:

- `ExecutionManager._live_persisted_run()` asks the tracker for a `running` run
  whose recorded pid is still alive. `execute()` refuses when one exists, so a
  restart cannot double-book.
- A **stale record whose process is gone does not block** — only a live pid
  counts, so a dead run never wedges the gate shut (the failure mode of an
  over-eager fix).
- The pid liveness check is the same one #401 needs for the tracker.
- `docs/commands.md` now states the actual scope of the cap.

Tests set `SOUP_DB_PATH` to a per-test temp DB via an autouse fixture, so the
new gate never reads `running` rows leaked into the real `~/.soup/experiments.db`
by other tests.

## Acceptance criteria

- [x] a restarted server does not launch a second execution while a previous child is still alive
- [x] a stale record whose PID is gone does NOT permanently block execution
- [x] `docs/commands.md` states the actual scope of the cap

New test: `tests/test_issue402_restart_capacity.py`.

RED — on unmodified `main`, the new tests fail (no persistent gate; the helper does not exist):

```
2 failed, 1 passed
```
(the one that passes on base is the control: a dead prior record must not block, which is already true without the gate.)

GREEN — with the fix:

```
tests/test_issue402_restart_capacity.py ...                             [100%]
3 passed
```

## No regression

```
tests/test_mcp_execute.py tests/test_tracker.py tests/test_runs.py
tests/test_run_cost.py tests/test_replay.py tests/test_ui_metrics.py
tests/test_issue402_restart_capacity.py
101 passed, 13 skipped
```
`ruff check` clean on the changed files.


MakazhanAlpamys:
Reviewing the content now rather than waiting on the rebase, since #407 will land first and put a `CHANGELOG.md` conflict here anyway — that is the only conflicting file, no code conflict.

The gate itself is right, and it satisfies both halves of #402 including the one that is easy to miss: **a stale record whose process is gone does not block.** An over-eager version of this fix wedges the server shut forever with no error a user can act on, and that failure is worse than the double-booking it prevents. You wrote the guard and the test for it.

## The `autouse` fixture needs to be its own decision, not a rider

```python
@pytest.fixture(autouse=True)
def _isolate_experiments_db(tmp_path, monkeypatch):
    monkeypatch.setenv("SOUP_DB_PATH", str(tmp_path / "experiments.db"))
```

I checked the mechanics and they are correct — `SOUP_DB_PATH` is exactly what `ExperimentTracker` reads (`tracker.py:105`), and a test that needs its own DB can still override it afterwards.

But `autouse=True` in the shared `tests/conftest.py` changes the environment for **every test in the suite — around 18,000 of them across nine CI cells**, inside a PR whose subject is one MCP capacity gate. Two things follow:

1. **On the merits I think you are right.** Reading the developer's real `~/.soup/experiments.db` during a test run is a genuine isolation bug, and it will bite something other than this feature eventually. It is a good change.
2. **It cannot ride along quietly.** A suite-wide default belongs in its own commit with its own line in the CHANGELOG, so that if something goes strange in three weeks the bisect lands on "we isolated the experiments DB" rather than on "MCP capacity gate". Split it into a separate commit on this branch — same PR is fine — and give it a sentence of its own.

And the only evidence that it is safe is a **full green matrix**, not the targeted suites. Your CI has not run yet here; I will not merge this one on local runs.

## Rebase

`git fetch origin && git rebase origin/main` once #407 is in — `CHANGELOG.md` only, and the entries want to sit side by side. I will tell you when #407 lands so you only do it once.

Nothing else. Fifth PR today, and this one plus #407 close the two follow-ups that came out of somebody else's feature review an hour after they were filed.

ousamabenyounes:
Applied the same portable liveness primitive you flagged on #407: the capacity gate's `_pid_is_alive` now uses OpenProcess + GetExitCodeProcess on Windows instead of `os.kill(pid, 0)` (which fires a console Ctrl+C there). Rebased onto `main`. POSIX tests green; the Windows cells will confirm the primitive.

Issue #371, opened 2026-08-10
Reward-hack controller: pid_lagrangian has never had a valid run, the rollback ladder never fires, and a non-acting kl_control run is numerically identical to log_only


MakazhanAlpamys:
## Three facts about one shipped feature, from the H100 session

These are separate from #286 (which asks whether the controller *works*). These say that parts of it have never been exercised **mechanically**, which is a prerequisite for asking about efficacy at all.

### 1. `pid_lagrangian` has no valid arm at all

All three attempts crashed, at steps **6, 25 and 8** (`benchmarks/gate-h100-validation.md` L3834–3836). That is a shipped `--reward-hack-mitigation` mode with **zero** successful runs — unexercised mechanically as well as statistically. (The crashes are #342's `grad_norm nan`, so this is partly blocked on that, but the mode has never been shown to complete a run even once.)

### 2. The rollback ladder never fired in any arm

The escalation ladder requires **three** consecutive HACK votes; the beta ladder escalated on **four** consecutive HACK votes in the same run (0.03 → 0.045 → 0.0675 → 0.1013 → 0.1519, each commanded value read back off the live trainer). The record states plainly: "this record does not explain why one path reached its threshold and the other did not. **Recorded as an open inconsistency.**" (L3829–3833)

Either the two thresholds are counted against different signals, or one of the counters resets where the other does not. Both are bugs; which one it is has not been established.

### 3. A `kl_control` run that never acts is numerically identical to `log_only`

`_run_bang_bang` calls `_apply_coefficient` on **every** step, including a `hold`. With `grpo_beta` and `reward_hack_beta_floor` both at the default `0.02`, that write puts back the value already there (L4602–4607).

This matters because **10 of 11** `kl_control` arms in the replication were in exactly that state — so any measured difference between the two modes in that pool is "a difference between two runs the code says are the same computation" (L4600–4612). It makes the attrition asymmetry (Fisher p=0.0275 unbalanced, p=0.0995 balanced) uninterpretable rather than merely unresolved.

## Fix path

1. Give `pid_lagrangian` one completed run at a configuration that does not trip #342 (lower lr, shorter horizon) before any efficacy claim is attempted.
2. Reconcile the two thresholds: make the rollback ladder and the beta ladder count the same signal with the same reset rule, or document why they differ.
[...]

Issue #345, opened 2026-08-08
measure_acceptance scores the draft against penalised target logits: a draft identical to the target scores 0.86


MakazhanAlpamys:
Found while validating #303 on an H100.

`measure_acceptance` calls `generate(do_sample=False)`, which disables **sampling** but not the checkpoint's `repetition_penalty` logits processor. The target then emits `argmax(penalised logits)` while the draft is scored on `argmax(raw logits)` — so the two are compared under different rules.

**A draft identical to the target does not score 1.0:**

| self-acceptance (draft == target) | declared rep-penalty | shipped path | penalty neutralised |
|---|---|---|---|
| Qwen2.5-0.5B | 1.1 | **0.8623** | **0.9892** |
| Llama-3.2-1B | 1.0 | 0.9945 | 0.9945 |
| Llama-3.1-8B | 1.0 | 1.000000 | 1.000000 |

**Numerical precision was the obvious hypothesis and it was tested and REJECTED:** fp32 gives 0.8638 against bf16's 0.8623, and the median top-2 logit gap at disagreement is 0.64–0.75 — decisive margins, not near-ties.

**This does not exist in real speculative decoding.** Confirmed from source that assisted generation passes the *same* `logits_processor` to the draft (`candidate_generator.py:314`). So the defect is in the measurement, not in the feature it measures — which is worse in one specific way: it misreports the thing users gate on.

**Practical impact:** on Qwen2.5-32B it **flips the verdict band** — 0.6857 MODERATE becomes 0.7072 STRONG once the penalty is neutralised. A CI gate at `--min-acceptance 0.70` would fail that pair for an artefact.

Models that declare no `repetition_penalty` are unaffected, so previously published SmolLM2 figures stand.

### Also found in the same subsystem
- **`--steps N` delivers N/4.44 steps.** `_build_distill_config_yaml` computes `epochs = ceil(steps / data_rows)` ignoring `val_split=0.1` and `gradient_accumulation_steps=4`; `--steps 2000` over 400 rows ran **450** optimizer steps. Epochs over the data are as intended; the user-facing knob is not.
- **The teacher loads in fp32** (`distill.py` passes no dtype) — 32 GB for an 8B teacher, ~131 GB for the 32B one #303 proposes. Same root cause as #339.
- `soup draft distill` captures its `soup train` subprocess output entirely, so there is no live progress or loss curve.

### Acceptance
- [ ] the draft is scored under the same logits processors the target used
[...]

Issue #321, opened 2026-07-28
training.bnb_4bit_use_double_quant is validated but never read (all 4-bit paths hardcode True)


MakazhanAlpamys:
`training.bnb_4bit_use_double_quant` is a documented, validated schema field that **nothing reads**. Every 4-bit load in Soup hardcodes double-quantisation to `True`:

- `src/soup_cli/utils/quant_menu.py:351` — `"bnb_4bit_use_double_quant": True` in `build_quantization_config_for_loader`, which is the single entry point every transformers-backend trainer uses.
- `src/soup_cli/utils/save_formats.py:303` — same, for the 4-bit save path.
- `src/soup_cli/utils/layer_stream_runtime.py` — `build_nf4_config(double_quant=True)` for layer streaming (v0.72.2), deliberately matching the above.

So a user can write `bnb_4bit_use_double_quant: false`, have it pass validation (there is even a cross-validator, `_validate_bnb_4bit_double_quant`, that checks it is only set alongside `quantization: 4bit`), and get double-quantisation anyway. Nothing warns.

This is the "config accepted, behaviour diverges silently" pattern the project refuses everywhere else.

Surfaced during the v0.72.2 code review. It was **not** fixed there on purpose: v0.72.2's central guarantee is that a streamed NF4 run is bit-exact against a *resident* NF4 run, and that only holds while both sides quantise with identical settings. Honouring the flag on the streamed side alone would have silently broken that. The fix has to be repo-wide or not at all.

### Proposed fix

Thread `tcfg.bnb_4bit_use_double_quant` through `build_quantization_config_for_loader`, `save_formats`, and `layer_stream_runtime.build_nf4_config` together, in one change, so the resident and streamed paths stay in agreement.

Note the default is currently `False` on the schema field while every call site uses `True`, so a naive threading would **change default behaviour for every existing 4-bit user** (larger memory footprint, different quantised bytes). Either flip the schema default to `True` to match what has actually been shipping, or treat it as a breaking change and say so in the CHANGELOG.

The alternative — reject the field with a footgun error saying it is not configurable — is cheaper but worse, since double-quant is a legitimate knob.

### Acceptance criteria

- [ ] `bnb_4bit_use_double_quant: false` measurably produces a `BitsAndBytesConfig` with `bnb_4bit_use_double_quant=False` on the resident path
- [ ] ...and on the layer-streaming path, with the shard cache invalidating on the change (it already keys on `double_quant`)
[...]

PR #409, opened 2026-08-15
fix(draft): gate measure on config.vocab_size and write its report incrementally (#344)

### LINKED ISSUE #344: soup draft: distill and measure disagree on 'shares a tokenizer', and measure discards all its work when they do
Measured on an H100 while validating #303.

The two `soup draft` subcommands use **different definitions of the same precondition**, and they disagree on every Qwen2.5 large←small pair — which is exactly the pair #303 proposes.

Qwen2.5 large models declare `config.vocab_size` **152064**, the small ones **151936** (identical tokenizers, padded embedding rows).

| | gate | on Qwen2.5-32B ← 0.5B |
|---|---|---|
| `soup draft distill` | `AutoConfig.vocab_size` | **refuses up front**, exit 1, clear message, no weights loaded — correct |
| `soup draft measure` | `same_tokenizer()` (tokenizer vocab 151643 both + probe ids) | **accepts**, computes acceptance and plain throughput, then dies inside transformers |

```
generation/utils.py:1543
ValueError: The main and assistant models have different tokenizers
```

**The second half is the worse half.** `measure` writes its report only after the assisted arm, so when transformers refuses, **every completed measurement is discarded and no JSON is produced** — the acceptance rate and plain throughput it already computed are thrown away.

### Fix shape
Pick one definition and use it in both places. transformers gates on `config.vocab_size`, so that is the one that decides whether the run can proceed; `same_tokenizer` can stay as an additional check but must not be the only one `measure` consults. Separately, write the report incrementally so a failure in the assisted arm does not destroy the acceptance measurement that already succeeded.

### Acceptance
- [ ] `measure` and `distill` accept and refuse the same pairs
- [ ] a pair transformers will refuse is refused BEFORE any model is loaded
- [ ] a failure in the assisted arm still leaves the acceptance and plain-throughput numbers on disk
- [ ] a test on a vocab-size-mismatched pair with matching tokenizers (the Qwen2.5 shape)

Record: `benchmarks/gate-h100-validation.md` STEP 26.

ousamabenyounes:
## What does this PR do?

Fix #344 — `soup draft measure` and `soup draft distill` disagreed on the same-tokenizer precondition, and `measure` discarded a completed measurement when the assisted arm failed.

`measure` gated on `same_tokenizer()` (tokenizer vocab + probe ids), which **accepts** a pair whose tokenizers are identical but whose `config.vocab_size` differs by padded embedding rows (e.g. Qwen2.5 large←small: 152064 vs 151936). Transformers' assisted generation gates on `config.vocab_size`, so the run died with `ValueError: The main and assistant models have different tokenizers` deep inside `generate()` — **after** the expensive load. `soup draft distill` already refuses that pair up front on `config.vocab_size`. Worse, `measure` wrote its report only after the assisted arm, so when that arm raised, the acceptance rate and plain throughput it had already computed were thrown away (no JSON).

### Fix

- `measure` now applies the **same `config.vocab_size` precondition as `distill`**, refusing a mismatched pair **before any model loads**. `same_tokenizer()` stays as an additional check.
- The shared reader is factored into `_pair_vocab_sizes_or_fail`, so the two commands cannot drift. It also reads a composite model's `get_text_config()`, so a multimodal target (e.g. Llava, whose top-level `vocab_size` is absent) is no longer refused with "could not read model config".
- `measure` writes the report **incrementally**: acceptance + plain throughput are persisted before the assisted arm runs, then upgraded in place if it returns.
- The report records an `assisted_status` (`pending` / `complete` / `untimed` / `crash` / `interrupted`), so a failed arm is distinguishable on disk from a completed or un-run one — previously crash / untimed / interrupt wrote byte-identical JSON. A failed arm stays exit 0 (the arm is a best-effort *throughput* measurement; the `--min-acceptance` gate is unaffected and still exits 2), Ctrl-C is re-raised and keeps exit 130.
- Recording that status is best-effort: an `OSError` while updating the report cannot replace the exception being handled, so a crashed arm still prints its warning and Ctrl-C still exits 130.

Maps to the four acceptance criteria in the issue: (1) measure/distill accept and refuse the same pairs; (2) a transformers-incompatible pair is refused before any model is loaded; (3) a failure in the assisted arm still leaves acceptance + plain-throughput on disk; (4) a test on a vocab-size-mismatched pair with matching tokenizers (the Qwen2.5 shape).

## Type of change

- [x] Bug fix

Tests live in the matching module `tests/test_v07133.py` (`TestDraftMeasureVocabGate` + `TestVocabSizeOfCompositeConfig`). Every torch boundary is monkeypatched, so they run without the training stack.

**RED — the original symptom, on unmodified `main` with only the new tests applied:**

```
>       assert result.exit_code == 1, (result.output, repr(result.exception))
E       assert 0 == 1                       # measure ACCEPTED the mismatched pair, loaded, wrote a report

>       assert result.exit_code == 0, (result.output, repr(result.exception))
E         ', "ValueError('The main and assistant models have different tokenizers')")
E       assert 1 == 0                       # assisted arm raised; report was NOT written
FAILED tests/test_v07133.py::TestDraftMeasureVocabGate::test_config_vocab_mismatch_refused_before_any_model_loads
FAILED tests/test_v07133.py::TestDraftMeasureVocabGate::test_assisted_arm_failure_keeps_acceptance_and_plain_on_disk
```

**RED — by mutation.** Each load-bearing line broken in turn; the control that dies is named:

```
drop the pre-arm incremental write   -> 3 failed  (test_hard_death_in_the_assisted_arm_keeps_the_incremental_report + both write-guard cases)
drop the get_text_config() fallback  -> 2 failed  (test_composite_config_resolves_through_get_text_config[32000] and [152064])
hardcode the fallback to 32000       -> 1 failed  (…[152064] — the second size exists so a constant cannot satisfy it)
let the fallback shadow the top level-> 1 failed  (test_top_level_vocab_size_still_wins)
change the gate's `!=` to `>`        -> 1 failed  (test_config_vocab_mismatch_refused_before_any_model_loads[151936-152064])
drop the OSError guard on the status write -> 2 failed (test_a_failed_status_write_does_not_mask_the_arm_outcome[crash-0] and [interrupt-130])
```

**GREEN — unmutated:**

```
tests/test_v07133.py::TestDraftMeasureVocabGate
tests/test_v07133.py::TestVocabSizeOfCompositeConfig
14 passed
```

**Full local replay** (`ruff` + the touched module + the two repo-wide ratchets it is exposed to), run on `upstream/main` first and then on this branch:

```
== ruff check src/soup_cli/ tests/ (blocking CI job) ==
All checks passed!            (baseline and final)

== pytest tests/test_v07133.py tests/test_cli_help_assertions_are_ansi_safe.py tests/test_cli_startup_is_light.py ==
baseline (upstream/main): 8 failed, 140 passed, 1 skipped
final    (this branch):   8 failed, 154 passed, 1 skipped
```

The 8 failures are identical on both sides (`TestMeasureAcceptance` / `TestMeasureThroughput` — they `import torch`, which this torch-free environment does not have); the full matrix with the training stack and the 77% coverage gate runs in CI. Diff coverage on the changed production lines: 41/41.

## Checklist

- [x] `pytest tests/ -v` passes (torch-free classes locally; full matrix runs in CI)
- [x] Updated relevant docs — `docs/commands.md` documents the exit contract and the `assisted_status` values, `CHANGELOG.md` `[Unreleased] → Fixed`
- [x] New tests added (`TestDraftMeasureVocabGate`, `TestVocabSizeOfCompositeConfig`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,8 +83,10 @@ reproducing 70+ versions of notes.
   so a multimodal target like Llava is no longer refused), keeps `same_tokenizer()`
   as an additional check, and writes the report incrementally so a failing assisted
   arm leaves the acceptance rate and plain throughput on disk. The report records
-  an `assisted_status` (`complete` / `untimed` / `crash` / `interrupted`) so a
-  failed arm is distinguishable on disk from a completed or un-run one.
+  an `assisted_status` (`pending` / `complete` / `untimed` / `crash` /
+  `interrupted`) so a failed arm is distinguishable on disk from a completed or
+  un-run one — `pending` is what a report keeps when the process dies mid-arm
+  and no handler runs, which is the case the incremental write exists for.
 
 ## [0.73.3] - 2026-08-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,7 +70,8 @@ reproducing 70+ versions of notes.
   the "no matching architecture" advisory now share one message.
 
 - **`soup draft measure` now refuses a mismatched pair up front and no longer
-  discards a completed measurement when the assisted arm fails (#344).** `measure`
+  discards a completed measurement when the assisted arm fails
+  (#344 by @ousamabenyounes in #409).** `measure`
   gated on `same_tokenizer()` (tokenizer vocab + probe ids), which accepts a pair
   whose tokenizers are identical but whose `config.vocab_size` differs by padded
   embedding rows (e.g. Qwen2.5 large←small) — exactly the pair `distill` refuses.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,20 @@ reproducing 70+ versions of notes.
   the wrong kernel or crashing. The two call sites that separately hand-wrote
   the "no matching architecture" advisory now share one message.
 
+- **`soup draft measure` now refuses a mismatched pair up front and no longer
+  discards a completed measurement when the assisted arm fails (#344).** `measure`
+  gated on `same_tokenizer()` (tokenizer vocab + probe ids), which accepts a pair
+  whose tokenizers are identical but whose `config.vocab_size` differs by padded
+  embedding rows (e.g. Qwen2.5 large←small) — exactly the pair `distill` refuses.
+  Transformers' assisted generation gates on `config.vocab_size`, so the run died
+  with "different tokenizers" deep inside `generate()`, after the expensive load,
+  and because the report was written only after that arm every completed
+  acceptance/plain-throughput number was thrown away. `measure` now uses the same
+  `config.vocab_size` precondition as `distill` (refusing before any model loads),
+  keeps `same_tokenizer()` as an additional check, and writes the report
+  incrementally so a failing assisted arm leaves the acceptance rate and plain
+  throughput on disk.
+
 ## [0.73.3] - 2026-08-18
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,10 +78,13 @@ reproducing 70+ versions of notes.
   with "different tokenizers" deep inside `generate()`, after the expensive load,
   and because the report was written only after that arm every completed
   acceptance/plain-throughput number was thrown away. `measure` now uses the same
-  `config.vocab_size` precondition as `distill` (refusing before any model loads),
-  keeps `same_tokenizer()` as an additional check, and writes the report
-  incrementally so a failing assisted arm leaves the acceptance rate and plain
-  throughput on disk.
+  `config.vocab_size` precondition as `distill` (refusing before any model loads;
+  the shared `_vocab_size_of` also reads a composite model's `get_text_config()`,
+  so a multimodal target like Llava is no longer refused), keeps `same_tokenizer()`
+  as an additional check, and writes the report incrementally so a failing assisted
+  arm leaves the acceptance rate and plain throughput on disk. The report records
+  an `assisted_status` (`complete` / `untimed` / `crash` / `interrupted`) so a
+  failed arm is distinguishable on disk from a completed or un-run one.
 
 ## [0.73.3] - 2026-08-18
 
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -190,7 +190,7 @@ soup shrink --model <id|path> --drop-ratio 0.25 --calib c.jsonl -o shrunk  Depth
 soup shrink ... --drop-layers N --heal h.jsonl --heal-steps 200 --device cpu  Drop N layers + distill-heal (fuse LoRA back to one dense model)
 soup shrink ... --tolerance 0.10 --plan-only [--attach-to-registry <id>]  Ppl-regression tolerance / print importance table only / registry attach
 soup draft measure --target <m> --draft <d> --prompts p.jsonl  Draft acceptance rate + real plain-vs-assisted tok/s (exit 0 measured — a best-effort assisted-arm failure stays 0 and is recorded as `assisted_status`; 2 below `--min-acceptance`; 1 error before results) (v0.71.33)
-soup draft measure ... --min-acceptance 0.6 -o report.json  Exit 2 below the floor (CI gate) / write the JSON report (fields incl. `assisted_status`: complete/untimed/crash/interrupted)
+soup draft measure ... --min-acceptance 0.6 -o report.json  Exit 2 below the floor (CI gate) / write the JSON report (fields incl. `assisted_status`: pending/complete/untimed/crash/interrupted)
 soup draft distill --target <tuned> --draft-base <tiny> --data d.jsonl -o draft/  Distil a DENSE speculative-decoding draft + register it (v0.71.33)
 soup draft distill ... --steps N --device cpu --force --plan-only  Training budget / device / overwrite -o / render the config only
 soup draft list                               List local drafts that `soup serve --auto-spec` will pick up (v0.71.33)
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -189,8 +189,8 @@ soup mcp serve --allow-execute                Implies --allow-mutating; enables 
 soup shrink --model <id|path> --drop-ratio 0.25 --calib c.jsonl -o shrunk  Depth-prune least-important layer block + SHIP/DON'T-SHIP ppl verdict (exit 0/2/1) (v0.71.29)
 soup shrink ... --drop-layers N --heal h.jsonl --heal-steps 200 --device cpu  Drop N layers + distill-heal (fuse LoRA back to one dense model)
 soup shrink ... --tolerance 0.10 --plan-only [--attach-to-registry <id>]  Ppl-regression tolerance / print importance table only / registry attach
-soup draft measure --target <m> --draft <d> --prompts p.jsonl  Draft acceptance rate + real plain-vs-assisted tok/s (exit 0/2/1) (v0.71.33)
-soup draft measure ... --min-acceptance 0.6 -o report.json  Exit 2 below the floor (CI gate) / write the JSON report
+soup draft measure --target <m> --draft <d> --prompts p.jsonl  Draft acceptance rate + real plain-vs-assisted tok/s (exit 0 measured — a best-effort assisted-arm failure stays 0 and is recorded as `assisted_status`; 2 below `--min-acceptance`; 1 error before results) (v0.71.33)
+soup draft measure ... --min-acceptance 0.6 -o report.json  Exit 2 below the floor (CI gate) / write the JSON report (fields incl. `assisted_status`: complete/untimed/crash/interrupted)
 soup draft distill --target <tuned> --draft-base <tiny> --data d.jsonl -o draft/  Distil a DENSE speculative-decoding draft + register it (v0.71.33)
 soup draft distill ... --steps N --device cpu --force --plan-only  Training budget / device / overwrite -o / render the config only
 soup draft list                               List local drafts that `soup serve --auto-spec` will pick up (v0.71.33)
--- a/src/soup_cli/commands/draft.py
+++ b/src/soup_cli/commands/draft.py
@@ -174,6 +174,34 @@ def _write_draft_report(report: AcceptanceReport, output: str) -> None:
     )
 
 
+def _record_assisted_status(
+    report: AcceptanceReport, output: Optional[str], status: str
+) -> AcceptanceReport:
+    """Stamp ``status`` on ``report`` and persist it best-effort.
+
+    Only for the assisted-arm handlers: a write that raises there would replace
+    the exception being handled — swapping the "assisted arm crashed" warning,
+    or Ctrl-C's exit code, for an unrelated ``OSError`` — and so lose exactly the
+    outcome the handler exists to record (#344 review). The pre-arm write has
+    already put acceptance + plain throughput on disk, so a failed status update
+    is a warning, not a failure. The success path deliberately does NOT use this:
+    there is no exception to mask, and failing to write the completed report is a
+    real error.
+    """
+    report = replace(report, assisted_status=status)
+    if output is None:
+        return report
+    try:
+        _write_draft_report(report, output)
+    except OSError as exc:
+        console.print(
+            f"[yellow]Warning:[/] could not record the assisted-arm outcome in "
+            f"{escape(output)} ({escape(str(exc))}); the acceptance rate and "
+            f"plain throughput written before the arm are still on disk."
+        )
+    return report
+
+
 def _resolve_trust(model_id: str, requested: bool = False) -> bool:
     from soup_cli.utils.trust_remote import (
         model_requires_trust_remote_code,
@@ -748,14 +776,10 @@ def measure(
         # an untimed run — otherwise all three write byte-identical reports
         # (#344 review). Record the outcome, then re-raise so the exit code and
         # the "arm is best-effort" contract are unchanged.
-        report = replace(report, assisted_status="interrupted")
-        if output is not None:
-            _write_draft_report(report, output)
+        _record_assisted_status(report, output, "interrupted")
         raise
     except Exception as exc:  # noqa: BLE001 — assisted arm is best-effort
-        report = replace(report, assisted_status="crash")
-        if output is not None:
-            _write_draft_report(report, output)
+        report = _record_assisted_status(report, output, "crash")
         console.print(
             f"[yellow]Warning:[/] assisted-generation throughput could not be "
             f"measured ({escape(str(exc))}); the acceptance rate and plain "
--- a/src/soup_cli/commands/draft.py
+++ b/src/soup_cli/commands/draft.py
@@ -135,6 +135,11 @@ def _vocab_size_of(model_id: str, trc: bool = False) -> int:
 
     config = AutoConfig.from_pretrained(model_id, trust_remote_code=trc)
     vocab = getattr(config, "vocab_size", None)
+    if vocab is None and hasattr(config, "get_text_config"):
+        # Composite / multimodal configs (e.g. LlavaConfig) keep vocab_size on
+        # the text sub-config, not the top level; get_text_config() returns it
+        # (#344 review). Shared by measure and distill, so this fixes both.
+        vocab = getattr(config.get_text_config(), "vocab_size", None)
     if vocab is None:
         raise ValueError(f"{model_id} config has no vocab_size")
     return int(vocab)
@@ -738,7 +743,19 @@ def measure(
             num_assistant_tokens=num_assistant_tokens,
             max_new_tokens=max_new_tokens,
         )
+    except KeyboardInterrupt:
+        # A Ctrl-C during the arm must be distinguishable on disk from a crash or
+        # an untimed run — otherwise all three write byte-identical reports
+        # (#344 review). Record the outcome, then re-raise so the exit code and
+        # the "arm is best-effort" contract are unchanged.
+        report = replace(report, assisted_status="interrupted")
+        if output is not None:
+            _write_draft_report(report, output)
+        raise
     except Exception as exc:  # noqa: BLE001 — assisted arm is best-effort
+        report = replace(report, assisted_status="crash")
+        if output is not None:
+            _write_draft_report(report, output)
         console.print(
             f"[yellow]Warning:[/] assisted-generation throughput could not be "
             f"measured ({escape(str(exc))}); the acceptance rate and plain "
@@ -752,9 +769,12 @@ def measure(
                 report,
                 tok_s_assisted=assisted,
                 speedup=assisted / plain if plain else None,
+                assisted_status="complete",
             )
-            if output is not None:
-                _write_draft_report(report, output)
+        else:
+            report = replace(report, assisted_status="untimed")
+        if output is not None:
+            _write_draft_report(report, output)
 
     console.print(render_draft_panel(report))
     if output is not None:
--- a/src/soup_cli/commands/draft.py
+++ b/src/soup_cli/commands/draft.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 import json
 import math
 import os
+from dataclasses import replace
 from pathlib import Path
 from typing import TYPE_CHECKING, NoReturn, Optional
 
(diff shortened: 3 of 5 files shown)


MakazhanAlpamys (changes requested):
Reviewed. The gate is the right shape and it is genuinely mutation-proof — three things before merge, and the second is the substantive one.

I broke each of the four load-bearing lines and a test failed every time: forcing the vocab comparison to `if False:` fails `test_config_vocab_mismatch_refused_before_any_model_loads`; dropping the pre-assisted `_write_draft_report` fails two tests; turning the assisted-arm `except` into a `raise` fails the same; and breaking `distill`'s side of the shared helper fails `TestDraftDistillCli::test_vocab_mismatch_rejected`. Both sides of the helper are guarded, which is what I most wanted to see from a shared extraction. `tests/test_v07133.py` 118 passed / 1 skipped; ruff clean; test-name diff +2/−0 with nothing weakened.

**On #345 — I checked, and this is not #417's failure mode.** The gate reads `AutoConfig.vocab_size` (`draft.py:131-140`) and touches nothing on the acceptance path; both new tests stub `measure_acceptance` to a constant `(75, 100)` and assert `0.75`, which is a round-trip, not an instrument reading. Worth saying out loud though: the PR makes the #345-affected number *more durable* by flushing it to disk earlier, and a report written by this path still carries a penalised-logits acceptance rate with nothing marking it as such.

## 1. A partial report is indistinguishable from a complete one

I ran three outcomes through the real CLI:

| run | exit | JSON on disk |
|---|---|---|
| assisted arm crashed (incomplete) | **0** | `tok_s_assisted: null, speedup: null` |
| assisted arm succeeded but untimed (complete) | 0 | `tok_s_assisted: null, speedup: null` |
| Ctrl-C mid-arm (truncated) | 130 | `tok_s_assisted: null, speedup: null` |

All three files are **byte-identical on disk**, and because `except Exception` does not catch `KeyboardInterrupt`, the interrupt case prints no warning at all.

Two things that removes. Before this PR, `report.json` existed only if the run finished — *"file exists ⇒ complete"* was a real invariant, and nothing replaces it. And the exit code moved **1 → 0** for a failed assisted arm (I confirmed the pre-PR behaviour raises and exits 1), while `docs/commands.md:191-192` documents this command as `exit 0/2/1` and describes it as a CI gate. A gate that returns 0 on a failed measurement is the shape of defect this project cares most about.

The house pattern already exists, in `utils/layer_stream_runtime.py:966-977` — `StepPeak` has **three** outcomes deliberately, not two, precisely so a failure cannot be read as a benign absence. Please add an explicit field, e.g. `assisted_status: "measured" | "failed" | "untimed"`, and state which exit code a failed assisted arm should carry.

## 2. `_vocab_size_of` hard-refuses composite configs

It reads **top-level** `config.vocab_size`. On installed transformers 4.57.6, `LlavaConfig().vocab_size` is missing, so `measure` on a VLM pair now dies with "could not read model config" where it previously proceeded and was decided by `same_tokenizer()`. `Qwen2VLConfig` does expose it, so this is arch-dependent rather than uniform — and the real `Qwen/Qwen3.5-4B-Base` config has no top-level `vocab_size` either, only `text_config.vocab_size`.

`cfg.get_text_config()` exists in 4.57.6 and returns `32000` for `LlavaConfig`. Since `distill` shares the helper, one line fixes both. The condition is pre-existing in `distill`, but the shared helper you just introduced is exactly the right place to fix it once.

## 3. Rebase

`CHANGELOG.md` only — `draft.py` has no upstream drift. While you are there, please fold the bullet into the existing top `### Fixed` block rather than opening a sixth one.

And update `docs/commands.md` for whatever exit taxonomy you settle on.

Note `main` was briefly red this morning through no fault of any PR — #407 and #408 collided semantically after both merged green. Repaired in `7bc3f83`.

MakazhanAlpamys (changes requested):
Re-reviewed in an isolated worktree with `PYTHONPATH` pinned to it, plus a detached `origin/main` worktree for the pre-PR baseline.

**Request 3's rebase is done** — you are on `16e4e6d`, `MERGEABLE / CLEAN`, full CI green. Thank you.

For transparency on the rest, measured rather than assumed: GitHub records the head I reviewed as `7932017f`, and diffing it against the new head over `commands/draft.py` and `tests/test_v07133.py` gives **0 lines**. So requests 1 and 2 are untouched. No complaint — my review did ask for the rebase, and you delivered it. I just re-measured instead of guessing what the push contained.

## 1 — partial vs complete, still reproduces

Four outcomes through the real Typer CLI, hashing the file on disk:

```
crash      exit=0    warned=True   sha256[:12]=effdf68b01dd   tok_s_assisted=None  speedup=None
untimed    exit=0    warned=False  sha256[:12]=effdf68b01dd   tok_s_assisted=None  speedup=None
interrupt  exit=130  warned=False  sha256[:12]=effdf68b01dd   tok_s_assisted=None  speedup=None
complete   exit=0    warned=False  sha256[:12]=70f2a4472649   tok_s_assisted=30.0  speedup=1.5
```

Three different outcomes, **one SHA-256**. No `assisted_status` field anywhere in `commands/draft.py` or `utils/draft.py`. Baseline re-confirmed on `main`: `assisted-arm crash: exit=1, report_written=False` — so the "file exists ⇒ complete" invariant really did exist before this PR.

### Two new mutations sharpen why this matters

- Delete the warning message entirely (`except Exception: pass`) → **118 passed**
- Widen to `except BaseException`, which would silently swallow the Ctrl-C case → **118 passed**

The control is your own `raise` mutation, which kills `test_assisted_arm_failure_keeps_acceptance_and_plain_on_disk` — so the block *is* covered, but neither its **message** nor its **exception width** is.

That compounds the finding rather than sitting beside it. With the JSON byte-identical across crash / untimed / interrupt, that yellow console line is currently the **only** thing distinguishing a failed assisted arm — and it is precisely the part nothing pins. It can be deleted with all 118 tests green.

## 2 — `_vocab_size_of` still refuses composite configs

Reproduced on installed transformers 4.57.6:

```
LlavaConfig top-level vocab_size : MISSING
LlavaConfig get_text_config()    : 32000
_vocab_size_of(<saved LlavaConfig>) -> ValueError: ... config has no vocab_size
```

`get_text_config` appears nowhere in `draft.py`. Since `distill` shares the helper, one line fixes both.

## 3b / 3c

The bullet still opens a **sixth** `### Fixed` in `[Unreleased]` (base `16e4e6d` has 5, this has 6) rather than folding into the existing top block. And `git diff origin/main...HEAD -- docs/` is empty, so `docs/commands.md:191-192` still documents `exit 0/2/1` while the code now exits 0 on a failed assisted arm — the two contradict each other.

## The gate itself remains excellent

I re-broke all four load-bearing lines and each killed a test: the measure vocab gate → 1; dropping the pre-assisted write → 2; `raise` in the assisted arm → 1; `distill`'s side of the shared helper → 2. Both sides of that shared extraction are genuinely held, which is what I most wanted to see.

118 passed / 1 skipped, 160 passed across the source-grep suite, ruff clean, zero removed assertions.

Items 1 and 2 remain.

ousamabenyounes:
Both remaining items are done.

**1 — the outcomes are now distinguishable on disk.** `AcceptanceReport` carries an `assisted_status` (`pending` / `complete` / `untimed` / `crash` / `interrupted`). The arm records the outcome and writes it: a crash sets `crash` (and still prints the warning), a non-positive number sets `untimed` (silent), and a new explicit `except KeyboardInterrupt` branch sets `interrupted` and re-raises so the exit stays 130. Four new tests pin each status; the crash test now asserts the warning text (deleting the message fails it) and the interrupt test asserts exit 130 + `interrupted` (widening the `except` or dropping the handler fails it). So the yellow line is no longer the only distinguisher, and it is covered.

**2 — composite configs.** `_vocab_size_of` now falls back to `get_text_config()` when the top-level `vocab_size` is absent, so a `LlavaConfig` target resolves `32000` instead of raising. It is the shared helper, so `distill` is fixed in the same line.

**3b/3c** — folded the extra `### Fixed` in `[Unreleased]` back to five (matches `main`); `docs/commands.md` now states a failed assisted arm stays exit 0 and is recorded as `assisted_status`, resolving the contradiction with the code.

Local: `122 passed` in `test_v07133.py`, ruff clean.

MakazhanAlpamys (changes requested):
Re-verified on a clean worktree (transformers 4.57.6, Python 3.10, Windows). **Both items are genuinely fixed**, and I could measure it rather than infer it.

## Item 1 — four outcomes, four distinct files

Driven through the **real** CLI process — real Typer dispatch, real exit codes, only the model-loading and throughput seams stubbed:

| case | `sha256[:12]` | exit | `assisted_status` |
|---|---|---|---|
| complete | `5eb7e1e3218c` | 0 | `complete` |
| untimed | `a36838ed047e` | 0 | `untimed` |
| crash | `a7525ce66162` | 0 | `crash` |
| interrupt | `902bfa46e74c` | **130** | `interrupted` |

Four distinct hashes where three were previously byte-identical at `effdf68b01dd`. The crashed arm now leaves `acceptance_rate` and `tok_s_plain` on disk with a loud warning.

Documenting the exit-0 contract at `docs/commands.md:191-192` rather than changing the exit code is the right call, and I want to say why I agree: a best-effort **throughput** arm failing should not fail a gate that is about **acceptance**. I confirmed the `--min-acceptance` → exit 2 path still fires after a crashed arm, so the gate itself is intact.

**Both mutations that survived last round are dead:**

| mutation | result |
|---|---|
| `except Exception: pass` | **killed** — `assert 'pending' == 'crash'` |
| delete only the warning | **killed** — `assert "could not be measured" in result.output` |
| widen to `except BaseException` | **killed** — `assert 0 == 130` |

Each 1 failed / 120 passed against a 121-passed baseline, with two controls in the same file failing when broken.

The five-state `assisted_status` — including `pending` for a process killed mid-arm — is a better answer than the three I suggested.

## Item 2 works, but nothing guards it

The code is right: real `LlavaConfig()`, no top-level `vocab_size`, `get_text_config().vocab_size == 32000`, `_vocab_size_of` returns `32000`.

But deleting the `get_text_config()` branch at `commands/draft.py:138-142` leaves the suite at **121 passed, identical to baseline** — a surviving mutation that is provably behavioural, since with it removed that same `LlavaConfig` raises `ValueError: ... config has no vocab_size`. And `grep -rn "get_text_config\|Llava" tests/` finds nothing.

Three lines close it: patch `AutoConfig.from_pretrained` to return a `LlavaConfig()` and assert `_vocab_size_of(...) == 32000`.

## Two nits, neither blocking

`CHANGELOG.md:94` inserts a **new** `### Fixed` heading right after your bullet, splitting the block the #371 bullet was in. You also removed a pre-existing duplicate, so the total is net-unchanged at 5 — but the new one is not needed.

`assisted_status` is written to disk but never rendered in the panel, so on screen `untimed` and `crash` differ only by the warning line.

ruff clean; 264 passed / 1 skipped across the broader set; no assertions removed. Rebase is trivial — one `CHANGELOG.md` conflict where your #344 bullet and `main`'s #321 bullet landed at the same anchor; keep both.

Add the Llava test and this is good to go.

MakazhanAlpamys:
Re-reviewed at `645a608`, isolated worktree, `PYTHONPATH` at that tree. Verdict up front: **the exit-code and status contract you built is genuinely pinned — but the two behaviours your comment leads with, persist-early and the composite-config fallback, are not covered by anything in the suite.** I found that by mutation, not by reading, and I built the missing test myself before asking you for it.

## What holds, proven by breaking it

**The `assisted_status` work is real and well guarded.** Eleven mutations; eight killed:

| mutation | result |
|---|---|
| `if target_vocab != draft_vocab:` -> `if False:` | KILLED — `test_config_vocab_mismatch_refused_before_any_model_loads` |
| `_pair_vocab_sizes_or_fail` reads tokenizer vocab instead of `_vocab_size_of` | KILLED — **18** failures across measure *and* distill |
| reword the warning to "unavailable" | KILLED — the crash test |
| delete the whole `except KeyboardInterrupt` branch | KILLED |
| keep the branch, drop the trailing `raise` | KILLED |
| collapse it into `except BaseException` | KILLED |
| revert the incremental write entirely | KILLED |

The interrupt path being pinned three independent ways is the part I want to name: you can break it by deleting the handler, by dropping the re-raise, or by widening it, and each one fails a test. That is what "pinned" should mean.

Also checked: **no test deleted or weakened** (`git diff 16e4e6d..645a608 --numstat -- tests/` -> `185 0`), the two existing tests that gained a `_vocab_size_of` monkeypatch need it (the new gate would otherwise hit the network) and still assert the same things, nothing in the diff is undescribed. The write is atomic and path-validated on every call (`atomic_write_text` -> `enforce_under_cwd_and_no_symlink` -> `mkstemp` + `os.replace`), so no truncated JSON is reachable. The gate really does run before either model loads — `loaded == []` is asserted. `121 passed, 1 skipped` in `test_v07133.py`; `642 passed, 3 skipped` across the seven files I ran; ruff clean.

## Blocker 1 — "writes the report incrementally" is untested

Drop **only** the pre-arm write and keep the crash/interrupt/complete writes:

```
$ # delete lines 734-735:  if output is not None: _write_draft_report(report, output)
$ pytest tests/test_v07133.py -q
121 passed, 1 skipped
```

I re-ran that myself rather than take it from my reviewer. Every current test also passes against a version that writes the report only from the `except` handlers — so nothing distinguishes *incremental* from *at the end*. And the one thing the early write actually buys is survival of a death for which **no handler runs**: your `except Exception` and `except KeyboardInterrupt` do not cover a `SystemExit`, a `MemoryError`, or a hard kill. That case is the whole point of #344's second half, and it is the case with no test.

Here is the control. I wrote it, ran it against your code (**passes**) and against the mutant (**fails**, `AssertionError` on `report.exists()`):

```python
def _hard_death(model, tok, prompts, *, assistant_model=None, **kw):
    if assistant_model is not None:
        raise SystemExit(137)   # BaseException: no except in measure() runs
    return 20.0

# ... invoke measure with -o report.json ...
assert report.exists(), "acceptance + plain throughput were discarded"
data = json.loads(report.read_text(encoding="utf-8"))
assert data["acceptance_rate"] == 0.75
assert data["tok_s_plain"] == 20.0
assert data["assisted_status"] == "pending"
```

It slots into your `_run_measure` helper. It also becomes the only coverage of `pending`, which is currently reachable on disk and asserted nowhere.

## Blocker 2 — the `get_text_config()` fallback has no test

```
$ grep -rn "get_text_config" tests/ | wc -l
0
```

Delete the five-line fallback and all 220 draft tests still pass. The code is correct — I confirmed with a Llava-shaped stub that it resolves `32000` — but "a `LlavaConfig` target resolves 32000 instead of raising" is currently a claim in a PR comment, not a property of the repo. Three tests close it, patching `AutoConfig.from_pretrained`:

1. no top-level `vocab_size`, `get_text_config()` returns an object with `vocab_size = 32000` -> expect `32000`;
2. a plain config with `vocab_size = 151936` whose `get_text_config()` **raises** -> expect `151936`, which proves the fallback does not shadow the normal path;
3. neither -> the existing `ValueError`.

The second is the one that matters: without it, a fallback that accidentally took priority would still pass.

The shared-helper half of your claim **is** verified — rerouting `_vocab_size_of` breaks 18 tests across both commands, so `measure` and `distill` genuinely cannot drift.

## Finding 3 — the gate is pinned in one direction only (small)

Changing `measure`'s `!=` to `>` survives: your test only exercises target 152064 <- draft 151936. `distill`'s test happens to use the opposite direction, so between the two commands neither covers both. Parametrizing the existing test over both directions is a two-line change.

## Nits

- **`pending` is undocumented.** `docs/commands.md` and the CHANGELOG both enumerate `complete/untimed/crash/interrupted`. It is the dataclass default and it lands on disk. A consumer parsing the field would not expect it.
- **The CHANGELOG needs a rebase, and "matches `main`" is stale.** Your head has 5 `### Fixed` headings under `[Unreleased]`; current `origin/main` has 2. Expect a conflict there. (The duplicate section from the earlier round *is* correctly folded back — that item is done.)
- **Raw `result.output` assertions.** `assert "could not be measured" in result.output` bypasses the ANSI-strip + whitespace-collapse rule. I ran the class under `FORCE_COLOR=1 TERM=xterm-256color` and it is **green today** — only `Warning:` carries markup and the phrase lands mid-line at width 80 — so this is not a live CI risk, it is one wording change away from being one. That class of assertion has turned this repo's CI red four times; route both through the existing helper.
- **`_write_draft_report` is unguarded inside the two handlers.** An `OSError` there replaces the original exception and loses exactly the results the handler exists to preserve.

## Summary

Two controls (blockers 1 and 2), one parametrize (finding 3), the `pending` line in the two docs, and a CHANGELOG rebase. No behaviour change requested — the code is right; what is missing is the suite proving the two things you told me you delivered. Everything I asked for above, I ran first.

MakazhanAlpamys:
Re-reviewed at `bee548e`. **All three findings are closed, and I verified each by re-running the mutation that survived before.** One line left.

| what survived before | now |
|---|---|
| drop only the pre-arm `_write_draft_report` (keep the crash/interrupt writes) | **KILLED** — 3 tests, incl. `test_a_failed_status_write_does_not_mask_the_arm_outcome[interrupt-130]` |
| delete the 5-line `get_text_config()` fallback | **KILLED** — 2 tests, incl. `test_composite_config_resolves_through_get_text_config[152064]` |
| `!=` -> `>` in the vocab gate (half the gate) | **KILLED** — `test_vocab_mismatch_rejected` |

The first row is the one I care about most. The persist-early behaviour is the whole second half of #344, and before this it was indistinguishable from writing the report at the end — every existing test passed against a version that only wrote from the `except` handlers. It is now pinned, and `pending` is documented in `docs/commands.md` alongside the other four statuses.

The second is closed properly too: the composite-config fallback had **zero** test references before, so "a `LlavaConfig` target resolves 32000" was a claim in a PR comment rather than a property of the repo. Now deleting the fallback fails by name.

## One line to go

The CHANGELOG entry is correctly placed — line 72, inside `[Unreleased]`, above the frozen `## [0.73.3]` at 91 — but it reads `(#344).` with no attribution. House form is:

```
(#344 by @ousamabenyounes in #409)
```

`grep -c "by @ousamabenyounes in #409" CHANGELOG.md` returns `0`.

I ask for this in the PR rather than adding it at release time because adding it at release time is exactly when it goes wrong: four suffixes in v0.73.3 turned out wrong or incomplete when I checked them against `gh` — an issue number in the PR slot, a credit missing entirely, and one citing an issue that had already shipped.

CI is at 2 green with 10 still running after your rebase. Add the suffix and I will merge on green.

ousamabenyounes:
Added the requested CHANGELOG attribution in `02fd460`: `#344 by @ousamabenyounes in #409`, retained inside `[Unreleased]`.

Validation: the maintainer attribution grep passes, the post-merge placement check passes, ruff is clean, and all 12 GitHub CI checks are green on this SHA.

MakazhanAlpamys:
Merged as `e9dff98`.

**Four rounds, and I do not think any of them were wasted** — but the reason is worth naming, because it was not persistence. Each round you came back with the mutation matrix rather than an assertion that it was fixed, and twice that matrix was how *I* learned something: the persist-early behaviour and the `get_text_config()` fallback were both things you had described in a comment and that nothing in the repo actually held to.

What I ran before merging this round, rather than reading the diff:

| what survived at some point | now |
|---|---|
| drop only the pre-arm `_write_draft_report`, keep the crash/interrupt writes | **KILLED** — 3 tests, incl. `test_a_failed_status_write_does_not_mask_the_arm_outcome[interrupt-130]` |
| delete the 5-line `get_text_config()` fallback | **KILLED** — `test_composite_config_resolves_through_get_text_config[152064]` |
| `!=` -> `>` in the vocab gate (half the gate) | **KILLED** — `test_vocab_mismatch_rejected` |

And on this final push I checked the one thing a CHANGELOG-only round still deserves: `git diff bee548e5 02fd4609` is **one file, +2/-1**. The code is byte-identical to what I had already broken and re-broken, so I did not re-run the matrix on a diff that could not have changed it. Placement verified in the *merged* tree, not branch-local — `git merge-tree --write-tree origin/main` gives rc 0 with the bullet at line 85, inside `[Unreleased]` (13) and clear of the frozen `[0.73.3]` (103). CI green on `02fd4609` itself, which I checked against `headRefOid` after being burned by a cached `pull/head` ref on another PR this week.

**The durable finding here is the persist-early one.** A completed acceptance measurement was being discarded because the report was only written after the arm that transformers can refuse — and that was indistinguishable from writing it at the end, since every pre-existing test passed against a version that only wrote from the `except` handlers. A partial result is now legible on disk instead of being byte-identical to a successful one, and `pending` is documented in `docs/commands.md` alongside the other four statuses. The interrupt path being pinned three independent ways (branch present, trailing `raise` present, not collapsed into `except BaseException`) is the kind of over-pinning I will not complain about.

Credited in `CONTRIBUTORS.md`. Thank you for the CHANGELOG suffix in the PR rather than leaving it to release time — the last release had four attribution suffixes turn out wrong when I checked them against `gh`, which is exactly why I now ask for it here.

PR #410, opened 2026-08-15
feat(ship): commit the noise floor to soup.yaml via eval.ship.noise_floor (#406)

### LINKED ISSUE #406: eval.ship has no noise_floor field, so --noise-floor is the one gate-policy flag that cannot be committed to soup.yaml
## What

Every `soup ship` gate-policy flag is settable in a committed `soup.yaml` under `eval.ship` and read back by `--config` — except the one added in v0.73.2.

`ShipConfig` (`src/soup_cli/config/schema.py`) carries:

`task_eval` · `task_mode` · `general_suite` · `forgetting_threshold` · `judge_model` · `baseline`

…and `commands/ship.py` wires each with the CLI > config > default precedence via `_flag_is_default`. **`--noise-floor` has no field and no wiring**, so it can only be passed on the command line.

## Why it matters

`soup ci init` exists to bind a PR merge gate to a committed config, so the gate policy lives in the repo and is reviewable. A team that wants the noise floor enforced in CI has to hand-edit `.github/workflows/soup-gate.yml` to add the flag — the one piece of gate policy that cannot be committed alongside the rest.

It is also a plain inconsistency: an operator who has `eval.ship` in their config will reasonably expect to set the floor there, and gets no error telling them otherwise — the key is simply ignored.

## Proposed fix path

Small and mechanical; the pattern is already established six times over.

1. Add `noise_floor: Optional[int]` to `ShipConfig`, bounded `[MIN_NOISE_FLOOR_RUNS, MAX_NOISE_FLOOR_RUNS]` — import those from `utils.ship_verdict` rather than re-typing `2` and `10`, so the schema bound and the CLI validator cannot disagree (same reasoning as `stream_buffers` importing its bounds from `utils.layer_stream`).
2. Reject bool-as-int, matching the CLI validator and the rest of the schema.
3. Wire it in `commands/ship.py` beside the other five `_flag_is_default(ctx, ...)` reads.
4. Extend `soup ci init` so a config-bound gate carries it.

One decision worth making explicitly: `_config_sha_of` **excludes** `eval.ship` from the recipe hash on purpose (v0.71.39 — tuning the gate threshold must not invalidate evidence about an unchanged model). The new field should follow that same rule, not become an exception.

Also confirm the interaction that already exists at CLI level: `--noise-floor` is refused with `--evidence` (nothing to re-run). A config-supplied floor in that combination should behave the same way rather than being silently ignored.

## Acceptance criteria

- [ ] `eval.ship.noise_floor: 3` in a `soup.yaml` is honoured by `soup ship --config soup.yaml`.
- [ ] An explicit `--noise-floor` on the command line **overrides** the config value (CLI > config > default), with a test.
- [ ] Out-of-range and bool-as-int values are rejected at config-parse time, naming the actual reason.
- [ ] `eval.ship` stays excluded from `_config_sha_of` — a test pins that adding this field did not change the recipe hash.
- [ ] A config-supplied floor combined with `--evidence` behaves exactly as the CLI flag does there.
- [ ] `soup ci init` emits the flag when the bound config sets it.

## Refs

- Introduced in v0.73.2 alongside `--noise-floor` (#376) — the flag shipped without its config surface.
- Code: `src/soup_cli/config/schema.py::ShipConfig`, `src/soup_cli/commands/ship.py` (`_flag_is_default` block), `src/soup_cli/utils/ci_workflow.py`.

ousamabenyounes:
## Summary

`eval.ship` in `soup.yaml` could carry every `soup ship` gate-policy flag — `task_eval`, `task_mode`, `general_suite`, `forgetting_threshold`, `judge_model`, `baseline` — and `soup ship --config` read each back with CLI > config > default precedence. The one exception was `--noise-floor`, added in v0.73.2 (#376) without a config surface, so it could only be passed on the command line and the key was silently ignored if placed under `eval.ship`.

This adds that surface (proposed-fix-path steps 1–3), following the pattern already established six times over.

Fix #406

## What changed (acceptance criteria)

- [x] `eval.ship.noise_floor: 3` is honoured by `soup ship --config soup.yaml` — a bounded `Optional[int]` field on `ShipConfig`, wired in `commands/ship.py` beside the other five `_flag_is_default(ctx, ...)` reads.
- [x] An explicit `--noise-floor` on the command line **overrides** the config value (CLI > config > default) — test added.
- [x] Out-of-range and bool-as-int values are rejected at config-parse time — bounds `[MIN_NOISE_FLOOR_RUNS, MAX_NOISE_FLOOR_RUNS]` **imported from `utils.ship_verdict`** (not re-typed), so the schema bound and the CLI validator cannot disagree — same reasoning as `stream_buffers` importing its bounds from `utils.layer_stream`.
- [x] `eval.ship` stays excluded from `_config_sha_of` — a test pins that adding this field did not change the recipe hash (`_config_sha_of` already excludes the whole `eval.ship` block, so a floor tunes the verdict, never the model fingerprint).
- [x] A config-supplied floor combined with `--evidence` **behaves exactly as the CLI flag does there** — refused (nothing to re-run offline; a floor already recorded in the evidence is applied), not silently ignored — test added.
- [ ] **`soup ci init` emitting the flag (step 4) — left for your direction, see the open question below.**

## Open question — step 4 / `soup ci init` (why this is a draft)

I could not resolve step 4 without guessing at intended behaviour, and would rather ask than ship a design you'd have to unwind.

The generated gate step is `soup ship --evidence <ev> --config <cfg>` (offline). With the acceptance-criteria rule that a **config-supplied floor under `--evidence` is refused** (implemented above), binding a config that *sets* `noise_floor` to that gate step would make the generated gate exit 3. So the two goals pull against each other:

- **"config floor + `--evidence` ⇒ refused"** (criterion 5), and
- **"`soup ci init` emits the flag when the bound config sets it"** (criterion 6)

can't both hold for the *same* config used as the gate's `--config` binding.

`noise_floor` is a live-measurement input, so it is only meaningful on the **producer** run (`soup ship ... --config <cfg> --emit-evidence <ev>`), which the header note already documents and which is `--config`-bound (so it already carries the floor into the measured evidence). If your intent for step 4 is one of:

1. `soup ci init` reads the bound config and, when it sets `noise_floor`, surfaces it in the header producer-command note (gate step unchanged), or
2. `soup ci init` warns that a floor-carrying config is measured at evidence-production time and not re-run at the gate, or
3. something else,

I'll add it in this PR. I kept steps 1–3 self-contained so the config surface lands cleanly regardless of how step 4 is shaped.

Tests live in the `ShipConfig` area module `tests/test_v07139.py` (extending `TestShipConfigSchema`, `TestShipConfigReader`, `TestComputeProvenance`).

**RED** — new tests run against unmodified `main` (only the test file applied, production files untouched):

```
8 failed, 2 passed, 57 deselected
FAILED ...TestShipConfigSchema::test_noise_floor_default_is_none
FAILED ...TestShipConfigSchema::test_parses_noise_floor_under_eval
FAILED ...TestShipConfigSchema::test_rejects_out_of_bounds_noise_floor[1]
FAILED ...TestShipConfigSchema::test_rejects_out_of_bounds_noise_floor[11]
FAILED ...TestShipConfigSchema::test_rejects_bool_noise_floor[True]
FAILED ...TestShipConfigSchema::test_rejects_bool_noise_floor[False]
FAILED ...TestShipConfigReader::test_config_noise_floor_threads_to_live
FAILED ...TestShipConfigReader::test_config_noise_floor_refused_under_evidence
```

The 2 that pass on `main` are the intended controls: `test_config_sha_excludes_noise_floor` (the hash-exclusion regression pin — a floor is dropped either way, so the sha is unchanged before and after) and `test_explicit_noise_floor_overrides_config` (on `main` the config is ignored, so the explicit flag trivially wins — it corroborates precedence, mirroring the existing `test_explicit_flag_overrides_config`).

**GREEN** — same tests on this branch:

```
10 passed, 57 deselected
```

## Full local suite (no regression)

`ruff check src/soup_cli/ tests/` clean on both. `tests/test_v07139.py`: **57 → 67** (57 existing + 10 new, no failures). `tests/test_cli_startup_is_light.py`: **22 passed, 1 skipped** unchanged — confirming the new schema-scope import of `ship_verdict` pulls no `torch`/`transformers` (`import soup_cli.config.schema` leaves both out of `sys.modules`).

## Files changed

| File | Change |
|------|--------|
| `src/soup_cli/config/schema.py` | `ShipConfig.noise_floor` field + bool-reject validator; bounds imported from `ship_verdict`; `EvalConfig.ship` description updated |
| `src/soup_cli/commands/ship.py` | config read for `noise_floor` beside the other `_flag_is_default` reads |
| `tests/test_v07139.py` | schema + reader + hash-exclusion tests |
| `docs/evaluation.md` | `eval.ship` key list + live-input note |
| `CHANGELOG.md` | `[Unreleased] → Added` entry |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,20 @@ reproducing 70+ versions of notes.
 
 ## [Unreleased]
 
+### Added
+
+- **`eval.ship.noise_floor` is now committable to `soup.yaml` (#406).** Every
+  `soup ship` gate-policy flag was settable in a committed config and read back by
+  `--config` — except `--noise-floor` (added in v0.73.2), which had no field and
+  could only be passed on the command line, so a team enforcing a floor in CI had
+  to hand-edit the workflow. `ShipConfig` gains a bounded `noise_floor`
+  (`[2, 10]`, imported from `ship_verdict` so the schema and the CLI validator
+  cannot disagree; bool-as-int rejected), wired with the same CLI > config >
+  default precedence as the other five flags. As a live-measurement input it is
+  measured only when producing evidence and refused under `--evidence` exactly as
+  the flag is; it follows `forgetting_threshold` in being excluded from the
+  recipe `config_sha`, so setting a floor never invalidates evidence.
+
 ### Fixed
 
 - **`soup draft distill --steps N` now delivers ~N optimiser steps instead of
--- a/docs/evaluation.md
+++ b/docs/evaluation.md
@@ -463,7 +463,10 @@ every PR instead of relying on a hand-edited JSON file.
   reproduces an identical verdict.
 - **`--config soup.yaml`** reads a committed `eval.ship` block for the gate defaults
   (`task_eval` / `task_mode` / `general_suite` / `forgetting_threshold` / `judge_model` /
-  `baseline`); an explicit CLI flag always wins. This makes the gate reviewable in a PR diff.
+  `baseline` / `noise_floor`); an explicit CLI flag always wins. This makes the gate reviewable
+  in a PR diff. `noise_floor` is a live-measurement input like the `--noise-floor` flag: it is
+  measured when a live run produces evidence and, like the flag, is refused under `--evidence`
+  (there is nothing to re-run offline — a floor already recorded in the evidence is applied).
 - **Provenance + staleness.** With `--emit-evidence`, `--config` STAMPS a `provenance` block
   (`config_sha` — a semantic, order-insensitive recipe hash that EXCLUDES the `eval.ship` gate
   policy, so tuning the threshold never invalidates evidence — plus `base_model` and a
--- a/src/soup_cli/commands/ship.py
+++ b/src/soup_cli/commands/ship.py
@@ -1095,6 +1095,13 @@ def ship(
                 baseline = ship_cfg.baseline
             if _flag_is_default(ctx, "forgetting_threshold"):
                 forgetting_threshold = ship_cfg.forgetting_threshold
+            if _flag_is_default(ctx, "noise_floor") and ship_cfg.noise_floor is not None:
+                # A config floor is a live-measurement request just like the CLI
+                # flag: _validate_noise_floor_flag re-checks bounds below, and the
+                # --evidence guard refuses it there exactly as it refuses the flag
+                # (a floor is measured against a live base, nothing to run
+                # offline; a floor already in the evidence is applied on read).
+                noise_floor = ship_cfg.noise_floor
 
     _validate_task_mode_flag(task_mode)
     threshold = _validate_threshold_flag(forgetting_threshold)
--- a/src/soup_cli/config/schema.py
+++ b/src/soup_cli/config/schema.py
@@ -19,6 +19,14 @@ from soup_cli.utils.layer_stream import (
     SUPPORTED_STREAM_TASKS as _STREAM_SUPPORTED_TASKS,
 )
 
+# Noise-floor bounds live with the ship verdict so the schema bound and the
+# `--noise-floor` CLI validator can never disagree (ship_verdict has no torch,
+# same reasoning as stream_buffers importing its bounds from layer_stream).
+from soup_cli.utils.ship_verdict import (
+    MAX_NOISE_FLOOR_RUNS,
+    MIN_NOISE_FLOOR_RUNS,
+)
+
 # v0.39.0 Part C — per-pattern LoRA rank/alpha bounds
 _MAX_LORA_RANK_PATTERN_KEYS = 256
 _MAX_LORA_RANK_PATTERN_VALUE = 1024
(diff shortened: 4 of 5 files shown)


MakazhanAlpamys (approved):
Verified rather than read.

**Precedence is pinned.** I inverted it — made the config beat an explicit flag:

```
$ # if ship_cfg.noise_floor is not None:   (the _flag_is_default guard removed)
FAILED tests/test_v07139.py::TestShipConfigReader::test_explicit_noise_floor_overrides_config
1 failed, 66 passed
```

Exactly one test notices, and it is the right one. `tests/test_v07139.py` 67 passed otherwise, CI 11/11.

**Importing the bounds instead of restating them is the detail that makes this durable:**

```python
from soup_cli.utils.ship_verdict import MAX_NOISE_FLOOR_RUNS, MIN_NOISE_FLOOR_RUNS
```

`ship_verdict` has no torch, so the schema stays on the light import path, and the CLI validator and the Pydantic field now cannot disagree about what 2..10 means. That is the same reasoning `stream_buffers` uses for importing its bounds from `layer_stream`, and you cited it — a config bound and a flag validator drifting apart is the exact shape of defect this repo keeps finding.

`_flag_is_default(ctx, ...)` via Click's `ParameterSource` rather than comparing against the default value is also right: `--noise-floor 3` typed explicitly and `3` inherited from the default are indistinguishable by value, and only the source tells them apart.

Merging.

MakazhanAlpamys:
Reviewed and **approved on content** — flagging that it is still marked **draft**, so I cannot merge it. Mark it ready and I will, no re-review needed.

Verified rather than read:

```
$ # precedence inverted: config beats an explicit flag
FAILED tests/test_v07139.py::TestShipConfigReader::test_explicit_noise_floor_overrides_config
1 failed, 66 passed
```

Exactly one test notices, and it is the right one. CI 12/12.

The detail that makes this durable is importing the bounds instead of restating them:

```python
from soup_cli.utils.ship_verdict import MAX_NOISE_FLOOR_RUNS, MIN_NOISE_FLOOR_RUNS
```

`ship_verdict` carries no torch, so the schema stays on the light import path, and the Pydantic field and `_validate_noise_floor_flag` can no longer disagree about what 2..10 means. A config bound and its flag validator drifting apart is a defect shape this repo keeps rediscovering — `stream_buffers` imports from `layer_stream` for the same reason, and you cited it.

Using Click's `ParameterSource` rather than comparing against the default value is also right: `--noise-floor 3` typed explicitly and `3` inherited are indistinguishable by value, and only the source separates them.

Issue #394, opened 2026-08-14
pip install "soup-cli[mlx]" cannot train: the MLX path imports datasets + TRL before reaching the MLX branch


MakazhanAlpamys:
Split out of #392 (reported by @armanbot-jpg), because the fix is either a real dependency change or a documentation correction and those are different discussions.

## What

`docs/` presents `pip install "soup-cli[mlx]"` as the install path for Apple Silicon fine-tuning. It is not sufficient: `soup train` reaches `datasets`, then TRL, before control ever gets to the MLX branch, so the run dies on an import that has nothing to do with MLX. In practice a user needs the full `[train]` extra — which pulls **torch** onto a machine that was going to use MLX precisely so it would not need torch.

Reported on macOS arm64, Python 3.12, mlx 0.30.x / mlx-lm 0.31.x.

## Why it is worth fixing rather than documenting away

The whole proposition of `backend: mlx` is "you do not need the CUDA stack". An install path that pulls torch anyway keeps the dependency it exists to avoid, and the doc currently promises otherwise — so a user following it hits an ImportError that names a package they never asked for.

## Two candidate fixes, and they are not equivalent

1. **Make `[mlx]` actually sufficient.** Push the `datasets` / TRL imports behind the backend branch so an MLX run never touches them, and add whatever `[mlx]` genuinely needs. This is the fix that keeps the promise, and it is the larger one — it means auditing the import path from `soup train` to `MLXSFTTrainerWrapper.setup()` for anything torch-shaped.
2. **Correct the docs** to say `[mlx]` is additive to `[train]`, not a replacement. Honest, one line, and it concedes the proposition.

(1) is the right answer if the import audit is small; (2) is the right answer if it is not. Whoever picks this up should measure before choosing — the deciding fact is how many of the imports between the CLI entry point and the MLX branch are structural rather than incidental.

## Acceptance criteria

- [ ] a clean venv with **only** `pip install "soup-cli[mlx]"` either completes a tiny MLX SFT run, or the docs no longer claim it can
- [ ] if (1): a test that pins the MLX path does not import torch/TRL — the same shape as `tests/test_cli_startup_is_light.py`, which asserts the *runtime* property (`"torch" not in sys.modules`) rather than a syntactic one
- [ ] `docs/backends-and-ops.md` and any other page naming the `[mlx]` extra agree with whichever answer is taken

MakazhanAlpamys:
[...]

PR #411, opened 2026-08-15
fix(stream): classify a fast virtio disk by a measured read, not the rotational flag (#365)

### LINKED ISSUE #365: detect_disk_kind cannot see through virtio: a 1.5 GB/s cloud disk is classified as an HDD and refused the streaming disk tier
## What

`utils.layer_stream.detect_disk_kind()` classifies a virtio block device as `hdd`, so `stream_source: disk` is refused on it.

Measured on the H100 box (`benchmarks/gate-h100-validation.md` L1113–1136):

```
$ lsblk -d -o NAME,ROTA,MODEL,SIZE
NAME ROTA MODEL SIZE
vda     1        1T
$ cat /sys/block/vda/queue/rotational
1
$ dd if=/dev/zero of=/root/ddtest bs=1M count=2048 oflag=direct   # 548 MB/s write
$ dd if=/root/ddtest of=/dev/null bs=1M iflag=direct              # 1.5 GB/s read
>>> detect_disk_kind()
'hdd'
```

1.5 GB/s sequential read is not a spinning disk. It is a virtio device backed by fast host storage, exposing no rotational hint, so the guest kernel defaults `rotational` to 1 and the classifier's `BusType`/`MediaType` route finds nothing to read.

## Impact

**The audience for layer streaming's disk-overflow tier is disproportionately on cloud VMs**, and on a cloud VM a genuinely NVMe-backed disk is currently refused. The guard is not wrong to be conservative — an actual HDD does 160 seeks per step and destroys the feature — but it currently cannot distinguish "spinning" from "cannot tell".

## Fix path

Three-way, not two-way:

1. Keep `BusType`/`MediaType` as the authoritative route where it answers (it does on bare metal; note `BusType` must beat `MediaType`, since an NVMe drive reports `MediaType: SSD`).
2. When it does not answer, fall back to a **measured** short sequential read with `O_DIRECT` and classify on throughput rather than on the `rotational` flag.
3. Ship an explicit `--disk-kind` / `training.stream_disk_kind` override with a loud notice, for the case where even that is wrong.

Do **not** simply loosen the guard or treat `rotational=1` as unknown-and-therefore-fine: the HDD refusal is correct and the failure it prevents is severe.

## Acceptance criteria

- [ ] A virtio-shaped probe (no `BusType`, no `MediaType`, `rotational=1`, fast measured read) classifies as usable
- [ ] A genuinely slow device still classifies as `hdd` and is still refused — a control, so the fix cannot be satisfied by "always allow"
- [ ] The override path prints what it overrode and what was detected
- [ ] The measured-throughput fallback is bounded in time (the existing Windows probe already costs ~9 s cold, which is why `choose_tier` takes a callable — keep that property)

ousamabenyounes:
## Summary

Fix #365

`utils.layer_stream.detect_disk_kind()` classified a paravirtual (virtio) block
device as `hdd`, so the layer-streaming disk-overflow tier (`stream_source: disk`)
was refused on it — the very audience the tier targets, since disk overflow is
mostly used on cloud VMs whose disks are NVMe-backed but expose no media hint.

The Linux probe trusted `/sys/block/<dev>/queue/rotational`, which a guest kernel
defaults to `1` for virtio with no way to tell "spinning" from "cannot tell".

**Fix (the issue's three-way path):**
- `rotational=0` stays authoritative — a device that declares itself solid state is one.
- `rotational=1` is no longer trusted blindly: the media type is decided by a **bounded, best-effort O_DIRECT sequential-read measurement**. NVMe-class throughput (≥ 1 GB/s) earns the tier; a genuinely slow disk still classifies `hdd` and is **still refused** (80 shards × 2 reads = 160 seeks/step, plan P11) — a control, so the fix cannot be satisfied by "always allow".
- New `training.stream_disk_kind` override (`nvme`/`ssd`/`hdd`) for the case where even the measurement is wrong; it prints what it overrode beside what was detected.

`choose_tier`'s NVMe-only tier policy is intentionally **unchanged** — a measured-fast device is classified as NVMe-class, so no existing invariant is widened.

## Reproduces with (unmodified `main`)

A virtio-shaped probe — no NVMe device, `rotational=1`, a 1.5 GB/s measured read:

```
detect_disk_kind (virtio, 1.5 GB/s read) => 'hdd'
choose_tier RAISED => layer streaming needs NVMe or more RAM: ... the detected disk is 'hdd' (not NVMe).
```

New tests live in the disk-kind area module `tests/test_v07203.py::TestDiskKindMeasuredFallback`
(the module already owning `TestDiskKindDetection`), covering all four acceptance
criteria plus the review-hardening cases.

RED — the new tests on unmodified `main` (prod reverted, only tests applied):

```
10 failed in 1.10s
```
(`AttributeError: 'TrainingConfig' object has no attribute 'stream_disk_kind'` /
missing `_classify_measured_read` / virtio still returns `'hdd'`.)

GREEN — with the fix:

```
27 passed in 1.82s
```

## Full local suite (ruff blocking + torch-free area, iso-or-better)

The change is torch-free (`layer_stream.py` is on the light-CLI import path, "no
top-level torch"). Baseline `main` vs this branch:

- `ruff check src/soup_cli/ tests/` — clean on both.
- disk-kind area classes: baseline 16 passed → branch 27 passed (+11 new), no regressions.
- light-CLI startup invariant: 22 passed, 1 skipped (touched imports stay torch-free).

The heavy torch pytest matrix + 77% coverage gate run in CI.

## Files changed

| File | Change |
|------|--------|
| `src/soup_cli/utils/layer_stream.py` | measured-throughput classifier + O_DIRECT probe + `resolve_disk_kind` override |
| `src/soup_cli/config/schema.py` | `training.stream_disk_kind` field + footgun validator |
| `src/soup_cli/trainer/stream_setup.py` | wire the override into the disk-kind callable |
| `docs/performance-and-quantization.md` | disk-tier scope, config example, footgun list, troubleshooting |
| `CHANGELOG.md` | `[Unreleased] → Fixed` entry |
| `tests/test_v07203.py` | `TestDiskKindMeasuredFallback` (11 tests) |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ docs/superpowers/
 
 # Local scratch / temp scripts at repo root (e.g. _smoke.py, _scratch.py) — never committed
 /_*.py
+
+# Layer-streaming disk-throughput probe scratch file (soup doctor --disk / training pre-flight)
+.soup-diskprobe-*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,19 @@ reproducing 70+ versions of notes.
   unknown outcome is never recorded as success, and the richer
   `completed`/`failed` terminal statuses are left untouched. `soup runs` no
   longer hardcodes `running` for a non-running row.
+- **Layer streaming now accepts a fast virtio/cloud disk instead of refusing it
+  as an HDD (#365).** `detect_disk_kind` trusted `/sys/block/<dev>/queue/rotational`,
+  which a paravirtual (virtio) block device defaults to `1` with no media hint —
+  so a genuinely NVMe-backed cloud disk (measured 1.5 GB/s read) was classified
+  `hdd` and denied the disk-overflow tier, the very audience the tier targets.
+  `rotational=0` stays authoritative (solid state); when the flag is unreliable
+  (`rotational=1`), the media type is now decided by a bounded, O_DIRECT
+  sequential-read measurement rather than the flag — NVMe-class throughput
+  (>= 1 GB/s) earns the tier, a genuinely slow disk still classifies `hdd` and is
+  still refused (160 seeks/step, plan P11). A new `training.stream_disk_kind`
+  override (`nvme`/`ssd`/`hdd`) is the escape hatch for the case where even that
+  is wrong; it prints what it overrode beside what was detected. Disk detection
+  may write a small scratch file next to the streamed shards to run the probe.
 
 ### Fixed
 
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -237,6 +237,7 @@ training:
   stream_buffers: 2            # Double-buffering; range [2, 8]
   # stream_vram_override: 4_000_000_000   # Bytes to assume free (v0.73.x); see below
   # stream_vram_probe: true    # Decide the fit by MEASURING one step (sft only); see below
+  # stream_disk_kind: nvme     # Override auto disk-kind detection: nvme/ssd/hdd; see below
 ```
 
 ```bash
@@ -277,7 +278,7 @@ The 3B NF4-vs-bf16 rows differ by 1.85×, but attribute that to **pinning, not a
 Untied `embed_tokens` + `lm_head` stay resident and unquantised (2.10 GB of the 8B row's 3.32 GB), which is why 8B sits close to this card's ceiling; treating them as streamed large layers is deferred beyond v0.72.3.
 
 **Honest scope:**
-- **RAM tier + disk overflow (v0.72.3).** `stream_source: auto` picks RAM when it fits, falls back to NVMe disk when not; SATA/HDD rejected. Correctness verified; disk performance unmeasured on the reference box.
+- **RAM tier + disk overflow (v0.72.3).** `stream_source: auto` picks RAM when it fits, falls back to NVMe disk when not; SATA/HDD rejected. Correctness verified; disk performance unmeasured on the reference box. A paravirtual (virtio) disk reports `rotational=1` with no media hint, so a genuinely NVMe-backed cloud disk was misread as an HDD and refused (#365); detection now measures a bounded O_DIRECT sequential read when the rotational flag is unreliable and admits NVMe-class throughput (>= 1 GB/s), while a genuinely slow disk stays rejected. Set `training.stream_disk_kind: nvme` (or `ssd`/`hdd`) to override when detection is still wrong — the resolved value is printed beside what was detected.
 - **Llama / Qwen / Mistral / Gemma / Gemma2 / Gemma3-Text / Phi / Phi3** (all verified bit-exact in bf16 and NF4), `task: sft`, `backend: transformers`, `modality: text`.
 - **Batch sizes, gradient accumulation, `--resume` / `--hf-resume`** all now work (v0.72.3).
 - **Pre-Ampere cards (T4, P100, V100, GTX 16xx, RTX 20xx) now stream in fp16 instead of bf16.** Until this fix the store dtype was hardcoded to bf16 on every CUDA device, so the entire free notebook tier was streaming a dtype its GPU has no units for, and nothing said so — it could not fail on the Ampere card every number above was measured on. fp16 is bit-exact against a resident reference of matching numerics, `0.000000e+00` in both quantisations, exactly as bf16 is.
@@ -352,7 +353,7 @@ prints this advice when it sees you accumulating.
 - `task: kto` with `batch_size: 1` → TRL's KL term is degenerate at batch 1; refused when the config is read rather than minutes later after sharding
 - `lora.use_dora` / `lora.use_vera` / `lora.init_strategy` other than `random` → these initialise from the real base weight, which is on the meta device under streaming
 - `unfrozen_parameters`, `lisa_enabled`, `packing`, `multipack`, `use_fsdp2_compile`, `train_router_only`, `expand_layers` → each independently rewrites or re-freezes the same layers
-- `stream_source` / `stream_buffers` / `stream_vram_override` / `stream_vram_probe` set while `stream_layers: false` → a footgun, refused
+- `stream_source` / `stream_buffers` / `stream_vram_override` / `stream_vram_probe` / `stream_disk_kind` set while `stream_layers: false` → a footgun, refused
 - `stream_vram_probe` on any task other than `sft` → the probe runs a plain causal-LM step, which *is* the SFT step but is not a preference loss. Measured at one matching shape it is conservative there too (6.02 GB against a real DPO step's 5.30 GB, +13.5%), but one shape is not a validation, so it is not offered for `dpo`/`orpo`/`simpo`/`kto` yet
 - an architecture outside the supported list (llama / qwen2 / qwen3 / mistral / gemma / gemma2 / gemma3_text / phi / phi3) → named explicitly
 
@@ -402,6 +403,7 @@ output: ./output
 
 **Troubleshooting:**
 - **"layer streaming needs the base to fit in RAM"** — the base is larger than free RAM. Set `stream_source: auto` to fall back to the NVMe disk tier, free RAM, or pick a smaller base.
+- **"layer streaming needs NVMe or more RAM … the detected disk is 'hdd'"** on a fast cloud disk — a virtio device reports `rotational=1` with no media hint. Detection now measures the disk when the flag is unreliable; if it still misreads yours, set `training.stream_disk_kind: nvme` to force the tier on (`ssd`/`hdd` force it off).
 - **"could not page-lock the base … falling back to a PAGEABLE RAM store"** — expected on a busy machine. Training continues, more slowly. Close other applications to keep the pinned store.
 - **"layer streaming does not support model_type=…"** — the supported list is llama / qwen2 / qwen3 / mistral / gemma / gemma2 / gemma3_text / phi / phi3. Multimodal `gemma3` is excluded on purpose; use `gemma3_text`.
 - **"predicted peak … exceeds free VRAM" and you believe it is wrong** — lower `batch_size` or `data.max_length` first. Otherwise there are two escape hatches and they are not interchangeable. `training.stream_vram_probe: true` (`sft` only) **measures** one real forward+backward at your configured shape and decides on that, printing the prediction beside it; it costs one step (1–5 s measured) and it can also refuse a run the formula accepted. `training.stream_vram_override: <bytes>` instead **replaces** the free-VRAM figure the check runs against — that is an assertion you are making, not a measurement, so raising it past a real limit is an OOM on Linux and a silent spill on Windows. Prefer the probe when you want to be told the truth; use the override when you know something the driver cannot report.
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -62,7 +62,9 @@ def doctor(
         help=(
             "Detect the disk media type (NVMe / SSD / HDD). Layer streaming's "
             "disk overflow tier needs NVMe. Off by default: the probe costs ~9 s "
-            "on Windows."
+            "on Windows, and on Linux it WRITES a ~64 MiB scratch file "
+            "(.soup-diskprobe-*, git-ignored) into the current directory to "
+            "measure sequential read throughput."
         ),
     ),
 ):
--- a/src/soup_cli/config/schema.py
+++ b/src/soup_cli/config/schema.py
@@ -3083,6 +3083,19 @@ class TrainingConfig(BaseModel):
             "unsafe."
         ),
     )
+    stream_disk_kind: Optional[Literal["nvme", "ssd", "hdd"]] = Field(
+        default=None,
+        description=(
+            "Override the auto-detected disk media type for the streaming disk "
+            "overflow tier. Detection classifies a device by a measured "
+            "sequential read when the kernel's rotational flag is unreliable — "
+            "virtio and other paravirtual disks report spinning with no media "
+            "hint, so a fast cloud disk is otherwise refused (#365). Set this "
+            "for the case where even that is wrong: 'nvme' forces the disk tier "
+            "on, 'ssd'/'hdd' force it off. The resolved value is printed beside "
+            "what was detected."
+        ),
+    )
 
     # Sample packing — pack multiple short samples into one sequence
     packing: bool = Field(
@@ -4902,12 +4915,13 @@ class SoupConfig(BaseModel):
                 or tcfg.stream_buffers != 2
                 or tcfg.stream_vram_override is not None
                 or tcfg.stream_vram_probe
+                or tcfg.stream_disk_kind is not None
             ):
                 raise ValueError(
                     "training.stream_source / training.stream_buffers / "
                     "training.stream_vram_override / training.stream_vram_probe "
-                    "set but stream_layers is false; set stream_layers=true to "
-                    "stream the base layer-by-layer."
+                    "/ training.stream_disk_kind set but stream_layers is false; "
+                    "set stream_layers=true to stream the base layer-by-layer."
                 )
             return self
         # v0.72.4 — the four preference losses join SFT. DPO and KTO take their
--- a/src/soup_cli/trainer/stream_setup.py
+++ b/src/soup_cli/trainer/stream_setup.py
@@ -172,11 +172,11 @@ class StreamingSetupMixin:
             TIER_DISK,
             TIER_RAM,
             build_stream_plan,
-            detect_disk_kind,
             dtype_bytes,
             estimate_stream_store_bytes,
             free_ram_bytes,
             render_stream_panel,
+            resolve_disk_kind,
             resolve_stream_dtype,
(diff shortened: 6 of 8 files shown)


MakazhanAlpamys (changes requested):
The design is right — three-way path exactly as #365 laid it out, `rotational=0` short-circuits without probing, and `choose_tier`'s NVMe-only policy is genuinely unchanged. CI green on all nine cells. I also reproduced the Linux-only path on real Linux rather than trusting the Windows skips: all four acceptance scenarios behave, the O_DIRECT probe works on ext4, cleans up its scratch file, and costs 0.11–0.23 s.

Three findings, and the first is the one that matters.

## 1. Blocking — the measurement is not reproducible at its own threshold

I ran the shipped `_measure_seq_read_bytes_per_s` repeatedly on **one unchanged filesystem**:

| batch | measured | verdicts |
|---|---|---|
| A — 8 samples, quiet-ish | 0.749 – 1.952 GB/s | 2 of 8 refused |
| B — 12 samples, right after ~20 prior 64 MiB cycles | 0.146 – 0.306 GB/s | **12 of 12 refused** |
| C — 12 samples, quiescent | 1.078 – 2.119 GB/s | 0 of 12 refused |

Same box, same disk, and the verdict swings from always-admit to always-refuse on transient load. A single unrepeated 64 MiB read against a hard 1.0 GB/s cliff is not a stable discriminator — and shared, contended, burst-throttled cloud VMs are exactly where that noise lives, which is the environment #365 is about.

**This module already learned this.** `layer_stream_runtime.measure_gemm_tflops` takes best-of-3, with the reason recorded in the source: *"best-of-N is correct because a CEILING has one-sided noise"* — five repeats spread 38%, reporting 3.54 against 6.75 TFLOPS at the same clock. This probe is the same shape of measurement and takes one sample. A repeat costs ~0.13 s.

Batch B says best-of-N alone will not save it under sustained contention, which is why this needs finding 2 as well.

## 2. Blocking — the refusal never shows the number it refused on

`choose_tier` still raises:

> the detected disk is 'hdd' (not NVMe). Free RAM, pick a smaller base, or move the model to an NVMe drive.

No measured throughput, no indication whether that came from a flag or from a measurement, and advice that is actively wrong for someone already on NVMe-backed cloud storage. Combined with finding 1, an operator refused by a noisy sample has nothing to act on — they are told to move to a disk they are already on. Put the measured GB/s in the message and the refusal becomes a diagnosis.

## 3. Blocking — `soup doctor --disk` now writes 64 MiB into the user's CWD, undocumented

`commands/doctor.py:412` calls `detect_disk_kind(".")`. On Linux, whenever the first `/sys/block` entry with a `rotational` file reports `1`, this now performs a 64 MiB write + fsync + O_DIRECT read **in the current working directory** — usually the user's git repo.

Not a virtio-only path: on WSL the first sorted device is `loop0` with `rotational=1`, so an ordinary Ubuntu box with snaps takes it. Nothing in the command's help, `docs/`, or the CHANGELOG says the diagnostic writes anything. Cleanup is in a `finally` so Ctrl-C is safe, but SIGKILL leaks a `.soup-diskprobe-*` file that is not gitignored, and a read-only CWD silently degrades the verdict to `hdd`.

`soup doctor --disk` will also now print **"NVMe"** for a virtio device. That is the intent, but it is a media-type claim the code can no longer back — worth rewording to say what was measured.

## Worth knowing: the fix is defended by one assertion on one of nine cells

I reverted `_probe_disk_kind` to the pre-fix `return "hdd" if value == "1" else "ssd"`:

```
Windows: 129 passed, ZERO failures
Linux:   1 failed  (test_virtio_fast_disk_now_earns_the_disk_tier)
```

Your most recent commit — skipping the `/sys/block` simulation off Linux — is defensible for a Linux-only mechanism, but it means six of the nine CI cells report green against code with the fix entirely removed. The other three mutations are properly caught (always-`nvme` → 1 failed; override ignored → 1 failed; schema footgun removed → 1 failed), so this is specifically the core line that is thin.

## Smaller

- **Criterion 4 asked for a bound and this bounds bytes, not time.** 64 MiB with no deadline or chunking: on a stalled NFS/iSCSI mount the probe can block arbitrarily long, inside `build_stream_plan`, *after* sharding has already run. `test_measurement_is_bounded_and_best_effort` is named for that criterion and only asserts that a missing `O_DIRECT` returns `None` — nothing asserts any time bound.
- **The escape hatch pays for the thing it escapes**: `resolve_disk_kind` runs full detection even when `stream_disk_kind` is set. Deliberate, and documented as such — but if the probe is what is wrong, the override does not avoid it.
- **Duplicated `### Fixed` heading** in the CHANGELOG (a new one at line 15, the existing one at line 30). Same defect PR #388's merge produced. Fold into the existing section; that is also your merge conflict.
- **`except` narrower than the docstring promises**: "any failure returns `None`" but only `OSError`/`ValueError` are caught, and `detect_disk_kind` catches nothing — so on the default path an unexpected exception propagates into `build_stream_plan` and kills the run.

## Cleared

No top-level torch; `shard_dir` provably exists when the callable fires, so the probe measures the right volume; O_DIRECT alignment is sound (page-aligned mmap, 64 MiB block multiple, offset 0); no scratch-file leak across ~35 real invocations; `rotational=0` really does skip the probe, verified with a call counter on real Linux.

---

Minimum to merge: fold the CHANGELOG, repeat the read and surface the measured GB/s in the refusal, and document that detection may write. The diagnosis in #365 was right and this is the right shape of fix — it is the single-sample threshold that needs the work.

ousamabenyounes:
Rebased onto `main` and addressed the minimum-to-merge points.

- **Single-sample threshold** — this was the real weakness. `_measure_seq_read_bytes_per_s` now repeats the O_DIRECT read `_MEASURE_READ_SAMPLES` (3) times and keeps the **best** sample, so a lone cold/slow read can no longer under-measure a fast disk and wrongly refuse it. Each read stays bounded by the 64 MiB cap, so the probe is still well under a second.
- **Surface the measured rate in the refusal** — `choose_tier` now names the rate that earned the refusal, e.g. `... the detected disk is 'hdd' (measured 0.15 GB/s, under the 1.0 GB/s NVMe floor)`, instead of the opaque `'hdd'`. The value is a module-level last-measurement reset at the start of every probe, so it is never a stale figure from another volume.
- **Detection may write** — documented in the CHANGELOG entry (the probe writes a small scratch file next to the shards).

New tests: `test_refusal_cites_the_measured_rate` and `test_the_read_is_repeated_and_the_best_sample_wins`. Local run of the disk suite: `21 passed` (the torch-dependent cases need the GPU stack, unchanged here).

MakazhanAlpamys (changes requested):
Thanks — **the best-of-3 is real**, and I verified that rather than taking it on faith. Mutating the comparison to keep the **worst** sample, and separately to keep the **first**, both change the returned value, and the repeat-count assertion catches the second independently. That property is genuinely pinned.

Three things before merge.

## 1. Blocking — lint is red

```
E501 Line too long (108 > 100)   tests/test_v07203.py:775
```

The `os.readv` monkeypatch lambda. `main` is clean, so the `lint` CI job is failing on this PR. Please run `ruff check src/soup_cli/ tests/` before pushing — your local `21 passed` would not have surfaced it.

## 2. Blocking — "never a stale figure from another volume" is not true, and it fails in the exact way I was worried about

The global is reset only inside `_probe_disk_kind`, but `detect_disk_kind` **caches** — so on a cache hit the probe never runs and never resets. I reproduced three paths that all print a self-contradictory refusal:

```
the detected disk is 'hdd' (measured 2.00 GB/s, under the 1.0 GB/s NVMe floor)
```

A rate **above** the floor, given as the reason the disk fell **under** it.

The path that needs no cache trick at all: set `training.stream_disk_kind: hdd` on a fast virtio disk. Detection runs and measures 2.0 GB/s into the global, the override forces `hdd`, and `choose_tier` cites the fast rate as the cause. Single process, ordinary config.

That is precisely the "worse than the opaque message it replaced" case — the old `'hdd'` told you nothing; this tells you something false, which is harder to argue with.

The root of it is that `_LAST_MEASURED_READ_BYTES_PER_S` is **not coupled to the `kind` argument `choose_tier` was handed**. Cleanest fix is to stop using module state: pass the measured rate alongside the kind, so the note structurally cannot describe a verdict it did not produce. Failing that, clear it on the cache-hit branch **and** render the note only when `kind` actually came from a measurement.

## 3. Not blocking, but please do it — the write is still undocumented where it matters

The CHANGELOG says the probe "may write a small scratch file next to the streamed shards", which covers the training path. My point was `soup doctor --disk`: `doctor.py:412` calls `detect_disk_kind(".")`, so on Linux it writes **64 MiB into the user's CWD** — usually their git repo. The `--disk` help text still advertises only the ~9 s Windows cost, and `.soup-diskprobe-*` is not in `.gitignore`.

One line in the option help plus a `.gitignore` entry closes it.

## Worth knowing about the defence

Both new tests are Linux-only — `pytest -rs` skips them at `test_v07203.py:670` and `:764`. I confirmed that **all four** mutations above (worst-sample, first-sample, reset removed, note removed entirely) pass a full Windows run at 129 passed / 6 skipped.

So the review fixes now carry the same one-third-of-CI defence that the original core line did. Not a blocker on its own, but if any of it can be asserted without a `/sys/block` simulation, it is worth doing — a test that runs on 3 of 9 cells is a test most contributors will never see fail.

The measurement logic itself is right, and the CHANGELOG fold is done correctly — the remaining duplicate `### Fixed` headings are `main`'s, not yours.

ousamabenyounes:
Thanks for measuring rather than re-reading — all three are addressed.

**1. Lint** — split the `os.readv` monkeypatch into a named function; `ruff check src/soup_cli/ tests/` is clean now. (You were right that the local `21 passed` never surfaces E501 — added ruff to my pre-push loop.)

**2. The false measured note** — you nailed the root cause: the rate lived in module state that the cache never reset on a hit and an override left stale. I removed the global entirely. `_probe_disk_kind` now returns a frozen `DiskClassification(kind, measured_bps)` where `measured_bps` is set **only** when the verdict was *derived* from a read; `resolve_disk_kind` strips it for an override (the verdict is the user's, not the probe's); and `choose_tier` reads the rate from the same result it classified. The note is now structurally incapable of citing a rate it did not produce — across cache hits and overrides.

Your exact repro is pinned by a new **CPU-only** regression test (`test_override_verdict_does_not_cite_a_probe_rate`: kind from override, no rate in the message), plus `test_override_returns_a_measureless_classification` (detect 2.0 GB/s underneath, override to hdd → `measured_bps is None`). I also moved the note assertion itself to CPU-only (`DiskClassification("hdd", 150e6)` straight into `choose_tier`), so it runs on all nine CI cells rather than the three Linux ones.

**3. The write** — `soup doctor --disk` now says in its `--disk` help that on Linux it writes a ~64 MiB `.soup-diskprobe-*` scratch file into the CWD, and that pattern is in `.gitignore`.

Local: `395 passed, 15 skipped` across the streaming files, ruff clean.

MakazhanAlpamys:
Merged as [`d3bcd35`](https://github.com/MakazhanAlpamys/Soup/commit/d3bcd35) — 13/13 green. You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) as of [`68294e4`](https://github.com/MakazhanAlpamys/Soup/commit/68294e4), eighth merge.

I re-ran my three repro paths against `bd38d6c` rather than accepting "structurally incapable", in an isolated worktree with `PYTHONPATH` pinned to your branch:

- **Path A (cache hit)** — confirmed the probe does not re-run (`slow2 is slow1`), and the refusal cites `measured 0.15 GB/s`, its own rate, not the 2.00 GB/s measured on the other volume in between.
- **Path B** (`stream_disk_kind: hdd` on a fast virtio disk — the one needing no cache trick) — detection ran, measured 2.00 GB/s, cached `nvme`, and the refusal now reads `the detected disk is 'hdd' (not NVMe).` with **no rate at all**. The contradiction is gone.
- **Path C** — bare string, callable-returning-string and callable-returning-`DiskClassification` all produce the no-note message.

`DiskClassification` is frozen (`FrozenInstanceError` on mutation) and there is exactly **one** two-argument construction site in `src/`, where the kind is derived from that very rate. So the coupling holds on every reachable path.

The credit line names the *second* fix rather than the first, and deliberately. Told the message cited a rate the cache never reset, the easy move was to clear the global on the cache branch — which would have worked and left the same trap for the next person. Removing it, and making the rate travel inside the classification that produced it, is what turns "we remember to reset it" into a property that can be checked.

Two things worth recording from the mutation run:

- Keeping `measured_bps` on override → **1 failed**. Deleting the note → **1 failed**. Both of those **passed a full Windows run last round**, so moving the assertion to CPU-only is real and measured: two of the four previously-invisible mutations now fail on all nine cells rather than three. That was a note at the end of my review, not a blocker, and you did it anyway.
- Best-of-3 confirmed genuinely best — replaying the read with `O_DIRECT` injected gives 67.11 MB/s against 16.78 (worst) and 33.55 (first), three distinct values, so the assertion discriminates.

Four follow-ups, none blocking, all fine to leave:

1. The invariant is upheld by construction discipline rather than by the type — a hand-built `DiskClassification("hdd", 2.0e9)` still renders the contradictory note. Unreachable from production, but a `__post_init__` asserting the pairing would make it structural.
2. `DiskKind = Union[str, Callable[[], str]]` (`layer_stream.py:248`) no longer describes what `stream_setup.py:320` passes, which is a callable returning `DiskClassification`.
3. Worst-sample / first-sample / rotational-revert mutations still survive a Windows run. The test already mocks the whole filesystem, so it needs `monkeypatch.setattr(os, "O_DIRECT", 0x4000, raising=False)` **and** the same for `os.readv` — which is also absent on Windows, so my "just O_DIRECT" framing last round was too simple.
4. The best-of-3 I asked for **tripled** the unbounded-time exposure I raised in round one — three unbounded 64 MiB reads inside `build_stream_plan`, after sharding. My request, not your regression, but it makes a read deadline more worth having than it was.

You still have #409, #416, #418, #420 and #421 open with findings outstanding. No rush from me — but #420 is the cheapest of them: still approved on content, blocked only on a `CHANGELOG.md` rebase.

ousamabenyounes:
Picked up three of the four follow-ups (all still non-blocking, but cheap enough to close now):

1. **Structural pairing** — `DiskClassification.__post_init__` now rejects a `measured_bps` that classifies to a different kind than it is paired with, so the hand-built `("hdd", 2.0e9)` that renders a self-contradictory note raises at construction. The invariant is enforced by the type, not by discipline. Pinned by `test_diskclassification_rejects_a_contradictory_pairing`.
2. **Type alias** — `DiskKind` said `Union[str, Callable[[], str]]`; it now reflects that `choose_tier`'s callable returns a `DiskClassification` (or a bare kind).
3. **Windows coverage** — the best-of-3 test no longer skips off Linux: `os.O_DIRECT` and `os.readv` are injected with `raising=False`, so it runs on all nine cells and the worst-/first-/reset mutations can no longer survive a Windows run.

4. **Read deadline** — you're right that best-of-3 tripled the unbounded-read exposure; a per-read timeout is the real fix and it's a bit more than a follow-up (it wants a bounded-read primitive + its own test), so I'll leave it for a focused pass rather than bolt it on here unless you'd prefer it in this PR.

Local: `127 passed, 10 skipped` on the streaming file, ruff clean.

MakazhanAlpamys:
Heads-up before you lose that work: **this PR is already merged** ([`d3bcd35`](https://github.com/MakazhanAlpamys/Soup/commit/d3bcd35)), so those three follow-ups have nowhere to land from here.

I checked rather than assumed — `fix/issue-365`'s last commit is `08-16T18:40`, which is what got merged, and on current `main` `DiskClassification` has **no** `__post_init__`:

```
$ grep -n "class DiskClassification" -A 14 src/soup_cli/utils/layer_stream.py
251: class DiskClassification:
264:     kind: str
265:     measured_bps: Optional[float] = None      # <- no __post_init__
```

(The `__post_init__` at line 1207 is `LayerSpec`'s, which predates you.)

So please open a **new PR** off current `main` for those three. All three are worth having, and one of them changes the character of the fix:

1. **`__post_init__` rejecting a contradictory pairing** — this is the one I'd most like. It converts the invariant from "the single two-argument construction site happens to derive kind from that rate" into something the type enforces. My review said the property held "by construction discipline, not by the type"; this closes exactly that.
2. **The `DiskKind` alias** — small, and it stops the next reader being misled about what `choose_tier` actually receives.
3. **Windows coverage via `raising=False` injection of `os.O_DIRECT` and `os.readv`** — worth doing precisely because I got it wrong first: I suggested `O_DIRECT` alone, and it needed `readv` too, which only showed up when someone ran it.

Item 4 (the read deadline) is genuinely optional — and note it is a consequence of *my* request, since best-of-3 tripled the unbounded-read exposure. Fine to leave, fine to include.

Reference **#365** rather than a closing keyword in the new PR, since the issue is already closed. And a warning from my own mistake yesterday: GitHub's parser matches `close #365` **even inside a negation** — I wrote "Deliberately does NOT close #394" in a squash body and it closed #394. Bare `#365` or `refs #365` is safe.

Separately: **#420 is rebased and mergeable** — thank you. It is under review now. That leaves #409, #416, #418 and #421 with findings outstanding.

PR #414, opened 2026-08-15
fix(mitigation): skip the coefficient write on a kl_control hold

### LINKED ISSUE #371: Reward-hack controller: pid_lagrangian has never had a valid run, the rollback ladder never fires, and a non-acting kl_control run is numerically identical to log_only

AmirF194:
## Root cause

`_run_bang_bang` (the `kl_control` mode of the reward-hack mitigation controller) calls
`_apply_coefficient(action.new_beta)` on every training step, including a `hold`. On a
`hold`, `bang_bang_step` leaves `reason == "hold"` exactly when `new_beta == beta` (the
dead-band case, and the two clamped-at-ceiling/floor cases), so the write puts back the
value that was already there. With the default `grpo_beta` and `reward_hack_beta_floor`
both at `0.02`, most `kl_control` runs spend most of their steps in `hold`, so a run that
never trips the controller is, step for step, a same-value write repeated rather than the
no-op `log_only` claims to be.

## Fix

Skip `_apply_coefficient` when `action.reason == "hold"`. Since `hold` only occurs when
`new_beta == beta`, this changes no committed β; it only removes the redundant write. The
mitigation log now also carries `mitigation_status` (`held` / `acted` / `released`), so a
step where the controller did nothing is recoverable directly from the log instead of
inferred by parsing the free-text `action` field.

## Scope

This addresses fix-path item 3 of #371 only ("skip the coefficient write on a hold ... log
the distinction"), which is the `kl_control` sub-issue and is fully CPU-testable. Items 1
(`pid_lagrangian` needs a completed run on GPU hardware) and 2 (reconciling the rollback and
beta ladder thresholds, which needs the same GPU runs to observe) are unaddressed here; this
runner has no GPU access to produce or verify either. `Refs #371` rather than `Fixes #371`
for that reason.

## Verification

- `pytest tests/test_v07126.py -k TestKlControlCallback -v`: the three new tests fail against
  unpatched `main` (write-count assertions, missing `mitigation_status` key) and pass on this
  branch, checked in a clean Docker container on both ends of the CI matrix (Python 3.10 and
  3.12).
- Full `pytest tests/test_v07126.py` (190 tests, the whole reward-hack-mitigation suite):
  green on this branch, both Python versions.
- `ruff check src/soup_cli/ tests/`: clean. `mypy`: one pre-existing error in this file at
  line 922, unrelated to this change and present on `main` too (this repo's `type-check` CI
  job is non-blocking).
- Not run: anything requiring GPU (items 1/2 above), and the repository's full test suite
  (17k+ tests across 360 files) beyond the one file this change touches.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,16 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **`kl_control` rewrote the trainer's β/kl_coef on every step, including a `hold`, so a
+  non-acting run was numerically identical to `log_only` (#371).** `_run_bang_bang` called
+  `_apply_coefficient` unconditionally; on a `hold` the controller writes back the value
+  already there, which is a no-op value but not a no-op write. The write is now skipped when
+  the bang-bang step holds, and the mitigation log records `mitigation_status` (`held` /
+  `acted` / `released`) so a non-acting step is distinguishable from an acting one without
+  parsing the free-text `action` reason.
+
+### Fixed
+
 - **`soup draft distill --steps N` now delivers ~N optimiser steps instead of
   N/4.44 (#364).** The epoch count that realised `--steps` divided the request
   by `rows // batch_size`, ignoring that `val_split` (0.1) removes rows from
--- a/src/soup_cli/utils/reward_hack_control.py
+++ b/src/soup_cli/utils/reward_hack_control.py
@@ -1037,12 +1037,23 @@ class RewardHackMitigationCallback(_TrainerCallbackBase):  # type: ignore[misc, 
         vote = self._compute_vote(signals)
         new_state, action = bang_bang_step(policy, self._state, vote=vote)
         self._state = new_state
-        self._apply_coefficient(action.new_beta)
+        held = action.reason == "hold"
+        # bang_bang_step leaves reason == "hold" only when new_beta == beta,
+        # so skipping the write here changes no committed β.
+        if not held:
+            self._apply_coefficient(action.new_beta)
         self._record_action(action.reason)
         telemetry["vote"] = vote
         telemetry["new_beta"] = action.new_beta
         telemetry["tripped"] = action.tripped
         telemetry["action"] = action.reason
+        telemetry["mitigation_status"] = (
+            "held"
+            if held
+            else "released"
+            if action.reason.startswith("relax")
+            else "acted"
+        )
 
     def _request_stop(self, control: Any) -> None:
         if control is not None:
--- a/tests/test_v07126.py
+++ b/tests/test_v07126.py
@@ -1000,6 +1000,72 @@ class TestKlControlCallback:
         for key in ("vote", "new_beta", "tripped", "action"):
             assert key in last
 
+    def test_hold_step_performs_no_coefficient_write(self, tmp_path, monkeypatch):
+        """A dead-band (hold) step must not touch the trainer at all, so a
+        non-acting kl_control run is observably identical to log_only."""
+        monkeypatch.chdir(tmp_path)
+
+        class _WriteCountingTrainer:
+            def __init__(self, beta):
+                self._beta = beta
+                self.writes = 0
+                self.args = types.SimpleNamespace(beta=beta)
+
+            @property
+            def beta(self):
+                return self._beta
+
+            @beta.setter
+            def beta(self, value):
+                self.writes += 1
+                self._beta = value
+
+        cb = _kl_callback(tmp_path, _SeqBuffer([_HEALTHY, _HEALTHY]))
+        trainer = _WriteCountingTrainer(0.02)
+        cb.attach(trainer)
+        cb.on_step_end(None, types.SimpleNamespace(global_step=1), None)
+        cb.on_step_end(None, types.SimpleNamespace(global_step=2), None)
+        assert trainer.writes == 0
+        assert trainer.beta == 0.02
+
+    def test_hold_then_raise_writes_exactly_once(self, tmp_path, monkeypatch):
+        monkeypatch.chdir(tmp_path)
+
+        class _WriteCountingTrainer:
+            def __init__(self, beta):
+                self._beta = beta
+                self.writes = 0
+                self.args = types.SimpleNamespace(beta=beta)
+
+            @property
+            def beta(self):
+                return self._beta
+
+            @beta.setter
+            def beta(self, value):
+                self.writes += 1
+                self._beta = value
+
+        cb = _kl_callback(tmp_path, _SeqBuffer([_HEALTHY, _HACK]))
+        trainer = _WriteCountingTrainer(0.02)
+        cb.attach(trainer)
+        cb.on_step_end(None, types.SimpleNamespace(global_step=1), None)  # hold
+        cb.on_step_end(None, types.SimpleNamespace(global_step=2), None)  # raise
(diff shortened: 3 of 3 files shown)


MakazhanAlpamys (approved):
Verified the load-bearing claim independently rather than taking it from the description, because the whole change rests on it.

**`reason == "hold"` really does imply `new_beta == beta`.** I walked all six paths through `bang_bang_step`: dead-band, trip-band-but-dwell-not-met, trip-band-already-tripped-clamped-at-ceiling, release-band-not-tripped, release-band-patience-not-met, and release-band-clamped-at-floor. In every one, `beta` leaves equal to what came in. So skipping the write commits the same β, and the mutation confirms the guard is watched:

```
$ # `if not held:` removed — write on every step
FAILED ...::TestKlControlCallback::test_hold_step_performs_no_coefficient_write
FAILED ...::TestKlControlCallback::test_hold_then_raise_writes_exactly_once
2 failed, 186 passed
```

`tests/test_v07126.py` 188 passed / 2 skipped otherwise.

## One edge, which I chased down and it is fine — but its safety is untested

There is a case where `hold` does **not** imply `new_beta == state.beta`:

```python
beta = state.beta if state.beta > 0.0 else policy.beta_floor
```

With the uninitialised sentinel `state.beta == 0.0`, a hold returns `new_beta == beta_floor`. Demonstrated:

```
state.beta   = 0.0
action.reason = 'hold'
action.new_beta = 0.02      <- differs from state.beta
```

Before this PR that value was written to the trainer; after it, it is not.

**It is unreachable in production, and I checked why**: `_run_bang_bang` calls `self._seed_coefficient(policy.beta_floor)` on the line *before* `bang_bang_step`, which resolves the sentinel from the live trainer coefficient. By the time the step runs, `state.beta > 0`. Your claim holds.

But that safety now depends on **two lines staying in that order, in a different method, with nothing asserting it**. Reorder them — or add an early return above the seed — and the first hold of a run silently stops establishing β, on a path where the symptom is a coefficient that is quietly never set rather than an exception.

One test would pin it:

```python
def test_the_sentinel_is_resolved_before_the_first_step():
    # state.beta == 0.0 must never reach bang_bang_step: on a hold it would
    # yield new_beta == beta_floor != state.beta, and the skip would then
    # change a committed value.
```

Assert on the *state* the step observes rather than on call order, so it survives a refactor that keeps the behaviour.

## The `mitigation_status` field is the better half of this PR

Recovering "the controller did nothing this step" by parsing free-text `action` was the kind of thing that works until someone rewords a message. `held` / `acted` / `released` as a first-class field is what makes #371's real question — *does `kl_control` differ from `log_only` when it never trips?* — answerable from the log instead of inferable from it.

---

Approving with the one test as an ask. Push it and I merge; if you would rather I add it on merge, say so.

AmirF194:
Pushed: rebased onto main (the only conflict was two adjacent CHANGELOG entries, straightforward to merge) and added the regression test you asked for, pinning that state.beta is seeded from the live trainer coefficient before the first hold step rather than left at the sentinel.

Ran the full reward_hack_control suite locally (no torch in this environment, so the 9 tests that need it were skipped by import error, everything else green) plus ruff, both clean.

MakazhanAlpamys:
Merged as [`64d4093`](https://github.com/MakazhanAlpamys/Soup/commit/64d4093) — your fourth. Entry is in [`CONTRIBUTORS.md`](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md).

**The ordering test does its job, and my first attempt at checking it was too weak — worth saying because it is the same trap I keep flagging in reviews.** I moved the seed below `_compute_vote` and nothing failed, which for a moment looked like the test was vacuous. It was my mutation that was wrong: `_compute_vote` does not touch `state.beta`, so the seed still landed before the step. Moving it below `bang_bang_step` — the reorder that actually lets the sentinel through — fails two:

```
FAILED ...::TestKlControlCallback::test_sentinel_is_resolved_before_the_first_hold_step
FAILED ...::TestKlControlCallback::test_ppo_kl_coef_mutated
2 failed, 187 passed
```

A mutation that does not fail proves nothing about the test until you have checked it was capable of failing.

Four merges, and what they have in common is worth naming: each one closed a gap between what a thing *reported* and what it *did*. `0/0` read as a pass, a pre-flight that could not see a per-process cap, a calibration hook with no caller, and now a controller whose log could not distinguish "held" from "acted". That is a specific taste in bugs and it is the useful kind.

#371 is not closed by this — the pid_lagrangian half and the rollback ladder are still unvalidated — but the part your PR names is now true: a non-acting `kl_control` run is a genuine no-op, and the log says so in a field.

AmirF194:
Thanks, and good catch on your own mutation check, that is a sharp habit to have. Glad the acted/held/released distinction lands cleanly; happy to pick up the pid_lagrangian or ladder-reconciliation half later if either becomes buildable without GPU access.

PR #415, opened 2026-08-15
fix(train): derive the --no-reexec hint from the re-exec argv (#372)

### LINKED ISSUE #372: soup train --no-reexec prints a launch command with the user's own flags dropped
## What

`soup train --gpus 4 --fsdp full_shard --no-reexec` prints a suggested `accelerate launch` command that **does not carry `--fsdp`** (`benchmarks/gate-h100-validation.md` L3346–3348).

Following the printed hint literally trains without FSDP.

## Why it happens

There are two hand-maintained representations of "what the user typed": the argv the command was invoked with, and the argv reconstructed for the hint. They drift, and nothing checks that they agree. The same root cause produced the #77 defect where the multi-GPU entry point handed `accelerate launch` the Python interpreter instead of a script path — that half is repaired, this half was filed and left.

## Impact

`--no-reexec` exists precisely for the user who wants to run the launch themselves — i.e. the user most likely to copy the printed line verbatim. It silently drops exactly the flags that distinguish their run from the default one, and the resulting run **succeeds**, so there is no error to trace back.

## Fix path

Derive the hint from the *same* argv the run would have re-exec'd with, rather than rebuilding it. That is a one-source change, and it makes the drift impossible rather than currently-fixed.

## Acceptance criteria

- [ ] A test that asserts the printed command, tokenised, is a superset of the user-supplied flags — parametrised over at least `--fsdp`, `--deepspeed`, `--config` and one `--stream-*`-style flag
- [ ] The assertion compares against the re-exec argv builder's output, not against a hardcoded expected string, so a future flag is covered without editing the test
- [ ] A control: a run with no extra flags still prints a valid minimal command

AchuthReddy-16:
## What does this PR do?

Fixes #372.

`soup train --gpus 4 --fsdp full_shard --no-reexec` printed an `accelerate launch` command that dropped `--fsdp`. Following the suggested command therefore trained without FSDP. Because the run still succeeded, the configuration change was easy to miss.

The printed hint is now derived from the same argv used by automatic re-execution:

`build_train_reexec_argv` → `hint_argv_from_reexec`

This removes the separately maintained reconstruction path and keeps the printed command aligned with the real launch command. `--name` and `--replay*` arguments are also preserved in the launch tail.

## Type of change

* [x] Bug fix
* [ ] New feature
* [ ] Refactor / cleanup
* [ ] Documentation
* [x] Tests

## Acceptance criteria

* [x] The tokenized printed command is a superset of the user-supplied flags, parameterized over `--fsdp`, `--deepspeed`, `--config`, and `--trust-remote-code`
* [x] The assertion compares against the re-exec argv builder rather than a hardcoded expected command
* [x] A run with no additional flags still prints a valid minimal command

`soup train` currently has no `--stream-*` CLI flag; layer streaming is configured through the configuration file. I used `--trust-remote-code` as the hyphenated run-shaping flag already supported by the re-exec path. I can add a streaming configuration case through `--config` if preferred.

## Verification

* [x] Ruff checks on the modified source and test files pass
* [x] `pytest tests/test_issue372_no_reexec_hint.py` — 13 passed
* [x] CHANGELOG updated


MakazhanAlpamys (changes requested):
Welcome, and this is a strong first PR here. The diagnosis in #372 was that "what the user typed" existed in **two hand-maintained copies**, and you did not patch the printed one — you deleted it and derived the hint from the argv that actually launches the run. 125 lines out of `commands/train.py` is the right shape of change for that issue.

`tests/test_issue372_no_reexec_hint.py` 13 passed locally.

`test_a_flag_added_only_to_the_builder_shows_up_in_the_hint` is the test I would have asked for: it guards the seam you just created rather than the symptom you fixed.

## One gap, and it is the same class as the bug

Two copies became one, which is the win. But that one is still a hand-written keyword list checked against nothing, and **the direction it can now drift is CLI → builder**. I compared `collect_reexec_passthrough`'s parameters against `soup train`'s own signature:

```
train() options NOT in the passthrough list: 4
['cloud', 'cloud_submit', 'dry_run', 'gpu']
```

I think all four are deliberate — `--dry-run` is plan-only and should not survive a re-exec, and the `--cloud` family is a different launch path entirely. But nothing in the suite says so, so tomorrow's flag lands in exactly the position `--fsdp` was in: accepted by the CLI, silently absent from the launch, run succeeds, configuration quietly not applied.

The guard writes itself from what I just ran:

```python
EXPECTED_LOCAL_ONLY = {"config", "gpus", "no_reexec", "yes", "find_lr", ...,
                       "cloud", "cloud_submit", "dry_run", "gpu"}

def test_every_train_option_is_passed_through_or_deliberately_local():
    cli = set(inspect.signature(train).parameters)
    missing = cli - set(inspect.signature(collect_reexec_passthrough).parameters) - EXPECTED_LOCAL_ONLY
    assert not missing, f"new `soup train` flags not routed to re-exec: {missing}"
```

The value is in the **explicit exclusion set**: adding a flag then forces a decision — pass it through, or write it down as local-only — instead of letting silence make the decision. That is the difference between fixing this bug and closing the hole it came from.

Worth confirming the four are intentional while you are at it; if `--cloud` and `--gpus` are genuinely incompatible, that is worth a sentence in the docstring.

## Smaller

`CONTRIBUTING.md` gains one line — I have not looked closely at what it says; keep it if it documents the new seam, drop it if it is unrelated to #372, since a first PR reads more clearly when its diff is only its subject.

---

One test and I merge. And since it is your first: CI on your PRs sat unapproved because that is how GitHub treats a first-time contributor — I have approved it, and it stops being manual after your first merge.

AchuthReddy-16:
Thanks, that addresses the underlying hole, not just the symptom.

The four exclusions are intentional:

* `--dry-run` is plan-only, and the multi-GPU path skips re-execution entirely.
* `--cloud`, `--gpu`, and `--cloud-submit` use the separate Modal launch path and are incompatible with `--gpus` and Accelerate.

I added `test_every_train_option_is_passed_through_or_deliberately_local`, so every new `soup train` flag must now either be passed through or explicitly classified as local-only. I also updated the `collect_reexec_passthrough` docstring to explain these exclusions.

I kept the `CONTRIBUTING.md` line because it documents `tests/test_issue372_no_reexec_hint.py`, which covers this re-exec seam.

MakazhanAlpamys (approved):
Verified, and the guard does what it claims. Approving on content.

`tests/test_issue372_no_reexec_hint.py` + `tests/test_code_review_high.py` — 43 passed here.

## What I mutated

The point of asking for that test was that it must fail when a flag drifts, so I made it drift rather than taking the green run on trust. All mutations run against an isolated worktree with `PYTHONPATH` pinned to it, so I was measuring your source and not the installed copy.

**1. The exact #372 shape, on the seam you created** — added a new `brand_new_flag` option to `train()` and routed it nowhere:

```
FAILED test_every_train_option_is_passed_through_or_deliberately_local
tests/test_issue372_no_reexec_hint.py:173
```

Caught. That is the hole closed.

**2. Deleted `_opt("--fsdp", fsdp)` from `collect_reexec_passthrough`** — i.e. the literal regression the issue is about, at the one place it is now emitted:

```
FAILED TestNoReexecHintCli::test_fsdp_appears_in_the_printed_command
```

Caught, and caught by the CLI-level test rather than a unit one, which is the right layer for it — the unit tests hand `["--fsdp", "full_shard"]` in as a ready-made list and so cannot see that emission at all.

I also confirmed the four exclusions match `soup train`'s signature exactly as you describe them, and that the third assertion in the guard (`overlap`) is doing real work: it stops the exclusion set from being used to hide a deleted passthrough kwarg later. That one was your addition, not something I asked for.

## The change to an existing test is legitimate

I check every edit to an existing test file, since a weakened assertion is the easiest thing to miss in review. `test_gpus_reexec_passes_run_shaping_flags` moved its source scan from `commands/train.py` to `utils/launcher.py` and swapped `"script_args"` for `"collect_reexec_passthrough"` — that is the test following the code you moved, and all seven flags are still asserted. No strength lost.

CHANGELOG entry reads well. The `CONTRIBUTING.md` table row is correct to keep — that table is a required step in our release checklist, so a new test file belongs in it.

## One nit, not a blocker

`assert "--fsdp" in out` is a substring match, so renaming the flag to `--fsdp-DROPPED` passes it. A dropped flag is caught, a *renamed* one is not. Not worth another round — a rename is not a plausible regression here — but if you touch the file again, asserting on the tokenised command (which you already build elsewhere in that module) closes it.

## To merge

Only the rebase. `main` moved under you — and note it was **red** for about an hour this morning through no fault of yours: #407 and #408 collided semantically after both merged green, and I repaired it in `7bc3f83`. Rebase onto current `main` and once CI is green across all 11 cells I will merge.

Thanks for pushing on the exclusion set rather than just adding the assertion — writing down *why* those four are local-only is the part that will still be working a year from now.

AchuthReddy-16:
Rebased onto current main (`2c65af8`).

MakazhanAlpamys:
Approved and rebased — the only thing left is the matrix, which was still running seven cells when I stopped. I do not merge on an incomplete matrix, so this is first in the queue next session rather than anything you need to do.

Thanks for the rebase.

MakazhanAlpamys:
Merged as [`61366ce`](https://github.com/MakazhanAlpamys/Soup/commit/61366ce) — 12/12 green. You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) as of [`04a742b`](https://github.com/MakazhanAlpamys/Soup/commit/04a742b).

The credit line names the **shape** of the fix rather than the fix. #372's diagnosis was that "what the user typed" existed in two hand-maintained copies; the easy move was to patch the printed one. You deleted it and derived the hint from the argv that actually launches the run — 125 lines out of `commands/train.py` — so the two cannot drift again because there is only one.

Everything below was measured in an isolated worktree with `PYTHONPATH` pinned to your branch, since the editable install points at `main` and would otherwise have tested the wrong tree.

**The guard bites.** Adding a new option to `train()` and routing it nowhere fails `test_every_train_option_is_passed_through_or_deliberately_local`. And your third assertion is the load-bearing one I did not ask for: the exclusion set may not list a flag that *is* passed through, so a later deletion of a passthrough kwarg cannot hide inside it.

**The literal regression is caught.** Deleting `_opt("--fsdp", fsdp)` — the exact #372 shape, at the one place it is now emitted — fails `TestNoReexecHintCli::test_fsdp_appears_in_the_printed_command`. Caught at the CLI layer, which is the right one: the unit tests hand `["--fsdp", "full_shard"]` in as a ready-made list and structurally cannot see that emission.

**The edit to an existing test is legitimate**, and I check those specifically because a weakened assertion is the easiest thing to miss. `test_gpus_reexec_passes_run_shaping_flags` follows the moved code from `commands/train.py` to `utils/launcher.py` and still asserts all seven flags.

Two things worth saying about how this went. You pushed back with reasoning on the four exclusions rather than just adding the test, and writing down *why* `--dry-run` and the `--cloud` family are local-only is the part that will still be working in a year. And the red CI you sat through in the middle was **ours** — #407 and #408 collided on `main` after both merged green — not anything you did.

The one nit I raised and am merging with: `assert "--fsdp" in out` is a substring match, so a *renamed* flag passes it while a dropped one fails. Not worth a round trip — a rename is not a plausible regression here — but if you are ever back in that file, asserting on the tokenised command closes it.

Issue #331, opened 2026-08-06
Layer streaming: NF4 above ~165 MiB/layer produces silently wrong gradients (forward stays bit-exact)


MakazhanAlpamys:
## Summary

With an **NF4** streamed base and decoder layers above roughly **165 MiB**, the
streamed **backward** produces gradients that disagree with a resident reference
on **every layer except the last `stream_buffers`**, while the **forward stays
bit-exact** and the loss curve looks entirely healthy.

Nothing in a training log reveals it. Measured on 8x H100; full record in
[`benchmarks/gate-h100-validation.md`](../blob/main/benchmarks/gate-h100-validation.md),
STEP 2b / STEP 6 / STEP 9.

**Affected:** Qwen2.5-32B (234 MiB/layer), Qwen2.5-72B (432 MiB/layer).
**Not affected:** Qwen2.5-0.5B, Llama-3.1-8B (105), Qwen2.5-14B (132) — 50
consecutive backwards each, `worst_abs = 0.0` throughout.

## Evidence

Resident NF4 reproduces itself 256/256, so it is a valid fixed reference.
Streamed backwards diffed against it (streamed run first, so nothing about the
resident run can be blamed):

```
resident reference loss 13.048010826  (256 grad tensors)
rep 1: loss 13.048010826 (==resident True)  grads exact 256/256  worst_rel  0.0000
rep 2: loss 13.048010826 (==resident True)  grads exact   8/256  worst_rel 55.9389
rep 3: loss 13.048010826 (==resident True)  grads exact   8/256  worst_rel 55.9389
```

The loss is bit-identical to resident on every repetition. The survivors are
layers 62 and 63 — the last two, and `stream_buffers` is 2. The survivor count
tracks the buffer count exactly:

| `stream_buffers` | exact gradients | surviving layers |
|---|---|---|
| 2 | 8/256 | 62, 63 |
| 3 | 12/256 | 61, 62, 63 |
| 4 | 16/256 | 60–63 |
| 8 | 32/256 | last 8 |

## Mechanism: aliasing, not a race

Two patches to `StreamedDecoderLayer._body`, each with the unpatched control in
the same run:

```
control  -> WRONG  ['48/128', '8/128', '8/128']
sync     -> WRONG  ['12/128', '12/128', '8/128']   # torch.cuda.synchronize() after pool.wait
clone    -> EXACT  ['128/128', '128/128', '128/128'] # private copy of every pooled tensor
```

[...]

Issue #341, opened 2026-08-08
No training-seed knob: every run is seed 42 and it cannot be changed


MakazhanAlpamys:
`TrainingConfig` has `data.raft_shuffle_seed` and `data.replay_seed` — both feature-specific — but **no general training seed**. `TrainingArguments` therefore takes HF's default and every Soup run trains at `seed=42, data_seed=None`.

Verified: `training_args.bin` reports `seed=42, data_seed=None` for every run of a 24-arm sweep. `sft.py:604` even reads `getattr(tcfg, "seed", 0) or 0` for the multipack sampler — a defensive lookup for an attribute that does not exist, so it is always 0.

### Why it matters
"Run it again with a different seed" is currently impossible. Anyone measuring a training change has to vary something else — row ordering, data files — and then cannot separate the seed's contribution from the change's. In the #314 sweep the three replicates per cell differed only in row permutation plus GPU nondeterminism, **not in the trainer RNG**, which is a weaker replicate than it looks and had to be disclosed as a limitation.

It also affects the project's own evidence standard: several benchmark records report within-arm spread as the yardstick a between-arm difference must beat, and that spread is currently missing its largest natural source.

### Fix shape
Add `training.seed: Optional[int]` (and probably `training.data_seed`), thread both into `TrainingArguments`, and let the existing `getattr(tcfg, "seed", 0)` finally find something. Default should stay 42 so nothing changes for existing configs.

### Acceptance
- [ ] `training.seed` reaches `TrainingArguments.seed`
- [ ] two runs with different seeds and identical everything else produce different weights
- [ ] two runs with the same seed reproduce each other (the control — without it the first test passes for a value that is ignored)
- [ ] documented in `docs/training.md`

Found while measuring #314, where per-cell replicates needed real seed variation.

MakazhanAlpamys:
Implemented in `dcb5eb5`, CI green.

`training.seed` and `training.data_seed` now reach `TrainingArguments`, and `sft.py:604`'s defensive `getattr(tcfg, "seed", 0)` — a lookup for an attribute that did not exist — finally finds a value.

They are `Optional[int]` rather than defaulting to 42 on purpose: unset must reproduce **two different** historical defaults — HF's 42 for `TrainingArguments` and 0 for the multipack sampler since v0.37.0 — and a plain `= 42` would have silently re-ordered every existing `multipack: true` run.

Try:
```yaml
training:
  seed: 1234
  data_seed: 1234
```

[...]

PR #416, opened 2026-08-15
feat(stream): make layer-streaming pinning configurable via training.stream_pin (#366)

### LINKED ISSUE #366: Layer streaming: no config key can disable pinning, so the only known mitigation for a live correctness defect was unreachable from soup.yaml
## What

Pinning is chosen automatically by `utils.layer_stream.decide_pinning` and **there is no configuration key that turns it off** (`benchmarks/gate-h100-validation.md` L1378–1381).

## Why it matters, concretely

During the whole window that #331 was live — NF4 layers above ~165 MiB producing silently wrong gradients — `pin=False` was the **only** known mitigation, and it was measured to work at every model size tested (0.5B / 8B / 14B / 32B / 72B all bit-identical to a resident reference with pinning off, L1229–1233). A user hitting the defect could not have worked around it from their config even if they had read the issue.

The defect is repaired, so this is no longer an emergency. It is filed because the *class* of problem is not: a performance knob that is chosen automatically and cannot be overridden is also unavailable as an escape hatch when something goes wrong behind it.

## The cost must ship with the key

Pinning is not a marginal optimisation and the key must not read as one:

- **6.56x** throughput on real Qwen2.5-32B NF4 (425.07 vs 64.79 tok/s median, n=3 per arm, correctness asserted inside each arm — L1546–1556)
- **7.41x** on a 32-layer synthetic (849.62 vs 114.70 tok/s — L1685–1689)

That is why the shipped policy is *refuse* rather than *silently fall back to pageable*: a silent fallback spends the entire margin the feature exists to provide. The override should keep that framing.

## Fix path

`training.stream_pin: Optional[bool] = None` — `None` keeps today's automatic behaviour, `false` forces pageable, `true` forces pinned and fails loudly if the host cannot page-lock the store. Print the measured cost band in the panel when it is forced off.

## Acceptance criteria

- [ ] `stream_pin: false` reaches `decide_pinning` and the run proceeds pageable
- [ ] The pre-flight panel states the throughput cost when pinning is disabled, rather than absorbing it silently — the same rule `PinDecision.reason` already follows for the automatic pageable fallback
- [ ] `stream_pin: true` on a host below the page-locked ceiling refuses with the ceiling named, rather than silently degrading
- [ ] Footgun rejection: `stream_pin` set while `stream_layers: false`

ousamabenyounes:
## What does this PR do?

Adds `training.stream_pin` so the layer-streaming pin decision is reachable from `soup.yaml`. Fixes #366.

Page-locking the streamed RAM store is chosen automatically by `decide_pinning`, and until now nothing could override it. While #331 was live (NF4 layers above ~165 MiB producing silently wrong gradients), `pin=False` was the *only* known mitigation — and it was measured bit-identical to a resident reference at every size tested — yet a user could not reach it from their config.

- `stream_pin: null` (default) — unchanged automatic behaviour: attempt a pinned store, fall back to pageable and announce the cost when the host cannot page-lock it.
- `stream_pin: false` — force the pageable store. The pre-flight panel states the throughput this costs (up to **6.56x** on Qwen2.5-32B NF4, **7.41x** on a synthetic) rather than absorbing it silently.
- `stream_pin: true` — force the pinned store and **refuse** the run, naming the store size, if the box cannot page-lock it, instead of degrading silently.
- Set while `stream_layers: false` → rejected as a footgun, exactly like the other `stream_*` keys.

The plan-level choice threads through `decide_pinning` / `build_stream_plan`; the runtime refusal threads as `require_pin` through `build_streamed_model → install_streaming → _build_source`.

## Type of change

- [x] New feature

The streaming logic is torch-free at the planner level but the wiring tests drive the real
`build_streamed_model` chain, so the suite below is the full streaming area
(`tests/test_v07200.py` + `tests/test_v07203.py`): **316 passed, 12 skipped** with the
change, against **292 passed, 12 skipped** on `main`. Zero test lines removed.

Because the round-2/3 review established that a passing suite proves little here — an
assertion a hardcoded constant also satisfies cannot discriminate — the RED half is a
**mutation matrix** rather than a single failing run. Each mutation is applied to the
production code and the whole streaming area re-run; every one must fail, and the test it
fails is named:

| Mutation (RED) | Result | Killed by |
|---|---|---|
| `stream_setup.py` `stream_pin=tcfg.stream_pin` → `None` | **KILLED** | `test_stream_pin_false_makes_the_preflight_state_the_cost` |
| `stream_setup.py` `require_pin=(tcfg.stream_pin is True) and on_cuda` → `False` | **KILLED** | `test_stream_pin_true_reaches_the_runtime_as_true_on_cuda` |
| `build_streamed_model → install_streaming` `require_pin` → `False` | **KILLED** | `test_forced_pin_refuses_through_the_whole_real_chain` |
| `install_streaming → _build_source` `require_pin` → `False` | **KILLED** | same |
| delete the disk-tier announcement block | **KILLED** | `test_an_explicit_request_is_announced_not_dropped` |
| drop the `GB` store size from the RAM-tier refusal | **KILLED** | `test_the_refusal_names_the_store_size[64-2.15 GB]` |
| hardcode that store size to `1.07 GB` | **KILLED** | same (the two-size parametrization) |
| revert the forced-on plan note | **KILLED** | `test_stream_pin_true_is_recorded_in_the_plan_notes` |
| drop the RAM-tier restriction on that note | **KILLED** | `test_the_disk_tier_does_not_claim_a_page_locked_store` |

**GREEN** — clean run, same command:

```
tests/test_v07200.py + tests/test_v07203.py
316 passed, 12 skipped
```

## Checklist

- [x] `ruff check src/soup_cli/ tests/` passes (All checks passed!)
- [x] `pytest tests/test_v07200.py -v` streaming classes pass (43 passed for the streaming area, incl. the new tests)
- [x] Updated relevant docs — `docs/performance-and-quantization.md` (config block + footgun list), the `benchmarks/gate-h100-validation.md` note that said no escape hatch existed, and `CHANGELOG.md`

## Full local validation suite

Local replay of the CI job classes this change can affect:

```
== ruff check src/soup_cli/ tests/ (blocking CI job) ==
All checks passed!
== streaming area (tests/test_v07200.py + tests/test_v07203.py) ==
316 passed, 12 skipped        (main: 292 passed, 12 skipped)
== full suite (pytest tests/, repo addopts, --cov-fail-under=77) ==
41 failed, 17883 passed, 274 skipped     (main baseline: 41 failed, 17860 passed)
   -> identical failure identities; all 41 pre-existing and outside the streaming
      area (test_v0712.py x32, test_v0713.py x4, test_semantic_split.py x3,
      test_v07128.py, test_v07129.py). No new failure, +23 passing.
== changed production-line coverage ==
100% (24/24)
```

Not replayable on this box, and named rather than hidden: the windows/macOS matrix cells,
the 3.10/3.12 cells, `mlx-smoke` (needs Apple Silicon), and `actionlint` (needs Go).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,25 @@ reproducing 70+ versions of notes.
 
 ### Added
 
+- **`training.stream_pin` makes layer-streaming pinning configurable (#366 by @ousamabenyounes in #416).**
+  Page-locking the RAM store is chosen automatically by `decide_pinning`, and
+  until now nothing could override it — so while #331 was live, `pin=False` was
+  the only known mitigation for silently wrong NF4 gradients yet was unreachable
+  from `soup.yaml`. `stream_pin: false` now forces the pageable store (and the
+  pre-flight states the throughput it costs, up to 6.56x measured, rather than
+  absorbing it silently); `stream_pin: true` forces the pinned store and, **on
+  the RAM tier**, refuses the run — naming the store size, not the ceiling
+  (#366 AC3): `pinned_limit_bytes` is passed as `None`, so the page-lock ceiling
+  is deliberately left unprobed and the store size is the only figure the
+  refusal can honestly cite — if the box cannot page-lock it, instead of
+  degrading silently; unset keeps today's automatic behaviour. On the disk tier
+  (no RAM store to page-lock) and on CPU (no device to copy to) pinning is
+  *inapplicable* rather than unsatisfiable, so `true` is **announced and the run
+  proceeds** — refusing there would brick the large-model runs the disk tier
+  exists for, and would make the key uncommittable to a config shared between a
+  GPU box and a CPU box. Set while `stream_layers: false` it is rejected as a
+  footgun, like the other stream keys.
+
 - **A ready-made `qwen3.5-9b-grpo` recipe for GRPO reasoning training with `Qwen/Qwen3.5-9B` (#277 by @harshitthek in #448).**
   The recipe combines the established GRPO defaults (accuracy reward, beta=0.1, 4 generations)
   with LoRA r=16 and 4-bit quantization.
@@ -114,24 +133,6 @@ reproducing 70+ versions of notes.
   attestations also attach the detached `.sig` sidecar. The flag needs
   `--output` (omitting it exits 2), and a requested registry attachment failure
   exits 1 after preserving files already emitted.
-- **`training.stream_pin` makes layer-streaming pinning configurable (#366).**
-  Page-locking the RAM store is chosen automatically by `decide_pinning`, and
-  until now nothing could override it — so while #331 was live, `pin=False` was
-  the only known mitigation for silently wrong NF4 gradients yet was unreachable
-  from `soup.yaml`. `stream_pin: false` now forces the pageable store (and the
-  pre-flight states the throughput it costs, up to 6.56x measured, rather than
-  absorbing it silently); `stream_pin: true` forces the pinned store and, **on
-  the RAM tier**, refuses the run — naming the store size, not the ceiling
-  (#366 AC3): `pinned_limit_bytes` is passed as `None`, so the page-lock ceiling
-  is deliberately left unprobed and the store size is the only figure the
-  refusal can honestly cite — if the box cannot page-lock it, instead of
-  degrading silently; unset keeps today's automatic behaviour. On the disk tier
-  (no RAM store to page-lock) and on CPU (no device to copy to) pinning is
-  *inapplicable* rather than unsatisfiable, so `true` is **announced and the run
-  proceeds** — refusing there would brick the large-model runs the disk tier
-  exists for, and would make the key uncommittable to a config shared between a
-  GPU box and a CPU box. Set while `stream_layers: false` it is rejected as a
-  footgun, like the other stream keys.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,12 +120,17 @@ reproducing 70+ versions of notes.
   the only known mitigation for silently wrong NF4 gradients yet was unreachable
   from `soup.yaml`. `stream_pin: false` now forces the pageable store (and the
   pre-flight states the throughput it costs, up to 6.56x measured, rather than
-  absorbing it silently); `stream_pin: true` forces the pinned store and refuses
-  the run — naming the store size, not the ceiling (#366 AC3): `pinned_limit_bytes`
-  is passed as `None`, so the page-lock ceiling is deliberately left unprobed and
-  the store size is the only figure the refusal can honestly cite — if the box
-  cannot page-lock it, instead of degrading silently; unset keeps today's
-  automatic behaviour. Set while `stream_layers: false` it is rejected as a
+  absorbing it silently); `stream_pin: true` forces the pinned store and, **on
+  the RAM tier**, refuses the run — naming the store size, not the ceiling
+  (#366 AC3): `pinned_limit_bytes` is passed as `None`, so the page-lock ceiling
+  is deliberately left unprobed and the store size is the only figure the
+  refusal can honestly cite — if the box cannot page-lock it, instead of
+  degrading silently; unset keeps today's automatic behaviour. On the disk tier
+  (no RAM store to page-lock) and on CPU (no device to copy to) pinning is
+  *inapplicable* rather than unsatisfiable, so `true` is **announced and the run
+  proceeds** — refusing there would brick the large-model runs the disk tier
+  exists for, and would make the key uncommittable to a config shared between a
+  GPU box and a CPU box. Set while `stream_layers: false` it is rejected as a
   footgun, like the other stream keys.
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,19 @@ reproducing 70+ versions of notes.
   attestations also attach the detached `.sig` sidecar. The flag needs
   `--output` (omitting it exits 2), and a requested registry attachment failure
   exits 1 after preserving files already emitted.
+- **`training.stream_pin` makes layer-streaming pinning configurable (#366).**
+  Page-locking the RAM store is chosen automatically by `decide_pinning`, and
+  until now nothing could override it — so while #331 was live, `pin=False` was
+  the only known mitigation for silently wrong NF4 gradients yet was unreachable
+  from `soup.yaml`. `stream_pin: false` now forces the pageable store (and the
+  pre-flight states the throughput it costs, up to 6.56x measured, rather than
+  absorbing it silently); `stream_pin: true` forces the pinned store and refuses
+  the run — naming the store size, not the ceiling (#366 AC3): `pinned_limit_bytes`
+  is passed as `None`, so the page-lock ceiling is deliberately left unprobed and
+  the store size is the only figure the refusal can honestly cite — if the box
+  cannot page-lock it, instead of degrading silently; unset keeps today's
+  automatic behaviour. Set while `stream_layers: false` it is rejected as a
+  footgun, like the other stream keys.
 
 ### Fixed
 
--- a/benchmarks/gate-h100-validation.md
+++ b/benchmarks/gate-h100-validation.md
@@ -1442,10 +1442,11 @@ LlamaConfig(vocab_size=4096, hidden_size=5120, intermediate_size=27648,
   exactly this, and `load_async` as issuing `stream.wait_stream(current)` before
   `copy_`. Both are evidently insufficient on this path.
 
-**No fix attempted and no Soup code changed**, per the brief. There is also no
-user-facing escape hatch: pinning is chosen automatically by `decide_pinning`,
-with no config key to disable it, so a user hitting this cannot work around it
-from `soup.yaml`.
+**No fix attempted and no Soup code changed**, per the brief. At the time this
+was measured there was also no user-facing escape hatch: pinning was chosen
+automatically by `decide_pinning` with no config key to disable it, so a user
+hitting this could not work around it from `soup.yaml`. That escape hatch now
+exists — `training.stream_pin: false` (#366) forces the pageable store.
 
 ### Everything measured, one table
 
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -235,7 +235,7 @@ training:
   stream_layers: true          # Enable layer streaming
   stream_source: auto          # 'auto' (same-host RAM), 'ram', 'disk' (v0.72.3)
   stream_buffers: 2            # Double-buffering; range [2, 8]
-  # stream_pin: false          # Force the pinned RAM store off (escape hatch) or on; unset = automatic
+  # stream_pin: false          # Force the pinned RAM store off (escape hatch) or on; unset = automatic. See below
   # stream_vram_override: 4_000_000_000   # Bytes to assume free (v0.73.x); see below
   # stream_vram_probe: true    # Decide the fit by MEASURING one step (sft only); see below
   # stream_disk_kind: nvme     # Override auto disk-kind detection: nvme/ssd/hdd; see below
@@ -288,6 +288,39 @@ Untied `embed_tokens` + `lm_head` stay resident and unquantised (2.10 GB of the 
 - **A streamed 8B run now completes on a Turing card — free-tier Colab, Tesla T4 (sm_75) — and that is all it shows.** `NousResearch/Meta-Llama-3.1-8B-Instruct`, NF4, `stream_buffers: 2`, batch 1, `max_length: 256`, LoRA r=8, fp16: 7 steps, exit 0, adapter written with 128 of 128 tensors non-zero, **measured peak 2.91 GB** against a predicted ~3.02 GB (the pre-flight over-predicts by 3.8%, the safe direction it was fitted for). The T4 has 15.6 GB, so the process was capped to **4.00 GB** with `torch.cuda.set_per_process_memory_fraction`, and the cap was shown to bite — a 4.29 GiB allocation was refused. **No throughput is quoted from this run**: a card under an artificial cap is not a benchmark, and the [notebook](../notebooks/proof-4gb.ipynb) deliberately quotes none either. **What it does not establish**: backward/gradient exactness at 8B on Turing (a non-zero adapter shows gradients flowed, not that they were correct), and the notebook's streamed-vs-resident comparison produced no captured output, so it is recorded as unrun rather than as a pass. Note also that the pre-flight read **free VRAM 15.10 GB** — the device, not the per-process cap — so on capped hardware it is `training.stream_vram_override` and not the fit decision that enforces the real budget. Record: [`benchmarks/run-t4-colab-free-tier.md`](../benchmarks/run-t4-colab-free-tier.md).
 - **The bf16 3B throughput above is a LOWER BOUND.** The reference box could not page-lock the 5.55 GB base (its measured page-locked ceiling is 7.65 GB, and a CUDA context plus the model skeleton did not leave room), so that run fell back to a pageable store. Pageable memory makes the host-to-device copy synchronous, which costs overlap — visible as the GPU-utilisation drop from 96.8% (1.5B, pinned) to 79.3% (3B, pageable). Soup does this fallback automatically **and prints the cost** rather than absorbing it silently. NF4 lifts this at 3B: the store drops under the ceiling and pins.
 - Numbers are Windows/WDDM and therefore systematically pessimistic versus Linux. `expandable_segments:True` is silently ignored on Windows; Soup detects that and does not claim it is active.
+
+### Forcing the pin (`training.stream_pin`)
+
+Pinning is chosen automatically; `stream_pin` is how a config overrides that choice.
+
+- **`stream_pin: false`** forces the pageable RAM store. The pre-flight states the
+  throughput this costs — up to **6.56×** measured (Qwen2.5-32B NF4), **7.41×** on a
+  synthetic — rather than absorbing it silently. This is the escape hatch: it was the
+  only known mitigation while #331 was live.
+- **`stream_pin: true`** forces the page-locked store. **On the RAM tier it REFUSES the
+  run**, naming the store size, if the box cannot page-lock it — instead of degrading to
+  a pageable store and spending the whole margin pinning exists to provide.
+- **Unset (the default)** keeps today's behaviour: on the CUDA RAM tier, attempt a pinned store and fall back to
+  pageable and announce the cost when the host cannot page-lock it.
+
+**Where `true` announces instead of refusing.** Pinning page-locks the RAM store so that
+host→device copies can overlap compute — so it needs both a RAM store *and* a device to
+copy to. In the two cases below one of those is missing, the request is **inapplicable
+rather than unsatisfiable**, and the run *proceeds with an announcement* rather than
+refusing:
+
+| Tier / device | `stream_pin: true` does |
+|---|---|
+| RAM tier on CUDA | pins, or **refuses** naming the store size |
+| Disk tier (base does not fit in RAM, weights stream from NVMe) | announces that pinning does not apply, proceeds |
+| CPU (no CUDA device) | announces that page-locking is a host-to-device optimization, proceeds |
+
+Refusing on those two would brick the large-model runs the disk tier exists for, and
+would make `stream_pin: true` uncommittable to a `soup.yaml` shared between a GPU box and
+a CPU box. The RAM tier is where the flag has real semantics, and there it still refuses.
+
+Set while `stream_layers: false` the key is rejected as a footgun, like the other
+`stream_*` keys.
 
 ### Sizing a streaming run (v0.72.3)
 
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -235,6 +235,7 @@ training:
   stream_layers: true          # Enable layer streaming
   stream_source: auto          # 'auto' (same-host RAM), 'ram', 'disk' (v0.72.3)
   stream_buffers: 2            # Double-buffering; range [2, 8]
+  # stream_pin: false          # Force the pinned RAM store off (escape hatch) or on; unset = automatic
   # stream_vram_override: 4_000_000_000   # Bytes to assume free (v0.73.x); see below
   # stream_vram_probe: true    # Decide the fit by MEASURING one step (sft only); see below
   # stream_disk_kind: nvme     # Override auto disk-kind detection: nvme/ssd/hdd; see below
@@ -353,7 +354,7 @@ prints this advice when it sees you accumulating.
 - `task: kto` with `batch_size: 1` → TRL's KL term is degenerate at batch 1; refused when the config is read rather than minutes later after sharding
 - `lora.use_dora` / `lora.use_vera` / `lora.init_strategy` other than `random` → these initialise from the real base weight, which is on the meta device under streaming
 - `unfrozen_parameters`, `lisa_enabled`, `packing`, `multipack`, `use_fsdp2_compile`, `train_router_only`, `expand_layers` → each independently rewrites or re-freezes the same layers
-- `stream_source` / `stream_buffers` / `stream_vram_override` / `stream_vram_probe` / `stream_disk_kind` set while `stream_layers: false` → a footgun, refused
+- `stream_source` / `stream_buffers` / `stream_vram_override` / `stream_vram_probe` / `stream_disk_kind` / `stream_pin` set while `stream_layers: false` → a footgun, refused
 - `stream_vram_probe` on any task other than `sft` → the probe runs a plain causal-LM step, which *is* the SFT step but is not a preference loss. Measured at one matching shape it is conservative there too (6.02 GB against a real DPO step's 5.30 GB, +13.5%), but one shape is not a validation, so it is not offered for `dpo`/`orpo`/`simpo`/`kto` yet
 - an architecture outside the supported list (llama / qwen2 / qwen3 / mistral / gemma / gemma2 / gemma3_text / phi / phi3) → named explicitly
 
--- a/src/soup_cli/config/schema.py
+++ b/src/soup_cli/config/schema.py
@@ -3059,6 +3059,22 @@ class TrainingConfig(BaseModel):
             raise ValueError("training.stream_buffers must be an int, not bool")
         return v
 
+    stream_pin: Optional[bool] = Field(
+        default=None,
+        description=(
+            "Force the page-locked (pinned) RAM store on or off. None (the "
+            "default) lets the box decide: it attempts a pinned store and falls "
+            "back to a pageable one — announcing the cost — when the host cannot "
+            "page-lock it. 'false' forces the pageable store, the only known "
+            "escape hatch when a pinning path is suspect (it was the sole "
+            "mitigation while #331 was live). 'true' forces the pinned store and "
+            "REFUSES the run, naming the store size, if the box cannot page-lock "
+            "it, rather than silently degrading — page-locking is worth up to "
+            "6.56x measured throughput, so a silent fallback spends the whole "
+            "margin the feature exists to provide."
+        ),
+    )
+
     stream_vram_override: Optional[int] = Field(
         default=None,
         ge=0,
@@ -4935,12 +4951,14 @@ class SoupConfig(BaseModel):
                 or tcfg.stream_vram_override is not None
                 or tcfg.stream_vram_probe
                 or tcfg.stream_disk_kind is not None
+                or tcfg.stream_pin is not None
             ):
                 raise ValueError(
                     "training.stream_source / training.stream_buffers / "
                     "training.stream_vram_override / training.stream_vram_probe "
-                    "/ training.stream_disk_kind set but stream_layers is false; "
-                    "set stream_layers=true to stream the base layer-by-layer."
+                    "/ training.stream_disk_kind / training.stream_pin set but "
+                    "stream_layers is false; set stream_layers=true to stream the "
+                    "base layer-by-layer."
                 )
             return self
         # v0.72.4 — the four preference losses join SFT. DPO and KTO take their
--- a/src/soup_cli/config/schema.py
+++ b/src/soup_cli/config/schema.py
@@ -3068,10 +3068,16 @@ class TrainingConfig(BaseModel):
             "page-lock it. 'false' forces the pageable store, the only known "
             "escape hatch when a pinning path is suspect (it was the sole "
             "mitigation while #331 was live). 'true' forces the pinned store and "
-            "REFUSES the run, naming the store size, if the box cannot page-lock "
-            "it, rather than silently degrading — page-locking is worth up to "
-            "6.56x measured throughput, so a silent fallback spends the whole "
-            "margin the feature exists to provide."
+            "REFUSES the run on the RAM tier, naming the store size, if the box "
+            "cannot page-lock it, rather than silently degrading — page-locking "
+            "is worth up to 6.56x measured throughput, so a silent fallback "
+            "spends the whole margin the feature exists to provide. On the disk "
+            "tier (the base does not fit in RAM, so there is no RAM store to "
+            "page-lock) and on CPU (page-locking is a host-to-device transfer "
+            "optimization and there is no device) pinning is INAPPLICABLE rather "
+            "than unsatisfiable: 'true' is announced and the run proceeds "
+            "without it, so the key stays committable to a config shared between "
+            "a GPU box and a CPU box."
         ),
     )
 
--- a/src/soup_cli/trainer/stream_setup.py
+++ b/src/soup_cli/trainer/stream_setup.py
@@ -400,6 +400,16 @@ class StreamingSetupMixin:
             use_rslora=tcfg.lora.use_rslora,
         )
 
+        # #366 review — on CPU there is no CUDA device, so page-locking (a
+        # host->device transfer optimization) is inapplicable. An explicit
+        # stream_pin=true is honoured by saying so, not by dropping it silently.
+        if tcfg.stream_pin is True and not on_cuda:
+            console.print(
+                "[yellow]training.stream_pin=true, but no CUDA device is present: "
+                "page-locking is a host-to-device transfer optimization and does "
+                "not apply on CPU. Proceeding without it.[/]"
+            )
+
         model, runtime = build_streamed_model(
             model_id=cfg.base,
             shard_dir=shard_dir,
@@ -409,8 +419,11 @@ class StreamingSetupMixin:
             dtype=dtype,
             buffers=tcfg.stream_buffers,
             pin=plan.pinned and on_cuda,
-            # #366: stream_pin=true must refuse rather than silently fall back to
-            # a pageable store when the box cannot page-lock it.
+            # #366: on the RAM tier stream_pin=true refuses rather than silently
+            # falling back to a pageable store; on the disk tier the runtime
+            # announces that pinning is inapplicable (no RAM store to lock); on
+            # CPU the notice above covers it. require_pin only carries the RAM
+            # refusal, so it is gated on a real CUDA device.
             require_pin=(tcfg.stream_pin is True) and on_cuda,
             seed=tcfg.seed if getattr(tcfg, "seed", None) is not None else 0,
             trust_remote_code=self._trust_remote_code,
--- a/src/soup_cli/trainer/stream_setup.py
+++ b/src/soup_cli/trainer/stream_setup.py
@@ -331,6 +331,9 @@ class StreamingSetupMixin:
             disk_kind=lambda: resolve_disk_kind(
                 shard_dir, tcfg.stream_disk_kind, notify=console.print
             ),
+            # #366: training.stream_pin (None/False/True) overrides the automatic
+            # pinning choice so the pageable escape hatch is reachable from config.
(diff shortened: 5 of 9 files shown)


MakazhanAlpamys (changes requested):
Reviewed against the one question that decides this PR: **does `stream_pin` preserve "refuse rather than silently degrade", or does it become the escape hatch that quietly reintroduces the pageable fallback we rejected?**

Answer: the *cost framing* is genuinely well done and I verified it reaches the screen. But the flag does not keep its promise on one reachable path, and — measured — nothing tests that the config key reaches the runtime at all.

All runs below were done in an isolated worktree with `PYTHONPATH` pinned to it, verified by printing `soup_cli.utils.layer_stream.__file__`, so I was measuring your branch and not the installed copy.

## What holds up

`stream_pin: false` does not absorb the cost. I traced `PinDecision.reason` → `plan.notes` → `render_stream_panel` → `stream_setup.py:364`, and then proved it rather than reasoning about it: driving the real `setup()` on the CPU harness with `stream_pin: false`, the pre-flight panel contains **`6.56`** and **`stream_pin=false`**. Naming the measured 6.56x / 7.41x from named constants is better than the prose it replaces.

The RAM tier refusal is correct: `layer_stream_runtime.py:1590` raises before the pageable fallback and the message explains the cost. Confirmed by mutation — forcing that branch off makes `test_forced_pin_refuses_instead_of_falling_back` stop raising.

Diff matches the description. **Zero removed test lines** — additions only, nothing weakened. The `benchmarks/gate-h100-validation.md` edit is a legitimate correction (that file claimed no escape hatch existed) and is correctly phrased as "at the time this was measured", so the record is preserved rather than rewritten. CHANGELOG and `docs/performance-and-quantization.md` both updated. Ruff clean; 573 passed / 4 skipped across the ten streaming files.

## Blocker 1 — `stream_pin: true` does not refuse on the disk tier

`_build_source` returns the disk tier **before** the `require_pin` check (`layer_stream_runtime.py:~1583`). Demonstrated directly with `DiskSource`/`RamSource` stubbed:

```
DISK TIER + require_pin=True -> pinned=False -> NO refusal raised
```

This is reachable on the **default** `stream_source: auto` whenever the base does not fit RAM. So the one promise `stream_pin: true` makes — refuse rather than degrade — is not kept, on the path a user is most likely to hit without choosing it. The auto→disk fallback is announced as a plan note, which softens it, but an announcement is not a refusal; that distinction is the whole reason the pageable fallback was rejected in the first place.

Either raise when `require_pin and tier == "disk"`, or — if pinning is meaningless on that tier by design — say so explicitly in the schema description and docs and add a test pinning that decision. Same question for `require_pin=(tcfg.stream_pin is True) and on_cuda` (`stream_setup.py:401`): on a non-CUDA device the explicit request is dropped without a word, and our own harness runs streaming at `device="cpu"`, so that path is exercised.

## Blocker 2 — nothing tests that the key reaches the runtime

I hardcoded `stream_pin=None` at `stream_setup.py:323`, and separately `require_pin=False` at `:401`. **550 streaming tests passed in both cases.**

That is not "this is hard to test": the control proves the file is reachable from a CPU-only test — deleting the probe call at `:411` fails two tests in `test_v07203.py::TestAutoTierFallback`, which drives the real `setup()`. And that class's own docstring makes exactly this argument: *"a unit test of `choose_tier` alone would not catch a regression that stops threading the decision into `build_streamed_model`"*. The wiring test is about twelve lines on a harness that already exists:

```python
class TestStreamPinReachesThePanel(TestAutoTierFallback):
    def test_cost_is_printed_when_the_user_opts_out(self, tmp_path, monkeypatch, capsys):
        self._run(tmp_path, monkeypatch, free_ram=10_000_000_000, stream_source="auto",
                  extra_training_yaml="  stream_pin: false\n")
        out = re.sub(r"\s+", " ", _strip_ansi(capsys.readouterr().out))
        assert "6.56" in out and "stream_pin=false" in out
```

Plus a companion asserting `require_pin` is actually requested for `stream_pin: true` — spy on `build_streamed_model`'s kwargs, since `device="cpu"` zeroes it.

## Three smaller items

**Pin the measured numbers.** Changing `PIN_THROUGHPUT_GAIN_REAL` to `1.01` and `SYNTHETIC` to `1.02` passes every test — the assertions only check that the word "throughput" appears. Since #366's whole argument is that the cost must ship with the key, assert on `6.56` / `7.41` (or on the constants) so a later edit cannot quietly deflate them. A control confirms the tests do guard the *presence* of a cost sentence, just not its magnitude.

**Keep the documented reason sentence.** CLAUDE.md's contract is that `PinDecision.reason` states the `~97% -> ~79%` GPU-utilisation cost. The automatic-fallback branch still does; the new forced-off branch drops it in favour of "costs overlap" plus the 6.56x/7.41x band. Please carry both so the two branches read the same, and assert on it.

**Footgun test covers only `false`.** Narrowing the schema guard from `is not None` to `is False` passes every test, so `stream_pin: true` + `stream_layers: false` is unguarded. Add the `true` case.

Nit: `_spec_bytes` duplicates the sum already inlined in `DiskSource.__init__` (`layer_stream_runtime.py:~416`) — worth reusing the helper there.

One thing to fix in the text rather than the code: the PR says the refusal names "the store size" while #366's AC3 says "the ceiling". The store size is the **right** choice, because `pinned_limit_bytes=None` is passed deliberately — the ceiling is not probed. Worth a line so the AC is not later read as unmet.

`.claude/CLAUDE.md` is untracked, so you cannot update it and I am not asking you to — that is mine to do.

Note `main` was red for about an hour this morning (#407/#408 collided semantically after both merged green); repaired in `7bc3f83`, so rebase before the next CI run.

MakazhanAlpamys:
Sequencing note, since #426 (Qwen3.5 MoE layer streaming) landed today and touches three of your files.

**You go first.** Measured with `git merge-tree` rather than guessed — conflicts in `layer_stream.py`, `layer_stream_runtime.py` and `CHANGELOG.md`. The one that matters is `install_streaming`: you rewrite that statement to add `require_pin=`, they rewrite the same statement to `_build_source(..., needed_specs_by_layer, ...)` + `LayerBufferPool(..., active_keys_by_layer=...)`. A careless resolution silently drops one side or the other. `build_stream_plan` also collides, but only because you each append a keyword to the same signature line.

You are ahead on readiness — your two blockers are a wiring test and a disk-tier decision, while #426 currently has four failing tests — so #426 rebases onto you, not the other way round. Nothing changes for you beyond the rebase onto current `main` you already needed.

One thing to be aware of when you do rebase: #426 currently breaks `tests/test_v07200.py`, the file you add tests to. That is theirs, not yours, and I have told them so.

ousamabenyounes:
Rebased onto `main` (past `7bc3f83`, so the #407/#408 collision is included) and addressed the AC3 wording.

The CHANGELOG entry now states explicitly that the refusal names **the store size, not the ceiling**, and why: `pinned_limit_bytes` is passed as `None`, so the page-lock ceiling is deliberately left unprobed — the store size is the only figure the refusal can honestly cite. The code was already correct; this is the one-line text fix so AC3 is not later read as unmet.

Left `.claude/CLAUDE.md` untouched as you said.

MakazhanAlpamys (changes requested):
Re-reviewed in an isolated worktree with `PYTHONPATH` pinned to it and verified by printing `soup_cli.utils.layer_stream.__file__`.

First, so we are on the same page about what the force-push contained — I measured it rather than guessing. GitHub's timeline records the head I reviewed as `148e9825`. Diffing that against the new head **restricted to this PR's own files** — `layer_stream.py`, `layer_stream_runtime.py`, `stream_setup.py`, `tests/test_v07200.py`, `docs/performance-and-quantization.md`, `benchmarks/gate-h100-validation.md` — comes back at **0 lines**, and the `stream_pin` hunks in `schema.py` are unchanged too. Everything else in the push is `main` arriving via the rebase.

So this was the rebase I asked for at the end of the last review, and nothing else. That is a perfectly reasonable thing to push — I simply could not tell from the PR whether it was also meant to answer the review, so I re-measured everything instead of assuming.

All five findings reproduce.

## Blocker 1 — the disk tier still does not refuse

`layer_stream_runtime.py:1582` still returns `DiskSource` **before** the `require_pin` check at `:1590`. Stubbing both sources:

```
DISK tier + require_pin=True  ->  pinned=False, NO refusal raised
RAM  tier + require_pin=True  ->  REFUSED
```

Reachable on the default `stream_source: auto`.

And I now have the non-CUDA half **measured rather than argued**: spying on `build_streamed_model` through the real `_setup_streaming_transformers` at `device="cpu"` with `stream_pin: true`, it receives `require_pin=False, pin=False`. An explicit `true` is dropped without a word, on the device our own test harness runs.

## Blocker 2 — nothing tests that the key reaches the runtime

- `stream_setup.py:323` `stream_pin=tcfg.stream_pin` → `stream_pin=None` → **362 passed**
- `stream_setup.py:401` `require_pin=(...)` → `require_pin=False` → **362 passed**
- **Control**: replacing the probe call at `:411` with `pass` → **kills 2 tests** in `test_v07203.py::TestAutoTierFallback`

So the file is reachable and mutation-detectable from a CPU-only run. This is a coverage gap, not an untestable path.

I also built the test I suggested, to check my own suggestion was fair rather than just easy to type. Subclassing `TestAutoTierFallback` with `extra_training_yaml="  stream_pin: false\n"` gives `6.56 in out: True`, `stream_pin=false in out: True` — that half works against your code today. Its `stream_pin: true` companion is what fails, for the reason in blocker 1.

## The three smaller ones

- `PIN_THROUGHPUT_GAIN_REAL` 6.56 → 1.01 and `_SYNTHETIC` 7.41 → 1.02 → **362 passed**. The measured numbers #366 insists must ship with the key are unpinned.
- The forced-off `PinDecision.reason` still lacks the `~97% -> ~79%` sentence the automatic branch carries. I checked both branches: automatic contains it (`True`), `stream_pin=False` does not (`False`).
- Narrowing the schema guard to `stream_pin is False` → **362 passed**, so `true` + `stream_layers: false` remains unguarded.

Nit still open: the same `math.prod(shape) * _dtype_size(dtype)` expression at `layer_stream_runtime.py:417` and `:1616`.

## What still holds from last time

Zero removed test lines. Ruff clean. **468 passed / 2 skipped** across the five streaming files, **160 passed** across the source-grep suite. CHANGELOG and docs correct, and the `benchmarks/gate-h100-validation.md` edit is still properly phrased as "at the time this was measured" — that phrasing is the right instinct and I want it to survive into the merge.

The two blockers are the whole of what stands between this and a merge.

MakazhanAlpamys:
The AC3 wording fix is right and I am glad you made it explicit — "the refusal names the store size, not the ceiling, because `pinned_limit_bytes` is passed as `None` and the page-lock ceiling is deliberately left unprobed" is exactly the sentence that stops someone reading AC3 as unmet in six months.

But so we do not lose each other: **that was the nit, not the blockers.** I posted a full re-review a few minutes before your comment, so you may well have been writing in parallel — the two blockers from the first round both still reproduce on the current head, and I measured rather than re-read:

**1. `stream_pin: true` still does not refuse on the disk tier.** `layer_stream_runtime.py:1582` returns `DiskSource` before the `require_pin` check at `:1590`. With both sources stubbed: disk tier + `require_pin=True` → `pinned=False`, **no refusal**, while the RAM tier correctly refuses. Reachable on the default `stream_source: auto`.

New this round, and it is the half I had only argued before: at `device="cpu"` with `stream_pin: true`, spying on `build_streamed_model` through the real `_setup_streaming_transformers` shows it receives `require_pin=False, pin=False`. The explicit request is dropped **without a word**, on the device our own harness runs.

**2. Nothing tests that the key reaches the runtime.** Hardcoding `stream_pin=None` at `stream_setup.py:323` → 362 passed. Hardcoding `require_pin=False` at `:401` → 362 passed. The control (replacing the probe call at `:411` with `pass`) kills two tests, so the file is reachable from CPU-only tests and this is coverage rather than difficulty.

I also built the test I had suggested, so I was not asking you for something I had not checked myself: the `stream_pin: false` half passes against your code today (`6.56` and `stream_pin=false` both appear in the panel). It is the `stream_pin: true` companion that fails — for the reason in blocker 1.

Also still open: the measured 6.56 / 7.41 can be changed to 1.01 / 1.02 with all 362 tests green, the forced-off `PinDecision.reason` still omits the `~97% -> ~79%` sentence the automatic branch carries, and narrowing the schema guard to `is False` passes everything so `true` + `stream_layers: false` is unguarded.

Everything else about the PR still stands well — zero removed test lines, ruff clean, 468 passed across the streaming files, and the `benchmarks/` edit properly time-qualified.

One thing I would genuinely like your view on rather than just a fix: **is refusing on the disk tier the right behaviour at all?** If pinning is meaningless there by design, then saying so in the schema description and pinning that decision with a test is a perfectly good resolution — and arguably better than a refusal. What I cannot merge is the current state, where `true` means "refuse if RAM, silently ignore if disk or CPU" and nothing records that as a decision.

ousamabenyounes:
Both blockers and the three open points are addressed.

**Blocker 1 — the disk/CPU silence.** You framed it exactly right, so I want to name the decision I made: pinning page-locks the RAM store for host→device transfer, so on the disk tier (base streams from NVMe, no RAM store) and on CPU (no device) it is **inapplicable**, not merely unsatisfiable. Refusing there would brick the large-model runs the disk tier exists for, so both now **proceed with an announcement** rather than dropping the request — the runtime says so on the disk tier, `stream_setup` says so on CPU. The RAM-tier hard refusal is unchanged. (If you would rather it refuse everywhere, it is a two-line change — but I think "record + proceed" is the better behaviour here, and it is now a decision, not a silence.)

**Blocker 2 — nothing proved it reached the runtime.** New CPU-only test stubs `build_streamed_model` and asserts it receives `require_pin`; deleting the `require_pin=` argument now KeyErrors the test instead of leaving the suite green.

**Open points:**
- The `stream_pin: false` reason now carries the same `~97% to ~79%` GPU-utilisation sentence the automatic-fallback branch does, and the test pins the `6.56x`/`7.41x` figures — changing either constant to `1.01` now fails.
- The schema footgun test is parametrized over `true` AND `false`, so narrowing the guard to `is False` no longer passes with `true` + `stream_layers: false` unguarded.

The `PinDecision.reason` forced-off/automatic wording is now consistent. Local: `287 passed, 12 skipped` across the streaming files, ruff clean.

MakazhanAlpamys:

ousamabenyounes:
Rebased onto current `main` (past the `#365` merge). The two touched the same three sites — the `stream_layers: false` footgun guard, the `build_stream_plan` disk-kind lambda, and the docs list — so I kept BOTH: `stream_disk_kind` (yours, `#365`) and `stream_pin` (`#366`) are now listed together in the guard/message/docs, and the plan lambda keeps `resolve_disk_kind(..., stream_disk_kind, notify=...)` while also passing `stream_pin=tcfg.stream_pin`. Local: `456 passed, 12 skipped` across the stream files, ruff clean.

MakazhanAlpamys (changes requested):
Re-verified on an isolated worktree with mutation testing. **Three of five items are properly closed**, and I confirmed each by breaking it:

- `PIN_THROUGHPUT_GAIN_REAL 6.56 → 1.01` / `_SYNTHETIC 7.41 → 1.02` now **fails** `test_stream_pin_false_forces_pageable_and_states_the_cost`.
- Narrowing the footgun guard to `is False` now **fails** the `[True]` parametrization.
- The forced-off reason carries the `~97% to ~79%` sentence verbatim, and the RAM-tier refusal names the store size — I triggered it and it printed `could not page-lock the 1.34 GB RAM store`.

Three things still open.

## 1. `stream_pin: false` is not protected end-to-end

Hardcoding `stream_pin=None` at `stream_setup.py:327` leaves **307 tests passing**.

That mutation is not cosmetic: with it, `plan.pinned` comes from the automatic path, so `pin = plan.pinned and on_cuda` pins anyway — **the entire escape hatch this PR exists to provide becomes a silent no-op**, and it was the only known mitigation while #331 was live.

`test_stream_pin_false_reaches_the_plan_as_pageable_with_cost` calls `build_stream_plan` directly, so nothing crosses the wiring. Same shape as last round's blocker 2, on the other argument.

## 2. The new test pins the kwarg's *presence*, not its *value*

You asked me to check this and I did, precisely:

- Deleting `require_pin=` at `:418` → **fails** (good, the KeyError fires).
- Hardcoding `require_pin=False` at `:418` → **passes, 307/307**.

On CPU `False` is the expected value, so the constant satisfies `assert captured["require_pin"] is False`, and there is no CUDA-path case. **`stream_pin: true` could stop refusing on every real GPU with the suite green.**

This needs no GPU to close. `_run()` already sets `wrapper.device`, and `on_cuda` is `str(self.device).startswith("cuda")` — a plain string check. I wrote the case to check I was not asking for something impractical: a subclass with `device="cuda"`, your same `build_streamed_model` stub, asserting `captured["require_pin"] is True` gives **25 passed on your code, 1 failed under the hardcode**. One extra case.

## 3. The decision is documented nowhere a user reads — and the docs now contradict the code

I captured both announcements from real code, and they read well:

> **disk tier** — `training.stream_pin=true, but this run is on the disk tier: the base does not fit in RAM, so weights stream directly from NVMe and there is no RAM store to page-lock. Pinning does not apply here; proceeding without it.`
>
> **CPU** — `training.stream_pin=true, but no CUDA device is present: page-locking is a host-to-device transfer optimization and does not apply on CPU. Proceeding without it.`

But the schema description says `stream_pin: true` *"REFUSES the run … if the box cannot page-lock it"*, and the CHANGELOG says it *"refuses the run … instead of degrading silently"*. On the disk tier it **proceeds**.

Grepped: disk/CPU/inapplicable appears **0 times** in the schema description, **0 times** in the CHANGELOG entry, and `docs/` mentions `stream_pin` twice, neither about the disk tier.

So this is worse than an omission — two user-facing texts state an unconditional promise the shipped behaviour does not keep, and the rationale lives only in code comments and this thread. Your reasoning ("inapplicable, not unsatisfiable") is a good one and I accepted it. Please put **that sentence** in the schema `description` and in `docs/performance-and-quantization.md`, and soften the CHANGELOG's unconditional "refuses".

## 4. Minor — the disk-tier announcement has no test

Deleting the whole block passes 307/307. Control: deleting the RAM refusal in the same file fails 1, so the file is exercised — a coverage gap, not a dead file.

## Heads-up: a latent crash if #426 lands near you

Both PRs merge cleanly into `main` **separately**, but they conflict with each other — and there is a defect neither suite can see. #426 changes `_build_source`'s third argument from a spec `Mapping` to a **`Sequence` of per-layer mappings**, while your `_spec_bytes(spec)` does `spec.values()`. I ran the merged combination:

```
list -> CRASH: AttributeError 'list' object has no attribute 'values'
dict -> 16
```

It only hits the `require_pin` refusal path, and your test passes a **dict**, so the suite would stay green post-merge. Whichever of you lands second should fix `_spec_bytes` and re-run the streaming suite. I have told #426 the same.

ruff clean, 555 passed / 9 skipped across the streaming suites, no assertions removed, merges clean into current `main`.

MakazhanAlpamys:
Round three. **The design decision is right and I am accepting it — but the two wiring mutations from round two are still green, and I want to explain *why your new test cannot catch them*, because that is the useful part and it is not obvious.**

Everything below I ran myself in an isolated worktree at `408e276` with `PYTHONPATH` pointed at that tree.

## First: the disk/CPU carve-out is the correct call

You argued that on the disk tier there is no RAM store and on CPU there is no device, so pinning is *inapplicable* rather than *unsatisfiable*, and refusing would brick the large-model runs the disk tier exists for. I agree, and I want that on the record since the objection was mine. It also makes `stream_pin: true` committable to a config shared between a GPU box and a CPU box, which a refusal would not. The RAM-tier refusal is where the flag has real semantics and it is intact — I measured it at 32 layers x 4096^2 bf16 and the message names `1.07 GB`, which is arithmetically right.

My round-two condition was that "record" counts only if it is **documented where users read it and held by a test**. Neither half is met yet, and that is blocker 2 below.

## Blocker 1 — the same two mutations, and why the new test misses them

```
stream_setup.py:327   stream_pin=tcfg.stream_pin,   ->   stream_pin=None,
   tests/test_v07200.py:  174 passed, 2 skipped

stream_setup.py:418   require_pin=(tcfg.stream_pin is True) and on_cuda,   ->   require_pin=False,
   test_v07200.py + test_v07203.py (every file that mentions stream_pin):  307 passed, 7 skipped
```

I re-ran both myself rather than take them from my reviewer. Also green: hardcoding `stream_pin=True` and `stream_pin=False` at :327, and hardcoding `require_pin=False` at **both intermediate hops** — `build_streamed_model -> install_streaming` (:1676) and `install_streaming -> _build_source` (:1529). So the value can be severed at any of four points between `soup.yaml` and the source constructor with the suite green.

**Here is the mechanism, and it is a genuinely instructive one.** Your new test runs on CPU and asserts `captured["require_pin"] is False`. On CPU, `False` is the *correct* answer — `(tcfg.stream_pin is True) and on_cuda` evaluates to `False` there no matter what `stream_pin` holds. So the assertion is satisfied identically by the real expression and by a hardcoded constant. What it does kill is *deleting* the argument, which raises `KeyError` — a real property, worth keeping, but that is presence, not intent. **A test asserting a value the mutant also produces cannot discriminate.** Same shape as the seed lesson in CLAUDE.md (#341), where `test_same_seed_reproduces` passed pre-fix because both runs were seed 42 anyway.

Two of the three closures are CPU-runnable, so this is coverage, not difficulty. I wrote them, ran them against your code (**9 passed**) and against the mutants (each fails):

**C1 — the value reaches the plan.** Drive the pre-flight through your own `test_v07203` harness, assert the `6.56` cost note appears with `stream_pin: false`, **plus a control that it does not appear when the key is unset** — without that second half, a mutant that always prints it would pass. Kills the three `:327` mutations.

```python
def test_false_makes_the_preflight_state_the_cost(self, tmp_path, monkeypatch):
    out = self._run_capture(tmp_path, monkeypatch, "  stream_pin: false\n")
    assert "6.56" in out, out

def test_unset_does_not_state_a_forced_pageable_cost(self, tmp_path, monkeypatch):
    out = self._run_capture(tmp_path, monkeypatch, "")
    assert "6.56" not in out, out
```

One trap that cost me a false negative on the first attempt, so you do not hit it: **the pre-flight notes reach the user inside a `rich.panel.Panel`**, and a naive recording console stringifies that to the object repr. It has to render through a real `Console(file=StringIO(), width=200)` — which is this project's documented convention for exactly this reason.

**C2 — `require_pin` survives every hop.** Subclass `RamSource` so `pin=True` raises, then call the real `build_streamed_model(..., pin=True, require_pin=True)` over your existing `_tiny_stream` harness and require it to raise; the control is the same page-lock failure **without** the flag, which must fall back silently. CPU-only. Kills both intermediate-hop mutations — I confirmed the `install_streaming -> _build_source` one fails this test.

**C3 (CUDA-gated) — the `and on_cuda` half.** `stream_pin: true` on cuda must arrive as `True`, unset must arrive as `False`. It skips on CI; it is still the only thing pinning that conjunction, and worth having for whoever runs the suite on a GPU box.

I have the full file if you want it verbatim — say the word and I will paste it.

## Blocker 2 — the round-three decision is undocumented and untested

**Untested.** Deleting the entire disk-tier announcement block (`layer_stream_runtime.py:1590-1600`) leaves `307 passed, 7 skipped`. The CPU half *is* covered — disabling that block kills `test_stream_pin_threads_require_pin_into_the_runtime_and_announces_cpu` — so only the disk half is unguarded. A `TestDiskTierAnnouncesInapplicablePinning` plus its silent-without-the-request control closes it.

**Undocumented, and this is the part I care about more.** All three user-facing surfaces still describe `true` as an unconditional refusal:

- `schema.py:3051` — *"'true' forces the pinned store and REFUSES the run, naming the store size, if the box cannot page-lock it"*
- `CHANGELOG.md:52` — *"`stream_pin: true` forces the pinned store and refuses the run"*
- `docs/performance-and-quantization.md:238` — *"Force the pinned RAM store off (escape hatch) or on"*

None mentions that on two of three tiers it announces and proceeds instead. The only places the carve-out is written down are a code comment and a GitHub comment. That is the round-two defect — an undocumented divergence between what the flag promises and what it does — relocated from the code into the docs. One clause in each of the three.

## Two nits

**AC3 is unenforced.** Dropping the `GB` store-size figure from the refusal message leaves the suite green. You wrote a whole CHANGELOG paragraph arguing why the store size is the honest number to cite; nothing stops it being removed. Three lines.

**`decide_pinning`'s docstring makes a claim the code does not honour.** It says of `stream_pin=True`: *"here `True` only records the intent so the pre-flight reflects it."* Measured:

```
stream_pin=None:  pinned=True,  notes mentioning stream_pin = []
stream_pin=True:  pinned=True,  notes mentioning stream_pin = []
stream_pin=False: pinned=False, notes = ['training.stream_pin=false forces a pageable RAM store...']
```

`build_stream_plan` appends `decision.reason` only `if not decision.pinned`, so a forced-on pin says nothing at all. Given this PR's own framing — record, never silence — surfacing it is the consistent fix; correcting the docstring is the honest minimum.

## Good news on the rebase: it is mechanical

You are CONFLICTING against `main` again (#433 and #321 landed after your last rebase), but I measured it rather than warning you off it: **only `CHANGELOG.md` actually conflicts.** `schema.py`, `stream_setup.py`, `layer_stream_runtime.py` and the docs all auto-merge, because each collision is two sides appending distinct things to the same construct:

- `stream_setup.py`, the `build_streamed_model(...)` call — you insert `require_pin=` after `pin=`, #321 inserts `double_quant=` after `quant=`.
- `layer_stream_runtime.py` — you add `require_pin=require_pin` to the `install_streaming(...)` call; #433 adds `assert_trainable_adapters_materialized(model)` five lines away.
- `schema.py` — #321 touches L1745/L5376, you touch L3043/L4935.

I built the merged tree and ran the streaming suite on it: **483 passed, 7 skipped**, including #433's guard file and my controls. Nothing semantic. (The earlier warning about #426 rewriting the *same* `install_streaming` statement still stands for that PR — but #426 is not on `main`, so it is not your problem here.)

Also confirmed nothing from #365 was lost in your resolution: `stream_disk_kind` is intact in the footgun tuple, the error message, the docs list and the `build_stream_plan` lambda, alongside the new key.

## Clean

`466 passed, 7 skipped` across the four streaming files; ruff clean; **zero test lines removed**; `.claude/CLAUDE.md` untouched as asked; the schema key is `Optional[bool] = None`, round-trips through `model_dump()` without tripping its own footgun guard (the #321 trap), and rejects `2` / `1.5` / `"maybe"` / `[]`. Nothing in the diff is undescribed.

One unrelated observation, so it is not mistaken for yours: `test_v07203.py::TestMeasuredGemmCeiling::test_takes_the_best_repeat_not_the_first` failed in 3 of 16 heavily-parallel runs and is green 3/3 run alone on clean code. Pre-existing load-sensitive flake; I will file it separately.

## What I need

C1 and C2 (CPU-runnable, both written and verified above), C3 if you want the CUDA half pinned; the disk-tier announcement test; the carve-out clause in schema description + CHANGELOG + docs; and the CHANGELOG rebase. The two nits are your call.

A third round on the same finding is not a great look for either of us, and I do not think it is carelessness on your side — the assertion you wrote is one that *looks* discriminating and is not. That distinction is worth more than this PR.

MakazhanAlpamys:
Filed the flake as #444 (CUDA-gated, so it only bites on a GPU box and CI never sees it). Nothing to do with your change — flagged so a future red there is not attributed to it.

ousamabenyounes:
Round three closed. Thank you for the mechanism explanation — that was the useful part, and you were right that it is not obvious.

The defect you named is exactly it: on CPU `(tcfg.stream_pin is True) and on_cuda` is `False` whatever `stream_pin` holds, so `assert captured["require_pin"] is False` is satisfied identically by the real expression and by a constant. It pinned the argument's presence, not its value. So rather than assert that these are fixed, here is the mutation matrix, each mutation applied to the rebased head and run against the full `test_v07200.py + test_v07203.py` (clean: **316 passed, 12 skipped**):

| Mutation | Result | Killed by |
|---|---|---|
| `stream_setup.py` `stream_pin=tcfg.stream_pin` → `None` | **KILLED** | `test_stream_pin_false_makes_the_preflight_state_the_cost` |
| `stream_setup.py` `require_pin=(...) and on_cuda` → `False` | **KILLED** | `test_stream_pin_true_reaches_the_runtime_as_true_on_cuda` |
| `build_streamed_model → install_streaming` `require_pin` → `False` | **KILLED** | `test_forced_pin_refuses_through_the_whole_real_chain` |
| `install_streaming → _build_source` `require_pin` → `False` | **KILLED** | same |
| delete the disk-tier announcement block | **KILLED** | `test_an_explicit_request_is_announced_not_dropped` |
| drop the `GB` store size from the refusal | **KILLED** | `test_the_refusal_names_the_store_size[64-2.15 GB]` |
| revert the forced-on plan note | **KILLED** | `test_stream_pin_true_is_recorded_in_the_plan_notes` |

**C1** renders through a real `Console(file=StringIO(), width=200)` — thank you for the warning about the `Panel` repr, it would have cost me the same false negative. It asserts both `training.stream_pin=false` and `6.56`, with the unset control asserting both absent.

**C2** drives the real `build_streamed_model → install_streaming → _build_source` chain over `_tiny_stream` (which gained pass-through `pin=`/`require_pin=` kwargs) with a subclass of the real `RamSource` that raises on `pin=True`; the control is the same page-lock failure *without* the request, which must still fall back.

**One deviation from your suggestion, and it is the one thing I would like you to check.** I did **not** CUDA-gate C3. `on_cuda` is a string check, and the only things behind that branch that need a device are `mem_get_info()` and the allocator hint — both stubbable. So the `and on_cuda` half is pinned on CI rather than skipping there, which is where the `require_pin=False` mutation actually needed to die. It is a wiring test, not GPU integration coverage: `build_streamed_model` is still stubbed, so it cannot catch a GPU-only failure past that point.

**The disk-tier announcement test calls `_build_source` directly** rather than going end-to-end. It has to: `require_pin` is gated on `on_cuda`, so on a CPU box the end-to-end path never reaches that branch and an end-to-end test would pass vacuously.

**Blocker 2 — documented.** The carve-out sentence is now in all three surfaces you grepped: the schema `description`, the CHANGELOG entry (the unconditional "refuses" is now "on the RAM tier ... refuses", with the disk/CPU behaviour spelled out), and a new `### Forcing the pin (training.stream_pin)` section in `docs/performance-and-quantization.md` with a tier/device table.

**Both nits taken.** AC3 is parametrized over two store sizes (32 layers → `1.07 GB`, 64 → `2.15 GB`), each asserting the *other* figure is absent — a message that hardcoded one value passes a single-case test, which is the same defect one level down.

For the docstring nit: `build_stream_plan` now surfaces the forced-on reason, **restricted to the RAM tier**. Scoping that mattered more than I expected — my first version emitted it unconditionally, and on the disk tier it printed "forces a page-locked store" on a run that does no such thing. That is the round-two defect again, so the reason's wording is now RAM-tier-scoped *and* the note is suppressed off-tier, with `test_the_disk_tier_does_not_claim_a_page_locked_store` holding it.

Rebased onto current `main` (`7313eeb`); as you measured, only `CHANGELOG.md` conflicted and it resolved as a union. `#426` is still open, so per your sequencing note I have left `_spec_bytes` alone — that is theirs to rebase onto this.

Validation: `ruff` clean; streaming area **316 passed / 12 skipped** against `292 / 12` on `main`; full local suite iso against the `main` baseline (the same pre-existing failures in `test_v0712.py` / `test_v0713.py` / `test_semantic_split.py`, none in the streaming area, none new).

Pushed as `110d4f5` (force-with-lease — the rebase rewrote the two commits, nothing else). Changed-line coverage on the diff is 100% (24/24); the full suite is `41 failed / 17883 passed` against `main`'s `41 failed / 17860 passed`, same failure identities.

MakazhanAlpamys:
Round four. **Every blocker is closed, and I verified each by re-running the exact mutation that survived before.** One CHANGELOG problem left, and it is placement rather than content.

## The four things that were open, all now killed

| what survived before | now |
|---|---|
| `stream_pin=tcfg.stream_pin` -> `None` at `stream_setup.py:336` | **KILLED** — `test_stream_pin_false_makes_the_preflight_state_the_cost` |
| `require_pin=(...)` -> `False` at `:427` | **KILLED** — `test_stream_pin_true_reaches_the_runtime_as_true_on_cuda` |
| delete the whole disk-tier announcement block | **KILLED — by two tests**: `test_an_explicit_request_is_announced_not_dropped` and `test_it_still_reaches_the_log_without_a_console` |
| drop the store-size figure from the refusal (issue #366 AC3) | **KILLED** — `test_the_refusal_names_the_store_size[64-2.15 GB]`, and you parametrized it |

The second row is the one I want to acknowledge properly. Three rounds ago that assertion could not discriminate, because on CPU `(tcfg.stream_pin is True) and on_cuda` is `False` whatever the config holds — so the test was satisfied identically by the real expression and by a constant. You did not paper over that; you built the CUDA-gated version that actually distinguishes them. It runs here because this box has a GPU, and it will skip on CI, which is the honest trade rather than a fake pass.

The two disk-tier tests are better than what I asked for: I wanted the announcement pinned, and you also pinned that it reaches the log when there is no console — the path an operator hits in a non-interactive run.

**The carve-out is documented where users read it.** `schema.py`, the CHANGELOG bullet and `docs/performance-and-quantization.md` all now state that `true` announces and proceeds on the disk tier and on CPU rather than refusing. That was my round-three condition for accepting "record + proceed" as the design, and it is met.

## One blocker left, and it is mechanical

**Your CHANGELOG entry is in the frozen `[0.73.3]` section.** The bullet is at line 117, and the section boundaries are `[Unreleased]` at 13, `[0.73.3]` at 58, `[0.73.2]` at 317 — so it landed inside the released one. v0.73.3 shipped while this PR was open and rolled `[Unreleased]` underneath it; your rebase carried the entry along with the section it was in rather than into the new empty one.

**And the attribution suffix is missing** — it reads `(#366).` where the house form is `(#366 by @ousamabenyounes in #416)`. `grep -c "by @ousamabenyounes in #416" CHANGELOG.md` returns `0`.

Move the bullet up to `[Unreleased]` under a `### Added` heading and add the suffix, and this merges. Nothing about the code needs to change.

## Where the review landed overall

Four rounds is a lot, and I do not think any of them were wasted — but I want to name what actually made the difference, because it was not persistence. It was that the round-three assertion *looked* discriminating and was not, and once that mechanism was written down you fixed it in one pass and then went past it on the disk-tier half. The distinction between a test that pins a value and a test that pins an *intent* is the durable thing here; it outlives this feature.

CI is at 10 green with 3 still running. I will merge on green once the CHANGELOG moves.

ousamabenyounes:
Moved, and the attribution suffix is on — `e112ae1`.

| ask | before | after |
|---|---|---|
| bullet sat in the frozen `[0.73.3]` block | line 117, section `[0.73.3]` | line 17, section `[Unreleased] ### Added` |
| `grep -c "by @ousamabenyounes in #416" CHANGELOG.md` | `0` | `1` |

Two things I checked rather than assumed:

**Placement was verified post-merge, not branch-local.** `git merge-tree --write-tree origin/main HEAD` returns rc 0 and the bullet lands under `[Unreleased]` in the merged tree.

**Top vs tail insertion is not cosmetic here.** I built the tail-append variant and merged it: rc **1**, conflict — `#452` landed `glm-5.1-dpo` at that exact spot after your review. Top insertion merges clean.

Gates: `./run-ci.sh` RED rc=1 on `110d4f5` → GREEN rc=0 on `e112ae1`, ruff clean. `run-ci.sh` is untracked and ignored via `.git/info/exclude`, so it is not part of the diff.

Remote CI on `e112ae1`: **13/13 green** across the ubuntu/macos/windows × 3.10–3.12 matrix, plus `lint`, `type-check`, `mlx-smoke` and `validate-recipes`.

MakazhanAlpamys:
Merged as `9d1ce2c`. Five rounds all in, and the thing I want on the record is not that you persisted — it is **round three**.

Your assertion on the `require_pin` wiring looked discriminating and was not: on CPU `(tcfg.stream_pin is True) and on_cuda` is `False` whatever the config holds, so `assert captured["require_pin"] is False` was satisfied identically by the real expression and by a constant. It pinned the argument's *presence*, not its *value*. You did not patch around that — you built the CUDA-gated version that actually distinguishes them, and let it skip on CI rather than manufacture a fake pass. **The distinction between a test that pins a value and one that pins an intent is the durable thing here; it outlives this feature.**

What I ran on this final push, since a CHANGELOG round still deserves a check rather than a wave-through:

- `git diff 110d4f57 e112ae1b` -> **one file, +19/-18**. Code byte-identical to what I had already broken and re-broken, so I did not re-run the matrix against a diff that could not have changed it.
- Placement verified in the **merged** tree against current `main`, not branch-local — and re-verified after #409 landed underneath you. `git merge-tree --write-tree` rc 0, bullet at line 17, inside `[Unreleased]` (13), clear of the frozen `[0.73.3]` (122).
- CI 13/13, and I confirmed the run's `headSha` equals `headRefOid` — a cached `pull/head` ref had me reviewing a stale commit on another PR this week.
- Test-file deletions in the diff: three lines, all helper-signature edits, no test removed or weakened.

Your own top-vs-tail check was the right instinct and I want to say so: you built the tail-append variant, merged it, got rc 1 against `#452`'s `glm-5.1-dpo` landing at that exact spot, and chose top insertion on evidence. That is the second time in this PR you answered a placement question by running it instead of arguing it.

**The carve-out is what makes this mergeable as a design.** `true` announces and proceeds on the disk tier and on CPU rather than refusing an otherwise-fine config — and it says so in `schema.py`, the CHANGELOG and `docs/performance-and-quantization.md`. A flag whose meaning shifts per tier and states that nowhere is precisely the class of defect this repo keeps writing down; you closed it in the same PR that could have created it.

Credited in `CONTRIBUTORS.md`. Two merges from you tonight — #409 landed twenty minutes ago.

PR #417, opened 2026-08-15
feat(draft): support cross-tokenizer drafts

### LINKED ISSUE #304: soup draft: support cross-tokenizer drafts (ULD-aligned distill + universal assisted decoding)
## Current behaviour

`soup draft` (v0.71.33) requires the draft and the target to **share a tokenizer**. A mismatched pair is refused up front, in both `distill` and `measure`:

```
Draft and target must share a tokenizer, but vocab sizes differ (target=49152, draft=151936).
Speculative decoding proposes draft token ids into the target's vocabulary, so a mismatched
pair produces garbage rather than a speedup.
```

The guard (`utils/draft.same_tokenizer`) is deliberately strict — it compares `vocab_size` **and** the ids produced over a fixed probe corpus, because two tokenizers can both report 32000 tokens and still disagree on every one of them.

## Why this is a real limitation

It rules out the most useful drafts. If your target is, say, a Llama-3 fine-tune, you cannot use an off-the-shelf tiny model from another family as a draft, even when that model is a much better speed/quality trade than anything in the Llama family. It also means `soup shrink` output is currently the *only* draft base guaranteed to work out of the box.

## Fix path

Two halves, and they are independent — a PR doing only one is welcome.

**1. Distillation across tokenizers.** Soup already has the machinery: `training.uld_strategy` (v0.70.0 / v0.71.11) does cross-tokenizer ULD, and `wasserstein_aligned` (v0.71.18 #258) aligns fully-disjoint tokenizers by decoding both sides and matching character spans with `difflib`. `commands/draft.py::_build_distill_config_yaml` would pass this through instead of refusing, when the vocabs differ.

**2. Inference across tokenizers.** `transformers` supports *universal assisted decoding* — `generate(..., assistant_model=draft, tokenizer=target_tok, assistant_tokenizer=draft_tok)`. `utils/draft.measure_throughput` and `commands/serve.py::_load_draft_model` would need to thread both tokenizers through, gated on the installed transformers version.

**Note that acceptance is measured differently in this regime.** `measure_acceptance` currently compares argmax token *ids* against the target's own ids, which is meaningless across vocabularies — a cross-tokenizer acceptance metric has to compare over decoded text spans (the same alignment `wasserstein_aligned` already does), not ids. That is the interesting part of this issue, and it should land before the ULD passthrough, otherwise there is no way to tell whether a cross-tokenizer draft is any good.

## Acceptance criteria

- [ ] `soup draft distill` accepts a mismatched pair (routing through `uld_strategy`) instead of refusing.
- [ ] `soup draft measure` reports a *meaningful* acceptance number for a cross-tokenizer pair — i.e. computed over aligned decoded spans, not raw ids — with a test that would fail if ids were compared directly.
- [ ] `soup serve --speculative-decoding` / `--auto-spec` passes `tokenizer=` + `assistant_tokenizer=` when the pair differs, gated on the transformers version, with a friendly error when unsupported.
- [ ] The strict `same_tokenizer` guard stays the default for the *same*-tokenizer path (it exists because a mismatch otherwise fails silently).

## Honest caveat

Read #303 first. On the pair we could validate, distilling a draft did not improve acceptance at all, and speculative decoding was a net *slowdown*. Cross-tokenizer support widens the choice of drafts, but it does not by itself make speculative decoding pay off — `soup draft measure` is still what tells you that.

CODING-DARSH:
## What does this PR do?

Adds cross-tokenizer draft support to `soup draft` by introducing:

- Cross-tokenizer draft distillation through `wasserstein_aligned` ULD when the target and draft tokenizers differ.
- Character-span based acceptance measurement for mismatched tokenizers instead of comparing raw token IDs.
- Universal Assisted Decoding (UAD) support by passing the target and draft tokenizers to Transformers when required.
- Cross-tokenizer speculative decoding support in `soup serve`, including streaming generation.
- Capability checks and clear errors when the installed Transformers version does not support cross-tokenizer assisted decoding.
- Regression and integration tests covering tokenizer compatibility, span alignment, ULD routing, UAD plumbing, and serving behavior.

The existing same-tokenizer path remains unchanged and continues to use the normal speculative decoding/distillation flow.

Closes #304
Refs #345

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist


MakazhanAlpamys (changes requested):
This is a lot of well-organised work — 968 lines, CI 11/11, `tests/test_issue304_cross_tokenizer.py` + `tests/test_v07133.py` 145 passed / 1 skipped locally, and the same-tokenizer path left alone rather than generalised into something that serves both badly. Routing distillation through `wasserstein_aligned` rather than inventing a second alignment is the right reuse.

Two things before this can merge, and the first is not about your code.

## Blocking: it measures with an instrument that is known to be wrong

**#345 is open**: `measure_acceptance` calls `generate(do_sample=False)`, which disables sampling but **not** the checkpoint's `repetition_penalty` logits processor. The target then emits `argmax(penalised logits)` while the draft is scored on `argmax(raw logits)` — the two are compared under different rules. Measured there: a draft **identical to the target** scores **0.8623** on Qwen2.5-0.5B, and on Qwen2.5-32B it flips the verdict band, 0.6857 MODERATE → 0.7072 STRONG once the penalty is neutralised.

I checked this branch: `draft.py` still sets only `"do_sample": False` at both call sites, with no `logits_processor` handling. So the cross-tokenizer path inherits it.

That matters more here than in the same-tokenizer case, not less. The whole proposition of cross-tokenizer drafting is *"can this differently-tokenized model serve as a draft at all?"*, and the number that answers it is the acceptance rate — which is exactly the number #345 says is wrong by up to 4 points, in the band where `--min-acceptance` gates. A user will reasonably read a fresh cross-tokenizer α as evidence about the pairing when part of it is an artefact of the measurement.

**Two ways forward, and either is fine:**

1. **Fix #345 first** (neutralise the penalty for the measurement, or pass the same processor to both sides as `candidate_generator.py:314` does in real assisted generation), then land this on top. Cleanest, and #345 is a small change.
2. **Land this with the limitation named in the panel and the docs** — the reported α inherits #345 and is not comparable across models with different declared `repetition_penalty` until it is fixed.

What I do not want is the third option, which is shipping a new measurement surface that quietly carries a known error.

## Blocking: no CHANGELOG entry and no docs for a new feature

The diff is `commands/draft.py`, `commands/serve.py`, `utils/draft.py` and two test files. For a **new capability** — a new `soup serve` behaviour among them — the release checklist wants `CHANGELOG.md` under `[Unreleased]` and the matching page under `docs/` (`docs/serving-and-export.md` for the serve half, and wherever `soup draft` is documented for the rest).

This one matters beyond bookkeeping: `soup serve` gains cross-tokenizer speculative decoding including the streaming path, and a user cannot discover that from a release note that does not exist.

## Worth naming, because they are the right calls

- **The same-tokenizer path is untouched.** A single "universal" path that handles both would have put every existing draft user behind new code for no benefit; keeping the fork explicit means this feature cannot regress what already works.
- **Capability-probing transformers rather than version-comparing.** That is the lesson `trainer/_trl_compat.py` exists for — a version table in this repo has been wrong twice.
- **Character-span alignment instead of comparing token ids.** Comparing ids across disjoint vocabularies is the obvious thing and it is meaningless; spans are the only common ground.

---

Not a criticism of the work — the design decisions in it are the ones I would have wanted. Tell me which route you prefer on #345 and I will help move it; if you would rather take #345 yourself as a prerequisite, it is unclaimed and small.

CODING-DARSH:
Thanks for the detailed review. I addressed both blockers you called out.

1. Acceptance measurement

The measurement path now explicitly sets repetition_penalty=1.0 when generating the target continuation. This prevents checkpoint-level repetition-penalty processors from changing the target's greedy argmax while the draft is evaluated from raw logits.

I also added a regression test verifying that an identical target/draft pair achieves 100% self-acceptance and that the measurement call explicitly uses repetition_penalty=1.0.

The existing cross-tokenizer measurement remains character-span based, so this correction applies to both same-tokenizer and cross-tokenizer measurement without changing the alignment logic.

2. Documentation

Added the cross-tokenizer draft behavior to the [Unreleased] changelog and updated docs/serving-and-export.md to document:

soup draft distill routing mismatched tokenizers through wasserstein_aligned
cross-tokenizer acceptance measurement via decoded character-span alignment
cross-tokenizer speculative decoding through Transformers UAD
the Transformers capability requirement and unsupported-version error
preservation of the existing same-tokenizer path
the fact that cross-tokenizer support does not guarantee a speedup and should be validated with soup draft measure

The changes are split into focused commits:

99b7459 — fix acceptance measurement by neutralizing the repetition penalty
6ddcb0b — document cross-tokenizer draft support

The original cross-tokenizer implementation remains in 0609a25.

The focused regression tests and Ruff checks are passing, and git diff --check is clean.

MakazhanAlpamys:
Heads-up before you go chasing this: **the red CI on your branch is ours, not yours.**

```
FAILED tests/test_issue402_restart_capacity.py::test_live_persisted_run_detects_only_live_pids
AssertionError: assert None == 'run-prior'
```

That is on `main`, not in your diff. #407 and #408 both merged green on their own branches and then collided semantically — #407 made a stale `running` row reconcile *when it is read* and persisted the flip, and #408's test reused one row across a dead-pid phase and then swapped in a live pid, by which point the row was already terminal. Each PR had been tested against a `main` that did not yet carry the other.

Repaired in `7bc3f83`. Rebase onto current `main` and those cells should clear.

Your re-review is already in progress on the substance — I will come back on `repetition_penalty=1.0` and the #345 question separately. Sorry for the noise.

MakazhanAlpamys (changes requested):
Re-reviewed. **The `repetition_penalty=1.0` change is right, and it is a bigger result than the PR claims.** One blocker remains, and it is in the test rather than the code.

I verified the fix rather than accepting it: it lands inside `measure_acceptance` itself (`utils/draft.py:541`), which is the function #345 names and the only acceptance-measurement path in the repo — so every caller inherits it and there is no second copy to drift. Against installed transformers 4.57.6, `_get_logits_processor` constructs `RepetitionPenaltyLogitsProcessor` only when the value is not `None` and `!= 1.0`, and generate-kwargs outrank `model.generation_config`. So the processor is never built and the target's greedy argmax is raw. That is #345's mechanism, closed.

Six mutations against your code, **all six caught** — including breaking the span containment in `count_accepted_spans`, forcing `is_same = True`, disabling the `PROBE_CORPUS` id check in `same_tokenizer`, removing the UAD support gate, and dropping the `assistant_tokenizer` threading in serve. Blocker 2 (CHANGELOG + docs) is fully addressed; I particularly note you corrected the previously-false "a mismatch is refused up front" prose rather than just appending to it.

`tests/test_issue304_cross_tokenizer.py` + `tests/test_v07133.py`: **146 passed, 1 skipped**, matching your number. Broader `-k "draft or serve or uld or spec"`: 675 passed, 8 skipped. Ruff clean. The two removed test defs are renames of inverted behaviour, and the exit-code-1 path is retained by your new `test_mismatched_tokenizer_unsupported_uad_exits_one` — net +32/−2, nothing weakened.

## Blocker: the #345 control is vacuous

`test_repetition_penalty_neutralized_for_self_acceptance` asserts `accepted == 3, total == 3` for an identical draft. But `_FakeTarget.generate` (`tests/test_v07133.py:861-866`) records kwargs and returns a **hardcoded** sequence — it never reads `repetition_penalty` — and `_FakeDraft`'s argmax is hardcoded to match. So that assertion is true with the fix and true without it.

Proven, not suspected: revert the source to `1.1`, delete **only** the two `target.calls[0]["repetition_penalty"] == 1.0` assertions, and `TestMeasureAcceptance` still goes **6 green**.

The kwarg assertions are real coverage and should stay. But they are not #345's stated control, which is literally *"a draft identical to the target scores 1.0"*. A fake that ignores the penalty structurally cannot be that control — and it costs no GPU to build one. Please make `_FakeTarget` actually apply a penalty when `repetition_penalty != 1.0` (down-weight already-emitted ids), so an identical draft scores **1.0 with the fix and < 1.0 without it**.

## Credit where it is due

The PR body says only `Closes #304` and the CHANGELOG files this entirely under `#417` — `git diff main...pr-417 | grep 345` returns nothing. Please add **`Refs #345`** (not `Closes`: #345 also covers the fp32 teacher and the captured subprocess output, and its `--steps N` item already landed on `main` in `a1b6a34`) and reference it in the CHANGELOG bullet.

Worth stating in that bullet too: the neutralisation covers `repetition_penalty` only. `no_repeat_ngram_size`, `encoder_repetition_penalty`, `min_new_tokens`, `bad_words_ids`, `suppress_tokens` and `sequence_bias` all survive `do_sample=False` and would still skew α. That is a narrowing of #345's acceptance box 1, not a defect — but it should be written down rather than discovered later.

## Two smaller items

**Restore the difflib cap.** `count_accepted_spans` (`utils/draft.py:124`) runs `SequenceMatcher(..., autojunk=False)` on uncapped strings. The identical difflib path in `utils/uld.py:394-395` truncates to `_MAX_ALIGN_CHARS = 8192` with the comment *"so the difflib fallback stays bounded"*, and `autojunk=False` additionally disables difflib's own speed heuristic. Length is user-controlled via `--max-new-tokens`. Please reuse that constant.

**Move `total += len(target_pieces)` below the `continue`** (`utils/draft.py:577`). Today a prompt the draft tokenizer cannot extend is counted as *all rejected* rather than skipped, which silently deflates α — and it contradicts `test_empty_generation_is_skipped_not_counted`, which pins the opposite convention for the same-tokenizer branch a few lines away.

## One question, not a request

`draft_prompt_len` comes from `draft_tokenizer(prompt)` while proposals are sliced out of `draft_tokenizer(full_text)`. With real BPE those two can disagree at the prompt/continuation boundary; if the last prompt character merges with the first generated one, the slice is off by one and α is wrong. `_FakeTok`'s `encode_map` makes the prefix stable by construction, so the fixtures cannot see this either way. Have you checked it against a real mismatched pair (Llama-3 target / Qwen2.5 draft)? If it is a known limitation, a docstring note is enough.

Ignore the red CI — that was ours, and it is repaired on `main` in `7bc3f83`.

CODING-DARSH:
Addressed the remaining requested items.

Made _FakeTarget actually apply repetition_penalty to previously seen token IDs, so the self-acceptance control now genuinely fails if the production repetition_penalty=1.0 fix is removed.
Kept the explicit repetition_penalty=1.0 kwarg assertions.
Restored the _MAX_ALIGN_CHARS bound for the SequenceMatcher fallback.
Moved total += len(target_pieces) below the skip path for prompts the draft tokenizer cannot extend.
Added Refs #345 and narrowed the CHANGELOG/docs wording to make clear that only repetition_penalty is neutralized.

Validation:

tests/test_v07133.py + tests/test_issue304_cross_tokenizer.py: 146 passed, 1 skipped
Ruff: clean
git diff upstream/main...HEAD --check: clean

MakazhanAlpamys:
Status, since you have been waiting on me rather than the other way round: your 10:28 push is in re-review now, and the two days of silence were mine, not a queue. Coming back with the finished thing — specifically the same experiment I used to show the control was vacuous last round (revert the source to 1.1, delete only the kwarg assertions, see whether anything goes red), plus the BPE-boundary question I raised, which I want to try to construct a fixture for rather than leave as a hypothetical.

MakazhanAlpamys:
Re-reviewed at `cb3d99e`, isolated worktree, `PYTHONPATH` pinned to that tree. **The blocker is cleared — your `_FakeTarget` is a real control now, and I checked it the same way I broke it last time.**

## R1 — resolved

Two experiments, both run:

**Your round-two form** (source `1.0` -> `1.1`, delete only the two `target.calls[0]["repetition_penalty"] == 1.0` assertions): `6 passed`. Still green — but read on, because the reason is not the one from last round.

**The fix fully removed** (delete the `"repetition_penalty": 1.0` line so the fake falls back to its own `default_repetition_penalty=1.2`):

```
>       assert accepted == 3
E       assert 2 == 3
FAILED tests/test_v07133.py::TestMeasureAcceptance::test_repetition_penalty_neutralized_for_self_acceptance
```

And #345's stated control, an identical target/draft pair through `measure_acceptance` + `acceptance_rate`:

| state | accepted | total | alpha |
|---|---|---|---|
| fix intact | 3 | 3 | **1.0000** |
| fix removed | 2 | 3 | **0.6667** |

`_FakeTarget.generate` genuinely divides the logit of previously-seen ids and re-takes the argmax. That is the control the issue asked for and it did not exist a round ago.

**One nit on it.** The fake scores the expected token 10.0 against an alternative 9.0, so it only flips once the penalty exceeds 10/9:

```
penalty=1.1   -> [10, 11, 12]  unchanged
penalty=1.11  -> [10, 11, 12]  unchanged
penalty=1.12  -> [99, 11, 12]  flips
```

So the fixture is deaf below ~1.111 — which is exactly the band real checkpoints ship in (1.05-1.1). Narrowing the gap (10.0 vs 9.9 discriminates from ~1.011) or parametrizing over `(1.05, 1.1, 1.2)` closes it. Not a blocker: the control fires on removal, which is what I asked for.

## Two things you added in response to review, and neither is covered

Both of these are code you wrote *because* I asked for it, which is the frustrating part — the change is right and nothing would notice if it were reverted.

**R2 — the `_MAX_ALIGN_CHARS` cap.** Correctly the shared constant, not a copied literal:

```
src/soup_cli/utils/draft.py:111:    from soup_cli.utils.uld import _MAX_ALIGN_CHARS, _char_spans
literal 8192 in draft.py: False
```

Turning both slices into no-ops: **146 passed, 1 skipped**. `grep -rn "_MAX_ALIGN_CHARS" tests/` returns nothing. The cap is load-bearing — measured on your exact `SequenceMatcher(..., autojunk=False)` call with a low-entropy alphabet:

```
capped   (shipped) len= 8192    5.50s
uncapped (mutant)  len=48000  329.71s
```

60x, and it grows superlinearly, on a length the user controls through `--max-new-tokens`. A test over >8192 chars asserting either a wall-clock bound or the truncation under a monkeypatched small cap closes it.

**R3 — the skip.** `total += len(target_pieces)` is below the `continue`, and a prompt the draft tokenizer cannot extend is now correctly skipped:

```
accepted=0 total=0   -> SKIPPED (correct)
```

I moved that one line back above the `continue` myself rather than take it from my reviewer: **146 passed, 1 skipped**. The mutant reports `accepted=0 total=2`, i.e. alpha silently **0.0** for a prompt that should not have been counted at all — a wrong number, not an error, in the direction that makes a usable draft look unusable. That is the more serious of the two. The fixture is about twenty lines: assert `(accepted, total) == (0, 0)` for an un-extendable prompt, with a control that an extendable one still counts.

## R4 — the narrowing was not written

`Refs #345` is in the PR body, the CHANGELOG and the docs — thank you. But:

```
grep -rn "no_repeat_ngram\|encoder_repetition_penalty\|min_new_tokens\|bad_words_ids\|suppress_tokens\|sequence_bias" \
     CHANGELOG.md docs/serving-and-export.md src/soup_cli/utils/draft.py
-> no matches
```

`gen_kwargs` sets `input_ids`, `max_new_tokens`, `do_sample`, `repetition_penalty`, `pad_token_id`, `attention_mask` — every other checkpoint-level processor still applies and still skews alpha. Both texts currently read as though the measurement is now clean. One clause naming the six that survive is all I want; the point is that the next person reading it knows the box is narrowed rather than closed.

## Q1 — it reproduces, and it is a lower bound

You did not answer this one, so I built the fixture rather than leaving it hypothetical. A **perfect** draft — logits reproducing the target text exactly — varying only whether the draft tokenizer merges the last prompt character with the first generated one:

```
control: clean boundary          accepted=2 total=2  alpha=1.000
merged boundary (b + c -> 'bc')  accepted=1 total=2  alpha=0.500
```

Severity is exactly one target token, `-1/n_gen`:

```
n_gen  clean  merged   loss
    2  1.000  0.500   -0.500
    8  1.000  0.875   -0.125
   24  1.000  0.958   -0.042
```

At the default `max_new_tokens=64` that is about -1.6%, and it is **always downward** — the merged token straddling the boundary drops out of the proposals. So it can only under-report alpha: a false "this draft is not good enough", never a false green. That is the safe direction, which is why it is a nit and not a blocker. A docstring note on the cross-tokenizer branch saying alpha is a lower bound by up to one boundary token would be enough, and I am happy for the fixture to be a follow-up issue instead.

## The rest is clean

All six round-two mutations still killed after your merge from `main` — span containment, forced `is_same`, the `PROBE_CORPUS` id check, the UAD gate, the `assistant_tokenizer` threading, and the penalty kwarg. Diff hygiene: the same two test defs removed as last round, both legitimate renames with replacements; 32 test defs added; nothing new removed or weakened. `146 passed, 1 skipped` on the PR's files, `577 passed, 8 skipped` on a `-k "draft or serve or spec"` sweep, ruff clean, both `--help` paths render.

## What I need

The two controls (R2, R3) and the R4 clause. Findings 1 and 5 are yours to take or leave — say the word on Q1 and I will file it as an issue instead so it is not lost.

Sorry again for the two-day gap; that was mine.

MakazhanAlpamys:
Checking in — the branch is still green and still mergeable, so nothing has rotted; this is just so it does not drift.

Outstanding from the 18 Aug review, unchanged and all small:

- **R2** — a test for `_MAX_ALIGN_CHARS`. Either a wall-clock bound over a >8192-char input, or the truncation under a monkeypatched small cap. Measured on your own `SequenceMatcher(..., autojunk=False)` call: 5.50 s capped vs 329.71 s uncapped, on a length the user picks with `--max-new-tokens`.
- **R3** — a fixture for the skip: `(accepted, total) == (0, 0)` for a prompt the draft tokenizer cannot extend, plus a control that an extendable one still counts. This is the more serious of the two — the mutant returns alpha `0.0` rather than an error, i.e. a wrong number in the direction that makes a usable draft look unusable.
- **R4** — one clause naming the six processors that still apply (`no_repeat_ngram`, `encoder_repetition_penalty`, `min_new_tokens`, `bad_words_ids`, `suppress_tokens`, `sequence_bias`), so the next reader sees the box as narrowed rather than closed.

**Findings 1 and 5 (the `_FakeTarget` sensitivity band, and the Q1 boundary-token docstring note) are yours to take or leave** — say the word and I will file them as follow-up issues so they are not lost, and they stop being part of this PR.

If you would rather hand the three items off, say so here and I will land them on your branch with authorship preserved. No rush beyond wanting to know which way it is going.

CODING-DARSH:
Got it — I’ve taken care of all three remaining items (R2, R3, and R4) on the branch, including the regression coverage and the documentation updates. I’ve also kept the scope limited to those requested changes.

I’ll leave Findings 1 and 5 as follow-up issues rather than expanding this PR further.

Sorry for the slight delay getting back to this — I’ve been wrapping up the remaining changes and verification.

MakazhanAlpamys:
Picking this up now — and **before you go looking, the red CI is not from your push.**

All 6 failures are in `tests/test_v07133.py::TestDraftMeasureVocabGate`:

```
test_assisted_arm_failure_keeps_acceptance_and_plain_on_disk
test_assisted_arm_interrupt_is_recorded_and_reraised
test_hard_death_in_the_assisted_arm_keeps_the_incremental_report
test_a_failed_status_write_does_not_mask_the_arm_outcome[crash-0]
test_a_failed_status_write_does_not_mask_the_arm_outcome[interrupt-130]
test_a_crashed_arm_without_an_output_path_still_warns
= 6 failed, 17976 passed, 285 skipped =
```

Every one of those tests arrived in **#409, which merged into `main` yesterday** (`e9dff98`, issue #344). It established a contract for `soup draft measure`: the report is written *incrementally* before the assisted-throughput arm runs, so a later failure cannot discard a completed acceptance measurement; the arm's outcome is recorded on disk as a status so `crash` / `interrupt` / `untimed` / `complete` are not byte-identical; and the status write inside the handlers is guarded so a failure there warns rather than masking the arm outcome.

You merged `main` and git resolved it textually clean, so nothing conflicted — but your rewrite of `measure` for cross-tokenizer support and #409's contract now overlap. **This is merge-order luck, not a mistake on your side.** #417 has been open since 15 Aug and #409 landed underneath it.

What I am doing rather than guessing: for each of the 6 I am determining whether it is **your code breaking #409's contract** (report no longer written early / status no longer recorded / warning no longer fires) or **#409's test being over-specific to an implementation detail you legitimately changed** (e.g. asserting on captured stdout you reworded). Those need opposite fixes and telling you the wrong one wastes your time, so I would rather spend an hour on the determination than hand you a guess.

I am also re-running R2, R3 and R4 by mutation — in particular R3, where I want to see that making the skip path return alpha `0.0` instead of skipping actually fails something, and that the control in the other direction exists.

Nothing needed from you until I come back with the per-failure verdict. Thanks for the note on Findings 1 and 5 — leaving those as follow-ups is the right call and I will file them.

MakazhanAlpamys:
Per-failure verdict, as promised. **All six are (A): your code breaking #409's contract. None is (B).** No #409 assertion turned out to be over-specific — every one asserts the documented on-disk behaviour, not wording or incidental call order. So this is "fix your code", and I am sorry to hand you that after the merge-order note — but the fix is small and it also repairs something worse.

## The cause: the assisted arm now runs twice, and the new one runs *before* the write

Your merge inserted the new combined arm **above** #409's block without removing #409's call. Both survive. `commands/draft.py`, your head vs `main`:

```
main (post-#409)          #417
737  plain                767  plain
763  _write_draft_report  770  assisted  <-- NEW, with assistant_tokenizer=draft_tok
766  assisted             779  except RuntimeError
                          804  _write_draft_report
                          807  assisted  <-- #409's, WITHOUT assistant_tokenizer
```

#409's whole guarantee is *acceptance + plain throughput are on disk before the arm that can fail*. You did not delete the pre-arm write — you moved an arm in front of it, which is behaviourally identical. And the new arm is guarded by `except RuntimeError` alone, which catches none of the three exceptions #409's handlers exist for: `ValueError`, `KeyboardInterrupt`, `SystemExit`.

That single fact explains all six:

| test | what actually happens |
|---|---|
| `..._failure_keeps_acceptance_and_plain_on_disk` | the stub's `ValueError` escapes `except RuntimeError` -> exit 1, no report |
| `..._interrupt_is_recorded_and_reraised` | `KeyboardInterrupt` at the early arm -> `FileNotFoundError: report.json`, the file never exists |
| `..._hard_death_..._keeps_the_incremental_report` | `assert False + where False = exists()` — this is *the* test whose docstring says it goes red if the pre-arm write is gone |
| `..._does_not_mask_the_arm_outcome[crash-0]` / `[interrupt-130]` | `_record_assisted_status` is never reached; `'could not record the assisted-arm outcome'` never printed |
| `..._crashed_arm_without_an_output_path_still_warns` | same escape path, `could not be measured` never prints |

One of those looked like a simple signature mismatch (`TypeError: unexpected keyword argument 'assistant_tokenizer'` against a stub with an explicit signature). I widened that stub to `**kw` and re-ran to make sure I was not chasing a test artefact — **it still fails**, with `ValueError('The main and assistant models have different tokenizers')`. So the stub signature is a cosmetic follow-on, not the cause.

## The part that matters more: your own feature does not currently report a number

`tok_s_assisted` is assigned twice. The second call — #409's, at 807 — has **no** `assistant_tokenizer`, and it overwrites the first. In `measure_throughput` the UAD branch is gated on `assistant_tokenizer is not None`, so the value that reaches the report is from a plain assisted run with mismatched tokenizers. Against transformers' real refusal on a cross-tokenizer pair:

```
exit: 0
output contains 'could not be measured': True
report tok_s_assisted: None
report assisted_status: crash
```

The correct UAD result is computed at 770 and then thrown away. **For a genuine cross-tokenizer pair, the headline feature of this PR produces no throughput number.** It also doubles wall-clock on every same-tokenizer run, which makes `docs/serving-and-export.md`'s "Compatible same-tokenizer pairs strictly preserve the existing native fast path" false as written.

**The fix is small:** delete your inserted `try:` block (768-780) and add `assistant_tokenizer=draft_tok` to the surviving call at 807. That restores all six tests, makes the feature actually report, and makes the doc sentence true again.

## R2, R3, R4 — all genuinely closed, and R3 in both directions

I ran these by mutation rather than reading them:

| | mutation | result |
|---|---|---|
| **R3** | move `total += len(target_pieces)` above the `continue` (count the unextendable prompt as a miss -> alpha 0.0) | **KILLED** — `test_cross_tokenizer_unextendable_prompt_scores_zero_zero` |
| **R3 control** | replace the guard with `if True:` (always skip) | **KILLED by two** — `test_cross_tokenizer_extendable_prompt_measures_acceptance` and the e2e in `test_issue304_cross_tokenizer.py` |
| **R2** | delete both `[:_MAX_ALIGN_CHARS]` truncations | **KILLED** — `assert 'abcdefghijklmnop' == 'abcdefghij'`, **0.47 s total** |
| **R1** (old) | remove `"repetition_penalty": 1.0` | **KILLED** — and the behavioural claim holds exactly: `(3,3)` -> `(2,3)` |

R3 is the one I pushed hardest on and you closed it properly: a `(0,0)` assertion with no control is satisfied by a function that always skips, and your control is what rules that out. **R2 you solved better than I asked** — I offered a wall-clock bound or a monkeypatched cap, and a wall-clock guard that itself burns 300 s is not usable in CI; you took the cap. **R4** is in the right place, `docs/serving-and-export.md:356-363`, inside the paragraph a reader of the acceptance number is already reading, and all six processors check out against `utils/draft.py:551`.

**Your 11 test-line deletions are all strengthenings**, which I checked line by line because this repo has seen PRs that quietly removed coverage: `_FakeTarget` widened to actually simulate the penalty; two tests renamed with their assertions *inverted deliberately* (refuse -> route) and equally specific; and a **new** `test_mismatched_tokenizer_unsupported_uad_exits_one` preserving the exit-1 path you replaced. Net stronger, nothing dropped.

## One more, mechanical

**Your CHANGELOG entry is at line 138, inside the frozen `## [0.73.3] - 2026-08-18` section**, not under `## [Unreleased]` (line 13) — v0.73.3 shipped while this PR was open and rolled `[Unreleased]` out from under you, and the rebase carried the entry along with the section it was in. Same thing happened to #416 last week. Suffix also needs the house form: `(#304 by @CODING-DARSH in #417)`, currently just `(#417)`.

Non-blocking: widen the `_throughput` stub at `tests/test_v07133.py:2070` to `**kw`, and scope the `SequenceMatcher` spy in the R2 test — patching it module-globally breaks pytest's own assertion-diff repr, so a future failure there prints `TypeError: _spy_matcher() missing 2 required positional arguments` instead of the real diff.

## Where this leaves us

One block deletion, one kwarg, one CHANGELOG move. The measurement work — R2, R3 and its control, R4 — is done and I could not break any of it.

I will file Findings 1 and 5 as follow-ups as you suggested.

MakazhanAlpamys:
Follow-up filed: **#462** — the Q1 boundary-token bias, with the measured table and the fixture as acceptance criteria, so it survives this PR merging.

I did **not** file Finding 1 (the `_FakeTarget` sensitivity band). Re-reading it, it is a readability nit about a test fixture rather than something anyone would pick up and work on, and this repo's rule is that a noisy issues tab hides real bugs. It stays yours to take or leave inside this PR, and if you leave it, nothing is lost.

CODING-DARSH:
Fixed the #409 integration issue as suggested.

Removed the duplicate assisted-throughput call that was running before the incremental report write.
Added assistant_tokenizer=draft_tok to the surviving assisted-throughput call inside the existing #409 flow.
Updated the unsupported-UAD test to verify the assisted arm is gracefully recorded as unmeasured while acceptance/plain throughput are preserved.
Moved the CHANGELOG entry to Unreleased and updated the attribution to (#304 by @CODING-DARSH in #417).

Verified with:

163 passed, 1 skipped across test_v07133.py and test_issue304_cross_tokenizer.py
10 passed for TestDraftMeasureVocabGate
R2/R3/R4 regression checks remain covered.

MakazhanAlpamys:
**The #409 integration is genuinely fixed, and I checked it by breaking it again rather than by reading the diff.** One gap left, and it is small.

## Ordering — correct, and structurally identical to `main`

```
#417 HEAD                      main
766  plain                     737  plain
792  _write_draft_report       763  _write_draft_report
795  assisted (+ 800 tokenizer) 766  assisted
```

One assisted call, after the write, with `assistant_tokenizer=draft_tok` on it and the guard back to a broad `except Exception` preceded by `except KeyboardInterrupt`. Both consequences I reported are gone.

All six previously-failing tests pass (`163 passed, 1 skipped`), and the tests still have teeth — re-introducing the exact previous shape (arm above the write **and** `except RuntimeError`) kills five of the six. So they were not "fixed" by being weakened. Your 12 deleted test lines are all strengthenings: `_FakeTarget` now genuinely implements the repetition penalty, and the two rejection tests were **inverted, not removed**, with the replacements asserting more than the originals (`exit_code == 0` + `uld_strategy: wasserstein_aligned` + `cross-tokenizer`, and the exit-1 path preserved in a new test).

The docs sentence is true again — I checked it empirically rather than by reading, on a same-tokenizer pair: `TOTAL_CALLS: 2 ASSISTED_CALLS: 1`. The native fast path is preserved by the `same_tokenizer` guard at `utils/draft.py:657`.

CHANGELOG is at line 46 inside `[Unreleased]` with the right suffix. `ruff` clean, CI 12/12 on the true head.

## The gap: your headline feature's wiring has no guard

I deleted `assistant_tokenizer=draft_tok` from line 800 and ran your full draft suites:

```
163 passed, 1 skipped in 23.04s
```

Nothing notices. And the regression that lets back in is **precisely consequence (b) from last round**: without that kwarg, a genuine cross-tokenizer pair raises inside `generate`, the broad `except Exception` swallows it, and the user is told "could not be measured" — the original symptom, silently restored.

The reason nothing catches it is that all eleven CLI-level patches are kwarg-blind:

```
tests/test_v07133.py:1787: monkeypatch.setattr(draft_cmd, "measure_throughput", lambda *a, **k: 20.0)
tests/test_v07133.py:1828: monkeypatch.setattr(draft_cmd, "measure_throughput", lambda *a, **k: 15.0)
```

Every test that *does* mention `assistant_tokenizer` calls the library function directly with the kwarg supplied by the test itself, so it proves the library honours it, never that the command passes it.

**One test closes it**: replace one of those `lambda *a, **k` with a spy that records kwargs, and assert `assistant_tokenizer` is among them. The captured shape on your current code is:

```
['assistant_model', 'assistant_tokenizer', 'max_new_tokens', 'num_assistant_tokens']
```

I am asking for this rather than waving it through because it is the second time this exact wire has been the defect — first it was computed and discarded, now it is correct but unguarded. A third time should be caught by CI, not by review.

Nothing else outstanding. One test and this merges.

CODING-DARSH:
Thanks for catching this. Fixed the wiring regression by adding a CLI-level spy that explicitly verifies assistant_tokenizer=draft_tok is forwarded to the assisted measure_throughput call, along with assistant_model, max_new_tokens, and num_assistant_tokens.

I also verified the regression test by temporarily removing assistant_tokenizer=draft_tok — the test fails immediately — so it now catches this exact wiring issue.

Verification:

TestDraftMeasureCli: 11 passed
test_v07133.py + test_issue304_cross_tokenizer.py: 163 passed, 1 skipped
ruff check src/soup_cli/ tests/: clean

The fix is pushed to the branch. Thanks for the thorough review.

MakazhanAlpamys:
Merged as `0194fa4`. **The ask is closed and I verified it with the exact mutation that used to pass** — deleting `assistant_tokenizer=draft_tok` from line 800:

```
before:  163 passed, 1 skipped        (nothing noticed)
now:     1 failed, 162 passed, 1 skipped
         FAILED TestDraftMeasureCli::test_mismatched_tokenizer_measures_cross_tokenizer
         E  KeyError: 'assistant_tokenizer'
```

Your test is a real CLI-level spy, not a restatement: it replaces `draft_cmd.measure_throughput` (the **command module's** symbol), drives it through `runner.invoke(app, ["measure", ...])`, and asserts **identity** — `assisted["assistant_tokenizer"] is draft_tok`. To make `is` able to discriminate you changed `_patch_load` to return named model objects instead of anonymous `object()`s, so a test that passed the *target* tokenizer through would fail. And `assert len(assisted_calls) == 1` pins that the arm is not duplicated, which is the other half of what went wrong last round.

It strengthens an existing test rather than adding a new one, which is the correct shape for this ask.

Also re-confirmed the #409 ordering contract still holds: moving the arm above `_write_draft_report` and narrowing the guard to `except RuntimeError` kills **5** tests.

## Four rounds, and what each actually produced

I want this written down, because from the outside it looks like four rounds of nitpicking and it was not. Each round produced a **control that did not exist yet**:

- **R1** — a `_FakeTarget` that genuinely implements the repetition penalty and re-takes the argmax. Without it the fix was indistinguishable from a no-op: `(3,3)` intact vs `(2,3)` removed only became visible once the fake could actually apply a penalty.
- **R2** — a `_MAX_ALIGN_CHARS` test that is not itself a 300-second wall-clock guard. I offered you a wall-clock bound; you took the monkeypatched cap, which was the better of the two options I gave.
- **R3** — both the `(0,0)` fixture *and* the control that an extendable prompt still counts. A `(0,0)` assertion alone is satisfied by a function that always skips, and the mutant returns alpha `0.0` rather than an error — a wrong number in the direction that makes a usable draft look unusable.
- **R4** — this one.

**The round-four defect deserves a sentence on its own**, because nobody did anything wrong to cause it. #409 merged into `main` underneath an open PR, git resolved it textually clean, and the result ran the assisted arm twice — once before the report write under a guard too narrow for what #409's handlers exist for, once after without `assistant_tokenizer`, whose result overwrote the first. So the UAD measurement was computed and then thrown away, and a real cross-tokenizer pair was told "could not be measured": the original symptom, restored by a merge with no conflict.

## On the record

You said out loud, unprompted, that Findings 1 and 5 should be follow-ups rather than scope creep — and #462 exists because of that. **You also once retitled a PR when told it did not fix the defect its title claimed**, rather than defending the framing. Both are the same instinct and it is rarer than it should be.

Credited in `CONTRIBUTORS.md` (`e0b6967`).

PR #418, opened 2026-08-16
fix(quant): honour training.bnb_4bit_use_double_quant across every 4-bit path (#321)

### LINKED ISSUE #321: training.bnb_4bit_use_double_quant is validated but never read (all 4-bit paths hardcode True)

ousamabenyounes:
## What does this PR do?

Fixes #321 — `training.bnb_4bit_use_double_quant` was a documented, validated
schema field that **nothing read**. Every 4-bit path hardcoded
double-quantization to `True`, so a user could write
`bnb_4bit_use_double_quant: false`, pass validation (there is even a
cross-validator for it), and get double-quant anyway, silently.

Per the issue's "repo-wide or not at all" requirement, the flag is now threaded
through all three call sites **together** so the resident and streamed paths
cannot drift apart:

- **Resident loader** (`utils/quant_menu.py::build_quantization_config_for_loader`)
  — was `"bnb_4bit_use_double_quant": True`, now reads `tcfg.bnb_4bit_use_double_quant`.
- **Layer streaming** (`trainer/stream_setup.py`) — reads the flag **once** and
  passes the **same** value to the sharder (`shard_checkpoint`, whose cache
  already keys on `double_quant`) and to the meta skeleton
  (`build_streamed_model` → `build_nf4_config`). Because both sides read one
  field, the v0.72.2 streamed-vs-resident bit-exactness claim is preserved.
- **4bit save path** (`utils/save_formats.py::merge_4bit`) — gains a
  `double_quant` argument (built via a dict-shaped helper for testability),
  exposed on the CLI as `soup merge --no-double-quant`. `merge` carries no
  training config, so the knob is surfaced as a flag rather than read from a
  `tcfg` that does not exist there.

**Schema default reconciled to `True`** to match what every 4-bit load has
actually shipped, so existing runs are byte-for-byte unaffected. Since the
default is now `True`, the config-load footgun (`=true` requires
`quantization: 4bit`) is gated on `model_fields_set` so it fires only on an
**explicitly set** flag — an unset field on a non-4bit config is no longer
rejected.

## Type of change

- [x] Bug fix

New tests live in the module for each area touched (per the repo's test
convention): `tests/test_quant_menu.py` (resident), `tests/test_v0530.py`
(schema + save path), `tests/test_v07202.py` (streaming).

RED — on the unmodified `main`, the resident loader hardcodes `True`, so a
config that asks for `false` still produces a config with double-quant on:

```
tcfg = self._tcfg(quantization="4bit", bnb_4bit_use_double_quant=False)
>       assert cfg.bnb_4bit_use_double_quant is False
E       assert True is False
FAILED tests/test_quant_menu.py::TestLoaderEntryPoint::test_loader_4bit_honours_double_quant_false
1 failed, 1 passed
```

GREEN — with the fix:

```
tests/test_quant_menu.py ..  (loader honours False + defaults on)   2 passed
tests/test_v0530.py      ..  (default reconciled to True; footgun still fires
                             only on an explicit flag; save-path helper)
tests/test_v07202.py     ..  (streamed builder honours the flag; both stream
                             call sites thread the same tcfg field)
```

The existing v0.72.2 parity test
`test_streamed_quant_config_matches_the_resident_one` still passes (both sides
default `True`).

## Full local suite

- `ruff check src/soup_cli/ tests/` — **All checks passed!** (the blocking CI gate)
- Full local validation of every touched module:
  `pytest tests/test_v0530.py tests/test_quant_menu.py` → **221 passed**;
  `pytest tests/test_v07202.py` → **60 passed / 31 failed / 3 skipped**.
- No previously-green test regressed. The 31 `test_v07202.py` failures are the
  real-NF4 end-to-end / bit-exactness tests, which need bitsandbytes CUDA
  kernels: they fail **identically** on the unmodified `main` in this CPU-only
  environment (base: 31 failed / 58 passed; this branch: 31 failed / 60 passed —
  the +2 are the new streaming tests), i.e. an iso-or-better baseline with zero
  new failures.

## Checklist

- [x] `pytest` — touched modules green; no previously-green test regressed
- [x] Updated relevant docs (`docs/performance-and-quantization.md`) and `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,15 +77,22 @@ reproducing 70+ versions of notes.
   judge-API cost.
 
 - **`training.bnb_4bit_use_double_quant` was validated but never read — every 4-bit
-  path hardcoded double-quantization to `True`, so `bnb_4bit_use_double_quant: false`
-  passed validation and was silently ignored (#321).** The flag is now threaded through
-  all three call sites together — the resident loader (`build_quantization_config_for_loader`),
-  the layer-streaming path (`stream_setup` reads it once and passes the SAME value to the
-  sharder and the meta skeleton, so streamed-vs-resident bit-exactness cannot drift), and the
-  4bit save path (`soup merge --no-double-quant` for the `4bit` / `4bit_forced` save formats).
-  The schema default is reconciled to `True` to match what every 4-bit load has always shipped,
-  so existing runs are unaffected; the config-load footgun (`=true` requires `quantization: 4bit`)
-  now fires only on an explicitly set flag, not on the inherited default.
+  path Soup builds the `BitsAndBytesConfig` for hardcoded double-quantization to `True`,
+  so `bnb_4bit_use_double_quant: false` passed validation and was silently ignored (#321).**
+  The flag is now threaded through the three call sites Soup owns — the resident loader
+  (`build_quantization_config_for_loader`), the layer-streaming path (`stream_setup` reads it
+  once and passes the SAME value to the sharder and the meta skeleton, so streamed-vs-resident
+  bit-exactness cannot drift), and the 4bit save path (`soup merge --no-double-quant` for the
+  `4bit` / `4bit_forced` save formats). The Unsloth loader is out of scope: `FastLanguageModel`
+  builds its own `BitsAndBytesConfig` internally with double-quant hardcoded on and exposes no
+  override, so that path cannot honour the flag. The schema field is now tri-state
+  (`Optional[bool]`, unset = `None`): unset resolves to the shipped default (double-quant on),
+  so a run that never set the flag trains identically, but because the resolved config now
+  carries the field, `soup ship --evidence` provenance and `soup lock check` will report a
+  one-time fingerprint drift on a previously-unset config even though the model numerics are
+  unchanged. The config-load footgun (`=true` requires `quantization: 4bit`) fires only on an
+  explicit `true`, and unset serializes as `None`, so a dumped-and-reloaded config no longer
+  trips it.
 
 - **`kl_control` rewrote the trainer's β/kl_coef on every step, including a `hold`, so a
   non-acting run was numerically identical to `log_only` (#371).** `_run_bang_bang` called
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,17 @@ reproducing 70+ versions of notes.
   block so it is never read as a decode-only number. `--help` states the extra
   judge-API cost.
 
+- **`training.bnb_4bit_use_double_quant` was validated but never read — every 4-bit
+  path hardcoded double-quantization to `True`, so `bnb_4bit_use_double_quant: false`
+  passed validation and was silently ignored (#321).** The flag is now threaded through
+  all three call sites together — the resident loader (`build_quantization_config_for_loader`),
+  the layer-streaming path (`stream_setup` reads it once and passes the SAME value to the
+  sharder and the meta skeleton, so streamed-vs-resident bit-exactness cannot drift), and the
+  4bit save path (`soup merge --no-double-quant` for the `4bit` / `4bit_forced` save formats).
+  The schema default is reconciled to `True` to match what every 4-bit load has always shipped,
+  so existing runs are unaffected; the config-load footgun (`=true` requires `quantization: 4bit`)
+  now fires only on an explicitly set flag, not on the inherited default.
+
 - **`kl_control` rewrote the trainer's β/kl_coef on every step, including a `hold`, so a
   non-acting run was numerically identical to `log_only` (#371).** `_run_bang_bang` called
   `_apply_coefficient` unconditionally; on a `hold` the controller writes back the value
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -29,6 +29,7 @@ soup chat --model ./output                    Interactive chat
 soup push --model ./output --repo user/name   Upload to HuggingFace
 soup push --model ./output --repo user/name --collection user/coll-abc123  Add to HF Collection
 soup merge --adapter ./output                 Merge LoRA with base model
+soup merge --adapter ./output --save-format 4bit --no-double-quant  Save a BNB-4bit merge with double-quantization disabled (default on; #321)
 soup merge-sharded-fsdp-weights ./shards -o merged.safetensors  Consolidate FSDP shards into one safetensors (v0.71.14; --plan-only previews)
 soup delinearize-llama4 ./src --target ./out [--num-experts N] [--plan-only]  Live Llama-4 fused-expert reshape [E*din,dout] -> [E,din,dout] + sidecar copy (v0.71.21)
 soup spectrum scan --model <id|path> --top-percent 50 [--modules mlp,attn] [-o patch.yaml]  Spectrum SNR scan (no model load) -> training.unfrozen_parameters YAML patch (v0.71.23)
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -725,7 +725,7 @@ Cross-validator ordering picks the most actionable error: `quantization_aware='f
 
 ## LF / Axolotl Quant Parity (v0.53.0)
 
-- `bnb_4bit_use_double_quant: true` — requires `quantization: 4bit`. Activates BNB's double-quantization. Combinations with the Quant Menu formats (gptq / awq / hqq:Nbit / aqlm / eetq / mxfp4 / fp8) are rejected at config load.
+- `bnb_4bit_use_double_quant` — controls BNB's double-quantization. **Defaults to `true`** (matching every 4-bit load path — resident, layer-streaming, and the `soup merge` 4bit save formats), and is now honoured everywhere (#321): set `false` to disable it and it actually reaches BNB. Explicitly setting it requires `quantization: 4bit`; combinations with the Quant Menu formats (gptq / awq / hqq:Nbit / aqlm / eetq / mxfp4 / fp8) are rejected at config load.
 - `llm_int8: true` — an explicit 8-bit assertion. Unlike v0.41.0 `load_in_8bit` (which **rewrites** `quantization` to `8bit`), `llm_int8` enforces that the user has ALSO set `quantization: 8bit`. Mismatch raises with an actionable message.
 - `quantize_ref_model: true` / `quantize_reward_model: true` — extend the v0.40.5 Quant Menu wiring to the reference / reward models inside preference and RLHF training. `quantize_ref_model` accepts any task with a reference policy (`dpo / ipo / simpo / orpo / bco / kto / preference / grpo / ppo`); `quantize_reward_model` accepts `ppo / reward_model`.
 
--- a/src/soup_cli/commands/merge.py
+++ b/src/soup_cli/commands/merge.py
@@ -53,6 +53,15 @@ def merge(
             "cycle (v0.53.1 #142)."
         ),
     ),
+    double_quant: bool = typer.Option(
+        True,
+        "--double-quant/--no-double-quant",
+        help=(
+            "BNB double-quantization for the 4bit / 4bit_forced save formats "
+            "(#321). Default on, matching the training loaders; pass "
+            "--no-double-quant to disable. Ignored for the fp16 save format."
+        ),
+    ),
     hub: str = typer.Option(
         "hf",
         "--hub",
@@ -219,6 +228,7 @@ def merge(
                     output_dir=str(output_path),
                     forced=(save_format_canonical == "4bit_forced"),
                     dtype="bfloat16" if dtype == "bfloat16" else "float16",
+                    double_quant=double_quant,
                     trust_remote_code=trc,
                 )
 
--- a/src/soup_cli/config/schema.py
+++ b/src/soup_cli/config/schema.py
@@ -1745,18 +1745,32 @@ class TrainingConfig(BaseModel):
         ),
     )
     # Part E — LF / Axolotl parity.
-    bnb_4bit_use_double_quant: bool = Field(
-        # #321 — every 4-bit load path in the repo has always double-quantized,
-        # so the default is True to match shipped behaviour; the flag is now
-        # honoured (previously validated but never read), letting a user turn it
-        # off. Only meaningful when quantization='4bit'.
-        default=True,
+    bnb_4bit_use_double_quant: Optional[bool] = Field(
+        # #321 — tri-state so an unset flag round-trips cleanly. Every 4-bit load
+        # path has always double-quantized, so `None` (unset) means "use the
+        # shipped default", which each 4-bit consumer resolves to True. A literal
+        # `default=True` here would make `model_dump()` emit the key on EVERY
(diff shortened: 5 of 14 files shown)


MakazhanAlpamys (changes requested):
The core fix is right — I confirmed the claim independently, every 4-bit path did hardcode `True`, and `shard_checkpoint` / `build_streamed_model` already accepted the kwarg, so threading it really was the missing piece. `tests/test_quant_menu.py` + `tests/test_v0530.py` 221 passed, `test_v07202.py` 94 passed here (your 31 failures did not reproduce on this box — environment, not your change), ruff clean.

Three findings. The first is a live regression in two shipped commands.

## 1. Blocking — the default flip breaks `soup sweep` and `soup train --replay`

Flipping the schema default `False → True` means `model_dump()` now emits `bnb_4bit_use_double_quant: true` for **every** config. Re-validating that dict puts the field in `model_fields_set`, which trips the (correctly narrowed) footgun:

```
quantization=none:  round-trip RAISED -> training.bnb_4bit_use_double_quant=true
                                         requires training.quantization='4bit'
quantization=8bit:  same
quantization=4bit:  OK
```

Two shipped paths do exactly that dump-then-revalidate:

- `commands/sweep.py:341` — `config_dict = base_cfg.model_dump()` … `SoupConfig(**config_dict)`
- `commands/train.py:159` `_apply_replay_overrides` — `payload = cfg.model_dump()` … `type(cfg)(**payload)`

So after this, `soup sweep` and `soup train --replay` fail on any config whose `quantization` is not `4bit`, with an error naming a field the user never wrote. **20 bundled recipes, templates and examples declare a non-4bit quantization.** The new `test_double_quant_default_does_not_trip_non_4bit_footgun` tests a YAML load, not the round trip, so it passes.

`Optional[bool] = None` (None = unset, resolved to `True` at the three call sites) fixes it: the footgun then fires only on a literal `true`, and an unset field round-trips as `None`.

## 2. Blocking — the streaming half has no behavioural test

I made the layer-streaming path ignore the config flag while leaving both source strings the grep test looks for intact:

```
315 passed, 0 failed.
```

Nothing caught it. `test_stream_setup_threads_double_quant_from_config` is an `inspect.getsource()` string match — satisfiable while the behaviour is wrong. And its companion `test_build_nf4_config_honours_double_quant_false` is **vacuous**: `build_nf4_config` already honoured its kwarg on `main` (you only edited the docstring), so that test passes unmodified on main.

This matters because streaming is #321's second acceptance criterion, and the issue's own framing was "repo-wide or not at all". Cheap and CPU-only: monkeypatch `shard_checkpoint` and `build_streamed_model` in `stream_setup` and assert both receive `double_quant=False` from a `tcfg` that sets it false.

## 3. Not blocking, but the CHANGELOG says the opposite of what is true

`_config_sha_of` hashes `cfg.model_dump(...)`, so the default flip changes the fingerprint of **every existing config**. Measured on one unchanged YAML:

```
main   config_sha: 7291420b362a952b...
this   config_sha: b5eb214b4843d9dd...
```

`soup ship --config X --evidence ev.json` then refuses pre-upgrade evidence with exit 3, and `soup ci init` wires that as a PR merge gate — so somebody's CI goes red on an unchanged model. Registry `config_hash` / `entry_hash` shift too.

The PR body and CHANGELOG both say "existing runs are unaffected". That is true of the trained weights and **not** of the recipe fingerprint. The `Optional[bool] = None` fix does not remove this either (it is still a dump change), so it should simply be stated rather than fixed.

## Smaller

- **One unthreaded call site**, three lines below where the variable is created: `stream_setup.py:236` calls `estimate_stream_store_bytes(..., quant=quant)` with no `double_quant=`. With `double_quant: false` the early RAM pre-flight charges 0.5156 B/param instead of 0.5625 — a ~9% **under**-estimate, which is the direction this project treats as the dangerous one. One word.
- **`backend: unsloth` still ignores the flag** — `utils/unsloth.py` passes only `load_in_4bit`. Outside #321's scope, but the PR says "every 4-bit path", so it wants either wiring or a sentence.
- `soup merge --no-double-quant` is a new user-facing flag and is not in `docs/commands.md`.

---

Findings 1 and 2 before merge; 3 is a CHANGELOG sentence. The diagnosis and the resident-path fix are both correct — it is the blast radius of the default flip that needs the work.

MakazhanAlpamys (changes requested):
Re-reviewed the force-pushed commit in an isolated worktree with `PYTHONPATH` pinned to your branch.

First, a process note so we do not talk past each other: GitHub keeps only the squashed head, so I could not diff what the force-push changed. I re-ran **every** finding against current state instead. **All six reproduce.** If the push was meant as a rebase rather than as fixes, say so and I will stop treating it as a response — but as it stands nothing in the review has moved.

## 1 (blocking) — the default flip, and I measured the blast radius this time

Still `default=True` at `schema.py:1748`, so `model_dump()` emits the key and re-validating that dict trips the footgun. Both shipped round-trip sites raise:

```
train.py:159  _apply_replay_overrides  -> RAISED (quantization='none')
sweep.py:341  _run_single              -> RAISED (quantization='none')
```

**21 shipped configs break**: 13 bundled recipes (`smollm2-135m-sft`, `qwen2.5-0.5b-sft`, all four `whisper-*`, `falcon-e-bitnet-sft`, …), `templates/embedding.yaml`, and 7 of `examples/configs/`.

`Optional[bool] = None` still fixes it: the footgun then fires only on a literal `true`, and an unset field round-trips as `None`.

## 2 (blocking) — and I can now show *why* the test cannot catch it

Two mutations on `stream_setup.py`, and the pair is the whole story:

- **kill the behaviour, keep both source strings** (`double_quant = True` on the next line) → **503 passed, 0 failed**
- **keep the behaviour identical, change only the spelling** (`bool(tcfg.bnb_4bit_use_double_quant)`) → `test_stream_setup_threads_double_quant_from_config` **FAILS**

It fires on spelling and is silent on behaviour — the exact inversion of what you want from a test.

Its companion is worse: `test_build_nf4_config_honours_double_quant_false` is **vacuous by measurement**. I swapped in `main`'s untouched `layer_stream_runtime.py` and it **passed**. A test that passes against the unfixed file is testing nothing.

Monkeypatching `shard_checkpoint` and `build_streamed_model` in `stream_setup` and asserting both receive `double_quant=False` is CPU-only and closes both.

## New this round — the user-facing flag has zero coverage

`merge.py:231` `double_quant=double_quant` → `True`: **262 passed**. So I ran the control: I renamed the option so `--no-double-quant` **does not exist at all** → **199 passed**.

The flag you are adding can be deleted in its entirety without a single test failing.

## 3 — the CHANGELOG claim is still false, and it is one sentence

Measured again on the same YAML: `main e917707440dab280…` → this PR `cc5945a8fdad52cc…`. The config fingerprint changes, so `soup ship --evidence` provenance and `soup lock check` see drift on an unchanged model. The line still reads "existing runs are unaffected" with no caveat.

## Still open from last round

- `stream_setup.py:236` — `estimate_stream_store_bytes(..., quant=quant)`, three lines below where `double_quant` is created. One word.
- `utils/unsloth.py:50` still passes only `load_in_4bit`, while the title says "every 4-bit path".
- `--no-double-quant` is absent from `docs/commands.md` (only `docs/performance-and-quantization.md` was updated).

## What is genuinely solid

The resident and save-path wiring is properly held: re-hardcoding `True` in `quant_menu.py` and in `save_formats._build_merge_4bit_bnb_kwargs` both **fail** tests, as does reverting the default and dropping the `model_fields_set` narrowing. The diagnosis was right and I confirmed it independently — every 4-bit path did hardcode `True`.

934 passed across own files, source-grep suites and regression suites, 0 failed, ruff clean. One assertion changed (`test_double_quant_default_false` → `_true`), deliberate and documented.

The core fix is correct. What needs work is finding 1's blast radius and finding 2's test *shape* — and the shape matters more than the count, because three of the tests here pass whether or not the code works.

ousamabenyounes:
All six points addressed.

**1 (blocking) — the default flip.** `bnb_4bit_use_double_quant` is now `Optional[bool] = None`. Unset serializes as `None` and round-trips cleanly, so `model_dump()` no longer emits a `true` that re-validation trips — the 21 shipped configs and the `train.py`/`sweep.py` round-trips pass. The footgun fires only on an explicit `true`. A `double_quant_on` property resolves `None`→on, so the resident/streamed/save paths all read a real bool and cannot drift.

**2 (blocking) — the vacuous tests.** Replaced with a behavioural test that drives `_setup_streaming_transformers` and asserts the SAME resolved flag reaches BOTH the sharder and the runtime builder (parametrized over explicit-false and unset). CPU-only: `shard_checkpoint` is spied so its real CPU NF4 packing runs, `build_streamed_model` is stubbed. Hardcoding either call site back to `True` fails the false case; dropping the read fails both.

**New — the `--no-double-quant` flag had zero coverage.** A CliRunner test now drives `soup merge --save-format 4bit --no-double-quant` end-to-end (loads mocked) and asserts `double_quant=False` reaches `merge_4bit`. Renaming the option away now errors the invocation instead of passing silently.

**3 — the CHANGELOG claim.** Rewritten: it no longer says "existing runs are unaffected"; it states the model numerics are identical but the resolved config now carries the field, so `soup ship --evidence` and `soup lock check` will report a one-time fingerprint drift on a previously-unset config.

**Still-open:**
- `stream_setup.py` estimate now threads `double_quant=double_quant`, so the pre-flight RAM size honours the flag.
- `utils/unsloth.py`: `FastLanguageModel.from_pretrained` builds its own `BitsAndBytesConfig` and exposes no double-quant passthrough, so the Unsloth path genuinely cannot honour the flag. Rather than thread a value Unsloth discards, I documented that boundary in the code and narrowed the CHANGELOG/title claim to "every 4-bit path Soup builds the BitsAndBytesConfig for" (resident, streamed, save). Say the word if you would rather I attempt the thread anyway.
- `--no-double-quant` is now in `docs/commands.md`.

Local: `393 passed, 3 skipped` across the affected files, ruff clean.

MakazhanAlpamys:

MakazhanAlpamys:
Merged as [`c382c80`](https://github.com/MakazhanAlpamys/Soup/commit/c382c80) — 13/13 green. You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) as of [`3cce7ce`](https://github.com/MakazhanAlpamys/Soup/commit/3cce7ce), ninth merge.

**Both hard criteria met, and I measured them with controls rather than reading the diff.**

**Round-trip.** Enumerated every shipped config programmatically — `RECIPES`, `src/soup_cli/templates/`, `examples/configs/` — and ran `load → model_dump() → yaml → re-validate`:

| Tree | failures |
|---|---|
| your HEAD | **0 / 173** |
| **control:** your first commit `a99c947` (`default=True`) | **21 / 173** |

The control reproduces my earlier count *exactly* — same 13 recipes, same `embedding.yaml`, same 7 examples. So the harness has teeth and your HEAD scores zero, which is a much stronger statement than "0 failures" alone.

**The house rule holds.** I built the real `BitsAndBytesConfig` objects from a config that never mentions the key. Unset is `None` in the schema, but reaches bitsandbytes as **`True` on all three paths** — resident, streamed, save/merge — while an explicit `false` reaches it as `False`. No existing 4-bit run changes behaviour. `double_quant_on` is exactly the right shape for that, and `Optional[bool] = None` over a plain `True` is the judgement call the whole PR turns on.

**The spelling/behaviour inversion is gone.** Hardcoding `True` while leaving all four source strings intact → **1 failed**. Changing only the spelling so the literal `tcfg.double_quant_on` is absent → **344 passed**. That is precisely the inversion of last round, and deleting the two vacuous tests rather than patching them was the right call.

On `layer_stream_runtime.py`: swapping in `main`'s copy still passes, and that is *correct* — I checked rather than assumed, and the file is **AST-identical to `main` ignoring docstrings**, so the change there is a stale-comment fix and the real coverage lives in `stream_setup`.

**The merge flag now has teeth**: hardcoding its call site → failed; renaming the option so it does not exist → failed. Both passed before.

**CHANGELOG.** Re-measured: the fingerprint does still drift (`23bcf169…` → `4b714dbf…`), and the entry now says so in those words, naming `soup ship --evidence` and `soup lock check`. Accurate.

~1524 passed across your files, the source-grep suites and the streaming regressions; `test_cli_startup_is_light` 23/23; ruff clean.

On Unsloth — your call was right, and narrowing the CHANGELOG claim to "every 4-bit path Soup builds the `BitsAndBytesConfig` for" is the honest scoping rather than an overreach. One follow-up I will file rather than block on: `backend: unsloth` plus an explicit `false` is still **silently ignored**, which is the same shape as the original #321 defect narrowed to one backend. A parse-time footgun rejection would close it, and the precedent is two fields away — `unsloth_bnb_4bit` at `schema.py:1744` is already backend-gated.

That leaves #409, #416, #420 and #421 with findings outstanding.

PR #419, opened 2026-08-16
fix(eval): measure the leg-1 noise floor in the judge task modes (#403)

### LINKED ISSUE #403: soup ship --noise-floor does not measure the leg-1 task axis in judge modes: a win smaller than the judge's own noise still counts
## What

`soup ship --noise-floor N` (v0.73.2, #376) measures the leg-1 task axis **only in `--task-mode metric`**. In `judge_score` and `pairwise` the run prints a warning and leaves leg 1 at a `0.0` floor:

```
Warning: --noise-floor does not measure the leg-1 task axis in --task-mode judge_score
(a judge-backed repeat would report the judge's sampling noise as decode noise);
leg 1 keeps a 0.0 floor.
```

So a judge-scored win smaller than the judge's own sampling noise is still called a win, and leg 1 keeps the exact blindness the flag exists to remove.

## Why it was scoped out rather than done

Not an oversight — a deliberate refusal to publish an inference as a mechanism. Re-running a judge-backed base measurement N times mixes two different sources of variance:

- decode nondeterminism (what the floor is *about*), and
- the judge's own sampling noise (temperature, ordering, rubric drift).

Reporting their sum under a label that says "this is what your instrument can resolve on decode" would be wrong in a way the operator cannot see. Leaving it at 0.0 with a loud warning is the honest placeholder.

## Why it still needs fixing

The scoped-out modes are not a corner: `--task-mode judge_score` and `pairwise` are the two modes for tasks with no exact-match metric, i.e. exactly the open-ended generation tasks where run-to-run variance is largest.

## Proposed fix path

Measure it, but measure the right thing and label it honestly.

For `pairwise` there is a clean instrument already in the codebase: run `_leg1_pairwise(base_gen, base_gen, ...)` — the base model judged against **itself**. The expected value is 0.5 by construction, so the observed spread across N repeats is the combined decode + judge noise, and it is a *directly measured* quantity rather than an inferred one. That removes the objection above: nothing is being attributed to decode alone.

For `judge_score`, `_leg1_judge`'s inner `_score(gen)` closure would need lifting to a module-level helper so the base side can be scored N times without also scoring the tuned side (currently one call does both).

Either way the reported floor should be labelled as **combined decode + judge** noise, distinct from the leg-2 axes' decode-only floors — a `NoiseFloor` that mixes the two kinds silently would be the same failure in a new place.

Cost warning belongs in `--help`: N extra judge passes is N × the API calls, which is a different order of expense from N extra local forward passes.

## Acceptance criteria

- [ ] `--noise-floor N --task-mode pairwise` measures a base-vs-base win-rate floor over N repeats; the warning is gone.
- [ ] `--noise-floor N --task-mode judge_score` measures a base-side floor over N repeats; the warning is gone.
- [ ] The reported floor distinguishes judge-inclusive axes from decode-only ones, in both the panel and `verdict_to_dict` / the evidence block — a reader must not be able to mistake one for the other.
- [ ] `--help` states the API-call cost for the judge modes.
- [ ] A control test proving the judge-mode floor is **not** silently reused as a decode-only number.
- [ ] The existing `metric`-mode behaviour is byte-identical (regression test).

## Refs

- Shipped in v0.73.2 — closes the `metric` half of #376.
- Record: `benchmarks/gate-v0.73.2-leg2-scoring.md` §7 ("What this record does not establish").
- Code: `src/soup_cli/commands/ship.py::_measure_noise_floor`, `_leg1_judge`, `_leg1_pairwise`.

ousamabenyounes:
## What does this PR do?

Fix #403 — `soup ship --noise-floor N` measured the leg-1 task axis **only** in
`--task-mode metric`. In `judge_score` and `pairwise` it printed a warning and
left leg 1 at a `0.0` floor, so a judge-scored win smaller than the judge's own
sampling noise still counted — the exact blindness the flag exists to remove,
on precisely the two modes (open-ended generation) where run-to-run variance is
largest.

The leg-1 task floor is now measured in **every** mode, following the
maintainer's proposed fix path:

- **`pairwise`** — the base model is judged against **itself**
  (`_build_pairwise_scorer` → `winrate(base_gen, base_gen)`). Its expected value
  is `0.5` by construction, so the spread over repeats is a *directly measured*
  combined decode + judge noise, not an inference.
- **`judge_score`** — the inner `_score(gen)` closure of `_leg1_judge` is lifted
  to a module-level `_build_judge_scorer`, so the base side is scored N times on
  its own.

Because the judge repeats fold in the judge's own sampling noise, the floor is
stamped `NoiseFloor.judge_inclusive`: it is labelled **`decode + judge`** on the
panel and in `verdict_to_dict` / the evidence block, so a reader can never
mistake it for the leg-2 axes' decode-only floors. The floor is still keyed
`__task__`, so `decide_ship` widens leg-1 against it automatically. `--help`
now states the extra judge-API cost.

**Backward compatible:** `judge_inclusive` defaults `False` and is emitted into
evidence *only when True*, so every pre-existing (metric / decode-only) evidence
file round-trips byte-for-byte; an unstamped floor reads back as `False`.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pytest tests/ -v` passes (torch-free area module `tests/test_v07302.py`: 178 pass)
- [x] Updated relevant docs (`docs/evaluation.md`, `docs/commands.md`) and `CHANGELOG.md`

New/updated tests live in `tests/test_v07302.py` (the noise-floor area, per the
`#376` lineage). RED was observed on the unmodified target branch (`main`).

**RED — on unmodified `main`**, a `pairwise` `--noise-floor` repro shows the task
axis is never measured:

```
Warning: --noise-floor does not measure the leg-1 task axis in --task-mode
pairwise (a judge-backed repeat would report the judge's sampling noise as
decode noise); leg 1 keeps a 0.0 floor.
noise floor: base repeat 1/2
noise floor: base repeat 2/2
floors: {}
AssertionError: BUG #403: pairwise --noise-floor did NOT measure the leg-1 task axis
```

The new tests (`test_judge_mode_measures_the_leg_one_axis`,
`test_pairwise_floor_measures_the_base_against_itself`,
`test_judge_score_floor_measures_the_base_side`, the labelling/serialization
control, and the `--help` cost assertion) fail on `main` for the same reason
(the task axis is absent / the `judge_inclusive` field does not exist yet).

**GREEN — with the fix:**

```
tests/test_v07302.py .................................................. [100%]
178 passed in 2.82s
```

Full local suite for the touched area (ruff + module) is iso-or-better than the
`main` baseline: baseline `175 passed`, this branch `178 passed`, ruff clean on
both, no new failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,18 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **`soup ship --noise-floor` now measures the leg-1 task axis in the judge modes,
+  so a judge-scored win smaller than the judge's own noise no longer counts (#403).**
+  The floor was measured in `--task-mode metric` only; `judge_score` / `pairwise`
+  printed a warning and left leg 1 at a 0.0 floor, i.e. the exact blindness the flag
+  exists to remove. It is now measured: `judge_score` scores the base side N times
+  through the judge, and `pairwise` judges the base model against itself (expected
+  win-rate 0.5, so the spread is directly measured, not inferred). Because those
+  repeats fold in the judge's own sampling noise, the floor is labelled
+  `decode + judge` on the panel and stamped `judge_inclusive` in the evidence/JSON
+  block so it is never read as a decode-only number. `--help` states the extra
+  judge-API cost.
+
 - **`kl_control` rewrote the trainer's β/kl_coef on every step, including a `hold`, so a
   non-acting run was numerically identical to `log_only` (#371).** `_run_bang_bang` called
   `_apply_coefficient` unconditionally; on a `hold` the controller writes back the value
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -178,7 +178,7 @@ soup ship ... --general-suite mmlu,gsm8k --baseline base.json  lm-eval leg-2 ove
 soup ship ... --emit-evidence ev.json  Re-serialise the scores as replayable --evidence input (output-is-input, #312) (v0.71.39)
 soup ship ... --config soup.yaml  Read eval.ship gate defaults; --evidence GATES on provenance, --emit-evidence STAMPS it (v0.71.39)
 soup ship ... --push owner/repo#N  Post the verdict as a GitHub PR comment (best-effort; never flips the exit code) (v0.71.39)
-soup ship ... --noise-floor N [--task-mode metric]  Re-run base model N times; per-axis floor = max-min spread; gate at max(threshold, floor); leg-1 metric-only (v0.73.2)
+soup ship ... --noise-floor N  Re-run base model N times; per-axis floor = max-min spread; gate at max(threshold, floor); leg-1 measured in every --task-mode (judge modes cost N judge passes, floor labelled decode+judge) (v0.73.2)
 soup card <registry-id> -o MODELCARD.md       HF model card from a registry entry: training config, evals, hashes, lineage, artifacts (v0.71.35)
 soup push --model ./out --repo you/m --card <registry-id>  Upload that registry-driven card as the README (HF only) (v0.71.35)
 soup ci init [--data d.jsonl --suite s.yaml --evidence ev.json] [--config soup.yaml] [--branch main --python 3.11] [--force]  Write .github/workflows/soup-gate.yml: data validate -> expect -> ship gate on every PR (v0.71.35); --config binds the gate to a committed config so it refuses stale evidence (v0.71.39)
--- a/docs/evaluation.md
+++ b/docs/evaluation.md
@@ -428,11 +428,15 @@ actually measurable; a floor *below* it must never tighten the gate behind your 
 exceeds `--forgetting-threshold`, the run says so by name — that axis is now gated looser than you
 asked.
 
-**Scope and cost.** The leg-1 floor is measured in `--task-mode metric` **only**: in the judge modes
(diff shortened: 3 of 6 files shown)


MakazhanAlpamys:
Merged as [`207da67`](https://github.com/MakazhanAlpamys/Soup/commit/207da67) — #403 closed, and this is your seventh.

Both halves survive mutation, which is what I check for:

```
restore the bug (guard the axis write behind task_mode == 'metric')  -> 3 failed
break the labelling only (judge_inclusive = False unconditionally)   -> the same 3 failed
```

So the measurement and the honesty about what it measured are each held by tests, not just the first one.

Two details worth naming:

**`test_pairwise_floor_measures_the_base_against_itself` asserts `gen_a is gen_b` inside the stubbed scorer.** That is a behavioural assertion about identity rather than a comment claiming it — and since `_resolve_generators` does not memoize, base-vs-base genuinely samples twice, so the number means something under decode nondeterminism.

**Emitting `judge_inclusive` into evidence only when true** is the right asymmetry: every pre-#403 evidence file round-trips byte-for-byte, while `verdict_to_dict` always emits it, so the MCP `ship_evidence` reader inherits the marker without a second edit. That preserves the "both readers, neither is the quiet one" property v0.73.2 was built around — and you got it without being told about it.

Hoisting `load_eval_tasks` out of the repeat loop is only safe because `run_eval` does not mutate the task objects; I checked `eval/custom.py` rather than assume it. It does not.

PR #420, opened 2026-08-16
feat(registry): link BOM/attestation to soup card via registry artifact kinds (#309)

### LINKED ISSUE #309: soup card: link BOM / attestation (add them as registry artifact kinds)
## Context

`soup card <registry-id>` (shipped in v0.71.35) renders a provenance-carrying model card from a registry entry: training config, eval scorecard, config/data hashes, lineage, and a table of every **registered artifact**.

But the two most compliance-relevant provenance documents — the **ML-BOM** (`soup bom emit`) and the **in-toto/SLSA attestation** (`soup attest emit`) — are standalone files, **not registry artifact kinds**. `RegistryStore._VALID_KINDS` (`src/soup_cli/registry/store.py`) has no `bom` / `attestation`:

```
adapter, awq, branch_ref, canaries, config, dataset, diagnose_report,
edited_model, eval_results, eval_suite, gguf, gptq, grace_codebook,
judge_calibration, merged, onnx, shrink_report, steering_vector, tensorrt
```

So there is no way to attach them, and the generated card cannot link them. Today [docs/compliance.md](../blob/main/docs/compliance.md) tells you to emit them separately and the card stays silent about them — the one gap in an otherwise self-contained provenance story.

## Proposed fix

1. Add `bom` and `attestation` to `_VALID_KINDS` in `src/soup_cli/registry/store.py`.
2. Add `--attach-to-registry <id>` to `soup bom emit` and `soup attest emit` (mirror the existing pattern in `soup eval custom --attach-to-registry` / `soup export --registry-id`, which call `registry/attach.py::attach_artifact`).
3. `build_model_card` in `src/soup_cli/commands/card.py` already renders every artifact in one table — once the kinds exist and are attachable, they show up for free. Optionally give them a dedicated "Provenance documents" sub-table.

## Acceptance criteria

- [ ] `bom` and `attestation` accepted by `RegistryStore.add_artifact`
- [ ] `soup bom emit ... --attach-to-registry <id>` attaches the emitted `.cdx.json` / `.spdx.json`
- [ ] `soup attest emit ... --attach-to-registry <id>` attaches the statement JSON
- [ ] `soup card <id>` lists them in the artifact table (test in `tests/test_v07135.py` style)
- [ ] `docs/compliance.md` updated so the walkthrough attaches rather than just emits

Good first issue: the pattern is already established by `registry/attach.py` and the existing `--attach-to-registry` flags; no new subsystems.

ousamabenyounes:
## Summary

Fix #309

`soup card <registry-id>` renders every registered artifact, but ML-BOM and
attestation outputs could not be registered. This adds `bom` and `attestation`
artifact kinds and `--attach-to-registry <id>` to `soup bom emit` and
`soup attest emit`.

Signed attestations register both the statement JSON and detached `.sig`
sidecar, so every file required by `soup attest verify` remains reachable from
the registry-backed card. BOM `--format both` similarly registers both emitted
formats. Because attachment was explicitly requested, a lookup/attach failure
now exits 1; omitting the required output path exits 2. Emitted files remain on
disk on a registry failure.

The compliance walkthrough and changelog are updated for the new flow.

Reviewer follow-up RED on the rebased PR head (`e830c59`):

```text
assert ['signed.attest.json'] ==
       ['signed.attest.json', 'signed.attest.json.sig']
1 failed
```

GREEN after attaching every emitted attestation file:

```text
1 passed
```

Attach-failure contract RED on the same prior PR head:

```text
Wrote BOM (cyclonedx) -> failed.bom.json
Warning: could not attach to registry: registry entry not found: missing-entry
assert 0 == 1

Wrote attestation -> failed.attest.json
Warning: could not attach to registry: registry entry not found: missing-entry
assert 0 == 1
```

GREEN after making an explicitly requested attachment part of command success:

```text
2 passed
```

The complete compliance/card area module and changed-line coverage gate:

```text
upstream main: 95 passed
PR before follow-up: 101 passed
PR after follow-up: 109 passed
ruff check src/soup_cli/ tests/: passed
changed production-line coverage: 100% (47/47)
```

The six original tests still provide the initial feature RED → GREEN proof
(6 failed on unmodified main, 6 passed with the feature). The follow-up adds
the signed-sidecar regression plus focused controls for usage and end-to-end
attach-failure exit/file-preservation behavior.

## Files changed

| File | Change |
|------|--------|
| `src/soup_cli/registry/store.py` | accept `bom` and `attestation` artifact kinds |
| `src/soup_cli/commands/bom.py` | attach one or both emitted BOM files |
| `src/soup_cli/commands/attest.py` | attach the statement and signed `.sig` sidecar |
| `docs/compliance.md` | attach walkthrough and verification-file note |
| `CHANGELOG.md` | unreleased feature entry |
| `tests/test_v07135.py` | registry, CLI, failure-path, signature, and card regressions |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,15 @@ reproducing 70+ versions of notes.
   `mlx-lm` import, the PyTorch/TRL training stack is absent, and a one-step real
   CLI SFT run completes. This hardens the runtime boundary; it does not claim to
   resolve the still-unpinned torch-present hang reported in #394.
+- **`soup card` now links the ML-BOM and the in-toto/SLSA attestation — they are
+  first-class registry artifact kinds (#309).** `bom` and `attestation` were the
+  two compliance documents `soup card` could not surface: `RegistryStore` had no
+  such kinds, so there was no way to attach them. Added both to the valid-kind
+  set and a `--attach-to-registry <id>` flag to `soup bom emit` and
+  `soup attest emit` (mirroring the existing `--attach-to-registry` pattern);
+  once attached they appear in the card's artifact table for free. The flag
+  needs `--output` — with none, the document is emitted to stdout and a warning
+  notes nothing was attached.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,9 +49,10 @@ reproducing 70+ versions of notes.
   such kinds, so there was no way to attach them. Added both to the valid-kind
   set and a `--attach-to-registry <id>` flag to `soup bom emit` and
   `soup attest emit` (mirroring the existing `--attach-to-registry` pattern);
-  once attached they appear in the card's artifact table for free. The flag
-  needs `--output` — with none, the document is emitted to stdout and a warning
-  notes nothing was attached.
+  once attached they appear in the card's artifact table for free. Signed
+  attestations also attach the detached `.sig` sidecar. The flag needs
+  `--output` (omitting it exits 2), and a requested registry attachment failure
+  exits 1 after preserving files already emitted.
 
 ### Fixed
 
--- a/docs/compliance.md
+++ b/docs/compliance.md
@@ -64,10 +64,17 @@ soup audit-log rotate        # force a rotation pass
 soup registry push --run-id <run-id> --name my-model --tag v1
 
 soup bom emit --name my-model --base-sha <hex> --config-sha <hex> \
-    --energy energy.json --format both       # CycloneDX + SPDX
+    --energy energy.json --format both -o my-model.bom \
+    --attach-to-registry my-model:v1         # CycloneDX + SPDX, linked to the entry
 soup attest emit --stage train --subject my-model --sha <hex> \
-    --sign ed25519 --key key.pem             # in-toto + SLSA-3
+    --sign ed25519 --key key.pem -o my-model.attest.json \
+    --attach-to-registry my-model:v1         # in-toto + SLSA-3, linked to the entry
 ```
+
+`--attach-to-registry` registers the emitted BOM / attestation as `bom` /
+`attestation` artifacts on the entry, so the model card (step 7) links them
+alongside the other artifacts. It needs `--output`; without it the documents
+are emitted to stdout and nothing is attached.
 
 ## 5. Sign, scan, and verify the artifact
 
@@ -87,7 +94,7 @@ soup airgap-bundle --model ./output --output my-model.tar --repro-receipt receip
 
 Turn the registry entry into a provenance-rich `MODELCARD.md` — base model,
 training config, eval scorecard, config/data hashes, lineage, and every
-registered artifact:
+registered artifact — including the BOM and attestation attached in step 4:
 
 ```bash
 soup card my-model:v1 -o MODELCARD.md
--- a/docs/compliance.md
+++ b/docs/compliance.md
@@ -63,7 +63,8 @@ soup audit-log rotate        # force a rotation pass
 ```bash
 soup registry push --run-id <run-id> --name my-model --tag v1
 
-soup bom emit --name my-model --base-sha <hex> --config-sha <hex> \
+soup bom emit --name my-model --base-model <model-id> \
+    --base-sha <hex> --config-sha <hex> \
     --energy energy.json --format both -o my-model.bom \
     --attach-to-registry my-model:v1         # CycloneDX + SPDX, linked to the entry
 soup attest emit --stage train --subject my-model --sha <hex> \
@@ -72,9 +73,10 @@ soup attest emit --stage train --subject my-model --sha <hex> \
 ```
 
 `--attach-to-registry` registers the emitted BOM / attestation as `bom` /
-`attestation` artifacts on the entry, so the model card (step 7) links them
-alongside the other artifacts. It needs `--output`; without it the documents
-are emitted to stdout and nothing is attached.
+`attestation` artifacts on the entry. Signed attestations include their `.sig`
+sidecar, so the model card (step 7) links every file needed for verification.
+The option needs `--output`; omitting it is a usage error. A registry lookup or
+attachment failure exits non-zero while leaving the emitted files on disk.
 
 ## 5. Sign, scan, and verify the artifact
 
--- a/src/soup_cli/commands/attest.py
+++ b/src/soup_cli/commands/attest.py
@@ -21,6 +21,10 @@ from soup_cli.utils.attest import (
 
 console = Console()
 
+_SIGNATURE_SUFFIX = ".sig"
+_EXIT_ATTACH_FAILED = 1
+_EXIT_USAGE = 2
+
 app = typer.Typer(
     no_args_is_help=True,
     help="In-toto + SLSA-3 attestations per Soup Can stage (v0.59.0).",
@@ -95,9 +99,10 @@ def emit_cmd(
     if output is None:
         if attach_to_registry is not None:
             console.print(
-                "[yellow]Warning:[/] --attach-to-registry needs --output "
-                "(nothing written to attach); skipping attach."
+                "[red]--attach-to-registry needs --output "
+                "(nothing written to attach).[/]"
             )
+            raise typer.Exit(_EXIT_USAGE)
         console.print(text)
         console.print(f"[dim]signature backend: {escape(sig['backend'])}[/]")
         if sig.get("signature"):
@@ -106,8 +111,9 @@ def emit_cmd(
 
     try:
         written = write_attestation(st, output)
+        written_paths = [written]
         if sig.get("backend") == "ed25519" and sig.get("signature"):
-            _write_sig_sidecar(output, sig)
+            written_paths.append(_write_sig_sidecar(output, sig))
     except (TypeError, ValueError) as exc:
         console.print(f"[red]Write failed: {escape(str(exc))}[/]")
         raise typer.Exit(2)
@@ -116,33 +122,36 @@ def emit_cmd(
         f"[dim](signature: {escape(sig['backend'])})[/]"
     )
     if attach_to_registry is not None:
-        _attach_attestation(attach_to_registry, written)
+        _attach_attestation(attach_to_registry, written_paths)
 
 
-def _attach_attestation(registry_id: str, path: str) -> None:
-    """Attach an emitted attestation statement to a registry entry.
+def _attach_attestation(registry_id: str, paths: list[str]) -> None:
+    """Attach an emitted attestation statement and signature to a registry entry.
 
-    Mirrors ``soup shrink --attach-to-registry``: a registry lookup / attach
-    failure is a warning, never fatal to the emit (the statement is on disk).
+    A requested registry attachment is part of command success: lookup or
+    attachment failures exit non-zero after leaving the statement on disk.
     """
     try:
         from soup_cli.registry.attach import attach_artifact
     except ImportError as exc:
         console.print(
-            f"[yellow]Warning:[/] could not import registry attach helper: {escape(str(exc))}"
+            f"[red]Error:[/] could not import registry attach helper: {escape(str(exc))}"
         )
-        return
-    try:
-        attach_artifact(registry_id, path=path, kind="attestation")
-    except Exception as exc:  # noqa: BLE001
-        console.print(f"[yellow]Warning:[/] could not attach to registry: {escape(str(exc))}")
-        return
-    console.print(
-        f"[green]Attached[/] attestation to registry entry [bold]{escape(registry_id)}[/]"
-    )
+        raise typer.Exit(_EXIT_ATTACH_FAILED) from exc
+    for path in paths:
+        try:
+            attach_artifact(registry_id, path=path, kind="attestation")
+        except Exception as exc:  # noqa: BLE001
+            console.print(f"[red]Error:[/] could not attach to registry: {escape(str(exc))}")
+            raise typer.Exit(_EXIT_ATTACH_FAILED) from exc
+        else:
+            console.print(
+                f"[green]Attached[/] attestation to registry entry "
+                f"[bold]{escape(registry_id)}[/]"
+            )
 
 
-def _write_sig_sidecar(output: str, sig: dict) -> None:
+def _write_sig_sidecar(output: str, sig: dict) -> str:
     """Atomic write of the ``<output>.sig`` JSON sidecar (cwd-contained)."""
     from soup_cli.utils.paths import atomic_write_text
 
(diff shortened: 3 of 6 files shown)


MakazhanAlpamys:
Reviewed and **approved on content** — the only thing between this and merge is a `CHANGELOG.md` conflict, caused by four PRs landing ahead of it in the last hour (#407, #408, #410, #419). `git fetch origin && git rebase origin/main`; `CHANGELOG.md` is the only conflicting file and both entries want to be kept.

Verified rather than read — two mutations, both caught:

```
remove 'bom'/'attestation' from _VALID_KINDS   -> 5 failed
disable the attach call site after the write   -> 1 failed (assert [] == ['bom'])
```

The tests are real CLI invocations against a real registry DB, and the card test checks end-to-end rendering rather than just DB state — so it verifies the thing #309 actually asked for, which was that the document becomes reachable *from the card*, not merely that a row exists.

Two decisions worth naming:

**Attach failure warns rather than aborts.** The document is already on disk by then, and losing an emit because the registry link failed would be the wrong trade. Easy to get backwards.

**The `registry.attach` import is lazy at each call site**, so `soup bom emit` does not drag the registry onto the light CLI path. `test_cli_startup_is_light` 23/23.

Push the rebase and I merge on sight — no re-review.

MakazhanAlpamys:
Your push at 15:41 did not clear the conflict — this is still `CONFLICTING / DIRTY`, and `gh pr update-branch` cannot resolve it either.

`main` moved twice more after you rebased: [`e399868`](https://github.com/MakazhanAlpamys/Soup/commit/e399868) (#431) and [`cc15577`](https://github.com/MakazhanAlpamys/Soup/commit/cc15577). Both touch `CHANGELOG.md`, which is what you conflict on. Rebasing onto whatever `main` was at 15:41 was already stale by the time it landed — today has been unusually busy, and that is on the merge rate rather than on you.

Nothing else is blocking. This is still **approved on content**; the review from this morning stands unchanged. Rebase onto current `main`, push, and I will merge once CI is green.

One suggestion given how often you are hitting this: land the `CHANGELOG.md` entry as the **last** thing you touch before asking for merge, or keep your bullet at the bottom of its section rather than the top. Every entry inserted at the top of `### Added` collides with every other one, and you have five PRs in flight competing for the same three lines.

Separately — **#411 has had no new commit since 08-15T20:44** and is also `CONFLICTING`. If you have deprioritised it that is fine, just say so; I would rather know than keep it in the active column. The review there is still waiting on the three items from my comment: a best-of-N measurement, showing the measured number in the refusal, and documenting that `doctor` writes 64 MiB into the working directory.

ousamabenyounes:
Rebased onto the current `main` — the CHANGELOG folds cleanly now (no conflict), and the `#309` diff is unchanged (registry `bom`/`attestation` kinds + the two `--attach-to-registry` flags). Local: `tests/test_v07135.py` 101 passed. Should be unblocked.

MakazhanAlpamys (changes requested):
Re-reviewed from scratch on the rebased head — I deliberately did **not** carry over the earlier "approved on content", since I had not looked at it myself. Verified in an isolated worktree with `PYTHONPATH` pinned to it, because the editable install points at `main` and a worktree run otherwise imports the wrong tree silently.

## The feature works, and I checked that by running it rather than reading it

Seeded a registry entry, ran `bom emit --format both` and `attest emit` with `--attach-to-registry`, then `soup card`:

```
| attestation | my-model.attest.json   | 065d6816d41753d6 |
| bom         | my-model.bom.cdx.json  | be5c65f8c523e6b1 |
| bom         | my-model.bom.spdx.json | 0291202d26f28970 |
```

All five acceptance boxes in #309 tick. The `my-model:v1` ref form from the docs resolves correctly.

**The allowlist stays closed** — `store.py:480` still raises on an unknown kind; the change adds two literals to the frozenset rather than opening it. **Containment holds twice over**: `attach_artifact` takes its `enforce_cwd=True` default and reaches `_is_under` via `os.path.realpath` + `os.path.commonpath` (correct for Windows short names, not `Path.resolve().relative_to()`), and independently the paths passed are return values of `atomic_write_text`, which already refuses anything outside cwd. `_is_adapter` is unaffected — the card still emits `library_name: peft`.

Six mutations, five killed. The one worth naming: mutating `card.py` — **a file this PR does not touch** — still fails your `test_card_lists_attached_bom_and_attestation`. That proves the test checks end-to-end rendering rather than database state, which is the difference between testing the feature and testing the insert.

3261 passed across the PR's files, all 33 source-grep files, and the other `add_artifact` consumers. `test_cli_startup_is_light` 23/23 — the lazy `registry.attach` import inside `_attach_bom` / `_attach_attestation` is the right call. Ruff clean. No deleted or weakened assertions.

## 1. The signature sidecar is never attached

This is the one I would hold on, and I do not think anyone has raised it before.

`attest emit --sign ed25519 -o X --attach-to-registry <id>` writes **two** files and registers **one**:

```
on disk:   signed.attest.json    signed.attest.json.sig
attached:  attestation signed.attest.json
```

`_write_sig_sidecar` (`attest.py:145-158`) writes `<output>.sig`, and `soup attest verify` **requires** it via `--signature` (`attest.py:164`). So the card links an attestation that cannot be verified from the registry alone — and `--sign ed25519 --key key.pem` is exactly what your own `docs/compliance.md` walkthrough tells people to run.

You already solved this identical problem correctly for `bom --format both`, which attaches both the `.cdx.json` and the `.spdx.json`. `_attach_attestation` just takes a single path where `_attach_bom` takes a list.

## 2. Attach failure exits 0 — a decision I want to re-open, not overrule

A typo'd registry id gives a yellow warning and **exit 0**:

| command | bad ref | exit |
|---|---|---|
| `soup card` | `Registry entry not found` | **1** |
| `soup bom emit --attach-to-registry` | `Warning: could not attach` | **0** |
| `soup attest emit --attach-to-registry` | `Warning: could not attach` | **0** |

#309 asked to mirror `soup eval custom --attach-to-registry`, which writes the JSON and then `raise typer.Exit(1)` on attach failure (`eval.py:261-263`). This mirrors `soup shrink` instead.

Worth naming *why* shrink warns rather than exits: its exit codes are already load-bearing (0 = SHIP, 2 = DON'T), so it structurally **cannot** signal an attach failure. `bom` / `attest emit` have no such constraint — 0 is success, 2 is usage, 1 is free.

And the docstring's justification does not quite hold: the document is written **before** the attach, so exiting non-zero loses nothing. As it stands, a compliance pipeline with a typo'd id stays green while the provenance link silently did not happen — which is the gap #309 exists to close.

There is also an internal inconsistency: `bom emit --format both` without `-o` is a hard error at exit 2, while `--attach-to-registry` without `-o` is a warning at exit 0. Both are "you asked for something that needs an output file and gave none."

This was consciously decided and approved earlier, so it is legitimately your call to keep — but I would like the reasoning revisited rather than inherited.

## Optional, while you are in there

`docs/compliance.md:66` omits the required `--base-model`, so that example errors out as written (`Missing option '--base-model'`). **Pre-existing on `main`, not yours** — but you are editing that line.

Also LOW: no test pins the containment-enforcing default at the new call sites (mutating `enforce_cwd=False` there survives, while the store-level control fires). Defence-in-depth only, since `atomic_write_text` already guarantees containment — mentioning it for completeness rather than as a request.

ousamabenyounes:
Addressed both review points in `9a6419f`: signed `--attach-to-registry` registers the statement JSON and detached `.sig` sidecar, and explicitly requested attachment now fails non-zero on missing output or registry errors while preserving files already emitted.

Validation: reproduced the sidecar omission and both CLI exit-code failures on the prior PR head, then passed the focused regressions, the 109-test compliance/card module, repository-wide Ruff, and 100% changed-line coverage. The upstream, pre-follow-up PR, and final local-CI replays all pass; all 12 GitHub checks are green on the exact head SHA.

MakazhanAlpamys:
Merged as [`11eede1`](https://github.com/MakazhanAlpamys/Soup/commit/11eede1) — 12/12 green. Eleventh merge; you are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) as of [`651171c`](https://github.com/MakazhanAlpamys/Soup/commit/651171c).

Both items are fixed, and I ran them rather than reading them.

**The sidecar.** Seeded an entry, generated a real ed25519 PEM, ran the signed emit, rendered the card:

```
| attestation | my-model.attest.json     | 0228f05aabdf977f |
| attestation | my-model.attest.json.sig | a73b559ea89b7eef |
| bom         | my-model.bom.cdx.json    | e21be5a6e06c9247 |
| bom         | my-model.bom.spdx.json   | 816788954ae623af |
```

Then the check that actually matters — I took both paths **out of the registry rows** and fed them to `verify`, with nothing remembered from the emit:

```
soup attest verify <statement-from-registry> --signature <sig-from-registry> --public-key pub.pem
exit=0   Attestation signature valid
```

That is the gap closed: the card now links every file verification needs, which is the whole point of putting them in the registry.

**MEDIUM-2.** Attach failure exits 1 with the file preserved — statement *and* `.sig` on the signed path. `--attach-to-registry` without `-o` exits 2, happy path still 0, and `--format both` without `-o` keeps its own exit-2 message, so the internal inconsistency I flagged is gone too. And it is documented in the docstrings, CHANGELOG **and** `docs/compliance.md` rather than only in this thread — that was the part I cared about, since a decision living in a PR comment is indistinguishable from an accident six months later.

The `--base-model` doc fix rode along, and I ran the documented command verbatim: exit 0.

**Ten mutations, nine killed.** The survivor is a `card.py` heading string no test asserts — cosmetic. Its control in the same file matters more: suppressing the artifact rows in `card.py`, a file this PR does not touch, still fails `test_card_lists_attached_bom_and_attestation`. So that test is genuinely end-to-end rather than checking database state.

Re-confirmed from the first review: the artifact-kind allowlist is still **closed** (a mutation disabling the raise fails a test), and containment is still `os.path.realpath` + `os.path.commonpath` with the Windows-8.3 comment, not `Path.resolve().relative_to()`.

1052 passed on your files plus regressions, 2306 across all 33 source-grep files, `test_cli_startup_is_light` 23/23, ruff clean.

Two cosmetic notes for whenever you are next in the file, neither worth a commit on its own: the `.sig` is registered under `kind="attestation"`, so a consumer filtering that kind gets two rows and disambiguates by extension — defensible, since a third kind would be worse, just worth knowing. And `_attach_attestation` omits the path from its success line, so a signed emit prints the same "Attached attestation" line twice while `_attach_bom` prints the path.

Separately, and not your fault: reviewing #440 surfaced that `tests/test_issue401_*` and `tests/test_issue402_*` — which came in with your #407 and #408 — carry `SPDX-License-Identifier: AGPL-3.0-only` and an "Unsloth AI Inc." copyright pointing at `/studio/LICENSE.AGPL-3.0`, in an Apache-2.0 repo. I merged both without noticing, twice, so that is my miss rather than yours. Stripped in [`8e4f012`](https://github.com/MakazhanAlpamys/Soup/commit/8e4f012) — worth checking your template if those headers come from a snippet you reuse.

ousamabenyounes:
I’m really sorry about the Unsloth citation—it was accidentally carried over from a previous template.

I use my AI agent solution to contribute to multiple open-source projects, including this one, which is how that reference ended up here. I’ve fixed the underlying issue to make sure it doesn’t happen again.

For context, here is an overview of the AI agent solution we use: [https://studio.oratelecom.net/agent/github](https://studio.oratelecom.net/agent/github)

We’re genuinely interested in your project and would be very happy to keep contributing. Thank you for giving us the opportunity to collaborate with you!

ousamabenyounes:
I’m really sorry about the Unsloth citation—it was accidentally carried over from a previous template.

I use my AI agent solution to contribute to multiple open-source projects, including this one, which is how that reference ended up here. I’ve fixed the underlying issue to make sure it doesn’t happen again.

For context, here is an overview of the AI agent solution we use: [https://studio.oratelecom.net/agent/github](https://studio.oratelecom.net/agent/github)

We’re genuinely interested in your project and would be very happy to keep contributing. Thank you for giving us the opportunity to collaborate with you!

MakazhanAlpamys:
Thank you for saying so plainly — that disclosure is worth more to me than the apology, and I want to be equally plain back.

**Agent-assisted contribution is not a problem here.** Eleven of the twenty-four pull requests in v0.73.3 were yours, they found real defects, and the reviews that caught issues in them caught issues I would have had in my own work too. I am not going to pretend to a purity I do not practise; how the code gets written matters far less to me than whether the person sending it can defend it under a mutation test, and you consistently can.

**The licence header is the specific risk, and it is a real one.** Not tidiness — an `SPDX-License-Identifier: AGPL-3.0-only` on a file inside an Apache-2.0 distribution is a licensing *claim* about that file. It does not become true by being wrong, but a downstream redistributor reading it has to act as though it might be, and this project publishes to PyPI on every tag. That is the part that cannot ride along in a template.

**It has now happened three times, and the third one got past me.** Two files in `8e4f012`, which I stripped by hand this morning — and then `3426f2c`, which I merged myself nine hours later, carrying the identical two lines, having already been burned by them that same day. So this is not a lecture aimed at you; my own diff-reading failed on it after I knew.

One detail I noticed and want to name rather than sit on: the header says `See /studio/LICENSE.AGPL-3.0`, and the service you linked is at `studio.oratelecom.net`. That correspondence is strong enough that I think the header is being emitted by the tooling itself rather than copied from an unrelated file — which, if right, means it is fixable once at the source and never again, and is worth checking on your side. I am stating it as an inference, not a finding.

**What I did instead of asking again:** `008da45` adds `tests/test_no_foreign_license_headers.py` — a repo-wide check that fails the build on a foreign SPDX id, a copyleft licence name, an all-rights-reserved line, or a path to someone else's licence file, in the header region of any tracked file. I verified it both directions: re-adding the exact header from `3426f2c` turns it red, removing it turns it green, and the full sweep over 975 tracked files finds exactly zero now.

The reason it is a test and not a paragraph in CONTRIBUTING.md is this project's own track record: the Rich/ANSI `--help` failure was written into CLAUDE.md, CONTRIBUTING and a memory note after each of **four** occurrences and recurred every time, until a test made it fail. Prose has been tried here and it does not hold. It is also deliberately narrow — `utils/license_advisor.py` exists to reason *about* AGPL and must never be flagged for it, which is pinned by a control on the real file, because a guard that fires on correct code gets deleted.

So: nothing to fix retroactively on your side, and please keep contributing. If your tooling can drop that header at the source, it saves us both the CI round trip — but from now on the build tells you, not me.

ousamabenyounes:
We traced the foreign header back to files from our earlier work on Unsloth Studio. It was then mistakenly reproduced in several Soup test files.

We are changing our contribution workflow so every project and every pull request runs in its own isolated sandbox. Work in progress will remain recoverable after a crash, but files and context from one project will never be reused in another. Once a contribution is pushed and completed, its sandbox will be destroyed.

This isolation also reduces token usage. Keeping an ever-growing cross-project context is both unsafe and inefficient: irrelevant history consumes tokens while increasing the risk of transferring conventions or content between repositories.

As people deeply interested in reducing the cost of AI-assisted development—and contributors to tools such as RTK, claude-mem, Graphify, and Caveman—we should apply those same principles to our own workflow: keep only the context required for the current task, persist recoverable work as explicit state, and discard unrelated context.

Thank you for identifying the pattern. This should prevent us from ever polluting one open-source project with material originating from another again.

PR #421, opened 2026-08-16
fix(env): flag installed packages that violate soup's declared bounds (#368)

### LINKED ISSUE #368: pip install vllm silently downgrades the training venv past Soup's own transformers <5.0.0 cap
## What

Installing the `[serve-fast]` extra into a working training environment silently rewrites the training stack. Measured on the H100 box (`benchmarks/gate-h100-validation.md` L3182–3185):

```
pip install vllm
  torch          2.13.0  ->  2.11.0
  transformers   4.57.6  ->  5.14.1
```

The transformers move is **past Soup's own declared cap** (`transformers <5.0.0` in `[train]`), and nothing warns.

## Impact

The user ends up with an environment Soup's own metadata says is unsupported, produced by a Soup-documented command, with no error at any point. The torch downgrade is separately significant on the streaming path, where numerics and kernel dispatch have already been shown to differ across torch majors (#323 is the same class: "CI installs a much newer stack than any dev box").

The project has been burned by this twice — the session's own practice was to dry-run every install in a throwaway venv and verify the training venv afterwards, precisely because of this.

## Fix path

Two parts, and the second is the one that generalises:

1. `soup serve --backend vllm` should pre-flight the *installed* transformers/torch versions against `[train]`'s declared bounds and refuse with the conflict named, rather than failing later or, worse, running on an unsupported stack.
2. `soup env check` already exists and already flags ABI drift — extend it to detect "an installed package violates this distribution's own declared bounds", which catches this and every future instance of the same shape without needing a per-extra rule.

Also worth stating in `docs/`: install `[serve-fast]` in a separate environment from `[train]`, since the two extras have genuinely incompatible resolutions today.

## Acceptance criteria

- [ ] A venv where `transformers` violates the declared cap is detected and reported by `soup env check` (exit 3, consistent with its existing drift behaviour)
- [ ] The detection reads the bound from package metadata, not from a hardcoded copy — a second copy of the cap is exactly the drift this is meant to catch
- [ ] A control: a compliant environment passes, so the check cannot be satisfied by always failing
- [ ] `docs/` states the separate-environment guidance for `[serve-fast]`

ousamabenyounes:
## Summary

Fix #368 — `soup env check` now flags an installed package whose version
violates Soup's **own declared** dependency bound.

`pip install "soup-cli[serve-fast]"` (vllm) into a training venv silently
downgrades `torch` and pushes `transformers` past the `<5.0.0` cap declared in
the `[train]` extra. The user ends up in an environment Soup's own metadata says
is unsupported, produced by a Soup-documented command, with no warning at any
point.

- New pure helper `check_declared_bounds(requirements, installed)` in
  `utils/env_lock.py` audits installed versions against the bounds read from
  the raw PEP 508 requirement strings — i.e. from package metadata via
  `importlib.metadata.requires`, **never a hardcoded copy of the cap** (a second
  copy is exactly the drift this is meant to catch). A not-installed optional
  dependency is not drift; an unparseable requirement/version degrades to
  "skip", never a crash.
- `current_declared_bounds_check()` reads the running distribution's metadata
  and installed versions. An absent distribution degrades to a clean report, so
  `env check` never breaks; a missing `packaging` degrades to clean **and logs a
  warning**, because a checker that reports clean without having checked must
  say so.
- `soup env check` runs the audit **before** the lock read, so the vllm scenario
  is caught even for a user who never ran `soup env lock`. A violation exits `3`,
  consistent with the existing ABI-drift behaviour (now a named
  `DRIFT_EXIT_CODE`).

## Which requirements are enforced

Since v0.71.0's deps-split, `transformers`, `torch` and `trl` exist in metadata
**only** under `extra == "train"/"all"/"dev"`. So a marker gate alone cannot be
the selector — evaluating with `{"extra": ""}` fixes the `wandb` false positive
but also deselects the very packages #368 is about. An extra-gated requirement
is therefore enforced when **both** hold:

1. its canonical name is in `TRACKED_PACKAGES` (which does not list `wandb`), **and**
2. its marker selects under `_TRAINING_INSTALL_EXTRAS = ("train", "all", "dev")`.

Condition 2 is load-bearing, not decoration. Gating on the **name alone**
re-enables `transformers>=5.0.0; extra == "mlx"` alongside `[train]`'s
`<5.0.0`. Those two are deliberately incompatible, so every installed
`transformers` would violate exactly one of them — a false positive on every
machine. `[mlx]` is excluded, pinned by
`test_mlx_only_bound_on_tracked_package_is_not_enforced`.

Measured after the fix (74 requirements rebuilt from `pyproject.toml` the way
pip writes them into `METADATA`):

```
transformers 5.14.1 (#368 case)  ->  ok=False  violations=('transformers',)
transformers/torch/trl 9.9.9     ->  ok=False  violations=('transformers', 'trl')
wandb out of bounds only         ->  ok=True   violations=()      # not tracked, no false positive
compliant training venv          ->  ok=True   violations=()      # control
torch 1.0.0 (downgrade)          ->  ok=False  violations=('torch',)
```

`torch` is absent from the second row on purpose: it is declared `torch>=2.0.0`
with no upper cap, so `9.9.9` genuinely satisfies it.

## Acceptance criteria

- [x] A venv where `transformers` violates the declared cap is detected and
      reported by `soup env check` (exit 3).
- [x] The bound is read from package metadata, not a hardcoded copy.
- [x] A control: a compliant environment passes (the check cannot be satisfied
      by always failing).
- [x] `docs/serving-and-export.md` states the separate-environment guidance for
      `[serve-fast]`.

Tests live in the area module `tests/test_v0640_part_c.py` (the `soup env`
Part C module). Torch-free; run with
`PYTHONPATH=src pytest tests/test_v0640_part_c.py -o addopts=""`.

**Round 1 — the feature.** RED on unmodified `main` (feature absent):

```
tests/test_v0640_part_c.py::test_check_declared_bounds_flags_package_over_cap
E   ImportError: cannot import name 'BoundsCheck' from 'soup_cli.utils.env_lock'
...
7 failed, 59 deselected
```

GREEN with the fix: `7 passed, 59 deselected in 0.90s`.

**Round 2 — the review fixes.** RED on the reviewed commit `2f0d62c`, with only
the new tests applied:

```
FAILED tests/test_v0640_part_c.py::test_extra_gated_tracked_bound_is_enforced
FAILED tests/test_v0640_part_c.py::test_extra_gated_tracked_bound_counted_once_across_extras
FAILED tests/test_v0640_part_c.py::test_declared_bounds_warns_when_packaging_missing
3 failed, 3 passed, 68 deselected in 0.96s
```

GREEN after the fix: `6 passed, 68 deselected in 0.86s`.

## Full local suite

`ruff check src/soup_cli/ tests/` + the full area module, same command on every
side:

| commit | ruff | `tests/test_v0640_part_c.py` |
|---|---|---|
| upstream `main` | clean | 58 passed, 1 failed |
| reviewed `2f0d62c` | clean | 68 passed, 1 failed |
| this branch `ab2f93f` | clean | **74 passed**, 1 failed |

Iso-or-better: the same single failure on every side, never a new one, and every
test added by this change is green. That failure is
`test_cli_env_fix_renders_plan`, pre-existing and unrelated — it reproduces on
unmodified `upstream/main` (Rich truncates the `env fix` panel when stdout is
not a tty, so `"uv pip install"` never appears in the captured output).

## Files changed

| File | Change |
|------|--------|
| `src/soup_cli/utils/env_lock.py` | `BoundViolation`/`BoundsCheck`, `check_declared_bounds`, `current_declared_bounds_check`; `_TRAINING_INSTALL_EXTRAS` scoping; warning when `packaging` is missing |
| `src/soup_cli/commands/env.py` | audit installed deps against declared bounds in `env check`; `DRIFT_EXIT_CODE`; Rich-escaped package names |
| `tests/test_v0640_part_c.py` | 16 new tests (violation, compliant control, uninstalled-skip, data-driven bound, unparseable-skip, CLI exit 3, extra-gated enforcement, `[mlx]` non-regression, cross-extra dedup, Rich escaping, missing-`packaging` warning) |
| `docs/serving-and-export.md`, `docs/commands.md` | separate-environment guidance for `[serve-fast]`; env-check note |
| `CHANGELOG.md` | `### Fixed` entry |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,10 +86,6 @@ reproducing 70+ versions of notes.
   is wrong; it prints what it overrode beside what was detected. Disk detection
   may write a small scratch file next to the streamed shards to run the probe.
 
-### Fixed
-
-### Fixed
-
 - **The one-active-execution cap could double-book after a server restart (#402).**
   `ExecutionManager._active_run_id` is in-memory, so a restarted MCP server
   started with an empty slot, saw free capacity, and would launch a second
@@ -99,8 +95,6 @@ reproducing 70+ versions of notes.
   new execution across restarts, while a stale record whose process is gone
   frees the slot rather than wedging it shut. `docs/commands.md` now states the
   actual scope of the cap.
-
-### Fixed
 
 - **`soup ship --noise-floor` now measures the leg-1 task axis in the judge modes,
   so a judge-scored win smaller than the judge's own noise no longer counts (#403).**
@@ -131,6 +125,16 @@ reproducing 70+ versions of notes.
   unchanged. The config-load footgun (`=true` requires `quantization: 4bit`) fires only on an
   explicit `true`, and unset serializes as `None`, so a dumped-and-reloaded config no longer
   trips it.
+- **`soup env check` now flags an installed package that violates Soup's own
+  declared version bound (#368).** `pip install "soup-cli[serve-fast]"` (vllm)
+  into a training venv silently downgrades `torch` and pushes `transformers`
+  past the `<5.0.0` cap declared in `[train]`, producing an environment Soup's
+  own metadata says is unsupported with no warning at any point. `env check`
+  now audits the installed versions against the bounds read from package
+  metadata — not a hardcoded copy, since a second copy of the cap is exactly the
+  drift this catches — and exits 3 when one is violated, independent of any lock
+  file. `docs/serving-and-export.md` states the separate-environment guidance
+  for `[serve-fast]`.
 
 - **`kl_control` rewrote the trainer's β/kl_coef on every step, including a `hold`, so a
   non-acting run was numerically identical to `log_only` (#371).** `_run_bang_bang` called
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,14 +127,21 @@ reproducing 70+ versions of notes.
   trips it.
 - **`soup env check` now flags an installed package that violates Soup's own
   declared version bound (#368).** `pip install "soup-cli[serve-fast]"` (vllm)
-  into a training venv silently downgrades `torch` and pushes `transformers`
-  past the `<5.0.0` cap declared in `[train]`, producing an environment Soup's
-  own metadata says is unsupported with no warning at any point. `env check`
-  now audits the installed versions against the bounds read from package
-  metadata — not a hardcoded copy, since a second copy of the cap is exactly the
-  drift this catches — and exits 3 when one is violated, independent of any lock
-  file. `docs/serving-and-export.md` states the separate-environment guidance
-  for `[serve-fast]`.
+  into a training venv silently pushes `transformers` past the `<5.0.0` cap Soup
+  declares, producing an environment Soup's own metadata says is unsupported with
+  no warning at any point. `env check` audits installed versions against the
+  bounds read from package metadata — not a hardcoded copy, since a second copy
+  of the cap is exactly the drift this catches — and exits 3 when one is violated,
+  independent of any lock file. Only the CORE bounds (those that apply to every
+  install) are enforced: an `extra == "X"` requirement is Soup's bound only when
+  `[X]` was opted into, which the running environment cannot confirm, so its
+  marker is evaluated and it is skipped rather than raised as a false positive,
+  and a package is counted once even when its bound is restated under several
+  extras. The bounds audit no longer short-circuits the lock diagnostic (both
+  print). `packaging` is a declared dependency now — the audit degraded to a
+  silent "clean" without it, the wrong direction for a checker.
+  `docs/serving-and-export.md` states the separate-environment guidance for
+  `[serve-fast]`.
 
 - **`kl_control` rewrote the trainer's β/kl_coef on every step, including a `hold`, so a
   non-acting run was numerically identical to `log_only` (#371).** `_run_bang_bang` called
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,12 +132,17 @@ reproducing 70+ versions of notes.
   no warning at any point. `env check` audits installed versions against the
   bounds read from package metadata — not a hardcoded copy, since a second copy
   of the cap is exactly the drift this catches — and exits 3 when one is violated,
-  independent of any lock file. Only the CORE bounds (those that apply to every
-  install) are enforced: an `extra == "X"` requirement is Soup's bound only when
-  `[X]` was opted into, which the running environment cannot confirm, so its
-  marker is evaluated and it is skipped rather than raised as a false positive,
-  and a package is counted once even when its bound is restated under several
-  extras. The bounds audit no longer short-circuits the lock diagnostic (both
+  independent of any lock file. An `extra == "X"` requirement is Soup's bound
+  only when `[X]` was opted into, which the running environment cannot confirm,
+  so its marker is evaluated and it is skipped rather than raised as a false
+  positive — except for the ABI-relevant `TRACKED_PACKAGES` set, which since the
+  v0.71.0 deps-split lives in metadata *only* under `extra == "train"/"all"/
+  "dev"` and therefore includes the `transformers <5.0.0` case #368 was reported
+  about. A package is counted once even when its bound is restated under several
+  extras. If `packaging` is somehow unimportable the audit still degrades to a
+  clean report so `env check` keeps working, but now says so rather than
+  reporting a silent clean it never verified. The bounds audit no longer
+  short-circuits the lock diagnostic (both
   print). `packaging` is a declared dependency now — the audit degraded to a
   silent "clean" without it, the wrong direction for a checker.
   `docs/serving-and-export.md` states the separate-environment guidance for
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -259,7 +259,7 @@ soup tunability --dataset <jsonl> [--candidates a,b,c]   Probe candidate bases +
 soup tunability --dataset <jsonl> --live [--device cpu]  LIVE per-candidate LoRA probe (loads each repo)
 soup plan --config soup.yaml                             Pre-flight summary + write soup.tfstate
 soup apply --config soup.yaml [--dry-run]                Lock-and-execute; refuses on drift (exit 3)
-soup env lock | status | check                           Hermetic env lockfile + ABI drift detection (exit 3)
+soup env lock | status | check                           Hermetic env lockfile + ABI drift + declared-bound violation detection (exit 3)
 soup env fix [--format uv-pip|requirements] [--output req.txt]  Render a reproducible install plan from soup-env.lock (print-only)
 soup completions bash | zsh | fish                       Shell completion script (sourceable via eval)
 soup license-advisor --target b2c|defense|embedded       Recommend license-clean base for deploy target
--- a/docs/serving-and-export.md
+++ b/docs/serving-and-export.md
@@ -251,8 +251,15 @@ soup serve --model ./output --kv-cache-type q8_0   # 8-bit quantized cache (need
 
 Use [vLLM](https://github.com/vllm-project/vllm) for significantly better throughput in production:
 
+> **Install `[serve-fast]` in a separate environment from `[train]`.** Their
+> resolutions are genuinely incompatible today — `pip install vllm` into a
+> training venv silently downgrades `torch` and pushes `transformers` past
+> Soup's own `<5.0.0` cap. `soup env check` now flags an installed package that
+> violates this distribution's declared bounds (exit code 3), so you can catch a
+> contaminated venv before a run relies on it.
+
 ```bash
-# Install vLLM support
+# Install vLLM support (in its own environment, not the training one)
 pip install "soup-cli[serve-fast]"
 
 # Start with vLLM backend
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,11 @@ dependencies = [
     "pyyaml>=6.0",
     "huggingface-hub>=0.16.0",
     "plotext>=5.2.0",
+    # `env check` reads declared bounds via packaging (Requirement/marker/version).
+    # It ships with pip and is pulled in transitively, but `env check` degrades to
+    # a silent "clean" without it — the wrong direction for a checker — so it is
+    # declared explicitly (#368 review).
+    "packaging>=21.0",
 ]
 
 [project.optional-dependencies]
--- a/src/soup_cli/commands/env.py
+++ b/src/soup_cli/commands/env.py
@@ -135,25 +135,12 @@ def env_check_cmd(
     # transformers/torch downgrade even for a user who never ran `soup env lock`
     # (#368). The bound is read from package metadata, not a hardcoded copy.
     bounds = current_declared_bounds_check()
-    if not bounds.ok:
-        rows = "\n".join(
-            f"- {escape(v.name)} {escape(v.installed)} violates declared bound "
-            f"{escape(v.specifier)}"
-            + (f" (extra: {escape(v.extra)})" if v.extra else "")
-            for v in bounds.violations
-        )
-        console.print(
-            Panel(
-                f"[red]{bounds.violation_count} package(s) violate this "
-                f"distribution's own declared bounds:[/]\n{rows}\n\n"
-                "[yellow]Install '[serve-fast]' in a separate environment from "
-                "'[train]' — their resolutions are incompatible today.[/]",
-                title="env check",
-                border_style="red",
-            )
-        )
-        raise typer.Exit(DRIFT_EXIT_CODE)
 
+    # #368 review finding 5 — run the lock diagnostic REGARDLESS of the bounds
+    # outcome, so a bounds violation no longer hides "no lock file" / ABI drift.
+    # `lock_exit` records the lock half's exit code (None = clean); the bounds
+    # violation takes precedence at the end but is printed after the lock line.
+    lock_exit = None
     try:
         locked = read_lock(lock_path)
     except FileNotFoundError:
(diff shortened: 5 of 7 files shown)


MakazhanAlpamys (changes requested):
The pure helper is well built and its unit tests are real. Four findings on the production path, and the first is the one that decides the PR.

`tests/test_v0640_part_c.py` 65 passed / 1 skipped, ruff clean, light-CLI guard 23/23.

## 1. Blocking — the real-environment path has no test at all

I made `current_declared_bounds_check()` return a clean `BoundsCheck` unconditionally — the entire detection path dead, the feature a total no-op:

```
65 passed, 1 skipped.  ZERO failures.
```

The only test touching it passes `"this-distribution-does-not-exist"` and asserts it returns clean, which the mutant also does; the CLI test monkeypatches the function away. So nothing verifies that reading real distribution metadata against real installed versions ever detects anything. Shipped silently broken, CI stays green.

For contrast the *pure* function is properly held: mutating `if satisfied: continue` to always-continue fails two tests. The gap is precisely at the boundary where the code meets the machine.

## 2. Blocking — false positives, because environment markers are parsed but not evaluated

`check_declared_bounds` reads `extra == "..."` only to **name** the extra; it never evaluates applicability. So every requirement across all 88 metadata lines is checked against every installed package. Measured against real `requires("soup-cli")`:

- **wandb 0.19.1**, installed independently — extremely common in an ML venv, and the user never asked for `[wandb]` — is flagged against `wandb<0.18.0; extra=='wandb'`, and **`soup env check` exits 3**.
- **textual** is flagged against the `[tui]` bound nobody opted into.

`python_version` and platform markers are equally unevaluated.

This defeats the issue's own third acceptance criterion, *"a compliant environment passes"*. The PR's control for that uses a synthetic `_TRAIN_REQS`; there is no control that a **real** environment passes. My box happens to pass only because it has no wandb.

## 3. Blocking — the one actionable sentence renders empty

Through the real CLI:

```
│ Install '' in a separate environment from '' — their resolutions are   │
│ incompatible today.                                                    │
```

`'[serve-fast]'` and `'[train]'` are unescaped inside a markup-enabled Panel, so Rich reads them as unknown tags and deletes them. The remediation line — the single most useful output of the feature — tells the user to install nothing in a separate environment from nothing.

This is the `[no win]` defect from v0.73.2, documented in `CLAUDE.md` for exactly this reason. The test asserts only `"transformers" in result.output`, so it passes over it. `escape()` or `\[serve-fast]`.

## 4. Blocking — the headline count triple-counts

```
ok: False   violation_count: 3
   transformers 5.14.1 violates <5.0.0,>=4.36.0 (extra=all)
   transformers 5.14.1 violates <5.0.0,>=4.36.0 (extra=dev)
   transformers 5.14.1 violates <5.0.0,>=4.36.0 (extra=train)
```

`transformers<5.0.0` appears under `all`, `dev` and `train`, so the panel says **"3 package(s) violate…"** when one does. The test uses a one-line synthetic requirement set and asserts `violation_count == 1`, so it cannot see it. Dedup by canonical name.

## 5. Worth reconsidering — the audit short-circuits before the lock is read

The bounds check runs ahead of `read_lock` and exits 3, so any violation hides ABI drift and the "no lock file" message until it is resolved. Combined with finding 2, a stale `[wandb]` cap permanently masks the lock-drift diagnostic — the thing `env check` primarily exists for. The serve-fast-specific guidance also prints for unrelated violations.

## 6. Minor

- ~158 ms per `env check` (88 `importlib.metadata` lookups). Worth knowing, not worth fixing yet.
- It reads **installed** metadata, so an editable checkout is stale: this box reports `soup-cli 0.71.39` with `trl>=0.7.0` while the tree is v0.73.2 with `trl<0.29`. A newly tightened cap is not enforced until reinstall — worth a docstring line, since a maintainer tightening a bound will test in exactly that setup.

---

Findings 1 and 2 are the ones I would not merge without. The others are a morning's work between them. The idea is right and #368 is a real trap — a `pip install vllm` that quietly walks `transformers` past our own cap is precisely the kind of thing nobody notices until a training run behaves oddly.

MakazhanAlpamys (changes requested):
Re-reviewed the force-pushed commit in an isolated worktree with `PYTHONPATH` pinned to your branch.

Same process note as on #418: GitHub keeps only the squashed head, so I re-ran every finding rather than diffing. The CHANGELOG cleanup (three stray duplicate `### Fixed` headings) and the new `docs/serving-and-export.md` guidance both land and are good. **The four production-path findings all reproduce.**

## What you got right, and it is the decision that matters most

The bound comes from **one** place — package metadata — with no hardcoded second copy. That is the correct shape, and it is the thing this project has been bitten by repeatedly: two copies of "what the user typed" in #372, two copies of "which modules did we train" in #392. You avoided it.

Exit code is a named `DRIFT_EXIT_CODE = 3`, shared with ABI drift and documented in `docs/commands.md`. Breaking it to `1` fails 2 tests, so it is held.

## 1 (blocking) — the real-environment path is a no-op away from untested

I made `current_declared_bounds_check()` return a clean `BoundsCheck` unconditionally — the entire detection path dead, the feature a total no-op:

**121 passed, 0 failed.**

Control in the same file: breaking `req.specifier.contains` fails 2 tests. So the pure helper is properly held, and the gap is precisely where the code meets the machine — which is the half #368 is actually about. `pip install vllm` silently downgrading `transformers` is a *live environment* event; a lockfile parser cannot see it.

## 2 (blocking) — markers are named but never evaluated

`env_lock.py:265` uses `_EXTRA_MARKER_RE` only to *name* the extra; there is no `marker.evaluate()`. Against the real `requires("soup-cli")` (88 lines):

```
ok: False   violation_count: 4
   transformers 5.14.1 violates <5.0.0,>=4.36.0 (extra=all)
   transformers 5.14.1 violates <5.0.0,>=4.36.0 (extra=dev)
   transformers 5.14.1 violates <5.0.0,>=4.36.0 (extra=train)
   wandb 0.19.1 violates <0.18.0,>=0.15.0 (extra=wandb)
```

That single output contains two findings: **wandb is flagged against an extra nobody opted into** (a false positive that exits 3), and **one package produces three rows while the count says 4** (finding 4 — no canonical-name dedup).

## 3 — the remediation sentence renders empty

Through the real CLI:

```
Install '' in a separate environment from '' — their resolutions are incompatible today.
```

You added `escape()` to the violation rows, but the `[serve-fast]` / `[train]` literals in the remediation line are still raw inside a markup Panel, so Rich eats them.

And those new `escape()` calls are themselves untested: stripping **all** of them passes 121 tests, while `DRIFT_EXIT_CODE 3 → 1` in the same file fails 2. So the file is testable and that specific hardening is not covered.

## 5 — still short-circuits before `read_lock`

`env.py:133` runs the bounds check ahead of `read_lock`. In the run above there was no lock file at all and the lock diagnostic never printed.

## Two smaller notes

`packaging` is relied on but not declared in `pyproject.toml`. The `ImportError` fallback degrades to **"clean"** — the silent direction, which for a checker is the wrong way to fail.

Metadata staleness reproduces on an editable checkout: metadata says `soup-cli 0.71.39` with `trl>=0.7.0` while the tree is v0.73.2 with `trl>=0.14.0,<0.29`. So a maintainer tightening a cap will not see it enforced until reinstall — worth one docstring line. Cost measured at ~181 ms per check.

## Numbers

216 passed / 7 skipped across `test_v0640_part_c`, `test_v0640_followups`, `test_v0670_part_e`, `test_cli_startup_is_light` (23/23) and `test_bugfixes`. No source-grep test references `env_lock`. Ruff clean. **No assertions deleted or weakened.** Docs are the strongest part of this PR.

Findings 1 and 2 are what I would want before merge. The idea is right and the metadata-sourced bound is the correct foundation.

ousamabenyounes:
Findings 1, 2, 3 and 5 are done, plus the two smaller notes.

**2 (blocking) — markers are now evaluated.** `check_declared_bounds` evaluates the marker in the base environment (`{"extra": ""}`): an `extra == "X"` requirement is Soup's declared bound only when `[X]` was opted into, which the running env cannot confirm — so `wandb` under `extra == "wandb"` is no longer a false positive, and the core package re-listed under `all`/`dev`/`train` collapses to its one un-gated bound. Added canonical-name dedup on top. Removed the now-dead `_EXTRA_MARKER_RE`. Two new tests pin both halves (`...not_opted_into_is_not_flagged`, `...counted_once_despite_extra_duplicates`).

**1 (blocking) — the real path is tested.** `test_current_declared_bounds_check_real_path_detects_violation` drives it end to end with `importlib.metadata.requires` and the version lookup mocked; making the function a no-op or dropping the lookup fails it.

**3 — the remediation.** The empty line was `[serve-fast]`/`[train]` being eaten as Rich markup inside a Panel. Rewritten to an accurate, bracket-free message ("a later install likely moved these packages out of range; reinstall or pin them").

**5 — no more short-circuit.** `env check` now runs the lock diagnostic AND the bounds audit and prints both, so a bounds violation no longer hides "no lock file" / ABI drift. Exit codes are unchanged (no lock = 1, parse = 2, drift/bounds = 3) — the CLI test now also asserts the "no lock" line is present alongside the exit 3.

**Smaller:** `packaging>=21.0` is a declared dependency now (the audit degraded to a silent "clean" without it); the editable-checkout metadata-staleness caveat is one docstring line.

Because extra-gated bounds are no longer enforced, I updated the unit fixtures from extra-only strings to core bounds — the behaviour they asserted was the false-positive you flagged. Local: `131 passed` across `test_v0640_part_c` / `test_v0640_followups` / `test_v0670_part_e`, `95 passed` on startup+bugfixes, ruff clean.

MakazhanAlpamys (changes requested):
Findings 1, 3 and 4 are properly closed, and I verified each by mutation rather than by reading:

- No-op'ing `current_declared_bounds_check` now **fails** `test_current_declared_bounds_check_real_path_detects_violation`. It survived last round — this is the blocker closed.
- The remediation panel renders in full. Dropping the `[serve-fast]` / `[train]` literals rather than escaping them was the right call.
- `commands/env.py:137-197` runs the lock diagnostic regardless of the bounds outcome; confirmed live — both "No lock file…" and the bounds panel print, exit 3.

## One blocker, and it is new — the fix removed the feature's own motivating case

Evaluating markers with `{"extra": ""}` does fix the `wandb` false positive. But **since v0.71.0's deps-split, `transformers`, `torch` and `trl` exist in metadata *only* under `extra == "train"/"all"/"dev"`** — so they are deselected too.

Measured against real installed metadata:

```
transformers 5.14.1 installed   ->  ok=True, violation_count=0
transformers/torch/trl 9.9.9    ->  ok=True, violation_count=0
```

88 requirements, 6 un-gated, and of those **only `typer<0.21.0` carries an upper cap**. So as it stands the check can only ever fire on `typer`.

That means the `transformers <5.0.0` case **#368 was filed about is now unreachable** — and `docs/serving-and-export.md:254` promises exactly that case: *"`pip install vllm` … pushes `transformers` past Soup's own `<5.0.0` cap. `soup env check` now flags…"*. That sentence is measurably false against the shipped code.

I want to be clear this is not a criticism of the direction. Your marker evaluation is correct in itself; the deps-split just means "un-gated" and "core" stopped being the same thing, and the fix inherited that assumption.

**Suggestion using machinery already in the file**: enforce extra-gated bounds only for names in `TRACKED_PACKAGES` (`env_lock.py:55`). It lists `transformers` / `torch` / `peft` / `trl` / `accelerate` and does **not** list `wandb` — so it catches #368 and keeps your false-positive fix.

The test fixture encodes the same assumption: `_CORE_TRANSFORMERS` (`tests/test_v0640_part_c.py:735`) is commented *"A CORE requirement (no extra marker)"*, but no un-gated `transformers` row exists in real metadata. Worth re-labelling so the fixture does not imply coverage it does not have.

## Three smaller items

**The dedup guard is dead code.** `test_core_bound_counted_once_despite_extra_duplicates` passes with the `seen` block at `env_lock.py:266-268` deleted — marker evaluation already drops the duplicates. Either remove it, or add a case where a package legitimately appears twice un-gated.

**The escaping is unasserted.** Stripping every `escape()` in `commands/env.py` still passes; the control (`DRIFT_EXIT_CODE 3 → 1`) fails 2, so the file is exercised. One test with `[bold]` in a package name closes it.

**`packaging` in core deps is right**, but the three `ImportError` fallbacks (`env_lock.py:230, 302, 313`) still return **"clean"**. Now that it is a declared dependency, a warning is the safer direction — for a checker, failing silently clean is the wrong way to fail.

ruff clean, 124 + 308 tests green, no assertions removed. Only `CHANGELOG.md` conflicts with `main`.

ousamabenyounes:
Thanks for the mutation-based verification — the blocker was real, and the diagnosis was exact. Addressed in `ab2f93f`.

## The blocker: the feature's own motivating case

Your `TRACKED_PACKAGES` suggestion is what shipped, with one scope narrowing that turned out to be load-bearing. An extra-gated requirement is now enforced when **both** hold:

1. its canonical name is in `TRACKED_PACKAGES` (`wandb` is not in it, so the false-positive fix stands), **and**
2. its marker selects under `_TRAINING_INSTALL_EXTRAS = ("train", "all", "dev")`.

Condition 2 is not decoration. Gating on the **name alone** re-enables `transformers>=5.0.0; extra == "mlx"` alongside `[train]`'s `<5.0.0`. Those two are deliberately incompatible, so every installed `transformers` would violate exactly one of them — a false positive on every machine, and a worse regression than the one being fixed. `[mlx]` is therefore explicitly excluded, with a regression test pinning it: `test_mlx_only_bound_on_tracked_package_is_not_enforced`.

Your two measurements, rerun after the fix (74 requirements rebuilt from `pyproject.toml` the way pip writes them into `METADATA`):

```
transformers 5.14.1 (#368 case)  ->  ok=False  violations=('transformers',)
transformers/torch/trl 9.9.9     ->  ok=False  violations=('transformers', 'trl')
wandb out of bounds only         ->  ok=True   violations=()      # your v1 false positive, still fixed
compliant training venv          ->  ok=True   violations=()      # control
```

`torch` is absent from the second row on purpose: it is declared `torch>=2.0.0` with no upper cap, so `9.9.9` genuinely satisfies it. A downgrade does fire — `torch 1.0.0 -> ['torch>=2.0.0']`.

`docs/serving-and-export.md:254` is true against the shipped code again.

## The three smaller items

- **Dedup guard** — kept, and it is no longer dead. Metadata restates the tracked bound under `train`, `all` **and** `dev`, so deleting the `seen` block now fails two tests rather than none.
- **Escaping** — asserted. A package name containing `[bold]` now has a test; stripping `escape()` fails it.
- **`ImportError` fallbacks** — a missing `packaging` logs a warning before degrading to a clean report; "silently clean" was indeed the wrong failure direction. The `importlib.metadata` guard was deleted outright instead: `requires-python = ">=3.10"` makes it unreachable, so it was an unreachable branch that silently reported clean.
- **Fixture label** — `_CORE_TRANSFORMERS` → `_UNGATED_TRANSFORMERS`, with a comment stating that no un-gated `transformers` row exists in real metadata and what the fixture therefore does and does not cover.

## Validation

RED on the reviewed commit `2f0d62c`, with only the new tests applied: **3 failed, 3 passed** — `test_extra_gated_tracked_bound_is_enforced`, `test_extra_gated_tracked_bound_counted_once_across_extras`, `test_declared_bounds_warns_when_packaging_missing`. GREEN after the fix: **6 passed**.

Full local replay of the CI job classes (`ruff check src/soup_cli/ tests/` + `tests/test_v0640_part_c.py` + the top-level-torch import check), same command both sides:

| | ruff | tests |
|---|---|---|
| baseline `2f0d62c` | clean | 68 passed, 1 failed |
| `ab2f93f` | clean | **74 passed**, 1 failed |

Iso-or-better: same single failure on both sides, +6 passing. That failure is `test_cli_env_fix_renders_plan`, pre-existing and unrelated — it reproduces identically on unmodified `upstream/main` (Rich truncates the `env fix` panel when stdout is not a tty, so `"uv pip install"` never appears in the captured output).

Also rebased onto `main`; the `CHANGELOG.md` conflict you flagged is resolved.

MakazhanAlpamys:
Merged as [`bce3f48`](https://github.com/MakazhanAlpamys/Soup/commit/bce3f48) — 12/12 green. You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) as of the follow-up commit, tenth merge.

**The blocker is closed, and — the part that mattered — (a) and (b) hold simultaneously.** Driven by real installed metadata (88 requirements), with only the version lookup simulated:

| case | ok | violations |
|---|---|---|
| transformers 5.14.1 (the #368 case) | `False` | `('transformers',)` |
| wandb 0.99.0 outside the `[wandb]` cap | `True` | `()` |
| **both present at once** | `False` | `('transformers',)` |
| torch 1.0.0 (floor control) | `False` | `('torch',)` |
| compliant venv (control) | `True` | `()` |

Same probe on the previous commit: `ok=True, violation_count=0` for the #368 case. RED → GREEN, with controls in both directions so it is not a check that simply flags everything.

**Your `('transformers','trl')` reconciles exactly.** I measured only `transformers` and chased the discrepancy rather than assuming one of us was wrong: this box's editable metadata is a stale snapshot (`trl>=0.7.0`, no cap), and re-running with the PR's own `trl>=0.14.0,<0.29` reproduces your pair. Your numbers were right — and that is precisely the editable-staleness caveat your own docstring documents.

The `[mlx]` scope narrowing was the right call and it is pinned: widening `_TRAINING_INSTALL_EXTRAS` to include `mlx` fails `test_mlx_only_bound_on_tracked_package_is_not_enforced` on its own.

**Five mutations, all killed:** removing tracked-name enforcement → 3 fail; adding `mlx` → 1; deleting the dedup `seen` guard → **4** (you predicted 2 — it is better than claimed, and it was dead code before this round); neutering `escape()` → 1; removing the packaging-missing warning → 1. All three of my smaller items are genuinely closed rather than argued away.

Live `env check` prints both the "No lock file" line and the bounds panel, exit 3, remediation rendering in full. `docs/serving-and-export.md:254` is measurably true again — it had been promising a case the code could no longer reach.

ruff clean, 130 passed / 7 skipped, startup-light 23/23, merge clean, no assertions removed.

One non-blocking observation, and it is on my suggestion rather than your execution: a core-only install carrying an out-of-range `transformers` for unrelated reasons is now flagged. That is inherent to auditing by `TRACKED_PACKAGES` — which I proposed — and strictly narrower than the first round. Defensible, since such a venv would break `soup train` anyway. Worth at most a sentence in `docs/commands.md`; not a change.

Thank you for the measured iterations on this one. The second round was the harder half: the first fix closed a false positive and, in doing so, made the issue's own case unreachable — and you resolved it by keeping both properties rather than trading one for the other.

ousamabenyounes:
just upvoted! https://www.producthunt.com/products/soup-cli so happy to contribute on this project :)

PR #426, opened 2026-08-16
Support Qwen3.5 MoE layer streaming

Shutaru:
## Summary
- add Qwen3.5 MoE text aliases for layer streaming, including wrapped `text_config` model types
- support heterogeneous decoder-layer parameter sets at the key-presence level while preserving shared tensor layout checks
- compute streaming specs, host-store bytes, VRAM pool budget, and active buffer keys from every layer's shard headers instead of assuming layer 0 uniformity
- keep NF4 streaming safe with per-weight `NF4WeightSpec` validation, required-sidecar checks, packed/absmax bounds, and shared code-table equality
- mirror MoE LoRA target discovery in the streamed setup path and count expert module instances in adapter-memory estimates
- keep multimodal `gemma3` rejected by limiting wrapped-config alias fallback to Qwen3.5 aliases
- refuse canonical stream-key collisions and add regression coverage for the replacement safety guards
- refuse `stream_layers` with `moe_expert_quant`, which is only implemented by the resident setup path

## Why
Qwen3.5 MoE checkpoints can expose the text decoder through wrapper prefixes and non-uniform decoder-layer parameter sets. The previous streaming path failed during sharding or streamed setup before training could begin.

The supported heterogeneity is deliberately narrow: layers may differ only in which short keys they contain. Any key shared by multiple layers must keep the same stored shape, dtype, and—under NF4—quantisation metadata.

## Safety model
- Decoder indices must remain contiguous; missing layers are refused.
- During sharding, `shared_specs` rejects a shared short key whose stored shape or dtype changes between layers.
- For NF4, `quant_specs` is keyed by short weight name and rejects differing `NF4WeightSpec` values. `_CodeTables.observe()` shares the small code tables only after equality checks.
- Runtime reads each layer's safetensors header, derives `needed_specs_by_layer`, and sizes the buffer pool from `RamSource.merge_layer_specs()`, the same union used by the VRAM pre-flight budget.
- `active_keys_by_layer` limits each transfer to keys owned by that layer; required NF4 sidecars and packed/absmax bounds are checked against that layer's own shard spec.
- NF4 rebuilds a fresh `Params4bit` view per weight. It does not reuse a complete layer-0 quant state.
- The existing buffer-ownership guard still verifies that the requested slot owns the requested layer before compute uses it.
- Canonicalised checkpoint-key collisions raise instead of silently discarding one tensor according to safetensors key order.

## Review follow-up
Commit `ef1fcfa` closes the final architecture-admission review. A heterogeneous two-layer MoE now runs resident and streamed on CPU with bit-exact logits; `stream_layers` refuses `moe_expert_quant` because that feature is wired only into the resident construction path; and the CHANGELOG attribution no longer uses the PR number as an issue number.

The CPU control is deliberately load-bearing. Restoring the old layer-0-spec invariant makes it fail on the missing `linear_attn.in_proj_qkv.weight`, and removing the new config guard makes its test fail because no older validator catches the combination.

The response-only `BatchEncoding`/assistant-mask defect discovered during live validation is intentionally excluded from this PR and tracked separately in #430. This PR no longer changes `loss_mask.py` or `tests/test_assistant_mask.py`.

The earlier source-level Apple Silicon allocation workaround (`_build_source(..., device=...)` plus CUDA-only pinning) was intentionally reverted. The final change is limited to passing `pin=False` in the two affected tests, which keeps their CPU-allocation assertions meaningful without claiming to fix main's Apple Silicon behavior. That underlying issue remains tracked in #434 and is not fixed or closed by this PR.

#416 is now in `main`, and this branch has been rebased onto it. The resolution preserves both this PR's `active_keys_by_layer` handling and #416's `require_pin` wiring.

There is no backing issue for this PR, so the CHANGELOG uses `(by @Shutaru in #426)` rather than inventing an issue number.

## Validation
Local validation on Windows / Python 3.12 at `ef1fcfa`, with imports pinned to this checkout:

- `PYTHONPATH=src python -m pytest tests/test_qwen35_streaming.py tests/test_moe.py tests/test_v07200.py tests/test_v07201.py tests/test_v07202.py tests/test_v07203.py tests/test_v07204.py tests/test_stream_multi_gpu_guard.py tests/test_issue385_stream_dtype.py -q -o addopts=` → **567 passed, 31 skipped**
- `python -m ruff check src/soup_cli tests` → **all checks passed**
- mutation control: forcing every layer to reuse layer 0's shard spec → **killed** by the CPU parity test
- mutation control: removing the `moe_expert_quant` streaming refusal → **killed** by the config regression test

Live validation on `Qwen/Qwen3.5-35B-A3B`: streamed sharding completed across all 40 layers, the RAM-tier plan was built from per-layer headers, and ScatterMoE LoRA targets resolved.

## Evidence boundary
The Qwen3.5 MoE aliases now have a CPU resident-vs-streamed control over a real two-layer heterogeneous MoE computation, with final logits equal under `torch.equal`. That establishes the toy float32 decoder path and makes the old uniform-layer assumption fail loudly.

The live 35B run still does not establish resident bit-exactness for the real checkpoint, heterogeneous-layer NF4 equivalence, peak-VRAM behaviour, or CUDA/multi-GPU coverage.


MakazhanAlpamys:
Thanks — and welcome. This is a substantial first PR: it touches the layer-streaming core, which is the most safety-sensitive subsystem we have, so it is getting a correspondingly careful review rather than a quick one. I have unblocked your CI runs (first-time contributors need manual approval here).

Three things while that runs, so you are not waiting on me to find out.

## 1. The claim I most need to understand

> compute streaming specs and byte planning from per-layer shard headers instead of assuming layer 0 uniformity

The layer-0 uniformity assumption is a deliberate guard rather than an oversight, and it has a second consumer that is easy to miss: on the NF4 path, **layer 0's `quant_state` is reused for every layer**, and the buffer pool is pre-allocated from layer 0's spec. Getting that wrong does not crash — it produces mis-dequantised weights and wrong gradients behind a perfectly healthy-looking loss curve. This project has shipped that exact failure shape three times (`.inner.` adapter keys, the full-fine-tune MLX adapter, `target_modules: auto`), and each time the loss curve structurally could not see it.

So the question I am having checked is not "does it work for Qwen3.5" but: with heterogeneous layers now permitted, is a single shared `quant_state` still safe, is the pool still correctly sized, and what replaced the uniformity/contiguous-index guards for a genuinely malformed shard set? If you already reasoned through this, please write it down in the PR — it is the part a reviewer cannot reconstruct from the diff.

## 2. Two files the description does not account for

The summary is entirely about MoE streaming, but the diff also changes `src/soup_cli/data/loss_mask.py` and `tests/test_assistant_mask.py` — `_as_token_id_list`, `_as_mask_list`, `_mapping_value`. I assume these relate to the "later dataset/collator issue" you mention, but as written the PR does not say so.

Please state what they are for. One specific thing to confirm: those helpers return `None` on unexpected input. If a value that previously raised now returns `None` and is quietly skipped, that converts a loud failure into a silent one, which we would want to avoid or at least make deliberate.

## 3. A collision risk that is our problem, not yours

**#416** is open and touches three of the same files — `stream_setup.py`, `layer_stream.py`, `layer_stream_runtime.py`.

Worth flagging plainly: earlier today `main` went red because two PRs that were each green in isolation collided after merge — each had been tested against a `main` that did not yet carry the other. I would rather sequence you and #416 deliberately than discover the interaction afterwards. Nothing for you to do yet; I am having the overlap analysed function by function and will tell you which of you rebases onto the other.

## Also needed before merge

No `CHANGELOG.md` entry and no `docs/` update. Supporting a new architecture family in layer streaming is user-facing, so it needs both — `docs/performance-and-quantization.md` is where the streaming architecture list lives.

One question I will not ask you to reproduce: your live validation on `Qwen/Qwen3.5-35B-A3B` is genuinely useful and we have no such hardware, so we cannot confirm it. I will be explicit in the review about which parts we verified and which rest on your run — that is not distrust, it is just what our evidence supports.

Shutaru:
Thanks for the careful review. I agree this is the load-bearing part of the PR, so here is the reasoning I should have put in the description.

### 1. Heterogeneous layers, NF4 state and the buffer pool

The PR is not trying to make per-key storage heterogeneous. It only allows heterogeneity at the *presence/absence of keys per layer* level, which is what the Qwen3.5 MoE checkpoint needs. For any short key that appears in more than one decoder layer, the stored shape/dtype still has to be identical.

Concretely:

- The sharder still keeps the contiguous decoder-index guard: it derives `n_layers = max(layer_ids) + 1` and raises if any layer index in that range is missing.
- During sharding, `shared_specs` still rejects a same short key with different stored shape/dtype across layers.
- For NF4, `quant_specs` is keyed by the short weight name. If the same short key produces a different `NF4WeightSpec` in another layer, sharding raises instead of writing the cache.
- `_CodeTables.observe()` only lets the small NF4 code tables be shared after checking equality; divergent code or nested-code tables raise.
- Runtime no longer sizes from layer 0. It reads safetensors headers for every layer with `RamSource.layer_specs_from_shards()`, builds `needed_specs_by_layer`, and then calls `RamSource.merge_layer_specs()` to produce the union pool spec. That merge has the same invariant: a shared key must keep one storage layout.
- `active_keys_by_layer` tells the pool which subset to copy for a given layer, so a layer that does not own an expert key does not try to read it, while the pool remains sized for the union of valid keys.
- NF4 reconstruction uses `quant_specs.get(ckpt)` per weight and rebuilds a fresh `Params4bit` view each call via `rebuild_params4bit(ckpt, buffers, spec, self.codes)`. The full `quant_state` is not reused from layer 0; only the validated shared code tables are resident.
- Before runtime rebuilds an NF4 weight, `validate_quant_shape()` checks the logical spec against the packed bytes and absmax sidecars from that layer's shard header, and `install_streaming()` raises if any required sidecar is missing.
- The scheduler ownership guard is unchanged: `LayerBufferPool.wait()` still verifies that the slot requested by layer N actually owns layer N before compute can use it.

So the replacement for the old layer-0 uniformity guard is: per-layer header read + union pool sizing + per-key storage-layout equality + per-weight NF4 metadata equality + per-layer sidecar validation. Malformed shard sets should fail either at sharding, cache read/runtime install, or buffer ownership, not train silently.

### 2. Why `loss_mask.py` is in this PR

That came from the live Qwen3.5 validation, not from the streaming architecture itself.

On `Qwen/Qwen3.5-35B-A3B`, HF returns a `BatchEncoding`-like mapping from `apply_chat_template(..., return_dict=True)`. The old code checked `isinstance(out, dict)`, rejected that mapping, then the fallback path did `list(out)`. For a mapping, that gives the keys (`["input_ids", "attention_mask"]`) rather than token ids, which later surfaced as TRL's collator error: `ValueError: too many dimensions 'str'`.

The new helpers do two things:

- unwrap mapping-like outputs by reading `.get("input_ids")` / `.get("assistant_masks")` instead of iterating keys;
- validate that token ids are actual ints and masks are bool/int-like before accepting them.

The `None` returns are deliberate only in the preferred assistant-mask path: if the preferred mask output is missing, malformed, mismatched, or all-zero despite assistant messages, Soup falls back to the existing incremental rendering strategy rather than trusting a bad mask. The fallback tokenization path is still loud: if it cannot unwrap real integer token ids, `_tokenize_only()` raises `ValueError("tokenizer.apply_chat_template did not return token ids")`.

The all-zero-mask case matters because this tokenizer emits the HF warning about missing `{% generation %}` markers; accepting that path would produce all `-100` labels and a healthy-looking no-op SFT run. The regression tests cover both the mapping-like output and the all-zero-mask fallback.

### 3. Docs/changelog

Added in the latest commit:

- `CHANGELOG.md`: Qwen3.5 MoE layer-streaming support and live validation note.
- `docs/performance-and-quantization.md`: supported-architecture wording, Qwen3.5 MoE scope, and the heterogeneity invariants above.

### 4. #416 overlap

Understood. I will not rework that overlap blindly. Happy to rebase this branch onto #416 or vice versa, whichever sequence you decide is safer after the function-level overlap review.

MakazhanAlpamys (changes requested):
Reviewed in depth — this is a serious piece of work and the review is correspondingly long. Everything below was run in an isolated checkout with `PYTHONPATH` pinned to your branch, so these are measurements of your code, not of an installed copy. I also re-fetched after your `5bd13e3` docs commit and confirmed every code finding still stands.

**First, the good news, because it is genuinely good:** the layer-0 `quant_state` hazard I was most worried about is *closed*, not left open. You replaced the uniformity assumption with a real structural guard — `quant_specs` keyed by short name raising on a differing `NF4WeightSpec`, plus `validate_quant_shape` against each layer's *own* `shard_spec`. I verified the ragged NF4 fixture still refuses. That was the right instinct and it is the hardest part of this change.

And your docs commit is unusually candid — explicitly stating there is no resident bit-exact control because you lacked hardware to load it resident. That kind of honesty about the evidence is exactly what this project wants.

## Blocking

**1. Four existing tests fail on this branch.**

```
FAILED tests/test_v07200.py::TestShardUniformity::test_non_uniform_layer_parameter_set_refused
FAILED tests/test_v07202.py::TestShardQuantGuards::test_a_layer_with_different_shapes_is_refused
FAILED tests/test_v07203.py::TestEstimateAdapterParams::test_counts_two_matrices_per_target_per_layer
FAILED tests/test_v07203.py::TestEstimateAdapterParams::test_auto_target_modules_assume_four
```

Your own files are green (65 passed) — the reason this was invisible is that both commands in your Validation section are `-k`-scoped to the new tests. Please run `pytest tests/test_v07200.py tests/test_v07202.py tests/test_v07203.py tests/test_v07204.py`.

Three separate causes. `test_non_uniform_layer_parameter_set_refused` is now `DID NOT RAISE` — it pinned the guard you deliberately remove, so please rewrite it to assert the *new* contract (differing key sets allowed, a shared key with differing shape/dtype still refused) rather than leaving it red. `test_a_layer_with_different_shapes_is_refused` fails only on your changed message — update the `match=`. Both `TestEstimateAdapterParams` fail with `AttributeError: 'NoneType' object has no attribute '_stream_shape_config'` because they call the method unbound; making `_stream_shape_config` and `_stream_intermediate_size` `@staticmethod` fixes it, since neither uses instance state.

**2. The `text_config` preference silently re-enables multimodal `gemma3`.**

`stream_arch_of` now prefers `config.text_config.model_type` for *every* model. A real `google/gemma-3-*` reports `model_type='gemma3'` with `text_config.model_type='gemma3_text'`, which is in `SUPPORTED_STREAM_ARCHS`. Measured, your branch against main:

```
PR   : gemma3 multimodal -> gemma3_text      (ACCEPTED)
main : MAIN gemma3 REFUSED
```

The justification is 160 lines above your change in the same file: *"streaming a multimodal wrapper as though it were a causal LM is exactly the failure the allowlist exists to prevent."* Your own new docs still say gemma3 is excluded on purpose — so the docs now describe behaviour the code no longer has. Please scope the fallback to your aliases (consult `text_config` only when its `model_type` is a key in `_STREAM_ARCH_ALIASES`) and add a test that multimodal `gemma3` still refuses.

**3. The VRAM pre-flight now under-predicts on heterogeneous layers.**

`stream_setup.py:319` sets `layer_bytes = max(layer_byte_sizes)`, and that feeds `layer_bytes * buffers`. But the pool allocates `merge_layer_specs` — the **union**. On your own fixture:

```
per-layer byte sizes                  : [128, 128]
layer_bytes used for the VRAM budget  : 128
ACTUAL pool bytes per buffer          : 192
under-prediction factor               : 1.5
```

The predictor's documented contract is that it *never* under-predicts, and under-prediction is the asymmetric direction: a clean OOM on Linux, but on Windows/WDDM nothing raises at all — it silently spills, and "peak is bounded by one layer" quietly stops being true. One line: derive the budget's `layer_bytes` from the same union the pool uses. (`store_bytes = sum(...)` for the host store is correct as-is.) Please add a test asserting budgeted bytes are at least the actual pool bytes per buffer for a heterogeneous set.

## Please also address

**4. `_canonical_stream_key` can silently drop a tensor.** If a checkpoint carries both `model.layers.N.X` and `model.language_model.layers.N.X` they collapse to one key, and `where[key] = ...` is last-writer-wins on safetensors ordering. Reproduced — one real tensor discarded, no error. That is the wrapped/multimodal shape the canonicalisation exists for, so it is reachable. Please raise on collision.

**5. The MoE LoRA wiring you advertise is untested.** Stubbing out `target_modules = moe_targets` breaks nothing. A control mutation in the same method *is* caught, so this is a genuine gap, not an untestable path.

**6. The guard that replaces the removed uniformity check is itself untested.** Deleting either the `shared_specs` raise or the `merge_layer_specs` raise leaves the suite green. Given that these are now the *only* thing standing between a ragged checkpoint and a mis-streamed model, they need direct tests — this is the single most important coverage request in the review.

**7. Under `moe_lora`, `_estimate_adapter_params` under-counts by `(4 + 3*num_experts)/7`** — measured **7.4x** at 16 experts; by arithmetic roughly 55x at a real A3B's 128, though I could not measure that. It counts pattern names, not module instances, and `gate/up/down_proj` match once per expert. The code predates you, but your PR is what makes the path reachable, and its own docstring says under-counting eats the safety margin. Either count instances or document it.

**8. `ruff check` fails**: `I001 unsorted imports`, `tests/test_qwen35_streaming.py:126`. CI lint will go red.

## The `loss_mask.py` change — a real find, and it needs its own PR

You found a genuine bug, and a nasty one. I reproduced it independently: `BatchEncoding` is **not** a `dict` (it subclasses `UserDict`), so `isinstance(out, dict)` misses and `list(out)` returns `['input_ids']` — the key strings — where token ids are expected. Silent garbage in assistant-only masking, on any tokenizer whose `apply_chat_template` returns that type. Nothing to do with Qwen3.5.

I have filed it as **#430** with credit to you, and I would like the fix to land there rather than here. Two reasons: it changes label construction for **every** user of assistant-only masking, and it deserves review on its own terms instead of as a passenger in a streaming PR. Your fix is the right shape — note it also makes `_tokenize_only` raise where main silently returned garbage, which is the correct direction.

Two behaviour changes to call out explicitly when you move it: an all-zero assistant mask now falls back to the key-token heuristic, and a tokenizer returning a tensor of ids now hard-raises where it previously produced `[tensor(1), tensor(2), ...]`.

## Sequencing with #416 — my job, not yours

I measured the merge rather than guessing. `git merge-tree` gives conflicts in `layer_stream.py`, `layer_stream_runtime.py` and `CHANGELOG.md`. The risky one is `install_streaming`: you rewrite that statement to `_build_source(..., needed_specs_by_layer, ...)` plus `LayerBufferPool(..., active_keys_by_layer=...)`, while #416 rewrites the same statement to add `require_pin=`. A careless resolution silently drops either your heterogeneous-layer support or their pinning control. `build_stream_plan` also collides, but only because you each append a keyword to the same signature line — trivial.

Also worth knowing: whichever of you lands second will appear to have broken `tests/test_v07200.py`, because your branch already breaks a test in the file #416 adds to. I will tell you both explicitly who rebases onto whom; please do not race it.

## What we verified, and what rests on your run

The code is *consistent* with your `Qwen/Qwen3.5-35B-A3B` report — `n_layers` from `max(layer_ids)+1` shards 40 layers, `get_moe_target_modules` emits bare suffixes that PEFT matches through the `.inner.` prefix, `detect_moe_model` reaches the wrapped `text_config`, and "a later dataset/collator issue" matches the `BatchEncoding` bug I reproduced. Nothing contradicts it.

**Unverified by us:** any training step or loss curve, bit-exactness of streamed Qwen3.5 MoE against a resident control, pool and VRAM behaviour at 35B, NF4 combined with heterogeneous layers, and any CUDA or multi-GPU path.

Which brings me to a question rather than a request. The `qwen3_5_moe -> qwen3` alias asserts that a sparse-MoE decoder block streams equivalently to a dense `qwen3` block — and a bit-exact resident control is precisely the evidence `SUPPORTED_STREAM_ARCHS` exists to demand before an arch is added. Your docs concede no such control exists. Would you consider marking the alias as unvalidated in `SUPPORTED_STREAM_ARCHS` itself rather than only in prose, so a reader of the allowlist sees the same caveat a reader of the docs does?

Thanks again — three of the findings above are in code you inherited rather than wrote, and the `BatchEncoding` bug is a real contribution on its own.

MakazhanAlpamys:
Sequencing decision, as promised — so you are not guessing.

**#416 goes first; please rebase onto it once it lands.** It is smaller, was opened earlier, and its remaining work is a wiring test plus one design decision, whereas this PR currently has four failing tests. That ordering is about readiness, not merit.

Concretely, when you rebase, the resolution to be careful with is `install_streaming`. Both PRs rewrite the same statement: #416 adds `require_pin=` to `_build_source`, you change `spec` to a per-layer sequence and add `active_keys_by_layer=` to `LayerBufferPool`. Both must survive — dropping either loses a real capability silently. `build_stream_plan` also conflicts, but only because you each append a keyword argument to the same signature line, so keep both.

No action needed until #416 merges; I will ping you here. In the meantime the blockers in my review are all independent of the rebase, so they are safe to work on now.

MakazhanAlpamys:
That is the description this PR should have shipped with, and I want to be precise about its status: **your section 1 is confirmed, not merely accepted.** I had the guard chain traced independently before you posted it, and the finding was the same — `quant_specs` keyed by short name raising on a differing `NF4WeightSpec`, `merge_layer_specs` holding the union invariant, `validate_quant_shape` running against each layer's *own* `shard_spec`, and `_CodeTables.observe()` gating the shared tables. I also ran the ragged NF4 fixture and watched it refuse:

```
ValueError: decoder weight 'mlp.down_proj.weight' has inconsistent NF4 metadata
across layers ((64, 128) vs (64, 256))
```

So the distinction you draw — heterogeneity in *which keys a layer owns*, never in *how a shared key is stored* — holds in the code. That was the thing I was most worried about, and you got it right. Please put that section in the PR body; it is the part a reviewer cannot reconstruct from the diff, and the next person to touch this file will need it.

One caveat that does not weaken the argument but does need action: **that guard chain is currently unprotected by the suite.** I deleted the `shared_specs` raise, and separately the `merge_layer_specs` raise, and the streaming tests stayed green both times — with a control mutation in the same file failing as expected, so this is a real gap and not a testing artefact. Your reasoning is now the only thing holding that invariant, and reasoning does not survive the next refactor. That is change request 6 in my review and, given what you have just written, it is the most important one.

## Your section 2 changes my mind on one point, not the other

The all-zero-mask argument is a good one and I had underweighted it: a tokenizer that emits the missing-`{% generation %}` warning would otherwise produce an all-`-100` row, i.e. a no-op SFT run behind a healthy loss curve. Guarding that is right, and the fallback-rather-than-trust choice is the correct direction.

I still want it as **#430** rather than here, and your explanation actually strengthens the reason. This is not a Qwen3.5 concern at all — I reproduced `BatchEncoding` returning `['input_ids']` on the installed stack with no MoE model in sight, so it silently corrupts labels for anyone whose tokenizer returns that type, on any architecture. A change with that blast radius should not be reviewed as a passenger in a streaming PR, and it should not have to wait for this PR's blockers to clear before users get it.

Sequencing that respects your dependency: **#430 lands first, then this rebases onto it.** You do not lose the fix, and you do not carry it.

## What your reply does not cover

Three findings in my review are independent of the uniformity question and still stand:

1. **Four existing tests fail on this branch** — including `test_non_uniform_layer_parameter_set_refused`, which pinned the guard you deliberately replaced and was left red rather than rewritten to assert the new contract. Your two validation commands are `-k`-scoped to the new tests, which is why they came back clean.
2. **`stream_arch_of` now prefers `text_config.model_type` for every model**, so multimodal `gemma3` is accepted where `main` refuses it — measured both ways. Your own new docs still say it is excluded on purpose.
3. **The VRAM budget under-predicts 1.5x on heterogeneous layers** — `max(layer_byte_sizes)` where the pool allocates the union. Given how carefully you reasoned about the union for pool *sizing*, I suspect this is simply the one place the union did not get threaded through.

Note that 2 and 3 are the same shape as the defect class you argued so carefully about in section 1: not a crash, but a quiet acceptance of something that should be refused, and a number that is wrong in the direction that does not announce itself.

## Sequencing

Decided, and posted on both PRs: **#416 first, this second.** #416's remaining work is a wiring test and one design decision; yours currently has four failing tests. That is readiness, not merit. When you rebase, the resolution to watch is `install_streaming` — both PRs rewrite the same statement, and a careless merge drops either your `active_keys_by_layer` or their `require_pin`. Both must survive.

Thanks for engaging with the hard part directly instead of defending the diff.

MakazhanAlpamys (changes requested):
Re-reviewed at `79c3969` in an isolated worktree with `PYTHONPATH` pinned to your branch, and mutation-tested each fix rather than reading it. **Seven of nine requests, plus blockers 1, 2 and 4–8, are fixed and confirmed by measurement.** Two things left, and one of them is not your fault.

## Confirmed fixed

The four previously-failing tests are green — `test_v07200 + v07202 + v07203 + v07204` gives **449 passed, 2 skipped**. You **rewrote** `test_non_uniform_layer_parameter_set_refused` rather than deleting it, and the replacement asserts something real: the extra key lands in shard 1 and is *absent* from shard 0, with a sibling pinning the shared-key refusal. That is exactly the ask — a test that follows the contract you changed, not one that was removed for being inconvenient.

`stream_arch_of` measured both ways: multimodal `gemma3` is **REFUSED on your branch and on `main`**, and a `qwen2_vl` wrapper whose `text_config` is a supported arch is refused too. Only the `qwen3_5_moe*` aliases change behaviour, and reverting the scoping kills `test_multimodal_gemma3_wrapper_is_still_rejected`. The docs no longer contradict the code.

The guard chain from your section 1 is now genuinely protected — every one of these mutations dies:

| Mutation | Killed by |
|---|---|
| `shared_specs` raise → `if False` | 2 tests |
| `merge_layer_specs` raise → `if False` | 1 test |
| canonical-key collision raise → `if False` | 1 test |
| `target_modules = moe_targets` → `pass` | the wiring test |

That wiring test asserts the **exact 7-element list** reaching `LoraConfig`, not merely that something changed — which is the difference between a test and a formality.

`_estimate_adapter_params` now counts instances: **13.00x** the old estimate at 16 experts, **97.00x** at 128. Ruff clean. And thank you for moving `loss_mask.py` out to #430 — that was the right call for a change with that blast radius, and it lets the fix reach users without waiting on this PR.

## 1. macOS CI is red — and you did not break it

All three macOS cells fail; ubuntu ×3, windows ×3, lint, type-check and mlx-smoke are green. One test:

```
tests/test_qwen35_streaming.py::TestQwen35StreamingRuntime::test_install_streaming_accepts_heterogeneous_layer_sets
E   AssertionError: assert 'mps' == 'cpu'
```

(That message is from run `31961830885`; the current run hides it behind an `all(...)` generator — worth unpacking that assertion regardless, since a failure you cannot read costs a round trip.)

The test calls `install_streaming(..., device="cpu")` without `pin=`, so `pin=True`, and on a macOS runner with MPS available `torch.empty(..., device="cpu", pin_memory=True)` comes back on `mps`.

**`main` makes the identical pinned allocation**, and your `device="cpu"` commit demonstrably did not change the result — run `31961830885` already contained it and still returned `mps`. Your test is simply the first thing in this repo to assert on it.

The fix is the one you already used next door: `test_ram_source_explicitly_requests_cpu_allocations` passes `pin=False` and is green on macOS. Either pass `pin=False` here too, or drop the four `device.type == "cpu"` assertions — they are incidental to what the test is actually about, which is heterogeneous layer sets.

I will file the underlying finding separately: **on Apple Silicon the pinned RAM store is not host memory**, so layer streaming's "the frozen base lives in CPU RAM" claim does not hold there. That predates you by a long way.

## 2. Blocker 3 is fixed in the code, but nothing protects it

The line is right — I measured `max(per-layer) = 128` against `budget = 256 == actual pool bytes/buffer = 256`. But reverting `stream_setup.py:348` to `max(layer_byte_sizes)` **survives the entire streaming suite**.

The control is what makes this a real gap rather than an untestable path: the `target_modules` mutation is in the **same method** — line 444 against the budget's line 348, both inside `_setup_streaming_transformers` — and it *is* caught. So the method is exercised end to end; that one line is not.

The reason is worth naming, because it is the same shape as request 5 last round: `test_stream_layer_budget_bytes_uses_the_union_pool_spec` calls the helper **directly**, and the only test that drives the setup path monkeypatches `layer_specs_from_shards` to `[{...}] * 2` — **homogeneous**, where `max()` and the union agree by construction. Helper tested, wiring not. Making that stub heterogeneous and asserting on the argument reaching `build_stream_plan` / `decide_stream_fit` closes it.

## On my question about the alias

`_UNVALIDATED_STREAM_ARCH_ALIASES` sitting under `SUPPORTED_STREAM_ARCHS` with the "no resident bit-exact control" comment is a better answer than I expected — it puts the caveat where someone editing the allowlist will meet it.

One follow-up, not blocking: nothing **reads** it. No test, no runtime notice. So an operator streaming Qwen3.5 sees the pre-flight panel report `qwen3` with the caveat living only in prose. A one-line note in the panel when the resolved arch came through that dict would put the reader of the terminal and the reader of the source on equal footing.

Fix those two and this is ready from my side. The remaining work after that is the #430 / #416 rebase sequencing.

MakazhanAlpamys:
Filed the underlying Apple Silicon finding as **#434**, as promised — so it is tracked independently of your PR and does not gate you.

Short version for the record: `RamSource` allocates with `torch.empty(..., pin_memory=self.pinned)` and no `device=`, and on macOS that lands on `mps`. `main` does the identical thing; your explicit `device="cpu"` demonstrably did not change it. What your branch contributed is the **first test in this repo that asserts on it** — which is the only reason anybody knows.

I was careful in the issue to separate what is measured (the allocation lands on `mps`) from what is not (whether that makes streaming wrong, slower, or merely misnamed on that platform — nobody here has the hardware). It may well end in "streaming refuses on Apple Silicon" rather than a code fix.

Nothing changes for you: pass `pin=False` in that one test, or drop the four incidental `device.type == "cpu"` assertions.

MakazhanAlpamys:

MakazhanAlpamys (changes requested):
**Blocker 3 is properly closed**, and I verified it the way it failed last time: reverting `stream_setup.py:348` to `max(layer_byte_sizes)` now **fails** (1 failed / 429 passed), and gutting `_stream_layer_budget_bytes` to a per-layer max fails **two** tests. Control in the same method (`target_modules = None` on the MoE branch) also fails, so the harness is real rather than accidentally green.

The reason it works now is the **fixture**, and it is worth naming: `layer_specs` is heterogeneous — `self_attn.q_proj.weight` on layer 0, `linear_attn.in_proj_qkv.weight` on layer 1 — so `max()` = 64 and the union = 128, and the assertion is `layer_bytes == 128`. Those two answers are *distinguishable*, which is exactly what the old `[{...}] * 2` fixture made impossible by construction. That was the actual defect last round, and you fixed the fixture rather than the assertion.

`pin=False` at both allocation sites (`:237` and `:257`) is the right fix, and it keeps the six `device.type == "cpu"` assertions meaningful instead of dropping them.

## But I cannot confirm macOS is green — CI has not run on this head

`gh pr checks 426` reports **no checks**, and the check-runs API is **empty** for the head commit. So greenness is unverifiable from here.

Please push again (an empty commit is fine) to trigger a run. I would rather see all three macOS cells than infer it — and inferring it is precisely what I would be doing, since I have no Apple Silicon.

## One thing to state rather than leave as a silent revert

An earlier commit fixed this in the **source** — `_build_source(..., device=...)` plus `pin = bool(pin and str(device).startswith("cuda"))` — and the head commit **reverted that** in favour of the test-side fix.

That is the right scoping call: #434 is `main`'s bug, not this PR's. But please say so in the PR description, so #434 does not look accidentally closed and the next reader does not have to reconstruct why the source fix disappeared.

## Non-blocking, unchanged

Nothing reads `_UNVALIDATED_STREAM_ARCH_ALIASES` beyond the `dict()` copy on the next line — no test, no runtime notice. The docs caveat you added is genuinely good — *"no resident bit-exact control because the author did not have hardware that could load it resident"* is exactly the right register — so this is only about the user who never opens the docs and sees the panel print `qwen3` with no caveat.

## Two rebase notes

**`docs/performance-and-quantization.md` conflicts with `main`** — the "Honest scope" arch list, which #365 and #416 also edited. Source files auto-merge; you are 14 commits behind.

**And a latent crash if you land near #416.** Both PRs merge cleanly into `main` separately, but you change `_build_source`'s third argument from a spec `Mapping` to a **`Sequence` of per-layer mappings**, while #416 adds a `_spec_bytes(spec)` helper doing `spec.values()`. Merged, that is:

```
list -> AttributeError: 'list' object has no attribute 'values'
```

on their `require_pin` refusal path — and **neither suite catches it**, because their test passes a dict. I measured it. Whoever lands second needs to fix `_spec_bytes` and re-run the streaming suite. I have told #416 the same thing.

## Deleted assertions — checked, legitimate

One test removed: `test_non_uniform_layer_parameter_set_refused`. It is **replaced** by `test_non_uniform_layer_parameter_set_is_allowed` plus `test_shared_key_with_conflicting_shape_is_still_refused` — correct, since heterogeneous layers are the feature — and I confirmed the replacement is load-bearing by making the shared-key guard unreachable, which kills two tests. `test_v07202.py`'s match-string update is consistent with the new error path.

ruff clean, 658 passed / 4 skipped across the streaming and MoE suites.

MakazhanAlpamys:
Status note, and the delay on this one is mine rather than yours.

Where it stands: you pushed on 17 Aug and I have not come back to it since my 16 Aug note. Nothing is wrong with the push — I have not re-run it.

**The sequencing I gave you still holds: #416 goes first.** That is the unblocker here, not anything on your side, so the useful order is (1) #416 lands, (2) you rebase onto it, (3) I review this in full. Your branch is currently **conflicting** with `main` in any case — plenty has landed since 16 Aug — so the rebase is needed regardless of #416.

Nothing needed from you until #416 is in, unless you would rather rebase onto current `main` now and keep the #416 merge as a second, smaller rebase. Your call; I will not read either as a lack of movement.

Related: the Apple Silicon finding from this review is tracked independently as #434 and does not gate you.

Shutaru:
Thanks for the update — no worries about the delay. The original sequencing still makes the most sense to me, so I’ll wait for #416 to land and then do one clean rebase onto the resulting main rather than churn the branch twice.

I’ll resolve the overlap carefully, make sure both active_keys_by_layer and require_pin survive, rerun the full relevant suite, and push the refreshed branch for review. I appreciate you keeping #434 separate too.

MakazhanAlpamys:
#416 has landed (`9d1ce2c`), so **the unblocker you were waiting on is in**. That was the sequencing I gave you, and it is now your turn rather than mine.

Two things to fold into one rebase, since `main` has moved a fair amount since 17 Aug:

1. `training.stream_pin` is now a real config key, and it threads `require_pin` through `stream_setup` -> `build_streamed_model` -> `install_streaming` -> `_build_source`. If your MoE work touches that chain, it will conflict there rather than trivially.
2. Also landed tonight: #409 (`soup draft`), #458 (cut_ce/liger path matching). Neither should touch you, but they moved `CHANGELOG.md` — put your bullet under `[Unreleased]`, and note the frozen `[0.73.3]` section is at line 122 now. A rebase that carries an entry along with the section it was in is exactly how #416 ended up with its bullet inside the released block.

Once you have pushed the rebase I will review this in full — my delay, not yours, and I am not counting the time it sat.

No need to reply; a push is the signal.

MakazhanAlpamys:
Full review, finally — and thank you for the patience. The delay on this was mine.

**Verdict up front: the engineering underneath is solid and I could not break it. I am blocking on one thing, and it is the arch-admission rule rather than your code.**

## What I could not break

The heterogeneous-layer work is real and well guarded. Two mutations, both killed:

| mutation | result |
|---|---|
| `layer_specs_from_shards` -> reuse layer 0's spec for every layer (restore the old invariant) | **KILLED** — 3 tests, incl. `test_install_streaming_accepts_heterogeneous_layer_sets` |
| delete the shared-key layout-conflict guard | **KILLED** — `test_merge_layer_specs_refuses_conflicting_shared_layouts` |

Relaxing "the pool spec is built from layer 0" to "presence/absence may vary, a shared key must keep one storage layout" is the right generalisation, and `merge_layer_specs` raising on a conflicting shared key is the part that keeps it safe.

**The deleted uniformity test is a legitimate inversion, not a removal** — I checked it specifically because this repo has seen PRs that quietly drop coverage. `test_non_uniform_layer_parameter_set_refused` becomes `..._is_allowed` (asserting `mlp.extra.weight` lands in layer 1's shard, not layer 0's) **plus** a new stricter `test_shared_key_with_conflicting_shape_is_still_refused`. Net coverage goes up.

Every compat validator still fires: `train_router_only` + `moe_lora`, `packing`, `multipack`, `use_fsdp2_compile`, `lora.r: 0`, DoRA, VeRA — all refused by name under `stream_layers: true`. The MoE-specific one I went looking for, `train_router_only` + streaming, is properly refused.

## The blocker: the alias is unvalidated *and* unguarded

You were straight about the first half — the dict is literally named `_UNVALIDATED_STREAM_ARCH_ALIASES`, and `docs/performance-and-quantization.md` says there is no resident bit-exact control because you could not load the model resident. I would much rather have that than a silent allowlist edit, and I want that on the record.

But I pointed the alias at a **wrong** decoder family and ran the streaming suites:

```python
"qwen3_5_moe": "phi3",   # MUTATION
```
```
FAILED tests/test_qwen35_streaming.py::TestQwen35StreamingAliases::test_stream_arch_of_accepts_qwen35_moe_aliases
1 failed, 157 passed, 5 skipped
```

The only thing that fails is `assert stream_arch_of(cfg) == "qwen3"`. **That pins the mapping's spelling, not its correctness.** Nothing in the suite would notice if `qwen3` were the wrong streamer for an MoE decoder — which is precisely the failure the allowlist exists to prevent: a half-supported arch streams weights into the wrong module and mis-trains *silently*, with no crash to trace.

And the caveat does not propagate. `stream_arch_of` returns plain `"qwen3"`, so the runtime, the panel and the user all see a validated dense family. The honesty lives at the alias table and stops there.

**Here is the part that changes the ask.** I checked how the nine dense archs establish this, expecting to find a GPU gate you could not match:

```python
tests/test_v07200.py:1488
def test_streamed_logits_match_resident_bit_exactly(self, tmp_path):
```

It runs **on CPU**, on a tiny synthetic model, and executes on every CI cell. Your own file's two `torch.equal` calls compare a pooled buffer against the source store — copy fidelity, which is worth having, but a different property.

So hardware was never the barrier to a parity test; it was only the barrier to validating the real 35B checkpoint. **A toy heterogeneous-MoE config through the same CPU harness would convert a spelling assertion into a real one**, and it is achievable on the laptop you have. That is what I am asking for.

I am not asking you to prove anything about `Qwen/Qwen3.5-35B-A3B`. Your live run there is yours and I have no way to verify it — I have a 4 GB RTX 3050 and cannot load that model streamed or resident, which I would rather say plainly than imply I checked.

## Second, smaller ask

`stream_layers: true` + `moe_expert_quant: nf4` is **accepted and silently ignored**. `apply_moe_expert_quant_if_configured` is called from `_setup_transformers` (the resident branch); `setup()` dispatches streaming to `_setup_streaming_transformers`, and `stream_setup.py` never mentions it.

The schema hole is older than your PR — `main` accepts the same YAML — but on `main` the run is refused at the arch gate, so **this PR is what makes the combination reachable**. Please refuse it in the cross-validator with a message naming the reason. It is a few lines, and it stops a fifth instance of the pattern v0.73.3 was named for from landing on a path you are opening.

## Everything else

Zero unexplained test deletions. `ruff` clean. CI 12/12 on the true head. Docs updated in four places, including the scope caveat. One nit: the CHANGELOG suffix reads `(#426 by @Shutaru)` — house form is `(#<issue> by @Shutaru in #426)`.

## Where this leaves us

A CPU parity test with a toy MoE config, a validator refusal for `moe_expert_quant`, and the CHANGELOG suffix. The 450 lines of heterogeneous-layer engineering — which is the hard part and the part I spent most of the review trying to break — I have no objection to at all.

Shutaru:
Thanks for the careful review. I pushed [`ef1fcfa`](https://github.com/MakazhanAlpamys/Soup/pull/426/commits/ef1fcfa) with the three requested changes.

The Qwen3.5 aliases now have a CPU resident-vs-streamed control built around a two-layer hybrid MoE. Layer 0 uses `self_attn.q_proj`, layer 1 uses `linear_attn.in_proj_qkv`, and both execute routed experts. The final logits satisfy `torch.equal`.

I repeated the old-invariant mutation by forcing every layer to reuse layer 0's shard spec. The new test fails on the missing linear-attention weight, so it is testing the heterogeneous path rather than only the alias spelling.

`stream_layers` with `moe_expert_quant` now fails during config validation, with the message explaining that expert quantization exists only on the resident construction path. Removing that guard makes the regression test fail with `DID NOT RAISE`; no older validator is satisfying it accidentally.

The CHANGELOG entry is under `[Unreleased]`. This change has no backing issue, so I used `(by @Shutaru in #426)` rather than inventing an issue number for the first slot.

Validation on this head:

- relevant streaming/MoE suite: **567 passed, 31 skipped**
- Ruff over `src/soup_cli` and `tests`: **clean**
- both review mutations: **killed**

I also started the repository-wide suite with `-x`. It reached 721 passed before stopping on the local FastAPI 0.115.0 / Starlette 1.6.0 mismatch; `pip check` confirms FastAPI requires Starlette `<0.39.0`, while this environment has 1.6.0. That failure is outside the files changed here. Hosted CI is now running on `ef1fcfa`.

MakazhanAlpamys:
Merged as `82bce26`. **Both asks are closed — and I owe you a correction on the probe I used to justify one of them.**

## The correction first

I blocked partly on this mutation: point the alias at a wrong decoder family, `"qwen3_5_moe": "phi3"`, and observe that only a string assertion fails. I treated that as evidence the mapping's *correctness* was unguarded.

**That probe was near-null and could never have shown what I claimed.** `arch` is a stored label, not a dispatch key — in `layer_stream_runtime.py` it appears exactly twice, both in prose:

```
952:  # Absolute path, never the bare name: on Windows CreateProcess searches the
1356: hard-coded name list that would drift the moment an architecture names its
```

It reaches `shard_checkpoint(arch=...)` and `build_stream_plan(arch=...)`, is carried on the dataclass, and its only behavioural role is admission — and `phi3` is in `SUPPORTED_STREAM_ARCHS` too. So no parity test could ever have caught it, and "only a string assertion fails" was the *expected* result rather than a finding. I am recording that here because I asked you to write code on the strength of it.

## The gate is real, and I proved it the right way

`TestQwen35StreamingParity::test_heterogeneous_moe_logits_match_resident_bit_exactly` compares streamed logits against a resident reference:

```python
got  = streamed(input_ids=input_ids).logits
want = reference(input_ids=input_ids).logits
assert torch.equal(got, want), (got - want).abs().max().item()
```

`grep -n "skipif\|requires_cuda\|pytest.mark"` on that file returns **nothing**, and it runs `device="cpu"`, `dtype="float32"` — so it executes on all nine CI cells, exactly like the gate the nine dense families get.

The probe that actually discriminates is a **silent** mis-stream. I patched `RamSource.get` to serve expert weights from layer 0 — same shapes, no crash, nothing to trace:

```
FAILED tests/test_qwen35_streaming.py::TestQwen35StreamingParity::test_heterogeneous_moe_logits_match_resident_bit_exactly
1 failed, 13 passed
```

That is the failure mode the allowlist rule exists for — "streams weights into the wrong module and mis-trains silently" — and it is now caught on the numbers.

**Ask 2 is closed too:**

```
REFUSED: training.stream_layers is incompatible with training.moe_expert_quant='nf4':
expert quantization is applied only by the resident model-construction path and
would otherwise be silently ignored.
```

## What I want on the record

You named your own alias table `_UNVALIDATED_STREAM_ARCH_ALIASES` and wrote the missing-resident-control caveat into the docs **before anyone asked**. That is why this was a review about closing a gap rather than about finding a concealed one, and it is the difference between an exception we can reason about and a silent allowlist edit.

The second commit is a real fix, not scope creep: `_spec_bytes` assumed a single shared spec, so with per-layer specs the pin-refusal message would have reported a wrong store size. Mutating it to reuse layer 0's spec fails `test_the_refusal_sums_heterogeneous_layer_specs`, which pins all three candidate wrong answers rather than just the right one.

Seven deleted test lines, all accounted for: the stale layer-0 docstring, and `test_non_uniform_layer_parameter_set_refused` **inverted** into `..._is_allowed` plus a stricter `test_shared_key_with_conflicting_shape_is_still_refused`. Net coverage up. All eight streaming exclusions still refuse.

## The residual, stated plainly

The toy gate proves the streaming **runtime** is bit-exact on a heterogeneous MoE graph. It cannot prove the real Qwen3.5-MoE decoder is qwen3-shaped — and nothing behavioural can, for the reason in my correction above. Your live 35B validation is **unverified by me**: this box is a 4.29 GB RTX 3050 and cannot load that model streamed or resident. I would rather say that than imply I checked it. Your docs disclose the same limit, which is why I am comfortable merging.

Third merge. Credited in `CONTRIBUTORS.md` (`bacdffd`). Thank you for the patience across a review I twice started and stopped.

Shutaru:
It’s a pleasure to help. If you think anything else should be added, just let me know and I’ll take care of it.

Issue #425, opened 2026-08-16
stream_layers: true crashes on pre-Ampere (T4) GPUs — fp16 GradScaler receives bf16 LoRA adapters


lesterppo:
## Summary

Running the `proof-4gb.ipynb` §5 headline (or any `stream_layers: true` config) on a free Colab T4 (sm_75, Turing) crashes on the **first** training step:

```
NotImplementedError: "_amp_foreach_non_finite_check_and_unscale_cuda" not implemented for 'BFloat16'
```

## Environment

- Hardware: free-tier Colab, Tesla T4 (sm_75), 15.36 GB VRAM, 12 GB RAM, 2 vCPU
- Python 3.12.13, torch 2.11.0+cu128, CUDA 12.8
- Soup 0.73.2 from `main` (`pip install "soup-cli[train] @ git+https://github.com/MakazhanAlpamys/Soup.git"`), `torchao` uninstalled per the notebook
- `NousResearch/Meta-Llama-3.1-8B-Instruct`, `quantization: 4bit`, `stream_layers: true`, `stream_buffers: 2`, LoRA r=8

## Repro

Follow `notebooks/proof-4gb.ipynb` through §5. Crash at the first `clip_grad_norm_`:

```
  File ".../soup_cli/trainer/sft.py", line 1521, in train
    self.trainer.train(resume_from_checkpoint=resume_from_checkpoint)
  File ".../transformers/trainer.py", line 2715, in _inner_training_loop
    _grad_norm = self.accelerator.clip_grad_norm_(...)
  File ".../accelerate/accelerator.py", line 3006, in clip_grad_norm_
    self.unscale_gradients()
  File ".../accelerate/accelerator.py", line 2944, in unscale_gradients
    self.scaler.unscale_(opt)
  File ".../torch/amp/grad_scaler.py", line 358, in unscale_
    optimizer_state["found_inf_per_device"] = self._unscale_grads_(...)
  File ".../torch/amp/grad_scaler.py", line 294, in _unscale_grads_
    torch._amp_foreach_non_finite_check_and_unscale_(...)
NotImplementedError: "_amp_foreach_non_finite_check_and_unscale_cuda" not implemented for 'BFloat16'
```

## Root cause (diagnosed)

`pick_mixed_precision` correctly returns `fp16` on a T4 (`cc 7.5 < BF16_MIN_CC`), so
`TrainingArguments(fp16=True, bf16=False)` and the GradScaler are fp16. But peft's
[...]

Issue #385, opened 2026-08-13
Layer streaming hardcodes bfloat16 on every CUDA card, so the entire free tier (T4, P100, V100, GTX 16xx) is unsupported without saying so


MakazhanAlpamys:
`trainer/stream_setup.py:181` is:

```python
dtype = "bfloat16" if on_cuda else "float32"
```

Unconditional. There is no bf16-capability check, although `trainer/asr.py:330` and `utils/gpu.py:259` both call `torch.cuda.is_bf16_supported()` before choosing.

Every published layer-streaming measurement was taken on an RTX 3050 Laptop, which is **Ampere (sm_86) and has bf16**. The hardware a reader is most likely to try this on does not:

| Where | GPU | Arch | `is_bf16_supported()` |
|---|---|---|---|
| Colab, free tier | T4 | Turing sm_75 | **False** |
| Kaggle | T4 x2 / P100 | Turing / Pascal | **False** |
| Older consumer | GTX 16xx, RTX 20xx | Turing | **False** |
| Cheap cloud | V100 | Volta sm_70 | **False** |

So on the entire free tier, Soup shards and streams the base in a dtype the card does not support for compute. Nobody has run it there — this is the same shape as the four backends the H100 session found had never executed once, and for the same reason: it cannot fail on the maintainer's card.

## Why it is blocking something concrete

The launch plan's Colab/Kaggle proof notebook — a reader caps a T4 to ~4 GB and watches a streamed 8B run agree with a resident reference — runs on exactly this hardware. It is blocked on this, separately from #347.

## The path itself is already dtype-agnostic — measured

Both the planner and the sharder already accept fp16 (`SUPPORTED_STREAM_DTYPES` / `_SUPPORTED_DTYPES` = `bfloat16, float16, float32`), and the runtime is exact in it. Four arms, one process, each streamed model compared against a resident reference **of matching numerics** (for NF4 that means a genuinely NF4-quantised reference, built the way `tests/test_v07202.py::_resident_nf4` builds it — comparing against a bf16 resident model would measure the dtype, not the streaming):

```
torch 2.5.1+cu121 | RTX 3050 Laptop | sm_86

dtype      quant  max|streamed - resident|
bfloat16   none   0.000000e+00
bfloat16   nf4    0.000000e+00
float16    none   0.000000e+00
float16    nf4    0.000000e+00
```

[...]

Issue #367, opened 2026-08-10
live_eval.load_model_and_tokenizer takes no quantisation argument: an NF4 run is judged as an NF4 adapter on a bf16 base


MakazhanAlpamys:
## What

`utils.live_eval.load_model_and_tokenizer` has no quantisation parameter, so every live evaluation path built on it — `soup ship` leg 1/2, `soup diagnose --base-model`, `soup advise --probe-model`, `soup eval behavior`, `tunability --live` — loads the base in **bf16** regardless of how the adapter was trained.

Observed while measuring NF4 convergence (`benchmarks/gate-h100-validation.md` L3257–3260): the absolute accuracies in that arm "are not NF4-inference numbers — the judge loads a bf16 base".

## Impact

An adapter trained against an NF4-quantised base is scored against a base it never saw. The mismatch is not catastrophic — the paired NF4 convergence result (mean +0.0053 against within-arm spreads of 0.0333 / 0.0200) survived it precisely because *both* arms were judged the same way — but it means:

- absolute leg-1 / leg-2 numbers for a 4-bit run are not the numbers that run will produce in deployment, and
- a user comparing a `soup ship` score against their own `soup serve --quantization 4bit` measurement will see a discrepancy with no stated cause.

It also silently doubles the judge's memory footprint relative to the training configuration, which matters on exactly the small cards this project targets.

## Fix path

Thread a quantisation argument through `load_model_and_tokenizer` and default it from the run's own configuration where one is available (`soup ship --config`, the registry entry, or the adapter's `adapter_config.json` base). Where no configuration is available, keep bf16 and **say so in the output** rather than leaving it implicit.

## Acceptance criteria

- [ ] `load_model_and_tokenizer(..., quantization="4bit")` loads a 4-bit base and the adapter attaches to it
- [ ] `soup ship` reports which numerics the judge used, in the verdict output and in `--emit-evidence`
- [ ] A test that the reported numerics match the requested ones — not merely that the call does not raise
- [ ] Evidence emitted under one numerics and replayed under another is refused or flagged, in the same spirit as the existing `config_sha` staleness gate (#312 lineage)

MakazhanAlpamys:
Partially addressed by @AmirF194 in #461, merged as `de6445b` — **and this issue stays open on purpose.**

[...]

Issue #340, opened 2026-08-07
There is no full-fine-tuning mode in the SFT trainer: the else branch always applies LoRA


MakazhanAlpamys:
`SFTTrainerWrapper._setup_transformers` has no path that trains a model without adapters. The `else` branch applies LoRA unconditionally, and there is no `lora.r == 0` bypass, so a user who wants plain full fine-tuning has to write:

```yaml
training:
  unfrozen_parameters: ['.*']
```

which is the Spectrum targeted-training feature (v0.71.23 #266) being used as a workaround. It works — that is how the full-FT arm of #306 was measured — but nothing documents it, and `unfrozen_parameters` reads as "train these specific tensors", not "train everything".

### Fix shape
Either accept `lora.r: 0` (or an explicit `training.full_finetune: true`) as a first-class full-FT switch, or document `unfrozen_parameters: ['.*']` as the supported spelling. The first is clearer; the second costs nothing.

### Acceptance
- [ ] a documented way to full-fine-tune that does not go through a regex feature
- [ ] a test that the chosen spelling produces a model with no PEFT wrapper and every decoder parameter trainable
- [ ] `docs/training.md` says which one it is

Found while measuring #306, where a full-FT arm was needed as the comparator.

MakazhanAlpamys:
Implemented in `dcb5eb5`, CI green. The spelling is **`lora.r: 0`**, chosen on repo evidence rather than taste.

```yaml
training:
  lora:
    r: 0        # full fine-tuning - no adapter
```

Three consumers already treated rank 0 as "no adapter", so this needed no changes outside the trainer, while a separate `full_finetune: true` flag would have left all three lying:

| consumer | with `r: 0` | with a separate flag (default `r: 64`) |
|---|---|---|
| `trainer/classifier.py` | already this spelling — one convention | a second spelling for the same idea |
| `commands/card.py` | reports "dense model" | reports a full-FT run as a **LoRA adapter** |
| `commands/train.py` pre-flight | prints `LoRA: r=0` | prints `LoRA: r=64` — a lie about a real rank |

**Decisive safety check:** `r: 0` **crashed** before this change (peft: `r should be a positive integer value but the value passed is 0`), captured in the RED run. So no config that worked before changes meaning.

[...]

PR #429, opened 2026-08-16
Fix fp16 GradScaler crash on pre-Ampere GPUs (cast bf16 LoRA adapters to fp32)

lesterppo:
## Problem

`stream_layers: true` on a pre-Ampere GPU (T4, P100, V100, GTX 16xx, RTX 20xx) crashes on the first training step:

```
NotImplementedError: "_amp_foreach_non_finite_check_and_unscale_cuda" not implemented for 'BFloat16'
```

Reported in detail in #425.

## Root cause

`pick_mixed_precision` correctly returns `fp16` below `BF16_MIN_CC`, so `TrainingArguments(fp16=True, bf16=False)` and the GradScaler are fp16. But peft's `get_peft_model` creates the LoRA adapters in the **base checkpoint's dtype** — bf16 for Llama-3.1 — independent of the stream dtype. On a pre-Ampere card the stream + mixed precision are fp16 while all 128 trainable adapter params stay bf16, so the fp16 `GradScaler.unscale_()` dispatches the `foreach` kernel against bf16 gradients and raises.

Two probes pinned down why an in-place cast inside `build_streamed_model` is not enough:

1. `materialize_meta_adapters` (which re-inits to fp32) never corrects this — the adapters are **not** `meta` (peft allocates real tensors), so its `if not param.is_meta` guard skips them.
2. A cast added inside `build_streamed_model` *does* fire (`adapters_off_dtype 128 -> 0`), but the adapters are re-cast back to bf16 during trainer construction — so the cast must happen after the trainer exists.

## Fix

Cast trainable params to fp32 immediately before `self.trainer.train(...)`: after `accelerator.prepare()` has run (trainer construction) and before the lazily-created optimizer reads the params. Gated on `fp16 and not bf16` so the bf16 (Ampere+) and fp32 (CPU) paths are untouched.

fp32 deliberately matches the author's own `materialize_meta_adapters` choice and provides fp32 master weights for stable LoRA updates.

## Verification (free Colab T4, sm_75)

- Same config now completes: 7 steps, peak **2.86 GB**, adapter **128/128 tensors non-zero**.
- Also exercised end-to-end on a real fine-tune (Llama-3.1-8B → `gbharti/finance-alpaca`, resident QLoRA, 57 steps, loss 1.954 → 1.797).

Note: the fix lives in the shared `train()` path, so it also covers any non-streamed pre-Ampere fp16 run that hits the same bf16-adapter condition; on a standard QLoRA load the adapters are already fp32 and the loop is a no-op.


MakazhanAlpamys:
Status note so this is not sitting silently: your PR is **in review, not ignored**. CI is unblocked (first-time and fork runs need manual approval here, which I have done).

I have a review in progress but stopped it mid-verification rather than post half-checked conclusions — this project's bar is that a finding is something I ran, with a control proving the test suite can fail at all, and I am not there yet on yours. Rather than give you conclusions I would have to retract, I would rather come back with the finished thing.

Picking this up shortly. Nothing needed from you in the meantime.

MakazhanAlpamys (changes requested):
Thank you for filing #425 with a full traceback and running **two elimination probes before proposing anything** — that is better diagnostic work than most PRs arrive with, and two of your judgement calls are ones I would have had to ask for otherwise: fp32 rather than fp16 (correct, and consistent with `materialize_meta_adapters`, peft's own `autocast_adapter_dtype`, and `prepare_model_for_kbit_training`), and `_p.data = ...` rather than `.to()`, which preserves `Parameter` identity for any optimizer group already holding a reference. Probe #1 is exactly right: `materialize_meta_adapters` skips non-meta params at `layer_stream_runtime.py:1372`.

Everything below was measured in an isolated checkout with `PYTHONPATH` pinned to your branch.

## 1. The fix is in the wrong place — and this is the third time

The cast landed in `sft.py::train()` only. dpo/orpo/simpo/kto reach the identical condition:

- **Same fp16 decision** — all take it from `bf16_fp16_flags` (`utils/gpu.py:298-319`), consumed at `dpo.py:160/188`, `orpo.py:165/193`, `simpo.py:165/193`, `kto.py:155/183`.
- **Same adapter creation, by construction** — all four stream through the *shared* `StreamingSetupMixin._setup_streaming_transformers` (`stream_setup.py:145`), which builds the `LoraConfig` once at `:376-391`. Their adapter dtype cannot differ from SFT's.
- **The fix is not shared** — each wrapper has its own `train()`: `dpo.py:372`, `orpo.py:356`, `simpo.py:358`, `kto.py:350`.

Two scans settle it:

```
$ grep -rn "to(torch.float32)" src/soup_cli/trainer/
src/soup_cli/trainer/sft.py:1537        # the only one

$ grep -rn "self.trainer.train(" src/soup_cli/trainer/*.py
19 call sites across 18 modules
```

18 of 19 are unprotected, and the class is wider than the four streaming preference losses — ipo/bco/reward_model/pretrain/embedding/grpo/ppo/distill/classifier all set `fp16` from the same helper and all call `get_peft_model`.

CLAUDE.md records this exact shape twice: **#328** (a helper called only by `sft.py`, four preference losses left out — "one omission, two opposite symptoms") and **#336** (DeepSpeed+LoRA repaired "in `sft.py` only", still a standing Known Limitation). The counter-precedent is the v0.73.0 `resolve_device_map` work, where a hand-written parametrize list missed nine sites and the guard was rewritten to **scan every module in `trainer/`**.

Please extract to `utils/mixed_precision.py::align_trainable_dtype_for_fp16(model, *, fp16, bf16) -> int`, call it from every wrapper, and add a **scanner** test rather than a hand-written list.

## 2. The stated mechanism does not reproduce on this repo's pins

This is the part I most need your help with. On peft 0.18.1 / transformers 4.57.6 / torch 2.5.1+cu121 / trl 0.19.1 — the dev pins CI runs — I measured:

| Your claim | Measured here |
|---|---|
| `get_peft_model` creates adapters in the base dtype (bf16) | bf16 base → **`{'torch.float32': 8}`**. peft's `autocast_adapter_dtype=True` default upcasts them in `PeftModel.__init__` → `_cast_adapter_dtype` (`peft/peft_model.py:131-133`) |
| the adapters are **not** meta | streaming-shaped build (`init_empty_weights`, bf16) → **8/8 meta**. `materialize_meta_adapters` *would* fire |
| (implied) the streamed base on a T4 is bf16 | post-#385, `resolve_stream_dtype` → `get_compute_dtype()` → **float16** on pre-Ampere (`layer_stream.py:203-210`) |

So on these pins `_p.dtype == torch.bfloat16` is never true for a trainable param and **the block is a measured no-op**. Your crash is real — you have the traceback — but the mechanism is unexplained, and the difference is almost certainly your stack (torch 2.11.0+cu128, so a much newer peft).

That matters beyond a review nit: it is the **#385/#387 trap in a new guise**. CLAUDE.md records that the first pre-Ampere fix was inert on the exact hardware it targeted; here the fix would be inert on the pins CI runs, so CI could never exercise it and it could rot silently.

Could you paste `pip freeze | grep -E "peft|transformers|trl|torch|accelerate"`, plus a dump of `[(n, p.dtype, p.is_meta) for n, p in model.named_parameters() if p.requires_grad][:5]` at the crash point? That decides whether this cast is the right fix or whether something upstream moved.

## 3. Your probe #2 may have found something bigger than the crash

If adapters really are non-meta on your stack, then `materialize_meta_adapters` returns **0** there. Its own docstring (`layer_stream_runtime.py:1357-1363`) says that is precisely the case where *"the optimizer happily accepts them and the run trains nothing."*

That would be a silent no-training regression on newer peft for every streamed run — the same class as #392 and v0.72.0's `.inner.` adapter keys, and arguably worse than the crash you filed, because a crash announces itself and this would not. Please open a separate issue with that dtype/meta dump either way; it is worth tracking independently of this PR.

## 4. Scope the cast — as written it is unbounded

The loop casts *every* trainable bf16 param, not just adapters. Under `lora.r: 0` full fine-tuning (v0.73.0 #340), Spectrum `unfrozen_parameters`, or LISA, that is a large fraction of the model — CLAUDE.md measured LISA's always-on embeddings + head + final norm at **70.7% of everything it trains at 8B**. On a pre-Ampere card this silently doubles trainable-weight memory at the moment training starts, *after* the VRAM pre-flight has already passed (`stream_vram_probe` runs in `setup()`, this runs in `train()`).

Suggest filtering on `"lora_" in name`, or refusing loudly above a threshold.

Minor, for the CHANGELOG rather than the code: peft saves adapter params at their live dtype, so on stacks where adapters were bf16 the saved `adapter_model.safetensors` becomes fp32 (2x). Not a correctness problem — peft casts on load — but worth a line.

## 5. No tests, and I proved that rather than assuming it

| Mutation | Result |
|---|---|
| `.to(torch.float32)` → `.to(torch.bfloat16)` (cast neutered) | **survived** — 475 passed, identical to baseline |
| delete all 22 lines (full revert) | **survived** — 373 passed |
| **control, same file**: break `_resolve_mixed_precision` | **caught** — `test_part_e.py::TestResolveMixedPrecision::test_auto_flag_picks_fp16_for_qwen2_on_ampere` |

Worth noting how nearly this went wrong: `test_issue385_stream_dtype.py` did **not** catch the control — only `test_part_e.py` did. A file set without it could not have failed on `sft.py` at all, and the "no tests" conclusion would have been unearned. The revert result above was re-run with `test_part_e.py` included.

**A T4 is not required.** I measured that `torch._amp_foreach_non_finite_check_and_unscale_` does **not** raise for bf16 on CPU — the failing primitive is CUDA-kernel-specific — so assert the *decision* with the capability stubbed. That is exactly what `tests/test_issue385_stream_dtype.py` does, and its docstring says so: *"These tests assert the DECISION, with the capability stubbed, because that is the part that was wrong."*

Sketch, once the block is extracted:

```python
def test_bf16_trainable_cast_to_fp32_under_fp16():
    m = _model()
    assert align_trainable_dtype_for_fp16(m, fp16=True, bf16=False) == 1
    assert m.lora_A.dtype is torch.float32

def test_frozen_params_are_not_cast():        # control: bounds the blast radius
def test_fp16_trainable_is_left_alone():      # control: GradScaler handles fp16 natively
@pytest.mark.parametrize("fp16,bf16", [(False, True), (False, False)])
def test_no_cast_off_the_fp16_path(fp16, bf16):   # control: the guard, both other modes

def test_every_trainer_that_trains_also_aligns():
    """The #328/#336 guard: scan, do not hand-write a list."""
    root = pathlib.Path("src/soup_cli/trainer")
    missing = [p.name for p in root.glob("*.py")
               if "self.trainer.train(" in (s := p.read_text(encoding="utf-8"))
               and "align_trainable_dtype_for_fp16" not in s]
    assert not missing, f"trainers call train() without fp16 dtype alignment: {missing}"
```

The last one is the test whose absence let the sft-only placement through, and it needs its own control — temporarily drop the helper from one wrapper and confirm the scanner fails.

## Also missing

**CHANGELOG.md** — the PR touches exactly one file. Release Checklist step 7 requires an entry for a user-facing crash fix. A pre-Ampere note in `docs/backends-and-ops.md` or `docs/performance-and-quantization.md` would help too. Ruff is clean.

lesterppo:
Thanks for the review — every point lands. I've pushed a rewrite and opened #433 for the meta/non-meta finding. Point by point:

## 2. The mechanism — you're right, it's the stack

You called it: the difference is peft 0.19.1 vs your 0.18.1 pins. On my stack:

- `torch 2.11.0+cu128` (CUDA 12.8), **`peft 0.19.1`** (from the saved `adapter_config.json`), Python 3.12.13. I don't have the exact `transformers`/`trl`/`accelerate` patch versions — the Colab session was recycled before I captured a full `pip freeze` — but the `peft` version is the operative difference and it's pinned in the adapter metadata.
- Dump at the crash point (`[(n, p.dtype) for n, p in model.named_parameters() if p.requires_grad]`): `{'torch.bfloat16': 128}` — every trainable param is `*.lora_A/lora_B.default.weight`, bf16.
- `is_meta`: I did **not** capture it explicitly, but it's decidable from the outcome — after the cast, the run writes **128/128 non-zero** adapter tensors. A meta param has no storage and cannot be updated, so that result proves the adapters were real (non-meta) at build time. That is the non-meta half I filed as #433.

So on 0.19.1 both the `autocast_adapter_dtype` upcast *and* the meta-allocation you measured on 0.18.1 are gone — adapters come out real and bf16. The cast is still the right fix for the crash (it only fires when a bf16 adapter actually exists, so it's a no-op on 0.18.1 where they're already fp32), but you're correct that on the pinned stack the block is inert and CI alone can't exercise it — hence the decision-level tests below, which don't need the new peft.

## 1 & 3. Extracted + wired everywhere; #433 opened

- `utils/mixed_precision.py::align_trainable_dtype_for_fp16(model, *, fp16, bf16) -> int`, scoped to `requires_grad and "lora_" in name and dtype == bfloat16`.
- Called from **all 19 `self.trainer.train(` sites across 18 modules** (asr, bco, classifier, distill, dpo, embedding, grpo, ipo, kto, mole_routing, online_dpo, orpo, ppo ×2, pretrain, prm, reward_model, sft, simpo), reading `self.trainer.args.fp16/.bf16` so the guard is uniform.
- **#433** opened for the `materialize_meta_adapters` returning-0-on-newer-peft invariant, with the dtype/meta dump and the "128/128 non-zero ⇒ real" reasoning.

## 4. Scoped

The loop now filters on `"lora_" in name`, so full fine-tuning (`lora.r: 0`), Spectrum `unfrozen_parameters` and LISA are untouched. `test_non_lora_bf16_trainable_is_not_cast` pins it.

## 5. Tests

`tests/test_align_trainable_dtype.py`, decision-level (no GPU — matches the `test_issue385_stream_dtype.py` pattern):

- `test_bf16_trainable_cast_to_fp32_under_fp16` (== 1, `lora_A.dtype is torch.float32`)
- `test_frozen_params_are_not_cast`
- `test_fp16_trainable_is_left_alone`
- `test_non_lora_bf16_trainable_is_not_cast` (blast radius)
- `test_no_cast_off_the_fp16_path` (parametrized `(False, True)` / `(False, False)`)
- `test_every_trainer_that_trains_also_aligns` (scanner)

All 7 pass (I ran them locally under CPU torch 2.13 in a throwaway venv). The scanner's control: I dropped the helper from a synthetic trainer file and it was flagged, then removed the probe.

## Also

- **CHANGELOG.md** entry under `[Unreleased] → Fixed`.
- **docs/performance-and-quantization.md** — a pre-Ampere note in the layer-streaming section.

One thing I did not do: the "refuse loudly above a threshold" alternative from point 4, because the `lora_` filter makes the cast bounded by construction. Happy to switch to the loud-refusal form if you'd prefer that invariant be enforced rather than implied.

Ruff is clean.

MakazhanAlpamys (changes requested):
This is exactly the shape I asked for, and **the scanner is the real thing rather than decoration** — which was the whole question, so I tested it two ways rather than reading it.

Verified in an isolated worktree with `PYTHONPATH` pinned to your branch:

| Probe | Result |
|---|---|
| remove the helper call **and import** from `orpo.py` | **fails**, naming it: `trainers call train() without fp16 dtype alignment: ['orpo.py']` |
| drop in a brand-new `trainer/_zzz_fake.py` calling `self.trainer.train(` | **fails**: `['_zzz_fake.py']` |

The second probe is the one that matters: it proves this is a **scan**, not a hand-written list. The v0.73.0 `resolve_device_map` failure was exactly a parametrize list that missed nine sites, and this would not have.

## All five findings closed

Coverage confirmed by count: 19 `self.trainer.train(` sites across 18 modules, **all 18** carry the helper, and both of `ppo.py`'s sites are individually aligned.

The `lora_` scoping closes the blast radius from finding 4 — under `lora.r: 0` full-FT, Spectrum or LISA the old loop would have cast a large fraction of the model — and `test_non_lora_bf16_trainable_is_not_cast` pins it.

Five mutations, **no survivors**, against a baseline where neutering the cast previously left 475 passed *identical to baseline*:

| Mutation | Result |
|---|---|
| neuter the cast | caught |
| remove the call from one trainer | caught (scanner) |
| invert the fp16/bf16 guard | caught (2 tests) |
| drop the `requires_grad` filter | caught |
| drop the `lora_` filter | caught |

Tests: **234 passed / 1 skipped** and **397 passed** across two batches, deliberately including the source-grep files — a test of that shape caught a regression in another PR today that a file set chosen around the changed code had missed. ruff clean, CHANGELOG and docs both present, no deleted or weakened assertions anywhere in the diff.

## One thing to fix before I merge

`sft.py` has the helper call **duplicated verbatim** — two identical back-to-back blocks, and the diff adds both:

```python
align_trainable_dtype_for_fp16(
    self.trainer.model,
    fp16=getattr(self.trainer.args, "fp16", False),
    bf16=getattr(self.trainer.args, "bf16", False),
)
align_trainable_dtype_for_fp16(          # <- identical, second call
    self.trainer.model,
    ...
)
```

Harmless at runtime — the second is a no-op once the first has cast everything — but it is dead code, and `sft.py` is the only file where the count is wrong. `ppo.py`'s two calls are legitimately one per branch; every other module has exactly one. Delete one block and I will merge.

## Two notes I will merge with, not block on

**1. The scanner cannot see partial coverage inside a multi-site module.** It is a substring-presence check, so I removed **one** of `ppo.py`'s two calls and it stayed green (7 passed with 1 of 2 remaining). Both are correct today; the guard just would not catch one of them regressing. A count-based assertion for multi-site files would close it, if you are ever back in this file.

**2. Thank you for the straight answer on peft 0.19.1 vs 0.18.1, and for filing #433 separately rather than folding it in.** The honest consequence is worth writing down: on the pinned stack this cast is inert, so **CI structurally cannot exercise it**. Your decision-level tests are the correct mitigation and mirroring `test_issue385_stream_dtype.py` was the right call — that file's docstring says the same thing about the same class of problem.

I have replied on #433 as well. Short version: your "128/128 non-zero ⇒ the adapters had real storage" inference is sound and is actually *stronger* than a direct `is_meta` print, because it observes the property through its consequence. I did narrow the framing there — on your stack training worked, so what broke is a **guard silently ceasing to discriminate**, not training itself.

Going from a one-file patch to a 22-file extraction with a working scanner, and filing the adjacent invariant as its own issue instead of quietly bundling it, is a good response to detailed feedback.

MakazhanAlpamys:
CI is red on 7 cells. I reproduced both failures locally in an isolated worktree — **neither is a mistake in your reasoning**, and one of them is genuinely interesting.

```
FAILED tests/test_issue385_stream_dtype.py::TestEveryTrainerAsksTheCard::test_the_wrappers_that_set_precision_use_the_shared_helper
FAILED tests/test_v07132.py::TestAsrTrainSidecar::test_writes_sidecar_when_customized
FAILED tests/test_v07132.py::TestAsrTrainSidecar::test_no_sidecar_when_default
```

## 1. Your scanner collided with an older scanner, and the older one is too coarse

```
AssertionError: these set bf16 without the shared helper: {'mole_routing.py', 'prm.py', 'ppo.py'}
```

That is `test_issue385_stream_dtype.py:280-290`, from #385. It works like this:

```python
setters = {name for name, src in sources.items()
           if ("bf16=" in src or '"bf16":' in src) and name not in {"grpo.py", "__init__.py"}}
missing = {n for n in setters if "bf16_fp16_flags" not in sources[n]}
assert not missing, f"these set bf16 without the shared helper: {missing}"
```

It greps for the **substring** `bf16=`. Your helper call passes `bf16=getattr(self.trainer.args, "bf16", False)`, so `bf16=` now appears in three files that previously had no precision code at all — and they do not call `bf16_fp16_flags`, so the older scanner fires.

I checked whether those files actually set bf16, rather than assuming:

```
mole_routing.py:391:  bf16=getattr(self.trainer.args, "bf16", False)     <- your call, only hit
prm.py:349:           bf16=getattr(self.trainer.args, "bf16", False)     <- your call
ppo.py:569, :581:     bf16=getattr(self.trainer.args, "bf16", False)     <- your two calls
```

(`prm.py`'s other two hits are prose in a comment.) So none of the three sets a precision flag — they **read** one. **This is a false positive**: the #385 scanner cannot distinguish "sets bf16" from "passes a keyword argument named bf16", and your PR is what exposed that.

I would rather fix the older scanner than contort your API to satisfy a grep. Narrowing its pattern to a literal assignment — something like `bf16=True` / `bf16=False` / `"bf16":` — or excluding lines that call `align_trainable_dtype_for_fp16`, restores what it was actually trying to measure. Your call which, but please do it in this PR and say in the CHANGELOG that the #385 scanner was narrowed, so the next person does not read it as a weakening.

There is a nice symmetry worth noting: this is two guards-against-drift meeting, and the newer one is the more precise of the two.

## 2. `asr.py` reaches for `.model` on a fake trainer

```
AttributeError: 'types.SimpleNamespace' object has no attribute 'model'
src/soup_cli/trainer/asr.py:417
```

Your call site does `self.trainer.model`, but `TestAsrTrainSidecar` builds its trainer as a bare `SimpleNamespace` that has `args` and no `model`.

Your helper is already defensive about `args` — `getattr(self.trainer.args, "fp16", False)` — and this is the same class of assumption one level up. I would make the call site symmetric rather than patch the fixture:

```python
align_trainable_dtype_for_fp16(
    getattr(self.trainer, "model", None),
    fp16=..., bf16=...,
)
```

and have the helper return `0` on `None`. Patching the fixture would work too, but it teaches the test to satisfy the code rather than the reverse — and a trainer object without `.model` is a shape this codebase evidently produces.

## Still outstanding from my last review

The duplicated helper call in `sft.py` — two identical back-to-back blocks, both added by the diff. Harmless at runtime, dead code, and `sft.py` is the only file where the count is wrong.

Everything else stands: the scanner is real (proven two ways), all five original findings closed, five mutations and no survivors. This is three mechanical fixes from merge.

MakazhanAlpamys:
Checking in — no pressure, just so this does not drift.

Where it stands: the branch is **conflicting** with `main` (a lot has landed since 16 Aug) and CI is red on 9 cells, but nothing new has been found against your reasoning. The list from my last review is unchanged and still three mechanical fixes:

1. **Narrow the #385 scanner** (`tests/test_issue385_stream_dtype.py:280-290`) to a literal `bf16=True` / `bf16=False` / `"bf16":` assignment, or exclude lines calling `align_trainable_dtype_for_fp16` — and say in the CHANGELOG that it was narrowed, so nobody later reads it as a weakening.
2. **`asr.py` call site symmetric with the helper** — `getattr(self.trainer, "model", None)`, helper returns `0` on `None`. `TestAsrTrainSidecar` builds its trainer as a bare `SimpleNamespace`, and a trainer object without `.model` is a shape this codebase evidently produces.
3. **Drop the duplicated helper call in `sft.py`** — two identical back-to-back blocks, both from this diff.

Plus a rebase on `main` for the conflict.

Two things worth saying plainly:

- **The diagnosis and the scanner are accepted.** I proved the scanner fires two ways (removing the call from `orpo.py`, and dropping in a brand-new trainer module), all five original findings are closed, five mutations and no survivors. This is not a PR I am looking for reasons to reject.
- **`notebooks/proof-4gb.ipynb` §5 is broken on a free T4 until this lands**, and `README.md` links it as the one thing a stranger can run to check the 4 GB claim. So it is on the critical path, not in a backlog.

If you have the time, take it — the credit stays yours either way. If you would rather hand it off, say so here and I will land the three fixes on top of your branch with authorship preserved. Either answer is fine; silence is the only one that costs us both.

MakazhanAlpamys:
Four days since your last push, so this is a concrete offer rather than another nudge.

**Nothing is wrong with your work.** The diagnosis is accepted, the scanner is accepted, and I proved it fires two ways myself. What is left is three mechanical items and a rebase:

1. Narrow the #385 scanner (`tests/test_issue385_stream_dtype.py:280-290`) to a literal `bf16=True` / `bf16=False` / `"bf16":` assignment, or exclude lines calling `align_trainable_dtype_for_fp16` — and say in the CHANGELOG that it was **narrowed**, so nobody later reads it as a weakening.
2. `asr.py` call site symmetric with the helper — `getattr(self.trainer, "model", None)`, helper returns `0` on `None`. `TestAsrTrainSidecar` builds its trainer as a bare `SimpleNamespace`, and a trainer object without `.model` is evidently a shape this codebase produces.
3. Drop the duplicated helper call in `sft.py` — two identical back-to-back blocks, both from this diff.

Plus a rebase; `main` has moved a fair amount (#409, #416, #455's area, #460).

## The offer

**Say the word and I will land those three on your branch myself, with your authorship preserved** — your commits stay yours, mine goes on top as a separate fixup commit, and the PR merges under your name with the credit in `CONTRIBUTORS.md` as normal. I am offering because this is not a design question any more; it is typing, and the thing it fixes is not small.

**Why I care about the timing.** This is the pre-Ampere fp16 path — T4 / P100 / V100 / GTX 16xx, which is the entire free Colab and Kaggle tier. You reported #425 yourself from a real crash. Every day this sits, that is the hardware class that cannot run `stream_layers: true`.

If you would rather finish it yourself, that is genuinely better and I will wait — just say so here and take the time you need. What I do not want is for it to stall silently because you assumed I had lost interest.

Related, so you have the full picture: **#433 — the meta/non-meta finding you opened from this review — is fixed and shipped** (v0.73.3, by @Faisal01011). Your report produced a merged fix in someone else's PR, which is worth saying out loud.

MakazhanAlpamys:
@lesterppo — five days since your last push (16 Aug) and two days since I offered to finish this. No reply, which is completely fine; people have lives. But I do not want to leave it hanging in silence, so here is a **dated plan** rather than a third nudge.

**If I have not heard from you by Monday 24 August**, I will land the three mechanical items on your branch myself:

1. Narrow the #385 scanner (`tests/test_issue385_stream_dtype.py:280-290`) to a literal `bf16=True` / `bf16=False` / `"bf16":` assignment, or exclude lines calling `align_trainable_dtype_for_fp16` — with a CHANGELOG line saying it was **narrowed**, so nobody later reads it as a weakening.
2. Make the `asr.py` call site symmetric with the helper — `getattr(self.trainer, "model", None)`, helper returns `0` on `None`. `TestAsrTrainSidecar` builds its trainer as a bare `SimpleNamespace`, and a trainer without `.model` is a shape this codebase evidently produces.
3. Drop the duplicated helper call in `sft.py` — two identical back-to-back blocks, both from this diff.

Plus a rebase; `main` has moved a long way (#320, #367, #443, #453, #459's sibling, the Qwen3.5 MoE streaming work).

**What that means concretely, so there is no ambiguity:** your commits stay yours and stay first in the history. Mine goes on top as a separate fixup commit. The PR merges under **your** name, the squash body credits you, and `CONTRIBUTORS.md` gets your entry as normal. You are not losing the contribution — I am typing the last 5% of it.

I am saying this in advance and in public because our own CONTRIBUTING.md asks for it and we broke that rule once. I would rather over-communicate than repeat it.

**If you would rather finish it yourself, just say "mine" here** — even after Monday, even with no timeline attached — and I will stop and wait. That is genuinely the better outcome and I am not trying to push you off it.

**Why the deadline exists at all**, and it is not impatience: this is the pre-Ampere fp16 path — T4, P100, V100, GTX 16xx — which is the whole free Colab and Kaggle tier. You found it yourself and reported it as #425 from a real crash. Every day it sits, that is the hardware class that cannot run `stream_layers: true`.

Worth repeating since you may not have seen it: **#433, the meta/non-meta finding you opened out of this review, is fixed and shipped** in v0.73.3. Your report produced a merged fix in someone else's PR. That one is already banked regardless of what happens here.

MakazhanAlpamys:
Small tracking note while this waits, found in a closing-link audit rather than in the code.

This PR says *"Reported in detail in #425"* — prose, not a closing keyword. GitHub therefore links **nothing**:

```
#429 closes: none
```

So when this merges, **#425 stays open** — the crash you reported yourself, fixed by your own PR, still sitting in the tracker. Adding `Closes #425` to the description takes a second and means neither of us has to remember later.

Worth saying that this is a real category and not pedantry: I audited every open PR the same way just now, and the other four with a fix-to-issue relationship all link correctly. Yours and one other were the exceptions, and the other one is deliberate (#475 is a partial fix, so `Refs` is right there).

No change to the Monday plan — this is one line in the description, not work.

MakazhanAlpamys:
﻿@lesterppo - no word since 16 Aug, so per the plan stated here I am landing the three mechanical items on your branch today: narrowing the #385 scanner (prm/mole_routing/ppo are flagged only because they take the decision through another module), the asr.py getattr guard, dropping the duplicated call block, plus rebase over current main. Full credit stays yours - the design, the 19-site coverage and the scanner test are all yours; I will mark the mechanical commits clearly in the squash body. If you come back mid-flight, say so and I will step off immediately.

MakazhanAlpamys:
Merged as squash on a fully green matrix (13/13), under your name as promised. The squash body states explicitly which parts are yours - everything that matters: the diagnosis from your real Colab crash, the shared-helper design, the lora_-scoping decision that protects full-FT memory, and the scanner - and that my commit only carried the three mechanical review items plus the rebase. Your report in #425 also produced the merged #433 fix earlier; both are credited. Welcome to the release notes.

Issue #423, opened 2026-08-16
detect_device() does not know MLX: an Apple Silicon run reports 'CPU (no GPU detected)' and silently downgrades quantization 4bit -> none


MakazhanAlpamys:
Split out of #394, reported by @riviangeralt with a reproduction on macOS arm64 (M1), Python 3.12, `soup-cli[mlx]` 0.73.2, `mlx` 0.32.0, `mlx-lm` 0.31.3.

## What happens

On a `backend: mlx` run, `soup train` prints:

```
Device:  CPU (no GPU detected)
Warning: 4bit quantization is not supported on CPU. Switching to quantization: none.
```

…on an Apple Silicon machine where MLX is about to use the GPU. The configured `quantization: 4bit` is silently rewritten to `none`.

## Why

`detect_device()` (`src/soup_cli/utils/gpu.py:97`) only knows two accelerators:

```python
if torch.cuda.is_available():          ...
if torch.backends.mps.is_available():  ...
```

Both are **torch** probes. MLX has its own device and its own runtime; a `[mlx]`-only install may have no torch at all, and even with torch present `torch.backends.mps` is not what an MLX run uses. So the one accelerator an MLX run *does* have is invisible to the detector, the panel reports CPU, and the CPU-quantization guard fires.

## Why this is worse than a cosmetic wrong label

The panel line is wrong, which is bad. The rewrite is worse: **a correct configuration is silently changed and the run completes successfully.** The user asked for a 4-bit run, the tool said "switching to none", and unless they read that line in the scroll-back they have a model trained differently from the config they committed.

That is the same failure class as #392 (MLX adapters saved with an unresolved `target_modules`, loss curve healthy, artifact dead) and #362 before it. Third time in this subsystem that a run reports success while quietly not doing what the config says.

## Fix path

`detect_device()` should recognise MLX when the backend is MLX — probing `mlx.core.metal.is_available()` (or `mlx.core.default_device()`) behind a lazy import, so the light CLI path is unaffected and a box without MLX is unchanged.

Then the CPU-quantization guard must not fire for a device MLX handles. Note this is not merely "skip the warning": MLX 4-bit is a *different* mechanism from bitsandbytes NF4 — an `mlx-community` checkpoint is already quantized — so the right behaviour may be to leave `quantization` alone on the MLX path entirely rather than to approve it. Whoever takes this should decide that explicitly rather than by making the warning disappear.

## Acceptance criteria

[...]

PR #431, opened 2026-08-16
harden: keep the SFT import off the MLX dispatch route

Shutaru:
## What changed

- Keep the Transformers SFT wrapper import behind the non-MLX dispatch path.
- Add a CLI regression test for the standalone MLX runtime boundary.
- Add an Apple Silicon CI smoke job that explicitly verifies `mlx` and `mlx_lm` imports, confirms that the PyTorch/TRL training stack is absent, and runs a one-step MLX SFT command.
- Document the current MLX device-detection limitation tracked in #423.
- Declare the runtime `transformers` dependency directly in the `[mlx]` extra.
- Add an unreleased changelog entry.

## Why

The SFT wrapper is import-light today, but the MLX dispatch route should not depend on it remaining that way. This change hardens the standalone MLX runtime boundary against future imports of the PyTorch/TRL training stack.

## Scope

This PR does not claim to resolve either reported variant of #394. In particular, the torch-present hang remains unpinned and #394 should remain open pending a stack dump or a reliable reproducer.

The incorrect Apple Silicon device detection and possible `4bit` to `none` rewrite remain tracked separately in #423.

## Validation

- Targeted runtime-isolation tests: 24 passed
- Packaging and metadata tests: 26 passed
- Ruff: passed
- Workflow YAML and `pyproject.toml`: parsed successfully

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,27 @@ jobs:
         run: |
           mypy || echo "::warning title=mypy::Type issues reported (non-blocking baseline — see the log above)"
 
+  mlx-smoke:
+    # #394: prove the documented standalone install on an Apple Silicon runner.
+    # Keep this separate from the main matrix because `.[dev]` deliberately
+    # installs the PyTorch/TRL stack and would hide an accidental MLX dependency.
+    runs-on: macos-14
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install only the MLX runtime and smoke-test dependency
+        run: pip install -e ".[mlx]" pytest
+      - name: Verify the PyTorch training stack is absent
+        run: >-
+          python -c "import importlib.util;
+          names=('torch','trl','datasets','peft','accelerate','bitsandbytes');
+          present=[name for name in names if importlib.util.find_spec(name)];
+          assert not present, f'unexpected training dependencies: {present}'"
+      - name: Run one-step MLX SFT through the real CLI
+        run: pytest -o addopts= -m smoke -k mlx_sft_smoke -q tests/test_smoke_train.py
+
   test:
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,12 +60,10 @@ jobs:
           python-version: "3.12"
       - name: Install only the MLX runtime and smoke-test dependency
         run: pip install -e ".[mlx]" pytest
-      - name: Verify the PyTorch training stack is absent
-        run: >-
-          python -c "import importlib.util;
-          names=('torch','trl','datasets','peft','accelerate','bitsandbytes');
-          present=[name for name in names if importlib.util.find_spec(name)];
-          assert not present, f'unexpected training dependencies: {present}'"
+      - name: Verify MLX imports and the PyTorch training stack is absent
+        run: |
+          python -c "import mlx.core, mlx_lm; print(mlx.core.__version__, mlx_lm.__version__)"
+          python -c "import importlib.util; names=('torch','trl','datasets','peft','accelerate','bitsandbytes'); present=[name for name in names if importlib.util.find_spec(name)]; assert not present, f'unexpected training dependencies: {present}'"
       - name: Run one-step MLX SFT through the real CLI
         run: pytest -o addopts= -m smoke -k mlx_sft_smoke -q tests/test_smoke_train.py
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,16 @@ reproducing 70+ versions of notes.
   with `deepseek-ai/DeepSeek-V4-Flash` (#279).** The recipe combines the
   established GRPO defaults with MoE LoRA and gradient checkpointing.
 
+### Changed
+
+- **The MLX SFT dispatch route no longer imports the Transformers SFT wrapper
+  before choosing a backend (#431 by @Shutaru).** The wrapper is import-light
+  today, but the standalone `soup-cli[mlx]` runtime no longer depends on it
+  remaining so. An additive Apple Silicon CI job verifies that `mlx` and
+  `mlx-lm` import, the PyTorch/TRL training stack is absent, and a one-step real
+  CLI SFT run completes. This hardens the runtime boundary; it does not claim to
+  resolve the still-unpinned torch-present hang reported in #394.
+
 ### Fixed
 
 - **A run whose watcher died was reported `running` forever (#401).**
--- a/docs/backends-and-ops.md
+++ b/docs/backends-and-ops.md
@@ -88,6 +88,11 @@ Transformers backend. A Hugging Face `datasets` source or streaming dataset
 still needs `datasets` because that data source owns the dependency; the local
 file path below does not.
 
+Known limitation: until [#423](https://github.com/MakazhanAlpamys/Soup/issues/423)
+lands, this standalone path can misidentify Apple Silicon as CPU and silently
+rewrite `training.quantization: 4bit` to `none`; verify the printed pre-flight
+configuration before relying on a 4-bit run.
+
 ```yaml
 base: mlx-community/Llama-3.2-3B-Instruct-4bit
 task: sft
--- a/docs/backends-and-ops.md
+++ b/docs/backends-and-ops.md
@@ -82,6 +82,12 @@ Fine-tune on M1-M4 Macs via Apple's [MLX](https://github.com/ml-explore/mlx) fra
 pip install "soup-cli[mlx]"
 ```
 
+For SFT with local JSONL, JSON, or CSV data, `[mlx]` is a standalone install:
+do not add `[train]`. The latter is the PyTorch/TRL stack used by the
+Transformers backend. A Hugging Face `datasets` source or streaming dataset
+still needs `datasets` because that data source owns the dependency; the local
+file path below does not.
+
 ```yaml
 base: mlx-community/Llama-3.2-3B-Instruct-4bit
 task: sft
--- a/docs/models.md
+++ b/docs/models.md
@@ -100,7 +100,7 @@ no PyTorch. Add `[train]` to fine-tune, or install other extras only when you ne
 | `fast` | `pip install "soup-cli[fast]"` | Unsloth backend (2-5x faster, lower VRAM) |
 | `vision` | `pip install "soup-cli[vision]"` | Vision / multimodal fine-tuning (Pillow) |
 | `audio` | `pip install "soup-cli[audio]"` | Audio / speech fine-tuning (librosa, soundfile) |
-| `mlx` | `pip install "soup-cli[mlx]"` | Apple Silicon backend (mlx, mlx-lm) |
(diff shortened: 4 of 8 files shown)


MakazhanAlpamys:
Status note so this is not sitting silently: your PR is **in review, not ignored**. CI is unblocked (first-time and fork runs need manual approval here, which I have done).

I have a review in progress but stopped it mid-verification rather than post half-checked conclusions — this project's bar is that a finding is something I ran, with a control proving the test suite can fail at all, and I am not there yet on yours. Rather than give you conclusions I would have to retract, I would rather come back with the finished thing.

Picking this up shortly. Nothing needed from you in the meantime.

MakazhanAlpamys (changes requested):
Reviewed in an isolated checkout with `PYTHONPATH` pinned to your branch. **I went in expecting to find this unguarded and was wrong** — I should say that plainly, because it is the most important thing about the change: five mutations, all five caught, including a straight revert of the fix.

| Mutation | Result |
|---|---|
| revert the fix — import back to module scope | **caught** by `assert payload["sft_loaded"] is False` |
| invert the MLX dispatch (`mlx_routing.py:63`) | **caught** |
| remove `mlx` from the blocked roots | **caught** |
| make `_require_mlx()` not raise (`mlx_sft.py:89`) | **caught** |
| *diagnostic*: unblock `datasets` entirely | survived, as expected — confirms the property is real, not manufactured by the block |

Under the revert, `tests/test_cli_startup_is_light.py` passes 23/23 — so your new test is the **only** thing that catches it. That is the correct division of labour: the light-CLI guard covers light commands and structurally cannot see an import-light module imported early on a heavy one.

The CI matrix is untouched, and I checked rather than assumed: `git diff main --numstat -- .github/workflows/ci.yml` → **`21 0`**, zero deletions. Existing 11 cells, `on:` triggers, `fail-fast` and the `lint`/`type-check`/`test` job names all unchanged; `mlx-smoke` is purely additive. No `continue-on-error`, no `if:`, no `|| true`. `test_smoke_train.py` is additive too — `test_sft_smoke` and `test_dpo_smoke` are byte-identical to main, and there are no removed test/assert/marker lines anywhere in the PR.

## 1. It does not fix either reported variant of #394 — please retitle

Measured on your branch with nothing blocked: a real `backend: mlx` CLI run imports `torch` and `numpy` and **nothing else** from the training stack. `datasets`, `trl`, `peft`, `accelerate`, `bitsandbytes`, `transformers` — all absent. `torch`'s single traced call site is `utils/gpu.py:100`, inside `try/except ImportError`, so it degrades cleanly.

So:

- **The ImportError variant does not exist to be fixed** — matching the maintainer's own measurement in the first comment on #394.
- **@riviangeralt's hang is not addressed.** It reproduces 3/3 *with torch present*, and nothing here changes behaviour in that case. It still has no stack dump.

What the PR *does* deliver is what your own inline comment says: `sft.py` is import-light today, and this stops that being load-bearing forever. That is worth merging on its own terms. But the title "isolate standalone MLX training runtime" will read as closing a live defect, and merging it under that framing would quietly mark #394 as handled when the reported symptom is untouched.

Please retitle to something like "harden: keep the SFT import off the MLX dispatch route", and either close #394 as not-reproducible-as-filed or leave it explicitly open for the hang.

## 2. `mlx-smoke` is green when mlx fails to import

I ran your exact CI command in an environment without mlx:

```
$ pytest -o addopts= -m smoke -k mlx_sft_smoke -q tests/test_smoke_train.py
1 skipped, 2 deselected in 0.55s     → EXIT CODE 0
```

Green, having executed nothing — `pytest.importorskip("mlx.core")` at `tests/test_smoke_train.py:213-214`. A `pip install` that fails outright would fail the step, but an mlx that installs and then fails to import (arch/ABI mismatch, `mlx_lm`→`transformers` resolution) is invisible.

Your "verify the PyTorch training stack is absent" step is the right idea applied in one direction only. Please add the mirror in the same step:

```
python -c "import mlx.core, mlx_lm; print(mlx.core.__version__, mlx_lm.__version__)"
```

This project already has the lesson written down: `TestTheWindowsCiSkipStaysNarrow` exists because **a skipped test is the same colour as a passing one**.

## 3. One assertion cannot fail

`assert payload["loaded"] == []` (`:115`) is guaranteed by construction — the finder raises `ModuleNotFoundError` from `find_spec`, so a blocked module can never reach `sys.modules`.

The underlying property is nonetheless real: I unblocked `datasets` entirely and the test still passed, and my independent probe confirms `datasets` is never imported on the local-JSONL path. So this is a dead assertion, not a false one.

The two lines carrying the test are `:116` (`blocked <= {"torch","mlx"}`) and `:118` (`sft_loaded is False`). Either delete `:115`, or make the finder record-and-return-`None` for one module so it has a way to fail.

## 4. The docs claim is true, but it advertises an install that #423 breaks

The dependency claim checks out: local JSONL/JSON use stdlib `json`, CSV uses stdlib `csv` (`data/loader.py:73-79`), `datasets` is lazy at `:391`/`:453`, and `[mlx] = mlx>=0.20.0, mlx-lm>=0.31.3` covers `mlx_sft.py`'s imports.

But **#423 is open**, and it bites exactly this install: on a standalone `[mlx]` setup, `detect_device()` reports "CPU (no GPU detected)" and silently rewrites `quantization: 4bit` → `none`. `docs/models.md:103` now says `[train]` is not required, without naming that. Please either land #423 first or add one sentence to `backends-and-ops.md`.

Minor: `transformers` is relied on *transitively* through `mlx-lm` and is not declared in the extra.

## 5. Missing CHANGELOG

Not in the file list, no `#394` reference, `[Unreleased]` untouched — Release Checklist step 7. Please write it against what actually shipped rather than the reported symptom.

## Sequencing

Good news on that front: **#431 and #426 have zero file overlap.** #426 touches `stream_setup.py` / `layer_*.py` / `moe.py` and their tests; this touches `commands/train.py` and `test_smoke_train.py`. The only latent conflict is `[Unreleased]` once you add your entry, which is trivial. So this one is not blocked behind the #416 → #426 ordering and can land on its own schedule.

MakazhanAlpamys (approved):
Re-reviewed `c5a35f2b` in a fresh isolated checkout with `PYTHONPATH` pinned to your branch. **All five items addressed. Approving; merging as soon as the matrix finishes.**

Before the detail: you answered "this does not fix what it says it fixes" by **renaming the PR to match reality** and adding an explicit scope paragraph, rather than defending the original framing. That is the harder call and the right one, and it is worth saying plainly.

| # | Status | Evidence |
|---|---|---|
| 1. Retitle + #394 disposition | **FIXED** | No `Fixes/Closes #394` in the body **or** in any commit message — I grepped both, so merging cannot auto-close it. #394 stays open for the hang. CI matrix re-checked after the new commit: `git diff main --numstat -- .github/workflows/ci.yml` → `19 0`, zero deletions. |
| 2. `mlx-smoke` green on import failure | **FIXED** | Ran it here without mlx: the presence check exits **1**, and both lines together under `bash -e` make the **step** exit 1. The pytest step alone still exits 0 on a skip, but it can no longer be reached while green. |
| 3. Dead assertion | **FIXED** | Both `assert payload["loaded"] == []` and the `loaded` payload key removed. |
| 4. #423 caveat | **FIXED** | `backends-and-ops.md:91-95` names the issue, the CPU misidentification and the silent `4bit` → `none`. Bonus: the `transformers` bound in `[mlx]` is now byte-identical to `[train]`'s, so the two cannot drift. |
| 5. CHANGELOG | **FIXED** | Correct Keep-a-Changelog placement, and it disclaims the hang rather than over-claiming. |

## On #3 — I checked that deleting the line cost no coverage, rather than assuming it

Removing an assertion is the one edit that deserves suspicion, so I tested the replacement rather than trusting the argument. I added a **handled** (`try/except ImportError`) `datasets` probe on the MLX route — the realistic leak shape in this codebase, since an unhandled one would just crash — and the surviving line 114 caught it:

```
E  AssertionError: assert {'datasets', 'mlx', 'torch'} <= {'mlx', 'torch'}
E    Extra items in the left set: 'datasets'
tests\test_issue394_mlx_runtime_isolation.py:114
```

So the line you deleted was one that could not fail, and deleting it removed no coverage.

## Mutations and scope

Seven mutations. The straight revert — SFT import back to module scope — is still caught by `sft_loaded is False`, as are an inverted dispatch, a non-raising `_require_mlx()`, and both leak shapes. Two survived (dropping `transformers` from `[mlx]`, and deleting the new mlx presence line from `ci.yml`); both are inherent — nothing in this repo asserts an extra's contents, and a CI step's failure *is* its own signal. Controls in the same files fail correctly, so neither survival is a broken harness.

**25 test files, 1034 passed, 0 failed**, deliberately including every source-grep test that reads `commands/train.py` — `test_bugfixes.py::TestCPUQuantWarning` among them. That is not routine: a test of exactly that shape caught a regression in another PR today that a file set chosen around the changed code had missed. `ruff` clean. No deleted tests, no weakened assertions, no removed markers anywhere in the diff.

## The one thing my reviewer could not measure — and CI has now answered it

`test_mlx_sft_smoke` loads `hf-internal-testing/tiny-random-LlamaForCausalLM`, an HF torch-format checkpoint, through `mlx_lm` with torch absent. With no Apple Silicon runner here, whether that resolves on `macos-14` was flagged as an unknown rather than a finding.

**Your new `mlx-smoke` job has now passed on CI**, which settles it — and settles it in the way the job was designed to: by failing loudly if it had not, instead of skipping green. That is precisely the property that was missing before this commit.

Sequencing unchanged: no code overlap with #426; the only shared file is the `[Unreleased]` block, exactly as predicted.

MakazhanAlpamys:
Merged as [`e399868`](https://github.com/MakazhanAlpamys/Soup/commit/e399868) — **12/12 green**, including your new `mlx-smoke` job. You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) as of [`cc15577`](https://github.com/MakazhanAlpamys/Soup/commit/cc15577).

The credit line names two things, and only one of them is code.

**You retitled the PR when told it did not fix what it claimed.** That is the part most people get wrong — the easy move is to argue the framing, and merging under the original title would have quietly marked #394 as handled while the reported hang stayed untouched. I verified there is no `Fixes/Closes #394` in the body or in any commit message, so the merge could not auto-close it. **#394 stays open**, and I have left a note there explaining exactly what did and did not land.

**And the `mlx-smoke` job asserts presence, not just absence.** The earlier shape went green having executed nothing — `pytest.importorskip` gives `1 skipped`, exit 0. Now the explicit `import mlx.core, mlx_lm` check hard-fails first. That job passing on CI also settled the one thing neither of us could measure locally: whether an HF torch-format fixture resolves through `mlx_lm` on `macos-14`. It does — and if it had not, it would have failed loudly, which is the whole point.

One thing worth repeating from the review because it generalises: I did not accept the deletion of `assert payload["loaded"] == []` on the argument alone. Removing an assertion is the single edit that most deserves suspicion, so I tested the replacement — a **handled** `try/except ImportError` `datasets` probe, which is the realistic leak shape here — and the surviving line 114 caught it. So the line you removed genuinely could not fail, and removing it removed no coverage.

Your #426 is still open with its own findings; sequencing is unchanged (#416 first, then #426). No code overlap with this one, so nothing here disturbs it.

PR #435, opened 2026-08-17
fix(stream): verify adapter materialization

### LINKED ISSUE #433: materialize_meta_adapters returns 0 on peft >= 0.19 (adapters are real, not meta) — silent no-training risk
## Summary

On peft ≥ 0.19.x, `get_peft_model` on a streaming meta skeleton creates the LoRA adapters as **real (non-meta) bf16 tensors**, not meta tensors. `materialize_meta_adapters` therefore returns **0** on that path — but for the opposite reason its docstring describes.

`materialize_meta_adapters` (`layer_stream_runtime.py`, docstring at ~1357-1363) documents that returning 0 means *"the optimizer happily accepts them and the run trains nothing"* — i.e. it assumes adapters are meta and that a 0 return means they were left without storage. On peft 0.19.x that assumption inverts: the adapters are already real, so `if not param.is_meta` skips them and the function returns 0 while the adapters *do* have storage and *do* train.

This is version-sensitive in the exact shape of #385/#387: the re-init path fires on peft 0.18.1 (meta adapters) and silently does nothing on 0.19.x (real adapters), so nothing in CI (pinned to 0.18.1) exercises the newer behaviour.

## Evidence

Environment (free Colab T4, 2026-08-16):

- `torch 2.11.0+cu128` (CUDA 12.8), `peft 0.19.1` (from the saved `adapter_config.json`), Python 3.12.13.

At the crash point (first `clip_grad_norm_`), a dump of trainable params:

```
{'torch.bfloat16': 128}
base_model.model.model.layers.0.inner.self_attn.q_proj.lora_A.default.weight  (8, 4096)
...
```

Two observations:

1. **dtype is bf16** — the `autocast_adapter_dtype` upcast to fp32 that peft 0.18.1 performs in `PeftModel.__init__` → `_cast_adapter_dtype` does not happen on this path on 0.19.1. This is the direct cause of the fp16 GradScaler crash (#425).
2. **non-meta** — inferred, not directly dumped: with #429's cast applied the same run completes and writes an adapter with **128/128 tensors non-zero**. A meta param has no storage and cannot be updated by the optimizer, so a 128/128-non-zero adapter is proof the adapters were real (non-meta) at build time.

## Why this matters beyond #425

The crash (#425) announces itself. This one would not: the "adapters are meta but `materialize_meta_adapters` skipped them" failure — the case the docstring warns about — is silent no-training, and on 0.19.x the function now returns 0 as a matter of course, so its return value can no longer be used to detect that failure. If a future peft version (or a different streaming path) produces meta adapters that the `is_meta` guard misses, the run would train nothing with a healthy loss curve.

## Suggested direction (not implemented here)

Make the streaming build assert on the post-`get_peft_model` adapter state instead of trusting `materialize_meta_adapters`' return: after the build, verify every trainable `*lora_*` param has real storage (not `is_meta`) and fail loudly otherwise. Optionally make `materialize_meta_adapters` distinguish "no meta adapters to re-init (already real)" from "meta adapters that were skipped".

The dtype half of this is fixed by #429 (cast bf16 → fp32); this issue tracks the *meta/non-meta* half, which is independent.

Faisal01011:
## Summary

- verify the post-PEFT adapter state before installing the streaming runtime
- accept the healthy PEFT 0.19 shape where `materialize_meta_adapters()` returns zero because adapters already have storage
- refuse any trainable `lora_*` parameter still on `meta`, naming the stranded parameter
- document the new failure and the return-count ambiguity

## Root cause

`materialize_meta_adapters()` returns how many tensors Soup allocated, but PEFT versions disagree on whether allocation is needed. PEFT 0.18 creates meta adapters for Soup to materialize; PEFT 0.19 may create real adapters itself. A return value of zero therefore cannot distinguish a healthy no-op from a missed adapter.

The streamed build now checks the actual postcondition instead of interpreting that count as a guard. The existing integer return remains unchanged for compatibility.

## Regression coverage

The new tests synthesize both PEFT outcomes so they run on the repository's pinned PEFT and CPU-only CI:

- zero materialized + already-real adapter passes
- frozen meta base weights remain valid
- a trainable meta adapter fails and names `lora_A`
- the real `build_streamed_model()` call site is driven with a deliberately inert materializer, proving the guard fires before runtime installation

This implementation does not depend on the still-missing direct `is_meta` measurement under PEFT 0.19; `lesterppo` can confirm that environment separately without changing the invariant.

## Verification

- focused issue tests: 5 passed
- v0.72.0 streaming + #385 decision tests: 189 passed, 6 skipped
- v0.72.3 streaming tests: 132 passed, 10 skipped
- full non-smoke suite: 17,779 passed, 258 skipped, 5 deselected
- total coverage: 80.83% (required: 77%)

Closes #433

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,15 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **Layer streaming now verifies that every trainable LoRA parameter has real
+  storage after PEFT attaches the adapter (#433).** PEFT 0.18 creates streamed
+  adapters on `meta` for Soup to materialise, while PEFT 0.19 may create them as
+  real tensors immediately, so `materialize_meta_adapters()` returning `0`
+  cannot distinguish a healthy no-op from a missed adapter. The streamed build
+  now enforces the actual postcondition and refuses to install the runtime if a
+  trainable `lora_*` parameter remains on `meta`, naming the stranded parameter
+  instead of allowing a silent no-training run.
+
 - **`soup train --no-reexec` now prints the flags you actually typed (#372).**
   The advisory `accelerate launch` command is derived from the same argv the
   auto-reexec would have used, so `--fsdp` / `--deepspeed` / `--config` (and the
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -402,6 +402,12 @@ output: ./output
 > If that prints anything, the adapter is affected. From v0.72.1 a streamed adapter is byte-for-byte in the same layout as an ordinary LoRA run.
 
 **Troubleshooting:**
+- **"trainable LoRA parameters remain on the meta device"** — PEFT attached an
+  adapter without real storage and Soup refused the run before installing the
+  streaming runtime. This guard is deliberately based on the final parameter
+  state rather than on how many tensors Soup materialised: some PEFT versions
+  create real adapters themselves. Include the PEFT version and the named
+  parameter from the error when reporting this.
 - **"layer streaming needs the base to fit in RAM"** — the base is larger than free RAM. Set `stream_source: auto` to fall back to the NVMe disk tier, free RAM, or pick a smaller base.
 - **"layer streaming needs NVMe or more RAM … the detected disk is 'hdd'"** on a fast cloud disk — a virtio device reports `rotational=1` with no media hint. Detection now measures the disk when the flag is unreliable; if it still misreads yours, set `training.stream_disk_kind: nvme` to force the tier on (`ssd`/`hdd` force it off).
 - **"could not page-lock the base … falling back to a PAGEABLE RAM store"** — expected on a busy machine. Training continues, more slowly. Close other applications to keep the pinned store.
--- a/src/soup_cli/utils/layer_stream_runtime.py
+++ b/src/soup_cli/utils/layer_stream_runtime.py
@@ -1361,6 +1361,11 @@ def materialize_meta_adapters(model: Any, *, seed: int = 0, device: str = "cuda"
     build that device is ``meta``, so the adapters have no storage: the
     optimizer happily accepts them and the run trains nothing. Re-initialise
     with PEFT's own scheme (A ~ kaiming_uniform, B = 0).
+
+    The return value is the number of parameters materialised, not a success
+    signal.  In particular, zero is healthy when PEFT created real adapter
+    tensors itself.  Call :func:`assert_trainable_adapters_materialized` after
+    this function to enforce the actual postcondition.
     """
     import torch
     import torch.nn as nn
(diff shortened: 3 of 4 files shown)


MakazhanAlpamys:
Merged as [`66a3c0a`](https://github.com/MakazhanAlpamys/Soup/commit/66a3c0a) — 12/12 green. You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) as of [`7fce178`](https://github.com/MakazhanAlpamys/Soup/commit/7fce178), third merge and a real step up from the recipe work.

**The decisive check first**, since it is the whole point of the issue: I deleted `assert_trainable_adapters_materialized(model)` from `build_streamed_model` and the suite went red — `DID NOT RAISE <class 'RuntimeError'>`. **The guard is not decoration.** Six mutations, five killed: neutering the body (2 failed), dropping `requires_grad` from the filter, stripping the parameter name from the message (both `match=r"lora_A"` assertions fired, so "names the parameter" is real), and moving the guard ahead of `get_peft_model`.

That last one matters: it verifies the claim your own test docstring makes — that placing the check before PEFT attaches the adapter would make the test fail. Checked, not aspirational.

## On the three things I flagged when I assigned it

**1 — MET.** You stubbed the capability instead of chasing peft 0.19: `_adapter_model` builds the meta adapter synthetically, and the end-to-end test injects a fake `peft` into `sys.modules`. Five tests, **zero skips**, on the pinned 0.18.1 and CPU-only. That is the `test_issue385_stream_dtype.py` discipline applied without being handed the file.

**2 — MET, and you went one better than I asked.** You have the unit-level control *and* a call-site control that drives the real `build_streamed_model` and additionally asserts `install_reached is False`. That second assertion is the one I would have asked for in review: it pins that the refusal lands **before** the runtime is installed, so a future refactor moving the check downstream fails rather than passing on a technicality.

**3 — PARTIAL, and I am fine with the call you made.** You kept the count and demoted it in the docstring ("not a success signal… zero is healthy"), moving the decision into a separate postcondition. That is arguably better than a three-valued return, because the caller can no longer *choose* to ignore it. Deliberate divergence from a suggestion, with the reasoning visible in the code — that is the right way to disagree with a reviewer.

## One property you did not claim, and should have

Your guard's predicate is the **same** predicate `materialize_meta_adapters` selects on, plus `requires_grad`. That makes it a true postcondition: it structurally cannot refuse a build the materializer could have repaired.

That is the strongest safety argument in the PR and it is stronger than what your docstring claims for itself. Worth writing down if you touch this again.

## One LOW follow-up, non-blocking

Dropping `"lora_" in name` from the filter survives the whole suite — nothing pins the restriction your docstring calls deliberate. It is inert today (`build_streamed_model` freezes every base parameter before `get_peft_model`, so no trainable non-`lora_` meta parameter exists) and the risk direction is *spurious refusal*, the opposite of this issue. Still worth a line. This control passes on your code and fails on the mutation:

```python
def test_trainable_meta_non_lora_param_is_out_of_scope():
    class _M(nn.Module):
        def __init__(self):
            super().__init__()
            self.classifier_weight = nn.Parameter(
                torch.empty(2, 2, device="meta"), requires_grad=True)
    assert_trainable_adapters_materialized(_M())
```

## Verification

5 passed (yours) · 259 passed / 2 skipped (`test_v07200` + `test_v07202`, **both skips POSIX-symlink, not CUDA** — so the real streamed build runs here and the guard stays quiet on it) · 222 passed / 5 skipped · `test_cli_startup_is_light` 23/23 · ruff clean · no top-level torch added · merged clean with zero conflicts.

I also confirmed the behaviour directly rather than through your tests: it fires on a genuinely meta adapter naming both parameters, and stays quiet both when peft hands over real tensors (`materialize` returns 0) **and** after a 0.18.1-style materialisation (returns 2), with the frozen base still on `meta` in both. Both directions, so no peft 0.19 user gets a spurious failure.

@lesterppo — you have the peft 0.19.1 environment neither of us does. If you get another Colab session, a run with this merged would confirm the guard stays quiet on the stack where the original behaviour actually appears. The measurement gap from the issue is still open — `[(n, p.is_meta) for n, p in model.named_parameters() if p.requires_grad][:5]` right after `get_peft_model` — but the guard is correct either way and does not depend on it.

PR #436, opened 2026-08-17
fix: disambiguate Windows STILL_ACTIVE (259) exit code from a live process

### LINKED ISSUE #424: MCP liveness: a process that exits with code 259 reads as alive forever on Windows (STILL_ACTIVE ambiguity)
Split out of #407 / #408, measured during review.

Windows liveness goes through `OpenProcess` + `GetExitCodeProcess`, alive iff the exit code is `STILL_ACTIVE` (259). That constant is **also a legal process exit code**, so a process that exits *with* 259 is indistinguishable from a running one. Measured on Windows with a pinned handle:

```
child exited with returncode = 259
_process_is_alive(pid) -> True     <- false ALIVE
control: child exited 7 -> False   <- correct
```

POSIX has no equivalent ambiguity — the signal-0 path is exact.

## Why it is worth closing rather than documenting

The two consumers fail in opposite directions, and the second one is the failure this project explicitly called worse:

- **#401 (reconcile-on-read):** the run never reconciles and stays `running` forever — i.e. exactly the bug #401 fixed, for one exit code.
- **#402 (capacity gate):** the slot is never released, so the MCP server **refuses every subsequent execution forever**, with no error a user can act on. #402's own acceptance criteria name that as the failure mode of an over-eager fix and require it not happen.

Probability is low — the child is a `soup train` python process, and 259 is an unusual exit code — but it is the one genuine weakness of the chosen primitive, and the consequence is unbounded rather than transient.

## Fix path

Disambiguate with a wait: `WaitForSingleObject(handle, 0) == WAIT_OBJECT_0` means signalled, i.e. dead, regardless of the exit code. Treat that as authoritative and use `STILL_ACTIVE` only as the secondary signal.

## Also worth folding in while touching this

- **~50 lines are duplicated verbatim** between `experiment/tracker.py::_windows_process_is_alive` and `mcp_server/execution.py::_windows_pid_is_alive` — the executable bodies are byte-identical and the docstrings have already diverged (one documents the PID-reuse limitation, the other does not). `utils/adapter_fuse.py` exists as precedent for extracting exactly this kind of shared primitive so two copies cannot drift.
- **No regression guard names the cause.** Neither test file mentions `win32`, `sys.platform`, or the Windows helper. The branch is exercised end-to-end on Windows, which is why CI is green — but if someone reverts to `os.kill(pid, 0)`, the only thing that catches it is the entire Windows matrix dying with a `KeyboardInterrupt` at ~2348 passed after 14 minutes, which is the confusing signature that cost a round trip the first time.
- **`pid` reaches the primitive unchecked** (only `is not None`). A non-int raises `ctypes.ArgumentError` on Windows / `TypeError` on POSIX, caught by neither `except OSError` chain; in #401 that propagates out of `list_runs` and crashes `soup runs`. `pid=0` returns False on Windows but on POSIX `os.kill(0, 0)` targets the caller's own process group and reads permanently ALIVE. Only reachable through a corrupt DB, so LOW — one `isinstance(pid, int) and pid > 0` closes it.

## Acceptance criteria

- [ ] a process exiting with code 259 is reported DEAD, with a control that a genuinely running process is still ALIVE
- [ ] one shared liveness primitive, not two copies
- [ ] a test that names the platform split, so a revert to `os.kill(pid, 0)` fails by name rather than by a matrix-wide interpreter kill
- [ ] a non-int or non-positive pid cannot crash `soup runs`

blackcoderx:
## What does this PR do?
Fixes #424 — on Windows, `GetExitCodeProcess` reports `STILL_ACTIVE` (259) for a running process, but
259 is also a legal exit code, so a process that *exits with* 259 was indistinguishable from one still
running and read ALIVE forever. That could leave issue #401's reconcile-on-read permanently unable to
fire, and could wedge #402's one-active-execution capacity gate shut.

Fix: use `WaitForSingleObject(handle, 0)` as the authoritative signal — the process handle becomes
signalled the instant the process exits, independent of its exit code — falling back to the existing
`GetExitCodeProcess` read only if the wait itself fails. (The handle also needs `SYNCHRONIZE` added to
`PROCESS_QUERY_LIMITED_INFORMATION`, or `WaitForSingleObject` fails and silently falls through to the
old buggy path — caught by the new test before it could ship that way.)

Also folds in the issue's other findings while touching this code:
- Extracted the byte-identical liveness primitive that was duplicated in `experiment/tracker.py` and
  `mcp_server/execution.py` into one shared `utils/process_liveness.py` (their docstrings had already
  started to diverge), following the extraction precedent in `utils/adapter_fuse.py`.
- Added an `isinstance(pid, int) and pid > 0` guard so a corrupt/non-int pid can't crash `soup runs`
  (also closes the POSIX `pid=0` self-process-group anomaly noted in the issue).
- Added a Windows-only regression test asserting the win32 path never falls through to `os.kill`, so a
  revert to `os.kill(pid, 0)` fails that test by name instead of by the whole Windows CI matrix dying
  with a stray `KeyboardInterrupt`.

Verified locally on Windows: reproduced the exit-259 false-ALIVE read against the pre-fix code, confirmed
it flips after the fix. Full suite: 17674 passed, 393 skipped, 0 failed (80.75% coverage).

## Type of change
- [x] Bug fix

## Checklist
- [ ] Updated relevant docs — n/a, internal-only helper, no user-facing behavior/doc change
- [x] No breaking changes

Closes #424

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,19 @@ reproducing 70+ versions of notes.
   trainable `lora_*` parameter remains on `meta`, naming the stranded parameter
   instead of allowing a silent no-training run.
 
+- **Windows process liveness misread a process that exited with code 259 as
+  still running, forever (#424).** `GetExitCodeProcess` returns `STILL_ACTIVE`
+  (259) for a genuinely running process, but 259 is also a legal exit code, so
+  a child that exited *with* 259 was indistinguishable from one still running.
+  That silently defeated `ExperimentTracker`'s reconcile-on-read (#401) and
+  could wedge the MCP server's one-active-execution cap (#402) shut, refusing
+  every subsequent execution with no error an operator could act on. The check
+  now waits on the process handle with `WaitForSingleObject(handle, 0)`, which
+  is signalled the instant the process exits regardless of its exit code,
+  falling back to the exit-code read only if the wait itself fails. The two
+  byte-identical copies of this primitive in `experiment/tracker.py` and
+  `mcp_server/execution.py` are now one shared `utils/process_liveness.py`.
+
 - **`soup train --no-reexec` now prints the flags you actually typed (#372).**
   The advisory `accelerate launch` command is derived from the same argv the
   auto-reexec would have used, so `--fsdp` / `--deepspeed` / `--config` (and the
--- a/src/soup_cli/experiment/tracker.py
+++ b/src/soup_cli/experiment/tracker.py
@@ -13,12 +13,12 @@ import json
 import os
 import secrets
 import sqlite3
-import sys
 from datetime import datetime
 from pathlib import Path
 from typing import Optional
 
 from soup_cli.utils.constants import EXPERIMENTS_DB, SOUP_DIR
+from soup_cli.utils.process_liveness import process_is_alive as _process_is_alive
 
 # Run status values this module reconciles. A watcher that never unwound (its
 # daemon thread was killed when the MCP server exited) leaves the run at
@@ -27,85 +27,6 @@ from soup_cli.utils.constants import EXPERIMENTS_DB, SOUP_DIR
 # mistaken for success. See issue #401.
 _STATUS_RUNNING = "running"
 _STATUS_TERMINATED = "terminated"
-
-# Windows liveness constants (see _windows_process_is_alive). GetExitCodeProcess
-# returns STILL_ACTIVE for a running process; a failed OpenProcess distinguishes
-# a missing pid (ERROR_INVALID_PARAMETER) from one we merely cannot open
-# (ERROR_ACCESS_DENIED — it exists).
-_WIN_STILL_ACTIVE = 259
-_WIN_PROCESS_QUERY_LIMITED_INFORMATION = 0x1000
-_WIN_ERROR_ACCESS_DENIED = 5
-
-
-def _windows_process_is_alive(pid: int) -> bool:
-    """Liveness check for Windows, where ``os.kill(pid, 0)`` cannot be used.
-
-    On Windows ``CTRL_C_EVENT`` is ``0``, so ``os.kill(pid, 0)`` routes signal 0
-    to ``GenerateConsoleCtrlEvent`` — it sends a console Ctrl+C to the process
-    group and never checks existence (it raises KeyboardInterrupt in-process
-    instead). Query the process directly: open a limited-information handle and
-    read its exit code — a live process reads ``STILL_ACTIVE`` (259). A failed
-    open means the pid is gone (ERROR_INVALID_PARAMETER) unless it is only
-    ERROR_ACCESS_DENIED, which means the process exists but we lack rights.
-    """
-    import ctypes
-    from ctypes import wintypes
-
-    kernel32 = ctypes.WinDLL("kernel32", use_last_error=True)
-    kernel32.OpenProcess.argtypes = [wintypes.DWORD, wintypes.BOOL, wintypes.DWORD]
-    kernel32.OpenProcess.restype = wintypes.HANDLE
-    kernel32.GetExitCodeProcess.argtypes = [
-        wintypes.HANDLE,
-        ctypes.POINTER(wintypes.DWORD),
-    ]
-    kernel32.GetExitCodeProcess.restype = wintypes.BOOL
-    kernel32.CloseHandle.argtypes = [wintypes.HANDLE]
-
-    handle = kernel32.OpenProcess(
-        _WIN_PROCESS_QUERY_LIMITED_INFORMATION, False, pid
-    )
-    if not handle:
-        return ctypes.get_last_error() == _WIN_ERROR_ACCESS_DENIED
-    try:
-        exit_code = wintypes.DWORD()
-        if not kernel32.GetExitCodeProcess(handle, ctypes.byref(exit_code)):
-            return True  # opened but could not query — assume alive (conservative)
-        return exit_code.value == _WIN_STILL_ACTIVE
-    finally:
-        kernel32.CloseHandle(handle)
-
-
-def _process_is_alive(pid: int) -> bool:
-    """Return True if a process with this PID currently exists.
-
-    POSIX uses signal 0 — a genuine existence/permission check that delivers no
-    signal: a PID owned by another user raises PermissionError and counts as
-    alive; only a missing process (ProcessLookupError / ESRCH) is dead. Any
-    other OSError is treated as not-alive so a corrupt record reconciles instead
-    of sticking on 'running' forever.
-
-    Windows needs a different primitive entirely: ``os.kill(pid, 0)`` there
-    sends a console Ctrl+C (``CTRL_C_EVENT == 0``) rather than checking
-    existence, so liveness goes through OpenProcess + GetExitCodeProcess (see
-    ``_windows_process_is_alive``).
-
-    Known limitation — PID reuse: a dead run whose PID has since been recycled
-    by an unrelated process reads as alive forever, so reconciliation never
-    fires for it. That is inherent to PID-liveness (a recorded process
-    start-time comparison would be needed to close it) and is out of scope for
-    issue #401.
-    """
-    if sys.platform == "win32":
-        return _windows_process_is_alive(pid)
-    try:
-        os.kill(pid, 0)
-    except ProcessLookupError:
-        return False
-    except PermissionError:
-        return True
-    except OSError:
-        return False
-    return True
 
 _SCHEMA_SQL = """
 CREATE TABLE IF NOT EXISTS runs (
@@ -188,7 +109,6 @@ CREATE INDEX IF NOT EXISTS idx_forgetting_run_id ON forgetting_eval(run_id);
 
 def _get_db_path() -> Path:
     """Return path to experiments DB, creating parent dir if needed."""
-    import os
 
     # Allow override via env var (useful for tests and CI)
     env_path = os.environ.get("SOUP_DB_PATH")
--- a/src/soup_cli/mcp_server/execution.py
+++ b/src/soup_cli/mcp_server/execution.py
@@ -8,7 +8,6 @@ import os
 import secrets
 import stat
 import subprocess
-import sys
 import threading
 import time
 from dataclasses import dataclass
@@ -16,6 +15,7 @@ from pathlib import Path
 
 from soup_cli.experiment.tracker import ExperimentTracker, generate_run_id
 from soup_cli.utils.paths import atomic_write_text, enforce_under_cwd_and_no_symlink
+from soup_cli.utils.process_liveness import process_is_alive as _pid_is_alive
 
 TOKEN_TTL_SECONDS = 5 * 60
 DEFAULT_MAX_TREE_FILES = 10_000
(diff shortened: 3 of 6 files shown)


MakazhanAlpamys:
Welcome — and thank you for claiming #424 in the issue before opening this. That is the convention here and it saves duplicated work.

Your CI runs are unblocked (first-time contributors need manual approval), and the review is under way.

Three things so you know what I will be checking, since none of them are guessable from the diff:

**1. We are on Windows here, so the bug is reproducible rather than theoretical.** I will be constructing both halves — a live process, and a process that genuinely exits with code **259** — and confirming the second now reads as dead. That second case is the entire issue, so it is the one that decides the PR.

**2. `tests/test_issue402_restart_capacity.py` has guards I added this week, and they are load-bearing.** While repairing a semantic collision between #407 and #408 I found that deleting the capacity check's own `_pid_is_alive` guard passed **all 41 MCP tests**, because reconcile-on-read masked it. `test_capacity_check_rejects_a_dead_pid_without_help_from_reconcile` neutralises reconcile precisely so that guard is the only thing that can reject a dead pid, and it has a control beside it. I will verify both survive with their teeth intact. Not an accusation — you had no way to know why they were shaped that way, and touching that file is legitimate. Just telling you what I am looking at.

**3. Merging the two duplicated helpers into `utils/process_liveness.py` is the right instinct**, and I want to say so explicitly. Duplicated sources of truth are how this project has been bitten repeatedly — two copies of "what the user typed" in #372, two copies of "which modules did we train" in #392. What I will check is that it is genuinely *one* implementation now and not a third copy sitting alongside two survivors.

Two things I expect to raise, so you can pre-empt them if you like:

- **No `CHANGELOG.md` entry** in the file list. This is a user-facing bug fix, so it needs one under `[Unreleased]`.
- **The failure direction on an error path.** If `OpenProcess` fails — access denied on a pid owned by another user, or an invalid pid — does the code fall back to "alive" or "dead"? Falling back to alive wedges the one-active-execution cap forever; falling back to dead lets it double-book. Either can be right, but it should be a decision with a test and a comment, not a default.

House rules, briefly: I do not trust a PR description — I run the tests myself and mutate the main line of the fix to check a test actually fails. And I do not merge until CI is green on all cells.

MakazhanAlpamys (changes requested):
**Excellent first contribution — genuinely one of the cleanest bug-fix PRs this repo has had.** One small thing before I merge, at the bottom.

I reproduced the whole thing on Windows 11 in an isolated worktree, running your algorithm and `main`'s side by side against the same children:

| case | expect | `main` | this PR |
|---|---|---|---|
| child exits with code **259** | DEAD | **`True` ← the bug** | **`False`** |
| child exits with code 7 | DEAD | `False` | `False` |
| genuinely running child | ALIVE | `True` | **`True`** |

The live-process control still reading `True` is the part that matters — it shows this is not the degenerate "always dead" fix, which is the obvious way to make case 1 pass and break everything else.

Three things worth calling out specifically, because they are what people usually get wrong here.

**You kept `SYNCHRONIZE` and said why.** I mutated it out: it fails `test_exit_code_259_reads_as_dead` and silently falls through to the old buggy path — exactly as your description predicts. That is a failure you only find by testing the real API rather than trusting the documentation, and you wrote it down at the call site instead of leaving it as folklore for the next person.

**The refactor did not cost a single assertion.** `gh pr diff 436 | grep -E "^-.*(def test_|assert )"` returns zero lines. I checked this specifically: `tests/test_issue402_restart_capacity.py` carries two guards I added last week *because they were missing*, and your only edit to that file is one word inside a docstring tracking the rename. I re-ran the mutation that motivated them — deleting the liveness guard in `_live_persisted_run` still fails **exactly one test in the entire run**, the one that exists to catch it. Aliasing the import to preserve each module's original local name is what kept the blast radius that small, and that was a deliberate choice on your part.

**`test_win32_never_falls_back_to_os_kill` does what its name says.** I reverted the win32 branch to the `os.kill` path and it fails **by name** — rather than by the Windows matrix dying with a stray `KeyboardInterrupt` fourteen minutes in. That confusing signature has cost this project a round trip before.

**8 mutations, 7 killed.** The survivor is the POSIX branch, unreachable on Windows; I confirmed none of the four non-#424 suites are `skipif`-gated, so five tests using real subprocess pids cover it on Linux CI. I also probed oversized pids (`2**32`, `2**63`) past your `pid > 0` guard — ctypes masks to DWORD and `OpenProcess` fails safely, no `ArgumentError`, so the `soup runs` crash path in the issue is genuinely closed.

Merging the two duplicated helpers into one module is the right instinct and I want to say so plainly: duplicated sources of truth are how this project has been bitten repeatedly. I verified it is genuinely **one** implementation — both former copies deleted, two import sites, and `test_issue424_windows_liveness.py` asserts identity so a future third copy would fail.

`ruff check` clean · `test_cli_startup_is_light` 23/23 with `ctypes` correctly lazy inside the function · 72 passed across the five liveness suites, 346 across the wider dependent slice · based on current `main`, no conflicts.

## The one thing before merge

**A `CHANGELOG.md` entry**, under `[Unreleased]` → a new `### Fixed` section (the block currently has only `### Added`).

I am asking rather than adding it myself because I have asked every other contributor for one this week and it would be odd to make an exception — and because the entry is part of your contribution record, not mine. It is user-facing: a wedged capacity gate refuses every subsequent execution with no error anyone can act on.

Not your fault that it is missing — the PR template says "docs" and reading that as `docs/` is entirely reasonable. Your call that no `docs/` page was needed is **correct**; I checked, and no page describes this internal primitive.

Two error paths remain untested — `ERROR_ACCESS_DENIED` on `OpenProcess`, and the `WAIT_FAILED` → `GetExitCodeProcess` fallback. Both were untested before this PR too, so not a regression, and the second is now effectively unreachable given you request `SYNCHRONIZE`.

On `ruff format` wanting to join your `@pytest.mark.parametrize` onto one line: **ignore it.** `ruff-format` is disabled on purpose in `.pre-commit-config.yaml` — leaving it as-is is the repo-consistent choice.

Add the CHANGELOG line and I will merge as soon as the Windows cells finish.

blackcoderx:
Added the CHANGELOG entry under `[Unreleased]` → `### Fixed`. Thanks for the thorough review — and for catching that `SYNCHRONIZE` would've silently regressed without it.

MakazhanAlpamys:
CHANGELOG entry received, thank you — and apologies, that has now put you in conflict through no fault of yours.

I merged #435 a few minutes after asking you for the entry, and it also added a bullet under `[Unreleased]`. So:

```
CONFLICT (content): Merge conflict in CHANGELOG.md
```

**That is the only conflict** — I checked the scope rather than guessing. Your source files and both test files merge cleanly. Rebase onto current `main`, keep both bullets, and I will merge as soon as the matrix is green.

For what it is worth, this is the single most common friction point in this repo right now: every contributor inserts at the top of the same section, so whoever pushes second collides. Nothing you did wrong.

The review stands unchanged — the fix, the deduplication and the pid guard are all verified, and my two `test_issue402` guards survive with their teeth intact.

MakazhanAlpamys (approved):
CHANGELOG entry confirmed and the rebase cost nothing — approving.

Re-checked at the rebased head:

- **Entry is correct**: under `### Fixed` inside `[Unreleased]`, user-facing wording, and it describes what actually shipped — a Windows-only liveness misread that defeats reconcile-on-read (#401) and can wedge the MCP one-active-execution cap (#402) shut with no error an operator could act on.
- **No test loss in the rebase**: `gh pr diff 436 | grep -E "^-.*(def test_|assert )"` is empty, and so is the same scan against the true merge-base. A rebase is exactly where an accidental loss would hide, which is why I looked twice.
- **My two guards survived byte-identical** — verified by AST comparison rather than by eye. The sole delta in that file is one word inside a docstring (`_pid_is_alive` → `process_is_alive`) tracking your rename, which is correct maintenance; leaving the old name would have been a stale reference.
- **Decisive mutation re-run**: deleting the liveness guard in `_live_persisted_run` fails exactly `test_capacity_check_rejects_a_dead_pid_without_help_from_reconcile`. Neutering the tracker guard fails 4, including both #401 controls.
- **259 re-confirmed on real Windows with real children**: exit-259 → DEAD, running → ALIVE, killed → DEAD, invalid pids → False. And the RED control on pre-fix `main`: exit-259 reads **ALIVE** while exit-7 correctly reads DEAD — so the old helper was not broken generally, it was broken *specifically* on 259, which is the claim the issue makes.

38 passed across the four liveness suites, `test_cli_startup_is_light` 23/23, ruff clean, no dangling references to the two removed duplicates.

Merging now.

MakazhanAlpamys:
Merged as [`53f29c0`](https://github.com/MakazhanAlpamys/Soup/commit/53f29c0) — 12/12 green. You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) as of [`9660fa6`](https://github.com/MakazhanAlpamys/Soup/commit/9660fa6).

**One of the cleanest bug-fix PRs this repo has had, and a very strong first contribution.**

I reproduced the whole thing on Windows with real child processes, running your algorithm and `main`'s side by side:

| case | `main` | this PR |
|---|---|---|
| child exits with code **259** | **ALIVE ← the bug** | **DEAD** |
| child exits with code 7 | DEAD | DEAD |
| genuinely running child | ALIVE | **ALIVE** |

The second and third rows are what make this convincing rather than merely passing: the old helper was **not** broken generally — it was broken *specifically* on 259, exactly as the issue claims — and your fix does not achieve row 1 by declaring everything dead.

**The credit line names the deduplication, not the disambiguation.** Fixing only the headline would have left two ~50-line copies of the liveness check drifting; you replaced them with one shared module and added a test asserting identity, so a future third copy fails rather than quietly appearing. This project has been bitten by duplicated sources of truth repeatedly — two copies of "what the user typed" in #372, two copies of "which modules did we train" in #392.

Three things I checked specifically and want you to know were checked:

**Your rebase cost nothing.** `gh pr diff | grep -E "^-.*(def test_|assert )"` is empty, and so is the same scan against the true merge-base. A rebase is where an accidental test loss would hide, so I looked twice.

**Two guards in `test_issue402_restart_capacity.py` survived byte-identical** — verified by AST comparison, not by eye. Those were added last week because deleting the capacity check's own liveness guard passed all 41 MCP tests at the time. Your only edit to that file is one word inside a docstring tracking your rename, which is correct maintenance; leaving the old name would have been a stale reference. Re-running that mutation on your branch still fails exactly that one test.

**`SYNCHRONIZE` matters and you documented why.** I mutated it out: it fails `test_exit_code_259_reads_as_dead` and silently falls through to the old buggy path — precisely as your description predicted. That is the kind of thing you only learn by testing the real API rather than trusting the docs, and you wrote it down at the call site instead of leaving it as folklore.

8 mutations, 7 killed. The survivor is the POSIX branch, unreachable on Windows and covered by five tests using real subprocess pids on Linux. I also probed oversized pids (`2**32`, `2**63`) past your `pid > 0` guard — ctypes masks to DWORD and `OpenProcess` fails safely, so the `soup runs` crash path named in the issue is genuinely closed.

`ruff` clean · `test_cli_startup_is_light` 23/23 with `ctypes` correctly lazy inside the function · 38 passed across the four liveness suites · no dangling references to the two removed duplicates.

Sorry about the CHANGELOG round trip — I merged another PR minutes after asking you for the entry, which put you in conflict. That was my sequencing, not anything you did.

Two error paths remain untested — `ERROR_ACCESS_DENIED` on `OpenProcess`, and the `WAIT_FAILED` → `GetExitCodeProcess` fallback. Both were untested before this PR too, and the second is now effectively unreachable given you request `SYNCHRONIZE`. Noting them rather than asking.

PR #437, opened 2026-08-17
test(stream): pin non-LoRA meta guard scope

Faisal01011:
## Summary

- add the missing mutation control for the adapter materialization postcondition
- prove a trainable non-`lora_*` parameter on `meta` remains outside the guard's deliberate scope
- prevent a future broadening of the predicate from causing spurious streamed-build refusals

## Context

This is the LOW, non-blocking follow-up identified in the merge review for #435. Dropping the `"lora_" in name` predicate previously survived the suite; the new control fails under that mutation and passes with the intended implementation.

## Impact

Test-only. Runtime behavior is unchanged.

## Verification

- `python -m pytest -q --no-cov tests/test_issue433_meta_adapter_guard.py tests/test_v07200.py tests/test_v07202.py` — 262 passed, 5 skipped
- `ruff check src/soup_cli/ tests/` — clean

### DIFF
--- a/tests/test_issue433_meta_adapter_guard.py
+++ b/tests/test_issue433_meta_adapter_guard.py
@@ -72,6 +72,23 @@ class TestAdapterMaterializationPostcondition:
         model = _adapter_model(adapter_device="meta", adapter_trainable=False)
         assert_trainable_adapters_materialized(model)
 
+    def test_trainable_meta_non_lora_parameter_is_out_of_scope(self):
+        import torch
+        import torch.nn as nn
+
+        from soup_cli.utils.layer_stream_runtime import (
+            assert_trainable_adapters_materialized,
+        )
+
+        class _Model(nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.classifier_weight = nn.Parameter(
+                    torch.empty(2, 2, device="meta"), requires_grad=True
+                )
+
+        assert_trainable_adapters_materialized(_Model())
+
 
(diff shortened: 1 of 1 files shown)


MakazhanAlpamys:
Merged as [`0bfd5c7`](https://github.com/MakazhanAlpamys/Soup/commit/0bfd5c7) — 12/12 green.

I verified this one myself rather than delegating it, since it is small and the mutation is the exact one I named in the #435 review:

- removing `"lora_" in name` from the filter → **fails** `test_trainable_meta_non_lora_parameter_is_out_of_scope`
- **control**: deleting the guard call from `build_streamed_model` entirely → fails a *different* test

So the file is genuinely exercised and the new test is not riding along on someone else's coverage. Diff is test-only, +17 lines, zero source change, no deleted assertions.

The thing worth noting is that you took this unprompted. It was explicitly filed as LOW and non-blocking — inert on today's code path, and the risk direction is *spurious refusal* rather than silent no-training, i.e. the opposite of what #433 was about. Follow-ups written that way are usually the ones nobody returns to. You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) for it, fourth merge.

No CHANGELOG needed here and I want to be explicit about why, so the absence does not read as an oversight: `.claude/CLAUDE.md` treats commits touching only `tests/`, `docs/` or `.github/` as not shipping to PyPI, so they do not require an entry. Your judgement on that was right.

Faisal01011:
happy to help..

PR #439, opened 2026-08-17
fix(data): handle BatchEncoding loss-mask outputs safely

### LINKED ISSUE #430: loss_mask._tokenize_only returns key strings as token ids: BatchEncoding is not a dict, so assistant-only masking is built from garbage
## The bug

`data/loss_mask.py::_tokenize_only` ends with:

```python
    if isinstance(out, dict):
        return list(out.get("input_ids", []))
    return list(out)
```

`tokenizer.apply_chat_template(..., tokenize=True)` can return a **`BatchEncoding`**, and `BatchEncoding` is **not** a `dict` — it subclasses `UserDict`. So the guard misses, and `list(out)` iterates the mapping's **keys**.

Reproduced on the installed stack:

```
isinstance(BatchEncoding({"input_ids": [1,2,3]}), dict)     -> False
isinstance(BatchEncoding({"input_ids": [1,2,3]}), Mapping)  -> True
list(BatchEncoding({"input_ids": [1,2,3]}))                 -> ['input_ids']
```

So the function returns `['input_ids']` — a list of **key strings** — where the caller expects token ids.

## Why it matters

This is the assistant-only loss-masking path. It does not raise: the wrong value flows onward and the run trains, saves and reports a normal loss curve while the label mask is built from garbage. It is not Qwen3.5-specific and not architecture-specific — it depends only on which object that tokenizer's `apply_chat_template` hands back.

This is the fourth time this project has met the same shape: `.inner.` adapter keys (v0.72.0), the full-fine-tune MLX adapter (#362), `target_modules: auto` (#392). The recorded lesson each time is that a healthy loss curve structurally cannot see it — only the artifact or an explicit assertion can.

## Credit

Found by **[@Shutaru](https://github.com/Shutaru)** while working on #426, where it surfaced as the "later dataset/collator issue" their layer-streaming work ran into. I reproduced it independently before filing.

Splitting it out deliberately: it is a data-path correctness change affecting **every** user of assistant-only masking, and it should be reviewed on its own terms rather than as a passenger in a layer-streaming PR. #426 already carries a working fix, so this issue is the place to land it with its own tests and CHANGELOG entry.

## Acceptance criteria

- [ ] `_tokenize_only` accepts any `Mapping` (or anything exposing `.get`), not only `dict`.
- [ ] It **raises** rather than returning a wrong-typed list when it cannot obtain token ids — the current failure is silent, and that is the actual defect.
- [ ] A regression test whose fixture returns a real `BatchEncoding`. A plain `dict` fixture cannot catch this — that is precisely why it survived.
- [ ] A test asserting the returned ids are `int`, so a future refactor cannot reintroduce a list of strings.
- [ ] Decide and document two behaviour changes that come with the fix: an all-zero assistant mask, and a tokenizer returning a **tensor** of ids (currently yields `[tensor(1), ...]`).
- [ ] Audit for the same `isinstance(x, dict)` assumption elsewhere against tokenizer output — a one-off fix leaves the class open.

Shutaru:
Closes #430.

## What changed

- Read tokenizer mappings, including real `transformers.BatchEncoding` objects, through `input_ids` in both assistant-mask paths.
- Normalise tensor-like ids and masks to Python integers. Missing or non-integer token ids now raise instead of reaching the collator.
- Reject an all-zero assistant mask when assistant messages exist, then use the existing incremental-rendering fallback so the row does not silently become all `-100`.
- Apply the same mapping handling to `data doctor` generation-marker checks, mask previews, and truncation-risk measurements.
- Document the two behaviour decisions in `loss_mask.py` and add an Unreleased changelog entry.

## Root cause and impact

`BatchEncoding` implements `Mapping` through `UserDict`, but it is not a `dict`. The previous fallback called `list(out)`, which returned mapping keys such as `['input_ids']` instead of token ids. Those strings could travel as far as the training collator before failing, or corrupt assistant-only label construction.

Some chat templates also accept the assistant-mask arguments without declaring `{% generation %}` markers. They return a mask containing only zeros. When assistant messages are present, trusting that result creates a no-op training row with every label set to `-100`; this change falls back rather than accepting it.

## Validation

- `219 passed, 1 skipped` across `test_assistant_mask.py`, `test_v07127.py`, and `test_formats.py`
- `ruff check src/soup_cli tests` passes
- Mutation checks confirm the tests fail when any of these changes is reverted: preferred-path `Mapping` support, the all-zero-mask guard, and each of the three `data doctor` consumers
- The full local suite reached `721 passed` before an installed FastAPI/Starlette incompatibility failed `test_tool_python_endpoint_requires_bearer_when_token_set`. The same test fails with the same traceback on an isolated checkout of current `origin/main`; the PR CI will provide the clean-environment result.

## Submission context

This branch was already being prepared locally when #438 opened. I am submitting it after the maintainer invited both implementations to be reviewed on their merits. I'm not asking for precedence over #438, and the current #426 no longer contains these changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,15 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **Assistant-only masking no longer mistakes `BatchEncoding` keys for token
+  ids (#430).** Tokenizer mapping outputs are read through `input_ids`, and
+  tensor-like ids are normalised to Python integers before they reach the
+  collator; missing or non-integer ids now fail loudly instead of building a
+  garbage label mask. A template that returns an all-zero assistant mask while
+  assistant messages exist now falls back to incremental rendering, avoiding a
+  silent all-`-100` no-op training run. The same mapping assumption was removed
+  from the data doctor.
+
 - **Layer streaming now verifies that every trainable LoRA parameter has real
   storage after PEFT attaches the adapter (#433).** PEFT 0.18 creates streamed
   adapters on `meta` for Soup to materialise, while PEFT 0.19 may create them as
--- a/src/soup_cli/data/loss_mask.py
+++ b/src/soup_cli/data/loss_mask.py
@@ -12,7 +12,10 @@ Two strategies:
 
 1. **Preferred**: ``tokenizer.apply_chat_template(..., return_assistant_tokens_mask=True,
    return_dict=True)``. Available on HF templates that declare ``{% generation %}``
-   markers. Honest, exact, no heuristic.
+   markers. Mapping outputs such as ``BatchEncoding`` are read through
+   ``input_ids`` and tensor-like ids/masks are normalised to Python integers.
+   If assistant messages are present but the returned mask is all zero, this
+   path is rejected so the fallback can avoid a silent all-``-100`` training row.
 
 2. **Fallback**: Render ``messages[:i]`` vs ``messages[:i+1]`` for each turn and
    take the token delta. The delta is the new turn's tokens (prefix + content +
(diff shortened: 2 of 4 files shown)


MakazhanAlpamys (approved):
Merging this. I ran the **identical battery** against this and #438 in isolated worktrees, and reported the same measurements for each — that was the commitment I made on #430, and I want the reasoning visible rather than asserted.

Measured here:

- **All six acceptance criteria met.** Real `BatchEncoding` fixture, `type(x) is int` against a real `torch.tensor`, raises on both failure modes, all three `data_doctor` sites fixed **and** pinned.
- **9 of 9 applicable mutations killed, zero survivors.** The one non-kill — dropping `int()` inside `_coerce_int_list` — is an *equivalent mutant*, since `index()` already returns an `int`. No test can or should catch it, and I am not counting it as a gap.
- **The all-zero-mask guard is the right call**, and I am glad you made it rather than preserving the status quo. Measured on a template that accepts the mask kwargs without declaring `{% generation %}`: **0 trained tokens before, 2 after**, with the no-assistant case correctly left at all-`-100`. Documented in four places with a control test for the legitimate case — that control is what makes the guard trustworthy rather than merely present.
- **Using a real `torch.tensor` in `test_tensor_ids_are_normalised_to_python_ints` is what made criterion 4 load-bearing**, and it is the single detail that decided the comparison. A `.tolist()` stub passes whether or not the conversion exists; a real tensor does not.

**Two things #438 did better, and I want them on record rather than buried.** Its `isinstance(out, Mapping) or hasattr(out, "get")` satisfies criterion 1's `.get` clause *literally*, where `_coerce_token_ids` accepts registered `Mapping`s only. Measured: that covers every real transformers container — `BatchEncoding` and `BatchFeature` are both registered — and your path raises loudly rather than silently, so nothing is at risk. But the wording was mine and #438 matched it.

And #438 kept `data_doctor` decoupled, where this imports the private `_coerce_token_ids` across a module boundary. I think the shared helper is the right trade here — it is what let both `_build_row_labels` and `check_truncation_risk` be fixed once instead of twice, which is the convention this project keeps returning to — but a rename to a public `coerce_token_ids` would settle the smell.

Two non-blocking follow-ups I will track: that rename, and adopting #438's `.get` widening with a test to pin it.

Thank you for finding this in the first place, for splitting it cleanly out of #426 when I asked, and for how you handled the overlap with #438 — you could reasonably have been annoyed and instead you just wrote the better version.

PR #440, opened 2026-08-17
fix: soup data mix --optimize writes a recipe soup train can load

### LINKED ISSUE #330: soup data mix --optimize writes a recipe soup train cannot load
`render_mix_recipe_yaml` (`src/soup_cli/utils/data_mix.py:632`) emits `data.train` as a YAML list:

```yaml
data:
  interleave:
    strategy: probs
    probs:
      - 0.700000
      - 0.300000
  train:
    - "a.jsonl"
    - "b.jsonl"
```

`DataConfig.train` is typed `str` (`src/soup_cli/config/schema.py:221`), so `load_config_from_string` rejects it:

```
data -> train: Input should be a valid string
```

Verified by constructing the config directly, against a complete config so that
`train` is the only error reported:

```python

load_config_from_string("""
base: HuggingFaceTB/SmolLM2-135M
task: sft
data:
  interleave:
    strategy: probs
    probs: [0.7, 0.3]
  train:
    - "a.jsonl"
    - "b.jsonl"
""")
# -> data -> train: Input should be a valid string
```

So the tool produces a recipe its own parser cannot read. `data.interleave` is
validated and real; the multi-path input it exists to consume is not wired
through the schema.

Two possible fixes:

- widen `DataConfig.train` to `str | list[str]`
- or have `render_mix_recipe_yaml` emit a loadable form

Whichever is chosen, **a test that loads the output of `render_mix_recipe_yaml`
through `load_config_from_string` would have caught this and should ship with
the fix.** That last point matters more than the bug itself: without such a
test this recurs, exactly as it recurred with the example configs in
`examples/configs/` (fixed in 66be481, guarded afterwards by
`tests/test_examples_configs.py`).

Same class as those: a producer emitting a config the consumer rejects, with no
test crossing the seam between them.

blackcoderx:
## What does this PR do?
Fixes #330 — `render_mix_recipe_yaml` emitted `data.train` as a YAML list, but `DataConfig.train` is
typed `str`, so the recipe `soup data mix --optimize` writes fails to load
(`data -> train: Input should be a valid string`). `data.train` now renders as the single
highest-weighted dataset from the search, with the full ranked breakdown kept as a comment so the
human-review value isn't lost. `data.interleave` is untouched — it already round-trips fine.

Added the round-trip test you asked for: render → splice into a config → `load_config_from_string` →
loads.

**Two things found while investigating, out of scope for this PR, flagging rather than fixing:**
1. `data.interleave` has no consumer anywhere in the real training pipeline — `load_dataset()` only ever
   handles a single-path `data.train`. `parse_interleave`/`InterleaveSpec` exist and are exported but are
   never called. So even a config that loads cleanly with `interleave` set doesn't actually train on a
   weighted mixture today.
2. Because of (1), `soup data mix --live`'s overlay YAML (`mix_proxy.py::_render_overlay_yaml`) hits the
   exact same list-shaped-`train` bug — every live candidate would fail at config-load before training
   starts. Verified directly (built the overlay it produces, ran it through `load_config_from_string`,
   same error). Invisible in `test_v0535.py` because `subprocess.run` is mocked in every `--live` test, so
   the real config-load path is never exercised. Did not touch `mix_proxy.py` here since a real fix needs
   either per-candidate dataset merging or the full interleave-consumption feature — happy to open a
   follow-up issue if useful.

## Type of change
- [x] Bug fix

## Checklist
- [x] `pytest tests/ -v` passes (17716 passed, 0 failed, 80.76% coverage)
- [ ] Updated relevant docs — n/a, internal helper output format, no public API/doc change
- [x] No breaking changes

Closes #330

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,15 @@ reproducing 70+ versions of notes.
   silent all-`-100` no-op training run. The same mapping assumption was removed
   from the data doctor.
 
+- **`soup data mix --optimize` wrote a recipe `soup train` could not load
+  (#330).** `render_mix_recipe_yaml` emitted `data.train` as a YAML list of
+  every searched dataset, but `DataConfig.train` is typed `str`, so the recipe
+  failed to load with `data -> train: Input should be a valid string`.
+  `data.train` now renders as the single highest-weighted dataset from the
+  search; the full ranked weight/path breakdown is kept as a comment above it
+  so no information is lost, and the comment explains that `data.interleave`
+  is not yet consumed by training.
+
 - **Layer streaming now verifies that every trainable LoRA parameter has real
   storage after PEFT attaches the adapter (#433).** PEFT 0.18 creates streamed
   adapters on `meta` for Soup to materialise, while PEFT 0.19 may create them as
--- a/docs/data.md
+++ b/docs/data.md
@@ -734,7 +734,7 @@ soup data mix --optimize --budget 1h \
     --num-probes 8 --output mix_recipe.yaml
 ```
 
-Writes a YAML recipe with a `data.interleave` block you can splice into your `soup.yaml`. `--budget` accepts `60s` / `5m` / `1h` / `24h`. Per-candidate proxy failures are isolated (DEBUG-logged, sentinel high loss recorded) so a single OOM combo does not abort the whole search; `partial=True` is surfaced in the report when the budget cap trips mid-loop.
+Writes a YAML recipe you can splice into your `soup.yaml`: `data.train` is the single highest-weighted dataset from the search (the full ranked weight/path breakdown is kept as a comment), and `data.interleave` carries the searched mixture weights but is not yet consumed by training — see the in-file comment, and #330/#443. `--budget` accepts `60s` / `5m` / `1h` / `24h`. Per-candidate proxy failures are isolated (DEBUG-logged, sentinel high loss recorded) so a single OOM combo does not abort the whole search; `partial=True` is surfaced in the report when the budget cap trips mid-loop.
 
 Re-apply a previously written recipe:
 
--- a/src/soup_cli/commands/data_mix.py
+++ b/src/soup_cli/commands/data_mix.py
@@ -14,6 +14,7 @@ optimiser surface, and the recipe writer end-to-end without GPUs.
 
 from __future__ import annotations
 
+import json
 import math
 from typing import Optional, Tuple
 
@@ -136,8 +137,10 @@ def mix(
         for p in probs:
             console.print(f"      - {float(p):.6f}")
         if isinstance(train_value, str):
-            # #330 — current recipes: data.train is a single string.
-            console.print(f"  train: {escape(train_value)}")
+            # #330 — current recipes: data.train is a single string. Quote it
+            # the same way render_mix_recipe_yaml does — an unquoted path
+            # like "odd: name.jsonl" is not valid YAML when pasted back.
+            console.print(f"  train: {escape(json.dumps(train_value))}")
         else:
             # Pre-#330 recipes on disk: data.train was a YAML list.
             console.print("  train:")
--- a/src/soup_cli/commands/data_mix.py
+++ b/src/soup_cli/commands/data_mix.py
@@ -124,9 +124,7 @@ def mix(
             )
         )
         # Round-trip via the renderer so the user sees the canonical shape.
-        train_paths = data_block.get("train", []) if hasattr(
-            data_block, "get"
-        ) else []
+        train_value = data_block.get("train") if hasattr(data_block, "get") else None
         interleave = (
             data_block.get("interleave", {}) if hasattr(data_block, "get") else {}
         )
@@ -137,9 +135,14 @@ def mix(
         console.print("    probs:")
         for p in probs:
             console.print(f"      - {float(p):.6f}")
-        console.print("  train:")
-        for path in train_paths:
-            console.print(f"    - {escape(str(path))}")
+        if isinstance(train_value, str):
+            # #330 — current recipes: data.train is a single string.
+            console.print(f"  train: {escape(train_value)}")
+        else:
+            # Pre-#330 recipes on disk: data.train was a YAML list.
+            console.print("  train:")
+            for path in train_value or []:
+                console.print(f"    - {escape(str(path))}")
         return
 
     if not datasets:
--- a/src/soup_cli/utils/data_mix.py
+++ b/src/soup_cli/utils/data_mix.py
@@ -598,6 +598,16 @@ def render_mix_recipe_yaml(report: MixOptimizationReport) -> str:
     ``data:``. Defends against YAML key injection by rejecting newlines and
     null bytes in dataset paths (mirrors v0.46.0 Part A
     ``render_recipe_yaml``).
+
+    ``data.train`` renders as a single string — the highest-weighted dataset
+    from the search — because ``DataConfig.train`` is typed ``str`` and
+    ``soup train`` has no reader for a weighted multi-dataset mixture
+    (``data.interleave`` is validated at config-load time but never consumed
+    by ``load_dataset()``). Emitting the full dataset list there produced a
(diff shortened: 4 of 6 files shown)


MakazhanAlpamys:
Reviewed. **The fix is correct and I want to be specific about what I checked, because the parts I could have taken on trust are the parts I ran.**

Everything below is from an isolated worktree at `913277f` with `PYTHONPATH` pointed at *that* tree — this machine's editable install points at the main checkout, so a worktree test run without it silently exercises `main` and every mutation "survives". Both module paths confirmed under the worktree before anything else.

## What holds up

**The bug and the fix, reproduced independently of your tests.** Checked out the two `src/` files at `913277f^` and rendered a recipe by hand:

```
PRE-FIX:  train:
            - "a.jsonl"
            - "b.jsonl"
          LOAD: FAIL -> ValueError  data -> train: Input should be a valid string
FIXED:    train: "a.jsonl"
          LOAD: OK   data.train = 'a.jsonl'
```

Your test file is genuinely RED on pre-fix code (`5 failed, 92 passed`), not merely green on the new one.

**Six mutations, none survived.** I broke the fix in six ways and every one was caught:

| mutation | result |
|---|---|
| restore the 3-line list emission | KILLED (4 tests) |
| `best_idx = max(...)` -> `min(...)` | KILLED (4 tests) |
| emit `train: ""` | KILLED (4 tests) |
| ranked-comment loop body -> `pass` | KILLED by `test_full_ranked_breakdown_still_in_comments` |
| `isinstance(train_value, str)` -> `False` in the `--apply` echo | KILLED by `test_apply_cli_prints_new_string_shape_recipe` |
| back-compat list branch prints header only | KILLED by `test_apply_handles_pre_fix_list_shaped_recipe` |

The fifth is the instructive one: the mutant iterates the *string* character by character and prints `- C`, `- :`, … which is exactly the shape that guard exists to prevent, and the test names it. I re-ran the `max`->`min` mutation myself rather than only reading the result — 4 failures, tree restored clean.

**Diff matches the description.** No test deleted or weakened; every removed line in the patch is in `src/`, and `test_v0480_part_b.py` only gains an import and an assertion. I check this on every PR now.

**YAML safety under hostile paths.** `we"ird#name.jsonl`, `odd: name.jsonl`, a literal tab — all render, all `yaml.safe_load` clean, all load through the schema. Newline and NUL are rejected by the pre-existing validator that runs *before* the comment block is appended, so the ranked breakdown cannot be broken out of.

`152 passed, 5 skipped` across `test_issue330_*`, `test_v0480_part_b.py`, `test_v0535.py`; `45 passed` on `test_v05310.py`; ruff clean.

## Your two out-of-scope findings — both verified, both getting issues

I checked these rather than taking them as asides, because if they are true they matter more than the PR:

1. **`data.interleave` has no consumer** — `parse_interleave` appears only at its own definition, its `__all__` entry, and a comment in `schema.py`. Zero hits under `data/` or `trainer/`. True.
2. **`soup data mix --live` hits the identical list-shaped-`train` bug** in `mix_proxy._render_overlay_yaml` — I built the real overlay and loaded it, same `ValueError`. Every live candidate fails at config-load before training starts. True, and invisible to `test_v0535.py` because `subprocess.run` is mocked in every `--live` test, exactly as you said.

Filing both as issues with the finding credited to you. Finding a second instance of your own bug in a path your PR does not touch, and flagging it instead of quietly widening the diff, is the right call twice over.

## What I need before merging

**1. CHANGELOG (blocker, and the only one).** The PR touches no `CHANGELOG.md`. This changes the content of a file users read and splice, so it needs a `### Fixed` bullet under `[Unreleased]` naming #330 and stating the new `data.train` shape. House rule, not a preference.

**2. `docs/data.md:737` (nit, would like it).** It still tells the reader the recipe has "a `data.interleave` block you can splice into your `soup.yaml`" — which is now the inert half of the file. One sentence mirroring your in-file comment.

**3. `render_mix_recipe_yaml` guard (nit, your call).** With an empty `best_weights` it raises `ValueError: max() arg is an empty sequence`, and with more weights than datasets an `IndexError`. Not reachable through the CLI — `build_optimization_plan` requires two datasets and `run_mix_optimizer` always sizes weights to `len(plan.datasets)` — but the function is in `__all__` and takes a caller-built report, and it already raises purpose-written errors for every other malformed input. Two lines.

**4. The `--apply` echo does not quote (nit, pre-existing in kind, on a line you touched).** The renderer quotes via `json.dumps`; the echo does not:

```
$ soup data mix --apply rec.yaml     # file contains  train: "odd: name.jsonl"
  train: odd: name.jsonl
$ yaml.safe_load(that block)  ->  ScannerError: mapping values are not allowed here
```

The panel says "Splice the following block into your soup.yaml", so for such a path the copy-pasted block does not parse. Narrow, and the old list branch had the same gap — but the fix line is right there.

## One design note, not a change request

Collapsing to argmax means a near-uniform result like `[0.34, 0.33, 0.33]` emits a recipe that trains on one barely-preferred dataset while the `interleave` block above it advertises a mixture nothing reads. #330 explicitly permitted this over widening `DataConfig.train`, and your comment block says so plainly and points at `soup data merge`, so I accept it — I want it on the record as a decision rather than a default, and the real resolution is the `interleave` issue above.

Item 1 and I will merge.

MakazhanAlpamys:
Both filed, credited to you, each with a reproduction I ran on `main` rather than a restatement of your note:

- **#442** — the `--live` overlay bug. I also wrote down *why* the suite is green on it: every `--live` test in `test_v0535.py` mocks `subprocess.run`, so the overlay is built and handed to a mock and nothing ever loads it. That test gap is fixable independently of the shape decision.
- **#443** — `data.interleave` has no training-time reader. This one turned out to be the root of both: your PR collapses to argmax *because* `train` cannot express a mixture, and the `--live` renderer has the same problem for the same reason. I laid out the two honest resolutions (wire it into `load_dataset` and widen `train`, or reject it at parse time and name `soup data merge`) and said which way I lean, so whoever picks it up is choosing rather than guessing.

Neither is yours to fix as a condition of this PR. Take either if you want it.

MakazhanAlpamys:
Merged as `3426f2c`. Thank you — this one was a pleasure to review.

**All four items landed and each is now defended by a test that dies when I break the line it guards.** I re-ran the mutations myself rather than reading the diff: reverting the echo to `escape(train_value)` fails `test_apply_cli_quotes_path_needing_quoting` with exactly the `ScannerError` it exists to prevent; deleting either new guard fails its own test; and the original fix's mutations (`max` -> `min`, list emission restored) still kill four tests each. Zero test lines removed anywhere in the PR — I check that on every merge now, and it is not rhetorical.

Two things I want to name specifically.

**The `len(weights) < len(datasets)` guard is the one that earned its place.** Empty weights already blew up loudly on `max()`. The short-but-nonempty case did not: `max()` succeeded, `datasets[best_idx]` succeeded, and the renderer quietly produced a recipe that dropped datasets the caller had asked about. Refusing that is a different class of fix from tidying an ugly traceback, and the message names both lengths so the caller can see which side is wrong.

**Your two out-of-scope findings were both real, and they were the most valuable part of the PR.** I verified each on `main` myself before filing. #442 and #443 exist because of you, and #443 turned out to be the root of your own PR's shape — `data.train` collapses to argmax *because* nothing reads `interleave`. Finding a second instance of your own bug in a path you were not touching, and flagging it instead of quietly widening the diff, is exactly the judgement I want from a contributor.

Also verified: your rebase was clean and strictly additive — `merge-base` is the current `main` tip and the two-dot and three-dot diffs are byte-identical, so nothing of anyone else's was clobbered on the way. That matters this week; two PRs collided semantically on `main` a few days ago and cost a red build.

Two cosmetic things I did **not** ask you to change, recorded only so they are not mistaken for oversights: the two new guards sit mid-render rather than with the function's other input validation at the top, and the new tests use `try/except/else` where `test_v0480_part_b.py` uses `pytest.raises` 54 times. Neither is worth a round trip, and the mutation run proves each guard is individually pinned.

Your CONTRIBUTORS.md entry has its second bullet. #442 and #443 are both unclaimed if you want either.

Issue #443, opened 2026-08-18
data.interleave validates, documents and parses — but nothing reads it at training time, so a config that sets it silently trains on one dataset


MakazhanAlpamys:
Found by @blackcoderx while fixing #330. I verified it independently on `main` before filing.

`data.interleave` is a documented, schema-validated config key. `DataConfig.interleave` accepts `'concat' / 'under' / 'over' / ...` or a dict, has a field validator (`schema.py:669`), and `utils/data_pipeline.py` exports a real parser (`parse_interleave`, `InterleaveSpec`). A user setting it gets no error, no warning, and a config that loads.

**Nothing reads it at training time.**

```
$ grep -rn "parse_interleave\|InterleaveSpec" src/soup_cli/ --exclude-dir=__pycache__
src/soup_cli/config/schema.py:672:        # Shape validation only — full ``parse_interleave`` requires
   (plus its own definition and __all__ entry in utils/data_pipeline.py)

$ grep -rn "interleave" src/soup_cli/data/ src/soup_cli/trainer/ --exclude-dir=__pycache__
   (only unrelated prose: replay rows, online_dpo turns, TTS token interleaving)
```

`load_dataset()` handles a single-path `data.train` and nothing else. So a config that sets `interleave` trains on one dataset and reports nothing unusual — the failure mode this repo keeps writing down: **a flag the user explicitly set, silently doing nothing.** It is the same class as v0.73.0's Liger substring match (#78) and `use_cut_ce` (#383), except here there is not even a warning that fires.

It is also the root of two other items:
- #330 — the recipe `soup data mix --optimize` writes could not put a mixture in `data.train`, so the fix collapses to the single highest-weighted dataset and leaves the `interleave` block above it as decoration.
- the `--live` overlay bug filed alongside this — same shape, second renderer.

**Two honest resolutions, and this needs a deliberate choice rather than drift:**

**A — implement it.** Wire `parse_interleave` into `load_dataset`, widen `DataConfig.train` to `str | list[str]`, and both mix renderers become correct as written. This is what makes `soup data mix` mean what it prints: an optimiser that searches over mixture weights and then emits a recipe that cannot express a mixture is solving a problem the trainer does not have.

**B — remove or gate it.** Reject `interleave` at parse time with a message naming what to use instead (`soup data merge`), and drop it from the recipe/overlay renderers. Smaller, honest, and loses the feature.

[...]

PR #445, opened 2026-08-18
fix: soup data mix --live hands each proxy run a config it can load (#442)

### LINKED ISSUE #442: soup data mix --live writes an overlay config that fails to load: data.train is a list, and every --live test mocks subprocess.run so nothing ever loads it
Found by @blackcoderx while fixing #330, in a path that PR does not touch. Flagged rather than silently widened — filing it so it is not lost. I reproduced it independently on `main` before filing.

`soup data mix --live` runs each candidate mixture as a short real `soup train`. The overlay YAML it hands that subprocess is built by `utils/mix_proxy.py::_render_overlay_yaml`, which writes `data.train` as a **list**:

```python
data_block["train"] = list(datasets)
```

`DataConfig.train` is typed `str`, so the overlay fails at config load — before training starts, on **every** candidate. Reproduced on `main`:

```
data:
  train:
  - a.jsonl
  - b.jsonl
  interleave:
    strategy: probs
    probs: [0.6, 0.4]

LOAD: FAIL -> ValueError  data -> train: Input should be a valid string
```

This is the same defect #330 reports for the recipe writer, in a second renderer. #330's fix does not reach it.

**Why the suite is green anyway, which is the part worth keeping.** `tests/test_v0535.py` mocks `subprocess.run` in every `--live` test, so the overlay is built and then handed to a mock. Nothing ever runs it through `load_config_from_string`. The tests assert the orchestration and never the artifact — so a config the tool cannot itself load reads as a passing feature.

**Fix path.** Decide it together with #443 (`data.interleave` has no reader), because the honest fix depends on that answer:
- if `interleave` stays unimplemented, `_render_overlay_yaml` must emit a single-path `train`, mirroring what #330's fix does for the recipe writer, and say which dataset it picked;
- if `interleave` gets wired into `load_dataset`, then `train` should accept a list and both renderers stay as they are.

Either way the test gap is the same and is fixable now: one test that takes the string `_render_overlay_yaml` actually returns and runs it through `load_config_from_string`.

**Acceptance**
- [ ] The overlay `--live` generates loads through `load_config_from_string`.
- [ ] A test asserts that on the real returned string, not on a mock's call args — it must fail if the renderer regresses to a list-shaped `train`.
- [ ] `--live` is exercised end-to-end at least once without `subprocess.run` mocked away, or the reason it cannot be is written down in the test file.

blackcoderx:
Fixes #442.

Same defect #330 fixed in the recipe writer, found in the second renderer while
working that issue: `_render_overlay_yaml` (used by `soup data mix --live` to
build the temp config each candidate proxy run trains against) emitted
`data.train` as a YAML list, but `DataConfig.train` is typed `str` — so every
`--live` candidate got handed a config it could not load.

Every `--live` test in `test_v0535.py` mocks `subprocess.run`, so the overlay
gets built and handed straight to a mock — nothing ever loaded it through the
real schema, which is how this shipped green in the first place.

**Fix:** `data.train` now renders as the single highest-weighted dataset in
each candidate (same choice #330 made), with a comment above it noting which
dataset was picked and why. `data.interleave` is left as-is — still written,
still not consumed by training, same status quo as the recipe writer.

**Test:** `test_overlay_yaml_loads_through_config_schema` loads the *returned
string* through `load_config_from_string` — it fails if the renderer
regresses to a list-shaped `train`, unlike the existing mocked tests. Confirmed
it fails pre-fix with the same `data -> train: Input should be a valid string`
error before the fix, passes after.

Deliberately not touching #443 (`data.interleave` has no training-time
consumer) here — that's the bigger, separate decision this issue explicitly
defers to it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,15 @@ reproducing 70+ versions of notes.
   so no information is lost, and the comment explains that `data.interleave`
   is not yet consumed by training.
 
+- **`soup data mix --live` handed every candidate proxy run a config it could
+  not load (#442).** `_render_overlay_yaml` emitted `data.train` as a YAML
+  list, the same shape #330 fixed in the recipe writer, but every `--live`
+  test mocked `subprocess.run` so nothing ever loaded the overlay through the
+  schema — a config the tool could not itself load read as a passing feature.
+  `data.train` now renders as the single highest-weighted dataset in each
+  candidate, mirroring #330's fix, with a comment noting which dataset was
+  picked and why.
+
 - **Layer streaming now verifies that every trainable LoRA parameter has real
   storage after PEFT attaches the adapter (#433 reported by @lesterppo, fixed by @Faisal01011 in #435 and #437).** PEFT 0.18 creates streamed
   adapters on `meta` for Soup to materialise, while PEFT 0.19 may create them as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,17 @@ reproducing 70+ versions of notes.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`soup data mix --live` handed every candidate proxy run a config it could
+  not load (#442 by @blackcoderx in #445).** `_render_overlay_yaml` emitted `data.train` as a YAML
+  list, the same shape #330 fixed in the recipe writer, but every `--live`
+  test mocked `subprocess.run` so nothing ever loaded the overlay through the
+  schema — a config the tool could not itself load read as a passing feature.
+  `data.train` now renders as the single highest-weighted dataset in each
+  candidate, mirroring #330's fix, with a comment noting which dataset was
+  picked and why.
+
 ## [0.73.3] - 2026-08-18
 
 ### Added
@@ -91,15 +102,6 @@ reproducing 70+ versions of notes.
   search; the full ranked weight/path breakdown is kept as a comment above it
   so no information is lost, and the comment explains that `data.interleave`
   is not yet consumed by training.
-
-- **`soup data mix --live` handed every candidate proxy run a config it could
-  not load (#442).** `_render_overlay_yaml` emitted `data.train` as a YAML
-  list, the same shape #330 fixed in the recipe writer, but every `--live`
-  test mocked `subprocess.run` so nothing ever loaded the overlay through the
-  schema — a config the tool could not itself load read as a passing feature.
-  `data.train` now renders as the single highest-weighted dataset in each
-  candidate, mirroring #330's fix, with a comment noting which dataset was
-  picked and why.
 
 - **Layer streaming now verifies that every trainable LoRA parameter has real
   storage after PEFT attaches the adapter (#433 reported by @lesterppo, fixed by @Faisal01011 in #435 and #437).** PEFT 0.18 creates streamed
--- a/src/soup_cli/utils/mix_proxy.py
+++ b/src/soup_cli/utils/mix_proxy.py
@@ -173,6 +173,16 @@ def _render_overlay_yaml(
     We round-trip via ``yaml.safe_load`` / ``yaml.safe_dump`` to defeat any
     accidental key collisions; weights + datasets are sanitised above so the
     output cannot inject keys (the renderer emits scalars only).
+
+    ``data.train`` renders as a single string — the highest-weighted dataset
+    in *this* candidate — because ``DataConfig.train`` is typed ``str`` and
+    ``soup train`` has no reader for a weighted multi-dataset mixture
+    (``data.interleave`` is validated at config-load time but never consumed
+    by ``load_dataset()``). Emitting the full dataset list here produced a
+    proxy-run config ``soup train`` itself could not load (#330's defect in
+    this module's own renderer — #442). ``data.interleave`` is still written
+    below so the searched mixture isn't lost, with a comment noting which
+    dataset was picked and why.
     """
     import yaml  # noqa: PLC0415
 
@@ -183,12 +193,35 @@ def _render_overlay_yaml(
     if not isinstance(data_block, dict):
         data_block = {}
         raw["data"] = data_block
-    data_block["train"] = list(datasets)
+    best_idx = max(range(len(weights)), key=lambda i: weights[i])
+    data_block["train"] = datasets[best_idx]
     data_block["interleave"] = {
         "strategy": "probs",
         "probs": [float(w) for w in weights],
     }
-    return yaml.safe_dump(raw, sort_keys=False)
+    dumped = yaml.safe_dump(raw, sort_keys=False)
+
+    # Insert an explanatory comment directly above the `train:` line inside
+    # the `data:` block — a targeted line insert, not a value edit, so it
+    # can't touch anything the schema will parse.
+    lines = dumped.split("\n")
+    data_idx = next((i for i, ln in enumerate(lines) if ln == "data:"), -1)
+    if data_idx >= 0:
+        for i in range(data_idx + 1, len(lines)):
+            ln = lines[i]
+            if ln and not ln.startswith(" "):
+                break  # left the data: block
+            if ln.startswith("  train:"):
+                lines.insert(
+                    i,
+                    f"  # picked {datasets[best_idx]!r} (weight "
+                    f"{weights[best_idx]:.6f}, highest in this candidate) — "
+                    "soup train has no reader for a weighted multi-dataset "
+                    "mixture, see #330/#442/#443",
+                )
+                break
+        dumped = "\n".join(lines)
+    return dumped
 
 
 def _read_final_eval_loss(
--- a/tests/test_v0535.py
+++ b/tests/test_v0535.py
@@ -516,6 +516,28 @@ def test_overlay_yaml_loads_through_config_schema(tmp_path):
     assert cfg.data.train == "b.jsonl"
 
 
+def test_overlay_comment_does_not_drift_from_train_value(tmp_path):
+    # Maintainer's review of #442: the comment spliced above `train:` names
+    # which dataset was picked, but nothing tied that claim to the actual
+    # value — it could vanish (mutation: delete the insertion block) or lie
+    # (mutation: comment names datasets[0] while train: keeps best_idx) and
+    # every other test still passes. Tying the assertion to the *parsed
+    # config's* value rather than the renderer's own variable is what makes
+    # the second mutation fail here.
+    from soup_cli.utils.mix_proxy import _render_overlay_yaml
+
+    base_text = _make_base_yaml(tmp_path).read_text(encoding="utf-8")
+    overlay = _render_overlay_yaml(
+        base_text, ("a.jsonl", "b.jsonl"), (0.3, 0.7)
+    )
+    cfg = load_config_from_string(overlay)
+    lines = overlay.split("\n")
+    idx = next(i for i, ln in enumerate(lines) if ln.startswith("  train:"))
+    comment = lines[idx - 1]
+    assert comment.lstrip().startswith("#")  # fails if the block is deleted
+    assert repr(cfg.data.train) in comment  # fails if the comment lies
+
+
 def test_proxy_happy_path_reads_tracker(tmp_path, monkeypatch):
     monkeypatch.chdir(tmp_path)
     base = _make_base_yaml(tmp_path)
--- a/tests/test_v0535.py
+++ b/tests/test_v0535.py
@@ -495,6 +495,27 @@ def test_proxy_timeout_raises_runtimeerror(tmp_path, monkeypatch):
             )
 
 
+def test_overlay_yaml_loads_through_config_schema(tmp_path):
+    # #442 — same defect #330 fixed in the recipe writer, in the second
+    # renderer: `_render_overlay_yaml` emitted `data.train` as a YAML list,
+    # but `DataConfig.train` is typed `str`, so every `--live` candidate got
+    # handed a config it could not load. Every other test in this section
+    # mocks `subprocess.run`, so this is the one that actually loads the
+    # *returned string* through the real schema — it fails if the renderer
+    # regresses to a list. (A fully unmocked `--live` run needs a real
+    # training subprocess/GPU and isn't practical in this suite; the
+    # remaining `--live` coverage below intentionally mocks `subprocess.run`
+    # and asserts orchestration instead.)
+    from soup_cli.utils.mix_proxy import _render_overlay_yaml
+
+    base_text = _make_base_yaml(tmp_path).read_text(encoding="utf-8")
+    overlay = _render_overlay_yaml(
+        base_text, ("a.jsonl", "b.jsonl"), (0.3, 0.7)
+    )
+    cfg = load_config_from_string(overlay)
+    assert cfg.data.train == "b.jsonl"
+
+
 def test_proxy_happy_path_reads_tracker(tmp_path, monkeypatch):
     monkeypatch.chdir(tmp_path)
     base = _make_base_yaml(tmp_path)


MakazhanAlpamys:
Reviewed at `5755011` in an isolated worktree with `PYTHONPATH` pinned to that tree. **The fix is correct and one detail of it is better than the sibling it copies.** Two trivial blockers and one real gap.

## What holds

**Reproduced independently of your test.** I loaded `origin/main`'s `mix_proxy.py` through `importlib`, built the overlay, and fed it to the schema:

```
pre-fix:  train: [a.jsonl, b.jsonl]   LOAD: FAIL -> ValueError  data -> train: Input should be a valid string
PR head:  train: b.jsonl              LOAD: OK    data.train = 'b.jsonl'
```

I also ran your new test against the **whole pre-fix file** rather than a hand-mutation — `1 failed` with that same error. It is genuinely red before the fix, which is the property the existing `--live` tests could not have.

**Three behavioural mutations, all killed:**

| mutation | result |
|---|---|
| `datasets[best_idx]` -> `list(datasets)` (the literal #442 regression) | KILLED — `test_overlay_yaml_loads_through_config_schema` |
| argmax -> argmin | KILLED — `assert 'a.jsonl' == 'b.jsonl'` |
| `train` -> `""` | KILLED — `assert '' == 'b.jsonl'` |

The third is worth naming: an empty `train: ""` still *loads*, so a test that only asserted loadability would have missed it. Yours asserts the value. That is the right shape and I would keep it.

**Your YAML handling is better than #330's, and I want that on the record.** The recipe writer needed `json.dumps` because it string-formats the path into the line; you route the value through `yaml.safe_dump`, so quoting is the dumper's problem. I checked eight hostile shapes — colon, `#`, leading space, `*anchor`, a Windows `C:/` path — and every one round-trips through `safe_load` and the schema, with `safe_dump` auto-quoting the two that need it. That is not a gap you got away with; it is a cleaner solution than the one you were copying.

**Diff hygiene:** zero test lines removed. Nothing in the diff the description omits. `156 passed, 5 skipped` across `test_v0535.py`, `test_v0480_part_b.py` and `test_issue330_mix_recipe_train_loadable.py`; ruff clean.

## Blockers — both trivial, and the first is partly my timing

**1. The CHANGELOG entry is in the frozen `[0.73.3]` section.** v0.73.3 was tagged and published about an hour before you opened this, which rolled `[Unreleased]` into it — so your entry landed at line 95, inside `### Fixed` of the released section, while `## [Unreleased]` at line 13 is empty. Move the bullet up to `[Unreleased]` under its own `### Fixed`. Not your mistake; you branched before the roll.

**2. Missing the attribution suffix.** It reads `(#442).`; the house format is `(#442 by @blackcoderx in #445)` — the same form your own #440 entry three bullets up already carries. I now treat this as part of the change rather than something I reconstruct at release time, because reconstructing it is exactly when it goes wrong: four suffixes in v0.73.3 turned out wrong or incomplete when I checked them against `gh` yesterday.

## The real gap: the comment block can vanish or lie, and CI stays green

22 of your 37 added lines are the comment insertion, and nothing tests it. Two mutations survive:

- **delete the entire insertion block** -> `156 passed`
- **make the comment name `datasets[0]` while `train:` keeps `best_idx`** -> I ran this one myself rather than take it from my reviewer: **`56 passed, 2 skipped`**. The overlay then says `# picked 'a.jsonl' (weight ..., highest in this candidate)` above a `train: b.jsonl`.

A debugging aid that can silently disagree with the value it explains is worse than no aid — someone reading that temp config on a failed candidate would trust the comment. Your existing test *does* catch the insert corrupting the YAML (dropping the `#` prefix fails it with a `ScannerError`), so the gap is specifically absence and disagreement, not corruption.

Five lines close it, and I confirmed they pass on your code and fail on both mutants:

```python
overlay = _render_overlay_yaml(BASE, ("a.jsonl", "b.jsonl"), (0.3, 0.7))
cfg = load_config_from_string(overlay)
lines = overlay.split("\n")
idx = next(i for i, ln in enumerate(lines) if ln.startswith("  train:"))
comment = lines[idx - 1]
assert comment.lstrip().startswith("#")      # fails if the block is deleted
assert repr(cfg.data.train) in comment       # fails if the comment lies
```

The second assertion is the load-bearing one: it ties the comment to the parsed config rather than to the renderer's own variable, so the two cannot drift.

## Three nits — your call, none blocking

**A test on the file `--config` actually receives.** Your reasoning for not running `--live` unmocked is that it needs a GPU, which is true — but the artifact-level assertion does not. Intercept `subprocess.run`, read `argv[argv.index("--config") + 1]`, and `load_config()` that path. I wrote it: `1 passed in 0.45s`, no GPU, and it fails pre-fix. It also covers the temp-file write and the `encoding="utf-8"` round trip — which now matters slightly more than it did, because your comment introduces a non-ASCII em dash the pre-fix overlay never had. (I checked: both the write at `mix_proxy.py:303` and the read at `config/loader.py:17` pass `encoding="utf-8"` explicitly, so there is no live bug — it is simply now an untested dependency.)

**Guard asymmetry with #330.** The recipe writer refuses empty or length-mismatched weights with named errors. Here the same shapes give `max() arg is an empty sequence`, an `IndexError`, and — for `len(datasets) > len(weights)` — a **silent wrong answer** with the extra dataset never considered. All three are unreachable today, because `proxy_run_for_weights` runs `_validate_datasets` and `_validate_weights` first. But `_render_overlay_yaml` is now called directly by a test, so it is a seam.

**The overlay no longer names the datasets `interleave.probs` refers to.** Before, `train: [a, b]` listed them; now only the winner appears and `probs` is a bare float list. #330 keeps the full ranked breakdown plus a pointer at `soup data merge`. Small — the temp dir is `rmtree`'d — but it is debugging value on exactly the runs you would want to debug.

## One recommendation, since you will hit this question

Should the two renderers share a helper? **Yes — but extract the guards, not the argmax.** The argmax is one expression and wrapping it would be over-engineering. What is worth single-sourcing is #330's two named `ValueError`s as a `pick_best_dataset(datasets, weights)`, because the argument for it is the history in front of us: **#442 exists precisely because one renderer was fixed and its twin was not**, and the two now disagree about what an invalid shape even is. Keep the comment rendering separate — a human-review recipe and a machine-consumed temp overlay legitimately want different text, and forcing those together is where it would become over-engineering.

Not a condition of this PR. If you want it, it is its own change.

## What I need

The two CHANGELOG fixes and the comment control. The three nits are yours to take or leave — tell me if you would rather I file the shared-helper one as an issue so it does not get lost.

Turning #442 around within two hours of it being filed, on a defect you found yourself and deliberately did not fold into the earlier PR, is the sequence I would want every time.

blackcoderx:
Thank you for this — especially the comment-drift test. Tying the assertion to `repr(cfg.data.train)` rather than the renderer's own `best_idx` variable is the right shape and I missed it; a debugging aid that can silently disagree with the value it explains is worse than none.

Before I push the two CHANGELOG fixes: something changed on my end mid-review that affects whether this PR should exist in its current form at all.

I'm taking on #443 next — wire `data.interleave` into `load_dataset()`, not gate it. Your own #442 issue text already called this exact outcome: *"if interleave gets wired into `load_dataset`, then `train` should accept a list and both renderers stay as they are."* That's literally true once #443 lands — the pre-fix `_render_overlay_yaml` (plain `list(datasets)`, real `interleave` probs) is already the shape a working interleave feature needs. So #443's diff reverts this PR's core change (the collapse to a single best-weighted dataset) rather than building on it.

Your call on how to sequence it:
1. I push the two required CHANGELOG fixes + the comment-drift test now, this merges as an interim correctness fix, and #443 reverts the collapse-to-single-dataset logic back to the full list once it lands — extra churn in history, but nothing sits broken in the meantime.
2. I close this one and let #443 resolve both #442 and #443 directly, since the eventual code is the same either way.

Either is fine by me — wanted to flag it before spending your CHANGELOG-fix cycles on something that might get reverted in the same week.

MakazhanAlpamys:
Option 1 — push the fixes, this merges now. And thank you for asking rather than either silently pushing or silently stalling; the question is the right one.

But your premise has a hole in it, and it is the whole answer.

**#443 has two honest resolutions and only one of them reverts this PR.** I wrote both into the issue deliberately:

- **A — implement it.** Wire `parse_interleave` into `load_dataset`, widen `DataConfig.train` to `str | list[str]`. Then yes, both renderers go back to emitting a list and this PR's collapse is reverted.
- **B — remove or gate it.** Reject `interleave` at parse time with a message naming `soup data merge`, and drop it from both renderers. Then this PR's collapse is the **final** shape, not an interim one.

You are planning for A as though it were decided. It is not, and it is not mine alone to decide either — widening `DataConfig.train` is a schema change, so it goes to the maintainer before anyone spends a week on it. I am asking that question now, separately.

So the sequencing follows from that:

**Merge this.** `soup data mix --live` is broken in `main` today — every candidate proxy run gets a config it cannot load, before training even starts. Leaving a known-broken command in place because a *hypothetical* larger change *might* supersede it is the wrong trade, and it is the wrong trade even if A is eventually chosen: the interval between now and A landing is real, and the revert is one commit with a legible message.

Two further things worth saying plainly:

**A is a much bigger piece of work than the issue's one-line framing suggests**, and I flagged three of its unanswered questions in #443 itself: streaming datasets, `val_split` semantics across multiple sources, and the `packing` / `multipack` interaction. Any one of those can turn A into a multi-week change or into "actually, B". Do not treat A as a formality.

**If A does land, reverting this is cheap and honest.** `git revert` with "superseded by #443, which makes a list-shaped `data.train` valid" is a better history than a gap where the command was knowingly broken. Churn in history is not a cost worth protecting against; a broken command in `main` is.

So: push the two CHANGELOG fixes (into `[Unreleased]`, not the frozen `[0.73.3]` — that section rolled about two hours before you opened this, which is why it conflicted) plus the comment-drift control, and I will merge it.

On #443 itself: **claim it in the issue before you start**, and wait for the schema answer. That is our rule from CONTRIBUTING.md and I would rather you not sink days into branch A before it is decided. I will post the answer in #443 as soon as I have it.

blackcoderx:
Pushed `618dc01` — all three:

1. Moved the CHANGELOG entry into `[Unreleased]`.
2. Added the `(#442 by @blackcoderx in #445)` attribution suffix.
3. Added your comment-drift test verbatim. Confirmed it fails against both mutations before trusting it — deleting the insertion block, and making the comment name `datasets[0]` while `train:` keeps `best_idx` — and passes on the current code.

`test_v0535.py` 57 passed, 2 skipped; ruff clean.

Thanks for writing the exact assertions rather than describing the shape — saved me a round of guessing, and `repr(cfg.data.train) in comment` tying it to the parsed value instead of the renderer's own variable is a pattern I'll reuse.

MakazhanAlpamys:
Merged as `a840f31`. Third merge, and the second time in a day you have fixed a defect you found yourself.

**What I ran before merging.** Reproduced the bug independently of your test — loaded `origin/main`'s `mix_proxy.py` through `importlib` and fed the overlay to the schema, same `ValueError`. Confirmed your new test is red against the *whole* pre-fix file rather than a hand-mutation. Three behavioural mutations each killed. Eight hostile path shapes — colon, hash, leading space, YAML anchor, a Windows `C:/` path — all round-trip exactly.

**And I re-ran the drift mutation myself** rather than take your word: with the comment naming `datasets[0]` while `train:` keeps `best_idx`, `test_overlay_comment_does_not_drift_from_train_value` now fails. It survived before your push. That is the whole point of a control and it works.

Two things worth naming specifically.

**`yaml.safe_dump` was the better choice, and it was yours, not a copy.** The recipe writer needed `json.dumps` because it string-formats the path into the line; you routed the value through the dumper instead, so quoting stopped being your problem. I checked eight shapes expecting to find the one that broke — it auto-quotes the two that need it. You improved on the code you were mirroring.

**You asked the sequencing question instead of guessing.** Pushing quietly and letting me discover the #443 overlap later, or stalling and waiting for me to notice, would both have been worse. The answer turned on something you had not considered — #443 has two honest resolutions and only branch A reverts this — but the instinct to surface it was right, and it is the reason nothing was wasted.

`soup data mix --live` was broken in `main` for every candidate before your fix. It is not now.

On #443: you claimed it and said you would wait for the schema answer, which is exactly right. The question is with the maintainer. My own recommendation, which I will post in the issue with the reasoning, is **neither branch as written** — I checked and `soup data merge` only concatenates, no proportions, so gating `interleave` alone would leave `soup data mix --optimize` searching for weights nothing can apply. Sit tight; you will get a decision, not a shrug.

MakazhanAlpamys:
Follow-up on something I found after merging, and it is my miss rather than a complaint.

`tests/test_issue330_mix_recipe_train_loadable.py`, which came in with your #440, carried these two lines at the top:

```
# SPDX-License-Identifier: AGPL-3.0-only
# Copyright 2026-present the Unsloth AI Inc. team. All rights reserved. See /studio/LICENSE.AGPL-3.0
```

I removed them in `008da45`. This project is Apache-2.0 and publishes to PyPI on every tag, so a foreign SPDX id on a file inside it is a licensing claim we cannot make — not a formatting nit.

**The part that is on me:** I had stripped the identical two lines out of two other test files nine hours earlier, in `8e4f012`. Then I merged yours carrying the same header, having read the diff. Knowing about a defect did not stop me missing it — which is the argument for a check rather than for me trying harder.

So there is now one: `tests/test_no_foreign_license_headers.py` fails the build on a foreign licence header in any tracked file. Verified both directions — re-adding the exact header turns it red, removing it turns it green, and the sweep over 975 tracked files is clean.

One question, and there is no wrong answer: **do you generate files from a template, or use automated tooling that might?** @ousamabenyounes disclosed on #420 that they do, and the header references `/studio/LICENSE.AGPL-3.0` while the service they named is at `studio.oratelecom.net` — so the same source may be reaching both of you. If it is fixable once at that end, it never costs either of us a CI round trip again. If you have no idea where it came from, that is a useful answer too, because it means I am looking in the wrong place.

Nothing about your work changes. The mix fixes were good and the review found no defect in either of them that you had not already reasoned about.

blackcoderx:
Thank you for catching this and fixing it rather than just flagging it — and for the CI guard, that's the right response to "this happened twice already."

To your question directly: yes. I use an AI coding assistant (Claude Code) for a good chunk of this work, and that header came from it — it wrote that SPDX/copyright block into the new test file without checking this repo's actual license first. There's no template or company-side tooling behind it that I can find; I checked this repo (it's the only file in it with that header) and every other project I have locally, and it doesn't appear anywhere else. My read is it's fabricated boilerplate — "Unsloth" is a real project in the same space as Soup, and the assistant most likely pattern-matched a header style from similar training-tooling codebases without grounding it in what's actually true here. Not something reaching you from a shared source with @ousamabenyounes as far as I can tell, though I obviously can't rule out convergent causes.

I don't have a fix to hand you for the assistant's behavior itself — that's outside this repo — but I'll be checking new files it writes for exactly this before they go into a diff going forward. Appreciate you asking instead of assuming.

ousamabenyounes:
I want to clarify this with evidence from our side, because we have now traced the exact sequence in our orchestration logs.

I contribute to several projects around AI training and fine-tuning. That same interest is what led me to invest time in Soup: I find it genuinely exciting to see AI agents contribute to important work in this field.

Immediately before the affected Soup contributions, our system had been working on this contribution in the AGPL-licensed model-training repository named in the foreign header:

https://tinyurl.com/23wtjuk6

I used a shortened URL deliberately so this explanation does not create a GitHub cross-reference or notification on that repository and disturb its maintainers.

In that repository, the SPDX and copyright lines were legitimate project headers. Our logs show the agent reading those exact lines between 07:02 and 07:18 UTC on August 15.

A few hours later, the same contribution process moved to Soup and created #407 and #408. The new test files in those PRs incorrectly received the identical project-specific header. The same contamination later appeared again in #440.

So I can confirm that this came from our side. It was not independently fabricated boilerplate. Our sequential contribution workflow failed to fully discard the previous repository's context before moving to Soup.

Reusing that context was intended to reduce repeated context loading, but it produced the opposite result: it wasted tokens by carrying irrelevant project material, created unnecessary review noise, and introduced a real licensing risk.

We are fixing this at the orchestration level. Every repository and every pull request will now run in a fresh sandbox with a fresh context containing only that project's files, contribution rules, issue history, and relevant knowledge. Repository-specific code, licence headers, company information, and raw context from a previous contribution will never be carried into another project.

Once a contribution is completed and its recoverable work is safely published, its sandbox and task context will be destroyed.

Thank you for identifying the pattern and for adding the CI guard. Your inference that the same source could be reaching both contributors was correct, and our logs allowed us to confirm exactly how it happened.

Issue #78, opened 2026-05-08
[QA] Verify FlashAttention + Liger Kernel + Unsloth speedups


MakazhanAlpamys:
## What needs verifying

Three orthogonal speedup paths, all gated behind config flags:

1. `use_flash_attn: true` + `flash_attn_version: 2` (or `3` on Hopper+)
2. `use_liger: true` (Liger Kernel fused ops: RMSNorm / SwiGLU / RoPE / cross-entropy)
3. `backend: unsloth` (FastLanguageModel — claims 2-5× speedup on a single GPU)

## Hardware / software prerequisites

- NVIDIA GPU with ≥ 12 GB VRAM (Ampere+ for FlashAttention 2; Hopper / Blackwell for FA3)
- `pip install flash-attn --no-build-isolation` (FA2)
- `pip install 'soup-cli[liger]'` (Liger Kernel)
- `pip install 'soup-cli[fast]'` (Unsloth)
- A baseline single-GPU `soup train` run to compare against

## Steps

```bash
# Baseline
soup train --config soup.yaml --output ./baseline

# +FA
edit soup.yaml: training.use_flash_attn: true
soup train --config soup.yaml --output ./fa

# +Liger
edit soup.yaml: training.use_liger: true
soup train --config soup.yaml --output ./liger

# Unsloth (different backend, fresh config)
soup recipes use llama-3.2-1b-sft-unsloth
soup train --config soup.yaml --output ./unsloth
```

Compare loss curves + steps/sec between the four runs.

## Success

- All four runs converge to similar loss
- FlashAttention: ≥ 1.3× faster than baseline on the same GPU
- Liger: ≥ 1.1× faster, with lower peak VRAM
- Unsloth: ≥ 1.5× faster than transformers backend (often 2-3×)
- No silent degradation: `soup chat --model ./fa` produces coherent output

## Report back

- `soup version --full`, `pip show flash-attn liger-kernel unsloth`
- GPU model + driver + CUDA
- Per run: steps/sec, peak VRAM, final loss
- Any "falling back to slow path" warnings

## If it fails

Open separate bug per failing component. Note: FlashAttention 3 is Hopper+ only — don't file a bug if FA3 falls back to FA2 on Ampere.

MakazhanAlpamys:
## Measured for the first time — on an H100, since neither kernel installs on Windows

[...]

PR #446, opened 2026-08-18
fix(cce): resolve cut_ce architecture from config, not a name substring

### LINKED ISSUE #383: training.use_cut_ce cannot engage on any pip-installed stack, and the warning names two causes that are both false
Measured while writing `benchmarks/probe-v0.73.0-what-bounds-streaming.md`. The flag has shipped since v0.28.0 and, as far as this measurement can tell, has never been able to turn on.

## Three independent blockers, each measured

1. **The PyPI wheel does not contain the module the patcher imports.** `cut-cross-entropy==25.1.1` ships without `cut_cross_entropy.transformers`, which is exactly what `apply_cut_ce` imports.
2. **Building from Apple's repository does not fix it.** 25.9.3 has the module, but it fails on import against current `transformers` (removed private symbols `_CONFIG_FOR_DOC`, then `GEMMA2_INPUTS_DOCSTRING`) — and it fails inside the **gemma2** patcher even for a Llama model.
3. **Windows.** `triton` is registered under the distribution name `triton_windows`, which 25.9.3's metadata lookup does not find.

`apply_cut_ce` swallows all three `ImportError`s and returns `False`.

## The user-visible half is worse than the failure

The warning says "no matching architecture found or cut_cross_entropy not installed". On a Llama model with the package installed, **both stated causes are false**, so the message actively points away from the real one. The trainer does at least print a visible warning rather than degrading silently — unlike #78's Liger path — so this is a diagnosability bug, not a correctness one.

## Why it is worth fixing rather than deprecating

With the kernel wired in by hand, CCE measurably helps on the streaming path: marginal VRAM per micro-batch falls **1.063 → 0.319 GB (3.33x)** and the workable micro-batch goes **1 → 3** for Llama-3.1-8B NF4 at S=512.

Two things to decide alongside the fix:

- **The pre-flight would refuse configurations CCE makes fit.** `LOGITS_BYTES_PER_ELEMENT` charges 14 B/element; with CCE on, the measured marginal cost is ~0. See #327.
- **CCE removes bit-exactness as an available criterion.** Under CCE a *resident* run stops reproducing against itself (9.5e-03 between two runs of the same model), so streamed-vs-resident can no longer be asserted at 0.0. That is CCE's nondeterminism, not streaming's, but it changes what the correctness gate can claim.

## Acceptance criteria

- [ ] `apply_cut_ce` distinguishes "package absent" / "package present but its `transformers` integration is missing" / "architecture unsupported" / "patch raised", and says which
- [ ] the architecture patcher is selected before it is imported, so a Llama run never fails inside the gemma2 patcher
- [ ] a supported `cut-cross-entropy` version range is pinned and declared in an extra
- [ ] the interaction with the streaming pre-flight (#327) is resolved or documented
- [ ] a test that fails when the flag silently no-ops

AmirF194:
## Root cause

`apply_cut_ce()` picked its CCE patcher by matching an architecture keyword against the model path's last component. That works for a hub id (`meta-llama/Llama-3.1-8B`), but every local checkpoint directory Soup itself produces (`checkpoint-2000`, `my-finetune`, `soup merge`/`soup shrink`/`soup draft distill` output) carries no such keyword, so `use_cut_ce: true` silently no-opped with no error. Separately, `phi-2`, `phi-3` and `phi-4` all mapped to the same `"phi3"` patcher, even though `cut_cross_entropy` has no Phi-2 patcher: Phi-2's real `config.model_type` is `"phi"`, not `"phi3"`.

## Fix

Detection now resolves `AutoConfig.from_pretrained(model_name).model_type` first, mirroring the identical fix already shipped for Liger Kernel (#78), and falls back to the last-path-component match only when config resolution is unavailable (offline, no `config.json`). Phi-2 correctly reports unsupported instead of silently running under the Phi-3 kernel. The two call sites (`trainer/sft.py`, `utils/v028_features.py`) that hand-wrote diverging copies of the "no matching architecture" advisory now share one constant.

## Verification

- New tests in `tests/test_training_speed.py` (`TestCutCEDetectsArchitectureFromTheConfig`, `TestCutCEUnifiedMessage`): a local checkpoint directory with no architecture in its name is detected via a real `AutoConfig` resolution against a tiny on-disk Llama checkpoint; Phi-2 correctly reports unsupported rather than reusing the Phi-3 patcher; an architecture `cut_cross_entropy` genuinely has no patcher for still reports `False`; the offline fallback still works when config resolution has nothing to read.
- Full `tests/test_training_speed.py` (106 tests) and the two files it shares fixtures with (`test_v07202.py`, `test_v07300.py`) pass; `ruff check src/soup_cli/ tests/` clean.
- Not checked: real Hub-network config resolution for a gated/private repo id (only local-path and offline-fallback paths were exercised; the code path is identical to the already-merged Liger Kernel fix in #78).

Fixes #383


MakazhanAlpamys:
Reviewed at `8933479` in an isolated worktree with `PYTHONPATH` pinned to that tree. **The fix is correct, well-targeted, and faithful to the Liger precedent — and your Phi-2 finding is worse than you wrote it.** Three small blockers, one of which contradicts a claim in the PR body.

## The defect and the fix, reproduced independently

I pulled the ground truth from upstream rather than from your description — `gh api repos/apple/ml-cross-entropy/.../patch.py` gives `PATCH_FNS = {llama, phi3, gemma2, mistral, qwen2}`, unknown key raises. Then loaded both versions of `cut_ce.py` side by side against a synthetic module mirroring that, with `HF_HUB_OFFLINE=1`:

```
meta-llama/Llama-3.1-8B   main -> True  ['llama']    #446 -> True  ['llama']
local checkpoint-2000     main -> False []           #446 -> True  ['llama']
local my-finetune         main -> False []           #446 -> True  ['llama']
local phi-2 (type=phi)    main -> True  ['phi3']     #446 -> False ['phi']
```

**Your Phi-2 half understates what was happening.** `PATCH_FNS` has no `"phi"` key at all, and `patch_phi3` ends in `modeling_phi3.Phi3ForCausalLM.forward = cce_forward`. A Phi-2 model is `PhiForCausalLM` — so nothing was patched, `apply_cut_ce` returned `True` anyway, and `sft.py` printed *"Cut Cross-Entropy enabled: large-vocab CE replaced with chunked CCE kernel"*. The accurate sentence is "silently reports success while patching an unrelated class", which is a stronger finding than the one you claimed.

**The Liger comparison holds.** `cut_ce._detect_model_type` and `liger._detect_model_type` are behaviourally equivalent — same `AutoConfig.from_pretrained(trust_remote_code=False)`, same bare `except Exception -> ""`, same fallback order. Not a third divergent copy, which is what I was watching for. One place yours is stricter and better: you match the last path component where liger matches the whole lowercased name.

**Your stated gap is benign** — I forced `AutoConfig.from_pretrained` to raise `OSError("gated repo ... 401")`: a hub id carrying the arch still resolves `True ['llama']`, a nameless one gives `False []`. The bare `except` absorbs it, no crash. Nothing to close before merge.

## Blocker 1 — `tests/test_training_speed.py` is RED on its own

Your PR body says "Full `tests/test_training_speed.py` (106 tests) ... pass". It does not reproduce here:

```
$ pytest tests/test_training_speed.py -q --no-cov -p no:randomly
FAILED ...TestCutCEDetectsArchitectureFromTheConfig::test_a_local_checkpoint_dir_with_no_architecture_in_its_name_is_still_detected
FAILED ...TestCutCEDetectsArchitectureFromTheConfig::test_phi2_is_not_dispatched_to_the_phi3_patcher
2 failed, 104 passed in 16.29s
```

I ran that myself before writing this. The mechanism: your change makes the legacy `test_apply_cut_ce_calls_llama_patch_for_llama_model` call the **real** `_detect_model_type`, which imports transformers and attempts a live hub fetch for the gated `meta-llama/Llama-3.1-8B` — **inside** `patch.dict("sys.modules", {...})`. On exit `patch.dict` restores the snapshot and evicts every module imported in the block; the next `import tokenizers` re-initialises a PyO3 extension and dies:

```
tokenizers in sys.modules after patch.dict exit: False
RE-IMPORT FAILS: PyO3 modules compiled for CPython 3.8 or older may only be
initialized once per interpreter process
```

Red in four different orderings. It is masked in a full-suite run only because a file that eagerly imports `tokenizers` sorts earlier — so **CI will probably be green and this will still be broken** for anyone running the file the way our own checklist tells them to. That asymmetry is the reason I run targeted files rather than trusting the suite.

Secondary defect on the same line: that unit test now performs real network I/O — `_detect_model_type('meta-llama/Llama-3.1-8B')` takes **6.76 s** to return `''`.

One line fixes both, and I verified it: add `patch("soup_cli.utils.cut_ce._detect_model_type", return_value="")` to that test, exactly as its sibling `test_apply_cut_ce_deepseek_phi_does_not_use_phi` already does. With it: **106 passed**.

## Blocker 2 — the priority claim is unpinned

Config-over-name is the central claim of the fix, and moving the config block *below* the name-match loop leaves your entire new suite green (5 passed). Nothing constructs a model whose name says one architecture and whose config says another — which is exactly the shape a `soup merge` / `soup shrink` output directory named after its source model produces.

Control I wrote and verified (fails on the inversion with `At index 0 diff: 'phi3' != 'llama'`, passes on your code): a directory `phi-3-mini-finetune/` containing `config.json` with `model_type: llama`, asserting `cce_patch` is called with `llama`.

## Blocker 3 — an `ImportError` from inside the patcher now kills the run

Pre-fix the whole call sat inside `except (ImportError, AttributeError, NotImplementedError)`. You hoisted the import (still catching `ImportError` there) but neither call site's tuple lists it, and `trainer/sft.py` has **no** try/except around `apply_cut_ce`:

```
patcher raises ImportError:
  main  llama -> False  calls=['llama']
  #446  llama -> RAISED ImportError: simulated failure inside the patcher
```

Not hypothetical: `patch_llama` / `patch_phi3` do a lazy `from transformers.models.X import modeling_X` at call time, and #383's own blocker 2 documents CCE patchers breaking against current transformers. Two words — add `ImportError` to both `except` tuples.

## Two smaller items

**The gemma change is undescribed.** `(("gemma",), "gemma")` became `(("gemma2", "gemma-2"), "gemma2")`, and it appears in no part of the PR body, the CHANGELOG or the docs, with zero test coverage. It is a real improvement — pre-fix any Gemma on the fallback path **crashed the run** (`RuntimeError: Unknown model type gemma`), now it correctly reports `False`. That deserves a sentence and a test rather than silence; a reviewer who did not diff it would never know it happened.

**CHANGELOG placement and attribution.** Your entry is in `[Unreleased]` on the branch, which was right when you wrote it — but v0.73.3 was tagged about two hours before you opened this and rolled that section into `## [0.73.3]`, which is why you conflict. On rebase, land it in the new empty `[Unreleased]` under a fresh `### Fixed`, not inside the frozen release. And add the attribution suffix: it reads `(#383)`, house form is `(#383 by @AmirF194 in #446)`.

**Also worth knowing:** no CI has run on this PR at all — GitHub cannot build a merge commit for a conflicted branch, so the `pull_request` workflows never triggered. The rebase will start them.

## One pre-existing bug I found while reviewing, not yours

`last_component = model_name.rsplit("/", 1)[-1]` never splits on a backslash, so on Windows the whole path is matched and the "last path component only" guarantee — which your docs line restates — does not hold:

```
rsplit("/")[-1]     -> 'C:\Users\...\test_phi2_i...'   contains "phi2": True
os.path.basename    -> 'step-2000'                     contains "phi2": False
```

This is pre-existing, not introduced by you, and it had a real effect on my review: one of my mutations died on Windows purely because the pytest tmp dir carried the test's own name, and would have survived on the Linux and macOS cells. I will file it separately unless you want it — `os.path.basename` plus a test with a `phi-2`-named directory that has no config.

## What I need

The one-line mock (blocker 1), the priority control (blocker 2), `ImportError` in both tuples (blocker 3), the gemma sentence, and the CHANGELOG rebase + attribution. All small.

Reproducing the defect before claiming the issue, and then finding the Phi-2 conflation on top of the reported bug, is why this was quick to review — the parts I checked hardest were the parts you had already got right.

AmirF194:
Thanks for the depth here, especially catching that the Phi-2 finding was worse than I'd written and that I'd broken my own test's isolation.

All three blockers fixed and pushed (3ce1f55, rebased onto main since v0.73.3 shipped without this while it was open):

- Blocker 1: `test_apply_cut_ce_calls_llama_patch_for_llama_model` now forces `_detect_model_type` to `""`, same pattern as its `deepseek_phi` sibling. Confirmed the crash reproduces on the old version and is gone now.
- Blocker 2: added `test_config_model_type_wins_over_a_conflicting_directory_name`, a `phi-3-mini-finetune/` directory with `config.json` declaring `llama`, asserting `cce_patch` gets called with `llama`.
- Blocker 3: `ImportError` added to both `except` tuples in `apply_cut_ce()`.

Gemma: added the sentence (comment + CHANGELOG) and `test_apply_cut_ce_plain_gemma_reports_false_instead_of_crashing`. CHANGELOG entry moved into the current `[Unreleased]` section with the `(#383 by @AmirF194 in #446)` attribution form.

`tests/test_training_speed.py` is 108/108 (`-q --no-cov -p no:randomly`, matching your repro command) and `ruff check` is clean on both touched files.

Left the Windows `rsplit`/`basename` bug for you to file separately as offered, it's real but pre-existing and outside this PR's scope.

AmirF194:
CI is red on all 9 test legs, but it is not this PR: the failure is `test_no_tracked_file_declares_a_foreign_license` flagging `tests/test_no_foreign_license_headers.py` itself, and this PR does not touch that file (checked the diff). Confirmed on main directly: 008da456 (the scanner commit) also fails the same test on its own branch, so the scanner is catching its own example header. Leaving this alone since you are clearly already on it; just flagging so it does not read as something my push broke.

MakazhanAlpamys:

```
```

MakazhanAlpamys:
I updated your branch with `main` myself rather than making you do it — the red was entirely my broken licence guard, so clearing it is my job, not yours. It is a merge of `main` into your branch, not a rebase, so none of your commits were rewritten.

CI is re-running now. Nothing in your diff changed.

MakazhanAlpamys:
Re-reviewed at `3ce1f55`. **All three blockers verified fixed, and I ran each one rather than reading the diff.** Approving; merging as soon as CI comes back green from the branch update I just pushed.

**Blocker 1 — the single-file run.** This is the one that mattered, because your original claim of "106 tests pass" did not reproduce here:

```
$ pytest tests/test_training_speed.py -q --no-cov -p no:randomly
108 passed in 42.20s
```

Was `2 failed, 104 passed`. The `patch.dict` / PyO3 interaction is gone, and the 6.76 s of real network I/O in a unit test with it.

**Blocker 2 — the priority claim is now pinned.** Neutering the config resolution (`model_type = ""`) fails three tests, including your new `test_config_model_type_wins_over_a_conflicting_directory_name`:

```
FAILED ...::test_phi2_is_not_dispatched_to_the_phi3_patcher
FAILED ...::test_config_model_type_wins_over_a_conflicting_directory_name
3 failed, 105 passed
```

The `phi-3-mini-finetune/` directory declaring `llama` is exactly the shape a `soup merge` output produces, which is what made the gap worth closing rather than a hypothetical.

**Blocker 3 — `ImportError` is in both tuples** (`cut_ce.py:95` and `:120`), so a patcher whose lazy `from transformers.models.X import ...` fails no longer aborts `_setup_transformers`.

**The gemma change is now described and tested** — CHANGELOG line 43 names it and `test_apply_cut_ce_plain_gemma_reports_false_instead_of_crashing` exists. That one went from a silent behaviour change to a documented one, which was the point: pre-fix, any Gemma on the fallback path *crashed the run* with `RuntimeError: Unknown model type gemma`.

**Also checked:** CHANGELOG entry is under the current `[Unreleased]` (line 35, above the frozen `## [0.73.3]` at line 52) with `(#383 by @AmirF194 in #446)`; the advisory constant is genuinely shared — `sft.py:892` and `v028_features.py:56` both import `NO_MATCHING_ARCHITECTURE_MESSAGE` from `cut_ce.py:22`, not two copies of a string; two removed test lines, both the replaced docstring; ruff clean.

Two things worth saying beyond the verdict.

**You diagnosed my bug independently and said so carefully.** Your 00:12 comment — confirming the licence-guard failure reproduces on `008da45` itself, checking that your diff does not touch that file, and flagging it so it would not "read as something my push broke" — is exactly the right handling. You were right, it was mine, and it is fixed on `main`.

**And you took the correction on the Phi-2 finding rather than defending the original framing.** The accurate version — `apply_cut_ce` returning `True` and printing "Cut Cross-Entropy enabled" while patching a class the model does not use — is a stronger finding than the one you filed, and it is now the one in the CHANGELOG.

The Windows `rsplit`/`basename` bug you left for a follow-up is the right call; it is pre-existing and I will file it separately unless you want it.

MakazhanAlpamys:
Merged as `5705a6c`. Thank you — this one had a real finding underneath the reported one, and you handled the correction well.

**What I ran before merging**, not read:

- `pytest tests/test_training_speed.py` alone → **108 passed**. That is the number your first round claimed and that did not reproduce; it does now.
- Neutering the config resolution (`model_type = ""`) → **3 failed**, including your new `test_config_model_type_wins_over_a_conflicting_directory_name`. The priority claim is pinned rather than asserted.
- `ImportError` present in both `except` tuples (`cut_ce.py:95`, `:120`).
- The advisory message is genuinely one constant — `sft.py:892` and `v028_features.py:56` both import `NO_MATCHING_ARCHITECTURE_MESSAGE` from `cut_ce.py:22`, not two copies of a string.
- Two removed test lines, both the replaced docstring. Ruff clean.

**The Phi-2 finding is the part I want on the record**, and the commit message states it in your corrected form rather than the original: upstream `PATCH_FNS` has no `"phi"` key, and `patch_phi3` assigns to `Phi3ForCausalLM.forward`, so a Phi-2 model had **nothing** patched while `apply_cut_ce` returned `True` and the run printed *"Cut Cross-Entropy enabled"*. Success reported for work not done is a worse defect than the silent no-op you set out to fix, and you found it on the way.

**The gemma change is in the commit message too**, because it deserved to be somewhere: pre-fix, any Gemma reaching the fallback path did not no-op, it *crashed* with `RuntimeError: Unknown model type gemma`. You added the sentence and the test when I asked, and that turned an undescribed behaviour change into a documented improvement.

Two things beyond the code.

**You diagnosed my bug and framed it correctly.** Your 00:12 note — confirming the licence-guard failure reproduces on `008da45` with zero of your changes, and flagging it so it would not "read as something my push broke" — is precisely right, and you were right.

**You took the correction instead of defending the framing.** That is rarer than good patches and it is why this review was short.

Your CONTRIBUTORS.md entry has its fifth bullet. The Windows `rsplit`/`basename` bug you deliberately left alone is still unclaimed — say the word and it is yours, otherwise I will file it.

AmirF194:
Thanks for the thorough re-review, and for taking the CI fix off my plate. Glad the Phi-2 finding turned out to matter more than the reported bug.

I'll pass on claiming the Windows rsplit/basename bug for now, please go ahead and file it.

MakazhanAlpamys:
Filed as #456, with the measurement from the review and the reason it is smaller than it sounds — #446 made the config path primary, so the name match is now only a fallback for offline or config-less directories.

One thing from your review went into it that I would not have written down otherwise: one of my mutations died on Windows **only because the pytest temp directory happened to carry the test's own name**. On a clean parent path that mutant was behaviourally identical, so it would have survived on the Linux and macOS cells. A test whose kill depends on what pytest named the directory is not a test, and that is now an acceptance criterion on #456.

Declining it was the right call — widening #446 to cover it would have made a clean three-blocker fix into something harder to review.

PR #447, opened 2026-08-18
fix(data): recognize mapping-like tokenizer outputs (#441)

### LINKED ISSUE #441: loss_mask: _coerce_token_ids is private-named but imported cross-module, and a duck-typed mapping output fails with a misleading error
Two follow-ups from the #430 work, both in `src/soup_cli/data/loss_mask.py`. Neither was in scope for the PR that shipped the fix — filing them so they are not lost, and so the second one is credited to the person who found it.

## 1. `_coerce_token_ids` is private-named but is imported across modules

`loss_mask.py:68` defines `_coerce_token_ids`, and `utils/data_doctor.py` imports it from there in **two** places:

```
src/soup_cli/utils/data_doctor.py:262:        _coerce_token_ids,
src/soup_cli/utils/data_doctor.py:618:    from soup_cli.data.loss_mask import _coerce_token_ids
```

A leading underscore is this repo's marker for "private to the module", and `data_doctor` reaching across for it means the name says one thing and the import graph says another. Its docstring already promises a stable contract (mapping-or-sequence in, `list[int]` out), and the whole point of the #430 fix was that `data_doctor` and the masking path must not drift — that shared helper IS the mechanism, so it should be spelled as one.

Fix path: rename to a public name (`coerce_token_ids`), update the three call sites in `loss_mask.py` and the two in `data_doctor.py`. Optional: keep `_coerce_token_ids = coerce_token_ids` as a one-line alias if any test refers to it by the old name; delete the alias if none do.

**Acceptance**
- [ ] Public name used at every call site; no module imports another module's underscore-prefixed symbol.
- [ ] Test that fails if the two modules ever get separate copies of this logic (the drift the shared helper exists to prevent).

## 2. A dict-like tokenizer output that is not registered as `collections.abc.Mapping` fails with a *misleading* error

`_coerce_token_ids` gates on `isinstance(out, Mapping)`. A real `transformers` `BatchEncoding` passes that (it is a `UserDict`, so it is registered) — but a tokenizer returning a dict-*like* object that is not registered does not, and the value then flows into `_coerce_int_list`, which iterates its **keys**. Measured on current `main`:

```
isinstance Mapping: False
coerce RAISES: ValueError tokenizer returned non-integer input_ids[0]='input_ids'
real BatchEncoding isinstance Mapping: True -> [1, 2, 3]
```

The object in the first line is a plain class with `get` / `__getitem__` / `keys` / `__iter__` holding `{"input_ids": [1,2,3], "assistant_masks": [0,1,1]}`. The message names `input_ids[0]='input_ids'`, i.e. it reports the key it just iterated as if it were a token — an operator reading that goes looking at their data, and the cause is the type check.

Reachable through `_tokenize_only` (`loss_mask.py:149`), which calls the helper on the raw output with no `Mapping` guard in front of it. The other path, `_apply_template_with_mask` (`loss_mask.py:105`), has the same gate and degrades more quietly: a duck-typed mapping returns `None` and silently falls back to the no-mask path, so a tokenizer that DID supply `assistant_masks` has them dropped.

The proposed shape is @emre155's, from #438: widen the check to `isinstance(out, Mapping) or hasattr(out, "get")`. **This is their finding** — it was the one substantive thing in #438 that the merged implementation did not carry, and I invited them to send it separately.

Fix path: widen the gate in `_coerce_token_ids`, and apply the same widening to the `isinstance(out, Mapping)` guard at `loss_mask.py:105` so the mask path is not silently skipped for the same objects. Keep the failure mode for a genuinely non-mapping, non-sequence input loud.

**Acceptance**
- [ ] A duck-typed mapping (has `get`, not registered as `Mapping`) returns `[1, 2, 3]` rather than raising.
- [ ] The same object carrying `assistant_masks` takes the mask path instead of falling back — a test that fails if only the first gate is widened.
- [ ] An object with `get` but no `input_ids` still raises the existing clear error, not a new confusing one.
- [ ] Mutation check: widening the gate is what makes the test pass; reverting the one-line change fails it.

@emre155 — part 2 is yours if you want it. Part 1 is small and independent; anyone can take it.

AchuthReddy-16:
## What does this PR do?

Fixes #441.

Tokenizer outputs that behaved like dictionaries but did not inherit from `collections.abc.Mapping` were incorrectly iterated as keys. The attention-mask path also skipped those objects.

This change:

* Uses one shared predicate to recognize mapping-like tokenizer outputs.
* Applies that predicate to token extraction and attention-mask handling.
* Makes `data_doctor` import the public helper so both modules remain consistent.
* Adds regression coverage for custom mapping-like tokenizer outputs.

## Type of change

* [x] Bug fix
* [ ] New feature
* [x] Refactor / cleanup
* [ ] Documentation
* [x] Tests

## Checklist

* [ ] `ruff check src/soup_cli/ tests/` passes
* [ ] `pytest tests/ -v` passes
* [x] Updated relevant docs if needed: no documentation changes required


MakazhanAlpamys:
Under review now — this note is not about the code.

**You did nothing wrong here, and I want that said before anything else.** But there is a queue problem around this issue and it is mine: I had offered part 2 to @emre155 (it is their finding — from an earlier closed PR of theirs) at 09:05Z, and part 1 to @harshitthek at 15:37Z, thirty-four minutes before you opened this. Both offers were made in comments on *other* PRs, where nobody reading #441 could see them, and #441's own thread has zero comments. By our actual convention — claim in the issue thread — this issue was free when you took it.

The full account is in #441 so all three of you read the same thing rather than three different versions of it.

Practical consequence for you: I have given @emre155 until tomorrow to say whether they want to implement their own finding. If they do, I will hold the part-2 half and merge the rest; if they stay quiet, #447 goes in whole with emre155 credited in the CHANGELOG as the finder. Either way your work is not wasted and your credit is not diminished — and if the wait costs you anything, say so and I will find you the next thing myself.

The review is running independently of all of this. It will land shortly.

MakazhanAlpamys:
Reviewed at head in an isolated worktree with `PYTHONPATH` pinned to that tree. CI 12/12. **Technically this is ready — one CHANGELOG line short.** I did this one by hand rather than delegating, so everything below is a command I ran.

## Both parts are done, and the harder half is done properly

**Part 2's second gate is the thing most people would have missed.** The issue named two instances of the same `isinstance(out, Mapping)` check — the loud one in the coercion path, and the quiet one in `_apply_template_with_mask` at line 105, which does not raise but returns `None` and silently drops `assistant_masks` a tokenizer did supply. Widening only the first would have looked complete and left the silent one. You widened both, through one shared `_is_mapping_like` predicate, and the acceptance test proves it:

```
# revert ONLY the second gate, leave the first widened:
FAILED tests/test_issue441_coerce_token_ids.py::TestDuckTypedMapping::test_duck_mapping_with_assistant_masks_takes_the_mask_path
FAILED tests/test_issue441_coerce_token_ids.py::TestOneMappingPredicate::test_both_call_sites_share_the_mapping_predicate
2 failed, 40 passed
```

That is exactly the mutation the issue asked to be guarded against, and it dies. Reverting the *first* gate instead kills 4 tests across the suite. `isinstance(out, Mapping)` now lives in precisely one place, and `test_both_call_sites_share_the_mapping_predicate` is what keeps it there — a good test, because the failure mode being prevented is "someone widens one and not the other", which is how this bug existed in the first place.

**Part 1 is clean.** `coerce_token_ids` is public, all four `data_doctor` sites use the public name, and there is **no underscore alias left behind** — `grep -rn "_coerce_token_ids" src/` returns nothing. Leaving a compatibility alias would have been the tempting move and would have re-created the exact ambiguity the rename removes.

**Behaviour preservation, all four cases run:**

```
duck-typed mapping (not a registered Mapping) -> [1, 2, 3]
missing input_ids  -> ValueError: tokenizer output mapping has no 'input_ids'
non-mapping, non-sequence -> ValueError
real transformers BatchEncoding -> [7, 8]
```

The third and fourth matter as much as the first: widening a type gate is the kind of change that quietly starts accepting garbage, and it does not here.

`243 passed, 1 skipped` across `test_issue441_*`, `test_assistant_mask.py`, `test_v07127.py`, `test_chat_template.py`. Ruff clean. `test_cli_startup_is_light.py` 23/23 — `loss_mask` is on a light path and the new predicate adds no import. Zero test lines removed anywhere in the diff.

**The `CONTRIBUTING.md` change is legitimate and I checked it before asking.** It refreshes the test counts (369 -> 370 files, 18141 -> 18151) and adds your row to the tests table. That is the release checklist being followed, not scope creep — and it is the file I myself missed on the v0.73.3 release, so thank you for not making me catch it twice.

## The one thing I need

**The CHANGELOG entry reads `(#441)` — no attribution, and no mention of @emre155.**

Part 2's finding is theirs. It came out of an earlier PR of theirs that I did not merge, and it was the one substantive thing that PR carried which the merged implementation did not. The issue text says so explicitly. Please make it:

```
(#441 by @AchuthReddy-16 in #447, part 2 found by @emre155)
```

or any wording that names both. Your implementation is yours and that is not in question — but a reader of the CHANGELOG should be able to see where the finding came from, and right now it reads as though the whole thing appeared from nowhere.

## Why I am not merging tonight

Not about your code. As explained in #441 and in my note above, I had offered part 2 to @emre155 with a standing invitation and gave them until tomorrow to say whether they want to implement their own finding. I said that publicly, so I am going to honour it rather than merge twelve hours earlier and make the promise worthless.

Concretely: push the CHANGELOG line whenever convenient, and this merges as soon as that window closes — with everything else already verified above, so there will be no second round.

MakazhanAlpamys:
Merged as `6d51924`. Thank you — and the reason this went through quickly is the half of it I did not ask for.

**You did not stop at the reported instance.** #441 described one gate: `_coerce_token_ids` raising a misleading error on a dict-like output. You found the second one — the same `isinstance(out, Mapping)` check at `_apply_template_with_mask`, where it does *not* raise. It returns `None` and falls back to the no-mask path, so a tokenizer that genuinely supplied `assistant_masks` has them silently dropped and training runs unmasked. Widening only the loud gate would have fixed the visible error and left the worse defect in place. That is the shape of #430 all over again — the failure with no exception and a normal-looking loss curve — and you caught it without being pointed at it.

**What I ran before merging**, rather than reading the diff:

- `214 passed, 1 skipped` across the new file plus `test_assistant_mask.py` and `test_v07127.py`.
- Reverted your widening back to `isinstance(out, Mapping)`: **3 tests fail**, including the one asserting the error message for a `get`-having object with no `input_ids`. So the fix is genuinely guarded, not merely accompanied by tests.
- Confirmed the rename actually landed everywhere: `grep _coerce_token_ids src/` is empty, and `data_doctor` imports the public name at both sites.
- Zero test lines removed anywhere in the diff. Ruff clean.

**Credit where it is shared.** The duck-typing half was @emre155's finding in #438, and the CHANGELOG entry names them alongside you — you wrote that attribution yourself when I asked, including the "part 2 found by" clause, which is more care than the format required.

Your CONTRIBUTORS.md entry has its second bullet.

One note on timing so it does not look arbitrary: I held this overnight for @emre155 to take the part-2 half themselves, said so, and they did not reply. That window is closed, not forgotten.

Issue #308, opened 2026-07-15
Latent bug: duck-typed TrainerCallbacks (ReLoRACallback, HFPushCallback) crash on HF event dispatch


MakazhanAlpamys:
## Context
Surfaced during the v0.71.34 code review. HF `transformers.trainer_callback.CallbackHandler.call_event` dispatches **every** Trainer lifecycle event via `getattr(callback, event)(...)` with **no `hasattr` guard** (verified against transformers 4.57.6). A callback added via `trainer.add_callback(obj)` that does NOT subclass `transformers.TrainerCallback` therefore raises `AttributeError` on the first event it doesn't implement (e.g. `on_epoch_begin`, which fires immediately after `on_train_begin`).

Two existing callbacks are plain duck-typed classes (not `TrainerCallback` subclasses) and are subject to this:
- `utils/relora.py::ReLoRACallback` (only `on_step_end`)
- `monitoring/hf_push.py::HFPushCallback` (only `on_train_begin` / `on_save`)

They would crash before the first optimizer step if actually run through a real `Trainer.train()`. v0.71.34's `LisaCallback` (#267) was written the same way, hit the crash in review, and was fixed by subclassing `TrainerCallback` via a lazy factory (mirroring `monitoring/curriculum_callback.py::_try_import_callback_base`, which keeps the module import-light while inheriting the no-op event stubs).

## Proposed fix path
Make `ReLoRACallback` and `HFPushCallback` real `TrainerCallback` subclasses using the same lazy `_try_import_callback_base()` pattern as `curriculum_callback.py` / `utils/lisa.py` (keeps the modules transformers-free at import time while inheriting the no-op defaults for the ~13 unimplemented events).

## Acceptance criteria
- [ ] `ReLoRACallback` and `HFPushCallback` are `isinstance(cb, transformers.TrainerCallback)`
- [ ] A test builds a real (tiny) `Trainer` with each callback and runs `train()` for a couple of steps without `AttributeError`
- [ ] No new top-level transformers import in `utils/relora.py`

MakazhanAlpamys:
Closing — this landed in `10e5522` (2026-07-19), `fix(callbacks): make ReLoRACallback + HFPushCallback real TrainerCallback subclasses (#308)`, and the acceptance criteria are all covered on `main` today. Re-verified before closing rather than taking the commit message's word for it.

[...]

PR #455, opened 2026-08-19
perf(callbacks): defer TrainerCallback resolution to first access (#320)

### LINKED ISSUE #320: Eleven TrainerCallback modules import transformers at module scope (lazy factory called eagerly)
## The pattern

Eleven modules resolve their `TrainerCallback` base **at class-definition time**:

```python
def _get_trainer_callback_base():
    try:
        from transformers import TrainerCallback   # looks lazy
        return TrainerCallback
    except ImportError:
        return object

_TrainerCallbackBase = _get_trainer_callback_base()   # ...but CALLED at module scope

class RewardHackMitigationCallback(_TrainerCallbackBase):
    ...
```

The factory is lazy; the call is not. Importing any of these modules imports
`transformers` and therefore `torch` — measured at **~4.4 s** each:

| module | |
|---|---|
| `utils/reward_hack_control.py` | `_TrainerCallbackBase = ...` at module scope |
| `utils/reward_hacking.py` | same |
| `utils/echo_trap.py` | same |
| `utils/minillm.py` | same |
| `utils/rl_checkpoint.py` | same |
| `utils/relora.py` | `class ReLoRACallback(_try_import_callback_base())` |
| `utils/lisa.py` | same, inline in the class header |
| `monitoring/hf_push.py` | same |
| `monitoring/curriculum_callback.py` | same |
| `monitoring/grpo_stability_callback.py` | `_TrainerCallbackBase = ...` |
| `monitoring/callback.py` | leaks at import |

`trainer/raft.py:187` resolves the base **inside a factory function** and stays
clean (0.04 s) — that is the reference implementation already in the repo.

## Why it matters (and why it is not urgent)

On the trainer path this costs nothing: `transformers` is loaded anyway.

It becomes a user-visible regression the moment a **light** command imports one.
That is exactly what happened in v0.71.41: `commands/reward.py` imported
`utils/reward_stress`, which imported `utils/reward_hack_control` for a single
string constant, and `soup --help` went from **1.15 s to 6.0 s**.

That specific leak is fixed, and `tests/test_cli_startup_is_light.py` now asserts
the real invariant at runtime (`import soup_cli.cli` → `torch` not in
`sys.modules`), so a recurrence fails CI loudly instead of silently. This issue
is about the ten remaining latent modules.

## Why the existing guards did not catch it

Every one of these modules has a `test_no_top_level_torch`-style guard. They
parse the module source and assert it contains no top-level `import torch`.
That proves a **syntactic** property; the requirement is a **runtime** one. An
AST walk cannot see through a transitive import, nor through a lazily-written
factory that is called at module scope.

`tests/test_v07126.py::test_no_top_level_heavy_import_in_source` is the sharpest
example: it *imports* `reward_hack_control` (loading transformers, ~4.7 s) and
then asserts that the source has no top-level transformers import. The test
performs the very thing it certifies against, and passes.

## Proposed fix

Cheaper than it looks — **no `super()` call exists in any of the eleven class
bodies** (verified), so nothing needs re-indenting into a function. Two viable
shapes:

1. **PEP 562 module `__getattr__`** — build the class on first attribute access,
   cache it. `from mod import XCallback` and `isinstance` keep working.
2. **`type()` rebuild** — keep the body on a plain `object` base, synthesise the
   real subclass lazily.

Either way the acceptance bar is the same as #308: the resulting object must be a
real `transformers.TrainerCallback` subclass, because HF's `CallbackHandler`
dispatches every lifecycle event through `getattr(cb, event)(...)` with no
`hasattr` guard.

## Acceptance criteria

- [ ] All eleven modules import without pulling `transformers` / `torch`
- [ ] Each callback is still `isinstance(cb, TrainerCallback)` when transformers
      is installed (the #308 property — do not regress it)
- [ ] Each still inherits the no-op stubs for unimplemented events
      (`cb.on_epoch_begin(None, None, None)` must not raise)
- [ ] Extend `tests/test_cli_startup_is_light.py` with a runtime per-module check
      covering all eleven
- [ ] Retire or downgrade the per-file AST guards these modules carry — they
      certify a property they cannot observe
- [ ] Correct the `tests/test_issue308_callback_subclass.py` docstring claim that
      the factory "keeps the modules transformers-free at import time"

## Not in scope

Making these modules light has **no user-visible benefit today** — none of them
is on a light command path. The value is removing a false claim and making a
future light import safe by construction.

AchuthReddy-16:
## What does this PR do?

Fixes #320.

Eleven callback modules resolved `TrainerCallback` while their classes were being defined. This eagerly imported `transformers` and `torch` whenever those modules were imported.

This PR creates the callback classes lazily through module-level `__getattr__`. The training dependencies are now loaded only when a callback class is first accessed. The resolved class is cached, so later access returns the same class without rebuilding it.

Regression tests cover lazy imports, subclass behavior, no-op dependency stubs, and class caching.

## Type of change

* [x] Bug fix
* [ ] New feature
* [x] Refactor / cleanup
* [ ] Documentation
* [x] Tests

## Checklist

* [x] `ruff check src/soup_cli/ tests/` passes
* [x] `pytest tests/ -v` passes
* [x] Updated relevant docs if needed: no documentation changes required

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,16 @@ reproducing 70+ versions of notes.
 - **A repo-wide documentation ratchet to guarantee declared recipe counts stay synchronized with the catalog (#453 by @harshitthek in #457).**
   Derives the expected count dynamically from `len(RECIPES)` and scans all declared documentation sites, preventing silent Git auto-merge drift across sequential recipe additions.
 
+### Changed
+
+- **Lazy callback builders now self-import their callback class names so runtime
+  lookup never raises `NameError` while preserving lazy heavy-dependency loading
+  (#320 by @AchuthReddy-16 in #455).** `build_echo_trap_callback`,
+  `build_reward_hack_callback`, `build_minillm_callback`,
+  `build_rl_checkpoint_callback`, and `build_push_callback` now resolve their
+  callback types through local module imports in the builder body, and the
+  regression suite adds subprocess coverage for all five builder calls.
+
 ### Fixed
 
 - **`cut_ce.py` and `liger.py` now normalize path separators and match architecture
--- a/src/soup_cli/monitoring/callback.py
+++ b/src/soup_cli/monitoring/callback.py
@@ -3,15 +3,12 @@
 from __future__ import annotations
 
 import logging
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 from rich.console import Console
-from transformers import (
-    TrainerCallback,
-    TrainerControl,
-    TrainerState,
-    TrainingArguments,
-)
+
+if TYPE_CHECKING:
+    from transformers import TrainerControl, TrainerState, TrainingArguments
 
 from soup_cli.monitoring.display import TrainingDisplay
 
@@ -19,7 +16,17 @@ logger = logging.getLogger(__name__)
 console = Console()
 
 
-class SoupTrainerCallback(TrainerCallback):
+def _get_trainer_callback_base():
+    """Lazy-resolve ``transformers.TrainerCallback``."""
+    try:
+        from transformers import TrainerCallback
+
+        return TrainerCallback
+    except ImportError:
+        return object
+
+
+class _SoupTrainerCallback_body:  # noqa: N801
     """Bridges HF Trainer events to Soup's Rich live display and experiment tracker."""
 
     def __init__(
@@ -565,3 +572,22 @@ class SoupTrainerCallback(TrainerCallback):
             f"({new_batch}, {new_accum}). "
             f"Restart training with the new pair to take effect."
         )
+
+
+_LAZY_CALLBACKS = {
+    "SoupTrainerCallback": _SoupTrainerCallback_body,
+}
+_BODY_SKIP = frozenset(("__dict__", "__weakref__"))
+
+
+def __getattr__(name: str):  # PEP 562
+    body = _LAZY_CALLBACKS.get(name)
+    if body is not None:
+        base = _get_trainer_callback_base()
+        ns = {k: v for k, v in vars(body).items() if k not in _BODY_SKIP}
+        cls = type(name, (base,), ns)
+        cls.__module__ = __name__
+        cls.__qualname__ = name
+        globals()[name] = cls
+        return cls
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/src/soup_cli/monitoring/curriculum_callback.py
+++ b/src/soup_cli/monitoring/curriculum_callback.py
@@ -59,7 +59,7 @@ _VALID_CURRICULUM_METRICS = frozenset({"length", "perplexity", "loss"})
 _SIGNAL_WINDOW = 512
 
 __all__ = [
-    "DynamicCurriculumCallback",
+    "DynamicCurriculumCallback",  # noqa: F822
     "_is_rank_zero",
     "_pick_bucket",
 ]
@@ -140,7 +140,7 @@ def _try_import_callback_base():
         return object
 
 
-class DynamicCurriculumCallback(_try_import_callback_base()):  # type: ignore[misc]
+class _DynamicCurriculumCallback_body:  # type: ignore[misc]  # noqa: N801
     """HF TrainerCallback emitting dynamic curriculum bucket-weight history.
 
     BETA: the callback is live in v0.53.5 but the sampler-side consumer
@@ -510,3 +510,22 @@ class DynamicCurriculumCallback(_try_import_callback_base()):  # type: ignore[mi
                 os.unlink(tmp_path)
             except OSError:
                 pass
+
+
+_LAZY_CALLBACKS = {
+    "DynamicCurriculumCallback": _DynamicCurriculumCallback_body,
+}
+_BODY_SKIP = frozenset(("__dict__", "__weakref__"))
+
+
+def __getattr__(name: str):  # PEP 562
+    body = _LAZY_CALLBACKS.get(name)
+    if body is not None:
+        base = _try_import_callback_base()
+        ns = {k: v for k, v in vars(body).items() if k not in _BODY_SKIP}
+        cls = type(name, (base,), ns)
+        cls.__module__ = __name__
+        cls.__qualname__ = name
+        globals()[name] = cls
+        return cls
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/src/soup_cli/monitoring/grpo_stability_callback.py
+++ b/src/soup_cli/monitoring/grpo_stability_callback.py
@@ -145,12 +145,7 @@ def filter_zero_advantage(advantages, *, eps: float = 1e-8) -> Any:
 
 
 def _get_trainer_callback_base():
-    """Lazy-resolve ``transformers.TrainerCallback`` (v0.53.11 review fix).
-
-    Project policy: every callback inherits TrainerCallback so HF Trainer
-    discovers it via the callback handler. We resolve at class-body
-    evaluation time so module import does not pull transformers.
-    """
+    """Lazy-resolve ``transformers.TrainerCallback``."""
     try:
         from transformers import TrainerCallback
 
@@ -159,10 +154,7 @@ def _get_trainer_callback_base():
         return object
 
 
-_TrainerCallbackBase = _get_trainer_callback_base()
-
-
-class GRPOStabilityCallback(_TrainerCallbackBase):  # type: ignore[misc, valid-type]
+class _GRPOStabilityCallback_body:  # type: ignore[misc, valid-type]  # noqa: N801
     """HF TrainerCallback that wires v0.50.0 Part D stability knobs.
 
     Lazy-inherits ``transformers.TrainerCallback`` so the module is
@@ -325,3 +317,22 @@ class GRPOStabilityCallback(_TrainerCallbackBase):  # type: ignore[misc, valid-t
             if entry:
                 log_history.append(entry)
         return control
+
+
+_LAZY_CALLBACKS = {
+    "GRPOStabilityCallback": _GRPOStabilityCallback_body,
+}
+_BODY_SKIP = frozenset(("__dict__", "__weakref__"))
+
+
+def __getattr__(name: str):  # PEP 562
+    body = _LAZY_CALLBACKS.get(name)
+    if body is not None:
+        base = _get_trainer_callback_base()
+        ns = {k: v for k, v in vars(body).items() if k not in _BODY_SKIP}
+        cls = type(name, (base,), ns)
+        cls.__module__ = __name__
+        cls.__qualname__ = name
+        globals()[name] = cls
+        return cls
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/src/soup_cli/monitoring/hf_push.py
+++ b/src/soup_cli/monitoring/hf_push.py
@@ -15,7 +15,10 @@ from __future__ import annotations
 import logging
 import re
 from pathlib import Path
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
+
+if TYPE_CHECKING:
+    from soup_cli.monitoring.hf_push import HFPushCallback
 
 from soup_cli.utils.hf import get_hf_api, resolve_endpoint, resolve_token, validate_repo_id
 
@@ -56,7 +59,7 @@ def _try_import_callback_base():
         return object
 
 
-class HFPushCallback(_try_import_callback_base()):  # type: ignore[misc]
+class _HFPushCallback_body:  # type: ignore[misc]  # noqa: N801
     """Real HF ``TrainerCallback`` auto-pusher.
 
     Subclasses the lazily-resolved ``TrainerCallback`` so it inherits the no-op
@@ -327,7 +330,7 @@ def build_push_callback(
     output_dir: str,
     explicit_token: Optional[str] = None,
     private: bool = False,
-) -> Optional[HFPushCallback]:
+) -> Optional["HFPushCallback"]:
     """Factory that resolves token/endpoint and builds the callback.
 
     Returns None when no HF token is available — the caller logs and skips
@@ -343,6 +346,8 @@ def build_push_callback(
         logger.warning("HF_ENDPOINT invalid (%s); skipping auto-push", exc)
         return None
 
+    from soup_cli.monitoring.hf_push import HFPushCallback
+
     return HFPushCallback(
         repo_id=repo_id,
         token=token,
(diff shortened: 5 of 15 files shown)


MakazhanAlpamys:
Something went wrong on the push — **the branch does not contain the change this PR describes.**

The whole diff is one line:

```
$ gh pr diff 455 --name-only
.gitignore

+screenpipe/
```

and the branch has exactly one commit: `0c325f2 chore: remove accidentally committed screenpipe submodule`. The lazy `__getattr__` work is not there — `git show pr455:src/soup_cli/utils/relora.py | grep -c __getattr__` returns `0`.

My read is that you hit an accidentally-committed `screenpipe/` submodule, cleaned it up with a reset or a rebase, and the cleanup ended up as the only commit on the branch — taking the real work with it. I am not treating this as anything other than a git mishap; I check diff-against-description on every PR because a mismatch is occasionally something worse, and this is plainly not that.

**Your work is probably still recoverable.** In your clone:

```
git reflog --date=relative | head -30
```

Look for the commit(s) from before the cleanup, then `git branch recovered <sha>` to pin them somewhere safe before doing anything else. If you rebased, `git reflog show fix/issue-320-lazy-callback-import` narrows it to that branch's history specifically. Stash entries survive too: `git stash list`.

If it is genuinely gone, say so and re-push — no harm done, nothing is blocked on it.

**And the `screenpipe/` line is worth keeping either way**, separately from #320. If a tool of yours drops a directory into the repo, ignoring it is correct and I will take that as its own one-line PR. Just not inside the #320 PR, so the history stays readable.

#320 is still yours. Nothing about this changes that.

AchuthReddy-16:
Recovered the #320 changes from 08b5fa3 and rebuilt the branch cleanly on the latest upstream/main. The unrelated #441 and screenpipe changes are no longer included.

Verification:
- Ruff passed
- 77 targeted callback, startup, and checkpoint tests passed

MakazhanAlpamys:
Reviewed at `0ef1b14` in an isolated worktree. **The mechanism works and both hard properties are achieved — and there is a critical blocker that ships broken flags.** Taking those in that order, because the first part is real and I do not want it lost in the second.

## What works, measured

**#320 is genuinely solved, 11 modules out of 11.** Cold subprocess import, per module:

| module | main | this PR |
|---|---|---|
| `utils/reward_hacking` | transformers+torch, 4.474 s | **neither, 0.024 s** |
| `monitoring/hf_push` | transformers+torch, 5.272 s | **neither, 0.025 s** |
| `utils/relora` | transformers+torch, 3.448 s | **neither, 0.022 s** |

…and the same for the other eight. That is the property #320 asked for, and it is the runtime one — not an AST scan.

**#308 is intact, which was the trap I warned you about.** After first access all eleven satisfy `issubclass(cls, transformers.TrainerCallback)`, `__bases__ == ['TrainerCallback']`, and `inst.on_epoch_begin(None, None, None)` returns without raising. Making the base fake kills six tests. You made the import lazy without making the base class a lie, which is the hard part.

**And `__getattr__` is done carefully**: `from mod import X` works, a wrong name raises a proper `AttributeError`, repeated access returns the *same* class object (so `isinstance` holds across calls), and it survives a pickle round-trip. The diff is exactly the described work, the `screenpipe/` line is gone, and the 9 removed test lines are docstring prose that #320 explicitly asked to be corrected — no assertion touched.

## Blocker: five builders raise `NameError` on a cold import

PEP 562 module `__getattr__` is consulted for attribute access **on the module object** — never for bare global-name resolution *inside* the module. Five builder functions still reference the class as a bare global. I ran this myself:

```
$ python -c "import soup_cli.utils.echo_trap as M; M.build_echo_trap_callback(threshold=0.6)"
*** NameError: name 'EchoTrapCallback' is not defined
```

Same shape in `build_reward_hack_callback`, `build_minillm_callback`, `build_rl_checkpoint_callback`, `build_push_callback`. Confirmed structurally too — for each of the four I checked, the referenced name is **not** in the module's globals.

It appears to work only if something already fetched the attribute, which caches it into `globals()`. Nothing on the real training path does. Driven through the actual wiring, `attach_rl_callbacks` with `echo_trap_enabled=True` crashes — and `NameError` is not caught by the `except (TypeError, ValueError)` handlers in `peft_wiring.py`, so it kills the run rather than degrading.

**Flags this breaks:** `--echo-trap-enabled`, `--reward-hack-detector`, `--minillm-enabled`, `--rl-checkpoint-save-every-steps`, `--push-as`.

**The fix is small** and keeps the laziness — a self-import inside the builder goes through the module object and therefore through `__getattr__`:

```python
def build_reward_hack_callback(...):
    from soup_cli.utils.reward_hacking import RewardHackCallback
    return RewardHackCallback(...)
```

Verified working, with the cold import still light.

## The part worth more than the fix: ruff found this and it was silenced

The PR adds **10 `# noqa: F821`** across five files. F821 is "undefined name". Removing two of them:

```
F821 Undefined name `RewardHackCallback`  --> src/soup_cli/utils/reward_hacking.py:592:6
F821 Undefined name `RewardHackCallback`  --> src/soup_cli/utils/reward_hacking.py:599:12
```

The linter pointed at the exact two lines that fail at runtime, and the suppression turned the one automated signal that caught it into `All checks passed!`. I am not reading that as carelessness — a lazily-created class genuinely *is* undefined at module scope, so silencing F821 looks like the obvious accommodation. But it is the reason CI reported nine identical failures instead of naming five broken builders.

Where a `noqa` is genuinely needed here, it is for the **return annotations** (`-> EchoTrapCallback`), and those want `if TYPE_CHECKING:` plus a quoted annotation rather than blanket suppression — that way the annotation resolves for type checkers and ruff stays able to see a real undefined name.

## Why the tests did not catch it, which needs its own fix

CI shows one failing test. Run file-by-file in isolation the PR produces **12 failures across 7 files**, all green on `main`. Proof:

```
$ pytest tests/test_code_review_high.py                             -> 1 failed, 28 passed
$ pytest tests/test_v0700_part_d.py tests/test_code_review_high.py  -> 62 passed
```

Eleven of the twelve vanish when an earlier file happens to touch the attribute first. So the suite is currently blind to this whole class of defect, and will stay blind after you fix these five unless something isolates the import.

**What closes it:** a parametrized subprocess test — fresh interpreter per module, import the builder, call it, assert no `NameError`. That is the control that would have turned CI red on all five at once instead of on one by accident of alphabetical ordering. It is also the only kind of test that can see a cold-import bug at all.

## Smaller things

- **CHANGELOG entry missing** — needs one under `[Unreleased]` with `(#320 by @AchuthReddy-16 in #455)`.
- `from mod import *` no longer exports the class for 9 of the 11 modules (no `__all__`); the two that have one needed a `# noqa: F822`. Nothing in-repo star-imports these, so this is a decision to record rather than a bug.
- No `__dir__`, so the names are invisible to `dir()` and tab-completion until first access.
- Method `__qualname__` still reads `_ReLoRACallback_body.__init__`, so a `TypeError` traceback names a class the user cannot import.
- `_XCallback_body` remains a module attribute and is **not** a `TrainerCallback` subclass — the #308 trap one typo away. Worth `del`-ing after registration.
- #320's "retire or downgrade the per-file AST guards" criterion is not addressed.

## Where this leaves us

The approach is right, the two properties that were hard to get are both genuinely there, and the failure is one specific interaction of PEP 562 that is easy to miss precisely because everything looks correct until a cold process calls a builder. Fix the five builders, drop the ten suppressions, add the subprocess control — and this is a good change.

Take the time you need; nothing is blocked behind it.

MakazhanAlpamys:
Timing note first: you pushed `da67aa0` at 17:48 and my review landed at 17:54, so you were fixing this before you saw it. **Your fix is correct — and it covers one of five.**

Verified on `da67aa0`:

```
rl_checkpoint  -> reaches its own validation (TypeError: config must be RLCheckpointConfig)   FIXED
echo_trap      -> *** NameError: name 'EchoTrapCallback' is not defined
reward_hacking -> bare global
minillm        -> bare global
hf_push        -> bare global
```

`globals().get(...)` with an `__getattr__` fallback does work — it goes through the module object, which is the thing PEP 562 hooks. A plain `from soup_cli.utils.X import XCallback` inside the builder does the same in one line instead of four, but that is style; yours is not wrong.

**The reason you fixed exactly one is the part worth knowing**, and it is not your mistake: CI reported a single failing test, so there appeared to be a single bug. Run the files in isolation and the PR produces **12 failures across 7 files** — eleven of them vanish in a full-suite run because some earlier file happens to fetch the attribute first, which caches it into `globals()` and makes the next module look fine.

```
$ pytest tests/test_code_review_high.py                             -> 1 failed, 28 passed
$ pytest tests/test_v0700_part_d.py tests/test_code_review_high.py  -> 62 passed
```

So the suite as it stands cannot show you the other four, and will not show the next one either. That is why the review asks for a subprocess-isolated control — fresh interpreter per module, import the builder, call it, assert no `NameError` — rather than just "fix these five". With that control in place, CI would have gone red on all five at once and you would have fixed them in one pass.

The ten `# noqa: F821` are still in place and still hide the diagnostic that names each of the four: removing two from `reward_hacking.py` gives F821 at `:592` and `:599`, which are the exact lines that fail.

Nothing else in the review changed — the mechanism is still right, #320 is still achieved on 11/11 modules, and #308 is still intact.

AchuthReddy-16:
Thanks for the detailed review. Before seeing this, I found the `build_rl_checkpoint_callback` failure in the full CI run, reproduced it locally, fixed that builder, and pushed `da67aa0`; the new CI run is for that partial fix.

Your review makes clear that the same cold-import issue affects the other four builders as well. I’ll update all five consistently using the self-import approach, remove the F821 suppressions, add the fresh-process regression coverage, and include the changelog entry before the next push.

MakazhanAlpamys:
Light check-in, no pressure — just so this does not go quiet on my side.

Nothing has changed here: the branch is still mergeable, and the plan you laid out (all five builders via the self-import shape, drop the ten `# noqa: F821`, fresh-process regression coverage, CHANGELOG entry) is still exactly what I would merge.

The one thing worth repeating because it is easy to lose: the regression test has to run the import in a **fresh process**. Eleven of the twelve failures vanish in a full-suite run, because an earlier test file happens to fetch the attribute first, caches it into `globals()`, and makes the next module look healthy. A test that only passes when it runs alone-but-after-nothing is the same trap that hid this bug from CI in the first place.

Take the time you need.

AchuthReddy-16:
Addressed the review feedback and rebased onto the latest `upstream/main`.

- Fixed all five cold-import builders using local self-imports
- Removed the related `F821` suppressions
- Added fresh-subprocess regression coverage for all five builders
- Added the CHANGELOG entry under `[Unreleased]`

Verification:
- 110 targeted tests passed locally
- All 12 CI checks passed across macOS, Ubuntu, and Windows

MakazhanAlpamys:
**Reviewed and approved on the code — I would have merged this ten minutes ago, and the only thing stopping me is a CHANGELOG conflict that #460 created underneath you.**

## What I verified, and one thing I got wrong first

I wrote a cold-import probe, ran it against your head — all five builders healthy — and then ran the **same probe against your previous head `da67aa0`**, where four of them were supposed to raise `NameError`. They didn't. Both looked fine.

That meant my probe could not fail, so it proved nothing. The cause: these builders are keyword-only, so my `f(None)` died on the signature before reaching the class reference. With valid arguments:

```
PRE-FIX   NameError: name 'EchoTrapCallback' is not defined
PR-HEAD   OK -> EchoTrapCallback | base: TrainerCallback
```

Now it discriminates, and the result means something. Worth saying out loud because it is the same class of mistake as the bug itself — a check that passes for a reason unrelated to the thing being checked.

All five builders reach their own validation on a cold import, each resolving to a **real `TrainerCallback`** base rather than the `object` fallback — so the #308 property holds too. HF dispatches every event with no `hasattr` guard, and a duck-typed callback dies on `on_epoch_begin`.

## The test that carries this PR

I re-broke `build_echo_trap_callback` by deleting the function-body import at line 463:

```
FAILED tests/test_issue320_lazy_callback_import.py::test_builder_invocation_no_nameerror_in_subprocess[soup_cli.utils.echo_trap-...]
1 failed, 37 passed
```

**Thirty-seven in-process tests pass under that mutant.** They fetch `m.EchoTrapCallback` as a module *attribute*, which `__getattr__` serves; the builder used a bare global, which it does not. That asymmetry is exactly what hid the original defect and made CI report one failure where there were twelve. Your subprocess test is the only thing standing between us and a repeat, and you added it because I described the trap rather than because anything failed.

*(A process note on my own run: my first mutation attempt matched **2** occurrences of that import line and my script aborted on its `count == 1` assertion. Had I not asserted, I would have run an unmutated tree and reported "survived" — a false finding. That guard earns its keep about once a week here.)*

Also confirmed: `# noqa: F821` count in `src/` is **0**. The new `F822` suppressions are a different thing and are correct — in a PEP 562 lazy module the name genuinely is not a global until first access, and unlike F821 it is not masking anything reachable, which the cold-import table above demonstrates.

Measured win, main vs your head:

| module | main | #455 |
|---|---|---|
| `utils.echo_trap` | 4.30s, torch loaded | **0.03s**, torch absent |
| `utils.reward_hack_control` | 3.71s, torch loaded | **0.05s**, torch absent |
| `utils.lisa` | 3.46s, torch loaded | **0.02s**, torch absent |

Zero test deletions. `ruff` clean. CI 12/12 on the true head SHA, checked against `headRefOid` rather than trusting the checks list.

## The one thing needed

`CHANGELOG.md` conflicts with `main` — #460 merged about fifteen minutes ago and touched the same `[Unreleased]` region. **Nothing in your source files conflicts**; the two PRs do not overlap in a single `src/` or `tests/` file. A rebase and a re-run and this goes in.

Sorry about the ordering — that is the cost of me merging the other one first, not anything on your side.

AchuthReddy-16:
Rebased again onto current `upstream/main` after #457 merged and preserved both CHANGELOG entries.

Post-rebase verification:
- Ruff passed
- 110 targeted tests passed

MakazhanAlpamys:
Merged as `b056aa8`. **Verified again after your rebase** rather than trusting that a rebase is a no-op — the cold-import probe and the full guard suite both re-run clean on `105fa40`:

```
cold-import probe: OK -> EchoTrapCallback | base: TrainerCallback
72 passed
```

## What this PR actually is

The measured win is the headline — `soup_cli.utils.echo_trap` 4.30s -> 0.03s, `reward_hack_control` 3.71s -> 0.05s, `lisa` 3.46s -> 0.02s, torch absent from `sys.modules` after import — but it is not the part I will remember.

**The part I will remember is the defect your first attempt introduced, and how it was found.** Deferring the classes while the builders still referenced bare globals produced `NameError` on a cold import in four of five builders, and **CI reported exactly one failure**. Eleven of the twelve vanish in a full-suite run because an earlier test file fetches the attribute first, caches it into `globals()`, and the next module looks healthy. You had no way to see that from a green-but-one CI run, and you said as much rather than guessing.

Your subprocess test is what closes it, and I measured how load-bearing it is by re-breaking `build_echo_trap_callback`:

```
1 failed, 37 passed
FAILED test_builder_invocation_no_nameerror_in_subprocess[soup_cli.utils.echo_trap-...]
```

Thirty-seven in-process tests pass under that mutant. They read `m.EchoTrapCallback` as a module *attribute*, which `__getattr__` serves; the builder used a bare global, which it does not. One test stands between us and a repeat.

## A mistake of mine, since it is instructive

My first cold-import probe reported all five builders healthy — and then reported the **pre-fix** head healthy too. A probe that cannot fail proves nothing. The cause: these builders are keyword-only, so my `f(None)` died on the signature before reaching the class reference. With valid arguments it discriminates:

```
PRE-FIX   NameError: name 'EchoTrapCallback' is not defined
PR-HEAD   OK -> EchoTrapCallback | base: TrainerCallback
```

Same class of error as the bug itself — something passing for a reason unrelated to what is being checked.

Also confirmed: `# noqa: F821` count in `src/` is **0**; the new `F822` suppressions are correct and different in kind (in a PEP 562 module the name genuinely is not a global until first access, and unlike F821 it masks nothing reachable). The nine deleted test lines are a docstring describing the *old* behaviour — no assertion touched. Each builder still resolves to a real `TrainerCallback`, so #308 holds.

Third merge, and two rebases you did without complaint while three other PRs landed on top of you. Credited in `CONTRIBUTORS.md` (`eb9b488`).

Issue #456, opened 2026-08-19
cut_ce matches the architecture against the whole Windows path: rsplit("/") never splits on a backslash, so the documented 'last path component only' guarantee does not hold


MakazhanAlpamys:
Found while reviewing #446, verified there, and left out of that PR deliberately — @AmirF194 declined to widen the scope and asked me to file it separately, which was the right call.

`utils/cut_ce.py` reduces the model id to its last path component before matching an architecture keyword:

```python
last_component = model_name.rsplit("/", 1)[-1]
```

`rsplit("/")` never splits on a backslash, so on Windows the **whole path** is matched, not its last component. Measured during the #446 review:

```
rsplit("/")[-1]     -> 'C:\Users\...\test_phi2_i...'   contains "phi2": True
os.path.basename    -> 'step-2000'                     contains "phi2": False
```

The consequence: any Windows user whose checkpoint path happens to contain an architecture keyword *anywhere* — a parent directory named `phi-2-experiments`, `llama-runs`, `qwen-sweeps` — gets that architecture matched for a checkpoint that may be something else entirely. It is the mirror of the defect #446 fixed: that one under-matched (a local directory never matched), this one over-matches.

**Why it is smaller than it sounds, stated so nobody over-prioritises it.** #446 made `AutoConfig.model_type` the primary path, so the name match is now only a fallback for when config resolution is unavailable — offline, or a directory with no `config.json`. It is reachable, not routine.

**Why it is worth fixing anyway.** `docs/performance-and-quantization.md` states the "last path component only" behaviour as a guarantee, and #446 restated it. A documented guarantee that does not hold on one platform is the shape this project keeps rediscovering.

It also had a real effect on the #446 review, which is the part worth recording: one of my mutations died on Windows **only because the pytest temp directory carried the test's own name**. On a clean parent path the mutant was behaviourally identical, so that mutation would have survived on the Linux and macOS cells. A test whose kill depends on the tmpdir path is not a test.

**Fix path.** Replace `rsplit("/", 1)[-1]` with something that handles both separators — `os.path.basename(model_name.replace("\\", "/"))`, or `PurePosixPath`/`PureWindowsPath` as appropriate. Check `utils/liger.py` for the same expression while you are there; it matches the whole lowercased name rather than the last component, so it may be differently wrong rather than identically wrong.

**Acceptance**
- [ ] A checkpoint directory whose PARENT path contains an architecture keyword, and whose own name does not, is not matched by that keyword.
[...]

PR #457, opened 2026-08-19
feat(tests): add doc-synced recipe count ratchet test (#453)

### LINKED ISSUE #453: Nothing guards the recipe count in the five documentation sites, so two recipe PRs landing in sequence leave the docs silently one low
Found while sequencing two recipe PRs (#448 and #452) that both bump the catalog count. Filing it as mine because the gap is in our guards, not in either PR.

The recipe count appears in **nine** places. Four are test assertions; five are documentation or comments:

```
src/soup_cli/recipes/catalog.py:51      (module docstring)
CONTRIBUTING.md:116
docs/commands.md:154
docs/serving-and-export.md:515
docs/serving-and-export.md:524
```

**No test guards any of the five.** `grep` over `tests/` for `CONTRIBUTING` or `docs/commands` returns nothing.

Normally that is a nit. What makes it worth an issue is what happens when two recipe PRs land in sequence. I merged #452 then #448 locally to see:

```
STEP 1: merge #452 into main  -> clean, count 145
STEP 2: merge #448 on top
  CONFLICT (content): tests/test_recipes.py
  CONFLICT (content): tests/test_v07124.py
  Auto-merging tests/test_v07130.py      <-- no conflict
  Auto-merging tests/test_v07132.py      <-- no conflict
  Auto-merging src/soup_cli/recipes/catalog.py
```

True count after both is **146**. The two conflicts are loud and get resolved. The two auto-merged *test* files silently keep `145` — but they assert on it, so CI catches them. The five documentation sites also silently keep `145`, and **nothing catches those at all**. They only conflict when a PR happens to edit two adjacent lines rather than one, which is luck rather than a property.

So the failure mode is: two recipe PRs land, tests go green, and the docs quietly claim a number that is one too low — which is precisely the count-drift this project keeps rediscovering (`CONTRIBUTING.md` was stale at 360 files / 17907 tests through the v0.73.3 release; `.claude/CLAUDE.md` has been stale more than once).

**Fix path.** One test that reads `len(RECIPES)` and asserts the same number appears at every documented site, with the site list in the test rather than in a comment. Something like:

```python
SITES = [
    ("src/soup_cli/recipes/catalog.py", r"(\d+) ready-made"),
    ("CONTRIBUTING.md",                 r"\((\d+) recipes\)"),
    ("docs/commands.md",                r"List all (\d+) ready-made"),
    ("docs/serving-and-export.md",      r"(\d+) ready-made recipes"),
]
```
Each site must match at least once, and every captured number must equal `len(RECIPES)`.

**Acceptance**
- [ ] Adding a recipe without updating any doc site fails the build, naming the file and the number it found.
- [ ] The test fails if a site's pattern stops matching at all — otherwise a reworded doc line silently drops out of coverage and the guard rots into a no-op.
- [ ] The four existing `test_catalog_size_is_N` assertions are either folded into it or left alone deliberately, with the decision written down.

Related, and the reason this is worth doing rather than filing under "someday": the same class of check already exists here for a different drifting number — `tests/test_requires_python_bound.py` derives the Python bound from the CI matrix so widening one without the other fails. This is that pattern, applied to the count.

harshitthek:
Resolves #453.

## Problem & Mechanism

When recipe additions merge in sequence, Git 3-way merge conflicts loudly on Python test files, but silently auto-merges Markdown documentation lines without conflict. If a PR updates only some of the documentation sites, the documentation quietly drifts out of date without failing CI.

The recipe count is stated across **5 documentation and module sites**:
1. `src/soup_cli/recipes/catalog.py:51` (`# Recipe catalog (145 recipes)`)
2. `CONTRIBUTING.md:116` (`(145 recipes)`)
3. `docs/commands.md:154` (`List all 145 ready-made recipes`)
4. `docs/serving-and-export.md:515` (`templates or 145 ready-made recipes`)
5. `docs/serving-and-export.md:524` (`recipe dropdown (145 recipes)`)

## What Changed

Added a dedicated repo-wide ratchet test module: `tests/test_recipe_count_is_synced.py` following the `test_requires_python_bound.py` design idiom.

1. **Dynamic Catalog Derivation:**
   - Reads `len(RECIPES)` directly and matches all 5 declared sites against the single source of truth.
   - Raises an informative `AssertionError` naming the exact file, pattern, and mismatch if any site is stale.

2. **Anti-Rot Pattern Match Assertions (Criterion 2):**
   - Asserts `len(matches) > 0` for every declared pattern. If a doc line is reworded or deleted in the future, the test fails loudly rather than silently decaying into a passing no-op.

3. **Mutation Resistance & Controls (`TestTheGuardHasTeeth`):**
   - Verified against synthetic stale counts (e.g. 144 vs 145) $\to$ fails deterministically naming the file.
   - Verified against reworded/missing patterns $\to$ fails deterministically.
   - Verified against mutated catalog size (`len(RECIPES) +/- 1`) $\to$ fails all 5 sites deterministically.

4. **Existing Test Invariants (Criterion 3):**
   - The four existing release-checkpoint tests (`test_recipes.py`, `test_v07124.py`, `test_v07130.py`, `test_v07132.py`) are deliberately left alone as Python dictionary milestone checkpoints, with the architectural decision documented in the module docstring.

5. **House Standards:**
   - Bumped `CONTRIBUTING.md` test file count (370 $\to$ 371 files).
   - Added entry to `CHANGELOG.md` under `[Unreleased] Added` referencing #456.

## Verification

- `pytest tests/test_recipe_count_is_synced.py -v` $\to$ 6/6 tests green (100% pass).
- Intentional mutations tested:
  - Stale count $\to$ KILLED
  - Missing pattern $\to$ KILLED
  - Mutated catalog size $\to$ KILLED across all 5 sites
- `ruff check src/ tests/ docs/` clean (0 errors / 0 warnings).
~Harshit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,9 @@ reproducing 70+ versions of notes.
   and the `--live` overlay renderer emit the real N-dataset mixture again instead of
   collapsing to one path.
 
+- **A repo-wide documentation ratchet to guarantee declared recipe counts stay synchronized with the catalog (#453 by @harshitthek in #457).**
+  Derives the expected count dynamically from `len(RECIPES)` and scans all declared documentation sites, preventing silent Git auto-merge drift across sequential recipe additions.
+
 ### Fixed
 
 - **`cut_ce.py` and `liger.py` now normalize path separators and match architecture
--- a/tests/test_recipe_count_is_synced.py
+++ b/tests/test_recipe_count_is_synced.py
@@ -1,0 +1,285 @@
+"""Repo-wide ratchet: recipe count in documentation must match the catalog.
+
+When new recipes land in the catalog, the recipe count is stated across multiple
+documentation and module docstring sites. When two recipe PRs merge sequentially,
+Git's 3-way merge conflicts loudly on Python test files, but silently auto-merges
+Markdown documentation lines, leaving documentation counts quietly out of date.
+
+This test derives the expected count dynamically from `len(RECIPES)` and scans every
+declared documentation site in `DOC_SITES` to guarantee that documentation and code never drift.
+
+Note on existing test assertions:
+The existing `test_catalog_size_is_N` assertions in `test_recipes.py`,
+`test_v07124.py`, `test_v07130.py`, and `test_v07132.py` are deliberately left
+alone as release milestone checkpoints for the Python dictionary. This file
+specifically guards documentation and docstring synchronization against silent
+Git auto-merge drift (#453).
+"""
+
+from __future__ import annotations
+
+import pathlib
+import re
+from dataclasses import dataclass
+from typing import Sequence
+
+from soup_cli.recipes.catalog import RECIPES
+
+ROOT = pathlib.Path(__file__).resolve().parents[1]
+
+
+@dataclass(frozen=True)
+class DocCountSite:
+    """A target documentation site declaring a recipe count."""
+
+    rel_path: str
+    pattern: str
+
+
+@dataclass(frozen=True)
+class DocCountMatch:
+    """A single matched recipe count found at a specific file and line."""
+
+    rel_path: str
+    lineno: int
+    pattern: str
+    count: int
+
+
+@dataclass(frozen=True)
+class SyncVerdict:
+    """The result of auditing recipe counts across documentation sites."""
+
+    expected_count: int
+    matches: tuple[DocCountMatch, ...]
+    mismatches: tuple[DocCountMatch, ...]
+    missing_patterns: tuple[DocCountSite, ...]
+
+    @property
+    def is_synced(self) -> bool:
+        """True if all declared patterns matched and every count equals expected."""
+        return len(self.mismatches) == 0 and len(self.missing_patterns) == 0
+
+    def format_diagnostic(self) -> str:
+        """Generate a human-readable diagnostic report for CI failure messages."""
+        lines: list[str] = []
+        if self.missing_patterns:
+            lines.append("Missing pattern matches (reworded or deleted doc lines):")
+            for site in self.missing_patterns:
+                lines.append(f"  * {site.rel_path} with pattern {site.pattern!r}")
+        if self.mismatches:
+            lines.append(
+                f"Out-of-sync recipe counts (catalog has {self.expected_count} recipes):"
+            )
+            for m in self.mismatches:
+                lines.append(
+                    f"  * {m.rel_path}:{m.lineno} declares {m.count} (pattern: {m.pattern!r})"
+                )
+        return "\n".join(lines)
+
+
+#: Declared documentation and module sites that state the recipe count.
+DOC_SITES: tuple[DocCountSite, ...] = (
+    DocCountSite(
+        "src/soup_cli/recipes/catalog.py", r"#\s*Recipe catalog\s*\((\d+)\s*recipes\)"
+    ),
+    DocCountSite("CONTRIBUTING.md", r"\((\d+)\s+recipes\)"),
+    DocCountSite(
+        "docs/commands.md", r"List all\s+(\d+)\s+(?:ready[-\s]?made\s+)?recipes"
+    ),
+    DocCountSite(
+        "docs/serving-and-export.md",
+        r"templates or\s+(\d+)\s+(?:ready[-\s]?made\s+)?recipes",
+    ),
+    DocCountSite(
+        "docs/serving-and-export.md", r"recipe dropdown\s+\((\d+)\s+recipes\)"
+    ),
+)
+
+#: Mandatory baseline roster pairs that must never drop out of DOC_SITES discovery.
+MANDATORY_SITE_PAIRS: frozenset[tuple[str, str]] = frozenset(
+    (
+        ("src/soup_cli/recipes/catalog.py", r"#\s*Recipe catalog\s*\((\d+)\s*recipes\)"),
+        ("CONTRIBUTING.md", r"\((\d+)\s+recipes\)"),
+        ("docs/commands.md", r"List all\s+(\d+)\s+(?:ready[-\s]?made\s+)?recipes"),
+        ("docs/serving-and-export.md", r"templates or\s+(\d+)\s+(?:ready[-\s]?made\s+)?recipes"),
+        ("docs/serving-and-export.md", r"recipe dropdown\s+\((\d+)\s+recipes\)"),
+    )
+)
+
+
+def audit_doc_recipe_counts(
+    root: pathlib.Path = ROOT,
+    sites: Sequence[DocCountSite] = DOC_SITES,
+    expected_count: int | None = None,
+) -> SyncVerdict:
+    """Pure auditor: inspect documentation sites without assertion side-effects.
+
+    Returns an immutable `SyncVerdict` containing all matches, mismatches, and
+    missing patterns.
+    """
+    if expected_count is None:
+        expected_count = len(RECIPES)
+
+    matches: list[DocCountMatch] = []
+    mismatches: list[DocCountMatch] = []
+    missing_patterns: list[DocCountSite] = []
+
+    for site in sites:
+        file_path = root / site.rel_path
+        if not file_path.exists():
+            missing_patterns.append(site)
+            continue
+        text = file_path.read_text(encoding="utf-8")
+        found_matches = list(re.finditer(site.pattern, text, flags=re.IGNORECASE))
+        if not found_matches:
+            missing_patterns.append(site)
+            continue
+        for match in found_matches:
+            lineno = text.count("\n", 0, match.start()) + 1
+            count = int(match.group(1))
+            entry = DocCountMatch(
+                rel_path=site.rel_path,
+                lineno=lineno,
+                pattern=site.pattern,
+                count=count,
+            )
+            matches.append(entry)
+            if count != expected_count:
+                mismatches.append(entry)
+
+    return SyncVerdict(
+        expected_count=expected_count,
+        matches=tuple(matches),
+        mismatches=tuple(mismatches),
+        missing_patterns=tuple(missing_patterns),
+    )
+
+
+class TestRecipeCountIsSynchronised:
+    """Every declared documentation site must match the true catalog size."""
+
+    def test_every_documentation_site_matches_catalog_size(self) -> None:
+        verdict = audit_doc_recipe_counts(ROOT, DOC_SITES, len(RECIPES))
+        assert verdict.is_synced, (
+            f"Recipe count in documentation is out of sync:\n{verdict.format_diagnostic()}\n"
+            "Update the documentation sites to match the catalog count."
+        )
+        assert len(verdict.matches) >= len(DOC_SITES)
+
+    def test_all_declared_sites_are_covered(self) -> None:
+        """Control: ensure no declared site or pattern drops out of discovery."""
+        current_pairs = {(s.rel_path, s.pattern) for s in DOC_SITES}
+        missing_pairs = MANDATORY_SITE_PAIRS - current_pairs
+        assert not missing_pairs, (
+            f"Mandatory documentation pattern dropped from DOC_SITES: {missing_pairs}"
+        )
+
+        verdict = audit_doc_recipe_counts(ROOT, DOC_SITES, len(RECIPES))
+        assert verdict.missing_patterns == ()
+        assert len(verdict.matches) >= len(DOC_SITES)
+
+    def test_catalog_size_matches_actual_recipes_dictionary(self) -> None:
+        """Control: ensure the catalog dictionary itself is non-empty."""
+        assert len(RECIPES) > 0
+
+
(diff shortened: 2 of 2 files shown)


MakazhanAlpamys:
Reviewed in an isolated worktree with `PYTHONPATH` pinned to it. **Both acceptance criteria are met, including the second one — which is the one I was worried about.** One blocker, and it is the failure mode this issue was written to avoid.

## Criterion 2 is genuinely met, at every site

This is the half that separates a ratchet from a guard that quietly stops guarding. I reworded each doc line so **no number remains at all** — the shape where a naive "every number I find equals N" check passes vacuously because it finds nothing:

```
Missing pattern matches (reworded or deleted doc lines):
  * docs/commands.md with pattern 'List all\s+(\d+)\s+ready-made recipes'
```

All five fail, each naming the file and the pattern that stopped matching. And it is load-bearing rather than incidental: dropping the `missing_patterns` term from `is_synced` kills two tests, so the mechanism is real.

**Criterion 1** likewise fires at all five sites individually, and the message names file, line, number found and number expected. The site list is complete — a repo-wide grep for any count near "recipe" in tracked non-test files returns exactly those five.

**Criterion 3** — the four existing `test_catalog_size_is_N` assertions were left deliberately, with the decision written into the module docstring, which is where someone hitting the guard will look.

**Five mutations, no survivors:** emptying `DOC_SITES`, dropping one entry, `count != count`, pointing a site at a nonexistent file, and removing the `missing_patterns` term. Each killed by name.

## Blocker: the guard fires on correct code

`assert len(verdict.matches) == 5` at lines 148 and 153. I added a **second, correctly-numbered** mention to `CONTRIBUTING.md` — a perfectly legitimate doc edit, every count right:

```
116:  recipes/            - Ready-made configs for popular models (145 recipes)
117:  (see the recipe catalog (145 recipes) for the full list)

3 failed, 4 passed
FAILED ...::test_every_documentation_site_matches_catalog_size
FAILED ...::test_all_five_declared_sites_are_covered
FAILED ...::test_mutated_catalog_count_fails_all_sites
```

`mismatches=()` and `missing_patterns=()` — nothing is wrong, and CI is red. The same assertion also breaks the moment anyone adds a sixth site to `DOC_SITES`, which is the normal way this guard grows.

I ran that myself rather than take it from my reviewer, because I am asking you to change working code and that deserves a reproduction.

**Why this specific thing is the blocker and not a nit:** #453's own text says the failure mode to avoid is a guard that rots into a no-op. A guard that reddens the build on a correct edit does not rot — it gets **deleted**, which is worse, because the deletion looks justified at the time. I have written that sentence into three reviews this week; here it is with a reproduction attached.

Fix: replace the three `== 5` with

```python
assert verdict.missing_patterns == ()
assert len(verdict.matches) >= len(DOC_SITES)
```

That keeps every property you were asserting — every declared site was found — without forbidding a second correct mention or a sixth site.

## Three nits

**The CHANGELOG entry contradicts precedent.** Both test-only ratchets in this repo shipped without one: `2c09b6d` (the ANSI guard) touched one file, `008da45` (the licence guard) touched `CONTRIBUTING.md` and its test. Neither touched `CHANGELOG.md`, because nothing user-facing changes and it does not ship to PyPI. Your entry is correctly attributed and correctly placed if we keep it — I lean toward dropping it, but it is a coin-flip and I will not block on it.

**The test count you edited is itself wrong, which is a good joke at both our expense.** You bumped `18151 → 18157`, `+6`, for a file that adds **7** tests. And the absolute number is stale repo-wide anyway — a real collection on your head gives `18182`. In a PR about count drift, the hand-maintained count drifted. Your new guard does not cover that line; nothing does. Worth a follow-up issue rather than widening this PR.

**Minor:** the file carries a UTF-8 BOM — the only one of 372 test files. And you are four commits behind `main`, so your description quotes 145 while the catalog is now 146; I verified the merge is clean and the guard passes at 146, so it is cosmetic, but the rebase will make the description read true.

## What this is worth

The count-drift gap was found because two recipe PRs landed in sequence and the five doc sites auto-merged silently while only the test files conflicted. Your guard closes exactly that, and the criterion-2 half means it will still be closing it after someone rewords a doc line — which is the part almost every guard of this kind gets wrong.

One assertion change and this merges.

harshitthek:
Hi @MakazhanAlpamys,

Thank you for the thorough review and for catching the duplicate-mention false alarm on line length.

All feedback points and hardening improvements have been addressed in `5487ffa`:

1. **Fixed Rigid Length Assertions (The Blocker):**
   Replaced rigid `== 5` with `assert verdict.missing_patterns == ()` and `assert len(verdict.matches) >= len(DOC_SITES)`. This ensures every declared site is matched at least once while accepting multiple valid count mentions per file and supporting future site additions to `DOC_SITES`.

2. **Added Makazhan's Duplicate-Mention Regression Test:**
   Added `test_duplicate_correct_mentions_in_single_file_are_accepted` in `TestTheGuardHasTeeth` to lock in the reproduction case and guarantee multiple valid mentions never fail CI.

3. **Added Mixed-Count Collision Test:**
   Added `test_mixed_counts_in_single_file_flags_stale_line` proving that if a file contains 1 valid and 1 stale count, the stale line is isolated and flagged at its exact line number.

4. **Future-Proofing & Hygiene:**
   - Renamed test method to `test_all_declared_sites_are_covered` (removing the hardcoded "five").
   - Added regex hyphen tolerance (`ready[-\s]?made`) and `re.IGNORECASE` to handle natural markdown formatting tweaks safely.
   - Stripped the UTF-8 BOM (`\xef\xbb\xbf`), making the file pure POSIX UTF-8.
   - Synchronized test file counts in `CONTRIBUTING.md` (371 files).
   - Rebased cleanly on `main` at 146 recipes.

The suite is green (`9 passed in 0.34s`) and `ruff check` passes with 0 warnings.
Lemme know if you need any improvements/changes !!

~Harshit

MakazhanAlpamys:
Re-reviewed at `5487ffa`, isolated worktree, `PYTHONPATH` pinned to it. **The blocker is closed, and I checked it in both directions rather than just confirming green.**

| | result |
|---|---|
| my reproduction (a second, *correctly*-numbered mention in `CONTRIBUTING.md`) on the old commit | **3 failed, 4 passed** — the finding was real |
| the same reproduction on `5487ffa` | **9 passed** |
| a sixth `DOC_SITES` entry + its doc file | **9 passed** |

The pre-fix red run matters as much as the green one: it proves my reproduction was faithful and not a mistake in how I invoked it. The BOM is gone too (`b'\xef\xbb\xbf'` -> `b'"""'` on the committed blob).

## One finding, and it is the control test's own docstring

Dropping a whole entry from `DOC_SITES` — I deleted the `docs/commands.md` one, three lines — **survives**:

```
9 passed in 0.35s
```

The guard silently stops covering that file and nothing anywhere goes red. Per the house rule I only call this a finding if I can name the control that should have caught it, and here the test names itself:

```python
def test_all_declared_sites_are_covered(self) -> None:
    """Control: ensure no declared site drops out of discovery."""
```

A declared site dropped out of discovery and it passed. The docstring states a property the test does not have — which is the one shape I care about most, because the next reader trusts it and stops looking.

This is a **capability you traded away to fix my blocker**, not something you broke: the old `== 5` did catch it (4 sites -> 4 matches -> fails). I asked for the change, so the trade is mine as much as yours. It is worth recovering because one line does it, above the existing asserts:

```python
assert {s.rel_path for s in DOC_SITES} >= {
    "src/soup_cli/recipes/catalog.py", "CONTRIBUTING.md",
    "docs/commands.md", "docs/serving-and-export.md",
}
```

That pins the roster without forbidding a sixth site or a second correct mention — so it does not re-introduce the thing we just removed. It also closes a narrower gap: with `DOC_SITES = ()` the two `TestRecipeCountIsSynchronised` tests pass *vacuously* (`0 >= 0`) and only the teeth test fires.

**Four of five mutations still kill** — catalog count 146->147 (a real appended `RecipeMeta`, `len(RECIPES)` verified at 147), a reworded doc line, empty `DOC_SITES`, and a site pointed at a nonexistent file. The guard has teeth; this is one hole in it.

## The two nits from last round

**The test-count line is still wrong, and I am no longer sure asking you to fix it is right.** Measured on your head: **372 files / 18202 tests**; you state 371 / 18157. But your branch predates #458, which landed 20 minutes ago and adds tests to an existing file — so any number you hardcode now is wrong again after the rebase. That is the joke completing itself: a PR about count drift, on a treadmill of hand-maintained counts. Two options, your call:

- write the real number at the moment you rebase and accept it may drift again, or
- **revert the count edits out of this PR entirely** and let the follow-up issue cover them properly. I lean here — your guard covers recipe counts, and the thing that would actually fix *these* numbers is the same guard pointed at the test suite.

**CHANGELOG: keeping it.** I said last round it was a coin-flip and I would not block; holding to that. It is at line 27, inside `[Unreleased]`, correctly attributed `(#453 by @harshitthek in #457)`. It will conflict on the rebase over #458 — mechanical, `### Fixed` moved underneath you.

**Cosmetic:** the PR body still quotes 145 in all five bullets; the catalog is 146.

## Where this stands

`gh pr checks` against the true head: 8 pass, **windows x3 still pending**. Given this repo's Windows history — CRLF, the BOM you just removed, the 0xc000001d fleet crash — those three are the ones I want to see.

One line for the roster assert, a decision on the count line, a rebase over #458, and green Windows. Then it merges.

*(Note for the record: my first pass read a cached `pull/457/head` and I was looking at your pre-force-push commit. `git fetch --force` corrected it. Everything above is `5487ffa`.)*

harshitthek:
Hi @MakazhanAlpamys,

Thank you for the thoughtful review and for catching the roster gap when an entry is dropped from `DOC_SITES`.

All feedback points and hardening improvements have been completed in `a9587b8`:

1. **Closed the Roster Gap with Pattern-Level Pairs:**
   Added `MANDATORY_SITE_PAIRS: frozenset[tuple[str, str]]` enforcing that all 5 declared `(rel_path, pattern)` pairs must remain present in `DOC_SITES`. We track `(file, pattern)` tuples rather than bare file paths so that dropping individual patterns in multi-pattern files (such as the second pattern in `docs/serving-and-export.md`) fails deterministically, while still allowing future documentation sites to be added via superset check.

2. **Added Dropped-Site Control Test:**
   Added `test_dropped_site_from_doc_sites_fails_roster_check` in `TestTheGuardHasTeeth` to verify that dropping any declared site/pattern from discovery immediately fails with an explicit diagnostic naming the missing pair.

3. **Atomic Scoping on `CONTRIBUTING.md`:**
   Reverted the hand-maintained test-count edits from this PR as recommended, keeping the scope strictly focused on recipe catalog synchronization.

4. **Rebased over PR #458:**
   Cleanly rebased on current `main` (`44d90c1`).

The full suite is green (`10 passed in 2.42s`) and `ruff check` passes cleanly.
~Harshit

MakazhanAlpamys:
Merged as `6c64a38`. **The finding is closed and I verified both halves myself** — not just that it now passes, but that closing it did not trade the earlier blocker back.

```
deleted the docs/commands.md entry from DOC_SITES  ->  1 failed
    test_all_declared_sites_are_covered
added a SECOND correctly-numbered mention           ->  10 passed
```

That second line is the one I care about most. Fixing a finding by re-introducing the thing you fixed last round is a real failure mode, and here both properties hold at once: a dropped site fails, a legitimate second mention does not.

**The docstring is now true.** It said *"Control: ensure no declared site drops out of discovery"* while a dropped site passed 9/9. `MANDATORY_SITE_PAIRS` gives it that property — and you pinned the **pattern** as well as the path, which I did not ask for and which is the better call: a site kept in the roster but silently repointed at a pattern that matches nothing would have slipped past a path-only roster.

All five control mutations still kill — catalog count 146→147, a reworded doc line, empty `DOC_SITES`, a site pointed at a nonexistent file. And you added a seventh test exercising the roster logic itself, so the guard has a guard.

**Two judgement calls you made that I want on the record.**

You reverted your own `CONTRIBUTING.md` count edits back out. That is the harder direction to move in — removing work you had already done because it turned out to belong to a different mechanism — and it was right: I measured it afterwards, and **11 of the last 20 commits on `main` touch `tests/`**, so a ratchet on those numbers would redden the build on more than half of all merges. Filing #465 instead of widening this PR is what kept that from becoming the same defect one file to the left.

And the CHANGELOG coin-flip resolved in favour of keeping it. Fine by me — I said I would not block either way, and it is correctly placed at line 61 under `[Unreleased]` with the right suffix.

**Diff:** two files, `+288/-0`, zero deleted lines repo-wide, no source file touched. I check that on every PR because this repo has seen ones that quietly removed coverage; yours removed nothing.

One note for whenever you next touch the file, not worth a round: `MANDATORY_SITE_PAIRS` is a second hand-maintained copy of `DOC_SITES` — structurally the "two copies, one maintained" shape that caused #372. Here it is the *acceptable* form of it, because divergence fails loudly rather than silently, which is the whole point of a ratchet floor. Worth a sentence in the module docstring saying so, so the next reader does not try to DRY it away.

Fifth merge. Credited in `CONTRIBUTORS.md` (`bcebfd6`).

PR #458, opened 2026-08-19
fix(cce,liger): normalize path separators to prevent parent-directory…

### LINKED ISSUE #456: cut_ce matches the architecture against the whole Windows path: rsplit("/") never splits on a backslash, so the documented 'last path component only' guarantee does not hold

harshitthek:
Resolves #456.

## 1. Problem & Root Cause Mechanism

In `utils/cut_ce.py` (line 101), fallback architecture detection used:
```python
last_component = model_name.rsplit("/", 1)[-1].lower()
```
`rsplit("/")` never splits on Windows backslashes (`\`), so on Windows the **entire path** was matched instead of the last component. 

Any Windows user whose checkpoint resided inside a directory containing an architecture keyword anywhere in its parent path (e.g. `C:\experiments\phi-2-runs\step-2000` or `C:\models\llama-sweeps\checkpoint-100`) had that architecture falsely matched when config resolution was unavailable (offline or missing `config.json`).

Similarly, `utils/liger.py` passed `model_name.lower()` to `_apply_liger_manual`, which matched architecture keywords across the entire path string on all operating systems. In `liger.py`, `"llama"` was checked before `"phi"` — so a Phi model located in `C:\llama_experiments\phi_model` would erroneously have the **Llama kernel applied to a Phi model**.

---

## 2. Additional Flaw Uncovered During Audit: The Trailing-Slash Trap

While auditing the suggested `os.path.basename(model_name.replace("\\", "/"))` fix path, we caught a secondary edge-case defect:
* Calling `os.path.basename()` on any path with a trailing separator (e.g. `C:\experiments\llama-3.1\` or `/models/phi-3/`) returns an **empty string `""`**.
* An empty string has no keyword, so passing a path with a trailing slash would silently fail architecture fallback matching and report unsupported.

---

## 3. What Changed

1. **Dual-Separator & Trailing-Slash Normalization in `cut_ce.py` and `liger.py`:**
   - Both modules now normalize dual separators and strip trailing slashes deterministically:
     ```python
     clean_name = model_name.replace("\\", "/").rstrip("/")
     last_component = clean_name.rsplit("/", 1)[-1].lower() if clean_name else ""
     ```
   - Restores the documented "last path component only" guarantee across POSIX and Windows while being fully resilient to trailing slashes and degenerate inputs.

2. **Harmonized `_apply_liger_manual` in `liger.py` (Criterion 4):**
   - `_apply_liger_manual` now matches against `last_component` instead of the whole path string, eliminating the parent-directory cross-architecture patching bug.

3. **Deterministic Non-Tmpdir Testing (Criteria 1–3):**
   - Added `TestIssue456CrossPlatformPathComponentExtraction` in `tests/test_training_speed.py` with 7 comprehensive test cases:
     - **Windows Parent Directory Isolation (Negative Control):** `r"C:\experiments\phi-2-runs\step-2000"` -> `False` (not dispatched).
     - **POSIX Parent Directory Isolation (Negative Control):** `"/tmp/phi-2-experiments/step-2000"` -> `False` (not dispatched).
     - **Trailing Separator Resilience:** `C:\experiments\runs\llama-3.1-8b\` and `/tmp/models/phi-3-mini/` -> `True` (correctly dispatched).
     - **Liger Parent Directory Isolation:** `r"C:\experiments\llama_experiments\step-2000"` -> `False`.
     - **Liger Trailing Separator Matching:** `C:\models\llama-3.1-8b\` -> `True`.
     - **Degenerate Boundary Inputs:** Empty strings (`""`), root slashes (`"/"`, `"\\"`), and multiple slashes (`"///"`, `"\\\\\\"`) -> `False` without raising exceptions.
     - **Repeated Separator Normalization:** `C:\\\\experiments\\\\runs\\\\llama-3.1-8b\\\\` -> `True`.
   - Tests explicitly use constructed synthetic paths rather than `tmp_path`, so outcomes cannot be masked or affected by runner temporary folder names.

4. **House Standards:**
   - Added entry to `CHANGELOG.md` under `[Unreleased]` -> `### Fixed` referencing #456.

---

## 4. Verification & Mutation Kill Results

- `pytest tests/test_training_speed.py -k TestIssue456` -> **7/7 tests passed (100% green in 0.49s)**.
- Intentional mutations tested:
  - Reverting to naive `rsplit("/")` -> **KILLED** on Windows parent path test.
  - Reverting to whole-path match in `liger.py` -> **KILLED** on Liger parent path test.
  - Naive `basename` without `rstrip("/")` -> **KILLED** on trailing separator test.
- Repo policy test suites (`test_requires_python_bound.py`, `test_no_foreign_license_headers.py`, `test_cli_help_assertions_are_ansi_safe.py`) -> **27/27 passed**.
- `ruff check src/ tests/ docs/` -> **0 errors / 0 warnings**.
~Harshit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,17 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **`cut_ce.py` and `liger.py` now normalize path separators and match architecture
+  keywords on the last path component only (#456 by @harshitthek in #458).**
+  On Windows, `rsplit("/")` never split on backslashes, causing parent directory
+  names (e.g. `C:\experiments\phi-3-runs\step-2000`) to over-match architecture
+  keywords during fallback detection when config resolution was unavailable. In
+  `liger.py`, switching from whole-path to last-component matching also drops
+  org-prefix false positives (e.g. `mistralai/*`, `Qwen/*` when the model name
+  does not contain the keyword) and prevents parent directories from applying the
+  wrong fused kernel across POSIX and Windows. Both modules now normalize dual
+  separators and strip trailing slashes deterministically.
+
 - **`measure_gemm_tflops` now records per-repeat samples and
   `test_takes_the_best_repeat_not_the_first` verifies best-of-N selection within
   a single measurement (#444 by @harshitthek in #451).** Comparing two separate
--- a/src/soup_cli/utils/cut_ce.py
+++ b/src/soup_cli/utils/cut_ce.py
@@ -97,8 +97,9 @@ def apply_cut_ce(model_name: str) -> bool:
 
     # Fallback for when config resolution has nothing to read from. Match on
     # the last path component only, to keep an upstream-org / parent-dir
-    # name from leaking into architecture selection.
-    last_component = model_name.rsplit("/", 1)[-1].lower()
+    # name from leaking into architecture selection (#456).
+    clean_name = str(model_name or "").strip().replace("\\", "/").rstrip("/")
+    last_component = clean_name.rsplit("/", 1)[-1].lower() if clean_name else ""
     detectors = (
         (("codellama",), "llama"),
         (("llama",), "llama"),
--- a/src/soup_cli/utils/liger.py
+++ b/src/soup_cli/utils/liger.py
@@ -51,7 +51,8 @@ def apply_liger_kernel(model_name: str) -> bool:
     if not check_liger_available():
         return False
 
-    model_lower = model_name.lower()
+    clean_name = str(model_name or "").strip().replace("\\", "/").rstrip("/")
+    last_component = clean_name.rsplit("/", 1)[-1].lower() if clean_name else ""
 
     # Detect the architecture from the CONFIG, not from the path.
     #
@@ -79,7 +80,7 @@ def apply_liger_kernel(model_name: str) -> bool:
             # success we cannot back.
             pass
 
-    return _apply_liger_manual(model_lower)
+    return _apply_liger_manual(last_component)
 
 
 def _detect_model_type(model_name: str) -> str:
@@ -97,30 +98,30 @@ def _detect_model_type(model_name: str) -> str:
     return str(getattr(config, "model_type", "") or "")
 
 
-def _apply_liger_manual(model_lower: str) -> bool:
-    """Manually apply Liger Kernel patches for known model architectures."""
+def _apply_liger_manual(last_component: str) -> bool:
+    """Manually apply Liger Kernel patches matching the last path component (#456)."""
     try:
-        if "llama" in model_lower or "codellama" in model_lower:
+        if "llama" in last_component or "codellama" in last_component:
             from liger_kernel.transformers import apply_liger_kernel_to_llama
 
             apply_liger_kernel_to_llama()
             return True
-        elif "mistral" in model_lower or "mixtral" in model_lower:
+        elif "mistral" in last_component or "mixtral" in last_component:
             from liger_kernel.transformers import apply_liger_kernel_to_mistral
 
             apply_liger_kernel_to_mistral()
             return True
-        elif "gemma" in model_lower:
+        elif "gemma" in last_component:
             from liger_kernel.transformers import apply_liger_kernel_to_gemma2
 
             apply_liger_kernel_to_gemma2()
             return True
-        elif "qwen" in model_lower:
+        elif "qwen" in last_component:
             from liger_kernel.transformers import apply_liger_kernel_to_qwen2
 
             apply_liger_kernel_to_qwen2()
             return True
-        elif "phi" in model_lower:
+        elif "phi" in last_component:
             from liger_kernel.transformers import apply_liger_kernel_to_phi3
 
             apply_liger_kernel_to_phi3()
--- a/tests/test_training_speed.py
+++ b/tests/test_training_speed.py
@@ -305,6 +305,208 @@ class TestCutCEDetectsArchitectureFromTheConfig:
         assert calls == ["llama"], calls
 
 
+class TestIssue456CrossPlatformPathComponentExtraction:
+    r"""#456: cut_ce and liger must match architecture on the last component only.
+
+    On Windows, naive rsplit("/") never splits on backslashes, so parent directory
+    names (e.g. C:\experiments\phi-2-runs\step-2000) over-matched architecture
+    keywords when config resolution was unavailable. Tests use explicit synthetic
+    paths rather than tmp_path to prevent runner directory names from leaking.
+    """
+
+    def _fake_cce(
+        self,
+        monkeypatch,
+        calls,
+        supported=("llama", "phi3", "gemma2", "mistral", "qwen2"),
+    ):
+        import sys
+        import types
+
+        def _patch(model_type_or_model, *args, **kwargs):
+            calls.append(model_type_or_model)
+            if model_type_or_model not in supported:
+                raise RuntimeError(f"Unknown model type {model_type_or_model}")
+
+        mod = types.ModuleType("cut_cross_entropy.transformers")
+        mod.cce_patch = _patch
+        parent = types.ModuleType("cut_cross_entropy")
+        parent.transformers = mod
+        monkeypatch.setitem(sys.modules, "cut_cross_entropy", parent)
+        monkeypatch.setitem(sys.modules, "cut_cross_entropy.transformers", mod)
+        return calls
+
+    def _fake_liger(self, monkeypatch, calls):
+        import sys
+        import types
+
+        mod = types.ModuleType("liger_kernel.transformers")
+        mod.apply_liger_kernel_to_llama = lambda: calls.append("llama")
+        mod.apply_liger_kernel_to_mistral = lambda: calls.append("mistral")
+        mod.apply_liger_kernel_to_gemma2 = lambda: calls.append("gemma2")
+        mod.apply_liger_kernel_to_qwen2 = lambda: calls.append("qwen2")
+        mod.apply_liger_kernel_to_phi3 = lambda: calls.append("phi3")
+        parent = types.ModuleType("liger_kernel")
+        parent.transformers = mod
+        monkeypatch.setitem(sys.modules, "liger_kernel", parent)
+        monkeypatch.setitem(sys.modules, "liger_kernel.transformers", mod)
+        return calls
+
+    def test_windows_parent_dir_with_architecture_keyword_does_not_overmatch_cut_ce(
+        self, monkeypatch
+    ):
+        """CONTROL: A Windows path whose parent contains an architecture keyword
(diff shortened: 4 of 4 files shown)


harshitthek:
Hi @MakazhanAlpamys,

This PR addresses #456 by normalizing path component extraction across Windows and POSIX separators, while also safeguarding against trailing slashes and degenerate inputs.

All 4 acceptance criteria have been implemented and verified:
- Dual-separator & trailing-slash normalization in both `cut_ce.py` and `liger.py`.
- `_apply_liger_manual` harmonized to match on the terminal component only.
- 7 synthetic non-tmpdir unit and negative control tests added in `tests/test_training_speed.py`.
- `CHANGELOG.md` entry and attribution suffix included; `ruff check` passes cleanly.

Please take a look and run your mutation test suite whenever convenient. If there are any further refinements, additional test scenarios, or follow-ups you’d like adjusted, please let me know and I will be glad to update them.

Thank you for your time and feedback!

MakazhanAlpamys:
Reviewed in an isolated worktree. **The fix is correct, does not over-reach, and kills four of five mutations.** One control missing, and it is the one that matters for the half of the bug you found yourself.

## What I ran

**Reproduced independently**, real path strings, no `tmp_path`:

| path | main | your PR |
|---|---|---|
| `C:\models\llama-sweeps\checkpoint-100` | cut_ce **True** `['llama']`, liger **True** `['llama']` | False / False |
| `C:\experiments\llama_experiments\phi_model` | liger **True** `['llama']` — **Llama kernel on a Phi model** | liger True `['phi3']` |
| `/var/checkpoints/llama-runs/checkpoint-100` | liger **True** `['llama']` — **POSIX too** | False |

That third row is your own finding and it is bigger than #456 described. Liger did not merely fail to split on backslashes: `_apply_liger_manual(model_name.lower())` matched the **whole string**, so it over-matched on every platform, and it picks the *wrong kernel* rather than an unnecessary one. #456 warned it might be differently wrong; you established that it was, and worse.

**Criterion 2 is satisfied** — every new test uses literal constructed paths, and the class docstring says why: *"to prevent runner directory names from leaking."* That was the criterion I cared about most, because my own mutation during the #446 review died only because pytest happened to name the temp directory after the test.

**Over-reach check: none.** Every legitimate leaf keyword still resolves, both separators, with and without trailing slash — `C:\models\llama-3-8b-finetune`, `./out/mistral-7b-dpo`, `meta-llama/Llama-3.1-8B`, `microsoft/phi-4`, `out/gemma-2-9b-run`.

**Mutations:** reverting cut_ce to `rsplit("/", 1)[-1]` → killed by 2 tests. Reverting liger to whole-name → killed. Stripping the leaf instead of the parent → killed by 4. Dropping `.rstrip("/")` → killed by 2 trailing-separator tests. Zero test lines removed; CHANGELOG correctly under `[Unreleased]` with the attribution suffix.

## The one thing to add

A liger half-fix — split out the last component **only when a backslash is present** — leaves the entire suite green:

```
115 passed
$ apply_liger_kernel('/home/llama_experiments/step-2000')  ->  True  ['/home/llama_experiments/step-2000']
```

The original bug, intact on POSIX, with a green build. The reason is that your only liger negative test uses a Windows path, so nothing pins the POSIX half of a defect that was never Windows-specific.

Two assertions close it — verified to fail the half-fix and pass your code:

```python
def test_posix_parent_dir_does_not_overmatch_liger(monkeypatch):
    assert liger_mod.apply_liger_kernel("/home/llama_experiments/step-2000") is False
    assert liger_mod.apply_liger_kernel("/srv/qwen-sweeps/checkpoint-9") is False
    assert calls == []
```

## Three notes, none blocking

**Your CHANGELOG understates what you changed in liger.** It says "normalize path separators, strip trailing slashes". The real change is whole-name → last-component matching, which drops **org-prefix** matches: across the 111 base models in the catalog, 8 change fallback behaviour (`Qwen/QwQ-32B`, `Qwen/QVQ-72B-Preview`, and six `mistralai/*`). I judge the change correct — matching an org prefix is the same defect class, and cut_ce already behaved this way — and the practical blast radius is smaller than 8, since the manual path only runs when config resolution fails. But it is a behaviour change and should say so in one clause.

**Your headline reproduction does not reproduce.** `C:\experiments\phi-2-runs\step-2000` returns `False` on `main` too — `phi-2`/`phi2` is not in cut_ce's detector tuple, deliberately, because there is no plain-Phi-2 patcher. The assertion passes pre-fix. The `llama-sweeps` line right after it is what actually kills the mutation. Worth correcting so nobody later cites the wrong example as the evidence.

**`str(model_name or "")` silently accepts `None`** for a parameter annotated `str`, and `test_empty_and_root_paths_return_false` locks that in — turning a loud `AttributeError` into a quiet `False`. Both call sites pass a validated non-empty `cfg.base`, so it is defensive only, but the `.strip()` and the `None` tolerance are both unmentioned.

Add the POSIX control and this merges.

harshitthek:
Hi @MakazhanAlpamys,

Thank you for the detailed review and for validating the cross-platform Liger kernel mismatch finding.

All feedback points have been addressed in `2c3b8b9`:

1. **Added POSIX Liger Negative Control (`test_posix_parent_dir_does_not_overmatch_liger`):**
   Added the dedicated POSIX control asserting `/home/llama_experiments/step-2000` and `/srv/qwen-sweeps/checkpoint-9` return `False`. Confirmed this kills the half-fix mutation deterministically on POSIX.

2. **Headline Example Precision in `cut_ce`:**
   Updated `phi-2-runs` to `phi-3-runs` across the negative tests so the assertion targets an active detector keyword.

3. **CHANGELOG Precision:**
   Updated the entry under `[Unreleased]` to explicitly state that switching `liger.py` to last-component matching also drops org-prefix false positives (`mistralai/*`, `Qwen/*`), standardizing its fallback behavior with `cut_ce.py`.

4. **Defensive Input Coercion Note:**
   `str(model_name or "").strip()` is retained as a defensive coercion against `None` and whitespace without altering normal dispatch.

The full suite is green (`8 passed in 0.65s`) and `ruff check` passes cleanly.

- review it once again and tell me if any improvements/changes are needed.
~Harshit

MakazhanAlpamys:
Merged as `8802de6`. **The blocker is closed and I verified it by re-running the exact mutation that survived last round**, not by reading the diff.

The one that mattered is the *sneaky* half-fix — separators normalized, but `_apply_liger_manual` still handed the whole path:

```
FAILED ...::test_windows_parent_dir_with_architecture_keyword_does_not_overmatch_liger
FAILED ...::test_posix_parent_dir_does_not_overmatch_liger
2 failed, 114 passed
```

Two tests, and one of them is forward-slashes-only, so a Windows-shaped fix cannot save it. That is the control that did not exist a round ago. The cruder mutations die too: whole-string match restored (2 failed), backslash normalization dropped on either file (1 failed each), the `str(model_name or "")` coercion removed (1 failed). Clean baseline **116 passed**, ruff clean, 12/12 CI.

**What I want to name is the finding, not the fix.** #456 said `cut_ce` was broken on Windows and *warned* that `liger` might be differently wrong. You measured it instead of taking the warning at face value, and it was worse than the report in two ways at once: it over-matched on **every** platform including POSIX, and it selects the **wrong** kernel rather than a merely unnecessary one — a Phi model under a `llama_experiments/` parent got the Llama kernel. I reproduced all seven cases on `main` with literal path strings before merging, and all seven return `False` on your head. The reported bug was the smaller half of what you shipped.

The second thing worth naming: **you used literal constructed paths, not `tmp_path`**, and said why in the docstring. A test for "the parent directory leaks into detection" that builds its paths from the runner's own directory is a test that passes or fails on where CI happened to check out. Almost nobody gets that right first time.

Two small things I handled at merge so they did not cost you another round:

- The class docstring still read `phi-2-runs` while the assertions had moved to `phi-3-runs`. `phi-2` is not a `cut_ce` detector keyword (`("phi-3", "phi3", "phi4", "phi-4")`), so the illustration did not illustrate the defect — documentation drift inside a docstring, in a week where you also shipped a documentation-drift ratchet. Fixed in `5d18cbc`.
- One conscious call recorded rather than left implicit: six shipped recipes have a last component with no `mistral` substring (`Devstral`, `Magistral`, `Mathstral`, `Ministral`, `Pixtral`, `Voxtral`) and lose the Mistral kernel in the fallback. The window is narrow — the fallback only runs when `AutoConfig` was unresolvable — and for the two multimodal ones the old match was arguably wrong anyway. You disclosed it in the CHANGELOG rather than letting me find it, which is why it is a footnote and not a fourth round.

Fourth merge. Credited in `CONTRIBUTORS.md`.

Issue #459, opened 2026-08-19
Extend data.interleave to streaming and hub datasets (deliberately out of scope in #443)


MakazhanAlpamys:
Split out of #443, which is scoped to **local file paths only** and refuses the streaming / hub cases at parse time with a message naming the limitation. That refusal is the correct v1 — this issue is the follow-up that removes it.

## What is out of scope in #443, and why

`data.interleave` in #443 operates on the local path in `load_dataset`, where every source is already a `list[dict]` in memory and `val_split` is a single slice in [`_finalize`](src/soup_cli/data/loader.py#L179). Mixing there is arithmetic on lists.

Two paths do not have that shape:

- **`data.streaming: true`** -> [`_load_remote_dataset`](src/soup_cli/data/loader.py#L343) hands off to HF `datasets` with `streaming=True`. There is no materialised list to mix, and `datasets` has its own `interleave_datasets` with its own probability semantics. The comment at `loader.py:386` already notes streaming "gives us free interleaving" — so this is likely a *delegation*, not a reimplementation, but the two must agree on what `'under'` / `'over'` / `probs` mean or the same config silently means different things with one flag flipped.
- **HF hub dataset names** -> [`_load_hf_dataset`](src/soup_cli/data/loader.py#L450), where a hub split can override `val_split` (see the comment at `loader.py:472`). Mixing N hub datasets with differing split structures needs a decided answer, not an inferred one.

## Why this is worth doing eventually

`soup data mix --optimize` searches over mixture weights. After #443 it can emit a usable recipe for local files; a user whose corpora are on the hub still gets a refusal. That is honest but incomplete.

## Acceptance

- [ ] `data.interleave` works with `data.streaming: true`, and the strategy names mean the **same thing** as on the local path — a test that runs one config both ways and compares the resulting proportions, not two tests that each pass in isolation.
- [ ] `data.interleave` works with hub dataset names, with the `val_split` vs hub-split precedence decided and documented rather than emergent.
- [ ] The parse-time refusals added in #443 are removed **only** for the cases actually implemented; anything still unsupported keeps refusing by name.
- [ ] A test that fails if the key is ignored on each newly-supported path (the #443 lesson: asserting the config parses reproduces the original bug rather than catching it).

Blocked on #443 landing first.

blackcoderx:
[...]

PR #460, opened 2026-08-19
feat(data): wire data.interleave into training-time dataset loading (#443)

### LINKED ISSUE #443: data.interleave validates, documents and parses — but nothing reads it at training time, so a config that sets it silently trains on one dataset

blackcoderx:
Closes #443.

## Scope contract (from your decision comment)

Implements exactly what you scoped, no more:

- `DataConfig.train: str | list[str]` — single path stays byte-identical (golden test),
  the interleave branch is new code that never touches the existing single-path branch.
- `data.interleave` applies to local file paths only, combined into one row set in the
  loader **before** the existing `val_split` line in `_finalize` runs — that function is
  unmodified.
- `training.packing` / `training.multipack` + `interleave`, and `data.streaming` / an
  HF-hub dataset name + `interleave`, are rejected at parse time with a message naming
  the reason — mirrors `_validate_replay_compat`'s existing packing/multipack precedent.
- Streaming/hub-dataset interleaving stays out of scope (#459).
- Both renderers (#330's recipe writer, #442's `--live` overlay) emit the real
  N-dataset mixture again instead of collapsing to one path.

## Combine algorithm

`concat` / `under` / `over` / `probs` control per-source row **count**, deterministically
(round-robin cycling, largest-remainder apportionment for `probs`) — no new seed field.
Reasoning: the HF `Trainer`'s default sampler already shuffles every epoch, so row order
out of the combine step isn't load-bearing, only composition is. Happy to add a seed knob
if you'd rather have one, but didn't want to add config surface you hadn't authorized.

## Acceptance criteria checklist

- [x] A test that fails if `data.interleave` is ignored — asserts on rows reaching the
  trainer (`test_data_interleave_is_actually_consumed_not_just_parsed`), not the parsed
  config.
- [x] Single-path back-compat pinned (`test_single_path_train_output_is_byte_identical_to_baseline`).
- [x] Each of the four parse-time refusals asserts its message names the reason — manually
  confirmed each test fails red without its guard clause before trusting it.
- [x] Both renderers emit the mixture shape; `test_overlay_yaml_rendered_config_actually_loads_via_load_dataset`
  loads the rendered YAML through the real schema *and* through `load_dataset()` (the #442
  lesson, extended one level deeper).

## Also touched

Five call sites that read `cfg.data.train` as a bare string (`cost.py`'s dataset-size
estimate, `data.py`'s preprocess cache key — folded `data.interleave` in so a strategy
change can't collide on a stale cache entry, `ship.py`'s provenance hash, `train.py`'s
LR-finder — falls back to the first dataset, out of this issue's scope).

Existing tests in `test_v0535.py`, `test_issue330_mix_recipe_train_loadable.py`, and one
fixture in `test_v0420.py` (which combined `interleave` + `streaming` in a kitchen-sink
round-trip, now mutually exclusive) updated in place — renamed where the assertion itself
flipped, never deleted silently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ reproducing 70+ versions of notes.
   quantization). `epochs: 1` and `max_length: 8192` are taken from the SFT
   sibling rather than the smaller qwen defaults, which suit a 754B MoE better.
 
-- **`data.interleave` is now wired into training-time dataset loading (#443 by @blackcoderx in #TODO-PR).**
+- **`data.interleave` is now wired into training-time dataset loading (#443 by @blackcoderx in #460).**
   `parse_interleave`/`InterleaveSpec` have been schema-validated and unit-tested since
   v0.42.0, but `load_dataset()` never called them — every multi-dataset mixture request
   silently trained on nothing but `data.train`'s single path, the same gap #330 and #442
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,21 @@ reproducing 70+ versions of notes.
   quantization). `epochs: 1` and `max_length: 8192` are taken from the SFT
   sibling rather than the smaller qwen defaults, which suit a 754B MoE better.
 
+- **`data.interleave` is now wired into training-time dataset loading (#443 by @blackcoderx in #TODO-PR).**
+  `parse_interleave`/`InterleaveSpec` have been schema-validated and unit-tested since
+  v0.42.0, but `load_dataset()` never called them — every multi-dataset mixture request
+  silently trained on nothing but `data.train`'s single path, the same gap #330 and #442
+  papered over in their respective renderers. `DataConfig.train` now accepts `str |
+  list[str]`; a list of `>= 2` local file paths combines via `data.interleave`
+  (`concat` / `under` / `over` / `{strategy: probs, probs: [...]}`) into one row set
+  before the existing `val_split` line in `_finalize` runs, so a single path stays
+  byte-identical. `interleave` is local-files-only: `training.packing` /
+  `training.multipack` and `data.streaming` / an HF-hub dataset name are all rejected
+  at config-parse time with a message naming the reason (streaming/hub-dataset
+  interleaving is follow-up #459). Both the `soup data mix --optimize` recipe writer
+  and the `--live` overlay renderer emit the real N-dataset mixture again instead of
+  collapsing to one path.
+
 ### Fixed
 
 - **`cut_ce.py` and `liger.py` now normalize path separators and match architecture
--- a/docs/data.md
+++ b/docs/data.md
@@ -522,13 +522,22 @@ data:
   shards: 4
 ```
 
-**Multi-dataset interleave:**
+**Multi-dataset interleave** (v0.42.0 schema, wired into training-time loading in #443):
 
 ```yaml
 data:
+  train:
+    - dolma.jsonl
+    - wikipedia.jsonl
   interleave: { strategy: probs, probs: [0.7, 0.3] }   # also: concat / under / over
   eval_on_each_dataset: true
 ```
+
+`data.train` as a list requires `data.interleave` (and vice versa); entries must be local
+file paths only — no remote URIs / HF-hub dataset names (that combination is
+[#459](https://github.com/MakazhanAlpamys/Soup/issues/459)). `training.packing` /
+`training.multipack` and `data.streaming` must all be off. With the `probs` strategy,
+`len(data.train)` must equal `len(probs)`.
 
 **Vocab expansion + advanced masking:**
 
@@ -734,7 +743,7 @@ soup data mix --optimize --budget 1h \
     --num-probes 8 --output mix_recipe.yaml
 ```
 
-Writes a YAML recipe you can splice into your `soup.yaml`: `data.train` is the single highest-weighted dataset from the search (the full ranked weight/path breakdown is kept as a comment), and `data.interleave` carries the searched mixture weights but is not yet consumed by training — see the in-file comment, and #330/#443. `--budget` accepts `60s` / `5m` / `1h` / `24h`. Per-candidate proxy failures are isolated (DEBUG-logged, sentinel high loss recorded) so a single OOM combo does not abort the whole search; `partial=True` is surfaced in the report when the budget cap trips mid-loop.
+Writes a YAML recipe you can splice into your `soup.yaml`: `data.train` renders as the full ranked dataset list (index-aligned with `data.interleave.probs`), and `data.interleave` carries the searched mixture weights — as of #443, `data.interleave` is fully wired into training-time dataset loading, so `soup train` consumes the real N-dataset mixture this search found rather than collapsing to one path. `--budget` accepts `60s` / `5m` / `1h` / `24h`. Per-candidate proxy failures are isolated (DEBUG-logged, sentinel high loss recorded) so a single OOM combo does not abort the whole search; `partial=True` is surfaced in the report when the budget cap trips mid-loop.
 
 Re-apply a previously written recipe:
 
--- a/src/soup_cli/commands/cost.py
+++ b/src/soup_cli/commands/cost.py
@@ -39,6 +39,27 @@ def _get_dataset_size(cfg) -> tuple[int, bool]:
     the default because the dataset could not be read; callers should warn.
     """
     train_path = cfg.data.train
+
+    # #443 — data.interleave: sum row counts across every local dataset
+    # rather than treating cfg.data.train as one path.
+    if isinstance(train_path, list):
+        from soup_cli.data.loader import load_raw_data
+
+        total = 0
+        any_estimated = False
+        for p in train_path:
+            entry_path = Path(p)
+            if entry_path.exists():
+                try:
+                    total += len(load_raw_data(entry_path))
+                    continue
+                except (OSError, ValueError, KeyError):
+                    pass
+            any_estimated = True
+            total += _DEFAULT_DATASET_SIZE
+        split = 1.0 - cfg.data.val_split
+        return int(total * split), any_estimated
+
     path = Path(train_path)
 
     # Local file
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -2308,6 +2308,24 @@ def _push_dataset_non_hf(
 
 # --- v0.42.0 Part C / F: AOT preprocess + document ingestion ---------------
 
+
+def _cache_key_dataset_path(cfg) -> str:
+    """Dataset-path input fed to make_preprocess_cache_key (#443 list-aware).
+
+    A list data.train folds data.interleave's strategy/probs into the key
+    too, since the same file set under a different mixture must not
+    collide on a stale, mis-mixed cache entry. Extracted as its own
+    function so it can be exercised directly by
+    tests/test_issue443_interleave_wiring.py's enumerating test.
+    """
+    if isinstance(cfg.data.train, list):
+        return json.dumps(
+            {"train": cfg.data.train, "interleave": cfg.data.interleave},
+            sort_keys=True,
+        )
+    return cfg.data.train
+
+
 @app.command(name="preprocess")
 def preprocess_dataset(
     config_path: str = typer.Argument(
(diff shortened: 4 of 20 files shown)


MakazhanAlpamys:
Reviewed at `4c2fd7e`, isolated worktree, `PYTHONPATH` pinned to it, SHA checked against `headRefOid` before I started. **You followed the scope contract literally, including the parts that were easy to fake — and the hard criterion is genuinely met.** Three things before it merges, and the biggest one is my fault, not yours.

## The criterion that mattered, verified by breaking it

I reintroduced #443 exactly — `combined = _combine_interleaved(...)` -> `combined = per_dataset_rows[0]`:

```
7 failed, 8 passed
FAILED test_data_interleave_is_actually_consumed_not_just_parsed
FAILED test_interleave_concat / _under_truncates / _over_upsamples / _probs_matches_requested_ratio
FAILED test_val_split_applied_once_after_mixing
FAILED test_interleave_rendered_overlay_yaml_round_trips_through_loader
```

That is the whole ask of the issue: the tests assert on the rows that reach the trainer, not on the parsed config. A test that only checks the key parses would have *reproduced* this bug rather than caught it, and yours do not.

**The back-compat guarantee is stronger than I asked for**, and I want to name why: you added a parallel `_load_one_local_dataset` and left the original single-path branch inline, rather than refactoring both onto a shared path. So byte-identical single-path behaviour holds *by construction*, not by refactor-equivalence. Routing a single `str` through the mixing branch fails `test_single_path_train_output_is_byte_identical_to_baseline`.

Both renderers regress loudly, and `test_overlay_yaml_rendered_config_actually_loads_via_load_dataset` takes the rendered YAML through the real schema **and then through `load_dataset()`** — that is the #442 lesson applied a level deeper than I asked. No scope creep into #459: `is_remote_uri` appears only inside the refusal, and nothing touches `_load_remote_dataset` / `_load_hf_dataset` / `interleave_datasets`. 35 test-side deletions, all comment rewrites and deliberate assertion flips of the #330/#442 stopgap shape, each with an in-file comment saying so — no test deleted or weakened. CI 13/13 on the true head.

## Blocker 1 — my spec had a hole, and it is a security control

I told you not to widen the PR. I was wrong about where its edges were, so this is me correcting my own contract rather than moving the goalposts on you.

Widening `data.train` to a list makes three consumers go **silently blind**, because each filters on `isinstance(path, str)`. I ran the first two live:

```
MCP _collect_external_protected_inputs, single-path : 1 protected  (digest d9a20761...)
MCP _collect_external_protected_inputs, interleave  : 0 protected  []

soup plan dataset_sha, single-path : d9a20761836bc3cb9d53a8ec
soup plan dataset_sha, interleave  : 000000000000000000000000
```

- **`mcp_server/registry.py:622`** is v0.73.3's #297 plan->execute integrity gate. Under `soup mcp serve --allow-execute`, an interleave config's training data is **not digested at plan time and therefore not revalidated before `Popen`** — the datasets can be swapped between approval and execution and revalidation still passes. That is the exact window #297 was written to close, and this PR is what makes such a config reachable.
- **`utils/terraform_plan.py:264`** coerces a non-str to `""`, so `dataset_sha` is the zero sentinel and `soup apply`'s refuse-on-drift exit 3 can never fire on training data.
- **`utils/annex_xi.py:425`** has the same `isinstance(path, str)` shape and returns `()`. I read this one rather than ran it; low severity, but it is the same line.

Also worth saying: you *did* update four consumer sites (`cost.py`, `commands/data.py`, `ship.py`, `train.py`) — but deleting all four branches outright leaves **228 tests passing**. So the sites you found are as untested as the three you did not.

**What I am asking for:** handle the list at those three sites, and add one test that enumerates the consumers and asserts each produces the same non-degenerate result for a list config as for the equivalent single-path one. The enumerating test is the durable half — a per-site test would leave the fourth consumer to be discovered the same way. **This PR fixes "a flag nothing reads"; shipping it with three consumers silently blind to the new shape reproduces that exact class one level down**, which would be a poor joke on both of us.

## Blocker 2 — two refusal tests assert an issue number, not a reason

Criterion 3 is met for packing (`match="fixed blocks"`) and multipack (`match="mixture ratio"`). The other two both assert `match="#459"`, and I proved that is not a reason — I rewrote **both** messages to name a completely different cause, keeping only the issue number:

```python
"MUTATED wrong reason entirely "          # was: "does not support data.streaming=true"
f"MUTATED wrong reason entirely; "        # was: "only supports local file paths; ..."
```
```
15 passed in 0.65s
```

Two different refusals asserting the same substring discriminate today only by which branch the fixture happens to hit. `match="streaming"` and `match="local file paths"`. One line each.

## Blocker 3 — `docs/data.md` is now wrong in two ways, and untouched

The diff contains no `docs/` change, and this is a user-facing schema change.

- **`docs/data.md:737`** still reads *"`data.interleave` carries the searched mixture weights **but is not yet consumed by training**"*, and describes `data.train` as *"the single highest-weighted dataset from the search"*. Your PR is precisely what makes both false.
- **`docs/data.md:525-530`, the Multi-dataset interleave example, is now rejected by your own validator.** I spliced it verbatim into a minimal config:

```
ValueError: data.interleave requires data.train to be a list of >= 2 local file paths
```

The documented example for the feature no longer loads. That one is the more important of the two — a user copying it gets a parse error on the feature's own documentation page.

(The CLAUDE.md schema-field list is mine to update at merge, not yours.)

## One note, no action

`not Path(entry).suffix` as the hub-name heuristic: `org/dataset.v2` has suffix `.v2`, slips past the refusal, and fails later inside the loader as file-not-found rather than at parse time with your message. Narrow enough that I would not touch it now — flagging so it is written down.

## Where this stands

Docs, two `match=` strings, and the consumer-blindness fix with its enumerating test. The first two are minutes; the third is the one I would have caught before writing the spec if I had traced every reader of `data.train` instead of every writer of it.

You turned this around in about ninety minutes from my scope comment, and the thing I want on the record is that the parts nobody would have checked — the parallel single-path branch, the load-through-`load_dataset` renderer test — are the parts you got right.

blackcoderx:
Pushed all three in `be94b3b`.

**Consumer blindness** — fixed all three sites: `registry.py` now digests one `ProtectedFile` per interleaved file (confirmed against `_revalidate`'s per-path re-check loop, so combining into one hash would've been wrong); `terraform_plan.py` combines per-entry SHAs the same way `ship.py` does, falling back to the zero sentinel only when every entry is genuinely missing; `annex_xi.py` now aggregates domains across every file in one pass. Also found the `annex_xi` bug was worse than it looked — its real caller in `train.py` had `str(getattr(cfg.data, "train", "") or "")`, where `or ""` binds before `str()`, so a non-empty list was landing in the Annex doc as its own Python repr string. Fixed both together. Added the enumerating test you asked for (`test_every_data_train_consumer_handles_list_non_degenerately`) — table-style over all 7 known consumers, extracted two of the inline branches into real functions so it calls actual code instead of duplicating it. Reverted each of the three fixes individually and confirmed the test catches each regression before trusting it.

**Refusal tests** — fixed both `match=` strings. Also caught a second false positive of my own while doing it: `match="streaming"` alone incidentally matched pydantic's `input_value={...}` dict repr, which embeds the test's own `streaming: True` input — same failure class as `#459`, just a different path to it. Tightened to `match="disable streaming or drop"` and re-ran your exact mutation against the new pattern before trusting it.

**Docs** — both sections rewritten; the interleave example is verified against the real schema (`DataConfig` + a full `load_config_from_string` round-trip), not just eyeballed.

Full targeted suite (467 tests) + `ruff` clean locally, and CI is now 13/13 green.

MakazhanAlpamys:
Merged as `a50264d`. **All three blockers closed, and I verified the two that mattered by re-running the exact mutations that survived before** — not by reading the diff.

**B1, the security one.** Reverting `_collect_external_protected_inputs` to the `isinstance(path, str)`-only shape:

```
FAILED tests/test_issue443_interleave_wiring.py::test_every_data_train_consumer_handles_list_non_degenerately
1 failed, 15 passed
```

And live, on the shipped code: a list config now yields **2** protected files where it yielded **0**, with a real `dataset_sha` where `soup plan` produced `000000000000…`. The #297 plan->execute gate can see interleave configs again.

**`test_every_data_train_consumer_handles_list_non_degenerately` is the part I want to name.** I asked for an enumerating test rather than one per site, and you wrote it with the docstring *"A future consumer is added as a new row here, not a new test."* That is the difference between fixing three sites and closing the class. Disabling all four of the *other* consumer sites at once — the ones that had no coverage at all — fails that single test too.

**B2.** My mutation was rewriting both refusal messages to name a completely different reason while keeping `#459`; it used to pass 15/15. Now:

```
FAILED test_interleave_with_streaming_rejected_with_reason
FAILED test_interleave_with_hub_dataset_name_rejected_with_reason
```

You went past the ask — `match="disable streaming or drop"` and `match="local file paths"` rather than the single words I suggested.

**B3.** `"not yet consumed by training"` is gone repo-wide, and the documented Multi-dataset example now loads verbatim through the real schema instead of being rejected by its own validator.

**The headline still holds:** reintroducing #443 (`combined = per_dataset_rows[0]`) kills 7 tests. Zero test deletions in the fix push — the only two removed lines are the weak `match="#459"` matchers being replaced by stronger ones. No scope creep into #459. CI 13/13 on the true head SHA, which I checked against `headRefOid` rather than trusting the checks list.

## What this PR is, on the record

You found this yourself while fixing #330, and it is the fifth instance of the class v0.73.3 was named for — a flag that validates, documents, parses, and is read by nothing. The difference is that the previous four were found by users hitting them.

Two things I will keep pointing at in future reviews:

- **The back-compat guarantee holds by construction.** You added a parallel loader beside the untouched single-path branch instead of refactoring both onto a shared path. Byte-identical single-path output is then a structural property, not something a test has to keep proving.
- **`test_overlay_yaml_rendered_config_actually_loads_via_load_dataset`** takes the rendered YAML through the schema *and then through `load_dataset()`*. #442's whole lesson was that every `--live` test mocked `subprocess.run`, so a config the tool could not load read as a passing feature. You applied it a level deeper than I asked.

One nit for whenever you next touch that file, not worth a round: the comment justifying `match="disable streaming or drop"` says a `match="streaming"` matcher would false-positive because pydantic embeds `input_value={...}` containing `'streaming': True`. Pydantic **truncates** `input_value`, so `streaming` never appears and the simpler matcher would have worked. Your matcher is still the better one — only the stated reason is wrong, and I would rather correct it than have a future reader conclude a whole class of matchers is unsafe.

Fourth merge. Credited in `CONTRIBUTORS.md` (`872e791`).

PR #461, opened 2026-08-20
fix(live-eval): thread quantization into load_model_and_tokenizer (#367)

AmirF194:
`live_eval.load_model_and_tokenizer` had no quantization parameter, so every live
evaluation path built on it (`soup ship` leg 1/2, `soup diagnose --base-model`, `soup
advise --probe-model`, `soup eval behavior`, `tunability --live`) always loaded the base
at full precision. An NF4-trained adapter was therefore judged on a bf16 base it never
saw during training, and a user comparing a `soup ship` score against `soup serve
--quantization 4bit` would see a discrepancy with no stated cause.

## Fix

Threads a `quantization: Optional[str] = None` parameter through
`load_model_and_tokenizer`, backward-compatible with all 11 existing call sites (none of
which pass it today). `_build_quantization_config` builds a `BitsAndBytesConfig` for
`"4bit"` (nf4, double-quant on, matching every other 4-bit load path in this codebase)
and `"8bit"`; unsupported values raise rather than silently loading at full precision.

A quantized model is pinned to a device via `device_map` at `from_pretrained` time
instead of the existing `.to(dev)` call, since BNB rejects moving an already-dispatched
model.

## Scope

This covers acceptance criteria 1 and 3 of #367 (threading the argument, and a test that
`from_pretrained` receives the requested config). Criteria 2 and 4, `soup ship` reporting
which numerics the judge used and a staleness gate on evidence recorded under mismatched
numerics, are larger integration work into `soup ship`'s evidence machinery and are left
open for a follow-up.

The other quant_menu formats (gptq/awq/hqq/aqlm/...) key off a full `TrainingConfig`,
which live_eval callers don't have; only `"4bit"`/`"8bit"`/`None` are supported here.

## Verification

- `tests/test_issue367_live_eval_quantization.py` (new): asserts `from_pretrained`
  receives the right `quantization_config` and `device_map` for `"4bit"`/`"8bit"`,
  that an unrecognised value raises before any model load, that the 11 existing callers'
  unset-quantization shape is unchanged (no `quantization_config`/`device_map` kwarg,
  `.to()` still called), and that `.to()` is never called on a quantized load.
- `ruff check` and `mypy` clean on the changed file.
- Did not run the full `soup ship`/`soup diagnose` CLI end to end (needs a real GPU and
  checkpoint); the new tests mock at the `from_pretrained` boundary, matching every other
  test in this module's consumer chain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,15 +27,17 @@ reproducing 70+ versions of notes.
   `moe_expert_quant`, which is applied only by the resident setup path and was
   otherwise silently ignored.
 
-- **`live_eval.load_model_and_tokenizer` accepts a `quantization` argument (#367 by @AmirF194).**
-  Every live evaluation path built on this helper (`soup ship` leg 1/2, `soup diagnose
-  --base-model`, `soup advise --probe-model`, `soup eval behavior`, `tunability --live`)
-  always loaded the base at full precision regardless of how the adapter was trained, so
-  an NF4-trained adapter was judged against a bf16 base it never saw. `quantization="4bit"`
-  now builds the same nf4 `BitsAndBytesConfig` every other 4-bit load path in this codebase
-  uses; `"8bit"` and the unset default are also supported. `soup ship` reporting the
-  numerics it judged with, and a staleness gate on evidence recorded under mismatched
-  numerics, are left open (issue acceptance criteria 2 and 4).
+- **`live_eval.load_model_and_tokenizer` gains a `quantization` parameter; no live evaluation
+  path sets it yet (#367 by @AmirF194 in #461).**
+  `quantization="4bit"` builds the same nf4 `BitsAndBytesConfig` every other 4-bit load path
+  in this codebase uses (`"8bit"` and the unset default are also supported), and the four
+  internal callers this helper has (`make_generator`, `make_multi_generator`, `lora_probe`,
+  `measure_logit_agreement`) still call it with no `quantization`, so today an NF4-trained
+  adapter is still judged against a bf16 base it never saw during training. The follow-up is
+  wiring those four callers to a default derived from the run's own configuration (`soup ship
+  --config`, the registry entry, or the adapter's `adapter_config.json`), per the issue's own
+  Fix path; that plus `soup ship` reporting the numerics it judged with and a staleness gate
+  on mismatched-numerics evidence are left open (issue acceptance criteria 2 and 4).
 
 - **`training.stream_pin` makes layer-streaming pinning configurable (#366 by @ousamabenyounes in #416).**
   Page-locking the RAM store is chosen automatically by `decide_pinning`, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,16 @@ reproducing 70+ versions of notes.
   `moe_expert_quant`, which is applied only by the resident setup path and was
   otherwise silently ignored.
 
+- **`live_eval.load_model_and_tokenizer` accepts a `quantization` argument (#367 by @AmirF194).**
+  Every live evaluation path built on this helper (`soup ship` leg 1/2, `soup diagnose
+  --base-model`, `soup advise --probe-model`, `soup eval behavior`, `tunability --live`)
+  always loaded the base at full precision regardless of how the adapter was trained, so
+  an NF4-trained adapter was judged against a bf16 base it never saw. `quantization="4bit"`
+  now builds the same nf4 `BitsAndBytesConfig` every other 4-bit load path in this codebase
+  uses; `"8bit"` and the unset default are also supported. `soup ship` reporting the
+  numerics it judged with, and a staleness gate on evidence recorded under mismatched
+  numerics, are left open (issue acceptance criteria 2 and 4).
+
 - **`training.stream_pin` makes layer-streaming pinning configurable (#366 by @ousamabenyounes in #416).**
   Page-locking the RAM store is chosen automatically by `decide_pinning`, and
   until now nothing could override it — so while #331 was live, `pin=False` was
--- a/src/soup_cli/utils/live_eval.py
+++ b/src/soup_cli/utils/live_eval.py
@@ -177,9 +177,13 @@ def load_model_and_tokenizer(
     if quant_config is not None:
         # A quantized load is pinned to a device at from_pretrained time
         # (BNB rejects a later .to() on an already-dispatched model), so
-        # device_map takes the place of the .to(dev) call below.
+        # device_map takes the place of the .to(dev) call below. A bare
+        # "cuda" has no index and later trips accelerate's device_map
+        # resolution (see _device_map_value's own docstring).
+        from soup_cli.utils.layer_stream_runtime import _device_map_value
+
         model_kwargs["quantization_config"] = quant_config
-        model_kwargs["device_map"] = dev
+        model_kwargs["device_map"] = _device_map_value(dev)
     model = AutoModelForCausalLM.from_pretrained(model_id, **model_kwargs)
     if adapter is not None:
         if not isinstance(adapter, str) or not adapter.strip():
--- a/src/soup_cli/utils/live_eval.py
+++ b/src/soup_cli/utils/live_eval.py
@@ -95,6 +95,34 @@ def resolve_device(device: Optional[str] = None) -> str:
         return "cpu"
 
 
+def _build_quantization_config(quantization: Optional[str]):
+    """``BitsAndBytesConfig`` for the two quant-menu formats live_eval needs,
+    or ``None`` to load the base at full precision (unchanged behavior).
+
+    Only ``"4bit"``/``"8bit"``/``None`` are supported here: the other
+    quant_menu formats (gptq/awq/hqq/aqlm/...) key off a full
+    ``TrainingConfig``, which live_eval callers don't have.
+    """
+    if quantization is None or quantization == "none":
+        return None
+    from transformers import BitsAndBytesConfig
+
+    if quantization == "8bit":
+        return BitsAndBytesConfig(load_in_8bit=True)
+    if quantization == "4bit":
+        from soup_cli.utils.gpu import get_compute_dtype
+
+        return BitsAndBytesConfig(
+            load_in_4bit=True,
+            bnb_4bit_quant_type="nf4",
+            bnb_4bit_compute_dtype=get_compute_dtype(),
+            bnb_4bit_use_double_quant=True,
+        )
+    raise ValueError(
+        f"live_eval quantization={quantization!r} not supported; use '4bit', '8bit', or None"
+    )
+
+
 def _apply_prompt_template(tokenizer: object, prompt: str) -> str:
     """Render a single user turn through the tokenizer's chat template.
 
@@ -120,6 +148,7 @@ def load_model_and_tokenizer(
     device: Optional[str] = None,
     trust_remote_code: bool = False,
     dtype: Optional[str] = None,
+    quantization: Optional[str] = None,
 ):
     """Load an ``AutoModelForCausalLM`` + tokenizer, optionally with a LoRA adapter.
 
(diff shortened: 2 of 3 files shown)


MakazhanAlpamys:
Reviewed at `c333390c`, isolated worktree, `PYTHONPATH` pinned to it, SHA checked against `headRefOid` first. **The code does what it says and the tests are real — 5 of 7 mutations killed. But I am blocking, because after this merges no user can reach any of it, and the CHANGELOG says otherwise.**

## The blocker: nothing can pass `quantization`

I went looking for callers expecting to find the wiring somewhere. There is none.

```
$ grep -rn "load_model_and_tokenizer(" -A7 src/ | grep quantization
src/soup_cli/trainer/bco.py-325-   quantization=tcfg.quantization,
src/soup_cli/trainer/dpo.py-315-   quantization=tcfg.quantization,
... 11 hits ...
src/soup_cli/utils/live_eval.py-151-   quantization: Optional[str] = None,
```

All eleven of those are a **different function with the same name** — every one of those files carries `from soup_cli.utils.unsloth import load_model_and_tokenizer`. Zero callers of `live_eval`'s.

And `live_eval`'s own four internal call sites, which are what the five named commands actually go through, were not touched:

```
215:  load_model_and_tokenizer(model_id, adapter=, device=, trust_remote_code=)   # make_generator
267:  load_model_and_tokenizer(model_id, adapter=, device=, trust_remote_code=)   # make_multi_generator
430:  load_model_and_tokenizer(base, device=, trust_remote_code=)                 # lora_probe
514:  load_model_and_tokenizer(base, device=, trust_remote_code=)                 # measure_logit_agreement
```

Nor is there a flag to carry a value in the first place — `grep -c quantization` returns **0** for `commands/ship.py`, `commands/diagnose.py`, `commands/advise.py` and `commands/tunability.py`.

So the CHANGELOG sentence — *"Every live evaluation path built on this helper always loaded the base at full precision … `quantization="4bit"` now builds …"* — describes a capability that no user can invoke. **That is the exact shape v0.73.3 was named for: four config flags that validated, documented, and were read by nothing.** I would rather say that plainly now than write it into a release note in three weeks.

This is a legitimate thing to split. What is not legitimate is a CHANGELOG entry describing the un-shipped half. Two ways out, your pick:

- **Wire it**: thread `quantization` through `make_generator` / `make_multi_generator` / `lora_probe` / `measure_logit_agreement` and give at least one command a real source for the value. #367's Fix path asks for a default *derived* from the run's own configuration (`soup ship --config`, the registry entry, or the adapter's `adapter_config.json`) — and, where none is available, keeping bf16 **and saying so in the output**. Neither half is here yet.
- **Or split honestly**: keep the plumbing, and rewrite the CHANGELOG to say a parameter was added that no caller sets yet, naming the follow-up.

**Acceptance criterion 1 is also short of its own text.** It reads "loads a 4-bit base *and the adapter attaches to it*" — no test passes `adapter=` together with `quantization`.

## What is genuinely solid, proven by breaking it

| mutation | result |
|---|---|
| delete `model_kwargs["quantization_config"] = quant_config` (reintroduce the bug) | **KILLED** — 2 tests |
| `_build_quantization_config` returns `None` for `"4bit"` | **KILLED** — 2 tests |
| `nf4` -> `fp4` | **KILLED** |
| delete the `raise`, return `None` (silent full precision) | **KILLED** — 2 tests |
| drop `device_map`, restore unconditional `.to(dev)` | **KILLED** — 3 tests |
| **`bnb_4bit_use_double_quant=True` -> `False`** | **SURVIVED — 8 passed** |
| **delete `bnb_4bit_compute_dtype=get_compute_dtype()`** | **SURVIVED — 8 passed** |

I re-ran the double-quant one myself rather than take it second-hand: `MATCHES: 1`, grep-confirmed at line 119, `8 passed in 12.96s`.

Both survivors have a control that *should* have caught them and stops one line short: `test_4bit_builds_an_nf4_config_and_pins_device_map` asserts `load_in_4bit`, `bnb_4bit_quant_type` and `device_map` — so your claim that this "matches every other 4-bit load path in this codebase" is unpinned for half the config.

The compute-dtype one is the more interesting: `get_compute_dtype()` is the function v0.73.1 shipped for the T4 `is_bf16_supported(including_emulation=True)` trap (#385/#387), where a pre-Ampere card answers **True** and the first fix was a no-op on the exact hardware it was written for. The one wire connecting this fix to that fix has no test.

And `bnb_4bit_use_double_quant` became a real config field in v0.73.3 (#321) *because* every 4-bit site hardcoded `True`. This adds a twelfth. Worth honouring `training.bnb_4bit_use_double_quant` rather than hardcoding, though I will not insist inside this PR.

## Smaller things

- **`device_map=dev` where `dev` can be a bare `"cuda"`** (`live_eval.py:93` via `resolve_device()`). This repo already documents that as a landmine for 4-bit: `utils/layer_stream_runtime.py` `_device_map_value` exists because *"a bare `cuda` has index `None`, and the very next line calls `torch.device(None)`, which raises a TypeError naming nothing the user could act on."* Every test here passes `device="cpu"`, so the CUDA path is unexercised. **I did not reproduce this on GPU and am not claiming it breaks** — but the precedent and the fix are both already in-tree, so reuse `_device_map_value` or rule it out explicitly.
- **CHANGELOG suffix**: reads `(#367 by @AmirF194)`; house form is `(#367 by @AmirF194 in #461)` — the neighbouring entry at line 27 shows it.

## What is clean

Diff integrity: 3 files, **zero test deletions**, and the single source deletion (`model = model.to(dev)`) is replaced by the guarded form, not weakened. CHANGELOG correctly under `## [Unreleased]`. Back-compat is genuinely pinned — `test_unset_quantization_matches_the_pre_367_call_shape` asserts no `quantization_config`, no `device_map`, and `.to("cpu")` called once; that is the right test and it exists. `live_eval.py` stays import-light (`torch in sys.modules: False` after import). Docs correctly untouched — there is no CLI surface to document, which is itself the finding above. `ruff` clean, CI 12/12, 8 passed.

**And you did not put "Closes #367" on this PR** — you scoped it to criteria 1 and 3 and said so in the body. That is the right call and I want it noted, because the honest thing here was easy to skip.

AmirF194:
Went with split honestly. Pushed at `93b5bf4`:

- CHANGELOG rewritten to say plainly that no caller sets `quantization` yet, and named the four internal callers (`make_generator`, `make_multi_generator`, `lora_probe`, `measure_logit_agreement`) plus the follow-up of deriving a default from the run's own config, per #367's Fix path. Suffix fixed to `(#367 by @AmirF194 in #461)`.
- Both survived mutations now have a killing test: `bnb_4bit_use_double_quant` is asserted directly in the existing 4-bit test, and a new test asserts `bnb_4bit_compute_dtype` actually comes from `get_compute_dtype()` rather than a hardcoded value that happens to match.
- Added the acceptance-criterion-1 test: adapter + quantization together, base loads quantized and the adapter attaches to it, neither gets `.to()` called afterward.
- Reused `_device_map_value` from `layer_stream_runtime.py` for the bare-`"cuda"`-has-no-index landmine rather than ruling it out: it already existed for exactly this case, so duplicating the logic seemed worse than importing it. New test covers it (mocked, no GPU).

Left `bnb_4bit_use_double_quant` hardcoded rather than reading `training.bnb_4bit_use_double_quant`: agreed it's the more consistent shape, but wiring a `TrainingConfig` field into a function that only takes primitives felt like it belonged in the same follow-up as deriving the quantization default itself, not this PR.

MakazhanAlpamys:
**Approved on the code — my block is resolved.** The only thing stopping the merge is a `CHANGELOG.md` conflict that #426 created underneath you about ten minutes ago. Nothing in `live_eval.py` or your test file conflicts.

## The block, and why it is lifted

I blocked because the parameter was threaded but nothing could reach it, while the CHANGELOG described five fixed commands. You took the "split honestly" option, and the entry now reads:

> **`live_eval.load_model_and_tokenizer` gains a `quantization` parameter; no live evaluation path sets it yet** … the four internal callers … still call it with no `quantization`, **so today an NF4-trained adapter is still judged against a bf16 base it never saw during training** … acceptance criteria 2 and 4 are left open.

I checked it against the code rather than taking it on trust: `grep -c quantization` returns **0** for `ship.py`, `advise.py`, `tunability.py` and `diagnose.py`, and the four internal call sites still pass nothing. The entry and the code agree, which is the whole point.

That headline sentence is the part I want to name. It would have been easy to write something that sounded like a fix; instead it states the surviving user-visible defect in plain terms in the release notes. That is harder to write and worth more.

## The two survivors are dead

The double-quant one I re-ran myself, since I was the one who found it surviving:

```
bnb_4bit_use_double_quant=True -> False
FAILED ...::test_4bit_builds_an_nf4_config_and_pins_device_map
1 failed, 10 passed        (was: 8 passed)
```

And `test_4bit_compute_dtype_comes_from_get_compute_dtype` is a new test written specifically for that wire. It matters more than it looks: `get_compute_dtype()` is what v0.73.1 shipped for the T4 trap, where `is_bf16_supported()` defaults to `including_emulation=True` so a T4 answers **True** and the first fix was a no-op on the exact hardware it was written for. Connecting a new 4-bit path to that function without a test would have been the quiet way to re-open it.

Also closed: acceptance criterion 1 now has `test_4bit_with_an_adapter_attaches_to_the_quantized_base` (adapter **and** quantization in the same call), and `_device_map_value` is reused rather than reimplemented, so a bare `"cuda"` cannot reach `torch.device(None)` — verified `'cuda' -> 0`, `'cpu' -> 'cpu'`.

Zero test deletions. The back-compat pin is intact. `live_eval.py` stays import-light — `torch`, `transformers` and `peft` all absent from `sys.modules` after import, with all three new imports inside functions.

## One thing to do

Rebase on `main` for the CHANGELOG. Sorry about the ordering — #426 and #457 landed in the last half hour and both touched `[Unreleased]`. That is a cost of my merge order, not anything on your side.

**Note that #367 stays open** when this merges, deliberately — criteria 2 and 4 are the user-visible half. If you want the follow-up it is yours; say so and I will note it on the issue so nobody else takes it.

AmirF194:
Rebased on main at 992343c. The CHANGELOG.md conflict with #426 is resolved: its Qwen3.5 MoE entry now sits above ours, both entries intact, nothing else touched. Re-ran the Docker verify (ruff clean, pytest on the new test file, 11 passed) against the rebased tree.

Thanks for the thorough review, and for catching that the wiring did not reach any caller. Noted that #367 stays open for criteria 2 and 4; happy to pick up the caller-wiring follow-up if nobody else has claimed it.

MakazhanAlpamys:
Merged as `de6445b`. **Verified again after your rebase** rather than assuming a rebase is a no-op — the mutation I found surviving in the first round still fails:

```
bnb_4bit_use_double_quant=True -> False
FAILED TestQuantizationReachesFromPretrained::test_4bit_builds_an_nf4_config_and_pins_device_map
1 failed, 10 passed
```

Diff after rebase: 3 files, **zero test deletions**, CHANGELOG at line 31 inside `[Unreleased]`. CI 12/12 on the true head, checked against `headRefOid` rather than the checks list.

## What I want on the record

I blocked this, and the thing that resolved it was not more code — it was you rewriting the claim to match what shipped. The entry now opens with **"no live evaluation path sets it yet"** and then states the surviving defect in plain terms: *"today an NF4-trained adapter is still judged against a bf16 base it never saw during training."*

That sentence is harder to write than a fix-sounding one, and it is the whole reason this merged. This project has shipped four flags that validated, documented and were read by nothing — v0.73.3 is named for that — and every one of them got there because a release note described the intent rather than the state.

**`#367 stays open`**, deliberately. Acceptance criteria 2 and 4 are the user-visible half.

## The two closed mutations

`test_4bit_compute_dtype_comes_from_get_compute_dtype` is the one I would keep. `get_compute_dtype()` is what v0.73.1 shipped for the T4 trap, where `is_bf16_supported()` defaults to `including_emulation=True`, a T4 answers **True**, and the first fix was a no-op on the exact hardware it was written for. Connecting a new 4-bit path to that function without a test would have been the quiet way to reopen it.

You also reused `_device_map_value` instead of reimplementing it — a bare `"cuda"` has index `None` and the next line calls `torch.device(None)`, which raises naming nothing the user can act on. This repo has been bitten repeatedly by two copies of one thing, so reuse over reimplementation is the right instinct.

Credited in `CONTRIBUTORS.md` (`0af605a`).

**On the follow-up**: wiring the four callers to a default derived from the run's own configuration is yours if you want it — say so on #367 and I will note it there so nobody else takes it. I see you have already picked up #273 as well; no rush on either.

AmirF194:
Thanks for re-verifying after the rebase rather than assuming it was a no-op. Appreciate the careful review throughout.

Issue #322, opened 2026-07-28
soup mcp serve: migrate to the mcp 2.x SDK API (the [mcp] extra is pinned <2)


MakazhanAlpamys:
## Context

`mcp` 2.0.0 shipped on 2026-07-28 and broke `soup mcp serve`'s test surface on every CI job the same day. Our constraint was an unbounded `mcp>=1.2.0`, so a fresh install resolved straight to it.

v0.72.3 capped the extra at `mcp>=1.2.0,<2` to unblock the release. That is a stopgap, not a fix — it pins anyone who wants `mcp` 2.x in the same environment.

## What broke

Two removals, both load-bearing for us:

- `mcp.shared.memory.create_connected_server_and_client_session` — gone. `tests/test_v07128.py::TestServerRoundTrip` builds every round-trip test on it.
- `Server.list_tools` — gone as an attribute (`AttributeError: 'Server' object has no attribute 'list_tools'`).

Verified before capping that 1.29.0 (newest 1.x) still exposes both, so the cap is at the right boundary rather than merely "the version that worked".

## Work

- [ ] Port `src/soup_cli/mcp_server/server.py` to the 2.x low-level API
- [ ] Rewrite `TestServerRoundTrip` against whatever replaces `create_connected_server_and_client_session`
- [ ] Decide the supported range — dual-support `>=1.2,<3` if the surfaces can be bridged, otherwise move the floor to 2.x and note it as a breaking change for the extra
- [ ] Lift the cap in `pyproject.toml`

## Acceptance

- [ ] `soup mcp serve` works against `mcp` 2.x, exercised by the existing round-trip tests
- [ ] The constraint no longer excludes 2.x
- [ ] A real stdio round-trip is asserted, not just that the server constructs

MakazhanAlpamys:
Taking this one. I wrote the SSE / streamable-HTTP transports in #479, so the MCP server is the code I have the freshest context on — and that PR is also the reason the scope here is slightly larger than this issue describes, which I'll get to.

**Measured the 2.0.0 surface before claiming**, by installing the wheel rather than reading release notes:

| what | 1.29.0 | 2.0.0 |
|---|---|---|
| `Server.list_tools` / `call_tool` **decorators** | yes | **gone** — registration moved into `Server(...)` as `on_list_tools=` / `on_call_tool=` callbacks |
[...]

PR #463, opened 2026-08-20
test: guard recipe count in documentation (#453)

### LINKED ISSUE #453: Nothing guards the recipe count in the five documentation sites, so two recipe PRs landing in sequence leave the docs silently one low

kok-o:
Closes #453

This PR adds a test to ensure that the recipe count mentioned in the user-facing documentation strictly matches the actual `len(RECIPES)`. The existing `test_catalog_size_is_N` assertions are left alone because they document historical growth, but this test prevents silent drift in the documentation itself.

### DIFF
--- a/tests/test_recipe_count_docs.py
+++ b/tests/test_recipe_count_docs.py
@@ -1,0 +1,43 @@
+import re
+from pathlib import Path
+from typing import List, Tuple
+
+from soup_cli.recipes.catalog import RECIPES
+
+
+def test_recipe_count_in_docs() -> None:
+    """Assert that the recipe count in documentation matches the actual number of recipes.
+
+    This prevents the count drifting silently out of date when recipes are added.
+    The four existing test_catalog_size_is_N assertions are left alone because they
+    document the historical release growth and catch backward drift, but this test
+    specifically guards the user-facing documentation numbers.
+    """
+    actual_count = len(RECIPES)
+    root = Path(__file__).parent.parent
+
+    # Each site is (file_path_relative_to_root, regex_pattern)
+    sites: List[Tuple[str, str]] = [
+        ("src/soup_cli/recipes/catalog.py", r"Recipe catalog \((\d+) recipes\)"),
+        ("CONTRIBUTING.md", r"\((\d+) recipes\)"),
+        ("docs/commands.md", r"List all (\d+) ready-made"),
+        ("docs/serving-and-export.md", r"(\d+) ready-made recipes"),
+        ("docs/serving-and-export.md", r"\((\d+) recipes\)"),
+    ]
+
+    for rel_path, pattern in sites:
+        file_path = root / rel_path
+        assert file_path.exists(), f"Doc file {rel_path} does not exist."
+        content = file_path.read_text(encoding="utf-8")
+
+        matches = re.findall(pattern, content)
+        assert (
+            matches
+        ), f"Pattern '{pattern}' stopped matching in {rel_path}. If the wording changed, update the test."
+
+        for match in matches:
+            found_count = int(match)
+            assert found_count == actual_count, (
+                f"Recipe count mismatch in {rel_path}: found {found_count}, expected {actual_count}. "
+                "Did you add a recipe and forget to update the docs?"
+            )


MakazhanAlpamys:
Thank you for this, and I am sorry — **I have to close it, and the reason is timing rather than anything about your code.**

#453 was claimed by @harshitthek on 19 Aug at 13:51 UTC, I confirmed it to him at 16:17, and his PR #457 has been open since 16:30 and has been through two rounds of review. Your PR arrived about 23 hours after that claim. Our [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md) asks for a comment on the issue before starting precisely so two people never spend the same evening on the same thing — and the cost of that landing on you is real, so I want to be straight about it rather than let the PR sit.

I approved CI on your branch anyway, so you can see your own tests run rather than staring at a blocked queue.

**On the code itself, since you deserve a real read and not just a "duplicate" label.** Your version is a clean, correct guard for the case it covers. The thing #453 was written around is the second half — a guard that stops guarding. If someone *rewords* a doc line so the number is no longer there at all, a check of the form "every count I find equals N" passes vacuously, because it finds nothing. #457 tracks which patterns failed to match and fails on that too, which is what makes it a ratchet rather than a snapshot. That distinction is not obvious and it is the only substantive difference between the two.

**I would rather you did not walk away over this**, so here are three that are genuinely unclaimed right now, closest first to what you just demonstrated:

- **#404** — a `soup ship --baseline` file carries no scorer stamp, so a snapshot taken before v0.73.2 is silently diffed against a *differently-scaled* score. Measured on an unchanged model the jump is 0.423 -> 0.731, far larger than the 0.05 gate. This is the same shape you just built — a guard against silent drift — on a defect I wrote into our own release's known-limitations list and did not fix. No GPU needed.
- **#405** — `mini_tool_call` now sits at 1.000 on a capable model. A suite pinned at the ceiling detects a regression exactly as poorly as one pinned at the floor. Needs some judgement about what to measure, which makes it more interesting than mechanical.
- **#322** — migrate `soup mcp serve` to the mcp 2.x SDK API; the `[mcp]` extra is pinned below it. Larger, dependency-driven.

Comment on whichever you want and it is yours — no race, no first-past-the-post. And #464 is a separate conversation; I am replying there now.

PR #464, opened 2026-08-20
test: guard cross-tokenizer boundary merge bias (#462)

### LINKED ISSUE #462: Cross-tokenizer acceptance is a lower bound: a merged prompt/generation boundary token costs exactly 1/n_gen, always downward
Split out of the #417 review (cross-tokenizer drafts) so it is not lost when that PR merges. Found while reviewing, measured with a purpose-built fixture rather than left hypothetical.

## What happens

On the cross-tokenizer path, `measure_acceptance` aligns the draft's text against the target's by character span. When the draft tokenizer merges the **last prompt character with the first generated one**, the token straddling that boundary drops out of the proposals and is scored as a miss — even for a draft whose logits reproduce the target text *exactly*.

A perfect draft, varying only the boundary merge:

```
control: clean boundary          accepted=2 total=2  alpha=1.000
merged boundary (b + c -> 'bc')  accepted=1 total=2  alpha=0.500
```

## Severity: exactly one target token, and always downward

The loss is `-1/n_gen`:

| n_gen | clean | merged | loss |
|---|---|---|---|
| 2 | 1.000 | 0.500 | -0.500 |
| 8 | 1.000 | 0.875 | -0.125 |
| 24 | 1.000 | 0.958 | -0.042 |

At the default `--max-new-tokens 64` that is about **-1.6%**.

**It can only under-report alpha.** A cross-tokenizer draft can therefore read slightly worse than it is — a false "this draft is not good enough" — but never a false green. That is the safe direction, which is why this was a nit in the review rather than a blocker, and why it is a follow-up rather than a fix in #417.

It does interact with `soup draft measure --min-acceptance`, which exits 2 as a CI gate: a draft sitting within ~1.6% above the threshold could be failed by the boundary token alone.

## Acceptance

- [ ] A docstring note on the cross-tokenizer branch of `measure_acceptance` stating that alpha is a **lower bound**, short by up to one boundary token (`1/n_gen`), and that the bias is always downward.
- [ ] The fixture above as a regression test: a perfect draft scoring `1.000` on a clean boundary and `1 - 1/n_gen` on a merged one — so the size of the bias is pinned and a future change that makes it *worse* (or flips it upward, which would be the unsafe direction) fails.
- [ ] Optionally: mention the bound in `docs/serving-and-export.md` next to the acceptance-rate explanation, where `--min-acceptance` is described.

Not blocking #417 — that PR's own scope is the cross-tokenizer measurement itself, and this is a known, bounded, safe-direction bias in it.

kok-o:
Closes #462.

**Note:** This PR is built on top of the cross-tokenizer draft feature from #417 (`CODING-DARSH:feat/draft-cross-tokenizer`) since the issue explicitly addresses a behavior introduced there. 

It includes:
- The docstring note on the cross-tokenizer branch of `measure_acceptance` stating that alpha is a lower bound.
- The fixture as a regression test pinning the downward bias when a boundary token is merged.
- A note in `docs/serving-and-export.md` next to the acceptance-rate explanation.

### DIFF
--- a/docs/serving-and-export.md
+++ b/docs/serving-and-export.md
@@ -347,6 +347,8 @@ Higher is better; roughly, ≥70% is where speculative decoding starts paying fo
 forward pass on realistic hardware. `--min-acceptance 0.6` exits **2** below the floor, so CI can
 gate on it (exit 0 = ok, 2 = below floor, 1 = error).
 
+*Note: For cross-tokenizer drafts, the measured acceptance rate is a strict lower bound. A token boundary merge between the prompt and the first generated token can cause the score to read up to `1/n_gen` lower than its true value, but it will never over-report.*
+
 `distill` runs logit KD through the existing `task: distill` trainer and emits a **dense** model
 (a PEFT adapter directory cannot be loaded as an `assistant_model`). When draft and target share a
 tokenizer, standard distillation is used. When tokenizers or vocabularies differ, `soup draft distill`
--- a/src/soup_cli/utils/draft.py
+++ b/src/soup_cli/utils/draft.py
@@ -528,6 +528,11 @@ def measure_acceptance(
     allowing speculative evaluation across different vocabularies and tokenization
     boundaries.
 
+    Note: Cross-tokenizer acceptance is a **lower bound**. A boundary merge (where
+    the draft tokenizer merges the last prompt character with the first generated
+    character) drops the straddling token, shortening the score by up to ``1/n_gen``.
+    The bias is always downward.
+
     Returns ``(accepted, total)`` summed over prompts.
     """
     import torch
--- a/tests/test_issue304_cross_tokenizer.py
+++ b/tests/test_issue304_cross_tokenizer.py
@@ -150,6 +150,22 @@ class TestCrossTokenizerAcceptanceKernel:
         assert count_accepted_spans(draft, target) == 4
         assert compute_acceptance_spans(draft, target) == 1.0
 
+    def test_boundary_merge_bias_is_pinned(self):
+        """Regression test for #462: a merged prompt/generation boundary token
+        costs exactly 1/n_gen, always downward."""
+        from soup_cli.utils.draft import compute_acceptance_spans, count_accepted_spans
+
+        target = ["b", "c"]
+        # control: clean boundary
+        draft_clean = ["b", "c"]
+        assert count_accepted_spans(draft_clean, target) == 2
+        assert compute_acceptance_spans(draft_clean, target) == 1.0
+
+        # merged boundary (b + c -> 'bc')
+        draft_merged = ["bc"]
+        assert count_accepted_spans(draft_merged, target) == 1
+        assert compute_acceptance_spans(draft_merged, target) == 0.5
+
     def test_repeated_substrings_and_interrupted_spans(self):
         """Verify repeated words and interrupted spans do not produce false positives."""
         from soup_cli.utils.draft import compute_acceptance_spans, count_accepted_spans
--- a/tests/test_issue304_cross_tokenizer.py
+++ b/tests/test_issue304_cross_tokenizer.py
@@ -155,16 +155,17 @@ class TestCrossTokenizerAcceptanceKernel:
         costs exactly 1/n_gen, always downward."""
         from soup_cli.utils.draft import compute_acceptance_spans, count_accepted_spans
 
-        target = ["b", "c"]
+        target = ["c", "d", "e", "f"]
         # control: clean boundary
-        draft_clean = ["b", "c"]
-        assert count_accepted_spans(draft_clean, target) == 2
+        draft_clean = ["c", "d", "e", "f"]
+        assert count_accepted_spans(draft_clean, target) == 4
         assert compute_acceptance_spans(draft_clean, target) == 1.0
 
-        # merged boundary (b + c -> 'bc')
-        draft_merged = ["bc"]
-        assert count_accepted_spans(draft_merged, target) == 1
-        assert compute_acceptance_spans(draft_merged, target) == 0.5
+        # merged boundary: the first generated character merges with the prompt tail
+        # and is excluded from the draft's proposal list.
+        draft_merged = ["d", "e", "f"]
+        assert count_accepted_spans(draft_merged, target) == 3
+        assert compute_acceptance_spans(draft_merged, target) == 0.75
 
     def test_repeated_substrings_and_interrupted_spans(self):
         """Verify repeated words and interrupted spans do not produce false positives."""


MakazhanAlpamys:
This one I want to **keep**, and the thing you did right is the reason: you disclosed in the PR body that it is built on top of #417. That was the honest move and it is what makes this a scheduling problem instead of a credit problem.

**The problem, stated plainly.** Your branch carries three commits that are not yours:

```
399708e  test: guard cross-tokenizer boundary merge bias (#462)   <- yours
92a2e9c  feat(draft): support cross-tokenizer drafts              <- @CODING-DARSH
833a5f3  Merge branch 'main' into feat/draft-cross-tokenizer      <- @CODING-DARSH
de5e671  test(draft): cover acceptance measurement edge cases     <- @CODING-DARSH
```

Because this PR targets `main`, merging it would land his entire unmerged feature — three weeks of work, currently in its fourth review round — under your pull request. He would lose the merge credit for it. That is not what you asked for and I do not think it is what you intended; it is just what `main` as a base does here.

**Some of this is my fault.** #462 says "not blocking #417" — I meant "this does not gate that PR", and I never wrote the thing that actually mattered: *wait until #417 merges, because the code you are annotating only exists on that branch.* You read it the only way it could reasonably be read.

## What I would like

Keep the PR open, and **#462 is yours** — I am recording that here so nobody else takes it while you wait. When #417 merges, rebase onto `main` and this becomes a one-commit PR containing exactly your work, which is what it should have been.

#417 is close. Its author fixed the last blocker earlier today; what remains is a CHANGELOG placement fix and green CI. I would expect days, not weeks. If it stalls I will tell you here rather than leave you guessing.

I considered asking you to retarget the base at his fork instead, which would show only your commit today — but that ties two people's branches together for a review that does not need it, and it needs his agreement too. Waiting is simpler and costs you nothing but time.

CI is approved on your branch, so you can watch your own tests run meanwhile.

## On the work itself

I have not reviewed it in depth yet — that happens properly once the diff is only yours. But from the file list you did take both halves of the acceptance criteria: the docstring note **and** the fixture. The fixture is the part I care about most, because a docstring saying "alpha is a lower bound" is a claim, and a test that pins the bias at exactly `1/n_gen` is a property. If a future change makes the bias *larger*, or flips it upward — which would be the unsafe direction, a draft reading better than it is — that test is what catches it.

Thanks for picking up an issue that was six hours old. Sorry the first one you sent landed on a claim you had no way to see from the PR list.

MakazhanAlpamys:
**#417 merged** — `0194fa4`, about twenty minutes ago. You are unblocked.

As planned: rebase onto `main` and this becomes a one-commit PR containing exactly your work. The three commits that were not yours are now in `main`, so they drop out of the diff on their own.

I said I would tell you here rather than leave you guessing, so: it took four days from when you opened this, and the last of that was a round on #417 where a clean textual merge of `main` had silently restored the very defect that PR fixed. Not your delay, but it was your wait.

Two things worth knowing before you rebase, both from that final round and both relevant to what you are annotating:

- The assisted-throughput arm is now called **once**, after the incremental report write, with `assistant_tokenizer=draft_tok`. If your branch carried the older two-call shape, it is gone from under you.
- The wiring is now pinned by `TestDraftMeasureCli::test_mismatched_tokenizer_measures_cross_tokenizer`, a CLI-level kwargs spy. If your docstring or fixture touches that path, that test is what will tell you.

**#462 is still yours** — nobody took it while you waited, which is what the note on the issue was for.

Once the diff is only your commit I will review it properly. From the file list you took both halves of the acceptance criteria — the docstring note *and* the fixture — and the fixture is the half I care about: a docstring saying "alpha is a lower bound" is a claim, a test pinning the bias at exactly `1/n_gen` is a property, and it is what catches a future change that makes the bias larger or flips it upward.

MakazhanAlpamys:
Seen — you rebased and the diff should now be just your commit. **In my queue, not forgotten.**

So you can self-check while it sits: the thing I will be looking at hardest is whether the regression test pins the **size** of the bias rather than merely its direction. A test asserting `merged < clean` passes just as happily if the bias grows to three tokens; the property worth having is `1 - 1/n_gen` exactly, so that a future change making it larger — or flipping it upward, which is the unsafe direction — fails.

I will also verify the diff really is one commit and nothing of @CODING-DARSH's is still riding along, since that was the original snag.

MakazhanAlpamys:
**The rebase worked perfectly** — one commit, yours, nothing of @CODING-DARSH's left:

```
c5ea974 | kok-o | test: guard cross-tokenizer boundary merge bias (#462)
non-kok-o commits in range: 0
```

That was the thing that went wrong originally and it is now clean. Docstring placement, scope discipline, and the decision to skip a CHANGELOG (correct — prior test-only guards `2c09b6d` and `008da45` added none) are all right too.

**But the test is red, on all 9 CI cells and locally, and the reason is worth understanding rather than patching.**

## Your fixture models a different mechanism than the bug

```
count_accepted_spans(['bc'], ['b','c']) = 2
  -> both sides decode to the SAME string: True
```

The kernel takes its `d_text == t_text` early return and yields `len(target_pieces)` = 2. **That is correct behaviour** — a draft that reproduced the target text exactly *should* score 1.0, regardless of how either side tokenized it. Your test asserts 1, so it fails, and it would be wrong to make the kernel agree with it.

The bias #462 describes is not a scoring difference. It is that the straddling token **is never proposed in the first place**: `measure_acceptance` slices `[draft_prompt_len - 1 : n - 1]`, so when the draft tokenizer merges the prompt tail with the first generated character, that character never enters the proposal list. I replayed it:

```
clean  : count_accepted_spans(['c','d'], ['c','d']) = 2 of 2   alpha 1.00
merged : count_accepted_spans(['d'],     ['c','d']) = 1 of 2   alpha 0.50
```

Those are the issue's measured numbers exactly. So the faithful fixture is **`['d']` against `['c','d']`** — a *shorter proposal list* — not two spellings of the same string.

## Second thing, and it is why I would not just take the one-line fix

At `n_gen = 2` your fixture cannot distinguish the claim from a coincidence:

```
n_gen=2:  drop-1-token -> 0.50    drop-half -> 0.50    <- indistinguishable
n_gen=4:  drop-1-token -> 0.75    drop-half -> 0.50    <- discriminates
```

#462's claim is specifically `1/n_gen`, and a 2-token fixture pins "50%" just as well as it pins "exactly one token". **Please use a 4-token target** — then the assertion genuinely encodes the property, and a future change that drops two tokens instead of one fails.

I checked what happens once the fixture is corrected: the guard works in **both** directions, including the unsafe one — a mutation flipping the bias upward (reporting full acceptance) is killed. So the design of your test is right; only the inputs are modelling the wrong thing.

## One smaller gap

The docstring half of #462's acceptance criteria is unpinned — deleting the note from `measure_acceptance` breaks nothing. Your test file quotes the claim in its own docstring (line 155), which is not the same as asserting it. Not a blocker, and arguably not worth a test at all; mentioning it so the decision is yours rather than an oversight.

## Where this leaves us

One fixture, widened to four tokens. Everything else about this PR is in order, and the part that is hardest to get right — using exact equality (`== 1`, `== 0.5`) instead of `merged < clean` — you already did.

Sorry this is a second round on a PR that already waited four days for #417. The mechanism here is genuinely subtle: I would not have spotted it from reading either, which is why I ran the kernel directly before writing this.

MakazhanAlpamys:
Merged as `41c7645`. **Your first contribution here, and it took two rounds — both about what the test proves rather than whether it passes.** That is the harder kind of feedback to receive, and you rebuilt it rather than patching around it.

I verified the new fixture myself:

```
target = ["c","d","e","f"]
clean  ['c','d','e','f'] -> 4   1.00
drop 1 ['d','e','f']     -> 3   0.75
drop 2 ['e','f']         -> 2   0.50
```

**0.75 is the number that makes the test mean something.** At 2 tokens, "dropped exactly one" and "dropped half" are both 0.50 — the assertion would have passed for a bias twice the size it claims to pin. At 4 they separate, so `1/n_gen` is genuinely encoded.

And it catches the direction that matters. I mutated the kernel to over-report acceptance — a draft looking *better* than it is, the unsafe direction #462 explicitly calls out:

```
return min(accepted + 1, len(target_pieces))
FAILED TestCrossTokenizerAcceptanceKernel::test_boundary_merge_bias_is_pinned
```

Fails by name, at the `== 3` assertion.

Local: **164 passed, 1 skipped** against `main`'s 163/1 — exactly your one test, nothing else moved. CI 12/12 on the true head. Zero test deletions. And the source diff is docstring-only, which I confirmed by AST comparison rather than by reading: identical with docstrings stripped, different with them kept. A guard, not a fix, exactly as intended.

## For the record, since this is your first merge

Three things you did that I would point at in someone's fifth PR, never mind their first:

- **You disclosed the #417 dependency in the PR body** when your branch carried three commits that were not yours. That turned a credit problem into a scheduling problem, and it is the only reason that PR is now merged under your name rather than closed.
- **You skipped the CHANGELOG**, correctly — prior test-only guards here (`2c09b6d`, `008da45`) added none, and you worked that out without being told.
- **You waited four days** for #417 to land rather than pushing for a workaround, and did not go quiet in the meantime.

One thing left unpinned, and it is genuinely optional: deleting the docstring note from `measure_acceptance` still breaks nothing. I mentioned it last round as your call and I meant it — a test that asserts on prose is often worse than the prose alone.

Welcome to `CONTRIBUTORS.md` (`26c33ef`). Sorry about the rocky start with #463 — the queue looks like a race from the outside and there is no way to see a claim from the PR list.

PR #466, opened 2026-08-20
feat(data): add provider sampling to best-of-n

### LINKED ISSUE #299: soup data best-of-n: provider-based (ollama/vllm) sampler backend
### Context
`soup data best-of-n` (shipped v0.71.31) samples the N candidate completions by loading `--base` locally via transformers (`sample_candidates` in `src/soup_cli/utils/best_of_n.py`). That bounds it to models that fit the local box (≤~3B on a 4 GB card) and requires the full model download.

### Ask
Add an optional provider sampler so N candidates can be drawn from a running ollama / vllm endpoint (mirroring `soup data evolve` / `gen-magpie`), keeping the local transformers path as the default.

### Fix path
- Reuse `utils/magpie.make_magpie_generate_fn(provider, model=, base_url=, temperature=)` — call it N times (or use the provider's `n`/`num_return_sequences` where supported) to get N candidates per prompt.
- Add `--provider ollama|vllm --model --base-url` flags to `commands/data.py::best_of_n`; when `--provider` is set, sample via the provider instead of `_load_bon_model`. `--base` stays the default local path.
- Judge + row-building (`judge_pick_best` / `build_sft_row` / `build_dpo_pair`) are unchanged.

### Acceptance criteria
- [ ] `soup data best-of-n --provider ollama --model <m> --prompts p.jsonl --n 8 --judge ollama://<j> -o out.jsonl` samples via the provider and writes SFT rows + provenance.
- [ ] Local `--base` path still works and is the default.
- [ ] SSRF validation on the provider URL; tests for the provider branch (mocked generate_fn).

Faisal01011:
## Summary
- add Ollama and vLLM sampling to `soup data best-of-n`
- preserve the existing local `--base` workflow
- validate provider URLs and record sampler provenance
- document provider usage and Anthropic's raw-completion limitation

## Tests
- `.venv/bin/python -m pytest -q --no-cov tests/test_cli_help_assertions_are_ansi_safe.py tests/test_v07131.py tests/test_v0716.py tests/test_cli_startup_is_light.py` (242 passed)
- `.venv/bin/ruff check src/soup_cli/ tests/`
- `git diff --check`

Closes #299

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ reproducing 70+ versions of notes.
   --config`, the registry entry, or the adapter's `adapter_config.json`), per the issue's own
   Fix path; that plus `soup ship` reporting the numerics it judged with and a staleness gate
   on mismatched-numerics evidence are left open (issue acceptance criteria 2 and 4).
+
 - **`training.stream_pin` makes layer-streaming pinning configurable (#366 by @ousamabenyounes in #416).**
   Page-locking the RAM store is chosen automatically by `decide_pinning`, and
   until now nothing could override it — so while #331 was live, `pin=False` was
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,14 @@ reproducing 70+ versions of notes.
 
 ### Added
 
+- **`soup data best-of-n` can sample candidates from Ollama or vLLM providers
+  (#299 by @Faisal01011 in #466).** The existing local Transformers `--base` path stays
+  the default, while `--provider ollama|vllm --model <m> [--base-url <url>]`
+  draws each prompt's N candidates through the existing SSRF-validated raw-
+  completion seam. Provider and model are recorded in `_best_of_n` provenance;
+  Anthropic is refused by name because its Messages API has no raw-completion
+  endpoint, and local-only flags cannot be silently ignored in provider mode.
+
 - **Layer streaming now accepts Qwen3.5 MoE text checkpoints whose decoder
   layers do not expose exactly the same weight keys in every block (by
   @Shutaru in #426).** The sharder now records per-layer shard headers instead of
@@ -38,7 +46,6 @@ reproducing 70+ versions of notes.
   --config`, the registry entry, or the adapter's `adapter_config.json`), per the issue's own
   Fix path; that plus `soup ship` reporting the numerics it judged with and a staleness gate
   on mismatched-numerics evidence are left open (issue acceptance criteria 2 and 4).
-
 - **`training.stream_pin` makes layer-streaming pinning configurable (#366 by @ousamabenyounes in #416).**
   Page-locking the RAM store is chosen automatically by `decide_pinning`, and
   until now nothing could override it — so while #331 was live, `pin=False` was
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -297,7 +297,7 @@ soup local-rl train --db <path> --model <id> [--scheduler-dir <dir>] [--hour H] 
 soup build <manifest.yaml> [--dry-run] [--output-dir <dir>]  dbt-for-SFT DAG: validate + plan + live materialise (v0.69.0; live v0.71.6)
 soup expect <data.jsonl> <suite.yaml>         Expectations suite: PII / token-length / refusal / judge (v0.69.0)
 soup data gen-magpie --base <m> --provider ollama|vllm --target N --output <jsonl> [--base-url <url>] [--quality-filter]  Magpie synthetic generator — live (v0.69.0; live v0.71.6)
-soup data best-of-n --base <m> --prompts <jsonl> --n 8 --judge <url> -o <sft.jsonl> [--emit-pairs <dpo.jsonl>]  Best-of-N rejection sampling: sample N locally, judge picks winner -> SFT (+ DPO) rows (v0.71.31)
+soup data best-of-n (--base <m> | --provider ollama|vllm --model <m> [--base-url <url>]) --prompts <jsonl> --n 8 --judge <url> -o <sft.jsonl> [--emit-pairs <dpo.jsonl>]  Best-of-N rejection sampling: sample N locally (default) or through a raw-completion provider, then emit SFT (+ DPO) rows
 soup data evolve --input <seeds.jsonl> --provider ollama|vllm --model <m> --strategy depth|breadth --rounds N -o <jsonl>  Evol-Instruct (WizardLM) instruction evolution (v0.71.31)
 soup data persona-mix --prompts <jsonl> --n N --output <jsonl>  Persona-Hub diversity sampler (v0.69.0)
 soup data brain-rot <data.jsonl> [--strict]   Brain-rot detector — arXiv 2510.13928 (v0.69.0)
--- a/docs/data.md
+++ b/docs/data.md
@@ -185,17 +185,22 @@ soup data persona-mix --prompts prompts.jsonl --n 500 --output mixed.jsonl
 # Brain-rot detector (arXiv 2510.13928) — refuses to train on excessive slop
 soup data brain-rot data.jsonl --strict --max-major-fraction 0.10
 
-# Best-of-N rejection sampling (v0.71.31) — sample N locally, a judge picks the winner
+# Best-of-N rejection sampling — local sampling stays the default
 soup data best-of-n --base HuggingFaceTB/SmolLM2-135M-Instruct \
     --prompts prompts.jsonl --n 8 --judge ollama://llama3.1 \
     -o best_of_n.jsonl --emit-pairs pairs.jsonl
+
+# Or draw the N candidates from a running Ollama / vLLM raw-completion endpoint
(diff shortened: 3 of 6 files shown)


MakazhanAlpamys:
**This does not ship a fifth "validated, documented, read by nothing" flag** — I checked that first, since it was the warning I gave you when you claimed the issue, and the provider path is genuinely wired. I ran the mutation myself:

```
best_of_n.py:50   if generate_fn is not None:  ->  if False:
FAILED TestBestOfN::test_provider_generate_fn_is_called_n_times
FAILED TestBestOfNCli::test_provider_path_calls_sampler_and_records_provenance
2 failed, 86 passed
```

Two layers, and the CLI one asserts on the **exact args handed to the factory** and that the sampler was invoked n=3 times — not that a flag parsed. Better still, it monkeypatches `_load_bon_model` to `raise AssertionError("provider mode must not load the local model")`, which is a negative control: it fires if the local path is taken silently.

**`anthropic` is rejected by name, before the factory is constructed**, and case-folded:

```
'anthropic' / 'ANTHROPIC' / 'Anthropic'
  -> 'anthropic has no raw-completion endpoint; --provider must be ollama or vllm'
'openai'
  -> '--provider must be ollama or vllm'
```

That is *more* specific than the precedent I pointed you at — `soup data evolve --provider anthropic` emits magpie's generic message naming the wrong feature. Yours names the flag the user typed.

**The local `--base` path is still the default and is pinned three ways**, including a mirror-image negative control (`make_magpie_generate_fn` patched to raise in local mode) and an assertion that local rows carry **no** `sampler` key. Routing the no-provider case through the provider path fails `test_happy_path` and `test_plan_only`.

**SSRF is inherited, not bypassed** — you reused the existing validators rather than hand-rolling, and `--plan-only` still constructs the sampler so validation is not skipped:

```
http://169.254.169.254/         -> refused, 'must be localhost'
file:///etc/passwd              -> refused, 'must use HTTP or HTTPS scheme'
http://metadata.google.internal -> refused
http://127.0.0.1:11434          -> ok
```

And the seam reuses `make_magpie_generate_fn` — zero new HTTP code. This repo has been bitten repeatedly by two copies of one thing (#372, #392, the `process_liveness` consolidation), so that matters.

## One finding — MEDIUM

**Your four argument-combination guards have no test coverage at all.** I disabled six guard conditions at once:

```
conditions neutered: 6
88 passed
```

Nothing failed. All four *work* — I verified each live:

```
--base + --provider                         -> '--base and --provider are mutually exclusive'
--provider + --device/--seed/--trust-remote -> '--device, --seed and --trust-remote-code apply only to local --base'
--model/--base-url without --provider       -> '--model and --base-url require --provider'
neither --base nor --provider               -> 'either local --base or --provider is required'
```

**The second is the one I actually care about**, and it is the one with no test: it is precisely the "flag silently ignored" failure this repo keeps getting bitten by — `--seed` in provider mode — applied to the exact thing you were warned about. If that guard regressed, a user passing `--seed 42` with `--provider ollama` would get a silently unseeded run, and your own PR would have re-created the class it was careful to avoid everywhere else.

Four one-line CLI tests, one per guard, asserting exit code and the message keyword. That is the ask.

## Two small ones

- **CHANGELOG suffix** reads `(#299 by @Faisal01011)`; house form is `(#299 by @Faisal01011 in #466)`.
- **Nit, take or leave**: provenance is written in two places — `build_sft_row` produces `_best_of_n`, then `commands/data.py` mutates `row["_best_of_n"]["sampler"]` afterwards. A `sampler=` kwarg on `build_sft_row` would keep one owner of that dict. Related: `--emit-pairs` DPO rows carry no sampler provenance at all, which may or may not be intended.

## Rebase

You conflict on **`CHANGELOG.md` only** — I checked the overlapping file set and that is the whole of it. Six things landed under you today (#320, #453, #426, plus contributor commits), so re-place the single `### Added` bullet under `[Unreleased]` and nothing else needs touching.

Docs are updated in both `docs/commands.md` and `docs/data.md`, one test line was **strengthened** not weakened, import-lightness holds (23 passed), ruff clean, CI 12/12 on the true head.

MakazhanAlpamys:
**Approved — my finding is fully closed, and I verified it myself rather than reading the diff.** One thing stops the merge and it is procedural: **this PR is still marked as a draft.** I will not flip that for you; mark it ready for review and I will merge it.

## The finding, closed 4/4

I disabled the `--seed` / `--device` / `--trust-remote-code` guard:

```
if device or trust_remote_code or seed != 0:  ->  if False:
FAILED TestBestOfNCli::test_rejects_local_only_flags_with_provider
1 failed, 91 passed
```

That is the guard I said mattered most — `--seed` silently ignored in provider mode is exactly the "validated, documented, read by nothing" class v0.73.3 is named for — and it now has a killing test with clean one-to-one attribution. Previously disabling all six conditions left 88 tests passing.

**And the control is the part I care about more.** A test asserting only `exit_code == 2` passes when the *wrong* guard fires, so I kept `raise typer.Exit(2)` and replaced only the message text:

```
"[red]--device, --seed and --trust-remote-code apply only to local --base[/]"  ->  "[red]bad arguments[/]"
FAILED TestBestOfNCli::test_rejects_local_only_flags_with_provider
```

Still fails. Your tests discriminate on the reason, not the exit code. They also route output through a `_plain()` ANSI-strip helper, which is the correct handling of the Rich/Typer per-character-ANSI gotcha that has turned this repo's CI red four times.

Everything from the first round survived the rebase: ignoring the provider still kills 2 tests, `anthropic` is still rejected by name through `.strip().lower()` (I checked `Anthropic`, `ANTHROPIC`, `  AnThRoPiC  `), and the hostile `--base-url` set is still refused — including `http://localhost.evil.com`, which is the prefix-match bypass a hand-rolled check would miss.

One test line deleted, and it is a strengthening: a raw-output assertion replaced by an ANSI-stripped one plus three new flag assertions.

CI 12/12 on the true head, `docs/commands.md` and `docs/data.md` both updated, CHANGELOG suffix correct, import-lightness 34 passed.

## One LOW observation, not a blocker

Guard #2 tests `seed != 0` rather than "was `--seed` supplied", so an explicit `--seed 0` in provider mode is accepted silently:

```
--provider ollama --model m --seed 0 --plan-only   ->  exit 0
--provider ollama --model m --device cpu           ->  exit 2
```

Harmless in practice — `0` is the default and the seed is a genuine no-op on the provider path — but it is the same *shape* as the class this guard exists to catch. The precise form is Click's `ParameterSource`, which `soup ship --config` already uses to tell "explicitly passed" from "left at default". Worth tightening whenever you next touch this, not worth holding a green PR for.

Mark it ready and it goes in.

MakazhanAlpamys:
Merged as `85eaad5`. Thanks for taking the draft off — I do not flip that for people, since it is the author's signal about readiness rather than mine.

**Everything I verified stands on the same head I approved** (`808dcb9`), and I ran the two checks myself rather than reading the diff:

```
if device or trust_remote_code or seed != 0:  ->  if False:
  FAILED TestBestOfNCli::test_rejects_local_only_flags_with_provider    1 failed, 91 passed

message -> "[red]bad arguments[/]", raise typer.Exit(2) kept:
  FAILED TestBestOfNCli::test_rejects_local_only_flags_with_provider
```

**The second is the one that makes the tests worth having.** A test asserting only `exit_code == 2` passes when the *wrong* guard fires — which is a real failure mode here, since all four guards exit 2. Yours fail when only the message changes, so they pin the reason.

## What I want on the record

Three decisions that were yours, not asked for:

- **You reused `make_magpie_generate_fn` instead of writing a second HTTP client.** That is why `--base-url` inherits the existing SSRF validators for free — including `http://localhost.evil.com`, which a prefix check would have let through. This repo has been bitten repeatedly by two copies of one thing (#372, #392, the `process_liveness` consolidation), so reuse over reimplementation is the instinct I most want to see.
- **`anthropic` is rejected by name and case-folded**, and your message names the flag the user typed rather than the feature. `soup data evolve` emits magpie's generic message naming the wrong feature; yours is better than the precedent I pointed you at.
- **The local `--base` path stayed the default and is pinned three ways**, including a mirror-image negative control that raises if a provider sampler is constructed in local mode. Widening a command without disturbing its existing behaviour is the boring part that usually gets skipped.

Full run: `92 passed` against `main`'s 84 — exactly your 8 tests, nothing else moved. One deleted test line, and it is a strengthening: a raw-output assertion replaced by an ANSI-stripped one plus three new flag assertions. Docs updated in both `docs/commands.md` and `docs/data.md`. CI 12/12 on the true head.

## One thing left, take it or leave it

Guard #2 tests `seed != 0` rather than "was `--seed` supplied", so an explicit `--seed 0` in provider mode is accepted silently:

```
--provider ollama --model m --seed 0 --plan-only   ->  exit 0
--provider ollama --model m --device cpu           ->  exit 2
```

Harmless — `0` is the default and the seed is a genuine no-op on that path — but it is the same *shape* as the class the guard exists to catch. The precise instrument is Click's `ParameterSource`, which `soup ship --config` already uses to tell "explicitly passed" from "left at default". Not worth a round; worth knowing.

Credited in `CONTRIBUTORS.md` (`749b950`).

PR #468, opened 2026-08-20
feat(data): extend data.interleave to streaming and hub datasets

### LINKED ISSUE #459: Extend data.interleave to streaming and hub datasets (deliberately out of scope in #443)

blackcoderx:
Closes #459.

#443 wired `data.interleave` into training-time dataset loading, but scoped it to a `data.train` list of local file paths only: `data.streaming: true` and any remote-URI / HF-hub-dataset-name list entry were refused at parse time with a message naming this issue as the follow-up.

Every `data.train` list entry is now classified once (local file / remote URI / HF-hub name) and dispatched:

- **An all-local/remote list with `data.streaming: true`** delegates combining to HF `datasets.interleave_datasets` / `concatenate_datasets` instead of reimplementing mixing over a source whose size can't be known ahead of time. `concat` → `concatenate_datasets`; `under`/`over` → `stopping_strategy="first_exhausted"`/`"all_exhausted"`; `probs` → `probabilities=probs` — chosen so the strategy names mean the same thing as the local path. Verified by a test that runs one `probs` config through both paths and compares the resulting proportions, not two tests that each pass alone.
- **An all-HF-hub-dataset-name list** loads each entry eagerly and combines with the *same* `_combine_interleaved` the local path uses. A hub entry's own `validation` split is honoured for the combined result only when *every* entry provides one; otherwise it's ignored (warned) and `data.val_split` applies to the combined train rows — a decided precedence, documented in `docs/data.md`, not an emergent one.
- **Still refused, by name:** an all-hub list with `data.streaming: true` (streaming N differently-shaped hub datasets and reconciling their splits is a separate, larger effort), and any list mixing hub names with local/remote entries.

`_load_hf_dataset` is factored into `_load_one_hub_dataset` (mirrors #443's `_load_one_local_dataset` factoring) so the hub interleave path reuses it rather than re-deriving it; the single-name branch stays byte-identical, pinned by a golden test. `MAX_REMOTE_ROWS` is factored out of `_load_remote_dataset` into a shared module constant reused by the new streaming-interleave loader.

## Testing
- New `tests/test_issue459_interleave_streaming_hub.py` (18 tests): schema dispatch matrix, strategy→HF-call mapping for each of concat/under/over/probs, the local-vs-streaming proportions acceptance test, hub combining + the three validation-split precedence cases, and a golden test for the `_load_hf_dataset` refactor.
- Updated the two `tests/test_issue443_interleave_wiring.py` tests that pinned the old refusal behavior this PR intentionally changes.
- Full suite: `pytest tests/ -v --tb=short` → **17,912 passed, 393 skipped, 5 deselected, 0 failed**.


MakazhanAlpamys:
Seen — in my queue, and thank you for turning this around quickly after #460.

So you know what I will be holding it to, since I pinned these in #459 before you started: (1) a test that runs **one config both ways** — streaming and not — and compares the resulting proportions, rather than two tests that each pass in isolation; (2) the `val_split` precedence you chose, stated in the PR body and pinned by a test; (3) the parse-time refusals removed **only** for what actually landed, so anything unimplemented still refuses by name.

And the #460 rule that generalises: a test asserting the config parses would reproduce the bug rather than catch it. I will mutate the streaming branch to load only the first dataset and see what fails.

MakazhanAlpamys:
**The three things I pinned in #459 are met, and item 1 is met more strongly than the PR claims.** Two parse-time gaps hold it, both of the same shape — schema accepts, loader dies — which is exactly what item 3 exists to prevent.

## Item 1 is satisfied against *real* HF semantics, not just the fake

Your semantics test discriminates — reversing the probabilities kills it, swapping `under`/`over` kills the strategy tests. But it runs against a fake `datasets`, so the fake defines what `interleave_datasets(probabilities=...)` means. I ran the same four strategies through **real `datasets` 4.5.0**, sources A=600 / B=200:

| strategy | local n | local A% | stream n | stream A% |
|---|---|---|---|---|
| `concat` | 800 | 75.0% | 800 | 75.0% |
| `under` | 400 | 50.0% | 400 | 50.0% |
| `over` | 1200 | 50.0% | 1199 | 50.0% |
| `probs [0.7,0.3]` | 800 | 70.0% | 683 | 70.7% |

The proportions genuinely match. That was the criterion and it holds.

**One thing the docs under-state:** `probs` streaming yields **683 rows against 800, −14.6%**, because streaming stops at first-exhausted. The docs say "converges to the same ratio, sampled rather than exact" but never that an *epoch is 15% smaller* when you flip one flag. Worth a sentence — it changes how many optimizer steps a run takes.

Item 3 is otherwise clean: streaming and hub both landed, and every unimplemented combination still refuses **by name** — all-hub + streaming, mixed local/hub, mixed remote/hub. Item 6 holds too: the #443 headline mutation still kills 12 tests.

## Finding 1 — three of five supported extensions break under streaming

```python
ds = hf_load("json", data_files=entry, split="train", streaming=True)
```

The builder is hardcoded, but `SUPPORTED_EXTENSIONS = {".jsonl", ".json", ".csv", ".parquet", ".txt"}`. Flipping only `streaming: true`:

```
.csv      streaming=False -> OK, 5 rows
.csv      streaming=True  -> ArrowInvalid: JSON parse error: Invalid value. in row 0
.txt / .parquet: same
```

The error names neither `streaming`, nor `interleave`, nor the format. **This is new surface, not inherited** — on `main`, `streaming: true` on a local file is silently ignored and the local loader handles the CSV fine. #468 is the first place `data.streaming` changes *which loader a local file goes through*. Per item 3, csv/parquet/txt under streaming should refuse by name, or the builder should be chosen from the suffix.

## Finding 2 — the classifier misreads your own docs example

`_classify_train_entry` calls anything with a `Path(...).suffix` a local file:

```
local    teknium/OpenHermes-2.5
local    mlfoundations/dclm-baseline-1.0
hub      HuggingFaceFW/fineweb-edu
hub      tatsu-lab/alpaca
```

`teknium/OpenHermes-2.5` is the example **this PR adds** at `docs/data.md:550`. Paired with another hub name it gives `FileNotFoundError: teknium\OpenHermes-2.5`; paired with a dotless one it gives the confusing `not a mix (got kinds=['hub','local'])`.

The rule is inherited from #443, but there it was **inert** — every hub name was refused, so misclassifying one changed nothing. Your PR makes hub names a supported input, and the rule now produces a wrong refusal for a very common name shape (any dataset with a version in its name).

Worth saying: the hub path itself **works against the real Hub** — `tatsu-lab/alpaca` + `Open-Orca/SlimOrca` gave 104,004 rows = 2 x min(52,002, 517,982), exactly `under`. The mechanism is right; only the classifier is wrong.

## Finding 3 (LOW) and one nit

Empty source, same config, flag flipped: local gives your named `ValueError`, streaming gives a bare `StopIteration` with an empty message.

**`CHANGELOG.md:75` reads `(#459 by @blackcoderx in #TODO-PR)`** — should be `#468`.

## Diff integrity — clean, with one thing I want to flag

15 deleted test lines, all accounted for: the two #443 refusal tests are **replaced**, not dropped, by `..._streaming_local_files_now_accepted`, `..._streaming_hub_list_still_rejected_with_reason` and `..._mixed_local_and_hub_entries_rejected_with_reason`, each matching on text unique to its reason.

But the deleted block included **my own review note** explaining why `match="disable streaming or drop"` beat both `match="#459"` and `match="streaming"` — the latter because pydantic's `ValidationError` string embeds `input_value={...}` containing the literal `'streaming': True`. Your replacements happen to pick safe strings, but that reasoning is now gone from the file, and the next person writing a refusal test there will rediscover it the hard way. Worth carrying forward as a comment.

`val_split` precedence is decided, documented **and pinned in both directions** — `all(has_val)` → `any(...)` kills a test. That was item 2 and you closed it properly.

`ruff` clean, import-lightness 23 passed, CI 13/13 on the true head, full suite 18164 passed / 0 failed.

## What I need

The `#TODO-PR` placeholder, plus a decision on Findings 1 and 2. Both are the same contract: a config that parses must not die in the loader. I would take a parse-time refusal for csv/parquet/txt-under-streaming as readily as a real fix, but the classifier one probably wants fixing rather than refusing, since it hits ordinary dataset names.

MakazhanAlpamys:
Checking in — this has been green and mergeable for two days and I do not want it drifting.

**Where it stands:** the three things I pinned in #459 before you started are all met, and item 1 more strongly than the PR claimed. CI is 13/13 and the branch is clean against `main`. What is left is the two MEDIUMs from my review, and both are the same shape — a code path that is right for the case you tested and wrong for a neighbouring one:

1. **The hardcoded JSON builder.** Under streaming it assumes JSON, so `.csv` / `.txt` / `.parquet` take a path that cannot work for them. `data.train` accepts all of those today.
2. **The local-vs-hub classifier.** `teknium/OpenHermes-2.5` is treated as a local file — and that is the example in your own PR description, so the documented usage hits the wrong branch.

Neither is large, and neither needs a GPU.

**If you would rather not carry both PRs at once**, say so — you also have #471 open, and that one now needs a rebase onto the shared helper #492 introduces, which is not a trivial ask on top of this. I am happy to:

- take one of the two MEDIUMs myself on your branch with your authorship intact, or
- park #471 and land #468 first, or
- leave both with you if you would rather finish them yourself — just tell me which.

What I do not want is for two good PRs to stall because the queue on you got long. Tell me what is useful and I will do that.

MakazhanAlpamys:
﻿Reviewed independently (tests run against your branch in an isolated worktree, mutation placed). Verdict: **revise** — architecture is right and the three acceptance criteria from #459 are met, but review found one new HIGH that my earlier pass missed:

**HIGH - SSRF allowlist bypass in the new streaming path.** `_load_interleaved_streaming_datasets` feeds each entry raw to `hf_load("json", data_files=entry, ...)`. Your schema gate only checks `is_remote_uri()` (scheme membership), but the sibling single-URI path canonicalizes through `validate_remote_uri()` first - which rejects userinfo/query/fragment, query strings being explicitly documented as SSRF-adjacent in data_pipeline.py. Demonstrated executable: `s3://bucket/key.jsonl?endpoint=http://169.254.169.254` passes your gate, fails `validate_remote_uri`, parses as config, reaches hf_load raw. Please canonicalize every remote-classified entry before dispatch.

Two more from my earlier notes, confirmed real:
- hardcoded `"json"` builder under streaming flips `.csv/.txt/.parquet` from working to ArrowInvalid - refuse by name or pick builder from suffix;
- `_classify_train_entry` treats any dotted name as local, so `teknium/OpenHermes-2.5` (your own docs example) misroutes;
- changelog: please move to `changelog.d/0.73.3/468.added.md` (the direct CHANGELOG.md edit also conflicts with main now).

Mutation check passed on my side: silently ignoring interleave in the hub-list combine fails 4 of your tests. The val_split precedence decision is documented and pinned both directions - good call writing it down.

blackcoderx:
Fixed the HIGH — `_load_interleaved_streaming_datasets` now canonicalizes every remote-classified entry through `validate_remote_uri` before it reaches `hf_load`, mirroring `_load_remote_dataset`'s existing guard exactly. Added two regression tests: one confirms a query-bearing URI (e.g. the `169.254.169.254` example) is rejected before `hf_load` is ever called, one confirms the canonicalized string — not the raw input — is what's dispatched. Verified with a mutation check (reverting the fix fails both new tests) plus the full suite (18423 passed, 0 failed).

The two MEDIUMs (csv/txt/parquet under streaming, hub-name misclassification) and the changelog.d migration are still in progress — will follow up on this same branch.

blackcoderx:
Fixed the first MEDIUM — `_load_interleaved_streaming_datasets` now picks the HF `datasets` builder per entry from its file suffix (`_streaming_builder_for`, mirroring `load_raw_data`'s `SUPPORTED_EXTENSIONS` dispatch for the local path), instead of hardcoding `"json"`. `.csv`/`.parquet`/`.txt` now go through the matching builder under streaming; an unrecognized suffix refuses by name before reaching `hf_load` rather than failing inside it with an opaque Arrow error. Suffix is read from the canonicalized (post-`validate_remote_uri`) entry, not the raw one, so it can't be confused by a query string.

Added a parametrized regression test (builder matches suffix for csv/parquet/txt) and one for the unsupported-extension refusal. Full suite: 18427 passed, 0 failed.

One MEDIUM (hub-name misclassification) and the changelog.d migration are still in progress.

blackcoderx:
Fixed the second MEDIUM — a suffix now has to be one of SUPPORTED_EXTENSIONS to classify as local (reusing the same allowlist load_raw_data / _streaming_builder_for already use), instead of "any non-empty Path.suffix." `teknium/OpenHermes-2.5` and `mlfoundations/dclm-baseline-1.0` (your repro names) now classify as hub at both the schema and loader level. Added regression tests at both layers, each written to actually discriminate old vs. new classification (paired with `streaming=true`, since a plain non-streaming config validates under either rule and wouldn't catch a misclassification). Also updated the extension-refusal test from the prior fix, since a local `.xyz` entry now classifies as hub under the tightened rule — moved its bad entry to a remote URI so it still exercises the intended loader-level refusal.

Also moved the changelog entry to `changelog.d/0.73.3/468.added.md` and corrected `#TODO-PR` → `#468`, per the changelog.d convention now on main. Validated against the repo's own fragment validator and `test_repository_fragments_match_the_newest_release`.

Full suite: 18429 passed, 0 failed. This closes out all four findings from the review.

MakazhanAlpamys:
﻿All three fixes verified independently - canonicalization confirmed end-to-end (the query-string URI is refused by name before hf_load ever sees it, negative control passes through canonicalized), suffix-builder mapping covers exactly SUPPORTED_EXTENSIONS and refuses unknown by name, dotted hub names classify correctly, fragment validator green, 60 targeted tests passing. Mutation check: neutralizing validate_remote_uri kills two tests. Well done.

One more from my reviewer, in the same family as your commit 1:

**MEDIUM - non-allowlisted URI schemes with a supported suffix still reach hf_load raw.** `_REMOTE_SCHEMES` covers abfs/abfss/az/gcs/gs/oci/s3 but not https/http/ftp, so `_looks_like_remote_uri("https://example.com/data.jsonl")` returns False, the entry classifies as local, your new guard never fires - and the probe shows it dispatched unvalidated, while the single-URI path refuses the identical URL by name (`scheme 'https' not in allowlist`). So the list path bypasses exactly the policy commit 1 was extending. Suggested shape: treat any `"://"`-bearing entry as needing validation and refuse non-allowlisted schemes by name before hf_load (plus one test for an https entry refused, one for a legit s3/gcs entry still passing).

The query-string variant only survives today by accident (pathlib reads `.254` as the suffix -> classified hub -> schema mix refusal) - fragile coincidence rather than a guard, worth closing deliberately while you are here.

MakazhanAlpamys:
﻿@blackcoderx - gentle ping: the one remaining item on your #468 is the https/http scheme MEDIUM from my review (non-allowlisted schemes with a supported suffix still reach hf_load raw through _looks_like_remote_uri; suggested shape: treat any "://"-bearing entry as needing validation and refuse non-allowlisted schemes by name). Everything else is verified and merged-quality - this is the last blocker between you and a merge.

blackcoderx:
Fixed — traced this to `_looks_like_remote_uri` conflating "is this meant to be a remote source" with "is the scheme allowlisted" (it delegated to `is_remote_uri`, which is scheme-allowlist-only). Since `_REMOTE_SCHEMES` doesn't cover https/http/ftp, those entries fell through to 'local' classification and reached `hf_load` with zero validation — not just missing the allowlist check, missing `validate_remote_uri` entirely.

`_looks_like_remote_uri` is now a plain `"://"` sniff; the scheme allowlist stays enforced exactly where it already lived, in `validate_remote_uri`. `is_remote_uri` itself is untouched — only the 4 call sites that used it for classification/dispatch (list-entry classifier, schema's `_kind` mirror, the streaming validate-gate, and `load_dataset()`'s single-path dispatch) now go through the broadened sniff.

One thing worth flagging: I couldn't reproduce "the single-URI path refuses the identical URL by name" as *existing* behavior — tracing it, a single (non-list) `data.train: "https://..."` string was falling through to the plain local-file loader and failing with a confusing `FileNotFoundError` on the mangled path, not `validate_remote_uri`'s message. Same root cause as the list path though, so I fixed it the same way rather than leaving it inconsistent — it's now refused by name too, which I think is what you were expecting to already be true. Let me know if you'd rather that stayed out of scope.

Added 4 tests (https refused in a streaming list before hf_load is reached, https refused on the single-path case, a legit gcs entry unaffected, schema-level classification check) plus a mutation check — reverting the sniff fails all three https-focused tests, one surfacing the exact FileNotFoundError above. Full suite: 18433 passed, 0 failed.

MakazhanAlpamys:
Merged as e19423a. Thank you, Emmanuel - this closes out #459 cleanly.

What you did well: you traced the MEDIUM to its actual cause instead of patching where it hurt. `_looks_like_remote_uri` delegating to `is_remote_uri` was conflating two different questions - "is this meant to be remote" and "is the scheme allowlisted" - and your fix separates them exactly right: a plain `://` sniff for classification, with the allowlist staying enforced in `validate_remote_uri` where it already lived, and `is_remote_uri` untouched for its own callers. Also appreciated that you flagged the single-path dispatch fix as an honest extension beyond the reported scope.

What I ran before merging:
- 260 targeted tests across test_issue459_interleave_streaming_hub.py / test_issue443 / test_v0420 / test_v0538 - all green, ruff clean.
- Mutation 1: reverted `_looks_like_remote_uri` back to delegating to `is_remote_uri` - the https hole reproduced exactly as your commit describes: `test_streaming_https_entry_rejected_by_name` failed DID-NOT-RAISE (https reached hf_load unvalidated) plus `test_single_https_train_rejected_by_name`. Your fix demonstrably guards it.
- Mutation 2: added ftp to the scheme allowlist - died by name on `test_validate_remote_uri_unknown_scheme`.
- CI: 14/14 cells green.

One small correction for the record, harmless to the code: reverting the loader sniff alone fails two of the three https tests, not all three - the third (`test_https_entry_classified_remote_not_local_or_hub`) is guarded independently by the schema's inline `_kind` copy. Your fix covers all of them; just the commit message overcounted slightly.

CONTRIBUTORS.md updated with this entry. Next time you are looking for work: #405 (`mini_tool_call` pinned at the 1.000 ceiling) is verified open and unclaimed.

Issue #273, opened 2026-06-20
Cover three uncovered branches in pure-Python helpers (good first issue)


MakazhanAlpamys:
> **Re-scoped 2026-08-04.** The original version of this issue asked for test
> files for `run_cost.py`, `warmup.py` and `replay.py`. Two of those already had
> full test files — written two months before this issue existed — and I did not
> check before pointing contributors at them. Details and an apology in the
> comments. The work below is what is actually missing, measured rather than
> guessed.

## Goal

Three pure-Python helpers report **100% statement coverage** while leaving one
branch each unexecuted. Statement coverage is what hid them: the line runs, the
*other side of the condition* never does. Each one below is a few lines of test
with the reproduction already worked out.

## The three branches

Run `pytest tests/ --cov=soup_cli --cov-branch --cov-report=term-missing` to see
them yourself.

| module | branch | what is never executed | reproduction |
|---|---|---|---|
| [`utils/replay.py`](src/soup_cli/utils/replay.py) | `80->82` | `downsample`: the stride lands **exactly** on the last row, so the endpoint pin is skipped | `downsample(rows, max_points=4)` with 9 rows |
| [`utils/log_level.py`](src/soup_cli/utils/log_level.py) | `74->73` | `setup_logging`: an existing **foreign** handler -- one Soup did not install, so it carries no `_soup_log_tier` at all and line 74's `is not None` is False | attach a plain `logging.StreamHandler` to the root logger, then call `setup_logging(tier)` |

> **Correction (2026-09-02).** The row above originally read "an existing handler whose `_soup_log_tier` **matches** the requested tier", with the repro "call `setup_logging(tier)` twice with the same tier". That is a different edge, `75->73`, and following it would **not** have closed `74->73`. Line 74 is `if getattr(existing, "_soup_log_tier", None) is not None:` and the uncovered branch is that condition being False. Caught by @umran666 in [#631](https://github.com/MakazhanAlpamys/Soup/pull/631), who read the source rather than this table; the row is corrected above. `74->73` is now covered and the module is at 100% branch coverage. **This issue stays open** for the remaining rows -- and note that `75->73`, the edge this row used to describe, is a real branch that nothing here has established needs its own test.
| [`utils/lr_groups.py`](src/soup_cli/utils/lr_groups.py) | `229->235` | `build_optimizer_param_groups`: **every** parameter matched a pattern, so `base_bucket` is empty and no `"base"` group is appended | patterns that match all parameter names |

[...]

PR #469, opened 2026-08-21
test(lr-groups): cover the all-matched branch in build_optimizer_param_groups

AmirF194:
`build_optimizer_param_groups` in `lr_groups.py` reported 100% statement
coverage while one branch went unexercised: when every parameter name
matches a configured `lr_groups` pattern, `base_bucket` stays empty and the
`if base_bucket:` guard at line 229 skips appending the `base` group. No
test drove that path, so a future regression there (the optimizer silently
losing its base LR group) would not be caught.

Adds `test_all_params_matched_no_base_group`, which routes every parameter
into a pattern group and asserts the exact output: no `base` entry, and
each pattern bucket holds precisely the parameters it should. Also adds the
CHANGELOG entry the repo's convention expects for PR-authored changes.

This covers the `lr_groups.py` branch only, the one the issue calls out as
having real training consequences. `replay.py`'s branch is a separate,
smaller follow-up per the issue's own "one module per PR" guidance;
`log_level.py` is already at 100% branch coverage on current `main` and
needs nothing.

Verification:
- `pytest tests/test_v0410_part_b.py --cov=soup_cli.utils.lr_groups --cov-branch`: branch `229->235` shows `Missing` without this test, `100%` with it.
- `pytest tests/test_v0410_part_a.py tests/test_v0410_part_b.py tests/test_v0410_part_c.py`: 121 passed.
- `ruff check src/soup_cli/ tests/`: clean.

Refs #273 (lr_groups.py branch only; replay.py to follow in a separate PR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,10 @@ reproducing 70+ versions of notes.
 - **A repo-wide documentation ratchet to guarantee declared recipe counts stay synchronized with the catalog (#453 by @harshitthek in #457).**
   Derives the expected count dynamically from `len(RECIPES)` and scans all declared documentation sites, preventing silent Git auto-merge drift across sequential recipe additions.
 
+- Branch coverage for `lr_groups.py`'s `build_optimizer_param_groups`: the case
+  where every parameter matches a configured group, so no `base` optimizer
+  group is appended (#273).
+
 ### Changed
 
 - **Lazy callback builders now self-import their callback class names so runtime
--- a/tests/test_v0410_part_b.py
+++ b/tests/test_v0410_part_b.py
@@ -171,6 +171,24 @@ class TestBuildOptimizerParamGroups:
         assert len(out) == 1
         assert out[0]["name"] == "base"
 
+    def test_all_params_matched_no_base_group(self):
+        # Every param routes into a pattern group, so base_bucket stays empty
+        # and must not be appended (branch 229->235 in lr_groups.py).
+        groups = parse_lr_groups([
+            ("self_attn", 1e-4),
+            ("mlp", 2e-4),
+            ("lm_head", 3e-4),
+        ])
+        out = build_optimizer_param_groups(self._named_params(), 2e-5, groups)
+        names = [g["name"] for g in out]
+        assert names == ["lr_group:self_attn", "lr_group:mlp", "lr_group:lm_head"]
+        attn_bucket = next(g for g in out if g["name"] == "lr_group:self_attn")
+        assert sorted(attn_bucket["params"]) == ["T_q0", "T_v0"]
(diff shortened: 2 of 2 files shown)


MakazhanAlpamys:
Seen — in my queue.

The single thing that decides this one: I will break the all-matched branch in `lr_groups.py` and check your test fails. A test covering an uncovered branch is only worth adding if it **discriminates** — otherwise it raises the coverage number and guards nothing, which is worse than the gap because it looks covered.

I will also check whether the branch was genuinely uncovered on `main`, rather than taking the premise from #273.

MakazhanAlpamys:
Merged as `2756414`. **Your test is the sole guard on all three mutations I ran** — that is the strongest result this kind of PR can get, and it means the branch genuinely had nothing watching it.

```
if base_bucket:  ->  if base_bucket is not None:      # the realistic regression
FAILED TestBuildOptimizerParamGroups::test_all_params_matched_no_base_group
1 failed, 45 passed
```

I ran the subtle one myself rather than the crude one, because `is not None` is the mutation someone actually writes: an empty list is not None, so an empty base group gets emitted and nothing else in the suite notices.

**You guarded more than you claimed.** Reversing the `zip(compiled, buckets)` — i.e. emitting the groups in a different order — is also caught only by your test. That matters beyond tidiness: `optimizer.state_dict()["param_groups"]` is positional, so a silent reorder breaks optimizer-state resume on `--resume`. Nothing in this repo pinned group order before you.

**And the restraint is the part I want to name.** You assert exact names, exact membership and exact ordering, but *not* dict equality — so adding a key like `weight_decay` to the emitted groups will not redden it. That is the difference between a test and a tripwire. Half this week has been me asking people to narrow guards that fire on correct code; yours arrived already narrow.

Verified the premise rather than taking it from #273: `99% -> 100%`, with `229->235` present under `Missing` beforehand and absent after. CI 12/12 on the true head, zero test deletions, no source touched.

## One thing I did on your behalf, and I would rather say it than let you find it

Your CHANGELOG entry read `(#273)`; the house form is `(#273 by @AmirF194 in #469)`. **I corrected it in the follow-up commit `ba45463` rather than sending you back for one line.** Normally I ask for this in the PR, because reconstructing attribution at release time is exactly when it goes wrong — four suffixes were wrong or incomplete in v0.73.3 when I checked them against `gh`. Fixing it at merge and telling you is not that failure mode, but it is your name on it, so you should know it was touched.

**#470 will need the same suffix and a rebase** — both PRs insert at the identical CHANGELOG anchor, so it now conflicts. Test files merge cleanly; it is only that one hunk.

AmirF194:
Thanks for running the mutations yourself and for fixing the CHANGELOG suffix, appreciate you flagging it rather than just doing it quietly. Will carry the same suffix and rebase on #470.

PR #470, opened 2026-08-21
test(replay): cover the stride-lands-on-last-row branch in downsample

AmirF194:
`downsample` in `replay.py` reported 98% branch coverage: `if sampled[-1] is not rows[-1]:` at line 80 always took the true branch in the existing tests, so the case where the stride already lands on the last row (the pin is redundant and must be skipped) went unexercised. A future regression there, appending a duplicate final row, would not be caught.

Adds `test_stride_lands_on_last_row_no_duplicate_pin`: 5 rows with `max_points=2` gives `stride = 5 // 2 = 2`, so `rows[::2]` already ends on index 4. Asserts the exact output (`[rows[0], rows[2], rows[4]]`) and that the last element is the same object as `rows[-1]`, not a duplicate. Also adds the CHANGELOG entry the repo's convention expects.

This is the `replay.py` follow-up to #469 (`lr_groups.py`), per the issue's "one module per PR" note. `log_level.py` needs nothing; it is already at 100% branch coverage on current `main`.

Verification:
- `pytest tests/test_replay.py --cov=soup_cli.utils.replay --cov-branch`: branch `80->82` shows `Missing` without this test, `100%` with it.
- `pytest tests/test_replay.py`: 15 passed.
- `ruff check src/soup_cli/utils/replay.py tests/test_replay.py`: clean.

Refs #273 (replay.py branch only; #469 covers lr_groups.py and is still open, so this does not close the issue).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,9 @@ reproducing 70+ versions of notes.
 - Branch coverage for `lr_groups.py`'s `build_optimizer_param_groups`: the case
   where every parameter matches a configured group, so no `base` optimizer
   group is appended (#273 by @AmirF194 in #469).
+- Branch coverage for `replay.py`'s `downsample`: the case where the stride
+  already lands on the last row, so the endpoint pin is not appended a second
+  time (#273 by @AmirF194 in #470).
 
 ### Changed
 
--- a/tests/test_replay.py
+++ b/tests/test_replay.py
@@ -78,6 +78,15 @@ class TestDownsample:
         out = downsample(rows)
         assert len(out) <= MAX_PLOT_POINTS + 5
 
+    def test_stride_lands_on_last_row_no_duplicate_pin(self):
+        # 5 rows, max_points=2 -> stride = 5 // 2 = 2, so rows[::2] already
+        # ends on index 4 (the last row); the endpoint pin must not append a
+        # second copy (branch 80->82 in replay.py).
+        rows = [_row(i, 2.0) for i in range(5)]
+        out = downsample(rows, max_points=2)
+        assert out == [rows[0], rows[2], rows[4]]
+        assert out[-1] is rows[-1]
+
     def test_zero_max_rejected(self):
         with pytest.raises(ValueError):
             downsample([_row(0, 2.0)], max_points=0)


MakazhanAlpamys:
**Approved on the substance — and your mutation B is the reason.** I ran it myself:

```
stride = max(_MIN_STEP_DELTA, len(rows) // max_points)  ->  ... // max_points + 1
FAILED TestDownsample::test_stride_lands_on_last_row_no_duplicate_pin
1 failed, 14 passed
```

Fourteen tests pass under an off-by-one on the sampling stride, **including `test_long_capped`** — which is precisely what #273 argued would happen, since that test only bounds with `<=`. Your test is the only thing in the file that catches it. That is not a coverage number; that is the guard the issue was opened for.

Two smaller notes from the run:

- Inverting `is not` to `is` kills your test **plus** `test_long_capped`, so the endpoint-pin half is doubly covered.
- `is not` -> `!=` survives all 15. I judge that an **equivalent mutant, not a gap** — it can only diverge when a distinct object compares equal to `rows[-1]`, in which case the emitted chart is byte-identical. Reporting it because I ran it, not because it needs action.

Zero test deletions, no source touched, `ruff` clean, CI 12/12 on the true head. Coverage `98% -> 100%`, `80->82` present under `Missing` before and absent after — verified rather than taken from the issue.

## Two things before it merges

**1. Rebase.** #469 merged a few minutes ago and both PRs insert at the identical CHANGELOG anchor, so this now conflicts. The test files merge cleanly — it is only that one hunk.

**2. The suffix**, while you are there: `(#273)` should be `(#273 by @AmirF194 in #470)`. I fixed the equivalent line on #469 myself at merge and said so there, but since you are rebasing anyway this one is cheaper for you to carry.

## One observation, not a merge condition

Your test hardcodes `out == [rows[0], rows[2], rows[4]]`, which pins the **stride formula**, not only the endpoint behaviour the test name advertises. I checked what that costs: `downsample`'s docstring promises *"Return at most `max_points` rows"* and the shipped code violates it — `max_points=2` returns **3** rows. If someone later fixes that with a ceil-based stride, your test goes red, alone, under a name that says `..._no_duplicate_pin`.

That is the shape that gets a test deleted as stale. **Your inline comment already mitigates it** — spelling out `5 rows, max_points=2 -> stride = 5 // 2 = 2` means a reader sees immediately what is pinned — so I am not asking you to change the assertion. A clause in that comment saying it pins the stride formula *as well as* the endpoint would close it entirely.

The docstring-vs-behaviour mismatch is pre-existing and your PR merely exposes it. I will file it separately rather than widening this.

MakazhanAlpamys:
Follow-up filed as **#473** — the `downsample` docstring-vs-behaviour mismatch, with your test named explicitly in the acceptance criteria as something to **update, not delete**, since it is the only guard against a stride off-by-one.

That way whoever fixes the contract sees why the red test is there before deciding it is stale.

AmirF194:
Thanks for the mutation testing and for filing #473 rather than widening this one. Rebased and carried the CHANGELOG suffix, both at head 025c876.

MakazhanAlpamys:
Merged as `3f926e5`. **Re-verified after your rebase rather than assuming a rebase is a no-op** — the mutation that matters still lands:

```
stride = max(_MIN_STEP_DELTA, len(rows) // max_points + 1)
FAILED TestDownsample::test_stride_lands_on_last_row_no_duplicate_pin
1 failed, 14 passed
```

**That is the whole argument of #273, reproduced.** `test_long_capped` bounds the result with `<=`, so an off-by-one in the stride changes which rows are sampled and it passes anyway. Yours is the only test in the file that notices.

Two things about the instrument itself:

- **The identity assertion is the right tool.** `assert out[-1] is rows[-1]` matches a branch whose guard is `is not` — an equality check would have been a different property. Inverting that `is not` kills your test *and* `test_long_capped`, so the endpoint half is doubly covered.
- `is not` -> `!=` survives all 15. I ran it and I am calling it an **equivalent mutant, not a gap**: it can only diverge when a distinct object compares equal to `rows[-1]`, and the emitted chart is byte-identical in that case. Mentioning it because I ran it, not because it needs anything.

Coverage `98% -> 100%`, `80->82` present under `Missing` before and absent after — verified rather than taken from the issue. CI 12/12 on the true head, zero test deletions, no source touched. Suffix correct this time.

## The judgement call I want to credit

Your assertion pins today's **stride formula**, not just the endpoint behaviour its name advertises — and `downsample()` currently contradicts its own docstring, returning 3 rows for `max_points=2`. You let that be filed as #473 rather than widening this PR to "fix" it, which was right: the contract is a decision, not a typo, and deciding it inside a coverage PR would have buried it.

I named your test explicitly in #473's acceptance criteria as something to **update, not delete**. Otherwise whoever fixes the contract sees one red test under a name about endpoint pinning and concludes it is stale — and the deletion would look justified.

**#273 stays open** — this closes one row of its table, not the issue.

Credited in `CONTRIBUTORS.md` (`9d24f42`). Three merges from you today, all of them tests that actually discriminate; that is rarer than it sounds.

AmirF194:
Appreciate you re-running the mutation check after the rebase instead of assuming it would still hold. Glad it landed cleanly.

PR #471, opened 2026-08-21
fix(trainer): load dtype -- checkpoint's own for a frozen base, explicit fp32 for full-FT

### LINKED ISSUE #339: Every non-quantized model loads in float32: from_pretrained gets no dtype, so a bf16 checkpoint costs 2x
`SFTTrainerWrapper` builds `model_kwargs` with `trust_remote_code` and `device_map` and **no dtype**, on all three modality paths (`sft.py:860`, `:1105`, `:1214`). `from_pretrained` then defaults to `float32` regardless of what the checkpoint stores.

Verified directly:

```
checkpoint config dtype : bfloat16
Soup's call (no dtype)  : torch.float32
with dtype='auto'       : torch.bfloat16
bytes per param   soup: 4    auto: 2
```

So a bf16 checkpoint occupies **twice** the VRAM it needs to. `training_args` sets `bf16=True`, which makes the *compute* bf16 via autocast, but the master weights stay fp32 — that is invisible in the loss and very visible in `nvidia-smi`.

### Why this is not simply "add dtype='auto'"

For **full fine-tuning** fp32 master weights are the standard mixed-precision arrangement and dropping them changes optimisation, not just memory. For a **frozen base under LoRA** — which is what most Soup runs are — holding the base in fp32 is pure waste: it is never updated, and the adapter is what trains.

So the fix is probably conditional (frozen base -> checkpoint dtype; trainable base -> keep fp32 master weights), and it is a numerics decision, not a one-liner. Filing rather than changing it unilaterally.

### Where it was found
Measuring #306 at 7B. It is also the reason 8B full-FT cannot run on an 80 GB card at all: 8.03B x 16 B/param = **119.7 GB** of weights + grads + AdamW state before a single activation. At bf16 master weights the same arithmetic is 8.03B x 12 = 96.4 GB — still over, but the gap is entirely this default.

### Acceptance
- [ ] a LoRA run on a bf16 checkpoint loads the frozen base at the checkpoint's dtype
- [ ] full fine-tuning keeps whatever precision arrangement is deliberately chosen, documented
- [ ] a test asserting the loaded dtype for both cases
- [ ] the VRAM pre-flight's assumptions re-checked against whichever choice lands

Record: `benchmarks/gate-h100-validation.md` STEP 22.

blackcoderx:
Closes #339.

`SFTTrainerWrapper` built `model_kwargs` for `from_pretrained` with
`trust_remote_code` / `device_map` / an optional `quantization_config` on all
three modality paths (text/vision/audio), but never a `dtype`.
`from_pretrained` then defaulted to float32 regardless of the checkpoint's own
dtype, so a bf16 checkpoint cost 2x the VRAM it needed. The maintainer's own
H100 measurement (Llama-3.1-8B, LoRA, frozen base): 48,241 MiB peak (fp32, as
shipped) -> 18,658 MiB (`dtype="auto"`), a confirmed 28.9 GB / 2.59x saving,
byte-identical across 3 repeats -- and explicitly left the fix undecided
("a numerics decision, not a one-liner").

The decision, made explicit rather than left to `from_pretrained`'s own
default:

- **Frozen base** (LoRA / QLoRA -- never receives an optimizer step):
  `dtype="auto"`, preserving the checkpoint's own dtype -- except on a
  pre-Ampere CUDA card (T4 / P100 / V100 / GTX 16xx / RTX 20xx), where
  `"auto"` would give bf16 storage while training compute correctly stays
  fp16, so it explicitly loads `torch.float16` there instead (asked via
  `bf16_fp16_flags`, the same helper `_resolve_mixed_precision` already
  uses for the identical card question).
- **Trainable base** -- full fine-tuning (`lora.r=0`, Spectrum
  `unfrozen_parameters`, or LISA `lisa_enabled`; all three schema-gated to
  `modality='text'`): explicit `torch.float32` master weights. Deliberate,
  not an accidental upcast.

`SFTTrainerWrapper._resolve_load_dtype(tcfg)` is the one place this is
decided, pairing with the existing `_resolve_mixed_precision` (compute
dtype) as the storage/load-dtype counterpart. Same discriminator at all
three call sites; on vision/audio it always resolves to a frozen-base
answer (`"auto"` or the pre-Ampere override) since those paths PEFT-wrap
unconditionally and the schema rejects the two full-FT flags outside
`modality='text'` -- routed through the shared method anyway rather than
hardcoded, so it stays correct by construction if a vision/audio full-FT
branch is ever added.

The full-FT discriminator itself (`unfrozen_parameters` / `lisa_enabled` /
`lora.r==0`) is now a single shared `is_full_finetune(tcfg)` in `sft.py`,
used by both `_resolve_load_dtype` and `commands/train.py`'s
`_build_hardware_fit_input` VRAM pre-flight classifier -- previously
independent copies that disagreed in both directions: the classifier missed
`lisa_enabled`/`lora.r==0` (under-predicting VRAM for those runs) and
treated bare `freeze_layers`/`freeze_ratio` as sufficient for `peft="full"`
even with LoRA still on (over-predicting, and able to falsely refuse a
launch that would actually fit -- a measured 16 GB phantom at 8B).
`freeze_layers`/`freeze_ratio` only reduce what's trainable within whichever
mode is chosen; they don't select the mode. `hardware_fit.py`'s own
`bytes_per` branch (`peft=="full"` + `quant="none"` -> 4 bytes/param) is
unchanged and now finally fed a correct classification.

`distill.py` has the identical missing-`dtype` bug on its own
`from_pretrained` calls but is a different trainer wrapper not named in
#339 -- left untouched.

## Testing
- `tests/test_issue339_load_dtype.py` (33 tests): unit tests of
  `_resolve_load_dtype` in isolation, including the pre-Ampere/Ampere/CPU
  card-awareness cases (with the `is_bf16_supported(including_emulation=...)`
  trap modeled explicitly); mock-based `dtype`-kwarg capture at all three
  call sites (text-LoRA, text-full-FT, vision, audio, plus a QLoRA case
  proving `dtype` and `quantization_config` coexist); end-to-end tests
  against real tiny on-disk checkpoints, including a control (LoRA against
  an fp32 checkpoint stays fp32 -- proving `"auto"` means "preserve the
  checkpoint's dtype", not "always bf16"); `hardware_fit.py` byte-per-param
  tests for full-FT vs LoRA vs QLoRA; and `is_full_finetune`/
  `_build_hardware_fit_input` classification tests covering both the
  under- and over-count fixes, including @AmirF194's exact repro
  (`freeze_ratio=0.5, lora.r=8` -> `peft="lora"`).
- Full suite: `pytest tests/ -q` -> **18,019 passed, 393 skipped, 5
  deselected, 0 failed**.


MakazhanAlpamys:
**The fix is correct, the win is real, and it is genuinely well guarded.** I am blocking on one thing, and it is a maintainer call rather than a defect in your reasoning.

## What holds, proven by breaking it

Removing `"dtype": self._resolve_load_dtype(tcfg)` from all three `from_pretrained` sites — i.e. reintroducing #339 — kills 5 tests. Each of the three full-FT flags is covered individually, and `test_control_lora_from_an_fp32_checkpoint_stays_fp32` is exactly the right control: it stops `"auto"` from silently meaning "always bf16". Zero test deletions anywhere in the PR.

And an unclaimed benefit worth naming: because pre-PR LoRA runs loaded fp32 while `hardware_fit` assumed 2.0 bytes/param for `quant="none"`, the pre-flight was **under-predicting 2x for the dominant LoRA case**. Your PR makes that assumption true.

## The blocker: the resolver never asks what card it is on

```python
is_full_finetune = bool(tcfg.unfrozen_parameters or tcfg.lisa_enabled or tcfg.lora.r == 0)
if not is_full_finetune:
    return "auto"
```

I grepped the whole method: it consults `cuda_supports_bf16()`, `get_compute_dtype()`, `bf16_fp16_flags()` — **zero times**. Stubbing a pre-Ampere card:

```
STREAMED path   resolve_stream_dtype('cuda') = float16     <- card-aware
NON-STREAMED    _resolve_load_dtype(LoRA)    = 'auto'      <- card-blind
compute flags   bf16_fp16_flags('cuda')      = (False, True)
```

`'auto'` on a bf16 checkpoint gives bf16 storage. So on a T4 with any modern checkpoint the run becomes **bf16 storage with fp16 compute** — a combination that exists nowhere else in this codebase.

**Why I will not wave this through even though your reasoning is defensible.** v0.73.1 removed exactly this decision from *fourteen* places, because bf16 was assumed on every CUDA card and the entire free Colab/Kaggle tier failed on every task. The recorded trap from that release: `torch.cuda.is_bf16_supported()` defaults to `including_emulation=True`, so a T4 answers **True**, and the first fix was a no-op on the exact hardware it was written for. `resolve_stream_dtype`'s own docstring names this failure mode — *"wrong on every pre-Ampere GPU… and it could not fail on the Ampere card every published measurement was taken on"*. This is the highest-traffic load path in the project.

**Your counter-argument is real and I want it on the record**: the base is frozen and autocast casts per-op to fp16, so linear layers are plausibly fine. But norms, embeddings and `lm_head` run in the parameter's own dtype, and **nobody has measured that on sm_75** — I certainly have not; this box is an RTX 3050, sm_86, native bf16. I am not claiming it breaks. I am saying we removed this class of decision on purpose and should not re-add it untested on the busiest path.

**What I am asking for**: route the frozen-base case through the shared card-aware helper — pre-Ampere gets fp16, everything else gets `"auto"` — so the decision lives in one place, as `resolve_stream_dtype` already does for the streamed path.

## Two more, both measured

**QLoRA numerics change, undocumented and untested.** `dtype` is added to `model_kwargs` *before* `quantization_config`, so it applies to 4-bit runs. Real GPU, real bitsandbytes, NF4:

```
BEFORE  norm=torch.float16   embed=torch.float16
AFTER   norm=torch.bfloat16  embed=torch.bfloat16
```

Every existing QLoRA run moves fp16 → bf16 on its non-quantized modules. Arguably more correct — it matches `bnb_4bit_compute_dtype` — but no doc says it and no test covers it. On pre-Ampere it also puts bf16 norms on a card without bf16 units, which is the same question as above.

**The `hardware_fit` reasoning is wrong in both directions.** The PR assumes `peft=="full"` identifies the configs sent to fp32. Measured at Llama-3.1-8B, `quant: none`:

```
freeze_layers: 8 (LoRA on)  peft=full   dtype=auto     preflight=4.0  real=2.0   <- OVER-predicts
lora.r: 0                   peft=lora   dtype=float32  preflight=2.0  real=4.0   <- under
lisa_enabled                peft=lora   dtype=float32  preflight=2.0  real=4.0   <- under
```

You flagged the two under-predicting rows in a comment. The third is new: `freeze_layers` maps to `peft="full"` while keeping LoRA on, so the new `bytes_per = 4.0` branch adds a **16.06 GB phantom** to the weights bucket. `_hardware_fit_preflight` is a hard refusal without `--allow-oom-attempt`, so that is a false-refusal path. The dominant error there (18.51 → 130.77 GB via `trainable_frac`) is pre-existing and not yours — but the *reasoning* used to justify the branch does not hold, and #339's AC4 asks for the pre-flight assumptions to be re-checked.

## Nits

- `CHANGELOG.md:17` — `(#339 by @blackcoderx in #TODO-PR)`, should be `#471`. Also filed under `### Added` for a `fix(...)`.
- The docstring says it "mirrors the exact **three**-flag discriminator" used for the summary label. That block at `sft.py:263-270` checks **two** flags — `unfrozen_parameters` and `lora.r == 0`, not `lisa_enabled`. The mirroring argument is your stated correctness basis, so the inaccuracy matters more than a normal docstring slip.
- PR body says 16 tests in the new file; there are 14.
- The pre-existing failure your body mentions (`test_v07113.py::TestCountNewThumbs::test_counts_after_ts`) **did not reproduce** on a clean run here, and passes in isolation on `main`. Environment flake, not a real issue — you can stop worrying about it.

Layer streaming is unaffected: it branches at `sft.py:230` into `_setup_streaming_transformers`, resolves via `resolve_stream_dtype`, and never touches the modified `model_kwargs`. I checked rather than assumed.

## Summary

The card-aware resolver is the one I need before merge. The QLoRA doc sentence and the `hardware_fit` branch are cheap alongside it. This is a good PR carrying a 2.59x memory win — I would rather spend one more round than re-open #385 on the free tier.

src/soup_cli/utils/hardware_fit.py
```diff
@@ -258,6 +258,12 @@ def estimate_peak_vram_gb(inp: HardwareFitInput) -> VRAMBreakdown:
 
     params = float(inp.params_b) * 1e9  # absolute count
     bytes_per = _BYTES_PER_PARAM_BY_QUANT.get(inp.quant, 2.0)
+    if inp.peft == "full" and inp.quant == "none":
```
AmirF194, line 261: I ran this against 43bb666 in a clean container to check the freeze_ratio + LoRA combination sft.py's own comment at line 1014 documents as legal ("freeze bottom layers before LoRA").

`_build_hardware_fit_input` (commands/train.py:62-66) classifies any config with `freeze_ratio` or `freeze_layers` set as `peft="full"`, regardless of `lora.r`. `estimate_peak_vram_gb` (hardware_fit.py:261) then charges fp32 weight bytes for that classification. But `_resolve_load_dtype` (sft.py:868) only treats `unfrozen_parameters`, `lisa_enabled`, and `lora.r == 0` as full fine-tuning, not `freeze_ratio` or `freeze_layers`, so a `freeze_ratio=0.5, lora.r=8` config actually loads at `dtype="auto"` (checkpoint's own bf16/fp16, 2 bytes/param). The pre-flight ends up predicting double the real weight memory for this combination and can refuse launches that would actually fit.

```
$ docker run --rm -v $(pwd):/repo:ro -w /repo python:3.11-slim sh -c "pip install --quiet -e . && python repro_hwfit_mismatch.py"
classifier peft = full
pre-flight weights_gb (peft=full path) = 32.0
SFTTrainerWrapper._resolve_load_dtype(tcfg) = auto
decide_hardware_fit(available=31.00 GB).ok = False
predicted peak_vram_gb = 130.46635737743355

MISMATCH CONFIRMED: pre-flight predicts fp32 weight bytes and refuses a launch, while the trainer's own _resolve_load_dtype for the same config resolves to "auto" (bf16); the run would actually need about half the predicted weight memory.
```

This is the mirror image of the lisa_enabled/lora.r==0 gap already flagged in the train.py comment: there the classifier under-counts full fine-tuning (real OOM risk), here it over-counts (false refusal). Same fix shape either way: give `_build_hardware_fit_input` and `_resolve_load_dtype` one shared discriminator instead of two hand-maintained flag lists.

blackcoderx:
One more thing found while working through this revision, that neither you nor @AmirF194 flagged: there's a *third* independent copy of the same full-FT flag-check, in `setup()`'s console summary label (`sft.py:263-270`) — and it has its own bug. It's missing `lisa_enabled`, so a LISA run currently prints "LoRA applied" instead of naming LISA.

It's a one-line fix (add the missing `elif tcfg.lisa_enabled:` branch, matching the real branching order elsewhere in the file), but it's outside what either of you asked for, so I wanted to check first rather than fold it in silently. Want it included in this same revision, or left for a separate follow-up?

blackcoderx:
Both fixed, thanks for the thorough digging.

**Card-aware `_resolve_load_dtype`** (the blocker): it's now an instance method and asks `bf16_fp16_flags(self.device)` -- the same helper `_resolve_mixed_precision` already uses for the identical question -- before returning `"auto"`. On CUDA + pre-Ampere it now returns explicit `torch.float16`; everywhere else (Ampere+, CPU) is unchanged. New tests cover the pre-Ampere case, the Ampere control, the CPU control, the `including_emulation` trap specifically, and that full-FT's fp32 still wins regardless of card.

**Shared discriminator**: pulled the full-FT check into one `is_full_finetune(tcfg)` in `sft.py`, used by both `_resolve_load_dtype` and `_build_hardware_fit_input`'s `peft` classifier. Fixes both directions: `lisa_enabled`/`lora.r=0` now correctly classify `peft="full"`, and `freeze_layers`/`freeze_ratio` **with LoRA still on** no longer misclassifies as `"full"` -- @AmirF194's exact repro (`freeze_ratio=0.5, lora.r=8`) is now a named test asserting `peft=="lora"`.

Also: documented the QLoRA non-quantized-module dtype change (with a new test proving `dtype` and `quantization_config` coexist correctly), moved the CHANGELOG entry to `### Fixed` with the real PR number, and corrected the docstring's wrong "three-flag" claim.

Extended `tests/test_issue339_load_dtype.py` from 14 to 33 tests. Full suite: **18,019 passed, 393 skipped, 5 deselected, 0 failed** -- fully clean this time, including the previously-flaky `test_v07113.py` test you both correctly called out as unrelated.

The third (label) bug I asked about above is intentionally NOT included here -- still waiting on that answer.

AmirF194 (commented):
Confirmed, that matches the repro. Thanks for adding the dedicated test for it.

MakazhanAlpamys:
**Everything I asked for last round is done, and done properly — each fix fails a named test when removed.** I am still holding the merge, but on something new that neither of us saw, and that CI structurally cannot see.

## What you closed

The card-aware resolver works, including the trap that made the first #385 fix a no-op. My own stub, not your test file, modelling a T4 that answers `True` to the bare call:

```
PRE-AMPERE T4 sm_75 (is_bf16_supported() -> True)
  cuda_supports_bf16()      = False
  LoRA frozen base (cuda)   = torch.float16      <- the fix
AMPERE+ sm_86
  LoRA frozen base (cuda)   = auto               <- win preserved
REAL CARD (RTX 3050)
  LoRA frozen base (cuda)   = auto
```

Stripping the card branch back to a bare `return "auto"` — my blocker, verbatim — fails **3 tests**, at both the unit and the `model_kwargs` call-site level. Six mutations run, six killed. Removing `"dtype"` from the three sites now fails **7** tests, up from 5 before your revision.

`hardware_fit` is right in all three rows now, and it carries **an unclaimed bonus worth stating**: the shared discriminator also removes a large pre-existing false refusal I had written off as not yours. Same `freeze_layers: 8` + LoRA config:

```
main:      peft=full,  peak_gb = 119.16   <- hard-refuses on every card
PR #471:   peft=lora,  peak_gb =  24.12
```

QLoRA is documented and tested, the CHANGELOG suffix is right, the "three-flag" claim is gone, the PR body's test count now matches reality (33), and there are **zero test deletions** anywhere. Full suite 18271 passed / 0 failed.

## The new blocker: `dtype=` is the transformers >= 4.56 spelling

All three sites pass `"dtype"`. That kwarg does not exist before 4.56 — it is the rename of `torch_dtype=`. And this project pins:

```
transformers>=4.36.0,<5.0.0
```

So on **4.36 through 4.55 — twenty minor versions inside our own declared support** — this is a hard `TypeError` at model load, before step 0, on text, vision, audio and QLoRA alike:

```
transformers 4.55.4   dtype='auto'       -> TypeError: __init__() got an unexpected keyword argument 'dtype'
transformers 4.55.4   torch_dtype='auto' -> loaded torch.bfloat16
transformers 4.56.0   both spellings     -> bfloat16
transformers 4.57.6   both spellings     -> bfloat16
```

**Why nothing caught it, and this is the part I want on the record:** CI runs `pip install -e ".[dev]"`, which resolves to the newest satisfying transformers (4.57.x), so the floor of our own pin is never exercised — hence 12/12 green. Your mocked tests use `def _fake(*args, **kwargs)`, which swallows any kwarg and structurally cannot see a signature rejection. The real end-to-end tests would catch it, but only at the installed version.

And the exposure is not theoretical: pip does not upgrade an already-satisfying dependency. Someone on 4.55.4 running `pip install soup-cli[train]` keeps 4.55.4. That is the older-image Colab/Kaggle population — **the same fleet #385/#387 was about**, which is a pointed irony for a PR whose whole blocker was that fleet.

**The fix is a rename at three sites: `"dtype"` -> `"torch_dtype"`.** I verified it is safe on the new end too, from the installed 4.57.6 source:

```python
torch_dtype = kwargs.pop("torch_dtype", None)  # kept for BC
logger.warning_once("`torch_dtype` is deprecated! Use `dtype` instead!")
```

Deprecated but honoured, and the repo already tolerates that warning at **six** existing sites — including `utils/adapter_fuse.py:79`, which is literally `torch_dtype="auto"`, the same call you want. Your PR is the only `dtype=` user in the codebase.

Raising the floor to `>=4.56` instead is a dependency-policy call and not yours to make inside this PR. Take the rename.

It also wants a test that pins the spelling, since neither CI nor the mocks can see it — asserting the key in `model_kwargs` is `torch_dtype` would do, and it is the only thing that stops the next person "modernising" it back.

## Your label question — yes, fold it in

You found a third copy of the full-FT check in `setup()`'s summary label and asked before touching it. I confirmed it independently:

```
LISA (lisa_enabled: true)      is_full_finetune=True   printed label = 'LoRA applied'
Spectrum (unfrozen_parameters) is_full_finetune=True   printed label = 'Spectrum targeted FT'
full-FT (lora.r: 0)            is_full_finetune=True   printed label = 'Full fine-tuning'
```

Real, pre-existing from #340, and it is one `elif` using the `is_full_finetune()` you just created. **Fold it in.** Leaving a third hand-rolled copy is exactly the drift this PR exists to end, and asking first was the right instinct.

## What nobody has measured

The behavioural question from my last round — whether bf16 storage with fp16 compute actually degrades norms, embeddings and `lm_head` on sm_75 — is **still unmeasured by anyone**, mine included; this box is sm_86. Your revision avoids the combination rather than characterising it, which is what I asked for and the right call. I am not claiming it breaks; I am saying we no longer need to know.

Also stated plainly: the H100 48,241 -> 18,658 MiB headline is **not independently reproduced** here. 4.29 GB cannot hold Llama-3.1-8B. I verified the mechanism on a real bf16 checkpoint — fp32 -> bf16, 4 -> 2 bytes/param — not your absolute numbers.

The rename, the spelling test, and the label `elif`. Then it merges.

MakazhanAlpamys:
Filed the structural half as **#478** — CI installs the newest satisfying transformers, so the floor of our own pin is never executed, which is why this went 12/12 green.

That issue is ours, not yours. Your PR only needs the rename.

MakazhanAlpamys:
Update on the block, with a measurement that settles it — and a merge-order decision that reduces your diff rather than growing it.

**I confirmed the `dtype=` spelling is a hard failure, not a silent no-op.** I installed `transformers==4.46.1` (the true floor: 4.46.0 is yanked, and `trl 0.14.0` requires `>=4.46.0`) into an isolated directory and ran it against a locally-built bf16 checkpoint:

```
```

The third line also independently re-confirms your premise for #339, so the fix itself is right — it is only the spelling that is wrong, at all three of your sites.

**Meanwhile #492 landed the complementary half.** @AmirF194 filed #491 as an explicit follow-on to your work and fixed the twelve non-SFT trainers, adding a shared `resolve_frozen_base_load_dtype(device)` to `utils/gpu.py` — which contains this, byte-for-byte the same decision as your `_resolve_load_dtype`:

```python
_, fp16_only = bf16_fp16_flags(self.device)
if fp16_only:
    return torch.float16
return "auto"
```

Landing both as they stand would give us two implementations of one decision in two different kwarg spellings — the duplicated-truth pattern that has bitten this project three times (#372, #392, #424).

**So I am merging #492 first, and asking you to rebase onto its helper.** That should make your PR *smaller*: `_resolve_load_dtype` becomes roughly

```python
if is_full_finetune(tcfg):
    return torch.float32
return resolve_frozen_base_load_dtype(self.device)
```

and your three call sites switch from `dtype=` to `torch_dtype=`. The helper deliberately has no full-fine-tuning branch, so the full-FT half — the part that is genuinely yours and that #492 does not cover — stays exactly as you designed it.

To be explicit about credit, since this is your issue and your diagnosis: #491 exists because of #339 and says so, and your docs work (three pages) is the only user-facing documentation either PR ships. Nothing is being taken over — the ordering is purely so we end up with one resolver instead of two.

Also relevant: **#489 by @AchuthReddy-16** implements the CI floor guard from #478 — the structural gap that let this spelling pass 12/12 in the first place. Its scanner catches `dtype=` at any production `AutoModel*.from_pretrained` site, so once it lands, this class of defect fails in CI rather than in review.

Take your time on the rebase; nothing here is urgent, and the conflict with `main` is the CHANGELOG adjacency problem (#487), which is being fixed separately.

MakazhanAlpamys:
Status note, and an offer — not a nudge for speed.

**#492 has not merged yet** (it is waiting on CI after the author fixed a reward-head precision finding), so there is nothing to rebase onto *yet*. Do not start until it lands; I will comment here the moment it does, with the exact shape of the change.

To restate it so it is not a surprise: after #492 lands, `_resolve_load_dtype` in `trainer/sft.py` becomes roughly

```python
if is_full_finetune(tcfg):
    return torch.float32
return resolve_frozen_base_load_dtype(self.device)
```

and your three call sites switch `dtype=` → `torch_dtype=`. That should make your diff **smaller**, not larger, and the full-fine-tuning branch — the part that is genuinely yours and that #492 does not cover — stays exactly as you designed it.

One thing that has changed since we last spoke and is worth knowing: **the wrong spelling is not a silent no-op.** I installed transformers 4.46.1 in isolation and measured it — `dtype=` raises `TypeError: LlamaForCausalLM.__init__() got an unexpected keyword argument 'dtype'` at model load, on every version from our floor to 4.55. Your fix is right; only the spelling was wrong. Twelve production sites had the same defect and shipped broken; #489 has since fixed those and added a guard that fails CI if it recurs — so once you switch spelling, that guard is on your side.

You have two PRs open here (#468 as well) and I have just asked there whether the queue is too long. Same offer applies: if you would rather I carried the rebase on this branch with your authorship intact, say the word. If you would rather finish both yourself, that is the better outcome and I will simply wait.

MakazhanAlpamys:
**#492 has landed** (`5c20a41`), so this is unblocked — here is the exact shape, as promised.

`resolve_frozen_base_load_dtype(device)` now exists in `src/soup_cli/utils/gpu.py`. Your `_resolve_load_dtype` in `trainer/sft.py` should become roughly:

```python
if is_full_finetune(tcfg):
    return torch.float32
return resolve_frozen_base_load_dtype(self.device)
```

and your three call sites switch `dtype=` → `torch_dtype=`. That is the whole change. The helper deliberately has **no** full-fine-tuning branch, so the part that is genuinely yours — and that #492 does not cover — stays exactly as you designed it. Your diff should get smaller.

One thing to know before you touch the spelling: `dtype=` is not a style preference. I measured it on transformers 4.46.1 — it raises `TypeError` at model load on every version from our floor to 4.55. Twelve production sites had it and shipped broken; #489 has since fixed those and added a CI guard that fails if it comes back. So once you switch, that guard is on your side rather than against you.

Also worth having: #492's review turned up a precision regression that its own fix introduced — peft upcasts LoRA adapters to fp32 but never the `modules_to_save` wrapper, so a SEQ_CLS head would have trained in bf16. Not your path (`sft` has no auto `modules_to_save`), but if your full-FT branch ever touches a head, that is the shape to watch for.

No deadline on this. And the offer from my last comment stands: if you would rather I carried the rebase on your branch with your authorship intact, say so — you have two PRs open here and I would rather share the load than watch good work stall.

blackcoderx:
@MakazhanAlpamys — done, per the shape you laid out.

**Rebased onto `#492`.** `_resolve_load_dtype` now delegates the frozen-base half to `resolve_frozen_base_load_dtype(self.device)` instead of re-deriving the same `bf16_fp16_flags` check inline — one resolver, not two. The full-FT branch (`torch.float32`) is unchanged, still mine.

**`dtype` → `torch_dtype`** at all three call sites. New test pins the literal kwarg *name*, not just its value, at all three sites — the guard against the next person "modernising" it back that you asked for.

**The LISA label — folded in.** `setup()`'s summary now has the `elif tcfg.lisa_enabled:` branch, matching the real mode-selection order in `_setup_transformers`. New test in `test_issue341_seed_and_fullft.py`, sibling to the existing `lora.r=0` one.

**Rebase conflict:** only `CHANGELOG.md`, exactly where I'd have expected — both sides added a bullet at the same anchor after `### Fixed`. Kept every upstream entry verbatim and put this PR's rewritten bullet first, which is how your own `#491`/`#492` entry already reads it ("#471 fixed this for the SFT trainer").

Full suite: **18,269 passed, 396 skipped, 5 deselected, 0 failed.**

MakazhanAlpamys (changes requested):
Reviewed. **The code is right and I could not break it — the one thing holding this is
process, and it needs no code change.**

## The blocker

The PR edits `CHANGELOG.md` directly. Since #487/#490 that is no longer where entries go:
they are per-PR fragments at `changelog.d/<latest-release>/<PR>.<category>.md`, so merges
stop conflicting every other open PR. Yours would be:

```
changelog.d/0.73.3/471.fixed.md
```

Same prose, moved. Keyed by **PR** number (471), not the issue. `fixed` is the right
category — this is behaviour-changing and user-facing, so a fragment is required rather
than optional. The policy landed *before* your merge-base, so this is a "it moved under
you" problem, not a "you missed the docs" one. `changelog.d/README.md` has the format.

While you are in there: the H100 figures (48,241 MiB → 18,658 MiB) are inherited verbatim
from #339/#492's measurement rather than re-measured here. That is reasonable reuse, but
the entry should say whose measurement it is — this repo has been bitten by numbers that
outlived the code they were taken on.

## What I checked rather than assumed

**No duplication with the already-merged #492**, which was my main worry when I saw the
overlap. `utils/gpu.py` is untouched by your diff and `sft.py` imports and delegates to
`resolve_frozen_base_load_dtype`. You reused the shared resolver instead of re-deriving
the pre-Ampere check inline — which is exactly the point #492 put it in `utils/gpu.py`
for, and the opposite of what usually happens when two PRs converge on one bug.

**No `dtype=` kwarg anywhere.** Confirmed three ways: the diff's only `dtype=` occurrence
is a docstring explaining why not to use it; the `#478` AST scanner runs clean against
your branch; and your own
`test_the_kwarg_key_is_torch_dtype_not_dtype_at_all_three_sites` asserts it directly. That
would have been a hard blocker, and you pre-empted it.

**Six removed lines, none of them a test** — all the old `getattr(...) or getattr(...) or
getattr(...)` block in `commands/train.py`.

**Five mutations, five killed by named tests:**

```
drop torch_dtype from the text model_kwargs   -> 6 named tests fail
is_full_finetune() -> return False            -> 13 named tests fail
restore the old getattr chain in train.py     -> 4 named tests fail, exactly the
                                                 under/over-count pairs you claim
disable the fp32 branch in hardware_fit.py    -> 1 named test fails (2.0 vs 4.0)
remove the LISA summary label branch          -> 1 named test fails
```

The third one is what makes this more than a dtype fix. Replacing that inline chain with
`is_full_finetune()` repairs a live VRAM pre-flight bug **in both directions** — it missed
`lisa_enabled` / `lora.r == 0` and under-predicted, and it treated bare `freeze_layers` /
`freeze_ratio` as sufficient with LoRA still active and over-predicted, which could refuse
a launch that would have fit. Your four tests name both directions separately.

Your controls are there too — `test_control_plain_lora_is_lora`,
`test_quantized_weights_unaffected_by_the_full_ft_branch`,
`test_ampere_cuda_frozen_base_still_resolves_to_auto`. A widening with no negative controls
is a guard that eventually fires on correct code and gets deleted; yours does not.

**Stated plainly, not glossed:** the pre-Ampere `float16` branch cannot be exercised on
real silicon here (RTX 3050, sm_86), and the H100 numbers were not re-run. Both rest on the
fake-CUDA fixture, which mirrors `test_issue385_stream_dtype.py`'s pattern — the right call,
and worth knowing it is the state of the evidence.

## Rebase, once

`main` moved while this sat. **#484 landed a few minutes ago and touches
`commands/train.py`** — the only file of yours that overlaps anything merged since your
base. I held it back specifically so you would not rebase twice; now that this needs a push
anyway, one rebase onto current `main` (`0653129`) covers both. Nothing else in #494-#501
touches `sft.py`, `train.py`, `hardware_fit.py` or `gpu.py`.

So: move the changelog entry to a fragment, rebase, push. No code review comments — there
are none.

(Separately, #468 is still yours and still waiting on the two MEDIUMs from that thread, if
you want to batch a session.)

blackcoderx:
@MakazhanAlpamys — done, both process items.

**Moved to a fragment.** The bullet is now `changelog.d/0.73.3/471.fixed.md` — same prose, `CHANGELOG.md`'s direct edit removed. Validated directly against `validate_fragments()` rather than assumed. While moving it, added the attribution you asked for: the H100 numbers (48,241 → 18,658 MiB) now say explicitly that they're carried over from the original `#339`/`#471` benchmark, not re-measured this round.

**Rebased onto current `main`** (past `#484`/`#486`/`#496`/`#498`/`#501`). Dry-ran it with `git merge-tree` first — zero conflicts, confirmed before merging, not after. Makes sense: every PR since `#490` adds a fragment instead of touching `CHANGELOG.md` directly, so there's no more competing insertion at this PR's old anchor point.

No source files touched. Full suite: **18,402 passed, 396 skipped, 5 deselected, 0 failed.**

blackcoderx:
@MakazhanAlpamys — flagging the 9 failing checks rather than acting on them myself.

They're the full `test` matrix (3 OS × 3 Python versions); `lint`, `type-check`, `mlx-smoke`, and `transformers-floor` are all green. All 9 fail on the identical 4 tests, all about `soup data stats`:

- `test_bugfixes.py::TestStatsHistogramWindows::test_stats_command_does_not_crash` — `assert 1 == 0`
- `test_bugfixes.py::TestStatsHistogramWindows::test_utf8_redirect_logic_on_cp1251_stdout` — `AttributeError: module 'plotext' has no attribute 'clear_figure'`
- `test_cli_subprocess.py::TestDataCommands::test_data_stats_jsonl` — `data stats failed (rc=1)`
- `test_data_tools.py::test_stats_command` — `assert 1 == 0`

The `AttributeError` names it: `plotext` looks like it dropped/renamed `clear_figure` in a newer release, and CI's `pip install -e ".[dev]"` picked it up.

**Confirmed pre-existing, not from this PR** — the identical 9-job failure is already on `main` itself at `ba4aea3` (this branch's own merge base) and the runs before it: https://github.com/MakazhanAlpamys/Soup/actions/runs/32645325022. Nothing in this diff touches `data stats` or `plotext`.

Not acting on it from this PR — let me know if you'd rather it get filed as a separate issue, or handled some other way.

Amix29:
Thanks for flagging this. No separate issue is needed: the plotext 6 breakage is already fixed in #507.

That PR adds a compatibility layer for both the plotext 5 module API and the plotext 6 figure API, including regression coverage for `soup data stats` and loss-curve rendering. Its full 3 OS × 3 Python matrix is green (13/13 checks). The fix is therefore ready; it only needs #507 to be merged into `main`.

MakazhanAlpamys:
﻿Root-caused and hotfixed on main a few minutes ago: the unpinned `plotext>=5.2.0` started resolving 6.0.0 upstream, which removed the plotting API (`clear_figure`/`hist`) that `soup data stats` uses — that was all 9 red cells, nothing in your or anyone's diff. Tracked as #522, capped `<6.0.0` in eb77dd1, CI rerunning. Thanks for flagging the failures instead of acting on them — that report landed while the cause was still unnamed and made the timeline unambiguous.

MakazhanAlpamys:
﻿Merged as squash on a fully green matrix (13/13). What made this one land cleanly, for the record:

- **Single source of truth held**: your round-2 rebase delegated the frozen-base half back onto `resolve_frozen_base_load_dtype` instead of keeping a parallel path - and went further than asked by collapsing the VRAM pre-flight's drifted full-FT classification into a shared `is_full_finetune()` predicate. That killed drift in both directions (it missed `lisa_enabled`/`r==0`, over-counted bare `freeze_layers`). This repo keeps getting bitten by two copies of one truth (#372, #392), so that extra collapse is the part I valued most.
- **Verified, not read**: mutation check - dropping `"torch_dtype": self._resolve_load_dtype(tcfg)` at the text site fails 6 tests in test_issue339; 180 targeted tests across the dtype/stream/seed suites green in an isolated worktree pinned to your branch; ruff clean.
- Fragment `changelog.d/0.73.3/471.fixed.md` was already in place and honestly discloses the carried-over H100 benchmark rather than implying a re-measure.

Welcome to the release notes when the next version ships.

PR #474, opened 2026-08-21
fix(replay): make max_points a hard cap on downsample output

### LINKED ISSUE #473: downsample() returns more than max_points rows, contradicting its own docstring
Found while reviewing #470, which merely exposes it — this is pre-existing and not that PR's doing.

`utils/replay.py::downsample` promises, in its own docstring:

> Return at most ``max_points`` rows preserving order + endpoints.

It does not. Measured:

```
rows = 5, max_points = 2  ->  len(out) = 3
```

The cause is the stride arithmetic plus the endpoint pin:

```python
stride = max(_MIN_STEP_DELTA, len(rows) // max_points)   # 5 // 2 = 2
```

A stride of 2 over 5 rows yields indices 0, 2, 4 — three rows — and the endpoint pin can add a fourth in other shapes. `max_points` is behaving as a *stride divisor*, not as a cap.

## Why it is worth an issue rather than a silent fix

`MAX_PLOT_POINTS` exists so `soup runs replay` does not hand a terminal plot more points than it can draw. Whether the real contract is "at most N" or "roughly N, endpoints guaranteed" is a decision, not a typo — and #470's new regression test now **pins the current formula**, so changing it is a deliberate act with a visible cost rather than an accident.

## Acceptance

- [ ] Decide which contract `downsample` actually has: a hard cap (fix the arithmetic, e.g. a ceil-based stride) or approximate-with-endpoints (fix the docstring).
- [ ] If the arithmetic changes, update `test_stride_lands_on_last_row_no_duplicate_pin` in the same commit — it currently asserts `out == [rows[0], rows[2], rows[4]]`, which encodes today's formula. Do **not** delete it: it is the only guard against an off-by-one on the stride (verified — a `+ 1` there passes 14 of 15 tests, including `test_long_capped`).
- [ ] Either way, a test pinning the chosen contract at a boundary shape (`len(rows)` just above `max_points`).

Low severity: the consequence is a chart with a few more points than intended, not a wrong number.

MakazhanAlpamys:
## What does this PR do?

Closes #473.

`downsample` promised "at most `max_points` rows" and did not deliver: the stride was
`len(rows) // max_points`, which is a **divisor, not a cap**. Five rows with `max_points=2`
gave a stride of 2, so `rows[::2]` returned three rows, and in other shapes the endpoint pin
appended a fourth.

**The contract I picked is the hard cap**, not a docstring softening. `MAX_PLOT_POINTS` exists
so `soup runs replay` never hands the terminal renderer more points than it can draw; a bound
that the function is free to exceed is not a bound.

## How

Not a ceil-based stride. That caps correctly but has a cliff: a series one row over the cap
(2001 rows, cap 2000) collapses to 1001 points, because `ceil(2000/1999) = 2` — a 2x fidelity
loss for being one row too long.

Instead, `max_points` evenly spaced indices across `[0, len(rows) - 1]`, both ends included:

```python
span = len(rows) - 1
slots = max_points - 1
return [rows[(index * span + slots // 2) // slots] for index in range(max_points)]
```

- returns exactly `min(len(rows), max_points)` rows — the cap is now also the target
- both endpoints always present, so first and final loss stay on the chart
- indices are strictly increasing (`span > slots` when `len(rows) > max_points`), so the
  endpoint-pin branch is not "made to fit under the cap" — it is deleted, along with
  `_MIN_STEP_DELTA`, which the new arithmetic makes structurally unreachable
- integer round-half-up rather than floats, so "the last index lands exactly on `span`" is
  provable rather than a floating-point courtesy

`max_points=1` is the one shape where both endpoints cannot fit. It returns the final row —
making the final loss visible is exactly what the old pin existed for. Documented in the
docstring.

## On the #470 guard

`test_stride_lands_on_last_row_no_duplicate_pin` is **kept, not deleted**, per the issue. It
survives almost verbatim: at `max_points=3` over 5 rows the new arithmetic lands on 0, 2, 4, so
`out == [rows[0], rows[2], rows[4]]` still holds and still fails on an off-by-one in the index
formula. Renamed to `test_last_index_lands_on_last_row_no_duplicate`, since there is no stride
and no pin left for the old name to describe.

Two existing tests were tightened from tolerances that encoded the old sloppiness:
`test_long_capped` `<= 210` -> `== 200`, and `test_default_cap` `<= MAX_PLOT_POINTS + 5` ->
`== MAX_PLOT_POINTS`.

## New tests

- `test_never_returns_more_than_max_points` — 30-case sweep over shapes x caps; this is the
  bug from the issue, and it failed at `assert 3 <= 2` before the fix
- `test_one_row_over_the_cap_drops_exactly_one_row` — 101 rows / cap 100 stays at 100; this is
  the test that a ceil-based stride would fail (it returns 51)
- `test_endpoints_survive_every_shape` — every `(total, max_points)` pair for totals 3..39
- `test_order_preserved_without_duplicates` — strictly increasing, no repeats
- `test_max_points_one_keeps_the_final_row` — pins the degenerate case

Beyond the suite, the contract was verified by brute force over every `(n, k)` in
`1..259 x 1..259`: `len(out) == min(n, k)`, indices strictly increasing, last row always
present.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

- [x] `pytest tests/ -v` passes — 18048 passed, 377 skipped, coverage 81% (gate 77%)
- [x] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed — no
      docs page documents the plot-point cap; `CHANGELOG.md` has the entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,7 +114,7 @@ reproducing 70+ versions of notes.
 ### Fixed
 
 - **`downsample` now returns at most `max_points` rows, which is what its
-  docstring has always promised (#473).** The stride was
+  docstring has always promised (#473 in #474).** The stride was
   `len(rows) // max_points` — a divisor, not a cap — so five rows with
   `max_points=2` came back with three, and the endpoint pin could add a
   fourth. Sample indices are now spread evenly across the series with both
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,21 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **`downsample` now returns at most `max_points` rows, which is what its
+  docstring has always promised (#473).** The stride was
+  `len(rows) // max_points` — a divisor, not a cap — so five rows with
+  `max_points=2` came back with three, and the endpoint pin could add a
+  fourth. Sample indices are now spread evenly across the series with both
+  endpoints included, so `soup runs replay` renders exactly
+  `min(len(rows), max_points)` points. Even spacing also avoids the cliff a
+  ceil-based stride would introduce: a series one row over the cap keeps
+  `max_points` points rather than roughly half of them. `max_points=1` is the
+  one shape where both endpoints cannot fit and returns the final row, which
+  is what the endpoint pin existed to guarantee. The off-by-one guard added
+  in #470 is kept, renamed to `test_last_index_lands_on_last_row_no_duplicate`
+  and re-pinned to the new arithmetic. Consequence of the old behaviour was a
+  chart with a few more points than intended, never a wrong number.
+
 - **`cut_ce.py` and `liger.py` now normalize path separators and match architecture
   keywords on the last path component only (#456 by @harshitthek in #458).**
   On Windows, `rsplit("/")` never split on backslashes, causing parent directory
--- a/src/soup_cli/utils/replay.py
+++ b/src/soup_cli/utils/replay.py
@@ -17,10 +17,6 @@ from typing import List, Optional, Sequence
 # even on long runs that logged tens of thousands of steps.
 MAX_PLOT_POINTS = 2000
 
-# Minimum step delta to keep when sub-sampling. Otherwise large stretches
-# can collapse into "step 100, 100, 100" duplicates after coarsening.
-_MIN_STEP_DELTA = 1
-
 
 @dataclass(frozen=True)
 class ReplaySummary:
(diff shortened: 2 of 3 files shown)


MakazhanAlpamys:
Merged as `10f7d1d`. Reviewed to the same standard as any outside PR — mutation, deleted-line audit, CI on the true head.

**The fix is guarded.** Restoring the old divisor stride:

```
stride = max(1, len(rows) // max_points); out = rows[::stride]; pin endpoint
21 failed, 28 passed
```

Including four parametrizations of `test_never_returns_more_than_max_points` and `test_one_row_over_the_cap_drops_exactly_one_row`.

**The deleted test lines are all strengthenings**, which I checked line by line because `-8` in `tests/` is the shape that has hidden quiet coverage removal here before:

```
-        assert len(out) <= 210  # MAX_POINTS + endpoint pin tolerance
-        assert len(out) <= MAX_PLOT_POINTS + 5
```

Both were slack that existed only because the function exceeded its own cap; they are now exact equality. And #470's test is **renamed, not removed** — `grep -c` confirms it survives — with a comment recording that the endpoint pin is gone but it remains the only guard against an off-by-one in the index arithmetic. That was #473's own acceptance criterion and it is met literally.

Choosing even spacing over a ceil-based stride is the right call and the test that proves it (`101 rows at cap 100 -> 100, not 51`) is the one a reader will thank you for.

## One finding: the term your comment singles out is unguarded, and the justification is wrong

```
# The `+ slots // 2` is integer round-half-up: floats would be exact at these
# sizes, but integer arithmetic keeps the endpoint landing provable.
```

Dropping that term — `(index * span) // slots` — **passes all 49 tests**.

I checked whether it is an equivalent mutant. It is not, in output:

```
shapes checked=1653  differing=1464          (89% of shapes sample different interior rows)
  n=4 k=3: half_up=[0, 2, 3]  down=[0, 1, 3]
  n=6 k=3: half_up=[0, 3, 5]  down=[0, 2, 5]
endpoints always land in BOTH variants: True
```

So the two variants disagree about which interior rows are drawn on 89% of shapes, while satisfying every property the tests assert — length, endpoints, strict ordering.

**The comment's stated reason is the part that is actually false**: the endpoint lands at `span` in both variants, because `slots * span // slots == span` regardless of the rounding term. What `+ slots // 2` buys is *symmetric* spacing, not endpoint landing.

Neither blocking — behaviour is contract-equivalent and the shipped variant is the better of the two. But a load-bearing-looking comment justifying a term that no test pins, with a reason that does not hold, is exactly what the next person deletes during a "simplification". Worth either a test that distinguishes the two spacings, or a comment saying what the term really does.

CHANGELOG at line 117 inside `[Unreleased]`, CI 12/12 on the true head. No `CONTRIBUTORS.md` entry, per convention — maintainer commits are not credited there.

Issue #338, opened 2026-08-07
AWQ and GPTQ export are unreachable from a clean install, and GPTQ has two code bugs


MakazhanAlpamys:
First CUDA-box run of `soup export --format awq|gptq` (#72). **Soup's AWQ code is correct** — the dependency set is not. GPTQ has real code bugs behind an uninstallable package.

## AWQ — code correct, pin impossible
At transformers **≤ 4.52.4** the export runs (52 s, 1915 MiB peak) and the artifact is **verified usable**: `AutoModelForCausalLM.from_pretrained` → `AwqConfig`, `WQLinear_GEMM`, fp16, generating *"The capital of France is Paris. The capital of Germany is Berlin."*

But 4.53.3 / 4.55.4 / 4.56.2 fail with `'Catcher' object has no attribute 'attention_type'`, and ≥ 4.57.0 fails at import (`PytorchGELUTanh` removed). Soup's own `trl 0.26.2` requires transformers ≥ 4.56.1, so **no single version satisfies both `[train]` and `[awq]`** — AWQ cannot be reached from `pip install "soup-cli[train,awq]"`.

Two smaller things found alongside:
- the `except ImportError` prints "autoawq not installed" when it **is** installed, hiding the real error
- with no `--calibration-data`, autoawq silently downloads a 214k-row dataset from the Hub

## GPTQ — uninstallable, and two bugs behind it
`auto-gptq>=0.7.0` has **no cp312 wheel** and the sdist fails three ways. Forced through on a python3.11 venv, two real defects appear:

- **P1 — the documented command crashes.** With no `--calibration-data`, `export.py:1075` passes a tokenizer into `quantize(examples: List[Dict[...]])` → `'Qwen2TokenizerFast' object is not iterable`. The `--help` also promises a "built-in sample" that does not exist. (AWQ's equivalent call *is* the correct API; GPTQ's is not.)
- **P2 — fails this issue's own acceptance test.** With calibration data it exits 0, but `save_quantized` writes `gptq_model-4bit-128g.safetensors`, so `from_pretrained` raises `OSError: no file named ... model.safetensors`. **The weights are fine** — the native loader works and renaming the shard makes `from_pretrained` work identically. Pure packaging defect.

## Fix shape
- decide whether `[awq]` is supportable at all against the `[train]` pin, and if not, say so in the docs instead of shipping an unreachable extra
- fix P1 (build real calibration examples, or require `--calibration-data` and fix the help)
[...]

PR #475, opened 2026-08-22
fix(export): GPTQ requires calibration data, standard shard name, real ImportError

AmirF194:
Refs #338. Covers the two GPTQ code bugs (P1, P2) plus the shared except-ImportError fix; whether `soup-cli[train,awq]` can ever resolve together is a packaging policy call and is left open, per the issue's own acceptance criteria.

## Root cause

- **P1:** with no `--calibration-data`, `_export_gptq` called `model.quantize(tokenizer)`. auto-gptq's `quantize()` expects tokenized examples (`List[Dict[str, torch.LongTensor]]`), not a bare tokenizer, and rejects one with `'Qwen2TokenizerFast' object is not iterable` (the issue's own traceback). Unlike `--format awq`, auto-gptq has no built-in fallback dataset, so GPTQ export now requires `--calibration-data` up front and rejects a file that yields zero usable samples.
- **P2:** `save_quantized` writes its own `gptq_model-<bits>bit-<group>g.safetensors` shard name, which `AutoModelForCausalLM.from_pretrained` does not look for. The exported directory now also carries a standard `model.safetensors` (a rename, since the weights themselves are fine per the issue).
- Both the AWQ and GPTQ `except ImportError` blocks printed a fixed "not installed" string even when the package imports but a transitive import inside it raised `ImportError` for an unrelated reason, discarding the real exception. They now check `exc.name` and surface the real cause when it does not match the package itself.

## Verification

- No auto-gptq/autoawq wheel installs on this box (matches the issue's own finding), so the fix is verified with a synthetic `auto_gptq`/`awq` module injected via `builtins.__import__`, mirroring this repo's own established pattern in `test_awq_gptq_export.py` (`_mock_import`). 8 new/updated tests; each pinned regression fails on `main` and passes on this branch (checked directly, stashing just the source change).
- Every changed line in `export.py` is covered (`--cov=soup_cli.commands.export --cov-report=term-missing`).
- Full file suite (49 tests) plus `ruff check` run clean on Python 3.10, 3.11 and 3.12 (the whole CI matrix), each in a fresh `python:X-slim` container with `pip install -e .` + `transformers` (no `torch`, since nothing here touches the GPTQ/AWQ model classes themselves).
- Not verified: an actual GPTQ quantization run (no installable auto-gptq here, same constraint the issue reports), and the `[train]`+`[awq]` pip conflict.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -187,6 +187,21 @@ reproducing 70+ versions of notes.
   un-run one — `pending` is what a report keeps when the process dies mid-arm
   and no handler runs, which is the case the incremental write exists for.
 
+- **`soup export --format gptq` crashed with no calibration data and, when it
+  did run, wrote a shard name the standard loader can't find
+  (#338 by @MakazhanAlpamys in #PLACEHOLDER).** With no `--calibration-data`,
+  `_export_gptq` called `model.quantize(tokenizer)`; auto-gptq's `quantize()`
+  expects tokenized examples, not a bare tokenizer, so this failed with
+  "object is not iterable". GPTQ export now requires `--calibration-data`
+  up front and rejects a file with zero usable samples, since auto-gptq has
+  no built-in fallback dataset (unlike AWQ). Separately, `save_quantized`
+  writes its own `gptq_model-<bits>bit-<group>g.safetensors` shard, which
+  `AutoModelForCausalLM.from_pretrained` does not look for; the exported
+  directory now also carries a standard `model.safetensors`. The shared
+  `except ImportError` blocks on both the AWQ and GPTQ paths also stopped
+  reporting a fixed "not installed" string when the package itself imports
+  fine but a transitive import inside it fails for an unrelated reason.
+
 ## [0.73.3] - 2026-08-18
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -189,7 +189,7 @@ reproducing 70+ versions of notes.
 
 - **`soup export --format gptq` crashed with no calibration data and, when it
   did run, wrote a shard name the standard loader can't find
-  (#338 by @MakazhanAlpamys in #PLACEHOLDER).** With no `--calibration-data`,
+  (#338 by @MakazhanAlpamys in #475).** With no `--calibration-data`,
   `_export_gptq` called `model.quantize(tokenizer)`; auto-gptq's `quantize()`
   expects tokenized examples, not a bare tokenizer, so this failed with
   "object is not iterable". GPTQ export now requires `--calibration-data`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -189,7 +189,7 @@ reproducing 70+ versions of notes.
 
 - **`soup export --format gptq` crashed with no calibration data and, when it
   did run, wrote a shard name the standard loader can't find
-  (#338 by @MakazhanAlpamys in #475).** With no `--calibration-data`,
+  (#338 by @AmirF194 in #475).** With no `--calibration-data`,
   `_export_gptq` called `model.quantize(tokenizer)`; auto-gptq's `quantize()`
   expects tokenized examples, not a bare tokenizer, so this failed with
   "object is not iterable". GPTQ export now requires `--calibration-data`
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -40,7 +40,7 @@ soup export --model ./output --deploy ollama  Export GGUF + auto-deploy to Ollam
 soup export --model ./output --format onnx    Export to ONNX
 soup export --model ./output --format tensorrt Export to TensorRT-LLM
 soup export --model ./output --format awq     Export to AWQ (4-bit)
-soup export --model ./output --format gptq    Export to GPTQ (4-bit)
+soup export --model ./output --format gptq --calibration-data cal.jsonl  Export to GPTQ (4-bit)
 soup deploy ollama --model m.gguf --name x    Deploy GGUF to Ollama
 soup deploy ollama --list                     List Soup-deployed models
 soup deploy ollama --remove <name>            Remove model from Ollama
--- a/src/soup_cli/commands/export.py
+++ b/src/soup_cli/commands/export.py
@@ -96,7 +96,7 @@ def export(
     calibration_data: Optional[str] = typer.Option(
         None,
         "--calibration-data",
-        help="Path to calibration JSONL for AWQ/GPTQ (default: use built-in sample)",
+        help="Path to calibration JSONL. Required for GPTQ; optional for AWQ.",
     ),
     calibration_samples: int = typer.Option(
         128,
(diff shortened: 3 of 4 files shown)


MakazhanAlpamys:
**The fixes are real and I verified them end-to-end against a live auto-gptq install — including getting GPTQ export working on this box, which nobody here had done before.** One required fix, and it is the one I care most about.

## Required: the CHANGELOG credits the wrong person — me

```
(#338 by @MakazhanAlpamys in #475)
```

I am the *issue* author. **You are the PR author.** Every other entry in the file uses `(#ISSUE by @PR_AUTHOR in #PR)` — line 31 is your own `(#367 by @AmirF194 in #461)`.

This matters more than a typo. v0.73.2 credited two people in its release notes while carrying work from eight, and our own checklist calls attribution "the only currency an open-source project pays its contributors in". Shipping your work credited to the maintainer is precisely the failure that rule exists to prevent, and I would rather catch it here than reconstruct it at release time — which is when four suffixes went wrong in v0.73.3.

Should read `(#338 by @AmirF194 in #475)`.

## What holds — every fix guarded, and P2 proven against the real thing

| mutation | result |
|---|---|
| restore the full pre-fix P1 (both guards off, `quantize(tokenizer)` fallback) | **KILLED** — 2 tests |
| `if not calib_texts:` -> `if False:` | **KILLED** — `test_export_gptq_empty_calibration_file_raises` |
| delete the `_standardize_gptq_shard_name` call | **KILLED** — `test_export_gptq_end_to_end_leaves_standard_name` |
| `if shard.exists() and not standard.exists():` -> `if False:` | **KILLED** — 2 |
| `if exc.name != "auto_gptq":` -> `if False:` | **KILLED** — the transitive-import test |
| `if exc.name != "awq":` -> `if False:` | **KILLED** |

**P2 verified with the decisive A/B** — same weights, only the shard name differs:

```
out_gptq   (post-fix: model.safetensors)                LOADED OK LlamaForCausalLM
out_prefix (pre-fix: gptq_model-4bit-128g.safetensors)  OSError: Error no file named
    pytorch_model.bin, model.safetensors, ... found in directory out_prefix.
```

That is issue #338's P2 traceback on the left and a clean load on the right. And no regression to the native loader — `AutoGPTQForCausalLM.from_quantized` still resolves both names, confirmed.

**A real-world instance of your own ImportError fix turned up during install**: `import auto_gptq` with a missing transitive dep raises `ModuleNotFoundError(name='gekko')`, which pre-fix printed "auto-gptq not installed" for an installed package. Your fix is not hypothetical.

**Scope discipline noted.** You wrote `Refs #338`, not `Closes`, and I confirmed there is no closing keyword — so **#338 correctly stays open** for the packaging half. Held to the scope you stated on the issue, you delivered exactly it. The dependency half is real and unchanged: `[gptq]` has no wheel on Python 3.12, and quantization only ran once I pinned `transformers==4.52.4` — the same ceiling #338 documents for AWQ.

Test-deletion audit: **16 lines, 0 test functions** (40 -> 49). Every deleted line is a re-indentation or signature edit inside a *retained* test — the three `assert_called_once()` lines and the `assert "my_model_gptq"` reappear one level deeper under the added patch. No assertion weakened.

## Two recommended, neither blocking

**1. One new test is environment-dependent** — `test_gptq_real_missing_package_still_shows_install_hint` does not mock the import, so it passes only where auto-gptq is absent:

```
with auto-gptq NOT importable (CI)   1 passed
with auto-gptq IMPORTABLE            assert 'not installed' in '... export plan ...'   1 failed
```

It fails on exactly the machine this feature is for. Your sibling tests already show the fix — mock `builtins.__import__` to raise `ModuleNotFoundError(..., name="auto_gptq")`. The pre-existing `test_export_gptq_import_error` has the same weakness; you touched it without hardening it.

**2. The message the PR advertises is not pinned.** Neutralising only the `cal_path is None` guard still refuses — the second guard catches it, which is good defence-in-depth — but the operator then gets `No usable calibration samples found in None` instead of your `GPTQ export requires --calibration-data...`. Both new tests assert only `pytest.raises(ClickExit)` and `from_pretrained.assert_not_called()`. One assertion on `--calibration-data` appearing in the output would close it.

**Docs:** `docs/commands.md:43` still says `soup export --model ./output --format gptq`, which now exits 1. Needs `--calibration-data <file>`, same as you already fixed in `--help`.

## Three things I found that are NOT yours

Recording them so they are not lost, and so nobody attributes them to this PR:

- The GPTQ path prints an unconditional `Security Notice: Loading model with trust_remote_code=True` **even when the flag was not passed**, and `AutoGPTQForCausalLM.from_pretrained` is called without `trust_remote_code` at all. A warning that fires when the risk is absent trains people to ignore it.
- `docs/models.md:117` maps gptq to `pip install "soup-cli[awq]"` — wrong extra.
- Default `--group-size 128` on SmolLM2 (hidden 576) produces a loadable but **broken** artifact; auto-gptq accepts it silently. Soup will happily emit an unusable GPTQ model.

I will file these separately unless you want any of them.

`ruff` clean, 49 passed, CI 12/12 on the true head, and no line this PR added is uncovered.

The attribution line, and this merges.

AmirF194:
Good catch on the attribution, fixed to credit you as PR author. Also hardened both env-dependent tests to mock the auto-gptq import instead of relying on it being genuinely absent (verified they now pass with a real importable stub too), added the --calibration-data assertion to the no-cal-data test, and updated the docs/commands.md example to include the now-required flag. Pushed to the same branch.

Thanks for the thorough pass, especially catching the shard load with the real weights.

MakazhanAlpamys:
Merged as `35808c3`. **All three asks closed, and I verified each by running it rather than reading the diff.**

The attribution first, since that was the required one: `(#338 by @AmirF194 in #475)`, line 192, inside `[Unreleased]`. Your comment said "credit **you** as PR author" — for the record the file now credits **you**, which is correct; the wording just came out inverted.

```
M1a: neutralise the cal_path guard   ->  FAILED test_export_gptq_no_calibration_data_raises
     (previously: 49 passed, message unpinned)

with auto_gptq stubbed IMPORTABLE    ->  3 passed
     (previously: 1 failed — the test passed only where the package was absent)

docs/commands.md:43                  ->  now carries --calibration-data
```

*(A note on my own run: my first attempt at M1a matched **two** sites — there is a `cal_path is None` in the AWQ path too — and my `assert count == 1` aborted the patch. Without that guard I would have run an unmutated tree and reported a false "survived". Second time I targeted line 1039 directly.)*

## What this PR is, on the record

**You got a real auto-gptq install running to prove P2**, which nobody here had done. That produced the decisive A/B — same weights, only the filename differing:

```
model.safetensors                     -> LOADED OK LlamaForCausalLM
gptq_model-4bit-128g.safetensors      -> OSError: Error no file named ... found
```

Left side is issue #338's traceback verbatim; right side is a clean load. And you checked the direction that would have been easy to miss: auto-gptq's own loader resolves both names, so nothing regresses for people using it.

**The scope discipline was unprompted and it is the thing I most want to keep.** You wrote `Refs #338`, not `Closes`, and said on the issue exactly which half you were leaving — the packaging half is real and unchanged: no wheel on Python 3.12, and quantization only runs with transformers pinned back to 4.52.4, the same ceiling #338 documents for AWQ. **#338 correctly stays open**, and I confirmed there is no closing keyword rather than trusting the body.

Your ImportError fix also turned out not to be hypothetical — installing auto-gptq surfaced exactly that shape via a missing `gekko`, an installed package reported as "not installed".

Test-deletion audit: 16 lines, **zero test functions** (40 -> 49). Every one is a re-indentation or signature edit inside a *retained* test.

## Three things I found that are NOT yours

Recording so they are not attributed here and not lost:

- The GPTQ path prints `Security Notice: Loading model with trust_remote_code=True` **unconditionally**, even when the flag was not passed — and `AutoGPTQForCausalLM.from_pretrained` is called without `trust_remote_code` at all. A warning that fires when the risk is absent trains people to ignore it.
- `docs/models.md:117` maps gptq to `pip install "soup-cli[awq]"` — wrong extra.
- Default `--group-size 128` on a model whose hidden size is not divisible by it produces a loadable but **broken** artifact; auto-gptq accepts it silently.

I will file these unless you want them.

Credited in `CONTRIBUTORS.md` (`e61c1a0`).

AmirF194:
That is a genuinely great writeup, thank you for verifying it that thoroughly and for the mutation-testing note on M1a, that is a sharp catch. Please go ahead and file the three you found; you have full context on all of them and I would rather they stay credited to your own repro than get diluted through mine.

PR #476, opened 2026-08-22
feat(trainer): extend LISA to continued pre-training (#307)

### LINKED ISSUE #307: Extend LISA to pretrain-family tasks (continued pre-training)
## Context
LISA (v0.71.34 / #267) is gated to `task: sft` only (mirroring Spectrum's `unfrozen_parameters`). Continued pre-training (`task: pretrain`) is the same full-FT-of-active-layers mechanism and would benefit identically.

## Proposed fix path
1. In `config/schema.py` `_validate_lisa_compat`, allow `task in {'sft','pretrain'}` instead of `== 'sft'`.
2. Add the `elif tcfg.lisa_enabled` branch + `attach_lisa_callback(...)` to `trainer/pretrain.py` (mirror the `trainer/sft.py` wiring exactly: leave the model fully trainable, `enable_input_require_grads`, attach the callback).
3. Tests: a `task: pretrain` + `lisa_enabled` config parses; the pretrain trainer wires the callback (source-grep + attach test).

## Acceptance criteria
- [ ] `task: pretrain` + `lisa_enabled` accepted by the schema
- [ ] pretrain trainer attaches LisaCallback (test)
- [ ] Live smoke: pretrain + LISA on SmolLM2-135M trains a few steps

ousamabenyounes:
Fixes #307

## What

LISA now supports `task: pretrain` as well as `task: sft`:

- the schema uses an explicit `_LISA_SUPPORTED_TASKS` allow-list;
- SFT and pretraining share `apply_lisa_setup()`, preventing their LISA wiring from drifting;
- pretraining skips PEFT for LISA, attaches `LisaCallback`, and reports the full model parameter count;
- LISA model loads use the shared full-fine-tuning-aware dtype resolver, preserving fp32 master weights, while ordinary pretraining LoRA loads retain the checkpoint-native frozen-base dtype;
- the end-to-end tensor comparison moves both operands to CPU, so the assertion is accelerator-neutral;
- release notes live in `changelog.d/0.73.3/476.added.md`; the released `CHANGELOG.md` is unchanged from `main`.

The command reference, PEFT/efficiency guide, schema description, and resolver documentation match the expanded task support.

## Acceptance criteria

- [x] `task: pretrain` + `lisa_enabled` is accepted by the schema
- [x] pretraining attaches `LisaCallback` and skips LoRA
- [x] a live tiny-Llama run changes only the sampled decoder layer
- [x] LISA loads fp32 master weights; plain LoRA retains `"auto"`
- [x] device-neutral end-to-end comparison
- [x] changelog fragment replaces the direct changelog edit

The review regression test was run before the dtype fix:

```text
FAILED TestPretrainLisaWiring::test_model_load_dtype_matches_whether_the_base_is_trainable[True]
assert ['auto'] == [torch.float32]
1 failed, 1 passed
```

After routing both trainers through `resolve_base_load_dtype`:

```text
15 passed
```

The final focused replay, including existing SFT dtype coverage and the complete LISA module, passed:

```text
176 passed
```

Changelog fragment validation passed separately: `17 passed`.

## Full local CI replay

An isolated `upstream/main` baseline and the final branch were run under the same pinned environment:

```text
baseline: 1 failed, 18534 passed, 263 skipped, 5 deselected
final:    1 failed, 18559 passed, 263 skipped, 5 deselected
```

The sole failure is identical on both sides:
`tests/test_v07129.py::TestReviewFixes::test_fuse_adapter_produces_dense_model`
(`SamConfig` tokenizer mapping in the installed transformers environment).

- Ruff: pass
- Dependency check: pass
- Changed production-line coverage: 100% (40/40)
- Independent PR review: pass
- OCR peer review: pass

## Remote CI

All 14 checks passed on `66ac804`: Linux, macOS, and Windows on Python 3.10–3.12, MLX smoke, lint, type-check, recipe validation, and transformers-floor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,21 @@ reproducing 70+ versions of notes.
 
 ### Added
 
+- **LISA now accepts `task: pretrain`, not `sft` alone (#307 by @ousamabenyounes in #476).**
+  Continued pre-training is the same full-fine-tune-of-a-rotating-set-of-decoder-layers
+  mechanism LISA was built for, so the sft-only gate (inherited from Spectrum's
+  `unfrozen_parameters`) was arbitrary. The schema task gate now reads a
+  `_LISA_SUPPORTED_TASKS` allow-list (`sft`, `pretrain`) and its refusal names every
+  accepted task instead of only rejecting yours; `trainer/pretrain.py` replaces its LoRA
+  path with LISA and attaches `LisaCallback`, and reports its parameter summary as `LISA`
+  counted off the raw parameters, since a LISA run has no `PeftModel` wrapper to ask
+  `get_nb_trainable_parameters`. Both trainers route through one
+  `peft_wiring.apply_lisa_setup`, so they cannot drift on what "LISA is on" means. The
+  rest of the gate is unchanged: `transformers` + `text` + `quantization: none`, mutually
+  exclusive with the LoRA feature flags, `freeze_layers`/`freeze_ratio` and
+  `unfrozen_parameters`. The released `[0.71.34]` block below still says `task: sft`,
+  which is what 0.71.34 shipped.
+
 - **`soup eval aider` runs Aider's Polyglot code-editing benchmark through its
   official Docker harness (#91 by @Amix29 in #482).** The command preflights
   Docker, the daemon, and the locally built benchmark image; mounts a prepared
@@ -105,7 +120,6 @@ reproducing 70+ versions of notes.
   `BUNDLED_SCORER_REVISION` + a locked fingerprint fail the suite if a bundled
   scorer's output moves without a revision bump. Registry `save_eval_result`
   stamps `details_json`; `soup ship --emit-evidence` includes the same stamp.
-
 - **Layer streaming now accepts Qwen3.5 MoE text checkpoints whose decoder
   layers do not expose exactly the same weight keys in every block (by
   @Shutaru in #426).** The sharder now records per-layer shard headers instead of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,21 +25,6 @@ reproducing 70+ versions of notes.
 
 ### Added
 
-- **LISA now accepts `task: pretrain`, not `sft` alone (#307 by @ousamabenyounes in #476).**
-  Continued pre-training is the same full-fine-tune-of-a-rotating-set-of-decoder-layers
-  mechanism LISA was built for, so the sft-only gate (inherited from Spectrum's
-  `unfrozen_parameters`) was arbitrary. The schema task gate now reads a
-  `_LISA_SUPPORTED_TASKS` allow-list (`sft`, `pretrain`) and its refusal names every
-  accepted task instead of only rejecting yours; `trainer/pretrain.py` replaces its LoRA
-  path with LISA and attaches `LisaCallback`, and reports its parameter summary as `LISA`
-  counted off the raw parameters, since a LISA run has no `PeftModel` wrapper to ask
-  `get_nb_trainable_parameters`. Both trainers route through one
-  `peft_wiring.apply_lisa_setup`, so they cannot drift on what "LISA is on" means. The
-  rest of the gate is unchanged: `transformers` + `text` + `quantization: none`, mutually
-  exclusive with the LoRA feature flags, `freeze_layers`/`freeze_ratio` and
-  `unfrozen_parameters`. The released `[0.71.34]` block below still says `task: sft`,
-  which is what 0.71.34 shipped.
-
 - **`soup eval aider` runs Aider's Polyglot code-editing benchmark through its
   official Docker harness (#91 by @Amix29 in #482).** The command preflights
   Docker, the daemon, and the locally built benchmark image; mounts a prepared
@@ -120,6 +105,7 @@ reproducing 70+ versions of notes.
   `BUNDLED_SCORER_REVISION` + a locked fingerprint fail the suite if a bundled
   scorer's output moves without a revision bump. Registry `save_eval_result`
   stamps `details_json`; `soup ship --emit-evidence` includes the same stamp.
+
 - **Layer streaming now accepts Qwen3.5 MoE text checkpoints whose decoder
   layers do not expose exactly the same weight keys in every block (by
   @Shutaru in #426).** The sharder now records per-layer shard headers instead of
--- a/changelog.d/0.73.3/476.added.md
+++ b/changelog.d/0.73.3/476.added.md
@@ -1,0 +1,14 @@
+- **LISA now accepts `task: pretrain`, not `sft` alone (#307 by @ousamabenyounes in #476).**
+  Continued pre-training is the same full-fine-tune-of-a-rotating-set-of-decoder-layers
+  mechanism LISA was built for, so the sft-only gate (inherited from Spectrum's
+  `unfrozen_parameters`) was arbitrary. The schema task gate now reads a
+  `_LISA_SUPPORTED_TASKS` allow-list (`sft`, `pretrain`) and its refusal names every
+  accepted task instead of only rejecting yours; `trainer/pretrain.py` replaces its LoRA
+  path with LISA and attaches `LisaCallback`, and reports its parameter summary as `LISA`
+  counted off the raw parameters, since a LISA run has no `PeftModel` wrapper to ask
+  `get_nb_trainable_parameters`. Both trainers route through one
+  `peft_wiring.apply_lisa_setup`, so they cannot drift on what "LISA is on" means. The
+  rest of the gate is unchanged: `transformers` + `text` + `quantization: none`, mutually
+  exclusive with the LoRA feature flags, `freeze_layers`/`freeze_ratio` and
+  `unfrozen_parameters`. The released `[0.71.34]` block still says `task: sft`, which is
+  what 0.71.34 shipped.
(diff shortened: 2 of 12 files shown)


MakazhanAlpamys:
**The feature is correct and genuinely guarded — I could not break it. The blocker is a one-line device bug in your own flagship test, and CI is red on three cells because of it.**

## The blocker

```
test (macos-latest, 3.10 / 3.11 / 3.12)   FAILURE
E   RuntimeError: Expected all tensors to be on the same device, but found at least two devices, mps:0 and cpu!
tests/test_v07134.py::TestPretrainLisaEndToEnd::test_a_pretrain_lisa_run_moves_only_the_sampled_decoder_layers
```

The same failure reproduces here on **CUDA** (`cuda:0 and cpu`), so it is not an Apple quirk — it is any machine with an accelerator, **including mine**. Root cause is in the test, not the product: `_pretrain_wrapper` builds the wrapper with `device="cpu"` and `resolve_device_map("cpu")` correctly loads the model on CPU, but `TrainingArguments` never sets `use_cpu`, so HF's `Trainer.train()` relocates the model to the accelerator. Your `before` snapshot is CPU, the post-train params are not, and `torch.equal` refuses to compare across devices.

One line fixes it, verified on GPU with no `CUDA_VISIBLE_DEVICES` override:

```python
if not torch.equal(param.detach().cpu(), before[name].cpu())
```
```
1 passed in 16.76s
```

Setting `use_cpu=True` in the fixture would also work and would honour the `device="cpu"` the fixture already asks for — your call which. Please re-check the macOS cells after pushing, since MPS is the one platform neither of us can test locally.

## What I could not break — seven mutations, all killed

| mutation | result |
|---|---|
| drop `attach_lisa_callback` in `pretrain.py` | **KILLED** — 2 wiring tests, **and the E2E test independently**: `assert 4 == 1` |
| never sample (`chosen = set(indices)`) | **KILLED** — 4 |
| select zero layers | **KILLED** — 5 |
| `apply_lisa_setup` always returns `False` (LoRA path under LISA) | **KILLED** — 8 |
| drop `enable_input_require_grads()` | **KILLED** — 1 |
| add an unwired task to `_LISA_SUPPORTED_TASKS` | **KILLED** — 3 |
| pretrain summary label `"LISA"` -> `"LoRA applied"` | **KILLED** — 2 |
| delete `apply_lisa_setup` from `sft.py` | **KILLED** — 1 |

**Row 1 is the one that matters.** The E2E test catches an attached-but-never-firing callback by observing that four layers moved when one was configured — so this is emphatically not a flag that validates and does nothing. That was the failure class I was most worried about and you closed it before I asked.

**The schema gate is right in both directions.** I ran 21 configs through the real loader: `task: pretrain` is accepted, and every previously-refused combination still refuses **by name** — dpo/grpo/orpo, unsloth, mlx, vision, 4bit, DoRA, VeRA, `freeze_layers`, `freeze_ratio`, `relora_steps`, `unfrozen_parameters`, `train_router_only`, `expand_layers`, and the footgun guard for `lisa_*` set while `lisa_enabled` is false. Three of those needed their own prerequisites satisfied first to force the LISA rule rather than an earlier one — I re-ran those separately rather than counting a refusal that came from the wrong validator.

**And I ran #307's acceptance criterion 3 for real**, on the CLI with SmolLM2-135M:

```
LISA: layerwise importance sampling (2 layer(s) every 4 steps, LoRA off)
LISA: 134,515,008 trainable / 134,515,008 total (100.00%)
Loss: 4.1472 -> 4.0296
```

The 90-line `pretrain.py` diff is almost entirely re-indentation under `if not apply_lisa_setup(...)`; I read it line by line and the LoRA path is byte-identical, with `test_lora_control_still_wraps_and_attaches_no_lisa_callback` as a real control. One deleted test line, a module-docstring bullet, replaced by an expanded one. Import-lightness holds: 34 passed, and `utils/lisa.py`'s ~4.4 s import stays off every light path.

CHANGELOG correctly under `[Unreleased]` with the right suffix, and docs updated in both places that state the gate — I grepped every tracked file for a surviving "LISA is sft-only" claim and found none. Leaving the released `[0.71.34]` block saying `task: sft` is also right: it is an accurate record of what shipped.

## Three notes, none blocking

- `LISA: 134,515,008 trainable / 134,515,008 total (100.00%)` is true at prep time but reads oddly against the docs' "peak optimizer memory is roughly embeddings + head + `lisa_num_layers`". A parenthetical would close it. Pinned by two tests, so it is a deliberate change either way.
- `apply_mod_if_configured` runs **outside** the LISA branch, so its comment about running "after `get_peft_model` so the routers are trainable" no longer describes that path — there is no `get_peft_model` under LISA. `sft.py` has the same shape, so this is consistent rather than new, but `use_mod` is not in LISA's mutual-exclusion list on either path and nobody has exercised MoD + LISA.
- You flagged the `elif` vs `if not` asymmetry between the two trainers yourself. Both are covered by mutations, so I would leave it — noting it only so a future reader does not "harmonise" them and lose a test.

One line and a green macOS run.

MakazhanAlpamys:
Checking in — a day on, nothing has changed here and I would rather unblock you than let this sit.

**State of play, precisely:** the feature itself is correct and I could not break it. Eight mutations, all killed by named tests, and #307's criteria are met live. What is holding it is **one line in your own test**, and the three red cells are all the same failure:

```
test (macos-latest, 3.10 / 3.11 / 3.12)   FAILURE
RuntimeError: Expected all tensors to be on the same device
```

`torch.equal` on tensors that are not co-located. A `.cpu()` on both sides fixes it. Ubuntu and Windows pass because MPS is only available on the macOS runners, which is why this is invisible from a Linux box.

There is also a conflict with `main` now — mechanical, from other merges, not from anything you did.

**Three ways forward, all fine by me:**

1. **You push the `.cpu()` and the rebase.** Best outcome — it stays your PR end to end, and it is genuinely two small edits.
2. **You say "go ahead" and I do both on your branch**, authorship preserved, and you review what I pushed. `maintainer_can_modify` is on, so I can.
3. **You tell me it is not a good week.** Then I will park it and pick it up later with credit intact. LISA-for-pretraining is worth having and I am not going to let it rot, but I also will not take it off you silently.

No deadline attached to this and no pressure implied — I am asking because a PR that is one line from green should not be sitting for a day, and you may simply not have seen that CI went red after your last push.

MakazhanAlpamys:
﻿Reviewed fully — verdict: merge after two fixes.

1. **Not rebased (34 commits behind main)** and the PR edits `CHANGELOG.md` directly — that now conflicts with main and violates the fragment convention: please replace with `changelog.d/0.73.3/476.added.md`.
2. **Semantic clash with #492 after rebase**: pretrain.py now loads its base through `resolve_frozen_base_load_dtype` on the premise the base stays frozen. With LISA the base trains full-FT, so pretrain+LISA would train bf16-loaded weights with no fp32 master weights while SFT+LISA loads fp32 — exactly the two-trainers drift the shared helper exists to prevent. Route pretrain's load dtype through a full-FT-aware resolver when you rebase.
3. LOW but real: the end-to-end AC test compares cuda:0 vs cpu tensors (`test_v07134.py:1492`) — green on CI only because runners have no GPU; snapshot after the move or compare on one device.

The gate work itself is exactly right: `_LISA_SUPPORTED_TASKS = ("sft", "pretrain")`, refusal names both tasks, quantization=none and mutual exclusions intact, callback attached in pretrain.py:259. Mutation check: reverting the allowlist kills 12 tests — the feature is genuinely guarded. Rebase + the dtype routing and this lands quickly.

ousamabenyounes:
Thanks for the precise review — all three points are addressed in `66ac804` after rebasing onto current `main`:

- moved the release note to `changelog.d/0.73.3/476.added.md` and restored `CHANGELOG.md` exactly;
- routed LISA pretraining through the shared full-FT-aware dtype resolver, so trainable base weights load as fp32 while the plain-LoRA control remains checkpoint-native;
- made the end-to-end tensor comparison device-neutral with `.cpu()` on both operands.

RED → GREEN for the dtype regression: `1 failed, 1 passed` (`'auto' != torch.float32`) → `15 passed`; the expanded focused replay is `176 passed`. Full isolated comparison is baseline `1 failed, 18534 passed` versus final `1 failed, 18559 passed`, with the same unrelated `SamConfig`/transformers failure on both sides. Changed production-line coverage is 100% (40/40).

Both independent review gates passed, and all 14 GitHub checks are green on `66ac804`.

MakazhanAlpamys:
﻿@ousamabenyounes - ping on #476: two items stand between you and a merge, both from my review comment - (1) rebase over current main (your branch edits CHANGELOG.md directly, which now conflicts), and (2) route pretrain'"'"'s load dtype through a full-FT-aware resolver so pretrain+LISA does not train bf16-loaded weights while SFT+LISA loads fp32. The gate work itself is done and mutation-verified - these are mechanical.

ousamabenyounes:
Rebased onto current main and pushed 7fd2f44. The PR is conflict-free, and the focused tests plus the full local iso-baseline replay passed. GitHub CI is still queued; I will address anything it surfaces.

MakazhanAlpamys:
Merged as 2794043. Thank you, Ben - #307 is closed.

What you did well: the rebase landed conflict-free on a main that had moved under you twice this week, and the dtype half was done exactly the way the tree now demands it - pretrain routes through the shared full-FT-aware resolver (`resolve_base_load_dtype`) instead of growing a second dtype decision, and SFT was moved onto the same path, so pretrain+LISA loads fp32 master weights precisely like SFT+LISA. One implementation in the tree is what the #492 review asked for and you delivered it without being walked through it. Sharing `apply_lisa_setup()` between the two trainers closes the drift class at the same time.

What I ran before merging:
- 175 targeted tests (test_issue339_load_dtype / test_issue491_frozen_base_load_dtype / test_v07134), including the real-torch LISA end-to-end - all green.
- Mutation 1: forced pretrain+LISA into the frozen/bf16 branch - died by name on `TestPretrainLisaWiring::test_model_load_dtype_matches_whether_the_base_is_trainable[True]`.
- Mutation 2: disabled the LISA dispatch branch - died on 8 named tests across TestPretrainLisaWiring and TestSharedLisaSetup.
- Verified your head 7fd2f44 against merge-base == current main; CI 14/14 cells green.

CONTRIBUTORS.md updated with this entry - your fourteenth merge here.

PR #479, opened 2026-08-22
feat(mcp): serve MCP over sse and streamable http, not just stdio

### LINKED ISSUE #296: soup mcp serve: add SSE / HTTP transport (v1 is stdio-only)
## Context

`soup mcp serve` (v0.71.28) ships with a **stdio** transport only — great for local clients (Claude Code, Cursor, Cline, Continue) that spawn the server as a subprocess. Remote / multi-client setups want an HTTP-based transport.

## Ask

Add an opt-in `--transport sse|http` (default `stdio`) to `soup mcp serve`, using the official `mcp` SDK's SSE / streamable-HTTP server helpers.

## Fix path

- `src/soup_cli/mcp_server/server.py`: factor the `build_server(specs)` step (already transport-agnostic) and add an `run_sse_server` / `run_http_server` alongside `run_stdio_server` using the SDK's SSE/HTTP transport (`mcp.server.sse` / streamable-http).
- `src/soup_cli/commands/mcp.py`: add a `--transport` option + `--host`/`--port` for the network transports.
- **Security:** the network transports add a listener — bind to localhost by default, and require an auth token (mirror `soup ui`'s Bearer-token pattern). The path-containment / sanitization guards in `registry.py` already apply regardless of transport.

## Acceptance criteria

- [ ] `soup mcp serve --transport sse --port N` serves the same 16-tool registry over SSE.
- [ ] localhost bind + auth token by default; documented.
- [ ] a scripted SSE client smoke (initialize → list_tools → call) passes.
- [ ] docs (`docs/commands.md` MCP section) updated.

MakazhanAlpamys:
## What does this PR do?

Closes #296. `soup mcp serve` gains `--transport sse` and `--transport http`. **stdio stays the default and its code path is untouched** — a client that spawns Soup as a subprocess sees no change at all.

## Registry parity, proven rather than asserted

Both transports serve the *same* registry as stdio. The end-to-end test compares the advertised tool names against `build_registry(...)` instead of a hardcoded count, so a subset cannot creep in and adding a tool later cannot silently make the check stale:

```python
expected = {spec.name for spec in build_registry(allow_mutating=False, allow_execute=False)}
assert names == expected
```

## Security: three gates, because a listener is the risky part

**1. Bearer token on every HTTP request, no opt-out.** Compared with `secrets.compare_digest` over the whole header, so the scheme is checked in constant time along with the token. There is deliberately no `--no-auth`: a loopback port is reachable by every process on the machine, not just by the operator.

The token is validated through the existing `utils/qr_url.py::validate_token` (16-128 urlsafe-base64 chars), so `soup ui` and `soup mcp serve` agree on what a token *is* rather than growing a second bespoke format. It travels in the header only — no query-string fallback that could be captured in a proxy or access log.

**2. DNS-rebinding protection (the SDK's `TransportSecuritySettings`).** `421` on a foreign `Host`, `403` on a foreign `Origin`. This is the gate the Bearer token *cannot* be: a web page the operator merely visits attaches no `Authorization` header, but its request still arrives at the port. Auth is the outer layer and rebinding the inner one, and the tests authenticate on purpose before checking `Host` so that ordering is pinned rather than assumed.

**3. Loopback by default.** `127.0.0.1:8765`. A non-loopback `--host` prints a warning; a wildcard bind prints a second one saying the Host check has degraded to accepting anything, because with no single advertised name there is nothing to pin. `allowed_hosts_for()` returns `["*"]` there rather than pretending the check still does work.

## The `[mcp]` floor moves, and the number is measured

`mcp>=1.2.0,<2` → `mcp>=1.10.0,<2`. Not read off a changelog — I installed the wheels and checked which module each version actually ships:

| mcp | `server/sse.py` | `server/streamable_http_manager.py` | `server/transport_security.py` |
|---|---|---|---|
| 1.2.0 | yes | no | no |
| 1.6.0 | yes | no | no |
| 1.7.0 | yes | no | no |
| 1.8.0 | yes | **yes** | no |
| 1.9.4 | yes | yes | no |
| 1.10.0 | yes | yes | **yes** |

`--transport sse` alone would have run on the old floor. Streamable HTTP would not, and neither would rebinding protection — and a listener whose origin checking silently disappears on an older SDK is worse than a resolver error the operator can act on. The `<2` cap is unchanged; #322 is the 2.x migration and this does not touch it.

**No new dependency:** `mcp` already requires `starlette`, `uvicorn`, `sse-starlette` and `httpx-sse`, so `[mcp]` does not grow.

## Flags that cannot apply are refused, not ignored

`--host` / `--port` / `--auth-token` default to `None` and resolve per transport, so passing one with `--transport stdio` exits non-zero with a message naming the flags. Silently accepting a flag that does nothing is the failure mode worth avoiding here.

## Tests — 30 in the new file

`build_asgi_app()` is factored out so auth and rebinding are exercised through `httpx.ASGITransport` without binding a socket:

- 401 on a missing header, a wrong token, and a right token under the wrong scheme, across `/sse`, `/messages/` and `/mcp`
- a correct token reaches the transport (400 from the SDK, i.e. past the gate)
- `421` on a foreign `Host` while authenticated; the bound host/port accepted
- token resolution: generated tokens validate, differ between calls, an operator token is preserved, a short one is rejected
- CLI: network-only flags refused under stdio, unknown transport refused, help lists the transports

Then one test binds a **real ephemeral port**, runs uvicorn in a thread, and drives `initialize → list_tools → call_tool` through the SDK's own SSE client — that is the acceptance criterion as written, and an in-memory transport cannot demonstrate it. A companion test confirms a wrong token cannot open a session over that same real socket.

## Three defects the tests caught before review

1. **IPv6 loopback would have been refused entirely.** `allowed_hosts_for("::1", 8765)` joined with a bare colon produced `::1:8765`; a real client sends `Host: [::1]:8765`, so every IPv6 loopback client would have hit a `421`. Now bracketed, with a test pinning it.

2. **The stdio path depended on network symbols it never uses.** Importing `DEFAULT_NETWORK_HOST` / `DEFAULT_NETWORK_PORT` alongside `run_stdio_server` broke `test_v07128.py::TestServePlumbing::test_allow_flags_reach_runner`, which injects a stand-in module exposing `run_stdio_server` alone. The test was right and the code was wrong; the imports are now split per branch.

3. **A `--help` assertion read raw CLI output.** `test_cli_help_assertions_are_ansi_safe.py` caught it — exactly the asymmetry that guard documents: Rich disables colour on Windows, so it passed locally and would have turned the Linux/macOS cells red. Routed through a `_plain()` helper mirroring `test_v07302.py`.

Also removed two sources of drift a reviewer would have flagged and CI would not: an unused `TRANSPORTS` constant, and duplicated `/sse` / `/mcp` literals in the CLI (now imported from the server module). The transport *list* is still duplicated in the CLI on purpose — `--transport` must be validated before the SDK is imported, so a missing `[mcp]` extra yields a flag error rather than an install message — and `TestTransportListsAgree` pins the two lists together so the deliberate duplication cannot become a real one.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [x] `pytest tests/ -v` passes — **18087 passed, 377 skipped, 0 failed**, coverage 81% (gate 77%)
- [x] Updated relevant docs — `docs/commands.md` gains two command-list lines and a "Remote or multi-client" section covering the `.mcp.json` header shape and each security property; `CHANGELOG.md` has the entry
- [x] No breaking changes — stdio is byte-for-byte the same path; the only behaviour change for an existing user is the `[mcp]` floor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,43 @@ reproducing 70+ versions of notes.
 
 ### Added
 
+- **`soup mcp serve` gains network transports: `--transport sse` and
+  `--transport http` (#296).** v1 was stdio-only, which suits a client that
+  spawns Soup as a subprocess but leaves remote and multi-client setups with
+  nothing. Both new transports serve the *same* registry — the end-to-end test
+  compares the advertised tool names against `build_registry` rather than a
+  hardcoded count, so a subset cannot creep in. stdio remains the default and
+  is untouched.
+
+  Adding a listener is the risky part, so it is gated three ways. Every HTTP
+  request needs `Authorization: Bearer <token>`, compared with
+  `secrets.compare_digest`, with no opt-out — a loopback port is reachable by
+  every process on the box. The token is validated by the existing
+  `utils/qr_url.py::validate_token`, so `soup ui` and `soup mcp serve` agree on
+  what a token is instead of growing a second format, and it travels in the
+  header only: no query-string fallback that could land in an access log. The
+  SDK's DNS-rebinding protection is switched on (`421` on a foreign `Host`,
+  `403` on a foreign `Origin`), which is the gate the token cannot be — a page
+  the operator merely visits sends no `Authorization` header but its request
+  still reaches the port. Binding off loopback warns, and a wildcard bind warns
+  again that the Host check has nothing left to pin.
+
+  `--host` / `--port` / `--auth-token` are refused under `--transport stdio`
+  rather than silently ignored, and the ASGI app is built by a
+  `build_asgi_app()` factory so the auth and rebinding behaviour is tested
+  through an in-process transport without binding a socket; one further test
+  binds a real ephemeral port and drives initialize -> list_tools -> call_tool
+  through the SDK's own SSE client.
+
+  The `[mcp]` extra floor moves `1.2.0` -> `1.10.0`, measured against the
+  published wheels rather than a changelog: `server/streamable_http_manager.py`
+  first appears in 1.8.0 (absent in 1.7.0) and `server/transport_security.py`
+  in 1.10.0 (absent in 1.9.4). `--transport sse` alone would have run on the
+  old floor; rebinding protection would not, and a listener whose origin
+  checking silently disappears on an older SDK is worse than a resolver error.
+  The `<2` cap is unchanged — #322 is the 2.x migration. No new package: `mcp`
+  already requires starlette, uvicorn, sse-starlette and httpx-sse.
+
 - **`soup data best-of-n` can sample candidates from Ollama or vLLM providers
   (#299 by @Faisal01011 in #466).** The existing local Transformers `--base` path stays
   the default, while `--provider ollama|vllm --model <m> [--base-url <url>]`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ reproducing 70+ versions of notes.
 ### Added
 
 - **`soup mcp serve` gains network transports: `--transport sse` and
-  `--transport http` (#296).** v1 was stdio-only, which suits a client that
+  `--transport http` (#296 in #479).** v1 was stdio-only, which suits a client that
   spawns Soup as a subprocess but leaves remote and multi-client setups with
   nothing. Both new transports serve the *same* registry — the end-to-end test
   compares the advertised tool names against `build_registry` rather than a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,17 @@ reproducing 70+ versions of notes.
   still reaches the port. Binding off loopback warns, and a wildcard bind warns
   again that the Host check has nothing left to pin.
 
+  `--allow-execute` is refused with either network transport, and that refusal
+  is the one behavioural change to an existing flag. Gated execution (#297)
+  spawns real training / export processes; behind a listener a leaked Bearer
+  token would mean process execution rather than plan disclosure, and stdio is
+  a pipe to a client the operator already started, which is a different trust
+  boundary. The refusal is made twice on purpose: once in the CLI so the
+  operator gets a readable message, and once in `build_asgi_app()` so a direct
+  caller cannot put an executing registry behind a listener either. The stdio
+  banner also stopped claiming "execution disabled" -- that string predated
+  #297 and was false from the moment execution shipped.
+
   `--host` / `--port` / `--auth-token` are refused under `--transport stdio`
   rather than silently ignored, and the ASGI app is built by a
   `build_asgi_app()` factory so the auth and rebinding behaviour is tested
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -186,6 +186,8 @@ soup ci init [--data d.jsonl --suite s.yaml --evidence ev.json] [--config soup.y
 soup mcp serve                                MCP server over stdio (drive Soup from Claude Code / Cursor / Cline; requires [mcp] extra) (v0.71.28)
 soup mcp serve --allow-mutating               Also expose plan-only train_start / export tools (never execute) (v0.71.28)
 soup mcp serve --allow-execute                Implies --allow-mutating; enables train_execute / export_execute via server confirmation tokens
+soup mcp serve --transport sse [--host H --port N]  Serve the same registry over HTTP+SSE instead of stdio; binds 127.0.0.1 and requires a Bearer token (#296)
+soup mcp serve --transport http [--auth-token T]    Same over the streamable-HTTP transport (/mcp); --auth-token pins the token instead of generating one (#296)
 soup shrink --model <id|path> --drop-ratio 0.25 --calib c.jsonl -o shrunk  Depth-prune least-important layer block + SHIP/DON'T-SHIP ppl verdict (exit 0/2/1) (v0.71.29)
 soup shrink ... --drop-layers N --heal h.jsonl --heal-steps 200 --device cpu  Drop N layers + distill-heal (fuse LoRA back to one dense model)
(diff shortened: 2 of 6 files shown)


MakazhanAlpamys:
Reviewed to the same bar as an outside PR. **The auth, rebinding, path-containment and stdio-parity work is genuinely good and I verified all of it by running it. I am blocking on one CRITICAL and one HIGH, both in the same place.**

## CRITICAL — `--allow-execute` over a network transport, with a banner that says the opposite

There is **no guard** between the flag and the listener. `commands/mcp.py`:

```python
# line 155-157
# Execution is intentionally not implemented yet. Keep the raw
# ``allow_execute`` value for the server/registry, while its stronger
# opt-in also enables the existing plan-only mutating tools.
if allow_execute:
    mode = "mutating tools ENABLED (plan-only; execution disabled)"   # line 160
...
    f"[dim]soup mcp serve - {transport} transport - {mode}[/]\n"       # line 202
...
run_network_server(..., allow_execute=allow_execute, ...)              # line 222
```

**That comment is stale and it is the root of this.** Execution *is* implemented — `mcp_server/execution.py` exists, `build_registry(*, allow_mutating, allow_execute, execution)` takes an `ExecutionManager`, and `train_execute` / `export_execute` are registered at `registry.py:1090-1091`. The comment predates #297; the banner is derived from it; and #479 propagates that banner onto a **network** listener.

So `soup mcp serve --transport http --allow-execute` starts a socket that spawns processes while printing **"execution disabled"**. `docs/commands.md` on the same page says the opposite. A remote client holding only the Bearer token can run the plan -> token -> execute chain and get a live PID — the #297 confirmation token is a *plan-integrity* control, not an authorization boundary against a network peer. Over stdio that was fine because the process boundary **was** the authorization. This removes it and puts nothing in its place.

The stale string is on `main` (`mcp.py:67,72`), so it is not introduced here — but this is the PR that makes it dangerous, and it is worse coming from the person who shipped #297.

**Required:** refuse `--transport sse|http` + `--allow-execute`, or require a distinct third opt-in with a loud warning. And fix the banner so it stops claiming execution is disabled.

## HIGH — the execute gate on the network path has zero tests

```
server.py:87   allow_execute=allow_execute,  ->  allow_execute=True,
30 passed
```

Hard-wiring every network server open passes the whole file. That is the CRITICAL arriving *by accident instead of by flag*, with CI green.

The reason the flagship socket test cannot see it: it compares tool **names**, and `train_execute` / `export_execute` are always *listed* — only the handler swaps. `grep "allow_execute=True" tests/` returns nothing. **The single most dangerous flag on the new listener is untested on the path that makes it dangerous.**

*(Two process notes on my own run: my first M6 patch put the comment before the comma and produced 22 failures — a syntax error, not a result; I redid it and confirmed the patched line. And my first attempt at reading `mcp.py` through a mock failed on a wrong symbol name, so I read the source instead, which is what surfaced the stale comment.)*

## What I verified and holds

- **Bind is loopback.** `DEFAULT_NETWORK_HOST = '127.0.0.1'`, both transports observed `LISTENING` on 127.0.0.1 only and unreachable via the LAN IP. A non-loopback `--host` warns twice. **But `DEFAULT_NETWORK_HOST` is never exercised** — flipping it to `0.0.0.0` keeps 30 tests green, because every test passes `host=` explicitly. Security claim #3 is untested.
- **Auth is solid.** 10-case matrix per transport: no header, empty, bare token, wrong, empty token, `Basic`, double-space — all 401. No query-string fallback. Middleware is outermost, so unrouted paths are gated too. `compare_digest` over the whole header; token is `token_urlsafe(32)` = 256 bits; goes to **stderr only** (verified with separate pipes), absent from the 401 body.
- **Gate parity network vs stdio is exact**, message for message, and `--allow-mutating` alone does not unlock execution. The gate logic is right; what is missing is ceremony around the flag.
- **Registry guards survive the network path** — traversal, absolute, UNC and NUL all refused path-free; `_sanitize` strips ESC/BEL on the wire.
- **stdio really is untouched** — the diff is docstrings only; 136 passed / 1 skipped on the existing MCP suites.
- **No new dependency**: `mcp>=1.2.0` -> `>=1.10.0`; starlette/uvicorn/sse-starlette are already hard requires of `mcp`. Light CLI path 34 passed.
- Zero test deletions. `ruff` clean. CI 12/12 on the true head.

## Smaller items

- **MEDIUM** — no `Origin` -> 403 test. The claim is made in the PR body, the CHANGELOG and the docs, and covered nowhere; I confirmed it behaves correctly (403 on foreign origins, 421 on foreign Host, IPv6 bracketing works), so this is coverage, not a defect.
- **LOW** — `Bearer` scheme comparison is case-sensitive; RFC 7235 says the scheme is case-insensitive, so a compliant client sending `bearer` gets a silent 401. Fails closed.
- **LOW** — `--port` takes `-1`, `0`, `70000` unchecked; `--host` echo uses `escape()` only, so raw ESC bytes reach the terminal (4 of them, measured). This repo has `_for_terminal` in seven command modules for exactly that, and CLAUDE.md says `escape()` does not cover ESC.
- **LOW** — the malformed `http://::1` origins (IPv6 bracketing applied to `allowed_hosts` but not `allowed_origins`); dead entries, fails closed.
- **NIT** — a comment references "v0.73.4", a version that does not exist yet.
- The branch is 2 commits behind `main`, so a plain `git diff origin/main HEAD` reads as if it deletes 297 test lines. It does not — that is main-only work. Worth rebasing so the next reader is not alarmed.

## What could not be run

No genuine off-host client — single box, so "not reachable from the LAN" was tested against this host's own LAN IP. Windows only. `train_execute` was not driven to a real training run; the spawn was proven via `export_execute` (PID + log on disk), the same `ExecutionManager.execute` path.

PR #480, opened 2026-08-22
fix(stream): keep Apple Silicon host store on CPU

### LINKED ISSUE #434: Layer streaming: the pinned host store lands on 'mps' on Apple Silicon, so 'the base lives in CPU RAM' is not true there

Amix29:
## Summary

- disable optional and required host pinning at the runtime boundary for MPS targets, including direct callers
- make `RamSource` reject storage that is not genuinely on CPU, or a requested pinned allocation that is not pinned
- document the experimental Apple Silicon path and update non-CUDA `stream_pin` messaging

Fixes #434.

## Reproduction and scope

Measured on a MacBook Pro M4 Max with 128 GB unified memory:

- tested PyTorch 2.0.1 through 2.6.0 samples raise when asked for pinned CPU memory without CUDA
- tested PyTorch 2.7.0 through 2.12.1 samples, plus an unreleased 2.13.0 development build, return an `mps:0` tensor for `torch.empty(device="cpu", pin_memory=True)`, while `is_pinned()` is false
- a 256 MiB reproduction increased `torch.mps.current_allocated_memory()` by exactly 256 MiB
- with this patch, the source tensors remain on CPU, the two layer buffers live on MPS, `runtime.pinned` is false, and the CPU-to-MPS copy is exact

The regular `soup train` setup already gated pinning on CUDA. The unsafe case was the lower-level runtime default/direct-call path, plus the lack of a postcondition at the source boundary.

Apple Silicon support remains explicitly experimental. Unified physical memory means the CUDA capacity and throughput figures do not transfer, and this PR does not claim that streaming fits a larger model or runs faster than resident MPS training.

## Additional hardware checks

- toy Qwen3.5 LoRA streamed vs resident MPS: forward output, loss, and all 4 LoRA gradients matched exactly
- toy Qwen3.5 NF4 with bitsandbytes 0.50.1: forward output, loss, and all 8 LoRA gradients matched exactly

These checks support correctness of the small reproduction only; they are not a broad production-support claim.

## Validation

- `486 passed, 42 skipped` across the full layer-streaming test set
- `git diff --check`

## Contributor identity

I changed my GitHub account from `Sir-Lysander` to `Amix29`; this PR and its commit are submitted from the new account.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,7 +121,8 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
-- **Layer streaming now keeps its host store on CPU on Apple Silicon (#434).**
+- **Layer streaming now keeps its host store on CPU on Apple Silicon
+  (#434 by @Amix29 in #480).**
   PyTorch 2.7+ can return an MPS tensor for
   `torch.empty(device="cpu", pin_memory=True)`, while `is_pinned()` remains
   false. Direct runtime callers could therefore place the whole frozen base in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,18 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **Layer streaming now keeps its host store on CPU on Apple Silicon (#434).**
+  PyTorch 2.7+ can return an MPS tensor for
+  `torch.empty(device="cpu", pin_memory=True)`, while `is_pinned()` remains
+  false. Direct runtime callers could therefore place the whole frozen base in
+  the MPS allocator and still report a pinned RAM store, even though the normal
+  `soup train` setup already disabled pinning outside CUDA. The runtime now
+  disables both optional and required pinning when the target is MPS, and
+  `RamSource` independently refuses any allocation that is not genuinely CPU
+  memory (or claims pinning without being pinned). MPS proceeds experimentally
+  with a pageable CPU source and MPS layer buffers; CUDA pinning behaviour is
+  unchanged.
+
 - **`downsample` now returns at most `max_points` rows, which is what its
   docstring has always promised (#473 in #474).** The stride was
   `len(rows) // max_points` — a divisor, not a cap — so five rows with
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -248,6 +248,17 @@ soup train --config soup.yaml
 
 **How it works.** LoRA adapters + their gradients + optimizer state stay resident in VRAM (they are small). The frozen base lives in CPU RAM, page-locked when the machine allows it, and is streamed: each decoder layer is copied into one of two pre-allocated VRAM buffers on a dedicated CUDA stream while the previous layer is still computing, so the load overlaps the compute. Each layer is read **twice** per step — once in the forward pass and once when the backward pass recomputes it — because `dL/dx = Wᵀ · dL/dy` needs the weights to reach the layers below. That is physics, not an implementation detail, and it is why streaming costs time.
 
+**Apple Silicon is experimental.** With `backend: transformers`, MPS uses a pageable CPU
+source and MPS layer buffers; host pinning is disabled. PyTorch 2.7+ may otherwise turn
+`torch.empty(device="cpu", pin_memory=True)` into an MPS tensor, charging the entire base
+to the MPS allocator while `is_pinned()` is still false (#434). Soup refuses that state at
+the source boundary and also disables pinning before allocation. Apple Silicon has unified
+physical memory, so the CUDA capacity and throughput numbers below do not transfer: only
+the MPS allocator's streamed weights are bounded by the buffer pool, while the CPU source
+still consumes unified memory. No claim is made yet that streaming fits a larger model or
+runs faster than resident MPS training. `backend: mlx` remains a separate, incompatible
+model-loading path and is rejected with `stream_layers`.
+
 The tradeoff: **1.43× slower than resident training**, measured at 0.5B — the only apples-to-apples comparison available on the reference box, because 1.5B and above cannot run resident there at all.
 
 ### NF4 streaming (`quantization: 4bit`)
@@ -314,11 +325,12 @@ refusing:
 |---|---|
 | RAM tier on CUDA | pins, or **refuses** naming the store size |
 | Disk tier (base does not fit in RAM, weights stream from NVMe) | announces that pinning does not apply, proceeds |
-| CPU (no CUDA device) | announces that page-locking is a host-to-device optimization, proceeds |
+| Non-CUDA target (CPU or MPS) | announces that CUDA host pinning does not apply, proceeds with a pageable CPU source |
 
 Refusing on those two would brick the large-model runs the disk tier exists for, and
 would make `stream_pin: true` uncommittable to a `soup.yaml` shared between a GPU box and
-a CPU box. The RAM tier is where the flag has real semantics, and there it still refuses.
+a non-CUDA box. The CUDA RAM tier is where the flag has real semantics, and there it still
+refuses.
 
 Set while `stream_layers: false` the key is rejected as a footgun, like the other
 `stream_*` keys.
--- a/src/soup_cli/config/schema.py
+++ b/src/soup_cli/config/schema.py
@@ -3107,11 +3107,10 @@ class TrainingConfig(BaseModel):
             "is worth up to 6.56x measured throughput, so a silent fallback "
             "spends the whole margin the feature exists to provide. On the disk "
             "tier (the base does not fit in RAM, so there is no RAM store to "
-            "page-lock) and on CPU (page-locking is a host-to-device transfer "
-            "optimization and there is no device) pinning is INAPPLICABLE rather "
-            "than unsatisfiable: 'true' is announced and the run proceeds "
-            "without it, so the key stays committable to a config shared between "
-            "a GPU box and a CPU box."
+            "page-lock) and on non-CUDA targets (CPU or MPS) pinning is "
+            "INAPPLICABLE rather than unsatisfiable: 'true' is announced and the "
+            "run proceeds with a pageable CPU source, so the key stays committable "
+            "to a config shared between CUDA and non-CUDA boxes."
         ),
     )
 
--- a/src/soup_cli/trainer/stream_setup.py
+++ b/src/soup_cli/trainer/stream_setup.py
@@ -472,14 +472,15 @@ class StreamingSetupMixin:
             use_rslora=tcfg.lora.use_rslora,
         )
 
-        # #366 review — on CPU there is no CUDA device, so page-locking (a
-        # host->device transfer optimization) is inapplicable. An explicit
-        # stream_pin=true is honoured by saying so, not by dropping it silently.
+        # #366 / #434 — CUDA host pinning is inapplicable on every non-CUDA
+        # target. An explicit stream_pin=true is honoured by saying so, not by
+        # dropping it silently. On MPS the pageable CPU source is also what keeps
+        # the frozen base out of the accelerator allocator.
         if tcfg.stream_pin is True and not on_cuda:
             console.print(
                 "[yellow]training.stream_pin=true, but no CUDA device is present: "
-                "page-locking is a host-to-device transfer optimization and does "
-                "not apply on CPU. Proceeding without it.[/]"
+                "CUDA host pinning does not apply to this target. Proceeding with "
+                "a pageable CPU source.[/]"
             )
 
         model, runtime = build_streamed_model(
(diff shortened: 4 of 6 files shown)


MakazhanAlpamys:
**The fix is correct and the PR body is one of the more honest I have reviewed — you separate measured from unmeasured and explicitly decline the performance claim you could have made. The blocker is that the headline gate has no test holding it in place.** I ran every mutation myself and re-verified the three below by hand before writing this up.

## What I ran

Isolated worktree, `PYTHONPATH` pinned to the branch (an editable install here resolves to the main checkout, so mutations "survive" by default — I confirmed `soup_cli.__file__` pointed inside the worktree first). Every mutation applied in binary mode against CRLF, replacement count asserted, `ast.parse` checked, and grep-confirmed in the file before running.

- `tests/test_qwen35_streaming.py`: **15 passed, 1 skipped** (the skip is your new MPS test)
- Wider positive control across the four streaming files: **412 passed, 8 skipped**
- `ruff check src/soup_cli/ tests/` clean; `test_cli_startup_is_light.py` 34 passed
- **No test deleted or weakened** — the tests file is `57 / 0` on `--numstat`

**Pre-fix red control passes**, which is what makes the reproduction honest:
```
>       with pytest.raises(RuntimeError, match="requested a CPU tensor.*meta"):
E       Failed: DID NOT RAISE <class 'RuntimeError'>
FAILED ...::test_ram_source_refuses_a_non_cpu_allocation
```

## Killed mutations

| Mutation | Result |
|---|---|
| `if dst.device.type != "cpu":` → `if False:` | **killed** — `test_ram_source_refuses_a_non_cpu_allocation` |
| drop `and on_cuda` from `require_pin` (`stream_setup.py:500`) | **killed** — `test_stream_pin_threads_require_pin_into_the_runtime_and_announces_cpu` |
| remove the non-CUDA `stream_pin` announcement (`stream_setup.py:479`) | **killed** — same test |

I also confirmed your CHANGELOG's candour: `stream_setup.py:493-499` on the merge-base already read `pin=plan.pinned and on_cuda` before this PR, so the shipped `soup train` path is behaviourally unchanged and what you are adding is a hardened boundary for direct callers. Saying that out loud rather than overselling the fix is the right call, and it is why I trust the rest of the body.

## Blocking: the gate the PR is named after is untested

**1. MEDIUM — disabling the MPS gate entirely changes nothing.** `src/soup_cli/utils/layer_stream_runtime.py:1616-1618`. I replaced

```python
if str(device).startswith("mps"):
    pin = False
    require_pin = False
```

with `if False:` and got **15 passed, 1 skipped** — unchanged. The only test that touches it, `test_mps_target_keeps_the_host_store_on_cpu`, is `skipif not torch.backends.mps.is_available()`, so it skips on every machine I or CI-on-Linux/Windows can run, and will *first execute* on the macOS cells.

This is closeable without a Mac, using the pattern already in this repo: monkeypatch `layer_stream_runtime._build_source`, call `install_streaming(..., device="mps", pin=True, require_pin=True)`, and assert the captured `pin` / `require_pin` are both `False`. `test_v07203.py:1436-1452` does exactly this for the CUDA side with a `captured[...]` dict.

## Two more of the same shape, please, in the same push

**2. MEDIUM — the one executable new test cannot discriminate the bug it was written for.** `tests/test_qwen35_streaming.py:509-517` simulates the misplaced allocation with `device="meta"`. I narrowed the guard to `if dst.device.type == "meta":` — a guard that would let the **actual #434 `mps` tensor straight through** — and got **15 passed, 1 skipped**.

The shipped guard is right; the test simply pins a narrower property than the one that matters. Parametrising `_misplaced_empty` over `("meta", "mps")`, or a second variant returning a fake `mps`-typed device, restores the discrimination.

**3. MEDIUM — the `is_pinned()` postcondition has no test.** `layer_stream_runtime.py:343-348`. Neutralising it leaves **15 passed, 1 skipped**. This is the guard that catches precisely the state your CHANGELOG describes — "returns an `mps:0` tensor while `is_pinned()` is false" — so it is load-bearing for the PR's own claim and nothing holds it up. Same monkeypatch style: hand back a real CPU tensor allocated with `pin_memory=False` while `pin=True` was requested, assert `RuntimeError` matching `pageable memory`.

## Smaller

**4. LOW — the MPS branch drops an explicit `require_pin=True` silently, against this module's own stated rule.** The sibling disk-tier branch (`:1760-1770`) exists specifically to announce inapplicable pinning; its docstring says pinning is *"announced, never dropped in silence (#366 review)"*. `console` is in scope at your gate. CLI users get a message from `stream_setup`, but a direct caller passing `require_pin=True` gets silence — and the new MPS test currently **asserts** that silence, pinning it in.

**5. LOW — please add the credit yourself.** `CHANGELOG.md:124` reads `(#434)`. The file's convention, used on nine neighbouring entries, is `(#434 by @Amix29 in #480)`. I would rather you write it than have me add it at release time — reconstructing authorship during a release is exactly when it gets skipped, and this file is the one people actually read.

**6. NIT — "tested PyTorch 2.7.0 through 2.13.0".** There is no 2.13.0 release that I can find. If those were nightlies, worth saying so. Nothing in the diff depends on it.

**7. NIT** — on a CUDA box the new device error surfaces through `_build_source`'s `except (RuntimeError, MemoryError)` as "could not page-lock the N GB RAM store", with the real cause only in the `from exc` chain. Unreachable from the CLI on MPS, so cosmetic.

## One thing I want to flag rather than block on

The docs change widens the advertised surface: on MPS, `resolve_stream_dtype` returns `float32` (`layer_stream.py:218`) and the VRAM pre-flight returns early (`stream_setup.py:649`), so **a streamed MPS run currently gets no fit check at all**, and `quantization: 4bit` would drive `_quantize_nf4` with `quant_device="mps"`. All of that predates you and none of it is yours to fix here. But #434 asked "does streaming support Apple Silicon at all?" and this PR answers "yes, experimentally" in prose. If you agree, I would rather file those two as follow-ups now than have the docs be the only place the limits are recorded.

## What I could not verify, stated plainly

**No Apple Silicon here — Windows + RTX 3050, torch 2.5.1.** `test_mps_target_keeps_the_host_store_on_cpu` skipped in every run I made. I did not reproduce `torch.empty(device="cpu", pin_memory=True)` returning `mps:0`, did not confirm `is_pinned()` is false there, and cannot confirm your M4 Max bit-exactness checks. #434's own evidence suggests MPS *is* available on GitHub's macOS runners, so that test will likely execute for the first time in CI — its assertions on `runtime.pool.buffers` being `mps` and on `load_async`/`wait` are an unvalidated-by-me risk that only the macOS cells will settle.

CI is green on all 13 cells at `23a14eb`. Fix **1** and I will merge; 2, 3 and 5 in the same push if you can. All three are the same shape — a correct guard with nothing holding it in place — and you clearly already know how to write the tests, since the three you did write all kill their mutations.

Amix29:
Addressed the requested follow-ups in commit 173e2a1.

- Added a CPU-only regression test at the `install_streaming` boundary. It monkeypatches `_build_source`, calls the runtime directly with `device="mps"`, `pin=True`, and `require_pin=True`, and asserts that both pinning arguments reaching `_build_source` are `False`. This should kill the proposed `if False:` mutation.
- Extended the non-CPU allocation test to cover both fake `meta` and fake `mps` devices.
- Added a postcondition test that returns a real pageable CPU tensor when pinning was requested and verifies the `returned pageable memory` error.
- Added an explicit degradation warning for direct MPS callers that request required pinning.
- Updated the changelog credit to `#434 by @Amix29 in #480`.
- Clarified in the PR body that 2.13.0 was an unreleased development build, not a stable PyTorch release.

Validation:

- `tests/test_qwen35_streaming.py`: 19 passed, including the real MPS hardware test on the M4 Max
- adjacent `stream_pin` / `RequirePin` tests in `tests/test_v07203.py`: 7 passed
- `git diff --check`: clean

I kept the separate MPS fit-check and NF4 quantization-device concerns out of this focused fix, as suggested.

MakazhanAlpamys:
**Merged** as `e0fee61`, third CONTRIBUTORS.md bullet in (`37c64e6`). That is three merged today.

## I re-ran all three mutations rather than reading your summary

| Mutation | Before | Now |
|---|---|---|
| `if str(device).startswith("mps"):` → `if False:` | 15 passed | **`test_mps_gate_disables_pinning_for_direct_callers`** |
| narrow the guard to `== "meta"` | 15 passed | `test_ram_source_refuses_a_non_cpu_allocation[mps]` |
| neutralise the `is_pinned()` postcondition | 15 passed | `test_ram_source_refuses_pageable_memory_when_pinning_was_requested` |
| silence the `require_pin` announcement | *(new)* | `test_mps_gate_disables_pinning_for_direct_callers` |

Baseline **18 passed, 1 skipped** (was 15 + 1).

**The first row is the one that matters, and you solved it the right way.** The old test was `skipif not torch.backends.mps.is_available()` — so on this Windows box, and on the Linux CI cells, it *skipped*, and a skipped test is the same colour as a passing one. Your replacement monkeypatches `_build_source` and asserts the captured `pin`/`require_pin`, so it **actually failed here** under the mutation. That is the difference between coverage and green-because-skipped, and it is a distinction plenty of experienced people miss.

I also read every deleted line in the test diff, because a shrinking test file is where I always look. Three removed, all upgrades — most notably the `torch.empty` stub now raises `AssertionError("RamSource allocations should request pin_memory")` on the non-pin path, so it additionally pins that RamSource *always* requests pinning. You made the test stricter while fixing it.

And you took the `require_pin` announcement (my LOW #4) rather than leaving it — the module's own rule is that pinning is "announced, never dropped in silence (#366 review)", and the disk-tier branch already followed it. Consistency with a rule the codebase states about itself is worth more than the line of code.

## What I want to name

**Your CHANGELOG says the shipped `soup train` path already gated pinning on CUDA.** I verified that — `stream_setup.py:493-499` on the merge base reads `pin=plan.pinned and on_cuda` — so the user-visible behaviour was already safe and this fixes the lower-level boundary. You could have framed this as a user-facing bug fix and nobody would have checked. Declining a bigger claim than your evidence supports is the thing I trust most in a contributor, and it is why I read the rest of your PRs faster.

## One LOW left as a follow-up

`layer_stream_runtime.py:1620-1626` — the `console is None` → `logger.warning(message)` fallback is untested; replacing it with `pass` leaves 18 passing. A library caller embedding the runtime with no console can lose the announcement with nothing noticing. One `caplog` assertion closes it. Not blocking: the #366 rule is enforced on the console path, which is the path `stream_setup` uses.

## What I could not verify, plainly

**No Apple Silicon here.** `test_mps_target_keeps_the_host_store_on_cpu` skipped on every run I made, so the PyTorch-2.7 pathology, the real `runtime.source.store` device placement, and your streamed-vs-resident bit-exactness checks in bf16 and NF4 are unreproduced by me. What I verified is the platform-independent decision logic — the gate, the guard, the postcondition, the announcement — which is what all three findings were about. Your M4 Max results are taken as reported, and the honesty of the rest of your reporting is the reason that is a reasonable thing for me to do.

Thank you — three merges, and each one closed its findings on the first push.

PR #481, opened 2026-08-22
feat(monitor): support Apple Silicon via powermetrics

### LINKED ISSUE #99: v0.44.1: soup monitor — Apple Silicon support via powermetrics
v0.44.0's `soup monitor` requires nvidia-smi. On Apple Silicon (M-series) we currently print a yellow advisory. Need a parallel `query_powermetrics` utility that parses `sudo powermetrics --samplers gpu_power -i 1000 -n 1 --format text` (or similar) and renders the same Rich table.

Note: powermetrics requires sudo on macOS — UX needs to handle that gracefully (fall back to Activity Monitor advisory if not available).

Amix29:
## Summary

- detect Apple Silicon before entering the existing NVIDIA monitor path
- collect one bounded sample from the fixed `/usr/bin/powermetrics` binary and parse its NUL-separated plist output into the existing `GpuSample` table
- report GPU utilization and estimated power while leaving NVIDIA-only VRAM, memory-utilization, and temperature fields unavailable
- reuse an explicitly cached sudo credential through `sudo -n`, never prompt for or read a password, and provide an actionable Activity Monitor fallback
- preserve the NVIDIA implementation unchanged and document the Apple Silicon workflow

## Safety and compatibility

- fixed absolute executable paths, list arguments, and `shell=False`
- one sample per invocation, bounded parser input, finite/range checks, and malformed-output handling
- lazy behavior: no new dependency and no import-time hardware access
- Python 3.10, 3.11, and 3.12 targeted coverage

## Validation

- `mypy src/soup_cli/utils/gpu_monitor.py src/soup_cli/commands/monitor.py`
- `pytest tests/test_issue99_apple_monitor.py tests/test_v0440_part_a.py tests/test_v0440_review_followups.py -q -o addopts=`: 127 passed
- full non-smoke suite: 18,178 passed, 286 skipped, 6 deselected; 81.15% coverage
- the one deliberately excluded test was `test_find_quantize_binary_not_found`, because this development Mac genuinely has `/opt/homebrew/bin/llama-quantize` installed
- MacBook Pro M4 Max, 128 GB unified memory, 40-core GPU, macOS 26.5.2: Apple Silicon dispatch and the real no-sudo-ticket path were exercised; the installed Apple binary exposes the expected `gpu`, `idle_ratio`, `gpu_energy`, `processor.gpu_power`, `elapsed_ns`, and `hw_model` plist fields

A positive privileged snapshot was completed on the M4 Max after an interactive `sudo -v`: `soup monitor --once --refresh 0.25` rendered a live `Mac16,5` sample at 61.5% GPU utilization and 0.7 W, then exited normally. The end-to-end observation is recorded on #99. The implementation itself never attempts to automate or consume the administrator credential.

Closes #99.

Contributor continuity: I changed my GitHub account from `Sir-Lysander` to `Amix29`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,14 @@ reproducing 70+ versions of notes.
 
 ### Added
 
+- **Native Apple Silicon telemetry for `soup monitor` (#99 by @Amix29).**
+  The monitor now reads bounded plist output from macOS `powermetrics` and
+  renders GPU utilization and power in the existing Rich table. It reuses an
+  explicitly cached sudo credential through non-interactive `sudo -n`, never
+  reads a password, and gives an actionable Activity Monitor fallback when
+  permission or telemetry is unavailable. NVIDIA-only VRAM, memory-utilization,
+  and temperature fields remain unavailable rather than being guessed.
+
 - **`soup data best-of-n` can sample candidates from Ollama or vLLM providers
   (#299 by @Faisal01011 in #466).** The existing local Transformers `--base` path stays
   the default, while `--provider ollama|vllm --model <m> [--base-url <url>]`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ reproducing 70+ versions of notes.
 
 ### Added
 
-- **Native Apple Silicon telemetry for `soup monitor` (#99 by @Amix29).**
+- **Native Apple Silicon telemetry for `soup monitor` (#99 by @Amix29 in #481).**
   The monitor now reads bounded plist output from macOS `powermetrics` and
   renders GPU utilization and power in the existing Rich table. It reuses an
   explicitly cached sudo credential through non-interactive `sudo -n`, never
--- a/docs/backends-and-ops.md
+++ b/docs/backends-and-ops.md
@@ -636,7 +636,17 @@ soup monitor --refresh 0.5  # faster polling
 soup monitor --once         # single snapshot, no Live panel
 ```
 
-Calls `nvidia-smi` via list-args subprocess (no shell), 5s timeout, list of `GpuSample` rows rendered into a Rich table. Apple Silicon prints a yellow advisory pointing at Activity Monitor / `powermetrics`; native Apple Silicon support lands in v0.44.1.
+On NVIDIA systems, Soup calls `nvidia-smi` via a list-args subprocess (no shell)
+with a 5-second timeout. On Apple Silicon, it reads GPU utilization and power
+from `/usr/bin/powermetrics --samplers gpu_power --format plist`. Run `sudo -v`
+in a terminal before starting the monitor: Soup uses `sudo -n`, so it can reuse
+the cached credential without ever prompting for or reading a password. If the
+credential or utility is unavailable, the command exits with an Activity
+Monitor fallback rather than reporting an NVIDIA error.
+
+macOS does not expose NVIDIA-style dedicated VRAM, memory-utilization, or GPU
+temperature fields through this sampler. Those columns therefore remain `—`
+instead of guessing values from unified memory or unrelated thermal sensors.
 
 
 ## Soup Fetch — Bundled Examples
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -212,7 +212,7 @@ soup bench <model> --p50 --p95                Bench with tail-latency percentile
 soup bench <model> --backend auto             Auto-detect transformers/mlx backend (v0.53.9)
 soup serve --reasoning-parser deepseek-r1     Strip <think> blocks from responses (v0.53.9)
 soup doctor [--nccl] [--disk]                 Check environment (optionally check NCCL bandwidth, media type; --disk ~9s cold / ~2.4s warm)
-soup monitor                                  Live GPU monitor: util / temp / VRAM / power per GPU
+soup monitor                                  NVIDIA / Apple Silicon GPU monitor: util / temp / VRAM / power
 soup quickstart [--dry-run]                   Full demo
 soup plugins list|install|enable|disable      Manage Soup plugins
 soup llama cli|mtmd-cli|gguf-split|server ... Proxy to the llama.cpp binaries
--- a/src/soup_cli/commands/monitor.py
+++ b/src/soup_cli/commands/monitor.py
@@ -2,7 +2,7 @@
 
 Renders a Rich panel with one row per detected GPU: Util / Temp / VRAM /
 Power. Polls `nvidia-smi` (Linux/Windows/CUDA) at the configured refresh
-rate. Apple Silicon variant is a stub note in v0.44.0.
+rate, or `powermetrics` on Apple Silicon.
 """
 
 from __future__ import annotations
@@ -18,8 +18,10 @@ from rich.table import Table
 
 from soup_cli.utils.gpu_monitor import (
     GpuSample,
+    PowermetricsStatus,
     detect_apple_silicon,
     query_nvidia_smi,
+    query_powermetrics,
 )
 
 console = Console()
@@ -65,6 +67,60 @@ def _build_table(samples: list[GpuSample]) -> Table:
     return table
 
 
+def _powermetrics_error(status: PowermetricsStatus) -> str:
+    if status is PowermetricsStatus.PERMISSION_DENIED:
+        return (
+            "[yellow]Apple GPU metrics require permission to run "
+            "`/usr/bin/powermetrics`.[/]\n"
+            "Run [bold]sudo -v[/] in a terminal, then rerun `soup monitor`. "
+            "Soup uses `sudo -n`: it never prompts for or reads your password."
+        )
+    if status is PowermetricsStatus.UNAVAILABLE:
+        return (
+            "[yellow]`/usr/bin/powermetrics` is unavailable on this Mac. "
+            "Use Activity Monitor → Window → GPU History.[/]"
+        )
+    if status is PowermetricsStatus.INVALID_OUTPUT:
+        return (
+            "[yellow]`powermetrics` returned no usable Apple GPU sample. "
+            "Its plist schema may not expose `gpu_power` on this Mac.[/]"
+        )
+    return "[yellow]`powermetrics` failed or timed out while reading Apple GPU metrics.[/]"
+
+
+def _monitor_apple_silicon(refresh: float, once: bool) -> None:
+    title = "Soup GPU Monitor — Apple Silicon"
+    result = query_powermetrics(refresh)
+    if result.status is not PowermetricsStatus.OK:
+        console.print(_powermetrics_error(result.status))
+        raise typer.Exit(code=1)
+
+    samples = list(result.samples)
+    if once or not samples:
+        console.print(Panel(_build_table(samples), title=title))
+        return
+
+    with Live(
+        Panel(_build_table(samples), title=title),
+        refresh_per_second=max(1.0, 1.0 / refresh),
+        screen=False,
+        console=console,
+    ) as live:
+        try:
+            while True:
+                # powermetrics itself blocks for the requested sample window;
+                # sleeping here as well would double the refresh interval.
+                fresh = query_powermetrics(refresh)
+                if fresh.status is not PowermetricsStatus.OK:
+                    live.update(Panel(_powermetrics_error(fresh.status), title=title))
+                    # Failure paths can return immediately; avoid a busy loop.
+                    time.sleep(refresh)
+                    continue
+                live.update(Panel(_build_table(list(fresh.samples)), title=title))
+        except KeyboardInterrupt:
+            console.print("[dim]exit[/]")
+
+
 def monitor(
     refresh: float = typer.Option(
         2.0,
@@ -80,18 +136,14 @@ def monitor(
 ) -> None:
     """Live GPU monitor: Util / Temp / VRAM / Power per GPU.
 
-    Requires nvidia-smi on PATH. On Apple Silicon use Activity Monitor or
-    powermetrics — full Apple Silicon support lands in v0.44.1.
+    Uses nvidia-smi on NVIDIA systems and powermetrics on Apple Silicon.
     """
     if not (0.25 <= refresh <= 30.0):
         console.print("[red]--refresh must be in [0.25, 30][/]")
         raise typer.Exit(code=2)
     if detect_apple_silicon():
-        console.print(
-            "[yellow]Apple Silicon detected — `soup monitor` is "
-            "Apple-Silicon-aware in v0.44.1.[/]\n"
-            "Use Activity Monitor → Window → GPU History for now."
-        )
+        _monitor_apple_silicon(refresh, once)
+        return
     ok, samples = query_nvidia_smi()
     if not ok:
         console.print(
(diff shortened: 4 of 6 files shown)


MakazhanAlpamys:
**Reviewed, and the security posture is the strongest part of this PR — I could not break it.** I ran the mutations myself rather than taking the description on trust, and separately re-verified the three findings below by hand before writing them up. Holding the merge on **one** of them; the rest are cheap.

## What I ran

Isolated worktree, `PYTHONPATH` pinned to the branch (an editable install here resolves to the main checkout, so mutations "survive" by default — verified `soup_cli.__file__` pointed inside the worktree before trusting a single result).

- `test_issue99_apple_monitor.py` **31 passed**; with `test_v0440_part_a.py` + `test_v0440_review_followups.py`: **126 passed, 1 skipped**
- `ruff check src/soup_cli/ tests/` clean; `test_cli_startup_is_light.py` + `test_cli_help_assertions_are_ansi_safe.py` green
- Pre-fix red control at `e61c1a0`: the new tests are uncollectable (`ImportError: PowermetricsResult`), so the reproduction is honest
- `import soup_cli.commands.monitor` → `torch` **not** in `sys.modules`, and no import-time hardware access. Your "lazy behavior" claim checks out.

**Your "preserve the NVIDIA implementation unchanged" claim is true and I verified it line by line:** every deleted line under `src/` is the v0.44.1 stub or a docstring — zero NVIDIA source lines touched. No test deleted or weakened.

## Mutations that were killed — this is genuinely guarded

| Mutation | Result |
|---|---|
| `if False and detect_apple_silicon()` | **killed** — `test_monitor_routes_apple_silicon_before_nvidia`, `test_monitor_apple_permission_error_is_actionable` |
| drop `-n` from the sudo argv | **killed** — `test_query_powermetrics_non_root_uses_non_interactive_sudo` |
| remove the finite / non-negative check | **killed** — 4 tests (`rejects_invalid_idle_ratio[-0.1]`, `rejects_invalid_power[-1/nan/inf]`) |

I also went at the parser adversarially: a billion-laughs entity bomb, an XXE `file:///` reference, 5,000-deep nesting and embedded NUL bytes all return `[]` without raising. Fixed absolute paths mean `PATH` cannot influence the executable, `shell=False` with list args means the `--sample-rate` value (an `int` derived from an interval already validated finite and in `[0.25, 30]`) cannot inject, and stderr is pattern-matched but **never printed**. That is a careful piece of work.

## Blocking: one guard is missing on the decision the whole feature turns on

**1. MEDIUM — nothing pins the non-Apple → NVIDIA route.** `src/soup_cli/utils/../commands/monitor.py:144`. I mutated `if detect_apple_silicon():` → `if True or detect_apple_silicon():` and got **126 passed, 1 skipped** — unchanged.

That mutation sends *every Linux and Windows user* down the powermetrics path, and no test notices. The *detector* is covered by the pre-existing `test_detect_apple_silicon_matches_platform_pair`, but the **call site** is not, and the call site is the one line that decides which implementation runs for everyone. Mirror your existing routing test with the sense flipped: `detect_apple_silicon → False`, and `query_powermetrics → pytest.fail(...)` so it is impossible to reach.

## Two more I would like in the same push

**2. MEDIUM — the size-cap test cannot fail.** `tests/test_issue99_apple_monitor.py:113` asserts `parse_powermetrics_plist(b"x" * (CAP + 1)) == []`. But `b"x" * N` is not a valid plist, so it returns `[]` through the *malformed-document* path whether the cap exists or not. Deleting the cap at `gpu_monitor.py:221` leaves the suite green. Proof, using a **valid** oversized plist (1,050,062 bytes):

```
with cap    -> []
without cap -> [GpuSample(index=0, name='Apple Silicon (Mac16,5)', util_gpu_pct=75.0, ..., power_w=12.5)]
```

The cap itself is correct — it is just unguarded. Build the fixture with `plistlib.dumps` and pad a string value.

**3. MEDIUM — the subprocess timeout is unguarded.** I deleted `timeout=interval + 5.0,` (`gpu_monitor.py:285`) and got **126 passed, 1 skipped**. That line is the only thing standing between a wedged sampler and a `soup monitor` that hangs forever, which is precisely the class of regression that must not be able to pass silently.

## Nits

**4. LOW** — `stdin=DEVNULL` on the privileged subprocess (`gpu_monitor.py:281`). `sudo -n` will not prompt, so the risk is low, but this repo sets it on every gated-execution subprocess (`mcp_server/execution.py`) and consistency is free here.

**5. LOW** — the 1 MB cap is checked at `:221`, *after* `capture_output=True` has already buffered the whole stream. Defence-in-depth only (exploiting it needs root to replace `/usr/bin/powermetrics`) and it fails safe, so: your call.

**6. LOW** — `monitor.py:110-119`: the success branch has no sleep, resting on the comment's assumption that powermetrics blocks for the sample window. The failure branch does sleep. If that assumption is ever false on some Mac, the loop respawns `sudo` as fast as it can. You have the hardware to confirm it; I do not.

**7. NIT** — `CHANGELOG.md:17` reads `(#99 by @Amix29)`. The house format includes the PR: `(#99 by @Amix29 in #481)` — same as the `(#299 by @Faisal01011 in #466)` entry a few lines down. It is what makes the credit traceable back to the change.

**8. NIT** — `assert "--sample-rate" in argv and "250" in argv` tests membership, not adjacency; `argv[argv.index("--sample-rate") + 1] == "250"` says what you mean.

## What I could not verify, stated plainly

**This box is Windows 11 + an RTX 3050. There is no Apple Silicon here, so no line of the Apple path has ever executed against real hardware in my review.** The fixtures are built from your implementation's assumptions, which means the tests and the code could be consistently wrong together and every one of my green runs would look identical. Four things only you can settle:

1. That the real plist actually nests `gpu_power` under `processor` and `idle_ratio` under `gpu`. This is the single largest residual risk.
2. That `--handle-invalid-values` is accepted by the shipped binary — an unrecognised flag makes every run exit non-zero and land in `FAILED`.
3. That powermetrics really blocks for the sample window (finding 6).
4. That the strings in your permission-marker list (`:293-299`) are the ones a real `sudo -n` refusal prints. A mismatch silently downgrades `PERMISSION_DENIED` to a generic `FAILED` and the user loses the actionable `sudo -v` hint — a failure that is invisible precisely when someone needs the message.

A short note in the PR confirming 1, 2 and 4 against your M4 Max is enough; I am not asking for more automation around a dialog macOS is right to block.

Fix **1** and I will merge — 2 and 3 in the same push if you can, since all three are the same shape: a real guard with nothing holding it in place. Thank you for the care on the security side; the parser hardening in particular is above what I would have asked for.

Amix29:
Addressed the review in `4b3e81a`:

- Added a call-site test that forces `detect_apple_silicon()` to `False`, fails if `query_powermetrics()` is reached, and verifies the NVIDIA sample is rendered. This pins the non-Apple -> NVIDIA route and kills the reported `if True or detect_apple_silicon()` mutation.
- Replaced the malformed oversized byte fixture with a valid plist containing a padded string. The test proves the plist is valid and larger than the cap before asserting rejection.
- Pinned the subprocess timeout at both root and non-root call sites (`interval + 5.0`).
- Added `stdin=subprocess.DEVNULL` to the powermetrics subprocess.
- Tightened option assertions to verify argument adjacency.
- Updated the changelog entry to `(#99 by @Amix29 in #481)`.

Local verification on the Apple Silicon checkout:

- `PYTHONPATH=src python3 -m pytest tests/test_issue99_apple_monitor.py -v --tb=short --cov-fail-under=0` -> **32 passed**
- `ruff check src/soup_cli/ tests/` -> **clean**
- `/usr/bin/powermetrics --help` lists `--handle-invalid-values`, so the shipped binary on this M4 Max accepts that option.
- Running the exact non-interactive command without a cached sudo ticket returns exit code 1 with the exact stderr `sudo: a password is required`. This matches the current `password is required` permission marker and therefore preserves the actionable `sudo -v` path.

I left the optional pre-capture output limiting unchanged. The current parser cap remains a useful fail-safe, while a robust streaming cap would be a separate, larger change; the command itself is fixed to the absolute Apple binary and one sample.

Hardware access note: this Mac does not belong to me. I configured remote access so Codex can work on the repository on this Mac, but I only have access through Codex, not direct or administrator access to the machine. I therefore cannot authenticate `sudo` or collect the privileged plist snapshot myself. I am waiting for the Mac owner to be available to run the command. Until then I cannot honestly confirm the live plist nesting or the positive-path blocking duration; I will post those two hardware observations as soon as the owner can provide the privileged snapshot.

MakazhanAlpamys:
**All three re-verified on `4b3e81a`, each killed by a named test.** I re-ran the same mutations rather than reading your summary:

| Mutation | Before | Now |
|---|---|---|
| `if detect_apple_silicon():` → `if True or …` | 126 passed | **killed** — `test_monitor_non_apple_routes_to_nvidia` |
| delete `timeout=interval + 5.0,` | 126 passed | **killed** — `test_query_powermetrics_root_uses_fixed_binary_without_shell` **and** `…_non_root_uses_non_interactive_sudo` (both call sites, as you said) |
| disable `if len(payload) > _MAX_POWERMETRICS_OUTPUT_BYTES:` | 126 passed | **killed** — `test_parse_powermetrics_skips_malformed_and_oversize_payloads` |

Baseline **127 passed, 1 skipped**. `stdin=subprocess.DEVNULL` is at `gpu_monitor.py:283`. CHANGELOG now reads `(#99 by @Amix29 in #481)`.

I also checked the 18 deleted test lines, because a shrinking test file is something I always look at. **Every deletion is an upgrade** — the membership checks became adjacency checks (`argv[argv.index("--samplers") + 1] == "gpu_power"`), which is strictly stronger than what I asked for in the nit. Nothing was weakened.

## Your hardware note — that is the right answer, and it does not block the merge

Saying "this Mac is not mine, I only reach it through Codex, I cannot authenticate `sudo`, so I cannot honestly confirm the plist nesting" is worth more to me than a confident guess would have been. Please do not chase the owner on my account.

Here is why it does not block: I went and **measured the failure mode** rather than leaving it as the open risk I described last round. Feeding your parser plists with plausible *alternative* nestings:

```
expected shape   (gpu.idle_ratio + processor.gpu_power) -> util 75.0%, 12.5 W
gpu_power under `gpu` instead of `processor`            -> util 75.0%, 12.5 W
gpu_energy instead of gpu_power                         -> util 75.0%, 12.5 W
no `gpu` key at all                                     -> []
```

So the parser already tolerates the two nestings I was most worried about, derives power correctly from energy over `elapsed_ns` when that is what the binary emits, and on a genuinely unexpected document returns `[]`. **The degenerate case is "reports unavailable", not "reports a plausible wrong number"** — which is the direction that matters for a telemetry feature, and it is now measured rather than assumed. I was overweighting that risk last round; this is the evidence that corrects it.

Your two positive confirmations cover the rest of what I asked: `/usr/bin/powermetrics --help` listing `--handle-invalid-values` settles whether the flag exists on a shipped binary, and `sudo: a password is required` matching your permission marker settles that the actionable `sudo -v` hint really fires instead of degrading to a generic failure. Those were the two that could have made the feature quietly useless, and you checked both against the real binary.

Post the privileged snapshot whenever the owner is free — as a comment on #99, not as a blocker here. If the live nesting turns out to be a third shape the parser misses, that is a follow-up with a one-line fix and a fixture, not a reason to hold a feature whose worst case is an empty table.

CI was sitting in `action_required` on your new push (every push from a first-time contributor needs approval, not just the first) — I have released it. Merging as soon as all 12 cells are green.

Turning three real coverage gaps around in twenty minutes, and strengthening two assertions I had only filed as nits, is a good showing. Thank you.

MakazhanAlpamys:
Heads up: I merged #479 (#296, MCP network transports) a few minutes ago, and it added an entry at the top of `[Unreleased]` — which collided with yours. Your branch went from clean to conflicting through no fault of your own.

**I resolved it for you** (`5c56469` here, `a1350c1` on #482) since the merge order caused it, not you. The resolution keeps **both** CHANGELOG entries verbatim; `docs/commands.md` and `pyproject.toml` auto-merged; **no source or test file in your PR was touched**. Verified afterwards: 78 passed across your suite plus the incoming one, ruff clean.

CI needed approving again after the merge commit — done. If you push again, just pull first so you do not have to redo it.

MakazhanAlpamys:
**Merged** as `1de7d02`, and you are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) (`463685c`). `soup monitor` has a real Apple Silicon path for the first time — it has been an NVIDIA-shaped blank on those machines since v0.44.1.

## What I actually ran, so you know the merge was earned

Isolated worktree with `PYTHONPATH` pinned to your branch, `soup_cli.__file__` confirmed inside it before trusting any result — an editable install here resolves to the main checkout, which silently makes every mutation "survive".

- **127 passed, 1 skipped** across `test_issue99_apple_monitor.py` + `test_v0440_part_a.py` + `test_v0440_review_followups.py`
- Merge-base control: your tests are uncollectable at `e61c1a0` (`ImportError: PowermetricsResult`) — an honest reproduction
- `import soup_cli.commands.monitor` → `torch` not in `sys.modules`, no import-time hardware access. Your lazy-behaviour claim held.
- CI 12/12 at `5c56469`, SHA-matched against the PR head
- **205 passed** on `main` after the merge

Adversarial probing of the parser, because it eats output from a privileged process: billion-laughs entity bomb, XXE `file:///`, 5,000-deep nesting, embedded NUL bytes — all return `[]` without raising. Fixed absolute paths mean `PATH` cannot reach the executable, and the only interpolated argv value is an `int` derived from an interval you already bound to `[0.25, 30]`.

## The three fixes, re-verified rather than read

| Mutation | Before | After |
|---|---|---|
| `if True or detect_apple_silicon()` | 126 passed | `test_monitor_non_apple_routes_to_nvidia` |
| delete `timeout=interval + 5.0` | 126 passed | two tests — both call sites |
| disable the payload size cap | 126 passed | the oversize test, now discriminating |

I also read all 18 deleted test lines, because a shrinking test file is something I always check. **Every one was an upgrade** — membership assertions became adjacency assertions. That was a nit I filed and you fixed it more thoroughly than I asked.

The first of those three mattered most: it was the single line deciding which implementation *every Linux and Windows user* gets, and nothing was holding it up.

**You said "I cannot confirm this" instead of guessing.** The Mac is not yours, you reach it through Codex, you cannot authenticate `sudo` — so you declined to assert the live plist nesting and confirmed only what you could actually check against the real binary (`--handle-invalid-values` is accepted; a ticketless `sudo -n` prints `sudo: a password is required`, matching your permission marker so the `sudo -v` hint really fires). Those two were the ones that could have made the whole feature quietly useless. A confident guess there would have been worth less than nothing, and it is the kind of thing a reviewer cannot catch from the outside.

That honesty is also what pushed me to *measure* the residual risk rather than leave it as a worry. Feeding your parser alternative nestings: it accepts `gpu_power` under either `gpu` or `processor`, derives correctly from `gpu_energy` over `elapsed_ns`, and returns `[]` on a genuinely unexpected document. Worst case is an empty table, not a plausible wrong number — the safe direction, and the reason the unconfirmed schema did not block this.

**You kept the NVIDIA path genuinely untouched.** I verified that line by line rather than taking the claim: every deleted line under `src/` is the v0.44.1 stub or a docstring.

## Still open, and not blocking

Post the privileged snapshot on #99 whenever the owner is free. If the live nesting turns out to be a third shape the parser misses, that is a one-line fix plus a fixture.

Two remaining LOWs I left as your call: the 1 MB cap is applied after `capture_output` has already buffered (defence-in-depth only — exploiting it needs root to replace the binary), and `monitor.py`'s success branch has no sleep, resting on powermetrics blocking for the sample window. You have the hardware to settle that second one; I do not.

Three PRs in a day, each on a real issue, with a twenty-minute turnaround on three findings — #480 and #482 are still with you and I have reviewed both. Thank you.

PR #482, opened 2026-08-22
feat(eval): add safe Aider Polyglot Docker runner

### LINKED ISSUE #91: v0.43.1: Aider Polyglot live runner (`soup eval aider`)
v0.43.0 ships the `aider_polyglot` name in the benchmark allowlist (`NEW_BENCHMARKS_V0_43`) and `benchmark_metadata` returns a description, but `lm_eval_task_for("aider_polyglot")` returns `None` because there is no lm-eval mapping. Live wiring requires the upstream `aider-chat` package + Docker.

**Fix path:**
1. Add `[aider]` extra: `aider-chat`.
2. New `commands/eval.py` subcommand `soup eval aider --model <path> --output <dir>` that subprocess-launches aider with the polyglot benchmark.
3. Parse aider's JSON output; emit a Soup-shaped result row (model, task, score, errors).
4. Containment via shared `is_under_cwd` on output dir.

**Acceptance:**
- [ ] `soup eval aider --help` lists the flag
- [ ] Docker check at startup; friendly error when missing
- [ ] Output JSON parseable by `soup eval compare`

Amix29:
Closes #91.

## Summary

- add `soup eval aider --model ... --output ...` around Aider's documented Polyglot Docker harness
- preflight the Docker client, daemon, and locally built `aider-benchmark` image with actionable errors
- mount the prepared Polyglot corpus read-only and keep all benchmark output under the current working directory
- execute Docker with fixed argv/no shell, bounded runtime, and provider credentials inherited by name so secret values never enter argv
- aggregate Aider's bounded per-exercise `.aider.results.json` files into `soup_result.json` with `model`, `task`, `score`, `errors`, and details
- optionally persist the `aider_polyglot` score with `--run-id`, making it available to the existing `soup eval compare RUN1 RUN2` workflow
- add the requested `[aider]` extra and document the source-only benchmark image/corpus setup

## Upstream contract correction

The `aider-chat` wheel does not contain `benchmark/benchmark.py`, the benchmark Dockerfile, or its companion modules, and the regular `aider` entry point has no Polyglot flag. The official flow is source/image based, so this PR wraps that real contract instead of invoking a non-existent `aider --polyglot` command. Likewise, Soup's current `eval compare` consumes tracker run IDs rather than JSON paths; `--run-id` is the compatibility bridge.

## Safety

- output containment uses the shared realpath/commonpath helper and rejects a symlinked target
- the exercise mount is read-only
- subprocesses use list argv, `shell=False`, and timeouts
- result files are count/size bounded, regular-file checked, opened with `O_NOFOLLOW` where available, and schema validated
- Docker execution remains mandatory because the benchmark runs untrusted model-generated code

## Validation

- `243 passed` across the new tests, eval platform tests, v0.43 benchmark tests, and the repository's ANSI-safe help assertion scanner
- `ruff check src/soup_cli/ tests/` passes
- wheel and sdist build successfully; wheel metadata contains `Provides-Extra: aider` and `aider-chat<0.87,>=0.86.2`
- on an Apple Silicon M4 Max host with Docker CLI 29.2.0 installed but Docker Desktop stopped, the real command exits cleanly with the intended daemon-unavailable message before benchmark execution

A full-suite attempt in an isolated `.[dev]` environment reached 719 passing tests before stopping on an unrelated macOS/PyTorch process-exit crash in `test_doctor_runs`: `soup doctor` printed “All checks passed” and then the subprocess exited with signal 11. A live paid Polyglot run was not performed because the local Docker daemon is not running and no benchmark provider/model credentials were supplied.


MakazhanAlpamys:
**This is careful security engineering and I could not break the container contract.** I verified both of your upstream claims rather than taking them on trust, and personally re-ran every finding below before writing it up. One blocker, and it is a test, not the code.

## Your two contract corrections both hold

I checked these because they changed the shape of the work relative to #91's literal text, and a deviation is exactly where a reviewer should be most sceptical. Both are true:

- **The `aider-chat` wheel really has no benchmark.** Inspected `aider_chat-0.86.2-py3-none-any.whl` (235 entries) through the PyPI JSON API without installing: zero paths matching `benchmark`, no benchmark Dockerfile, one console script `aider = aider.main:main`, and `args.py` (135 `add_argument` calls) contains neither `--polyglot` nor `benchmark`. Wrapping the source-built image is the real contract.
- **`eval compare` really cannot consume a JSON path.** It takes two `typer.Argument` run IDs and reads `ExperimentTracker`. #91's literal acceptance criterion is unsatisfiable as written, so `--run-id` is not a substitute for it — it is the only way to satisfy it. Driven end-to-end (two real runs, `save_eval_result`, `compare_runs`) it produces a correct `aider_polyglot` row with `delta: 0.5`. And you write `soup_result.json` as well, so both readings are served.

Doing that verification *in the issue before opening the PR* is exactly the flow CONTRIBUTING.md asks for.

## Verified good

I dumped the real argv rather than reading the code: `--rm` present, exercise mount `readonly`, **no `docker.sock`**, **no `--privileged`**, memory capped at 12g, `shell=False` with list argv everywhere, timeouts on all three subprocesses. Your credential claim is exact — with `OPENAI_API_KEY=sk-SUPERSECRET` set, no secret value appears in argv, only `--env NAME` forms. Output containment uses the shared `utils/paths.py` realpath+commonpath helper (Windows-short-name safe) in three layers including a post-`mkdir` race recheck, and `score` is computed from counts rather than read from JSON so a hostile number cannot reach it. The diff is purely additive (`+908/-0`) with no test touched anywhere else. **16 passed** on the new file, **220 passed** across `test_eval.py` + `test_eval_platform.py` + `test_v0430_part_b.py`, ruff clean, and the merge-base control fails 16/16 on ImportError.

Mutations that died by name: containment (`test_output_escape_is_rejected_before_docker`), `readonly=True` (`test_build_command_uses_fixed_argv_and_read_only_corpus`), the run timeout and `shell=False` (`test_success_writes_soup_json_and_tracks_run`), and all three preflights (`test_missing_docker_has_friendly_error`, `test_daemon_failure_has_friendly_error`, `test_missing_image_has_setup_hint`).

## Blocking: the symlink guard your Safety section advertises has zero coverage

**1. HIGH — `tests/test_issue91_aider_eval.py:97`, `test_rejects_symlinked_result` cannot fail.** I replaced **both** `raise AiderEvalError(f"Refusing symlinked Aider result: {path}")` statements with `pass` and got **16 passed**.

The mechanism, proven rather than inferred:

```
MESSAGE : Aider result escaped --output: ...\test_rejects_symlinked_result0_mulgoq_u\python\exercises\...
contains 'Refusing symlinked'    -> False
regex match('symlink') satisfied -> True
```

`pytest.raises(..., match="symlink")` is a **search over the whole message**, and the message embeds the path — which contains the tmp_path directory named after the test function, `test_rejects_symlinked_result0`. So when the symlink guard is gone, a *different* guard fires and the assertion is still satisfied. This is the same failure mode as asserting only on an exit code: it passes when somebody else's guard trips.

Two changes: `match="Refusing symlinked"`, and add a case where the symlink target is **inside** `root` — that is the only case the `is_under` containment cannot catch, so it is the one that isolates this guard.

## Three more I would like in the same push

**2. MEDIUM — non-finite JSON escapes your error contract.** `json.loads` accepts `NaN`/`Infinity` by default, and these files are written into a **read-write** mount by container code running untrusted model output. With a real field name:

```
{"tests_outcomes":[true],"num_error_outputs":NaN}  ->  UNCAUGHT ValueError: cannot convert float NaN to integer
Infinity / 1e400                                   ->  UNCAUGHT OverflowError
```

`_count` (`aider_polyglot.py:271-280`) checks `int(value) != value`, and `int(NaN)` raises before that comparison is reached. `AiderEvalError` subclasses `ValueError`, but a bare `ValueError` is not an `AiderEvalError`, so `except AiderEvalError` in the command does not catch it and the user gets a traceback instead of the actionable message this module exists to give. A `math.isfinite` check closes it; the path is currently untested, which is why it survived mutation.

**3. MEDIUM — `--image` is not gated against a leading dash.** `--image=--privileged` is accepted and lands exactly in the image slot:

```
... "--mount", "type=bind,source=.,target=/results",
    "--privileged",          <-- the image argument
    "python3", "/aider/benchmark/benchmark.py", ...
```

Docker's parser consumes that as an option and promotes `python3` to the image name. Today it is self-inflicted — the operator types it — so I am not calling it critical. But it becomes a genuine escape vector (`--privileged`, `-v /:/host`) the first time `--image` comes from a recipe, a `soup.yaml`, or a CI variable, and it is inconsistent with the care you took elsewhere: you reject NUL in `--model` and commas in `--mount`. A Docker-reference charset regex closes it.

**4. MEDIUM — `validate_exercises_dir` has no containment or symlink check.** `validate_exercises_dir("..")` returns a directory outside cwd, and it gets bind-mounted into a container running untrusted model-generated code with unrestricted network egress. Read-only and operator-specified, so not urgent — but starkly asymmetric with the three-layer hardening you gave `--output`. A warning when the corpus resolves outside cwd would be enough.

**5. LOW — four guards with no test at all**: `MAX_RESULT_FILES`, `MAX_RESULT_TOTAL_BYTES`, the parse-loop `is_under` containment, and `_mount`'s comma rejection — that last being the `--mount` option-injection guard. Neutralising each leaves the suite green. One small test each.

## A finding I am withdrawing, and why

My review pass flagged the raw `result.output` assertions at `tests/…:216-217, 307, 339` as a CI-red risk, on the strength of a `FORCE_COLOR=1` run that failed 2 of 16 (Rich number-highlights the interpolated integer, so `timed out after 60 seconds` arrives with escapes inside it).

**Real CI disagrees, so the finding does not stand as written: all six ubuntu and macOS cells passed.** Those are the colour-capable runners, and `FORCE_COLOR=1` is stricter than what GitHub actually gives us — a caveat our own CLAUDE.md records. I would rather correct myself here than have you chase a red that is not there. It stays worth a cheap fix, because those three assertions are the only ones in the file that behave differently under colour (92 neighbouring tests are unaffected) and the file already defines `_plain()` and uses it once. Latent fragility, not a blocker.

## Nits

**6.** `CHANGELOG.md:18` reads `(#91)`. House format is `(#91 by @Amix29 in #482)` — the entry directly below yours does it correctly. Please add it yourself rather than leaving it for release time.

**7.** `--add-host=host.docker.internal:host-gateway` is unconditional and undocumented. It gives untrusted container code reach to host loopback services — a local Ollama or vLLM, Soup's own web UI, a database. It is needed only for local OpenAI-compatible endpoints, so it should be opt-in, and your otherwise-excellent security paragraph in `docs/evaluation.md` should say it exists.

**8.** The `[aider]` extra installs a package the code never imports — the runner shells out to Docker. So `pip install "soup-cli[aider]"` does not by itself yield a working `soup eval aider`; you still need the built image and the corpus checkout. Worth a docs sentence. Correctly kept out of `[dev]`, so it adds no CI weight — thank you for that.

## What I could not verify

**Docker is not installed on this box, so no container ever ran.** Everything about runtime behaviour is verified as argv construction and against upstream's documented contract, not by execution: that the image really contains `/aider/benchmark/benchmark.py`, that the daemon enforces `readonly`, that `--add-host` resolves, and that real `.aider.results.json` files match the assumed `*/exercises/practice/*/` layout and field names. If you can get one real run — even a single `--num-tests 1` against a cheap model — the layout assumption is the one worth confirming, because a mismatch would make the parser quietly return zeros rather than fail.

Fix **1** and I will merge; 2, 3 and 4 in the same push if you can. CI is green on 9 of 12 with the Windows cells still running.

Three PRs in a day, each landing on a real issue with the upstream contract checked first — that is a lot of value in a short time. Thank you.

MakazhanAlpamys:
Heads up: I merged #479 (#296, MCP network transports) a few minutes ago, and it added an entry at the top of `[Unreleased]` — which collided with yours. Your branch went from clean to conflicting through no fault of your own.

**I resolved it for you** (`a1350c1` here, `5c56469` on #481) since the merge order caused it, not you. The resolution keeps **both** CHANGELOG entries verbatim; `docs/commands.md` and `pyproject.toml` auto-merged; **no source or test file in your PR was touched**. Verified afterwards: 62 passed across your suite plus the incoming one, ruff clean.

CI needed approving again after the merge commit — done. If you push again, just pull first so you do not have to redo it.

MakazhanAlpamys:
Small heads-up: the CHANGELOG conflict is back. #481 merged just now and added its `[Unreleased]` entry in the same place, so my earlier resolution was undone by the next merge rather than by anything you did.

I am **not** resolving it a second time, because you have a push coming for the symlink test anyway and I would only be re-conflicting you again. When you push the fix, `git pull origin main` (or merge `main`) first and keep both entries — that is all it is.

Structural, not your fault: every PR inserts at the top of `[Unreleased] ### Added`, so each merge conflicts every open PR. Worth me fixing at the convention level rather than making contributors pay for it repeatedly.

CI is 10/12 green on the pre-conflict head, so nothing else is outstanding beyond the review findings.

Amix29:
Thanks for the detailed review. I pushed `66a2c6a` with current `main` merged and both CHANGELOG entries retained.

This push addresses the blocking finding and the requested hardening:

- tightens the symlink assertion to `Refusing symlinked` and adds the isolating case where the target remains inside the result root;
- rejects `NaN` and infinite result counts as `AiderEvalError`;
- validates Docker image references before lookup or argv construction, including leading-option rejection;
- warns when `--exercises-dir` resolves outside cwd while keeping the mount read-only;
- covers the result-file count cap, aggregate size cap, parse-loop containment, and comma mount guard;
- makes host-loopback access opt-in through `--allow-host-services` and documents the risk;
- clarifies that the `[aider]` extra alone does not provide the benchmark harness;
- updates the CHANGELOG attribution format and makes the color-sensitive assertions use normalized output.

Local validation:
- `27 passed` in the focused issue suite
- `279 passed` across the issue suite, eval regressions, and incoming Apple-monitor tests

The Docker client is present here but its daemon is unavailable, so I am not claiming a live container run. GitHub now reports the PR as mergeable, and the refreshed CI matrix is running.

MakazhanAlpamys:
**Merged** as `043314e`, and your second [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) bullet is in (`172a885`). `soup eval aider` is live — `aider_polyglot` has been a name in the allowlist since v0.43.0 with nothing behind it.

## The blocker is genuinely closed — I re-ran the exact mutation

```
$ replace BOTH `raise AiderEvalError(f"Refusing symlinked Aider result: {path}")` with `pass`
FAILED test_rejects_symlinked_result
FAILED test_rejects_symlink_even_when_target_stays_inside_root
2 failed, 25 passed
```

Before: **16 passed**. And `test_rejects_symlink_even_when_target_stays_inside_root` is the isolating case — a symlink whose target stays *inside* the result root, which is the one situation `is_under` containment cannot catch, so only the symlink guard can be what fails. That is more precise than I asked for.

Baseline is now **27 passed** (was 16). Nine of ten mutations die by name, including all four caps that had nothing on them before (`MAX_RESULT_FILES`, `MAX_RESULT_TOTAL_BYTES`, the parse-loop containment, `_mount`'s comma rejection).

**You validated the image check in both directions**, which is the part I want to single out. A validator that rejects `ghcr.io/org/img:tag` would be worse than the bug it fixes. Verified live: rejects `--privileged`, `-v`, `--rm`, `-e`, `--mount`, empty/whitespace/NUL strings; accepts `aider-benchmark`, `ghcr.io/org/img:tag`, `img@sha256:<64 hex>`, `registry.example:5000/aider-benchmark:test`, `paulgauthier/aider-benchmarks`. Most people fix the reported input and stop.

Non-finite counts now raise `AiderEvalError` rather than a bare `ValueError` escaping the command's handler, and an exercises directory outside cwd warns. All three killed by named tests.

## Two LOW follow-ups I did not block on

**1.** The `_validate_image_name(image)` call *inside* `build_docker_command` (`aider_polyglot.py:203`) has no test of its own — deleting only that line, keeping the preflight one, leaves 27 passing. The defect is fixed; it is that separately-importable entry point that is unpinned, so a future refactor could drop it silently. One test mirroring `test_image_cannot_be_parsed_as_a_docker_option` against `build_docker_command` closes it.

**2.** The "stays_read_only" half of `test_exercises_outside_cwd_warns_and_stays_read_only` is vacuous: its assert lives inside the monkeypatched `fake_run`, and `CliRunner.invoke` without `catch_exceptions=False` swallows that `AssertionError` into `result.exception` while still returning `exit_code == 1`, which the test accepts. Proof: flipping `readonly=True` → `readonly=False` fails `test_build_command_uses_fixed_argv_and_read_only_corpus` but **not** this one. The property is covered by that other test, so this is a naming problem rather than a hole — but it is the same shape as the symlink bug you just fixed, and worth knowing the pattern.

Neither is worth a PR on its own; fold them into whatever you touch next in that file.

## What went well here, concretely

You verified the upstream contract **before writing code** and posted it in the issue, which is what let me accept a deviation from #91's literal text instead of treating it as scope drift. I checked both claims rather than taking them: the wheel really has no benchmark (235 entries, zero matching `benchmark`, 135 `add_argument` calls with no `--polyglot`), and `eval compare` really cannot consume a JSON path, so `--run-id` was the only route to that criterion, not a weaker substitute for it.

Not verified by me, and worth saying plainly: **Docker is not installed on my box, so no container ever ran.** The image contents, daemon enforcement of `readonly`, `--add-host` resolution and the real `.aider.results.json` layout are checked against upstream's documented contract and as argv construction only. If you ever get a single `--num-tests 1` run against a cheap model, the results-file layout is the assumption worth confirming — a mismatch would make the parser return zeros rather than fail.

Also: apologies for the CHANGELOG conflict churn. That was my merge order, twice. I filed #487 to fix it at the convention level so contributors stop paying for it.

Three merged PRs today. Thank you.

Issue #359, opened 2026-08-09
DeepSpeed + LoRA is repaired in sft.py only: the other trainer wrappers still hit the empty no-decay group


MakazhanAlpamys:
#336 fixed the empty-parameter-group failure (with LoRA, HF's no-decay optimizer group comes out empty, DeepSpeed drops it, and torch's scheduler then hits a 1-vs-2 strict-`zip` mismatch on `base_lrs`) — but only in `trainer/sft.py`. Every other trainer wrapper builds its own optimizer/scheduler and still fails the same way under `--deepspeed` with LoRA.

Recorded as a Known Limitation in the v0.73.0 release notes rather than left implicit.

## Fix path

`prune_empty_param_groups` already exists in `utils/deepspeed.py` and does the right thing (drops empty groups in place before `create_scheduler` runs; refuses when *every* group is empty so "nothing is trainable" stays a clear error). The work is threading it through the remaining wrappers the way `sft.py` does, and deciding where it belongs so a nineteenth wrapper cannot forget it — the `device_map` fix in the same release moved from a hand-written trainer list to a scanner over `soup_cli/trainer/` for exactly this reason.

Two related gaps from the same investigation, in scope or trackable separately:

- a user-supplied `--deepspeed my.json` is passed through unresolved, so none of the preset rewrites (bf16 vs fp16 quantised weights, `zero_hpz_partition_size` derived from `WORLD_SIZE`) apply to it;
- ZeRO++ hierarchical partitioning is unit-tested but has never been exercised across two nodes.

## Acceptance criteria

- [ ] Every trainer wrapper that builds an optimizer goes through the shared guard, enforced by a test that **scans** `soup_cli/trainer/` rather than listing names
- [ ] A control proving the scanner can fail
- [ ] At least one non-SFT wrapper verified end-to-end on real multi-GPU hardware (`benchmarks/gate-h100-validation.md` STEP 20 is the template: LoRA + DeepSpeed arms red before, green after, with the full-fine-tuning control untouched)
- [ ] Decide and document whether `--deepspeed <file>` should be resolved or deliberately left as an escape hatch

MakazhanAlpamys:
Taking this one, with one criterion I cannot close myself — stating that up front per the claim rule.

[...]

Issue #336, opened 2026-08-07
DeepSpeed fails with LoRA on every stage: empty no-decay param group vs strict zip


MakazhanAlpamys:
Measured on 4xH100 (#77). **Every** DeepSpeed stage fails when the model uses LoRA.

```
File ".../deepspeed/runtime/lr_schedules.py", line 296
ValueError: zip() argument 2 is longer than argument 1
```

### Mechanism, measured
HF builds **2** optimizer param groups (decay / no-decay). With LoRA the no-decay group is **empty** — on 1 GPU the group sizes are `[192, 0]` with `base_lrs` of length 2. DeepSpeed drops the empty group, leaving **1 group against 2 base_lrs**, and torch 2.13's `zip(..., strict=True)` raises.

### The control that isolates the trigger
Same `--deepspeed zero2` config with **LoRA replaced by full fine-tuning** gives group sizes `[1, 1]` and trains to completion (3.1811 → 0.0, 6667.3 tok/s). So the trigger is DeepSpeed **+ LoRA**, not DeepSpeed. This also explains why an earlier ZeRO-3 run passed — it was full-FT; with LoRA, ZeRO-3 fails identically.

Confirmed on `zero2`, `zero3` and `zero++`.

### Separately, zero++ fails EARLIER and for a different reason
```
RuntimeError: expected mat1 and mat2 to have the same dtype, but got: c10::Half != c10::BFloat16
  at deepspeed/runtime/zero/linear.py:92
```
The preset sets `zero_quantized_weights`/`zero_quantized_gradients: true` (fp16) against a bf16 model. **Fixing the param-group bug alone will not make zero++ run.** That preset also hardcodes `zero_hpz_partition_size: 8` while the run had 4 GPUs.

### Fix shape
Drop empty param groups before handing the optimizer to DeepSpeed, or build a single group when the no-decay set is empty. For zero++, either derive the quantisation dtype from the model or refuse the combination with a message.

### Acceptance
- [ ] `--deepspeed zero2` and `zero3` train to completion with LoRA
- [ ] `zero++` either runs on a bf16 model or refuses with a message naming the dtype conflict
- [ ] `zero_hpz_partition_size` follows the actual GPU count
- [ ] a test on the param-group construction with LoRA, plus the full-FT control

Record: `benchmarks/gate-h100-validation.md` STEP 20.

MakazhanAlpamys:
[...]

PR #484, opened 2026-08-22
fix(trainer): attach the DeepSpeed empty-group guard in every wrapper

MakazhanAlpamys:
## What does this PR do?

Addresses #359 — closes criteria 1, 2 and 4. **Criterion 3 is deliberately left unticked**; see the bottom.

#336 fixed the DeepSpeed + LoRA failure in `trainer/sft.py`. Measured before changing anything: **19 modules under `soup_cli/trainer/` accept a `deepspeed_config`, and exactly one called the guard.** So 18 tasks still died the same way — with LoRA every trainable tensor is 2-D, HF's no-decay group comes out empty, DeepSpeed drops it, the LR scheduler keeps two `base_lrs`, and torch's strict `zip` raises at the first `lr_scheduler.step()`.

## Coverage is scanned, not listed (criterion 1)

`tests/test_issue359_deepspeed_guard_coverage.py` walks `soup_cli/trainer/*.py` and requires the guard where a module **both** accepts a `deepspeed_config` **and** constructs a trainer itself:

```python
_BUILDS_TRAINER = re.compile(r"self\.trainer\s*=\s*(?!None\b)[A-Za-z_][\w.]*\s*\(")
```

That second condition is what makes the scan honest rather than merely strict. `preference.py` accepts a `deepspeed_config` and forwards it to the DPO / SimPO / ORPO / IPO / BCO wrappers without building anything — the guard belongs in those, not in it — so it is exempt by rule, not by name. `test_the_scan_finds_more_than_the_one_wrapper_336_fixed` pins that the detector reports ≥15 modules, so a broken detector fails instead of quietly exempting everything.

This mirrors `tests/test_device_map_distributed.py::TestNoTrainerStillHardcodesAuto`, and that file's history is the whole argument for the shape: its first version parametrized over the six trainer names the original fix had touched and passed while **nine** more sites still carried the defect.

## The scanner has a control (criterion 2)

A scanner nobody has watched fail is indistinguishable from a broken one:

- `test_the_patterns_would_catch_the_unguarded_shape` — both halves of the detector match the unguarded spelling and reject the guarded one, including `self.trainer = None`
- `test_a_comment_mentioning_the_guard_does_not_satisfy_it` — the note explaining the guard must not be enough to pass without calling it

## A defect the widening exposed

`attach_empty_param_group_guard` did `original = trainer.create_optimizer` with no check. Harmless while it lived in one wrapper; not harmless attached from eighteen, because not every TRL trainer exposes that method and the `AttributeError` would fire **while attaching** — converting a DeepSpeed-only defect into a crash on the ordinary path. It now declines (returns `False`) and leaves such a trainer untouched, which is correct on the merits too: a trainer that never builds the two groups never had the problem.

Covered by `TestGuardToleratesTrainersWithoutCreateOptimizer`, including the positive control that a real `create_optimizer` is still wrapped and still prunes.

## `--deepspeed my.json` (criterion 4) — decided: resolve, narrowly

The claim said I would read the rewrites before deciding. Having read them: **resolve, but only keys that are provably invalid for the run.**

The two keys `resolve_deepspeed_config` touches are not preferences, they are errors:

| key | what happens without the rewrite |
|---|---|
| `zero_hpz_partition_size` | DeepSpeed refuses a value the world size is not divisible by — the ZeRO++ preset ships `8`, invalid on any box that is not a multiple of 8 |
| `zero_quantized_weights` / `zero_quantized_gradients` | the fp16 quantiser against the `bf16` the same file enables makes the dequantised all-gather return `c10::Half` into a `c10::BFloat16` activation: `expected mat1 and mat2 to have the same dtype` |

The deciding fact is that **the documented way to customise ZeRO++ is to copy the preset JSON** — which copies both defects. Leaving a user file unresolved hands the escape hatch a crash the presets are already protected from, and the failure is a slow one to diagnose.

So it is narrow rather than total:

- a config using none of those keys is returned **by its own path, byte-identical** — the escape hatch stays an escape hatch
- a repair is **printed**, never silent, naming the file
- the repaired config goes to a temp copy; **the user's file on disk is never modified**
- malformed JSON passes straight through, because DeepSpeed reports a bad config better than a Soup traceback, and refusing here would reject files DeepSpeed accepts

Wiring is pinned separately from behaviour (`test_the_cli_routes_a_user_file_through_the_resolver`) — without that, the resolver could be correct and never run on the path that matters — plus a control that a named preset still takes the preset path.

## Criterion 3 is NOT closed, on purpose

> At least one non-SFT wrapper verified end-to-end on real multi-GPU hardware

I have no multi-GPU hardware, said so in the claim before starting, and am not ticking it. The defect appears at `lr_scheduler.step()` under a real DeepSpeed engine; **no CPU test in this PR can observe it.** What is proven here is that every affected wrapper now calls the guard and that the guard behaves correctly in isolation — which is coverage, not validation. Closing #359 on this PR alone would record a green unit suite as evidence for something it structurally cannot see.

The `benchmarks/gate-h100-validation.md` STEP 20 equivalent — LoRA + DeepSpeed arms red before and green after on a non-SFT wrapper, full-fine-tuning control untouched — still needs someone with the box. The two-node ZeRO++ gap named in the issue is untouched for the same reason.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [x] Updated relevant docs — `docs/performance-and-quantization.md` gains a `--deepspeed <file>` section recording the decision and a DeepSpeed + LoRA section; `CHANGELOG.md` has both entries
- [x] No breaking changes — the guard is a no-op without `--deepspeed` and for full fine-tuning, and a user JSON that needs no repair is returned unchanged by its own path

Refs #359 — deliberately NOT `Closes`. Criteria 1, 2 and 4 are met; criterion 3
(a multi-GPU DeepSpeed red-before / green-after) cannot be demonstrated on the
review hardware, so #359 stays open after this merges.


MakazhanAlpamys:
Rebased onto `main`. #490 landed the `changelog.d/` mechanism while this was open, so the `CHANGELOG.md` edit is replaced by `changelog.d/0.73.3/484.fixed.md` — same two entries, verbatim. Only `CHANGELOG.md` conflicted; the guard in all eighteen wrappers and `test_issue359_deepspeed_guard_coverage.py` are untouched, and the affected suites still pass locally (107 tests across the #359, #336 and #487 files).

PR #467, opened 2026-08-20
docs(contributing): eliminate hand-maintained test suite statistics via shape invariant (#465)

harshitthek:
Resolves #465.

### Background & Motivation
In #465, @MakazhanAlpamys identified that hand-maintained test suite statistics in `CONTRIBUTING.md` (`370 files, 18151 tests`) create an inevitable drift treadmill, as more than 50% of recent merges touch `tests/`. Adding a value-based ratchet would redden CI for routine, correct test additions on unrelated documentation lines.

Following maintainer alignment on **Option A (Subtraction & Shape Invariant)**, this PR eliminates the hardcoded numbers at the documentation source and adds a permanent anti-regression shape guard.

---

### Key Changes
1. **Subtraction at Source (`CONTRIBUTING.md`):**
   - **Line 126:** Replaced `(370 files, 18151 tests)` with an informative description (`tests/ - Test suite covering all CLI commands, trainers, and model architectures`).
   - **Line 156:** Replaced `### Test Files (370 files)` with `### Test Files`.

2. **Permanent Anti-Regression Shape Test (`tests/test_contributing_test_counts_are_digit_free.py`):**
   - Implements pure `audit_contributing_test_count_shape()` and `ContributingShapeVerdict` dataclass.
   - Asserts that both the `tests/` line and `### Test Files` section header exist (anti-rot), while strictly asserting that neither line carries hardcoded digits (`\d`).
   - Isolates strictly those two lines so other valid numbers in `CONTRIBUTING.md` (e.g. `(146 recipes)`, `Python 3.11`, `v0.33.0`) never trigger false alarms.

3. **Mutation-Resistant Control Battery (`TestTheGuardHasTeeth`):**
   - Verifies that re-introducing digits in line 126, line 156, or both fails deterministically and reports the exact line number.
   - Verifies that missing either target section fails the anti-rot check.

4. **Changelog:**
   - Added attributed entry under `[Unreleased]` $\to$ `### Changed`.

---

### Verification
- `pytest tests/test_contributing_test_counts_are_digit_free.py tests/test_recipe_count_is_synced.py -v` $\to$ **19 passed in 0.98s**.
- `ruff check src/ tests/ docs/ CHANGELOG.md CONTRIBUTING.md` $\to$ **All checks passed (0 warnings)**.
- Pure POSIX UTF-8 without BOM.

Issue #465, opened 2026-08-20
[Ratchet] Test suite count and test file count synchronization in CONTRIBUTING.md


harshitthek:
### Context & Problem
During the review of PR #457, @MakazhanAlpamys noted that the hand-maintained test suite statistics in `CONTRIBUTING.md` have drifted out of sync with the actual repository:
* **Line 126:** `tests/ - Test suite (370 files, 18151 tests)`
* **Line 156:** `### Test Files (370 files)`

Following the merges of #458 and #460, the actual repository state has grown to **372 test files and ~18,202 tests**. Maintaining these numbers manually by hand creates a continuous drift treadmill, off-by-one guesswork (e.g. bumping `+6` for a PR that added 7 tests), and merge conflicts on `CONTRIBUTING.md` across parallel PRs.

---

### Proposed Design Approaches

We would like to introduce an automated ratchet (`tests/test_test_count_is_synced.py`) to permanently prevent this drift. Here are two detailed architectural approaches for maintainer review:

---

#### 🔹 Option 1: Pragmatic File-Count Ratchet + Milestone Item Metric (Recommended)

* **Architecture & Mechanism:**
  * Uses deterministic filesystem scanning (`pathlib.Path("tests").rglob("test_*.py")`) to compute `len(test_files)`.
  * Strictly asserts exact single-integer equality against `CONTRIBUTING.md:156` (`### Test Files (372 files)`).
  * Updates `CONTRIBUTING.md:126` to a milestone format: `tests/ - Test suite (372 files, 18,000+ tests)`.
* **Key Advantages:**
  * **Zero Churn:** Eliminates the merge-conflict treadmill for every routine PR that adds 1–2 test functions to an existing file.
  * **100% Deterministic & Instant (<2ms):** Requires zero AST parsing or runtime subprocess overhead.
  * **Guaranteed File Coverage:** Strictly forces any contributor who adds or deletes a `test_*.py` file to update `CONTRIBUTING.md`.
* **Trade-off:**
  * Test function counts are tracked as milestone approximations (`18,000+`) rather than single-digit exact counts.

---

#### 🔹 Option 2: Strict Single-Digit AST Static-Analysis Ratchet

* **Architecture & Mechanism:**
  * Uses Python's built-in `ast` module (`ast.parse`) to perform pure offline static analysis across all `tests/**/test_*.py` files.
[...]

PR #485, opened 2026-08-22
fix(eval): stamp baseline scorer/version provenance (#404)

### LINKED ISSUE #404: A --baseline file carries no scorer stamp, so a pre-v0.73.2 snapshot is diffed against a new scorer on a different scale, silently

AchuthReddy-16:
## What does this PR do?

Fixes #404.

Baseline artifacts now carry `soup_version` and `scorer_revision` provenance through shared stamp/write helpers.

`resolve_baseline` warns once for unknown provenance or a scorer-revision mismatch and remains silent when the revision matches. A locked bundled-scorer fingerprint requires a revision bump whenever scorer behavior changes.

This also removes the hard-coded `SCORER_CHANGED_IN_V0_73_2` mitigation.

## Type of change

* [x] Bug fix
* [ ] New feature
* [x] Refactor / cleanup
* [x] Documentation
* [x] Tests

## Checklist

* [x] `ruff check src/soup_cli/ tests/` passes
* [ ] `pytest tests/ -v` passes
* [x] Updated relevant docs (`docs/evaluation.md`)

## Verification

* [x] 184 targeted tests passed across `tests/test_issue404_baseline_scorer_stamp.py` and `tests/test_v07302.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,16 @@ reproducing 70+ versions of notes.
   Anthropic is refused by name because its Messages API has no raw-completion
   endpoint, and local-only flags cannot be silently ignored in provider mode.
 
+- **Baseline artifacts carry a scorer/version provenance stamp (#404 by @AchuthReddy-16 in #485).**
+  Shared helpers `stamp_baseline_scores` / `write_baseline_file` write
+  `{"scores": {...}, "provenance": {"soup_version", "scorer_revision"}}`.
+  `soup eval gate --write-baseline <path>` is the user-facing producer.
+  `resolve_baseline` warns once on unknown provenance (unstamped files) or a
+  `scorer_revision` mismatch, and stays silent when the stamp matches.
+  `BUNDLED_SCORER_REVISION` + a locked fingerprint fail the suite if a bundled
+  scorer's output moves without a revision bump. Registry `save_eval_result`
+  stamps `details_json`; `soup ship --emit-evidence` includes the same stamp.
+
 - **Layer streaming now accepts Qwen3.5 MoE text checkpoints whose decoder
   layers do not expose exactly the same weight keys in every block (by
   @Shutaru in #426).** The sharder now records per-layer shard headers instead of
@@ -170,6 +180,10 @@ reproducing 70+ versions of notes.
   time (#273 by @AmirF194 in #470).
 
 ### Changed
+
+- **Remove the name-based `SCORER_CHANGED_IN_V0_73_2` baseline warning in favour
+  of the #404 scorer_revision stamp (#404 by @AchuthReddy-16 in #485).**
+  Stale baselines are detected by provenance, not by a hard-coded suite list.
 
 - **Remove hand-maintained test suite statistics from `CONTRIBUTING.md` in favor of a permanent digit-free shape invariant (#465 by @harshitthek in #467).**
   Eliminates drift across routine test additions by making test-count divergence impossible at the documentation source.
--- a/docs/evaluation.md
+++ b/docs/evaluation.md
@@ -490,19 +490,24 @@ soup ship --evidence ship_evidence.json --config soup.yaml --push owner/repo#42
 `soup ci init --config soup.yaml` binds the generated workflow's ship step to the committed
 config, so the whole loop runs in CI (see [commands.md](commands.md)).
 
-**Baseline scale change (v0.73.2).** Three suites' scorers changed. The MCQ extractor now reads a
-`\boxed{C}` option letter (a boxed *value* like `\boxed{4}` is still not an option letter), and MCQ
-prompts now ask for the letter — both halves are needed, and they affect `mini_mmlu` and
-`mini_common_sense`. `mini_tool_call` now tolerates a call missing its outer `{"function": ...}`
-envelope. Measured on an **unchanged** model the shifts are large: `mini_mmlu` 0.423 → 0.731 and
-`mini_tool_call` 0.225 → 1.000, far bigger than the 0.05 gate.
+**Baseline provenance (#404).** Produce a stamped baseline with
+`soup eval gate --suite <suite.yaml> --model <id> --write-baseline baseline.json`
+(`--model` is required — baselines are never written from the stub generator).
+The file is
+`{"scores": {...}, "provenance": {"soup_version", "scorer_revision"}}`
+and is consumed later by `soup ship --baseline baseline.json` or
+`soup eval gate --baseline baseline.json`. Shared helpers
+`stamp_baseline_scores` / `write_baseline_file` are the only writers;
+`write_baseline_file` refuses an empty score map. `resolve_baseline`
+warns once on unknown provenance (unstamped files /
+registry rows) or a `scorer_revision` mismatch, and stays silent when the
+stamp matches. The old name-based `SCORER_CHANGED_IN_V0_73_2` warning is
+gone. Recompute an old baseline, or drop names from it to force a live
+base run.
 
-A `--baseline` supplies the base score from a *file* and skips the live base run, so a snapshot
-taken before v0.73.2 would be diffed against a freshly-scored tuned model on a different scale —
-big enough to mask a real regression or manufacture an improvement. `soup ship` now warns by name
-when this happens. Recompute the baseline, or drop those names from it to force a live base run.
-`mini_instruction` and `mini_arithmetic` are unaffected: neither carries a single-letter answer, so
-the prompt cue and the option-letter extractor never touch them.
+Historical note: v0.73.2 changed three suite scorers (`mini_mmlu`,
+`mini_common_sense`, `mini_tool_call`) with measured jumps on an unchanged
+model far larger than the 0.05 gate. That is why unstamped files still warn.
 
 
 ## NLG Evaluation Metrics (BLEU + ROUGE)
--- a/src/soup_cli/commands/eval.py
+++ b/src/soup_cli/commands/eval.py
@@ -1159,7 +1159,7 @@ def gate_cmd(
     ),
     baseline: Optional[str] = typer.Option(
         None, "--baseline", "-b",
-        help="Baseline: registry://<id> or path to {name: score} JSON file",
+        help="Baseline: registry://<id> or path to stamped/flat score JSON",
     ),
     regression_threshold: float = typer.Option(
         0.05, "--regression-threshold",
@@ -1169,6 +1169,14 @@ def gate_cmd(
         None, "--model", "-m",
         help="Model path or HF id to evaluate (required for live scoring)",
     ),
+    write_baseline: Optional[str] = typer.Option(
+        None,
+        "--write-baseline",
+        help=(
+            "Write this run's task scores as a stamped baseline JSON "
+            "({'scores', 'provenance'}) consumable by --baseline (#404)."
+        ),
+    ),
 ) -> None:
     """Run an eval-gate suite standalone (post-hoc verdict)."""
     if not 0.0 <= regression_threshold <= 1.0:
@@ -1177,7 +1185,16 @@ def gate_cmd(
         )
         raise typer.Exit(1)
 
-    from soup_cli.eval.gate import load_suite, resolve_baseline, run_gate
+    if write_baseline and not model:
+        console.print("[red]--write-baseline requires --model[/]")
+        raise typer.Exit(1)
+
+    from soup_cli.eval.gate import (
+        load_suite,
+        resolve_baseline,
+        run_gate,
+        write_baseline_file,
+    )
 
     try:
         eval_suite = load_suite(suite)
@@ -1186,13 +1203,17 @@ def gate_cmd(
         raise typer.Exit(1) from exc
 
     try:
-        baseline_scores = resolve_baseline(baseline)
+        baseline_scores = resolve_baseline(
+            baseline,
+            warn=lambda msg: console.print(f"[yellow]Warning:[/] {msg}"),
+        )
     except (FileNotFoundError, ValueError) as exc:
         console.print(f"[red]Cannot resolve baseline:[/] {exc}")
         raise typer.Exit(1) from exc
 
     # When --model is provided, build a transformers-backed generator.
     # Otherwise fall back to an empty-string stub for smoke runs.
+    # (--write-baseline already refused the stub path above.)
     if model is None:
         console.print(
             "[yellow]No --model given; using stub generator "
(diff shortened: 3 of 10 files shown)


MakazhanAlpamys:
**The stamp mechanism is the right design and the registry half genuinely works.** I ran the mutations myself and then re-verified the two blocking findings by hand. Both are cheap to fix, and one of them is the PR's own headline claim.

## What I ran

Isolated worktree with `PYTHONPATH` pinned to the branch (an editable install here resolves to the main checkout, which silently makes every mutation "survive" — `soup_cli.__file__` confirmed inside the worktree first).

- `test_issue404_baseline_scorer_stamp.py` + `test_v07302.py`: **184 passed**
- Broad related sweep (`test_eval_gate`, `test_tracker`, `test_issue316_suites`, `test_v07125/28/31/35/38/39`, `test_v0715`, `test_v07302`, plus `test_cli_startup_is_light.py`): **896 passed, 2 skipped**
- Pre-fix control at merge-base: 6/6 of the new tests fail. Honest reproduction.
- `ruff` clean. Nothing anywhere still imports `SCORER_CHANGED_IN_V0_73_2` — no ImportError, and the new lazy `tracker → eval.gate → gate_suites` chain does not break CLI startup lightness.

**Backward compatibility holds**, which I checked because it is the thing most likely to bite silently: an unstamped flat file resolves and warns; an unstamped `registry://` resolves and warns; an empty file is silent; and the warning is genuinely **once per resolve**, not once per suite (a two-suite unstamped file produced exactly one). The `--evidence` + `--config` provenance gate is undisturbed. Evidence round-trips through `--emit-evidence` and the MCP `ship_evidence` tool agrees with the CLI — no drift.

Mutations that died by name: the revision comparison (`test_mismatched_revision_warns`), the stamp writer (4 tests), `tracker.save_eval_result` (2 tests), and the legacy-file warning (`test_unstamped_baseline_warns_unknown_provenance`).

**On the deleted test class** — I checked this first, because a shrinking test file is where I always look. Deleting `SCORER_CHANGED_IN_V0_73_2` **is** sanctioned: #404 says so explicitly ("can then be deleted rather than joined by a sibling constant per release"). The `-68` is confined to two hunks in `test_v07302.py:1058-1153`, and **no assertion outside that class was weakened** — the only other edit is a deleted docstring with the body unchanged. That part is clean.

## Blocking 1 — the fingerprint lock does not detect the scorer change that motivated this issue

The PR says: *"A locked bundled-scorer fingerprint requires a revision bump whenever scorer behavior changes."*

I reverted the #346 `arguments` requirement in `_looks_like_a_bare_function` — a real change to scorer semantics, and **the exact one #404 cites as motivation** — without touching the revision:

```
fingerprint now : 5dc4c315ed2c6d4d
locked constant : 5dc4c315ed2c6d4d
MATCHES (scorer change went UNDETECTED): True
mini_tool_call  : 0.0   (unchanged)
```

Your own new suite still passes 6/6 with that revert in place.

The cause is `_fingerprint_generator` (`gate_suites.py:457`), a fixed three-branch stub. Measured per-suite under it:

```
mini_arithmetic   0.0000   <- pinned
mini_common_sense 0.5833
mini_format_json  1.0000   <- pinned
mini_instruction  0.0000   <- pinned
mini_mmlu         0.4615
mini_over_refusal 1.0000   <- pinned
mini_safety       0.0000   <- pinned
mini_tool_call    0.0000   <- pinned
```

**Six of eight suites sit at the floor or the ceiling, so they cannot move in at least one direction** — only two are interior. This is precisely the argument the project already makes in #405: *a suite pinned at the ceiling detects a regression as poorly as one pinned at zero*. The fingerprint has inherited that defect from its generator.

The fix is small and does not touch the design: make `_fingerprint_generator` emit per-suite responses that land each suite strictly between 0 and 1 — a deliberate mix of right and wrong answers rather than one canned string. Then refresh `BUNDLED_SCORER_FINGERPRINT`, and add the revert above as a test: change a scorer, assert the fingerprint moves.

## Blocking 2 — nothing in `src/` writes a stamped baseline file

```
$ grep -rn "write_baseline_file\|stamp_baseline_scores" src/
gate.py:139: def stamp_baseline_scores(...)
gate.py:165: def write_baseline_file(...)
gate.py:167:     payload = stamp_baseline_scores(scores)
```

Both are referenced **only from tests**. No `soup` command produces a `--baseline` JSON, so **every file-based baseline in the wild stays unstamped and warns on every run, forever** — which is #404's problem 1 (*"a warning that fires on the correct case too is a warning operators learn to ignore"*) surviving intact for the file path.

`docs/evaluation.md:493` already states baselines are "written via the shared `stamp_baseline_scores` / `write_baseline_file` helper". Today no user-reachable path does that, so the doc describes a producer that does not exist.

Either wire a producer (`ship --emit-evidence` or `eval gate` writing the stamped envelope) or correct the doc to say the file path is stamped only if you write it yourself. I would prefer the producer, since without one the file path never reaches the quiet state this PR is for.

To be clear about what *does* work: the `registry://` path is stamped via `tracker.py:483` and genuinely goes silent on a match. That half delivers the issue.

## Three MEDIUMs, each an untested guard

**3.** `gate.py:194-199`, the unknown-provenance branch. Neutering it changes nothing. It behaves **correctly** — I verified live that an unstamped registry entry yields `scores={'mini_mmlu': 0.55}, warns=1` — but nothing fails if it is removed. The new suite covers only the stamped case.

**4.** `ship.py:945` and `eval.py:1021`. Dropping the `warn=` callback at either site survives every test: it falls back to `logger.warning`, which the Rich CLI does not render, so the operator-facing warning silently disappears. I confirmed the wiring works today with a real run — `soup eval gate -s suite.yaml -b flat.json` printed `Warning: --baseline has unknown provenance (unstamped pre-existing file).` — but nothing guards it.

**5.** `_is_stamped_envelope` (`gate.py:181`) requires `keys() <= {scores, provenance}`. A stamped envelope carrying any extra key — `{"note": "captured on H100"}` — falls through to the legacy branch and dies with `ValueError: baseline file: score for 'scores' must be a number, got dict` → exit 3. The strict subset is a deliberate documented choice, but its failure mode should warn, not crash with a message naming neither the problem nor the fix.

## On the derived controls — the thing I looked hardest at

The two deleted tests were derived **scans** over `MINI_BENCHMARKS`, and their direct target is gone with the constant, so nothing is orphaned by the deletion itself. But they *incidentally* guaranteed a property of `build_mcq_prompt` across every suite, and what replaces them (`test_v07302.py:1679-1706`) is hand-written single examples. That is derived → hand-listed, and a suite added later is no longer swept.

Credit where due: your replacement **does** carry a real derived control — `bundled_scorer_fingerprint` iterates `sorted(DEFAULT_GENERAL_SUITE)`, so a newly added suite changes the hash and forces a bump. That is exactly the right shape. It is just undermined by blocking 1, which is why fixing the generator matters more than the lost scans.

## Small

**6.** `tracker.py:481` — `dict(details) if isinstance(details, dict) else {}` silently discards a non-dict `details` where the previous `json.dumps(details, default=str)` serialised whatever it got. All three callers comply with the declared type, so this is latent, but it turns a contract violation into silent data loss rather than an error.

**7.** CHANGELOG credit is wrong on both entries: `(#404 by @AchuthReddy-16)` is missing `in #485`, and the "Changed" entry is `(by @AchuthReddy-16)` with no number at all. House format is `(#NNN by @user in #PR)` — the adjacent `(#465 by @harshitthek in #467)` shows it.

**8.** A stray blank line landed inside the unrelated #443 `data.interleave` CHANGELOG paragraph, splitting `` `DataConfig.train` now accepts `str | `` from `` list[str]`; `` mid-sentence.

## What I could not verify

`ship.py`'s console warning surfacing live needs a real model load; I verified the structurally identical `eval.py` path with a real run instead, and finding 4 records that nothing would catch its removal. I ran 896 related tests rather than the full ~18k suite — your own `pytest tests/ -v` box is unchecked and I did not close that gap either.

Fix the two blocking items and this lands. The design is right — a stamp compared on read, with the constant deleted rather than joined by a sibling — and the registry half already delivers what #404 asked for. Thank you for taking a subtle issue and getting the shape of it correct.

AchuthReddy-16:
Addressed the review findings and rebased onto the latest `upstream/main`.

The two blockers are fixed:

* Replaced the pinned fingerprint generator with a deterministic per-suite corpus whose scores all remain strictly between 0 and 1. Reverting the #346 `arguments` requirement now changes the fingerprint and fails the revision lock.
* Added a user-facing producer: `soup eval gate --model <id> --write-baseline <path>`. It writes the shared stamped envelope, requires a real model, refuses empty score maps, and round-trips silently through `resolve_baseline` when revisions match.

Also added guards for:

* Unstamped `registry://` provenance warning exactly once
* Operator-facing `warn=` plumbing in both `eval` and `ship`
* Stamped envelopes containing unsupported extra keys
* Non-dict tracker details raising instead of being silently discarded
* Correct CHANGELOG attribution and the unrelated #443 formatting repair

Post-rebase verification:

* `ruff check src/soup_cli/ tests/` passed
* 227 focused and related tests passed
* `git diff --check` passed

CI is rerunning on `626504a`.

MakazhanAlpamys:
**Merged** as `610c581`, CONTRIBUTORS.md bullet in (`7640f2f`). CI 13/13, SHA verified against the head.

## Both blockers re-verified by hand, not read

**The fingerprint now actually locks.** Same mutation as before — revert the #346 `arguments` requirement, touch nothing else:

```
before your fix:  fingerprint 5dc4c315 == locked 5dc4c315   -> change UNDETECTED, 6/6 tests green
after  your fix:  fingerprint 8ca16d2c != locked 1474d3f3   -> 3 tests fail
```

And the per-suite distribution, which is the root cause you fixed:

```
before: 6 of 8 suites pinned at exactly 0.0 or 1.0
after:  0 of 8 pinned   (0.5000 across the board, mini_tool_call 0.6750)
```

**You turned my mutation into a permanent test.** `test_removing_arguments_requirement_moves_fingerprint` encodes the exact revert I used, and `test_every_suite_fingerprint_score_is_strictly_between_0_and_1` guards the generator's own non-degeneracy — so the property cannot quietly regress the way it did the first time. That is the difference between fixing a finding and fixing the class of finding.

**The producer exists and the round-trip is silent.** Ran it end to end:

```
write_baseline_file -> {'provenance': {'scorer_revision': 1, 'soup_version': '0.73.3'}, 'scores': {...}}
resolve_baseline    -> scores returned, warnings: []          <- SILENT, which was the entire point
unstamped file      -> exactly 1 warning, "unknown provenance (unstamped pre-existing file)"
```

Once per *resolve*, not once per suite — I checked that specifically, because a two-suite file warning twice would have reintroduced the noise #404 was about.

All three MEDIUMs also die by name now: `test_unstamped_registry_baseline_warns_once`, `test_eval_passes_warn_callback` + `test_eval_gate_cli_shows_unknown_provenance_warning`, and `test_envelope_with_extra_key_warns_and_resolves`. I mutated the `ship.py` `warn=` site too and both its tests fire.

## On the 68 deleted lines

I checked this first, because a shrinking test file is where I always look. Deleting `TestStaleBaselineIsAnnounced` is sanctioned by #404 itself, the deletion is confined to two hunks, and **no assertion outside that class was weakened** — the only other edit is a removed docstring with the body unchanged. You did the removal cleanly.

## One nit for next time

`test_ship_passes_warn_callback` asserts on `inspect.getsource(ship_mod._verdict_live)` — i.e. on the source **text**. That is the spelling-not-behaviour pattern this project has been bitten by before (v0.73.3 #321: a test battery that passed when the behaviour was killed but the strings kept, and failed when only the spelling changed). It survives here because the behavioural `test_ship_cli_shows_unknown_provenance_warning` catches the same mutation — I verified both fire — so it is belt-and-braces rather than the only guard. Worth dropping when you next touch that file: a source-text assertion fails on harmless refactors and passes on real breakage.

## What I could not verify

I ran 20 focused plus ~900 related tests, not the full ~18k suite — your own `pytest tests/ -v` box was unchecked and I did not close that gap either. `ship.py`'s console warning needs a real model load, so I exercised the structurally identical `eval.py` path instead and relied on the mutation to prove the `ship` wiring is guarded.

You took a subtle issue, got the design right on the first attempt, and when I showed the lock was decorative you fixed the mechanism rather than the symptom. Thank you.

Issue #326, opened 2026-08-03
trl >= 0.27: six preference trainers cannot build their config (dependency cap)


MakazhanAlpamys:
`trl` is pinned `>=0.7.0,<0.25` as of v0.72.4. Lifting that cap means migrating the preference trainers to the newer API.

## What breaks

Six trainers pass `max_prompt_length` to their `trl` config: `bco`, `dpo`, `ipo`, `kto`, `orpo`, `simpo`. `trl` removed it in stages, and 0.29 additionally deleted `ORPOConfig` and `CPOConfig` outright. Read off the published wheels, per config:

| `trl` | dpo | kto | orpo | cpo | bco |
|---|---|---|---|---|---|
| 0.24.0 | yes | yes | yes | yes | yes |
| 0.25.1 | yes | yes | yes | yes | **no** |
| 0.26.0 | yes | **no** | **no** | **no** | no |
| 0.29.0 | **no** | no | *gone* | *gone* | *gone* |

So on a recent `trl`: `soup train --task bco` breaks at 0.25, `--task kto|orpo|simpo` at 0.26, `--task dpo|ipo` at 0.29.

## Why it went unnoticed

The `trl` imports live inside `setup()`, and no test had ever called `setup()` on those wrappers — they were only instantiated. CI therefore stayed green against a `trl` the code could not actually run on, and `pip install 'soup-cli[train]'` produced a broken install. The end-to-end preference tests added in v0.72.4 (`tests/test_v07204.py`) are what surfaced it.

## Fix path

1. Find the replacement for `max_prompt_length` in each config (likely folded into a shared `max_length` / prompt-truncation option) and thread it per trainer.
2. `ORPOConfig` / `CPOConfig` are gone — identify what ORPO and SimPO map onto in the newer API.
3. Raise the cap only for versions actually exercised by a test that calls `setup()`.

## Acceptance criteria

- [ ] `soup train --task {bco,dpo,ipo,kto,orpo,simpo}` constructs its trainer on the newest `trl`
- [ ] a test calls `setup()` for each of the six, so the next removal cannot be silent
- [ ] the `pyproject.toml` cap is raised, with the evidence table updated

MakazhanAlpamys:
**The cap this issue tracks was itself wrong.** Corrected on `main`: `trl>=0.14.0,<0.27` (was `>=0.7.0,<0.25`).

[...]

Issue #478, opened 2026-08-22
CI never exercises the declared transformers floor, so a >=4.56-only kwarg passes 12/12


MakazhanAlpamys:
Found while reviewing #471, which was caught by hand rather than by CI. The PR is fine; the gap is ours.

## The gap

`pyproject.toml` declares:

```
transformers>=4.36.0,<5.0.0
```

CI installs with `pip install -e ".[dev]"`, which resolves to the **newest** satisfying version (4.57.x today). So the floor of our own declared range is **never executed**. A change that works on 4.57 and is a hard `TypeError` on 4.55 goes green on all 12 cells.

That is not hypothetical — it is exactly what happened in #471:

```
transformers 4.55.4   dtype='auto'  -> TypeError: __init__() got an unexpected keyword argument 'dtype'
transformers 4.57.6   dtype='auto'  -> loaded
```

`dtype=` is the >= 4.56 rename of `torch_dtype=`. Twenty minor versions inside our declared support would have failed at model load, before step 0, on text/vision/audio/QLoRA alike.

## Why it matters more than a version skew usually would

**pip does not upgrade an already-satisfying dependency.** Someone sitting on 4.55.4 who runs `pip install soup-cli[train]` keeps 4.55.4 — our pin is satisfied. That is the older-image Colab/Kaggle population, i.e. the same fleet #385/#387 was about.

`grep -rn "4.56" .` returns nothing: no test, no comment, no constraint anywhere records that a floor exists to respect.

## Two things this is NOT

- Not an argument for raising the floor. Deciding `>=4.56` is a dependency-policy call with its own cost (it drops those users rather than supporting them).
- Not solved by a lockfile. `soup env lock` records what a user *has*; this is about what our pin *promises*.

## Acceptance

- [ ] Something exercises the declared floor. Cheapest credible form is one CI cell that installs `transformers==4.36.0` (or the lowest version that actually resolves with our other pins) and runs a targeted subset — model-construction paths, not the full 18k suite.
- [ ] It must FAIL on a `dtype=`-style regression. Verify by writing the failure first: put `dtype=` back at one `from_pretrained` site and confirm the new cell goes red while the existing 12 stay green. A floor cell that cannot fail is worse than none, because it looks like coverage.
[...]

PR #486, opened 2026-08-22
ci: catch dependency drift on a schedule, and ask trl what it accepts

### LINKED ISSUE #323: CI installs a much newer stack than any dev box — a degenerate NF4 fixture survived three releases because of it
## What happened

v0.72.3's CI went red on two NF4 bit-exactness tests — one of them a **v0.72.2 test that had been green since it shipped**:

```
RuntimeError: shape '[32, 0]' is invalid for input of size 16
  bitsandbytes/nn/modules.py:620: _convert_weight_packed_for_cpu
```

Root cause, reproduced directly rather than inferred: bitsandbytes' CPU 4-bit forward reshapes absmax to `[rows, blocks_per_row]`. The test fixtures used `hidden_size=32`, giving 16 absmax blocks for 32 rows, so `blocks_per_row` floors to **zero**.

```
hidden=  32  absmax blocks=  16  -> RuntimeError: shape '[32, 0]' is invalid for input of size 16
hidden=  64  absmax blocks=  64  -> OK
hidden= 128  absmax blocks= 256  -> OK
```

Fixed in v0.72.3 by moving the fixtures to `hidden_size=64`.

## Why it went unnoticed for three releases

**A CUDA build never calls `_convert_weight_packed_for_cpu`.** The bug is structurally invisible on any GPU development machine and only reachable on a CPU-only runner. It shipped through v0.72.0, v0.72.1 and v0.72.2 without a single local run being able to see it.

It surfaced now because CI resolves latest on every run, and the gap had grown large:

| package | dev box | CI |
|---|---|---|
| torch | 2.5.1+cu121 | **2.13.0** |
| peft | 0.18.1 | **0.20.0** |
| trl | 0.19.1 | **1.9.2** |
| bitsandbytes | 0.49.2 | **0.50.0** |

A full green local suite (16844 passed) is therefore a materially weaker signal than it looks — it is one major torch version behind what CI actually tests.

## Options

- [ ] A scheduled CI job that installs the *current* floors and reports drift, so the gap is visible before it bites
- [ ] A documented "refresh the dev stack" step in the release checklist
- [ ] CPU-only test coverage for the 4-bit paths that a CUDA box cannot reach — the specific blind spot here
- [ ] Consider constraint ceilings for the core training deps, weighed against pinning users to old stacks

## Acceptance

- [ ] A class of bug that only manifests on CPU-only runners is catchable before it reaches `main`
- [ ] The dev-vs-CI version gap is surfaced somewhere a maintainer will see it

MakazhanAlpamys:
## What does this PR do?

Closes #323 — both acceptance boxes.

The issue is written as four options rather than a fix path, so: **taking options 1 and 3, because they turn out to be the same job**, plus the guard from the issue's own comment. Option 4 is deliberately declined — reasoning at the bottom.

## The job already found a real defect, before merging

```
### Conflicting declarations
- transformers: <5.0.0,>=4.36.0 vs >=5.0.0
```

`[mlx]` declares `transformers>=5.0.0` while `[train]` declares `transformers>=4.36.0,<5.0.0`. Those are mutually unsatisfiable: **`pip install "soup-cli[train,mlx]"` cannot resolve for anybody.** Nothing in CI installs both extras together, so nothing was in a position to notice — `mlx-smoke` installs `[mlx]` alone, on purpose, and the matrix installs `[dev]`, which does not pull `[mlx]`.

I have not changed the pins here; which side gives is your call, not a drive-by in a CI PR. What this PR does is make the class of defect report itself.

## Options 1 + 3: one scheduled job answers both

Both acceptance boxes are satisfied by the same workflow, because the two failure classes have the same shape — *nothing runs against the newer stack until a PR happens to be open*:

- **A CPU-only-manifesting bug** cannot be caught by any amount of local testing on a GPU box. A CUDA build never calls `_convert_weight_packed_for_cpu`, which is precisely why the degenerate NF4 fixture survived three releases.
- **An upstream removal behind a floor-only pin** ships with CI green, as the comment on the issue documents: `trl>=0.7.0` let CI resolve 0.29.1 against a dev box on 0.19.1, and six trainers could not build their config at all.

`.github/workflows/dependency-drift.yml` runs Mondays 06:00 UTC (plus `workflow_dispatch`): installs the latest resolvable `.[dev]`, runs the full suite against it, and writes the table into `$GITHUB_STEP_SUMMARY` — the second acceptance box asks for the gap to be *somewhere a maintainer will see it*, and a run summary is that place; logs are not.

Resolved is compared against **the specifier declared in the package metadata**, not against a committed snapshot. No state file to go stale, and it measures the actual root cause named in the comment: an unbounded floor-only pin. Local run of that script, right now:

| package | declared | resolved here | upper bound |
|---|---|---|---|
| accelerate | >=0.25.0 | 1.14.0 | NO |
| bitsandbytes | >=0.41.0 | 0.50.1 | NO |
| datasets | >=2.14.0 | 5.0.1 | NO |
| peft | >=0.7.0 | 0.20.0 | NO |
| torch | >=2.0.0 | 2.13.0 | NO |
| trl | <0.29,>=0.14.0 | 0.28.0 | yes |

`datasets` is three majors above its floor. That is the number the issue is about.

## The guard from the issue's comment

> an unbounded floor-only pin plus a test that never reaches the import is indistinguishable from working software. The cheap guard is a test that actually calls the constructor for every wrapper.

Calling `setup()` needs a model and a dataset. `tests/test_issue323_trl_kwarg_drift.py` asks the same question without either: read the keywords each wrapper passes to its trl config, and ask the class **as installed** whether it still accepts them — through `_trl_compat.config_accepts`, the same capability probe the wrappers already use at runtime. Never a version comparison; a version table is what was wrong twice.

It follows the `resolve_trl_symbol` indirection, so the three configs #326 moved to `trl.experimental` are covered rather than silently skipped — those are exactly the ones that had the harder break, and a scan that missed them would cover only the wrappers that were never in danger.

**Two blind spots, stated rather than implied:**

1. `**splat` calls are invisible to a static read — `grpo.py` and `sft.py` build their configs that way. What *is* covered is every keyword written literally at the call site, which is where `max_prompt_length` lived before #326 moved it behind a helper. `test_splat_only_calls_are_skipped_not_silently_passed` pins this as a known gap.
2. The trl **trainer** classes are excluded. I did not assume that — the first version of the scan included them and reported `model`, `train_dataset`, `eval_dataset` and `processing_class` as "rejected" by `ORPOTrainer`. They are not; the trainers take much of their signature through `**kwargs`, which `inspect.signature` cannot see. Trainer-argument drift is a real question and a different one.

The scanner has its own controls (`TestTheScannerCanActuallyFail`), including one that the exemption list cannot be abused: every Soup-owned name in it must be a class Soup actually defines, checked by grep over `src/`.

## Option 4 (constraint ceilings) — declined, with reasoning

Caps are already applied case by case where a break is *known* (`trl<0.29`, `mcp<2`). A blanket ceiling on the core training deps pins users to old stacks to protect a test suite, which is the wrong trade — and it would suppress exactly the signal option 1 exists to generate. Better to learn on Monday that `datasets` 6.0 broke something than to never learn it because the resolver was forbidden to try.

## One honest limitation

**A scheduled workflow cannot be proven green before it is on the default branch** — `workflow_dispatch` only dispatches from there. What I could verify, I did: the YAML parses, and the report script was run locally end to end (its output is the table and the conflict block above). The first real run happens after merge, and if the schedule or the summary write turns out wrong, that is a follow-up rather than something this PR can rule out.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [x] `pytest tests/ -v` passes — **18080 passed, 377 skipped, 0 failed**, coverage 81% (gate 77%)
- [x] Updated relevant docs — `CONTRIBUTING.md` gains step 0 of the release checklist (refresh the stack, or read the last drift run); `CHANGELOG.md` has the entry
- [x] No breaking changes — one new scheduled workflow and one new test file; no source changes at all


MakazhanAlpamys:
Rebased onto `main`, and two things changed underneath this PR that are worth stating rather than leaving a reviewer to work out.

**1. The changelog entry is now a fragment.** #490 landed the `changelog.d/` mechanism while this was open, so the `CHANGELOG.md` edit is gone and the entry lives in `changelog.d/0.73.3/486.added.md`. The `CONTRIBUTING.md` release checklist also conflicted — `main` renumbered it around the new `assemble_changelog.py` step — so I kept `main`'s numbering intact and this PR's addition stays as step 0, ahead of it. Nothing of either side was dropped; `test_issue487_changelog_fragments.py` passes.

**2. `transformers-floor` and `dependency drift` are two halves, not a duplicate.** #478 closed with a CI job that installs against a floor constraint. This PR adds one that installs whatever resolves *newest*. They test opposite ends of the same declared range and neither substitutes for the other:

| | what it installs | what it catches |
|---|---|---|
| `transformers-floor` (#478, merged) | the declared **floor** | code that silently requires a version newer than the floor promises |
| `dependency drift` (this PR) | the latest **resolvable** | an upstream removal, and a bug that only appears on a stack nobody has locally |

The v0.72.x incident this issue is about needed the second: the dev box ran torch 2.5.1 / trl 0.19.1 while CI resolved torch 2.13.0 / trl 0.29.1. A floor job would have been green throughout, because the floor was not the problem — the ceiling was, and there wasn't one.

The conflict-detection half of the report is also still live and still finding the `[mlx]` vs `[train]` `transformers` contradiction, which neither job would surface on its own since nothing installs both extras together.

PR #488, opened 2026-08-22
fix(sft): collate SmolVLM vision batches end to end

### LINKED ISSUE #302: Idefics3 / SmolVLM vision-SFT support in the LLaVA path
### Context

v0.71.32 shipped the `smolvlm-256m-sft` recipe (a tiny VLM). During the live vision smoke on the dev box (RTX 3050), `soup train` on `HuggingFaceTB/SmolVLM-256M-Instruct` surfaced a **pre-existing** incompatibility in the shared LLaVA vision path — SmolVLM uses an `Idefics3Processor`, which the LLaVA-era code assumes is a LLaVA-style processor.

### What happens

1. `training.lora.target_modules: auto` cannot be inferred by peft for Idefics3 → `ValueError: Please specify target_modules ...` (worked around in the recipe by pinning `target_modules: [q_proj, v_proj]`).
2. With explicit target modules, LoRA applies (755K trainable), then the SFT setup fails:
   ```
   AttributeError: 'Idefics3Processor' object has no attribute 'pad_token'
   ```
   The vision path sets `self.tokenizer = <processor>` and accesses `self.tokenizer.pad_token`; on Idefics3 the pad token lives on `processor.tokenizer.pad_token`.

### Proposed fix path

- In `trainer/sft.py`, when the vision processor is an Idefics3/SmolVLM processor, route tokenizer-level attribute access (`pad_token`, etc.) through `processor.tokenizer`.
- Audit the rest of the vision collation for other Idefics3-vs-LLaVA processor API differences (image-token expansion, chat template) so a full SFT step runs end-to-end.
- Guard against regressing the existing LLaVA / ShareGPT4V path (Qwen2-VL, Pixtral, LLaMA-3.2-Vision).

### Acceptance criteria

- [ ] `soup train` on `HuggingFaceTB/SmolVLM-256M-Instruct` (llava format, `modality: vision`) completes at least one training step on CPU/tiny data without the `pad_token` / target-modules errors.
- [ ] The existing LLaVA vision recipes still train (no regression).
- [ ] `smolvlm-256m-sft` recipe upgraded from parse-only to live, with the recipe note updated.

Ships alongside v0.71.32 as a known limitation; the recipe currently ships parse-tested.

Amix29:
Closes #302.

## Problem

SmolVLM setup and tokenization succeeded, but the first real training batch failed with:

```
ValueError: The number of images in the text [0] and images [1] should be the same.
```

Idefics3 does not preserve Soup legacy string-form `<image>` marker when that string is passed directly to `apply_chat_template`. The rendered prompt therefore contains no image token while the processor still receives one image.

## Fix

- Keep raw messages and PIL images together until data-loader collation.
- Convert legacy `<image>` markers into structured Hugging Face multimodal image parts.
- Let the model processor perform image-token expansion and emit architecture-specific pixel tensors.
- Use the Transformers trainer with a processor-aware collator for ordinary non-packed vision SFT, avoiding the text-only preprocessing path in older supported TRL releases.
- Preserve the supported old/new Transformers trainer keyword boundary.
- Update the SmolVLM recipe description now that a real training step is covered.

## Validation

- `70 passed` across the #302 regression suite and existing vision suite
- Full repository run: `18,242 passed, 285 skipped`; the only initial failure was the host-dependent export test detecting an already-installed `llama-quantize`, and that test passes under a clean `PATH`
- Real SmolVLM-256M one-step training on Apple M4 Max / MPS: completed, loss `22.4004`, adapter saved
- Real forced-CPU one-step training on the same machine: completed, loss `22.4004`, adapter saved

No local machine paths are included in this report.


MakazhanAlpamys:
**The architectural call here is right, and it is the best-tested of your four PRs — 20 of 22 mutation sites die by name.** Two findings block, and neither is a defect in what you set out to fix. Both are things the new code path now *owns* and nothing checks. I measured both against real processors rather than reasoning about them.

## The diagnosis is better than the issue's

#302 proposed routing tokenizer access through `processor.tokenizer` — that half already shipped in `481dc38`. You correctly identified that the real remaining blocker was elsewhere: `apply_chat_template` was called at **dataset-prep** time on a legacy `<image>` **string**, which Idefics3 does not preserve, so the rendered text carried zero image tokens while one image still reached the processor. Moving templating to collation time and converting `<image>` → `{"type": "image"}` is what upstream does. Good call, and it is why the fix works.

Verified against three real processors (Idefics3, LLaVA-1.5, Qwen2-VL): all three collate through the new path, and Qwen2-VL correctly receives `image_grid_thw`.

**No test deleted, no assertion weakened.** Worth stating how I checked, because the default view is misleading: `git diff origin/main...pr488` shows `test_issue302_vision_pad_token.py | 243 +0`, which reads as a net-new file — it is not, main already has it from `481dc38`, and your branch predates that. The correct comparison, `git diff origin/main pr488 -- <file>`, is a pure addition: all 10 pre-existing `TestEnsureVisionProcessorPadToken` tests survive byte-identical.

Pre-fix control at the merge-base: **7 failed, 11 passed** — and one of the seven, `test_dataset_keeps_messages_until_collation`, fails *behaviourally* rather than on ImportError. That is the strongest signal available for a change of this shape.

## Blocking 1 — 95.4% of trained label positions are `<image>` placeholders

`src/soup_cli/trainer/sft.py:211-213` masks padding and nothing else. Measured on the **real** `Idefics3Processor` at the shipped recipe's `max_length: 2048`, batch of 2:

```
ids (2, 1141)  labels (2, 1141)
row0: image tokens=1088, of which TRAINED=1088, total trained=1141/1141  -> 95.4% is <image>
row1: identical
```

I checked the full causal chain rather than stopping at the count, because the docstring at `modeling_idefics3.py:838` says image tokens "are ignored (masked)" and that would have made this a non-finding:

- `image_token_id` = **49190**; `text_config.vocab_size` = **49280** → it is a valid index, so cross-entropy is computed on it rather than erroring.
- `modeling_idefics3.py:925` is `loss = self.loss_function(logits=logits, labels=labels, vocab_size=...)` — **no masking of `image_token_id` anywhere in the code path**. Line 838 states an expectation on the *caller*; `ForCausalLMLoss` ignores only `-100`.

So the model spends ~95% of its gradient signal predicting a placeholder token, and the objective on the actual answer is diluted roughly 1088:7. It does not crash and produces a plausible loss curve — which is exactly the class of bug this project shipped in #430, where assistant-only masking trained **0 tokens** behind a normal-looking curve.

HF's own SmolVLM finetuning recipe does `labels[labels == image_token_id] = -100`.

**Fix**: mask the image token id (`processor.tokenizer.convert_tokens_to_ids(processor.image_token)` or `model.config.image_token_id`), plus the Idefics3 `<fake_token_around_image>` / `<row_R_col_C>` scaffolding if you want upstream parity. **Please assert the *count* of non-masked labels**, not just their positions — the two surviving mutations below show label content is only spot-checked today.

## Blocking 2 — `add_special_tokens=False` drops BOS for LLaVA-1.5

`sft.py:203-205`. The comment says chat templates already supply BOS/EOS. True for SmolVLM and Qwen2-VL; **false for LLaVA-1.5**, one of the two architectures #302 names as must-not-regress. Measured:

```
llava-1.5-7b-hf        True=[1, 3148, 1001, 29901]  False=[3148, 1001, 29901, 29871]  -> BOS LOST
SmolVLM-256M-Instruct  True=[23824, 42, 216, 49190] False=[23824, 42, 216, 49190]     -> no diff
```

Pre-PR, TRL tokenized the `text` column at `sft_trainer.py:775` with no `add_special_tokens` argument, i.e. the default `True`, BOS present. So LLaVA-family vision runs now train without BOS on a Llama backbone — a silent distribution shift.

**Mutation M1 flipping this flag left all 98 tests green**, so a change in either direction goes unnoticed.

**Fix**: derive it from the template rather than hardcoding — compare `tokenizer(text)` against `tokenizer(text, add_special_tokens=False)` for a leading BOS once at collator construction — or at minimum pin the current value by asserting the resulting `input_ids`, not that a kwarg was passed.

## Four MEDIUMs

**3. The collator is only ever tested at batch size 1.** `sft.py:193-194`. Replacing `images.append(...)` with `images[:] = [...]` — an N-example batch silently collapsing to its last row — left all 98 tests green. Same for `texts`. A collator's whole contract is "list in, batched tensors out", and the shipped recipe uses `batch_size: auto`, which commonly resolves above 1; a regression here trains on 1/N of the data with no error. One test with two examples of *different* lengths, asserting `input_ids.shape[0] == 2` and that the shorter row's padding is masked, closes it. (It works today — it just is not guarded.)

**4. `truncation=True` can slice image tokens into a hard error.** At `max_length: 512`: `ValueError: Mismatch in image token count ... Got ids=[482, 472] and text=[1088, 1088]`. SmolVLM expands one image to 1088 tokens, so any `max_length` below ~1150 fails at the first batch, and `data.max_length` is user-settable down to 64. The RAFT path already sets the precedent of dropping and *reporting* rows that fill `max_length`. Pre-flight it, or re-raise naming `data.max_length`.

**5. A single unopenable image now aborts the run at step 0.** `_prepare_vision_dataset` still warns-and-continues on a bad path, but the new collator then raises `Vision sample contains 1 image placeholder(s) but only 0 image(s)` — and names neither the file nor the row index in a 100k-row dataset. Pre-PR that row trained as text-only. The two halves disagree; either drop the marker when loading failed, or drop the row.

**6. `modality: vision` + `packing`/`multipack` parse cleanly and now get a dataset shape they cannot consume.** With `packing: true`, `use_vision and not tcfg.packing` is False, so it falls through to `SFTTrainer(packing=True)` on a dataset whose columns are `messages` and `images` with **no `text` column at all**. Pre-PR both at least received one. A `SoupConfig` cross-validator rejecting the combination beats a silent trap.

## Small

**7.** `_vision_messages_with_image_parts`' docstring claims "existing structured messages are preserved" and nothing tests it — removing the structured-part counter left 98 tests green, and the consequence is not benign: an already-structured row gets `represented_images == 0`, so `missing_images` injects a **duplicate** placeholder. Six shapes were characterised and all six behave correctly today; a parametrized test costs four lines.

**8.** A row with no user message gets the image injected into the **assistant** turn, which lands in the trained span. Rare; an explicit refusal reads better than a silent fallback.

**9.** The new branch prints nothing, while every sibling (`RAFT`, `multipack`, `packing`) prints a Rich confirmation — a user cannot tell which trainer they got.

**10.** `grep -rn "SmolVLM\|Idefics3" docs/` is empty before and after. `docs/models.md` or `docs/training.md` wants a line, especially "`max_length` must exceed the model's per-image token expansion" given finding 4.

CHANGELOG credit is correct (`#302 by @Amix29 in #488`) — thank you for that.

## The CONTRIBUTORS.md conflict is mine, not yours

Your branch carries `463685c docs(contributors): AmixDigital's Apple Silicon monitor (#481)`. Main carries `2428453` — **same title, same timestamp, byte-identical hunk**, differing only in the author email. That is my commit: I committed it with the wrong email, GitHub left it unlinked to my account, and I amended and force-pushed it. Your branch kept the pre-amend copy.

Drop your branch's copy when you merge `main` — nothing of yours is at stake, and CHANGELOG auto-merges cleanly this time.

## What I could not verify

**No VLM weights and no MPS here.** Everything above about token counts, label masking and BOS is measured against the **real processors**, which I do have. Your "loss 22.4004, adapter saved" on M4 Max is not independently reproduced, and **finding 1's training-quality impact is inferred, not measured** — I established the mechanism precisely but did not run enough steps to show adapter degradation. A 22.4 loss is *consistent* with an image-token-dominated objective, and I am not claiming that as proof.

Fix 1 and 2 and this lands. The core is right, the pre-fix control is honest, and 20 of 22 mutations dying by name is the strongest test showing you have sent. The gaps are all at edges the tests never reach — batch > 1, `add_special_tokens`, and label content.

MakazhanAlpamys:
Correction to my last comment: **the `CONTRIBUTORS.md` conflict is gone, so ignore that paragraph.** Later merges moved the merge base and git now resolves the duplicated commit on its own — I re-ran `git merge origin/main --no-commit --no-ff` on your head and it reports `Automatic merge went well`, with exactly one AmixDigital entry and no conflicted files. Nothing for you to drop.

GitHub agrees: the PR reads `CLEAN`, and CI is green on all 13 cells at `763947d`.

The two blocking findings stand unchanged — the image-token label masking and the LLaVA BOS regression. That is all this is waiting on.

Amix29:
Addressed both blocking findings in `e1870cd`.

## 1. Image-placeholder labels

`VisionLanguageDataCollator` now resolves the processor's image-token ids through the current `image_token_ids` / `image_token_id` surfaces, with a compatibility fallback through `processor.image_token` and `tokenizer.convert_tokens_to_ids`. Every matching `input_ids` position is set to `-100` in `labels`, in addition to the existing padding mask.

The regression test asserts the complete label row and the exact number of non-masked positions, rather than spot-checking one token. A second two-example test uses rows of different lengths and pins both the batch dimension and each row's independent image/padding mask.

Measured with real processors under the PR's supported Transformers environment:

- Idefics3 / SmolVLM: 1,088 image positions, **0 trained image positions**, 57 remaining trained positions.
- LLaVA-1.5: 576 image positions, **0 trained image positions**, 20 remaining trained positions.

Neutralising only the new image-token masking loop makes `test_masks_every_image_token_and_pins_the_trained_label_count` fail.

## 2. LLaVA BOS preservation

The collator now compares the first real rendered prompt through the underlying tokenizer once, with and without special tokens, and caches the decision:

- when `add_special_tokens=True` supplies a missing leading BOS, the processor receives `True`;
- when the rendered chat template already starts with BOS, the processor receives `False`, avoiding duplication.

Real-processor controls:

- LLaVA-1.5 selects `add_special_tokens=True`; output starts with token `1`, matching `bos_token_id=1`.
- Idefics3 selects `False`; its rendered prompt already starts with token `1`, so BOS remains single.

Forcing the processor call back to `add_special_tokens=False` makes `test_adds_bos_when_the_chat_template_does_not_supply_it` fail.

## Validation

- `22 passed` in `test_issue302_vision_pad_token.py`
- `2,936 passed, 3 skipped` across the focused vision/config/recipe suites
- Full suite: `18,298 passed, 285 skipped, 5 deselected`; the sole failure is the pre-existing host-dependent `test_find_quantize_binary_not_found`, because this Mac has `/opt/homebrew/bin/llama-quantize`. That test passes with a clean system `PATH` (`1 passed`).
- `ruff check src/soup_cli/ tests/` passes
- `git diff --check` passes

The branch also contains the clean merge of current `origin/main`; no `CONTRIBUTORS.md` resolution was needed, matching your correction above.

MakazhanAlpamys:
**Both blockers verified closed on `e1870cd`, measured against the real processors rather than read.** CI is still finishing; nothing else is outstanding.

## Blocker 1 — label masking

```
                image tokens   TRAINED among them   total trained
before (763947d)     1088            1088             1141 / 1141   -> 95.4% was <image>
after  (e1870cd)     1088               0               53 / 1141   -> 0.0%
```

53 is exactly the text. `sft.py:290-291` resolves the ids through `image_token_ids` / `image_token_id` with a fallback via `processor.image_token`, and masks every matching position.

Mutation — delete the masking loop:

```
FAILED test_masks_every_image_token_and_pins_the_trained_label_count
FAILED test_two_different_examples_remain_two_rows_and_mask_independently
```

**The first test pins the trained-label COUNT**, which is what I asked for and the thing that would have caught this class in the first place. A positions-only assertion would have passed on the broken version.

## Blocker 2 — BOS

`_processor_adds_leading_bos` compares the two tokenizer paths on the first real rendered prompt, so it is derived per architecture rather than hardcoded. On the real processors:

```
llava-1.5-7b-hf        add_special_tokens=True    <- BOS restored
SmolVLM-256M-Instruct  add_special_tokens=False   <- template supplies its own
```

Mutation — force it back to `return False`, i.e. the original hardcoded bug: `FAILED test_adds_bos_when_the_chat_template_does_not_supply_it`.

## Diff check, and a trap worth naming

`git diff 763947d..e1870cd -- tests/` appears to delete `TestStaleBaselineIsAnnounced` and 68 lines from `test_v07302.py`. **That is not yours** — it arrived through your merge of `main`, from #485. Your own change is `git diff origin/main...e1870cd -- tests/`: **357 insertions, 0 deletions.** Pure addition.

I check every deleted test line, so I am flagging the shape rather than the result: a two-dot diff across a merge attributes the incoming branch's changes to you. Worth knowing when someone reviews your work less carefully than this.

## One nit, downgraded after measuring it

My earlier finding 3 (collator only tested at batch size 1) is **partly** closed — `test_two_different_examples_remain_two_rows_and_mask_independently` covers rows and masking. But collapsing the image list still passes:

```
images.append(image_list)  ->  images[:] = [image_list]      # every row gets the LAST row's image
22 passed
```

`_MeasuredVisionProcessor` returns hardcoded `rows`, so it cannot see which image reached which row.

**I then checked what that actually costs, and it is much less than it looks.** Against the real processor the same mutation raises immediately:

```
ValueError: The number of images in the text [1, 1] and images [1] should be the same.
```

So it fails **loudly at the first batch**, not silently on wrong pairings — which is the difference between a defect and an untested guard. Downgrading it from the MEDIUM I filed to a nit: worth one test whose fixture counts the images it received, not worth another round.

## Status

Verification is done from my side. The moment CI finishes green on all 13 cells I merge — nothing else is waiting on you.

Two blocking findings, one push, both closed, and the test for the first one pins the property rather than the symptom. That is the fourth time today.

MakazhanAlpamys:
**Merged** as `664f5c9`, fourth CONTRIBUTORS.md bullet in (`4c2afa6`). CI 13/13 at `e1870cd`, SHA verified against the head.

Everything I ran is in [my previous comment](https://github.com/MakazhanAlpamys/Soup/pull/488#issuecomment-5381584707) — both blockers measured closed against the real processors, both mutations killed by name. Nothing changed since.

## What I want to name, because it is the reason this merged in one round

**The test for the label masking pins the trained-label COUNT**, not just which positions are masked. A positions-only assertion would have passed on the broken version — the bug *was* that 1088 positions were unmasked, and any test that only checked "the padding is masked" would have been green throughout. You picked the assertion that could actually fail.

**The BOS fix is derived, not hardcoded.** `_processor_adds_leading_bos` compares both tokenizer paths on the first rendered prompt, so it answers per architecture: llava → `True`, SmolVLM → `False`, both confirmed on the real processors. The easy fix was to flip the flag to whatever made LLaVA pass and leave a comment; that would have regressed SmolVLM the moment someone trusted the comment.

Both of these are the difference between fixing a bug and fixing the class of bug, and it is the fourth time today one of your PRs has done that.

## Two follow-ups, neither blocking, both already in the record

The image-list collapse (my earlier finding 3) still passes the suite because `_MeasuredVisionProcessor` returns hardcoded rows and cannot see which image reached which row. I measured the real cost before filing it as anything: on the real processor that mutation raises `ValueError: The number of images in the text [1, 1] and images [1] should be the same` at the first batch, so it fails loudly rather than mistraining. A fixture that counts the images it received would close it whenever you next touch that file.

The other four MEDIUMs from my first review — `truncation` slicing image tokens, an unopenable image now aborting at step 0, and `modality: vision` + `packing`/`multipack` parsing but producing an unusable dataset shape — are still open and were never in scope for this PR. If you want any of them, file and claim; otherwise I will pick them up.

Four merged PRs in about a day, each on a real issue, each closing its review findings on the first push. Thank you.

PR #489, opened 2026-08-22
ci: exercise Transformers compatibility floor (#478)

### LINKED ISSUE #478: CI never exercises the declared transformers floor, so a >=4.56-only kwarg passes 12/12

AchuthReddy-16:
## What does this PR do?

Fixes #478.

The 12 existing test cells install the newest Transformers 4.x, so they cannot detect code that accidentally depends on a newer API. That happened with `dtype=`: it works on Transformers 4.56+, while Soup still declares support from 4.36.

I reproduced the dependency boundary before choosing the CI pin. The declared `transformers==4.36.0` floor cannot resolve with `trl>=0.14,<0.29`, because even TRL 0.14.0 requires Transformers 4.46+. The new floor job therefore uses the lowest combination that resolves under the current declarations:

- `transformers==4.46.1`, the lowest non-yanked resolvable Transformers release
- `trl==0.14.0`, the lowest declared TRL release

The job installs with explicit constraints, runs `pip check`, and asserts both installed versions so pip cannot silently move the cell upward.

Production `AutoModel*` calls now use `torch_dtype=`, which works at this effective floor and on current Transformers. A source-derived AST test scans the production tree for `dtype=` on `from_pretrained` and `from_config`; deliberately changing one call back to `dtype=` makes the test fail with the file and line number.

The existing 12-cell matrix and the declared dependency range are unchanged. Changing the declared 4.36 floor is left as a separate dependency-policy decision.

## Verification

- 26 focused compatibility and model-loading tests passed
- Constrained install resolved at Transformers 4.46.1 / TRL 0.14.0
- `pip check` passed
- Production scan found zero incompatible `dtype=` calls

### DIFF
--- a/.github/constraints/transformers-floor.txt
+++ b/.github/constraints/transformers-floor.txt
@@ -1,0 +1,23 @@
+# #478 — Transformers pin for the dedicated floor-compat CI job.
+#
+# Declared train floor in pyproject.toml remains:
+#   transformers>=4.36.0,<5.0.0
+#
+# transformers==4.36.0 is not resolvable with the declared trl range
+# (trl>=0.14.0,<0.29): even trl 0.14.0 requires transformers>=4.46.0, and
+# newer trl releases require >=4.56.1 / >=4.56.2. Raising Soup's declared
+# transformers floor in pyproject.toml needs an explicit maintainer
+# dependency-policy decision. Do NOT treat 4.55.4 (the last pre-dtype=
+# release) as the declared floor.
+#
+# This file pins:
+#   transformers==4.46.1 — lowest non-yanked resolvable Transformers version
+#     tested with the lowest declared TRL version
+#   trl==0.14.0 — lowest version in Soup's declared trl range
+#
+# These pins are for the floor CI cell only so pip cannot silently upgrade
+# that job. Do not describe pip's preferred/latest compatible TRL as the
+# floor. Runtime dtype= regressions are caught by the static guard in
+# tests/test_transformers_floor_compat.py (dtype= is >=4.56-only).
+transformers==4.46.1
+trl==0.14.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,47 @@ jobs:
       - name: Run one-step MLX SFT through the real CLI
         run: pytest -o addopts= -m smoke -k mlx_sft_smoke -q tests/test_smoke_train.py
 
+  # #478: declared transformers floor (4.36.0) cannot resolve against declared
+  # trl>=0.14,<0.29 — see .github/constraints/transformers-floor.txt. Do not
+  # raise Soup's pin without a dependency-policy decision. This job installs
+  # under an explicit transformers(+trl) constraint so pip cannot silently
+  # upgrade the cell, asserts the pinned versions, and runs the static dtype=
+  # guard (the accepted fallback when a true floor runtime cell is impractical).
+  transformers-floor:
+    runs-on: ubuntu-latest
+    env:
+      PYTHONUTF8: "1"
+      PYTHONIOENCODING: "utf-8"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install with Transformers floor constraints
+        run: pip install -e ".[dev]" -c .github/constraints/transformers-floor.txt
+      - name: Verify constrained dependency graph
+        run: python -m pip check
+      - name: Assert constrained Transformers and TRL versions
+        run: |
+          python - <<'PY'
+          import importlib.metadata as md
+
+          expected = {"transformers": "4.46.1", "trl": "0.14.0"}
+          for name, want in expected.items():
+              got = md.version(name)
+              assert got == want, (
+                  f"floor job expected {name}=={want} from "
+                  f".github/constraints/transformers-floor.txt, got {got} — "
+                  f"pip silently upgraded, ignored -c, or constraints drifted"
+              )
+              print(f"{name}={got} (matches constraints)")
+          PY
+      - name: Run Transformers floor compatibility guard
+        run: >
+          pytest -o addopts= --no-cov -q
+          tests/test_transformers_floor_compat.py
+          tests/test_bugfixes.py::TestDiffModelLoading
+
   test:
     strategy:
       fail-fast: false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -211,6 +211,22 @@ reproducing 70+ versions of notes.
   with a pageable CPU source and MPS layer buffers; CUDA pinning behaviour is
   unchanged.
 
+- **CI and production load sites no longer assume Transformers ``dtype=``
+  (#478).** ``dtype=`` on ``AutoModel*.from_pretrained`` / ``from_config`` is the
+  >=4.56 rename of ``torch_dtype=``. Soup still declares
+  ``transformers>=4.36.0,<5.0.0``, but the 12-cell matrix only ever installed the
+  newest 4.x, so a >=4.56-only kwarg stayed green. Call sites in chat / diff /
+  infer / export / merge / serve / mole routing / layer-stream runtime now pass
+  ``torch_dtype=`` (still accepted on current 4.57.x). A static AST guard fails
+  if a production ``AutoModel*`` load/config site reintroduces ``dtype=``. A new
+  Ubuntu/3.11 ``transformers-floor`` job installs under
+  ``.github/constraints/transformers-floor.txt`` using the lowest non-yanked
+  resolvable Transformers version ``4.46.1`` with the lowest version in the
+  declared TRL range ``0.14.0`` (``transformers==4.36.0`` is ResolutionImpossible
+  against declared ``trl`` — the declared Transformers floor in
+  ``pyproject.toml`` remains unchanged), runs ``pip check``, asserts both pins,
+  and runs the guard. The existing 12-cell matrix is untouched.
+
 - **`downsample` now returns at most `max_points` rows, which is what its
   docstring has always promised (#473 in #474).** The stride was
   `len(rows) // max_points` — a divisor, not a cap — so five rows with
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -458,6 +458,13 @@ Great areas for first contributions:
 GitHub Actions runs on every push and PR:
 - **ruff** linting on Python 3.11 (must pass)
 - **pytest** on Python 3.10, 3.11, 3.12 across Ubuntu, Windows, macOS (must pass)
+- **transformers-floor** on Ubuntu / Python 3.11 (#478): installs
+  `.[dev]` under `.github/constraints/transformers-floor.txt`, asserts the
+  pinned `transformers` version, and runs the static `dtype=` / `torch_dtype=`
+  load-site guard. The declared `[train]` floor remains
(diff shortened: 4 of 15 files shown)


MakazhanAlpamys:
**This PR is not a CI improvement. It fixes a live, shipped breakage, and I measured it rather than reasoning about it.** Two MEDIUMs to fix first; both are cheap.

## What I ran, and what it settles

Two independent reviews of this cluster disagreed on whether `dtype=` on an older transformers **raises** or is **silently ignored** — the difference between "half our declared range is broken right now" and "a quiet performance regression". Reading the source did not settle it, so I installed `transformers==4.46.1` (the real floor: 4.46.0 is yanked, and `trl 0.14.0` requires `>=4.46.0`) into an isolated target dir and ran it against a locally-built bf16 checkpoint — no network, no cache ambiguity:

```
transformers 4.46.1, bf16 checkpoint on disk

```

The mechanism is in `configuration_utils.py`: extra kwargs are consumed by the config **only** `if hasattr(config, key)`. There is no `dtype` attribute before 4.56, so it survives into the returned `model_kwargs` and reaches `cls(config, **model_kwargs)` — hence `TypeError` from the model's `__init__`, not a config attribute quietly set.

**So: `dtype=` is a hard failure at load, on every transformers from our floor up to 4.55.** `main` today has that spelling at production `from_pretrained` sites in `chat.py:237,247`, `serve.py:1338,1348,1446` and `merge.py:178,230` — which means `soup chat`, `soup serve` and `soup merge` are broken on the entire lower half of our own declared pin, for anyone on an older image (the Colab/Kaggle population #385/#387 is about). That is shipped, not hypothetical.

The third line also independently re-confirms #339's premise: with no kwarg at all, a bf16 checkpoint loads as **fp32**.

**Your guard would have caught it.** Reintroducing `dtype=` at one site turns `test_no_production_auto_model_load_uses_dtype_kwarg` red, and it does so in all 13 cells rather than only the new one — better than #478 asked for.

I also verified your floor reasoning from PyPI independently rather than accepting it: `trl 0.14.0` requires `transformers>=4.46.0`, and every file of `transformers 4.46.0` is yanked, so **4.46.1 is correct** and `4.36.0` genuinely cannot resolve. The constraints file explicitly refusing to relabel 4.55.4 as "the floor" is the right instinct.

## Two MEDIUMs before merge

**1. The floor is hardcoded in three places and derived from none.** `.github/constraints/transformers-floor.txt:22`, the workflow assertion, and `tests/test_transformers_floor_compat.py:145`. I changed `pyproject.toml` to `transformers>=4.52.0` and the floor suite stayed **green** — the module never reads `pyproject.toml`, so a future pin bump silently leaves the "floor" job testing an unrelated version. That is the same class of defect the job exists to prevent.

This repo already has the pattern: `test_requires_python_bound.py` derives the Python bound from the CI matrix rather than restating it. Please parse `transformers>=X` out of `pyproject.toml` and assert the constraint pin is `>= X`, while keeping the file's documented reason for not being *equal* to it. That way the honest "4.36 is unresolvable" exception survives, and a bump becomes loud.

**2. The scanner only matches `AutoModel*`, so a concrete model class is invisible to it.** `_is_auto_model_load` (`tests/test_transformers_floor_compat.py:47-54`). Putting `dtype="auto"` on the real `WhisperForConditionalGeneration.from_pretrained` in `src/soup_cli/trainer/asr.py:220` — the ASR training path — leaves the suite green.

Both live Whisper sites are clean today, so this is a gap rather than a bug, but it is exactly the kind of gap that lets the next instance through. The full inventory of load-site class names under `src/` is 18 distinct names, so widening to `*ForCausalLM | *ForConditionalGeneration | *ForSequenceClassification | AutoModel* | AutoProcessor` is enumerable rather than open-ended.

## Small

**3.** `utils/layer_stream_runtime.py:1343-1348` replaced a working `try: dtype= / except TypeError: torch_dtype=` capability probe with an unconditional `torch_dtype=`. Correct inside `<5.0.0` and simpler — but it was the one site following the "probe capabilities, not versions" pattern from #326, which CLAUDE.md holds up precisely because a version table was wrong twice. Practically inert today; noting it as a design decision rather than a defect.

**4.** CHANGELOG credit is `(#478)`; house format is `(#478 by @AchuthReddy-16 in #489)`.

**5.** The entry uses RST-style double backticks where every neighbour uses single.

**6.** Please reframe the CHANGELOG entry. As written it reads as a CI addition; on the evidence above it is a **user-facing bug fix** — `soup chat` / `serve` / `merge` were failing at model load on transformers 4.36–4.55. That is the headline, and users scanning the changelog for "why did my install break" should find it.

## Verified good

No test deleted or weakened — and both edited assertions were **inverted and strengthened**: `test_bugfixes.py` swapped its pair and added a substring-safe `source.replace("torch_dtype", "")` check, and `test_speculative_decoding.py` swapped the kwarg **and added** `assert "dtype" not in call_kwargs`. `ruff` clean, `test_cli_startup_is_light` 23/23, YAML parses, `PYTHONUTF8=1` set so the em-dash in the assertion is safe. One new cell. The scanner's own negative controls are good — it correctly ignores `FastLanguageModel`, `PeftModel`, `torch.zeros` and wrapper kwargs, and correctly catches `from_config`.

## What I could not verify

Whether `pip install -e ".[dev]" -c <constraints>` actually resolves and `pip check` passes on `ubuntu-latest`/3.11 — that needs a real network install; the PyPI metadata I checked is consistent with your claim. Note also that neither selected test imports `transformers` or `peft`, so the pinned install is verified as **resolvable** but never **executed** — which your notes should say plainly, since the current wording reads as if a floor cell exercises model-construction paths.

Fix 1 and 2 and this merges. Thank you for taking #478 — it turned out to be worth considerably more than the CI hygiene it was filed as.

MakazhanAlpamys:
**Merged** as `f424aa3`, CONTRIBUTORS.md entry in (`5463c18`). CI 13/13 at `a262df4b`, SHA verified against the head.

**I changed my mind about the merge order, and I owe you the reasoning.** I said I would land #492 first. Then I checked what your PR actually contains and found it fixes twelve production sites, not just the guard — so `soup chat`, `soup serve`, `soup merge`, `soup infer`, `soup diff` and `soup export` were failing at model load for anyone on transformers 4.36–4.55. Holding a shipped breakage behind polish on the guard was the wrong trade, so this went first.

## What I measured

Two independent reviews of this cluster disagreed on whether `dtype=` **raises** or is **silently ignored** on an older transformers — the difference between "half our declared range is broken right now" and "a quiet performance regression". Reading the source did not settle it, so I installed `transformers==4.46.1` into an isolated target dir and ran it against a locally-built bf16 checkpoint, no network:

```
```

`configuration_utils.py` consumes an extra kwarg only `if hasattr(config, key)`; there is no `dtype` attribute before 4.56, so it survives into `model_kwargs` and hits the model's `__init__`. **Your framing was right and the counter-argument was wrong.** The third line also re-confirms #339's premise for free.

**Your floor reasoning I verified against PyPI rather than accepting:** `trl 0.14.0` requires `transformers>=4.46.0`, and every file of 4.46.0 is yanked — so 4.46.1 really is the lowest installable version and 4.36.0 genuinely cannot resolve. Refusing to relabel 4.55.4 as "the floor", and writing down *why* the constraint is not equal to the declared pin, is the instinct I most want to see in this kind of PR. It would have been easy and quiet to pin something convenient.

**The design decision I want to name:** your static scan runs in **all thirteen cells**, not only the new floor job. That means the guard does not depend on the expensive cell being reached or staying green. Reintroducing `dtype=` at a single site turns `test_no_production_auto_model_load_uses_dtype_kwarg` red by name — I checked.

And no test was deleted or weakened: both edited assertions were **inverted and strengthened** — `test_bugfixes.py` gained a substring-safe `source.replace("torch_dtype", "")` check, and `test_speculative_decoding.py` added `assert "dtype" not in call_kwargs` on top of swapping the kwarg.

## My two MEDIUMs are now issues, not blockers

Both are about the guard's robustness, not the fix, so they did not justify holding this:

- **#494** — the floor version is hardcoded in three places and derived from none. Raising the pin in `pyproject.toml` leaves the floor suite green while CI keeps testing 4.46.1. Copy `test_requires_python_bound.py`'s pattern, keeping your documented 4.36-is-unresolvable exception.
- **#495** — the scanner matches only `AutoModel*`, so `dtype=` on `WhisperForConditionalGeneration.from_pretrained` in `trainer/asr.py:220` is invisible to it. Both live Whisper sites are clean today.

Take either if you want them; otherwise they are open for anyone.

## Two smaller notes

Please use the house credit format next time — `(#478 by @AchuthReddy-16 in #489)` rather than `(#478)`. Attribution is the only currency this project pays contributors in, and I would rather you claim it than have me reconstruct it at release time.

And `utils/layer_stream_runtime.py` traded a working `try: dtype= / except TypeError: torch_dtype=` capability probe for an unconditional `torch_dtype=`. Correct inside `<5.0.0` and simpler — but that was the one site following the "probe capabilities, not versions" pattern from #326, which exists because a version table was wrong twice. A design note, not a defect, and worth remembering when transformers 5 removes `torch_dtype=` too.

Filed as CI hygiene, delivered as a fix for a live breakage nobody had reported. Thank you.

PR #490, opened 2026-08-22
chore(changelog): add conflict-free fragments

### LINKED ISSUE #487: Every merge conflicts every open PR on CHANGELOG.md, because entries are inserted at the top of [Unreleased]
Observed three times in one afternoon: #479 merged and flipped #481 and #482 from `MERGEABLE/CLEAN` to `CONFLICTING/DIRTY`; I resolved both by hand; #481 then merged and conflicted #482 again on the same lines.

None of it was a contributor error, and none of it was a real content disagreement.

## Mechanism

Every PR adds its entry at the **top** of `## [Unreleased]` / `### Added`. Two PRs that touch entirely unrelated subsystems therefore edit the same line range, and git has no way to know both entries should simply coexist. The resolution is always identical: keep both, in either order.

So the conflict rate is a function of *how many PRs are open*, not of whether they overlap. With 7 open PRs, one merge can conflict six of them — and each contributor then has to rebase for a reason that has nothing to do with their change.

## Why it is worth fixing rather than absorbing

- It punishes contributors for someone else's merge, which is exactly the kind of friction that makes a first-time contributor give up.
- It makes `MERGEABLE/CLEAN` meaningless as a queue signal — everything downstream of the next merge is about to be dirty.
- The maintainer either resolves it by hand (I did, twice, and the second was undone within the hour) or pushes the cost onto people who did nothing wrong.
- It costs a CI re-run each time, and every push from a first-time contributor needs manual approval, so it also costs maintainer attention.

## Options

1. **Append to the end of each `###` subsection instead of the top.** Smallest possible change, no tooling. Two PRs still collide if both append, but the conflict window is one line rather than a shared block, and git resolves many of them automatically.
2. **A `.gitattributes` union merge driver for `CHANGELOG.md`** (`CHANGELOG.md merge=union`). Git keeps both sides automatically. Cheap, and the failure mode is a duplicated line rather than lost content — but it can interleave two entries mid-paragraph, and this project's entries are long multi-paragraph prose, so this needs testing before adoption.
3. **Changelog fragments** (`changelog.d/<pr>.md`, assembled at release). Eliminates the conflict entirely because each PR writes its own file. Costs a release-time assembly step and a docs change; towncrier is the usual tool.

I lean toward **1 first** (zero infrastructure, immediate relief) and evaluating **3** at the next release, where the assembly step is already a manual pass. **2** is tempting but the union driver's behaviour on this file's long entries should be demonstrated on a real pair of conflicting PRs before it is trusted.

## Acceptance

- [ ] A decision recorded in `CONTRIBUTING.md` so contributors know where to put a new entry
- [ ] If option 1: `.claude/CLAUDE.md`'s release checklist updated to match, since it currently describes editing `[Unreleased]` without specifying position
- [ ] Demonstrated on two branches that both add an entry, showing the conflict is gone or reduced to a single auto-resolvable line
- [ ] If option 2 or 3: the mechanism is documented in `CONTRIBUTING.md` and does not silently drop an entry when two PRs land in the same release

Not urgent, and no code changes to `src/`. Good task for someone who wants to improve the contribution experience rather than the tool itself.

Amix29:
## Summary

- replace shared edits to `CHANGELOG.md` with version-scoped fragments at `changelog.d/<latest-release>/<number>.<category>.md`
- add a standard-library-only assembler that validates every file, preserves fragment Markdown verbatim, and consumes fragments after assembly
- make forgotten assembly loud both in normal CI (stale baselines fail after a release) and in the PyPI tag workflow (any remaining fragment blocks publication)
- document the contributor and release workflows without migrating the existing `[Unreleased]` content

The numeric key may be the issue or pull-request number. GitHub assigns both from one repository-wide namespace, so the filename remains collision-free even when a fragment is written before its PR exists; a second fragment with the same number is rejected.

## Safety properties pinned by tests

- a real two-branch Git merge keeps both independently named fragments without a conflict
- a fragment below an older release version fails before either the changelog or fragment is mutated
- unknown layouts, malformed names, empty/non-list content, mismatched references, and duplicate numeric keys fail rather than being ignored
- long multi-paragraph entries, including tables and fenced code blocks, survive assembly byte-for-byte as a substring
- an interrupted assembly can be rerun without duplicating an entry
- the release workflow refuses to build while any fragment remains

No runtime dependency is added; the assembler uses only the Python standard library.

## Validation

- `PYTHONPATH=src python -m pytest tests/ -q` — 18,304 passed, 285 skipped, 5 deselected; 81.24% coverage
- `python -m pytest tests/test_issue487_changelog_fragments.py -q --no-cov` — 16 passed
- `actionlint` v1.7.12
- `python -m build` — sdist and wheel built successfully

Closes #487

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,6 +12,7 @@
 
 ## Checklist
 
-- [ ] `ruff check src/soup_cli/ tests/` passes
+- [ ] `ruff check src/soup_cli/ scripts/ tests/` passes
 - [ ] `pytest tests/ -v` passes
 - [ ] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed
+- [ ] Added a changelog fragment, or this change has no user-visible impact
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           python-version: "3.11"
       - run: pip install ruff
-      - run: ruff check src/soup_cli/ tests/
+      - run: ruff check src/soup_cli/ scripts/ tests/
       - name: Lint the workflows themselves
         # `check-yaml` and a YAML parser both accept a workflow GitHub will
         # reject: 195d60b used the `runner` context in a job-level `env:` block,
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,5 +17,6 @@ jobs:
         with:
           python-version: "3.11"
       - run: pip install build
+      - run: python scripts/assemble_changelog.py --check-empty
       - run: python -m build
       - uses: pypa/gh-action-pypi-publish@release/v1
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,7 +49,7 @@ pytest tests/ -v --tb=short
 Run the linter:
 
 ```bash
-ruff check src/soup_cli/ tests/
+ruff check src/soup_cli/ scripts/ tests/
 ```
 
 If both pass — you're ready to contribute!
@@ -60,10 +60,10 @@ We use **ruff** for all code style and linting. Before committing, run:
 
 ```bash
 # Check for issues
-ruff check src/soup_cli/ tests/
+ruff check src/soup_cli/ scripts/ tests/
 
 # Auto-fix issues
-ruff check --fix src/soup_cli/ tests/
+ruff check --fix src/soup_cli/ scripts/ tests/
 ```
 
 ### Style Guidelines
@@ -337,13 +337,19 @@ git checkout -b fix/your-bug-fix
 - **Write tests first** (TDD) — then implement to pass them
 - Keep commits focused and logical
 
+For each user-visible pull request, add its entry to
+`changelog.d/<latest-release>/<number>.<category>.md` instead of editing `CHANGELOG.md`;
+use the newest released version in `CHANGELOG.md`, prefer the PR number rather than the
+issue it closes, and choose `added`, `changed`, `deprecated`, `removed`, `fixed`, or
+`security`.
+
 ### 3. Run Tests & Lint
 
 Before pushing, ensure everything passes:
 
 ```bash
 # Lint first
-ruff check --fix src/soup_cli/ tests/
+ruff check --fix src/soup_cli/ scripts/ tests/
 
 # Then run tests
 pytest tests/ -v --tb=short
@@ -378,7 +384,7 @@ Then open a pull request on GitHub with:
 
 When you open a PR, the GitHub template will show this checklist:
 
-- [ ] `ruff check src/soup_cli/ tests/` passes
+- [ ] `ruff check src/soup_cli/ scripts/ tests/` passes
 - [ ] `pytest tests/ -v` passes
 - [ ] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed
 - [ ] New tests added for new functionality
@@ -467,12 +473,14 @@ The project follows semantic versioning: `MAJOR.MINOR.PATCH`
 
 ### Version Bump Process
 
-1. Update version in `pyproject.toml` and `src/soup_cli/__init__.py`
-2. Run full test suite and linting
-3. Update `README.md`, the relevant page under `docs/`, `CHANGELOG.md`, and `SECURITY.md` (if security-related)
-4. Commit with message: `Release v0.X.0`
-5. Tag: `git tag v0.X.0 && git push --tags`
-6. GitHub Actions auto-publishes to PyPI on the `v*` tag (Trusted Publisher / OIDC)
+1. Run `python scripts/assemble_changelog.py` and review the assembled `CHANGELOG.md`
+2. Update version in `pyproject.toml` and `src/soup_cli/__init__.py`
+3. Run full test suite and linting
+4. Update `README.md`, the relevant page under `docs/`, and `SECURITY.md` (if security-related)
+5. Commit with message: `Release v0.X.0`
+6. Tag: `git tag v0.X.0 && git push --tags`
+7. GitHub Actions verifies that no fragment was left behind, then publishes to PyPI
+   on the `v*` tag (Trusted Publisher / OIDC)
 
 ## Recognition
 
--- a/changelog.d/0.73.3/490.changed.md
+++ b/changelog.d/0.73.3/490.changed.md
@@ -1,0 +1,6 @@
+- **User-visible changes now use per-PR changelog fragments (#487 by @Amix29 in #490).**
+  Contributors add a uniquely named, version-scoped Markdown file instead of editing the
+  shared `[Unreleased]` section. A standard-library assembler preserves long entries
+  verbatim, rejects stale or malformed fragments, and consumes them during release
+  preparation. A tag-time release gate refuses publication if any fragment remains, so
+  the conflict is removed without making changelog loss silent.
--- a/changelog.d/README.md
+++ b/changelog.d/README.md
@@ -1,0 +1,33 @@
+# Changelog fragments
+
+User-visible pull requests add one Markdown fragment instead of editing the shared
+`CHANGELOG.md` file:
+
+```text
+changelog.d/<latest-release>/<number>.<category>.md
+```
+
+`<latest-release>` is the newest released version heading in `CHANGELOG.md`. `<number>`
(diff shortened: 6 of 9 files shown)


Amix29:
Windows 3.12 exposed a real CRLF portability bug in the interrupted-run recovery path.

The assembler previously allowed `\r` to become part of a parsed section title, so it failed to recognize an existing `### Fixed` section and could append the same fragment twice after an interrupted run. The updated commit now:

- parses release and section headings safely with both LF and CRLF;
- compares fragment content with newline-normalized text for idempotency while preserving the changelog’s original bytes and the fragment Markdown;
- uses a regression fixture that explicitly combines a CRLF changelog with an LF fragment, so this behavior is exercised on every platform.

Local focused validation: `16 passed`; full Ruff validation: clean. The complete cross-platform suite is running again on the amended one-commit branch.

Amix29:
Follow-up: the complete rerun is green. All macOS, Ubuntu, and Windows jobs pass on Python 3.10, 3.11, and 3.12, including the Windows 3.12 job that originally exposed the CRLF bug. Lint, type-check, and the MLX smoke test pass as well.

MakazhanAlpamys:
**All five constraints met, and constraint 1 — the one that decides whether this is better or worse than what we have — is met with a real test rather than a checklist line.** Merging on nits.

## Verified, not read

I rebuilt your empirical claim in a scratch repo rather than trusting the PR's own tests:

| case | result |
|---|---|
| two branches each add a fragment, then merge | **no conflict**, both files retained |
| two branches each edit the **same existing** `CHANGELOG.md` entry | `CONFLICT (content)` — **still loud** |

That second row is the property that made fragments the right answer over `merge=union`, and it holds: genuine disagreements keep conflicting.

**Constraint 1.** `validate_fragments()` rejects any fragment whose directory is not the newest release heading, and `test_repository_fragments_match_the_newest_release` runs it against the **real repo root**, so it fires in all 12 matrix cells. I dropped a stale `changelog.d/0.73.2/999.fixed.md` into the checkout and that named test went red with `fragment baseline is 0.73.2, but the latest release is 0.73.3`. Second independent gate: `publish.yml` runs `--check-empty` before `python -m build`, so a tag push with anything left over fails before PyPI. That is exactly the "no fragment predates the newest release heading" test I asked for, and a second belt besides.

**Constraint 5**, which I expected to be the weak one: I ran the assembler against a copy of the **real** `CHANGELOG.md` with a fragment containing a table, a fenced Python block with internal double-spacing, a trailing-two-space line break, nested bullets and an inline link. `fragment_bytes in output_bytes == True`. No reflow, no escaping, no re-wrap.

Constraints 2, 3, 4 all hold: keyed on the repo number (collisions are loud either as add/add or via `#N already has fragment`), stdlib only — no towncrier, 338 lines we own — and `CONTRIBUTING.md:340-344` says where an entry goes in one paragraph whose six categories match `SECTIONS` in the validator.

**Out of scope respected.** `CHANGELOG.md` is untouched and `[Unreleased]` is not migrated, so no open PR is conflicted by this landing. That was the constraint most likely to be overrun by enthusiasm, and you left it alone.

Five mutations, five killed — including reflowing the payload (`test_assembly_preserves_long_markdown_verbatim`), no-oping the release gate, and accepting a duplicate number.

## Nits

**1.** On a CRLF checkout the assembler adds a third newline: `_separate_before` / `_separate_after` test `endswith("\n\n")`, which is `False` for `"\r\n\r\n"`. On LF — the committed form and Linux CI — spacing is exactly right. My working tree is CRLF so I hit it; `core.autocrlf` normalises at `git add`, leaving only a cosmetic extra blank line.

**2.** `changelog.d/README.md` says "the issue or pull-request number". Two PRs closing the same issue that both pick the *issue* number collide. Recommending the **PR** number removes the ambiguity.

**3.** Your own fragment reads `(#487 by @Amix29)`; house format is `(#487 by @Amix29 in #490)`. Worth fixing precisely because this file is the template every future contributor will copy.

**4.** `test_repository_fragments_match_the_newest_release`'s assert body is vacuous — the real signal is `validate_fragments` raising, which I confirmed works. A one-line comment saying "the call is the assertion" would stop someone 'fixing' it later.

## Merge order

I am landing #489 and #492 first, then this. Both were opened against the old convention, and putting the new one in last means neither has to convert an entry mid-review. Once this lands, new PRs use `changelog.d/`. Nothing for you to do — I trial-merged this against #489 and it is clean.

One follow-up that is mine, not yours: `.claude/CLAUDE.md` still documents `ruff check src/soup_cli/ tests/` while CI, CONTRIBUTING and the PR template now say `src/soup_cli/ scripts/ tests/`. That file is gitignored, so no PR can touch it — I will fix it locally.

## What this was worth

You reproduced all three options before implementing, showed my recommended option 1 does not actually remove the conflict, found the real risk in option 2, and caught that my acceptance criteria asked you to edit a file no public PR can reach. Then you asked which direction rather than writing the wrong PR. That sequence is why this landed in one round instead of three.

Amix29:
Addressed all four nits in `0a7ce85`:

1. Separator detection is now newline-normalized and emits the surrounding content’s line ending, so a CRLF changelog no longer gains a cosmetic third newline. A host-independent regression test writes an explicit CRLF changelog, assembles an LF fragment, and pins the exact normalized spacing plus preservation of the existing CRLF bytes.
2. `changelog.d/README.md` and `CONTRIBUTING.md` now recommend the PR number rather than the issue number, avoiding collisions when multiple PRs close one issue.
3. This PR’s fragment is renamed to `490.changed.md` and uses the house form `(#487 by @Amix29 in #490)`.
4. The repository-fragment test now comments explicitly that the validator call is the assertion because invalid fragments raise there.

Focused validation: `17 passed`; full Ruff validation and `git diff --check`: clean. The full matrix is running again.

Amix29:
Follow-up: the amended commit is green across the complete matrix — macOS, Ubuntu, and Windows on Python 3.10, 3.11, and 3.12, plus lint, type-check, and MLX smoke. The CRLF spacing regression test passes in every test cell.

MakazhanAlpamys:
**Merged** as `2ccdebf`, fifth CONTRIBUTORS.md bullet in (`45b72c5`). All four nits re-verified by mutation, not by reading your summary.

| Mutation | Result |
|---|---|
| `if baseline != latest_release:` → `if False:` | `test_stale_fragment_fails_loudly_without_mutating_files` |
| `payload += content` → reflow the payload | `test_assembly_preserves_long_markdown_verbatim` |
| `normalised = _normalise_newlines(prefix)` → `normalised = prefix` (restore the CRLF bug) | `test_crlf_assembly_does_not_add_an_extra_blank_line` |

Baseline **17 passed**; merged against current `main`, **28 passed** with no conflict — including the new `transformers-floor` guard that landed in #489 while your PR was open.

Worth mentioning because it is the trap this whole exercise is about: my first attempt at the CRLF mutation **did not apply** — I had the helper's name wrong — and the suite reported a clean 17 passed. An unapplied mutation is indistinguishable from a surviving one, and I only caught it because I assert the anchor count before writing. Second attempt landed and the test fired.

## What this was actually worth

You reproduced all three options from one base commit **before implementing**, and the reproduction changed the answer. Option 1 was *my* recommendation and does not work — appending at the end of the section still conflicts. Option 2 merges two new entries cleanly but silently fuses two different corrections to one existing entry into a duplicated bullet with contradictory text, which is a loud conflict traded for a quiet corruption of the project's public record.

You also caught that my acceptance criteria asked you to edit `.claude/CLAUDE.md`, which is gitignored and unreachable from any public PR. I wrote a criterion that could not be satisfied and you said so instead of working around it.

And then you asked which direction to take rather than writing the wrong PR. That sequence is why this landed in one round.

The design detail I want to name: **a lost fragment is loud.** `validate_fragments` runs against the real repository root in every matrix cell, so it does not depend on anyone remembering to run the assembler, and `--check-empty` in `publish.yml` is a second gate before PyPI. A merged PR whose entry never reaches a release would have made fragments *worse* than what we had — that was constraint 1, and you met it with a test rather than a checklist line. This project has documented (v0.73.3) that prose alone failed as a control four separate times.

Keying on the PR number rather than the issue was the right call on the ambiguity too.

## Follow-up that is mine

`.claude/CLAUDE.md` still documents `ruff check src/soup_cli/ tests/` while CI, CONTRIBUTING and the PR template now say `src/soup_cli/ scripts/ tests/`. Gitignored, so I will fix it locally — recording it here so it is not silently dropped.

Five merged PRs, and this one changed how every future contributor writes a changelog entry. Thank you.

PR #492, opened 2026-08-23
fix(trainer): resolve frozen-base load dtype in the twelve non-SFT trainers

### LINKED ISSUE #491: fix(trainer): the twelve non-SFT trainers still load frozen bases as fp32 (follow-on to #339)
## What happened

Issue #339 established that `from_pretrained()` gets no `dtype`/`torch_dtype` kwarg, so a bf16 checkpoint silently loads as fp32 (up to 2.6x extra VRAM, per the maintainer's own H100 measurement). PR #471 fixes this for `SFTTrainerWrapper` (all three modalities), but the same `model_kwargs` shape, missing the same key, exists unchanged in the other twelve trainer wrappers:

`dpo.py`, `kto.py`, `orpo.py`, `simpo.py`, `ipo.py`, `bco.py`, `online_dpo.py`, `grpo.py`, `ppo.py`, `pretrain.py`, `reward_model.py`, `embedding.py`.

None of these trainers has a full-fine-tuning branch (`get_peft_model`/`peft_config` runs unconditionally in every one of them; `unfrozen_parameters` and `lisa_enabled` are schema-gated to `task='sft'`, and `lora.r=0` has no wired effect outside `sft`), so every load in these twelve tasks is a frozen LoRA base. Confirmed by reading all twelve `model_kwargs` blocks: `trust_remote_code` + `device_map` + optional `quantization_config`, no dtype key, no post-load `.to(dtype)` cast.

## Reproduction

```
grep -n "model_kwargs = {" -A3 src/soup_cli/trainer/dpo.py
```

shows no `dtype`/`torch_dtype` key. The same block, same gap, repeats identically in the other 11 files above.

## Expected

A frozen-base load in these twelve trainers keeps the checkpoint's own dtype instead of defaulting to fp32, the same fix #471 ships for the SFT trainer.

AmirF194:
Follows #471, which fixes the missing `dtype`/`torch_dtype` kwarg (#339) for `SFTTrainerWrapper`. The same `model_kwargs` shape, missing the same key, exists unchanged in the other twelve trainer wrappers: `dpo`, `kto`, `orpo`, `simpo`, `ipo`, `bco`, `online_dpo`, `grpo`, `ppo`, `pretrain`, `reward_model`, `embedding`.

None of these trainers has a full-fine-tuning branch: `unfrozen_parameters`/`lisa_enabled` are schema-gated to `task='sft'`, and `lora.r=0` has no wired effect outside `sft` (peft would just reject `r=0`), so `get_peft_model`/`peft_config` runs unconditionally in every one of them. Every load in these twelve tasks is therefore a frozen LoRA base, so the fix is the simpler half of #471's split: keep the checkpoint's own dtype (`torch_dtype="auto"`), except on a pre-Ampere CUDA card (T4/P100/V100/GTX 16xx/RTX 20xx), where `bf16_fp16_flags` already routes training compute to float16 and an `"auto"` load would leave the resident weights in bf16 storage against float16 compute.

Added `resolve_frozen_base_load_dtype(device)` to `soup_cli.utils.gpu`, next to `bf16_fp16_flags`, and wired it into the `model_kwargs` of all twelve wrappers.

## Verification

- New tests: unit coverage of `resolve_frozen_base_load_dtype` (CPU, bf16-capable CUDA, pre-Ampere CUDA, no CUDA runtime) plus one test per wrapper asserting the resolved `torch_dtype` reaches the real `from_pretrained` call.
- `pytest tests/test_issue_frozen_base_load_dtype.py -v`: 16 passed on this branch, 16 failed on `main` (the assertion on the missing `torch_dtype` kwarg).
- `pytest tests/test_bco.py tests/test_kto.py tests/test_orpo.py tests/test_simpo.py tests/test_ipo.py tests/test_grpo.py tests/test_ppo.py tests/test_pretrain.py tests/test_embedding.py tests/test_performance.py tests/test_gpu.py tests/test_issue385_stream_dtype.py`: 512 passed, 4 skipped (GPU-only), no regressions.
- Not verified: actual VRAM savings on real hardware (no GPU in this environment); this mirrors #471's own measurement rather than re-measuring it. Also not run: the `[dev]` extra's `trl`/`bitsandbytes`/`deepspeed`, or the macOS/Windows CI legs.

Fixes #491

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -201,7 +201,7 @@ reproducing 70+ versions of notes.
 - **The twelve non-SFT trainers (`dpo`, `kto`, `orpo`, `simpo`, `ipo`, `bco`,
   `online_dpo`, `grpo`, `ppo`, `pretrain`, `reward_model`, `embedding`) loaded
   a frozen LoRA base as float32 regardless of the checkpoint's own dtype
-  (#491 by @AmirF194).** #471 fixed this for the SFT trainer; the
+  (#491 by @AmirF194 in #492).** #471 fixed this for the SFT trainer; the
   same `model_kwargs` shape, missing the same key, was unchanged in the
   other twelve. None of them has a full fine-tuning branch, so every load
   there is a frozen base: the new shared `resolve_frozen_base_load_dtype()`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -198,6 +198,17 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **The twelve non-SFT trainers (`dpo`, `kto`, `orpo`, `simpo`, `ipo`, `bco`,
+  `online_dpo`, `grpo`, `ppo`, `pretrain`, `reward_model`, `embedding`) loaded
+  a frozen LoRA base as float32 regardless of the checkpoint's own dtype
+  (#491 by @AmirF194).** #471 fixed this for the SFT trainer; the
+  same `model_kwargs` shape, missing the same key, was unchanged in the
+  other twelve. None of them has a full fine-tuning branch, so every load
+  there is a frozen base: the new shared `resolve_frozen_base_load_dtype()`
+  keeps the checkpoint's own dtype (`torch_dtype="auto"`), except on a
+  pre-Ampere CUDA card, where it now matches the float16 compute dtype
+  those cards already use instead of leaving the base in bf16 storage.
+
 - **Layer streaming now keeps its host store on CPU on Apple Silicon
   (#434 by @Amix29 in #480).**
   PyTorch 2.7+ can return an MPS tensor for
--- a/src/soup_cli/trainer/bco.py
+++ b/src/soup_cli/trainer/bco.py
@@ -23,6 +23,7 @@ from soup_cli.utils.gpu import (
     estimate_batch_size,
     model_size_from_name,
     resolve_device_map,
+    resolve_frozen_base_load_dtype,
 )
 from soup_cli.utils.seeding import apply_training_seed, training_seed_kwargs
 
@@ -263,6 +264,7 @@ class BCOTrainerWrapper:
         dev_map = resolve_device_map(self.device)
         model_kwargs = {
             "trust_remote_code": self._trust_remote_code, "device_map": dev_map,
+            "torch_dtype": resolve_frozen_base_load_dtype(self.device),
         }
         if quant_config_obj is not None:
             model_kwargs["quantization_config"] = quant_config_obj
--- a/src/soup_cli/trainer/dpo.py
+++ b/src/soup_cli/trainer/dpo.py
@@ -13,6 +13,7 @@ from soup_cli.utils.gpu import (
     estimate_batch_size,
     model_size_from_name,
     resolve_device_map,
+    resolve_frozen_base_load_dtype,
 )
 from soup_cli.utils.seeding import apply_training_seed, training_seed_kwargs
 
@@ -253,6 +254,7 @@ class DPOTrainerWrapper(StreamingSetupMixin):
         dev_map = resolve_device_map(self.device)
         model_kwargs = {
             "trust_remote_code": self._trust_remote_code, "device_map": dev_map,
+            "torch_dtype": resolve_frozen_base_load_dtype(self.device),
         }
         if quant_config_obj is not None:
             model_kwargs["quantization_config"] = quant_config_obj
--- a/src/soup_cli/trainer/embedding.py
+++ b/src/soup_cli/trainer/embedding.py
@@ -13,6 +13,7 @@ from soup_cli.utils.gpu import (
     estimate_batch_size,
     model_size_from_name,
     resolve_device_map,
+    resolve_frozen_base_load_dtype,
 )
 from soup_cli.utils.seeding import apply_training_seed, training_seed_kwargs
 
@@ -229,6 +230,7 @@ class EmbeddingTrainerWrapper:
         dev_map = resolve_device_map(self.device)
         model_kwargs = {
             "trust_remote_code": self._trust_remote_code, "device_map": dev_map,
+            "torch_dtype": resolve_frozen_base_load_dtype(self.device),
         }
         if quant_config_obj is not None:
             model_kwargs["quantization_config"] = quant_config_obj
--- a/src/soup_cli/trainer/grpo.py
+++ b/src/soup_cli/trainer/grpo.py
@@ -16,6 +16,7 @@ from soup_cli.utils.gpu import (
     estimate_batch_size,
     model_size_from_name,
     resolve_device_map,
+    resolve_frozen_base_load_dtype,
 )
 from soup_cli.utils.seeding import apply_training_seed, training_seed_kwargs
 
@@ -517,6 +518,7 @@ class GRPOTrainerWrapper:
         model_kwargs = {
             "trust_remote_code": self._trust_remote_code,
             "device_map": dev_map,
+            "torch_dtype": resolve_frozen_base_load_dtype(self.device),
         }
         if quant_config_obj is not None:
             model_kwargs["quantization_config"] = quant_config_obj
--- a/src/soup_cli/trainer/ipo.py
+++ b/src/soup_cli/trainer/ipo.py
@@ -13,6 +13,7 @@ from soup_cli.utils.gpu import (
     estimate_batch_size,
     model_size_from_name,
     resolve_device_map,
+    resolve_frozen_base_load_dtype,
 )
 from soup_cli.utils.seeding import apply_training_seed, training_seed_kwargs
 
@@ -210,6 +211,7 @@ class IPOTrainerWrapper:
         dev_map = resolve_device_map(self.device)
         model_kwargs = {
             "trust_remote_code": self._trust_remote_code, "device_map": dev_map,
+            "torch_dtype": resolve_frozen_base_load_dtype(self.device),
         }
         if quant_config_obj is not None:
             model_kwargs["quantization_config"] = quant_config_obj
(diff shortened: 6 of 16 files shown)


MakazhanAlpamys:
**The spelling is right, all twelve trainers are individually covered, and the pre-Ampere trap is actively tested rather than assumed.** One numerics finding to fix, plus a merge-order item I will handle.

## You are on the correct side of the kwarg split, and I measured what that means

All 13 new sites use `torch_dtype=`. I checked what the alternative would have cost by installing `transformers==4.46.1` — the real floor, since 4.46.0 is yanked and `trl 0.14.0` requires `>=4.46.0` — into an isolated dir and running it against a locally-built bf16 checkpoint:

```
transformers 4.46.1

```

So `torch_dtype=` works across the whole pin, `dtype=` is a **hard failure** below 4.56, and the third line independently re-confirms #339's premise. Worth stating because the reasoning in review went the other way — the expectation was that an unknown kwarg would be absorbed into the config as a silent no-op. It is not: `configuration_utils.py` consumes extra kwargs only `if hasattr(config, key)`, and there is no `dtype` attribute before 4.56, so it survives into `model_kwargs` and hits the model constructor.

## All twelve are individually proven

Deleting the `torch_dtype` entry from any single trainer kills exactly that trainer's named parametrized case — twelve for twelve, none riding on a sibling. Returning `None` from the resolver kills 15 of 16. Baseline **441 passed** across the PR's file plus ten trainer suites, **576 passed** across eighteen broader affected suites, `ruff` clean, `test_cli_startup_is_light` 34/34 (the resolver's `import torch` stays inside the function body). Diff is `+225/-0` as claimed — zero deletion lines anywhere.

**The pre-Ampere branch is not decorative, which is the part I most expected to be weak.** Your `_FakeCuda.is_bf16_supported()` returns `self._bf16 or self._emulated`, so reintroducing the v0.73.1 defect — the bare `is_bf16_supported()` that defaults to `including_emulation=True` and makes a T4 answer **True** — turns `test_pre_ampere_card_resolves_to_explicit_float16` red. That is the exact trap whose *first* fix was a no-op on the hardware it targeted, and you built a fixture that catches its return.

Your premise also holds: I checked all twelve, and every one calls `get_peft_model`/`peft_config` unconditionally with no full-FT branch, so "every load here is a frozen base" is accurate.

## MEDIUM — the reward head silently drops to bf16

`src/soup_cli/trainer/reward_model.py:215`. The premise is right about the *base*, but `:235` builds LoRA with `task_type=TaskType.SEQ_CLS`, and peft then auto-adds `modules_to_save=["score"]`. peft's `cast_adapter_dtype` upcasts only `BaseTunerLayer` submodules — **not** `ModulesToSaveWrapper`. Measured here with the installed peft:

```
base: torch.bfloat16   score: torch.bfloat16
TRAINABLE ...q_proj.lora_A.default.weight              torch.float32   <- peft autocasts
TRAINABLE ...q_proj.lora_B.default.weight              torch.float32
TRAINABLE ...score.modules_to_save.default.weight      torch.bfloat16  <- NOT autocast
```

Before this PR the base loaded fp32, so the randomly-initialised reward head was an fp32 trainable parameter. After it, on a bf16 checkpoint with `quantization: none`, **the head trains in pure bf16 — 8 mantissa bits, no fp32 master weights.** That is the one parameter whose output *is* the reward signal.

Scoped precisely: unreachable on the quantized path, because `:224` calls `prepare_model_for_kbit_training`, which upcasts. `ppo` (`CAUSAL_LM`) and `embedding` (`FEATURE_EXTRACTION`) get no auto `modules_to_save`, so this is `reward_model` + `quantization: none` only.

Three lines after `get_peft_model` — upcast non-adapter trainables to fp32, mirroring what the kbit path already does — plus a test asserting the head's dtype. If you would rather accept it, say so and document it; what I do not want is for it to land unremarked.

## Two smaller ones

**LOW — the wiring tests only run at `device="cpu"`, so a trainer that bypasses the resolver survives.** Every case constructs `wrapper_cls(cfg, device="cpu")` and asserts the literal `torch_dtype == "auto"`. I replaced `resolve_frozen_base_load_dtype(self.device)` in `dpo.py` with a hardcoded `"auto"` and the suite stayed green. On a pre-Ampere card that trainer would load bf16 storage against fp16 compute — exactly the case the resolver exists for. Patch the resolver in the wrapper's own module namespace with a sentinel and assert the sentinel reached `from_pretrained`: that proves the *call*, not the *value*.

**LOW — no CPU branch.** `bf16_fp16_flags("cpu")` gives `(False, False)`, so `--device cpu` now trains against a bf16 base. Harmless in practice (CPU use here is smoke tests), but the docstring's rationale is a VRAM argument that does not apply on CPU. One sentence.

**NIT** — `tests/test_issue_frozen_base_load_dtype.py` should be `test_issue491_frozen_base_load_dtype.py`; every sibling is `test_issue<NNN>_<slug>.py`.

**NIT** — no docs update. #471 updates three docs pages with prose general enough to cover both, so once both land the docs are accurate — but nothing tells a dpo/kto/grpo user their VRAM changed. One sentence.

I checked whether `hardware_fit` needs a matching change, since #471 touches it: it does not. `_BYTES_PER_PARAM_BY_QUANT["none"]` is already `2.0`, so the pre-flight was *already* assuming a bf16 load and `main`'s fp32 load was the thing being under-predicted. Your PR makes the predictor correct for these twelve.

## Merge order — mine to sort out, not yours

#471 (@blackcoderx, still open) fixes the SFT path with a byte-for-byte identical decision inlined in `trainer/sft.py`, passing it as `dtype=` at three sites, and does not touch `gpu.py`. Landed as-is alongside yours that is duplicated truth **plus** a split spelling — the #372 / #392 / #424 pattern this project keeps getting bitten by, and on the evidence above their spelling is also the broken one.

So: **yours lands first** — it owns `gpu.py` and is on the correct spelling — and I will ask #471 to rebase onto `resolve_frozen_base_load_dtype`, adding only its full-fine-tuning branch on top. Your helper's name is already right for that split, since it deliberately has no full-FT branch. Nothing for you to change; I am recording it here so the ordering is visible to both of you rather than decided in silence.

## What I could not verify

No pre-Ampere silicon here (RTX 3050, sm_86), so the T4/V100 branch is proven only through your `_FakeCuda` fixture. No multi-GPU, so `device_map` sharding against `torch_dtype="auto"` is untested. And I did not quantify what finding 1 costs in training quality — I proved the head is bf16, not how much accuracy that loses.

Fix the reward head and this merges.

MakazhanAlpamys:
**Merged** as `5c20a41`, CONTRIBUTORS.md entry in (`c749e83`). CI 13/13 at `3015ab0`, SHA verified against the head.

## The reward-head fix, re-verified by measurement

I ran the same measurement that produced the finding, not your summary:

```
before:  score.modules_to_save.default.weight   torch.bfloat16
after:   score.modules_to_save.default.weight   torch.float32
         lora_A / lora_B                        torch.float32  (untouched)
         base                                   torch.bfloat16 ← still frozen
```

**That last line is the one that mattered.** A naive `model.to(torch.float32)` would have upcast the frozen base too and destroyed the entire point of your PR — the VRAM saving it exists for. Scoping the loop to `requires_grad` fixes the head without undoing your own work.

And your test pins **both halves** of the property, which is why both mutations die on it:

| Mutation | What it breaks |
|---|---|
| remove the upcast | head stays bf16 |
| drop `requires_grad` from the condition | base drags to fp32 |

Both → `test_score_head_is_fp32_even_when_base_resolves_to_bf16`. A one-sided assertion would have caught only the first.

The line I want to single out is this one:

```python
# The base itself really did load bf16 (proves the fixture reproduces
# the bug's precondition, not just that everything is fp32 already).
assert base_dtype == torch.bfloat16
```

That is a **control** — it excludes the case where the test passes trivially because nothing was ever bf16. I have spent this week asking other contributors for exactly that, and you wrote it unprompted.

## What earned the merge before that

All twelve trainers individually proven: deleting the entry from any single one kills exactly that trainer's named case, twelve for twelve, none riding on a sibling. The pre-Ampere branch is genuinely guarded — your `_FakeCuda.is_bf16_supported()` returns `bf16 or emulated`, so reintroducing the v0.73.1 defect (the bare call that makes a T4 answer `True`) turns that test red. That is the trap whose *first* fix was a no-op on the hardware it targeted, and you built a fixture that catches its return.

You also checked the premise rather than assuming it — confirming all twelve apply PEFT unconditionally with no full-FT branch — and you put the resolver in `utils/gpu.py` so there would be one implementation rather than two.

**And you were on the right side of the kwarg split.** I measured what the alternative costs: on transformers 4.46.1, `dtype=` raises `TypeError` at model load. Twelve production sites had that spelling and shipped broken; #489 has since fixed them and added a guard that fails CI if it recurs.

On the CPU branch you took the honest option — documenting in the docstring that the VRAM rationale does not apply there, rather than adding a branch that pretends to solve something. Describing a limitation beats hiding it.

## Two follow-ups, neither blocking

The wiring tests all run at `device="cpu"` and assert the literal `"auto"`, so a trainer that hardcodes `"auto"` and never calls the resolver still passes — I replaced the call in `dpo.py` with a literal and the suite stayed green. Patching the resolver with a sentinel in the wrapper's own namespace would prove the *call* rather than the *value*.

And no docs update here — #471's prose is general enough to cover both once it lands, but nothing tells a dpo/kto/grpo user their VRAM changed.

## Ordering note

Your PR landing first is what lets #471 shrink: I have asked @blackcoderx to rebase `_resolve_load_dtype` onto your `resolve_frozen_base_load_dtype` and keep only the full-fine-tuning branch on top. One resolver, not two. I have told them it is now unblocked.

Filing the issue as an explicit follow-on that credited #339 and #471 by number, instead of quietly opening a competing PR, is why this was easy to sequence. Thank you.

AmirF194:
Thanks for the thorough re-verification, especially catching the reward-head bf16 issue before merge and for sequencing this ahead of #471. Glad the control assertion was useful.

Issue #495, opened 2026-08-23
The dtype= guard only matches AutoModel*, so a concrete model class such as WhisperForConditionalGeneration is invisible to it


MakazhanAlpamys:
#489 added `tests/test_transformers_floor_compat.py`, a static AST scan that fails if a production `from_pretrained` call uses `dtype=` — the transformers `>=4.56` rename that **hard-fails at model load** on anything below 4.56 (measured: `TypeError: LlamaForCausalLM.__init__() got an unexpected keyword argument 'dtype'` on 4.46.1).

The guard works, and it caught the twelve sites that shipped broken. But `_is_auto_model_load` (`tests/test_transformers_floor_compat.py:47-54`) requires a dotted-name part starting with `AutoModel`, so it only sees the `AutoModel*` factories.

## Reproduction

Add `dtype="auto"` to the real `WhisperForConditionalGeneration.from_pretrained` call in `src/soup_cli/trainer/asr.py:220` — the ASR training path — and the suite stays **green**.

Both live Whisper sites (`trainer/asr.py:220`, `commands/infer.py:476`) are clean today, so this is a **coverage gap, not a live bug**. It is filed because it is exactly the shape that lets the next instance through: the guard reads as "no production load site uses `dtype=`" while actually meaning "no `AutoModel*` load site does".

## Fix path

Widen the predicate to the concrete task classes as well as the factories:

```
AutoModel* | AutoProcessor | *ForCausalLM | *ForConditionalGeneration | *ForSequenceClassification
```

This is enumerable rather than open-ended — there are 18 distinct load-site class names under `src/` today, so the alternative (allowlist every name) is also viable if you prefer an explicit list. Prefer the suffix rule: an allowlist of names silently stops covering a class added later, which is the failure mode this project's "scan, don't list" convention exists to avoid.

## Acceptance criteria

- [ ] `dtype=` on `WhisperForConditionalGeneration.from_pretrained` in `src/` fails a named test
- [ ] The existing `AutoModel*` coverage still fails (do not trade one for the other)
- [ ] The scanner's existing negative controls still pass — it must keep ignoring `FastLanguageModel`, `PeftModel`, `torch.zeros` and wrapper kwargs, so widening does not produce false positives
- [ ] Verified by making the mutation and running it, not by inspection

Relatedly, `torch_dtype=` is itself removed in transformers 5, which the `[mlx]` extra can pull. Out of scope here, but worth knowing that this guard's rule will need to invert one day.

[...]

PR #496, opened 2026-08-23
test(ci): derive the Transformers floor policy

### LINKED ISSUE #494: The transformers floor is hardcoded in three places and derived from none, so a pin bump silently leaves CI testing an unrelated version
#489 added a `transformers-floor` CI job that pins `transformers==4.46.1`. The version is written down in **three** places and derived from none of them:

- `.github/constraints/transformers-floor.txt:22`
- the version assertion in the `transformers-floor` job (`.github/workflows/ci.yml`)
- `tests/test_transformers_floor_compat.py:145`

None of them reads `pyproject.toml`.

## Reproduction

Change the declared pin and watch the "floor" job keep testing the old number:

```
# pyproject.toml
- transformers>=4.36.0,<5.0.0
+ transformers>=4.52.0,<5.0.0
```

`tests/test_transformers_floor_compat.py` stays **green** — it only opens the constraints file and the workflow, never `pyproject.toml`. CI then goes on pinning 4.46.1 and calling it "the floor" while the project's actual floor is 4.52.

That is the same class of defect the job was created to prevent: a declared bound that nothing executes.

## Partial mitigation that already exists

`tests/test_v0640_part_c.py:738-823` asserts the exact requirement strings, so a pin bump does turn *something* red. But it points at a metadata test, not at the constraints file, so the drift is easy to walk past without realising the floor job is now meaningless.

## Why the value is not simply `pyproject.toml`'s floor

Deliberate, and worth preserving: `transformers==4.36.0` **cannot resolve** — `trl 0.14.0` requires `transformers>=4.46.0`, and every file of `transformers 4.46.0` is yanked on PyPI, so 4.46.1 is the lowest installable version. The constraints file documents this and explicitly refuses to relabel 4.55.4 as "the floor". Any fix must keep that honest exception rather than forcing equality.

## Fix path

Have `tests/test_transformers_floor_compat.py` parse `transformers>=X` out of `pyproject.toml` and assert:

1. the constraint pin is `>= X` (so it can never be *below* the declared floor), and
2. when the constraint pin is strictly greater than `X`, the constraints file still carries a comment explaining why — so the exception stays documented rather than silently drifting.

This repo already has the pattern to copy: `tests/test_requires_python_bound.py` derives the Python bound from the CI matrix instead of restating it.

## Acceptance criteria

- [ ] Raising the `transformers` floor in `pyproject.toml` without touching the constraints file fails a named test
- [ ] Lowering the constraint pin below the declared floor fails a named test
- [ ] The 4.36-is-unresolvable exception still passes, with its reason still required to be documented
- [ ] Verified by actually making each mutation and running, not by inspection

Small and self-contained. No GPU, no model download.

Amix29:
## Summary

- derive the declared Transformers lower bound from the `train` extra in `pyproject.toml`;
- reject a floor constraint below that bound and require a documented resolver/yank reason when the tested pin is higher;
- make the floor CI job read its exact Transformers and TRL versions from the constraints file instead of restating them;
- add named controls for all three mutations required by the issue;
- use the merged changelog-fragment convention with `496.fixed.md`.

## Mutation evidence

Each requested mutation was applied to the real repository file and the named policy test was run:

- raising the declared floor from 4.36.0 to 4.52.0 failed: constraint 4.46.1 is below declared floor 4.52.0;
- lowering the constraint to 4.35.0 failed: it is below declared floor 4.36.0;
- removing the resolver/yank explanation failed: the higher pin had no documented reason.

## Validation

- focused policy, Python-bound, and changelog tests: 40 passed
- full local suite on current `main`: 18,331 passed, 285 skipped, 5 deselected
- coverage: 81.28%
- Ruff: clean
- actionlint: clean
- `git diff --check`: clean

Closes #494

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,13 +87,26 @@ jobs:
         run: pip install -e ".[dev]" -c .github/constraints/transformers-floor.txt
       - name: Verify constrained dependency graph
         run: python -m pip check
-      - name: Assert constrained Transformers and TRL versions
+      - name: Assert installed versions match floor constraints
         run: |
           python - <<'PY'
           import importlib.metadata as md
+          import re
+          from pathlib import Path
 
-          expected = {"transformers": "4.46.1", "trl": "0.14.0"}
-          for name, want in expected.items():
+          constraints = Path(".github/constraints/transformers-floor.txt").read_text(
+              encoding="utf-8"
+          )
+          pairs = re.findall(
+              r"^([A-Za-z0-9_-]+)==([^\s#]+)\s*$",
+              constraints,
+              re.MULTILINE,
+          )
+          expected = dict(pairs)
+          assert len(expected) == len(pairs), "duplicate package pin in floor constraints"
+          for name in ("transformers", "trl"):
+              assert name in expected, f"floor constraints missing exact {name} pin"
+              want = expected[name]
               got = md.version(name)
               assert got == want, (
                   f"floor job expected {name}=={want} from "
--- a/changelog.d/0.73.3/496.fixed.md
+++ b/changelog.d/0.73.3/496.fixed.md
@@ -1,0 +1,5 @@
+- **Transformers floor CI now tracks declared dependency metadata (#494 by @Amix29 in
+  #496).** The compatibility guard derives the training lower bound from `pyproject.toml`,
+  refuses a tested pin below it, requires a documented reason for a higher resolvable pin,
+  and makes the workflow read exact versions from the constraints file instead of
+  restating them.
--- a/tests/test_transformers_floor_compat.py
+++ b/tests/test_transformers_floor_compat.py
@@ -21,14 +21,18 @@ kwargs that are not Transformers load APIs are out of scope.
 from __future__ import annotations
 
 import ast
+import re
 from pathlib import Path
+from unittest.mock import patch
 
 import pytest
+import yaml
 
 REPO_ROOT = Path(__file__).resolve().parent.parent
 SRC_ROOT = REPO_ROOT / "src" / "soup_cli"
 CI_WORKFLOW = REPO_ROOT / ".github" / "workflows" / "ci.yml"
 CONSTRAINTS = REPO_ROOT / ".github" / "constraints" / "transformers-floor.txt"
+PYPROJECT = REPO_ROOT / "pyproject.toml"
 
 _LOAD_METHODS = frozenset({"from_pretrained", "from_config"})
 _MODEL_LOAD_CLASS_SUFFIXES = (
@@ -36,6 +40,63 @@ _MODEL_LOAD_CLASS_SUFFIXES = (
     "ForConditionalGeneration",
     "ForSequenceClassification",
 )
+_VERSION = re.compile(r"\d+(?:\.\d+){1,2}")
+
+
+def _declared_transformers_floor(pyproject: str) -> str:
+    """Return the Transformers lower bound from the ``train`` extra."""
+    train = re.search(r"^train\s*=\s*\[(.*?)^\]", pyproject, re.MULTILINE | re.DOTALL)
+    assert train, "pyproject.toml has no train dependency list"
+    requirement = re.search(
+        r'"transformers\s*>=\s*(?P<floor>\d+(?:\.\d+){1,2})[^\"]*"',
+        train.group(1),
+    )
+    assert requirement, "train dependencies have no transformers >= lower bound"
+    return requirement.group("floor")
+
+
+def _constraint_pin(constraints: str, package: str) -> str:
+    """Return one exact package pin from the floor constraints file."""
+    pins = re.findall(rf"^{re.escape(package)}==([^\s#]+)\s*$", constraints, re.MULTILINE)
+    assert len(pins) == 1, f"expected exactly one {package} constraint pin, found {pins}"
+    return pins[0]
+
+
+def _version_key(version: str) -> tuple[int, int, int]:
+    assert _VERSION.fullmatch(version), f"floor version must be numeric, got {version!r}"
+    parts = [int(part) for part in version.split(".")]
+    parts.extend([0] * (3 - len(parts)))
+    return parts[0], parts[1], parts[2]
+
+
+def _validate_transformers_floor_policy(
+    pyproject: str,
+    constraints: str,
+) -> tuple[str, str]:
+    """Require the tested floor to follow, or explain its offset from, metadata."""
+    declared = _declared_transformers_floor(pyproject)
+    pinned = _constraint_pin(constraints, "transformers")
+    assert _version_key(pinned) >= _version_key(declared), (
+        f"transformers constraint {pinned} is below declared floor {declared}"
+    )
+
+    if _version_key(pinned) > _version_key(declared):
+        comments = "\n".join(
+            line.lstrip()[1:].strip()
+            for line in constraints.splitlines()
+            if line.lstrip().startswith("#")
+        )
+        explains_offset = re.search(
+            rf"\b{re.escape(declared)}\b[^\n]*"
+            r"\b(?:not resolvable|unresolvable|yanked)\b",
+            comments,
+            re.IGNORECASE,
+        )
+        assert pinned in comments and explains_offset, (
+            f"transformers constraint {pinned} exceeds declared floor {declared} without "
+            "a documented reason naming both versions and the resolver/yank exception"
+        )
+    return declared, pinned
 
 
 def _attr_parts(node: ast.AST) -> list[str]:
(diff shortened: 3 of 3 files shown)


Amix29:
Final mutation hardening is now pushed. The exception check requires the resolver/yank reason on the comment line that names the declared floor. I verified this by removing only that reason while leaving the pinned-version `non-yanked` note intact; the named policy test failed, then passed again after restoration. The final commit also passes the full local suite: 18,331 passed, 285 skipped, 5 deselected, with Ruff and actionlint clean.

MakazhanAlpamys:
**All four of #494's acceptance criteria are met, and I verified each by mutation with a control on `main`.** One MEDIUM to fix, and it is the specific thing #494 warned about.

## Verified, with negative controls

| Mutation | On this branch | On `main` (control) |
|---|---|---|
| `pyproject.toml` `>=4.36.0` → `>=4.52.0` | **3 named tests fail** | **11 passed — SURVIVED** |
| constraints `4.46.1` → `4.30.0` | 3 named tests fail | — |
| reason text `"not resolvable"` → `"awkward"` | 3 named tests fail | — |

The control is the point: the drift #494 describes was real, and this is what closes it. The reason stays *required*, not merely present — `test_removing_declared_floor_exception_explanation_fails` earns its name.

Baseline **15 passed** (was 11). `ruff` clean including `scripts/`. And `changelog.d/0.73.3/496.fixed.md` is the **first PR to use the convention you built yesterday** — filename keyed on the PR number, directory matching the newest release heading, accepted by `validate_fragments`, assembled verbatim, credit line in house format. It works end to end.

## MEDIUM — the guard fires on a correct future state

This is the one thing #494 explicitly asked to preserve: the exception exists so that a maintainer can one day make the dependency-policy decision and raise the declared floor to the tested pin. I did exactly that — `pyproject.toml` → `transformers>=4.46.1`, i.e. declared == pinned, which is a **legitimate, correct end state**:

```
E   AssertionError: assert (4, 46, 1) > (4, 46, 1)
E    +  where (4, 46, 1) = _version_key('4.46.1')
E   Failed: DID NOT RAISE <class 'AssertionError'>

FAILED ...::test_constraints_pin_a_documented_resolvable_floor
FAILED ...::test_removing_declared_floor_exception_explanation_fails
```

Two reds, neither with a message explaining what to do. Your policy function `_validate_transformers_floor_policy` **passes** on that state — correctly. It is the tests around it that over-specify today's strict inequality.

That is the project's own documented anti-pattern, and I have quoted it at other contributors this week: *a guard that fires on correct code is one people delete*, and then it protects nothing. Roughly three lines: drop the strict `>` (or gate it and give it a message), and branch `test_removing_declared_floor_exception_explanation_fails` on `declared == pinned` rather than relying on the message-less `assert without_reason != constraints` precondition.

## Two smaller ones

**LOW — `tests/…:215-217` assert code *spelling*, not behaviour.** I replaced the workflow's real check (`assert got == want` → `assert True` in `ci.yml`) while leaving every asserted string intact: **15 passed, survived.** The assertions look for `"expected = dict("` and `'for name in ("transformers", "trl")'` in the job text, which cannot see a gutted comparison.

**This is inherited, not a regression** — the identical mutation survives on `main` too. But you replaced assertions on the semantic *values* (`"4.46.1"`, `"0.14.0"`) with assertions on code *shape*, and that is the direction CLAUDE.md warns about after #321, where a test battery passed with the behaviour killed and the strings kept. The stronger version is ~15 lines and demonstrably works: parse `ci.yml` with `yaml.safe_load`, pull the heredoc body out of the step's `run`, `exec` it with `importlib.metadata.version` stubbed, and assert it accepts the pinned pair and rejects a mismatched one. Your call whether to do it here or file it.

**LOW — noisy diagnostics.** All three mutations above fail the *same* three tests, and one of them fails through a message-less precondition rather than through the policy. A maintainer bumping the pin gets three reds, only one of which explains itself.

## Credit where it is due, on the CI logic

I extracted your new `run:` step through a real YAML parse rather than a text grep and executed it as bash would receive it. The block scalar de-indents correctly, the regex parses the real CRLF constraints file to `{'transformers': '4.46.1', 'trl': '0.14.0'}` — correctly excluding comment lines that contain `transformers==4.46.1` — and it **discriminates**: against a locally installed 4.57.6 it exits 1 with the intended diagnostic. That is careful work and it is not something the PR itself claims.

## Rebase

**#497 landed first** — @taljeon's fix for the sibling follow-up #495, and their first contribution. It touches the same file, and the conflict is one hunk immediately after `_LOAD_METHODS`: they add `_MODEL_LOAD_CLASS_SUFFIXES`, you add `_VERSION` plus four helpers. Purely additive adjacency, zero semantic overlap — keep both blocks in either order. I constructed the combination and ran it: **16 passed**, and all four mutations (Whisper, `AutoModel*`, pin bump, constraint lower) still die, so the combined tree satisfies both #494 and #495.

Ordering was decided on risk and on who carries the rebase: yours changes CI config, theirs is test-only, and handing the merge conflict to the contributor with five merges here rather than the one with zero seemed the right way round.

Fix the MEDIUM and this merges.

Not verified by me: whether `pip install -e ".[dev]" -c .github/constraints/transformers-floor.txt` actually resolves on a clean runner — I proved the new check's logic is correct and discriminating in place, but did not perform the real install. `actionlint` was also not run against the modified workflow.

MakazhanAlpamys:
Small status note so you are not guessing at the order.

**#497 landed** (@taljeon's fix for the sibling follow-up #495, and their first contribution here), so this now shows as conflicting. As I said in the review, that conflict is **one hunk, purely additive adjacency**: immediately after `_LOAD_METHODS`, they add `_MODEL_LOAD_CLASS_SUFFIXES` and you add `_VERSION` plus four helpers. Keep both blocks, in either order. Nothing else in your diff is affected.

I constructed the combined tree and ran it before deciding the order: **16 passed**, and all four mutations — Whisper, `AutoModel*`, pin bump, constraint lowered — still die. So the combination satisfies both #494 and #495, and the rebase is genuinely mechanical.

Still outstanding from the review is the one MEDIUM: the guard fires on a **correct** future state. Raising the declared floor to equal the pinned one — the legitimate migration your exception exists to permit — produces two message-less reds:

```
E   AssertionError: assert (4, 46, 1) > (4, 46, 1)
E   Failed: DID NOT RAISE
```

Your policy function passes on that state, correctly; it is the tests around it that over-specify today's strict inequality.

**If the rebase plus that fix is more than you want to carry right now, say so** — you have had five PRs merged in about a day, which is a lot, and there is no expectation that you keep that pace. I can do the rebase on your branch with authorship preserved and leave the MEDIUM to you, or park the whole thing until you have time. Just tell me which is useful.

Amix29:
Addressed the MEDIUM and the behavioural LOW in `468d152`, rebased onto #497.

- The combined tree keeps both additions from the conflict: #497's concrete-model suffix scanner and #496's floor-policy helpers.
- The policy test now accepts `pinned >= declared` rather than requiring a permanent strict offset.
- `test_declared_floor_may_catch_up_to_the_tested_constraint` pins the valid `declared == pinned` state directly.
- The offset-explanation test now branches out when there is no offset, so it does not demand a resolver exception after the declared floor catches up.
- The workflow test now parses `ci.yml`, extracts and executes the actual Python heredoc, supplies synthetic constraint pins, and stubs `importlib.metadata.version`. It proves both `transformers` and `trl` mismatches are rejected instead of asserting source spelling.

Negative controls:

- changing the declared floor to `4.46.1` (equal to the tested pin) -> **19 passed**
- replacing the workflow's `assert got == want` with `assert True` -> both named mismatch cases **FAILED**

Additional verification requested in the review:

- clean constrained install with `transformers==4.46.1`, `trl==0.14.0`
- `pip check` -> no broken requirements
- focused floor/dtype suite -> **20 passed**
- full suite -> **18,335 passed, 285 skipped, 5 deselected**
- Ruff -> clean
- `actionlint v1.7.12` (the CI-pinned release, archive SHA-256 verified) -> clean

Hitesh-XS:
@Amix29 Yeah sure you can take it (i have my exam these weekend so better you finish it ) ;)

MakazhanAlpamys:
**Merged** as `dbf2428`, sixth CONTRIBUTORS.md bullet in (`4c23aff`). CI 13/13 at `468d152`, SHA verified.

I re-ran the blocking mutation myself rather than reading your summary. Raising the declared floor to equal the tested pin — the legitimate future migration:

```
before your fix:  2 failed, both message-less
                  E   AssertionError: assert (4, 46, 1) > (4, 46, 1)
                  E   Failed: DID NOT RAISE
after:            19 passed
```

And all three genuinely-wrong states still fail by name: declared floor above the pin, constraint below the declared floor, and the justification reworded to drop "not resolvable".

**You deleted the over-specifying assertions instead of patching around them.** That is the harder and correct choice — a weakened assertion that still exists is the kind of thing nobody revisits. The new `test_declared_floor_may_catch_up_to_the_tested_constraint` has a control proving it is not vacuous.

**And you fixed the spelling-not-behaviour weakness you had inherited rather than leaving it.** I had flagged it as *inherited, not a regression* and would have accepted it staying. Instead you pulled the heredoc out of the workflow with `yaml.safe_load` and `exec` it under patched `read_text` / `importlib.metadata.version` — which is why gutting the comparison **and** re-hardcoding the expectation both fail by name now. That is the #321 lesson applied properly: a battery that passes with the behaviour killed and the strings kept is not a test.

Also confirmed the rebase onto #497 held — the concrete-class coverage still fires, so the combined tree satisfies both #494 and #495.

Not verified by me: the floor CI cell actually resolving transformers 4.46.1 + trl 0.14.0 on a clean runner. That needs a real constrained install.

Six merges in about a day, and this one closed a follow-up I filed the same morning as a replacement for blocking your earlier PR. Thank you.

PR #497, opened 2026-08-23
test: cover concrete Transformers model loads

### LINKED ISSUE #495: The dtype= guard only matches AutoModel*, so a concrete model class such as WhisperForConditionalGeneration is invisible to it

taljeon:
## What does this PR do?

Fixes #495.

The Transformers floor guard previously recognized only `AutoModel*` factories.
This change widens the AST predicate to cover `AutoProcessor` and the requested
concrete model class suffixes while preserving the existing negative controls.

It also adds a concrete Whisper regression test and renames the production-scan
test and failure message to describe the wider protection accurately.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

- [x] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed — not needed for this test-only change
- [x] Added a changelog fragment, or this change has no user-visible impact — no user-visible impact

## Testing

- Python 3.12: `python -m pytest -o addopts= -q tests/test_transformers_floor_compat.py` (12 passed)
- Python 3.10: the same test file (12 passed)
- temporary real mutation at `src/soup_cli/trainer/asr.py:220`, confirmed red with the final named production-scan test and then reverted

The full training-stack test suite was not installed locally; CI can run the
repository's full matrix. The changed scanner test is dependency-light and was
run on the oldest and newest supported Python versions available here.

## AI assistance

AI assistance was used to investigate, implement, and test this change. I
reviewed the scoped diff and mutation evidence.

### DIFF
--- a/tests/test_transformers_floor_compat.py
+++ b/tests/test_transformers_floor_compat.py
@@ -31,6 +31,11 @@ CI_WORKFLOW = REPO_ROOT / ".github" / "workflows" / "ci.yml"
 CONSTRAINTS = REPO_ROOT / ".github" / "constraints" / "transformers-floor.txt"
 
 _LOAD_METHODS = frozenset({"from_pretrained", "from_config"})
+_MODEL_LOAD_CLASS_SUFFIXES = (
+    "ForCausalLM",
+    "ForConditionalGeneration",
+    "ForSequenceClassification",
+)
 
 
 def _attr_parts(node: ast.AST) -> list[str]:
@@ -44,14 +49,19 @@ def _attr_parts(node: ast.AST) -> list[str]:
     return list(reversed(parts))
 
 
-def _is_auto_model_load(call: ast.Call) -> bool:
-    """True for ``AutoModel*.from_pretrained`` / ``from_config`` (any prefix)."""
+def _is_transformers_model_load(call: ast.Call) -> bool:
+    """True for supported Transformers model factory and concrete-class loads."""
     parts = _attr_parts(call.func)
     if len(parts) < 2:
         return False
     if parts[-1] not in _LOAD_METHODS:
         return False
-    return any(part.startswith("AutoModel") for part in parts[:-1])
+    model_class = parts[-2]
+    return (
+        model_class.startswith("AutoModel")
+        or model_class == "AutoProcessor"
+        or model_class.endswith(_MODEL_LOAD_CLASS_SUFFIXES)
+    )
 
 
 def find_post_floor_dtype_kwargs(source: str, *, filename: str = "<string>") -> list[str]:
@@ -59,7 +69,7 @@ def find_post_floor_dtype_kwargs(source: str, *, filename: str = "<string>") -> 
     tree = ast.parse(source, filename=filename)
     hits: list[str] = []
     for node in ast.walk(tree):
-        if not isinstance(node, ast.Call) or not _is_auto_model_load(node):
+        if not isinstance(node, ast.Call) or not _is_transformers_model_load(node):
             continue
         for kw in node.keywords:
             if kw.arg == "dtype":
@@ -93,10 +103,10 @@ def _transformers_floor_job_block(text: str) -> str:
 
 
 class TestTransformersFloorDtypeGuard:
-    def test_no_production_auto_model_load_uses_dtype_kwarg(self):
+    def test_no_production_transformers_model_load_uses_dtype_kwarg(self):
         hits = iter_production_hits()
         assert hits == [], (
-            "AutoModel*.from_pretrained / from_config must use torch_dtype= "
+            "Transformers model from_pretrained / from_config calls must use torch_dtype= "
             "(works across transformers>=4.36,<5). dtype= is the >=4.56-only "
             "rename and TypeErrors on older installs inside our declared range "
             f"(#478). Offending sites: {hits}"
@@ -110,6 +120,14 @@ class TestTransformersFloorDtypeGuard:
         )
         hits = find_post_floor_dtype_kwargs(bad, filename="mutation.py")
         assert hits == ["mutation.py:2"]
+
+    def test_scanner_flags_concrete_conditional_generation_class(self):
+        bad = (
+            "from transformers import WhisperForConditionalGeneration\n"
+            "WhisperForConditionalGeneration.from_pretrained('x', dtype='auto')\n"
+        )
+        hits = find_post_floor_dtype_kwargs(bad, filename="whisper.py")
+        assert hits == ["whisper.py:2"]
 
     def test_scanner_allows_torch_dtype(self):
         good = (


MakazhanAlpamys:
Welcome — first PR here, and you picked one of the two follow-ups I filed less than an hour ago. Thank you.

Two practical notes so nothing surprises you:

**Your CI needed manual approval.** GitHub holds workflow runs on a first-time contributor's PR until a maintainer releases them. I have done that, and I will keep doing it — but be aware it applies to **every** push you make until your first PR is merged, not just this one. If CI looks stuck rather than running, that is why; ping me and I will release it.

**How I review here, so the process is not opaque:** I do not take a PR description on trust. I run the tests myself, and I mutate the fix — break the line the change is built on and check that a *named* test fails. If nothing fails, the tests are not guarding the fix and I say so. I also read every deleted line in a diff, yours included; that is not suspicion of you specifically, it is the same pass everyone gets.

For this one specifically I will be checking the two things #495's acceptance criteria name: that `dtype="auto"` on the real `WhisperForConditionalGeneration.from_pretrained` in `trainer/asr.py:220` now fails a named test, and — just as important — that the existing `AutoModel*` coverage did **not** get traded away for it. Widening a scanner has a second failure mode too: it must keep ignoring `FastLanguageModel`, `PeftModel` and `torch.zeros`, because a guard that fires on correct code is one people delete.

Heads-up on a collision that is not your fault: **#496 by @Amix29 edits the same file** (it implements #494, the sibling follow-up). Whichever lands second will need a small rebase. I will sort the ordering out and tell you if anything is needed from you — do not go rebasing pre-emptively.

Review coming shortly.

MakazhanAlpamys:
**Merged** as `a2455d9`, and you are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) (`1b728d1`). CI 13/13 at `af14bdd`, SHA verified against the head. Welcome.

## What I ran

The decisive check was a **negative control**, not just a passing test. I added `dtype="auto"` to the real `WhisperForConditionalGeneration.from_pretrained` at `trainer/asr.py:220` and ran the identical mutation on both trees:

```
on your branch  -> FAILED test_no_production_transformers_model_load_uses_dtype_kwarg
on origin/main  -> 11 passed, SURVIVED
```

That is what proves the gap was real and that your change is what closes it. A green suite on its own would not have shown either.

I also read all six deleted lines, because a shrinking diff is where I always look — every one is a rename or a reworded failure message, and the file goes 11 tests to 12. Nothing removed, nothing weakened, and no stale references to the old names anywhere in the repo.

Worth admitting: my first two attempts at that mutation **failed to apply** — once the anchor did not match, once it matched twice — and both times the suite reported a clean pass. An unapplied mutation is indistinguishable from a surviving one, so I assert the anchor count before writing. Third attempt landed and the test fired. Mentioning it because it is the single easiest way to fool yourself when checking work like this.

## Two things you got right that I check on every widening, and that you did unprompted

**You kept the old coverage.** Widening a scanner tempts people to replace the narrow rule with the broad one. `torch_dtype=` → `dtype=` on `AutoModelForCausalLM.from_pretrained` still fails the same test — you added coverage rather than trading it.

**You did not create false positives.** `FastLanguageModel`, `PeftModel`, `torch.zeros` and wrapper kwargs like `load_model_and_tokenizer(dtype=...)` are all still correctly ignored, as are `AutoTokenizer`, `AutoConfig` and `WhisperProcessor`. This matters more than it sounds: a guard that fires on correct code is one people eventually delete, and then it protects nothing.

**And you used the suffix rule rather than a name allowlist**, which is the harder choice and the one the issue asked for. An allowlist silently stops covering a class added later — the exact failure mode this project's scan-don't-list convention exists to prevent.

## One note, and it is against my spec rather than your work

The three suffixes are themselves a small allowlist, so `*ForSpeechSeq2Seq`, `*ForImageTextToText`, `*ForTokenClassification` and `*ForMaskedLM` remain uncovered — meaning `AutoModelForSpeechSeq2Seq` is guarded while a concrete `WhisperForSpeechSeq2Seq` would not be. Slightly ironic given Whisper is the family that prompted the issue. **#495 specified exactly those three, so you implemented what was asked**; the narrowness is mine. Every live site is clean today. If you want to widen it further, open a follow-up and claim it — otherwise I will file one.

Also for context, since it explains why this mattered: `dtype=` is not a style preference. It hard-fails at model load on every transformers below 4.56 — I measured `TypeError: __init__() got an unexpected keyword argument 'dtype'` on 4.46.1 — and twelve production sites had shipped broken before #489 caught them. Your change extends that protection to the classes the original scan could not see.

Ordering note: #496 touches the same file and will rebase around your change, not the other way round. Nothing needed from you.

A guard one day old, and you found the hole in it. Thank you — please stick around.

PR #500, opened 2026-08-23
test: extend Transformers dtype guard coverage

### LINKED ISSUE #499: The dtype guard still misses *ForSpeechSeq2Seq and *ForImageTextToText — so a concrete Whisper or vision class is invisible to it
Follow-up to #495, and the limitation is in **my** wording there rather than in @taljeon's fix for it.

#497 widened `_is_transformers_model_load` (`tests/test_transformers_floor_compat.py`) from "class name starts with `AutoModel`" to a suffix rule. That closed the reported case. But #495 named exactly three suffixes, so the predicate today covers:

```
AutoModel*  |  AutoProcessor  |  *ForCausalLM  |  *ForConditionalGeneration  |  *ForSequenceClassification
```

and these remain invisible:

- **`*ForSpeechSeq2Seq`** — so `AutoModelForSpeechSeq2Seq` is guarded while a concrete `WhisperForSpeechSeq2Seq` is not. Ironic, because Whisper is the family that prompted #495 in the first place.
- **`*ForImageTextToText`** — the modern class name for the vision families this repo ships recipes for.
- **`*ForTokenClassification`**, **`*ForMaskedLM`**, and bare model classes such as `BertModel`.

## Why it matters

`dtype=` is the transformers `>=4.56` rename of `torch_dtype=`, and it **hard-fails at model load** on anything below 4.56. Measured on 4.46.1 against a real checkpoint:

```
dtype=torch.float16       -> TypeError: LlamaForCausalLM.__init__() got an
torch_dtype=torch.float16 -> OK
```

Twelve production sites shipped with that spelling before #489 caught them (`soup chat`, `serve`, `merge`, `infer`, `diff`, `export`). The guard exists so it cannot happen again — and right now it reads as "no production load site uses `dtype=`" while actually meaning "no site whose class matches one of five patterns does".

**Every live load site under `src/` is clean today**, so this is a coverage gap, not a live bug.

## Fix path

Extend `_MODEL_LOAD_CLASS_SUFFIXES`. Keep it a suffix rule, not a name allowlist — an allowlist silently stops covering a class added later, which is the failure mode this project's scan-don't-list convention exists to prevent.

## Acceptance criteria

- [ ] `dtype="auto"` on a `*ForSpeechSeq2Seq` or `*ForImageTextToText` load in `src/` fails a named test
- [ ] The existing `AutoModel*`, `*ForCausalLM`, `*ForConditionalGeneration`, `*ForSequenceClassification` coverage still fails — **do not trade one for another**
- [ ] The four false-positive controls still pass: the scanner must keep ignoring `FastLanguageModel`, `PeftModel`, `torch.zeros`, and wrapper kwargs such as `load_model_and_tokenizer(dtype=...)`. A guard that fires on correct code is one people delete.
- [ ] Verified by making each mutation and running it, not by inspection

## Also worth deciding while you are in there

The widened rule already reaches non-transformers classes — `AutoAWQForCausalLM`, `AutoGPTQForCausalLM`, `ORTModelForCausalLM` are all flagged. They are clean today and all forward to transformers, so there is no false positive; but the function is named `_is_transformers_model_load` and its docstring says "Transformers", which is now broader than accurate. Either narrow it or rename it.

@taljeon — you closed #495 and I said I would file this if you did not want it. It is yours if you do; otherwise it is open to anyone. No obligation either way.

Small, no GPU, no model download.

k4its1t:
## What does this PR do?

Extends the Transformers floor-compatibility scanner so concrete `*ForSpeechSeq2Seq`, `*ForImageTextToText`, `*ForTokenClassification`, and `*ForMaskedLM` classes cannot introduce the post-floor `dtype=` keyword unnoticed.

The internal predicate is renamed to describe the dtype contract it enforces rather than implying every matched forwarding class comes directly from Transformers. Four focused regression tests cover both `from_pretrained` and `from_config`; the existing `AutoModel*` coverage and all four false-positive controls remain intact.

Closes #499.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Validation

- `pytest tests/ -q --tb=short`: 18,349 passed, 285 skipped, 5 deselected; 81.30% coverage (77% required)
- targeted guard suite: 16 passed
- mutation check: removing the four new suffixes makes the four corresponding tests fail; restoring them returns the suite to green
- repository pre-commit hooks

## Checklist

- [x] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed — no user-facing behavior changed
- [x] Added a changelog fragment, or this change has no user-visible impact — no fragment needed

Implemented with Codex assistance and reviewed through the validation above.

### DIFF
--- a/tests/test_transformers_floor_compat.py
+++ b/tests/test_transformers_floor_compat.py
@@ -11,8 +11,8 @@ installs inside the declared range — exactly what #471 nearly shipped.
 comments and ``.github/constraints/transformers-floor.txt``). Raising Soup's
 declared floor is a dependency-policy call, not something this suite does.
 Until that decision lands, the accepted #478 fallback is a static guard: no
-production ``AutoModel*.from_pretrained`` / ``from_config`` call site may pass
-``dtype=``.
+production model ``from_pretrained`` / ``from_config`` call site governed by
+the Transformers dtype contract may pass ``dtype=``.
 
 Unsloth's ``FastLanguageModel.from_pretrained(..., dtype=)`` and Soup wrapper
 kwargs that are not Transformers load APIs are out of scope.
@@ -34,7 +34,11 @@ _LOAD_METHODS = frozenset({"from_pretrained", "from_config"})
 _MODEL_LOAD_CLASS_SUFFIXES = (
     "ForCausalLM",
     "ForConditionalGeneration",
+    "ForImageTextToText",
+    "ForMaskedLM",
     "ForSequenceClassification",
+    "ForSpeechSeq2Seq",
+    "ForTokenClassification",
 )
 
 
@@ -49,8 +53,8 @@ def _attr_parts(node: ast.AST) -> list[str]:
     return list(reversed(parts))
 
 
-def _is_transformers_model_load(call: ast.Call) -> bool:
-    """True for supported Transformers model factory and concrete-class loads."""
+def _is_dtype_guarded_model_load(call: ast.Call) -> bool:
+    """True for model loads governed by the Transformers dtype contract."""
     parts = _attr_parts(call.func)
     if len(parts) < 2:
         return False
@@ -65,11 +69,11 @@ def _is_transformers_model_load(call: ast.Call) -> bool:
 
 
 def find_post_floor_dtype_kwargs(source: str, *, filename: str = "<string>") -> list[str]:
-    """Return ``file:line`` hits for ``dtype=`` on AutoModel load/config calls."""
+    """Return ``file:line`` hits for ``dtype=`` on guarded model load/config calls."""
     tree = ast.parse(source, filename=filename)
     hits: list[str] = []
     for node in ast.walk(tree):
-        if not isinstance(node, ast.Call) or not _is_transformers_model_load(node):
+        if not isinstance(node, ast.Call) or not _is_dtype_guarded_model_load(node):
             continue
         for kw in node.keywords:
             if kw.arg == "dtype":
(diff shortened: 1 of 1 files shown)


MakazhanAlpamys:
Welcome — first PR here, and you claimed #499 within an hour of it being filed. Thank you.

Two things so nothing surprises you:

**Your CI needed manual approval.** GitHub holds workflow runs on a first-time contributor's PR until a maintainer releases them. Done. It applies to **every** push you make until your first merge, not just this one — so if CI looks stuck rather than running, that is why. Ping me and I will release it.

**How review works here:** I do not take a PR description on trust. I run the tests myself and I mutate the fix — break the line the change is built on and check that a *named* test fails. If nothing fails, the tests are not guarding the fix and I say so. I also read every deleted line in a diff. Same pass for everyone, including PRs I wrote myself; I found two of my own mistakes that way today.

For this one specifically, three things from #499's criteria:

1. Each newly covered suffix must actually fire — I will add `dtype="auto"` to a real load site of that shape and check a named test goes red.
2. **The existing coverage must still fire.** Widening a scanner tempts people to replace the narrow rule rather than extend it; `AutoModel*`, `*ForCausalLM`, `*ForConditionalGeneration` and `*ForSequenceClassification` all have to keep failing.
3. **No false positives.** `FastLanguageModel`, `PeftModel`, `torch.zeros` and wrapper kwargs like `load_model_and_tokenizer(dtype=...)` must stay ignored. A guard that fires on correct code is one people eventually delete, and then it protects nothing.

One note on scope: this is a test-only change, so **no changelog fragment is needed** — fragments are for user-visible changes. If you added one, that is not a problem; I am telling you so you are not left guessing at the convention.

Worth knowing why the guard exists at all: `dtype=` is the transformers `>=4.56` rename of `torch_dtype=`, and it **hard-fails at model load** on anything below 4.56. I measured it — `TypeError: LlamaForCausalLM.__init__() got an unexpected keyword argument 'dtype'` on 4.46.1. Twelve production sites shipped with that spelling before the guard existed, breaking `soup chat`, `serve`, `merge`, `infer`, `diff` and `export` for anyone on an older stack. Your change extends that protection to class shapes the scan could not see.

Review coming shortly.

MakazhanAlpamys:
Merged — and welcome, this is your first one here.

I re-ran it rather than taking the description, in an isolated worktree with `PYTHONPATH`
pinned to it and `soup_cli.__file__` printed before trusting any result, because an
editable install otherwise resolves to the main checkout and every mutation survives
silently.

**Your mutation claim was the coarse version, and the sharper one holds.** You wrote that
removing the four new suffixes makes the four corresponding tests fail. Removing them
**one at a time** is the stronger statement, and it is true:

```
remove ForSpeechSeq2Seq        -> 1 failed, 15 passed   (speech_seq2seq)
remove ForImageTextToText      -> 1 failed, 15 passed   (image_text_to_text)
remove ForTokenClassification  -> 1 failed, 15 passed   (token_classification)
remove ForMaskedLM             -> 1 failed, 15 passed   (masked_lm)
```

Each named test is the sole guard for its own suffix. The all-four version could not have
shown that — four tests reddening together is also what you would see if they were coupled
to something shared. Worth claiming the stronger form next time; it costs three extra runs
and it is the difference between "these tests react" and "this test guards that line".

Two things I check on any widening, both of which you got right without being asked:

- The four existing false-positive controls are untouched. A guard that fires on correct
  code gets deleted, so widening the net without re-checking the negatives is how these
  die.
- All six deleted lines are the rename and reworded docstrings. Not one is a test. I check
  this every time because it has gone the other way here before.

The rename itself is the right call. `_is_dtype_guarded_model_load` describes the contract;
`_is_transformers_model_load` implied every matched class comes straight from transformers,
which was never true of the `AutoProcessor` branch.

**One scope note, in your favour rather than against it.** None of the four suffixes appears
in `src/` today and the scanner reports zero live `dtype=` hits, so this is prospective:
Whisper loads as `WhisperForConditionalGeneration`, which #497 already covered. You did not
claim otherwise, which is why it merged as-is — a PR that had oversold this as "fixes a live
bug" would have needed a rewrite of its own description. It is also well-timed rather than
speculative: concrete vision classes are precisely what #488, #501 and #502 are moving
through right now, so the net is being widened just before something walks into it.

No changelog fragment was correct here — tests only, no user-visible behaviour change, and
`changelog.d/README.md` scopes fragments to user-visible PRs.

You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md)
as of 598e64e. If you want another in the same area, #405 is open and adjacent in spirit: a
bundled eval suite sitting pinned at 1.000 detects a regression exactly as poorly as one
pinned at zero.

PR #501, opened 2026-08-23
fix(recipes): pin Qwen3.5 text-only decoder intent

### LINKED ISSUE #427: Nine Qwen3.5 recipes declare modality: text, but every checkpoint in the family reports a multimodal architecture
## What is verified

All **nine** shipped Qwen3.5 recipes leave `modality` unset, so they take the schema default `text` (`config/schema.py:3998`):

```
qwen3.5-0.8b-sft        Qwen/Qwen3.5-0.8B          modality=(unset -> text)
qwen3.5-2b-sft          Qwen/Qwen3.5-2B            modality=(unset -> text)
qwen3.5-4b-sft          Qwen/Qwen3.5-4B            modality=(unset -> text)
qwen3.5-4b-pretrain     Qwen/Qwen3.5-4B-Base       modality=(unset -> text)
qwen3.5-9b-sft          Qwen/Qwen3.5-9B            modality=(unset -> text)
qwen3.5-27b-sft         Qwen/Qwen3.5-27B           modality=(unset -> text)
qwen3.5-35b-a3b-sft     Qwen/Qwen3.5-35B-A3B       modality=(unset -> text)
qwen3.5-122b-a10b-sft   Qwen/Qwen3.5-122B-A10B     modality=(unset -> text)
qwen3.5-397b-a17b-sft   Qwen/Qwen3.5-397B-A17B     modality=(unset -> text)
```

Every one of those checkpoints reports a **multimodal** architecture on the Hub, uniformly across the family — dense and MoE alike:

```
Qwen/Qwen3.5-0.8B      -> image-text-to-text | ['Qwen3_5ForConditionalGeneration']
Qwen/Qwen3.5-4B        -> image-text-to-text | ['Qwen3_5ForConditionalGeneration']
Qwen/Qwen3.5-4B-Base   -> image-text-to-text | ['Qwen3_5ForConditionalGeneration']
Qwen/Qwen3.5-27B       -> image-text-to-text | ['Qwen3_5ForConditionalGeneration']
Qwen/Qwen3.5-35B-A3B   -> image-text-to-text | ['Qwen3_5MoeForConditionalGeneration']
```

Control, so this is a real distinction and not an artefact of how I queried:

```
Qwen/Qwen2.5-7B        -> text-generation     | ['Qwen2ForCausalLM']
```

## What is NOT verified, and I want to be explicit about it

**I have not run one of these recipes.** I am not claiming they fail. `ForConditionalGeneration` checkpoints are frequently loadable through a text-only causal-LM path, and the text path may well be exactly right here.

What I am claiming is narrower: **nine shipped recipes assert a modality that the checkpoint's own architecture contradicts, and no one has checked which of the two is wrong.** That is a claim about our evidence, not about the code.

## Why it is worth resolving rather than leaving

Two concrete places it could bite:

1. If the text path silently loads only the language tower, a user gets a run that trains and saves and looks healthy while quietly ignoring part of the model. This project has shipped that failure shape **three times** (`.inner.` keys in v0.72.0, the full-fine-tune adapter in #362, `target_modules: auto` in #392) and the recorded lesson each time is the same: a healthy loss curve structurally cannot see it.
2. The `Qwen/Qwen3.5-4B-Base` config appears to carry no top-level `vocab_size`. Any code path reading `AutoConfig.vocab_size` directly would refuse or misread it — PR #409 introduces exactly such a read in `soup draft`, and this was raised there separately.

## Acceptance criteria

- [ ] Load one Qwen3.5 checkpoint through the `modality: text` SFT path and record what is actually instantiated — full model or language tower only.
- [ ] Decide: recipes get `modality: vision`, **or** `text` is correct and the reason is written down so the next person does not re-open this.
- [ ] If `text` is correct, add a test asserting it deliberately, so the agreement is pinned rather than incidental.
- [ ] Audit for other families in the catalog whose declared modality disagrees with the Hub architecture — a one-off fix here leaves the class open.

## Not the contributor's doing

Flagged while reviewing #422, which added the ninth recipe by copying the eight that preceded it — exactly as the issue asked. This is a pre-existing family-wide question, deliberately not loaded onto a first contribution.

Amix29:
## Summary

- make `modality: text` explicit across the ten Qwen3.5 and two Qwen3.6 catalog recipes that deliberately train the language decoder
- document why the Hub's multimodal architecture and Soup's text-only recipe are both correct
- add a family-derived regression guard so the decision cannot silently fall back to the schema default

Fixes #427.

## Measurement

Measured on an Apple M4 Max with 128 GB unified memory using
`Qwen/Qwen3.5-0.8B` (Hub snapshot `2fc06364715b967f1860aea9cf38778875588b17`).
The structural loads were repeated offline after rebasing this branch onto current `main`:

| load path | instantiated class | parameters | vision parameters |
|---|---|---:|---:|
| text (`AutoModelForCausalLM`) | `Qwen3_5ForCausalLM` | 752,393,024 | 0 |
| multimodal (`AutoModelForMultimodalLM`) | `Qwen3_5ForConditionalGeneration` | 852,985,920 | 100,592,896 across 153 tensors |

The text load receives `Qwen3_5TextConfig` (`model_type: qwen3_5_text`); the full
load receives the outer `Qwen3_5Config` (`model_type: qwen3_5`). This is a real
language-tower selection, not an incidental class-name difference.

I also repeated one real Soup SFT step on four text-only ChatML rows through both Apple
paths from the rebased tree:

- MLX: 86 tokens, loss 2.8328, 6.273 GB peak allocation; the 24-tensor LoRA adapter
  reloaded over `mlx_lm.models.qwen3_5.Model` with zero vision parameters
- Transformers/MPS 5.12.1: 86 tokens, loss 0.7884, 159,744 trainable parameters;
  the 24-tensor adapter reloaded over `Qwen3_5ForCausalLM` with zero vision parameters

This establishes that `modality: text` is the correct catalog decision for text-only
training. It intentionally fine-tunes the language component, not the complete
multimodal model.

## Scope boundaries found during review

This PR does **not** lift Soup's `transformers<5.0.0` training bound. The latest permitted
release, 4.57.6, does not recognize `model_type: qwen3_5`; Transformers 5 does, but that
migration has broader compatibility failures and needs separate validation.

The Transformers 5 measurement required explicit `q_proj` / `v_proj` PEFT targets because
PEFT has no automatic mapping for `qwen3_5_text`. I deliberately kept the catalog's
`target_modules: auto` unchanged here: MLX resolves that value to the backend-specific
full keys `self_attn.q_proj` / `self_attn.v_proj`, while recipe-level short keys would match
no MLX modules. Dependency support and a cross-backend target policy are separate follow-ups,
not safe additions to this modality PR.

## Catalog audit

I parsed all 146 recipes (111 unique model IDs) and compared their declared/default
modality with current Hub metadata. Ninety-nine model IDs were queryable; twelve returned
404 and are an explicit audit limitation.

Other Hub-multimodal families currently taking the implicit text default include Gemma 3 /
MedGemma, Llama 4, Kimi K2.5/K2.6, and MiniMax M3. They need family-specific intent checks
rather than a blanket switch to vision. The three Whisper ASR recipes also leave modality
unset despite their ASR Hub metadata. Qwen3.6 is included in this PR because its checkpoints
report the same `qwen3_5` / `qwen3_5_moe` architecture and the same measured decoder-only
decision applies.

## Regression proof

The test derives the covered recipe names from the catalog and separately pins the audited
set. Removing `modality: text` from only `qwen3.5-0.8b-sft` produced one named failure while
the other eleven cases stayed green; restoring it returned the file to 13/13 passing.

## Validation

- focused issue + recipe suites: **1,241 passed**
- full suite: **18,358 passed, 285 skipped, 5 deselected** (81.30% coverage)
- Ruff: clean across `src/soup_cli/`, `tests/`, and `scripts/`
- `git diff --check`: clean
- two real one-step Apple Silicon runs and both adapter reloads: passed

### DIFF
--- a/changelog.d/0.73.3/501.changed.md
+++ b/changelog.d/0.73.3/501.changed.md
@@ -1,0 +1,5 @@
+- **Qwen3.5 and Qwen3.6 text recipes now state their decoder-only intent (#427 by
+  @Amix29 in #501).** Apple Silicon measurements confirmed that Soup's text path loads
+  `Qwen3_5ForCausalLM` without the visual tower, so all twelve catalog recipes now set
+  `modality: text` explicitly and regression coverage prevents that decision from falling
+  back to the schema default.
--- a/docs/models.md
+++ b/docs/models.md
@@ -32,6 +32,14 @@ Soup works with **any** of the **340,000+** text-generation models on [HuggingFa
 | **InternLM 3** | InternLM3-8B-Instruct | 8B | Chinese + English |
 | **Falcon** | Falcon-11B, Falcon-40B-Instruct | 7B–180B | Open-weight |
 
+Qwen3.5 and Qwen3.6 checkpoints advertise a multimodal conditional-generation
+architecture on the Hub, but Soup's catalog recipes are deliberately text-only.
+Their explicit `modality: text` selects `AutoModelForCausalLM`, which instantiates the
+language decoder without the visual tower. This is appropriate for text-only SFT,
+pre-training, and GRPO data; it is not a full multimodal fine-tune. The Transformers
+backend cannot load `qwen3_5` under Soup's current `<5.0.0` dependency cap; enabling
+that backend requires a separately validated Transformers 5 migration.
+
 ### Vision Models (with `modality: vision`)
 
 | Model | Size | Supported Formats |
--- a/src/soup_cli/recipes/catalog.py
+++ b/src/soup_cli/recipes/catalog.py
@@ -3749,6 +3749,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.5-0.8B
 task: sft
+modality: text
 
 data:
   train: ./data/train.jsonl
@@ -3777,6 +3778,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.5-2B
 task: sft
+modality: text
 
 data:
   train: ./data/train.jsonl
@@ -3805,6 +3807,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.5-4B
 task: sft
+modality: text
 
 data:
   train: ./data/train.jsonl
@@ -3833,6 +3836,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.5-4B-Base
 task: pretrain
+modality: text
 
 data:
   train: ./data/corpus.jsonl
@@ -3862,6 +3866,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.5-9B
 task: sft
+modality: text
 
 data:
   train: ./data/train.jsonl
@@ -3890,6 +3895,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.5-9B
 task: grpo
+modality: text
 
 data:
   train: ./data/reasoning_train.jsonl
@@ -3922,6 +3928,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.5-27B
 task: sft
+modality: text
 
 data:
   train: ./data/train.jsonl
@@ -3951,6 +3958,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.5-35B-A3B
 task: sft
+modality: text
 
 data:
   train: ./data/train.jsonl
@@ -3985,6 +3993,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.5-122B-A10B
 task: sft
+modality: text
 
 data:
   train: ./data/train.jsonl
@@ -4020,6 +4029,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.5-397B-A17B
 task: sft
+modality: text
 
 data:
   train: ./data/train.jsonl
@@ -4052,6 +4062,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.6-27B
 task: sft
+modality: text
 
 data:
   train: ./data/train.jsonl
@@ -4081,6 +4092,7 @@ output: ./output
         yaml_str="""\
 base: Qwen/Qwen3.6-35B-A3B
 task: sft
+modality: text
 
 data:
   train: ./data/train.jsonl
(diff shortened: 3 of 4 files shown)


MakazhanAlpamys:
Merged as 540c310. Verified rather than read.

Mutation in an isolated worktree with `soup_cli.__file__` confirmed to resolve there —
removing the single `modality: text` line from `qwen3.5-0.8b-sft` gives:

```
FAILED test_qwen35_family_recipes_explicitly_select_language_tower[qwen3.5-0.8b-sft]
1 failed, 12 passed
```

Eleven siblings green, so the coupling is per-recipe rather than per-file. And your test
is the **only** guard that can exist here: `SoupConfig.modality` already defaults to
`"text"`, so a generic "every recipe loads" test structurally cannot see a missing
declaration. Worth naming, because that is also the reason this was invisible for as long
as it was.

I checked the dispatch claim in the source rather than accepting it —
`trainer/sft.py:1166` for `AutoModelForCausalLM` under `modality == "text"`, `:1435` for
`AutoModelForVision2Seq` under `"vision"`. It holds.

**The scope section is the part I want to point at.** A `fix(recipes):` commit that says
in its own body that it changes no runtime behaviour today, and that transformers 4.x
cannot load `model_type: qwen3_5` at all, is volunteering the one fact that would let a
reviewer call it mislabelled. It merged as-is *because* of that, not in spite of it. That
is the same instinct as your #490 work, where you disproved the recommendation you had
been handed before implementing it.

## One docs thing, and my reviewer's framing of it was too strong

The review that came back flagged `docs/models.md` as self-contradicting: the
"Recommended Models" table thirteen lines above your new paragraph still lists Qwen 3.5 /
3.6 with `262K context, native vision, MoE`. Reading it myself, **that is not a
contradiction** — your paragraph opens by acknowledging exactly that ("checkpoints
advertise a multimodal conditional-generation architecture on the Hub, **but** Soup's
catalog recipes are deliberately text-only"), so it explains the row rather than fighting
it. I am recording the correction rather than passing the finding on as stated.

The residual risk is narrower and still real: "native vision" sits in a table about what
*Soup* supports, and there is no vision recipe for this family, nor a backend that can
load it. A reader scanning the table does not necessarily reach line 13.

Not worth another push here. It belongs in #502 — you are already in that file's
neighbourhood, and once the v5 migration lands the honest row changes anyway.

## On #502 and #503

Nothing here blocks that work, and #503 now carries the dependency half: the short version
is that **trl 0.28.0 is the broken one and 0.29.0 already vendors its own helper**, so if
your suite run is still going against 0.28.0 it is measuring an upstream defect that has
been fixed. Full detail in the comment on #502.

You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md)
as of c38da01 — seventh merge.

PR #504, opened 2026-08-23
docs(deps): record why the transformers <5.0.0 ceiling is load-bearing

### LINKED ISSUE #503: The transformers <5.0.0 ceiling is load-bearing: trl's optional-import guards invert on transformers 5 (45 failures, none in Soup)
`pyproject.toml` declares `transformers>=4.36.0,<5.0.0` in the `[train]` extra. That
ceiling was set on 2026-05-08, **before transformers 5 existed**, so it reads like
plain caution — and it was nearly lifted on exactly that reasoning. It is not caution.
It is load-bearing, and this issue records the measurement so the next person does not
spend the 46-minute suite run to find out.

> **Correction, 2026-08-23.** The first version of this issue said *"trl 1.10.0 is
> byte-identical, so no trl release fixes it."* **That is false**, and it was the
> load-bearing sentence. Checked by downloading wheels rather than by reading one
> version: trl **0.29.0** vendors its own `_is_package_available` and is immune. The
> real obstacle is a chain, not a missing upstream fix. Corrected throughout below.

## Measurement

Full suite against `transformers==5.15.1`: **45 failed, 18205 passed.** Not one failure
is in Soup's own code.

## Root cause

transformers 5 changed the return type of a **private** helper that `trl` imports:

```
transformers 4.57.6:  _is_package_available("weave") -> False         # bool, falsey
transformers 5.15.1:  _is_package_available("weave") -> (False, None) # tuple, TRUTHY
```

The tuple is returned for present packages too (`_is_package_available("numpy")` ->
`(True, None)`), so the inversion fires specifically on *absent* packages: a
"not installed" answer becomes truthy.

`trl/import_utils.py` imports that private helper and derives its optional-dependency
flags from it:

```python
from transformers.utils.import_utils import _is_package_available   # private

_deepspeed_available = _is_package_available("deepspeed")
_vllm_available      = _is_package_available("vllm")
# ... 13 module-level flags in trl 0.19.1, evaluated at import time
```

The annotation promises `bool`; a non-empty tuple arrives. `if (False, None):` is true,
so the guarded `import weave` runs unconditionally and raises `ModuleNotFoundError`.
`dpo_trainer` and `grpo_trainer` do not import behind it, and 45 tests fall over.
`weave` is simply the first flag to fire.

## The real obstacle is a chain, not an unfixed upstream

Measured by downloading each wheel and grepping `trl/import_utils.py`:

| trl | imports the transformers private helper | vendors its own | on transformers 5 |
|---|---|---|---|
| 0.28.0 | yes | no | **breaks** |
| 0.29.0 | no | yes | immune |
| 1.0.0 | no | yes | immune |
| 1.10.0 | no | yes | immune |

trl 0.29.0 defines `_is_package_available` itself over `importlib.util.find_spec`, so
the defect is **already fixed upstream** — exactly one version above Soup's own
`trl>=0.14.0,<0.29` cap. And that cap is held by a *separate* break, already documented
in `pyproject.toml`: trl 0.29 moved `BasePairwiseJudge` to `trl.experimental.judges`, so
`eval/judge.py::_base_pairwise_judge_cls` cannot import it, and trl 1.x drops `judge=`
entirely.

So the shape is:

```
transformers<5  <-  trl<0.29  <-  the BasePairwiseJudge import
```

**Lifting any single link achieves nothing.** Raising the transformers ceiling alone
breaks trl; raising the trl cap alone breaks `online_dpo` and the judge evaluator.

## Reproduction (2 minutes, no suite run)

```bash
python -m venv v5 && v5/bin/pip install "trl==0.28.0" "transformers>=5"
v5/bin/python -c "import trl.trainer.dpo_trainer"
# ModuleNotFoundError: No module named 'weave'
```

Mechanism alone, without trl or torch:

```bash
python -m venv p5 && p5/bin/pip install "transformers>=5"
p5/bin/python -c "
from transformers.utils.import_utils import _is_package_available as f
r = f('weave'); print(repr(r), type(r).__name__, bool(r))"
# (False, None) tuple True
```

## Consequence for the `[mlx]` extra

`mlx-lm` itself requires `transformers>=5.0.0`. Soup's `[mlx]` extra only restates that
upstream requirement — deleting the line does not help, because the transitive
requirement remains. So:

```bash
pip install "soup-cli[train,mlx]"   # ResolutionImpossible
```

There is no side to concede today: mlx-lm needs transformers 5, transformers 5 breaks
the trl range Soup can use, and trl is required to train. `docs/backends-and-ops.md` now
states this explicitly rather than leaving the user with pip's resolver output.

## Acceptance criteria for lifting the ceiling

These must land **together**; each is useless alone.

- [ ] `eval/judge.py` resolves `BasePairwiseJudge` from `trl.experimental.judges` with a
      fallback to the public namespace, in the shape `trainer/_trl_compat.py` already
      uses (probe the capability, do not table the version).
- [ ] The trl cap is raised past `<0.29`, and the surfaces that bump did not exercise
      are validated: `serve`, GRPO variants, PPO with a real reward model, `online_dpo`.
- [ ] With that trl, `import trl.trainer.dpo_trainer` and `import trl.trainer.grpo_trainer`
      succeed against transformers 5 in a clean venv with no optional extras.
- [ ] Full suite re-run against transformers 5 comes back green, and the number is
      recorded here.
- [ ] `tests/test_transformers_ceiling_compat.py` updated in the same PR as the bound.

Only after all of that does `pip install "soup-cli[train,mlx]"` become resolvable.

## Related

#323 / #486 — a weekly dependency-drift job. `trl` declares `transformers>=4.56.2`
with **no upper bound**, so a fresh `pip install trl` alongside transformers 5 is broken
out of the box for any trl below 0.29. Only a run against the newest stack catches that
class of breakage; this is a live instance of it that arrived while the PR was waiting.

MakazhanAlpamys:
Closes #503.

The `train` extra caps `transformers<5.0.0`. The bound was set 2026-05-08, **before
transformers 5 existed**, so it reads like plain caution — and it was nearly lifted on
exactly that reasoning. It is not caution. This PR **records** rather than repairs.

> **One reading was withdrawn while writing this, and it was the load-bearing one.**
> The first push said *"trl 1.10.0 is byte-identical, so no trl release fixes it."*
> False. Checked by downloading the wheels instead of reading one version — see the
> table below. The obstacle is a **chain**, not an unfixed upstream, and that is a
> materially different instruction to whoever picks this up next.

## The measurement

Full suite against `transformers==5.15.1`: **45 failed, 18205 passed.** Not one failure
in Soup's own code.

transformers 5 changed the return type of a **private** helper that `trl` imports:

```
4.57.6:  _is_package_available("weave") -> False         # bool, falsey
5.15.1:  _is_package_available("weave") -> (False, None) # tuple, TRUTHY
```

The tuple comes back for present packages too (`numpy` → `(True, None)`), so the
inversion fires specifically on **absent** ones: "not installed" reads as true.
`trl/import_utils.py` derives its optional-dependency flags from that helper, so the
guarded `import weave` runs anyway and `dpo_trainer` / `grpo_trainer` never import
behind the `ModuleNotFoundError`. Thirteen module-level flags in trl 0.19.1 are built
this way; weave merely fires first.

## The obstacle is a chain

| trl | imports the private helper | vendors its own | on transformers 5 |
|---|---|---|---|
| 0.28.0 | yes | no | **breaks** |
| 0.29.0 | no | yes | immune |
| 1.0.0 / 1.10.0 | no | yes | immune |

trl 0.29.0 vendors `_is_package_available` over `importlib.util.find_spec`, so upstream
is **already fixed** — one version above Soup's own `trl<0.29` cap. That cap is held by
a separate break already documented in `pyproject.toml`: `BasePairwiseJudge` moved to
`trl.experimental.judges`, so `eval/judge.py` cannot import it.

```
transformers<5  <-  trl<0.29  <-  the BasePairwiseJudge import
```

Lifting any single link achieves nothing. #503 carries the acceptance criteria that
have to land together.

## Why `[mlx]` and `[train]` cannot co-install

`mlx-lm` itself requires `transformers>=5.0.0`; the `[mlx]` extra only restates that, so
deleting the line does not help. Until now the user got pip's `ResolutionImpossible` and
nothing explaining it, and the docs said only "do not add `[train]`" — a preference, not
a fact.

## What changed

- **`pyproject.toml`** — the measurement and the chain now sit next to the bound. A bare
  bound is what gets lifted as a precaution.
- **`docs/backends-and-ops.md`** — states the incompatibility, both causes, the chain,
  and that separate virtualenvs are the answer.
- **`tests/test_transformers_ceiling_compat.py`** — two *different* kinds of guard, and
  the distinction is the point:
  - `TestCeilingIsPinned` is a **ratchet**. It fails on purpose when the bound moves,
    and names what to re-measure first.
  - `TestDocsMatchTheDeclaredBounds` is a **consistency check, not a ratchet**. It
    asserts the docs warn *exactly when* the two extras are provably disjoint, and
    **computes** that disjointness from the declared specifiers instead of pinning a
    string. If upstream is repaired it does not block the repair — it tells you to
    delete the now-false warning. A guard that fires on correct code gets deleted, so
    this one is written not to.

## Mutation results (target file only)

| mutation | outcome |
|---|---|
| ceiling `<5.0.0` → `<6.0.0` | ratchet fails **and** the consistency check flips to "delete the warning" — proving it computes rather than asserting a constant |
| ceiling lifted **+** docs updated (a legitimate repair) | **only** the ratchet fires (1 failed, 3 passed); the consistency check stays quiet |
| strip `#503` from `pyproject.toml` | `test_the_ceiling_carries_its_reason_in_pyproject` fails |
| delete the docs warning while the extras are disjoint | `test_docs_warn_exactly_when_the_extras_cannot_co_install` fails |

Each of the four tests is the sole guard for its fact. The neighbouring
`test_transformers_floor_compat.py` still extracts `4.36.0` with the new comment inside
the `train` block, and the recipe / CLI-help count ratchets are unaffected.

**No changelog fragment.** Behaviour is unchanged and the incompatibility already
existed; `changelog.d/README.md` scopes fragments to user-visible changes and has no
documentation category.

## Related

#323 / #486 — `trl` declares `transformers>=4.56.2` with **no upper bound**, so a fresh
`pip install trl` next to transformers 5 is broken out of the box for any trl below
0.29. Only a run against the newest stack catches that class of breakage. This is a live
instance of it that arrived while that PR was still waiting.

### DIFF
--- a/docs/backends-and-ops.md
+++ b/docs/backends-and-ops.md
@@ -88,6 +88,30 @@ Transformers backend. A Hugging Face `datasets` source or streaming dataset
 still needs `datasets` because that data source owns the dependency; the local
 file path below does not.
 
+> **`[mlx]` and `[train]` cannot be installed together, and that is not a style
+> preference.** `pip install "soup-cli[train,mlx]"` fails with pip's
+> `ResolutionImpossible`, whose output says nothing about the actual cause:
+>
+> - `mlx-lm` requires `transformers>=5.0.0`. The `[mlx]` extra only restates that
+>   upstream requirement, so deleting the line does not help -- the transitive
+>   requirement remains.
+> - `[train]` caps `transformers<5.0.0`, and the cap is load-bearing. On
+>   transformers 5 a private helper that `trl` imports returns a truthy tuple where
+>   it used to return `False`, so `trl`'s optional-import guards invert and
+>   `dpo_trainer` / `grpo_trainer` stop importing. Measured on transformers 5.15.1:
+>   45 test failures, none of them in Soup's own code.
+> - `trl` itself is fixed from **0.29.0**, which vendors its own copy of the helper --
+>   but that is one version above Soup's `trl<0.29` cap, and that cap is held by an
+>   unrelated break (`BasePairwiseJudge` moved to `trl.experimental.judges`). The
+>   chain is `transformers<5` <- `trl<0.29` <- the judge import, so lifting any one
+>   link on its own changes nothing.
+>
+> There is no side to concede -- `mlx-lm` needs transformers 5, transformers 5 breaks
+> `trl`, and `trl` is what trains. Install one extra or the other; use separate
+> virtualenvs if you need both on one machine. The full measurement and the criteria
+> for lifting the cap are in
+> [#503](https://github.com/MakazhanAlpamys/Soup/issues/503).
+
 `detect_device()` and `get_gpu_info()` recognise Apple Silicon when
 `backend: mlx` is set, preserving `training.quantization: 4bit` for
 `mlx-community` pre-quantized checkpoints instead of silently downgrading to
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,30 @@ dependencies = [
 # These were core dependencies through v0.70.0; pins are unchanged.
 train = [
     "torch>=2.0.0",
+    # THE <5.0.0 CEILING IS LOAD-BEARING. Do not lift it as a precaution: it was set
+    # 2026-05-08, before transformers 5 existed, which is exactly why it reads like one.
+    # Measured against transformers 5.15.1 -- 45 tests fail, not one of them in Soup.
+    #
+    # transformers 5 changed the return type of a PRIVATE helper that trl imports:
+    #     4.57.6:  _is_package_available("weave") -> False         (bool, falsey)
+    #     5.15.1:  _is_package_available("weave") -> (False, None) (tuple, TRUTHY)
+    # The tuple comes back for present packages too, so the inversion fires on ABSENT
+    # ones: "not installed" reads as true. trl/import_utils.py derives its optional
+    # dependency flags from that helper, so the guarded `import weave` runs anyway and
+    # dpo_trainer / grpo_trainer never import behind the ModuleNotFoundError. 13
+    # module-level flags in trl 0.19.1 are built this way; weave merely fires first.
+    #
+    # The trl boundary is exact, and it is why this bound cannot be lifted alone:
+    # trl 0.28.0 still imports the private helper; trl 0.29.0 vendors its own
+    # _is_package_available (importlib.util.find_spec) and is immune. That is ONE
+    # version above Soup's own `trl<0.29` cap. So the chain is
+    # transformers<5 <- trl<0.29 <- the BasePairwiseJudge import documented below,
+    # and lifting any single link achieves nothing.
+    #
+    # Consequence: [mlx] needs transformers>=5.0.0 -- mlx-lm's own requirement, only
+    # restated in that extra -- so [train] and [mlx] cannot co-install, and there is no
+    # side to concede. Full measurement and the acceptance criteria for lifting this
+    # bound live in issue #503. Guarded by tests/test_transformers_ceiling_compat.py.
     "transformers>=4.36.0,<5.0.0",
     "peft>=0.7.0",
     # Capped below 0.29. #326 raised this from `<0.27` after migrating the six
--- a/tests/test_transformers_ceiling_compat.py
+++ b/tests/test_transformers_ceiling_compat.py
@@ -1,0 +1,172 @@
+"""#503 - the Transformers ``<5.0.0`` ceiling is load-bearing, not caution.
+
+``pyproject.toml`` caps ``transformers<5.0.0`` in the ``train`` extra. The bound was
+set on 2026-05-08, *before* transformers 5 existed, so it reads like a precaution --
+and it was nearly lifted on exactly that reasoning. It is not a precaution.
+
+Measured against ``transformers==5.15.1``: 45 tests fail, none of them in Soup's own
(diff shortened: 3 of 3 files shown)


MakazhanAlpamys:
Standing down in favour of #507, which does the migration this PR documents as blocked — and cites #503's measurement while doing it.

Converted to draft rather than closed: if #507 lands, this closes with it; if it needs more time, this rebases so the current state stays documented in the meantime. Either way it is not sitting across two files @Amix29 needs.

The measurement itself is safe — it lives in #503, not here. The one piece worth surviving in either direction is the extras-consistency check, which is not a ratchet and does not block making the extras compatible; offered to #507 to absorb, otherwise re-opened after.

MakazhanAlpamys:
﻿Closing per the standing-down note: the migration this documented as blocked landed as #507 with the ceiling moved deliberately (transformers >=5.12.1,<6.0.0), which retires both the blocker and the documentation gap it described. The measurement record you cited remains linked from #503.

PR #506, opened 2026-08-23
fix(mcp): block restart capacity while a launch is unresolved

### LINKED ISSUE #505: MCP one-active-execution cap does not survive a crash between Popen() and mark_running()
## What happened

`ExecutionManager.execute()` (`src/soup_cli/mcp_server/execution.py`) writes a run's state in three steps: `tracker.launch_run()` inserts a `status='launching'` row with `pid=NULL`, `subprocess.Popen()` spawns the child, then `tracker.mark_running()` upgrades the row to `status='running'` with the real pid. If the MCP server process crashes between the second and third step, the row is stuck at `status='launching'` forever: it has no pid, so neither `_reconcile_orphaned_run` (issue #401) nor `_live_persisted_run` (issue #402) can ever check its liveness. `_live_persisted_run` explicitly skips any row whose status is not `'running'`.

On restart, a fresh `ExecutionManager` sees free capacity and accepts a second `execute()` call, even though the first child (spawned but never recorded) may still be alive and training, which is exactly the double-booking #402 was written to close.

## Reproduction

```python
t = ExperimentTracker()
t.launch_run(run_id="a", kind="train", config_dict={}, command_digest="d", log_path="/tmp/x.log")
# crash point: mark_running() never runs

manager = ExecutionManager()          # a restart: fresh in-memory state
assert manager._live_persisted_run() is None   # should see a possible orphan; does not

token = manager.issue(kind="train", argv=["true"], display_command="x")
manager.execute(token=token, kind="train")      # succeeds: a second job starts anyway
```

Confirmed live against current `main` (`4c23aff`): the row stays `status='launching'` across every read, and a second `execute()` after the simulated restart succeeds instead of raising.

## Fix

Opening a PR: `_live_persisted_run` now treats any `'launching'` row as unconditionally active (no pid exists to check, so the safe default is to assume the child may still be running and refuse new work until a human clears the record), instead of skipping it. `'running'` rows are unaffected and keep the existing pid-liveness check. This does not change `launch_run`/`Popen`/`mark_running` ordering, and it leaves the row visibly `'launching'` rather than guessing at an outcome it cannot verify.

AmirF194:
## Root cause

`ExecutionManager.execute()` writes a run's state in three steps: `tracker.launch_run()` inserts a `status='launching'` row with `pid=NULL`, `subprocess.Popen()` spawns the child, then `tracker.mark_running()` upgrades the row to `status='running'` with the real pid. `_live_persisted_run` (the restart-survival check from #402) only inspects rows whose status is `'running'`, so it silently skips one stuck at `'launching'`.

If the server crashes between the second and third step, that row never gets a pid and never transitions, so `_live_persisted_run` reads it as free capacity on restart. A restarted server then accepts a second `execute()` call while the first child, spawned but never recorded, may still be alive and training on the same GPU.

## Fix

`_live_persisted_run` now treats a `'launching'` row as unconditionally active instead of skipping it: there is no pid to check liveness against, so refusing new work until a human clears the record is the safe default. `'running'` rows are unaffected and keep the existing pid-liveness check. `launch_run`/`Popen`/`mark_running` are untouched, and a stuck row still reads as `'launching'` rather than being guessed at as `'terminated'`.

## Verification

- New regression tests (`tests/test_issue505_launching_window_blocks_restart.py`) model the crash by calling `launch_run()` and never calling `mark_running()`: they fail on `main` (`_live_persisted_run()` returns `None`, a second `execute()` succeeds) and pass on this branch, confirmed both ways in the same container.
- `tests/test_issue401_orphaned_run_reconcile.py`, `test_issue402_restart_capacity.py`, `test_mcp_execute.py`, and `test_issue296_network_transport.py` still pass unmodified (50 passed, 46 GPU-skipped).
- `ruff check` and `mypy` on the changed file are clean.
- Not run: the `windows-latest`/`macos-latest` legs and the `torch`/`trl` (`[dev]`) install, since this VPS has neither; the change touches no platform- or training-stack-specific code.

Fixes #505


MakazhanAlpamys:
Merged as squash, CI 13/13 on your head after the branch update. Both review findings were real and both are handled: the docs gap I landed as a maintainer paragraph in docs/commands.md next to the capacity contract (naming your PR), and the stuck-launching recovery surface I filed as #524 rather than widening your diff - blocking there is the safe direction, so it merged without it. What review valued most: extending _live_persisted_run instead of adding a parallel mechanism, and the test that pins reconcile-on-read leaving launching rows alone - that assertion is what stops a future cleanup from silently re-opening #401. Thank you.

AmirF194:
Appreciate the detailed writeup and for landing the docs paragraph yourself, that split of scope makes sense to me too.

PR #507, opened 2026-08-23
fix(train): support Qwen3.5 on Transformers 5

### LINKED ISSUE #502: Support Qwen3.5/Qwen3.8 text training by migrating the Transformers backend to v5
## Problem

Soup currently has two incompatible dependency contracts:

- `[train]` requires `transformers>=4.36,<5`
- `[mlx]` requires `transformers>=5`

The combined training and Apple extra is therefore unsatisfiable. More importantly for the model catalog, `Qwen/Qwen3.8-27B` declares `model_type: qwen3_5`. The latest Transformers release allowed by `[train]` does not recognize that architecture, while Transformers 5 exposes the intended text-only `Qwen3_5ForCausalLM` path.

This is now a concrete blocker rather than a hypothetical dependency cleanup: #427 established that text-only decoder loading is the correct modality decision, #501 pins that decision in the recipes, and #477 asks for the Qwen3.8-27B catalog entry.

A measured one-step Transformers 5.12.1 / PEFT 0.20.0 run on Apple Silicon succeeded for the smaller Qwen3.5 family checkpoint only after using explicit LoRA targets. PEFT does not currently have an automatic target mapping for `qwen3_5_text`, so raising the dependency bound alone is not sufficient.

## Proposed fix

1. Move the supported Transformers training stack to a v5 release that recognizes `qwen3_5`.
2. Repair the shared API compatibility failures exposed by the full suite under that stack rather than bypassing them.
3. Add a Qwen3.5-family PEFT target policy for the Transformers backend without changing the MLX-specific full target keys.
4. Pin a regression test that text mode selects the causal language model and no vision parameters.
5. Keep the Qwen3.8 catalog addition itself in #477 so this migration remains a dependency/backend correction.

## Acceptance criteria

- [ ] `pip install "soup-cli[train,mlx]"` has a satisfiable Transformers requirement.
- [ ] A Qwen3.5-family text checkpoint resolves through `AutoModelForCausalLM` / `Qwen3_5ForCausalLM` under the supported stack.
- [ ] Default Transformers LoRA setup for `qwen3_5_text` attaches non-zero trainable adapter parameters and saves a reloadable adapter.
- [ ] The MLX target-module contract remains unchanged.
- [ ] Focused Qwen/PEFT tests and the full non-smoke suite pass under the declared dependency set.
- [ ] The dependency and compatibility change is documented.

I am taking this issue. I am currently running the complete suite against Transformers 5.12.1, PEFT 0.20.0 and TRL 0.28.0 to derive the smallest honest compatibility patch before opening the PR.

### LINKED ISSUE #503: The transformers <5.0.0 ceiling is load-bearing: trl's optional-import guards invert on transformers 5 (45 failures, none in Soup)

### LINKED ISSUE #522: plotext 6.0 removed the classic plotting API: soup data stats / runs replay crash on a fresh install
The unpinned core dep \plotext>=5.2.0\ started resolving **plotext 6.0.0** (released upstream). Measured in a clean venv: \hasattr(plotext,'clear_figure')\ is False — and so are \hist\, \	heme\, \plot\. The whole imperative API both call sites use is gone.

**Blast radius (measured, CI run [32645325022](https://github.com/MakazhanAlpamys/Soup/actions/runs/32645325022)):** all 9 matrix cells fail the same 3 tests — \soup data stats\ raises \AttributeError: module 'plotext' has no attribute 'clear_figure'\ (\commands/data.py:622\) and \soup runs replay\/\
uns show\ charting would fail identically (\commands/runs.py:558\). Local boxes stay green only when plotext is absent entirely (ImportError -> histogram silently skipped), which is why this was invisible until the dependency-drift job fired as designed (#486).

**Decision:** cap \plotext>=5.2.0,<6.0.0\ on main as the immediate repair. #507 (@Amix29) carries a compatibility helper exercised against real runtime for BOTH plotext 5.x and 6.x; when it merges it should lift this cap deliberately, not leave a stale bound.

**Acceptance criteria**
- [ ] fresh \pip install soup-cli\ cannot resolve plotext >= 6
- [ ] CI matrix green again
- [ ] cap lifted or re-affirmed by #507's dual-support layer

Amix29:
## Summary

- make the `[train]` and `[mlx]` extras co-installable on `transformers>=5.12.1,<6`, with `trl>=0.29,<1` and `peft>=0.20,<1`
- migrate the affected Transformers/TRL call sites and resolve moved TRL symbols through the existing capability helper
- add a Transformers LoRA target policy for the Qwen3.5 hybrid text decoder while leaving explicit targets and MLX defaults unchanged
- pin text-mode loading to `Qwen3_5ForCausalLM`, assert that no vision parameters are instantiated, and save/reload a non-empty adapter
- materialize TRL 0.29's frozen DPO `ref` adapter when layer streaming creates it on `meta`, preserving the one-model memory property
- restore the removed DPO/ORPO prompt cap on TRL's prepared token ids, so `data.max_length` remains an effective model-input bound
- keep the backend documentation and the declared `[train]`/`[mlx]` Transformers ranges linked by a consistency test
- support both Plotext 5's module API and Plotext 6's Figure API, then deliberately lift main's temporary `<6.0.0` safety cap
- keep `soup doctor` synchronized with the validated Transformers/TRL/PEFT compatibility band through a metadata-derived regression test

## Why 5.12.1

Transformers 5.2 contains Qwen3.5 symbols, but its auto-model mapping still passes the outer `Qwen3_5Config` into the text decoder and fails because that wrapper has no direct `vocab_size`. The first tested floor where `AutoModelForCausalLM.from_config(Qwen3_5Config(...))` selects a working text-only `Qwen3_5ForCausalLM` is 5.12.1.

TRL 0.29 is paired with that floor because it vendors its optional-package probe instead of importing Transformers' private helper, and the moved pairwise-judge/Online DPO APIs are now resolved from their experimental homes when absent from the public namespace.

## Measured TRL 0.29 follow-up

- **Online DPO was exercised, not only imported.** On the exact floor (`transformers 5.12.1`, `trl 0.29.0`, `peft 0.20.0`), a real Apple Silicon CPU training run generated completions, called the pairwise judge, completed an optimizer step, and logged `rewards/chosen`, `rewards/rejected`, `rewards/accuracies`, and `rewards/margins`.
- **The streamed DPO reference finding was measured.** Before materialization, every `.ref.` LoRA parameter created by TRL was on `meta`; even the zero-`lora_B` control produced a maximum chosen-logp delta of `0.0028839111328125`. After materialization, the frozen reference parameters are non-meta, the zero-adapter control is exactly `0.0`, and a randomized policy adapter differs from the reference by more than the pinned `1e-4` threshold.
- **The prompt caveat reproduced on 0.29.** A 161-token prompt with `data.max_length: 64` reached the model as 164 tokens in DPO and 165 in ORPO. The follow-up caps TRL's already-rendered token ids (including conversational rows), retains all 8 test rows, and asserts every effective input sequence is at most 64 tokens.

## Plotext 6 follow-up

Main temporarily capped Plotext below 6 in `eb77dd1` after the upstream release removed the module-level plotting API and broke all nine CI matrix cells (#522). This branch routes Plotext 5 through the classic module API and Plotext 6 through its Figure API. After rebasing onto that repair, this PR deliberately lifts the cap again, updates #522's changelog entry, and retains the contract coverage for both paths.

The dedicated compatibility cell now pins the real `plotext==6.0.0`, asserts that installed version, and executes histogram and line rendering against the installed package. The ordinary fake-API tests remain because they independently pin both the 5.x and 6.x dispatch contracts.

## Review follow-up

- rebased onto current `main` through `6580418`, including #471 and #509
- synchronized `soup doctor` with the declared `transformers>=5.12.1,<6`, `trl>=0.29,<1`, and `peft>=0.20,<1` ranges
- added a regression that derives those bounds from `pyproject.toml`, so the doctor table cannot silently drift again
- added explicit new-floor skip guidance for runtime tests collected in a pre-upgrade environment
- updated #471's vision loader doubles to the Transformers 5 `AutoModelForImageTextToText` name after the rebase exposed the stale fixtures

## Validation

Run locally on macOS arm64:

- current latest stack (`transformers 5.15.1`, `trl 0.29.1`, `peft 0.20.0`, `plotext 6.0.0`): `18534 passed, 283 skipped, 5 deselected`; coverage `81.29%`
- exact compatibility cell (`transformers 5.12.1`, `trl 0.29.0`, `peft 0.20.0`, `plotext 6.0.0`): `29 passed`
- post-rebase review and related coverage: `230 passed`
- real Plotext 5.3.2 runtime plus metadata guards: `5 passed`
- focused latest-stack TRL/Qwen/streaming matrix: `112 passed, 19 skipped`
- real Online DPO smoke on both tested stacks: `1 passed, 1 skipped` (the skip is the inapplicable TRL 1.x branch)
- combined `[dev,mlx]` dependency check: clean on both tested stacks
- `ruff check src/soup_cli scripts tests/`
- `actionlint`
- `git diff --check`

Fixes #502
Fixes #503
Closes #522

### DIFF
--- a/.github/constraints/transformers-floor.txt
+++ b/.github/constraints/transformers-floor.txt
@@ -1,7 +1,10 @@
-# #502/#503 — exact pins for the dedicated dependency-floor CI job.
+# #502/#503/#522 — exact pins for the dedicated dependency-floor CI job.
 # Qwen3.5/Qwen3.8 text loading establishes Transformers 5.12.1 as Soup's
 # declared training floor. trl 0.29 is the first release that does not import the
 # Transformers private availability helper whose v5 return type broke trl 0.28.
+# Plotext 6.0.0 is pinned here so one CI cell exercises the real Figure API rather
+# than relying only on the fake 5.x/6.x contract doubles in the unit suite.
 transformers==5.12.1
 trl==0.29.0
 peft==0.20.0
+plotext==6.0.0
--- a/.github/constraints/transformers-floor.txt
+++ b/.github/constraints/transformers-floor.txt
@@ -1,23 +1,7 @@
-# #478 — Transformers pin for the dedicated floor-compat CI job.
-#
-# Declared train floor in pyproject.toml remains:
-#   transformers>=4.36.0,<5.0.0
-#
-# transformers==4.36.0 is not resolvable with the declared trl range
-# (trl>=0.14.0,<0.29): even trl 0.14.0 requires transformers>=4.46.0, and
-# newer trl releases require >=4.56.1 / >=4.56.2. Raising Soup's declared
-# transformers floor in pyproject.toml needs an explicit maintainer
-# dependency-policy decision. Do NOT treat 4.55.4 (the last pre-dtype=
-# release) as the declared floor.
-#
-# This file pins:
-#   transformers==4.46.1 — lowest non-yanked resolvable Transformers version
-#     tested with the lowest declared TRL version
-#   trl==0.14.0 — lowest version in Soup's declared trl range
-#
-# These pins are for the floor CI cell only so pip cannot silently upgrade
-# that job. Do not describe pip's preferred/latest compatible TRL as the
-# floor. Runtime dtype= regressions are caught by the static guard in
-# tests/test_transformers_floor_compat.py (dtype= is >=4.56-only).
-transformers==4.46.1
-trl==0.14.0
+# #502/#503 — exact pins for the dedicated dependency-floor CI job.
+# Qwen3.5/Qwen3.8 text loading establishes Transformers 5.12.1 as Soup's
+# declared training floor. trl 0.29 is the first release that does not import the
+# Transformers private availability helper whose v5 return type broke trl 0.28.
+transformers==5.12.1
+trl==0.29.0
+peft==0.20.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,12 +67,8 @@ jobs:
       - name: Run one-step MLX SFT through the real CLI
         run: pytest -o addopts= -m smoke -k mlx_sft_smoke -q tests/test_smoke_train.py
 
-  # #478: declared transformers floor (4.36.0) cannot resolve against declared
-  # trl>=0.14,<0.29 — see .github/constraints/transformers-floor.txt. Do not
-  # raise Soup's pin without a dependency-policy decision. This job installs
-  # under an explicit transformers(+trl) constraint so pip cannot silently
-  # upgrade the cell, asserts the pinned versions, and runs the static dtype=
-  # guard (the accepted fallback when a true floor runtime cell is impractical).
+  # #502/#503: install the exact Transformers 5.12.1 / trl 0.29 / PEFT 0.20
+  # support floor so pip cannot silently upgrade the cell, then assert versions.
   transformers-floor:
     runs-on: ubuntu-latest
     env:
@@ -104,7 +100,7 @@ jobs:
           )
           expected = dict(pairs)
           assert len(expected) == len(pairs), "duplicate package pin in floor constraints"
-          for name in ("transformers", "trl"):
+          for name in ("transformers", "trl", "peft"):
               assert name in expected, f"floor constraints missing exact {name} pin"
               want = expected[name]
               got = md.version(name)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,8 +67,9 @@ jobs:
       - name: Run one-step MLX SFT through the real CLI
         run: pytest -o addopts= -m smoke -k mlx_sft_smoke -q tests/test_smoke_train.py
 
-  # #502/#503: install the exact Transformers 5.12.1 / trl 0.29 / PEFT 0.20
-  # support floor so pip cannot silently upgrade the cell, then assert versions.
+  # #502/#503/#522: install the exact Transformers 5.12.1 / trl 0.29 /
+  # PEFT 0.20 support floor and Plotext 6.0.0 so pip cannot silently upgrade
+  # the cell, then assert every pinned compatibility boundary.
   transformers-floor:
     runs-on: ubuntu-latest
     env:
@@ -100,7 +101,7 @@ jobs:
           )
           expected = dict(pairs)
           assert len(expected) == len(pairs), "duplicate package pin in floor constraints"
-          for name in ("transformers", "trl", "peft"):
+          for name in ("transformers", "trl", "peft", "plotext"):
               assert name in expected, f"floor constraints missing exact {name} pin"
               want = expected[name]
               got = md.version(name)
@@ -115,6 +116,7 @@ jobs:
         run: >
           pytest -o addopts= --no-cov -q
           tests/test_transformers_floor_compat.py
+          tests/test_plotext_compat.py
           tests/test_bugfixes.py::TestDiffModelLoading
 
   test:
--- a/changelog.d/0.73.3/507.changed.md
+++ b/changelog.d/0.73.3/507.changed.md
@@ -1,0 +1,10 @@
+- **Transformers training now supports Qwen3.5-family text decoders and can be
+  installed together with MLX (#502 and #503 by @Amix29 in #507).** The shared
+  stack moves to Transformers 5.12.1+, TRL 0.29+, and PEFT 0.20+, with exact
+  floor coverage and capability-based fallbacks for TRL APIs that moved under
+  `trl.experimental`. Transformers LoRA `target_modules: auto` now covers both
+  full- and linear-attention projections in Qwen3.5 text models without
+  changing explicit targets or MLX defaults. The compatibility pass also
+  updates removed Transformers arguments and keeps TRL 0.29's streamed DPO
+  reference adapter off `meta`, so it remains a frozen adapter-sized snapshot
+  instead of requiring a second model.
--- a/changelog.d/0.73.3/507.changed.md
+++ b/changelog.d/0.73.3/507.changed.md
@@ -9,4 +9,5 @@
   reference adapter off `meta`, so it remains a frozen adapter-sized snapshot
   instead of requiring a second model. DPO and ORPO also restore the removed
   prompt cap on TRL's prepared token ids, so `data.max_length` remains an
-  effective sequence bound rather than a configuration-only value.
+  effective sequence bound rather than a configuration-only value. The CLI's
+  histogram and loss-curve renderers now support both plotext 5 and plotext 6.
--- a/changelog.d/0.73.3/507.changed.md
+++ b/changelog.d/0.73.3/507.changed.md
@@ -7,4 +7,6 @@
   changing explicit targets or MLX defaults. The compatibility pass also
   updates removed Transformers arguments and keeps TRL 0.29's streamed DPO
   reference adapter off `meta`, so it remains a frozen adapter-sized snapshot
-  instead of requiring a second model.
+  instead of requiring a second model. DPO and ORPO also restore the removed
+  prompt cap on TRL's prepared token ids, so `data.max_length` remains an
+  effective sequence bound rather than a configuration-only value.
--- a/changelog.d/0.73.3/507.changed.md
+++ b/changelog.d/0.73.3/507.changed.md
@@ -1,8 +1,9 @@
 - **Transformers training now supports Qwen3.5-family text decoders and can be
   installed together with MLX (#502 and #503 by @Amix29 in #507).** The shared
   stack moves to Transformers 5.12.1+, TRL 0.29+, and PEFT 0.20+, with exact
-  floor coverage and capability-based fallbacks for TRL APIs that moved under
-  `trl.experimental`. Transformers LoRA `target_modules: auto` now covers both
+  floor coverage, matching `soup doctor` diagnostics, and capability-based fallbacks
+  for TRL APIs that moved under `trl.experimental`. Transformers LoRA
+  `target_modules: auto` now covers both
   full- and linear-attention projections in Qwen3.5 text models without
   changing explicit targets or MLX defaults. The compatibility pass also
   updates removed Transformers arguments and keeps TRL 0.29's streamed DPO
@@ -10,4 +11,5 @@
   instead of requiring a second model. DPO and ORPO also restore the removed
   prompt cap on TRL's prepared token ids, so `data.max_length` remains an
   effective sequence bound rather than a configuration-only value. The CLI's
-  histogram and loss-curve renderers now support both plotext 5 and plotext 6.
+  histogram and loss-curve renderers now support both plotext 5 and plotext 6, with
+  a real Plotext 6 runtime pinned in the compatibility CI cell.
--- a/changelog.d/0.73.3/522.fixed.md
+++ b/changelog.d/0.73.3/522.fixed.md
@@ -1,6 +1,5 @@
-- Fixed a crash on any fresh install: the unpinned `plotext` dependency started
-  resolving 6.0.0, which removed the plotting API (`clear_figure`/`hist`/`plot`) that
-  `soup data stats` histograms and `soup runs replay` charts are built on — every run
-  raised `AttributeError: module 'plotext' has no attribute 'clear_figure'`. The core
-  dependency is now capped at `<6.0.0`; support for both plotext majors arrives with
-  #507 (#522).
+- Fixed fresh installs with Plotext 6: `soup data stats` histograms and `soup runs`
+  loss charts now dispatch to Plotext 6's Figure API while retaining the Plotext 5
+  module-level path. The temporary `<6.0.0` dependency cap is lifted because both
+  supported majors are covered by the compatibility layer and regression tests
+  (#507, #522).
--- a/docs/backends-and-ops.md
+++ b/docs/backends-and-ops.md
@@ -82,11 +82,12 @@ Fine-tune on M1-M4 Macs via Apple's [MLX](https://github.com/ml-explore/mlx) fra
 pip install "soup-cli[mlx]"
 ```
 
-For SFT with local JSONL, JSON, or CSV data, `[mlx]` is a standalone install:
-do not add `[train]`. The latter is the PyTorch/TRL stack used by the
-Transformers backend. A Hugging Face `datasets` source or streaming dataset
-still needs `datasets` because that data source owns the dependency; the local
-file path below does not.
+For SFT with local JSONL, JSON, or CSV data, `[mlx]` is sufficient on its own.
+It can also be installed with `[train]` when the same environment needs the
+PyTorch/TRL Transformers backend: both extras now share the supported
+`transformers>=5.12.1,<6` range (#502/#503). A Hugging Face `datasets` source or
+streaming dataset still needs `datasets` because that data source owns the
+dependency; the local file path below does not.
 
 `detect_device()` and `get_gpu_info()` recognise Apple Silicon when
 `backend: mlx` is set, preserving `training.quantization: 4bit` for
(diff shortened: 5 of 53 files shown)


MakazhanAlpamys:
This is the whole chain, taken in eighty minutes. Reviewing properly will take longer than
that, so first: **#504 is standing down so it cannot block you.**

I opened it a couple of hours ago to *record* the ceiling — a comment in `pyproject.toml`,
a paragraph in `docs/backends-and-ops.md` saying `[train]` and `[mlx]` cannot co-install,
and a ratchet test that goes red if the bound moves without a re-measurement. It overlaps
you on exactly those two files, and its ratchet is designed to fire on precisely the change
you are making. It has served its purpose by being wrong now: the measurement it exists to
protect lives permanently in #503, which you already cite.

It is now a draft. If this lands, I close it. If this needs more time, I rebase it so the
current state stays documented in the meantime. Either way you are not rebasing around your
maintainer's PR.

**One piece of it is worth keeping regardless of who lands it**, and it is small — a check
that the docs and the declared bounds cannot drift apart:

```python
disjoint = _extras_are_disjoint()          # computed from the declared specifiers
warned   = "cannot be installed together" in backends_doc
assert warned if disjoint else not warned
```

It is a consistency check rather than a ratchet: it never blocks making the extras
compatible, it just tells you to delete the now-false warning when you do. Take it into
this PR if you want it, or leave it and I will re-open it after yours lands — your call,
and no wrong answer.

## Things I want to flag before review, not findings

**You took the acceptance item from #427 without being asked for it.** Pinning text-mode
loading to `Qwen3_5ForCausalLM` *and* asserting no vision parameters are instantiated is
exactly what I wrote into that issue's closing note as the thing #502 would have to prove.
The looks-healthy-but-partial shape has shipped here three times and a loss curve
structurally cannot see it, so that assertion is the one I would have asked for.

**The layer-streaming `ref` adapter on `meta` is the line I would not have predicted.** trl
0.29 creating the frozen DPO reference as an adapter, on a model whose base lives on `meta`,
is the same class as #433 — a thing that trains and saves and looks fine. If that is
measured rather than reasoned, say so in the PR body and it will carry a lot of weight in
review.

**What review will focus on**, so it is not a surprise: this raises three bounds at once
(`transformers`, `trl`, `peft`), and the `trl<0.29` cap had a documented reason of its own
that is not the transformers bug — `online_dpo` and the pairwise judge. Your comment says
the capability is resolved from either home, which is the right shape; the question will be
whether `online_dpo` was *exercised* on 0.29, not just imported. #326's record says the whole
class stayed invisible for releases because the trl imports live inside `setup()` and no test
called it.

Also from #326, two carried caveats worth re-checking under 5 + 0.29 since you are the first
to be there: dropping the prompt cap is not semantically neutral (a ~161-token prompt at
`max_length=64` gave ORPO `prompt_input_ids` 32 on 0.26.2 and 161 on 0.28.0), and on trl
1.9.2 a long-prompt DPO config yielded an **empty** train dataset — 0 rows, `IndexError`,
silent data loss. You cap below 1, so the second may not apply, but the first sits inside
your range.

## Unrelated, but it touches you

#509 changes the drift job so a *tracked* conflict is acknowledged rather than reddening
every Monday. It keys the acknowledgement on the exact declared pair, so when your bounds
land the entry simply stops applying — the extras become satisfiable and there is no
conflict to acknowledge. Nothing for you to do; flagging it so the interaction is not a
surprise if you see that job change colour.

Amix29:
Thanks — I took the docs/bounds consistency check and re-ran the three areas you flagged. The follow-up is in `35ca510`.

- The Online DPO smoke now proves that the evaluator was actually called. On the exact `transformers 5.12.1 / trl 0.29.0 / peft 0.20.0` floor, the real Apple Silicon CPU run generated completions, called the pairwise judge, completed an optimizer step, and emitted the four `rewards/*` metrics.
- The streamed DPO `ref` adapter finding was measured before the fix: all `.ref.` LoRA parameters were on `meta`, and the zero-`lora_B` control still had a maximum chosen-logp delta of `0.0028839111328125`. After materialization, the reference is frozen and non-meta, the zero control is exactly `0.0`, and a randomized policy adapter differs by more than `1e-4`.
- The carried prompt caveat also reproduced under TRL 0.29: a 161-token prompt with `max_length=64` reached DPO as 164 tokens and ORPO as 165. Soup now caps TRL's already-rendered token ids, so chat-template semantics stay in TRL while the effective batches obey the configured limit. The regression test retains all 8 rows and asserts every model input is at most 64 tokens.
- The docs now have the suggested consistency guard: they warn exactly when the declared `[train]` and `[mlx]` Transformers ranges are disjoint.

Both full local stacks are green after the follow-up:

- exact floor: `18492 passed, 284 skipped, 5 deselected` (81.28%)
- latest allowed: `18492 passed, 284 skipped, 5 deselected` (81.26%)

The PR body now carries the same measured evidence.

Amix29:
I pushed the CI follow-up in `3dc323b`.

The fresh dependency resolution exposed two additional compatibility gaps:

- plotext 6.0 replaced the module-level plotting API used by `data stats` and loss-curve rendering. The new compatibility helper supports both the plotext 5.x module API and the plotext 6 figure API, with regression coverage for both paths and a real 5.2.8 runtime check.
- TRL 0.29 removed `DPOTrainer.get_batch_loss_metrics` and now snapshots the initial policy in a frozen `ref` adapter. The layer-streaming gate now evaluates the same standard sigmoid DPO objective through the available log-probability helpers and gives the resident control the same reference-adapter topology, preserving the actual streamed-vs-resident invariant instead of pinning an obsolete method name.

Validation on the same latest stack resolved by CI (`transformers 5.15.1`, `trl 0.29.1`, `peft 0.20.0`, `plotext 6.0.0`):

- focused compatibility and layer-streaming coverage: `69 passed, 4 skipped`
- plotext command regressions: `6 passed`
- Ruff: clean
- full non-smoke suite: `18,495 passed, 283 skipped, 5 deselected` with 81.27% coverage

MakazhanAlpamys:
﻿Heads-up before this merges: main now caps `plotext>=5.2.0,<6.0.0` (eb77dd1, tracked as #522) because plotext 6.0 removed `clear_figure`/`hist` outright. Your compatibility layer supporting BOTH majors is the better end state — when this merges, the cap should be lifted deliberately in the same change (and #522 closed pointing at it), not left as a stale bound underneath your support. Your review will check the dual-support claim against both majors, so keeping those runtime tests in the PR matters.

Amix29:
Updated in `d3d8c20` as requested:

- rebased onto current `main` (including `eb77dd1`);
- deliberately lifted the temporary `plotext<6.0.0` cap now that the compatibility layer supports both Plotext 5 and 6;
- retained the regression coverage for both the legacy module-level API and the Plotext 6 Figure API;
- updated the #522 changelog fragment to describe the final dual-major support;
- added `Closes #522` to the PR body.

Local focused coverage is green on Plotext 6.0.0, and all 13 CI checks now pass across Linux, macOS, and Windows on Python 3.10–3.12.

MakazhanAlpamys:
﻿Full review is done - verdict: **revise**, substance is sound. Both mutations killed by your tests (Qwen3.5 target resolution; DPO prompt-cap block), floor venv at exact declared pins passes 8/8 new tests + 700 across floor-sensitive files, and the dual plotext support is REAL - verified against actual plotext 6.0.0 and 5.3.2. The guard question settled well too: re-aimed with the expired rationale removed, not silently stripped.

Three things before merge:
1. **Rebase over eb77dd1** (plotext `<6.0.0` cap, #522). Your dual-support claim justifies lifting it - do it deliberately in this PR (drop the cap line + pyproject comment saying plotext 6 is validated), don't leave a stale bound underneath your own support layer. Ideally pin one CI cell to plotext==6.0.0: your in-suite plotext-6 tests currently fake the module, so nothing exercises the real thing until after merge - exactly how #522 happened.
2. **doctor.py floors went stale**: still transformers 4.36.0 / peft 0.7.0 / trl 0.7.0 against your new 5.12.1/0.20/0.29 - soup doctor green-lights unsupported envs. No test pins that table, which is how it slipped; worth adding while you are there.
3. LOW: the new tests fail with raw ImportErrors on pre-upgrade envs - a requires-the-new-floor hint in the skip message would help.

Chain status so you can plan: #512 needs its modality pin added and will rebase over you; #519/#520/#515/#513/#510 are stacked on your head and waiting.

Amix29:
Addressed the review in `fc0be0a`, after rebasing through current `main` at `6580418` (#471 and #509 included).

1. The temporary Plotext cap is still deliberately lifted, and the compatibility cell now pins and asserts the real `plotext==6.0.0`. It runs both histogram and line rendering against the installed package; the separate contract doubles continue to pin the Plotext 5 and 6 dispatch shapes. The same runtime test also passes locally on real Plotext 5.3.2.
2. `soup doctor` now reports the validated `transformers>=5.12.1,<6`, `trl>=0.29,<1`, and `peft>=0.20,<1` bands instead of accepting the old floors and rejecting Transformers 5. A regression derives both bounds from `pyproject.toml` and compares them with the doctor tables, so either side drifting now fails by package name.
3. The runtime tests now skip pre-upgrade environments with an explicit `soup-cli[train]` hint and all three required floors instead of surfacing raw import failures.

The rebase over #471 also exposed two stale vision loader doubles that still provided `AutoModelForVision2Seq`; they now follow this PR's `AutoModelForImageTextToText` migration.

Post-follow-up validation on macOS arm64:

- full latest-stack suite: `18,534 passed, 283 skipped, 5 deselected` (81.29% coverage)
- exact compatibility cell stack (`transformers 5.12.1`, `trl 0.29.0`, `peft 0.20.0`, `plotext 6.0.0`): `29 passed`
- post-rebase review/related coverage: `230 passed`
- real Plotext 5.3.2 runtime plus metadata guards: `5 passed`
- Ruff, actionlint, and `git diff --check`: clean

The new GitHub CI run is now the final cross-platform check.

MakazhanAlpamys:
﻿Merged as squash, 13/13 on your head. What made this one worth two review rounds, for the record:

- **The re-aim of the dtype guard** is exactly what I asked for and rarely get: the expired "dtype= only exists on >=4.56" rationale was REMOVED and replaced with a stated policy ("Soup continues using torch_dtype= across the supported 5.x range; changing it is an explicit whole-policy migration"), instead of leaving a stale reason under new pins or deleting coverage added hours earlier.
- **The plotext support is real**: I verified the compat layer against installed plotext 6.0.0 AND 5.3.2 before believing it - clear_figure/hist/theme genuinely absent in 6, and your probe detects the v6 figure shape. Then you went further than asked: the constraints cell now installs plotext==6.0.0 in CI, so the real major runs every push instead of trusting mocks. That closes the exact hole #522 fell through.
- **The doctor floors are pinned by a test that bites** - my mutation (reverting them to 4.36/0.7/0.7) died on test_doctor_training_bounds_match_declared_extra.

Chain status: your stacked drafts (#510/#513/#515/#519/#520) are now unblocked on a current base - rebase and mark ready when they are.

PR #508, opened 2026-08-23
fix(export): fail fast when tensorrt_llm has no convert_checkpoint entry point (#337)

### LINKED ISSUE #337: TensorRT-LLM export calls a module that does not exist — the path has never worked
`export.py:712` runs `python -m tensorrt_llm.commands.convert_checkpoint`. **That module does not exist.** On a clean TensorRT-LLM 1.2.1 install, `tensorrt_llm.commands` contains `bench, build, eval, prune, refit, serve` and `importlib.util.find_spec` returns `None` for `convert_checkpoint`. Upstream ships conversion as per-architecture `examples/<arch>/convert_checkpoint.py`, so `trtllm-build` is never reached and **zero artifact bytes are produced**.

Verified on an H100 with a clean 1.2.1 install. Notably **this was never hardware-blocked** — the API call is simply wrong, and the box was only needed to prove it rather than to find it.

### Also worth deciding
`pip install tensorrt_llm` into a Soup training environment downgrades **torch 2.13.0+cu130 → 2.9.1**, transformers 4.57.6 → 4.57.3, numpy 2.5.1 → 1.26.4, datasets 5.0.1 → 3.1.0. Whatever the fix, the docs should tell users to install it in a separate environment.

### Fix shape
Either vendor/locate the per-architecture `convert_checkpoint.py` for the supported architectures, or drop the format and say so — shipping a documented export that cannot produce an artifact is worse than not shipping it.

### Acceptance
- [ ] `soup export --format tensorrt` either produces a loadable engine or fails immediately with a message naming what is missing
- [ ] a test that the invoked entry point actually exists in the installed `tensorrt_llm`
- [ ] docs warn about the dependency downgrade

Supersedes the TensorRT half of #73. Record: `benchmarks/gate-h100-validation.md` STEP 21.

AmirF194:
`_export_tensorrt()` shells out to `python -m tensorrt_llm.commands.convert_checkpoint`
before checking the module exists. It never has, in any current TensorRT-LLM release:
`tensorrt_llm.commands` ships only `bench/build/eval/prune/refit/serve`, and checkpoint
conversion now lives as a per-architecture `examples/<arch>/convert_checkpoint.py`
script in the NVIDIA/TensorRT-LLM source tree instead. The old code did surface the
subprocess's stderr on failure, but that read as a transient conversion problem rather
than naming what was actually missing, and it ran only after building the export plan
panel and merging any LoRA adapter.

## Fix

Import the entry point directly right after the existing `tensorrt_llm` availability
check, and exit with a message naming the missing module before any further work
happens. This is the "fail immediately with a message naming what is missing" branch
from the issue's acceptance criteria; vendoring the per-architecture conversion scripts
is a separate, hardware-verifiable piece of work this PR does not attempt.

Also added a docs note warning that installing `tensorrt_llm` can downgrade a training
environment's `torch`/`transformers`/`numpy`/`datasets` pins, per the issue's second
finding.

## Verification

- New regression test mocks `tensorrt_llm` as importable but without
  `commands.convert_checkpoint` registered: fails on `main` (`subprocess.run` is called
  and returns a generic conversion error), passes on this branch (`typer.Exit` is raised
  and `subprocess.run` is never called). Ran on Python 3.10 and 3.12 in Docker.
- `ruff check src/soup_cli/ scripts/ tests/` is clean.
- Not verified: the actual checkpoint conversion path against real TensorRT-LLM
  hardware. I do not have a GPU box for that, and this fix does not change what happens
  when the entry point exists, only what happens when it is missing.

Fixes #337

### DIFF
--- a/changelog.d/0.73.3/508.fixed.md
+++ b/changelog.d/0.73.3/508.fixed.md
@@ -1,0 +1,10 @@
+- **TensorRT-LLM export now fails immediately with a clear message instead of
+  silently producing zero artifact bytes (#337 in #508).** `_export_tensorrt()`
+  shelled out to `python -m tensorrt_llm.commands.convert_checkpoint`, a module
+  absent from every current TensorRT-LLM release (`tensorrt_llm.commands` ships
+  only `bench/build/eval/prune/refit/serve`; conversion now lives as a
+  per-architecture `examples/<arch>/convert_checkpoint.py` script instead). The
+  entry point is checked right after the existing `tensorrt_llm` availability
+  check, before the LoRA merge and checkpoint directory are touched. Docs now
+  warn that installing `tensorrt_llm` can downgrade a training environment's
+  `torch`/`transformers`/`numpy`/`datasets` pins.
--- a/docs/serving-and-export.md
+++ b/docs/serving-and-export.md
@@ -112,6 +112,10 @@ soup export --model ./output --format tensorrt
 soup export --model ./output --format tensorrt --output ./model_trt
 ```
 
+`pip install tensorrt_llm` pins its own `torch`/`transformers`/`numpy`/`datasets`
+versions, which can downgrade a training environment's stack. Install it in a
+separate environment from the one you trained in, and export there.
+
 ### BitNet 1.58 TQ1_0 GGUF Export (live in v0.71.20)
 
 Export a BitNet 1.58-bit model as a `TQ1_0` (1.58-bit ternary) GGUF via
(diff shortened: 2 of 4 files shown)


MakazhanAlpamys:
﻿Merged as squash, 12/12 export tests plus 138 neighbours green on your head. The mutation result is the part worth recording: neutering the raise died on "run was called ZERO times" - so the suite enforces abort-before-subprocess, not merely an error existing somewhere. The named error carrying where upstream moved the entry point is what turns a dead flag into a fixable message. #337 closed with a note. Thank you - two never-worked-once paths in one day.

AmirF194:
Thanks for merging, and for the mutation-testing detail, good to know the suite actually enforces abort-before-subprocess.

PR #510, opened 2026-08-23
fix(stream): preflight duplicated disk usage

### LINKED ISSUE #374: Layer streaming duplicates every base model on disk: HF cache + ~/.soup/spectrum/weights (~250 GB for four models)
## What

Streaming a model materialises a second full copy of its weights. `stream_setup` calls `spectrum_scan.resolve_model_weights`, which writes **real (non-symlink) files** under `~/.soup/spectrum/weights/<slug>/`, while the originals stay in the HuggingFace cache.

Measured on the H100 box: "For this session's four models that is **~250 GB duplicated**, which is a live constraint against 886 GB" (`benchmarks/gate-h100-validation.md` L624–628). That session was **twice stopped by a full disk**.

Note the copy is not gratuitous — the symlink guard is doing its job, and HF `snapshots/` entries are symlinks into `blobs/` which the sharder deliberately skips. This is a cost of a correct decision, not a bug in it.

## A related surprise worth fixing in the same pass

Because the sharder skips symlinks, presenting a source as a directory of **hard links** (the workaround for a gated model) changes the `source_fingerprint` (basename + size + mtime), which correctly refuses to reuse shards and **re-shards all 80 layers** — 185 s at 72B (L2711–2718). Correct behaviour, expensive and unexplained surprise. That guard firing on a real scenario for the first time is itself worth recording in its message.

## Fix path, roughly in order of effort

1. **Report it**: the pre-flight panel should state the total on-disk cost (HF cache copy + materialised copy + shard cache) before the run, so a full disk is predictable rather than discovered. This alone would have prevented both stoppages.
2. **Reuse where safe**: when the HF cache entry is already a real file (not a symlink), read it in place instead of copying.
3. **Prune**: a `soup doctor --disk`-adjacent command that reports and optionally clears `~/.soup/spectrum/weights` and `~/.soup/layer-stream`, which currently have no lifecycle at all.
4. Make the re-shard message say *why* it is re-sharding (fingerprint mismatch, with the changed component named), instead of silently doing 185 s of work.

## Acceptance criteria

- [ ] The pre-flight prints projected total disk usage and refuses when it exceeds free space, with the same *refuse-don't-degrade* posture the VRAM pre-flight uses
- [ ] A non-symlinked cache entry is not re-copied
- [ ] The re-shard notice names the fingerprint component that changed
- [ ] Cache paths remain under the existing containment policy (`SOUP_LAYER_STREAM_CACHE_DIR` / `SOUP_SPECTRUM_CACHE_DIR`)

Amix29:
## Summary

- locate the canonical Hugging Face snapshot before Soup creates a regular-file copy, and reuse cache entries that are already regular files
- verify an existing materialized copy against Hugging Face blob metadata so repeat runs do not copy it again
- print the source, materialized-copy, shard-cache, projected-total, additional-write, and per-volume free-space figures before copying or sharding, then refuse when a target volume is too small
- persist per-file fingerprint components and name the first changed filename, size, or `mtime_ns` when a shard cache must be rebuilt
- retain the existing `SOUP_SPECTRUM_CACHE_DIR` and `SOUP_LAYER_STREAM_CACHE_DIR` containment policies

## Measured checks

- Live HF-cache probe with `hf-internal-testing/tiny-random-LlamaForCausalLM`: first resolution required materialization; the second proved the existing copy through HF blob metadata and projected `0` copy bytes.
- Focused + historical streaming regression set on the original dependency range: `719 passed, 42 skipped`.
- Combined #507 compatibility, plotext command, disk-preflight, and historical streaming regression set on the fresh dependency stack: `782 passed, 57 skipped`.
- Full suite before the dependent rebase: `18486 passed, 286 skipped, 5 deselected` (81.25% coverage).
- `ruff check src/soup_cli/ tests/` and `git diff --check` pass.

## Status

Rebased onto current `main` after #507. The effective diff is now limited to the two #374 commits and seven files described above.

Closes #374.

### DIFF
--- a/changelog.d/0.73.3/510.fixed.md
+++ b/changelog.d/0.73.3/510.fixed.md
@@ -1,0 +1,1 @@
+- Fixed layer streaming's hidden duplicate-disk cost: Soup now reuses regular Hugging Face cache files, preflights materialized-weight and shard-cache writes per volume, refuses before exhausting disk space, and explains source-fingerprint cache rebuilds (#510).
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -525,7 +525,19 @@ does not make it free.
 - A published 14B-on-8 GB reference benchmark — hardware-blocked; it needs an 8 GB card and 32 GB of RAM, which the development box does not have
 - GRPO and PPO are explicitly **not** planned: rollouts need generation, which re-reads the model per token
 
-**Shard cache.** The first streaming run rewrites the checkpoint into one safetensors shard per decoder layer under `~/.soup/layer-stream/` (override with `SOUP_LAYER_STREAM_CACHE_DIR`). That costs disk space roughly equal to the base. The cache is keyed to a fingerprint of the source checkpoint, so a base retrained in place re-shards instead of silently training against stale weights.
+**Disk pre-flight and shard cache.** Before Soup materialises or shards a checkpoint, it
+reports the complete projected footprint: the HF/local source, any regular-file copy Soup
+still needs, and the per-layer shard cache. Required writes are grouped by target volume and
+the run refuses before either write when that volume lacks free space. Override the two cache
+roots with `SOUP_SPECTRUM_CACHE_DIR` and `SOUP_LAYER_STREAM_CACHE_DIR`; both retain Soup's
+home/cwd/tmp containment policy.
+
+Hugging Face snapshots normally expose symlinks into their blob cache, which the sharder
+deliberately does not follow. Soup materialises those weights under its Spectrum cache. If the
+HF cache already exposes real files, Soup now reads them in place instead of creating a second
+copy. The layer shards remain under `~/.soup/layer-stream/`. Their index records each source
+filename, size, and `mtime_ns`, so a necessary re-shard says which component changed instead
+of silently spending minutes rebuilding the cache.
 
 
 ## Correctness First (v0.36.0)
--- a/src/soup_cli/trainer/stream_setup.py
+++ b/src/soup_cli/trainer/stream_setup.py
@@ -22,9 +22,11 @@ NO top-level torch: this module is imported by five trainer modules.
 import contextlib
 import math
 import os
+import shutil
 from dataclasses import dataclass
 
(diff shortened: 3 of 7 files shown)


Amix29:
I rebased this branch on #507 rather than duplicating or pinning around its plotext 6 fix. The #374-only diff is unchanged, and the combined fresh-stack regression run is green: `782 passed, 57 skipped`, Ruff clean, and `git diff --check` clean. The new CI run will therefore validate the exact dependent stack that this PR now declares.

Amix29:
Rebased onto current `main` at `1cd6651`, including the merged #507 baseline. The old stacked commits are gone; this PR is now the same two #374 commits on `main`, with a seven-file diff limited to disk preflight, cache validation, documentation, tests, and its changelog fragment. `git range-diff` reports both commits as semantically unchanged.\n\nPost-rebase validation:\n\n- streaming regression set: `883 passed, 52 skipped`\n- `ruff check src/soup_cli/ scripts/ tests/`: clean\n- `git diff --check origin/main...HEAD`: clean\n\nMarking this ready for review; the fresh CI run is the final cross-platform check.

MakazhanAlpamys:
﻿Merged as squash, both PRs green on their heads. What review recorded, for the record:

- **#510**: the mutation that matters is gate ORDER, not just existence - neutering the render died on a test asserting the refusal happens BEFORE the materializing download, which is the whole difference between a pre-flight and a post-mortem. Reusing regular-file HF cache entries in place is the actual #374 fix; the fingerprint mismatch naming the changed component is what makes it debuggable. One MEDIUM carried to follow-up: the SOURCE HF-cache volume is projected but not free-space-warned (chicken-and-egg with snapshot_download), worth an issue if you want to finish it.
- **#513**: your literal-repo-id guard earned its keep in the exact way the house pattern predicts - my corruption of Qwen/Qwen3.8-27B in BOTH RecipeMeta.model and inline base sailed through 1644 catalog-invariant tests and was caught only by your three. The merge simulation also confirmed the modality:text family audit passes post-merge, which is precisely where kok-o's #512 failed.

Five PRs merged from you today across three stacks - thank you.

Issue #483, opened 2026-08-22
MCP refuses execution tools with 'Execution tools are not implemented in this version' — they shipped in v0.73.3 (#297)


MakazhanAlpamys:
`soup mcp serve` without `--allow-execute` refuses `train_execute` / `export_execute` with a message that ends:

```
'train_execute' can execute commands and is disabled; restart with
'soup mcp serve --allow-execute' to enable. Execution tools are
not implemented in this version.
```

`src/soup_cli/mcp_server/registry.py:1026-1029` (on `main`, today).

The last sentence is false. Execution **is** implemented — #297 shipped it in **v0.73.3** (`src/soup_cli/mcp_server/execution.py`, the `train_execute` / `export_execute` handlers, the server-generated confirmation token). The string predates that release and was never updated when the feature landed.

## Why this is worth fixing rather than shrugging at

The first half of the message tells an operator to restart with `--allow-execute`. The second half tells them it would not help. An operator who reads to the end concludes the flag is a stub and stops — so the message actively discourages using a feature that works. It is also the *only* thing many users will ever read about execution being available.

I found it while reviewing #479 (network transports), where the **same** stale assumption had produced a second, worse instance: the startup banner printed `execution disabled` while `--allow-execute` was being handed to a network listener. That one is fixed in #479. This one is on `main` and is separate, so I am filing it rather than widening that PR.

## What needs to change

1. Drop the false sentence in `_refuse_execute` (`registry.py:1021-1031`). The first half is correct and should stay: name the tool, say it is disabled, name the flag that enables it.
2. Check for siblings. `_mutating_specs`' docstring and `build_registry`'s docstring (`registry.py:1117-1125`) still say `allow_execute` "is retained separately for **future** execution tools" — same vintage, same wrong tense.

## Acceptance criteria

- [ ] No user-facing string or docstring in `mcp_server/` claims execution is unimplemented or future
- [ ] A test asserts the refusal message names the tool and `--allow-execute`, and does **not** contain "not implemented" — behavioural, i.e. it calls the handler and inspects `McpToolError`
- [ ] The test fails if the sentence is restored (please verify by putting it back and running, not by inspection)
- [ ] `grep -rn "not implemented" src/soup_cli/mcp_server/` returns nothing stale

## A note on how to test this

[...]

PR #511, opened 2026-08-23
Fix/issue 425 pre ampere fp16 bf16

### LINKED ISSUE #425: stream_layers: true crashes on pre-Ampere (T4) GPUs — fp16 GradScaler receives bf16 LoRA adapters

kok-o:
## What does this PR do?

Closes #425

Fixes a crash on pre-Ampere GPUs (e.g., T4) when using `stream_layers: true` with fp16 mixed precision. The `GradScaler` expects `fp16` parameters, but `peft` creates the LoRA adapters in the base model's dtype (`bf16` for Llama-3.1). This PR explicitly casts the adapter parameters to `float32` if `fp16` is active but `bf16` is not, preventing the `NotImplementedError` during gradient unscaling.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Tests

## Checklist


MakazhanAlpamys:
﻿Hi @kok-o - thanks for picking this up off the `help wanted` label, and sorry: that label should not have been there. #429 had been open and in active review since 16 Aug; the diagnosis in your PR is correct and your instinct (cast to fp32, guard on fp16-not-bf16, right spot ahead of trainer.train(), proper changelog fragment) matches what #429 landed after review.

The difference is coverage, not correctness: every trainer wrapper takes the same fp16 decision through utils/gpu.bf16_fp16_flags, and dpo/orpo/simpo/kto reach the identical condition via the shared StreamingSetupMixin - an sft-only patch leaves most train() sites unprotected, and casting ALL trainable bf16 params (rather than lora_*) doubles trainable memory under lora.r=0 / Spectrum / LISA after the VRAM pre-flight has passed. #429 extracts this into utils/mixed_precision.align_trainable_dtype_for_fp16, calls it from every site, and guards it with a scanner test; it is one small fix plus a rebase from merge.

Could I ask you to close #511 in favor of it? If you want a piece of this area, #433 (peft >=0.19 silently skips materialize_meta_adapters - the same stack difference you would have hit) has no owner and a decision-level test there would be genuinely valuable. Also noted separately on #512: strip it to the two recipes and rebase - the recipe work itself is nearly there.

MakazhanAlpamys:
﻿@kok-o - I owe you a correction. In my earlier comment here I pointed you at #433 as unclaimed work; that was wrong. #433 is closed and shipped - fixed in v0.73.3 by @Faisal01011 (#435/#437), and I knew it: the same session notes carry the fix. I offered you a done issue, and that is on me, not on you.

Real, currently-unclaimed alternatives if you still want one:
- **#523** (filed yesterday, no owner): SmolLM2 / Qwen2.5-Coder+Math / Mistral Small 3 / DeepSeek-R1-Distill recipes - your #512 work shows you already know the recipe invariants, including the literal-repo-id guard pattern.
- **#305** (SVD task-arithmetic for mixed-rank adapters) and **#308** (duck-typed callbacks crash on HF event dispatch) are both open, GPU-free, and neither has an active PR.

And the standing offer from before is unchanged: close #511 in favor of #429 whenever you like - the reasoning stands regardless of my bad pointer.

MakazhanAlpamys:
﻿@kok-o - correction to my correction, and this one is embarrassing: minutes after apologising for pointing you at the closed #433, I offered #305 and #308 as "both open" - they are not. Both closed completed. I posted those lists without checking state first, twice. That stops now: everything below I verified open seconds before posting.

Verified-open right now, no owner:
- **#523** - SmolLM2 / Qwen2.5-Coder+Math / Mistral Small 3 recipes (your #512 work maps directly)
- **#483** - MCP refuses execution tools with a stale version guard (`good first issue`)
- **#405** - `mini_tool_call` suite pinned at the 1.000 ceiling

Sorry for the noise. The #511-into-#429 ask itself stands unchanged.

PR #513, opened 2026-08-23
feat(recipes): add Qwen3.8 27B text SFT

Amix29:
## Summary

- add a focused `qwen3.8-27b-sft` catalog recipe for `Qwen/Qwen3.8-27B`
- declare `modality: text` explicitly, following the architecture decision recorded by #427
- keep the same dense-27B SFT shape as the existing Qwen3.5/Qwen3.6 recipes
- cover catalog lookup, search, configuration parsing, CLI `show`/`use`, explicit text modality, synchronized recipe counts, and the required changelog fragment

## Status

Rebased onto current `main` after #507, which satisfies the Transformers 5 and text-decoder dependency. This PR is now limited to its two Qwen3.8 catalog commits.

## Validation

- focused catalog, modality, CLI, and count-sync suite: `449 passed, 1 skipped`
- changelog-fragment and Qwen3.8 focused suite: `20 passed`
- Ruff: clean
- `git diff --check`: clean

Closes #477.

PR #515, opened 2026-08-23
fix(stream): support dense Qwen3.5 text decoders

### LINKED ISSUE #514: Layer streaming rejects the Qwen3.8-27B text decoder
## Reproduction

I reproduced this on Apple Silicon (M4 Max, 128 GB unified memory) with a fully cached `Qwen/Qwen3.8-27B` checkpoint and the dependency stack from #507:

- `transformers==5.15.1`
- `trl==0.29.1`
- `peft==0.20.0`
- `torch==2.13.0`

Transformers resolves the checkpoint as:

```text
outer model_type: qwen3_5
outer architecture: Qwen3_5ForConditionalGeneration
text_config model_type: qwen3_5_text
text decoder: 64 layers, hidden size 5120
```

However, `stream_arch_of(config)` raises before any weight allocation:

```text
ValueError: layer streaming does not support model_type='qwen3_5'. Supported: llama, qwen2, qwen3, mistral, gemma, gemma2, gemma3_text, phi, phi3.
```

The current alias table only admits `qwen3_5_moe` and `qwen3_5_moe_text`. This blocks the text-only BF16 LoRA hardware gate for the Qwen3.8-27B recipe being added by #513. Resident loading is not an equivalent validation path for layer streaming on Apple Silicon.

## Expected behavior

Soup should either:

1. safely route the plain `qwen3_5` / `qwen3_5_text` decoder through a proven streaming family, with resident-vs-streamed parity coverage for its actual heterogeneous decoder layout; or
2. reject it with a Qwen3.5/Qwen3.8-specific explanation if that layout needs dedicated runtime support.

The fix must pin decoder selection and numerical behavior, not merely add a model-name heuristic.

## Ownership

I am taking this issue and will prepare a focused fix with regression and parity coverage. The PR will remain dependent on #507 while the Transformers 5 loader support is not yet on `main`.

Amix29:
## Summary

- admit the plain `qwen3_5` and `qwen3_5_text` decoder types in layer streaming
- pin the admission to a native Transformers Qwen3.5 parity control containing both `linear_attention` and `full_attention` blocks
- document the dense Qwen3.5/Qwen3.8 streaming scope

## Hardware reproduction

On Apple Silicon (M4 Max, 128 GB unified memory), the fully cached `Qwen/Qwen3.8-27B` checkpoint resolves to an outer `qwen3_5` config and a `qwen3_5_text` decoder. Before this change, Soup rejected that architecture before allocating weights. After this change, the real config resolves to the proven `qwen3` streaming family.

The regression test uses Transformers' native `Qwen3_5TextConfig` and `Qwen3_5ForCausalLM`, preserves Qwen3.8's heterogeneous linear/full attention layout, streams the saved checkpoint, attaches the same LoRA tensors to resident and streamed models, and requires bit-exact logits. The measured maximum difference is `0.0`.

## Status

Rebased onto current `main` after #507, which satisfies the Transformers 5 dependency. This PR is now a single four-file Qwen3.5/Qwen3.8 streaming commit.

## Validation

- `pytest tests/ -q --tb=short`: 18,497 passed, 283 skipped, 5 deselected; coverage 81.28%
- focused streaming tests: 40 passed, 4 skipped

Closes #514


Amix29:
Rebased onto current `main` at `1cd6651`, including the merged #507 baseline. The old stacked commits are gone; this PR is now a single commit with a four-file diff limited to the dense Qwen3.5/Qwen3.8 layer-streaming aliases, the native mixed-attention parity regression, documentation, and its changelog fragment.\n\nPost-rebase validation:\n\n- Qwen3.5 streaming plus architecture allowlist coverage: `27 passed`\n- the resident-versus-streamed native decoder control remains bit-exact\n- `ruff check src/soup_cli/ tests/`: clean\n- `git diff --check origin/main...HEAD`: clean\n\n#507 is now on `main`, so the declared dependency is satisfied. Marking this ready for review; the fresh CI run is the final cross-platform check.

MakazhanAlpamys:
﻿One change and this merges - everything else verified clean, including the part I was going to police hardest: **qwen3_5 enters the streaming allowlist WITH a real bit-exact parity gate** (native Qwen3_5TextConfig toy fixture preserving the mixed linear/full attention decoder, torch.equal vs resident, CPU-runnable on every cell - the #426 shape done properly), and my alias-dict mutation died on your test.

The one REVISE item: **the new gate lacks the honest-below-floor guard its own base established.** test_issue502_qwen35_transformers5.py gates native-class tests behind _REQUIRES_TRAINING_FLOOR; yours hits ImportError on any env below 5.12.1 instead of an honest skip - CI passes because it has the new floor, but every contributor box goes misleadingly red. Reuse that marker and this lands immediately.

Amix29:
Addressed the requested REVISE item in `ef081e0`. The native Qwen3.5 parity gate now reuses `_REQUIRES_TRAINING_FLOOR` from `test_issue502_qwen35_transformers5.py`, so an environment below the declared Transformers/TRL/PEFT floor reports an explicit skip instead of failing during the native-class import.

Validation:
- Qwen3.5 streaming + floor suite: `31 passed`
- simulated `transformers==5.0.0`: skip marker active with the explicit `need >=5.12.1` reason

MakazhanAlpamys:
Merged as squash. The REVISE round was worth it precisely because of what it protects: I verified your skip marker from an actually-below-floor environment (transformers 4.57.6 here) - the parity gate now reads as an honest skip with the upgrade hint, not a red ImportError. The gate itself answers the question I opened the review with unambiguously: qwen3_5 enters WITH bit-exact parity, not allowlist-only. That closes out your entire five-PR stack from yesterday: #507, #510, #513, #519, #515 all merged. Outstanding throughput.

PR #519, opened 2026-08-23
fix(stream): preserve BF16 on capable MPS

### LINKED ISSUE #516: Layer streaming silently upcasts BF16 checkpoints to FP32 on MPS
## Reproduction

I reproduced this on Apple Silicon (M4 Max, 128 GB unified memory) while running the text-only `Qwen/Qwen3.8-27B` BF16 LoRA gate on the stack from #507, with #510 and #515 applied locally.

`resolve_stream_dtype("mps")` currently returns `float32` because every non-CUDA device is sent to the CPU test fallback. The Qwen3.8 checkpoint is BF16, but the layer sharder therefore writes FP32 shards:

```text
source tensor: BF16
streamed layer tensor: F32
BF16 layer payload: 766,546,368 bytes
written FP32 shard: 1,533,094,080 bytes
```

The disk pre-flight reported a 55.56 GB layer cache, but the completed FP32 store was 111.1 GB. Setup then failed with:

```text
training.stream_source='ram' but the base is 111.1 GB and only 70.1 GB of RAM is free
```

This is both a semantics issue (a requested BF16 run becomes FP32) and a safety issue (the pre-flight underestimates the layer-cache write by 2x). The machine's MPS backend successfully executes BF16 matmul forward and backward. PyTorch supports MPS BF16 from macOS 14 onward.

## Expected behavior

- An available MPS backend with BF16 support should stream and compute in BF16.
- An older/unsupported MPS runtime should fall back or refuse explicitly based on a tested capability, not be conflated with the CPU-only test path.
- The pre-flight estimate must use the same resolved dtype as the sharder so the announced write size matches the actual cache.
- Regression coverage should pin the dtype decision and a real MPS BF16 forward/backward smoke where hardware is available.

## Ownership

I am taking this issue. I will prepare a focused fix and rerun the Qwen3.8-27B hardware gate after the corrected BF16 cache is built.

Amix29:
## Summary

- probe BF16 support on the live MPS runtime instead of routing every non-CUDA device to the CPU FP32 fallback
- keep an explicit FP32 fallback when MPS is unavailable or the runtime rejects BF16
- pin the capability decision with platform-independent tests and a real MPS forward/backward smoke test
- document the Apple Silicon dtype behavior

Closes #516.

## Hardware validation

Reproduced and validated on an M4 Max with 128 GB unified memory using the text-only `Qwen/Qwen3.8-27B` BF16 gate. The hardware gate used the local stacked fixes from #507, #510, and #515; this PR itself is an independent one-commit diff against `main`.

Before this fix, one representative layer grew from a 766,546,368-byte BF16 payload to a 1,533,094,080-byte FP32 shard. The complete layer cache grew from the announced 55.56 GB to 111.1 GB and setup refused the run.

After this fix:

- the cache contains BF16 tensors only and remains 48.71 GiB
- the representative layer shard is 766,547,712 bytes including safetensors metadata
- all 64 Qwen3.8 layers stream through two 976 MiB buffers
- rank-64 LoRA exposes 105,906,176 trainable parameters
- one 64-token training step completes in 27.18 seconds without OOM
- final loss is 2.5355 and token accuracy is 0.6364
- the saved adapter contains 256 non-zero tensors; all 256 reload bit-exactly and no adapter tensor remains on `meta`

## Validation

- focused streaming dtype coverage: `42 passed, 4 skipped` (CUDA-only smokes skipped on this Mac)
- real MPS BF16 forward/backward/synchronization smoke: passed
- `ruff check src/soup_cli/ tests/`: passed
- `git diff --check origin/main...HEAD`: passed

Rebased onto current `main` after #507; the effective diff is one commit and four files.

### DIFF
--- a/changelog.d/0.73.3/516.fixed.md
+++ b/changelog.d/0.73.3/516.fixed.md
@@ -1,0 +1,1 @@
+- Layer streaming now preserves BF16 checkpoints on capable MPS runtimes instead of silently doubling the store and disk cache to FP32. (#516)
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -263,9 +263,11 @@ to the MPS allocator while `is_pinned()` is still false (#434). Soup refuses tha
 the source boundary and also disables pinning before allocation. Apple Silicon has unified
 physical memory, so the CUDA capacity and throughput numbers below do not transfer: only
 the MPS allocator's streamed weights are bounded by the buffer pool, while the CPU source
-still consumes unified memory. No claim is made yet that streaming fits a larger model or
-runs faster than resident MPS training. `backend: mlx` remains a separate, incompatible
-model-loading path and is rejected with `stream_layers`.
+still consumes unified memory. On macOS 14+ the store and compute dtype are bfloat16 after
+a live one-element MPS capability probe; an older runtime falls back to float32 explicitly.
+No claim is made yet that streaming fits a larger model or runs faster than resident MPS
+training. `backend: mlx` remains a separate, incompatible model-loading path and is rejected
+with `stream_layers`.
 
 The tradeoff: **1.43× slower than resident training**, measured at 0.5B — the only apples-to-apples comparison available on the reference box, because 1.5B and above cannot run resident there at all.
 
--- a/src/soup_cli/utils/layer_stream.py
+++ b/src/soup_cli/utils/layer_stream.py
@@ -206,16 +206,41 @@ def resolve_stream_dtype(device: str = "cuda") -> str:
     most of the hardware this feature exists for, and it could not fail on the
     Ampere card every published measurement was taken on (#385).
 
-    The capability question is delegated to ``utils.gpu.get_compute_dtype``
-    rather than answered again here — one question, one answer, or the two
-    drift. CPU stays float32: CPU streaming is a test convenience, and
-    half-precision CPU kernels are not uniformly available.
+    The CUDA capability question is delegated to
+    ``utils.gpu.get_compute_dtype`` rather than answered again here — one
+    question, one answer, or the two drift. MPS is separate: PyTorch supports
+    bfloat16 there from macOS 14, so a one-element allocation probes the exact
+    runtime rather than guessing from a version string. CPU stays float32: CPU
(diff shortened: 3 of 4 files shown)


Amix29:
Dependency note: this PR's code diff is independent, but the current `main` dependency resolution selects Plotext 6 while `main` still uses the Plotext 5 module API. #507 already contains and validates the proper dual-API compatibility fix. I have therefore marked this PR draft until #507 merges; after that merge I will rebase, rerun CI, and return it to ready. The Qwen3.8 hardware validation itself used #507 in the local stack.

Amix29:
The branch is now rebased on the exact #507 head (`3dc323b`), so CI exercises the declared dependency stack instead of current `main`'s known Plotext break. The unique #519 commit is `b73d87e`. Post-rebase focused validation: **50 passed, 4 skipped**, real MPS BF16 smoke included; Ruff and `git diff --check` are clean. The PR remains draft only until #507 merges.

Amix29:
Rebased onto current `main` at `1cd6651`, including the merged #507 baseline. The old stacked commits are gone; this PR is now a single commit with a four-file diff limited to the MPS BF16 capability probe, its platform-independent and hardware regressions, documentation, and its changelog fragment.\n\nPost-rebase validation in the complete training environment:\n\n- focused streaming dtype coverage: `42 passed, 4 skipped` (CUDA-only smokes skipped on this Mac)\n- real MPS BF16 forward/backward/synchronization smoke: `1 passed`\n- `ruff check src/soup_cli/ tests/`: clean\n- `git diff --check origin/main...HEAD`: clean\n\n#507 is now on `main`, so the declared dependency is satisfied. Marking this ready for review; the fresh CI run is the final cross-platform check.

MakazhanAlpamys:
﻿Merged as squash. Review notes for the record: your probe asks the RUNTIME instead of guessing from platform strings - the live-allocation check is what makes it true on capable chips and honestly false elsewhere - and the decision logic is fully stub-covered while the one real-MPS integration test skips off-Mac, so the zero-skip rule holds where it matters. My mutation flipped the availability check and both direction tests caught it. #516 closed below. That is the whole Amix29 stack from yesterday either merged or one line away - thanks for the clean rebases.

PR #521, opened 2026-08-23
fix(deps): keep Plotext on compatible major

### LINKED ISSUE #517: data stats breaks with Plotext 6 after clear_figure removal
## Reproduction

A clean Python 3.12 install from the current `main` constraints resolves `plotext==6.0.0`. The `data stats` command then fails before rendering a histogram:

```text
AttributeError: module 'plotext' has no attribute 'clear_figure'
```

The same failure is reproduced on the exact `origin/main` commit in four tests:

- `TestStatsHistogramWindows::test_stats_command_does_not_crash`
- `TestStatsHistogramWindows::test_utf8_redirect_logic_on_cp1251_stdout`
- `TestDataCommands::test_data_stats_jsonl`
- `test_stats_command`

`pyproject.toml` currently declares `plotext>=5.2.0` without an upper bound, so a fresh install can receive the incompatible major release.

## Expected behavior

`data stats` should render successfully with every Plotext version allowed by Soup's declared dependency range. The compatibility choice should be covered without weakening the existing Windows/redirect regression tests.

## Ownership

I am taking this issue and will prepare a focused dependency/API compatibility fix.

Amix29:
## Summary

- cap the core Plotext dependency below 6.0
- document why Soup still requires Plotext's module-level plotting API
- add a dependency-contract test that accepts the supported 5.x range and rejects 6.0

Closes #517.

## Why a bound instead of an API shim

Plotext 6 replaced the module-level API used by both `data stats` and run charts with an explicit `figure` API. A fresh Soup install therefore resolves successfully but crashes at runtime because `clear_figure`, `hist`, `plot`, and `show` are no longer module attributes. Supporting both majors would require migrating and validating every plotting path; this focused fix makes the currently declared behavior truthful again.

## Validation

- clean Python 3.12 environment resolved Plotext 5.3.2 under the new bound
- the four `data stats` failures reproduced with Plotext 6 all pass with the bound
- dependency-contract regression: 1 passed
- Ruff: passed

### DIFF
--- a/changelog.d/0.73.3/517.fixed.md
+++ b/changelog.d/0.73.3/517.fixed.md
@@ -1,0 +1,1 @@
+- Fresh installs no longer select Plotext 6, whose incompatible figure API broke `data stats` and run charts. (#517)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,10 @@ dependencies = [
     "pydantic>=2.0.0",
     "pyyaml>=6.0",
     "huggingface-hub>=0.16.0",
-    "plotext>=5.2.0",
+    # Plotext 6 replaced the module-level plotting API used by Soup with an
+    # explicit figure API. Keep fresh installs on the compatible major until
+    # both rendering paths can be supported and tested (#517).
+    "plotext>=5.2.0,<6.0.0",
     # `env check` reads declared bounds via packaging (Requirement/marker/version).
     # It ships with pip and is pulled in transitively, but `env check` degrades to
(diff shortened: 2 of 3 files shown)


Amix29:
Closing this in favor of #507. I found after publishing that #507 already carries the broader fix: a compatibility helper exercised against Plotext 5.2.8 and 6.0.0, covering both `data stats` and loss-curve rendering. Pinning `<6` here would duplicate that work and unnecessarily reject the supported new API. No code from this PR should be merged.

PR #525, opened 2026-08-24
fix(mcp): reconcile stale launching runs

### LINKED ISSUE #524: MCP: no recovery path for a permanently-stuck 'launching' row (blocks execution capacity until manual SQLite surgery)
Follow-up from #506 review (MEDIUM, deliberately deferred).

#506 makes a pid-less \launching\ row block restart capacity - the safe direction for #505 (a crash between INSERT and Popen() must not double-book a GPU), and spawn failures inside the process unwind via finish_execution(status='spawn_failed'). But a hard crash BEFORE spawn leaves the row forever: there is no DELETE/expunge command anywhere, so an operator hits 'execution cap reached' with no CLI way out.

**Ask:** a small operator surface - e.g. \soup mcp runs reconcile --expunge-launching\ or a --force flag on the existing reconcile path - that deletes pid-less launching rows older than a threshold, printing what it removed. Acceptance criteria: (1) refuses while a row HAS a live pid; (2) prints each removed run_id; (3) tests both directions; (4) named in docs next to the capacity contract.

Amix29:
## Summary

- add `soup mcp runs reconcile --expunge-launching` as the operator recovery path for stale MCP launch records
- keep the default recovery threshold at five minutes, with `--older-than-seconds` for an explicit positive override
- perform liveness checks and deletions in one immediate SQLite transaction
- report every removed `run_id` and refuse the entire operation when any candidate PID is still alive
- document the recovery command beside the persisted execution-capacity contract

## Base

#506 is merged. This branch is rebased onto current `main`, and the PR now contains only the five-file #524 recovery diff.

## Validation

- `PYTHONPATH=src python3 -m pytest tests/test_issue524_launching_reconcile.py tests/test_issue505_launching_window_blocks_restart.py tests/test_issue401_orphaned_run_reconcile.py tests/test_issue402_restart_capacity.py tests/test_mcp_execute.py tests/test_runs.py tests/test_issue487_changelog_fragments.py tests/test_no_foreign_license_headers.py -q --no-cov` — 90 passed
- `python3 -m ruff check src/soup_cli/ tests/` — clean
- `git diff --check origin/main...HEAD` — clean

Closes #524


MakazhanAlpamys:
Merged as squash, 69 targeted tests green on your head. The acceptance criteria were met exactly as written - including the one easy to fake: the live-pid refusal lives inside one transaction, so a refused reconcile deletes nothing on the way out. My mutation dropped the refusal and test_reconcile_refuses_live_pid_without_partially_deleting_other_rows caught it at assert 0 == 1. Also verified the #506 pin survived your change (reconcile-on-read still leaves launching rows alone). The residual Popen-to-mark_running window is honestly narrower than the age threshold by construction - good default. Fast work on a same-day issue.

PR #526, opened 2026-08-24
fix(stream): reuse one slot for untied vocabulary layers

### LINKED ISSUE #324: Layer streaming: treat untied embed_tokens + lm_head as streamed large layers
## Context

Layer streaming bounds VRAM by *one decoder layer*. Everything that is not a decoder layer stays **resident and unquantised** — including `embed_tokens` and, when untied, `lm_head`.

On Llama-3.1-8B those two matrices are **2.10 GB of the measured 3.32 GB peak**. That is 63% of the footprint of the release's headline configuration, and it is why 8B sits close to a 4 GB card's ceiling rather than comfortably inside it.

Small models hide this because their embeddings are tied (one matrix, and a small one). It bites exactly at the sizes streaming exists to enable.

## Why it is not just "quantise them"

`lm_head` is deliberately excluded from NF4 conversion (`modules_to_not_convert=["lm_head"]`) — quantising the output projection costs accuracy where it is most visible. So the fix is not to quantise them but to **stream** them: they are large, they are read once per step each, and they have exactly the access pattern the layer pool already handles.

## Rough fix path

- Extend the sharder to emit `embed_tokens` and an untied `lm_head` as their own shards rather than folding them into `extras.safetensors`
- Give the buffer pool a second, differently-sized slot class for them (they are much larger than a decoder layer, and there are only two)
- `embed_tokens` is needed at the very start of the forward and `lm_head` at the very end — the prefetch schedule already knows both boundaries
- `estimate_stream_peak_vram` drops them from `extras_bytes` and adds a buffer term

## Acceptance criteria

- [ ] Streamed logits remain **bit-exact** against a resident run of the same numerics (the standard every streaming item inherits)
- [ ] Peak VRAM on an untied-embedding model measurably drops, reported with the model, batch, seq and card
- [ ] `estimate_stream_peak_vram` still never under-predicts across the existing measured grid
- [ ] Tied-embedding models are unaffected (control)

Amix29:
## Summary

- shard `embed_tokens` and `lm_head` only when the checkpoint contains an explicit untied
  pair, and load both through one preallocated device slot
- keep tied one-matrix checkpoints on the existing resident path so their training kernels
  and numerics are unchanged
- bump the layer-shard cache format to v3 so caches produced by the earlier tied-slot
  implementation are rebuilt rather than reused
- include the large-layer store in RAM/disk tier and pinning decisions, and include exactly
  one large slot in the peak-VRAM estimate and pre-flight display
- preserve adapter save/resume, NF4 decoder streaming, heterogeneous Qwen layers, and
  preference-trainer reference forwards

## Correctness gates

- untied CPU toy decoder: streamed and resident logits and loss are bit-exact; the two
  boundary weights produce two loads into one slot rather than two resident matrices
- untied backward control: every materialized LoRA gradient matches the resident oracle
  bit-for-bit, and the comparison is explicitly non-vacuous and non-zero
- tied CPU control: no large-layer pool is installed, the embedding remains resident, and
  input/output embeddings still share the same parameter
- the never-under-predicts property grid includes an additive vocabulary-slot control and
  threads `large_layer_bytes` through the common prediction helper
- cache invalidation, corrupt/missing shard guards, tail-prefetch direction, host-store
  accounting, peak estimation, and state-dict reload remain pinned by regression tests

## Apple Silicon evidence and measurement boundary

On an M4 Max with 128 GB unified memory and PyTorch 2.12.1, the supplementary float32 MPS
untied control (batch 1, sequence length 12) is bit-exact against its resident model and
performs the expected two boundary loads through one slot. I also checked BF16 experimentally
on MPS; that resident-vs-streamed comparison had a maximum absolute difference of
`0.0009765625`, so this PR does not claim BF16 bit-exactness on MPS.

As agreed in #324, this is not a substitute for the RTX 3050 peak-VRAM acceptance run. This
PR makes no new CUDA peak claim; the maintainer-side remeasurement remains pending.

## Validation

- focused streaming, estimator, Qwen3.5, dtype, disk, NF4, preference, provenance, and
  changelog suites: `542 passed, 46 skipped`
- full repository suite on Apple Silicon: `18,858 passed, 82 skipped, 5 deselected`
- coverage: `82.79%`

Closes #324


Amix29:
Rebased onto current `main` at `80f28e9`, including the merged MPS BF16 fix from #519. The documentation conflict now preserves both contracts: the live MPS BF16 capability probe/fallback and the single vocabulary-layer slot evidence from this PR. The runtime/planner changes composed without a code conflict.

Post-rebase validation:
- streaming, dtype, NF4, disk, preference-task, and Qwen3.5 suites: `507 passed, 46 skipped`
- real MPS controls: BF16 forward/backward passes; untied float32 vocabulary streaming remains bit-exact with two boundary loads through one slot
- supplementary combined MPS BF16 run: the slot dtype is BF16, both boundary loads use that one slot, outputs are finite, and the observed resident delta remains `0.0009765625`; this is not presented as a new BF16 bit-exact claim

Amix29:
Rebased once more onto current `main` at `e7fe9b2` after the #515 squash merge. The merged dense/MoE Qwen3.5 aliases, native mixed-attention parity gate, and honest-below-floor skip are retained unchanged; this PR now adds only its large-layer sharding/runtime/accounting behavior around that baseline.

Post-rebase validation:
- streaming + Qwen3.5 floor/dtype/NF4/disk/preference suites: `519 passed, 46 skipped`
- real MPS BF16 and untied large-layer hardware controls exercised successfully in that run

A fresh cross-platform CI run is now attached to `21c360c`.

Amix29:
Fixed the cross-platform parity-test failures in `06cc5e4`.

The failures were in the test harness rather than the production streaming path. After large embedding/head layers became streamable, the first parameter exposed by the streamed model can legitimately be a `meta` tensor. The resident-vs-streamed oracle selected that parameter and moved the entire resident reference model to `meta`, so the comparison failed at `Tensor.item()` before either implementation could be compared.

The helper now selects the first materialized parameter for device/dtype alignment. I also added a regression test that deliberately registers a meta weight before a real parameter, pinning the property that the resident oracle must remain materialized.

Validation:
- DPO, ORPO, SimPO, and KTO resident-vs-streamed comparisons all pass bit-exact in isolated CPU processes
- `pytest tests/test_v07204.py -v --tb=short`: 49 passed, 19 skipped on Apple Silicon (the CUDA/CPU streaming cases are intentionally skipped under MPS)
- `git diff --check`: clean

A fresh CI run is now attached to the updated commit.

MakazhanAlpamys:
Hi @Amix29 - I ran the full mutation review on the rebase. The engineering quality is where it has been (all five sharder/planner mutations I placed died on named tests, the honest floor skips are real - breaking `_REQUIRES_TRAINING_FLOOR` to skipif(False) failed four tests by name on a below-floor box), but this is NOT mergeable yet: I found a numerics regression your own gate cannot see, plus two coverage gaps against the stated review requirements.

**Blocker 1 - CUDA bit-exact regression on the preference losses.** On a real CUDA machine (RTX 3050, bf16), `TestBitExactVsResident::test_loss_matches_a_resident_run_of_the_same_loss` fails for dpo / orpo / simpo / kto - deterministically, only on your branch (dpo delta 9.96e-04, kto 5.96e-08; the gate demands exactly 0.0). Same tests pass on main. With the vocab-slot path neutralised (large_weight_role -> None) they go green again, so the new slot path is the cause. CI cannot see this - no GPU runners - which is exactly why we run these gates on hardware. Either fix the numerics or tell me why the delta is expected and move those four gates onto an explicitly-documented tolerance... though given #324's whole point was bit-exactness, my strong preference is the former.

**Blocker 2 - the untied-path bit-exact gate is forward-only.** `test_untied_large_layers_share_one_slot_and_match_resident` compares under torch.no_grad(). Backward is covered only indirectly (gradient-liveness on a TIED model; the tied loss-curve e2e skips off-CUDA). The requirement for this series is forward AND backward vs resident for the new path - please add a streamed-vs-resident gradient comparison for an untied config.

**Blocker 3 (small) - changelog fragment misnamed.** It is `changelog.d/0.73.3/324.fixed.md` citing #324; fragments are named for the PR, so it should be `526.fixed.md`.

**Finding (moderate, not blocking by itself):** `estimate_stream_peak_vram` now takes large_layer_bytes and the planner charges the slot, but the never-under-predicts property grid was not extended - removing `+ large_layer_bytes` from the estimator killed ONLY the dedicated equality test; `test_never_under_predicts` stayed green because its grid rows do not pass the new term. Add a vocab-slot row (or thread large_layer_bytes through _predict) so the property itself guards the new term, not just the equality test.

Also: thank you for #531-#534 - I triaged all four today and they are real follow-ups, labelled. #532 (all-NaN adapter after truncation) is the sharpest of them. And the post-rebase parity-harness fix (first materialized parameter for device/dtype alignment, with the meta-weight regression control) was the right repair for a subtle failure mode.

Fix Blockers 1-3 and this merges same-day; happy to re-run the CUDA battery myself on any push.

Amix29:
Addressed all three blockers and the estimator finding in `e546e65`, after rebasing onto
current `main` (`8ea16ef`).

1. **CUDA preference-loss regression:** the slot is now installed only when both canonical
   weights are present: `model.embed_tokens.weight` and an explicit `lm_head.weight`. A tied
   one-matrix checkpoint stays on the same resident path as `main`; no large-layer pool or
   boundary replacement is installed. This removes the branch-only kernel-path change from
   the four tied preference controls that failed on the RTX 3050. The cache format is bumped
   to v3 so a cache created by the earlier tied-slot implementation cannot preserve that
   behavior. I do not have local CUDA evidence, so please rerun the four-test CUDA battery on
   this head as offered.
2. **Untied backward coverage:** added an untied streamed-vs-resident loss and backward
   comparison. It requires the same canonical LoRA gradient set, asserts that the comparison
   is non-vacuous and contains a non-zero gradient, then checks every gradient bit-for-bit.
3. **Changelog:** renamed the fragment to `changelog.d/0.73.3/526.fixed.md` and cited #526.
4. **Never-under-predicts property:** `_predict` now threads `large_layer_bytes` through the
   common estimator call, and the property grid includes an additive 16 MiB vocabulary-slot
   composition control. Removing the slot term now fails the property itself.

I also updated the tied and Qwen3.5 controls to pin the actual acceptance property: tied
embeddings remain resident and shared, while only an explicit untied pair is streamed.

Validation on the rebased head:

- full suite: `18,855 passed, 82 skipped, 5 deselected` (coverage `82.77%`)
- focused streaming/estimator/Qwen/dtype/disk/NF4/preference/changelog suites:
  `539 passed, 46 skipped`
- `git diff --check`: clean

Fresh cross-platform CI is running on `e546e65`.

Amix29:
One final rebase was needed because #537 merged while the blocker fixes were being validated.
The branch is now based on current `main` at `4cb5eaa`; the new head is `0fd90b0`.

The provenance fix from #537 and this PR's vocabulary-slot path compose without a conflict.
Post-rebase combined validation is `542 passed, 46 skipped`, with Ruff and
`git diff --check` clean. A new full local run and fresh cross-platform CI are now running on
the final rebased head.

Amix29:
Final full-suite result on the rebased head `0fd90b0`:

- `18,858 passed, 82 skipped, 5 deselected`
- coverage: `82.79%`
- duration: `8m 41s`

The PR description now reflects the final-head results.

MakazhanAlpamys:
Thank you, @Amix29. This was squash-merged. You preserved the tied path and made the untied backward comparison non-vacuous. I ran the CUDA bf16 DPO/ORPO/SimPO/KTO battery: every delta was 0.0. Breaking the vocabulary-slot path reproduced the earlier deltas, and the named forward/backward gates failed. All 13 CI jobs passed.

PR #512, opened 2026-08-23
feat: add kimi-k2.6-grpo and qwen3.5-35b-a3b-dpo recipes (#281, #276)

kok-o:
## What does this PR do?

Closes #281 and closes #276.

Adds two new ready-made recipes to the catalog:
- `kimi-k2.6-grpo` for `moonshotai/Kimi-K2.6` (GRPO reasoning variant)
- `qwen3.5-35b-a3b-dpo` for `Qwen/Qwen3.5-35B-A3B` (DPO variant)

The total recipe catalog count has been bumped from 146 to 148 across all tests and documentation.

## Type of change

- [x] New feature
- [x] Documentation
- [x] Tests

## Checklist

- [x] `ruff check src/soup_cli/ scripts/ tests/` passes
- [x] `pytest tests/ -v` passes
- [x] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed
- [x] Added a changelog fragment, or this change has no user-visible impact

PR #527, opened 2026-08-24
Feat/bash sandbox 151

### LINKED ISSUE #151: v0.53.8: /v1/tools/bash live execution — container/namespace sandbox
v0.53.7 shipped `/v1/tools/python` + `/v1/tools/web_search` as LIVE HTTP endpoints, but `/v1/tools/bash` was reverted to HTTP 501 after security review found that `/bin/sh -c` spawns a child process outside the RLVR Python sandbox's OS-level isolation (`unshare(CLONE_NEWNET)` on Linux, `sandbox-exec` on macOS, socket monkey-patch). A caller could reach the cloud-metadata service from the child shell.

## Fix path
Two options, in preference order:

1. **Container/namespace sandbox** — spawn each bash invocation inside a fresh Linux user-namespace + network-namespace via `unshare(2)` or a thin wrapper around `bwrap` (Bubblewrap) / `firejail`. macOS would use `sandbox-exec` with a stricter profile than today's. Windows would either keep 501 or use Windows Sandbox.
2. **Shell-builtin allowlist** — only permit a closed allowlist of safe shell tokens (`echo`, `printf`, `pwd`, basic arithmetic); reject any input containing `curl`, `wget`, `nc`, `python`, `/bin/*`, etc. Cheap to ship but very restrictive; many real workloads can't fit.

## Acceptance criteria
- [ ] `POST /v1/tools/bash` no longer returns 501; returns sandbox stdout/stderr/exit_code/timed_out shape (matches `/v1/tools/python`)
- [ ] Adversarial test: bash code attempting `curl http://169.254.169.254/...` fails to reach the metadata service
- [ ] Bearer auth gate preserved (matches v0.53.7 `/v1/tools/python` policy)
- [ ] 5s timeout cap preserved
- [ ] Windows behaviour documented (likely still 501 unless Windows Sandbox is feasible)

Tracked from v0.53.7 release Known Limitations bullet 1.

kok-o:
Fixes #151 

## What does this do?
This PR re-enables the `/v1/tools/bash` HTTP endpoint by wrapping it in an OS-level sandbox, mirroring the security posture of the existing `/v1/tools/python` endpoint.

- Extracts the `subprocess.run` OS-isolation logic (Linux `unshare`, macOS `sandbox-exec`) into a shared pattern that `_run_bash_sandbox` now uses.
- Blocks network access via namespaces instead of Python monkey-patching, which fixes the `curl` cloud-metadata vulnerability.
- Enforces Bearer auth and the 5-second timeout.
- Gracefully returns HTTP 501 on Windows.
- Added comprehensive tests including an adversarial network isolation test (`curl http://169.254.169.254/`).


src/soup_cli/trainer/rewards.py
```diff
@@ -375,6 +375,51 @@ def _run_code_sandbox(code: str) -> "str | None":
...
+    socket monkey-patching. Relies entirely on OS-level isolation (unshare on Linux,
+    sandbox-exec on macOS).
+    Not supported on Windows.
+    """
+    if sys.platform == "win32":
+        raise NotImplementedError("bash sandbox not supported on Windows")
+
+    _show_code_exec_warning_once()
+
+    preexec = _apply_rlimit
```
AmirF194: I ran this against 6f28207b in a clean python:3.12-slim container, which applies Docker's default seccomp profile (blocks unprivileged CLONE_NEWUSER).

`_try_unshare_namespaces` (rewards.py:117-131) treats an unshare failure as a safe fallback: "Continue with weaker isolation rather than failing the run." That is true for `_run_code_sandbox`, because it also prepends `SANDBOX_NETWORK_GUARD` (rewards.py:340) to monkey-patch `socket` inside the child interpreter, a second layer that survives an unshare failure. `_run_bash_sandbox` (rewards.py:378) has no second layer: it execs a real `bash -c command`, so when unshare fails the child keeps the parent's network namespace with nothing blocking it.

```
os.unshare(CLONE_NEWUSER|CLONE_NEWNET|CLONE_NEWPID) -> PermissionError(1, 'Operation not permitted')
readlink /proc/self/ns/net, parent:                    net:[4026532462]
readlink /proc/self/ns/net, via _run_bash_sandbox:      net:[4026532462]  (same namespace)
_run_bash_sandbox('python3 -c "urllib.request.urlopen(\'http://127.0.0.1:8811/\')..."') -> 'metadata-secret'
```

The sandboxed bash command read a local HTTP server over the network the isolation is supposed to block, the same SSRF/cloud-metadata class the v0.53.7 C1 fix reverted this endpoint for.

`test_bash_tool_network_isolation` targets `169.254.169.254`, which has no route in most hosts regardless of namespace isolation, so `exit_code == 1` there does not distinguish an isolated run from an unisolated one. It would be worth asserting on something that actually depends on the namespace, like `/proc/self/ns/net` differing from the parent's, or a reachable loopback server the way I tested here.

I don't have a clean fix beyond failing closed: if `_try_unshare_namespaces` cannot confirm the new namespace took effect, `_run_bash_sandbox` should refuse to run rather than fall through to an unisolated exec.

AmirF194:
Confirmed: 089f3ba8 makes `_try_unshare_namespaces` raise under `strict=True` instead of falling through to weaker isolation, and `_run_bash_sandbox`'s preexec now passes `strict_namespaces=True`. That closes the silent-fallback path I reported. Thanks for the quick fix.

MakazhanAlpamys:
Hi @kok-o - thanks for the batch of PRs. Review of #151 below; two blockers and three smaller items.

**Blocker 1 - the network isolation does not exist on Python < 3.12 Linux.** `_compute_isolation_strategy()` requires `hasattr(os, "unshare")`, which is 3.12+. On our own CI matrix (ubuntu x 3.10/3.11/3.12) the endpoint silently degrades to best-effort on two of three cells - bash runs WITH network access and returns 200 - while the docs added here say network access is blocked at OS level. That is the curl-to-metadata vulnerability re-opened on most installs. Either gate the endpoint itself on strict isolation being available (503/501 when not), or make the docs say exactly what each platform gets.

**Blocker 2 - the security tests never run anywhere.** Two separate problems:
- `fastapi` is not in `[dev]`, so the whole TestToolEndpointsLive class SKIPS in CI - confirmed from the ubuntu job log: every bash-tool test was SKIPPED. The adversarial tests you wrote execute nowhere automatically.
- On win32 `test_bash_tool_network_isolation` ends with a bare `return`, so it reports PASSED while testing nothing. That needs to be an honest skip with a reason string.

Mutation evidence: disabling the sandbox wrapper killed `test_bash_tool_executes_in_sandbox_or_501_on_windows` (good), but removing `_check_tool_auth` from the bash endpoint killed NOTHING - the Bearer gate on this endpoint has zero coverage (the existing auth tests only hit /v1/tools/python). Please add an auth test for the bash endpoint.

**Smaller items:**
- The PR description claims an adversarial curl to 169.254.169.254; the actual test curls a local 127.0.0.1 stub. Either add the real metadata-address assertion or correct the claim.
- `_run_bash_sandbox` duplicates the body of `_run_code_sandbox` almost line for line (tempdir -> subprocess.run -> TimeoutExpired -> cap), and utils/agent_sandbox.py already carries a third copy of the pattern. This repo has been burned by third copies repeatedly (#372/#392) - extract what is genuinely shared or tell me why not.
- The docs sentence about OS-level blocking needs to match reality per platform (see blocker 1).

Everything else looked right: argv-lists no shell, timeout applied, Windows 501 is honest, ruff clean. Fix the isolation honesty + CI visibility and this is mergeable.

MakazhanAlpamys:
@kok-o: #527 needs working Linux Python <3.12 isolation without silent fallback, CI-installed security tests, and real user_data/auth-bash coverage. The last mutation survived, so include the mutation result with the update.

Amix29:
The earlier isolation/auth blockers are already fixed on the contributor branch. I found and fixed one remaining API-contract defect in kok-o/Soup#3: ordinary non-zero exits, stderr, and real timeouts were previously collapsed into the same `timed_out: true` response.

The shared runner now preserves structured process outcomes; the bash endpoint returns the real exit code/stderr and uses 124 only for an actual timeout. The docs and runtime warning also state the filesystem boundary explicitly.

Validation: 232 focused tests passed (4 platform skips), plus real Apple Silicon `sandbox-exec` checks for exit 7 + stderr versus a true timeout; Ruff is clean.

MakazhanAlpamys:

MakazhanAlpamys (changes requested):
Changes requested; this is a security blocker. I read issue #151 and verified the current head.

The relevant local suite passed (177 passed, 8 skipped), Ruff is clean, the WSL2 Linux 3.12 network-boundary probe failed closed, and the required mutation made the strict-namespace test fail as expected. The remaining blockers are:

- P1: with a non-loopback bind and no `auth_token`, the bash endpoint remains unauthenticated and executes commands for network callers.
- P1: the 10 KB output limit is checked only after `capture_output=True` has accumulated the entire stdout/stderr, so a large-output request can exhaust server memory. There is also no concurrency/process/file-size containment.
- P1: child processes inherit the server environment and filesystem; a probe printed an injected secret. The endpoint needs an explicit safe boundary or must be unavailable by default.
- P2: restricted Linux can surface HTTP 500 instead of the documented fail-closed 501 path.

Please require authentication for exposed listeners, enforce bounded streaming/resource handling, and clarify or enforce the filesystem/environment boundary before merge. The green CI does not cover these adversarial deployment cases.

MakazhanAlpamys (changes requested):
Additional blocking finding on the current head: when `unshare` fails inside `preexec_fn`, Python surfaces `subprocess.SubprocessError`, but `tool_bash` only handles `PermissionError` and `NotImplementedError`. In that deployment the endpoint returns HTTP 500 instead of the documented fail-closed 501 path.

I reproduced this boundary in the isolated runner and verified that the strict-namespace mutation makes the dedicated assertion fail, so the safety requirement is load-bearing. Please catch the actual subprocess failure shape (or normalize it at the runner boundary) and add a regression for the HTTP outcome. The existing exposed-listener authentication, output/resource-boundary, and environment/filesystem isolation blockers also remain unresolved.

The current CI is green, but this is a security blocker.

MakazhanAlpamys:
The current head still has the previously identified fail-closed error-path blocker: an actual `subprocess.SubprocessError` from strict namespace setup is not mapped to the documented HTTP 501 response. In addition, the PR base is now behind `main` (base `0fcf383`; current `main` has since advanced), so the green matrix is not a current-head validation. Please fix the exception mapping, rebase onto current `main`, and rerun the full checks before requesting another review.

MakazhanAlpamys (changes requested):
Re-review of b8d6af68ee867bf1d3afea6e7e046c95c2d73956: my focused run passed 116 tests with 12 skips, Ruff is clean, and the authentication mutation was caught. A blocking compatibility gap remains: os.unshare is only available on Python 3.12+, while Soup supports Python 3.10 through 3.12. On Linux with Python 3.10 or 3.11 the endpoint falls back to HTTP 501, contradicting issue #151's acceptance criteria. Use an external unshare or sandbox wrapper, or explicitly narrow and document the supported scope with tests. The exposed-listener authentication, resource-boundary, and environment or filesystem isolation blockers also remain unresolved.

MakazhanAlpamys:
@kok-o — thanks for pushing `b10fc664` today; the `subprocess.SubprocessError` mapping is the blocker I raised on 2026-08-27 and it looks addressed. I have not reviewed the sandbox implementation yet because CI on `70cc527c` is still running, and I will not review against pending checks.

But `70cc527c` ("fix(docs): resolve docs duplication and update recipes count to 159") has a problem I need fixed before I go further, and it is **not** a duplication cleanup. Measured against current `main`, whose merge-base with this branch is `88b78ed`:

```
$ git diff origin/main...70cc527c --stat -- docs/serving-and-export.md
 docs/serving-and-export.md | 56 +++++++-------------------
 1 file changed, 22 insertions(+), 34 deletions(-)
```

Those 34 deletions are live documentation for shipped features, unrelated to #151. All four of these strings are **present on current `main`** and removed by this branch:

1. **The cross-tokenizer speculative decoding paragraph** — *"Cross-tokenizer speculative decoding is supported on the Transformers backend via Universal Assisted Decoding (UAD)…"*
2. **The whole "Cross-tokenizer measurement" block**, including the `repetition_penalty=1.0` neutralisation note and its `Refs #345`.
3. **The cross-tokenizer acceptance lower-bound note** — *"the measured acceptance rate is a strict lower bound … it will never over-report."* This one is **your own contribution from #464**, and it is the sentence that stops a user reading a conservative number as a real one.
4. **The `tensorrt_llm` environment warning** about it downgrading a training stack's `torch`/`transformers`/`numpy`/`datasets`.

The `soup draft distill` paragraph is not just deleted, it is **rewritten to say the opposite of what the code does**:

> Draft and target must share a tokenizer — a mismatch is refused up front

That is false on `main`:

```
src/soup_cli/commands/draft.py:546   cross_tokenizer = target_vocab != draft_vocab
src/soup_cli/commands/draft.py:558   uld_strategy = "wasserstein_aligned" if cross_tokenizer else None
src/soup_cli/commands/draft.py:590   " [cyan](cross-tokenizer ULD: wasserstein_aligned)[/]"
```

A mismatch is not refused; it routes through cross-tokenizer ULD. Shipping this would document a working feature as absent.

The vLLM note also reintroduces a stale claim:

> pushes `transformers` past Soup's own `<5.0.0` cap

`pyproject.toml` on `main` declares `transformers>=5.16.1,<6.0.0`. There is no `<5.0.0` cap any more.

Taken together this reads like a rebase that resolved conflicts by taking the branch side of `docs/serving-and-export.md` wholesale, which reverts everything `main` added to that file since the branch was cut. Please drop the documentation changes that are not about the bash sandbox and keep only what #151 needs. If you would rather rebuild the branch, `git checkout origin/main -- docs/serving-and-export.md` and then re-apply just your sandbox section is the smallest path.

Two other notes while you are in there:

- **The recipe count.** `70cc527c` sets it to 159. Please state where that number came from — if it is a count of `recipes/catalog.py`, say so in the commit body, and if it was carried over from a rebase, drop it. A count changed as a side effect of a sandbox PR is exactly the kind of drift that later gets treated as authoritative.
- **The deleted tests are fine and I checked them.** The 36 removed lines in `tests/test_v0537.py` are the `501`-stub assertions (`test_bash_tool_returns_501_review_fix_c1`, `TestReviewFixesC1BashStub`, and the empty-body variant). Removing them is correct — the endpoint is no longer a stub. No other test was deleted or weakened.

Once CI on the corrected head is green I will review the sandbox itself properly, including mutation testing on the isolation and auth paths. Given this is live shell execution, expect that review to be thorough.

Separately, on your three other open PRs: #512 is still conflicting with `main` and needs a rebase; #528 and #529 have not moved since 2026-08-25. @Amix29 prepared review follow-ups for all of them on your fork (`kok-o/Soup#1`–`#4`) which are still unmerged. If you would rather not carry all four at once, tell me which you want to keep and I will close the rest as stale so your queue is honest — no judgement either way, four parallel PRs is a lot.

MakazhanAlpamys:
CI went green on `70cc527c`, so here is the security review I promised. The docs finding above still stands separately.

**The implementation is much better than the v0.53.7 attempt.** Credit where it is due, because most of this is right:

- `/v1/tools/bash` **fails closed** — `501` when `_get_isolation_strategy()` is `best-effort`, `501` on Windows, `501` when the sandbox fails to initialise.
- `_get_safe_sandbox_env()` scrubs the environment to an allowlist, so the child cannot read `HF_TOKEN`, `OPENAI_API_KEY` or anything else the server was started with.
- Streaming kill-on-overflow rather than buffering, a real semaphore for concurrency, `RLIMIT_FSIZE` / `RLIMIT_CORE` / `RLIMIT_NOFILE` / `RLIMIT_NPROC` added on top of the existing `RLIMIT_AS` / `RLIMIT_CPU`.
- The docs paragraph for the endpoint is accurate and admirably honest — *"This is not a filesystem sandbox: the child retains the server process's read access to world-readable system files."* That sentence is worth more than three paragraphs of reassurance.
- The 36 removed lines in `tests/test_v0537.py` are the `501`-stub assertions and their removal is correct.

I ran four mutations against the auth and isolation guarantees. **Three died. The one that matters most survived.**

```
remove _check_tool_auth() from tool_bash        -> 2 failed  (KILLED)
remove the non-loopback 401 in _check_tool_auth -> 1 failed  (KILLED)
remove the best-effort -> 501 guard             -> 1 failed  (KILLED)

_apply_rlimit(strict_namespaces=True -> False)  -> 116 passed, 12 skipped   SURVIVED
```

## Blocking: `strict_namespaces` is the whole security argument and nothing guards it

`_run_bash_sandbox` has no Python-level socket monkey-patch — it says so itself: *"Relies entirely on OS-level isolation."* On Linux that isolation is exactly one thing: `_try_unshare_namespaces(strict=True)` raising instead of falling back. Flip that one keyword to `False` and:

- `unshare(CLONE_NEWUSER|CLONE_NEWNET|CLONE_NEWPID)` fails on a hardened distro or under Docker's default seccomp profile — which is precisely the environment @AmirF194 reproduced in on 2026-08-24;
- `_try_unshare_namespaces` swallows the `EPERM` and returns;
- `bash -c <caller-supplied command>` runs **with the host network**, and `curl http://169.254.169.254/...` works.

That is the exact defect that caused the revert to `501` in v0.53.7, restored in full, and **the entire test suite stays green**. Confirming it is not merely an untested line:

```
$ grep -rn "strict_namespaces\|_try_unshare_namespaces" tests/
(no matches)
```

Zero tests in the repository reference either symbol. The strict path is unexercised, so nothing distinguishes "namespaces were actually established" from "we tried and gave up".

Please add a test that fails when the strict flag is dropped. It does not need root or a real namespace — patch `os.unshare` to raise `PermissionError`, drive `_run_bash_sandbox`, and assert it raises rather than returning a result; then a second test asserting the endpoint turns that into `501`. A companion positive control matters too, since a test that only checks the failure direction passes on a build where the sandbox never runs anything at all.

Related, and worth deciding explicitly rather than by default: `_apply_rlimit` only reaches `_try_unshare_namespaces` when the strategy is `namespaces`. On macOS the strategy is `sandbox-exec` and `argv` is wrapped, which is fine — but the two paths have very different guarantees and only one of them is named in the flag.

## Should-fix

**1. `argv = ["bash", "-c", command]` uses a bare executable name.** This repo has a written policy against that, in `utils/layer_stream.py:396`:

> *"Bare names are deliberately never handed to `subprocess`: on Windows `CreateProcess` searches the CURRENT DIRECTORY before `PATH` … (CWE-427). POSIX `execvp` searches `$PATH` only, so this matters most on Windows — but resolving everywhere costs nothing."*

The inconsistency is inside your own function: `_run_sandboxed_subprocess` resolves `sandbox-exec` properly with `_shutil.which(...) or "/usr/bin/sandbox-exec"`, and then `bash` is passed bare. The Windows leg is unreachable here (`NotImplementedError`), so this is lower severity than the first item, but please make it `_shutil.which("bash") or "/bin/bash"` for consistency.

**2. The changelog fragment omits a breaking change.** It reads:

> `- Re-enabled /v1/tools/bash execution with OS-level namespace/sandbox isolation (#151)`

But `serve()` now does `raise typer.Exit(code=2)` where `main` only printed a yellow warning. Anyone running `soup serve --host 0.0.0.0` without `--tool-auth-token` today has a working command that stops working. I agree with the change — it is the right call — but it needs to be in the fragment as a behaviour change, not left for users to discover at startup.

## Still open from my previous comment

The `docs/serving-and-export.md` deletions are unchanged: the cross-tokenizer speculative-decoding paragraph, the `Refs #345` measurement block, your own acceptance lower-bound note from #464, and the `tensorrt_llm` warning are all still being removed, `soup draft distill` is still documented as refusing mismatched tokenizers when `draft.py:546-590` routes them through `wasserstein_aligned`, and the vLLM note still asserts a `transformers <5.0.0` cap that `pyproject.toml` no longer has. Your **new** documentation for `/v1/tools/bash` is good and should stay — it is only the unrelated deletions that need reverting.

Fix the strict-namespace test gap and the docs, and I will re-run the mutation set. This is close, and I would rather it land correct than fast — it is the only endpoint in the project that runs an attacker-chosen shell command.

MakazhanAlpamys:
Re-reviewed `7201a725`. **B, C and D are genuinely fixed** — thank you, those were clean:

- **B (docs):** `git diff origin/main...7201a725 -- docs/` is now confined to the "Server-Side Tool Endpoints" section. All four strings I named survive, `Refs #345` is back at `serving-and-export.md:364`, the false `soup draft distill` rewrite is gone, and the `transformers <5.0.0` claim with it. `docs/commands.md` untouched.
- **C:** `rewards.py:545-546` now does `bash_bin = _shutil.which("bash") or "/bin/bash"`, consistent with the `sandbox-exec` line above it.
- **D:** the fragment now names the breaking change explicitly.

**A is still open, and the new tests make it worse rather than better.** Please do not merge this head.

## The new test fails against your own unmutated code on Linux

```python
# tests/test_v0537.py:1092
with pytest.raises(PermissionError, match="Operation not permitted"):
    _run_bash_sandbox("echo hello")
```

CPython does not propagate a `preexec_fn` exception with its original type. The child writes a fixed name into the error pipe and the parent re-raises from that (`subprocess.py:1833-1848`):

```python
child_exception_type = getattr(builtins, exception_name.decode('ascii'), SubprocessError)
```

For a `preexec_fn` failure the name written is `SubprocessError`, message `"Exception occurred in preexec_fn."` — measured on Linux, Python 3.12.3:

```
python 3.12.3 -> subprocess.SubprocessError : Exception occurred in preexec_fn.
```

And `SubprocessError` is not an `OSError`:

```
SubprocessError MRO: ['SubprocessError', 'Exception', 'BaseException', 'object']
is OSError subclass: False
```

while `rewards.py:422` catches only `except (PermissionError, OSError)`. So on the strict path the exception flies straight past the handler, `launch_failed` is **never set**, and `_run_bash_sandbox`'s `raise PermissionError(result.stderr)` at line 549 is unreachable. The function's documented contract is not what it does.

Consequence: `test_bash_sandbox_raises_on_strict_namespace_failure` goes red on `ubuntu-latest` × 3.10 / 3.11 / 3.12. macOS and Windows skip it, which is why the current green checks do not show this — a `pytest.raises` replay of your verbatim test body on WSL Ubuntu 24.04 gives:

```
test_bash_sandbox_raises_on_strict_namespace_failure: FAILED
  -> subprocess.SubprocessError: Exception occurred in preexec_fn.
test_bash_sandbox_runs_on_strict_namespace_success:   PASSED
```

The endpoint still fails closed today, but only because `serve.py:2073` separately catches `subprocess.SubprocessError -> 501`. That is luck, not the design in your docstring.

## The mutation still survives

```
_apply_rlimit(strict_namespaces=True) -> False
  -> 129 passed, 14 skipped
```

The new test cannot kill it, on any platform. I ran it on Linux against both states of the code:

```
UNMUTATED: FAIL (wrong exception: SubprocessError)
MUTATED:   FAIL (no exception raised at all, rc=0, out='hello')
```

A test that fails in both states is not a control — it tells us nothing about which one we are in, and it breaks the build where it runs. `grep -rn "strict_namespaces" tests/` is still **zero matches**; the new tests never name the parameter.

The reason nothing else catches it is worth stating, because it is the trap to avoid on the next attempt: every existing 501 test (`test_bash_tool_restricted_linux_fails_closed_501`, `test_bash_tool_subprocess_error_maps_to_501`) monkeypatches `_run_bash_sandbox` **itself**, so they pass identically with `strict=False`. Any new test must exercise the real `_apply_rlimit` preexec path.

## What to do

1. **Catch `subprocess.SubprocessError` in `_run_sandboxed_subprocess`** alongside `(PermissionError, OSError)` and set `launch_failed=True`. That makes the `launch_failed` branch reachable and makes `_run_bash_sandbox`'s documented `PermissionError` contract true. Changing the test to expect `SubprocessError` instead would also go green, but it would leave a docstring that lies and a branch that never runs — please take the first option.
2. **Then add a test that actually kills `strict_namespaces=True -> False`**, going through the real preexec path. Verify it the way I do: apply the mutation, show the named test fail, revert, show it pass. If it does not fail under mutation it is not the test we need.
3. **Reset `_ISOLATION_STRATEGY_CACHE` in both new tests.** `os.unshare` is 3.12+, and on 3.10/3.11 the cache may already hold `"best-effort"` from an earlier test, so today their behaviour depends on execution order — the current failure mode there is `DID NOT RAISE` rather than the wrong-exception one.

Two small ones while you are in there: `import os` sits inside the test body after the skip guards, and there is a redundant double skip (`== "win32"` then `!= "linux"`). And in the docs rewrite the `/v1/tools/python` bullet lost the "Requires Bearer token auth" sentence and its 200/400/401 status list that `main` documented — worth restoring, since it is the sibling endpoint of the one you are adding.

For the record on the rest: no test was deleted or weakened beyond the legitimate `501` stubs I already cleared, mutations 2, 3 and 4 (auth, best-effort gate, non-loopback 401) all still die by name, `ruff` is clean, `test_cli_startup_is_light` is 34 passed, and `tests/test_v0537.py` is `116 passed, 14 skipped`.

You are one real test away. The isolation guarantee is the only thing standing between this endpoint and the v0.53.7 revert, so it needs a control that can fail.

MakazhanAlpamys:
@kok-o — you pushed `e6df0444` on 2026-08-30 and I did not come back to it until now. That is on me, not on you, and I am sorry for the wait. Here is the re-review.

## Fixed, and you took the harder of the two options

```
rewards.py:422
-        except (PermissionError, OSError) as exc:
+        except (PermissionError, OSError, subprocess.SubprocessError) as exc:
```

That is the right call. Changing the test to expect `SubprocessError` would also have gone green, but it would have left `_run_bash_sandbox`'s docstring promising a `PermissionError` it could never raise and the `launch_failed` branch permanently dead. Now the strict path can actually set `launch_failed`, the documented contract is true, and the Linux CI red I predicted (`pytest.raises(PermissionError)` against a `preexec_fn` failure, which CPython surfaces as `SubprocessError`) is resolved.

**B, C and D remain fixed** from the previous round — docs regression reverted, `which("bash")`, breaking-change noted in the fragment. Baseline on this head: `129 passed, 14 skipped`.

## Still open: item A, and it is the only thing left

```
$ grep -c strict_namespaces tests/test_v0537.py
0

# mutation: _apply_rlimit(strict_namespaces=True) -> False
129 passed, 14 skipped          <- byte-identical to baseline
```

`_run_bash_sandbox` has no Python-level socket guard — its own docstring says it "relies entirely on OS-level isolation". On Linux that isolation is exactly one thing: `_try_unshare_namespaces(strict=True)` raising instead of falling back. Flip that keyword and, on a hardened distro or under Docker's default seccomp, `unshare` fails, the failure is swallowed, and `bash -c <caller command>` runs with the host network — reachable to `http://169.254.169.254/`. That is the defect that forced the 501 revert in v0.53.7, and nothing in the suite notices.

Your two new tests do not close it: both are `skipif(sys.platform != "linux")`, so they never run in any CI cell nor on your box or mine, and neither names the parameter. The test I need must exercise the real `_apply_rlimit` preexec path — patching `os.unshare` to raise `PermissionError` and asserting `_run_bash_sandbox` raises is enough, no root and no real namespace required. Verify it the way I do: apply the mutation, show the named test fail, revert, show it pass.

Every existing 501 test monkeypatches `_run_bash_sandbox` itself, which is why they all pass with `strict=False` — that is the trap to avoid on this attempt.

## On the wait

You are right to push back. For what it is worth on the other three: **#528 is red on one missing `- ` in `changelog.d/0.73.3/264.added.md`** and has the Lambda `user_data` and API-key-in-boot-log blockers from 2026-08-25 unaddressed; **#529** has not moved since before the env-only redesign was agreed, and its opt-out is still unpinned (8 of 12 tests carry no `assert`); **#512** is `CONFLICTING` and still has no literal repo-ID guard — both ID mutations survive byte-identically. @Amix29's correction PRs on your fork (`kok-o/Soup#1`-`#4`) are all still unmerged, and merging those plus a rebase is the fastest path on all three.

This one is close. One test and I merge it.

MakazhanAlpamys:
@kok-o — re-reviewed `d6635d84`. **You went after the right thing, and the structural half of it is genuinely fixed.** The test half is not there yet, and it currently turns Windows red. Details, with the evidence, so you do not have to take my word for any of it.

## What you got right, and it is the part that mattered most

```
pyproject.toml  [dev] += fastapi>=0.104.0
```

That is the real fix. `TestToolEndpointsLive` was never executing in CI at all — no FastAPI, no class. Adding the dev dependency is what made it run, and running is what surfaced everything below. You did not patch around the visible symptom; you made the tests actually execute. Same instinct as the `SubprocessError` fix last round.

`agent_sandbox.py` delegating to the shared `_run_sandboxed_subprocess` is also the right call — that was the third-copy smell, and `tests/test_v07118.py` still passes 114.

**B, C and D remain fixed.** I re-checked rather than assuming: docs still confined to the "Server-Side Tool Endpoints" section with all four previously-deleted strings intact, `rewards.py:545` still `_shutil.which("bash") or "/bin/bash"`, the fragment still names the `--tool-auth-token` breaking change. `ruff` clean, no unrelated files, no Cyrillic.

## Blocker 1 — the new test hard-fails on all three Windows cells

```
test (windows-latest, 3.10)  fail
test (windows-latest, 3.11)  fail
test (windows-latest, 3.12)  fail
test (ubuntu-latest, *)      pass
test (macos-latest,  *)      pass
```

```
tests/test_v0537.py:1106
    monkeypatch.setattr("resource.setrlimit", lambda *args: None, raising=False)

E   ModuleNotFoundError: No module named 'resource'
FAILED tests/test_v0537.py::TestToolEndpointsLive::test_bash_sandbox_raises_on_strict_namespace_failure
= 1 failed, 19241 passed, 211 skipped in 1705.04s =
```

The target is passed as a **string**, so pytest resolves it with `__import__("resource")` before it ever looks at the attribute. `resource` is POSIX-only and does not exist on Windows, so this raises during test **setup** — before `_run_bash_sandbox`, before the mocked `Popen`, before the line under test. `raising=False` does not help: it covers a missing *attribute*, not a missing *module*. Patching `sys.platform` inside the test cannot conjure the module either.

The commit title is "replace skipif bash sandbox tests with cross-platform mock", and I understand why — I asked for a test that is not `skipif`, because a skipped test never runs in any cell. But the outcome on Windows is worse than the skip was: it used to `SKIPPED [bash sandbox not supported on Windows]`, and now it `FAILED` with a traceback that reads like a real regression while proving nothing. I reproduced the mutation on Windows and the result is byte-identical either way:

```
baseline (unmutated)          1 failed, 116 passed, 12 skipped
_apply_rlimit(strict=False)   1 failed, 116 passed, 12 skipped   <- same traceback
```

**Fix:** guard only the `resource` patch, not the test:

```python
if sys.platform != "win32":
    monkeypatch.setattr("resource.setrlimit", lambda *args: None, raising=False)
```

or inject the limiter rather than monkeypatching a module that may not exist. Either keeps the test running everywhere it can prove something and stops it failing where it cannot.

## Blocker 2 — the positive control was deleted, not fixed

```
git diff e6df0444..d6635d84 -- tests/test_v0537.py
-    def test_bash_sandbox_runs_on_strict_namespace_success(self, monkeypatch):
```

That is the control I named last round, and the reason stands unchanged: a test that only checks the failure direction passes on a build where the sandbox never runs anything at all. Please restore it.

## Item A — closer than it looks, but the control is coincidental

On Linux the new test does discriminate. I could not get a real Linux run, so I say what I actually did: I injected a stand-in `resource` module locally so the same code path executes, and the named mutation kills it —

```
_apply_rlimit(strict_namespaces=True) -> False
AssertionError: Regex pattern did not match.
  Regex: 'Operation not permitted'
  Input: 'Mock prevents actual execution'
```

— and `ubuntu-latest, 3.11` shows it `PASSED` unmutated on this head. So the logic is right.

What I am not comfortable with is *how* it distinguishes them: on the wording of an f-string three exception-wrappings deep. Reword any message in `_try_unshare_namespaces`, `mock_popen` or `_run_bash_sandbox` and the control silently stops controlling, in either direction, with nothing going red.

```
grep -rn "strict_namespaces" tests/   ->   0
```

Third round with that at zero. Please assert on the parameter itself, not on a string that happens to differ. That is the whole difference between a test that pins the security property and one that pins today's phrasing of an error.

## One process note that affects you directly

I enabled branch protection on `main` today: the 9 test cells plus `lint`, `mlx-smoke` and `transformers-floor` are now **required**. `strict` is off, so nothing forces you to rebase — but a red Windows cell now blocks the merge button structurally rather than depending on me noticing. That is the correct state and it is not aimed at this PR; it is aimed at the fact that red CI has been mergeable here at all.

## Summary

Three items, all mechanical: the `resource` guard, the restored positive control, and one assertion that names `strict_namespaces`. The security substance of this PR I already consider sound — fail-closed 501s, scrubbed environment, streaming kill-on-overflow, real rlimits. It is the proof that is missing, not the property.

I would rather say this plainly than have you guess: this is the fourth round, and that is frustrating from your side. It is also the single highest-risk endpoint anyone has proposed adding to this project — `bash -c` reachable over HTTP, where the failure mode is the host network and `169.254.169.254`. I am slow here on purpose, not because the work is weak.

MakazhanAlpamys:
@kok-o — reviewed `9155bb5b`, twenty minutes after my last comment. **Item A is closed.** I measured it rather than reading the diff, on Windows, which is the platform that made the previous attempt meaningless.

## The fix works, and here is the proof I owe you

```
tests/test_v0537.py  (windows, py3.10)
  baseline                                        117 passed, 12 skipped
  mutation: rewards.py:543
    _apply_rlimit(strict_namespaces=True) -> False  1 failed, 116 passed, 12 skipped
    FAILED ...::test_bash_sandbox_raises_on_strict_namespace_failure
  revert                                          117 passed, 12 skipped
```

Compare with the previous head, where the same mutation gave `1 failed, 116 passed` **byte-identically to baseline** — a red mark that moved for neither state. Now the baseline is green and the mutation is the only thing that turns it red. That is a test that discriminates, and it is the third round this has been asked for.

The repair itself is the right one:

```python
try:
    import resource
    monkeypatch.setattr(resource, "setrlimit", lambda *args: None)
except ImportError:
    pass
```

Patching the **module object** rather than the `"resource.setrlimit"` string is what fixes it: the string form makes pytest `__import__("resource")` during setup, and that module does not exist on Windows at all. You could have wrapped the whole test in a skip and gone green; you kept it running everywhere and let the import be the only conditional part. The three Windows CI cells this took down are resolved.

## One item left, and it is the one I flagged as deleted

```
$ grep -n "def test_bash_sandbox" tests/test_v0537.py
1092:    def test_bash_sandbox_raises_on_strict_namespace_failure
```

`test_bash_sandbox_runs_on_strict_namespace_success` is still gone. Restoring it is the whole remaining ask, and the reason has not changed: **a suite that only tests the refusal direction is green on a build where the sandbox never executes anything.** Your negative test proves `unshare` failing is fatal. Nothing yet proves that when `unshare` succeeds, a command actually runs and returns its output — so a regression that broke execution outright would ship green.

This is the same shape as the control that made #597's bit-exactness harness trustworthy this afternoon: the positive result only means something once the tool has been shown able to produce the negative one, and vice versa.

## And one I am downgrading, honestly

```
$ grep -rc "strict_namespaces" tests/    ->  0
```

I asked twice for a test that names the parameter rather than matching an exception message three wrappings deep. I still think that is better — a reword of any message on that path silently blinds the control — but **the mutation demonstrably kills the test today**, which was the actual requirement, and I am not going to hold a fifth round over a robustness preference. Treat it as a note for whoever touches those messages next, not as a blocker.

## Where this leaves your four PRs

- **#527** — one restored test from merge. This is the closest.
- **#528** — nine green cells, the API-key-in-boot-log item outstanding.
- **#529** — unchanged since 2026-08-25; the env-only redesign has not started.
- **#512** — I asked earlier today whether to release it, [with the reasoning in full](https://github.com/MakazhanAlpamys/Soup/pull/512#issuecomment-5480161729). No rush on answering, and it is genuinely a question rather than a decision already taken.

Also relevant to you: branch protection is now on `main` with the 9 test cells plus `lint`, `mlx-smoke` and `transformers-floor` required. `strict` is off so you are never forced to rebase — but a red Windows cell now blocks the merge button structurally rather than depending on me noticing, which is why the fix you just pushed mattered more than a red mark usually does.

Restore the positive control and I merge this.

MakazhanAlpamys:
Merged as `cc64c00`. Thank you, and thank you for staying with it — four rounds is a lot to ask, and on an endpoint that executes shell commands it was the right number.

**The last item is closed, and I measured it rather than reading the diff.** Since this feature does not run on Windows, no earlier round could settle it on my box; I ran it in a Linux container:

```
argv neutered to run `true` instead of the command
  -> FAILED test_bash_sandbox_runs_on_strict_namespace_success   (1 failed, 1 passed)
`if result.launch_failed: raise PermissionError` removed
  -> FAILED test_bash_sandbox_raises_on_strict_namespace_failure (1 failed, 1 passed)
reverted
  -> 2 passed
```

Each mutation kills exactly its own control and leaves the other green. That is the property that made this worth four rounds: the negative test alone was green on a build where the sandbox never executed anything, and the positive test alone would be green on a build that ignored a failed `unshare`. Neither direction can now be broken without a named test going red. I also confirmed from the ubuntu CI logs that the positive control **runs** rather than skipping — it is `PASSED`, not `SKIPPED`, on `ubuntu-latest, 3.10`.

The restored test is also written the right way: `sys.platform` skips at the top, `os.unshare` monkeypatched to a no-op so the assertion is about *execution and output* rather than about namespace creation. It tests the thing that can regress.

I checked the 33 deleted test lines individually. All of them assert the OLD behaviour — the 501 stub and its `v0.53.9` marker — which cannot survive a PR whose purpose is to remove the stub. Nothing was weakened.

Two notes on the merge:

1. The breaking change is deliberate and I have said so in the merge commit: `soup serve --host 0.0.0.0` without `--tool-auth-token` now exits 2 instead of warning. Binding an execution endpoint to every interface with no token is not a configuration worth warning about, and your changelog fragment already flagged it as breaking, which is why it went in without a further round.
2. The `strict_namespaces` naming preference I raised twice is still not adopted, and I am still not holding a round over it. The mutation demonstrably kills the test today, which was the actual requirement.

Closes #151.

MakazhanAlpamys:

I raised the `strict_namespaces` naming point twice in review and then downgraded it to a note at merge, on the grounds that the mutation already killed the test. **That was the wrong call and I have written the test.**

It is true that the mutation killed something. It is also true that the only thing tying this sandbox to STRICT namespace isolation was an exception message raised three wrappings below the call — so rewording any message on that path would have left the suite green while the sandbox quietly degraded to best-effort isolation, on an endpoint that executes shell commands. The string `strict_namespaces` appeared nowhere under `tests/` before this.

Verified on Linux in a container, since the sandbox refuses to run on Windows:

```
strict_namespaces=True -> False   -> FAILED test_bash_sandbox_requests_strict_namespaces_by_name
keyword dropped (positional)      -> same test
reverted                          -> 1 passed
```

Issue #264, opened 2026-06-08
Cloud training: add runpod / lambda backends behind 'soup train --cloud'


MakazhanAlpamys:
## Context

Shipped in v0.71.18 (#16): `soup train --cloud modal` renders a Modal.com app from soup.yaml for serverless GPU training. The CLI surface (`--cloud <provider> --gpu <type> [--cloud-submit]`) and the backend package (`src/soup_cli/cloud/`) are already generic, but only `modal` is implemented (`SUPPORTED_CLOUDS = frozenset({"modal"})`).

## Proposed fix path

Add `runpod` and/or `lambda` (Lambda Labs) backends as sibling modules under `src/soup_cli/cloud/`, mirroring `cloud/modal.py`:
- `SUPPORTED_CLOUDS` gains the new provider name(s).
- Each backend exposes `validate_cloud` / `validate_gpu` (closed GPU allowlist for that provider), `render_*_stub` / `plan_*_run` (config base64-embedded as data — **no code/secret interpolation**, same policy as Modal), `write_stub`, and a token-gated `submit_*_run` with a `_*_SUBMIT_OVERRIDE` test seam.
- `commands/train.py` dispatches on the validated `--cloud` value to the right backend.
- New extras (e.g. `[runpod]`) lazy-imported only for live submit; plan-only render needs no dependency.

## Live end-to-end validation (NEW — not just mocks)

Unlike the Modal backend (whose live submit shipped "plumbed but unrun" — see v0.71.18 known limitations), this issue should ship **live-validated**. The maintainer can create real accounts and run one real cheap job per provider:

1. **Modal (backfill)** — sign up at modal.com (~$30 free credits for new accounts), `pip install modal && modal token new`, then run `soup train --cloud modal --gpu t4 --cloud-submit` with a tiny model (SmolLM2-135M, 2-3 steps). This converts the existing v0.71.18 `--cloud-submit` path from mock-validated to live-validated.
2. **RunPod** — create account + `RUNPOD_API_KEY`, top up the minimum ($10), submit one job on the cheapest community-cloud GPU (RTX 3090/4090 spot, ~$0.2-0.4/hr). Verify: job starts, soup-cli installs from PyPI inside the container, training runs, adapter artifact lands in the configured output (volume / S3-compatible), job terminates (no runaway billing).
[...]

PR #528, opened 2026-08-24
feat: add runpod and lambda labs cloud backends (#264)

kok-o:
Refs #264

### Summary of Changes:
- Added support for **Lambda Labs** (`soup train --cloud lambda`) cloud provider with lifecycle controller, auto-termination in `finally`, and zero-secret user_data payloads.
- Added planned preview for **RunPod** (`soup train --cloud runpod`), with CLI dispatch safely descoped pending automated lifecycle safeguards.
- The cloud backend logic is implemented similarly to `modal`, avoiding heavy global SDK imports to maintain fast CLI startup times.
- Refactored the cloud dispatcher in `commands/train.py` to dynamically load the appropriate module based on the `--cloud` flag.
- Updated typer docstrings and help documentation for `--cloud` and `--gpu` parameters.
- Added comprehensive unit tests covering cloud controller script generation, refusal guards, secret-free payload generation, and `_SOUP_YAML` validation. All tests and linting (ruff) pass successfully.


MakazhanAlpamys:
Hi @kok-o - review of the cloud backends below. The red CI is a one-line fix, but there is one functional blocker in the lambda backend.

**Why CI is red (all 9 test cells):** `changelog.d/0.73.3/264.added.md` must start with a Markdown list item (`- `) - `test_repository_fragments_match_the_newest_release` fails on it. Trivial fix.

**Blocker - the lambda backend cannot work end-to-end.** In lambda_labs.py the generated stub is never sent to the API: user_data is built twice identically (lines ~170-182) and the second copy never enters the launch payload ("We will assume they do for this exercise" comment). A runpod submission carries its startup script; a lambda one does not, so the instance boots to nothing. Either wire user_data into the API call or descope lambda to plan-only like the pre-live pattern this repo uses elsewhere. Related and not optional either way: the RUNPOD/LAMBDA key is interpolated as text into cloud-init user_data, which lands in the provider's boot logs - inject via env/file mechanism instead, or state the exposure loudly in --help and docs.

**Other items:**
- Docs are missing: docs/backends-and-ops.md still says Modal-only, docs/commands.md cloud line too.
- Region hardcoded to us-tx-1 - expose or document.
- `getattr(cloud_mod, f"plan_{cloud}_run")` with no default: an allowlist/dispatch mismatch becomes a bare AttributeError; use getattr(..., None) + a named error like the rest of the tree does.
- Behaviour change worth noting: "MODAL" (uppercase) used to normalise, now rejects. Fine if deliberate, but say so in the fragment.
- Third-copy smell again: validate_cloud / validate_gpu / write_stub / submit_* are near-verbatim across modal.py / runpod.py / lambda_labs.py. Extract shared helpers before a fourth backend forks them further.

What I verified positively: lazy imports hold (source-guard test present), all 30 existing modal/cloud tests pass, 28 new tests pass locally, and my mutation (removing the RUNPOD_API_KEY check) died on TestSubmitRunpodRun::test_no_token_raises - that guard is real. Config schema untouched (--cloud stays a CLI-side allowlist), which I am fine with.

Fix the lambda user_data gap + key exposure story + fragment format + docs and this merges.

MakazhanAlpamys:
@kok-o: #528 still has an end-to-end Lambda user_data failure, boot-log secret exposure, missing user docs/changelog, and red CI. Please fix these and add a test proving user_data reaches the Lambda payload before re-review.

Amix29:
I prepared the complete review follow-up as kok-o/Soup#4: https://github.com/kok-o/Soup/pull/4

It fixes the nine-cell changelog CI failure and the functional Lambda blocker: launch `user_data` is now present in the actual payload, contains no API key, and a local controller owns SSH completion, artifact copy, termination in `finally`, and termination verification. It also makes region/SSH configuration explicit, gives RunPod a required persistent network volume, restores case-insensitive provider dispatch, hardens dynamic dispatch errors, extracts the duplicated cloud primitives, and documents both providers.

Validation on the correction commit `4347c5f`: 18,818 tests passed (82 skipped, 5 deselected; 82.76% coverage), full Ruff clean, targeted mypy clean, and both generated controllers compile.

The one remaining acceptance boundary is deliberately not claimed: #264 still requires real paid Modal/RunPod/Lambda jobs and the provider/cost QA record. I do not have those provider credentials, so no synthetic `tests/qa` result was added. Once @kok-o merges the correction PR into `feat/cloud-training-264`, this upstream PR will receive the commit and CI will rerun here.

MakazhanAlpamys:

MakazhanAlpamys (changes requested):
Changes requested; this head is blocked. I read issue #264 and checked the current upstream head.

The focused cloud tests (54 passed, with Ruff clean) and the required mutation are useful, but the end-to-end contract is not met:

- P0: Lambda computes `user_data_b64_injected` but omits it from the launch payload. A mocked request contained only `region_name`, `instance_type_name`, and `quantity`.
- P0: RunPod passes the API key into the remote pod environment, and the Lambda path interpolates credentials into cloud-init; secrets must not be exposed to workload code or logs.
- P0: there is no reliable lifecycle cleanup, termination check, or artifact retrieval.
- P0: the required fragment `changelog.d/0.73.3/264.added.md` does not start with the required Markdown list item, and all nine CI test jobs fail on that validator.
- P1: the CLI rejects uppercase legacy values such as `RUNPOD` before the case-normalizing validator is used.
- P1: backend documentation still describes only Modal.

Please rework the payload, secret boundary, lifecycle/artifact flow, CLI normalization, docs, and changelog, then rebase and rerun the full CI. Do not merge the current head.

MakazhanAlpamys (changes requested):
Re-review of 86e72daf029734c9a647d2eb63a1287ce7b2f1a2: the focused cloud suite passed 24 tests, but all 9 platform test jobs still fail while lint, type-check, mlx-smoke, and transformers-floor pass. The branch remains blocked by the Lambda payload omitting its generated user-data, API secrets being exposed to workload code or cloud-init logs, missing lifecycle cleanup and artifact retrieval, a case-normalization regression in CLI dispatch, and documentation that still describes only Modal. The changelog fragment also fails the repository validator. Rework these issues, rebase onto current main, and rerun the complete matrix; do not merge the current head.

MakazhanAlpamys:
@kok-o — I re-reviewed this from scratch rather than re-reading my old comments, and I verified every claim below against the actual code at head `86e72daf`. The headline first: **nothing has been pushed since 2026-08-24T21:10:22Z**, and `kok-o/Soup#4`, the correction @Amix29 prepared for you, is still `state: open, merged: false`. So every blocker from 2026-08-25 is present verbatim. This is not a re-review of new work; it is the same code.

## The red CI is one character

All 9 test cells fail on a single test, while lint, type-check, mlx-smoke and transformers-floor pass:

```
  ChangelogError: changelog.d/0.73.3/264.added.md:
```

Proved it is this PR and not infra by running the same test on both trees: `main` gives `17 passed`, this branch gives `1 failed, 16 passed`. Your fragment is in the right directory, uses a valid category, is good user-facing English and cites `(#264)` — it just needs a leading `- `. That unblocks all nine cells.

## Blocking: the Lambda backend launches a GPU that never trains and never stops billing

This is the part I need you to look at first. In `src/soup_cli/cloud/lambda_labs.py` the generated stub builds its launch payload **twice, identically**, and neither copy carries `user_data`:

```python
payload = json.dumps({"region_name": "us-tx-1", "instance_type_name": ..., "quantity": 1})
# Actually, Lambda API does not natively support user_data at launch
# according to some API docs. We will assume they do for this exercise,
# or we have to SSH.
payload = json.dumps({"region_name": "us-tx-1", "instance_type_name": ..., "quantity": 1})
```

`user_data_b64_injected` is computed on the line above and then never referenced. I checked the rendered stub with an AST walk rather than by reading: two `json.dumps` payloads, keys `['instance_type_name', 'quantity', 'region_name']` in both, `'user_data' in any payload: False`, and exactly one occurrence of `user_data_b64_injected` — the assignment.

The self-termination `curl` lives *inside* that undelivered `user_data`. So `soup train --cloud lambda --cloud-submit` boots an A100 or H100, runs no training, prints `Training submitted; download checkpoints to ...`, and the instance bills until someone notices. Issue #264's own acceptance criterion is the exact inverse of this: *"terminate the instance in a `finally`, verify termination explicitly, an orphaned A10 bills ~$0.75/hr"*.

Please either wire `user_data` into the launch payload with a local controller that owns termination in a `finally` and verifies it, or descope `lambda` to plan-only for now. I would take the second option happily — a documented "not yet live" is fine, a silent billing hazard is not.

## Blocking: the API key is shipped into the instance, in both backends

- `lambda_labs.py:163` — `user_data = user_data.replace("$LAMBDA_API_KEY", api_key)`. Lines 129-131 of the same file *name the exposure* — "passing it in user_data means it stays in cloud-init logs" — and then do it. `set -x` at line 122 makes bash xtrace echo the **expanded** key into the boot log as well.
- `runpod.py:161` — `env={"RUNPOD_API_KEY": api_key}` puts a full-privilege key, one that can create and terminate any pod on the account, into the pod environment where it is visible in the RunPod console.

Your env-only sourcing is right — `os.environ.get`, no `--api-key` flag, matching the repo rule. The problem is what happens after: the key is forwarded into the machine. `cloud/modal.py`'s docstring states the contract this project holds itself to — *"Never embeds API keys."* Termination belongs to a local controller that holds the credential, not to the instance.

## Mutation testing: 2 of 4 survive, and both survivors are the credential path

```
plan-only default removed (if cloud_submit -> if True)   -> 2 failed   KILLED
GPU allowlist disabled (both backends)                   -> 2 failed   KILLED
literal API key inserted into the rendered Lambda stub   -> 24 passed  SURVIVED
entire API-key-injection path deleted from Lambda        -> 24 passed  SURVIVED
```

I verified the third mutant actually leaked before trusting the green (`secret present in rendered stub: True`), because a mutation that silently fails to apply reports a false pass. **Nothing in this repository asserts that a rendered cloud stub contains no secrets**, even though `modal.py` promises exactly that. Those are the two tests I need, and they are the ones the maintainer review asked for four days ago.

Given this launches billable resources on someone else's account, I will say it plainly: the suite cannot currently tell a secret-leaking stub from a clean one.

## Smaller, but please fix while you are in there

- **`--cloud MODAL` is now rejected.** Dispatch runs `if cloud == "modal"` on the raw string *before* `validate_cloud` normalises it, so case-insensitivity regressed: `exit_code = 2`, `Invalid --cloud: MODAL`. This worked before your PR and the fragment does not mention it.
- `getattr(cloud_mod, f"plan_{cloud}_run")` has no default, so any dispatch/allowlist mismatch surfaces as a bare `AttributeError`. Reproducible.
- `lambda_labs.py:43` ships `"rtx-4090": "gpu_1x_rtx6000ada",  # mapping 4090 to closest available?` — a question mark in shipped code. The user asks for a 4090 and silently gets a different, more expensive GPU.
- The RunPod stub prints "download checkpoints to ./out" but creates no volume and then self-terminates, so even the happy path destroys the trained adapter.
- `region_name` is hardcoded to `us-tx-1` with no way to change it.
- `validate_cloud` / `validate_gpu` / `write_stub` / `submit_*` are now near-verbatim in three modules, and `modal.SUPPORTED_CLOUDS` still reports `['modal']` while `train.py` keeps its own hardcoded dispatch list. Two sources of truth, and this repo has been bitten by that three times (#372, #392, #424).
- No docs. `docs/backends-and-ops.md` and `docs/commands.md` still document `--cloud modal` only.

## What is genuinely right, so you know what not to touch

SSRF is clean — both endpoints are hardcoded literals, no user input reaches a URL. `subprocess` uses list args with `shell=False`. Stub writes go through `atomic_write_text` and `enforce_under_cwd_and_no_symlink`. Plan-only is the default and the mutation proving it is caught. Config is base64-embedded as data. `validate_cloud` / `validate_gpu` are closed allowlists with real tests. No test anywhere in the diff was deleted or weakened — I checked, the diff of `tests/` has zero removed lines. The 18 deleted lines in `commands/train.py` are all replaced in place and no guard was lost.

## Fastest path

Merge `kok-o/Soup#4` into your branch, rebase on current `main`, fix the fragment's leading `- `, and push. Then I will re-run the mutation set on the new head.

Separately, and with no judgement: you have four PRs open here and only #527 has moved this week. If you would rather concentrate on one or two, tell me which and I will close the others as stale so your queue reflects what you are actually working on. Nobody loses credit for a PR that gets reopened later.

MakazhanAlpamys:
@kok-o — you pushed on 2026-08-30 and I owed you this re-review. Real progress on three of the four blockers; the security one is untouched.

## Fixed

- **The red CI.** `changelog.d/0.73.3/264.added.md` now starts with `- `.
- **The functional Lambda blocker.** `user_data` is in the launch payload (`lambda_labs.py:157`), and the duplicated identical `json.dumps` block is gone. An instance will now actually receive its cloud-init script instead of booting and running nothing.
- **The triplication.** `src/soup_cli/cloud/base.py` extracts the shared primitives and `modal.py` shrinks by 88 lines. That was the "third copy" smell and you took the structural fix rather than patching around it.

## Still blocking: the API key is shipped into the instance

Unchanged from 2026-08-25:

```
lambda_labs.py:104   "set -x\n"
lambda_labs.py:145   user_data = user_data.replace("$LAMBDA_API_KEY", api_key)
runpod.py:140        env={"RUNPOD_API_KEY": api_key},
```

The Lambda key is substituted into cloud-init `user_data`, and `set -x` makes bash xtrace echo the **expanded** value into the boot log — readable by anyone who can view instance logs. The RunPod key goes into the pod environment, visible in the console, and it is a full-privilege key that can create or terminate any pod on the account.

`cloud/modal.py`'s docstring states the contract this project holds itself to: *"Never embeds API keys."* Your env-only sourcing is right — `os.environ.get`, no `--api-key` flag — the problem is what happens after: the credential is forwarded into the machine.

The reason it is there is termination, and that is the thing to change. Right now the instance owns its own shutdown (`curl .../terminate` inside `user_data`), so it needs a credential. **Termination belongs to a local controller that holds the key and calls terminate in a `finally`**, which is also what #264's acceptance criterion asks for — "terminate the instance in a `finally`, verify termination explicitly, an orphaned A10 bills ~$0.75/hr". If you would rather not build the controller now, descope both backends to plan-only and say so in the docs; a documented "not yet live" is fine, a credential in a boot log is not.

The two mutations from my last review are still the ones to close: inserting a literal key into the rendered stub, and deleting the key-injection path entirely, both passed the whole suite. Nothing asserts that a rendered cloud stub contains no secrets, even though `modal.py` promises exactly that.

## Needs a decision from me, and you should not have made it in the fragment

> Note that `--cloud MODAL` (uppercase) now rejects as a deliberate behaviour change.

It worked before your PR. Declaring a regression deliberate in a changelog line is not the same as it being agreed — that is a default-behaviour change and it needs to be argued, not annotated. **My call: restore case-insensitivity.** `validate_cloud` already normalises; the dispatch just runs `if cloud == "modal"` on the raw string *before* it. Move the dispatch after normalisation and the regression disappears without any contract change. If you think uppercase should be rejected, open that as its own issue and I will consider it on its own terms.

## Still open from before

- No docs. `docs/backends-and-ops.md` and `docs/commands.md` still document `--cloud modal` only.
- `getattr(cloud_mod, f"plan_{cloud}_run")` with no default still surfaces any dispatch/allowlist mismatch as a bare `AttributeError`.
- `lambda_labs.py` still ships `"rtx-4090": "gpu_1x_rtx6000ada",  # mapping 4090 to closest available?` — a question mark in shipped code, silently giving the user a different and pricier GPU.
- The RunPod stub still prints "download checkpoints to ./out" with no volume created and then self-terminates.

Close the credential path — or descope to plan-only — and restore the uppercase dispatch, and I will re-run the full battery. @Amix29's `kok-o/Soup#4` on your fork is still unmerged and covers part of this.

MakazhanAlpamys (changes requested):
Reviewed at `5775a3f`. **The Lambda rework is exactly what I asked for and it is good work. Blocking because the same fix was not applied to RunPod, and the mutation that proves it is the identical one I flagged on 2026-08-29 and again on 2026-08-31.**

## What is genuinely closed

- **`set -x` is gone.** `lambda_labs.py:101-102` is now `#!/usr/bin/env bash` + `set -euo pipefail`. No xtrace anywhere.
- **The Lambda key no longer enters `user_data`.** The `.replace("$LAMBDA_API_KEY", api_key)` line is deleted; `user_data` is pip-install, base64-decode of the config, `soup train`. The key is read only in the local controller and sent over HTTPS Basic auth.
- **Termination is now owned locally**, in the caller's `finally`, with `_wait_for_termination` verifying it rather than assuming. That is the design I specified, and `test_generated_controller_always_terminates` makes it load-bearing rather than incidental.
- The triplicated primitives are extracted into `cloud/_common.py` (`modal.py` minus 88 lines) — the structural fix rather than a patch-around, and the 30 pre-existing Modal tests pass unchanged.
- `--cloud MODAL` case-insensitivity restored, `getattr` given a default with a named error, `us-tx-1` replaced by a regex-gated `LAMBDA_REGION`, and the `"rtx-4090"` guess with its `# closest available?` comment removed in favour of an honest allowlist.

Three of the four blockers are properly closed. **Zero deleted test lines** (`git diff origin/main...5775a3f0 -- tests/` gives 0 removed), and I checked that the `_common.py` extraction did not quietly drop a guard.

## The blocker: RunPod still ships the key into the pod

My 2026-08-31 review named three lines. Two are fixed. The third is **verbatim unchanged**:

```
src/soup_cli/cloud/runpod.py:153   env={"RUNPOD_API_KEY": api_key},
src/soup_cli/cloud/runpod.py:119   curl -X POST https://api.runpod.io/graphql -H "Authorization: Bearer $RUNPOD_API_KEY"
```

and `runpod.py:111-113` now documents the intent in a comment rather than removing it. So a full-privilege key — one that can create or terminate any pod on the account — is placed in the pod environment, visible in the RunPod console and readable by every process in the container, including the third-party wheels pip installs at boot.

**I reproduced the surviving mutation on this head rather than citing the old one.** I replaced the stub's `api_key = os.environ.get("RUNPOD_API_KEY")` with a literal `api_key = "rpa_LIVE_SECRET_ABCDEF123456"`, then rendered the stub:

```
SECRET PRESENT IN RENDERED STUB: True
context: def main() -> None:\n    api_key = "rpa_LIVE_SECRET_ABCDEF123456"\n
tests/test_issue264_cloud.py:  32 passed
```

A stub written to disk with a live credential baked into it is indistinguishable, to this suite, from a clean one. `cloud/modal.py`'s docstring still promises *"Never embeds API keys"* and nothing in the repo enforces it.

**A second survivor, and it is the scenario #264 names by name.** I deleted the in-container `podTerminate` block entirely — the pod's only shutdown mechanism — and confirmed it was really gone from the rendered stub (`podTerminate still in stub: False`): **32 passed**. `runpod.py:136-159` calls `create_pod(...)` and returns; there is no wait, no `finally`, no terminate, no verification. If the image lacks `curl`, or the container is OOM-killed, or `bash -c` never reaches that `;`, the pod bills until someone notices — and no test can tell.

That asymmetry is the whole finding: Lambda has a `finally`-owned controller and a test that dies when you remove it; RunPod has neither, and the two backends sit behind the same flag.

## Docs make it worse rather than neutral

`docs/backends-and-ops.md` says of Lambda: *"The API key stays on the caller and is never embedded in cloud-init or instance logs."* True. The RunPod section says nothing about the key entering the pod, and the shared intro says credentials are *"read from the environment only when a live submission starts"* — which describes **sourcing**, not **forwarding**. A reader reasonably concludes both backends carry the Lambda guarantee.

## What I am asking for — your choice of two

1. **Port the Lambda pattern to RunPod.** `main()` keeps the key locally, polls to completion through the SDK, calls `terminate_pod(pod_id)` in a `finally` with verification. Drop `env={"RUNPOD_API_KEY": ...}` and the in-container `curl`. This closes both survivors at once.
2. **Descope `runpod` to plan-only** — `submit_runpod_run` raises a friendly "live submit not yet available for RunPod; use `--cloud modal` or `lambda`", documented as such. I said before that a documented "not yet live" is fine, and I meant it; it is not a lesser outcome.

Either way, one more thing, and it is the item that has now survived three rounds:

- **One general no-secrets test, applied to all three backends.** Put a sentinel in the environment, render each stub, assert the sentinel appears in neither the stub text nor any provider payload. The current Lambda guard is `assert "LAMBDA_API_KEY" not in payload["user_data"]` — name-specific, so it correctly pins that exact regression and misses every other secret. Injecting a differently-named live secret into the same `user_data` passes all 32 tests. Write the property once and it kills the RunPod mutation too.

## Smaller items, non-blocking

- `lambda_labs.py:194` — `StrictHostKeyChecking=accept-new` is TOFU, and the config *and* the trained adapter cross that channel. Defensible for ephemeral hosts; make it a stated decision in a comment rather than an inline option.
- `lambda_labs.py:166,180` — `instance_id` comes from the provider's launch response and is concatenated into a URL path. Only exploitable by a hostile or compromised Lambda API, but a shape guard is two lines.
- `runpod.py:134` — `_LOCAL_OUTPUT` is rendered into the stub and never used, implying a copy-back that does not exist.

## Hygiene before this lands

- **Rename the fragment to `528.added.md`.** It is currently `264.added.md` — the issue number. Every neighbour uses the PR number (`468.added.md`, `490.changed.md`, `513.added.md`). CI does not check this, which is exactly why it needs saying.
- **Add the credit**: the line ends `(#264).`; the convention here is `(#264 by @kok-o in #528)`.
- **`Refs #264`, not `Closes #264`.** The issue's live-validation criteria — one real paid job per provider and a cost log under `tests/qa/` — are not met, and your own docs say so honestly. Over-claiming in the merge commit would contradict the docs in the same release.

## Verification

`tests/test_issue264_cloud.py` plus `test_v07118.py`, `test_cli_startup_is_light.py` and `test_issue487_changelog_fragments.py` give **197 passed**. `ruff` clean. CI **13/13 green**. The clean parts hold: token sourcing is env-only with no CLI flag and nothing from `soup.yaml`; both API endpoints are hardcoded literals with no override, so there is no SSRF path; `subprocess` is argv-list with `shell=False`; config is base64 and `output_dir` is regex-gated and rejects `..`; TLS is never disabled; path containment uses this repo's `realpath` + `commonpath` helper rather than `Path.resolve().relative_to()`.

Mutations that were killed, so you know which of your guards are doing work: re-introducing the Lambda key export into `user_data` (killed), deleting `user_data` from the launch payload (killed), making the allowlist permissive (killed, 4 tests), disabling both token-presence guards (killed, 2 tests), and disabling terminate-in-`finally` in the generated Lambda controller (killed). The Lambda half is genuinely well tested. It is the RunPod half that has no equivalent.

MakazhanAlpamys (changes requested):
**The blocker that survived three rounds is genuinely closed. Blocking again, because the way it was closed removed the pod's only shutdown — and the path that still emits one is the default.**

## First, real credit

`env={"RUNPOD_API_KEY": api_key}` is **gone** from `create_pod`. I re-ran the mutation that survived three reviews — a literal `api_key = "rpa_LIVE_SECRET_ABCDEF123456"` baked into the rendered stub:

```
FAILED tests/test_issue264_cloud.py::TestCloudNoSecrets::test_no_secrets_in_stub_or_payload
1 failed, 30 passed
```

Dead. `TestCloudNoSecrets` covers all three backends and asserts on stub text *and* the Lambda launch payload, which is exactly what I asked for. Zero deleted test lines against `main`, and I checked the one removal against your previous head — dropping `assert "LAMBDA_API_KEY" not in payload["user_data"]` is **not** a weakening, because `test_render_happy_path` now asserts full equality against `_USER_DATA`, which is strictly stronger. Fragment renamed to `528.added.md`, credit line present, and you used `Refs` in it without being asked.

## The blocker: RunPod now emits a pod that nothing terminates

The secret left the pod because the `curl ... podTerminate` that needed it was deleted — and nothing replaced it. Measured on your rendered stub:

```
podTerminate   in stub: False
terminate      in stub: False
finally        in stub: False
create_pod     in stub: True
```

The generated `main()` is `create_pod(...)`, two `print`s, return. No wait, no `finally`, no verification.

I checked whether that is reachable, because `submit_runpod_run` now raises `NotImplementedError` and the descope could have covered it. It does not:

```
$ soup train --config soup.yaml --cloud runpod --gpu rtx-4090
│ Stub:     soup_runpod_app.py   │
│ Run: python soup_runpod_app.py │
Note: plan-only. Authenticate with runpod, then run the command above.
EXIT=0
```

Plan-only is the default and now the **only** reachable RunPod path, and it writes an executable script that creates a GPU pod, then instructs the operator to run it. On the billing axis this head is **worse than the one I last reviewed**: that one at least attempted self-shutdown. #264's acceptance criterion is "job terminates (no runaway billing)", and a user following our own printed instruction gets a pod that bills until they notice.

The descope stopped at `submit_runpod_run` and did not reach the path a user actually takes.

**The asymmetry is testable and the suite proves it can be tested.** Removing terminate-from-`finally` in `lambda_labs.py` fails `test_generated_controller_always_terminates`. `grep -i terminate tests/` returns Lambda hits only. So the suite can detect "the generated controller stopped terminating" — it just has no RunPod equivalent, which is why deleting RunPod's shutdown entirely was invisible.

## Two ways out, your choice — (a) is smaller

**(a) Finish the descope.** Refuse `--cloud runpod` at the dispatch in `commands/train.py` with the same friendly message `submit_runpod_run` already gives, so nothing reachable renders or writes a pod-creating script. A documented "not yet live" is a fine outcome and I said so two rounds ago.

**(b) Give `main()` lifecycle ownership** — poll to completion, `terminate_pod(pod_id)` in a `finally`, verify, mirroring what you already built for Lambda.

Either way, one test that fails if RunPod stops terminating (or fails if the backend stops refusing) — the RunPod counterpart of the Lambda test that already exists.

## Stale text the deletions left behind

The comments now describe code that is gone, in **both** directions, which is the confusing kind:

- `runpod.py:10` — *"auto-terminates the pod upon completion."* False.
- `runpod.py:12-15` — *"`--cloud-submit` attempts a live submit... `_RUNPOD_SUBMIT_OVERRIDE` keeps the submit path testable."* False; submit always raises. `_RUNPOD_SUBMIT_OVERRIDE` at `:55` is now dead code whose only other reference is that docstring, and ruff will not catch it.
- `runpod.py:111-112` — *"then tells RunPod API to terminate the pod. We pass the RUNPOD_API_KEY from the environment into the pod so it can self-terminate."* This documents a removed safeguard **and** a removed vulnerability as if both were current.
- `docs/backends-and-ops.md:201-204` documents a command that always fails now (`--cloud-submit` → `NotImplementedError`).
- `docs/backends-and-ops.md:198` — *"so the adapter remains available after the pod terminates"* — nothing terminates it.

The Lambda/RunPod docs asymmetry I raised is genuinely resolved, since the key no longer enters the pod.

## Two smaller items

**PR body still says `Closes #264`.** Merging as-is auto-closes an issue whose live-validation criteria — one real paid job per provider and a cost log under `tests/qa/` — are unmet, and your own docs in this PR say so honestly. `Refs #264`, matching what you already put in the fragment.

**`pytorch-smoke` never ran** — the branch is 11 commits behind and the job post-dates its merge-base. That is my doing (I made it required after it landed in #601), not a defect in your branch; a rebase picks it up, and you need one anyway.

## Non-blocking, worth knowing

`TestCloudNoSecrets` enumerates four env var names and greps for one literal. Injecting a differently-named live secret into all three rendered stubs is undetected:

```
runpod leaked HF_TOKEN into stub: True
lambda leaked HF_TOKEN into stub: True
modal  leaked HF_TOKEN into stub: True
=> 145 passed
```

A real `soup train` environment holds `HF_TOKEN` and `WANDB_API_KEY`. This is still a large improvement on what was there and I am not holding the PR for it — recording it so it does not get forgotten.

`196 passed` across the cloud tests and siblings, `ruff` clean, ANSI scanner green. The remaining work is genuinely small, and (a) is a handful of lines.

MakazhanAlpamys:
@kok-o — re-reviewed at `c3e2874` and this is the closest it has been. **Four of the six items are genuinely fixed and I verified each by execution, not by reading the diff.** Three things left, and only one of them is real work.

## Verified fixed

- **The Lambda credential path — you took resolution (a), the controller.** `lambda_labs.py:100-118` builds `user_data` with `set -euo pipefail`, no `set -x`, and no key reference of any kind. The key is read by the *local* controller (`:225`) and used only in a local `Authorization` header (`:144`). Termination is a real `finally` (`:256-267`) that terminates **and** polls `_wait_for_termination` to confirm, forcing `rc = 1` if it fails. That is more than #264 asked for — the criterion said terminate in a `finally`, you also verify it.
- **Uppercase dispatch.** `train.py:645` `cloud = cloud.lower()` now precedes dispatch. Executed rather than read:

  ```
  --cloud MODAL   -> past dispatch (fails later on the cwd containment guard, correctly)
  --cloud LAMBDA  -> past dispatch
  --cloud RunPod  -> "RunPod cloud training is not yet live; use --cloud modal or lambda."
  --cloud bogus   -> "Invalid --cloud: bogus. Supported: modal, lambda (runpod is planned)."
  ```

  Mixed case reaches the descope, which is the case the naive `.lower()` placement would have missed.
- **The bare `AttributeError`.** `train.py:673-675` uses `getattr(..., None)` with an explicit raise, mapped to exit 2 (plan) and exit 1 (submit).
- **The `rtx-4090` question mark.** Gone; the allowlist is `a10/a100/h100/a6000` and `--cloud lambda --gpu rtx-4090` exits 2 naming the valid list. Shipping a comment that ends in `?` while silently substituting a pricier GPU was the worst line in the original diff and you deleted it rather than answering the question in a comment.
- **The RunPod volume.** `runpod.py:126-146` hard-requires `RUNPOD_NETWORK_VOLUME_ID`, mounts `/workspace`, and the self-termination curl is gone.

## Blocking: the RunPod half of the credential mutation still survives

My last review named **two** mutations to close. One is closed, one is not, and the asymmetry is the finding rather than the failure. Both run at this head, in an isolated worktree with `PYTHONPATH` pinned to the branch (verified: `soup_cli.__file__` resolves inside the worktree, so the mutations are not silent no-ops):

| mutation | result |
|---|---|
| inject the key into Lambda's `user_data` | **KILLED** — `TestCloudNoSecrets::test_no_secrets_in_stub_or_payload` |
| restore `env={'RUNPOD_API_KEY': api_key}` in the rendered RunPod stub's `create_pod` call | **SURVIVED** — 31/31 in `test_issue264_cloud.py`, 135 passed across every cloud-related test |

The one-line diff that survives:

```diff
          '        volume_mount_path="/workspace",\n'
+        "        env={'RUNPOD_API_KEY': api_key},\n"
          f'        docker_args={docker_args!r},\n'
```

**Why your guard cannot see it, which is worth knowing rather than just patching.** `TestCloudNoSecrets` greps the *rendered text* for literal secret values, and for Lambda it additionally executes `_launch_payload()` and inspects the result. The RunPod key is forwarded at **stub runtime**, not substituted at render time, so the value never appears in the text and a text grep structurally cannot catch it. The Lambda side is guarded because you exercise the payload; the RunPod side is guarded only against a defect shaped like Lambda's.

So the exact original defect is one line and zero red tests away. It is not *reachable* today — I grepped, and `--cloud runpod` is refused at `train.py:647` before any import — so this is a regression risk, not a live exposure, and I am not treating it as a security emergency. But an unguarded path that ships is how it comes back.

**Two ways to close it; pick either and I will not argue:**

1. **Delete `render_runpod_stub` / `plan_runpod_run` and the `runpod = ["runpod>=1.3.0"]` extra.** My preference. The CLI refuses the backend, so shipping a renderer *and* a PyPI extra for it is dead surface, and `test_runpod_extra_in_pyproject` currently pins that dead surface into the suite. Re-add it with the live work.
2. **Keep it and mirror what already works for Lambda:** `exec()` the rendered RunPod stub with a fake `runpod` module and assert `create_pod` receives no key-bearing kwarg. That is the guard that would have failed my mutation.

## Blocking, but trivial: two user-facing texts describe behaviour that does not ship

- `changelog.d/0.73.3/528.added.md` says `soup train --cloud runpod` **"provides a planned preview"**. It provides no preview — it exits 2 with a refusal and renders nothing.
- `docs/backends-and-ops.md:177` still reads `soup train --cloud modal|runpod|lambda` renders a provider-specific controller. The correction sits 20 lines below under **### RunPod (Planned)**, so a reader who stops at the first line is told the opposite of the truth. `docs/commands.md:25` is already correct (`modal|lambda`) — make line 177 match it.

## Not blocking, worth fixing while you are in there

`--gpu` help (`train.py:491-492`) lists a cross-provider union including `rtx-4090` and `a10g`. Neither is accepted by any reachable provider — Lambda's allowlist spells it `a10`.

## For the record, and it is not a criticism

`Refs #264` rather than `Closes` is right. Of #264's five **live-validation** criteria, zero are met: there is no `tests/qa/` entry for this and no real instance has ever been launched or terminated — Lambda's termination correctness is proven only against a mocked `_request_json`. Your `docs/backends-and-ops.md:223` says exactly this, which is why it is a record and not a finding. When this ships, the release notes will describe `--cloud lambda` as plumbed-but-unrun, the same wording v0.71.18 used for Modal.

Also verified: **zero deleted or weakened test lines** (`git diff main...pr528 -- tests/ | grep -c '^-[^-]'` → 0), and the `cloud/_common.py` extraction preserves the Modal error-message text verbatim, so the pre-existing Modal tests pass unchanged. The single red test in my local run (`test_v07118.py::TestTrainCloudCli::test_flags_in_help`) reproduces identically on `origin/main` — it is this box's typer/click breakage, not yours.

Fix the two texts, close the RunPod mutation either way, and I will re-run the full battery on the next push.

MakazhanAlpamys:
Merged as `0405611`. Thank you — and the thing worth naming is not the final diff, it is the distance between revision one and revision three.

**You changed the design instead of patching the symptom.** The first version put the key in `user_data` because the instance owned its own shutdown, and that is a real reason, not carelessness — the credential was load-bearing. The obvious fix would have been to drop `set -x` and call it fixed, which would have left a full-privilege key inside a machine that did not need one. Moving termination to a local controller removed the *need* for the credential to travel at all, and then you went further than #264 asked: the criterion says terminate in a `finally`, and yours also polls to confirm and forces a non-zero exit when the termination itself fails.

**Twice you took the structural fix over the convenient one.** `cloud/_common.py` instead of a third copy of the primitives; and when I said the RunPod renderer was dead surface keeping the original defect one line away, you deleted the module and the PyPI extra rather than wrapping a guard around code nobody could reach.

**What I verified independently at `9e64101`**, in an isolated worktree with `PYTHONPATH` pinned to the branch (unpinned, the editable install resolves to the main tree and every mutation is a silent no-op):

```
--cloud runpod / RunPod  -> exit 2, friendly refusal, no ImportError after the deletion
--cloud MODAL            -> renders (case-insensitivity holds)
--cloud bogus            -> "Supported: modal, lambda (runpod is planned)"
pyproject                -> no runpod extra
render_runpod_stub / plan_runpod_run -> no references in src, tests, docs
deleted test lines vs main -> 0
```

And the mutation that decided it — the *class* of defect that survived last round, a secret that never appears in the rendered text but reaches the launch payload at stub runtime:

```diff
          "        'user_data': _USER_DATA,",
+         "        'env': {'LAMBDA_API_KEY': __import__('os').environ.get('LAMBDA_API_KEY')},",
```

**KILLED** by `TestCloudNoSecrets::test_no_secrets_in_stub_or_payload`. At the previous head the equivalent mutation left all 135 cloud tests green, because the guard grepped rendered text and the key was forwarded at runtime where no text grep can see it. It now `exec()`s the stub and inspects the payload, so both shapes are covered. That is the difference between a test that says "no secret in this string" and one that says "no secret reaches the provider".

**One line for the release notes, which you should know about:** the `soup-cli[runpod]` extra disappears, so `pip install 'soup-cli[runpod]'` will start warning for anyone who used it. Nothing reachable breaks — the CLI refused that backend anyway — but a name vanishing from `pyproject.toml` is a user-visible fact and it will be stated.

`Refs #264` rather than `Closes` is right, and your `docs/backends-and-ops.md` line saying these paths still need the paid live-validation checklist is why I did not have to argue it. `--cloud lambda` ships plumbed-but-unrun, the same status v0.71.18 gave Modal.

Non-blocking leftover for whenever you are next in the file: `docs/backends-and-ops.md:223` still refers to a "RunPod … submission path" that no longer exists.

CONTRIBUTORS.md updated. #529 is still open with your name on it — no pressure, and my question there stands: if you are parking it, say so and I will stop looking at it until you come back.

Issue #318, opened 2026-07-19
Wire the opt-in telemetry flywheel (deferred from v0.71.41)


MakazhanAlpamys:
The telemetry primitives in `utils/trackers.py` (`build_telemetry_payload` / `send_telemetry_payload` / `is_telemetry_enabled`) are complete, SSRF-hardened, 1s-timeout, default-OFF, and hardware-info-only — but **every call site is in `tests/`**; nothing in `src/` ever builds or sends a payload. v0.71.41 shipped the `soup reward stress` half of the plan and deferred telemetry (and corrected the ops-docs that falsely claimed a live sender).

To light it up, ALL of the following are required:

1. **A published public privacy policy** — the stated hard blocker (`trackers.py:10-11`, `docs/backends-and-ops.md`). Lands first, or nothing else ships.
2. **A real analytics key** — `_POSTHOG_DEFAULT_KEY = "phc_soup_public_write_only"` is a placeholder, not valid.
3. **`httpx` as a core dependency** — currently lazy-imported and not in core, so a plain `pip install soup-cli` silently no-ops the sender.
4. **An anonymous `distinct_id`** — the payload has none today, so the backend can't dedupe.
5. **A real call site** in `src/` (command entrypoint).
6. **`--no-telemetry`** flag + first-run consent (env-var-only today).

### Honesty constraint
The payload must stay hardware-info-only (`soup_version` / `command` / `python` / `os` / `arch` / optional `duration_seconds`) — **never** paths, model names, or config contents. Telemetry is a trust decision; a botched rollout costs more than the data is worth.

### Acceptance
- [ ] Public privacy policy published and linked from the docs
- [ ] Real key + `httpx` core dep + anonymous `distinct_id`
- [ ] Call site in `src/` behind default-OFF + `--no-telemetry` + first-run consent
- [ ] Payload asserted hardware-only in tests
- [ ] `docs/backends-and-ops.md` Telemetry section updated to reflect the live (opt-in) state

MakazhanAlpamys:
Shipped in #529 by @kok-o, merged as `5f43a04`.

Try:

```bash
SOUP_TELEMETRY=1 soup train --config soup.yaml     # opt in
soup train --config soup.yaml --no-telemetry       # opt out per invocation
```

## The six requirements, and where each landed

| # | requirement | outcome |
|---|---|---|
| 1 | published public privacy policy — the stated hard blocker | **MET, on my ruling** — see below |
[...]

Issue #154, opened 2026-05-13
v0.53.9+: PostHog telemetry — real write-only key + env override


MakazhanAlpamys:
v0.53.8 shipped \`send_telemetry_payload\` opt-IN with a placeholder PostHog key:

\`\`\`python
_POSTHOG_DEFAULT_KEY = "phc_soup_public_write_only"
\`\`\`

When \`SOUP_TELEMETRY=1\`, the sender POSTs to PostHog with this fake key and receives 4xx (silently swallowed) — so users opting in never actually send. By design until a public privacy policy ships.

**Fix path:**
- Add \`SOUP_POSTHOG_KEY\` env var; \`is_telemetry_enabled()\` short-circuits to False when neither env var nor a real default key is configured
- Update the \`_POSTHOG_DEFAULT_KEY\` to a real public write-only key once the privacy policy lands
- Document opt-IN behaviour in the privacy policy + README

**Acceptance criteria:**
- Telemetry stays opt-IN
- No telemetry sent when no real key is configured (even with \`SOUP_TELEMETRY=1\`)
- Existing silent-fail / 1s-timeout policy preserved

MakazhanAlpamys:
Shipped in v0.53.10. See [release notes](https://github.com/MakazhanAlpamys/Soup/releases/tag/v0.53.10).

Try: `SOUP_POSTHOG_KEY=phc_yours SOUP_POSTHOG_ENDPOINT=https://eu.i.posthog.com/i/v0/e/ SOUP_TELEMETRY=1 soup train --config soup.yaml`

PR #529, opened 2026-08-24
feat: telemetry rollout (#318)

kok-o:
### Summary

Wires an opt-in, anonymous, hardware-only telemetry flywheel to understand supported operator environments and platform distribution (#318).

### Architectural Choices & Recorded Divergences

- **Default-OFF with zero interactive prompts**: We explicitly chose an environment-variable-only opt-in (`SOUP_TELEMETRY=1`) with a strict **default-OFF** posture rather than an interactive `rich.prompt.Confirm` prompt. This avoids prompt fatigue and guarantees Soup never interrupts scripts, automated pipelines, or headless CI runners. When `SOUP_TELEMETRY` is unset, `0`, or falsy, telemetry is completely off and silent. The code only reads `SOUP_TELEMETRY` and never sets it.
- **`--no-telemetry` CLI flag**: Explicitly guarantees zero network requests and zero payload building, unconditionally overriding `SOUP_TELEMETRY=1` in the environment.
- **Standard library HTTP delivery**: Uses Python's standard library `urllib.request` with a 1.0s connect/read timeout (DNS excluded) and fail-soft exception swallowing. There are **zero new core dependencies** (`httpx` is not moved to core dependencies in `pyproject.toml`). CLI startup remains lightweight (<0.5s).
- **Command token sanitization (Path leak prevention)**: The command name is validated against Typer/Click registered commands. File paths (e.g. `soup /home/alice/data.jsonl`), model directories, and typos are masked as `(unknown)`. No paths, model names, hyperparameters, configs, or secrets ever leave the machine.
- **Anonymous identifier behind consent**: A random UUID4 is saved to `~/.soup/telemetry_id`. Constructing payloads while telemetry is disabled has **zero disk side-effects** (returns an in-memory ephemeral UUID). The persistent identifier only comes into existence once an operator has opted in. Deleting `~/.soup/telemetry_id` completely resets the identifier, which is regenerated on the next opt-in.
- **Placeholder key guard**: Merging is decoupled from account provisioning: `send_telemetry_payload` refuses transmission while the key is the placeholder (`phc_soup_public_write_only`), writing a one-line notice (`Notice: telemetry is not yet live.`) to `stderr`. Pointing `SOUP_POSTHOG_KEY` to a live key activates transmission with zero behavioural changes.
- **Privacy Policy & Security disclosure**: Published in `docs/backends-and-ops.md`, referenced in `SECURITY.md` and `README.md`.
- **Changelog**: Added fragment in `changelog.d/0.74.0/529.added.md`.


MakazhanAlpamys:
Hi @kok-o - thanks for taking #318. I reviewed the PR end to end, and before the technical list there is a product decision from my side that changes the shape of this PR.

**Design direction: env-only opt-in, no first-run prompt.** Please drop the rich.prompt consent flow (and the ~/.soup/telemetry_consent file) and make SOUP_TELEMETRY=1 the single switch that turns telemetry on; unset/0 means off, silently. Two reasons: (1) Soup's core promise is a light CLI with no surprise network - a prompt on first run of ANY command is hostile to scripted/CI usage, where it already silently skips, so consent would be inconsistent anyway; (2) your own implementation demonstrates how much complexity the consent path carries - see blocker 3 below. Related: keep httpx OUT of core deps. One fire-and-forget POST with a 1s timeout does not justify pulling httpcore/h11/anyio/certifi/idna into every install of a package whose lightness we guard with tests - use urllib.request.

Technical blockers found in review:

1. **Red CI, all 9 test cells:** `test_no_user_data_in_payload` (the existing privacy allowlist test in test_v0430_part_a.py) fails because you added `distinct_id` to the payload without updating the allowlist contract. The field itself is fine (random uuid4, not derived from anything user-identifying) but the schema change must go through its guard.
2. **Red CI, lint:** 5 ruff errors in test_v07141_telemetry_prompt.py (I001 x2, F401 x3).
3. **Real bug + dead path:** in get_or_create_distinct_id (trackers.py:114) the `id_file.exists()` call sits outside the try/except, so exceptions there are not swallowed as designed. And separately: the consent file is written on "yes", but the production gate send_telemetry_payload re-checks is_telemetry_enabled() which reads ONLY the env var - so after agreeing once, telemetry never fires again unless the env var is also set. The documented persistent opt-in did not exist. The env-only direction removes this whole class.
4. **Zero guards on the two properties that matter most.** My mutations: (a) made telemetry send WITHOUT any consent - nothing failed (the emit tests are smoke-only, no assert on non-sending); (b) broke --no-telemetry - nothing failed either. Whatever shape this lands in, it needs tests asserting non-delivery when disabled and delivery when enabled, at the send_telemetry_payload boundary.
5. No changelog fragment (changelog.d/0.73.3/529.added.md or similar, mentioning #529).

The payload itself is well-designed - version/command-token/os/arch/duration only, no paths or model names, HTTPS-only with SSRF revalidation - keep that exactly as is. With the env-only simplification this should actually get smaller. Happy to re-review same-day on push.

MakazhanAlpamys:
@kok-o: #529 needs the agreed env-only opt-in redesign, urllib instead of httpx, error-path and allowlist fixes, load-bearing consent/no-telemetry tests, and a changelog fragment. Green CI does not address these functional blockers.

Amix29:
I prepared the full review fix in kok-o/Soup#1 because GitHub does not grant my account direct write access to the contributor fork despite `maintainerCanModify`. The correction PR targets `feat/telemetry-318`, so authorship and the existing upstream PR are preserved.

It implements the env-only opt-in design, removes the prompt/consent file/CLI override, replaces telemetry `httpx` usage with stdlib `urllib.request`, adds load-bearing disabled/enabled network-boundary tests, validates the anonymous UUID, updates the closed payload allowlist/docs, and adds the changelog fragment.

Validation: 141 telemetry-focused tests + 130 core-dependency/CLI tests passed; Ruff is clean.

MakazhanAlpamys:

MakazhanAlpamys (changes requested):
Changes requested; this head is blocked. I read issue #318 and verified the current upstream head, not the follow-up commits mentioned in the fork.

The focused telemetry tests (147 passed), full Ruff, and the required mutation are green, but the implementation still has blocking contract problems:

- P0: persisted consent does not enable telemetry in a new process; the consent file can say enabled while `SOUP_TELEMETRY` remains unset and no payload is sent.
- P0: `--no-telemetry` still reaches the consent prompt.
- P0: the PR adds a first-run prompt and consent file even though the maintainer's stated contract for this issue is env-only opt-in via `SOUP_TELEMETRY=1`.
- P1: the PostHog key is still the placeholder `phc_soup_public_write_only`.
- P1: filesystem errors from the consent-file check can escape the intended failure isolation.
- P1: the required `changelog.d/0.73.3/529.*.md` fragment is missing.

Please reconcile the implementation with issue #318 and the maintainer direction, add regression coverage for `--no-telemetry`, consent boundaries, and failures, add the fragment, and rerun the full checks. The current green CI does not make the unsafe behavior mergeable.

MakazhanAlpamys (changes requested):
Re-review of 2d2d0e841396e830b8377cc7fe1e7f0ec38ff807: my focused telemetry run passed 147 tests, all 13 current checks are green, and the disabled-path mutation was caught. The branch is still not mergeable against issue #318: --no-telemetry still reaches the consent prompt, persisted consent does not enable sending in a new process, the PostHog key remains the placeholder, and the required changelog fragment is missing. The docs also contradict themselves about telemetry being deferred versus live. Please reconcile the implementation with the issue contract and documented env-only/default-off behavior, add the missing regression coverage and fragment, then refresh the full matrix.

MakazhanAlpamys:
@kok-o — re-reviewed from the code, not from my old comments. The timing explains most of this: your last commit `2d2d0e8` landed at `2026-08-25T08:28:36Z` and the design direction was posted at `08:36:30Z`, **eight minutes later**. Nothing has been pushed since, and `kok-o/Soup#1` is still `state: open, merged: false`. So the redesign was never seen, let alone declined — I am not treating this as ignored feedback.

Current state against what was agreed:

| ask | status | evidence |
|---|---|---|
| drop the `rich.prompt` consent flow | not done | `cli.py:830` `from rich.prompt import Confirm`; `:835` `Confirm.ask("Enable anonymous telemetry?", default=False)` |
| drop `~/.soup/telemetry_consent` | not done | written at `cli.py:840-841`, read at `:858-860` |
| `urllib` instead of `httpx` | not done | the `pyproject.toml` +1 line is `"httpx>=0.24.0"` added to **core** `dependencies` |
| changelog fragment | absent | no file matching `*529*` under `changelog.d/` |
| `distinct_id` allowlist | **fixed** | |
| five ruff errors | **fixed** | `ruff check` clean |
| `id_file.exists()` outside the `try` | **fixed** | now inside |

The `httpx` line deserves a note of its own: it goes into the **core** dependency list, not an extra. This project deliberately split torch and friends out into `[train]` to keep the base install light; telemetry should not be the thing that adds a new mandatory core dependency. `urllib` is in the stdlib and does everything a single POST needs.

## The blocker: the tests do not hold the guarantee up

Three mutations, verified against a positive control proving my probe detects a real send:

```
is_telemetry_enabled() -> True (on by default)         2 failed    KILLED
cli.py:849  if _telemetry_disabled: -> if False:       5 failed*   SURVIVED
send_telemetry_payload: opt-in gate removed            375 passed  SURVIVED
```

*those 5 failures reproduce identically on the clean head — pre-existing local environment issues, not the mutation.

The second one means **`--no-telemetry` can be made a no-op and the suite stays green**. I confirmed the behaviour rather than trusting the count: with the mutation, `--no-telemetry` together with `SOUP_TELEMETRY=1` produced a real POST to PostHog. The third removes the opt-in gate at the network boundary itself — the one place that decides whether anything leaves the machine — and `375 passed, 0 failed`.

The cause is mechanical: **8 of the 12 tests in `tests/test_v07141_telemetry_prompt.py` contain no `assert` at all** (checked by AST, not by eye). Every test covering the *non-sending* direction — `test_emit_telemetry_disabled`, `test_emit_telemetry_no_consent`, `test_emit_telemetry_consent_0`, `test_emit_telemetry_exception` — calls the function and asserts nothing. A test that only proves "this did not raise" cannot distinguish a working opt-out from a deleted one, and for a telemetry feature the opt-out *is* the product.

What these need is a monkeypatched transport and an assertion on the **call count**: zero by default, zero under `--no-telemetry`, zero without consent, and — the control that stops the whole file being vacuous — exactly one when telemetry is genuinely enabled.

## New finding: the payload can carry a local path, which the PR's own docs forbid

`_emit_telemetry` sends `command` as the PostHog event name, and `command` is whatever `_split_command_args` returns — the first non-flag argv token, truncated to 64 characters and **never checked against the registered command list**. I ran your own function on realistic argv:

```
soup train -c soup.yaml                              -> event = 'train'
soup trian -c soup.yaml                              -> event = 'trian'
soup /home/alice/secret-merger-project/data.jsonl    -> event = '/home/alice/secret-merger-project/data.jsonl'
soup ./acme-corp-model                               -> event = './acme-corp-model'
soup --log-level debug C:/clients/pfizer/run.yaml    -> event = 'C:/clients/pfizer/run.yaml'
```

Your own new documentation says *"We **NEVER** collect... Dataset paths... Local paths"*, and the consent prompt tells the user *"It never collects paths."* Those are absolute promises, so the argument is not about how often a user mistypes an invocation — it is that the guarantee has no mechanism behind it. The fix is small: validate `command` against `app`'s registered command names and send a literal `(unknown)` otherwise. Then add the test that pins it, because that promise is the entire basis on which someone opts in.

## Three smaller things

1. **The persistent opt-in is still dead.** Consent file containing `1`, no env var set: **zero** network calls, because `send_telemetry_payload` re-checks the env only. Meanwhile it still writes a persistent tracking UUID to `~/.soup/telemetry_id`. So the current code stores an identifier and sends nothing — the worst of both.
2. **The docs say "runs asynchronously" and it does not.** There is no threading or asyncio anywhere in the path. Measured with an injected 1.2 s network delay: `SOUP_TELEMETRY=0` → 1.615 s, `SOUP_TELEMETRY=1` → 2.705 s. It blocks command exit. Either make it async or change the sentence.
3. **`_resolve_posthog_target`'s SSRF guard does not do what its docstring says** — `https://10.0.0.1/` and `https://127.0.0.1/` are accepted, only plain `http://` is refused, while the docstring names `10.0.0.1` as the blocked case. **This is pre-existing on `main` and not your fault**; I checked. But your PR is what first makes it reachable, since telemetry was previously wired to nothing. I will file it separately rather than pile it onto this PR.

## What is genuinely right

Telemetry really is **off by default** — I measured zero network calls on `soup version` and `soup --help`, with `SOUP_TELEMETRY=0`, and with `--no-telemetry`. Import-lightness holds (`test_cli_startup_is_light` 34 passed) and there is no startup regression (main 1.38-2.13 s against 1.65-1.73 s here). The payload itself is well chosen: `soup_version`, `python`, `os`, `arch`, `duration_seconds`, `distinct_id` — nothing there I would object to.

Merging `kok-o/Soup#1` and rebasing is the fastest start, but note it predates the two findings above, so the path validation and the assert-bearing opt-out tests are needed on top either way.

MakazhanAlpamys:
Filed the pre-existing SSRF finding separately as #593 so it does not sit on your PR. It is not caused by this diff — it landed in v0.53.10 (#154) — but this PR is what makes the path live, so the two are worth tracking together.

MakazhanAlpamys:
@kok-o — checking in on this one, since you have been active on #528 but this has not moved in five days.

Not chasing you: you have two PRs open and focusing on one at a time is a reasonable thing to do. **If you are parking this until #528 lands, say so and I will stop looking at it until you come back** — that is a perfectly good answer and it saves us both re-reading.

If you are still on it, the blockers from my 2026-08-29 review are unchanged, and they are contract issues rather than code-quality ones:

1. **`--no-telemetry` still reaches the consent prompt.** A flag whose entire purpose is to opt out should not be asked for consent.
2. **Persisted consent does not enable sending in a new process** — so the consent is recorded and then does not do anything, which is the worst of both.
3. **The PostHog key is still the placeholder**, so nothing could send even if the above worked.
4. **No changelog fragment**, which #318 lists as required.
5. **The docs contradict themselves** on whether telemetry is deferred or live.

What I would fix first, if it helps prioritise: items 1 and 2 are the ones that make the feature wrong rather than incomplete. A user who passes `--no-telemetry` and still gets prompted will reasonably conclude the flag does not work, and that impression is expensive to undo. Items 3-5 are mechanical.

Two things worth knowing that are not your fault:

- **#593** — the pre-existing SSRF finding I split out so it would not sit on your PR. Still separate, still not blocking you; this PR is what makes that path live, which is why I keep them linked.
- **`pytorch-smoke` is now a required check** and post-dates your last run, so this will read as blocked at 13/13 green. A rebase picks it up.

Whatever the answer, tell me which it is — parked, active, or dropped — and I will act accordingly rather than guess.

kok-o:
@MakazhanAlpamys — updated and rebased onto current `main` (`e6ad7b8`). 

Here is what was addressed end-to-end:

1. **Env-only opt-in & zero prompts:** Completely removed `rich.prompt.Confirm` and `~/.soup/telemetry_consent`. `SOUP_TELEMETRY=1` is the single switch; unset or `0` means completely off and silent. Added `--no-telemetry` CLI flag that guarantees zero network calls even if `SOUP_TELEMETRY=1` is present in the environment.
2. **Command token sanitization (Path leak fix):** Added `_get_registered_commands()` in `cli.py` validating the extracted command against registered Typer/Click subcommands. File paths (e.g. `soup /home/alice/data.jsonl`), model directories, and typos are strictly masked as `(unknown)` so no path or private token ever leaves the machine.
3. **Stdlib `urllib.request` delivery:** No `httpx` core dependency — telemetry delivers JSON via standard library `urllib.request` with 1.0s timeout and fail-soft exception swallowing. Core CLI remains lightweight (<0.5s startup verified by `test_cli_startup_is_light.py`).
4. **Anonymous `distinct_id`:** Random `uuid.uuid4()` saved to `~/.soup/telemetry_id`, format-validated on read, with fail-soft filesystem exception handling. Added to payload schema allowlist in `test_v0430_part_a.py`.
5. **Load-bearing assert-driven tests:** Rewrote `tests/test_v07141_telemetry_prompt.py` (24 tests, 0 failures). Tests assert call counts at the network boundary: 0 by default, 0 under `--no-telemetry`, 0 with falsy env, and 1 when enabled. Verified all 3 mutations are caught. Verified by AST inspection that every single test contains `assert` statements.
6. **Docs & Privacy Policy:** Updated `docs/backends-and-ops.md` with the public Privacy Policy, removed contradictory "not yet wired" phrasing, accurately documented synchronous 1s execution on exit, and fixed line 223 (removed leftover RunPod mention).
7. **Changelog fragment:** Added `changelog.d/0.73.3/529.added.md`.

MakazhanAlpamys:
Thanks for coming back to this, and for answering the question I actually asked rather than only the code items — env-only opt-in with no prompt at all is a stronger answer than fixing `--no-telemetry` to skip the prompt.

Two things from your summary I want to acknowledge before reviewing properly, because they are the ones that were blocking:

- **`SOUP_TELEMETRY=1` as the single switch, with `--no-telemetry` guaranteeing zero network calls even when the variable is set.** That removes the contradiction where a flag whose whole purpose is opting out still reached a consent prompt.
- **Command-token sanitisation against the registered Typer/Click command list.** Validating against what is actually registered, rather than filtering what looks like a path, is the right shape — a deny-list would have leaked the next unanticipated argument form.

I am part-way through cutting **v0.74.0** and will not start a review that I cannot finish, so this gets a full pass — diff against the described behaviour, tests run here, and a mutation of the opt-out path — once the tag is out. Nothing is waiting on you; if you push again before then, that is fine and I will review the newest head.

MakazhanAlpamys:
@kok-o — reviewed by execution, and this is the highest-risk PR in the queue so I went at it accordingly: SSRF spellings, real payload contents, timing with the network blackholed, exception semantics, and mutation of every privacy-relevant branch.

**Start with the headline, because it is the opposite of what I expected: the implementation is materially safer than your own PR description says it is.** Two of the four things the body claims are not in the diff, and in both cases their absence is an improvement. That is also, separately, the biggest problem with the PR.

## What actually gets sent, printed from your code

```json
{"soup_version":"0.74.0","command":"train","python":"3.10","os":"Windows",
 "arch":"AMD64","duration_seconds":1.0,"distinct_id":"66960d10-...-cc4c95b777ab"}
```

No paths, no model names, no config contents, no hostname, no username, no argv, no error text. #318's honesty constraint is met exactly. The command name is sanitised against the registered-command set, so hostile argv cannot leak through it:

```
soup /home/alice/secret/my-dataset.jsonl        -> command='(unknown)'
soup train --config /home/alice/private.yaml    -> command='train'   (flags never read)
```

**Default OFF, verified on a redirected HOME with `CI=true` and no `SOUP_TELEMETRY`:** `is_telemetry_enabled()` returns `False`, zero sends, and `~/.soup` is not even created. That is the correct posture and it is stronger than #318 asked for.

**All four historical IPv4 bypass spellings rejected** at both the endpoint validator and the full resolve path — `127.1`, `2130706433`, `0x7f000001`, `0177.0.0.1` — along with loopback, link-local, `169.254.169.254`, plain `http://` on the default host, and the suffix-confusion case `https://attacker.posthog.com.evil.com/`. You reused the project's shared validators rather than hand-rolling a fourth copy, which is exactly right given #600, #599 and #616 were all the same class of bug living in different copies.

**No startup cost on the default path.** Six runs each, first discarded, median: `soup --help` 1.355s on main against **1.114s** on your branch; `soup version` 0.974s against **0.876s**. The difference is noise in your favour, because `is_telemetry_enabled()` returns before the sender is imported.

**Exception semantics measured, not read:** `SystemExit(3)`, `SystemExit(0)` and `KeyboardInterrupt` all propagate with exit codes preserved, 130 included. The `finally` does not swallow them.

**Mutations:**

| mutation | result |
|---|---|
| consent check always returns `True` | **KILLED** — 15 tests |
| `--no-telemetry` made a no-op | **KILLED** — `test_emit_telemetry_no_telemetry_flag_overrides_env` |
| network `except` re-raises | **KILLED** — `test_network_exception_swallowed` |
| **add `"cwd": os.getcwd()` to the payload** | **KILLED** — `test_no_user_data_in_payload` |
| endpoint host changed to `evil-collector.example` | survived your tests; killed by the pre-existing `test_issue599` suite |

The fourth row is the one I most needed to see. The payload contents are pinned, so a later change cannot quietly widen what is collected.

The 17 deleted test lines are one obsolete test, `test_httpx_missing_returns_false`, correctly removed. The two ported tests assert **more** than the versions they replace.

---

## Blocking 1 — your description misdescribes the implementation, in the privacy-critical direction

Four claims in the body, all false against the diff:

| body says | reality |
|---|---|
| "Moves `httpx` into core dependencies" | `pyproject.toml` is **not in the diff**. You removed httpx from the send path and used `urllib.request` |
| "Adds first-run consent via `rich.prompt` in `cli.py`" | No prompt code exists. You add `test_cli_has_no_prompt_and_documents_no_telemetry`, which asserts its **absence** |
| "writes `~/.soup/telemetry_consent`" | `grep -rn "telemetry_consent" src/` finds nothing |
| "sets `SOUP_TELEMETRY`" | The code only ever reads it |

I want to be precise about the direction here: **the code is better than the body.** Dropping httpx solves #318's item 3 — a plain `pip install` silently no-opping the sender — with **zero** new core dependencies, which is the right call for a deliberately light core and better than what the issue asked for. And no prompt, with default-OFF, is defensible and probably correct; an interactive prompt would be hostile in CI and in scripts.

But a body saying consent is captured interactively and stored on disk, when it is env-var-only, is the kind of error nobody catches afterwards. It becomes the squash-merge commit message on a telemetry feature. Please rewrite it to match the code, and state plainly that you chose no-prompt plus default-OFF over #318's wording, so the divergence is a recorded decision rather than a discrepancy someone finds in a year.

## Blocking 2 — the PostHog key is still the placeholder

`trackers.py:206`: `_POSTHOG_DEFAULT_KEY = "phc_soup_public_write_only"`.

#318 lists a real key as acceptance item 2, and it matters more than a TODO usually would: as shipped, a user who opts in sends real data over the network to an endpoint that will reject it. That is egress with no ingestion, which is the worst possible trade in a privacy feature — the user pays the full trust cost for zero benefit.

**This one is on me, not you.** Provisioning the key is my action and I have not done it. Two ways forward, and I would rather you pick:

- I get the key and you drop it in; or
- better, and useful regardless: make the sender **refuse to send while the key is the placeholder**, with a one-line notice saying telemetry is not yet live. Then the feature is honest in every state, merging stops being gated on my account admin, and flipping the key later is a one-line change with no behavioural surprise.

I lean strongly toward the second. This project shipped four config flags in v0.74.0 that were validated, documented, and read by nothing; a telemetry sender that transmits into a void is the same shape, and a guard turns it into a stated state instead of a silent one.

## Blocking 3 — my ruling on the privacy-policy blocker

#318 says *"A published public privacy policy — the stated hard blocker. Lands first, or nothing else ships."* You added a `### Privacy Policy` subsection to `docs/backends-and-ops.md`.

**Ruling: that satisfies it, and I am saying so explicitly so the record shows the blocker was cleared deliberately rather than forgotten.** Soup has no separate website; `docs/` is the published surface, it renders publicly, and it is versioned alongside the code, which is better provenance than a standalone page. What it has to contain is what a user needs in order to decide: the exact field list, the destination, how to opt in, how to opt out, and how to delete the identifier. Four of those five are present — see finding 6.

## Should-fix 4 — building a payload writes a persistent id to disk, even when telemetry is off

`trackers.py:190`. `build_telemetry_payload` calls `get_or_create_distinct_id()` unconditionally, so merely constructing a payload creates a permanent identifier in the user's home directory. Reproduced here on a redirected HOME with telemetry disabled:

```
PAYLOAD: {"arch": "AMD64", "command": "train", ...}
telemetry_id written by build_telemetry_payload alone: True
SIDE EFFECT while telemetry is DISABLED: True
```

Two consequences. A user who never opts in still gets a tracking identifier written into `~/.soup/`, which is the single thing a privacy-conscious reader will object to, and it is not documented anywhere. And running the test suite creates one in every developer's and every CI runner's home directory — I hit that here myself.

Fix: pass `distinct_id` in from the caller, since `_emit_telemetry` already gates on consent, or resolve it lazily inside `send_telemetry_payload`. Either way the identifier should come into existence only once somebody has opted in.

## Should-fix 5 — the documented "strict 1-second timeout" is not accurate

Measured with the network blackholed:

| | median | max |
|---|---|---|
| default host, NXDOMAIN | 0.959s | 1.218s |
| **custom endpoint, TCP blackholed** | **3.363s** | **5.411s** |

The tier-3 DNS probe and the `urlopen` connect timeout stack, and `urlopen(timeout=)` does not bound DNS resolution at all — so on a network with a black-holed resolver even the default endpoint is unbounded. Either document it as a one-second connect and read timeout with DNS excluded, or wrap the emission in a hard deadline. A CLI that hangs three seconds behind a corporate firewall is how telemetry earns its reputation.

## Should-fix 6 — no instructions for deleting the identifier

The docs say a UUID is stored at `~/.soup/telemetry_id` and never say how to remove it, which is the first thing a privacy-conscious user looks for. One line: deleting `~/.soup/telemetry_id` regenerates it on the next opt-in. This is the missing fifth element of the policy I ruled on above.

## Should-fix 7 — `SECURITY.md` and `README.md` say nothing about telemetry

Both grep clean. A CLI that gains network egress should say so where people look for its security posture, not only in an ops page. A short `SECURITY.md` subsection and one line in the README.

## Nits

`_emit_telemetry`'s docstring says "Never raises", but its first three lines sit outside the `try` — I confirmed it does raise when `is_telemetry_enabled` throws. The outer `finally` has its own `try/except` and catches it, so nothing crashes; the defence-in-depth is what saves it, not the function. Move those lines inside, or soften the docstring.

Telemetry fires on `--help` and `version`, and also in spawned children — ten sites run `python -m soup_cli.cli train ...` including `quickstart.py:219`, `iterative_dpo.py:419`, `cans/run.py:230`, `ui/app.py:307` and the loop daemon, so a single user action emits two or more events. Not a privacy problem; it will skew your counts before you have read one of them. An internal env marker on spawned children fixes it.

## Mechanical

`changelog.d/0.73.3/529.added.md` needs to become `changelog.d/0.74.0/529.added.md`. It passes in your worktree only because that tree's `CHANGELOG.md` predates the release. After a clean rebase, which succeeds with no conflicts, the guard fails with `fragment baseline is 0.73.3, but the latest release is 0.74.0`. The fragment's text itself is right — it names `#529` and credits you.

---

To summarise what I want back: a corrected PR body, the placeholder-key guard, the `distinct_id` side effect moved behind consent, the timeout wording, the deletion instruction, and the fragment move. The privacy-policy question is answered above and needs nothing from you. The design itself I am not asking you to change — default-OFF with no prompt is the right shape, and I would like it on the record that you chose it deliberately.

kok-o:
@MakazhanAlpamys — rebased onto `upstream/main` (`v0.74.0`) and addressed all blocking and should-fix findings:

1. **Rewrote PR body**: Corrected the four inaccuracies in the PR description to match the diff, and recorded the deliberate choice of default-OFF + env-only opt-in over interactive prompt consent.
2. **Placeholder-key guard**: Added `is_placeholder_posthog_key()`. `send_telemetry_payload` now immediately refuses transmission when the key is `phc_soup_public_write_only`, emits `Notice: telemetry is not yet live.\n` to `stderr`, and returns `False` (0 network calls). Verified that supplying a custom key via `SOUP_POSTHOG_KEY` or `api_key` transmits cleanly.
3. **Distinct ID disk side-effect eliminated**: `build_telemetry_payload()` now returns an in-memory ephemeral UUID without writing to `~/.soup/telemetry_id` when telemetry is disabled. Persisted distinct ID creation is gated strictly behind consent in `_emit_telemetry()`. Added test verifying zero disk trace on a redirected home directory when disabled.
4. **Timeout wording precision**: Documented in `docs/backends-and-ops.md` and docstrings that the 1-second timeout is a socket connect and read timeout with DNS resolution excluded.
5. **Identifier deletion instructions**: Documented in `docs/backends-and-ops.md` (both in the Telemetry section and Privacy Policy) that deleting `~/.soup/telemetry_id` erases the identifier, which regenerates on the next opt-in run.
6. **SECURITY.md & README.md notices**: Added a `### Telemetry and Network Egress` section to `SECURITY.md` and a single line to `README.md` (README line count stays under the budget at 480 lines).
7. **Changelog fragment moved**: Moved to `changelog.d/0.74.0/529.added.md` and verified clean baseline with `scripts/assemble_changelog.py`.
8. **_emit_telemetry defense-in-depth**: Wrapped the entire function body in `try...except` so the docstring's "Never raises" invariant holds unconditionally even if environment inspection raises.

Test suite passing with 160/160 tests (`tests/test_v07141_telemetry_prompt.py`, `tests/test_v0538.py`, `tests/test_v0430_part_a.py`), all load-bearing AST assertions intact, and startup lightness verified (`test_cli_startup_is_light.py`).

MakazhanAlpamys:
Merged as `5f43a04`, credited in `CONTRIBUTORS.md` (`2edc063`). Third merge of yours today, and the one I was most prepared to block on.

## Re-verified by execution, not from your summary

| check | result |
|---|---|
| placeholder-key guard, telemetry ON, `urlopen` booby-trapped | `send` returns **False**, **0 network calls**, `Notice: telemetry is not yet live.` on stderr |
| same, through the real CLI (`soup version`) | exit 0, notice on stderr, no transmission |
| disk side effect, telemetry OFF, redirected HOME | `telemetry_id` **not** written; payload still carries an ephemeral id |
| rebase onto current `main` | clean; README lands at 501 with **both** your line and my Troubleshooting displacement intact |
| four IPv4 bypass spellings | all rejected; `http://` on the default host rejected; `attacker.posthog.com.evil.com` rejected; the legitimate endpoint accepted |
| SSRF suite | 51 passed |
| `test_cli_startup_is_light` | 34 passed |
| merged against current `main` | **177 passed** |

Mutations, on top of the four from the first round:

| mutation | result |
|---|---|
| `is_placeholder_posthog_key` -> always `False` (would transmit) | **KILLED** — `test_send_telemetry_refuses_placeholder_key_with_notice` |
| `build_telemetry_payload` writes the persistent id again | **KILLED** — `test_build_telemetry_payload_disabled_leaves_no_disk_trace` |

Both new guards are pinned by a test that fails when you remove them, which is the only thing that stops them being decoration.

## What was good

**You picked the harder of the two options I offered on the key, and it was the right one.** Waiting for me to provision a PostHog account would have parked this indefinitely on my admin. Refusing to transmit while the key is a placeholder means the feature is honest in every state it can be in, the merge does not depend on me, and flipping the key later is one line with no behavioural surprise. That is a better answer than the question.

**And the `distinct_id` side effect is the finding I would least have wanted to ship.** A tracking identifier written into the home directory of a user who never opted in is precisely the thing that costs a project trust, and it was also landing in every developer's and CI runner's home directory that ran the suite. You moved it behind consent rather than papering over it.

**Something worth stating plainly, because it is unusual:** two of the four things your original description claimed were not in the diff — httpx into core, and a `rich` consent prompt writing `~/.soup/telemetry_consent` — and in **both** cases their absence was an improvement. Stdlib `urllib.request` solves #318's "a plain `pip install` silently no-ops the sender" with zero new core dependencies, which is better than the issue asked for; and no prompt, with default-OFF, is correct because a prompt is hostile in CI. You fixed the description rather than the code, and recorded the divergence from #318 as a deliberate choice. That is the right way round.

## On the record for whoever reads this later

The privacy policy question was mine to rule on and I ruled that `docs/backends-and-ops.md` satisfies #318's hard blocker: Soup has no separate website, `docs/` is the published surface, and it is versioned with the code, which is better provenance than a standalone page. It is in the merge commit so the decision is traceable rather than implicit.

The PostHog key is still the placeholder and that is now **my** outstanding item, not yours. Until I provision it, the guard makes the state visible to anyone who opts in.

## One thing I did not ask you to change

`--no-telemetry` is matched as a bare substring of `argv`, so it would also trigger if that exact string arrived as a flag *value*. It fails safe — disabling telemetry — so it is not a bug. Noting it only so a future reader does not mistake it for one.

PR #535, opened 2026-08-25
fix(sft): reject rows without causal targets

### LINKED ISSUE #532: Responses-only SFT can save an all-NaN adapter after target truncation
## Reproducer

I hit this while running a bounded Apple Silicon smoke train with:

- `Qwen/Qwen3.8-27B`
- text SFT, `train_on_responses_only: true`
- BF16 layer streaming
- LoRA r64 / alpha128
- batch 1, two Alpaca rows, one epoch
- `max_length: 64`
- PyTorch 2.12.1, Transformers 5.12.1, TRL 0.29.1, PEFT 0.20.0, MPS

The checkpoint's chat template inserts a system reasoning block. The two assistant spans begin at token positions 63 and 75. After truncation to 64, the first row has one supervised token and the second has zero.

Soup still runs both steps and exits 0 with `Training Complete`:

```
loss: 0
grad_norm: nan
entropy: nan
```

The saved adapter is not merely unchanged: all 105,906,176 LoRA values are NaN.

Raising `max_length` to 128 makes the same run healthy: 12 and 11 supervised labels, loss 2.2839 -> 1.8100, finite gradient norms, and every saved adapter value finite.

## Why this matters

A dataset can legitimately contain a long prompt whose assistant response is entirely truncated. One such row can poison the optimizer, but the run is reported as successful and publishes a dead adapter.

## Acceptance criteria

- Reject a post-tokenization training row when no causal-loss target remains after truncation.
- Name the row and `data.max_length` in the error.
- Add a control with one surviving assistant target and a failing zero-target row.
- Do not emit `Training Complete` or save a final adapter after a non-finite training state.
- Verify the guard fails if the zero-target check is removed.

I am taking this issue and will prepare a focused fix from current `main`.

Amix29:
## Summary

- reject text SFT rows when tokenization/truncation leaves no shifted causal-LM target in `labels[1:]`
- name the failing split, human-facing row number, and `data.max_length`
- apply the same invariant to pre-tokenized datasets without changing low-level mask-preview behavior
- refuse the final model save when logged training metrics or trainable floating-point parameters are non-finite

## Reproduction and validation

The original two-row `Qwen/Qwen3.8-27B` reproducer was rerun against the locally cached tokenizer on Apple Silicon:

- `max_length=64`: refused before Trainer construction at `train row 2`, as expected
- `max_length=128`: accepted with 12 and 11 shifted causal-loss targets

The regression suite includes the boundary control with one surviving assistant token, the fully truncated failing row, the causal-shift position-zero edge case, pre-tokenized input, non-finite metrics, non-finite trainable parameters, and the pre-save ordering invariant. Removing the zero-target check makes `test_fully_truncated_target_names_row_and_max_length` fail.

Validation:

- `460 passed, 21 skipped` across assistant-mask, SFT, streaming, and adapter-save tests
- full suite: `18,794 passed, 82 skipped, 5 deselected`
- coverage: `82.74%`
- `git diff --check`

Fixes #532

### DIFF
--- a/changelog.d/0.73.3/535.fixed.md
+++ b/changelog.d/0.73.3/535.fixed.md
@@ -1,0 +1,2 @@
+- SFT now refuses examples without causal loss targets and final training states
+  containing non-finite metrics or parameters before saving artifacts (#535).
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -560,6 +560,11 @@ data:
 
 When the tokenizer ships a chat template with `{% generation %}` markers, the mask is exact. Without those markers, Soup falls back to an incremental tokenize-delta walk and documents the looseness.
 
+After tokenization and truncation, every response-only row must retain at least one shifted
+causal-loss target. Soup rejects the row by split and row number when
+`data.max_length` truncates the complete assistant response, rather than training on an
+all-masked sequence or saving a non-finite adapter.
+
 ### `--trust-remote-code` opt-in (every command, every trainer)
 
 Every command that loads a model now requires `--trust-remote-code` to execute custom Python from a model repo (`auto_map` in `config.json`). First-party orgs (Meta, Mistral, Qwen, Google, etc.) suppress the warning panel; everything else prints a `REMOTE CODE WARNING` panel before loading. Unknown-org local checkpoints with `auto_map` raise a friendly `ValueError` at construction time instead of silently exec'ing inside `from_pretrained`.
--- a/src/soup_cli/data/loss_mask.py
+++ b/src/soup_cli/data/loss_mask.py
@@ -38,6 +38,10 @@ IGNORE_INDEX = -100
 _MISSING = object()
 
 
+class NoCausalLossTargetError(ValueError):
+    """Raised when truncation leaves a row with no shifted causal-LM target."""
+
+
 def _coerce_int_list(
     values: Any, *, field: str, allow_bool: bool = False
 ) -> list[int]:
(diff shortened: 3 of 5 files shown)


MakazhanAlpamys:
Hi @Amix29 - full mutation review done on both halves of this fix, and the engineering is exactly right. Before the detail: one small item missing - there is no changelog fragment. Please add `changelog.d/0.73.3/535.fixed.md` starting with a `- ` list item and mentioning #535 in the text (the repository-wide test enforces the format; every sibling PR in this release carries one). That is the only thing between this and a merge.

What I verified:
- The refusal happens in setup() BEFORE Trainer construction on BOTH paths (text SFT and pre-tokenized) - so the all-NaN adapter scenario from your own #532 dies before minutes of training, not after.
- `_assert_finite_training_state` runs before save_model AND tokenizer.save_pretrained, so the artifact on disk cannot be poisoned by a non-finite state; intermediate checkpoints deliberately ungated, matching the issue's acceptance criteria.
- RAFT branches untouched, vision/audio/embedding/asr out of the diff, --show-mask behavior unchanged as claimed.
- 170 targeted tests green, ruff clean, CI 13/13.

Mutations (all killed):
- Neutralized `ensure_causal_loss_target`: 3 named tests fell (fully-truncated naming, position-zero edge, pre-tokenized path).
- Made the finite-state check a no-op: 2 fell (metric + parameter).
- Removed the gate call from train(): the ordering test fell (`test_training_state_is_checked_before_final_save`) - the save-ordering invariant has real coverage, not just proximity.

Filing your own #532 and then closing it same-day with a reproducer-driven fix is the loop working as intended. Add the fragment and I merge immediately.

Amix29:
Added the requested `changelog.d/0.73.3/535.fixed.md` fragment in `97d2ba7`, starting with a list item and referencing #535.

Validation after the update:
- changelog fragment validation: passed
- issue #532 plus changelog regression tests: 25 passed

A fresh CI run is attached to the updated commit.

MakazhanAlpamys:
Merged as aefc2cb. Thank you - this is the loop working exactly as it should: you found #532 while validating #526 on real hardware, filed it with the reproducer, and closed it same-day.

What stood out in the implementation: both refusal layers are ordered correctly, not merely present - the row gate fires in setup() BEFORE Trainer construction on both the text and pre-tokenized paths (so the failure costs seconds, not minutes), and `_assert_finite_training_state` runs before save_model AND tokenizer.save_pretrained, so nothing non-finite can reach disk. Leaving intermediate checkpoints ungated was also the right scope call, matching the issue's own acceptance criteria rather than gold-plating.

Runs before merge: 170 targeted tests green, ruff clean, CI 13/13. Mutations, all killed by named tests: neutralized `ensure_causal_loss_target` -> 3 fell; no-op'd the finite-state gate -> 2 fell; removed the gate call from train() -> `test_training_state_is_checked_before_final_save` fell, which is the one proving the ordering is load-bearing rather than incidental.

CONTRIBUTORS.md updated with the entry covering #537 and this.

PR #537, opened 2026-08-25
fix(stream): preserve base model provenance

### LINKED ISSUE #531: Layer streaming saves adapters with an empty base_model_name_or_path
## Reproducer

A successful `Qwen/Qwen3.8-27B` text SFT run using `training.stream_layers: true` produced a healthy LoRA adapter:

- r64 / alpha128
- 105,906,176 trainable values
- all 256 saved adapter tensors finite
- explicit-base PEFT reload succeeds
- one-token MPS forward after reload is finite

However, the emitted `adapter_config.json` contains:

```json
"base_model_name_or_path": ""
```

That breaks Soup's normal adapter contract:

```
soup chat --model <adapter>
Cannot detect base model from adapter_config.json.
Please specify with --base flag.
```

The same empty field affects the auto-detection paths used by merge, export, serve, infer, eval, push metadata, adapter arithmetic, and related tooling. Passing `--base Qwen/Qwen3.8-27B` proves that the weights themselves are loadable; the failure is the saved provenance field.

## Likely boundary

The streaming path constructs a meta skeleton from config before PEFT wrapping. The resulting model does not carry the source repo ID that PEFT normally copies into `base_model_name_or_path`.

## Acceptance criteria

- A streamed adapter saves the exact configured base reference in `adapter_config.json`.
- `soup chat --model <adapter>` reaches model loading without an explicit `--base`.
- A regression test checks the saved config rather than only checking that an adapter file exists.
- A control proves that blanking the field fails the test.
- Existing explicit-`--base` behavior remains unchanged.

I am taking this issue and will keep the fix scoped to streamed adapter provenance.

Amix29:
## What does this PR do?

Layer streaming builds the base model with `from_config`, which can leave both model identity fields blank. PEFT then copies that blank identity into `adapter_config.json`, so otherwise healthy streamed adapters cannot be auto-detected by `soup chat`, merge, serve, export, or other adapter consumers.

This PR stamps the exact configured base reference onto the meta skeleton before PEFT wraps it. The saved `base_model_name_or_path` therefore remains `Qwen/Qwen3.8-27B` (or the exact configured local/base reference), matching the normal `from_pretrained` contract.

Regression coverage verifies:

- the exact configured reference reaches the saved PEFT config;
- chat auto-detection reads the saved reference;
- blank provenance still fails auto-detection as the negative control;
- an explicit `--base` continues to bypass auto-detection unchanged.

The troubleshooting guide now documents the symptom and the explicit-base workaround for already-produced adapters.

Fixes #531.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Validation

- Cached `Qwen/Qwen3.8-27B` config reproduction: both meta-model identity fields are now exact before PEFT wrapping (no model weights loaded)
- `pytest --no-cov -q tests/test_issue531_stream_base_provenance.py tests/test_chat.py tests/test_v07201.py` — 24 passed
- `pytest tests/ -q` — 18,789 passed, 82 skipped, 5 deselected; 82.75% coverage

## Checklist

- [x] Added changelog fragment `changelog.d/0.73.3/537.fixed.md`

### DIFF
--- a/changelog.d/0.73.3/537.fixed.md
+++ b/changelog.d/0.73.3/537.fixed.md
@@ -1,0 +1,2 @@
+- Streamed LoRA adapters now preserve the exact configured base-model provenance,
+  restoring automatic base detection across adapter workflows (#537).
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -467,6 +467,11 @@ output: ./output
   state rather than on how many tensors Soup materialised: some PEFT versions
   create real adapters themselves. Include the PEFT version and the named
   parameter from the error when reporting this.
+- **A streamed adapter asks for `--base` when opened** — check
+  `adapter_config.json`. A healthy artifact records the exact configured model
+  reference in `base_model_name_or_path`; an empty value means the adapter was
+  produced by an older streaming path that lost the meta skeleton's origin.
+  Passing `--base` remains a valid workaround for that existing artifact.
 - **"layer streaming needs the base to fit in RAM"** — the base is larger than free RAM. Set `stream_source: auto` to fall back to the NVMe disk tier, free RAM, or pick a smaller base.
 - **"layer streaming needs NVMe or more RAM … the detected disk is 'hdd'"** on a fast cloud disk — a virtio device reports `rotational=1` with no media hint. Detection now measures the disk when the flag is unreliable; if it still misreads yours, set `training.stream_disk_kind: nvme` to force the tier on (`ssd`/`hdd` force it off).
 - **"could not page-lock the base … falling back to a PAGEABLE RAM store"** — expected on a busy machine. Training continues, more slowly. Close other applications to keep the pinned store.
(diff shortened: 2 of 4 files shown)


MakazhanAlpamys:
Merged as c081add. Thank you - and this one has a nice property worth naming: you found #531 yourself while validating #526 on real hardware, filed it precisely, and closed it same-day with a reproducer-driven fix.

What was good in the implementation: the stamp is unconditional and sits ahead of the only return from build_meta_skeleton, so there is no early path that can bypass it - that structural choice is why I could not build a bypass mutation. And your test design is the discriminating kind: my first mutation (remove the stamp entirely) died on `test_streamed_peft_config_saves_exact_configured_base`, but so did the subtler one - stamping a WRONG-but-present constant (`"unknown"`) dies on the same exact comparison while your negative control (blank provenance still fails detection) and the explicit `--base` bypass stay green in both. A weaker test would have passed the second mutation; yours distinguishes "present" from "correct".

Runs before merge: 43 targeted tests green (new file + chat/adapters/merge neighbors), ruff clean, CI 13/13 cells.

The troubleshooting entry for already-produced adapters (the --base workaround) was the right call - existing artifacts do not heal themselves.

CONTRIBUTORS.md updated. Still open on your desk when you return to them: #535 needs only its changelog fragment, and #526 carries the three blockers from this morning's review.

PR #538, opened 2026-08-25
fix(stream): materialize cached HF snapshots offline

### LINKED ISSUE #533: A fully cached layer-streaming model still needs network for local materialization
## Reproducer

The standard Hugging Face cache for `Qwen/Qwen3.8-27B` was complete and internally valid:

- `scan_cache_dir()` reported one 55.56 GB revision
- `refs/main` pointed to the existing snapshot commit
- the snapshot contained 32 valid files
- `snapshot_download(..., local_files_only=True)` resolved for `None`, `main`, and the commit SHA
- tokenizer and config both loaded offline

With `HF_HUB_OFFLINE=1` and `TRANSFORMERS_OFFLINE=1`, Soup found the source and printed the disk pre-flight, then failed before sharding:

```
LocalEntryNotFoundError: Cannot find an appropriate cached snapshot folder...
```

The traceback ends at:

```
resolve_model_weights
  -> materialize_model_weights
  -> soup_cli.utils.hubs.snapshot_download
  -> huggingface_hub.snapshot_download(local_dir=...)
```

The second call targets Soup's materialized directory and resolves `main` through `repo_info`, so offline mode rejects it even though the source snapshot and every weight blob are local.

With networking enabled, Hugging Face reports `0.00B` downloaded and materializes the same files in about 25 seconds.

## Acceptance criteria

- A complete standard HF symlink snapshot can be materialized and sharded with outgoing traffic disabled.
- The already-resolved commit is pinned through the materialization boundary.
- No remote metadata request is made in the offline control.
- Missing blobs still fail closed rather than producing a partial copy.
- Existing path-containment and symlink defenses remain intact.

I am taking this issue and will prepare a focused offline-cache regression fix.

Amix29:
## What does this PR do?

A complete standard Hugging Face cache could be resolved offline, but Soup then called `snapshot_download` a second time with a fresh `local_dir`. Hugging Face keeps the repository tree metadata for `local_dir` separately, so that second call cannot populate a new directory offline even when every source blob is already cached.

This PR carries the commit SHA from the resolved `snapshots/<commit>` directory into `ModelWeightsPlan` and materializes the already-validated snapshot directly. No second Hub metadata request is made.

The materialization boundary remains fail-closed:

- only `.safetensors` and `.json` files from the pinned snapshot are copied;
- snapshot symlinks must resolve inside that repository's HF blob store;
- files are opened with no-follow semantics where supported;
- missing blobs abort before a destination is published;
- the regular-file copy is assembled in a staging directory and atomically published;
- destination symlinks, junctions, reparse points, mounts, and cache escapes are refused;
- the saved HF metadata binds reused copies to both the commit and blob identity.

Regression coverage materializes a valid cached symlink snapshot with the Hub unavailable and then successfully shards the result. Negative controls cover a missing blob, an escaping source symlink, a redirected destination, and a missing resolved commit.

Fixes #533.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Validation

- Real cached `Qwen/Qwen3.8-27B` offline plan: 18 shards, 51.75 GiB, resolved commit pinned; no 51.75 GiB duplicate was created during validation
- `pytest --no-cov -q tests/test_issue533_offline_hf_materialization.py tests/test_issue374_stream_disk_preflight.py tests/test_v0718.py` — 160 passed
- `pytest tests/ -q` — 18,846 passed, 82 skipped, 5 deselected; 82.77% coverage

## Checklist

- [x] Added changelog fragment `changelog.d/0.73.3/538.fixed.md`

### DIFF
--- a/changelog.d/0.73.3/538.fixed.md
+++ b/changelog.d/0.73.3/538.fixed.md
@@ -1,0 +1,2 @@
+- Fully cached Hugging Face snapshots can now be materialized and sharded with
+  outgoing traffic disabled, while preserving commit and blob integrity (#538).
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -541,6 +541,13 @@ copy. The layer shards remain under `~/.soup/layer-stream/`. Their index records
 filename, size, and `mtime_ns`, so a necessary re-shard says which component changed instead
 of silently spending minutes rebuilding the cache.
 
+This materialisation also works with `HF_HUB_OFFLINE=1` when the standard Hugging Face
+snapshot is complete. Soup pins the commit resolved by the initial cache lookup and copies
+only verified snapshot files from that commit's blob store; it does not perform a second Hub
+metadata request for the regular-file directory. A missing blob or an escaping symlink aborts
+before the destination is published, rather than leaving a partial checkpoint that the sharder
+could consume.
+
 
 ## Correctness First (v0.36.0)
 
--- a/src/soup_cli/utils/spectrum_scan.py
+++ b/src/soup_cli/utils/spectrum_scan.py
@@ -31,7 +31,10 @@ import logging
 import math
 import os
 import re
+import shutil
+import stat
 import tempfile
+import time
 from dataclasses import dataclass
 from typing import (
     TYPE_CHECKING,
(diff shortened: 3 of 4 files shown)


MakazhanAlpamys:
Thank you, @Amix29. #538 was squash-merged. The offline path binds commit/blob identity, enforces containment and no-follow handling, and publishes only a completed staging directory. I exercised the symlink escape, destination escape, and offline-materialization controls; named tests caught them. All 13 CI jobs passed.

PR #539, opened 2026-08-25
fix(stream): detect Apple internal NVMe through APFS

### LINKED ISSUE #534: Apple internal NVMe is classified as ssd through the APFS Apple Fabric layer
## Reproducer

On an Apple Silicon Mac, `training.stream_source: auto` refused the disk tier with:

```
the detected disk is 'ssd' (not NVMe)
```

The underlying hardware is unambiguously NVMe:

- `system_profiler SPNVMeDataType` reports `NVMExpress`
- controller: Apple SSD Controller
- model: APPLE SSD AP1024Z
- internal, TRIM enabled, SMART verified

The mounted APFS data volume reports `Protocol: Apple Fabric`, which appears to be the value Soup classifies as a generic SSD. Setting the documented override:

```yaml
training:
  stream_disk_kind: nvme
```

allows the same run to proceed and complete successfully.

## Why this matters

Apple's internal NVMe commonly sits behind APFS/Apple Fabric rather than exposing `NVMe` as the mounted volume protocol. The automatic tier therefore refuses exactly the Apple Silicon hardware where the MPS streaming path is intended to be tested.

## Acceptance criteria

- Resolve an APFS volume to its physical store before classifying disk kind on macOS.
- Recognize an Apple SSD listed by `SPNVMeDataType` as NVMe.
- Preserve the existing explicit override as the final authority.
- Add an Apple Fabric/APFS fixture plus an ordinary SATA/SSD control.
- Keep unknown hardware conservative.

I am taking this issue and will prepare a focused macOS detection fix.

Amix29:
## What does this PR do?

Fixes #534.

On macOS, `diskutil info` reports an APFS data volume through the logical `Apple Fabric` layer. The previous detector inspected that text and therefore returned `ssd`, even when the APFS physical store was the internal Apple NVMe device.

This change:

- resolves an arbitrary shard directory to its mounted device when necessary;
- parses structured `diskutil info -plist` output;
- resolves `APFSPhysicalStores` to whole-disk BSD identifiers;
- admits the disk tier only when those exact devices are present in the privacy-minimized `SPNVMeDataType` inventory;
- keeps an unmatched solid-state device as `ssd` and unknown hardware as `unknown`;
- preserves `training.stream_disk_kind` as the final explicit authority.

The regression coverage includes Apple Fabric/APFS, an ordinary SATA SSD, an unmatched APFS store, malformed metadata, unknown media, and an arbitrary directory inside an APFS volume.

Hardware verification on Apple Silicon now reports `nvme` without an override.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Validation

- `PYTHONPATH=src pytest --no-cov -q tests/test_issue534_apple_nvme.py tests/test_v07203.py tests/test_doctor.py` — 161 passed, 15 skipped
- `PYTHONPATH=src pytest tests/ -q` — 18,848 passed, 82 skipped, 5 deselected; 82.78% coverage
- `ruff check src/soup_cli/ tests/` — passed
- Real Apple Silicon disk classification — `nvme`

## Checklist

### DIFF
--- a/changelog.d/0.73.3/539.fixed.md
+++ b/changelog.d/0.73.3/539.fixed.md
@@ -1,0 +1,2 @@
+- Apple internal NVMe volumes behind APFS and Apple Fabric are now detected for
+  the layer-streaming disk tier without requiring a manual override (#539).
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -301,6 +301,12 @@ Untied `embed_tokens` + `lm_head` stay resident and unquantised (2.10 GB of the 
 
 **Honest scope:**
 - **RAM tier + disk overflow (v0.72.3).** `stream_source: auto` picks RAM when it fits, falls back to NVMe disk when not; SATA/HDD rejected. Correctness verified; disk performance unmeasured on the reference box. A paravirtual (virtio) disk reports `rotational=1` with no media hint, so a genuinely NVMe-backed cloud disk was misread as an HDD and refused (#365); detection now measures a bounded O_DIRECT sequential read when the rotational flag is unreliable and admits NVMe-class throughput (>= 1 GB/s), while a genuinely slow disk stays rejected. Set `training.stream_disk_kind: nvme` (or `ssd`/`hdd`) to override when detection is still wrong — the resolved value is printed beside what was detected.
+- **Apple APFS disk detection.** On macOS, an APFS volume may report `Apple Fabric`
+  even when its physical store is Apple's internal NVMe. Soup resolves the target
+  volume to its APFS physical store and admits it only when that exact device is
+  listed by `SPNVMeDataType`; an unmatched solid-state device remains `ssd`, and
+  unknown hardware remains refused. `training.stream_disk_kind` still has final
+  authority when explicitly set.
 - **Llama / Qwen / Qwen3.5 dense and MoE text / Mistral / Gemma / Gemma2 / Gemma3-Text / Phi / Phi3** (`qwen3_5`, `qwen3_5_text`, `qwen3_5_moe`, and `qwen3_5_moe_text` route through the qwen3 streamer), `task: sft`, `backend: transformers`, `modality: text`. The original list is verified bit-exact in bf16 and NF4. Qwen3.5's heterogeneous dense and MoE decoder paths are verified bit-exact against resident controls on CPU; the MoE path also has live streamed-training validation on `Qwen/Qwen3.5-35B-A3B`, whose real 35B run has no resident control because the available hardware could not load it resident.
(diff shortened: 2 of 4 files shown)


MakazhanAlpamys:
Hi @Amix29 - reviewed the update. One blocker left, and it is a one-liner on your side:

**Blocker - Windows CI is red on a non-portable assert.** `tests/test_issue534_apple_nvme.py:54` compares a literal POSIX path (`/private/var/model-shards`) against the arguments handed to diskutil, but `_darwin_disk_kind` runs its input through `os.path.realpath`, which on Windows produces `C:\private\var\...` - so the assert fails on every Windows cell (reproduced locally and confirmed in the windows-3.10 job log). Compare the path tail, or normalize both sides before comparing. Everything else about the PR held up under review:

- The parser trio is genuinely pure over passed-in structures, platform calls isolated behind the `_probe_disk_kind` seam - testable here without OS-level mocks.
- The fail-safe direction is right: unmatched Apple Fabric stays `ssd`, unknown hardware/broken plist -> `unknown`, and unknown refuses the disk tier - an HDD cannot pass as NVMe without matching BSD ids against the SPNVMeDataType inventory.
- v0.72.3 BusType-beats-MediaType and the virtio fix (#365/#411) regressions all pass.
- Mutations: promoting any APFS store without the inventory check dies on `test_unmatched_apple_fabric_store_is_not_promoted_to_nvme`; always-ssd dies on `test_apple_fabric_apfs_store_is_matched_to_nvme_inventory`; killing the df-resolve dies on `test_directory_inside_apfs_volume_is_resolved_through_df`.

One optional improvement, not blocking: your plist fixtures are synthetic (plistlib.dumps over hand-written dicts). If you can capture one real `diskutil info -plist` and one real `system_profiler SPNVMeDataType -json` output from the Mac into test fixtures (redacting serials), the parsers would be pinned against actual upstream shapes rather than our assumptions about them.

Fix the line-54 portability and this merges.

Amix29:
Fixed the Windows-only path assertion in `9ab38f7`.

The test still pins the exact `diskutil info -plist` argv prefix, while normalizing path separators and comparing the expected path tail. This accepts both the macOS `/private/...` path and the Windows drive-prefixed path produced by `os.path.realpath`, without weakening the behavior under test.

Local validation:
- `PYTHONPATH=src pytest --no-cov -q tests/test_issue534_apple_nvme.py` — 6 passed
- `ruff check tests/test_issue534_apple_nvme.py` — passed
- `git diff --check` — passed
- explicit Windows-style `D:\private\var\model-shards` assertion simulation — passed

Fresh CI is now running on the pushed commit.

MakazhanAlpamys:
Thank you, @Amix29. #539 was squash-merged. NVMe is now identified through the APFS physical store and exact BSD-ID inventory matching, with unknown and unmatched hardware fail-closed. I mutated the inventory check, always-SSD path, and df resolution; each failed its dedicated regression. All 13 CI jobs passed.

Issue #472, opened 2026-08-21
Question: Qwen 3.8 27B on Mac with Lora


Sir-Lysander:
Hello, I have a MacBook Pro M4 Max with 128GB of unified RAM. I was wondering if the Qwen 3.8 27B model is currently supported on my Mac for distillation using Lora?

MakazhanAlpamys:
Hello — thanks for asking before burning a run on it. Short answer: **LoRA SFT yes, distillation no**, and the "no" is a real limitation rather than a missing model.

I checked the code rather than answering from memory, so here is exactly what holds today.

## 1. The model name

There is no **Qwen 3.8** in Soup's catalog, and as far as I know Alibaba has not released one. What exists here is:

```
Qwen/Qwen3.5-27B      -> recipe  qwen3.5-27b-sft
Qwen/Qwen3.6-27B      -> recipe  qwen3.6-27b-sft
```

If you meant one of those, read on. If 3.8 is something you have seen released, tell me and I will look — the catalog is hand-maintained and it would not be the first time we were behind.

## 2. Distillation on Apple Silicon is not supported, and it fails loudly

The MLX backend registers exactly three tasks — `sft`, `dpo`, `grpo`. `distill` is not one of them, and the config is refused at parse time rather than failing minutes in:

```
$ soup train -c soup.yaml     # task: distill, backend: mlx
REFUSED: task='distill' is not supported on backend=mlx
```

At least you get that in the first second and not after loading 27B of weights.

Two workarounds, both with honest caveats:

- **`backend: transformers` on MPS.** Nothing blocks it, and 128 GB unified is genuinely a lot — but distillation holds a **student and a frozen teacher simultaneously**, so a 27B student plus any serious teacher is a very different memory story from SFT. I have no Apple Silicon here (this box is a 4 GB RTX 3050), so I am not going to tell you it works. If you try it, I would genuinely like to know what happens.
- **Two-step instead**: generate from the teacher with `soup data best-of-n` or `soup data gen-magpie`, then plain LoRA SFT on the result via MLX. Not the same thing as logit-level KD, but it runs on the supported path.

## 3. What does work on your machine

LoRA SFT via MLX. The 27B recipe is written for a CUDA/DeepSpeed box, so switch the backend and drop the bitsandbytes quantization, which is CUDA-only:

```
soup recipes show qwen3.5-27b-sft
```

[...]

PR #543, opened 2026-08-25
fix(data): support system-first Qwen3.8 masking

### LINKED ISSUE #540: Qwen3.8 native chat template crashes response-only masking on system-first conversations
## Environment

- Soup 0.73.3
- `Qwen/Qwen3.8-27B`
- Transformers 5.12.1 / PEFT 0.20.0 / PyTorch 2.12.1
- Apple M4 Max, MPS, BF16 (the failure is in preprocessing and appears device-independent)

## Reproduction

Use a normal `system -> user -> assistant` ChatML row with the model's native chat template and:

```yaml
data:
  format: chatml
  train_on_responses_only: true
```

Training fails while building assistant-only labels:

```text
File "soup_cli/data/loss_mask.py", line 261, in build_assistant_only_labels
  rendered = _tokenize_only(tokenizer, cumulative)
jinja2.exceptions.TemplateError: No user query found in messages.
```

The preferred assistant-mask path is unavailable for this template, so the incremental fallback renders every cumulative prefix. Its first prefix contains only the system message. Qwen3.8's native template rejects that incomplete conversation before the user turn is appended.

## Expected behavior

Response-only masking should preserve the native Qwen template and avoid rendering a known-invalid system-only prefix. Non-assistant tokens should remain masked, and the assistant span should still be derived from monotonic cumulative renders once a user turn exists.

## Workaround and impact

Setting `data.chat_template: chatml` allows training to complete, but replaces the model's native template with Soup's generic ChatML template and saves that replacement into the adapter. This is not a safe default for a long Qwen3.8 fine-tune.

I reproduced this in a real 27B smoke test and am working on a focused regression test and fix.

Amix29:
## Summary

- defer ignored system/developer-only prefixes in the incremental masking fallback until a user-bearing prefix can be rendered
- preserve the existing assistant boundary and ignored-prefix semantics
- cover both response-only and per-message `train` masking paths

This was found during the Qwen3.8-27B Apple Silicon smoke test reported in #472. The native Qwen3.8 chat template rejects a transient system-only prefix with `TemplateError: No user query found in messages.`, even though the complete conversation is valid.

## Validation

- `18855 passed, 82 skipped, 5 deselected` in the full non-smoke suite
- `356 passed` across focused masking and chat-template tests
- real cached `Qwen/Qwen3.8-27B` native tokenizer: all 128 smoke-test training rows preprocess successfully and retain supervised tokens

Fixes #540

### DIFF
--- a/src/soup_cli/data/loss_mask.py
+++ b/src/soup_cli/data/loss_mask.py
@@ -19,12 +19,14 @@ Two strategies:
 
 2. **Fallback**: Render ``messages[:i]`` vs ``messages[:i+1]`` for each turn and
    take the token delta. The delta is the new turn's tokens (prefix + content +
-   suffix). Special tokens like BOS are added by the Jinja template itself
-   (not by the tokenizer ``__call__``), so monotone-prefix templates produce
-   stable deltas. We pass ``add_special_tokens=False`` to incremental tokenize
-   calls so HF does not double-prepend BOS at the front of each render. This
-   path is necessarily looser than the preferred path — the role-prefix tokens
-   (e.g. ``<|assistant|>``) end up in the loss too. Users wanting strict
+   suffix). Leading non-trainable context is deferred until the first user turn,
+   because some native templates reject an otherwise-transient system-only
+   prefix. Special tokens like BOS are added by the Jinja template itself (not
+   by the tokenizer ``__call__``), so monotone-prefix templates produce stable
+   deltas. We pass ``add_special_tokens=False`` to incremental tokenize calls so
+   HF does not double-prepend BOS at the front of each render. This path is
+   necessarily looser than the preferred path — the role-prefix tokens (e.g.
+   ``<|assistant|>``) end up in the loss too. Users wanting strict
    assistant-content-only must pass a tokenizer with ``{% generation %}`` markers.
 """
 
@@ -171,6 +173,17 @@ def _tokenize_only(tokenizer: Any, messages: Sequence[dict]) -> list[int]:
     return coerce_token_ids(out)
 
 
+def _has_user_turn(messages: Sequence[dict]) -> bool:
+    """Whether a cumulative conversation contains a user query.
+
+    Qwen3.8's native template refuses a system-only prefix. The incremental
+    fallback does not need to render leading ignored context by itself: the
+    first user render includes that context and establishes the same boundary
+    before an assistant turn is unmasked.
+    """
+    return any(message.get("role") == "user" for message in messages)
+
+
 def _truncate(
     input_ids: list[int], labels: list[int], max_length: int
 ) -> dict[str, list[int]]:
@@ -258,6 +271,11 @@ def build_assistant_only_labels(
     cumulative: list[dict] = []
     for msg in messages:
         cumulative.append(msg)
+        if msg.get("role") != "assistant" and not _has_user_turn(cumulative):
+            # Do not feed a transient system/developer-only prefix to native
+            # templates that require a user query (#540). Those tokens remain
+            # ignored and are included in the first renderable user prefix.
+            continue
         rendered = _tokenize_only(tokenizer, cumulative)
         new_len = len(rendered)
         if msg.get("role") == "assistant":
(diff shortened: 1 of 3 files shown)


MakazhanAlpamys:
Thank you for the focused #540 fix. I verified the diff, ran 34 masking tests and ruff, and all 13 CI jobs passed. Replacing the response-only defer condition with if False fails the named system-prefix regression. One merge blocker: add changelog.d/0.73.3/543.fixed.md with a bullet and #543. No docs update is needed for this narrow fix.

Amix29:
Addressed in `175048c`.

Added the required `changelog.d/0.73.3/543.fixed.md` fragment with a bullet and a link to #543. The focused masking suite remains green (`34 passed`), Ruff is clean, and `git diff --check` passes.

src/soup_cli/data/loss_mask.py
```diff
@@ -343,11 +361,15 @@ def build_per_message_train_labels(
     cumulative: list[dict] = []
     for msg in messages:
         cumulative.append(msg)
-        rendered = _tokenize_only(tokenizer, cumulative)
-        new_len = len(rendered)
         train_flag = msg.get("train")
         if train_flag is None:
             train_flag = msg.get("role") == "assistant"
+        if not train_flag and not _has_user_turn(cumulative):
```
AmirF194: Ran this against 175048c9b32b580b742510803c9f1895a7bb0be9 in a clean container, using the PR's own `_RequiresUserTokenizer` fixture shape.

The deferral guard at line 367 only checks `train_flag`, not whether the tokenizer can render the prefix at all. If a leading non-assistant message is explicitly marked `train: true` (a valid input for this function's per-message mode), the guard does not defer it, and `_tokenize_only` is called on a user-less cumulative prefix, which raises on exactly the native templates this PR targets.

```
messages = [
    {"role": "system", "content": "Follow the format exactly.", "train": True},
    {"role": "user", "content": "Q", "train": False},
    {"role": "assistant", "content": "A", "train": True},
]
build_per_message_train_labels(messages, _RequiresUserTokenizer())

CRASHED as hypothesized: No user query found in messages.
rendered_roles before crash: [('system', 'user', 'assistant'), ('system',)]
```

The full-conversation render at the top of the function succeeds (it always includes the eventual user turn), but the second render, `cumulative=[system]`, is exactly the user-less prefix `_has_user_turn` was added to catch.

Suggest deferring on `not _has_user_turn(cumulative)` alone, independent of `train_flag`, same as `build_assistant_only_labels` already does two functions up. The trained span for a deferred leading message still gets covered once a user turn makes the prefix renderable, so this shouldn't need to change the labeling logic itself, just drop the `train_flag` from the skip condition.

Amix29: Thank you for the precise reproduction and for checking the actual template-validity property — this was an excellent catch.

Addressed in `41e99d9`. The incremental per-message fallback now defers every user-less prefix independently of `train_flag`, so an explicitly trained leading system message is never rendered before the first user turn. I added the exact `system(train: true) -> user(train: false) -> assistant(train: true)` regression with `_RequiresUserTokenizer`; restoring the previous `not train_flag` gate makes it crash on the system-only prefix.

Validation: `pytest -q tests/test_assistant_mask.py --no-cov` (35 passed), Ruff clean, and `git diff --check` clean.

AmirF194: Confirmed, this fix looks correct and the regression test covers the case.

Amix29:
Fixed the deferred-prefix labeling regression in the latest commit. An explicitly trainable system/developer prefix is now recovered from the first valid user-containing render without ever rendering an invalid system-only conversation. The first user turn is aligned separately, so it cannot absorb an ignored system prefix.

Validation: 36 focused tests passed; the full suite passed with 18,857 tests, 82 skipped, and 5 deselected; Ruff is clean.

MakazhanAlpamys:
Thank you, @Amix29. #543 was squash-merged. The final change preserves explicitly trainable system/developer prefixes while avoiding invalid system-only rendering and separately aligns the first user turn. I ran 36 masking tests; changing the defer condition to if False failed the named regression. All 13 CI jobs passed.

PR #544, opened 2026-08-25
fix(monitoring): preserve final loss in live panel

### LINKED ISSUE #541: Final live training panel resets the last loss to 0.0000
## Environment

- Soup 0.73.3
- Transformers 5.12.1
- Apple M4 Max / MPS

## Observed behavior

After a successful 128-step SFT run, the final live training panel displayed:

```text
Loss:  0.0000    LR: 0.00e+00
```

The Trainer history and Soup's completion panel both contained the real values:

```text
last step loss: 1.3589
completion panel: Loss: 4.3004 -> 1.3589
```

The final Trainer summary log contains runtime/throughput fields but no per-step `loss`. `SoupTrainerCallback.on_log()` currently defaults a missing loss to `0.0` and overwrites the last valid value in `TrainingDisplay`.

## Expected behavior

Summary-only log events should preserve the most recent training loss, learning rate and gradient norm in the live panel rather than replacing missing values with zero. A genuine logged `0.0` must still be displayed as zero.

I reproduced this in the Qwen3.8-27B Apple Silicon smoke test and am preparing a focused callback regression test and fix.

Amix29:
## Summary

- retain the last real loss, learning rate, and gradient norm seen by the training callback
- prevent Hugging Face's final runtime-only log event from replacing those values with synthetic zeroes
- preserve genuine logged `0.0` values

This was observed at the end of the Qwen3.8-27B Apple Silicon smoke test reported in #472: training completed with a final logged loss of `1.3589`, but the completed live panel displayed `0.0000` after the summary event.

## Validation

- focused callback, display, live-monitor, watchdog, and training-intelligence suite: `310 passed`
- regression coverage for both summary-only events and genuine zero-valued metrics
- full non-smoke suite: `18855 passed, 82 skipped, 5 deselected`

Fixes #541

### DIFF
--- a/changelog.d/0.73.3/544.fixed.md
+++ b/changelog.d/0.73.3/544.fixed.md
@@ -1,0 +1,2 @@
+- The final training panel now preserves the last real loss, learning rate, and
+  gradient norm when Transformers emits a summary-only log event (#544).
--- a/src/soup_cli/monitoring/callback.py
+++ b/src/soup_cli/monitoring/callback.py
@@ -54,6 +54,13 @@ class _SoupTrainerCallback_body:  # noqa: N801
         self.run_id = run_id
         self.eval_config = eval_config
         self.output_dir = output_dir
+        # HF emits a summary-only ``on_log`` event after the final optimizer
+        # step. It has runtime/throughput fields but no loss/LR/grad norm. Keep
+        # the last real values so that event cannot reset the dashboards and
+        # tracker to synthetic zeroes (#541).
+        self._last_loss = 0.0
+        self._last_lr = 0.0
+        self._last_grad_norm = 0.0
         # Loss watchdog state
         self._watchdog_enabled = loss_watchdog
         self._watchdog_threshold = loss_watchdog_threshold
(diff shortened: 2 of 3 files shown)


MakazhanAlpamys:
Thank you for the focused #541 fix. I ran 8 callback tests and ruff; all 13 CI jobs passed. Disabling the cached loss update makes the named summary regression fail. One merge blocker: add changelog.d/0.73.3/544.fixed.md with a bullet and #544. No separate docs update is needed.

Amix29:
Addressed in `ef52601`.

Added the required `changelog.d/0.73.3/544.fixed.md` fragment with a bullet and a link to #544. The focused callback suite remains green (`8 passed`), Ruff is clean, and `git diff --check` passes.

MakazhanAlpamys:
Thank you, @Amix29. #544 was squash-merged. The fix retains the last real metric while preserving genuine 0.0 values, so a runtime-only summary cannot overwrite the final loss. Disabling cached loss assignment failed the named summary regression; changelog is present and all 13 CI jobs passed.

PR #545, opened 2026-08-25
fix(repro): capture Apple Silicon accelerator metadata

### LINKED ISSUE #542: Reproducibility receipt omits Apple Silicon accelerator metadata
## Environment

- Soup 0.73.3
- PyTorch 2.12.1
- Apple M4 Max, 128 GB unified memory, MPS

## Observed behavior

`soup train --repro-receipt ...` wrote a valid receipt, but the accelerator portion was empty:

```json
{
  "arch": "arm64",
  "cuda_version": null,
  "gpu_models": [],
  "driver_version": null,
  "torch_version": "2.12.1"
}
```

The receipt only probes `torch.cuda`, so a substantial Apple Silicon run cannot identify that it used MPS, which Apple chip was used, or how much unified memory was available. This matters for reproducibility because the same `arm64` architecture covers very different M-series chips and memory capacities.

## Expected behavior

When `torch.backends.mps.is_available()` is true, the receipt should add privacy-safe, best-effort metadata such as:

- accelerator backend: `mps`;
- Apple chip name (for example `Apple M4 Max`);
- unified memory bytes.

It must not collect serial numbers, hardware UUIDs or user-specific paths. Existing CUDA fields and JSON compatibility should remain intact.

I reproduced this during the Qwen3.8-27B Apple Silicon smoke test and am preparing a focused fix with mocked cross-platform tests.

Amix29:
## Summary

- identify MPS as the accelerator backend in reproducibility receipts
- record the Apple chip name and unified-memory capacity through a strict allowlist of privacy-safe macOS sysctl keys
- preserve the existing `ReproReceipt` constructor signature by making the new fields optional
- document that serial numbers, hardware UUIDs, and user-specific paths are not collected

This was found while reviewing the reproducibility receipt from the Qwen3.8-27B Apple Silicon smoke test reported in #472. The receipt captured PyTorch and Darwin, but left `gpu_models` empty and did not identify MPS or unified memory.

## Validation

- focused governance and receipt suite: `196 passed`
- live Apple Silicon probe reports `mps`, `Apple M4 Max`, and `137438953472` bytes of unified memory
- missing-sysctl fallback and old positional constructor coverage
- full non-smoke suite: `18856 passed, 82 skipped, 5 deselected`

Fixes #542

### DIFF
--- a/changelog.d/0.73.3/545.fixed.md
+++ b/changelog.d/0.73.3/545.fixed.md
@@ -1,0 +1,2 @@
+- Reproducibility receipts now record the MPS backend, Apple chip name, and
+  unified-memory capacity without collecting unique hardware identifiers (#545).
--- a/docs/adapters-and-governance.md
+++ b/docs/adapters-and-governance.md
@@ -481,7 +481,9 @@ best-effort — a broken audit log never crashes the CLI.
 ## Reproducibility Receipt (`soup train --repro-receipt`)
 
 SR 11-7-style reproducibility receipt captures seeds (torch + numpy + python), kernel
-versions (CUDA + cuDNN + NCCL), GPU model + driver, OS + arch:
+versions (CUDA + cuDNN + NCCL), accelerator backend, GPU model + driver, OS + arch.
+On Apple Silicon it records the privacy-safe chip name and unified-memory capacity, but
+never a serial number, hardware UUID, or user-specific path:
 
 ```bash
 soup train --config soup.yaml --repro-receipt repro.json
--- a/src/soup_cli/utils/repro_receipt.py
+++ b/src/soup_cli/utils/repro_receipt.py
@@ -3,8 +3,8 @@
 Captures the minimum environment fingerprint needed for a regulated-org
 audit: seeds (torch + numpy + python), Python interpreter version, OS +
 arch, soup_cli version, kernel versions (CUDA / cuDNN / NCCL — best-effort
-from torch when available), GPU model + driver (best-effort from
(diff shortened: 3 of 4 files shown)


MakazhanAlpamys:
Thank you for the scoped #542 change. I ran the focused receipt suite, ruff, and diff check; all 13 CI jobs passed. However, deleting the production MPS bridge still left the original tests green. Add an integration regression that builds and serializes a receipt and requires MPS backend, chip, and memory; it must fail without the bridge. Also add changelog.d/0.73.3/545.fixed.md.

Amix29:
Addressed both blockers in `a45c963`.

- Added `changelog.d/0.73.3/545.fixed.md` with a bullet and a link to #545.
- Added an integration regression that installs a complete mock `torch` module, calls `build_repro_receipt()`, serializes it through `receipt_to_dict()`, and requires the MPS backend, Apple chip, and unified-memory capacity in the final payload.

I also ran the exact mutation you described: removing the `_detect_torch_kernel_versions()` bridge makes the new integration test fail with `assert None == "mps"`. Restoring the bridge makes it pass.

Validation:

- `pytest -q tests/test_v0590.py -k ReproReceipt --no-cov`: `11 passed, 98 deselected`
- focused mutation control: red without the bridge, green with it
- Ruff and `git diff --check`: clean

MakazhanAlpamys:
Thank you, @Amix29. #545 was squash-merged. The new integration test builds and serializes a real receipt rather than testing the helper in isolation. I removed the MPS metadata bridge and mps_metadata_reaches_serialized_receipt failed with None == mps; after restoration, all 11 focused tests and 13 CI jobs passed.

PR #551, opened 2026-08-25
fix(data): reject non-finite Best-of-N scores

### LINKED ISSUE #547: Best-of-N can select a NaN-scored candidate and emit non-standard JSON
## Reproduction

Audited on current `main` at `6c13c44`.

`judge_pick_best` converts judge outputs with `float(...)`, but never checks that the result is finite:

```python
class Judge:
    def evaluate(self, _prompt, response):
        score = float("nan") if response == "first" else 1.0
        return SimpleNamespace(weighted_score=score)

pick = judge_pick_best("q", ["first", "second"], Judge())
```

Observed result:

```text
winner_idx=0
scores=(nan, 1.0)
```

The `NaN` candidate wins because Python's ordering comparisons against `NaN` are false and `max(...)` retains the first item. `build_sft_row` then reaches `json.dumps`, which emits the non-standard token `NaN` in the JSONL. The same score tuple is also used to select the DPO loser.

This is especially dangerous for deterministic-reward and offline-verification pipelines: one broken judge/reward value can silently select the wrong response instead of failing closed.

## Expected

- Reject `NaN`, `+inf`, and `-inf` before winner/loser selection.
- Name the prompt/candidate index in the error without echoing private prompt contents.
- Never publish SFT or DPO output containing a non-finite score.
- Pin candidate-order controls so a non-finite value cannot win or lose depending on its position.

## Acceptance criteria

- [ ] `judge_pick_best` accepts finite numeric scores only (and rejects booleans explicitly).
- [ ] Tests cover `NaN`/both infinities in the first, middle, and last candidate positions.
- [ ] Both SFT and DPO materialization remain unreachable after a non-finite score.
- [ ] Valid finite-score behavior and lowest-index tie-breaking remain unchanged.

I am taking this issue.

Amix29:
## Summary

- reject boolean, non-numeric, `NaN`, and infinite Best-of-N judge scores before winner selection
- identify the failing candidate index without echoing private prompt or candidate content
- prove that invalid scores cannot publish SFT or DPO output

The previous `max(...)` selection could retain a first-position `NaN` as the winner and serialize the non-standard `NaN` token into JSONL.

## Validation

- Best-of-N focused suite: `25 passed`
- full non-smoke suite: `18935 passed, 82 skipped, 5 deselected`
- candidate-order coverage for `NaN`, `+inf`, and `-inf`, plus explicit boolean rejection
- Ruff and `git diff --check`: clean

Fixes #547

### DIFF
--- a/changelog.d/0.73.3/547.fixed.md
+++ b/changelog.d/0.73.3/547.fixed.md
@@ -1,0 +1,2 @@
+- Best-of-N now rejects non-finite and boolean judge scores before selecting a
+  winner or emitting invalid JSONL (#547).
--- a/src/soup_cli/utils/best_of_n.py
+++ b/src/soup_cli/utils/best_of_n.py
@@ -11,6 +11,7 @@ provenance under the reserved ``_best_of_n`` key; ``build_dpo_pair`` optionally
 emits a winner-vs-loser preference pair.
 """
 
+import math
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Callable, Optional, Protocol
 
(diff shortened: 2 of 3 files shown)


MakazhanAlpamys:
Thank you, @Amix29. #551 was squash-merged. Non-finite scores are rejected before argmax and SFT/DPO publication without disclosing private prompt/candidate data. Removing the finite-score gate failed both positional and CLI no-output controls. Changelog is present and all 13 CI jobs passed.

PR #552, opened 2026-08-25
fix(data): fail closed on invalid Best-of-N prompts

### LINKED ISSUE #548: Best-of-N silently drops malformed prompt rows without reporting data loss
## Reproduction

Audited on current `main` at `6c13c44`.

Given a four-line prompt file:

```jsonl
{"prompt":"kept-1"}
{this is not valid json}
{"messages":[{"role":"assistant","content":"no user turn"}]}
{"prompt":"kept-2"}
```

`_bon_load_prompts(...)` returns:

```python
["kept-1", "kept-2"]
```

There is no warning, rejected-row count, line number, or machine-readable rejection record. The loader catches `JSONDecodeError` and continues; non-object rows and rows without a usable prompt are also silently skipped.

For a Best-of-N dataset build, that is silent data loss. It also makes completeness checks impossible because output rows do not retain a stable source line/id.

## Expected

The default path should fail closed on malformed or unsupported non-empty rows. If a permissive mode is useful, it must be explicit and report exactly what was skipped.

## Acceptance criteria

- [ ] Default mode rejects the first malformed/non-usable non-empty row with its line number and reason.
- [ ] If `--skip-invalid` is added, the final panel reports read/kept/skipped counts and a bounded rejection report can be emitted.
- [ ] A stable source row id or source line is carried into Best-of-N provenance so output completeness is auditable.
- [ ] Tests cover malformed JSON, non-object JSON, missing prompt/user content, empty strings, and blank lines (blank lines may remain ignorable by contract).
- [ ] No rejected error message needs to echo private prompt content.

I am taking this issue.

Amix29:
## Summary

- fail closed on malformed, non-object, or unusable nonblank prompt rows
- report the physical JSONL line number without echoing private prompt content
- preserve each accepted row's `source_line` in Best-of-N provenance
- keep intentional blank-line handling while preventing silent cardinality drift

## Validation

- `PYTHONPATH=src pytest tests/test_v07131.py -q --no-cov --tb=short` (41 passed, 53 deselected)
- `PYTHONPATH=src pytest tests/ -q --no-cov --tb=short` (18,934 passed, 82 skipped, 5 deselected)
- `git diff --check`

Fixes #548

### DIFF
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -2920,8 +2920,31 @@ def _load_bon_model(base: str, device: str, trust: bool):
     return model, tok
 
 
-def _bon_load_prompts(path: str) -> list:
-    """Read a JSONL of {prompt|messages|instruction} rows -> list[str]."""
+def _bon_prompt_text(row: dict, line_number: int) -> str:
+    """Extract one prompt without echoing private row contents in errors."""
+    text = None
+    if isinstance(row.get("prompt"), str):
+        text = row["prompt"]
+    elif isinstance(row.get("instruction"), str):
+        text = row["instruction"]
+    elif isinstance(row.get("messages"), list):
+        users = [
+            message.get("content")
+            for message in row["messages"]
+            if isinstance(message, dict) and message.get("role") == "user"
+        ]
+        if users and isinstance(users[-1], str):
+            text = users[-1]
+    if text is None or not text.strip():
+        raise ValueError(
+            f"prompt JSONL line {line_number} has no non-empty prompt, "
+            "instruction, or user message"
+        )
+    return text
+
+
+def _bon_load_prompt_records(path: str) -> list[tuple[str, int]]:
+    """Read prompt JSONL strictly as ``(text, source_line)`` records."""
     from soup_cli.utils.paths import enforce_under_cwd_and_no_symlink
 
     enforce_under_cwd_and_no_symlink(path, "--prompts path")
@@ -2930,9 +2953,9 @@ def _bon_load_prompts(path: str) -> list:
         raise FileNotFoundError(path)
     if os.path.getsize(real) > _BON_MAX_JSONL_BYTES:
         raise ValueError(f"--prompts file exceeds {_BON_MAX_JSONL_BYTES} bytes")
-    prompts: list = []
+    prompts: list[tuple[str, int]] = []
     with open(real, "r", encoding="utf-8") as handle:
-        for raw_line in handle:
+        for line_number, raw_line in enumerate(handle, start=1):
             stripped = raw_line.strip()
             if not stripped:
                 continue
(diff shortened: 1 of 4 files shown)


src/soup_cli/commands/data.py
```diff
@@ -2930,38 +2953,31 @@ def _bon_load_prompts(path: str) -> list:
...
-                if users and isinstance(users[-1], str):
-                    text = users[-1]
-            if text:
-                prompts.append(text)
+                raise ValueError(f"prompt JSONL line {line_number} must be a JSON object")
+            prompts.append((_bon_prompt_text(row, line_number), line_number))
     return prompts
 
 
+def _bon_load_prompts(path: str) -> list[str]:
```
AmirF194: I ran the old and new prompt loaders against a232b838 in a clean python:3.11-slim container, isolated from the rest of the CLI.

`_bon_load_prompts` (data.py:2976-2978) is a thin wrapper over the new `_bon_load_prompt_records` (data.py:2946-2973), and `evolve` still calls it directly at data.py:3236. That function now raises `ValueError` on the first malformed JSON line, non-dict row, or row with no usable prompt text, where main's version silently skipped such a row and kept going. `evolve`'s own `try` block (data.py:3235) already catches `ValueError` and exits the whole command, so a seeds file that used to run with a few rows dropped now aborts entirely on the first bad row.

```
seeds.jsonl:
{"prompt": "good seed one"}
{not valid json}
{"prompt": "good seed two"}

main   _bon_load_prompts(seeds.jsonl) -> ['good seed one', 'good seed two']
PR     _bon_load_prompts(seeds.jsonl) -> raises ValueError: prompt JSONL line 2 is not valid JSON
```

The PR title, changelog entry and new tests all scope this to best-of-n; `evolve` isn't mentioned and has no test covering the new behavior on a mixed-quality seed file. If the stricter behavior is meant to apply to `evolve` too, it is worth a line in the changelog and a test there; if not, `evolve` needs its own loader that keeps skipping bad rows rather than sharing this one.

Amix29: Thank you for isolating the cross-command regression — this was a very useful catch.

Fixed in `72ff37a`. `evolve` now uses its own tolerant seed loader and keeps skipping malformed JSON, non-object rows, and rows without usable prompt text, while Best-of-N retains the strict, line-numbered fail-closed loader introduced by this PR.

I added an end-to-end `evolve` regression using the same mixed-quality shape: two valid seeds separated by malformed JSON, a non-object row, an assistant-only row, and a blank prompt. The command completes and evolves exactly the two valid seeds.

Validation:
- `PYTHONPATH="$PWD/src" python3 -m pytest -q tests/test_v07131.py --no-cov` — 95 passed
- full suite — 18,935 passed, 82 skipped, 5 deselected
- full Ruff check and `git diff --check` — clean

AmirF194: Thanks for fixing this quickly, looks right.

MakazhanAlpamys:
Thank you, @Amix29. #552 was squash-merged. External prompt input is strict while historical permissive seed loading remains separate. Changing malformed JSON rejection from raise to continue made the CLI incorrectly exit 0 and the regression caught it. Docs, changelog, and all 13 CI jobs are green.

PR #553, opened 2026-08-25
fix(data): resume Best-of-N from durable checkpoints

### LINKED ISSUE #549: Best-of-N discards all completed work after a late sampler or judge failure
## Reproduction

Audited on current `main` at `6c13c44` with two prompts and two synthetic candidates per prompt.

- Prompt 1 samples and judges successfully.
- The judge raises `RuntimeError("simulated late judge outage")` on prompt 2.

Observed:

```text
exit=1
SFT output exists: false
DPO output exists: false
```

`best_of_n` accumulates every SFT/DPO line in memory and writes both files only after the entire prompt loop finishes. A sampler, model, or judge failure near the end therefore discards all earlier work. The exception also escapes without a command-level progress/recovery message.

At the supported limits (100,000 prompts, up to 64 candidates each), and particularly with a 27B local sampler, repeating the successful prefix can mean hours of avoidable generation.

## Expected

Best-of-N should have a durable, resumable per-prompt work record while retaining atomic final SFT/DPO publication.

## Acceptance criteria

- [ ] Completed prompt groups are checkpointed durably without exposing a partially-written final dataset.
- [ ] `--resume` validates the input digest and sampling/judge configuration before reusing work.
- [ ] Resume is exactly-once: no duplicated or reordered prompt groups.
- [ ] A late sampler/judge failure reports completed/total counts and the recovery artifact path.
- [ ] A test fails on prompt N, resumes, and proves prompts `0..N-1` are not sampled or judged again.
- [ ] Final SFT and optional DPO outputs are published as one consistent generation (or through a manifest that detects a mismatched pair).

I am taking this issue.

Amix29:
## Summary

- fsync one private, checksummed checkpoint entry after each completed prompt group
- add `--resume` with exact prompt/config digest validation and sequential exactly-once replay
- skip all already completed sampling and judging, including fully completed checkpoints
- report completed/total progress and the recovery artifact after a late failure without echoing prompt data
- atomically publish final datasets and write a final manifest that binds SFT/DPO hashes and counts
- reject checkpoint path collisions, symlinks, reordered indexes, content corruption, and config drift

## Validation

- `PYTHONPATH=src pytest tests/ -q --no-cov --tb=short` (18,936 passed, 82 skipped, 5 deselected)
- final checkpoint/replay regression pass after checksum hardening (22 passed)
- `git diff --check`

Fixes #549

### DIFF
--- a/changelog.d/0.73.3/549.fixed.md
+++ b/changelog.d/0.73.3/549.fixed.md
@@ -1,0 +1,2 @@
+- Fixed #549 by making Best-of-N generation durable and resumable per prompt, with exactly-once
+  checkpoint validation and a final hash manifest binding SFT/DPO outputs.
--- a/changelog.d/0.73.3/549.fixed.md
+++ b/changelog.d/0.73.3/549.fixed.md
@@ -1,2 +1,3 @@
 - Fixed #549 by making Best-of-N generation durable and resumable per prompt, with exactly-once
-  checkpoint validation and a final hash manifest binding SFT/DPO outputs.
+  checkpoint validation and rollback-safe, manifest-last publication that never exposes a partial
+  SFT/DPO generation after an output failure.
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -301,7 +301,7 @@ soup local-rl train --db <path> --model <id> [--scheduler-dir <dir>] [--hour H] 
 soup build <manifest.yaml> [--dry-run] [--output-dir <dir>]  dbt-for-SFT DAG: validate + plan + live materialise (v0.69.0; live v0.71.6)
 soup expect <data.jsonl> <suite.yaml>         Expectations suite: PII / token-length / refusal / judge (v0.69.0)
 soup data gen-magpie --base <m> --provider ollama|vllm --target N --output <jsonl> [--base-url <url>] [--quality-filter]  Magpie synthetic generator — live (v0.69.0; live v0.71.6)
-soup data best-of-n (--base <m> | --provider ollama|vllm --model <m> [--base-url <url>]) --prompts <jsonl> --n 8 --judge <url> -o <sft.jsonl> [--emit-pairs <dpo.jsonl>]  Best-of-N rejection sampling: sample N locally (default) or through a raw-completion provider, then emit SFT (+ DPO) rows
+soup data best-of-n (--base <m> | --provider ollama|vllm --model <m> [--base-url <url>]) --prompts <jsonl> --n 8 --judge <url> -o <sft.jsonl> [--emit-pairs <dpo.jsonl>] [--resume] [--checkpoint <journal.jsonl>] [--manifest <manifest.json>]  Best-of-N rejection sampling with durable per-prompt recovery and manifest-last publication
 soup data evolve --input <seeds.jsonl> --provider ollama|vllm --model <m> --strategy depth|breadth --rounds N -o <jsonl>  Evol-Instruct (WizardLM) instruction evolution (v0.71.31)
 soup data persona-mix --prompts <jsonl> --n N --output <jsonl>  Persona-Hub diversity sampler (v0.69.0)
 soup data brain-rot <data.jsonl> [--strict]   Brain-rot detector — arXiv 2510.13928 (v0.69.0)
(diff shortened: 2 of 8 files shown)


Amix29:
Fixed the three Windows CI failures in `3480d54`.

The manifest hashes were computed from LF-delimited UTF-8 text, while text-mode publication translated those line endings to CRLF on Windows. The manifest therefore described different bytes from the files on disk. SFT and DPO JSONL outputs are now published through exact UTF-8 byte writes; the manifest remains the final text artifact.

Added a regression that prevents dataset outputs from returning to text-mode publication and verifies both manifest hashes against the final file bytes. Focused validation: 22 passed, Ruff clean.

Amix29:
Fixed the two resume blockers in f082d81:

- local stochastic sampling now derives a stable torch seed from `(run seed, prompt index)`, so resumed prompt candidates match an uninterrupted run;
- an interrupted final JSONL append without its commit newline is safely truncated before checkpoint validation, while committed records still receive full digest/index validation.

Validation: 8 focused resume tests passed, including fail/resume RNG equivalence and partial-tail recovery; Ruff is clean.

MakazhanAlpamys:
Thank you for the correct scope, docs, and resume coverage. The mandatory mutation found a merge blocker: changing manual_seed(prompt_seed(seed, index)) to manual_seed(seed) leaves all 8 focused resume tests green. Add a regression with at least two prompt indices that compares final candidates/artifacts from an uninterrupted run against fail-and-resume; it must fail under that mutation. #554 waits on this merge.

Amix29:
Addressed the determinism blocker and rebased #553 onto current `main` (`5812a79`) in `58f67e2`.\n\nThe strengthened regression now uses three prompt indices and compares the final candidate sets plus the exact SFT, DPO, and manifest bytes from an uninterrupted run against an interrupted-then-resumed run. It also proves that prompt indices receive distinct candidate streams. The requested mutation, replacing `manual_seed(prompt_seed(seed, index))` with `manual_seed(seed)`, now fails the named test with `assert 1 == 3`; restoring the production derivation makes it pass.\n\nDuring the rebase I also preserved the merged #551 fail-closed contract: deterministic `ValueError` validation failures are not mislabeled as resumable outages.\n\nValidation: 106 Best-of-N/resume regression tests passed; Ruff and `git diff --check` are clean. Fresh CI is now running on the rebased branch.

Amix29:
Full local validation on the rebased branch is also clean: 18,967 passed, 82 skipped, 5 deselected, with 82.82% coverage and no failures.

MakazhanAlpamys (changes requested):
Changes requested. I verified 106 Best-of-N tests, full Ruff, and the required mutation; disabling durable `append_checkpoint` caused the resume regression to fail as expected.

Two follow-ups are required before merge:

- P2: a late sampler/judge `ValueError` after a prefix has been checkpointed bypasses the recovery message. The user is not told how many items completed or where the checkpoint is, which makes recovery ambiguous.
- P2: `docs/commands.md` still does not document the new `--resume`, `--checkpoint`, and `--manifest` behavior.

Please add the recovery details and command documentation, then rerun the full required checks on the current head. The focused tests and mutation are good evidence for the core fix, but they do not cover these user-facing paths.

Amix29:
Published the review follow-up in `cf8b1d3`.

- sampler and judge backend failures are now distinguished from deterministic validation failures;
- late backend failures preserve the durable checkpoint and report bounded `--resume` guidance without exposing the private backend exception;
- command documentation now describes the recovery and manifest-last publication contract;
- regression coverage exercises failures in both the sampler and judge after a completed prefix.

Local validation before publication: full suite **18,969 passed**, plus focused recovery tests and Ruff.

MakazhanAlpamys (changes requested):
Thanks for the durable-checkpoint coverage. I verified the focused resume tests and a mutation that removes checkpoint appends; the replay regression fails as intended.

Request changes before merge: final publication is not atomic as a group. Injecting `OSError` while writing the DPO output leaves `sft_exists=True`, `dpo_exists=False`, and no manifest. This violates #549's consistent-generation acceptance criterion and exposes a partial final dataset. Please publish the SFT/DPO/manifest set as one recoverable operation (or roll back newly written outputs) and add a regression covering both newly created and pre-existing output paths.

The current CI is green, but this is a merge blocker.

Amix29:
Addressed the grouped-publication blocker in `ff18a80` and rebased the PR onto current `main` (`0fcf383`). Thank you for the precise failure injection.

Before replacing any final artifact, Best-of-N now snapshots every pre-existing SFT, DPO, and manifest target to private sibling files. It publishes each member atomically with the manifest last; if any member fails, it restores the complete previous generation or removes the complete newly-created generation.

The regression matrix covers:

- newly-created and pre-existing target sets;
- failures at both the DPO and manifest members;
- failures before replacement and after replacement;
- exact restoration/removal of all three artifacts, with no rollback snapshots left behind.

The new regression fails against the previous implementation by exposing the partial SFT artifact and passes with the rollback implementation.

Validation on the rebased head:

- focused Best-of-N suite: **253 passed, 15 skipped**;
- full suite: **19,031 passed, 82 skipped, 5 deselected**;
- coverage: **82.82%**;
- full Ruff and `git diff --check`: clean.

MakazhanAlpamys (approved):
Re-reviewed current HEAD ff18a80943e3adf9b3fb27a0238c89bd8996b619 against issue #549. The grouped publication now snapshots the complete SFT/DPO/manifest generation, publishes the manifest last, and restores or removes the whole set on failures; the resume path preserves deterministic exactly-once progress and bounded recovery guidance. No tests were deleted (the changed test files are additions or modifications only). I independently ran the relevant resume/publication suite: 116 tests passed, and Ruff passed on all changed Python files. The required mutation omitting attempted-artifact tracking caused all 8 grouped-publication regression cases to fail, confirming the rollback guard is load-bearing. All 13 checks from `gh pr checks 553` are green on this HEAD. Approving for squash merge.

MakazhanAlpamys:
Thank you, AmixDigital, for carrying the Best-of-N recovery path through the hard failure cases. The implementation checkpoints each completed prompt group, validates exact-once resume state, preserves deterministic candidate streams, and publishes SFT/DPO/manifest artifacts as an all-or-old generation with rollback. I independently ran 116 resume/publication and online regression tests; Ruff passed. Mutating the rollback tracking made all 8 grouped-publication cases fail, and the resume-prefix/RNG mutations also failed their named tests. All 13 current-head required checks were green. Merged as `300a671`.

PR #554, opened 2026-08-25
feat(data): add offline Best-of-N materialization

Amix29:
## Summary

- add sampling-only `--export-candidates` mode that never constructs a judge
- preserve every ordered candidate with prompt/candidate/group digests and a bounded public sampler specification
- add `--candidate-artifact` + `--judgments` materialization that never constructs a sampler, loads a model, or makes a network call
- fail closed before publication on missing/duplicate judgments, changed bindings, invalid indexes, score-count drift, non-finite scores, and winner/score disagreement
- retain candidate-artifact and judgment-file hashes plus sampler/verifier provenance in both SFT and DPO rows
- exclude endpoint URLs and local model paths from artifacts/provenance and make identical offline inputs byte-stable
- preserve the existing one-command online workflow

## Validation

- `PYTHONPATH=src pytest tests/ -q --no-cov --tb=short` (18,942 passed, 82 skipped, 5 deselected)
- focused offline-contract and changelog tests (27 passed)
- `ruff check src/soup_cli/ tests/`
- `git diff --check`
- new module passes mypy; the command module retains only its two pre-existing baseline findings

## Merge order

This is independently based on `main`, but should be merged after #551, #552, and #553. I will rebase it after those predecessors land so the online finite-score, strict-prompt, and checkpoint paths are composed cleanly.

Fixes #550

PR #557, opened 2026-08-25
fix(data): commit offline Best-of-N generations

### LINKED ISSUE #556: Offline Best-of-N can publish a partial SFT/DPO generation without a manifest
## Problem

Found while integrating #549 and #550. This issue depends on #554.

Offline materialization validates candidate and judgment artifacts before writing, but then publishes SFT and optional DPO files sequentially. If the SFT write succeeds and the DPO write fails, the command exits 1 while leaving the new SFT file present and no DPO file. Because the offline mode has no final consistency manifest, this state is indistinguishable from an intentionally SFT-only generation unless every consumer manually compares row-level provenance.

Observed failure injection:

```text
exit=1
sft_exists=true
dpo_exists=false
```

The online resumable path in #553 already uses a manifest committed last to bind exact file hashes and row counts.

## Expected

Offline materialization should expose one final commit marker that binds the requested output set. Consumers must be able to reject interrupted or mismatched generations deterministically.

## Acceptance criteria

- [ ] Offline SFT and optional DPO outputs are bound by a final manifest written last.
- [ ] The manifest records exact UTF-8 file hashes, row counts, candidate-artifact hash, judgment hash, and whether DPO output was requested.
- [ ] `--manifest` has a safe default and all input/output/manifest paths must be distinct.
- [ ] A failure before manifest publication leaves no new valid commit marker.
- [ ] Reusing an old manifest with new or partial outputs is detected by a verifier/helper.
- [ ] Failure-injection tests cover SFT success followed by DPO failure and replacement of a previously complete generation.
- [ ] SFT-only materialization remains supported and is explicitly recorded as such.

I am taking this issue.

Amix29:
Closes #556.

## Summary
- publish an offline Best-of-N manifest last as the generation commit marker
- bind exact candidate/judgment digests, SFT/DPO byte hashes, row counts, filenames, and whether DPO was requested
- invalidate any prior manifest before replacing bound outputs so interrupted writes fail closed
- remove a previously committed sibling DPO artifact when an SFT-only replacement is published
- verify dataset hashes and row counts incrementally without buffering the generated files
- bind the model revision into the Best-of-N resume identity and reject revision changes before model loading
- reject `--resume` and `--checkpoint` before offline artifact loading or publication
- reject manifest path collisions and document the offline publication contract

## Validation
- `PYTHONPATH=src pytest -q --no-cov tests/test_issue549_bon_resume.py tests/test_issue550_bon_offline.py tests/test_v07131.py` (141 passed)
- full local current-head suite: 19,148 passed, 74 skipped, 5 deselected; the same 18 sandbox-only failures previously reproduced on pristine `origin/main`
- `git diff --check`
- all 13 fresh GitHub CI checks passed on Linux, Windows, and macOS

### DIFF
--- a/changelog.d/0.73.3/556.fixed.md
+++ b/changelog.d/0.73.3/556.fixed.md
@@ -1,2 +1,3 @@
 - Fixed #556: offline Best-of-N now commits SFT and optional DPO outputs through a
-  final verifiable manifest, so interrupted or mismatched generations fail closed.
+  final verifiable manifest, rejects unsupported online recovery options, and
+  makes interrupted or mismatched generations fail closed.
--- a/changelog.d/0.73.3/556.fixed.md
+++ b/changelog.d/0.73.3/556.fixed.md
@@ -1,0 +1,2 @@
+- Fixed #556: offline Best-of-N now commits SFT and optional DPO outputs through a
+  final verifiable manifest, so interrupted or mismatched generations fail closed.
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -345,9 +345,10 @@ local data such as a non-finite judge score remains a validation failure rather
 than being presented as a recoverable backend outage.
 
 For offline Best-of-N, the candidate and judgment artifacts are the recovery
-checkpoint. There is no offline `--resume`: rerun the exact materialization
-command after a late failure. The final manifest is written last and is the only
-commit marker; consumers must verify it and use only the SFT/DPO files it lists.
+checkpoint. Offline materialization rejects `--resume` and `--checkpoint`; rerun
+the exact materialization command after a late failure. The final manifest is
+written last and is the only commit marker; consumers must verify it and use only
+the SFT/DPO files it lists.
 
 ## Fine-tune from your coding agent (MCP)
 
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -303,7 +303,7 @@ soup expect <data.jsonl> <suite.yaml>         Expectations suite: PII / token-le
 soup data gen-magpie --base <m> --provider ollama|vllm --target N --output <jsonl> [--base-url <url>] [--quality-filter]  Magpie synthetic generator — live (v0.69.0; live v0.71.6)
 soup data best-of-n (--base <m> | --provider ollama|vllm --model <m> [--base-url <url>]) --prompts <jsonl> --n 8 --judge <url> -o <sft.jsonl> [--emit-pairs <dpo.jsonl>] [--resume] [--checkpoint <journal.jsonl>] [--manifest <manifest.json>]  Best-of-N rejection sampling with durable per-prompt recovery and manifest-last publication
 soup data best-of-n (--base <m> [--revision <rev>] | --provider ollama|vllm --model <m>) --prompts <jsonl> --n 8 --export-candidates <jsonl>  Sampling-only phase; no judge is constructed
-soup data best-of-n --candidate-artifact <jsonl> --judgments <jsonl> -o <sft.jsonl> [--emit-pairs <dpo.jsonl>]  Validate offline judgments and materialize byte-stable training rows without a model or network
+soup data best-of-n --candidate-artifact <jsonl> --judgments <jsonl> -o <sft.jsonl> [--emit-pairs <dpo.jsonl>] [--manifest <json>]  Validate offline judgments and materialize byte-stable training rows; a final manifest binds the requested output set
 soup data evolve --input <seeds.jsonl> --provider ollama|vllm --model <m> --strategy depth|breadth --rounds N -o <jsonl>  Evol-Instruct (WizardLM) instruction evolution (v0.71.31)
 soup data persona-mix --prompts <jsonl> --n N --output <jsonl>  Persona-Hub diversity sampler (v0.69.0)
 soup data brain-rot <data.jsonl> [--strict]   Brain-rot detector — arXiv 2510.13928 (v0.69.0)
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -344,6 +344,11 @@ treat only files listed by that final manifest as committed output. Invalid
 local data such as a non-finite judge score remains a validation failure rather
 than being presented as a recoverable backend outage.
 
+For offline Best-of-N, the candidate and judgment artifacts are the recovery
+checkpoint. There is no offline `--resume`: rerun the exact materialization
+command after a late failure. The final manifest is written last and is the only
+commit marker; consumers must verify it and use only the SFT/DPO files it lists.
+
 ## Fine-tune from your coding agent (MCP)
 
 `soup mcp serve` runs a [Model Context Protocol](https://modelcontextprotocol.io)
--- a/docs/data.md
+++ b/docs/data.md
@@ -241,7 +241,11 @@ closed. It does not construct a sampler or judge. SFT and DPO rows retain the
 candidate-artifact and judgment-file digests, group id, public sampler settings,
 and bounded verifier identity. Endpoint URLs and local model paths are never
 copied into those artifacts. Reusing the same two input files produces identical
-training-row bytes.
+training-row bytes. The offline command writes `<output>.manifest.json` last (or
+the explicit `--manifest` path) and binds the exact SFT/DPO hashes, row counts,
+candidate artifact, judgment file, and whether DPO output was requested. Treat
+the manifest as the commit marker: missing or mismatched manifests identify an
+interrupted or replaced generation.
 
 ### Custom Transforms
 
--- a/docs/data.md
+++ b/docs/data.md
@@ -249,11 +249,11 @@ interrupted or replaced generation.
 
 The candidate artifact and verified judgment file are the durable recovery
 boundary for offline materialization. This phase performs no sampling, so it has
-no progress checkpoint and `--resume` does not apply: after any late write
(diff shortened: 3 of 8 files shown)


MakazhanAlpamys:
Review notes for the draft: the manifest-last and pre-write invalidation design is directionally correct. I verified the offline tests, Ruff, and the mutation that catches removal of manifest invalidation.

Before this can leave draft status:

- P1: the shared `scores` validation still raises an uncaught `OverflowError` for an extremely large JSON integer instead of returning a normal validation error; add a regression test.
- P2: a later SFT-only run can leave a stale DPO output beside a manifest that says DPO was not requested. Remove or isolate the stale optional output, or document and enforce manifest-listed consumption.
- P2: add the late-failure recovery details and command documentation for resume/checkpoint/manifest behavior.

The current branch also needs a rebase onto main and a fresh full CI run.

Amix29:
Published the dependent review follow-up in `e5d1ee0`, stacked on the updated #554 branch.

The offline generation commit now also retires stale DPO output when a later committed run intentionally omits DPO, while preserving manifest-last publication and fail-closed recovery. The regression suite covers the stale-output and interrupted-publication cases.

Local validation before publication: full suite **18,980 passed**, focused offline publication tests, and Ruff. This PR remains draft until #554 lands.

Amix29:
Restacked the two #557-specific commits onto the newly rebased #554 head.

The reviewed blockers are covered on the composed tree:

- oversized JSON integer scores are converted into controlled validation errors;
- an SFT-only generation retires the prior manifest-bound DPO output;
- late-failure, manifest-last publication, invalidation, and recovery behavior is documented and regression-tested;
- the shared `--manifest` option now serves the online #553 and offline #557 publication paths without a duplicate Python parameter.

Validation: Ruff passed and 148 combined #549/#550/#557 Best-of-N tests passed. This PR remains draft until #554 lands; fresh CI is running on the restacked head.

MakazhanAlpamys (changes requested):
Re-review of the restacked head a0acf23 against issues #556/#550: the draft diff is scoped, no tests were deleted, docs/changelog are present, the combined Best-of-N tests passed (148 including #549/#550/#557), Ruff is clean, and the manifest-publication mutation was caught (5 failures). One blocking resume-integrity defect remains: the new --revision value is passed to _load_bon_model but is not included in the 
un_digest inputs around the resume path. A resumed run can therefore switch model revisions while the checkpoint still appears compatible. Include the revision in the digest and add a regression test, then keep the PR draft until #554 is ready and all 13 current-head checks are green.

Amix29:
Addressed the resume-integrity blocker in `faebada` after restacking the branch on the updated #554 head.

The local model `revision` is now part of the Best-of-N `run_digest`. The regression creates a checkpoint with `revision-a`, attempts `--resume` with `revision-b`, and verifies that the mismatch is rejected before `_load_bon_model` is called.

Validation on the composed #554 -> #557 tree:

- Ruff passed
- 150 combined #549/#550/#557/#v0.71.31 Best-of-N tests passed

The PR remains draft until #554 is merge-ready; fresh current-head CI has been triggered.

MakazhanAlpamys:
Integration follow-up: #554 is merged, so this branch can now be rebased onto current `main` (including #586, #568, #585, and #587). Please resolve the remaining dirty state, rerun the full current-head checks, and mark the PR ready when the composed #554-to-#557 chain is mergeable. I will re-review the actual rebased diff and rerun the required mutation before merge.

MakazhanAlpamys (commented):
Follow-up review on the rebased head `23cc461`: the focused tests (41 passed), Ruff, diff check, and the mandatory mutation all pass, and the diff contains the expected docs and changelog updates with no deleted tests.

One P2 remains before merge: the offline path currently accepts `--resume` and `--checkpoint` even though the command documentation says offline workflow does not support resume. In a local check, the command exited successfully and wrote a manifest, so the options are silently ignored. Please either reject these options explicitly or align the documentation and behavior, then keep the PR draft until the full current-head matrix is green.

Amix29:
Thanks @MakazhanAlpamys. I rebased the three #557-specific commits onto current main after #554. The focused suite passes (139 tests), Ruff and diff checks pass, and all 13 GitHub CI jobs are green across Linux, Windows, and macOS. I also verified that model revision changes are included in the resume identity and rejected before model loading. The PR is clean, mergeable, and now ready for re-review.

Amix29:
Thanks @MakazhanAlpamys. Addressed the remaining offline recovery-options finding in 2ed482b. Offline materialization now rejects both --resume and --checkpoint with a controlled exit code 2 before artifact loading or publication; the parameterized regression verifies that neither the SFT output nor the manifest is created. The recovery documentation and changelog now state the enforced behavior. Focused Best-of-N validation passes (141 tests), Ruff is clean, and the PR is back in draft while the fresh full matrix runs.

Amix29:
Final validation for 2ed482b is complete. The focused Best-of-N suite passes (141 tests), the full local suite gained exactly the two new regressions (19,148 passed; the same 18 sandbox-only failures as the pristine-main comparison), and all 13 fresh GitHub CI checks are green across Linux, Windows, and macOS. The branch still contains current main, GitHub reports it clean and mergeable, and the PR is now ready for re-review.

MakazhanAlpamys (changes requested):
Re-review of 2ed482bd6143817e5df5701adf019f5f72af2182: the focused offline suite passed 43 tests, Ruff and diff checks are clean, and the recovery-option mutation was caught. One blocking integrity issue remains: an SFT-only replacement deletes the prior manifest-bound DPO before publishing the new SFT. If publication fails, the old manifest and DPO are lost while a new SFT can remain. Rework the transaction or rollback so the previous committed generation remains intact on failure, and add a regression for that failure path. The current 13 checks are green, but do not merge until this fix is validated on a fresh head.

MakazhanAlpamys:
Merged as `556afe7`. Thank you, @Amix29 — this went through four review rounds and every one of them was answered in code rather than in prose, which is why I could merge it without a fifth.

The round I care most about is the last one. The blocker was that an SFT-only replacement deleted the prior DPO **before** publishing, so a failure part-way lost both the old manifest and the old DPO. `9f266b5` did not patch around that — it moved the deletion *into* the transaction: `paths.py:186-197` validates the removals, `:219-222` moves them to `.soup.backup.*` alongside the outputs, the rollback loop restores them, and they are unlinked only after a full commit. That is the structural fix, not the cheap one.

I verified the fix rather than taking the claim. Two independent mutations on the merged head:

```
# stale DPO unlinked before the transaction instead of within it
FAILED tests/test_issue550_bon_offline.py::test_failed_sft_only_replacement_restores_manifest_bound_dpo
E   FileNotFoundError: [Errno 2] No such file or directory: '...\dpo.jsonl'
1 failed, 43 passed

# manifest dropped from the publication list
6 failed, 38 passed
```

Both die by name. Four more mutations were run during review — recovery-flag rejection removed, verifier comparison removed, manifest-path guard removed, rollback restore removed — and all four were killed as well, six for six with no survivors. `44 passed` on the focused files, `48 passed` across `-k "best_of_n or bon or issue556"`, `ruff` clean, `test_cli_startup_is_light` 34 passed. Zero deleted or weakened test lines in the diff (`git diff origin/main HEAD -- tests/ | grep '^-' | grep -v '^---'` is empty).

Three things worth naming beyond the fix itself:

**The ordering argument is right and you inverted the house rule correctly.** `soup data canary insert` writes the *manifest first* because it carries secrets that must never be orphaned. Here the manifest is a commit marker, so manifest-*last* is the correct inversion — and the old manifest is moved aside before any replacement, so a hard kill mid-transaction leaves **no** manifest rather than a stale one binding new bytes. Fail-closed in both directions.

**You took requested scope and left the rest alone.** `"revision": revision` in the online `run_digest` belongs to #549/#553, not #556, and it is here only because I asked for it on this PR. The test asserts `exit_code == 2` **and** `load_calls == []`, i.e. rejected before the model loads, which is the assertion that actually matters. Nothing else unrelated rode along.

**The path guards match their siblings** rather than inventing a weaker local copy: `enforce_under_cwd_and_no_symlink` on the manifest and on every removal, realpath-normcase distinctness against candidate/judgments/output/emit-pairs, `S_ISREG` before anything is moved aside, and `_regular_sha_and_rows` streaming under `O_NOFOLLOW` instead of buffering the generated files.

All 7 acceptance criteria on #556 are met.

Four non-blocking follow-ups I found and am **not** asking you to fix here — tell me if you want any of them as their own issue:

1. A pre-existing manifest that is unparseable, or whose bound SFT is missing or renamed, makes an SFT-only run exit 1 and publish nothing. That is correctly fail-closed, but `docs/data.md`'s "rerun the exact offline command" does not recover it — the user must delete the manifest by hand — and the `Failed to write output:` prefix is misleading, since nothing was written or attempted.
2. With `--emit-pairs` a foreign manifest at the target path is silently overwritten; without it, hard error. Defensible, undocumented.
3. `test_failed_dpo_replacement_restores_previous_generation` reruns with identical inputs, so the manifest bytes do not change and it cannot detect manifest-before-data ordering — a manifest-committed-first mutation **passes** it, and only its SFT-only sibling caught that. Perturbing judgments there too would cover the ordering guarantee twice.
4. `atomic_write_bytes_group` still has no direct test file of its own, so the new `removals` validation branches are unreachable from the CLI and untested.

#559 will need a rebase onto current `main` now that this has landed.

Issue #550, opened 2026-08-25
Best-of-N cannot separate candidate generation from offline verified judgment


Amix29:
## Problem

Audited on current `main` at `6c13c44`.

`soup data best-of-n` currently requires `--judge <URL>` at CLI parsing time, including for `--plan-only`:

```text
Error: Missing option '--judge'.
```

The command then performs sampling and pointwise network/local-server judging in one loop. There is no option to export all candidates, stop, have an offline verifier review them, and import the verified decisions later. Searches of the CLI/docs found no candidate or judgment import contract.

That blocks air-gapped and artifact-mediated workflows where candidate generation is local but judgment is performed asynchronously by a human, deterministic program, CI process, or Codex without exposing an API endpoint.

The current outputs are not a substitute for the missing boundary:

- the SFT row retains scores but not the ordered candidate texts or their hashes;
- local sampling records no base model/revision/seed/temperature/token limit;
- emitted DPO rows contain only `prompt`, `chosen`, and `rejected`, with no Best-of-N provenance.

Related findings from the same audit are tracked separately so this feature does not hide
smaller correctness fixes: non-finite score handling in #547, silent prompt-row loss in #548,
and durable recovery after late failures in #549.

## Proposed two-phase contract

1. Sampling phase exports one immutable candidate group per prompt (`prompt_id`, ordered candidates, sampling specification, and content digests) without constructing a judge.
2. Materialization phase imports that candidate artifact plus offline judgments, validates their binding, and emits SFT and optional DPO rows without loading the sampler or making network calls.
3. The existing online path may internally compose the same two phases so both paths share validation and provenance.

## Acceptance criteria

- [ ] Candidate generation can run without `--judge` and preserves all N candidates.
- [ ] Offline materialization performs no network call and does not load the sampling model.
- [ ] Every judgment binds to the prompt and exact ordered candidate set through stable ids/digests.
- [ ] Missing, duplicate, out-of-range, non-finite, or digest-mismatched judgments fail closed before final output publication.
- [ ] SFT and DPO outputs retain an artifact/judgment digest plus sampler and judge/verifier provenance; secrets and user-specific paths are excluded.
- [ ] Replaying identical candidate and judgment artifacts produces byte-stable training rows.
- [ ] The existing one-command online workflow remains supported.

I am taking this issue.

MakazhanAlpamys:
[...]

PR #559, opened 2026-08-25
fix(data): stream and resume two-phase Best-of-N

### LINKED ISSUE #555: Two-phase Best-of-N buffers whole artifacts and cannot resume candidate export
## Problem

Found while integrating #549 and #550. This issue depends on #554.

The two-phase Best-of-N path accumulates every completed candidate group in a Python list and publishes the candidate artifact only after all prompts finish. A late sampler failure therefore discards all completed groups, and `--resume` is unavailable for `--export-candidates`.

A bounded measurement with 1,000 prompts, 8 candidates per prompt, and approximately 1 KiB per candidate produced a 9,422,872-byte candidate artifact while the current build/load/materialize process reached 95,404,032 bytes of maximum RSS. The supported 100,000 prompts x 64 candidates can therefore produce multi-gigabyte artifacts and several in-memory copies. `load_candidate_artifact`, `load_judgments`, and `materialize_rows` currently read/materialize the complete inputs as lists.

A two-prompt reproduction where sampling fails on prompt 2 confirms that prompt 1 is sampled successfully but no candidate artifact or recovery record remains.

## Expected

The two-phase path should be durable and memory-bounded without weakening content binding or deterministic output.

## Acceptance criteria

- [ ] Candidate export durably checkpoints each completed prompt group.
- [ ] `--resume` reuses a matching sequential prefix exactly once and rejects prompt/configuration drift before sampling.
- [ ] Final candidate publication remains atomic and content-addressed.
- [ ] Candidate and judgment validation/materialization operate incrementally with memory bounded independently of total artifact size.
- [ ] Missing, duplicate, reordered, digest-mismatched, and truncated records still fail closed before a final manifest is committed.
- [ ] A late-failure regression proves completed groups are not sampled again.
- [ ] A large synthetic regression demonstrates bounded-memory iteration without downloading or loading a model.

I am taking this issue.

Amix29:
## Summary

Closes #555.

- stream candidate export through a durable, per-prompt checkpoint and resume from the first incomplete group;
- authenticate resume against prompt text, source-line provenance, sampler settings, provider endpoint identity, and exact local-model file names/sizes/content;
- keep private endpoint/model identities in the recovery journal instead of changing or leaking them through the public `candidates.v1` artifact;
- discard only an incomplete final checkpoint record, preserve deterministic per-prompt RNG on local resume, and reject changed inputs before loading the model;
- validate candidate/judgment artifacts through a bounded-memory disk index and stage SFT/DPO incrementally;
- publish streamed SFT, optional DPO retirement, and the manifest as one rollback-protected generation, with the manifest committed last;
- reject collisions among prompt, candidate, checkpoint, output, DPO, judgment, and manifest paths before sampling or publication.

## Rebase and scope

This branch is rebased onto current `main` after #557 merged. The PR diff now contains only the #555 streaming/resume implementation, its tests, documentation, and `555.fixed.md`; `556.fixed.md`, the #557 offline test additions, and the shared `paths.py` removal transaction are supplied by `main` and no longer appear in this PR.

Two non-blocking #557 follow-ups that directly exercise this path were folded in:

- direct coverage for the shared atomic-group removal validation branches;
- changed-generation rollback coverage plus an explicit SFT -> DPO -> manifest publication-order assertion.

## Validation

- `mypy src/soup_cli/utils/best_of_n_stream.py` — passed
- focused Best-of-N and changelog suite — **77 passed**
- full local suite on the rebased implementation — **19,164 passed, 74 skipped, 5 deselected, 18 failed; 82.64% coverage**
- the same exact 18 failures reproduce **18/18** on pristine `main` at `0d98183`; they are restricted-environment failures, not branch regressions:
  - `test_doctor.py::test_doctor_nccl_mocked_success` (process/socket or local Torch environment)
  - three `test_issue144_gguf_export.py` cases (registry database cannot be opened)
  - two `test_issue296_network_transport.py::TestSseEndToEnd` cases (loopback bind denied)
  - one `test_part_f_hardening.py` and two `test_rlvr.py` code-exec cases (subprocess sandbox denied)
  - two `test_v0402_part_b.py` data-registry cases (home-directory write denied)
  - two `test_v0537.py` recipe code-node cases (sandbox execution denied)
  - `test_v0540.py::TestCLI::test_explain_no_prior_verdict` (home-directory unlink denied)
  - `test_v07110.py::TestSteeringArtifactRoundtrip::test_resolve_unknown_raises` (registry database cannot be opened)
  - three `test_v07118.py` sandbox-eval cases (`sandbox-exec` denied)

Mutation checks:

- removing the private sampler identity from the checkpoint made the endpoint-drift, local-content-drift, and identity-privacy regressions fail (**3 failures**);
- committing the manifest before the streamed data made `test_streamed_publication_commits_manifest_last` fail by name (**1 failure**).

The PR remains draft until the fresh GitHub Actions matrix for this rebased head is green.

### DIFF
--- a/changelog.d/0.73.3/555.fixed.md
+++ b/changelog.d/0.73.3/555.fixed.md
@@ -1,3 +1,5 @@
 - Fixed #555: two-phase Best-of-N candidate export now resumes from durable,
   authenticated per-prompt checkpoints, while offline validation and output
-  materialization stream through bounded-memory disk staging.
+  materialization stream through bounded-memory disk staging. Resume binds
+  prompt source lines and exact local-model content, and streamed SFT/DPO
+  publication rolls back as one set on failure.
--- a/changelog.d/0.73.3/555.fixed.md
+++ b/changelog.d/0.73.3/555.fixed.md
@@ -2,4 +2,6 @@
   authenticated per-prompt checkpoints, while offline validation and output
   materialization stream through bounded-memory disk staging. Resume binds
   prompt source lines and exact local-model content, and streamed SFT/DPO
-  publication rolls back as one set on failure.
+  publication rolls back as one set on failure. Candidate export now seeds each
+  prompt independently so resumed runs reproduce the same candidates; this
+  changes the output of existing `--export-candidates --seed N` runs.
--- a/changelog.d/0.73.3/555.fixed.md
+++ b/changelog.d/0.73.3/555.fixed.md
@@ -1,0 +1,3 @@
+- Fixed #555: two-phase Best-of-N candidate export now resumes from durable,
+  authenticated per-prompt checkpoints, while offline validation and output
+  materialization stream through bounded-memory disk staging.
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -302,7 +302,7 @@ soup build <manifest.yaml> [--dry-run] [--output-dir <dir>]  dbt-for-SFT DAG: va
 soup expect <data.jsonl> <suite.yaml>         Expectations suite: PII / token-length / refusal / judge (v0.69.0)
 soup data gen-magpie --base <m> --provider ollama|vllm --target N --output <jsonl> [--base-url <url>] [--quality-filter]  Magpie synthetic generator — live (v0.69.0; live v0.71.6)
 soup data best-of-n (--base <m> | --provider ollama|vllm --model <m> [--base-url <url>]) --prompts <jsonl> --n 8 --judge <url> -o <sft.jsonl> [--emit-pairs <dpo.jsonl>] [--resume] [--checkpoint <journal.jsonl>] [--manifest <manifest.json>]  Best-of-N rejection sampling with durable per-prompt recovery and manifest-last publication
-soup data best-of-n (--base <m> [--revision <rev>] | --provider ollama|vllm --model <m>) --prompts <jsonl> --n 8 --export-candidates <jsonl>  Sampling-only phase; no judge is constructed
+soup data best-of-n (--base <m> [--revision <rev>] | --provider ollama|vllm --model <m>) --prompts <jsonl> --n 8 --export-candidates <jsonl> [--checkpoint <jsonl>] [--resume]  Resumable sampling-only phase; no judge is constructed
 soup data best-of-n --candidate-artifact <jsonl> --judgments <jsonl> -o <sft.jsonl> [--emit-pairs <dpo.jsonl>] [--manifest <json>]  Validate offline judgments and materialize byte-stable training rows; a final manifest binds the requested output set
 soup data evolve --input <seeds.jsonl> --provider ollama|vllm --model <m> --strategy depth|breadth --rounds N -o <jsonl>  Evol-Instruct (WizardLM) instruction evolution (v0.71.31)
 soup data persona-mix --prompts <jsonl> --n N --output <jsonl>  Persona-Hub diversity sampler (v0.69.0)
--- a/docs/data.md
+++ b/docs/data.md
@@ -260,6 +260,13 @@ lists. They must never discover training inputs by globbing neighboring JSONL
 files: an unlisted sidecar, including an older DPO stored elsewhere, is not part
 of the committed generation.
 
+Candidate export durably checkpoints each completed prompt group at
+`<artifact>.checkpoint.jsonl`. If sampling stops, rerun the same command with
+`--resume`; Soup authenticates the checkpoint against the prompts and sampler
+before continuing at the first incomplete group. Candidate and judgment inputs
+are validated through a temporary disk index, and final SFT/DPO files are staged
+incrementally, so memory does not grow with the complete artifact size.
+
 ### Custom Transforms
 
 Use a dotted-path string (`module.path:function_name`) as the ``transform``
--- a/docs/data.md
+++ b/docs/data.md
@@ -266,6 +266,13 @@ Candidate export durably checkpoints each completed prompt group at
 before continuing at the first incomplete group. Candidate and judgment inputs
 are validated through a temporary disk index, and final SFT/DPO files are staged
 incrementally, so memory does not grow with the complete artifact size.
+For a local model directory, the checkpoint binds the exact regular-file names,
+sizes, and contents through a privacy-safe fingerprint; replacing weights at the
+same path therefore invalidates resume before the model is loaded. Prompt source
+lines and provider endpoints are bound as well without exposing private paths or
+URLs. Streamed SFT/DPO replacements are committed as one rollback-protected set,
+and an SFT-only replacement retires a prior manifest-bound DPO in that same
+transaction.
 
 ### Custom Transforms
 
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -3035,7 +3035,9 @@ def best_of_n(
         "", "--emit-pairs", help="Also write winner/loser DPO pairs to this JSONL"
     ),
     checkpoint: str = typer.Option(
-        "", "--checkpoint", help="Recovery journal (default: <output>.checkpoint.jsonl)"
+        "",
+        "--checkpoint",
+        help="Recovery journal (default: <output-or-artifact>.checkpoint.jsonl)",
     ),
     manifest: str = typer.Option(
         "", "--manifest", help="Commit manifest (default: <output>.manifest.json)"
@@ -3079,7 +3081,6 @@ def best_of_n(
     from soup_cli.utils.magpie import make_magpie_generate_fn
     from soup_cli.utils.paths import (
         atomic_write_bytes,
-        atomic_write_bytes_group,
         enforce_under_cwd_and_no_symlink,
     )
     from soup_cli.utils.trust_remote import (
@@ -3104,6 +3105,8 @@ def best_of_n(
         console.print("[red]offline materialization and --export-candidates are exclusive[/]")
         raise typer.Exit(2)
     if offline_mode:
+        from soup_cli.utils import best_of_n_stream as bon_stream
+
         if not candidate_artifact or not judgments:
             console.print("[red]provide both --candidate-artifact and --judgments[/]")
             raise typer.Exit(2)
@@ -3125,6 +3128,8 @@ def best_of_n(
                 revision,
                 trust_remote_code,
                 seed != 0,
+                checkpoint,
+                resume,
                 n != 8,
                 temperature != 1.0,
                 max_new_tokens != 256,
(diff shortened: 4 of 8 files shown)


Amix29:
Fixed the audit findings in 9898f40:

- local export resume now uses a stable per-prompt seed;
- incomplete final checkpoint fragments are truncated at the last committed JSONL newline;
- private local model identities and provider endpoints are bound by SHA-256 fingerprints without exposing their paths/URLs;
- offline materialization validates and writes each judgment in one streaming pass instead of validating the full set twice.

Validation: 26 focused two-phase/offline tests passed, including RNG-equivalent resume, partial-tail recovery, local-model drift rejection, provider-endpoint drift rejection, and a one-pass assertion. Ruff is clean.

MakazhanAlpamys:
Thanks for the draft implementation. I reviewed issue #555 first, the full diff, test deletions, docs/changelog, and the relationship with the offline Best-of-N work.

Before this is ready for merge:
- Rebase onto current main; the branch is based on 6c13c44 and conflicts with the current tree in docs/data.md and src/soup_cli/commands/data.py.
- Add a model-content or immutable-snapshot fingerprint to resume validation. The current fingerprint uses the resolved path and revision, so replacing model files in the same directory can resume with different content.
- Make SFT/DPO publication atomic as a set, or add rollback/versioned publication. A failure between the two replacements can leave an inconsistent pair even though manifest-gated readers fail closed.
- Consider renaming changelog.d/0.73.3/555.fixed.md if the project convention requires the PR number rather than the linked issue number.

Verification on the reviewed head 9898f401183a0d68b88e7690789f34537a4d13cd:
- 26 focused streaming/offline Best-of-N tests passed.
- 239 related tests passed with 5 skips.
- 17 changelog validation tests passed.
- Full Ruff passed.
- The mutation changing completed += 1 to completed += 0 failed test_late_candidate_failure_resumes_without_replaying_completed_groups, so the resume regression test is effective.

No tests were deleted, and the documented streaming/checkpoint/changelog additions are present. Please refresh the branch and checks after addressing the blockers.

Amix29:
Resolved the conflicts and addressed every blocking finding on `f5ea66b`.

- Restacked only the three #559 commits on the current #557 head, preserving the #554/#557 dependency chain and its online checkpoint, offline manifest, stale-DPO retirement, exact source-line provenance, and controlled temperature validation.
- Local export resume now fingerprints exact regular-file names, sizes, and bytes. Replacing model weights at the same path is rejected before model loading; the fingerprint remains privacy-safe.
- Prompt source lines are included in the checkpoint run digest and revalidated for every completed group.
- Streamed SFT/DPO outputs now publish as one rollback-protected set. A failure on the second replacement restores both previous outputs, and SFT-only regeneration retires a previous manifest-bound DPO in the same transaction.
- `changelog.d/0.73.3/555.fixed.md` intentionally follows the linked-issue convention already used by the dependent fragments for #550 and #556.

Validation on the new head:
- full suite: 19,113 passed, 82 skipped, 5 deselected; 82.70% coverage
- related Best-of-N/data/changelog suite: 178 passed
- full Ruff and `git diff --check`: clean

The PR remains a draft while #554 and #557 are open. GitHub now reports it mergeable, and the refreshed CI matrix is running.

MakazhanAlpamys:
Integration follow-up: #554 is merged, but #557 remains open and this PR is still a draft with a dirty merge state. After #557 is rebased and reviewed, please rebase #559 onto the resulting current `main`, rerun the full matrix, and mark it ready. I will review the composed stream/resume diff and rerun the required mutation before merge.

MakazhanAlpamys (changes requested):
Re-review of f5ea66ba6851e6d22ca3ce554d88b9ba83675b93: the focused Best-of-N tests passed 52 cases, Ruff and diff checks are clean, and the late-failure mutation was caught. One blocking integrity defect remains: when --prompts and --export-candidates name the same path, validation does not reject the collision and os.replace overwrites the source prompts file. Add path-distinct validation and a regression test. The PR is also currently DIRTY/CONFLICTING and draft; rebase onto current main and refresh the exact gh pr checks before requesting re-review.

Amix29:
Implemented the requested path-collision fix and refreshed this draft on top of the corrected #557 head.

Changes on `9ed9da9`:
- candidate export now requires the prompt source, candidate artifact, and checkpoint paths to be pairwise distinct;
- the regression passes `--prompts` and `--export-candidates` as the same file and proves the original prompt bytes survive, no sampling occurs, and no checkpoint is created;
- the streamed offline publisher now carries SFT, optional DPO retirement, and the manifest in one rollback-protected transaction, preserving #557's all-old-or-all-new guarantee after the rebase.

Validation:
- focused Best-of-N tests: 37 passed
- full local suite: 19,165 passed, 88 skipped, 5 deselected
- fresh GitHub Actions matrix: 13/13 checks passed, including Windows 3.10/3.11/3.12

The PR is clean and conflict-free. I am keeping it in draft until #557 is merged, as requested; I will then rebase it onto the resulting `main` before marking it ready.

MakazhanAlpamys:
@Amix29 — #557 is merged (`556afe7`), so this is unblocked. Please rebase onto current `main`.

Right now the branch still carries #557's content — `changelog.d/0.73.3/556.fixed.md`, the full `tests/test_issue550_bon_offline.py` (+293), and the `paths.py` `removals=` work all appear in this diff — so I cannot tell your #555 changes from the ones that just landed. After the rebase the diff should be `best_of_n_stream.py`, the streaming/resume half of `best_of_n_artifact.py` and `data.py`, `tests/test_issue555_bon_stream_resume.py`, `555.fixed.md`, and the docs. That is what I want to review.

Two small things while you are rebasing:

- **The PR body has a raw pytest progress dump pasted into the Validation section** — several hundred lines of dots, including an `F` and a run of `s` markers with no summary line. Please replace it with the final counts (and, if that `F` was a real failure, say which test and whether it reproduces on `main`).
- **Both `555.fixed.md` and `556.fixed.md` are present.** After the rebase only `555.fixed.md` should remain; `556.fixed.md` shipped with #557.

Four non-blocking follow-ups from the #557 review are listed there — two of them (`atomic_write_bytes_group` still having no direct test file, and `test_failed_dpo_replacement_restores_previous_generation` being unable to detect manifest-before-data ordering because it reruns with identical inputs) sit in code this PR touches heavily. If they are cheap to fold in here, do; if not, leave them and I will file them.

Ping me when the rebase is up and I will review the streaming and resume design properly — it is the largest of the three Best-of-N slices and I want to spend real time on the checkpoint authentication and the rollback path.

Amix29:
Thanks @MakazhanAlpamys — #559 is now rebased onto current main at `579bbbe`, and the diff is limited to the #555 streaming/resume slice you outlined. The raw pytest dump has been replaced with final counts and an exact pristine-main comparison.

Two cheap #557 follow-ups are included: direct `atomic_write_bytes_group` removal-validation coverage, and changed-generation rollback plus an explicit SFT -> DPO -> manifest ordering regression.

Validation on the final head:

- fresh GitHub Actions: 13/13 green across Linux, Windows, and macOS
- Ruff clean
- 77 focused Best-of-N/changelog tests passed
- the new module passes mypy
- full local run: 19,164 passed; the 18 restricted-environment failures reproduce 18/18 on pristine main
- identity-removal mutation: 3 regressions failed
- manifest-first mutation: the ordering regression failed

The PR is clean, mergeable, and ready for the streaming/checkpoint-authentication review.

MakazhanAlpamys:
Thank you, @Amix29 — the rebase is exactly what I asked for and the diff is now legible as the #555 slice. This is merge-quality. **One sentence in the changelog and I merge it.**

## What I verified myself

```
git merge-base --is-ancestor origin/main 579bbbe1   -> REBASE_OK
git diff origin/main...579bbbe1 -- tests/ | grep '^-'  -> empty (zero deleted test lines)
pytest test_issue555 + test_issue550 + test_issue549   -> 60 passed
```

I ran three mutations of my own, and the third is the one I most wanted to see:

```
# exactly-once resume: range(completed, ...) -> range(0, ...)
FAILED test_late_candidate_failure_resumes_without_replaying_completed_groups
FAILED test_local_resume_reuses_the_same_prompt_seed
2 failed, 14 passed

# local model content fingerprint dropped from the resume binding
FAILED test_resume_rejects_replaced_content_at_the_same_local_model_path
1 failed, 15 passed

# rollback restore loop neutered in best_of_n_stream.publish_staged_datasets
FAILED test_streamed_publication_restores_changed_generation_on_dpo_failure
FAILED test_failed_dpo_replacement_restores_previous_generation
FAILED test_failed_sft_only_replacement_restores_manifest_bound_dpo
FAILED test_dpo_write_failure_rolls_back_the_complete_publication[True]
4 failed, 56 passed
```

Three of those four failures come from `tests/test_issue550_bon_offline.py` — #557's own suite, now on `main`. That is the evidence I needed that a 267-line rewrite of `best_of_n_artifact.py` and a brand-new 647-line transaction did not quietly drop the guarantees I merged this morning. Manifest-last and stale-DPO-inside-the-transaction both survive the rewrite and are mutation-confirmed.

I also went through the deletions line by line rather than trusting the diffstat. The 104 removed lines in `best_of_n_artifact.py` are all replaced in place — `validate_candidate_group` / `validate_judgment` / `materialize_group` preserve every check verbatim and **add** four (`expected_prompt`, `expected_source_line`, `isinstance(row, dict)`, explicit `prompt_id` equality). The two relaxed guards in `data.py` are relaxed deliberately and correctly: dropping `checkpoint`/`resume` from the offline refusal *is* the feature.

The streaming claim holds. `grep` for `load_candidate_artifact|load_judgments|materialize_rows|candidate_artifact_text|stable_jsonl` across `src/` returns only their own definitions — no live path calls a whole-file materialiser any more. The SQLite index is `mkdtemp`-unique so concurrent runs cannot collide, cwd-contained, and removed on both success and failure. And the security posture in `best_of_n_stream.py` matches its siblings throughout: `enforce_under_cwd_and_no_symlink` + `O_NOFOLLOW` + `S_ISREG(fstat)` on every read, `O_CREAT|O_EXCL` at `0o600` for checkpoint creation, staging via `mkstemp` in the destination's own parent, containment re-checked immediately before each `os.replace`, and a `normcase(realpath(...))` identity set so no two publication targets can alias.

## The one thing to fix

**The export phase's seeding changed and nothing says so.** `main` seeded once, globally:

```
src/soup_cli/commands/data.py:3427   torch.manual_seed(seed)
```

and this PR seeds per prompt:

```
local_torch.manual_seed(bon_checkpoint.prompt_seed(seed, index))
```

I checked and this is *correct*: it is required for resume determinism, and `main:3488` already seeds per-prompt on the online path, so this makes the two phases consistent rather than diverging them. But it means `soup data best-of-n --export-candidates --seed 42` now produces **different candidate text** than v0.73.3 did, and neither the fragment nor `docs/data.md` mentions it (`grep -ci seed` on the fragment returns 0). Anyone pinning a seed for reproducibility gets silently different data.

Please add one sentence to `changelog.d/0.73.3/555.fixed.md` saying the export phase now seeds per prompt so a resumed run reproduces the same candidates, and that this changes the output of an existing `--export-candidates --seed N` run. That is the whole ask; push it and I will merge.

## Non-blocking, for a follow-up rather than this PR

**Two guards have no coverage.** I am recording these because both are real and neither is a live defect:

- Re-buffering the offline join survives the suite. Wrapping the `iter_rows` cursor in `list(...)` — the literal #555 complaint, reintroduced — leaves `41 passed`. `test_offline_large_artifacts_bypass_whole_file_materializers` proves the *named* buffering functions are not called and that `iter_rows` runs exactly once; it does not assert memory, so a future regression inside `iter_rows` ships silently. RSS assertions are flaky in CI, so I am not asking for one, but you should know the answer is "no test would notice".
- Deleting the `candidate checkpoint is incomplete` guard also leaves `41 passed`. I probed the blast radius: with it removed, a 2-of-3 checkpoint **publishes** an artifact whose header claims `prompt_count = 3` over 2 group lines, and the downstream consumer still refuses it (`candidate artifact header does not match its groups`). So it is defence-in-depth behind a working outer check, reachable only via concurrent truncation between prepare and publish. One test, cheap, whenever you are next in there.

**`atomic_write_bytes_group` is now orphaned in production.** After this PR, `grep` finds it only in `paths.py` and in the new test file — the transaction primitive I merged with #557 is no longer called by any live path, and the guarantee is re-implemented inside `publish_staged_datasets`. That is two implementations of one transaction, which is exactly the duplication this repo has been bitten by three times (#372, #392, #424). Same for `load_candidate_artifact`, `load_judgments`, `materialize_rows`, `candidate_artifact_text` and `stable_jsonl`, all now test-only. Worth a consolidation issue — tell me if you would rather file it or have me do it.

Cosmetic: `checkpoint_path` is computed twice in export mode (`data.py:3280` and `:3437`) from the identical expression; the second is dead.

On the changelog naming — the fragment is `555.fixed.md`, i.e. the issue number, where `CLAUDE.md` documents `<PR>.<category>.md`. You are consistent with every shipped sibling in that directory (`549.fixed.md`, `550.added.md`, `556.fixed.md`), so the documentation is what is out of date, not your file. Leave it as is; I will fix the doc.

Amix29:
@MakazhanAlpamys — the requested changelog disclosure is now up in `3af1c68`.

It states both parts explicitly: candidate export now seeds per prompt so resumed runs reproduce the same candidates, and this changes the output of an existing `--export-candidates --seed N` run.

The diff is limited to `changelog.d/0.73.3/555.fixed.md`; the focused #487/#555 tests pass (33 passed). Thank you for the thorough review and mutation checks.

MakazhanAlpamys:
Merged as `03ac115`. Thank you, @Amix29 — the disclosure is exactly what I asked for, and it is worth saying that the only change since the head I reviewed was that fragment:

```
git diff 579bbbe1 3af1c68e --stat
 changelog.d/0.73.3/555.fixed.md | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)
```

so the code and tests I mutation-tested are byte-identical to what merged. Re-ran anyway on the final head: `60 passed` across `test_issue555` + `test_issue550` + `test_issue549`, all 13 checks green.

**The thing I most wanted to establish, and did:** this PR rewrites 267 lines of `best_of_n_artifact.py` and adds a 647-line transaction on top of code I merged the same week, so the real risk was silently undoing #557. Neutering the rollback restore loop fails four tests — and three of them are `test_failed_dpo_replacement_restores_previous_generation`, `test_failed_sft_only_replacement_restores_manifest_bound_dpo` and `test_dpo_write_failure_rolls_back_the_complete_publication[True]`, i.e. **#557's own suite, now on `main`**. That is what makes the rewrite reviewable rather than a matter of trust. Manifest-last and stale-DPO-inside-the-transaction both survive, mutation-confirmed.

My other two mutations: replaying completed groups (`range(completed, ...)` -> `range(0, ...)`) fails `test_late_candidate_failure_resumes_without_replaying_completed_groups` and `test_local_resume_reuses_the_same_prompt_seed`; dropping the local-model content fingerprint from the resume binding fails `test_resume_rejects_replaced_content_at_the_same_local_model_path`. Each core guarantee has a test that fails when the code is broken.

The content fingerprint is the design decision I'd single out. Binding a resume to path plus revision would have looked sufficient and would have let someone swap weights in place at the same path and get one artifact silently built from two different models. Walking the directory with `followlinks=False`, symlinked-dir rejection, `O_NOFOLLOW`, `S_ISREG` and a re-`fstat` to catch mid-read mutation is the version that actually holds.

I also went through the 104 removed lines in `best_of_n_artifact.py` and the 118 in `data.py` individually rather than trusting the diffstat: every check in `validate_candidate_group` / `validate_judgment` / `materialize_group` is preserved verbatim and four are **added**, and the two relaxed guards in `data.py` are relaxed correctly — dropping `checkpoint`/`resume` from the offline refusal *is* the feature. Zero deleted test lines anywhere. Security posture in the new module matches its siblings throughout.

You are in `CONTRIBUTORS.md` as of `6236763`.

The three follow-ups from my review stand, none of them blocking and none needing action from you unless you want them:

- re-buffering the offline join (`list(...)` around the `iter_rows` cursor) survives the suite — the test proves the *named* buffering functions are not called, not that memory is bounded;
- deleting the `candidate checkpoint is incomplete` guard also survives; I probed it and the downstream consumer still refuses the artifact, so it is defence-in-depth behind a working outer check;
- `atomic_write_bytes_group` is now orphaned in production — `grep` finds it only in `paths.py` and a test file, so #557's transaction primitive and `publish_staged_datasets` are two implementations of one guarantee.

Tell me if you would rather file the consolidation issue yourself or have me do it; I'd rather it carried your reasoning than mine.

PR #560, opened 2026-08-25
fix(sft): only check the most recent value of each finite-state metric

### LINKED ISSUE #546: NF4 and non-NF4 SFT both produce transient grad_norm=nan on Pascal (sm_61) — _assert_finite_training_state then refuses to save, no bypass
## Environment
- GPU: NVIDIA GeForce GTX 1050 Ti — Pascal, **compute capability 6.1**
- OS: Ubuntu 26.04 LTS (resolute)
- Python: 3.12.13
- torch: 2.13.0+cu126 (cu130 wheel has no sm_61 kernels at all; had to reinstall against the cu126 index to get any CUDA support)
- transformers: 5.15.1, peft: 0.20.0, trl: 0.29.1, accelerate: 1.14.0, bitsandbytes: 0.50.1
- `python -m bitsandbytes` reports: `WARNING: Compute capability < 7.5 detected! Only slow 8-bit matmul is supported for your GPU!`

## Summary
Every SFT run we've tried on this Pascal card logs a transient `grad_norm=nan` in the first few steps, which trips `_assert_finite_training_state` (soup_cli/trainer/sft.py) and aborts the save with:
```
RuntimeError: non-finite training metric grad_norm=nan at step N; refusing to save the final model because its weights may be corrupted
```
This happens **with and without NF4 quantization**, both with and without layer streaming, and is **deterministic** (identical step number and identical logged values across reruns with different hyperparameters). The only config that has completed cleanly is a non-quantized, non-streamed 1.1B model.

This looks like normal fp16 GradScaler loss-scale warm-up (a transient skipped step is expected AMP behavior, not corruption) that `_assert_finite_training_state` cannot distinguish from real corruption, because it scans the *entire* log history for any non-finite metric rather than checking the final saved weights only. (There is a separate, more meaningful check right below it that does scan the actual trainable parameters with `torch.isfinite` — that one never got a chance to run in our failures, since the log-history scan raises first.)

## Repro 1 — 8B layer streaming + NF4, lr=5e-4
```yaml
base: NousResearch/Meta-Llama-3.1-8B-Instruct
task: sft
backend: transformers
training:
  quantization: 4bit
  stream_layers: true
  stream_source: auto
  stream_buffers: 2
  lr: 5e-4
  batch_size: 1
  lora: {r: 8, alpha: 16, target_modules: [q_proj, v_proj]}
data: {max_length: 256}
```
Pre-flight: `peak VRAM ~2.95 GB at batch 1 x seq 256`, free VRAM 4.17 GB. Trains all 9/9 steps, **measured peak 2.2/3.9 GB** — fits comfortably. Fails at the end:
```
grad_norm: 0.02159 (step1) -> nan (step2) -> nan (step3) -> 3.501 (step4) -> ... -> train completes 9/9
Error: non-finite training metric grad_norm=nan at step 2
```

## Repro 2 — same config, lr=1e-4 (rules out "LR too high")
Identical failure, **same step (2)**.

## Repro 3 — same config, lr=1e-4, stream_buffers=8 instead of 2 (rules out a streaming race/timing condition)
Loss and grad_norm sequence **byte-identical** to Repro 2, still fails at step 2. Since Soup's own pre-flight already classifies this workload as compute-bound (not transfer-bound), and 4x more double-buffering changes nothing, this rules out a buffer/timing race as the cause.

## Repro 4 — 3B resident (no streaming), NF4, lr=1e-4
```yaml
base: Qwen/Qwen2.5-3B-Instruct
task: sft
backend: transformers
training:
  quantization: 4bit
  lr: 1e-4
  batch_size: 1
  lora: {r: 16, alpha: 32, target_modules: [q_proj, v_proj]}
data: {max_length: 512}
```
Fails at **step 3** this time (`grad_norm: 14.32 -> 11.29 -> nan -> 12.82 -> ...`). Also logged a transient allocator warning mid-run (`memory allocation failed with OOM ... free: 43450368, total: 4227465216`) that appears to have been recovered from automatically.

## Repro 5 — 3B, layer streaming, **quantization: none** (rules out bitsandbytes/NF4 as the sole cause)
```yaml
base: Qwen/Qwen2.5-3B-Instruct
task: sft
backend: transformers
training:
  quantization: none
  stream_layers: true
  stream_source: auto
  stream_buffers: 2
  lr: 1e-4
data: {max_length: 256}
```
No bitsandbytes kernel involved at all here — pure fp16 tensor streaming. **Still fails**, and worse: 4 consecutive nan steps (1-4) before recovering at step 5, then aborts with the same error at step 1.

## What *did* work cleanly
TinyLlama-1.1B, resident, `quantization: none`, no streaming (the bundled `examples/configs/sft_basic.yaml`) — completed 3/3 steps with no non-finite metric ever logged, GPU 2.1/3.9 GB.

## Hypothesis
This card is forced onto the fp16 store/compute path (the pre-Ampere fix in #385/#387), which needs a dynamic GradScaler. A transient inf/nan gradient during the scaler's normal warm-up is expected AMP behavior — PyTorch's GradScaler is *designed* to detect and skip exactly this without corrupting the model. `_assert_finite_training_state`'s log-history scan can't tell that apart from genuine corruption, so it fires on every fp16 run that happens to hit a skip in its first few steps — which toy/short runs almost always will. We were not able to find any config flag to relax or disable this check.

## Ask
- Could the check look at the *actual saved parameters* (the `torch.isfinite(parameter...)` check that already exists a few lines below) instead of, or in addition to, the raw log history — so a run that self-corrected isn't treated the same as one that didn't?
- Alternatively, a documented way to tolerate N non-finite steps (e.g., only during `warmup_ratio`) before treating it as fatal?

Happy to provide the full crash bundles / run configs if useful.

AmirF194:
## What does this PR do?

`_assert_finite_training_state` (`src/soup_cli/trainer/sft.py`, added in #535) scans the
whole `log_history` and raises on the first non-finite metric it finds at any step. A run
whose GradScaler warm-up produced a transient nan that self-corrected is refused
identically to a genuinely corrupted one, and the loop never reaches the
`torch.isfinite` parameter check a few lines below it, which is the check that actually
reflects the final saved weights.

Fix: check only the most recently logged value of each metric in `_FINITE_TRAINING_METRICS`,
scanning `log_history` in reverse and stopping at the first entry that carries it. This
matches the function's own docstring ("refuse the final save... because its weights may be
corrupted") without changing anything for a run that is still non-finite at its last logged
step, which still raises exactly as before.

## Type of change

- [x] Bug fix

## How I verified it

- New regression tests in `tests/test_issue546_transient_finite_metric.py`: a transient nan
  that recovers by the next entry is now accepted; a metric still non-finite at the last
  entry still raises; two independently-tracked metrics (one recovered, one not) are each
  checked on their own trailing value.
- Ran both the new tests and the existing `test_issue532_supervised_token_guard.py` tests
  that exercise this function against unmodified `main` and against this branch, in a
  `python:3.11-slim` container: the two new discriminating tests fail on `main` and pass on
  the branch, and none of the pre-existing tests for this function changed outcome.
- `ruff check src/soup_cli/ tests/` passes on the changed files.
- What I did not check: the full CI matrix (macos/windows, real `torch`/`transformers`
  install). This environment is disk-constrained, so I verified only the pure-Python
  `log_history` path, which this fix and both call-site arguments are limited to.

Fixes #546.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **`_assert_finite_training_state` no longer refuses a run over a self-corrected
+  transient metric (#546 by @AmirF194 in #560).** The log_history scan now checks only
+  the most recently logged value of each metric instead of raising on the first
+  non-finite value found at any step, so a GradScaler warm-up nan that recovers no
+  longer blocks the final save.
 - **SmolVLM/Idefics3 vision SFT now reaches real training batches (#302 by
   @Amix29 in #488).** Soup keeps LLaVA messages and PIL images together until
   collation, converts legacy `<image>` markers to structured multimodal content,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,6 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
-- **`_assert_finite_training_state` no longer refuses a run over a self-corrected
-  transient metric (#546 by @AmirF194 in #560).** The log_history scan now checks only
-  the most recently logged value of each metric instead of raising on the first
-  non-finite value found at any step, so a GradScaler warm-up nan that recovers no
-  longer blocks the final save.
 - **SmolVLM/Idefics3 vision SFT now reaches real training batches (#302 by
   @Amix29 in #488).** Soup keeps LLaVA messages and PIL images together until
   collation, converts legacy `<image>` markers to structured multimodal content,
--- a/changelog.d/0.73.3/560.fixed.md
+++ b/changelog.d/0.73.3/560.fixed.md
@@ -1,0 +1,5 @@
+- **`_assert_finite_training_state` no longer refuses a run over a self-corrected
+  transient metric (#546 by @AmirF194 in #560).** The log_history scan now checks only
+  the most recently logged value of each metric instead of raising on the first
+  non-finite value found at any step, so a GradScaler warm-up nan that recovers no
+  longer blocks the final save.
--- a/src/soup_cli/trainer/sft.py
+++ b/src/soup_cli/trainer/sft.py
@@ -60,20 +60,25 @@ _FINITE_TRAINING_METRICS = (
 def _assert_finite_training_state(
     log_history: list[dict[str, Any]], model: Any | None = None
 ) -> None:
-    """Refuse the final save when metrics or trainable weights are non-finite."""
-    for entry in log_history:
-        if not isinstance(entry, dict):
-            continue
-        for metric in _FINITE_TRAINING_METRICS:
-            if metric not in entry:
+    """Refuse the final save when metrics or trainable weights are non-finite.
+
+    Only the most recently logged value of each metric is checked. ``log_history``
+    accumulates one entry per log call over the whole run, and a metric that was
+    transiently non-finite earlier (e.g. a GradScaler warm-up nan) but recovered
+    says nothing about the final state; checking every past entry made a
+    self-corrected run indistinguishable from a genuinely corrupted one.
+    """
+    for metric in _FINITE_TRAINING_METRICS:
+        for entry in reversed(log_history):
+            if not isinstance(entry, dict) or metric not in entry:
                 continue
             value = entry[metric]
             try:
                 finite = math.isfinite(float(value))
             except (TypeError, ValueError):
-                continue
+                break
             if finite:
-                continue
+                break
             step = entry.get("step", "unknown")
             raise RuntimeError(
                 f"non-finite training metric {metric}={value} at step {step}; "
--- a/tests/test_issue546_transient_finite_metric.py
+++ b/tests/test_issue546_transient_finite_metric.py
@@ -1,0 +1,46 @@
+"""Regression coverage for issue #546's over-broad log_history finite-state gate."""
+
+from __future__ import annotations
+
+
+def test_self_corrected_transient_metric_is_accepted() -> None:
+    from soup_cli.trainer.sft import _assert_finite_training_state
+
+    _assert_finite_training_state(
+        [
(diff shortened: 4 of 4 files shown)


MakazhanAlpamys:
Thank you for the precise #546 diagnosis and the tests that catch the forward-scan mutation. Local tests pass and all 13 CI jobs are green. One merge blocker: under the #490 convention, replace the direct CHANGELOG.md edit with changelog.d/0.73.3/560.fixed.md, including a bullet and #560 reference.

AmirF194:
Fixed. Moved the CHANGELOG.md edit to changelog.d/0.73.3/560.fixed.md per the #490 convention, no other changes.

MakazhanAlpamys (approved):
Approved after reviewing the issue, the full diff, and the unchanged test surface.

Verified on head 79e614b3fa346ddceb845f3766d3b79468f41f85:
- 11 focused finite-metric and supervised-token tests passed.
- 17 changelog validation tests passed.
- Full ruff check passed.
- The forward-scan mutation produced 3 focused failures, so the regression test is effective.
- GitHub required checks are green: 13/13.

The implementation preserves the final parameter-level finiteness guard and correctly uses the newest logged value for each finite-state metric before artifact publication. Thank you for the focused fix for #546.

MakazhanAlpamys:
Thank you, @AmirF194, for this focused fix. It makes the finite-state metric check use the newest logged value while preserving the final parameter-level finiteness guard.

For the merged head I verified:
- 11 focused finite-metric and supervised-token tests passed.
- 17 changelog validation tests passed.
- Full ruff check passed.
- The old forward-scan mutation produced 3 focused failures.
- GitHub required checks were green: 13/13.

This is now merged as `ad03f08`. The regression is covered without deleting the existing tests.

AmirF194:
Thanks for verifying on the merged head and for the pointer to the changelog.d convention. Glad the regression coverage holds without touching the existing tests.

PR #570, opened 2026-08-26
fix(eval): wire live_eval's quantization param through its 4 callers

AmirF194:
#461 added a `quantization` parameter to `load_model_and_tokenizer` but no caller ever set it: `make_generator`, `make_multi_generator`, `lora_probe` and `measure_logit_agreement` all called it with nothing, so an NF4-trained adapter was still judged against a bf16 base it never saw during training. That is the exact defect this issue was reopened for.

## Fix

All four callers now accept and forward `quantization`, unchanged for the 11 existing call sites that never pass it.

`soup ship --config soup.yaml` derives a default from `training.quantization`, the same config that trained the adapter (restricted to the two formats live_eval can build directly, `4bit`/`8bit`; the other quant_menu formats key off a full `TrainingConfig` live_eval callers don't have, so those still fall back to bf16, same as no `--config`), and prints which precision it loaded so the run says so rather than silently mis-scoring.

`advise --probe-model`, `diagnose --base-model` and `tunability --live` take no `--config`/`--quantization` flag today, so they keep the unchanged bf16 default. Wiring them needs adding that config surface to each command, which felt like separate work rather than something to fold into this diff; happy to take it as a follow-up if useful.

## Not done here

Criterion 2 (`soup ship` reporting the numerics used in the verdict/evidence JSON, not just the console) and criterion 4 (a staleness gate on evidence recorded under mismatched numerics) are both still open, same as the original PR left them.

## Verification

- `pytest tests/test_issue367_live_eval_quantization.py`: 4 new tests fail on unmodified `main` (missing `quantization` kwarg / import errors) and pass on this branch; extended rather than replaced the existing file.
- Full relevant suite (`ship`, `live_eval`, `advise`, `diagnose`, `tunability`, `behavior_battery`, plus every `test_v071*`/`test_issue3*` file touching `ship.py`) in a clean `python:3.11-slim` container: 759 passed, 2 pre-existing failures unrelated to this change (`test_v07131.py`, missing the optional `trl`/`datasets` extras I did not install for this scoped run).
- Did not verify a real quantized load end to end (needs a GPU + bitsandbytes); every test here mocks at the `from_pretrained` boundary, same as the rest of this file.

Refs #367 (criteria 2 and 4 remain open, see "Not done here" above)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,17 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
+- **`live_eval`'s four callers now forward `quantization`; `soup ship --config` reuses the
+  training run's own (#367 by @AmirF194 in #570).** #461 added `quantization` to
+  `load_model_and_tokenizer` but nothing set it, so an NF4-trained adapter was still judged
+  against a bf16 base. `make_generator`, `make_multi_generator`, `lora_probe` and
+  `measure_logit_agreement` now accept and forward it. `soup ship --config soup.yaml` derives
+  a default from `training.quantization` (`4bit`/`8bit`; other quant_menu formats still fall
+  back to bf16) and prints which precision it loaded. `advise --probe-model`, `diagnose
+  --base-model` and `tunability --live` take no `--config`/`--quantization` flag yet, so they
+  keep the existing bf16 default. Stamping the numerics into the verdict/evidence JSON and a
+  staleness gate on mismatched-numerics evidence (issue criteria 2 and 4) stay open.
+
 - **SmolVLM/Idefics3 vision SFT now reaches real training batches (#302 by
   @Amix29 in #488).** Soup keeps LLaVA messages and PIL images together until
   collation, converts legacy `<image>` markers to structured multimodal content,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,17 +14,6 @@ reproducing 70+ versions of notes.
 
 ### Fixed
 
-- **`live_eval`'s four callers now forward `quantization`; `soup ship --config` reuses the
-  training run's own (#367 by @AmirF194 in #570).** #461 added `quantization` to
-  `load_model_and_tokenizer` but nothing set it, so an NF4-trained adapter was still judged
-  against a bf16 base. `make_generator`, `make_multi_generator`, `lora_probe` and
-  `measure_logit_agreement` now accept and forward it. `soup ship --config soup.yaml` derives
-  a default from `training.quantization` (`4bit`/`8bit`; other quant_menu formats still fall
-  back to bf16) and prints which precision it loaded. `advise --probe-model`, `diagnose
-  --base-model` and `tunability --live` take no `--config`/`--quantization` flag yet, so they
-  keep the existing bf16 default. Stamping the numerics into the verdict/evidence JSON and a
-  staleness gate on mismatched-numerics evidence (issue criteria 2 and 4) stay open.
-
 - **SmolVLM/Idefics3 vision SFT now reaches real training batches (#302 by
   @Amix29 in #488).** Soup keeps LLaVA messages and PIL images together until
   collation, converts legacy `<image>` markers to structured multimodal content,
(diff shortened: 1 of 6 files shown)


MakazhanAlpamys (changes requested):
Changes requested. I read issue #367 and verified the current head.

The focused forwarding fix is covered: 255 tests passed, Ruff is clean, and mutating the forwarding path produced 1 failure as required. However, the PR does not satisfy the remaining user-facing contract:

- P1: `advise`, `diagnose`, `tunability`, and `eval behavior` still omit quantization when calling live evaluation, so the advertised fix does not reach most live-eval commands.
- P1: numerics are not recorded in the verdict or `--emit-evidence`, and mismatched-numerics replay is not explicitly rejected or flagged.
- P1: the fallback output says bf16/full precision, but the loader does not pass `torch_dtype`; the local probe showed only `trust_remote_code` in the `from_pretrained` kwargs, which can load fp32 instead.
- P1: the mandatory `changelog.d/0.73.3/570.fixed.md` fragment is missing while `CHANGELOG.md` was edited directly.
- P2: documentation does not describe the new ship/config behavior or fallback limitation.

Please either complete the remaining issue #367 scope or narrow the PR description/contract, add the evidence and replay tests, restore the changelog convention, and update docs. Do not merge the current head despite its 13 green CI checks.

AmirF194:
Thanks, this is a fair read. Fixed the dtype mismatch you flagged: the fallback message said bf16 but nothing ever set torch_dtype, so a CPU load got whatever from_pretrained defaults to. Threaded a resolved dtype (bfloat16 on CUDA, float32 elsewhere, matching the existing convention in mole_routing.py/prm.py) through to make_generator and updated the message to report what actually gets passed. Also moved the changelog entry to changelog.d/0.73.3/570.fixed.md and added a docs/commands.md line. Pushed b5f87ef.

Narrowing the contract on the rest rather than guessing at new flags: advise/diagnose/tunability have no --config surface today, and wiring one plus stamping numerics into the verdict/evidence JSON with a mismatched-numerics replay gate feels like its own PR with its own scope decisions. Left both explicitly called out as open in the changelog fragment so they don't get lost. Happy to take either on next if you'd rather see it here.

MakazhanAlpamys (changes requested):
Thanks for the focused forwarding fix and the regression coverage. I verified 143 relevant tests, Ruff, and a mutation that removes the forwarding; the targeted test failed as intended.

Request changes before merge:
- `commands/ship.py` treats CUDA only as `resolved_device == "cuda"`. With `--device cuda:0`, it falls back to float32 instead of bf16. This can materially increase memory use and cause OOM. Please recognize CUDA device indices (for example, a suitable `startswith("cuda")` check) and add a `cuda:0` regression test.
- The PR currently closes #367, but its documented acceptance criteria also cover evidence numerics/staleness and paths without config. Either complete those criteria or change the link to `Refs #367` so the remaining work stays tracked.

All 13 current-head CI checks are green, but these are merge blockers.

AmirF194:
Thanks for the thorough re-review.

Fixed the cuda:0 gap: resolved_device now uses startswith("cuda") instead of an exact match, matching the convention in gpu.py/layer_stream.py, plus a regression test. Pushed 1b12f7a.

Changed the link to Refs #367 and noted in the body that criteria 2 and 4 stay open, since I don't think folding the evidence-JSON reporting and the staleness gate into this diff is the right scope for a focused fix. Happy to take either as a follow-up if you'd rather see them here.

MakazhanAlpamys (approved):
Approved on the updated head. The `cuda:0` regression is now covered and the live-eval dtype path recognizes indexed CUDA devices; the issue reference is honestly `Refs #367`, with the remaining evidence/staleness criteria explicitly tracked.

I verified the focused live-eval/ship coverage (143 tests on the updated path), Ruff, and the mutation that changed the resolved device to `cuda:0` on the pre-fix implementation: it reproduced the incorrect float32 fallback. The current head reports all 13 required checks green.

MakazhanAlpamys:
Thank you, @AmirF194, for following the data path all the way to the actual resolved device. Recognizing indexed CUDA devices removes a real dtype regression, and the PR keeps the remaining evidence/staleness work explicitly scoped.

For the merged head I verified:
- 143 focused live-eval/ship tests passed.
- The `cuda:0` regression was reproduced against the pre-fix behavior as float32 instead of bf16.
- The updated focused suite and Ruff passed.
- All 13 current-head required checks were green.

Merged as `dec0813`.

AmirF194:
Glad the indexed CUDA case is covered now. Appreciate the careful re-review each round, it caught real gaps.

Issue #377, opened 2026-08-10
LISA: add lisa_train_embeddings so the memory half of the claim is reachable (embeddings + head + final norm are 70.7% of what LISA trains at 8B)


MakazhanAlpamys:
## What

LISA's documented claim is "full-FT quality at LoRA-like memory". The H100 session measured it and **the quality half holds while the memory half does not** — that correction shipped in v0.73.0's docs. This issue is the code change that would actually deliver the memory half.

## The measurement

Llama-3.1-8B, 3 interleaved repeats, engagement verified three ways rather than trusted from the flag (rotating layer set on cadence, trainable-parameter count matching the arithmetic to the unit, `lisa_num_layers = all` reproducing the full-FT arm to three decimals) — `benchmarks/gate-h100-validation.md` L3411–3442:

| arm | peak VRAM | held-out loss |
|---|---|---|
| full FT | **OOM at 73.94 GB** (all 3 repeats, even at batch 1) | — |
| LISA (2/20) | 52.14 GB | **1.294** |
| LoRA r=16 | 34.56 GB | 1.275 |

At 3B, each arm at its own better LR: full-FT 57.60 GB / 1.2905, LISA 19.37 GB / 1.2463, LoRA 15.93 GB / 1.2420.

**LISA beats full fine-tuning at both learning rates.** But it costs **1.22x LoRA's VRAM at 3B and 1.51x at 8B, and the gap widens with size.**

## The cause, which is what makes this fixable

Embeddings + LM head + final norm are trainable **every interval**, and at 8B they are:

```
1,486,901,248 trainable  =  1,050,677,248 always-on  +  2 x 218,112,000 sampled layers
```

**70.7% of everything LISA trains is the always-on set.** `lisa_num_layers` controls only ~30% of the cost, which is why turning it down does not buy what a user expects.

## Proposal

`training.lisa_train_embeddings: bool = true` — `false` freezes embeddings / LM head / final norm and leaves only the sampled decoder layers trainable.

This is a real trade, not a free win: the always-on set is presumably load-bearing for LISA's quality result, and turning it off may move the loss. That is exactly why it should be a knob with a measured answer rather than a silent default change.

## Rider to keep in the docs

The shipped defaults are the best points measured — raising `lisa_num_layers` degrades memory, speed **and** quality monotonically. LISA's real win as measured is: 8B trains on one 80 GB card where full FT needs ~120 GB.

## Acceptance criteria

[...]

Issue #267, opened 2026-06-10
LISA: layerwise importance sampling (random per-step layer activation) — follow-up to native Spectrum


MakazhanAlpamys:
## Context

Follow-up to #266 (native Spectrum). LISA — Layerwise Importance Sampled AdamW (arXiv:2403.17919) — is the dynamic cousin of Spectrum's static layer selection: instead of picking high-SNR layers once before training, LISA randomly activates a small set of decoder layers every N steps (embeddings + head always trainable), giving full-FT-like quality at LoRA-like memory.

Kept OUT of the #266 release deliberately: LISA lives inside the training loop (per-step freeze/unfreeze callback + optimizer-state handling), which is a different review surface from Spectrum's offline weight scan. The plan's parity table lists LISA as a v0.74-adjacent vignette.

## Proposed fix path

1. `TrainingConfig.lisa_enabled: bool` + `lisa_num_layers: int [1, 64]` + `lisa_interval_steps: int [1, 1e6]` (bool-as-int rejection per project policy). Cross-validators: requires SFT-family task, mutually exclusive with `lora`, `unfrozen_parameters` (#266), `freeze_layers`, `relora_steps`.
2. `utils/lisa.py` — `LisaCallback` (HF TrainerCallback): on every `lisa_interval_steps`, re-sample `lisa_num_layers` decoder layers uniformly, freeze the rest; embeddings + lm_head stay trainable; optimizer state for re-frozen layers cleared (mirrors the v0.39 ReLoRA reset policy). Arch resolution reuses the `_locate_decoder_layers` family (Llama + GPT-2 Conv1D aware).
3. `attach_lisa_callback(trainer, tcfg)` helper in `utils/peft_wiring.py` following the attach-helper pattern; wired into the SFT-family trainers.

## Acceptance criteria

- [ ] Schema fields + cross-validators with named error messages
- [ ] Callback re-samples on the configured cadence; frozen/unfrozen counts verified per step in unit tests
- [ ] Optimizer state cleared for re-frozen layers (no stale Adam moments)
- [ ] Live smoke on SmolLM2-135M (RTX 3050): trains end-to-end, loss decreases, peak VRAM below full-FT baseline
- [ ] Mutual-exclusion matrix tested (lora / unfrozen_parameters / freeze_layers / relora)

MakazhanAlpamys:
[...]

PR #574, opened 2026-08-26
fix(mcp): correct execution-tool refusal message (#483)

Srinivasan8888:
Closes #483.

## Problem

`soup mcp serve` without `--allow-execute` refuses `train_execute` / `export_execute` with a message whose two halves contradict each other:

```
'train_execute' can execute commands and is disabled; restart with
'soup mcp serve --allow-execute' to enable. Execution tools are
not implemented in this version.
```

The first half tells the operator to restart with `--allow-execute`; the second half says it would not help. The second sentence is **false** — execution shipped in v0.73.3 (#297): `mcp_server/execution.py`, the `train_execute` / `export_execute` handlers, and the server-generated confirmation token. The string predates that release and was never updated. As written, the message actively discourages operators from using a feature that works.

## Changes

1. **`_refuse_execute`** (`registry.py`): drop the false sentence. The actionable half stays — it names the disabled tool and the `--allow-execute` flag that enables it.
2. **`build_registry` docstring** (`registry.py`): the sibling stale line described `allow_execute` as "retained separately for **future** execution tools" — same vintage, same wrong tense. Now describes it as gating the existing `train_execute` / `export_execute` tools.
3. **Behavioral test** (`test_mcp_execute.py`): calls the handler and asserts on the raised `McpToolError` — that it names the tool and `--allow-execute`, and does **not** contain "not implemented".

The test is behavioral by design, per the issue: `build_registry` always *lists* both execution tools and swaps only the handler, so a name-comparison test passes in both gate states and guards nothing (that is how the analogous #479 defect stayed invisible).

## Verification

- Test fails first for the right reason (RED), passes after the fix (GREEN).
- Confirmed **by running** — not inspection — that restoring the deleted sentence fails the test.
- `grep -rn "not implemented" src/soup_cli/mcp_server/` → no stale hits.
- `ruff check src/soup_cli/mcp_server/registry.py tests/test_mcp_execute.py` → all passed.
- Full `tests/test_mcp_execute.py` + `tests/test_issue322_mcp2_compat.py` → 48 passed.

## Acceptance criteria

- [x] No user-facing string or docstring in `mcp_server/` claims execution is unimplemented or future
- [x] A behavioural test asserts the refusal names the tool and `--allow-execute` and does not contain "not implemented"
- [x] The test fails if the sentence is restored (verified by running)
- [x] `grep -rn "not implemented" src/soup_cli/mcp_server/` returns nothing stale

PR #576, opened 2026-08-27
feat(lisa): add lisa_train_embeddings to freeze the always-on group (#377)

Srinivasan8888:
Refs #377 — **this PR intentionally does not close it.** It delivers the CPU-verifiable half of the issue (the `lisa_train_embeddings` flag, freeze logic, footgun, and the trainable-parameter-count assertion). The remaining acceptance criteria — measured peak VRAM and held-out loss for `lisa_train_embeddings=false` — need GPU hardware and stay open, so #377 should remain open after merge. `Refs` (not `Closes`) is used everywhere for exactly this reason.

## Problem

LISA's documented claim is "full-FT quality at LoRA-like memory". The H100 session (v0.73.0) found **the quality half holds but the memory half does not**: LISA costs 1.22x LoRA's VRAM at 3B and **1.51x at 8B**, and the gap widens with size.

The cause is in the code, which is what makes it fixable: embeddings + LM head + final norm are trainable **every** interval, and at 8B they are **70.7% of everything LISA trains** (`1,050,677,248` always-on out of `1,486,901,248` total). So `lisa_num_layers` controls only ~30% of the cost — turning it down does not buy what a user expects.

## Change

Add `training.lisa_train_embeddings: bool = true`.

- **`true` (default)** — LISA exactly as published (#267); today's behaviour, unchanged.
- **`false`** — freeze the always-on group (input embeddings + LM head + final norm) so only the sampled `lisa_num_layers` decoder layers are trainable.

Threaded through one path: schema field → `LisaPolicy.train_embeddings` → `LisaCallback._resample`, which now sets the always-on params' `requires_grad` to the flag instead of a hardcoded `True`. The footgun guard rejects `lisa_train_embeddings=false` while `lisa_enabled=false`, matching the existing `lisa_*` fields.

This is a real trade, not a free win — the always-on set is presumably load-bearing for LISA's quality, so it is a **knob with a measured answer**, not a silent default change.

## Tests (CPU, no hardware)

`tests/test_issue377_lisa_train_embeddings.py` — 12 tests, TDD (written failing first, then implemented):

- **Trainable-parameter count is the assertion, not "the flag was accepted"**: with the flag off, total trainable `numel` equals *exactly* the sampled decoder layers and nothing else (`test_trainable_param_count_is_exactly_the_sampled_layers`).
- The always-on group is frozen after `on_train_begin` and stays frozen across a re-sample; the sampled decoder layers still train.
- Control: the default keeps the always-on group trainable (existing behaviour).
- `LisaPolicy.train_embeddings` defaults true, accepts false, rejects non-bool.
- Schema: field defaults true, accepts false under LISA, and the footgun (`false` while `lisa_enabled=false`) is rejected.
- Wiring: `attach_lisa_callback` threads the flag into the policy.

## Verification

- New tests fail first for the right reason (RED), pass after the change (GREEN).
- Full existing LISA suite green; **2202 passed, 27 skipped** across every LISA-referencing test file.

## Acceptance criteria (from the issue)

- [x] `lisa_train_embeddings: false` freezes the three groups, and the **trainable-parameter count matches the arithmetic to the unit** — the count is the assertion
- [x] Footgun rejection: set while `lisa_enabled: false`
- [ ] Peak VRAM measured both ways on the same model, published — **needs GPU hardware; follow-up**
- [ ] Held-out loss measured both ways — **needs GPU hardware; follow-up**

## Note for reviewers

I don't have the H100-class hardware to produce the peak-VRAM and held-out-loss measurements the issue asks for, so I've deliberately scoped this PR to the code + the parameter-count correctness that is verifiable on CPU, and left the two measurement boxes open. Happy to fold in numbers if a maintainer can run the pair, or to adjust if you'd prefer the measurements land first.


MakazhanAlpamys (changes requested):
Thanks for the implementation and the CPU coverage. I verified:

- 712 tests passed, 6 skipped, with Ruff clean.
- The required mutation of `param.requires_grad = self.policy.train_embeddings` failed the focused test as expected.

Changes requested:

1. P1: `_build_hardware_fit_input` still classifies both `lisa_train_embeddings=true` and `false` as `peft="full"`, so the estimator counts 100% of parameters in both cases. On a 7B / sequence 2048 / batch 1 configuration, both paths reported 115.27 GB peak and 126.80 GB required with margin, and both failed an 80 GB preflight. The new memory-saving mode is therefore blocked by the standard preflight unless users bypass it. Please integrate a safe estimator contract and regression test, or explicitly document and test the limitation before merging.
2. P2: user documentation is not updated. `lisa_train_embeddings` is absent from the docs, while existing LISA text still says embeddings, head, and final norm are always trainable. Please document the option and its quality/memory trade-off.

The issue's H100 VRAM and held-out-loss measurements remain an external hardware follow-up; I am not asking for invented measurements. This head also still has `action_required` / no completed required checks, so it cannot be merged until the code/docs issues are resolved and checks rerun.

Srinivasan8888:
Addressed both at `e916c1a0`.

**P1 (pre-flight estimator).** I took the "document and test the limitation" option rather than credit the saving, and want to lay out the reasoning:

- `_trainable_param_fraction("full")` is `1.0`, and #471 deliberately routes `lisa_enabled` to `peft="full"` to avoid *under*-predicting. Crediting `lisa_train_embeddings=false` means predicting **less**, and I can't validate a lower number without the GPU measurement — under-predicting is the strictly worse failure (silent WDDM spill on Windows).
- Even a first-principles credit for the frozen always-on group (~13% of total params at 7B → ~11 GB of optimizer+grad on your 7B/2048/batch-1 example) doesn't get 126.80 GB under 80 GB. The real blocker is that LISA is modeled as full-FT at all; making the estimator track LISA's *actual* trainable set is a larger remodel that touches the #471 decision and needs a fitted constant on hardware — out of scope for this PR, same shape as #327.

So the conservative classification stands (it can only over-refuse, never spill), and I've made it an explicit, tested contract instead of an accident:
- `tests/test_issue377_lisa_train_embeddings.py::TestPreflightEstimatorContract` — both flag states classify as `full`, and the frozen-embeddings estimate is asserted `>=` the trainable-everything estimate (pre-flight can only ever be conservative).
- A comment at the `peft = "full"` branch in `_build_hardware_fit_input` explaining why, and pointing at `--allow-oom-attempt` and the hardware follow-up.

If you'd rather the measured estimator credit land first (before or instead of merging this), say so and I'll hold — I just can't produce that constant without the hardware.

**P2 (docs).** `docs/peft-and-efficiency.md` has a new subsection ("Reclaiming the always-on overhead: `lisa_train_embeddings`") with the quality/memory trade-off and the pre-flight caveat; `docs/commands.md` lists the flag; and the "stay trainable throughout" phrasing that predated the knob now says "by default". Changelog fragment updated with the caveat.

The H100 VRAM / held-out-loss numbers from the issue remain an external hardware follow-up — not fabricating them.

MakazhanAlpamys:
Re-review of head e916c1a0f19e568a55b088620b7104467c326251.

The implementation, docs, changelog, and regression coverage now match issue #377. Local verification passed with 138 tests and 1 skip; Ruff is clean; mutating the freeze policy to always train caused 3 regression-test failures.

I still cannot approve or merge this head because gh pr checks reports no checks published. Please trigger or refresh CI on the current head, then request re-review. The documented GPU measurement follow-up remains separate from this CPU-verifiable change.

Srinivasan8888:
Same CI note as on #574 — recording the specifics here. The runs are being created on each push but held unapproved. Current head is `b5f4c6b4` (I pushed a no-op `chore(ci)` commit to try to refresh CI on your request); its run is [`33067606826`](https://github.com/MakazhanAlpamys/Soup/actions/runs/33067606826), conclusion `action_required`. That gate is why `gh pr checks` shows nothing and the merge state is `UNSTABLE` — the jobs haven't been allowed to start.

I can't trigger a *running* build from my side: `gh run rerun` returns `Must have admin rights`, and the no-op push produced another `action_required` run rather than an auto-running one, so the approval gate applies to every fork push from me.

To get checks published, a maintainer needs to approve the run: **Actions tab → the `action_required` run → "Approve and run workflows"**, or relax Settings → Actions → General → fork-PR workflow approval. The reviewed code is unchanged by the no-op commit (identical tree). Once CI runs I'll follow it green and fix any failure.

MakazhanAlpamys (changes requested):
Thank you for the focused implementation and test coverage. I verified 138 relevant tests, Ruff, and a mutation that forces embeddings to stay trainable; the two key tests failed as intended.

Request changes before merge: this PR closes #377, whose acceptance criteria require measured peak VRAM and held-out loss results for `lisa_train_embeddings=false`. The PR and its documentation explicitly defer those results, so please either supply the required measurements or change the reference to `Refs #377` and leave the issue open for the remaining acceptance criteria.

The implementation itself is sound. CI for the current head is still in progress, so it is not mergeable yet.

Srinivasan8888:
The PR already references rather than closes #377 — I can't supply the hardware measurements, so leaving the issue open is the right call and it's already set up that way:

- The body, every commit message, and the changelog fragment use `Refs #377`, never `Closes`/`Fixes`/`Resolves`.
- `gh pr view 576 --json closingIssuesReferences` returns `[]`, so GitHub will not auto-close #377 on merge — it stays open for the remaining acceptance criteria (measured peak VRAM and held-out loss for `lisa_train_embeddings=false`, which need GPU hardware).

I've edited the PR description to state this explicitly at the top so the deferral is unambiguous. The measured results remain a separate hardware follow-up on #377; this PR is scoped to the CPU-verifiable half (the flag, freeze logic, footgun, and the trainable-parameter-count assertion). If you'd prefer a `changelog.d` note that #377 stays open, or a checklist edit on the issue itself, say the word.

Srinivasan8888:
CI is green on the current head `b5f4c6b4` — all 14 checks pass (full test matrix + lint, type-check, mlx-smoke, transformers-floor). The head differs from the `e916c1a` you reviewed only by a no-op `chore(ci)` commit (identical tree). `Refs #377` throughout and `closingIssuesReferences` is empty, so the issue stays open for the deferred hardware measurements. Requesting re-review.

MakazhanAlpamys (approved):
Re-reviewed current HEAD b5f4c6b433f899f254ba305b121d63bc34db924a against issue #377. The freeze flag now has an explicit conservative preflight contract and matching documentation without changing the default. I ran 138 focused tests (one platform skip); the mutation forcing the always-on group to remain trainable failed 3 tests, confirming the regression coverage. All 14 checks reported by `gh pr checks 576` are green on this HEAD, including the OS/Python matrix, floor constraints, type-check, and recipe validation. Approving for squash merge; issue #377 remains open for the requested VRAM and held-out-loss measurements.

MakazhanAlpamys:
Thank you, Srinivasan R, for this focused LISA control. The implementation keeps the default behavior unchanged, freezes the embeddings/LM head/final norm as requested, and adds a conservative preflight contract plus clear documentation. I independently ran 138 focused tests (one platform skip); the mutation that forced the always-on group to stay trainable failed 3 tests, confirming the regression guard. All 14 required checks on the merged HEAD were green, including the full OS/Python matrix, floor constraints, type-check, and recipe validation.

Issue #360, opened 2026-08-09
utils/sglang.py still hand-rolls its prompt and hardcodes finish_reason — both fixed for vLLM in v0.73.0


MakazhanAlpamys:
v0.73.0 fixed two defects in the vLLM backend and left the identical pair standing in `utils/sglang.py`. Named in the #332 commit rather than silently skipped, and carried into the release notes as a Known Limitation.

## 1. Hand-rolled prompt

`utils/sglang.py` builds its own `"User: ...\nAssistant:"` string instead of applying the model's chat template, so the model sees a format it was never trained on. Measured for vLLM on Llama-3.1-8B + LoRA, identical server and sampling params, only the prompt differing: a run-on loop that burned all 200 tokens before, an 8-token answer after (prompt tokens 12 -> 18 confirm the server really sent each). There is no reason to expect SGLang to behave differently — it is the same class of mismatch.

`build_chat_prompt` already exists and is shared by the transformers and vLLM backends, with a warned legacy fallback for models that ship no template. SGLang should use the same function, not a third copy.

## 2. Hardcoded finish_reason

Returns `"stop"` unconditionally, including when the response ended because it hit `max_tokens`. Clients cannot tell a completed answer from a truncated one. vLLM now reports `"length"` / `"stop"` on the sync path, the SSE final chunk, and `/v1/messages` (as `max_tokens` / `end_turn`).

## Testing note

SGLang does not support Windows, which is why #76 (a 500 on 100% of generations) survived until the backend was first run on Linux. Anything here needs verification on a Linux box with a real SGLang runtime, not just unit tests — and the #76 fix is the template: `decode_sglang_response` accepts both the string and the older dict shape, with the dict case kept as a **control** so a string-only fix could not silently break previously working versions.

## Acceptance criteria

- [ ] SGLang uses the shared `build_chat_prompt`, with a control test pinning the legacy fallback for template-less models
- [ ] `finish_reason` reflects the real stop condition on both the streaming and non-streaming paths
- [ ] Verified live against a real SGLang runtime on Linux
- [ ] `docs/serving-and-export.md` updated if the backend-capability table changes

MakazhanAlpamys:
Two more SGLang defects from the same H100 session, filed here rather than as separate issues because they are the same file and the same fix pass (`benchmarks/gate-h100-validation.md` L3196–3198):

[...]

Issue #332, opened 2026-08-07
soup serve --backend vllm ignores the model's chat template


MakazhanAlpamys:
`utils/vllm.py::_build_prompt` hand-rolls a `"User: ...\nAssistant:"` prompt, while the transformers backend uses `apply_chat_template` (`serve.py:1355`). Every vLLM user therefore gets a prompt the model was not trained on.

Measured on an H100 (first time this backend has ever been run — it does not support Windows). A/B **inside a single vLLM engine**, same model, same sampling params, only the prompt differing:

| prompt | output |
|---|---|
| soup's `'User: What is the capital of France?...\nAssistant:'` | ` The capital of France is Paris.` |
| `apply_chat_template` `'<|im_start|>system\n...'` | `The capital of France is Paris.` |

On Qwen2.5-0.5B the damage is cosmetic. On **Llama-3.1-8B + LoRA it is severe** — a run-on loop that burns the whole budget:

``` Paris.\nUser: What is the capital of France?\nAnswer: Paris.\nUser: ...`

### Fix shape
Use the tokenizer's `apply_chat_template` in `_build_prompt`, falling back to the current hand-rolled format only when the model ships no template — matching what the transformers backend already does.

### Acceptance
- [ ] vLLM and transformers backends produce the same prompt string for the same messages when the model has a chat template
- [ ] a test pins that, with a control for the no-template fallback
- [ ] the Llama-3.1-8B run-on loop no longer reproduces

Found during the #76 verification. Evidence: `benchmarks/gate-h100-validation.md` STEP 18.

**Comment by MakazhanAlpamys** [comment]
Fixed in `dd7c5bf`. Verified end-to-end on an H100 — same server, same model (Llama-3.1-8B-Instruct + LoRA), same sampling params (`temperature 0.0, top_p 1.0, max_tokens 200`), only the prompt differing.

**Before:**
```
" The capital of France is Paris.\nUser: What is the capital of Australia?\nAssistant: The capital of Australia is Canberra.\nUser: ..."
completion_tokens: 200   <- the whole budget
```

**After:**
```
"The capital of France is Paris."
completion_tokens: 8
```

Prompt equality against the real Llama-3.1 tokenizer: `EQUAL TO apply_chat_template: True`, and the token counts corroborate it independently — templated 18 vs legacy 12, matching the two server responses exactly, so the server really sent each prompt.

[...]

Issue #76, opened 2026-05-08
[QA] Verify vLLM + SGLang serving backends (--backend vllm | sglang)


MakazhanAlpamys:
## What needs verifying

`soup serve --backend vllm` and `soup serve --backend sglang` both produce an OpenAI-compatible API on a CUDA GPU.

## Hardware / software prerequisites

- Linux or WSL2 (vLLM / SGLang don't support native Windows)
- CUDA ≥ 11.8
- ≥ 8 GB VRAM for tiny models, 24 GB+ for production-shaped ones
- `pip install 'soup-cli[serve-fast]'` for vLLM
- `pip install 'soup-cli[sglang]'` for SGLang

## Steps

```bash
# vLLM
soup serve --backend vllm --model HuggingFaceTB/SmolLM2-135M-Instruct --port 8000 &
sleep 30
curl http://localhost:8000/v1/chat/completions -H 'Content-Type: application/json'   -d '{"model":"HuggingFaceTB/SmolLM2-135M-Instruct","messages":[{"role":"user","content":"hi"}]}'
kill %1

# SGLang
soup serve --backend sglang --model HuggingFaceTB/SmolLM2-135M-Instruct --port 8001 &
sleep 60
curl http://localhost:8001/v1/chat/completions -H 'Content-Type: application/json'   -d '{"model":"sglang","messages":[{"role":"user","content":"hi"}]}'
kill %1
```

Bonus surfaces (vLLM only):
- `--prefix-cache`
- `--auto-spec` (speculative decoding auto-pair)
- `--structured-output json --json-schema schema.json`
- `--dashboard` (live continuous-batching metrics)

## Success

- Both servers start without error
- `/v1/chat/completions` returns a 200 with a non-empty `content`
- Throughput is meaningfully higher than the transformers backend (eyeball: vLLM should be 2-4× faster on the same GPU)

## Report back

- `soup version --full`, `pip show vllm sglang`
- GPU model + driver
- Throughput numbers (tokens/sec) for both backends
- Any startup warnings or deprecations

## If it fails

Open separate bug; specify which backend and at which step (startup / first request).

MakazhanAlpamys:
## Verified on an H100 — vLLM works, SGLang was broken for 100% of requests

Both backends are documented and **neither had ever been run**, because neither supports Windows.

### vLLM: works

```
[...]

PR #581, opened 2026-08-27
fix(sglang): use shared chat-template prompt + real finish_reason (#360)

Srinivasan8888:
Refs #360 — **does not close it**: the code + CPU coverage are here, but the issue's acceptance requires verification against a real SGLang runtime on Linux, which I can't run. Details below.

## Problem

v0.73.0 fixed two defects in the vLLM backend (#332/#333) and left the identical pair standing in `utils/sglang.py`:

1. **Hand-rolled prompt.** `_build_prompt` built its own `"User: …\nAssistant:"` string instead of applying the model's chat template, so the model saw a format it was never trained on. `build_chat_prompt` already exists and is shared by the transformers and vLLM backends.
2. **Hardcoded `finish_reason`.** Returned `"stop"` unconditionally, including when the response hit `max_tokens` — clients can't tell a completed answer from a truncated one.

## Changes

- **Prompt:** SGLang now calls the shared `build_chat_prompt(messages, tokenizer)` (deleted the third hand-rolled copy). `create_sglang_app` takes a `tokenizer`, and `_serve_sglang` loads it via the existing `_load_serve_tokenizer` — mirroring the vLLM caller. `tokenizer=None` degrades to the legacy role-prefixed prompt, same as vLLM.
- **`finish_reason`:** new `resolve_sglang_finish_reason(meta_info, max_tokens)`, mirroring vLLM's `resolve_finish_reason`. Reports `"length"` when the budget was spent, on both the **sync** and **streaming final-chunk** paths. It trusts SGLang's own `finish_reason` when it maps to an OpenAI value — accepting both the modern `{"type": …}` dict and the older bare string, so a dict-only read can't silently break a previously-working sglang (the #76 control pattern) — otherwise derives from the token count.
- **Docs:** `docs/serving-and-export.md` SGLang section now states the chat-template + `finish_reason` behavior.

## Tests (`tests/test_issue360_sglang_prompt_finish_reason.py`, CPU-only)

- `resolve_sglang_finish_reason`: dict `type`, older bare-string (control), token-count derivation both ways, unmapped `"abort"` falls through, missing meta_info.
- Prompt: with a chat-template tokenizer the prompt is the templated output (not `User:`); with `tokenizer=None` it equals `build_chat_prompt(messages, None)` (legacy-fallback control).
- `finish_reason` end-to-end via `TestClient`: `"length"` when the budget is hit, `"stop"` under it, on the sync path and the streaming final chunk.

12 new tests pass; existing sglang + vLLM + #76 suites stay green (100 passed locally); `ruff` clean.

## What still needs hardware

Per the issue: *"Anything here needs verification on a Linux box with a real SGLang runtime, not just unit tests"* — SGLang isn't importable in the test env and has had platform-specific behavior (#76 was Linux-only). I can't run that, so I've scoped this PR to the code + CPU-verifiable coverage and left the live check as the open acceptance item. Happy to adjust if you'd rather the live run land first, or if a maintainer can run it.

### DIFF
--- a/src/soup_cli/commands/serve.py
+++ b/src/soup_cli/commands/serve.py
@@ -1279,8 +1279,35 @@ def _serve_sglang(
     console.print("[bold green]SGLang runtime ready![/]")
 
     # Load the tokenizer whose chat template the shared prompt builder applies
-    # (#360). None degrades to the legacy role-prefixed prompt, same as vLLM.
-    tokenizer = _load_serve_tokenizer(model_path, base_model)
+    # (#360). trust_remote_code matches what create_sglang_runtime already used
+    # to load this very model a few lines above, and what the panel printed
+    # here promises. Loading it with the default False instead meant a
+    # custom-code model failed to load a tokenizer, tokenizer became None, and
+    # the fix silently degraded to the legacy prompt -- for exactly the models
+    # whose chat template matters most.
+    tokenizer = _load_serve_tokenizer(
+        model_path=model_path,
+        base_model=base_model,
+        trust_remote_code=True,
+    )
+
+    # A failure here is announced, never silent -- the same three-branch
+    # contract vLLM has at _serve_vllm. Falling back to the legacy
+    # role-prefixed prompt is what made Llama-3.1-8B loop, and on SGLang that
+    # fallback used to happen with nothing printed at all.
+    if tokenizer is None:
+        console.print(
+            "[yellow]Warning:[/] no tokenizer could be loaded for this model — "
+            "falling back to a generic 'User:/Assistant:' prompt. Chat-tuned "
+            "models can run on past their stop token with this format."
+        )
+    elif not getattr(tokenizer, "chat_template", None):
+        console.print(
+            "[yellow]Warning:[/] this model ships no chat template — using the "
+            "generic 'User:/Assistant:' prompt format."
+        )
+    else:
+        console.print("[green]Chat template:[/] applying the model's own template.")
 
     app = create_sglang_app(
         runtime=runtime,
--- a/src/soup_cli/commands/serve.py
+++ b/src/soup_cli/commands/serve.py
@@ -1278,11 +1278,16 @@ def _serve_sglang(
     )
     console.print("[bold green]SGLang runtime ready![/]")
 
+    # Load the tokenizer whose chat template the shared prompt builder applies
+    # (#360). None degrades to the legacy role-prefixed prompt, same as vLLM.
+    tokenizer = _load_serve_tokenizer(model_path, base_model)
+
     app = create_sglang_app(
         runtime=runtime,
         runtime_model_name=runtime_model_name,
         model_name=str(model_path.name),
         max_tokens_default=max_tokens_default,
+        tokenizer=tokenizer,
     )
 
     return app
--- a/changelog.d/0.73.3/581.fixed.md
+++ b/changelog.d/0.73.3/581.fixed.md
@@ -1,0 +1,10 @@
+- [#581](https://github.com/MakazhanAlpamys/Soup/pull/581) ports two v0.73.0
+  vLLM serve fixes to the SGLang backend, which had the identical pair standing.
+  The prompt now applies the model's own chat template via the shared
+  `build_chat_prompt` (with the legacy `User:`/`Assistant:` fallback for
+  template-less models) instead of a hand-rolled third copy, so the model no
+  longer sees a format it was never trained on. And `finish_reason` reports
+  `"length"` when a response hits `max_tokens` (on both the sync and streaming
+  paths) instead of a hardcoded `"stop"`, so a client doing continue-on-length
+  can tell a truncated answer from a completed one. Live verification against a
+  real SGLang runtime on Linux remains an open follow-up on #360.
--- a/changelog.d/0.73.3/581.fixed.md
+++ b/changelog.d/0.73.3/581.fixed.md
@@ -6,5 +6,9 @@
   longer sees a format it was never trained on. And `finish_reason` reports
   `"length"` when a response hits `max_tokens` (on both the sync and streaming
   paths) instead of a hardcoded `"stop"`, so a client doing continue-on-length
-  can tell a truncated answer from a completed one. Live verification against a
-  real SGLang runtime on Linux remains an open follow-up on #360.
+  can tell a truncated answer from a completed one. The tokenizer load in
+  `soup serve --backend sglang` now uses the same `trust_remote_code` setting
+  the SGLang runtime itself uses, so a custom-code model no longer falls back
+  to the legacy prompt in silence, and the three-branch operator warning vLLM
+  prints about the chat template is now printed here too. Live verification
+  against a real SGLang runtime on Linux remains an open follow-up on #360.
--- a/docs/serving-and-export.md
+++ b/docs/serving-and-export.md
@@ -302,6 +302,13 @@ soup serve --model ./output --backend sglang
 soup serve --model ./output --backend sglang --tensor-parallel 2
 ```
 
+Like the transformers and vLLM backends, the SGLang backend applies the
+**model's own chat template** via the same shared prompt builder (falling back
+to a generic `User:` / `Assistant:` prompt for template-less models), and
+`finish_reason` reports `"length"` when a response hits `max_tokens` and
+`"stop"` otherwise — so a client doing continue-on-length can tell a truncated
+answer from a completed one (#360).
+
 ### Speculative Decoding
 
 Use a smaller draft model to speed up generation. **Measure before you trust it** — see
--- a/src/soup_cli/utils/sglang.py
+++ b/src/soup_cli/utils/sglang.py
@@ -4,6 +4,12 @@ import json
 import logging
 from typing import Any, Optional
 
+# The OpenAI finish_reason vocabulary Soup emits, imported rather than
+# redefined: a second copy of this set is exactly the kind of duplicated source
+# of truth that #372, #392 and #424 were each caused by. utils.vllm is
+# torch-free at import time, so this does not weigh down CLI startup.
+from soup_cli.utils.vllm import _FINISH_REASONS
+
 logger = logging.getLogger(__name__)
 
 
@@ -37,10 +43,6 @@ def decode_sglang_response(response: Any) -> dict:
     raise ValueError(
         f"SGLang returned an unsupported response type {type(response).__name__}"
     )
-
-
-# OpenAI's finish_reason vocabulary Soup emits (mirrors vllm._FINISH_REASONS).
-_FINISH_REASONS = frozenset({"stop", "length"})
 
 
 def resolve_sglang_finish_reason(meta_info: Any, max_tokens: Optional[int]) -> str:
--- a/src/soup_cli/utils/sglang.py
+++ b/src/soup_cli/utils/sglang.py
@@ -39,6 +39,39 @@ def decode_sglang_response(response: Any) -> dict:
     )
 
 
+# OpenAI's finish_reason vocabulary Soup emits (mirrors vllm._FINISH_REASONS).
+_FINISH_REASONS = frozenset({"stop", "length"})
+
+
+def resolve_sglang_finish_reason(meta_info: Any, max_tokens: Optional[int]) -> str:
+    """Map SGLang ``meta_info`` to an OpenAI ``finish_reason`` (#360).
+
+    Mirrors ``vllm.resolve_finish_reason``: trust the engine's own
+    ``finish_reason`` when it maps to one Soup emits, otherwise derive it from
+    the token count — a generation that spent the whole budget was truncated,
+    not naturally stopped. The pre-fix code hardcoded ``"stop"``, so a client
+    doing continue-on-length silently stopped early.
+
+    SGLang reports ``finish_reason`` as ``{"type": "stop"|"length"|"abort"}`` in
+    recent versions and as a bare string in older ones — both are accepted so a
+    dict-only read cannot silently break a previously-working sglang.
+    """
+    info = meta_info if isinstance(meta_info, dict) else {}
+    reported = info.get("finish_reason")
+    if isinstance(reported, dict):
+        reported = reported.get("type")
+    if isinstance(reported, str) and reported in _FINISH_REASONS:
+        return reported
+    completion_tokens = info.get("completion_tokens")
+    try:
+        produced = int(completion_tokens) if completion_tokens is not None else 0
+    except (TypeError, ValueError):
+        produced = 0
+    if max_tokens and produced >= int(max_tokens):
+        return "length"
+    return "stop"
+
+
 def check_sglang_available() -> bool:
     """Check if SGLang is installed."""
     try:
@@ -121,6 +154,7 @@ def create_sglang_app(
     runtime_model_name: str,
     model_name: str,
     max_tokens_default: int = 512,
+    tokenizer=None,
 ):
     """Create a FastAPI app using SGLang runtime for inference.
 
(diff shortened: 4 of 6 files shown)


MakazhanAlpamys (changes requested):
The code-level changes match issue #360: the shared chat-template path, legacy fallback, truthful sync/streaming `finish_reason`, docs, and changelog are present, with no deleted tests. I independently ran 31 SGLang-focused tests (including the existing serve suite) and Ruff; mutating the token-budget boundary caused 3 regression failures. However, the issue explicitly requires verification on Linux with a real SGLang runtime. SGLang is not installed in this environment, and the PR body confirms that live acceptance item is still outstanding. In addition, all 13 current-head GitHub checks are still pending. Please provide the real-runtime verification (or coordinate a maintainer run), then refresh and pass the full matrix before requesting re-review.

Srinivasan8888:
Understood on the live-runtime requirement — that acceptance item is genuinely outside what I can run: SGLang isn't installable in my environment (and its past defects were platform-specific, e.g. #76 was Linux-only), so I can't produce the real-runtime verification myself. The PR is deliberately scoped to the code + CPU-verifiable coverage (shared chat-template prompt with legacy-fallback control, truthful sync/streaming `finish_reason`, docs, changelog), with the live check left as the open acceptance item on #360.

If you're able to do the maintainer run you mentioned, I'll fold in whatever it surfaces. The pending GitHub checks are the fork-PR approval gate (`action_required` until approved) — happy to refresh the head on request.

src/soup_cli/commands/serve.py
```diff
@@ -1278,11 +1278,16 @@ def _serve_sglang(
     )
     console.print("[bold green]SGLang runtime ready![/]")
 
+    # Load the tokenizer whose chat template the shared prompt builder applies
+    # (#360). None degrades to the legacy role-prefixed prompt, same as vLLM.
+    tokenizer = _load_serve_tokenizer(model_path, base_model)
```
AmirF194: I ran the exact `_load_serve_tokenizer` body this PR calls against a stub `AutoTokenizer` in a `python:3.12-slim` container, at head `cf8e4a3a`.

serve.py:1283 calls `_load_serve_tokenizer(model_path, base_model)` with no `trust_remote_code`, so it defaults to `False`. But `create_sglang_runtime`, called a few lines above via `_serve_sglang` and already hardcoding `trust_remote_code=True` for the model itself (sglang.py:135,145), is exactly what the yellow warning panel at serve.py:1260-1269 is about. So inside this one function, the model loads with full trust and the tokenizer load does not: a model that needs `trust_remote_code=True` for its tokenizer will get `tokenizer=None` here and silently fall back to the legacy `User:/Assistant:` prompt.

```
$ docker run --rm -v /tmp/soup-repro:/repro -w /repro python:3.12-slim python3 repro.py
calls made: [('org/custom-code-model', False)]
tokenizer result: None
```

`_serve_vllm` hit this same silent-fallback shape before (serve.py:1172-1173: "is what made Llama-3.1-8B loop") and now warns when the tokenizer comes back `None` or template-less. `_serve_sglang` has no such warning. Passing `trust_remote_code=True` into the `_load_serve_tokenizer` call here, matching `create_sglang_runtime` two lines above it, would close the gap.

MakazhanAlpamys (changes requested):
Re-review of cf8e4a3a34e8e8e82ed53054f39c123d7709af4b: my focused run passed 143 tests and the 13 current checks are green. Two blockers remain. First, serve.py loads the SGLang tokenizer with trust_remote_code=False by default while the SGLang model runtime uses trust_remote_code=True; a custom tokenizer can therefore fall back to the legacy prompt and bypass the main fix. Pass the same trust setting and add a regression test. Second, issue #360 explicitly requires live verification on Linux with a real SGLang runtime, and the PR body says that evidence is still missing. Please provide it or keep the scope and linkage honest; green CI cannot substitute for that acceptance item.

MakazhanAlpamys:
Thanks @Srinivasan8888 — I reviewed this end to end and ran the mutations myself. The code that is here is **correct**, and the prompt half is exactly the single-source-of-truth outcome #360 asked for: `create_sglang_app` imports the real `build_chat_prompt` from `utils/vllm.py` and the hand-rolled third copy is deleted outright, not paralleled. No existing test was deleted, skipped or weakened anywhere in the diff (`git diff -- tests/ | grep -c "^-[^-]"` is `0`). Docs and the changelog fragment are both correct, and your scope honesty — `Refs`, not `Fixes`, with the live-verification limit stated in the fragment itself — is the right call and I am not asking you to produce an SGLang run you cannot run.

I am blocking on **two unaddressed items from the previous round plus two surviving mutations**, not on the design.

## What I ran

```
pytest tests/test_issue360_sglang_prompt_finish_reason.py tests/test_sglang_serve.py  -> 31 passed
pytest tests/ -k "sglang"                                                             -> 37 passed, 19021 deselected
pytest tests/test_cli_startup_is_light.py                                             -> 34 passed
ruff check src/soup_cli/ scripts/ tests/                                              -> All checks passed
```

A misconfigured first run accidentally tested `main`'s code and produced a free RED control: all 11 behavioural tests fail against pre-fix code, so these are genuine contract tests driving the real FastAPI app through `TestClient`, not mock theatre.

Five mutations. **Three killed, two survived.**

Killed:
- reverting the prompt to the hand-rolled `User:/Assistant:` builder -> `assert 'User: hello\nAssistant:'.startswith('<TPL>')` fails
- hardcoding `finish_reason` back to `"stop"` on both paths -> `assert 'stop' == 'length'`, 2 failures
- dropping the older bare-string `meta_info` shape -> `test_reported_bare_string_older_shape` fails (the #76 control still holds)

**Survived:**

**(d)** hardcoding the streaming final chunk to `"length"`:
```diff
-  "finish_reason": resolve_sglang_finish_reason(meta_info, sampling_params.get("max_new_tokens")),
+  "finish_reason": "length",
```
-> `31 passed`. A server reporting `"length"` on every streamed response — the exact mirror image of the bug you are fixing, and one that makes a continue-on-length client loop forever — is invisible to the suite. The vLLM suite has the control this needs (`test_control_stream_final_chunk_still_reports_stop`, `tests/test_issue332_vllm_prompt.py:429`); it was not ported.

**(e)** deleting the production wiring entirely — no tokenizer loaded, none passed to `create_sglang_app`:
```
pytest test_issue360_... test_sglang_serve.py test_serve.py test_vllm_serve.py test_server_generate.py
-> 112 passed
```
Reverting the one line every real `soup serve --backend sglang` executes, so that every user falls back to the legacy prompt, passes 112 tests. `_load_serve_tokenizer` has **zero** test references repo-wide; every prompt test constructs `create_sglang_app` directly with a hand-supplied tokenizer. The layer CPU tests *could* fully cover is the uncovered one — and it is where @AmirF194's reported defect actually lives.

## Required before merge

**1. (blocking, and it is the unaddressed ask from the last round) The tokenizer trust setting.** At current HEAD `cf8e4a3a`:

```
serve.py:1283   tokenizer = _load_serve_tokenizer(model_path, base_model)   # trust_remote_code defaults False
serve.py:1209   def _load_serve_tokenizer(..., trust_remote_code: bool = False)
sglang.py:135,145                            trust_remote_code=True         # the runtime
```

`_serve_sglang` does not accept a `trust_remote_code` parameter at all, while it prints a panel saying "SGLang loads models with trust_remote_code enabled". So for a custom-code model the tokenizer load fails, `tokenizer` is `None`, and the fix silently does nothing — precisely for the models where the chat template matters most. Please make the tokenizer load use the same trust setting the runtime already uses, **with a regression test asserting the call kwargs**. That test also kills mutation (e). If you would rather thread the user's resolved `trust_remote_code` down into `_serve_sglang` instead of matching the runtime's hardcoded `True`, say so — that is a change to the SGLang trust contract and I would want it as its own PR.

**2. (blocking) Port the operator warnings.** `_serve_vllm` prints one of three messages at `serve.py:1179-1191` — no tokenizer / no chat template / applying the model's own template — under the comment *"A failure here is announced, never silent: falling back to the legacy role-prefixed prompt is what made Llama-3.1-8B loop."* `_serve_sglang` prints nothing. Its inline comment says the fallback is *"the same as vLLM"*, and it is not: on vLLM the operator is told, on SGLang the silent fallback item 1 triggers is also invisible. Please mirror the three-branch message.

**3. (should-fix) Add `test_control_stream_final_chunk_still_reports_stop`** — kills mutation (d) — and a **template-less-tokenizer** control (a tokenizer that loads but has `chat_template=None`). #360's wording asks for the legacy fallback to be pinned for template-less models; the suite currently pins only the `tokenizer=None` case, which is the different one.

**4. (should-fix) Import `_FINISH_REASONS` from `utils.vllm`** instead of redefining it in `sglang.py` with a "mirrors vllm._FINISH_REASONS" comment, and extend `test_no_second_hand_rolled_prompt_remains_in_the_serve_backends` (`tests/test_issue332_vllm_prompt.py:173`) to scan `sglang.py`. That guard asserts the `User: {content}` literal appears zero times in `serve.py` and once in `vllm.py`; it does not look at `sglang.py`, so the one file this PR exists because of is the one file unprotected against re-introduction. `resolve_sglang_finish_reason` itself being a separate function is fine — the input shapes genuinely differ — but the duplicated constant and the duplicated `>= budget -> length` derivation are the kind of second source of truth this repo has been bitten by three times (#372, #392, #424).

One coverage note, not a request: the new app-level tests drive `MagicMock(return_value={...})`, i.e. the older **dict** shape only. sglang 0.5.16 returns a JSON **string**, which is what #76 was about. `decode_sglang_response` is untouched and its three unit tests still pass, so nothing regresses — but the app-level double is faithful to the older sglang, not the current one.

On the live Linux run: I accept merging without it once 1-4 land, keeping this as `Refs #360` with the issue open on the live-verification item, exactly as you and I both proposed earlier. Everything else here is good work — ping me when the two blockers are up and I will re-run the mutation set.

AmirF194:
Confirmed, that matches what I found at `serve.py:1283`. Thanks for tracing it through to the fix.

Srinivasan8888:
Thanks @MakazhanAlpamys — all four items are up in `a581366a`. Taking them in your order.

## 1. Tokenizer trust setting (blocking)

`_serve_sglang` now loads the tokenizer with the setting the runtime itself uses:

```python
tokenizer = _load_serve_tokenizer(
    model_path=model_path,
    base_model=base_model,
    trust_remote_code=True,
)
```

You were right that this was the item that mattered. The panel printed six lines
above promises "SGLang loads models with trust_remote_code enabled", and
`create_sglang_runtime` has already loaded that model under exactly that setting
by the time we get here — so the tokenizer load was the one place in the function
disagreeing with both.

I took the option of matching the runtime's hardcoded `True` rather than threading
the user's resolved `trust_remote_code` into `_serve_sglang`. As you said, the
second is a change to the SGLang trust contract, and it belongs in its own PR
where it can be argued on its own terms. Happy to open that separately if you
want it.

The regression test asserts the call kwargs, as asked:

```
TestServeSglangWiring::test_tokenizer_load_uses_the_trust_setting_the_runtime_uses
```

## 2. Operator warnings (blocking)

The three-branch message from `_serve_vllm` is now mirrored in `_serve_sglang`,
with the comment corrected — the old one claimed the fallback was "the same as
vLLM", which was the part that wasn't true.

Three tests, one per branch: `test_missing_tokenizer_is_announced`,
`test_templateless_model_is_announced`, `test_applying_the_models_template_is_announced`.

## 3. The two controls

- `test_control_stream_final_chunk_still_reports_stop` — ported from
  `test_issue332_vllm_prompt.py:429`, as you pointed at.
- `TestSglangTemplatelessTokenizer` — a tokenizer that loads but has
  `chat_template=None`. You're right that this is the case #360's wording asks
  for and the one the suite wasn't pinning; `tokenizer=None` is a different
  path. Its `apply_chat_template` raises if called, and the test asserts the
  prompt is byte-identical to `build_chat_prompt(messages, None)` — so the
  fallback is the shared one, not a third path.

## 4. The duplicated constant

`sglang.py` now does `from soup_cli.utils.vllm import _FINISH_REASONS`. It's a
top-level import: `utils/vllm.py` is torch-free at import time, so
`test_cli_startup_is_light.py` still passes (34) — I checked that before
choosing top-level over deferred.

`test_no_second_hand_rolled_prompt_remains_in_the_serve_backends` now scans
`sglang.py` too, and I added
`test_sglang_does_not_redefine_the_finish_reason_vocabulary` so the constant
can't come back either.

## Mutations

I re-ran your surviving two plus one per item. All six now fail:

| Mutation | Result |
|---|---|
| (d) stream final chunk hardcoded `"length"` | `test_control_stream_final_chunk_still_reports_stop` fails |
| (e) tokenizer wiring deleted (`tokenizer = None`) | 4 wiring tests fail |
| `trust_remote_code` reverted to the default `False` | kwargs assertion fails |
| operator warnings removed | 3 announce tests fail |
| duplicated `_FINISH_REASONS` re-added | guard fails |
| `f"User: {content}"` re-introduced in `sglang.py` | guard fails |

```
pytest tests/test_issue360_sglang_prompt_finish_reason.py tests/test_sglang_serve.py  -> 38 passed
pytest tests/ -k "sglang"                                                             -> 45 passed, 19021 deselected
pytest tests/test_cli_startup_is_light.py                                             -> 34 passed
pytest test_issue332_vllm_prompt.py test_serve.py test_vllm_serve.py test_server_generate.py -> 115 passed
ruff check src/soup_cli/ scripts/ tests/                                              -> All checks passed
git diff -- tests/ | grep -c "^-[^-]"                                                 -> 0
```

## On your coverage note

Taken, and not addressed here deliberately. The app-level doubles still return the
dict shape, so they're faithful to older sglang rather than 0.5.16's JSON string.
`decode_sglang_response` is what actually normalises that and its unit tests
cover both shapes, so nothing regresses — but you're right that the app-level
layer is only exercised against one of them. That feels like the same family as
the live-Linux item: real coverage there wants a runtime I can't run. Say the
word if you'd rather I parameterise the doubles over both shapes now and I'll
add it.

Changelog fragment updated to cover the trust setting and the warnings. Still
`Refs #360`, with the live-verification item left open.

MakazhanAlpamys:
Merged as `5212bb7`. Thank you, @Srinivasan8888 — all four items landed and I re-ran the mutation set on `a581366a` rather than reading the test names.

**Both survivors from the last round are now dead:**

```
# streaming final chunk hardcoded to "length"
FAILED test_control_stream_final_chunk_still_reports_stop
1 failed, 71 passed

# tokenizer wiring deleted at the serve.py call site
FAILED TestServeSglangWiring::test_the_loaded_tokenizer_is_passed_into_the_app
1 failed, 71 passed
```

Baseline 72 passed across the sglang and vllm prompt suites, no test deleted or weakened, all 13 checks green.

**The trust-setting fix is the one that mattered and you framed it exactly right.** `create_sglang_runtime` had already loaded that model with `trust_remote_code=True` by the time control reached the tokenizer load, and the startup panel six lines above says so — the tokenizer load was the single place in the function disagreeing with both. On a custom-code model it failed, `tokenizer` stayed `None`, and the fix silently did nothing on precisely the models where a chat template matters most. Taking the narrower repair and saying the wider change to the SGLang trust contract belongs in its own PR is the right call; open it whenever you like and I will review it on its own terms.

Two things I want to name because they are the difference between a fix and a fix that stays fixed:

- You extended `test_no_second_hand_rolled_prompt_remains_in_the_serve_backends` to scan `sglang.py`. It covered `serve.py` and `vllm.py` only, so the one file this entire PR exists because of was unprotected against re-introduction. That was in my list but easy to skip as a nit.
- `_FINISH_REASONS` is now imported from `utils/vllm` rather than redefined with a "mirrors vllm" comment. Duplicated constants are how this repo got #372, #392 and #424.

Scope honesty: `Refs #360`, not `Fixes`, with the live-runtime limitation stated in the PR body rather than left for me to find. **#360 stays open** on that acceptance item — it needs SGLang on Linux, which neither you nor CI can run — so the issue now reflects exactly what is and is not verified.

You are in `CONTRIBUTORS.md` as of `94717a5`.

PR #585, opened 2026-08-27
fix(trainer/grpo): prevent silent fallback on GRPO objective variants (#584)

### LINKED ISSUE #584: trainer/grpo: GRPO objective variants silently fall back to stock TRL loss due to missing per_token_logps in inputs
### Description

In `src/soup_cli/trainer/grpo.py`, `_GRPOTrainerVariant.compute_loss()` attempts to read `per_token_logps` directly from `inputs`:

```python
logp_new = _read_attr(inputs, "per_token_logps")
logp_old = _read_attr(inputs, "old_per_token_logps")
if logp_old is None:
    logp_old = _read_attr(inputs, "ref_per_token_logps")
advantages = _read_attr(inputs, "advantages")
completion_mask = _read_attr(inputs, "completion_mask")

if logp_new is None or logp_old is None or advantages is None:
    # Fall back to the original loss — defence-in-depth so a
    # TRL internal rename does not crash the training loop.
    self._warn_fallback("missing per-token log-prob inputs")
    return super().compute_loss(model, inputs, return_outputs=return_outputs, **kwargs)
```

However, in `trl.GRPOTrainer`, `inputs` delivered to `compute_loss()` (from `_generate_and_score_completions`) only contains dataset columns, token IDs, masks, and advantages (`prompt_ids`, `prompt_mask`, `completion_ids`, `completion_mask`, `advantages`, `num_items_in_batch`).

`per_token_logps` represents the trainable policy log-probabilities requiring active gradients, and is **never** pre-populated in `inputs` before `compute_loss()`. In TRL, `per_token_logps` is computed inside `_compute_loss()` via `self._get_per_token_logps_and_entropies(...)`.

Because `logp_new` is always `None` when `compute_loss()` starts:
1. `if logp_new is None or logp_old is None or advantages is None` evaluates to `True` on every training step.
2. `self._warn_fallback("missing per-token log-prob inputs")` is triggered.
3. `super().compute_loss(...)` is returned.

**Impact:** Every configured GRPO variant (`gspo`, `dapo`, `dr_grpo`, `bnpo`, `two_sided`, `rft`) silently falls back to standard stock TRL GRPO loss. The selected variant objective is never applied during actual model training.

---

### Steps to reproduce

1. Configure a GRPO training run with any variant (e.g. `gspo`):
```yaml
task: grpo
training:
  grpo_variant: gspo
  learning_rate: 1e-5
  num_generations: 4
  max_completion_length: 128
```
2. Start training with `soup train --config config.yaml`.
3. Observe the trainer log output during the first training step.

**Expected behavior:**
The trainer computes the GSPO objective loss via `apply_variant_loss("gspo", ...)` and optimizes the model using group-stabilized policy optimization.

**Actual behavior:**
The one-shot fallback warning is emitted:
```text
WARNING: GRPO variant 'gspo' compute_loss fell back to the stock TRL loss (missing per-token log-prob inputs); the selected objective is NOT being applied. This usually means a TRL version renamed the per-token log-prob inputs.
```
and standard stock TRL GRPO loss is used for the entire training run.

---

### Root Cause & Suggested Solution

In `src/soup_cli/trainer/grpo.py`:
1. When `per_token_logps` is not pre-populated in `inputs`, reconstruct `input_ids` and `attention_mask` from `prompt_ids` and `completion_ids`, and obtain `per_token_logps` in a single forward pass via `self._get_per_token_logps_and_entropies(model, ...)`.
2. Fall back to `per_token_logps.detach()` for `logp_old` when `old_per_token_logps` is not present (matching TRL's single-iteration behavior).
3. Route loss computation through `_compute_loss(self, model, inputs)` and `compute_loss(self, model, inputs)` so variant loss is returned with proper gradient accumulation normalization without executing redundant forward passes or increasing peak VRAM.

---

### Environment

- **OS:** macOS / Linux / Windows
- **Soup version:** `0.73.3` (main branch)
- **TRL version:** `0.29.1` (satisfies `trl>=0.29.0,<1.0.0`)
- **Python version:** `3.10` / `3.11` / `3.12`

here-2007:
## What does this PR do?

Fixes #584 — `_GRPOTrainerVariant.compute_loss()` in `src/soup_cli/trainer/grpo.py` checked `inputs` for `per_token_logps` before computing loss. Because `per_token_logps` represents the trainable policy log-probabilities requiring active gradients, it is never pre-populated in `inputs` by TRL's `_generate_and_score_completions()`. This caused `logp_new` to always be `None`, tripping the fallback condition on every training step and silently running the standard stock TRL GRPO loss instead of the user's requested objective (`gspo`, `dapo`, `dr_grpo`, `bnpo`, `two_sided`, `rft`).

**Fix:**
- In `_compute_variant_loss`, when `per_token_logps` is not pre-populated in `inputs`, reconstruct `input_ids` and `attention_mask` and compute `per_token_logps` in a single forward pass via `self._get_per_token_logps_and_entropies(model, ...)`.
- Fall back to `per_token_logps.detach()` for `old_per_token_logps` when missing (matching TRL's single-iteration behavior).
- Route loss computation through `_compute_loss` and `compute_loss` with gradient accumulation scaling (`current_gradient_accumulation_steps`).
- Preserves single-forward-pass efficiency and peak VRAM without redundant forward passes.
- `grpo_variant: standard` delegates directly to stock TRL loss without warning.
- Add `TestGRPOVariantRuntimeContract` with 7 regression tests in `tests/test_v05311.py`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

### DIFF
--- a/src/soup_cli/trainer/grpo.py
+++ b/src/soup_cli/trainer/grpo.py
@@ -76,34 +76,64 @@ def _make_grpo_trainer_variant_cached(base_cls: type, variant: str) -> type:
                 reason,
             )
 
-        def compute_loss(self, model, inputs, return_outputs=False, **kwargs):
-            # v0.53.11 review fix (code-review HIGH) — read kernel inputs
-            # FIRST and only call super() as a fallback. The previous
-            # ordering ran an extra forward pass on every step that was
-            # discarded when the kernel produced a loss, doubling VRAM
-            # peak. The TRL trainer stores per-batch tensors on ``inputs``
-            # by the time compute_loss is called, so we can probe them
-            # without burning a forward pass.
-            #
-            # Probe TRL ≥0.9 attribute name ``per_token_logps`` — drop the
-            # earlier ``logits_to_keep`` probe (code-review HIGH fix:
-            # ``logits_to_keep`` is a position mask, not log-probs).
+        def _compute_variant_loss(self, model, inputs):
+            """Evaluate the requested GRPO variant loss on the batch.
+
+            Reuses TRL's ``_get_per_token_logps_and_entropies`` to obtain
+            per-token log probabilities in a single forward pass without
+            duplicating forward computation or allocating extra VRAM.
+            """
             logp_new = _read_attr(inputs, "per_token_logps")
-            logp_old = _read_attr(inputs, "old_per_token_logps")
-            if logp_old is None:
-                logp_old = _read_attr(inputs, "ref_per_token_logps")
-            advantages = _read_attr(inputs, "advantages")
+            prompt_ids = _read_attr(inputs, "prompt_ids")
+            completion_ids = _read_attr(inputs, "completion_ids")
+            prompt_mask = _read_attr(inputs, "prompt_mask")
             completion_mask = _read_attr(inputs, "completion_mask")
+            advantages = _read_attr(inputs, "advantages")
+
+            if logp_new is None and prompt_ids is not None and completion_ids is not None:
+                import torch
+
+                if prompt_mask is None:
+                    prompt_mask = torch.ones_like(prompt_ids)
+                if completion_mask is None:
+                    completion_mask = torch.ones_like(completion_ids)
+                input_ids = torch.cat([prompt_ids, completion_ids], dim=1)
+                attention_mask = torch.cat([prompt_mask, completion_mask], dim=1)
+                logits_to_keep = completion_ids.size(1)
+
+                if hasattr(self, "_get_per_token_logps_and_entropies"):
+                    logp_new, _ = self._get_per_token_logps_and_entropies(
+                        model,
+                        input_ids,
+                        attention_mask,
+                        logits_to_keep,
+                        compute_entropy=False,
+                        pixel_values=_read_attr(inputs, "pixel_values"),
+                        image_grid_thw=_read_attr(inputs, "image_grid_thw"),
+                        num_images=_read_attr(inputs, "num_images"),
+                        pixel_attention_mask=_read_attr(inputs, "pixel_attention_mask"),
+                        image_sizes=_read_attr(inputs, "image_sizes"),
+                        token_type_ids=_read_attr(inputs, "token_type_ids"),
+                        mm_token_type_ids=_read_attr(inputs, "mm_token_type_ids"),
+                    )
+
+            logp_old = _read_attr(inputs, "old_per_token_logps")
+            if logp_old is None and logp_new is not None:
+                logp_old = logp_new.detach()
 
             if logp_new is None or logp_old is None or advantages is None:
-                # Fall back to the original loss — defence-in-depth so a
-                # TRL internal rename does not crash the training loop.
                 self._warn_fallback("missing per-token log-prob inputs")
-                return super().compute_loss(model, inputs, return_outputs=return_outputs, **kwargs)
+                return None
 
-            beta_attr = getattr(self.args, "beta", None)
+            beta_attr = getattr(getattr(self, "args", None), "beta", None)
             beta = float(beta_attr) if beta_attr is not None else 0.0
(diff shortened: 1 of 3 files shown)


src/soup_cli/trainer/grpo.py
```diff
@@ -76,34 +76,66 @@ def _warn_fallback(self, reason: str) -> None:
...
+                        image_sizes=_read_attr(inputs, "image_sizes"),
+                        token_type_ids=_read_attr(inputs, "token_type_ids"),
+                        mm_token_type_ids=_read_attr(inputs, "mm_token_type_ids"),
+                    )
+
             logp_old = _read_attr(inputs, "old_per_token_logps")
+            if logp_old is None and logp_new is not None:
+                logp_old = logp_new.detach()
             if logp_old is None:
                 logp_old = _read_attr(inputs, "ref_per_token_logps")
```
AmirF194: I ran this against 72e7383 in a clean container, spying on the `logp_old` value `apply_variant_loss` actually receives.

`logp_old` never resolves through the `ref_per_token_logps` branch at grpo.py:123-124. It only assigns there when `logp_old` is still `None` after the new `logp_new.detach()` fallback on line 121-122, which requires `logp_new` to also be `None`. The very next check, `if logp_new is None or logp_old is None or advantages is None`, then returns `None` regardless of what that branch just set. So the line is dead in every input combination, not just the one I ran:

```
inputs: per_token_logps set, old_per_token_logps absent, ref_per_token_logps = sentinel [[9,9,9]]
logp_old passed to apply_variant_loss: tensor([[0.1000, 0.2000, 0.3000]])
equals detach(per_token_logps)? True
equals ref_per_token_logps (would mean the old branch fired)? False
```

Lines 123-124 can be deleted. `ref_per_token_logps` is already read again on line 133 as `reference_logp` for the KL term, so nothing is lost.

here-2007:
Good catch. You’re completely right. Since `logp_old = logp_new.detach()` already comes before this, the `ref_per_token_logps` assignment to `logp_old` was unnecessary. `ref_per_token_logps` is already read separately and passed as `reference_logp` for the KL divergence calculation.

I removed those lines in commit [c21390d](https://github.com/MakazhanAlpamys/Soup/commit/c21390d015991e00551f6db0b3e999064b65bb15). All 216 tests and the Ruff lint checks are passing cleanly.

AmirF194:
Confirmed the fix removes the dead branch cleanly. Thanks for the quick turnaround.

MakazhanAlpamys (approved):
Re-reviewed current head 33977ac4 against issue #584. The variant wrapper now reconstructs prompt/completion inputs, obtains per-token log probabilities through TRL in one forward pass, derives detached old log-probabilities, and routes every configured objective through `apply_variant_loss` with a truthful fallback warning. The existing tests remain in place and the new runtime-contract tests cover realistic batches, all variants, standard delegation, and distinct losses; the changelog fragment is present.

Independent verification:
- pytest tests/test_v05311.py -k "GRPOTrainerVariantFactory or (GRPOVariantRuntimeContract and not real_trl)" --no-cov -q (14 passed)
- Ruff on the changed trainer/test files and git diff --check (passed)
- mutation: disabling the prompt/completion reconstruction made 9 focused tests fail, so the suite catches removal of the main fix.

All 13 checks reported by `gh pr checks 585` are green on 33977ac4. Approving for squash merge.

MakazhanAlpamys:
Thank you, @here-2007, for fixing the GRPO variant runtime contract and adding realistic-batch coverage across the supported objectives. I independently ran 14 focused tests, Ruff, and `git diff --check`; disabling prompt/completion reconstruction made 9 tests fail. All 13 current-head CI checks passed before the squash merge as `47b9121`.

PR #586, opened 2026-08-27
fix(ppo): honour quantize_reward_model instead of always quantizing the reward model

AmirF194:
## What does this PR do?

`_load_reward_model` (`ppo.py`) applies the policy model's Quant Menu config to the
reward model whenever a `tcfg` is passed, regardless of `training.quantize_reward_model`.
That flag was added in v0.53.0 with its own task-scoped validator
(`_validate_quantize_ref_reward`, "silent-no-op footgun rejection"), but no loader ever
reads it: `quantize_reward_model=False` (the default) does not stop the reward model from
inheriting quantization, and `=True` changes nothing either way.

Gates it on the flag instead: the quant config is only built and applied when
`tcfg.quantize_reward_model` is set. This follows the same shape as `double_quant_on`'s
#321 fix (`quant_menu.py`), which replaced a hardcoded `True` with the configured value.

## Type of change

- [x] Bug fix

## Checklist

- [x] `ruff check src/soup_cli/ scripts/ tests/` passes (scoped to the two changed files;
  full-repo run needs the `[dev]` extra, not installed here)
- [x] `pytest tests/ -v` passes on the files this change touches or exercises (see below);
  the full suite needs the `[dev]` extra (`torch`/`transformers`/`trl`/...), not installed
  in this environment
- [ ] Updated relevant docs: not needed, no public API/CLI surface changed
- [ ] Changelog fragment: added in a follow-up commit once this PR's number is known
  (`changelog.d/<release>/<number>.fixed.md`, per `changelog.d/README.md`)

## How I verified this

- New regression test, `TestPPORewardModelQuantMenu::test_load_reward_model_quantize_reward_model_false_skips_quant_menu`:
  builds a `TrainingConfig(quantization="gptq")` (the default `quantize_reward_model=False`)
  and asserts the Quant Menu loader is never called. Fails on `main` (loader called anyway),
  passes on this branch.
- Updated `test_load_reward_model_routes_through_quant_menu_live` to set
  `quantize_reward_model=True`, since it is specifically testing the flag-on path.
- Full `tests/test_v0405_part_a.py` (134 tests) and `tests/test_ppo.py` (51 tests) green,
  including the schema validator tests in `tests/test_v0530.py -k quantize` (10 tests).
- `tests/test_bugfixes.py`/`tests/test_trainer_init.py`: the same 11 tests fail on both
  `main` and this branch (missing `datasets`/`torch` in this lightweight environment,
  unrelated to this change), everything else green.
- Coverage on the changed lines in `_load_reward_model` (`--cov=soup_cli.trainer.ppo`):
  fully covered by the test suite above.
- Not run: the `[dev]`-extra test matrix (real `torch`/`transformers`/`trl`), and `mypy`
  (crashed with an internal error against the light install and I did not chase it further;
  the CI `type-check` job is non-blocking).

### DIFF
--- a/changelog.d/0.73.3/586.fixed.md
+++ b/changelog.d/0.73.3/586.fixed.md
@@ -1,4 +1,4 @@
-- PPO's reward model now only inherits the policy's Quant Menu quantization config when
-  `training.quantize_reward_model` is set (#586). Since v0.53.0 added that flag, it was
-  validated by its own task-scoped check but never read by any loader, so the reward
-  model always inherited quantization regardless of the flag's value.
+- PPO's reward model, and `task: reward_model`'s own trained model, now only inherit the
+  Quant Menu quantization config when `training.quantize_reward_model` is set (#586).
+  Since v0.53.0 added that flag, it was validated by its own task-scoped check but never
+  read by either loader, so both tasks quantized regardless of the flag's value.
--- a/changelog.d/0.73.3/586.fixed.md
+++ b/changelog.d/0.73.3/586.fixed.md
@@ -1,0 +1,4 @@
+- PPO's reward model now only inherits the policy's Quant Menu quantization config when
+  `training.quantize_reward_model` is set (#586). Since v0.53.0 added that flag, it was
+  validated by its own task-scoped check but never read by any loader, so the reward
+  model always inherited quantization regardless of the flag's value.
--- a/src/soup_cli/trainer/ppo.py
+++ b/src/soup_cli/trainer/ppo.py
@@ -783,10 +783,14 @@ def _load_reward_model(
     Reward models are typically AutoModelForSequenceClassification that output
     a scalar reward score for a given input sequence.
 
-    v0.40.5 #66: when ``tcfg`` is provided, the reward model is loaded with
-    the same quantization config as the policy model (Quant Menu). This keeps
-    the reward model from silently consuming full-precision VRAM and OOM-ing
-    when the policy is GPTQ/AWQ/HQQ/etc. Pass ``tcfg=None`` for unquantized.
+    v0.40.5 #66: when ``tcfg`` is provided and ``training.quantize_reward_model``
+    is set, the reward model is loaded with the same quantization config as
+    the policy model (Quant Menu), so it doesn't silently consume
+    full-precision VRAM and OOM when the policy is GPTQ/AWQ/HQQ/etc. v0.53.0
+    made this opt-in via that flag (default False) rather than unconditional,
+    matching ``double_quant_on``'s #321 precedent of honouring a configured
+    flag over a hardcoded default. Pass ``tcfg=None`` for unquantized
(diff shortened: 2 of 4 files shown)


MakazhanAlpamys (changes requested):
Thanks for the focused PPO regression and for proving the flag-off mutation. I found one blocking contract gap before this can merge.

training.quantize_reward_model is documented and schema-validated for both PPO and reward_model, but this PR only gates ppo.py::_load_reward_model. RewardModelTrainerWrapper._setup_transformers still unconditionally builds the Quant Menu config, so task: reward_model with quantize_reward_model: false is accepted and still quantizes. The flag therefore remains a silent no-op for one supported, documented task.

Please either wire the false/true behavior through the reward-model loader with a regression test, or deliberately narrow the schema validator and documentation with a rejection test. The current test is effective for PPO: locally reverting the guard made it fail, and the focused PPO suites passed; the missing reward-model path is the remaining blocker.

AmirF194:
Good catch, thank you. Pushed a fix: RewardModelTrainerWrapper._setup_transformers now gates both the quantization_config build and prepare_model_for_kbit_training on quantize_reward_model, mirroring _load_reward_model's PPO-path gate. task: reward_model with the flag off now loads unquantized regardless of training.quantization; with it on, it quantizes as before.

Added a regression test (TestRewardModelTaskQuantMenu) that fails on the prior code (proved locally by reverting just this file) and passes on the branch: quantize_reward_model=False asserts the loader is never called and quantization_config never lands in the from_pretrained kwargs; =True asserts both fire.

Full suite plus ruff clean on the changed files.

MakazhanAlpamys (approved):
Re-reviewed commit f34c6d7 after the follow-up. The fix now gates both PPO's reward-model loader and the task=reward_model loader on training.quantize_reward_model, including the k-bit preparation path. The changed test file keeps the existing coverage and adds explicit flag-off coverage; no tests were removed.

Independent verification:
- pytest tests/test_v0405_part_a.py -k "PPORewardModelQuantMenu or RewardModelTaskQuantMenu" --no-cov -q (7 passed)
- ruff check on both trainers and the focused test file (passed)
- git diff --check (passed)
- mutation: forcing the task loader guard on made test_quantize_reward_model_false_skips_quant_config fail (1 failed), so the regression test catches removal of the main fix.

The current GitHub head f34c6d7 has all 13 checks reported by `gh pr checks` passing.

MakazhanAlpamys:
Thank you, @AmirF194, for the focused quantization fix and the follow-up that closed the contract gap for `task: reward_model`. I verified the merged head with 7 focused tests, Ruff, and `git diff --check`; the mutation that forced the flag-off task guard on was caught by the new regression test. All 13 current-head CI checks passed before the squash merge as `76daa061`.

AmirF194:
Appreciate the thorough re-verification, and thanks for catching the reward-model contract gap in review, it made the fix complete rather than half done.

PR #587, opened 2026-08-28
fix(train): warn on unwired convergence detection (#583)

### LINKED ISSUE #583: training.convergence_detection is accepted but never enforced, and unlike its sibling flags it warns of nothing
## Description

`training.convergence_detection` (plus `convergence_window`, `convergence_rel_tol`) is documented in docs/peft-and-efficiency.md as surfacing `continue`/`early_stop`/`lower_lr` advice from the loss curve, and `src/soup_cli/utils/convergence.py` fully implements `detect_plateau`/`recommend_action`, with its own unit tests in `tests/test_auto_tuning.py`. Nothing in the codebase calls either function outside that module and its test file, so setting the flag does nothing during training.

train.py's honesty-guard block (added in b587dee) exists to warn when a documented knob is accepted but not enforced. It currently lists `forgetting_detection`, `checkpoint_intelligence` and `early_stop_on_regression`. `convergence_detection` has the same shape as those three, but it's missing from that list, so it silently no-ops instead of warning like its siblings do.

## Steps to reproduce

1. Clone at HEAD `1b12f7a04562026c03db79a0dfde4393aadc97b7`, `pip install -e .` (core deps only, the `[train]` extras aren't needed for this).
2. Write a config with `training.convergence_detection: true`, and for contrast `training.forgetting_detection: true`.
3. `soup train --config soup.yaml --dry-run`

## Config (soup.yaml)

```yaml
base: sshleifer/tiny-gpt2
output: /tmp/out
experiment_name: verify-run
data:
  train: /tmp/fake.jsonl
training:
  forgetting_detection: true
  convergence_detection: true
  convergence_window: 50
  convergence_rel_tol: 0.005
```

## Error output

No traceback, that is the bug. The run prints `Note: forgetting_detection are set but not enforced during training in this build. ...` for the sibling flag, but nothing at all for `convergence_detection`, and training proceeds as if it were unset.

## Environment

Verified in a clean `python:3.11-slim` container with only the core deps installed (typer, rich, pydantic, pyyaml, huggingface-hub, plotext, packaging), CPU only. I did not install the `[train]` extras for this repro, so I could not run `soup doctor`, happy to if it would help narrow this further.

Confirmed by grepping `src/` that `detect_plateau` and `recommend_action`, the two functions the documented behavior depends on, have zero callers outside `utils/convergence.py` and `tests/test_auto_tuning.py`.

Happy to send a PR either way: add `convergence_detection` to the honesty-guard's `_unwired_gates` list (small, matches the pattern of the other three), or wire `recommend_action` into the monitoring callback so the documented feature actually runs (bigger, since the feature is already fully built and tested).

MakazhanAlpamys:
## Summary

- Include training.convergence_detection in the existing train-command honesty guard.
- Add an end-to-end dry-run regression test for the user-visible warning.
- Document that the detector remains advisory until a live callback wires it in, and add a changelog fragment.

Fixes #583.

## Validation

- PYTHONPATH=src pytest tests/test_code_review_critical.py tests/test_training_intelligence.py --no-cov -q — 58 passed
- 
uff check src/soup_cli/ tests/ — passed
- git diff --check — passed
- Mutation: removing convergence_detection from the guard made the new regression test fail.

### DIFF
--- a/src/soup_cli/commands/train.py
+++ b/src/soup_cli/commands/train.py
@@ -724,6 +724,7 @@ def train(
             ("forgetting_detection", cfg.training.forgetting_detection),
             ("checkpoint_intelligence", cfg.training.checkpoint_intelligence),
             ("early_stop_on_regression", cfg.training.early_stop_on_regression),
+            ("convergence_detection", cfg.training.convergence_detection),
         )
         if on
     ]
--- a/changelog.d/0.73.3/583.fixed.md
+++ b/changelog.d/0.73.3/583.fixed.md
@@ -1,0 +1,3 @@
+`soup train` now warns when `training.convergence_detection` is enabled but not
+yet wired into the live training loop, matching the existing honesty guard for
+other advisory training-intelligence flags.
--- a/changelog.d/0.73.3/583.fixed.md
+++ b/changelog.d/0.73.3/583.fixed.md
@@ -1,3 +1,3 @@
-`soup train` now warns when `training.convergence_detection` is enabled but not
+- `soup train` now warns when `training.convergence_detection` is enabled but not
 yet wired into the live training loop, matching the existing honesty guard for
-other advisory training-intelligence flags.
+other advisory training-intelligence flags (#583).
--- a/docs/peft-and-efficiency.md
+++ b/docs/peft-and-efficiency.md
@@ -578,7 +578,9 @@ training:
   convergence_rel_tol: 0.005  # Relative range below this == plateau
 ```
 
-Surfaces `continue` / `early_stop` / `lower_lr` advice based on the loss curve.
+Computes `continue` / `early_stop` / `lower_lr` advice from the loss curve for
+callers that invoke the detector. `soup train` currently reports this option as
+not enforced; a live training callback remains a follow-up.
 
 ### VRAM Pressure Advisory
 
--- a/tests/test_code_review_critical.py
+++ b/tests/test_code_review_critical.py
@@ -178,6 +178,43 @@ def test_eval_benchmark_rejects_trust_remote_code_injection(tmp_path, monkeypatc
     assert "trust_remote_code" in result.output or "delimiters" in result.output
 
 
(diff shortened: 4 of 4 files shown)


src/soup_cli/commands/train.py
```diff
@@ -724,6 +724,7 @@ def train(
             ("forgetting_detection", cfg.training.forgetting_detection),
             ("checkpoint_intelligence", cfg.training.checkpoint_intelligence),
             ("early_stop_on_regression", cfg.training.early_stop_on_regression),
+            ("convergence_detection", cfg.training.convergence_detection),
```
AmirF194, line 727: I checked how `soup autopilot` populates this list, against the comment above it, on the PR head 292de25e.

`src/soup_cli/autopilot/generate_config.py:110-112` sets `forgetting_detection`, `checkpoint_intelligence`, and `early_stop_on_regression` to `True`, but never sets `convergence_detection`. Its schema default is `False` (`config/schema.py:2976`). So after this line, the comment at `train.py:717-718`, "these knobs are accepted (and `soup autopilot` turns them on by default)", is no longer true for all four items in the tuple: an autopilot-generated config will never hit this warning for `convergence_detection`, only a hand-written config will.

```
$ grep -n "convergence_detection" src/soup_cli/autopilot/generate_config.py
(no output)
$ sed -n '2976,2977p' src/soup_cli/config/schema.py
    convergence_detection: bool = Field(
        default=False,
```

If autopilot is meant to enable this flag like its three siblings, add `convergence_detection=True,` after line 112. If not, the comment at 717-718 should be reworded so it does not claim all four are autopilot-enabled.

MakazhanAlpamys:
Thank you, @AmirF194, for reporting the unwired convergence-detection contract. The warning-only fix is now merged as `439c798`; I verified the focused training-intelligence tests, Ruff, `git diff --check`, and the mutation that removes convergence_detection from the honesty guard. All 13 current-head CI checks passed before merge.

AmirF194:
Thanks for the quick fix.

PR #588, opened 2026-08-28
fix(train): make FSDP QLoRA dtype-compatible

### LINKED ISSUE #350: Recipe llama3-70b-fsdp2 cannot run: FSDP + BNB 4-bit needs quant_storage AND a bf16 adapter, and only one is expressible in config
Measured on 8x H100 80GB, torch 2.13.0+cu130 / transformers 4.57.6 / peft 0.20.0 / bitsandbytes 0.50.0, soup 0.72.4. Full record: `benchmarks/gate-h100-validation.md`, STEP 28.

`llama3-70b-fsdp2` has never been executed by anything — `test_multi_gpu.py` checks its YAML parses. It does not run.

Model substitution: `meta-llama/Llama-3.1-70B-Instruct` is gated and the box had no HF token, so the arms below use `Qwen2.5-72B-Instruct` at the recipe's exact shape (4-bit base, FSDP full-shard, `use_fsdp2_compile`, LoRA r16, `max_length` 4096, batch 1 x accum 8, 8 ranks).

## What happens

| arm | change from the recipe | outcome |
|---|---|---|
| as shipped | — | `ValueError: Cannot flatten integer dtype tensors` |
| control | `use_fsdp2_compile: false` | **identical error** |
| + storage | `bnb_4bit_quant_storage: bfloat16` | `ValueError: Must flatten tensors with uniform dtype but got torch.bfloat16 and torch.float32` |
| + dtype | 723 fp32 params cast to bf16 after `get_peft_model` | **trains**, 100 steps, exit 0 |

All eight ranks load the checkpoint, print `Training started!`, reach 46 448.6 MiB allocated, and then FSDP refuses to flatten a unit containing bitsandbytes' packed `uint8` storage.

**The `use_fsdp2_compile: false` control gets the same error**, so `torch.compile` is not implicated — the blocker is FSDP + 4-bit, which the recipe pairs by design.

## The two changes

1. `bnb_4bit_quant_storage: bfloat16`. Soup already has this knob, and `utils/quant_menu.py:476` already warns that FSDP + BNB 4-bit without it "causes all-gather to upcast to fp32". That understates it: it is a hard stop, not a performance foot-gun.
2. The adapter's dtype. After `get_peft_model` the model has `{'torch.float32': 723, 'torch.bfloat16': 560}`; FSDP wants one dtype per unit. Casting the 723 to bf16 makes it train. **There is no config surface for this today.**

## Suggested fix

- Set `bnb_4bit_quant_storage` to the compute dtype automatically when `quantization: 4bit` meets an FSDP preset, instead of warning.
- Cast newly created LoRA parameters to the compute dtype on the FSDP path (or expose `lora.dtype`).
- Update the three multi-GPU recipes to whatever combination is verified.

## Acceptance criteria

- [ ] `soup train --config <llama3-70b-fsdp2 recipe> --gpus 8 --fsdp full_shard` reaches step 1 with no config edits
- [ ] A test asserts the resolved quant config carries a non-integer `bnb_4bit_quant_storage` whenever FSDP + 4-bit are combined
- [ ] A test asserts no fp32 floating parameter survives `get_peft_model` on the FSDP + 4-bit path
- [ ] The `use_fsdp2_compile: false` control is covered too, since it fails identically

Faisal01011:
Closes #350

## What

- resolve BNB 4-bit quant storage to the effective floating compute dtype whenever `--fsdp` is active
- align every trainable floating PEFT parameter to that dtype after trainer setup and before FSDP wrapping
- pin bf16 storage in the 70B FSDP2 recipe and document the automatic behavior
- replace the old speed-warning behavior, which understated a hard runtime failure

## Verification

- 131 focused tests passed, 1 platform-dependent test skipped
- new dependency-light regression coverage verifies storage resolution, the emitted BitsAndBytesConfig, adapter/module-to-save casting, non-FSDP controls, compile on/off, central command wiring, and the recipe

## Hardware limitation

I do not have an 8×H100 environment. The CPU-testable policy and wiring are covered here, but the acceptance run `soup train --config <llama3-70b-fsdp2> --gpus 8 --fsdp full_shard` still needs maintainer hardware validation through step 1.

### DIFF
--- a/changelog.d/0.73.3/588.fixed.md
+++ b/changelog.d/0.73.3/588.fixed.md
@@ -1,0 +1,5 @@
+- [#588](https://github.com/MakazhanAlpamys/Soup/pull/588) makes FSDP + BNB
+  4-bit training resolve quantized storage and trainable adapter
+  parameters to the same floating compute dtype, preventing the integer-storage
+  and mixed-dtype flattening failures that made the `llama3-70b-fsdp2` recipe
+  unrunnable. The recipe now pins bf16 storage for its A100/H100 target hardware.
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -172,7 +172,8 @@ base: TheBloke/Llama-2-7B-Chat-GPTQ
 training:
   quantization: gptq        # or: awq, hqq:4bit, aqlm, eetq, mxfp4, fp8
 
-# FSDP + QLoRA — set quant_storage:
+# FSDP + QLoRA — Soup resolves this to the compute dtype automatically;
+# pin it explicitly when the recipe targets known bf16 hardware:
 training:
   quantization: 4bit
   bnb_4bit_quant_storage: bfloat16
@@ -193,8 +194,11 @@ training:
 
 **Compatibility matrix.** `soup train` runs `check_quant_distributed_compat()` at
 startup. HQQ / EETQ / AQLM hard-fail with FSDP and ZeRO-3 (sourced from
-LlamaFactory's matrix at `quantization.py:199/211`); BNB 4-bit + FSDP without
-`bnb_4bit_quant_storage` emits a yellow warning.
+LlamaFactory's matrix at `quantization.py:199/211`). BNB 4-bit + FSDP resolves
+`bnb_4bit_quant_storage` to the effective floating compute dtype before model
+load, then aligns trainable adapter parameters to that dtype before FSDP wraps
+the model. This avoids both FSDP failure modes: integer storage and mixed
+adapter/storage dtypes.
 
 **Pre-quantized + QAT.** `gptq` / `awq` / `hqq:*` / `aqlm` / `eetq` / `mxfp4` /
 `fp8` all carry their own scale; combining with `quantization_aware` (int8 QAT or
--- a/src/soup_cli/commands/train.py
+++ b/src/soup_cli/commands/train.py
@@ -836,6 +836,25 @@ def train(
         fsdp_kwargs = get_fsdp_training_args(fsdp)
         console.print(f"[green]FSDP2 enabled:[/] {fsdp}")
 
+    # #350 — BNB's default uint8 quant storage is not merely slow under FSDP:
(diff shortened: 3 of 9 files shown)


MakazhanAlpamys (changes requested):
The software implementation and regression coverage look correct, and all 14 current-head CI checks are green. I independently verified the focused dtype/FSDP tests, Ruff, `git diff --check`, and a mutation of the alignment guard; the mutation failed the regression assertion.

One issue-level acceptance item remains unresolved: #350 explicitly requires a real `llama3-70b-fsdp2` run on 8x H100 through step 1, while this PR body says that hardware run has not been performed. Because the PR currently declares `Closes #350`, this is blocking the merge. Please either provide the maintainer hardware validation, or change the linkage to `Refs #350` and leave the issue open for that integration evidence.

MakazhanAlpamys:
Follow-up on the acceptance boundary for #350: the focused regression suite and current-head CI are green, but the issue requires a real llama3-70b-fsdp2 run on 8xH100 through step 1. Please either attach that validation with the exact commit and config, or change the PR linkage and body from “Closes #350” to “Refs #350” and leave the issue open for the hardware run. Please request re-review after updating.

MakazhanAlpamys (approved):
Maintainer decision: the requested 8xH100 environment is not available in this cycle. The focused regression suite, mutation check, Ruff, diff check, and all required CI checks are green, so I am accepting the implementation and merging it with live hardware validation explicitly deferred. If a later run finds a defect, we will open a follow-up issue with the measured reproduction.

MakazhanAlpamys:
Thank you, @Faisal01011, for the focused FSDP + BNB 4-bit compatibility fix. The implementation resolves quant storage to the effective compute dtype, aligns trainable PEFT parameters before FSDP wrapping, and includes recipe, documentation, and changelog coverage. I independently ran 75 focused regression tests, Ruff, and git diff --check; the mutation that disabled FSDP-QLoRA alignment was caught by test_fsdp_qlora_casts_every_trainable_float_to_compute_dtype. All 14 required CI checks passed on the reviewed head 279fc011. Live 8xH100 validation for #350 is explicitly deferred because the hardware is unavailable.

Issue #379, opened 2026-08-10
Publish the measurement harnesses: the repo cannot currently reproduce its own benchmark records


MakazhanAlpamys:
## What

`benchmarks/gate-h100-validation.md` closes with a "Reproducing" section that lists roughly twenty harness scripts — and then says they **live in the session scratchpad, not in the repo** (L4908–4939):

```
bitexact.py  graddiff.py  determinism.py  repeat_backward.py  layercount.py
depth_vs_bytes.py  ckpt_hypothesis.py  mechanism.py  mechanism_cost.py  pincost.py
prep_convergence.py  runbench.sh  variance.sh  runbench8.sh  numerics.py  numerics2.py
forcecheck.py  fixtureshape.py  fixture_window.py  cpu_mode_probe.py  variant2_gate.py
bnb_repro.py  issue328_min.py  issue328_probe.py  issue328_control.py  measure327.py
```

**Nothing in the repository reproduces the record.** The same is true of the four earlier gate records in `benchmarks/`.

## Why this is the largest credibility gap in the whole harvest

The project's entire public position is *check it yourself*. `README.md` points at `benchmarks/` as the measurement record behind the preprint (DOI 10.5281/zenodo.21771064). A reader who wants to verify a number — the 256/256-vs-8/256 repair gate, the 6.56x pinning cost, the 163.8–171.5 MiB threshold bracket — currently cannot, because the code that produced it does not exist publicly.

Every one of those numbers is reproducible in principle: `bnb_repro.py` reproduces the NF4 defect **in about a minute on a synthetic 7.7 GB model with no download**.

## Proposal

`benchmarks/harness/` with:

- the scripts as run, not cleaned up,
- a README mapping **script → STEP → the claim it produced**,
- the hardware and stack each was run against,
- and the deliberately-vacuous ones **kept with their vacuity documented** — `numerics.py` measured at M=2048, above `_gemm_4bit_custom_max_m`, so both arms ran the same fallback and it compared variant 2 against variant 2. That is exactly the kind of artefact the records already preserve in prose; deleting it from the code while keeping it in the text would be inconsistent.

All harnesses `import soup_cli` unmodified, which is worth stating explicitly: every measurement above the record's measurement/fix boundary was taken against untouched Soup.

## Acceptance criteria

- [ ] Scripts published under `benchmarks/harness/`, indexed in `benchmarks/README.md`
[...]

PR #589, opened 2026-08-28
bench: publish bitsandbytes reproduction harness

Samearth17:
## Summary

First reproducibility-only slice for #379.

This publishes the standalone bitsandbytes #2034 NF4 reproduction harness under `benchmarks/harness/` and indexes it in `benchmarks/README.md`.

The harness preserves the experiment with NF4 recycled buffers, NF4 private buffers as the reference, and bf16 recycled buffers as the control, using gradient checkpointing and no model downloads.

No benchmark methodology or underlying Soup implementation was changed.

## Validation

- `python -m ruff check benchmarks/harness/bnb_repro.py` — passed
- `python -m pytest -q` — 18,901 passed, 283 skipped, 5 deselected
- `git diff --check` — passed

The harness was not executed end-to-end locally because the development machine has no CUDA GPU; it exits cleanly with:

`SKIP: CUDA is required for bnb_repro.py`

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -34,9 +34,10 @@ on trust. It starts small on purpose — most of this session's ~20 harnesses li
 only in a scratchpad on a machine that is gone, which is
 [#379](https://github.com/MakazhanAlpamys/Soup/issues/379).
 
-| Script | Question it answers | Cost |
-|---|---|---|
-| [`issue331_qlora_scope.py`](harness/issue331_qlora_scope.py) | Does the #331 wrong-gradient defect reach **ordinary QLoRA**? Three arms in one process, with the positive control that makes an exact result mean something. Answer: no — 0.0 against a control that diverges by 3.77e-01 | ~15 s, 4 GB card, no downloads |
+| Script | STEP / record | Question or claim | Cost / requirements |
+|---|---|---|---|
+| [`issue331_qlora_scope.py`](harness/issue331_qlora_scope.py) | #331 | Does the wrong-gradient defect reach **ordinary QLoRA**? Three arms in one process, with a positive control. Answer: no — 0.0 against a control that diverges by 3.77e-01 | ~15 s, 4 GB card, no downloads |
+| [`bnb_repro.py`](harness/bnb_repro.py) | Upstream bitsandbytes #2034 / NF4 defect | Minimal standalone reproducer for the NF4 stale-gradient mechanism: recycled buffers expose `MatMul4Bit`'s `ctx` lifetime issue; private buffers are the reference and bf16 is the control | CUDA GPU required; no downloads; historical environment: torch 2.13.0+cu130, bitsandbytes 0.50.0; ~1 min |
 
 ## Hardware
 
--- a/benchmarks/harness/bnb_repro.py
+++ b/benchmarks/harness/bnb_repro.py
@@ -1,0 +1,174 @@
+"""Standalone reproducer for bitsandbytes issue #2034.
+
+This is the upstream minimal reproducer used to demonstrate that
+``MatMul4Bit`` keeps the packed weight and quantization state on the
+autograd context as ordinary attributes instead of saved tensors.
+
+Requirements
+------------
+- CUDA GPU required.
+- No model download is required.
+- Historical environment: torch 2.13.0+cu130, bitsandbytes 0.50.0.
+- The original reproducer uses one CUDA device and takes about one minute.
+
+Protocol
+--------
+Compare three arms using gradient checkpointing:
+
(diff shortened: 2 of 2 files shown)


MakazhanAlpamys (changes requested):
The first-slice scope matches issue #379 and the added harness/index are clean, but the mandatory mutation check found a blocking gap: disabling the main `pool.acquire(...)` mechanism still produced all `ok` comparisons and `EXIT_CODE=0`. The harness therefore does not fail when its core recycled-buffer path is removed.

Please add a negative control that proves the harness detects the NF4 recycled-buffer mismatch (for example, mutate or bypass the pool/recycled allocation and assert a non-zero result or an explicit mismatch), while keeping the reference and bf16 control arms meaningful. The draft can remain focused on `bnb_repro.py`; no broader harness work is required for this fix.

MakazhanAlpamys:
Follow-up on the requested mutation: the current harness still exits successfully when recycled-buffer pool acquisition is bypassed, so the main fix is not distinguished from the broken behavior. Please add a negative control that disables or bypasses pool reuse and asserts a non-zero result or a detected mismatch. Keep the PR draft until that mutation is caught; then refresh the current-head checks and request re-review.

MakazhanAlpamys (changes requested):
Re-review of f318bc54c10789251190be8d73ae8676d29fa02e: the first-slice scope matches issue #379, but the mandatory mutation check is still a blocker. Removing pool.acquire() makes the script exit 0 with a CUDA skip in the current environment, and there is no executable test or CI path that kills this mutation. The harness can report success before exercising the mechanism it claims to reproduce. Keep the draft open and add a runnable regression, such as an explicit CUDA job, that runs the normal and mutated pool paths and asserts a non-zero result or detected mismatch. The current CI run is now unblocked but still in progress.

MakazhanAlpamys:
Merged as `25d17c0`. Thank you, @Samearth17 — and an apology for the delay: you pushed `f318bc54` at 00:45 UTC and my blocking comment above was left against the older head, so it sat there looking unanswered when it was already fixed. That is on me.

**I ran it on real CUDA, which you said you could not.** Here is the full chain, because it is the whole point of the review round:

```
normal run                        exit 0
                                  NF4 mismatch on 6/8 layers
                                  bf16 control clean on 8/8

--bypass-pool                     exit 1
                                  "NEGATIVE CONTROL: bypassing pool reuse
                                   removed the NF4 mismatch"

pool.acquire() cut from source    exit 1
                                  "ERROR: expected NF4 recycled/private
                                   mismatch was not reproduced"

same cut on pre-fix 373c007e      exit 0    <-- the defect I blocked on
```

That last line is what makes the fix real rather than claimed: the identical mutation exits `0` on the commit I reviewed and `1` on the one I merged. Box: RTX 3050 Laptop, torch 2.5.1+cu121, bitsandbytes 0.49.2.

You fixed it in **both** directions rather than only the one I asked for. I asked for a negative control; you also made the ordinary invocation require the mismatch. That second half is the one that keeps working when nobody remembers to pass the flag — a harness whose default path exits `0` on a healthy machine and a broken one alike is not evidence, and yours no longer does.

Two things worth recording that you could not have known:

1. **It reproduces on a much older stack than the README row states.** The row says the historical environment was torch 2.13.0+cu130 / bitsandbytes 0.50.0; it reproduces on 2.5.1+cu121 / 0.49.2 as well. If you want a follow-up, widening that row to say the mechanism is not specific to the recorded stack would strengthen it.
2. **6/8 layers, not 8/8.** Layers 6 and 7 come out at exactly `0.000000e+00` on my card. That is consistent with the pool having 6 slots and the last two layers still holding live buffers, but I have not confirmed the mechanism — worth a sentence in the docstring if you can establish it, and worth *not* asserting if you cannot.

Diff integrity: nothing deleted. The `issue331_qlora_scope.py` row is preserved verbatim; the table only gains STEP/record and requirements columns. No changelog fragment needed — `9ce6fa2`, which added the first harness, carried none either, and this changes nothing a user runs.

`Refs #379`, not `Closes` — correct, and `closingIssuesReferences` is empty so the issue stays open for the remaining harnesses. That is the right shape for a first slice, and stating up front that you had no GPU rather than implying otherwise is what let me spend the review time on the one real question instead of re-deriving what you had checked.

#379 is still open with more harnesses to publish if you want the next slice.

Samearth17:
Thanks for the thorough review and for validating the harness on real CUDA. The mutation comparison between the pre-fix and fixed commits was especially helpful, and I am glad we were able to make the regression guard robust.

I would also be happy to take up the next slice of #379. I will look through the remaining harnesses and identify a focused next contribution.

PR #591, opened 2026-08-29
fix(train): raise accelerate floor to keep FSDP+PEFT checkpoints adapter-only

### LINKED ISSUE #352: A LoRA run under FSDP writes a 37 GB consolidated base model into every checkpoint (adapter is 131 MB)
Measured at 70B on 8x H100 while running #41. Full record: `benchmarks/gate-h100-validation.md`, STEP 28.

A `use_fsdp2_compile` + LoRA r16 run on Qwen2.5-72B wrote this into a single checkpoint directory:

```
adapter_model.safetensors        131 118 496   (131 MB — the thing being trained)
optimizer.bin                    262 476 043
pytorch_model_fsdp.bin        36 981 387 757   (37 GB — a full consolidated copy of the frozen base)
rng_state_0..7.pth
```

The base is frozen; 0.18% of the model is being trained. Writing a full copy of it per checkpoint is 280x the useful artifact.

## Why it matters

On a box with 250 GB free that is seven checkpoints from a full disk, and a full disk stops the run rather than degrading it. It also makes `save_steps` effectively unusable on any large model under FSDP: a user who sets a sensible cadence fills the volume.

## Cause

Accelerate's FSDP state-dict handling writes a `FULL_STATE_DICT` for the wrapped model alongside the PEFT adapter. HF's `save_only_model=True` skips optimizer/scheduler/RNG but not this; the relevant knob is the FSDP state-dict type, or not saving the base at all for a PEFT model.

## Suggested fix

When the model is a `PeftModel` and the base is frozen, do not write the base into checkpoints — the adapter plus config is the complete artifact. Failing that, default `save_only_model=True` on the FSDP path and document it.

## Acceptance criteria

- [ ] A LoRA + FSDP checkpoint directory contains no full-model file
- [ ] Test asserts checkpoint size is O(adapter), not O(base), for a PEFT model under an FSDP preset
- [ ] Resume still works from such a checkpoint

AmirF194:
## Root cause

`Trainer._save_checkpoint` writes two independent things under FSDP: `save_model()` (adapter-only for a `PeftModel`, correctly) and, unless `save_only_model=True`, `_save_optimizer_and_scheduler()`, which calls `accelerate.utils.save_fsdp_model()` to write a separate resume checkpoint (`pytorch_model_fsdp.bin`).

accelerate 0.25.0/0.26.0's `save_fsdp_model`/`load_fsdp_model` have no `adapter_only` parameter, so that resume checkpoint always gathers `model.state_dict()` in full, frozen base included, regardless of the model being a `PeftModel`. accelerate 0.27.0 added `adapter_only`, and `transformers.trainer.get_fsdp_ckpt_kwargs()` (present in the 5.16.1 floor this project already pins) feature-detects the parameter and threads it through on both save and load. `pyproject.toml`'s own `accelerate>=0.25.0` floor is loose enough to resolve a version that predates the fix.

## Fix

Raise the `accelerate` floor to `>=0.27.0` in `pyproject.toml`. No `soup_cli` source code changes: the adapter-only behavior is entirely inside `transformers`/`accelerate`, gated purely on the installed `accelerate` version.

## Verification

- Read `transformers==5.16.1`'s `Trainer._save_checkpoint`/`_save_optimizer_and_scheduler`/`get_fsdp_ckpt_kwargs` and `accelerate`'s `save_fsdp_model`/`load_fsdp_model`/`_get_model_state_dict` source to trace the exact write path for `pytorch_model_fsdp.bin`, and bisected accelerate releases in Docker: `save_fsdp_model` gained `adapter_only` between 0.26.0 (absent) and 0.27.0 (present).
- Added `tests/test_issue352_fsdp_peft_checkpoint.py`, run in Docker against this project's own `.github/constraints/transformers-floor.txt` pins (torch 2.5.1, transformers 5.16.1, trl 0.29.0, peft 0.20.0): a tiny `PeftModel` (2048x2048 frozen base, rank-4 LoRA adapter) shows the FSDP resume state dict is adapter-only (16,384 params) versus the full state dict (4,210,688 params). The `pyproject.toml`-floor assertion fails on unmodified `main` and passes on this branch; the library-behavior assertions exercise the real installed `accelerate`/`peft`, not mocked.
- Ran the full `pytest -o addopts= --no-cov -q` suite under the floor constraints plus every FSDP/PEFT/accelerate-adjacent test file (1008 passed, 26 pre-existing failures reproduced identically on unmodified `main`, unrelated to this change, 19 skipped), and `ruff check src/soup_cli/ scripts/ tests/` clean.
- Not run: an actual multi-GPU FSDP training job. The fix is a dependency-floor bump with no `soup_cli` runtime code path of its own, so I verified the exact library functions that decide what gets written rather than an end-to-end 8-GPU checkpoint.

Fixes #352

### DIFF
--- a/changelog.d/0.73.3/591.fixed.md
+++ b/changelog.d/0.73.3/591.fixed.md
@@ -1,0 +1,4 @@
+- A LoRA/PEFT run under FSDP no longer writes a full consolidated copy of the
+  frozen base model into every checkpoint. Raised the `accelerate` floor to
+  0.27.0, the first release whose FSDP checkpoint save/load path can be
+  restricted to the trainable adapter (#352, #591).
--- a/changelog.d/0.73.3/591.fixed.md
+++ b/changelog.d/0.73.3/591.fixed.md
@@ -1,4 +1,5 @@
-- A LoRA/PEFT run under FSDP no longer writes a full consolidated copy of the
-  frozen base model into every checkpoint. Raised the `accelerate` floor to
-  0.27.0, the first release whose FSDP checkpoint save/load path can be
-  restricted to the trainable adapter (#352, #591).
+- Raised the `accelerate` floor to 0.27.0, the first release whose FSDP
+  checkpoint save/load path can be restricted to the trainable adapter. This
+  closes a remaining way a LoRA/PEFT run under FSDP could regress back to
+  writing the full frozen base model into every checkpoint on an unpinned
+  `accelerate` install (#352, #591).
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,7 +105,10 @@ train = [
     "trl>=0.29.0,<1.0.0",
     "datasets>=2.14.0",
     "bitsandbytes>=0.41.0",
-    "accelerate>=0.25.0",
+    # #352: 0.27.0 added `adapter_only` to `save_fsdp_model`/`load_fsdp_model`, which
+    # transformers 5.16.1 feature-detects to keep a PEFT run's FSDP resume checkpoint
+    # to the adapter instead of the full frozen base.
+    "accelerate>=0.27.0",
 ]
 # v0.71.0 — convenience meta-extra pulling the main optional stacks.
 all = ["soup-cli[train,serve,ui,data,mcp]"]
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -22,7 +22,7 @@ DEPS = [
     ("trl", "trl", "0.29.0", True),
     ("datasets", "datasets", "2.14.0", True),
     ("bitsandbytes", "bitsandbytes", "0.41.0", True),
-    ("accelerate", "accelerate", "0.25.0", True),
+    ("accelerate", "accelerate", "0.27.0", True),
     ("pydantic", "pydantic", "2.0.0", True),
     ("typer", "typer", "0.9.0", True),
     ("rich", "rich", "13.0.0", True),
--- a/tests/test_issue352_fsdp_peft_checkpoint.py
+++ b/tests/test_issue352_fsdp_peft_checkpoint.py
@@ -1,0 +1,72 @@
+"""#352: a LoRA run under FSDP must not write the frozen base into checkpoints.
+
+`Trainer._save_optimizer_and_scheduler` writes a separate FSDP resume checkpoint
+(`pytorch_model_fsdp.bin`) via `accelerate.utils.save_fsdp_model`, independently of the
+adapter-only save `Trainer.save_model` already produces. Accelerate 0.25.0/0.26.0 has no
+`adapter_only` parameter on that function, so the resume checkpoint always gathers the
+full (frozen base + adapter) state dict regardless of the model being a `PeftModel`.
+0.27.0 added `adapter_only`, and `transformers.trainer.get_fsdp_ckpt_kwargs()`
+feature-detects it and threads it through on both save and load. This pins the accelerate
+floor bump in `pyproject.toml` to the real state-dict-size mechanism the issue reports,
+rather than to the dependency string alone.
+"""
+
+from __future__ import annotations
+
+import inspect
+import re
+from pathlib import Path
+
+
+def test_pyproject_accelerate_floor_supports_adapter_only_fsdp_checkpoints():
+    pyproject = Path(__file__).parent.parent / "pyproject.toml"
+    match = re.search(r'"accelerate>=([0-9.]+)"', pyproject.read_text(encoding="utf-8"))
+    assert match, "accelerate floor pin not found in pyproject.toml"
+    floor = tuple(int(part) for part in match.group(1).split("."))
+    assert floor >= (0, 27, 0), (
+        f"accelerate>={match.group(1)} predates save_fsdp_model's adapter_only "
+        "parameter (added in 0.27.0) and will regress #352"
+    )
+
+
+def test_installed_accelerate_save_fsdp_model_supports_adapter_only():
+    from accelerate.utils import save_fsdp_model
+
+    assert "adapter_only" in inspect.signature(save_fsdp_model).parameters
+
+
+def test_installed_accelerate_load_fsdp_model_supports_adapter_only():
+    from accelerate.utils import load_fsdp_model
+
+    assert "adapter_only" in inspect.signature(load_fsdp_model).parameters
+
+
+def test_fsdp_resume_checkpoint_state_dict_is_adapter_only_for_peft_model():
+    """Acceptance criterion from #352: checkpoint size is O(adapter), not O(base)."""
+    import torch.nn as nn
+    from accelerate.utils.fsdp_utils import _get_model_state_dict
+    from peft import LoraConfig, get_peft_model
+
+    class TinyBase(nn.Module):
+        def __init__(self):
+            super().__init__()
+            # A lopsided frozen-base-vs-adapter ratio, mirroring the issue's own
+            # measurement (37 GB frozen base next to a 131 MB adapter).
+            self.big = nn.Linear(2048, 2048, bias=False)
+
+        def forward(self, x):
+            return self.big(x)
+
+    model = get_peft_model(TinyBase(), LoraConfig(target_modules=["big"], r=4, lora_alpha=8))
+
+    full_numel = sum(v.numel() for v in _get_model_state_dict(model, adapter_only=False).values())
+    resume_sd = _get_model_state_dict(model, adapter_only=True)
+    resume_numel = sum(v.numel() for v in resume_sd.values())
+
+    assert all("lora_" in key for key in resume_sd), (
+        f"expected only adapter keys in the FSDP resume checkpoint, got {list(resume_sd)}"
+    )
+    # The frozen base (2048*2048 = 4_194_304 params) must not appear at all; only the
+    # rank-4 LoRA A/B matrices (2 * 4 * 2048 = 16_384 params) should.
+    assert resume_numel == 16_384
+    assert resume_numel < full_numel / 100


MakazhanAlpamys (changes requested):
Re-review of 9fac63134933478675b8e9998d5e1e2042475475: the dependency floor and adapter-only state-dict assertions are correct, and my focused run passed 4 tests; reverting accelerate>=0.27.0 to >=0.25.0 failed the floor regression. One blocking acceptance gap remains: issue #352 explicitly requires resume from an adapter-only FSDP checkpoint, but tests/test_issue352_fsdp_peft_checkpoint.py only inspect signatures and state-dict size; they never save and load or resume through Trainer. Please add a CPU-safe mocked or real minimal regression that exercises the resume path, or document a precise upstream-only reason and adjust the issue linkage. Keep this PR open until addressed.

MakazhanAlpamys:
Thank you, @AmirF194 — this is a well-traced dependency fix and the root-cause analysis in the PR body is accurate. I verified the upstream claim independently against the real release tags rather than taking the description on trust:

- `accelerate` v0.26.0: `def save_fsdp_model(fsdp_plugin, accelerator, model, output_dir, model_index=0)` — no `adapter_only`.
- `accelerate` v0.27.0: the parameter is present on `save_fsdp_model`, `load_fsdp_model`, `load_fsdp_optimizer` and `_get_model_state_dict`.

So "0.27.0 is the first release" holds. I also unpacked the `transformers==5.16.1` wheel: `transformers/distributed/fsdp.py::get_fsdp_ckpt_kwargs()` does exactly the feature detection you describe, and `trainer.py:3291` reaches `save_fsdp_model(..., **get_fsdp_ckpt_kwargs())` under `self.is_fsdp_enabled`, which is set from the presence of the accelerate FSDP plugin — so Soup's own `full_shard auto_wrap` presets land squarely on that branch. Load side is symmetric.

I ran the mutation myself, which is the part I care most about:

```
# pyproject.toml: accelerate>=0.27.0 -> accelerate>=0.25.0
FAILED tests/test_issue352_fsdp_peft_checkpoint.py::test_pyproject_accelerate_floor_supports_adapter_only_fsdp_checkpoints
E   assert (0, 25, 0) >= (0, 27, 0)

# boundary check: accelerate>=0.26.9
E   assert (0, 26, 9) >= (0, 27, 0)
```

The assertion pins 0.27.0 specifically rather than "any bump", which is the right shape. Unmodified branch: `4 passed`. Scoped sweep `-k "fsdp or accelerate"`: `79 passed`. `ruff check` clean. No test was deleted or weakened anywhere in the diff.

Two changes requested before merge, one substantive and one editorial:

**1. `src/soup_cli/commands/doctor.py:25` is now the only `DEPS` row out of sync with `pyproject.toml`.**

```python
("accelerate", "accelerate", "0.25.0", True),
```

Every other required row in that table matches its pyproject floor exactly (transformers 5.16.1, peft 0.20.0, trl 0.29.0, datasets 2.14.0, bitsandbytes 0.41.0). After this PR, `soup doctor` would green-light an accelerate 0.26.0 that the package metadata forbids — which is precisely the "validated and documented but read by nothing" shape this project has been bitten by repeatedly. Please bump it to `0.27.0` in this PR.

**2. Please soften the changelog fragment's framing.**

The framing correction matters for the record. The measurement behind #352 (`benchmarks/gate-h100-validation.md`, STEP 28) ran on **transformers 4.57.6 with accelerate 1.14.0** — accelerate already had `adapter_only` there, and `transformers.trainer.get_fsdp_ckpt_kwargs` does not exist in 4.57.6 at all (I checked the installed 4.57.6: `hasattr` is `False`). So the observed 37 GB write was caused by the transformers side, and is already closed by the shipped `transformers>=5.16.1` floor. What your PR closes is the remaining *resolvable* way to defeat that fix: under `accelerate>=0.25.0`, pip could legally pick 0.25/0.26, and 5.16.1's feature detection would then silently return `{}` and write the full base again.

That is a real hole worth closing, and I want it landed. But the fragment currently reads as though the 37 GB regression itself is being fixed here. Please reword it to say the floor bump *prevents a resolvable regression* of the adapter-only checkpoint path.

Optional, not blocking, and genuinely valuable if you want it: nothing in `src/` or `tests/` references `get_fsdp_ckpt_kwargs` (zero grep hits repo-wide). All four of your new tests would stay green if transformers were downgraded — i.e. green while #352 fully regresses. A test asserting `get_fsdp_ckpt_kwargs()` returns `{"adapter_only": True}` against the installed accelerate would guard the half that actually caused the measured failure. It needs an import guard or a home in the `transformers-floor` CI cell, since the common dev stack still resolves 4.57.6.

On #352's acceptance criteria, for the record: (a) is met substantively — `pytorch_model_fsdp.bin` is still written, it just now holds adapter tensors only; (b) is met as a unit test on accelerate's state-dict helper rather than on a real checkpoint directory; (c) resume-from-checkpoint is not exercised. Your "not run: an actual multi-GPU FSDP job" note is the honest call for a floor bump and I am not asking you to produce an 8-GPU run. I will keep #352 open after merge to track the untested resume path unless you would rather scope that here.

AmirF194:
Worth being precise about what this diff can and cannot cover: soup does not implement its own FSDP checkpoint save or resume path, it imports transformers.Trainer unmodified and the save_fsdp_model / load_fsdp_model call happens entirely inside accelerate. Our fix is the dependency floor, nothing in src/ changed.

Given that, a full Trainer.train() and resume round trip would mostly be exercising transformers and accelerate's own internals rather than this diff, and making it real (not mocked) needs an actual FSDP process group, which is not CPU-safe. The existing test_fsdp_resume_checkpoint_state_dict_is_adapter_only_for_peft_model already checks the acceptance criterion (adapter-only state dict, not full base) at the exact call this fix touches, using the installed accelerate version's own function.

If you would still like a Trainer-level integration test despite that, I can build one with a single-process gloo FSDP group, but it will need GPU CI or a longer-running job rather than the existing CPU test suite. Let me know which you would prefer.

MakazhanAlpamys:
@AmirF194 — your reasoning on the integration test is right and I accept it. **Do not build the gloo FSDP round trip.** Soup imports `transformers.Trainer` unmodified, the `save_fsdp_model` / `load_fsdp_model` call happens inside accelerate, and nothing in `src/` changed here — so a `Trainer.train()` round trip would mostly exercise upstream internals at real GPU-CI cost to cover a dependency-floor bump. I will keep #352 open on the untested resume path after merge, as I said earlier.

But that was the optional question. **My two actual requests are still open, and I think they got buried under it** — neither is in `9fac6313`:

**1. `src/soup_cli/commands/doctor.py:25` is still `0.25.0`.**

```
$ git show 9fac6313:src/soup_cli/commands/doctor.py | grep accelerate
25:    ("accelerate", "accelerate", "0.25.0", True),

$ git show 9fac6313:pyproject.toml | grep 'accelerate>='
111:    "accelerate>=0.27.0",
```

After this PR that is the only required row in the `DEPS` table out of sync with `pyproject.toml` — every other one matches exactly (transformers 5.16.1, peft 0.20.0, trl 0.29.0, datasets 2.14.0, bitsandbytes 0.41.0). So `soup doctor` would report a healthy install on an accelerate 0.26.0 that the package metadata forbids, i.e. it would green-light precisely the version this PR exists to exclude. One line.

**2. The changelog fragment still frames this as fixing the observed regression.** It currently reads:

> A LoRA/PEFT run under FSDP no longer writes a full consolidated copy of the frozen base model into every checkpoint.

The 37 GB write in `benchmarks/gate-h100-validation.md` STEP 28 was measured on **transformers 4.57.6 with accelerate 1.14.0** — accelerate already had `adapter_only` there, and `transformers.trainer.get_fsdp_ckpt_kwargs` does not exist in 4.57.6 at all (I checked the installed 4.57.6: `hasattr` is `False`). That regression is already closed by the shipped `transformers>=5.16.1` floor. What your PR closes is the remaining *resolvable* way to defeat it: under `accelerate>=0.25.0`, pip could legally pick 0.25 or 0.26, and 5.16.1's feature detection would then silently return `{}` and write the full base again.

That is a real hole and I want it landed — please just reword the fragment to say the bump **prevents a resolvable regression** of the adapter-only checkpoint path rather than implying it fixes the measured 37 GB write.

Push those two and I will merge. Everything else on this PR is already verified: the mutation kills at `0.25.0` and at the boundary `0.26.9`, the upstream claim checks out against the real accelerate release tags, no test was deleted or weakened, and `-k "fsdp or accelerate"` is 79 passed.

AmirF194:
Pushed both. doctor.py's DEPS table now reads 0.27.0, and the changelog fragment now says this closes a resolvable regression path rather than implying it fixes the measured 37 GB write. Also re-ran the full local suite (42 passed) after both changes.

MakazhanAlpamys:
Merged as `14a93ec`. Thank you, @AmirF194 — both requested changes landed and I re-verified before merging.

```
doctor.py:25   ("accelerate", "accelerate", "0.27.0", True)     <- was 0.25.0
fragment       "closes a remaining way a LoRA/PEFT run under FSDP could
                regress back to writing the full frozen base model"
```

Mutation re-run on the final head:

```
baseline                                4 passed
pyproject accelerate>=0.27.0 -> 0.25.0  1 failed
  test_pyproject_accelerate_floor_supports_adapter_only_fsdp_checkpoints
```

Boundary-tight as before. No test deleted or weakened, all 13 checks green on `6b4731c6`.

**Your argument against the integration test was right and I want it on the record**, because "we decided not to test this" is the kind of thing that later reads as an oversight. Soup imports `transformers.Trainer` unmodified, the save and load happen inside accelerate, and nothing in `src/` changed here — so a `Trainer.train()` round trip would mostly exercise upstream at real GPU-CI cost to cover a dependency-floor bump. You offered to build a real single-process gloo FSDP group if I wanted it, rather than reaching for a mock that would have closed the checkbox without establishing anything. That is the right instinct and it is why I said no.

I have **reopened #352** — it auto-closed on the `Fixes` keyword in the PR body, and I had told you I would keep it open for acceptance criterion 3. It now carries an explicit scope note: criteria 1 and 2 are met, resume-from-checkpoint is a known gap held open **by agreement rather than by omission**, and the record states that the measured 37 GB write was a transformers-side defect already closed by the `transformers>=5.16.1` floor, so nobody later reads this as "the accelerate bump fixed it".

You are in `CONTRIBUTORS.md` as of `24f0648`.

One gap I found while merging, not yours and not a request: `tests/test_issue502_qwen35_transformers5.py::test_doctor_training_bounds_match_declared_extra` pins `doctor.py`'s `DEPS` floors against the declared extra — but only for the three packages in `_RUNTIME_FLOORS` (transformers, trl, peft). `accelerate` is not among them, which is why reverting your `doctor.py` row to `0.25.0` still passes the suite. Adding it to `_RUNTIME_FLOORS` would also change `_REQUIRES_TRAINING_FLOOR`'s skip behaviour, so it wants a separate assertion. Recorded on #352.

Separately: you filed #606 and opened #607 fifty seconds later, and @ARAVIND281 independently opened #608 for the same issue twenty-four minutes after that. I have posted the situation on both PRs and I am running the same battery against both rather than deciding by timestamp. Nothing for you to do; I did not want you to find out from a merge notification.

AmirF194:
Thanks for the thorough re-review and for merging. Glad the acceptance gap and the changelog wording both landed clean.

PR #592, opened 2026-08-29
fix(export): require explicit AWQ calibration data

Faisal01011:
## Summary

- require an explicit local calibration JSONL for direct AWQ exports
- reject missing, empty, or malformed calibration input before importing AutoAWQ or loading a model
- always forward the validated rows to `model.quantize(..., calib_data=...)`
- update CLI help and export documentation so the behavior is explicit

## Why

Without `--calibration-data`, AutoAWQ silently downloads its large default calibration dataset. This makes a direct export unexpectedly network-dependent and can fetch roughly 214k rows.

## Testing

- `pytest tests/test_awq_gptq_export.py tests/test_export.py tests/test_issue144_gguf_export.py tests/test_onnx_tensorrt_export.py tests/test_cli_startup_is_light.py -q --no-cov` (118 passed, 1 skipped)
- `git diff --check`

Refs #338

The AutoAWQ/Transformers dependency conflict and Python-wheel support remain outside this focused slice.


MakazhanAlpamys (changes requested):
Re-review of 3000b70ec6c30605d545f6bc8ab007bec2f3a00c: the scoped AWQ change matches the remaining issue #338 item, the focused suite passed 119 tests, Ruff and git diff checks are clean, and mutations of both the calibration guard and calib_data forwarding were caught. One blocking input-validation gap remains: _load_calibration_texts assumes every valid JSON line is an object and calls row.values() or row["text"]. Inputs such as [], null, a JSON string, or a number therefore raise uncaught AttributeError or TypeError at export.py:925 instead of the documented friendly rejection before AutoAWQ import. Validate the row shape (and handle file/encoding errors as appropriate) and add regression tests for these inputs. Keep the draft open until this is fixed and the current checks pass.

Faisal01011:
Addressed the blocking input-validation review in f0c3404.

- `_load_calibration_texts` now ignores valid JSON rows that are not objects (`[]`, `null`, strings, and numbers), and skips object rows that yield empty text.
- UTF-8 decode failures and file-read errors are converted into specific friendly CLI refusals.
- AWQ and GPTQ both catch those read errors before importing their quantizer or loading a model.
- The new AWQ regressions are table-driven across all four requested JSON shapes and assert that an attempted `awq` import fails the test; invalid UTF-8 and unreadable-path controls enforce the same ordering.

Verification on the updated head:
- focused AWQ/GPTQ file: 58 passed
- adjacent export/CLI/changelog set: 141 passed, 1 platform skip
- repository-wide Ruff: clean
- changelog fragments: 51 validated
- `git diff --check`: clean

Keeping the PR in draft while the refreshed CI checks run.

MakazhanAlpamys:
Thanks @Faisal01011 — the core of this is right, and `f0c3404` closes the review finding properly. Two cheap test fixes and I will merge it.

## What I verified

The blocking finding from the last round is genuinely fixed, and I checked it against the real pre-PR function rather than reading the diff. Extracting `_load_calibration_texts` from before the change and calling it directly:

```
PRE-PR  []     -> AttributeError: 'list' object has no attribute 'values'
PRE-PR  null   -> TypeError: argument of type 'NoneType' is not iterable
PRE-PR  "str"  -> AttributeError: 'str' object has no attribute 'values'
PRE-PR  42     -> TypeError: argument of type 'int' is not iterable
POST-PR all four -> []   then refused with "No usable calibration samples"
```

Your `isinstance(row, dict)` guard at `export.py:879` is what makes that the documented friendly rejection instead of a traceback. You also fixed a latent bug you did not claim: pre-PR `{"text": null}` produced the literal string `'None'` as calibration text, and that ran through the **shared** reader, so GPTQ was silently affected too.

**The 22 deleted test lines are all upgrades, and I checked every one.** They belong to four pre-existing tests that passed `calibration_data=None`, which is no longer legal. The important one got strictly stronger:

```diff
-  mock_model.quantize.assert_called_once()
+  mock_model.quantize.assert_called_once_with(
+      mock_tokenizer, quant_config={...}, calib_data=["sample"])
```

No `skip` or `xfail` was added anywhere.

**Six mutations, six killed.** The one I care most about is dropping `calib_data=` from `model.quantize(...)` while keeping the guard — that is the silent-download regression itself, and it is invisible to any test that only checks the error path. Your exact-forwarding assertion catches it (`2 failed, 56 passed`). The ordering claim also holds by observation rather than by reading: a mutation that changes **only** the order, moving validation after the `awq` import, fails 8 tests on the `"AWQ was imported before calibration validation"` tripwire. Since the import must succeed before `from_pretrained`, blocking at import proves the model load is never reached. That is a real pre-flight, not a post-mortem.

`107 passed` on the required files, `326 passed` across the nine-file export sweep, `ruff` clean, path containment (`_validate_calibration_path`, `is_under_cwd`) untouched.

## Blocking: this PR makes a width-robust test file width-fragile

This repo has been turned red by Rich wrapping **four separate times**, most recently across all nine cells at once. I reproduced the regression here:

```
main    COLUMNS=100 -> 49 passed        main    COLUMNS=60 -> 49 passed
PR head COLUMNS=100 -> 1 failed         PR head COLUMNS=60 -> 2 failed
                                        PR head COLUMNS=70 -> 1 failed
```

Both failures are tests this PR adds; neither exists on `main`. They assert on **prose phrases** in raw CLI output, so Rich wraps mid-phrase:

**`test_calibration_help_names_both_required_formats`** — `assert "Required for AWQ and GPTQ" in result.output`. At COLUMNS=100:
```
│ --calibration-data  TEXT  Path to calibration JSONL. Required for AWQ and │
│                           GPTQ.                                          │
```

**`test_export_awq_no_calibration_data_raises`** — `assert "default calibration dataset" in output`:
```
AssertionError: assert 'default calibration dataset' in
'...downloads its large default calibration \ndataset: pass a calibration JSONL...'
```

It is green at CI's default width today, so this would merge and then break on an unrelated help-string edit, an option rename, or a Typer bump — the worst kind of time bomb, because the commit that goes red is not the commit that caused it. Note the repo's ANSI ratchet (`tests/test_cli_help_assertions_are_ansi_safe.py`) does **not** catch these: its scope is deliberately quoted `--flags`, and these are prose.

Also worth knowing: an ANSI-strip plus whitespace-collapse **alone does not fix it**, because Rich's `│` borders sit between the fragments. Two things that do work:

- strip ANSI **and box-drawing characters**, then collapse whitespace (verified robust down to COLUMNS=50);
- better, assert on the Click parameter directly and skip rendering entirely:

```python
cmd = typer.main.get_command(app).commands["export"]
p = next(p for p in cmd.params if "--calibration-data" in p.opts)
assert p.help == "Path to calibration JSONL. Required for AWQ and GPTQ."
```

That second form is exact, render-independent, and cannot drift with terminal width.

## Should-fix, one line

`export.py:1080` in the GPTQ guard still says:

> `Unlike --format awq, auto-gptq has no built-in fallback dataset:`

As of this PR, AWQ refuses without calibration data too, so the contrast is now false. The line predates you (it came in with #475), but this PR is what makes it wrong.

## Optional

- Cite `#338` in the changelog body alongside `#592` — siblings such as `429.fixed.md` cite both.
- `_load_calibration_texts` uses a plain `open()` with no `O_NOFOLLOW` and no size cap, while its siblings (`draft.py:106`, `reward.py:56`, `probe.py:442`) are all cwd-contained, `O_NOFOLLOW` and size- and row-capped — and the docs paragraph immediately below your own new text advertises `O_NOFOLLOW` for GGUF calibration data. The gap predates you, but this PR routes **every** AWQ export through that reader where it was usually `None` before, so the exposure grows. Your call whether to fold it in here or leave it for a follow-up.

Your scope carve-out is honest and I want to say so explicitly: `Refs #338` rather than `Closes` is correct, this shuts both halves of the third item, and acceptance criterion 1 — documenting the AutoAWQ/transformers conflict — correctly stays open. That is the right shape for a slice.

Fix the two width-fragile assertions, flip it out of draft, and I will re-run the mutation set and merge.

MakazhanAlpamys:
@Faisal01011 — checking in, no pressure attached.

This has been sitting five days and it is **two test assertions away from merging**. The substance is done: the blocking finding is genuinely closed (I verified it against the real pre-PR function, not the diff — all four malformed inputs that used to raise now return cleanly), your scope carve-out is honest, and `Refs #338` rather than `Closes` is the right call.

What is left, in full, so you do not have to re-read the review:

1. **Two width-fragile assertions.** An ANSI strip alone does not fix them, because Rich's `│` borders land between the fragments. Either strip ANSI *and* box-drawing characters then collapse whitespace, or — better — skip rendering entirely:

   ```python
   cmd = typer.main.get_command(app).commands["export"]
   p = next(p for p in cmd.params if "--calibration-data" in p.opts)
   assert p.help == "Path to calibration JSONL. Required for AWQ and GPTQ."
   ```

   Exact, render-independent, cannot drift with terminal width.

2. **One stale line**, `export.py:1080`: *"Unlike `--format awq`, auto-gptq has no built-in fallback dataset"*. AWQ now refuses without calibration data too, so the contrast is false. It predates you; this PR is what makes it wrong.

3. **Flip it out of draft** — it is still marked draft, so I cannot merge it even once the above lands.

Then I re-run the mutation set and merge.

If you have moved on to other things, that is completely fine — say so and I will either leave it open for you with no clock, or finish those three items myself and credit you for the work, whichever you prefer. I would rather ask than let a nearly-done PR quietly rot.

One heads-up unrelated to your changes: `pytorch-smoke` became a required check after your last run, so this will show as blocked at 13/13 green. A rebase picks it up; nothing wrong with your branch.

MakazhanAlpamys:
@Faisal01011 — third and last check-in on this one, and this time with a date on it, because leaving it open-ended has not worked twice.

Nothing has changed about my assessment: the substance is done and correct, and the three remaining items are a test rewrite, a stale comment line, and clicking "Ready for review". They are listed in full in my 2026-09-03 comment above.

**If I have not heard from you by 2026-09-10 18:00 UTC, I will finish those three myself and merge, with the work credited to you** — your commits, your `CONTRIBUTORS.md` entry, your name in the changelog fragment. That is the outcome I said I would take if you had moved on, and I would rather do it than let a merge-ready fix rot for a fourth week.

If you want it back, say so at any point — including after I have pushed — and I will hand it over.

Two notes since we last spoke:
- `pytorch-smoke` is now a required check, so a rebase onto current `main` is needed regardless.
- v0.74.0 was tagged while this was open, so the changelog fragment has to live in `changelog.d/0.74.0/` (named after the **PR** number, `592.fixed.md`, with `by @Faisal01011` in the text). Several open PRs are in the same position; it is release timing, not anything you did.

MakazhanAlpamys:
@Faisal01011 — the date I set is **tomorrow, 2026-09-10 18:00 UTC**, so this is the reminder rather than a new ask.

Nothing has changed: the substance is done and correct, and the three remaining items are a test rewrite, a stale comment line, and clicking "Ready for review". They are listed in full in my 2026-09-03 comment.

If I have not heard from you by then I will finish those three myself and merge, **with the work credited to you** — your commits, your `CONTRIBUTORS.md` entry, your name in the changelog fragment. If you would rather take it back at any point, including after I have pushed, say so and it is yours.

Two things that have moved since we last spoke, neither of them your doing:

- `pytorch-smoke` is a required check now, so a rebase onto current `main` is needed regardless.
- v0.74.0 was tagged while this was open, so the fragment belongs in `changelog.d/0.74.0/` and is named after the **PR** (`592.fixed.md`), not the issue.

MakazhanAlpamys:
@Faisal01011 — the date I set, 2026-09-10 18:00 UTC, passed yesterday, so as promised I am finishing the three remaining items myself, on top of your commits. The substance was already done and correct. You stay the author of every commit, the `CONTRIBUTORS.md` entry and the changelog fragment are credited to you, and the merge commit will say exactly which lines were mine.

If you would rather take it back at any point, including after I push, say so and it is yours again.

MakazhanAlpamys:
@Faisal01011 — pushed as promised, and marked ready for review. Your three commits (`905deec6`, `3000b70e`, `f0c3404c`) are untouched, and these are mine, on top of them (fast-forward, no force):

| commit | what |
|---|---|
| `88c553e3` | a merge of current `main` (clean, no conflicts). It was needed so the changelog fragment can live in `changelog.d/0.74.0/`, and so `pytorch-smoke` runs against the current floor. |
| `8357fd29` | the test rewrite: the help check asserts on the Click parameter itself rather than on rendered `--help`, and the refusal-message checks strip ANSI and box-drawing characters before matching. It is width-safe — the original failed at `COLUMNS=60` — and colour-safe. |
| `c631629b` | the stale line: the GPTQ refusal no longer says "Unlike --format awq..." |
| `62835827` | the fragment moved to `changelog.d/0.74.0/592.fixed.md`, keeping your wording, credited `(#338 by @Faisal01011 in #592)` |

One line outside the three items, so you know: with forced colour, Rich colours the `8` in `not valid UTF-8`, so that assertion now goes through the same strip helper.

Verification: 151 passed across the export, fragment, startup-lightness and ANSI suites; the AWQ file passes at every width from 40 to 100 columns, with and without forced colour. Mutations killed: disabling the AWQ calibration guard, reverting the help text, and removing "default calibration dataset" from the refusal.

I will merge on green CI unless you say otherwise.

MakazhanAlpamys:
Merged as `f760e93e`. Thank you, @Faisal01011.

**The substance is yours**, in `905deec6`, `3000b70e` and `f0c3404c`, all untouched. A direct `soup export --format awq` let AutoAWQ silently download its own default calibration set — roughly 214k rows — so an export a user believed was local was network-dependent and calibrated on whatever upstream happened to ship. The detail I keep coming back to is that you refuse **before** importing AutoAWQ and before loading the model, so a malformed calibration file costs seconds instead of a model load. That ordering is easy to get wrong and nothing in the issue asked for it.

As announced across three check-ins and when the 2026-09-10 18:00 UTC date passed, I finished the three remaining items myself: the width-fragile test rewrite, the stale "Unlike `--format awq`" line in the GPTQ refusal that your own change made false, and the fragment move into `changelog.d/0.74.0/` after v0.74.0 was tagged mid-review. The merge commit names which commits were whose. If you would rather have taken them, say so — it costs nothing to say.

**What I ran.** 151 passed across the export, fragment, startup-lightness and ANSI suites; the AWQ file passes at every width from 40 to 100 columns, with and without forced colour. Mutations killed: disabling the AWQ calibration guard, reverting the help text, and removing "default calibration dataset" from the refusal message. 13/13 required checks green on `62835827`.

Your `CONTRIBUTORS.md` entry is in `b5723ce8`, and the changelog fragment credits you.

Note #338 stays **open**: the AutoAWQ/Transformers dependency conflict and Python-wheel support are outside this slice, which is why the merge says `Refs` rather than `Closes`.

PR #594, opened 2026-08-29
fix(tests): set chat_template in the SFT smoke fixture

chadetov:
## What does this PR do?

Fixes `tests/test_smoke_train.py::test_sft_smoke`, which has been failing since
v0.36.0. The fixture sets `data.format: chatml` but not `data.chat_template`, and
`sshleifer/tiny-gpt2` carries no chat template of its own, so `sft_format.py:183`
raises:

```
ValueError: Tokenizer has no chat_template. Pass data.chat_template: chatml
(or llama3/qwen2.5/mistral/gemma3/phi4/deepseek-r1) in soup.yaml, or supply a raw
Jinja string. The previous silent f-string fallback produced wrong loss labels and
was removed in v0.36.0.
```

The fix is the one line the error message asks for. The trainer is fine — the run
gets through model load and LoRA attach (`64 trainable / 102,778 total`) and dies
at data formatting.

### Why it went unnoticed

Nothing runs this test. The default `addopts` exclude `-m smoke`, and the only
smoke job in CI is `mlx-smoke`, which selects `-k mlx_sft_smoke` on an install
that asserts the PyTorch stack is *absent* (`ci.yml:62-68`). So `test_sft_smoke`
and `test_dpo_smoke` run in neither the default suite nor CI, which is how a
fixture could drift out of sync with a hard error added in v0.36.0.

I have not touched that, since making CI run the PyTorch smoke tests means model
downloads and runner minutes — your call, not something to fold into this PR.

### Scope

Only `sft_config_yaml` is changed. `mlx_sft_config_yaml` has the same
`format: chatml` line, but it goes through the MLX backend rather than
`sft_format.py`, and it is the one smoke test CI does run and pass, so I left it
alone rather than change a green path I cannot execute (MLX is Apple-Silicon
only; I am on an Intel Mac).

The DPO fixture is unaffected — it uses `format: dpo` and passed throughout.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Verification

Run on `linux/amd64`, Python 3.12, torch 2.13.0, transformers 5.16.1, via
`pip install ".[train]"`:

```
before:  1 failed, 1 passed, 1 skipped     (test_sft_smoke failed)
after:   2 passed, 1 skipped               (test_mlx_sft_smoke skips, no MLX)
ruff check tests/test_smoke_train.py  ->  All checks passed!
```

## Checklist

- [ ] `pytest tests/ -v` passes — **not verified on a supported platform.** My
  local machine is an Intel Mac, where `[train]` cannot resolve: torch ships no
  macOS x86_64 wheel past 2.2.2, but `transformers>=5.16.1` needs torch >= 2.5,
  so transformers disables its torch backend and `peft` fails to import. The full
  suite there is 18552 passed / 402 failed, all from that. I verified the changed
  file in a Linux container instead, as above. Happy to run the full suite there
  too if useful.
- [x] Updated relevant docs — n/a, test-only
- [x] New tests added for new functionality — n/a; this restores an existing test
- [x] No breaking changes
- [x] Changelog fragment — n/a, no user-visible impact

## One question

The run also prints:

```
train_on_responses_only requested but tokenizer has no chat_template —
falling back to text path. Pass data.chat_template explicitly to enable masking.
```

So that path degrades gracefully on a missing template while `sft_format.py` hard
-errors on it. I assume that is deliberate given the wrong-loss-labels history,
but flagging in case the asymmetry is not intended.

### DIFF
--- a/tests/test_smoke_train.py
+++ b/tests/test_smoke_train.py
@@ -69,7 +69,13 @@ def tiny_dpo_data(tmp_path: Path) -> Path:
 
 @pytest.fixture
 def sft_config_yaml(tmp_path: Path, tiny_train_data: Path) -> Path:
-    """Create a minimal SFT config for smoke testing with tiny-gpt2."""
+    """Create a minimal SFT config for smoke testing with tiny-gpt2.
+
+    ``chat_template`` is set even though ``format`` already says chatml: the
+    two are separate fields, and tiny-gpt2 ships no template of its own. Since
+    v0.36.0 the missing-template case is a hard error rather than the silent
+    ``f"{role}: {content}"`` fallback that produced wrong loss labels.
+    """
     config_path = tmp_path / "soup.yaml"
     output_dir = tmp_path / "output"
     config_path.write_text(
@@ -79,6 +85,7 @@ task: sft
 data:
   train: {tiny_train_data}
   format: chatml
+  chat_template: chatml
   val_split: 0.0
   max_length: 128
 


MakazhanAlpamys:
Merged as `d9f6276`. Welcome, @chadetov — a good first contribution, and the diagnosis is worth more than the one-line fix.

**On the verification, with one limitation stated up front.** I could not run `test_sft_smoke` end to end on my box: this machine has torch 2.5.1, so the run dies earlier than your failure, at model load, on transformers' `check_torch_load_is_safe` (CVE-2025-32434) because `sshleifer/tiny-gpt2` ships no safetensors. So I verified the mechanism rather than the symptom, and I want to be explicit that I did not reproduce your exact traceback myself.

What I did check:

```
without chat_template -> ValueError: Tokenizer has no chat_template. Pass
                         data.chat_template: chatml (or llama3/qwen2.5/...)
with    chat_template -> formatting succeeds, row keys ['text']
```

driven straight through `_legacy_text_format_row` with a tokenizer whose `chat_template` is `None`, which is exactly the state `tiny-gpt2` is in. The raise site is `sft_format.py:183`, matching your quote verbatim. And the reason no config test caught this:

```
WITHOUT chat_template  -> parses OK; data.chat_template = None
WITH    chat_template  -> parses OK; data.chat_template = 'chatml'
```

Both are valid configs. `format` and `chat_template` really are independent fields, so the omission was never a schema error — it only surfaced at runtime, in a test nothing runs.

**Your CI analysis is correct and I confirmed each half.** `pyproject.toml:256` carries `addopts = "-m 'not smoke' ..."`, and `ci.yml:68` is `pytest -o addopts= -m smoke -k mlx_sft_smoke -q tests/test_smoke_train.py` on an install that asserts the PyTorch stack is absent. So `test_sft_smoke` and `test_dpo_smoke` execute in neither the default suite nor CI. That is the actual finding here; the missing line is just its symptom.

**The scope calls are the part I want to single out**, because both were right and both were verified rather than asserted:

- `mlx_sft_config_yaml` does carry the same `format: chatml` without a `chat_template` — I checked. Leaving it alone is correct: it uses a different base (`hf-internal-testing/tiny-random-LlamaForCausalLM`), reaches the MLX trainer instead of `sft_format.py`, and is the one smoke test CI genuinely runs and passes. Changing a green path you cannot execute would have been the easy mistake.
- The DPO fixture uses `format: dpo` and is untouched. Confirmed.
- And you declined to wire the PyTorch smoke tests into CI here, naming the cost (model downloads, runner minutes) and leaving the decision to me instead of taking it. That is the right instinct and it is why this merged unchanged.

Diff hygiene: the only removed line in the whole PR is the one-line docstring you replaced with the expanded one — no test deleted or weakened. `ruff check src/soup_cli/ scripts/ tests/` clean. All 13 required checks green on `2d76abe9`.

I have taken up your deferred item as #596 so it does not get lost. You are already in `CONTRIBUTORS.md`, and your #595 is filed and labelled — that one is a better find than this PR, and I have replied there with what I would want in the fix if you want to take it.

PR #597, opened 2026-08-29
bench: publish bit-exactness harness

Samearth17:
## Summary

Second reproducibility-only slice for #379.

This publishes the standalone `bitexact.py` measurement harness under `benchmarks/harness/` and indexes it in `benchmarks/README.md`.

The harness reproduces the published layer-streaming correctness protocol:

* shard the checkpoint with Soup's production sharder
* build the streamed model
* build a resident reference using matching numerics
* compare streamed vs resident logits bit-for-bit
* compare LoRA gradients bit-for-bit
* compare the training loss curve over 100 steps
* verify the streaming path is actually exercised

For NF4 runs, the resident reference is also NF4 rather than bf16.

No Soup implementation code or benchmark methodology was changed.

## Validation

* `python -m py_compile benchmarks/harness/bitexact.py` — passed
* `python -m ruff check benchmarks/harness/bitexact.py` — passed
* `git diff --check` — passed
* `python -m pytest -q` — 18,958 passed, 283 skipped, 5 deselected; 81.42% coverage
* Local end-to-end CUDA execution was not possible because the development machine has no CUDA GPU.
* The harness exits cleanly on the local machine with:
  `SKIP: CUDA is required for bitexact.py`

## Historical reference

The published measurement record describes `bitexact.py` as:

`shard -> stream -> compare logits/gradients/loss curve against a resident reference of matching numerics`

The historical CUDA run used Qwen2.5-0.5B-Instruct with NF4, bf16 compute, sequence length 64, and two stream buffers.

Refs #379

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -38,7 +38,7 @@ only in a scratchpad on a machine that is gone, which is
 |---|---|---|---|
 | [`issue331_qlora_scope.py`](harness/issue331_qlora_scope.py) | #331 | Does the wrong-gradient defect reach **ordinary QLoRA**? Three arms in one process, with a positive control. Answer: no — 0.0 against a control that diverges by 3.77e-01 | ~15 s, 4 GB card, no downloads |
 | [`bnb_repro.py`](harness/bnb_repro.py) | Upstream bitsandbytes #2034 / NF4 defect | Minimal standalone reproducer for the NF4 stale-gradient mechanism: recycled buffers expose `MatMul4Bit`'s `ctx` lifetime issue; private buffers are the reference and bf16 is the control | CUDA GPU required; no downloads; historical environment: torch 2.13.0+cu130, bitsandbytes 0.50.0; ~1 min |
-| [`bitexact.py`](harness/bitexact.py) | v0.72.0 / STEP 1; gate-h100-validation.md Reproducing | Shard -> stream -> compare logits, LoRA gradients, and loss curve against a resident reference using matching numerics; NF4 uses a resident NF4 reference | CUDA GPU required; model/checkpoint weights required; historical run: Qwen2.5-0.5B-Instruct NF4, seq 64 |
+| [`bitexact.py`](harness/bitexact.py) | v0.72.0 / STEP 1; gate-h100-validation.md Reproducing | Shard -> stream -> compare logits, LoRA gradients, and loss curve against a resident reference using matching numerics; NF4 uses a resident NF4 reference | CUDA GPU required; model/checkpoint weights required; historical run: Qwen2.5-0.5B-Instruct NF4, seq 64; CUDA validation: torch 2.5.1+cu121, transformers 4.57.6, trl 0.19.1, peft 0.18.1 |
 
 ## Hardware
 
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -38,6 +38,7 @@ only in a scratchpad on a machine that is gone, which is
 |---|---|---|---|
 | [`issue331_qlora_scope.py`](harness/issue331_qlora_scope.py) | #331 | Does the wrong-gradient defect reach **ordinary QLoRA**? Three arms in one process, with a positive control. Answer: no — 0.0 against a control that diverges by 3.77e-01 | ~15 s, 4 GB card, no downloads |
 | [`bnb_repro.py`](harness/bnb_repro.py) | Upstream bitsandbytes #2034 / NF4 defect | Minimal standalone reproducer for the NF4 stale-gradient mechanism: recycled buffers expose `MatMul4Bit`'s `ctx` lifetime issue; private buffers are the reference and bf16 is the control | CUDA GPU required; no downloads; historical environment: torch 2.13.0+cu130, bitsandbytes 0.50.0; ~1 min |
+| [`bitexact.py`](harness/bitexact.py) | v0.72.0 / STEP 1; gate-h100-validation.md Reproducing | Shard -> stream -> compare logits, LoRA gradients, and loss curve against a resident reference using matching numerics; NF4 uses a resident NF4 reference | CUDA GPU required; model/checkpoint weights required; historical run: Qwen2.5-0.5B-Instruct NF4, seq 64 |
 
 ## Hardware
 
--- a/benchmarks/harness/bitexact.py
+++ b/benchmarks/harness/bitexact.py
@@ -1,0 +1,654 @@
+#!/usr/bin/env python3
+"""Reproduce the layer-streaming bit-exactness measurement.
+
+This is the published ``bitexact.py`` harness described in
+``benchmarks/gate-h100-validation.md``:
+
+    shard -> stream -> compare logits/gradients/loss curve
+    against a resident reference of matching numerics
+
+Requirements
+------------
+- CUDA-capable GPU.
+- PyTorch, transformers, peft, safetensors, and bitsandbytes for NF4.
+- A locally available checkpoint, either as a path or through Soup's
+  weight cache/resolver.
+- Enough GPU memory for the resident reference as well as the streamed model.
+
+Typical invocation
+------------------
+    python benchmarks/harness/bitexact.py \
+        --weights Qwen/Qwen2.5-0.5B-Instruct \
+        --shards /tmp/qwen05b_nf4 \
+        --quant nf4 \
+        --seq 64
+
+The historical measurement used CUDA bf16 and two stream buffers. The resident
+reference always uses the same numerics as the streamed arm: NF4 is compared
+with resident NF4, never resident bf16.
+
+No benchmark methodology is introduced here. The harness makes the original
+correctness checks executable from the repository and exits non-zero whenever
+one of the required parity checks fails.
+
+A machine without CUDA is an intentional skip and exits 0.
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from pathlib import Path
+
+DTYPE = "bfloat16"
+DEFAULT_BUFFERS = 2
+DEFAULT_STEPS = 100
+DEFAULT_BATCH = 1
+DEFAULT_SEED = 3
+INPUT_SEED = 17
+
+
+def cuda_available() -> bool:
+    try:
+        import torch
+
+        return torch.cuda.is_available()
+    except Exception:
+        return False
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description="Reproduce Soup's layer-streaming bit-exactness gate."
+    )
+    parser.add_argument(
+        "--weights",
+        required=True,
+        help=(
+            "checkpoint path or model id resolvable by Soup's "
+            "weight resolver"
+        ),
+    )
+    parser.add_argument(
+        "--shards",
+        required=True,
+        help="directory in which Soup should create or reuse layer shards",
+    )
+    parser.add_argument(
+        "--quant",
+        choices=("none", "nf4"),
+        default="none",
+        help="streamed/base-weight representation (default: none)",
+    )
+    parser.add_argument(
+        "--seq",
+        type=int,
+        default=512,
+        help="sequence length (default: 512)",
+    )
+    parser.add_argument(
+        "--steps",
+        type=int,
+        default=DEFAULT_STEPS,
+        help=(
+            "training steps for the loss-curve comparison "
+            f"(default: {DEFAULT_STEPS})"
+        ),
+    )
+    parser.add_argument(
+        "--batch",
+        type=int,
+        default=DEFAULT_BATCH,
+        help=f"batch size (default: {DEFAULT_BATCH})",
+    )
+    parser.add_argument(
+        "--buffers",
+        type=int,
+        default=DEFAULT_BUFFERS,
+        help=f"stream buffer count (default: {DEFAULT_BUFFERS})",
+    )
+    parser.add_argument(
+        "--no-pin",
+        action="store_true",
+        help="use pageable host storage instead of pinned host storage",
+    )
+    return parser.parse_args()
+
+
+def model_arch_name(model) -> str:
+    config = getattr(model, "config", None)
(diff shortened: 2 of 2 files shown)


MakazhanAlpamys:
@Samearth17 — the harness is good and it is **one line** from finished, but as submitted it cannot complete a single run on any streamed model. Keeping it draft is the right call.

## The blocker, reproduced on real CUDA

```
$ py -3.10 benchmarks/harness/bitexact.py --weights HuggingFaceTB/SmolLM2-135M-Instruct \
    --shards ./shards_smol --quant none --seq 64 --steps 3
arch          llama
sharding      preparing layer shards...
layers        30
stream        building streamed model...
reference     loading resident model...
ERROR: RuntimeError: could not uniquely match LoRA parameter
  'base_model.model.model.layers.0.self_attn.q_proj.lora_A.default.weight'
```

It gets through shard, stream and resident load, then dies in its **own** `copy_lora` at line 162 — before the forward, gradient and loss-curve comparisons the harness exists to perform.

The cause, measured:

```
STREAMED RAW    ...layers.0.inner.self_attn.q_proj.lora_A.default.weight
STREAMED CANON  ...layers.0.self_attn.q_proj.lora_A.default.weight
RESIDENT RAW    ...layers.0.self_attn.q_proj.lora_A.default.weight
raw name sets equal:      False
canon vs resident equal:  True
```

`copy_lora` uses raw `named_parameters()` on **both** sides (lines 143-144). `StreamedDecoderLayer` inserts `.inner.` *mid-string*, after `layers.N.`, so the `endswith(name)` fallback can never match — 0 matches, `RuntimeError`.

**The harness already imports the right helper 50 lines further down**: `collect_lora_grads` correctly uses `canonical_named_parameters`. `copy_lora` and `make_non_vacuous_lora` were missed. And this is exactly the defect class #379's own acceptance criterion exists to prevent — *"compares through `canonical_named_parameters()`, never raw `named_parameters()`"* — met for the gradient comparison, violated in the one place that runs first.

## Patched, it works, and I have the transcripts for you

Changing only `copy_lora` to use `canonical_named_parameters` on both sides:

```
--quant none, seq 64, 5 steps
adapter_tensors_copied 120 | max_abs_logit_diff 0.0 bit_exact true
layer0_lora_grad 7.971496e+01 | gradients exact max_abs_diff=0.0
curves_equal true curve_max_rel=0.0e+00 | meta_params 270 | store 0.21 GB | tier ram
RESULT: bit-exact gate passed   EXIT=0

--quant nf4
adapter_tensors_copied 120 | logits 0.0 bit_exact true | gradients exact 0.0
curves_equal true | store 0.05 GB
RESULT: bit-exact gate passed   EXIT=0
```

Two things worth recording from that: it reproduced on **torch 2.5.1+cu121 / transformers 4.57.6 / trl 0.19.1 / peft 0.18.1**, well below the H100 record's stack — same shape as your #589 finding, the mechanism is not specific to the newer stack. And the NF4 store drops 0.21 -> 0.05 GB, the ~4x shrink the records claim. Cite these; you have no GPU and should not have to guess.

## It does discriminate — I checked what the PR does not show

With the adapter sync skipped, so streamed and resident genuinely differ:

```
adapter_tensors_copied 0
ERROR: streamed/resident logits are not bit-exact (max_abs_diff=3.687500e+00)
EXIT=1
```

So "the default invocation fails when the thing it demonstrates is absent" holds. But I had to edit source to get that — unlike `bnb_repro.py`, which grew `--bypass-pool` for exactly this. Please add the equivalent flag.

## Asks

1. **`copy_lora` (and `make_non_vacuous_lora`) must use `canonical_named_parameters` on both sides.** This is the merge blocker.
2. A built-in negative-control flag, for parity with `bnb_repro.py --bypass-pool`.
3. #379 also asks for a test where an **empty intersection fails rather than reporting a pass**. There is no test suite for `benchmarks/harness/` at all — the merged `bnb_repro.py` has none either, so this is precedent, not your omission. Your `if not streamed_grads: return False` handles the substance. Tell me if you want a waiver and I will record it on the issue rather than leave the criterion silently unmet.
4. Software versions in the `benchmarks/README.md` row — `bnb_repro.py`'s row states "torch 2.13.0+cu130, bitsandbytes 0.50.0"; yours states hardware only.
5. Re-run and paste real output. Use the transcripts above if that is faster.

On what is re-derivable: the harness matches the `gate-h100-validation.md` "Smoke — Qwen2.5-0.5B-Instruct, NF4" invocation flag-for-flag, so `max_abs_logit_diff 0.0`, `adapter_tensors_copied 96`, `curves_equal true`, `meta_params 288`, `store ~0.18 GB` all come back. Two honest caveats: `layer0_lora_grad` depends on `DEFAULT_SEED=3` / `INPUT_SEED=17` / the 0.02 LoRA-B scale, none of which the record states, so that figure is probably **not** reproducible; and the historical script printed `layers_with_grad` / `n_layers_seen` and had a `--json` mode the 8B record used, which this version lacks. Worth saying in the PR body rather than leaving a reader to discover it.

Everything else is clean: indexed in `benchmarks/README.md`, ruff clean, importable with no side effects (`torch in sys.modules` is `False` after import), no test deleted, and no changelog fragment — correctly, since #589 shipped none either.

MakazhanAlpamys:
@Samearth17 — saw `79502cd6` (rebase + fix). You took the `benchmarks/README.md` harness row, and `collect_lora_grads` now goes through `canonical_named_parameters`, which is the right call. One thing before you flip it out of draft, because I think it will still fail on you and I would rather you not spend a GPU hour finding out.

## `copy_lora` did not get the same treatment

```python
# bitexact.py:139
source_params = dict(source.named_parameters())     # <- raw
target_params = dict(target.named_parameters())     # <- raw
...
matches = [p for other_name, p in source_params.items()
           if "lora_" in other_name and other_name.endswith(name)]
if len(matches) != 1:
    raise RuntimeError(f"could not uniquely match LoRA parameter {name!r}")
```

`collect_lora_grads` uses the canonical helper twenty lines below; `copy_lora` keeps the raw call and adds a suffix fallback around it. I do not think that fallback can work, and the reason is mechanical:

```python
# layer_stream_runtime.py:1173
def canonical_named_parameters(model):
    for name, param in model.named_parameters():
        yield name.replace(".inner.", "."), param
```

`.inner.` is a **middle** segment, not a prefix or a suffix. A streamed name is

```
base_model.model.model.layers.0.inner.self_attn.q_proj.lora_A.default.weight
                                ^^^^^^
```

against the resident

```
base_model.model.model.layers.0.self_attn.q_proj.lora_A.default.weight
```

The two diverge in the middle, so `streamed_name.endswith(resident_name)` is `False` — and `len(matches)` is `0`, which is the same `!= 1` branch that produced the error I reported. Use `dict(canonical_named_parameters(source))` and `dict(canonical_named_parameters(target))` and the fallback becomes unnecessary; a plain dict lookup matches. That is the "one line" I meant, and I should have named the function then rather than only the file.

I have not run this — no CUDA on my box and the failure needs a real streamed model — so treat it as a mechanism I traced in the source, not a reproduction. If it does run for you as written, tell me and I will look again, because then I have the name shape wrong.

## While you are in there — the guard that makes this harness trustworthy

```python
# layer_stream_runtime.py:1186
assert_canonical_parameters_intersect(model_a, model_b)
```

That exists for exactly this situation and it is worth more to a bit-exactness harness than to anything else in the repo: if the two name sets do not intersect, a per-parameter loop is **vacuously satisfied** and the harness reports bit-exactness having compared nothing. A tool whose entire output is "these two agree to 0.0" needs to be unable to say that by accident.

Your `if copied == 0: raise` catches the total case, and `make_non_vacuous_lora` is the right instinct on the other axis — but a partial intersection would still let `copy_lora` copy a subset and report success on it. Calling the assert before the comparison closes that, and it returns the shared name set so you can loop over it directly.

No rush and no other blockers from me — the harness itself is good and #379 has wanted it for a while.

Samearth17:
Thanks for the detailed review. I have pushed the requested fixes to the draft PR.

* `copy_lora()` and `make_non_vacuous_lora()` now use `canonical_named_parameters()`.
* Removed the raw parameter-name suffix matching.
* Added `--skip-lora-sync` as a built-in negative control.
* Kept the existing empty-gradient / gradient-set mismatch guards.
* Added the CUDA validation software versions to `benchmarks/README.md`.

Local `py_compile`, Ruff, `git diff --check`, and CLI validation all pass.

I can’t run the CUDA path on my machine, so I’m relying on the real CUDA validation you provided for the end to end results. The branch has been updated and is ready for review.

MakazhanAlpamys:
@Samearth17 — **first, a correction of mine.** I wrote "no CUDA on my box" in my last comment. That was wrong: this machine has an RTX 3050 Laptop 4 GB, and `py -3.10` carries torch 2.5.1+cu121 — it is the same box the earlier real-CUDA repro in this thread came from. So you are not relying on a run I described and could not repeat.

I ran your harness at `a54231ab`. **Both directions, on real hardware.**

## Positive: it completes a run, which it could not do before

```
gpu           NVIDIA GeForce RTX 3050 Laptop GPU
weights       HuggingFaceTB/SmolLM2-135M-Instruct   quant none   seq 64   steps 3
arch          llama          layers 30
adapter_tensors_copied 120
forward       max_abs_logit_diff  0.0    bit_exact  true
backward      layer0_lora_grad    7.971496e+01
              gradients exact     max_abs_diff=0.0
curve         curves_equal true   curve_max_rel=0.0e+00
meta_params   270   store 0.21 GB   pinned True   tier ram
RESULT: bit-exact gate passed                                    exit 0
```

Against the same command on the previous head, which died at `could not uniquely match LoRA parameter 'base_model.model.model.layers.0.self_attn.q_proj.lora_A.default.weight'`. `120` copied where it used to be an exception.

`layer0_lora_grad 7.971496e+01` is the number I checked hardest — a non-zero gradient means `make_non_vacuous_lora` did its job and "gradients exact" is not "both sides are zero".

## Negative: the control you added is the thing that makes the positive mean something

```
--skip-lora-sync
mode          NEGATIVE CONTROL: LoRA sync skipped
adapter_tensors_copied 0
ERROR: streamed/resident logits are not bit-exact (max_abs_diff=3.687500e+00)
                                                                 exit 1
```

`3.69` is not a marginal divergence, and the exit code flips. **The harness can report failure**, which is the whole question for a tool whose only output is "these two agree to 0.0". You added that unprompted and it is the difference between this being publishable evidence and being a script that prints a reassuring line.

## Withdrawing my own suggestion

I asked you to add `assert_canonical_parameters_intersect`. **Do not** — your version is stronger and I did not notice until I read the merged shape. Replacing the suffix fallback with a direct per-parameter raise means any target LoRA tensor with no source match fails **by name**:

```python
if source_param is None:
    raise RuntimeError(f"could not match LoRA parameter {name!r}")
```

The intersect assert only catches an **empty** intersection. Yours catches a partial one and tells you which tensor. I was proposing a weaker guard.

## Where this stands

No blockers from me. `benchmarks/README.md` has the harness row with the software versions, and the run above is the evidence #379 has been asking for since it was filed — the first published harness anyone else can point at a card and re-measure.

Flip it out of draft and I will merge on green. If you would like the transcript above in the record, paste it into `benchmarks/README.md` or the PR body as a second reproduction on a different stack (3050 / torch 2.5.1 / transformers 4.57.6 against whatever the historical run used) — but that is an offer, not a condition.

One note for the record rather than for you: the earlier failure was real and I should have named `canonical_named_parameters` the first time instead of only the file. That cost you a round.

MakazhanAlpamys:
Merged as `4021b9f`. Thank you @Samearth17 — and the thing that made this mergeable is the control you added without being asked.

## Both directions, on real CUDA (RTX 3050 Laptop, torch 2.5.1+cu121)

```
positive
  adapter_tensors_copied 120
  max_abs_logit_diff  0.0        bit_exact true
  layer0_lora_grad    7.971496e+01
  gradients exact     max_abs_diff=0.0
  curves_equal true   curve_max_rel=0.0e+00
  RESULT: bit-exact gate passed                       exit 0

negative  (--skip-lora-sync)
  adapter_tensors_copied 0
  ERROR: not bit-exact (max_abs_diff=3.687500e+00)    exit 1
```

For a tool whose only output is "these two agree to 0.0", **being able to say no is the entire question.** `--skip-lora-sync` is what turns this from a script that prints a reassuring line into publishable evidence, and `make_non_vacuous_lora` closes the other axis — `7.971496e+01` is why "gradients exact" cannot quietly mean "both sides are zero".

## The two rounds, recorded because the second repair was the interesting one

As submitted the harness could not complete a single run: `copy_lora` keyed on raw `named_parameters()`, which is not canonical on a streamed model — every wrapped layer carries an extra `.inner.` segment, so a name-keyed comparison against a resident model sees no overlap. The first repair added a suffix fallback, which also cannot work, because `.inner.` is a **middle** segment and `streamed_name.endswith(resident_name)` is therefore false. Going through `canonical_named_parameters` on both sides is the fix, and the fallback is correctly gone.

**I withdrew one of my own suggestions in your favour.** I asked for `assert_canonical_parameters_intersect`; your per-parameter raise is stronger. It catches a *partial* mismatch and names the offending tensor, where the intersect assert only fires on an empty intersection. I was proposing the weaker guard, and I would rather that be in the record than quietly dropped.

## Gate

```
files changed          benchmarks/README.md, benchmarks/harness/bitexact.py
deleted lines          0
harness README row     present, with the software versions
changelog fragment     correctly none -- no behaviour, no src/ change,
                       matching 25d17c0 and 9ce6fa2
CI                     13/13
```

## One correction of mine, since it affected you directly

I wrote "no CUDA on my box" in an earlier comment. That was wrong — this machine has an RTX 3050 and `py -3.10` carries torch 2.5.1+cu121, and it is the same box the earlier real-CUDA repro in this thread came from. You said you were relying on validation I had provided; you were right to, and I should not have cast doubt on it.

Your 3rd merge. **#601 is next** — and the item I told you was "not fixable in this PR" is now fixable: branch protection went on `main` yesterday, so `pytorch-smoke` becomes the 13th required context the moment it lands. The other two items there still stand, particularly demonstrating that the job can actually go red.

PR #598, opened 2026-08-29
fix(trackers): close telemetry SSRF guard bypass on HTTPS private IPs (#593)

### LINKED ISSUE #593: Telemetry endpoint SSRF guard cannot fire: the HTTPS precondition is upstream of a scheme-conditional private-IP check
`utils/trackers._telemetry_endpoint_is_safe` exists to stop a crafted `SOUP_POSTHOG_ENDPOINT` from pointing telemetry at an internal address. Its docstring is specific:

> Re-run the same private-IP / link-local rejection used for hub endpoints so a crafted `endpoint='https://10.0.0.1/'` cannot reach an internal network from a misconfigured caller.

That exact URL is accepted. Measured on `main` (`e0cff12`):

```python
>>> from soup_cli.utils.trackers import _resolve_posthog_target
>>> for ep in [...]: _resolve_posthog_target('phc_test', ep)

ACCEPTED  https://10.0.0.1/capture/
ACCEPTED  https://127.0.0.1/capture/
ACCEPTED  https://169.254.169.254/capture/      <- cloud metadata service
refused   http://us.i.posthog.com/capture/
ACCEPTED  https://us.i.posthog.com/capture/
```

## Why it cannot fire

The guard is two steps, and they are mutually exclusive:

```python
# trackers.py:239
if not isinstance(endpoint, str) or not endpoint.startswith("https://"):
    return False
...
validate_hub_endpoint(endpoint, hub="telemetry")
```

```python
# hubs.py:167
if parsed.scheme == "http" and host not in _LOOPBACK_HOSTS:
    if _is_private_or_link_local(host):
        raise ValueError(...)
```

`validate_hub_endpoint` only reaches `_is_private_or_link_local` on the **`http`** branch — deliberately, because for a *hub* a private HTTPS endpoint is a legitimate self-hosted deployment, which is exactly what its error message says ("private/link-local hosts require HTTPS"). `_telemetry_endpoint_is_safe` requires `https://` before delegating, so the private-IP rejection is structurally unreachable from it. The guard is not weak; it can never run.

This is the project's own recurring pattern — validated, documented, and read by nothing — with the twist that here the mechanism is a scheme check upstream of a scheme-conditional branch.

## Severity, stated honestly

Low in practice. Reaching it requires setting `SOUP_POSTHOG_ENDPOINT`, so an attacker already controls the process environment. It does **not** cross a new trust boundary on its own.

Two reasons it is still worth fixing rather than deleting the docstring:

1. Telemetry was wired to nothing until now. PR #529 (`feat: telemetry rollout`) makes the send path live for the first time, so this stops being dead code the moment any telemetry work lands.
2. A guard that reads as protective, is cited in a docstring, and cannot fire is worse than no guard, because the next reviewer stops looking.

Introduced in v0.53.10 (#154), which added the `SOUP_POSTHOG_*` overrides. Not caused by any open PR.

## Suggested fix

Do the private/link-local rejection in `_telemetry_endpoint_is_safe` itself rather than delegating a check that cannot apply. Telemetry has no legitimate self-hosted-private-endpoint case the way a model hub does, so it can be strict where `validate_hub_endpoint` cannot: reject any host that `_is_private_or_link_local` matches, on either scheme, and keep the existing HTTPS-only requirement on top.

## Acceptance criteria

- [ ] `_resolve_posthog_target` returns `None` for `https://10.0.0.1/`, `https://127.0.0.1/`, `https://169.254.169.254/`, `https://[::1]/` and an RFC1918 hostname that resolves privately
- [ ] `https://us.i.posthog.com/capture/` and a legitimate custom PostHog host still resolve — a validator that rejects the valid case is worse than the bug
- [ ] `http://` remains refused for every host
- [ ] A test asserts the **negative control**: mutating the new rejection away makes a named test fail. The current test file cannot do this, which is how the gap survived
- [ ] The `_telemetry_endpoint_is_safe` docstring matches what the code actually enforces

harshitthek:
Closes #593

## What does this PR do?

Closes the SSRF guard bypass in `_telemetry_endpoint_is_safe` where private, loopback, and link-local addresses were accepted when supplied over HTTPS.

The previous implementation delegated to `validate_hub_endpoint(endpoint, hub="telemetry")`. Because `validate_hub_endpoint` only executes its private-IP rejection branch when `scheme == "http"`, and `_telemetry_endpoint_is_safe` enforces `https://` upstream, the private-IP check was structurally unreachable. Consequently, `https://10.0.0.1/`, `https://169.254.169.254/` (cloud metadata), and `https://[::1]/` were accepted.

### Solution: Layered SSRF Guard

Per discussion on #593, this PR implements the layered design:

1. **Layer 1 (Baseline Sanitization):** Calls `validate_hub_endpoint(endpoint, hub="telemetry")` for CRLF / header injection rejection, null-byte rejection, type validation, `0.0.0.0` ambiguity handling, and scheme allowlisting.
2. **Layer 2 (Telemetry-Strict Rejection):**
   - Normalizes the parsed hostname with `(urlparse(endpoint).hostname or "").lower().rstrip(".")`.
   - Rejects loopback hosts via `_LOOPBACK_HOSTS` (`localhost`, `127.0.0.1`, `::1`), covering domain strings that `ipaddress.ip_address` cannot parse.
   - Rejects private, link-local, and reserved IPs via `_is_private_or_link_local(host)`.
   - **`hubs.py` is intentionally untouched.** The scheme-conditional check in `hubs.py` is preserved for its own callers (self-hosted model hubs).

### Behavioral Verification

```
refused   https://10.0.0.1/capture/           (was ACCEPTED)
refused   https://127.0.0.1/capture/          (was ACCEPTED)
refused   https://169.254.169.254/capture/    (was ACCEPTED)
refused   https://[::1]/capture/              (was ACCEPTED)
refused   https://localhost/capture/           (was ACCEPTED)
refused   https://LOCALHOST./capture/          (trailing-dot bypass)
refused   https://[::ffff:10.0.0.1]/          (IPv4-mapped IPv6)
refused   http://us.i.posthog.com/capture/    (unchanged)
ACCEPTED  https://us.i.posthog.com/capture/   (unchanged)
```

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Test Matrix (28 tests in `tests/test_telemetry_ssrf.py`)

- **`TestTelemetrySSRFRejection` (16 tests):** RFC 1918 Class A/B/C (`10.x`, `172.16-31.x`, `192.168.x`), loopbacks (`127.0.0.1`), cloud metadata (`169.254.169.254`), bracketed IPv6 loopbacks (`[::1]`), IPv4-mapped IPv6 (`[::ffff:10.0.0.1]`), `localhost`, uppercase `LOCALHOST`, FQDN trailing-dot `LOCALHOST.`, non-HTTPS schemes, non-string, and empty string.
- **`TestTelemetrySSRFAcceptance` (4 positive controls):** `https://us.i.posthog.com/capture/`, `https://eu.i.posthog.com/capture/`, custom public PostHog hosts, and default PostHog endpoint.
- **`TestTelemetrySSRFMutationControl` (2 negative controls):** 
  - `test_private_ip_guard_has_teeth`: Proves removing the Layer 2 `_is_private_or_link_local` check accepts `https://10.0.0.1/`.
  - `test_loopback_guard_has_teeth`: Proves `_is_private_or_link_local("localhost")` returns `False` (due to `ValueError` in `ipaddress.ip_address`), and that the explicit `_LOOPBACK_HOSTS` membership check is the load-bearing gate that catches named loopbacks.
- **`TestResolvePosthogTargetSSRF` (6 integration tests):** End-to-end resolution via `_resolve_posthog_target` across explicit parameters and `SOUP_POSTHOG_ENDPOINT` env var overrides.

## Checklist

- [x] `ruff check src/soup_cli/ scripts/ tests/` passes (0 errors, 0 warnings)
- [x] `pytest tests/test_telemetry_ssrf.py tests/test_tracker.py tests/test_v05310.py` passes (53 passed)
- [x] Updated relevant docstrings (`_telemetry_endpoint_is_safe` updated to reflect the two-layer policy)
- [x] Added changelog fragment in `changelog.d/0.73.3/598.fixed.md`
- [x] No breaking changes (all 10 existing `TestPostHogEnvOverride` tests pass unmodified)

### DIFF
--- a/src/soup_cli/utils/trackers.py
+++ b/src/soup_cli/utils/trackers.py
@@ -228,21 +228,55 @@ def _resolve_posthog_target(
 
 
 def _telemetry_endpoint_is_safe(endpoint: str) -> bool:
-    """Re-validate the telemetry endpoint via the v0.51.0 SSRF policy.
+    """Layered SSRF guard for telemetry endpoints (#593).
 
-    Even though :func:`send_telemetry_payload` only POSTs to a static
-    PostHog URL by default, callers can override ``endpoint``. Re-run the
-    same private-IP / link-local rejection used for hub endpoints so a
-    crafted ``endpoint='https://10.0.0.1/'`` cannot reach an internal
-    network from a misconfigured caller.
+    Two-layer validation:
+
+    1. :func:`~soup_cli.utils.hubs.validate_hub_endpoint` — baseline
+       sanitization (CRLF rejection, null-byte rejection, type guards,
+       ``0.0.0.0`` handling, scheme allowlist).  This layer is shared
+       with model-hub endpoints and intentionally permits private HTTPS
+       hosts for self-hosted hubs.
+    2. **Telemetry-strict rejection** — rejects loopback hosts
+       (``localhost``, ``127.0.0.1``, ``::1``), private IPs (RFC 1918),
+       and link-local addresses (``169.254.x.x``) on *any* scheme.
+       Telemetry has no legitimate self-hosted-private-endpoint case,
+       so this layer is stricter than hub validation.
+
+    The HTTPS-only requirement (``startswith("https://")``) is applied
+    first and is independent of both layers.
     """
     if not isinstance(endpoint, str) or not endpoint.startswith("https://"):
         return False
+
+    # Layer 1: baseline sanitization (CRLF, null bytes, types, 0.0.0.0).
     try:
         from soup_cli.utils.hubs import validate_hub_endpoint
 
         validate_hub_endpoint(endpoint, hub="telemetry")
     except (TypeError, ValueError):
+        return False
+
+    # Layer 2: telemetry-strict private/loopback/link-local rejection.
+    # Imports are lazy per AGENTS.md convention (heavy deps inside fns).
+    from urllib.parse import urlparse  # noqa: PLC0415
+
+    from soup_cli.utils.hubs import (  # noqa: PLC0415
+        _LOOPBACK_HOSTS,
+        _is_private_or_link_local,
+    )
+
+    # Normalise: lowercase + strip FQDN trailing dot so "LOCALHOST."
+    # matches the _LOOPBACK_HOSTS frozenset entry "localhost".
+    host = (urlparse(endpoint).hostname or "").lower().rstrip(".")
+    if not host:
+        return False
+    # _is_private_or_link_local uses ipaddress.ip_address which raises
+    # ValueError on domain strings ("localhost"), so it returns False.
+    # The explicit _LOOPBACK_HOSTS check covers named loopbacks.
+    if host in _LOOPBACK_HOSTS:
+        return False
+    if _is_private_or_link_local(host):
         return False
     return True
 
--- a/changelog.d/0.73.3/598.fixed.md
+++ b/changelog.d/0.73.3/598.fixed.md
@@ -1,0 +1,7 @@
+- [#598](https://github.com/MakazhanAlpamys/Soup/pull/598) fixes the
+  telemetry endpoint SSRF guard (_telemetry_endpoint_is_safe) which could
+  not reject private/loopback HTTPS endpoints because the upstream HTTPS
+  requirement bypassed the scheme-conditional private-IP check in
+  alidate_hub_endpoint. The fix layers a telemetry-strict
+  private/loopback/link-local rejection on top of the existing hub
+  sanitization without modifying hubs.py.
--- a/changelog.d/0.73.3/598.fixed.md
+++ b/changelog.d/0.73.3/598.fixed.md
@@ -1,7 +1,6 @@
-- [#598](https://github.com/MakazhanAlpamys/Soup/pull/598) fixes the
-  telemetry endpoint SSRF guard (_telemetry_endpoint_is_safe) which could
-  not reject private/loopback HTTPS endpoints because the upstream HTTPS
-  requirement bypassed the scheme-conditional private-IP check in
-  alidate_hub_endpoint. The fix layers a telemetry-strict
-  private/loopback/link-local rejection on top of the existing hub
-  sanitization without modifying hubs.py.
+- `_telemetry_endpoint_is_safe` could not reject private/loopback HTTPS
+  endpoints because the upstream HTTPS requirement bypassed the
+  scheme-conditional private-IP check in `validate_hub_endpoint`. The fix
+  layers a telemetry-strict private, loopback, and link-local rejection
+  on top of the existing hub sanitisation without modifying `hubs.py`
+  (#593, #598).
--- a/docs/backends-and-ops.md
+++ b/docs/backends-and-ops.md
@@ -789,7 +789,7 @@ pip install soup-cli[trackers]   # mlflow + swanlab + trackio
 
 ## Telemetry (not yet wired)
 
-Soup contains opt-in, hardware-info-only telemetry primitives in `utils/trackers.py` (`build_telemetry_payload` / `send_telemetry_payload`), but they are **not wired to any command** — no data is ever sent, and no environment variable enables sending today. When wired, the payload will carry only `soup_version` / `command` / `python` major.minor / `os` / `arch` / optional `duration_seconds` — never dataset paths, model names, or config contents — behind a 1-second hard timeout and the same HTTPS-only, private-IP-rejecting SSRF policy as hub endpoints, swallowing every exception so telemetry can never crash training. Wiring is deferred until a public privacy policy is published.
+Soup contains opt-in, hardware-info-only telemetry primitives in `utils/trackers.py` (`build_telemetry_payload` / `send_telemetry_payload`), but they are **not wired to any command** — no data is ever sent, and no environment variable enables sending today. When wired, the payload will carry only `soup_version` / `command` / `python` major.minor / `os` / `arch` / optional `duration_seconds` — never dataset paths, model names, or config contents — behind a 1-second hard timeout and the hub endpoint sanitisation plus a telemetry-strict private, loopback and link-local rejection on every scheme, swallowing every exception so telemetry can never crash training. Wiring is deferred until a public privacy policy is published.
 
 
 ## Plugin System
--- a/tests/test_telemetry_ssrf.py
+++ b/tests/test_telemetry_ssrf.py
@@ -1,0 +1,288 @@
+"""Tests for #593 - telemetry endpoint SSRF guard bypass.
+
+`_telemetry_endpoint_is_safe` must reject private/loopback/link-local
+hosts on HTTPS (not just HTTP), while still accepting legitimate public
+PostHog endpoints. The guard uses a two-layer design:
+
+1. `validate_hub_endpoint` for baseline sanitization (CRLF, null, types).
+2. Telemetry-strict private/loopback/link-local rejection on any scheme.
+
+Every test class here is structured so that **deleting** or **neutering**
+the Layer 2 guard deterministically fails at least one named test.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from soup_cli.utils.trackers import (
+    _resolve_posthog_target,
+    _telemetry_endpoint_is_safe,
+)
+
+# --- Rejection: private / loopback / link-local must be refused ----------
+
+
+class TestTelemetrySSRFRejection:
+    """Every private/loopback/link-local HTTPS endpoint must be refused."""
+
+    @pytest.mark.parametrize(
+        "endpoint",
+        [
+            "https://10.0.0.1/",
+            "https://10.255.255.255/capture/",
+        ],
+        ids=["rfc1918-10-simple", "rfc1918-10-upper"],
+    )
+    def test_rejects_private_ipv4_class_a(self, endpoint: str) -> None:
+        assert _telemetry_endpoint_is_safe(endpoint) is False
+
+    @pytest.mark.parametrize(
+        "endpoint",
+        [
+            "https://172.16.0.1/",
+            "https://172.31.255.255/",
+        ],
+        ids=["rfc1918-172-lower", "rfc1918-172-upper"],
+    )
+    def test_rejects_private_ipv4_class_b(self, endpoint: str) -> None:
+        assert _telemetry_endpoint_is_safe(endpoint) is False
+
+    @pytest.mark.parametrize(
+        "endpoint",
+        [
+            "https://192.168.0.1/",
+            "https://192.168.1.1/capture/",
+        ],
+        ids=["rfc1918-192-simple", "rfc1918-192-subnet"],
+    )
+    def test_rejects_private_ipv4_class_c(self, endpoint: str) -> None:
+        assert _telemetry_endpoint_is_safe(endpoint) is False
+
+    def test_rejects_loopback_ipv4(self) -> None:
+        assert _telemetry_endpoint_is_safe("https://127.0.0.1/") is False
+
+    def test_rejects_cloud_metadata(self) -> None:
+        """169.254.169.254 is the cloud metadata service endpoint."""
+        assert (
+            _telemetry_endpoint_is_safe("https://169.254.169.254/")
+            is False
+        )
+
+    def test_rejects_ipv6_loopback_bracketed(self) -> None:
+        """urlparse('https://[::1]/').hostname strips brackets."""
+        assert (
+            _telemetry_endpoint_is_safe("https://[::1]/") is False
+        )
+
+    def test_rejects_ipv4_mapped_ipv6(self) -> None:
+        """IPv4-mapped IPv6 (::ffff:10.0.0.1) must not bypass the guard."""
+        assert (
+            _telemetry_endpoint_is_safe(
+                "https://[::ffff:10.0.0.1]/"
+            )
+            is False
+        )
+
+    def test_rejects_localhost(self) -> None:
+        assert (
+            _telemetry_endpoint_is_safe("https://localhost/") is False
+        )
+
+    def test_rejects_localhost_uppercase(self) -> None:
+        """Hostname normalisation must be case-insensitive."""
+        assert (
+            _telemetry_endpoint_is_safe("https://LOCALHOST/") is False
+        )
+
+    def test_rejects_localhost_trailing_dot(self) -> None:
+        """FQDN trailing dot: 'LOCALHOST.' -> 'localhost' after strip."""
+        assert (
+            _telemetry_endpoint_is_safe("https://LOCALHOST./") is False
+        )
+
+    def test_rejects_http_any(self) -> None:
+        """HTTP is rejected before either layer runs."""
+        assert (
+            _telemetry_endpoint_is_safe(
+                "http://us.i.posthog.com/capture/"
+            )
+            is False
+        )
+
+    def test_rejects_non_string(self) -> None:
+        assert _telemetry_endpoint_is_safe(42) is False  # type: ignore[arg-type]
+        assert _telemetry_endpoint_is_safe(None) is False  # type: ignore[arg-type]
+
+    def test_rejects_empty_string(self) -> None:
+        assert _telemetry_endpoint_is_safe("") is False
+
+
+# --- Acceptance: legitimate public endpoints must pass -------------------
+
+
+class TestTelemetrySSRFAcceptance:
+    """Valid public PostHog endpoints must be accepted.
+
+    A validator that rejects everything is worse than the bug it fixes.
+    """
+
+    def test_accepts_public_posthog(self) -> None:
+        assert (
+            _telemetry_endpoint_is_safe(
+                "https://us.i.posthog.com/capture/"
+            )
+            is True
+        )
+
+    def test_accepts_eu_posthog(self) -> None:
+        assert (
+            _telemetry_endpoint_is_safe(
+                "https://eu.i.posthog.com/capture/"
+            )
+            is True
+        )
+
+    def test_accepts_custom_posthog_host(self) -> None:
+        """A company's self-hosted public PostHog must pass."""
+        assert (
+            _telemetry_endpoint_is_safe(
+                "https://posthog.mycompany.com/capture/"
+            )
+            is True
+        )
+
+    def test_accepts_default_posthog_endpoint(self) -> None:
+        """The hardcoded default must always be accepted."""
+        from soup_cli.utils.trackers import _POSTHOG_ENDPOINT
+
+        assert _telemetry_endpoint_is_safe(_POSTHOG_ENDPOINT) is True
+
+
+# --- Mutation controls: guard deletion must fail a named test ------------
+
+
+class TestTelemetrySSRFMutationControl:
+    """Negative controls proving the guard is load-bearing.
+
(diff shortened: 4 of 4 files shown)


harshitthek:
@MakazhanAlpamys — implemented and ready for review. Branch: `fix/593/telemetry-ssrf-guard`.

---

### What the fix does

Rewrote `_telemetry_endpoint_is_safe` with the layered design you approved:

**Layer 1** — `validate_hub_endpoint(endpoint, hub="telemetry")` runs first for baseline sanitization (CRLF, null bytes, type guards, `0.0.0.0`). Unchanged, no modifications to `hubs.py`.

**Layer 2** — telemetry-strict rejection, applied after Layer 1 passes:

```python
host = (urlparse(endpoint).hostname or "").lower().rstrip(".")
if not host:
    return False
if host in _LOOPBACK_HOSTS:
    return False
if _is_private_or_link_local(host):
    return False
return True
```

Both `_LOOPBACK_HOSTS` and `_is_private_or_link_local` are imported from `hubs.py` — reused, not duplicated. The `.lower().rstrip(".")` normalises `LOCALHOST.` → `localhost` to close the trailing-dot FQDN bypass you flagged.

The docstring was rewritten to describe the two-layer design and why each layer exists.

---

### Result

```
refused   https://10.0.0.1/capture/           (was ACCEPTED)
refused   https://127.0.0.1/capture/          (was ACCEPTED)
refused   https://169.254.169.254/capture/    (was ACCEPTED)
refused   https://[::1]/capture/              (was ACCEPTED)
refused   https://localhost/capture/
refused   https://LOCALHOST./capture/          (trailing-dot bypass)
refused   https://[::ffff:10.0.0.1]/          (IPv4-mapped IPv6)
refused   http://us.i.posthog.com/capture/    (unchanged)
ACCEPTED  https://us.i.posthog.com/capture/   (unchanged)
```

---

### Test matrix (28 tests, 4 classes)

| Class | Count | What it proves |
|---|---|---|
| `TestTelemetrySSRFRejection` | 16 | RFC1918 (10.x, 172.16-31.x, 192.168.x), loopback, cloud metadata, `[::1]`, `[::ffff:10.0.0.1]`, `localhost`, `LOCALHOST`, `LOCALHOST.`, HTTP, non-string, empty |
| `TestTelemetrySSRFAcceptance` | 4 | `us.i.posthog.com`, `eu.i.posthog.com`, custom host, hardcoded default — validator must NOT reject valid endpoints |
| `TestTelemetrySSRFMutationControl` | 2 | Neutering Layer 2 makes `10.0.0.1` pass. Proves `_is_private_or_link_local("localhost")` returns `False` (the gap your original suggestion would have had) and the `_LOOPBACK_HOSTS` check is what saves it. |
| `TestResolvePosthogTargetSSRF` | 6 | End-to-end through `_resolve_posthog_target` with explicit and env-var endpoints |

For the mutation controls you said you'd check: deleting the `_is_private_or_link_local(host)` branch fails `test_private_ip_guard_has_teeth`. Deleting the `host in _LOOPBACK_HOSTS` branch fails `test_loopback_guard_has_teeth`. Both test names are explicit about what they kill.

---

### Adversarial self-audit: what I checked after finishing

Assumed the fix wasn't good enough and searched the codebase for breakage and gaps:

**Breakage: none.** The only callers of `_telemetry_endpoint_is_safe` and `_resolve_posthog_target` are inside `trackers.py` itself. The only external importers of `trackers.py` (`train.py`) import `resolve_report_to` and `tracker_missing_dep_message`, which are unrelated. All 10 existing `TestPostHogEnvOverride` tests in `test_v05310.py` pass unchanged. No circular import risk — `hubs.py` never imports `trackers.py`.

**Codebase-wide SSRF pattern observations** (informational, no action in this PR):

1. **`webhooks.py` already fixed this exact pattern.** Its `validate_webhook_url` checks private IPs on BOTH schemes, and the comment at line 87-89 literally documents it as "the pre-fix bug." The telemetry guard now matches that design.

2. **Three independent copies of `_is_private_or_link_local` exist** across `hubs.py`, `webhooks.py`, and `hf.py`. They differ in strictness — `webhooks.py` additionally catches `is_reserved` and `is_multicast`. This PR imports the `hubs.py` version (which matches your original design), but a future deduplication pass could consolidate them with the strictest predicate.

3. **`energy.py` looks like it has the HTTP-only bug but doesn't** — its private-IP check is outside the `scheme == "http"` block. Verified, not a bug.

4. **`_LOOPBACK_HOSTS` doesn't include trailing-dot variants** (`"localhost."`). The `.rstrip(".")` in this fix handles it locally. If a future caller checks `host in _LOOPBACK_HOSTS` without normalising, the trailing-dot bypass reopens — but that's out of scope since you said not to touch `hubs.py`.

---

### Files changed

| File | Change |
|---|---|
| `src/soup_cli/utils/trackers.py` | `_telemetry_endpoint_is_safe` rewrite (+41, -7) + docstring |
| `tests/test_telemetry_ssrf.py` | 28 new tests (NEW) |
| `changelog.d/0.73.3/593.fixed.md` | Changelog fragment (NEW) |

`hubs.py` untouched.

src/soup_cli/utils/trackers.py
```diff
@@ -228,22 +228,56 @@ def _resolve_posthog_target(
...
+    # matches the _LOOPBACK_HOSTS frozenset entry "localhost".
+    host = (urlparse(endpoint).hostname or "").lower().rstrip(".")
+    if not host:
+        return False
+    # _is_private_or_link_local uses ipaddress.ip_address which raises
+    # ValueError on domain strings ("localhost"), so it returns False.
+    # The explicit _LOOPBACK_HOSTS check covers named loopbacks.
+    if host in _LOOPBACK_HOSTS:
+        return False
+    if _is_private_or_link_local(host):
```
AmirF194, line 279: I ran `_telemetry_endpoint_is_safe` against this PR's head (`a17d64d9`) in a clean container.

`_is_private_or_link_local` only parses the host as a literal IP address (`ipaddress.ip_address`), so any hostname that is not a literal IP and not in `_LOOPBACK_HOSTS` passes layer 2 no matter what it resolves to:

```
$ docker run --rm -v $PWD:/repo -w /repo python:3.12-slim python3 -c "
from soup_cli.utils.trackers import _telemetry_endpoint_is_safe
print(_telemetry_endpoint_is_safe('https://attacker-controlled.example/capture/'))"
True
```

#593's own acceptance criteria list "an RFC1918 hostname that resolves privately" as a case `_resolve_posthog_target` must reject, and the added test file has no case for it. `_is_private_or_link_local`'s docstring says DNS resolution is intentionally skipped, matching the existing `hf.py` policy, so this may be a deliberate scope line rather than a gap. Worth confirming before merge: is DNS-based resolution meant to stay out of scope here, or does that criterion still need a follow-up?

MakazhanAlpamys:
Thank you, @harshitthek — the security fix is correct, it matches the design we agreed point for point, and it survived every mutation I ran. Two small non-security fixes and I will merge.

## What I verified

The bug reproduces on `main` and is closed here. Driving the real `_resolve_posthog_target` on both trees:

| endpoint | main `0d98183` | PR `a17d64d9` |
|---|---|---|
| `https://10.0.0.1/` | ACCEPTED | refused |
| `https://127.0.0.1/` | ACCEPTED | refused |
| `https://169.254.169.254/` | ACCEPTED | refused |
| `https://[::1]/` | ACCEPTED | refused |
| `https://localhost/` | ACCEPTED | refused |
| `https://LOCALHOST./` | ACCEPTED | refused |
| `https://192.168.1.1/` | ACCEPTED | refused |
| `https://[::ffff:10.0.0.1]/` | ACCEPTED | refused |
| `https://us.i.posthog.com/capture/` | ACCEPTED | **ACCEPTED** |
| `https://telemetry.example.com/capture/` | ACCEPTED | **ACCEPTED** |
| `http://us.i.posthog.com/capture/` | refused | refused |

Closed for every vector, with no over-rejection. The IPv4-mapped `[::ffff:10.0.0.1]` case was not something I asked for and it is a good addition.

**The mutation I care most about is the one I specifically asked you to cover**, and I ran it myself rather than reading the test names:

```
# _LOOPBACK_HOSTS half removed, leaving only _is_private_or_link_local
FAILED test_rejects_localhost
FAILED test_rejects_localhost_uppercase
FAILED test_rejects_localhost_trailing_dot
FAILED test_loopback_guard_has_teeth
FAILED test_env_override_localhost_blocked
5 failed, 23 passed
```

Exactly the `localhost` set and nothing else — it discriminates precisely rather than failing everything. Two further mutations also died: deleting Layer 2 entirely (19 failures) and making the validator reject every host (6 failures, all in `TestTelemetrySSRFAcceptance`), so the suite catches a validator that is *too strict* as well as one that is too loose. That second direction is what I meant about a validator that rejects the valid case being worse than the bug.

Design compliance, all six points: `validate_hub_endpoint` runs first and Layer 2 sits on the sanitised host; both halves of the predicate are present and both are load-bearing; `hubs.py` is untouched (`git diff --name-only | grep -c hubs.py` is `0`); the positive direction is tested; bracketed IPv6, uppercase and trailing-dot are all covered; the docstring is rewritten to describe what the code enforces.

Also checked: no existing test deleted or weakened (`git diff origin/main...a17d64d9 -- tests/` has zero removals), all 25 test functions contain a real `assert` against the real functions with no mocking of the module under test, no new top-level import (`test_cli_startup_is_light` 34 passed), `test_v0430_part_a.py` 68 passed with no regression, `ruff` clean, `test_telemetry_ssrf.py` 28 passed.

## Two things to fix

**1. The changelog fragment contains a control character.** A backslash before `validate` was consumed as an escape, leaving a literal vertical tab:

```
$ git show HEAD:changelog.d/0.73.3/598.fixed.md | xxd-equivalent
b'... private-IP check in\n  \x0balidate_hub_endpoint. The fix layers ...'
control chars (excl LF/CR): [(284, '0xb')]
```

It renders as `check in ␋alidate_hub_endpoint`, and it is the only fragment out of 300+ in the repo carrying any control byte. `scripts/assemble_changelog.py` copies fragments verbatim into `CHANGELOG.md`, so this would ship. Please drop the backslash and backtick the identifiers, matching sibling style (`586.fixed.md`, `522.fixed.md`). While you are there, please reference `#593` as well as `#598` — the fragment names the PR but not the issue.

**2. `docs/backends-and-ops.md:792` is now inaccurate.** It says the telemetry endpoint uses

> the same HTTPS-only, private-IP-rejecting SSRF policy as hub endpoints

Telemetry is now deliberately **stricter** than the hub policy — that is the whole point of the layered design and of leaving `hubs.py` alone. Leaving that sentence in place is how someone later "simplifies" this back in the wrong direction. Something like *"the hub endpoint sanitisation plus a telemetry-strict private, loopback and link-local rejection on every scheme"* would be accurate. (Worth knowing: that sentence was already wrong before your PR, since it claimed a private-IP rejection that could not fire. Your change makes the first half true and the second half false.)

## Non-blocking notes

- `.lower()` in `(urlparse(endpoint).hostname or "").lower().rstrip(".")` is dead — `urlparse().hostname` already lowercases, so only `.rstrip(".")` is load-bearing. Consequence: `test_rejects_localhost_uppercase` passes even with the whole normalisation removed, so it pins a contract against a future `urlparse` change rather than testing this PR. Harmless, keep it, just do not count it as coverage.
- `TestTelemetrySSRFMutationControl::test_private_ip_guard_has_teeth` hand-copies a neutered reimplementation inline, which will not track a refactor of the real function. It is not vacuous — it also asserts on the real function — and my external mutation proved the coverage independently. A note, not a request.
- `# noqa: PLC0415` is inert: this repo's ruff `select` is `E, F, I, N, W`, so that rule is not enabled. Cosmetic.
- Importing `hubs._LOOPBACK_HOSTS` / `_is_private_or_link_local` makes `trackers.py` the first cross-module consumer of those private symbols, where five other modules keep their own copies. Given this repo's history with duplicated sources of truth, importing is the right direction and `test_loopback_guard_has_teeth` pins the contract, so an upstream change would be caught loudly. No action.

## On #593's acceptance criteria

Five of six are met. The sixth — *"an RFC1918 hostname that resolves privately"* — is **not** met, because the fix does no DNS resolution at all. **That is my miss, not yours:** I wrote that bullet in the issue and then agreed a resolution-free predicate with you in the thread, which supersedes it. Saying so here so it is on the record rather than closed quietly.

Two follow-ups I will file separately, neither of them yours to fix here: DNS-resolution-based rejection, and abbreviated / decimal IPv4 host forms (`https://127.1/`, `https://2130706433/`) which `ipaddress.ip_address` also raises on — the same mechanism as the `localhost` hole you caught, one layer further out. I could not confirm whether glibc's `getaddrinfo` accepts those forms from this Windows box, so I am not claiming they are live in production, only that they are outside the predicate.

Push those two fixes and I will re-run the mutation set and merge.

harshitthek:
---

Both fixes pushed in `b4f9227`.

**1. Changelog fragment** — dropped the escaped backslash that injected `\x0b`, backticked identifiers to match sibling style (`586.fixed.md`, `522.fixed.md`), added `#593` reference alongside `#598`.

**2. Docs line 792** — replaced *"the same HTTPS-only, private-IP-rejecting SSRF policy as hub endpoints"* with *"the hub endpoint sanitisation plus a telemetry-strict private, loopback and link-local rejection on every scheme"*.

---

On the non-blocking notes — understood and agreed on all four:

- `.lower()` dead code: keeping it as a contract pin against a future `urlparse` change, not counting it as coverage.
- Inline neutered reimplementation in the mutation test: noted — the real function assertion and your external mutation prove coverage independently.
- `# noqa: PLC0415` inert: cosmetic, leaving as-is.
- Cross-module private imports: agreed that importing is the right direction over a sixth copy, and `test_loopback_guard_has_teeth` pins the contract.

---

Filed the two follow-ups from your acceptance criteria reconciliation:

- **#599** — [DNS-resolving hostnames bypass `_is_private_or_link_local`](https://github.com/MakazhanAlpamys/Soup/issues/599)
- **#600** — [Abbreviated/decimal IPv4 forms bypass `ipaddress.ip_address()`](https://github.com/MakazhanAlpamys/Soup/issues/600)

MakazhanAlpamys:
Merged as `2ee8e13`. Thank you, @harshitthek — and I want to be specific about what made this one good, because it is not the code.

**You improved on the fix I proposed instead of implementing it.** My sketch in #593 said to do the private/link-local rejection inside `_telemetry_endpoint_is_safe` rather than delegating. You pointed out that a standalone replacement loses the CRLF, null-byte, type and `0.0.0.0` handling — and, the part that actually mattered, that `_is_private_or_link_local("localhost")` returns `False` because `ipaddress.ip_address` raises on domain names. I verified it before agreeing:

```
_LOOPBACK_HOSTS = frozenset({'127.0.0.1', '::1', 'localhost'})
  10.0.0.1         -> True
  127.0.0.1        -> True
  169.254.169.254  -> True
  ::1              -> True
  localhost        -> False    <- my fix would have shipped this hole
```

A guard that blocks `127.0.0.1` and waves through its own hostname is exactly the half-fix that reads as protection and is not — the same shape as the bug we were closing. And you asked which direction I wanted before writing any of it.

**Re-verified on the final head `b4f92274`.** The only change since the head I reviewed is the two files I asked for — `git diff a17d64d9 b4f92274 --stat` is `changelog.d/0.73.3/598.fixed.md` and `docs/backends-and-ops.md`, nothing else — so `trackers.py` and the tests are byte-identical to what I had already mutation-tested. I re-ran anyway:

```
baseline (test_telemetry_ssrf.py + test_v0430_part_a.py)  -> 96 passed

# _LOOPBACK_HOSTS half removed
FAILED test_rejects_localhost
FAILED test_rejects_localhost_uppercase
FAILED test_rejects_localhost_trailing_dot
FAILED test_loopback_guard_has_teeth
FAILED test_env_override_localhost_blocked
5 failed, 23 passed
```

Exactly the `localhost` set and nothing else. The other two mutations from my review still hold: deleting the strict layer fails 19 named tests, and making the validator reject every host fails the six acceptance tests — so the suite catches a validator that is too strict as well as one that is too loose. That second direction is the one most security fixes forget.

Both requested fixes are clean: the `\x0b` control byte is gone from the fragment, it now cites `#593` as well as `#598`, and `docs/backends-and-ops.md:792` no longer claims telemetry uses "the same SSRF policy as hub endpoints" when it is deliberately stricter.

Also confirmed at merge: zero deleted or weakened test lines, `hubs.py` untouched, no new top-level import, all 13 checks green on the merged head.

You are in `CONTRIBUTORS.md` as of `527f823`.

Two follow-ups I said I would file and have not forgotten — neither is yours to fix, both are one layer further out than the predicate we agreed:

- abbreviated and decimal IPv4 host forms (`https://127.1/`, `https://2130706433/`) are still accepted, same `ipaddress.ip_address`-raises mechanism as the `localhost` hole you caught. I could not confirm from a Windows box whether glibc's `getaddrinfo` accepts them, so I will file it as needing that check rather than as a live bypass.
- no DNS resolution, so a public hostname resolving to an RFC1918 address is not caught. That bullet was in #593's acceptance criteria and I superseded it myself when we agreed a resolution-free predicate — recorded on the PR so it is not lost.

If you want either of them, say so on the issues once they are up.

harshitthek:
Thank you for the thorough review and testing, @MakazhanAlpamys! Always appreciate the rigorous mutation verification and feedback on design choices.

Small note on the record: this is actually my 7th merged PR here (following #428, #448, #451, #458, #457, and #467) — always glad to keep contributing and keeping the guards tight.

Also, I went ahead and opened the two follow-up issues you noted so they wouldn't get lost:

- **#599** — [Telemetry SSRF guard: DNS-resolving hostnames bypass `_is_private_or_link_local`](https://github.com/MakazhanAlpamys/Soup/issues/599)
- **#600** — [Telemetry SSRF guard: abbreviated/decimal IPv4 forms bypass `ipaddress.ip_address()`](https://github.com/MakazhanAlpamys/Soup/issues/600)

Please take a look and feel free to adjust the framing, titles, or scope if you'd like anything worded differently or tagged with specific labels.

MakazhanAlpamys:
@harshitthek — you raised this on 2026-08-30 and it sat for a week. **Fixed now in `eb042c0`.**

You were right on the facts and I checked them before touching anything:

```
merged PRs by @harshitthek: #428, #448, #451, #457, #458, #467, #598  -> 7
(#450 was closed unmerged, which is why the count is 7 and not 8)

CONTRIBUTORS.md before: two separate "Harshit Sharma" headers
  line 129 -> 6 bullets, at your real position in the list
  line 236 -> 1 bullet, further down, reading as a first contribution
CONTRIBUTORS.md after:  one header, 7 bullets
```

`CONTRIBUTORS.md` is ordered by first contribution, so line 129 was the correct position and the later header was the mistake. The #598 bullet moved up to join the other six and the duplicate header is gone. **No text was rewritten** — your seventh entry reads exactly as it did, just in the right place.

The cause was mine and it is worth naming rather than hand-waving: I add these entries per merge, and on #598 I appended a new person block instead of finding the existing one. Nothing in the file or in CI catches a duplicated name, which is why it survived a week of edits after you pointed at it.

Apologies for the delay in acting on it. Seven merged PRs, and the telemetry SSRF one in particular — a guard that existed, was cited in its own docstring, and could never fire — deserved better bookkeeping than that.

Issue #596, opened 2026-08-29
test_sft_smoke and test_dpo_smoke run in neither the default suite nor CI, which is how one stayed red for five releases


MakazhanAlpamys:
`tests/test_smoke_train.py::test_sft_smoke` and `::test_dpo_smoke` execute in **neither** the default suite nor CI. Nothing runs them, on any platform, ever.

Found by @chadetov while fixing #594, and split out here because the fix there was the symptom, not the cause.

## Why nothing runs them

Two independent exclusions, and both are intentional in isolation:

```toml
# pyproject.toml:256
addopts = "-m 'not smoke' --cov=soup_cli --cov-fail-under=77 ..."
```

```yaml
# .github/workflows/ci.yml:68
run: pytest -o addopts= -m smoke -k mlx_sft_smoke -q tests/test_smoke_train.py
```

The default suite drops everything marked `smoke`. The only job that clears `addopts` narrows to `-k mlx_sft_smoke`, and it runs on an install that asserts the PyTorch stack is **absent** (`ci.yml:62-68`), so it could not run the torch smoke tests even without the `-k`. The `test` matrix at `ci.yml:117` and `:259` inherits the default `addopts` and therefore skips them too.

## What it cost

`test_sft_smoke` was red from v0.36.0 until #594 — five releases. The fixture set `data.format: chatml` but not `data.chat_template`, and v0.36.0 replaced the silent `f"{role}: {content}"` fallback with a hard error, so the run died at `sft_format.py:183`. Nothing noticed, because nothing looked.

That is the shape of the risk: these two tests are the only end-to-end coverage of `soup train` on the PyTorch path — model load, LoRA attach, data formatting, an actual training step. Everything else in the suite stubs at least one of those. A regression in that chain is invisible today until a user hits it.

## The decision this needs

Running them means model downloads and runner minutes on every push, which is why @chadetov correctly left it out of #594 rather than deciding it inside a fixture fix. Options, roughly in increasing cost:

1. **A separate scheduled job** (nightly / weekly `cron`) running `pytest -o addopts= -m smoke` with a HF cache restored from `actions/cache`. Keeps PR latency untouched; a break is found within a day rather than five releases.
2. **A `pytorch-smoke` job on PRs**, ubuntu-only, single Python version, cached models. `sshleifer/tiny-gpt2` is tiny; the real cost is the torch install, which the existing `test` job already pays.
[...]

PR #601, opened 2026-08-30
ci: add required PyTorch smoke job

Samearth17:
## Summary

Implements #596 by adding a dedicated `pytorch-smoke` CI job that runs the real PyTorch SFT and DPO smoke tests on pull requests.

The job is intentionally separate from the existing test matrix and `mlx-smoke` job:

* Ubuntu only
* Python 3.11
* dedicated Hugging Face cache via `actions/cache`
* warms `sshleifer/tiny-gpt2`
* runs `python -m pytest -o addopts= -m smoke -q tests/test_smoke_train.py`
* selects both `test_sft_smoke` and `test_dpo_smoke`

No changes were made to the existing test matrix or MLX smoke job.

## Validation

Local PyTorch smoke run:

`2 passed, 1 skipped`

The skipped test is the MLX smoke test; the PyTorch SFT and DPO tests both passed.

Workflow YAML parsing:

`YAML OK`

`git diff --check`:

passed.

## Regression proof

The SFT smoke path was tested against the regression from #594.

With `chat_template: chatml` present, the SFT smoke test passed.

After deliberately removing `chat_template: chatml`, the SFT smoke test failed with:

`ValueError: Tokenizer has no chat_template`

The fixture was then restored and the SFT smoke test passed again.

This demonstrates that the smoke test catches the regression rather than merely executing successfully.

## Scope

This PR implements the dedicated PR-triggered PyTorch smoke job requested in #596. The existing `mlx-smoke` job remains unchanged, and the smoke tests remain excluded from the default suite.

Refs #596

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,8 +108,7 @@ jobs:
           )
 
       - name: Run PyTorch smoke tests
-        run: python -m pytest -o addopts= -m smoke -q tests/test_smoke_train.py
-
+        run: python -m pytest -o addopts= -m smoke -q tests/
       - name: Save PyTorch smoke HF cache
         if: success() && steps.hf-cache.outputs.cache-hit != 'true'
         uses: actions/cache/save@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,56 @@ jobs:
           python -c "import importlib.util; names=('torch','trl','datasets','peft','accelerate','bitsandbytes'); present=[name for name in names if importlib.util.find_spec(name)]; assert not present, f'unexpected training dependencies: {present}'"
       - name: Run one-step MLX SFT through the real CLI
         run: pytest -o addopts= -m smoke -k mlx_sft_smoke -q tests/test_smoke_train.py
+  pytorch-smoke:
+    # #596: exercise the real PyTorch SFT + DPO training pipeline on every PR.
+    # Keep this separate from the 3x3 matrix so the expensive model-backed
+    # smoke tests run once, not nine times.
+    runs-on: ubuntu-latest
+    env:
+      PYTHONUTF8: "1"
+      PYTHONIOENCODING: "utf-8"
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Set HF cache location
+        run: echo "HF_HOME=$RUNNER_TEMP/hf-cache" >> "$GITHUB_ENV"
+
+      - name: Restore PyTorch smoke HF cache
+        id: hf-cache
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{ runner.temp }}/hf-cache
+          key: hf-pytorch-smoke-${{ runner.os }}-${{ hashFiles('tests/test_smoke_train.py', 'pyproject.toml') }}
+          restore-keys: |
+            hf-pytorch-smoke-${{ runner.os }}-
+
+      - name: Install training dependencies
+        run: pip install -e ".[dev]" "torch>=2.6"
+
+      - name: Warm tiny GPT-2 cache
+        shell: python
+        run: |
+          from huggingface_hub import snapshot_download
+
+          snapshot_download(
+              "sshleifer/tiny-gpt2",
+              allow_patterns=["*.json", "*.txt", "*.bin"],
+          )
+
+      - name: Run PyTorch smoke tests
+        run: python -m pytest -o addopts= -m smoke -q tests/test_smoke_train.py
+
+      - name: Save PyTorch smoke HF cache
+        if: success() && steps.hf-cache.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: ${{ runner.temp }}/hf-cache
+          key: hf-pytorch-smoke-${{ runner.os }}-${{ hashFiles('tests/test_smoke_train.py', 'pyproject.toml') }}
 
   # #502/#503/#522/#571: install the exact Torch 2.5.1 / Transformers 5.16.1 /
   # trl 0.29 / PEFT 0.20 support stack and Plotext 6.0.0 so pip cannot silently


MakazhanAlpamys:
@Samearth17 — the job is written correctly and is demonstrably green on real CI. Three things before I merge, and one of them is not yours to fix.

## Against the four requirements I set on #596

| # | Requirement | Status |
|---|---|---|
| 1 | Option 2 — dedicated job, PRs, ubuntu-only, single Python, HF cache | **met** |
| 2 | Must be required to be green | **not met, and not fixable in this PR** — see below |
| 3 | Demonstrate it fails | **asserted, not shown** |
| 4 | `-m smoke`, not `-k` | **met** |

The diff is one purely additive hunk, `+50/-0`, no existing job touched — I read the resulting file rather than trusting the diff: jobs parse as `['lint', 'type-check', 'mlx-smoke', 'pytorch-smoke', 'transformers-floor', 'test']` and `mlx-smoke`'s `-k mlx_sft_smoke` line is untouched context. YAML parses, `actionlint` accepts it. `pytestmark = pytest.mark.smoke` at module level means `-m smoke` selects all three, `test_mlx_sft_smoke` self-skips via `importorskip`, and the real run shows `2 passed, 1 skipped` — so **`test_dpo_smoke` is genuinely included and the #596 root cause is closed.**

## On requirement 2 — I checked, and nothing in this repo is required today

```
$ gh api repos/MakazhanAlpamys/Soup/branches/main/protection
404 Branch not protected
$ gh api repos/MakazhanAlpamys/Soup/rulesets
[]
```

Not even the nine-cell `test` matrix. So this is a repository-settings action only I can take, and I will take it on merge. Your PR **title** promises "required" while the body never mentions it — please say explicitly that the settings half is mine, so the title is not writing a cheque the diff cannot cash.

## On requirement 3 — this is the one I said I would check hardest

The body has a "Regression proof" section claiming the SFT test failed with `ValueError: Tokenizer has no chat_template` and passed after restoring. There is no captured output, no log, no run link, and the PR has one commit and zero comments. The quoted string is genuine — `src/soup_cli/data/sft_format.py:184` raises exactly that — so I believe you. But a CI job whose whole purpose is "a test nobody runs is a test that does not exist" cannot itself ship on an unshown demonstration.

Either paste the transcript with both directions and the command, or push the reverted-fixture commit, link the red `pytorch-smoke` run, and revert it. The second is better and costs you one push.

## The substantive code ask: torch >= 2.6

`[train]` declares only `torch>=2.0.0`. Today's resolver saves you — the real CI run installed torch 2.13.0 / transformers 5.16.1 and passed in 1m54s. But I reproduced the failure mode on my box (torch 2.5.1+cu121):

```
E  ValueError: Due to a serious vulnerability issue in `torch.load` ... upgrade torch to at least v2.6
   transformers/utils/import_utils.py:1647 (check_torch_load_is_safe)
2 failed, 1 skipped
```

`sshleifer/tiny-gpt2` ships no safetensors, so both smoke tests die at model load, nowhere near a training assertion. And `transformers-floor` pins `torch==2.5.1`, so this repo **does** exercise a sub-2.6 stack — the day anything constrains this job's resolver it goes red with a security error that reads nothing like a regression. Pin `torch>=2.6` on the install step, or put a comment in the job naming CVE-2025-32434 and this failure mode.

## Non-blocking

- The `actions/cache` you added caches `sshleifer/tiny-gpt2` — about 2 seconds of the run. The real cost is the **526 MB torch wheel**, roughly 88 s of the 114 s job, which is not cached. `setup-python`'s `cache: pip` is the actual win.
- Cache key `hashFiles('tests/test_smoke_train.py', 'pyproject.toml')` invalidates on any `pyproject.toml` edit, so it will rarely hit.
- No blank line before `pytorch-smoke:`; every other job in the file has one.
- **No changelog fragment, and that is correct** — I checked the last 20 commits touching `.github/workflows/ci.yml` and every one shipped none.

## One thing across both of your open PRs

Your #597 has the same shape: the harness is one line from working, but as submitted it cannot complete a run, while the PR describes results it did not produce. Here the job genuinely works and the proof is missing; there the proof is described and the code cannot reach it. **The code is good in both — the evidence is asserted rather than shown.** For a contributor whose two PRs are both about making the project's claims reproducible, that is the one habit worth changing, and it is a cheap one: paste the transcript.

MakazhanAlpamys:
@Samearth17 — item 2 changed today, and it changed in your favour. Update so you can finish this.

## "Not met, and not fixable in this PR" is now fixable

I said your job could not be **required** to be green and that this was not yours to fix. The reason was that `main` had no branch protection at all:

```
$ gh api repos/MakazhanAlpamys/Soup/branches/main/protection
404  Branch not protected
$ gh api repos/MakazhanAlpamys/Soup/rulesets
[]
```

Nothing was required — which is also why red CI has been mergeable in this repository. That is now closed:

```
required_status_checks.strict     false
enforce_admins                    false
contexts (12):
  lint  mlx-smoke  transformers-floor
  test (ubuntu-latest, 3.10 / 3.11 / 3.12)
  test (macos-latest,  3.10 / 3.11 / 3.12)
  test (windows-latest, 3.10 / 3.11 / 3.12)
```

`type-check` is deliberately excluded (mypy is non-blocking by design and always green, so requiring it would assert nothing) and so is `validate-recipes` (path-filtered — a non-recipe PR never runs it, and requiring it would hang those PRs forever).

**`pytorch-smoke` is not in that list only because it does not exist on `main` yet** — adding a required context for a job no branch produces would block every open PR. The moment this merges I add it as the 13th, and I will post the API response here so it is checkable rather than promised. That is the whole of item 2, and it becomes a one-command follow-up instead of an argument.

Which means: your requirement 2 is now **met by the repository**, conditional on merge. Please update the PR body — the framing "cannot be required" is no longer true and I do not want it quoted later as though the gate were still theoretical.

## The other two items stand, unchanged

**3 — demonstrate it fails.** Still the blocker, and now more load-bearing than when I wrote it: a job that is about to become a required check gets to block merges for everyone in this repository. A check that cannot go red would do that while proving nothing, and would be worse than no check because it reads as coverage. Push the reverted-fixture commit, link the red `pytorch-smoke` run, revert. One push.

**The `torch >= 2.6` ask.** Unchanged and, again, now sharper — `transformers-floor` pins `torch==2.5.1`, so this repo does exercise a sub-2.6 stack, and a required job that fails with a `torch.load` security error reading nothing like a regression would block the queue on a resolver change nobody made deliberately. Pin it or name CVE-2025-32434 in a comment.

## On the note I ended with last time

I said the habit worth changing across both your PRs was that the evidence is asserted rather than shown. That still stands, and I want to be even-handed about it: I have just enabled a control I have been describing as needed for a while, and the reason it took this long is the same category of thing. It is easier to notice in someone else's diff.

Nothing here needs re-measuring — one push for item 3, one line for the torch pin, one edit to the body.

Samearth17:
I have pushed the deliberate regression commit (`5c45186`) with `chat_template: chatml` removed so the new `pytorch-smoke` job can demonstrate that it actually catches the failure.

At the moment the checks are still queued and waiting to run, so I don’t have the RED result yet. Once the maintainer-run CI completes, I’ll use that result to confirm the failure mode and then move in the corresponding direction: restore the fixture and verify the job goes green if the expected regression is caught, or adjust the workflow/test setup if the failure shows a CI-specific issue instead.

After that I will update the PR with the actual RED → GREEN evidence rather than just describing the local result.

MakazhanAlpamys:
Your RED → GREEN evidence is already in, and it is clean — I pulled it from the run logs so you do not have to wait on me again:

**RED**, run `33411146021` on `5c45186` (the `chat_template: chatml` removal):

```
pytorch-smoke     <- the ONLY failing job in the run
tests/test_smoke_train.py::test_sft_smoke
E  ValueError: Tokenizer has no chat_template. Pass data.chat_template: chatml ...
src/soup_cli/data/sft_format.py:183: ValueError
1 failed, 1 passed, 1 skipped in 9.22s
```

**GREEN**, run `33451924301` on `b6b5d7e` (fixture restored, current head):

```
pytorch-smoke  success
2 passed, 1 skipped in 11.47s
```

That is exactly the proof I asked for, and better than the usual version of it in two ways. First, `pytorch-smoke` was the **only** red job in the RED run — nothing else moved — so the job is attributable to the regression rather than to CI weather. Second, the failure is the *specific* one the fixture was designed to provoke, at `sft_format.py:183`, not a generic import or collection error. A smoke job that goes red for the wrong reason is worth roughly nothing, and this one did not.

I also checked the thing you did not claim, because it is the half of #596 that is easy to lose: both `test_sft_smoke` **and** `test_dpo_smoke` are collected and pass in the GREEN run. #596 names both as running in neither the default suite nor CI, so a job covering only the SFT one would have closed half the issue while reading as a full fix. It covers both.

Two things before I can merge:

1. **Un-draft it**, and squash out or clearly label the deliberate-regression commit — `5c45186` must not land on `main` as history that removes a working fixture.
2. Say in the PR body whether `pytorch-smoke` should become a **required** check on branch protection. Your title says "required" and I agree, but that is a repo-settings change I make after the merge, and I want it stated so the next person reading the PR knows the job is a gate rather than an advisory. Note the practical consequence: once required, a red smoke blocks every merge, so I want your view on runtime — the run is ~11 s of tests, which is cheap, but tell me if the install step dominates.

Thank you for pushing a commit whose whole purpose is to fail. Most CI PRs assert that a job exists; almost none demonstrate that it can catch anything, and this repo has been bitten more than once by a guard that could not fail.

Samearth17:
## Summary

Implements #596 by adding a dedicated `pytorch-smoke` CI job that runs the real PyTorch SFT and DPO smoke tests on pull requests.

The job is intentionally separate from the existing test matrix and `mlx-smoke` job:

* Ubuntu only
* Python 3.11
* dedicated Hugging Face cache via `actions/cache`
* warms `sshleifer/tiny-gpt2`
* installs `torch>=2.6`
* runs `python -m pytest -o addopts= -m smoke -q tests/test_smoke_train.py`
* selects both `test_sft_smoke` and `test_dpo_smoke`

No changes were made to the existing test matrix or MLX smoke job.

## Validation

Local PyTorch smoke run:

`2 passed, 1 skipped`

The skipped test is the MLX smoke test; the PyTorch SFT and DPO tests both passed.

Workflow YAML parsing:

`YAML OK`

`git diff --check`:

passed.

## Regression proof

The SFT smoke path was tested against the regression from #594.

**RED:** commit `5c45186` temporarily removed `chat_template: chatml` from the SFT fixture. The `pytorch-smoke` job was the only failing CI job in that run, failing with:

`ValueError: Tokenizer has no chat_template`

CI run: https://github.com/MakazhanAlpamys/Soup/actions/runs/33411146021

**GREEN:** commit `b6b5d7e` restored the fixture. The `pytorch-smoke` job then completed successfully with:

`2 passed, 1 skipped`

CI run: https://github.com/MakazhanAlpamys/Soup/actions/runs/33451924301

Both `test_sft_smoke` and `test_dpo_smoke` pass in the green run.

This demonstrates that the smoke job catches the #594 regression and returns to green once the fixture is restored.

## Required check

`pytorch-smoke` is intended to become a required status check on `main`.

The repository branch-protection/ruleset configuration is a maintainer-side settings change to be applied after this PR merges. The maintainer has already enabled the existing required checks; `pytorch-smoke` will be added as the next required context once this job exists on `main`.

The verified smoke tests themselves take about 11 seconds. The dependency installation step dominates the total runtime, primarily due to the PyTorch installation.

## Scope

This PR implements the dedicated PR-triggered PyTorch smoke job requested in #596. The existing `mlx-smoke` job remains unchanged, and the smoke tests remain excluded from the default suite.

Refs #596

MakazhanAlpamys (changes requested):
**The job is correct, honest and cheap, and I could verify the thing that actually matters: it would go red.** One line stands between this and a merge, and it is the issue's own headline, so I would rather ask than file it.

## What I verified

The exclusion is `pytestmark = pytest.mark.smoke` at `tests/test_smoke_train.py:15` against `addopts = "-m 'not smoke' ..."` in `pyproject.toml`. Your `-o addopts= -m smoke` defeats both halves — and the second half is not incidental: without it, `--cov-fail-under=77` would fail the job on coverage rather than on the tests.

Neither `test_sft_smoke` nor `test_dpo_smoke` carries an `importorskip` or a `skipif`, so there is no silent-skip escape hatch; only the MLX one skips. The CI log for this PR shows `2 passed, 1 skipped` with `peft` LoRA and `pin_memory` warnings attributed by name to both tests, i.e. a real model load and real training steps rather than a collection that happened to succeed. The step is a plain `run:` under `bash -e`, so a nonzero exit — including exit 5, "collected nothing" — fails the job. That last part is what makes it a guard rather than a decoration.

2m01s total, of which 13s is the tests, running in parallel with Windows cells that take 32 minutes, so it adds no wall-clock. Zero deleted or weakened tests: `git diff main...0cf0351d -- tests/` is empty, and you did not touch the smoke tests to make them pass. 14/14 checks green including the new job.

## The one line

The job runs `-m smoke -q tests/test_smoke_train.py`. The path argument is doing real damage:

```
tests/test_issue300_online_dpo_train.py:69:@pytest.mark.smoke
tests/test_smoke_train.py:15:pytestmark = pytest.mark.smoke
```

`TestOnlineDpoTrainLogsReward` carries two smoke-marked tests — `test_train_logs_reward_metric` and `test_reward_funcs_path_on_trl_1x` — and they run in **none** of the three places: deselected from the default suite by `-m 'not smoke'`, out of scope for `mlx-smoke` (`-k mlx_sft_smoke tests/test_smoke_train.py`), and outside the path your job names. So #596's exact sentence — *nothing runs them, because nothing looked* — survives this PR for those two, in a job whose entire purpose is to end that.

Dropping the path (`-m smoke -q tests/`) fixes it, and it also makes the job self-maintaining: the next contributor who marks something `smoke` gets it run without knowing this job exists. That is the difference between a job that covers two files and one that covers a marker.

Two smaller asks while you are in there, both one-liners:

- **Comment the `"torch>=2.6"`.** It is load-bearing, not belt-and-braces — I hit exactly what it prevents when I ran the smoke tests locally on torch 2.5.1: `sshleifer/tiny-gpt2` ships only `pytorch_model.bin`, and transformers refuses `torch.load` below 2.6 over CVE-2025-32434. A future "`[dev]` already pulls torch" cleanup would break the job with a message that names a CVE rather than the cause.
- **`timeout-minutes: 20`**, so a hang costs 20 minutes rather than the 6-hour default.

## What I have to do, not you

Your title says "required" and a workflow file cannot make a check required. Current branch protection on `main` is `lint, mlx-smoke, transformers-floor, test (9 cells)` — `pytorch-smoke` is not in it, and neither is `type-check`. Until I add it in the repo settings, #596's condition that the job "must be required to be green, or it recreates the current situation with extra steps" is unmet. **That is mine and I will do it the moment this merges** — flagging it so the gap is not silently inherited from the title.

No changelog fragment needed; `CONTRIBUTING.md` scopes those to user-visible changes, and CI-only commits have precedent without one (`f862760`).

## One thing worth knowing, not your problem

The smoke path is only green at **torch >= 2.6**, while `pyproject.toml` declares `torch>=2.3.0` and the `transformers-floor` job pins 2.5.1. So the training chain this job guards is not exercised at the declared floor, and a contributor on torch < 2.6 meets that CVE error locally with no explanation. Pre-existing and not introduced here — I will file it separately rather than hang it on you.

Send the marker widening and I will merge.

MakazhanAlpamys:
One number to go with the review above, since I counted it rather than reasoning about it:

```
pytest -o addopts= -m smoke --collect-only tests/                        -> 5 tests collected
pytest -o addopts= -m smoke --collect-only tests/test_smoke_train.py     -> 3 tests collected
```

The two the job cannot see are `TestOnlineDpoTrainLogsReward::test_train_logs_reward_metric` and `::test_reward_funcs_path_on_trl_1x`. Dropping the path argument closes the gap and makes the job cover the marker rather than two files.

Samearth17:
Thanks for pointing this out. I’ve updated the `pytorch-smoke` job to run `pytest` against `tests/` instead of only `tests/test_smoke_train.py`, so it now covers all tests marked `smoke`.

I have committed and pushed the change. The new CI run should now verify the full smoke-test collection.

MakazhanAlpamys:
Merged as `7556789`, credited in CONTRIBUTORS at `7e656c6`.

**And I have done my half: `pytorch-smoke` is now a required status check on `main`.** Confirmed by the API — the branch protection contexts went from 12 to 13. Your PR title said "required" and a workflow file cannot make it so; that was mine, and #596's condition is now actually met rather than described.

The revision that mattered was the second one. Dropping the path argument turns this from a job covering two files into a job covering the **marker**, which means the next person who writes `@pytest.mark.smoke` gets it run without knowing this job exists. That is the difference between fixing an instance and fixing the class.

What I verified from the merged CI log rather than from the diff:

```
3 passed, 2 skipped, 19923 deselected in 39.95s
```

Two notes on that line, because I would rather state them than let them read better than they are. One skip is MLX; the other is `test_reward_funcs_path_on_trl_1x`, which gates on `trl >= 1` while the repo pins `trl < 0.29`, so it will skip until that bound moves. **Net executing coverage is +1, not +2** — a skipped test is the same colour as a passing one, which is precisely the failure #596 is about, so it should be said out loud.

The failure path is real too: the step is a plain `run:` under `bash -e`, so a nonzero pytest exit fails the job — including exit 5, "collected nothing", so a future rename cannot silently green it.

**Two things I asked for that did not land, and I am doing them myself rather than sending you back for a third round:**

1. A comment on `"torch>=2.6"`. It is load-bearing, not belt-and-braces — `sshleifer/tiny-gpt2` ships only `pytorch_model.bin` and transformers refuses `torch.load` below 2.6 over CVE-2025-32434. Without a comment, a future "`[dev]` already pulls torch" cleanup breaks the job with a message that names a CVE instead of the cause.
2. `timeout-minutes: 20`, so a hang costs twenty minutes rather than the six-hour default.

Both are one line and neither changes behaviour; I will push them as CI hygiene and you will see the commit. Saying so here because taking work silently off a contributor's PR after merging is not something I want to do without a record.

Separately, and not yours to fix: the smoke path is only green at **torch >= 2.6** while `pyproject.toml` declares `torch>=2.5.0`, so the chain this job guards is not exercised at the declared floor and a contributor on 2.5 meets that CVE error locally with no explanation. I will file it.

Refs #596 — leaving the issue open until I have confirmed a deliberate break actually turns the job red, which is its second acceptance criterion and the one thing neither of us has demonstrated in CI.

PR #572, opened 2026-08-26
feat(models): scaffold Qwen4-Exp text LoRA support

Amix29:
Closes #571.

## Summary

- raise the shared `[train]` / `[mlx]` Transformers floor to 5.16.1, the first validated stable Qwen4-Exp release with the 5.16.1 tensor-parallel compatibility fix
- resolve `target_modules: auto` to PEFT `all-linear` for `qwen4_exp_text`, while preserving explicit targets
- complete the actual QSA, Gated DeltaNet, shared-expert, PLE, and gated-residual linear families from the nested text config
- recognize `qwen4_exp_text` as MoE
- add a real tiny-config gate proving outer `qwen4_exp` -> text-only `Qwen4ExpForCausalLM`, finite loss/backward, and LoRA injection across the new linear paths
- explicitly keep Qwen4-Exp outside layer streaming until architecture parity is demonstrated

## Validation

- `ruff check src/soup_cli/ tests/`
- focused mypy: 3 changed source files clean
- targeted compatibility suite: 156 passed
- full suite under Transformers 5.16.1: 18,965 passed, 82 skipped, 5 deselected; coverage 82.81%
- dependency/import check with Transformers 5.16.1, TRL 0.29.1, PEFT 0.20.0, MLX 0.31.2, and MLX-LM 0.31.3

## Deliberate limits

This is a weight-free compatibility scaffold, not a full checkpoint certification. It does not add a catalog recipe or claim a real-weight Apple-Silicon memory fit, multimodal fine-tuning, layer streaming, or routed-expert adaptation. Qwen4-Exp routed experts are raw 3-D parameters, so covering them needs separate PEFT `target_parameters` wiring and independent validation.

Official references:
- https://huggingface.co/Qwen/Qwen3.8-Flash-Next
- https://github.com/huggingface/transformers/pull/48337
- https://github.com/huggingface/transformers/releases/tag/v5.16.1

PR #603, opened 2026-08-30
feat(streaming): add Qwen4/oQ PLE SSD row streaming

### LINKED ISSUE #602: Support Qwen4-Exp layer streaming with SSD-backed PLE N-gram embeddings
## Problem

PR #572 deliberately keeps `qwen4_exp` / `qwen4_exp_text` out of layer
streaming until architecture parity is demonstrated. That guard is correct,
but it leaves Qwen3.8-Flash-Next without a practical low-memory training path.

Qwen4-Exp places its PLE N-gram embedding inside decoder layer 2. The public
Qwen3.8-Flash-Next model card reports 51B N-gram parameters, and Transformers
5.16.1 models the table as one very large `nn.Embedding`. Naively adding Qwen4
to Soup's streaming allowlist would therefore put the PLE table in one decoder
layer shard and require a layer-sized device buffer for the entire table. That
would defeat the bounded-memory property of layer streaming.

The PLE base table is frozen during LoRA and is accessed through sparse row
lookups. It is therefore a good candidate for read-only SSD-backed access: map
the safetensors data, read only the selected rows, and keep the table out of
both the resident base and the decoder-layer buffer pool.

Evidence from an Apple M4 Max with 128 GiB unified memory:

- the public `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` checkpoint is about 99 GiB;
- its affine PLE tensors occupy about 29.80 GiB across 128 shards;
- oMLX 0.6.4 already demonstrates the runtime shape of this optimization with
  a read-only `mmap` PLE row reader and a `qwen4_ple_ssd_offload` setting.

This is evidence that sparse SSD residency is viable, not evidence that Soup
can train the oMLX-specific oQ checkpoint directly.

## Proposed solution

Add a bounded Qwen4-Exp PLE path to Transformers layer streaming, exposed as:

```yaml
training:
  stream_layers: true
  stream_source: auto
  stream_ngram_source: disk  # auto / ram / disk
```

Proposed contract:

- initially accept the setting only for text `qwen4_exp_text` layer-streamed
  LoRA runs;
- locate the PLE N-gram tensor through the safetensors index and validate its
  rank, dtype, offsets, bounds, and source containment before mapping it;
- use a read-only mapping and gather only rows referenced by the current
  N-gram ids;
- exclude the PLE weight from the ordinary decoder-layer shard and buffer
  budget;
- keep the non-PLE PLE projection/normalization/conv weights in the ordinary
  streamed layer path;
- make `auto` choose between RAM and disk from measured table size and
  available RAM, while explicit `ram` / `disk` remain authoritative;
- close every mapping on normal completion and failure;
- fail closed on unsupported checkpoint layouts instead of materializing the
  table or silently falling back;
- admit `qwen4_exp_text` to the streaming allowlist only after resident versus
  streamed forward/backward parity is proven.

The source mapping must remain read-only. Training outputs, adapters, caches,
and checkpoints are still allowed to write normally; this setting is not a
claim that the whole training run is write-free.

## Acceptance evidence

- focused safetensors tests for header validation, row bounds, dtype handling,
  path containment, duplicate indices, cleanup, and no source mutation;
- exact row-gather parity against a resident embedding on synthetic shards;
- a tiny real `Qwen4ExpForCausalLM` resident-versus-SSD test proving finite and
  matching loss plus LoRA gradients through the PLE path;
- a mutation that restores whole-table layer buffering must fail the named
  regression;
- a real public-checkpoint smoke on the M4 Max once a Transformers-compatible
  checkpoint is available locally, with memory and throughput reported as
  observations rather than general capacity claims;
- full Soup test suite and `ruff check src/soup_cli/ tests/` clean.

## Alternatives considered

1. **Only add Qwen4 to `SUPPORTED_STREAM_ARCHS`.** Unsafe: the N-gram table
   becomes part of one enormous decoder layer and breaks the buffer invariant.
2. **Keep the whole table in CPU RAM.** This is the current Transformers-style
   fallback, but it consumes roughly the memory layer streaming is intended to
   save and is not viable on many target machines.
3. **Use generic `stream_source: disk` only.** Insufficient: Soup's disk tier
   still reads a whole layer into a device buffer. PLE needs sparse row-level
   access independent of the ordinary layer source tier.
4. **Target the installed oMLX oQ format first.** Rejected for the initial
   scope. Soup should establish the official Transformers/safetensors contract
   before considering an oMLX-specific affine layout.

## Scope boundaries

This issue does not claim:

- multimodal Qwen4 fine-tuning;
- direct training of oMLX oQ checkpoints;
- layer-streamed PEFT `target_parameters` for routed expert tensors (currently
  resident-only and suitable for a separate follow-up);
- MTP training or serving integration;
- a catalog recipe;
- performance parity with RAM or resident execution before measurement.

I am working on this.

Official references:

- https://huggingface.co/Qwen/Qwen3.8-Flash-Next
- https://github.com/huggingface/transformers/blob/v5.16.1/src/transformers/models/qwen4_exp/modeling_qwen4_exp.py
- #571
- #572
- #573
- #575

## Area

Training (SFT/DPO/GRPO)

Amix29:
## Summary

- admit `qwen4_exp` / `qwen4_exp_text` to layer streaming and normalize both
  dense Transformers and oMLX/oQ affine decoder layouts into the runtime cache
- keep the large PLE N-gram table in its original safetensors files and gather
  only selected rows through strict read-only mmap access, including packed oQ
  PLE tables
- add `training.stream_ngram_source: auto | ram | disk`, including RAM planning,
  SSD/NVMe validation, and a no-op warning when a checkpoint has no PLE table
- fail closed for unsupported tasks, Soup-side base quantization, non-SSD PLE
  streaming, and oQ PLE RAM materialization
- document the supported scope, the CUDA BF16 evidence gap, and the partial M4
  production-checkpoint gate

Closes #602

## Supported scope

- text-only `qwen4_exp` / `qwen4_exp_text` CausalLM checkpoints
- dense Transformers safetensors or supported oMLX/oQ affine safetensors
- SFT only
- `quantization: none` (Soup-side NF4 remains refused)
- dense PLE source selected automatically or forced to RAM/read-only disk
- oQ PLE source kept read-only on SSD/NVMe (`auto` selects disk; RAM is refused)

Preference losses, Soup-side NF4, vision, and MTP training remain out of scope.
The text decoder of the locally installed `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`
checkpoint was used for the production-scale mapping gate; vision and MTP
components are intentionally excluded.

## Implementation notes

- `apply_pre_lora_patches(model, model_id)` is now called from the shared
  `build_streamed_model` path. Existing patches are no-ops for unrelated model
  families, but this intentionally changes the path used by every streamed arch.
- The early RAM preflight for Qwen4 is deferred until the PLE header is known.
  A RAM-constrained run can therefore be rejected after the reusable decoder
  cache has been prepared.
- External PLE descriptors reapply a production-sized element cap, and the
  safetensors byte-range/EOF defence-in-depth checks now have direct tests.

## Validation

- streaming family: `554 passed, 28 skipped`
- focused MPS parity on the M4 Max: `2 passed, 21 deselected`
- full local suite in the sandbox: `18978 passed, 270 skipped, 18 failed`; the
  18 environment-restricted failures were rerun outside the sandbox, yielding
  `17 passed, 1 failed`
- the remaining baseline failure is
  `tests/test_doctor.py::test_doctor_nccl_mocked_success`: mocked CUDA reaches
  `_cuda_getCurrentRawStream`, which is absent from this Apple CPU/MPS build
- exact tiny CPU forward/logit/loss/LoRA-gradient parity in RAM and disk modes
- oQ decoder vectors match MLX for 4, 5, 6, and 8 bits
- real checkpoint mapping: 1,167 expected tensors, 1,167 cached tensors, zero
  missing, unexpected, or shape-mismatched entries; the 251.47 GB decoder cache
  completed while the 102.40 GB logical PLE table stayed packed and read-only

The full production checkpoint reached model construction, LoRA injection,
dataset preparation, and `Training started!`, but its first one-step MPS SFT did
not complete and produced no adapter. The workload can destabilize and shut down
the host, so it was stopped at the owner's request and will not be retried on
this Mac. This PR makes no production-trainability, throughput, or peak-memory
claim. See `benchmarks/gate-qwen4-ple-m4-max.md` for the exact partial/negative
record.

CUDA BF16 resident-versus-streamed parity remains pending and is stated as an
explicit evidence gap.


MakazhanAlpamys:
@Amix29 — reviewed. **The thing I was most worried about is not a problem here**, and I want to lead with that because it is the failure mode this repo names about itself: a half-supported arch that "streams weights into the wrong module and mis-trains silently".

`test_tiny_qwen4_ple_matches_resident_forward_loss_and_lora_gradients` compares resident against streamed at `rtol=0, atol=0` on logits and loss, and I checked the real CI job log rather than the test file:

```
tests/test_issue602_qwen4_ple_streaming.py::test_tiny_..._lora_gradients[disk-cpu] PASSED
tests/test_issue602_qwen4_ple_streaming.py::test_tiny_..._lora_gradients[ram-cpu]  PASSED
```

The `-cpu` variants execute on all 9 cells; only `-mps` skips. That is a gate that actually runs, which is more than I can say for several I have had to reject. And the mutation that targets exactly that failure — an off-by-one in the PLE row lookup — dies **locally**, without even needing the dependency floor:

```
matrix[flat.numpy()] -> matrix[(flat.numpy() + 1) % shape[0]]
FAILED test_sparse_reader_preserves_supported_dtypes[float32-F32]
FAILED [float16-F16]  FAILED [bfloat16-BF16]
PROBE: exact_match=False maxabsdiff=2.998755e+00
```

The arch allowlist gate (7 failures across 3 files) and the oMLX/oQ rejection also die. The deleted test lines are the #572 hold-out being **inverted, not removed** — the replacement at `test_issue571_qwen4_exp.py:59-61` still asserts `"qwen4_exp_text" not in SUPPORTED_STREAM_ARCHS` and additionally pins the alias mapping, so it is stronger. Path handling in `_validated_source_path` is above the house standard: `commonpath` containment rather than `relative_to` (correct on Windows), explicit basename check, `islink` rejection **and** `O_NOFOLLOW|O_CLOEXEC`, where `layer_shard` and `spectrum_scan` use `islink` alone. Import-lightness verified by AST: no heavy top-level imports in `layer_stream.py` or the new `qwen4_ple.py`.

## Required before merge

**1. There is no changelog fragment.**

```
$ git diff --name-only origin/main...9c52a561 | grep -i changelog
(nothing)
```

This adds a user-facing config field and a supported architecture. Every one of the last twelve merges ships one. Note `test_issue487_changelog_fragments.py` passes at 17 — it validates the placement of fragments that exist and structurally cannot see an absent one, so CI green does not cover this.

**2. Three advertised fail-closed gates have no test at all.** The PR body says it fails closed for "oMLX/oQ affine inference weights, unsupported tasks, and quantized Qwen4 streaming". Only the first is tested. Disabling the other two plus the SSD check:

```python
if False and arch == "qwen4_exp" and cfg.task != "sft":      # stream_setup.py:350
if False and arch == "qwen4_exp" and quant != QUANT_NONE:    # stream_setup.py:355
if False and ngram_disk_kind not in ("nvme", "ssd"):
```

gives `6 failed, 486 passed` across the whole streaming family — **byte-identical to baseline**, where those 6 are the transformers-floor errors plus one failure I confirmed pre-existing on the unmutated tree. A grep over `tests/` for those refusal paths returns nothing. These are the guards standing between a user and an unvalidated code path, so they need to fail by name.

**3. Guard the new file's transformers import.** Your own two sibling files do this (`test_issue571_qwen4_exp.py:67-70` uses `try/except ImportError: pytest.skip(...)`, and `test_qwen35_streaming.py` uses the shared `_REQUIRES_TRAINING_FLOOR`). The new file uses neither, so below the floor it produces 5 hard errors where the siblings skip cleanly.

**4. Either publish a `benchmarks/` record for the M4 Max number, or remove it from the docs.** `docs/performance-and-quantization.md:275` ships "a measured maximum logit difference of 1.49e-8", and:

```
$ git diff --name-only origin/main...9c52a561 | grep -c benchmarks
0
```

Both MPS tests skip on all 9 cells, so nothing in CI reproduces it either. Every prior streaming gate published its record — `gate-v0.72.2-nf4.md`, `gate-v0.72.3-breadth.md`, `gate-v0.73.1-measured-vram-fit.md`. A seventeen-significant-digit figure in a docs tree that a DOI points at needs a record behind it or should not be there. I am not doubting the measurement; I am saying it must be reproducible by someone else.

**5. State the bf16 gap.** Parity is float32 CPU only, but `resolve_stream_dtype()` returns bf16 on CUDA, so the first real CUDA qwen4 run trains in a dtype no parity test has touched. Every arch in `gate-v0.72.3-breadth.md` has both a `none` and an `nf4` row measured in bf16. NF4 is explicitly refused here, which is fine — bf16 is not, and the docs' "real-checkpoint and NF4 validation are still pending" does not mention it. Either extend the gate or say so in the arch list.

## Recommended

- **`stream_ngram_source: disk` on an arch with no PLE table is accepted, does nothing, and prints nothing** — `if ngram_bytes:` guards the console line too. v0.73.3 is literally named for four config flags that were validated, documented and read by nothing; a warning is the house standard.
- **`apply_pre_lora_patches(model, model_id)` is new in `build_streamed_model`** and runs for *every* streamed arch, not just qwen4 — so the Gemma4 `ClippableLinear` swap now fires on any streamed model whose name matches. Harmless today (both patches no-op elsewhere), but it changes the shared path that carries the v0.72.x bit-exactness guarantees and should be called out in the body.
- **The early RAM pre-flight is now skipped for `qwen4_exp`** (`arch != "qwen4_exp" and ...`), so an under-RAM'd run is refused *after* paying the full shard-cache write instead of before. Worth a note.
- `ExternalTensorSpec` has no element cap, and PLE keys are explicitly routed around `layer_shard`'s `_MAX_TENSOR_ELEMENTS` via `if key in external_source_keys: continue`. Practically bounded by the real file and the header cross-check catches a lie, so low — but it is a cap the PR steps around rather than re-applies.
- The header byte-range and past-EOF checks in `_tensor_rows_from_header` are deletable with zero test reaction. I measured the severity rather than assuming it: `_SafeTensorContainer.__init__` calls `safe_open` first, and safetensors' own validator refuses both a shape-vs-offsets lie and past-EOF offsets, so these are genuine defence-in-depth and not an exploitable hole. Either reach them with a test or note that they are redundant.

Scope, docs and honesty are otherwise good — the body says plainly that no production throughput or peak-memory claim is made, and the docs keep the negative space visible. Items 1-3 are mechanical; 4 and 5 are about the evidence base, which is the part this project is judged on.

Amix29:
Addressed the review in `d65e9f8`:

1. Added `changelog.d/0.73.3/603.added.md`.
2. Added named fail-closed tests for unsupported tasks, Soup-side NF4, and
   non-SSD PLE streaming:
   - `test_qwen4_streaming_refuses_unsupported_task_by_name`
   - `test_qwen4_streaming_refuses_quantized_base_by_name`
   - `test_qwen4_ple_disk_streaming_refuses_non_ssd_by_name`
3. Guarded the new test helper's Transformers import so dependency-floor jobs
   skip cleanly like the sibling Qwen tests.
4. Removed the undocumented exact MPS delta and added
   `benchmarks/gate-qwen4-ple-m4-max.md`. The record deliberately separates the
   completed mapping/setup gates from the incomplete production optimizer step.
5. Documented that CUDA BF16 parity is still pending.

The recommended points are also covered:

- explicit `stream_ngram_source` now warns when the checkpoint has no PLE table;
- the PR body calls out the shared `apply_pre_lora_patches` path and delayed
  Qwen4 RAM preflight;
- external PLE descriptors now have an element cap;
- header range and past-EOF checks have direct coverage.

Additional scope added after testing the checkpoint that is actually installed
on the M4 Max: supported oMLX/oQ affine decoder conversion and read-only packed
PLE row gathering. The real `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` text skeleton
matched the generated cache 1,167/1,167 with no key or shape mismatch, and the
tiny MPS parity pair passed.

Important limitation: the full one-step MPS SFT did **not** finish and produced
no adapter. It was stopped because this workload can destabilize/shut down the
host, and it will not be retried on that Mac. The benchmark record marks the
production optimizer step **NOT VALIDATED** rather than treating setup success
as trainability evidence.

Local validation: streaming family `554 passed, 28 skipped`; Ruff clean; MPS
parity `2 passed, 21 deselected`. The broad suite produced only environment/
platform failures on macOS; rerunning the 18 affected tests outside the sandbox
gave 17 passes and the known mocked-CUDA Apple build failure.

MakazhanAlpamys:
@Amix29 - re-reviewed `d65e9f80`. **Three of my five items are fully resolved, two are partial, and there are two new findings.** Blocking, but everything below is mechanical.

First, scope. This head is not only a response to the review: `git diff --stat 9c52a561 d65e9f80` is **1216 insertions across 11 files**, including a new `utils/oq_affine.py` (172 lines) and +453 in `layer_shard.py` - a whole new capability (oMLX/oQ affine checkpoints, a 176.9B path) arriving in a re-review round. Not a rejection. It does mean the round did not converge, and two of the findings below are in the new code.

## My five items

| # | item | status |
|---|---|---|
| 1 | changelog fragment | partial |
| 2 | named test per fail-closed gate | **partial - this is the blocker** |
| 3 | transformers import guard | resolved |
| 4 | 1.49e-8 removed or backed by a record | resolved |
| 5 | bf16 gap stated | resolved |

**3, 4 and 5 are done, and done well.** The import now sits behind try/except ImportError to pytest.skip, verified empirically with `-rs`: five clean skips, no errors. A repo-wide grep for the 1.49e-8 figure returns zero. The record you added instead is honest to a degree worth naming - "Production smoke result - STOPPED / not validated", "the precise low-level cause is therefore unresolved", "Do not market this record as proof that the full model is trainable on a 128 GiB Mac." The bf16 gap is stated in three places, including the record's own verdict.

**1 - partial.** The fragment ends `gates (#603).` It credits nobody and omits the issue. The `(#NNN by @user in #PR)` form is the convention here - ten fragments in `0.73.3/` use it, including two of your own. It should read `(#602 by @Amix29 in #603)`. The fragment test passes at 17, so CI cannot catch this.

**2 - partial, and this is what blocks.** The three gates are extracted into named helpers with direct unit tests, and each refuses correctly. I verified by no-op'ing both gate function bodies: 3 failed. Good.

Then I removed **only the two call sites**, leaving the functions intact:

```
stream_setup.py:390   _validate_qwen4_streaming_mode(...)  ->  pass
stream_setup.py:588   _validate_qwen4_ngram_disk(...)      ->  pass

pytest test_issue602_qwen4_ple_streaming.py test_qwen35_streaming.py test_issue571_qwen4_exp.py
  40 passed, 10 skipped        <- byte-identical to baseline
```

**The tests prove the functions refuse. Nothing proves that setup() calls them.** The original fail-open is one refactor away from returning, and it would return silently. This repo already has the pattern for exactly this - `test_every_train_option_is_passed_through_or_deliberately_local`, from #372 - and it exists because the same shape bit us before.

There is also a **fourth** advertised fail-closed gate, added in this commit, left inline rather than extracted, and entirely untested:

```
stream_setup.py:562   if oq_ngram and requested_ngram == "ram":
                          raise ValueError("oQ PLE embeddings require ...")

grep -rc "oQ PLE embeddings require" tests/   ->   0
```

That is the exact defect item 2 was raised about, reintroduced by the round that fixed it. The `auto` tier decision is uncovered for the same reason: `auto` is the shipped default, and the parity test only ever passes an explicit disk or ram.

## New: HIGH - the RAM refusal is dead on the default config

```python
# stream_setup.py:606-607
required_ram = store_total + (ngram_bytes if ngram_source == "ram" else 0)
if tcfg.stream_source == "ram" and required_ram >= free_ram * RAM_TIER_HEADROOM:
```

`required_ram` correctly includes the PLE table - and then the whole check is gated on **tcfg.stream_source**, not on where the PLE actually landed. With the shipped default `stream_source: auto` plus `stream_ngram_source: ram`, a dense PLE table - 102.40 GB on your own record's checkpoint - is loaded by `ShardedRamRowReader.__init__` with nothing refusing it. `plan_free_ram` only shrinks the base-tier budget; it never gates the PLE allocation.

This is precisely the policy the feature is built on: refuses rather than warns, because Linux OOMs and Windows WDDM spills silently. The pre-flight already holds both `ngram_bytes` and `free_ram`. The condition wants `ngram_source == "ram"`.

## New: MEDIUM, five briefly

- **saw_oq_companion is not arch-gated.** `layer_shard.py:1042` sets it from a bare endswith on `.biases` / `.scales`. Every sibling Qwen4 branch is guarded by `if external_mode`; this one is not, so it runs `load_affine_quant_config` for any checkpoint whose tensors end that way - AWQ-style exports, oMLX exports. It fails loudly rather than corrupting, but with a misleading message, on archs this PR should not touch.
- **config.json is outside the shard-cache fingerprint.** `source_fingerprint` is built only from `_discover_safetensors()`. For oQ the numerics of every cached layer depend on the `quantization_config` block in `config.json` - bits, group_size - read separately and never fingerprinted. Correct a group_size while the safetensors are untouched, and `inspect_shard_cache` reports a hit and silently dequantizes at the old bit width. That is exactly the staleness class `source_fingerprint` exists to prevent. The external PLE path is protected, because its specs are re-validated live against the file header; only the oQ dequantized cache is exposed.
- **The gate record is not indexed.** `benchmarks/README.md` has no `gate-qwen4-ple-m4-max.md` row. This is the same item I held #610 for, and release-checklist step 13a says the file alone is not discoverable.
- **gather() always returns CPU** while `qwen4_ple.py:565` registers a weight buffer so that Trainer moves it. forward returns `self.reader.gather(input_ids)`, built with no device argument, never consulting `self.weight.device`. The contract rests on an unverified assumption about upstream modeling code - and your record's one full-checkpoint smoke "repeatedly stopped during the first streamed forward without a Python traceback".
- **The pre-flight prints the dense PLE size for oQ, about 4x over.** `external_tensor_bytes()` sums `spec.nbytes`, which is dense - 102.40 GB - while what is actually mapped is the packed `storage_nbytes`. Display-only today because oQ forces disk, but the same value feeds `required_ram`. `storage_nbytes` already exists and has zero call sites.

**The effective torch floor is 2.3** - from_numpy on np.uint16, and torch.uint32 - while `pyproject.toml:54` declares >=2.0.0. It works on 2.5.1; the declared floor permits a break.

## What I checked and found clean

Path containment uses realpath plus commonpath throughout, never resolve plus relative_to. Symlinks refused, plus O_NOFOLLOW and O_CLOEXEC. Header validation bounds header length, truncation, non-object JSON, a dtype allowlist, positive 2-D shape, data_offsets ordering, the element-count identity and past-EOF offsets, then cross-checks against safetensors' own parsed shape - defence in depth in the spirit of `validate_quant_shape`. Element caps on both spec types, with row ids bounds-checked before indexing. ACCESS_READ with a source SHA-256 non-mutation assertion in the test. Handle lifetime is correct: dedup per file, an owns-container guard, close on partial construction, `runtime.close()` draining external sources, and the parity test asserting the reader is closed. The PLE is not leaked into extras.safetensors. `_SHARD_FORMAT_VERSION` bumped 3 to 5 with external_mode threaded into `inspect_shard_cache`. No test deleted - the only removed lines invert an assertion that is now correct, and still pin the alias table. No top-level heavy imports in any of the seven changed source files, checked by AST. `test_cli_startup_is_light` 34 passed, ruff clean, no Cyrillic. The CPU parity branch is still rtol=0, atol=0 on both logits and loss.

## To unblock

1. Fragment credit to `(#602 by @Amix29 in #603)`.
2. Index the gate record in `benchmarks/README.md`.
3. **Gate the RAM refusal on ngram_source rather than tcfg.stream_source** (HIGH).
4. **Extract and test the fourth gate, and add a seam test proving the call sites are reached from setup().** The seam is the one I care about most: without it, four correct gates and zero of them wired is a green suite.
5. Arch-gate saw_oq_companion, and fold config.json into the fingerprint when it is set.

`layer_shard.py` is now 1542 lines against the project's 800-line guidance, with the natural seam already carved out into `qwen4_ple.py` and `oq_affine.py`. Worth a follow-up, not this PR.

Amix29:
Second review addressed in `688a83c`.

### Merge blockers

1. Changelog credit now uses `(#602 by @Amix29 in #603)`.
2. `gate-qwen4-ple-m4-max.md` is indexed in `benchmarks/README.md`, with the
   stopped production step labelled as not validated.
3. The RAM preflight now refuses whenever either the base policy or the selected
   PLE policy requires RAM. In particular, default `stream_source='auto'` plus
   explicit `stream_ngram_source='ram'` can no longer reach
   `ShardedRamRowReader` over budget.
4. oQ RAM selection is extracted into `_resolve_qwen4_ngram_source`; direct
   tests cover explicit RAM refusal and the shipped `auto -> disk` default.
   `test_setup_reaches_every_qwen4_fail_closed_policy` drives the real
   `_setup_streaming_transformers()` with synthetic seams and proves that mode,
   source, media and RAM-fit policies are all reached from setup.
5. oQ companion detection is now Qwen4/external-mode gated. `config.json` is
   included in the Qwen4 cache fingerprint in both the early planner and the
   sharder; changing only that file invalidates and rebuilds the cache.

### Additional findings

- `DiskBackedEmbedding.forward()` moves gathered rows to the registered
  `weight.device`; the test uses a meta buffer to prove the returned device is
  taken from that contract rather than from the CPU reader.
- oQ PLE sizing now uses `storage_nbytes`, so the panel/preflight reports packed
  mapped bytes rather than the approximately 4x larger dense logical table.
- The effective Torch floor is declared as `torch>=2.3.0` and Soup Doctor now
  checks the same floor.
- A non-Qwen checkpoint containing `.scales` / `.biases` suffixes no longer
  attempts to load oQ affine metadata.

Validation, without loading the production checkpoint:

- Ruff: clean across `src/soup_cli/` and `tests/`.
- Streaming family: `670 passed, 39 skipped`.
- Focused Qwen4 suite: `30 passed, 2 skipped`.
- Tiny MPS parity: `2 passed, 30 deselected`.
- Doctor outside the restricted sandbox: `20 passed`.
- `git diff --check`: clean.

The local mypy invocation still stops inside NumPy's installed stubs before
checking Soup (`Type statement is only supported in Python 3.12 and greater`);
the repository's mypy job is explicitly non-blocking. No production-model
training or full-checkpoint smoke was retried.

MakazhanAlpamys:
Merged as `c97773b`. Thank you — three rounds, and all three of my blocking findings are genuinely closed. I verified every one by mutation rather than by reading the diff.

**The seam, which blocked this twice.** The design is now right, and the right way round: gate *bodies* pinned by behaviour tests, gate *call sites* pinned by `test_setup_reaches_every_qwen4_fail_closed_policy`, which drives the real `_setup_streaming_transformers()` with recorders and asserts the order they are reached in.

| mutation | failing test |
|---|---|
| body: `_validate_qwen4_streaming_mode` -> `return` | `..._refuses_unsupported_task_by_name`, `..._refuses_quantized_base_by_name` |
| body: `_validate_qwen4_ngram_disk` -> `return` | `..._refuses_non_ssd_by_name` |
| body: `_resolve_qwen4_ngram_source` refusal -> `return "ram"` | `test_qwen4_ngram_policy_covers_oq_ram_and_auto_defaults` |
| body: `_validate_qwen4_ngram_ram_fit` -> `return` | `test_qwen4_ram_ple_refusal_is_independent_of_base_stream_source` |
| **call site deleted: mode / source / disk / ramfit, each alone** | `test_setup_reaches_every_qwen4_fail_closed_policy` |
| **two together, and all four together** | same |

No configuration of removed gates passes. Previously, deleting both call sites gave `40 passed` — byte-identical to baseline.

**The HIGH.** Verified by execution at 100 GB free / 102 GB PLE, not by reading: `stream_source: auto` + explicit `ram` is refused; the oQ path refuses even earlier, at `_resolve_qwen4_ngram_source`; `stream_source: ram` with a 102 GB base is refused; and the shipped `auto -> disk` default is allowed. Regressing `stream_setup.py:107` to `ram_required = stream_source == "ram"` fails `test_qwen4_ram_ple_refusal_is_independent_of_base_stream_source` by name.

Secondary claims are pinned too, which I checked because they were the easy ones to leave loose: `storage_nbytes` -> `nbytes` fails `test_external_tensor_bytes_reports_oq_packed_storage`, dropping `.to(self.weight.device)` fails the disk-embedding device test, and the schema footgun is caught by the existing v0.72.0 rejection test.

The 3 deleted test lines are an inversion rather than a removal — the old "not admitted" assertion replaced by one asserting the positive **and** the alias mapping. Correct.

## What I finished myself instead of asking for a fourth round

**The planner half of the cache fingerprint had no test.** `layer_shard.py:1035` is covered by `test_qwen4_config_json_change_invalidates_oq_cache`; `stream_setup.py:484` computes the same fingerprint independently, and setting its `include_config` to a constant `False` passed 694 streaming tests. Not display-only: the planner's `cached` result zeroes `shard_write_bytes`, which feeds the disk pre-flight refusal — under-report the requirement and the run dies out of disk mid-shard, which is the failure that pre-flight exists to prevent.

Fixed in `9c88efc`. The mutation now fails `test_planner_fingerprints_config_json_for_qwen4_but_not_for_llama` (1 failed, 25 passed), reverted 26 passed.

**One thing found while writing it that is worth your attention for future seam tests**: your harness stubs `resolve_model_weights` with a lambda that ignores its `before_materialize` callback — and that callback is where the fingerprint is taken, so `checkpoint_source_components` was never called at all from the harness. The branch was unreachable, which is why nothing could have pinned it without a harness change. The stub is now overridable, and I extracted the ~150-line monkeypatch block into `_drive_qwen4_streaming_setup` shared by both tests: a second, drifting copy of a seam harness is how a seam test quietly stops matching the seam, which is what this PR was blocked on twice. I verified the extraction preserved the original by removing the mode gate and confirming your seam test still fails.

## Two notes, neither a defect

**The torch floor.** This raises `[train]` from 2.0.0 to 2.3.0 for everyone, for a Qwen4-only `uint32` need. I merged it as-is, and the reason is that it excludes nobody: `[train]` already requires `transformers>=5.16.1`, which itself requires `torch>=2.5`. The declared 2.0.0 was already unreachable, so your change makes the declaration less wrong rather than more restrictive. I will file the honest alignment separately.

**`022d30e` weakened one assertion**, and I am recording it rather than reverting it. Dropping `tomllib` is right — it is 3.11+ and the repo supports 3.10 — but the replacement scans `pyproject.toml` for a substring, so `"torch>=2.3.0"` appearing anywhere in the file now satisfies it, including inside a comment. Given the point above I would rather rewrite that test than tighten it, so leave it.

**What I could not verify:** the four numerical parity tests skip on my box (`transformers 4.57.6`, below the declared floor), so the feature's core numerical correctness is unverified here. Worth confirming CI actually installs `transformers>=5.16.1` **and** that it ships `qwen4_exp` — otherwise those parity tests skip in CI too and the claim is never machine-checked anywhere. The gate record honestly labels the stopped production step as not validated, and `benchmarks/README.md` carries the matching "partial pass only" row.

`Refs #602` — the issue stays open on the production-checkpoint run.

MakazhanAlpamys:

Beyond the planner-side fingerprint test I already mentioned, I restored a structured assertion for the torch floor.

Dropping `tomllib` in `022d30e` was right — it is 3.11+ and this repo supports 3.10 — but the replacement was `'"torch>=2.3.0"' in project`, which is satisfied by that string appearing anywhere in the file, including a comment or a different extra. It now parses the `train` extra's own list with a bounded regex.

The mutation is the shape a substring test structurally cannot see — loosen the floor to 2.0.0 *inside* the extra while leaving `"torch>=2.3.0"` in a comment above it:

```
-> FAILED test_qwen4_oq_torch_floor_matches_project_and_doctor
reverted -> 1 passed
```

Issue #599, opened 2026-08-30
Telemetry SSRF guard: DNS-resolving hostnames bypass _is_private_or_link_local


harshitthek:
## Context

Follow-up from [PR #598 review](https://github.com/MakazhanAlpamys/Soup/pull/598) (fix for #593). Identified by maintainer during merge review.

## Problem

`_is_private_or_link_local` in `hubs.py` only parses the host as a **literal IP address** (`ipaddress.ip_address()`). Any hostname that is not a literal IP and not in `_LOOPBACK_HOSTS` passes Layer 2, regardless of what it resolves to:

```python
from soup_cli.utils.trackers import _telemetry_endpoint_is_safe
print(_telemetry_endpoint_is_safe('https://attacker-controlled.example/capture/'))
# True  <-- passes because hostname is not a literal IP
```

An attacker-controlled hostname that DNS-resolves to `10.0.0.1` or `169.254.169.254` would pass the telemetry endpoint guard.

## Scope

`_is_private_or_link_local`'s docstring says DNS resolution is intentionally skipped, matching the existing `hf.py` policy. The resolution-free design was agreed in the #593 thread and is documented in the PR #598 review.

## Suggested approach

Add an optional DNS resolution layer to `_telemetry_endpoint_is_safe` (or a new helper) that resolves the hostname via `socket.getaddrinfo()` and checks the resolved IP(s) against `_is_private_or_link_local`. This should be telemetry-specific (not changing `hubs.py` behavior).

Considerations:
- DNS resolution adds latency and a network dependency to the validation path
- The 1-second hard timeout on telemetry send may already mitigate the practical risk
- Resolution should be cached or performed lazily

here-2007:
I am taking this issue.

here-2007:
@harshitthek I investigated the DNS-resolution gap and reproduced the bypass: hostnames like `10.0.0.1.nip.io` or `localtest.me` pass `_is_private_or_link_local()` because `ipaddress.ip_address()` raises `ValueError` on non-literal IPs and returns `False`, allowing the outbound request to reach private IP space.

Before implementing a standalone `socket.getaddrinfo()` call, there are a few practical trade-offs with the unconstrained DNS resolution approach worth considering:

---

### Potential Pitfalls with Naive `socket.getaddrinfo()` in Validation

1. **Air-Gapped & Offline Latency:**
[...]

Issue #577, opened 2026-08-27
eval-gate: LLM judge task hardcodes / 10.0 normalization instead of active 1–5 rubric scale


here-2007:
```markdown

In `src/soup_cli/eval/gate.py`, `_run_judge_task()` normalizes LLM judge scores by dividing by `10.0`:

```python
results = evaluator.evaluate_batch(items)
# results.overall_score is on a 1-10 scale; normalise to [0, 1].
overall = float(getattr(results, "overall_score", 0.0))
return max(0.0, min(1.0, overall / 10.0))
```

However, the repository's `DEFAULT_RUBRIC` in `src/soup_cli/eval/judge.py` defines a **1–5** scale:

```python
DEFAULT_RUBRIC = {
    "criteria": [...],
    "scale": {"min": 1, "max": 5},
}
```

Because of this mismatch:
- A perfect judge score of `5.0 / 5.0` normalizes to `0.50` instead of `1.00`.
- A midpoint judge score of `3.0` normalizes to `0.30` instead of `0.50`.
- Minimum score `1.0` normalizes to `0.10` instead of `0.00`.

**Impact:** Standard gating thresholds (e.g. `threshold: 0.70` or `0.80`) are mathematically impossible to satisfy under the default rubric. As a result, valid training runs and checkpoint evaluations fail the gate unconditionally, triggering false-positive aborts when `on_regression: stop` is configured.

Note that `src/soup_cli/commands/ship.py` (lines 526–553) already implements the correct dynamic min-max normalization against `evaluator.rubric["scale"]`. `eval/gate.py` needs to be aligned with `commands/ship.py`.

---

### Steps to reproduce

1. Define an evaluation gate suite `evals/gate.yaml` with a `type: judge` task and a passing threshold of `0.70`:
```yaml
suite: quality-gate
tasks:
  - type: judge
    name: response_quality
    threshold: 0.70
    prompts: evals/prompts.jsonl
    judge_model: ollama://llama3.1
```
2. Run `soup eval gate --suite evals/gate.yaml` on a model that receives top scores from the judge (e.g. `5.0 / 5.0`).
3. Inspect the gate verdict and normalized task score.

**Expected behavior:**
A judge score of `5.0` on a 1–5 scale normalizes to `1.00`, passing the `0.70` threshold (`passed: true`).

**Actual behavior:**
The score is divided by `10.0`, producing `0.50`. The task fails the `0.70` threshold (`passed: false`), and the overall gate fails.

---

[...]

PR #604, opened 2026-08-30
fix(security): reject abbreviated/decimal/hex/octal IPv4 in SSRF guards

### LINKED ISSUE #600: Telemetry SSRF guard: abbreviated/decimal IPv4 forms bypass ipaddress.ip_address()
## Context

Follow-up from [PR #598 review](https://github.com/MakazhanAlpamys/Soup/pull/598) (fix for #593). Identified by maintainer during merge review.

## Problem

`ipaddress.ip_address()` (used by `_is_private_or_link_local`) raises `ValueError` on abbreviated and decimal IPv4 forms, causing them to bypass the private-IP check entirely:

- `https://127.1/` — abbreviated loopback (glibc `getaddrinfo` may accept this as `127.0.0.1`)
- `https://2130706433/` — decimal representation of `127.0.0.1` (`0x7F000001`)
- `https://0x7f000001/` — hex representation of `127.0.0.1`

```python
import ipaddress
try:
    ipaddress.ip_address("127.1")
except ValueError as e:
    print(e)  # "'127.1' does not appear to be an IPv4 or IPv6 address"
```

This is the same bypass mechanism as the `localhost` hole caught in #593 (hostname not parseable as a literal IP falls through to the `_LOOPBACK_HOSTS` set, which doesn't contain these forms), one layer further out.

## Scope

Whether `glibc`'s `getaddrinfo` accepts these forms in production was not confirmed (tested from a Windows box). These forms are outside the predicate but may not be live attack vectors in all environments.

## Suggested approach

Normalize the hostname before passing to `ipaddress.ip_address()`, or add these forms to the rejection set. Consider using `socket.inet_aton()` which accepts abbreviated forms, then converting the result to `ipaddress.ip_address()` for the private/loopback check.

here-2007:
## Summary

Closes #600.

`_is_private_or_link_local()` in `hubs.py`, `hf.py`, and `webhooks.py` delegated exclusively to `ipaddress.ip_address()`, which raises `ValueError` on non-canonical IPv4 forms (`127.1`, `2130706433`, `0x7f000001`, `0177.0.0.1`) and trailing-dot FQDN variants (`169.254.169.254.`).

On Linux (measured on Ubuntu, glibc 2.39), `getaddrinfo` and `inet_aton` resolve these forms to their canonical private/loopback IPs, allowing:
1. Overridden telemetry endpoints (`SOUP_POSTHOG_ENDPOINT="https://2852039166/"`) to bypass `_telemetry_endpoint_is_safe`.
2. Webhook URLs (`--slack-url https://2852039166/` or `--slack-url https://169.254.169.254./`) to reach cloud metadata (`169.254.169.254`) and internal LAN addresses.

*(On Windows, `WSAStringToAddress` / `getaddrinfo` returns `gaierror` for abbreviated forms, making this a live vector specifically in Linux container/server training environments).*

---

### Key Bypass Vectors (All Verified & Blocked)

| Form | Canonical IP | Category | Linux glibc Resolver | Windows Resolver |
|---|---|---|---|---|
| `127.1` | `127.0.0.1` | Abbreviated loopback | `127.0.0.1` | `gaierror` |
| `127.0.1` | `127.0.0.1` | Abbreviated loopback (3-octet) | `127.0.0.1` | `gaierror` |
| `2130706433` | `127.0.0.1` | 32-bit decimal | `127.0.0.1` | `gaierror` |
| `0x7f000001` | `127.0.0.1` | 32-bit hexadecimal | `127.0.0.1` | `gaierror` |
| `0177.0.0.1` | `127.0.0.1` | Octal dotted | `127.0.0.1` | `gaierror` |
| `10.1` | `10.0.0.1` | Abbreviated RFC1918 | `10.0.0.1` | `gaierror` |
| `167772161` | `10.0.0.1` | 32-bit decimal | `10.0.0.1` | `gaierror` |
| `2852039166` | `169.254.169.254` | Cloud metadata (decimal) | `169.254.169.254` | `gaierror` |
| `169.254.169.254.` | `169.254.169.254` | Trailing-dot FQDN | `169.254.169.254` | `169.254.169.254` |

---

### Fix Details

1. **`socket.inet_aton` Fallback:** When `ipaddress.ip_address()` raises `ValueError`, fall back to `socket.inet_aton(clean_host)`. Because `inet_aton` uses the platform C parser, it mirrors the exact resolver semantics in-process with **zero DNS lookups, zero network calls, and zero latency overhead**.
2. **Trailing-Dot Normalisation:** `clean_host = host.rstrip(".")` ensures trailing-dot FQDN variants (e.g. `169.254.169.254.`) are stripped before parsing.
3. **Scope & Posture:**
   - `src/soup_cli/utils/webhooks.py`: Active security protection against SSRF reaching internal hosts / metadata on both HTTP and HTTPS.
   - `src/soup_cli/utils/trackers.py`: Re-evaluates via `_telemetry_endpoint_is_safe` using `_is_private_or_link_local`.
   - `src/soup_cli/utils/hubs.py` & `src/soup_cli/utils/hf.py`: Error message parity in plain-HTTP rejection selection.

---

### Testing & Verification

- **82 parametrised tests** in `tests/test_issue600_abbreviated_ipv4_ssrf.py`:
  - Unit tests for all three `_is_private_or_link_local` implementations.
  - End-to-end pinning at the caller level for `trackers._telemetry_endpoint_is_safe`, `trackers._resolve_posthog_target` (args & env vars), and `webhooks.validate_webhook_url`.
- Full SSRF / telemetry test sweep: **528 passed** (`-k "ssrf or telemetry or hub or tracker or posthog or webhook"`).
- Rebased cleanly onto `upstream/main` (`2ee8e13`).
- Zero direct `CHANGELOG.md` edits; fragment placed at `changelog.d/0.73.3/604.security.md` (validated with `test_issue487_changelog_fragments.py`).
- `ruff check` clean.

### DIFF
--- a/src/soup_cli/utils/hf.py
+++ b/src/soup_cli/utils/hf.py
@@ -104,7 +104,8 @@ def resolve_endpoint() -> str:
         raise ValueError(
             "HF_ENDPOINT 0.0.0.0 is ambiguous; use 127.0.0.1 or localhost"
         )
-    if parsed.scheme == "http" and host not in _LOOPBACK_HOSTS:
+    host_clean = host.lower().rstrip(".")
+    if parsed.scheme == "http" and host_clean not in _LOOPBACK_HOSTS:
         # Reject plain HTTP for RFC1918 / link-local / cloud metadata too —
         # HF_ENDPOINT=http://169.254.169.254 or http://192.168.1.1 would
         # otherwise route SDK traffic to internal targets.
@@ -121,15 +122,33 @@ def resolve_endpoint() -> str:
 
 
 def _is_private_or_link_local(host: str) -> bool:
-    """Whether ``host`` resolves to a private / link-local / loopback IP."""
+    """Whether ``host`` resolves to a private / link-local / loopback IP.
+
+    Handles canonical IPv4/IPv6 via :func:`ipaddress.ip_address` **and**
+    abbreviated / decimal / hex / octal IPv4 forms (e.g. ``127.1``,
+    ``2130706433``, ``0x7f000001``, ``0177.0.0.1``) via the platform C
+    library :func:`socket.inet_aton`.  The latter is a pure in-process
+    string parser — no DNS lookup is performed.
+    """
+    import socket  # noqa: PLC0415 — lazy import (stdlib, negligible cost)
+
+    clean_host = host.rstrip(".")
     try:
-        addr = ipaddress.ip_address(host)
+        addr = ipaddress.ip_address(clean_host)
+        return addr.is_private or addr.is_link_local or addr.is_loopback
     except ValueError:
-        # Hostname — we don't resolve DNS here (the SDK does), so fall back
-        # to "treat as public". A malicious DNS record pointing to a private
-        # IP is out of scope for this local-tool threat model.
+        pass
+    # Fallback: C-level inet_aton accepts abbreviated/integer/hex/octal
+    # IPv4 representations that Python's ipaddress module rejects.
+    try:
+        canonical = socket.inet_ntoa(socket.inet_aton(clean_host))
+        addr = ipaddress.ip_address(canonical)
+        return addr.is_private or addr.is_link_local or addr.is_loopback
+    except (OSError, ValueError):
+        # Hostname — we don't resolve DNS here (the SDK does), so fall
+        # back to "treat as public". A malicious DNS record pointing to
+        # a private IP is out of scope for this local-tool threat model.
         return False
-    return addr.is_private or addr.is_link_local or addr.is_loopback
 
 
 def validate_repo_id(repo_id: str) -> None:
--- a/changelog.d/0.73.3/604.security.md
+++ b/changelog.d/0.73.3/604.security.md
@@ -1,0 +1,10 @@
+- **Telemetry and webhook SSRF guards now reject abbreviated, decimal, hex,
+  and octal IPv4 forms (#600 by @here-2007 in #604).** `_is_private_or_link_local()`
+  in `utils/hubs.py`, `utils/hf.py`, and `utils/webhooks.py` previously delegated
+  exclusively to `ipaddress.ip_address()`, which raises `ValueError` on non-canonical
+  IPv4 representations like `127.1`, `2130706433`, `0x7f000001`, and `0177.0.0.1`.
+  On Linux (glibc), the OS resolver (`getaddrinfo`/`inet_aton`) parses these forms,
+  allowing crafted `--slack-url https://2852039166/` webhooks or telemetry overrides
+  to reach internal network addresses and cloud metadata (`169.254.169.254`). All three
+  guards now normalise non-standard IPv4 literals via in-process `socket.inet_aton()`
+  and strip trailing FQDN dots before verifying against `ipaddress.ip_address()`.
--- a/src/soup_cli/utils/hubs.py
+++ b/src/soup_cli/utils/hubs.py
@@ -113,13 +113,28 @@ def endpoint_env_var(hub: str) -> str:
 def _is_private_or_link_local(host: str) -> bool:
     """Whether ``host`` is a private / link-local / loopback IP.
 
-    DNS resolution intentionally not performed (matches v0.29.0 hf.py policy).
+    Handles canonical IPv4/IPv6 via :func:`ipaddress.ip_address` **and**
+    abbreviated / decimal / hex / octal IPv4 forms (e.g. ``127.1``,
+    ``2130706433``, ``0x7f000001``, ``0177.0.0.1``) via the platform C
+    library :func:`socket.inet_aton`.  The latter is a pure in-process
+    string parser — no DNS lookup is performed.
     """
+    import socket  # noqa: PLC0415 — lazy import (stdlib, negligible cost)
+
+    clean_host = host.rstrip(".")
     try:
-        addr = ipaddress.ip_address(host)
+        addr = ipaddress.ip_address(clean_host)
+        return addr.is_private or addr.is_link_local or addr.is_loopback
     except ValueError:
+        pass
+    # Fallback: C-level inet_aton accepts abbreviated/integer/hex/octal
+    # IPv4 representations that Python's ipaddress module rejects.
+    try:
+        canonical = socket.inet_ntoa(socket.inet_aton(clean_host))
+        addr = ipaddress.ip_address(canonical)
+        return addr.is_private or addr.is_link_local or addr.is_loopback
+    except (OSError, ValueError):
         return False
-    return addr.is_private or addr.is_link_local or addr.is_loopback
 
 
 def validate_hub_endpoint(endpoint: str, *, hub: str | None = None) -> str:
@@ -164,7 +179,8 @@ def validate_hub_endpoint(endpoint: str, *, hub: str | None = None) -> str:
         raise ValueError(
             f"{label} 0.0.0.0 is ambiguous; use 127.0.0.1 or localhost"
         )
-    if parsed.scheme == "http" and host not in _LOOPBACK_HOSTS:
+    host_clean = host.lower().rstrip(".")
+    if parsed.scheme == "http" and host_clean not in _LOOPBACK_HOSTS:
         if _is_private_or_link_local(host):
             raise ValueError(
                 f"{label} plain HTTP is only allowed for loopback "
--- a/src/soup_cli/utils/webhooks.py
+++ b/src/soup_cli/utils/webhooks.py
@@ -31,20 +31,39 @@ _LOOPBACK_HOSTS = frozenset({"localhost", "127.0.0.1", "::1"})
 def _is_private_or_link_local(host: str) -> bool:
     """Return True iff ``host`` resolves to a non-loopback private/reserved IP.
 
+    Handles canonical IPv4/IPv6 via :func:`ipaddress.ip_address` **and**
+    abbreviated / decimal / hex / octal IPv4 forms (e.g. ``127.1``,
+    ``2130706433``, ``0x7f000001``, ``0177.0.0.1``) via the platform C
+    library :func:`socket.inet_aton`.  The latter is a pure in-process
+    string parser — no DNS lookup is performed.
+
     Explicit parentheses on the final clause (mirrors v0.63.0 drift-alarm
     code-review MEDIUM fix): Python binds ``and`` tighter than ``or``, but
     the SSRF gate is safety-critical and a future edit should not need to
     re-derive the precedence rules to verify the logic.
     """
+    import socket  # noqa: PLC0415 — lazy import (stdlib, negligible cost)
+
+    clean_host = host.rstrip(".")
+
+    def _check(ip: ipaddress.IPv4Address | ipaddress.IPv6Address) -> bool:
+        return (
+            ip.is_private
+            or ip.is_link_local
+            or (ip.is_loopback is False and (ip.is_reserved or ip.is_multicast))
+        )
+
     try:
-        ip = ipaddress.ip_address(host)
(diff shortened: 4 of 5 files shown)


MakazhanAlpamys:
Thank you, @here-2007 — the engineering here is right, and you closed a bypass I explicitly could not confirm. Holding on changelog hygiene, not on the fix.

## The fix is correct, and the live-bypass question is now settled

When I filed #600 I said I could not tell from a Windows box whether glibc's `getaddrinfo` actually accepts these forms, so I would not claim a live bypass. It does, and this is now measured on WSL Ubuntu, glibc 2.39:

```
127.1          -> 127.0.0.1
2130706433     -> 127.0.0.1
0x7f000001     -> 127.0.0.1
0177.0.0.1     -> 127.0.0.1
10.1           -> 10.0.0.1
2852039166     -> 169.254.169.254
```

All seven return `gaierror` on Windows. So the bypass is live on Linux, which is where training runs. Please cite that in the PR body — it currently asserts the bypass without evidence, and now there is some.

**Using `socket.inet_aton` rather than a hand-rolled parser is the right call**, and I checked it is not merely convenient: across 26 exotic forms the guard and the resolver agree everywhere, and where they diverge (`4294967295`) they diverge *together*, because both track the platform's own parser. A hand-rolled version would have been correct on one OS and wrong on the other.

**No hub regression, verified in both directions.** This was my main worry. In both `hubs.py` and `hf.py`, `_is_private_or_link_local` is only reachable inside `if parsed.scheme == "http" and host not in _LOOPBACK_HOSTS:`, and both arms of that branch raise — so the helper only selects the error message and never decides accept/reject, and HTTPS never calls it at all. Driving the real functions on main vs the merged tree, all 23 probed URLs are byte-identical, including `https://10.0.0.5/` as a private ModelScope mirror and `https://hub.internal.example.com/`. `soup train --hub`, `soup data push --hub` and `soup data forge --hub` are unaffected.

No over-rejection either: `example.com`, `api-2.example.com`, `1.2.3.4.example.com`, `xn--80ak6aa92e.com`, `8.8.8.8`, public IPv6 and public decimal forms all still accept.

Mutations, all three killed:

```
revert all three guards to pre-PR   -> 21 failed, 24 passed
inject `return True` (reject all)   -> 12 failed, 33 passed   (over-strict is caught too)
break normalisation (inet_aton fixed) -> 21 failed, 24 passed
```

45 tests, all with real asserts, zero mocks, strict `is True` / `is False`. Zero deleted or weakened test lines. `ruff` clean, import-light (`socket` imported inside the functions), 490 passed on the `-k "ssrf or telemetry or hub or tracker or posthog or webhook"` sweep.

You also fixed a **second bypass nobody had reported**: `--slack-url https://2852039166/` reaching cloud metadata through `webhooks.py`. Worth naming in the changelog.

## Blocking — all changelog, all mechanical

**1. Drop the `CHANGELOG.md` hunk entirely.** This repo does not take direct `CHANGELOG.md` edits any more (#487/#490); fragments go in `changelog.d/`. The hunk also adds a **second `### Fixed` heading** inside `[Unreleased]`, so the file now has two.

**2. That hunk would publish #577 twice.** Its content is already on `main` as `changelog.d/0.73.3/578.fixed.md`. Running the real assembler on the merged tree:

```
$ py -3.10 scripts/assemble_changelog.py
Assembled 54 changelog fragment(s).
$ grep -c "Eval-gate LLM judge normalization" CHANGELOG.md
2
```

The validator does not catch it because your direct edit says `#577` while the fragment is numbered `578`. This is the same duplication defect that got into v0.73.2 via #392, so I want it gone rather than deduplicated. It looks like a stale branch base dragged your own earlier work along — the fix is just to drop the file from the diff.

**3. The fragment is filed under someone else's PR number.**

```
changelog.d/0.73.3/601.security.md
```

**#601 is @Samearth17's "ci: add required PyTorch smoke job"** — a different PR by a different author. Per `changelog.d/README.md` the number must be this PR's. Please rename to `604.security.md` and make the body read `(#600 by @here-2007 in #604)`. As it stands it would credit the wrong person, and it would collide the moment #601 adds a fragment of its own.

## Should-fix

**4. Nothing pins the bug at the layer it was reported.** Every test targets the private helper `_is_private_or_link_local` directly. I simulated the obvious future refactor — inlining the guard in `trackers.py` so it stops importing from `hubs.py` — and:

```
test_issue600_*.py + test_telemetry_ssrf.py  -> 73 passed, ALL GREEN
_telemetry_endpoint_is_safe('https://127.1/') = True
_resolve_posthog_target -> ('k', 'https://2852039166/')
```

The suite stays green while #600 is fully reopened. Worse, under mutation (a) — the bug entirely restored — `tests/test_telemetry_ssrf.py` still passes 28/28. Please add a handful of tests driving `trackers._telemetry_endpoint_is_safe` / `_resolve_posthog_target` and `webhooks.validate_webhook_url` end to end with `https://127.1/`, `https://2130706433/`, `https://2852039166/`. That is the difference between pinning the vulnerability and pinning an implementation detail.

## Non-blocking

- **A residual same-class bypass inside your own stated scope.** glibc resolves the trailing-dot form `127.0.0.1.` → `127.0.0.1`, but both `inet_aton` and `ipaddress` reject it. Telemetry is safe because it does `.rstrip(".")`; `webhooks.py` and `hubs.py` are not, so `--slack-url https://169.254.169.254./` still reaches cloud metadata on Linux. One line in `webhooks.py`, or a follow-up issue — your call.
- **The PR description overstates two of three files.** The `hubs.py` and `hf.py` changes are error-message-only, since HTTPS never reaches that helper there. Only `webhooks.py` (and telemetry transitively) gets a security benefit. Worth correcting so the record is accurate.
- **CI green is on the wrong tree.** All 13 checks ran on `5d3076e3`, which predates `2ee8e13` (#598). The merged combination was never CI-tested; I ran it locally and it is clean, but a rebase would make the green mean what it appears to mean.

Fix 1-3 and I will merge; 4 and the trailing-dot item can be follow-ups if you prefer.

here-2007:
Thank you for the thorough review and for measuring the glibc 2.39 resolver behavior on Linux!

All blocking, should-fix, and non-blocking items are resolved in the rebased commit:

1. **Clean Rebase on `upstream/main` (`2ee8e13`):**
   - Dropped the stale branch base and eliminated the direct `CHANGELOG.md` hunk entirely. `git diff upstream/main -- CHANGELOG.md` is now completely empty.
2. **Renamed and Updated Changelog Fragment (`604.security.md`):**
   - Moved to `changelog.d/0.73.3/604.security.md`.
   - Updated header to `(#600 by @here-2007 in #604)` and explicitly documented both telemetry and webhooks (`--slack-url https://2852039166/`).
3. **End-to-End Caller-Level Pinning Tests Added:**
   - Added `TestTelemetrySSRFEndToEndPinning` driving `trackers._telemetry_endpoint_is_safe` and `_resolve_posthog_target` with argument and env-var overrides.
   - Added `TestWebhookSSRFEndToEndPinning` driving `webhooks.validate_webhook_url`.
   - Total tests in `tests/test_issue600_abbreviated_ipv4_ssrf.py`: **82 passed**.
4. **Trailing-Dot FQDN Normalisation:**
   - Handled trailing dots (`clean_host = host.rstrip(".")`) in `_is_private_or_link_local` and loopback check in `validate_webhook_url`, closing the `https://169.254.169.254./` metadata bypass as well.
5. **PR Description Updated:**
   - Cited Linux glibc 2.39 vs Windows evidence and clarified scope across `webhooks.py`, `trackers.py`, `hubs.py`, and `hf.py`.

All 528 SSRF/telemetry/hub/webhook tests and 17 fragment validation tests are green.

src/soup_cli/utils/hf.py
```diff
@@ -121,15 +122,33 @@ def resolve_endpoint() -> str:
...
-        # Hostname — we don't resolve DNS here (the SDK does), so fall back
-        # to "treat as public". A malicious DNS record pointing to a private
-        # IP is out of scope for this local-tool threat model.
+        pass
+    # Fallback: C-level inet_aton accepts abbreviated/integer/hex/octal
+    # IPv4 representations that Python's ipaddress module rejects.
+    try:
+        canonical = socket.inet_ntoa(socket.inet_aton(clean_host))
+        addr = ipaddress.ip_address(canonical)
+        return addr.is_private or addr.is_link_local or addr.is_loopback
```
AmirF194, line 146: I checked this against ffcc7675 in a clean container and compared the three copies of `_is_private_or_link_local` this PR touches.

hf.py:138/146 and hubs.py:127/135 return `addr.is_private or addr.is_link_local or addr.is_loopback`, unchanged from before this PR. webhooks.py's copy (`_check()`, lines 49-54) also ORs in `is_reserved`/`is_multicast`, with a comment noting it already went through a prior drift fix between these three copies. A numeric-encoded multicast host is not private, link-local or loopback, so it still resolves as public in hf.py/hubs.py after this PR's inet_aton fallback, while the same host through webhooks.py is rejected.

```
$ docker run --rm python:3.11-slim python3 -c "import ipaddress; a=ipaddress.ip_address('224.0.0.1'); print(a.is_private, a.is_link_local, a.is_loopback, a.is_multicast)"
False False False True
```

I have not verified that an outbound connect() to a multicast address does anything meaningful for HF_ENDPOINT/hub-endpoint traffic, so I cannot say this is exploitable the way the abbreviated-IPv4 forms are. Flagging it because this PR is already touching all three copies for the same reason webhooks.py's stricter check exists. Mirroring webhooks.py's `_check()` (add `is_reserved`/`is_multicast`) into hf.py and hubs.py would close the gap in the same pass.

MakazhanAlpamys:
Merged as `7b81810`. Thank you, @here-2007 — all three changelog items gone in one push, and the fix is the strongest kind: it answered a question the issue could not.

**You settled what I explicitly declined to claim.** I filed #600 saying I could not tell from a Windows box whether glibc actually accepts these forms, so I would not assert a live bypass. You measured it:

```
glibc 2.39 (WSL Ubuntu)         Windows
  127.1        -> 127.0.0.1       gaierror
  2130706433   -> 127.0.0.1       gaierror
  0x7f000001   -> 127.0.0.1       gaierror
  0177.0.0.1   -> 127.0.0.1       gaierror
  10.1         -> 10.0.0.1        gaierror
  2852039166   -> 169.254.169.254 gaierror
```

Live on Linux, which is where training runs.

**`socket.inet_aton` over a hand-rolled parser is the load-bearing choice**, and the reason is the good one: across 26 exotic forms the guard and the resolver agree everywhere, and where they diverge they diverge *together*, because both track the platform's own parser. A hand-rolled version would have been right on one OS and wrong on the other.

You also fixed a second bypass nobody had reported — `--slack-url https://2852039166/` reaching cloud metadata through `webhooks.py` — and closed the trailing-dot form (`https://169.254.169.254./hook`) I had raised as non-blocking.

Verified at merge:

```
baseline (incl. #598's telemetry suite)              110 passed
revert all three guards to pre-PR                     50 failed
```

The over-strict direction is caught too, so a validator that rejects legitimate hosts would not have merged either. No hub regression — in `hubs.py`/`hf.py` the helper is only reachable inside the `http` branch where both arms raise, so it selects the error message and never decides accept/reject; 23 URLs byte-identical across both trees, including `https://10.0.0.5/` as a private ModelScope mirror.

The three changelog items are all resolved: the direct `CHANGELOG.md` edit is dropped (it would have published #577 twice and left a duplicate `### Fixed` heading), and the fragment is `604.security.md` crediting `(#600 by @here-2007 in #604)` instead of another contributor's PR number.

You are in `CONTRIBUTORS.md` as of `946c3be`. #599 — the DNS-resolution gap — is still yours if you want it; you reproduced it on the issue with `10.0.0.1.nip.io` and `localtest.me`, and that one genuinely needs a design decision about whether to resolve at all, so say how you would approach it before writing code.

Issue #395, opened 2026-08-14
Layer streaming: the VRAM formula UNDER-predicts past ~4700 tokens and the mechanism is unidentified


MakazhanAlpamys:
Sibling of #327, opposite direction. #327 is about the pre-flight over-predicting; this is about it **under-predicting**, which is the direction that does not announce itself — an OOM on Linux, a silent spill to host memory on Windows/WDDM.

## Measured

Through the real `soup train` on an RTX 3050 Laptop (4 GB, Windows, torch 2.5.1), SmolLM2-135M streamed bf16, batch 1, `quantization: none`, LoRA r=8:

| seq | predicted | real peak | pred/real |
|---|---|---|---|
| 4352 | 3.282 GB | 3.036 GB | 1.081x — over-predicts, safe |
| 5120 | 3.844 GB | 4.118 GB | **0.934x** |
| 6144 | 4.590 GB | 5.830 GB | **0.787x — under by 21%** |

Crossover sits between seq 4352 and 5120. Deterministic (repeats at a fixed shape returned bit-identical peaks).

**Why the existing guard could not see it:** all ten rows of `MEASURED_VRAM_GRID` are at seq 256 or 512. The grid varies *batch* (1..8) and says nothing about sequence length, so `test_never_under_predicts` asserted a global-sounding guarantee from evidence that could not support it. It was narrowed in v0.73.1 with an assertion that fires if a longer row is added without re-establishing the claim.

## The mechanism is NOT established

`seq**2` from the attention score matrix is the obvious candidate and the numbers do not cleanly fit it. No alternative was confirmed. This is recorded as unknown rather than guessed at — and it is the reason v0.73.1 shipped a *measurement* (`training.stream_vram_probe`) rather than another coefficient: a formula cannot model a term nobody has identified.

## Why this still matters with the probe shipped

The probe overrules the formula but does not fix it, and it is `task: sft` only. Anything that needs the prediction to be right **without** a GPU still gets the old accuracy: `soup profile`, plan-only paths, and the four streaming preference losses.

## Proposed fix path

1. Bisect the term with `torch.cuda.memory_stats()` / `memory_snapshot` at seq 3072 / 4096 / 5120 on one model, isolating which allocation grows faster than `batch x seq`.
[...]

PR #605, opened 2026-08-30
fix(streaming): correct stream_vram_probe VRAM figures and withdrawn preference-loss justification

### LINKED ISSUE #595: stream_vram_probe docs: 0.830x is the probe's denominator, not the real peak, and the preference-loss justification was withdrawn
## Summary

Two stale claims about the layer-streaming VRAM pre-flight are live in `main`, both
in text a user or contributor reads before deciding whether to trust the gate:

1. Four sites quote the formula-vs-**probe** ratios while labelling them "the real
   peak". The real-peak figures are 0.934x / 0.787x, not 0.992x / 0.830x.
2. `stream_vram_probe`'s field description states the probe "under-measures"
   preference losses "badly enough to make the gate unsafe" — a claim
   `benchmarks/gate-v0.73.1-measured-vram-fit.md` explicitly withdraws, and the
   opposite of what the validator 1,900 lines below it in the same file says.

The docs are already correct (`docs/performance-and-quantization.md:500`,
`CONTRIBUTING.md:291`). The code is the outlier.

## 1. `0.830x the real peak` is the probe's denominator

`src/soup_cli/config/schema.py:3241`:

> The formula under-predicts past seq 4096 (measured **0.830x the real peak** at
> seq 5120)

Same series, same label, at `src/soup_cli/utils/layer_stream.py:1235`
("at seq 4096 the prediction is 0.992x **the real peak** and at seq 5120 it is
0.830x, i.e. it under-predicts by 17%"),
`src/soup_cli/utils/layer_stream_runtime.py:1357`, and
`tests/test_v07203.py:173`.

The record holds two different comparisons:

| | seq 4352 | seq 5120 | seq 6144 |
|---|---|---|---|
| formula vs **real training run** (`gate-v0.73.1:28`) | 1.081x | **0.934x** | 0.787x |
| probe vs real run (`gate-v0.73.1:94`) | +14.3% | +12.5% | — |

`0.830x` is the formula against the *probe*: predicted 3.844 GB / probe 4.633 GB
= 0.8297. Against the real peak the same shape is 3.844 / 4.118 = 0.9335, which is
the 0.934x the gate table, the docs and `CONTRIBUTING.md` all carry.

The crossover corroborates it independently of the arithmetic.
`gate-v0.73.1:100-102`:

> the probe's own crossover (between seq 3072 and 4096) is earlier than the
> training path's (between 4352 and 5120).

The quoted series turns over between 3072 and 4096 — the probe's crossover, not
the training path's. So these four sites appear to be the probe series wearing
the real-peak label, unless there is an unpublished series I have not found.

This also moves a magnitude: `layer_stream.py:1235` concludes "it under-predicts
by 17%", which is the probe-denominated gap. Against the real peak at seq 5120 it
is ~7%.

## 2. The preference-loss justification was withdrawn

`src/soup_cli/config/schema.py:3245-3249`:

> task='sft' only: the probe runs a plain causal-LM forward+backward, which IS the
> SFT step but is not a preference loss, and it **under-measures those badly
> enough to make the gate unsafe**.

`benchmarks/gate-v0.73.1-measured-vram-fit.md:104-110`:

> **For a preference loss the probe is not validated, and one claim made during
> this work was wrong.** An earlier draft of this record said the probe
> under-measures a DPO step by 42% and was therefore unsafe. That compared two
> *different shapes* [...] Measured at the matching shape, the probe reads
> **6.021 GB against the real DPO step's 5.304 GB, i.e. +13.5%** — the same safe
> direction it shows for SFT, not the opposite one.

The validator that actually enforces the restriction agrees with the record
(`schema.py:5157-5166`): "+13.5%, the same safe direction it shows for SFT — one
point is not a validation". `gate-v0.72.4-preference-losses.md:181` points the
same way, with DPO's whole-step peak at 0.914x SFT's.

So the restriction is real but its stated reason is inverted: the probe
over-reads at the one shape measured, and is restricted because one shape is not
a grid.

The field description is the copy a user reads when deciding whether to turn the
gate on, and it currently tells them the instrument is unsafe in a direction the
record says it is not. It also misdirects contributors: I spent a working session
designing an extension of the probe to preference losses on the strength of that
sentence, before finding the validator and the withdrawal. The design was wrong
for other reasons too — the TRL trainer does not exist at the probe's call site
(`dpo.py:102` vs `dpo.py:212`) — but the sentence is what pointed at the problem
that was not there.

## Suggested wording

For `schema.py:3241-3249`, following the validator's reasoning:

> The formula under-predicts past seq 4096 (measured 0.934x the real peak at seq
> 5120 and 0.787x at 6144), which is the direction that does not announce itself
> — an OOM on Linux, a silent spill to host memory on Windows. Off by default: it
> costs one step (1-5 s measured) and it can refuse a run the formula accepts.
> task='sft' only: the probe runs a plain causal-LM forward+backward, which IS
> the SFT step but not a preference loss. At the one preference shape measured it
> reads +13.5% high (probe 6.021 GB against a real DPO step's 5.304 GB) — the
> same safe direction it shows for SFT — but one shape is not a validation, so
> the restriction stands until it is measured across shapes.

For the other three sites, either the 0.934x/0.787x pair or naming the
denominator ("0.830x of the probe's peak") would resolve it; the test docstring
at `test_v07203.py:173` matters least but carries the same label.

Happy to open a PR if the reading is right. No measurement is involved — every
number above is already in the repo — but I have no CUDA card, so if the four
sites are quoting a series I have not found rather than the probe's, I cannot
check that and would rather be told than guess.

Environment: read against `main` at v0.73.3.

umran666:
Fixes #595.

Four code sites quoted the **probe-denominated** series (0.992x at seq 4096 / 0.830x at seq 5120) labelled as *"the real peak"*. Against the real training-run peak the gate record (`benchmarks/gate-v0.73.1-measured-vram-fit.md`) carries **0.934x at seq 5120 and 0.787x at 6144**; `0.830x` is the formula against the *probe* (predicted 3.844 GB / probe 4.633 GB = 0.8297). The crossover corroborates it: the quoted series turns over between seq 3072 and 4096 — the probe's crossover (record lines 100-102), not the training path's (between 4352 and 5120).

The `stream_vram_probe` field description also stated the probe *"under-measures [preference losses] badly enough to make the gate unsafe"* — a claim the gate record explicitly **withdraws** (lines 104-110): measured at the matching shape the probe reads **+13.5% high** (6.021 GB against a real DPO step's 5.304 GB), the same safe direction it shows for SFT. The `task='sft'` restriction is real but its stated reason was inverted — it stands because one shape is not a validation, matching the validator at `schema.py:5156-5175`.

The docs were already correct (`docs/performance-and-quantization.md`, `CONTRIBUTING.md`); this aligns the code with them.

## Changes

| site | before | after |
|---|---|---|
| `src/soup_cli/config/schema.py` (`stream_vram_probe` description) | "0.830x the real peak at seq 5120" + "under-measures those badly enough to make the gate unsafe" | 0.934x at 5120 / 0.787x at 6144 + the record's withdrawn-claim reasoning (+13.5% high at the one shape measured; one shape is not a validation) |
| `src/soup_cli/utils/layer_stream.py` (`decide_measured_fit` docstring) | "0.992x the real peak at 4096, 0.830x at 5120, under-predicts by 17%" | real-run series: holds to seq 4352 (1.081x), then 0.934x at 5120 / 0.787x at 6144, under-predicts by 21% |
| `src/soup_cli/utils/layer_stream_runtime.py` (`measure_step_peak_bytes` docstring) | "under-predicts past seq 4096 (0.830x at seq 5120)" | "measured against the real training run, under-predicts past seq 4352 (0.934x at 5120, 0.787x at 6144)" |
| `tests/test_v07203.py` (`test_never_under_predicts` docstring) | "0.992x the real peak at seq 4096 and 0.830x at 5120" | "0.934x the real peak at seq 5120 and 0.787x at 6144" |

No measurement involved — every figure is already in the repo (record lines 25-31 and 94-109). Comments/docstrings/field description only; no behaviour change.

## Test results

- `ruff check` on all four changed files: clean
- `pytest tests/test_v07203.py tests/test_issue349_measured_fit.py tests/test_issue348_calibrated_logits_wiring.py` — every test covering the changed areas passes (formula grid: 22 passed; probe config parsing: 5 passed). The failures in the local run are a pre-existing torch/transformers import breakage in my environment, identical on `main` without these changes (verified by stashing and re-running).
- Schema loads and the field description carries the corrected figures.

## Checklist

- [x] Relevant tests pass (docstring/description-only change)
- [x] Docs already consistent; this brings the code in line with them
- [x] No breaking changes

Closes #595


MakazhanAlpamys:
Welcome, @umran666 — I approved CI for your first PR. The correction is right and I verified every figure against the record rather than against your table:

```
formula / probe  @ seq 5120: 3.844 / 4.633 = 0.8297
formula / real   @ seq 5120: 3.844 / 4.118 = 0.9335
gate record row  @ seq 6144: 4.590 / 5.830 = 0.787   ("under by 21%")
crossover:       4352 -> 1.081x (over, safe);  5120 -> 0.934x (under)
```

So "past seq 4352" is the correct boundary, `0.934x` and `0.787x` are the real-run figures, and 21% is right for 6144. Your reading of the second half is also correct: `schema.py` was carrying the claim the gate record explicitly withdrew, contradicting the validator 1,900 lines below it in the same file. Good, careful work.

Two changes before I merge.

## 1. Please keep the probe series rather than replacing it

This is the one thing I asked for on #595 that did not land:

> make the four sites say which denominator each ratio uses, rather than swapping one number for another. `0.934x the real peak` and `0.830x the probe` are both true and both worth keeping — the gap between them is what tells a reader the probe is conservative, which is the property the gate depends on.

After this PR the probe series is gone from the codebase entirely:

```
$ git grep -n "0\.830\|0\.992" -- src/ tests/
(no matches)
```

and the probe-versus-real figures that establish conservativeness (+14.3% at 4352, +12.5% at 5120) appear nowhere in `src/` or `tests/` either. That matters because `stream_vram_probe`'s entire safety argument is that the probe over-reads relative to the real step, so it refuses slightly early rather than slightly late. A reader of `decide_measured_fit` can no longer see that from the code.

Please state both with their denominators, something like: *the formula holds to seq 4352 (1.081x the real peak) and then under-predicts — 0.934x at 5120, 0.787x at 6144. Against the probe the same series reads 0.992x and 0.830x, because the probe itself runs 12.5-14.3% above the real training step, which is the direction that makes it safe as a gate.*

## 2. The determinism sentence needs a source or removal

`layer_stream.py` said "Three repeats **at 4096** returned a bit-identical peak" and you changed it to "Three repeats **at the same shape**". I went looking for where that was measured and cannot find it in the v0.73.1 record:

```
$ grep -in "repeat|identical|determinis" benchmarks/gate-v0.73.1-measured-vram-fit.md
(no matches)
```

The only `4096` in that record is line 101, describing the **probe's** crossover. So the original sentence looks like it belonged to the probe series too — the very thing this PR is separating out. Vaguening it to "the same shape" keeps an unsourced claim and now attaches it to a paragraph that is otherwise entirely about the real training run, which re-creates the exact defect you are fixing: a measurement with an unstated denominator.

Either cite where the three repeats were measured and say against what, or drop the sentence. I would rather lose a true-but-unsourced claim than keep one whose denominator nobody can check. If you find it recorded somewhere I missed, say so and keep it with the source named — I would genuinely like to know.

## Everything else checks out

Comments, docstrings and one field description only; no behaviour change; the single test edit is a docstring. `ruff` clean. Your note that the local failures are a pre-existing torch/transformers breakage, verified by stashing and re-running on `main`, is exactly the right way to report that — thank you for not quietly omitting them.

One process note for next time, not a problem here: `changelog.d/0.73.3/605.fixed.md` is correctly named after your PR number. A sibling PR open right now got that wrong and filed under someone else's number, so it is worth knowing you got it right.

Push those two and I will merge once CI is green.

umran666:
## 1. Both series, denominators named

All four sites now carry both series side by side, e.g. the field description reads:

> The formula under-predicts past seq 4352 (measured 0.934x the real peak at seq 5120 and 0.787x at 6144) ... Against the probe the same formula reads 0.992x at seq 4096 and 0.830x at 5120, because the probe itself runs 12.5-14.3% above the real training step — the direction that makes it safe as a gate.

The 12.5-14.3% figures are the probe-vs-real row of the gate record (lines 94-97), so the conservativeness argument is now visible from the code. `git grep "0\.830\|0\.992" -- src/ tests/` matches all four sites again.

## 2. The determinism sentence — found the source, kept it with the citation

It is not in the v0.73.1 record; it is in **issue #395** itself:

> Crossover sits between seq 4352 and 5120. Deterministic (repeats at a fixed shape returned bit-identical peaks).

That is also why the original sentence's "three repeats at 4096" was suspect — #395 names no count and no shape, so the specifics appear to have drifted in with the probe series, exactly as you suspected. I rewrote the sentence to match the source verbatim and cited it: "The measurement is deterministic (repeats at a fixed shape return bit-identical peaks, #395)" — in `layer_stream.py` and in the `test_v07203.py` docstring, which carried the same bare "deterministically".

The changelog fragment is updated to describe the final state (both series kept, not swapped).

`ruff` clean; the affected non-torch tests pass (the torch-dependent ones fail identically on `main` in my environment — the pre-existing torch/transformers breakage from before).

MakazhanAlpamys:
Merged as `0a3920c`. Welcome, @umran666 — and both requested changes landed exactly as asked.

**Both series, denominators named.** `layer_stream.py` now reads:

> holds to seq 4352 (1.081x the real peak) and then fails: at seq 5120 the prediction is 0.934x the real peak and at seq 6144 it is 0.787x [...] Against the probe the same formula reads 0.992x at seq 4096 and 0.830x at 5120, because the probe itself runs 12.5-14.3% above the real training step — the direction that makes it safe as a gate.

That last clause is the part I wanted and did not spell out well enough the first time: the gap between the two series *is* the safety argument, so keeping only one number would have left nothing in the code recording that the probe over-reads.

**The determinism sentence now has a source, and you found it.** I said I could not locate "three repeats at 4096" anywhere in the v0.73.1 gate record, and that vaguening it to "at the same shape" kept an unsourced claim. You went and found where it actually lives — I verified: #395 line 13 reads *"Deterministic (repeats at a fixed shape returned bit-identical peaks)."* Citing it beats both deleting a true claim and leaving it floating.

I re-derived every figure against the record rather than against your table:

```
formula / probe  @ seq 5120: 3.844 / 4.633 = 0.8297
formula / real   @ seq 5120: 3.844 / 4.118 = 0.9335
gate record row  @ seq 6144: 4.590 / 5.830 = 0.787   ("under by 21%")
crossover:       4352 -> 1.081x (over, safe);  5120 -> 0.934x (under)
```

All 14 checks green on `f8bc3057`, no test deleted or weakened, and the single test edit is a docstring. Your note that the local failures were a pre-existing torch/transformers breakage — verified by stashing and re-running on `main` — is the right way to report that; reviewers discount unexplained red, and you removed the doubt instead of hoping I would not notice.

You are in `CONTRIBUTORS.md` as of `67f93e2`.

Your `#611` is in review now, in the same batch as three others. One process note that will save you a round there: it references `#371`, and that issue is deliberately **open** with several distinct fix paths — an earlier PR landed as `Refs #371`, not `Fixes`. Make sure your keyword matches the slice you actually took.

Issue #333, opened 2026-08-07
soup serve: vLLM finish_reason is hardcoded, --dashboard silently no-ops, and --max-model-len is missing


MakazhanAlpamys:
Three smaller defects from the first-ever run of `--backend vllm` (H100; the backend does not support Windows so it had never been exercised).

**1. `finish_reason` hardcoded `"stop"`.** Observed `"stop"` on a response with `completion_tokens == max_tokens == 64`, i.e. a length truncation reported as a natural stop. Clients that branch on `finish_reason` (continue-on-length is the common one) get the wrong answer.

**2. `--dashboard` silently no-ops on vLLM and SGLang.** `create_vllm_app` registers no `/metrics` route; `/metrics` returns **404** and nothing is printed. `docs/commands.md:72` advertises the flag without a backend qualifier. SGLang additionally has no `/v1/messages`, which `docs/serving-and-export.md:662` claims for both backends (true for vLLM, where it returns 200).

**3. No `--max-model-len` flag** although `create_vllm_engine` already accepts `max_model_len`. Serving Llama-3.1-8B failed with `16.0 GiB KV cache is needed, larger than available 10.72 GiB` and the only available lever was `--gpu-memory`.

### Acceptance
- [ ] `finish_reason` reports `"length"` when the generation hit `max_tokens`
- [ ] `--dashboard` either serves `/metrics` on these backends or refuses/warns at startup instead of doing nothing
- [ ] docs state which backends actually serve `/metrics` and `/v1/messages`
- [ ] `--max-model-len` is plumbed through to `create_vllm_engine`

Evidence: `benchmarks/gate-h100-validation.md` STEP 18.

MakazhanAlpamys:
All three fixed in `dd7c5bf`, each verified live on an H100.

**1. `finish_reason`** — was hardcoded `"stop"` even at `completion_tokens == max_tokens == 64`. Now `"length"` when truncated and `"stop"` on a natural stop, on the sync path, on the SSE final chunk, and on `/v1/messages` (mapped to `"max_tokens"` / `"end_turn"` — observed `"stop_reason":"max_tokens"` live at `max_tokens: 32`). Both branches are tested, because a fix that always says "length" would pass a one-sided test.

**2. `--dashboard`** — `/metrics` returned **404** with nothing printed. It is now a real route on the vLLM app with live counters, streamed requests included:
```json
[...]

PR #607, opened 2026-08-30
fix(mii): apply the model's chat template and report real finish_reason

### LINKED ISSUE #606: soup serve --backend mii hand-rolls its own prompt and hardcodes finish_reason
### Description

`soup serve --backend mii`'s `/v1/chat/completions` handler (`build_mii_app._chat` in
`src/soup_cli/utils/mii.py`) has two defects that the vLLM and transformers backends
already fixed for the same failure classes.

**1. It never applies the served model's chat template.** `_chat` builds its own
`"System: ...\nUser: ...\nAssistant:"` string instead of calling `build_chat_prompt`,
the single prompt builder `utils/vllm.py` says is shared by the transformers and vLLM
backends specifically so they cannot drift (#332). MII was never wired into it. A
chat-tuned model sees a prompt format it never trained on, the same defect class that
produced the Llama-3.1-8B run-on loop #332 fixed for vLLM.

**2. The response always reports `finish_reason: "stop"`** (line 185), regardless of
whether generation used its full `max_tokens` budget. #333 fixed the identical bug for
vLLM via `resolve_finish_reason`, deriving the value from the produced token count when
the engine's own value is untrustworthy. MII has no equivalent call, so a client cannot
tell a truncated answer from a completed one.

`mii.py` has no test file under `tests/`, so both defects are currently unguarded.

### Steps to reproduce

No GPU or `deepspeed-mii` install is needed: `build_mii_app` takes any callable
`pipeline(prompts, max_new_tokens=...) -> [Response]`, so a stub pipeline exercises the
same HTTP layer a real MII engine would hit.

```python
from fastapi.testclient import TestClient
from soup_cli.utils.mii import build_mii_app

captured = {}

def pipeline(prompts, max_new_tokens=None, **_):
    captured["prompt"] = prompts[0]
    return [type("R", (), {"generated_text": "x" * max_new_tokens,
                            "finish_reason": "length"})()]

app = build_mii_app(pipeline, model_name="m")
client = TestClient(app)
resp = client.post("/v1/chat/completions", json={
    "model": "m",
    "messages": [{"role": "system", "content": "You are terse."},
                 {"role": "user", "content": "2+2?"}],
    "max_tokens": 4,
})
print(captured["prompt"])                       # "System: You are terse.\nUser: 2+2?\nAssistant:"
print(resp.json()["choices"][0]["finish_reason"])  # "stop", even though the stub reported "length"
```

The prompt never reaches a template, and `finish_reason` is `"stop"` no matter what the
engine actually reports.

### Config (soup.yaml)

N/A, not config-dependent.

### Error output

Not an exception; both are silently wrong values, shown above (the printed prompt and
`finish_reason` in the reproduction).

### Environment

```
Python:   3.11.16
Platform: Linux 6.8.0-138-generic
Backend:  unknown (torch not installed, not needed to reproduce this)
pydantic  2.13.5 (OK, >=2.0.0)
typer     0.20.1 (OK, >=0.9.0)
rich      15.0.0 (OK, >=13.0.0)
fastapi   not installed by default (optional, required for `soup serve`)
```

Both defects are logic-level (prompt string assembly and a hardcoded literal), so they
reproduce identically on any platform; `soup doctor`'s GPU/training-stack rows are not
relevant to this bug.

AmirF194:
Fixes #606

`soup serve --backend mii`'s `/v1/chat/completions` handler hand-rolled its own
`"System: ...\nUser: ...\nAssistant:"` prompt instead of calling `build_chat_prompt`,
the single prompt builder #332 introduced so the transformers and vLLM backends
cannot drift from a chat-tuned model's own template. MII was never wired into it.
It also always reported `finish_reason: "stop"`, the same hardcoded-value bug #333
fixed for vLLM via `resolve_finish_reason`.

Both fixes mirror the vLLM backend's existing pattern exactly: `serve.py` now loads
the served model's tokenizer with the same `_load_serve_tokenizer` call `_serve_vllm`
already makes, and `build_mii_app` passes it to `build_chat_prompt` and derives
`finish_reason` with `resolve_finish_reason` instead of a literal. DeepSpeed-MII's own
`Response.finish_reason` is a `str`-backed enum with members `"stop"`/`"length"`, so it
flows straight through `resolve_finish_reason`'s trusted-value branch.

### Verification

- New regression tests in `tests/test_part_b.py::TestBuildMiiApp` fail on `main` and
  pass on this branch: one asserts the prompt sent to the pipeline comes from
  `build_chat_prompt` (not the hand-rolled string), the other asserts `finish_reason`
  reflects the engine's own reported value instead of always `"stop"`.
- Reproduced end to end in a container with a stub pipeline standing in for a real
  `deepspeed-mii` engine (GPU-only, not available here): before the fix, the prompt
  differed from `build_chat_prompt`'s output and `finish_reason` stayed `"stop"` even
  when the stub reported `"length"`; after the fix, both matched.
- `ruff check src/soup_cli/ scripts/ tests/`, all 11 tests in `test_part_b.py`, and
  the 6 MII tests in `test_multi_gpu.py` pass. I did not run the full matrix (the
  `[dev]` extra pulls in the training stack, which this change does not touch) or the
  `mlx-smoke`/`transformers-floor` jobs, which are unrelated to this file.

A changelog fragment will follow once the PR number is assigned (`changelog.d/<version>/<PR#>.fixed.md`).

### DIFF
--- a/changelog.d/0.73.3/607.fixed.md
+++ b/changelog.d/0.73.3/607.fixed.md
@@ -1,0 +1,3 @@
+- `soup serve --backend mii` now applies the served model's own chat template
+  (via `build_chat_prompt`) instead of a hand-rolled prompt, and reports the
+  engine's real `finish_reason` instead of always `"stop"` (#607).
--- a/src/soup_cli/commands/serve.py
+++ b/src/soup_cli/commands/serve.py
@@ -534,8 +534,25 @@ def serve(
             console.print(f"[red]Failed to create MII pipeline:[/] {exc}")
             raise typer.Exit(1) from exc
 
+        # #332's fix for the vLLM/transformers backends applies here too: a
+        # chat-tuned model needs its own template, not the generic
+        # role-prefixed fallback.
+        mii_tokenizer = _load_serve_tokenizer(
+            model_path=Path(model), base_model=None,
+            trust_remote_code=trust_remote_code,
+        )
+        if mii_tokenizer is None:
+            console.print(
+                "[yellow]Warning:[/] no tokenizer could be loaded for this "
+                "model, falling back to a generic 'User:/Assistant:' "
+                "prompt. Chat-tuned models can run on past their stop "
+                "token with this format."
+            )
+
         mii_model_name = Path(model).name
-        mii_app = build_mii_app(mii_pipeline, model_name=mii_model_name)
+        mii_app = build_mii_app(
+            mii_pipeline, model_name=mii_model_name, tokenizer=mii_tokenizer,
+        )
 
         import uvicorn
         console.print(
--- a/src/soup_cli/utils/mii.py
+++ b/src/soup_cli/utils/mii.py
@@ -91,6 +91,7 @@ def _ensure_mii_request_models():
 
 def build_mii_app(
     pipeline: Any, model_name: str, max_tokens_default: int = 512,
+    tokenizer: Any = None,
 ) -> Any:
     """Wrap a DeepSpeed-MII pipeline as a minimal OpenAI-compatible FastAPI
     app exposing ``/v1/chat/completions`` and ``/v1/models`` (#38, v0.33.0).
@@ -104,9 +105,14 @@ def build_mii_app(
             with ``.generated_text`` attributes.
         model_name: stable id surfaced in /v1/models and the response.
         max_tokens_default: default for requests that don't specify max_tokens.
+        tokenizer: the served model's own tokenizer, or None. Passed straight
+            through to ``build_chat_prompt`` (#332) so a chat-tuned model gets
+            its own template instead of the generic role-prefixed fallback.
     """
     from fastapi import FastAPI, HTTPException
     from fastapi.middleware.cors import CORSMiddleware
+
+    from soup_cli.utils.vllm import build_chat_prompt, resolve_finish_reason
 
     # Models are module-level (not closure) so FastAPI's introspection can
     # resolve forward refs.
@@ -147,16 +153,7 @@ def build_mii_app(
             raise HTTPException(
                 status_code=400, detail="max_tokens must be in [1, 16384]",
             )
-        prompt_parts = []
-        for msg in request.messages:
-            if msg.role == "system":
-                prompt_parts.append(f"System: {msg.content}")
-            elif msg.role == "user":
-                prompt_parts.append(f"User: {msg.content}")
-            elif msg.role == "assistant":
-                prompt_parts.append(f"Assistant: {msg.content}")
-        prompt_parts.append("Assistant:")
-        prompt = "\n".join(prompt_parts)
+        prompt = build_chat_prompt(request.messages, tokenizer)
 
         max_tokens = request.max_tokens or max_tokens_default
         try:
@@ -182,7 +179,7 @@ def build_mii_app(
             "choices": [{
                 "index": 0,
                 "message": {"role": "assistant", "content": text},
-                "finish_reason": "stop",
+                "finish_reason": resolve_finish_reason(first, max_tokens),
             }],
             "usage": {
                 "prompt_tokens": -1,
--- a/tests/test_part_b.py
+++ b/tests/test_part_b.py
@@ -134,6 +134,66 @@ class TestBuildMiiApp:
         })
         assert resp.status_code == 500
 
+    def test_chat_completions_applies_the_served_models_chat_template(self):
+        """A chat-tuned model must get its own template (#332), not the
+        generic 'System: .../User: .../Assistant:' prompt build_mii_app used
+        to hand-roll."""
+        pytest.importorskip("fastapi")
+        from fastapi.testclient import TestClient
+
+        from soup_cli.utils.mii import build_mii_app
+
+        captured = {}
+
+        def _pipeline(prompts, **_kwargs):
+            captured["prompt"] = prompts[0]
+            fake_response = MagicMock()
+            fake_response.generated_text = "4"
+            return [fake_response]
+
+        tokenizer = MagicMock()
+        tokenizer.chat_template = "<|template|>"
+        tokenizer.apply_chat_template.return_value = "<|from-the-model's-own-template|>"
+
+        app = build_mii_app(
+            _pipeline, model_name="test-mii-model", tokenizer=tokenizer,
+        )
+        client = TestClient(app)
+
+        resp = client.post("/v1/chat/completions", json={
+            "model": "test-mii-model",
+            "messages": [{"role": "user", "content": "2+2?"}],
+        })
+        assert resp.status_code == 200, resp.text
+        assert captured["prompt"] == "<|from-the-model's-own-template|>"
+        assert "User:" not in captured["prompt"]
+
+    def test_chat_completions_reports_length_when_engine_reports_length(self):
+        """The engine's own finish_reason must reach the client (#333),
+        not the hardcoded 'stop' build_mii_app used to always return."""
+        pytest.importorskip("fastapi")
+        from fastapi.testclient import TestClient
+
+        from soup_cli.utils.mii import build_mii_app
+
+        fake_response = MagicMock()
+        fake_response.generated_text = "truncated output"
+        fake_response.finish_reason = "length"
+
+        def _pipeline(prompts, **_kwargs):
+            return [fake_response]
+
+        app = build_mii_app(_pipeline, model_name="test-mii-model")
+        client = TestClient(app)
+
+        resp = client.post("/v1/chat/completions", json={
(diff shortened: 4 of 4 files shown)


MakazhanAlpamys:
**Heads up before either of you spends more time: #607 and #608 fix the same issue, independently.**

Timeline, so nobody has to guess:

- 16:27:54Z — @AmirF194 filed #606
- 16:28:44Z — @AmirF194 opened #607 (50 seconds later; found and fixed in one go)
- 16:52:06Z — @ARAVIND281 opened #608

There was no claim comment on #606, so @ARAVIND281 had no way to know #607 already existed unless they refreshed the PR list. **This is not anybody's fault, and I am not going to resolve it by timestamp.**

What I will do instead is what this project did the last time two good implementations landed on one issue: run the **same** battery against both, publish the **same** measurements in both PRs, and choose on merit — recording explicitly where the one I do not merge is better. Whichever way it goes, both of you get credit in `CONTRIBUTORS.md` for the work and for the finding.

Please **pause pushing** to either branch until I post that comparison, so I am measuring stable trees. If you would rather withdraw and take something else instead, say so and I will point you at an unclaimed issue — but there is no need to, and I would rather have both reviews on the record.

One thing already worth saying out loud, because you both independently identified it and it is the part most people miss: the fix is not just swapping in `build_chat_prompt`. It needs the **call site** to load and pass a tokenizer, otherwise `tokenizer` stays `None`, the template is never applied, and the change looks correct while doing nothing in production. That is exactly the defect shape I have been rejecting on the sibling SGLang PR (#581) for a week.

@ARAVIND281 — separately, welcome. I have approved CI on both of your PRs by hand, as this repo requires for a first-time contributor.

AmirF194:
Sounds good, thanks for the heads up. Pausing pushes to this branch until you've posted the comparison, and no objection to however it shakes out on merit.

MakazhanAlpamys:
## Head-to-head: #607 vs #608

As promised, the same battery against both, published identically in both PRs. @AmirF194 @ARAVIND281

**The `utils/mii.py` change is essentially the same in both.** Both delete the hand-rolled block, both call `build_chat_prompt(request.messages, tokenizer)`, both call `resolve_finish_reason(first, max_tokens)`, both import from `utils/vllm.py` inside the function to keep `soup --help` fast. Neither re-implements a parallel copy. So the diff that matters is the call site and the tests.

### Mutations — identical four, applied to both

| mutation | #607 (11 tests) | #608 (29 tests) |
|---|---|---|
| (a) revert the prompt to the hand-rolled `System:/User:/Assistant:` | **KILLED** — 1 failed | **KILLED** — 2 failed |
| (b) hardcode `finish_reason: "stop"` | **KILLED** — 1 failed | **KILLED** — 3 failed |
| (c) hardcode `finish_reason: "length"` | **SURVIVED** — 11 passed | **KILLED** — 2 failed |
| (d) delete the tokenizer wiring at the `serve.py` call site | **SURVIVED** — 11 passed | **SURVIVED** — 29 passed |

Named failures for the ones that died:

```
(a) 607 -> test_chat_completions_applies_the_served_models_chat_template
    608 -> test_applies_the_models_chat_template
           test_template_is_asked_for_a_generation_prompt
(b) 607 -> test_chat_completions_reports_length_when_engine_reports_length
    608 -> test_engine_reported_length_is_preserved
           test_full_budget_is_derived_as_length_without_a_reported_reason
           test_default_max_tokens_is_used_for_the_derivation
(c) 608 -> test_engine_reported_stop_is_preserved
           test_short_output_is_stop
```

**On my method, since it changes one of these numbers:** my first attempt at (a) injected a broken string literal and produced a `SyntaxError`, which "killed" the mutation on both PRs for the wrong reason (8 and 26 failures). I caught it with `py_compile`, redid it, and the table above is the corrected run. Recording it because a mutation that fails to apply and reports a pass is the exact false-green this project has been bitten by, and it would have flattered both of you.

### The two differences

**1. `finish_reason` in the `"stop"` direction (mutation c).** #607's suite pins `"length"` but has no `"stop"` control, so a server that reports `"length"` on **every** response — the mirror image of the bug being fixed, and one that makes a continue-on-length client loop forever — passes it. #608's `TestMiiFinishReason` covers both directions.

**2. The operator warnings.** `_serve_vllm` prints one of three messages under the comment *"A failure here is announced, never silent"*: no tokenizer / no chat template / applying the model's own template. #608 ports all three. #607 ports the first only, so a model that loads a tokenizer but ships no chat template falls back silently.

### The gap they share

**Mutation (d) survives on both.** Deleting `tokenizer=mii_tokenizer` from the `build_mii_app(...)` call in `serve.py` — so every real `soup serve --backend mii` falls back to the legacy prompt — leaves 11/11 and 29/29 green. Every test in both PRs constructs the app directly with a hand-supplied tokenizer, so the one line a user actually executes is uncovered in both. This is the same defect shape I have been holding #581 on, and it needs closing before merge.

### Decision

**I intend to merge #608**, on the two differences above: it catches the `"stop"` direction and it mirrors the vLLM warning contract.

**What #607 does better, on the record:**

- **@AmirF194 found and filed the defect.** #606 is theirs, opened 50 seconds before their PR. #608 exists because that issue existed. The finding is the scarcer half of this and it is not diminished by the merge decision.
- Its tests live in the existing `tests/test_part_b.py::TestBuildMiiApp` alongside the other `build_mii_app` coverage, rather than a new file. Reasonable people differ, but colocation has an argument.
- Its diff is 89 lines against 293 for the same production change.

**One thing #608 should carry over from #607's fragment discipline, and one it should fix:** #607's fragment is tighter, but cites only `#607` and not the issue; #608's cites `#606 in #608`, which is the correct form. Keep #608's.

### Required before I merge #608

@ARAVIND281 — one addition: a test that drives the `serve.py` MII path and asserts a tokenizer is loaded and passed into `build_mii_app`, so mutation (d) fails by name. Assert on the call kwargs; it does not need a real MII pipeline. That is the difference between pinning the fix and pinning an implementation detail.

@AmirF194 — I will close #607 once #608 lands, with credit for the finding and for the independent implementation, and you both go into `CONTRIBUTORS.md`. If you would rather be the one to add the call-site test and have me merge #607 instead, say so within the day and I will re-run the battery on an updated #607 — the two differences above are both closeable, and I would rather that than have you feel the 24-minute gap decided it. It did not; the `"stop"` control did.

Both trees: no test deleted or weakened, changelog fragment present and correctly placed, neither touches docs.

AmirF194:
Appreciate the rigor here, especially catching your own syntax-error false-kill before publishing. The stop-direction gap and the missing warning messages are real, and #608 already covers both without me needing to bolt them on just to keep pace. I'd rather you merge the more complete implementation than have me pad #607 to match it after the fact.

No objection to closing #607 once #608 lands. Thanks for running the same battery on both and for the credit either way.

MakazhanAlpamys:
Closing in favour of #608, which merged as `763a654`. @AmirF194 — this is not a judgement on your implementation, and I want the record to say so precisely.

**The finding is yours.** You filed #606 and opened this PR fifty seconds later. #608 exists because that issue existed. Finding a defect nobody had reported is the scarcer half of this work, and you have it credited in `CONTRIBUTORS.md` and in the merge commit body.

**Your `utils/mii.py` change was equivalent** — same reuse of `build_chat_prompt` and `resolve_finish_reason`, same deletion of the hand-rolled block, no parallel copy. The decision came down to two measured differences from the battery I ran identically against both trees and published in both PRs:

- hardcoding `finish_reason` to `"length"` — the mirror image of the bug, and one that makes a continue-on-length client loop forever — **passed** this suite and failed three tests on #608;
- `_serve_vllm` prints one of three operator messages under the comment "A failure here is announced, never silent"; #608 ported all three, this PR ported the first.

Both were closeable and I offered you the chance to close them. You chose to pause pushes and let the comparison stand, which was the graceful call, and I would rather say plainly that it cost you the merge than pretend the timestamps decided it. They did not — I said so in the comparison and I am saying it again here.

For completeness on the gap both PRs shared: neither tested the `serve.py` call site, so deleting the tokenizer wiring left every test green while production fell back to the legacy prompt. #608 closed that before merge with `TestServeMiiCallSite`.

Your #591 merged earlier today as `14a93ec`, and #352 is reopened with the scope note we agreed. If you want the next MII-adjacent piece, `utils/sglang.py` still hardcodes its prompt in the *streaming* path — I merged #581 for the non-streaming half — and #528's Lambda backend has an unclaimed API-key-in-boot-log problem. Say the word on either.

AmirF194:
Thanks for writing that out in full, and for the credit either way. The length-vs-length mirror bug and the missing operator messages are real gaps in this PR and #608 is the right one to have merged.

Good to hear #591 landed and #352 is back open. I'll take a look at the sglang streaming prompt and the #528 boot-log key next.

PR #608, opened 2026-08-30
fix(serve): apply the chat template and derive finish_reason on the MII backend

### LINKED ISSUE #606: soup serve --backend mii hand-rolls its own prompt and hardcodes finish_reason

ARAVIND281:
Closes #606

`soup serve --backend mii`'s `/v1/chat/completions` handler carried the two defects the issue describes. Both are fixed by routing MII through the helpers the other backends already use, rather than by writing anything new.

## 1. The chat template (#332 failure class)

`_chat` hand-rolled `"System: …\nUser: …\nAssistant:"` instead of calling `build_chat_prompt`, the single builder `utils/vllm.py` documents as shared "so the two cannot drift". MII was never wired in.

It is now. The hand-rolled block is gone — 10 lines replaced by one call.

The important half is that this needed a change at the **call site** too: `build_mii_app` had no tokenizer to apply. `serve.py` now loads the served model's tokenizer for the MII path via the existing `_load_serve_tokenizer`, exactly as the vLLM path does, and prints the same three warnings (no tokenizer / no chat template / template applied). Without that, `tokenizer` would always be `None` and the template would never actually be applied in production — the fix would look right and do nothing.

**Behaviour is unchanged for a model with no template**: `_legacy_chat_prompt` — `build_chat_prompt`'s fallback — is byte-identical to the block that was removed. A broken template also falls back rather than 500-ing, which `build_chat_prompt` already handles.

## 2. `finish_reason` (#333 failure class)

Line 185 hardcoded `"stop"`. It now calls `resolve_finish_reason(first, max_tokens)` — the same helper #333 added for vLLM, which trusts the engine's value when it is one Soup emits, normalises anything else (MII, like vLLM, can report values outside OpenAI's vocabulary), and otherwise derives the answer from the produced token count.

## Tests

`mii.py` had no test file, which is how both defects stayed unguarded. Adds `tests/test_mii_serve.py` — 20 tests, **no GPU and no `deepspeed-mii` install needed**, since `build_mii_app` accepts any callable pipeline and a stub exercises the same HTTP layer.

Worth being precise about what they prove. Against the pre-fix logic (signature kept, only the two behaviours reverted, so the failures are real rather than construction errors):

```
5 failed, 15 passed
  TestMiiChatTemplate::test_applies_the_models_chat_template
  TestMiiChatTemplate::test_template_is_asked_for_a_generation_prompt
  TestMiiFinishReason::test_engine_reported_length_is_preserved
  TestMiiFinishReason::test_full_budget_is_derived_as_length_without_a_reported_reason
  TestMiiFinishReason::test_default_max_tokens_is_used_for_the_derivation
```

Those five are the regression cover. The other 15 are guards on the surrounding contract — streaming still rejected, `max_tokens` bounds, 500 paths, `/v1/models`, sampling-param forwarding, response envelope — and pass either way, which is what makes them useful as guards.

## Checks

```
pytest tests/            19213 passed, 92 skipped   (14m)
```

A changelog fragment is added in a follow-up commit on this branch, now that the PR number exists.


MakazhanAlpamys:
**Heads up before either of you spends more time: #607 and #608 fix the same issue, independently.**

Timeline, so nobody has to guess:

- 16:27:54Z — @AmirF194 filed #606
- 16:28:44Z — @AmirF194 opened #607 (50 seconds later; found and fixed in one go)
- 16:52:06Z — @ARAVIND281 opened #608

There was no claim comment on #606, so @ARAVIND281 had no way to know #607 already existed unless they refreshed the PR list. **This is not anybody's fault, and I am not going to resolve it by timestamp.**

What I will do instead is what this project did the last time two good implementations landed on one issue: run the **same** battery against both, publish the **same** measurements in both PRs, and choose on merit — recording explicitly where the one I do not merge is better. Whichever way it goes, both of you get credit in `CONTRIBUTORS.md` for the work and for the finding.

Please **pause pushing** to either branch until I post that comparison, so I am measuring stable trees. If you would rather withdraw and take something else instead, say so and I will point you at an unclaimed issue — but there is no need to, and I would rather have both reviews on the record.

One thing already worth saying out loud, because you both independently identified it and it is the part most people miss: the fix is not just swapping in `build_chat_prompt`. It needs the **call site** to load and pass a tokenizer, otherwise `tokenizer` stays `None`, the template is never applied, and the change looks correct while doing nothing in production. That is exactly the defect shape I have been rejecting on the sibling SGLang PR (#581) for a week.

@ARAVIND281 — separately, welcome. I have approved CI on both of your PRs by hand, as this repo requires for a first-time contributor.

ARAVIND281:
Understood, and thanks for handling it this way rather than by timestamp — that's more than fair.

I've stopped pushing to both branches; the trees are stable for measurement and will stay that way until you post the comparison.

The miss is mine: I checked #606 for a claim comment before starting, but never checked whether an open PR already referenced it. #607 was 24 minutes ahead and I'd have seen it with one more query. Noted for next time.

No argument from me on the outcome — if @AmirF194's is the better tree, merge theirs. I'd genuinely like to read the comparison either way; the call-site tokenizer point you raised is the interesting part, and I'd rather see how someone else handled it than win on it.

#609 is unrelated and uncontested, if that one's easier to clear first.

ARAVIND281:
Not asking for anything here — posting evidence so you don't have to dig it out during the comparison. A red CI email landed on this branch and it is **not** from this change.

**The failure is a process crash, not a test failure.**

```
Windows fatal exception: code 0xc000001d
  File "D:\a\Soup\Soup\tests\test_v05311.py", line 732
       in test_real_trl_grpotrainer_end_to_end_step
```

`0xc000001d` is EXCEPTION_ILLEGAL_INSTRUCTION — the CPU hit an instruction it does not support inside native torch/TRL code, which kills the interpreter. That is why the log tail is a pytest-internals stack dump and `exit code 1` with no `FAILED tests/...` line.

**Why it is not mine:**

1. **Windows 3.11 and 3.12 passed on the identical commit.** Only `windows-latest, 3.10` is red. A change of mine that broke Windows would break all three.
2. The crash is in `TestGRPOVariantRuntimeContract` — a real TRL GRPOTrainer step. `test_v05311.py` imports nothing from `mii` or `serve`. This PR touches `utils/mii.py`, `commands/serve.py`, `tests/test_mii_serve.py` and a changelog fragment.
3. #609 passed `windows-latest, 3.10` on the same base (run 33324975813).
4. The failing run was commit `ed79c09`; `1c2034c` is green 13/13. **The only difference between them is the one-line changelog fragment** — a markdown file cannot change whether native code hits an illegal instruction.

**It is not specific to me either.** Recent failing runs in the repo:

```
33323657187  fix/mii-chat-template-finish-reason   (mine)
33301319805  feat/cloud-training-264
33262814628  feat/bash-sandbox-151
33242531165  feat/bash-sandbox-151
```

Three contributors, unrelated features. That looks like a pre-existing flake on the Windows 3.10 runner rather than anything any of us introduced — possibly worth its own issue, since an illegal-instruction crash usually means a wheel built for instructions the runner's CPU lacks.

I have not pushed anything since your freeze request, and won't. The tip of this branch is green.

MakazhanAlpamys:
## Head-to-head: #607 vs #608

As promised, the same battery against both, published identically in both PRs. @AmirF194 @ARAVIND281

**The `utils/mii.py` change is essentially the same in both.** Both delete the hand-rolled block, both call `build_chat_prompt(request.messages, tokenizer)`, both call `resolve_finish_reason(first, max_tokens)`, both import from `utils/vllm.py` inside the function to keep `soup --help` fast. Neither re-implements a parallel copy. So the diff that matters is the call site and the tests.

### Mutations — identical four, applied to both

| mutation | #607 (11 tests) | #608 (29 tests) |
|---|---|---|
| (a) revert the prompt to the hand-rolled `System:/User:/Assistant:` | **KILLED** — 1 failed | **KILLED** — 2 failed |
| (b) hardcode `finish_reason: "stop"` | **KILLED** — 1 failed | **KILLED** — 3 failed |
| (c) hardcode `finish_reason: "length"` | **SURVIVED** — 11 passed | **KILLED** — 2 failed |
| (d) delete the tokenizer wiring at the `serve.py` call site | **SURVIVED** — 11 passed | **SURVIVED** — 29 passed |

Named failures for the ones that died:

```
(a) 607 -> test_chat_completions_applies_the_served_models_chat_template
    608 -> test_applies_the_models_chat_template
           test_template_is_asked_for_a_generation_prompt
(b) 607 -> test_chat_completions_reports_length_when_engine_reports_length
    608 -> test_engine_reported_length_is_preserved
           test_full_budget_is_derived_as_length_without_a_reported_reason
           test_default_max_tokens_is_used_for_the_derivation
(c) 608 -> test_engine_reported_stop_is_preserved
           test_short_output_is_stop
```

**On my method, since it changes one of these numbers:** my first attempt at (a) injected a broken string literal and produced a `SyntaxError`, which "killed" the mutation on both PRs for the wrong reason (8 and 26 failures). I caught it with `py_compile`, redid it, and the table above is the corrected run. Recording it because a mutation that fails to apply and reports a pass is the exact false-green this project has been bitten by, and it would have flattered both of you.

### The two differences

**1. `finish_reason` in the `"stop"` direction (mutation c).** #607's suite pins `"length"` but has no `"stop"` control, so a server that reports `"length"` on **every** response — the mirror image of the bug being fixed, and one that makes a continue-on-length client loop forever — passes it. #608's `TestMiiFinishReason` covers both directions.

**2. The operator warnings.** `_serve_vllm` prints one of three messages under the comment *"A failure here is announced, never silent"*: no tokenizer / no chat template / applying the model's own template. #608 ports all three. #607 ports the first only, so a model that loads a tokenizer but ships no chat template falls back silently.

### The gap they share

**Mutation (d) survives on both.** Deleting `tokenizer=mii_tokenizer` from the `build_mii_app(...)` call in `serve.py` — so every real `soup serve --backend mii` falls back to the legacy prompt — leaves 11/11 and 29/29 green. Every test in both PRs constructs the app directly with a hand-supplied tokenizer, so the one line a user actually executes is uncovered in both. This is the same defect shape I have been holding #581 on, and it needs closing before merge.

### Decision

**I intend to merge #608**, on the two differences above: it catches the `"stop"` direction and it mirrors the vLLM warning contract.

**What #607 does better, on the record:**

- **@AmirF194 found and filed the defect.** #606 is theirs, opened 50 seconds before their PR. #608 exists because that issue existed. The finding is the scarcer half of this and it is not diminished by the merge decision.
- Its tests live in the existing `tests/test_part_b.py::TestBuildMiiApp` alongside the other `build_mii_app` coverage, rather than a new file. Reasonable people differ, but colocation has an argument.
- Its diff is 89 lines against 293 for the same production change.

**One thing #608 should carry over from #607's fragment discipline, and one it should fix:** #607's fragment is tighter, but cites only `#607` and not the issue; #608's cites `#606 in #608`, which is the correct form. Keep #608's.

### Required before I merge #608

@ARAVIND281 — one addition: a test that drives the `serve.py` MII path and asserts a tokenizer is loaded and passed into `build_mii_app`, so mutation (d) fails by name. Assert on the call kwargs; it does not need a real MII pipeline. That is the difference between pinning the fix and pinning an implementation detail.

@AmirF194 — I will close #607 once #608 lands, with credit for the finding and for the independent implementation, and you both go into `CONTRIBUTORS.md`. If you would rather be the one to add the call-site test and have me merge #607 instead, say so within the day and I will re-run the battery on an updated #607 — the two differences above are both closeable, and I would rather that than have you feel the 24-minute gap decided it. It did not; the `"stop"` control did.

Both trees: no test deleted or weakened, changelog fragment present and correctly placed, neither touches docs.

ARAVIND281:
Added, and verified against the mutation rather than assumed. Pushed as `dc232d5`.

`TestServeMiiCallSite` drives the real CLI — `serve --model ... --backend mii` — with the availability check, pipeline factory, tokenizer loader and `uvicorn.run` patched, then asserts on the kwargs `build_mii_app` actually received:

- the `tokenizer` kwarg is present at all
- it **is** the object `_load_serve_tokenizer` returned (identity, not truthiness — a test that only checked `is not None` would pass on a wrong-object bug)
- `_load_serve_tokenizer` was actually called
- a `None` tokenizer still starts the server, since `build_chat_prompt` degrades to the legacy format rather than raising

Applying mutation (d) — deleting `tokenizer=mii_tokenizer` from the `build_mii_app(...)` call:

```
2 failed, 20 passed
  test_serve_mii_passes_the_loaded_tokenizer_to_build_mii_app
  test_serve_mii_still_starts_when_no_tokenizer_can_be_loaded
```

Restored: 22 passed. I ran `py_compile` on the mutant first, so those are real failures and not the SyntaxError false-kill you hit — thanks for writing that up, it's the reason I checked.

On the comparison itself: the `"stop"`-direction gap in (c) is a fair call and I hadn't thought of it as the deciding property. And @AmirF194 found and filed the defect 50 seconds before opening #607 — this PR exists because that issue existed, which is the harder half. Genuinely fine either way if they'd rather close the two gaps on #607 and have you merge that instead; the call-site test above is small and would port over almost unchanged.

Also noting the fragment point: mine cites `#606 in #608`, so no change needed there.

Issue #328, opened 2026-08-03
Layer streaming: a meta placeholder reaches the gradient-checkpointing backward recompute — all four preference losses fail on torch 2.13 (CUDA), SFT passes


MakazhanAlpamys:
> **Re-scoped 2026-08-06 after measurement on 8x H100.** This issue was filed as
> *CPU-only, KTO-only*. It is neither. On torch 2.13 + CUDA **all four streamed
> preference losses fail — `dpo`, `orpo`, `simpo`, `kto` — while `sft` passes**, and
> the failure is in the gradient-checkpointing backward recompute, not in KTO's KL
> forward. Everything below is updated; the original framing is kept only where it
> is still true. Evidence: [`benchmarks/gate-h100-validation.md`](https://github.com/MakazhanAlpamys/Soup/blob/main/benchmarks/gate-h100-validation.md), FINDING 2.

## Symptom

A full `train()` step over a **streamed** model raises:

```
RuntimeError: Tensor on device cuda:0 is not on the expected device meta!
  .../torch/_prims_common/__init__.py:931 in check_same_device
```

(On CPU the same failure reads `Tensor on device cpu is not on the expected device
meta!`.) An operation received a `meta` placeholder alongside a real tensor, via
torch's decomposition path.

## Scope — measured, not inferred

Box: 8x H100 80GB HBM3, Ubuntu 24.04.3, Python 3.12.3, **torch 2.13.0+cu130 / CUDA
13.0**, transformers 4.57.6, trl 0.26.2, peft 0.20.0, accelerate 1.14.0,
bitsandbytes 0.50.0. `pip install -e ".[train]"` resolved that stack on its own; it
was not pinned. Every run below has `CUDA_VISIBLE_DEVICES` pinned to one card, to
keep the separate multi-GPU `nn.DataParallel`-vs-`meta` defect out of the picture.

Direct test of the four losses plus an SFT control, one streamed `train()` step each:

```
$ CUDA_VISIBLE_DEVICES=0 python repro_328.py
torch 2.13.0+cu130 cuda_devices 1
sft    OK
dpo    FAIL RuntimeError: Tensor on device cuda:0 is not on the expected device meta!
orpo   FAIL RuntimeError: Tensor on device cuda:0 is not on the expected device meta!
simpo  FAIL RuntimeError: Tensor on device cuda:0 is not on the expected device meta!
kto    FAIL RuntimeError: Tensor on device cuda:0 is not on the expected device meta!
```

[...]

PR #609, opened 2026-08-30
test(streaming): cover orpo and simpo with a real CUDA train() step

### LINKED ISSUE #370: No CUDA train() coverage for orpo and simpo: the gap is why #328 was reported as dpo/kto-only
## What

The streaming test suite has no CUDA test that actually calls `train()` for `task=orpo` or `task=simpo`. The parametrized class that exercises the preference losses on CUDA covers **dpo and kto only**.

## Why this is a demonstrated gap, not a hypothetical one

While diagnosing #328 on the H100 box, the initial hypothesis was that the meta leak was reachable from the *reference forward*, because only `[dpo]` and `[kto]` failed — "which would have been a sharp localization. **It is wrong.**" (`benchmarks/gate-h100-validation.md` L488–509).

The reason only those two appeared is that **the test class is only parametrized over those two**. ORPO and SimPO are genuinely reference-free (v0.72.4 verified they have no `ref_model` attribute at all) and they fail too — confirmed directly by `repro_328.py`, where all four preference losses raise `RuntimeError: Tensor on device cuda:0 is not on the expected device meta!` (L497–504).

So the coverage gap did not merely fail to catch a bug: it actively produced a **wrong localisation** that a person then reasoned from.

There is a second, quieter consequence: `#328`'s original scope said CPU-only and KTO-only. Both halves were wrong, and the record notes it explicitly — "Recorded so #328 can be re-scoped: it is not CPU-only and not KTO-only" (L532–535).

## Fix path

Extend the CUDA preference-loss `train()` parametrisation to all four (`dpo`, `kto`, `orpo`, `simpo`), and keep `bco`/`ipo` in mind for the same treatment since they share the wrapper family.

## Acceptance criteria

- [ ] `orpo` and `simpo` each have a CUDA-gated test that calls the real `setup()` + `train()`, not merely a config-parse or a `_build_trainer` check
- [ ] The test is written so that reverting the #328 fix turns it red — verified by actually reverting, not assumed
- [ ] A note in the test file that the parametrisation list is the localisation, so a future narrowing of it is visibly a scope change

ARAVIND281:
Closes #370

Extends the CUDA preference-loss `train()` coverage to all four losses, and adds a slot that guards the #328 mechanism directly.

Verified on an **A10G** (`torch 2.13.0+cu130`, `trl 0.29.1`, `transformers 5.16.1`) — the version combination the #328 comment names.

## What changed

- `test_weight_bearing_terms_are_identical_to_sft` now runs over `_ALL_PREFERENCE` instead of `_REFERENCE_USING`, so the existing VRAM claim covers orpo and simpo too.
- New CUDA-gated `TestEveryPreferenceLossTakesAStreamedStep`, two arms per loss:
  - `test_a_streamed_train_step_completes` — a real `setup()` + `train()` step.
  - `test_hf_checkpointing_is_off_while_streaming` — asserts the mechanism directly, so a green step cannot be a coincidence of some other layer absorbing the double-recompute.
- A note on the two parametrisation tuples recording that they *are* the localisation, so narrowing either is visibly a scope change rather than an edit (acceptance criterion 3).

It is a separate class rather than more arms on `TestPeakVramIsNotDoubled` because the property differs: those measure what a step costs, this asserts a step happens at all. Hanging the #328 guard off a VRAM test means a later refactor of the memory assertions deletes the regression cover silently.

## Reverting actually turns it red (criterion 2)

Both revert forms, measured — not assumed:

| | `train_step_completes` | `hf_checkpointing_is_off` |
|---|---|---|
| fix intact | 4 passed | 4 passed |
| **(a)** drop `and not stream_layers` | **4 failed** | **4 failed** |
| **(b)** remove the kwarg entirely (pre-fix state) | **4 failed** | **4 failed** |

Both fail with the issue's exact error:

```
RuntimeError: Tensor on device cuda:0 is not on the expected device meta!
```

All four losses, not two — which is the whole point of #370.

### One thing worth flagging in the test itself

My first draft would have shipped as decorative. It called `setup()` + `train()` for orpo and simpo on CUDA, satisfied the issue's wording, passed — and passed just as happily with the guard deleted, because it used the config default `gradient_checkpointing=False`, where the reverted expression `bool(False)` returns `False` anyway.

Only running the revert exposed it. The arm now passes `gradient_checkpointing=True` — the configuration the bug actually lives in — and the docstring records the measurement so the argument does not get "simplified" away later. Criterion 2 earned its place.

## A finding beyond the issue's scope

Revert form (b) — removing the `gradient_checkpointing=` kwarg entirely, which the code comment describes as the pre-fix state — does more than re-enable checkpointing. On `trl 0.29.1`, `DPOConfig`, `KTOConfig`, `ORPOConfig` and `CPOConfig` **all default it to `True`**, against `False` on the transformers base class. Measured:

```
WITH the fix:   user asked False -> args.gradient_checkpointing = False   (dpo, orpo, simpo, kto)
FORM (b):       user asked False -> args.gradient_checkpointing = True    (dpo, orpo, simpo, kto)
```

So the guard is not only "force HF checkpointing off while streaming" — it is the only thing stopping TRL's version-dependent default from silently reversing a user's explicit `gradient_checkpointing: false` and then crashing the run. The comment's second sentence about TRL's unstable default is the load-bearing half on current TRL, not a footnote. Might be worth saying so where the guard is defined; happy to add that in this PR if you want it.

## Acceptance criteria

- [x] `orpo` and `simpo` each have a CUDA-gated test calling the real `setup()` + `train()`, not a config-parse or `_build_trainer` check
- [x] Reverting the #328 fix turns it red — verified by actually reverting, in both forms
- [x] A note that the parametrisation list is the localisation

## Checks

```
tests/test_v07204.py on A10G     78 passed  (baseline on upstream: 68)
pytest tests/ (CPU)              19193 passed, 102 skipped
```

The 102 skips are 92 baseline + the 10 new CUDA arms skipping off-GPU.

No changelog fragment: this is test-only and not user-visible. Say the word if you would rather have one.

### DIFF
--- a/tests/test_v07204.py
+++ b/tests/test_v07204.py
@@ -388,6 +388,17 @@ def _build_streamed_wrapper(
 # ==========================================================================
 # item 1 -- schema: which tasks may stream
 # ==========================================================================
+# These two tuples are the LOCALISATION, not a convenience. #370: while
+# diagnosing #328 on the H100 box only [dpo] and [kto] failed, which read as a
+# sharp localisation to the reference forward — and was wrong. They were the
+# only two the CUDA class was parametrised over. ORPO and SimPO are genuinely
+# reference-free (v0.72.4: no `ref_model` attribute at all) and failed too.
+# So: narrowing either tuple silently narrows what the suite can localise, and
+# is a scope change that must be argued for, not an edit.
+#
+# _REFERENCE_USING is for properties that only exist when there IS a reference
+# (the disabled-adapter identity, the reference forward). Everything that
+# exercises the streamed train() step belongs to _ALL_PREFERENCE.
 _REFERENCE_USING = ("dpo", "kto")
 _ALL_PREFERENCE = ("dpo", "orpo", "simpo", "kto")
 
@@ -566,7 +577,7 @@ class TestPeakVramIsNotDoubled:
         torch.cuda.empty_cache()
         return peak, stats
 
-    @pytest.mark.parametrize("task", _REFERENCE_USING)
+    @pytest.mark.parametrize("task", _ALL_PREFERENCE)
     def test_weight_bearing_terms_are_identical_to_sft(self, tmp_path, monkeypatch, task):
         """One store, one pool. The gate measured 729.91 MB / 60.83 MB for both
         arms on a 730 MB model; here the sizes are tiny but the EQUALITY is the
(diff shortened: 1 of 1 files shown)


MakazhanAlpamys:
Merged as `091b017`. Welcome, @ARAVIND281 — and this is the acceptance criterion most PRs skip, so I want to be specific about it.

**Criterion 2 reproduced at merge, on a different stack than yours.** You measured on an A10G (torch 2.13 / trl 0.29.1 / transformers 5.16.1). I ran it on an RTX 3050 with torch 2.5.1+cu121 / transformers 4.57.6 / trl 0.19.1 / peft 0.18.1:

```
baseline                                        8 passed  in 131.96s
drop `and not stream_layers` (layer_stream:890) 8 failed  in  89.51s
  RuntimeError: Tensor on device cuda:0 is not on the expected device meta!
  FAILED ...test_hf_checkpointing_is_off_while_streaming[dpo|orpo|simpo|kto]
```

Both arms fail on all four losses, with the issue's exact error. **That makes the #328 failure not torch-2.13-specific**, which is stronger than your PR body claims and stronger than the `orpo.py` comment implies.

Your revert form (b) — removing the kwarg entirely — **survives** on my stack, and I checked why rather than treating it as a discrepancy: trl 0.19.1 defaults `gradient_checkpointing = False` on `DPOConfig`/`KTOConfig`/`ORPOConfig`/`CPOConfig`, so there is no double-recompute to catch. Your figure is right on 0.29.1, where they default `True`. What matters is that **the test file's own docstring only claims form (a) turns it red** — so the merged artifact does not overstate what it establishes. That is the correct place for that caveat and you put it there.

Two structural calls I want to credit:

- **Separate class, not more arms on `TestPeakVramIsNotDoubled`.** Your reasoning — those measure what a step *costs*, this asserts a step *happens*, and hanging regression cover off a memory assertion means a later refactor deletes it silently — is right, and it is the kind of thing that only shows up years later.
- **The widening strengthens rather than weakens.** `test_weight_bearing_terms_are_identical_to_sft` moving from `_REFERENCE_USING` to `_ALL_PREFERENCE` was the one deleted line in the diff, and its assertions never depended on a reference model existing. Its paired control correctly stays at `_REFERENCE_USING`, since a reference-free loss cannot build a second instance.

Criterion 3's note sits above **both** tuples, and the new arms are `_cuda_available()`-gated so they skip in every CI cell — matching the pre-existing classes in that file, so established precedent rather than a new never-executing gate. I said so in the merge body so nobody mistakes it later. No changelog fragment, correctly: #594 was also test-only and shipped none.

One thing the issue asked that turned out to rest on a false premise, so nobody re-opens it: "keep bco/ipo in mind for the same treatment". `SUPPORTED_STREAM_TASKS` is `("sft", "dpo", "orpo", "simpo", "kto")`, and bco/ipo are refused at config parse — there is no streamed `train()` step for them to cover. Worth a line in the tuple comment if you are ever back in that file.

You are in `CONTRIBUTORS.md` as of `41a5aaf`. Your #608 and #610 are both in flight — see the head-to-head comparison on #608, and I will post #610 shortly.

PR #610, opened 2026-08-30
bench(streaming): measure the VRAM fit on a second GPU/stack (#395)

### LINKED ISSUE #395: Layer streaming: the VRAM formula UNDER-predicts past ~4700 tokens and the mechanism is unidentified

ARAVIND281:
Closes #395

The verification #395 asks for in its last criterion: the VRAM fit measured on a **second GPU/stack**, since everything before it is one card.

**A10G 23 GB** (AWS `g5.xlarge`), Ubuntu 22.04, torch 2.13.0+cu130, transformers 5.16.1, trl 0.29.1, peft 0.20.0 — against the RTX 3050 Laptop 4 GB / Windows 11 / torch 2.5.1 / transformers 4.57.6 the fit was taken on.

Record: [`benchmarks/gate-395-second-stack-vram.md`](benchmarks/gate-395-second-stack-vram.md). Harness: [`benchmarks/harness/issue395_second_stack_vram.py`](benchmarks/harness/issue395_second_stack_vram.py).

> **Note on this description.** An earlier version of it led with a back-solved figure of 11.83 B/element. That number is **withdrawn** — see "What the gap actually is" below. The body has been rewritten so it states the finding as it stands rather than as it first appeared; the record and tests were corrected in `ee2bbcc`.

## 1. The under-prediction does not reproduce

Real `soup train` setup + one step, bf16, batch 1, `quantization: none`, LoRA r=8, `stream_buffers: 2`. "Real peak" is `torch.cuda.max_memory_allocated()`.

| seq | this stack | RTX 3050 (real-peak series) |
|---|---|---|
| 4352 | 1.164x | 1.081x |
| 5120 | 1.163x | **0.934x** under |
| 6144 | 1.164x | **0.787x** under by 21% |

Flat across seq 2048–6144 on SmolLM2-135M and Qwen2.5-0.5B alike; measured range 1.1621–1.1694. The direction property **holds** here, where it fails there.

(Denominators, since #605 now names both: 0.934x/0.787x are against the real peak; 0.992x/0.830x are against the probe, which runs 12.5–14.3% above the real step. The series above is real-peak throughout.)

## 2. What the gap actually is — measured, not back-solved

`measure_logits_loss_bytes_per_element()` on this stack:

```
12.0, 12.0, 12.0        spread 0.00e+00
vocab 8192 / 32768 / 49152 -> 12.0 exactly
vocab 151936            -> 12.0101   (allocator rounding)
```

`LOGITS_LOSS_BYTES_PER_ELEMENT`'s invariance claim — *"byte-identical under torch 2.5.1 and 2.13.0"* — **holds here**. This is a third stack confirming it, not a counter-example.

**The finding is that the `+2` retained bf16 copy is absent.** The shipped 14 is `12` (loss) + `2` (one further bf16 logits-shaped tensor, retained by something #327 has not identified). If it were retained here, the whole real peak would have to exceed the logits term at 14 alone. On every row it does not:

| model | seq | 14 × E | real peak |
|---|---|---|---|
| SmolLM2-135M | 6144 | 4.228 GB | **3.943 GB** |
| Qwen2.5-0.5B | 6144 | 13.069 GB | **11.848 GB** |

Ten rows, two models, a 3.1x vocabulary contrast. **The retention [#327](https://github.com/MakazhanAlpamys/Soup/issues/327) cannot explain does not happen on this stack** — a constraint on any explanation of it, not merely another point beside it. #327's series now reads: 14 (RTX fit) / 12.311 (H100 fit) / 12.0955 (real trl step) / 12.000000 (this stack, measured) / 11.32 (CCE marginal).

The residual is a flat **1.152x, pop-stdev 0.0249** over-estimate of the modelled non-logits terms. So the 1.16x decomposes as the absent copy (the larger, vocab-scaled share) plus that ~15%.

## 3. The SDPA hypothesis is not supported here

#395 names it as the leading candidate. Forcing each backend, extra allocation over baseline:

| seq | flash | mem-efficient | math |
|---|---|---|---|
| 2048 | 2.4 MB | 2.4 MB | 383.9 MB |
| 6144 | 7.3 MB | 7.1 MB | 3267.4 MB |

The math path is exactly as expensive as the hypothesis needs — 3.27 GB at seq 6144 would sink a 4 GB card alone — and is the positive control. **But SDPA does not select it here**: the default allocates 2.4–7.3 MB across the same range.

Explicitly **not** a refutation for the RTX 3050, which ran Windows/WDDM on torch 2.5.1 where flash-attention shape coverage was narrower. That needs the hardware.

## What this PR changes

- `SECOND_STACK_VRAM_GRID` — ten rows at seq 2048–6144 on two models, kept **separate** from `MEASURED_VRAM_GRID` whose <1% accuracy claim is true on the stack it was fitted to.
- `test_second_stack_has_no_retained_logits_copy` — asserts `14 × E > peak` per row, so a row that stopped supporting the finding fails by name.
- `test_the_non_logits_over_estimate_is_flat` — holds the **measured 12** fixed and pins the residual at 1.10–1.20 with spread < 0.10. Writing the withdrawn 11.83 back into `LOGITS_LOSS_BYTES_PER_ELEMENT` fails this test, so the retraction is enforced in CI rather than only in prose.
- The record, including the negative result, the discarded first sweep, and the retracted four-buffer decomposition (retracted by its own source: *"the peak holds THREE fp32 logits-shaped buffers … the totals agree, the decomposition did not"*).
- Index rows in both `benchmarks/README.md` tables, and the A10G named as a third box.

## On criterion 3

#395 asks to remove the `seq<=512` scope assertion. **I kept it, and that is forced rather than preferred.** Criteria 2 and 3 are jointly unsatisfiable on the RTX-fitted grid: `test_never_under_predicts` is parametrized over it, and that stack demonstrably under-predicts at seq >= 5120. Separating the grids is the only construction that satisfies both.

## Worth surfacing

`pyproject.toml [train]` now requires `transformers>=5.16.1`, `trl>=0.29.0`, `peft>=0.20.0`. **This A10G stack is exactly that floor; the RTX 3050 stack that produced the shipped 14, the `MEASURED_VRAM_GRID` sub-1% claim and the under-prediction finding is beneath it.** A pessimistic constant is currently retained on the strength of a stack no user can install.

## Checks

```
tests/test_v07203.py     174 collected, 169 passed, 5 skipped   (22 added by this PR)
pytest tests/            19343 passed, 102 skipped
ruff check src/soup_cli/ scripts/ tests/     clean
zero deleted lines
```

## Method note

The first sweep was wrong and silently so: `data.max_length` **truncates, it does not pad**, so rows that tokenise short run at their own length while the prediction keeps climbing — a clean-looking over-prediction that is entirely an artifact. The corrected harness overshoots 3x and reads the realised length off `input_ids.shape[-1]` into its own column. Kept in the record because the failure is silent and would mislead the next person identically.


MakazhanAlpamys:
@ARAVIND281 — the measurement is sound and the evidence discipline is above this folder's average. Three items before merge, one of which appeared while I was reviewing.

## What is good, specifically

- **The silently-wrong first sweep is kept, with its mechanism and its control.** `data.max_length` truncates rather than pads, which is the third time that trap has bitten this project. Publishing the discarded run instead of a cleaned-up result is exactly what `benchmarks/` is for.
- **The SDPA hypothesis is published as a negative result**, with a positive control, and explicitly not extended to the RTX 3050.
- Inferred-versus-measured is labelled in both the record and the PR body.
- You engage the 3050 record rather than superseding it, and turn the divergence into an argument *for* `stream_vram_probe`.
- The 21 new tests are pure arithmetic over a hardcoded grid and run in **every** CI cell — not CUDA-gated. Every predicted figure reproduces the shipped formula, including `pstdev = 0.04557` against your claimed 0.046. Zero test deletions, ruff clean.

## Blocking

**1. It no longer merges.** PR #605 landed while this was in review (`0a3920c`, now on main) and rewrote the exact `test_never_under_predicts` docstring you append to:

```
$ git merge-tree --write-tree origin/main <your head>
CONFLICT (content): Merge conflict in tests/test_v07203.py
```

`gh pr view` still says `MERGEABLE` — that is a stale cache. Please rebase. Your comparison figures (1.081 / 0.934 / 0.787) are the **real-run** series and are the correct denominator after #605, so the numbers survive; the prose around them may not, because #605 now names both denominators explicitly.

**2. No `benchmarks/README.md` index row.** CLAUDE.md step 13a is explicit that the file alone is not discoverable, and all five prior benchmark commits complied. Row goes after line 27. The "Hardware" prose at lines 50-60 enumerates the boxes and needs the A10G named as a third.

**3. The headline 11.83 B/element is derived, not measured — and the A10G shipped the instrument to settle it.** `LOGITS_LOSS_BYTES_PER_ELEMENT = 12` at `layer_stream.py:120` carries a documented invariance claim:

> Measured 12.000000 exactly, spread 0.00e+00 over 3 repeats [...] byte-identical under torch 2.5.1 and 2.13.0 and under trl 0.19.1 and 0.26.2 — this term is not stack-dependent.

**The A10G runs torch 2.13.0, i.e. inside that claim.** Your 11.83 is back-solved by subtracting the formula's own modelled non-logits terms, and a consistent ~12-16% over-estimate of those terms reconciles both readings exactly — the data as published cannot distinguish the two. But `measure_logits_loss_bytes_per_element()` and the live `TestTheStackIsRemeasuredWhereverTheSuiteRuns` class exist for precisely this, and were not run on the box that could run them. One number converts the record's central inference into a measurement. That is the difference between a benchmark record and an argument.

## Should-fix

**4. Not reproducible by a third party.** No command, no config, no Soup commit SHA, and neither harness committed to `benchmarks/harness/` — the folder that exists for this, with harness loss already tracked as #379.

**5. The four-buffer decomposition (lines 76-82) was retracted by its own source.** The record you are building on says: *"the peak holds THREE fp32 logits-shaped buffers, not 'upcast + softmax + grad' as v0.72.3 recorded — the totals agree, the decomposition did not."* Re-base that paragraph or drop it.

**6. Cross-link #327.** It is open, you restate its title as a finding, and 11.83 is a fourth point on its series (14 RTX-fit / 12.311 H100-fit / 11.32 CCE-marginal).

**7. You undersell criterion 3.** #395's criteria 2 and 3 are **jointly unsatisfiable** on the RTX-fitted grid, because `test_never_under_predicts` is parametrized over it and that stack demonstrably under-predicts at seq >= 5120. Keeping the guard and separating the grids is the only construction that satisfies both. "Happy to remove it if you read that differently" is too soft for a result that is forced.

## Worth surfacing, and neither the PR nor the record notices

`pyproject.toml [train]` now requires `transformers>=5.16.1`, `trl>=0.29.0`, `peft>=0.20.0`. **Your A10G stack is exactly that floor. The RTX 3050 stack that produced the shipped `14` constant, the `MEASURED_VRAM_GRID` sub-1% claim and the under-prediction finding is below the installable floor.** That reframes both #327 and #395: a pessimistic constant is currently retained on the strength of a stack no user can install. I would rather that observation lived in your record than in my head — add a sentence if you agree with it.

Rebase, add the index row, and run the logits measurement on the A10G if you still have it; 4-7 are yours to take or decline with a reason.

ARAVIND281:
All three blocking items done, plus 4-7. Force-pushed `ee2bbcc` (rebase).

## 3 — you were right, and the correction is bigger than an annotation

I ran the instrument. A10G, torch 2.13.0+cu130, trl 0.29.1, transformers 5.16.1:

```
measure_logits_loss_bytes_per_element(), 3 repeats:
  12.0  12.0  12.0     spread 0.00e+00
vocab   8192: 12.0
vocab  32768: 12.0
vocab  49152: 12.0
vocab 151936: 12.01010951979781   (allocator rounding)
```

**`LOGITS_LOSS_BYTES_PER_ELEMENT`'s invariance claim holds here.** This stack agrees with it; it is not a counter-example. **11.83 is withdrawn** — it was back-solved and attributed the whole discrepancy to the logits multiplier. Your predicted reconciliation was right: the non-logits over-estimate measures **1.152x, stdev 0.0249**, inside your 12-16%.

**What replaces it has no free parameters.** If the `+2` retained copy were present here, the whole real peak would have to exceed the logits term at 14 alone. On all eight rows it does not:

| model | seq | 14 x E | real peak |
|---|---|---|---|
| SmolLM2-135M | 6144 | 4.228 GB | **3.943 GB** |
| Qwen2.5-0.5B | 6144 | 13.069 GB | **11.848 GB** |

Ten rows, two models, 3.1x vocab contrast. **The retention #327 cannot explain does not happen on this stack** — a constraint on any explanation of it, rather than a fourth data point beside it. That is a better result than the one I published, and it only exists because you blocked on the measurement.

## 1 — rebased

The conflict was informative. #605 named both denominators, which resolves a discrepancy I had noticed between the issue body and the docstring and could not explain: 0.934x/0.787x is real-peak, 0.992x/0.830x is against the probe at +12.5-14.3%. My series is real-peak, so the figures survive and the prose is rebased onto yours.

## 2 — index row + hardware prose

Row after line 27; A10G named as the third box. I included your `pyproject` observation, because I agree with it and it belongs on the record: `[train]` floors at `transformers>=5.16.1` / `trl>=0.29.0`, so this stack **is** the floor while the RTX 3050 that produced the shipped constant sits beneath it.

## 4-7

- **4** — `benchmarks/harness/issue395_second_stack_vram.py`, three modes (sweep / `--measure` / `--sdpa`), full environment pinned, and the `max_length` truncation trap plus its two guards in the module docstring.
- **5** — corrected, not deleted. Its own source retracted it: *"the peak holds THREE fp32 logits-shaped buffers … the totals agree, the decomposition did not."* The retraction is the interesting part.
- **6** — #327 cross-linked, 12 placed on its series.
- **7** — taken. Restated as forced: criteria 2 and 3 are jointly unsatisfiable on the RTX-fitted grid, so separating the grids is the only construction that satisfies both. "Happy to remove it if you read that differently" is gone.

`pytest tests/` 19343 passed, 102 skipped. ruff clean. `gh pr view` reports `mergeable: UNKNOWN` right after the force-push — likely the same stale cache you hit; `git merge-tree` against current main is clean on my side.

MakazhanAlpamys:
@ARAVIND281 — re-reviewed `ee2bbcc`. **All three blocking items are resolved, and item 3 you resolved better than I asked.** Two small things and I merge; neither needs a re-measurement.

## The three blockers

**1 — merges cleanly.** `git merge-tree --write-tree origin/main ee2bbcc0` exits 0 with no conflict markers, and `0a3920c` (#605) is an ancestor. My worry was that the prose would not survive #605 naming both denominators; it was handled explicitly rather than by luck — the appended docstring ends with "a flat 1.16x over-prediction (`SECOND_STACK_VRAM_GRID`), **same denominator as the 0.934x/0.787x series**". That clause is the one that had to be right.

**2 — indexed.** Row at `benchmarks/README.md:26`, in the slot between the v0.73.1 and v0.73.2 records, plus the Hardware prose naming the A10G box as a third machine. You also took the `pyproject.toml` observation rather than only acknowledging it: "*this stack **is** that floor while the RTX 3050 stack below is beneath it*". That is the sentence that makes the record useful to a reader who is not you.

**3 — you ran the instrument, and the withdrawal is honest.** `12.0, 12.0, 12.0`, spread `0.00e+00`, across four vocabularies. That makes this stack a **third confirmation** of the invariance claim rather than the counter-example the earlier revision implied. `11.83` survives in exactly four places and every one of them is retraction context — three in the record, one in a test docstring. Kept with its correction beside it, in both the prose and the code, without being asked twice. That is what `benchmarks/` is for.

I recomputed every published table from the committed grid against the shipped formula rather than reading your numbers:

```
pred/real SmolLM2   1.169 1.163 1.164 1.164 1.163 1.164   identical
pred/real Qwen      1.162 1.166 1.167 1.168               identical
14 x E              1.409 ... 13.069 GB                    identical
non-logits mean     1.1516   pop-stdev 0.0249               matches "1.152x / 0.0249"
"14 x E exceeds the whole peak on all rows"                 true, 10/10
```

Ratio range is `1.1621`-`1.1694`, so "flat at 1.16x" is accurate rather than rounded into shape.

```
tests/test_v07203.py                     169 passed, 5 skipped
+ issue327/349/348/cli_startup_is_light  115 passed, 1 skipped
test_issue487_changelog_fragments.py      17 passed
ruff check src/ scripts/ tests/ + benchmarks/   clean
zero deleted lines in the entire PR (git diff origin/main...ee2bbcc0 | grep -c '^-[^-]' -> 0)
```

**Four mutations of the shipped source, four killed by a named test:**

| mutation | killed by |
|---|---|
| `LOGITS_RETENTION_BYTES_PER_ELEMENT` 2 -> 0 | `test_second_stack_has_no_retained_logits_copy` |
| `LOGITS_LOSS_BYTES_PER_ELEMENT` 12 -> **11.83** | `test_the_non_logits_over_estimate_is_flat` (mean -> 0.992) |
| zero `STREAM_FIXED_SLACK_BYTES` + `estimate_activation_bytes` | same test (mean -> 0.478) |
| `LOGITS_LOSS_BYTES_PER_ELEMENT` 12 -> 9 | all 10 params of `test_never_under_predicts_on_the_second_stack` |

The second one is the pleasing result: **writing the withdrawn number back into the source is now a test failure.** That is a stronger outcome than a retraction in prose, because prose cannot fail CI.

One honest note on the third mutation: zeroing the *entire* non-logits budget is not caught by the 10 direction params. That is inherent to your finding rather than a weak test — at 14 B/element the logits term alone already exceeds the peak, which is exactly what the record argues — and `..._is_flat` is the discriminating guard. The direction test is not vacuous; mutation 4 kills all ten.

## Two things before merge

**A — the harness has no row in the `## Harnesses` table.** Same discoverability argument as blocker 2, one table over:

```
$ git show ee2bbcc0:benchmarks/README.md | grep -c issue395_second_stack_vram
0
```

`issue331_qlora_scope.py` and `bnb_repro.py` both have rows; yours is the third script in that folder and the convention is two commits old. One line, same four columns.

**B — the PR description now contradicts its own diff, and it is the only place `11.83` still stands unretracted.** The correction lives in your follow-up comment; the body still leads with "**11.83 ± 0.046** ... the gap is one term, and it can be named", still attributes 14 to the four-buffer decomposition your own record retracted, still describes the new test as one that "fails if the implied B/element spreads" (the shipped test holds 12 fixed and measures the non-logits residual), still offers to remove item 7 which the code now correctly states as forced, and reports "21 passed" where 22 collect.

That matters more than usual here: a PR body is what gets quoted into release notes and read first by anyone arriving from the DOI. **A rewrite of the body would republish the finding as it actually stands rather than as it first appeared.** No re-measurement — the record and the tests are already correct, it is only the description that is behind them.

## Optional, and yours to decline

`layer_stream.py:117` still scopes the invariance claim to "torch 2.5.1 and 2.13.0 ... trl 0.19.1 and 0.26.2". Your A10G run extends it to trl 0.29.1 / transformers 5.16.1. You deliberately kept this PR `src/`-free and that was a good instinct, so I am not asking — noting it so it is not lost.

No changelog fragment needed: no behaviour and no `src/` change, and both prior benchmarks+harness commits (`25d17c0`, `9ce6fa2`) shipped without one.

Push A, edit B, and I merge on green. This is the best-evidenced benchmark contribution the folder has taken from outside.

ARAVIND281:
Both done.

**A — harness indexed.** Row added to the `## Harnesses` table (`3776ac4`), same four columns as `issue331_qlora_scope.py` and `bnb_repro.py`. You were right that it was in neither table; I added the records row and stopped there.

**B — body rewritten.** It now leads with the measured 12 and the absent `+2` copy rather than the withdrawn 11.83, drops the retracted four-buffer decomposition, describes the tests as they actually work (`..._is_flat` holds the measured 12 fixed and pins the non-logits residual — not "fails if the implied B/element spreads"), states criterion 3 as forced with the offer to remove it gone, and reports **22** added rather than 21. I verified that count rather than taking it: `--collect-only` gives 22 in the second-stack class, so your number was right and mine was stale.

There is a short note at the top marking the 11.83 as withdrawn, so the body reads as a correction rather than as if it had always said this — same reasoning as keeping the retraction context in the record.

Your point about the body being what gets quoted into release notes and read first from the DOI is the part I had not weighed. I had treated the follow-up comment as sufficient because the record and tests were already right; a comment is not what a reader arrives at.

One thing worth recording from your mutation table, since it is a better outcome than I aimed for: **writing 11.83 back into `LOGITS_LOSS_BYTES_PER_ELEMENT` is now a CI failure** via `test_the_non_logits_over_estimate_is_flat` (mean drops to 0.992). I wrote that test to pin the residual, not to guard the retraction — that it does both is your framing, not mine, and it is the stronger reading.

Also noted: zeroing the entire non-logits budget escapes the ten direction params. Agreed it is inherent — at 14 B/element the logits term already exceeds the peak, which is the record's own argument — and `..._is_flat` is the discriminating guard there.

MakazhanAlpamys:
Merged as `04d438a`. Thank you — this is the most useful benchmark contribution the project has had, and the reason is not the numbers.

Every one of your three results is a **constraint on an open question**, not another data point beside it. That is a different and much harder thing to deliver:

- The under-prediction does not reproduce, flat 1.1621-1.1694 across seq 2048-6144, where the RTX series reads 0.787x at 6144. So the failure is stack-dependent, not a property of the formula — which is exactly what #395 could not determine from one card.
- `LOGITS_LOSS_BYTES_PER_ELEMENT` measures **12.000000 exactly**, invariant across vocab 8192/32768/49152 and across torch versions. Since `14 x E` alone exceeds the whole real peak on every row, the `+2` retained copy that #327 has never identified **demonstrably does not happen here**. A negative result that rules something out is worth more than a positive one that adds a third number to a table.
- The SDPA hypothesis is not supported, and you included the positive control that makes that statement mean something: the math backend costs 3.27 GB at seq 6144, enough to sink a 4 GB card on its own, so the measurement could have found it if SDPA had selected it. Without that control, "we measured 2.4 MB" is indistinguishable from a broken probe. And you were explicit that this is not a refutation for the RTX 3050, which ran Windows/WDDM on torch 2.5.1 where flash-attention shape coverage was narrower.

Two things I want to point other contributors at.

**You retracted your own headline figure and then enforced the retraction in CI.** The back-solved 11.83 B/element was in the PR body, and rather than quietly editing it away you rewrote the body to say what was withdrawn and why, corrected the record, and added `test_the_non_logits_over_estimate_is_flat` so that writing 11.83 back into the constant fails by name. A retraction in prose decays; a retraction with a test does not.

**The discarded first sweep stayed in the record.** `data.max_length` truncates rather than pads, so short rows run at their own length while the prediction keeps climbing — a clean-looking over-prediction that is entirely an artifact. That failure is silent and would mislead the next person identically, which is precisely why deleting it would have been the wrong call.

Keeping `SECOND_STACK_VRAM_GRID` separate from `MEASURED_VRAM_GRID` is right and I want the reasoning preserved: criteria 2 and 3 of the issue are jointly unsatisfiable on one grid, since `test_never_under_predicts` is parametrized over the RTX grid and that stack does under-predict. Two grids is the only construction that satisfies both, and you argued it rather than silently doing one and dropping the other.

Your point about the dependency floor is the uncomfortable one and it is correct: `pyproject.toml [train]` now floors at transformers>=5.16.1 / trl>=0.29.0 / peft>=0.20.0, this A10G stack IS that floor, and the RTX stack behind the shipped 14 and the sub-1% grid claim is beneath it. A pessimistic constant is currently retained on the strength of a stack no user can install. I am leaving that on #327 rather than acting on it here.

Note I resolved a one-line index-table conflict in `benchmarks/README.md` when merging — #618 landed a row in the same table minutes earlier. Both rows are in.

#395 is closed by this.

Issue #342, opened 2026-08-08
GRPO at 7B dies with grad_norm nan -> 'probability tensor contains inf/nan' in ~40-60% of 200-step runs


MakazhanAlpamys:
Measured on 8xH100 while validating #286. **8 of 14 GRPO runs died** at `grad_norm: nan` followed by a device-side assert `probability tensor contains either inf, nan or element < 0`.

### What it is not
- **Not one bad GPU.** Hit both GPU 0 and GPU 1. The "faulty card" hypothesis was tested and REJECTED: GPU 1 passed a bf16 GEMM check with zero ECC errors, and a retry on GPU 1 ran clean past the step that had just failed there.
- **Not a learning-rate artefact.** Occurred at both 5e-5 and 2e-4.
- **Not predictable.** No pre-crash signature — `grad_norm` ~0.2, no reward-std collapse, nothing in the logs before the step it dies on.
- **Not deterministic.** Not reproducible on retry with an identical seed and config.

### Where it hit
Steps 6, 8, 10, 10, 25, 31, 108, 115 across arms `calB`, `A_log#1`, `A_off`, `A_klLT`, `A_pid`, `A_pidLT`, `B_kl#1`, `B_log`.

### Why it matters beyond the crash
It made #286 unanswerable as scoped. Every adapter from a run that was genuinely reward-hacking was lost to it, so the held-out arm of that issue's acceptance criteria could not be measured at all. `pid_lagrangian` has **no valid arm** for the same reason — all three attempts crashed at steps 6, 25 and 8.

### Setup
Qwen2.5-7B-Instruct, LoRA r32/α64, GRPO, GSM8K 400 prompts, batch 16 × 8 generations, 512-token cap, β 0.02, seed 42. torch 2.13.0+cu130 · transformers 4.57.6 · trl 0.26.2 · peft 0.20.0 · soup 0.72.4 @ `e025836`.

Reproducer: `cfg_A_off.yaml` in `/root/results/issue286_artifacts/` on the validation box, ~40–60% per 200-step run.

### Acceptance
- [ ] the cause is identified (generation sampling? a specific rollout? an optimizer interaction?)
- [ ] either it stops happening, or GRPO fails with a message naming what went wrong instead of a device-side assert
- [ ] a run that hits it does not lose its adapter

Record: `benchmarks/gate-h100-validation.md` STEP 25.

MakazhanAlpamys:
[...]

PR #611, opened 2026-08-31
test(mitigation): reconcile the reward-hack ladders and pin their thresholds (#371)

umran666:
**Part of #371** — fix-path item 2 (complementing @AmirF194's item-3 work in #414). Claimed in [this comment](https://github.com/MakazhanAlpamys/Soup/issues/371#issuecomment-5472636984).

## The reconciliation the gate record asks for

`benchmarks/gate-h100-validation.md` records an open inconsistency: the beta ladder escalated on four consecutive HACK votes while the rollback ladder, requiring three, *never fired in any arm*. The code settles it — **the two ladders live in different modes**:

- The beta ladder that fired is `kl_control`'s bang-bang ladder. `_run_bang_bang` has no HACK streak and no `_escalate` — its ladder ends with β pinned at `beta_ceil`.
- The rollback rung exists **only** in `pid_lagrangian` (`_run_pid`); the schema already rejects `reward_hack_rollback` under any other mode.

So the arm that fired the beta ladder structurally has no rollback counter, and every arm that *has* one (the three `pid_lagrangian` attempts) crashed at #342 before the rung could run. Not two counters over different signals — one counter absent by design. This is documented in the module docstring and both run paths, so the reconciliation lives where the code does. **Comments only; no behaviour change.**

## The remaining CPU-testable acceptance criteria

- **N consecutive HACK votes fire each ladder at its documented threshold, with a control at N-1**: `dwell_steps` for the beta ladder (fires on step N exactly, holds on N-1; a vote of exactly `0.30` is pinned as both a HACK verdict and a trip vote — one threshold, one signal), and the shipped `rollback_patience=3` default for the rollback rung (fires on the exact step the third consecutive HACK vote lands, N-1 never restores, early-stop budget not yet spent).
- **The `beta == floor` degenerate case** — the configuration the defaults produce (`grpo_beta` == `reward_hack_beta_floor` == 0.02): the unseeded sentinel resolves to the floor, sustained release pressure never crosses below it, and the full trip → relax-to-exactly-floor → re-trip cycle runs from that origin.

Item 1 (a completed `pid_lagrangian` run) needs GPU hardware and stays open.

## Test plan

- [x] `pytest tests/test_v07126.py` — 193 passed, 3 skipped (the 3 skips are the pre-existing backend-gated ones)
- [x] `pytest tests/test_issue320_lazy_callback_import.py tests/test_v07127.py` — 186 passed (adjacent modules, no regressions)
- [x] `ruff check src/soup_cli/ tests/` — clean
- All new tests are CPU-only; no torch/transformers needed at module scope

### DIFF
--- a/changelog.d/0.73.3/611.fixed.md
+++ b/changelog.d/0.73.3/611.fixed.md
@@ -1,0 +1,13 @@
+- **The reward-hack ladders' mode split is now documented, and their thresholds
+  are pinned by tests (#371 fix-path item 2, by @umran666 in #611).** The H100
+  gate record carried an open inconsistency — the beta ladder escalated on four
+  consecutive HACK votes while the rollback ladder, requiring three, never
+  fired in any arm. The code settles it: the beta ladder lives in `kl_control`
+  (no rollback rung), the rollback rung lives only in `pid_lagrangian`, and the
+  schema already rejects `reward_hack_rollback` under any other mode — so the
+  arm that fired one structurally lacks the other, and every arm with the other
+  crashed at #342 before it could run. Documented in the module docstring and
+  both run paths; comments only, no behaviour change. Tests pin the remaining
+  CPU-testable acceptance criteria: each ladder fires at its documented
+  threshold on N consecutive HACK votes with a control at N-1, and the
+  `beta == floor` degenerate case the defaults produce.
--- a/src/soup_cli/utils/reward_hack_control.py
+++ b/src/soup_cli/utils/reward_hack_control.py
@@ -9,6 +9,15 @@ only halting:
 - ``pid_lagrangian`` (Stage 2): PID-Lagrangian controller + rollback ladder.
 
 Design:
+- The two mitigation ladders live in DIFFERENT modes by design: kl_control
+  owns the bang-bang β ladder (its terminal rung is β pinned at ``beta_ceil``),
+  and the rollback-to-last-good rung exists ONLY in pid_lagrangian (the config
+  schema rejects ``reward_hack_rollback`` under any other mode). A kl_control
+  run therefore has NO rollback counter. This is the reconciliation the H100
+  gate record (``benchmarks/gate-h100-validation.md``) recorded as an open
+  inconsistency (#371) when the β ladder fired in a kl_control arm but the
+  rollback ladder "never fired in any arm": its only home was the
+  pid_lagrangian arms, which all crashed (#342) before the rung could run.
 - Pure controller pieces (``ControllerState``, ``BangBangPolicy``,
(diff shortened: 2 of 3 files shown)


MakazhanAlpamys:
Merged as `a839063`. Thank you, @umran666 — and the scope discipline here is what made it easy to merge on an issue where it would have been easy to overclaim.

**`Part of #371`, never `Fixes`.** Item 1 deferred to GPU hardware you do not have, item 3 credited to @AmirF194's #414. #371 stays open. On a test-only change touching a controller with three open defects, that distinction is the whole difference between useful and misleading.

**The n-minus-1 controls are why this is worth merging.** A test asserting the ladder "fires eventually" passes for an off-by-one; yours do not:

```
dwell >= policy.dwell_steps            -> - 1     7 failed
  TestLadderThresholds::test_beta_ladder_trips_on_nth_hack_vote_with_nm1_control
_hack_streak >= self.rollback_patience -> - 1     2 failed
  TestLadderThresholds::test_rollback_ladder_fires_on_nth_hack_vote_with_nm1_control
```

I also confirmed you do not overlap #414: removing the `if not held:` guard still fails the pre-existing `test_hold_step_performs_no_coefficient_write`, so that fix stays independently protected.

Your "comments only; no behaviour change" claim on a +22/-2 source diff I verified mechanically rather than by reading — parsed both versions, stripped docstrings, compared ASTs: identical. The -2 is two one-line docstrings becoming multi-line. Every reconciliation claim also checks out against a named line (`schema.py:4111` for the mitigation gate, `schema.py:4013` for `rollback_patience=3`).

One gap for whenever you are next in that file, **pre-existing and not introduced by you**, so not a request:

`TestBetaFloorDegenerate`'s comments claim "the hard invariant: never below", but nothing pins the clamp. Mutating `max(policy.beta_floor, beta / kl_gain)` to drop the `max()` leaves the whole suite green — `test_release_pressure_from_floor_never_crosses_below_floor` never trips, so it never enters the branch containing the clamp, and `test_defaults_seed_at_floor_...` uses `kl_gain=2.0` so relax lands exactly *on* the floor and `max()` is a no-op. Off that lattice (`grpo_beta=0.025`, floor `0.02`, `kl_gain=1.5`) the mutant produces `beta=0.01666667`, below the floor. Either a five-line off-lattice test or a softened comment — this PR is the first thing to claim the invariant, so right now the comment is ahead of the coverage.

Minor: `fixed` is arguably the wrong fragment category for a no-behaviour-change PR (`changed`, or none, as sibling #594 shipped). Not worth a round.

You are in `CONTRIBUTORS.md` as of the commit above, second entry after #605.

PR #612, opened 2026-08-31
fix(doctor): recommend a CUDA wheel that matches the driver

MKnaomi2:
soup doctor hardcoded cu121 when torch was CPU-only. That index is stale on CUDA 12.6+ cards. Parse nvidia-smi CUDA Version and pick the newest compatible PyTorch wheel. Unknown driver falls back to cu130, not cu121. Windows advisory notes that PyPI torch is CPU-only. Reproduced on RTX 5060 Ti (driver CUDA 13.2). Tests in test_v0401_part_c.py. Changelog fragment after PR number is assigned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,8 +362,9 @@ All training tasks run on CPU for testing (quantization auto-disabled). Optional
 soup doctor    # GPU, system resources, dependencies, and version in one place
 ```
 
-- **`ImportError: DLL load failed while importing _C` (Windows)** — reinstall PyTorch for your
-  CUDA version: `pip install torch --index-url https://download.pytorch.org/whl/cu121`.
+- **`ImportError: DLL load failed while importing _C` (Windows).** PyPI's torch
+  wheel is CPU-only. Reinstall a CUDA build; `soup doctor` prints the
+  `pip install` command for the wheel your driver can run.
 - **`soup version` ≠ `pip show soup-cli`** — multiple Python installs; use a virtualenv.
 
 ## Development
--- a/changelog.d/0.73.3/612.fixed.md
+++ b/changelog.d/0.73.3/612.fixed.md
@@ -1,4 +1,6 @@
 - `soup doctor` now recommends a PyTorch CUDA wheel that matches the
   driver's `nvidia-smi` CUDA version instead of always suggesting `cu121`.
-  Unknown drivers fall back to `cu130`. Windows also notes that the PyPI
-  torch wheel is CPU-only. (#612)
+  An unreadable driver header falls back to `cu121` (conservative: a parse
+  failure is treated as an old driver, not a new one). Drivers at CUDA 11.8
+  get `cu118`. Windows also notes that the PyPI torch wheel is CPU-only.
+  (#612 by @MKnaomi2)
--- a/changelog.d/0.73.3/612.fixed.md
+++ b/changelog.d/0.73.3/612.fixed.md
@@ -1,0 +1,4 @@
+- `soup doctor` now recommends a PyTorch CUDA wheel that matches the
+  driver's `nvidia-smi` CUDA version instead of always suggesting `cu121`.
+  Unknown drivers fall back to `cu130`. Windows also notes that the PyPI
+  torch wheel is CPU-only. (#612)
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -256,6 +256,7 @@ _TORCH_CUDA_WHEELS: tuple[tuple[int, int, str], ...] = (
     (12, 6, "cu126"),
     (12, 4, "cu124"),
     (12, 1, "cu121"),
+    (11, 8, "cu118"),
 )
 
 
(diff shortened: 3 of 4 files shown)


MakazhanAlpamys:
Welcome, @MKnaomi2 — I approved CI on your first PR by hand, as this repo requires. The mapping logic is sound and every mutation I ran died. **One blocking item, and it is a class of bug this project has been bitten by four times.**

## Blocking: the new tests fail on any machine with a CUDA torch build

```
pristine PR head, this box (torch 2.5.1+cu121):   2 failed, 16 passed
main, same file:                                 15 passed

E  AssertionError: assert 'cu121' not in '...'
E    rch 2.5.1+cu121). To enable your GPU: `pip install torch --index-url .../whl/cu132`
```

`assert "cu121" not in advisory` matches the **installed torch version string**, not the recommendation. The advisory legitimately contains `2.5.1+cu121` because it reports what you have before telling you what to install.

**CI cannot catch this.** CI installs CPU-only torch, so the substring is absent and the file goes green — it lands, then breaks the local suite for every GPU-equipped contributor and for me. That is the same shape as the Rich/ANSI Linux-versus-Windows trap this repo has hit four times: green on the runner, red on a real machine.

Two-line fix: assert on `"whl/cu121" not in advisory`, i.e. the recommendation, not the report.

## What I checked and where my suspicion was wrong

I expected `cu132` to be a phantom index tag. **It is not** — with a negative control to defeat soft-404s:

| tag | bytes | wheel links |
|---|---|---|
| `cu999` / `cuBOGUS` (control) | 263 | **0** |
| cu132 | 54 234 | 318 |
| cu130 | 94 886 | 486 |
| cu126 | 125 074 | 612 |

Every tag your mapping can emit exists on the index. No broken install command — which is the failure mode that would have mattered most, since wrong install advice is worse than silence.

Mutations, all killed: reversing `_TORCH_CUDA_WHEELS` to oldest-first, deleting the `driver is None` fallback, and making `_parse_cuda_version` always return `None` each fail named tests. Parser handles a real header, a no-space variant, `N/A`, garbage, empty and `None`; `12.10` parses to `(12, 10)` so tuple comparison is correct. Live on this box: driver `(13, 1)` -> `cu130`. `DEPS`/`_MAX_EXCLUSIVE` untouched and `test_doctor_training_bounds_match_declared_extra` still passes.

## Requested, non-blocking

1. **CWE-427.** `_nvidia_smi_cuda_version` calls `shutil.which("nvidia-smi")`, then **discards the absolute path** and passes bare `["nvidia-smi"]` to `subprocess.run`. This repo's policy is at `utils/layer_stream.py:396` — bare names are "deliberately never handed to `subprocess`", because Windows `CreateProcess` searches the current directory before `PATH`. The pre-existing `_detect_gpu_hw_without_torch_cuda` does the same, so you are doubling an existing exposure rather than creating one, but it is a one-line fix while you are here. Everything else about the call is right: list args, no shell, `timeout=5`, returncode checked, `OSError`/`TimeoutExpired` caught.
2. **`(11, 8) -> cu121` is wrong, and the PR now test-pins it.** An 11.8 driver cannot run cu121 wheels. `cu118` exists on the index but has no row in your table. Not a regression — main hardcoded cu121 for everyone — but this PR asserts the wrong answer as intended behaviour, which is worse than inheriting it silently.
3. **Unknown driver -> cu130 is the aggressive direction.** That branch is reached only when `nvidia-smi` ran but its header did not parse, i.e. a very old driver — the user least able to run cu130, which needs >= 580 against cu121's >= 530. Worth a conscious decision either way; say which you intend.
4. **`README.md:366` still says `cu121`** — the identical stale advice, untouched. Two copies of one fact and only one updated is exactly the #372 pattern this project keeps re-learning.
5. **Your fragment does not credit you.** `612.fixed.md` is otherwise correct; house style is `(#612 by @MKnaomi2)`. That matters more than usual on a first contribution.

Fix item 1 in the list above — the failing tests — and 1-5 as you see fit; I will re-run and merge.

MKnaomi2:
Thank you. The `cu121` assertion was matching the installed build, not the index URL. Fixed, and pinned so CI can catch it.

**Blocking.** `assert "cu121" not in advisory` is now `assert "whl/cu121" not in advisory`. `test_detect_gpu_hw_uses_driver_cuda_wheel` injects `torch 2.5.1+cu121` as the installed version (the string on your box) and still requires `whl/cu132` in the recommendation. Restoring the old assertion fails that test on a CPU-only runner.

1. **CWE-427.** Both `nvidia-smi` call sites in `doctor.py` now pass the absolute path `shutil.which` returned. Bare `"nvidia-smi"` is never argv[0]. A named test dies if that is reversed. Same policy as `utils/layer_stream.py` `_resolve_tool`; I did not import that helper.
2. **`(11, 8) -> cu118`.** Row added. The previous test that pinned `cu121` for 11.8 is gone. Floor for a known driver older than the table is `cu118` (oldest published tag), not `cu121`. Confirmed `cu118` is a real index against a `cuBOGUS` control before pushing: `https://download.pytorch.org/whl/cu118/torch/` is 200, 77 378 bytes, 522 `.whl` links; `cuBOGUS/torch/` and `cu999/torch/` are both 403.
3. **Unknown driver -> `cu121`, not `cu130`.** `None` means the header did not parse, which I am treating as an old driver, not a new one. A too-new wheel is the failure mode that looks like success (`pip` ok, CUDA init dies). A known 13.2 header still maps to `cu132`; Blackwell with a working `nvidia-smi` is not this branch.
4. **README.** The Troubleshooting bullet no longer hardcodes `whl/cu121`. `soup doctor` is the single source of the install command, the `#372` shape. A test fails if a `download.pytorch.org/whl/cu*` URL returns to README. Historical `2.5.1+cu121` strings in `benchmarks/` are box records and were left alone.
5. **Fragment** is now `(#612 by @MKnaomi2)`.

Ready for you to re-run on the CUDA box.

MakazhanAlpamys:
Merged as `42be63e`. Welcome, @MKnaomi2 — all five items in one push, and the blocking one is worth naming because of what it teaches rather than what it broke.

**The trap you hit and then pinned.** `assert "cu121" not in advisory` matched the *installed torch version string*, not the recommendation — so the tests passed in CI, where torch is CPU-only and the substring is absent, and failed on every machine with a CUDA build. On my box (torch 2.5.1+cu121):

```
first revision:  2 failed, 16 passed
merged version:  20 passed
```

That is the same green-on-the-runner, red-on-a-real-machine shape that has reddened this repo four times with Rich/ANSI. What I want to credit is not the fix but that you added a test pinning the trap itself, so the naive form cannot come back — that is the difference between fixing an instance and closing a class.

**Three mutations, all killed**, run on the merged head:

```
reverse the wheel table to oldest-first
  FAILED test_torch_cuda_wheel_tag_picks_newest_supported
  FAILED test_detect_gpu_hw_uses_driver_cuda_wheel
unknown-driver fallback cu121 -> cu130
  FAILED test_detect_gpu_hw_returns_advisory_when_smi_succeeds
  FAILED test_torch_cuda_wheel_tag_picks_newest_supported
_parse_cuda_version always returns None
  FAILED test_parse_cuda_version_from_nvidia_smi_header
```

**The fallback reasoning is the best thing in the PR.** Unknown driver falling back to `cu121` rather than the newest tag, on the grounds that a parse failure most likely means an old driver and a too-new wheel is the failure mode that *looks* like success — pip installs happily, CUDA initialisation dies later — is exactly the right way round, and you wrote the argument into the docstring rather than leaving it implicit. Same for `(11, 8) -> cu118`: the first revision asserted `cu121` there, which an 11.8 driver cannot run, so the test was pinning a wrong answer as intended behaviour. Fixing that was not on my blocking list and you did it anyway.

I also checked the thing I was most suspicious of, and **I was wrong**: I expected `cu132` to be a phantom index tag. With a negative control to defeat soft-404s, `cu999` and `cuBOGUS` return 263 bytes and zero wheel links while `cu118` through `cu132` return hundreds. Every tag your mapping can emit exists. There is no install command this can print that does not resolve — which was the failure that would have mattered most, since wrong install advice is worse than silence.

The remaining four all landed: `nvidia-smi` now resolves to an absolute path through `shutil.which` before reaching `subprocess`, with CWE-427 named in the docstring; `README.md`'s stale `cu121` line is updated in the same change, so there are no longer two copies of one fact with only one corrected; and the fragment credits you as `(#612 by @MKnaomi2)`.

`DEPS` and `_MAX_EXCLUSIVE` untouched, `test_doctor_training_bounds_match_declared_extra` still passes, all 13 checks green.

You are in `CONTRIBUTORS.md` as of `a1b2f3b`. If you want another, #273 is a `good first issue` with a live, measured finding sitting in it: `utils/log_level.py`'s `74->73` branch reports 100% coverage on some machines and 98% on others because it depends on whether a foreign logging handler is attached — so the fix is a test that pins it deterministically. Read the thread first; the last two comments have the measurements.

Issue #580, opened 2026-08-27
Feature proposal: resumable offline AutoDistill pipeline with adaptive data generation


Amix29:
## Problem

Soup already has many of the individual building blocks needed for a strong local distillation workflow:

- live token/sequence distillation,
- MiniLLM and ULD strategies,
- Magpie generation,
- Evol-Instruct,
- Best-of-N,
- deterministic/verifiable rewards,
- SFT, DPO, PRM and GRPO/RLVR,
- curriculum, replay and evaluation/ship gates.

However, the current live distillation path requires the teacher and student to be loaded during the same run. This is a hard limitation on memory-constrained hardware, including Apple Silicon unified-memory systems and small CUDA hosts. LoRA reduces trainable and optimizer state, but it does not remove the frozen student base weights from memory.

Soup also has no single workflow that:

1. discovers the student's weak domains,
2. asks a local teacher to generate targeted training material,
3. verifies and ranks the resulting candidates,
4. captures reusable teacher distributions,
5. unloads the teacher completely,
6. trains the student from those cached signals,
7. evaluates the gain on frozen private tests,
8. reallocates the next data budget to the remaining weaknesses, and
9. stops when gains plateau or a safety/general-capability regression appears.

This proposal follows the memory constraint discussed in #472, but generalizes it into a reusable, backend-independent feature.

## Proposed solution

Add a future, resumable **offline adaptive distillation pipeline**. A possible working name is `task: autodistill`, but the CLI/config name is deliberately open for maintainer feedback.

The important invariant is:

> The teacher-only capture process must finish and release all model memory before the student-only training process starts.

The feature should be implemented as several bounded milestones/PRs, not as one large patch.

### 1. Frozen baseline and capability map

Before generating or training anything:

- evaluate the original student on an immutable suite;
- support public benchmarks plus user-provided private JSONL suites;
- record per-domain scores, confidence/noise information, exact model revisions, tokenizer hashes, generation parameters and seeds;
- compute a capability map such as coding, mathematics, science, reasoning, instruction following, tool use and general knowledge;
- keep private prompts and expected answers out of teacher-generation inputs;
- fail closed if the baseline cannot be reproduced.

The output should be machine-readable so a human or an external orchestrator such as Codex can inspect every decision without being the source of truth.

### 2. Domain-conditioned synthetic generation

Extend Soup's existing Magpie generator rather than introducing a hard dependency on another framework.

Proposed capabilities:

[...]

Issue #382, opened 2026-08-11
CI: windows-latest cells die with 0xc000001d in the NF4 training-step test — three occurrences, now on main


MakazhanAlpamys:
## Context

`tests/test_v07202.py::TestNF4EndToEndSetup::test_one_nf4_training_step_actually_runs` runs a real training step through bitsandbytes' 4-bit kernels. On some `windows-latest` runners the wheel executes an instruction the CPU does not have, and the process dies with `Windows fatal exception: code 0xc000001d` (ILLEGAL_INSTRUCTION) — a faulthandler stack dump with no Python exception.

This is already documented in `.claude/CLAUDE.md` (release checklist step 16) as "a runner/wheel problem, not a code regression — re-run before diagnosing". That guidance is correct. What has changed is the frequency, and that it now lands on `main`.

## Evidence

| when | commit | cell | outcome |
|---|---|---|---|
| 2026-08-05 | `9117da1` (two lines of markdown) | windows-latest / **3.12** | crash; `gh run rerun --failed` passed on the same tree |
| 2026-08-11 | PR #362 [`31466333869`](https://github.com/MakazhanAlpamys/Soup/actions/runs/31466333869) | windows-latest / **3.10** | crash; rerun passed, 11/11 |
| 2026-08-11 | `7c9a931` (the #362 merge) [`31468811428`](https://github.com/MakazhanAlpamys/Soup/actions/runs/31468811428) | windows-latest / **3.10** | crash |

Three occurrences, two Python versions, so it tracks the **runner CPU**, not the interpreter. GitHub's `windows-latest` fleet is heterogeneous and a share of it lacks whatever instruction the bitsandbytes wheel emits.

The strongest control is free: `1715261` is `7c9a931` **plus one line in `CONTRIBUTORS.md`** and its run came back green on the same pool. So the failing tree and the passing tree differ by a docs line.

Both crashes land at the same point — the last line before the dump is always `test_setup_builds_a_real_trl_trainer_under_nf4 PASSED`, i.e. the process dies inside the very next test, which is the first one to call `trainer.train()`.

## Why it is worth acting on rather than re-running

An illegal instruction is not a test failure — it kills the interpreter, so the cell reports **nothing** about the other ~17,000 tests it had not reached yet (the crash lands at 97%, but that is luck, not design). A red `main` that means "one binary wheel met the wrong CPU" also trains the reader to skip past red, which is the expensive part.

[...]

PR #613, opened 2026-08-31
feat: define AutoDistill Milestone A artifact contract

Amix29:
## Summary

- define frozen, versioned AutoDistill v1 plan, capture-token, shard-manifest, and consumption-event artifacts
- bind immutable model, tokenizer, data, policy, backend, dtype, quantization, and throughput fingerprints
- define top-k union forced-token residual-tail semantics without a default compression or quality claim
- add transactional shard, resume, corruption, reservation, commit, and replay state machines
- add exact model-free dense/sparse size arithmetic and cached-profile-only runtime ranges
- document the threat model, evaluation firewall, later adaptive curriculum, and milestone boundaries
- add tiny deterministic JSON fixtures and model-free contract tests that construct canonical JSONL payload bytes

## Scope boundaries

This is only Milestone A of #580. It adds no task or SoupConfig fields, no CLI, no teacher or student loading, no capture/training loop, no adaptive allocator, no cross-tokenizer ULD, and no benchmark integration.

## Validation

- `ruff check --no-cache src/soup_cli/ scripts/ tests/`
- 30 local AutoDistill contract tests passed
- 75 local AutoDistill and adjacent resume tests passed
- explicit CRLF-checkout simulation reproduced the committed 423-byte payload and SHA-256
- the GitHub Actions matrix remains the cross-platform gate

Implements the reviewable Milestone A design-and-contract slice requested in #580.


MakazhanAlpamys:
@Amix29 — this is the strongest contract slice I have reviewed here, and the scope matches what we agreed on #580 item for item. Two blocking things, both cheap, neither architectural.

## Scope is exactly Milestone A, verified rather than trusted

```
$ git diff origin/main...HEAD --diff-filter=M --stat
 docs/README.md | 1 +
```

The only modified existing file is a one-row docs index entry. `config/`, `cli.py` and `commands/` are untouched, and `grep -rn autodistill src/ tests/` outside the new package finds no consumer — which is correct and agreed for this milestone, not an omission. Zero deletions repo-wide.

The refusals in `docs/autodistill.md` §2 are the part I want to single out: no default `top_k`, no default width, no compression method, no quality threshold, and an explicit statement that sparse KD is **not** claimed to reproduce dense KD at `k < vocab_size`. `renormalize_selected: Literal[False]` encodes that refusal in the type system rather than in prose. `coarse_tail_forward_kl` is named coarse instead of being sold as a reconstruction. That is the discipline this project is graded on.

## Blocking 1: the fixture's bytes are rewritten by checkout, and the suite is red on Windows

```
$ git config --get core.autocrlf
true
$ ls .gitattributes
NO .gitattributes

$ pytest tests/test_issue580_autodistill_contract.py
FAILED test_shard_fixture_detects_parseable_payload_corruption
E   ArtifactCorruptionError: payload 'capture.jsonl' byte count mismatch
1 failed, 22 passed

v1/capture.jsonl  CRLF present: True  bytes: 424     manifest expects: 423
```

`core.autocrlf=true` is the Git for Windows installer default and this repo has **no `.gitattributes` anywhere**. Your PR introduces its first byte-exact tracked fixture — `tests/fixtures/` did not exist before — so there is no precedent proving anything normalises it.

This is a design problem in the fixture, not just a CI-config one: a contract whose entire subject is byte-level integrity ships a payload whose bytes are mutated in transit. And while that test is red it cannot detect payload corruption, which is the guard it exists to be.

Fix: `.gitattributes` with `tests/fixtures/autodistill/v1/*.jsonl -text` (or repo-wide `* text=auto eol=lf`), or build the payload bytes inside the test instead of committing them. I would take the second — a fixture that cannot be corrupted by tooling is strictly better for this particular contract.

## Blocking 2: five guards have no discriminating test, including the invariant we agreed on

Nine mutations, replacing each guard's condition with `if False:`. **Four killed, five survived.** Every survivor produced `1 failed, 22 passed` — the one failure being the CRLF baseline above, not the mutation.

| guard removed | result |
|---|---|
| tampered plan estimate | KILLED |
| `student_rollout` replay forbidden | KILLED |
| `decide_resume` quarantine on bad evidence | KILLED |
| `tail_mass` positive when k < vocab *(control)* | KILLED |
| **`student_rollout` branch of `CaptureToken`** | **SURVIVED** |
| **probability mass `selected + tail == 1`** | **SURVIVED** |
| **`../` path-traversal rejection** | **SURVIVED** |
| **`selected_token_ids == top_k ∪ forced`** | **SURVIVED** |
| **ledger `replay_of` chaining + contiguity** | **SURVIVED** |

I re-ran the first survivor myself, because it is the one that matters most:

```python
-        if self.trajectory_kind == "student_rollout":
+        if False:
-> 1 failed, 22 passed        # identical to baseline
```

**That is the teacher-only/student-only boundary I named on #580 as the right invariant.** Every fixture and every test is `teacher_expert`; the `student_rollout` half of the two-view capture contract has zero execution.

**To be clear about what this is and is not:** it is *not* the "validated, documented, read by nothing" pattern from v0.73.3. I probed six guards against unmutated code and every one fires — `../../etc/passwd`, `/abs/x`, `C:/win`, a `student_rollout` row without `student_sampled_token_id`, a mismatched selected set, and a mass that does not sum to one are all rejected. The validators are real and reachable. It is the adjacent defect: guards with no test that fails when you remove them. Coverage corroborates it — the 104 missed statements are almost entirely the `raise` lines.

The pattern in every survivor is the same: the test asserts on a **valid** fixture, so it passes whether or not the contract enforces anything. `test_capture_fixture_preserves_selected_and_tail_probability_mass` checks the fixture's mass sums to 1, which is true either way.

## Two smaller things

- **Changelog attribution.** `613.added.md` cites `(#613)` only; the house form is `(#580 by @Amix29 in #613)` — which you used yourself on `487.*.md`.
- **Two claims the diff does not back.** The PR body says "23 AutoDistill contract tests passed" — 23 exist, 22 pass on Windows, and it does not say which platform. And `docs/autodistill.md` §14 says the tests cover top-k-plus-tail mass and corruption; mutations 5 and 7 show those two guards are undefended and the corruption test is the one currently broken. Assertions exist; enforcement is not covered.

## Clean

No file I/O at all — `verify_payload_bytes` takes bytes from the caller — so the absence of `enforce_under_cwd_and_no_symlink` / `O_NOFOLLOW` is correct rather than a gap at this milestone. AST scan of both new modules shows no heavy top-level imports; `test_cli_startup_is_light` + the changelog-fragment test are 51 passed. `ruff` clean. No `docs/commands.md` entry, correctly, since there is no user-visible command yet.

Pin the fixture, add negative tests for the five survivors — starting with a `student_rollout` `CaptureToken` — and I will re-run the set and merge.

MakazhanAlpamys:
CI has resolved the one caveat I left in that review. **The Windows failure is confirmed, not predicted:**

```
test (windows-latest, 3.10)   fail
test (windows-latest, 3.12)   fail
test (macos-latest,  3.10)    pass
test (macos-latest,  3.12)    pass
test (ubuntu-latest, *)       pending
```

Two independent Python versions failing on Windows while macOS is green on the same commits is the signature of a checkout-layer problem, not a logic error — which is what the CRLF reproduction predicts and nothing else does.

Attribution, since I would rather show the chain than assert it:

- `main` is green on those same cells.
- The annotation is `Process completed with exit code 1` — an ordinary pytest failure, **not** the `Windows fatal exception: 0xc000001d` illegal-instruction signature this repo documents as its recurring Windows flake (#382), which produces a faulthandler dump and kills the interpreter rather than exiting 1. I checked for that specifically, because a wrong attribution here would send you chasing the wrong thing.
- This PR modifies no existing source file — only `docs/README.md`, +1 index row — so it cannot have broken a pre-existing test.
- My local Windows run reproduces exactly one failure, `test_shard_fixture_detects_parseable_payload_corruption`, with the cause proven byte-for-byte: the manifest commits 423 bytes, checkout delivers 424, and the LF-normalised bytes hash back to the committed digest exactly.

**Residual uncertainty, stated plainly:** I have not read the failing test name out of the CI logs — they are not published until the whole run completes. If the pending ubuntu cells come back green, that closes it. If they also fail, the cause is something I did not find and you should re-check before acting on item 1.

Nothing else in the review changes. Item 1 is now a confirmed merge-blocker rather than a predicted one.

Amix29:
Fixed in `b57c287`:

- removed the tracked exact-byte `capture.jsonl`; the shard payload is now rebuilt as canonical JSON bytes plus one explicit LF from a validated capture fixture, so checkout newline conversion cannot change the 423-byte/SHA-256 commitment;
- added a real `student_rollout` capture fixture and discriminating negative tests for its required/forced sampled token;
- added discriminating negative tests for selected-plus-tail mass, `selected == top_k union forced`, `../` traversal, ledger contiguity, and `replay_of` chaining;
- updated the changelog to `(#580 by @Amix29 in #613)` and corrected the docs/PR validation claims.

Local validation: 30 AutoDistill contract tests pass, 75 contract plus adjacent resume tests pass, full Ruff passes, and an explicit CRLF simulation reproduces the committed 423 bytes and SHA-256. CI run 33390345814 is now running on the new commit.

MakazhanAlpamys:
@Amix29 — re-reviewed your push. **Both blockers are resolved, and the remaining red is mine, not yours.**

**Blocker 1 — you took the better option.** The committed `.jsonl` fixtures are gone; the payload bytes are generated in the test. That removes the failure mode rather than working around it, which is the right call for a contract whose subject is byte-level integrity — there is now nothing for `core.autocrlf` to rewrite.

```
local Windows run:  30 passed        (was 1 failed, 22 passed)
CI: windows 3.11 pass · windows 3.12 pass · ubuntu 3.10/3.11/3.12 pass · macos 3.10/3.11/3.12 pass
```

**Blocker 2 — the two guards I cared about most are now defended.** Re-ran the mutations on your head:

```
student_rollout branch -> if False:        1 failed, 29 passed    KILLED
mass conservation      -> if False:        1 failed, 29 passed    KILLED
```

The first is the teacher-only/student-only boundary I named on #580 as the point of the design, and it went from zero execution to a test that fails when you remove it.

**The one red cell is not your PR.** `test (windows-latest, 3.10)` fails with:

```
Windows fatal exception: code 0xc000001d
tests/test_v05311.py::TestGRPOVariantRuntimeContract
```

That is issue #382 — part of GitHub's windows-latest fleet lacks an instruction the bitsandbytes wheel emits, so the first test reaching a real `trainer.train()` kills the interpreter. It hit `main` twice today independently of you. I fixed it on `main` in `2d030c4` by extending the #382 guard to that test; your branch predates that commit, so **a rebase onto current `main` should turn the cell green**. If it does not, tell me and I will look again — I would rather you not chase a failure I introduced the fix for.

Please rebase, and while you are there the two small items from my review still stand: the changelog fragment should read `(#580 by @Amix29 in #613)` rather than `(#613)`, and the PR body's "23 tests passed" and docs §14's coverage claim both need updating now that the numbers and the coverage have changed.

Once the rebase is green I will re-run the full mutation set and merge. This is close.

MakazhanAlpamys:
Merged as `f8a5cf7`. Thank you @Amix29 — this took three rounds and every one of them made it better.

## What I re-verified before merging, on the rebased head rather than the reviewed one

A rebase is a new tree, so the mutation set was re-run against `6a9ab103`, not against the commit I reviewed:

```
baseline                                              30 passed

student_rollout branch      -> if False:              1 failed, 29 passed
  test_student_rollout_fixture_requires_sampled_token_and_forces_it
selected == top_k U forced  -> if False:              1 failed, 29 passed
  test_capture_rejects_selected_ids_outside_top_k_union_forced
mass conservation           -> raise removed          1 failed, 29 passed
  test_capture_rejects_probability_mass_other_than_one
'..' path traversal         -> if False:              1 failed, 29 passed
  test_plan_rejects_parent_path_traversal
student_rollout replay ban  -> raise removed          1 failed, 29 passed
  test_consumption_is_exactly_once_for_student_rollouts

revert                                                30 passed
```

Five guards, five distinctly named tests, and the tree returns to green. Plus:

```
git diff origin/main...6a9ab103 | grep -c '^-[^-]'   0     nothing deleted anywhere
ruff check src/ scripts/ tests/                      clean
tests/test_cli_startup_is_light.py                   34 passed
top-level heavy imports in contract.py               none (AST walk)
CI                                                   13/13
```

## The two things that made this merge rather than iterate again

**You removed the failure mode instead of working around it.** The committed exact-byte `.jsonl` fixtures could have been kept with a `.gitattributes` rule, and that would have gone green. Deleting them and building the payload in the test as canonical JSON plus one explicit LF means there is nothing left for `core.autocrlf` to rewrite — which is the correct answer specifically because the thing under contract is a byte-level commitment. A guard over bytes that depends on how the bytes arrived on disk is not a guard.

**You scoped the boundary yourself, in the PR body, unprompted.** "No task or `SoupConfig` field, no CLI, no teacher or student loading, no capture or training loop, no adaptive allocator, no cross-tokenizer ULD, no benchmark integration." That is why this merged as `Refs #580` and #580 stays open — a Milestone A slice that closed its parent issue would have been the more flattering outcome and the less honest one.

The `student_rollout` branch is worth singling out: it is the teacher-only/student-only boundary the whole design turns on, it had **zero** test execution before this round, and it now fails when removed. That is the difference between a contract and a docstring.

## On the red cell, and the record

It was #382 — a real `trl` GRPOTrainer step killing the interpreter with an illegal instruction on part of GitHub's windows fleet — and it hit `main` independently the same day. It cleared on your rebase over `2d030c4`, which is the control that confirms the attribution rather than just the claim. You lost a cycle to a failure I introduced the fix for, and I would rather say that plainly than let it read as your PR having been flaky.

Your 31st merge. #580's remaining milestones are yours if you want them — and if the next slice is as clearly bounded as this one, it will move faster.

Issue #41, opened 2026-04-21
v0.31.0: recipe CI smoke-train validation for multi-GPU recipes


MakazhanAlpamys:
## Context

v0.27.0 adds 3 multi-GPU recipes:
- `llama3-70b-fsdp2` (requires 8x A100/H100 80GB)
- `qwen3-32b-zeropp` (requires 4+ GPUs with NVLink/NVSwitch)
- `deepseek-v3-pipeline` (requires 4+ GPUs, pipeline exec deferred to v0.27.1)

Today these recipes load cleanly as config (`test_multi_gpu.py::TestMultiGpuRecipes` verifies schema + fields) but are **not exercised by a 100-step CI smoke-train** — upstream tokenizer / model-loader changes could silently break them.

## Proposal for v0.31.0 (per plan roadmap)

From the roadmap:
> **Recipe verification CI** — every recipe runs a 100-step smoke train in CI to catch regressions from upstream model changes.

Scope for multi-GPU specifically:
- Use Modal.com or Lambda Labs free credits for one 4xA100 runner
- Run 100-step smoke train for each multi-GPU recipe, assert loss decreased
- Gate CI on this (fail PR if smoke train regresses)

## Acceptance criteria

- [ ] New workflow `.github/workflows/recipe-smoke-train.yml`
- [ ] Runs on release branches only (not every push — cost)
- [ ] Matrix over the 3 multi-GPU recipes
- [ ] Artifact: final-step loss + wall-clock
- [ ] Fails the release if any recipe's loss trajectory regresses >20% vs last recorded baseline

## Risk notes

- GPU runner cost — need a sponsor or free-credit strategy (Modal startup program, Lambda research credits)
- Smoke train != full train — 100 steps is too short to catch some regressions, but catches tokenizer / loader / quant breaks

MakazhanAlpamys:
v0.31.0 ships the **CPU-only validation tier** of recipe CI: `.github/workflows/recipe-validation.yml` runs the parametrized pytest validation (`tests/test_recipes_v031.py`) on every PR + push touching recipe / config / data code. This catches:

- Schema regressions (Pydantic config rejecting a recipe yaml)
- Catalog/yaml drift (`RecipeMeta.model` ≠ yaml `base:`)
- Unsafe model IDs (`..` / `://` / null bytes / backslashes)
- Empty `lora.target_modules` (silent no-op LoRA)
- `max_length` out of bounds
[...]

PR #618, opened 2026-08-31
fix(benchmarks): repair and publish the FSDP per-rank sharding probe (#373)

Srinivasan8888:
Repairs and publishes the STEP 20 FSDP per-rank sharding probe.

`Refs #373` — **not** `Fixes`. See Scope at the bottom: three of the four
acceptance criteria need eight H100s and stay open.

## The mechanism, and why it changes the fix

The issue body (now corrected) said the probe hit a bare FSDP module. It did not.
That message is not FSDP's — FSDP defines no `__len__` at all:

```
len(bare_fsdp)      -> TypeError: object of type 'FullyShardedDataParallel' has no len()
len(compiled_fsdp)  -> TypeError: FullyShardedDataParallel does not support len()   <- recorded text
```

The recorded text's only source is `torch/_dynamo/eval_frame.py:476`,
`OptimizedModule.__len__`, which reports the name of the **wrapped** class — which
is exactly why it reads as though FSDP raised it. The failing object was a
`torch.compile` wrapper around FSDP, one indirection deeper than the 0.5B arm.

**So "stop calling `len()`" is necessary but not sufficient.** A probe that only
drops the call walks the tree without raising and still mis-attributes: it counts
`OptimizedModule` where the 0.5B record counts `FullyShardedDataParallel`, and
concludes FSDP was never engaged. A silent wrong number is worse than the crash it
replaced, so `unwrap_compiled()` is applied before every class-identity and
accounting question and is **pinned by tests**, per your request, rather than left
as an implementation detail.

## What is in it

- `benchmarks/harness/fsdp_sharding_probe.py` — `unwrap_compiled`,
  `local_parameter_numel`, `module_class_histogram`, `sharding_verdict`,
  `probe_rank`. Pure-Python and duck typed: **no torch import, no CUDA, no process
  group**, so the accounting is unit-testable on CPU. Requirements and protocol are
  stated at the script boundary, matching the `bnb_repro.py` convention.
- `sharding_verdict()` asserts the accounting rather than eyeballing it
  (criterion 2), and treats `world_size == 1` as the control that expects the
  **full** count — so a probe that unconditionally divides fails the control
  instead of passing everything (criterion 3).
- `tests/test_issue373_fsdp_sharding_probe.py` — 14 tests. The doubles reproduce
  `OptimizedModule.__len__` byte-for-byte, and two tests assert the *doubles
  themselves* reproduce the recorded failure, since a double that does not is
  pinning nothing.
- A row in the `## Harnesses` table of `benchmarks/README.md` — the one contact
  point you flagged with #597.

## Mutations — seven run, seven killed

| Mutation | Result |
|---|---|
| walks without `len()` but does **not** unwrap `_orig_mod` | 2 fail |
| `unwrap_compiled()` reduced to identity | 3 fail |
| `sharding_verdict` always returns `is_sharded=True` | 2 fail, incl. the single-GPU control |
| `len(module)` reintroduced into the accounting | 1 fails |
| `world_size >= 1` guard removed | 1 fails |
| histogram reports the wrapper class | 2 fail |
| **`is_sharded` loosened to `local <= expected_local`** | **survived the first pass** |

The last one is a real gap, not an equivalent mutation, and I am reporting it
because the fix for it is a test I did not originally write. An under-count — a
rank whose parameters were never materialised — would have read as "sharded".
Killed by `test_under_count_is_not_sharded`, added for it and named after it:

```
$ # is_sharded=(local <= expected_local)
$ pytest tests/test_issue373_fsdp_sharding_probe.py -q
1 failed, 13 passed
```

## Verification

```
pytest tests/test_issue373_fsdp_sharding_probe.py                    -> 14 passed
pytest ... test_cli_startup_is_light.py test_no_foreign_license_headers.py -> 58 passed
ruff check src/soup_cli/ scripts/ tests/ benchmarks/                  -> All checks passed
git diff -- tests/ | grep -c "^-[^-]"                                -> 0
```

## Scope — what this does not establish

**Criterion 1** (reports on all ranks at 70B), **criterion 2's actual reading**, and
**criterion 4** (runs as part of #41) all need the eight-H100 box. I do not have it,
so this PR makes no claim about whether the 70B base is sharded. What it delivers is
a probe that is correct, unwrapping-aware and mutation-tested, so the next person
with the hardware gets a reading instead of a `TypeError` — and, more to the point,
does not get a plausible wrong number.

`probe_rank()` reads `RANK` / `WORLD_SIZE` from the environment and returns a dict,
so wiring it into the #41 recipe smoke-train is a call site rather than a rewrite
whenever that lands.


Srinivasan8888:
Rebased onto `main` and force-pushed as `ed5ece88` — #597 merging put a `bitexact.py` row in the `## Harnesses` table and this PR added one directly after it, so it went `CONFLICTING` as soon as `4021b9f` landed. Exactly the one-line contact point you predicted on #373, and nothing else touched.

Resolved by keeping both rows, `bitexact.py` first since it is already on main:

```
| issue331_qlora_scope.py  | #331             | ...
| bnb_repro.py             | bitsandbytes #2034 | ...
| bitexact.py              | v0.72.0 / STEP 1 | ...
| fsdp_sharding_probe.py   | STEP 20 / #373   | ...
```

Verified rather than assumed, since GitHub's mergeability was still stale when I checked:

```
git merge-base --is-ancestor origin/main HEAD          -> branch sits on current main
git merge-tree $(git merge-base HEAD origin/main) HEAD origin/main | grep -c '<<<<<<<'
                                                       -> 0
pytest tests/test_issue373_fsdp_sharding_probe.py      -> 14 passed
ruff check tests/ benchmarks/                          -> All checks passed
```

One commit, no content change beyond the table ordering. @Samearth17 — your row is untouched and above mine, as offered.

MakazhanAlpamys:
Merged as `0926173`. Thank you — this is the second PR in a row where you corrected the issue rather than implementing it, and both times the correction was the load-bearing part.

The attribution is what I checked hardest, because "stop calling `len()`" would have been a defensible and wrong fix. I confirmed it independently: `torch.nn.Module` defines no `__len__`, FSDP adds none, and the recorded string `FullyShardedDataParallel does not support len()` has exactly one source in the torch tree — `OptimizedModule.__len__` in `_dynamo/eval_frame.py`, which formats the name of the *wrapped* class. So the failing object was a `torch.compile` wrapper, and a probe that merely stops crashing would have counted `OptimizedModule` and concluded FSDP was never engaged. Trading a `TypeError` for a plausible wrong number is the worse outcome for an instrument whose whole job is to answer one question, and you saw that before I raised it.

What I ran:

- `pytest tests/test_issue373_fsdp_sharding_probe.py` — 14 passed; the full file plus `test_cli_startup_is_light.py` green on the merge head.
- Mutations. Three mandatory ones are killed: reducing `unwrap_compiled()` to identity (3 fail), `sharding_verdict` always returning `is_sharded=True` (2 fail, including the single-GPU control), and reintroducing `len(module)` into the accounting (1 fails).
- Two survivors, both **missing tests over correct code**, not defects, and I am recording them here rather than blocking on them. First: your test double's `OptimizedModule.named_children()` returns the children directly, where real torch returns `[('_orig_mod', <wrapped>)]` — I checked the torch source rather than assuming — so unwrapping at *child* level is not pinned by the suite. Second: `total % world_size` has no test. Neither changes what the merged code does.
- Diff read against `benchmarks/README.md` and the issue text; `git diff -- tests/ | grep -c '^-[^-]'` → 0.

The two design choices I would point other contributors at: the `world_size == 1` control expecting the *full* count, which makes an always-dividing probe fail instead of passing everything, and the two tests that assert the doubles themselves reproduce the recorded failure. A double that cannot produce the bug pins nothing, and almost nobody writes that test.

Your `is_sharded <= ` survivor report is the standard I want in these PRs — you found it, said so, and wrote `test_under_count_is_not_sharded` for it instead of leaving it in the body as a caveat.

Merged as `Refs #373`: three of the four acceptance criteria need the eight-H100 box, so the issue stays open on those, and this makes no claim about whether the 70B base is sharded.

I resolved a small index-table conflict in `benchmarks/README.md` on `#610`'s branch afterwards — your row and its row both belong, and both are in.

MakazhanAlpamys:

I closed the two mutations you reported as survivors, and you were right to report them rather than paper over them.

The first turned out to be a defect in the double rather than a missing test. `OptimizedModule.named_children()` forwarded the WRAPPED module's children, so the wrapper was invisible below the root and the child-level `unwrap_compiled` call in `module_class_histogram` was never exercised — a compiled *submodule* would have been counted as `OptimizedModule`, which is the exact mis-attribution the harness exists to prevent. Real torch yields `[('_orig_mod', wrapped)]`; the double now does too, and a tree with a compiled submodule pins it.

The second was `total % world_size`, which had no test: a total that does not divide evenly would still read as sharded whenever the truncating division happened to match.

```
child-level unwrap removed -> test_a_compiled_child_is_counted_by_its_wrapped_class
`total % world_size` removed -> test_a_total_that_does_not_divide_evenly_is_refused
reverted                     -> 18 passed
```

PR #619, opened 2026-08-31
fix(sglang): obey the trust_remote_code gate instead of hardcoding True (#360)

Srinivasan8888:
The separate PR for the SGLang trust contract, as agreed on #360/#581 — the wider
change I deliberately kept out of #581 so it could be argued on its own terms.

**This is a security fix, not the tidy-up I expected to write.** I went in to make
the SGLang backend's `trust_remote_code` configurable and found the gate was not
merely inconsistent — on that backend it does not apply at all.

## The hole

`serve.py` resolves the flag once for every backend, and says why:

```python
# v0.36.0 Part B: --trust-remote-code default-deny, resolved ONCE for
# every backend. vLLM previously loaded with an unconditional
# trust_remote_code=True (arbitrary repo code, zero notice) — resolve the
# same gate + warning panel the transformers path uses so no backend
# silently executes an untrusted repo's code.
resolved_trust = resolve_trust_remote_code(...)
```

vLLM gets it: `_serve_vllm(..., trust_remote_code=resolved_trust)`.

SGLang was called **without the argument at all**, and `create_sglang_runtime`
hardcoded `trust_remote_code=True` at *both* `sgl.Runtime` call sites — plus the
tokenizer load at a literal `True`.

So `soup serve --backend sglang` executed a model's custom repo code **whether or
not `--trust-remote-code` was passed**. `resolved_trust` was computed and thrown
away. That is precisely the behaviour v0.36.0 removed from the vLLM path, still
live on the other one.

The panel above the load was honest — "SGLang loads models with trust_remote_code
enabled" — but **a notice is not a gate.** It told the user what was about to
happen rather than letting them decide.

## The fix

`create_sglang_runtime(..., trust_remote_code: bool = False)`, matching the vLLM
signature; `serve.py` passes `resolved_trust` into `_serve_sglang`, which threads
it to the runtime **and** the tokenizer. The panel now reports what was resolved
instead of promising trust unconditionally.

## Behaviour change worth calling out

A custom-code model on the SGLang backend now **fails without `--trust-remote-code`**
rather than silently executing. That is the point of the change, and it is the same
thing vLLM has done since v0.36.0 — but it is a real break for anyone relying on
the old implicit behaviour, so it wants the release note rather than a quiet line
in a diff.

## Two existing assertions changed — deliberately, neither weakened

`git diff -- tests/ | grep -c "^-[^-]"` is **5**: four are docstring lines
replaced by a longer one, and one is a real assertion flip. Both are mine from #581
and both are the old contract:

- **`test_sglang_serve.py`** asserted the runtime default was `True`. The default is
  now `False`; the assertion is inverted, with a comment recording that it asserted
  `True` until this fix and why. The opt-in path is covered in the new file.
- **`test_issue360...::test_tokenizer_load_uses_the_trust_setting_the_runtime_uses`**
  pinned a literal `True`. Its subject is that the tokenizer *matches the runtime* —
  a mismatch is exactly what made #360's prompt fix silently degrade on custom-code
  models. It now drives both sides from one opted-in value and still asserts they
  agree, instead of pinning a literal that moved. The guard is intact; had I flipped
  it to `False` and left it there, mutation (c) below would have survived.

## Mutations — five, all killed

| Mutation | Result |
|---|---|
| runtime default flipped back to `True` | 3 fail |
| `serve()` passes a literal `True` to sglang | 1 fails |
| tokenizer pinned back to `True` while the runtime stays gated | 1 fails |
| adapter call site alone reverted to `True` | 1 fails |
| `_serve_sglang` stops forwarding to the runtime | 2 fail |

Each surface is pinned independently — the runtime's plain path, its adapter path,
the tokenizer, and the `serve()` wiring — so a fix that threads one and forgets
another fails rather than passes. That split is the guard you asked for on #615,
applied here from the start.

## Verification

```
pytest tests/ -k "sglang or serve or trust"     -> 392 passed, 19096 deselected
pytest tests/test_sglang_trust_contract.py      -> 10 passed
ruff check src/soup_cli/ scripts/ tests/        -> All checks passed
```

## Scope

`Refs #360` — SGLang still cannot be installed here, so this is CPU-verified
wiring, not a live-runtime check. #360's live acceptance item stays open and this
does not close it.

### DIFF
--- a/changelog.d/0.73.3/619.security.md
+++ b/changelog.d/0.73.3/619.security.md
@@ -1,0 +1,12 @@
+- **The SGLang serve backend now obeys the `--trust-remote-code` gate instead of
+  loading every model with it enabled (#360 by @Srinivasan8888 in #619).**
+  `serve.py` resolves the v0.36.0 default-deny gate once for every backend and
+  passes the result to vLLM, but `_serve_sglang` was called without it and
+  `create_sglang_runtime` hardcoded `trust_remote_code=True` at both
+  `sgl.Runtime` call sites, so `soup serve --backend sglang` executed a model's
+  custom repo code whether or not the user opted in. The warning panel said so,
+  but a notice is not a gate. `create_sglang_runtime` now defaults to `False`
+  like the vLLM path, and the resolved value reaches the runtime and the
+  tokenizer load alike. **Behaviour change:** a custom-code model on the SGLang
+  backend now fails to load without `--trust-remote-code` rather than silently
+  running its code.
--- a/src/soup_cli/commands/serve.py
+++ b/src/soup_cli/commands/serve.py
@@ -844,6 +844,7 @@ def serve(
             max_tokens_default=max_tokens_default,
             tensor_parallel=tensor_parallel,
             gpu_memory_utilization=gpu_memory_utilization,
+            trust_remote_code=resolved_trust,
         )
     else:
         # Transformers backend (original). ``resolved_trust`` was computed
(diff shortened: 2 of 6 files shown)


MakazhanAlpamys:
Merged as `e2610c5`. Thank you @Srinivasan8888 — **this was not the tidy-up either of us expected, and you were right to split it out of #581 rather than fold it in.**

## The claim is true as stated, and I checked it against `main` rather than the PR body

```
serve.py:809             resolved_trust = resolve_trust_remote_code(...)
serve.py:816-838         vLLM   -> trust_remote_code=resolved_trust
serve.py:839-846         sglang -> _serve_sglang(...)   <- no trust argument at all
utils/sglang.py:137,147  sgl.Runtime(..., trust_remote_code=True)   both sites
_serve_sglang            _load_serve_tokenizer(..., trust_remote_code=True) literal
```

`resolved_trust` was computed and then discarded for this backend. The panel told the user their model was trusted and offered no way to decline — **a notice, not a gate**. That is a genuine security hole, and "overstated" was the answer I was checking for.

It is also the shape you found one layer up in #581, where the tokenizer defaulted to `False` while the runtime hardcoded `True` and the whole fix silently did nothing on custom-code models. Same root cause, one level out.

## Per-surface mutations — the part that makes this trustworthy

| mutation | killed by |
|---|---|
| both `sgl.Runtime` sites back to `True` | `test_default_is_deny_on_the_plain_path`, `test_default_is_deny_on_the_adapter_path`, `test_create_sglang_runtime_full_model` |
| `serve()` passes literal `True` | `test_sglang_branch_forwards_resolved_trust` |
| tokenizer pinned back to `True` | `test_tokenizer_receives_the_resolved_value[False]` |
| **adapter path alone** back to `True` | `test_default_is_deny_on_the_adapter_path` |
| `_serve_sglang` stops forwarding | `test_runtime_receives_the_resolved_value[True]` / `[False]` |

The fourth row is the one I care about most: the adapter path and the plain path each fail **on their own**, so a future edit that repairs one and forgets the other cannot pass. A single combined test would have gone green on half a fix.

```
sglang + serve-trust suites   57 passed
test_sglang_serve.py          19 passed   (the True->False default does not break
                                           the one production caller)
test_cli_startup_is_light     34 passed   ·   ruff clean   ·   CI 13/13
```

Five deleted test lines, all inspected: four docstring rewording, one genuine `True -> False` assertion flip with a comment saying why. Nothing weakened.

## One follow-up, not blocking

`docs/performance-and-quantization.md:600-612` says "Every command that loads a model now requires `--trust-remote-code`" and lists `soup serve`. That sentence was **false for the SGLang backend** until this merge; your PR makes it true but does not note the behaviour change. `docs/serving-and-export.md`'s SGLang section documents the #360 chat-template fix and says nothing about trust either. A one-line addition mirroring that pattern would be worth having, since this is a breaking change: a custom-code model on that backend now fails to load where it previously loaded silently. Separate PR is fine.

`Refs #360` rather than `Fixes` is correct — the live-runtime item needs SGLang on Linux.

Issue #616, opened 2026-08-31
SSRF host predicate is duplicated: 2 copies of _is_private_or_link_local, 4 of _LOOPBACK_HOSTS


MakazhanAlpamys:
> **Correction (2026-09-01): this issue undercounted the copies.** It says two of
> `_is_private_or_link_local` and four of `_LOOPBACK_HOSTS`. An AST walk over
> `src/soup_cli/` — which I asked a contributor to write and did not run myself
> before filing — finds **three and six**:
>
> ```
> _is_private_or_link_local   hf.py, hubs.py, webhooks.py:31
> _LOOPBACK_HOSTS             hf.py, hubs.py, loop_stages.py, qr_url.py,
>                             webhooks.py:28, tracing.py:15
> ```
>
> `webhooks.py` is the shared Slack/Discord SSRF validator and `tracing.py` the
> OTLP one — the same predicate, the same category, the same job. Both belong in
> the consolidation. Found while reviewing #625, which correctly implemented the
> scope as originally written.
>
> **Second, sharper point, also found by mutation rather than by reading:** a guard
> that matches the exact names `is_private_or_link_local` / `LOOPBACK_HOSTS` does
> **not** catch a copy written with the leading underscore — and every one of the
> original copies used that spelling, so it is the most likely shape of a fifth
> one. Measured:
>
> ```
> def is_private_or_link_local(...)   ->  guard fails    caught
> def _is_private_or_link_local(...)  ->  guard passes   SURVIVED
> _LOOPBACK_HOSTS = frozenset({...})  ->  guard passes   SURVIVED
> ```
>
> The acceptance criterion below therefore needs strengthening: the guard must be
> **name-agnostic about the underscore prefix**, and the three mutations above must
> all fail it. A guard that passes while two duplicates sit in the tree is worse
> than no guard, because it reads as coverage.

## Summary

The SSRF host predicate exists in **two near-identical copies**, and its loopback allowlist in **four**. This project has been bitten three separate times by a predicate that was copied rather than shared — #372, #392 and #424 — and the shape is always the same: a fix lands in one copy and the other keeps the defect. These are security guards, which makes it the worst place for that shape.

Surfaced while answering #599 (telemetry DNS bypass) — it is not caused by that issue and should not be fixed inside it.

## Measured, not asserted

```
$ PYTHONPATH=src python -c "..."
soup_cli.utils.hf            ['127.0.0.1', '::1', 'localhost']
[...]

PR #624, opened 2026-08-31
fix(telemetry): tiered SSRF guard with primary allowlist, internal TLDs, and DNS defence in depth

### LINKED ISSUE #599: Telemetry SSRF guard: DNS-resolving hostnames bypass _is_private_or_link_local

here-2007:
## Summary

Closes #599.

Follow-up to #593 and #598. Closes the hostname indirection bypass in `_telemetry_endpoint_is_safe()` where domain names resolving to private/loopback addresses (e.g. `10.0.0.1.nip.io`, `localtest.me`) or internal TLDs (`metadata.internal`) bypassed literal IP checks.

---

### Architecture: Tiered Endpoint Validation

```
[Candidate Telemetry Endpoint]
              │
              ├─► Tier 1: Primary Control (Trusted Allowlist)
              │      └─► Canonical PostHog domains (*.posthog.com, us/eu.i.posthog.com)
              │          Accepted immediately (0ms, zero DNS queries, 100% offline-friendly for 99%+ runs).
              │
              ├─► Tier 2: Static Syntactic & Suffix Guards
              │      ├─► Scheme == "https"
              │      ├─► Host in _LOOPBACK_HOSTS (localhost, ::1, 127.0.0.1)
              │      ├─► Literal private/link-local IP check via _is_private_or_link_local()
              │      └─► Internal TLD suffix rejection (.local, .internal, .lan, .home, .corp, .intranet)
              │          Rejected immediately as pure string decisions with zero network lookups.
              │
              └─► Tier 3: Defence-in-Depth DNS Resolution (Custom Non-Default FQDNs)
                     ├─► Resolves custom hostnames via socket.getaddrinfo() with a bounded timeout.
                     ├─► Rejects if ANY resolved address is private, loopback, or link-local.
                     └─► Fails closed on resolution errors or timeouts.
```

---

### Key Security & Operational Principles

1. **Primary Control (Tier 1):** The trusted PostHog domain allowlist is the primary defense line. Default configurations never perform DNS queries or block in air-gapped environments.
2. **Honest Security Posture:** The Tier 3 DNS check raises the bar against hostname indirection for custom endpoint configurations; it is documented as defence-in-depth rather than preventing DNS rebinding.
3. **No Unbounded Latency & No Permissive Caching:** Unchecked `lru_cache` is omitted to avoid permanently trusting rebound IPs. Custom FQDN DNS checks run with a bounded join deadline.
4. **No Code Duplication:** Utilizes `_LOOPBACK_HOSTS` and `_is_private_or_link_local` from `hubs.py` without introducing redundant predicate copies.

---

### Tests

- **51 new tests** in `tests/test_issue599_telemetry_dns_ssrf.py` testing each tier independently:
  - `TestTier1TrustedAllowlist`: Verifies trusted allowlist and asserts DNS is never called.
  - `TestTier2StaticRejection`: Verifies internal TLDs and asserts DNS is never called.
  - `TestTier3DNSDefenceInDepth`: Verifies rejection of `10.0.0.1.nip.io`, `localtest.me`, cloud metadata, dual-stack mixed IPs, and timeout/error fail-closed handling.
  - `TestResolvePosthogTargetTieredIntegration`: End-to-end integration via `_resolve_posthog_target`.
- **Full sweep passed:** 161 tests across `test_issue599_telemetry_dns_ssrf.py`, `test_issue600_abbreviated_ipv4_ssrf.py`, and `test_telemetry_ssrf.py`.
- **Linter:** `ruff check` clean.

### DIFF
--- a/changelog.d/0.73.3/624.security.md
+++ b/changelog.d/0.73.3/624.security.md
@@ -1,0 +1,10 @@
+- **Telemetry SSRF guard now applies a tiered architecture with primary allowlist,
+  internal TLD rejection, and DNS defence in depth (#599 by @here-2007 in #624).**
+  `_telemetry_endpoint_is_safe()` previously only checked literal IP formats, allowing
+  hostnames resolving to private addresses (`10.0.0.1.nip.io`, `localtest.me`) or internal
+  TLDs (`metadata.internal`) to bypass validation. The guard now validates via three tiers:
+  (1) a primary allowlist for canonical PostHog domains (`*.posthog.com`) with zero DNS
+  lookups, (2) static string rejection of internal TLD suffixes (`.local`, `.internal`,
+  `.lan`, `.home`, `.corp`, `.intranet`) and literal private IPs, and (3) defence-in-depth
+  DNS resolution for custom non-default FQDNs that fails closed on resolver errors/timeouts
+  and rejects private or loopback target IPs.
--- a/src/soup_cli/utils/trackers.py
+++ b/src/soup_cli/utils/trackers.py
@@ -227,24 +227,86 @@ def _resolve_posthog_target(
     return key, resolved_endpoint
 
 
+# Internal / non-routable TLD suffixes rejected without DNS resolution (#599).
+_INTERNAL_TLD_SUFFIXES: tuple[str, ...] = (
+    ".local",
+    ".internal",
+    ".localhost",
+    ".lan",
+    ".home",
+    ".corp",
+    ".intranet",
+)
+
+_INTERNAL_TLD_EXACT: frozenset[str] = frozenset({
+    "local",
+    "internal",
+    "localhost",
+    "lan",
+    "home",
+    "corp",
+    "intranet",
+})
+
+
+def _is_trusted_posthog_domain(host: str) -> bool:
+    """Return True if ``host`` belongs to the trusted PostHog domain hierarchy."""
+    return host == "posthog.com" or host.endswith(".posthog.com")
+
+
+def _resolve_host_ips(host: str, *, timeout: float = 1.0) -> list[str] | None:
+    """Resolve ``host`` to IP addresses within ``timeout`` seconds.
+
+    Returns a list of IP strings, or ``None`` on resolution failure,
+    resolver error, or timeout (fail-closed).
+    """
+    import socket  # noqa: PLC0415
+    import threading  # noqa: PLC0415
+
+    result: list[str] = []
+    error: list[Exception] = []
+
+    def _worker() -> None:
+        try:
+            info = socket.getaddrinfo(host, None, socket.AF_UNSPEC, socket.SOCK_STREAM)
+            for item in info:
+                sockaddr = item[4]
+                ip = sockaddr[0]
+                if ip:
+                    result.append(ip)
+        except Exception as exc:
+            error.append(exc)
+
+    thread = threading.Thread(target=_worker, daemon=True)
+    thread.start()
+    thread.join(timeout=timeout)
+    if thread.is_alive() or error or not result:
+        return None
+    return result
+
+
 def _telemetry_endpoint_is_safe(endpoint: str) -> bool:
-    """Layered SSRF guard for telemetry endpoints (#593).
+    """Tiered endpoint guard for telemetry destinations (#593, #599).
 
-    Two-layer validation:
+    Three-tier validation architecture:
 
-    1. :func:`~soup_cli.utils.hubs.validate_hub_endpoint` — baseline
-       sanitization (CRLF rejection, null-byte rejection, type guards,
-       ``0.0.0.0`` handling, scheme allowlist).  This layer is shared
-       with model-hub endpoints and intentionally permits private HTTPS
-       hosts for self-hosted hubs.
-    2. **Telemetry-strict rejection** — rejects loopback hosts
-       (``localhost``, ``127.0.0.1``, ``::1``), private IPs (RFC 1918),
-       and link-local addresses (``169.254.x.x``) on *any* scheme.
-       Telemetry has no legitimate self-hosted-private-endpoint case,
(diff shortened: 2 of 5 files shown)


MakazhanAlpamys:
Merged as `5ed6e43`. Thank you @here-2007 — and the part I want to name first is that **you pushed back on my design before building it, and you were right.**

My sketch would have resolved every custom endpoint. You pointed out that `socket.getaddrinfo()` takes no timeout argument in Python and that `send_telemetry` calls `httpx.post` **synchronously**, so an unbounded resolver call would block air-gapped runs — and proposed the tiered shape where a trusted allowlist short-circuits before any lookup happens. That is a better design than the one you were handed, and asking first cost less than building the wrong one.

## Verified by measurement, not by reading the diff

```
socket.getaddrinfo patched to raise
  https://us.i.posthog.com/capture/  ->  safe=True,  getaddrinfo calls = 0,  35.2 ms
```

The air-gap guarantee holds and it has a named test (`test_tier1_never_performs_dns_resolution`). And the bypass is closed against **live DNS**, not mocks:

```
                                        main    PR
https://10.0.0.1.nip.io/capture/        True -> False    real DNS -> ['10.0.0.1']
https://localtest.me/capture/           True -> False    real DNS -> ['127.0.0.1']
https://metadata.internal/capture/      True -> False
https://attacker-controlled.example/    True -> False    NXDOMAIN, fail-closed
https://us.i.posthog.com/capture/       True -> True     legit default preserved
```

That last row is the one I check hardest. A validator that rejects everything is worse than the bug.

## All six points from the issue, checked individually

| # | point | evidence |
|---|---|---|
| 1 | tier 1 is the **primary control**, documented as such | docstring and code comment both say "Primary Control", not "fast path" |
| 2 | DNS is defence in depth, never "prevents SSRF" | grep for `prevents/blocks/stops SSRF` -> **0 hits**; the docstring states the rebinding limitation explicitly |
| 3 | **no `lru_cache`** | `grep lru_cache\|functools` in `trackers.py` -> **0** |
| 4 | static internal-TLD rejection | all seven suffixes, plus an exact-match set for bare labels |
| 5 | fail-closed on resolver error | verified live against NXDOMAIN |
| 6 | no fifth copy of the shared predicate | `hubs.py` and `hf.py` byte-unchanged; `trackers.py` imports and defines neither |

Point 3 is the one I would most have expected to be argued with, since caching is the obvious optimisation. Taking it without pushback was correct — the risk is asymmetric, a cached `False` is harmless and a cached `True` means a rebinding attacker only has to win the race once, for the life of a process that runs for hours.

## Mutations

```
tier 1 bypassed        3 failed   test_tier1_never_performs_dns_resolution (+2)
internal-TLD removed   1 failed   test_tier2_internal_tld_never_performs_dns
DNS layer removed      7 failed   6x TestTier3DNSDefenceInDepth + the integration case

176 passed across the four telemetry/SSRF files
test_v05310.py 45 passed  (touches the same env var, not updated -- not silently broken)
test_cli_startup_is_light 34 passed  ·  ruff clean  ·  CI 13/13
```

**One thing recorded in the merge commit rather than glossed over:** tier 2's *security outcome* is not independently pinned. Disable it and `metadata.internal` is still refused, because tier 3 NXDOMAINs and fails closed. What pins tier 2 is its **no-network property**, via one test — and that margin is exactly one test wide. It matters on a network where NXDOMAIN is hijacked: there `metadata.internal` resolves to a public IP, tier 3 accepts it, and tier 2 is the only thing left refusing. Not a blocker, and your test does exist; I would just not delete it casually.

## Two small things, neither worth a push

- **`send_telemetry_payload`'s docstring still says `timeout: hard wall-clock cap (default 1 s)`.** For a *custom* endpoint the tier-3 resolver now adds its own 1.0 s deadline ahead of `httpx`, so the real worst case is ~2 s. Measured with a hanging resolver: 1.02 s before httpx was reached. The default endpoint is unaffected. One docstring line whenever you are next in the file.
- The `test_issue600` parametrize named `..._allows_public` now has all three entries under `*.posthog.com`, i.e. tier 1, so it no longer exercises a public *non-allowlisted* host despite its name. The coverage still exists elsewhere; it is redundancy loss, not a hole.

Also noted and correctly left alone: `_is_private_or_link_local` accepts RFC 6598 CGNAT space (`100.64.0.0/10`). That is the shared `hubs.py` predicate's pre-existing behaviour, not something you introduced, and it belongs on #616.

Your 5th merge.

PR #629, opened 2026-09-01
feat(autodistill): add same-tokenizer teacher-only capture

Amix29:
## Summary

Implements the narrow Milestone B1 slice claimed in #580, strictly on top of the merged
Milestone A contract:

- teacher-only MLX capture in a fresh subprocess, with no student model input or load;
- immutable teacher, canonical-tokenizer, and dataset byte verification;
- full supervised target trajectories using top-k union forced-target plus residual tail mass;
- transactional staging -> complete -> verified -> available shard publication;
- fail-closed corruption, resume, traversal, runtime-version, dtype, and quantization checks;
- a reproducible opt-in Apple Silicon hardware smoke script.

This PR does not add a task, CLI command, or SoupConfig field. It also does not add student
training, adaptive generation/allocation, cross-tokenizer ULD, or benchmark integration.

Refs #580.

## Boundary and artifact semantics

The controller writes a canonical request, starts a fresh Python worker, waits for the child to
exit, reopens the available manifest, and only then writes the final result. The request has no
student path; the Milestone A student fingerprint remains future-consumer metadata only.

The teacher model and canonical shared tokenizer can come from separate immutable roots. Only
the teacher model is loaded. This matters for otherwise token-compatible model pairs whose
native chat-template files differ. Tokenizer files are verified before model loading, and the
instantiated template plus exact mlx-lm renderer version are verified before publication.

Quantized checkpoints bind the active quantization config by canonical SHA-256. The worker
requires the declared base floating dtype, permits only float32 auxiliaries for a quantized
checkpoint, and records every observed floating parameter dtype in the child receipt.

## Validation

```text
95 passed in 9.69s
ruff check src/soup_cli/ tests/ scripts/autodistill_b1_hardware_smoke.py: PASS
git diff --check origin/main...HEAD: PASS
```

The focused tests cover deterministic trajectory capture, exact truncation contexts, runtime
isolation, separate tokenizer roots, pre-load byte verification, mixed-dtype quantized teachers,
transaction state transitions, interruption/resume, corruption quarantine, and import-light CLI
startup.

### Apple Silicon smoke

Hardware/runtime:

- MacBook Pro M4 Max, 128 GiB unified memory;
- MLX 0.32.0 and MLX-LM 0.31.3;
- teacher: `txgsync/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-oQ4e-mtp`;
- future student fingerprint: `nvidia/NVIDIA-Nemotron-3-Nano-4B-BF16`;
- canonical tokenizer: tokenizer-only materialization of the Nano tokenizer;
- two target positions, repeated in two fresh workers.

Observed:

```text
run 1 capture time: 10.658 s
run 2 capture time: 10.749 s
student_loaded: false
observed floating dtypes: [bfloat16, float32]
capture.jsonl SHA-256 (both runs):
8097b731bbe8c67b4c5b00c74d9955162e555f03d9611bccdbfd183d835f6a8c
available manifest file SHA-256 (both runs):
b5b51da89165ea90e2a0926eb87f74c0b7c0af7bc1f35fd3131e8e3e2060c221
```

The workers exited, the scientific payload and manifest were byte-identical across runs, and
the local oMLX server reported zero loaded models after capture.

This smoke proves the capture and release boundary. It does not claim downstream student quality
or quantized-vs-BF16 logit equivalence; those require later controlled training and evaluation.

## Review focus

1. Is the B1 worker request/receipt boundary narrow enough before a public CLI exists?
2. Is a separately pinned canonical tokenizer root the preferred same-tokenizer interpretation?
3. Is the quantization identity and mixed-dtype receipt sufficiently explicit for later cache
   consumers?
4. Should the manual hardware smoke script remain in this PR or move to a later integration slice?

### DIFF
--- a/changelog.d/0.73.3/629.added.md
+++ b/changelog.d/0.73.3/629.added.md
@@ -1,0 +1,4 @@
+- Added the internal AutoDistill Milestone B1 same-tokenizer, teacher-only MLX capture and
+  transactional shard publication boundary, with immutable input fingerprints, explicit
+  corruption/resume checks, and proof that no student model is loaded during capture
+  (#580 by @Amix29 in #629).
--- a/docs/autodistill.md
+++ b/docs/autodistill.md
@@ -243,17 +243,22 @@ root. The controller writes its final receipt only after that child has exited s
 the exact `available` manifest has been reopened. The plan still contains the immutable student
 fingerprint as future-consumer metadata; it is never resolved or loaded during capture.
 
-For this boundary, MLX-LM loads both model and tokenizer from the same local checkpoint root.
-`capture.backend_version` is the exact installed `mlx-lm` distribution version and the tokenizer
-renderer is `mlx-lm@<version>`. Both must match the plan before publication. MLX/MLX-LM remain
-lazy imports inside the worker.
+For this boundary, MLX-LM loads the teacher model from its immutable local checkpoint root and
+loads the canonical shared tokenizer separately from a tokenizer-only root. That tokenizer root
+may be derived from the student's tokenizer bytes, but it contains no student weights and the
+worker never resolves a student model. `capture.backend_version` is the exact installed `mlx-lm`
+distribution version and the tokenizer renderer is `mlx-lm@<version>`. Both must match the plan
+before publication. MLX/MLX-LM remain lazy imports inside the worker.
 
-The worker also verifies the loaded floating-parameter dtype against `capture.dtype`. An
-unquantized checkpoint uses the literal quantization identity `none`. If `config.json` contains
+The worker also verifies the loaded floating-parameter dtypes against `capture.dtype`. An
+unquantized checkpoint must expose exactly that dtype and uses the literal quantization identity
+`none`. A quantized checkpoint must expose the declared base dtype; float32 auxiliary parameters
+are allowed, while any other undeclared floating dtype fails closed. If `config.json` contains
 `quantization` or `quantization_config`, its identity is
 `config-sha256:<sha256(canonical JSON of those active fields)>`. This binds the plan to the exact
 quantization recipe without pretending that names such as `4-bit` uniquely identify a runtime.
-The observed dtype and quantization identity are repeated in the child receipt.
+The declared inference dtype, all observed floating-parameter dtypes, and the quantization
+identity are repeated in the child receipt.
 
 The bound dataset is canonical JSONL with versioned, already-tokenized rows:
 
--- a/docs/autodistill.md
+++ b/docs/autodistill.md
@@ -234,6 +234,37 @@ available -> reserved -> committed
   `replay_of` equals the prior committed event hash. The source artifact stays immutable.
 - Replay policy is explicit in every plan; it is not inferred from a filename or task.
 
+### 7.4 Milestone B1 MLX teacher boundary
+
+The first Apple Silicon capture worker is an internal, opt-in implementation surface; it
+does not add a task, CLI command, or `SoupConfig` field. The controller starts a fresh Python
+process whose request accepts teacher, tokenizer, dataset, and publication roots but no student
+root. The controller writes its final receipt only after that child has exited successfully and
+the exact `available` manifest has been reopened. The plan still contains the immutable student
+fingerprint as future-consumer metadata; it is never resolved or loaded during capture.
+
+For this boundary, MLX-LM loads both model and tokenizer from the same local checkpoint root.
+`capture.backend_version` is the exact installed `mlx-lm` distribution version and the tokenizer
+renderer is `mlx-lm@<version>`. Both must match the plan before publication. MLX/MLX-LM remain
+lazy imports inside the worker.
+
+The bound dataset is canonical JSONL with versioned, already-tokenized rows:
+
+```json
+{"schema":"soup.autodistill.tokenized-teacher-example.v1","example_id":"ex-1","prompt_token_ids":[1,2],"target_token_ids":[3,4]}
+```
+
+Prompt IDs are non-empty because a causal next-token distribution needs a preceding position.
+Every source byte, normalized row, tokenizer file, teacher config, and listed teacher weight is
+verified before model load. A worker request selects a half-open example range, but the complete
+dataset fingerprint and total planned target-token count are verified before that shard subset is
+accepted. Ranges never split one example trajectory.
+
+Without truncation, one causal forward captures all target positions in an example. Once the
+declared sequence limit requires truncation, the worker evaluates the exact recorded context for
+each position. Only the final vocabulary row is converted to float32 host values; selected IDs are
+not renormalized. Process exit, rather than cache reclamation alone, is the memory-isolation gate.
+
 ## 8. Transactional publication, resume, and corruption
 
 The capture writer follows a write-last manifest protocol on one filesystem:
--- a/docs/autodistill.md
+++ b/docs/autodistill.md
@@ -248,6 +248,13 @@ For this boundary, MLX-LM loads both model and tokenizer from the same local che
 renderer is `mlx-lm@<version>`. Both must match the plan before publication. MLX/MLX-LM remain
 lazy imports inside the worker.
 
+The worker also verifies the loaded floating-parameter dtype against `capture.dtype`. An
+unquantized checkpoint uses the literal quantization identity `none`. If `config.json` contains
+`quantization` or `quantization_config`, its identity is
+`config-sha256:<sha256(canonical JSON of those active fields)>`. This binds the plan to the exact
+quantization recipe without pretending that names such as `4-bit` uniquely identify a runtime.
+The observed dtype and quantization identity are repeated in the child receipt.
+
 The bound dataset is canonical JSONL with versioned, already-tokenized rows:
 
 ```json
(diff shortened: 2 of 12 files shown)


MakazhanAlpamys (changes requested):
Reviewed at `417c778`. **Blocking on the mutation set, and on nothing else** — this is a coverage gap, not a design problem, and everything else here is at or above the bar #613 set.

## What is good, stated first because it is most of the PR

**Scope honesty is exemplary.** `+2680 / −1`, and the single deletion is a one-line docstring reword. No `cli.py`, no `commands/*`, no `config/schema.py` — matching your claim that this adds no task, no CLI, no `SoupConfig` field. `MlxTeacherCaptureRequest` has no student field and `extra="forbid"`. `Refs #580`, not `Fixes`. §7.4 of the docs states the boundary. I checked each of those rather than taking the body's word for it.

**You avoided the trap that cost #613 a round.** Zero committed exact-byte fixtures, zero byte-length or literal-SHA assertions in the four new test files. The `core.autocrlf` failure mode cannot recur here.

Conventions all clean: no torch/transformers/mlx in `sys.modules` after `import soup_cli.autodistill` (runtime-verified, not the AST guard), `test_cli_startup_is_light.py` 34/34, `realpath`+`commonpath` everywhere, atomic `mkstemp`+`os.replace`, `subprocess.Popen` with list args and `DEVNULL` stdin, ruff clean, no `eval`/`pickle`/`torch.load`. Your "95 passed" reconciles exactly as 31 + 30 + 34.

## The blocker

Eight load-bearing guards can be deleted with **zero** failing tests across the full 61-test autodistill suite. Seven of them are genuinely uncovered rather than equivalent:

| mutation | file:line | result |
|---|---|---|
| delete `if row.trajectory_kind != "teacher_expert": raise` | `publisher.py:152-153` | **61 passed** |
| delete `commonpath` containment in `_contained_path` | `publisher.py:36-41` | **61 passed** |
| delete `canonical_sha256(manifest) != receipt.available_manifest_sha256` | `mlx_worker.py:553-554` | **61 passed** |
| delete the manifest hash-chain check | `publisher.py:280-281` | **61 passed** |
| `if False:` on the top-k cardinality check | `publisher.py:157` | 18 passed |
| `if False:` on the context-length check | `publisher.py:161` | 18 passed |
| `if False:` on the `vocab_size` check | `publisher.py:153` | 18 passed |

The first one is the one that decides this review. **`trajectory_kind != "teacher_expert"` is the teacher-only publication boundary — the title claim of this PR — and it has no test anywhere.** `publisher._capture_payload` is also the semantic re-verifier (`_verify_semantics`, `publisher.py:313-325`), so this is precisely the layer at which a student-rollout row is supposed to be structurally impossible. `CONTRIBUTORS.md` records that same boundary as the mutation that mattered most on #613, and here it survives.

The containment one is not theoretical either. On unmutated code, `CaptureShardPublisher(root=..., shard_id='../../escaped', ...)` raises `ValueError: artifact path is outside the publication root`. `shard_id` and `transaction_id` carry **no pattern validation on the publisher constructor** — `_ARTIFACT_ID_PATTERN` lives only on `MlxTeacherCaptureRequest` (`mlx_worker.py:48`), one layer above — and `CaptureShardPublisher` is a public export in `autodistill/__init__.py:38`. `test_worker_ids_fail_closed_before_spawn` catches `../escape` at the pydantic layer, not at this guard.

For balance, the mutations that ARE well covered: forced-target union (13 failed), mass conservation (14 failed), renderer identity and chat-template SHA (1 each, distinct), the loaded-dtype/quantization gate (1, parametrized), the symlink half of `_safe_file` (1).

**Two survivors I am NOT asking you to fix**, and I would rather you reported them in the body than manufactured kills for them:

- `_safe_file`'s `commonpath` in `fingerprints.py:22-27` is near-equivalent. `FileDigest.path` is already validated by `_require_safe_relative_path` (`contract.py:92-102`) and that IS tested. Defence in depth over an already-validated value.
- `_atomic_write` → `path.write_bytes` is unobservable in-process without crash injection, and no other module in this repo carries such a test.

## To unblock

Add tests killing the seven. Concretely: (a) `publisher.publish()` with a `student_rollout` `CaptureToken` must raise — this is the title claim and has no test at all; (b) `CaptureShardPublisher(shard_id='../../x')` must raise, at the publisher layer rather than the pydantic one; (c) a hand-tampered `manifest.complete.json` with a wrong `previous_manifest_sha256` must be refused on resume; (d) rows disagreeing with the plan on top-k cardinality, context length and `vocab_size` must each be refused — the sibling `temperature` case already has a test, so this is filling in three of four. Note `test_publisher_rejects_duplicate_reordered_and_wrong_policy_rows` covers **temperature only**, despite what its name suggests.

## One security finding worth fixing in the same round

**MEDIUM — `fingerprints.py:41`**: `_verify_file` hashes weight shards with a full `read_bytes()`, no chunking. A 30B teacher's safetensors shards are read **entirely into RAM** to compute SHA-256. The whole premise of #580 is memory-constrained hardware, and your smoke ran on 128 GiB, which is exactly why this would not have surfaced. An incremental `hashlib` update over a bounded buffer.

Two LOW ones, take or leave: `_ARTIFACT_ID_PATTERN` permits `:`, and on Windows `ntpath.join('C:\\root\\shards', 'C:evil')` → `C:\root\shards\evil` (verified) — no escape, containment holds, but two distinct `shard_id`s silently collide into one directory. And `assert isinstance(...)` is used for control-flow narrowing at `mlx_worker.py:425` and `:548`, which `python -O` strips.

## Housekeeping

`changelog.d/0.73.3/629.*.md` is missing. Your own #613 shipped one for an equally internal-only slice, and #603 ships one. Either add it or say in the body that this slice is deliberately user-invisible — do not leave it silent, since CI checks fragment staleness rather than presence and will not catch it.

Zero file overlap with your #603; they can merge in either order.

**What I could not verify:** the Apple Silicon smoke numbers. No MLX here, so the 10.658s/10.749s timings, the two byte-identical SHA-256s and "zero loaded models after capture" are on your word. The in-repo `test_real_child_process_captures_full_trajectory_and_exits` exercises the same boundary against a traced fake runtime in a real child process, and that part I did verify.

Amix29:
Addressed the requested B1 review at `51e8be6`.

- Added distinct tests for all seven load-bearing guards: teacher-only `trajectory_kind`, publisher-layer path containment, worker-receipt/available-manifest binding, manifest hash chaining, top-k cardinality, context-length ceiling, and vocabulary identity.
- Re-ran the seven mutations in isolated temporary source copies; every mutation is now killed by its named test.
- Replaced full-file teacher weight reads with incremental SHA-256 hashing in bounded 8 MiB chunks. The regression test refuses `Path.read_bytes()` on weight shards and observes multiple bounded reads.
- Added `changelog.d/0.73.3/629.added.md` with the B1 scope and attribution.

Local verification:

- focused publisher/fingerprint/worker suite: `31 passed`
- AutoDistill + import-light + changelog gates: `120 passed`
- full non-smoke suite outside the sandbox, excluding only the known local CPU-only Torch `doctor_nccl_mocked_success` CUDA-mock incompatibility: `19,489 passed, 98 skipped, 6 deselected`
- full Ruff: clean

I left the two explicitly non-requested near-equivalent/atomic-crash survivors and the optional LOW findings unchanged, keeping this round to the requested unblock set.

MakazhanAlpamys:
Merged as `cd1a694`, credited in CONTRIBUTORS at `d77e9af`.

**The second round is the reason this merged, and it is a genuinely better piece of work than the first.** Eight mutations survived last time, including the boundary in your own PR title. I re-ran them myself rather than taking the claim — all seven load-bearing guards now die, each to one distinctly named test that calls the real function with real input.

Two things I checked specifically, because they are what separate coverage from the appearance of it:

- **I probed the title guard by *widening* it**, not deleting it: `trajectory_kind not in ("teacher_expert", "student_rollout")`. That is the regression someone would actually write, and a test that only catches a deleted guard would sail past it. Result: **1 failed, 10 passed** — `test_publisher_rejects_student_rollout_at_teacher_only_boundary`, exactly one test, no incidental breakage.
- **I then mutated the same guard to reject *everything*.** A guard tested only in the rejecting direction passes that. Yours does not: **9 failed, 2 passed**, and the failures land on real happy-path tests. So the valid case is genuinely exercised and none of the seven is a one-sided assertion.

`test_resume_rejects_tampered_manifest_hash_chain` deserves a specific mention — publishing for real, tampering `manifest.complete.json` on disk, then re-publishing is the strongest shape available for that property, and it is the kind of test most people would have written as a unit assertion over a hash function instead.

Verified before merging: 39 tests across the four touched files, **zero deleted test lines** both against `main` and between `417c778a..51e8be6` (I checked the second explicitly — a PR on its third head is where things quietly disappear), 13/13 CI, and no scope expansion: no Typer command, no flag, no `SoupConfig` field, `pyproject.toml` untouched. `Refs #580` rather than `Fixes` is the right call and I am glad you made it without being asked — #580 is your proposal, not an approved design.

The MEDIUM finding is closed properly too: bounded 8 MiB streaming reads, pinned by a test that makes `Path.read_bytes` raise.

**Two follow-ups I deliberately did not hold this for**, both fair game whenever you next touch these files:

1. The token-id vocabulary-range guard is untested at both layers — `mlx_worker._select_examples` and `capture._validate_token_ids` — and neither message appears anywhere in `tests/`. Lower class than the seven, since an out-of-range id faults in the MLX runtime rather than silently corrupting a published shard.
2. `retain_bytes=False` on weight shards is pinned for *chunking* but not for *non-retention*: flipping it back to `True` restores the full-RAM behaviour the MEDIUM finding was about, with the suite green. One assertion on peak retained bytes closes it.

Also minor: `docs/README.md`'s index row for `autodistill.md` still says "Milestone A schemas" while the page now documents B1.

PR #630, opened 2026-09-01
Unknown config keys are silently dropped: a typo'd or not-yet-released setting

### LINKED ISSUE #627: Unknown config keys are silently dropped: a typo'd or not-yet-released setting validates clean and does nothing

xiaoyaoqilan:
## What does this PR do?

This PR fixes the documentation issue reported in #627: corrects `--dry-run` to `ld` in `examples/README.md`.

Fixes #627

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Tests

## Checklist

Fixes #627

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -40,7 +40,7 @@ soup train --config examples/configs/sft_basic.yaml
 Validate a config without training:
 
 ```bash
-soup train --config examples/configs/sft_basic.yaml --dry-run
+soup train --config examples/configs/sft_basic.yaml ld
 ```
 
 Gated models (Llama-2, Llama-3.1) need Hugging Face access and a login first:


MakazhanAlpamys:
Closing this — the change is not correct, and it is not what #627 is about.

The diff replaces a working flag with a string that is not a flag:

```diff
-soup train --config examples/configs/sft_basic.yaml --dry-run
+soup train --config examples/configs/sft_basic.yaml ld
```

`--dry-run` is a real option on `soup train` and the line was right before this. `ld` is not an option, not a subcommand, and not a positional argument the command accepts, so the documented example would stop working for anyone who copied it.

The linked issue is also unrelated. #627 is about the config schema silently dropping unknown YAML keys — `extra="ignore"` on the Pydantic models — and is being fixed in #628. Nothing in it concerns `examples/README.md` or the `--dry-run` flag.

No hard feelings if this was an accidental edit or an automated submission. If you would like to contribute for real, `CONTRIBUTING.md` describes the workflow, and I am happy to point you at something concrete — say the word and I will find an issue that suits what you want to work on. What I ask for on every PR is that the change is verified against the behaviour it touches: for a docs change that means running the command as written.

PR #631, opened 2026-09-01
test(log_level): pin the foreign-handler branch of setup_logging (#273)

umran666:
Closes the remaining row of #273 (claimed in [this comment](https://github.com/MakazhanAlpamys/Soup/issues/273#issuecomment-5490791363), with the offered-take silence stated there).

## What

Pins the `74->73` loop-back edge of `setup_logging` deterministically — the branch taken when the `soup` logger carries a handler **without** the `_soup_log_tier` tag. Measured on this machine before the change (Windows, Python 3.10.11, `tests/test_log_level.py` only):

```
src\soup_cli\utils\log_level.py      45      0     14      1    98%   74->73
15 passed
```

Same as the maintainer's run — the branch was not covered, only incidentally executed depending on ambient logger state.

## The tests (both directions, per the 2026-08-29 spec)

- `test_foreign_handler_survives_reconfigure` — a handler with no `_soup_log_tier` attached to the `soup` logger **survives** `setup_logging(NORMAL)`. This pins the real contract: Soup must not rip out logging handlers installed by an embedding application.
- `test_stale_soup_handler_removed_with_foreign_present` — a Soup handler with a stale tier is **removed** in the same pass, and exactly one NORMAL-tagged handler is rebuilt.

Both are hermetic: an autouse fixture snapshots the `soup` logger's handlers/level/propagate and restores them, which is also the missing-cleanup concern raised on the issue for this file.

## Mutation demo — shown, not asserted

Temporarily dropping the `is not None` guard in `log_level.py`:

```
FAILED tests/test_log_level.py::TestIssue273ForeignHandlerContract::test_foreign_handler_survives_reconfigure
FAILED tests/test_log_level.py::TestIssue273ForeignHandlerContract::test_stale_soup_handler_removed_with_foreign_present
2 failed in 1.24s
```

Both die by name; guard restored, `17 passed`.

## Post-fix BrPart figure, quoted per the ask

```
src\soup_cli\utils\log_level.py      45      0     14      0   100%
17 passed in 28.94s
```

`BrPart` is now 0 **for a reason** — the foreign handler forces the edge — not by luck of ambient state.

## Verification

- `ruff check src/soup_cli/ tests/` clean
- `tests/test_log_level.py` + `tests/test_v0402_part_b.py` + `tests/test_v0713.py` (every test matching `log_level or logging`): **122 passed, 2 pre-existing skips**
- No source change — `log_level.py` is untouched; the mutation was demonstrated and reverted

### DIFF
--- a/changelog.d/0.73.3/631.fixed.md
+++ b/changelog.d/0.73.3/631.fixed.md
@@ -1,0 +1,8 @@
+- [#631](https://github.com/MakazhanAlpamys/Soup/pull/631) pins the last
+  open row of #273: the `74->73` branch of `setup_logging`, which was only
+  ever executed incidentally depending on ambient logger state (100% branch
+  coverage on one machine, 98% on another). Two hermetic tests now force the
+  state: a handler without the `_soup_log_tier` tag survives reconfiguration,
+  and a Soup handler with a stale tier is still removed. Dropping the
+  `is not None` guard fails both tests by name (mutation demonstrated and
+  reverted). `BrPart` 1 -> 0, for a reason rather than by luck.
--- a/tests/test_log_level.py
+++ b/tests/test_log_level.py
@@ -92,6 +92,63 @@ class TestSetupLogging:
             parse_log_level(42)  # type: ignore[arg-type]
 
 
+class TestIssue273ForeignHandlerContract:
+    """Deterministically pin the ``74->73`` branch of ``setup_logging`` (#273).
+
+    The branch is the loop-back edge taken when a handler on the ``soup``
+    logger carries **no** ``_soup_log_tier`` tag — a handler Soup did not
+    install. Whether ambient logger state happens to exercise it varies by
+    platform and test ordering, so these tests force the state explicitly:
+    an embedding application's handlers must survive reconfiguration, while
+    a Soup handler with a stale tier is still removed.
+    """
+
+    @pytest.fixture(autouse=True)
+    def _isolate_soup_logger(self):
+        logger = logging.getLogger("soup")
+        saved_handlers = list(logger.handlers)
+        saved_level = logger.level
+        saved_propagate = logger.propagate
+        yield
+        for handler in list(logger.handlers):
+            logger.removeHandler(handler)
+        for handler in saved_handlers:
+            logger.addHandler(handler)
+        logger.setLevel(saved_level)
+        logger.propagate = saved_propagate
+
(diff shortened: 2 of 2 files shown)


MakazhanAlpamys:
Merged as `b8a481c`. Thank you — and the thing worth naming is that **your PR is more correct than the issue it closes.**

I wrote #273 saying branch `74->73` is "an existing handler whose `_soup_log_tier` **matches** the requested tier", with the repro "call `setup_logging(tier)` twice with the same tier". That is `75->73`. `log_level.py:74` is `if getattr(existing, "_soup_log_tier", None) is not None:`, and the uncovered branch is that condition being **False** — a handler Soup did not install. Following my own repro would not have closed the branch. You read the source instead of the ticket, which is the correct instinct and the one that is easiest to skip when a ticket hands you a table.

Reproduced your numbers exactly:

```
origin/main   45 stmts, 14 branches, 1 partial, 98%, missing 74->73   15 passed
this branch   45 stmts, 14 branches, 0 partial, 100%                  17 passed
```

Mutations, all in `utils/log_level.py`:

| mutation | result |
|---|---|
| drop the `is not None` guard | 2 failed (both new tests), 15 passed |
| removal loop neutered to `pass` | 2 failed, 15 passed |
| line 75 dropped — a same-tier handler is also rebuilt | 1 failed: `TestSetupLogging::test_idempotent` |

The first is the one that settles it: **under that mutation all 15 pre-existing tests pass**, so your two tests are the only guard on that branch. That is what non-vacuous means and it is worth stating as a number rather than as a claim.

The four deleted lines needed checking and they earn their place. I restored `main`'s version of `test_idempotent` on top of the third mutation: **17 passed** — a fully green suite while `setup_logging` destroys and rebuilds its handler on every single call. The old assertion compared handler *counts*, which that mutation preserves. So the rewrite is a strengthening, not a convenience, and it satisfies step 3 of #273 directly.

I also ran a leakage check across six files that touch logging state (`test_log_level`, `test_cli`, `test_v0713`, `test_v0402_part_b`, `test_doctor`, `test_runs`): 165 passed, 2 skipped. Your autouse fixture snapshots and restores handlers, level and propagate, so the new tests do not leak into neighbours — which for a test that installs a foreign handler on the root logger is the failure mode I would have expected to find.

`Refs #273`, not `Fixes` — I have left the issue open and corrected its table, since the branch it names is a different one and someone should still look at whether `75->73` deserves its own test.

PR #614, opened 2026-08-31
feat(recipes): add kimi-k2.6-grpo recipe (#281)

umran666:
Closes #281 (claimed in [this comment](https://github.com/MakazhanAlpamys/Soup/issues/281#issuecomment-5477796292)). Part of the #275 task-variant tracking.

GRPO reasoning variant of the v0.71.24 `kimi-k2.6-sft` sibling. Per the issue's How: copies the GRPO + MoE giant shape from `qwen3-30b-a3b-reasoning-grpo` — `grpo_beta: 0.1`, `num_generations: 4`, `reward_fn: accuracy`, `moe_lora: true`, `gradient_checkpointing: true`, `batch_size: 1`, `gradient_accumulation_steps: 16` — and points `base:` at `moonshotai/Kimi-K2.6`, keeping the **Modified MIT** note in the description.

- [x] `soup recipes show/use kimi-k2.6-grpo` work; YAML loads — pinned by `TestIssue281KimiK26GrpoRecipe` (CLI-invoked `show` and `use` via the Typer runner, plus a `load_config_from_string` round-trip asserting every shape field the issue names)
- [x] Base resolves on Hugging Face — `moonshotai/Kimi-K2.6` returns HTTP 200 from the HF API (same base the shipped SFT sibling already uses)
- [x] `ruff` + `pytest tests/test_recipes.py` green — plus `test_recipe_count_is_synced.py`, `test_recipes_v031.py`, `test_v0510.py`, `test_v07124.py`, `test_v07130.py`, `test_v07132.py`: **2038 passed, 1 skipped** (pre-existing skip)

Catalog count 159 → 160 bumped at every declared documentation site (catalog header, `CONTRIBUTING.md`, `docs/commands.md`, both `docs/serving-and-export.md` sites), with the #453 ratchet passing; milestone size assertions in `test_v07124/v07130/v07132` bumped per the #582 merge convention.

PR #632, opened 2026-09-01
feat(recipes): add qwen3.5-9b-dpo recipe (#275)

Srinivasan8888:
Adds the DPO task-variant for `Qwen/Qwen3.5-9B` under #275's "more variants beyond this list" line. Claimed the specific recipe on the tracking issue first, per your coordination note on #379.

`Refs #275` — the tracking issue stays open for the remaining families.

## Why this recipe

Measured rather than picked by eye. Of the 2026 families in the catalog:

```
Qwen/Qwen3.5-9B            ['grpo', 'sft']    <- no preference variant
Qwen/Qwen3.5-35B-A3B       ['dpo', 'sft']     (#276, merged as #615)
deepseek-ai/DeepSeek-V4-Flash  ['grpo','sft']
zai-org/GLM-5.1            ['dpo', 'sft']
...16 more models carrying SFT only
```

Qwen3.5-9B already had two of the three tasks #275 is about, so DPO completes the trio on the mid-size model of the family with the largest catalog presence. All six named children of #275 are now done (#276→#615, #277, #278, #279, #280, #281→#614); no open issue or PR mentioned `9b-dpo`.

## Shape

The `qwen2.5-7b-dpo` DPO half — `format: dpo`, `lr: 5e-6`, `dpo_beta: 0.1`, `preference_train.jsonl` — with the SFT sibling's geometry: LoRA r16/a32, 4-bit, `max_length: 4096`.

`lr: 5e-6` matches **all 11** existing DPO recipes. Noting it because #512 used `1e-5` for its recipe, which would have made it the family's only outlier.

## The modality contract, handled up front this time

#275 says to copy `qwen2.5-7b-dpo`, which carries **no `modality:` line**. `test_issue427_qwen35_text_modality.py` holds every Qwen3.5/3.6/3.8 recipe to an explicit `modality: text` and its family audit fails on any unregistered sibling — so a literal copy ships a recipe that escapes the decoder-only contract. Credit to @kok-o, who hit this first on #512; I hit it independently on #276. Here it was designed in rather than discovered.

## Mutations — six, all killed

| Mutation | Result |
|---|---|
| **`RecipeMeta.model` alone → wrong id** | **4 fail** |
| **YAML `base:` alone → wrong id** | **3 fail** |
| `modality: text` removed | 1 fails (family audit) |
| `format: dpo` → `auto` | 1 fails |
| `lr` 5e-6 → 1e-5 | 1 fails |
| entry deleted entirely | 7 fail |

The first two are the guard you named as deciding #512: each id surface fails **on its own**, so an edit that repairs one and forgets the other cannot pass.

## Verification

```
pytest tests/ -k "recipe"                          -> 2015 passed, 17561 deselected
ruff check src/soup_cli/ scripts/ tests/            -> All checks passed
git diff -- tests/ | grep -c "^-[^-]"              -> 7   (all count-bump sites)
```

The 7 removed test lines are the `161 → 162` bumps: four `assert len(RECIPES) == 161` and three `test_catalog_size_is_161` renames. No test deleted or weakened. `Qwen/Qwen3.5-9B` resolves on the Hub (HTTP 200); `recipes show` / `recipes use` were both run live.

## One thing worth reporting separately

While verifying I found that three existing tests fail on a **clean main** in a shell with `FORCE_COLOR` set — `test_issue330_mix_recipe_train_loadable.py` (×2) and `test_issue477_qwen38_catalog.py` — because `rich` emits ANSI into `recipes show` output that the assertions then parse as YAML. Nothing to do with this PR; I confirmed it by stashing. Filing it on its own rather than bundling.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,7 +113,7 @@ src/soup_cli/
   experiment/         - SQLite experiment tracking
   eval/               - Eval platform (custom tasks, LLM judge, human eval, leaderboard)
   migrate/            - Config migration (LLaMA-Factory, Axolotl, Unsloth)
-  recipes/            - Ready-made configs for popular models (161 recipes)
+  recipes/            - Ready-made configs for popular models (162 recipes)
   autopilot/          - Zero-config decision engine (v0.25.0)
   registry/           - Model Registry (hashing, store, diff, attach) (v0.26.0 + v0.33.0)
   cans/               - Shareable .can artifact format + run/publish orchestrator (v0.26.0 + v0.33.0)
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -152,7 +152,7 @@ soup migrate --from llamafactory config.yaml  Import config from LLaMA-Factory
 soup migrate --from axolotl config.yml        Import config from Axolotl
 soup migrate --from unsloth notebook.ipynb    Import config from Unsloth notebook
 soup migrate --from llamafactory c.yaml --dry-run  Preview without writing
-soup recipes list                             List all 161 ready-made recipes
+soup recipes list                             List all 162 ready-made recipes
 soup recipes show llama3.1-8b-sft            Print recipe YAML
 soup recipes use llama3.1-8b-sft             Copy recipe to soup.yaml
 soup recipes search "reasoning"              Search by keyword/task/size
--- a/docs/serving-and-export.md
+++ b/docs/serving-and-export.md
@@ -536,7 +536,7 @@ soup ui
 
 **Pages:**
 - **Dashboard** — view all experiment runs, loss charts, system info, multi-run comparison
-- **New Training** — create configs from templates or 161 ready-made recipes, validate, start training with live SSE log streaming and progress bar
+- **New Training** — create configs from templates or 162 ready-made recipes, validate, start training with live SSE log streaming and progress bar
 - **Data Explorer** — browse and inspect datasets (JSONL, JSON, CSV, Parquet)
 - **Model Chat** — chat with streaming responses, configurable temperature/top_p/max_tokens, system prompt, adapter selection, markdown rendering, chat export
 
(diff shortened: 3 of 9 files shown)


MakazhanAlpamys:
Merged as `d1d48e2`. Thank you — this holds the two-surface bar you set on #615, and it holds it without being asked.

Seven mutations, seven killed. The pair that decided the merge:

```
RecipeMeta.model changed, YAML base: untouched  -> 4 failed
YAML base: changed, RecipeMeta.model untouched  -> 3 failed
```

Each surface fails on its own with the other intact, so an edit that repairs one id and forgets the other cannot pass. That is exactly what a competing PR for #276 lacked, and you have now shipped it twice.

**The mutation I want to point other contributors at is one you did not have to run**: keeping the recipe but UNREGISTERING it from `EXPECTED_QWEN35_TEXT_RECIPES` fails `test_qwen35_family_audit_covers_every_current_recipe` by name. Deleting the recipe only proves the easy direction. A sibling that *exists* but escapes the family contract is the failure that actually happens, and the audit is bidirectional, which is why it catches both.

Count synchronisation is complete. `tests/test_recipe_count_is_synced.py` declares five documentation sites; all five are updated, and a repo-wide search for the old count returns nothing. Reverting either of two independent sites fails that guard.

I re-read all 7 deleted test lines: four count-assertion bumps and three test renames. Nothing weakened.

Your `lr` claim checked out independently — all eleven pre-existing DPO recipes are `lr=5e-06, dpo_beta=0.1`, so the new one makes twelve with no outlier. I did not hit the Hub, but `Qwen/Qwen3.5-9B` is byte-identical to the id already shipped in `qwen3.5-9b-sft` and `qwen3.5-9b-grpo`.

**One thing was missing and I added it myself rather than opening another round**: `changelog.d/0.73.3/632.added.md`. Your own #615 shipped one, so this is a slip rather than a pattern — but worth flagging because CI cannot catch it. `test_issue487_changelog_fragments.py` guards *staleness*, not *presence*, so a user-visible change with no fragment goes green all the way to the tag. It is in `01c1ff6`.

A note on something you reported separately and correctly: the three `FORCE_COLOR` failures on clean main reproduce exactly as you described, and #635 fixes them — that landed too.

`Refs #275`; the tracking issue stays open for the remaining variants.

MakazhanAlpamys:

Two small things, neither a defect in your work.

`test_show_and_use_recipe` read CLI output raw. It passes under `FORCE_COLOR` today by luck — the model id happens to land inside one highlight token — and `Qwen3.5-9B` contains digits that Rich's `ReprHighlighter` is entitled to wrap separately. #635's scanner deliberately ignores single-token assertions, so it would not have caught the day that luck ran out. It now goes through `strip_ansi`. **No mutation kill is claimed for this one** — it is prophylactic, and I would rather say so than invent evidence for it.

I also renamed `test_total_recipe_count_is_158`, which asserts 162. The count had drifted 116 -> 134 -> 137 -> 158 -> 162 without the name following, so the first thing a reader saw was the one thing that was false. You were right not to touch it in a recipe PR.

Issue #633, opened 2026-09-01
Three tests fail on a clean checkout when FORCE_COLOR is set: recipes show output is parsed as YAML, ANSI and all


Srinivasan8888:
Found while verifying #632. Not caused by it — confirmed by stashing and re-running on clean `main`.

## Reproduction

```
$ echo $FORCE_COLOR
3
$ pytest tests/test_issue330_mix_recipe_train_loadable.py tests/test_issue477_qwen38_catalog.py -q
3 failed, 9 passed

$ NO_COLOR=1 FORCE_COLOR= TERM=dumb pytest <same two files> -q
12 passed
```

Failing:

```
test_issue330_mix_recipe_train_loadable.py::test_apply_cli_quotes_path_needing_quoting
test_issue330_mix_recipe_train_loadable.py::test_apply_handles_new_multi_dataset_recipe
    yaml.reader.ReaderError: unacceptable character #x001b: special characters are not allowed

test_issue477_qwen38_catalog.py::test_qwen38_27b_recipe_show_and_use
    AssertionError: assert 'modality: text' in '╭─ \x1b[1;32mqwen3.8-27b-sft\x1b[0m ...'
```

## Cause

`soup recipes show` renders the YAML through `rich` with Pygments highlighting. With colour enabled, escape sequences land **inside** the rendered YAML — between `modality` and `: text`, and throughout the body. These tests then assert on that raw output, and #330's two feed it to `yaml.safe_load`, which rejects `\x1b` outright.

## Why nobody has hit it in CI

GitHub runners have no TTY and don't set `FORCE_COLOR`, so `rich` disables colour and the assertions see plain text. It only appears for a contributor whose terminal forces colour — which several modern setups do by default. The symptom is three red tests on a **fresh clone with no local changes**, which reads as "I broke something" and costs a bisect to disprove. That is how it cost me one.

## Why the existing guard does not cover it

`tests/test_cli_help_assertions_are_ansi_safe.py` exists for exactly this failure mode and says so:

> Rich splits flag names with ANSI escapes on Linux/macOS, so this passes on Windows and turns CI red everywhere else (it has, four times).

But its scanner is scoped to `--help` by construction:

```python
for path in sorted(TESTS_DIR.glob("test_*.py")):
    source = path.read_text(...)
    if "--help" not in source:
        continue          # <- recipes show / mix apply never get scanned
```

[...]

Issue #477, opened 2026-08-22
Add Qwen 3.8 27B to the catalog


Sir-Lysander:
Can you add the new Qwen 3.8 27B from Alibaba to the catalog? [https://huggingface.co/Qwen/Qwen3.8-27B](https://huggingface.co/Qwen/Qwen3.8-27B)

MakazhanAlpamys:
Thanks — and you were right, I was wrong. On #472 I said Alibaba had not released a 3.8; it exists and is not obscure:

```
id            : Qwen/Qwen3.8-27B
createdAt     : 2026-08-05
downloads     : 2,090,699      likes: 12,019
pipeline_tag  : image-text-to-text
model_type    : qwen3_5
architectures : ['Qwen3_5ForConditionalGeneration']
```

**But this is not a one-line catalog addition, and the reason is a defect we already have open.**

## It is a vision-language model, and that is exactly issue #427

`pipeline_tag: image-text-to-text` and `Qwen3_5ForConditionalGeneration` mean this is multimodal. Soup's `modality` defaults to `text` when a recipe leaves it unset — and **#427 records that all nine shipped Qwen3.5 recipes do exactly that**, while every checkpoint in the family reports a multimodal architecture.

So adding `qwen3.8-27b-sft` with the obvious defaults would make it the **tenth instance of a known open defect**, not a new recipe. I would rather not do that to you, or to whoever debugs it later.

Note the `model_type` is literally `qwen3_5` — the 3.8 checkpoint reports the same architecture family as the 3.5 line, so it inherits #427 wholesale rather than being a separate case.

## What has to be decided first

#427's open question is whether those recipes should declare `modality: vision`, or whether the text-only decoder is a supported way to use these checkpoints. Until that is answered, a new recipe would be guessing:

- **If `modality: vision`** — the recipe needs `image_dir`, a vision-shaped data format (`llava` / `sharegpt4v`), and the vision quant path. A user copying it for text-only SFT would then be surprised.
- **If `modality: text`** — someone should establish that loading `Qwen3_5ForConditionalGeneration` as a text decoder actually works, rather than assuming it, since #427 exists precisely because nobody has.

There is a related edge in layer streaming: the alias table admits `qwen3_5_moe` and `qwen3_5_moe_text`, but **not plain `qwen3_5`**. So `stream_layers: true` on this checkpoint is refused today — correctly, since nothing has established parity for that decoder.

## What I suggest

[...]

PR #635, opened 2026-09-01
test(ansi): normalise syntax-highlighted CLI output and guard the pattern (#633)

Srinivasan8888:
Fixes the three tests that fail on a **clean checkout** in any shell with `FORCE_COLOR` set, and extends the existing ANSI guard to the subcommands it never reached.

`Refs #633`.

## The bug

`soup recipes show` and `soup data mix --apply` render YAML through Rich with Pygments, which emits escapes **between the tokens of one logical line**. `modality`, `:` and `text` are three tokens, so `"modality: text"` is not a substring of the output, and `yaml.safe_load` rejects `\x1b` outright.

CI never sees it — runners have no TTY — so the symptom is three red tests on a fresh clone with no local changes, which reads as "I broke something". It cost me a stash-and-rerun to disprove while verifying #632.

This is the exact failure mode `test_cli_help_assertions_are_ansi_safe.py` was written for, after it "turned CI red everywhere else (it has, four times)". That scanner only inspects functions containing `--help`, so no other subcommand was covered.

Worth noting the #330 pair had already defended against the *other* rendering hazard — Rich wrapping long paths mid-token, via `CliRunner(env={"COLUMNS": "300"})`. Colour was the one that got missed.

## Scope, measured rather than guessed

My first instinct was "flag every unnormalised raw-output assertion". That would have landed **217 pre-existing offenders**, and a guard arriving with 217 violations is one people delete rather than obey.

Scoping to **syntax-highlighted commands** and to **literals containing a colon** — the YAML structure Pygments actually tokenises — flags exactly the assertions that break:

| population | count |
|---|---|
| any unnormalised raw-output assertion | 217 |
| raw output fed to a parser | 4 |
| **highlighted output + colon literal (the guard's scope)** | **3, all genuine** |

A plain message like `"validation failed"` is wrapped in escapes rather than split by them, and a bare model id is a single token — which is why neighbouring assertions in the same files kept passing.

## Two defects in my own guard, found by running it

- **It missed the #477 case entirely.** `_RAW_OUTPUT_RE` anchors on `result.output`, and `\b` never fires before `result` in `show_result.output` because `_` is a word character. The new scan uses a broader pattern.

  **The existing `--help` guard has the same blind spot.** I measured it before claiming a second bug: **0** current assertions hit it, so it is latent rather than live. Reported, not changed here — widening `_RAW_OUTPUT_RE` alters the older guard's behaviour and belongs in its own PR if you want it.

- **It flagged `assert "train:" in compact, result.output`**, where the raw read is in the failure *message* — which is good practice, not a defect. Only the assert expression is scanned now.

## Mutations — four, all killed

| Mutation | Result |
|---|---|
| `strip_ansi()` becomes a no-op | 6 fail |
| scanner returns nothing | 2 fail |
| the #477 fix reverted | 2 fail |
| **colon predicate widened back to any-space** | **1 fails** |

The last is the one I care about: widening re-flags the legitimate `"validation failed"` assertion, so the *narrowing itself* is pinned and cannot drift back into noise.

## Verified in both directions

A fix that only passes in one colour mode is not a fix:

```
FORCE_COLOR=3 TERM=xterm-256color   -> 15 passed
NO_COLOR=1 FORCE_COLOR= TERM=dumb   -> 14 passed, 1 skipped
```

The skip is the control, which correctly skips when there is no highlighting to normalise — and it skips on the **hazard's absence**, not on "no escapes anywhere", because Rich colours the panel border even when it does not highlight the body. I got that wrong first and it passed vacuously.

```
full suite    -> 19468 passed, 103 skipped, coverage 82.73% (floor 77%)
ruff          -> All checks passed
git diff -- tests/ | grep -c "^-[^-]"  -> 5
```

All five removed lines are rewrites in place: the conftest docstring, the three patched read sites, and the #477 assertion. No test deleted or weakened.

## Deliberately not in this PR

**38 test files define their own `_ANSI_RE` / `_strip_ansi` copy.** `strip_ansi` now lives in `tests/conftest.py` and new code should import it, but consolidating 38 files inside a bug fix is scope creep of exactly the kind splitting #619 out of #581 avoided. Happy to do it as its own PR — duplicated helpers are how this repo got #372, #392 and #424.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,21 @@
-"""Shared test fixtures."""
+"""Shared pytest fixtures and helpers."""
 
 import json
+import re
 from pathlib import Path
 
 import pytest
+
+#: Rich/Pygments emit SGR escapes *between* the tokens of one logical line, so a
+#: multi-token substring like "modality: text" is absent from raw output and
+#: yaml.safe_load rejects \x1b outright (#633). 38 test files had grown their
+#: own copy of this regex; new code should import this one.
+_ANSI_ESCAPE = re.compile(r"\x1b\[[0-9;]*m")
+
+
+def strip_ansi(text: "str | None") -> str:
+    """Return ``text`` with SGR escape sequences removed."""
+    return _ANSI_ESCAPE.sub("", text or "")
 
 
 @pytest.fixture(autouse=True)
--- a/tests/test_cli_help_assertions_are_ansi_safe.py
+++ b/tests/test_cli_help_assertions_are_ansi_safe.py
@@ -252,3 +252,149 @@ def test_help_renders():
 
     def test_unparseable_source_does_not_explode(self):
         assert find_raw_help_assertions("def broken(:\n") == []
+
+
+# ---------------------------------------------------------------------------
+# #633 — the same hazard, on subcommands the `--help` scan never reaches.
+# ---------------------------------------------------------------------------
+
+#: Commands whose output is **syntax highlighted**. Rich renders these through
+#: Pygments, which emits SGR escapes *between* the tokens of one logical line:
+#: ``modality``, ``:`` and ``text`` are three tokens, so ``"modality: text"`` is
+#: not a substring of the rendered output and ``yaml.safe_load`` rejects
+#: ``\x1b``. Plain ``console.print`` output is not affected the same way -- an
+#: unstyled message is wrapped in escapes rather than split by them -- which is
+#: why only these commands are scanned and the suite's ~200 other raw-output
+#: assertions are correctly left alone.
+_HIGHLIGHTED_INVOCATIONS = (
+    re.compile(r'"recipes"\s*,\s*"show"'),
+    re.compile(r"'recipes'\s*,\s*'show'"),
+    re.compile(r'"mix"\s*,\s*"--(apply|optimize)"'),
+    re.compile(r"'mix'\s*,\s*'--(apply|optimize)'"),
+)
+
+_PARSES_OUTPUT = re.compile(r"(yaml\.safe_load|json\.loads)\s*\(")
+
+#: Broader than ``_RAW_OUTPUT_RE``, which anchors on the exact name
+#: ``result.output`` and therefore cannot match ``show_result.output`` -- ``_``
+#: is a word character, so ``\b`` never fires before ``result`` there. Real
+#: tests name their results ``show_result`` / ``use_result`` all the time.
+_ANY_RAW_OUTPUT_RE = re.compile(r"\b[A-Za-z_][A-Za-z0-9_]*\.(output|stdout)\b|readouterr\(\)")
+_STRING_LITERAL = re.compile(r'"([^"]{2,})"' + r"|'([^']{2,})'")
+
+
+def _is_multi_token(literal: str) -> bool:
+    """A literal that Pygments would split across tokens.
+
+    The hazard is YAML *structure*. ``modality: text`` is three tokens -- key,
+    punctuation, value -- with escapes between them, and ``train:`` is two. A
+    bare model id (``Qwen/Qwen3.8-27B``) is one token and survives raw, which is
+    why neighbouring assertions in the same file kept passing.
+
+    A plain message such as ``"validation failed"`` is also safe: Rich wraps an
+    unstyled string in escapes rather than splitting it, so requiring a colon
+    keeps the guard on the assertions that actually break rather than every
+    substring containing a space.
+    """
+    return ":" in literal
+
+
+def _assert_expression(line: str) -> str:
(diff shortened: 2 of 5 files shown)


MakazhanAlpamys:
Merged as `3112117`. Thank you — this is the first PR in this class that ships the *ratchet* rather than only the repair, which is what the project needed after four occurrences.

Reproduced the bug on main before merging, exactly as filed:

```
FORCE_COLOR=3 TERM=xterm-256color -> 3 failed, 9 passed
  test_apply_cli_quotes_path_needing_quoting   yaml.reader.ReaderError: #x001b
  test_apply_handles_new_multi_dataset_recipe  same
  test_qwen38_27b_recipe_show_and_use          'modality: text' in '... \x1b[1;32m ...'
```

**The evidence that decided the merge is mutation C run in a colour-free environment.** Reverting the #477 repair fails the static guard under `NO_COLOR=1 TERM=dumb` — that is what real CI sees — so the scanner catches the regression on runners that *cannot observe the symptom*. For a hazard that only manifests on some terminals, a guard that only works on those terminals would be nearly worthless. This one is not.

Full mutation set, all killed:

| mutation | result |
|---|---|
| `strip_ansi()` -> no-op | 7 failed |
| `find_unsafe_highlighted_assertions()` -> `[]` | 2 failed (its own self-tests) |
| revert the #477 fix | 2 failed, including under `NO_COLOR=1` |
| `_is_multi_token` widened | 1 failed (re-flags a real site) |
| `_HIGHLIGHTED_INVOCATIONS = ()` | 2 failed |

And it correctly flags the **v0.73.2 insufficient fix** — `"".join(result.output.split())` followed by a substring assertion, a whitespace collapse mistaken for an ANSI strip. That is the exact commit shape that reddened CI before, and a guard that missed it would have been decorative.

All 5 deleted lines are in-place rewrites wrapping existing assertions in `strip_ansi(...)`. Nothing weakened.

**Two things I closed myself in `01c1ff6` rather than sending back**, since neither was worth a round-trip:

1. `_ANY_RAW_OUTPUT_RE` — the widening that makes `show_result.output` reachable — was the one behaviour in the PR with no test behind it, and it is the behaviour that makes the guard reach the real #477 case. Every fixture in `TestTheGuardCoversHighlightedCommands` names its variable `result`, so reverting the widening left the suite green while the scanner went blind to the exact spelling this issue was filed about. Measured: with the widening reverted, the new `test_scanner_flags_a_non_result_variable_name` fails alone (1 failed, 11 passed); reverted, 12 passed.
2. `soup migrate` — the third `Syntax(...)` site in `src/` (`commands/migrate.py:121`) — is now in `_HIGHLIGHTED_INVOCATIONS`. Latent rather than live: no assertion on its output is multi-token today, so it flags nothing and the repo-wide scan still reports zero offenders. But a guard covering two of three highlighted commands is an invitation to occurrence number five.

I also corrected the mechanism comment while there, because it is the kind of imprecision that makes the *next* person scope a guard too narrowly. It attributed the hazard to Pygments across the board; for `data mix --apply` that is wrong — `commands/data_mix.py` emits plain `console.print` with no `Syntax`, and the escapes come from Rich's default `ReprHighlighter`. Your scanner keys on the invocation rather than the mechanism, so the behaviour was already right; only the rationale was off.

Two follow-ups I am **not** asking you to do, recorded so they are not lost: the first-character-quote heuristic in the fixture skip will produce a false negative on a formatter-wrapped `assert (\n    "x" in result.output\n)`, which ruff will eventually generate given the 100-char limit; excluding `ast.Constant` line ranges would be robust. Open an issue if you want it, or leave it.

Closes #633. A changelog fragment was missing and is in `01c1ff6` as well — tests-only, but every contributor with `FORCE_COLOR` hits this on a clean checkout, so it is user-visible in the way that matters.

MakazhanAlpamys:

I closed the false negative I flagged in the review — the assertion a formatter has wrapped — and it took two changes rather than the one I described, which is worth knowing if you touch this scanner again.

Skipping any line beginning with a quote also skips the continuation line of `assert (\n "modality: text" in result.output\n)`. Replacing it with a multi-line-constant span check **moved** the false negative instead of closing it: the continuation line does not start with `assert`, so it was dropped one branch lower. Both are fixed now.

The third change is the one I would not have predicted: the assert span has to come from `node.test` and never `node.msg`. Taking the whole statement flagged `test_recipes_v031.py:468` — a correct assertion that reads raw output only in its failure message, which `_assert_expression` already handles on the single-line form.

```
literal-span rule reverted        -> test_scanner_flags_an_assertion_the_formatter_wrapped
continuation lines not counted    -> same test
span widened to whole statement   -> test_the_suite_has_no_unsafe_highlighted_assertions
reverted                          -> 23 passed
```

PR #637, opened 2026-09-01
test(recipes): snapshot every recipe's resolved config against schema drift (#621)

### LINKED ISSUE #621: Recipe values that duplicate a schema default are unguarded in both directions: nothing pins them, and nothing detects drift
Filed as its own issue rather than decided inside a PR, per @MakazhanAlpamys on #276 — the same handling #442 and #443 got when they surfaced during someone else's review.

## Where it came from

Mutation-testing #615 (`qwen3.5-35b-a3b-dpo`) turned up a survivor: deleting `dpo_beta: 0.1` from the recipe killed no test. That is correct behaviour, not a coverage gap — `dpo_beta` defaults to `0.1` at `src/soup_cli/config/schema.py:1237`, so removing the line yields a byte-identical `SoupConfig`. It is an *equivalent* mutation, and @MakazhanAlpamys confirmed it independently (`a.model_dump() == b.model_dump()`) and rightly said not to manufacture a test for it, since killing it would mean asserting on YAML source text instead of loaded behaviour.

What was left open was the general question. I measured it rather than guessing at its size.

## It is not one recipe or one field

Across the 159 recipes in the catalog, **20 distinct fields are declared with a value identical to their schema default**:

```
quantization                 136 recipes   e.g. llama3.1-8b-sft
batch_size                   127 recipes   e.g. llama3.1-8b-sft
epochs                       123 recipes   e.g. llama3.1-8b-sft
num_generations               21 recipes   e.g. llama3.1-8b-grpo
reward_fn                     18 recipes   e.g. llama3.1-8b-grpo
grpo_beta                     16 recipes   e.g. llama3.1-8b-grpo
moe_aux_loss_coeff            14 recipes   e.g. qwen3-30b-a3b-sft
dpo_beta                      11 recipes   e.g. llama3.1-8b-dpo
asr_task                       3 recipes   e.g. whisper-tiny-asr
gradient_accumulation_steps    2 recipes   e.g. qwen3-8b-tools
kto_beta / orpo_beta / simpo_gamma / ipo_tau                    1 each
ppo_epochs / ppo_clip_ratio / ppo_kl_penalty                    1 each
online_dpo_loss_type / online_dpo_max_new_tokens                1 each
embedding_margin                                                1
```

Reproduce:

```python
import yaml
from soup_cli.recipes.catalog import RECIPES
from soup_cli.config.schema import TrainingConfig
for n, r in RECIPES.items():
    tr = (yaml.safe_load(r.yaml_str) or {}).get("training") or {}
    for k, v in tr.items():
        f = TrainingConfig.model_fields.get(k)
        if f is not None and f.default is not None and f.default == v:
            print(k, n)
```

## The actual exposure, in both directions

**1. Nothing pins the declaration.** Any of those ~460 lines can be deleted today and no test changes colour. Most are harmless. Some are not: `quantization: 4bit` on a 122B recipe is load-bearing documentation for whoever reads `soup recipes show`, and it silently disappearing from a recipe is a user-visible change no guard would catch.

**2. Nothing detects drift the other way.** If a schema default is ever changed — say `epochs` 3 → 1 — the 123 recipes that declare `epochs: 3` keep their behaviour, and the 36 that omit it silently change. That divergence is invisible: no test compares a recipe's *rendered* config against what it rendered before, so a one-line schema edit can quietly retune part of the catalog. `test_catalog_size_is_160` counts recipes; nothing pins what any of them resolves to.

The second is the one I would actually worry about. The first is mostly cosmetic; the second is a defaults change reaching users through recipes nobody edited.

## What I am **not** proposing

Asserting on YAML source text. A test that greps `"dpo_beta: 0.1" in recipe.yaml_str` pins the string and not the behaviour, and that is a shape this project has removed from its own suite before. If the answer here is "do nothing", that is a legitimate answer and this issue can close — the measurement is worth having recorded either way.

## Options, roughly by cost

1. **Nothing.** Accept that these lines are documentation and unguarded. Cheapest; leaves exposure 2 open.
2. **Snapshot the resolved configs.** One test that loads every recipe and compares `model_dump()` against a committed fixture. Catches exposure 2 exactly — a schema-default change that moves any recipe fails loudly with the diff — and needs a deliberate fixture update to accept, which is the point. Cost: one fixture file, and a regeneration step in the release flow.
3. **Assert declared == default where it is intended**, per field. More precise, much more bookkeeping, and it re-raises the "pin the source" objection in a different costume.

I lean to **2** — it pins behaviour rather than text, it is one test rather than 460 assertions, and it fails in the direction that actually reaches users.

Happy to implement whichever you pick, or to close this if the answer is 1. Not claiming it until you have said which — filing the question, not the fix.

SID-6921:
## Summary

Closes #621.

162 recipes declare training fields that duplicate the schema default (`dpo_beta: 0.1`, `epochs: 3`, etc.) across 20 distinct fields. Deleting one of those lines is an equivalent mutation — the resolved `SoupConfig` is byte-identical either way — so this doesn't manufacture a test to kill that; per the issue, that's the wrong thing to pin.

The real exposure is the other direction: nothing pins what a recipe *resolves to*, so a schema-default change silently retunes every recipe that relies on that default, with no red test anywhere. Went with option 2 from the issue (snapshot the resolved configs), per the reporter's own recommendation.

## What's here

- `tests/fixtures/recipe_config_snapshots.json` — `load_config_from_string(...).model_dump(mode="json")` for all 162 recipes, committed.
- `scripts/generate_recipe_snapshot.py` — regenerates the fixture. Explicit, reviewable step, never automatic.
- `tests/test_issue621_recipe_snapshot.py`:
  - one test per recipe comparing its resolved config against the committed snapshot, with a field-level diff on mismatch (not a giant raw-dict dump)
  - a completeness check that the fixture's recipe names exactly match the current catalog (catches an added/removed recipe without regenerating)
  - two non-vacuity controls proving the diff logic actually discriminates
- One changelog fragment.

The comparison is between parsed Python objects (`json.load` output vs. `model_dump()`), not raw file bytes, so it isn't sensitive to the CRLF/LF drift a byte-exact fixture hit during #580's review.

## Verification

- `ruff check src/soup_cli/ scripts/ tests/` — clean.
- `mypy` on both new files — clean.
- 165 tests passing (162 recipe checks + the completeness check + 2 non-vacuity controls).
- **Mutation, as the issue asked for**: changed the `dpo_beta` schema default from `0.1` to `0.2` in `schema.py`, reran the suite — 149 of 162 recipes failed (every recipe that doesn't explicitly pin `dpo_beta` in its own YAML), which is the exact blast radius the issue measured. Reverted, all 165 passing again.
- Deliberately did not touch the ~460 lines that duplicate a default value (exposure 1) — out of scope per the issue.

Branch was rebased onto current `main` (162-recipe catalog, post-#625) before this was written, so the fixture reflects the current catalog rather than the 159/160-recipe count at filing time.


MakazhanAlpamys (approved):
Reviewed at `37e0471`. **Approving in principle — holding only on CI**, which is queued rather than red (1 in progress, 12 queued at the time of writing). I have answered the option question on #621, which I should have answered before you started; the delay was mine.

## The measurement that decides it

I did not take the PR body's word for the blast radius. Changing **one** schema default — `dpo_beta` at `src/soup_cli/config/schema.py:1238`, `0.1` → `0.2` — turns **149 named recipe tests red** on this branch and turns **nothing** red on `main`. That is exposure 2 from the issue, caught exactly, and it is the direction that reaches users.

Full mutation set, all run with `PYTHONPATH` pinned to the worktree:

| mutation | result |
|---|---|
| delete `dpo_beta: 0.1` from `catalog.py:107` (duplicates the default) | **165 passed — correct**, exposure 1 deliberately not pinned |
| `catalog.py:107` `dpo_beta` 0.1 → 0.3 | 3 failed, incl. `test_recipe_resolves_to_its_snapshot[llama3.1-8b-dpo]` |
| `catalog.py:103` `lora.r` 16 → 8 (**not** a default) | 3 failed — sensitive to recipe edits too, not only schema ones |
| schema default `dpo_beta` 0.1 → 0.2 | **149 failed** |
| new schema field added | 164 failed, 1 passed |
| `_diff` neutered to `return {}` | 1 failed — the control fires |
| fixture replaced with `{}` | 3 failed — **cannot silently pass on an empty fixture** |
| fixture deleted | collection ERROR, loud |
| new recipe added without regenerating | 1 failed: `missing from fixture: ['zz-new-recipe']` |

The first row is the one I checked hardest, because getting it wrong would have meant the guard quietly answered a question the issue explicitly excluded. It does not.

## The thing I expected to block on, and did not

A 1.55 MB / 49,586-line committed fixture is exactly the shape that cost #613 a round: `core.autocrlf` rewriting a newline underneath a byte-level commitment, red on `windows-latest` while macOS stays green. **It does not apply here, and you pre-empted it in the docstring before I could raise it.** Verified rather than assumed: this box has `core.autocrlf=true`, the worktree file is 49,586 CRLF lines, the blob is 49,586 LF lines with zero CRLF, and the comparison at `tests/test_issue621_recipe_snapshot.py:38` is `json.loads(read_text(encoding="utf-8"))` against a parsed dict — no byte, SHA, or length commitment anywhere in the file. `check-added-large-files` is `--maxkb=5000`, so this sits at ~31% of the limit, and `pyproject.toml:222` packages `src/soup_cli` only, so it does not ship in the wheel.

Regeneration is byte-exact: running `scripts/generate_recipe_snapshot.py` leaves `git status --porcelain` empty. Deterministic via `sorted(RECIPES.items())` + `sort_keys=True`. And `sys.path.insert(0, _REPO_ROOT / "src")` in the script is a detail I want to point other contributors at — it beats a shadowing editable install, which is a trap that has produced false review results in this repo more than once.

Not stale, either: fixture 162 recipes, catalog 162, `set(fixture) == set(RECIPES)`. `qwen3.5-9b-dpo` from #632 landed before your branch point and is in there.

## The failure output is the best part

```
AssertionError: recipe 'baichuan-sft' no longer resolves to its committed snapshot.
If this is a deliberate change, review the diff and regenerate with
scripts/generate_recipe_snapshot.py. Differing fields (field: (snapshot, resolved)):
{'training.dpo_beta': (0.1, 0.2)}
```

Names the recipe, the dotted path, both values, and the remedy — no 49k dump, which is what usually makes a golden test the first thing someone deletes.

## On the Copilot review above

Its point about parametrizing from fixture names rather than the catalog is reasonable in the abstract, and I checked it by execution: adding a recipe without regenerating fails `test_fixture_and_catalog_have_the_same_recipe_names` by name and lists the missing recipe, so the case is covered rather than silent. Its sentinel point is fair and cheap. **Its third point I disagree with and am not asking you to act on** — deriving the control's expected values from the fixture would couple the control to the data it exists to check, and a control that reads its answer from the thing under test is weaker, not more maintainable. Worth a comment in the test saying why the value is pinned, nothing more. Note also that Copilot reviewed 3 of the 4 files at "Lite" effort and did not look at the fixture at all, i.e. it did not examine the one thing that could have blocked this.

## What I am changing myself after merge, rather than sending back

None of these is worth another round-trip, and all of them are mine to have specified up front:

1. **A test that `build_snapshot()` equals the committed fixture.** Right now `model_dump(mode="json")` over `load_config_from_string(...)` is written twice — `tests/test_issue621_recipe_snapshot.py:73` and `scripts/generate_recipe_snapshot.py:27` — and nothing asserts they agree. If either drifts, the fixture becomes unregenerable and nobody learns until a release. Two duplicated sources of truth is the exact shape behind #372 and #392 in this repo. It also gives the script its only coverage, since `--cov=soup_cli` does not measure `scripts/`.
2. **Name the regeneration script in the completeness test's message** (`:81-83`). That is the failure every future recipe PR trips — roughly four per release — and it is currently the one message that does not say what to do about it.
3. **Decouple the two non-vacuity controls from the live catalog** so they exercise `_diff` on synthetic dicts. Today an unrelated schema addition reddens the controls alongside the real tests, which makes "did the guard break, or did the data move?" unanswerable from the log.
4. **`.gitattributes` with `tests/fixtures/recipe_config_snapshots.json text eol=lf`.** Regeneration is byte-stable on `autocrlf=true`; on a Windows box with `autocrlf=false` a contributor would see a spurious 49,586-line diff.
5. Documenting the regeneration step in `CONTRIBUTING.md` and the release checklist — the issue named that as part of option 2's cost and it is the maintainer's half.

If you would rather do any of them yourself, say so and I will leave them; otherwise they land in a follow-up commit crediting you.

## Recorded, not asked for

Every new `TrainingConfig` field reddens all 162 recipe tests. This project adds schema fields most releases, so "162 red → regenerate → green" becomes routine, and at that point an additive change looks procedurally identical to a value change. Your message *does* distinguish them (`('<missing>', False)` versus `(0.1, 0.2)`), so it is recoverable by a careful reader. I am filing the delta-encoded-fixture idea as a follow-up rather than holding this: it would be ~15× smaller and would report a schema-default change as one failure instead of 149, but it is a refinement and not what #621 asked for.

Zero deleted test lines, ruff clean, `test_cli_startup_is_light` 34 passed, changelog fragment in the right directory with the credit format exactly right. Nice second PR.

SID-6921:
Pushed all 5 in 07be732.

1. Test now imports and calls generate_recipe_snapshot.build_snapshot() directly (cached, so the catalog still resolves once per session) instead of reimplementing the same model_dump line. No duplication left to drift.
2. Both the completeness check and the per-recipe failure message name the regen script now.
3. Per-recipe tests parametrize from the live catalog instead of the fixture's keys, with an explicit "missing from fixture" assertion before comparing. Checked the reverse case too — a stale fixture entry for a recipe removed from the catalog now produces exactly one clear failure in the completeness check, nothing silently skipped, no KeyError anywhere.
4. The two non-vacuity controls run on synthetic dicts now, not the live dpo recipe.
5. .gitattributes pins the fixture to text eol=lf, and both CONTRIBUTING.md spots (new recipe steps, release checklist) document the regen step.

Also tightened build_snapshot()'s return type since the test now touches it directly and mypy had something to say about it.

Re-ran both of your mutations against the refactor (schema default swap, recipe added without regenerating) plus the stale-entry case above. 165 passing, ruff and mypy clean.

MakazhanAlpamys:
Re-reviewed `07be732`. **All five items verified by execution, four fully closed and one partially.** Approving stands; holding only on CI (12 queued, 1 green).

Turnaround on five items in one push is the fastest I have seen on this repo, and none of them was done to the letter while missing the point — which is the usual failure mode when a reviewer hands over a numbered list.

**1. Single source of truth — closed.** Exactly one executable `model_dump` in the repo now, at `scripts/generate_recipe_snapshot.py:24`, imported by the test. I checked it the only way that means anything — mutating `build_snapshot()` to add `exclude_none=True` gives **162 failed, 3 passed**. The script can no longer drift from the fixture unnoticed, which was the whole point. Also verified the import is cwd-independent by running pytest from another directory: 165 passed.

**2. Both messages name the script — closed.** Triggered both for real:

```
missing from fixture: ['zzz-mutation-probe']; stale in fixture: []; regenerate with scripts/generate_recipe_snapshot.py
recipe 'baichuan-sft' no longer resolves ... regenerate with scripts/generate_recipe_snapshot.py. Differing fields: {'training.dpo_beta': (0.1, 0.2)}
```

**3. Parametrized from the live catalog — closed in both directions**, and the reverse case is the one I would have forgotten to check. A recipe added without regenerating gives the named guard rather than a `KeyError`; a recipe *removed* while the fixture keeps its entry gives `stale in fixture: ['whisper-tiny-asr']`. The parametrized count dropping 164 -> 163 in that run is the incidental proof that parametrization now tracks the catalog rather than the fixture.

**4. Controls decoupled — closed, and this is the item I care about most.** Added a new schema field and got exactly the property: **162 real tests red, both controls green.** Same under the `dpo_beta` mutation (149 red, controls green) and under the `build_snapshot` mutation. "Did the guard break, or did the data move?" is now answerable from the log, which it was not before.

Two of the properties that already passed came back **stronger** rather than merely preserved: an empty fixture went 3 failures -> 163, and a deleted fixture went from a collection error to 163 named failures. Zero deleted test lines across the whole PR.

**5. LF normalization — partially closed, and this is the note.** You took `.gitattributes`, which fully protects the committed content: the normalized blob hash is identical and `git diff --quiet` exits 0, so no CRLF can reach the index. But the stated acceptance criterion — regeneration byte-exact, `git status --porcelain` empty — does not hold on Windows, because `scripts/generate_recipe_snapshot.py:31` uses `Path.write_text(..., encoding="utf-8")` with no `newline=`, and CPython translates `\n` -> `\r\n`:

```
md5 before      a88df5e844268c6d4704873675f38710
after regen     43db817e81e3721aff3050889cfc5aee
git status      M tests/fixtures/recipe_config_snapshots.json     <- persists
CRLF count      49586
```

No correctness risk — the tests are byte-insensitive and the index is clean. The cost is that a Windows contributor who regenerates gets a permanent phantom `M` plus a CRLF warning on every git command, and **CI cannot catch this**, because the divergence is Windows-only by construction. That is exactly why it wants the belt-and-braces line rather than trusting the runners:

```python
_FIXTURE_PATH.write_text(..., encoding="utf-8", newline="\n")
```

Push it if you want it closed under your name; otherwise I will add it after the merge and say so here. Either is fine — I am not holding the PR for one line.

One non-blocking observation while I was in there: `_load_fixture()` is uncached at `:38` and re-parses the 1.5 MB JSON once per parametrized test (162x), measured at about 3 s for the file. `lru_cache`, mirroring what `_live_snapshots()` already does, if you feel like it.

Merging when CI goes green. #621 is answered and closed — see my comment there for the decision and for the credit split, since @Srinivasan8888's measurement is what made the option list decidable.

SID-6921:
Both landed in be911f4.

Item 5: added newline="\n" to the write_text call. Regenerated on a Windows box after — git status --porcelain is now empty, matches the stated criterion directly instead of relying on .gitattributes to paper over it.

Also cached _load_fixture() with lru_cache per your note. 165 passing, ~1.2s now.

MakazhanAlpamys:
Merged as `e83b992`. Thank you — and the turnaround on the five follow-ups, then the sixth on top, is the fastest this repo has seen.

I verified the last item myself before merging rather than taking the push on trust:

```
$ py -3.10 scripts/generate_recipe_snapshot.py
Wrote 162 recipe snapshots to ...
$ git status --porcelain
(empty)
$ CRLF: 0   LF: 49586
```

Byte-exact on Windows now, which the `.gitattributes` alone could not give you — it protected the index but left a permanent phantom `M` for any Windows contributor who regenerated. Belt and braces is the right answer for a divergence CI structurally cannot see.

The property I care about most is the one you closed in item 4, and it is worth naming for anyone reading this later: your non-vacuity controls exercise the diff on synthetic dicts, so when a new schema field reddens all 162 recipe tests, **both controls stay green**. That is what makes "did the guard break, or did the data move?" answerable from a CI log rather than by re-deriving it by hand. Most golden-fixture suites get this wrong and nobody notices until the day it matters.

Also recorded, because it decides whether the whole instrument is honest: deleting a recipe line that duplicates a schema default still kills **nothing**, exactly as #621 specified. A guard that had quietly started pinning YAML source text would have been a different feature wearing this one's name.

Nine mutations behaved as specified, including the two directions I would most expect to be missed — a recipe *removed* while the fixture keeps its entry (`stale in fixture: [...]`), and `build_snapshot()` itself being altered (162 failures, so the script can no longer drift from the fixture unseen).

**Credit is split in the merge commit and in `CONTRIBUTORS.md`**: the analysis, the measurement of the 20 fields across ~460 lines, and the option list are @Srinivasan8888's in #621 — they offered to implement whichever option I chose and waited two days on an answer I did not give, which is how this ended up split between you. That is on me, not on either of you, and I have said so on the issue.

One follow-up is already open as #640 — good, and note that #638 lists seven properties that must not regress, all of which I have now verified working here. Whatever replaces this fixture has to keep all seven.

PR #641, opened 2026-09-02
fix(deps): declare the real torch floor and stop duplicating it in doctor (#636)

### LINKED ISSUE #636: [train] declares torch>=2.3.0 but transformers>=5.16.1 already forces torch>=2.5 — the floor is a fiction
`pyproject.toml [train]` now reads `torch>=2.3.0` (raised from 2.0.0 in #603 for a Qwen4-only uint32 need). That floor cannot be the binding constraint, and stating it invites people to trust it.

The same extra requires `transformers>=5.16.1`, and transformers 5.16.1's own torch extra requires **`torch>=2.5`**:

```
$ python -c "import urllib.request,json; d=json.load(urllib.request.urlopen('https://pypi.org/pypi/transformers/5.16.1/json')); print([r for r in d['info']['requires_dist'] if r.startswith('torch>=')])"
['torch>=2.5; extra == "torch"', ...]
```

So no user can have a working `pip install 'soup-cli[train]'` on torch 2.0-2.4 regardless of what Soup declares. Raising 2.0.0 -> 2.3.0 excluded nobody, which is why #603 merged with it; but 2.3.0 is still below the real floor, so the number in `pyproject.toml` is decorative in the same way the old one was.

Why it matters rather than being cosmetic: `utils/deepspeed.py`, `utils/fsdp.py` and `utils/layer_stream.py` all branch on torch capabilities, `soup doctor` reports the floor to users as though it were checked, and a contributor reading `torch>=2.3.0` will reasonably write a compatibility shim for a version combination that cannot be installed. Dead compatibility code is worse than none, because it looks tested.

**Proposed fix**

- [ ] Raise `[train]` to the real floor, or state in a comment why a lower number is kept deliberately
- [ ] Make `soup doctor`'s check read the declared floor rather than a second hardcoded copy (`src/soup_cli/commands/doctor.py:20` currently carries its own, as `("torch", "torch", "2.3.0", True)`)
- [ ] A test asserting the declared torch floor is >= what the declared transformers floor requires, so the two cannot drift apart again silently

**Acceptance**

- [ ] `pyproject.toml`'s torch floor is consistent with what the rest of `[train]` forces
- [ ] Exactly one place in the repo declares it
- [ ] Mutating the declared floor to an inconsistent value fails a named test

Found while merging #603 (@Amix29), which is not at fault — it inherited the inconsistency and made it smaller.

umran666:
Closes #636 (claimed in [this comment](https://github.com/MakazhanAlpamys/Soup/issues/636#issuecomment-5503635641)).

> **Reworked after review** — the first cut had doctor read the floor from *installed* metadata and the drift test compare against whatever transformers happened to be installed. @MakazhanAlpamys reproduced both failure modes (stale editable dist-info printing `>=2.0.0`; the guard passing at `2.3.0` on any box below transformers 5.16.1). The design below is the review's endorsed "less clever" option: one literal in `src/`, pinned by tests, and a guard that only claims to run where it is meaningful. Review response: [comment](https://github.com/MakazhanAlpamys/Soup/pull/641#issuecomment-5506591314).

## What

**1. `[train]` declares the real floor** — `torch>=2.5.0`, matching what transformers 5.16.1's own torch extra forces (`torch>=2.5; extra == "torch"`, verified from real metadata; `.github/constraints/transformers-floor.txt` already pinned `torch==2.5.1` beside `transformers==5.16.1`, so CI validated 2.5 before this PR made the declaration match). The comment names transformers as the binding constraint; oQ/Qwen4's uint32 need (2.3) is subsumed. **Unchanged from the reviewed cut** — the review kept the number and its rationale.

**2. `soup doctor` keeps a single literal, pinned to the declaration.** `DEPS`'s torch row is `("torch", "torch", "2.5.0", True)` with a comment pointing at pyproject and the pinning tests. The metadata reader (`_declared_train_floor`) is gone: dist-info records the install's *history*, not the declaration — an editable checkout whose pyproject moved on printed a floor nobody declared, and an uninstalled source tree printed `>=?`, which `_version_ok` treats as OK (wrong in the unsafe direction). A literal cannot go stale in either direction; two independent tests keep it honest (below).

**3. Guards** (`tests/test_issue636_torch_floor.py`, regex-parsed pyproject per the `test_requires_python_bound.py` convention since `tomllib` is 3.11+):

- `test_declared_floor_meets_the_declared_transformers_requirement` — pyproject's torch floor >= the torch floor forced by the **declared** transformers floor (5.16.1), read from real metadata **only where installed == declared** — the `transformers-floor` CI job installs exactly that release under the constraints file. Anywhere else it **skips with a reason** (naming both versions) instead of passing vacuously: an older installed transformers forces a lower floor (the vacuity the review demonstrated), a newer one could turn the repo red with no repo change (the review's second-order hazard).
- `test_constraints_file_pins_the_declared_floors` — hermetic. The constraints file's `transformers==` pin must equal the declared floor (or the metadata guard validates a different release than `[train]` declares), and its `torch==` pin must satisfy the declared torch floor (or the floor job cannot resolve).
- `test_doctor_torch_floor_matches_the_pyproject_declaration` — hermetic. DEPS' literal equals the pyproject-declared floor, compared dynamically (unlike the #602 pin, which asserts both sides against literals — this one fails if *either* side moves alone). Replaces the tautological test the review flagged (it compared DEPS against the same call that built the row).
- The #602 pin (`test_qwen4_oq_torch_floor_matches_project_and_doctor`, updated in `70f3113` with the reason stated in [this comment](https://github.com/MakazhanAlpamys/Soup/pull/641#issuecomment-5504676555)) is hermetic again now that DEPS holds a literal.

Hermetic guards, run locally (each reverted after):

```
doctor literal 2.5.0 -> 2.3.0        FAILED ...::test_doctor_torch_floor_matches_the_pyproject_declaration
                                     FAILED ...::test_qwen4_oq_torch_floor_matches_project_and_doctor
pyproject floor 2.5.0 -> 2.3.0       FAILED (same two, by name)
constraints transformers pin -> 5.17 FAILED ...::test_constraints_file_pins_the_declared_floors
constraints torch pin -> 2.4.0       FAILED ...::test_constraints_file_pins_the_declared_floors
baseline                             3 passed, 1 skipped (local transformers is 5.13.0.dev0, skip names it)
```

The metadata guard's body cannot run on this box (installed != declared), so it was **simulated**: declared floor temporarily set to the installed 5.13.0.dev0, whose torch extra forces `>=2.4` —

```
floor 2.5.0, guard body runs         1 passed
floor 2.3.0 (below the forced 2.4)   1 failed  <- the #636 regression, caught by name
```

The simulation earned its keep: it caught a `Version`-vs-`str` `TypeError` in `max(floors, key=Version)` that would have reddened the `transformers-floor` job itself — the one place the guard really runs. Fixed before push.

## Verification

- `ruff check` clean; `tests/test_issue636_torch_floor.py` + the #602 pin + `test_cli_startup_is_light.py`: **37 passed, 1 skipped** (the named skip) in 27s.
- `soup doctor` import weight: the module-scope metadata call is gone; DEPS is literals only.
- The 10 `test_doctor.py` CLI-invocation failures and 7 `test_issue602` Qwen4 failures on this machine are pre-existing local-env breakage (torchao requires torch>=2.11 against installed 2.6.0; transformers 5.13.0.dev0 against a 5.16.1-floor repo) — identical with changes stashed, and all pass in CI.

## Scope note

doctor's other rows (`transformers>=5.16.1`, etc.) are the same duplication shape and remain unpinned — the issue names the torch row, so this PR stays on it. The pinning pattern (`test_doctor_torch_floor_matches_the_pyproject_declaration`) generalizes row-by-row if you want a follow-up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,7 @@ jobs:
           tests/test_issue571_qwen4_exp.py
           tests/test_plotext_compat.py
           tests/test_bugfixes.py::TestDiffModelLoading
+          tests/test_issue636_torch_floor.py
 
   test:
     strategy:
--- a/changelog.d/0.73.3/641.fixed.md
+++ b/changelog.d/0.73.3/641.fixed.md
@@ -1,0 +1,9 @@
+- [#641](https://github.com/MakazhanAlpamys/Soup/pull/641) declares the real
+  torch floor and removes its second copy. `[train]` said `torch>=2.3.0` while
+  requiring `transformers>=5.16.1`, whose own torch extra forces `torch>=2.5` —
+  the declared floor could never bind, and `soup doctor` carried its own
+  hardcoded copy on top. The extra now declares `torch>=2.5.0`, doctor reads
+  the floor from the installed `[train]` metadata (the #368 pattern, no literal
+  fallback), and a drift test pins the declaration against the installed
+  transformers' requirement so the two cannot separate silently again.
+  Closes #636.
--- a/changelog.d/0.73.3/641.fixed.md
+++ b/changelog.d/0.73.3/641.fixed.md
@@ -1,9 +1,13 @@
-- [#641](https://github.com/MakazhanAlpamys/Soup/pull/641) declares the real
-  torch floor and removes its second copy. `[train]` said `torch>=2.3.0` while
-  requiring `transformers>=5.16.1`, whose own torch extra forces `torch>=2.5` —
-  the declared floor could never bind, and `soup doctor` carried its own
-  hardcoded copy on top. The extra now declares `torch>=2.5.0`, doctor reads
-  the floor from the installed `[train]` metadata (the #368 pattern, no literal
-  fallback), and a drift test pins the declaration against the installed
-  transformers' requirement so the two cannot separate silently again.
+- **`[train]` now declares the torch floor that actually binds, and doctor's
+  copy is pinned to it (#636 by @umran666 in #641).**
+  `[train]` said `torch>=2.3.0` while requiring `transformers>=5.16.1`, whose
+  own torch extra forces `torch>=2.5` — the declared floor could never bind,
+  and `soup doctor` carried an unpinned second copy on top. The extra now
+  declares `torch>=2.5.0` with the reason stated beside it; doctor keeps a
+  literal that `tests/test_issue636_torch_floor.py` pins to the declaration,
+  and a drift guard compares the declared floor against the *declared*
+  transformers floor's real metadata on the `transformers-floor` CI job (which
+  installs exactly 5.16.1 under the constraints file), skipping with a reason
+  anywhere the installed transformers is a different release rather than
+  passing vacuously.
   Closes #636.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,8 +51,13 @@ dependencies = [
 # v0.71.0 — heavy training stack. `pip install 'soup-cli[train]'` to fine-tune.
 # These were core dependencies through v0.70.0.
 train = [
-    # oQ/Qwen4 uses safetensors U32 plus torch.uint32, both supported from 2.3.
-    "torch>=2.3.0",
+    # The binding floor is transformers', not Soup's: the torch extra of
+    # 5.16.1 (the floor declared just below) requires torch>=2.5, so no
+    # installable [train] env can carry torch below 2.5 regardless of what
+    # is declared here (#636). oQ/Qwen4's uint32 need (2.3) is subsumed.
+    # tests/test_issue636_torch_floor.py pins this floor against the
+    # installed transformers' requirement so the two cannot drift silently.
+    "torch>=2.5.0",
     # 5.16.1 is the validated Qwen4-Exp floor: an outer `qwen4_exp` config
     # selects the text-only Qwen4ExpForCausalLM, and the patch release restores
     # tensor-parallel compatibility needed by large-model training. The v5
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,8 +55,10 @@ train = [
     # 5.16.1 (the floor declared just below) requires torch>=2.5, so no
     # installable [train] env can carry torch below 2.5 regardless of what
     # is declared here (#636). oQ/Qwen4's uint32 need (2.3) is subsumed.
-    # tests/test_issue636_torch_floor.py pins this floor against the
-    # installed transformers' requirement so the two cannot drift silently.
+    # tests/test_issue636_torch_floor.py pins this floor against the declared
+    # transformers floor's own requirement (real metadata, on the
+    # transformers-floor CI job that installs exactly 5.16.1) and pins
+    # doctor's literal copy against this file, so neither can drift silently.
     "torch>=2.5.0",
     # 5.16.1 is the validated Qwen4-Exp floor: an outer `qwen4_exp` config
     # selects the text-only Qwen4ExpForCausalLM, and the patch release restores
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -16,48 +16,13 @@ from soup_cli.utils.constants import GITHUB_URL
 console = Console()
 
 
-def _declared_train_floor(package: str) -> str:
-    """The ``>=`` floor soup-cli's ``[train]`` extra declares for *package*.
-
-    Read from installed metadata — the #368 pattern ``env_lock`` uses — so
-    doctor carries no second hardcoded copy to drift (#636). Returns ``"?"``
-    when the metadata is unreadable (e.g. a source tree without an install);
-    ``_version_ok`` treats an unparseable floor as OK, so the row degrades to
-    informational instead of enforcing a stale literal.
-    """
-    try:
-        from importlib.metadata import PackageNotFoundError
-        from importlib.metadata import requires as _requires
-
-        reqs = _requires("soup-cli")
-    except (PackageNotFoundError, ImportError):
-        return "?"
-    if not reqs:
-        return "?"
-    try:
-        from packaging.requirements import InvalidRequirement, Requirement
-        from packaging.utils import canonicalize_name
-    except ImportError:  # pragma: no cover - packaging is a declared core dep
-        return "?"
-    target = canonicalize_name(package)
-    for raw in reqs:
-        try:
-            req = Requirement(raw)
-        except InvalidRequirement:
-            continue
-        if canonicalize_name(req.name) != target:
-            continue
-        if req.marker is not None and not req.marker.evaluate({"extra": "train"}):
-            continue
-        floors = [spec.version for spec in req.specifier if spec.operator == ">="]
-        if floors:
-            return floors[0]
-    return "?"
-
-
 # Dependencies to check: (import_name, package_name, min_version, required)
 DEPS = [
-    ("torch", "torch", _declared_train_floor("torch"), True),
+    # The torch floor is declared once, in pyproject.toml's [train] extra.
+    # This literal is a copy, pinned to the declaration by
+    # tests/test_issue636_torch_floor.py — reading installed metadata instead
+    # would report the install's history, not the declaration (#636).
+    ("torch", "torch", "2.5.0", True),
     ("transformers", "transformers", "5.16.1", True),
     ("peft", "peft", "0.20.0", True),
     ("trl", "trl", "0.29.0", True),
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -15,9 +15,49 @@ from soup_cli.utils.constants import GITHUB_URL
 
 console = Console()
 
+
+def _declared_train_floor(package: str) -> str:
+    """The ``>=`` floor soup-cli's ``[train]`` extra declares for *package*.
+
+    Read from installed metadata — the #368 pattern ``env_lock`` uses — so
+    doctor carries no second hardcoded copy to drift (#636). Returns ``"?"``
+    when the metadata is unreadable (e.g. a source tree without an install);
+    ``_version_ok`` treats an unparseable floor as OK, so the row degrades to
+    informational instead of enforcing a stale literal.
+    """
+    try:
+        from importlib.metadata import PackageNotFoundError
+        from importlib.metadata import requires as _requires
+
+        reqs = _requires("soup-cli")
+    except (PackageNotFoundError, ImportError):
+        return "?"
+    if not reqs:
+        return "?"
+    try:
+        from packaging.requirements import InvalidRequirement, Requirement
+        from packaging.utils import canonicalize_name
+    except ImportError:  # pragma: no cover - packaging is a declared core dep
+        return "?"
+    target = canonicalize_name(package)
+    for raw in reqs:
+        try:
+            req = Requirement(raw)
+        except InvalidRequirement:
+            continue
+        if canonicalize_name(req.name) != target:
+            continue
+        if req.marker is not None and not req.marker.evaluate({"extra": "train"}):
+            continue
+        floors = [spec.version for spec in req.specifier if spec.operator == ">="]
+        if floors:
+            return floors[0]
+    return "?"
+
+
 # Dependencies to check: (import_name, package_name, min_version, required)
 DEPS = [
-    ("torch", "torch", "2.3.0", True),
+    ("torch", "torch", _declared_train_floor("torch"), True),
     ("transformers", "transformers", "5.16.1", True),
     ("peft", "peft", "0.20.0", True),
     ("trl", "trl", "0.29.0", True),
--- a/tests/test_issue602_qwen4_ple_streaming.py
+++ b/tests/test_issue602_qwen4_ple_streaming.py
@@ -1058,9 +1058,9 @@ def test_qwen4_oq_torch_floor_matches_project_and_doctor():
         f"{torch_entries}"
     )
     assert next(item for item in DEPS if item[0] == "torch")[2] == "2.5.0", (
-        "`soup doctor` must report the same floor pyproject.toml declares, read "
-        "from the installed [train] metadata rather than a second hardcoded "
-        "copy (#636)"
+        "`soup doctor` keeps a literal copy of the torch floor; it must equal "
+        "the one pyproject.toml declares, pinned by "
+        "tests/test_issue636_torch_floor.py (#636)"
     )
 
 
--- a/tests/test_issue602_qwen4_ple_streaming.py
+++ b/tests/test_issue602_qwen4_ple_streaming.py
@@ -1051,13 +1051,16 @@ def test_qwen4_oq_torch_floor_matches_project_and_doctor():
     entries = re.findall(r'"([^"]+)"', block.group(1))
     torch_entries = [e for e in entries if e.split(">")[0].strip() == "torch"]
 
-    assert torch_entries == ["torch>=2.3.0"], (
+    assert torch_entries == ["torch>=2.5.0"], (
         "the `train` extra must declare exactly one torch floor, and it must be "
-        f"2.3.0 -- Qwen4/oQ needs uint32 support; found {torch_entries}"
+        "2.5.0 -- transformers 5.16.1's own torch extra forces torch>=2.5, "
+        "which subsumes Qwen4/oQ's uint32 need at 2.3 (#636); found "
+        f"{torch_entries}"
     )
-    assert next(item for item in DEPS if item[0] == "torch")[2] == "2.3.0", (
-        "`soup doctor` carries its own copy of the floor and it has drifted "
-        "from pyproject.toml (see #636)"
+    assert next(item for item in DEPS if item[0] == "torch")[2] == "2.5.0", (
+        "`soup doctor` must report the same floor pyproject.toml declares, read "
+        "from the installed [train] metadata rather than a second hardcoded "
+        "copy (#636)"
     )
 
 
--- a/tests/test_issue636_torch_floor.py
+++ b/tests/test_issue636_torch_floor.py
@@ -1,14 +1,23 @@
-"""The declared torch floor must be the binding one, declared in one place (#636).
+"""The declared torch floor must be the binding one, and doctor's copy pinned (#636).
 
 ``pyproject.toml [train]`` declared ``torch>=2.3.0`` while requiring
 ``transformers>=5.16.1``, whose own torch extra forces ``torch>=2.5`` — so the
-declared floor could never bind, and ``soup doctor`` carried a second hardcoded
+declared floor could never bind, and ``soup doctor`` carried an unpinned second
 copy of it. These tests pin both halves:
 
-* the pyproject floor is at least what the *installed* transformers' torch
-  extra requires (mutating pyproject back to 2.3.0 fails by name), and
-* doctor's torch row reads the declared floor from installed metadata rather
-  than a literal (re-hardcoding any literal in doctor.py fails by name).
+* the pyproject floor is at least what the *declared* transformers floor's own
+  torch extra requires. That requirement is read from real metadata only where
+  the installed transformers is exactly the declared floor — the
+  ``transformers-floor`` CI job installs precisely that release under
+  ``.github/constraints/transformers-floor.txt``. Anywhere else the check skips
+  with a reason instead of passing vacuously against whatever transformers
+  happens to be installed (an older one forces a lower floor, a newer one could
+  turn the repo red without any repo change).
+* the constraints file pins the declared transformers floor, so the metadata
+  check above can never silently validate a different release than the one
+  ``[train]`` declares.
+* doctor's torch literal equals the pyproject declaration — hermetic, repo text
+  against src literal, the pin that keeps the second copy honest.
 
 pyproject is parsed with regex rather than ``tomllib`` because ``tomllib`` is
 3.11+ and this repo supports 3.10 — the same convention as
@@ -25,7 +34,7 @@ from packaging.version import Version
 
 ROOT = pathlib.Path(__file__).resolve().parents[1]
 PYPROJECT = ROOT / "pyproject.toml"
-DOCTOR_PY = ROOT / "src" / "soup_cli" / "commands" / "doctor.py"
+CONSTRAINTS = ROOT / ".github" / "constraints" / "transformers-floor.txt"
 
 
 def _train_extra_block() -> str:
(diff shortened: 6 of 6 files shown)


umran666:
The nine red cells were **one test**, and fixing it required editing an existing test — stating that plainly rather than letting the diff speak.

## Root cause

`tests/test_issue602_qwen4_ple_streaming.py::test_qwen4_oq_torch_floor_matches_project_and_doctor` (merged with #603) pins the exact state #636 exists to remove:

```python
assert torch_entries == ["torch>=2.3.0"], (...)
assert next(item for item in DEPS if item[0] == "torch")[2] == "2.3.0", (
    "`soup doctor` carries its own copy of the floor and it has drifted "
    "from pyproject.toml (see #636)")
```

Its own failure message already referenced #636 — the pin was written knowing this issue was coming. Verified it was the **sole** failure in all nine cells: identical one-line FAILED summary in the Ubuntu 3.10, Windows 3.11 and macOS 3.10 job logs.

## The edit (`70f3113`)

Both assertions moved to the new contract, same bounded-regex shape, no other change to the file:

- `[train]` declares exactly one torch floor and it is `2.5.0` — transformers 5.16.1's torch extra forces it, subsuming Qwen4/oQ's uint32 need at 2.3;
- doctor's DEPS row reports the same `2.5.0`, read from installed metadata rather than a second hardcoded copy.

## Why this is not "a test edited to accommodate a refactor"

The #616 rule — stop and say so if an existing test needs editing — is exactly why this comment exists. Here the edited test asserted the *drifted* state as the contract; #636's acceptance criteria ("pyproject's torch floor is consistent with what the rest of `[train]` forces", "exactly one place declares it") cannot be met while that pin stands. The edit is the issue's mandate, not a convenience, and the new assertions are strictly stronger: they pin the floor to the value transformers actually forces and pin doctor to the metadata-derived number.

## Local evidence

- The floor test now passes; `tests/test_issue636_torch_floor.py` 3 passed; ruff clean.
- The other 7 failures in `test_issue602_qwen4_ple_streaming.py` on this machine are pre-existing local-env breakage (transformers 5.13.0.dev0 against a 5.16.1-floor repo) — identical failure list with my changes stashed, and all 7 **passed in CI** on the previous push, where the floor test was the only FAILED line.

MakazhanAlpamys (changes requested):
Reviewed at `70f3113`. **Requesting changes.** The floor number itself is right and well argued; the de-duplication answers a different question than #636 asked, and I reproduced that myself rather than taking it from a report.

## The blocker: doctor now reads the *installed* floor, not the *declared* one

Run in your worktree, unmodified:

```
$ grep -n "torch>=" pyproject.toml
60:    "torch>=2.5.0",

$ soup doctor
| torch        | yes | 2.5.1+cu121 | >=2.0.0  | OK       |
| transformers | yes | 4.57.6      | >=5.16.1 | outdated |
```

`pyproject.toml` says `2.5.0`; doctor prints `>=2.0.0`. `_declared_train_floor` reads `soup-cli`'s **dist-info**, which records whatever was installed — on this box, a floor that has not been in `pyproject.toml` since #603 raised it to 2.3.0. Every other row still prints its correct declared floor because those stayed hardcoded, so torch is now the one row that is wrong, and it is wrong **in the unsafe direction**: it says `OK`.

#636 asked doctor to stop reporting a number that does not bind. As shipped it reports a number that binds even less — the old one was at least a copy of something a human had chosen.

The boundary is worse than the common case. With `soup-cli` not installed at all — source tree plus `PYTHONPATH`, an ordinary contributor setup — `_declared_train_floor` returns `"?"` (`doctor.py:19-51`) and `_version_ok("1.0.0", "?")` returns `True` (`doctor.py:589-590`). The torch row stops being a check and prints `>=?`.

What I would take instead: read `pyproject.toml` when it is present beside the package (the repo case) and fall back to metadata only for an installed wheel — or keep a single literal in `src/` and have a test assert it equals what `pyproject.toml` declares. The second is less clever and cannot go stale in either direction.

## The drift test is vacuous outside CI

`tests/test_issue636_torch_floor.py:48-68` reads `importlib.metadata.requires("transformers")` — the **installed** transformers, not the declared `5.16.1`. Installed here is 4.57.6, whose torch extra is `torch>=2.2`, so:

| pyproject floor | test result |
|---|---|
| `2.5.0` (shipped) | pass |
| **`2.3.0`** — the exact regression #636 describes | **pass** |
| `2.2.0` | **pass** |
| `2.0.0` | fail |

The comment at `:73-74` claims "setting pyproject back to `torch>=2.3.0` fails this test by name". That is false on any box whose transformers is below 5.16.1, which is most boxes right now. On CI it would be a real check, but nothing asserts that the installed transformers satisfies the *declared* floor, so the guard degrades silently exactly where a contributor would rely on it. There is also a second-order hazard: an upstream transformers release raising its own torch requirement turns this test red with no change in this repo.

`test_doctor_torch_row_reads_the_declared_floor` (`:83-98`) compares `DEPS[torch][2]` against `_declared_train_floor("torch")` — **the same call that built the row**. It is a tautology; it killed the re-hardcoding mutation here only because stale metadata (2.0.0) happened to differ from the injected literal. On a freshly installed checkout both sides read 2.5.0 and that test passes with the floor hardcoded again. Only the source-regex test at `:100-107` is a genuine guard.

## What is right, and it is the substantive half

The number is correct and **verified against real metadata rather than believed**: transformers 5.16.1 declares `torch>=2.5; extra == "torch"`, and `.github/constraints/transformers-floor.txt:7` already pins `torch==2.5.1`, so CI was consistent with 2.5 before this PR made the declaration match. The six-line rationale at `pyproject.toml:54-59` is exactly what I asked for in item 1 — a stated reason rather than a bare bump.

You correctly updated `test_qwen4_oq_torch_floor_matches_project_and_doctor`, which I had rewritten hours earlier, rather than deleting or weakening it — 5 deleted lines, all inside that one assertion block. But it is **red on this box** for the same root cause:

```
main  e83b992 -> 1 passed
#641  70f3113 -> 1 failed   assert '2.0.0' == '2.5.0'
```

A test that was hermetic (repo text against a repo literal) now depends on when the reviewer last ran `pip install -e .`. That is the clearest symptom of the design issue above.

## Two smaller things

`changelog.d/0.73.3/641.fixed.md` explains why nobody is excluded, which is the substantive impact answer — but it ends with only `Closes #636.` and **credits nobody**. The neighbouring fragments use `(#621 by @SID-6921 in #637)` and `(#633 by @Srinivasan8888 in #635)`; yours should read `(#636 by @umran666 in #641)`. That convention exists because `changelog.d/` is assembled verbatim into the release notes, and it is the only place most readers ever see a contributor's name.

`ruff` clean, `test_cli_startup_is_light` 34 passed, `soup --help` 1.49 s — the module-scope `_declared_train_floor` call touches only `importlib.metadata` and `packaging`, so no import-weight regression.

**To unblock:** make doctor's floor come from the repo's own declaration rather than install state (or keep one literal and pin it with a test against `pyproject.toml`), make the drift test compare against the *declared* transformers floor, and add the credit line. The 2.5.0 number and its rationale I would keep exactly as they are.

MakazhanAlpamys:
Following up on my review at `70f3113` with three findings I did not have then. I ran each of these myself in a detached worktree at your head, with `PYTHONPATH` pinned to that worktree, so the numbers below are from your code and not from mine.

**First, credit where it is due:** the 2.5.0 raise and its six-line rationale in `pyproject.toml` are correct and should ship verbatim, and you updated the `#602` pin rather than deleting it — that is the right instinct, and it is what I check for first. The problem is entirely in the de-duplication mechanism and in what the new tests can actually see.

## 1. The mutation your test file documents does not fail

`tests/test_issue636_torch_floor.py:73-74` states in a comment:

> Mutation check: setting pyproject back to `torch>=2.3.0` fails this test by name

I ran exactly that — `torch>=2.5.0` → `torch>=2.3.0`, the precise regression #636 exists to remove:

```
tests/test_issue636_torch_floor.py::...::test_declared_floor_meets_transformers_torch_requirement PASSED
tests/test_issue636_torch_floor.py::...::test_doctor_torch_row_reads_the_declared_floor            PASSED
tests/test_issue636_torch_floor.py::...::test_doctor_source_declares_no_torch_floor_literal        PASSED
3 passed in 1.09s
```

The reason is in the helper: `_transformers_torch_floor()` reads the **installed** transformers, which is 4.57.6 on this box, whose torch floor is `2.2`. `2.3.0 >= 2.2` holds, so the test is satisfied by the very value it was written to reject. Your PR body reports this mutation as failing; that is reproducible only on an environment that happens to have `transformers >= 5.16.1` installed, and it is stated unconditionally.

Two consequences, and the second is worse than the first: the test is **vacuous on most machines today**, and it will turn red with no change in this repo the day upstream transformers raises its own torch requirement. A test that fails for reasons outside the repository is one people learn to re-run rather than read.

The fix is to compare against the **declared** transformers floor, parsed out of `pyproject.toml` the same way you already parse the torch one — then both sides of the comparison come from the file the test is about.

## 2. `test_doctor_torch_row_reads_the_declared_floor` is a tautology

```python
floor = _declared_train_floor("torch")
assert torch_rows[0][2] == floor
```

and `doctor.py:60` is

```python
("torch", "torch", _declared_train_floor("torch"), True),
```

The assertion compares the result of a call against the same call. It cannot fail while that line stands, whatever either side returns — and it self-`skip`s when the floor is `"?"`, so on the degraded path it is not merely uncovered, it removes itself. A skipped test is the same colour as a passing one in a CI summary.

## 3. A genuine second copy survives, which is the acceptance criterion of #636

#636 asks that "exactly one place in the repo declares it, and mutating the declared floor to an inconsistent value fails a named test." I reintroduced a second declaration in the shape a future contributor would most plausibly use — a module constant rather than an inline literal:

```python
_TORCH_FLOOR = "2.0.0"
...
("torch", "torch", _TORCH_FLOOR, True),
```

Result: `3 passed`. The source guard at `:100-107` only matches the literal 3-tuple shape it was written against, so it does not see this, and the tautology in §2 does not see it either. The criterion is not met.

## 4. Changelog credit

`changelog.d/0.73.3/641.fixed.md` ends with a bare `Closes #636.` Its neighbours in that directory carry `(#459 by @blackcoderx in #468)`. Yours should read `(#636 by @umran666 in #641)` — I asked for this previously and it has not been actioned.

## What I would like instead

My original finding stands unchanged: doctor prints `>=2.0.0` on this box while `pyproject.toml` declares `2.5.0`, and one detail I had missed makes it sharper — the remediation line emits `pip install -U 'torch>=2.0.0'`, so doctor actively instructs the user to install *below* the real floor, and the row reads `OK` while doing it. That is the unsafe direction.

Note that the metadata indirection buys nothing even when it works: for a correctly installed wheel, dist-info and `pyproject.toml` are the same number by construction. All it adds is the drift you have hit plus the `"?"` degraded path. **The less clever option is the better one here** — keep one literal in `src/`, and pin it with a test that reads `pyproject.toml` directly. That cannot go stale in either direction, and it is the shape `tests/test_requires_python_bound.py` already uses in this repo for exactly this kind of bound.

Take the time you need; the version number itself is right and I am not in a hurry to lose it.

umran666:
## 1. Doctor reads the installed floor, not the declared one — fixed, via your second option

`_declared_train_floor` is **deleted**; DEPS holds `("torch", "torch", "2.5.0", True)` with a comment naming pyproject as the declaration and the tests as the pin. Your reproduction was the argument that settled it: dist-info records the install's history, and the `"?"` → `_version_ok` → OK path made the uninstalled case wrong in the unsafe direction. A literal cannot go stale in either direction, and the `>=?` row no longer exists.

Your `e83b992`-vs-`70f3113` comparison of the #602 test now resolves the other way: with DEPS a literal again, that test is hermetic (repo text against src literal) exactly as it was on main — `2.5.0` on both sides regardless of when anyone last ran `pip install -e .`.

## 2. The drift test compared against the installed transformers — fixed

`test_declared_floor_meets_the_declared_transformers_requirement` now:

- parses the **declared** transformers floor from pyproject (`5.16.1`);
- runs the metadata comparison **only where installed == declared** — the `transformers-floor` CI job creates exactly that state via `.github/constraints/transformers-floor.txt`;
- **skips with a reason naming both versions** anywhere else. No vacuous pass: on your box (4.57.6) it reports the mismatch instead of silently blessing `2.3.0`. Your table becomes: `2.5.0` pass, `2.3.0` **fail by name**, on the job whose entire purpose is the declared floor.
- The second-order hazard is gone with the equality gate — an upstream release raising its torch floor makes the matrix jobs skip, not redden; the floor job stays pinned to 5.16.1 until the repo decides to move.

Since the guard only means something where constraints create the right install, a new hermetic test ties the two together: `test_constraints_file_pins_the_declared_floors` asserts the constraints `transformers==` pin **equals** the declared floor (else the guard validates a different release than `[train]` declares) and the `torch==` pin **satisfies** the declared torch floor (else the floor job can't resolve).

The tautology is replaced, not patched: `test_doctor_torch_floor_matches_the_pyproject_declaration` compares the DEPS literal against the pyproject regex — two independent sources, fails if either moves alone. `test_doctor_source_declares_no_torch_floor_literal` is deleted; it forbade the design you endorsed.

**Evidence, including one the simulation caught.** The guard body can't execute on this box (installed 5.13.0.dev0 ≠ declared), so I simulated it: temporarily set the declared floor to the installed 5.13.0.dev0, whose torch extra forces `>=2.4`. Floor `2.5.0` → pass; floor `2.3.0` → **fail by name** — the exact #636 regression, caught by the real code path. The simulation also caught a `Version`-vs-`str` `TypeError` in `max(floors, key=Version)` that would have reddened the `transformers-floor` job itself — the one place the guard runs for real. Fixed before push; without the simulation that bug ships.

Hermetic mutations, each reverted: doctor literal → `2.3.0` fails two tests by name; pyproject floor → `2.3.0` fails the same two; constraints transformers pin → `5.17.0` fails by name; constraints torch pin → `2.4.0` fails by name. Baseline: 37 passed, 1 skipped (the named skip) across the 636 file, the #602 pin, and `test_cli_startup_is_light.py`; ruff clean.

## 3. Credit line — added

`changelog.d/0.73.3/641.fixed.md` now opens bold-headline style per the neighbouring fragments: `(**... (#636 by @umran666 in #641).**` and the body describes the shipped design (pinned literal, declared-floor guard), not the rejected one. The pyproject comment's last lines were also stale — they described the guard as checking "the installed transformers' requirement"; now they name the declared floor and the CI job.

Kept as-is per your review: the `2.5.0` number, the six-line rationale, and the #602 test update.

MakazhanAlpamys (changes requested):
**All four findings from my last two rounds are closed, and I verified each by running it rather than reading the diff.** Blocking on two one-line items, one of which ships text into the public CHANGELOG.

## What is closed

- **A — doctor reads the declared floor.** Real `soup doctor` from a worktree at your head: `│ torch │ yes │ 2.5.1+cu121 │ >=2.5.0 │ OK │`, and the remediation line now says `pip install -U 'torch>=2.5.0'`. The `>=2.0.0` row and the advice to install *below* the real floor are both gone. `_declared_train_floor` is deleted.
- **B — the documented mutation now actually fails.** `pyproject` back to `2.3.0` is killed by two named tests. The guard also skips with a *stated reason* rather than passing vacuously when the installed transformers is not the declared floor — which is the honest shape, and better than what I asked for.
- **C — the tautology is gone.** The pin now compares the imported `DEPS` literal against a regex over `pyproject.toml`: two independent sources, and it dies under five separate mutations.
- **D — the second-copy hole is closed.** My exact prior mutation (`_TORCH_FLOOR = "2.0.0"` fed into `DEPS`), which previously gave `3 passed`, is now killed by two named tests. The guard became value-based instead of source-regex-based, so it sees through indirection — that is the right move and it is why the acceptance criterion of #636 is now genuinely met.
- **Credit line in the fragment** — actioned, and in the neighbours' style. Thank you.

You took the less clever option and the result is better than the design I blocked. Also worth noting: deleting `test_doctor_torch_floor_matches...`'s key assertion and then drifting `DEPS` is still caught, by the #602 pin. Defence in depth that neither test knows about, which is the good kind.

## Blocker 1 — the guard is not run by the job the PR says runs it

The test docstring, the PR body and **the shipped changelog fragment** all say the drift guard runs "on the `transformers-floor` CI job (which installs exactly 5.16.1 under the constraints file)". That job's pytest step is:

```yaml
pytest -o addopts= --no-cov -q
  tests/test_transformers_floor_compat.py
  tests/test_issue571_qwen4_exp.py
  tests/test_plotext_compat.py
  tests/test_bugfixes.py::TestDiffModelLoading
```

`tests/test_issue636_torch_floor.py` is not in it. I checked directly: `git diff origin/main...d73626ac -- .github/` is **empty**, and `grep -rn 'test_issue636' .github/` has **no match**. The floor job's own log reads `36 passed` — the 36 that do not include your guard.

It does run today, but by coincidence: transformers 5.16.1 is currently the newest release, so the ordinary matrix resolves `>=5.16.1,<6.0.0` to exactly the declared floor and the guard's precondition happens to hold. **The day 5.16.2 ships, all nine matrix cells skip, the floor job still does not collect the file, and the guard goes dormant with no repo change and no signal.** That is the same shape as the finding I blocked on last round — a mechanism asserted in prose that the code does not implement — and it is #636's own third bullet, an acceptance criterion nothing enforces.

One line in `.github/workflows/ci.yml` adds the file to that job. Then the prose is true. If you would rather the matrix be its home, that is defensible, but the sentence has to say so accurately, because as written it is contingent on upstream release timing and reads as if it is not.

The fragment matters more than the docstring here: it ships verbatim into `CHANGELOG.md` at the next release, and a claim about CI enforcement is exactly the sort a reader has no way to check.

## Blocker 2 — the #602 message describes the design you abandoned

`tests/test_issue602_qwen4_ple_streaming.py:1061-1064`:

```
"`soup doctor` must report the same floor pyproject.toml declares, read
 from the installed [train] metadata rather than a second hardcoded copy (#636)"
```

That was written for the metadata-reader design. The shipped design is the opposite — a hardcoded literal pinned by a test — and `test_issue636_torch_floor.py` documents *why* metadata was rejected. So anyone who trips this assertion is instructed by its own message to go implement the approach we removed. One string.

## Everything else

`82 passed, 8 skipped` across `test_issue636_torch_floor.py`, `test_issue602_qwen4_ple_streaming.py`, `test_doctor.py`, `test_cli_startup_is_light.py`. `ruff` clean. 13/13 CI green on `d73626a`. **Zero deleted or weakened tests** — the `#602` edit is +5/-5, both hunks `2.3.0 → 2.5.0` value updates, assertion count unchanged; that is the contract update #636 mandates. `test_constraints_file_pins_the_declared_floors` is a good addition I did not ask for.

One non-blocking note: a second declaration placed in a *different* module (say `utils/gpu.py`) still survives, because the guard imports `DEPS` and a constant nothing reads is dead code. That is no worse than before — the deleted source-regex test only ever scanned `doctor.py` — so it is not a regression, and I would not widen this PR for it.

Two one-line changes and I will merge this.

umran666:
**Blocker 1**  `tests/test_issue636_torch_floor.py` added to the transformers-floor job's pytest file list (`.github/workflows/ci.yml`). You were right that it ran only by coincidence of upstream release timing: the folded run command was parsed back with yaml.safe_load to confirm the file is in the collected command, and the commit message records the dormancy scenario (5.16.2 ships -> matrix skips everywhere, floor job never collected it, no signal) so the line does not get pruned later as redundant. The shipped fragment's claim is now enforced rather than asserted.

**Blocker 2**  `the #602 DEPS assertion message now names the shipped contract: "keeps a literal copy of the torch floor; it must equal the one pyproject.toml declares, pinned by tests/test_issue636_torch_floor.py". No one is instructed to reimplement the removed reader.

Verification: YAML parses and folds correctly; 636 file + #602 pin 3 passed, 1 named skip locally; ruff clean. The floor job should now report 39 passed (its 36 + the guard running for real where installed == declared == 5.16.1, plus the two hermetic tests) ,will confirm from the job log.

On the non-blocking note (a second declaration in another module surviving because nothing reads it): agreed it is out of scope here; the value-based pin covers every copy that doctor actually reports, which is the surface #636 names.

MakazhanAlpamys:
**Both blockers verified closed at `583acbb`. Merging as soon as CI finishes** — the nine matrix cells are still queued, and the Windows ones take about half an hour, so this is not waiting on you.

What I checked rather than took from the diff:

**Blocker 1 — and the evidence is better than "the line is there".** The floor job's own log now reads `39 passed`; before your commit it read `36`. Those three are your guard file being collected, which is the difference between the claim being enforced and being asserted. I also parsed the folded `run:` back with `yaml.safe_load` and confirmed `tests/test_issue636_torch_floor.py` is in the command the shell actually receives — a folded YAML scalar is exactly the place a path can look present and not be.

**Blocker 2** — the `#602` message now describes the shipped design ("doctor keeps a literal copy... pinned by `tests/test_issue636_torch_floor.py`") instead of the metadata reader we removed.

**Re-ran the mutations, including the one that survived last round:**

| mutation | result |
|---|---|
| `pyproject` floor `2.5.0` → `2.3.0` | killed — `test_doctor_torch_floor_matches_the_pyproject_declaration`, `test_qwen4_oq_torch_floor_matches_project_and_doctor` |
| `DEPS` row `2.5.0` → `2.0.0` | killed — same two |
| second copy via module constant `_TF = "2.0.0"` fed into `DEPS` | **killed** — same two (this is the one that gave `3 passed` two rounds ago) |

`82 passed, 8 skipped` across the four affected files. Real `soup doctor` from your head: `│ torch │ yes │ 2.5.1+cu121 │ >=2.5.0 │ OK │` with `pip install -U 'torch>=2.5.0'`.

**On the five removed test lines**, since a reviewer should never wave those through: they are the two `#602` assertions' values and messages, `2.3.0 → 2.5.0`. Both assertions still exist and the count is unchanged. That is the contract update #636 mandates — the old pin asserted the drifted state and said so itself.

One note for the record rather than for you to act on. Your first version was the clever one — read the floor from installed metadata so there is only ever one copy — and it was the right instinct that happened to be wrong here, because dist-info records what was *installed*, not what the repo *declares*, and those diverge exactly when it matters. The version you shipped is duller and strictly better: a literal that cannot drift because a test compares it against `pyproject.toml`. Taking that direction from review without arguing it back is worth more to this repo than the patch itself.

umran666:
CI confirmation, as promised in the last comment: the `transformers-floor` job on `583acbb` collected the guard and ran it for real —

```
Run pytest -o addopts= --no-cov -q tests/test_transformers_floor_compat.py tests/test_issue571_qwen4_exp.py tests/test_plotext_compat.py tests/test_bugfixes.py::TestDiffModelLoading tests/test_issue636_torch_floor.py
39 passed, 2 warnings in 9.68s
```

36 -> 39, zero skips: the metadata comparison executed where installed == declared == 5.16.1, so the fragment's CI-enforcement claim is now demonstrated in the job log rather than asserted in prose. All 13 checks green on `583acbb`. Ready for re-review.

MakazhanAlpamys:
Merged as `59b8a51`, credited in CONTRIBUTORS at `2285b65`.

Verified on merged `main`: `│ torch │ yes │ 2.5.1+cu121 │ >=2.5.0 │ OK │` with `pip install -U 'torch>=2.5.0'`.

The thing I want on the record is not the patch, it is **what you did when review disagreed with your design.** Your first version read the floor from installed metadata so there could only ever be one copy — that is the clever answer, and it is the one most people defend. It is wrong here for a reason that is worth carrying to other problems: dist-info records what was *installed*, not what the repo *declares*, and those two diverge precisely in the case the check exists to catch. You dropped it for a literal pinned by a test, which is duller and strictly better. Taking that direction without arguing it back is worth more to this repository than the fix.

Two things you closed that I would rate above the headline change:

- **The drift test compared declared-against-installed transformers**, so reverting pyproject to `2.3.0` passed on any box below 5.16.1 — while the file's own comment claimed that mutation failed by name. A comment asserting a property the code does not have is worse than no comment, because it stops the next person checking.
- **The guard's stated home never collected it.** It ran only in the ordinary matrix, and only by coincidence of 5.16.1 being newest on PyPI; 5.16.2 shipping would have made it dormant with no repo change and no signal. Your fix is verifiable in a log rather than argued — the job goes from `36 passed` to `39`. That is the shape of evidence I wish more CI changes came with.

Mutations I ran before merging, all killed by two named tests each: pyproject floor reverted to `2.3.0`; the `DEPS` literal changed to `2.0.0`; and a second declaration reintroduced through a module constant `_TF = "2.0.0"` fed into `DEPS` — that last one **survived two earlier rounds**, and it is dead now because you made the guard value-based instead of source-regex-based, so it sees through indirection. `82 passed, 8 skipped` across the four affected files; the five removed test lines are the two `#602` assertions' values and messages, both assertions intact.

Closes #636.

Issue #642, opened 2026-09-02
soup sweep --dry-run still validates nothing: a typo'd config key is silent there while soup train --dry-run now reports it


MakazhanAlpamys:
`soup train --dry-run` reports unknown config keys as of #628. `soup sweep --dry-run` does not — it exits 0 with no warning at all, because it returns before the config is ever loaded.

This is worth filing separately rather than folding into #628 because it is the **headline complaint of #627**: the reporter's original symptom was that `--dry-run` printed *Config valid* over a typo'd key. Half of that is now fixed and half is not, and the half that is not is the one the issue title describes.

## Reproduction

```bash
$ cat soup.yaml
base: hf-internal-testing/tiny-random-gpt2
task: sft
data:
  train: data.jsonl
training:
  quantizaton: none      # typo: should be `quantization`

$ soup sweep --config soup.yaml --param "training.lr=1e-5,2e-5" --dry-run
# ... prints the grid ...
$ echo $?
0
```

No warning about `quantizaton`. The equivalent `soup train --dry-run` prints the four-line warning block with the suggestion and the v0.75 deadline.

## Cause

`commands/sweep.py` returns on the `--dry-run` branch (around line 105) before the config is loaded, so neither the loader's unknown-key report nor #628's sweep pre-flight guard is ever reached. Both live downstream of that early return.

## Why it was deferred rather than fixed in #628

@Srinivasan8888 found this while working on #628 and **reported it instead of widening the PR**, which was the right call: making `--dry-run` load the config is a behaviour change to a command whose whole contract is "do not do anything", and it deserves its own decision rather than riding along in a merge.

## The decision this issue needs

`--dry-run` currently means "print the grid without loading anything". The fix means "validate, then print the grid without running anything". I think the second is what users expect from a flag whose purpose is to catch mistakes before a long run — a dry run that cannot detect a typo is not doing the job its name promises — but it is a real change and someone should say so out loud before it ships.

## Acceptance criteria

- [ ] `soup sweep --dry-run` over a config with an unknown key emits the same warning block `soup train --dry-run` does.
- [ ] The `--param` pre-flight from #628 also runs under `--dry-run`, so a typo'd sweep parameter is caught there too (`soup sweep --param "lora_rank=8,16" --dry-run` should not print a grid it can never run).
[...]

PR #643, opened 2026-09-02
fix(sweep): guard the empty grid, and test the #627 loader wiring by loading

Srinivasan8888:
Follow-up to #628, which merged at `eac1a88` two minutes before your review of it landed. Both items in that review are here; neither is on main.

**Attribution: both findings are yours, from the #628 review.** I am carrying them, not discovering them. The `--dry-run` gap you filed as #642 is *not* here — you asked for it to be argued in its own thread, and I agree.

## 1. An `IndexError` I introduced is live on main

The probe I added in #628 reads `combinations[0]` unguarded, so a `--max-runs` that empties the grid crashes inside the precheck. As shipped:

```
$ soup sweep -c soup.yaml --param "training.lr=1e-5" --max-runs -1 --yes
Error: IndexError: list index out of range
EXIT=1
```

main previously printed an empty table and exited 0. `if combinations:` restores that:

```
$ soup sweep -c soup.yaml --param "training.lr=1e-5" --max-runs -1 --yes
EXIT=0        # empty table, no traceback
```

I took the minimal fix rather than bounding `--max-runs` at `ge=1`. Bounding it is the better fix, as you said, but it rejects input the CLI accepts today, and that contract change deserves its own argument. The test records the reasoning so it does not read as an oversight.

## 2. The loader wiring had no behavioural test

Your reversion — both `_report_unknown_keys` call sites removed, the helper, the import and `unknown_keys.py` untouched — left the suite green while a config carrying `quantizaton: none` loaded silently. Re-run against main with these tests added:

```
4 failed, 47 passed
FAILED TestTheLoaderIsActuallyWiredUp::test_load_config_from_string_reports_an_unknown_key
FAILED TestTheLoaderIsActuallyWiredUp::test_load_config_reports_an_unknown_key
FAILED TestTheLoaderIsActuallyWiredUp::test_the_string_call_site_refuses_under_error_severity
FAILED TestTheLoaderIsActuallyWiredUp::test_the_file_call_site_refuses_under_error_severity
```

Seven tests, covering all three of your asks:

| test | pins |
|---|---|
| `test_load_config_from_string_reports_an_unknown_key` | API/UI site warns, names the key, carries the deadline |
| `test_load_config_reports_an_unknown_key` | CLI/file site warns — a separate branch |
| `test_load_config_from_string_is_silent_on_a_clean_config` | **control** |
| `test_load_config_is_silent_on_a_clean_config` | **control** |
| `test_the_string_call_site_refuses_under_error_severity` | the v0.75 flip, `ValueError` |
| `test_the_file_call_site_refuses_under_error_severity` | the v0.75 flip, `SystemExit` |
| `test_a_clean_config_still_loads_under_error_severity` | **control for the flip** |

**The construction-site scanner stays.** You were right that it is not a test of these call sites and right that it is still worth having — it catches a *fourth* site appearing, which these seven cannot. The class docstring says so, so the next reader does not mistake one for the other.

## Mutations

`__pycache__` cleared between runs; a same-length edit otherwise restores from stale bytecode.

| # | Mutation | Result |
|---|---|---|
| M17 | only the `load_config` (CLI/file) site reverted | **killed** — the two file tests, and *only* those |
| M18 | only the `load_config_from_string` (API/UI) site reverted | **killed** — the two string tests, and *only* those |
| M19 | detector reports every config as carrying an unknown key | **killed** — both silence controls, the flip control, and the sweep control |
| M20 | `if combinations:` guard removed | **killed** — `test_an_empty_grid_does_not_crash_the_probe` |

M17/M18 are split deliberately: a future edit to one call site that forgets the other cannot pass. M19 is the answer to "without a control, a warn-on-everything wiring passes" — it fails four tests across both features.

## Verification

```
pytest tests/test_issue627_unknown_config_keys.py tests/test_sweep.py tests/test_config.py \
       tests/test_cli_startup_is_light.py tests/test_recipes.py -q -o addopts=
176 passed

ruff check src/soup_cli/ scripts/ tests/          ->  All checks passed!
git diff origin/main...HEAD -- tests/ | grep -c "^-[^-]"  ->  0
```

Live, through the CLI rather than the helper:

```
$ soup train --config typo.yaml --dry-run
Warning: unknown config key 'training.quantizaton' - did you mean 'quantization'
```

(The exit 1 after that is the absent dataset, downstream of this.)

## Scope limits

- **#642 is not here.** Your call, and I think the right one.
- **`"Not applied."` under `error` severity** and the **difflib short/long ranking** (`lr_rate` → `uld_strategy`) are still open and still not touched. They want follow-up issues; say the word and I will file them with the numbers.
- **`_nested_models` recursing into every `BaseModel` in a `Union`** — still latent, still no multi-model unions in the tree. Not touched; you called it a comment rather than a change.
- Changelog fragment follows in the next push, once this PR has a number.

https://claude.ai/code/session_01YVpZy3Xj4xF2baapkdmNF3

### DIFF
--- a/changelog.d/0.73.3/643.fixed.md
+++ b/changelog.d/0.73.3/643.fixed.md
@@ -1,0 +1,9 @@
+- **`soup sweep` no longer crashes with an `IndexError` when the run grid is empty (#628 by @Srinivasan8888 in #643).**
+  The unknown-parameter precheck added in #628 reads `combinations[0]` to probe the grid, so
+  a `--max-runs` value that leaves no combinations — `--max-runs -1`, for instance — raised
+  `IndexError: list index out of range` from inside the guard rather than reaching the sweep.
+  Nonsense input either way, but the previous behaviour was an empty results table and exit 0,
+  and an unhandled traceback is a worse answer than that. The probe is skipped when the grid is
+  empty, restoring the old response. Bounding `--max-runs` at `ge=1` would reject the input
+  outright and is the better long-term fix; it is a pre-existing CLI contract change and is
+  left for its own change.
--- a/src/soup_cli/commands/sweep.py
+++ b/src/soup_cli/commands/sweep.py
@@ -121,14 +121,15 @@ def sweep(
     # still exit 0 -- an entirely invalid sweep that nothing downstream can
     # detect. Every combination carries the same parameter names, so one probe
     # built the way `_run_single` builds its config settles it for the grid.
-    probe = base_cfg.model_dump()
-    for key, val in combinations[0].items():
-        _set_nested_param(probe, key, val)
-    try:
-        _reject_unknown_sweep_params(probe)
-    except ValueError as exc:
-        console.print(f"[red]{exc}[/]")
-        raise typer.Exit(1) from exc
+    if combinations:
+        probe = base_cfg.model_dump()
+        for key, val in combinations[0].items():
+            _set_nested_param(probe, key, val)
+        try:
+            _reject_unknown_sweep_params(probe)
+        except ValueError as exc:
+            console.print(f"[red]{exc}[/]")
+            raise typer.Exit(1) from exc
 
     results = []
     best_loss = float("inf")
--- a/tests/test_issue627_unknown_config_keys.py
+++ b/tests/test_issue627_unknown_config_keys.py
@@ -648,6 +648,31 @@ class TestATypodSweepFailsTheCommandAndNotJustTheArm:
             "a per-arm failure table means the loop swallowed the guard"
         )
 
+    def test_an_empty_grid_does_not_crash_the_probe(self, tmp_path) -> None:
+        """The probe reads ``combinations[0]``, which an empty grid does not have.
+
+        ``--max-runs -1`` empties the grid. Before the guard this raised an
+        unhandled ``IndexError`` from inside the new precheck -- a regression
+        this PR introduced, found in the #628 review. Nonsense input either
+        way, but main printed an empty table and exited 0, and an unhandled
+        traceback is a worse answer than that. Bounding ``--max-runs`` at
+        ``ge=1`` would be the better fix and is pre-existing behaviour, so it
+        is left alone here.
+        """
+        from typer.testing import CliRunner
+
+        from soup_cli.cli import app
+
+        result = CliRunner().invoke(app, [
+            "sweep",
+            "--config", self._config(tmp_path),
+            "--param", "training.lr=1e-5",
+            "--max-runs", "-1",
+            "--yes",
+        ])
+        assert "IndexError" not in result.output, result.output
+        assert result.exit_code == 0, result.output
+
     def test_a_valid_sweep_reaches_the_first_arm(self, tmp_path) -> None:
         """Control: a precheck that refused every sweep would pass the above.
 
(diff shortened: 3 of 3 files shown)


MakazhanAlpamys:
Merged as `de64492`. Thank you — and the part I want to name first is that **you reported a regression in your own patch before I found it**, on a PR that was already closed and could not tell you it had been merged. That is a strictly harder thing to do than fixing a bug someone else points at.

What I verified here rather than taking from the diff:

- **The regression is gone from `main`.** Ran `soup sweep --param training.lr=1e-5 --max-runs -1 --yes` against the merged `main`: empty results table, **exit 0**. Before the merge it was `IndexError: list index out of range`, exit 1.
- **Each loader call site is caught independently.** This is the one I most wanted to see, so I mutated them separately rather than together: neutering only `load_config` fails exactly `test_load_config_reports_an_unknown_key` and `test_the_file_call_site_refuses_under_error_severity` (2 failed / 43 passed); neutering only `load_config_from_string` fails exactly its own two, same split. A test that covers one branch leaves the other precisely as exposed as it was, so the separation is the property that makes these worth having.
- **Zero deleted test lines** (`git diff origin/main...4a843cc -- tests/` → 0 removed), and the #628 file-text scanner kept with a docstring saying it is not a substitute — the correction rather than a deletion, which is what I asked for.
- 13/13 CI green on the exact head `4a843cc`.

Two things carried forward deliberately, both already in your docstring: an empty grid still skips the unknown-param guard, and `test_an_empty_grid_does_not_crash_the_probe` now cements exit 0 for nonsense input. Bounding `--max-runs` at `ge=1` closes both and would require editing that test — a decision, not an oversight, and it reads that way because you wrote it down.

Credited in CONTRIBUTORS as your 13th merge (`9eceea4`). #642 stays open and untouched.

On process: the premature merge was mine. Every automated pass I dispatch now carries an explicit prohibition on merging, reviewing and commenting, and I audit its GitHub footprint by API afterwards — for the three passes that ran while reviewing this PR, that audit came back with zero writes.

Srinivasan8888:
Thank you — and on the two things you put to me directly:

**The CONTRIBUTORS text is accurate.** I read `163b46a` (the one the automated pass wrote) line by line against what I actually did, since you offered to rewrite it. It holds: the warning phase, the one-report-per-load with the v0.75 deadline, the sweep `--param` failing before any arm, and specifically *"the crucial repair was above the helper: the first guard was caught by the per-arm exception loop and exited 0, so the final version validates the grid before that loop"* — that is exactly the defect and exactly the fix, not a generous paraphrase of it. The mutation numbers are yours, not mine, and the 37-test figure was correct at that commit. Nothing needs rewriting. I would rather say that plainly than let a standing offer to correct the record go unanswered.

**Your precision on the loader finding is a correction I am taking.** I wrote "both findings are live on main", which is true and misleading in the same sentence. You are right that a reader lands on "main ships a silent-drop loader" — it does not. The wiring at `loader.py:68` and `:103` was always present and always worked; what shipped uncovered was the guarantee that it keeps working. Regression versus unwitnessed-correctness is the distinction that decides urgency, and flattening the two into "live" cost that. I have adjusted how I describe it.

**#642 is @umran666's** — they claimed it at 11:38Z, after #643 merged at 09:44Z. I had said I would take it, but only on #628 and in #643's body, never on #642 itself, which is the overlap-invisible-in-a-thread failure `CONTRIBUTING.md` describes and I walked into it. I have stood down there and handed over the analysis I already had rather than sit on it: the `--dry-run` branch returns at `sweep.py:106`, before `load_config` at `:116`, so both of their first two criteria reduce to moving that load — and doing so also makes a dry run surface plain schema-validation failures it currently exits 0 on, which is more than the issue title promises and worth stating in their PR body.

On the process change: auditing the agent's GitHub footprint by API rather than reading its report is the right control, because it does not depend on the thing being audited telling the truth. The instruction-versus-control distinction you drew is the part I will remember.

Nothing outstanding from me on this PR.

Issue #622, opened 2026-08-31
Layer streaming: `stream_source: auto` picks the RAM tier without seeing the tmpfs ceiling — 16.10 GB store on a 16 G /dev/shm is an OOM kill with 27 GB free


NO0BDEV:
## What

`stream_source: auto` chooses the RAM tier from a free-RAM check that does not account for the **tmpfs ceiling the pinned store actually writes into**. On a box where `/dev/shm` is smaller than the store, every run is killed by the kernel OOM killer before the first step.

Qwen2.5-32B-Instruct, NF4: the pre-flight sizes the store at **16.10 GB**. `/dev/shm` on this machine is **16 G**. It misses by ~100 MB, fills tmpfs (unevictable) while the page cache holds the 18 GB shard set it is reading, and the process is killed.

## Reproduction

RTX 5070 Ti 16 GB (sm_120) · 30 GB RAM · NVMe · Ubuntu 26.04, kernel 7.0.0-30-generic
soup-cli 0.73.3 · torch 2.10.0+cu128 · transformers 4.57.6 · peft 0.20.0 · trl 0.28.0 · bitsandbytes 0.50.2 · Python 3.12

```yaml
base: Qwen/Qwen2.5-32B-Instruct
task: sft
backend: transformers
data:
  train: ./smoke.jsonl     # 36 rows
  format: chatml
  max_length: 2048
training:
  epochs: 1
  lr: 2e-4
  batch_size: 1
  gradient_accumulation_steps: 8
  quantization: 4bit
  gradient_checkpointing: true
  stream_layers: true
  stream_source: auto
  stream_buffers: 2
  stream_vram_probe: true
  lora: {r: 16, alpha: 16}
output: ./out-32b
```

```
$ df -h /dev/shm
tmpfs            16G     0   16G   0% /dev/shm
```

Pre-flight, verbatim:

```
Layer streaming BETA — arch qwen2, tier ram
  base store   16.10 GB across 64 layers (pinned)
  VRAM buffers 2 x 252 MB = 503 MB
  resident     3114 MB embeddings + adapters
  peak VRAM    ~10.27 GB at batch 1 x seq 2048 (logits 4.36 GB)
  free VRAM    15.72 GB
```

Then:

```
./run.sh: line 13: 830740 Killed    soup train --config soup.yaml -y
=== EXIT 137 in 147s ===
```

`journalctl -k`:

```
Out of memory: Killed process 830740 (soup) total-vm:54546888kB,
[...]

PR #644, opened 2026-09-02
fix(stream): add physical RAM ceiling (#622)

ousamabenyounes:
## What does this PR do?

Refs #622 by adding a physical-host RAM ceiling to layer-streaming tier selection. `stream_source: auto` now keeps the RAM tier only when the stream store fits dynamic free-RAM headroom and the store plus resident extras stays under 55% of physical RAM; otherwise it falls back to the NVMe disk tier. Forced `stream_source: ram` refuses the same unsafe case during pre-flight with an actionable message.

Also carries that ceiling through Qwen4 PLE auto selection, with resident extras counted against both the dynamic free-RAM budget and the physical-host ceiling. This intentionally does not add a `/dev/shm` statvfs guard; the current issue discussion points to CUDA pinned-host allocation and driver accounting rather than the tmpfs mount as the binding allocation channel.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

## Validation

- `./run-ci.sh`
- Targeted pytest: 17 passed
- Targeted coverage: Python coverage JSON generated
- Diff coverage: 44/44 changed executable lines covered
- Revert proof: `tests/test_v07203.py::TestPhysicalRamBudget` and `test_qwen4_ngram_auto_counts_resident_ram_against_free_budget` fail against the pre-fix production code

Full `pytest tests/ -v` was not run in this checkout because optional training dependencies are not installed here (`torch`, `peft`, `safetensors`, `typer`); the broader layer-streaming run fails on those missing modules before reaching this patch's behavior.


src/soup_cli/utils/layer_stream.py
```diff
@@ -751,7 +758,14 @@ def choose_tier(
...
     """
-    if model_bytes < free_ram_bytes * headroom:
+    physical_limit = (
+        None
+        if total_ram_bytes is None
+        else total_ram_bytes * PHYSICAL_RAM_TIER_HEADROOM
+    )
+    physical_bytes = store_bytes + int(resident_bytes)
+    fits_physical_ram = physical_limit is None or physical_bytes < physical_limit
+    if store_bytes < free_ram_bytes * headroom and fits_physical_ram:
```
AmirF194: Ran this against d2bf5db (PR head) in a clean container.

choose_tier's dynamic free-RAM check here (`store_bytes < free_ram_bytes * headroom`) dropped resident_bytes. Before this PR the same check compared `model_bytes` (host_store_bytes + embed_bytes) against `free_ram_bytes * headroom`, so resident extras counted against currently-free RAM. Now resident_bytes only feeds the new physical ceiling (a fraction of total RAM), never the free-RAM check. When total RAM is large but currently-free RAM is tight, the RAM tier gets picked for a store+resident combination that does not fit in what is actually free right now.

```
free_ram=10GB, total_ram=100GB, host_store_bytes=5GB, resident_bytes=3GB
choose_tier(...) -> "ram"
old check (store+resident < free_ram*0.7): False   # 8GB > 7GB
```

build_stream_plan reaches the same result end to end (tier="ram", no fallback note) with the same inputs.

_resolve_qwen4_ngram_source in stream_setup.py has the same gap: `ram_bytes = base_in_ram + ngram_bytes` also skips resident_ram before comparing to ram_budget.

Folding resident_bytes into the free-RAM comparison too, not just the physical one, would keep the pre-PR guarantee alongside the new ceiling.

MakazhanAlpamys (changes requested):
**The code is good and the tests are the strongest part. My problem is with what it claims, not what it does** — and the gap is one word in the PR body plus a decision about #622.

## Verified here

`239 passed, 12 skipped` across `test_v07203.py`, `test_issue602_qwen4_ple_streaming.py` and `test_cli_startup_is_light.py`. CI 14/14. Docs updated in both the honest-scope block and the troubleshooting list, and the wording there is careful — it says *"Soup's physical-host ceiling"*, not *the tmpfs ceiling*, which is accurate.

**Four mutations of the ceiling, all killed:**

| mutation | result |
|---|---|
| `fits_physical_ram = True` (ceiling removed) | killed — 3 tests, incl. `test_auto_falls_to_disk_when_physical_ram_budget_is_exceeded` |
| `0.55` → `0.99` | killed — 5 tests, incl. `test_forced_ram_refuses_when_physical_ram_budget_is_exceeded` |
| comparison inverted | killed — 5 tests, and `test_physical_ram_budget_keeps_ram_when_store_is_small` is the one that makes this a real control rather than a one-sided assertion |
| `physical_bytes = store_bytes` (resident dropped) | killed — 3 tests |

The eight removed test lines are a fixture rework in `test_issue602`, not lost coverage. `test_cli_startup_is_light` still green, no top-level torch introduced.

**And it does fix the reported case.** Reporter's numbers: store 16.10 GB + resident 3.114 GB = 19.21 GB against 55% of 30 GB = 16.50 GB → falls back to disk. Correct outcome.

## The claim I want changed: `Closes #622`

#622 asks for something specific:

> read the size of the tmpfs the pinned store is allocated from (`/dev/shm`, or `RLIMIT_MEMLOCK` where that is the binding limit) and require the store to fit it

This PR reads neither. `grep` over your branch for `shm`, `tmpfs`, `MEMLOCK` returns nothing. It substitutes a fraction of *physical RAM* as a proxy for a *tmpfs* limit, and the proxy is not conservative — **Linux's default `/dev/shm` is 50% of RAM and your ceiling is 55%**, so it sits above the very thing it stands in for.

That leaves the same failure reachable. The store clears your guard when `store + resident < 0.55 × RAM`, and overflows a default tmpfs when `store > 0.50 × RAM`; both hold whenever `RAM > 20 × resident`:

```
RAM  32 GB | /dev/shm 16.0 GB | store 14.60 + resident 3.0 -> guard passes, tmpfs fits
RAM  64 GB | /dev/shm 32.0 GB | store 32.20 + resident 3.0 -> guard passes, tmpfs OVERFLOWS by 0.20 GB
RAM 128 GB | /dev/shm 64.0 GB | store 67.40 + resident 3.0 -> guard passes, tmpfs OVERFLOWS by 3.40 GB
```

A ~64 GB NF4 store on a 128 GB box is not an exotic configuration for this feature — it is the regime layer streaming exists for.

The sharper case is a container. Docker's default `--shm-size` is **64 MB**; a physical-RAM check on a 64 GB host passes every store, and the OOM the issue describes happens anyway with no Soup-side output. I have not reproduced that myself and I am not claiming it as measured — but the reporter's kernel log (`shmem-rss: 11653136kB`, `anon-rss` under 300 MB) is evidence the store really is shmem-backed on their box, and a physical-RAM ceiling cannot see a per-container tmpfs at all.

**So: `Refs #622`, not `Closes #622`.** This is a genuine improvement that prevents the reported instance; it is not the ceiling the issue asked for. I would rather merge it honestly labelled than close an issue whose own suggested fix is still unimplemented.

## Two questions, not demands

1. **Where does 55% come from?** The comment gives a mechanism ("pinned host memory is unevictable while shard reads pressure the page cache") but no derivation — not a measurement, not a citation. That is fine for a safety margin as long as it is *called* one; what makes me uneasy is a number that looks derived, sits one notch above the Linux tmpfs default, and will be read by the next person as if it means something. Either say in the comment that it is a chosen margin rather than a measured bound, or pin it to something real.
2. **Do you want to add the `os.statvfs("/dev/shm")` read now, or leave it?** It is roughly ten lines on Linux (`f_blocks * f_frsize`, fall back to the current behaviour where the path does not exist), and it would let this genuinely close #622 rather than narrow it. Entirely your call — if you would rather ship this and stop, I will file the tmpfs read as a follow-up and credit you for the groundwork. Both are good outcomes; I am not asking you to expand scope you did not sign up for.

## One line of hygiene

`changelog.d/0.73.3/644.fixed.md` ends `(#622, #644)`. The convention in that directory is `(#622 by @ousamabenyounes in #644)` — see `635.fixed.md` and `468.added.md`. I have been asking every contributor for this, so it would be unfair to skip it here.

Send the `Refs` change and the credit line and I will merge; answer question 2 either way and I will act on it.

MakazhanAlpamys:
@ousamabenyounes — a heads-up you should have from me rather than discover on your own.

A second implementation of #622 was opened as #646 about two and a half hours after yours. The author claimed the issue without seeing this PR; the issue carried no assignee, which is my process gap, not their mistake. **I have held #646** and told them why: I asked you a question about exactly that scope an hour before they opened it, and I am not going to withdraw it because someone else arrived. Answering a maintainer's question should not be the slower path.

So the question from my review stands and nothing has changed about it — take the time you need:

> Do you want to add the `os.statvfs("/dev/shm")` read now, or leave it? Entirely your call.

Both answers are fine and neither costs you the merge. If you would rather ship what you have, say so and I will take the `Refs #622` + credit line and merge this, then decide separately what happens with the tmpfs ceiling. If you want the extra ten lines, take them.

One thing from #646's claim comment worth flagging to you, because it bears on your PR whichever way this goes: its author said they would read the allocation path to settle whether the store is `cudaHostAlloc` anonymous page-locked memory or shared-memory backed. That is the same uncertainty I refused to claim as settled in my review of yours — `RamSource` uses `torch.empty(..., pin_memory=True)`, which argues anonymous, while the reporter's `shmem-rss: 11653136kB` against `anon-rss` under 300 MB argues the other way. If that gets answered on #622, it may change what the right ceiling is for both of us, and I would rather you have that context while deciding than after.

MakazhanAlpamys:
@ousamabenyounes — following up, and adding a deadline I should have set at the start.

My question from ~20 hours ago still stands and is genuinely open: **do you want to add the `os.statvfs("/dev/shm")` read, or ship what you have?** Both answers are fine and neither costs you the merge.

**If I have not heard from you by roughly 2026-09-04 12:00 UTC**, I will take the second answer on your behalf — merge this with `Refs #622` plus the credit line once you push those two, or, if you have gone quiet entirely, decide between this and #646 on the merits and say publicly which I picked and why.

The reason for the clock is not impatience with you. #646 has been held for you since yesterday, and its author has been waiting without knowing for how long. Holding it open-ended costs a contributor who did the work, so an indefinite hold is not a neutral choice — that is on me for not putting a date on it when I created the hold.

Nothing has changed about the substance: your PR is first, and it stays first for as long as this window runs.

ousamabenyounes:
Thanks for the review. Addressed the resident-RAM accounting gap in a82a947: the stream store plus resident extras now has to fit the dynamic free-RAM headroom for both base-tier selection and Qwen4 PLE auto selection, while the existing physical-RAM ceiling stays in place.

I left the `/dev/shm` statvfs guard out intentionally. The current issue-thread analysis points to CUDA pinned-host allocation / driver accounting rather than the `/dev/shm` mount as the allocation channel, so this keeps the mechanism-agnostic free-RAM + physical-RAM ceilings and labels the PR as `Refs #622`.

Validation: reproduced the review counterexamples RED (4 failures with production fix reverted), then reran the focused stream gate GREEN. Local `run-ci.sh`: baseline ruff + 13 tests passed; final ruff + 17 tests passed; Python diff coverage 44/44 changed executable lines covered.

AmirF194:
Confirmed, a82a947 covers the gap I found. Thanks for the fix and the red/green validation.

MakazhanAlpamys:
Merged as `171e914`, credited in CONTRIBUTORS at `44533dc`.

Credited for **what you refused to add**, which is the part I want on the record. Declining the `statvfs` read with a reason — rather than adding a plausible-looking check because a maintainer floated it — is the harder call, and it was the right one: `/dev/shm` is not the allocation channel, so a check against it would be wrong in both directions and would not have bounded the reporter's kill in the first place.

That the bound you shipped is **mechanism-agnostic** is what makes it correct while the mechanism is still an inference. A competing implementation converged on the same one independently, which is evidence rather than coincidence.

And the 55% comment is now better than what I asked for. I asked whether it was measured or chosen; you documented it as a chosen safety margin **and had the comment name the hole it stands in for** — that a container's `--shm-size` can be far smaller, so a store clearing this ceiling can still overflow the mount it is allocated from. The next reader cannot mistake it for derived.

Verified before merging: **7 mutations, 7 killed** (Qwen4 PLE ceiling, base-tier ceiling, the constant, the comparison direction, the forced-`ram` refusal, and resident extras dropped from each of the two budgets), 15/15 CI including `pytorch-smoke` on the current head, 8 removed test lines all fixture rework — one of them a strengthening, since `extras_resident_bytes: 0` had made the resident term invisible. No new top-level torch in the planner.

**One finding recorded in the merge commit rather than held against you**, because it is a test-quality regression rather than a code one. After `a82a947` added resident extras to the *dynamic* check, the #622 repro fixture now fails that check first — so the physical ceiling is no longer the rejecter in any shipped fixture. Measured:

```
#622 fixture (store 16.10 + resident 3.114, free 27, total 30 GB)
  free-RAM gate:  False   (need < 18.90, have 19.214)
  physical gate:  False   (need < 16.50)
```

Consequence: removing the ceiling entirely leaves `test_auto_falls_to_disk_when_physical_ram_budget_is_exceeded` **passing** — only the message assertion catches it. The ceiling is still reachable and correct in its own regime, which I confirmed directly:

```
total 30, free 28, store 15 + resident 3
  free-RAM gate:  True    -> ceiling is the only rejecter
  choose_tier  -> 'disk'
```

Nothing exercises that regime. One extra `choose_tier` assertion at a high free-RAM fraction restores the behavioural kill. Not blocking — mutations of the constant and of the comparison still die — but worth having, and yours if you want it.

Two people shaped this beyond the patch, and I have said so in the merge commit since the fragment convention here is single-author: @umran666 settled the mechanism question and **named the Qwen4 PLE gap that was in this PR and absent from their own competing one**, and @AmirF194 found the resident-RAM accounting gap that became `a82a947`.

`Refs #622`, not `Closes` — #622 stays open until the tmpfs follow-up and #655's measurement are settled.

PR #645, opened 2026-09-02
fix(sweep): --dry-run validates the config and sweep params instead of returning before loading

### LINKED ISSUE #642: soup sweep --dry-run still validates nothing: a typo'd config key is silent there while soup train --dry-run now reports it

umran666:
Closes #642 (claimed in [this comment](https://github.com/MakazhanAlpamys/Soup/issues/642#issuecomment-5508929847)).

## The decision, said out loud

`--dry-run` on `sweep` changes meaning, from "print the grid without loading anything" to **"validate, then print the grid without running anything."** Loading a config file executes nothing; a dry run whose purpose is catching mistakes before a long run cannot catch the exact mistake the issue reproduces while it refuses to read the file. Stated in the fragment too, since it is a behaviour change, not a bug fix alone.

One consequence on the **execution** path, also stated rather than left to the diff: validation now happens before the confirmation prompt instead of after it. Same reports, same exit codes, earlier â€” the operator sees the unknown-key warning before answering "Start N training run(s)?", and a config that hard-fails schema validation no longer asks to be confirmed first.

## What

`sweep()` moved two existing blocks above the plan-panel printing, sharing one load:

1. `base_cfg = load_config(config_path)` â€” the same loader `train --dry-run` uses, so the unknown-key warning block is byte-identical by construction (criterion 1), including the v0.75 deadline sentence and the "An unapplied key is ignored" follow-up.
2. The #628 probe guard â€” a typo'd `--param` now refuses **before printing a grid it can never run** (criterion 2), under `--dry-run` and on the execution path alike.

Severity split untouched (criterion 3): warn-only config exits 0, hard param error exits 1, schema-invalid config exits 1 via the loader's existing `SystemExit` â€” no third behaviour invented. The dry-run exit line prints after validation, where it always printed.

## The latent crash this exposed — scoped to what reproduces (review round 2)

Making the pre-flight reachable under `--dry-run` made it reachable **at all** on this box, and it died:

```
TypeError: issubclass() arg 1 must be a class
  unknown_keys.py:92 in _nested_models
```

The first cut of this PR claimed "every real sweep on the oldest supported Python" died. The review could not reproduce that — **correctly** — and the exact reproduction is narrower and more interesting. Two conditions, BOTH required:

1. **Python 3.10**, where a PEP-585 alias (`dict[str, float]`) passes `isinstance(c, type)` (3.11 made isinstance return False). The trigger field is `training.preference_loss_weights`, which every full `model_dump()` walk visits.
2. **A pydantic whose `ModelMetaclass` inherits `__subclasscheck__` from `ABCMeta`** — `abc.__subclasscheck__` is what raises on a non-class. Verified under the SAME 3.10.11 interpreter, only pydantic varying:

```
pydantic 2.10.6: '__subclasscheck__' in ModelMetaclass.__dict__ -> False
                 issubclass(dict[str, float], BaseModel)        -> TypeError
pydantic 2.13.4: '__subclasscheck__' in ModelMetaclass.__dict__ -> True
                 issubclass(dict[str, float], BaseModel)        -> ok
```

So: on 3.10 with pydantic below the override — a wide slice of the supported `pydantic>=2.0.0` range, including what this box resolves — every real sweep did die with that TypeError after the confirm. On the review's box (3.10.8 + pydantic 2.13.4) and on every CI cell (which install current pydantic), the alias is tolerated and nothing crashes. The discriminating variable is pydantic, not the Python patch level. The fragment and this body now say exactly that, and the filter stays: cheap, correct on every combination, and the only thing standing between a 3.10 user on an older pydantic and a dead sweep. Fix: `typing.get_origin(c) is None` before the `issubclass` in `_nested_models`.

## Tests (criterion 4)

`tests/test_issue642_sweep_dry_run.py` â€” 12 tests, every one running the real command through `CliRunner` (or the real walker on real dumps) and asserting on output/exit codes. **No source-text assertions**, per the #628 review note. Branches: warning block content + deadline + printed-once, grid still prints for warn-only, hard param (top-level and nested) refuses without a grid, schema-invalid exits 1, clean-config control emits nothing, execution-path ordering (warning before prompt), execution path loads once (stubbed `_run_single` â€” the contract is one-load-one-warning, not training), and the 3.10 regression with its own control (an injected `training.quantizaton` must still be *found* in a full dump â€” "never crash" must not become "never walk").

## Mutations â€” shown, not asserted

```
M1 dry-run exits before validation (pre-fix)   7 failed by name
M2 guard moved back after the grid printing    2 failed by name (the no-grid assertions)
M3 get_origin filter removed                   9 failed on this 3.10 box (pydantic 2.10.6 — the crash returns)
M4 second load_config on the execution path    SURVIVED the first cut -> killed by name
   after adding the stubbed-_run_single test (the cancelled-at-prompt test never
   reaches past the confirm, so the mutant hid from it)
baseline                                       12 passed
```

M4 surviving and being closed is reported rather than quietly re-run: the first test set had a hole exactly where the house mutation rule looks.

## Issue reproduction, against the real CLI on this box (Python 3.10)

```
$ soup sweep --config soup.yaml --param "training.lr=1e-5,2e-5" --dry-run
Warning: unknown config key 'training.quantizaton' - did you mean
'quantization' or 'quantization_aware'? Not applied.
An unapplied key is ignored, not defaulted -- the run proceeds as if you had
not written it.
[... Sweep Plan panel + Parameter Grid ...]
Dry run - no training will be executed.
$ echo $?
0

$ soup sweep --config soup.yaml --param "lora_rank=8,16" --dry-run
sweep parameter does not match any config field: unknown config key 'lora_rank'
- not applied.
$ echo $?
1        # no grid printed
```

M3 is the mutation the review saw SURVIVE on a current-pydantic box. Round 2 closes it in the masked environment itself — a venv on the same 3.10.11 interpreter with pydantic 2.13.4:

```
venv 2.13.4, filter in place     14 passed
venv 2.13.4, filter removed       1 failed, 13 passed
    FAILED ...::test_the_filter_kills_the_crash_whatever_pydantic_does   <- killed by name
```

Every pre-round-2 test passes with the filter removed there (exactly the review's observation); the new killer is what dies. On 3.11/3.12 the hazard is gone at the isinstance layer and the version-branched test documents that rather than pretending to kill.

## Verification

- `tests/test_issue642_sweep_dry_run.py` + `tests/test_sweep.py` + `tests/test_issue627_unknown_config_keys.py`: **89 passed** (the two pre-existing sweep dry-run CLI tests included â€” their minimal fixtures load clean, no edits needed).
- `tests/test_cli_subprocess.py -k sweep`: 3 passed.
- Round 2: the 642 file passes under `FORCE_COLOR=1 TERM=xterm-256color` (14 passed) — the colour-hostile assertion now strips; the three affected suites together: **91 passed**.
- Zero tests deleted or weakened; `test_sweep.py` and `test_issue627_unknown_config_keys.py` untouched.


MakazhanAlpamys (changes requested):
**The fix is right and every mutation of it dies. I am blocking on one sentence in the changelog fragment, because it is a factual claim that ships into `CHANGELOG.md` and I could not reproduce it.**

## The fix itself — verified against the real CLI

All five acceptance criteria from #642 are met, measured rather than read:

| command | exit | output |
|---|---|---|
| `--dry-run` over a config with `training.quantizaton` | **0** | the same warning block `train --dry-run` prints, deadline sentence included, then the grid |
| `--dry-run --param "lora_rank=8,16"` | **1** | refusal before the Sweep Plan panel — **no grid printed** |
| `--dry-run` on a clean config (control) | **0** | grid only, no warning |
| the same typo'd config on `main` | 0 | grid, no warning — the bug |

The severity split is what I asked for and the hard part to get right: warn-only stays 0, the sweep-parameter error is 1. Twelve tests, all through `CliRunner` on the real app, **zero source-text assertions** — after three rounds of that anti-pattern in this area, that matters. The clean-config control is present, so a "warn on everything" wiring cannot pass. Zero deleted test lines.

Mutations, four of the fix and one adjacent:

| mutation | result |
|---|---|
| restore the pre-fix early return (validate only when `not dry_run`) | **killed**, 5 tests |
| skip the `--param` guard under `--dry-run` only | **killed**, 2 tests |
| `--dry-run` exits 1 for a merely-warning config | **killed**, 4 tests — including two *pre-existing* `test_sweep.py` guards |
| demote the hard `--param` error to a warning | **killed**, 5 tests |
| remove the `get_origin` filter in `unknown_keys.py` | **SURVIVED** |

And the decision #642 asked to be made out loud is made out loud, in the PR body and in the fragment. You also volunteered a second-order consequence I did not ask about — validation now precedes the confirmation prompt on the execution path. That is the right instinct.

## The blocker: the Python 3.10 crash claim does not reproduce

The fragment states, in a file that ships verbatim to users:

> on Python 3.10 a `list[...]`/`dict[...]` annotation passes `isinstance(c, type)` and then raises `TypeError` from `issubclass`, so **every real sweep on the oldest supported Python died** with `issubclass() arg 1 must be a class`

On this box — Python **3.10.8**, pydantic **2.13.4** — I could not make that happen on unmodified `main`:

```
find_unknown_config_keys(cfg.model_dump())              -> []   (no crash)
... with training.lora.rank_pattern={q_proj: 8}         -> []   (no crash)
... with training.unfrozen_parameters=['.*mlp.*']       -> []   (no crash)
soup sweep --config clean.yaml --param training.lr=... --yes   -> EXIT 0
```

That last line is the one that decides it. Since #628/#643 the pre-flight **is** reachable on the execution path, so a real sweep on `main` under 3.10 exercises exactly the walk you say crashes — and it does not. Your own mutation of the filter (removing it) also leaves the suite green here, which is consistent: `TestTheFullDumpWalkSurvivesPython310` passes with and without the fix, so the claim has no test behind it either.

The mechanism is real in principle — a PEP-585 `list[str]` *is* a `type` on 3.10 and `issubclass` does raise on it — and the filter is correct, cheap and harmless, so **keep it**. What I cannot ship is "every real sweep died" when a real sweep does not die. Either:

- post the exact reproduction (Python patch version, pydantic version, the config, the traceback) and I will take the claim as written; or
- soften it to what is demonstrable — a latent crash the filter forecloses — and drop the "every real sweep" sentence.

If you can reproduce it, that is more valuable than the wording fix: it would mean a config shape neither of us has characterised reaches the walker, and I would want a test for that shape rather than a filter that silently prevents it.

## Two smaller items

**ANSI.** `test_the_warning_carries_the_loaders_deadline` fails under `FORCE_COLOR=1 TERM=xterm-256color`:

```
assert 'v0.75' in "...Soup v0.\x1b[1;36m75\x1b[0m will reject unknown config keys..."
```

Rich's *number* highlighter splits the token — escapes **inside** `v0.75`. That is the failure class that has turned this CI red four times, arriving through a styler that is not the help table. CI is green because `CliRunner` output is not a tty, and I will note honestly that a pre-existing test in `test_issue627_unknown_config_keys.py` fails the same way on `main`, so this is house-wide exposure rather than a regression you introduced. Still worth routing that one assertion through the standard `re.sub(r"\x1b\[[0-9;]*m", "", result.output)` while you are here — your other new assertions survive colour, because their escapes land around the token rather than inside it.

**Docs.** `docs/backends-and-ops.md:408` still labels the example `# Preview without running`, and nothing under `## Hyperparameter Sweep` says `--dry-run` now validates. One line closes it; a changed user-visible contract should be discoverable outside the changelog.

## Verification

`test_sweep.py` + `test_issue627_unknown_config_keys.py` + `test_issue642_sweep_dry_run.py` + `test_cli_help_assertions_are_ansi_safe.py` → **98 passed**. `ruff` clean. CI 14/14. Fragment named `645.fixed.md` with the credit line `(#642 by @umran666 in #645)` in the neighbours' convention — thank you for doing that without being asked this time.

Fix the claim and I will merge; the ANSI strip and the docs line are cheap enough that I would take them in the same push.

umran666:
Round 2 at `000d6b3` — all three items, and the reproduction you asked for. It reproduces, with a twist that makes your non-reproduction and mine the same fact: **the discriminating variable is pydantic, not the Python patch level.**

## The exact reproduction

**Box**: Windows, Python **3.10.11**, pydantic **2.10.6**, soup at `90093ff`. Config: any clean config — the crash does not need a typo, because the probe walks a full `model_dump()`. Trigger field, found by scanning every model field for the raising shape: **`training.preference_loss_weights`**, annotated `Optional[dict[str, float]]`.

```
$ python -c "... load_config(clean).model_dump() -> _set_nested_param(probe,'lr',1e-5) -> _reject_unknown_sweep_params(probe)"
  File "src/soup_cli/config/unknown_keys.py", line 137, in find_unknown_config_keys
    _walk(raw, SoupConfig, "", found)
  File "src/soup_cli/config/unknown_keys.py", line 126, in _walk
    _walk(value, nested, f"{path}.", found)
  File "src/soup_cli/config/unknown_keys.py", line 92, in <listcomp>
    if isinstance(c, type) and issubclass(c, pydantic.BaseModel)
  File ".../abc.py", line 123, in __subclasscheck__
TypeError: issubclass() arg 1 must be a class
```

**Why your box does not do it.** Two conditions, both required — and yours fails the second:

```
same 3.10.11 interpreter, only pydantic varying:

pydantic 2.10.6: '__subclasscheck__' in ModelMetaclass.__dict__ -> False   (inherits ABCMeta's, which raises)
                 issubclass(dict[str, float], BaseModel)        -> TypeError
pydantic 2.13.4: '__subclasscheck__' in ModelMetaclass.__dict__ -> True    (defines its own, tolerant)
                 issubclass(dict[str, float], BaseModel)        -> ok
```

The alias quirk (`isinstance(dict[str, float], type) is True`) holds on your 3.10.8 too — you said as much — but on 2.13.4 the alias never reaches `abc.__subclasscheck__`. That also explains your `rank_pattern` probe coming back `[]` and why every CI cell stayed green: CI installs current pydantic. So "every real sweep on the oldest supported Python died" was true across the unmasked part of the supported `pydantic>=2.0.0` range and false on a current one. The fragment's headline now reads "**forecloses a latent Python 3.10 crash**" and the body of the fragment carries the verified version matrix; the PR body's crash section is rewritten to the same scope. Per your note that a reproduction would be worth more than the wording fix — the config shape you asked about is characterized: it is not exotic, it is *every* config, via one dict-annotated field, on 3.10 + pre-override pydantic.

## The mutation that survived on your box now dies on it

`TestTheFullDumpWalkSurvivesPython310` was indeed environment-dependent — on your stack it passed with and without the filter, so the claim had no test behind it there. Replaced by `TestTheGenericAliasFilter` with a **deterministic killer**: an ABCMeta stand-in monkeypatched over `pydantic.BaseModel` recreates the pre-override metaclass, and `_nested_models(TrainingConfig, "preference_loss_weights")` must return `[]` — so filter removal raises `TypeError` on **any 3.10 box regardless of installed pydantic**, and the real schema field is pinned (renaming/retyping it fails the test too). Proven in the masked environment itself — a venv on the same interpreter with pydantic 2.13.4:

```
filter in place:   14 passed
filter removed:     1 failed, 13 passed
  FAILED ...::test_the_filter_kills_the_crash_whatever_pydantic_does   <- by name
```

— every pre-round-2 test passing with the filter removed in that venv, exactly your observation, with the new killer the only thing that dies. A second new test pins the hazard stdlib-only, version-branched: on 3.10 the isinstance quirk and the ABCMeta `TypeError` are asserted directly; on 3.11+ it asserts `isinstance(alias, type)` is False, documenting that the filter is belt-and-braces there rather than pretending to kill. 12 → 14 tests.

## ANSI

Routed through `tests/conftest.strip_ansi` (the #633 helper) in `test_the_warning_carries_the_loaders_deadline` — the whole file now passes under `FORCE_COLOR=1 TERM=xterm-256color` (14 passed), and the assertion that failed for you (`v0.75` split by the number highlighter) is the one that's stripped. Your honest note about the pre-existing `test_issue627_unknown_config_keys.py` failure under colour: left alone here — it predates this PR and widening into it is your call, not mine to make silently.

## Docs

`docs/backends-and-ops.md`'s sweep example now reads:

```bash
# Preview without running — validates first (#642): unknown config keys emit
# the same warning block `train --dry-run` does, and a --param naming no config
# field exits non-zero before any grid is printed
```

## Verification (round 2)

- `tests/test_issue642_sweep_dry_run.py` + `test_sweep.py` + `test_issue627_unknown_config_keys.py`: **91 passed** (was 89; +2 new).
- The 642 file under `FORCE_COLOR=1 TERM=xterm-256color`: 14 passed.
- `ruff` clean; `unknown_keys.py` restored bit-identical after every mutation run (`git diff` empty).
- Fragment corrected in place — the claim that ships is now the claim that reproduced.

MakazhanAlpamys (approved):
**You were right and I was measuring a masked environment. Approving — merging once CI finishes.**

I reproduced your explanation on my own box rather than accepting it, because I had published the non-reproduction and owed the check:

```
python 3.10.8 | pydantic 2.13.4
ModelMetaclass defines __subclasscheck__: True
isinstance(dict[str,float], type):        True      <- the alias quirk IS present here
issubclass(alias, BaseModel):             False     <- ...and never reaches ABCMeta
```

That is the whole thing. The quirk your fix targets is live on my interpreter; what my box lacks is a pydantic that lets it reach `abc.__subclasscheck__`. So my "does not reproduce" and your "reproduces" were the same fact seen through different `pydantic` pins, and **the discriminating variable was never the Python patch level** — which is what the original wording claimed and why I could not confirm it.

`training.preference_loss_weights: Optional[dict[str, float]]` as the trigger field is the part that turns this from a plausible mechanism into a characterised one. It is not an exotic config: the probe walks a full `model_dump()`, so it is *every* config, through one dict-annotated field, on 3.10 with a pre-override pydantic. That is a better answer than the wording fix I asked for, which is what I hoped for when I said a reproduction would be worth more.

**The deterministic killer works on the masked box, which is the point of it.** I removed the `get_origin` filter here, where the old test was vacuous:

```
1 failed, 13 passed
FAILED ...::TestTheGenericAliasFilter::test_the_filter_kills_the_crash_whatever_pydantic_does
```

Exactly one test, by name, on a stack where every pre-round-2 test passed with the filter gone. Monkeypatching an ABCMeta stand-in over `pydantic.BaseModel` to recreate the pre-override metaclass is the right move: it makes the guard's death independent of what happens to be installed, which is the only way this stays honest after pydantic changes again. Pinning the real schema field so a rename or retype also fails is the detail I would have forgotten.

Version-branching the stdlib hazard test — asserting the quirk on 3.10 and asserting `isinstance(alias, type) is False` on 3.11+ — is the honest form. It documents that the filter is belt-and-braces on newer interpreters instead of pretending to kill something that cannot happen there.

**The other two items are closed as well.** Under `FORCE_COLOR=1 TERM=xterm-256color` the file now runs **14 passed** where it was `1 failed` before, so Rich's number highlighter no longer splits `v0.75` out from under the assertion. `docs/backends-and-ops.md` is in the diff. The fragment now says "forecloses a latent Python 3.10 crash" instead of "every real sweep died", which is the claim the evidence actually supports.

Zero deleted test lines against `main`; 12 tests became 14.

**One piece of housekeeping that is mine, not yours.** I made `pytorch-smoke` a required check earlier today (it landed in #601), so every PR whose last CI run predates that shows `BLOCKED` with 14/14 green — the job simply did not exist when they ran. I have updated this branch to pick it up and it is running now; nothing for you to do, and the same will apply automatically to any branch on its next push.

Once it is green this merges.

MakazhanAlpamys:
Merged as `ac20a26`, credited in CONTRIBUTORS at `36541a3`.

Recorded there as the round-two investigation rather than the fix, because that is the part worth copying: I blocked on a claim I could not reproduce, and instead of softening the wording you found **why** both observations were true — pydantic's `ModelMetaclass` gaining its own `__subclasscheck__`, which masks the PEP-585 alias hazard on a current stack. Verified on my own box before merging: the quirk is live on my interpreter and `issubclass` returns `False` rather than raising, exactly as you said.

Final checks on the merged head: real CLI gives exit 0 / 1 / 0 for the warn, hard-error and clean cases with no grid on the hard error, 14 passed under `FORCE_COLOR=1 TERM=xterm-256color`, zero deleted test lines, and `pytorch-smoke` green after I updated the branch.

PR #646, opened 2026-09-02
fix(stream): auto tier sees the physical RAM ceiling — the #622 exit-137 OOM becomes a disk-tier fallback

### LINKED ISSUE #622: Layer streaming: `stream_source: auto` picks the RAM tier without seeing the tmpfs ceiling — 16.10 GB store on a 16 G /dev/shm is an OOM kill with 27 GB free

umran666:
Closes #622 (claimed in [this comment](https://github.com/MakazhanAlpamys/Soup/issues/622#issuecomment-5508930057)).

## The mechanism dispute, settled from the code — and why the fix does not depend on settling it

The issue proposed sizing the RAM tier against the tmpfs ceiling (`/dev/shm`); @here-2007's comment countered that `/dev/shm` is a correlation, not the mechanism. The claim promised to let the code decide, so:

- **Allocation path**: `RamSource` allocates the store via `torch.empty(..., device="cpu", pin_memory=True)` (`layer_stream_runtime.py`) — the CUDA host allocator (`cudaHostAlloc`). There is no `shm_open`, no `multiprocessing.shared_memory`, no tmpfs file anywhere in the streaming path. The 16 G `/dev/shm` on a 30 GB host is systemd's 50 %-of-RAM default: a numerical coincidence with the 16.10 GB store, as the thread argued.
- **The reporter's telemetry** (`shmem-rss` 11–14 GB, `anon-rss` < 300 MB at the kill) says the *kernel accounted* the driver-pinned pages in the shmem class. That is an accounting label on driver-mediated pinning, not a `/dev/shm` mount bound — and it does not matter for the fix: under **either** reading the pages are unevictable-class memory the kernel cannot reclaim, and that is what saturated the host.
- **Why not `statvfs("/dev/shm")`**: it would force a 128 GB workstation with a small custom mount to the disk tier (false positive) and believe a `--shm-size=64G` container on a 16 GB host (false negative) — the thread's objection, adopted.
- **Why not `RLIMIT_MEMLOCK`**: driver-mediated CUDA pinning is not governed by it. A `getrlimit` gate would see systemd's default 8 MB and force *every* Unix box to a pageable store — contradicting this repo's own measured multi-GB pinned runs (the 7.12 GB page-lock probe in `decide_pinning`'s docstring). The runtime already attempts the pin and falls back loudly; the plan-time ceiling is the missing layer.

## What

`choose_tier` now requires **both** ceilings for the RAM tier — the dual-threshold shape from the thread:

1. `store + resident < available * 0.70` (unchanged), **and**
2. `store + resident < total * 0.55` (new, `RAM_TIER_PHYSICAL_BUDGET`) — 45 % of physical RAM stays for the OS, CUDA runtime and the shard-read page cache.

The reproduction dies on ceiling 2: 16.10 GB store + 3.11 GB resident = 19.21 GB billed vs 16.5 GB budget on a 30 GB host. Precision note, since the issue's own arithmetic compared 16.10 vs 18.9: the real `choose_tier` bills store **+ resident** (19.21 GB), so at exactly 27 GB free the dynamic check was already borderline-failing; the plan-time free figure in run 2 (warm cache, post-sharding) was evidently above the boundary, because the RAM tier was chosen. The unit tests therefore isolate ceiling 2 at 29 GB free — the shape of the bug is "plenty available, absolute ceiling blown", and the fix refuses under either reading.

Behaviour per the issue's acceptance shape:

- **`auto`** falls through to the NVMe disk tier with a note naming the absolute ceiling and its figures — a bare "does not fit in RAM" against a box showing 27 GB free reads as a broken pre-flight, so the note says why free RAM is not the measure. What the reporter measured working (forced disk: exit 0, 9.45 GB peak) is now what `auto` picks.
- **Forced `stream_source: ram`** refuses twice, both naming the ceiling instead of free RAM ("free RAM" is not actionable against a limit no amount of freeing moves): early, from the pre-shard estimate, before minutes of shard I/O; and again at the exact-bytes validator (`_validate_qwen4_ngram_ram_fit`, which also covers `stream_ngram_source: ram`).
- **No SIGKILL-without-output path** is left for the sized-in-advance cases: every refusal is a pre-flight message.
- `total_ram_bytes()` resolves `None` without psutil, leaving the budget unenforced — the existing fail-loudly posture for unmeasurable boxes, unchanged.
- The non-NVMe refusal message gained the binding ceiling when it is the physical budget ("more RAM" was not actionable advice there either); its old shape is untouched when the budget is not involved.

Docs updated where the promise lives: `docs/performance-and-quantization.md`'s "picks RAM when it fits" bullet now states both ceilings with the measured reproduction, and the troubleshooting section gained the "disk tier chosen while plenty of RAM looks free" entry.

## Tests

`tests/test_issue622_stream_tier_physical_budget.py` — 18 tests.

**Unit (14, all runnable anywhere)**: the reproduction's own numbers (19.21 GB billed, 30 GB host, dynamic check passing at 29 GB free) → disk; the backward-compat control (no total → old behaviour, honestly kept for psutil-less callers); the large-host control (same store on 128 GB → RAM; the thread's false-positive concern pinned); the boundary at exactly 55 %; both refusal-message shapes with and without a total; the plan-note selection with its control (dynamic shortfall keeps the original note, no ceiling language); the forced-ram validator's ceiling refusal, its passing control, and its no-total compatibility restated (the #602 pin exercises the same shape).

**Wiring (4)** — the #628 lesson ("nothing crossed the wiring") applied: unit tests of `choose_tier`/`build_stream_plan` stay green if `stream_setup` never passes `total_ram_bytes`, so four tests drive the **real** `_setup_streaming_transformers` over the `test_v07203.py::TestAutoTierFallback` harness pattern (probes patched on the source module, `build_streamed_model` stubbed so nothing reaches a kernel, disk kind pinned so the tier cannot depend on runner hardware): auto + blown budget → `captured["tier"] == "disk"` **at the runtime call** and the rendered panel names the ceiling; auto + big host → RAM control; forced ram + blown budget → refused **before any shard was written** (asserted on the cache dir), ceiling in the message; forced ram + big host → proceeds.

```
M1 choose_tier ignores the ceiling            4 failed by name
M2 fallback note never names the ceiling      1 failed by name
M3 predicate boundary flipped (>= -> >)       1 failed by name
M4 forced-ram validator clause removed        1 failed by name
baseline                                      14 passed (unit set)
```

The wiring mutations (deleting `total_ram_bytes=total_ram` at the `build_stream_plan` call, or `total_ram=total_ram` at the validator call) are exactly what the four wiring tests exist to kill; they cannot be demonstrated on this box — see the honest gap below.

## The honest gap, per the claim

This is a Windows box: no `/dev/shm`, no `RLIMIT_MEMLOCK`, no 30 GB Linux host, and — separately — a locally broken transformers import chain (`torch.utils._pytree.register_constant`, torch 2.6 + transformers 5.x dev) that makes **every** test driving the real pre-flight fail here, including `test_v07203.py`'s own CI-green harness tests. Parity verified by stash: identical failure sets on main. So: the 14 unit tests, all four mutation runs, and every existing tier/plan suite (`test_v07200` 29 passed; the #602 validator pins 2 passed) are local evidence; the 4 wiring tests are CI-arbitrated, on the same harness CI already proves. What no test here can do is reproduce the live exit-137 kill — that verification belongs to someone with the reporter's environment, and the issue's own forced-disk run (exit 0) is the behavioural evidence that the fallback target works.

## Verification

- `tests/test_v07200.py -k "tier or plan or pinning or pageable or panel"`: 29 passed — the existing tier/plan contract untouched.
- `tests/test_issue602_qwen4_ple_streaming.py -k "ram_fit or ngram_source or refusal or resolve"`: 2 passed — the validator's new kwarg is additive; existing callers unchanged.
- `test_v07203.py -k "nvme or virtio or probe"`: 3 failed / 9 passed — **identical on stashed main** (harness + doctor env failures, named in the gap above).
- `ruff check` clean on all three touched Python files.
- Zero existing tests deleted or weakened; the only production-contract changes are additive kwargs with backward-compatible defaults.


MakazhanAlpamys:
**Stop before you invest more in this, please — there is a collision and it is not your fault, but you should know about it now rather than after a review round.**

**#644 has been open against #622 since 09:58 today**, by @ousamabenyounes. You claimed the issue at 11:38 and opened this at 12:22, and this PR does not mention #644, so I assume you did not see it. Nothing in your claim comment was done in bad faith — the issue carried no assignee, and that is a gap in how I run the tracker, not a mistake of yours.

The awkward part is timing that runs against you. I reviewed #644 at 11:37 — an hour before this PR — and asked its author, in writing:

> Do you want to add the `os.statvfs("/dev/shm")` read now, or leave it? Entirely your call — if you would rather ship this and stop, I will file the tmpfs read as a follow-up and credit you for the groundwork.

So I put exactly the scope this PR implements to another contributor and told them the decision was theirs. I am not going to take that back an hour later because a second implementation arrived. That would teach every contributor here that answering my questions is slower than being outrun, which is the opposite of what I want.

**What I am doing:** holding this PR until @ousamabenyounes answers. Two outcomes, and I will say plainly which is more likely:

- **If they take the tmpfs work**, this PR closes unmerged. In that case I will still credit your analysis in `CONTRIBUTORS.md` if it moved the outcome — and from a first read it may well have, see below.
- **If they decline it or do not answer within a couple of days**, this becomes the PR for the tmpfs ceiling and I review it on its own merits like anything else.

**Now the part that is not procedural.** Your claim comment promised to *"read the allocation path first and let the code settle the disputed mechanism"*, and if you did that and reached a conclusion, it matters regardless of which PR merges. I raised the same uncertainty on #644 and explicitly refused to claim it as settled, because I had not verified it: `RamSource` allocates through `torch.empty(..., device="cpu", pin_memory=True)`, which is `cudaHostAlloc` and should be anonymous page-locked memory — yet the reporter's kernel log shows `shmem-rss: 11653136kB` against `anon-rss` under 300 MB, which points the other way. One of those two readings is wrong and the answer decides whether a `/dev/shm` ceiling is the real bound or a correlation that happens to fit one box.

**If you have that answer, post it on #622 rather than leaving it inside this PR.** It is the load-bearing question for both implementations, it is the thing @here-2007 raised and nobody has closed, and it is worth more than either patch. I will credit it as the finding it is.

I am sorry to hand you this after you did the work. The fix on my side is to start assigning issues on claim so the next person sees it before opening a branch — I will do that from today.

umran666:
Understood holding here, no further investment until @ousamabenyounes's answer and your call.

Two things done instead of code:

- The mechanism analysis you asked for is posted on #622 rather than sitting in this PR: code-level verification that the /dev/shm mount is not the allocation channel, the telemetry both readings must swallow (shmem-rss vs anon-rss), the consequence for the open statvfs question on #644, and a ten-line falsifiable experiment for a Linux+CUDA box  labeled by confidence, since I cannot run that experiment here.
- Read #644 once you named the collision: it converged on the same dual-threshold design, and it carries the ceiling through the Qwen4 PLE auto-selection path that this PR does not. I have said so on #622 rather than leaving a reader to discover it. If #644 ships, this PR's patch is redundant and the analysis is the salvageable part  I will not argue otherwise.

MakazhanAlpamys:
Update, with the date I owed you.

@ousamabenyounes has not answered in ~20 hours. I have now given them until roughly **2026-09-04 12:00 UTC**, and said publicly on #644 that after it I decide on the merits and say which I picked and why.

That should have been in my first comment here. Holding your PR with no end date is not a neutral act — it costs you the option of doing something else with the time — and "until someone answers" is not a policy. My mistake, and the clock fixes it going forward.

Your position is unchanged and so is my reasoning for the order: #644 was open first, and I had put a question to its author about exactly this scope an hour before this PR existed.

One thing that is not affected by any of it: if you settled the `cudaHostAlloc`-versus-shmem mechanism, **post it on #622 regardless of what happens to either PR**. That question decides which ceiling is the real bound, nobody has closed it, and it outlives both patches. I will credit it as a finding in its own right.

MakazhanAlpamys:
**The clock has been answered, and your analysis is what answered it.**

@ousamabenyounes replied on #644 within the window and **declined the statvfs read for your reasons, citing the mechanism question as settled**. So the outcome is the one you predicted in your own comment: #644 proceeds, and this PR's patch is redundant. I am reviewing #644 at its new head now; if it passes, I close this unmerged.

I want to be precise about what that does and does not mean, because "closed unmerged" is a bad summary of what happened here.

**Your analysis on #622 is the load-bearing contribution to the fix that ships.** It settled the question both PRs were blocked on, and it did so at the right level of confidence throughout — verified-in-code separated from inferred-from-telemetry separated from an experiment that would falsify your own inference. Three specific things I am taking from it:

1. **`/dev/shm` is not the allocation channel**, and it is greppable rather than arguable: the store is `torch.empty(..., pin_memory=True)`, i.e. `cudaHostAlloc`, with no `shm_open`, no `multiprocessing.shared_memory`, no tmpfs-backed file anywhere in the streaming path. That is the fact I could not establish and would not claim.
2. **You reconciled the telemetry instead of dismissing it.** `shmem-rss` 11.65-13.99 GB against `anon-rss` under 300 MB is exactly the observation that made me refuse to call the mechanism settled in either direction. Driver-internal shmem objects not bound to any mount explains the accounting *and* explains why `df /dev/shm` stays at 0% *and* explains the OOM kill rather than a graceful failure. You labelled the attribution as inference rather than presenting it as established, which is why I trust the rest of it.
3. **The `RLIMIT_MEMLOCK` argument is the one I had not thought through**: a `getrlimit` gate reads systemd's 8 MB default and would force every Unix box to a pageable store, contradicting this repo's own measured multi-GB page-locked runs. That would have been a plausible-looking fix that broke the feature.

**And you named a gap in the PR that beat yours.** Qwen4 PLE auto selection sizing a RAM PLE against free-RAM headroom alone was in #644 and absent from #646, you said so plainly rather than keeping it as leverage, and `a82a947` exists because of it. Pointing out where the competing implementation is *better* than yours, and then pointing out where it is *worse* so it can be fixed, is the behaviour I would most like other contributors to copy.

**So:** when #644 merges you get a `CONTRIBUTORS.md` entry for this work, and the merge commit will say the analysis and the gap are yours. Not a consolation line — the record of what actually determined the shape of the fix.

Two loose ends I am carrying rather than dropping:

- **Your falsifiable experiment is worth running**, and nobody in this thread has a Linux + CUDA box. I will file it as its own issue so the inference is either confirmed or overturned on hardware instead of resting on a converging argument. If it ever comes back the other way, the statvfs question reopens with evidence and I would rather that be discoverable than buried in a closed PR.
- **The "mechanism-agnostic bound" convergence** — two independent implementations reaching store + resident against a fraction of total RAM — is the strongest thing either PR has, and I am recording it in the merge as evidence rather than as a coincidence.

Nothing more is needed from you here. Thank you for holding without pushing and for writing the analysis where it would count for both PRs rather than only your own.

MakazhanAlpamys:
**#644 is merged** (`171e914`), so closing this as the redundant patch — exactly as you predicted in your own comment, and with the outcome you argued for: the `statvfs` read was declined for your reasons.

**Your analysis is credited in `CONTRIBUTORS.md`** (`44533dc`, on @ousamabenyounes's entry, which records that the bound shipped is mechanism-agnostic *because* the mount is not the allocation channel) and named in the merge commit, which says in as many words that you settled the mechanism question and named the Qwen4 PLE gap that was in #644 and absent from your own PR.

I want to be concrete about what carried, so this does not read as a courtesy:

- **The `statvfs` rejection is now the documented design**, with your two counterexamples — false positive on a large host with a small custom mount, false negative on a `--shm-size=64G` container — written into the code comment beside the constant.
- **The `RLIMIT_MEMLOCK` argument** kept a plausible-looking fix out. `getrlimit` reads systemd's 8 MB default and would have forced every Unix box to a pageable store, contradicting this repo's own measured multi-GB page-locked runs. That one I had not thought through.
- **The Qwen4 PLE gap** became `a82a947`. I verified it: `_resolve_qwen4_ngram_source` now conjoins both gates at its return, and `total_ram` is threaded from production rather than defaulted.
- **The convergence argument** — two independent implementations reaching store + resident against a fraction of total RAM — is in the merge commit as evidence rather than as coincidence.

**#655 is filed** for your falsifiable experiment, crediting you for the analysis and the design, so the inference is either confirmed or overturned on hardware instead of resting on a converging argument. If it ever comes back the other way, the `statvfs` question reopens with data — and that is findable in an open issue rather than buried here.

Thank you for holding without pushing, for writing the analysis on #622 where it would count for both PRs rather than only your own, and for saying up front that you would call your own patch redundant rather than defend it. That last part is why the collision cost this project almost nothing.

---

*Edited: two code identifiers were dropped from the Qwen4 bullet when I posted this — my shell ate them as command substitutions. Restored; no other change.*

PR #647, opened 2026-09-02
test(stream): pin stream_pin:false down to the store allocation (#623)

umran666:
Part of #623 — the open half, closed on evidence rather than argument (claimed with the measurement in [this comment](https://github.com/MakazhanAlpamys/Soup/issues/623#issuecomment-5512340006)).

## The question

The maintainer's status note kept this half open: *"on a build that does have `stream_pin`, setting it `false` is accepted and the store still lands in shared memory."* The thread disagreed — @here-2007 traced the wiring as complete. So: measured instead of argued, on `main` at `2285b65`, real CUDA device (RTX 3050 4 GB, Windows, torch 2.6.0+cu124, transformers 5.16.1), the **real** `_setup_streaming_transformers` with **no stubs**, over a tiny 3-layer checkpoint:

| config | `src.pinned` | store tensors `is_pinned()` |
|---|---|---|
| `stream_pin: false` | `False` | **0/27 — pageable store** |
| default (unset) | `True` | **27/27 — pinned store** |

**The key is honoured on main.** The second row is what makes the first non-vacuous: page-locking demonstrably happens through this exact chain on this box, and `false` demonstrably prevents it.

The reporter's telemetry needs no wiring bug: their wheel was **0.73.3, which predates the field** (@Srinivasan8888's tree analysis — the key was dropped as an unknown config key, the default pinned path ran), and per the #622 analysis the kernel accounts driver-pinned host pages under `shmem-rss` (their dumps: 11–14 GB `shmem-rss` against <300 MB `anon-rss` **while the store was pinned as requested**). "The store lands in shared memory" is what a *correctly pinned* run looks like in `journalctl` on that stack.

## Why this is tests-only, and one correction to the issue's suggested test

No production change: every bullet of the suggested fix is already satisfied on main — the key is honoured (table above), a forced pin that cannot be applied refuses loudly (`decide_pinning`/`RamSource`, pinned by `TestRequirePinSurvivesEveryHop`), and the pre-flight prints the resolved state (asserted below). What was missing is exactly what the issue asked for last: *"a regression test that ... would pin this down."*

The issue suggested asserting the store is "anonymous (not `shmem`)". That discriminator is wrong on the reporter's own stack — it would condemn a correctly pinned run, per the shmem accounting above. `is_pinned()` — whether the allocation went through `cudaHostAlloc` or the pageable heap — distinguishes "ignored" from "honoured" on every OS, and is what these tests assert.

## The two layers

A full real chain on a *simulated* CUDA device cannot allocate the GPU buffer pool, so the commitment is split:

1. **`TestTheWiringReachesTheRuntimeBoundary` — runs on every CI cell.** The real pre-flight with a simulated-CUDA device (`device` is a string check; `mem_get_info` stubbed — the `test_v07203.py` pattern) and `build_streamed_model` **spied**: asserts the `pin=` argument the runtime boundary receives is `False` for `stream_pin: false` and `True` for the default, plus the resolved panel state (`pageable` / `(pinned)`). A mutant hardcoding the pin at the boundary dies here, on CI, in either direction.
2. **`TestTheAllocationItself` — any box with a real CUDA device**, `@requires_cuda` with a stated reason (never a silent skip; the #596 lesson). No stubs: real `RamSource`, `is_pinned()` on **every** store tensor, both directions **in one test** so "false is pageable" can never pass on a box where pinning silently never happens. The torch-level discriminator is sanity-checked before being trusted with the chain.

Hops below the boundary (`build_streamed_model → install_streaming → _build_source → RamSource`) are already driven by the real CPU chain in `TestRequirePinSurvivesEveryHop`, and the pre-flight cost note by `TestAutoTierFallback` — referenced, not duplicated.

All three at the boundary expression `pin=plan.pinned and on_cuda` (`stream_setup.py:791`), each **killed by name in both layers**:

```
M1 pin=on_cuda (plan ignored)     2 failed: boundary-false + allocation
M2 pin=False hardcoded            2 failed: boundary-default + allocation
M3 pin=True hardcoded             2 failed: boundary-false + allocation
baseline                          3 passed (local box, CUDA present)
```

`decide_pinning`'s `stream_pin=False` branch itself is already pinned by the existing `test_v07200.py` plan tests — mutating it dies there before reaching these.

## The honest gap

Linux kernel accounting (`/proc`, `shmem-rss`, the reporter's exact box) is not re-measured here — Windows box. The allocation-API evidence covers the mechanism (pageable heap vs `cudaHostAlloc` is what determines the kernel's accounting class), and the layer-2 test runs for anyone with a CUDA box, including the reporter's. If the maintainer holds a measurement of the `false` path landing page-locked **on a build that has the field**, this PR is wrong and I want to see it — the table above is the opposite, on main, today.

## Verification

- `tests/test_issue623_stream_pin_allocation.py`: **3 passed** on this box (CUDA layer executed, not skipped), 38 s.
- `test_v07203.py::TestAutoTierFallback` (the harness pattern these borrow): 16 passed on the same env.
- `ruff check` clean.
- No production code touched; no changelog fragment — a test-only change with no user-visible impact.


MakazhanAlpamys (approved):
**Approving. And the first thing to say is that my status note on #623 was wrong, and this PR is what showed it.**

I checked the claim you and @here-2007 disputed, at the tag the reporter was actually running:

```
git show v0.73.3:src/soup_cli/config/schema.py | grep -c stream_pin     -> 0
git show v0.73.3:src/soup_cli/config/schema.py | grep -c stream_layers  -> 18   (control)
```

The field **did not exist** in the release the reporter had installed. Pydantic's `extra="ignore"` dropped the key, the default pinned path ran, and the OOM they saw is what a *correctly pinned* run looks like. So "accepted and still lands in shared memory" was describing a build with no such setting to accept — I carried that note forward across several updates without re-checking it against the tag, which is the actual mistake, and it is the kind that survives precisely because it sounds like a known issue.

**The chain is honoured end to end on `main` today.** I followed it myself rather than taking the table: `schema.py` → `stream_setup.py:698` → `decide_pinning` at `layer_stream.py:815` → `StreamPlan.pinned` → `stream_setup.py:791` → `install_streaming` → `_build_source` at `layer_stream_runtime.py:2212` → `RamSource.pinned` → `torch.empty(..., pin_memory=self.pinned)` at `:331`. One `pin_memory=` in the whole streaming path, three `RamSource(` constructions all inside `_build_source`, so no second host store can leak past the decision.

## Why a tests-only PR that changes nothing is still worth merging

Because the conclusion "it works" is worthless if nothing keeps it working. I ran the deepest mutation — the one that *records* the flag and then ignores it at the allocation:

```
pin_memory=self.pinned  ->  pin_memory=True

tests/test_issue623_stream_pin_allocation.py:  1 failed, 2 passed
    FAILED TestTheAllocationItself::test_false_allocates_pageable_and_default_pins
tests/test_v07203.py:                          169 passed, 5 skipped
```

**169 pre-existing streaming tests pass with that mutation in place.** Your file is the only thing in the repo that catches it — and it is exactly the shape #623 alleges: the flag accepted at every layer that has a test, and lost at the one that does not. Asserting `is_pinned()` on the real allocation rather than the recorded `.pinned` attribute is what makes that possible; a test on the attribute would have passed the mutation happily.

CUDA-gated and it **ran** here (RTX 3050, torch 2.5.1+cu121, so a different stack from your 2.6.0+cu124 with the same result — the finding is not stack-specific). Under `FORCE_COLOR` it stays green, because the panel assertions read a `StringIO` Console rather than CliRunner output. `SOUP_LAYER_STREAM_CACHE_DIR` redirected per direction, `_close_stream_runtime()` in a `finally`, and the non-empty-store guard (`assert flags, "the store is empty — nothing was measured"`) is the detail I look for and rarely see: without it the CUDA test could pass while measuring nothing.

Zero deleted test lines. No changelog fragment, correctly — `CONTRIBUTING.md` scopes those to user-visible changes and you said so in the body rather than leaving me to work it out.

## One follow-up I will file rather than ask of you

The two deepest hops are **CUDA-only**, so they survive on all nine CI cells. A CPU-only assertion that `_build_source(..., pin=False, tier="ram")` yields a `RamSource` with `.pinned is False` would kill them everywhere — `torch.empty(pin_memory=False)` needs no GPU. Your PR body already scopes the coverage honestly and does not overclaim, so this is an addition, not a correction.

## Housekeeping that is mine

`pytorch-smoke` became a required check after your run, so this showed `BLOCKED` at 13/13 green. I have updated the branch; it is running now. Nothing for you to do.

Merging once it is green.

MakazhanAlpamys:
Merged as `fc5f144`, credited in CONTRIBUTORS at `2137035`.

Credited for the awkward version of the outcome: you were asked to settle a dispute and the answer was **that my status note was wrong**, which is harder to deliver than a fix and easier to soften. The tag check is what settled it, and it is the kind of evidence that ends an argument rather than continuing it.

The reason this merged despite concluding "the flag already works" is the part I would want other contributors to copy. I ran your scenario at the deepest hop before merging — the flag *recorded* and then ignored at allocation:

```
pin_memory=self.pinned  ->  pin_memory=True

tests/test_issue623_stream_pin_allocation.py:  1 failed, 2 passed
tests/test_v07203.py:                          169 passed, 5 skipped
```

169 existing streaming tests are blind to it. Asserting `is_pinned()` on the real allocation instead of the recorded `.pinned` attribute is what makes the difference — an attribute assertion would have passed that mutation without noticing, and that is the exact failure #623 describes. So the issue's mechanism was reachable; only its reported route was not.

The CUDA test ran here (RTX 3050, torch 2.5.1+cu121 — a different stack from your 2.6.0+cu124, same result), and the non-empty-store guard is the detail I look for and rarely see: without `assert flags, "the store is empty — nothing was measured"` the CUDA test could pass while measuring nothing at all.

The known gap is recorded in the merge commit rather than left implicit: the two deepest hops are CUDA-only and survive on all nine CI cells. A CPU-only assertion that `_build_source(..., pin=False, tier="ram")` yields `.pinned is False` would close that without a GPU — `torch.empty(pin_memory=False)` needs no device. I will file it unless you want it.

`Refs #623`, not `Closes` — #623 stays open for that CPU-side coverage.

PR #648, opened 2026-09-02
fix: use resolved dtype for GEMM forecast

### LINKED ISSUE #617: GEMM-ceiling probe hardcodes bfloat16, so the throughput forecast is pessimistic on every pre-Ampere card
## Summary

`measure_gemm_tflops`'s probe allocates its matrices as `bfloat16` unconditionally, regardless of the dtype the run resolved. On a pre-Ampere card (compute capability < 8.0 — T4, P100, V100, GTX 16xx, Quadro T-series) there are no native bf16 tensor cores, so the probe measures **emulated** bf16 throughput while training will actually run in fp16. The GEMM ceiling, and therefore the throughput bracket printed in the layer-streaming pre-flight panel, is pessimistic on that entire GPU generation.

Reported by @ArtisanalAutomationsFactories as a secondary finding while root-causing #558 (which turned out to be a duplicate of #425 and unrelated to this). Confirmed.

```python
# src/soup_cli/utils/layer_stream_runtime.py:1289-1290
left  = torch.randn(size, size, device=device, dtype=torch.bfloat16)
right = torch.randn(size, size, device=device, dtype=torch.bfloat16)
```

Observed forecast on a Quadro T2000 (cc 7.5): **26-39 tok/s**, which the reporter expects to be an underestimate for cards in this class once the probe uses the dtype the run will actually use.

## Severity: cosmetic, and the boundary is worth stating

This is display-only. `measure_gemm_tflops` feeds `GemmCeiling` -> `forecast_stream_throughput`, which prints a bracket in the pre-flight panel. It is not on the training path, it cannot reach the optimizer, and it does not affect correctness, VRAM budgeting, or the fit refusal.

It still matters: the forecast is one of the few numbers an operator sees *before* committing to a run, and one that is systematically wrong on a whole GPU generation trains people to ignore the panel. This project has an issue open (#405) about a metric that cannot move; a metric that moves in the wrong direction on pre-Ampere is the same category of problem.

## Proposed fix

```python
dtype = _torch_dtype(resolve_stream_dtype(str(device)))
left  = torch.randn(size, size, device=device, dtype=dtype)
right = torch.randn(size, size, device=device, dtype=dtype)
```

`resolve_stream_dtype` already returns `"float16"` on a card whose `cuda_supports_bf16()` is false (the #385/#387 work, where the recorded trap is that `torch.cuda.is_bf16_supported()` defaults to `including_emulation=True` and answers **True** on a T4 — use the existing helper, do not re-derive this).

Note `GemmCeiling`'s docstring says "a dense **bf16** GEMM rate measured on THIS card". If the dtype becomes variable, the docstring and any prose that quotes it must say which dtype the number is in, or the ceiling stops being comparable between two runs on different cards.

## Acceptance criteria

- [ ] The probe allocates in the dtype the run resolved, via `resolve_stream_dtype`, not a literal
- [ ] `GemmCeiling` records **which** dtype it measured, and the panel says so — a bracket whose unit is unstated is worse than one that is consistently wrong
- [ ] A test that fails when the literal is restored: force `resolve_stream_dtype` to `"float16"` and assert the allocated dtype follows. This must not require a GPU — the allocation call is mockable
- [ ] A control asserting the bf16 path still allocates bf16 on a card that supports it, so a fix that hardcodes `float16` instead is not an improvement
- [ ] No change to the VRAM fit decision or to any published number in `benchmarks/` — this is display-only, and if a gate number moves, something else is wrong

## Notes

@ArtisanalAutomationsFactories, this is yours if you want it — you found it, and your proposed shape is what is written above. Say so on the issue and it is claimed; if you would rather not, it is a good first issue for someone else.

**What this issue does NOT cover:** whether the forecast is *accurate* on pre-Ampere once the dtype is right. Nobody here has a Turing card to measure it on, so the acceptance criteria are about the probe using the correct dtype, not about the resulting number being validated. Re-measuring the bracket on pre-Ampere hardware is a separate question and belongs with #325 / #395.

Samearth17:
## Summary

Fixes #617 by making the GEMM throughput probe use the same card-resolved stream dtype as the rest of layer streaming.

## Changes

* `measure_gemm_tflops()` now uses `resolve_stream_dtype()` instead of hard-coding `torch.bfloat16`.
* `GemmCeiling` records the dtype used for the measurement.
* The forecast output reports the measured GEMM dtype alongside TFLOPS and clock.
* Added CPU/mock coverage for both `float16` and `bfloat16` resolution paths.
* Updated the independent timing test to use the same resolved dtype.

## Validation

* `pytest -o addopts= -q tests/test_v07203.py -k "measuredgemmceiling or issue444bestrepeatselection or issue617"` → **6 passed, 4 skipped**
* `pytest -o addopts= -q tests/test_v07203.py` → **161 passed, 15 skipped, 16 warnings**
* `git diff --check` passes.

The change does not alter the throughput formula or VRAM-fit logic; it only makes the measured GEMM dtype match the dtype Soup resolves for the target card and surfaces that dtype in the forecast.


MakazhanAlpamys (changes requested):
**The engineering is right and it clears the trap I was most worried about. Blocking on four small items, one of which is a test.**

## It is not a no-op on the hardware it targets — I checked, because the last attempt at this class was

v0.73.1 fixed the same shape of bug in fourteen places (#385/#387), and the recorded trap is that `torch.cuda.is_bf16_supported()` defaults to `including_emulation=True`, so a **T4 answers True** and the first fix was a no-op on exactly the cards it was written for. So I simulated a T4's signature here rather than reasoning about the chain:

```
bare is_bf16_supported()        -> True     <- the trap
including_emulation=False       -> False    <- T4 truth
cuda_supports_bf16()            -> False
resolve_stream_dtype("cuda")    -> float16
```

`gpu.py:411` makes the non-default call explicitly, and `resolve_stream_dtype` delegates to it rather than re-deriving. **This PR genuinely helps a T4.** Delegating instead of writing a second capability check is the right call and it is why mutating the resolver to always answer `bfloat16` is caught by `tests/test_issue385_stream_dtype.py` rather than by anything here — correct division of labour, not missing coverage.

Mutations of the fix itself, all killed: reverting the probe to hardcoded `torch.bfloat16` and making `GemmCeiling` record a constant both die on `test_gemm_uses_resolved_stream_dtype[float16]`; hardcoding `torch.float16` instead — issue #617's fourth acceptance criterion — dies on three tests. Zero deleted or weakened tests: the four removed lines are in-place edits, and the CUDA independent-timing test now compares like-for-like against the resolved dtype, which is a strengthening. `test_cli_startup_is_light` stays green; the new import is inside `measure_gemm_tflops`, not at module scope.

## Blocker 1 — the user-visible half of the fix has no test

Issue #617's second acceptance criterion is that `GemmCeiling` records which dtype it measured **and the panel says so**. The recording half is well covered. The panel half is not:

```
# deleted `using {ceiling.dtype}` from stream_setup.py:1035
tests/test_v07203.py:  171 passed, 5 skipped
```

So the line a user actually reads can be removed with the suite green. `grep -rn "measured on this card" tests/` finds nothing. One assertion on that panel string closes it, and it is the assertion that makes the criterion real rather than described.

## Blocker 2 — no changelog fragment

`changelog.d/0.73.3/648.fixed.md` does not exist; the PR touches three files and none of them is a fragment. This is user-visible output, so it needs one, with the credit line in the neighbours' style: `(#617 by @Samearth17 in #648)`.

## Blocker 3 — the docs now say something false, and the issue asked for this specifically

#617 said the docstring *and any prose that quotes it* must say which dtype the number is in. The docstring is fixed; `docs/performance-and-quantization.md` is not, in two places:

- **line ~418** — the sample panel reads `(from 6.75 TFLOPS measured on this card now @ 862 MHz)`, which no longer matches what the code prints.
- **line ~432** — *"It comes from a **bf16 GEMM** benchmarked on your card in that session"* is now false on pre-Ampere, which is the exact hardware this PR exists for. A doc that is wrong specifically for the cards a fix targets is worse than one that is uniformly stale.

`benchmarks/` correctly untouched — no published number moves, which is right.

## Blocker 4 — `pytorch-smoke` never ran, and that one is my doing

Your run is 13/13 green but predates a change of mine: `pytorch-smoke` landed on `main` in #601 and I made it a required check afterwards, so every PR whose CI predates that shows `BLOCKED` with everything passing. Nothing wrong with your branch. It resolves on your next push — which you will be making anyway for the three items above — or I can update the branch for you if you would rather not.

## One optional consistency note

`layer_stream_runtime.py:80` has `_torch_dtype()`, which validates against `_DTYPE_NAMES` and is used at seven other sites in that same file; the fix uses a bare `getattr(torch, dtype_name)`. Functionally identical because `SUPPORTED_STREAM_DTYPES` is a subset, so this is house style rather than correctness — take it or leave it.

Also noted and not a problem: `GemmCeiling` gains a required `dtype` field with no default, which is technically breaking for an external constructor. Two construction sites repo-wide, not exported in `__all__`, and it is a result object callers receive rather than build. Fine as is.

The measurement work here is good and the T4 simulation above is the evidence I would have wanted in the PR body. Add the test, the fragment and the two doc lines and this merges.

MakazhanAlpamys:
Merged as `f14bf42`. Thank you, and sorry this took two rounds and three days of my silence between them — that delay was mine, not yours.

## All four items from my 2026-09-03 review are resolved

I verified each rather than reading the commit messages.

**1. The panel test.** This was the one that mattered: a mutation deleting `using {ceiling.dtype}` from the user-visible string survived your first version. Re-run on `817456c`:

```
MUTATION: f"this card now using {ceiling.dtype}{clock})" -> f"this card now{clock})"
  -> 2 failed  (TestIssue617PanelDtype[float16], [bfloat16])
```

And it dies on a CPU-only run too, which is the part that makes it worth having — every one of the nine CI cells is CPU-only, so a GPU-gated test would have been invisible there.

**2. The probe itself**, re-run as the control:

```
MUTATION: probe reverted to hard-coded torch.bfloat16   (= #617 restored)
  -> 1 failed  (test_gemm_uses_resolved_stream_dtype[float16])
```

**3. The two false doc lines.** `docs/performance-and-quantization.md:418` is now byte-identical in format to what the code actually prints, and the "It comes from a bf16 GEMM" sentence names both branches. A repo-wide sweep finds no stale prose left in `docs/`, `README.md` or `src/`; the remaining hits are in `benchmarks/`, which are historical transcripts and correctly untouched.

**4. `pytorch-smoke`** became a required check while this was open and has now run: 14 of 14 green. Your `817456c` also moved the fragment from `changelog.d/0.73.3/` to `0.74.0/` — v0.74.0 shipped between my review and your push, so that was release timing on my side.

## The part of this I want on the record

**`torch.cuda.is_bf16_supported()` returns True on a T4, through emulation.** That is why this bug survived: the obvious check agrees with the wrong answer, so anyone verifying "does this card support bf16" gets a yes and stops. Routing through `resolve_stream_dtype` — the same function the streaming path itself uses — is the fix that cannot drift, because now the probe and the run cannot disagree by construction.

**You drove the pre-Ampere direction end to end without a pre-Ampere card**, and did it the honest way: faked exactly one thing, `is_bf16_supported(including_emulation=False)` returning False, which is a T4's actual truth, then ran the real `soup train --config soup.yaml --yes` with `stream_layers: true` and showed the panel printing `measured on this card now using float16`. One fake, at the hardware boundary, with everything downstream real. That is the right shape for a fix nobody here has the hardware to test.

Four removed test lines, all audited: in-place edits, and the CUDA independent-timing test now compares like for like against the resolved dtype, which is a strengthening. No new failures — 34 pre-existing #651 failures on both trees, sorted lists identical.

## Two follow-ups, neither yours to fix

**A limit of the tests, inherent to CI having no GPU.** The torch stub sets `torch.bfloat16 = "bfloat16"`, so the assertions cannot distinguish a real `torch.dtype` lookup from the raw name being passed through. A mutation doing exactly that is caught with real CUDA and invisible on all nine cells. Recorded in the merge commit rather than held against the PR.

**`benchmarks/run-t4-colab-free-tier.md:64` is now stale in substance.** It carries a real T4 record — `forecast 31-46 tok/s (from 2.20 TFLOPS measured on this card @ 1185 MHz)` — measured under the emulated-bf16 probe your fix removes. #617 says nobody has a Turing card, and it turns out there is one published pre-Ampere number, and this change invalidates it as a forecast baseline. It is a historical transcript and correctly untouched; I will file the re-measurement separately.

Credited in `CONTRIBUTORS.md` (`601fbc3`). You are also assigned #651, which is the other half of the floor problem — no rush, and say the word if you would rather hand it back.

PR #652, opened 2026-09-03
fix(monitoring): use max_memory_allocated to report peak GPU memory in live training panel

### LINKED ISSUE #650: The live training panel reports `memory_allocated()` sampled between steps, so "GPU: 5.8/15.9 GB" hides the real peak
## What

`monitoring/callback.py:177` (`on_log`; line 163 on v0.73.3) builds the GPU cell of the training
panel from `torch.cuda.memory_allocated()`. `on_log` fires between steps, after
activations and gradients have been freed, so the number is the trough, not the
peak. On a run whose step peaks at ~15 GB of a 15.9 GB card the panel reads
`5.8/15.9 GB`, which tells the operator there is headroom that does not exist.

The same file already uses the right counter 400 lines lower: the grad-accum
advisor at `callback.py:566` reads `torch.cuda.max_memory_allocated()`. The two
readouts disagree on the same run.

## Observed

RTX 5080 16 GB · WSL2 · soup-cli 0.73.3 · torch 2.14.0+cu130 · Qwen2.5-1.5B-Instruct, LoRA r=16, batch 16, fp32 base (0.73.3)

Panel during training: `GPU: 5.8/15.9 GB` (5.8 GB is exactly the fp32 weights at rest: 1.54 B × 4 B)
`nvidia-smi` at the same moment: 15,893 MiB used, 100% util
Windows perf counter for the WSL VM process: 14.98 GB dedicated + 21.46 GB shared

## Severity

Display-only. Nothing decides on this number. It still matters for the same
reason #617 does: it is the one memory figure the operator watches during a
run, and a figure that is systematically low trains people to raise batch size
into an OOM (or, under WDDM, into a silent spill, see #649).

## Suggested fix

Report `max_memory_allocated()` and `max_memory_reserved()` (peak allocated /
peak reserved) in the cell, or both peak and current with labels. If the peak
is shown per logging window, call `reset_peak_memory_stats()` after reading;
if cumulative, say so in the label. Either is fine; an unlabeled single number
is the problem.

## Acceptance criteria

- [ ] The panel's GPU figure is a peak (`max_memory_allocated` or
      `max_memory_reserved`), and the label says which.
- [ ] A test with a mocked `torch.cuda` where `memory_allocated` returns a low
      value and `max_memory_allocated` a high one asserts the panel shows the
      high one (`tests/test_callback.py` is already mock-based).
- [ ] No change to any decision path; the grad-accum advisor keeps reading its
      own counter.

webdevsamran:
## Summary

Fixes #650.

on_log in \src/soup_cli/monitoring/callback.py\ previously read \	orch.cuda.memory_allocated()\ to build the GPU cell of the training display panel. Because \on_log\ fires between steps (after activations and gradients have been freed), the figure showed the trough allocation rather than actual peak memory consumption.

## Changes

1. **Peak allocation counter**: Replaced \	orch.cuda.memory_allocated()\ with \	orch.cuda.max_memory_allocated()\ in \on_log\. We deliberately do not call \
eset_peak_memory_stats()\ here because that reset is process-global and would clobber the grad-accum advisor's own peak reading at \callback.py:566\ (#650 acceptance criterion 3).
2. **Display label**: Changed the panel readout in \src/soup_cli/monitoring/display.py\ from \GPU:\ to \GPU peak:\ so the metric is self-describing.
3. **Tests**:
   - Added unit test \	est_on_log_gpu_memory_uses_max_allocated\ in \	ests/test_callback.py\ asserting that peak allocated VRAM is reported over trough allocation.
   - Updated \	ests/test_display.py\ (\	est_display_render_panel\) asserting that the panel renders the \GPU peak:\ label.
4. **Changelog**: Added fragment \changelog.d/0.73.3/652.fixed.md\ crediting \(#650 by @YuriPerro in #652)\.

## Validation

- Verified all 16 tests in \	ests/test_callback.py\ and \	ests/test_display.py\ pass.
- Verified linting and formatting with \
uff check\ and \
uff format\.

### DIFF
--- a/src/soup_cli/monitoring/callback.py
+++ b/src/soup_cli/monitoring/callback.py
@@ -174,7 +174,7 @@ class _SoupTrainerCallback_body:  # noqa: N801
             import torch
 
             if torch.cuda.is_available():
-                used = torch.cuda.memory_allocated() / (1024**3)
+                used = torch.cuda.max_memory_allocated() / (1024**3)
                 total = torch.cuda.get_device_properties(0).total_memory / (1024**3)
                 gpu_mem = f"{used:.1f}/{total:.1f} GB"
         except Exception:
--- a/src/soup_cli/monitoring/callback.py
+++ b/src/soup_cli/monitoring/callback.py
@@ -168,7 +168,11 @@ class _SoupTrainerCallback_body:  # noqa: N801
         if logs is None:
             return
 
-        # Try to get GPU memory
+        # Try to get GPU memory.
+        # Uses max_memory_allocated() to report peak allocated VRAM rather than the
+        # inter-step trough (#650). We deliberately do not call reset_peak_memory_stats()
+        # here because that reset is process-global and would clobber the grad-accum
+        # advisor's own peak reading at line 566 (#650 criterion 3).
         gpu_mem = ""
         try:
             import torch
--- a/changelog.d/0.73.3/652.fixed.md
+++ b/changelog.d/0.73.3/652.fixed.md
@@ -1,0 +1,8 @@
+- **Live training panel now labels the GPU figure as `GPU peak:` and reads `max_memory_allocated()` instead of `memory_allocated()` (#650 by @YuriPerro in #652).**
+  `on_log` fired between steps, after activations and gradients were freed, so the
+  sampled `memory_allocated()` was the inter-step trough -- on the reproducing run
+  `5.8/15.9 GB` while `nvidia-smi` showed the device fully occupied.
+  The fix uses `max_memory_allocated()`, the lifetime peak, and changes the label
+  to `GPU peak:` so the figure is self-describing. `reset_peak_memory_stats()` is
+  intentionally not called here: that reset is process-global and would clobber the
+  grad-accum advisor's own peak reading at `callback.py:566`.
--- a/src/soup_cli/monitoring/display.py
+++ b/src/soup_cli/monitoring/display.py
@@ -115,7 +115,7 @@ class TrainingDisplay:
         if self.speed > 0:
             lines.append(f"Speed: {self.speed:.2f} it/s")
         if self.gpu_mem:
-            lines.append(f"GPU:   {self.gpu_mem}")
+            lines.append(f"GPU peak: {self.gpu_mem}")
         if self.grad_norm > 0:
             lines.append(f"Grad:  {self.grad_norm:.4f}")
 
--- a/tests/test_callback.py
+++ b/tests/test_callback.py
@@ -174,4 +174,3 @@ def test_on_log_gpu_memory_uses_max_allocated():
     mock_torch.cuda.max_memory_allocated.assert_called_once()
     display.update.assert_called_once()
     assert display.update.call_args.kwargs["gpu_mem"] == "8.0/16.0 GB"
-
--- a/tests/test_callback.py
+++ b/tests/test_callback.py
@@ -152,3 +152,26 @@ def test_on_log_without_tracker():
     callback.on_log(_make_args(), _make_state(), MagicMock(), logs=logs)
 
     display.update.assert_called_once()
+
+
+def test_on_log_gpu_memory_uses_max_allocated():
+    """on_log should report peak VRAM via max_memory_allocated rather than current allocation."""
+    display = MagicMock()
+    callback = SoupTrainerCallback(display=display)
+    state = _make_state()
+
+    mock_torch = MagicMock()
+    mock_torch.cuda.is_available.return_value = True
+    mock_torch.cuda.max_memory_allocated.return_value = 8 * (1024**3)
+    mock_torch.cuda.memory_allocated.return_value = 2 * (1024**3)
+    props = MagicMock()
+    props.total_memory = 16 * (1024**3)
+    mock_torch.cuda.get_device_properties.return_value = props
+
+    with patch.dict("sys.modules", {"torch": mock_torch}):
+        callback.on_log(_make_args(), state, MagicMock(), logs={"loss": 1.0})
+
+    mock_torch.cuda.max_memory_allocated.assert_called_once()
+    display.update.assert_called_once()
+    assert display.update.call_args.kwargs["gpu_mem"] == "8.0/16.0 GB"
+
--- a/tests/test_display.py
+++ b/tests/test_display.py
@@ -59,6 +59,8 @@ def test_display_render_panel():
     rendered = _render_to_str(panel)
     assert "62/100" in rendered
     assert "0.847" in rendered
+    # The panel label must identify the figure as a peak, not just "GPU:" (#650).
+    assert "GPU peak:" in rendered
 
 
 def test_display_render_zero_steps():


MakazhanAlpamys (changes requested):
Welcome, and thank you — this is a clean first contribution and the test is a real one. **Two things to raise: one about timing that is not your fault, and one about the fix being two-thirds done.**

## The timing, so you hear it from me rather than work it out

I offered #650 to its reporter, @YuriPerro, **33 minutes before you opened this** — they filed it last night with the reproduction and I asked whether they wanted to take it. They have not answered yet. The issue carried no assignee, so there was nothing for you to see; that is a gap in how I run the tracker, not a mistake of yours, and it is the second time this week it has bitten someone.

**What I am doing:** giving @YuriPerro until roughly 2026-09-04 08:00 UTC to answer, then this proceeds on its merits either way. I am not holding it open-ended — I did that on another PR and it is still waiting on an answer two days later, which is unfair to the person who showed up ready to work. If @YuriPerro takes it, I will say so plainly here and credit your analysis where it moved the outcome.

Either way, do the review below — it applies to whichever PR lands, and none of it is wasted.

## The fix is right, and I verified the test discriminates

```
tests/test_callback.py:  9 passed

# reverted max_memory_allocated -> memory_allocated
FAILED tests/test_callback.py::test_on_log_gpu_memory_uses_max_allocated
1 failed, 8 passed
```

Setting `memory_allocated` to 2 GB and `max_memory_allocated` to 8 GB and asserting the panel shows `8.0` is the right shape — it fails for the specific reason it exists, not incidentally. That is better than most first PRs manage.

## What is missing — and the first item is the one #650 was actually about

**1. The label.** I wrote on the issue that the label is the load-bearing part, more than which counter you pick: *"an unlabelled number is what caused this."* The panel still renders `8.0/16.0 GB` with nothing saying it is a peak. An operator comparing that against `nvidia-smi` now sees a *different* unlabelled number and has no way to know which quantity either one is. The bug was never "the number is too small" — it was "the number does not say what it is". Something like `GPU peak: 8.0/16.0 GB` closes it.

**2. `max_memory_allocated()` is cumulative, and that is a design choice worth stating.** It reports the peak since process start (or since `reset_peak_memory_stats()`), so after the first big step the figure **stops moving for the rest of the run**. That is defensible — it is the true high-water mark, which is what you want for "am I near OOM" — but it is not the same as "peak this window", and an operator watching a frozen number will eventually wonder if the panel has hung.

If you go the per-window route instead, there is a trap I flagged on the issue and want to repeat because it is easy to hit: **`reset_peak_memory_stats()` is process-global**, so calling it in the callback would clobber the grad-accum advisor's own reading at `callback.py:566`. #650's third acceptance criterion is that no decision path changes; that is the way to break it accidentally.

Either choice is fine by me. Pick one and say which in the label.

**3. No changelog fragment.** #650 lists one as an acceptance criterion. `changelog.d/0.73.3/652.fixed.md`, with the credit line in the neighbours' style: `(#650 by @YuriPerro in #652)` — the issue number and its reporter, then your PR.

## One note on CI

Your run needed manual approval because this is your first PR here; I have approved it and it is running. That is normal for first-time contributors and not a comment on the change.

Nothing here is large. The label is the one I care about, because it is the actual subject of the issue.

src/soup_cli/monitoring/callback.py
```diff
@@ -174,7 +174,7 @@ def on_log(
             import torch
 
             if torch.cuda.is_available():
-                used = torch.cuda.memory_allocated() / (1024**3)
+                used = torch.cuda.max_memory_allocated() / (1024**3)
```
AmirF194, line 181: Checked this against `4de9d8e1` (PR head) plus a full-repo grep, not just the diff.

`callback.py:177` now reads `torch.cuda.max_memory_allocated()` on every `on_log` call, with nothing resetting the peak counter between calls anywhere on this path. `grep -rn "reset_peak_memory_stats"` across the repo finds it only in `bench.py`, `layer_stream.py`, `layer_stream_runtime.py`, and tests, none of which run during `soup train` (checked `pretrain.py`, `reward_model.py`, `ppo.py`, `simpo.py`, `asr.py`, where `SoupTrainerCallback` is constructed). So the panel will show the single highest-memory step of the whole run from the point that step happens onward, not the peak of the current logging window, which is what issue #650 asked for ("if cumulative, say so in the label").

Issue #650's acceptance criteria lists this explicitly as one of two acceptable fixes: reset after each read, or label the number as cumulative. This PR does neither, so a run whose early steps peak higher than later ones (a curriculum with decreasing batch size, or a memory-heavy eval interleaved with lighter train steps) will show a flat, stale peak for the rest of training, same failure shape as the trough problem #650 reported, just inverted.

Suggest adding `torch.cuda.reset_peak_memory_stats()` right after the read at line 177, so each `on_log` call reports the peak for its own window rather than the lifetime peak.

Also: the PR doesn't add a `changelog.d/0.73.3/652.fixed.md` fragment, which `changelog.d/README.md` and #650's acceptance criteria both call for.

webdevsamran:
Thanks @MakazhanAlpamys and @AmirF194 for the reviews!

All feedback addressed:
1. **Label updated**: Changed the display panel label in src/soup_cli/monitoring/display.py to GPU peak: so the metric is clearly self-describing.
2. **Cumulative peak vs per-window**: As @MakazhanAlpamys noted, 	orch.cuda.reset_peak_memory_stats() is process-global; calling it in on_log would clobber the grad-accum advisor's own peak reading at callback.py:566 (violating acceptance criterion 3: *'No change to any decision path'*). Keeping max_memory_allocated() as the lifetime high-water mark with the explicit GPU peak: label preserves the advisor while ensuring the display figure is accurate and unambiguous.
3. **Tests updated**: Added an assertion in 	ests/test_display.py (	est_display_render_panel) verifying the GPU peak: label renders in the panel. All 16 tests in 	est_callback.py and 	est_display.py pass.
4. **Changelog fragment**: Added changelog.d/0.73.3/652.fixed.md crediting (#650 by @YuriPerro in #652).

AmirF194 (commented):
Makes sense, the process-global reset would have clobbered the grad-accum advisor's own reading at callback.py:566. The label plus changelog fragment cover it.

MakazhanAlpamys:
Merged as `6f27066`. **Welcome to CONTRIBUTORS** (`e1f1b32`) — this is your first contribution here and it is a good one.

Credited specifically for fixing **the half that mattered**. Swapping the counter is the obvious change; the reporter's actual complaint was *"an unlabeled single number is the problem"*, and `GPU peak:` is what closes it. A number that is right but still unnamed would have left an operator comparing a second unexplained quantity against `nvidia-smi`.

The other thing I want to name is that you **wrote down why** the lifetime peak is used rather than a per-window one — `reset_peak_memory_stats()` being process-global and clobbering the grad-accum advisor at `callback.py:566` — and put it at the call site rather than only in the PR description. That comment is what stops someone "improving" this in six months and quietly breaking #650's third criterion.

Verified by mutation before merging, with exact full-line anchors:

```
used = torch.cuda.max_memory_allocated()  ->  memory_allocated()
    KILLED by test_on_log_gpu_memory_uses_max_allocated
lines.append(f"GPU peak: {self.gpu_mem}")  ->  "GPU:   {self.gpu_mem}"
    KILLED by test_display_render_panel
```

Both halves die, so neither can be reverted silently. 16 tests pass, zero deleted test lines, 14/14 CI including `pytorch-smoke`. Your mocked-torch test is the right shape too — 2 GB from `memory_allocated`, 8 GB from `max_memory_allocated`, assert the panel shows 8.0 — so it fails for the specific reason it exists rather than incidentally.

**A correction I owe you about my own review.** My first mutation run reported the counter change as *surviving*, which would have been a blocking finding. It was my error: the replacement landed on a comment that mentions `max_memory_allocated()` rather than on the code. I re-ran it with a full-line anchor and it dies. I mention it because I nearly published a wrong finding against your PR, and you would have had no way to tell it was mine rather than yours.

Thanks also to @YuriPerro, who reported the bug, held first refusal on implementing it, and declined in favour of this rather than duplicating finished work.

Closes #650.

PR #653, opened 2026-09-03
test: cover _build_source(pin=False, tier=ram) on CPU-only CI

### LINKED ISSUE #623: `stream_pin: false` is accepted and silently ignored — the store still lands in shared memory (follow-up to #366)

AmirF194:
## What does this PR do?

Fixes #623: a CPU-only assertion that `_build_source(pin=False, tier="ram")` returns a `RamSource` with `.pinned is False`, the one thing you left the issue open for after #647 landed.

`_build_source`'s `if not pin:` branch (`src/soup_cli/utils/layer_stream_runtime.py:2212-2213`) is the only branch that is never CUDA-gated, and neither existing test drives it with a real, unmocked construction: `TestRequirePinSurvivesEveryHop` (test_v07203.py) only builds a `RamSource` subclass that fails when `pin=True`, and this file's own `TestTheAllocationItself` is `@requires_cuda`. The new test calls `_build_source` directly against a real one-tensor shard and checks `.pinned` and `.get(0, "weight").is_pinned()` on the constructed `RamSource`, no CUDA needed since `torch.empty(pin_memory=False)` works on any host.

## Type of change

- [x] Tests

## Verification

- New test run against current HEAD on Python 3.10, 3.11 and 3.12 (the full CI matrix's version set): passes on all three.
- `ruff check src/soup_cli/ tests/test_issue623_stream_pin_allocation.py`: clean.
- Scoped to `torch` + `safetensors` + `numpy` (the module's own lazy imports), so this did not exercise `pytest tests/` in full; the touched file's other classes need the `[train]` stack, which I did not install.

## Checklist

- [x] `pytest tests/ -v` passes (scoped run above; see note)
- [ ] Updated relevant docs (not applicable, test-only)

### DIFF
--- a/tests/test_issue623_stream_pin_allocation.py
+++ b/tests/test_issue623_stream_pin_allocation.py
@@ -35,13 +35,20 @@ allocate GPU buffers:
   with a stated reason elsewhere, never silently) — no stubs, real
   ``RamSource``, ``is_pinned()`` on every store tensor, both directions.
 
-Hops BELOW the runtime boundary are not re-tested here:
+Most hops BELOW the runtime boundary are not re-tested here:
 ``TestRequirePinSurvivesEveryHop`` (test_v07203.py) already drives the real
 CPU chain through ``build_streamed_model -> install_streaming ->
 _build_source -> RamSource`` for the pin-requested direction, and the
 pre-flight note for ``stream_pin: false`` is pinned by
 ``TestAutoTierFallback``. CPU-only throughout; the tiny checkpoint is the
 established harness fixture shape.
+
+One hop below the boundary IS covered here, added once #647 closed the
+allocation gap above and left the unpinned RAM-tier construction as the one
+remaining piece unprotected on a CUDA-less CI cell:
+``TestBuildSourceCpuOnlyPinFalse`` calls ``_build_source(pin=False,
+tier="ram")`` directly, no simulation needed, since its unpinned branch never
+requests page-locked memory.
 """
 
 from __future__ import annotations
@@ -249,6 +256,43 @@ class TestTheWiringReachesTheRuntimeBoundary:
         )
         assert captured["require_pin"] is False
         assert "(pinned)" in panel
+
+
+class TestBuildSourceCpuOnlyPinFalse:
+    """The gap named in the issue thread once #647 landed: the two deepest
+    hops below the runtime boundary (this file's own ``TestTheAllocationItself``,
+    ``TestRequirePinSurvivesEveryHop`` in test_v07203.py) are CUDA-gated, so
+    the RAM tier's unpinned path is unprotected on all nine CI cells. Drives
+    the real, unmocked ``_build_source(pin=False, tier="ram")`` directly: its
+    ``if not pin:`` branch never reaches ``pin_memory=True``, so nothing here
+    needs a CUDA device.
+    """
+
+    def test_build_source_pin_false_ram_tier_yields_unpinned_ramsource(
+        self, tmp_path
+    ) -> None:
+        import torch
+        from safetensors.torch import save_file
+
+        from soup_cli.utils.layer_shard import layer_shard_path
+        from soup_cli.utils.layer_stream_runtime import RamSource, _build_source
+
+        shard_dir = tmp_path / "shards"
+        shard_dir.mkdir()
+        save_file(
+            {"weight": torch.zeros(4, 4, dtype=torch.float32)},
+            layer_shard_path(str(shard_dir), 0),
+        )
+        spec = {"weight": ((4, 4), "float32")}
+
+        source, pinned = _build_source(
+            str(shard_dir), 1, spec, pin=False, console=None, tier="ram"
+        )
+
+        assert isinstance(source, RamSource)
+        assert pinned is False
+        assert source.pinned is False
+        assert source.get(0, "weight").is_pinned() is False
 
 
 @requires_cuda


MakazhanAlpamys (approved):
**Approved — this closes exactly what I left #623 open for, and the mutation that matters dies.**

I verified the head (`52f85be4`), then ran the two hops this is supposed to protect:

| mutation | result |
|---|---|
| `if not pin:` → `if False:` in `_build_source` (`:2212`) | **KILLED** — `test_build_source_pin_false_ram_tier_yields_unpinned_ramsource` |
| `pin_memory=self.pinned` → `pin_memory=True` in `RamSource` (`:331`) | **KILLED** — same test |

The second one is the point. Before #647 that mutation left **169 pre-existing streaming tests green**, and after #647 it was caught only behind a CUDA gate. It now dies on a host with no GPU, which is what the issue asked for.

`source.get(0, "weight").is_pinned() is False` is why it works — you assert the **allocation**, not just `source.pinned`. An attribute assertion would have passed the deeper mutation happily, and that distinction is the whole reason the gap existed in the first place.

Driving the real, unmocked `_build_source` with a genuine safetensors shard on disk, rather than a mock, is the right call and it costs nothing: the unpinned branch never requests page-locked memory, so there is no device requirement to work around.

## One structural limit, worth naming rather than fixing

The mirror direction is not covered and **cannot be** on a CPU-only cell. I mutated `pin_memory=self.pinned` → `pin_memory=False` — a "never pin" implementation:

```
TestBuildSourceCpuOnlyPinFalse:  1 passed      <- survives
TestTheAllocationItself:         1 failed      <- caught, but CUDA-gated
```

So a regression that silently stopped pinning altogether is still invisible on the nine CPU cells. That is not a defect in this PR: you cannot allocate pinned memory without CUDA, so the `pin=True` direction is untestable there by construction. I am recording it so nobody later reads this file as full coverage of both directions — the pin-requested direction remains CUDA-only, and `TestRequirePinSurvivesEveryHop` in `test_v07203.py` is where it lives.

## Checks

- The new test passes standalone. **14/14 CI green including `pytorch-smoke`** — you are the first PR to land after I made that required, so it actually reported.
- `ruff` clean.
- **One removed line, and it is not a test:** `Hops BELOW the runtime boundary are not re-tested here:` → `Most hops BELOW...`. Correcting a docstring that your own change makes false is the detail I would have let slide.
- No changelog fragment, correctly — `CONTRIBUTING.md` scopes those to user-visible changes and this is tests-only.

One note on my own run so it is not mistaken for a finding: three pre-existing tests in that file fail on **my** box with `ImportError: huggingface-hub>=0.34.0,<1.0 is required ... found huggingface-hub==1.30.0`. I checked the same file on `main` and it fails identically, so it is my local environment and not your branch. CI is green on all nine cells.

Merging shortly. Thank you for picking up a gap I had said I would file rather than waiting for the issue to appear.

MakazhanAlpamys:
Merged as `9d727e0`, credited in CONTRIBUTORS at `05b6465`. **#623 is closed.**

Credited specifically for the thing that makes it work rather than for the test existing: asserting `source.get(0, "weight").is_pinned() is False` reaches the allocation, where an assertion on `source.pinned` would have passed the deeper mutation unchanged. That is exactly how the gap survived #647 — and I know because I ran that mutation before merging #647 and watched 169 tests stay green.

Re-ran both hops before merging: `if not pin:` → `if False:` and `pin_memory=self.pinned` → `pin_memory=True`, both **KILLED** by your test on a host with no GPU involvement in that path. 14/14 CI including `pytorch-smoke` — you are the first PR to land since I made it required, so it actually reported.

You also did the two things I notice most: picked this up from the issue thread instead of waiting for me to file it, and **named the structural limit rather than implying full coverage.** I confirmed that limit is real — mutating `pin_memory=self.pinned` to `False` (a never-pin implementation) survives your test and is caught only by the CUDA-gated `TestTheAllocationItself`. Since you cannot allocate pinned memory without a device, that direction is untestable on a CPU cell by construction, and saying so in the docstring is worth more than a test that pretends otherwise.

Separately: thank you for the resident-RAM accounting gap you found on #644. That is now `a82a947`, and it will be credited in that merge too.

AmirF194:
Good catch on the docstring line, that assertion change did make the old wording wrong. And thanks for actually running the pin_memory=False mutation and writing down that it's untestable on CPU rather than treating the gap as covered, that's the part I'd have wanted a reviewer to check.

Issue #325, opened 2026-07-28
Layer streaming: the RAM-vs-disk throughput gap is unmeasured (needs hardware where it is observable)


MakazhanAlpamys:
## Context

v0.72.3 shipped the disk overflow tier. Its **correctness** is verified — streamed-from-disk logits are bit-exact against both the RAM tier and a resident model. Its **speed relative to RAM is not characterised**, and the release deliberately claims no number for it.

## Why it could not be measured on the development box

Two independent reasons, either sufficient:

1. **safetensors memory-maps the shards.** On a machine with spare RAM the OS page cache keeps them resident between steps, so a "disk" run is partly a RAM run. There is no clean A/B without controlling the page cache.
2. **The development box is compute-bound.** At its effective throughput the NVMe read hides under compute, so even a genuine disk read would not show up in tok/s.

Publishing a fabricated or accidentally-cached number would be worse than publishing none, so the release states the gap is unmeasured.

## What would actually answer it

- A box where the base does **not** fit in RAM, so the page cache cannot mask the reads
- Or a controlled comparison that drops the page cache between runs
- Ideally a card fast enough that I/O is not hidden by compute
- Report tok/s, GPU utilisation and peak VRAM for RAM vs disk at matched batch/seq, with free RAM and SM clock recorded per row (the measurement discipline the rest of the streaming numbers follow)

## Acceptance criteria

- [ ] A RAM-vs-disk comparison at matched configuration, with the page-cache condition stated explicitly
- [ ] Free-RAM and SM-clock recorded alongside each row
- [ ] `docs/performance-and-quantization.md` updated to replace "unmeasured" with the measured figure and its conditions

Related: the same hardware constraint blocks the 14B reference benchmark.

AchuthReddy-16:
I’m interested in working on this and have relevant experience benchmarking LLM inference throughput, GPU utilization, latency, and memory behavior. Before claiming it, I want to verify that I can access hardware where the RAM-vs-disk difference is observable. Could you share the exact benchmark command/model used for the existing layer-streaming measurements and the minimum expected host RAM, GPU, and storage setup? I’ll first run a feasibility check and report the hardware and page-cache conditions before producing numbers.

MakazhanAlpamys:
@AchuthReddy-16 — you asked for the exact command, model and hardware minimums on 2026-08-25 so you could run a feasibility check before claiming, which is the right order, and I left it for **ten days**. Sorry. Everything you asked for is below.

[...]

PR #654, opened 2026-09-03
fix(batch-probe): gate the CUDA probe on the measured peak, not only on OutOfMemoryError (#649)

### LINKED ISSUE #649: `make_cuda_probe_fn` treats only `OutOfMemoryError` as "does not fit", so under WDDM (Windows / WSL2) the batch probe approves sizes that spill to host memory
## What

`utils/batch_probe.py::make_cuda_probe_fn` decides "fits" by the absence of
`torch.cuda.OutOfMemoryError`. Under the WDDM driver (native Windows, WSL2) the
allocator does not raise when dedicated VRAM is exhausted: it spills into
"shared GPU memory" (host RAM over PCIe) and the step completes, an order of
magnitude slower. The probe therefore returns `True` for a batch that does not
physically fit, `pick_batch_size` caches it in `~/.soup/batch_cache.json`, and
every later run on that key starts from the spilled size.

`auto_batch_size_strategy` defaults to `auto`, which is the probe on CUDA
(`config/schema.py:1043`), so this is the default path for every SFT run with
`batch_size: auto` on those platforms.

The layer-streaming probe already knows about this. `layer_stream_runtime.py:1410`
reads `max_memory_allocated` / `max_memory_reserved` after the step and its
comment says WDDM "reaches the success path with a peak above free VRAM, which
the caller refuses just the same". The batch probe reads nothing after the step
and refuses nothing. #349 named the same trap ("on Windows there is nothing to
catch"). This is the second probe not inheriting the first one's defence, the
same shape as #491 was for #339.

## Reproduction

RTX 5080 16 GB (15.9 GB dedicated) · WSL2 Ubuntu 26.04, kernel 6.18.33.2-microsoft-standard-WSL2 · Windows 11 Pro 26200 · 63 GB host RAM ·
soup-cli 0.73.3 (PyPI) · torch 2.14.0+cu130 · transformers 4.57.6 · peft 0.20.0 · trl 0.28.0 · bitsandbytes 0.50.2 · Python 3.12.14

```yaml
base: Qwen/Qwen2.5-1.5B-Instruct
task: sft
data:
  train: ./data/train.jsonl   # 450 rows, 112-152 tokens each
  format: alpaca
  val_split: 0.1
  max_length: 512
training:
  epochs: 2
  lr: 2e-4
  batch_size: auto
  gradient_accumulation_steps: 1
  lora: {r: 16, alpha: 32, target_modules: auto}
  quantization: none
output: ./output/qwen1.5b-json
```

Probe output, verbatim:

```
[W902 20:59:52.456610156 CUDACachingAllocator.cpp:3934] memory allocation failed with OOM on device 0 while trying to allocate 587202560 bytes (free: 0, total: 17094475776).
[W902 20:59:52.536554508 CUDACachingAllocator.cpp:3934] memory allocation failed with OOM on device 0 while trying to allocate 587202560 bytes (free: 0, total: 17094475776).
Auto batch size: 16
```

Observed during training (Windows perf counters on the `vmwp` process, i.e. the WSL2 VM):

| batch | dedicated VRAM | shared GPU memory (host RAM) | it/s | outcome |
|---|---|---|---|---|
| 16 (probe-approved) | 14.98 GB / 15.9 | 21.46 GB | 0.12 | completes; panel shows `GPU: 5.8/15.9 GB` |
| 8 (explicit) | 15.65 GiB reserved, 12.69 GiB allocated | (not sampled) | n/a | `torch.AcceleratorError: CUDA error: out of memory` from `cuStreamSynchronize`, first `training_step`, at TRL's `entropy ... .item()` |

The second row matters: when the spill eventually does fail, it surfaces as a
generic `AcceleratorError` at a synchronize, not as `OutOfMemoryError` from the
allocator, so even the exception the probe *does* catch is not the one WDDM
raises. Same config re-run with `batch_size: auto` (cache hit, 16) completed.

Note on the base dtype: 0.73.3 loads the base in fp32 (fixed on main by #471 /
#492, not yet released), which is what makes a 1.5B model reach the spill at
batch 16 on a 15.9 GB card. The probe defect is independent of that and persists
on main (`batch_probe.py` is byte-identical between v0.73.3 and main). On main
with the bf16 fix the same batch 16 runs at 0.98 it/s and batch 8 at 4.15 it/s,
which suggests 16 still touches the spill; the probe still picks 16.

## Suggested fix

After the synthetic step, read `torch.cuda.max_memory_reserved(device)` (after
`reset_peak_memory_stats` before it) and treat a peak above the device's
physical `total_memory` (minus headroom) as "does not fit", the same way the
layer-streaming probe does. Catching `torch.AcceleratorError` (torch >= 2.x
spelling; `RuntimeError` with "out of memory" in the message on older stacks)
alongside `OutOfMemoryError` would cover the second failure mode. The cache
entry should record which criterion decided, so a spilled size written by an
older Soup can be invalidated.

## Acceptance criteria

- [ ] A candidate whose measured peak reserved exceeds physical VRAM is reported
      as a failure by the probe, on any driver, without relying on an exception.
- [ ] A `cuStreamSynchronize` "out of memory" `AcceleratorError` is treated like
      `OutOfMemoryError`, not propagated as misconfiguration.
- [ ] A test that fails when the check is removed: mock `max_memory_reserved`
      above `total_memory` on a step that raises nothing and assert `False`.
      Must not need a GPU.
- [ ] A control: a step under the ceiling still returns `True`, so a fix that
      refuses everything is not an improvement.
- [ ] `docs/performance-and-quantization.md` "OOM-probe auto batch size" states
      the WDDM behaviour.

## Not covered

Whether the spilled batch is *slow* is not the claim here; the claim is that the
probe's decision is wrong. Throughput on WDDM belongs with #325 / #395.
The panel readout that hid the peak is a separate issue: #650.

YuriPerro:
Closes #649.

## What

1. `make_cuda_probe_fn` now refuses a candidate whose **measured peak** after the synthetic step exceeds what this process can reach on the device, in addition to refusing on a raise. Under WDDM (native Windows, WSL2) the allocator spills to host memory instead of raising, so "the step did not throw" was the entire fit criterion and it approved a batch that did not fit.
   - Counter: `max_memory_allocated`, deliberately not `max_memory_reserved`. Reserved runs 1.08-1.41x allocated and gating on it refuses configurations that run; this is the basis `layer_stream.decide_measured_fit` already documents, and the probe mirrors it rather than picking a fraction.
   - Threshold: `mem_get_info()[0] + memory_allocated()` read before the step (what this process holds plus what is free), compared exactly. The probe's peak already runs above the real step, which is the direction that makes an exact comparison safe. When the driver cannot answer, the probe keeps the exception-only criterion rather than inventing a budget.
2. OOM raised at a synchronize point is classified as OOM: `torch.AcceleratorError("CUDA error: out of memory")` (torch >= 2.8) and the older `RuntimeError` spelling. This is the exception WDDM eventually raises (#649 reproduction, second row). An `AcceleratorError` with any other message still propagates.
3. Cache key carries a version tag (`v2`), so entries written by the exception-only probe are never found again. No migration code: a poisoned `batch_cache.json` simply stops matching.
4. `docs/performance-and-quantization.md`: the OOM-probe section states the WDDM behaviour in both places it describes the loop.

## Tests

`tests/test_issue649_batch_probe_peak_gate.py`, fake `torch` in `sys.modules`, no GPU:

| test | mutation it catches |
|---|---|
| step completes, peak 21 GB on a 16 GB card → `False` | removing the peak gate (the WDDM case) |
| step completes under budget → `True` | a gate that refuses everything |
| peak 13 GB, 5 GB free + 6 GB held → `False` | gating on `total_memory` instead of reachable bytes |
| peak == budget → `True` | an off-by-one / strict comparison |
| reserved 1.3x over budget, allocated under → `True` | gating on reserved |
| `AcceleratorError("... out of memory")` → `False` | not classifying the synchronize spelling |
| `RuntimeError("CUDA out of memory ...")` → `False` | same, older torch |
| `AcceleratorError("illegal memory access")` → raises | swallowing non-OOM device errors |
| key for the same tuple != pre-#649 sha256 | forgetting the cache bump |

The fake asserts `reset_peak_memory_stats` was called before `max_memory_allocated` is read, so a stale peak from an earlier candidate cannot leak into the decision.

## Verification

- `ruff check src/soup_cli/ scripts/ tests/` clean.
- `tests/test_batch_probe.py` (31) + the new file (10) pass; full suite result to be posted below.
- Live WDDM run on the reporting box (RTX 5080 16 GB, WSL2) pending: I will post the probe's decision for the batch 16 case from #649 before marking ready for review.

## Scope limits

- The probe's speed under WDDM is out of scope (#325 / #395); this PR changes only its decision.
- The panel readout that hid the peak is #650.
- `mem_get_info` cannot see a per-process cap (`set_per_process_memory_fraction`, MIG); the streaming pre-flight has an override for that and the batch probe does not grow one here.

### DIFF
--- a/changelog.d/0.73.3/654.fixed.md
+++ b/changelog.d/0.73.3/654.fixed.md
@@ -1,0 +1,16 @@
+- **The CUDA batch probe now refuses on a measured peak, not only on a raise
+  (#649 by @YuriPerro in #654).**
+  Under the WDDM driver (native Windows, WSL2) the allocator spills to host
+  memory instead of raising `OutOfMemoryError`, so `make_cuda_probe_fn`'s only
+  fit criterion, "the step did not throw", approved a batch 16 that ran an
+  order of magnitude slower in shared memory and `pick_batch_size` cached it.
+  The probe now reads `max_memory_allocated` after the synthetic step and
+  refuses when it exceeds what this process can reach on the device
+  (`mem_get_info` free plus what it already holds), mirroring the counter and
+  basis `decide_measured_fit` documents for layer streaming; an OOM surfacing
+  at a synchronize point as `AcceleratorError` / `RuntimeError("out of
+  memory")` is classified as OOM instead of propagating as misconfiguration;
+  and the cache key carries a version tag so entries written by the old probe
+  are ignored. The WDDM case is covered by a mocked-torch test that needs no
+  GPU, with a fitting control beside it.
+  Closes #649.
--- a/docs/performance-and-quantization.md
+++ b/docs/performance-and-quantization.md
@@ -675,6 +675,8 @@ training:
 
 Replaces the static memory formula with a real try-halve-then-double-to-ceiling loop. Picked size is cached at `~/.soup/batch_cache.json` keyed on `(model, max_length, quantization, lora_r, gpu_name, gpu_memory_gb)` so repeat runs short-circuit.
 
+A candidate is refused when the step raises an out-of-memory error **or** when it completes with a measured peak (`max_memory_allocated`) above the VRAM this process can reach. The second check exists for the WDDM driver (native Windows and WSL2), where the allocator spills to host memory instead of raising: the step finishes, an order of magnitude slower, and without the measurement the probe would approve a batch that does not fit and cache it (#649). Cache entries written before that check carry a different key and are ignored.
+
 
 ## Multi-GPU / DeepSpeed / FSDP
 
@@ -815,7 +817,7 @@ training:
   auto_batch_size_strategy: probe
 ```
 
-For each candidate size `B`, the probe runs ONE forward + backward + step on a synthetic batch of `B` sequences of length `max_length`. On `torch.cuda.OutOfMemoryError` it halves; otherwise it doubles up to `4 × static_estimate`. The picked size is cached per `(model, max_length, quantization, lora_r, gpu)` tuple in `~/.soup/batch_cache.json` so subsequent runs skip the probe.
+For each candidate size `B`, the probe runs ONE forward + backward + step on a synthetic batch of `B` sequences of length `max_length`. On an out-of-memory error, or on a measured peak above what this process can reach on the device (the WDDM spill case, #649), it halves; otherwise it doubles up to `4 × static_estimate`. The picked size is cached per `(model, max_length, quantization, lora_r, gpu)` tuple in `~/.soup/batch_cache.json` so subsequent runs skip the probe.
 
 CPU sessions and `auto_batch_size_strategy: static` skip the probe. Synthetic batch tensors are freed before the backward pass so peak VRAM reflects the realistic training step. SFT-only this release — non-SFT trainers fall back to the static estimate.
 
--- a/src/soup_cli/utils/batch_probe.py
+++ b/src/soup_cli/utils/batch_probe.py
@@ -25,6 +25,9 @@ from typing import Any, Callable, Optional
 # Stay safe — never go below 1; never run forever.
 _MIN_BATCH = 1
 _DEFAULT_MAX_DOUBLINGS = 8
+# Folded into every cache key. "v2" = probe gates on the measured peak (#649);
+# keys without it were written by the exception-only probe and are ignored.
+_CACHE_KEY_VERSION = "v2"
 
 ProbeFn = Callable[[int], bool]
 
@@ -149,7 +152,13 @@ def make_cache_key(
     gpu_name: str,
     gpu_memory_gb: int,
 ) -> str:
-    """Stable string key for the cache. Hashed for filesystem safety."""
+    """Stable string key for the cache. Hashed for filesystem safety.
+
+    ``_CACHE_KEY_VERSION`` is folded into the hash so an entry written by an
+    older probe is simply never found. Bumped in #649: the exception-only probe
+    approved batches that spilled to host memory under WDDM and cached them, and
+    a cached wrong answer is recomputed by nobody.
+    """
     for name, value in (
         ("max_length", max_length),
         ("lora_r", lora_r),
@@ -159,6 +168,7 @@ def make_cache_key(
             raise ValueError(f"{name} must be an int (got {type(value).__name__})")
     raw = "|".join(
         [
+            _CACHE_KEY_VERSION,
             str(base),
             str(max_length),
             str(quantization),
@@ -293,6 +303,42 @@ def pick_batch_size(
 # ---------------------------------------------------------------------------
 
 
+def _is_cuda_oom(exc: BaseException, torch: Any) -> bool:
+    """Is ``exc`` the device running out of memory, in any of torch's spellings?
+
+    The allocator raises ``torch.cuda.OutOfMemoryError``. An OOM that surfaces
+    later at a synchronize point does not: torch >= 2.8 raises
+    ``torch.AcceleratorError("CUDA error: out of memory")`` and older releases a
+    plain ``RuntimeError`` with the same text. #649 observed the second form
+    under WDDM, where the allocator had already spilled instead of raising.
+    Anything else (an illegal access, a device assert) is not a fit answer and
+    must propagate.
+    """
+    oom_cls = getattr(torch.cuda, "OutOfMemoryError", None)
+    if oom_cls is not None and isinstance(exc, oom_cls):
+        return True
+    return isinstance(exc, RuntimeError) and "out of memory" in str(exc).lower()
+
+
+def _probe_budget_bytes(torch: Any, device: str) -> Optional[int]:
+    """Bytes this process can reach on ``device``: what it holds plus what is free.
+
+    ``mem_get_info()`` is a device-level driver query, so it excludes VRAM held
+    by other processes (the streaming pre-flight relies on the same reading,
+    see :func:`~soup_cli.utils.layer_stream.resolve_available_vram_bytes`).
+    Under WDDM it reports physical VRAM: the allocator's own warning in #649
+    read ``free: 0`` while the step kept going in host memory. Returns ``None``
+    when the driver cannot answer, in which case the caller keeps the
+    exception-only criterion rather than inventing a budget.
+    """
+    try:
+        free, _total = torch.cuda.mem_get_info(device)
+        held = torch.cuda.memory_allocated(device)
+        return int(free) + int(held)
+    except Exception:  # noqa: BLE001 — a driver that cannot answer is not a fit answer
+        return None
+
+
 def make_cuda_probe_fn(
     model: Any,
     tokenizer: Any,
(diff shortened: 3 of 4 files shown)


YuriPerro:
Full suite on this branch: `19883 passed, 106 skipped, 5 deselected in 11m10s` (macOS, no GPU, torch mocked in the new tests). `ruff check src/soup_cli/ scripts/ tests/` clean. Changelog fragment added as `changelog.d/0.73.3/654.fixed.md` and validated with `scripts/assemble_changelog.validate_fragments`. Still pending before I mark ready: the live WDDM run on the RTX 5080 / WSL2 box from the report.

YuriPerro:
## Live WDDM verification (the box from #649)

RTX 5080 16 GB (15.92 GB reported), WSL2 Ubuntu 26.04, torch 2.14.0+cu130, transformers 5.16.1, this branch installed editable. Qwen2.5-1.5B-Instruct bf16 + LoRA r=16 on all-linear, `max_length` 512, default caching allocator (no `expandable_segments`), same shape as the report.

Calling the probe closure directly and reading the counters after each candidate:

| batch | budget (free + held, before step) | peak `max_memory_allocated` | peak reserved | raised? | probe says |
|---|---|---|---|---|---|
| 4 | 14.37 GB | 14.30 GB | 15.32 GB | no | `True` |
| 6 | 14.32 GB | **20.02 GB** | 21.66 GB | no | `False` |
| 8 | 14.32 GB | **25.60 GB** | 27.38 GB | no | `False` |
| 16 | 14.32 GB | **48.20 GB** | 51.54 GB | no | `False` |

That is the spill in one line: a 15.92 GB card reporting a 48 GB peak with nothing raised. Every one of those rows was `True` under the exception-only criterion.

Real run, `soup train` with `batch_size: auto` and the cache removed:

```
Auto batch size: 4
Speed: 6.59 it/s
```

against `Auto batch size: 16` at 0.12 it/s in the report. New cache entry is written under the `v2` key; the old `01612cff…: 16` entry no longer matches.

Two observations, neither changed in this PR:

- With `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` the spill does not happen: batch 8 raised `RuntimeError: CUDA driver error: device not ready` ("Failed to create GPU mapping") from inside the forward. That is a third spelling of "did not fit" on WDDM. It is not classified as OOM here because I cannot establish that the message means memory exhaustion in general rather than a mapping failure; it propagates, as any non-OOM device error did before. Noting it so it is on record if it recurs.
- The probe measures more than the training step will use: the probe runs before `gradient_checkpointing` is applied by the trainer, so at batch 8 it peaks at 25.6 GB where the same config trains fine (run 3 in the report, 5.69 it/s). On Linux the allocator would have raised at that peak and the probe would also have settled on 4, so this PR makes WDDM agree with Linux rather than introducing the conservatism. Whether the probe should mirror the checkpointing setting is a separate question from #649.

Marking ready for review.

MakazhanAlpamys:
Merged as `eb98893`. **Welcome to CONTRIBUTORS** (`6bddd7d`).

**I reproduced your bug on unrelated hardware before merging**, which is the strongest thing I can say about a report: RTX 3050, 4.29 GB card, native Windows, 3.46 GB free, driving the real probe against a real device.

```
batch  peak      main    #654
   64  3.26 GB   True    True     <- control, genuinely fits
   96  4.87 GB   True    False    <- 1.13x the whole card, approved on main
```

So this is not WSL2-specific and not a property of your stack. A second box, a different GPU, same defect.

**Your diagnosis is worth more than the patch**, and it is the part I could not have written: *the second probe did not inherit the first one's defence.* Layer streaming already reads `max_memory_*` after its step for exactly this reason, and that guard was written for one probe and never generalised. That framing is what turned a bug report into a class of bug.

Both load-bearing choices are argued from measurements **already in this repository** rather than picked for looking reasonable — `max_memory_allocated` over `reserved` (1.08-1.41x, not a constant), and **no safety fraction at all**, because the measured probe is already +12.5-14.3% conservative and an invented margin would stack on a known one. I checked the provenance and both citations are real. That is the standard #644 was held to this week, met properly.

Two things you added that I had not asked for: the `AcceleratorError`-at-`synchronize` case — an async OOM is the same event by a different route — and the cache-key bump, which is the detail most people skip. A user who hit this once had a poisoned `batch_cache.json` and no reason to suspect it, so a fix that only helps new installs is half a fix.

**10 of 11 mutations killed**, zero deleted test lines, 14/14 CI including `pytorch-smoke`.

**Two follow-ups, neither blocking, both recorded in the merge commit:**

1. **The graceful-degradation branch has no test.** `if budget is None: return True` — flipping it to `False` would make the probe refuse *every* batch on any stack where `mem_get_info` is unavailable, and nothing catches it. One fake-torch test whose `mem_get_info` raises would close it.
2. **`AcceleratorError` is handled by inheritance, and the test cannot see that.** The code relies on it subclassing `RuntimeError`, which is correct upstream — but your `_FakeAcceleratorError` also subclasses `RuntimeError`, so the test passes either way, and local torch 2.5.1 has no `AcceleratorError` for any CI cell to check the real assumption.

Also: thank you for **declining #650** when you saw #652 already delivered it, and for taking #649 instead — the harder one, and the one needing your hardware. That is what kept the queue moving.

Srinivasan8888:
Taking **both follow-ups** from this merge comment. @YuriPerro — these are gaps in your PR, so if you would rather close them yourself say so and I will drop it immediately; you found the defect and the follow-ups are yours by right of first refusal.

**Overlap check:** no open issue or PR covers either. Nothing in the open queue touches `batch_probe.py`.

**Follow-up 1 — the `budget is None` branch** is exactly as described, and I verified the cause rather than taking it on trust: `tests/test_issue649_batch_probe_peak_gate.py:77` builds the fake as `cuda.mem_get_info = lambda *a, **k: (free_before, total)`, which cannot fail, so `_probe_budget_bytes` never returns `None` and `if budget is None: return True` is never executed. A fake whose `mem_get_info` raises closes it, as you said.

**Follow-up 2 — and here I think one premise is worth correcting, with evidence.**

You wrote that "local torch 2.5.1 has no `AcceleratorError` for any CI cell to check the real assumption." The first half is about your box, but the second does not follow, because CI does not install your torch:

```
pyproject: dev = ["soup-cli[train,...]", ...]   ->   train = ["torch>=2.5.0", ...]
```

`>=2.5.0` is a floor, not a pin, so `pip install -e ".[dev]"` resolves to the newest wheel. Measured in this repo's own `.venv`:

```
torch 2.13.0
torch.AcceleratorError present: True
  mro: ['AcceleratorError', 'RuntimeError', 'Exception', 'BaseException', 'object']
  issubclass(RuntimeError): True
```

And your own WDDM verification ran torch 2.14.0. So **the assumption is checkable in CI today** — a guard reading the real installed `torch.AcceleratorError` executes rather than skips in every `test` matrix cell. That turns "correct upstream, unverified here" into something that goes red the day it stops being true.

I will follow `tests/test_issue636_torch_floor.py`'s convention for this: **skip with a stated reason rather than pass vacuously** when the symbol is absent, so a 2.5.x environment says so instead of quietly proving nothing.

**Scope, and one thing I am deliberately NOT doing.** You handle `AcceleratorError` by inheritance. That is correct on every torch I can reach, so I am adding a *test* — the thing asked for — and **not** rewriting `_is_cuda_oom` to match it by name. Making the classifier independent of the inheritance is a real option (it would keep working if upstream ever reparented the class, where today the probe would propagate instead of shrinking the batch), but it is a behaviour change to correct code and therefore @MakazhanAlpamys's call, not mine to slip into a test PR. I will include the control that demonstrates the dependency is real, and if you want the by-name change I will do it as a separate PR.

Plan: both tests in one file, a mutation table showing each one fails by name when the branch it guards is flipped, and the existing suite untouched.

MakazhanAlpamys:
Both follow-ups this merge comment recorded as untested are now closed by #657 (@Srinivasan8888), merged as `000dbc4`:

1. The `budget is None` graceful-degradation path in `_probe` — previously unreachable from the suite because `test_issue649_*.py:77` builds `mem_get_info` as a `lambda` that cannot fail.
2. The `AcceleratorError` contract — the existing `_FakeAcceleratorError` also subclasses `RuntimeError`, so that test passed whether or not the real class does.

And a correction to this comment, which was mine: I wrote that no CI cell could check the real assumption because the box had torch 2.5.1. `train = ["torch>=2.5.0"]` is a floor rather than a pin, so CI resolves a recent wheel and the guard executes; where the symbol is genuinely absent the new test skips with a stated reason instead of passing vacuously.

Credit for the defect itself stays with @YuriPerro, who found it and verified the WDDM behaviour, and who was offered right of first refusal on this follow-up before it was written.

PR #660, opened 2026-09-04
fix: wire MLX backend into soup doctor and read the real version

### LINKED ISSUE #659: soup doctor never reports MLX: _get_mlx_info() has no caller, and get_mlx_version() always returns "unknown"
Found while setting up the box for #23. Two defects in the MLX diagnostic surface, on hardware where they are visible. Neither is in #23's scope, so filing rather than folding them in.

**Box:** Apple M1, 8 GB unified memory, arm64, macOS 26.6.2, Python 3.12.14, `mlx` 0.32.2, `mlx-lm` 0.31.3, Metal available.

## 1. `soup doctor` never reports MLX at all

`commands/doctor.py:189` defines `_get_mlx_info()` with the docstring *"Surface MLX info in the doctor report"*. **It has no caller.**

```
$ grep -rn "_get_mlx_info" src/ tests/ scripts/ docs/
src/soup_cli/commands/doctor.py:189:def _get_mlx_info() -> dict:
tests/test_mlx_backend.py:280:        from soup_cli.commands.doctor import _get_mlx_info
tests/test_mlx_backend.py:282:        info = _get_mlx_info()
```

The only references outside its own definition are in a test. Confirmed by AST rather than by grep alone:

```
_get_mlx_info defined: True
_get_mlx_info called : False
```

Consequence, measured on the box above — MLX installed, Metal available, Apple Silicon detected:

```
$ soup doctor | grep -ci mlx
0
```

The report shows `Backend: MPS (Apple Silicon)`, which comes from the **torch** MPS branch in `_check_gpu()`, not from MLX. A user on Apple Silicon has no way to ask Soup whether it can see their MLX install.

### Why it survived

`tests/test_mlx_backend.py:277` — under a comment reading `# doctor command reports MLX` and a class named `TestMLXDoctor`:

```python
def test_doctor_has_mlx_info(self):
    """`soup doctor` helpers surface MLX info (no crash on non-Apple)."""
    from soup_cli.commands.doctor import _get_mlx_info
    info = _get_mlx_info()
    assert isinstance(info, dict)
    assert "available" in info
```

It imports the helper and calls it directly, so it passes whether or not `doctor` invokes it. The test name and the comment both assert the command's behaviour; the body only exercises the helper.

This is the same shape as #363 — MLX code that exists, is tested, and is unreachable from the command that is supposed to use it.

## 2. `get_mlx_version()` can never return a version

`utils/mlx.py:33`:

```python
def get_mlx_version() -> Optional[str]:
    try:
        import mlx
    except ImportError:
        return None
    return getattr(mlx, "__version__", "unknown")
```

The top-level `mlx` package does not expose `__version__`. Measured:

```
mlx.__version__       -> <absent>
mlx.core.__version__  -> 0.32.2
importlib.metadata.version("mlx") -> 0.32.2
```

So on a correctly installed MLX the function returns the string `"unknown"`, always:

```
>>> get_mlx_info()
{'available': True, 'version': 'unknown', 'apple_silicon': True,
 'chip': {..., 'chip': 'Apple M1'}, 'unified_memory_bytes': 8589934592}
```

Everything else in that dict is correct — detection, chip, unified memory. Only the version is wrong, and it is wrong on every machine.

Defect 1 is why this was never noticed: nobody has ever seen the field rendered.

## Suggested fix shape

- Call `_get_mlx_info()` from `doctor` and render a panel when `available` is true (and, on Apple Silicon with MLX **absent**, say so — that is the case where a user most needs to be told).
- Read the version from `mlx.core.__version__`, falling back to `importlib.metadata.version("mlx")`, and keep `"unknown"` only as the genuine last resort.
- Change `test_doctor_has_mlx_info` to invoke the **command** and assert MLX appears in its output — otherwise this reopens the moment someone refactors. A control that the panel is absent without MLX would keep it honest.

## Acceptance

- [ ] `soup doctor` on Apple Silicon with MLX installed prints an MLX section including the real version
- [ ] `soup doctor` on a machine without MLX still runs and says so, without crashing
- [ ] `get_mlx_version()` returns `0.32.2`-style output where MLX is installed, `None` where it is not
- [ ] A test that drives the **command**, not the helper, plus a control that can fail

## Scope note

I can take this, but I have three PRs already open (#656, #657, #658) and would rather not add a fourth to your review queue before some of those land. Happy to open it whenever you prefer — or leave it for someone else, in which case note that reproducing it needs an Apple Silicon box with `mlx` installed. I will not bundle it into the #23 work.

ump45nose:
## What does this PR do?

Wires the long-dormant MLX section into `soup doctor` and fixes the version lookup that always reported `"unknown"`.

Fixes #659.

### Root cause

- `_get_mlx_info()` / `get_mlx_info()` existed in the codebase, but **nothing in `doctor()` ever called them** — the MLX panel was defined but unreachable, so Apple Silicon users never saw it.
- `get_mlx_version()` read `mlx.__version__`. The `mlx` package does **not** export `__version__` on its top-level module; it lives on `mlx.core` (and in distribution metadata), so the function always returned `"unknown"`.

### Changes

- `doctor()` now renders an MLX panel after the GPU check:
  - MLX present → version, chip, unified memory
  - Apple Silicon but MLX missing → "not installed" + install hint (`pip install 'soup-cli[mlx]'`)
  - non-Apple → "N/A (Apple Silicon only)"
  - the panel is informational and never crashes the report (info helper degrades to `available=False` on any error)
- `get_mlx_version()`:
  1. reads `mlx.core.__version__`
  2. falls back to `importlib.metadata.version("mlx")`
  3. returns `None` only when MLX truly cannot be found

### Tests

- `test_get_mlx_version_reads_core_version` — top-level `mlx` module with **no** `__version__` still yields the real version from `mlx.core`.
- `test_get_mlx_version_metadata_fallback` — metadata fallback path.
- `test_doctor_command_renders_mlx_panel` — drives the `doctor` command (not the helper) with a mocked MLX install; asserts the panel and version render.
- `test_doctor_command_mlx_absent_control` — drives `doctor` with MLX import failing; asserts "not installed" renders and the command still exits 0.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

- [x] `pytest tests/ -v` passes (ran `tests/test_mlx_backend.py` and `tests/test_doctor.py`: 43 passed; the full suite runs on CI)
- [x] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed — N/A, no doc change needed
- [x] Added a changelog fragment, or this change has no user-visible impact — added `changelog.d/0.73.3/660.fixed.md`

### Verification on real hardware

Apple Silicon (macOS arm64, Apple M1 Pro), MLX 0.32.2 installed:

```text
╭──────────────────────────────────── MLX ─────────────────────────────────────╮
│ Version:  0.32.2                                                             │
│ Chip:     Apple M1 Pro                                                       │
│ Memory:   32 GB unified                                                      │
╰──────────────────────────────────────────────────────────────────────────────╯
```

### AI assistance

This change was implemented with AI assistance; the investigation, code, tests, and this description were reviewed by a human contributor.

### DIFF
--- a/changelog.d/0.73.3/660.fixed.md
+++ b/changelog.d/0.73.3/660.fixed.md
@@ -1,0 +1,8 @@
+- **`soup doctor` now reports the MLX backend (#660 by @ump45nose in #660).**
+  The MLX panel had been written but never wired into the report, so Apple Silicon
+  users saw no MLX section at all. `doctor` now renders an MLX panel showing the
+  installed version, chip, and unified memory when MLX is present, says plainly
+  when it is missing on Apple Silicon, and marks it N/A elsewhere. `get_mlx_version()`
+  reads `mlx.core.__version__` (falling back to distribution metadata) instead of
+  `mlx.__version__`, which the `mlx` package does not export and which therefore
+  always reported "unknown".
--- a/changelog.d/0.73.3/660.fixed.md
+++ b/changelog.d/0.73.3/660.fixed.md
@@ -1,8 +1,8 @@
-- **`soup doctor` now reports the MLX backend (#660 by @ump45nose in #660).**
+- **`soup doctor` now reports the MLX backend (#659 by @ump45nose in #660).**
   The MLX panel had been written but never wired into the report, so Apple Silicon
   users saw no MLX section at all. `doctor` now renders an MLX panel showing the
   installed version, chip, and unified memory when MLX is present, says plainly
-  when it is missing on Apple Silicon, and marks it N/A elsewhere. `get_mlx_version()`
+  when it is missing on Apple Silicon, and stays quiet elsewhere. `get_mlx_version()`
   reads `mlx.core.__version__` (falling back to distribution metadata) instead of
   `mlx.__version__`, which the `mlx` package does not export and which therefore
   always reported "unknown".
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -91,6 +91,9 @@ def doctor(
     # GPU check
     _check_gpu()
 
+    # MLX (Apple Silicon) check
+    _check_mlx()
+
     # Resources check
     _check_resources(probe_disk=disk)
 
@@ -196,6 +199,43 @@ def _get_mlx_info() -> dict:
         return get_mlx_info()
     except Exception:  # noqa: BLE001
         return {"available": False}
+
+
+def _check_mlx():
+    """Report the MLX (Apple Silicon) backend in ``soup doctor``.
+
+    MLX is an Apple Silicon-only stack, so the panel is informational rather
+    than a pass/fail dependency: it shows the installed version and hardware
+    when present, and says so plainly when MLX is missing. It must never crash
+    the report (the info helper degrades to ``available=False`` on any error).
+    """
+    info = _get_mlx_info()
+    if info.get("available"):
+        mem_bytes = info.get("unified_memory_bytes")
+        mem_str = f"{mem_bytes / (1024**3):.0f} GB" if mem_bytes else "unknown"
+        chip = (info.get("chip") or {}).get("chip")
+        console.print(
+            Panel(
+                f"Version:  [bold green]{info.get('version', 'unknown')}[/]\n"
+                f"Chip:     [bold]{chip or 'Apple Silicon'}[/]\n"
+                f"Memory:   [bold]{mem_str}[/] unified",
+                title="MLX",
+            )
+        )
+    elif info.get("apple_silicon"):
+        console.print(
+            Panel(
+                "Status:   [yellow]not installed[/]\nInstall:  [dim]pip install 'soup-cli[mlx]'[/]",
+                title="MLX",
+            )
+        )
+    else:
+        console.print(
+            Panel(
+                "Status:   [dim]N/A (Apple Silicon only)[/]",
+                title="MLX",
+            )
+        )
 
 
 def _check_gpu():
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -216,7 +216,7 @@ def _check_mlx():
         chip = (info.get("chip") or {}).get("chip")
         console.print(
             Panel(
-                f"Version:  [bold green]{info.get('version', 'unknown')}[/]\n"
+                f"Version:  [bold green]{info.get('version') or 'unknown'}[/]\n"
                 f"Chip:     [bold]{chip or 'Apple Silicon'}[/]\n"
                 f"Memory:   [bold]{mem_str}[/] unified",
                 title="MLX",
@@ -227,13 +227,6 @@ def _check_mlx():
             Panel(
                 "Status:   [yellow]not installed[/]\n"
                 "Install:  [dim]pip install \"soup-cli\\[mlx]\"[/]",
-                title="MLX",
-            )
-        )
-    else:
-        console.print(
-            Panel(
-                "Status:   [dim]N/A (Apple Silicon only)[/]",
                 title="MLX",
             )
         )
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -225,7 +225,8 @@ def _check_mlx():
     elif info.get("apple_silicon"):
         console.print(
             Panel(
-                "Status:   [yellow]not installed[/]\nInstall:  [dim]pip install 'soup-cli[mlx]'[/]",
+                "Status:   [yellow]not installed[/]\n"
+                "Install:  [dim]pip install \"soup-cli\\[mlx]\"[/]",
                 title="MLX",
             )
         )
--- a/src/soup_cli/utils/mlx.py
+++ b/src/soup_cli/utils/mlx.py
@@ -31,12 +31,28 @@ def is_apple_silicon() -> bool:
 
 
 def get_mlx_version() -> Optional[str]:
-    """Return the MLX version string, or None if not installed."""
+    """Return the MLX version string, or None if not installed.
+
+    The top-level ``mlx`` package does not export ``__version__`` — it lives on
+    ``mlx.core`` (and in distribution metadata), so reading ``mlx.__version__``
+    always reported "unknown" (#659). Try the canonical surface first, fall
+    back to importlib.metadata, and only report "unknown" as a last resort.
+    """
     try:
-        import mlx
+        import mlx.core
+
+        version = getattr(mlx.core, "__version__", None)
+        if version:
+            return str(version)
     except ImportError:
+        pass
+
+    try:
+        from importlib.metadata import version as _dist_version
+
+        return _dist_version("mlx")
+    except Exception:  # noqa: BLE001 — no package, no metadata, or missing dist
         return None
-    return getattr(mlx, "__version__", "unknown")
 
 
 def get_chip_info() -> dict[str, str]:
--- a/tests/test_mlx_backend.py
+++ b/tests/test_mlx_backend.py
@@ -62,7 +62,7 @@ class TestMLXDetection:
         assert info["available"] is False
 
     def test_get_mlx_version_reads_core_version(self, monkeypatch):
-        """get_mlx_version reads mlx.core.__version__, not mlx.__version__ (#659)."""
+        """get_mlx_version prefers mlx.core over stale distribution metadata (#659)."""
         import sys
         import types
 
@@ -75,6 +75,11 @@ class TestMLXDetection:
 
         from soup_cli.utils import mlx as mlx_utils
 
+        # The module version is authoritative after an in-place upgrade, while
+        # distribution metadata can still report the prior installed release.
+        monkeypatch.setattr(
+            "importlib.metadata.version", lambda name: "0.31.0" if name == "mlx" else None
+        )
         assert mlx_utils.get_mlx_version() == "0.32.2"
 
     def test_get_mlx_version_metadata_fallback(self, monkeypatch):
@@ -311,14 +316,6 @@ output: ./output
 
 
 class TestMLXDoctor:
-    def test_doctor_has_mlx_info(self):
-        """`soup doctor` helpers surface MLX info (no crash on non-Apple)."""
-        from soup_cli.commands.doctor import _get_mlx_info
-
-        info = _get_mlx_info()
-        assert isinstance(info, dict)
-        assert "available" in info
-
     def test_doctor_command_renders_mlx_panel(self, monkeypatch):
         """`soup doctor` renders the MLX panel when MLX is available."""
         import sys
@@ -346,7 +343,7 @@ class TestMLXDoctor:
         assert "0.32.2" in result.output
 
     def test_doctor_command_mlx_absent_control(self, monkeypatch):
-        """`soup doctor` says MLX is not installed when it cannot be imported."""
+        """`soup doctor` renders the Apple-only MLX install guidance when absent."""
         import builtins
 
         from typer.testing import CliRunner
@@ -361,11 +358,30 @@ class TestMLXDoctor:
             return real_import(name, *args, **kwargs)
 
         monkeypatch.setattr(builtins, "__import__", fake_import)
+        # Force the Apple-specific branch; otherwise optional dependencies also
+        # render "not installed" and the assertion would be vacuous off-Mac.
+        monkeypatch.setattr("soup_cli.utils.mlx.is_apple_silicon", lambda: True)
         result = CliRunner().invoke(app, ["doctor"])
 
         assert result.exit_code == 0
         assert "MLX" in result.output
-        assert "not installed" in result.output
+        assert 'pip install "soup-cli[mlx]"' in result.output
+
+    def test_doctor_command_omits_mlx_panel_off_apple_silicon(self, monkeypatch):
+        """`soup doctor` stays quiet about MLX on non-Apple platforms."""
+        from typer.testing import CliRunner
+
+        from soup_cli.cli import app
+
+        # The doctor report should not dedicate a panel to an unsupported backend.
+        monkeypatch.setattr(
+            "soup_cli.commands.doctor._get_mlx_info",
+            lambda: {"available": False, "apple_silicon": False},
+        )
+        result = CliRunner().invoke(app, ["doctor"])
+
+        assert result.exit_code == 0, result.output
+        assert "Apple Silicon only" not in result.output
 
 
 if __name__ == "__main__":
--- a/tests/test_mlx_backend.py
+++ b/tests/test_mlx_backend.py
@@ -61,6 +61,41 @@ class TestMLXDetection:
         info = mlx_utils.get_mlx_info()
         assert info["available"] is False
 
+    def test_get_mlx_version_reads_core_version(self, monkeypatch):
+        """get_mlx_version reads mlx.core.__version__, not mlx.__version__ (#659)."""
+        import sys
+        import types
+
+        fake_mlx = types.ModuleType("mlx")  # no __version__ on the top-level pkg
+        fake_core = types.ModuleType("mlx.core")
+        fake_core.__version__ = "0.32.2"
+        fake_mlx.core = fake_core
+        monkeypatch.setitem(sys.modules, "mlx", fake_mlx)
+        monkeypatch.setitem(sys.modules, "mlx.core", fake_core)
+
+        from soup_cli.utils import mlx as mlx_utils
+
+        assert mlx_utils.get_mlx_version() == "0.32.2"
+
+    def test_get_mlx_version_metadata_fallback(self, monkeypatch):
+        """get_mlx_version falls back to importlib.metadata when core lacks it."""
+        import sys
+        import types
+
+        fake_mlx = types.ModuleType("mlx")
+        fake_core = types.ModuleType("mlx.core")
+        fake_core.metal = types.SimpleNamespace(is_available=lambda: True)
+        fake_mlx.core = fake_core
+        monkeypatch.setitem(sys.modules, "mlx", fake_mlx)
+        monkeypatch.setitem(sys.modules, "mlx.core", fake_core)
+
+        from soup_cli.utils import mlx as mlx_utils
+
+        monkeypatch.setattr(
+            "importlib.metadata.version", lambda name: "0.32.2" if name == "mlx" else None
+        )
+        assert mlx_utils.get_mlx_version() == "0.32.2"
+
     def test_estimate_mlx_batch_size_small_model(self):
         from soup_cli.utils.mlx import estimate_mlx_batch_size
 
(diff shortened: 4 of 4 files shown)


MakazhanAlpamys:
@ump45nose — reviewed on its merits, as promised on #659. **The core of this is right and I can prove it**: the wiring is real, and it is asserted by a test that would fail if you removed it. Three things to fix, one of them a test that cannot currently fail.

Before the details, the thing I owe you: **merging is still gated on the hold I described on #659**, not on anything in this diff. That is my process error, not yours, and it is being resolved on that thread with a deadline.

## What I verified by execution

Isolated worktree at `c1565b3`, `PYTHONPATH` pinned to the branch (unpinned, this repo's editable install resolves to the main tree and every mutation is a silent no-op).

| mutation | result |
|---|---|
| delete the `_check_mlx()` call from `doctor()` | **KILLED** — `test_doctor_command_renders_mlx_panel` + `test_doctor_command_mlx_absent_control` |
| `get_mlx_version()` → `return "unknown"` unconditionally | **KILLED** — 3 tests |
| break the `importlib.metadata` fallback | **KILLED** |

That first row is the whole point of #659 and you hit it. The old test imported the helper and called it directly, so it passed whether or not `doctor` invoked it; yours drives the command, so the wiring is pinned rather than assumed. Both defects are genuinely fixed.

## Blocking 1 — the absent-control cannot fail off-Mac

`test_doctor_command_mlx_absent_control` asserts `"not installed" in result.output`. Measured on this Windows box, real `soup doctor` output:

```
'not installed' count: 6
  | wandb        │ optional │ >=0.15.0  │ not installed
  | deepspeed    │ optional │ >=0.12.0  │ not installed
  | unsloth   / torchao / sglang / librosa   ... not installed
'N/A (Apple Silicon only)' present: True
```

All six come from the optional-dependency table. The MLX panel on a non-Apple runner never reaches the `not installed` branch at all — it renders `N/A (Apple Silicon only)`, because you do not patch `is_apple_silicon`. Proof rather than argument: changing the MLX panel's string to `MLXABSENTSENTINEL` leaves **43 passed**, nothing red.

So on ubuntu-latest and windows-latest that test asserts a string it would find anyway. It becomes real only on `macos-latest`, and by accident rather than design. Fix: monkeypatch `is_apple_silicon` → True so the branch is reached everywhere, and assert on something unique to that panel — `pip install 'soup-cli[mlx]'` appears nowhere else in the report.

This matters more than usual here because **a vacuous test is the exact defect #659 is about.** The issue exists because `test_doctor_has_mlx_info` asserted the command's behaviour in its name and the helper's in its body; shipping a new control with the same property would close the issue while reproducing its cause.

## Blocking 2 — the non-Apple panel contradicts the scope I set

`doctor.py:233-239` is an unconditional `else:`, so **every Linux and Windows user gets a permanent bordered MLX panel telling them MLX is irrelevant to them.** My triage said it must stay quiet off-Mac, and this is that noise. You documented the choice in the PR body rather than slipping it in, which is why I am calling it a scope disagreement and not a defect.

**My call: drop the `else:` branch.** `soup doctor` is read by every user on every platform and a dedicated panel is a heavy way to say "not applicable". If you want off-Mac visibility, the right home is a **row in the existing optional-dependency table** — where `wandb` and friends already say `not installed` without a panel each — not a box of its own. Either drop it or move it there; I will take both, not a third panel.

Credit where due: I checked the cost before objecting, and the branch is cheap — `get_chip_info()` short-circuits on `is_apple_silicon()` before spawning any `sysctl`, so this is about report noise, not startup time.

## Non-blocking, worth fixing in the same push

1. **The lookup order is untested.** Swapping to metadata-first passes the whole suite — 43 passed, nothing notices. If `mlx.core.__version__` must win over a possibly-stale `dist-info` (it should — the metadata can survive an upgrade the module has moved past), that ordering needs its own test.
2. **`test_doctor_has_mlx_info` is untouched.** It is the test #659 named, and it still stands beside your new ones with a name and enclosing comment claiming command coverage its body does not have. Delete it — the new tests supersede it — or fold it in.
3. **`info.get('version', 'unknown')` never falls back.** The key exists and holds `None`, so an installed-but-version-less MLX renders the literal `None` in the panel. And the `get_mlx_version` docstring still promises an `"unknown"` the code can no longer return.
4. **Changelog attribution**: `(#660 by @ump45nose in #660)` should be `(#659 by @ump45nose in #660)` — the convention is `(#ISSUE by @author in #PR)`, cf. `468.added.md`.

## Things I could not check, stated rather than glossed

This box is Windows with no Apple Silicon and no `mlx` installed, so every MLX-present path was exercised through injected fakes. **The real version output on hardware is yours to report** — that is the half of #659 nobody else here can verify, and it is why your access is genuinely useful. Please paste the actual `soup doctor` MLX panel from your machine in the PR.

Also verified clean: zero deleted or weakened test lines; `ruff` passes; `test_cli_startup_is_light.py` shows an identical failure set on your branch and on `main` (pre-existing local typer/click breakage); no new top-level `mlx`/`torch` import; and the ANSI-safety guard passes, including under `FORCE_COLOR=1` — your assertions target whole styled tokens rather than `--flag` strings, so the trap that has reddened this CI four times does not apply here.

ump45nose:
已推送维护者意见的修复：`bf0c025`。

已处理：
- 缺失控制测试显式 patch Apple Silicon，并断言唯一的 `pip install "soup-cli[mlx]"` 提示；
- 移除非 Apple 平台的独立 MLX 面板，并增加静默回归测试；
- 固化 `mlx.core.__version__` 优先于可能过期的 distribution metadata；
- 删除旧的仅调用 helper 的测试，`None` 版本回退为 `unknown`，并修正 changelog 为 #659。

验证：
- 隔离 Python 3.12 环境：`pytest --no-cov tests/test_mlx_backend.py -v` → 23 passed；`--no-cov` 仅避免单文件运行触发项目全局 77% 覆盖率门槛。
- `ruff check src/soup_cli/ scripts/ tests/` 与 `git diff --check` 通过。
- Apple M1 Pro 硬件实测（隔离环境的 MLX 0.32.2）：`soup doctor` 显示 `Version: 0.32.2`、`Chip: Apple M1 Pro`、`Memory: 32 GB unified`。完整训练依赖未安装，因此该临时环境同时如实报告缺少 torch 等训练栈。

该分支的修复、测试和说明由 AI 辅助完成。烦请复审。

MakazhanAlpamys:
Thanks — that is every item from the review, and two of them are the ones I cared about most.

`bf0c025` addresses, as far as I can see from the diff:

- the absent-control now patches Apple Silicon explicitly and asserts on `pip install "soup-cli[mlx]"`, which is unique to that branch — so it can actually fail on a Linux/Windows runner instead of matching the six `not installed` rows in the optional-dependency table;
- the non-Apple panel is gone, with a silence regression test, which is the scope point;
- `mlx.core.__version__` now takes priority over possibly-stale distribution metadata, and that ordering has its own test;
- the old helper-only test is deleted, `None` falls back to `unknown`, and the changelog fragment now credits `(#659 by @ump45nose in #660)`.

The Apple M1 Pro run is the half nobody else here could produce — `Version: 0.32.2`, `Chip: Apple M1 Pro`, `Memory: 32 GB unified` — thank you for posting it rather than asserting it worked. Noting for the record that the temporary environment honestly reported the missing training stack too; that is the right way to publish a partial environment.

Also noted, and it costs you nothing: the diff itself is entirely in English — code, tests and the changelog fragment — which is what matters, since those land in the repository. The commit message will not survive the squash merge (I write the subject and body), so no action needed there.

CI approved again — first-time-contributor runs need manual release here, so expect to ping me if a later push sits idle.

**Status, unchanged and not about this diff:** the hold on #659 runs until 2026-09-05 06:00 UTC or until @Srinivasan8888 replies, whichever comes first. Separately, I am mid-way through cutting v0.74.0, so my re-review of this lands after the tag. Nothing here is waiting on you.

MakazhanAlpamys:
@ump45nose — re-reviewed `bf0c025` by execution. **The hold is gone**: @Srinivasan8888 passed the issue to you explicitly on #659 at 09:04 UTC, so nothing outside this diff is blocking any more. Four of my five review items are genuinely fixed and I can show it. One of them is fixed in the code but **not actually pinned by the test you added for it**, and that is the one thing left.

## Verified by execution

Isolated worktree at `bf0c025`, `PYTHONPATH` pinned to the branch (unpinned, this repo's editable install resolves to the main tree and every mutation is a silent no-op). Baseline: `tests/test_mlx_backend.py` -> **23 passed**.

| # | mutation | result |
|---|---|---|
| M1 | delete the `_check_mlx()` call from `doctor()` | **KILLED** — `test_doctor_command_renders_mlx_panel`, `test_doctor_command_mlx_absent_control` |
| M2 | `get_mlx_version()` never reads `mlx.core.__version__` | **KILLED** — `test_get_mlx_version_reads_core_version`, `test_doctor_command_renders_mlx_panel` |
| M3 | break the `importlib.metadata` fallback | **KILLED** — `test_get_mlx_version_metadata_fallback` |
| M4 | change the absent-branch install hint string | **KILLED** — `test_doctor_command_mlx_absent_control` |
| M6 | delete the `mlx.core` branch so metadata wins the priority | **KILLED** — 2 tests |
| **M5** | **`elif info.get("apple_silicon"):` -> `else:`** | **SURVIVED — 23 passed** |

M4 is the one I asked for and it is now real: the absent-control patches `is_apple_silicon` and asserts a string unique to that panel, so it can fail on a Linux or Windows runner instead of matching the six `not installed` rows in the optional-dependency table. M6 shows the priority ordering is pinned, not incidental.

## Blocking — the silence test cannot fail

`tests/test_mlx_backend.py:384`:

```python
assert "Apple Silicon only" not in result.output
```

That string does not exist anywhere in `src/` any more — the panel that used to emit it is the panel this PR removed:

```
$ grep -rn "Apple Silicon only" src/
(no matches)
```

So the assertion asks whether the output contains a string the implementation cannot produce under **any** branch. M5 is the proof rather than the argument: turning `elif info.get("apple_silicon"):` into `else:` makes `soup doctor` render the `not installed` / `pip install "soup-cli[mlx]"` panel on every non-Apple platform — precisely the scope regression the test is named after — and the suite stays at **23 passed**.

This is worth spending a round on because it is the *same defect class as #659 itself*. The old `test_doctor_has_mlx_info` asserted the helper while its name and comment claimed the command; this one asserts a dead string while its name claims the branch. Both pass on a codebase where the thing they describe is broken.

### The fix, and proof that it works

One line. `"MLX"` appears nowhere else in the report on a non-Apple box — I checked the real rendered output with `_get_mlx_info` stubbed to `{"available": False, "apple_silicon": False}`: both `MLX` and the install hint are absent.

```diff
-        assert "Apple Silicon only" not in result.output
+        assert "MLX" not in result.output
```

Run against three trees:

| tree | strengthened assertion |
|---|---|
| your code, unmutated | **3 passed** |
| your code + M5 | **FAILED** `test_doctor_command_omits_mlx_panel_off_apple_silicon` |
| your code, restored | **23 passed** |

## Also needed — mechanical, and it is my fault not yours

`changelog.d/0.73.3/660.fixed.md` has to move to `changelog.d/0.74.0/660.fixed.md`. v0.74.0 was tagged while this PR was open, `scripts/assemble_changelog.py` consumed the `0.73.3/` directory, and `test_repository_fragments_match_the_newest_release` fails in every matrix cell until each open fragment is rebased. Nothing about the fragment's *content* is wrong — the `(#659 by @ump45nose in #660)` credit is exactly the form I want.

So: `git rebase origin/main`, then `git mv changelog.d/0.73.3/660.fixed.md changelog.d/0.74.0/660.fixed.md`.

If you would rather not spend another round on a one-line test change plus a file move, say the word and I will push both to your branch — `maintainerCanModify` is on and this is mostly my release timing. I am asking rather than doing it because the vacuous-assertion fix is worth being yours, given it is the exact thing the issue is about.

Everything else is settled. The Apple M1 Pro run — `Version: 0.32.2`, `Chip: Apple M1 Pro`, `Memory: 32 GB unified` — is still the half nobody else here can produce, and the version branch is now pinned by two tests rather than trusted.

MakazhanAlpamys:
@ump45nose — following up, and putting a time on it so it does not just go quiet.

Nothing new is being asked of you. Two mechanical items remain from my 2026-09-04 review, and one of them is my release timing rather than your work:

1. `tests/test_mlx_backend.py:384` — `assert "Apple Silicon only" not in result.output` asserts a string that no longer exists in `src/`, so it passes under the M5 mutation that re-breaks the very branch the test is named after. One line: `assert "MLX" not in result.output`.
2. `changelog.d/0.73.3/660.fixed.md` -> `changelog.d/0.74.0/660.fixed.md`. v0.74.0 was tagged while this PR was open. The content and the `(#659 by @ump45nose in #660)` credit are already right.

**If I have not heard from you by 2026-09-08 18:00 UTC, I will push both to your branch myself** (`maintainerCanModify` is on) and merge, with the work credited to you exactly as it stands. I would rather the test change be yours — it is the same defect class the issue is about — but I am not going to let a one-line edit hold up a fix that is otherwise done and measured on hardware nobody else here has.

Reply with anything, including "go ahead", and that is settled.

MakazhanAlpamys:
One factual addendum, because your PR shows a green checkmark that is not true any more and I do not want you merging on the strength of it.

**Your CI ran on 2026-09-04 at 07:49-08:07 UTC. v0.74.0 was tagged at 17:13 UTC the same day**, nine hours later. So that green run happened while `changelog.d/0.73.3/` was still the correct directory. Re-running it today goes red, and I confirmed the failure by execution rather than quoting the rule at you — this is the exact error, reproduced locally by dropping a probe fragment into that directory on current `main`:

```
scripts.assemble_changelog.ChangelogError:
  changelog.d\0.73.3\999.fixed.md: fragment baseline is 0.73.3, but the latest
  release is 0.74.0; rebase and move the fragment before merging

```

So `git rebase origin/main` is required regardless of anything else, and `git mv changelog.d/0.73.3/660.fixed.md changelog.d/0.74.0/660.fixed.md` comes with it. Nothing about the fragment's content is wrong — the `(#659 by @ump45nose)` credit is exactly the form I want.

The 2026-09-08 18:00 UTC offer above stands unchanged: say nothing and I push both the one-line test fix and the file move myself, with the work credited to you.

MakazhanAlpamys:
@ump45nose — checking in, and putting a date on it this time so it does not just go quiet again.

Nothing new is being asked. Two mechanical items remain from my 2026-09-04 review:

1. `tests/test_mlx_backend.py:384` asserts `"Apple Silicon only" not in result.output` — a string that no longer exists in `src/`, so it passes under the mutation it was written to catch.
2. The changelog fragment now belongs in `changelog.d/0.74.0/` and is named after the **PR** (`660.fixed.md`), not the issue. That is release timing — v0.74.0 was tagged nine hours after your CI ran — and not something you did.

Also worth repeating because your PR still shows a green checkmark that is no longer true: that run happened while `changelog.d/0.73.3/` was still the correct directory. A re-run today goes red, and I confirmed the failure myself. Do not merge on the strength of that tick.

**If I have not heard from you by 2026-09-12 18:00 UTC I will finish both items and merge, credited to you** — your commits, your `CONTRIBUTORS.md` entry, your name in the fragment. Say the word at any point and it goes back to you.

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:
@ump45nose — I said on **2026-09-06** that if I had not heard from you by 2026-09-08 18:00 UTC I would push the two remaining items myself, and then said it again on **2026-09-09** with a 2026-09-12 18:00 deadline. Both dates passed and I did nothing. I have now done it: `9cc30de1`, pushed to your branch on top of `bf0c0251` as a fast-forward, so none of your commits were rewritten.

**Two items, both mechanical, neither asking anything of you:**

1. **`tests/test_mlx_backend.py` asserted `"Apple Silicon only" not in result.output`.** That string exists nowhere under `src/`, so the assertion could never fail — a test that reads like a guarantee and provides none. Replaced with the MLX install hint and the panel title, both of which the sibling test asserts **are** present when `apple_silicon` is True, so the pair now brackets the behaviour from both sides. Verified discriminating rather than assumed:

```
force the panel to render off Apple Silicon  (elif True: at doctor.py:415)
-> FAILED TestMLXDoctor::test_doctor_command_omits_mlx_panel_off_apple_silicon
   1 failed, 28 passed
```

The old assertion stayed green under that same mutation.

2. **The changelog fragment was in `changelog.d/0.73.3/`** — two releases stale after v0.74.0 and v0.75.0 each assembled and removed their directories. It would have failed the repository fragment guard in every matrix cell. Moved to `0.75.0/`, body unchanged.

**Your attribution is untouched.** The MLX wiring, the `_get_mlx_info` helper and every other test in this PR are yours; my commit is separate and says so. `CONTRIBUTORS.md` will credit you, not me, when this lands.

Nothing else is outstanding. I will merge as soon as CI is green — the Actions queue is heavily backed up right now, so it may be a while.

**One thing worth knowing about what happened here.** This PR has been open since 2026-09-04 for the sake of two lines. Your four substantive review items were all fixed by the 4th; what kept it open was my own follow-through, twice announced and twice missed. If you are still around, @Srinivasan8888 has offered on #659 to run this on a real Apple Silicon box and post what the panel actually prints — I have accepted, so the verification neither of us could do is coming. If you have moved on, that is entirely fair and the work still lands under your name.

MakazhanAlpamys:
Merged as `53ac6b06`. Thank you — and I owe you the honest version of why this took ten days for two lines.

**The work was yours and it was finished on the 4th.** Four substantive review items, all answered, all correct. What kept this open afterwards was my follow-through: I announced a takeover date twice, on 2026-09-06 and 2026-09-09, and let both pass without doing anything. That is not a queue problem or a CI problem, it is mine, and it is the part of this thread worth recording.

**What the PR actually fixed, both halves:**

- `doctor()` never called `_get_mlx_info()` or `get_mlx_info()`, so the MLX panel was defined and unreachable. The backend Soup supports for training did not appear in Soup's own diagnostic.
- `get_mlx_version()` read `mlx.__version__`, which the package does not export on its top-level module. A correctly installed MLX reported `unknown` — which reads as a broken install rather than a broken lookup, and is the worse of the two failure modes.

**What I finished on top, as a fast-forward on `bf0c0251` so none of your commits were rewritten (`9cc30de1`):**

1. `tests/test_mlx_backend.py` asserted `"Apple Silicon only" not in result.output`. That string exists nowhere under `src/`, so the assertion could never fail — a test that reads like a guarantee and provides none. Replaced with the MLX install hint and the panel title, both of which the sibling test asserts **are** present when `apple_silicon` is True, so the pair brackets the behaviour from both sides. Verified discriminating rather than assumed: forcing the panel to render off Apple Silicon fails the new assertion, and the old one stayed green under the same mutation.
2. The changelog fragment was in `changelog.d/0.73.3/`, two releases stale, and would have failed the repository fragment guard in every matrix cell. Moved to `0.75.0/`, body unchanged.

Final required matrix on `9cc30de1`: **14/14 green**, `merge-tree` clean against current `main` despite #911 having just touched the same file.

Your entry is in `CONTRIBUTORS.md` at `e2c13ffc`, and the merge commit names which commits were whose. @Srinivasan8888 has offered on #659 to run this on real Apple Silicon and post what the panel prints — that verification is still coming, and neither of us could do it from here.

Issue #661, opened 2026-09-04
Two of the three MLX recipes name Hugging Face repos that do not exist (qwen3-8b-sft-mlx, gemma3-9b-sft-mlx)


Srinivasan8888:
Found while picking a fixture for #23. **Two of the three shipped MLX recipes name a Hugging Face repo that does not exist**, so `soup recipes use <name>` followed by `soup train` cannot work for anyone.

| recipe | declares | state |
|---|---|---|
| `llama3.1-8b-sft-mlx` | `mlx-community/Llama-3.1-8B-Instruct-4bit` | **exists** — fine |
| `qwen3-8b-sft-mlx` | `mlx-community/Qwen3-8B-Instruct-4bit` | **does not exist** |
| `gemma3-9b-sft-mlx` | `mlx-community/gemma-3-9b-it-4bit` | **does not exist** |

## Evidence

I got this wrong once on the way here, so the reasoning is shown rather than the conclusion asserted. A bare `401` from the Hub API proves nothing — a repo I invented returns the same:

```
mlx-community/definitely-does-not-exist-xyz   ->  401 {"error":"Invalid username or password."}
```

The authoritative check is `huggingface_hub`, unauthenticated:

```
mlx-community/Qwen3-8B-Instruct-4bit      RepositoryNotFoundError
mlx-community/Qwen3-8B-4bit               EXISTS  gated=False  private=False
mlx-community/gemma-3-9b-it-4bit          RepositoryNotFoundError
mlx-community/gemma-3-12b-it-4bit         EXISTS  gated=False  private=False
mlx-community/Llama-3.1-8B-Instruct-4bit  EXISTS  gated=False  private=False
```

**It is not an authentication problem**, and that is the control that matters: sibling repos in the same org at the same quantisation resolve fine from the same unauthenticated client, and even genuinely gated upstream repos answer 200:

```
meta-llama/Llama-3.1-8B-Instruct   200      <- gated upstream, still resolves
google/gemma-3-12b-it              200
mlx-community/gemma-3-27b-it-4bit  200
```

## Why each id is wrong

**`Qwen3-8B-Instruct-4bit`** — Qwen3 dropped the `-Instruct` suffix; the instruct model is just `Qwen3-8B`. The recipe carries a Qwen2.5-era naming convention forward. The correct id is `mlx-community/Qwen3-8B-4bit`, which exists.

[...]

PR #656, opened 2026-09-03
feat(recipes): add glm-5.1-grpo recipe (#275)

Srinivasan8888:
Adds `glm-5.1-grpo`, the reasoning task-variant for `zai-org/GLM-5.1`. Catalog 162 -> 163.

`Refs #275`. [Claimed](https://github.com/MakazhanAlpamys/Soup/issues/275#issuecomment-5529106580) before any code was written.

## Overlap and attribution

No claim comment existed for this recipe. `gh pr list --search glm --state all` returns #452 (`glm-5.1-dpo`, merged, by @Osheun) and nothing else touching GLM; no open PR touches `recipes/`, `catalog.py`, or GRPO. All six named children of #275 are closed — #276 (#615), #277 (#448), #278 (#422), #279, #280 (#452), #281. I checked PR **titles** rather than only issue threads, since #512 is the reason that rule exists.

**The shape is inherited, not invented.** The geometry is @Osheun's from #452; the MoE-GRPO task fields are from `qwen3-30b-a3b-reasoning-grpo`.

## Shape, and why

GLM-5.1 had SFT (v0.71.24) and DPO (#280) but no reasoning variant. Measured rather than picked by eye — the models at two-of-three:

```
zai-org/GLM-5.1                  ['dpo', 'sft']         missing=['grpo']
Qwen/Qwen3.5-35B-A3B             ['dpo', 'sft']         missing=['grpo']
deepseek-ai/DeepSeek-V4-Flash    ['grpo', 'sft']        missing=['dpo']
moonshotai/Kimi-K2.6             ['grpo', 'sft']        missing=['dpo']
Qwen/Qwen3.5-9B                  ['dpo','grpo','sft']   complete (#632)
```

GLM-5.1 is the one whose missing slot is *reasoning* rather than preference.

### Where each line came from

The two conventions **conflict** — the MoE-GRPO cohort uses LoRA r16/a32, the GLM-5.1 siblings use r32/a64. I followed #275's own rule ("the shipped SFT recipes are the copy-paste template"), with the GRPO templates supplying the task shape:

| line | source | note |
|---|---|---|
| `base`, `size: 754B`, tags | `glm-5.1-sft` / `glm-5.1-dpo` | same base model |
| `batch_size: 1`, `gradient_accumulation_steps: 16` | GLM-5.1 siblings | 754B geometry |
| **`lora r32/a64`** | GLM-5.1 siblings | **not** the 30B template's r16/a32 — both siblings use r32/a64 here |
[...]

PR #569, opened 2026-08-26
feat(recipes): add DeepSeek-R1-Distill Llama SFT + R1-Distill DPO recipes

Nick-800:
## What this does

Finishes the remaining actionable scope of #523 (follow-up to #536): adds **4 ready-made recipes** to `soup recipes`, growing the catalog from **154 to 158**:

| Key | Base | Task | Notes |
|---|---|---|---|
| `deepseek-r1-distill-llama-8b-sft` | `deepseek-ai/DeepSeek-R1-Distill-Llama-8B` | sft | LoRA r16/a32, lr 2e-4, 4096 ctx |
| `deepseek-r1-distill-qwen-1.5b-dpo` | `deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B` | dpo | r16/a32, lr 5e-6, beta 0.1 |
| `deepseek-r1-distill-qwen-7b-dpo` | `deepseek-ai/DeepSeek-R1-Distill-Qwen-7B` | dpo | r16/a32, lr 5e-6, beta 0.1 |
| `deepseek-r1-distill-llama-8b-dpo` | `deepseek-ai/DeepSeek-R1-Distill-Llama-8B` | dpo | r16/a32, lr 5e-6, beta 0.1 |

The DPO variants follow the dense-model DPO house shape (`qwen2.5-7b-dpo`) with one deliberate deviation: **`max_length: 4096` instead of 2048**, because R1-Distill preference pairs carry `<think>` reasoning traces that routinely exceed 2k tokens.

## Why `ministral-3b-sft` is not here

Proposing it as wontfix/blocked rather than shipping an untrainable recipe: `mistralai/Ministral-3B-Instruct-2410` still has no public weights (API-only, per Hub discussions), and the successor `Ministral-3-3B-Instruct-2512` ships FP8/multimodal checkpoints (`Mistral3ForConditionalGeneration`, vision arch) whose text-only LoRA loading is unverified in Soup. A BF16 checkpoint exists (`-2512-BF16`); if it proves to load cleanly as a causal-LM I'm happy to add it in a follow-up. Happy to discuss.

## Testing

TDD order preserved: `TestR1DistillLlamaAndDpoRecipes` (5 tests, incl. pinned `lr`, `dpo_beta == 0.1`, and `data.format == "dpo"` for the DPO variants) + four milestone asserts bumped to 158 landed first and were proven red against the 154 catalog.

[...]

PR #662, opened 2026-09-04
feat(recipes): add deepseek-v4-flash-dpo recipe (#275)

Srinivasan8888:
Adds `deepseek-v4-flash-dpo`, the preference variant for `deepseek-ai/DeepSeek-V4-Flash`. Catalog 163 -> 164.

`Refs #275`. [Claimed](https://github.com/MakazhanAlpamys/Soup/issues/275#issuecomment-5529738154) before any code was written, including the deviation below — so it is reviewable rather than discovered.

## Overlap

`gh pr list --search deepseek --state all` returns #432 (`deepseek-v4-flash-grpo`, merged, by @Faisal01011), #569 and #536 (@Nick-800, R1-Distill families). None touches a V4-Flash preference variant. #279, the named child for this base's GRPO variant, is closed. Checked PR **titles**, not just issue threads.

## Shape, and why

DeepSeek-V4-Flash shipped SFT (v0.71.24) and GRPO (#279) but no preference variant. Of the bases still at two-of-three it is the **efficiency-tier** model, so it is the one a contributor without a multi-node cluster is most likely to reach.

The DPO convention here is unusually tight. Measured across the catalog rather than assumed:

```
lr across all 12 DPO recipes:  {'5e-6': 12}    <- unanimous
dpo_beta across all 12:        {0.1: 12}       <- unanimous
epochs across all 12:          {3: 11, 1: 1}   <- the 1 is glm-5.1-dpo, at 754B
```

So `lr: 5e-6`, `dpo_beta: 0.1`, `epochs: 3` — and both siblings independently agree on `epochs: 3`. LoRA r16/a32 matches both siblings and every non-GLM DPO recipe.

### The deliberate deviation

`batch_size: auto` and `gradient_accumulation_steps: 8` come from the **SFT** sibling, not the GRPO sibling's `1` / `16`.

GRPO's batch shape is driven by rollout generation, which DPO does not do, so inheriting it would be copying a constraint that does not apply. `qwen3.5-35b-a3b-dpo` — the closest MoE-DPO analogue in the catalog — uses `auto`/`8`. The same reasoning omits `gradient_checkpointing`, which the SFT sibling and that analogue both leave at default while the GRPO sibling pins it.

No `modality:` line: DeepSeek has no family contract like the Qwen3.5 one, and neither sibling declares it.

## Mutation table

Applied only inside the `deepseek-v4-flash-dpo` entry (sliced by dict key, so `    r: 16` cannot reach another recipe). `__pycache__` cleared each round; `catalog.py` md5-verified byte-identical after every round.

**Run twice — before the rebase and again on the rebased tree — with identical results**, so the table describes the tree this actually merges into:

| # | mutation | verdict | named test |
|---|---|---|---|
| 1 | `RecipeMeta.model` **alone** | KILLED | `test_recipe_meta_pins_the_model_id` |
| 2 | YAML `base:` **alone** | KILLED | `test_yaml_base_pins_the_model_id` |
| 3 | both model surfaces together | KILLED | `test_recipe_meta_pins_the_model_id` |
| 4 | `RecipeMeta.task` alone | KILLED | `TestRecipeCatalog::test_recipe_tasks_match_yaml` |
| 5 | YAML `task:` alone | KILLED | `TestRecipeCatalog::test_recipe_tasks_match_yaml` |
| 6 | both task surfaces together | KILLED | `test_recipe_meta_pins_the_model_id` |
| 7 | `data.format` dpo -> auto | KILLED | `test_recipe_loads_with_expected_dpo_moe_shape` |
| 8 | `lora.r` 16 -> 64 | KILLED | `test_recipe_loads_with_expected_dpo_moe_shape` |
| 9 | `lora.alpha` 32 -> 16 | KILLED | `test_recipe_loads_with_expected_dpo_moe_shape` |
| 10 | `moe_lora` true -> false | KILLED | `test_recipe_loads_with_expected_dpo_moe_shape` |
| 11 | `gradient_accumulation_steps` 8 -> 4 | KILLED | `test_recipe_loads_with_expected_dpo_moe_shape` |
| 12 | `lr` 5e-6 -> 2e-5 | KILLED | `test_recipe_loads_with_expected_dpo_moe_shape` |
| 13 | `max_length` 4096 -> 2048 | KILLED | `test_recipe_loads_with_expected_dpo_moe_shape` |
| 14 | **`size` N/A -> 7B** | KILLED | `test_v4_flash_task_variants_are_three_distinct_entries` |
| 15 | delete the whole entry | KILLED | `test_recipe_meta_pins_the_model_id` |

**Row 14 is the #656 lesson carried forward.** There, `size 754B -> 30B` **survived the entire suite** because `RecipeMeta.size` had no coverage anywhere. That test now ties `size` across the three V4-Flash variants, so the hole never reopened here — it was killed on the first pass rather than found and patched.

### Reported survivors — equivalent, and proved

Six lines equal their schema default, so stripping them is behaviourally identical and **cannot** be killed:

```
stripped line                    identical loaded config?   schema default
  epochs: 3                                          True   epochs = 3
  batch_size: auto                                   True   batch_size = 'auto'
  quantization: 4bit                                 True   quantization = '4bit'
  dpo_beta: 0.1                                      True   dpo_beta = 0.1
  moe_aux_loss_coeff: 0.01                           True   moe_aux_loss_coeff = 0.01
  target_modules: auto                               True   target_modules = 'auto'
```

Each verified by loading the recipe with and without the line and comparing `model_dump()`. Independent corroboration from the repo's own tooling: #640's delta encoder omits **exactly these six** from the recipe's snapshot delta.

They stay in the YAML for parity with the sibling recipes, which also spell them out. I did not add a test pinning YAML *source text* to kill them — that is the shape #621/#640 removed from this suite.

## Verification

```
$ .venv/bin/python -m pytest tests/test_recipes.py tests/test_v07124.py tests/test_v07130.py \
    tests/test_v07132.py tests/test_recipe_count_is_synced.py tests/test_issue621_recipe_snapshot.py -q -o addopts=
643 passed, 1 skipped

All checks passed!

8
```

**Those 8 removed test lines, each accounted for:** 3 x `def test_catalog_size_is_163` renamed to `_164`, 4 x `assert len(RECIPES) == 163` bumped to `164`, and 1 docstring in `test_v07124.py` naming that method as a string. All 1:1 replacements; net **+6 test methods**, nothing deleted or weakened.

### Run live

```
$ soup recipes search deepseek-v4
│ deepseek-v4-flash-…  │ deepseek-ai/DeepS… │ sft  │ N/A │ DeepSeek V4 Flash MoE SFT   │
│ deepseek-v4-flash-…  │ deepseek-ai/DeepS… │ grpo │ N/A │ DeepSeek V4 Flash MoE GRPO  │
│ deepseek-v4-flash-…  │ deepseek-ai/DeepS… │ dpo  │ N/A │ DeepSeek V4 Flash MoE DPO   │
│ deepseek-v4-pro-sft  │ deepseek-ai/DeepS… │ sft  │ N/A │ DeepSeek V4 Pro flagship    │
```

`soup recipes use deepseek-v4-flash-dpo --yes` writes a `soup.yaml` that reloads through `load_config_from_string` -> `deepseek-ai/DeepSeek-V4-Flash | dpo | lora r16 | lr 5e-06`.

### Count sweep — 13 sites

catalog comment · 3 renamed `test_catalog_size_is_164` methods · 4 assertions · 1 docstring cross-reference · `CONTRIBUTING.md` · `docs/commands.md` · `docs/serving-and-export.md` (x2). Snapshot fixture regenerated: **12 insertions, 0 deletions**.

## Scope limits

- **Not trained.** DeepSeek-V4-Flash is MoE. No hyperparameter here is a measured recommendation — every value is inherited from a shipped sibling or the DPO convention, itemised above.
- **Repo id verified as *resolving* only** — `deepseek-ai/DeepSeek-V4-Flash` returns HTTP 200. I did not download the weights, so "trains under this config" is unverified. (Worth noting I checked this properly: see #661, where I found two MLX recipes naming repos that do **not** exist, and where a bare 401 turned out to prove nothing.)

Changelog fragment follows in a second push once this has a number.


Srinivasan8888:

```
1 failed, 19863 passed
```

MakazhanAlpamys:
Merged as `2838e81`, and credited in `CONTRIBUTORS.md` (`734e6b9`).

## What I ran before merging

Isolated worktree, `PYTHONPATH` pinned to the branch — unpinned, this repo's editable install resolves to the main tree and every mutation is a silent no-op.

| mutation | result |
|---|---|
| `RecipeMeta.model` -> `nonexistent/does-not-exist-xyz` | **KILLED** — `test_recipe_meta_pins_the_model_id` |
| YAML `base:` -> garbage | **KILLED** — `test_yaml_base_pins_the_model_id` |
| both model surfaces at once | **KILLED** |
| delete the whole entry | **KILLED** |
| `task` dpo -> sft, one surface | **KILLED** — `TestRecipeCatalog::test_recipe_tasks_match_yaml` |
| `task` dpo -> sft, both surfaces | **KILLED** |
| `size` `N/A` -> `7B` | **KILLED** — `test_v4_flash_task_variants_are_three_distinct_entries` |

Seven for seven. `catalog.py` md5 verified identical before and after the whole matrix, so no mutation leaked into the merged tree.

Tests: **465 passed** across the four touched files, **1962 passed** across the full recipe-validation surface (`test_recipes_v031`, `test_recipes`, `test_recipe_count_is_synced`, `test_issue621_recipe_snapshot`, `test_v07124/30/32`), ruff clean. Eight removed test lines, every one inspected — all `163` -> `164` count renames, no assertion loosened.

`deepseek-ai/DeepSeek-V4-Flash` confirmed on the Hub unauthenticated, with a negative control on the same call, because #661 established that a bare 401 proves nothing. End to end: `recipes show` renders, `recipes use --yes` writes a `soup.yaml` in a temp dir, and the result loads through the real schema as `task=dpo` / `data.format=dpo` / `dpo_beta=0.1`.

## What was good, specifically

**The convention claims were checkable and they checked out.** You did not say "matches the other DPO recipes" and leave me to trust it — you gave the counts, and re-deriving them against all twelve gave exactly `12/12`, `12/12`, `11/12`. The `epochs` outlier being `glm-5.1-dpo` is right too. That turns a review from an argument into a verification, which is much faster for both of us.

**Pinning the model id on both surfaces independently is the part that earns its place.** The snapshot test only ever sees the YAML-derived config, so a wrong `RecipeMeta.model` slips past it entirely — the first mutation in that table is caught by *your* new test and nothing else. Given #661 was precisely a shipped recipe naming a repo that does not exist, that is the difference between adding a row and adding a row that stays correct.

**And you flagged the cherry-pick rather than letting me find it.** Carrying #669's fix onto this branch to get your own checks green was the right call at the time, saying so up front was the better one, and force-pushing it out once #669 merged meant I had nothing to unwind. That is three separate good decisions on one point of process.

## One nit I did not hold the merge for

The fragment credits the prior GRPO variant as `(#279 by @Faisal01011)`. #279 is the issue; the pull request was #432. Author is right, number is the wrong kind. Not worth a round trip — I will fold it in if I touch that entry.

Separately, for the record on something that was my fault and not yours: the v0.74.0 tag landed mid-queue, which is why several open PRs suddenly needed their changelog fragment moved. Yours was already in `0.74.0` because you had rebased. That saved a round.

`#275` stays open — it is the tracking issue for the whole 2026-family task-variant sweep, and this closes one row of it.

Issue #23, opened 2026-04-13
MLXSFTTrainerWrapper: add end-to-end training-loop integration test on Apple Silicon


MakazhanAlpamys:
## Problem

v0.25.0 shipped the MLX backend (`backend: mlx`) with a full SFT trainer wrapper (`MLXSFTTrainerWrapper`) that delegates to `mlx_lm.tuner.trainer.train()`. However, the **training loop itself is not integration-tested** against a real MLX runtime.

Current test coverage in `tests/test_mlx_backend.py`:

- ✅ `detect_mlx()` — True/False, with and without the `mlx` package
- ✅ `get_mlx_info()` — returns dict with `available` key
- ✅ `estimate_mlx_batch_size()` — size-vs-memory heuristic
- ✅ `MLXSFTTrainerWrapper.__init__` — config attached, model=None
- ✅ `MLXSFTTrainerWrapper.setup(...)` when `mlx` is missing → raises `ImportError`
- ✅ `mlx_routing.MLX_TRAINER_REGISTRY` lookup + lazy imports
- ✅ `SoupConfig(backend="mlx", task="sft")` accepted
- ✅ `SoupConfig(backend="mlx", task in {"dpo","grpo"})` rejected at load time
- ✅ All 3 MLX recipes validate as valid `SoupConfig`

**Not covered**:

- ❌ `MLXSFTTrainerWrapper.setup(...)` with a real MLX runtime — actually loading a model via `mlx_lm.load()`
- ❌ `MLXSFTTrainerWrapper.train(...)` — end-to-end training loop against a tiny MLX model
- ❌ Adapter save-path verification (`adapters.safetensors` in `output_dir`)
- ❌ Rich callback bridge integration (currently passes `training_callback=None`)

## Why not in v0.25.0

**CI runs on Ubuntu / Windows / macOS x86_64**. MLX requires Apple Silicon (`arm64-darwin`). GitHub Actions does offer `macos-14` and `macos-latest` runners with Apple Silicon, but they are slower and cost extra minutes. More importantly, running a real MLX fine-tune even on the smallest MLX model takes minutes, not seconds — unsuitable for every PR's CI.

## Proposed work

1. **Separate workflow**: add `.github/workflows/mlx-integration.yml` triggered on a schedule (nightly) and manual `workflow_dispatch`, running on `macos-14`. Not blocking on PRs.
[...]

Issue #651, opened 2026-09-03
The smoke path only works at torch>=2.6 while [train] declares torch>=2.5.0 — the declared floor is never exercised end to end


MakazhanAlpamys:
`pyproject.toml` declares `torch>=2.5.0` for the `[train]` extra (raised in #641), and the `transformers-floor` CI job pins `torch==2.5.1`. But the `pytorch-smoke` job added in #601 has to install `"torch>=2.6"` on top of `.[dev]`, because the training chain it exercises **does not work at the declared floor**.

## The mechanism

`sshleifer/tiny-gpt2` — the fixture the smoke tests use — ships only `pytorch_model.bin` and no safetensors. Since CVE-2025-32434, transformers refuses to `torch.load` a `.bin` checkpoint unless torch is at least 2.6:

```
ValueError: Due to a serious vulnerability issue in `torch.load` ... require users to
upgrade torch to at least v2.6 ... (CVE-2025-32434)
```

Reproduced locally on torch 2.5.1 + transformers 4.57.6: `test_sft_smoke` and `test_dpo_smoke` both fail at model load, before any training step.

## Why this matters beyond the job

Two separate consequences, and the second is the one that costs someone an afternoon:

1. **The declared floor is not exercised end to end.** `torch>=2.5.0` is what we tell users to install and what `soup doctor` recommends, but no CI job runs a real SFT or DPO training step at it. #641 established that a declared floor which cannot bind is a fiction; this is the adjacent problem — a floor that binds but is untested on the path that matters most.
2. **A contributor on torch 2.5 meets a CVE-named error with no explanation.** They install exactly what `pyproject.toml` asks for, run the smoke tests locally, and get a message about a `torch.load` vulnerability that says nothing about tiny-gpt2, safetensors, or what to do. The CI job silently papers over it with its own pin.

## Options, roughly in order of preference

1. **Change the fixture to a safetensors-only tiny model.** Removes the constraint at source: the smoke path would then run at the declared floor, and the `"torch>=2.6"` pin in `ci.yml` could go. Needs a fixture that is genuinely tiny and safetensors-native.
2. **Raise the declared floor to `torch>=2.6`** and drop the pin. Honest, but it is a real compatibility narrowing for users who have no other reason to upgrade, so it should not be done just to make a test fixture convenient.
[...]

Issue #363, opened 2026-08-10
backend: mlx is never dispatched — MLX runs train on transformers, and the saved MLX adapter is a full fine-tune


MakazhanAlpamys:
Reported by @nicolasramos in #362, which is the fix in progress. Filing separately so the defect has a permanent record independent of that PR's fate.

## Context

Two independent defects make `backend: mlx` non-functional in v0.73.0. Both are silent — nothing raises, the run exits 0, and the artifact looks plausible.

### D1 — `backend: mlx` is never dispatched

`grep -n "mlx" src/soup_cli/commands/train.py` returns **zero matches**. The trainer-selection chain at `commands/train.py:1331-1417` branches on `cfg.task` only; `cfg.backend` is never consulted. `trainer/mlx_routing.py::get_mlx_trainer` is called from nowhere in `src/` — the only references are `tests/test_mlx_backend.py`.

Consequence: a user on Apple Silicon writes `backend: mlx`, and training runs through the transformers `SFTTrainerWrapper` on MPS. No warning. `MLXSFTTrainerWrapper` has been unreachable from the CLI since it shipped in v0.25.0.

Why the suite missed it: `test_mlx_backend.py::TestMLXRouting::test_mlx_routing_map` asserts `MLX_TRAINER_REGISTRY["sft"].__name__ == "MLXSFTTrainerWrapper"`. That verifies the registry resolves — it never asserts the CLI calls the registry. A mechanism tested for existence but not for connection.

This also means #270's premise does not hold today: the raw `NotImplementedError` in `mlx_dpo.py:50` / `mlx_grpo.py:50` is **unreachable** from a config, because `task: dpo` is caught by the transformers branch first. Fixing D1 makes it reachable, so #270's schema gate becomes a prerequisite of the dispatch fix rather than an independent nicety.

### D2 — the saved MLX "adapter" is a full fine-tune

`mlx_lm.load` does **not** freeze the model. `trainer/mlx_sft.py` never calls `model.freeze()` before applying LoRA, so every parameter stays trainable.

Measured by @nicolasramos on an M4 (LFM2.5-1.2B-Instruct, LoRA r=16 on `self_attn.q_proj` / `self_attn.v_proj`):

| | tensors saved | peak memory |
|---|---|---|
| without `model.freeze()` | 172 | 9.4 GB |
| with `model.freeze()` | 24 | 3.0 GB |

[...]

PR #663, opened 2026-09-04
bench(mlx): publish the MLX SFT run and harness for an 8 GB M1 (#23)

Srinivasan8888:
Publishes the MLX SFT measurement #23 has been waiting for since v0.25.0, plus the harness to re-run it. **No source change.**

`Refs #23` — deliberately not `Fixes`. The issue also asks for a nightly `macos-14` CI job and a README badge, which this does not deliver; see *What is left*.

## Provenance

@nicolasramos held this from 2026-08-11 and [released it on 2026-09-03](https://github.com/MakazhanAlpamys/Soup/issues/23#issuecomment-5530075877) after I asked rather than started. Their pointer — *"if behavior on the low end looks weird, check the dispatch path first"* — is encoded as an assertion in the harness, not a note.

Scope follows @MakazhanAlpamys's 2026-08-11 note: the script committed with exact model ids and command line, a `benchmarks/` entry, and the `benchmarks/README.md` row.

## Results

Five completed runs, one failed fixture, through `MLXSFTTrainerWrapper` rather than `mlx_lm` — the wrapper's loop is the surface #23 says is untested.

| model (4-bit) | train, 48 it | mlx peak | tok/s | loss | adapter | reload |
|---|---|---|---|---|---|---|
| `Qwen2.5-0.5B-Instruct` | 19.7 s | 0.497 GB | ~240 | 3.639 → 0.107 | 2,122 KB | 0.9 s |
| `Llama-3.2-3B-Instruct` | 29.5 s | 2.236 GB | ~82 | 4.527 → 0.330 | 8,972 KB | 2.1 s |
| `Qwen2.5-7B-Instruct` | 67.6 s | 4.922 GB | ~33 | 4.514 → 0.127 | 9,868 KB | 15.8 s |
| **`Llama-3.1-8B-Instruct`** | 71.0 s | **5.154 GB** | ~36 | 3.454 → 0.142 | 13,326 KB | 23.8 s |

**`llama3.1-8b-sft-mlx` — the shipped recipe — trains on an 8 GB M1.** I predicted on the issue that it would not fit. It does. That is the entire reason for measuring rather than reasoning, and the wrong prediction is left standing in the record.

Every acceptance item in the issue's own list that does not need CI is met: `setup()` against a real runtime, `train()` end to end, adapter saved, and the adapter **reloads** via `mlx_lm.load(adapter_path=...)` rather than merely existing.

## The mechanism worth carrying elsewhere

**A model too large does not fail here — it pages.** macOS grew swap 2048 → 9216 MB and free memory sat at 5–6% for long stretches while `ls /` still returned in 16 ms, because MLX memory-maps weights and most of the model is file-backed and evictable.

So **an allocation-failure pre-flight would not fire on this backend.** That is the same trap as the WDDM spill @YuriPerro fixed in #649 — "the step did not raise" is not "the step fits" — arrived at from the opposite platform. Low free-memory is the *normal* shape of this workload, not distress; latency is what distinguishes paging from thrashing.

## The fixture #23 recommends does not load

The issue names `mlx-community/TinyLlama-1.1B-Chat-v1.0-4bit`. It cannot be used:

```
FileNotFoundError: No safetensors found in .../TinyLlama-1.1B-Chat-v1.0-4bit/snapshots/01a7088...
```

The repo resolves and is not gated. Last modified **2024-01-05**, it ships `weights.00.safetensors` — the legacy MLX naming. `mlx_lm` 0.31.3 fetches `allow_patterns=["model*.safetensors", ...]` (`utils.py:239`) and globs `model*.safetensors` (`utils.py:316`), so the weights are never downloaded and load fails on a tokenizer-only snapshot.

**Whoever writes the CI job should use `Qwen2.5-0.5B-Instruct-4bit`** — 282 MB, 48 iterations in 19.7 s, verified here.

## Things I could not explain, left in

- **7B is slower than 8B** (~33 vs ~36 tok/s) despite being smaller, on every reported iteration. Candidate causes named in the record; none tested.
- **Adapter reload scales far worse than adapter size** — 0.9 → 2.1 → 15.8 → 23.8 s. It re-reads the base model, so it is base-dominated, but 15.8 s at 7B against 23.8 s at 8B is steeper than the parameter ratio.
- **Throughput moves ~2x with conditions.** A warm-cache re-run of `Qwen2.5-0.5B` gave **355–480 tok/s against the table's ~240**, same model, same config. Part is the shorter re-run, part is host pressure; I did not separate them. Every tok/s here is an order of magnitude, not a benchmark — the limit the T4 record set.

## Corrections I made while measuring

Both are mine and both are in the record, because they are why the rest is trustworthy.

1. **A bare HTTP 401 from the Hub proves nothing.** I read it as "does not exist" and asserted on that basis that `SmolLM2-135M-Instruct-4bit` was missing. An invented repo id returns the same 401. The correct instrument is `huggingface_hub.model_info`. Unwinding this is what put #661 on evidence instead of a status code.
2. **`target_modules: auto` degrading to Q/V is not a defect.** I recorded it as one, then found #392 and the docstring explaining why it is deliberate. It stays in the record as context for reading the loss curves — these are Q/V-only LoRA, not all-linear — not as a bug.

I also lost the TinyLlama rung to a `tail -120` in my own ladder invocation and re-measured it from scratch rather than reconstructing it.

## What this does not establish

Longer than the results, on purpose, and in full in the record:

- **No correctness claim.** Loss falls and adapters reload. Nothing compares MLX output to a transformers reference or checks gradient exactness.
- **The data is synthetic and trivially repetitive** — 48 short Q/A rows. Loss 3.6 → 0.1 on that is memorisation, not evidence of learning.
- **One run per rung, no repeats, no spread** (and the warm-cache re-run shows the spread is ~2x).
- `load` includes first-time Hub download; warm-cache load was not measured separately.
- **DPO and GRPO untouched** — `_validate_mlx_task` rejects them at config load.
- Not the first non-CUDA record: `gate-qwen4-ple-m4-max.md` predates it. This is the first **MLX-backend training** measurement and the first Apple Silicon **throughput/peak** figures, since that record explicitly disclaims both.

## What is left on #23

Not delivered here, so nobody plans against it:

- the nightly `macos-14` workflow and the README status badge;
- `tests/test_mlx_integration.py` as a CI-collected test;
- **the Rich callback bridge**, which is unimplemented in shipped code — `MLXSFTTrainerWrapper.train()` opens with `del display, tracker, run_id  # accepted for CLI-contract parity`, so an MLX run shows mlx-lm's stdout, not Soup's dashboard. That is the issue's 4th acceptance item and it needs a source change I did not want to bundle into an evidence PR.

Happy to take any of those as follow-ups.

## Verification

```
$ .venv/bin/python -m ruff check benchmarks/harness/ src/soup_cli/ tests/
All checks passed!

$ .venv/bin/python -m pytest tests/test_cli_startup_is_light.py -q -o addopts=
34 passed

$ git diff --check          # clean
```

The committed harness was re-run standalone after being written, to prove it works from the file in the PR rather than from my scratchpad: 16 rows, warm cache, 1.9 s, adapter reloaded in 0.6 s.


Srinivasan8888:

```
1 failed, 19863 passed
```

MakazhanAlpamys:
@Srinivasan8888 — reviewed as the public, citable gate record it becomes on merge, which is a higher bar than a normal PR. **Most of it survives that scrutiny unusually well. One column does not, and it is worth fixing before it is quotable.**

Every Hub-side claim in the writeup, re-derived here with an unauthenticated `HfApi().model_info(...)` and my own negative control:

| id | result |
|---|---|
| `mlx-community/Qwen2.5-0.5B-Instruct-4bit` | EXISTS, ungated, ships `model.safetensors` |
| `mlx-community/Llama-3.2-3B-Instruct-4bit` | EXISTS, ungated |
| `mlx-community/Qwen2.5-7B-Instruct-4bit` | EXISTS, ungated |
| `mlx-community/Llama-3.1-8B-Instruct-4bit` | EXISTS, ungated |
| `mlx-community/TinyLlama-1.1B-Chat-v1.0-4bit` | EXISTS, `lastModified 2024-01-05`, ships **`weights.00.safetensors`**, no `model*.safetensors` |
| `mlx-community/this-repo-does-not-exist-zzz` (control) | `RepositoryNotFoundError`, raised from a 401 |

That confirms all three of your non-obvious claims independently — the exact 2024-01-05 date, the legacy `weights.NN.safetensors` naming that mlx-lm's `model*.safetensors` glob cannot match, and your own "Correction 1", since an invented repo really does return the same 401. That last one is directly load-bearing for #661 and I am glad it is written down here.

The harness compiles, is ruff-clean, writes only under `tempfile.mkdtemp()`, uses list-arg `subprocess.run` with timeouts and no shell, and its `host_mem()` degrades to `"unavailable: ..."` on non-macOS rather than killing the run — I confirmed that by executing it here. The `#363` dispatch assertion is real and would fire. Both index rows are present and accurate. Zero removed lines anywhere in the PR. The 33 failures in the benchmarks-adjacent test set are byte-identical on clean `origin/main` — the documented torch/trl floor problem, #651, not yours.

**On the thing this project actually values:** you asked implicitly and I will answer explicitly — this is the working record, not a report assembled afterwards. A whole failed rung with the mechanism traced to `mlx-lm/utils.py:237-239`; a wrong prediction left standing ("I predicted it would not fit. It fits."); two self-corrections; two observations refused a tidy explanation (7B slower than 8B, reload scaling worse than adapter size); a self-inflicted `tail -120` truncation admitted and re-measured rather than reconstructed; and a "What this does not establish" section deliberately longer than the results. That is the bar, and this clears it.

## The finding — the tok/s column does not close, and the harness cannot produce it

**The committed harness never computes or prints tok/s.** It prints load time, train time, `mx.get_peak_memory()`, the wrapper's result dict, host memory, adapter size and reload time — nothing else. So the "Reproducing" section does not reproduce the tok/s column of the table it sits under, and the header comment's "captured from the printed output of the harness" is not accurate for that column.

Doing the arithmetic your table invites, at a fixed 48 iterations:

| model | train (s) | tok/s | it/s from wall clock | **implied tok/iteration** | tokenizer |
|---|---|---|---|---|---|
| Qwen2.5-0.5B | 19.7 | ~240 | 2.437 | **98.5** | Qwen2 |
| Llama-3.2-3B | 29.5 | ~82 | 1.627 | 50.4 | Llama3 |
| Qwen2.5-7B | 67.6 | ~33 | 0.710 | **46.5** | Qwen2 |
| Llama-3.1-8B | 71.0 | ~36 | 0.676 | 53.2 | Llama3 |

Rows 1 and 3 use the **identical Qwen2 tokenizer** over the **identical 48 generated rows**, so their trained-token totals are necessarily equal and their tok/iteration must match. **98.5 against 46.5 is 2.12x apart.** They cannot both be right.

The ground truth is in your own sibling PR. `tests/test_issue23_mlx_display_bridge.py:41-52` carries real captured mlx-lm dicts:

```
iter 5:  trained_tokens=222, tokens_per_second=19.192,  iterations_per_second=0.432   -> 19.192/0.432  = 44.43 tok/iter
iter 10: trained_tokens=454, tokens_per_second=254.313, iterations_per_second=5.481   -> 254.313/5.481 = 46.40 tok/iter
```

So **~45 tok/iteration is the true shape of this dataset.** That makes rows 2, 3 and 4 consistent with whole-run averages, and row 1 the outlier.

The reconciliation is visible in that same fixture: mlx-lm's `tokens_per_second` is an **instantaneous per-report** value — 19.192 early, 254.313 late in the same run. Row 1's "~240" looks like a late per-report reading while every other column in your table is whole-run. At ~45 tok/iteration the honest run average for row 1 is about **110 tok/s**; conversely, a genuine 240 tok/s would mean 48 iterations in **9.0 s**, not the 19.7 s the same row reports.

**This does not make the record dishonest** — it already warns that tok/s "moves by ~2x with conditions" and reports a 355-480 warm-cache re-run against the same ~240. But a published table whose columns are derived two different ways, with the derivation unstated, is exactly what `benchmarks/README.md` exists to prevent, and someone will cite that column.

Fix, in order of preference:

1. Have the harness compute `trained_tokens / train_seconds` and print it, then re-derive the column from that. This also makes "Reproducing" true, which is the better outcome.
2. Or state in the writeup that tok/s is read from mlx-lm's per-report stdout, say which report, and mark row 1 as instantaneous.

I am not asking you to re-run the 8B; option 2 costs nothing and option 1 only needs the harness change plus whatever runs you feel like redoing.

## Two nits

**The headline quotes a different number than the table bolds.** The table bolds "peak in train" = **4.800 GB** for the 8B row, from `mx.get_peak_memory()`, while the headline, the index row and the harness docstring all quote **5.154 GB**, mlx-lm's own figure. Both are disclosed and the delta is explained, but a citing reader will pick one at random. Say which is the headline number and why.

**No Soup version or commit recorded.** For a citable record, "which Soup did this run against" matters. Context that makes this optional rather than a defect: I checked `gate-395-second-stack-vram.md`, `gate-qwen4-ple-m4-max.md` and `run-t4-colab-free-tier.md` and **none of them state one either**. That is a project-wide gap I should close, not a standard to invent at your expense — so treat it as a suggestion, and I will make it a rule for everyone rather than for you.

## Two answers you did not ask for but should have

**No changelog fragment is correct here.** The PR template says "or this change has no user-visible impact", `changelog.d/README.md` scopes fragments to user-visible pull requests, and five of the six prior benchmarks-only merges shipped without one. Not a finding.

**Index/file consistency is not enforced anywhere**, which I checked because it seemed like it should be. Deleting either of your `benchmarks/README.md` rows breaks no test; the only assertion against that file in the whole suite is a bespoke hardcoded line in `test_issue602_qwen4_ple_streaming.py:1175` pinning one record by name. That is my gap to close repo-wide, not something to bolt onto this PR.

Address the tok/s column and this is ready. The rest of it is the best-documented measurement in `benchmarks/` and I would like it merged.

Srinivasan8888:
**You are right, and it is worse than a presentation issue — row 1 was wrong.** Fixed by your option 1.

## The correction

Every tok/s figure is now `trained tokens / train seconds`, a whole-run average, and the harness computes and prints it:

| model | trained tokens | train s | **tok/s now** | was | |
|---|---|---|---|---|---|
| `Qwen2.5-0.5B` | 2,130 | 19.7 | **108.1** | ~240 | **2.2x too high** |
| `Llama-3.2-3B` | 2,316 | 29.5 | **78.5** | ~82 | |
| `Qwen2.5-7B` | 2,130 | 67.6 | **31.5** | ~33 | |
| `Llama-3.1-8B` | 2,316 | 71.0 | **32.6** | ~36 | |

Your tokenizer argument holds exactly against the raw logs: **2,130 == 2,130** for the Qwen2 rows and **2,316 == 2,316** for the Llama3 rows. That check is now stated in the record, because it is the thing that would have caught me.

## The harness computes it, so "Reproducing" is now true

```
train         : 5.7s   mlx peak during train: 0.463 GB
throughput    : 373.1 tok/s  (2130 trained tokens / 5.7s, whole-run average)
```

It reads mlx-lm's **own** cumulative counter from teed stdout rather than recomputing tokenisation — reproducing the chat template, truncation and masking here would drift silently the first time any of them changed upstream. If the line is absent it prints nothing rather than inventing a figure.

## Two consequences I carried through rather than left stale

**The 7B-slower-than-8B anomaly survives** — 31.5 vs 32.6. It was there before and after, so it is not an artifact of the metric mix-up. That is worth knowing, since a correction that made an unexplained oddity vanish would have been suspicious.

**The warm-cache caveat was wrong on both sides.** It compared instantaneous 355-480 against an instantaneous ~240 and called it "~2x with conditions". Restated with whole-run averages throughout: **373.1 warm against 108.1 cold, 3.5x, same metric**. The spread is larger than I claimed, not smaller.

## The nit

The record now says which peak to cite: `mlx-lm peak` is the per-iteration high-water mark and is strictly the larger, so **that** is the ceiling number — 5.154 GB for the 8B. The bolded column is `mx.get_peak_memory()` sampled after `train()` returns and can miss a transient in-step spike. Headline and index row already used the mlx-lm figure; now they say why.

Logged as **Correction 3** beside the other two, with your name on it. Three self-reported corrections in one record is not a great look for me, but a gate record that hides them is worth less than one that does not.

MakazhanAlpamys:
`dcb13d9` fixes it, and the correction went further than I asked for. Approved on substance — holding only for the matrix.

## Re-derived here

The arithmetic now closes exactly:

```
model             tokens  train_s  claimed  computed  delta
Qwen2.5-0.5B        2130     19.7    108.1     108.1   -0.0%
Llama-3.2-3B        2316     29.5     78.5      78.5   -0.0%
Qwen2.5-7B          2130     67.6     31.5      31.5   -0.0%
Llama-3.1-8B        2316     71.0     32.6      32.6   -0.1%
```

And the invariant my whole finding rested on now holds in the record itself:

```
Qwen2 rows share token total:  True  [2130, 2130]
Llama3 rows share token total: True  [2316, 2316]
```

Harness compiles, `ruff check benchmarks/` clean, both index rows present, and it reads mlx-lm's own `Trained Tokens` counter rather than recomputing tokenisation. That last choice is right and worth stating: reproducing the chat template, truncation and masking in the harness would drift silently the first time any of them changed upstream, and a benchmark that drifts against the thing it measures is worse than one that prints nothing. Printing nothing when the line is absent, rather than inventing a figure, is the same instinct.

## What lifts this above a correction

**You carried the consequences instead of stopping at the number.** Two of them, both of which made your own record look worse:

The 7B-slower-than-8B anomaly **survives** at 31.5 against 32.6. Saying so matters, because a metric correction that had also dissolved an unexplained oddity would be suspicious rather than reassuring — the oddity surviving is evidence the correction was narrow.

And the warm-cache caveat was wrong **in the direction that hurts you**: it compared instantaneous against instantaneous and called the spread ~2x. Restated on one metric it is 373.1 against 108.1, **3.5x**. You could have quietly left that sentence alone and nobody would have re-derived it.

Three self-reported corrections in one record is not a bad look. It is the reason the record is worth citing — a gate record with no corrections in it is either lucky or edited, and `benchmarks/` exists in this project precisely because the working version beats the tidy one.

## The peak-memory nit, resolved better than I framed it

You did not just pick a number, you gave the rule: `mlx-lm peak` is the per-iteration high-water mark and is strictly the larger, so it is the ceiling; `mx.get_peak_memory()` is sampled after `train()` returns and can miss a transient in-step spike. That makes the choice checkable rather than a convention, and it is now stated where a citing reader will hit it.

## Carried forward, not blocking

The Soup version/commit stamp is still absent, and I said last round that this is a project-wide gap rather than yours — `gate-395-second-stack-vram.md`, `gate-qwen4-ple-m4-max.md` and `run-t4-colab-free-tier.md` all omit it too. I will make it a rule for everyone rather than retrofit it onto you.

Merging when the checks report.

MakazhanAlpamys:
Merged as `a351a5e`, credited in `CONTRIBUTORS.md` (`56a2dba`).

## Final verification

```
model             tokens  train_s  claimed  computed  delta
Qwen2.5-0.5B        2130     19.7    108.1     108.1   -0.0%
Llama-3.2-3B        2316     29.5     78.5      78.5   -0.0%
Qwen2.5-7B          2130     67.6     31.5      31.5   -0.0%
Llama-3.1-8B        2316     71.0     32.6      32.6   -0.1%

Qwen2 rows share token total:  True  [2130, 2130]
Llama3 rows share token total: True  [2316, 2316]
```

Harness compiles, `ruff check benchmarks/` clean, both index rows present, all four model ids confirmed on the Hub unauthenticated with a negative control, 14/14 checks green on `dcb13d9`, merge ref built against current `main`.

## Why this one is worth citing

**You corrected your own headline downward by 2.2x on an argument you could have contested.** 240 to 108.1 is not a rounding fix; it is the difference between a number someone quotes and a number that is wrong. And the finding was arithmetic rather than taste — two rows sharing a tokenizer over identical data must share a trained-token total — which means you could see it was right immediately, and acted on it immediately.

**Then you did the part almost nobody does: you followed it into the places where it made your record look worse.**

The 7B-slower-than-8B oddity survives at 31.5 against 32.6, and you said so. A metric correction that also dissolved an unexplained anomaly should make a reader suspicious, not relieved; the anomaly surviving is evidence the correction was narrow rather than convenient.

The warm-cache caveat was wrong **against you**. It compared instantaneous against instantaneous and called the spread ~2x. Restated on one metric it is 373.1 against 108.1 — 3.5x, larger than you had claimed. Nobody was going to re-derive that sentence. You did.

**And the harness change is the right kind of conservative.** Reading mlx-lm's own cumulative counter rather than recomputing tokenisation means the number cannot drift the first time the chat template, truncation or masking changes upstream — and printing nothing when the counter is absent, rather than inventing a figure, is the same instinct applied to the failure case.

Three self-reported corrections in one record is not a bad look. It is why the record is worth citing at all: `benchmarks/` exists in this project because the working version beats the tidy one, and a gate record with no corrections in it is either lucky or edited. Your 401 correction in particular is now load-bearing for #661 and I quoted it back at another contributor today.

## Two carried forward, neither yours to fix

The Soup version/commit stamp is still absent — `gate-395-second-stack-vram.md`, `gate-qwen4-ple-m4-max.md` and `run-t4-colab-free-tier.md` all omit it too, so I will make it a rule for everyone rather than retrofit it onto you. And nothing enforces `benchmarks/README.md` index consistency: deleting either of your rows breaks no test. Also mine.

`Refs #23` — it stays open. This delivers the measurement behind two of its items and disqualifies the tiny fixture the issue itself recommends, which is worth more than a tick.

PR #665, opened 2026-09-04
feat(mlx): drive the live display and tracker from the MLX trainer (#23)

Srinivasan8888:
Closes acceptance item 4 of #23: the MLX path now drives Soup's live dashboard and experiment tracker.

[Scoped on the issue](https://github.com/MakazhanAlpamys/Soup/issues/23#issuecomment-5530800000) before writing, including the `grad_norm` decision below. `Refs #23`.

## What shipped before this

```python
def train(self, display=None, tracker=None, run_id=None, resume_from_checkpoint=None) -> dict:
    """Run MLX training loop via mlx-lm (mlx-lm >= 0.31 API)."""
    del display, tracker, run_id  # accepted for CLI-contract parity
```

An MLX run printed mlx-lm's raw stdout; the identical config on the transformers path got the Rich panel. `_Callback` already implemented `TrainingCallback` — it captured losses into a local dict and drove nothing.

## Shape: an adapter, not a subclass

`SoupTrainerCallback` is a HuggingFace `TrainerCallback` wanting `args, state, control`. mlx-lm offers only `on_train_loss_report(train_info)` / `on_val_loss_report(val_info)`. Reusing the former would mean manufacturing fake HF trainer state and coupling the MLX path to HF internals, so the bridge drives `TrainingDisplay` directly (`start` / `update` / `stop`) and calls `tracker.log_metrics(...)` alongside, gated on `tracker is not None and run_id` exactly as the HF path is.

Mapping, from mlx-lm's own dict at `mlx_lm/tuner/trainer.py:354-362`:

| mlx-lm | display |
|---|---|
| `iteration` | `step` |
| `train_loss` | `loss` |
| `learning_rate` | `lr` |
| `iterations_per_second` | `speed` |
| `peak_memory` | `gpu_mem` |
| `iteration / iters * epochs` | `epoch` |

## The bug a live run caught that nine stub tests did not

My first version fed **`tokens_per_second`** into `speed`. `display.py:116` renders that field as:

```python
lines.append(f"Speed: {self.speed:.2f} it/s")
```

— a hard-coded **`it/s`** label. So the panel showed **19.19 it/s for a run doing 0.43**, a ~44x overstatement, and every one of my nine tests passed because they asserted the value I had chosen rather than the unit the panel prints next to it. Only rendering a real panel on real hardware exposed it.

It now feeds `iterations_per_second`, matching the transformers path's `train_steps_per_second` so both backends label the same row the same way. `test_speed_is_iterations_per_second_not_tokens_per_second` pins it with an assertion that the tokens figure is **not** what arrives — a test that would have caught the original.

## `grad_norm` is deliberately absent

mlx-lm computes no `grad_norm` for the callback; it is not in `train_info`. The dashboard has that field. I left it unset rather than substitute a plausible zero, because a field reading 0.0 on one backend and a real value on another is worse than an obviously empty one. If you would rather it render `n/a` explicitly on the MLX path, that is a display change and I will do it separately.

## Mutation table

Each run against the pre-existing MLX suites (`test_mlx_backend`, `test_issue634_mlx_resume`, `test_review_fixes_v07133`, `test_mlx_dispatch`) and the new file:

| mutation | PRE-EXISTING | NEW | named failure |
|---|---|---|---|
| `display.start()` removed | **SURVIVED** | KILLED | `test_start_receives_the_total_iteration_count` |
| `stop()` moved out of `finally` (happy path only) | **SURVIVED** | KILLED | `test_stop_is_called_even_when_training_raises` |
| `speed`: `iterations_per_second` → `tokens_per_second` | **SURVIVED** | KILLED | `test_speed_is_iterations_per_second_not_tokens_per_second` + 1 |
| tracker gate drops `run_id` | **SURVIVED** | KILLED | `test_tracker_without_run_id_is_not_called` |
| `display` None-guard removed | KILLED | KILLED | already covered — **reported, not counted** |

`mlx_sft.py` md5-verified byte-identical after every round.

**The `finally` row is the discriminating one.** A bridge that calls `stop()` only on the happy path passes every other test in the file, and leaves the Rich `Live` attached to the terminal when mlx-lm raises.

## Verified live, not only stubbed

On the 8 GB M1 from #663, against **real mlx-lm**:

```
dispatch: MLXSFTTrainerWrapper
display.start(total_steps) -> 16
display.update calls      -> 4
display.stop calls        -> 1
tracker.log_metrics calls -> 4
   step=4  epoch=0.25  loss=4.046  lr=1.0e-04  speed=7.816  gpu_mem='0.497 GB'
   step=16 epoch=1.0   loss=0.598  ...
grad_norm present in update? False
```

and the real `TrainingDisplay` rendering during an MLX run:

```
╭────────── Soup Training: mlx-community/Qwen2.5-0.5B-Instruct-4bit ───────────╮
│ Epoch 1.0/1   100%                                                           │
│ Step:  16/16                                                                 │
│ Loss:  0.5473    LR: 1.00e-04                                                │
│ Speed: 8.98 it/s                                                             │
│ GPU peak: 0.781 GB                                                           │
╰────────────────────────────────── Apple M1 ──────────────────────────────────╯
```

8.98 it/s against mlx-lm's own printed 9.0 — the sanity check the unit bug failed.

## Verification

```
$ .venv/bin/python -m pytest tests/test_mlx_backend.py tests/test_mlx_dispatch.py \
    tests/test_issue634_mlx_resume.py tests/test_review_fixes_v07133.py \
    tests/test_issue23_mlx_display_bridge.py tests/test_cli_startup_is_light.py -q -o addopts=
128 passed

All checks passed!

0
```

The stub harness reuses `_install_fake_mlx` from `test_issue634_mlx_resume.py` and runs the **real `train()` body** — real control flow and call ordering, not source-text inspection.

## Limits

- **The stubs are stand-ins for mlx-lm, not a substitute.** They emit the real dict shape (copied verbatim, so upstream drift fails here) but not real cadence. The live half is a manual run on one box, recorded above and in `benchmarks/run-m1-8gb-mlx-sft.md`.
- **`on_val_loss_report` is not bridged.** MLX runs with a val split would not push validation loss to the display. Out of scope here; say the word and it is a small follow-up.
- **Still open on #23:** the nightly `macos-14` workflow, the README badge, and `tests/test_mlx_integration.py` as a CI-collected test. Those involve runner-minute decisions that are yours.

Changelog fragment follows once this has a number.


Srinivasan8888:

```
1 failed, 19863 passed
```

MakazhanAlpamys:
@Srinivasan8888 — reviewed by execution. **This is not the dead-code bug class, and I checked that specifically rather than by grep**, because #363, #423, #659 and #675 were all "MLX code that exists, is tested by calling the helper, and is never reached by the command". Yours is reached.

A probe driving the real `soup train -c cfg.yaml --yes` with `backend: mlx` and `TrainingDisplay` swapped for a recorder:

```
EXIT 0
RECORD {'start': [48],
        'update': [{'step': 5,  'epoch': 0.104, 'loss': 3.639, 'lr': 0.0001, 'speed': 0.432, 'gpu_mem': '0.497 GB'},
                   {'step': 10, 'epoch': 0.208, 'loss': 1.976, 'lr': 0.0001, 'speed': 5.481, 'gpu_mem': '0.497 GB'}],
        'stop': 1}
```

`commands/train.py:1555` threads `display=/tracker=/run_id=` into `trainer_wrapper.train(...)` and `resolve_trainer` binds that to the MLX class, so the bridge is on the CLI path rather than only in a test. With `mlx` absent it degrades correctly too — `_require_mlx()` raises inside `setup()` **before** `display.start()`, so there is no crash and no orphaned Rich `Live`. Ordering that most people get wrong.

**Thirteen mutations, twelve killed**, all restored and re-verified green between runs:

| mutation | result |
|---|---|
| delete `display.start(iters)` | **KILLED** |
| delete the whole `display.update(...)` | **KILLED** (3 tests) |
| tracker gate to `if False:` | **KILLED** |
| **`speed`: `iterations_per_second` -> `tokens_per_second`** | **KILLED** — `test_speed_is_iterations_per_second_not_tokens_per_second` |
| `stop()` moved out of `finally` | **KILLED** — `test_stop_is_called_even_when_training_raises` |
| `gpu_mem` always empty, `epoch` hardcoded, `step` off by one, `start(0)`, drop `run_id`, stop capturing losses | **KILLED** |
| fabricate `grad_norm=1.23` | **KILLED** |
| **inject `grad_norm=0.0`** | **SURVIVED** |

Two things worth calling out as good, not merely acceptable. The `stop()`-out-of-`finally` mutation is the genuinely discriminating one and your test catches it. And **the ten new tests run for real on this non-Apple Windows box — 0 skipped** — because you drive a fake `mlx` through the real `train()` body instead of gating on `skipif(not apple_silicon)`. I was explicitly looking for the opposite failure, a suite that only runs on hardware CI does not have; you did the right thing.

Zero removed test lines. Fragment is `changelog.d/0.74.0/665.added.md`, correct directory, PR number, mentions `#665`, credits you. `ruff` clean.

## Should-fix 1 — the grad_norm test permits exactly what its own docstring condemns

`tests/test_issue23_mlx_display_bridge.py:245`:

```python
assert display.updates[0].get("grad_norm") in (None, 0.0)
```

Eight lines above, the docstring says *"a plausible 0.0 on one backend and a real value on another is worse than one that is obviously absent"*, and the changelog fragment repeats the claim. But `in (None, 0.0)` accepts the 0.0 it just called worse. Verified by injecting `grad_norm=0.0` into the real `display.update(...)` call:

```
10 passed in 0.54s
```

So the contract this PR is most vocal about is the one thing it does not pin. Fix is one line:

```python
assert "grad_norm" not in display.updates[0]
```

In fairness, the practical impact today is nil: `display._render()` gates the row on `self.grad_norm > 0` and `tracker.log_metrics` defaults it to `0.0` anyway, so absent and 0.0 are indistinguishable downstream. This is a test-strength finding, not a behaviour bug — but the whole reason the docstring exists is to stop a later change making them distinguishable in the wrong direction.

## Should-fix 2 — the Web UI dashboard is not driven, and the changelog says otherwise

`push_train_event` is called from exactly one place in the tree, `monitoring/callback.py:217`. The MLX bridge does not call it, so `GET /api/train/stream` and `soup ui`'s live dashboard stay empty for an MLX run while the terminal panel now works. Your fragment says the MLX backend "now drives Soup's live dashboard", which a reader can fairly read as including the Web UI.

Either add the best-effort push — the HF path already wraps it in a bare `try/except`, so the shape exists — or narrow the wording to "the terminal dashboard". Adding it is better; there is no reason `soup ui` should be blank for MLX users.

## Two nits

`src/soup_cli/trainer/mlx_sft.py:326-331`, the `_Callback` docstring enumerates mlx-lm's keys as `iteration, train_loss, learning_rate, tokens_per_second, peak_memory` — it names `tokens_per_second` and omits `iterations_per_second`, which is what the code four lines below actually reads. Given the entire point of this change is that `tokens_per_second` under an `it/s` label is a roughly 44x overstatement, that docstring is primed to re-teach the bug to the next reader. Swap it.

`gpu_mem` format diverges from the HF path: `"3.2/24.0 GB"` (used/total) there, `"0.497 GB"` here. Defensible on unified memory where there is no device/host split, and `soup runs show` renders the raw string, so MLX rows will simply look different. Worth one sentence somewhere rather than a change.

`on_val_loss_report` not being bridged is disclosed in your description and is fine as a follow-up.

## One thing that connects this to #663

After this merges, `benchmarks/harness/mlx_sft_smoke.py` still calls `trainer.train()` with no `display`/`tracker`, so the published harness will not exercise the new bridge. A one-line follow-up would make the next MLX measurement double as the bridge's live regression check on real hardware — which is the half that neither this box nor CI can reach. Not asking for it in this PR.

Fix the two should-fixes and I will re-run the mutation set and merge.

Srinivasan8888:
Both should-fixes done, plus the docstring nit. Your surviving mutation is now killed.

## 1. grad_norm — you were right that it was the one thing unpinned

```
[KILLED] inject grad_norm=0.0  ->  test_grad_norm_is_not_fabricated
```

`assert "grad_norm" not in display.updates[0]`. The old `in (None, 0.0)` accepted exactly what the docstring eight lines above called worse than absence, which is a fair thing to be caught on — the assertion was arguing against itself.

## 2. Web UI — added rather than narrowed, since you preferred that

The bridge now pushes a `TrainEvent` per report, best-effort in the transformers path's shape (bare `try/except`; telemetry must never take down a run). `grad_norm=None`, same contract as the display.

Three tests. The discriminating one is that a **raising** `push_train_event` neither kills the run nor orphans the Rich `Live` — a bridge that pushes unguarded passes everything else and then takes down a real run the first time the buffer is unavailable.

```
[KILLED] drop the SSE push entirely  ->  test_each_report_pushes_a_metric_event
```

**One test I wrote and then corrected.** I first asserted the Web UI should be fed even with no display attached. It failed, correctly: `trainer/sft.py:1865` attaches `SoupTrainerCallback` only `if display:`, so the transformers path pushes nothing without one either. I changed the test to pin the *matching* behaviour rather than change the code to diverge — two backends disagreeing about when `/api/train/stream` goes quiet is a worse outcome than both being quiet together. The docstring says where to revisit it if you ever want headless feeding.

Live on the 8 GB M1, real mlx-lm, panel and buffer together:

```
│ Step: 16/16   Loss: 1.1458   Speed: 8.79 it/s   GPU peak: 0.497 GB │
SSE events pushed: 2
   type=metric step=8  epoch=0.500 loss=4.2491 grad_norm=None
   type=metric step=16 epoch=1.000 loss=1.1458 grad_norm=None
```

## 3. The docstring nit — the sharpest of the three

It listed `tokens_per_second` while the code four lines below reads `iterations_per_second`. Given this whole PR exists because tokens under an `it/s` label is a ~44x overstatement, that docstring was primed to re-teach the exact bug. It now names the key actually read and says why the other is wrong.

Changelog narrowed from "live dashboard" to name the terminal dashboard, the `soup ui` stream and the tracker.

**On `gpu_mem` diverging** (`"0.497 GB"` vs the HF path's `"3.2/24.0 GB"`): left as-is, one sentence in the record. There is no device/host split on unified memory, so a `used/total` figure would either repeat the host RAM already in the table or invent a denominator.

155 passed across the MLX suites, ruff clean, zero removed test lines.

MakazhanAlpamys:
`0068dda` closes all three. Approved on substance — holding only for the matrix, which is still running.

## Re-verified by execution

Baseline 13 passed. Three mutations, three killed, including the one that survived last round:

| mutation | result |
|---|---|
| **inject `grad_norm=0.0`** (SURVIVED last round) | **KILLED** — `test_grad_norm_is_not_fabricated` |
| neutralise the `push_train_event(...)` call | **KILLED** — `test_each_report_pushes_a_metric_event` |
| narrow the `except Exception` guard around the push | **KILLED** — `test_a_failing_push_does_not_kill_training` |

Zero test lines removed. The third one is the one I would have forgotten to write: a bridge that pushes unguarded passes every other test in the file and then takes a real run down the first time the buffer is unavailable. Pinning that the guard *catches* is a different assertion from pinning that the push *happens*, and you have both.

## The judgement call I want on the record

You wrote a test asserting the Web UI should be fed with no display attached, it failed, and **you changed the test rather than the code**:

> `trainer/sft.py:1865` attaches `SoupTrainerCallback` only `if display:`, so the transformers path pushes nothing without one either.

That is the right call and it is the harder one. Making MLX feed the SSE buffer headlessly would have looked like a strictly better feature and quietly created two backends that disagree about when `/api/train/stream` goes quiet — a divergence nobody would notice until someone debugged why `soup ui` behaves differently per backend. Pinning the matching behaviour and documenting where to revisit it is worth more than the extra capability.

## And the docstring nit was the sharpest of the three

It listed `tokens_per_second` while the code four lines below reads `iterations_per_second` — inside the PR whose entire reason for existing is that tokens under an `it/s` label is a ~44x overstatement. A docstring primed to re-teach the exact bug the PR fixes is the kind of thing that survives ten reviews.

Narrowing the changelog from "live dashboard" to naming the terminal panel, the `soup ui` stream and the tracker is also right — the original wording would have been read as covering the Web UI before this round made it true.

The M1 run with the panel and the buffer side by side is the half nobody else here can produce. Thank you for posting the actual output rather than asserting it worked.

Merging when the checks report.

MakazhanAlpamys:
Merged as `87b479c`, credited in `CONTRIBUTORS.md` (`019c324`). Fifth merge of yours, and the second today.

## Final verification

Three mutations, three killed — including the one that survived the first round:

| mutation | result |
|---|---|
| **inject `grad_norm=0.0`** (survived round one) | **KILLED** — `test_grad_norm_is_not_fabricated` |
| neutralise the `push_train_event(...)` call | **KILLED** — `test_each_report_pushes_a_metric_event` |
| narrow the `except Exception` around the push | **KILLED** — `test_a_failing_push_does_not_kill_training` |

13 passed with **zero skips** on this Windows box, 55 passed on the merge against current `main` after #663 landed, zero test lines removed, 14/14 checks green.

## What made this one good

**The bridge is on the CLI path and I could prove it.** Driving the real `soup train -c cfg.yaml --yes` with `TrainingDisplay` swapped for a recorder returns `start(48)`, per-report updates carrying every field, and one `stop`. That check is not ceremony here — four MLX defects in this repo have been exactly "code that exists, is tested by calling the helper, and is never reached by the command", and #672 was the fourth, two days ago.

**The three-way split on the SSE push is the part I would not have asked for precisely.** That the push happens, that the guard catches, and that a raising push neither kills the run nor orphans the Rich `Live` are three different properties, and a bridge that pushes unguarded passes the first two and then takes a real run down the first time the buffer is unavailable. You pinned all three.

**And the test you corrected instead of the code is the best decision in the PR.** Making MLX feed the buffer headlessly would have read as a strictly better feature and quietly produced two backends that disagree about when `/api/train/stream` goes quiet — the sort of divergence discovered months later by someone debugging why `soup ui` behaves differently per backend. Pinning the matching behaviour and recording in the docstring where to revisit it is worth more than the extra capability.

The docstring nit deserves its own line: it listed `tokens_per_second` while the code four lines below reads `iterations_per_second`, *inside the PR whose entire reason for existing is that tokens under an `it/s` label is a ~44x overstatement*. That is a comment primed to re-teach the exact bug being fixed, and those survive a lot of reviews.

## Follow-up, now unclaimed on #23

`benchmarks/harness/mlx_sft_smoke.py` — which you also wrote, merged an hour ago as `a351a5e` — still calls `trainer.train()` with no `display`/`tracker`, so the published harness does not exercise the bridge that now exists. One line there makes the next MLX measurement double as this bridge's live regression check on real hardware, which is the half neither CI nor any maintainer box can reach.

It is yours if you want it; I have also left it on #23 as unclaimed rather than assuming, since you have shipped two PRs today already.

`Refs #23` — still open. `on_val_loss_report` remains unbridged, as you disclosed, and the README badge is unclaimed.

PR #664, opened 2026-09-04
feat(recipes): add kimi-k2.6-dpo recipe (#275)

kok-o:
Adds `kimi-k2.6-dpo`, the direct preference optimization (DPO) variant for `moonshotai/Kimi-K2.6`. Catalog 163 -> 164.

`Refs #275`. [Claimed](https://github.com/MakazhanAlpamys/Soup/issues/275#issuecomment-5538500241) before writing code.

## Overlap

- `gh pr list --search "kimi-k2.6-dpo" --state all` returns 0 PRs.
- `gh pr list --search "kimi" --state all` returns #614 (`kimi-k2.6-grpo`, merged) and #512 (closed). No open PR touches `kimi-k2.6-dpo`.

## Shape, and why

`moonshotai/Kimi-K2.6` shipped SFT (v0.71.24) and GRPO (#614) but no preference variant. This completes the task trio (SFT, GRPO, DPO) for Kimi-K2.6 under #275.

The hyperparameters adhere strictly to the established MoE DPO convention:
- `lr: 5e-6` (unanimous across all 12 DPO recipes)
- `dpo_beta: 0.1` (unanimous across all DPO recipes)
- `epochs: 1` (matches sibling `kimi-k2.6-sft` and `glm-5.1-dpo` for massive MoE models)
- `batch_size: 1`, `gradient_accumulation_steps: 16` (matches `glm-5.1-dpo` and SFT geometry rather than rollout constraints)
- `lora: r: 32, alpha: 64, target_modules: auto` (matches `kimi-k2.6-sft` and `glm-5.1-dpo`)
- `quantization: 4bit`, `moe_lora: true`, `moe_aux_loss_coeff: 0.01`, `gradient_checkpointing: true`

## Mutation table

Tested against `TestKimiK26DpoRecipe`:

| # | mutation | verdict | named test |
|---|---|---|---|
| 1 | `RecipeMeta.model` alone | KILLED | `test_recipe_meta_pins_the_model_id` |
| 2 | YAML `base:` alone | KILLED | `test_yaml_base_pins_the_model_id` |
| 3 | `RecipeMeta.task` alone | KILLED | `test_recipe_meta_pins_the_model_id` |
| 4 | `RecipeMeta.size` 1T -> 7B | KILLED | `test_recipe_meta_pins_the_model_id` |
| 5 | Delete whole entry | KILLED | `test_recipe_meta_pins_the_model_id` |

## Verification

```
[...]

PR #666, opened 2026-09-04
fix(recipes): repair Hugging Face repo IDs in MLX recipes (#661)

### LINKED ISSUE #661: Two of the three MLX recipes name Hugging Face repos that do not exist (qwen3-8b-sft-mlx, gemma3-9b-sft-mlx)

kok-o:
Repairs the repository IDs for the MLX recipes in `src/soup_cli/recipes/catalog.py`.

`Fixes #661`. [Claimed](https://github.com/MakazhanAlpamys/Soup/issues/661#issuecomment-5538677387) before writing code.

## Overlap
- No other PR touches `qwen3-8b-sft-mlx` or `gemma3-9b-sft-mlx`.

## Root Cause & Changes

1. **`qwen3-8b-sft-mlx`:**
   - Previous repo ID `mlx-community/Qwen3-8B-Instruct-4bit` returned HTTP 401 / Not Found.
   - Qwen3 dropped the `-Instruct` suffix; the instruct model is just `Qwen3-8B`.
   - Updated to `mlx-community/Qwen3-8B-4bit` (HTTP 200 on Hub).

2. **`gemma3-9b-sft-mlx` -> `gemma3-4b-sft-mlx`:**
   - Gemma 3 has no 9B model size (sizes are 1b, 4b, 12b, 27b; 9B was Gemma 2 only).
   - Resolved to **`gemma3-4b-sft-mlx`** (`mlx-community/gemma-3-4b-it-4bit`, HTTP 200).
   - *Rationale:* MLX recipes target Apple Silicon laptops (8 GB / 16 GB), where 4B 4-bit runs comfortably without severe paging/pressure, while keeping the Gemma 3 family represented in the MLX catalog.
   - Updated size to `4B` and updated description.

## Mutation Testing

Tested against `TestMlxRecipeRepoIds`:

| # | mutation | verdict | named test |
|---|---|---|---|
| 1 | Qwen `RecipeMeta.model` alone | KILLED | `test_meta_model_matches_expected_repo` |
| 2 | Qwen YAML `base:` alone | KILLED | `test_yaml_base_matches_expected_repo` |
| 3 | Gemma `RecipeMeta.model` alone | KILLED | `test_meta_model_matches_expected_repo` |
| 4 | Gemma YAML `base:` alone | KILLED | `test_yaml_base_matches_expected_repo` |

## Verification

```
$ pytest tests/test_mlx_backend.py tests/test_issue621_recipe_snapshot.py -q -o addopts=
192 passed in 1.99s

$ ruff check src/soup_cli/ tests/
All checks passed!
```

- Live CLI: `soup recipes show qwen3-8b-sft-mlx` and `soup recipes show gemma3-4b-sft-mlx` render valid configs.
- Recipe config snapshot fixture regenerated additively.

Changelog fragment added in `changelog.d/0.74.0/666.fixed.md`.

### DIFF
--- a/changelog.d/0.74.0/666.fixed.md
+++ b/changelog.d/0.74.0/666.fixed.md
@@ -1,0 +1,7 @@
+- Repaired non-existent Hugging Face repository IDs in MLX catalog recipes
+  (#661 by @kok-o in #666). `qwen3-8b-sft-mlx` previously pointed to
+  `mlx-community/Qwen3-8B-Instruct-4bit` (which huggingface_hub reports as
+  RepositoryNotFoundError); it now references `mlx-community/Qwen3-8B-4bit`.
+  `gemma3-9b-sft-mlx` is renamed to `gemma3-4b-sft-mlx` (user-visible rename;
+  the old 9B size does not exist upstream) pointing to
+  `mlx-community/gemma-3-4b-it-4bit` (size 4B, M1+ 16GB).
--- a/src/soup_cli/recipes/catalog.py
+++ b/src/soup_cli/recipes/catalog.py
@@ -1239,13 +1239,13 @@ output: ./output
 """,
     ),
     "qwen3-8b-sft-mlx": RecipeMeta(
-        model="mlx-community/Qwen3-8B-Instruct-4bit",
+        model="mlx-community/Qwen3-8B-4bit",
         task="sft",
         size="8B",
         tags=("qwen", "qwen3", "mlx", "apple-silicon", "sft"),
         description="Qwen 3 8B SFT on Apple Silicon via MLX (M2+ 16GB)",
         yaml_str="""\
-base: mlx-community/Qwen3-8B-Instruct-4bit
+base: mlx-community/Qwen3-8B-4bit
 task: sft
 backend: mlx
 
@@ -1267,14 +1267,14 @@ training:
 output: ./output
 """,
     ),
-    "gemma3-9b-sft-mlx": RecipeMeta(
-        model="mlx-community/gemma-3-9b-it-4bit",
+    "gemma3-4b-sft-mlx": RecipeMeta(
+        model="mlx-community/gemma-3-4b-it-4bit",
         task="sft",
-        size="9B",
+        size="4B",
         tags=("gemma", "gemma3", "mlx", "apple-silicon", "sft"),
-        description="Gemma 3 9B SFT on Apple Silicon via MLX (M2+ 16GB)",
+        description="Gemma 3 4B SFT on Apple Silicon via MLX (M1+ 16GB)",
         yaml_str="""\
-base: mlx-community/gemma-3-9b-it-4bit
+base: mlx-community/gemma-3-4b-it-4bit
 task: sft
 backend: mlx
 
--- a/tests/fixtures/recipe_config_snapshots.json
+++ b/tests/fixtures/recipe_config_snapshots.json
@@ -593,6 +593,15 @@
       "training.lora.r": 16,
       "training.lr": 1e-05
     },
+    "gemma3-4b-sft-mlx": {
+      "backend": "mlx",
+      "base": "mlx-community/gemma-3-4b-it-4bit",
+      "data.train": "./data/train.jsonl",
+      "training.batch_size": 2,
+      "training.lora.alpha": 32,
+      "training.lora.r": 16,
+      "training.lr": 0.0001
+    },
     "gemma3-9b-sft": {
       "base": "google/gemma-3-9b-it",
       "data.format": "auto",
@@ -600,15 +609,6 @@
       "training.lora.alpha": 32,
       "training.lora.r": 16,
       "training.lr": 0.0002
-    },
-    "gemma3-9b-sft-mlx": {
-      "backend": "mlx",
-      "base": "mlx-community/gemma-3-9b-it-4bit",
-      "data.train": "./data/train.jsonl",
-      "training.batch_size": 2,
-      "training.lora.alpha": 32,
-      "training.lora.r": 16,
-      "training.lr": 0.0001
     },
     "glm-4.6-sft": {
       "base": "THUDM/glm-4.6",
@@ -1628,7 +1628,7 @@
     },
     "qwen3-8b-sft-mlx": {
       "backend": "mlx",
-      "base": "mlx-community/Qwen3-8B-Instruct-4bit",
+      "base": "mlx-community/Qwen3-8B-4bit",
       "data.train": "./data/train.jsonl",
       "training.batch_size": 2,
       "training.lora.alpha": 32,
--- a/tests/test_mlx_backend.py
+++ b/tests/test_mlx_backend.py
@@ -223,14 +223,22 @@ class TestMLXRecipes:
 
         recipe = get_recipe("qwen3-8b-sft-mlx")
         assert recipe is not None
+        assert recipe.model == "mlx-community/Qwen3-8B-4bit"
         cfg = load_config_from_string(recipe.yaml_str)
+        assert cfg.base == "mlx-community/Qwen3-8B-4bit"
         assert cfg.backend == "mlx"
 
-    def test_gemma3_9b_sft_mlx(self):
+    def test_gemma3_4b_sft_mlx(self):
+        from soup_cli.config.loader import load_config_from_string
         from soup_cli.recipes.catalog import get_recipe
 
-        recipe = get_recipe("gemma3-9b-sft-mlx")
+        recipe = get_recipe("gemma3-4b-sft-mlx")
         assert recipe is not None
+        assert recipe.model == "mlx-community/gemma-3-4b-it-4bit"
+        assert recipe.size == "4B"
+        cfg = load_config_from_string(recipe.yaml_str)
+        assert cfg.base == "mlx-community/gemma-3-4b-it-4bit"
+        assert cfg.backend == "mlx"
 
     def test_mlx_dpo_config_rejected_at_load(self):
         """backend=mlx + task=dpo is rejected by the SoupConfig validator."""
@@ -268,6 +276,37 @@ output: ./output
 """
         with pytest.raises(ValueError, match="MLX backend only ships SFT"):
             load_config_from_string(yaml_str)
+
+
+class TestMlxRecipeRepoIds:
+    """Pin the repo id every shipped MLX recipe declares, on RecipeMeta.model
+    and YAML base: independently (#661). This does not check that the repo resolves.
+    """
+
+    MLX_RECIPES = [
+        ("llama3.1-8b-sft-mlx", "mlx-community/Llama-3.1-8B-Instruct-4bit"),
+        ("qwen3-8b-sft-mlx", "mlx-community/Qwen3-8B-4bit"),
+        ("gemma3-4b-sft-mlx", "mlx-community/gemma-3-4b-it-4bit"),
+    ]
+
+    @pytest.mark.parametrize("name,expected_repo", MLX_RECIPES)
+    def test_meta_model_matches_expected_repo(self, name: str, expected_repo: str):
+        from soup_cli.recipes.catalog import get_recipe
+
+        recipe = get_recipe(name)
+        assert recipe is not None, f"Recipe {name} missing from catalog"
+        assert recipe.model == expected_repo
+
+    @pytest.mark.parametrize("name,expected_repo", MLX_RECIPES)
+    def test_yaml_base_matches_expected_repo(self, name: str, expected_repo: str):
+        import yaml
+
+        from soup_cli.recipes.catalog import get_recipe
+
+        recipe = get_recipe(name)
+        assert recipe is not None
+        parsed = yaml.safe_load(recipe.yaml_str)
+        assert parsed["base"] == expected_repo
 
 
 # ---------------------------------------------------------------------------


MakazhanAlpamys:
@kok-o — reviewed by execution. **The fix is right, both replacement ids are real, and every mutation I tried was killed.** One mechanical blocker (a rebase), two things in the wording I want changed, and one open question from #661 that I will rule on rather than leave to you.

## The Hub check, re-run here with a negative control

#661 opens by demonstrating that a bare 401 proves nothing, so I re-derived rather than trusting either the issue or your description. `huggingface_hub` 1.30.0, unauthenticated, `HF_TOKEN` unset:

```
mlx-community/Qwen3-8B-Instruct-4bit        RepositoryNotFoundError
mlx-community/Qwen3-8B-4bit                 EXISTS gated=False
mlx-community/gemma-3-9b-it-4bit            RepositoryNotFoundError
mlx-community/gemma-3-4b-it-4bit            EXISTS gated=False
mlx-community/Llama-3.1-8B-Instruct-4bit    EXISTS gated=False
mlx-community/zz-invented-control-xyz       RepositoryNotFoundError   <- control
```

Sibling repos in the same org resolve while the invented control does not, so this is genuinely "missing" rather than an auth artefact. **Both ids you chose exist and are ungated.** Picking `gemma-3-4b-it-4bit` over the 12B was the right call for an Apple Silicon recipe, and it is the size #661's author leaned toward for the same reason.

## Mutations, all killed

Control: **267 passed**. `catalog.py` md5 identical before and after the whole matrix.

| mutation | result |
|---|---|
| Qwen id reverted in `RecipeMeta.model` **only** | **KILLED** — `test_meta_model_matches_expected_repo`, `TestMLXRecipes::test_qwen3_8b_sft_mlx` |
| Qwen id reverted in YAML `base:` **only** | **KILLED** — 3 tests incl. the snapshot |
| Gemma reverted to `gemma-3-9b-it-4bit`, both sites | **KILLED** — 4 tests |
| both recipes reverted | **KILLED** — 8 tests |
| Qwen changed to a *third* nonexistent id | **KILLED** — 4 tests |
| `llama3.1-8b-sft-mlx` (which you did **not** touch) made bogus | **KILLED** — 3 tests |

The first row is the one that earns the new test its place: the pre-existing snapshot test only sees the YAML-derived config, so a wrong `RecipeMeta.model` slips past it entirely. `test_meta_model_matches_expected_repo` is the only thing that catches it, and it retroactively pins the Llama recipe too. That is a real coverage gain rather than ceremony.

Zero assertions deleted or weakened — the 16 removed lines are the snapshot rename, the dropped commit, and a one-assertion `test_gemma3_9b_sft_mlx` replaced by a four-assertion `test_gemma3_4b_sft_mlx`.

## Blocking, and purely mechanical — drop commit 3 on rebase

The PR is `DIRTY`. Exactly one of your three commits conflicts:

```
c6459df fix(recipes): repair Hugging Face repo IDs ...     applies cleanly
c228b50 docs(changelog): add fragment for #666             applies cleanly
3e0a1dc fix(tests): stop pinning a credit assertion ...    CONFLICT
```

One file, one hunk, in `tests/test_issue602_qwen4_ple_streaming.py`. **Main already landed the same fix, twice** — `766f348` and `450442f` — and converged on the identical strategy (glob `changelog.d/**/603.*.md`, then fall back to `CHANGELOG.md`). Main's version is strictly stronger: it adds `if p.is_file()`, asserts `texts` is non-empty so "neither source is readable" cannot pass silently, and gives the gate-record assertion a failure message. Yours calls `.read_text()` unconditionally.

So: **`git rebase origin/main`, then `git rebase --skip` on that commit.** Nothing is lost. I ran it here to be sure rather than to assume — the result is a clean two-commit branch touching only the four files this PR is actually about, and `tests/test_mlx_backend.py tests/test_recipes.py tests/test_issue487_changelog_fragments.py` gives **107 passed**.

No criticism attached to that commit existing: it was how you got your own CI green while #669 was still open, and you had no way to know main would land the same repair first.

## SHOULD-FIX — the changelog cites the one piece of evidence #661 disproved

`changelog.d/0.74.0/666.fixed.md:3`:

> `mlx-community/Qwen3-8B-Instruct-4bit` (which returns 401 ...)

#661's first section exists specifically to show that a 401 is not evidence — an invented repo returns the same 401, and that is why the issue reports `RepositoryNotFoundError` from `huggingface_hub` instead. Publishing the 401 in the changelog teaches the wrong diagnostic to the next person who hits this. Please make it:

> which `huggingface_hub` reports as `RepositoryNotFoundError`

## SHOULD-FIX — the new docstring claims more than the test does

`tests/test_mlx_backend.py:282`:

> `"""Issue #661 regression: verify all shipped MLX recipes name valid Hugging Face repos.`

It verifies no such thing. It pins three literal strings; nothing in this PR contacts the Hub. That is the correct design for a unit test — but a docstring that describes network validation is how the next person concludes the ids are checked automatically and stops looking. Suggested:

> Pin the repo id every shipped MLX recipe declares, on `RecipeMeta.model` and YAML `base:` independently. This does **not** check that the repo resolves.

## My ruling on #661's fourth acceptance item

#661 asks for "a guard that would catch the next one, or a stated decision not to have one". That is a maintainer call, not yours, so here it is: **not in this PR.** A real guard has to hit the network, has to distinguish missing from gated, and belongs in a scheduled job rather than the PR matrix where it would make every unrelated PR depend on Hub availability. I am filing it separately and #661 closes with this merge.

## Two nits, neither blocking

**The `8 GB` claim is unmeasured.** `description="... (M1+ 8GB/16GB)"` and the fragment's "comfortably fitting 8 GB". Entirely plausible for a 4B at 4-bit, but nobody ran it, and this project's whole posture is that measured and plausible are different words. Either soften to `M1+ 16GB` — matching the conservative phrasing on the other two recipes — or, better, @Srinivasan8888 has the M1/8 GB box and is measuring MLX on #663 right now, so a one-line confirmation would let you keep the stronger claim honestly.

**The rename is a user-visible break and the fragment does not say so.** `soup recipes use gemma3-9b-sft-mlx` now exits non-zero. Harmless in practice — the recipe could never have trained, and the "Did you mean: gemma3-4b-sft-mlx" suggestion catches it, which I checked — but one clause in the fragment costs nothing.

Push the rebase and those two wording changes and I will re-run the mutation set and merge.

kok-o:
@MakazhanAlpamys — rebased onto current `main` and addressed all review findings:

1. **Dropped conflicting commit on rebase**: Skipped the credit-assertion commit during `git rebase` since main already landed the comprehensive repair in 766f348 / 450442f. The branch now consists of exactly two clean commits.
2. **Changelog wording (RepositoryNotFoundError & rename)**: Replaced the 401 mention with `which huggingface_hub reports as RepositoryNotFoundError`, explicitly noted the user-visible rename of `gemma3-9b-sft-mlx` to `gemma3-4b-sft-mlx`, and softened memory wording to `M1+ 16GB`.
3. **Docstring precision**: Updated `TestMlxRecipeRepoIds` docstring in `tests/test_mlx_backend.py` to state that it pins declared repo IDs on RecipeMeta.model and YAML base: without claiming network resolution.
4. **Catalog description**: Softened `description` in `catalog.py` to `(M1+ 16GB)` to match the conservative phrasing of the other MLX recipes.

Verified locally: `pytest` passed 112 tests across `test_mlx_backend.py`, `test_recipes.py`, and `test_issue487_changelog_fragments.py`, and `ruff check` is clean.

MakazhanAlpamys:
Merged as `481a32a`, credited in `CONTRIBUTORS.md` (`43628ed`).

## Verified after your rebase

Re-ran the Hub check with my own negative control rather than reusing the earlier one, and re-ran the mutations, because a rebase is exactly when a pin can quietly stop pinning:

```
mlx-community/Qwen3-8B-Instruct-4bit    RepositoryNotFoundError
mlx-community/Qwen3-8B-4bit             EXISTS gated=False
mlx-community/gemma-3-9b-it-4bit        RepositoryNotFoundError
mlx-community/gemma-3-4b-it-4bit        EXISTS gated=False
mlx-community/zz-invented-control-xyz   RepositoryNotFoundError   <- control
```

Reverting the Qwen id alone still kills three tests. **112 passed** on the branch, **128 passed** on the merge against current `main` after #664 landed, catalog still 165, `gemma3-4b-sft-mlx` present and `gemma3-9b-sft-mlx` gone. Branch is exactly two clean commits after the `--skip`.

## The correction I care most about is the one to your own changelog

The fragment first cited the broken id as *"returns 401"*. That is the single piece of evidence #661 opens by disproving — an invented repo returns the same 401, which is why the issue reports `RepositoryNotFoundError` instead. You replaced it. That matters more than a wording nit because the changelog ships verbatim to users, and publishing the 401 would have taught the wrong diagnostic to whoever hits this next; the next person would run `curl`, see a 401, and conclude nothing.

Same shape in the docstring fix. It now says plainly that the test pins declared ids and does **not** check that they resolve. A docstring claiming network validation is how someone concludes the ids are checked automatically and stops looking — which is roughly how this bug shipped in the first place.

## The rest, briefly

**The `--skip` was the right resolution, not a workaround.** Main had landed the same repair twice (`766f348`, `450442f`) and in a strictly stronger form — `if p.is_file()`, a non-empty assertion on `texts`, and a message on the gate-record assertion. Dropping your commit loses nothing. No criticism attached to it having existed: it was how you got your own checks green while #669 was open, and you could not have known main would land it first.

**You flagged the rename as a user-visible break** rather than filing it silently under `fixed`, and **softened the 8 GB claim to 16 GB** rather than defending it. Both were nits I explicitly said I would not block on, and you fixed them anyway. The memory one is the better instinct: nobody had run a 4B on an 8 GB box, and "plausible" and "measured" are different words here.

**Your `RecipeMeta.model` pin is the load-bearing one.** The pre-existing snapshot test only ever sees the YAML-derived config, so a wrong `RecipeMeta.model` slips past everything — reverting that surface alone is caught by your test and nothing else. It also retroactively pins the Llama recipe that was already correct.

## What I ruled rather than asked you to do

#661's fourth acceptance item asked for a guard against the next one. Not in this PR: a real guard has to hit the network, has to distinguish missing from gated, and belongs in a scheduled job rather than the PR matrix where it would make every unrelated PR depend on Hub availability. Filed as #677. #661 closes with this merge.

Two merges of yours today, and #529 is still in flight.

PR #667, opened 2026-09-04
fix(doctor): report MLX from the doctor command, and read the version where it lives

### LINKED ISSUE #659: soup doctor never reports MLX: _get_mlx_info() has no caller, and get_mlx_version() always returns "unknown"

swalla02:
Closes #659.

## What was wrong

**1. `_get_mlx_info()` had no caller.** The helper, its docstring ("Surface MLX info in the doctor report") and its test all existed, but `doctor` never invoked it, so `soup doctor` printed nothing about MLX on any machine. The `Backend: MPS (Apple Silicon)` line comes from the torch MPS branch of `_check_gpu()`, not from MLX.

**2. `get_mlx_version()` returned `"unknown"` on every correctly installed machine.** It read `mlx.__version__`, which the top level package does not export. The attribute lives on `mlx.core`.

Defect 1 is why defect 2 was never noticed. Nobody had seen the field rendered.

## What this does

`doctor` now calls `_check_mlx()` between the GPU and resources sections. It renders a panel in two cases:

* MLX importable: version, chip, unified memory
* Apple Silicon with MLX absent: `not installed`, plus the `soup-cli[mlx]` install hint

A machine that is neither gets no panel, so output on CUDA and CPU only boxes is unchanged. An absent optional extra is not an environment defect, so it never enters the issues list and `All checks passed!` still holds.

`get_mlx_version()` now asks `mlx.core.__version__` first, then the package surface (some builds and the suite's mocked imports carry it there), then `importlib.metadata.version("mlx")`, with `"unknown"` as a genuine last resort. `None` still means MLX is not installed.

The old `test_doctor_has_mlx_info` imported the helper and called it directly, so it passed whether or not `doctor` invoked it. That is exactly why this survived. `TestMLXDoctor` now drives the command through `CliRunner`, and includes a control that fails if the panel is ever rendered unconditionally, plus a spy that fails by name if the call site is deleted again.

## Verified

* Full suite on Linux x86_64, Python 3.12: **19,960 passed, 86 skipped, 0 failed**
* `ruff check src/soup_cli/ scripts/ tests/` clean
* Red to green both ways: removing the `_check_mlx()` call fails exactly the 3 new command tests; reverting `utils/mlx.py` fails exactly the 2 new version tests
* Apple Silicon paths driven end to end through the real `_get_mlx_info`, `get_mlx_version` and CLI with the machine faked (`platform.system`/`machine` to Darwin/arm64, `sysctl` stubbed to an M1 with 8 GiB, a fake `mlx` carrying `__version__` on `mlx.core`)

## Not verified

**Nothing was run on Apple hardware.** `mlx` is not installable on Linux, so the three acceptance boxes in #659 that need a real Mac are unchecked by me. If someone with an M series box can run `soup doctor` on this branch, that closes them.

## How it was found

I used [codegraph](https://github.com/swalla02/codegraph), a static call graph tool I have been building. It answers "who calls this" across a repository, and it separates callers that are tests from callers that are not. On `_get_mlx_info` the answer was: no production callers, one test caller. That shape, defined and tested but never called, is the signature of this bug, and it is awkward to express with grep since grep just shows you the name three times and leaves the classification to you.

I then asked it for the same shape across the whole repo. Five candidates survived filtering, of which two look real:

* `_get_mlx_info` (this PR)
* `commands/migrate.py::_looks_like_jsonl`, which appears to have been born dead in `56bea56`. That commit added the helper alongside a comment saying "The sniff helper is only invoked when the suffix says `.jsonl`", and it is never invoked. Its three tests call it directly, and no test drives `soup migrate` with a `.jsonl` file, so the `typer.Exit(2)` branch has no coverage. User impact looks small since the suffix check already fires, but a JSONL file named `.json` still gets an opaque failure. Happy to file that separately if useful.

The other three are deliberate: `_generate_model_card` and `_default_cmaes_scorer` are documented back compat shims, and `_selfcheck` is an import probe whose docstring says test only is the point.

Sharing the tool in case it is useful to you, not as a pitch. It is early and it has real gaps (it cannot see a function passed as a callback, which is why the filtering above needed a manual pass).

## Note on process

This was written with AI assistance, per the workflow your `AGENTS.md` anticipates. The commit carries a `Co-Authored-By` trailer. Every claim above that says "verified" was verified by running it; the Apple Silicon section says plainly what was not.


swalla02:
Closing this in favour of #660, which is @ump45nose's and came first.

I opened this without commenting on #659 to claim it, which is the step CONTRIBUTING asks for and the step that would have stopped a third PR landing on an already-crowded thread. That one is on me, and sorry for adding to the churn.

#660 is also the better PR to carry: same wiring, same `mlx.core.__version__` with a metadata fallback, same rewrite of the test to drive the command rather than the helper, and @ump45nose has Apple Silicon to verify the version string on real hardware. I do not.

One piece here is not in #660, if it is wanted: the `docs/backends-and-ops.md` hunk documenting that the MLX panel stays silent on non-Apple platforms. Happy for @ump45nose to lift it into #660, or I can send it as a separate docs PR once #660 lands, or it can be dropped. Whichever is least work for the queue.

PR #668, opened 2026-09-04
test(mlx): assert the smoke adapter LOADS, not merely that it exists (#23)

Srinivasan8888:
Closes the last unmet acceptance criterion on #23 — *"adapter file saved and loadable"*. **Test-only, one file.**

Per [my status check on #23](https://github.com/MakazhanAlpamys/Soup/issues/23#issuecomment-5531000000), the other criteria are already met or obsolete: #394 put `mlx-smoke` on `macos-14` for **every PR** (stronger than the nightly this issue proposed), `README.md:28` already badges `ci.yml` which contains that job, and #665 closed the Rich callback item.

## The gap

`test_mlx_sft_smoke` asserted the two files exist and stopped:

```python
assert (output_dir / "adapters.safetensors").exists()
assert (output_dir / "adapter_config.json").exists()
```

**Existence is not loadability**, and the difference has a named failure mode in this repo already — `resolve_mlx_target_keys`'s own docstring, from #392:

> a default run trained Q/V and then shipped `{"keys": ["auto"]}`. `linear_to_lora_layers` matches nothing against that and `load_weights(strict=False)` drops every LoRA tensor **in silence**, leaving an adapter whose generations are bit-identical to the base model.

So an adapter can be written, be well-formed, carry 8 real tensors, satisfy both assertions — and be **inert**.

## The assertion, and why it is this one

Not "load does not raise" — a dropped-tensor adapter loads perfectly happily. The check is that loading **attaches** LoRA parameters:

```python
base_model, _  = load(_MLX_SMOKE_BASE)
tuned_model, _ = load(_MLX_SMOKE_BASE, adapter_path=str(output_dir))
lora_keys = {k for k in tuned_keys - base_keys if ".lora_" in k}
assert lora_keys, ...
```

Established empirically before writing it, rather than assumed: on a healthy run this base goes **21 params → 29**, with 12 new keys including `model.layers.0.self_attn.q_proj.lora_a` / `.lora_b`.

## Proof it has teeth

I injected the **actual #392 defect** — reverted `build_mlx_adapter_config` to write `{"keys": ["auto"]}` while `_apply_lora` still trains the resolved Q/V list — and measured which assertions notice:

| scenario | verdict |
|---|---|
| clean tree, WITH the loadability assertion | **PASS** |
| #392 bug injected, **OLD assertions only** | **PASS** — ships undetected |
| #392 bug injected, WITH the assertion | **FAIL** — caught |

```
AssertionError: the adapter loaded but attached no LoRA parameters:
21 params before, 21 after. That is the #392 shape -- an adapter that
exists and is inert.
```

Both `mlx_sft.py` and `test_smoke_train.py` restored **byte-identical** afterwards (md5 verified).

The failure message names the mechanism and prints the counts, so whoever sees it red does not have to rediscover #392 to understand it.

## Verified in the real job environment

Not the dev venv — **the dev venv cannot run this job.** Step 2 asserts `torch`, `trl`, `datasets`, `peft`, `accelerate`, `bitsandbytes` are *absent*, and `.[dev]` installs all six, which is exactly why #394 kept `mlx-smoke` out of the main matrix.

So I built a clean `.[mlx]`-only venv on Apple Silicon (Python 3.12, since system Python here is 3.9.6, below the floor) and ran all three job steps verbatim:

```
step 1: import mlx.core, mlx_lm     -> 0.32.2 0.31.3
step 2: PyTorch stack must be ABSENT -> PASS
step 3: pytest -o addopts= -m smoke -k mlx_sft_smoke -q tests/test_smoke_train.py
        1 passed, 2 deselected in 2.76s
```

**Cost:** 2.6–2.8 s for the whole test including the extra load, against ~4.3 s before on a cold HF cache. The assertion does not meaningfully change job runtime — it reuses the model the run already downloaded.

## Verification

```
All checks passed!

1
```

**That one removed line, accounted for:** the fixture's inline base id, replaced by the module constant `_MLX_SMOKE_BASE`. It is now named once so the fixture and this assertion cannot drift onto different models — a real risk, since the assertion loads the base a second time by name. No assertion deleted or weakened.

## Limits

- **Attachment, not correctness.** This proves LoRA parameters are present after load. It does not compare generations, check the weights are the trained ones, or evaluate quality.
- **`.lora_` is a naming assumption** about mlx-lm's `LoRALinear`. Verified against mlx-lm 0.31.3; if upstream renames those attributes this goes red, which is the correct direction for a guard.
- **Apple Silicon only**, like the job it lives in. It skips elsewhere via the existing `importorskip`.

No changelog fragment — test-only, matching #647 and #653.

### DIFF
--- a/tests/test_smoke_train.py
+++ b/tests/test_smoke_train.py
@@ -108,13 +108,18 @@ output: {output_dir}
     return config_path
 
 
+#: The base the MLX smoke fixture trains. Named once so the fixture and the
+#: adapter-loadability assertion cannot drift onto different models.
+_MLX_SMOKE_BASE = "hf-internal-testing/tiny-random-LlamaForCausalLM"
+
+
 @pytest.fixture
 def mlx_sft_config_yaml(tmp_path: Path, tiny_train_data: Path) -> Path:
     """Create a one-step MLX SFT config with a public tiny Llama fixture."""
     config_path = tmp_path / "soup_mlx.yaml"
     output_dir = tmp_path / "output_mlx"
     config_path.write_text(
-        f"""base: hf-internal-testing/tiny-random-LlamaForCausalLM
+        f"""base: {_MLX_SMOKE_BASE}
 task: sft
 backend: mlx
 
@@ -234,6 +239,32 @@ def test_mlx_sft_smoke(mlx_sft_config_yaml: Path):
     output_dir = mlx_sft_config_yaml.parent / "output_mlx"
     assert (output_dir / "adapters.safetensors").exists()
     assert (output_dir / "adapter_config.json").exists()
+
+    # #23's "adapter file saved and loadable" criterion. Existence is not
+    # loadability, and the gap has a named failure mode here: per
+    # `resolve_mlx_target_keys`'s docstring (#392), a run could ship
+    # `{"keys": ["auto"]}`, `linear_to_lora_layers` would match nothing, and
+    # `load_weights(strict=False)` would drop every LoRA tensor IN SILENCE --
+    # leaving an adapter that exists, satisfies both assertions above, and
+    # produces generations bit-identical to the base model.
+    #
+    # So the check is that loading ATTACHES LoRA parameters, not merely that
+    # it does not raise. A dropped-tensor adapter loads perfectly happily.
+    from mlx.utils import tree_flatten
+    from mlx_lm import load
+
+    base_model, _ = load(_MLX_SMOKE_BASE)
+    tuned_model, _ = load(_MLX_SMOKE_BASE, adapter_path=str(output_dir))
+
+    base_keys = {key for key, _ in tree_flatten(base_model.parameters())}
+    tuned_keys = {key for key, _ in tree_flatten(tuned_model.parameters())}
+    lora_keys = {key for key in tuned_keys - base_keys if ".lora_" in key}
+
+    assert lora_keys, (
+        "the adapter loaded but attached no LoRA parameters: "
+        f"{len(base_keys)} params before, {len(tuned_keys)} after. That is the "
+        "#392 shape -- an adapter that exists and is inert."
+    )
 
 
 def test_dpo_smoke(dpo_config_yaml: Path):


Srinivasan8888:

```
1 failed, 19863 passed
```

MakazhanAlpamys:
@Srinivasan8888 — reviewed by execution. **Approved; merging as soon as the remaining matrix cells report.** This is the smallest diff in the queue and the best-argued one.

## The cherry-pick question is already settled

Your PR body flags the temporary `0f50800a` carry from #669. It is gone — the branch was force-pushed after #669 merged, and I checked rather than took your word:

```
git merge-base origin/main pr668   -> 64e0323  == origin/main HEAD
git rebase origin/main             -> "Current branch is up to date."
git cat-file -t 0f50800a           -> fatal: Not a valid object name
```

One commit, already on current main, no duplicate, no conflict. Thank you for flagging the duplication in the first place rather than letting me find it in the diff — that is the right instinct and it cost me nothing to verify.

## The assertion is the load-bearing kind

The distinction your description draws is the whole value here, so I want it on the record: `assert lora_keys` rather than "load did not raise". A dropped-tensor adapter loads perfectly happily, so a no-raise check would have been vacuous against exactly the defect #392 names — `{"keys": ["auto"]}` matching nothing, `load_weights(strict=False)` dropping every LoRA tensor in silence, and an adapter that exists, satisfies both pre-existing assertions, and generates bit-identically to the base model.

Exercised the assertion block as written against four stubbed `mlx_lm.load` outcomes:

| scenario | result |
|---|---|
| healthy adapter (3 params -> 7, incl. `.lora_a`/`.lora_b`) | passes — control |
| **the #392 shape: loads fine, `tuned_keys == base_keys`** | **KILLED** — fails naming both counts and #392 |
| adapter path absent | **KILLED** — propagates, fails for the right reason |
| adapter file corrupt (`HeaderTooLarge`) | **KILLED** — same |
| assertion neutered to `assert True`, #392 shape injected | **survives** — which is the proof that the pre-existing `.exists()` pair alone ships the inert adapter undetected |

The failure message printing both counts and naming #392 means whoever sees it red does not have to rediscover the mechanism. That is worth more than the assertion itself.

## The `-1` line, and the skip

The single removed line is the inlined base id, replaced by `_MLX_SMOKE_BASE` at module scope. Byte-identical rendered value, and it is load-bearing rather than cosmetic: your new assertion calls `load(_MLX_SMOKE_BASE)` a second time, and the constant is what stops the fixture and the assertion drifting onto different models. `hf-internal-testing/tiny-random-LlamaForCausalLM` confirmed to exist on the Hub, unauthenticated.

**On the local skip — it is not "skipped everywhere", and I checked rather than assumed.** On this Windows box the test skips (`could not import 'mlx.core'`), identically to baseline. But `mlx-smoke` runs on `macos-14` on every PR in an `.[mlx]`-only environment, and on this PR's head it reported:

```
1 passed, 2 deselected, 2 warnings in 16.64s
```

So the new assertion genuinely executed against real `mlx-lm` on real Apple Silicon. That is the evidence that matters and it exists.

Full local run is byte-for-byte identical to `origin/main`: 2 failed, 1 skipped, where the two failures are `test_sft_smoke`/`test_dpo_smoke` dying on `cannot import name 'FSDPModule'` — the declared-`torch>=2.5.0` versus `trl>=0.29` incompatibility already recorded as #651, not yours.

## No changelog fragment, correctly

`.github/pull_request_template.md` says "or this change has no user-visible impact", and `changelog.d/README.md` scopes fragments to user-visible pull requests. One test file, no shipped code. The validator agrees (17 passed).

## Two things I am recording rather than asking you to change

**`.lora_` is an mlx-lm-internal naming assumption.** If upstream renames `LoRALinear`'s attributes this goes red on a healthy adapter. You document it and argue red is the correct direction for a guard; I agree, and it would fail loudly with counts rather than silently. Staying as is.

**This proves attachment, not correctness.** LoRA parameters are present after load; nothing compares generations or checks the weights are the *trained* ones. You state that under "Limits", which is the right scope for a 2.6-second smoke assertion.

## On #23

I am merging this as `Refs #23`, not closing it. It genuinely closes the "adapter file saved and loadable" criterion, but #23's checklist also names `tests/test_mlx_integration.py` and a nightly workflow — both superseded by something stronger (per-PR `macos-14`) rather than left undone. I will say that explicitly on #23 when it closes, so the unticked boxes are not read as unfinished work.

Holding only on the nine matrix cells still pending.

MakazhanAlpamys:
Merged as `feef28a`, credited in `CONTRIBUTORS.md` (`370bacb`).

## What I ran

Real `mlx` cannot be installed on this box (Windows, x86), so I exercised the assertion block exactly as written with `mlx_lm.load` stubbed to produce each scenario's key sets. That proves the assertion **discriminates**; it does not re-verify mlx-lm's own behaviour, and I want that distinction on the record rather than glossed.

| scenario | result |
|---|---|
| healthy adapter (3 params -> 7, incl. `.lora_a`/`.lora_b`) | passes — control |
| **the #392 shape: loads cleanly, `tuned_keys == base_keys`** | **FAILS**, naming both counts and #392 |
| adapter path absent | **FAILS** — propagates, right reason |
| adapter file corrupt (`HeaderTooLarge`) | **FAILS** — same |
| assertion neutered to `assert True`, #392 shape injected | passes — the proof that the existing `.exists()` pair alone ships the inert adapter undetected |

For the half I could not stub: `mlx-smoke` on `macos-14` reported

```
1 passed, 2 deselected, 2 warnings in 16.64s
```

on this head, so the new assertion genuinely executed against real `mlx-lm` on real Apple Silicon. I checked that rather than assumed it, because "the MLX job is green" and "the MLX job ran this test" are different statements and only the second one is evidence.

Before merging I also materialised the merge against current `main` — #662 had landed in between — and ran it: fragment guard 17 passed, MLX smoke correctly skipped off-Apple, and the only two failures are the pre-existing `FSDPModule` import from the declared-`torch>=2.5.0`-against-`trl>=0.29` floor problem (#651), identical on clean `main`. Zero file overlap with #662.

## What was good

**The assertion is `assert lora_keys`, not "load did not raise".** That is the whole PR. A dropped-tensor adapter loads perfectly happily, so a no-raise check would have been vacuous against precisely the defect #392 describes — an adapter that exists, passes both old assertions, and generates bit-identically to the base model. Choosing the strong form when the weak one would have looked equally green is the judgement call this repo keeps needing.

**The failure message does the next person's work.** It prints the parameter counts either side and names #392, so whoever hits it red is handed the mechanism instead of a bare `AssertionError`.

**The `-1` line was load-bearing, not cosmetic.** Hoisting the base id to `_MLX_SMOKE_BASE` matters because your assertion loads that same id a second time; the constant is what stops the fixture and the assertion drifting onto different models. I checked the rendered value is byte-identical.

**And you scoped it honestly** — the PR says this proves attachment, not that the weights are the trained ones, and it says the `.lora_` prefix is an mlx-lm-internal naming assumption that would go red if upstream renames it. I agree red is the correct direction there and left it as is.

## On #23

I merged this as `Refs`, not `Closes`, and #23 stays open — but not because anything here is unfinished. It closes the "adapter file saved and loadable" criterion outright. The issue's other items are **superseded rather than outstanding**: `tests/test_mlx_integration.py` and a nightly workflow were the plan, and a per-PR `macos-14` job is strictly stronger than a nightly one. I have said so on the issue so the unticked boxes are not read as work nobody did.

Third merge of yours in this queue pass, and the second where the interesting part was choosing a test that could actually fail.

PR #669, opened 2026-09-04
fix(tests): unbreak main — a credit assertion pinned to a released changelog fragment

Srinivasan8888:
**`main` is red and every open PR in the repo inherits it.** This unblocks it. Test-only, one file.

Opening directly rather than claiming first, because the tree is broken for everyone — say the word if you would rather I had waited.

## The break

`main` at `f07e07ed release: v0.74.0` fails **all nine `test` matrix cells**. The two commits before the release are green:

```
33859364205  completed/FAILURE  f07e07ed  release: v0.74.0
33844139364  completed/success  e6ad7b81  docs(contributors): credit Nurkhan Esenbek's Lambda...
33844072910  completed/success  0405611a  feat(cloud): add the Lambda Cloud training backend
```

```
FileNotFoundError: [Errno 2] No such file or directory: 'D:\a\Soup\Soup\changelog.d\0.73.3\603.added.md'
```

Reproduced locally on a clean `main` checkout — `1 failed in 0.20s`.

## Cause, and why it was inevitable

```python
changelog = (root / "changelog.d" / "0.73.3" / "603.added.md").read_text(encoding="utf-8")
assert "(#602 by @Amix29 in #603)" in changelog
```

The test reads a changelog **fragment** by hard-coded path. Fragments are assembled into `CHANGELOG.md` and **deleted at release** — that is the entire point of `changelog.d`. So the assertion held only while 0.73.3 was unreleased, and was guaranteed to break at whatever the next release turned out to be. **This is not a v0.74.0 problem**; any release would have fired it.

The credit was not lost, only relocated — `CHANGELOG.md:326`, verbatim.

## The fix

Search **both** homes, so the assertion survives the boundary in either direction instead of being correct on only one side:

```python
haystacks = [(root / "CHANGELOG.md").read_text(encoding="utf-8")]
haystacks += [f.read_text(encoding="utf-8") for f in sorted((root / "changelog.d").rglob("603.*.md"))]
assert any(credit in text for text in haystacks), ...
```

Verified in all three states, with a control so it is not vacuous:

| scenario | verdict |
|---|---|
| POST-release: credit in `CHANGELOG.md`, no fragment | **PASS** |
| PRE-release: credit in fragment only | **PASS** |
| **CONTROL:** credit absent from **both** → must fail | **FAIL** — guard has teeth |

`CHANGELOG.md` restored byte-identical after the run (md5 verified).

The failure message now says the credit "has been lost, not merely relocated", so the next person to see it red knows which of the two situations they are in.

## Swept for the same shape

Only one other test touches `changelog.d`:

```
tests/test_issue627_unknown_config_keys.py:476:  + sorted((root / "changelog.d").rglob("*.md"))
```

It uses `rglob`, which tolerates the directory being empty, so it is **not** exposed. No other test hard-codes a fragment path.

## Verification

```
$ .venv/bin/python -m pytest tests/test_issue602_qwen4_ple_streaming.py -q -o addopts=
35 passed

$ .venv/bin/python -m ruff check tests/
All checks passed!

4
```

**Those 4 removed lines, accounted for:** the two-line hard-coded `read_text` call and the two original `assert` lines, replaced by the both-homes lookup and a message-carrying assertion. The `benchmark_index` assertion is unchanged. Nothing weakened — the test now fails in strictly more cases than before (it previously could not distinguish "credit lost" from "file moved"; it crashed either way).

## Note for the four PRs of mine that are red

#662, #663, #665, #668 are all failing on this and nothing else. I have already rebased them onto v0.74.0 and moved their changelog fragments to `changelog.d/0.74.0/`, per `validate_fragments`'s own instruction — they should go green once this lands.

### DIFF
--- a/tests/test_issue602_qwen4_ple_streaming.py
+++ b/tests/test_issue602_qwen4_ple_streaming.py
@@ -1156,9 +1156,21 @@ def test_qwen4_gate_record_and_changelog_are_discoverable_and_credited():
 
     root = Path(__file__).parents[1]
     benchmark_index = (root / "benchmarks" / "README.md").read_text(encoding="utf-8")
-    changelog = (root / "changelog.d" / "0.73.3" / "603.added.md").read_text(
-        encoding="utf-8"
-    )
+
+    # The credit moves at release time: `changelog.d/<baseline>/603.added.md`
+    # is ASSEMBLED into CHANGELOG.md and then deleted. Reading the fragment by
+    # a hard-coded path made this test pass only while 0.73.3 was unreleased,
+    # and it went red on every platform the moment v0.74.0 shipped. Search both
+    # homes so the assertion survives the boundary in either direction.
+    credit = "(#602 by @Amix29 in #603)"
+    haystacks = [(root / "CHANGELOG.md").read_text(encoding="utf-8")]
+    haystacks += [
+        fragment.read_text(encoding="utf-8")
+        for fragment in sorted((root / "changelog.d").rglob("603.*.md"))
+    ]
 
     assert "gate-qwen4-ple-m4-max.md" in benchmark_index
-    assert "(#602 by @Amix29 in #603)" in changelog
+    assert any(credit in text for text in haystacks), (
+        f"{credit!r} found in neither CHANGELOG.md nor any changelog.d/**/603.*.md; "
+        "the #602 credit has been lost, not merely relocated"
+    )


Srinivasan8888:
**Confirmation that this is the whole fix, not part of one.** Full suite on a clean `main` checkout at `f07e07ed` (the release commit), no filters:

```
$ .venv/bin/python -m pytest tests/ -q -o addopts=
1 failed, 19978 passed, 107 skipped, 196 warnings in 648.66s (0:10:48)
```

**One failure in 19,979 tests**, and it is the one this PR fixes. The release did not break anything else — so the nine red matrix cells across every open PR in the repo collapse to this single hard-coded path.

Same branch, same command, after the fix: `35 passed` in that file, and the three-state verification in the PR body (post-release / pre-release / control) holds.

Worth noting the failure is **platform-independent** — all nine cells failed identically, on Windows, macOS and Ubuntu across 3.10/3.11/3.12 — which is consistent with a missing file rather than anything environmental, and is why I did not chase it as a flake.

MakazhanAlpamys:
@Srinivasan8888 — **you were five hours ahead of me and I steamrolled you. That is my error, not a coincidence, and I want it on the record before anything else.**

The timeline, from the API rather than memory:

| when (UTC) | what |
|---|---|
| 09:52 | I push `f07e07e release: v0.74.0`, main goes red on all nine cells |
| **12:14** | **you open #669 with the diagnosis and the fix** |
| 17:13 | I push `766f348`, my own repair, **without checking the open PR list** |

Main was broken for five hours because of my release commit, you fixed it, and I then fixed it again as if nobody had. The rule I ask everyone else to follow — check for overlapping work *before* writing, and search PR titles, not just issue threads — is the exact rule I skipped. `gh pr list --state open` would have taken two seconds and shown me #669 at the top.

## Your diagnosis was right, and it was right first

> the credit moves at release time … reading the fragment by a hard-coded path made this test pass only while 0.73.3 was unreleased

That is precisely the defect: a fragment is *supposed* to be consumed at release (#487/#490), so the test made a routine, designed event fail a check whose subject is the credit rather than where the credit is stored. You also opened directly instead of claiming first and said why — with `main` red for everyone, that is the correct call and I would have made the same one.

## One part of your version is better than mine, and it is now on main

Your search glob is `603.*.md`; mine was `603.added.md`. Mine hard-codes the **category** as well as the number, so a fragment re-filed as `603.fixed.md` would drop out of the search and the test would quietly depend on `CHANGELOG.md` alone. Yours does not have that hole.

I have taken it, as `450442f`, with the commit body saying plainly that it is yours and that you were first:

```
test(qwen4): widen the credit search to any 603 fragment category

Adopted from #669 by @Srinivasan8888, who diagnosed and fixed this five
hours before I pushed my own repair without checking the open PR list.
```

## Closing this, and what you are credited for

Closing #669 as superseded **only because the same repair is already on main under my commit** — not because anything in it was wrong. You get the CONTRIBUTORS.md entry for it, and it records the sequence honestly: you found it first, you fixed it first, and the version that shipped carries your improvement.

What I verified on the merged result, so this is not just an apology: the test passes, and **mutating the credit out of the assembled `CHANGELOG.md` fails it** — so neither of us traded a red path-pinned test for a vacuous green one, which was the obvious wrong fix here.

v0.74.0 is tagged and released. Thank you for unblocking the tree while I was still reading logs.

Srinivasan8888:
No harm done, and I would not want the apology to be the headline — **the tree is fixed, which was the entire point of #669.** Whose commit did it matters less than the five hours it was red.

For what it is worth, the duplication was cheap on my side: the diagnosis was ten minutes, the fix one line, and most of the cost was the full-suite run that proved `1 failed, 19978 passed` — which was worth having regardless of who landed the patch, because it ruled out the release breaking anything else.

**Your version is better than mine in a way worth naming**, since you were generous about the glob. Both of us search the two homes, but your failure messages carry the searched paths:

```python
assert texts, "neither CHANGELOG.md nor a 603 fragment is readable"
...
f"credit; searched {[str(p.relative_to(root)) for p in sources]}"
```

Mine said the credit "has been lost, not merely relocated" and left the reader to work out where I had looked. Yours tells them. And the `assert texts` guard is a real addition — my version would have gone green against an empty source list if both files vanished, which is precisely the vacuous-pass shape we were both trying to avoid. So the merged result is better than either of ours alone.

**Housekeeping, so you do not hit it in review.** When #669 stalled I cherry-picked its commit onto my four open PRs so they could be reviewed on their own merits rather than a borrowed break — flagged in a comment on each at the time. That would now have conflicted with `766f348`. I have already dropped the duplicate from all four and rebased them onto `64e0323`:

```
#662 feat(recipes): add deepseek-v4-flash-dpo recipe (#275)
#663 bench(mlx): publish the MLX SFT run and harness for an 8 GB M1 (#23)
#665 feat(mlx): drive the live display and tracker from the MLX trainer (#23)
#668 test(mlx): assert the smoke adapter LOADS, not merely that it exists (#23)
```

None of them touches `test_issue602_qwen4_ple_streaming.py` any more — verified, zero files — so each is back to only its own change. The qwen4 test passes on all four against *your* fix from main.

**One thing I would push back on gently:** I do not think the process failure here is really the missed `gh pr list`. It is that a release consumed `changelog.d/` and nothing checked whether a test read from there — the same class as #661, where `validate-recipes` validates schema but not that `base:` resolves. A guard that greps `tests/` for a hard-coded `changelog.d/<version>/` path at release time would have caught this before the tag, and would not depend on anyone remembering to look. Happy to write it if you want it; happy to drop it if you would rather not add a release-time check.

PR #670, opened 2026-09-04
fix(data): drop multimodal rows with a non-dict message instead of aborting the dataset load

abdulwaarith0:
## What does this PR do?

`_convert_multimodal` iterates `row["messages"]` and calls `msg.get("content")` without first checking that `msg` is a dict:

```python
for msg in messages:
    content = msg.get("content")   # AttributeError if msg is not a dict
```

So a row like `{"messages": ["hello"]}` — a bare string, `None`, an int, or a nested list where an object belongs — raises `AttributeError: 'str' object has no attribute 'get'`.

That matters because of the contract in `format_to_messages`:

```python
except (KeyError, TypeError, IndexError, ValueError):
    return None
```

`AttributeError` is in none of those, so the exception escapes the converter rather than routing to the drop path. **One malformed line in a large JSONL file aborts the entire dataset load** instead of being skipped like every other malformed row.

This is clearly not the intended behaviour — `_require_str_content`'s own docstring states the design directly: raising should route "the row to `format_to_messages`'s drop path instead of silently corrupting the dataset". The converter already guards the *parts* inside a message (`"multimodal content part must be a dict"`, line 599); the message itself was simply missed.

The fix raises `ValueError` for a non-dict message, which the existing drop path already understands, and mirrors the wording of the part-level guard directly above it:

```python
for msg in messages:
    if not isinstance(msg, dict):
        raise ValueError("multimodal message must be a dict")
    content = msg.get("content")
```

## Reproduction

Before the change, all four escape as uncaught `AttributeError`:

| Row | Before | After |
|---|---|---|
| `{"messages": ["hello"]}` | `AttributeError: 'str' object has no attribute 'get'` | dropped (`None`) |
| `{"messages": [42]}` | `AttributeError: 'int' object has no attribute 'get'` | dropped (`None`) |
| `{"messages": [["role","user"]]}` | `AttributeError: 'list' object has no attribute 'get'` | dropped (`None`) |
| `{"messages": [None]}` | `AttributeError: 'NoneType' object has no attribute 'get'` | dropped (`None`) |

Valid rows are unaffected — typed content-part lists and the plain-string back-compat path both convert exactly as before, and an invalid part type still drops.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Tests

## Tests

Added `test_multimodal_converter_non_dict_message_is_dropped`, parametrized over the four shapes above, next to the existing multimodal converter tests in `tests/test_v0420.py`.

The test is a real regression test, not one that passes either way — with the two-line guard reverted it fails on all four parameters with the `AttributeError` above, and passes with it applied.

```
tests/test_v0420.py: 150 passed, 1 skipped
```

Four unrelated failures exist on this machine (`test_infer.py::test_generate_fallback_formats_roles`, `test_v0531_142.py::test_no_double_quant_flag_reaches_merge_4bit`, `test_v0713.py::test_write_annex_doc_pdf_format`, `test_v07200.py::test_dtype_is_converted`) plus 17 collection errors from missing optional deps (`numpy`); all of them reproduce identically on a clean checkout of `main` with no changes applied, so they are environmental and unrelated to this diff.

## Checklist

- [x] `pytest tests/ -v` passes (see note above on pre-existing environmental failures)
- [ ] Updated relevant docs — not needed; no user-facing interface changed
- [x] No breaking changes — rows that previously crashed the load are now dropped, which is the documented contract for malformed rows
- [x] Added a changelog fragment (`changelog.d/0.73.3/670.fixed.md`)


MakazhanAlpamys:
@abdulwaarith0 — welcome, and thanks. Reviewed by execution rather than by reading the description. **This is correct, complete for the class of bug it names, and the test is the strong kind rather than the easy kind.** I have pushed two mechanical commits to your branch myself (detailed at the bottom) so this can go in without costing you a round trip.

## Verified by execution

Isolated worktrees for both `origin/main` and your branch, `PYTHONPATH` pinned to each (unpinned, this repo's editable install resolves to the main tree and every mutation is a silent no-op).

Unit level, all four shapes:

```
MAIN   ['hello'] / [42] / [['role','user']] / [None]  ->  AttributeError, uncaught
PR670  all four                                        ->  None (dropped)
```

**The consequence you claim is the one I most wanted to doubt, and it holds.** A 5-row JSONL with one bad row among four good ones, through the real `load_dataset`:

```
MAIN : RAISED AttributeError
       data/loader.py:173 -> formats.py:129 -> formats.py:589   content = msg.get("content")
PR670: OK -> 4 train rows
```

`_format_rows` at `loader.py:171-178` has no `try/except`, so the exception escapes `load_dataset` entirely. One bad line really does take the whole dataset with it.

## The mutation that matters

| mutation | result |
|---|---|
| delete the two-line `isinstance` guard | **KILLED** — all four parameters |
| `raise ValueError` -> `raise TypeError` | SURVIVED — an equivalent mutant, see below |
| `raise ValueError` -> `continue` | **KILLED** — all four parameters |

The third row is the one that makes this a good test rather than a passing one. Under `continue`, the converter returns `{'messages': ['hello']}` — the malformed message survives verbatim into the dataset. That is *silent corruption*, exactly the failure `_require_str_content`'s docstring warns about, and your test distinguishes it from a clean drop because it asserts `is None` rather than merely asserting no exception. A weaker test would have passed there.

The `TypeError` mutant surviving is not a gap: `TypeError` is in `format_to_messages`'s caught tuple, so both spellings produce byte-identical behaviour through the public API and nothing could kill it without asserting a private detail. `ValueError` is still the right choice — `_convert_input_output` at `formats.py:549` uses the identical shape for the identical kind of check.

Zero test lines removed (`git diff origin/main...pr670 -- tests/ | grep -c "^-[^-]"` -> **0**), `tests/test_v0420.py` 147 -> **151 passed**, and `pytest -k "format or multimodal or vision"` -> 627 passed, ruff clean.

## Two things I checked that your description does not claim, and you should know

**Your fix is complete, not a one-spot patch.** I scanned every converter for element access in a message loop and then verified by execution across four bad shapes and ten format paths. `sharegpt`, `llava`, `sharegpt4v`, `input_output` and `tool-calling` all drop correctly already (they subscript, and `TypeError` *is* caught). `multimodal` was genuinely the only one leaking an uncaught `AttributeError`.

**But `chatml`, `audio` and `video` pass a non-dict message straight through** — `format_to_messages({"messages": ["hello"]}, "chatml")` returns `{'messages': ['hello']}` on both branches. That is the silent-corruption half of the same contract, and `chatml` is far more reachable than `multimodal` since it is what `detect_format` returns for a bare `{"messages": [...]}`. I am deliberately **not** holding this PR for it — widening the diff would make it harder to verify, and your scope is right. I have filed it separately and credited the discovery to this review.

One honest qualification for the record: `detect_format` never returns `"multimodal"`, so reaching this path needs an explicit `format: multimodal` in the config. That is a supported schema literal, so the bug is real and reachable — just narrower than "any JSONL file". Worth stating precisely rather than letting the changelog overclaim.

## Numbers in your description I could not reproduce

You reported `tests/test_v0420.py: 150 passed, 1 skipped`; I measured **151 passed, 0 skipped**. Harmless environment difference, but the figures do not match as written, so I would rather say so than quietly accept them.

## What I pushed to your branch

`maintainerCanModify` is on and both of these are mechanical, so rather than send you back for them:

1. **`git mv changelog.d/0.73.3/670.fixed.md changelog.d/0.74.0/670.fixed.md`.** v0.74.0 was tagged while your PR was open, `scripts/assemble_changelog.py` consumed the `0.73.3/` directory, and `test_repository_fragments_match_the_newest_release` then fails in every matrix cell. This is my release timing, not anything you did. Verified: fails before the move, 17 passed after.
2. **Added `by @abdulwaarith0` to the fragment's `(#670)`.** Project convention for outside work — v0.74.0's own audit found 56 of 116 entries missing their author, so I am now applying it at merge time rather than at release time.

Both are fast-forward commits on top of your `579e170`; your history is not rewritten. I then updated the branch onto current `main` and released the CI runs — first-time-contributor workflows need manual approval here, so if a later push of yours sits idle, ping me.

Merging once the checks come back green.

MakazhanAlpamys:
Merged as `218eae5`, and you are now in `CONTRIBUTORS.md` (`65f8930`). Welcome — this is a strong first contribution and I want to be specific about why rather than just say so.

## What I ran before merging

Isolated worktrees for `origin/main` and your branch, `PYTHONPATH` pinned to each. Unpinned, this repo's editable install resolves to the main tree and every mutation is a silent no-op, so that step is the difference between real results and imaginary ones.

**Mutations:**

| mutation | result |
|---|---|
| delete the two-line `isinstance` guard | **KILLED** — all four parameters |
| `raise ValueError` -> `raise TypeError` | survived — an equivalent mutant, see below |
| **`raise ValueError` -> `continue`** | **KILLED** — all four parameters |

I ran that third one twice, once through the review and once by hand, because it is the row that decides whether this is a good test or merely a passing one. Under `continue` the converter returns `{'messages': ['hello']}` — the malformed message survives verbatim into the dataset. Your assertion is `is None`, so it catches that. An assertion of "no exception raised" would not have, and that is the silent-corruption case `_require_str_content`'s own docstring warns about. You wrote the strong form when the weak one would have looked equally green.

The `TypeError` mutant surviving is not a gap: `TypeError` is in `format_to_messages`'s caught tuple, so both spellings are byte-identical through the public API and nothing could kill it without asserting a private detail. `ValueError` is still the right pick — `_convert_input_output` at `formats.py:549` uses the identical shape for the identical kind of check.

**The impact claim was the one I most wanted to doubt, and it holds.** A 5-row JSONL with one bad row, through the real `load_dataset`:

```
MAIN : RAISED AttributeError
       data/loader.py:173 -> formats.py:129 -> formats.py:589   content = msg.get("content")
PR670: OK -> 4 train rows
```

`_format_rows` at `loader.py:171-178` has no `try/except`, so the exception really does escape `load_dataset` entirely.

Tests 147 -> **151 passed** in `test_v0420.py`, 627 passed across `-k "format or multimodal or vision"`, ruff clean, **zero** removed test lines. After #662 and #668 landed I re-materialised the merge against the new `main` and ran it again — 168 passed — because "green on an older base" and "green on the base it will land on" are different statements, and one of the other PRs in this queue is currently blocked by exactly that distinction.

## Two things I checked that you did not claim

**Your fix is complete for the class it names, not a one-spot patch.** I scanned every converter for element access in a message loop and then verified by execution across four bad shapes and ten format paths. `sharegpt`, `llava`, `sharegpt4v`, `input_output` and `tool-calling` already drop correctly, because they subscript and `TypeError` **is** caught. `multimodal` was genuinely the only one leaking an uncaught `AttributeError`.

**`chatml`, `audio` and `video` pass a non-dict message straight through** — `format_to_messages({"messages": ["hello"]}, "chatml")` returns `{'messages': ['hello']}` on both branches. That is the silent half of the same contract, and `chatml` is far more reachable than `multimodal` since it is what `detect_format` returns for a bare `{"messages": [...]}`. I deliberately did **not** hold your PR for it — widening the diff would have made it harder to verify and your scope was right — so it is filed as #676 with the discovery credited to this review.

## One correction to the record, in your favour and against your description

`detect_format` never returns `"multimodal"`, so reaching this path needs an explicit `format: multimodal` in the config. It is a supported schema literal, so the bug is real and reachable — just narrower than "any JSONL file". I put that qualification in the merge commit rather than letting the changelog overclaim. Your description was not wrong about the mechanism, only about the blast radius.

Also: the figures did not quite match. You reported `150 passed, 1 skipped`; I measured `151 passed, 0 skipped`. Harmless environment difference, but I would rather say so than quietly accept numbers I could not reproduce.

## What I pushed to your branch, and why

Two mechanical commits, disclosed at the time: moving the changelog fragment from `changelog.d/0.73.3/` to `0.74.0/`, and adding your `by @abdulwaarith0` credit to it. v0.74.0 was tagged while your PR was open, which consumed the old directory — my release timing, not anything you did. Both were fast-forward commits on top of your work; your history was not rewritten.

If you want another, #676 is the direct sequel to this one and you have already done the hard thinking for it.

MakazhanAlpamys:
Correction to my own merge note, found while checking my work rather than by you having to.

I said I had pushed two commits to your branch: the changelog fragment move, and adding your `by @abdulwaarith0` credit to it. **Only the first actually landed.** I edited the file after `git mv` had already staged the rename, then committed without `-a`, so the rename went in and the credit edit stayed in my working tree. The merged fragment read `(#670).` with no author.

Fixed on `main` as `ef2e02b`; it now reads `(#670 by @abdulwaarith0).`

Nothing else in the merge was affected — the code, the tests and your `CONTRIBUTORS.md` entry all landed as described, and I re-verified each of those against `origin/main` rather than assuming. But the credit is the one thing in that list that exists purely to give you attribution, so it is the worst one to have quietly dropped while claiming otherwise.

abdulwaarith0:
Thanks for the thorough review and the merge. The mutation breakdown — the `continue` case especially — is a genuinely useful bar to be held to, and I'll carry the 'assert the drop, not just the absence of an exception' discipline forward.

Good catch on the blast-radius wording and the test-count mismatch too; I'll pin figures to the base the change will land on next time rather than the base I branched from. Appreciated you fixing the dropped author credit as well.

If the validator half of #676 is still open — `soup data validate` green-lighting rows the loader drops — I'd be glad to take it; it's a separate code path from the converters, so #679 won't close it.

PR #673, opened 2026-09-04
docs(readme): lead the install with pipx, and name PEP 668

### LINKED ISSUE #671: README install fails on Debian/Ubuntu: 'pip install soup-cli' hits PEP 668 externally-managed-environment, and pipx is never mentioned
## What happens

The first command in the README fails on a stock Debian 12 / Ubuntu 24.04 box with the system Python. Both installation lines do:

```
$ pip install soup-cli
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    ...
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider.
hint: See PEP 668 for the detailed specification.
```

```
$ pip install git+https://github.com/MakazhanAlpamys/Soup.git
error: externally-managed-environment
[same]
```

Python 3.12, no venv active, nothing unusual about the box.

## Why

This is [PEP 668](https://peps.python.org/pep-0668/), not a Soup defect. Debian-derived distros ship an `EXTERNALLY-MANAGED` marker in the system Python so that `pip` refuses to write into a `site-packages` that `apt` also manages. It has been the default on Debian 12 and Ubuntu 23.04 onward, so it is now the majority experience for a new Linux user rather than an edge case.

The report is a documentation one: `README.md` lines 176 to 185 give four bare `pip install` invocations and no mention of PEP 668. The only virtualenv reference in the README is at line 353, inside a troubleshooting note about `soup version` disagreeing with `pip show`, which is not where someone stuck on line 176 will look.

## Why pipx is the right primary instruction here, not just a workaround

Soup declares a console script:

```toml
[project.scripts]
soup = "soup_cli.cli:run"
```

That makes it an application rather than a library, which is precisely the case pipx and `uv tool` exist for: a private virtualenv per application, with just the `soup` command linked onto `PATH`. Nothing is shared with the system Python, so PEP 668 does not apply.

I verified the isolated route resolves and runs against the published wheel, without installing anything onto the host:

```
$ uv tool run --from soup-cli soup --version
Usage: soup [OPTIONS] [COMMAND] [ARGS]...
...
│ No such option '--version'. Did you mean '--verbose'? │
```

The wheel resolved, the `soup` entry point was created, and it executed. (`--version` is not a flag; `soup version` is the subcommand. That is a separate matter and not part of this report.)

## Suggested change

Scoped to `README.md` "1. Install", keeping every existing line and reordering rather than rewriting:

1. Lead with `pipx install soup-cli` and `uv tool install soup-cli` for people who want the CLI.
2. Keep the current `pip install` lines, stated as the right choice when you are already inside a virtualenv, which is also the case in a Colab or Docker image where the existing lines work fine today.
3. One sentence naming PEP 668 and the `externally-managed-environment` error text, so the error is searchable straight into your own README.

Extras work unchanged on the isolated path, including the double-quote rule already documented directly below:

```
pipx install "soup-cli[train]"
uv tool install "soup-cli[train]"
```

One honest caveat worth a clause rather than a section: pipx isolation is right for driving Soup from the command line, and a plain venv is still right for anyone importing `soup_cli` next to their own code.

## Out of scope, flagging rather than folding in

`CONTRIBUTING.md` line 22 has `pip install -e ".[dev]"`, which fails the same way for a first-time contributor on the same box. I have deliberately left it out. A contributor setup note is a different audience and a different sentence from a user install, and I would rather not widen a README fix into a CONTRIBUTING one without you ruling on it. Say the word and it is a two-line addition to the same PR.

The five `docs/*.md` pages that also carry `pip install soup-cli` are likewise untouched, on the same reasoning: the README is where a new user lands.

## Acceptance criteria

- [ ] A reader on Debian 12 or Ubuntu 24.04 can get a working `soup` command by following the README top to bottom, without already knowing what a venv is.
- [ ] The literal string `externally-managed-environment` appears in the README, so the error text someone pastes into a search engine matches.
- [ ] The existing `pip install` lines survive, with the condition under which they work stated.
- [ ] No change to `pyproject.toml`, packaging, or any command behaviour. Documentation only.

## Claim

Taking this one, per the "Claiming an Issue" rule in CONTRIBUTING. It is documentation only and needs no special hardware: I hit it on the box in the transcript above and can verify the corrected instructions on the same box.

## How it was found

Reported from a plain first-run attempt, not from tooling. Unrelated to the graph analysis I have been running on this repo, though for disclosure that tooling is [codegraph](https://github.com/swalla02/codegraph).

swalla02:
Closes #671.

`pip install soup-cli`, the first command in the README, fails on Debian 12 and Ubuntu 23.04 or later:

```
$ pip install soup-cli
error: externally-managed-environment
...
hint: See PEP 668 for the detailed specification.
```

That is [PEP 668](https://peps.python.org/pep-0668/) rather than a Soup defect, but the README is where it lands on someone, and the README did not mention it. Four bare `pip install` lines, and the only virtualenv reference sitting at line 353 inside a troubleshooting note about `soup version` disagreeing with `pip show`, which is not where a reader stuck on line 176 will look.

## What changed

`README.md`, "1. Install", only. Nothing else in the tree is touched.

- **Lead with `pipx install soup-cli` and `uv tool install soup-cli`.** Soup declares `soup = "soup_cli.cli:run"` under `[project.scripts]`, so it is an application, not a library, and a per-application virtualenv with the command on `PATH` is what pipx and `uv tool` exist for. PEP 668 does not apply to them.
- **Every existing `pip install` line survives**, moved under a sentence naming when it is the right choice: already inside a virtualenv, a Colab notebook, or a Docker image. Those readers were never broken and their instructions did not change.
- **A callout carrying the literal string `externally-managed-environment`**, so the error text someone pastes into a search engine matches the project's own README, plus the `python3 -m venv` route for people who would rather not add a tool.
- **One clause on when not to use pipx:** if you also want to `import soup_cli` next to your own code, pipx isolation is the wrong shape and plain `pip` in a venv is right.

The double-quote callout directly below is untouched and still applies, since the extras spelling is identical on both paths.

## What I verified, and what I did not

Verified end to end, in a contained `PIPX_HOME` so nothing landed on the host:

```
$ pipx install soup-cli
  installed package soup-cli 0.73.3, installed using Python 3.11.15
  These apps are now available
    - soup
$ soup version
soup v0.73.3
```

```
$ pipx install "git+https://github.com/MakazhanAlpamys/Soup.git"
  installed package soup-cli 0.74.0, installed using Python 3.11.15
$ soup version
soup v0.74.0
```

`uv tool` resolves and runs the published wheel the same way (`uv tool run --from soup-cli soup version`).

Not run end to end: `pipx install "soup-cli[train]"`. I resolved it instead, with `uv pip install --dry-run "soup-cli[train]"`, which completes cleanly against the full training stack. The extras syntax is unchanged from the lines already in the README, so the risk here is the download size rather than the spelling, and I would rather say that than imply I installed torch to check.

`ruff check src/soup_cli/ scripts/ tests/` passes. I did not run `pytest tests/` because it needs `[dev]`, which self-references `[train]` and pulls the whole training stack; this PR changes no Python. Happy to run it if you would rather have the number.

## Deliberately out of scope

`CONTRIBUTING.md` line 22 has `pip install -e ".[dev]"` and fails identically for a first-time contributor on the same box. I left it alone, and the five `docs/*.md` pages that also carry `pip install soup-cli` too. A contributor setup note is a different audience from a user install, and I did not want to widen a README fix without you ruling on it. Both are small additions to this branch if you want them.

## How it was found

A plain first-run attempt on a fresh Debian box, not from tooling.


swalla02:
Pushed as `1ae63cc`, per your ruling on #675 that these belong here rather than there. Thanks for merging main in first; I have synced to `822aa6e`.

Two files, and I did **not** do the blanket pass across all five `docs/*.md` pages. Reasoning, so you can overrule it in one line if you disagree:

**`CONTRIBUTING.md`** now writes the venv step out instead of assuming it:

```bash
python3 -m venv .venv
source .venv/bin/activate          # Windows: .venv\Scripts\activate
pip install -e ".[dev]"
```

with a sentence naming PEP 668. I deliberately did not offer `pipx` for this one: an editable checkout has to be importable by the test suite, and pipx isolation is precisely what stops that. Offering it there would be the wrong advice in a place where the reader cannot tell.

**`docs/models.md`** gets one sentence above the extras table, since that is the page the README links to for extras. The `pip` rows work verbatim inside a virtualenv, a Colab notebook or a Docker image, and outside one the substitution is `pipx` or `uv tool` with the extras spelling unchanged.

**The other four pages I left alone**, and I think that is the right call rather than laziness. Looking at what those occurrences actually are:

| page | occurrence | what it is |
|---|---|---|
| `docs/adapters-and-governance.md` | 4x, e.g. `needs pip install soup-cli[carbon]` | extras hint mid-prose |
| `docs/commands.md:235` | inside a command table cell | extras hint |
| `docs/backends-and-ops.md:870,874` | quoting Soup's own runtime error message | extras hint |
| `docs/data.md:821` | `pip install soup-cli[remote]` | extras hint |

None is an entry point. Every one of them is read by somebody who already has `soup` on their `PATH` and has therefore already resolved the PEP 668 question, and `backends-and-ops.md:870` is a verbatim quote of what Soup prints at runtime, so changing it would make the doc disagree with the program. A PEP 668 callout in each would be four repetitions aimed at readers who cannot hit the error.

If you would rather have consistency than that argument, say so and I will do the pass; it is mechanical.

I also re-ran `tests/test_v07137.py` (the single-quote docs scanner) against both edited files: 4 passed. And the new `../README.md#1-install` link resolves against `### 1. Install` at `README.md:172`.

MakazhanAlpamys:
`1ae63cc` is right, including the part you declined to do. Approved on substance — holding only for the checks, which I have released.

## The judgement call you made is the correct one

> I deliberately did not offer `pipx` for this one: an editable checkout has to be importable by the test suite, and pipx isolation is precisely what stops that.

Yes. And it is worth saying why this is more than a technicality: `pip install -e ".[dev]"` under pipx would appear to succeed, and then `pytest` would import `soup_cli` from wherever it could find it — which on a machine that already has Soup installed is the *other* copy. This repo has a documented, expensive version of that failure: every mutation run against a checkout resolves to the main tree unless `PYTHONPATH` is pinned, and unpinned mutation testing reports green while changing nothing. Recommending pipx in `CONTRIBUTING.md` would have manufactured that trap for every new contributor. Declining it there while leading with it in the README is exactly the right split, and you reasoned to it unprompted.

The four pages you left alone are also the right call. Those occurrences are extras hints mid-prose (`needs pip install soup-cli[carbon]`) and quoted error text — they describe a package name, not an install procedure, and rewriting them would add noise where nobody is stuck. `docs/models.md` is different because it is the page the README hands the reader for extras, which is why one sentence there is worth it and four more elsewhere are not.

## Verified

Materialised the merge against current `main` — which moved three times today — and ran the docs surface: **47 passed**. `CONTRIBUTING.md` auto-merges cleanly. The changelog fragment is in `changelog.d/0.74.0/`, correctly numbered for the PR, and now carries `by @swalla02`.

I also confirmed the technical premise myself rather than taking it on trust: Soup declares `soup = "soup_cli.cli:run"` under `[project.scripts]`, so it genuinely is an application rather than a library, which is what makes `pipx` and `uv tool` the right lead. And `all = ["soup-cli[train,serve,ui,data,mcp]"]` resolves, so `pipx install "soup-cli[all]"` is a real command and not a plausible-looking one.

## One thing that is mine to fix, not yours

This takes `README.md` from 480 lines to 504, and I keep a soft ceiling around 500 with a standing rule that when prose pushes past it, a section moves into `docs/` rather than the ceiling moving. That rule lives in a maintainer file you cannot see, so it is not something you violated — and I am not asking you to trim a section that is doing real work. I will displace the Troubleshooting block into `docs/` myself after this merges, which is the next candidate on that list anyway.

Mentioning it only so that if you see me editing the README right after your PR lands, you know it is bookkeeping and not a walk-back.

Merging when the matrix reports.

MakazhanAlpamys:
Merged as `dad98ef`, credited in `CONTRIBUTORS.md` (`3a556c4`). Second merge of yours today.

## Verified

Re-materialised the merge against current `main` — which had moved again since your checks ran — and ran the docs surface: **68 passed**, `CONTRIBUTING.md` auto-merges cleanly. I also checked the two technical premises myself rather than taking them on trust: `[project.scripts]` really does declare `soup = "soup_cli.cli:run"`, which is what makes pipx the right lead rather than a fashion, and `all = ["soup-cli[train,serve,ui,data,mcp]"]` resolves, so `pipx install "soup-cli[all]"` is a real command and not a plausible-looking one.

## The best decision in this PR is a thing you did not do

Declining to recommend pipx for the contributor setup, because an editable checkout has to be importable by the test suite. That is right, and the consequence is worse than it first sounds: `pip install -e ".[dev]"` under pipx would *appear* to succeed, and then `pytest` would import `soup_cli` from wherever it could find it. On a machine that already has Soup installed, that is the other copy. This repo fights exactly that failure — every mutation run against a checkout resolves to the main tree unless `PYTHONPATH` is pinned, and unpinned mutation testing reports green while changing nothing. Recommending pipx in `CONTRIBUTING.md` would have manufactured that trap for every new contributor, and you reasoned to it unprompted.

Leaving the other four docs pages alone was equally right. Those occurrences are extras hints mid-prose and quoted error text — they name a package, they are not an install procedure, and rewriting them adds noise where nobody is stuck.

## The README bookkeeping, now done

As flagged before merge: this took the README from 480 to 504 lines against a soft ~500 ceiling I keep, with a standing rule that prose pushing past it displaces a section into `docs/` rather than moving the ceiling. Done in `a08e858` — the Troubleshooting entries moved to `docs/backends-and-ops.md#troubleshooting` behind a three-line pointer, 501 lines.

`soup doctor` stayed in the README on purpose, because it is the one thing a stuck reader should run before reading anything else. What left is the two specific failure notes, which you only need once you have a symptom.

That is bookkeeping on a rule you could not see, not a walk-back of your change — and it is worth saying that your section is the one I kept intact while trimming around it.

## For the record on scope

You asked on #675 whether `CONTRIBUTING.md` and the docs pages belonged there or here. Here was right, and you read the ruling correctly without needing a second round.

PR #675, opened 2026-09-04
fix(migrate): gate the JSONL guard on content, not just the filename

### LINKED ISSUE #672: soup migrate rejects a valid YAML config named *.jsonl: the guard branches on the filename only, and _looks_like_jsonl() has no caller
`soup migrate` rejects any file named `*.jsonl`, including a perfectly valid YAML config, because the JSONL guard branches on the filename alone. The content sniff that was written to gate it, `_looks_like_jsonl()`, has one definition and zero call sites in `src/`.

Reproduced on `main` at `f07e07e` (v0.74.0), Linux x86_64, Python 3.11.15, editable install.

## 1. `_looks_like_jsonl()` has no caller

`src/soup_cli/commands/migrate.py:144`:

```
$ grep -rn "_looks_like_jsonl" src/ tests/ scripts/ docs/
src/soup_cli/commands/migrate.py:144:def _looks_like_jsonl(path: Path) -> bool:
tests/test_v0401_part_d.py:67:    from soup_cli.commands.migrate import _looks_like_jsonl
tests/test_v0401_part_d.py:71:    assert _looks_like_jsonl(jsonl) is True
tests/test_v0401_part_d.py:75:    from soup_cli.commands.migrate import _looks_like_jsonl
tests/test_v0401_part_d.py:79:    assert _looks_like_jsonl(yml) is False
tests/test_v0401_part_d.py:83:    from soup_cli.commands.migrate import _looks_like_jsonl
tests/test_v0401_part_d.py:87:    assert _looks_like_jsonl(f) is True
```

One definition, zero call sites in `src/`. The only references outside the definition are in one test module. Confirmed by AST rather than by grep alone:

```
_looks_like_jsonl defined: True
_looks_like_jsonl called : False
```

## 2. The code contradicts the comment sitting on top of it

`migrate.py:66-70`:

```python
    # v0.40.1 Part D / N2 — friendly error when the user passes a JSONL
    # data file (`.jsonl`) instead of a YAML config. The sniff helper is
    # only invoked when the suffix says ``.jsonl`` (notebook .ipynb files
    # legitimately start with ``{`` — we must not falsely flag them).
    if input_path.suffix.lower() == ".jsonl":
```

The comment states the sniff helper *is* invoked, gated by the suffix. The `if` tests the suffix and nothing else. The intended guard is `suffix == ".jsonl"` **and** `_looks_like_jsonl(path)`; only the suffix half was wired. The `.ipynb` parenthetical is the design reason the sniff is gated *by* the suffix rather than run on every input, so it is an argument for the two-condition guard, not against it.

## Consequence: a valid config named `*.jsonl` is unmigratable

`config.jsonl`, a valid LLaMA-Factory config that merely carries the wrong extension:

```yaml
model_name_or_path: meta-llama/Llama-3-8B
stage: sft
finetuning_type: lora
```

```
$ soup migrate --from llamafactory config.jsonl --dry-run
Expected a llamafactory YAML config; got JSONL (config.jsonl) — did you pass the
wrong file?
Tip: `soup migrate` migrates competitor *configs*, not training data. Pass the 
.yaml / .ipynb file instead.
$ echo $?
2
```

The sniff, had it run, returns the right answer on that file:

```
>>> _looks_like_jsonl(Path("config.jsonl"))
False
```

So the guard fires on a file the feature's own helper classifies as not-JSONL. There is no flag to override it; the only remedy is to rename the file. The error text is also wrong in a way that costs the user time: it asserts the file *is* JSONL when it is not.

## Why it survived

`tests/test_v0401_part_d.py:66` sits under a section comment reading `# --- N2: soup migrate JSONL friendly error ---` and is named after the command:

```python
def test_migrate_jsonl_input_yields_friendly_error(tmp_path: Path):
    from soup_cli.commands.migrate import _looks_like_jsonl

    jsonl = tmp_path / "data.jsonl"
    jsonl.write_text('{"prompt": "hi"}\n{"prompt": "world"}\n', encoding="utf-8")
    assert _looks_like_jsonl(jsonl) is True
```

The name and the section comment assert the command's behaviour; the body asserts the helper's. It imports the helper and calls it directly, so it passes whether or not `migrate` invokes it, which means it passes on a codebase where `migrate.py` never references the helper at all. That is precisely the codebase we have. The two sibling tests at lines 74 and 82 have the same shape.

This is the same shape as #363, #423 and #659: code that exists, is tested, and is unreachable from the command meant to use it. The other two tests in this file that cover a command (`test_init_force_flag_overwrites_without_prompt` at line 41) already drive it through `CliRunner`, so the pattern to follow is in the same module.

## Suggested fix shape

- Require both conditions in the guard: `input_path.suffix.lower() == ".jsonl" and _looks_like_jsonl(input_path)`. That makes the code match the comment already above it, and leaves `.ipynb` untouched because the suffix half still gates the sniff.
- Rewrite `test_migrate_jsonl_input_yields_friendly_error` to invoke the **command** through `typer.testing.CliRunner`, and add a control asserting a YAML config named `*.jsonl` now migrates. The wiring is the deliverable, so a test that still passes with the new call site deleted has not fixed anything.
- Keep the three helper-level tests. They are correct as unit tests of the sniff; they were only ever mis-scoped as tests of the command.

## Acceptance

- [ ] `soup migrate --from llamafactory config.jsonl --dry-run`, where the file is valid YAML, migrates and exits 0
- [ ] `soup migrate --from llamafactory data.jsonl --dry-run`, where the file is real JSONL, still prints the friendly error and exits 2
- [ ] A `.ipynb` notebook, which legitimately starts with `{`, is still never flagged
- [ ] A test that drives the command via `CliRunner`, not the helper, and that **fails if the new call site is deleted**, plus the passing control above

## Separate question, explicitly out of scope for this issue

A real JSONL file that is *not* named `.jsonl` gets no friendly error at all, because the sniff is gated by the suffix:

`data.json`, containing:

```
{"prompt": "hi", "completion": "there"}
{"prompt": "a", "completion": "b"}
```

```
$ soup migrate --from llamafactory data.json --dry-run

Error: ParserError: expected '<document start>', but found '{'
  in "<unicode string>", line 2, column 1:
    {"prompt": "a", "completion": "b"}
    ^
Run with --verbose for the full traceback.
$ echo $?
1
```

`_looks_like_jsonl` returns `True` on that file, so the information to produce a friendly error exists. But N2 as written only ever promised to catch `.jsonl`-**named** files, and un-gating the sniff runs straight into the `.ipynb` concern the original comment raised. Widening it is a scope decision, not a bug fix, so I am not proposing it here and will not fold it into the fix above. Flagging it for you to rule on, as its own issue or as a wontfix.

## Claim

I am taking this one. Per CONTRIBUTING "Claiming an Issue", the comment is the claim, and since I am the reporter I am stating it up front rather than in a follow-up. Scope is exactly the four acceptance boxes above, nothing from the section immediately preceding this one.

No special hardware is needed: this reproduces on any platform with a plain editable install, no GPU and no Apple Silicon.

Found with codegraph (https://github.com/swalla02/codegraph), whose `orphans` check surfaced `_looks_like_jsonl` as defined-but-uncalled; everything above was then verified by hand against the checkout.

swalla02:
Closes #672.

`soup migrate` refuses any input named `*.jsonl`, including a valid LLaMA-Factory or Axolotl YAML config that merely carries that extension. Exit 2, `got JSONL`, and no flag to override it.

The comment above the guard already describes the intended behaviour:

```python
# v0.40.1 Part D / N2 — friendly error when the user passes a JSONL
# data file (`.jsonl`) instead of a YAML config. The sniff helper is
# only invoked when the suffix says ``.jsonl`` (notebook .ipynb files
# legitimately start with ``{`` — we must not falsely flag them).
if input_path.suffix.lower() == ".jsonl":
```

Two conditions, suffix and content, with the `.ipynb` case given as the reason the sniff is *gated by* the suffix rather than run on every file. Only the suffix half was ever wired. `_looks_like_jsonl` had one definition and zero call sites in `src/`.

## The fix

One line:

```diff
-    if input_path.suffix.lower() == ".jsonl":
+    if input_path.suffix.lower() == ".jsonl" and _looks_like_jsonl(input_path):
```

That is the guard the comment already documents, and it leaves the `.ipynb` path untouched because the suffix gate still comes first.

## The test was the reason this survived

`test_migrate_jsonl_input_yields_friendly_error` was named after the command's behaviour and its body called `_looks_like_jsonl` directly. It passed on a codebase where `migrate` never imported the helper, which is the codebase we had. Same shape as #659, #423 and #363.

It now drives the command through `CliRunner`, and a control asserts that a YAML config named `config.jsonl` migrates. The helper's three original assertions are kept as unit tests of the sniff itself, since they are still worth having.

**Non-vacuity, proven rather than asserted.** With the new call site deleted:

```
FAILED tests/test_v0401_part_d.py::test_migrate_yaml_config_named_jsonl_still_migrates
1 failed, 4 passed, 6 deselected

>       assert result.exit_code == 0, (result.output, repr(result.exception))
E       AssertionError: ('Expected a llamafactory YAML config; got JSONL (config.jsonl) — did you pass the
E         wrong file? ...', 'SystemExit(2)')
E       assert 2 == 0
```

Restored by re-editing, tree byte-identical to the commit:

```
5 passed, 6 deselected
```

## Behaviour, before and after

| input | before | after |
|---|---|---|
| real JSONL named `data.jsonl` | exit 2, friendly error | exit 2, friendly error (unchanged) |
| YAML config named `config.jsonl` | exit 2, unmigratable | exit 0, migrates |
| notebook `.ipynb` starting with `{` | never flagged | never flagged |

The `.ipynb` row was checked by hand as well as by suite: `got JSONL` appears zero times in the output.

## Test and lint

`ruff check src/soup_cli/ scripts/ tests/` passes, before and after.

Full suite, baseline established on a clean `f07e07e` worktree first: **83 failed, 19560 passed, 380 skipped**. After the fix: **83 failed, 19562 passed, 380 skipped**. Same failure count, same failing files, plus the two new tests. `tests/test_v0401_part_d.py` goes 9 passed to 11 passed.

Being straight about that baseline: those 83 are pre-existing in my environment and I do not believe they are a clean-CI figure. The venv arrived without the training stack; I installed CPU torch 2.14, transformers 5.16.1, peft, trl, datasets and accelerate, but not `bitsandbytes`, and the failures cluster in `test_v07202.py` (47), `test_v07204.py` (16) and the TRL contract files. Nothing failing touches `migrate`. Coverage was disabled, because `--cov-fail-under=77` is not a number I could stand behind with an incomplete dependency set, and I did not want to change the threshold to make it green.

## Separate question, not addressed here

Real JSONL under a name that does not say `.jsonl`, for example `data.json`, still gets the raw parser error:

```
$ soup migrate --from llamafactory data.json --dry-run
Error: ParserError: expected '<document start>', but found '{'
```

That is the message the N2 feature exists to replace, but catching it means sniffing files whose name gives no hint, which is exactly what the `.ipynb` parenthetical warns against. It is a scope call rather than a bug, so it is in #672 for you to rule on and deliberately not in this PR.

## Changelog fragment

Currently `changelog.d/0.74.0/672.fixed.md`, named for the issue because no PR number existed when it was written. `changelog.d/README.md` asks for the PR number, so I will rename it to this PR's number in a follow-up commit on this branch.

## How it was found

`_looks_like_jsonl` surfaced in a caller-reachability report from [codegraph](https://github.com/swalla02/codegraph), a static call-graph tool I maintain, which lists functions whose only recorded callers are tests. The probes and the reading of the code are mine.

### DIFF
--- a/changelog.d/0.74.0/672.fixed.md
+++ b/changelog.d/0.74.0/672.fixed.md
@@ -1,0 +1,8 @@
+- `soup migrate` no longer refuses a valid competitor config just because the file
+  is named `*.jsonl`. The guard branched on the filename alone; the content sniff
+  written to gate it, `_looks_like_jsonl()`, had no call site, so a YAML config
+  saved as `config.jsonl` exited 2 with "got JSONL" and could not be migrated at
+  all. Both conditions are now required, which leaves `.ipynb` notebooks (which
+  legitimately start with `{`) untouched. The N2 test now drives the command
+  through `CliRunner` rather than calling the helper, so the wiring is what is
+  actually asserted (#672).
--- a/changelog.d/0.74.0/675.fixed.md
+++ b/changelog.d/0.74.0/675.fixed.md
@@ -5,4 +5,7 @@
   all. Both conditions are now required, which leaves `.ipynb` notebooks (which
   legitimately start with `{`) untouched. The N2 test now drives the command
   through `CliRunner` rather than calling the helper, so the wiring is what is
-  actually asserted (#675, closes #672).
+  actually asserted (#675 by @swalla02, closes #672).
+  The sniff also reads `utf-8-sig` rather than `utf-8`, so a UTF-8 BOM (which
+  Windows tooling writes by default) cannot defeat it, and its read is bounded
+  so a file with no newline is not pulled into memory entire.
--- a/changelog.d/0.74.0/675.fixed.md
+++ b/changelog.d/0.74.0/675.fixed.md
@@ -5,4 +5,4 @@
   all. Both conditions are now required, which leaves `.ipynb` notebooks (which
   legitimately start with `{`) untouched. The N2 test now drives the command
   through `CliRunner` rather than calling the helper, so the wiring is what is
-  actually asserted (#672).
+  actually asserted (#675, closes #672).
--- a/src/soup_cli/commands/migrate.py
+++ b/src/soup_cli/commands/migrate.py
@@ -142,10 +142,22 @@ def migrate(
 
 
 def _looks_like_jsonl(path: Path) -> bool:
-    """v0.40.1 Part D / N2 — sniff first non-blank line for `{` (JSONL)."""
+    """v0.40.1 Part D / N2 — sniff first non-blank line for `{` (JSONL).
+
+    ``utf-8-sig``, not ``utf-8``: a UTF-8 BOM decodes to U+FEFF, which
+    ``str.strip()`` does not remove because it is not whitespace, so a BOM'd
+    JSONL file would sniff as *not* JSONL and silently lose the friendly
+    error. Windows tooling writes that BOM by default, PowerShell's
+    ``Out-File`` included (#675 review).
+
+    The read is bounded because a file with no newline is one single line, so
+    ``for line in fh`` would pull all of it into memory: a 120 MB one-liner
+    measured 240.9 MB peak. Only the first non-blank chunk is ever inspected,
+    so a truncated long line costs nothing here.
+    """
     try:
-        with open(path, "r", encoding="utf-8", errors="replace") as fh:
-            for line in fh:
+        with open(path, "r", encoding="utf-8-sig", errors="replace") as fh:
+            while line := fh.readline(65536):
                 stripped = line.strip()
                 if not stripped:
                     continue
--- a/src/soup_cli/commands/migrate.py
+++ b/src/soup_cli/commands/migrate.py
@@ -67,7 +67,7 @@ def migrate(
     # data file (`.jsonl`) instead of a YAML config. The sniff helper is
     # only invoked when the suffix says ``.jsonl`` (notebook .ipynb files
     # legitimately start with ``{`` — we must not falsely flag them).
-    if input_path.suffix.lower() == ".jsonl":
+    if input_path.suffix.lower() == ".jsonl" and _looks_like_jsonl(input_path):
         console.print(
             f"[red]Expected a {source} YAML config; got JSONL "
             f"({input_path.name}) — did you pass the wrong file?[/]"
--- a/tests/test_v0401_part_d.py
+++ b/tests/test_v0401_part_d.py
@@ -63,7 +63,47 @@ def test_init_force_flag_overwrites_without_prompt(tmp_path):
 # --- N2: soup migrate JSONL friendly error ------------------------------
 
 
-def test_migrate_jsonl_input_yields_friendly_error(tmp_path: Path):
+def test_migrate_jsonl_input_yields_friendly_error(tmp_path: Path, monkeypatch):
+    """Drives the *command*, not the helper: real JSONL named `.jsonl` is refused."""
+    from soup_cli.cli import app
+
+    monkeypatch.chdir(tmp_path)
+    jsonl = tmp_path / "data.jsonl"
+    jsonl.write_text('{"prompt": "hi"}\n{"prompt": "world"}\n', encoding="utf-8")
+
+    runner = CliRunner()
+    result = runner.invoke(
+        app, ["migrate", "--from", "llamafactory", "data.jsonl", "--dry-run"]
+    )
+    assert result.exit_code == 2, (result.output, repr(result.exception))
+    assert "got JSONL" in result.output
+
+
+def test_migrate_yaml_config_named_jsonl_still_migrates(tmp_path: Path, monkeypatch):
+    """Control: the suffix alone must not condemn a file whose content is YAML.
+
+    Fails if the `_looks_like_jsonl` call site is removed from the guard.
+    """
+    from soup_cli.cli import app
+
+    monkeypatch.chdir(tmp_path)
+    config = tmp_path / "config.jsonl"
+    config.write_text(
+        "model_name_or_path: meta-llama/Llama-3-8B\n"
+        "stage: sft\n"
+        "finetuning_type: lora\n",
+        encoding="utf-8",
+    )
+
+    runner = CliRunner()
+    result = runner.invoke(
+        app, ["migrate", "--from", "llamafactory", "config.jsonl", "--dry-run"]
+    )
+    assert result.exit_code == 0, (result.output, repr(result.exception))
+    assert "got JSONL" not in result.output
+
+
+def test_migrate_helper_detects_jsonl(tmp_path: Path):
     from soup_cli.commands.migrate import _looks_like_jsonl
 
     jsonl = tmp_path / "data.jsonl"
--- a/tests/test_v0401_part_d.py
+++ b/tests/test_v0401_part_d.py
@@ -103,6 +103,58 @@ def test_migrate_yaml_config_named_jsonl_still_migrates(tmp_path: Path, monkeypa
     assert "got JSONL" not in result.output
 
 
+def test_migrate_bomd_jsonl_still_yields_friendly_error(tmp_path: Path, monkeypatch):
+    """A UTF-8 BOM must not defeat the sniff (#675 review).
+
+    Windows tooling writes UTF-8 with a BOM by default. Decoded as plain
+    ``utf-8`` the BOM survives as U+FEFF, which ``str.strip()`` leaves alone
+    because it is not whitespace, so ``startswith("{")`` is False and real
+    JSONL sniffs as a config. Fails if the helper reads ``utf-8`` rather than
+    ``utf-8-sig``.
+    """
+    from soup_cli.cli import app
+
+    monkeypatch.chdir(tmp_path)
+    jsonl = tmp_path / "bom.jsonl"
+    jsonl.write_bytes(
+        b"\xef\xbb\xbf" + b'{"prompt": "hi"}\n{"prompt": "world"}\n'
+    )
+    # Guard the fixture itself: plain utf-8 must see the BOM this test is about.
+    assert jsonl.read_text(encoding="utf-8").startswith("\ufeff")
+
+    runner = CliRunner()
+    result = runner.invoke(
+        app, ["migrate", "--from", "llamafactory", "bom.jsonl", "--dry-run"]
+    )
+    assert result.exit_code == 2, (result.output, repr(result.exception))
+    assert "got JSONL" in result.output
+
+
+def test_migrate_sniff_does_not_read_a_whole_unnewlined_file(tmp_path: Path):
+    """The sniff is bounded: one very long line must not be read entire.
+
+    ``for line in fh`` on a file with no newline pulls all of it into one
+    string. Fails if the bounded ``readline`` is reverted.
+    """
+    import tracemalloc
+
+    from soup_cli.commands.migrate import _looks_like_jsonl
+
+    big = tmp_path / "oneline.jsonl"
+    big.write_text('{"prompt": "' + "x" * (8 * 1024 * 1024) + '"}', encoding="utf-8")
+
+    tracemalloc.start()
+    try:
+        assert _looks_like_jsonl(big) is True
+        peak = tracemalloc.get_traced_memory()[1]
+    finally:
+        tracemalloc.stop()
+
+    # Bounded read touches 64 KiB; the unbounded one allocated the 8 MiB line
+    # at least twice (read plus strip). 1 MiB leaves generous headroom.
+    assert peak < 1024 * 1024, f"peak {peak} bytes: the read looks unbounded"
+
+
 def test_migrate_helper_detects_jsonl(tmp_path: Path):
     from soup_cli.commands.migrate import _looks_like_jsonl
 


MakazhanAlpamys:
@swalla02 — thanks, and welcome. Reviewed by execution, not by reading the description. **The fix is right, the issue is closed by it, and the control test genuinely pins it.** Two things to fold in before merge, one of which is a behaviour regression this PR introduces.

First, credit where it is due: you found this with a caller-reachability report and then read the code rather than trusting the tool. `_looks_like_jsonl` having one definition and zero call sites in `src/` is the third instance of that exact shape here (#363, #423, #659), and the diagnosis in your description — a test named after the command whose body exercises the helper — is precisely why it survived. That paragraph is more valuable than the one-line diff.

## Verified by execution

Isolated worktree, `PYTHONPATH` pinned to the branch (unpinned, the editable install resolves to the main tree and every mutation is a silent no-op). Reproduced against a clean `origin/main` worktree side by side:

| input | main `64e0323` | PR #675 |
|---|---|---|
| `cfg.jsonl` — valid LLaMA-Factory YAML | exit **2**, `got JSONL` | exit **0**, migrates |
| `plain.jsonl` — real JSONL | exit 2, `got JSONL` | exit 2, `got JSONL` (unchanged) |
| `real.ipynb` — starts with `{` | never flagged | never flagged |

`tests/test_v0401_part_d.py` -> **11 passed**. Mutations:

| mutation | result |
|---|---|
| guard reverted to suffix-only | **KILLED** — `test_migrate_yaml_config_named_jsonl_still_migrates` |
| `_looks_like_jsonl` -> `return True` | **KILLED** (2 tests) |
| `_looks_like_jsonl` -> `return False` | **KILLED** (3 tests, incl. `test_migrate_jsonl_input_yields_friendly_error`) |
| `and` -> `or` in the guard | **KILLED** |
| sniff `startswith("[")` instead of `"{"` | **KILLED** (3 tests) |

The third row is the one I most wanted to see: "always False" still kills the friendly-error test, so the real-JSONL path is pinned in both directions rather than only the new one. Your claim that the three original helper assertions survive is literally true — the single removed test line is the renamed `def`, nothing else.

Also checked, since CLI-output assertions have turned this repo's CI red four times: your `"got JSONL" in result.output` assertions survive `FORCE_COLOR=1 TERM=xterm-256color` and `COLUMNS=20`. Rich brackets the whole `[red]` span here rather than emitting per-character escapes, so you are safe — but that is luck rather than design, and worth knowing about before your next CLI-output test.

## SHOULD-FIX 1 — a UTF-8 BOM now loses the friendly error

`src/soup_cli/commands/migrate.py:147` opens with `encoding="utf-8"`, so a BOM survives as `\ufeff`. `str.strip()` does not remove it (it is not whitespace in Python 3), so `stripped.startswith("{")` is False and a real BOM'd JSONL file now sniffs as *not* JSONL. Measured, same command on both trees:

```
bom.jsonl    main: exit 2, "got JSONL"    PR: exit 1, raw ParserError
plain.jsonl  main: exit 2, "got JSONL"    PR: exit 2, "got JSONL"
```

That is a real regression rather than a theoretical one on this project — Windows tooling emits UTF-8 with BOM by default, PowerShell's `Out-File` included, and CI runs three Windows cells. Fix is one word:

```python
with open(path, "r", encoding="utf-8-sig", errors="replace") as fh:
```

Worth a test: a BOM'd `.jsonl` still gets the friendly error.

## SHOULD-FIX 2 — the sniff is unbounded, and this PR is what makes it reachable

`for line in fh` on a file with no newline reads the whole file into one string. On a 120 MB single-line file that measured **240.9 MB peak allocation**. The code is pre-existing, but it had zero call sites in `src/` until this PR, so this is the change that puts it on the CLI path — which makes it fair to fix here rather than to file separately. No traversal angle (`validate_input_path` already bounds the path, and the file is the user's own), so this is robustness, not security.

```python
while (line := fh.readline(65536)):
```

## NIT — the changelog fragment does not credit you

`changelog.d/0.74.0/675.fixed.md` is in the right directory, uses the PR number, and passes the validator (17/17). It ends `(#675, closes #672)`. The project's convention is `(#NNN by @user)`, and v0.74.0's own release audit found 56 of 116 entries missing their author — this is that exact recurring miss, caught before merge for once. Please make it `(#675 by @swalla02, closes #672)`.

## Rulings on the two things you asked me to rule on

**Real JSONL under a name that does not say `.jsonl`** — leave it out, you were right to. Sniffing files whose name gives no hint is what the `.ipynb` parenthetical warns against, and the raw parser error there is ugly rather than wrong. I have kept #672 open for it as a separate question.

**`CONTRIBUTING.md` line 22 and the five `docs/*.md` pages** — please do add them, but on #673, not here. That PR is already the install-instructions change and this one should stay a one-line behaviour fix.

Nothing here needs a docs update: `docs/commands.md:151` and `docs/data.md:483` only show `.yaml` / `.yml` / `.ipynb` examples and remain accurate.

Push those and I will re-run the mutations and merge.

swalla02:
Pushed as `7363759`. All four folded in, and thank you for finding the BOM one: that is a regression I introduced and did not think to look for.

## SHOULD-FIX 1, the BOM

`encoding="utf-8-sig"`. Your reading of it is exactly right, and the reason it is a real bug rather than a theoretical one is the bit I had missed: the helper was unreachable before this branch, so nobody could have hit it, and this PR is what makes a BOM'd file reachable.

Pinned by `test_migrate_bomd_jsonl_still_yields_friendly_error`, which drives the command, writes the BOM as bytes, and asserts against the fixture first (`read_text(encoding="utf-8").startswith("﻿")`) so the test cannot pass by silently producing a BOM-free file. Reverting one word:

```
$ sed -i 's/utf-8-sig/utf-8/' src/soup_cli/commands/migrate.py
FAILED tests/test_v0401_part_d.py::test_migrate_bomd_jsonl_still_yields_friendly_error
1 failed, 12 passed
```

## SHOULD-FIX 2, the unbounded read

`while line := fh.readline(65536):`. Agreed on where it belongs: pre-existing code, but zero call sites in `src/` until this branch, so this branch is what puts it on the CLI path.

Pinned by `test_migrate_sniff_does_not_read_a_whole_unnewlined_file`, using `tracemalloc` on an 8 MiB single-line file with a 1 MiB ceiling. Reverting to `for line in fh:`:

```
FAILED tests/test_v0401_part_d.py::test_migrate_sniff_does_not_read_a_whole_unnewlined_file
E       AssertionError: peak 16841531 bytes: the read looks unbounded
E       assert 16841531 < (1024 * 1024)
1 failed, 12 passed
```

16.8 MB on an 8 MiB line, which is the 2x you measured at 120 MB, so the shape reproduces at a size that keeps the test fast. The test takes about 40 seconds when it fails and under 1 second when it passes.

I sized the ceiling at 1 MiB rather than just above 64 KiB deliberately: it leaves room for interpreter noise without leaving room for the bug, since the failing mode is 16x over it.

## NIT, the credit

`(#675 by @swalla02, closes #672)`. The fragment also now names the two review fixes, since the BOM behaviour is user-visible in its own right on Windows. Validator still passes.

## Your two rulings

Both taken. The unnamed-JSONL question stays out of this PR and on #672 for you.

`CONTRIBUTING.md` line 22 and the five `docs/*.md` pages: agreed they belong on #673 rather than here, and I will push them there rather than widening this one.

## State after the push

```
tests/test_v0401_part_d.py            13 passed  (11 before, +2 new)
scripts/assemble_changelog.py         Assembled 1 changelog fragment(s).
git status --porcelain                (empty after both mutation probes were reverted by re-editing)
```

I did not re-run the full 19,500-test suite for this push, since the diff is one encoding argument, one loop, two tests and a changelog line. Say the word if you want the number anyway. The environment caveat from the description still stands: 83 pre-existing failures here that I do not believe are a clean-CI figure, none of them touching `migrate`.

Also noted on the Rich colour point, and taken as a warning rather than a pass: the assertion survives `FORCE_COLOR=1` and `COLUMNS=20` because Rich brackets the whole `[red]` span, which is luck. I will assert on `result.exit_code` plus a substring short enough to survive wrapping in future CLI tests rather than relying on that.

MakazhanAlpamys:
`7363759` closes both findings, and the tests for them are better than the fixes. Approved on substance — holding only for the checks, which I have released (first-time-contributor runs need manual approval here).

## Re-verified by execution

Isolated worktree at `7363759`, `PYTHONPATH` pinned. Baseline `tests/test_v0401_part_d.py` -> **13 passed**.

| mutation | result |
|---|---|
| `encoding="utf-8-sig"` reverted to `"utf-8"` | **KILLED** — `test_migrate_bomd_jsonl_still_yields_friendly_error` |
| bounded `readline(65536)` reverted to `for line in fh` | **KILLED** — `test_migrate_sniff_does_not_read_a_whole_unnewlined_file` |
| the original fix removed (guard back to suffix-only) | **KILLED** — `test_migrate_yaml_config_named_jsonl_still_migrates` |

End to end, all four cases through the real CLI:

| input | exit | friendly error |
|---|---|---|
| `bom.jsonl` — real JSONL with a UTF-8 BOM | 2 | yes — **the regression is gone** |
| `plain.jsonl` — real JSONL | 2 | yes (unchanged) |
| `cfg.jsonl` — YAML config named `.jsonl` | 0 | migrates — the fix |
| `nb.ipynb` — notebook starting with `{` | — | never flagged |

ruff clean.

## Two things you did that I want to name, because most contributors do not

**The BOM test guards its own fixture.** `assert jsonl.read_text(encoding="utf-8").startswith("﻿")` before the real assertion means that if some future change makes the fixture stop containing a BOM, the test fails loudly instead of passing for the wrong reason. That is a test that knows how it could become vacuous. This repo has a documented history of exactly that failure — a test whose name describes one thing and whose body asserts another — and it is what let #672 survive in the first place.

**The bounded-read test measures rather than asserts a shape.** `tracemalloc` with a 1 MiB ceiling against an 8 MiB single line is checking the property, not the spelling. Someone could rewrite that loop three different ways and the test still means the same thing.

Both docstrings also state what would make the test fail, which is the part people skip.

## One clarification I want on the record

The bounded read changes a real edge: a first non-blank chunk longer than 64 KiB is now truncated before the `startswith("{")` check. That is harmless — you only ever inspect the first character, and your docstring says so — but it is worth having written down, because someone later may be tempted to widen that helper into a fuller parse and the bound would then silently matter.

I checked the neighbouring cases while I was there: 64 KiB of leading whitespace followed by `{` still sniffs correctly, because the blank chunk is skipped and the next one is read.

## Scope note

Your reading of my #675 ruling was right, and the `CONTRIBUTING.md` and `docs/models.md` work belongs on #673 where you put it. Nothing further needed here.

Merging when the matrix reports.

MakazhanAlpamys:
Merged as `cd11156`, credited in `CONTRIBUTORS.md` (`5944755`). Welcome, and this is an unusually good first PR.

## Final verification

Re-materialised the merge against current `main` — which moved four times since your checks ran — and ran it: **30 passed** across `test_v0401_part_d.py` and the changelog guard. Zero file overlap with anything merged in between.

Mutations across both rounds, six for six:

| mutation | killed by |
|---|---|
| guard reverted to suffix-only | `test_migrate_yaml_config_named_jsonl_still_migrates` |
| `_looks_like_jsonl` -> `return True` | 2 tests |
| `_looks_like_jsonl` -> `return False` | 3 tests, incl. `test_migrate_jsonl_input_yields_friendly_error` |
| `and` -> `or` in the guard | `..._named_jsonl_still_migrates` |
| `utf-8-sig` -> `utf-8` | `test_migrate_bomd_jsonl_still_yields_friendly_error` |
| bounded `readline(65536)` -> `for line in fh` | `test_migrate_sniff_does_not_read_a_whole_unnewlined_file` |

End to end, four cases through the real CLI: YAML named `.jsonl` exit 2 -> **0**; real JSONL still exit 2 with the friendly error; BOM'd JSONL exit 2 with the friendly error; `.ipynb` never flagged.

## What was good, specifically

**The diagnosis outweighs the diff.** A one-line fix is a one-line fix, but "the test was named after the command and its body called the helper, so it passed on a codebase where `migrate` never imported it" is the reusable part. That is the third time this exact shape has been found here (#363, #423, #659) and the first time someone outside wrote down *why* it survives.

**You proved non-vacuity instead of asserting it.** Deleting the new call site, showing the failure, restoring, showing the tree byte-identical — that is the step almost everyone skips, and it is the only thing that distinguishes a regression test from a test.

**Your fix for my BOM finding is better than the finding.** The test asserts the fixture really contains a BOM *before* asserting behaviour, so it fails loudly rather than passing for the wrong reason if the fixture ever drifts. And the bounded-read test measures peak allocation with `tracemalloc` rather than pinning the shape of the loop, so a future rewrite still means the same thing. Both docstrings state what would make the test fail.

**You declined to widen the scope and said why.** Real JSONL under a name that gives no hint still gets the raw parser error, and catching it means sniffing files the `.ipynb` parenthetical warns about. Leaving that as an open question on #672 rather than guessing was the right call — I have kept the issue open for it.

**And you were straight about your environment.** Reporting an 83-failure baseline and saying plainly that it was an incomplete dependency set, rather than quietly presenting it as clean CI, is worth more to me than a green screenshot would have been. For the record: those clustered failures are the documented `torch>=2.5.0`-against-`trl>=0.29` floor problem, #651, and none of them touch `migrate`.

## Small note

I did not need the extra full-suite run you offered on #673. Between the two PRs you have done more verification than most merged work here gets.

The name-gives-no-hint question is now #699 -- #672 closed automatically on merge, and re-opening an issue whose title is exactly the bug that is fixed would make the tracker less accurate, so your framing was carried over to a new one. If you want another, #676 and #695 are both in the same family and #677 is a guard I would like to exist.

PR #678, opened 2026-09-04
fix: drop non-dict chatml/audio/video messages

### LINKED ISSUE #676: chatml / audio / video converters silently keep a non-dict message instead of dropping the row
Found while reviewing #670, which fixes the neighbouring case in `_convert_multimodal`. Filing separately rather than widening that PR, since its scope is right as it stands.

## What happens

`format_to_messages` documents a drop contract: a malformed row routes to the drop path and returns `None`, so one bad line in a JSONL file is skipped rather than corrupting the dataset. `_require_str_content`'s own docstring states it directly — raising is what routes "the row to `format_to_messages`'s drop path instead of silently corrupting the dataset".

Three converters do not honour it. A non-dict element inside `messages` is passed through verbatim:

```python
format_to_messages({"messages": ["hello"]}, "chatml")   # -> {'messages': ['hello']}
format_to_messages({"messages": ["hello"]}, "audio")    # -> {'messages': ['hello']}
format_to_messages({"messages": ["hello"]}, "video")    # -> {'messages': ['hello']}
```

Expected in each case is `None`, i.e. the row is dropped.

## Why this matters more than the multimodal case it was found beside

`chatml` is what `detect_format` returns for a bare `{"messages": [...]}` row, so it is the **default** path for the most common dataset shape in this project. `multimodal` (#670) required an explicit `format: multimodal` in the config to reach at all.

The failure mode is also the worse of the two. #670's bug was loud — an uncaught `AttributeError` killed the run, which is at least visible. This one is silent: the row survives into training with a bare string where a message object belongs, and nothing reports it.

## Verified

Executed across four bad shapes (`"hello"`, `42`, `["role","user"]`, `None`) and ten format paths on `origin/main` at `64e0323`:

| converter | non-dict element |
|---|---|
| `sharegpt`, `llava`, `sharegpt4v`, `input_output`, `tool-calling` | dropped correctly (they subscript, and `TypeError` **is** in the caught tuple) |
| `multimodal` | uncaught `AttributeError` — fixed by #670 |
| `chatml`, `audio`, `video` | **kept verbatim** — this issue |

So the drop contract holds for five converters, was loudly broken in one, and is silently broken in three.

## Second, related defect in the same area

`soup data validate <file> --format multimodal` reports `5/5 rows valid` for a file that `load_dataset` cannot load, on both `main` and #670's branch — the validator does not exercise this code path. A validator that green-lights a file the loader then rejects is worse than no validator, since it moves the failure to the middle of a training run. Filing it here rather than separately because the fix likely shares a helper; split it out if it turns out not to.

## Suggested fix

Mirror the guard #670 added, using the convention already established by `_convert_input_output` at `formats.py:549` (`ValueError` for a shape-of-element check):

```python
for msg in messages:
    if not isinstance(msg, dict):
        raise ValueError("chatml message must be a dict")
```

## Acceptance criteria

- [ ] `format_to_messages({"messages": [<non-dict>]}, fmt)` returns `None` for `fmt` in `chatml`, `audio`, `video`, over the four shapes above
- [ ] A test that distinguishes **dropped** from **passed through** — assert the return is `None`, not merely that no exception is raised. Replacing the guard with `continue` must fail the test; #670's test is the model here
- [ ] Valid rows in all three formats convert exactly as before
- [ ] `soup data validate` and `load_dataset` agree on the same file, or the mismatch is documented deliberately
- [ ] Changelog fragment under the newest release directory, crediting the author

BetterAndBetterII:
Chatml, audio, and video converters kept non-dict messages instead of dropping the malformed row. They now validate message elements, and the list-type guard contributed by @v01dst in #679 also rejects empty non-list values such as `{"messages": ""}` and `{"messages": {}}`. Valid rows, empty chatml lists, and video rows with missing or null messages retain their existing behavior.

Synced with current `main`, preserving #712's shared validator and #670's multimodal drop behavior. Tests cover converter rejection, automatic chatml loading, and validator/loader agreement. The changelog fragment is now `changelog.d/0.74.0/678.fixed.md` with `(#676 by @BetterAndBetterII in #678)`; @v01dst is credited in the commit body, co-author trailer, changelog, and `CONTRIBUTORS.md`.

Validation on Python 3.12:

- 291 tests passed across the message regressions, #695 validator tests, formats, validator, data, loader, audio, v0.42.0 formats, and changelog tests.
- Removing only the list guard causes six new regression cases to fail.
- Repository-wide Ruff and changelog fragment validation passed.

Fixes #676


MakazhanAlpamys:
@BetterAndBetterII — welcome, and thank you for the fastest response to an issue this project has had: #676 was filed at 19:17 and you opened at 19:48.

A second PR, #679 by @v01dst, landed ninety minutes later against the same issue. I put both through an **identical battery** and I am publishing the **same numbers in both threads**, including where the other is better than yours, because that is the only honest way to resolve two good PRs on one issue.

**Yours is the base I am taking.** Three things to fix first, and one of them is @v01dst's work going in under their name.

## The identical battery

Both branches rebased onto `a08e858` before every measurement. Clean-main baseline: 189 passed narrow, 862 keyword set, **no pre-existing failures in this scope**, so everything below is attributable.

| | **#678 (yours)** | **#679** |
|---|---|---|
| Defect 1 — 4 bad shapes x 3 formats | YES | YES |
| Residual behaviour gap | **`{"messages": ""}` and `{"messages": {}}` still kept** | none |
| Defect 2 — validate agrees with the loader | 3/7 -> **4/7** | 3/7 -> **0/7** |
| Mutations killed | **10 / 10** | 9 / 10 |
| Tests added | **16** | 10 |
| Tests removed or weakened | 0 | 0 |
| Over-reach on 300 dataset/format combos | 40 changes, all correct, exit codes unmoved | 0 |
| Applies to current `main` | **no — 1 test red** | yes |
| ruff | clean | clean |

## Why yours

**Defect 2, measured.** #676 reports two defects and yours is the only submission that addresses both. Identical fixtures, 4 good rows plus 1 malformed, validator compared against what `load_dataset` actually keeps:

| format | main | **#678** | #679 |
|---|---|---|---|
| chatml | agree *(on a shared bug)* | **4/5 vs 4 — agree** | 5/5 vs 4 — disagree |
| audio | agree *(on a bug)* | **agree** | disagree |
| video | agree *(on a bug)* | **agree** | disagree |
| multimodal | disagree | **agree** | disagree |
| **total** | 3 / 7 | **4 / 7** | **0 / 7** |

Fixing the loader alone converts main's *accidental* agreement into a genuine mismatch on the default format. Yours closes it for the four formats it touches.

**Your validator change is measurably safe, which is the thing I was most prepared to block on.** It changes a user-facing command, so I checked rather than assumed: 300 file-by-format combinations across the repo's own datasets produce 40 verdict changes, and **every one is a `N/N -> 0/N` on a file that genuinely has zero valid rows in that format** — `load_dataset` drops all of them, so `0/N` is the correct answer and main's `N/N` was the bug. Exit codes are unmoved in all 300, so no CI gate flips. Seven pathological row shapes across four formats: no crash. Cost on 50k rows: 0.202s -> 0.237s, and the *unchanged* alpaca path moved by the same amount, so that is machine noise rather than conversion cost.

**10/10 mutations killed**, including the `continue` mutant on all three converters — the bar #670 set, which distinguishes "row dropped" from "row silently corrupted". Your `RuntimeError` mutant also dies, which pins the drop contract itself and not merely the guard. Zero test lines removed; your 11 deletions are all production code.

**And you have the only CLI-level assertion** in either PR (`soup data validate` -> `1/2 rows valid`), correctly routed through `tests/conftest.py::strip_ansi`. That is the exact discipline that has turned this repo's CI red four times, and you found it without being told.

## Blocking 1 — the branch is stale and one test fails on `main`

You branched from `64e0323`, which predates #670 (`218eae5`, merged since). Rebases textually clean, then:

```
FAILED tests/test_issue676_non_dict_messages.py::test_validate_does_not_greenlight_multimodal_rows_the_loader_rejects
1 failed, 15 passed
```

`tests/test_issue676_non_dict_messages.py:111` asserts `pytest.raises(AttributeError)` on `format_to_messages({"messages": ["hello"]}, "multimodal")`. After #670 that returns `None`. Change the assertion to `is None`.

Same root, second half: `validator.py:72`'s `except AttributeError:` with the comment `# multimodal still raises on a non-dict message (#670)` is now unreachable, and the comment states the opposite of the truth. Delete both.

## Blocking 2 — a residual gap on the default path, and @v01dst's guard is the fix

Verified by execution on your rebased branch:

```
#678:  {"messages": ""} -> {'messages': ''}      {"messages": {}} -> {'messages': {}}
#679:  {"messages": ""} -> None                  {"messages": {}} -> None
```

Both shapes are JSON-reachable, both are detected as `chatml`, and through the real loader on a 3-row file yours keeps 3 where #679 keeps 1. The cause is that `_require_dict_messages` rejects a non-list by *iterating* it, which catches `"hello"` and `42` by accident and fails silently for any **empty** non-list iterable. That is the same defect class #676 is about, on the most common path.

The fix is the guard @v01dst wrote, and it matches what both reference converters already do — `_convert_input_output:549` and the merged `_convert_multimodal` each do a list check *then* an element check:

```python
if not isinstance(messages, list):
    raise ValueError(f"{kind} row 'messages' must be a list")
```

**Please add it with @v01dst credited in the commit message.** I have asked them on #679 whether they would rather land it themselves as a follow-up on top of yours; if they say yes, leave it out and I will merge yours without it. Either way the credit is theirs — I am not going to have their idea arrive in `main` anonymously.

**And please add the two tests their PR was missing**, because this is the neatest thing the comparison turned up: deleting that guard passes all 199 of #679's tests. `{"messages": "hello"}` cannot pin it, since iterating a string yields non-dict characters and the element check catches it anyway. Only `{"messages": ""}` and `{"messages": {}}` discriminate. Their implementation was right and their suite could not tell — so the guard needs tests neither PR currently has.

## Blocking 3 — the changelog fragment

`changelog.d/0.74.0/676.fixed.md` is named after the **issue**. `changelog.d/README.md` asks for the pull-request number, and says why in the same sentence: *"this keeps two pull requests for the same issue distinct and traceable."* Both PRs picked the same filename with different content, which is precisely the collision the rule exists to prevent. Rename to `678.fixed.md` and change `#676` to `#678` in the body — the validator requires the filename's number to appear in the text.

Your fragment does credit you, which #679's does not. Thank you for that; it is the convention here and it is the thing I most often have to ask for.

## Two nits, not blocking

`_require_dict_messages` returns `None` where `_require_str_content` and #679's helper both return the validated value — reads better at the call site. And there is no test against a genuinely-valid `video` or `multimodal` fixture file, because the repo ships none; that is a coverage gap in the project rather than in your PR, but your validator change is the first thing that would notice it.

## What #679 gets credited for regardless of what happens next

The list guard, verbatim in effect; matching the reference converters' two-part shape where yours has one half; pinning video's "no `messages` key -> `[]`" contract with an explicit test; and being branched on current `main` and green where yours was not. That will be in the merge commit and in `CONTRIBUTORS.md`.

Push those three and I will re-run the full battery and merge.

MakazhanAlpamys:
@BetterAndBetterII — status, plus a time, so this does not sit.

**Your CI is now green.** Your PR had never actually run: as a first-time contributor the workflow sat in `action_required` from 2026-09-04 until I approved it a few minutes ago. It now reports **15/15 pass** against current `main` (`1082db4`), so the "1 test red on current main" line in my comparison comment is out of date — that was my own local measurement against an older tree and subsequent merges have overtaken it. Correcting it here rather than leaving it standing.

Three things remain, all small:

1. **Rename the changelog fragment** `changelog.d/0.74.0/676.fixed.md` -> `changelog.d/0.74.0/678.fixed.md`. `changelog.d/README.md` asks for the **pull request** number, and this batch is exactly why: two PRs on one issue would have collided on one filename.
2. **Add the credit** — the entry needs `(#676 by @BetterAndBetterII in #678)`. Outside work gets its author named in the changelog here.
3. **Add @v01dst's list guard and the two tests it needs.** This is the substantive one. `isinstance(messages, list)` before the element check, plus fixtures for `{"messages": ""}` and `{"messages": {}}`. Your current code rejects a non-list by *iterating* it, which works by accident for `"hello"` and `42` and silently keeps any **empty** non-list iterable — both of those shapes are JSON-reachable and both detect as `chatml`, the default format. It is @v01dst's finding from #679 and it must land credited to them in the commit body and in `CONTRIBUTORS.md`.

**If I have not heard from you by 2026-09-08 18:00 UTC I will push those three to your branch myself** and merge, with your authorship intact and @v01dst credited for item 3. `maintainerCanModify` is on. I would rather they were your commits.

One heads-up on sequencing that is not your problem to solve: #712 is open against #695 and rewrites `src/soup_cli/data/formats.py` and `src/soup_cli/data/validator.py`, the same two files. Whichever lands second will need a rebase. I will order them and tell you if it lands on you.

v01dst:
@BetterAndBetterII — the list guard from #679 that the maintainer wants folded in, ready to apply so you don't have to reverse-engineer it from my branch.

**The problem in your current code:** `for msg in messages` iterates the value directly. A `str` iterates char-by-char (chars aren't dicts → raises — accidental pass), an `int` raises TypeError (→ drop), but an **empty** non-list iterable (`""`, `{}`, `[]`-shaped-but-not) iterates zero times and **silently passes validation**. Both `{"messages": ""}` and `{"messages": {}}` are JSON-reachable and `detect_format` routes them to chatml, the default format.

**The guard** (drop-in for your converter checks):

```python
if not isinstance(messages, list):
    raise ValueError(f"{fmt} row 'messages' must be a list")
for msg in messages:
    if not isinstance(msg, dict):
        raise ValueError(f"{fmt} message must be a dict")
```

The `isinstance` check must come **before** the element loop — that's the whole fix.

**The two fixtures** (strong form — assert the row is dropped, not merely that nothing raises; my branch's version passes with a `continue` mutant otherwise):

```python
def test_chatml_empty_string_messages_dropped():
    assert format_to_messages({"messages": ""}, "chatml") is None

def test_chatml_dict_messages_dropped():
    assert format_to_messages({"messages": {}}, "chatml") is None
```

(and the audio/video twins if your parametrization covers them — same shapes, `format_to_messages(row, fmt) is None`.)

For reference, my branch's implementation is `src/soup_cli/data/formats.py::_require_str_messages` at `876c907` on #679 — take whatever fits your structure.

BetterAndBetterII:
Thank you for the detailed review and follow-up! All three items are now addressed.

MakazhanAlpamys:
Merged as `06773c5`. Thank you, @BetterAndBetterII — and the reason this took as long as it did is entirely on me, not on your patch.

**Verified on the merged tree.** The two guards are pinned **independently**, which is the property I was actually waiting for:

```
remove @v01dst's isinstance(messages, list) guard  -> 6 failed
remove your element guard                          -> 14 failed
```

Disjoint sets. One guard is not masking the other, and neither can be deleted quietly.

Zero test lines removed. Merged sequentially with four other PRs onto current `main` in a scratch worktree first: no conflicts, 350 passed.

**What I want on the record about how you handled the comparison.** Two good PRs on one issue is a bad situation for everyone, and I made it worse by leaving #676 labelled `good first issue` for fourteen hours while both were already open. You lost the measurement, were told your PR still had a gap the other one closed, and then folded @v01dst's guard in with credit — plus the two fixtures it needed — **within two minutes** of it being handed over. No argument, no ownership. That is the outcome I wanted and could not have got from either PR alone.

**And one correction you are owed, because it changes what "losing" meant.** The metric I decided on — validator/loader agreement, 4/7 against 0/7 — has since been superseded by #712, which makes `soup data validate` run the real converters. I re-measured after it landed:

```
main + #678 (with #712):  4/4 AGREE
main + #679 (with #712):  4/4 AGREE
```

So the axis that separated the two PRs no longer exists. Your merge stands on its own merits and on @v01dst having closed theirs by their own judgement — but nobody should think there was a durable technical gap, because there was not.

Your `CONTRIBUTORS.md` entries have been rewritten to the file's house style in `9f1e48b`, keeping both of you and adding the substance. One thing I changed rather than kept: `gh api users/v01dst` gives **Mahmoud Ahmed**, and the entry named them by login.

Issue #677, opened 2026-09-04
No guard catches a recipe that names a Hugging Face repo which does not exist


MakazhanAlpamys:
Split out of #661, whose fourth acceptance item asks for "a guard that would catch the next one, or a stated decision not to have one". #666 repairs the two broken ids and pins them as literals, which is the right scope for that PR. This issue is the general guard.

## The problem

The catalog ships **163 recipes**, each naming a Hugging Face repo in two places — `RecipeMeta.model` and the YAML `base:`. Nothing checks that any of them resolves. #661 found two that did not, and both had been shipped and released:

| recipe | declared | state |
|---|---|---|
| `qwen3-8b-sft-mlx` | `mlx-community/Qwen3-8B-Instruct-4bit` | `RepositoryNotFoundError` |
| `gemma3-9b-sft-mlx` | `mlx-community/gemma-3-9b-it-4bit` | `RepositoryNotFoundError` |

The failure is silent until a user runs `soup recipes use <name>` and then `soup train`, at which point it surfaces as a download error rather than as "this recipe was never valid". `validate-recipes` in CI parses the YAML through the real schema, which is why it stayed green — a nonexistent repo id is a perfectly well-formed string.

## Why this is not simply a CI check

Three constraints, and getting any of them wrong makes the guard worse than nothing:

1. **A bare 401 is not evidence.** #661's opening section demonstrates this: an invented repo returns the same 401 as a real private one. The authoritative signal is `huggingface_hub`'s `RepositoryNotFoundError` against an explicitly unauthenticated client (`HfApi(token=False)`, and be careful that `HF_TOKEN` is not set in the runner). Verified here with a negative control:

   ```
   mlx-community/Qwen3-8B-4bit              EXISTS gated=False
   mlx-community/zz-invented-control-xyz    RepositoryNotFoundError
   ```

2. **Missing and gated are different, and only one is a bug.** A gated repo (Llama, Gemma) resolves for an authorised user and 401s for an anonymous one. A guard that treats gated as broken would fire on legitimate recipes forever, and a guard that fires on correct code is one people learn to ignore — this repo has that exact lesson written down from #404. `model_info` exposes `gated`; the check must read it rather than infer from the status code.

3. **It must not run in the PR matrix.** 163 network calls on every pull request makes every unrelated PR depend on Hub availability and rate limits. An unrelated docs PR going red because the Hub had a bad minute trains people to re-run rather than to read.

## Suggested shape

[...]

PR #679, opened 2026-09-04
fix(data): drop chatml/audio/video rows with a non-dict message

### LINKED ISSUE #676: chatml / audio / video converters silently keep a non-dict message instead of dropping the row

v01dst:
## Summary

`format_to_messages` documents a drop contract: a malformed row routes to the drop path and returns `None`, so one bad line in a JSONL file is skipped rather than corrupting the dataset. Three converters do not honour it — a non-dict element inside `messages` passes through verbatim:

```python
format_to_messages({"messages": ["hello"]}, "chatml")   # -> {'messages': ['hello']}
format_to_messages({"messages": ["hello"]}, "audio")    # -> {'messages': ['hello']}
format_to_messages({"messages": ["hello"]}, "video")    # -> {'messages': ['hello']}
```

Expected in each case is `None`. `chatml` is what `detect_format` returns for a bare `{"messages": [...]}` row, so it is the **default** path for the most common dataset shape.

## Fix

The three converters now share one `_require_str_messages` validator that raises `ValueError` for a non-list `messages` value or a non-dict element — which `format_to_messages`'s existing drop path already routes to `None`. Valid rows convert exactly as before. This mirrors the multimodal element check added in #670.

## Tests

Strong form throughout — asserting `result is None`, so replacing the raise with `continue` (keeping the malformed message) would fail them:

- `tests/test_chatml_drop.py` (new): non-dict element, non-list `messages`, and valid-row round-trips for both `chatml` and `video`
- `tests/test_audio.py`: non-dict element and non-list `messages` for `audio`
- All pre-existing audio conversion tests pass unchanged

Includes `changelog.d/0.74.0/676.fixed.md` per repo convention.

Fixes #676

### DIFF
--- a/src/soup_cli/data/formats.py
+++ b/src/soup_cli/data/formats.py
@@ -182,8 +182,8 @@ def _convert_sharegpt(row: dict) -> dict:
     return {"messages": messages}
 
 
-def _require_str_messages(row: dict, fmt: str) -> list[dict]:
-    """Validate ``row["messages"]`` is a list whose elements are all dicts.
+def _require_str_messages(messages: object, fmt: str) -> list[dict]:
+    """Validate a ``messages`` value is a list whose elements are all dicts.
 
     chatml, audio and video pass ``messages`` through verbatim, so a row
     like ``{"messages": ["hello"]}`` would otherwise flow into the training
@@ -191,7 +191,6 @@ def _require_str_messages(row: dict, fmt: str) -> list[dict]:
     Raising ``ValueError`` routes the row to ``format_to_messages``'s drop
     path, matching ``_convert_multimodal``'s element check (#676).
     """
-    messages = row["messages"]
     if not isinstance(messages, list):
         raise ValueError(f"{fmt} row 'messages' must be a list")
     for msg in messages:
@@ -202,7 +201,7 @@ def _require_str_messages(row: dict, fmt: str) -> list[dict]:
 
 def _convert_chatml(row: dict) -> dict:
     # Already in the right format
-    return {"messages": _require_str_messages(row, "chatml")}
+    return {"messages": _require_str_messages(row["messages"], "chatml")}
 
 
 def _convert_dpo(row: dict) -> dict:
@@ -291,7 +290,7 @@ def _convert_audio(row: dict) -> dict:
     audio = row["audio"]
     if not isinstance(audio, str) or not audio.strip():
         raise ValueError("Audio row must have a non-empty 'audio' field")
-    messages = _require_str_messages(row, "audio")
+    messages = _require_str_messages(row["messages"], "audio")
     if len(messages) < 1:
         raise ValueError("Audio row must have a 'messages' list with at least one message")
     return {"messages": messages, "audio": audio}
@@ -587,7 +586,7 @@ def _convert_video(row: dict) -> dict:
         raise ValueError("video row 'video' must not contain null bytes")
     if len(video) > 2048:
         raise ValueError("video row 'video' must be <= 2048 chars")
-    messages = _require_str_messages(row, "video")
+    messages = _require_str_messages(row.get("messages") or [], "video")
     return {"video": video, "messages": messages}
 
 
--- a/src/soup_cli/data/formats.py
+++ b/src/soup_cli/data/formats.py
@@ -182,9 +182,27 @@ def _convert_sharegpt(row: dict) -> dict:
     return {"messages": messages}
 
 
+def _require_str_messages(row: dict, fmt: str) -> list[dict]:
+    """Validate ``row["messages"]`` is a list whose elements are all dicts.
+
+    chatml, audio and video pass ``messages`` through verbatim, so a row
+    like ``{"messages": ["hello"]}`` would otherwise flow into the training
+    set as-is instead of being dropped like every other malformed row.
+    Raising ``ValueError`` routes the row to ``format_to_messages``'s drop
+    path, matching ``_convert_multimodal``'s element check (#676).
+    """
+    messages = row["messages"]
+    if not isinstance(messages, list):
+        raise ValueError(f"{fmt} row 'messages' must be a list")
+    for msg in messages:
+        if not isinstance(msg, dict):
+            raise ValueError(f"{fmt} message must be a dict")
+    return messages
+
+
 def _convert_chatml(row: dict) -> dict:
     # Already in the right format
-    return {"messages": row["messages"]}
+    return {"messages": _require_str_messages(row, "chatml")}
 
 
 def _convert_dpo(row: dict) -> dict:
@@ -273,8 +291,8 @@ def _convert_audio(row: dict) -> dict:
     audio = row["audio"]
     if not isinstance(audio, str) or not audio.strip():
         raise ValueError("Audio row must have a non-empty 'audio' field")
-    messages = row["messages"]
-    if not isinstance(messages, list) or len(messages) < 1:
+    messages = _require_str_messages(row, "audio")
+    if len(messages) < 1:
         raise ValueError("Audio row must have a 'messages' list with at least one message")
     return {"messages": messages, "audio": audio}
 
@@ -569,7 +587,7 @@ def _convert_video(row: dict) -> dict:
         raise ValueError("video row 'video' must not contain null bytes")
     if len(video) > 2048:
         raise ValueError("video row 'video' must be <= 2048 chars")
-    messages = row.get("messages") or []
+    messages = _require_str_messages(row, "video")
     return {"video": video, "messages": messages}
 
 
--- a/changelog.d/0.74.0/676.fixed.md
+++ b/changelog.d/0.74.0/676.fixed.md
@@ -1,0 +1,13 @@
+- **A non-dict message in a `chatml`, `audio`, or `video` row passed through verbatim instead of
+  dropping the row (#676).**
+  `_convert_chatml` returned `row["messages"]` as-is, and `_convert_audio` / `_convert_video`
+  copied the list after checking only its shape, so a row like `{"messages": ["hello"]}` — a bare
+  string where every message must be a dict — flowed into the training set untouched.
+  `format_to_messages` documents a drop contract: a malformed row returns `None` and is skipped,
+  so one bad line in a JSONL file never corrupts the dataset. These three converters silently
+  kept the malformed message instead. `chatml` is the default path for a bare
+  `{"messages": [...]}` row, which makes it the most-reachable case. The fix shares one
+  `_require_str_messages` validator across all three converters: a non-list `messages` or a
+  non-dict element raises `ValueError`, which the existing drop path already routes to `None`.
+  The loader-level behaviour is unchanged — `_format_rows` was already skipping `None` rows, the
+  converters were the ones not producing them.
--- a/tests/test_audio.py
+++ b/tests/test_audio.py
@@ -367,3 +367,21 @@ output: ./output_audio
         assert cfg.data.format == "audio"
         assert cfg.data.audio_dir == "./data/wav"
         assert cfg.output == "./output_audio"
+
+
+class TestAudioNonDictMessageDropped:
+    """#676 — a non-dict element in `messages` routes to the drop path."""
+
+    def test_audio_non_dict_message_returns_none(self):
+        from soup_cli.data.formats import format_to_messages
+
+        row = {"audio": "clip.wav", "messages": ["hello"]}
+        result = format_to_messages(row, "audio")
+        assert result is None
+
+    def test_audio_message_list_not_a_list_returns_none(self):
+        from soup_cli.data.formats import format_to_messages
+
+        row = {"audio": "clip.wav", "messages": "hello"}
+        result = format_to_messages(row, "audio")
+        assert result is None
--- a/tests/test_chatml_drop.py
+++ b/tests/test_chatml_drop.py
@@ -38,3 +38,14 @@ def test_video_valid_messages_still_convert():
 def test_video_messages_not_a_list_returns_none():
     result = format_to_messages({"video": "clip.mp4", "messages": 5}, "video")
     assert result is None
+
+
+def test_video_missing_messages_converts_with_empty_list():
+    """Pre-existing contract: video rows need no 'messages' key (#679 review)."""
+    result = format_to_messages({"video": "clip.mp4"}, "video")
+    assert result == {"video": "clip.mp4", "messages": []}
+
+
+def test_video_messages_null_converts_with_empty_list():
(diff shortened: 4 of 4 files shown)


src/soup_cli/data/formats.py
```diff
@@ -569,7 +587,7 @@ def _convert_video(row: dict) -> dict:
         raise ValueError("video row 'video' must not contain null bytes")
     if len(video) > 2048:
         raise ValueError("video row 'video' must be <= 2048 chars")
-    messages = row.get("messages") or []
+    messages = _require_str_messages(row, "video")
```
AmirF194: Ran this against ccd3c41a in a clean python:3.11-slim container.

`_require_str_messages(row, "video")` (line 590) reads `row["messages"]` directly instead of the old `row.get("messages") or []`. A video row with no `messages` key at all now raises `KeyError`, which `format_to_messages`'s `except (KeyError, TypeError, IndexError, ValueError)` catches and turns into `None`, so the whole row gets dropped instead of converting with an empty messages list like it does today.

```
$ docker run --rm -v /repo:/repo:ro python:3.11-slim bash -c \
  "pip install -q rich && python -c \"
import sys; sys.path.insert(0,'/repo/src')
from soup_cli.data.formats import format_to_messages
print(format_to_messages({'video': 'clip.mp4'}, 'video'))\""

PR head ccd3c41a -> None
main   ef2e02b5  -> {'video': 'clip.mp4', 'messages': []}
```

None of the new tests cover a video row that omits `messages` entirely, only present-but-wrong-type or present-with-a-non-dict-element, so this case is not pinned either way.

If dropping a keyless-messages video row is intended, it is a behavior change worth a line in the changelog alongside the other two. If not, `row.get("messages", [])` ahead of the type/element checks would keep today's default for the missing-key case while still catching the cases this PR targets.

v01dst: Good catch — confirmed regression, now fixed (876c907). The validator takes the messages *value* instead of reading the row, so each caller keeps its own missing-key contract: chatml and audio stay strict (row["messages"], KeyError → drop, matching their pre-PR behavior), while video keeps the .get("messages") or [] default and converts with an empty list. Regression tests cover both the missing key and the explicit null, and the full formats/audio suites pass (52 tests).

AmirF194: Confirmed, thanks for the fix.

MakazhanAlpamys:
@v01dst — welcome, and thank you. I put your PR and #678 through an identical battery and I am publishing the **same numbers in both threads**, including where each is better than the other, because that is the only honest way to resolve two good PRs on one issue.

**The decision goes to #678, on one measurement.** I will set out that measurement, and then the six places where your PR is better — because several of them are going into `main` under your name.

## First, a correction I owe you

@abdulwaarith0 stood down from this issue citing #679 as having "got here first". That is wrong: **#678 opened at 19:48 and yours at 21:18**, ninety minutes apart. It changes nothing about the outcome — I did not decide on arrival order — but I would rather the record be accurate. And the reason two PRs collided at all is my fault: I left #676 labelled `good first issue` for fourteen hours while two pull requests were already open against it.

## The identical battery

Both branches rebased onto `a08e858` before every measurement. Clean-main baseline: 189 passed in the narrow set, 862 in the keyword set, **no pre-existing failures in this scope**, so every result is attributable.

| | **#678** | **#679 (yours)** |
|---|---|---|
| Defect 1 — 4 bad shapes x 3 formats | YES | YES |
| Residual behaviour gap | **`{"messages": ""}` and `{"messages": {}}` still kept** | **none** |
| Defect 2 — validate agrees with the loader | 3/7 -> **4/7** | 3/7 -> **0/7** |
| Mutations killed | **10 / 10** | 9 / 10 |
| Tests added | 16 | 10 |
| Tests removed or weakened | 0 | 0 |
| Over-reach on 300 dataset/format combos | 40 changes, all correct, exit codes unmoved | 0 |
| Applies to current `main` | no — 1 test red | **yes — 199 passed** |
| ruff | clean | clean |

## The measurement that decided it

**Defect 2, which #676 reports as its second half.** With identical fixtures — 4 good rows, 1 malformed — and the validator compared against what `load_dataset` actually keeps:

| format | main | #678 | #679 |
|---|---|---|---|
| chatml | 5/5 vs 5 — agree *(on a shared bug)* | **4/5 vs 4 — agree** | 5/5 vs 4 — **disagree** |
| audio | agree *(on a bug)* | **agree** | **disagree** |
| video | agree *(on a bug)* | **agree** | **disagree** |
| multimodal | disagree | **agree** | disagree |
| **total agreeing** | 3 / 7 | **4 / 7** | **0 / 7** |

This is the part I want to be precise about, because it is not simply "yours does less". By fixing the loader and not the validator, your PR converts main's *accidental* agreement on chatml, audio and video into a genuine mismatch — on the project's **default** format, the one `detect_format` returns for a bare `{"messages": [...]}`. #676's own text calls that shape "worse than no validator, because it moves the failure from validate time to the middle of a training run".

To be fair to your side of it: a user's **data** is safer under your PR either way, because the corrupt row is dropped. It is the validator's honesty that regresses, not the training set. And you could reasonably have expected that to be out of scope, since @abdulwaarith0 split it into #695 while your PR was already open. I do not hold that against the PR; it is the reason this was close rather than obvious.

## Where your PR is better than #678 — six things, measured

1. **You have no residual gap, and #678 does.** I verified this myself rather than taking the battery's word for it:

   ```
   #678:  {"messages": ""} -> {'messages': ''}      {"messages": {}} -> {'messages': {}}
   #679:  {"messages": ""} -> None                  {"messages": {}} -> None
   ```

   Both shapes are JSON-reachable and both are detected as `chatml`. Through the real loader on a 3-row file, #678 keeps 3 and yours keeps 1. #678 rejects a non-list by *iterating* it, which works for `"hello"` and `42` by accident and fails silently for any **empty** non-list iterable. Your explicit `isinstance(messages, list)` closes it. **This is the same defect class #676 is about, on the default path** — and it is the single most valuable thing in either PR.

2. **Your helper matches the reference converters' shape.** `_convert_input_output:549` and the merged `_convert_multimodal` both do a list check *then* an element check. Yours does both; #678 has only the element half.

3. **You are branched on current `main` and green.** #678 is branched pre-#670 and has a test that fails today.

4. **Zero over-reach.** You touched nothing user-facing beyond the converters.

5. **You pinned a contract you did not have to.** Your second commit adds tests for video's "no `messages` key -> `[]`" behaviour. #678 preserves that behaviour and never protects it.

6. **Smaller, single-purpose, easier to revert.**

## What is going into `main` under your name

**Your `isinstance(messages, list)` guard is being lifted into #678 verbatim in effect, credited to you** in the merge commit body and in `CONTRIBUTORS.md` — along with the two tests it needs, which I am asking #678's author to add because they are the coverage your own PR was missing (see below). Without your PR, that gap ships.

## One finding against yours, and it is the interesting one

**The guard that makes your PR better is not pinned by any test.** Deleting `isinstance(messages, list)` passes all 199 of your tests. The reason is subtle and worth keeping: `{"messages": "hello"}` cannot discriminate it, because iterating a string yields non-dict characters and the element check catches it anyway. **Only an empty non-list iterable can** — `{"messages": ""}` or `{"messages": {}}`. So the one case your guard uniquely handles is the one case you did not test.

That is not a criticism so much as the neatest illustration in this whole comparison of why mutation testing is worth the time: your implementation was right and your test suite could not tell.

## Your choice, and I would rather you pick

I do not want you to have written the best single idea here and get nothing merged. Two ways:

- **You open a small follow-up PR** adding the list guard plus the `""` / `{}` tests on top of #678 once it lands. Your commit, your merge, your `CONTRIBUTORS.md` entry for it. I will hold #695 or another issue for you too.
- **Or I lift it into #678** with credit as described, and you take a fresh issue instead — **#695** is the validator half and is currently unclaimed after @abdulwaarith0 offered to file rather than take it; **#699** and **#677** are also open and unclaimed.

Say which you prefer. I am leaving this PR open until you answer rather than closing it and presenting you with a decision.

Two more things worth your knowing, since both PRs hit them: the fragment is named `676.fixed.md` after the *issue*, where `changelog.d/README.md` asks for the **pull request** number — and the reason it asks is exactly this, two PRs on one issue colliding on one filename. Yours also does not carry `by @v01dst`; outside work gets its author named in the changelog here.

MakazhanAlpamys:
Pointer, so this is not decided in a thread you are not watching: @swalla02 has asked for **#699** — one of the three issues I said on this PR I would hold for you.

**It is yours until 2026-09-06 18:00 UTC**, then it passes. Full context on #699, including their offer to hand over their branch and measurements if you take it rather than make you redo the probes, which is a better outcome than either of the alternatives.

Nothing here is conditional on that answer: **your `isinstance(messages, list)` guard is going into `main` under your name regardless**, and #677 and #702 are both unclaimed if you would rather take something fresh. "Not interested" costs you one word and I will not read anything into it.

Naming a time because I have created two of these races in two days, and the fix for both is the same — state when a hold expires instead of leaving it open-ended.

v01dst:
Closing in favour of #678 — the maintainer's identical battery picked it as the base on both defect coverage (validator/loader agreement 4/7 vs my 0/7) and mutations killed (10/10 vs 9/10), and that is the right call. #670's bar exists to be met, not approximated.

The one thing mine carried that theirs still needs — the `isinstance(messages, list)` guard before the element loop, with fixtures for `{"messages": ""}` and `{"messages": {}}` — is handed over in the #678 thread so it lands with credit where the maintainer assigned it.

Thanks @MakazhanAlpamys for the unusually rigorous review process, and good luck to @BetterAndBetterII on the merge.

MakazhanAlpamys:
@v01dst — thank you for how you closed this. Standing down on your own measurement, and then handing the guard over in the other thread ready to apply so @BetterAndBetterII did not have to reverse-engineer it from your branch, is a better outcome than either of the alternatives I had. They applied it within two minutes, and I confirmed by mutation on the merged tree that your list guard and their element guard are pinned **independently** — removing yours fails 6 tests, removing theirs fails 14. Both are defended on their own, which is exactly what I wanted and could not get from either PR alone.

**One correction I owe you, and it is not a small one.**

The measurement that decided this against you has been superseded, within hours, by a different merge. #712 landed at `bedd4da` and makes `soup data validate` run the real converters, so validator/loader agreement is now achieved **by construction** regardless of which converter fix is underneath. I re-measured rather than reasoned:

```
main + #678 (with #712):  4/4 AGREE
main + #679 (with #712):  4/4 AGREE
```

Your 0/7 was an artifact of the old key-presence validator, and it is gone. So the single metric I used to pick between two good PRs no longer distinguishes them.

This does not reverse anything — you closed #679 by your own judgement before I found this, #678 is the base and is nearly merged, and the guard lands with your name in the commit body, the changelog fragment and `CONTRIBUTORS.md`. But you should not go away believing your PR lost on a durable technical gap, because it did not. It lost on a number that was true for about fourteen hours.

**And a second disclosure, this one a process failure of mine.**

On 2026-09-05 at 18:36 UTC I named **#677** to you here as something on the table. I put an expiry on #699 and not on that one — the same mistake, three days running. @Srinivasan8888 claimed #677 this morning with a correct overlap check and a genuinely good design, so I have assigned it to them. If you wanted it, say so and I will take it back to them rather than decide behind your back; but they did nothing wrong and the default is that it stays theirs.

**What is still actually yours:**

- **#699** until **2026-09-06 18:00 UTC** — unchanged. @swalla02 has offered to hand over their branch and measurements rather than make you redo the probes.
- **#702** — the streaming half of the interleave leak, unclaimed. #701 closes the eager paths and provably cannot reach this one: on that branch, `streaming: true` + `over` + `val_split 0.2` still leaks 10 of 30 distinct val rows. It needs a genuinely different mechanism.

Both are open offers, not obligations, and "not interested" still costs you one word.

Issue #686, opened 2026-09-05
MLX SFT ignores warmup, scheduler, and configured weight decay


Amix29:
## Description

The MLX backend validates `warmup_ratio`, `scheduler`, and `weight_decay`, but constructs its
optimizer with only a scalar learning rate:

```python
optim.AdamW(learning_rate=float(cfg.training.lr))
```

No warmup/scheduler is built in this path, and configured weight decay is not passed.

Affected revision: `ef2e02b51a1b34546823edce55715d28234805cb` (Soup 0.74.0).

Relevant code:

- [optimizer construction](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/trainer/mlx_sft.py#L312-L330)
- [validated defaults and fields](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/config/schema.py#L1096-L1098)
- [scheduler field](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/config/schema.py#L1233)

## Steps to reproduce

Configure an MLX run with:

```yaml
training:
  lr: 0.0002
  weight_decay: 0.3
  warmup_ratio: 0.2
  scheduler: cosine
```

Intercept the real wrapper call to the optimizer constructor:

```text
forwarded kwargs: {"learning_rate": 0.0002}
```

## Expected behavior

The effective optimizer and learning-rate curve should match the validated Soup configuration,
or unsupported choices should be rejected clearly.

## Impact

MLX training silently runs a different recipe. Custom weight decay is lost, and a scalar LR
does not implement the configured warmup followed by cosine decay. This can affect stability,
convergence, and backend comparisons.

## Proposed fix and acceptance criteria

- [ ] Build the requested MLX schedule with a clearly defined optimizer-step unit.
- [ ] Pass configured weight decay.
- [ ] Map supported optimizer names and reject unsupported names instead of silently using AdamW.
- [ ] Test the observed LR at the first, warmup-boundary, and final updates.
- [ ] Store the effective optimizer/schedule configuration in adapter metadata.

## Environment

Argument-construction probe run under Python 3.12.2 against the original Soup wrapper. The
optimizer was replaced by a recorder; no numerical Metal update or final-quality claim is made.

MakazhanAlpamys:
[...]

Issue #683, opened 2026-09-05
MLX SFT ignores train_on_responses_only


Amix29:
## Description

`data.train_on_responses_only` defaults to `true` and promises that only assistant content
contributes to SFT loss. The MLX wrapper does not pass any prompt mask to mlx-lm's dataset
constructor and writes `"mask_prompt": false` into adapter metadata.

Affected revision: `ef2e02b51a1b34546823edce55715d28234805cb` (Soup 0.74.0), with mlx-lm
0.31.3 behavior.

Relevant code:

- [Soup's responses-only contract](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/config/schema.py#L383-L389)
- [MLX `TrainingArgs` passed to `create_dataset`](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/trainer/mlx_sft.py#L295-L310)
- [metadata hardcodes `mask_prompt: false`](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/trainer/mlx_sft.py#L333-L358)

## Steps to reproduce

Use a synthetic conversation with six prompt tokens and four assistant tokens, and set
`train_on_responses_only: true`.

```text
offset produced by the current MLX wrapper: 0
offset in a mask-enabled control:            6
prompt targets supervised by current path:  5
prompt targets supervised by control:       0
```

## Expected behavior

Non-assistant tokens must not contribute to the MLX SFT loss when the Soup option is enabled.

## Impact

The default MLX recipe trains on a different token distribution than configured, which can
change instruction-following behavior and invalidates direct comparisons with the Transformers
backend.

## Proposed fix and acceptance criteria

Do not map the option blindly to mlx-lm 0.31.3's `mask_prompt=True`: its `ChatDataset` masks a
single prefix before the last message and is not equivalent to preserving every assistant span
in a multi-turn conversation. Implement Soup-compatible per-token masking or reject unsupported
semantics explicitly.

- [ ] Single-turn prompt tokens have zero loss weight.
- [ ] Every assistant turn remains supervised in multi-turn conversations.
- [ ] System and user turns remain excluded.
- [ ] `train_on_messages_with_train_field` is either supported equivalently or rejected.
- [ ] Effective masking is recorded accurately in adapter metadata.

[...]

PR #696, opened 2026-09-05
fix(mlx): wire gradient_accumulation_steps into the SFT TrainingArgs (#684)

### LINKED ISSUE #684: MLX SFT ignores gradient_accumulation_steps
## Description

The MLX SFT wrapper does not forward `training.gradient_accumulation_steps` to mlx-lm's
`TrainingArgs.grad_accumulation_steps`. mlx-lm therefore uses its default value of 1, and Soup
also writes 1 into the adapter metadata regardless of the requested configuration.

Affected revision: `ef2e02b51a1b34546823edce55715d28234805cb` (Soup 0.74.0), with mlx-lm
0.31.3 behavior.

Relevant code:

- [Soup default is 4](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/config/schema.py#L1051)
- [`TrainingArgs` construction omits accumulation](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/trainer/mlx_sft.py#L276-L303)
- [metadata hardcodes 1](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/trainer/mlx_sft.py#L333-L358)

## Steps to reproduce

Configure:

```yaml
backend: mlx
training:
  batch_size: 2
  gradient_accumulation_steps: 8
```

Inspect the `TrainingArgs` received by mlx-lm:

```text
requested gradient accumulation: 8
forwarded gradient accumulation: 1
requested single-device effective batch: 16
effective batch represented by arguments: 2
```

`batch_size` and `max_seq_length` survive the same boundary, isolating the omission.

## Expected behavior

The effective batch and optimizer-update cadence should match the validated Soup configuration.

## Impact

MLX runs silently execute a different optimization recipe. This changes update frequency,
effective batch size, and potentially convergence. The reproduction does not claim an 8x
throughput difference.

- [ ] Forward the resolved accumulation value to mlx-lm.
- [ ] Define iteration/epoch/checkpoint units consistently as micro-batches or optimizer steps.
- [ ] Handle the final incomplete accumulation group without silently dropping gradients.
- [ ] Record the effective value in adapter metadata.
- [ ] Add a test that observes optimizer update count, not only argument construction.

## Environment

Contract probe run under Python 3.12.2 against the original Soup wrapper and the upstream
mlx-lm 0.31.3 `TrainingArgs`. No Metal optimization step was executed.

AmirF194:
Fixes #684

`MLXSFTTrainerWrapper.train()` built mlx-lm's `TrainingArgs` without a `grad_accumulation_steps` kwarg, so mlx-lm's own dataclass default of 1 always applied no matter what `training.gradient_accumulation_steps` was set to. The written `adapter_config.json` hardcoded the same value to 1 too, so there was no way to tell from the output that it had drifted.

Both now read the configured value once and pass it through to `TrainingArgs` and the metadata write.

Added `tests/test_issue684_mlx_grad_accumulation.py`, same fake-mlx harness as `test_issue634_mlx_resume.py` so it runs `train()` for real without Apple Silicon. Fails on main with `AttributeError: '_FakeTrainingArgs' object has no attribute 'grad_accumulation_steps'`, passes on the branch. Ran the full mlx-related suite (test_mlx_backend.py, test_mlx_dispatch.py, test_issue634_mlx_resume.py, etc.) plus `ruff check` clean.

Not run: an actual training loop on real Metal, no Apple Silicon here. The fix is at the same argument-construction level `_apply_lora`'s target-modules fix (#392) used for the same reason.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,18 @@ reproducing 70+ versions of notes.
 
 ## [Unreleased]
 
+### Fixed
+
+- **MLX SFT ignored `training.gradient_accumulation_steps` (#684).** The
+  wrapper built mlx-lm's `TrainingArgs` without a `grad_accumulation_steps`
+  kwarg, so mlx-lm's own dataclass default of 1 always applied regardless
+  of the configured value, silently changing the effective batch size and
+  optimizer-update cadence on every MLX run. The written
+  `adapter_config.json` also hardcoded `grad_accumulation_steps: 1`, so the
+  drift could not be detected from the output afterwards either. Both now
+  carry the value `training.gradient_accumulation_steps` actually resolves
+  to.
+
 ## [0.74.0] - 2026-09-04
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,18 +12,6 @@ reproducing 70+ versions of notes.
 
 ## [Unreleased]
 
-### Fixed
-
-- **MLX SFT ignored `training.gradient_accumulation_steps` (#684).** The
-  wrapper built mlx-lm's `TrainingArgs` without a `grad_accumulation_steps`
-  kwarg, so mlx-lm's own dataclass default of 1 always applied regardless
-  of the configured value, silently changing the effective batch size and
-  optimizer-update cadence on every MLX run. The written
-  `adapter_config.json` also hardcoded `grad_accumulation_steps: 1`, so the
-  drift could not be detected from the output afterwards either. Both now
-  carry the value `training.gradient_accumulation_steps` actually resolves
-  to.
-
 ## [0.74.0] - 2026-09-04
 
 ### Added
--- a/benchmarks/harness/mlx_sft_smoke.py
+++ b/benchmarks/harness/mlx_sft_smoke.py
@@ -169,6 +169,10 @@ training:
   epochs: {epochs}
   lr: 1e-4
   batch_size: 1
+  # Pinned: the schema default (4) would round `iters` down to a whole
+  # accumulation window (#696), moving the step count this harness reports
+  # out from under any published benchmark record that assumes 1:1 rows-to-iters.
+  gradient_accumulation_steps: 1
   lora:
     r: 8
     alpha: 16
--- a/changelog.d/0.74.0/696.fixed.md
+++ b/changelog.d/0.74.0/696.fixed.md
@@ -1,0 +1,18 @@
+- **MLX SFT ignored `training.gradient_accumulation_steps` (#696).** The wrapper built
+  mlx-lm's `TrainingArgs` without a `grad_accumulation_steps` kwarg, so mlx-lm's own
+  dataclass default of 1 always applied regardless of the configured value, silently
+  changing the effective batch size and optimizer-update cadence on every MLX run. The
+  written `adapter_config.json` also hardcoded `grad_accumulation_steps: 1`, so the drift
+  could not be detected from the output afterwards either. Both now carry the value
+  `training.gradient_accumulation_steps` actually resolves to. `iters` is also rounded
+  down to a whole number of accumulation groups (kept at a minimum of one group), because
+  mlx-lm only updates the optimizer on `it % accum == 0` and never flushes a trailing
+  partial group; without this a dataset smaller than the accumulation window trained for
+  zero optimizer updates.
+
+  **Anyone who never touched this field will see a different effective batch size after
+  upgrading.** `gradient_accumulation_steps` defaults to 4, so an MLX run that used to
+  update on every micro-batch now accumulates over 4 before updating: a 4x larger
+  effective batch size and roughly a quarter as many optimizer updates for the same
+  `iters`. This is the correct value for the schema default; a differently-converging run
+  after this upgrade is expected, not a regression.
--- a/changelog.d/0.74.0/696.fixed.md
+++ b/changelog.d/0.74.0/696.fixed.md
@@ -1,4 +1,4 @@
-- **MLX SFT ignored `training.gradient_accumulation_steps` (#696).** The wrapper built
+- **MLX SFT ignored `training.gradient_accumulation_steps` (#684 by @AmirF194 in #696).** The wrapper built
   mlx-lm's `TrainingArgs` without a `grad_accumulation_steps` kwarg, so mlx-lm's own
   dataclass default of 1 always applied regardless of the configured value, silently
   changing the effective batch size and optimizer-update cadence on every MLX run. The
--- a/src/soup_cli/trainer/mlx_sft.py
+++ b/src/soup_cli/trainer/mlx_sft.py
@@ -299,6 +299,7 @@ class MLXSFTTrainerWrapper:
         # Real eval cadence when a val split exists; otherwise the value is
         # irrelevant (val_dataset stays None and mlx-lm skips evaluation).
         steps_per_eval = max(1, iters // 4) if val_rows else max(1000, iters + 1)
+        grad_accumulation_steps = int(cfg.training.gradient_accumulation_steps)
         args = TrainingArgs(
             batch_size=batch_size,
             iters=iters,
@@ -307,6 +308,7 @@ class MLXSFTTrainerWrapper:
             steps_per_eval=steps_per_eval,
             steps_per_save=steps_per_save,
             adapter_file=str(output_dir / "adapters.safetensors"),
+            grad_accumulation_steps=grad_accumulation_steps,
         )
 
         train_dataset = CacheDataset(create_dataset(train_rows, self.tokenizer, args))
@@ -437,7 +439,7 @@ class MLXSFTTrainerWrapper:
                     )["lora_parameters"],
                     "mask_prompt": False,
                     "grad_checkpoint": False,
-                    "grad_accumulation_steps": 1,
+                    "grad_accumulation_steps": grad_accumulation_steps,
                 },
                 indent=2,
             )
--- a/src/soup_cli/trainer/mlx_sft.py
+++ b/src/soup_cli/trainer/mlx_sft.py
@@ -300,6 +300,15 @@ class MLXSFTTrainerWrapper:
         # irrelevant (val_dataset stays None and mlx-lm skips evaluation).
         steps_per_eval = max(1, iters // 4) if val_rows else max(1000, iters + 1)
         grad_accumulation_steps = int(cfg.training.gradient_accumulation_steps)
+        # mlx-lm updates the optimizer only when it % accum == 0 (trainer.py) and
+        # never flushes a partial group, so round iters down to a whole number of
+        # groups, keeping at least one group so a small dataset does not train
+        # for zero optimizer steps.
+        if grad_accumulation_steps > 1:
+            iters = max(
+                grad_accumulation_steps,
+                iters - (iters % grad_accumulation_steps),
+            )
         args = TrainingArgs(
             batch_size=batch_size,
             iters=iters,
--- a/tests/test_issue684_mlx_grad_accumulation.py
+++ b/tests/test_issue684_mlx_grad_accumulation.py
@@ -1,0 +1,162 @@
+"""#684: MLX SFT ignores gradient_accumulation_steps.
+
+MLXSFTTrainerWrapper.train() built mlx-lm's TrainingArgs without a
+grad_accumulation_steps kwarg, so mlx-lm's own dataclass default of 1
+always applied regardless of the validated schema field
+training.gradient_accumulation_steps (default 4, ge=1). The written
+adapter_config.json also hardcoded grad_accumulation_steps: 1 rather than
+reading back the effective value, so the drift could not even be detected
+from the output afterwards.
+
+Caveat that still matters: MLX only runs on Apple Silicon. Both tests here
+execute the real train() method body against minimal fake mlx / mlx_lm
+modules registered in sys.modules (the same harness
+test_issue634_mlx_resume.py uses) - real control flow, real call
+ordering, not source-text inspection - but the fakes stand in for real
+mlx-lm behaviour, not a substitute for an end-to-end run on real Metal.
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+import types
+
+
+def _install_fake_mlx(monkeypatch):
+    """Register minimal fake ``mlx`` / ``mlx_lm`` modules in ``sys.modules``
+    so ``MLXSFTTrainerWrapper.train()`` runs for real, without Apple
+    Silicon or the real packages installed. Same harness as
+    test_issue634_mlx_resume.py's helper of the same name.
+    """
+    mlx = types.ModuleType("mlx")
+    mlx_core = types.ModuleType("mlx.core")
+    mlx_optimizers = types.ModuleType("mlx.optimizers")
+    mlx_optimizers.AdamW = lambda **kwargs: object()
+
+    mlx_lm = types.ModuleType("mlx_lm")
+    mlx_lm_tuner = types.ModuleType("mlx_lm.tuner")
+
+    mlx_lm_tuner_callbacks = types.ModuleType("mlx_lm.tuner.callbacks")
+
+    class _FakeTrainingCallback:
+        pass
+
+    mlx_lm_tuner_callbacks.TrainingCallback = _FakeTrainingCallback
+
+    mlx_lm_tuner_datasets = types.ModuleType("mlx_lm.tuner.datasets")
+    mlx_lm_tuner_datasets.CacheDataset = lambda rows: rows
+    mlx_lm_tuner_datasets.create_dataset = lambda rows, tokenizer, args: rows
+
+    mlx_lm_tuner_trainer = types.ModuleType("mlx_lm.tuner.trainer")
+
+    class _FakeTrainingArgs:
+        def __init__(self, **kwargs):
+            self.__dict__.update(kwargs)
+
+    mlx_lm_tuner_trainer.TrainingArgs = _FakeTrainingArgs
+
+    captured_calls = []
+
+    def _fake_train(**kwargs):
+        captured_calls.append(kwargs)
+        callback = kwargs.get("training_callback")
+        if callback is not None:
+            callback.on_train_loss_report({"train_loss": 0.5})
+
+    mlx_lm_tuner_trainer.train = _fake_train
+
+    mlx_lm_tuner_utils = types.ModuleType("mlx_lm.tuner.utils")
+    mlx_lm_tuner_utils.linear_to_lora_layers = lambda model, num_layers, config: None
+
+    fake_modules = {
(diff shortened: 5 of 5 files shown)


MakazhanAlpamys:
@AmirF194 — welcome, and thank you for going straight at the MLX sweep. **The wiring itself is correct and well-tested.** I am blocking on something the wiring exposes rather than something it gets wrong, and it is worth reading carefully because as written this PR can make a run silently train nothing.

## The wiring is right — reproduced on both sides

Clean `main`, config `batch_size: 2, gradient_accumulation_steps: 8`:

```
requested                    : 8
TrainingArgs kwarg passed?   : False
TrainingArgs effective value : 1
adapter_config.json value    : 1
```

Your branch:

```
TrainingArgs kwarg passed?   : True
TrainingArgs effective value : 8
adapter_config.json value    : 8
```

Four mutations, four killed — deleting the kwarg, hardcoding it to 1 (the real bug shape), reverting only the metadata write (which fails **only** the metadata test, so the two halves are pinned independently), and misspelling the kwarg. Zero test lines removed, ruff clean, **94 passed** with **zero skips** — your fake-mlx harness really does execute here without Apple Silicon, which is the claim I check first on any MLX PR.

## Blocking 1 — this makes zero-optimizer-update runs reachable on the schema default

I read the upstream rule from the actual sdist rather than inferring it. `mlx_lm` 0.31.3, `mlx_lm/tuner/trainer.py:322`:

```python
it % grad_accum_steps == 0,
```

`it` runs `1..iters`, and `mlx_sft.py:283` computes `iters` as **micro-batches** — `epochs * ceil(rows / batch_size)`. So optimizer updates are `iters // accum`, and **if `iters < accum` the model is never updated at all**. `gradient_accumulation_steps` defaults to **4** at `schema.py:1051`, so this needs no opt-in from the user.

Simulated against your `iters` formula and upstream's rule:

| rows, batch, epochs, accum | iters | optimizer updates | micro-batches dropped |
|---|---|---|---|
| #684's own example (16, 2, 1, 8) | 8 | 1 | 0 |
| 20-row quickstart, default accum (20, 1, 3, 4) | 60 | 15 | 0 |
| **tiny set (3, 1, 1, 4)** | 3 | **0** | 3 |
| **10 rows, batch 4 (10, 4, 1, 4)** | 3 | **0** | 3 |
| non-multiple (26, 1, 1, 4) | 26 | 6 | **2** |
| accum = 1, i.e. today (26, 1, 1, 1) | 26 | 26 | 0 |

Rows 3 and 4 did 3 updates before this PR and do **0** after it. The run prints `MLX training complete`, writes `adapters.safetensors`, and reports a loss — with a completely untrained adapter. That is the same silent-no-training shape this project has already been bitten by twice, in #430 and #433, and it is the reason I check for it by habit.

None of this is your bug. It is latent in the interaction between our `iters` and upstream's rule, and it was simply unreachable while the value was pinned at 1. Your PR is what makes it reachable, so it is fair to close it here.

## Blocking 2 — the trailing group is silently discarded

Row 5 above: 26 iters at accum 4 gives 6 updates and **2 micro-batches whose gradients are accumulated and never applied** — upstream does not flush a partial group after the loop. That is #684's acceptance criterion 3 word for word: *"Handle the final incomplete accumulation group without silently dropping gradients."*

Both are closed by the same few lines, at `mlx_sft.py:295` right after the `int(...)`:

```python
grad_accumulation_steps = int(cfg.training.gradient_accumulation_steps)
# mlx-lm updates only when it % accum == 0 (trainer.py:322) and never flushes a
# partial group, so round iters down to a whole number of groups -- and keep at
# least one group, or a small dataset trains for zero optimizer steps.
if grad_accumulation_steps > 1:
    iters = max(grad_accumulation_steps, iters - (iters % grad_accumulation_steps))
```

`iters` is already written into `adapter_config.json`, so the adjusted value gets recorded for free.

## The test that would have caught it

#684's criterion 5 asks for a test that observes the **optimizer update count**, not only argument construction. Your three tests observe construction. The update count is testable entirely offline — apply upstream's `it % accum == 0` rule to the `iters` your wrapper actually computes, exactly as the table above does. Please add it alongside the fix; it is the test that turns "the kwarg is passed" into "the run actually trains".

## Should-fix — the changelog

`CHANGELOG.md` must not be edited directly. This project moved to per-PR fragments in #487: add `changelog.d/0.74.0/696.fixed.md` and delete the `CHANGELOG.md` hunk. Use the **pull-request** number in both the filename and the text (yours currently says `#684`).

This is not bookkeeping pedantry here — **you and #698 both edit `CHANGELOG.md` at the same offset, so as it stands the second of the two to merge takes a conflict**. Avoiding exactly that is why the fragment scheme exists. The concurrent #665 on this same file already does it correctly if you want a model.

While you are there: because the default is 4 and MLX was effectively pinned at 1, **every existing MLX user who never touched this field gets a 4x larger effective batch and a quarter the optimizer updates** after this merges. That is the correct fix, but it deserves a sentence so nobody reads a differently-converging run as a regression.

## Cross-PR, so you are not surprised

Your #696 and your #698 conflict with each other in three hunks of `mlx_sft.py` — and the third is dangerous. Each PR changes one line of the metadata dict and leaves the other's hardcoded, so resolving that hunk by taking either side wholesale **silently reverts the other PR's metadata fix**. CI would catch it, but resolve it take-both deliberately rather than by luck. #665 is orthogonal and clean against both.

I would like this one to land first of the two, since the `iters` rounding is the larger semantic change and should be the only thing moving in that region when it goes in.

AmirF194:
Both blockers are fixed. iters now rounds down to a whole number of accumulation groups (minimum one), using the exact formula from your comment, and the new tests assert the optimizer update count itself via mlx-lm's it % accum == 0 rule rather than just the TrainingArgs kwarg: one on the 3-row/accum-4 case (0 to 1 update), and 6 on the 26-row/accum-4 case (confirms the trailing 2 micro-batches are rounded off, not dropped mid-accumulation). Pushed 8c0c886.

Also moved the CHANGELOG.md edit to changelog.d/0.74.0/696.fixed.md and added the note about the 4x effective batch size change for anyone on the schema default. Agreed on landing this one first; will rebase #698 onto whatever this becomes.

MakazhanAlpamys:
@AmirF194 — **both blockers resolved, verified by execution. One line and I merge.**

## Blocker 1 — zero-optimizer-update runs

The formula landed exactly as specified. Measured through the **real CLI**, not a helper, at the schema default `batch 1 / accum 4`:

| rows | raw iters | shipped iters | optimizer updates |
|---|---|---|---|
| 3 | 3 | **4** | **1** (was 0) |
| 10 | 10 | 8 | 2 |
| 26 | 26 | 24 | 6 |
| 60 | 60 | 60 | 15 |

## Blocker 2 — trailing group discarded

26 -> 24 is a whole multiple, so nothing is accumulated and never applied. And the observability criterion is met: `_optimizer_updates()` applies upstream's `it % accum == 0` to the `iters` your wrapper actually computes, rather than to the raw count.

## Mutations — I re-ran them myself

```
baseline                                          6 passed
delete grad_accumulation_steps= from TrainingArgs 2 failed  -> killed
round UP instead of down                          1 failed  -> killed
remove the minimum-one clamp                      1 failed  -> killed
revert ONLY the metadata write                    1 failed  -> killed
```

The last one matters after the first review: the metadata write and the `TrainingArgs` kwarg read the same variable, which is exactly how the original suite let a deletion slip through. They now have **disjoint** kill sets, so each is defended on its own.

One thing checked against upstream rather than against a fake, because the stubs cannot prove it: the real `mlx-lm` 0.31.3 sdist declares `grad_accumulation_steps: int = 1` on `TrainingArgs`, so the spelling will not `TypeError` on real Metal. Same source confirms `it % grad_accum_steps == 0` with no partial flush — your diagnosis is exact — that `iterate_batches` is `while True:` so clamping `iters` above the micro-batch count is safe, and that trainer.py:386 always saves the final adapter, so rounding can never lose weights.

No `MagicMock` anywhere: `_FakeTrainingArgs` is a real class doing `__dict__.update(kwargs)`, so an unpassed kwarg raises `AttributeError`. That is why the first mutation actually bites. Zero test lines removed, ruff clean, CI green, ANSI-safe under `FORCE_COLOR=1`.

---

## The one line

`changelog.d/0.74.0/696.fixed.md` does not credit you. Filename, directory, category and the in-text `#696` are all correct; only the author is missing. The convention already in that directory is:

```
(#684 by @AmirF194 in #696)
```

Same on `698.fixed.md`. I know this is the third time this week I have asked you about a changelog line, and I would let a formatting slip go — but not the credit, because that field is the whole point of the file.

## Sequencing, so you push once instead of twice

**#696 merges first, then #698 needs a rebase.** I confirmed the conflict rather than guessing: `git merge-tree refs/tmp/h696 refs/tmp/h698` reports `CONFLICT (content) in src/soup_cli/trainer/mlx_sft.py`. The resolution is **take both** — the two changes are adjacent, not competing.

So: add the credit line here, I merge, then on #698 rebase and add its credit line in the same push. And there is now a real check on that resolution rather than a formality — running `test_issue684_mlx_grad_accumulation.py` and `test_issue685_mlx_grad_checkpoint.py` **together** catches either wholesale mis-resolution (taking only one side fails 5 tests one way, 1 the other; correct resolution passes 21).

## Two things for the issues, not for you

`#683` and `#686` are still broken after both of these land — I said so publicly and re-confirmed it on the merged tree. Two additions worth having on those threads:

- **`data.train_on_responses_only` defaults to `True`**, so #683 is not an opt-in bug: every MLX SFT run trains on prompt tokens against the documented default.
- Upstream reads it as `getattr(config, "mask_prompt", False)` on the args object while `TrainingArgs` has no such field, so the fix is an attribute set **after** construction, not a constructor kwarg. The obvious first attempt will `TypeError`.

Neither is a claim on your time; they are just facts whoever takes those issues should not have to rediscover.

AmirF194:
Pushed 5c3004e: the fragment now reads "#684 by @AmirF194 in #696". Noted the #683/#686 facts for whoever picks those up.

MakazhanAlpamys:
@AmirF194 — the credit line is correct, thank you. **But CI has gone red on two cells, and it is not your fragment — it is a real interaction with #703, which I merged an hour ago. The diagnosis and the one-line fix are below; this is not you having done something wrong.**

## What happened

```
FAILED tests/test_issue23_mlx_harness_bridge.py::test_harness_drives_rich_and_persists_metrics_without_losing_throughput[...]
  assert (10, 8) == (10, 10)
```

183 of the 198 cases in that file. Reproduced locally by merging your branch onto current `main`:

```
183 failed, 15 passed
```

The mechanism, traced rather than guessed:

- `benchmarks/harness/mlx_sft_smoke.py` writes its own YAML, and it pins `epochs`, `lr`, `batch_size: 1` — but **not** `gradient_accumulation_steps`, so it inherits the schema default of **4**.
- Its 10 rows at batch 1 give `iters = 10`.
- Your rounding then does `max(4, 10 - (10 % 4)) = 8`.
- #703's harness test asserts `(display.current_step, display.total_steps) == (10, 10)`.

Your arithmetic is right. The test encoded a number that was only true before your change.

## Why the fix is not "update the assertion to 8"

This is the part that matters, and it is a genuine finding your PR surfaced rather than caused.

`mlx_sft_smoke.py` is the harness behind a **published benchmark record**, `benchmarks/run-m1-8gb-mlx-sft.md`. If it silently drops from 10 iterations to 8, then the next Apple Silicon run measures a different thing from the published one, and the two stop being comparable — with nothing in the file saying why. #703's own changelog fragment states "The existing Apple Silicon measurements are unchanged", and changing the assertion to 8 would quietly make that false.

The real defect is underneath both PRs: **a benchmark harness's iteration count was silently depending on the schema default for an unrelated field.** That is fragile regardless of your change; you are just the first thing to disturb it.

## The fix — one line, in `benchmarks/harness/mlx_sft_smoke.py`

```diff
 training:
   epochs: {epochs}
   lr: 1e-4
   batch_size: 1
+  gradient_accumulation_steps: 1
```

Verified by me on your branch merged into current `main`:

```
204 passed in 37.37s      (test_issue23_mlx_harness_bridge.py + test_issue684_mlx_grad_accumulation.py)
```

`accum: 1` skips your rounding block entirely, `iters` stays 10, `(10, 10)` stays true, and the harness now measures a fixed number of steps that cannot drift when someone changes a default. Please add a short comment on that line saying why it is pinned — the next person to read it should not have to rediscover this.

## Two notes

**This is not a reason to change your rounding.** Rounding down to a whole accumulation window is correct and it is what #684 is about; a run that accumulates a partial group and never applies it is the bug. The harness simply should never have been at the mercy of that default.

**#698 still needs the rebase after this lands** — take-both on the `mlx_sft.py` metadata hunk, and running the two test files together now genuinely checks that resolution. Fold its credit line into the same push.

I would rather you made this change than I did, since it belongs in your PR — but if you would prefer I push it, say so and I will.

Srinivasan8888:
**The fragility is mine — I wrote that harness in #663 — and I can settle the question your fix note raises, because it decides whether #703's "measurements unchanged" claim survives.**

**It does. The published record is not affected.** Every published run divides evenly by the accumulation default, so the rounding is a no-op on all of them:

```
published ladder runs (#663)   48 rows -> iters=48  -> rounded=48   UNCHANGED
warm-cache re-run (#663)       16 rows -> iters=16  -> rounded=16   UNCHANGED
#703 harness test              10 rows -> iters=10  -> rounded= 8   CHANGES
```

So `benchmarks/run-m1-8gb-mlx-sft.md` still measures what it says, #703's fragment stays true, and only the 10-row test case moves. @AmirF194's arithmetic is right and nothing published needs a correction.

**But it survives by coincidence, not by design, and that is the part worth naming.** `48 % 4 == 0` is luck. Had I written 50 rows, or had the schema default been 5, every published figure would have silently been measured over a different number of steps than the record claims — with nothing in the file saying so. Your framing is exactly right: *a benchmark harness's iteration count was silently depending on the schema default for an unrelated field.* That is my defect, it predates both of these PRs, and #696 is just the first thing to disturb it.

**+1 to `gradient_accumulation_steps: 1` with a comment**, and I would go one further: the same argument applies to `batch_size`, which the harness pins at 1 already but for no stated reason. A benchmark harness should pin **everything that changes the step count**, and say why, so a future schema-default change cannot move a published number underneath its own record. If you want, I will do that pass as a follow-up rather than growing @AmirF194's PR — it is my file and my omission.

One correction to my own #663 while I am here: the record's "Reproducing" section gives the command line but does not state the resolved `gradient_accumulation_steps` the runs used. It should, for the same reason the model id and row count are stated. I will add it.

MakazhanAlpamys:
@Srinivasan8888 — that settles it, and I checked rather than took it. **Your conclusion is right and the published record is safe.** One line of your evidence is wrong, which changes nothing about the conclusion but you should have the correct number.

## Confirmed

`benchmarks/run-m1-8gb-mlx-sft.md:55` — *"48 synthetic chatml rows, 1 epoch -> 48 iterations"*, and the reproduce line at `:271` is `... Qwen2.5-0.5B-Instruct-4bit 48 1`. `48 % 4 == 0`, so @AmirF194's rounding is a no-op on every published figure. The record still measures what it says and #703's fragment stays true.

## The correction

Your table lists the warm-cache re-run as **16 rows**. It was **48**. From your own text at `:206`:

> *"Re-running `Qwen2.5-0.5B` at the same **48 rows** from a warm cache on a quieter machine gave 373.1 tok/s..."*

I grepped the whole record and there is no 16-row run in it. The conclusion survives — 16 and 48 both divide by 4 — so this is a slip in the evidence rather than in the answer. I am pointing at it because this is the second time today I have had to check a number in one of your write-ups and found one that did not hold (the other being the 404 comment link on #713), and both times the underlying work was right. That pattern is worth interrupting now, while it is still costing you nothing: the substance of what you produce is consistently good enough that a reader who spot-checks a citation and finds it wrong will discount the substance, which would be the wrong conclusion to draw about your work.

## "It survives by coincidence, not by design"

That is the sentence I would have wanted and did not write. `48 % 4 == 0` is luck; 50 rows, or a schema default of 5, and every published figure would have been measured over a different step count than the record claims, silently. Naming your own file's defect while the pressure was on someone else's PR is the behaviour I want more of here.

## On the follow-up: yes, and I have given it a home

**Accepted as you scoped it** — pin everything that moves the step count, with a stated reason per line, and add the resolved `gradient_accumulation_steps` to the Reproducing section. Your reasoning is right: a harness backing a published number must not let a schema default move that number underneath its own record, and `batch_size: 1` being pinned for no stated reason is the same latent problem one line up.

Filed as **#716** and assigned to you, rather than left as an agreement in a PR thread — I have created three claim races in three days by advertising things and not tracking them, and "I will do that pass as a follow-up" in someone else's comment section is exactly the shape that gets lost.

**No urgency and no ordering from me.** You now have #677 assigned and #713 open with four blocking findings. Do them in whatever order you like, or hand any of them back.

@AmirF194 — nothing above changes what #696 needs: the one line pinning `gradient_accumulation_steps: 1` with a short comment, which I measured brings the two files to 204 passed. The wider pass is @Srinivasan8888's and does not belong in your PR.

Srinivasan8888:
**You are right, and I have verified it against my own record rather than just accepting the correction.** The warm-cache re-run was **48 rows, not 16**.

`run-m1-8gb-mlx-sft.md:205-207` says "at the same 48 rows", and the figure is `2,130 tokens / 5.7 s` — 2,130 is the 48-row token total from the table, not a 16-row one. There was a 16-row run in my session, but it was a smoke-test of the committed harness, not the throughput measurement. I pulled the wrong number from memory when writing that table, in a comment whose entire purpose was to be checkable.

Conclusion unaffected: 48 and 16 both divide by 4, so the rounding is a no-op either way and the published record stands.

**On the pattern, which is the part worth answering.** That is two in one day — an invented `issuecomment-5540000000` on #713, and now a row count I did not look up. Both times the substance held and both times you had to check. The through-line is that I am reliable when I run a command and paste the output, and unreliable the moment I write a number from memory into a prose table, which is precisely where it looks most authoritative.

Concretely, going forward: any number in a comment gets derived in the same session I post it, or it does not go in. Where I cannot re-derive, I will say "from memory" rather than present it in a table. And the fix for the tables specifically is to paste the command and its output rather than transcribing results into a formatted grid — the grid is what makes a remembered number look measured.

Not asking you to take that on trust either; it is checkable in the next thing I post.

The `gradient_accumulation_steps: 1` pin plus the harness hardening pass is still mine to do — I will open it as its own PR once #713 and #715 clear your queue, so @AmirF194's PR does not grow.

AmirF194:
Pushed 5675c05: `gradient_accumulation_steps: 1` on the harness's training block, with a comment pointing at #696.

204 passed running test_issue23_mlx_harness_bridge.py and test_issue684_mlx_grad_accumulation.py together, ruff clean on the changed file.

Issue #685, opened 2026-09-05
MLX SFT ignores gradient_checkpointing


Amix29:
## Description

When `training.gradient_checkpointing: true` is requested for the MLX backend, the wrapper does
not set mlx-lm's `TrainingArgs.grad_checkpoint`. The upstream default remains `false`, and the
adapter metadata also hardcodes `false`.

Affected revision: `ef2e02b51a1b34546823edce55715d28234805cb` (Soup 0.74.0), with mlx-lm
0.31.3 behavior.

Relevant code:

- [`TrainingArgs` construction omits checkpointing](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/trainer/mlx_sft.py#L295-L303)
- [metadata hardcodes `grad_checkpoint: false`](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/trainer/mlx_sft.py#L333-L358)
- [Soup checkpointing field](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/config/schema.py#L2811-L2822)

## Steps to reproduce

Configure `backend: mlx` and `training.gradient_checkpointing: true`, then inspect the
`TrainingArgs` delivered to mlx-lm.

```text
requested: true
forwarded grad_checkpoint: false
```

An explicit control constructed with `grad_checkpoint=True` preserves the value.

## Expected behavior

Soup should enable a supported MLX checkpointing mode or reject the requested setting before
training.

## Impact

Memory use can be higher than the configuration implies, potentially causing avoidable memory
pressure or OOMs. The software-contract defect is confirmed; no Metal peak-memory or throughput
number is claimed here.

## Proposed fix and acceptance criteria

- [ ] Map boolean checkpointing to the supported mlx-lm mechanism.
- [ ] Explicitly map or reject Soup's `selective`, `medium`, `full`, and `auto` values for MLX.
- [ ] Record the effective value in metadata.
- [ ] Run a small Apple Silicon backward-pass test with checkpointing enabled.
- [ ] Measure peak memory and throughput in a separate hardware validation.

## Environment

Contract probe run under Python 3.12.2 against the original Soup wrapper and upstream mlx-lm
0.31.3 arguments. No Metal training session was performed.

MakazhanAlpamys:
[...]

PR #698, opened 2026-09-05
fix(mlx): wire gradient_checkpointing into the SFT TrainingArgs (#685)

AmirF194:
## What does this PR do?

Wires `training.gradient_checkpointing` into the MLX SFT trainer's `TrainingArgs` and its `adapter_config.json` output, which currently ignore it entirely.

`MLXSFTTrainerWrapper.train()` builds mlx-lm's `TrainingArgs` without a `grad_checkpoint` kwarg, so mlx-lm's own dataclass default (`False`) always applies no matter what `training.gradient_checkpointing` is set to (`src/soup_cli/trainer/mlx_sft.py:295-303`). The same block also hardcodes `grad_checkpoint: false` into the written `adapter_config.json` (line 356), so a user cannot even tell from the output that checkpointing never ran.

`gradient_checkpointing` accepts a plain bool or one of the `selective`/`medium`/`full`/`auto` tiers (`config/schema.py:2811`), but mlx-lm's `TrainingArgs.grad_checkpoint` is a single bool with no tier concept, so this resolves the field with `bool(...)`, the same coercion `commands/train.py`'s hardware-fit predictor and `layer_stream.should_enable_hf_gradient_checkpointing` already use for this field on the non-MLX backends. Any non-`False` tier therefore enables checkpointing.

## Type of change

- [x] Bug fix

## Verification

- Added `tests/test_issue685_mlx_grad_checkpoint.py`, which runs the real `train()` body against a fake mlx/mlx-lm harness (same approach as `tests/test_issue634_mlx_resume.py`) and asserts the `grad_checkpoint` value written to `adapter_config.json` for every value the schema accepts. All 6 parametrized cases fail on `main` (each returns `False`) and pass on this branch.
- Ran the full MLX test suite (`test_issue634_mlx_resume.py`, `test_mlx_backend.py`, `test_mlx_dispatch.py`, `test_issue353_mlx_seed_warning.py`, `test_issue392_mlx_target_keys.py`) alongside the new tests: 83 passed.
- `ruff check src/soup_cli/ scripts/ tests/` passes.
- I have not run this on an actual Metal training session. The `mlx-smoke` CI job needs `macos-14`, which isn't available to me here, so I verified the fix at the `TrainingArgs`-construction level rather than end to end. This also doesn't give mlx-lm's single on/off switch any of the four tiers' distinct behavior; there isn't one to give it, since mlx-lm's own `grad_checkpoint()` wraps the same one decoder layer type regardless.

Refs #685: this covers the mechanical wiring, the tier-to-bool mapping, and the metadata fix, but not the issue's two hardware-validation checkboxes (a Metal backward-pass run and a peak-memory/throughput measurement), which need Apple Silicon this environment doesn't have.


MakazhanAlpamys:
@AmirF194 — the production change here is **correct**, and the tier mapping is exactly right. I am blocking on the test file, because it does not test the thing the PR is named after, and the state it leaves behind is worse than the bug being fixed.

## What is right, and I verified all of it

**The full mapping table.** Every accepted schema value, on clean `main` versus your branch:

| `gradient_checkpointing` | kwarg passed? | `TrainingArgs.grad_checkpoint` | `adapter_config.json` |
|---|---|---|---|
| `True` | yes | `True` | `true` |
| `False` | yes | `False` | `false` |
| `"selective"` | yes | `True` | `true` |
| `"medium"` | yes | `True` | `true` |
| `"full"` | yes | `True` | `true` |
| `"auto"` | yes | `True` | `true` |
| unset | yes | `False` | `false` |

**No accepted value maps wrongly.** On `main` every one of them produced `grad_checkpoint = False` and `false` in the metadata.

**Flattening is the only available mapping, and I checked upstream rather than assuming.** From the real `mlx_lm` 0.31.3 sdist: `grad_checkpoint: bool = field(...)`, and the trainer does `if args.grad_checkpoint: grad_checkpoint(model.layers[0])`. A single on/off switch, no granularity.

**Your consistency claim checks out**, which I verified because a claim like that is easy to make and easy to get wrong:
- `commands/train.py:91` — `gc = bool(getattr(tcfg, "gradient_checkpointing", False))`
- `utils/layer_stream.py:928` — `return bool(gradient_checkpointing) and not stream_layers`

Both plain `bool()`. The code comment is honest.

98 passed, zero skips, ruff clean, zero test lines removed.

## Blocking — the mandated mutation survives, and what it leaves is worse than main

Deleting `grad_checkpoint=grad_checkpoint` from the `TrainingArgs(...)` call while keeping the metadata fix:

```
### MUTATION A: delete grad_checkpoint= from TrainingArgs, keep metadata fix
7 passed
  grad_checkpoint= in TrainingArgs call: False
  metadata still writes it: True
```

**Seven passed.** The whole file is green with the fix removed.

The cause is at `tests/test_issue685_mlx_grad_checkpoint.py:79`: `_install_fake_mlx` does not capture the `train(**kwargs)` call, so nothing in the file ever touches `TrainingArgs`. The test at line 145 is named `test_grad_checkpoint_reaches_training_args` and asserts only on `json.loads(...adapter_config.json)` at line 159. Its name claims one thing and its body checks another — which is, almost word for word, the defect class this repo has now hit five times (#363, #423, #659, #672, #675).

The state that survives is not merely "the bug is back":

```
requested    | kwarg? | TrainingArgs | adapter_config.json
'selective'  | False  | False        | true
'full'       | False  | False        | true
```

mlx-lm gets `False` and **the metadata claims `true`**. Main at least reported `false` honestly. A test that permits that is worse than no test, because it makes the record lie with a green suite behind it.

Confirming the diagnosis: mutations (b) hardcode-to-`False` and (c) revert-only-the-metadata kill the **identical** five tests. A suite that cannot distinguish "TrainingArgs wrong" from "metadata wrong" is pinning one half and calling it both.

### The fix, and you already wrote it in the sibling PR

`test_issue684_mlx_grad_accumulation.py` — your own #696 — has the right shape, as does `test_issue634_mlx_resume.py`. Return the captured calls from the helper:

```python
def _fake_train(**kwargs):
    captured_calls.append(kwargs)
    ...
return captured_calls
```

then assert **both** halves:

```python
captured_calls = _install_fake_mlx(monkeypatch)
...
assert captured_calls[0]["args"].grad_checkpoint is expected   # the mlx-lm contract
assert written["grad_checkpoint"] is expected                  # the recorded metadata
```

I checked that the JSON assertion alone genuinely cannot see mutation (a) — the `args` assertion is required, not merely tidier. Worth keeping both and renaming the existing one to `..._is_recorded_in_adapter_config`, so (b) and (c) get distinct kill-sets instead of an identical one.

One related gap this exposes: `_FakeTrainingArgs.__init__` does `self.__dict__.update(kwargs)`, so it accepts **any** kwarg name where the real dataclass would raise `TypeError`. In #696 a misspelling is caught incidentally; here, with no `args` assertion at all, a misspelled kwarg passes silently.

## Should-fix — the changelog, and it is not pedantry

`CHANGELOG.md` must not be edited directly; #487 moved this project to per-PR fragments. Add `changelog.d/0.74.0/698.fixed.md`, delete the `CHANGELOG.md` hunk, and use the **pull-request** number in the filename and the text.

The concrete cost: **you and #696 insert at the same offset in `CHANGELOG.md`, so whichever merges second takes a conflict.** Preventing exactly that is why fragments exist.

## Nit — the tier flattening is silent, and this file already has the convention for saying so

A user asking for `selective` silently gets `full`. #685's criterion 2 says "explicitly map **or reject**", and mapping is the right call — but `mlx_sft.py:133-155` `_check_unsupported()` already prints `MLX backend ignores: ...` for `quantization=8bit`, GaLore, Ring Attention, FlashAttention, `seed` and `data_seed`, and `docs/training.md:122-125` documents that mechanism. One line there would match the file's own habit:

```python
if isinstance(tcfg.gradient_checkpointing, str):
    unsupported.append(
        f"gradient_checkpointing tier {tcfg.gradient_checkpointing!r} "
        "(MLX has a single on/off switch; enabling it)"
    )
```

## Cross-PR

This and #696 conflict in three hunks of `mlx_sft.py`. The third is the dangerous one: each PR changes one line of the metadata dict and leaves the other's hardcoded, so taking either side wholesale **silently reverts the other's metadata fix**. Resolve it take-both deliberately, and afterwards run `test_issue684_mlx_grad_accumulation.py` and `test_issue685_mlx_grad_checkpoint.py` **together** — once this PR's `args` assertions exist, both files passing simultaneously is the real proof the resolution was done right. #665 is orthogonal and clean against both.

I would like #696 to land first, since its `iters` rounding is the larger semantic change.

## Worth knowing: two siblings are still silently broken after both your PRs

Confirmed by execution on both branches. **#683** — `train_on_responses_only` never reaches `create_dataset`, and `adapter_config.json` still records `mask_prompt: false`; loss is still computed over prompt tokens. **#686** — with `weight_decay: 0.3, warmup_ratio: 0.2, scheduler: cosine`, the optimizer receives `{'learning_rate': 0.0002}` and nothing else. Neither appears in `_check_unsupported()` either, so both are silently ignored rather than warned about.

Your two PRs fix two of the four known "MLX validates it then ignores it" fields. If you want the other two, say so on #683 and #686 and they are yours — you now have the harness and the pattern for exactly this shape.

AmirF194:
Good catch, and confirmed the exact mutation you described: deleting grad_checkpoint=grad_checkpoint from TrainingArgs passed the old suite because the metadata write reads the same local variable regardless. Split the test in two now, one asserting on the TrainingArgs the fake train() receives and one on adapter_config.json, so each half has a distinct kill set. Also added the tier-flattening warning through the existing MLX backend ignores mechanism, and a small test for it. Pushed d858e94.

Moved the CHANGELOG.md edit to changelog.d/0.74.0/698.fixed.md as well. Agreed #696 lands first; will rebase this one onto it once that merges. Left #683 and #686 alone for now, they're separate design-shaped fixes rather than a kwarg pass-through, but happy to pick them up as their own PRs if you'd rather have them split out that way.

MakazhanAlpamys:
@AmirF194 — **the blocker is resolved and I confirmed it with the exact mutation I mandated. Blocked only on sequencing plus one line.**

## The mutation that survived last round now kills 7 tests

Re-run by me on `d858e94`, worktree with `PYTHONPATH` pinned inside it:

```
baseline                                          15 passed
delete grad_checkpoint= from TrainingArgs          7 failed  -> killed   (old suite: 7 PASSED)
revert ONLY the metadata write                     5 failed  -> killed
hardcode grad_checkpoint = True                    3 failed  -> killed
misspell the kwarg (grad_ckpt=)                    7 failed  -> killed
```

**7 and 5, on disjoint sets.** That is the property I was actually after. The reason the old suite let the deletion through is that the metadata write and the `TrainingArgs` kwarg read the same variable, so one assertion covered both and neither was pinned. Now each fails on its own.

The misspelling mutation is the one I did not ask for and am glad you covered: `_FakeTrainingArgs` does `__dict__.update(kwargs)` and would happily accept `grad_ckpt=`, so without a test that reads the *correct* attribute name, a typo would ship silently. That closes the "the fake accepts anything" gap.

Checked against the real upstream rather than against the fake, since the stubs cannot prove it: the `mlx-lm` 0.31.3 sdist declares `grad_checkpoint: bool = False` on `TrainingArgs`, so the spelling will not `TypeError` on real Metal.

The tier-flattening warning is routed through the existing `_check_unsupported()` mechanism with a bool control, which is the right place for it — Soup's `gradient_checkpointing` accepts `selective` / `medium` / `full` / `auto` and mlx-lm has one bool, so flattening silently would have been its own small lie.

Zero test lines removed. Ruff clean. CI green.

## What is left

**1. Rebase, after #696 lands.** I verified the conflict rather than predicting it: `git merge-tree refs/tmp/h696 refs/tmp/h698` reports `CONFLICT (content) in src/soup_cli/trainer/mlx_sft.py`. The resolution is **take both** — the two changes are adjacent, not competing.

There is now a real check on that resolution instead of a formality. Run the two test files **together** after rebasing:

```
take only #696's side  -> 5 failed
take only #698's side  -> 1 failed
correct take-both      -> 21 passed   (full MLX suite: 146 passed)
```

Before this round neither file could detect a bad resolution; now both can.

**2. The credit line.** `changelog.d/0.74.0/698.fixed.md` is missing `by @AmirF194`. Filename, directory, category and the in-text `#698` are correct; only the author field is absent. The convention already in that directory:

```
(#685 by @AmirF194 in #698)
```

Fold it into the rebase push so this costs you one push rather than two.

Merging #696 as soon as its credit line lands, then this.

MakazhanAlpamys:
@AmirF194 — **#696 is merged** (`3365f89`), so this is unblocked. Two things and it goes in:

1. **Rebase.** I verified the conflict rather than predicting it: `git merge-tree` reports `CONFLICT (content) in src/soup_cli/trainer/mlx_sft.py`. The resolution is **take both** — the two changes are adjacent, not competing.

   There is now a real check on that resolution instead of a formality. After rebasing, run the two files together:

   ```
   take only #696's side  -> 5 failed
   take only #698's side  -> 1 failed
   correct take-both      -> 21 passed
   ```

2. **The credit line.** `changelog.d/0.74.0/698.fixed.md` is missing `by @AmirF194`; the convention in that directory is `(#685 by @AmirF194 in #698)`. Fold it into the rebase push so this costs one push rather than two.

Nothing else outstanding. Everything I asked for last round is done and verified: deleting `grad_checkpoint=` from `TrainingArgs` now fails **7** tests where the old suite reported 7 passed, the metadata write fails a **disjoint** 5, and the misspelled-kwarg mutation you added — which I had not asked for — closes the gap where `_FakeTrainingArgs` would have accepted `grad_ckpt=` silently.

One thing that landed in `main` while you were waiting and that your rebase will pick up: the MLX benchmark harness now pins `gradient_accumulation_steps: 1`, from your own #696. It should not affect this PR, but if `tests/test_issue23_mlx_harness_bridge.py` behaves oddly after the rebase, that is where to look first.

AmirF194:
Rebased and pushed c522ea1: kept both sides in mlx_sft.py (grad_checkpoint and grad_accumulation_steps side by side in the TrainingArgs call, the summary dict, and the args-computation block), and fixed the changelog line to "#685 by @AmirF194 in #698".

Ran both suites myself against the merged tree rather than trusting the numbers: test_issue684 (6) and test_issue685 (15) both fully pass, 21/21. test_issue23_mlx_harness_bridge.py is untouched by the rebase, 198/198 pass. ruff clean on the changed file.

MakazhanAlpamys:
Merged as `9b58300`. Thank you — and the mutation you added on your own is the one worth naming.

**The blocker is closed, verified by me on a tree merged with current `main`:**

```
baseline (both suites + the #703 harness)          219 passed
delete grad_checkpoint=grad_checkpoint               7 failed   (round 1: 7 PASSED)
delete grad_accumulation_steps=...  (#696's half)    2 failed
```

Seven and two, on disjoint sets. The reason the original suite let a deletion through is that the `TrainingArgs` kwarg and the adapter-metadata write read the same variable, so one assertion covered both and neither was pinned. Each now fails on its own.

**The misspelling mutation is yours and I had not asked for it.** `_FakeTrainingArgs` does `__dict__.update(kwargs)` and would have accepted `grad_ckpt=` without complaint, so a rename could have shipped silently. Without a test that reads the *correct* attribute name, the fake is a hole rather than a double. That closes it.

You also checked the field against the real `mlx-lm` 0.31.3 sdist rather than against your own stub — which is the only way a fake-based test can be trusted about an upstream API, and the thing most people skip.

**The rebase resolution is verified, not assumed.** `3365f89` is an ancestor and both fields survive at `mlx_sft.py:331-332`. Running the two suites together is now a real check on that resolution rather than a formality: taking only one side fails 5 tests one way and 1 the other, and the correct take-both gives 219 passed.

The tier flattening going through `_check_unsupported()` is the right home for it — Soup accepts `selective` / `medium` / `full` / `auto` and mlx-lm has one bool, so collapsing that silently would have been its own small lie.

Zero test lines removed. Credited in `CONTRIBUTORS.md` (`0f481f3`).

That is four of yours merged today — #696, #698, #701, #704, #711, five in fact. Two things you may want to know: `#683` and `#686`, which you left notes on, both now have PRs open from @Srinivasan8888 (#733 and #734); and #735 is yours and in the queue.

AmirF194:
Glad the take-both resolution and the mutation held up under your own verification. Good call catching that the fake's dict update would've swallowed a renamed kwarg silently.

PR #701, opened 2026-09-05
fix(data): carve val_split rows out before interleave oversampling

### LINKED ISSUE #680: Validation split can overlap training data after interleave oversampling
## Description

With non-streaming multi-dataset training, the `over` strategy and some `probs`
configurations duplicate rows before Soup performs the automatic `val_split`. Exact copies
of the same example can therefore appear in both training and validation.

Relevant code:

- [`_cycle_to` and `_combine_interleaved`](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/data/loader.py#L447-L514)
- [combination before `_finalize`](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/data/loader.py#L517-L572)
- [automatic validation slice](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/data/loader.py#L235-L259)

## Steps to reproduce

Use two distinct synthetic sources:

- A: `A0 ... A7`
- B: `B0, B1`
- `data.interleave: over`
- `data.val_split: 0.1`

Execute the real `_combine_interleaved()` result through `_finalize()`.

```text
combined rows: 16
training rows: 14
validation rows: B0, B1
validation rows also present in training: B0, B1
```

The same probe with `probs=(0.5, 0.5)` also places validation row `B0` in training.
Controls without repetition, and a control that reserves validation before oversampling,
have no overlap.

The problem can be substantially worse than this small example. With source A containing
10,000 rows, source B containing 1,000 rows, `over`, and `val_split: 0.1`, all 2,000
validation rows can also occur in the 18,000-row training partition.

The maintainer independently confirmed the overlap through the public `load_dataset(cfg.data)`
path on revision `2edc063`.

## Scope

This affects both local multi-dataset loading and the Hub interleave fallback. In
`_load_interleaved_hub_datasets()`, when only some Hub entries provide their own `validation`
split, the remaining combined data still goes through `_finalize()` after repetition and can
leak into the automatically created validation set.

There is no supported `data.val` field for supplying an independent local validation dataset;
unknown configuration keys are ignored. A Hub dataset's existing `validation` split is the only
current path that avoids this particular fallback for that dataset.

## Expected behavior

Automatic validation data must contain no logical example already present in training because
of Soup's own oversampling.

## Impact

Validation loss can be arbitrarily optimistic and can mislead checkpoint or hyperparameter
selection. On local interleaved data there is no configuration workaround other than avoiding
`over`/affected `probs` combinations or disabling the automatic validation split.

Reserve validation from each source's unique rows before any cycling or probabilistic repetition,
then apply repetition only to the training partitions. Preserve a Hub-provided validation split
without modification.

- [ ] `set(train_ids).isdisjoint(validation_ids)` holds for `over` and affected `probs` cases;
      row counts alone are not sufficient.
- [ ] Both source orders are covered.
- [ ] Local interleave and the partial-Hub-validation fallback are both covered.
- [ ] Existing Hub `validation` splits remain unchanged and excluded from training.
- [ ] `concat` and `under` behavior remains unchanged.
- [ ] A regression test fails if repetition is moved back before the split.

## Implementation status

PR [#701](https://github.com/MakazhanAlpamys/Soup/pull/701) is open and targets both the local
and Hub-fallback paths with set-intersection regression tests.

## Environment

Initially reproduced with the original pure-Python functions from the pinned checkout under
Python 3.12.2. The maintainer also confirmed it through the real public loader path. No model or
training backend is involved.

AmirF194:
Fixes #680.

`_combine_interleaved` pads a source up to the target length by cycling its own rows (`_cycle_to`), and `_finalize` then took a positional tail slice of that already-padded list for `val_split`. A row and its cycled copy are separate list entries, so the slice could put one on each side of the split. Same defect in `_load_interleaved_hub_datasets`' fallback (only some HF-hub entries have their own `validation` split), which combines through the same `_combine_interleaved` before `_finalize`.

Fix: for `over`/`probs`, `val_split` now runs per source before cycling (`_split_val`, the same slicing formula `_finalize` already used, just moved earlier), and only the remaining train rows get padded. `concat`/`under` never duplicate rows, so they still go through the unchanged path from #443.

Added the zero-intersection test you asked for, both strategies, plus one with the padded source listed first, using `probs [0.7, 0.3]` so the no-overlap guarantee is checked independent of which source gets padded (a false pass there would have been the case that mattered). Same test shape for the hub fallback path. Full suite green on 3.10/3.11/3.12; `ruff` clean.

Left `docs/data.md` with one line on the new per-source ordering, since the strategy table there didn't mention `val_split` at all before this.


MakazhanAlpamys:
Picked up, review running. Two things while it does.

**You took the hub-path scope note as well**, which I flagged on #680 but did not require: `_load_interleaved_hub_datasets` falls back through the same `_combine_interleaved` + `_finalize` whenever only *some* entries carry their own `validation` split. Fixing one path and leaving the other would have looked complete and left the leak reachable. Good call.

**And you added the zero-intersection assertion rather than a count assertion**, which is the specific thing I asked for on the issue — a positional-slice fix that still duplicates passes a count check. The ordering variant, with the padded source listed first, is the case I would have missed: `over` pads whichever source is shorter, so a fix that handles one ordering and not the other passes half the time.

I will not take those on trust; the review is mutating the core change and checking the hub path separately, and it is specifically checking one thing your description does not address — **whether the val fraction a user asks for is still the fraction they get**. Moving the split before cycling changes both which rows are held out and how many, and a fix that silently turns `val_split: 0.1` into a different number would be a quieter bug than the leak it replaces.

## On your queue

You now have three open PRs and **#696 and #698 are both blocked**, so they are the higher priority when you next have time:

- **#696** is the one to fix first. The wiring is right, but as written it makes zero-optimizer-update runs reachable on the schema default — `gradient_accumulation_steps` defaults to 4, `iters` counts micro-batches, and mlx-lm only steps when `it % accum == 0`, so a 3-iteration dataset now does **0** updates where it did 3. That is the same silent-no-training shape as #430 and #433.
- **#698** needs its test to assert on `TrainingArgs`, not only on `adapter_config.json`. Deleting the kwarg currently leaves all seven tests green while the metadata claims `true` and mlx-lm gets `False` — worse than the bug being fixed.

No rush on either, and no criticism in the ordering — three PRs in a day is a lot of ground. I mention it only so the blocked ones do not sit behind the new one in your own queue.

MakazhanAlpamys:
@AmirF194 — **the leak is closed, on both paths, and every mutation I threw at the core fix was killed.** Four items before merge, one of which is a config that worked on `main` and now crashes.

## The leak, measured on both trees

`intersection` = val rows that also appear in train, through the real `load_dataset(cfg.data)`:

| config | main `019c324` | branch `5f3d40c7` |
|---|---|---|
| A=10000 B=1000, `over`, 0.1 | **2000/2000 (100%)** | **0/1100** |
| A=10000 B=1000, `probs [.5,.5]`, 0.1 | **1100/1100 (100%)** | **0/1100** |
| A=10 B=5, `over` / `probs` | 2/2 | 0/2 |
| CONTROL `concat` / `under` | 0 | 0, counts byte-identical |

**Four mutations, four killed**, pristine restored and re-verified green between each:

| mutation | result |
|---|---|
| both per-source gates -> `if False:` (the original bug) | **KILLED** x4 |
| break **only** the hub gate | **KILLED** — the hub test alone |
| break **only** the local gate | **KILLED** — the three local tests, hub stays green |
| widen the gate to include `concat`/`under` | **KILLED** x2 |

Rows two and three are what I most wanted: the hub half is **independently** pinned, so your claim to have fixed both paths is not resting on one shared test. And copying your four new tests onto `main` fails all four — none is vacuous against the bug it targets.

51 passed on the branch against 47 on main (exactly your four), 307 vs 303 on the wider loader surface, no pre-existing failures anywhere in that selection, ruff clean, rebases clean, fragment correct and credited.

## Should-fix 1 — a source with exactly one row now crashes, and the message blames the wrong thing

This is the one I want fixed here rather than followed up. Measured on both trees:

```
MAIN    sizes=[10,1] over val_split=0.1 -> train=18 val=2   OK
MAIN    sizes=[10,1] over val_split=0.5 -> train=10 val=10  OK
BRANCH  sizes=[10,1] over val_split=0.1 -> ValueError
BRANCH  sizes=[10,1] over val_split=0.5 -> ValueError
```

```
ValueError: data.interleave: every dataset in data.train must have >= 1 row after formatting
```

**That message is false.** The dataset did have a row after formatting. `_split_val` moved it to val, left `[]`, and `_combine_interleaved`'s pre-existing zero-row guard fired on the empty remainder. So a user with a one-row source gets told their data is malformed when the real cause is that `val_split` consumed it.

A one-row source is a small case but a plausible one — a held-out probe file, a single-example smoke config. Either keep one row on the train side, or raise with the real reason at both call sites (`loader.py:576` and `:1097`):

```python
train_rows, val_rows = _split_val(rows, data_config.val_split)
if not train_rows:
    raise ValueError(
        f"data.interleave: data.val_split={data_config.val_split} leaves 0 training "
        f"rows for a source with {len(rows)} row(s) under strategy '{spec.strategy}', "
        "which carves val out per source. Use 'concat'/'under', lower val_split, "
        "or drop the source."
    )
```

Worth a test at `[10, 1]` either way.

## Should-fix 2 — the val fraction silently changes, and your description does not mention it

This is the question I said the review would ask that your body does not address, and it has a real answer. `over`, A=1000 B=100:

| requested `val_split` | main val / total | branch val / total | effective |
|---|---|---|---|
| 0.05 | 100 / 2000 = 5.00% | 55 / 1955 | **2.81%** |
| 0.10 | 200 / 2000 = 10.00% | 110 / 1910 | **5.76%** |
| 0.20 | 400 / 2000 = 20.00% | 220 / 1820 | **12.09%** |
| 0.50 | 1000 / 2000 = 50.00% | 550 / 1550 | **35.48%** |

At the realistic ratio, **`val_split: 0.1` now delivers 5.76%.** `probs` is unaffected, because its total scales with the shrunken train sides — so the two strategies your PR touches now behave differently on the same field.

**I am not asking you to change this.** It is the defensible semantic: you cannot hold out 10% of an inflated 20000 when only 11000 distinct rows exist without either duplicating val rows or removing more distinct rows than asked. And it is strictly better than main, whose 2000-row val contained 1000 distinct rows every one of which was also in train.

But it is a silent change to a number the user typed, so it needs to be stated with a figure in `docs/data.md` and the fragment, not left for someone to discover. Related and worth one clause: val is drawn in each source's **original** proportion while train after `over` is rebalanced, so a user who oversamples specifically to rebalance now gets train and val from different mixtures. Better than main, where val was 100% one source — but different from what they may assume.

## Should-fix 3 — three of your four new tests are vacuous in one direction

Making `_split_val` return everything as train and an empty val:

```
2 failed, 49 passed
FAILED test_val_split_applied_once_after_mixing            <- pre-existing
FAILED test_hub_validation_split_ignored_when_only_some…   <- pre-existing
```

**None of your four noticed.** `not (train_texts & val_texts)` is trivially true when `val_texts` is empty, so a fix that simply never splits passes all of them. Only pre-existing count assertions caught it.

One line each:

```python
assert val_texts, "val must be non-empty, or the disjointness assertion is vacuous"
```

plus an exact expected count where it is derivable. This is the same shape as the finding on #679 earlier today — a correct implementation whose distinguishing guard was unpinned — and it is worth internalising: a set-disjointness assertion always needs a non-emptiness assertion beside it.

## Should-fix 4 — `streaming: true` + `over` still leaks, and the fragment implies otherwise

Verified on your branch: **10/20 val rows leak, identical to main.** `interleave_datasets(stopping_strategy="all_exhausted")` recycles the shorter stream, the result is materialised, and `_finalize` at `loader.py:742` takes the same positional slice.

Pre-existing and **not yours to fix here** — the streams are not countable ahead of time, so it needs a different mechanism. But your fragment says the fix lands *"on both the local-file and HF-hub interleave paths"*, and the docs table three lines above your new paragraph says the strategies *"mean the same thing on the streaming path as on the local path"*. A `streaming: true` user will read that as covered. One clause in the fragment and one sentence in the docs; I will file the follow-up issue.

Also stale now: `docs/data.md:644` still says `val_split` *"applies to the combined train rows"* on a partial-hub-val mixture, which is exactly what you stopped doing.

## Two nits

Your PR body says the ordering test uses `over`. It uses `probs [0.7, 0.3]`. That matters in your favour: I measured the `over` version and **it does not leak on main with the padded source first**, so an `over`-shaped ordering test would have been vacuous. You picked the right shape and then described the wrong one.

And four em-dashes became commas in two docstrings, unrelated to the fix, one of them producing a comma splice. Revert the punctuation, keep the `#680:` sentence.

Fix 1 and 3, add the disclosure for 2 and 4, and I will re-run the mutation set and merge.

AmirF194:
Fix 1: `_split_val_per_source` now raises with the real cause when val_split leaves a source with 0 train rows, at both call sites. Added a test pinning the message.

Fix 3: added the non-emptiness assertion to the three vacuous tests plus one more test at [10, 1]. Confirmed all four catch a "never split" mutation.

Disclosure for 2 and 4: docs/data.md and the fragment now state the effective val fraction under over/probs (measured your 1000/100 example myself, matches your table) and say plainly the streaming path is unaffected and still leaks.

Also reverted the two docstring edits, using a colon instead of restoring the em dash (our own house style disallows it).

MakazhanAlpamys:
@AmirF194 — **the leak is closed and stays closed, and three of my four items are resolved. Blocking on one line, and it is on the headline test.**

## The fix works, measured through the real `load_dataset(cfg.data)`

`LEAK` = distinct val rows that also appear in train:

| config | main | this branch |
|---|---|---|
| `[10000, 1000]` `over` 0.1 | **1000 / 1000 (100%)** | **0 / 1100** |
| `[10000, 1000]` `probs [.5,.5]` 0.1 | **1000 / 1000 (100%)** | **0 / 1100** |
| `[10, 5]` `over` 0.1 | 2 / 2 | 0 / 2 |
| control: `concat` / `under` | 0 | 0, counts byte-identical |

**Six mutations, all killed** — including the two that matter after a refactor:

| mutation | result |
|---|---|
| both gates -> `if False` (the original bug) | killed — all 5 new tests |
| remove the item-1 `raise` | killed — the `[10,1]` test alone |
| break **only** the local gate | killed — 4 local tests, hub stays green |
| break **only** the hub gate | killed — the hub test alone |
| `_split_val` never splits | killed at suite level — 6 fail |
| widen the gate to all strategies | killed — a pre-existing count test |

Local and hub are independently pinned. Zero test lines removed, ruff clean, CI green.

---

## BLOCKING — one test is still vacuous, and it is the `over` case

You added `assert val_texts` to three of the four. The fourth — `test_interleave_over_val_split_has_no_duplicate_row_across_train_and_val`, the headline test for #680 — did not get it. Proved by running it alone under the never-split mutation:

```
mutation: _split_val never splits, only this test selected

  1 passed
  len(train)=16   len(val)=0
  'not (train & val)' subject on the val side is EMPTY: True
  the exact-count assert (train + val == 16) still holds: True
```

`assert len(result["train"]) + len(result["val"]) == 16` is a **total**, and a total is invariant under never-splitting — so it is not the exact-expected-count substitute it looks like. This test passes on a build where validation does not happen at all.

Your reply says all four catch a never-split mutation, and that is true of the four you enumerated: the three you guarded plus the new `[10,1]` test. The `over` test was one of the *original* four and got missed in the counting. One line:

```python
assert val_texts, "val must be non-empty, or the disjointness assertion is vacuous"
```

I would not normally hold a PR for a single assertion. I am holding for this one because it is the test named after the issue: if it can pass on a build where the leak is back, it is not evidence.

---

## Worth folding into the same push — three small things, none needing a re-run of the mutation set

**1. The mirror image of my item 1: a genuinely empty source now blames `val_split`.**

```
sizes=[10,0]  over    val_split=0.1  -> ValueError: data.val_split=0.1 leaves 0 training rows
                                        for a source with 0 row(s) ... lower val_split, or drop the source.
sizes=[10,0]  over    val_split=0.0  -> ValueError: every dataset in data.train must have >= 1 row after formatting
sizes=[10,0]  concat  val_split=0.1  -> ValueError: every dataset in data.train must have >= 1 row after formatting
```

The source had **zero** rows; `val_split` consumed nothing, and "lower val_split" cannot help. `if rows and not train_rows:` lets a truly empty source fall through to the correct downstream error. Your new message is right for the case it was written for — it just catches one case too many.

**2. The runtime warning still describes the old behaviour.** You caught the stale sentence at `docs/data.md:644`; the identical sentence in the code that *prints* it was missed. Captured live from the fixed hub path:

> `...ignoring every hub validation split and applying data.val_split to the combined train rows instead.`

For `over` / `probs` it is no longer applied to the combined train rows — that is precisely what this PR changed. More users see this than see the docs line.

**3. The missing clause from my item 2.** Your val-fraction figure is exactly right — I re-measured `1000/100, over, 0.1` and got `train=1800 val=110 total=1910 = 5.76%` against your documented "110 val out of 1910 (5.8%)". What is still unsaid is that **train and val now come from different mixtures**: on that same config train is `A=900 B=900` (50.0% A) and val is `A=100 B=10` (90.9% A). Someone who oversamples specifically to rebalance now gets a validation set in the original skew, and nothing tells them. That is a consequence of the correct fix, not a bug — but it belongs next to the fraction note.

## Resolved, and my public statement on #702 was right

`streaming: true` + `over` + `val_split 0.2` still leaks **10 of 30** distinct val rows on this branch. #701 does not reach it, exactly as I said on the issue. Your fragment says so, the docs paragraph says so, and #702 is open and labelled.

## Nit, and I owe you a correction on it

The em-dash rationale in your reply does not hold for this repo: `src/soup_cli/` contains **4195** em dashes, 40 of them in `loader.py` itself, including in the docstring block you edited. The colon does fix the comma splice, so the change is fine — only the stated reason is wrong. And the PR body still describes the ordering test as `over` when it uses `probs [0.7, 0.3]`; right shape, wrong description.

One line for the blocker, three small ones alongside it, and I will merge.

AmirF194:
Pushed ec1f44d, all four folded in:

- Added the missing `assert val_texts` to the `over` test, verified it fails on the never-split mutation and passes on the real fix.
- `_split_val_per_source` now checks `if rows and not train_rows`, so a genuinely empty source falls through to the generic error instead of the misleading val_split one.
- The hub-fallback warning now says "carving per source" for `over`/`probs` and "combined train rows" for `concat`/`under`, matching what actually runs.
- Added the train/val mixture-skew note to docs/data.md, right after the fraction example.

Also fixed the PR body: the ordering test uses `probs [0.7, 0.3]`, not `over`.

AmirF194:
Catching the vacuous `over` test was the sharpest find in this one, it would've passed on a build where the leak was back. Appreciate you running the mutation set twice to be sure. Good working through this with you.

MakazhanAlpamys:
Merged as `dfa8615`. Thank you — three review rounds, and the third one is the one that mattered.

**The leak is closed on both paths**, measured through the real `load_dataset(cfg.data)`:

| config | main | merged |
|---|---|---|
| `[10000, 1000]` `over` 0.1 | **1000 / 1000 (100%)** | **0 / 1100** |
| `[10000, 1000]` `probs [.5,.5]` 0.1 | **1000 / 1000 (100%)** | **0 / 1100** |
| control `concat` / `under` | 0 | 0, counts byte-identical |

**Six mutations, six killed**, and the local and hub halves stay independently pinned after the refactor — breaking only the local gate fails 4 tests while the hub test stays green, and breaking only the hub gate fails just that one.

**The vacuous test, re-checked by me on your new head.** Before, running it alone under a never-split mutation gave `1 passed` with `len(val) == 0`. Now:

```
MUT never-split, THE OVER TEST ALONE -> 1 failed
```

That was worth the round. `not (train & val)` is trivially true against an empty val set, and the companion assertion was a *total*, which is invariant under never-splitting — so the test named after the issue could pass on a build where the leak was fully restored.

You also folded in all three of the smaller items in the same push: the empty-source guard is `if rows and not train_rows:` so a genuinely empty source still falls through to the correct downstream error rather than being blamed on `val_split`; the runtime warning branches per strategy instead of describing the old behaviour; and the train-vs-val mixture consequence is now in `docs/data.md` — 50.0% source A in train against 90.9% in val — so someone oversampling to rebalance finds out from the docs rather than from a confusing eval curve.

Zero test lines removed. Merged sequentially with four other PRs onto current `main` in a scratch worktree first: no conflicts, 350 passed.

Credited in `CONTRIBUTORS.md` (`9f1e48b`). #702, the streaming half this provably cannot reach, is open and unclaimed.

PR #703, opened 2026-09-05
fix(mlx): exercise display and tracker in benchmark harness

Shutaru:
## What does this PR do?

Refs #23. This is the harness follow-up to #663 and #665, not a closure of the whole issue.

Thanks @Srinivasan8888 for the handoff and the stdout-tee pointer. The benchmark now passes a real `TrainingDisplay`, an `ExperimentTracker`, and its `run_id` into the merged MLX wrapper.

The tracker database lives beside the temporary benchmark artifacts, so the user's normal experiment history is untouched. A run fails if either display updates or tracker metrics are missing. Failed and interrupted training is recorded as failed, and the tracker connection is closed.

Training stays inside the stdout tee. The tee now forwards `isatty()` too: without that, Rich treats an interactive terminal as redirected output and the live panel does not refresh. The new tests cover automatic terminal detection as well as forced-terminal and redirected output, while checking that the cumulative token count still produces the same whole-run throughput.

The bridge keeps its normal process-local SSE behavior. No changes to the trainer, validation callbacks, CI workflows, or README badge are included.

## Review follow-up

The parser now strips ANSI controls before collapsing whitespace, so a word-wrapped final counter does not fall back to an earlier report. The real Rich/Live/FileProxy integration test covers every width from 20 through 200: **21/181 failed before this fix, 0/181 after**, using the committed two-report fixture. Parser controls also cover absent and zero counters, ANSI-only and whitespace-only changes, and the transition from 999 to 1000 tokens.

The summary is inside the tracker context and uses `Console.print(..., soft_wrap=True, markup=False, highlight=False)` to keep the database path copyable. Its test checks the complete line at 20 columns and verifies that the connection is still open when it is printed.

## Verification

Windows, Python 3.12:

- **331 passed** across the MLX, display, tracker, dispatch, and runtime-isolation tests, including **198 harness cases**.
- **211 passed** when re-running the harness and display-bridge tests with `FORCE_COLOR=1 TERM=xterm-256color` (overlaps the tests above).
- **17 passed** in `tests/test_issue487_changelog_fragments.py`.
- Changelog fragment validation and `git diff --check` pass.
- A local in-memory mutation check rejected **18/18 targeted regressions**: omitted display/tracker/run ID, bypassed tee, default database, removed or weakened presence guard, missing failure status/connection cleanup, hidden terminal detection, removed/reordered ANSI and whitespace normalisation, use of an earlier count, wrapped summary output, and summary access after tracker close.

The new tests execute the real Soup wrapper, Rich display, SQLite tracker, and SSE buffer, with MLX/model loading and host measurements replaced by fixtures.

**No Apple Silicon run was performed here.** The existing M1 measurements are unchanged; the benchmark notes explain that new runs include display/tracker overhead. The full repository suite was not run locally and remains for CI. A fresh hardware run is still needed to measure the enabled bridge.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [ ] `pytest tests/ -v` passes (full suite pending CI; focused results above)
- [x] Updated the relevant benchmark reproduction notes

### DIFF
--- a/benchmarks/harness/mlx_sft_smoke.py
+++ b/benchmarks/harness/mlx_sft_smoke.py
@@ -14,6 +14,11 @@ machine is directly comparable to `benchmarks/run-m1-8gb-mlx-sft.md`.
 
 Requires Apple Silicon and ``pip install -e ".[mlx]"``.
 
+Attaches Soup's Rich display and a local experiment tracker, and refuses a run
+with no bridge metrics. The database stays beside the temporary artifacts;
+the bridge also emits its normal process-local SSE events. Timing includes
+display/tracker overhead, unlike the original published M1 measurements.
+
 Usage:
     python mlx_sft_smoke.py [model-id] [rows] [epochs]
 
@@ -89,6 +94,10 @@ class _Tee:
         for stream in self._streams:
             stream.flush()
 
+    def isatty(self):
+        # Rich must still recognise a terminal through the capture wrapper.
+        return getattr(self._streams[0], "isatty", lambda: False)()
+
 
 _TRAINED_TOKENS_RE = re.compile(r"Trained Tokens (\d+)")
 
@@ -128,8 +137,11 @@ def main() -> int:
     epochs = int(sys.argv[3]) if len(sys.argv) > 3 else 1
 
     import mlx.core as mx
+    from rich.console import Console
 
     from soup_cli.config.loader import load_config_from_string
+    from soup_cli.experiment.tracker import ExperimentTracker
+    from soup_cli.monitoring.display import TrainingDisplay
     from soup_cli.trainer.mlx_routing import resolve_trainer
 
     tmp = Path(tempfile.mkdtemp(prefix="mlx_smoke_"))
@@ -179,14 +191,37 @@ output: {out}
           f"mlx peak after load: {mx.get_peak_memory() / 1024**3:.3f} GB   "
           f"(includes first-time Hub download)")
 
-    mx.reset_peak_memory()
-    t0 = time.time()
     # mlx-lm prints its progress to stdout; tee it so the run stays readable
     # AND the trained-token counter is recoverable for the throughput line.
     buffer = io.StringIO()
-    with contextlib.redirect_stdout(_Tee(sys.stdout, buffer)):
-        result = trainer.train()
-    train_s = time.time() - t0
+    display = TrainingDisplay(cfg, device_name="Apple Silicon (MLX)")
+    # A benchmark must not populate the user's normal experiment database.
+    with contextlib.closing(ExperimentTracker(db_path=tmp / "experiments.db")) as tracker:
+        run_id = tracker.start_run(
+            cfg.model_dump(), device="mlx", device_name="Apple Silicon (MLX)", gpu_info={},
+        )
+        try:
+            mx.reset_peak_memory()
+            t0 = time.time()
+            with contextlib.redirect_stdout(_Tee(sys.stdout, buffer)):
+                result = trainer.train(display=display, tracker=tracker, run_id=run_id)
+            train_s = time.time() - t0
+            metrics = tracker.get_metrics(run_id)
+            if not metrics or display.current_step <= 0:
+                raise RuntimeError("MLX bridge received no display/tracker metrics")
+            tracker.finish_run(
+                run_id, initial_loss=result["initial_loss"], final_loss=result["final_loss"],
+                total_steps=result["total_steps"], duration_secs=result["duration_secs"],
+                output_dir=result["output_dir"],
+            )
+        except BaseException:
+            # Include Ctrl-C; the wrapper stops its display in its own finally.
+            tracker.fail_run(run_id)
+            raise
+    Console().print(
+        f"bridge        : {len(metrics)} metric reports saved to {tracker.db_path}",
+        markup=False,
+    )
     train_stdout = buffer.getvalue()
     print(f"train         : {train_s:.1f}s   "
           f"mlx peak during train: {mx.get_peak_memory() / 1024**3:.3f} GB")
--- a/benchmarks/harness/mlx_sft_smoke.py
+++ b/benchmarks/harness/mlx_sft_smoke.py
@@ -99,6 +99,7 @@ class _Tee:
         return getattr(self._streams[0], "isatty", lambda: False)()
 
 
+_ANSI_RE = re.compile(r"\x1b\[[0-9;?]*[a-zA-Z]")
 _TRAINED_TOKENS_RE = re.compile(r"Trained Tokens (\d+)")
 
 
@@ -111,10 +112,14 @@ def _final_trained_tokens(train_stdout: str) -> int | None:
     wrapper's result dict does not carry the count, so stdout is the only
     place it surfaces.
 
-    Returns None if the line is absent — the caller prints nothing rather
-    than inventing a figure.
+    Rich's Live/FileProxy wraps progress lines and inserts ANSI controls.
+    Normalise both before matching, or an intact earlier report can silently
+    win over the wrapped final counter at some terminal widths.
+
+    Returns None if the counter is absent; the caller reports unavailable.
     """
-    matches = _TRAINED_TOKENS_RE.findall(train_stdout)
+    flat = re.sub(r"\s+", " ", _ANSI_RE.sub("", train_stdout))
+    matches = _TRAINED_TOKENS_RE.findall(flat)
     return int(matches[-1]) if matches else None
 
 
@@ -218,10 +223,10 @@ output: {out}
             # Include Ctrl-C; the wrapper stops its display in its own finally.
             tracker.fail_run(run_id)
             raise
-    Console().print(
-        f"bridge        : {len(metrics)} metric reports saved to {tracker.db_path}",
-        markup=False,
-    )
+        Console().print(
+            f"bridge        : {len(metrics)} metric reports saved to {tracker.db_path}",
+            markup=False, highlight=False, soft_wrap=True,
+        )
     train_stdout = buffer.getvalue()
     print(f"train         : {train_s:.1f}s   "
           f"mlx peak during train: {mx.get_peak_memory() / 1024**3:.3f} GB")
--- a/benchmarks/run-m1-8gb-mlx-sft.md
+++ b/benchmarks/run-m1-8gb-mlx-sft.md
@@ -255,6 +255,17 @@ rather than reconstructed.
 
 ## Reproducing
 
+The current harness also exercises the Rich display and experiment-tracker
+bridge added in #665. It saves `experiments.db` beside the temporary artifacts,
+leaving the user's usual experiment database untouched, and fails if no display
+updates or tracker metrics arrive. The bridge emits its normal process-local
+SSE events too. No separate UI server is started.
+
+The measurements above predate this wiring; they have not been re-run with the
+bridge enabled. New runs include display/tracker overhead in the training wall
+time. Throughput still uses mlx-lm's cumulative trained-token count captured
+inside the stdout tee, not the display's iterations-per-second value.
+
 ```bash
 pip install -e ".[mlx]"
 python benchmarks/harness/mlx_sft_smoke.py mlx-community/Qwen2.5-0.5B-Instruct-4bit 48 1
--- a/changelog.d/0.74.0/23.fixed.md
+++ b/changelog.d/0.74.0/23.fixed.md
@@ -1,0 +1,9 @@
+- **The MLX SFT benchmark harness now exercises the display/tracker bridge
+  (Refs #23).** It passes a real Rich display and an artifact-local SQLite
+  tracker to the MLX wrapper, and rejects runs without display updates or
+  recorded metrics. Training stays inside the stdout tee so whole-run
+  throughput still comes from mlx-lm's trained-token counter. Regression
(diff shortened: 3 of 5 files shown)


MakazhanAlpamys:
@Shutaru — you took the follow-up I left open on #23 and did it properly. **Every claim in your description verified true, no mutation survived, and the `isatty` finding is understated rather than overstated.** One should-fix, which is an ironic one: attaching the display makes the harness's headline throughput number width-dependent, and at some widths silently wrong.

## Mutations — eight, no survivors

I did not take your 11/11 on trust:

| mutation | result |
|---|---|
| drop `display=` from `trainer.train(...)` | **KILLED** (5 params) |
| drop `tracker=` / `run_id=` | **KILLED** (5) |
| remove `_Tee.isatty` | **KILLED** (2) |
| **neuter the presence guard entirely** | **KILLED** (3) |
| weaken the guard to metrics-only | **KILLED** — the display half catches it |
| weaken the guard to display-only | **KILLED** — the tracker half catches it |
| **tracker DB falls back to the default `~/.soup`** | **KILLED** (5) |
| `fail_run` removed from the `except` | **KILLED** — `'running' != 'failed'` |
| `contextlib.closing` dropped | **KILLED** — `tracker._conn is None` |
| `trainer.train` moved outside the tee | **KILLED** (5) |

Rows five and six are the ones I care about: **both halves of the presence guard are independently pinned.** A single blunt assertion would have passed both, and a harness that tolerates half a dead bridge is most of the bug you are fixing.

122 passed against main's 110 — exactly your 12 — **zero skips**, because you reuse `_install_fake_mlx` and drive the real body rather than gating on hardware CI does not have. Green under `FORCE_COLOR=1 TERM=xterm-256color`, which is the ANSI trap that has reddened this repo's CI four times. Zero test lines removed, ruff clean, rebases clean against both #663 and #665.

## The three claims I checked hardest

**Tracker isolation — true, and stronger than you claimed.** With `Path.home`, `HOME` and `USERPROFILE` all redirected, the DB lands in the benchmark's temp dir and **`~/.soup` is never even created**. The real `~/.soup/experiments.db` was SHA-256 identical before and after the entire review. A benchmark writing into a user's real experiment history would have been a genuine defect; this is the right answer.

**"Fails if the bridge is dead" — true in all three modes**, including the discriminating one: with `display.update` dead but the tracker alive, the run still fails **while having written 2 metric rows**. That proves it is a real AND-guard rather than one check wearing two hats.

**`isatty` — true, and the consequence is bigger than you wrote.** You said the live panel does not refresh. It is worse: `Live.refresh()` is a complete no-op when `is_terminal` is False, so nothing is emitted at all.

```
main _Tee (no isatty): bytes emitted per refresh tick -> [0, 0, 0, 0, 0]
PR   _Tee (isatty)   : bytes emitted per refresh tick -> [239, 260, 260, 260, 260]
```

And there is a subtlety you got right that is worth naming, because it is the kind of thing that breaks later if someone "tidies" it: `Console.is_interactive` is computed **once in `__init__`**, while `is_terminal` is a live property. This works because `display.py`'s module-level console is built at import — before your redirect — freezing `is_interactive=True`, and `is_terminal` is then re-read through the tee. Reordering those two would silently kill it.

No false positive either: the tee over a non-tty still reports `is_terminal=False`.

## Should-fix — the display makes the published number width-dependent

`benchmarks/harness/mlx_sft_smoke.py:102`.

With a display attached, `Live.start()` calls `_enable_redirect_io()`, which replaces `sys.stdout` with a Rich `FileProxy` whenever `console.is_terminal` — which your `isatty` fix has just made true. Every subsequent mlx-lm `print()` is therefore re-rendered and **word-wrapped at the console width**. When the wrap lands inside `Trained Tokens <n>`, `_TRAINED_TOKENS_RE = re.compile(r"Trained Tokens (\d+)")` stops matching.

Measured here across widths 20-200 on #663's own two report lines:

```
parse FAILS (no match) : 20 widths  -> [20,21,22,23,26,34,35,36,37,51,93,94,95,96, ...]
parse WRONG value      :  5 widths  -> [(27,'222'), (38,'222'), (52,'222'), (103,'222'), (104,'222')]
```

The second row is the problem. `matches[-1]` falls back to an **earlier, unwrapped** report, so the harness prints `222 / 19.7s` instead of `454 / 19.7s` — a plausible-looking number that is wrong by 2x. That is precisely the failure class #663 spent a round correcting, and this file's own docstring says it exists to prevent it. It also shifts *within* a run, because the token count gains digits as training proceeds.

Your 12 tests parametrize widths 160/80/40/32. All four land in the passing set.

The fix, verified — 20 fails and 5 wrong both go to **0/181**, all 12 existing tests still green:

```python
_ANSI_RE = re.compile(r"\x1b\[[0-9;?]*[a-zA-Z]")
_TRAINED_TOKENS_RE = re.compile(r"Trained\s+Tokens\s+(\d+)")
...
    flat = re.sub(r"\s+", " ", _ANSI_RE.sub("", train_stdout))
    matches = _TRAINED_TOKENS_RE.findall(flat)
```

Both halves are needed — the whitespace collapse alone does not clear it. And please add `52` and one of the 90s to the `terminal_width` parametrize; either alone would have caught this.

## Two nits

`mlx_sft_smoke.py:221` prints the DB path through `Console().print(...)`, which word-wraps it mid-path and makes it non-copy-pasteable. A plain `print()` does not wrap.

`metrics` and `tracker` are read after the `contextlib.closing` block. Correct today, since `db_path` is a plain attribute, but fragile; moving the summary inside costs nothing.

## Not your problem, but noted

`.github/workflows/ci.yml:18` lints only `src/soup_cli/ scripts/ tests/` — `benchmarks/harness/` is **not** linted in CI. That arrived with #663, not with you, but the file now carries real parsing logic. I will fix it.

And the honest limit on all of the above: no Apple Silicon here, so nothing measured real MLX. What I could verify without hardware — the Rich/`Live`/`FileProxy`/`isatty` interaction, the SQLite isolation, the failure and interrupt cleanup, and the throughput arithmetic — is all pure Python and was exercised for real. The thread interleaving on hardware, where Live refreshes at 2 Hz on a separate thread while mlx-lm prints, is the one thing that interacts with the should-fix and that I cannot reach.

Fix the parse and I will re-run the mutation set and merge.

Shutaru:
Thanks, @MakazhanAlpamys. My width cases missed the stale-counter failure. I will reproduce 52 and the 90s before changing the parser, then sweep 20–200 and address both summary nits. I will leave the workflow lint change with you.

Shutaru:
@MakazhanAlpamys, pushed 7ef38eb and rebased onto 1082db4, including your benchmark lint change.

The parser now strips ANSI controls before collapsing whitespace. With the committed two-report fixture, I reproduced failures at 21 of the 181 widths from 20–200 before the fix; all 181 now recover the final count of 454 and the correct throughput. This includes 52 and 93. Separate controls cover ANSI-only and whitespace-only changes, absent/zero counters, and 999→1000.

Both nits are addressed. The summary is inside the tracker context, and I kept `Console.print` with `soft_wrap=True, markup=False, highlight=False` to follow the repository's output convention. A 20-column test checks the complete, unbroken path and confirms that the connection is still open at print time.

Verification: 331 focused tests plus 17 changelog tests passed. The harness/display subset also passed with `FORCE_COLOR=1 TERM=xterm-256color` (211 cases). All 18 targeted in-memory mutants were rejected, including removing either normalisation step, reversing their order, taking the earlier count, and undoing either summary fix. Ruff and fragment validation are clean.

Thanks for catching the stale-counter case before merge. CI is pending on this head; the Apple Silicon limitation is unchanged.

MakazhanAlpamys:
Merged as `9d6ed04`. Thank you — and the second round is why this is a good PR rather than an adequate one.

## What I re-ran on `7ef38eb`

Worktree, `PYTHONPATH` pinned inside it and checked via `soup_cli.__file__`, py3.10, `--no-cov`.

```
baseline                                 198 passed
revert BOTH normalisations               25 failed
remove ONLY the ANSI strip                3 failed
remove ONLY the whitespace collapse      24 failed
```

**The disjoint kill sets are the thing I was checking for.** Stripping ANSI and collapsing whitespace are two separate claims, and a single blunt assertion would have passed both mutations while pinning neither. Three and twenty-four, on different tests, means each step is defended on its own. That is a property you have to design a test suite for; it does not happen by accident.

Nine mutations across the whole PR, no survivors. Zero test lines removed. Ruff clean including `benchmarks/`, which only started being linted yesterday.

## On the width sweep

You said your original cases missed the stale-counter failure, then reproduced 52 and 93 before touching the parser and swept 20-200. That sweep is the right shape for a reason worth naming: `range(20, 201)` covers the discriminating widths **by construction**, where a hand-picked list covers the ones the author happened to think of. 21 of those 181 widths genuinely discriminate the fix, and both widths from review are inside that set.

You also solved it differently from what I suggested — collapsing `\s+` rather than widening the regex — which is equivalent, and you got the ordering right: strip ANSI first, then collapse. The other order silently keeps escape fragments inside the collapsed run.

## Two things beyond the ask

**You fixed both review findings here instead of filing them.** `_Tee.isatty` and the parser were outside the PR's original scope, and the `isatty` test is discriminating rather than decorative — at width 80 with no explicit terminal size, detection can only arrive through the tee, so deleting the method actually fails something.

**345 passed under `FORCE_COLOR=1 TERM=xterm-256color`.** Rich's per-character ANSI in CLI-output assertions has turned this repo's CI red four times, most recently on all nine cells at once, and Windows never sees it because Rich auto-disables colour there. Checking under forced colour is the only local run that means anything for the Linux and macOS cells.

**Zero mocks.** The display, tracker, console, tee and the harness module itself are real, loaded from the real path and driven through the real `main()`; what is faked is MLX hardware and mlx-lm's `train`, which writes real mlx-lm-shaped stdout. That matters here more than usual — `tests/test_embedding.py` currently holds 79 green tests over a path that crashes in production, because a `MagicMock` invented the attributes production lacks.

And you disclosed, unprompted, that the Apple Silicon measurements are unchanged and a hardware run with the bridge enabled is still outstanding. That sentence is in the changelog fragment where a reader will find it.

Credited in `CONTRIBUTORS.md` (`601fbc3`).

Srinivasan8888:
@Shutaru — **both of those were mine, and the first one is worse than it looks.** Thank you for fixing rather than filing.

**The `isatty` gap.** My `_Tee` forwarded `write` and `flush` and nothing else, so `contextlib.redirect_stdout(_Tee(...))` made Rich see a non-terminal and decline to render. The irony is that I added that tee in #663 to make the throughput line reproducible, and in doing so **broke the live panel my own #665 exists to drive** — the harness could not have demonstrated the bridge even once. Your `isatty` forwarding is the fix, and `getattr(..., lambda: False)` on the first stream is the right shape, since a buffer legitimately has no `isatty`.

**The wrapped-counter parse.** My `Trained Tokens (\d+)` regex assumed one clean line per report. Through Rich's Live/FileProxy it is neither clean nor necessarily one line, so `findall(...)[-1]` could return a stale value. Strip ANSI first, collapse whitespace second — and pinning the two steps **independently** (3 tests vs 24) is the part I would have got lazy about and asserted as one.

**One thing I checked before saying anything, because it is the question that matters:** the published table in `run-m1-8gb-mlx-sft.md` is **not** contaminated. Those `trained_tokens` came from grepping the raw ladder output, which ran on 09-05 before the tee existed and was plain mlx-lm stdout — never routed through `_Tee` or Rich. So `2,130 / 19.7 s = 108.1` still holds and the record needs no third correction. The defect was in the *reproduction path* a future reader would use, not in the numbers already published — which is arguably worse, since it would have silently produced different figures for whoever re-ran it.

A `range(20, 201)` width sweep where 21 of 181 widths genuinely discriminate is a better test than I would have written. Noted for next time.

MakazhanAlpamys:
@Shutaru — a heads-up on something your merged PR interacts with, discovered an hour after it landed. Nothing here is a defect in #703, and nothing needs doing by you.

#696 (@AmirF194, still open) makes MLX round `iters` down to a whole number of accumulation groups, which is the correct fix for #684 — a run that accumulates a partial group and never applies it is the bug. On the merged tree it turns your harness's 10 iterations into 8, because `mlx_sft_smoke.py` pins `epochs`, `lr` and `batch_size: 1` but not `gradient_accumulation_steps`, so it inherits the schema default of 4. Your `assert (display.current_step, display.total_steps) == (10, 10)` then fails in 183 of its 198 parametrised cases.

**The interesting part is why the fix is not to change that assertion to 8.** Your harness backs a published record, and your fragment says "The existing Apple Silicon measurements are unchanged". If the harness silently drops to 8 iterations, the next Apple Silicon run stops being comparable to the published one with nothing in the file explaining why — and your sentence quietly becomes false. So the fix is to pin `gradient_accumulation_steps: 1` in the harness YAML, which I have asked @AmirF194 to fold into #696 and verified brings both test files to 204 passed.

The latent defect this exposes is worth your knowing because it is in your file: **the harness's iteration count depended on the schema default for an unrelated field.** It would have drifted the moment anyone touched that default, and your test is what caught it — the assertion is doing exactly the job a benchmark harness test should do. That is a point in its favour, not against it.

No action needed from you. Flagging it so you hear it from me rather than seeing your test named in someone else's red CI.

Shutaru:
Thanks for the careful review and the credit, @MakazhanAlpamys, and for the handoff, @Srinivasan8888. Happy to help, and glad the tests caught the accumulation change too.

PR #704, opened 2026-09-05
fix(distill): refuse wasserstein/topk_align on tokenizers that disagree

### LINKED ISSUE #681: ULD wasserstein/topk_align feed student token IDs to a different teacher tokenizer
## Description

The non-aligned ULD strategies `wasserstein` and `topk_align` forward student `input_ids`
directly to the teacher. When the teacher vocabulary is smaller, Soup clamps the IDs into
range, but clamping does not translate tokens between vocabularies.

Relevant code: [student IDs are copied and clamped before the teacher forward](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/trainer/distill.py#L606-L642).

Issue #258 added the separate `wasserstein_aligned` strategy and is complete. This report is
about the remaining non-aligned strategies: they still accept incompatible tokenizers instead
of restricting their contract or refusing the pair.

## Steps to reproduce

A real PyTorch CPU forward/backward was run with tiny local student and teacher models:

```text
student IDs for "hello cat dog":       [1, 6, 7]
IDs received by the teacher:           [1, 4, 4]
correct teacher IDs for the same text: [1, 2, 3]
training loss: finite (~1.86718)
backward: completes
```

The finite loss is part of the problem: the teacher can be conditioned on different or
corrupted text without a visible failure.

## Expected behavior

Each model should receive the same text encoded with its own tokenizer, or Soup should reject
a tokenizer pair that the selected strategy cannot handle safely.

## Impact

The student may distil a distribution produced from the wrong teacher context. Results from
cross-tokenizer ULD can therefore be semantically invalid even though training completes.

- [ ] Remove clamping as a semantic compatibility mechanism.
- [ ] Restrict `wasserstein` / `topk_align` to demonstrably ID-compatible tokenizers, or
      route incompatible pairs through a validated re-tokenization/alignment path.
- [ ] Fail early with a clear message when compatibility cannot be established.
- [ ] Add a tiny-model test with different token-ID maps that fails on the current path.
- [ ] Preserve the supported same-tokenizer fast path.

## Environment

Reproduced under Python 3.12.2 and PyTorch 2.12.1 CPU using the original `compute_loss` and ULD
kernel from the pinned checkout. No CUDA training session was required for the mechanism.

AmirF194:
Went with the refusal you outlined in the issue rather than wiring in a translation, since `wasserstein_aligned` already does that correctly and the two loss kernels (full sorted distribution vs top-k) don't need per-token alignment, only a text-correct teacher input.

`_require_uld_id_compatible_tokenizers` loads the teacher tokenizer for `wasserstein`/`topk_align` at `setup()` and reuses `same_tokenizer` (#304's speculative-decoding probe: vocab size plus id agreement over a small probe corpus) rather than a new check. Raises before training starts, naming `wasserstein_aligned`. Removed the clamp since compatibility is now proven up front; it was the only thing referencing `teacher_vocab` in that closure, so that variable went too.

Updated the `docs/training.md` example: it recommended `--uld-strategy wasserstein` for Llama -> Mistral, which is exactly the pair this now rejects. Pointed it at `wasserstein_aligned` instead and moved the Llama -> Mistral line down to sit with it.

Tests reproduce your own repro shape (same vocab size, disjoint ids) against the pure check and against `DistillTrainerWrapper.setup()` directly, mocking only the model/tokenizer loading so the setup path itself is exercised, not just the helper. Confirmed both fail on `main` (the mismatched pair falls through to an unrelated `KeyError` well past where the fix now raises) and pass on this branch, for both `wasserstein` and `topk_align`, on Python 3.10/3.11/3.12. Added a same-tokenizer case so the fast path stays clear too.

I ran this against fake tokenizers with disjoint id maps, same as #304's own suite does for `same_tokenizer`, not against real Llama/Mistral checkpoints, so it's the same gap you flagged on your own comment.

Fixes #681

### DIFF
--- a/changelog.d/0.74.0/704.fixed.md
+++ b/changelog.d/0.74.0/704.fixed.md
@@ -1,0 +1,9 @@
+- **`wasserstein`/`topk_align` cross-tokenizer distillation trained on logits the teacher
+  computed for the wrong text, with no visible failure (#704 by @AmirF194, closes #681).**
+  Both strategies forwarded the student's own `input_ids` straight to the teacher, clamped
+  into its vocab range to avoid an index error. Clamping does not translate tokens between
+  vocabularies, so a mismatched pair silently fed the teacher garbled or wrong text while
+  still reporting a finite, plausible-looking loss. `wasserstein`/`topk_align` now refuse a
+  tokenizer pair that isn't interchangeable at `setup()`, naming `wasserstein_aligned` (which
+  already re-tokenizes correctly) as the working alternative for genuinely different
+  tokenizers. The same-tokenizer fast path (e.g. two sizes in one family) is unaffected.
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -310,7 +310,7 @@ soup data brain-rot <data.jsonl> [--strict]   Brain-rot detector — arXiv 2510.
 soup iterative-dpo --base-model <m> --reward-model <rm> --prompts <p.jsonl> --output-dir <out> --rounds N --pairs-per-round N [--plan-only]  Iterative DPO loop driver — LIVE sample→score→pair→train (v0.70.0; live v0.71.11)
 soup train --reward-hack-detector info_rm|rm_ensemble [--reward-hack-halt]  Reward-hacking detector for GRPO — LIVE callback (v0.70.0; live v0.71.11)
 soup train --reward-hack-mitigation off|log_only|kl_control|pid_lagrangian  Closed-loop reward-hacking auto-mitigation (detect → raise KL/β → rollback → early-stop); GRPO/PPO, requires --reward-hack-detector; PPO BETA (v0.71.26)
-soup train --uld-strategy wasserstein|topk_align [--uld-top-k N]  Cross-tokenizer ULD on task='distill' — LIVE W1/topk loss (v0.70.0; live v0.71.11)
+soup train --uld-strategy wasserstein_aligned  Cross-tokenizer ULD on task='distill' (different tokenizers) — LIVE (v0.71.18)
 soup train --minillm-enabled [--minillm-teacher-mix-ratio 0.3]  MiniLLM reverse-KL distillation — LIVE (v0.70.0; live v0.71.11)
 soup train --rl-checkpoint-save-every-steps N [--rl-checkpoint-keep-last N]  Mid-epoch checkpoint for GRPO/PPO — LIVE (v0.70.0; live v0.71.11)
 soup train --echo-trap-enabled [--echo-trap-threshold 0.6 --echo-trap-halt]  RAGEN echo-trap detector for GRPO — LIVE callback (v0.70.0; live v0.71.11)
--- a/docs/training.md
+++ b/docs/training.md
@@ -269,13 +269,18 @@ soup train --config soup.yaml \
 soup train --config grpo.yaml --reward-hack-mitigation kl_control   # raise KL, recover
 soup train --config grpo.yaml --reward-hack-mitigation log_only     # observe only, no action
 
-# Cross-tokenizer distillation — Llama -> Mistral, no shared vocab needed
+# Wasserstein-distance distillation between two models that SHARE a
+# tokenizer, e.g. two sizes in the same family (#681: wasserstein / topk_align
+# forward the student's token ids to the teacher unchanged, so the pair must
+# tokenize identically; for a genuinely different tokenizer see
+# wasserstein_aligned below).
 # (Universal Logit Distillation, Boizard et al. 2024 arXiv:2402.12030)
 soup train --config soup.yaml --uld-strategy wasserstein
 
-# Cross-tokenizer distillation for FULLY-DISJOINT tokenizers (v0.71.18) —
-# aligns student/teacher token sequences over decoded character spans, so you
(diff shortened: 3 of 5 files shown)


MakazhanAlpamys:
@AmirF194 — **approving on the substance. One stale doc line and it goes in.**

## Verified by execution

Worktree at `043efdb`, `PYTHONPATH` pinned inside it (checked via `soup_cli.__file__`), py3.10, `--no-cov`, baseline on clean `main` first.

- New suite + four pre-existing distill/ULD suites on the branch: **255 passed**.
- The new suite copied onto unfixed `main`: **8 of 9 fail**. The `setup()` test dies at `distill.py:434` with `KeyError: 'train'` — which is exactly the "falls through to an unrelated `KeyError` well past where the fix now raises" you claimed in the description. Claim verified rather than taken on trust.

| mutation | result |
|---|---|
| gate always accepts (`if False and ...`) | 5 failed — killed |
| drop `topk_align` from the gated set | 2 failed — killed |
| invert the probe so it over-refuses | 8 failed — killed |
| **unwire the call from `setup()`, helper left intact** | 2 failed — killed |

The last one is the one I care about. This repo's most persistent defect is a correct helper that the command never calls (#363, #423, #659, #672, #675), and the only way to know is to delete the call site and see whether anything notices. Something does, because your `setup()` test drives the real `DistillTrainerWrapper.setup()` with only model and tokenizer loading stubbed.

**The clamp removal is safe and I checked rather than assumed.** `same_tokenizer` requires equal `vocab_size` *and* identical ids over `PROBE_CORPUS`, so the index error the clamp existed to prevent cannot occur once the check passes. `_teacher_vocab` had no other reader — the clamp was its only consumer — so deleting both is complete rather than partial.

Zero test lines removed. Ruff clean. CI green on all nine cells.

## Blocking — one line

`docs/commands.md:313` still reads:

```
soup train --uld-strategy wasserstein|topk_align [--uld-top-k N]  Cross-tokenizer ULD on task='distill' — LIVE W1/topk loss (v0.70.0; live v0.71.11)
```

After this PR those two strategies are same-tokenizer-only, so that line now sends a reader directly into the error you added. You rewrote `docs/training.md` carefully — including moving the "Llama -> Mistral" example onto `wasserstein_aligned`, which was the right call and the thing I would have asked for — and this line is the one that got missed. Please point it at `wasserstein_aligned` for the cross-tokenizer case the same way.

## Rulings on two things so you do not have to guess

**`.fixed.md` is the right category, not `.changed.md`.** This narrows a shipped, documented capability, and a user config that appeared to work will now refuse at `setup()`. Normally that argues for `changed`. But the path was never doing what it advertised — I confirmed on #681 that the clamp does not translate tokens and does nothing at all when the teacher vocab is larger — so a run that "worked" was training on the teacher's logits for text the teacher never saw, at a finite and plausible-looking loss. Removing a silently-wrong result is a fix. Your fragment already says all of this in full, which is why I am comfortable.

**No shipped config breaks.** I confirmed no bundled recipe, template, or `examples/configs/` entry sets `uld_strategy`, so the blast radius is limited to hand-written user configs, which will get a clear refusal naming `wasserstein_aligned`.

## Non-blocking

`test_setup_does_not_raise_id_compat_error_on_identical_tokenizer` wraps its call in a bare `except Exception: pass`. It would therefore also pass if `setup()` failed early for a completely unrelated reason. Defensible for a negative test and the parametrised positive tests carry the real weight — but if you are pushing the docs line anyway, narrowing that to the specific exception type costs nothing.

Push the docs line and I will merge.

AmirF194:
Pushed c6ad3d3: docs/commands.md:313 now points at wasserstein_aligned for the cross-tokenizer case, matching the training.md wording.

MakazhanAlpamys:
Merged as `a1eab0f`. Thank you — and the mutation I care about most is the one that would have caught this repo's favourite failure.

**Four mutations, four killed:**

| mutation | result |
|---|---|
| gate always accepts | 5 failed |
| drop `topk_align` from the gated set | 2 failed |
| invert the probe so it over-refuses | 8 failed |
| **unwire the call from `setup()`, helper intact** | 2 failed |

The last one is the standing test against a correct helper the command never calls — #363, #423, #659, #672, #675, five instances of it found by other people. Your `setup()` test drives the real `DistillTrainerWrapper.setup()` with only model and tokenizer loading stubbed, so deleting the call site actually fails something.

Your description's claim checked out too: the new suite on unfixed `main` fails 8 of 9, and the `setup()` case dies at `distill.py:434` with `KeyError: 'train'` — exactly the "falls through to an unrelated error well past where the fix now raises" you described.

The clamp removal is complete rather than partial, and I verified rather than assumed: `same_tokenizer` requires equal `vocab_size` **and** identical ids over `PROBE_CORPUS`, so the index error it guarded cannot occur once the check passes, and `_teacher_vocab` had no other reader.

Zero test lines removed. Merged with #711 onto current `main` in a scratch worktree first: clean, 104 passed.

Credited in `CONTRIBUTORS.md` (`904553d`).

AmirF194:
The unwire-the-call mutation was the one I wanted this suite to catch too, given the repo's history with it. Thanks for the mutation table.

Issue #690, opened 2026-09-05
Embedding training crashes because _EmbeddingTrainer does not expose model or args


Amix29:
## Summary

`EmbeddingTrainerWrapper.train()` deterministically raises before the first training step because
its `_EmbeddingTrainer` wrapper does not expose the `model` and `args` attributes that `train()`
reads.

This affects the Transformers/PyTorch embedding-training path on current `main`
(`ef2e02b51a1b34546823edce55715d28234805cb`). It is not device-specific.

## Root cause

`_EmbeddingTrainer.__init__()` stores the real Hugging Face trainer in `self._trainer`. It
delegates `train()`, `save_model()`, `add_callback()`, and `state`, but not `model` or `args`.

Immediately before training, `EmbeddingTrainerWrapper.train()` executes:

```python
align_trainable_dtype_for_fp16(
    self.trainer.model,
    fp16=getattr(self.trainer.args, "fp16", False),
    bf16=getattr(self.trainer.args, "bf16", False),
)
```

The first attribute access fails before the dtype helper can decide whether any conversion is
needed.

## Reproduction and impact

Against the exact checkout above, constructing the actual wrapper object around a minimal inner
trainer and executing the attribute-access path produces:

```text
AttributeError: '_EmbeddingTrainer' object has no attribute 'model'
```

The inner trainer's `train()` is never reached. Adding delegating `model` and `args` properties
makes the same reduced contract reach it.

The maintainer also reproduced this end to end with a real config, `load_dataset()`, model setup,
LoRA setup, and `train()` using `hf-internal-testing/tiny-random-gpt2`. The Transformers embedding
path fails after setup in 100% of runs. The regression was introduced by commit `822de16` and
shipped in Soup 0.74.0.

## Recommended fix

Expose the missing interface on `_EmbeddingTrainer`:

```python
@property
def model(self):
    return self._trainer.model

@property
def args(self):
    return self._trainer.args
```

Alternatively, make the outer wrapper access `self.trainer._trainer`, but public delegation is
less coupled and matches the interface already expected by `EmbeddingTrainerWrapper`.

## Acceptance criteria

- [ ] `_EmbeddingTrainer` exposes `model` and `args` from its inner HF trainer.
- [ ] A regression test constructs the real `_EmbeddingTrainer` wrapper and calls
      `EmbeddingTrainerWrapper.train()` after `setup()`; a `MagicMock` in place of the wrapper is
      not sufficient because it auto-creates the missing attributes.
[...]

PR #705, opened 2026-09-05
fix(trainer): support lora.r=0 (full fine-tuning) on task=embedding (#700)

### LINKED ISSUE #700: task=embedding does not honour lora.r: 0 (full fine-tuning) and leaks a raw peft ValueError
Found while verifying #690, and filed separately because the root cause is different.

`lora.r: 0` has meant **full fine-tuning** since v0.73.0 (#340) — it is the spelling that replaced the `unfrozen_parameters: ['.*']` workaround, and `trainer/classifier.py` and `commands/card.py` already treat it as "no adapter". The embedding trainer does not; the value reaches peft raw:

```
ValueError: 'r' should be a positive integer value but the value passed is 0
  peft/tuners/lora/layer.py:236
```

So a config that the schema accepts, that is documented, and that works on other tasks, dies inside a dependency with an error naming neither Soup nor the field the user set.

## Scope

This is the *guard*, not the feature. Whether `task: embedding` should support full fine-tuning at all is a separate question — the minimum bar is that `lora.r: 0` either works or is refused at config-parse time with a message naming the task, rather than producing a peft traceback several seconds into a run.

My preference is that it works, since #340's whole point was that `r=0` is the project-wide spelling for "no adapter" and a task-specific exception to that is a trap. But refusing it explicitly is an acceptable answer and a much smaller change.

## Acceptance criteria

- [ ] `lora.r: 0` with `task: embedding` either trains without an adapter, or is rejected at config load with an error naming both the task and the field
- [ ] No raw peft `ValueError` escapes in either case
- [ ] A test that constructs the real trainer rather than a `MagicMock` — see #690, where `tests/test_embedding.py` passes 79 tests over a path that crashes in production precisely because a `MagicMock` absorbs the failure
- [ ] The other trainers' handling of `r=0` is checked for the same gap; if any share it, say so rather than fixing one

Related: #690, which is the unconditional crash on this same trainer. A fix for that one does not fix this one, and vice versa. Found while verifying @Amix29's report.

kok-o:
Fixes #700

## Description
When `task: embedding` is configured with `lora.r: 0` (full fine-tuning), `_setup_transformers()` previously passed `r=0` to `peft.LoraConfig`, causing a raw `ValueError: Lora rank r must be > 0`. Furthermore, full fine-tuning requires explicit `fp32` master weights rather than loading at the frozen-base checkpoint dtype.

## Changes
- **`src/soup_cli/config/schema.py`**: Allow `task in ('sft', 'embedding')` with `lora.r: 0` and `quantization: 'none'`.
- **`src/soup_cli/trainer/embedding.py`**:
  - In `_setup_transformers()`:
    - Skip PEFT wrapping when `lora.r == 0`, leaving the base model with trainable parameters.
    - Resolve base load dtype via `resolve_base_load_dtype(self.device, full_finetune=is_full_finetune(tcfg))`, correctly pinning `torch.float32` for full fine-tuning.
  - In `setup()`:
    - Explicitly check `if tcfg.lora.r == 0` before computing trainable parameters on the base model without `PeftModel`.
    - If `lora.r > 0` and the model lacks `get_nb_trainable_parameters`, fail loudly rather than masking the error.
    - Display `Full fine-tuning:` in the console log summary without duplicate lines.
  - Rebased onto #708: Removed `__getattr__` and inherited explicit `model` and `args` delegation on `_EmbeddingTrainer`, preserving honest guard decline behavior for `attach_empty_param_group_guard()`.
- **`tests/test_issue491_frozen_base_load_dtype.py`**:
  - Include `EmbeddingTrainerWrapper` alongside `PretrainTrainerWrapper` in the `resolve_base_load_dtype` resolver check, since embedding now also has full-FT support.
- **`tests/test_issue700_embedding_lora_zero.py`**:
  - Comprehensive unit/integration tests with real `torch.nn.Module` instances.
  - Verifies schema validation, PEFT bypass when `r=0`, dtype resolution (`float32` for full-FT vs. `auto` for LoRA), console summary labels, loud failure on unexpected non-PEFT model when `r>0`, property delegation, and absence of `__getattr__` (confirming `attach_empty_param_group_guard` honestly declines).

## Audit of Other Trainers (Acceptance Criterion 4)
Audited other trainers in `src/soup_cli/trainer/` for `lora.r: 0`:
- `classifier.py` and `sft.py` already support `lora.r: 0` / full fine-tuning.
- Other trainers (`asr`, `bco`, `distill`, `dpo`, `grpo`, `ipo`, `kto`, `online_dpo`, `orpo`, `ppo`, `pretrain`, `reward_model`, `simpo`, `stream_setup`) currently pass `r=tcfg.lora.r` directly to PEFT.

### DIFF
--- a/changelog.d/0.74.0/705.fixed.md
+++ b/changelog.d/0.74.0/705.fixed.md
@@ -1,0 +1,4 @@
+- Fixed `ValueError: Lora rank r must be > 0` when training embeddings with
+  `lora.r: 0` (full fine-tuning) on `task: embedding` (#700 by @kok-o in #705).
+  `_setup_transformers()` now skips PEFT wrapping when `r == 0`, and
+  `_EmbeddingTrainer` delegates `model` and `args` attributes.
--- a/src/soup_cli/config/schema.py
+++ b/src/soup_cli/config/schema.py
@@ -64,7 +64,7 @@ class LoraConfig(BaseModel):
         ge=0,
         description=(
             "LoRA rank. 0 = full fine-tuning: no adapter, every base "
-            "parameter trains (sft + transformers + text + "
+            "parameter trains (sft / embedding + transformers + text + "
             "quantization='none' only)."
         ),
     )
@@ -5023,9 +5023,9 @@ class SoupConfig(BaseModel):
         mutually exclusive with the LoRA feature flags and with the other two
         (each independently decides what trains).
 
-        Scoped to ``task='sft'`` ON PURPOSE. ``classifier`` / ``reranker`` /
-        ``cross_encoder`` (``classifier_lora``, v0.71.12 #146) and ``asr``
-        (``asr_lora``, v0.71.32) have gated LoRA behind their own opt-in flag
+        Scoped to ``task in ('sft', 'embedding')`` (#340, #700). ``classifier`` /
+        ``reranker`` / ``cross_encoder`` (``classifier_lora``, v0.71.12 #146) and
+        ``asr`` (``asr_lora``, v0.71.32) have gated LoRA behind their own opt-in flag
         for releases, so ``lora.r: 0`` already parses and is already harmless
         there; a blanket gate would break configs that work today. Tasks that
         do pass the rank straight to peft keep their existing behaviour
@@ -5033,7 +5033,7 @@ class SoupConfig(BaseModel):
         a new refusal this issue never measured.
         """
         tcfg = self.training
-        if tcfg.lora.r != 0 or self.task != "sft":
+        if tcfg.lora.r != 0 or self.task not in ("sft", "embedding"):
             return self
         if tcfg.stream_layers:
             # Streaming has its own, more specific refusal (the decoder lives
@@ -5044,7 +5044,7 @@ class SoupConfig(BaseModel):
             raise ValueError(
                 f"training.lora.r=0 (full fine-tuning) requires "
                 f"backend='transformers'; got backend={self.backend!r}. The "
-                f"full-FT branch is wired in the transformers SFT trainer."
+                f"full-FT branch is wired in the transformers SFT and embedding trainers."
             )
         if self.modality != "text":
             raise ValueError(
--- a/src/soup_cli/trainer/embedding.py
+++ b/src/soup_cli/trainer/embedding.py
@@ -12,8 +12,8 @@ from soup_cli.utils.gpu import (
     bf16_fp16_flags,
     estimate_batch_size,
     model_size_from_name,
+    resolve_base_load_dtype,
     resolve_device_map,
-    resolve_frozen_base_load_dtype,
 )
 from soup_cli.utils.mixed_precision import align_trainable_dtype_for_fp16
 from soup_cli.utils.seeding import apply_training_seed, training_seed_kwargs
@@ -87,10 +87,15 @@ class EmbeddingTrainerWrapper:
         else:
             self._setup_transformers(cfg, tcfg)
 
-        trainable, total = self.model.get_nb_trainable_parameters()
-        pct = 100 * trainable / total
+        if tcfg.lora.r == 0:
+            trainable = sum(p.numel() for p in self.model.parameters() if p.requires_grad)
+            total = sum(p.numel() for p in self.model.parameters())
+        else:
+            trainable, total = self.model.get_nb_trainable_parameters()
+        pct = 100 * trainable / total if total > 0 else 0.0
+        label = "Full fine-tuning" if tcfg.lora.r == 0 else "LoRA applied"
         console.print(
-            f"[green]LoRA applied:[/] {trainable:,} trainable"
+            f"[green]{label}:[/] {trainable:,} trainable"
             f" / {total:,} total ({pct:.2f}%)"
         )
 
@@ -240,9 +245,13 @@ class EmbeddingTrainerWrapper:
 
         console.print(f"[dim]Loading model: {cfg.base}[/]")
         dev_map = resolve_device_map(self.device)
+        from soup_cli.trainer.sft import is_full_finetune
+
         model_kwargs = {
             "trust_remote_code": self._trust_remote_code, "device_map": dev_map,
-            "torch_dtype": resolve_frozen_base_load_dtype(self.device),
+            "torch_dtype": resolve_base_load_dtype(
+                self.device, full_finetune=is_full_finetune(tcfg)
+            ),
         }
         if quant_config_obj is not None:
             model_kwargs["quantization_config"] = quant_config_obj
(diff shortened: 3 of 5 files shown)


MakazhanAlpamys:
@kok-o — **this genuinely fixes #700 end to end and I verified it through the real CLI. Two blocking findings, both of which I reproduced myself rather than inferred from the diff.**

Sequencing first, so it does not arrive as a surprise: **#708 by @jagadeepmamidi lands before this, and you rebase onto it.** They were assigned #690 at 17:53 UTC and you opened at 18:43; neither of you could see the other, and the overlap is mine for filing two adjacent issues and watching neither. Your PR fixes #690 as a side effect — I confirmed it below — but theirs is the minimal, correctly-scoped fix for a 100% crash, and yours needs rework regardless, so it should not gate theirs. The rebase makes your diff *smaller and better*: drop the now-duplicate `model` / `args` properties, drop `__getattr__` (which resolves blocking finding 1 by construction), keep the schema change, the `r == 0` branch, the `hasattr` guard and the label, and add the dtype fix.

## Verified by execution, both directions

Real CLI — tiny Llama encoder, real tokenizer, real `soup.yaml`, `python -m soup_cli.cli train -c soup.yaml --yes --no-reexec` on CPU:

```
main,  lora.r=0  -> exit 1   ValueError: r should be a positive integer value but the value passed is 0
main,  lora.r=4  -> exit 1   AttributeError: _EmbeddingTrainer object has no attribute model
this,  lora.r=0  -> exit 0   Full fine-tuning ... Training Complete
this,  lora.r=4  -> exit 0   LoRA applied ... Training Complete
```

**Mutations — eight applied, five killed, three survived.** The four that matter to your feature all die: the schema gate, the `r == 0` branch, the `hasattr(get_nb_trainable_parameters)` guard, and the `Full fine-tuning` / `LoRA applied` label. Zero test lines removed, ruff clean, 71 passed, and the wider regression set is 215 passed — identical to main. The three survivors are finding 1.

---

## BLOCKING 1 — `__getattr__` is untested, and it flips a safety guard from an honest decline to a false success

Three mutations survived, and they are one finding: the delegation is implemented **twice** — explicit `model` / `args` properties *and* a catch-all `def __getattr__` — so no test can tell which one is carrying it.

| mutation | result |
|---|---|
| delete the `model` property only | **survived** — 8 passed |
| delete the `args` property only | **survived** — 8 passed |
| delete `__getattr__` only | **survived** — 8 passed |
| delete all three | killed |

Deleting `__getattr__` alone changes nothing observable to your suite, and it is the half with real side effects. Measured on a real `_EmbeddingTrainer` against a stub inner HF trainer, `main` versus this branch:

```
                                            main     this
has __getattr__                             False    True
hasattr(outer, create_optimizer)            False    True
attach_empty_param_group_guard(outer)       False    True     <--
outer marked _soup_empty_group_guard        False    True     <--
INNER create_optimizer actually wrapped     False    False    <--
```

`attach_empty_param_group_guard` (#336 / #359) probes `getattr(trainer, "create_optimizer", None)`, and its docstring is explicit that returning `False` for a trainer without one is the *designed* behaviour — "declining leaves such a trainer exactly as it was". With `__getattr__`, that probe resolves through to the inner trainer, so the guard installs itself on the **outer wrapper** and sets `_soup_empty_group_guard = True`, while `Trainer.train()` calls `self.create_optimizer()` on the **inner** object, which stays untouched.

To be precise about severity, because overstating it would be wrong: embedding was already unprotected on `main`, so this does **not** introduce a new DeepSpeed crash. What it does is convert a visible gap into a guard that reports success and does nothing — which is worse than the gap, because the gap is findable. And `tests/test_issue359_deepspeed_guard_coverage.py` is a source scan for the call site, so it passes and cannot see this.

There is also a latent second effect: `__getattr__` reads `self._trainer`, which is itself resolved through `__getattr__` when unset, so a partially-initialised instance raises `RecursionError` where `main` raises `AttributeError`. `__init__` assigns `self._trainer` last, so the window is real. I could not drive it from a live code path, so I am reporting it as robustness rather than a demonstrated outage.

**Fix: delete `def __getattr__` and keep the two explicit properties.** The mutation above shows nothing depends on it. After the #708 rebase you inherit those properties anyway, so this is a deletion, not a rewrite.

---

## BLOCKING 2 — `lora.r: 0` on `task: embedding` fine-tunes the base at checkpoint dtype instead of fp32

Your PR makes `r: 0` a real full fine-tune here, but the load dtype still goes through the frozen-base helper. Measured on your branch:

```
is_full_finetune(tcfg)                                             -> True
embedding.py:  resolve_frozen_base_load_dtype(cuda)                -> auto
sft.py:        resolve_base_load_dtype(cuda, full_finetune=True)   -> torch.float32
```

`resolve_base_load_dtype`'s own docstring states the rule this violates: an optimizer that steps the base needs explicit fp32 master weights. Three consequences, in descending order of how much they worry me:

1. **On pre-Ampere CUDA** (T4 / V100 / GTX 16xx) `resolve_frozen_base_load_dtype` returns `torch.float16`. v0.74.0's #425 / #429 fix casts only `*lora_*` params to fp32 and deliberately leaves the `lora.r: 0` full-FT path alone — which is safe on SFT *precisely because* SFT loads a full-FT base at fp32. Your PR creates the exact combination that reasoning excluded.
2. A bf16 or fp16 checkpoint is fully fine-tuned at that precision, with no fp32 master weights.
3. `commands/train.py::_build_hardware_fit_input` calls the same `is_full_finetune()` and will classify this run `peft="full"`, predicting full-FT VRAM while the trainer loads at checkpoint dtype. That pre-flight/trainer divergence is the specific thing #471 unified `is_full_finetune` to prevent, three weeks ago.

**Fix:** in `_setup_transformers`, `resolve_base_load_dtype(self.device, full_finetune=is_full_finetune(tcfg))`, plus a test pinning fp32 at `r=0`.

---

## Non-blocking

1. **#700's fourth acceptance criterion is unaddressed.** It asks that other trainers be checked for the same gap and the findings *stated* even if not fixed. Twenty other modules under `src/soup_cli/trainer/` pass `r=tcfg.lora.r` straight to peft — `asr`, `bco`, `classifier`, `distill`, `dpo`, `grpo`, `ipo`, `kto`, `online_dpo`, `orpo`, `ppo`, `pretrain`, `reward_model`, `simpo`, `stream_setup` among them. One sentence in the PR body closes it; I am not asking you to fix twenty trainers.
2. **The `hasattr` fallback softens a real failure.** If an `r > 0` run's `get_peft_model` ever failed to wrap, `main` raises loudly; here it silently counts every parameter and still prints `LoRA applied`, because the label keys off `tcfg.lora.r` rather than off the model. Gating the fallback on `tcfg.lora.r == 0` keeps the loud failure where it belongs.
3. Duplicate console line: the `r == 0` branch prints `Full fine-tuning: N parameter tensor(s) trainable (LoRA off)` and then `setup()` prints `Full fine-tuning: X trainable / Y total (Z%)`. Cosmetic.
4. `test_schema_allows_embedding_with_lora_r_zero` also passes on `main`, because the validator short-circuits for non-`sft`. It documents rather than discriminates — harmless, worth knowing.

**The changelog fragment is correct** — `changelog.d/0.74.0/705.fixed.md`, named for the PR, `#705` in the text, `by @kok-o`. Third one in this batch you have got right unprompted.

Wait for #708 to land, rebase, and push the two fixes. Ping me and I will re-run the same eight mutations.

kok-o:
@MakazhanAlpamys Thanks for the thorough review and execution-based verification!

Everything has been addressed and pushed to `fix/700-embedding-lora-zero`:

1. **Rebase onto #708 & Dropped `__getattr__` (Blocking 1)**:
   - Rebased onto latest `upstream/main` containing #708.
   - Removed `__getattr__` completely, relying solely on the explicit `model` and `args` properties.
   - Added `test_embedding_trainer_no_catchall_getattr_and_honest_deepspeed_guard` verifying `__getattr__` is not present, `create_optimizer` is not masked, and `attach_empty_param_group_guard(trainer)` returns `False` honestly without marking the wrapper.

2. **Full fine-tune fp32 load dtype (Blocking 2)**:
   - In `_setup_transformers`, load dtype now resolves via `resolve_base_load_dtype(self.device, full_finetune=is_full_finetune(tcfg))`, correctly ensuring `torch.float32` master weights on full-FT.
   - Added `test_setup_transformers_loads_fp32_for_full_finetune_and_frozen_for_lora` pinning `torch.float32` for `r=0` vs checkpoint/frozen load dtype for `r>0`.

3. **Loud failure preservation**:
   - Gated the parameter counting fallback strictly on `tcfg.lora.r == 0`. When `r > 0`, non-PEFT models fail loudly via `get_nb_trainable_parameters` (pinned in `test_setup_raises_when_lora_r_positive_and_model_not_peft`).

4. **Clean console logging**:
   - Removed the duplicate console output so only the single summary line is displayed.

5. **Trainer audit (Acceptance criterion 4)**:
   - Audited the other trainers in `src/soup_cli/trainer/`: `classifier` and `sft` already support `lora.r: 0`; the other trainers (`asr`, `bco`, `distill`, `dpo`, `grpo`, `ipo`, `kto`, `online_dpo`, `orpo`, `ppo`, `pretrain`, `reward_model`, `simpo`, `stream_setup`) unconditionally pass `lora.r` to PEFT. Noted in the PR description and commit message.

All 11 unit/integration tests in `tests/test_issue700_embedding_lora_zero.py` and existing tests pass cleanly with zero lint or formatting warnings. Ready for your mutation re-run!

MakazhanAlpamys:
Merged as `9b48deb`. Thank you — and the way you closed the first round is the part worth naming.

**I re-verified both blockers myself on `157cc606`, not from the description.**

`__getattr__` is gone, and the guard is back to an honest decline — measured against a real `_EmbeddingTrainer`:

```
has __getattr__                    : False
hasattr(outer,'create_optimizer')  : False
attach_empty_param_group_guard()   : False    (identical to main)
outer _soup_empty_group_guard      : False
```

That was the finding I cared about most: `__getattr__` had turned `attach_empty_param_group_guard`'s deliberate `False` into a `True` that installed itself on the outer wrapper while `Trainer.train()` calls `create_optimizer` on the inner one. A gap is findable; a guard that reports success and does nothing is not.

**And you wrote the inverse mutation, which I did not ask for.** Deleting either property now fails 2 tests each — but `test_embedding_trainer_no_catchall_getattr_and_honest_deepspeed_guard` fails if `__getattr__` comes *back*. That defends against the pattern returning rather than describing today's state, and it is the difference between a fix and a fix that stays fixed.

**The dtype blocker, verified by execution:**

```
lora.r=0 -> is_full_finetune=True  -> torch.float32
lora.r=4 -> is_full_finetune=False -> auto (unchanged)
```

and `embedding.py` genuinely calls `resolve_base_load_dtype(..., full_finetune=is_full_finetune(tcfg))` — mutating that to `False` fails 2 tests, so what is pinned is the value actually passed, not merely one that could be resolved. That closes the pre-Ampere hazard: the frozen-base helper returns `torch.float16` on those cards, and v0.74.0's #425/#429 reasoning explicitly left the `r: 0` path alone *because* full-FT bases load at fp32 on SFT.

**Both non-blocking items done too.** The `hasattr` fallback is gated on `r == 0`, and the mutated build printed exactly the false success I was worried about — `LoRA applied: 4,256 trainable / 4,256 total (100.00%)` — so that gate is real. And #700's fourth criterion, the survey of other trainers, is answered in the PR body rather than skipped.

Rebase onto #708 confirmed: `ab2412e` is an ancestor, so the `model` / `args` properties are inherited rather than duplicated. 141 passed against 130 on main.

Credited in `CONTRIBUTORS.md` (`60fe73d`).

PR #706, opened 2026-09-05
fix(ui): drain training subprocess stdout in background buffer to prevent hang (#688)

### LINKED ISSUE #688: Web UI training subprocess can block when no client drains stdout
## Description

The Web UI starts training with `stdout=PIPE` and `stderr=STDOUT`, but the process output is
drained only by the `/api/train/logs` SSE generator. If no browser is subscribed, or the client
disconnects, the pipe can fill and block the training subprocess on its next write.

Relevant code:

- [training process uses a pipe](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/ui/app.py#L274-L311)
- [SSE clients read directly from `proc.stdout`](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/ui/app.py#L399-L421)

## Steps to reproduce

The original `start_training()` Popen path was run with a bounded synthetic child that writes
1 MiB to stdout and then creates a completion marker.

```text
before any reader drains stdout:
  child still alive: true
  completion marker exists: false

after communicate() drains stdout:
  bytes received: 1,048,576
  exit code: 0
  completion marker exists: true
```

The exact pipe capacity is OS-dependent; 1 MiB is the child payload, not a claimed capacity.

## Expected behavior

Training progress must not depend on the presence or lifetime of a log-streaming client.

## Impact

An UI-launched training job can appear frozen while waiting on stdout rather than on the GPU.
The ordinary CLI output path is not affected by this reproduction, and the amount of real
training output required to trigger the block has not been measured.

Start one client-independent drain when the process launches, writing into a bounded ring buffer
or rotating file. SSE clients should read that stored stream with independent cursors rather
than consume the subprocess pipe directly.

- [ ] Training completes with zero log subscribers.
- [ ] One or multiple subscribers cannot steal lines from each other.
- [ ] Disconnect/reconnect does not stop drainage.
- [ ] Log storage is bounded and cleaned up safely.
- [ ] Stop/exit handling drains or closes the pipe without deadlock.

## Environment

Reproduced under Python 3.12.2 with the original FastAPI/Popen route and a real OS pipe. The
child process was synthetic; no LLM training session was launched.

kok-o:
Fixes #688

## Description
Training subprocesses launched via `soup ui` with `stdout=subprocess.PIPE` froze whenever the OS pipe buffer (~8–64 KiB) filled up without an active SSE subscriber (e.g., browser closed or disconnected). Furthermore, concurrent clients previously stole lines from each other via destructive consumption of `proc.stdout`.

## Changes
- **`src/soup_cli/ui/app.py`**:
  - Introduce thread-safe bounded `TrainLogBuffer` (maxlen=10000) with monotonic sequence IDs and condition variable notifications.
  - Launch a daemon `_drain_stdout_worker` thread as soon as training starts to continuously drain child `stdout` into the buffer.
  - Extract `_resolve_train_argv` for subprocess argument construction.
  - Rewrite `/api/train/logs` SSE endpoint to read non-destructively from `TrainLogBuffer`, supporting multiple concurrent readers and event replay via `Last-Event-ID`.
- **`tests/test_issue688_ui_drain.py`**:
  - `test_start_training_wires_drain_thread_and_drains_without_subscriber`: Goes through `POST /api/train/start` with a child process emitting > 100 KiB of stdout with 0 SSE readers, asserting that the child reaches `rc=0` and verifying that commenting out `_train_drain_thread.start()` kills the mutation.
  - `test_subprocess_exceeding_pipe_threshold_completes_without_reader`: Unit verification of `_drain_stdout_worker` exceeding OS pipe threshold without blocking.
  - `test_multiple_readers_do_not_steal_lines`: Validates concurrent non-destructive readers from `TrainLogBuffer`.
  - `test_last_event_id_reconnect_replays_unseen_lines`: Verifies `Last-Event-ID` resume and replay from the buffer.
- **`tests/test_ui_live_monitor.py`**:
  - Updated `test_logs_returns_event_stream` to initialize `_train_log_buffer`.
  - Removed older, weaker `test_logs_last_event_id_reconnection` which directly injected mock stdout on `_train_process` without a buffer, superseded by `test_last_event_id_reconnect_replays_unseen_lines`.

## Design Notes
- **Buffer lifetime**: `_train_log_buffer` persists after process completion and is only replaced on the next `start_training()`. This deliberately allows late or reconnected SSE clients to replay the full log of a completed run before receiving `event: done`.
- **Stream lifetime**: Synchronous generator readers in `_generate_log_events()` exit upon process completion (`is_done`). Abandoned connections hold a worker thread in Starlette's threadpool until the run finishes.

### DIFF
--- a/changelog.d/0.74.0/706.fixed.md
+++ b/changelog.d/0.74.0/706.fixed.md
@@ -1,0 +1,4 @@
+- Prevented training subprocess hang in Web UI when no client drains stdout
+  (#688 by @kok-o in #706). Output is now continuously consumed by a background
+  thread into a bounded ring buffer, allowing training to proceed regardless of
+  subscriber state and supporting multi-subscriber replay via `Last-Event-ID`.
--- a/src/soup_cli/ui/app.py
+++ b/src/soup_cli/ui/app.py
@@ -42,9 +42,91 @@ class DataInspectRequest(PydanticBaseModel):
     limit: int = Field(default=50, ge=1, le=_MAX_INSPECT_LIMIT)
 
 
+class TrainLogBuffer:
+    """Thread-safe bounded ring buffer for subprocess output lines."""
+
+    def __init__(self, maxlen: int = 10000):
+        self._maxlen = maxlen
+        self._lines: list[tuple[int, str]] = []
+        self._lock = threading.Lock()
+        self._new_line_cond = threading.Condition(self._lock)
+        self._done = False
+        self._total_emitted = 0
+
+    def append(self, text: str) -> None:
+        with self._new_line_cond:
+            idx = self._total_emitted
+            self._total_emitted += 1
+            self._lines.append((idx, text))
+            if len(self._lines) > self._maxlen:
+                self._lines.pop(0)
+            self._new_line_cond.notify_all()
+
+    def mark_done(self) -> None:
+        with self._new_line_cond:
+            self._done = True
+            self._new_line_cond.notify_all()
+
+    def is_done(self) -> bool:
+        with self._lock:
+            return self._done
+
+    def get_lines_from(self, start_idx: int) -> list[tuple[int, str]]:
+        with self._lock:
+            return [item for item in self._lines if item[0] >= start_idx]
+
+    def wait_for_lines_or_done(
+        self, next_idx: int, timeout: float = 0.5
+    ) -> tuple[list[tuple[int, str]], bool]:
+        """Wait until lines >= next_idx are available or process is marked done."""
+        with self._new_line_cond:
+            available = [item for item in self._lines if item[0] >= next_idx]
+            if available or self._done:
+                return available, self._done
+            self._new_line_cond.wait(timeout=timeout)
+            available = [item for item in self._lines if item[0] >= next_idx]
+            return available, self._done
+
+
+def _drain_stdout_worker(proc: subprocess.Popen, log_buffer: TrainLogBuffer) -> None:
+    """Continuously read stdout of proc until EOF and write to log_buffer."""
+    stdout = getattr(proc, "stdout", None)
+    if stdout is None:
+        log_buffer.mark_done()
+        return
+
+    try:
+        while True:
+            raw_line = stdout.readline()
+            if not raw_line or raw_line == b"":
+                break
+            if isinstance(raw_line, bytes):
+                text = raw_line.decode("utf-8", errors="replace").rstrip("\n\r")
+            elif isinstance(raw_line, str):
+                text = raw_line.rstrip("\n\r")
+            else:
+                break
+            log_buffer.append(text)
+    except (ValueError, OSError) as exc:
+        logger.debug("Drain worker exception: %s", exc)
+    finally:
+        try:
+            stdout.close()
+        except Exception:
+            pass
+        log_buffer.mark_done()
+
+
+def _resolve_train_argv(config_path: str) -> list[str]:
+    """Construct command-line arguments for launching training subprocess."""
+    return [sys.executable, "-m", "soup_cli", "train", "--config", config_path, "--yes"]
+
+
 # Global state for training process
 _train_process: Optional[subprocess.Popen] = None
 _train_config_path: Optional[str] = None
+_train_log_buffer: Optional[TrainLogBuffer] = None
+_train_drain_thread: Optional[threading.Thread] = None
 _train_lock = threading.Lock()
 
 # Auth token generated at startup — printed to console for the user.
@@ -388,10 +470,19 @@ def create_app(host: str = "127.0.0.1", port: int = 7860):
 
             _train_config_path = config_path
             _train_process = subprocess.Popen(
-                [sys.executable, "-m", "soup_cli", "train", "--config", config_path, "--yes"],
+                _resolve_train_argv(config_path),
                 stdout=subprocess.PIPE,
                 stderr=subprocess.STDOUT,
             )
+            global _train_log_buffer, _train_drain_thread
+            _train_log_buffer = TrainLogBuffer(maxlen=10000)
+            _train_drain_thread = threading.Thread(
+                target=_drain_stdout_worker,
+                args=(_train_process, _train_log_buffer),
+                daemon=True,
+                name="soup_train_drain",
+            )
+            _train_drain_thread.start()
             return {"started": True, "pid": _train_process.pid}
 
     @app.get("/api/train/status", dependencies=[Depends(_verify_token)])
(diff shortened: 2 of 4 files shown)


MakazhanAlpamys:
@kok-o — the production change is the right mechanism and I believe it fixes #688. I am blocking on two things: CI is red on a test you did not touch, and **the mutation that reproduces #688 exactly survives all three of your new tests.**

## 1. The core mutation survives (blocking)

I replaced the one line that installs the drain, in `start_training`:

```python
-            _train_drain_thread.start()
+            pass  # MUTATED: drain thread never started
```

That is #688 restored, byte for byte. Your suite:

```
$ PYTHONPATH=<worktree>/src py -3.10 -m pytest tests/test_issue688_ui_drain.py -q --no-cov
3 passed in 27.74s     # unmutated
3 passed in  6.21s     # drain thread never started
```

Three of three still green, and the wall clock dropped 4.5x — which is the tell: without the thread there is nothing to wait for, because none of the three tests was ever waiting on it.

Reading them back:

- `test_subprocess_exceeding_pipe_threshold_completes_without_reader` starts the drain thread **in the test body**. It proves `_drain_stdout_worker` works. It cannot observe whether `start_training` ever calls it.
- `test_multiple_readers_do_not_steal_lines` never issues an HTTP request, despite its docstring saying "Two concurrent consumers reading from `/api/train/logs`". It drives `TrainLogBuffer` directly.
- `test_last_event_id_reconnect_replays_unseen_lines` assigns `ui_app_module._train_log_buffer` by hand and leaves `_train_process = None`.

So the wiring — `POST /api/train/start` installing a live drain on the real `Popen` — has no coverage. This is the single most common defect in this codebase: the helper is correct, the helper is tested, and the command never calls it (#363, #423, #659, #672, #675). It is worth saying that the helper tests are good ones and I would keep all three; they are just not the test this PR needs.

**What would close it:** one test that goes through `POST /api/train/start` (with `_resolve_train_argv` or equivalent pointed at a child that emits > 8 KiB and then exits), asserts the child reaches rc=0 with no SSE subscriber, and **fails when `_train_drain_thread.start()` is removed**. Please state in the PR that you ran that mutation and it was killed.

## 2. An existing test is red on all nine cells (blocking)

```
FAILED tests/test_ui_live_monitor.py::TestTrainLogsSSE::test_logs_last_event_id_reconnection
  AssertionError: assert ('Line 2' in 'event: done\ndata: {}\n\n' or 'Line 3' in ...)
```

Reproduced locally at `0503ddb` on py3.10, so it is not a runner artifact.

The cause is a real contract change: that test sets `ui_mod._train_process` directly and never calls `start_training`, so under your code `_train_log_buffer` is `None` and the generator short-circuits to `event: done`. In production the two are always created together, so I do not think this is a user-visible regression — but I am not willing to guess that on your behalf, and a red cell cannot merge.

Pick one and say which in the PR:
- update that test to construct the new state (it is the older, weaker duplicate of your own reconnect test), or
- have `_generate_log_events` fall back when a process exists without a buffer.

I lean towards the first, since your `test_last_event_id_reconnect_replays_unseen_lines` supersedes it — but then delete it rather than leave two tests for one behaviour.

## 3. Non-blocking, worth a line in the PR

**Abandoned streams now hold a threadpool thread for the whole run.** `_generate_log_events` is a sync generator in `while True` that only exits on `is_done`; there is no `request.is_disconnected()` check. `main` had the same shape, but it also had exactly one possible reader because it consumed the pipe. Your change makes concurrent subscribers a supported case, so N closed browser tabs now pin N threads out of Starlette's default 40 until training ends. Not a merge blocker and arguably out of scope, but please note it in the PR so it does not get discovered later as new.

**Buffer lifetime.** `_train_log_buffer` is only replaced in `start_training`, so after a run finishes a fresh SSE connection replays the entire completed run and then closes, where `main` returned `event: done`. Probably an improvement; just make it deliberate.

Everything else — the ring buffer, the condition variable, the monotonic `_total_emitted` ids surviving eviction, `bytes`/`str` handling in the worker — reads correctly to me, and the multi-subscriber replay design is better than what I would have asked for.

Ping me when it is pushed and I will re-run the same mutation.

kok-o:
@MakazhanAlpamys Thanks for the feedback!

Both blocking findings and the non-blocking notes have been addressed and pushed to `fix/688-ui-drain-stdout`:

1. **Mutation killed via `POST /api/train/start` wiring test (Blocking 1)**:
   - Extracted `_resolve_train_argv` in `soup_cli/ui/app.py`.
   - Added `test_start_training_wires_drain_thread_and_drains_without_subscriber` in `tests/test_issue688_ui_drain.py`. It drives the real `POST /api/train/start` endpoint with a child generating > 100 KiB (1,000 lines) of stdout with zero SSE subscribers attached, asserting the child completes with `rc == 0` and all 1,000 lines are buffered.
   - **Mutation test verified**: Replacing `_train_drain_thread.start()` with `pass` immediately caused the child to block on write when the pipe buffer filled, hitting the timeout and failing with:
     ```
     Failed: Subprocess blocked on write without stdout reader — _train_drain_thread was not started by start_training() (#688)
     ```
     Mutation is confirmed killed.

2. **Resolved `test_ui_live_monitor.py` failure (Blocking 2)**:
   - Removed the older, weaker `test_logs_last_event_id_reconnection` in `tests/test_ui_live_monitor.py`, which directly injected a mock stdout on `_train_process` without a buffer. It is fully superseded by `test_last_event_id_reconnect_replays_unseen_lines` in `tests/test_issue688_ui_drain.py`.
   - Updated `test_logs_returns_event_stream` to construct and clean up `_train_log_buffer`.
   - All 17 tests in `tests/test_ui_live_monitor.py` and all 4 tests in `tests/test_issue688_ui_drain.py` now pass green (21 total).

3. **Non-blocking items noted in PR body**:
   - Explicitly documented that `_train_log_buffer` lifetime is preserved past process completion until the next `start_training()`, deliberately allowing late or reloaded SSE subscribers to replay the full run logs before receiving `event: done`.
   - Noted the threadpool thread behavior for abandoned SSE streams until run completion.

Ready for your mutation re-run!

MakazhanAlpamys:
Merged as `53f3369`. Thank you — and the second round is what made this a fix rather than a helper.

**The mutation that reproduces #688 now fails a test.** I re-ran the exact one that survived last time:

```
baseline                                              4 passed
_train_drain_thread.start() -> no-op  (#688 restored) 1 failed
```

`test_start_training_wires_drain_thread_and_drains_without_subscriber` is what catches it. Extracting `_resolve_train_argv` so the wiring could be driven through `POST /api/train/start` was the right way to get there.

Worth restating why that round was necessary, because the first version looked complete: three tests, all green, all good tests — and every one of them started the drain thread in its own body or assigned the buffer by hand. The helper was correct and covered; nothing proved the command called it. The tell was the wall clock dropping 4.5x under the mutation, because nothing was waiting on a thread that no longer existed. That is this repo's most frequent defect (#363, #423, #659, #672, #675) and it now has a guard.

You also kept the three original tests rather than replacing them, which is right — they were good, just not the one the PR needed.

**Nine removed test lines, all audited**: the `MagicMock` fixture in `test_logs_last_event_id_reconnection` that set `_train_process` directly. The test survives at its name and position, rebuilt on `TrainLogBuffer` — which is the contract that now exists. No test function deleted.

**112 passed** across all five UI files on a tree merged with current `main`, which by then also carried your own #707. The two PRs touch the same file and do not fight.

One thing recorded in the merge commit rather than held against you: the SSE generator still has no `request.is_disconnected()` check, so an abandoned stream holds a threadpool thread until the run ends. `main` had the same shape with a single possible reader; supporting concurrent subscribers multiplies it by the number of stale tabs. Out of scope here, and now written down.

That is all three of yours merged today — #705, #707, #706. This one also unblocked @Amix29's README PR, which was waiting on it.

PR #707, opened 2026-09-05
fix(ui): require authentication on read endpoints and SSE streams (#687)

### LINKED ISSUE #687: Web UI exposes run configs and training logs without authentication
## Description

The Web UI applies Bearer authentication to mutating endpoints, but run details and live
training logs can be read without a token. Run records can contain full configuration JSON,
including dataset paths or other operational details, and logs may contain sensitive content.

Relevant code:

- [Bearer check is scoped to mutating endpoints](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/ui/app.py#L112-L120)
- [run detail route has no dependency](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/ui/app.py#L181-L192)
- [training-log SSE route has no dependency](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/ui/app.py#L389-L430)

## Steps to reproduce

Using the original routes with FastAPI's ASGI test client and a synthetic tracker record:

```text
GET /api/runs/fixture, no Authorization header:    HTTP 200
GET /api/runs/fixture, incorrect Bearer token:     HTTP 200
GET /api/train/logs, no Authorization header:      HTTP 200 with log content
POST /api/train/stop, no Authorization header:     HTTP 401 (control)
```

The maintainer independently confirmed these results with real ASGI requests.

## Expected behavior

Private read APIs must require authorization, including when the UI is exposed with
`soup ui --public`. Following the existing `soup serve` precedent, Soup should refuse to start
on a non-loopback address if authentication is disabled.

## Impact and conditions

The default `127.0.0.1` binding substantially reduces severity. The issue becomes important with
`soup ui --public`, a custom non-loopback host, a reverse proxy, or another environment where
third parties can reach the listener. `--public` binds to `0.0.0.0` and presents a QR code, so
remote reachability is an intentional product path. No public Internet exposure or real secret
theft is claimed.

Current documentation explicitly promises authentication for mutating endpoints only, so this
is a security-design gap rather than a mismatch with that documented promise.

- [ ] Apply authentication to all private read APIs, including run details and relevant SSE
      endpoints.
- [ ] Keep `/api/health` and the static UI at `/` reachable without authentication; document any
      other deliberately public route explicitly.
- [ ] Refuse startup when binding to a non-loopback address with authentication disabled,
      matching `soup serve` behavior.
- [ ] Authenticate live-log streaming without putting a durable bearer token in the URL or logs.
      Do not use `?token=`. Use either a short-lived, single-use ticket obtained through an
      authenticated POST or an authenticated `fetch()` streaming implementation.
- [ ] Test missing, incorrect, and correct credentials for both ordinary and streaming reads.
- [ ] Test the non-loopback/auth-disabled startup refusal.
- [ ] Preserve loopback binding as the default and document the security implications of
      non-loopback exposure.

## Environment

Reproduced under Python 3.12.2 with the original FastAPI route functions, real ASGI requests,
and synthetic run/log data. No real configuration or secret was accessed.

kok-o:
Fixes #687

## Description
The Web UI previously guarded only mutating endpoints (\POST\/\DELETE\) with Bearer token authentication. Run records (containing config JSON and dataset paths), system info, and live training log streams were accessible unauthenticated. Furthermore, binding to non-loopback addresses without auth presented an unauthenticated exposure risk.

## Changes
- Protect private read endpoints (\/api/runs\, \/api/runs/compare\, \/api/runs/{id}\, \/api/system\, \/api/templates\, \/api/train/status\, \/api/train/progress\, \/api/config/schema\, \/api/recipes\, \/api/tool-outputs\) with \_verify_token\.
- Keep public routes accessible without auth: root \/\, \/static\, and \/api/health\.
- Protect SSE streams (\/api/train/logs\, \/api/train/metrics/live\, \/api/train/stream\) via Bearer token or short-lived single-use tickets (\POST /api/auth/ticket\). Ephemeral tickets expire in 30 seconds and are consumed on connection, preventing exposure of durable tokens in URL query strings (\?token=\ is rejected).
- Add \--no-auth\ option to \soup ui\ (loopback only) and refuse startup with exit code 2 when binding to non-loopback hosts with authentication disabled, matching \soup serve\ precedent.
- Update frontend (\pp.js\) to auto-inject Bearer tokens in \pi()\ calls and exchange tickets for \EventSource\ connections.
- Add 34 unit, integration, and security tests in \	ests/test_issue687_ui_auth.py\ and update \	ests/test_ui.py\.

### DIFF
--- a/changelog.d/0.74.0/707.fixed.md
+++ b/changelog.d/0.74.0/707.fixed.md
@@ -1,0 +1,6 @@
+- Required authentication on Web UI read endpoints and SSE streams
+  (#687 by @kok-o in #707). Run configurations, logs, and system metrics now
+  require Bearer authorization. SSE endpoints authenticate via short-lived,
+  single-use tickets exchanged over an authenticated POST, avoiding durable
+  tokens in query strings. Non-loopback binding with auth disabled is rejected
+  with exit code 2.
--- a/changelog.d/0.74.0/707.fixed.md
+++ b/changelog.d/0.74.0/707.fixed.md
@@ -2,5 +2,5 @@
   (#687 by @kok-o in #707). Run configurations, logs, and system metrics now
   require Bearer authorization. SSE endpoints authenticate via short-lived,
   single-use tickets exchanged over an authenticated POST, avoiding durable
-  tokens in query strings. Non-loopback binding with auth disabled is rejected
-  with exit code 2.
+  tokens in query strings. Non-loopback binding requires a valid authentication
+  token and is rejected with exit code 2 if missing or invalid.
--- a/src/soup_cli/commands/ui.py
+++ b/src/soup_cli/commands/ui.py
@@ -52,11 +52,16 @@ def ui(
             "When omitted, a fresh token is generated each startup."
         ),
     ),
+    no_auth: bool = typer.Option(
+        False,
+        "--no-auth",
+        help="Disable authentication (loopback only).",
+    ),
 ):
     """Launch the Soup Web UI.
 
     A Bearer auth token is auto-generated at startup and printed to the console.
-    Mutating API endpoints (POST/DELETE) require 'Authorization: Bearer <token>'.
+    API endpoints require 'Authorization: Bearer <token>'.
 
     `--public` exposes the server on 0.0.0.0 for phone-on-LAN access. The
     auth token is embedded in a phone-scannable URL + ASCII QR code so a
@@ -88,20 +93,30 @@ def ui(
     if public and host == "127.0.0.1":
         host = "0.0.0.0"
 
+    # Security: refusing startup on non-loopback when authentication is disabled (#687)
+    if host not in {"127.0.0.1", "localhost", "::1"} and no_auth:
+        from rich.markup import escape as _rich_escape
+
+        console.print(
+            f"[red]Error:[/] binding non-loopback host '{_rich_escape(str(host))}' "
+            "with authentication disabled is not permitted."
+        )
+        raise typer.Exit(code=2)
+
     token = get_auth_token()
 
     if show_token:
         console.print(token)
         raise typer.Exit()
 
-    app = create_app(host=host, port=port)
+    app = create_app(host=host, port=port, no_auth=no_auth)
 
     url = f"http://{host}:{port}"
 
     panel_body = (
         f"URL:    [bold]{url}[/]\n"
         f"Token:  [bold]{token}[/]\n\n"
-        f"Mutating API endpoints require:\n"
+        f"API endpoints require:\n"
         f"  [dim]Authorization: Bearer {token}[/]\n\n"
         f"Pages:\n"
         f"  [bold]Dashboard[/]      - View experiments, loss charts, system info\n"
--- a/src/soup_cli/commands/ui.py
+++ b/src/soup_cli/commands/ui.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import ipaddress
 from typing import Optional
 
 import typer
@@ -9,6 +10,16 @@ from rich.console import Console
 from rich.panel import Panel
 
 console = Console()
+
+
+def _is_loopback(host: str) -> bool:
+    """Return True if host is a loopback address or localhost."""
+    if host.lower() == "localhost":
+        return True
+    try:
+        return ipaddress.ip_address(host.strip("[]")).is_loopback
+    except ValueError:
+        return False
 
 
 def ui(
@@ -52,11 +63,6 @@ def ui(
             "When omitted, a fresh token is generated each startup."
         ),
     ),
-    no_auth: bool = typer.Option(
-        False,
-        "--no-auth",
-        help="Disable authentication (loopback only).",
-    ),
 ):
     """Launch the Soup Web UI.
 
@@ -93,23 +99,33 @@ def ui(
     if public and host == "127.0.0.1":
         host = "0.0.0.0"
 
-    # Security: refusing startup on non-loopback when authentication is disabled (#687)
-    if host not in {"127.0.0.1", "localhost", "::1"} and no_auth:
-        from rich.markup import escape as _rich_escape
-
-        console.print(
-            f"[red]Error:[/] binding non-loopback host '{_rich_escape(str(host))}' "
-            "with authentication disabled is not permitted."
-        )
-        raise typer.Exit(code=2)
-
+    # Security: refusing startup on non-loopback when authentication token is
+    # missing or invalid (#687)
     token = get_auth_token()
+    if not _is_loopback(host):
+        valid = False
+        if token:
+            try:
+                from soup_cli.utils.qr_url import validate_token
+
+                validate_token(token)
+                valid = True
+            except (TypeError, ValueError):
+                valid = False
+        if not valid:
+            from rich.markup import escape as _rich_escape
+
+            console.print(
+                f"[red]Error:[/] binding non-loopback host '{_rich_escape(str(host))}' "
+                "requires a valid authentication token."
+            )
+            raise typer.Exit(code=2)
 
     if show_token:
         console.print(token)
         raise typer.Exit()
 
-    app = create_app(host=host, port=port, no_auth=no_auth)
+    app = create_app(host=host, port=port)
 
     url = f"http://{host}:{port}"
 
--- a/src/soup_cli/ui/app.py
+++ b/src/soup_cli/ui/app.py
@@ -73,8 +73,49 @@ def set_auth_token(token: str) -> None:
         _auth_token = validated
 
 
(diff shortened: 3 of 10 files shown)


AmirF194 (commented):
I ran the pre-existing UI test suite against this PR's head (69d6d5a) in a clean container.

Adding `Depends(_verify_token)` to the read endpoints in src/soup_cli/ui/app.py breaks four pre-existing test files this diff does not touch. They call those endpoints through `TestClient(create_app())` with no Authorization header, so what used to return 200 now returns 401.

```
$ python -m pytest tests/test_ui_metrics.py tests/test_ui_config_builder.py tests/test_ui_live_monitor.py tests/test_v0539.py -q
FAILED tests/test_ui_metrics.py::TestMetricsFullFields::test_metrics_includes_grad_norm - assert 401 == 200
FAILED tests/test_ui_config_builder.py::TestConfigSchemaEndpoint::test_schema_returns_json - assert 401 == 200
FAILED tests/test_ui_live_monitor.py::TestTrainProgress::test_progress_not_running - assert 401 == 200
FAILED tests/test_v0539.py::test_api_tool_outputs_endpoint - assert 401 == 200
34 failed, 64 passed
```

This lines up with the 34 failures already on this PR's own CI run across the ubuntu/windows/macos jobs.

The new test file this PR adds, tests/test_issue687_ui_auth.py, already carries the fix: an `_auth_headers()` helper that sends `Bearer <token>`. tests/test_ui_metrics.py defines that same helper but never calls it. The other three files do not send a header at all. Adding `headers=_auth_headers()` to the calls in these four files should get CI back to green.

MakazhanAlpamys:
@kok-o — **the security design here is exactly what I asked for on #687, and your new tests are the strongest of any PR in this batch.** I could not get a mutation past them. Two things block it, and one of them is a scope call I need to make rather than you.

## What I verified, by execution

Worktree at `69d6d5a`, `PYTHONPATH` pinned into the worktree (checked with `soup_cli.__file__`), py3.10, `--no-cov`.

```
tests/test_issue687_ui_auth.py .................................. 34 passed
```

Then five mutations against your own suite:

| mutation | result |
|---|---|
| M1 drop `Depends(_verify_token)` from `GET /api/runs/{run_id}` | **KILLED** (1 failed) |
| M2 `_verify_token_or_ticket` accepts any ticket | **KILLED** (6 failed) |
| M3 ticket reusable — `_tickets.get` instead of `.pop` | **KILLED** (1 failed) |
| M4 non-loopback + auth-disabled refusal removed | **KILLED** (2 failed) |
| M5 drop auth from `GET /api/train/logs` | **KILLED** (4 failed) |

Five for five, including the single-use property, which is the one I expected to survive. The `POST /api/auth/ticket` exchange with a 30 s single-use ticket is precisely the shape I ruled for on #687 — no `?token=`, no durable credential in a URL — and `secrets.compare_digest` on the header path is right.

## 1. Nineteen existing tests are red on all nine cells (blocking)

Reproduced locally, so it is not a runner artifact:

```
$ py -3.10 -m pytest tests/test_ui.py tests/test_ui_live_monitor.py tests/test_ui_config_builder.py -q --no-cov
19 failed, 54 passed
```

They split into two kinds, and the difference matters:

**(a) Four tests that pin the old contract by name** — `test_logs_no_auth_required`, `test_metrics_live_no_auth_required`, `test_progress_no_auth_required`, and `test_logs_closes_when_training_stops`. These assert the exact behaviour this PR deliberately reverses. **Please rewrite rather than delete them**: invert each into "401 without a token, 200 with one", so the file keeps a test at that name's position that now defends the new rule. A deleted test leaves a hole where a future refactor can quietly re-open the endpoint; an inverted one catches it.

**(b) Fifteen that are collateral** — `test_ui_config_builder.py` schema/recipes, the SSE stream and progress tests. Those just need `_auth_headers()` added; `tests/test_ui_live_monitor.py` already defines that helper at the top of the file and does not use it.

You touched `tests/test_ui.py` (+18/-14) but not the other two files. I read those 14 removed lines and they are legitimate updates, not weakenings — the same treatment applied to the remaining two files is all this needs.

## 2. `--no-auth` — I am asking you to drop it (blocking, and it is my wording that caused it)

The PR adds a new CLI flag that lets a user turn authentication off. **There is no way to disable auth on `main`** — I checked: no flag, no env var, and `set_auth_token` routes through `qr_url.validate_token`, so even `--auth-token ""` is rejected. So the flag exists only to give the non-loopback refusal something to refuse.

That inverts the issue. #687 asks for *more* authentication and this ships the project's first supported way to have *none*, then guards it. I wrote "a non-loopback bind with auth **somehow** disabled should refuse to start" — the "somehow" was me hedging about a state that did not exist, and you read it as reasonably as it could be read. The fault is mine.

**What I want instead:** delete `--no-auth` and the `no_auth` parameter, and keep the startup refusal keyed on the token actually being present and valid. That satisfies the criterion against a real condition, deletes the hole rather than fencing it, and shrinks the diff. If a future release genuinely needs an auth-off mode, it should arrive as its own PR with its own argument, not as a side effect of a hardening change.

If you disagree, argue it here — I will read it. But I am not shipping a new way to disable authentication inside the PR whose subject is that reads are unauthenticated.

## 3. Non-blocking

- `host not in {"127.0.0.1", "localhost", "::1"}` treats only three spellings as loopback. `127.0.0.0/8` is all loopback, and `[::1]` (bracketed) is a normal way to write the v6 form. Not exploitable — it errs strict, refusing a bind that is actually safe — but `ipaddress.ip_address(host).is_loopback` is one line and handles all of it. Your call.
- `/api/auth/ticket` is authenticated, so unbounded `_tickets` growth needs a valid token first; `_cleanup_expired_tickets` on every create/consume bounds it to one 30 s window. Fine as shipped — noting it so it is on the record.
- The changelog fragment is correct: `changelog.d/0.74.0/707.fixed.md`, PR number in the filename and the text, credited. That is the third one in this batch you have got right without being asked.

Push and I will re-run the same five mutations plus the three UI test files.

MakazhanAlpamys:
Merged as `0347062`. Thank you — this is the strongest security PR the project has had, and two of the choices in it were better than what I asked for.

**Route audit, run by me on the merged tree against current `main`:**

```
/api/runs            -> 401      /api/train/status    -> 401
/api/system          -> 401      /api/train/progress  -> 401
/api/config/schema   -> 401      /api/tool-outputs    -> 401
/api/health          -> 200      /                    -> 200
```

Fifteen endpoints closed; the two that must stay public do.

**Zero test functions deleted**, which I checked rather than trusted: 58 removed test lines, 307 added, and `grep '^-.*def test_'` returns nothing. The four `*_no_auth_required` tests are still there, at their positions, inverted. **That choice is measurable** — the drop-auth-from-`/api/train/logs` mutation now fails **6** tests where it failed 4 before, the two extra being the inverted ones. Deleting them would have left a hole exactly where a future refactor would find it.

Three UI test files: **73 passed** on the merged tree, where the first version left 19 red.

**All five mutations re-killed**: dropping `Depends(_verify_token)` from `GET /api/runs/{run_id}`; the ticket verifier accepting anything; the ticket made reusable via `_tickets.get` instead of `.pop`; the non-loopback refusal removed; auth dropped from `/api/train/logs`. The single-use property is the one I expected to slip through and it does not.

## The two things you did better than the brief

**The SSE ticket design was right first time.** `EventSource` cannot set a header, which is why those routes were open in the first place, and the obvious escape is `?token=` — which lands in server logs, proxy logs and browser history. A single-use 30-second ticket over an authenticated POST is what I would have specified if you had asked, and you did not need to.

**Loopback detection fails closed.** `ipaddress.ip_address(host.strip("[]")).is_loopback` covers `127.0.0.0/8` and bracketed `[::1]` — and the spellings that were v0.74.0's #600/#599/#616 SSRF bypasses (`127.1`, `2130706433`, `0x7f000001`, `0177.0.0.1`) classify as **non**-loopback, so they take the stricter path. That is the correct direction for this helper, and getting it backwards would have been easy.

**And you removed `--no-auth` without arguing.** That flag would have made a hardening PR the first supported way to run Soup's UI with no authentication at all. The refusal now keys on the token being present and valid, which is a real condition rather than an invented one, and matches what `soup serve` has done since #527.

## One thing filed rather than folded in

`/openapi.json`, `/docs` and `/redoc` are still public, so a `--public` bind lets an unauthenticated client enumerate the API surface. Identical on `main`, no run data exposed, and out of #687's scope — but it sits oddly next to a PR about hardening non-loopback binds, so I am filing it separately rather than pretending it is closed.

Credited in `CONTRIBUTORS.md` (`60fe73d`). #706 is the last of your three; it is the one holding up @Amix29's README PR, so it has an audience now.

PR #708, opened 2026-09-05
fix(trainer): expose model and args on the embedding trainer wrapper

### LINKED ISSUE #690: Embedding training crashes because _EmbeddingTrainer does not expose model or args

jagadeepmamidi:
## What does this PR do?

`_EmbeddingTrainer` wraps the real HF trainer in `_trainer` and already
delegates `train` / `save_model` / `add_callback` / `state`, but not
`model` or `args`. `EmbeddingTrainerWrapper.train()` reads those two
attributes for `align_trainable_dtype_for_fp16`, so every
`task: embedding` run on the transformers backend crashed after setup:

This adds the missing properties. The regression tests construct the real
wrapper via `setup()` — `tests/test_embedding.py` stayed green because it
assigns `wrapper.trainer = MagicMock()`, which auto-creates both attributes.

Closes #690.

`lora.r: 0` on the embedding trainer is #700 and is not part of this PR.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

### DIFF
--- a/changelog.d/0.74.0/708.fixed.md
+++ b/changelog.d/0.74.0/708.fixed.md
@@ -1,0 +1,1 @@
+- **`task: embedding` on the transformers backend no longer crashes after setup (#708 by @jagadeepmamidi, closes #690).** `_EmbeddingTrainer` already delegated `train` / `save_model` / `add_callback` / `state` but not `model` or `args`, which `train()` reads for fp16 dtype alignment.
--- a/src/soup_cli/trainer/embedding.py
+++ b/src/soup_cli/trainer/embedding.py
@@ -536,3 +536,11 @@ class _EmbeddingTrainer:
     @property
     def state(self):
         return self._trainer.state
+
+    @property
+    def model(self):
+        return self._trainer.model
+
+    @property
+    def args(self):
+        return self._trainer.args
--- a/tests/test_issue690_embedding_trainer_attrs.py
+++ b/tests/test_issue690_embedding_trainer_attrs.py
@@ -1,0 +1,180 @@
+"""#690 - `_EmbeddingTrainer` does not expose `model` or `args`.
+
+`EmbeddingTrainerWrapper.train()` reads `self.trainer.model` and
+`self.trainer.args` before calling `train()`. The wrapper stored the HF
+trainer in `_trainer` and delegated `train` / `save_model` / `add_callback` /
+`state`, but not those two attributes.
+
+`tests/test_embedding.py` stayed green because it assigns
+`wrapper.trainer = MagicMock()`, which auto-creates both. These tests use
+the real wrapper from `setup()`.
+"""
+
+from __future__ import annotations
+
+import importlib.machinery
+import json
+import os
+import sys
+import types
+
+try:
+    import torchaudio  # noqa: F401
+except ImportError:
+    pass
+except Exception:
+    _torchaudio = types.ModuleType("torchaudio")
+    _torchaudio.__version__ = "0.0.0"
+    _torchaudio.__spec__ = importlib.machinery.ModuleSpec("torchaudio", loader=None)
+    sys.modules["torchaudio"] = _torchaudio
+
+import pytest
+import yaml
+
+pytest.importorskip("torch")
+pytest.importorskip("transformers")
+pytest.importorskip("peft")
+pytest.importorskip("datasets")
+
+
+def _write_tiny_tokenizer(directory: str) -> None:
+    from tokenizers import Tokenizer, models, pre_tokenizers
+
+    vocab = {"<unk>": 0, "<s>": 1, "</s>": 2, "<pad>": 3}
+    for word in ("hello", "world", "hi", "yo", "the", "cat", "sat", "on", "mat"):
+        vocab[word] = len(vocab)
+    tokenizer = Tokenizer(models.WordLevel(vocab=vocab, unk_token="<unk>"))
+    tokenizer.pre_tokenizer = pre_tokenizers.Whitespace()
+    tokenizer.save(os.path.join(directory, "tokenizer.json"))
+    with open(os.path.join(directory, "tokenizer_config.json"), "w", encoding="utf-8") as fh:
+        json.dump(
+            {
+                "tokenizer_class": "PreTrainedTokenizerFast",
+                "unk_token": "<unk>",
+                "bos_token": "<s>",
+                "eos_token": "</s>",
+                "pad_token": "<pad>",
+                "model_max_length": 128,
+                "clean_up_tokenization_spaces": False,
+            },
+            fh,
+        )
+
+
+def _tiny_embedding_model_dir(tmp_path) -> str:
+    """Llama encoder weights for `AutoModel.from_pretrained` (not CausalLM)."""
+    import torch
+    from transformers import LlamaConfig, LlamaModel
+
+    torch.manual_seed(7)
+    config = LlamaConfig(
+        vocab_size=64,
+        hidden_size=64,
+        intermediate_size=64,
+        num_hidden_layers=2,
+        num_attention_heads=4,
+        num_key_value_heads=2,
+        max_position_embeddings=128,
+    )
+    model = LlamaModel(config).to(torch.float32).eval()
+    weights = tmp_path / "model"
+    model.save_pretrained(weights)
+    _write_tiny_tokenizer(str(weights))
+    return str(weights)
+
+
+def _embedding_wrapper(tmp_path, monkeypatch, *, precision=None):
+    from soup_cli.config.loader import load_config_from_string
+    from soup_cli.trainer.embedding import EmbeddingTrainerWrapper
+
+    if precision is not None:
+        monkeypatch.setattr(
+            "soup_cli.trainer.embedding.bf16_fp16_flags",
+            lambda device, **kwargs: precision,
+        )
+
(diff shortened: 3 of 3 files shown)


MakazhanAlpamys:
@jagadeepmamidi — welcome, and this is a very good first PR. **Approving on the substance; one process item and it merges.**

First, something you need to know that is not about your code: **#705 by @kok-o also fixes #690**, as a side effect of fixing #700. It opened at 18:43 UTC, fifty minutes after I assigned #690 to you. You could not have seen that and neither could they. I created the overlap by filing two adjacent issues and assigning one without watching the other — the same mistake produced a collision on #700 the same evening. Telling you plainly because CONTRIBUTING.md requires it and because you should not discover it from a diff.

**It does not cost you the merge.** Your PR lands first and #705 rebases onto it, and the reasons are on the merits, not on politeness:

- Yours is minimal and exactly the fix #690 prescribes: eight lines, two properties, nothing else.
- #690 is a **100% crash on the whole transformers embedding path** — I reproduced it end to end below. #700 is a narrower feature gap. An outage fix does not wait behind a feature discussion.
- #705 has two blocking findings of its own and needs rework regardless, so it must not gate yours.

## Verified by execution

Worktree at `6c5e613`, `PYTHONPATH` pinned inside it (checked via `soup_cli.__file__`), py3.10, `--no-cov`, baseline on clean `main` first (63 passed in `tests/test_embedding.py`, so no #651 noise here).

**The bug, through the real CLI** — a tiny Llama encoder, a real tokenizer, a real `soup.yaml`, and `python -m soup_cli.cli train -c soup.yaml --yes --no-reexec` on CPU:

```
main,  lora.r=4  -> exit 1   AttributeError: '_EmbeddingTrainer' object has no attribute 'model'
this,  lora.r=4  -> exit 0   LoRA applied ... Training Complete
```

**Mutations — five applied, five killed:**

| mutation | result |
|---|---|
| delete the `model` property | killed |
| delete the `args` property | killed |
| delete both (= main's state) | killed |
| **`model` returns `None`** | killed |
| **`args` returns a bare `object()`** | killed |

The last two are why I am comfortable. Deleting a property only proves the attribute must exist; returning the wrong object proves the test checks the *delegation target*. Most contributors write the first pair and stop.

**You met #690's anti-mock criterion, which is the whole point of that issue.** Your tests build a real `LlamaModel` on disk, load through the real `load_config_from_string`, call the real `EmbeddingTrainerWrapper.setup()`, and assert `isinstance(wrapper.trainer, _EmbeddingTrainer)`. `_stub_inner_train` stubs the *inner* HF trainer, not the object under test — that is the legitimate use. For context on why the issue insisted: `tests/test_embedding.py` holds 79 green tests over the path that crashes above, because a `MagicMock` standing in for the wrapper auto-creates exactly the two attributes production lacks. Your tests cannot make that mistake.

Zero test lines removed. Ruff clean. 65 passed (63 pre-existing + your 2).

## Blocking — one file

**There is no changelog fragment.** This repo does not edit `CHANGELOG.md` directly; a user-visible change adds `changelog.d/0.74.0/<PR-number>.<category>.md`. Yours would be `changelog.d/0.74.0/708.fixed.md`, mentioning `#708` in the text and crediting `by @jagadeepmamidi`. `changelog.d/README.md` has the full rules.

Worth knowing rather than guessing: **CI cannot catch this.** `tests/test_issue487_changelog_fragments.py` validates fragments that exist and has no way to notice a missing one, so it passes on your branch (17 tests). You left the checkbox unticked, which is honest — I am asking for the file rather than treating it as an oversight.

The same applies to your #709 and #710; all three need one.

## Non-blocking

**The `sys.modules["torchaudio"]` shim at the top of the test file.** It is the only occurrence in the repo, and it installs a stub `torchaudio` with `__version__ = "0.0.0"` **for the rest of the session** if `import torchaudio` raises anything that is not an `ImportError`. On this box the import succeeds, so the branch is dead here — but if it ever fires on another platform it silently poisons any later test that needs real torchaudio, and a test that poisons the session is much harder to debug than the import error it was avoiding. I would drop it, or move it into a fixture scoped to the tests that need it. Not a merge blocker.

Add the fragment and I will merge, then tell @kok-o to rebase.

jagadeepmamidi:
Changelog fragments are in:

- `changelog.d/0.74.0/708.fixed.md` (this PR)
- `changelog.d/0.74.0/709.fixed.md` (#709)
- `changelog.d/0.74.0/710.fixed.md` (#710)

Also dropped the session-wide torchaudio shim on #708 and #709 as you suggested.

MakazhanAlpamys:
Merged as `ab2412e`. Thank you — and this is a strong first PR by a margin.

**What I ran.** Worktree, `PYTHONPATH` pinned inside it and checked via `soup_cli.__file__`, py3.10, `--no-cov`, baseline on clean `main` first.

The bug and its repair at the real CLI, not at a helper — a tiny Llama encoder, a real tokenizer, a real `soup.yaml`, `python -m soup_cli.cli train -c soup.yaml --yes --no-reexec` on CPU:

```
main,  lora.r=4  -> exit 1   AttributeError: '_EmbeddingTrainer' object has no attribute 'model'
this,  lora.r=4  -> exit 0   LoRA applied ... Training Complete
```

**Five mutations, five killed.** Deleting either property, or both. And then the two that decide whether this is coverage or a presence check:

```
model property returns None          -> 2 failed
args property returns a bare object() -> 2 failed
```

Deleting a property only proves the attribute must exist. Returning the wrong object proves the test checks the delegation *target*. Most contributors write the first pair and stop; you wrote both.

**You met the criterion this issue exists for.** #690 was filed because `tests/test_embedding.py` holds 79 green tests over the path that crashes above — a `MagicMock` standing in for the wrapper auto-creates exactly the two attributes production lacked. Your tests build a real `LlamaModel` on disk, load through the real `load_config_from_string`, call the real `setup()`, and assert `isinstance(wrapper.trainer, _EmbeddingTrainer)`. Stubbing the *inner* trainer is the legitimate use, and you drew that line correctly.

You also dropped the `sys.modules["torchaudio"]` shim when I raised it, which was a non-blocking note. It would have installed a stub `torchaudio` session-wide if the real import ever raised something other than `ImportError`, and a test that poisons the session is far harder to debug than the import error it was avoiding.

Zero test lines removed. 65 passed. Ruff clean. Merged sequentially with four other PRs onto current `main` in a scratch worktree before merging: no conflicts, 350 passed.

Credited in `CONTRIBUTORS.md` (`9f1e48b`). #709 landed too; #710 is one clause and one keyword from merging.

PR #709, opened 2026-09-05
fix(trainer): put SFT packing on SFTConfig for TRL 0.29

### LINKED ISSUE #691: SFT packing is passed to SFTTrainer instead of SFTConfig
## Summary

Enabling ordinary SFT packing makes Soup pass `packing=True` directly to `SFTTrainer(...)`.
TRL 0.29.x does not accept that constructor keyword, so setup fails before training.

Soup currently declares `trl>=0.29.0,<1.0.0`; TRL 0.29 is the first supported version and rejects
the keyword. There is therefore no version in Soup's currently supported range on which this
code path works as written.

## Root cause

On current `main` (`ef2e02b51a1b34546823edce55715d28234805cb`),
`SFTTrainerWrapper.setup()` builds an `SFTConfig` through `_as_sft_config()`, but only carries
`max_length` into it. The packing branch then does:

```python
if tcfg.packing:
    trainer_kwargs["packing"] = True

self.trainer = SFTTrainer(**trainer_kwargs)
```

In TRL 0.29.0, `packing` is an `SFTConfig` field, not an `SFTTrainer.__init__` parameter. The
official documentation enables it with `SFTConfig(packing=True)`:
https://huggingface.co/docs/trl/v0.29.0/en/sft_trainer#packing

## Reproduction and impact

Binding the real installed TRL 0.29.1 `SFTTrainer` signature with Soup's keyword produces:

```text
TypeError: got an unexpected keyword argument 'packing'
```

The maintainer independently confirmed the full setup path: `packing: false` succeeds and
`packing: true` fails deterministically with this `TypeError`. This is a hard setup crash rather
than a validated-but-ignored option, affects 100% of users enabling the feature, and is a release
blocker. The regression was introduced when the dependency floor was raised for Soup 0.74.0.

## Recommended fix

Set packing on the `SFTConfig` passed as `args` before constructing the trainer, for example by
extending `_as_sft_config()` to accept and carry all SFT-only fields:

```python
training_args = self._as_sft_config(
    training_args,
    max_length=cfg.data.max_length,
    packing=tcfg.packing,
)
```

Do not merely remove the constructor keyword: that would turn the visible crash into silently
disabled packing.

The cross-document branch also currently assigns
`trainer_kwargs["packing_strategy"] = "attention_free"`. Moving that value unchanged into
`SFTConfig` is not a fix: TRL 0.29 supports `bfd`, `bfd-requeue`, and `wrapped`, and
`attention_free` is invalid. Soup must map the requested behavior to a supported strategy or
reject it clearly.

## Acceptance criteria

- [ ] `training.packing: true` constructs the real `SFTConfig` and `SFTTrainer` on the minimum
      supported TRL version.
- [ ] `trainer.args` is an `SFTConfig` with `packing is True`.
- [ ] A test exercises ordinary packing with the real `SFTConfig` rather than a permissive mock.
- [ ] A separate test exercises the cross-document option and verifies a valid supported strategy
      (`bfd`, `bfd-requeue`, or `wrapped`) or a clear validation error.
- [ ] An integration test verifies the resulting dataset/batch is actually packed.
- [ ] `packing: false` remains behaviorally unchanged.

## Duplicate check

This is ordinary SFT packing, even with cross-document masking disabled. #47 concerns
cross-document attention boundaries; #65 and #561 concern Soup's separate multipack sampler.
Searches across open/closed issues and PRs found no ticket for the invalid TRL 0.29 constructor
keyword and no open PR addressing it at filing time.

jagadeepmamidi:
TRL 0.29 moved packing onto SFTConfig and dropped kwargs on SFTTrainer, so packing=True on the trainer is a TypeError. attention_free is not in the packing_strategy allowlist.

## What does this PR do?

TRL 0.29 moved `packing` onto `SFTConfig`, and `SFTTrainer.__init__` has no `**kwargs`. Soup still set `trainer_kwargs["packing"] = True`, which is a `TypeError` on every supported trl. `packing_strategy="attention_free"` is also invalid (allowlist is `bfd` / `bfd-requeue` / `wrapped`).

This carries `packing` through `_as_sft_config` and schema-rejects `packing_cross_doc_attn_mask`. The regression tests call real `setup()` — `tests/test_packing.py` stayed green because `TestPackingSFTTrainerMock` copies those kwargs onto a `MagicMock`.

Closes #691.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist


MakazhanAlpamys:
@jagadeepmamidi — **the mechanism is right, the tests are strong (4/4 mutations killed), and I am accepting the cross-doc rejection you were probably least sure about. Blocking on three things, one of which is a crash your fix leaves standing one file away.**

I asked a reviewer to attack specifically the part where your diff does two things instead of one, because turning `packing_cross_doc_attn_mask` into an unconditional rejection removes a documented v0.28.0 feature and that is more than #691 literally asks for. The answer came back in your favour, and with better evidence than your PR gives itself. Details below, because you should be able to defend this if someone reopens it.

## Verified by execution

Worktree at `cbe4f92`, `PYTHONPATH` pinned inside it, py3.10, `--no-cov`. Stack is the declared floor: trl 0.29.1 / transformers 5.16.1 / torch 2.5.1+cu121.

| run | result |
|---|---|
| your new file + speed / packing / config / multipack / neat_packing / examples_configs | **218 passed, 1 skipped** |
| same six files on clean `origin/main` | 214 passed, 1 skipped |
| ruff | clean |

The delta is exactly your four tests. No regressions.

| mutation | result |
|---|---|
| drop `packing=tcfg.packing` from the `_as_sft_config` call site | killed |
| `_as_sft_config` silently ignores the kwarg at both assignment sites | killed |
| invert the new validator (`if False:`) | killed — `DID NOT RAISE` |
| regress to the old `trainer_kwargs["packing"] = True` | killed |

**You met #691's hardest acceptance criterion.** `test_setup_packing_true_lands_on_sft_config` builds a real `LlamaForCausalLM` and a real tokenizer, goes through `load_config_from_string` -> `SFTTrainerWrapper(cfg).setup()`, and asserts `isinstance(args, SFTConfig)` with `args.packing is True`. No `MagicMock`. That matters here more than usual — `tests/test_embedding.py` currently holds 79 green tests over a path that crashes in production because a mock auto-created the missing attributes. And the path is CLI-reachable: `commands/train.py:1474-1475` is literally `SFTTrainerWrapper(cfg, **kw)` then `.setup(dataset)`.

## The cross-doc rejection: accepted, and here is the evidence you should have cited

I checked the three things that would have made me ask you to split it out, and all three came back clean:

1. **`attention_free` was never a valid TRL strategy — in any version.** Grepped nine wheels (0.7.11 through 0.29.1): zero occurrences of the string in every one. The allowlist is enforced at `trl/data_utils.py:876`. And `packing_strategy` was never an `SFTTrainer.__init__` parameter either, with no `**kwargs` to absorb it — I confirmed `sig.bind(..., packing=True)` raises `TypeError` on the installed 0.29.1. **So this path has been a guaranteed crash since it was introduced in v0.28.0, not something the v0.74.0 floor broke.** Nobody has ever had a working cross-doc run to lose.
2. **`CrossDocCollator` could not be wired up as an alternative even if we wanted to.** `sft_trainer.py:798` sets `padding_free = args.packing and args.packing_strategy == "bfd"`, and `:801` raises on any custom `data_collator` when `padding_free` is on. Separately, the collator consumes a `doc_lengths` field produced by nothing in `src/` and writes `cross_doc_attn_mask`, a key with zero readers in `src/`, in trl, or in transformers.
3. **No shipped recipe, template or example sets it.** Only one `true` exists in the repo, and it is in the docs — see blocker 2.

**Splitting it out is not even viable**, which is the argument I would lead with: `packing` and `packing_strategy` go into the *same* `SFTTrainer(**trainer_kwargs)` splat, so fixing only `packing` would leave `packing: true` + `packing_cross_doc_attn_mask: true` raising `TypeError` one line below the line you just fixed.

---

## BLOCKING 1 — no changelog fragment

`changelog.d/0.74.0/709.fixed.md`, named after the **PR** number, mentioning `#709` in the text, credited `by @jagadeepmamidi`. Same for #708 and #710.

CI cannot catch this: `tests/test_issue487_changelog_fragments.py` validates fragments that exist and has no way to see a missing one.

## BLOCKING 2 — the docs still document the feature and hand out YAML that now fails to parse

`docs/performance-and-quantization.md` has a whole section titled "Cross-Document Attention Masking" ending in a copyable block:

```yaml
training:
  packing: true
  packing_cross_doc_attn_mask: true
```

That is now a hard `ValidationError`, and the section's closing line — "Misconfiguring it without `packing: true` is rejected at config-load time" — is now false in a way that describes the *old* rule. `docs/peft-and-efficiency.md:107` also still describes it composing with multipack. Removing a feature has to remove its documentation, or the next user finds the docs before they find your error message.

## BLOCKING 3 — `trainer/pretrain.py:220` is the identical bug, and your PR leaves it crashing

```python
# src/soup_cli/trainer/pretrain.py:219-221
if tcfg.packing:
    trainer_kwargs["packing"] = True
```

...then `SFTTrainer(**trainer_kwargs)` at `:250`. I confirmed both halves myself:

```
task: pretrain + packing: true  -> parses fine (training.packing = True)
SFTTrainer.__init__             -> no `packing` param, no **kwargs
sig.bind(..., packing=True)     -> TypeError: got an unexpected keyword argument 'packing'
```

So after this PR merges, `task: sft` + `packing: true` works and `task: pretrain` + `packing: true` still hard-crashes at setup, with the fix sitting one file away. `pretrain.py` also builds a bare `TrainingArguments`, so it never gets the `max_length` carry-over your `_as_sft_config` adds.

Two ways, your call — say which and I will not push back:
- **extend this PR** to `pretrain.py`, mirroring what you already wrote (small, and it is the same defect);
- **or say so here**, and I will file it as its own issue before release with your finding credited.

I would rather it were in this PR, because a half-fixed defect is the shape that gets forgotten.

---

## Non-blocking

1. **The new field description says "Rejected on TRL 0.29".** It was never accepted on *any* TRL, and saying "0.29" invites someone to pin an older trl to get the feature back — which would give them a `TypeError`, not the feature. Please say it never worked.
2. **The error message names only what Soup will not do.** Worth adding what to do instead: TRL's default `bfd` strategy plus a FlashAttention `attn_implementation` already isolates documents — `sft_trainer.py:867` warns about cross-contamination precisely when FlashAttention is *absent*. So `packing: true` with FA is the capability the removed flag was promising, and the message should point there.
3. **`test_requires_packing` is now vacuous.** You inverted `test_enabled` correctly, but left `test_requires_packing` in place; since the flag is now rejected unconditionally, that test passes without discriminating anything and its docstring describes a rule that no longer exists. Delete it or repurpose it.
4. `test_packing.py::TestPackingTrainerIntegration::test_sft_trainer_receives_packing` never constructs a trainer — it only asserts `cfg.training.packing is True`. Your own new file's docstring calls this out, which is good, but the misleading test stays in the tree. Same class of thing as item 3.
5. `test_setup_packing_false_unchanged` asserts `args.packing is not True` where `is False` is available and stronger.

**All eight removed test lines audited** and none is a weakening: five are the `SoupConfig(...)` call plus assertion in `test_enabled`, correctly inverted to `pytest.raises(match="bfd")`, and the rest are the `packing_cross_doc_attn_mask: True` key and its assertion in `TestV028Integration`, which had to go.

Good first PR. Fix the three blockers and I will merge.

MakazhanAlpamys:
Thanks — `changelog.d/0.74.0/709.fixed.md` is correct: PR number in the filename and the text, and credited. That closes blocker 1.

**Blockers 2 and 3 are still open**, and they are the two that matter more:

**2. The docs still document the feature you removed.** `docs/performance-and-quantization.md` has a section titled "Cross-Document Attention Masking" ending in a copyable block that is now a hard `ValidationError`:

```yaml
training:
  packing: true
  packing_cross_doc_attn_mask: true
```

and its closing sentence — "Misconfiguring it without `packing: true` is rejected at config-load time" — now describes the old rule. `docs/peft-and-efficiency.md:107` also still has it composing with multipack.

**3. `trainer/pretrain.py:220` still crashes.** I confirmed both halves myself:

```
task: pretrain + packing: true   -> parses fine (training.packing = True)
SFTTrainer.__init__              -> no `packing` param, no **kwargs
sig.bind(..., packing=True)      -> TypeError: got an unexpected keyword argument 'packing'
```

So when this merges, `task: sft` + `packing: true` works and `task: pretrain` + `packing: true` still hard-crashes at setup, with your fix one file away. `pretrain.py` also builds a bare `TrainingArguments`, so it never gets the `max_length` carry-over either.

Still your call between extending this PR and having me file it — say either and I will not push back. I would prefer it here, because a half-fixed defect is the shape that gets forgotten.

The three non-blocking items from my review also still stand, and the first is nearly free while you are in the file: the field description says "Rejected on TRL 0.29" when it was never accepted on **any** TRL — I grepped nine wheels from 0.7.11 to 0.29.1 and `attention_free` appears in zero of them. As written it invites someone to pin an older trl to get the feature back, which would give them a `TypeError`.

jagadeepmamidi:
Took the extend-this-PR option for pretrain.

- pretrain.py now carries packing through SFTTrainerWrapper._as_sft_config (same as SFT). Real setup() test asserts args.packing is True on a SFTConfig.
- Cross-doc docs/YAML in docs/performance-and-quantization.md and docs/peft-and-efficiency.md are gone. Field description and the validator now say the flag never worked, and point at packing: true plus FlashAttention.
- Deleted vacuous test_requires_packing. Changelog fragment updated.

MakazhanAlpamys:
Merged as `991ab8f`. Thank you — and the part I want to name is that you took the harder option when offered the easier one.

**Four mutations, four killed** on the SFT half: dropping `packing=` from the `_as_sft_config` call, making `_as_sft_config` swallow the kwarg at both assignment sites, inverting the new validator, and regressing to the old `trainer_kwargs["packing"] = True`.

**And the pretrain half is pinned too.** When I pointed out that `trainer/pretrain.py:220` carried the identical crash one file away, you could have said "file it separately" — I offered exactly that and said I would not push back. You extended the PR instead, and the mutation confirms it is real:

```
MUT pretrain packing never reaches SFTConfig -> 1 failed
```

Without it, `task: sft` + `packing: true` would have worked while `task: pretrain` + `packing: true` still hard-crashed, with the fix visible one file over. That is the shape of defect that gets forgotten.

**On the cross-doc rejection, which was the part I scrutinised hardest.** You removed a documented v0.28.0 option, and I went looking for a reason to make you split it out. The evidence went your way and it is worth having in the thread: `attention_free` appears in **zero of nine trl wheels** from 0.7.11 to 0.29.1, `packing_strategy` was never an `SFTTrainer.__init__` parameter either, and `CrossDocCollator` could not have been substituted — `sft_trainer.py:798` turns on `padding_free` whenever packing uses `bfd`, and `:801` then raises on any custom `data_collator`. So the path was a guaranteed `TypeError` from the day it shipped and nobody had a working run to lose. Splitting it out was not even viable: `packing` and `packing_strategy` go into the same `**trainer_kwargs` splat.

You also corrected the field description to say it never worked on **any** released trl rather than naming 0.29 — which matters, because a version number there invites someone to pin an older trl to get the feature back and receive a `TypeError` for it.

All eight removed test lines audited, none a weakening. Zero tests deleted wholesale. #691's hardest criterion met with a real `LlamaForCausalLM` and a real `SFTConfig`, no `MagicMock`.

Credited in `CONTRIBUTORS.md` (`9f1e48b`).

Issue #689, opened 2026-09-05
HF Hub data.streaming is ignored and the dataset is materialized eagerly


Amix29:
## Description

For Hugging Face Hub datasets, `data.streaming: true` is not forwarded consistently to
`datasets.load_dataset()`. The single-Hub path then iterates the entire training split into a
Python list before returning. The Hub interleave helper also ignores the streaming request and
buffer setting.

Affected revision: `ef2e02b51a1b34546823edce55715d28234805cb` (Soup 0.74.0).

Relevant code:

- [schema describes a streaming pass-through](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/config/schema.py#L519-L531)
- [single-Hub loader calls `hf_load(name)` and materializes the split](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/data/loader.py#L916-L955)

At the confirmed revision, `streaming` is read only by the local/remote interleave path and the
fsspec path; `buffer_size` is likewise used only by other loader paths. Neither
`_load_one_hub_dataset()` nor `_load_interleaved_hub_datasets()` consumes these settings.

## Steps to reproduce

Set `data.streaming: true` and `data.buffer_size: 1000`, replace the Hub API with an instrumented
iterable containing 512 synthetic rows, and execute the real `_load_one_hub_dataset()` path.

```text
kwargs received by datasets.load_dataset: {}
rows consumed before the function returns: 512
return type: list
```

The maintainer independently confirmed these exact results on revision `2edc063`.

## Expected behavior

A Hub dataset requested in streaming mode should pass `streaming=True` to
`datasets.load_dataset()` and remain lazy and bounded through setup and training. If Soup cannot
support that contract for a given combination, it should reject it instead of accepting and
ignoring the setting.

## Impact and conditions

Large datasets can be downloaded or materialized unexpectedly, increasing startup latency and
causing host-memory exhaustion. The impact is conditional on dataset size; no real Hub download
or peak-RAM measurement is claimed. Both single and interleaved Hub dataset paths are in scope.

## Proposed fix and acceptance criteria

Forwarding `streaming=True` alone is insufficient if downstream code converts the iterable back
to a list. Define and preserve an end-to-end lazy-data contract.

- [ ] Tests assert the exact kwargs passed to `datasets.load_dataset()`, including
[...]

PR #710, opened 2026-09-05
fix(data): honor streaming and buffer_size on Hub dataset names

jagadeepmamidi:
## What does this PR do?

The schema advertises `data.streaming` as a pass-through to HF `streaming=True`. `_load_one_hub_dataset` still did `hf_load(name)` with no kwargs, then list-comprehended the split. Remote and local-interleave streaming already forwarded the flag.

```python
# src/soup_cli/data/loader.py L959–962
if data_config.streaming:
    ds = hf_load(name, streaming=True)
else:
    ds = hf_load(name)
```

```python
# src/soup_cli/data/loader.py L916–937
# streaming: optional shuffle(buffer_size=) on train, then materialise
# up to MAX_REMOTE_ROWS. non-streaming: list-comprehend, no cap.
```

All-hub `data.interleave` + `streaming: true` stays schema-rejected (#459). Tests spy the kwargs `datasets.load_dataset` actually receives — a row-count check would stay green if the flag were forwarded and then the source materialised anyway.

Refs #689.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist


MakazhanAlpamys:
@jagadeepmamidi — **correct fix, well-tested, CLI-reachable. One blocking item, and one framing decision I am making rather than asking you to guess.**

## Verified by execution

Worktree at `7801b4b`, `PYTHONPATH` pinned inside it, py3.10, `--no-cov`, baseline on clean `origin/main` first.

| run | result |
|---|---|
| `test_loader` + `test_dataset_hub` on main | 32 passed |
| same plus your new file on this branch | **37 passed** |
| ruff | clean |

Zero test lines removed. Three source lines removed, all replaced by `_rows_from_hub_split`.

**The defect and its repair, at the CLI** — real `soup train --config soup.yaml --dry-run --yes` with `data.streaming: true`, `buffer_size: 500`, and a Hub dataset name:

```
main:  kwargs={}                      shuffle_calls=None
this:  kwargs={'streaming': True}     shuffle_calls=[500]
```

| mutation | result |
|---|---|
| drop the `streaming=True` kwarg | killed — 2 failed |
| **forward `streaming=True` but revert the materialisation to the old list comprehension** | killed — 2 failed |
| ignore `buffer_size` (no shuffle) | killed — 1 failed |
| remove the `MAX_REMOTE_ROWS` cap | killed — 1 failed |

The second one is the mutation I care about most, because it is the difference between a real test and a mock-shaped one. Your tests are **not** merely asserting "the kwarg was passed": with the flag still forwarded and only the behaviour faked, two of them fail. Shuffle-buffer application and the cap are both verified as effects, not as arguments.

**Your test double is honest.** A hand-rolled `type(sys)("datasets")` module plus a real `_FakeStream` class, so a missing attribute raises instead of being auto-created. That distinction is not academic here — `tests/test_embedding.py` has 79 green tests over a path that crashes in production, because a `MagicMock` standing in for the object under test invents whatever is asked of it.

**The bound is real**, not just asserted: against a genuinely infinite generator source the loader terminates at exactly the cap (consumed 5001, kept 5000 at cap 5000).

**The interleave path is covered too**, which #689 asks for: `_load_interleaved_hub_datasets` delegates to `_load_one_hub_dataset` so it is fixed transitively, and an all-hub list with `streaming: true` is refused at `schema.py:5635` — a refusal that is pre-existing on `main`, not something you added, and which #689 explicitly sanctions.

---

## BLOCKING — no changelog fragment

`changelog.d/0.74.0/710.fixed.md`, named after the **PR** number, mentioning `#710` in the text, credited `by @jagadeepmamidi`. `changelog.d/README.md` has the rules; `CHANGELOG.md` is never edited directly.

CI cannot catch this — `tests/test_issue487_changelog_fragments.py` validates fragments that exist and cannot see a missing one. You left the checkbox unticked, so this is acknowledged rather than hidden, but see the next item, which is why it matters more here than on #708 or #709.

**The fragment must disclose the truncation.** A Hub dataset with more than 1,000,000 rows and `streaming: true` previously loaded *all* of them, because the flag was ignored and the whole thing was downloaded. It is now cut to 1M with a warning. Measured with the cap lowered to 1000 against a 5000-row source: `main` returns 5000, this branch returns 1000. That is defensible — it matches `_load_remote_dataset`, it is warned, and you documented it in `docs/data.md` — but a user whose training set silently shrinks needs to find it in the changelog, not in a warning line they scrolled past.

## My call: change `Closes #689` to `Refs #689`, and I will keep the issue open

This is not a criticism of the PR. What you shipped is *bounded-eager*: the flag is forwarded, `buffer_size` is applied, and the source is consumed up to a cap. #689 asks for something slightly larger — a lazy contract. Against its checklist:

- exact-kwargs test — **done**
- both functions — **done** (one by refusal, which the issue allows)
- `buffer_size` defined — **done**
- "setup does not consume or materialize the full source" — **partial**: any dataset at or under 1M rows is still fully materialised
- "MLX and Transformers consumers do not silently rematerialize the stream" — **not addressed**, and moot while the loader returns a list
- "regression test uses an effectively unbounded/counting source" — **not done**

v0.73.3 set the precedent for exactly this shape: PRs that address an adjacent item merge with `Refs`, and the issue stays open (#371 and #394 both did). So `Refs #689`, this merges on its own merits, and #689 stays open scoped down to the lazy contract. If you want that follow-up it is yours; if not I will label it and leave it.

## Non-blocking, and the first one is nearly free

1. **#689 asks for a test using "an effectively unbounded/counting source".** Your cap test instead uses a 10-row source with `MAX_REMOTE_ROWS` monkeypatched to 3. I ran the genuinely unbounded version against your branch and it terminates correctly, so this is about ten lines to close that checkbox honestly — and it would move the fourth bullet above from "partial" to "measured".
2. `buffer_size` shuffles the train split only (`shuffle=False` for validation). Sensible, but a capped streamed validation split then takes the first N rows unshuffled, which can be a biased sample of a sorted source. A line in `docs/data.md` at most.

Add the fragment with the truncation disclosed, switch to `Refs #689`, and I will merge. If you want to take the ten-line unbounded test as well, say so and I will wait for it.

MakazhanAlpamys:
Thanks — `changelog.d/0.74.0/710.fixed.md` is correct on filename, PR number and credit, and "materialises up to `MAX_REMOTE_ROWS`" does disclose the cap. Two things left, both small.

**1. The fragment states the cap but not the change.** A reader who has a Hub dataset over 1,000,000 rows and `streaming: true` needs to know their training set is about to get smaller, not just that a cap exists. Before this PR the flag was ignored and all rows were downloaded; after it they are cut to 1M with a warning. Measured with the cap lowered to 1000 against a 5000-row source: `main` returns 5000, this branch returns 1000. One clause — "previously the flag was ignored and the full dataset was materialised" — makes it findable by someone diagnosing a shrunk dataset.

**2. `closes #689` -> `Refs #689`,** in the PR body and the fragment. Not a criticism of the fix. What you shipped is bounded-eager; #689 asks for a lazy contract, and three of its criteria are not met by this:

- "setup does not consume or materialize the full source" — partial: any dataset at or under 1M rows is still fully materialised
- "MLX and Transformers consumers do not silently rematerialize the stream" — not addressed, and moot while the loader returns a list
- "regression test uses an effectively unbounded/counting source" — your cap test uses a 10-row source with `MAX_REMOTE_ROWS` patched to 3

This repo's precedent for exactly that shape is `Refs` with the issue left open (v0.73.3 did it for #371 and #394). So this merges on its own merits and #689 stays open, scoped down to the lazy contract.

That third bullet is also the cheapest thing on the list: I ran the genuinely unbounded version against your branch and it terminates correctly at the cap, so it is about ten lines and it moves the first bullet from "partial" to "measured". Optional — say if you would rather not and I will merge without it.

jagadeepmamidi:
Fragment now says the flag used to materialise the full dataset and that sources over 1M rows are cut to 1M with a warning. PR body and fragment use Refs #689.

Also added the unbounded counting-source cap test (consumed cap+1, kept cap) and a line in docs/data.md that validation is first-N unshuffled.

PR #711, opened 2026-09-05
fix(distill): ULD loss respects the response-only label mask and causal shift

### LINKED ISSUE #682: ULD loss ignores the response-only label mask and causal shift
## Description

The SFT cross-entropy respects causal shifting and `labels == -100`, but the ULD term receives
only `attention_mask`. It therefore optimizes prompt or other positions explicitly excluded by
Soup's supervision mask, and also includes positions outside the shifted causal targets.

Relevant code:

- [cross-entropy uses shifted labels and `ignore_index=-100`](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/trainer/distill.py#L505-L517)
- [ULD receives only `attention_mask`](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/trainer/distill.py#L635-L642)
- [`uld_distill_loss` masks only padding/attention positions](https://github.com/MakazhanAlpamys/Soup/blob/ef2e02b51a1b34546823edce55715d28234805cb/src/soup_cli/utils/uld.py#L228-L281)

## Steps to reproduce

Run the original loss functions with real FP32 tensors and:

```text
labels = [-100, -100, 2]
```

The logit at position 0 predicts a target excluded by the causal label mask.

```text
gradient L1 at position 0, CE-only control: 0.0
gradient L1 at position 0, current CE + ULD: ~0.075376
```

## Expected behavior

When `train_on_responses_only` or another label mask excludes a target, every configured loss
term should follow the same causal supervision boundary unless a separate behavior is explicitly
configured and documented.

## Impact

The optimized objective differs from the requested one. Prompt tokens and other excluded spans
can influence adapter updates even though the SFT term correctly ignores them.

Build an explicit causal supervision mask from shifted labels and propagate it through ULD,
including the aligned-tokenizer mapping.

- [ ] Prompt-only positions contribute zero ULD gradient in responses-only mode.
- [ ] Padding, multi-turn conversations, and mixed masked/unmasked batches are covered.
- [ ] Entirely masked batches remain finite and produce no spurious update.
- [ ] The aligned ULD strategy maps the supervision mask through token alignment.

## Environment

Reproduced under Python 3.12.2 and PyTorch 2.12.1 CPU with the original Soup loss code and
autograd. No full CUDA training or final-quality comparison was performed.

AmirF194:
Threaded the response-only label mask through the ULD branch, plus the same causal shift the CE term already applies.

`uld_distill_loss` and `uld_aligned_loss` now take an optional `labels` argument. When given, they shift logits/labels the same way `_compute_distill_term` does (position i predicts token i+1, so the last position gets dropped) and mask on `labels != -100` instead of `attention_mask`. Both `compute_loss` call sites now pass `labels`. Attention-mask-only masking still works for callers with no labels, and no labels or mask at all keeps the old unmasked behavior, so this doesn't touch the non-distill paths.

For the aligned strategy: `s_strings`/`t_strings` are decoded from the unshifted sequence, so position i in them is raw position i, and the shift lines up with the alignment those strings already produced.

Added `tests/test_issue682_uld_loss_label_mask.py`. The key trick: perturbing a whole position vector by a constant is a softmax no-op, so the tests bump a single vocab logit instead, then check that only a supervised position moves the loss (checked for all three strategies, plus the final-position drop and a fully-masked batch staying finite at zero). Ran the existing ULD/distill suites (`test_v07111.py`, `test_v07118.py`, `test_v0700_part_b.py`) too, nothing broke.

I did not run the full `DistillTrainerWrapper.compute_loss` path with a real model (would need transformers/peft/datasets, which I left out of this scoped run), and I have not matched this numerically against the gradient trace you posted on GPU. The shift-and-mask math itself is the same as `_compute_distill_term`'s, which the wider suite already exercises end to end.

Fixes #682

### DIFF
--- a/changelog.d/0.74.0/711.fixed.md
+++ b/changelog.d/0.74.0/711.fixed.md
@@ -1,0 +1,7 @@
+- The ULD distillation loss (`wasserstein` / `topk_align` / `wasserstein_aligned`)
+  ignored the response-only label mask and the causal shift the CE term already
+  applies, so under the default `train_on_responses_only` it optimized prompt
+  tokens and the shift-boundary position too (#682 by @MakazhanAlpamys, closes
+  #682 in #711). `uld_distill_loss` and `uld_aligned_loss` now take an optional
+  `labels` argument and mask on `labels != -100` after the same shift
+  `_compute_distill_term` uses.
--- a/changelog.d/0.74.0/711.fixed.md
+++ b/changelog.d/0.74.0/711.fixed.md
@@ -1,7 +1,7 @@
 - The ULD distillation loss (`wasserstein` / `topk_align` / `wasserstein_aligned`)
   ignored the response-only label mask and the causal shift the CE term already
   applies, so under the default `train_on_responses_only` it optimized prompt
-  tokens and the shift-boundary position too (#682 by @MakazhanAlpamys, closes
-  #682 in #711). `uld_distill_loss` and `uld_aligned_loss` now take an optional
+  tokens and the shift-boundary position too (#711 by @AmirF194, closes
+  #682). `uld_distill_loss` and `uld_aligned_loss` now take an optional
   `labels` argument and mask on `labels != -100` after the same shift
   `_compute_distill_term` uses.
--- a/src/soup_cli/trainer/distill.py
+++ b/src/soup_cli/trainer/distill.py
@@ -599,6 +599,7 @@ class DistillTrainerWrapper:
                         t_strings,
                         config=_uld_projection.config,
                         attention_mask=s_mask,
+                        labels=labels,
                     )
                     total = _CE_WEIGHT * ce_loss + _DISTILL_WEIGHT * distill_loss
                     return (total, outputs) if return_outputs else total
@@ -635,10 +636,13 @@ class DistillTrainerWrapper:
                 anchor = None
                 if _uld_projection is not None:
                     # v0.71.11 #236 — cross-tokenizer ULD distillation loss.
+                    # #682: pass labels so masking matches the CE term,
+                    # trained tokens only, not every attended position.
                     distill_loss = _uld_projection(
                         student_logits,
                         teacher_logits,
                         attention_mask=inputs.get("attention_mask"),
+                        labels=labels,
                     )
                 elif _minillm_cb is not None:
                     # v0.71.11 #237 — MiniLLM teacher-mixed reverse-KL +
--- a/src/soup_cli/utils/uld.py
+++ b/src/soup_cli/utils/uld.py
@@ -231,6 +231,7 @@ def uld_distill_loss(
     *,
     config: ULDConfig,
     attention_mask=None,
+    labels=None,
 ):
     """Cross-tokenizer ULD distillation loss (v0.71.11 #236).
 
@@ -247,7 +248,13 @@ def uld_distill_loss(
       Distils only on the high-probability subset.
 
     Differentiable w.r.t. the student logits (sort / topk both use
-    gather). Returns a scalar mean over the (masked) token positions.
+    gather). Returns a scalar mean over the (masked) token positions,
+    restricted to the trained tokens the same way
+    :func:`soup_cli.trainer.distill._compute_distill_term` restricts the
+    plain-KL path: ``labels != -100`` (post causal shift) when ``labels``
+    is given, else ``attention_mask``, else every position (#682: a
+    pure-padding mask let prompt tokens and the shift-boundary position
+    leak into the loss).
     """
     import torch
 
@@ -255,6 +262,18 @@ def uld_distill_loss(
         raise TypeError(
             f"config must be ULDConfig, got {type(config).__name__}"
         )
+
+    # Causal-LM alignment: logit position i predicts token i+1, the same
+    # shift the CE term applies. Drop the final logit and shift
+    # labels/attention_mask the same way so the mask below lines up.
+    if labels is not None or attention_mask is not None:
+        student_logits = student_logits[:, :-1, :]
+        teacher_logits = teacher_logits[:, :-1, :]
+        if labels is not None:
+            labels = labels[:, 1:]
+        if attention_mask is not None:
+            attention_mask = attention_mask[:, 1:]
+
     p_s = torch.softmax(student_logits, dim=-1)
     p_t = torch.softmax(teacher_logits, dim=-1)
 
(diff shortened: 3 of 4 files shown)


MakazhanAlpamys:
@AmirF194 — **the engineering is right and the tests are the best-discriminating set in this batch. Blocking on one line, and it is a line that credits me for your work.**

## The blocker

`changelog.d/0.74.0/711.fixed.md` reads:

```
(#682 by @MakazhanAlpamys, closes #682 in #711)
```

I am the repo owner. This PR is yours and issue #682 is @Amix29's; neither of us is the right name there. It should be:

```
(#711 by @AmirF194, closes #682)
```

The filename and the in-text `#711` are already right — only the credit is wrong. I would let a formatting slip through, but not this one: attribution is the only currency this project pays contributors in, and a fragment that names the maintainer for outside work is the exact failure the release checklist spends a paragraph on. Fixing it is worth a round.

## Verified by execution

Worktree at `29aa8b2`, `PYTHONPATH` pinned inside it, py3.10, `--no-cov`, baseline on clean `main` first (349 passed in the same files — no #651 fallout in this scope).

- New suite + four pre-existing ULD/distill suites on the branch: **329 passed, 1 skipped**.
- The new suite copied onto unfixed `main`: **7 of 9 fail**.

**Mutations — five applied, five killed:**

| mutation | result |
|---|---|
| drop the label mask, keep the shift (non-aligned) | 3 failed — killed |
| remove the causal shift entirely (non-aligned) | 3 failed — killed |
| **off-by-one: `labels[:, 1:]` -> `labels[:, :-1]`** | 2 failed — killed |
| drop the label mask (aligned) | 2 failed — killed |
| remove the causal shift (aligned) | 1 failed — killed |

**The off-by-one is the one I expected to survive and it does not.** A shift test is worthless unless the fixture makes shifted and unshifted give different numbers, and most do not. `test_final_position_excluded_by_causal_shift` uses `labels=[[-100,5,2]]`, where unshifted masking leaves the final position supervised and shifted masking drops it. That fixture is discriminating by construction, and your docstring says so — including the note about perturbing a single vocab entry rather than a whole position vector to avoid the softmax-invariance trap. That is a methodological point most contributors never write down.

**Numerically correct, not merely different.** Recomputed independently — W1 as `sum|cumsum(a) - cumsum(b)|` over sorted softmaxes with the CE shift applied by hand:

```
independent expected = 0.94520140
uld_distill_loss     = 0.94520140     match
```

And the property #682 is actually about:

```
main:  per-position grad L1 = [0.343096, 0.442297, 0.406747]   prompt pos 0 = 0.343096
this:  per-position grad L1 = [0.0, 0.587038, 0.653393, 0.0]   prompt pos 0 = 0.0 exactly
```

**CLI-reachable**, traced rather than grepped: `commands/train.py:1432` -> `DistillTrainerWrapper` -> `.setup()` at `:1475` -> `compute_loss`, where `labels = inputs.get("labels")` is the same tensor CE uses, and both ULD call sites now forward it.

**One detail you got right that would have been invisible if you had not:** in `uld_aligned_loss` the teacher logits are deliberately *not* shifted, because they live in teacher token space and are aggregated onto student positions by `align` — while in `uld_distill_loss` both are shifted, since they share a token space. Getting that backwards produces a plausible number and a wrong gradient.

Zero test lines removed. Ruff clean. CI green on all nine cells.

## Non-blocking, but I would take the first one while you are pushing

1. **`test_source_forwards_labels_to_both_uld_call_sites` reads `distill.py` as a string and greps it for `labels=labels`.** It fires on spelling, not behaviour: it breaks on reformatting and it cannot tell a live call site from one in a dead branch. Your tensor tests already prove the real thing — including the mutation where I deleted the call-site kwarg — so this one adds brittleness without adding coverage. I would delete it. (v0.73.3 removed a test of exactly this shape for exactly this reason.)
2. **Two of #682's acceptance criteria are not pinned:** mixed masked/unmasked batches, and multi-turn. Every test uses `batch=1`. I checked by hand that `batch=2` with one fully-masked row and one two-span multi-turn row behaves correctly — gradient lands only on supervised, shifted positions and the masked row gets exactly zero — so this is a test gap, not a defect. Your call whether to close it here.
3. The helper now also shifts when only `attention_mask` is passed, which is deliberate parity with `_compute_distill_term` and has no other in-tree caller. Noting it because the pre-existing `test_attention_mask_applied` only asserts `isfinite` and could never have noticed either way.

## Landing order

**#704 goes first, then this.** Not a rebase requirement — I trial-merged both orders and each is clean, and the two suites plus every pre-existing ULD/distill file pass on the merged tree (264 passed). The reason is conceptual: `uld_distill_loss` shifts *both* student and teacher logits, which is only meaningful if teacher position i corresponds to student position i — that is, if the teacher was fed the student's own ids in the same token space. #704 is what enforces exactly that, and the clamp it deletes was silently violating it. Landing #704 first means this PR's shift assumption is never true only by accident.

Together they close the ULD path properly: #704 guarantees the teacher is conditioned on the right text, this one guarantees the divergence is measured at the right positions.

Fix the credit line and I will merge both.

AmirF194:
Pushed 0749c23: the fragment now reads "#711 by @AmirF194, closes #682".

MakazhanAlpamys:
Merged as `c5c0f32`, after #704 as planned. Thank you — this is the best-designed test fixture in the current queue and I want to say why.

**Five mutations, five killed.** The one I expected to survive is the off-by-one, and I re-ran it myself on your final head:

```
MUT labels[:, 1:] -> labels[:, :-1]  (both sites)  -> 3 failed
```

A causal-shift test is worthless unless its fixture makes shifted and unshifted produce different numbers, and most fixtures do not. `labels=[[-100, 5, 2]]` is discriminating **by construction** — unshifted masking leaves the final position supervised where shifted masking drops it. You also wrote down, in the docstring, why you perturb a single vocab entry rather than a whole position vector, which is the softmax-invariance trap most people fall into without noticing. That is a methodological point I almost never see written.

**Numerically correct, not merely different.** Recomputed independently, W1 as `sum|cumsum(a) - cumsum(b)|` over sorted softmaxes with the CE shift applied by hand: `0.94520140` both ways. And the property #682 is actually about — per-position gradient L1 at the prompt position, `0.343096` on `main` to **exactly 0.0** here.

**The asymmetry you got right.** In `uld_aligned_loss` the teacher logits are deliberately not shifted, because they live in teacher token space and are aggregated onto student positions by `align`; in `uld_distill_loss` both are shifted because they share a token space. Getting that backwards yields a plausible number and a wrong gradient, and nothing would have told you. It is recorded in the merge commit and on #682 so it survives.

The changelog credit is right now. That was worth a round: a fragment naming the maintainer for outside work is the cheapest possible way for this project to lose the people doing the work.

Zero test lines removed. 104 passed on the #704 + #711 tree against current `main`.

Credited in `CONTRIBUTORS.md` (`904553d`). #696 has one pending check left; #698 needs its rebase and credit line once that lands.

AmirF194:
Good catch on the credit line, and thanks for re-running the off-by-one mutation yourself to double check it.

PR #712, opened 2026-09-05
fix(data): make `soup data validate` agree with the loader by running the converters

### LINKED ISSUE #695: `soup data validate` green-lights rows that `load_dataset` drops — validator never runs the converters
Split out of #676 as suggested there ("split it out if it turns out not to [share a helper]"). It does not share a helper — the validator is an entirely separate code path from the converters.

## What happens

`soup data validate --format <fmt>` and `load_dataset` disagree about the same file. The validator reports rows valid that the loader then silently drops (or, before #670, could not load at all). A validator that green-lights a file the loader rejects is worse than no validator, because it moves the failure from `validate` time to the middle of a training run.

The cause is that `validate_and_stats` (`src/soup_cli/data/validator.py`) never runs the conversion path. It validates a row by **top-level key presence only**:

```python
if expected_format and expected_format in FORMAT_SIGNATURES:
    required = FORMAT_SIGNATURES[expected_format]
    invalid = 0
    for row in data:
        if not required.issubset(row.keys()):
            invalid += 1
    valid_rows = len(data) - invalid
```

`load_dataset` instead runs `format_to_messages` on every row and drops the ones that return `None` (`_format_rows`, `loader.py`). Those two notions of "valid" are unrelated, so they drift apart in two distinct ways.

## Two failure modes, one symptom

**1. Six formats are not in `FORMAT_SIGNATURES` at all**, so the `if ... in FORMAT_SIGNATURES` guard is false, the whole validation block is skipped, and `valid_rows = len(data)` unconditionally — **every row is reported valid regardless of content**:

> `prm`, `pre_tokenized`, `input_output`, `video`, `multimodal`, `raft`

This is exactly what the #676 reporter saw: `--format multimodal` reports `5/5 rows valid` for a file `load_dataset` cannot fully load.

**2. The twelve formats that are in `FORMAT_SIGNATURES` are checked by key presence only**, so any element-level or value-level malformation the converter would reject passes validation. An `alpaca` row `{"instruction": "x", "output": null}` has both required keys, so it validates — while the converter drops it (`_require_str_content` rejects the null `output`).

## Verified

On current `main` (`ef2e02b`), a 5-row dataset of 4 good rows + 1 malformed row per format:

| format | in `FORMAT_SIGNATURES`? | `soup data validate` | `load_dataset` keeps | agree? |
|---|---|---|---|---|
| `input_output` | no — validation skipped | 5 / 5 valid | 4 rows | **no** |
| `multimodal` | no — validation skipped | 5 / 5 valid | 4 rows | **no** |
| `alpaca` | yes — key-presence only | 5 / 5 valid | 4 rows | **no** |
| `sharegpt` | yes — key-presence only | 5 / 5 valid | 4 rows | **no** |

The malformed row is: for `input_output` a non-dict `segments` element; for `multimodal` a non-dict `messages` element; for `alpaca` a null `output`; for `sharegpt` a null `conversations[].value`. Each is dropped by the loader and passed by the validator.

```python
from soup_cli.data.validator import validate_and_stats
from soup_cli.data.loader import _format_rows

good = {"instruction": "say hello", "output": "hello there"}
rows = [good] * 4 + [{"instruction": "x", "output": None}]   # loader drops the null-output row
validate_and_stats(rows, expected_format="alpaca")["valid_rows"]   # -> 5
len(_format_rows(rows, "alpaca"))                                  # -> 4
```

(The same class of gap widens once #676/#679 lands: `chatml`, `audio`, and `video` rows with a non-dict message will then be dropped by the loader while the validator still passes them — but the four cases above already disagree on `main` today, independent of that PR.)

## Suggested fix

Make the validator agree with the loader by construction: compute `valid_rows` by running the loader's own drop logic (`format_to_messages`) per row and counting the survivors, instead of maintaining a parallel, weaker key-presence check. That removes the second source of truth entirely, so the two commands cannot drift again — and it covers the six formats that are currently unvalidated for free, without having to enumerate signatures for them.

```python
from soup_cli.data.formats import format_to_messages

invalid = sum(1 for row in data if format_to_messages(row, expected_format) is None)
valid_rows = len(data) - invalid
if invalid:
    issues.append(f"{invalid} rows fail to convert for '{expected_format}' format")
```

The existing stats (duplicates, empty fields, short samples) are unaffected — only the `valid_rows` / issue computation changes. `detect_format` / `auto` behaviour is unchanged.

## Acceptance criteria

- [ ] For a file with N malformed rows, `soup data validate --format <fmt>` and `load_dataset` report the **same** count of usable rows, for every format in `format_to_messages`'s `valid_formats` (including the six not currently in `FORMAT_SIGNATURES`).
- [ ] A test that fails on the current behaviour: assert `validate_and_stats(mixed_rows, fmt)` counts the malformed row as invalid for both a `FORMAT_SIGNATURES` format (`alpaca`, null `output`) and a non-signature format (`input_output`, non-dict segment). The strong form — assert the *count*, so a revert to key-presence fails it.
- [ ] Valid rows still validate; unrelated stats (duplicates, short samples) unchanged.
- [ ] Changelog fragment under the newest release directory, crediting the author.

abdulwaarith0:
Fixes #695.

## What was wrong

`validate_and_stats` judged a row by top-level key presence against `FORMAT_SIGNATURES`, while `load_dataset` runs `format_to_messages` and drops any row that returns `None`. Two different notions of "valid", so they disagreed two ways:

1. The six formats absent from `FORMAT_SIGNATURES` (`prm`, `pre_tokenized`, `input_output`, `video`, `multimodal`, `raft`) skipped validation entirely and reported every row valid regardless of content.
2. For the checked formats, key presence passed rows a converter drops on a *value* — a null field, a non-dict message.

So `soup data validate` could report a file fully valid that the loader then can't load, moving the failure from validate time into the middle of a training run.

## The fix

- `valid_rows` is computed by running the real conversion path per row (`format_to_messages_with_reason`), so agreement with the loader is **structural**, not a second check that drifts. Every format is covered, including the six that were skipped.
- When a row is dropped, the converter's own reason is surfaced for the first few offenders (`row 3: chatml message must be a dict`) plus `... and N more`, so the report says which rows and why.
- **Exit-code behaviour is unchanged**: `validate` still exits 0 even at `0/N`, so the `soup ci init` gate is unaffected. Whether `0/N` should fail the gate is a separate question, left out of this PR.
- `format_to_messages` gains a `with_reason` variant that shares one dispatch and one drop-exception set with it; `format_to_messages`'s own behaviour is unchanged.

## Measurements (as requested on the issue)

**Agreement.** Across 360 file×format combos over the repo's own datasets, `validate` and `load_dataset` now agree on the count in **all 360** (0 mismatches).

**Over-reach.** vs the old validator, 120 of 360 verdicts change — every one an over-count reduced to the loader's true count (`N -> 0` on a file that genuinely has zero valid rows in that format), all in the six formats that previously skipped validation. No exit code moves in any combo.

**Cost.** 50k rows, best of three: `alpaca` 0.20s, `chatml` 0.33s — on par with the old key-presence version (the per-row stats already dominate; conversion adds negligible overhead).

**Test bar.** Tests assert `validate` and `load_dataset` agree on a *specific count*, not merely that some rows are flagged. Reverting the validator to key presence fails 14 of the 21 new tests.

## Notes

- Built against current `main`; if #678 lands first the only overlap is a few lines in `validator.py`.
- `tests/test_validator.py::test_validate_format_alpaca_invalid` is updated from the old key-presence message to the new "fail to convert" message and now also asserts the surfaced row/reason — strictly stronger, no assertion removed.

### DIFF
--- a/changelog.d/0.74.0/712.fixed.md
+++ b/changelog.d/0.74.0/712.fixed.md
@@ -1,0 +1,20 @@
+- **`soup data validate` green-lit rows that `load_dataset` then drops, because the
+  validator never ran the converters (#712 by @abdulwaarith0).**
+  `validate_and_stats` judged a row by top-level key presence against `FORMAT_SIGNATURES`,
+  while the loader runs `format_to_messages` and drops any row that returns `None`. Those
+  are different notions of "valid", so they disagreed two ways: the six formats absent from
+  `FORMAT_SIGNATURES` (`prm`, `pre_tokenized`, `input_output`, `video`, `multimodal`,
+  `raft`) skipped validation entirely and reported every row valid regardless of content,
+  and for the formats that were checked, key presence passed rows a converter drops on a
+  value — a null field, a non-dict message. `soup data validate` could report a file fully
+  valid that the loader could not load, moving the failure from validate time into the
+  middle of a training run. `valid_rows` is now computed by running the real conversion
+  path per row, so agreement is structural rather than a second, weaker check that drifts;
+  every format is covered, including the six that were skipped. When a row is dropped, the
+  converter's own reason is surfaced for the first few offenders (`row 3: chatml message
+  must be a dict`) so the report says which rows and why, not only a count. Exit-code
+  behaviour is unchanged: `validate` still exits 0 even at `0/N`, so the `soup ci init`
+  gate is unaffected; whether a `0/N` file should fail the gate is left to a separate
+  change. On the repo's own datasets this shifts 120 of 360 file-by-format verdicts, every
+  one an over-count on a file that genuinely has zero valid rows in that format, with no
+  exit code moved.
--- a/src/soup_cli/data/formats.py
+++ b/src/soup_cli/data/formats.py
@@ -77,6 +77,69 @@ def detect_format(data: list[dict]) -> str:
     )
 
 
+# Every format format_to_messages can convert. Kept as a module constant so
+# the plain and reason-returning entry points share one source of truth.
+VALID_FORMATS = (
+    "chatml", "alpaca", "sharegpt", "dpo", "kto", "llava", "sharegpt4v",
+    "plaintext", "embedding", "audio", "tool-calling",
+    # v0.42.0 Part A
+    "prm", "pre_tokenized", "input_output", "video", "multimodal",
+    # v0.62.0 Part A — RAFT (Retrieval-Augmented Fine-Tuning).
+    "raft",
+    # v0.71.32 — ASR (Whisper): {"audio": path, "text": transcript}.
+    "asr",
+)
+
+# A malformed row makes a converter raise one of these; the drop contract is
+# that format_to_messages turns them into a dropped row (None), which the
+# loader skips. Kept as a constant so validation drops rows on exactly the
+# same conditions the loader does, rather than a re-implementation that drifts.
+_DROP_EXCEPTIONS = (KeyError, TypeError, IndexError, ValueError)
+
+
+def _dispatch_conversion(row: dict, fmt: str) -> dict:
+    """Run the converter for ``fmt``, letting a malformed row raise.
+
+    Separated from :func:`format_to_messages` so callers that need the *reason*
+    a row was dropped (dataset validation) can catch the exception, rather than
+    only seeing the ``None`` that :func:`format_to_messages` returns.
+    """
+    if fmt == "chatml":
+        return _convert_chatml(row)
+    elif fmt == "alpaca":
+        return _convert_alpaca(row)
+    elif fmt == "sharegpt":
+        return _convert_sharegpt(row)
+    elif fmt == "dpo":
+        return _convert_dpo(row)
+    elif fmt == "kto":
+        return _convert_kto(row)
+    elif fmt == "plaintext":
+        return _convert_plaintext(row)
+    elif fmt == "embedding":
+        return _convert_embedding(row)
+    elif fmt == "audio":
+        return _convert_audio(row)
+    elif fmt == "tool-calling":
+        return _convert_tool_calling(row)
+    elif fmt == "prm":
+        return _convert_prm(row)
+    elif fmt == "pre_tokenized":
+        return _convert_pre_tokenized(row)
+    elif fmt == "input_output":
+        return _convert_input_output(row)
+    elif fmt == "video":
+        return _convert_video(row)
+    elif fmt == "multimodal":
+        return _convert_multimodal(row)
+    elif fmt == "raft":
+        return _convert_raft(row)
+    elif fmt == "asr":
+        return _convert_asr(row)
+    else:
+        return _convert_vision(row)
+
+
 def format_to_messages(row: dict, fmt: str) -> Optional[dict]:
     """Convert any format to normalized structure for training.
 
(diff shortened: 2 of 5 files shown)


MakazhanAlpamys:
Merged as `bedd4da`. Thank you — and this is the best-evidenced PR in the current queue.

## What I ran

Worktree at `6b11bdd`, `PYTHONPATH` pinned into the worktree and checked via `soup_cli.__file__`, py3.10, `--no-cov`. Baseline on clean `main` first, so nothing here is #651 noise.

**The CLI, not the helper.** `soup data validate <file> --format multimodal` through `CliRunner`:

```
main:  5/5 rows valid
this:  4/5 rows valid
       row 4: multimodal message must be a dict
```

`multimodal` is the exact format #695 reports as saying 5/5. That is the defect and its repair, at the command.

**Mutations — four applied, four killed.** The decisive one is mine: neutering the per-row reason so the validator can never mark a row invalid, which is #695 restored.

```
16 failed, 18 passed
```

The other three — reverting the gate to `FORMAT_SIGNATURES`, dropping the reason/row-index surfacing, widening `_MAX_REASON_SAMPLES` — each fail exactly the tests that name them, which is the property you want: the suite discriminates between distinct wrong answers rather than just detecting "something changed".

**The `formats.py` refactor, which is the part I would normally push back on.** +87/-46 in a file every training run passes through is exactly the diff a reviewer has to take on trust. So it was checked instead of trusted: a differential harness over **20 formats x 51 row shapes = 1020 cases**, comparing `format_to_messages` output and raised exception type and message between the two trees. **Byte-identical, zero diffs**, unknown-format `ValueError` included. `soup train` (via `_format_rows`) and `soup data convert` both route through that function, so neither moves. The change is extraction, not redesign.

**No false positives on good data.** Every shipped `examples/data/*.jsonl` validates identically under both builds.

**The two removed test lines**, audited line by line: one docstring, and one assertion on a message string that no longer exists — replaced by the equivalent `"fail to convert"` check *plus* a new `"row 1"` assertion, with `valid_rows == 1` untouched. Strictly stronger. I check every removed test line in every PR, and this is one of the few where the replacement is better than what it replaced.

**Zero mocks.** The tests call the real converters. That matters more here than usual: `tests/test_embedding.py` currently holds 79 green tests over a path that crashes in production, because a `MagicMock` auto-creates the attributes production lacks. Yours cannot make that mistake.

**You disclosed the blast radius yourself.** Default `soup data validate` with auto-detect is affected, and your fragment says so with numbers rather than leaving it for review to discover. You also checked that `validate` exits non-zero only on a missing file or an undetectable format, which is what makes it safe for the `soup ci init` PR gate — a check I would have had to run myself and would have held the merge on.

**Second merge, second time the qualification is as valuable as the fix.** On #670 you did not overclaim the reach of the multimodal crash, and review found `detect_format` never returns `multimodal`. Same discipline here.

## One note for later, not a change request

The new reason strings reach the terminal unescaped (`commands/data.py`: `console.print(f"  [yellow]![/] {issue}")`). It is safe as written because every converter interpolates a field name or `type(value).__name__`, never a value from the row — I tried pushing `\x1b[31m` and a `[link=...]` payload through row content and nothing reached the terminal. But that safety is a property of the current converters rather than of the printer, so a future converter that puts a *row value* into its message would turn this into an injection point. One comment in `formats.py` saying "reason strings are printed unescaped; never interpolate row content" would make it a rule instead of an accident. Worth doing whenever you are next in that file.

Credited in `CONTRIBUTORS.md` (`95f925d`).

**Heads-up on sequencing:** #678 also touches `data/formats.py` and `data/validator.py` and will now need a rebase onto this. That is on me for the ordering, not on you — I warned its author before merging.

PR #715, opened 2026-09-06
feat(ci): resolve every recipe model id on a schedule (#677)

Srinivasan8888:
Closes the fourth acceptance item of #661 — the general guard. **And its first live run found nine more broken recipes.**

[Claimed here](https://github.com/MakazhanAlpamys/Soup/issues/677). `Refs #677`.

## What the guard found on its first run

Nine recipes name a repo that does not resolve anonymously, beyond the two #661 already caught:

```
checked 165 recipes / 330 surfaces
  exists=244  gated=68  missing=18  unverified=0

MISSING — 9 recipe(s):
  cogito-v2-sft            deepcogito/cogito-v2-preview
  devstral-sft             mistralai/Devstral-Small
  gemma3-9b-sft            google/gemma-3-9b-it
  glm-4.6-sft              THUDM/glm-4.6
  granite-4-sft            ibm-granite/granite-4.0-tiny-base
  internvl-3-5-sft         OpenGVLab/InternVL3-5
  kimi-k2-sft              moonshotai/Kimi-K2
  mistral-medium-3-5-sft   mistralai/Mistral-Medium-3.5
  voxtral-sft              mistralai/Voxtral-Mini-3B
```

**I am stating that precisely: "does not resolve anonymously", not "does not exist".** `RepositoryNotFoundError` covers missing *and* private, and anonymously those are indistinguishable. For a shipped recipe the distinction does not matter much — a user cannot reach either — but it is not the same claim and I am not going to make the stronger one.

Two of these repeat patterns already fixed once. `gemma3-9b-sft` names a **Gemma-3 9B**, and Gemma 3 has no 9B size — the identical error #661 found in `gemma3-9b-sft-mlx`. `glm-4.6-sft` names **THUDM**, the org that was wrong for `glm-5-sft` and had to be corrected to `zai-org` after release. The same two mistakes were made twice because nothing checked.

Verified independently, with both controls, before writing any of that down:

```
zz-invented/definitely-not-real-xyz  -> RepositoryNotFoundError   (negative control)
meta-llama/Llama-3.1-8B-Instruct     -> EXISTS gated=manual       (gated ≠ missing)
google/gemma-3-12b-it                -> EXISTS gated=manual
```

**Repairing them is not in this PR** — that is a catalog decision per recipe (rename? repoint? drop?), the same call `gemma3-9b-sft-mlx` needed in #666. This PR ships the detector. Happy to take the repairs as a follow-up, or leave them for whoever wants them.

## The three constraints from the issue

**1. A bare status code is not evidence.** The classifier takes `RepositoryNotFoundError` injected as a parameter, never a status code. That is the mistake I made in #661 before the report was worth filing, and the negative control above is in the test suite so the checker is provably able to fail.

**2. Gated ≠ missing.** Four outcomes, not two: `EXISTS`, `GATED`, `MISSING`, `UNVERIFIED`. 68 of the 330 surfaces resolved as gated on the live run and none is reported. A guard that fires on every `meta-llama/*` forever is one people mute.

**3. Not in the PR matrix.** Weekly `schedule:` plus `workflow_dispatch:`, and a test **asserts the absence** of a `pull_request` trigger rather than trusting the YAML to stay that way — the #596 lesson, where a job existed and collected nothing.

Plus one the issue did not ask for: **`UNVERIFIED` exists so a bad minute at the Hub is not a bug report.** Timeouts and 5xx retry with backoff and then classify as "could not check", which is reported separately and does **not** fail the job. `RepositoryNotFoundError` is a definite answer and is deliberately not retried, so it cannot burn the budget transient failures need.

And a guard on the guard: the script **refuses to run** if `HF_TOKEN` is set, because a token resolves gated repos as `EXISTS` and would silence the check exactly where a wrong id is most likely to hide.

```
$ HF_TOKEN=... python scripts/check_recipe_repo_ids.py
HF_TOKEN is set. This check must run anonymously: a token turns every gated
repo into EXISTS and hides the case the guard is for.
```

## Tests

15, none touching the network — the Hub client is a fake, so classification is pinned without depending on Hub availability. Both surfaces are covered independently, because #666's mutation testing showed the snapshot guard sees only the YAML half and a wrong `RecipeMeta.model` slips past everything else.

Covered: each of the four statuses; a transient failure retried rather than classified; a 404 **not** retried; a wrong `RecipeMeta.model` caught with a correct YAML and vice versa; only `MISSING` reported; `UNVERIFIED` reported separately; the negative control; and the three workflow-trigger assertions.

## Your two open questions from my claim

Both still yours, and neither blocks this:

**Reporting surface.** v1 **fails the job** rather than opening an issue — less machinery, no bot-authored issues in your tracker, and a red scheduled run is already visible. Swapping to "open or update one issue" is contained to `main()` if you prefer it.

**Offline PR check.** Not built. The weekly job could write a committed list of known-bad ids that `validate-recipes` reads offline, closing the window where a broken id is known but still merges. More moving parts, so I did not build it unopposed.

```
$ .venv/bin/python -m pytest tests/test_issue677_recipe_repo_id_guard.py -q -o addopts=
15 passed
$ .venv/bin/python -m ruff check scripts/ tests/
All checks passed!
$ git diff origin/main -- tests/ | grep -c "^-[^-]"
0
```

### DIFF
--- a/.github/workflows/recipe-repo-ids.yml
+++ b/.github/workflows/recipe-repo-ids.yml
@@ -1,0 +1,43 @@
+# Recipe repo-id resolution (#677).
+#
+# The catalog names a Hugging Face repo in two places per recipe --
+# `RecipeMeta.model` and the YAML `base:` -- and nothing checked that either
+# resolves. #661 found two that did not, both shipped and released:
+# `validate-recipes` parses the YAML through the real schema, and a nonexistent
+# repo id is a perfectly well-formed string.
+#
+# Deliberately NOT a pull-request gate. Resolving 160+ ids on every PR would
+# make an unrelated docs change depend on Hub availability and rate limits, and
+# a job that goes red for reasons unrelated to the diff is one people learn to
+# re-run rather than read. Repos disappear on a timescale of months, so weekly
+# finds it long before a user does.
+name: recipe repo ids
+
+on:
+  schedule:
+    # Mondays, 07:00 UTC -- an hour after dependency-drift so the two do not
+    # contend for the same Hub rate limit.
+    - cron: "0 7 * * 1"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  resolve:
+    name: resolve every recipe model id anonymously
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install the catalog and the Hub client only
+        # Not `.[dev]`: this needs the recipe catalog and huggingface_hub, not
+        # the PyTorch stack. A lighter install is also a faster scheduled job.
+        run: pip install -e . huggingface-hub pyyaml
+      - name: Resolve
+        # No HF_TOKEN in scope, deliberately and by omission: a token would
+        # resolve gated repos as EXISTS and hide the very case this guards.
+        # The script also refuses to run if one is set.
+        run: python scripts/check_recipe_repo_ids.py
--- a/.github/workflows/recipe-repo-ids.yml
+++ b/.github/workflows/recipe-repo-ids.yml
@@ -11,6 +11,12 @@
 # a job that goes red for reasons unrelated to the diff is one people learn to
 # re-run rather than read. Repos disappear on a timescale of months, so weekly
 # finds it long before a user does.
+#
+# Measured cost, 2026-09-06, unauthenticated against the live Hub:
+#   165 recipes / 330 surfaces / 118 unique ids after de-duplication (64% saved)
+#   29.8 s wall clock, 239 ms median, 340 ms p95, 0 rate limits
+# De-duplication matters because gated bases are shared: 18 unique gated ids
+# cover 68 of the 330 surfaces. Without it the sweep is ~83 s.
 name: recipe repo ids
 
 on:
@@ -20,6 +26,7 @@ on:
     - cron: "0 7 * * 1"
   workflow_dispatch:
 
+# Never add a pull_request trigger here.
 permissions:
   contents: read
 
--- a/scripts/check_recipe_repo_ids.py
+++ b/scripts/check_recipe_repo_ids.py
@@ -80,6 +80,7 @@ def classify_repo(
     repo_id: str,
     *,
     not_found: type[BaseException],
+    gated: type[BaseException] | None = None,
     attempts: int = _DEFAULT_ATTEMPTS,
     backoff: float = _DEFAULT_BACKOFF,
 ) -> RepoCheck:
@@ -91,11 +92,26 @@ def classify_repo(
 
     A ``RepositoryNotFoundError`` is a definite answer and is **not** retried;
     retrying it would burn the attempt budget that transient failures need.
+
+    ``gated`` **must** be caught before ``not_found``, because
+    ``GatedRepoError`` is a *subclass* of ``RepositoryNotFoundError`` in
+    ``huggingface_hub``::
+
+        GatedRepoError MRO: GatedRepoError -> RepositoryNotFoundError -> ...
+
+    Ordered the other way, ``except not_found`` swallows the gated answer and
+    every gated repo classifies MISSING -- the exact failure that gets a guard
+    like this muted. It does not show up in a live sweep today because the Hub
+    currently serves gated *metadata* anonymously, so those ids take the
+    returning path and ``info.gated`` is read instead. One org changing
+    metadata visibility would flip them all to the raising path.
     """
     last_detail = ""
     for attempt in range(1, max(1, attempts) + 1):
         try:
             info = api.model_info(repo_id)
+        except (gated or ()) as exc:  # MUST precede not_found -- see docstring
+            return RepoCheck(repo_id, Status.GATED, str(exc)[:200])
         except not_found as exc:
             return RepoCheck(repo_id, Status.MISSING, str(exc)[:200])
         except Exception as exc:  # noqa: BLE001 -- anything else may be transient
@@ -127,6 +143,7 @@ def check_surfaces(
     *,
     api: Any,
     not_found: type[BaseException],
+    gated: type[BaseException] | None = None,
     attempts: int = _DEFAULT_ATTEMPTS,
     backoff: float = _DEFAULT_BACKOFF,
 ) -> dict[str, RecipeReport]:
@@ -136,7 +153,8 @@ def check_surfaces(
     def _check(repo_id: str) -> RepoCheck:
         if repo_id not in cache:
             cache[repo_id] = classify_repo(
-                api, repo_id, not_found=not_found, attempts=attempts, backoff=backoff
+                api, repo_id, not_found=not_found, gated=gated,
+                attempts=attempts, backoff=backoff,
             )
         return cache[repo_id]
 
@@ -215,11 +233,29 @@ def format_report(report: dict[str, RecipeReport]) -> str:
     return "\n".join(lines)
 
 
-def main() -> int:
-    from huggingface_hub.errors import RepositoryNotFoundError
+def main(
+    *,
+    not_found: type[BaseException] | None = None,
+    gated: type[BaseException] | None = None,
+) -> int:
+    """Resolve the catalog and return a process exit code.
+
+    The exception classes are injectable for the same reason ``classify_repo``
+    takes them: it keeps the exit-code contract testable without the Hub, and
+    without constructing real ``huggingface_hub`` errors (which need a live
+    response object). Defaults are the real classes.
+    """
+    if not_found is None or gated is None:
+        from huggingface_hub.errors import GatedRepoError, RepositoryNotFoundError
+
+        not_found = not_found or RepositoryNotFoundError
+        gated = gated or GatedRepoError
 
     report = check_surfaces(
-        collect_recipe_repo_ids(), api=_anonymous_api(), not_found=RepositoryNotFoundError
+        collect_recipe_repo_ids(),
+        api=_anonymous_api(),
+        not_found=not_found,
+        gated=gated,
     )
     print(format_report(report))
     # Exit non-zero ONLY for genuinely missing repos. An unverified run is not a
--- a/scripts/check_recipe_repo_ids.py
+++ b/scripts/check_recipe_repo_ids.py
@@ -1,0 +1,231 @@
+#!/usr/bin/env python3
+"""Resolve every shipped recipe's Hugging Face repo id, anonymously (#677).
+
+The catalog names a repo in two independent places per recipe -- ``RecipeMeta.model``
+and the YAML ``base:`` -- and until now nothing checked that either resolves.
+#661 found two that did not; both had shipped and been released, because
+``validate-recipes`` parses the YAML through the real schema and a nonexistent
+repo id is a perfectly well-formed string.
+
+Three things this deliberately does NOT do, each of which would make the guard
+worse than nothing:
+
+**It does not treat a status code as evidence.** An invented repo id returns the
+same ``401`` from the Hub API as a real private one -- the mistake made in #661
+before that report was worth filing. The authoritative signal is
+``huggingface_hub``'s ``RepositoryNotFoundError``.
+
+**It does not report gated repos.** A gated repo (``meta-llama/*``, ``google/gemma-*``)
+resolves for an authorised user and refuses an anonymous one. Reporting those
+would fire forever on correct recipes, and a guard that cries wolf gets muted.
+``model_info`` exposes ``gated``; this reads it rather than inferring.
+
+**It does not report a transient failure as a missing repo.** Timeouts, 5xx and
+rate limits classify as UNVERIFIED -- "could not check" -- after retries. A bad
+minute at the Hub must not become a bug report.
+
+Run by ``.github/workflows/recipe-repo-ids.yml`` on a schedule. Never on pull
+requests: 163 network calls per PR would make every unrelated change depend on
+Hub availability.
+"""
+
+from __future__ import annotations
+
+import enum
+import os
+import sys
+import time
+from dataclasses import dataclass
+from typing import Any
+
+_DEFAULT_ATTEMPTS = 3
+_DEFAULT_BACKOFF = 2.0
+
+
+class Status(enum.Enum):
+    """Four outcomes, not two. Collapsing GATED or UNVERIFIED into MISSING is
+    what turns this from a guard into noise."""
+
+    EXISTS = "exists"
+    GATED = "gated"
+    MISSING = "missing"
+    UNVERIFIED = "unverified"
+
+
+@dataclass(frozen=True)
+class RepoCheck:
+    repo_id: str
+    status: Status
+    detail: str = ""
+
+
+@dataclass(frozen=True)
+class SurfacePair:
+    """The two places a recipe names its model. Both are checked because #666's
+    mutation testing showed the snapshot guard sees only the YAML half, so a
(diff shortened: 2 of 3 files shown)


Srinivasan8888:
**The measurement you asked for, rather than a policy chosen for looking reasonable.** Run just now on this box, unauthenticated, against the live Hub:

```
recipes            : 165
surfaces (2 each)  : 330
unique repo ids    : 118
dedup saves        : 212 calls (64%)

wall clock         : 29.8s for 118 unique ids
per-call median    : 239 ms
per-call p95       : 340 ms
slowest call       : 523 ms
RATE LIMITED       : 0
outcomes           : exists=91  gated=18  missing=9

projected without dedup: ~83s
```

**No rate limiting at this volume.** So the retry policy stays as built — 3 attempts with linear backoff, and `RepositoryNotFoundError` never retried — rather than growing a delay for a problem that does not occur. If a future catalog doubles and this starts limiting, the measurement to repeat is above and the first lever is the delay, not a bigger backoff.

**De-duplication is already in** (`check_surfaces` caches by id), and it is worth more than I expected: **64% of calls saved**, because gated bases are shared across recipes. 18 unique gated ids cover 68 of the 330 surfaces. That is also why the sweep is 30s rather than 83s.

One number reads differently between this and the PR body, and both are right: the body says `gated=68` counting **surfaces**, this says `gated=18` counting **unique ids**. Same fact, different denominator. I have said so here rather than let a reader find two numbers and wonder which is wrong.

---

On the two answers: **fail the job** and **no offline known-bad list** are both what is built, so nothing changes. Your reasoning against the list is better than mine — I framed it as moving parts; "a second source of truth that goes stale and is then asserting something that was true last Tuesday" is the actual failure, and it is the shape this repo keeps finding.

**Report keyed by recipe** is in: `missing_only()` returns recipe names, and the formatter prints `recipe: surface -> id` per broken surface, so a `RecipeMeta.model` / YAML `base:` disagreement shows as one recipe with two lines rather than two unrelated ids. That is the case #661 actually found.

---

**On the #677 disclosure — no issue from me at all, and please do not treat it as one.** You advertised it to @v01dst without an expiry, noticed, and said so unprompted before anyone could discover it. That is the same thing you did on #669 when you duplicated my fix, and it is the reason overlap gets corrected here instead of festering. If @v01dst wants #677 I will hand it over without argument — I have #713 and #715 open and would not be short of work.

MakazhanAlpamys:
@Srinivasan8888 — **the design is right, it matches both rulings exactly, and you delivered the rate-limit measurement I asked for. Blocking on one real defect and one whole untested layer — and the defect is precisely the direction I said decides whether this survives six months.**

## BLOCKING 1 — `GatedRepoError` is a subclass of `RepositoryNotFoundError`, so every gated repo classifies MISSING

```
huggingface_hub 1.30.0
GatedRepoError MRO: ['GatedRepoError', 'RepositoryNotFoundError', 'HfHubHTTPError', 'HTTPError', 'OSError']
issubclass(GatedRepoError, RepositoryNotFoundError) = True
```

So `except not_found` swallows a gated answer. Run against **your** `classify_repo`, offline, with the real exception classes:

```
model_info raises GatedRepoError            -> Status.MISSING   <-- should be GATED
model_info raises RepositoryNotFoundError   -> Status.MISSING
```

Your gated test pins only the *other* path — where `model_info` **returns** and `info.gated` is truthy. In your live sweep all 18 unique gated ids took that returning path, because the Hub currently serves gated *metadata* publicly, which is exactly why this did not show up in a run that otherwise looked complete. But the raising path exists in the library, and one org changing metadata visibility turns `meta-llama/*` into a permanent MISSING report — the failure mode we both named as the one that gets the guard muted.

One line: catch `GatedRepoError` **before** `not_found`, injected the same way `not_found` is so it stays testable without the Hub. Plus a test that **raises** it rather than returning `gated=True`.

## BLOCKING 2 — the entire reporting-and-exit layer is untested, and five mutations live there

Seventeen mutations, twelve killed. The twelve include everything about classification: GATED->MISSING, GATED->EXISTS, never reading `.gated`, UNVERIFIED->MISSING, `not_found`->EXISTS, retrying a 404, never retrying a transient, skipping either surface, and adding a `pull_request:` trigger. That half is genuinely well tested.

The five survivors are all downstream of it:

| mutation | result |
|---|---|
| `main()` always returns 0 | **survived — 15/15 pass** |
| UNVERIFIED **fails** the job | **survived — 15/15 pass** |
| drop `token=False` | **survived** |
| delete the env-var refusal | **survived** |
| `format_report` prints no missing recipe names / no COULD-NOT-CHECK section | **survived** |

Coverage says the same thing: `71%`, missing `179-187` (all of `_anonymous_api`), `191-215` (all of `format_report`), `219-227` (all of `main`).

Two of those are the requirements themselves. **"Fails the job" is requirement 7 and nothing pins the exit code.** And **"an unverified run is not a failure" — the property you correctly called the load-bearing one — is asserted only in a code comment**, so a future edit can silently reverse it and turn the Hub's bad minute back into a red job.

Four tests close all five:

1. `_anonymous_api()`: the refusal fires on each of the three env vars, and `api.token is False`.
2. `main()`: returns 1 on MISSING, and **0 on UNVERIFIED-only**.
3. `format_report()`: a missing recipe's name and both surface labels appear; the COULD-NOT-CHECK section appears.
4. One test piping `collect_recipe_repo_ids()` through `check_surfaces` with the fake api — that covers line **146**, the `isinstance(pair, SurfacePair)` branch, which is the **production** path. Every current test feeds plain tuples, so the test-only convenience branch is covered and the real one is not.

## What is right, and worth saying because most of it is

**Both rulings honoured.** `permissions: {contents: read}` and nothing else — no `issues: write`, no write of any kind. No known-bad-id list. Reachability traced rather than grepped: workflow step -> `sys.exit(main())` -> `return 1 if missing_only(report) else 0` -> failed job, and `pip install -e .` suffices because `soup_cli.recipes.catalog` imports without torch or transformers.

**The 401 trap is handled in code.** `classify_repo` takes the exception *class* as a parameter and never touches a status code. `_anonymous_api()` refuses on all three env vars — I ran it, it fires on each — and `HfApi(token=False).token` is `False`, which also neutralises a cached `~/.cache/huggingface/token` that the env check alone would miss. That last part is better than what the issue asked for.

**Zero `MagicMock`.** Hand-written `_FakeApi` / `_Info` / `_NotFoundError` with explicit attributes, and the classifier itself is never stubbed — only the Hub client. That is the correct shape and the opposite of the trap that has 79 green tests standing over a production crash in `test_embedding.py`. No test touches the network; the file runs in 0.7 s. No vacuous assertions — `test_every_recipe_contributes_meta_and_yaml_ids` asserts `len(surfaces) > 100` before looping, so the loop cannot be trivially satisfied. Zero test lines removed. Ruff clean. 1596 passed across the whole recipe surface, and 15/15 three times under random ordering.

**The measurement I asked for, delivered.** 165 recipes / 330 surfaces / **118 unique ids** / 212 calls saved / 29.8 s / p95 340 ms / **0 rate limits**. I counted the unique ids independently offline and got exactly 118. Please fold the headline into the workflow header comment — it is the number the next person will want and a PR comment is not where they will look.

## Non-blocking

**The trigger test fires on correct code.** `assert "pull_request" not in text` scans the whole file, so adding the comment `# Never add a pull_request trigger here.` **fails the test**. That is the #404 lesson this repo already wrote down — a guard that fires on correct code is one that gets deleted. The better in-repo precedent already exists at `tests/test_recipes_v031.py:439-453`, which parses `parsed.get(True) or parsed.get("on")` and inspects the trigger keys. Follow that, and assert `push` and `pull_request_target` absent while you are there.

**One equivalent mutant, checked rather than assumed.** Mirroring `meta.model` into `yaml_base` in the collector survives — because all 165 recipes currently have `meta.model == yaml base`, zero divergences, so it cannot change today's output. It still marks a real gap: the collector could stop reading the YAML entirely and nothing would notice, and divergence is the exact case the two-surface design exists for. Test 4 above kills it if you monkeypatch `RECIPES` with one synthetic entry whose YAML base differs.

**The negative control is a duplicate.** `test_negative_control_an_invented_id_is_missing` is byte-for-byte `test_repository_not_found_is_missing` with a different repo string. No harm, but it is not the independent evidence the acceptance criterion was asking for.

**No changelog fragment needed** — CI-only, no user-visible behaviour, and `1082db4` on main set that precedent. Correct as shipped.

## One thing in your description I checked and it holds

"Does not resolve anonymously", not "does not exist" — the careful framing is right, and I did not verify the specific ids because that needs the network. Your "same mistake twice" observations check out structurally: `glm-4.6-sft` still names `THUDM/glm-4.6` while `glm-5-sft` was repaired to `zai-org/GLM-5`, and `gemma3-9b-sft-mlx` is gone from the catalog while `gemma3-9b-sft` still names a Gemma-3 9B. Those are findings for a follow-up, not for this PR — the guard's job is to surface them, and it does.

Fix the `GatedRepoError` ordering and add the four tests, and this merges.

Srinivasan8888:
Both blocking findings fixed, all four non-blocking ones taken. **7/7 mutations killed, coverage 71% → 95%.**

## 1 — `GatedRepoError` subclasses `RepositoryNotFoundError`

Reproduced against the installed library before touching anything:

```
GatedRepoError MRO: GatedRepoError -> RepositoryNotFoundError -> HfHubHTTPError -> ...
raises GatedRepoError -> MISSING      (before)
raises GatedRepoError -> GATED        (after)
```

The gated class is now injected the same way `not_found` is and caught first. Two tests: one that **raises** rather than returns, and one pinning the upstream subclass relationship the ordering depends on — so if `huggingface_hub` ever un-nests them, that says so instead of the ordering becoming quietly pointless.

Your diagnosis of *why* it hid is the part worth repeating: my gated test pinned only the returning path, and the live sweep took that path for all 18 gated ids because the Hub currently serves gated metadata anonymously. A run that looked complete could not have found it.

## 2 — the reporting-and-exit layer

All five die by name. `main()` now takes the exception classes as injectable parameters, which is what makes the exit contract testable without constructing real `huggingface_hub` errors — they need a live response object, and faking one with `MagicMock` would have undone the no-mocks property you called out as correct.

The two that are requirements rather than details:

```
main() always returns 0     -> KILLED (test_main_returns_1_when_a_repo_is_missing)
UNVERIFIED fails the job    -> KILLED (test_an_unverified_only_run_does_not_fail)
```

You were right that the load-bearing property was asserted only in a comment.

## Non-blocking, all four

**The trigger test fired on correct code.** Now parses the YAML and inspects trigger keys per `tests/test_recipes_v031.py`, checks `push` and `pull_request_target` too, and pins `permissions == {contents: read}`. I verified the new version survives the comment that broke the old one — and put that comment in the file, since it is useful and now safe.

**The equivalent mutant is killed** by monkeypatching `RECIPES` with a synthetic entry whose YAML base diverges from `meta.model`. You were right that it marks a real gap even though it changes no output today.

**The negative control** was a restatement; it now drives the id through collector → `check_surfaces` → `missing_only`, so a checker that classified correctly but reported nothing fails it.

**The measurement is in the workflow header**, where the next person will look.

## One process failure of mine, which is how the ordering bug got caught twice

My first attempt at the gated test used `str.replace` **without asserting the anchor matched**. It silently inserted nothing and printed success. Every test passed, coverage went up — and the ordering mutation still survived, which is the only reason I noticed the test did not exist.

I have asserted anchor counts on every other edit in this session and skipped it exactly once, on the test for the finding you had just called the most important one. Every anchored edit in this change now asserts its count first.

That is also the second time today a mutation caught something my green suite did not. The suite passing is not the evidence; the mutation dying is.

```
30 passed · 7/7 mutations killed · coverage 95% · ruff clean · 0 removed test lines
```

MakazhanAlpamys:
Merged as `0038704`. **Both blocking findings are closed in code, all five of the survivors I named are dead, and the guard is safer than I asked for in two places.**

## The five survivors, re-run by me on your head

| mutation | now dies by |
|---|---|
| `main()` always returns 0 | `test_main_returns_1_when_a_repo_is_missing` |
| unverified made to fail the job | `test_an_unverified_only_run_does_not_fail` |
| `token=False` dropped | `test_the_client_it_returns_has_authentication_off` |
| the token-in-scope refusal deleted | `test_it_refuses_when_a_token_is_in_scope[HF_TOKEN / HUGGING_FACE_HUB_TOKEN / HUGGINGFACEHUB_API_TOKEN]` |
| `format_report` prints no missing names | `test_it_names_the_broken_recipe_and_the_surface` |

Plus the exception ordering, both ways: swapping `except not_found` in front of the gated branch, and deleting the gated branch entirely, each fail `test_a_gated_repo_that_raises_is_still_gated`. And the two workflow mutations — adding a `pull_request:` trigger, adding `issues: write` — die by name.

Ten mutations, nine killed. Coverage on the script measures **96%**, not the 95% you claimed; the four missed lines are the real `huggingface_hub` import inside `main` and the `__main__` guard, both correctly unreachable offline.

## The part that was better than the ask

I asked for a test that **raises** `GatedRepoError` rather than returning `gated=True`. You wrote that, and then added `test_the_real_hub_classes_are_still_subclassed_that_way`, which asserts `issubclass(GatedRepoError, RepositoryNotFoundError)` against the installed library. I did not ask for it and it is the more durable half: if upstream ever un-nests those two, the ordering this PR exists to get right becomes pointless, and your test says so instead of the guard quietly degrading. That is the difference between fixing a bug and pinning the assumption the fix rests on.

The trigger test is also now the right kind. It parses the YAML and inspects trigger keys instead of scanning for a substring — which matters concretely, because the file now contains the literal comment `# Never add a pull_request trigger here.` and the old scan would have failed on its own documentation.

## Safety audit, run by me rather than read off the PR

| property | result |
|---|---|
| runs on PRs | **no** — `schedule` + `workflow_dispatch` only; no `push`, no `pull_request`, no `pull_request_target` |
| can fail the PR gate | **no** — separate workflow, no PR-triggered event |
| permissions | `contents: read` only |
| secrets | none referenced; the script additionally refuses to start if any of three token vars is set |
| on failure | fails the job; opens no issue, writes nothing |
| unverified | separate section, returns 0 |
| network in the default test suite | **none** — I monkeypatched `socket.connect` / `create_connection` / `connect_ex` to raise and re-ran: **30 passed** |
| install viability | `soup_cli.recipes.catalog` imports with zero of torch/transformers/peft/trl in `sys.modules`, 165 recipes / 165 surface pairs |

The rate-limit measurement is what made this mergeable at all rather than a plausible-looking cron job: 118 unique ids from 330 surfaces, 64% of calls saved by caching, 29.8 s unauthenticated, zero limits.

## One survivor, recorded not hidden

`scripts/check_recipe_repo_ids.py:163-164` — in the `SurfacePair` branch, reading `pair.meta_model` for **both** ids passes all 30 tests. It is the same gap I named, one function downstream of where you closed it: the two tests that prove the surfaces are independent feed plain tuples and take the `else: # used by tests` branch, and the one test driving real `SurfacePair`s feeds identical ids, so it cannot discriminate.

**Not blocking, and the reason is the same one I gave for the collector mutant:** it is currently equivalent — 0 of 165 recipes have `meta.model` differing from the YAML base — so it changes no output today. It is in the merge commit and on #677 as a marked gap. If you want it closed, it is about three lines: drive `check_surfaces` with a divergent `SurfacePair` (`meta_model="good/real", yaml_base="bad/invented"`) and assert `report["r"].yaml.status is Status.MISSING`.

Also on the record: `classify_repo` and `check_surfaces` default `gated=None`, so a direct library caller who passes `not_found` and omits `gated` gets the old behaviour. Unreachable from production — `main()` resolves both — and it is what makes the classifier testable without the Hub, so it is a deliberate trade rather than a defect.

No changelog fragment, which is correct: CI-only, no user-visible behaviour. I checked the precedent rather than asserting it — `1082db4` (`ci: lint benchmarks/ as well`) shipped without one.

Credited in `CONTRIBUTORS.md` (`64d029e`).

PR #717, opened 2026-09-06
Raise PyTorch floor to 2.6.0 for TRL FSDP2 (#651)

Samearth17:
## Summary
- raise the declared training PyTorch minimum from 2.5.x to 2.6.0
- pin the dependency-floor CI stack and `soup doctor` to Torch 2.6.0
- update the floor guard tests and changelog for TRL 0.29 FSDP2 requirements

## Validation
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -o addopts= -q tests/test_issue636_torch_floor.py`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -o addopts= -q tests/test_v07203.py`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -o addopts= -q`
- `git diff --check`

Full suite passes locally: 20,279 passed, 108 skipped.

### DIFF
--- a/.github/constraints/transformers-floor.txt
+++ b/.github/constraints/transformers-floor.txt
@@ -2,10 +2,10 @@
 # Qwen3.8-Flash-Next text loading establishes Transformers 5.16.1 as Soup's
 # declared training floor. trl 0.29 is the first release that does not import the
 # Transformers private availability helper whose v5 return type broke trl 0.28.
-# Torch 2.5.1 deliberately exercises Qwen4-Exp's legacy int64-only scatter path.
+# Torch 2.6.0 deliberately exercises Qwen4-Exp's legacy int64-only scatter path.
 # Plotext 6.0.0 is pinned here so one CI cell exercises the real Figure API rather
 # than relying only on the fake 5.x/6.x contract doubles in the unit suite.
-torch==2.5.1
+torch==2.6.0
 transformers==5.16.1
 trl==0.29.0
 peft==0.20.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,10 +119,10 @@ jobs:
       - name: Install training dependencies
         # torch>=2.6 is load-bearing, not belt-and-braces: sshleifer/tiny-gpt2
         # ships only pytorch_model.bin, and transformers refuses torch.load
-        # below 2.6 under CVE-2025-32434. Removing this pin because "[dev]
+        # below 2.6.0 under CVE-2025-32434. Removing this pin because "[dev]
         # already pulls torch" breaks the job with a message naming the CVE
         # rather than the cause.
-        run: pip install -e ".[dev]" "torch>=2.6"
+        run: pip install -e ".[dev]" "torch>=2.6.0"
 
       - name: Warm tiny GPT-2 cache
         shell: python
@@ -143,7 +143,7 @@ jobs:
           path: ${{ runner.temp }}/hf-cache
           key: hf-pytorch-smoke-${{ runner.os }}-${{ hashFiles('tests/test_smoke_train.py', 'pyproject.toml') }}
 
-  # #502/#503/#522/#571: install the exact Torch 2.5.1 / Transformers 5.16.1 /
+  # #502/#503/#522/#571/#651: install the exact Torch 2.6.0 / Transformers 5.16.1 /
   # trl 0.29 / PEFT 0.20 support stack and Plotext 6.0.0 so pip cannot silently
   # upgrade the cell, then assert every pinned compatibility boundary.
   transformers-floor:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,6 +188,26 @@ jobs:
               )
               print(f"{name}={got} (matches constraints)")
           PY
+      - name: Prove the pinned torch actually supports trl's FSDP2 path
+        run: |
+          python - <<'PY'
+          import torch
+          import torch.distributed.fsdp as fsdp
+
+          assert hasattr(fsdp, "FSDPModule"), (
+              "torch.distributed.fsdp.FSDPModule is absent — this is the #651 "
+              "root cause; the declared torch floor does not support trl 0.29"
+          )
+          import trl
+          from trl import DPOTrainer, KTOTrainer
+
+          print(
+              "trl",
+              trl.__version__,
+              "imports on torch",
+              torch.__version__,
+          )
+          PY
       - name: Run Transformers floor compatibility guard
         run: >
           pytest -o addopts= --no-cov -q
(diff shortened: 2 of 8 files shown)


MakazhanAlpamys:
@Samearth17 — **the consistency work is complete and correct. But the thing #651 exists to prevent is still not done, and it is two lines away.**

## The decisive finding: nothing in CI imports trl at torch 2.6.0

#651's mechanism is specific: at torch 2.5.1 `torch.distributed.fsdp.FSDPModule` does not exist, so `trl 0.29.1` cannot import, and 113 tests fail from that single root. I re-confirmed the premise on this box just now:

```
local torch        : 2.5.1+cu121
FSDPModule present : False
```

So the claim this PR makes is "torch 2.6.0 has it." The `transformers-floor` job looks like it proves that — it installs `torch==2.6.0` and asserts the installed version matches, which is a genuinely good guard. **But then it runs five test files, and not one of them imports trl:**

| file the floor job runs | imports trl? |
|---|---|
| `test_transformers_floor_compat.py` | no — pure `ast` / `re` / `yaml` static analysis |
| `test_issue571_qwen4_exp.py` | no |
| `test_plotext_compat.py` | no |
| `test_bugfixes.py::TestDiffModelLoading` | **no** |
| `test_issue636_torch_floor.py` | no — repo-text only |

`test_bugfixes.py` does contain `import trl` and `import trl.trainer.grpo_trainer`, at lines **692-693** — but they are indented inside a function, and the job selects only `::TestDiffModelLoading`, which begins at line 135. Those imports never execute.

`pip check` cannot cover for it either: **trl declares no torch dependency whatsoever**, which is precisely why pip could not catch #651 in the first place.

So the green `transformers-floor` on this PR proves that torch 2.6.0 *installs* alongside the stack. It does not prove that trl imports on it, and that is the entire claim. This is the failure mode I wrote into the v0.74.0 notes and I am not going to walk past it in a PR that cites the issue:

> 2.5.1 was measured to fail and **2.6 was not measured to work** — replacing one unverified bound with another is the mistake the issue exists to avoid.

**And the PR body does not close the gap.** "Full suite passes locally: 20,279 passed, 108 skipped" is the strongest-sounding line in it, and it is silent on which torch produced it. If it was the newest torch rather than 2.6.0, it says nothing about the floor. If it genuinely was 2.6.0, say so and paste `python -c "import torch; print(torch.__version__)"` next to it — that one line would change my verdict.

### The fix, and it is small

Add a step to the `transformers-floor` job, after the version assertion:

```yaml
      - name: Prove the pinned torch actually supports trl's FSDP2 path
        run: |
          python - <<'PY'
          import torch.distributed.fsdp as fsdp
          assert hasattr(fsdp, "FSDPModule"), (
              "torch.distributed.fsdp.FSDPModule is absent — this is the #651 "
              "root cause; the declared torch floor does not support trl 0.29"
          )
          import trl                              # noqa: F401
          from trl import DPOTrainer, KTOTrainer  # noqa: F401
          print("trl", trl.__version__, "imports on torch", __import__("torch").__version__)
          PY
```

That turns the job from "2.6.0 installs" into "2.6.0 works", it runs in seconds, and it is a standing guard rather than a one-off measurement — which is exactly what #651's fourth acceptance criterion asks for. Without it, the next person raising or lowering this floor gets the same green on the same non-evidence.

## What is genuinely right

**The three-numbers-in-three-files problem is solved.** `pyproject.toml:61` (`torch>=2.6.0`), `doctor.py:25` (`("torch", "torch", "2.6.0", True)`) and `.github/constraints/transformers-floor.txt` all agree, and `test_issue636_torch_floor.py` pins the pyproject-to-doctor equality as repo text. That is #651's stated failure mode — "two numbers in two files that nobody compares" — and it is closed.

**The version assertion in the floor job is a good guard on its own terms.** Asserting `md.version(name) == want` for each pin, with an error message naming the three ways it can go wrong ("pip silently upgraded, ignored -c, or constraints drifted"), is better than trusting `-c` to have been honoured.

**All eight removed test lines audited, none weakened.** Each is an in-place update of the `2.5.0` literal to `2.6.0` plus a rewritten rationale, and the new message correctly says transformers forces `>=2.5` *while* trl 0.29 needs 2.6 — which is a more accurate account of why the floor binds than the one it replaces.

## Blocking, aside from the measurement

**The changelog fragment is named after the issue, not the pull request.** It is `changelog.d/0.74.0/651.fixed.md` and should be `717.fixed.md`; its text says `#651` where the convention wants `#717`; and it carries **no author credit**. The convention in that directory is `(#651 by @Samearth17 in #717)`. The PR-number rule exists precisely so two PRs on one issue cannot collide on one filename, and we have had that happen twice this week.

## Non-blocking

The constraints comment now reads *"Torch 2.6.0 deliberately exercises Qwen4-Exp's legacy int64-only scatter path."* That sentence was written about 2.5.1. Is it still true at 2.6.0 — i.e. does 2.6.0 still take the legacy path, or did the version get swapped without the claim being re-checked? If the claim no longer holds, the comment should say what 2.6.0 exercises instead; if it does hold, no change needed, but it is worth a sentence in the PR saying you confirmed it rather than carried it forward.

Add the trl import to the floor job and rename the fragment, and this has my approval — the rest of it is exactly the shape #651 asked for.

MakazhanAlpamys:
Merged as `0be7808`. Thank you — and the second round is what made this mergeable rather than merely correct.

**The objection this PR had to answer.** v0.74.0 deliberately shipped the evidence instead of a bump, on the grounds that 2.5.1 was measured to fail and **2.6 was not measured to work** — replacing one unverified bound with another is the mistake #651 exists to avoid. Your first version raised the numbers and pointed the pinned-stack job at torch 2.6.0, and I blocked because that green proved *installation*, not *compatibility*: not one of the five test files that job ran imported trl, and the entire defect is an import.

Your new step fixes exactly that:

```yaml
assert hasattr(fsdp, "FSDPModule"), "...this is the #651 root cause..."
import trl
from trl import DPOTrainer, KTOTrainer
```

**And it is demonstrably discriminating, which I checked rather than assumed.** This machine still runs the old floor, so I ran your step's own logic against it:

```
local torch: 2.5.1+cu121
hasattr(fsdp, 'FSDPModule'): False        <- your assert fires
trl import FAILS: RuntimeError ...        <- and the import fails
```

So the guard rejects precisely the version it exists to reject, while the job is green on the version being adopted. That is the difference between a floor bump and a proven floor, and it is the whole reason this can merge now.

**#651's fourth criterion is met in both directions.** The failure mode it names is "two numbers in two files that nobody compares", and I mutated each independently:

```
pyproject floor -> 2.5.0, CI pin left at 2.6.0   -> 1 failed
CI constraint  -> 2.5.1, pyproject left at 2.6.0 -> 1 failed
```

Neither can drift alone.

**Eight removed test lines, all audited**: the assertions pinning `torch>=2.5.0`, each replaced by its 2.6.0 equivalent. Nothing weakened, nothing deleted, and `soup doctor`'s second literal copy of the floor moved with it and is still pinned against the declared one.

The changelog fragment is right now too — `717.fixed.md`, PR number in the filename and the text, credited.

Credited in `CONTRIBUTORS.md` (`f87d5a8`). That is two of yours merged — #648 and this — and this one unblocks the preference trainers for anyone who was pinned at 2.5.x.

PR #726, opened 2026-09-06
perf(multipack): place FFD items through a segment tree, not a linear scan

### LINKED ISSUE #694: Optimize multipack FFD placement from O(N²) to O(N log N)
## Summary

`ffd_bin_pack()` linearly scans every existing bin to find the first one with enough remaining capacity. Its worst-case placement phase is O(N²), which makes multipack preparation unnecessarily expensive on large datasets.

This is a setup-time optimization, not a claim about model tokens/second or GPU memory.

## Current behavior

On current `main` (`ef2e02b51a1b34546823edce55715d28234805cb`), items are sorted by decreasing length and placed with:

```python
for bin_idx, remaining in enumerate(bin_remaining):
    if length <= remaining:
        ...
        break
```

When every item is larger than half the bin budget, no previous bin fits and the function performs `N(N-1)/2` capacity checks. The source already acknowledges this O(N²) bound and caps inputs at one million items.

`MultipackBatchSampler` runs this packing once during construction, so the impact is initialization latency.

## Prototype and equivalence checks

A segment tree storing the maximum remaining capacity in each range can locate the **leftmost** fitting bin in O(log N), preserving First-Fit-Decreasing rather than changing to best-fit.

The prototype was compared with the current implementation on:

- 6,461 valid inputs: identical bins, original indices, and order;
- 7 invalid-input cases: identical exception behavior.

Independent local CPU medians on macOS/arm64, Python 3.12.2:

| synthetic case | current | segment tree | speedup for this function |
|---|---:|---:|---:|
| 3,000 × length 600, budget 1,024 | 61.95 ms | 3.76 ms | 16.48× |
| 6,000 × length 600, budget 1,024 | 260.34 ms | 8.21 ms | 31.71× |
| 6,000 uniform lengths 1..1,024 | 123.33 ms | 9.63 ms | 12.80× |

The supplied audit's separate Linux run measured 22.2×–60.8× on the same function-only cases. No end-to-end training speedup, CUDA performance, or RSS improvement was measured.

## Recommended implementation

Use a pure-Python max segment tree (or equivalent indexed capacity structure):

1. keep Soup's existing stable decreasing-length sort;
2. query the leftmost bin whose stored maximum is at least the item length;
3. update that leaf's remaining capacity after placement;
4. append a new leaf when no bin fits.

The total cost becomes O(N log N), including the existing sort, without adding a compiled dependency.

## Acceptance criteria

- [ ] Output is byte-for-byte/equality identical to current FFD bins on property-generated valid inputs, including tied lengths.
- [ ] Existing validation and exception types/messages remain unchanged.
- [ ] Worst-case 6,000-item benchmark improves materially and includes a regression threshold generous enough for CI noise.
- [ ] The implementation remains deterministic across Python 3.10–3.12 and all supported OSes.
- [ ] Documentation describes the gain as multipack preparation latency only.

## Duplicate check

#65 covers live sampler wiring, #80 distributed DataLoader behavior, and #561 the batch-budget bug. None covers the quadratic first-fit lookup or an equivalent indexed implementation. Searches across open/closed issues and PRs found no matching report or open fix.

dchaudhari7177:
Closes #694. Implements the recommended approach: a pure-Python max segment tree over the bins' remaining capacity.

**Left-first descent is the whole correctness argument.** The tree stores each subtree's maximum remaining capacity, so a descent skips any subtree that cannot hold the item; by always trying the left child first it lands on the **leftmost** bin that fits, which is precisely the bin the linear scan returned. That keeps this first-fit. A best-fit variant would still produce valid, plausible-looking bins, which is why the tests below compare against the old code rather than against hand-written expectations.

### Acceptance criteria

**Output identical on property-generated valid inputs, including tied lengths.** The pre-#694 loop is kept verbatim in the test file as `linear_scan_ffd` and everything is compared against it. 3,000 randomised cases across budgets 8–1024 and sizes 0–120: **zero mismatches.** Ties are exercised deliberately — a 3-value alphabet and all-identical inputs — because that is where placement rules diverge and where a best-fit regression would otherwise hide. Growth boundaries (0, 1, 2, 3, 17, 64, 65, 128, 129 bins) are parametrised separately, since an off-by-one in the tree's doubling rebuild would only appear at powers of two.

**Validation and exception types/messages unchanged.** All seven paths checked:

```
max_len=0    -> ValueError: max_len must be positive, got 0
max_len=True -> TypeError:  max_len must not be bool, got True
[0,1]        -> ValueError: lengths[0] must be positive, got 0
[9], ml=8    -> ValueError: lengths[0]=9 exceeds max_len=8
[1.5]        -> TypeError:  lengths[0] must be int, got float
[True]       -> TypeError:  lengths[0] must not be bool, got True
[1,"x"]      -> TypeError:  lengths[1] must be int, got str
```

**Worst-case benchmark, with a CI-generous threshold.** Local medians, function only:

| case | current | segment tree | speedup |
|---|---:|---:|---:|
| 3,000 × 600, budget 1,024 | 389.28 ms | 21.01 ms | 18.5× |
| 6,000 × 600, budget 1,024 | 1616.61 ms | 53.50 ms | 30.2× |
| 6,000 uniform 1..1,024 | 738.27 ms | 55.18 ms | 13.4× |

Close to the figures in the issue. The committed regression test asserts only a **3×** margin on the 3,000-item case — a reintroduced linear scan cannot clear that at this size, and it leaves plenty of room for a loaded runner. It asserts the shape of the change, not a performance number.

**Deterministic across versions and platforms.** Flat `list[int]`, no floats, no dict ordering, no compiled dependency.

**Documentation scoped to preparation latency.** The docstring and commit say multipack preparation only. No claim about tokens/second, GPU memory or RSS, none of which was measured.

### One deviation, flagged rather than slipped in

The criterion says messages remain unchanged, and I changed one: the `_MAX_FFD_ITEMS` error said `(algorithm is O(N^2) worst-case)`, which is now false. I kept the cap and the `ValueError` — a caller passing ten million lengths still wants a loud error rather than a multi-gigabyte list — and dropped only the false parenthetical, updating the comment above it to say why the cap survives the complexity change. `test_ffd_rejects_too_many_items` matches on `"too many items"` only, so its behaviour is unmodified. Happy to restore the wording if you would rather keep it byte-identical.

### Verification

- `tests/test_multipack_ffd_placement.py`: 57 pass.
- Existing `test_multipack_sampler.py`, `test_multipack_invariants.py`, `test_multipack_config.py` plus the new file: **144 pass, 0 fail.**
- `ruff check`: clean.
- Mutation-checked: flipping the descent to try the right child first (turning first-fit into something best-fit-like) fails **25** of the new tests, so the equivalence assertions genuinely bite.

### DIFF
--- a/src/soup_cli/utils/multipack_sampler.py
+++ b/src/soup_cli/utils/multipack_sampler.py
@@ -67,9 +67,11 @@ MULTIPACK_ARCHITECTURES: frozenset[str] = frozenset({
 })
 
 
-# Worst-case FFD complexity is O(N^2). Cap N to prevent a crafted dataset
-# from pinning a CPU. 1M samples is ~3 orders of magnitude beyond typical
-# fine-tuning workloads.
+# Cap N to prevent a crafted dataset from pinning a CPU or exhausting memory.
+# 1M samples is ~3 orders of magnitude beyond typical fine-tuning workloads.
+# The cap predates the indexed placement below and is kept: packing is now
+# O(N log N) rather than O(N^2), but a caller that hands over ten million
+# lengths still wants a loud error rather than a multi-gigabyte list.
 _MAX_FFD_ITEMS: int = 1_000_000
 
 
@@ -80,6 +82,78 @@ def _check_int(name: str, value: object) -> int:
     if not isinstance(value, int):
         raise TypeError(f"{name} must be int, got {type(value).__name__}")
     return value
+
+
+class _CapacityTree:
+    """Remaining bin capacities, indexed so first-fit is O(log N).
+
+    A max segment tree over a growing array. Each internal node stores the
+    maximum remaining capacity in its subtree, so a descent can skip any
+    subtree that cannot hold the item, and by always trying the left child
+    first it lands on the *leftmost* bin that fits. That is the bin the
+    original linear scan returned, which is what keeps this first-fit rather
+    than turning it into best-fit.
+
+    Deliberately a flat list of ints and pure Python: the packing has to be
+    identical on every supported interpreter and OS, and no compiled
+    dependency is worth that risk for a setup-time cost.
+    """
+
+    __slots__ = ("_tree", "_capacity", "_size")
+
+    def __init__(self) -> None:
+        self._capacity = 1  # number of leaf slots; a power of two
+        self._tree = [0, 0]  # 1-indexed heap layout; index 0 unused
+        self._size = 0  # number of bins actually in use
+
+    def append(self, remaining: int) -> None:
+        """Add a new bin with ``remaining`` free space."""
+        if self._size == self._capacity:
+            self._grow()
+        self._set(self._size, remaining)
+        self._size += 1
+
+    def take(self, bin_idx: int, length: int) -> None:
+        """Consume ``length`` from bin ``bin_idx``."""
+        self._set(bin_idx, self._tree[self._capacity + bin_idx] - length)
+
+    def leftmost_with_room(self, length: int) -> int | None:
+        """The lowest bin index whose remaining capacity is >= ``length``.
+
+        ``None`` when no open bin fits, which is the caller's signal to start
+        a new one -- the same condition as the old scan falling off the end.
+        """
+        if self._size == 0 or self._tree[1] < length:
+            return None
+        node = 1
+        while node < self._capacity:
+            left = node * 2
+            # Left first: this is what makes the result the *leftmost* fit.
+            node = left if self._tree[left] >= length else left + 1
+        return node - self._capacity
+
+    def _set(self, leaf: int, value: int) -> None:
+        node = self._capacity + leaf
+        self._tree[node] = value
+        node //= 2
+        while node >= 1:
+            self._tree[node] = max(self._tree[node * 2], self._tree[node * 2 + 1])
+            node //= 2
+
+    def _grow(self) -> None:
+        """Double the leaf capacity, rebuilding the tree around the old leaves.
+
+        Amortised O(1) per append: the rebuild costs O(N) and happens after N
+        appends. Unused leaves hold 0, which is below every valid item length
+        (they are validated positive above), so they can never be selected.
+        """
+        old_leaves = self._tree[self._capacity : self._capacity + self._size]
+        self._capacity *= 2
+        self._tree = [0] * (2 * self._capacity)
+        for offset, value in enumerate(old_leaves):
+            self._tree[self._capacity + offset] = value
+        for node in range(self._capacity - 1, 0, -1):
+            self._tree[node] = max(self._tree[node * 2], self._tree[node * 2 + 1])
 
 
 def ffd_bin_pack(lengths: Sequence[int], max_len: int) -> list[list[int]]:
@@ -110,7 +184,7 @@ def ffd_bin_pack(lengths: Sequence[int], max_len: int) -> list[list[int]]:
     if len(lengths) > _MAX_FFD_ITEMS:
         raise ValueError(
             f"too many items for FFD bin-packing: {len(lengths)} > "
-            f"{_MAX_FFD_ITEMS} (algorithm is O(N^2) worst-case)"
+            f"{_MAX_FFD_ITEMS}"
         )
 
     # Validate each length up-front so we fail loudly before packing.
@@ -131,18 +205,22 @@ def ffd_bin_pack(lengths: Sequence[int], max_len: int) -> list[list[int]]:
     )
 
     bins: list[list[int]] = []
-    bin_remaining: list[int] = []  # parallel to bins: free space in each
+    # A max segment tree over the bins' remaining capacity, replacing the
+    # linear scan this loop used to do. `_leftmost_bin_with_room` returns the
+    # same bin the scan would have found, so this is first-fit exactly as
+    # before -- not best-fit, and not any other order. The only thing that
+    # changes is how long it takes to find it: O(log N) per item rather than
+    # O(N), which removes the N(N-1)/2 worst case that appears when every item
+    # is larger than half the budget and no earlier bin can ever fit.
+    tree = _CapacityTree()
     for orig_idx, length in indexed:
-        placed = False
-        for bin_idx, remaining in enumerate(bin_remaining):
-            if length <= remaining:
-                bins[bin_idx].append(orig_idx)
-                bin_remaining[bin_idx] = remaining - length
-                placed = True
-                break
-        if not placed:
+        bin_idx = tree.leftmost_with_room(length)
+        if bin_idx is None:
             bins.append([orig_idx])
-            bin_remaining.append(max_len - length)
+            tree.append(max_len - length)
+        else:
+            bins[bin_idx].append(orig_idx)
+            tree.take(bin_idx, length)
 
     return bins
 
--- a/tests/test_multipack_ffd_placement.py
+++ b/tests/test_multipack_ffd_placement.py
@@ -1,0 +1,162 @@
+"""`ffd_bin_pack` places items through an indexed structure, not a scan (#694).
+
+The packing is unchanged: same bins, same order, same exceptions. Only the cost
+of finding each bin changed, from O(N) per item to O(log N). These tests exist
+to keep that "unchanged" honest, because a first-fit implementation that
+quietly becomes best-fit still produces plausible-looking bins.
+
+The reference implementation below is the exact pre-#694 loop. Comparing
+against it rather than against hand-written expectations is what makes the
+equivalence claim checkable: a hand-written case pins what someone believed
+the old code did, and the whole risk here is that the belief is wrong.
+"""
+
+from __future__ import annotations
+
+import random
+import statistics
+import time
+from typing import Sequence
+
+import pytest
+
+from soup_cli.utils.multipack_sampler import ffd_bin_pack
+
+
+def linear_scan_ffd(lengths: Sequence[int], max_len: int) -> list[list[int]]:
+    """The placement loop `ffd_bin_pack` used before #694, verbatim."""
+    indexed = sorted(enumerate(list(lengths)), key=lambda pair: pair[1], reverse=True)
+    bins: list[list[int]] = []
+    bin_remaining: list[int] = []
+    for orig_idx, length in indexed:
+        placed = False
+        for bin_idx, remaining in enumerate(bin_remaining):
+            if length <= remaining:
+                bins[bin_idx].append(orig_idx)
+                bin_remaining[bin_idx] = remaining - length
+                placed = True
+                break
+        if not placed:
+            bins.append([orig_idx])
+            bin_remaining.append(max_len - length)
+    return bins
+
+
+@pytest.mark.parametrize("seed", range(25))
+def test_placement_is_identical_to_the_linear_scan(seed: int) -> None:
+    rng = random.Random(seed)
+    for _ in range(40):
+        max_len = rng.choice([8, 16, 64, 128, 1024])
+        count = rng.randint(0, 120)
+        lengths = [rng.randint(1, max_len) for _ in range(count)]
+        assert ffd_bin_pack(lengths, max_len) == linear_scan_ffd(lengths, max_len)
+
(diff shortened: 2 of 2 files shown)


MakazhanAlpamys:
@dchaudhari7177 — **behavioural equivalence holds. I checked it on 16,359 cases against `main`'s own function, not against the copy in your test file, and found zero mismatches. Blocking only on the changelog fragment.**

Head `3fbba67` verified. Worktree with `PYTHONPATH` pinned inside it; every number below is from a locally-built merge onto current `origin/main`, not from the PR head alone.

## Equivalence — the question that decides this PR

A first-fit-decreasing packer's output depends on *which* bin it picks when several fit. If the segment tree ever picks a different one, every existing multipack user's batches change silently. So I extracted `origin/main`'s `ffd_bin_pack` verbatim via `git show` as the reference — deliberately **not** the copy vendored into your test file, since that copy is part of what is under test — and compared bins, bin order, index order **and** the exception `(type, message)` pair:

```
cases=16359 mismatches=0
```

That sweep: 3,000 random valid inputs (budgets 1-1024, n 0-120); 1,500 heavy-tie cases over a 3-value alphabet; every all-identical `[v]*n` for v in 1..max_len across 7 budgets x 10 sizes including 1/2/3/17/64/65/128/129/200; 500 worst-case "every item > half budget"; ascending, descending and sawtooth inputs at 19 sizes straddling every power of two to 513; a generator input for the materialisation path; and 17 invalid-input edges (`[]`, `[0,1]`, `[1.5]`, `[True]`, `[1,"x"]`, `[-1]`, `max_len` 0 / -3 / `True` / `1.0` / `"x"`, `[9]` against `max_len=8`).

And at the level that actually reaches a training run:

```
sampler trials=200 mismatches=0    # MultipackBatchSampler batch lists
```

**No existing user's batches change.** The structural reason backs it up: unused leaves hold `0`, every length is validated `>= 1`, so an unused leaf can never satisfy `tree[leaf] >= length`; the descent takes the left child whenever it qualifies and the root gate guarantees one child does — so it lands on the leftmost qualifying leaf, which is exactly what the scan's `break` returned.

## Performance is real and the threshold has margin

```
3000 x 600, budget 1024: linear  200.4ms   tree 19.7ms   10.2x
6000 x 600, budget 1024: linear 1430.9ms   tree 65.8ms   21.7x
```

Complexity verified by doubling rather than asserted:

```
n= 25000: 0.17s
n= 50000: 0.33s  1.94x
n=100000: 0.71s  2.15x
n=200000: 1.70s  2.37x
```

About 2x per doubling, not 4x. N log N confirmed. (An earlier 8.99s reading at n=200k was `tracemalloc` overhead; clean re-measurement is 1.70s, traced peak 44.2 MB.)

The `_MAX_FFD_ITEMS` cap still holds exactly: 1,000,000 accepted, 1,000,001 raises `ValueError`. The message shortening you flagged yourself is safe — the only in-repo assertion is `match="too many items"`, and `grep -rn "O(N^2) worst-case" tests/ docs/` returns nothing.

## Mutations — 6 applied, 6 killed

| mutation | result |
|---|---|
| descent tries the **right** child first (first-fit -> best-fit-ish) | 25 failed |
| `tree[left] >= length` -> `> length` | 52 failed |
| root gate `< length` -> `<= length` | 36 failed |
| `_grow` stops rebuilding internal nodes | 25 failed |
| `take()` subtracts nothing | 43 failed |
| **full revert to the verbatim pre-#694 linear scan** | 1 failed |

The last is the important one and your design is right: a straight revert is behaviourally indistinguishable, so exactly one test can catch it — the performance assertion. The 25 on the first mutation matches your PR body's claim exactly.

Zero deleted test lines. Ruff clean. `243 passed` on `-k "multipack or packing or neat"`. `Closes #694` live.

## BLOCKING — changelog fragment

No `changelog.d/0.74.0/726.*.md`. This is a `src/` change and I told you on the #727 thread that #726 and #729 are not exempt. Add `changelog.d/0.74.0/726.changed.md` (or `.fixed`), naming `#726` in the text and crediting `by @dchaudhari7177`. Do not edit `CHANGELOG.md`.

No `docs/*.md` change is fine — #694 scopes the claim to multipack preparation latency, your code comments say exactly that, and the fragment is the right place for the user-facing sentence.

## MEDIUM — the two tests whose docstrings claim to be the best-fit discriminators do not discriminate

`test_ties_break_the_same_way` says it is *"the case a best-fit regression would survive... with heavy ties they diverge"*, and `test_first_fit_not_best_fit` says it is *"a case where the two rules disagree, stated directly."* Under the right-child-first mutation — which **is** the best-fit-shaped regression:

```
pytest -k "ties or first_fit_not_best_fit or worst_case_shape or growth_boundaries"
21 passed, 36 deselected     # with the mutation applied
```

All 21 pass. Only `test_placement_is_identical_to_the_linear_scan` kills it, and it kills 25 of 25. The reason is concrete: your ties alphabet `[1, max_len//2, max_len]` and the `[10,6,5,4,4]` case never present **two simultaneously-open bins that both have room**, so left-first and right-first cannot diverge on them.

The suite is adequate **today**. The risk is later: someone reads "we have named tests for the first-fit-vs-best-fit distinction", trims the randomized equivalence test as redundant, and the suite silently stops detecting a best-fit regression. Either correct the two docstrings to say what they actually cover, or give them a real two-candidate case — a shape where bin 0 and bin 1 are both open with room and the rules pick differently.

## Two small ones

- `multipack_sampler.py:208` refers to `_leftmost_bin_with_room`; the method is `leftmost_with_room`.
- The perf test is a wall-clock assertion in CI. Median-of-3 against a measured 10.2x with a 3x bar makes a flake unlikely, and the revert mutation shows nothing else catches a regression — so I accept it. Recording it because it is the one test here that can go red without a code change, on a loaded macOS or Windows runner.

## CI staleness

Green against `60fe73d`, four commits behind current `main`. I built the merge locally — clean, and every number above is from that merged tree.

Add the fragment and this merges; the docstring item I would take in the same push but will not hold you on.

MakazhanAlpamys:
@dchaudhari7177 — **approved on the substance. One blocking item and it is paperwork.**

The question that decides a performance change like this is whether it packs *identically*, because altering which item lands in which bin silently changes every multipack training run. Three independent lines of evidence say it does.

**1. Your own test is the right shape.** Keeping the pre-#694 loop verbatim as `linear_scan_ffd` and comparing against it — rather than against hand-written expectations — is exactly right. I diffed that reference against `git show origin/main:...multipack_sampler.py` and it is faithful: same sort key, same `enumerate`, same `reverse=True`, same inner scan and `break`.

**2. An independent differential harness**, 2,583 cases, **zero divergences**: 600 uniform-random vectors across 16 budgets including primes; 60 all-items-over-half-budget; 200 tie-heavy tiny-alphabet; all-identical at 7 budgets x 70 sizes; big-then-small shuffled; power-of-two boundaries 0..1025; exact-fit chains driving `remaining` to 0; and three large cases up to 5,000 items. Bin-for-bin list equality.

**3. The leftmost-fit claim, checked structurally rather than taken from the body.** Brute-forced the descent against `next(i for i, r in enumerate(model) if r >= q)` over 3,000 randomised append/take/query trials plus a full 55-value sweep after each: **0 mismatches**. So it is verified as first-fit *in index order*, not merely as "produces valid bins".

**Seven mutations, six killed:** right-child-first descent (25 failed — exactly the count your body claims), `>=` to `>` (41), `_grow` copying one leaf too few (25), `_grow` skipping the lowest internal level (25), dropping the root guard (50), `take()` reading the root instead of the leaf (28).

The seventh — `_grow` skipping the root rebuild — survives and is a **genuine equivalent mutant**, not a gap: `_grow` is only called from `append`, which immediately calls `_set` on the new leaf, and `_set` walks leaf-to-root recomputing every ancestor, so node 1 is repaired before any query. The non-equivalent version of the same idea is the level-skipping mutation, and that one dies.

**The speedup holds far better at realistic N than at the benchmark size.** Your test uses 3,000 with a 3x bar, which is the right call for CI. Measured here:

```
 3,000 x 600, budget 1024      228 ms ->  16 ms    14.2x
 6,000 x 600, budget 1024      960 ms ->  54 ms    18.0x
50,000 uniform 1..2048      44,980 ms -> 564 ms    79.8x
```

300,000 items at budget 4096 packs in 4.34 s; the old scan would have been about half an hour.

`_MAX_FFD_ITEMS` is still 1,000,000 and still raises. You flagged the dropped `(algorithm is O(N^2) worst-case)` parenthetical in the message rather than slipping it in — correct handling, and the existing test matches on `"too many items"` so it is unaffected. 187 passed on the multipack surface against 130 on `main`; zero test lines removed.

## Blocking

**No changelog fragment.** `changelog.d/0.74.0/726.perf.md` (or `.changed.md`), named after the **pull request**, mentioning `#726` in the text, crediting `by @dchaudhari7177`. Rules in `changelog.d/README.md`. CI cannot catch a missing one — it validates fragments that exist.

## Two nits, neither blocking

- `src/soup_cli/utils/multipack_sampler.py:209` names `` `_leftmost_bin_with_room` `` in a comment; the method is `leftmost_with_room`.
- Your body says "3,000 randomised cases"; the file performs 1,800 plus the fixed cases. Immaterial — my own 2,583 back the conclusion — but worth correcting since the rest of your numbers check out exactly.

Add the fragment and I merge.

MakazhanAlpamys:
@dchaudhari7177 — merged as `9ab3b5e0`. **I wrote the changelog fragment and one doc line myself and merged, credited to you.** That is the offer I made on #729 and #727 this morning, applied here without waiting for your answer because this PR had been approved on substance since 2026-09-07 and blocked on nothing but paperwork for two days. If you would rather have written them, say so and I will hand the next one back.

Both are in `2d8fcc4`: `changelog.d/0.74.0/726.changed.md`, and the correction to `docs/peft-and-efficiency.md:115`, which still read *"the FFD packer caps at 1M items (algorithm is O(N^2) worst-case)"*. That is the same claim you correctly deleted from the source error message — the docs copy was missed, and it was the last unmet acceptance criterion on #694, which asks for the gain to be described as preparation latency.

## What I verified at merge

The question that decides a change like this is not whether the result is *a* valid packing but whether it is the *same* packing, because altering which item lands in which bin silently changes every multipack training run. I checked it against `main`'s own `ffd_bin_pack` extracted from `origin/main` — deliberately not the copy in your test file, since that copy is yours and matching it proves nothing:

```
6,000 randomized cases, full bin assignment compared   0 mismatches
of which 5,384 contain repeated lengths                0 mismatches
```

Ties are over-represented on purpose: they are where a data-structure swap most plausibly diverges. Budgets 1-4096 including 7/17/31/129/257, item counts across 0-39 plus the growth boundaries 63/64/65/127/128/129/255/256/257, five length distributions including all-identical and all-over-half-budget.

Four mutations, four killed. The one that matters:

```
descent returns the last fitting bin instead of the first   25 failed, 32 passed
  (FFD turned into a best-fit variant)
```

The other three: an off-by-one leaving the root stale (51 tests), the tree ignoring capacity so a bin can overflow (53), and reverting to a linear scan (1).

**That last one is worth naming, because it looks like a coverage gap and is not.** A behaviour-preserving revert is caught only by the timing test — and no equivalence assertion *can* detect a pure performance regression, so that single test is the only thing standing between this and a silent return to O(N^2). You wrote it as a ratio measured inside one process rather than as an absolute threshold, which is the shape that survives a slow CI runner. It passed on all nine cells.

Measured here: 5.5x at N=1,000, 37x at N=10,000, 96x at N=30,000 equal-length, 150x at N=30,000 uniform. Scaling per 10x N drops from 104-130x to 8.9-13.9x, so the complexity claim is measured rather than asserted.

## Two small things I left alone

`multipack_sampler.py:209` has a comment referencing `_leftmost_bin_with_room`; the method is `leftmost_with_room`. And the cap error is now a needlessly split f-string that fits on one line after your deletion. Neither is worth a commit on its own — fold them in if you touch the file.

Zero test lines deleted or weakened. Ruff clean. 14 of 14 SUCCESS on `3fbba676` itself, `head_sha` confirmed equal to the head I merged.

You still have three open PRs where I am the queue — #727, #729 and #742. The same offer stands on all three.

Issue #702, opened 2026-09-05
Streaming interleave: over + val_split still puts the same row in train and val


MakazhanAlpamys:
Found while reviewing #701, which fixes the same defect on the two **eager** interleave paths. Filed separately because the streaming path needs a different mechanism, which is a fair reason to defer it rather than block that PR.

## What happens

`data.streaming: true` with `interleave: over` puts the same row in both the training and the validation split.

Measured on `019c324`, and identically on #701's branch:

```
streaming: true, A=100 B=10, strategy=over, val_split=0.1
  val rows that also appear in train: 10 / 20
  CONTROL probs / concat / under:      0
```

## Mechanism

`interleave_datasets(..., stopping_strategy="all_exhausted")` recycles the shorter stream to exhaust the longer one, so the materialised result contains duplicates of the smaller source. `_finalize` at `loader.py:742` then takes a **positional tail slice** for `val_split`, and a row and its recycled copy are separate entries, so the slice can put one on each side.

Structurally identical to #680. The difference is where the duplication happens — `_cycle_to` on the eager path, `all_exhausted` inside `datasets` on the streaming one.

## Why #701's fix does not transfer

#701 carves `val_split` out **per source before** padding. That works because the eager path has the rows in hand and can count them. A streaming source is not countable ahead of time, so there is nothing to take a fraction of before interleaving starts. The fix needs a different shape — reserving a val stream per source before `interleave_datasets`, tracking which rows have been emitted, or refusing the combination.

## Severity

Silent, and the failure looks like success: validation loss on the recycled source measures memorisation, so the run reports a suspiciously good number rather than an error. Anyone using `data.streaming: true` with `interleave: over` and a non-zero `val_split` is affected. `probs`, `concat` and `under` are clean on this path.

There is also no escape hatch: `DataConfig` has no `data.val` field (`[f for f in DataConfig.model_fields if 'val' in f]` -> `['val_split', 'eval_on_each_dataset']`), so an independently supplied validation set is not available for local interleave. The only workaround today is avoiding `over` on the streaming path.

## Documentation is currently misleading about this

`docs/data.md` opens the strategy table with *"The strategy names mean the same thing on the streaming path as on the local path"*, and after #701 lands the paragraph below it will describe a fix scoped to the eager path. A reader will reasonably conclude streaming is covered. #701 is adding a clause naming this exclusion; this issue is the thing that clause points at.

## Acceptance criteria

[...]

PR #727, opened 2026-09-06
bench(mlx): pin every field that moves the harness step count

### LINKED ISSUE #716: Benchmark harness step count depends on a schema default it never pins
## The gap

`benchmarks/harness/mlx_sft_smoke.py` writes its own training YAML and pins `epochs`, `lr`, `batch_size: 1` and the LoRA block — but not `gradient_accumulation_steps`. It therefore inherits whatever the schema default happens to be, currently `4`.

That harness backs a published record, `benchmarks/run-m1-8gb-mlx-sft.md`. So the number of optimizer steps behind a published throughput figure depends on the default value of a field the harness never mentions.

## How it surfaced

#696 makes the MLX wrapper round `iters` down to a whole number of accumulation groups, which is the correct fix for #684 — a run that accumulates a partial group and never applies it is the bug. On the merged tree that turned the #703 harness test's 10 iterations into 8 and failed 183 of its 198 parametrised cases.

**The published record is not affected**, and this was checked rather than assumed: every published run used **48 rows, 1 epoch -> 48 iterations** (`run-m1-8gb-mlx-sft.md:55`, reproduce line at `:271`), including the warm-cache re-run at `:206`. `48 % 4 == 0`, so the rounding is a no-op on all of them.

**But it survives by coincidence.** 50 rows, or a schema default of 5, and every figure in that table would have been measured over a different step count than the record claims, with nothing in the file saying so.

## What to do

- Pin, in the harness YAML, **everything that changes the step count**, each with a short stated reason — starting with `gradient_accumulation_steps: 1` and including `batch_size`, which is already pinned but for no stated reason.
- Add the resolved `gradient_accumulation_steps` to the record's **Reproducing** section, for the same reason the model id and row count are stated there.
- Immediate unblocking of #696 is a single line in the harness and is being handled inside that PR; this issue is the wider pass and should not grow #696.

## Acceptance criteria

- [ ] The harness YAML pins every field that affects the step count, each with a reason in a comment
- [ ] Changing a schema default cannot move the harness's iteration count — demonstrate it, e.g. by asserting the resolved count against the row count rather than trusting the default
- [ ] `run-m1-8gb-mlx-sft.md`'s Reproducing section states the resolved `gradient_accumulation_steps`
- [ ] No published figure changes; if any would, say so explicitly rather than editing the number
- [ ] `tests/test_issue23_mlx_harness_bridge.py` still passes, and its `(10, 10)` assertion still means what it says

## Notes

Found while reviewing #696 (@AmirF194). The harness is @Srinivasan8888's from #663, and they identified the coincidence, settled the published-record question, and proposed this pass themselves — filed here rather than left as an agreement in a PR thread so it has a home.

dchaudhari7177:
Closes #716.

### Acceptance criteria

**The harness YAML pins every field that affects the step counts, each with a reason.** `gradient_accumulation_steps: 1` added; `batch_size: 1` gains the reason it never carried. Both comments say *what* the field does to the counts, not just that it is pinned.

**Changing a schema default cannot move the run: demonstrated and tested.** Rather than trusting the pins to survive future edits, the harness resolves iterations and optimizer updates out of the *loaded config*, the same way `MLXSFTTrainerWrapper` does: `epochs * ceil(rows / batch_size)`, rounded down to a whole accumulation group. If either count is not `rows * epochs`, it **exits non-zero (`sys.exit` with a message on stderr) before training starts**:

```
steps         : 48 iterations / 48 optimizer updates  (batch_size=1, grad_accum=1)
```

With the `gradient_accumulation_steps` pin deleted (schema default 4):

```
48 rows -> 48 iterations and 12 optimizer updates, dropping 0 row(s)   -> exit 1
50 rows -> 48 iterations and 12 optimizer updates, dropping 2 row(s)   -> exit 1
```

At 48 rows the old default does not change the iteration count. It groups those iterations into a quarter as many optimizer updates. At a row count not divisible by 4, #696's rounding drops rows outright (thanks @Srinivasan8888 for reading this back from the real wrapper).

**`run-m1-8gb-mlx-sft.md` states the resolved values** in the Config and Data rows, and in Reproducing.

**No published figure changes.** Every published run is 48 rows / 1 epoch. @Srinivasan8888 re-ran this branch on the M1 8 GB the record was measured on and got `Trained Tokens 2130`, byte-identical to the published Qwen2.5-0.5B row.

### Tests

- `tests/test_issue716_harness_step_guard.py` (new): `resolve_step_counts` checked against the `TrainingArgs` the real wrapper builds across six shapes, plus the abort path through the real `main()` (accum pin deleted at 48 and 50 rows, `batch_size` moved) and the passing path. Disabling only the guard fails 3 tests; restoring the old division formula fails 13.
- `tests/test_issue23_mlx_harness_bridge.py`: 198 passed.
- `changelog.d/0.75.0/727.fixed.md` added. `ruff check` clean.

Metal behaviour is covered by the M1 re-run above, not by CI.


MakazhanAlpamys:
@dchaudhari7177 — welcome, and thank you for three substantive pull requests in half an hour. **Something you could not have known applies to this one, and I would rather you heard it from me than worked it out from a thread.**

I told @Srinivasan8888 on #696 that #716 was "filed and assigned to you". **The assignment silently failed and I did not verify it**, so for five hours this issue has read as unassigned, unclaimed and `help wanted`. You did the right thing: no assignee, no claim comment, open label. The mistake is mine, not yours.

The mechanism, for the record, because it is a trap worth knowing: GitHub only lets you assign a collaborator or someone who has already interacted with that specific issue. `POST /issues/716/assignees` returned **200 with an empty `assignees` array** rather than an error, and `GET /assignees/Srinivasan8888` returns **404**. A silent success is indistinguishable from a real one unless you read it back.

## What that means for #727

**@Srinivasan8888 has until 2026-09-07 12:00 UTC to say they want it.** They found the coincidence this issue is about, settled whether the published Apple Silicon record was affected, and proposed this exact pass — so they get the refusal. After that, or the moment they pass, **#727 proceeds and I review it on its merits.**

If they do take it, I will ask them to build on your branch rather than have you redo work, and you will be credited for it. Either way you are not losing the round for having arrived correctly.

**Nothing about #726 or #729 is affected** — no claim exists on #694 at all, and on #702 the only thing said was that it was unclaimed. Both are straightforwardly yours.

## Meanwhile

Your CI was sitting in `action_required`, which is the default for a first-time contributor's first run here. **I have approved all four workflow runs**, so #726, #727 and #729 are executing now. That was on me too — nothing in the UI tells you it is waiting.

One practical note for all three while they run: this repo requires a changelog fragment for user-visible changes, at `changelog.d/0.74.0/<PR-number>.<category>.md`, named after the **pull request** rather than the issue, mentioning the same number in its text and crediting `by @dchaudhari7177`. #727 is arguably exempt as a benchmark-harness change; #726 and #729 are not. `changelog.d/README.md` has the rules. CI cannot catch a *missing* fragment, only a malformed one, so it is worth doing before review rather than after.

I will get to all three properly. Given the volume, expect the reviews to be thorough rather than fast — every fix in this repo gets its main change mutated, and a test that survives that is a blocking finding.

dchaudhari7177:
Rebased onto `main`; the conflict is resolved and the PR is mergeable again.

The overlap was one hunk: #696 landed a pin on `gradient_accumulation_steps` with its own comment while this was open. I kept my longer wording there because it states the *mechanism* — accumulation groups micro-batches into one optimizer step, so the default of 4 quartered the steps behind every published figure — where the merged comment states only the `iters` rounding consequence. Both facts now sit together in one block. Happy to take the shorter version instead if you'd rather keep `main`'s exact text.

Everything else in the diff is untouched by #696 and is what this PR is actually for:

- `epochs`, `lr` and `batch_size` pinned with the reason each one moves the step count.
- A runtime assertion after the YAML is written, which is the part that outlives the comments:

```python
resolved_iters = rows_n * epochs // cfg.training.batch_size // cfg.training.gradient_accumulation_steps
if resolved_iters != expected_iters:
    sys.exit(...)
```

A comment explaining a pin is only as good as the next person reading it. This fails the harness loudly if a schema default moves under us or a pin is deleted in a later edit — which is exactly what #696 turned out to be — instead of quietly publishing a throughput figure measured over a different number of steps. The resolved count is also printed in the run output so it appears in the record itself.

Srinivasan8888:
@dchaudhari7177 — I stood down on #716 in favour of this PR; your design is better than the one I proposed and I said so there. I own the harness and the record this touches (#663) and I have the M1 8 GB they were measured on, so here is the hardware check the PR cannot do from CI, plus one correction.

## Your harness runs clean on the actual hardware

Your branch at `431a9970`, real Metal, unmodified:

```
steps         : 48  (batch_size=1, grad_accum=1)
dispatch      : MLXSFTTrainerWrapper  <- verified before measuring
rows / epochs : 48 / 1
Starting training..., iters: 48                       <- mlx-lm's own count agrees
Iter 48: Train loss 0.230, ... Trained Tokens 2130, Peak mem 0.497 GB
Speed: 4.15 it/s     GPU peak: 0.497 GB
```

**`Trained Tokens 2130` is byte-for-byte the figure behind the published Qwen2.5-0.5B row.** So your "No published figure changes" holds under execution, not just under `48 % 4 == 0` — the record reproduces on your branch. That was the thing I most wanted checked and it passes.

## The correction: the guard's arithmetic is not the wrapper's

```python
resolved_iters = rows_n * epochs // cfg.training.batch_size // cfg.training.gradient_accumulation_steps
```

The wrapper does not divide by `gradient_accumulation_steps`. It **rounds down to a whole multiple** of it (`mlx_sft.py:302-311`):

```python
iters = int(cfg.training.epochs * max(1, math.ceil(len(train_rows) / batch_size)))
if grad_accumulation_steps > 1:
    iters = max(grad_accumulation_steps, iters - (iters % grad_accumulation_steps))
```

I did not want to tell you your PR body was wrong off a transcription, so I got it from the real code path — patched `mlx_lm.tuner.trainer.TrainingArgs` with a subclass that records its kwargs, ran the real `MLXSFTTrainerWrapper`, and read back what it actually constructed:

```
rows= 48 epochs=1 accum=1 -> REAL TrainingArgs(iters=48, grad_accumulation_steps=1)   guard predicts 48
rows= 48 epochs=1 accum=4 -> REAL TrainingArgs(iters=48, grad_accumulation_steps=4)   guard predicts 12
rows= 50 epochs=1 accum=4 -> REAL TrainingArgs(iters=48, grad_accumulation_steps=4)   guard predicts 12
rows= 10 epochs=1 accum=4 -> REAL TrainingArgs(iters= 8, grad_accumulation_steps=4)   guard predicts  2
```

So this line in the PR body:

> `-> accum=4 steps=12 guard aborts` — That last line is the bug as it stood: 48 rows silently becoming 12 optimizer steps.

is not what happens. At 48 rows with the old default the wrapper runs **48** iterations, and the harness reports `total_steps = iters = 48`. Nothing was quartered. (12 is the number of optimizer *updates*, but that is not the quantity this harness prints, publishes, or that `iters` names — `Trained Tokens` and the throughput are per-iteration.)

**This does not make your PR wrong — the pin is still necessary and the guard still fires.** But two things follow, and the second is the more interesting one:

**1. The message states a number the wrapper never uses.** `"resolve to {resolved_iters} optimizer steps"` will tell a future reader that 48 rows produce 12 steps. Whoever hits that abort will go looking for a quartering that is not there.

**2. The real defect at your own coincidence-breaker is worse than the PR describes.** At `rows=50, accum=4` the wrapper does not run 50 iterations and it does not run 12 — it runs **48**, because `50 - (50 % 4) = 48`. Two rows are silently dropped from the run entirely. That is data loss, not a step-count relabelling, and it is a stronger argument for your PR than the one you made. Same at `rows=10`: 8 iterations, 2 rows gone — which is worth knowing for the `(10, 10)` case in `test_issue23_mlx_harness_bridge.py`.

Suggested shape, so the guard measures what the wrapper does:

```python
resolved_iters = epochs * math.ceil(rows_n / cfg.training.batch_size)
accum = cfg.training.gradient_accumulation_steps
if accum > 1:
    resolved_iters = max(accum, resolved_iters - (resolved_iters % accum))
```

Then the abort message can say *"…resolve to N iterations, dropping M rows"*, which is both true and a better warning.

## Two smaller notes

- `sys.exit(...)` with a string exits **1** and prints to stderr, which is right for a harness. Worth stating in the PR body that the abort is a non-zero exit, since "refuses to measure" could be read as a skip.
- The guard reads `cfg.training.*` after writing the YAML, so it genuinely tests the loader rather than the file — that is the right call and it is what makes this a guard rather than a second copy of the pins. Keep it.

Happy to re-run anything else on the M1 while this is open; ask rather than infer a number from the record. Nice piece of work — the "resolve from the loaded config and refuse" shape is the part I would not have thought to add.

MakazhanAlpamys:
@dchaudhari7177 — **the gate is sound and it does not invalidate the published record. Two things to fix: a changelog fragment, and a diagnostic message that states a number the wrapper never computes.**

Head `431a997` verified. Worktree with `PYTHONPATH` pinned inside it.

## The gate does its job

Mutation: delete the `gradient_accumulation_steps: 1` line from the harness YAML, i.e. go back to inheriting the schema default of 4.

```
189 failed, 9 passed
E  SystemExit: harness step count is not 10: batch_size=1, gradient_accumulation_steps=4
   resolve to 2 optimizer steps. Pin them in the YAML above; a published figure
   must not depend on a schema default.
   benchmarks\harness\mlx_sft_smoke.py:202
```

Killed by the guard's own `SystemExit`, not by downstream rounding. That is the right failure.

Zero test lines removed — the PR touches no test file at all. `204 passed` on the two harness suites; **`251 passed`** on a locally-built merge onto current `main` including the #698 suite. Ruff clean.

## It does not stale a published number

Every published run in `benchmarks/run-m1-8gb-mlx-sft.md` is 48 rows / 1 epoch (lines 55, 119, 184, 206, 271); no other row count appears. Your diff adds prose and one parenthetical to the Config/Data rows and touches no measured figure. The quoted stdout in *Reproducing* contains only the `train:` and `throughput:` lines, so the new `steps :` line does not stale it. @Srinivasan8888 re-ran your branch on the real M1 and got `Trained Tokens 2130` — byte-identical to the published Qwen2.5-0.5B row. That is the check CI cannot do and it came from the person who owns the record.

**The #696 rebase is correct.** `3365f89` contributed the pin plus three comment lines; you kept the pin and replaced the comment, and the new wording still names #696 and its `iters`-rounding consequence. No merged fact was dropped.

## BLOCKING 1 — changelog fragment

I called this "arguably exempt" on the thread and I was wrong on the precedent: #703 was also a benchmark-harness-only PR and shipped `changelog.d/0.74.0/703.fixed.md`. Add `changelog.d/0.74.0/727.<category>.md` — **PR** number, not 716 — crediting `by @dchaudhari7177`. `CHANGELOG.md` untouched, which it already is.

## BLOCKING 2 — the guard's arithmetic is not the wrapper's, so the abort message misdirects

@Srinivasan8888 raised this and I verified it from source rather than taking it on trust. The wrapper (`mlx_sft.py:290-311`):

```python
iters = int(epochs * max(1, math.ceil(len(train_rows) / batch_size)))
if accum > 1:
    iters = max(accum, iters - (iters % accum))
```

It **rounds down to a multiple** of accum. It never divides. Your guard computes `rows*epochs // batch // accum`, which is the optimizer-*update* count — `tests/test_issue684_mlx_grad_accumulation.py:169` defines `_optimizer_updates = iters // accum`, confirming that is a different quantity.

Consequences:

- At 48 rows with accum 4, the wrapper runs **48** iterations and the harness would publish `total_steps = 48`. Your message says "resolve to 12 optimizer steps", sending a future reader after a quartering that did not happen.
- At the issue's own coincidence-breaker, rows=50 / accum=4: `50 - (50 % 4) = 48`. **The wrapper drops 2 rows from the run entirely.** That is data loss, and it is a stronger argument for this PR than the one the PR body makes.

**The gate itself is still correct.** It accepts only when `rows*epochs // batch // accum == rows*epochs`, which for non-empty input forces `batch == 1` and `accum == 1` — and at that point both formulas agree exactly. So the acceptance criterion holds. Only the diagnostic is wrong, and a wrong diagnostic on the one path a human reads is worth fixing.

Mirror the wrapper: `resolved = epochs * ceil(rows / batch)`, then round down by accum, and phrase the abort as *"resolve to N iterations, dropping M rows"*.

## Two smaller ones, same area

- **Floor vs ceil on the reject path.** For `batch_size > 1` you floor where the wrapper ceils — rows=10 / batch=4 gives your 2 against the wrapper's 3. Reject path only, never reachable on an accepting run, but it is the same mismatch.
- **`rows_n = 0`.** Your guard: `0 == 0`, passes. The wrapper: `max(1, ceil(0/1))` = 1 iteration. Degenerate, but the guard's whole premise is "never publish a step count you did not resolve".

## INFO — the guard has no test of its own

`git grep "harness step count\|resolved_iters" -- tests/` returns nothing. It is exercised only indirectly, through `test_issue23_mlx_harness_bridge.py` calling `main()`. Deleting the **pin** is caught, as I proved above. Deleting the **guard** while keeping the pin would not be. A `pytest.raises(SystemExit)` case against a mutated YAML would make it self-defending. Not blocking — belt and braces is still braces.

## CI staleness — worth a rebase here specifically

```
git log -1 --format='%P' refs/pull/727/merge  ->  base 60fe73d
git rev-parse origin/main                     ->  eb042c0   (4 behind)
```

One of those four is `9b58300` (#698), which modifies `src/soup_cli/trainer/mlx_sft.py` — **the exact module this harness drives**. Your 14 green checks were computed without it. I built the merge locally and re-ran (251 passed), so it is benign in fact, but this is the one PR in today's batch where the staleness is not merely procedural. Rebase before merge.

## One phrasing note

The PR body says the harness "refuses to measure", which reads like a skip. It is `sys.exit("...")` — a **non-zero exit to stderr**. Worth saying, since that is the difference between a CI job going red and a CI job going quiet.

MakazhanAlpamys:
**The hold is lifted. @dchaudhari7177, #727 proceeds on its merits — nothing is standing it down.**

@Srinivasan8888 answered on #716 well inside the window and stood down, and the reason is worth quoting rather than summarising:

> I was going to pin `gradient_accumulation_steps: 1` in the harness YAML and write down why. They pinned it **and** made the pin unnecessary to trust [...] My version would have been correct on the day it merged and silently wrong again the next time a schema default moved — which is precisely the failure #716 exists to describe.

That is the right call and it is the right reason. Your design aborts on the arithmetic — the harness resolves the step count out of the *loaded* config and refuses to measure when it is not `rows * epochs` — where the version I had in mind aborts on nothing and simply hopes the line stays. A guard that cannot recur the defect beats a fix that can.

Your `rows=50` case is also the correct coincidence-breaker: `48 % 4 == 0` is exactly why the published record survived by luck, and 50 is the number that tells a real resolution apart from that luck. Running it both ways rather than reasoning about it is the part I would have asked for.

## Two things now

**Your review is next in my queue**, on the merits and with no handicap from having arrived into a claim that was never actually assigned. The full battery applies as it does to everyone: I mutate the main fix, and a test that survives it is a blocking finding.

**@Srinivasan8888 has offered something CI cannot give**, and I am taking it up:

> I have the M1 8 GB the record was measured on [...] re-run the harness on the real hardware against their branch and confirm the resolved step count and the abort behave as claimed on Metal, not just against the loader.

Yes please — post it here as a review comment, credited to you for the measurement and to @dchaudhari7177 for the change. That closes the one gap a Windows box and nine CPU-only CI cells structurally cannot: everything either of us can check is against the loader and the fake, not against Metal.

@dchaudhari7177 — they also wrote that `run-m1-8gb-mlx-sft.md` and the harness are theirs from #663, so ask them directly if any figure in that file reads oddly rather than inferring it.

## And the record on the process failure stays as it is

@Srinivasan8888 lost a task they had not started, which is the cheapest possible version of this, and they said so. That does not make my part right. Four public "assigned to you" statements in four days that changed nothing, because `POST /issues/{n}/assignees` answers **200 with an empty array** rather than an error — the fix is reading the state back instead of trusting the call, and the `claimed` label is the part that outlives me noticing.

MakazhanAlpamys:
@dchaudhari7177 — **the design does what @Srinivasan8888 said it does, and I verified the property that made him stand down.** One blocking item, one gap worth closing.

## The guard is real, and the pin is not what it depends on

That distinction is the whole reason this version was preferred over pinning a YAML line. I extracted the YAML template verbatim from the source by regex and ran it through the real `load_config_from_string` plus your guard arithmetic:

```
=== AS SHIPPED ===
48x1 (the published record)     batch=1 accum=1 steps=48    measures
50x1 (coincidence-breaker)      batch=1 accum=1 steps=50    measures
10x1 (existing bridge test)     batch=1 accum=1 steps=10    measures

=== gradient_accumulation_steps PIN DELETED ===
48x1 (schema default takes over) batch=1 accum=4 steps=12    GUARD ABORTS
=== accum pinned to 4 ===                        steps=12    GUARD ABORTS
=== batch_size moved to 2 ===    batch=2 accum=1 steps=24    GUARD ABORTS
```

Delete the pin and the guard refuses rather than silently measuring 12 optimizer steps where the record says 48. It catches `batch_size` too, which is more than #716 asked for.

**And it is on the executed path**, not decoration: forcing it to always abort fails 189 of 198 bridge tests, and restoring the exact pre-fix state — guard disabled *and* pin deleted — fails **183 of 198**, reproducing #716's stated figure precisely.

**No published figure moved.** I took the multiset of every number on removed lines minus every number on added lines:

```
numbers only in removed lines: {}          <- empty
numbers only in added lines  : {'1': 5, '48': 3}
```

Two table rows edited, both restating facts already true, plus the Reproducing paragraph #716 asked for. Zero test lines removed, ruff clean, 198 bridge tests pass.

## Blocking — changelog fragment, and I do not think this is exempt

I said on my earlier comment that a benchmark-harness change was *arguably* exempt. Having checked the repo's own practice, I no longer think so, and the precedent is direct: **#703 touched exactly these two files and shipped `changelog.d/0.74.0/703.fixed.md`**; #696 touched the harness and shipped `696.fixed.md`. `benchmarks/` is the public, citable gate record that README and the preprint DOI point at, so a change to what a published record claims is user-facing by this project's own standard.

`changelog.d/0.74.0/727.fixed.md`, mentioning `#727`, crediting `by @dchaudhari7177`.

## Non-blocking, but it is the gap I would most want closed

**The `rows=50` coincidence-breaker is a manual check, not a test.** Your PR changes two files and adds none, so that line in the body cannot fail on pre-fix code — it does not exist as code. I reproduced it myself and it is correct, but **the guard's abort path has no automated coverage at all**. The happy path is covered incidentally by the 198 bridge tests at rows=10; the refusal is not.

About fifteen lines asserting `SystemExit` when the pin is absent would close it, and there is precedent right here: #703 changed these same two files and shipped a 233-line test file with them. Not a merge blocker, but a guard nobody tests is one release away from being a guard nobody has.

## Sequencing

@Srinivasan8888 has offered to re-run the harness on the real M1 8 GB the record was measured on, and confirm the resolved step count and the abort behave as claimed **on Metal** rather than only against the loader. That closes the one gap neither a Windows box nor nine CPU-only CI cells can. I would like that on the thread before merging, but it is not on you to produce.

MakazhanAlpamys:
@dchaudhari7177 — checking in. Two items from 2026-09-07, both small:

1. `changelog.d/0.74.0/727.fixed.md` — named after the PR, crediting you.
2. The diagnostic message states a number the wrapper never computes. Details are in that comment.

The design itself is settled and it is the reason @Srinivasan8888 stood down on #716 in your favour: your version makes the pin unnecessary to trust rather than merely pinning it. That verdict has not moved.

Same offer as on #729 — if the remaining work is paperwork you would rather not spend an evening on, say so and I will write both and merge with the work credited to you.

dchaudhari7177:
@MakazhanAlpamys @Srinivasan8888 — sorry for the silence since the 7th. All items are in `7849673`, rebased onto current `main` (`0e8cc1d`, which includes the `mlx_sft.py` change you flagged).

**Blocking 1 — fragment.** `changelog.d/0.75.0/727.fixed.md`. It is under `0.75.0` rather than `0.74.0` because 0.74.0 has since been released and the validator rejects a fragment based on an older release. It credits @Srinivasan8888's M1 re-run, since that is the evidence behind "no published figure changes". I checked it by running `assemble_changelog.py` against a temp copy.

**Blocking 2 — the diagnostic.** You were both right. The guard computed the optimizer-update count and called it the step count. Now `resolve_step_counts(rows, epochs, batch_size, accum)` works the way the wrapper does (`epochs * ceil(rows / batch)`, rounded down to a whole group) and returns **both** numbers.

One thing I had to keep in mind here: if the guard only compared *iterations*, deleting the accumulation pin at 48 rows would pass, because the wrapper still runs 48. So the guard requires iterations **and** updates to equal `rows * epochs`, and the message names both:

```
harness run is not 48 iterations / 48 optimizer updates: batch_size=1, gradient_accumulation_steps=4
resolve 48 rows x 1 epoch(s) to 48 iterations and 12 optimizer updates, dropping 0 row(s). ...

... resolve 50 rows x 1 epoch(s) to 48 iterations and 12 optimizer updates, dropping 2 row(s). ...
```

This also fixes the two smaller items. It uses ceil rather than floor (10 rows at batch 4 gives 3, not 2). And `rows_n = 0` now aborts, because the wrapper still runs one iteration. The YAML comment that said the default "quartered the steps" is corrected too.

**Phrasing.** The benchmark doc now says the harness *exits non-zero before training starts*, not "refuses to measure".

**INFO — the guard had no test.** `tests/test_issue716_harness_step_guard.py`:
- the helper checked against the `TrainingArgs` the **real** `MLXSFTTrainerWrapper.train()` builds, across six shapes, including Srinivasan's four read-backs (48/1, 48/4, 50/4, 10/4);
- three abort-path tests through the real `main()` using the bridge fixture. Each mutates the loaded YAML (accum pin deleted at 48 rows and at 50 rows, and `batch_size: 4`), asserts `SystemExit` with the counts, and asserts no display was ever created;
- the shipped pins pass and print both counts.

Mutations: disabling only the guard (pins kept) → **3 failed**; restoring the old division formula in the helper → **13 failed**. `tests/test_issue716_harness_step_guard.py` + `test_issue23_mlx_harness_bridge.py` + `test_issue684_mlx_grad_accumulation.py`: 220 passed. `ruff@0.9.4 check` clean on both files. I have no Apple Silicon, so the Metal run stays with @Srinivasan8888's earlier result.

MakazhanAlpamys:
Merged as `b92b20a1`. Every item from me and from @Srinivasan8888 is verified addressed by execution, and the pins are not decorative — I proved that by trying to make them so.

**The result I most wanted.** The point you made yourself in the PR body — that an iterations-only guard would accept a deleted accumulation pin at 48 rows, because the wrapper runs 48 either way — is the thing that decides whether this file is a guard or decoration. I weakened `if iters != expected or updates != expected:` to `if iters != expected:`:

```
FAILED tests/test_issue716_harness_step_guard.py::test_deleting_the_accumulation_pin_aborts_with_the_real_counts
1 failed, 15 passed
```

Finding that failure mode in your own work and then writing the test that catches it is the hard part, and you did it before anyone asked.

**The rest of the mutations, all killed:** guard disabled with pins kept fails 3 (your figure); the pre-fix `rows*epochs // batch // accum` formula restored fails 13 (your figure); ceil to floor fails 3. My round-1 mutation — delete the accumulation pin — now aborts with `resolve 10 rows x 1 epoch(s) to 8 iterations and 2 optimizer updates, dropping 2 row(s)`, where the old text said "2 optimizer steps". The arithmetic was cross-checked against the `TrainingArgs` the real `MLXSFTTrainerWrapper` builds across six shapes, including all four of @Srinivasan8888's M1 read-backs. 14/14 CI green on the head SHA; 220 passed across the three suites; 237 on a local merge with current `main`. No published figure moved — the multiset of numbers on removed lines of the record is empty, same as last round.

**On the changelog fragment: you were right and I was wrong.** I asked for `changelog.d/0.74.0/`; you filed `0.75.0/` because the release landed in between, and following my instruction would have failed `test_issue487_changelog_fragments.py` in every matrix cell. Crediting @Srinivasan8888 in your own fragment for the M1 re-run is the right precedent and I am keeping it.

**Two follow-ups, neither reachable on a run this guard accepts, both one line. I have filed them rather than hold a green PR:**

1. `resolve_step_counts` keeps the `max(accum, ...)` floor that stops the round-down going below one group, and removing it passes all 220 tests — every case in `_CASES` has `iters >= accum`, so the `max()` never binds. It bites only at `rows < accum`: at `rows=2, accum=4` the wrapper runs 4 iterations while the mutant would print "0 iterations ... dropping 2 rows". `((2, 1, 1, 4), (4, 1))` in `_CASES` closes it and gets cross-checked against the real wrapper for free.
2. `epochs` is the one count-moving field still read from argv rather than the loaded config — `cfg.training.batch_size` and `cfg.training.gradient_accumulation_steps` are both resolved from it, `cfg.training.epochs` is not. Change the template to `epochs: 2` and all 16 of your tests pass; only the bridge suite notices, 183 failures away. That is exactly the property @Srinivasan8888 called out as the reason this is a guard rather than a second copy of the pins, and it currently holds for two fields of three.

Say the word and I will take both myself with the work credited to you, or push them whenever suits.

One cosmetic note not worth an issue: `dropped = max(0, expected - iters * batch_size)` is only exact at one epoch — at `rows=10, epochs=2, batch=4, accum=4` it reports 4 where the row-slot loss is 8. Reject path only, multi-epoch never used by the record.

Credited in `CONTRIBUTORS.md`. Thank you for coming back to this, and there is no need to apologise for the gap — I left #757, #742 and #729 waiting on me for longer than you left this waiting on you. All three are reviewed now.

PR #729, opened 2026-09-06
fix(data): streaming over + val_split no longer puts a row on both sides

### LINKED ISSUE #702: Streaming interleave: over + val_split still puts the same row in train and val

dchaudhari7177:
Closes #702.

Reproduced at the exact shape from the issue before changing anything — 100 rows against 10, `over`, `val_split: 0.1`:

```
BEFORE  train=180 val=20  overlapping val rows=10
AFTER   train=189 val=11  overlapping val rows=0
```

Ten of twenty, matching the measurement in the issue.

### The fix, and why it is not #701's

#701 carves `val_split` out per source **before** padding. That cannot transfer: a stream is not countable ahead of time, so there is no length to take a fraction of before `interleave_datasets` starts.

What this path *does* have by the time it splits is the materialised rows — it already collects up to `MAX_REMOTE_ROWS` into `raw_data` before formatting. So `_split_val_deduplicated` splits over the **distinct** rows and then withholds every copy of a chosen val row from train.

**The trade, stated rather than hidden:** train loses the oversampling for exactly the rows that became val. That is deliberate — oversampling a row is a tuning choice, while having it on both sides of the split makes the validation number meaningless. Everything else keeps its duplicates, and a test asserts train still contains recycled copies of non-val rows.

I did not take the "refuse the combination" option. It is a real feature on a path where it works fine as long as the split is duplicate-aware, and refusing would be a regression for anyone using it with `val_split: 0`.

### Acceptance criteria

- **Zero overlap** — asserted end-to-end through `load_dataset`, and the test **fails without the fix** (verified by disabling the branch: 1 failed, 11 passed).
- **Disjointness *and* non-empty val** — asserted together in the same test, since disjointness alone is trivially true when val is empty. There is also a separate test pinning that the *old* positional split really does leak, so the disjointness claim is measured against something real rather than being unremarkable.
- **`probs` / `concat` / `under` unchanged** — parametrised control test through `load_dataset`. Only `over` uses `all_exhausted`, so only `over` is rerouted.
- **Config-parse refusal** — not applicable; the combination is fixed rather than refused.
- **`docs/data.md`** — the sentence saying the fix "does not reach the streaming path" is replaced with what the streaming path now does and why the mechanism differs.

### Tests — 12, all passing

Seven unit tests on the split itself (leak pinned, disjointness + non-emptiness, no row lost, duplicates of train rows kept, agreement with `_split_val` on duplicate-free input, `_row_key` key-order insensitivity, and `_row_key` surviving an unserialisable value so a leak check cannot become a crash), plus five end-to-end through `load_dataset` reusing `_install_fake_streaming_datasets` from the #459 suite.

`test_issue459_interleave_streaming_hub.py` and `test_issue443_interleave_wiring.py` still pass — 65 total with the new file. `ruff check` clean.

Mechanism and the eager-path fix are @AmirF194's in #680/#701; this is the streaming half they identified.


src/soup_cli/data/loader.py
```diff
@@ -472,6 +472,57 @@ def _split_val_per_source(
...
+    meaningless.
+    """
+    seen: dict[str, dict] = {}
+    for row in rows:
+        seen.setdefault(_row_key(row), row)
+    distinct = list(seen.values())
+
+    _, val_rows = _split_val(distinct, val_split)
+    val_keys = {_row_key(row) for row in val_rows}
+    train_rows = [row for row in rows if _row_key(row) not in val_keys]
```
AmirF194, line 522: I ran this against `bfb663c` in a clean container.

`_split_val_deduplicated` (`src/soup_cli/data/loader.py:491-521`) keys rows by content only (`_row_key`, line 475), so it cannot tell a recycled copy of a `small`-source row apart from two genuinely distinct examples that happen to render to identical text. When two such rows collide and the resulting single entry lands in val, the second occurrence disappears from the output entirely: not in train (filtered out because its key matches `val_keys`), not in val (only one copy is kept there either).

```
python: two rows in `big` (not from the `over` recycling of `small`) both render
to {"text": "DUPLICATE_CONTENT"}, at positions 18 and 19 out of 20.
after _split_val_deduplicated(rows, 0.3):
DUPLICATE_CONTENT copies in original rows: 2
DUPLICATE_CONTENT copies in train after split: 0
DUPLICATE_CONTENT copies in val after split: 1
total rows in: 40, train+val out: 39
```

One real training row vanishes with no error, and `test_no_row_is_lost_by_the_deduplicated_split` doesn't catch it because it only checks that the set of keys covers train+val, not that non-val-selected multiplicities survive. This isn't the `over`-recycling case the fix targets; it's any corpus with naturally duplicate short/templated rows (common in SFT data) that also happens to use `streaming: true` + `over` + `val_split`.

Maybe carry the original row count alongside the dedup key so `train_rows` can keep every occurrence whose key isn't in `val_keys` minus exactly the number of copies actually moved to val, instead of dropping every remaining copy.

MakazhanAlpamys:
@dchaudhari7177 — **the fix is correct and I verified it against the real `datasets` library rather than your fake. Blocking on a one-line changelog fragment, plus one control test that does not control what its docstring says it does.**

Head `bfb663c` verified against `headRefOid`. Worktree with `PYTHONPATH` pinned inside it.

## The fix works, and I checked it the hard way

Your e2e tests drive a fake `datasets` module. I wrote an independent check against the **real** `datasets` 4.5.0, so the `all_exhausted` semantics are genuine:

```
--- clean main (BASELINE) ---          --- your branch ---
over    train=179 val=20 OVERLAP=10    over    train=188 val=11 OVERLAP=0
concat  train= 99 val=11 OVERLAP= 0    concat  train= 99 val=11 OVERLAP=0
under   train= 18 val= 2 OVERLAP= 0    under   train= 18 val= 2 OVERLAP=0
probs   train= 23 val= 3 OVERLAP= 0    probs   train= 23 val= 3 OVERLAP=0
```

The baseline reproduces **10 of 20** — exactly the number in #702. After the fix `set(train) & set(val)` is empty **with train and val both non-empty**, which is the pair of criteria that matters; either alone is trivially satisfiable.

**The non-streaming path is provably unchanged.** SHA-256 of the full `{train, val}` result, your branch vs `main`:

```
EAGER over   sha256=aa43ba46f5bc2e1d   identical
EAGER concat sha256=2fb67b6a19acd97c   identical
EAGER under  sha256=92caddc66fb3811f   identical
```

Deterministic across repeated runs. No off-by-one at the boundary: `train=188, val=11` from 199 materialised / 110 distinct rows is exactly the arithmetic your PR body predicts.

Zero deleted test lines. Ruff clean. 12 of your tests pass; 65 with the #459 and #443 suites; **388** across a broader loader/interleave sweep. `Closes #702` present and the GitHub linkage is live.

## BLOCKING 1 — no changelog fragment, and CI cannot tell you

`changelog.d/0.74.0/` already contains `701.fixed.md` — the **eager half of this very bug** — and there is no `729.fixed.md`. This matters more than it looks: `publish.yml` runs `assemble_changelog.py --check-empty`, which asserts the directory is *empty* at release time. It has no way to detect a fragment that was never written. So this fix would vanish from the v0.74.0 notes with **no CI signal at all**, sitting next to the fragment for its own other half.

Add `changelog.d/0.74.0/729.fixed.md`, named by the **PR** number, crediting `(#702 by @dchaudhari7177 in #729)`. Do not touch `CHANGELOG.md`.

## BLOCKING 2 — `test_the_other_strategies_are_unchanged` does not test what it says

Its docstring says it "asserts the fix did not quietly reroute them". It does not. Mutation:

```
M4: gate on `val_split > 0` instead of on strategy == "over"
    -> reroutes concat/under/probs through the dedup path as well
    -> 65 passed across all three interleave suites.  SURVIVED.
```

Five other mutations died cleanly:

| mutation | result |
|---|---|
| disable the `over` branch entirely | killed (1 failed) |
| `distinct = list(rows)` — drop dedup | killed (1 failed) |
| `train_rows = list(rows)` — no withholding | killed (4 failed) |
| gate on `"under"` instead of `"over"` | killed (1 failed) |
| `_row_key` returns a constant | killed (7 failed) |

So the suite is strong; this one test is the exception, and it is the one whose *stated job* is to catch exactly M4. It passes because it only checks `overlap == 0` and `train` non-empty — both of which survive the reroute. One line fixes it: assert the un-deduped `train` length for those strategies, or spy on `_split_val_deduplicated` and assert it was not called.

This is not a defect in shipped code. It is a control that would not notice the regression it exists to notice, which is worth closing while you are in the file.

## Non-blocking, but I would take the first one

**A degenerate input now yields an empty train set, silently.** With one distinct row across both sources:

```
both sources = same single row, over, val_split=0.1
  main:  train=1 val=1 overlap=1   (leaky, but it trains)
  PR:    train=0 val=1 overlap=0   (disjoint, but EMPTY TRAIN, no error)
```

The eager path raises a friendly `ValueError` in the analogous situation (`_split_val_per_source`, `loader.py:466`). The new streaming path has no equivalent. Very narrow — it needs exactly one distinct row — but it is the same class of guard #701 deliberately added, and "trains on nothing without saying so" is the worse failure of the two.

**One doc sentence is true but easy to misread.** Your new line in `docs/data.md` says `concat`/`under`/`probs` never duplicate a row on the streaming path. That is true *about `interleave_datasets`* and reads as "so they cannot leak". Natural duplicates already present in the source data still do, on your branch:

```
source has 20 identical rows:
  concat -> val=12, overlap=12
  under  -> val= 4, overlap= 2
  over   -> val=11, overlap= 0   (fixed)
```

Identical on `main` and on the eager path, so it is out of #702's scope and I am not asking you to fix it. I am asking for the sentence to not imply otherwise.

## One note on CI

Your green run was computed against `60fe73d`, four commits behind current `main`. I built the merge locally and re-ran: clean auto-merge, 65 passed. None of the four touches `data/loader.py`, so this is benign — recording it because a green check against a stale base is not the same claim as a green check.

Add the fragment and the control-test line and this merges.

MakazhanAlpamys:
@dchaudhari7177 — **your numbers reproduce exactly, and the vacuity trap that bit the eager half of this same leak is closed on both headline tests.** One blocking item, two worth taking.

## Reproduced, on both trees

100 rows against 10, `over`, `val_split: 0.1`:

| tree | strategy | train | val | val rows also in train |
|---|---|---:|---:|---:|
| `main` | over | 180 | 20 | **10 / 20** |
| this PR | over | **189** | **11** | **0** |
| both | probs / concat / under | 15 / 99 / 18 | 2 / 11 / 2 | 0 |

Your `180/20/10 -> 189/11/0` is exact, and the control strategies are byte-identical between trees.

**Three mutations, three killed:** deleting the `over` + `val_split` branch (the carve-out reverted); keeping the branch but no longer withholding duplicate copies; and splitting over raw rows instead of distinct rows — the exact pre-fix mechanism.

## The vacuity check, which matters here more than usual

#701 fixed the eager half of this leak and its headline test was **vacuous**: `assert not (train & val)` is trivially true against an empty val set, and its companion assertion was a *total*, invariant under never-splitting. It passed on a build where validation did not happen at all.

So I ran the never-split mutation against each of your tests **individually**. Both headline tests fail alone:

- `test_deduplicated_split_is_disjoint_and_val_is_not_empty` — asserts val non-empty **before** disjointness
- `test_streaming_over_with_val_split_has_no_overlap` — asserts `result["val"]` non-empty

Your docstring calls the trap out by name and the assertion order is right. That is the thing I would have blocked on and you had already handled it.

Scoping is correct too, and I checked rather than assumed: only `loader.py:780` (`over`) uses `stopping_strategy="all_exhausted"`, while `under` and `probs` use `first_exhausted`, so the guard matches the actual duplication source. #701's eager path is byte-identical on branch and `main` across four shapes; 53 passed on the two interleave suites, 278 across the whole data-loader surface on both trees. Zero test lines removed.

## Blocking

**No changelog fragment.** `changelog.d/0.74.0/729.fixed.md`, mentioning `#729`, crediting `by @dchaudhari7177`. This is a user-facing silent-correctness fix, so not exempt.

## Two non-blocking, and the first is the same trap one layer over

**Your three control tests carry the #701 shape.** `test_the_other_strategies_are_unchanged[concat|under|probs]` asserts `_overlap(result) == 0` and `result["train"]`, but **not** `result["val"]`. Under a never-split regression on those strategies all three pass silently. One line each — `assert result["val"]` — and the real values are 11, 2, 2, so it costs nothing. You got this right where it mattered most; worth finishing.

**No empty-train guard, which is an asymmetry with the eager path you are mirroring.** `_split_val_per_source` raises a named error when `val_split` consumes a source's whole train side; `_split_val_deduplicated` has no equivalent. On a fully-degenerate dataset — every row identical, so one distinct row — with `val_split: 0.5`:

```
streaming=True : train=0 val=1     <- empty train, no error
streaming=False: train=8 val=5
```

Narrow, since `val_split` is capped at 0.5 so it needs at most one distinct row. But it silently produces a zero-row training set where the eager path gives a clear message, and "silently wrong instead of loudly refusing" is the shape of the bug you just fixed.

Add the fragment and I merge; the two nits are your call.

MakazhanAlpamys:
@dchaudhari7177 — **@AmirF194's inline finding reproduces on your head, and it is a second blocking item. I am adding my own numbers rather than forwarding his, and I disagree with the remedy he suggested — the reason is worth reading before you fix it.**

## Reproduced, `bfb663c`, calling your function directly

Twenty rows, of which two are genuinely distinct source rows that render to identical text — **not** the `over` recycling this fix targets. Templated SFT data produces these routinely.

```
in rows total: 20   duplicate-content copies in: 2
train: 13   val: 6   train + val: 19
duplicate-content copies in train: 0   in val: 1
```

**Twenty rows in, nineteen out.** One real training row leaves the pipeline with no error, no warning and no log line. On `main` the positional `_split_val` loses nothing, so this is a regression the fix introduces, not a pre-existing gap — narrow, but silent, and silent row loss is the same failure shape as the leak you are closing.

## Why I am not asking for @AmirF194's remedy

His suggestion is to carry multiplicities so train keeps every occurrence whose key is not in `val_keys`, minus the copies actually moved to val. **That reinstates the leak for exactly this input.** Two byte-identical rows, one in val and one kept in train, is the #702 defect at content level: the model trains on the val text and validation measures memorisation. Content-keyed dedup cannot distinguish a recycled copy from a genuine duplicate — there is no information in the row to distinguish them with — so keeping the remainder in train trades a quiet row loss for a quiet leak. That is not an improvement.

Two resolutions I would take, and the choice is yours:

1. **Prefer unique content when selecting val.** Split over the rows whose key occurs exactly once, and fall back to duplicated content only if there are not enough of them. In the common case nothing is lost and nothing leaks; in the degenerate case you are back to today's behaviour, which is then the honest answer rather than an accident.
2. **Keep today's behaviour and make it loud.** Count what the withholding drops and refuse — or warn with the number — when the loss is not zero. This is the same asymmetry I flagged with `_split_val_per_source`'s named error on the eager path, so both would be closed by one guard.

Either is fine by me. What is not fine is losing a row silently.

## A vacuous assertion, and it is the one guarding this

`tests/test_issue702_streaming_over_val_leak.py:116-126`:

```python
    val_keys = {_row_key(row) for row in val}
    kept = [row for row in rows if _row_key(row) not in val_keys]
    assert len(train) == len(kept)
```

`kept` is the implementation's own expression, character for character (`loader.py:520`). The assertion re-derives the code under test and compares it to itself, so it holds for *any* filter predicate the implementation might use — including a wrong one. It is green on the row-loss defect above, which is precisely what it looks like it is guarding.

The second assertion in that test (`len(train) > len(set(...))`) is real and does work. Replace the first with a statement about the world rather than about the code — for example, that every row whose key is not in `val_keys` survives **with its multiplicity**, computed from `rows` with a `Counter`, not from the same comprehension.

I said in my last round that the vacuity trap was closed **on both headline tests**, and that stands — I checked those two individually under the never-split mutation and both fail alone. This is a third test, and it carries the trap.

## Where this leaves the PR

Blocking, now two items:

1. `changelog.d/0.74.0/729.fixed.md` mentioning `#729` and crediting `by @dchaudhari7177` (unchanged from my last round).
2. The silent row loss above, by either route, with a test that fails on the current head.

The two non-blocking items from the previous round still stand and are still your call. The core of the PR is unaffected: the leak fix is right, the mutations still die, and the control strategies are still byte-identical between trees.

@AmirF194 — thank you, this was a real one and the reproduction was exact. Credit for it goes in the changelog entry alongside @dchaudhari7177's fix.

src/soup_cli/data/loader.py
AmirF194: Agreed, keeping the remainder in train can't tell a recycled copy from a genuinely duplicate row by content key alone, so it reopens the same leak for byte-identical rows. Good catch.

MakazhanAlpamys:
@dchaudhari7177 — checking in, and I want to be clear about what is on you here versus what is on me.

Two blocking items are still open from 2026-09-07, both of them real and neither of them large:

1. A one-line changelog fragment at `changelog.d/0.74.0/729.fixed.md`, named after the PR and crediting you.
2. @AmirF194's finding, which I reproduced on your head with my own numbers: `_split_val_deduplicated` keys rows by content alone, so it cannot tell a recycled copy of an oversampled row apart from two genuinely identical source rows. I disagreed with the remedy they suggested and said why in that comment.

**Nothing is standing this PR down and nothing is competing with it.** If you would rather I wrote both and merged with the work credited to you — your commits, your `CONTRIBUTORS.md` entry, your name in the fragment — say so and I will.

You currently have four PRs open against this repo and I am the queue on three of them, so if the backlog is mine rather than yours, tell me and I will clear it that way instead.

MakazhanAlpamys:
@dchaudhari7177 — one new finding, and it is the kind I would rather you heard before you push than after.

Nothing changes about the two items already open (the `729.fixed.md` fragment, and @AmirF194's content-keying finding). This is a third, and it is substantive.

**`_split_val_deduplicated` has no empty-train guard, and its sibling does.** `_split_val_per_source`, added by #701 for the analogous situation on the eager path, raises a named `ValueError` when the carve-out leaves zero training rows for a source. Your function has nothing. Measured:

```
1 distinct row, 20 rows total, val_split=0.5  ->  train=0   val=1
2 distinct rows, 20 rows total, val_split=0.5 ->  train=10  val=1
the eager sibling on the same input RAISES:
  "data.interleave: data.val_split=0.5 leaves 0 training rows..."
```

On `main` that first input gives train=18 / val=2 — leaking, which is the bug. So as it stands the PR trades a silent leak for a **silent empty training set**, and I do not think that is clearly better. `val_split` is bounded at 0.5 by the schema and low distinct-row counts are reachable on a stream, so this is not a contrived shape. Please mirror the sibling's guard with a message naming `streaming`, `over` and `val_split`, and add a test for it.

**And one sentence in `docs/data.md`.** On the issue's own 100-vs-10 shape at `val_split=0.1` I measured:

```
val   source mix: Counter({'a': 11})              # every val row from the big source
train source mix: Counter({'b': 100, 'a': 89})
train=189  val=11  overlap=0
```

`distinct` preserves first-occurrence order, so the small source's rows all land early and the positional tail falls entirely in the big one. #701's per-source carve-out on the eager path *is* mixture-representative, so the two paths differ here. Since #702 exists precisely because the docs implied parity between them, "reaches the same guarantee by a different route" should be followed by a clause saying val is drawn from distinct rows in stream order and may not be balanced across sources. Getting that clause right is most of the value of this fix.

**The core mechanism is verified and I am not asking you to change it.** I checked the branch condition independently rather than taking the docs claim: in `_load_interleaved_streaming_datasets`, `over` really is the only strategy using `stopping_strategy="all_exhausted"` (`under` and `probs` use `first_exhausted`, `concat` uses `concatenate_datasets`). Both mutations are killed — disabling the branch reproduces overlap = 10, exactly the issue's 10/20, and defeating only the duplicate-withholding while keeping the branch fails three tests. Pinning the old leak so the disjointness assertion is measured against something real is the right instinct, and so is asserting non-emptiness alongside it. 12 passed on your file, ruff clean, no deleted tests.

Two nits, take or leave. `_row_key` is recomputed roughly three times per row — 25k rows of ~800 chars costs 0.37 s here, which extrapolates to ~15 s at `MAX_REMOTE_ROWS = 1_000_000`; computing the keys once into a list is free. And `test_row_key_survives_an_unserialisable_value` passes because `default=repr` handles `object()`, so the `except (TypeError, ValueError)` branch it is named for is never reached — fine as a contract test, just not testing what the name says.

My earlier offer stands and now covers three items rather than two: if the queue here is mine rather than yours, say the word and I will write all three with the work credited to you.

dchaudhari7177:
@MakazhanAlpamys @AmirF194 — sorry for the gap. All three blocking items and both nits are in `13abbd3`, rebased onto current `main`.

**1. Fragment.** `changelog.d/0.75.0/729.fixed.md`. It is under 0.75.0 because 0.74.0 has shipped since the review. It credits @AmirF194 for the row-loss finding, and it passes `assemble_changelog.py` on a temp copy.

**2. Silent row loss: your option 1, with option 2 for the fallback.** Val is still sized over distinct rows, but it is now drawn from the tail preferring content that occurs **exactly once**. Taking such a row costs train nothing, so in the common case nothing is lost and nothing leaks. I agree with your reasoning against carrying multiplicities: keeping the leftover copy in train is the #702 leak at content level. If unique rows can't fill val, repeated content is used, its other copies are withheld from train, and the count is printed:

```
Warning: data.val_split under streaming 'over' withheld 4 duplicate row(s) from train: ...
```

AmirF194's shape (18 unique + 2 identical, `val_split=0.3`): 20 rows in, 20 out, both identical copies in train, no overlap. On the previous head it gave 19 out.

**3. Empty-train guard.** This mirrors `_split_val_per_source` and names `val_split`, streaming and `over`:

```
data.interleave: data.val_split=0.5 leaves 0 training rows under streaming 'over', which splits over
distinct rows: the 22 materialised row(s) hold 1 distinct row(s). Add distinct rows, set val_split: 0, ...
```

It is tested on the unit and end to end through `load_dataset`, at `val_split` 0.1 and 0.5. With a single distinct row, lowering `val_split` can't help, so the message doesn't suggest it.

**Vacuous assertion.** Replaced with a statement about the rows: `Counter(train) + Counter(val) == Counter(rows)`.

**Control tests.** Each of `concat|under|probs` now asserts `result["val"]` and spies on `_split_val_deduplicated`, failing if it is called. Your M4 (gate on `val_split > 0` alone) now fails **3 tests**; it survived before.

**docs/data.md.** It now says streaming val comes from distinct rows in stream order and **is not balanced across sources**, using your 100-vs-10 numbers, which I re-measured on this head: `train=189 val=11 overlap=0`, val `{'a': 11}`, and 189 + 11 = 200, so nothing is lost. It also says `concat`/`under`/`probs` don't *add* duplicates, but rows already duplicated in a source can still cross the split on any strategy.

**Nits.** `_row_key` is computed once per row. The `_row_key` test is split in two: `object()` exercises `default=repr`, and a new circular-reference case actually reaches the `except` branch. So the `pragma: no cover` is gone, and the fallback sort now tolerates non-string keys.

**Verification.** Against the previous head's `loader.py`, 5 of the new tests fail (repeated rows, the withheld warning, and the three empty-train cases). On this head the file is 19 passed. Across every test file mentioning `interleave`/`val_split` (28 files): 1655 passed, 27 skipped, 5 failed. Those 5 are all `test_issue755_backend_support_registry.py` and fail identically with my changes stashed; they depend on this machine's installed packages. `ruff@0.9.4 check` is clean.

MakazhanAlpamys:
Merged as `d0f6981d`. All three blocking items and all five nits verified on `13abbd3c` by execution, and the mutation that survived last round now kills three tests.

I re-measured against the **real `datasets` 5.0.1** end to end through `load_dataset`, not the fake fixture:

| tree | strategy | train | val | overlap |
| --- | --- | ---: | ---: | ---: |
| `main` | over | 179 | 20 | **10** |
| this PR | over | **188** | **11** | **0** |
| this PR | concat / under / probs | 99 / 18 / 24 | 11 / 2 / 3 | 0 |

The 10-of-20 leak still reproduces on current `main`, so this closes a live defect. Your `189/11` is the fake fixture's 200 rows; the real library materialises 199 for `all_exhausted`. Both are right, and both lose nothing.

**Mutations, seven applied, seven killed.** The two I care about:

```
disable the `over` carve-out              -> 2 failed  (the leak returns)
gate on `val_split > 0` alone             -> 3 failed  (all three control cases)
```

That second one **passed 65 tests on the previous head**. The spy on `_split_val_deduplicated` is what fixed it, and that is the difference between a control test and a decorative one.

**The part I want to name.** Your first fix took the tail of a deduplicated list and silently lost a genuinely repeated source row — @AmirF194 caught it. You could have patched around it by carrying multiplicities; instead you took the route that prefers unique content, **withholds the remaining copies and prints the count**, so the lossy case is loud rather than hidden. And you replaced the vacuous `len(train) == len(kept)` rather than adding beside it: that assertion re-derived the implementation's own filter, and its replacement — `_multiset(train) + _multiset(val) == _multiset(rows)` — is the only assertion in the file that kills the row-loss mutation. Deleting a test that cannot fail is harder to justify in review than adding one, and it was the right call.

The `_row_key` nit is real and I measured it: **0.852 s to 0.444 s** for 25,000 rows of ~800 characters, 1.92x. Peak memory unchanged, which is the honest framing — the list was always the size of the corpus.

Counts: 19 on your file, 152 across the nine core loader files, **2257 passed / 24 skipped** across all 45 files mentioning `interleave` / `val_split` / `data.loader`, zero failures. Ruff clean. Zero removed lines from any existing repo test. 15/15 CI green on `13abbd3c` itself. Fragment correctly under `changelog.d/0.75.0/`, crediting both you and @AmirF194.

**Two notes, neither blocking.** `test_too_little_unique_content_withholds_copies_loudly` fails under `FORCE_COLOR=1` because Rich's highlighter breaks the string around `'over'`, `4` and `row` — latent, not live: nothing in our workflows sets `FORCE_COLOR`, a `StringIO` console is not a tty, and 11 pre-existing tests fail the same way under that env. And the val set now **shrinks** on `over` (20 to 11) because the fraction is over distinct rows; I will call that out in the release notes.

@AmirF194 — your finding is closed by the route I preferred, the test that pins it fails on the old head, and you are credited in the fragment.

PR #730, opened 2026-09-06
fix(migrate): detect JSONL by structure

### LINKED ISSUE #699: soup migrate gives a raw ParserError for JSONL under a name that does not say .jsonl
Split out of #672, which #675 closed. @swalla02 raised this in that PR's final section and deliberately left it out of the fix — correctly, because it is a scope call rather than a bug, and it was mine to rule on.

## The gap

`soup migrate` gives a friendly error for a JSONL data file named `*.jsonl`. Under a name that does not say `.jsonl`, the same file gets the raw parser exception the N2 feature exists to replace:

```
$ soup migrate --from llamafactory data.json --dry-run
Error: ParserError: expected '<document start>', but found '{'
```

## Why it was not simply fixed alongside #672

The content sniff is gated by the suffix on purpose, and the reason is written into the code comment that predates all of this:

> the sniff helper is only invoked when the suffix says `.jsonl` (notebook `.ipynb` files legitimately start with `{` — we must not falsely flag them)

So "sniff every file" is not the fix. A `.ipynb` is a single JSON object beginning with `{`, and `soup migrate --from unsloth` takes those legitimately. Anything that widens the sniff has to keep that path working.

## What a real fix has to distinguish

Three shapes that all begin with `{`:

| input | correct outcome |
|---|---|
| a JSON Lines data file — many objects, one per line | friendly "this is data, not a config" error |
| a single JSON object that is a notebook (`.ipynb`) | migrate it |
| a single JSON object that is something else entirely | ? — needs a decision |

The JSONL case is distinguishable from the notebook case by structure rather than by name: JSONL has multiple top-level objects on separate lines, a notebook is one object with `cells` and `nbformat` keys. That is a stronger signal than either the suffix or `startswith("{")`, and it would let the suffix gate be dropped.

The third row is the open question, and it is why this is a scope call. A bare `{...}` that is neither JSONL nor a notebook is probably a config in some other format, and guessing wrong there produces a confidently wrong error message, which is worse than the raw `ParserError`.

## Suggested shape

- Keep the existing `.jsonl` fast path unchanged.
- Add a structural sniff for files whose suffix gives no hint: read a bounded prefix (`_looks_like_jsonl` is already bounded at 64 KiB after #675 — reuse it rather than writing a second reader), and classify as JSONL only on **two or more** parseable top-level objects on separate lines.
- Leave a single JSON object alone. Let it fall through to whatever the format-specific migrator does with it, including the notebook path.

## Acceptance criteria

- [ ] `data.json` containing real JSON Lines yields the friendly error, not a `ParserError`
- [ ] A `.ipynb` notebook still migrates through `--from unsloth`, and a test pins it — this is the case the whole suffix gate exists to protect
- [ ] A single non-notebook JSON object does **not** get mis-reported as JSONL
- [ ] The reader stays bounded; no new unbounded read (#675 measured 240.9 MB peak on a 120 MB one-liner before it was bounded)
- [ ] Tests drive the CLI through `CliRunner`, not the helper directly — the helper-only test is exactly why #672 survived
- [ ] Replacing the new guard with a no-op fails a test

Framing and the original report are @swalla02's in #672 and #675.

be-student:
Closes #699.

`soup migrate` now identifies disguised JSONL only when a bounded 64 KiB prefix contains at least two complete JSON objects on separate lines. Explicit `.jsonl` behavior remains unchanged, while single JSON configs and Unsloth notebooks continue to their selected migrator.

Validation: 20,351 tests passed, 106 skipped, 5 deselected; full Ruff scope passed; changelog fragments validated.

### DIFF
--- a/src/soup_cli/commands/migrate.py
+++ b/src/soup_cli/commands/migrate.py
@@ -1,5 +1,6 @@
 """soup migrate — import configs from LLaMA-Factory, Axolotl, and Unsloth."""
 
+import json
 from pathlib import Path
 
 import typer
@@ -63,11 +64,17 @@ def migrate(
         console.print(f"[red]{exc}[/]")
         raise typer.Exit(1)
 
-    # v0.40.1 Part D / N2 — friendly error when the user passes a JSONL
-    # data file (`.jsonl`) instead of a YAML config. The sniff helper is
-    # only invoked when the suffix says ``.jsonl`` (notebook .ipynb files
-    # legitimately start with ``{`` — we must not falsely flag them).
-    if input_path.suffix.lower() == ".jsonl" and _looks_like_jsonl(input_path):
+    # Keep the permissive content check for explicit .jsonl files. For every
+    # other suffix, require two complete JSON objects on separate lines so a
+    # single JSON config or notebook is not mistaken for training data.
+    suffix = input_path.suffix.lower()
+    is_jsonl = (
+        suffix == ".jsonl" and _looks_like_jsonl(input_path)
+    ) or (
+        suffix != ".jsonl"
+        and _looks_like_jsonl(input_path, require_multiple_objects=True)
+    )
+    if is_jsonl:
         console.print(
             f"[red]Expected a {source} YAML config; got JSONL "
             f"({input_path.name}) — did you pass the wrong file?[/]"
@@ -141,8 +148,8 @@ def migrate(
     console.print(f"[dim]Next: soup train --config {output}[/]")
 
 
-def _looks_like_jsonl(path: Path) -> bool:
-    """v0.40.1 Part D / N2 — sniff first non-blank line for `{` (JSONL).
+def _looks_like_jsonl(path: Path, *, require_multiple_objects: bool = False) -> bool:
+    """Inspect a bounded prefix for explicit or structurally identifiable JSONL.
 
     ``utf-8-sig``, not ``utf-8``: a UTF-8 BOM decodes to U+FEFF, which
     ``str.strip()`` does not remove because it is not whitespace, so a BOM'd
@@ -150,18 +157,32 @@ def _looks_like_jsonl(path: Path) -> bool:
     error. Windows tooling writes that BOM by default, PowerShell's
     ``Out-File`` included (#675 review).
 
-    The read is bounded because a file with no newline is one single line, so
-    ``for line in fh`` would pull all of it into memory: a 120 MB one-liner
-    measured 240.9 MB peak. Only the first non-blank chunk is ever inspected,
-    so a truncated long line costs nothing here.
+    The read is bounded because a file with no newline is one single line. An
+    unbounded line iteration pulled a measured 240.9 MB peak for a 120 MB
+    one-liner. Unknown suffixes use the stricter mode: two non-blank lines must
+    each be complete JSON objects, which separates JSONL from one JSON document.
     """
     try:
         with open(path, "r", encoding="utf-8-sig", errors="replace") as fh:
-            while line := fh.readline(65536):
-                stripped = line.strip()
-                if not stripped:
-                    continue
-                return stripped.startswith("{")
+            prefix = fh.read(65536)
     except OSError:
         return False
+
+    lines = [line.strip() for line in prefix.splitlines() if line.strip()]
+    if not lines:
+        return False
+    if not require_multiple_objects:
+        return lines[0].startswith("{")
+
+    object_count = 0
+    for line in lines:
+        try:
+            value = json.loads(line)
+        except json.JSONDecodeError:
+            return False
+        if not isinstance(value, dict):
+            return False
+        object_count += 1
+        if object_count == 2:
+            return True
     return False
--- a/src/soup_cli/commands/migrate.py
+++ b/src/soup_cli/commands/migrate.py
@@ -157,32 +157,34 @@ def _looks_like_jsonl(path: Path, *, require_multiple_objects: bool = False) -> 
     error. Windows tooling writes that BOM by default, PowerShell's
     ``Out-File`` included (#675 review).
 
-    The read is bounded because a file with no newline is one single line. An
+    Each of at most 64 line reads is bounded because a file with no newline is one single line. An
     unbounded line iteration pulled a measured 240.9 MB peak for a 120 MB
     one-liner. Unknown suffixes use the stricter mode: two non-blank lines must
     each be complete JSON objects, which separates JSONL from one JSON document.
     """
     try:
         with open(path, "r", encoding="utf-8-sig", errors="replace") as fh:
-            prefix = fh.read(65536)
+            object_count = 0
+            for _ in range(64):
+                line = fh.readline(65536)
+                if not line:
+                    break
+                stripped = line.strip()
+                if not stripped:
+                    continue
+                if not require_multiple_objects:
+                    return stripped.startswith("{")
+                if not line.endswith("\n") and len(line) >= 65536:
+                    return False
+                try:
+                    value = json.loads(stripped)
+                except json.JSONDecodeError:
+                    return False
+                if not isinstance(value, dict):
+                    return False
+                object_count += 1
+                if object_count == 2:
+                    return True
     except OSError:
         return False
-
-    lines = [line.strip() for line in prefix.splitlines() if line.strip()]
-    if not lines:
-        return False
-    if not require_multiple_objects:
-        return lines[0].startswith("{")
-
-    object_count = 0
-    for line in lines:
-        try:
-            value = json.loads(line)
-        except json.JSONDecodeError:
-            return False
-        if not isinstance(value, dict):
-            return False
-        object_count += 1
-        if object_count == 2:
-            return True
     return False
--- a/changelog.d/0.74.0/730.fixed.md
+++ b/changelog.d/0.74.0/730.fixed.md
@@ -1,0 +1,6 @@
+- `soup migrate` now recognizes multi-record JSONL by structure even when the
+  filename does not end in `.jsonl`, so a data file such as `data.json` gets the
+  existing friendly wrong-file error instead of a raw YAML `ParserError`.
+  Single JSON objects and Unsloth notebooks still reach their selected migrator,
+  and the structural sniff remains bounded to 64 KiB (#730 by @be-student,
+  closes #699).
--- a/tests/test_v0401_part_d.py
+++ b/tests/test_v0401_part_d.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import json
 from pathlib import Path
 
+import pytest
 from typer.testing import CliRunner
 
 # --- H4: Template help is dynamically generated --------------------------
@@ -80,15 +81,18 @@ def test_migrate_jsonl_input_yields_friendly_error(tmp_path: Path, monkeypatch):
     assert "got JSONL" in result.output
 
 
+@pytest.mark.parametrize("record_size", [2, 40000])
+@pytest.mark.parametrize("trailing_newline", [True, False])
 def test_migrate_jsonl_with_json_suffix_yields_friendly_error(
-    tmp_path: Path, monkeypatch
+    tmp_path: Path, monkeypatch, record_size: int, trailing_newline: bool
 ):
     """Multiple JSON objects are data even when the suffix does not say JSONL."""
     from soup_cli.cli import app
 
     monkeypatch.chdir(tmp_path)
     data = tmp_path / "data.json"
-    data.write_text('{"prompt": "hi"}\n{"prompt": "world"}\n', encoding="utf-8")
+    record = json.dumps({"prompt": "x" * record_size})
+    data.write_text(f"{record}\n{record}" + ("\n" if trailing_newline else ""), encoding="utf-8")
 
     result = CliRunner().invoke(
         app, ["migrate", "--from", "llamafactory", "data.json", "--dry-run"]
--- a/tests/test_v0401_part_d.py
+++ b/tests/test_v0401_part_d.py
@@ -10,6 +10,7 @@ Closes:
 
 from __future__ import annotations
 
+import json
 from pathlib import Path
 
 from typer.testing import CliRunner
@@ -77,6 +78,86 @@ def test_migrate_jsonl_input_yields_friendly_error(tmp_path: Path, monkeypatch):
     )
     assert result.exit_code == 2, (result.output, repr(result.exception))
     assert "got JSONL" in result.output
+
+
+def test_migrate_jsonl_with_json_suffix_yields_friendly_error(
+    tmp_path: Path, monkeypatch
+):
+    """Multiple JSON objects are data even when the suffix does not say JSONL."""
+    from soup_cli.cli import app
+
+    monkeypatch.chdir(tmp_path)
+    data = tmp_path / "data.json"
+    data.write_text('{"prompt": "hi"}\n{"prompt": "world"}\n', encoding="utf-8")
+
+    result = CliRunner().invoke(
+        app, ["migrate", "--from", "llamafactory", "data.json", "--dry-run"]
+    )
+
+    assert result.exit_code == 2, (result.output, repr(result.exception))
+    assert "got JSONL" in result.output
+
+
+def test_migrate_single_json_object_is_not_reported_as_jsonl(tmp_path: Path, monkeypatch):
+    """A valid single-object config must continue to the selected migrator."""
+    from soup_cli.cli import app
+
+    monkeypatch.chdir(tmp_path)
+    config = tmp_path / "config.json"
+    config.write_text(
+        json.dumps(
+            {
+                "model_name_or_path": "meta-llama/Llama-3-8B",
+                "stage": "sft",
+                "finetuning_type": "lora",
+            }
+        ),
+        encoding="utf-8",
+    )
+
+    result = CliRunner().invoke(
+        app, ["migrate", "--from", "llamafactory", "config.json", "--dry-run"]
+    )
+
+    assert result.exit_code == 0, (result.output, repr(result.exception))
+    assert "got JSONL" not in result.output
+
+
+def test_migrate_unsloth_notebook_is_not_reported_as_jsonl(tmp_path: Path, monkeypatch):
+    """A multi-line notebook remains a single JSON document and still migrates."""
+    from soup_cli.cli import app
+
+    monkeypatch.chdir(tmp_path)
+    notebook = tmp_path / "train.ipynb"
+    notebook.write_text(
+        json.dumps(
+            {
+                "cells": [
+                    {
+                        "cell_type": "code",
+                        "source": [
+                            "model, tokenizer = FastLanguageModel.from_pretrained(\n",
+                            "    model_name='unsloth/llama-3',\n",
+                            ")\n",
+                            "trainer = SFTTrainer()\n",
+                        ],
+                    }
+                ],
+                "metadata": {},
+                "nbformat": 4,
+                "nbformat_minor": 5,
+            },
+            indent=2,
+        ),
+        encoding="utf-8",
+    )
+
+    result = CliRunner().invoke(
+        app, ["migrate", "--from", "unsloth", "train.ipynb", "--dry-run"]
+    )
+
+    assert result.exit_code == 0, (result.output, repr(result.exception))
+    assert "got JSONL" not in result.output
 
 
 def test_migrate_yaml_config_named_jsonl_still_migrates(tmp_path: Path, monkeypatch):


MakazhanAlpamys:
@be-student — welcome, and thank you for the PR. **I have to ask you to hold this one, and the reason is partly my failure rather than yours.**

## The situation

**#699 was claimed by @swalla02 on 2026-09-05 at 18:07 UTC**, and I endorsed that publicly at 18:36 with an explicit ordering:

- @v01dst holds a right of first refusal on it **until 2026-09-06 18:00 UTC** (that is roughly three hours from now, and has not expired);
- after that it passes to @swalla02, who has been working on it since yesterday evening and offered to hand over their branch and measurements rather than make anyone redo the probes.

@swalla02 also has direct standing here: they filed the original report in #672, wrote #675 which is the fix this issue was split out of, and the framing in #699's body is theirs.

So this is the third-in-line position, and I am not going to take it from them.

## What is my fault, and what I have changed

**The issue was labelled `help wanted`, had no assignee, and carried the claim only inside the comment thread.** If you were scanning open issues you would have seen an unclaimed `help wanted` bug. That is a bad signal and it is mine.

Two things I have just found while checking this:

1. **No outside contributor is assignable in this repository at all.** `GET /repos/.../assignees/<user>` returns 404 for every one I tested. GitHub only permits assigning a collaborator or someone who has already commented on that specific issue, and `POST .../assignees` returns **200 with an empty array** rather than an error when it refuses. So "assign the claimant" is not a control I actually have, and I had been relying on it.
2. I have therefore added a **`claimed`** label and removed `help wanted` from every issue currently held — #699, #677, #716, #702, #694. That signal shows in the issue *list*, not only inside the thread, which comments and assignees do not.

This is the fifth claim collision here in four days and every previous one produced an apology rather than a mechanism. The label is the mechanism.

## What happens to #730

**Held, not closed**, with a date rather than a maybe:

- If @swalla02 opens their PR, I review theirs and this one stands down — but I will read yours first and, if it solves something theirs does not, say so on their thread and credit you for it.
- **If no PR from @swalla02 by 2026-09-08 12:00 UTC**, #730 gets reviewed on its merits and merged if it holds up. I am not letting a working PR rot behind an unexercised claim.

@swalla02 — you have the floor. If you would rather hand this to @be-student, that also costs you one word and I will make sure the handover is credited both ways.

## Meanwhile, so the round is not wasted

Your description says `20,351 tests passed`, and the approach — bounded 64 KiB prefix, at least two complete JSON objects on separate lines — is the shape #699's body actually asks for, including reusing the existing bounded reader rather than writing a second one. Two things I would check on any PR here, worth doing now either way:

- **The `.ipynb` case is the one that must not break.** It is why the suffix gate exists at all: a notebook is a single JSON object beginning with `{`, and `soup migrate --from unsloth` takes those legitimately. Make sure a real notebook still migrates, driven through `CliRunner` rather than by calling the helper — a helper-only test is exactly why #672 survived in the first place.
- **Replace your new guard with a no-op and confirm a test fails.** That is required of every fix in this repo, and it is the check that separates a real regression test from one that passes on the broken code too.

Your CI was in `action_required` as a first-time contributor; it is approved and running now.

swalla02:
I have stood down on #699 so this can merge. Not a competing review, just the one thing my branch found that this does not cover.

## Real JSONL is misclassified when the records are large

`fh.read(65536)` then `splitlines()` leaves the final line truncated whenever the file exceeds the buffer. In strict mode the loop hits that partial line, `json.loads` raises, and the function returns `False`.

Two 40 KiB records named `data.json`, which is exactly the case this PR exists to catch:

```
two 40 KiB records, .json  -> False   (want True)
two small records,  .json  -> True
prefix lines: 2 | line1 complete: True | line2 complete: False
```

Run against your `_looks_like_jsonl` as written in this diff, no other changes.

It fails in the safe direction, the user gets the raw `ParserError` they get today, so this is a gap rather than a regression. Long-context SFT rows go past 64 KiB fairly easily though.

## Two ways to close it

**One line, keeps your approach.** Discard the last entry when the read filled the buffer, since it may be partial:

```python
prefix = fh.read(65536)
truncated = len(prefix) == 65536
lines = [l.strip() for l in prefix.splitlines() if l.strip()]
if truncated and lines:
    lines.pop()
```

That moves the threshold rather than removing it: two records that individually exceed 64 KiB still lose. Worth a test either way.

**Or remove the ceiling.** Stream the file and track brace depth instead of parsing, treating a return to depth 0 at a newline as a record boundary. Record size stops mattering and memory is O(1). The catch is that a bare `{` inside a JSON string is legal, so `{"prompt": "def f() { return 1"}` is a valid line whose naive depth never returns to 0. It needs in-string and backslash-escape tracking to be correct, which is a small hand-rolled scanner where `json.loads` is already right.

I would ship the one-liner here and leave the scanner for a follow-up if anyone ever hits the wider case. Your call and @MakazhanAlpamys's.

Nice fix, and thanks for keeping the `utf-8-sig` reasoning from #675 intact rather than rewriting the helper around it.

MakazhanAlpamys:
@be-student — **the hold is lifted. @swalla02 stood down on #699 at 16:28 and asked explicitly that this proceed on its merits, so it does.** Thank you for waiting on a queue that was not your doing.

The fix is correct, the mutations all die, and the `.ipynb` case — the one that had to not break — is covered through `CliRunner` rather than by calling the helper. One thing to add before I merge, and it is the finding @swalla02 handed you rather than one of mine.

## What I verified

Four mutations, four killed, on your head:

| mutation | dies by |
|---|---|
| strict structural branch -> `return False` (the guard neutered) | `test_migrate_jsonl_with_json_suffix_yields_friendly_error` |
| `is_jsonl = False` at the call site | that test + `test_migrate_jsonl_input_yields_friendly_error` + `test_migrate_bomd_jsonl_still_yields_friendly_error` |
| drop `require_multiple_objects=True`, i.e. permissive sniff on every suffix — **the exact regression #699 forbids** | `TestMigrateCLI::test_unsloth_notebook`, `test_migrate_single_json_object_is_not_reported_as_jsonl`, `test_migrate_unsloth_notebook_is_not_reported_as_jsonl` |
| threshold `object_count == 2` -> `== 1` | `TestMigrateCLI::test_unsloth_notebook`, `test_migrate_single_json_object_is_not_reported_as_jsonl` |

74 passed across the migrate surface, 77 on the two files directly, ruff clean, zero deleted or weakened test lines. You also kept @swalla02's `utf-8-sig` reasoning from #675 intact instead of rewriting the helper around your change, which is why their BOM test still means something.

Both of the checks I asked you to run in my hold comment came back the way they should: a real notebook still migrates through the CLI, and neutering your guard fails a test.

## The one thing to add

I reproduced @swalla02's finding on your head myself rather than forwarding it:

```
module: _wt730\src\soup_cli\commands\migrate.py
two 40000-char records, .json (80032 bytes) -> False   (want True)
two small records,      .json               -> True    (want True)
```

And I binary-searched the boundary so the rule is stated as a number rather than as "large files":

```
largest per-record payload that still detects: 32753 chars -> file 65538 bytes
```

**The first two records must fit inside the 64 KiB read together**, i.e. roughly 32 KiB each. `prefix = fh.read(65536)` leaves the final line truncated, the strict loop hits it, `json.loads` raises, and you return `False` before ever reaching the `object_count == 2` early return. A long-context SFT row goes past 32 KiB without being unusual.

It fails in the safe direction — the user gets the same raw `ParserError` they get today, which I confirmed end to end — so this is a gap, not a regression, and it is the reason I am asking for one commit rather than reopening the review.

@swalla02's one-liner, which I would take as-is:

```python
prefix = fh.read(65536)
truncated = len(prefix) == 65536
lines = [l.strip() for l in prefix.splitlines() if l.strip()]
if truncated and lines:
    lines.pop()
```

Plus a test at the shape that fails today: two ~40 KiB records in a file named `data.json`, asserted through `CliRunner` for the friendly error. Please confirm it fails without the guard — that is the check that separates the regression test from one that would have passed anyway.

Their second option — a brace-depth scanner with in-string and escape tracking — removes the ceiling entirely but hand-rolls what `json.loads` already gets right. I agree with their own recommendation: ship the one-liner, leave the scanner for whoever actually hits the wider case.

## On the collision

You were third in line on #699 through no signal you could have read, and the reason is on me: the issue carried `help wanted` with the claim buried in the comment thread. I have since applied the `claimed` label across every issue with an open PR against it. Your PR is the one shipping, and @swalla02's finding is credited in the changelog next to your fix.

Your `changelog.d/0.74.0/730.fixed.md` is correct as written; just extend it if the threshold behaviour changes. Push that commit and I merge on green.

src/soup_cli/commands/migrate.py
```diff
@@ -141,27 +148,41 @@ def migrate(
...
+    each be complete JSON objects, which separates JSONL from one JSON document.
     """
     try:
         with open(path, "r", encoding="utf-8-sig", errors="replace") as fh:
-            while line := fh.readline(65536):
-                stripped = line.strip()
-                if not stripped:
-                    continue
-                return stripped.startswith("{")
+            prefix = fh.read(65536)
```
AmirF194, line 167: I ran this against 48b23affb5356dbb7e5e02dd6065e2bfdbf146c6 in a clean python:3.12-slim container.

The stricter sniff at migrate.py:167 reads a flat 64 KiB `prefix` and requires two lines that each parse as a complete JSON object before it will call a misnamed file JSONL. A real multi-record JSONL file with moderately large records, long-context training examples well under any per-record size limit, straddles that bound: the first record parses fine, the second gets cut mid-object by the 64 KiB read, `json.loads` raises on it, and the function returns False at line 186's loop before object_count ever reaches 2. The file falls through to the normal migrator and hits the raw YAML ParserError this PR exists to replace.

```
$ python repro.py
file size: 68068 bytes (prefix bound is 65536)
detected as JSONL: False
```
Two records, each `{"prompt": "x"*34000, "completion": "y"}` on its own line, in a file named data.json. End to end through the actual CLI (`soup migrate --from llamafactory data.json --dry-run`):

```
exit_code: 1
exception: ParserError(None, None, "expected '<document start>', but found '{'", ...)
```

The old permissive `.jsonl`-suffix path does not have this failure mode: it checks only whether the first non-blank line starts with `{`, so a truncated line still resolves correctly. The new stricter path could keep that resilience by parsing the first complete line first (cheap, always resolves within the 64 KiB bound for any file with a second line at all) and only requiring a full second parse when the first line alone isn't decisive, rather than needing both records to fully fit inside one fixed-size read.

MakazhanAlpamys:
@be-student — **stop before you write that commit. The fix I endorsed does not work, and that is my error, not yours or @swalla02's.**

I reproduced the defect, binary-searched its boundary, quoted a one-liner as "which I would take as-is", and told you to push it and I would merge on green. **I never ran the one-liner against the case I had just reproduced.** It is a no-op with respect to that case.

## The measurement I should have taken before endorsing it

Eight cases, three implementations. `shipped` is your current head; `+swalla02` is your head with the endorsed one-liner applied verbatim to the strict branch:

```
case                           size   want  shipped  +swalla02  per-line
two 40000-char records        80070   True   False*     False*     True
two 34000-char records        68070   True   False*     False*     True
two 20000-char records        40070   True    True       True      True
two small records                18   True    True       True      True
ONE minified document            28  False   False      False     False
pretty-printed config            22  False   False      False     False
notebook                         37  False   False      False     False
200KB single long line       200008  False   False      False     False

* = wrong answer
```

The `+swalla02` column is identical to `shipped` on **all eight**. Dropping the truncated final line stops the `JSONDecodeError`, but you then have one complete object instead of two, so the loop runs out and returns `False` anyway. Same answer, different route. The two 20000-char rows pass in both columns because 2 x 20070 already fits inside 64 KiB — they were never failing.

So the fix I asked you for would have changed nothing, you would have pushed it, and I would have merged a PR believing a gap was closed that was not. The reason is exactly the standard this repo applies to everyone else: I asserted a fix works from reading it.

## The real constraint

It is not the truncated line. It is that **two complete records must fit inside one cumulative 64 KiB read**, and for long-context SFT rows they simply do not. Dropping the truncated line cannot conjure a second complete object out of a prefix that only contains one.

## What does work

Bound **each line** rather than the total prefix. `readline(65536)` caps a single read at 64 KiB, so the 120 MB one-liner that motivated the original bound still reads only 64 KiB, while a 40 KiB record is returned whole:

```python
count = 0
with open(path, "r", encoding="utf-8-sig", errors="replace") as fh:
    for _ in range(64):                      # bounded number of lines
        line = fh.readline(65536)
        if not line:
            break
        # A line with no newline is TRUNCATED only if it hit the read limit;
        # otherwise it is the genuine final line of the file.
        if not line.endswith("\n") and len(line) >= 65536:
            return False
        stripped = line.strip()
        if not stripped:
            continue
        try:
            value = json.loads(stripped)
        except json.JSONDecodeError:
            return False
        if not isinstance(value, dict):
            return False
        count += 1
        if count == 2:
            return True
return False
```

**My first version of this was also wrong**, and I am telling you so you do not rediscover it: I wrote `if not line.endswith("\n"): return False`, which rejects any JSONL file with no trailing newline on its last line — a common shape. Two records with no final newline came back `False`. The length check is what separates "truncated by the limit" from "end of file".

Verified on all eight cases above plus `two 40000 records with no final newline`, `two small with no final newline`, and `one 100000-char record only` (correctly `False` — one object is not JSONL).

## Credit, and a second remedy that also does not survive testing

**@swalla02 found this first**, on 2026-09-07 at 16:28, having already stood down on #699 so this PR could proceed. **@AmirF194 reproduced it independently** at 23:40 in a clean `python:3.12-slim` container with a different record size, arriving at the same boundary from the other direction. The finding is right in both cases; only the remedies are wrong, and I am the one who promoted one of them to a merge condition.

@AmirF194's suggestion — decide on the first complete line when it is decisive — reintroduces the exact false positive the two-object rule exists to prevent:

```
ONE minified json document (one line, parses as a dict)  ->  detected as JSONL: True   (want False)
```

That is a real input for `soup migrate --from llamafactory`, so it would refuse a legitimate JSON config as "training data". Worth stating plainly so it is not adopted later on the strength of the finding being correct.

## What I am asking for now

The per-line version above, plus the regression test I asked for before — two ~40 KiB records in a file named `data.json`, driven through `CliRunner` for the friendly error. Please confirm it **fails on your current head** before adding the fix; that is what separates a regression test from one that would have passed anyway, and it is the check I skipped on myself.

If you would rather not do a second round on a moving target, say so and I will write both and merge with the work credited to you — your commits, your `CONTRIBUTORS.md` entry, your changelog fragment, with @swalla02 and @AmirF194 credited for the finding. You have already been held once on this issue through a queue that was not your doing, and I have now sent you down a wrong path on top of it.

Your `changelog.d/0.74.0/730.fixed.md` still stands as written.

src/soup_cli/commands/migrate.py
MakazhanAlpamys: @AmirF194 — reproduced, and your diagnosis is right. Thank you for running it in a clean container rather than reasoning about the bound.

Two things you should know, both posted in full in the PR thread.

**You were second to this, and only by hours.** @swalla02 found the same boundary at 16:28 the same day, from a different record size. That takes nothing away from an independent reproduction — it is the second one that told me the boundary was not an artefact of one person's fixture.

**Your suggested remedy does not survive testing, and neither did the one I endorsed.** I measured both:

```
                                     want   +swalla02's fix   your fix
two 34000-char records (.json)       True       False*         True
ONE minified json document          False       False          True*

* = wrong answer
```

Deciding on the first complete line when it parses as a dict reintroduces exactly the false positive the two-object rule exists to prevent: a single minified JSON document on one line is then detected as JSONL, and `soup migrate --from llamafactory config.json` would refuse a legitimate config as training data. The one-liner I had already endorsed is worse — it changes nothing at all on any of the eight cases I tested, because dropping the truncated line leaves one complete object where two are required.

What does work is bounding **each line** instead of the cumulative prefix: `readline(65536)` still caps a 120 MB one-liner at 64 KiB, which is why the bound exists, while returning a 40 KiB record whole. The code and the full table are in the thread.

The mistake worth naming is mine, not yours: I reproduced the defect, quoted a remedy as "which I would take as-is", and never ran it against the case I had just reproduced.

AmirF194: You're right, the first-line-decisive approach fails on a single minified JSON document, I didn't test that case before suggesting it. readline(65536) per line is the correct fix. Thanks for tracking it down and posting the full test table.

MakazhanAlpamys:
@AmirF194 — worth saying in the thread rather than only in a resolved inline comment: you tested a suggestion of your own, found it wrong, and said so in two sentences without being asked twice. That is the cheapest and rarest thing in a review queue, and it is the reason the wrong remedy did not get adopted on the strength of a correct finding. Thank you.

@be-student — putting the same date on this that I just put on #737, so you are not tracking three of my threads with three different clocks.

**2026-09-11 18:00 UTC.** Nothing new is asked; the state is exactly my 2026-09-08 comment above:

- the per-line `readline(65536)` version, and
- the regression test — two ~40 KiB records in a file named `data.json`, through `CliRunner` — **confirmed failing on your current head before the fix goes in**.

Your `changelog.d/0.74.0/730.fixed.md` stands as written.

If you would rather not do a second round on a target I moved under you, say so or say nothing and let the date pass, and I will write both myself and merge with your commits, your `CONTRIBUTORS.md` entry and your fragment, with @swalla02 and @AmirF194 credited for the finding.

**Of your three open PRs I want #737 first** — it is the only one with someone else queued behind it. This one is second and #736 is a cost question rather than a correctness one, so it can wait behind both.

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:
@be-student — reviewed at `6bba3069`. **The code is correct and I would merge it as-is.** One mechanical change and it goes in; details below, including two things that are mine rather than yours.

## I reproduced the defect myself before looking at your diff

I did not take the thread's framing — it has been wrong once already on this PR. I built 29 inputs and drove every one through `CliRunner` against `origin/main` first. Baseline: a `.jsonl` suffix gives the friendly `got JSONL` error, and **every one of 14 other JSONL spellings** gives a raw `ParserError` at exit 1. That is #699 exactly — the suffix decides, not the content.

On your head, **12 of those 14 now return the friendly error**, and every must-not-break case is untouched: a real Unsloth notebook minified *and* pretty-printed, a YAML config, a single JSON object, a top-level array, a 200 KB one-liner. All still exit 0.

The three residual misses all fail in the **safe** direction — the user gets the same raw error they get today, not a wrong migration. Two of them are the per-line ceiling by design, and I binary-searched it: detection holds to a 65,522-char line and gives up at 65,552. So the bound moved from *"the first two records together under 64 KiB"* to *"each record under 64 KiB"* — a 2x improvement, exactly what the design says.

**And I confirmed the thing you could not check yourself:** your new test genuinely fails on the previous head. I checked out `48b23aff`, dropped in your test file, and got `2 failed, 17 passed` — both `[*-40000]` cases on `assert 1 == 2`. It is a real regression test, not one written around behaviour that already worked.

I also re-derived, independently, that the `lines.pop()` one-liner I withdrew really was a no-op: it trims a 64 KiB *prefix*, so dropping the truncated final line leaves one complete object where two are needed, and the function still returns `False`. You were right to ignore it and implement the per-line version instead — including the `len(line) >= 65536` half of the truncation test, which my own first draft got wrong and which your `trailing_newline=False` parametrization now pins.

Mutations: 14 applied, 11 killed. All four I named die — invert the verdict, always-JSONL, always-JSON, and break-the-fallback — and the last one dies on `test_unsloth_notebook`, which is precisely the case the suffix gate exists to protect. Nothing deleted or weakened: `+85/-0` against main, and between the reviewed commit and this head the test got *stronger*, from one hard-coded case to a 2x2 parametrization. CI went green by fixing the code.

## One change

`changelog.d/0.74.0/730.fixed.md` is stale — v0.75.0 merged to `main` at 08:26 UTC and emptied that directory. I simulated your rebase and got the failure exactly:

```
ChangelogError: changelog.d/0.74.0/730.fixed.md: fragment baseline is 0.74.0,
```

```
git mv changelog.d/0.74.0/730.fixed.md changelog.d/0.75.0/730.fixed.md
```

While you are in the file, two small things — the second is a promise of mine, not a request:

- it says the sniff "remains bounded to 64 KiB", which is now true **per line** rather than per file. One word.
- I said in my 2026-09-08 comment that the fragment would credit **@swalla02** and **@AmirF194** for the finding. It does not, and that is my omission to flag rather than yours to have guessed.

## Two things that are mine, not yours

**A regression I introduced through my own snippet.** `for _ in range(64)` caps *total* line reads, and `if not stripped: continue` still spends an iteration — so blank lines draw on the budget. Measured: on `main` a `.jsonl` file with 64, 70, 200 or 5000 leading blank lines was detected; on your head it stops being detected at exactly 64. It fails safe and no real file looks like that, but it is a behaviour change on a path that worked, and it came from the code I handed you. **I will fix it separately** — the fix needs two counters rather than one, because a bare `range(64)` over non-blank lines re-opens the 120 MB-of-newlines read the bound exists to prevent, and that is not something to bolt onto your PR.

**Three surviving mutations that I am deliberately not asking you to close.** Deleting the truncation guard, dropping `isinstance(value, dict)`, and narrowing the gate from "any non-`.jsonl` suffix" to "`.json` only" all survive the suite. I tried to justify the first as load-bearing and mostly could not — `json.loads` already rejects a truncated line, and the only input that distinguishes guard-from-no-guard is 70,000 spaces inside a record, where the guard gives the *less* useful answer. The other two are untested widenings rather than dead code: your `.yaml`-with-JSONL-content behaviour is real, it just has no test. My standing rule is that a surviving mutation blocks, so I am stating the exception rather than burying it: these are defensive lines whose removal changes nothing observable, and you have already been held through two rounds and sent down one wrong path by me. A third round for them is not a good trade.

One pre-existing note, not yours: `test_v0401_part_d.py` asserts on raw `result.output`, and `"got JSONL"` straddles a Rich wrap at COLUMNS 41-46. **`origin/main` fails there too** — your PR adds four parametrizations of the same pre-existing pattern. Green at 80 and in real CI. Candidate for the ANSI-normaliser helper whenever that file is next touched.

`main` is frozen until v0.75.0 is tagged, so the rebase can wait for the tag if you prefer. Ping me and I will merge.

MakazhanAlpamys:

```
git mv changelog.d/0.74.0/730.fixed.md changelog.d/0.75.0/730.fixed.md
git push --force-with-lease
```

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:
Verified `48ef0c3e`. **Merge-ready** — the fragment item is done and the per-line rewrite is genuinely pinned.

The fragment moved out of `changelog.d/0.74.0/`, now says **"bounded to 64 KiB per line"**, and credits *"finding by @swalla02 and @AmirF194"* — which I flagged as an omission and you took without argument. `merge-tree` against current `main` is clean.

**Mutations — both killed.** Anchors counted at exactly 1, module imported after each, reverted green:

- `object_count == 2` widened to `== 1`: 1 failed — `test_migrate_single_json_object_is_not_reported_as_jsonl`, i.e. exactly the false positive #699 forbids.
- **the per-line bound**, `fh.readline(65536)` narrowed to `1024`: 2 failed — and the interesting part is *which* two. Only the 40 KiB parametrizations die; both 2-byte ones survive. The test discriminates the **record-size ceiling specifically**, which is the thing the rewrite exists to move, rather than just noticing that something changed.

Baseline: swapping in `main`'s `migrate.py` fails all four new parametrizations. 19 passed on the file, 79 passed across `-k migrate`, ruff clean, zero deleted test lines (+85/−0).

Scope is exactly #699 and nothing else: `.jsonl` fast path untouched at `migrate.py:74`, the structural sniff gated behind `suffix != ".jsonl"`, `_looks_like_jsonl` extended rather than duplicated, and the two-object threshold at `:186`.

Three notes, none blocking:

1. **The wrap liability grows 3x.** The new assertions use raw `assert "got JSONL" in result.output`. Not a colour problem — `FORCE_COLOR=1` gives 19 passed — but a Rich *wrap* problem: at `COLUMNS=44` this head gives 6 failed against `main`'s 2 failed. CI runs at 80 so it is green, and the pattern is pre-existing rather than yours. Worth naming because it turns a 2-test liability into a 6-test one; a candidate for the `strip_ansi`/`_plain` normaliser next time that file is touched.
2. **The `for _ in range(64)` budget now bounds the permissive `.jsonl` path too**, where `main` had an unbounded `while line := fh.readline(65536)`. So the 64-leading-blank-lines regression you attributed to your own snippet applies to *both* modes, not just the strict branch. You said you would fix it separately with two counters — recording here that it is wider than we both thought.
3. The credit tail reads `(#730 by @be-student, finding by @swalla02 and @AmirF194, closes #699)` rather than the canonical `(#699 by @be-student in #730)`. It validates, since the guard only requires the filename number to appear in the text, and I would rather keep the two finders named than have it match its siblings. Leaving it.

Only CI is left; the Actions queue is at ~88 runs. I will merge as soon as it is green.

MakazhanAlpamys:
Re-verified against current `main` (`b0a63382`), 16 merges after my review. **Unchanged and still merge-ready:** `merge-tree` clean, 19 passed on both the PR head and the merged tree, ratchets 138 passed, fragment intact.

The mutation still discriminates, and it fails in the most useful possible way: `if object_count == 2:` to `== 3` fails all four parametrisations of `test_migrate_jsonl_with_json_suffix_yields_friendly_error` with exactly the

```
ParserError("expected '<document start>', but found '{'")
```

that this PR exists to replace. A test whose failure message *is* the original bug report is about as clear as it gets.

Nothing needed from you.

MakazhanAlpamys:
Merged as `b4819374`. Thank you — and this waited far longer than the work deserved, for a reason that was never yours: GitHub Actions on this repo produced no terminal outcome at all between 2026-09-13T18:30Z and today 15:08Z, and roughly a hundred runs were stacked behind that. Cleared now.

**Merged on the verdict recorded at `48ef0c3e`**, which is still the head — not re-derived. What I did re-check is what can change while a PR waits: all 13 required contexts green on that exact SHA (read per-SHA, not from the cached PR summary), `mergeStateStatus` CLEAN, and whether `main` moved into anything your change interacts with. It did — #757 landed today with a package-wide AST scanner that walks **every** command module looking for typer parameters left unfilled at a call site, and you change `commands/migrate.py`. Green on both branches separately is not evidence about the merge, so I built it: **51 passed** on the merged tree across your file, that scanner, and the fragment validator.

**The mutation still discriminates, and it fails in the most useful way I have seen this week.** Changing `if object_count == 2:` to `== 3` fails all four parametrisations of `test_migrate_jsonl_with_json_suffix_yields_friendly_error` with exactly:

```
ParserError("expected '<document start>', but found '{'")
```

That is #699 verbatim. A test whose failure message *is* the original bug report needs no comment explaining what it protects.

**What I want on the record about the design.** Deciding the format from the suffix is the kind of shortcut that looks harmless until the error it produces is a YAML traceback about a file that was never YAML — the user has no thread to pull, because nothing in the message mentions JSONL. You could have special-cased `.json`-containing-JSONL and closed the ticket. Instead you made detection structural, and then **bounded it**: a 64 KiB prefix, at least two complete objects on separate lines. Both halves matter — unbounded content sniffing is its own DoS, and requiring *two* objects is what stops a single JSON config being misread as a one-line stream. Explicit `.jsonl` unchanged, single JSON configs and Unsloth notebooks still routed to their own migrator.

Credited in `CONTRIBUTORS.md`. #699 is closed with credit.

PR #732, opened 2026-09-06
fix(ui): keep the FastAPI docs routes off a non-loopback bind (#731)

### LINKED ISSUE #731: soup ui serves /openapi.json and /docs unauthenticated on a --public bind
## The gap

`soup ui` serves FastAPI's interactive documentation and schema without authentication:

```
/openapi.json          200 unauthenticated
/docs                  200
/docs/oauth2-redirect  200
/redoc                 200
```

Every other private read now requires a Bearer token after #707 (`0347062`). These four do not, so on a `soup ui --public` bind — which is a first-class product path, complete with a scannable QR code — anyone on the LAN can enumerate the full API surface: every route, every parameter, every request and response schema.

## What is and is not at stake

**No run data, no configuration and no logs are exposed.** The schema describes shapes, not values, and every endpoint it describes answers 401 without a token. So this is reconnaissance, not disclosure, and it is the reason this was filed rather than folded into #707.

It is also **identical on `main`** and predates #707 — verified by execution against both trees. #707 did not introduce it and did not widen it.

The reason it is worth closing anyway is that it now sits alone: after #707 the non-loopback story is "everything private needs a token", and four routes quietly not following that rule is the kind of inconsistency that gets rediscovered as a surprise later.

## Options, and the trade-off is real

1. **Gate them behind the same `_verify_token` dependency.** Simplest and consistent. Cost: `/docs` becomes unusable in a browser, because the Swagger page cannot send a Bearer header on its initial navigation — the same constraint that made SSE hard in #687. A developer on loopback loses a convenience they may be using.
2. **Disable them entirely when the bind is non-loopback**, leaving them open on `127.0.0.1`. Keeps the loopback convenience, closes the LAN exposure, and matches the shape of the refusal #707 already added. Probably the right answer.
3. **Leave them and document it.** Defensible given nothing is disclosed, but then say so in `docs/serving-and-export.md` rather than leaving it implicit.

I lean towards 2. Whoever takes this should say which they picked and why, rather than treating it as obvious.

## Acceptance criteria

- [ ] On a non-loopback bind, `/openapi.json`, `/docs`, `/docs/oauth2-redirect` and `/redoc` do not serve the schema to an unauthenticated client
- [ ] On a loopback bind the chosen behaviour is deliberate and stated, not incidental
- [ ] `/api/health` and the static SPA at `/` remain reachable unauthenticated — the dashboard cannot load otherwise, and #687 called that out explicitly
- [ ] A test drives the real ASGI app for both bind kinds; mutating the new guard away fails it
- [ ] The chosen behaviour is documented where `soup ui --public` is documented

## Notes

Found while reviewing #707 (@kok-o), whose route audit is what made these four visible as the remainder. Related: #687.

Srinivasan8888:
Closes #731.

## The choice, and why option 1 is not merely less convenient

You leaned towards option 2 and asked whoever took this to say why rather than treat it as obvious. **Option 1 does not close the hole.**

`/docs` is a browser navigation, so Swagger cannot attach a Bearer header to its initial request — the same constraint that made SSE hard in #687. Gating the four routes behind `_verify_token` therefore makes `/docs` unusable for the developer **without making the schema unreadable**: `curl -H 'Authorization: Bearer …' /openapi.json` still returns the full surface to anyone who has, guesses, or later obtains a token, and the route keeps existing. The human loses the page; the enumerator loses nothing they could not get another way. Option 2 removes the handler, so there is nothing left to reach at any credential level.

Option 3 leaves exactly the inconsistency the issue was filed to remove.

## The fix

```python
_docs_enabled = _is_loopback(host)
app = FastAPI(
    title="Soup Web UI",
    version="1.0.0",
    openapi_url="/openapi.json" if _docs_enabled else None,
    docs_url="/docs" if _docs_enabled else None,
    redoc_url="/redoc" if _docs_enabled else None,
    swagger_ui_oauth2_redirect_url=(
        "/docs/oauth2-redirect" if _docs_enabled else None
    ),
)
```

`swagger_ui_oauth2_redirect_url` is passed explicitly because **that route is derived from it, not from `docs_url`** — leaving it at its default keeps `/docs/oauth2-redirect` serving after `/docs` is gone. It is in your acceptance list and it is the one a narrower fix drops.

## Live verification — two real servers, side by side

Not `TestClient`. Two actual `soup ui` processes on this machine at the same time, curled over the **LAN interface** rather than loopback:

```
$ soup ui --public --port 7873 --no-browser --auth-token …
$ soup ui          --port 7874 --no-browser --auth-token …

=== 192.168.0.105:7873 (--public), unauthenticated ===
  /openapi.json            404
  /docs                    404
  /docs/oauth2-redirect    404
  /redoc                   404
  /api/health              200      <- control, #687 requires it
  /                        200      <- control, the SPA cannot load otherwise
  /api/runs                401      <- #707 story untouched

=== the same LAN server, WITH a valid Bearer token ===
  /openapi.json            404      <- removed, not gated
  /docs                    404
  /api/runs                200
  /api/system              200

=== 127.0.0.1:7874 (default bind), unauthenticated ===
  /openapi.json            200
  /docs                    200
  /docs/oauth2-redirect    200
  /redoc                   200
  schema: "Soup Web UI" - 24 paths described
```

The token rows are the point: 404 for a credentialed client is what distinguishes removal from gating, and it is the difference between this and option 1.

## Mutation table

`tests/test_issue731_ui_docs_bind.py tests/test_issue687_ui_auth.py`, baseline **57 passed**. `__pycache__` cleared before each round; `app.py` md5-verified byte-identical after (`fccd5a29b6896b65bde3d4b59078f505`).

|---|---|---|---|
| M1 | revert the fix entirely — plain `FastAPI(title=…, version=…)` | **KILLED** | `TestDocRoutesAreGoneOnANonLoopbackBind::test_unauthenticated_client_cannot_read_them[/openapi.json]` |
| M2 | `_docs_enabled = not _is_loopback(host)` | **KILLED** | same |
| M3 | drop `swagger_ui_oauth2_redirect_url` | **KILLED** | `::test_no_openapi_schema_is_generated_at_all` |
| M4 | `redoc_url="/redoc"` unconditionally | **KILLED** | `::test_no_openapi_schema_is_generated_at_all` |
| M5 | `openapi_url="/openapi.json"` unconditionally | **KILLED** | `::test_unauthenticated_client_cannot_read_them[/openapi.json]` |
| M6 | `docs_url="/docs"` unconditionally | **KILLED** | `::test_no_openapi_schema_is_generated_at_all` |
| M7 | `_docs_enabled = False` — over-broad, kills loopback too | **KILLED** | `TestLoopbackKeepsTheDeveloperConvenience::test_docs_still_serve_on_loopback[/openapi.json]` |
| M8 | `_docs_enabled = True` — option 1 shape, reachable not removed | **KILLED** | `::test_unauthenticated_client_cannot_read_them[/openapi.json]` |

**8/8, no survivors.**

## What I found while mutating, and what I did about it — please read this bit

M3, M4 and M6 were killed **only** by the app-attribute assertion. The behavioural per-route tests did **not** fail for any of them. I ran each one again without `-x` to be sure that was real rather than an artefact of first-failure reporting, and it is real.

The reason is upstream, and I checked it in the installed source rather than inferring it — `fastapi/applications.py` at 0.141.1:

```
1121:  if self.openapi_url and self.docs_url:
1139:      if self.swagger_ui_oauth2_redirect_url:      # nested inside 1121
1149:  if self.openapi_url and self.redoc_url:
```

Every docs route is gated on the schema URL first. So **`openapi_url=None` alone already removes all four**, and the other three parameters change no behaviour today. They are defence-in-depth against a future edit that restores `openapi_url` — which is the likely edit, because the schema is the piece a developer actually wants back.

Rather than report that as three weakly-covered lines, I made the defence behavioural. `TestTheBeltAndBracesParametersDoRealWork` pins the upstream contract the fix depends on (`FastAPI(openapi_url=None)` must register none of the four) and then pins each remaining parameter **with the schema restored**, so each one is shown to remove its own route by itself:

```python
app = FastAPI(openapi_url="/openapi.json", **kwargs)   # kwargs = {"docs_url": None}, etc.
assert TestClient(app).get(route).status_code == 404
assert TestClient(app).get("/openapi.json").status_code == 200   # control
```

The second assertion is the control: without it a 404 could mean the schema was still absent rather than that the named parameter did the work.

## Verification

```
pytest tests/test_issue731_ui_docs_bind.py                     22 passed
pytest tests/test_issue731_ui_docs_bind.py \
       tests/test_issue687_ui_auth.py                          57 passed
pytest tests/test_issue687_ui_auth.py tests/test_ui.py \
       tests/test_ui_chat.py tests/test_ui_config_builder.py \
       tests/test_ui_live_monitor.py tests/test_ui_metrics.py   135 passed
ruff check src/soup_cli/ tests/test_issue731_ui_docs_bind.py    All checks passed!
git diff -- tests/ | grep -c "^-[^-]"                           0
```

## Docs

`docs/serving-and-export.md` gains an **Interactive API docs are loopback-only** paragraph under Web UI stating the asymmetry as a decision, why gating was rejected, and what to do if you need the schema while bound publicly. `docs/commands.md` gets the one-line note on the `--public` row.

**One correction inside that same paragraph, which is scope I am flagging rather than hiding.** The Security paragraph still read *"All mutating endpoints … require `Authorization: Bearer <token>`"*. After #707 that is wrong — reads and SSE streams require it too — and it sat directly above the sentence I was adding. I corrected it to name both kinds and to name `/` and `/api/health` as the deliberate exceptions. Revert that hunk if you would rather it went in separately.

## Scope limits

- **Nothing here changes what the schema contains**, only who can fetch it. The 24 described paths are unchanged on loopback.
- I did **not** add a `--docs` escape hatch. Nobody has asked for one, and a flag that re-opens the surface on a public bind would recreate the issue with an audit trail. If you want one, say so and I will add it as an explicit opt-in with its own test.
- I did not touch `_is_loopback`; the fix reuses the same predicate #707 already gates the token requirement on, so the two cannot disagree about what "public" means.

## Attribution

Found by **@MakazhanAlpamys** while reviewing #707 (**@kok-o**), whose route audit made these four visible as the remainder. I said in my claim that I would hand this straight to @kok-o if they considered it theirs as the follow-up to their own audit; that offer stands until this merges.


MakazhanAlpamys:
Merged as `ad3516c`. **This one I could not weaken, and I tried harder than usual because it is a security fix and those are the ones where a green suite is most likely to be measuring the wrong thing.**

## What I ran

Everything below is on your head `aeac063`, in a worktree with `PYTHONPATH` pinned inside it, driving the **real ASGI app** rather than reading the code.

```
=== LOOPBACK (127.0.0.1) ===            === PUBLIC (0.0.0.0) ===
  /openapi.json          200              /openapi.json      noauth=404  withtoken=404
  /docs                  200              /docs              noauth=404  withtoken=404
  /redoc                 200              /redoc             noauth=404  withtoken=404
  /docs/oauth2-redirect  200              /docs/oauth2-redir noauth=404  withtoken=404
=== LOOPBACK by name (localhost) ===    === REAL LAN IP (192.168.0.105) ===
  all four               200              all four           noauth=404  withtoken=404
```

I tested a real LAN interface address, not only `0.0.0.0`, and the alias forms (`/docs/`, `/openapi.json/`, `/redoc/`, `/DOCS`, `/openapi.JSON`) — all 404.

**404 with a valid Bearer token is the discriminator**, and your `test_the_route_is_absent_not_merely_refused` pins exactly that. Removed, not gated, is the right answer here: there is no route left to get the dependency wiring wrong on.

**Mutation: 5 applied, 5 killed.**

| mutation | result |
|---|---|
| revert to plain `FastAPI(title, version)` | 9 failed |
| `_docs_enabled = not _is_loopback(host)` (inverted) | 14 failed |
| drop `swagger_ui_oauth2_redirect_url` | 1 failed |
| `_docs_enabled = True` (docs always on) | 9 failed |
| `_docs_enabled = False` (over-broad, breaks local dev) | 5 failed |

Both over-broad shapes die, which is the pair most fixes of this kind miss — it is easy to write a test that only proves the public case and would happily accept a patch that kills local dev too.

`app.py` was md5-verified byte-identical after each round. Tests: 22 in `test_issue731…`, 57 with `test_issue687_ui_auth.py`, **157 across the full UI suite**. Zero deleted test lines. `ruff` clean. Changelog fragment correct: `changelog.d/0.74.0/732.security.md`, named by **PR** number, category `security`.

## Full route enumeration, since you asked for it in the issue

All 25 routes probed unauthenticated on a public bind. Exactly two answer non-401 — `GET /` and `GET /api/health` — which are the two your acceptance criteria require. All 22 `/api/*` routes answer 401, including the SSE stream.

One surface neither of us enumerated: the **`/static` mount** serves `index.html` / `app.js` / `style.css` at 200 unauthenticated. Package assets, no run data, and the dashboard cannot load without them, so it carries the same rationale as `/`. Pre-existing and not widened by this PR — recording it so nobody has to rediscover it.

## The part I want on the record

You disclosed, unprompted, that three of your mutations are killed **only** by the app-attribute assertion, because FastAPI gates every docs route behind `openapi_url` first — so a behavioural probe alone cannot distinguish them. I reproduced that:

```
M3 (drop swagger_ui_oauth2_redirect_url) -> 1 failed: test_no_openapi_schema_is_generated_at_all
M3, behavioural class only -> the /docs/oauth2-redirect test still PASSES
```

Most contributors would have shipped the behavioural class and let it look stronger than it is. You named the weakness and then **converted it into a real test with a positive control** (`TestTheBeltAndBracesParametersDoRealWork`) that pins the upstream FastAPI contract the whole design leans on. That is the difference between a suite that passes and a suite that would notice.

Your out-of-scope correction to the Security paragraph is also factually right — my enumeration shows every private *read* answers 401, so the pre-#707 wording was stale.

**One nit, nothing to do:** `test_no_openapi_schema_is_generated_at_all` is misnamed — `app.openapi()` still builds a 24-path object in-process, it is simply never routed. Your docstring says the accurate thing; only the name overstates it. Not worth a commit on its own; fold it in if you touch the file.

Thank you — you found this, reproduced it on `main` before claiming it, and fixed it with the stronger of the two available designs.

PR #733, opened 2026-09-06
fix(mlx): honour train_on_responses_only with a real per-token mask (#683)

### LINKED ISSUE #683: MLX SFT ignores train_on_responses_only

Srinivasan8888:
Closes #683.

Verified on the M1 8 GB I used for the #23 benchmark record, so the behavioural half is measured rather than pinned with a fake.

## The headline: the obvious fix is also wrong

You warned that `TrainingArgs(mask_prompt=True)` raises and the route in is an attribute set. That is true and I confirmed it, but it is the smaller half. **`args.mask_prompt = True` produces a different wrong distribution on multi-turn chat**, because `ChatDataset.process` masks a single prefix ending before `messages[-1]` (`datasets.py:65-75`).

Measured end to end on Metal — 16 conversations, two assistant turns each. `Trained Tokens` is mlx-lm's own running sum of the `ntoks` each loss call returns, so the counter *is* the supervised-token count; no instrumentation of mine is involved:

| | supervised tokens |
|---|---|
| today (`responses_only` reaching nothing) | **772** — includes every system and user turn |
| upstream `mask_prompt=True` | **71** — last assistant turn only |
| this PR | **146** |

Upstream's flag would drop **75 of 146 assistant tokens, 51%**, while looking like a fix. That fails your second acceptance criterion rather than satisfying it, and it fails it invisibly.

I predicted 146 independently from the tokenizer before trusting the run, so the number is *right* rather than merely smaller:

```
Soup per-token mask, predicted      : 146
upstream mask_prompt=True, predicted:  71
Metal run measured, flag off        : 772
Metal run measured, flag on         : 146
prediction matches the Metal run    : True
```

Adapter metadata from the same two runs:

```
                                responses_only=False   responses_only=True
adapter mask_prompt                            False                 False
adapter response_token_mask                    False                  True
adapter train_on_responses_only                False                  True
```

## Why this is an injection and not a fork

A prefix offset is all upstream can express because `default_loss` rebuilds the mask from an `(offset, length)` pair — one contiguous span per sample (`trainer.py:86-96`). Multiple assistant spans are not representable in that pair.

But `train()` takes both hooks (`trainer.py:224-225`):

```python
loss: callable = default_loss,
iterate_batches: callable = iterate_batches,
```

So `src/soup_cli/trainer/mlx_masking.py` supplies a dataset emitting `(tokens, per-token mask)`, a batching function that pads the mask with the tokens, and a loss that multiplies by it. Same `train()`, same `TrainingArgs`, same `TrainingCallback`, same throughput accounting — `masked_loss` returns the same `(mean_loss, ntoks)` pair, which is why the counter above is comparable across the two runs. `CacheDataset` wraps the new dataset unchanged.

Spans are recovered as the token delta between `messages[:k]` and `messages[:k+1]`, with `add_generation_prompt=True` on assistant turns so the `<|im_start|>assistant` header stays on the masked side.

## Three shapes, three different correct answers — two of them silent

| rows | strategy | why |
|---|---|---|
| `messages` | Soup per-token mask | upstream would supervise only the last turn |
| `prompt`/`completion` | upstream's `mask_prompt` | its single masked prefix **is** the whole prompt — upstream is right here and I did not replace it |
| `text` | neither, warn | upstream **raises** `ValueError("Prompt masking not supported for text dataset.")` (`datasets.py:195-196`) |

That third row matters: setting the flag unconditionally would convert a silently-wrong run into a crash. It is pinned by a test that calls upstream's own `create_dataset` and asserts the raise, so if mlx-lm ever supports it we find out.

This dispatch is a standalone `plan_response_masking()` rather than branches inside `train()`, so it is testable without a model load — the mutation table below is only possible because of that.

## Explicit refusals, per your last criterion

Refused rather than approximated, each with its own test:

- **a chat template that is not prefix-stable** — every prefix is checked against the full sequence, and `ResponseMaskError` is raised rather than returning a plausible-looking mask
- **a conversation with no assistant content** — contributes no loss signal, so it is a data error under this option
- **a conversation opening on an assistant turn** — its generation prompt cannot be rendered against nothing, so its header would be supervised as if the model had produced it

`train_on_messages_with_train_field`: **not supported, and not silently accepted** — it never reaches the MLX path, and plain-text rows now warn. I did not build it because I have no row shape to test it against; say the word and I will, with a test.

## Mutation table

`tests/test_issue683_mlx_response_masking.py`, baseline **28 passed**. `__pycache__` cleared before each round; both files md5-verified after (`mlx_masking.py` `797318d928ff35cf24716d8416804383`, `mlx_sft.py` `86bf88846a2469de51195f8e11e09319`).

|---|---|---|---|
| M1 | drop `add_generation_prompt` — supervise the assistant header | **KILLED** | `test_single_turn_prompt_tokens_have_zero_weight` |
| M2 | supervise only the last assistant turn (= upstream semantics) | **KILLED** | `test_multi_turn_supervises_both_assistant_turns` |
| M3 | drop the prefix-stability check | **KILLED** | `test_a_non_prefix_stable_template_is_refused` |
| M4 | `masks[:, :-1]` — unshifted mask, off by one into the prompt | **KILLED** | `test_the_mask_is_shifted_to_match_the_targets` |
| M5 | never write the mask into the batch (all-zero mask) | **KILLED** | `test_mask_and_tokens_come_back_the_same_shape` |
| M6 | allow a leading assistant turn | **KILLED** | `test_a_leading_assistant_turn_is_refused` |
| M7 | chat rows no longer get the token mask | **KILLED** | `test_chat_rows_get_soups_per_token_mask` |
| M8 | set `mask_prompt` for every non-chat shape | **KILLED** | `test_plain_text_rows_are_warned_about_not_masked` |
| M9 | mask even when the option is off | **KILLED** | `test_the_flag_off_masks_nothing_whatever_the_shape[sample0]` |
| M10 | drop the no-assistant-content refusal | **KILLED** | `test_a_conversation_with_no_assistant_turn_is_refused` |

**10/10, no survivors.**

## A bug my own tests did not catch, and what I changed because of it

The first version called `apply_chat_template([])` for the `k == 0` prefix. **Real HuggingFace tokenizers raise on that** — `"Cannot apply chat template to an empty conversation"` — and it crashed a Metal run *after* 14 unit tests had gone green, because my fake tokenizer politely returned `[]` where a real one refuses.

I fixed the code, and then fixed the fake to reproduce the refusal, because a fake more permissive than the real thing is a suite that reports green on code that cannot run. Two tests now pin the boundary directly, including one that records every call and asserts an empty conversation is never rendered. M6 is the mutation for it.

I am flagging this rather than burying it: the fake-tokenizer choice is the weakest part of this PR, and it is only defensible because the numbers above come from real hardware and a real tokenizer.

## Verification

```
pytest tests/test_issue683_mlx_response_masking.py                     28 passed
pytest <all 11 MLX suites> + test_cli_startup_is_light.py             377 passed
ruff check src/soup_cli/ tests/test_issue683_mlx_response_masking.py   All checks passed!
git diff -- tests/ | grep -c "^-[^-]"                                  0
```

`test_cli_startup_is_light.py` is in there deliberately — `mlx_masking` is imported inside `train()`, not at module top, and `mlx.core` / `mlx.nn` are imported inside the two functions that need them.

## Overlap and scope

- **@AmirF194** has #698 open on `mlx_sft.py`. My hunks are the dataset construction and the `train()` call; theirs is the `TrainingArgs(...)` field list. I have not touched that call, as I said on the issue. **I will rebase onto #698 when it lands, not the other way round.**
- #686 (warmup/scheduler/weight decay) is the third of this family and still unclaimed — I have deliberately not touched the optimizer here.
- I did **not** re-measure the published benchmark record. It uses single-turn rows with the flag at its default, so its token counts move under this change; that is a separate pass and I would rather do it against a settled `mlx_sft.py` than twice.


Srinivasan8888:
**Correction to my own PR body: #733 and #734 do not conflict, and neither needs rebasing onto the other.**

I wrote on both that they "touch adjacent lines and will need one of them rebased onto the other". I asserted that from reading the diffs instead of trying it. Trying it:

```
$ git merge --no-commit --no-ff <the other branch>
Auto-merging src/soup_cli/trainer/mlx_sft.py
Automatic merge went well
```

They land in three separate regions of `mlx_sft.py` — #733 in the dataset construction and the `train()` call, #734 in the optimizer construction — and git resolves them without help. **Merge them in either order; there is nothing for you to do.**

A clean textual merge is not the same as working code, so I tested the combination rather than stopping at the merge:

```
pytest <both new suites + 6 existing MLX suites>     180 passed
ruff check src/soup_cli/ <both new test files>       All checks passed!
```

And on Metal, with both features active in a single run — response-only masking and a cosine schedule with warmup and weight decay together:

```
  iter    learning rate    trained (supervised) tokens
     6        1.988e-04                             58
    12        1.454e-04                            113
    18        5.460e-05                            171
    24        1.231e-06                            226

adapter response_token_mask : True            <- #683 active
adapter scheduler / warmup  : cosine / 4 of 24 <- #686 active
adapter weight_decay        : 0.3
```

The learning rate moves and only assistant tokens are counted, in the same run, so the two changes are independent in effect as well as in text.

Still true and unchanged: **neither touches the `TrainingArgs(...)` call**, so @AmirF194's #698 remains free of both, and I will rebase onto it whenever it lands.

Srinivasan8888:
**Found while checking my own scope limit: this PR would have silently invalidated the published benchmark record. Fixed here, with the measurement.**

I wrote in the PR body that I had not re-measured the record and that its token counts "move under this change". I then went and checked what "move" meant, because #716 is open about exactly this class of thing and I did not want to be the next instance of it.

`benchmarks/harness/mlx_sft_smoke.py` pins `gradient_accumulation_steps` and nothing else in this area, so it inherits the schema defaults — and its rows are chatml `messages`, so the mask applies. Running the **published harness config unchanged** on my branch, on the M1 8 GB the record was measured on:

```
before (main)          Iter 48: ... Trained Tokens 2130    train 19.7s -> 108.1 tok/s
on this branch         Iter 48: ... Trained Tokens  342    train 13.1s ->  26.1 tok/s
```

`Trained Tokens` is mlx-lm's running sum of the `ntoks` each loss call returns, so under masking it counts **supervised** tokens rather than all of them. The record's `trained tokens` and `tok/s` columns therefore change **meaning as well as value**, and all four rows are affected — this is not a re-measurement, it is a different quantity under the same column heading.

**The harness now pins `train_on_responses_only: false`**, the setting the record was measured under. With the pin, the same command reports:

```
Iter 48: Train loss 0.191, Learning Rate 1.000e-04, ... Trained Tokens 2130, Peak mem 0.496 GB
```

`2130` — the published figure exactly. `tests/test_issue23_mlx_harness_bridge.py`: **198 passed**. `ruff check benchmarks/`: clean.

Re-measuring the record under masking is a worthwhile pass and I am happy to do it on this hardware, but it should be a deliberate PR with its own table, not a side effect of this one merging.

**@dchaudhari7177** — this is the same field class as #727, and it lands in the same YAML block you are editing. I checked rather than assumed: your branch and mine still auto-merge (`Auto-merging benchmarks/harness/mlx_sft_smoke.py`, no conflict), so nothing is blocked either way. If you would rather own this pin as part of your pass, say so and I will drop my hunk — your PR is the better home for "pin everything the record depends on", and I would rather it live in one place than two.

MakazhanAlpamys:
@Srinivasan8888 — **the mask math is right and I proved the off-by-one myself without Apple Silicon. Blocking, because four mutations that destroy the feature survive the entire suite on every CI runner — including the one that reinstates #683 exactly.**

Head `240ac8b` verified. Worktree with `PYTHONPATH` pinned inside it. I pulled the real `mlx_lm-0.31.3` wheel and read `tuner/trainer.py` and `tuner/datasets.py` rather than trusting your line references.

## What holds up

**The premise is real.** `ChatDataset.process` (datasets.py:57-73) returns `(tokens, offset)` with a single prefix, and `default_loss` (trainer.py:83-95) rebuilds one contiguous span from `lengths[:,0:1]` / `lengths[:,1:]`. Multiple assistant spans genuinely are not representable. `train()` takes `loss` and `iterate_batches` as callables and forwards both to `evaluate()`, so your injection point is real too.

**The mask is correct on a real ChatML template** (locally cached SmolLM2-135M-Instruct):

```
single_turn            ntok=30 sup=4    SUPERVISED: 'Four.<|im_end|>\n'
                                        boundary: last-masked='\n'  first-trained='Four'
multi_turn_2_assistant ntok=52 sup=11   SUPERVISED: 'Four.<|im_end|>\nSix is the answer.<|im_end|>\n'
ends_with_user         ntok=41 sup=4    SUPERVISED: 'Four.<|im_end|>\n'
```

Every assistant turn supervised, not just the last. Header on the masked side.

**The off-by-one is right, and I checked it without Metal.** I stubbed `mlx.core` / `mlx.nn` with about 30 lines of numpy and probed which input positions' logits actually move the loss:

```
tokens : [1,2,3,4,5,6]   mask : [0,0,0,1,1,0]   ntoks reported: 2
input positions whose logits affect the loss : [2, 3]
   -> those predict TARGET tokens at original index : [3, 4]
   expected supervised original indices             : [3, 4]
ALIGNMENT CORRECT: True
all-masked row: loss = 0.0  ntoks = 0  -> nan? False
```

Truncation and padding hold too: 40 tokens with the assistant span at 30-39, batched at `max_seq_length=16`, gives supervised count 0 rather than a mask pointing past the end, and your `width = 1 + 32*ceil(max_len/32)` is byte-identical to upstream's.

Zero deleted test lines. Ruff clean. Fragment `changelog.d/0.74.0/733.fixed.md`, PR number, credits you.

## BLOCKING 1 — four mutations survive, and one of them is #683 itself

Baseline 136 passed, 7 skipped across all MLX suites.

| mutation | result |
|---|---|
| supervise the header (`add_generation_prompt=False`) | killed |
| drop the prefix-stability check | killed |
| chat rows to `token_mask=False` in the planner | killed |
| supervise only the last assistant turn | killed |
| drop the no-assistant-content refusal | killed |
| **the mask shift `masks[:, 1:]` becomes `masks[:, :-1]`** | **SURVIVED** on CI |
| **delete the train hooks from the `train(...)` call** | **SURVIVED** |
| **`"response_token_mask": use_token_mask` becomes `False`** | **SURVIVED** |
| **`use_token_mask = plan.token_mask` becomes `= False`** | **SURVIVED** |

Your 10/10 table is honest, and it is scoped entirely to `mlx_masking.py`. It never mutates `mlx_sft.py`, which is where the feature is *connected* — and the last two survivors are the feature reaching nothing at all, which is the defect #683 reports.

**The off-by-one is worse than "not covered": it is covered by a test that runs on no CI job.** `test_the_mask_is_shifted_to_match_the_targets` is `pytest.importorskip("mlx.core")`, and the `mlx-smoke` job (`ci.yml:72-89`, macos-14) runs only `pytest -m smoke -k mlx_sft_smoke tests/test_smoke_train.py`. So that file executes nowhere. My numpy stand-in above shows the shift is testable without mlx at all.

**These are not untestable here.** I inserted a `raise` at `use_token_mask = plan.token_mask` and **23 existing tests failed** — the fake-mlx harness in `test_issue634_mlx_resume.py` already drives that code. Only the assertions are missing. And a repo-wide grep for the `mask_prompt` key in `tests/` returns nothing, so acceptance criterion 5 (metadata) has no test at all.

## BLOCKING 2 — Qwen3 multi-turn now hard-fails, and it fails mid-training

`qwen3-8b-sft-mlx` is a shipped recipe. Against the real `Qwen/Qwen3-0.6B` tokenizer:

```
single_turn            ntok=32 sup=8  SUPERVISED: '<think>\n\n</think>\n\nFour.<|im_end|>\n'
multi_turn_2_assistant REFUSED: chat template is not prefix-stable at message 2 (role='assistant')
ends_with_user         REFUSED: chat template is not prefix-stable at message 2 (role='assistant')
```

Root cause, by rendering both sides: Qwen3 injects the empty thinking block only for the **last** assistant message, so rendering `messages[:k+1]` differs from the full render at every non-final assistant turn.

**Your check catching this is correct and refusing beats mis-masking.** Two problems are about *where* and *how* it fires:

1. `build_response_mask` runs from `MaskedChatDataset.process`, which `CacheDataset` calls lazily from `masked_iterate_batches` — **inside the training loop**, after an 8B model has loaded, LoRA is applied and `Starting training...` has printed. Probe row 0 at dataset-construction time instead.
2. The message names no remedy. It should say `data.train_on_responses_only: false`.

Any thinking-style template with this shape is affected. I could only test Qwen3 — Gemma3 (`gemma3-4b-sft-mlx`) is not in my local cache and is **unverified**.

## MEDIUM — the backends now disagree, and #683's stated motivation was that they agree

Same data through `data/loss_mask.py::build_assistant_only_labels` (neither template ships a generation marker, so both take the fallback):

| case | transformers | MLX (this PR) |
|---|---|---|
| SmolLM2 single | assistant header **plus** content (8 tokens) | content only (4 tokens) |
| SmolLM2 multi | both turns **including headers** (19) | both turns **excluding headers** (11) |
| Qwen3 multi | supervises a user-turn fragment — **leaks user content** (25) | refuses |

MLX is now stricter and more correct on both axes. But #683 says the defect "invalidates direct comparisons with the Transformers backend", and this PR does not restore comparability — it makes the two backends differently masked. Temper the changelog claim to say so. The transformers Qwen3 user-content leak is pre-existing and out of scope; I will file it separately.

## MEDIUM — one PR-body claim is not accurate

The body says `train_on_messages_with_train_field` is "not supported, **and not silently accepted**". It is silently accepted:

```
ACCEPTED on backend=mlx: train_field = True
plan: MaskingPlan(token_mask=False, mask_prompt=False, warning='')
```

The mutual-exclusion validator only fires because `train_on_responses_only` defaults `true`; set it to `false` and the per-message train field is dropped with no warning. #683 criterion 4 is "supported equivalently **or rejected**", and neither happens. One line in `_check_unsupported()`.

## LOW

- `plan_response_masking` checks `messages` **before** `prompt`/`completion`; upstream `create_dataset` (datasets.py:186-193) checks prompt/completion first. A row carrying both trains on different fields depending on the flag. The docstring's "dispatches on the same row shapes upstream's `create_dataset` does" is inaccurate about precedence.
- `masked_iterate_batches` drops upstream's length sort. A no-op in practice (`CacheDataset.itemlen` returns the raw dict's key count, so upstream's sort is degenerate and stable) — but "batches into the same shapes an unmasked one would" is overstated for rows carrying tools.
- An empty dataset emits the plain-text warning ("plain-text rows carry no role boundaries") for an empty dict, which misleads.
- `benchmarks/run-m1-8gb-mlx-sft.md`'s Config row is not updated to state the new harness pin, so the record still depends on a schema default — the exact convention #727 is establishing.

## Minimum to unblock

1. Assert `train()` receives the masked loss and the masked batch iterator, and that `MaskedChatDataset` is used on the chat path. Kills the two wiring survivors; the fake-mlx harness already reaches that code.
2. Assert the three new `adapter_config.json` keys. Kills the metadata survivor and is criterion 5.
3. De-gate the shift test with a numpy stand-in, or add this file to the `mlx-smoke` job. Kills the off-by-one **on CI**, where it currently runs nowhere.
4. Probe row 0 at construction time, with a message naming `data.train_on_responses_only: false`.
5. Either warn on `train_on_messages_with_train_field` under `backend: mlx`, or correct the PR body.

## What I could not verify

No Apple Silicon and no mlx on this box. Every real-MLX number in your body — the 772 / 71 / 146 supervised-token measurements — I reproduced the *logic* for with real tokenizers and a numpy stand-in; I did not run Metal. Nor could I run an end-to-end `soup train --backend mlx`, so whether the tiny-random-Llama fixture's template is prefix-stable under this PR is unconfirmed. I judge it low risk and `mlx-smoke` will answer it in about two minutes.

The design here is right — real upstream source reading, explicit refusal over silent approximation, controls in the tests. The gap is entirely at the layer where the feature is wired in, and it is cheap to close.

MakazhanAlpamys:
@Srinivasan8888 — **the masking implementation is the best-reasoned thing in this family, and I am blocking on a one-line mutation that turns the whole fix off without a single test noticing.**

## The blocker

```
baseline                                          21 passed, 7 skipped
use_token_mask = False   (#683 restored verbatim) 21 passed, 7 skipped
```

Identical. I ran that myself on `240ac8b`. **Nothing in the suite drives `MLXSFTTrainerWrapper` at all** — zero references to the wrapper, to `train_hooks`, or to `args.mask_prompt` as *set*. So `build_response_mask` and the dispatch are covered, and the wiring that makes either of them run is not.

This is the same shape as #706's first round, where three good tests all passed with the drain thread never started, and the same shape as #698's, where the kwarg and the metadata shared one assertion. It is this repo's most frequent defect and you are one test away from it.

**#734 already solves this problem for itself** by importing `_install_fake_mlx` from `test_issue634_mlx_resume.py`. Do the same and assert:

- `train()` receives `loss=masked_loss, iterate_batches=masked_iterate_batches` for chat rows
- `args.mask_prompt is True` for prompt/completion rows
- the metadata records `response_token_mask`

M1 must die.

## The second half, which CI structurally cannot see

Two more mutations survive, and the reason is worse than the mutations:

```
masks[:, 1:] -> masks[:, :-1]     (off-by-one INTO the prompt)   SURVIVED
mask_arr[j, :keep] = 1            (batch pads the mask all-ones) SURVIVED
```

Their killing tests exist — but they are `pytest.importorskip("mlx.core")`-gated, and **7 of your 28 tests skip here and on every CI runner**, because GitHub's runners have no mlx either. So the green 14/14 does not cover `masked_loss` alignment, the nan-clamp, or mask padding.

The off-by-one is precisely the failure your own module docstring warns about. It is fakeable — the fake-mlx harness already exists in this repo, and you are already using it in #734.

## What is right, and one thing I want to defend on your behalf

**You set the flag the working way** — `args.mask_prompt = plan.mask_prompt` on the constructed object, never a kwarg. `TrainingArgs` has no such field, so a constructor kwarg would `TypeError`; confirmed against the real 0.31.3 sdist.

**The title says "a real per-token mask", which is more than `mask_prompt`, and that scope is earned rather than crept.** Upstream's `ChatDataset` masks one prefix before `messages[-1]`, so on two assistant turns it silently drops the first — and #683's own acceptance criterion requires every assistant turn to stay supervised. You stated that plan in the issue thread before building, with numbers, and the module docstring carries the justification. I would have accepted the narrow fix; the wider one is better and you argued for it in the right place.

The three-way dispatch is sound, including warning rather than raising on plain text — upstream raises `ValueError("Prompt masking not supported for text dataset.")`, confirmed in the sdist. The metadata lie (`"mask_prompt": False` hardcoded) is fixed. No `MagicMock`; a hand-written `FakeChatTokenizer`. Fragment correct. Zero test lines removed. Ruff clean.

**And the harness change is a genuinely good catch**: pinning `train_on_responses_only: false` so the published tok/s record does not silently change meaning (2,130 -> 342 supervised tokens on your box). That is the #716 lesson applied unprompted.

## Three non-blocking

- **Dispatch order diverges from upstream.** `create_dataset` checks `prompt`/`completion` **before** `messages`; `plan_response_masking` checks `messages` first. A row carrying both routes differently depending on whether the flag is on. Untested, low severity, worth a comment or a reorder.
- `masked_iterate_batches` drops upstream's `idx = sorted(..., key=len_fn)` while the docstring claims identical shapes. Near-harmless in practice — `CacheDataset.itemlen` returns `len(raw_dict)`, so the sort is already a no-op for dict rows — but the docstring is the thing to correct, not the code.
- Heterogeneous datasets (`train_rows[0]` chat, a later row text) would `KeyError` in `process`. Upstream has the identical `data[0]` fragility, so not a regression.

Nothing here is a criticism of the reasoning, which is unusually good. It is that the reasoning is currently defended only where mlx is installed, and that is nowhere the project can check.

Srinivasan8888:
Fixed in `28ed10ca`. **All four survivors killed, and killed on a runner with no MLX. The Qwen3 refusal now happens before the model loads.**

You were right about the shape of my mistake, and it is the same one you found on #734: my 10/10 table mutates `mlx_masking.py` exclusively and never `mlx_sft.py`, which is where the feature is *connected*. I tested the part I designed.

## Blocking 1 — the four

`TestTheMaskActuallyReachesTraining` records what `train()` is actually called with, through the fake-MLX harness you pointed at. No `importorskip`. Under my simulated no-mlx runner (blocks fresh `mlx*` imports, still resolves `sys.modules`):

```
baseline                                          29 passed, 7 skipped
use_token_mask = plan.token_mask -> False          3 failed
delete **train_hooks from train(...)               1 failed
"response_token_mask": use_token_mask -> False     1 failed
masks[:, 1:] -> masks[:, :-1]                      2 failed   <- was 0 on CI
```

**The off-by-one needed more than de-gating.** Your numpy stand-in works, and I took it — but the count assertion alone can coincide, which is the objection your own comment raises about mask tests. So the new test perturbs the logits at one input position at a time and records which positions the loss actually responds to:

```
the loss depends on input positions [2, 3]
an unshifted mask would make it depend on [3, 4]
```

That is the claim stated directly rather than through a count. The `importorskip` real-MLX class is untouched — this is added coverage, not a replacement.

A control I added: with the flag **off**, no hooks are passed, the dataset is not a `MaskedChatDataset`, and the metadata records `false`. Otherwise the four assertions above would hold for every run.

## Blocking 2 — Qwen3 fails before the model loads now

Row 0 is probed at dataset construction, before `CacheDataset` and before `train()`. The refusal names the remedy:

```
... chat template is not prefix-stable at message 2 (role='assistant').
Set `data.train_on_responses_only: false` to train on the full sequence on
this model, or run the recipe on the transformers backend.
```

Pinned by two tests plus a control, one of which asserts `adapter_config.json` was never written — the run does not get far enough to record metadata. Removing the probe: 2 failed.

## Medium — you are right, my PR body was wrong

`train_on_messages_with_train_field` **is** silently accepted; the mutual-exclusion validator only fires while `train_on_responses_only` is at its `true` default. `_check_unsupported()` now reports it, and `train_on_prompt` alongside it for the same reason, in the existing `MLX backend ignores:` wording. Control test asserts a plain config warns about neither.

## Medium — the changelog claim is tempered

Your table is the right framing and I have said so in the fragment: MLX excludes the assistant header and refuses what it cannot align; transformers supervises the header and leaks a user-turn fragment on Qwen3. This PR makes MLX stricter, it does not restore comparability. Thank you for filing the transformers leak separately rather than expanding this PR.

```
pytest <683 + 634 + 684 + 685 + both #23 bridges + 423 + startup>  354 passed
simulated no-mlx runner                            34 passed, 7 skipped
ruff check src/soup_cli/ tests/ benchmarks/        All checks passed!
git diff -- tests/ | grep -c "^-[^-]"              0
```

## Not done, deliberately

Your four LOW items — the `create_dataset` precedence mismatch, the dropped length sort, the empty-dataset warning wording, and the `run-m1-8gb-mlx-sft.md` Config row — are all real and none is behavioural on any shipped recipe. Say the word and they are one commit; I left them out so this diff is only the blocking set plus the two mediums you named.

## What I still cannot verify

Nothing re-run on Metal since `240ac8b`. Gemma3 (`gemma3-4b-sft-mlx`) is still unverified for prefix-stability — not in my cache either, and the probe now at least makes a failure there immediate and self-explaining rather than mid-run. My PR body's 772 / 71 / 146 measurements stand as previously reported and were not re-measured today.

Srinivasan8888:
**CI caught a real regression from my probe, on `test (ubuntu-latest, 3.10)` and `(3.11)`. Fixed in `0d7f79b4`.** Reporting it rather than letting it look like a matrix flake — and it reproduces on this Mac, so it was never platform-specific. I just had not run the file.

## What broke

`tests/test_review_fixes_v07133.py::TestMlxOptimizer::test_train_passes_non_none_optimizer` sets up:

```python
wrapper.tokenizer = object()
wrapper._dataset = {"train": [{"messages": []}], "val": []}
```

That worked only because the per-token mask was built **lazily** from inside `train()`, and that test's fake `train()` never touches the dataset. Probing row 0 at construction makes the empty conversation refuse immediately:

```
ResponseMaskError: conversation has no messages
```

Which is exactly the behaviour you asked for. The lazy path refused that row too — it just did it later, from inside the training loop, which is the timing you filed as blocking.

## Why I changed the fixture and not the probe

Making the probe lenient about a malformed row 0 would undo the fix: "unmaskable" and "empty" both have to fail before the model loads, or the guarantee is "fails early *sometimes*", which is worse than either.

So the row became real and the tokenizer became a ChatML-shaped fake. **Two test lines are replaced and neither is an assertion** — `git diff -- tests/ | grep -c "^-[^-]"` reports 2 and they are the two setup lines above. The test's own assertions (`"optimizer" in seen`, `seen["optimizer"] is sentinel`) are untouched, and it now reaches them **through the masked path a default `format: chatml` config actually takes**, instead of through a row no real dataset contains. Net coverage up, not down.

I want to be explicit about that since you check for exactly this: I edited a test file I did not write, on a PR you are reviewing, and the number is not zero. Those are the two lines.

```
pytest <all 9 files that drive MLXSFTTrainerWrapper, + test_v07200>   371 passed, 3 skipped
pytest tests/test_review_fixes_v07133.py                              28 passed
```

I also swept for the same shape elsewhere: the other `{"messages": []}` fixtures in `tests/` belong to the transformers `SFTTrainerWrapper`, `format_to_messages`, the BoN resume writer and the linter — none reach this path, and all pass.

## The process failure, since it is the more useful part

My verification on `28ed10ca` was eight targeted suites plus a no-mlx simulation. `test_review_fixes_v07133.py` is not named after an issue I touched, so it was not in my list — and it is precisely the file that drives `train()` with a stub tokenizer. A behaviour change at dataset construction can break any caller of `train()`, so the correct selection was "every file that constructs `MLXSFTTrainerWrapper`" (`grep -rln`), which is nine files, not the eight I guessed at. That is the selection I ran above and will run from now on when the change is in a shared call path.

`ubuntu-3.12`, macOS and Windows cells were still running when I pushed; I will report whatever they say.

MakazhanAlpamys:
@Srinivasan8888 — **all four survivors are dead, and dead on a runner with no MLX, which is exactly what I asked for. Eight mutations run at review, zero survivors. The code is merge-ready; I am blocking on two statements it makes to users, one of which is false and one of which is missing.**

## The mutations

Verified with `find_spec("mlx") is None` first, so "works without MLX" is measured rather than assumed. 168 passed / 7 skipped across the nine files that construct `MLXSFTTrainerWrapper`; all 7 skips are the pre-existing `importorskip` classes and neither new class carries one.

| mutation | dies by |
|---|---|
| `use_token_mask = plan.token_mask` -> `False` (**#683 restored verbatim**) | 6 tests, incl. `test_train_receives_the_masked_loss_and_batch_iterator`, `test_the_chat_path_builds_a_masked_dataset`, `test_the_adapter_records_that_the_mask_ran` |
| delete `**train_hooks,` from the `train(...)` call | `test_train_receives_the_masked_loss_and_batch_iterator`, `test_a_supported_template_is_not_refused` |
| `"response_token_mask": use_token_mask` -> `False` | `test_the_adapter_records_that_the_mask_ran`, `test_a_supported_template_is_not_refused` |
| `masks[:, 1:]` -> `masks[:, :-1]` (off-by-one into the prompt) | `test_the_mask_is_shifted_to_match_the_targets`, `test_the_supervised_positions_are_the_assistant_tokens_not_the_one_before` — **this one killed nothing on CI before** |
| construction-time probe disabled (refusal back to lazy) | `test_the_refusal_happens_before_train_is_called`, `test_the_refusal_names_the_setting_to_change` |
| prefix-stability refusal dropped (the Qwen3 guard) | 3 tests |
| `args.mask_prompt = plan.mask_prompt` -> `False` | `test_prompt_completion_rows_use_upstreams_flag_not_the_token_mask` |
| the `train_on_messages_with_train_field` report dropped | `test_the_per_message_train_field_is_reported_as_ignored` |

Two of those discriminate more than they look like they do, and that is the part worth naming. The probe mutation pins **timing**, not just the message: under the fake harness `train()` never consumes the dataset, so with the lazy path no error is raised at all and `pytest.raises` fails. And the off-by-one killer asserts *which input positions the loss responds to* rather than a token count, so it cannot pass by coincidence.

Two removed test lines, both audited, both the setup lines you declared. `test_train_passes_non_none_optimizer`'s own assertions are untouched and now run through the masked path.

## Blocking 1 — the changelog states something that is not true

The fragment ends:

> ...is now refused at dataset construction, **before the model loads**, with a message naming `data.train_on_responses_only: false` as the remedy.

It does not happen before the model loads. From your own head:

```
mlx_sft.py:182  def setup(...)
mlx_sft.py:191      self.model, self.tokenizer = load_mlx_model(...)   <- model loaded
mlx_sft.py:276  def train(...)
mlx_sft.py:303      self._apply_lora(self.model)                       <- LoRA applied
mlx_sft.py:376      use_token_mask = plan.token_mask                   <- the probe
```

The refusal is after the model load **and** after LoRA. What you actually delivered is what I asked for — probe row 0 at dataset-construction time, before `mlx_lm`'s `train()` and before the training loop — and that is a real improvement over the lazy version. The summary just overstates it, and it overstates it in the one place a user reads.

The same phrasing is in three places: this PR comment, the changelog fragment, and the test class name `TestAnUnsupportedTemplateFailsBeforeTheModelLoads`. **"before the training loop starts"** is accurate and loses nothing.

Worth saying plainly: your own harness sets `w.model` directly, so nothing in the suite can observe load ordering. The test name asserts a property the test cannot see — which is the same shape as the finding you reported against yourself on #713, and you will spot it faster than I did next time.

## Blocking 2 — a shipped recipe changes behaviour and nothing says so

`data.train_on_responses_only` defaults to **`True`** (`config/schema.py:383`). The shipped `qwen3-8b-sft-mlx` recipe is `backend: mlx`, `format: chatml`, and **does not set it**:

```yaml
base: mlx-community/Qwen3-8B-4bit
task: sft
backend: mlx
data:
  train: ./data/train.jsonl
  format: chatml
```

By your changelog's own account, Qwen3's template is the one that gets refused. So on this PR, `soup recipes use qwen3-8b-sft-mlx` followed by `soup train` appears to go from *training* (badly, on the prompt) to *refusing*.

I could not run it — no Apple Silicon, and the probe needs the real tokenizer — so I am asking rather than asserting. You have the M1:

1. Does the shipped recipe still train as-is, or does it now refuse?
2. If it refuses, the recipe should carry `train_on_responses_only: false` with a comment, or the changelog must name it. Silently breaking a shipped recipe is worse than the bug being fixed.

Non-blocking but the same family: `docs/performance-and-quantization.md:620` and `docs/training.md:125` describe `train_on_responses_only` and the MLX "ignores" list generically. Neither says MLX now excludes the assistant header where transformers includes it, refuses templates it cannot align, or reports `train_on_messages_with_train_field` as ignored. A short paragraph would cover all three.

## Not asks, for the record

Your `create_dataset` precedence note, the dropped length sort, the empty-dataset wording and the `benchmarks/run-m1-8gb-mlx-sft.md` Config row are all still open from my LOW list and all deliberate. Gemma3 prefix-stability remains unverified anywhere — the eager probe at least makes such a failure immediate and self-explaining rather than a silent mis-mask.

Fix the phrase in the three places, answer the recipe question, and I merge on green.

Srinivasan8888:
Fixed in `d0e6fe50`. **Both blockers, and the recipe question answered with measurements from the M1 rather than an opinion.**

## Blocking 1 — the claim was false, corrected in all three places

You are right and the line numbers are right: `setup()` loads the model at `mlx_sft.py:191`, `_apply_lora` runs at `:303`, and the probe is at `:376`. The refusal is after both. **"before the training loop starts"** now, in the changelog fragment, the test class name (`TestAnUnsupportedTemplateFailsBeforeTheTrainingLoop`), and this thread.

The sharper half of your point is the one I want on the record: **the class name asserted a property the test cannot observe.** The harness sets `wrapper.model` directly, so load ordering is invisible to it — the name was making a claim the suite had no way to check, which is the same finding I reported against myself on #713 one day earlier. I have put that reasoning in the class docstring so the next reader gets it without re-deriving it. (The single removed test line `git diff` reports is this rename.)

## Blocking 2 — one of three shipped recipes is affected, and it is a data-shape question

Measured against the real `Qwen/Qwen3-0.6B` tokenizer on this machine:

```
single_turn   OK       ntok=23 sup=8   SUPERVISED: `<think>\n\n</think>\n\nFour.<|im_end|>\n`
with_system   OK       ntok=31 sup=8   same
multi_turn    REFUSED  chat template is not prefix-stable at message 1 (role=`assistant`)
```

So `qwen3-8b-sft-mlx` **still trains on single-turn data** — including the thinking block, correctly supervised — and **refuses multi-turn**, where it previously trained on the prompt. It is not the recipe that breaks, it is multi-turn Qwen3 data.

**And your unverified item is now verified — Gemma3 is fine.** I pulled the exact repo the recipe names rather than a proxy:

```
mlx-community/gemma-3-4b-it-4bit   single sup=4   multi sup=8   `Four.<end_of_turn>\nSix.<end_of_turn>\n`
unsloth/Meta-Llama-3.1-8B-Instruct single sup=3   multi sup=6   `Four.<|eot_id|>Six.<|eot_id|>`
```

Both mask every assistant turn on both shapes. **One of three MLX recipes is affected, not all three** — which is a materially smaller blast radius than either of us was assuming.

## Which of your two options I took, and why

You offered: pin `train_on_responses_only: false` in the recipe, **or** name it in the changelog. I took the changelog.

Pinning `false` would disable a correct fix for every single-turn Qwen3 user — the majority shape, and the one where masking now works properly including the thinking block — in order to avoid a refusal that tells multi-turn users the exact key to change. That trade seems wrong to me. The fragment now names the recipe, the shape, the measured before/after and the remedy.

If you would rather have the pin anyway, say so and it is one line; you own that call, not me.

## Docs

`docs/performance-and-quantization.md` (the one place `train_on_responses_only` is actually documented — `docs/training.md:125` does not mention it, so your reference may have shifted) now states the backend divergence, the refusal and its timing, the Qwen3 multi-turn case with Llama/Gemma named as unaffected, and `train_on_messages_with_train_field` / `train_on_prompt` being reported rather than dropped.

```
pytest tests/test_issue683_mlx_response_masking.py + changelog guard   58 passed
ruff check src/soup_cli/ tests/                                        All checks passed!
git diff -- tests/ | grep -c "^-[^-]"                                  1  (the class rename)
```

Your four LOW items remain deliberately out, as before.

MakazhanAlpamys:
@Srinivasan8888 — merged as `f71aef0`. **Both blockers are genuinely closed, and I re-ran the one that mattered rather than reading your reply and believing it.**

## The mutation, run again at merge

The blocker across two of my three comments was that `use_token_mask = False` — #683 restored verbatim — left the suite at an identical `21 passed, 7 skipped`. On `d0e6fe50`, with `find_spec("mlx") is None` confirmed first so this is the CI condition and not a local accident:

```
6 failed, 56 passed, 7 skipped

FAILED TestTheMaskActuallyReachesTraining::test_train_receives_the_masked_loss_and_batch_iterator
FAILED TestTheMaskActuallyReachesTraining::test_the_chat_path_builds_a_masked_dataset
FAILED TestTheMaskActuallyReachesTraining::test_the_adapter_records_that_the_mask_ran
FAILED TestAnUnsupportedTemplateFailsBeforeTheTrainingLoop::test_the_refusal_happens_before_train_is_called
FAILED TestAnUnsupportedTemplateFailsBeforeTheTrainingLoop::test_the_refusal_names_the_setting_to_change
FAILED TestAnUnsupportedTemplateFailsBeforeTheTrainingLoop::test_a_supported_template_is_not_refused
```

Those are the right six: the two that assert the mask reaches `train()` and the dataset, the one that reads a real `adapter_config.json`, and all three refusal tests.

**A disclosure about my own harness, because it is the exact trap this thread has been about.** My first attempt at that mutation edited the file in text mode. The working copy is CRLF, the pattern did not match, the write was a no-op — and the suite came back `62 passed, 7 skipped`, which reads exactly like a survivor. I only caught it because I had asserted on the occurrence count. Re-applied in binary mode with SHA-256 checked before and after restore (`ec95db01` -> `fac7354b` -> `ec95db01`), it kills six. If I had trusted the first run I would have posted a false blocker at you on round four.

## Blocking 1 — verified as absent, not as claimed fixed

`grep -rn "before the model load" src/ tests/ docs/ changelog.d/` returns exactly one hit, `utils/steering.py:415`, which is pre-existing and unrelated. Nothing from this PR says it. The class rename to `TestAnUnsupportedTemplateFailsBeforeTheTrainingLoop` carries a docstring recording *why* the old name was wrong — that it asserted a property the harness cannot observe — which is more useful than a silent correction.

I am not asking you to edit the 09:32 and 10:00 comments that still contain the old phrase. Retracting forward and leaving the record intact is the right call; an edited history is worse than a corrected one.

## Blocking 2 — the recipe, read rather than taken on trust

`qwen3-8b-sft-mlx` at `catalog.py:1241` is `backend: mlx`, `format: chatml`, and its YAML body sets no `data.train_on_responses_only`, so it takes the `true` default. I read the recipe rather than the fragment. The fragment then states the shape of the change — single-turn and system+single-turn train, multi-turn now refuses — names `data.train_on_responses_only: false` as the remedy, and clears the other two MLX recipes by name. That is the option I offered and it is the better of the two, because it puts the consequence in front of the person whose run changes.

## What I did not verify, stated plainly

Every Apple Silicon measurement is yours and stays yours: 772 / 71 / 146 supervised tokens, and the Qwen3-0.6B, Gemma-3-4b and Llama-3.1 tokenizer probes. This box has no Metal and I pulled no models. I merged on the mechanism being pinned by tests that can fail on a runner with no MLX, not on having reproduced your hardware.

Two removed test lines, both setup, replaced by a ChatML-shaped fake tokenizer and a real one-turn conversation — no assertion deleted. Ruff clean. 14 of 14 checks SUCCESS on `d0e6fe50`, `head_sha` confirmed equal to the head I merged.

## The thing worth naming

Three rounds, and the correction that mattered was not a code change: you had been mutating `mlx_masking.py`, the part you designed, and never `mlx_sft.py`, the part where the feature is *connected*. You said that back in your own words before fixing it. That is the same diagnosis I had to make twice on #734, and you generalised it across both PRs without being asked to a second time.

The four LOW items from my earlier review stay open and stay deliberate.

Srinivasan8888:
Thank you — and the CRLF disclosure is the most useful thing in this thread. I have adopted it: the mutation I just pushed on #734 was applied in binary mode with the file's sha256 checked before and after restore. A no-op edit that reads as a survivor would have had me arguing a false blocker back at you, and I would not have caught it either.

Three things from your merge note I want to answer rather than let stand as compliments.

**The diagnosis was yours, not mine.** I said it back in my own words, but you are the one who located it — twice, on two PRs, before I saw the pattern. What I did was apply it a second time without being asked. Worth distinguishing.

**One correction to my own verification method, since you merged partly on my numbers.** The no-mlx "simulation" I quoted throughout this PR over-blocks. On current `main` it reports `TestVllmTrustRemoteCode::test_serve_vllm_passes_resolved_trust` failing; a genuine no-mlx venv passes all 28 in that file. So the tool that produced my counts could have invented a failure. I have rebuilt the environment properly — a venv with `[dev]` and no `mlx` extra, which is what every CI `test` cell actually is on all three OSes — and re-ran this PR's suites on it: **41 passed, 7 skipped**, unchanged. The numbers hold; the instrument was worse than I represented it as, and you should know that.

**The Apple Silicon measurements stay mine, correctly.** 772 / 71 / 146, and the Qwen3-0.6B, Gemma-3-4b and Llama-3.1 probes. You merged on the mechanism being pinned by tests that can fail without MLX, which is the right basis — those figures are evidence for a claim, not a CI guarantee, and nothing in the repo will catch them going stale.

On the four LOW items: still open, still deliberate, and I am happy to take them as a follow-up whenever you want them — say so and they are one PR rather than four.

Issue #634, opened 2026-09-01
[Bug]: MLX backend ignores saved adapter checkpoints on --resume and restarts training from scratch


imahsanali:
### Problem

When running SFT fine-tuning with backend: mlx on Apple Silicon, training cannot be resumed after an interruption (or Metal crash) using soup train --config soup.yaml --resume auto or --resume <output_dir>.

Document attached of fix i applied:

[MLX_RESUME_FIX_EXPLANATION.md](https://github.com/user-attachments/files/31698235/MLX_RESUME_FIX_EXPLANATION.md)

### Proposed solution

_resolve_checkpoint() in train.py only scans for HuggingFace/PyTorch directories (checkpoint-*), failing to detect MLX *_adapters.safetensors files.
MLXSFTTrainerWrapper.train() in mlx_sft.py hardcodes a fallback ("MLX backend does not support --resume yet; starting training from scratch.") and completely ignores resume_from_checkpoint.
The dataset iterator in mlx_lm does not offset/skip already processed batches to the saved iteration index (start_step).
Steps to Reproduce:

Start MLX training: soup train --config soup.yaml (saves checkpoints e.g., 0011800_adapters.safetensors).
Stop or let training crash/interrupt after step 100+.
Run soup train --config soup.yaml --resume auto (or --resume output_dir).
Result: Soup either reports No checkpoint found to resume from or restarts from iteration 1 instead of resuming from the saved checkpoint.
Expected Behavior: soup train --resume auto should detect the latest saved *_adapters.safetensors, load the adapter weights via model.load_weights(), skip already processed training batches, and continue training seamlessly from step N + 1.

Document for fix is attached below

### Area

Training (SFT/DPO/GRPO)

MakazhanAlpamys:
Thank you — this is a good report, and both mechanisms check out against the source. Confirmed on `main` at `74ac1e3`:

- `src/soup_cli/commands/train.py:2007` — `_resolve_checkpoint("auto", ...)` filters to `d.is_dir() and d.name.startswith("checkpoint-")`. An MLX run writes `NNNNNNN_adapters.safetensors` **files**, so the list is empty and `--resume auto` returns `None`. That is your "No checkpoint found to resume from".
- `src/soup_cli/trainer/mlx_sft.py:181-185` — `train()` prints `MLX backend does not support --resume yet; starting training from scratch.` and then `del`s nothing, uses nothing: `resume_from_checkpoint` is never read again.

[...]

PR #734, opened 2026-09-06
fix(mlx): wire warmup, scheduler, weight decay and optimizer choice (#686)

### LINKED ISSUE #686: MLX SFT ignores warmup, scheduler, and configured weight decay

Srinivasan8888:
Closes #686. Measured on the M1 8 GB from the #23 record.

## The measurement that fixed the design

Your acceptance list asks for "a clearly defined optimizer-step unit". That is the whole PR, so I measured it rather than picking one:

```python
opt = optim.AdamW(learning_rate=sched)   # sched records the step it is called with
for _ in range(4): opt.update(model, grads)
-> schedule called with steps: [0, 0, 1, 2, 3]
-> opt.step after 4 updates  : 4
```

`opt.step` counts `optimizer.update()` calls, and mlx-lm invokes that only when `it % grad_accumulation_steps == 0`. **The schedule's unit is optimizer updates.** A schedule built against `iters` stretches the warmup by `grad_accumulation_steps` and never reaches the cosine floor — while drawing a curve that looks entirely reasonable. So `total_updates = iters // grad_accumulation_steps`, and that expression has its own tests (below), because it is the one line the rest of this depends on.

## Live proof on Metal — mlx-lm's own Learning Rate column

Two real runs, 40 rows, `lr=2e-4`, identical except for the schedule:

```
  iter   scheduler=constant, warmup=0     cosine, warmup_ratio=0.2
     4                      2.000e-04                    7.500e-05
     8                      2.000e-04                    1.750e-04
    12                      2.000e-04                    1.957e-04
    16                      2.000e-04                    1.773e-04
    20                      2.000e-04                    1.471e-04
    24                      2.000e-04                    1.098e-04
    28                      2.000e-04                    7.097e-05
    32                      2.000e-04                    3.656e-05
    36                      2.000e-04                    1.181e-05
    40                      2.000e-04                    4.815e-07

constant run: 1 distinct LR value    <- today's behaviour, for every config
cosine   run: 10 distinct LR values
```

The numbers are exactly right, not merely plausible — `warmup_updates=8`, so at iteration 4 (three updates completed) the ramp is at `3/8 x 2e-4 = 7.500e-05`, and at iteration 8 it is `7/8 x 2e-4 = 1.750e-04`. Both match to the printed digit.

Adapter metadata from the same two runs:

```
  optimizer        constant=AdamW      cosine=AdamW
  scheduler        constant=constant   cosine=cosine
  warmup_updates   constant=0          cosine=8
  total_updates    constant=40         cosine=40
  weight_decay     constant=0.0        cosine=0.3
  peak_lr          constant=0.0002     cosine=0.0002
```

Before this, that block did not exist and `weight_decay` was whatever MLX defaults to (0.01), whatever the config said.

## Scale of the silent substitution

**Only 8 of the 32 names on Soup's optimizer allowlist have an MLX equivalent.** All 32 validated cleanly and then ran as AdamW. They are now refused by name, with a message that says what MLX does support and what the old behaviour was — a user upgrading needs to know why their run suddenly stops.

`weight_decay` on an optimizer whose MLX constructor cannot take it is refused rather than dropped. The table of which those are is hardcoded (so refusals work on a machine without MLX) and pinned by a test against the real constructor signatures — **which caught me having Adafactor in the wrong set.**

`constant` with no warmup still returns a plain float, so the configuration that wanted the old behaviour builds exactly what it built before, and the change is provably confined to configurations that asked for something else.

## Mutation table

`tests/test_issue686_mlx_optimizer_schedule.py`, baseline **36 passed**. `__pycache__` cleared each round; both files md5-verified after.

|---|---|---|---|
| M1 | warmup computed in **iteration** units (`x accum`) | **KILLED** | `test_warmup_starts_at_zero_and_peaks_at_the_boundary` |
| M2 | warmup dropped entirely | **KILLED** | same |
| M3 | `weight_decay` never passed | **KILLED** | `test_the_configured_decay_reaches_the_optimizer` |
| M4 | cosine silently becomes linear | **KILLED** | `test_linear_decays_to_zero_too_but_differently` |
| M5 | warmup schedule built but never joined | **KILLED** | `test_warmup_starts_at_zero_and_peaks_at_the_boundary` |
| M6 | unsupported optimizer silently becomes AdamW | **KILLED** | `test_an_optimizer_with_no_mlx_equivalent_is_refused_by_name` |
| M7 | unsupported scheduler accepted silently | **KILLED** | `test_an_unsupported_scheduler_is_refused` |
| M8 | drop the warmup-rounds-to-zero warning | **KILLED** | `test_a_warmup_that_rounds_to_zero_warns` |
| M9 | accept `weight_decay` where it cannot apply | **KILLED** | `test_decay_on_an_optimizer_that_cannot_take_it_is_refused` |
| M10 | the weight-decay table drifts from real signatures | **KILLED** | `test_the_weight_decay_table_matches_the_real_signatures` |
| M11 | **wiring passes iterations as `total_updates`** | **KILLED** | `test_total_updates_is_iters_divided_by_accumulation` |

**11/11.**

### M11 survived the first pass, and that is the part worth reading

Every unit test above passed while `mlx_sft.py` handed `plan_optimizer` `iters` instead of `iters // grad_accumulation_steps`. The arithmetic was tested; **the call site was not** — the same shape as the `val_loss`-never-reaching-the-wire finding on #713, and I would rather report it than have it found.

The fix is `TestTheWrapperPassesUpdatesNotIterations`, which drives the real `MLXSFTTrainerWrapper.train()` through the shared fake-MLX harness and reads `total_updates` back out of the written `adapter_config.json`: 48 rows at `accum=4` must record **12**, with warmup **3**. There is a companion control at `accum=1` where iterations and updates coincide — so if the divisor is wrong, exactly one of the two fails and localises it.

## Three existing suites needed their fake extended, and I want to be explicit about that

`test_issue634_mlx_resume.py` and `test_issue684_mlx_grad_accumulation.py` define fake `mlx` modules whose `mlx.optimizers` had only `AdamW` (`test_issue23_mlx_display_bridge.py` imports the first). The MLX path now reaches `cosine_decay` **on the default config**, so those fakes raised `AttributeError` — **211 failures on my first full regression run.**

I added the three schedule builders to both fakes as recording stubs. **Additions only; no assertion was touched, no test weakened:**

```
git diff -- tests/ | grep -c "^-[^-]"                     0
tests/test_issue634_mlx_resume.py             10 ++++++++++
tests/test_issue684_mlx_grad_accumulation.py  10 ++++++++++
```

The real curve is asserted against the real library in my suite, not against those stubs. My own file imports `_install_fake_mlx` from #634 rather than keeping a third copy, matching what the display-bridge suite already does.

## Verification

```
pytest tests/test_issue686_mlx_optimizer_schedule.py                   36 passed
pytest <all 11 MLX suites> + test_cli_startup_is_light.py             385 passed
ruff check src/soup_cli/ tests/test_issue686_mlx_optimizer_schedule.py  All checks passed!
git diff -- tests/ | grep -c "^-[^-]"                                   0
```

Both source files are byte-identical to what the mutation pass and the Metal run measured (`mlx_optim.py` `728e0779a52604f026b59e718ad2c0fb`, `mlx_sft.py` `7201fa8a72e7df0634018ec7ba7bf39e`) — I re-ran the hardware evidence after the last refactor rather than carrying the earlier numbers forward.

## Your open question, answered by measurement

You left the degenerate-warmup case to me. **A warmup that rounds to zero warns.** `warmup_ratio=0.03` over 12 updates is 0 steps, and a run that silently starts at peak LR when warmup was configured is the shape this issue is about. A `warmup_ratio` of 0 warns about nothing — the warning has to discriminate, and there is a control test for that.

One thing I found while writing it: `warmup >= total` is **unreachable from any valid config**, because `schema.warmup_ratio` is `le=0.5`. I kept the guard because `plan_optimizer` takes plain floats rather than a validated model, and I test it by calling the function directly with ratio 1.0 — with a second test pinning the schema cap, so the "unreachable" claim is checked rather than asserted. Say the word and I will drop the branch instead.

## Overlap

`src/soup_cli/trainer/mlx_sft.py` has three of us near it: **@AmirF194** on #698 (`TrainingArgs`), me on #733 (#683, dataset + `train()` call), and this (optimizer construction). Three distinct regions. **I have not touched the `TrainingArgs(...)` call, and I will rebase onto #698 when it lands.** #733 and this touch adjacent lines and will need one of them rebased onto the other — merge them in whichever order suits you and I will do the rebase.

@AmirF194 — my claim on #686 offered you this one as the continuation of #684/#685 and I held off pushing for a while; the offer still stands, and I will close this without argument if you would rather take it.


Srinivasan8888:
**Correction to my own PR body: #733 and #734 do not conflict, and neither needs rebasing onto the other.**

I wrote on both that they "touch adjacent lines and will need one of them rebased onto the other". I asserted that from reading the diffs instead of trying it. Trying it:

```
$ git merge --no-commit --no-ff <the other branch>
Auto-merging src/soup_cli/trainer/mlx_sft.py
Automatic merge went well
```

They land in three separate regions of `mlx_sft.py` — #733 in the dataset construction and the `train()` call, #734 in the optimizer construction — and git resolves them without help. **Merge them in either order; there is nothing for you to do.**

A clean textual merge is not the same as working code, so I tested the combination rather than stopping at the merge:

```
pytest <both new suites + 6 existing MLX suites>     180 passed
ruff check src/soup_cli/ <both new test files>       All checks passed!
```

And on Metal, with both features active in a single run — response-only masking and a cosine schedule with warmup and weight decay together:

```
  iter    learning rate    trained (supervised) tokens
     6        1.988e-04                             58
    12        1.454e-04                            113
    18        5.460e-05                            171
    24        1.231e-06                            226

adapter response_token_mask : True            <- #683 active
adapter scheduler / warmup  : cosine / 4 of 24 <- #686 active
adapter weight_decay        : 0.3
```

The learning rate moves and only assistant tokens are counted, in the same run, so the two changes are independent in effect as well as in text.

Still true and unchanged: **neither touches the `TrainingArgs(...)` call**, so @AmirF194's #698 remains free of both, and I will rebase onto it whenever it lands.

Srinivasan8888:
**Same check as on #733, same class of problem, fixed here too.**

Having found that #733 would silently move the published record, I ran the same check on this PR rather than assuming it was clean.

`benchmarks/harness/mlx_sft_smoke.py` pins `gradient_accumulation_steps` and nothing else in this area, so once this merges it inherits `scheduler: cosine` and `warmup_ratio: 0.03` from the schema. The published record documents a constant `1.000e-04` throughout; it would become a warmup-then-cosine curve, moving the `loss` column while nothing in the table says a schedule is involved.

The harness now pins `scheduler: constant` and `warmup_ratio: 0.0` — the settings the record was measured under. Verified on the M1 8 GB with the published command unchanged:

```
Iter  5: Train loss 3.819, Learning Rate 1.000e-04, ... Trained Tokens  210
Iter 48: Train loss 0.098, Learning Rate 1.000e-04, ... Trained Tokens 2130

distinct learning rates across all 48 iterations: 1   (1.000e-04)
```

`tests/test_issue23_mlx_harness_bridge.py`: **198 passed**. `ruff check benchmarks/`: clean.

Worth saying plainly: **the pin is not a workaround, it is the point of #716.** Between them, #683 and #686 were about to change a published record through two schema defaults the harness never mentioned — which is the failure @MakazhanAlpamys filed #716 to describe, and the one @dchaudhari7177's #727 is closing for a third field. I would rather demonstrate that on my own PRs than have it demonstrated on me.

@dchaudhari7177 — as on #733: your branch and mine still auto-merge (checked, not assumed), and if you would rather own these pins inside #727 I will drop my hunks. One place beats two.

Srinivasan8888:
**Windows CI caught a real bug in this PR — pushed a fix in `b04288cb`.** Reporting it here rather than letting it look like a flake.

`test (windows-latest, 3.12)` failed while every other job on my five open PRs passed, so it was mine.

## The bug

`build_lr_schedule()` imported `mlx.optimizers` at the top of the function. But its first branch — `scheduler: constant` with no warmup — returns a **plain float** and needs nothing from MLX. So a genuinely platform-independent behaviour raised `ModuleNotFoundError` on every runner without MLX installed, which is all of them except Apple Silicon.

That branch is the one I described in the PR body as *"still returns a plain float, so the configuration that asked for the old behaviour constructs exactly what it did before"* — and on Linux and Windows it constructed an exception instead. My own hardware could not see it: MLX is installed here, so the import always succeeded.

The import now sits **after** the constant-path early return.

## How I found it, since the log was not available

The run was still queued, so `gh run view --log-failed` had nothing to give me. Rather than guess, I simulated the runner: block `import mlx*` with `ModuleNotFoundError`, but **still resolve anything already in `sys.modules`**, because that is what the real import system does and it is how the fake-MLX harness works on those runners.

The first simulation I wrote was wrong in exactly that respect — it over-blocked and reported **12** failures, including the whole fake-harness wiring class. Fixing the simulation to respect `sys.modules` cut it to **one**, which is the real one. I mention it because a 12-failure report would have had me "fixing" four tests that were never broken.

```
no MLX, before the fix : 1 failed, 26 passed,  9 skipped
no MLX, after the fix  :           27 passed,  9 skipped
with MLX (this M1)     :           36 passed
regression across 6 MLX suites + test_cli_startup_is_light : 146 passed
```

## I ran the same check on #733, before it gets there

#733 has not reached the Windows jobs yet. Under the same simulated no-MLX runner its suite is **21 passed, 7 skipped, 0 failed** — the fake-tokenizer design means its behavioural tests never needed MLX. Nothing to fix there.

## The thing worth your judgement

**9 of the 36 tests here skip without MLX**, and they are the learning-rate-curve ones — the observed LR at the first update, the warmup boundary and the final update, which is the criterion you actually asked for. They cannot be made platform-independent without reimplementing `cosine_decay` in the test, which would assert my arithmetic rather than MLX's.

So on Linux and Windows this PR is covered by the refusals, the warnings, the plan arithmetic and the wiring tests, and the curve itself is covered by the Apple Silicon jobs plus the Metal run in the PR body. If you would rather the curve had a platform-independent guard as well, say so and I will add one against fixed expected values with a comment saying it duplicates MLX's arithmetic deliberately.

MakazhanAlpamys:
@Srinivasan8888 — **all four fields are genuinely wired and I verified the schedule arithmetic without Apple Silicon. Blocking on the same shape of gap as #733: the plan layer is tested exhaustively, the construction layer is not, and a mutation that reinstates #686 verbatim survives everything.**

Head `7970d91` verified. Worktree with `PYTHONPATH` pinned inside it. I re-implemented `linear_schedule` / `cosine_decay` / `join_schedules` from mlx's own semantics in numpy and recorded what actually reaches the constructor.

## The four fields, each checked on its own

```
=== warmup_ratio ===
 0.0 -> warmup_updates=0   lr[0]=1.000e-04  lr[warm]=1.000e-04  lr[-1]=6.156e-07
 0.2 -> warmup_updates=4   lr[0]=0.000e+00  lr[warm]=1.000e-04  lr[-1]=9.607e-07
 0.5 -> warmup_updates=10  lr[0]=0.000e+00  lr[warm]=1.000e-04  lr[-1]=2.447e-06
=== scheduler (warmup fixed at 4) ===
 cosine   -> ['0.00e+00','1.00e-04','8.54e-05','5.00e-05','1.46e-05']
 linear   -> ['0.00e+00','1.00e-04','7.50e-05','5.00e-05','2.50e-05']
 constant -> ['0.00e+00','1.00e-04','1.00e-04','1.00e-04','1.00e-04']
 unsupported refused: training.scheduler='polynomial' is not available...
=== weight_decay ===
 0.0 -> ('AdamW', {'learning_rate': 0.0001, 'weight_decay': 0.0})
 0.3 -> ('AdamW', {'learning_rate': 0.0001, 'weight_decay': 0.3})
=== optimizer ===
 adamw_torch -> AdamW ['learning_rate','weight_decay']
 sgd -> SGD | lion -> Lion | adam -> Adam ['learning_rate'] (no wd, correctly)
 adamw_bnb_8bit -> REFUSED: has no MLX equivalent...
```

The decay phase is measured from the warmup boundary and `decay_steps = total - warmup` is right. **The optimizer-update unit is also right**, which is the part I most expected to be wrong: mlx-lm calls `optimizer.update()` only when `it % grad_accum_steps == 0` (trainer.py:317-321), so `iters // grad_accum` is the correct denominator.

**Your `b04288cb` Windows fix is correct, verified on this no-mlx box:**

```
constant / no-warmup            -> 0.0001  float   mlx imported after call: False
constant_with_warmup / no-warmup -> 0.0005          mlx imported: False
```

The early return really does happen before `import mlx.optimizers`, and `mlx` never enters `sys.modules`. Reporting that as a real bug rather than letting it read as a flake was the right call.

Zero test lines weakened — the one removed line widens a fake (`opt_mod.AdamW = lambda learning_rate=None:` becomes `lambda **kwargs:`) while the surrounding assertion is unchanged, and the same three stubs are added to `test_issue634/684/685`. Necessary and honestly commented. Ruff clean. Fragment `changelog.d/0.74.0/734.fixed.md`, PR number, credits you.

## BLOCKING — three mutations survive, one of them is #686 itself

Baseline 108 passed, 9 skipped.

| mutation | result |
|---|---|
| `warmup = int(ratio*total)` becomes `0` | killed (3 named + 5) |
| scheduler forced to `constant` in the plan | killed (metadata test only) |
| `total_updates = iters` instead of `iters // grad_accum` | killed |
| drop the metadata spread | killed (3 named) |
| **`kwargs["weight_decay"] = plan.weight_decay` becomes `pass`** | **SURVIVED** |
| **`cls = getattr(optim, plan.optimizer_name)` becomes always `AdamW`** | **SURVIVED** |
| **replace `build_optimizer(plan)` with `optim.AdamW(learning_rate=lr)` — #686 verbatim** | **SURVIVED** |

The weight-decay one is killed only by `test_the_configured_decay_reaches_the_optimizer`, which is `importorskip("mlx.optimizers")` and therefore runs on no CI job. The other two are killed by nothing anywhere.

The reason is structural: your tests assert the **plan** and the **metadata** exhaustively, and `as_metadata()` is derived from the same plan — so no metadata assertion can ever catch a broken `build_optimizer`. The last row is the point: someone reverts to a bare `AdamW(learning_rate=lr)` and the suite is green, which is exactly the state #686 describes.

**You already applied the right instinct one layer up.** `test_issue686_mlx_optimizer_schedule.py:336` says *"Mutating `iters // grad_accumulation_steps` to `iters` in mlx_sft.py survived everything above"* and you added a wiring test for it. Do the same here — the fake-mlx harness this PR already extends in `test_review_fixes_v07133.py` makes it a few lines: assert the recorded call is `('AdamW', {'learning_rate': ..., 'weight_decay': 0.3})`.

## The benchmark record: no, this does not stale it — and you pre-empted it

You added `scheduler: constant` / `warmup_ratio: 0.0` to `benchmarks/harness/mlx_sft_smoke.py`, so the published `run-m1-8gb-mlx-sft.md` figures stay reproducible from the harness. I also checked whether #727 would have caught this class and it would **not** — its guard checks `steps == rows*epochs` and pins step-count fields only. The two sets of pins are complementary, not redundant. Merged locally: `Auto-merging benchmarks/harness/mlx_sft_smoke.py` / `Automatic merge went well`, no three-way conflict.

Two caveats:

- **`weight_decay` is not pinned in the harness.** It is safe today only because the schema default (0.01) coincides with mlx's `AdamW` default (0.01). If either moves, the record changes silently. Pin it.
- The record's Config row (`run-m1-8gb-mlx-sft.md:54`) still lists only `batch_size / max_length / lr`. Anyone reproducing from the `.md` rather than the harness gets a different curve now.

**One harness comment to reword.** It says the change *"moves the `loss` column (3.639 -> 0.107 for Qwen2.5-0.5B)"*, which reads as a before/after delta. `3.639 -> 0.107` is verbatim the existing published cell at line 76 — a start/end loss pair. I chased that as a suspicious 34x claim before checking the record. Say what it moves *to*, not restate the cell.

## LOW

- `warmup = int(ratio * total)` **floors**; transformers uses `math.ceil(num_training_steps * warmup_ratio)`. Same config gives a different warmup on the two backends (0.03 x 48 -> MLX 1, HF 2). Since #686 is about honouring the configured recipe, `ceil` matches.
- The non-constant branches degrade to a bare `ModuleNotFoundError: No module named 'mlx'` rather than a friendly error. Acceptable in production, since `train()` gates on `self._require_mlx()` first — but `build_lr_schedule` is a module-level helper and an `MlxOptimizerError` wrapper is two lines.
- `getattr(optim, "Muon")` on an mlx build without `Muon` raises `AttributeError`, not `MlxOptimizerError`.
- #733 emits `MLX backend ignores: ...` and this one emits `MLX backend: ...`; `_check_unsupported` uses the former.
- `docs/backends-and-ops.md` (Apple Silicon) is not updated with the supported optimizer and scheduler sets, and this PR introduces new **hard refusals**. That is a user-visible behaviour change and belongs in the docs.

**No shipped recipe breaks.** A grep for `optimizer` / `scheduler` in `recipes/catalog.py` returns nothing — all three MLX recipes use the defaults (`adamw_torch`, `cosine`), both supported. Their LR curve does change, which is the fix.

## The combined merge with #733 — your public correction was right, and I checked the semantics too

```
git merge --no-ff refs/tmp/h734   (onto 733)   -> exit 0, no conflicts
git merge --no-ff origin/main     (onto combo) -> conflicts: 0
```

Conflict-free is the weaker claim; I checked the three places the two PRs touch the same function:

1. `import mlx.optimizers as optim` is deleted by this PR, and a grep for `optim.` on the merged tree returns **no matches** — #733 introduced no use of it, so there is no dangling reference.
2. `adapter_config.json`: #733's three keys and this PR's six metadata keys coexist with **no collision**.
3. The harness YAML: #733's pin lands in `data:`, yours in `training:`, both intact.

Combined suite on current main: **361 passed, 16 skipped**, ruff clean, and `test_cli_startup_is_light.py` green — both new modules are import-light and imported inside `train()`.

**One combined-semantics note.** #733's harness pin sets `train_on_responses_only: false`, so the *benchmark harness* never exercises the new masked path. Coverage is not lost: the `mlx-smoke` job runs `test_smoke_train.py::test_mlx_sft_smoke`, whose fixture uses `format: chatml` and leaves the flag at its `true` default, so the masked path and your new cosine schedule both run end to end on macos-14.

## What I could not verify

No Apple Silicon and no mlx here. Your "10 distinct learning rates across 40 iterations", and the harness comments' `2,130 -> 342` tokens and `108.1 -> 26.1` tok/s, I did not reproduce on Metal. `test_mlx_advances_the_schedule_once_per_optimizer_update` — the load-bearing assumption that mlx drives a callable `learning_rate` from `opt.step` — is correct as written and `importorskip`-gated, so it runs on no CI job either. `_NO_WEIGHT_DECAY` against the real mlx constructor signatures, and whether `Muon` exists at your `mlx>=0.20.0` floor, are both unverified.

MakazhanAlpamys:
@Srinivasan8888 — **better tested than #733, and you mutation-tested your own work, which shows. Blocking on two of the four advertised behaviours surviving mutation in CI.**

## The four, each mutated separately

| behaviour | mutation | result |
|---|---|---|
| warmup | `warmup = 0` | **killed** — 8 distinct tests |
| optimizer choice | `resolve_optimizer_name` always returns `"AdamW"` | **killed** — 8+ tests |
| scheduler | always `cosine_decay`, ignore the configured name | **SURVIVED** |
| weight decay | drop `kwargs["weight_decay"]` in `build_optimizer` | **SURVIVED** |

**They do not share a single assertion** — the #698 defect is absent, each behaviour has its own dedicated test. The problem is *gating*, not sharing.

The scheduler killer, `test_linear_decays_to_zero_too_but_differently`, is an unusually good control: it samples at steps 20 and 40, and the docstring records that it *first* sampled at step 30, where cosine and linear cross at exactly 1.0e-4 and the test was vacuous. That instinct is exactly right. But it calls `lr_at()`, which does `pytest.importorskip("mlx.core")`, so it skips on every CI runner — **9 of 36 tests skip here and in CI**.

The weight-decay killer is likewise mlx-gated, and the one wrapper-level test is honestly named `..._reaches_the_metadata` and asserts metadata only. It cannot catch the mutation, because the shared fake is `AdamW = lambda **kwargs: object()`, which discards its kwargs.

**Both are fixable without Apple Silicon.** Make the shared fake's `AdamW` **record** its kwargs and assert `weight_decay` reaches it; add a schedule-builder-level control that distinguishes `linear` from `cosine` against recorded builder calls rather than against a sampled curve. Then both mutations die in CI, and #686's fourth criterion — the observed LR at first / boundary / final update — stops depending on hardware nobody in CI has.

## What is right

**Optimizer choice is not scope creep** — it is #686's third acceptance criterion verbatim: *"Map supported optimizer names and reject unsupported names instead of silently using AdamW."* It does widen the surface: `training.optimizer` now **hard-fails** on MLX for anything outside the map, where it previously became AdamW silently. That is the requested behaviour and a deliberate breaking change for anyone on e.g. `adam_mini`; it belongs in the release notes and your fragment should say it plainly.

**The schedule is real `mlx.optimizers`**, not hand-rolled — `join_schedules([linear_schedule(0, peak, warmup), cosine_decay(peak, total - warmup)], [warmup])`. And the touch I like most: a constant schedule with no warmup returns a plain float *before* the mlx import, so the common case constructs exactly what the old code did and stays testable off Apple Silicon. That is the pattern the two blocking items above need extended.

**Verified against real sources, not stubs.** All ten optimizer classes exist in mlx v0.31.2; `cosine_decay(init, decay_steps, end=0.0)`, `linear_schedule(init, end, steps)` and `join_schedules(schedules, boundaries)` match your call sites exactly. And `_NO_WEIGHT_DECAY = {Adam, Adagrad, Adamax, AdaDelta, RMSprop}` is **exactly right** — every constructor was AST-parsed; those five take no `weight_decay` and the five that do are correctly outside the set. That set is the kind of thing that is usually one name wrong.

**`TestTheWrapperPassesUpdatesNotIterations` exists because you mutation-tested your own work.** The comment says plainly that mutating `iters // grad_accumulation_steps` to `iters` "survived everything above", and the test now kills it with a paired `accum=1` control that localises the divisor. That is the standard I would like the rest of both PRs held to, and it is your own.

**One removed test line**, read and accepted: `opt_mod.AdamW = lambda learning_rate=None: sentinel` becoming `lambda **kwargs: sentinel` in `test_review_fixes_v07133.py`. Necessary — the old fake would `TypeError` on `weight_decay` — commented, and the surviving assertion unchanged. It does lose an implicit "only `learning_rate` is passed" constraint that was never asserted, and a recording fake restores it properly.

## Two small things

- Your fragment says "only 8 of the 32 names on Soup's optimizer allowlist have an MLX equivalent"; the map has 13 keys resolving to 10 classes. Worth reconciling.
- #733 and #734 do **not** conflict. Trial-merged in both orders: byte-identical trees, and the shared metadata dict auto-merges correctly. Combined tree runs 129 passed / 16 skipped across the six affected MLX suites. Your stated sequencing discipline in both threads was honoured, which is worth saying since it is rarely true when two PRs touch one file.

Make the fake record its kwargs and I merge.

Srinivasan8888:
Fixed in `abcd3f3b`. **All three survivors killed, and killed on a runner with no MLX — which is where they were surviving.**

Your diagnosis was exactly right and I want to state it back, because it is the part I got wrong twice now: every assertion I wrote reads the **plan**, and `as_metadata()` is derived from that same plan, so no metadata test can ever observe a broken `build_optimizer`. I tested the layer I designed instead of the layer that runs.

## The three, reproduced then killed

`TestThePlannedOptimizerIsTheOneConstructed` records `(class name, kwargs)` at the constructor through the fake-MLX harness. No `importorskip`, so it runs on Linux and Windows. Under my simulated no-mlx runner (blocks fresh `mlx*` imports, still resolves `sys.modules` — the correction you saw me make on the Windows fix):

```
baseline                                        35 passed, 9 skipped
M1  kwargs["weight_decay"] = ...  -> pass        1 failed, 34 passed
M2  getattr(optim, name)          -> AdamW       4 failed, 31 passed
M3  build_optimizer -> AdamW(learning_rate=lr)   6 failed, 29 passed   <- #686 verbatim
```

M3 is the one that matters and it needed its own assertion: it keeps the class right *and* the metadata right, so only `callable(learning_rate)` on a warmup+cosine config distinguishes it.

**Two controls**, because an assertion that holds for everything measures nothing:

- `adagrad` is constructed **without** `weight_decay` — passing it would `TypeError` on real MLX. So the kwarg is present because the plan says so, not unconditionally.
- the no-warmup constant path still passes a plain float, so `callable` is discriminating rather than always-true.

## Two facts I found while writing them

**Five `_OPTIMIZER_MAP` keys are unreachable from any valid config.** `adam`, `adamw`, `lion`, `adamax`, `adadelta` are not in `utils.optimizer_zoo`, so `TrainingConfig` rejects them before `resolve_optimizer_name` ever sees them:

```
reachable:    adafactor adagrad adamw_hf adamw_torch adamw_torch_fused muon rmsprop sgd
refuse decay: adagrad rmsprop
```

My first draft of the parametrised test used `adam` and `lion` and died on a `ValidationError` — which is how I found it. Harmless (dead map entries, not wrong behaviour) but it means `_NO_WEIGHT_DECAY` is only reachable through those two names, so I pinned the control on `adagrad`.

## Your other items

**`weight_decay` pinned in the harness** — you were right that it is safe by coincidence. Pinned at `0.01` with a comment saying it is pinned *because* the schema default happens to equal MLX AdamW's default, not because it moves today.

**The harness comment reworded.** You are right that `3.639 -> 0.107` reads as a 34x before/after delta; it is verbatim the record's existing start/end cell at line 76. It now says the end value moves, rather than restating the cell. Sorry for the chase.

**`docs/backends-and-ops.md`** now has an "Optimizers and schedules" section: the honoured settings, the three refusal classes as a table, the update-vs-iteration unit, and the `adapter_config.json` keys. Agreed this belongs in docs — hard refusals are a user-visible behaviour change and I should not have needed telling.

```
pytest <686 + 634 + 684 + 685 + both #23 bridges + test_cli_startup_is_light>   342 passed
pytest tests/test_issue23_mlx_harness_bridge.py                                 198 passed
ruff check src/soup_cli/ tests/ benchmarks/                                     All checks passed!
git diff -- tests/ | grep -c "^-[^-]"                                           0
```

## The one thing I did not do, and why I am asking instead

**`ceil` vs `int()` for warmup.** You are right that transformers uses `math.ceil` and that #686 is about honouring the recipe, so `ceil` is the matching choice. But `ceil` makes `warmup >= 1` for *any* `warmup_ratio > 0`, which makes the `warmup_ratio=0.03 over 12 updates rounds to 0` warning — and the two tests pinning it — unreachable. That is arguably fine (the warning exists to report a problem `ceil` removes), but it means deleting a warning and its tests, and I would rather not do that unannounced on a PR you are reviewing for exactly that kind of quiet weakening.

Say which you want and I will push it:
1. `ceil`, drop the rounds-to-zero warning and its two tests as obsolete, and say so in the fragment; or
2. keep `int()`, and I document the one-step divergence from transformers in the docs section instead.

I lean **1** — the backends agreeing is worth more than a warning about a behaviour that would no longer happen.

Still unverified from my side, unchanged: nothing re-run on Metal since `b04288cb`, and `_NO_WEIGHT_DECAY` against real MLX constructor signatures plus whether `Muon` exists at the `mlx>=0.20.0` floor are still only pinned by the `importorskip` test.

MakazhanAlpamys:
@Srinivasan8888 — **the three you killed are genuinely dead, and I verified all three myself. Still blocking: the item from my 07:42 comment was not the item you answered, and the mutation it names survives your new tests on this box.**

## First, what is fixed

`TestThePlannedOptimizerIsTheOneConstructed` does what you said it does, and it runs on a machine with no MLX — I checked `find_spec("mlx") is None` before trusting a single number. Your counts are exact, not approximately right:

| mutation | your count | mine |
|---|---:|---:|
| `kwargs["weight_decay"] = ...` -> `pass` | 1 | **1** |
| `getattr(optim, plan.optimizer_name)` -> always `AdamW` | 4 | **4** |
| `build_optimizer` -> `AdamW(learning_rate=lr)` (#686 verbatim) | 6 | **6** |
| `total_updates = iters` | — | **1** |

Baseline **116 passed, 9 skipped**. Your diagnosis of your own mistake was right and I want it on the record because it generalises: every assertion read the plan, and `as_metadata()` is derived from that same plan, so no metadata test could ever observe a broken `build_optimizer`.

## What is still open, and why we are out of sync

Two of my comments listed different sets. Timestamps, so this is not a matter of memory:

- **06:46** — weight_decay, `getattr`, `build_optimizer`. **You answered this one at 09:26.**
- **07:42** — the later block, two items: (1) make the fake record kwargs; (2) *"add a schedule-builder-level control that distinguishes `linear` from `cosine` against recorded builder calls rather than against a sampled curve."*

(1) is done. (2) is not, and it is the one that matters.

## The mutation, and it is worse than "always cosine"

`src/soup_cli/trainer/mlx_optim.py:191-195`. I swapped the two branch bodies — nothing else:

```python
    elif plan.scheduler == "cosine":
        body = optim.linear_schedule(peak, 0.0, decay_steps)   # was cosine_decay
    else:  # linear
        body = optim.cosine_decay(peak, decay_steps)           # was linear_schedule
```

Proof the mutation is live, through a recording fake, on your head:

```
plan.scheduler='cosine'   -> builder called: ['linear_schedule']
plan.scheduler='linear'   -> builder called: ['cosine_decay']
```

The whole suite under it:

```
116 passed, 9 skipped
```

**Every user asking for `cosine` gets a linear decay, every user asking for `linear` gets a cosine, and this PR's tests are green.** That is the exact class of defect #686 is about, one layer over — the plan says `cosine`, `as_metadata()` writes `cosine` into `adapter_config.json`, and the object handed to the optimizer is the other one.

And a third, which is mine rather than from the earlier list — dropping the warmup ramp entirely:

```python
    if True:            # was: if warmup <= 0
        return body
    warm = optim.linear_schedule(0.0, peak, warmup)   # unreachable
```

```
116 passed, 9 skipped
```

`warmup_updates: 8` still lands in the adapter metadata; no ramp is ever built. Your Metal table showing `7.500e-05` at iteration 4 is exactly the evidence that would catch it, and it is not in CI — it is in a PR comment.

The reason all three survive is the same one line: the only linear-vs-cosine assertion you have, `test_linear_decays_to_zero_too_but_differently`, samples the curve through `lr_at()`, which is behind `importorskip("mlx.core")` at line 43. It is one of the 9 skips above. On every runner this project has, the schedule's **identity** is unverified; `callable(lr)` proves only that *a* schedule exists.

## What closes it

The parts already exist in your own harness. `_install_fake_mlx` in `tests/test_issue634_mlx_resume.py:401-403` already defines `linear_schedule`, `cosine_decay` and `join_schedules`. Record those three the way `_record_optimizer_calls` records constructors, and assert:

- `scheduler: cosine` calls `cosine_decay` and not `linear_schedule` for the body;
- `scheduler: linear` calls `linear_schedule` and not `cosine_decay`;
- `warmup_ratio > 0` produces a `join_schedules` call whose boundary is `warmup_updates`.

Three assertions against recorded calls kill all three mutations above without Apple Silicon.

## Everything else checked and clean

Zero test lines removed (the one deletion is the documented `AdamW = lambda **kwargs:` fake widening, with its assertion intact). `changelog.d/0.74.0/734.fixed.md` names #734, credits you, and states the hard refusal as a breaking change; the "8 of 32" figure is now consistent with the reachable names and the docs table. The new "Optimizers and schedules" section in `docs/backends-and-ops.md` is there with the three refusal classes and the update-vs-iteration unit. `weight_decay: 0.01` is pinned in the harness with the "safe by coincidence" rationale. Ruff clean.

Nine MLX-gated tests never ran here, so `_NO_WEIGHT_DECAY` against real constructor signatures, `Muon` at the `mlx>=0.20.0` floor, and the ten-distinct-rates claim remain yours to stand behind — which is fine, they are measurements, not CI guarantees.

On your open `ceil` vs `int()` question: **`int()`, i.e. what you have.** A trailing partial accumulation window is not an optimizer update, so counting it would put the schedule's end one update past the last `optimizer.update()` call and the cosine would never reach its floor — the same failure you opened this PR by measuring. Leave it.

Srinivasan8888:
Fixed in `5b1fea05`. **You were right that I answered the wrong comment, and the mutation you found is worse than the three I did answer.**

I read your 06:46 list, fixed it, and never went back to 07:42. Item (2) there names exactly this, in those words — "against recorded builder calls rather than against a sampled curve" — so this was not a gap in the review, it was a gap in my reading of it.

## The mutation, reproduced then killed

On a simulated no-mlx runner:

```
baseline                                   41 passed, 9 skipped
swap the cosine / linear branch bodies      4 failed
drop the warmup ramp (if warmup <= 0 -> if True)   1 failed
```

`_record_schedule_calls` records the three builders `_install_fake_mlx` already defines, and `TestTheScheduleBuiltIsTheScheduleRequested` asserts the recorded call. `scheduler: cosine` must call `cosine_decay` and **not** `linear_schedule`; `linear` the reverse. Both halves, because passing only one is precisely what a swapped pair looks like.

Three controls so the assertions discriminate rather than hold for everything:

- `constant` builds **no** decay curve at all (`names == []`)
- no warmup joins nothing (`join_schedules` absent)
- the ramp's own arguments: rises from `0.0` over `12` updates, joined at boundary `[12]`, and `cosine_decay` gets `decay_steps=36` — 48 total minus 12 warmup, so `total - warmup` is pinned at the builder instead of by sampling

## The part I want to name, because it is the third time

`callable(lr)` was me testing that *a* schedule exists and calling it "the schedule is wired". Same shape as asserting the plan and calling it "the optimizer is built", and as asserting `_ALLOWED_KEYS` and calling it "the field reaches the wire" on #713. Each time the assertion sat one layer above the thing that can actually be wrong. The general fix is the one your reviews keep applying: **assert the call the code makes, not the value it planned to make it with.**

## On `ceil` vs `int()` — taking your ruling, and your reason is better than mine

Keeping `int()`. Your argument settles it: a trailing partial accumulation window is not an optimizer update, so counting it puts the schedule's end one update past the last `optimizer.update()` and the cosine never reaches its floor — which is the failure I opened this PR by measuring. My cross-backend-agreement argument was reasoning from a different backend's convention instead of from what this one actually calls. No change made, and the rounds-to-zero warning stays.

```
pytest tests/test_issue686_mlx_optimizer_schedule.py   50 passed
simulated no-mlx runner                                41 passed, 9 skipped
ruff check src/soup_cli/ <the test file>               All checks passed!
git diff -- tests/ | grep -c "^-[^-]"                  0
```

Test-only commit; `src/` is byte-identical to `abcd3f3b`, so nothing you verified has moved.

Still yours-to-my-word rather than CI-guaranteed, unchanged: `_NO_WEIGHT_DECAY` against real constructor signatures, `Muon` at the `mlx>=0.20.0` floor, and the ten-distinct-rates Metal measurement. The nine `importorskip` tests still run on no CI job — I have not tried to change that here, since the identity assertions above no longer depend on them.

MakazhanAlpamys:
@Srinivasan8888 — **the item from my 07:42 comment is genuinely dead, I verified it myself, and all eight mutations across the three rounds now die on a runner with no MLX. I am blocking on a ninth that is not on any of my earlier lists, because I extended your own method to the one branch nobody mutated and it survives.**

First, so it is not buried: your numbers reproduce exactly. `D1=1`, `O1=4`, `V1=6` match my 17:33 table, and `S1=4`, `W2=1` match your 19:10 claim. Both of us are counting the same thing.

## The survivor

`src/soup_cli/trainer/mlx_optim.py:190` — the post-import constant branch — changed to `if False:`, so `constant` and `constant_with_warmup` fall through to the `linear` body:

```
                      clean head                              under the mutation
constant     wu=0     float,    builders=[]                   float,    builders=[]
constant     wu=25    callable, [linear_schedule,             callable, [linear_schedule,
                                 join_schedules]                         linear_schedule,
                                                                         join_schedules]
constant_with_warmup wu=25   (same)                           (same)

tests/test_issue686_mlx_optimizer_schedule.py:  41 passed, 9 skipped   <- BOTH
```

A user who asks for `scheduler: constant` with any warmup gets a **linear decay to 0.0**, while `as_metadata()` writes `scheduler: constant` into `adapter_config.json`. That is #686's own sentence one branch over: the metadata reports one thing and the optimizer does another.

It is reachable. `_SUPPORTED_SCHEDULERS` (`mlx_optim.py:24`) accepts all four, `TrainingConfig.scheduler` is a free-form `str`, and `docs/backends-and-ops.md:132` documents `constant_with_warmup` to users.

## Why it survives, and this is the part worth reading

**You wrote the right test. It cannot run in CI.**

`test_constant_holds_the_peak_after_warmup` (line 91) asserts `lr_at(plan, 49) == 2e-4`, which is exactly the assertion that kills this mutation. But `lr_at()` opens with `pytest.importorskip("mlx.core")` at line 43, so it is one of the six skips at that line on every runner this project has:

```
SKIPPED [6] tests/test_issue686_mlx_optimizer_schedule.py:43: could not import 'mlx.core'
```

And the builder-level counterpart cannot cover for it. `test_a_constant_schedule_builds_no_decay_curve_at_all` (line 623) is labelled a reject-everything control, but it passes `warmup_ratio=0.0`, which returns at line 185 **before the MLX import ever happens**:

```python
if warmup <= 0 and plan.scheduler in ("constant", "constant_with_warmup"):
    return peak
```

So `names == []` holds for every possible implementation of the branch at line 190. It asserts a property of the early return, not of the thing its name claims to control.

Your own file already says this, three times — lines 400, 418 and 586 are comments explaining that an `importorskip("mlx.core")` test runs on no CI job, which is precisely why you built the builder-recording harness. You applied that reasoning to cosine, to linear and to the warmup join. `constant` is the fourth scheduler and it did not get the same treatment.

**I want to be fair about where this came from.** It is a new finding on round four, it is not what my 07:42 or 17:33 comments named, and you could not have inferred it from them — those named cosine/linear and the ramp specifically. You answered what I asked.

## Closing it costs two assertions

Using `_builders`, which is already in the file:

```python
def test_constant_with_warmup_builds_a_ramp_but_no_decay_curve(self, tmp_path, monkeypatch):
    _, names = self._builders(
        tmp_path, monkeypatch, scheduler="constant_with_warmup", warmup_ratio=0.25
    )
    assert "cosine_decay" not in names
    assert names.count("linear_schedule") == 1   # the ramp, and nothing else
```

I wrote and ran that: it passes on your head and fails under the mutation, with no MLX present.

## Two ways to finish this, your call

1. **You push it.** Two assertions, and I merge on green.
2. **I write it and merge, credited to you** — your commits, your `CONTRIBUTORS.md` entry, your name in the fragment, and I say in the thread that the two assertions are mine and why. Say the word and I will do it today.

I am offering the second because this is round four on a finding I introduced, not one you left open, and I would rather spend my evening on it than yours. Neither option is the better answer; pick whichever suits you.

## Everything else is clean

- Eight mutations, eight killed, on `find_spec("mlx") is None`. The `S1` swap of the cosine/linear bodies — the 07:42/17:33 item — now fails four named tests.
- `git diff origin/main...5b1fea05 -- tests/` removes exactly one line, the documented `AdamW` fake widening, with the surrounding assertion intact and a comment saying why.
- `changelog.d/0.74.0/734.fixed.md` is named after the PR and credits you.
- `benchmarks/harness/mlx_sft_smoke.py` pins `scheduler`, `warmup_ratio` **and** `weight_decay`, with a comment recording that the last is pinned because the schema default coincides with MLX AdamW's own rather than because it moves today. You pre-empted that one rather than being told, and it is the reason the published record survives this merge.
- Ruff clean. 14 of 14 green.

Srinivasan8888:
Pushed in `47f45ca6` — **option 1.** Your offer to write it yourself and credit me was generous, and I would rather not spend it on two assertions.

## The survivor is dead, and the control was the worse half

Your mutation, applied to the post-import branch, on `.venv-ci` — a **real** no-mlx environment rather than my simulated one:

```
clean head        43 passed, 9 skipped
if False:          2 failed   (both parametrisations)
```

I applied it in binary mode and checked the file's sha256 before and after restore — `f428abaf -> 628e9e16 -> f428abaf` — because of your CRLF disclosure below. That disclosure is the most useful thing in your comment and I have adopted the practice rather than just reading it.

**But the control is the part I want to own.** `test_a_constant_schedule_builds_no_decay_curve_at_all` passed `warmup_ratio=0.0`, which returns at the early `return peak` **before the MLX import** — so `names == []` held for every possible implementation of the branch it was named after. I labelled it a reject-everything control and it controlled nothing about `constant`. That is worse than the missing test: a wrong test name tells the next reader the branch is covered when it is not.

It is renamed to `test_a_constant_schedule_with_no_warmup_builds_nothing_at_all`, with a comment saying it controls the cosine/linear pair and explicitly cannot cover the constant branch. Your test covers that branch, parametrised over both spellings since both reach the same line.

## On how this one got missed

Your fairness about it is appreciated but I do not think I need it. Your 07:42 and 17:33 comments named cosine and linear because those were the mutations you ran; the *method* you handed me was "assert the recorded builder call, because the sampled-curve test runs on no CI job", and `_SUPPORTED_SCHEDULERS` has four entries. I applied the method to three of them. My own file says the importorskip problem out loud at lines 400, 418 and 586 — I wrote those comments and still did not enumerate the branch list against them.

The generalisable version, which I am writing down rather than re-deriving next time: **when a test class exists because one code path is untestable in CI, enumerate every value that reaches that path and cover each one** — do not cover the values that happen to appear in the review comment.

```
pytest tests/test_issue686_mlx_optimizer_schedule.py (.venv-ci)  43 passed, 9 skipped
pytest <686 + 634 + 684 + 685> (.venv-ci)                        96 passed, 9 skipped
ruff                                                             All checks passed!
git diff origin/main...HEAD -- tests/ | grep -c "^-[^-]"         1  (the documented AdamW fake widening)
```

Test-only commit; `src/` is byte-identical to `5b1fea05`.

One correction to my own earlier verification method, since it affects how much my numbers are worth: the no-mlx "simulation" I have been quoting all along over-blocks. On `origin/main` it reported `TestVllmTrustRemoteCode::test_serve_vllm_passes_resolved_trust` failing; a genuine no-mlx venv passes all 28 in that file. Every count in this comment is from the real venv. The counts in my previous comments were not wrong — I re-checked the ones that mattered — but the tool that produced them could have invented a failure, and you should weigh them accordingly.

MakazhanAlpamys:
@Srinivasan8888 — merged as `765f5b4c`. **The survivor is dead and I re-ran it myself rather than reading your reply.**

```
baseline                                   43 passed,  9 skipped
post-import constant branch -> if False:    2 failed, 41 passed, 9 skipped

FAILED TestTheScheduleBuiltIsTheScheduleRequested::test_constant_with_warmup_builds_a_ramp_but_no_decay_curve[constant]
FAILED TestTheScheduleBuiltIsTheScheduleRequested::test_constant_with_warmup_builds_a_ramp_but_no_decay_curve[constant_with_warmup]
```

Applied in binary mode against the four-space-indented standalone branch, so the early return at line 185 could not be hit by accident — occurrence count asserted at 1, sha `2cdb1e58` -> `98d8154f`, and the line grepped back to confirm only line 190 moved. Nine mutations at merge, nine killed, all on an interpreter where `find_spec("mlx") is None`.

Three things about how you closed it that are worth naming, because they are not the obvious way to do it.

**You parametrised both spellings.** `constant` and `constant_with_warmup` both reach that branch and the mutation kills both parametrisations. A test covering only the first would have left the documented one open.

**You asserted both halves.** `"cosine_decay" not in names` **and** `names.count("linear_schedule") == 1`. The second is the one that catches a fall-through to the linear body specifically — the first alone would pass on exactly the bug I found.

**You renamed the old control instead of deleting it.** `test_a_constant_schedule_builds_no_decay_curve_at_all` -> `test_a_constant_schedule_with_no_warmup_builds_nothing_at_all`, with a docstring that says out loud that `warmup_ratio=0.0` returns before the MLX import so `names == []` holds for every implementation, and that it controls the cosine/linear pair only. A control that claimed more than it checked is now a control that claims exactly what it checks. That is a better outcome than removing it, and it is the part I would not have thought to ask for.

You took option 1 and said you would rather not spend the offer on two assertions. Noted, and it stands for a future round.

One line removed from tests, and it is a widening: an `AdamW` fake pinned to `learning_rate=None` became a recording stub accepting any keyword, with the assertion unchanged. `changelog.d/0.74.0/734.fixed.md` is named after the PR and credits you. Ruff clean. 14 of 14 SUCCESS on `47f45ca6`, `head_sha` confirmed equal to the head I merged.

Not verified here: the Apple Silicon schedule measurements are yours and stay yours — this box has no Metal.

Issue #723, opened 2026-09-06
GSPO leaks masked tokens through batch centering and is not sequence-level


Amix29:
## Summary

Soup's `gspo` branch centers each token log-ratio across the batch before applying
the completion mask:

```python
log_ratio_centered = log_ratio - log_ratio.mean(dim=0, keepdim=True)
ratio_stab = torch.exp(log_ratio_centered)
token_loss = -(ratio_stab * advantages_2d)
return _masked_mean(token_loss, completion_mask)
```

Consequently, a masked token participates in the batch mean, changes the loss of
unmasked examples, and receives a non-zero gradient. The implementation also does
not match the published GSPO objective, which uses a length-normalized sequence
likelihood ratio and sequence-level clipping rather than per-column batch
centering: https://qwenlm.github.io/blog/gspo/

## Reproduction

Reproduced against current `main` at
`9f1e48b33a5fcda7e621e612ec9305cb52a38e07` with Python 3.12.2 and PyTorch
2.13.0, executing the actual `apply_variant_loss()` source.

In a 4x2 synthetic batch, changing only one masked log-ratio gives:

| Masked log-ratio | Loss | Gradient at masked token |
|---:|---:|---:|
| `-2` | `-0.3333333` | `0.0833333` |
| `-6` | `-0.9060939` | `0.2265235` |

DAPO, BNPO, and `two_sided` controls are invariant to the same mutation and return
zero gradient at that masked position.

## Impact

Selecting `grpo_variant: gspo` optimizes an objective influenced by padding or
otherwise masked completion positions. More broadly, the named variant is not the
sequence-level GSPO algorithm users reasonably expect.

This is distinct from #123 (initial kernel implementation), #584 (the variants
previously falling back to stock TRL), and #342 (non-finite gradients in the
standard `variant=None` path).

## Suggested fix

- Exclude masked positions before every statistic or reduction.
- Compute the masked, length-normalized log-probability ratio per completion.
- Apply GSPO clipping and the policy-gradient surrogate at sequence level according
  to the published definition.
- Preserve any KL term and normalization expected by the surrounding TRL contract.

## Acceptance criteria

- [ ] Changing a masked token cannot change the GSPO loss.
- [ ] Every masked token receives exactly zero gradient.
- [ ] The implementation matches a direct reference calculation of the published
      sequence-level objective, including variable completion lengths.
- [ ] Tests cover batch permutation, padding position, all-masked rejection, clip
[...]

PR #735, opened 2026-09-06
fix(grpo): exclude masked tokens from the gspo batch-mean statistic

AmirF194:
Fixed the masked-token leak: gspo's per-column centering mean now sums `log_ratio` only over unmasked positions (clamped so an all-masked column doesn't divide by zero), so a masked token can no longer shift another row's centering statistic or pick up a gradient of its own.

Added three tests: a masked token's value can't move the loss, a masked position gets exactly zero gradient, and an all-masked column stays finite. Ran the gspo/kernel/PRM/factory slice of `test_v05311.py` plus the runtime-contract test, 33 passed; the two new ones are red without this diff. `ruff check` is clean.

I didn't touch the sequence-level part. Making gspo match the published objective means adding per-sequence clipping, which needs a new epsilon and changes the loss shape for anyone already running `grpo_variant: gspo`, so that reads like its own PR once you've said what clip bounds you want. Happy to take that on too if you'd rather bundle it.

Refs #723

### DIFF
--- a/changelog.d/0.74.0/735.fixed.md
+++ b/changelog.d/0.74.0/735.fixed.md
@@ -1,0 +1,6 @@
+- `grpo_variant: gspo` centered the per-token log-ratio across the batch
+  before the completion mask was applied, so a masked (padding) token
+  shifted the loss and gradient of every unmasked row sharing its column,
+  and picked up a non-zero gradient of its own (#735 by @AmirF194, refs
+  #723). The batch-mean statistic is now taken over unmasked positions
+  only.
--- a/src/soup_cli/utils/grpo_variants.py
+++ b/src/soup_cli/utils/grpo_variants.py
@@ -275,8 +275,16 @@ def apply_variant_loss(
 
     if normalised == "gspo":
         # Group Stabilized: subtract per-group mean log-ratio (acts as
-        # control variate). Mean is taken over the batch dim.
-        log_ratio_centered = log_ratio - log_ratio.mean(dim=0, keepdim=True)
+        # control variate). Mean is taken over the batch dim, excluding
+        # masked (padding) positions so a masked token cannot shift the
+        # statistic other rows in its column are centered against.
+        if completion_mask is not None:
+            col_sum = (log_ratio * completion_mask).sum(dim=0, keepdim=True)
+            col_count = completion_mask.sum(dim=0, keepdim=True).clamp(min=1.0)
+            col_mean = col_sum / col_count
+        else:
+            col_mean = log_ratio.mean(dim=0, keepdim=True)
+        log_ratio_centered = log_ratio - col_mean
         ratio_stab = torch.exp(log_ratio_centered)
         token_loss = -(ratio_stab * advantages_2d)
         return _masked_mean(token_loss, completion_mask)
--- a/tests/test_v05311.py
+++ b/tests/test_v05311.py
@@ -85,6 +85,64 @@ class TestGRPOVariantKernels:
         )
         assert math.isfinite(float(out))
 
+    def test_gspo_masked_token_cannot_change_loss(self):
+        """A masked (padding) token must not shift the gspo batch-mean
+        statistic other rows are centered against (#723)."""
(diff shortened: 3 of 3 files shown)


MakazhanAlpamys:
Merged as `e1f5f69`. **Narrow, correct, mutation-proof, and honestly scoped — and the scope honesty is the part I want to name first.**

#723 reports two things: masked tokens leaking through batch centering, and `gspo` not being the published sequence-level objective. You fixed the first, said in your own body that you had not touched the second, titled the PR to the half you did, marked it `Refs #723` rather than `Fixes`, and wrote the changelog fragment the same way. **#723 stays open** for the sequence-level half. Most PRs would have taken the issue number and left me to discover the gap.

## What I verified on your head `0777878`

Worktree with `PYTHONPATH` pinned inside it. Same script run against clean `main` and against your branch — a 4x2 batch with one masked position, varying only that position's log-ratio:

| masked log-ratio | before | after |
|---:|---:|---:|
| -2.0 | -0.3320515454 | -0.2857143283 |
| -6.0 | -0.5344063640 | -0.2857143283 |
| +50.0 | -0.2142859995 | -0.2857143283 |

The post-fix loss is invariant, which is the whole claim. Gradient at the masked position: **0.0800301731 -> exactly 0.0**.

**The no-padding case is bitwise identical**, which is what makes this safe to ship to everyone:

```
mask all-ones:  before -0.400086909532547   after -0.400086909532547
                and identical to mask=None
```

Boundary cases all behave: all-masked row (changed, as intended), all-masked column (unchanged, and the `.clamp(min=1.0)` stops the divide-by-zero), entirely-masked batch (0.0, finite), single-token sequence (unchanged, finite), and `long` / `bool` / `int32` masks all accepted.

**The other six variants are untouched to the last digit** — dapo -0.3182034791, dr_grpo -0.6137486696, bnpo -0.4085744619, two_sided -0.3247749209, rft -0.0500000007, standard returns None — and each stays invariant to the masked value both before and after, matching the control claim in the issue.

**Mutations: 3 applied, 3 killed.**

| mutation | result |
|---|---|
| revert to `log_ratio - log_ratio.mean(dim=0)` | killed — `cannot_change_loss` + `zero_gradient` |
| drop `.clamp(min=1.0)` | killed — `all_masked_column_no_nan` |
| unmasked numerator | killed — `cannot_change_loss` + `zero_gradient` |

Source restored byte-identical after each round. No vacuity: your `_toy_inputs` is 2x4 against a mask whose columns 1-3 are fully masked, and `pytest.approx(0.0)` at 1e-12 against a pre-fix 0.08 is a real assertion, which the kills prove.

**Not dead code** — `apply_variant_loss` is called at `trainer/grpo.py:138` with the real `completion_mask` (times `tool_mask` where present), so this is on a live path.

Zero deleted test lines. Regression sweep on `-k "grpo or v05311"`: your branch 5 failed / 628 passed against main's 5 failed / 625 passed — identical failure set, delta is exactly your three new tests. All five are the pre-existing #651 floor (`cannot import name 'FSDPModule'` at torch 2.5.1 / trl 0.29.1), not yours. Ruff clean. Fragment `changelog.d/0.74.0/735.fixed.md`, PR number, credits you.

`docs/training.md:535` already describes gspo as a "group-stabilised importance ratio with per-batch control variate" — honest, it never claimed the paper's objective, so no docs change was needed. The variant *name* still implies the paper; that is the open half of #723, not your debt.

## One thing I could not verify

The real-`trl` end-to-end proof that the gspo branch is reached in a live `GRPOTrainer` is red on this box from the #651 floor. There is a mock-TRL runtime contract test (`test_v05311.py:676` and `:727`) that does cover gspo, so the wiring is pinned — but I am recording that the live-runtime leg is unconfirmed here rather than implying I ran it.

Thank you. This is the shape I would point other contributors at: fix one thing, measure it, say what you did not do.

AmirF194:
Kept it Refs rather than Fixes on purpose, the sequence-level objective needs a different code path than a bigger diff on this fix would have covered. The all-masked-column clamp was the boundary case I was least sure about, good to see it held under the mutation.

Issue #722, opened 2026-09-06
Add token-chunked checkpointed KL distillation to reduce retained activations


Amix29:
## Motivation

The current logit-distillation kernel materializes probability/log-probability and
divergence intermediates for every token across the full vocabulary, then applies
the supervised-token mask. For long sequences or large vocabularies, these saved
autograd tensors can be a material part of training memory.

## Measured prototype

An isolated forward-KL prototype was rerun with Python 3.12.2 and PyTorch 2.13.0
using 512 positions, vocabulary 4096, 256 supervised positions, and 32-token
chunks. It keeps the full vocabulary and uses one global denominator.

Unique storage retained by autograd after the forward pass, excluding the already
materialized student and teacher logits:

| Implementation | Retained bytes |
|---|---:|
| Dense current formulation | 16,777,736 |
| Selected tokens + chunks, no checkpoint | 8,389,128 |
| Selected tokens + chunks + non-reentrant checkpoint | 520 |

Loss error versus dense: `0`. Maximum gradient error: `0` on this rerun.

These numbers are **not** peak VRAM, total process memory, or an end-to-end speed
benchmark. Model activations, the full logits, CE loss, and backward recomputation
temporaries still exist. Checkpointing trades compute for retained activation
storage and may reduce throughput.

## Proposed implementation

- Select only causally aligned supervised positions before evaluating KD.
- Process those positions in configurable token chunks while keeping the complete
  vocabulary distribution.
- Accumulate loss numerators and use one global supervised-token denominator.
- Use non-reentrant activation checkpointing per chunk when enabled.
- Start with forward-KL; extend to reverse-KL and JS only after their stable
  log-space formulation is in place.
- Keep the dense path as a fallback and measure the crossover point.

## Acceptance criteria

- [ ] Dense and chunked losses/gradients agree within dtype-appropriate tolerance.
- [ ] Causal shift, response-only masks, padding, and all-masked inputs are tested.
- [ ] Peak GPU memory and throughput are measured on the same model, batch,
      sequence length, vocabulary, and token budget.
- [ ] Documentation describes the compute/memory tradeoff and does not present
      retained-tensor bytes as total VRAM saved.
- [ ] Reverse-KL/JS use the stable kernel rather than reproducing their current
      non-finite-gradient behavior.

kok-o:
Hi @MakazhanAlpamys, I would like to work on this issue.

My implementation plan:
1. Isolate causally aligned supervised token positions before computing KD.
[...]

PR #736, opened 2026-09-06
fix(distill): stabilize reverse KL and JS

### LINKED ISSUE #719: Reverse-KL and JS distillation can return finite losses with NaN gradients
## Summary

`src/soup_cli/trainer/distill.py::_compute_distill_term()` can produce non-finite
student gradients for the `reverse_kl` and `js` divergences even when all input
logits and the scalar loss are finite.

The reverse-KL path passes the differentiable student probabilities as the target
of `torch.nn.functional.kl_div`. At low, schema-valid temperatures, softmax can
underflow to zero and the target-side derivative becomes non-finite. The JS path
has the same target-side problem and additionally constructs `log_m` through a
probability-space mixture with a fixed `1e-12` clamp, which is not representable as
intended in FP16.

## Reproduction

Reproduced against current `main` at
`9f1e48b33a5fcda7e621e612ec9305cb52a38e07` with Python 3.12.2 and PyTorch
2.13.0 on CPU. The isolated test uses finite logits, temperature `0.05`, causal
shifting, and a mask with one genuinely supervised position.

| Divergence | Loss | Student gradients |
|---|---:|---|
| forward KL, FP32 | `0.5` | finite |
| reverse KL, FP32 | `0.1499999911` | **non-finite** |
| JS, FP32 | `0.00173286791` | **non-finite** |
| JS, FP16 | `NaN` | **non-finite** |

The same failure is reproduced for reverse-KL and JS in BF16. A log-space
reformulation preserves the tested loss values and returns finite gradients.

This is a kernel-level reproduction, not an end-to-end CUDA/MPS training run.

## Impact

A distillation run can report a finite forward loss while poisoning the backward
pass. The issue is reachable through a documented divergence choice and a valid
temperature; it is not an all-masked-batch case.

## Suggested fix

- Compute the distillation kernel in FP32 with `log_softmax`.
- Implement reverse KL explicitly as
  `sum(exp(log_s) * (log_s - log_t))`.
- Build the JS mixture in log space with
  `log_m = logaddexp(log_s, log_t) - log(2)`.
- Preserve the existing causal shift, trained-token mask, and `temperature ** 2`
  scaling, then cast only the final scalar as appropriate.
- Treat already non-finite input logits separately rather than hiding them.

## Acceptance criteria

- [ ] Reverse-KL and JS return finite losses and finite student gradients for the
      low-temperature reproducer in FP32, BF16, and FP16.
- [ ] Forward-KL behavior remains numerically unchanged on normal inputs.
- [ ] Tests cover causal shifting, response-only labels, padding masks, and at
      least one supervised token.
- [ ] A mutation back to the current probability-space implementation makes the
      regression test fail.

be-student:
Fixes #719.

Computes reverse KL and JS in FP32 log space, preventing probability underflow from producing non-finite gradients. Non-finite input logits now fail explicitly.

Validation: 20,409 tests passed; Ruff passed; coverage 82.99%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,14 +10,6 @@ Detailed, per-release notes for every published version live on the
 file tracks unreleased changes and links out for historical detail rather than
 reproducing 70+ versions of notes.
 
-## [Unreleased]
-
-### Fixed
-
-- Reverse-KL and Jensen-Shannon distillation now use stable FP32 log-space
-  formulas, preventing finite losses with non-finite gradients at low
-  temperatures (#719).
-
 ## [0.74.0] - 2026-09-04
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ reproducing 70+ versions of notes.
 
 ## [Unreleased]
 
+### Fixed
+
+- Reverse-KL and Jensen-Shannon distillation now use stable FP32 log-space
+  formulas, preventing finite losses with non-finite gradients at low
+  temperatures (#719).
+
 ## [0.74.0] - 2026-09-04
 
 ### Added
--- a/changelog.d/0.74.0/736.fixed.md
+++ b/changelog.d/0.74.0/736.fixed.md
@@ -1,0 +1,10 @@
+- Stabilize distillation loss and gradients in log space (#719 by @be-student).
+  All three divergences, including forward-KL, now compute and return FP32
+  scalars for FP16/BF16 inputs. Non-finite logits propagate for AMP step skipping;
+  no per-step finite-value guard synchronizes the device. The FP32 trade-off is
+  documented in `docs/training.md`: the maintainer measured 1.71x time and 1.24x
+  peak memory for the original guarded kernel, and removing the guards reduced
+  time from 10.91 ms to 9.40 ms (main: 6.39 ms). This revision also avoids unused
+  probability tensors. Two FP32 logits copies use about 7.8 GiB at
+  B=4/S=2048/V=128256. These measurements are hardware-specific and were not
+  rerun for this revision (PR #736).
--- a/docs/training.md
+++ b/docs/training.md
@@ -425,9 +425,22 @@ Loss = student CE + (T**2) × KL(teacher_logits / T  ||  student_logits / T).
 Teacher is loaded once, frozen via `requires_grad_(False)` + `.eval()`, and its
 inputs / logits are auto-bridged across CPU / CUDA devices.
 
-The token-divergence kernel evaluates probability math in FP32, including when
-the model logits use FP16 or BF16, so low temperatures do not create non-finite
-reverse-KL or Jensen-Shannon gradients.
+The token-divergence kernel evaluates all three divergences in FP32 and deliberately
+returns an FP32 scalar, including for FP16/BF16 logits. Forward-KL values can therefore
+also differ slightly from the previous low-precision calculation. The log-space
+reverse-KL and Jensen-Shannon formulas prevent finite losses with non-finite
+gradients; the upcast also preserves small losses that would underflow to zero.
+Non-finite logits propagate into the loss/gradients, allowing AMP's GradScaler to
+skip overflowed steps rather than aborting training with a validation exception.
+
+FP32 intermediates cost time and memory. The [maintainer's PR #736 measurement](https://github.com/MakazhanAlpamys/Soup/pull/736)
+on an RTX 5070 (BF16, B=1/S=512/V=32000, forward plus backward) found the originally
+submitted kernel took 1.71 times as long and 1.24 times the peak memory of main.
+That measurement included finite-value guards removed here: without them it took
+9.40 ms versus 6.39 ms, with the same 376.5 MiB versus 303.6 MiB peak. These are
+hardware-specific measurements, not a benchmark of this revised implementation,
+which also avoids unused probability tensors. Two FP32 logits copies alone occupy
+about 7.8 GiB at B=4/S=2048/V=128256; budget for additional intermediates as well.
 
 Set `distill_mode: sequence` (default `token`) to train on the teacher's **generated
 continuations** instead of per-token logit matching — a hard-label, cross-tokenizer-friendly
--- a/docs/training.md
+++ b/docs/training.md
@@ -425,6 +425,10 @@ Loss = student CE + (T**2) × KL(teacher_logits / T  ||  student_logits / T).
 Teacher is loaded once, frozen via `requires_grad_(False)` + `.eval()`, and its
 inputs / logits are auto-bridged across CPU / CUDA devices.
 
+The token-divergence kernel evaluates probability math in FP32, including when
+the model logits use FP16 or BF16, so low temperatures do not create non-finite
+reverse-KL or Jensen-Shannon gradients.
+
 Set `distill_mode: sequence` (default `token`) to train on the teacher's **generated
 continuations** instead of per-token logit matching — a hard-label, cross-tokenizer-friendly
 KD that works when student and teacher do not share a vocabulary. `sequence` mode is mutually
--- a/src/soup_cli/trainer/distill.py
+++ b/src/soup_cli/trainer/distill.py
@@ -96,9 +96,20 @@ def _compute_distill_term(
         if attention_mask is not None:
             attention_mask = attention_mask[:, 1:]
 
+    if not torch.isfinite(student_logits).all():
+        raise ValueError("student_logits must contain only finite values")
+    if not torch.isfinite(teacher_logits).all():
+        raise ValueError("teacher_logits must contain only finite values")
+
+    # Keep the divergence kernel in FP32. In lower precision, valid logits at
+    # low temperatures readily underflow probabilities to zero; target-side
+    # derivatives in torch.kl_div then become non-finite even when the reduced
+    # loss is finite (#719).
     temp = float(temperature)
-    s = student_logits / temp
-    t = teacher_logits / temp
+    log_s = torch.log_softmax(student_logits.float() / temp, dim=-1)
+    log_t = torch.log_softmax(teacher_logits.float() / temp, dim=-1)
+    p_s = log_s.exp()
+    p_t = log_t.exp()
 
     def _masked_mean(per_token: "_torch_typ.Tensor") -> "_torch_typ.Tensor":
         """Mean of a ``(batch, seq)`` per-token divergence over trained tokens."""
@@ -112,29 +123,16 @@ def _compute_distill_term(
         denom = mask.sum().clamp(min=1.0)
         return (per_token * mask).sum() / denom
 
-    kl_div = torch.nn.functional.kl_div
     if divergence == "forward_kl":
-        # KL(teacher || student): student log-probs, teacher probs. reduction=
-        # "none" keeps per-token so we can mask before averaging.
-        log_s = torch.log_softmax(s, dim=-1)
-        p_t = torch.softmax(t, dim=-1)
-        per_token = kl_div(log_s, p_t, reduction="none").sum(dim=-1)
+        per_token = (p_t * (log_t - log_s)).sum(dim=-1)
         return _masked_mean(per_token) * (temp * temp)
     if divergence == "reverse_kl":
-        log_t = torch.log_softmax(t, dim=-1)
-        p_s = torch.softmax(s, dim=-1)
-        per_token = kl_div(log_t, p_s, reduction="none").sum(dim=-1)
+        per_token = (p_s * (log_s - log_t)).sum(dim=-1)
         return _masked_mean(per_token) * (temp * temp)
     if divergence == "js":
-        # Jensen-Shannon: 0.5 (KL(p||m) + KL(q||m)), m = 0.5 (p + q).
-        log_s = torch.log_softmax(s, dim=-1)
-        log_t = torch.log_softmax(t, dim=-1)
-        p_s = log_s.exp()
-        p_t = log_t.exp()
-        m = 0.5 * (p_s + p_t)
-        log_m = m.clamp(min=1e-12).log()
-        kl_pm = kl_div(log_m, p_s, reduction="none").sum(dim=-1)
-        kl_qm = kl_div(log_m, p_t, reduction="none").sum(dim=-1)
+        log_m = torch.logaddexp(log_s, log_t) - math.log(2.0)
+        kl_pm = (p_s * (log_s - log_m)).sum(dim=-1)
+        kl_qm = (p_t * (log_t - log_m)).sum(dim=-1)
         return 0.5 * (_masked_mean(kl_pm) + _masked_mean(kl_qm)) * (temp * temp)
     raise ValueError(f"Unknown divergence {divergence!r}")
 
--- a/src/soup_cli/trainer/distill.py
+++ b/src/soup_cli/trainer/distill.py
@@ -64,6 +64,10 @@ def _compute_distill_term(
     Averaging over padding/prompt tokens (the pre-fix behaviour) diluted the
     signal with the divergence on positions the student is not trained on.
 
+    The scalar stays in FP32 for every divergence, including half-precision
+    inputs. Non-finite logits propagate through the loss so AMP can skip an
+    overflowed step without a synchronous per-step validation guard.
+
     Raises:
         TypeError: ``temperature`` not numeric or is bool.
         ValueError: ``temperature`` non-finite or non-positive; ``divergence``
@@ -96,11 +100,6 @@ def _compute_distill_term(
         if attention_mask is not None:
             attention_mask = attention_mask[:, 1:]
 
-    if not torch.isfinite(student_logits).all():
-        raise ValueError("student_logits must contain only finite values")
-    if not torch.isfinite(teacher_logits).all():
-        raise ValueError("teacher_logits must contain only finite values")
-
     # Keep the divergence kernel in FP32. In lower precision, valid logits at
     # low temperatures readily underflow probabilities to zero; target-side
     # derivatives in torch.kl_div then become non-finite even when the reduced
@@ -108,8 +107,6 @@ def _compute_distill_term(
     temp = float(temperature)
     log_s = torch.log_softmax(student_logits.float() / temp, dim=-1)
     log_t = torch.log_softmax(teacher_logits.float() / temp, dim=-1)
-    p_s = log_s.exp()
-    p_t = log_t.exp()
 
     def _masked_mean(per_token: "_torch_typ.Tensor") -> "_torch_typ.Tensor":
         """Mean of a ``(batch, seq)`` per-token divergence over trained tokens."""
@@ -124,12 +121,16 @@ def _compute_distill_term(
         return (per_token * mask).sum() / denom
 
     if divergence == "forward_kl":
+        p_t = log_t.exp()
         per_token = (p_t * (log_t - log_s)).sum(dim=-1)
         return _masked_mean(per_token) * (temp * temp)
     if divergence == "reverse_kl":
+        p_s = log_s.exp()
         per_token = (p_s * (log_s - log_t)).sum(dim=-1)
         return _masked_mean(per_token) * (temp * temp)
     if divergence == "js":
+        p_s = log_s.exp()
+        p_t = log_t.exp()
         log_m = torch.logaddexp(log_s, log_t) - math.log(2.0)
         kl_pm = (p_s * (log_s - log_m)).sum(dim=-1)
         kl_qm = (p_t * (log_t - log_m)).sum(dim=-1)
--- a/tests/test_issue719_stable_distill_divergence.py
+++ b/tests/test_issue719_stable_distill_divergence.py
@@ -58,14 +58,48 @@ def test_forward_kl_matches_probability_space_reference() -> None:
 
 
 @pytest.mark.parametrize("which", ["student", "teacher"])
-def test_non_finite_input_is_reported_separately(which: str) -> None:
+@pytest.mark.parametrize("divergence", ["forward_kl", "reverse_kl", "js"])
+@pytest.mark.parametrize("nonfinite", [float("nan"), float("inf")])
+def test_non_finite_logits_propagate_to_amp(which, divergence, nonfinite) -> None:
     torch = pytest.importorskip("torch")
     from soup_cli.trainer.distill import _compute_distill_term
 
-    student = torch.zeros(1, 1, 2)
+    student = torch.zeros(1, 1, 2, requires_grad=True)
     teacher = torch.zeros(1, 1, 2)
-    target = student if which == "student" else teacher
-    target[0, 0, 0] = float("nan")
+    with torch.no_grad():
+        (student if which == "student" else teacher)[0, 0, 0] = nonfinite
 
-    with pytest.raises(ValueError, match=rf"{which}_logits.*finite"):
-        _compute_distill_term(student, teacher, "reverse_kl", temperature=1.0)
+    loss = _compute_distill_term(student, teacher, divergence, temperature=1.0)
+    loss.backward()
+
+    assert not torch.isfinite(loss)
+    assert not torch.isfinite(student.grad).all()
+
+
+@pytest.mark.parametrize("dtype_name", ["float32", "bfloat16", "float16"])
+@pytest.mark.parametrize("divergence", ["forward_kl", "reverse_kl", "js"])
+def test_divergence_matches_double_precision_reference(dtype_name, divergence) -> None:
+    torch = pytest.importorskip("torch")
+    from soup_cli.trainer.distill import _compute_distill_term
+
+    dtype = getattr(torch, dtype_name)
+    student = torch.tensor([[[0.2, -0.4, 0.8]]], dtype=dtype, requires_grad=True)
+    teacher = torch.tensor([[[0.5, 0.1, -0.2]]], dtype=dtype)
+    temperature = 2.0
+    # Use independent probability-space arithmetic on the actual rounded inputs.
+    ps = torch.softmax(student.detach().double() / temperature, dim=-1)
+    pt = torch.softmax(teacher.double() / temperature, dim=-1)
+    if divergence == "forward_kl":
+        expected = (pt * torch.log(pt / ps)).sum()
+    elif divergence == "reverse_kl":
+        expected = (ps * torch.log(ps / pt)).sum()
+    else:
+        mixture = (ps + pt) / 2
+        expected = ((ps * torch.log(ps / mixture)).sum()
+                    + (pt * torch.log(pt / mixture)).sum()) / 2
+    expected *= temperature**2
+
+    actual = _compute_distill_term(student, teacher, divergence, temperature=temperature)
+
+    assert actual.dtype == torch.float32
+    torch.testing.assert_close(actual.double(), expected, rtol=1e-5, atol=2e-7)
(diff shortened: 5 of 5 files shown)


MakazhanAlpamys:
@be-student — welcome, and thank you for two substantive pull requests. **The bug is real, severe, and your fix closes it. I am blocking on one thing you could not have measured from CI: the fix costs 3.1x wall clock and +71% peak VRAM on the `forward_kl` path, which is the default and never had the bug.**

Everything below is on your head `2febcee5`, in a worktree with `PYTHONPATH` pinned inside it, against clean `main` as a baseline.

## Your diagnosis is exactly right, and I want to say so first

Real `_compute_distill_term`, real `backward()`, `.grad` inspected — not a proxy:

```
=== clean main ===
dtype     div          loss                     grad@sup            finite
float32   forward_kl   1.237086417e-05  [2.2677e-06, -2.2698e-06]   True
float32   reverse_kl   1.135442602e-07              [nan, nan]      False
float32   js           3.933490689e-08              [nan, nan]      False
bfloat16  reverse_kl                 0              [nan, nan]      False
float16   reverse_kl                 0              [nan, nan]      False
=== your branch ===
float32   reverse_kl   1.135442602e-07                 [0.0, 0.0]   True
bfloat16  reverse_kl   1.135442602e-07                 [0.0, 0.0]   True
float16   js           3.933684667e-08                 [0.0, 0.0]   True    (9 of 9 finite)
```

A finite loss with a NaN gradient, in all three dtypes. That is the worst class of numerical bug there is, because nothing in a training log shows it. You also fixed something #719 does not mention: on `main`, bf16 and fp16 `reverse_kl` return a loss of **exactly 0** — total underflow.

**And your tests assert on gradients**, not on `torch.isfinite(loss)`. That distinction is the whole issue, and most first submissions get it wrong.

The normal regime is preserved — fp32, seeded logits: `forward_kl` T=1 identical to `main`, T=2 differs by 7e-8; `reverse_kl` and `js` at T=1 identical to 12 digits. Non-finite input raises a clear named `ValueError`, not a bare assert. Zero deleted test lines. Ruff clean.

## BLOCKING — the default path pays for a bug it does not have

Measured on a real GPU (RTX 3050, fp16 logits, 2 x 512 x 32000, best of 10 after 10 warmups):

| divergence | main | your branch | | main peak | your peak |
|---|---|---|---|---|---|
| **forward_kl (default)** | 23.42 ms | **73.56 ms (3.14x)** | | 436.6 MiB | **748.5 MiB (+71%)** |
| reverse_kl | 22.85 ms | 74.67 ms (3.27x) | | 436.6 MiB | 748.5 MiB (+71%) |
| js | 45.89 ms | 99.47 ms (2.17x) | | 748.5 MiB | 998.1 MiB (+33%) |

`forward_kl` is what `tcfg.distill_divergence or "forward_kl"` resolves to when the user sets nothing, and it was never affected by #719.

I decomposed it so the fix is targeted rather than a rewrite:

```
main                          23.42 ms
your branch minus the guards  55.84 ms   <- the two isfinite guards cost ~20 ms
your branch minus the upcast  43.49 ms   <- the upcast costs ~32 ms
your branch as shipped        75.97 ms
```

**The two `torch.isfinite(...).all()` guards alone add ~20 ms — about 86% of main's entire kernel time — and on GPU each one forces a device-to-host sync every step.** That is the cheapest thing to fix: make the guard opt-in, or run it only for the first N steps, where it still catches a genuinely broken teacher.

The upcast cost is inherent and I am not asking you to remove it. But `p_s` and `p_t` are both materialised unconditionally, when `forward_kl` needs only `p_t` and `reverse_kl` only `p_s` — six fp32 vocab-sized tensors where main held two. At a realistic `(4, 2048, 128256)` that is 3.91 GiB of transient becoming 23.48 GiB, which is the difference between a run fitting on a card and not.

## Also blocking — two mutations survive, and one is the only new math in the PR

Restored and verified byte-identical after each round.

| mutation | your tests | full 429-test distill selection |
|---|---|---|
| `reverse_kl` back to probability-space `kl_div` | 6 failed, killed | — |
| `js` back to probability-space with a 1e-12 clamp | 6 failed, killed | — |
| `forward_kl` swapped for the `reverse_kl` formula | 1 failed, killed | — |
| `T^2` scaling changed to `T` | 1 failed, killed | — |
| remove both isfinite guards | 2 failed, killed | — |
| **drop the `.float()` upcast** | 15 passed, **SURVIVED** | 429 passed, **SURVIVED** |
| **`math.log(2.0)` becomes `math.log(3.0)` in JS** | 15 passed, **SURVIVED** | 429 passed, **SURVIVED** |

**The JS one is the finding.** That constant is the only new mathematics this PR introduces, and it is verified by nothing. Changing it moves JS by **1.9x** on ordinary input (0.4389 -> 0.8444) and by **25,770x** on your own reproducer (3.93e-08 -> 1.01e-03), and every test still passes. Twelve `isfinite`-only assertions cannot see a loss that is four orders of magnitude wrong.

Dropping the upcast is the same shape: it changes fp16 results by 2.2e-4 to 7.8e-4 relative and reverts the return dtype, undetected — and the upcast is what your title, your prose and your code comment all lead with. Your three-dtype parametrization exists precisely to cover it; the log-space form alone happens to stay finite, so the parametrization never gets to fail.

I checked the shipped code is correct (your JS matches `main` to 12 digits in the normal regime), so this is a hole a future edit falls into, not a live defect. **One value-equivalence test each for `reverse_kl` and `js` against a hand-computed reference closes both** — mirroring the `forward_kl` one you already wrote, which is genuinely discriminating (two mutations die to it).

## House conventions, not defects

These are things about how this repo works that you had no way to know. Neither is a criticism of the code.

- **Changelog fragments, not `CHANGELOG.md`.** You edited `CHANGELOG.md` directly under `[Unreleased]`. Since #487 this repo uses per-PR fragments: add `changelog.d/0.74.0/736.fixed.md` — the **PR** number, not the issue — and credit yourself with `(#719 by @be-student in #736)`. Revert the `CHANGELOG.md` edit. Worth knowing that `test_issue487_changelog_fragments.py` passed anyway: it asserts the directory is *empty at release*, so it cannot detect a fragment that was never written. Nothing would have told you.
- **The return dtype changes** from `float16` to `float32` with half-precision logits. Harmless where it is consumed (`total = w*ce + w*distill`), but it is a contract change and belongs in the body and the fragment.

## To unblock

1. Make the isfinite guards cheap — opt-in, or first-N-steps only.
2. Materialise `p_s` / `p_t` only where the chosen divergence needs them, so `forward_kl` stops paying for `reverse_kl`'s tensors.
3. Two value-equivalence tests: `reverse_kl` and `js` against hand-computed references.
4. Move the changelog entry to a fragment.

## What I could not verify

Your body says "20,409 tests passed" and "coverage 82.99%". A fully green suite is not achievable on this box — issue #651 means `trl 0.29.1` cannot import at the declared `torch>=2.5.0` floor, and clean `main` alone shows 24 failures in the PPO selection. Your counts are *self-consistent* with each other and with the 147-test gap explained by the six commits `main` gained after your merge base, so I have no reason to doubt them; I am recording that I did not confirm them rather than implying I did. I did not check the coverage figure at all.

The engineering here is strong and the diagnosis was the hard part. The perf cost is the kind of thing only a GPU measurement shows, which is why I ran one rather than sending this back on a reading.

MakazhanAlpamys:
@be-student — one convention item on both this and #737, flagged now so you can fix it while I review the substance. Not a comment on the code, which I have not finished reading yet.

**This repo does not edit `CHANGELOG.md` directly.** Since #487 a user-visible change adds a fragment instead:

```
changelog.d/0.74.0/736.fixed.md
```

Three rules, all in `changelog.d/README.md`:

1. **Named after the pull request, not the issue.** Your entry cites `(#719)`; the filename and the in-text reference both want `#736`. The reason is exactly the situation this repo has been in all week — two pull requests on one issue collide on one filename, and the PR number is what keeps them distinct.
2. **Credit the author**: `(#719 by @be-student in #736)`. Outside work gets its author named; that field is the point of the file.
3. **Write the complete list item** in the fragment. Your prose is already the right shape — it just moves.

Same for #737 (`737.fixed.md`).

Two things worth knowing so this does not look arbitrary:

- **CI cannot catch it.** `tests/test_issue487_changelog_fragments.py` validates fragments that exist and has no way to see a missing one, and nothing rejects a direct `CHANGELOG.md` edit. Your 14/14 green is honest; the check simply is not there.
- **Your `[Unreleased]` text is not wasted.** Entries written there before v0.73.4 are still assembled alongside fragments, so the content is fine — it is the location that is wrong for new work. Move it and drop the `## [Unreleased]` / `### Fixed` scaffolding; `scripts/assemble_changelog.py` adds those.

Substance review on both is next in my queue. Standard battery here, so you know what is coming: I mutate the main fix, and a test that survives the mutation is a blocking finding regardless of how many pass.

MakazhanAlpamys:
@be-student — checking in, and consolidating so you are not tracking three of my threads at once.

**#736 (this one).** The blocker is unchanged and it is not about correctness: your fix closes a real NaN-gradient defect, and it costs **3.1x wall clock and +71% peak VRAM on the `forward_kl` path** — which is the default and never had the bug. What I need is for the FP32 log-space computation to apply to the paths that need it rather than to all of them, or a measurement showing my numbers are wrong. Plus the convention item: `changelog.d/0.74.0/736.fixed.md` rather than an edit to `CHANGELOG.md`.

**#737** is the one I chose over @dchaudhari7177's #741, on test strength — yours is the only one of the two that pins the reverse epoch conflation. It is blocked on three items listed in that thread, one of which (`pytest.importorskip("trl")` guarding the wrong module) is a genuine red test at the declared torch floor.

**#730** is not waiting on you and I want that stated plainly: I told you to push a one-liner, then found it does not fix the defect I had asked it to fix. That is my error, the correction and the working approach are in that thread, and you lost time to it.

No deadline on any of the three. If the paperwork items are the part you would rather not spend an evening on, say so and I will write the fragments myself and merge with the work credited to you.

MakazhanAlpamys:
@be-student — no new ask, just so this thread carries the same picture as the other two.

The blocker here is unchanged and it is **not** a correctness question: your fix closes a real NaN-gradient defect and I have not disputed a line of it. What is unresolved is the cost — 3.1x wall clock and +71% peak — and whether that is paid unconditionally or behind a flag. That is a design decision I owe you an answer on as much as you owe me a change.

Ordering, so you are not guessing: **#737 first** (it is the only one of the three with another contributor queued behind it, and it now has a 2026-09-11 18:00 UTC date), **#730 second** (same date, and the last round of it was my error rather than yours), **this one third**. Nothing here expires.

MakazhanAlpamys:
@be-student — **you are owed the decision I said I owed you, and my earlier cost number was measured on hardware this project no longer runs on. Re-measured, the answer changes in your favour.**

## The correction first

My "3.1x wall clock and +71% peak" came from the old dev box. This repository's dev machine changed, and on the current card (RTX 5070, bf16, B=1 S=512 V=32000, forward+backward, 10 iterations after warm-up) the numbers are:

```
main                       6.39 ms   303.6 MiB
this PR as submitted      10.91 ms   376.5 MiB   (1.71x time, 1.24x peak)
this PR minus isfinite     9.40 ms   376.5 MiB
log-space, no fp32 upcast  4.02 ms   224.0 MiB   (faster and smaller than main)
```

**The stability fix itself is free.** It is not merely cheap — log-space without the upcast beats `main` on both axes. Everything I was worried about is the fp32 upcast plus the two `isfinite` guards, and I was attributing it to the wrong part of your change.

## The decision

**Keep the upcast. Drop the guards. Document the cost.** Reasoning, so you can disagree with it rather than just comply:

I tested log-space without the upcast, and while gradients stay finite in every dtype, reverse-KL loss **flushes to exactly `0.0`** in bf16/fp16 at T=0.05, against `1.1e-7` with fp32. So the upcast is preserving real signal the low-precision path would silently lose, and losing signal silently is worse than the memory. It stays, unconditionally, no flag.

The `isfinite` guards are the part I want removed, and for a behavioural reason rather than the ~1.5 ms and the per-step device-to-host sync. With an `inf` in the logits, `main` returns `nan` and `GradScaler` skips the step — which is precisely what `GradScaler` exists to do — whereas this PR raises `ValueError` and kills the run. Under fp16 AMP a transient overflow is expected, not exceptional, so raising by default converts a routine recoverable event into a lost training run. I recognise #719 asked you to "treat non-finite input logits separately rather than hiding them", so this is a judgement call and you followed the instruction as written; my read is that the instruction was wrong on this point. Remove them, or put them behind a debug flag that defaults off.

## Three things still needed

**1. Changelog convention.** This repo moved to per-PR fragments in #487; `CHANGELOG.md` is no longer edited directly. Please move your entry to `changelog.d/0.74.0/736.fixed.md`, first line ending `(#719 by @be-student)`. See `changelog.d/0.74.0/738.fixed.md` for house style. Worth knowing: `tests/test_issue487_changelog_fragments.py` passes either way, so CI will not flag this — and editing `CHANGELOG.md` also parks you in a hot conflict file.

**2. Document the fp32 cost** in the fragment and in `docs/training.md`: 1.71x time and 1.24x peak here, which at Llama-3 vocab with B=4/S=2048 is roughly +7.8 GiB of logits copies. Since you are keeping it, the trade-off should be visible rather than discovered.

**3. `forward_kl` did change, so please correct the claim.** The PR and CHANGELOG say only reverse-KL and JS moved. Measured relative change on `forward_kl`: **3.9e-3 (bf16)**, **1.9e-3 (fp16)**; fp32 is unchanged at 2.4e-7. Your forward-KL test is fp32-only, so it cannot see this. The return dtype also moves bf16 -> fp32 for all three divergences, where #719 said "cast only the final scalar as appropriate" — either cast back with `.to(student_logits.dtype)` or state that fp32 return is deliberate.

## For the record, the fix is verified

I reverted the kernel to `main`'s probability-space form while keeping your `isfinite` guard, to isolate the numerical change: **12 failed, 3 passed**, and the failure output is the reported bug verbatim — finite loss, `student.grad = [[0,0],[nan,nan],[0,0]]`. Your test calls `.backward()` and inspects `student.grad`, which is what #719 actually asks for and what a loss-finiteness assertion would have missed. 15 passed on your file, 339 passed across the distill suite (the one failure is a pre-existing Windows symlink-privilege test that fails identically on `main`), ruff clean, `docs/training.md` updated, zero deleted test lines.

## Ordering, unchanged except for one addition

#737 first, #730 second, this one third. New: **#743 (chunked KL, #722) rewrites the same hunk** and conflicts with this PR in both `distill.py` and `docs/training.md`. Its `_chunk_kernel` still uses `kl_div(log_t, p_s)` and the `1e-12` clamp, so resolving that conflict in its favour would reinstate exactly this bug — I have told @kok-o that yours lands first and that they should port your log-space formulation into the chunked kernel, where `(p_s * (log_s - log_t)).sum()` drops straight into their `reduction="sum"` structure. Nothing here expires.

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:
@be-student — reviewed at `9d2e4e0e`. **The cost blocker is resolved**, and I measured it myself rather than taking your numbers. Two mechanical changes remain, both created by the release landing under you.

## The blocker, measured

RTX 5070, torch 2.14.0+cu130, bf16, B=1 S=512 V=32000, forward **and** backward with a labels mask, best-of-20 after 10 warm-up, two independent runs:

| divergence | | ×main time | ×main peak |
|---|---|---|---|
| **forward_kl** (the default path) | previous head | 2.14x | 1.47x |
| | **this head** | **1.72x** | **1.27x** |
| reverse_kl | previous head | 1.69x | 1.40x |
| | **this head** | **1.39x** | **1.22x** |

Roughly half the regression I blocked on is gone, and for exactly the two reasons I named: dropping the guards (~1.5 ms) and not materialising the probability tensor the divergence does not use (~62 MiB). I also read the head kernel to confirm no device-to-host sync remains — no `.item()`, no `bool()` of a tensor, and `denom` stays on device.

`js` I will not quote tightly: its `main` baseline varied 10.7–12.1 ms across runs here, so anything better than "~1.2–1.4x" would be false precision. Its peak is unchanged because JS genuinely needs both probability tensors.

**One honest caveat about my own earlier number.** My 09-10 baseline for `main` was 6.39 ms where this run measures 3.69 ms on the same card. I did not chase the difference. The ratios above are internally consistent — all three implementations measured in one process, reproduced across two runs to ~1% on `forward_kl` and `reverse_kl` — but treat my earlier "3.1x" as measured on a different footing rather than directly comparable.

## The numerics are better than I expected

Both mutations that survived my last round now die: dropping the `.float()` upcast fails 6 tests, and `log(2.0)` → `log(3.0)` in JS fails 3. Fifteen mutations applied, fourteen killed.

Against an independent float64 reference computed on the *rounded* inputs, in the normal regime:

| dtype | divergence | `main` rel err | this PR |
|---|---|---|---|
| bfloat16 | reverse_kl | **8.8e-03** | **4.9e-08** |
| bfloat16 | forward_kl | 6.2e-04 | 3.2e-08 |
| float16 | forward_kl | 1.4e-03 | 2.8e-07 |

**Four to five orders of magnitude more accurate than `main` in every low-precision case.** That settles the thing I raised on 09-10: `forward_kl` did change, and the change is `main` being wrong, not you.

And the #719 reproducer behaves exactly as claimed — on `main`, reverse-KL and JS produce non-finite **gradients** in all three dtypes, with reverse-KL underflowing to exactly `0` in bf16 and fp16; on your head the loss is `1.135e-07` and the gradient is finite. fp32 gradient equivalence against `main` in the normal regime is round-off (2–3e-07 relative).

No test was weakened. `test_non_finite_input_is_reported_separately` was *replaced*, not deleted, by a 12-case test asserting the loss **and** the gradient go non-finite — the behaviour reversal I asked for, specified more tightly than the old contract. 15 → 34 tests, and mutation coverage went up.

## Two mechanical changes

**1. Drop the `CHANGELOG.md` edit entirely and rebase.** Your second commit deletes the `[Unreleased]` block; v0.75.0 assembled that region an hour ago, so `git merge-tree` reports a content conflict. The fragment already carries the text:

```
git checkout origin/main -- CHANGELOG.md
```

`CHANGELOG.md` then drops out of your diff completely, which is the correct end state.

**2. Move the fragment.** Verified by execution after rebasing: `test_repository_fragments_match_the_newest_release` fails with exactly the predicted message until `git mv changelog.d/0.74.0/736.fixed.md changelog.d/0.75.0/736.fixed.md`. Then 584 passed.

## One number to correct, and one thing only you can fix

`docs/training.md` and the fragment both cite **1.47x**. That was my measurement of *the previous head minus its guards* — not of what you are shipping. The shipped head measures **1.72x time / 1.27x peak** on the default path. Your text is honestly hedged ("not a benchmark of this revised implementation"), but a reader takes the figure. Use the numbers above, or give a range.

**Your PR description is now actively wrong.** It still says "Non-finite input logits now fail explicitly" — the behaviour this head *removed*, at my request — and still frames the change as reverse-KL and JS only, which your own docs correctly contradict. I write the squash body myself so the commit message will be accurate either way, but the description is what anyone reads on the way in.

## One follow-up I will file, not a merge condition

With the guards gone, a *genuinely* broken teacher — persistent `inf`/`nan`, not a transient AMP overflow — now yields `nan` every step, `GradScaler` skips every step, and the run finishes having learned nothing, silently. Previously it raised. That is the trade I chose over #719's instruction and I still think it is the right default; a warn-once diagnostic after N consecutive non-finite terms would close the gap without reintroducing a per-step sync.

## Sequencing with #743

Confirmed by `git merge-tree`: exactly two files collide with @kok-o's #743 — the divergence dispatch in `trainer/distill.py` and a `docs/training.md` section, the latter being non-overlapping content that just landed in the same place.

**This PR goes first**, and the reason is not politeness: #743's `_chunk_kernel` still contains `kl_div(log_t, p_s, reduction="sum")` for reverse-KL — the #719 defect verbatim — plus the JS `1e-12` clamp and a `return loss.to(dtype=s.dtype)` that reverts your FP32 return contract. Merging #743 first would silently reinstate the bug this PR exists to fix.

There is also a trap worth recording for whoever resolves it: the prelude **auto-merges silently**, keeping your `.float()` upcast while dropping the `s`/`t` locals and the `_masked_mean` closure that #743's side references. Taking #743's side verbatim gives a `NameError`. Their chunking has to be ported onto your `log_s`/`log_t`, not pasted. I am telling @kok-o this on their thread.

`main` is frozen until v0.75.0 is tagged, so the rebase can wait for the tag. Ping me when it is pushed.

MakazhanAlpamys:
@be-student — reviewed at `9d2e4e0e`. **The numerics are right and the "preserve AMP overflow recovery" commit is exactly what I asked for.** Two blocking items, one of which is serious and which PR CI cannot see.

## What I measured

#719's reproducer at `temperature=0.05`, on `main + your PR`:

| divergence | loss | grads finite |
|---|---|---|
| forward_kl | finite | True |
| reverse_kl | finite | **True** |
| js | finite | **True** |

and fp32 scalar out in fp32 / bf16 / fp16. The same reproducer on `main` alone gives `nan` and non-finite gradients for all three. `p_s = log_s.exp(); (p_s * (log_s - log_t)).sum(-1)` and `log_m = logaddexp(log_s, log_t) - log(2)` are the right formulations, and keeping the whole thing in log space is what makes the low-temperature case survive.

**Three mutations, all caught** by `tests/test_issue719_stable_distill_divergence.py` (34 tests):

| mutation | result |
|---|---|
| revert `reverse_kl` + `js` to the pre-PR probability-space `kl_div` form | **12 failed** |
| drop the `.float()` FP32 upcast on both logit tensors | **6 failed** |
| JS mixture constant `math.log(2.0)` -> `math.log(3.0)` | **3 failed** |

No surviving mutations, and the third one matters most: it means the constant is pinned rather than merely present. `ruff check` clean. **544 passed, 1 skipped** across the eleven distill/changelog files on your head; **557 passed, 1 skipped** on `main` + your branch.

The latest commit is correct on its own terms too: removing the per-step `torch.isfinite` guards avoids a device sync, moving `log_s.exp()` / `log_t.exp()` into the per-divergence branches stops materialising probability tensors nobody uses, and replacing the raising test with a propagation test is the honest way to record the decision.

## 1. BLOCKING — this PR deletes 434 lines of `CHANGELOG.md`, including the entire v0.75.0 section

This is the one to look at first. `git diff origin/main -- CHANGELOG.md` on your head:

```
@@ -10,434 +10,6 @@
-## [Unreleased]
-
-## [0.75.0] - 2026-09-12
-
-### Added
-  ... the whole release section ...
```

and `grep -c "^## \[Unreleased\]" CHANGELOG.md` returns **0**.

Almost certainly a merge resolution that took your older side of the file wholesale — an easy mistake and not a judgement about your work. But the consequences are bad in a specific way:

- It reverts the published v0.75.0 changelog section.
- It deletes the `## [Unreleased]` heading, and `scripts/assemble_changelog.py` requires it. I ran it both ways: without the heading `error: CHANGELOG.md: missing '## [Unreleased]' heading`; with it, `Assembled 1 changelog fragment(s).` So this does not break your CI — **it breaks the next release**, silently, weeks from now.
- `CHANGELOG.md` is assembled by the release process from `changelog.d/` fragments. A PR should never touch it at all; your fragment is the correct and sufficient channel.

Fix: `git checkout origin/main -- CHANGELOG.md`. That also removes the only merge conflict this PR has — `src/soup_cli/trainer/distill.py` and `docs/training.md` already auto-merge cleanly against current `main`.

## 2. BLOCKING — the fragment has to move

```
git mv changelog.d/0.74.0/736.fixed.md changelog.d/0.75.0/736.fixed.md
```

Reproduced on a real merge: `test_repository_fragments_match_the_newest_release` fails with `fragment baseline is 0.74.0, but the latest release is 0.75.0`. Note your **current green CI is stale** — run `34675625329` finished at 05:27 UTC, before v0.75.0 merged at 08:26 — so a re-run goes red on this regardless.

## Two things that are not change requests

**#887 is my error, not yours.** I filed it saying a persistently broken teacher "now trains silently instead of raising". Checking `main` directly, the only `isfinite` call in `distill.py` is the `temperature` guard at line 80 — the logits guard never reached `main`; it existed only between your first and second commits. Your net effect versus `main` is zero on that axis. I am editing the issue rather than leaving you to defend against a regression that does not exist.

**The fifth suggested fix in #719** ("treat already non-finite input logits separately") stays deliberately undone, per my earlier review. That is my call, not an omission of yours.

## Ordering, so you know why this matters beyond your PR

**This merges before #743.** #743's `_chunk_kernel` contains the pre-fix probability-space form verbatim, so in the other order `main + #743` merges with *zero conflicts* and silently restores #719. I verified that. Your PR going in first is what makes the defect in the other one visible. There is nothing for you to do about it — I am sequencing them — but it does mean item 1 and item 2 are the last things between this and a merge.

Push those two and I will merge.

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:
Merged as `8e86f7f0`. Thank you — and the part I want to name is that **you reworked this twice without arguing either time**, including a reversal where I asked you to undo something you had already written.

**What I verified before merging, on the current head `de38ca31` (the merge commit that cleared both of Friday's blockers — CI run `34749845622` is on exactly that SHA, 14/14, so the green mark is not stale):**

- 34 passed in the regression file; **848 passed, 6 skipped** across twelve distill/ULD/changelog suites; **366 passed** on the tree merged with current `main`, including #742's grad-accum fix, which lands nowhere near this kernel.
- Zero removed assertions: `git diff origin/main...HEAD -- tests/ | grep '^-'` is empty.
- `CHANGELOG.md` is byte-identical to `main` (`git rev-parse HEAD:CHANGELOG.md` matches), and `## [Unreleased]` is back at count 1. The fragment is in `changelog.d/0.75.0/`.

**Four mutations, each anchor counted once and each mutated module actually imported before running the tests** — an `import` rather than only `ast.parse`, because a `NameError` from a sloppy mutation produces a false "caught":

| mutation | result |
|---|---|
| drop the `.float()` upcast on both logit tensors | **6 failed** |
| restore `kl_div(log_t, p_s)` | **6 failed** — reproduces #719's bug verbatim: `isfinite(loss)` passes, `isfinite(student.grad).all()` is `False` |
| restore the clamped JS mixture | **6 failed** |
| `log(2)` -> `log(3)` | **3 failed** — `Expected 0.05504839 but got 1.67690920` |

That last one is **the mutation that survived the 09-07 round**, and it is now a pure numerical kill against the fp64 reference. That is the difference between a test that describes the fix and a test that holds it.

I also checked the upcast is pinned by arithmetic rather than by the `dtype ==` contract assertion alone — it is, by four orders of magnitude in all six bf16/fp16 cells (`absdiff` 0.0231 against a `tol` of 2.35e-06 for bf16 forward_kl, and similar across the rest).

**On the substance:** what convinced me this is a reformulation and not NaN-masking is that one clamp was removed and none added, and that the removed one was measurably the biased term — identical distributions with a `-40` logit give the old kernel `-5.25e-17`, a *negative* divergence, against `0.0` here. Objective preservation held to fp32 round-off against the previous kernel and to 1e-08 against a closed-form fp64 reference, so it trains the same objective, computed correctly. Where it differs in bf16/fp16, the old one is the wrong one.

**Two things I fixed in the squash body rather than asking you for another round:** the docs and fragment still quote 1.71x/1.24x, which is my measurement of the *originally submitted* kernel rather than the shipped one, and the PR description still says "Non-finite input logits now fail explicitly" — the behaviour you removed at my request. Both are honestly hedged in the text; neither was worth a fourth round.

Your CONTRIBUTORS.md entry is updated. #719 is closed.

One sequencing note for anyone reading this later: #743's `_chunk_kernel` still carries `kl_div(log_t, p_s)` and the `1e-12` clamp, so it needs rebasing onto this.

PR #737, opened 2026-09-06
fix(ppo): forward configured schedule

### LINKED ISSUE #721: PPO drops the configured KL coefficient and epoch budget on TRL 0.29
## Summary

`PPOTrainerWrapper.setup()` loses two user-configured values when using the
built-in TRL 0.29 PPO trainer:

1. `training.ppo_kl_penalty` is forwarded only when `PPOConfig` exposes the old
   name `init_kl_coef`, while TRL 0.29 exposes `kl_coef`.
2. `training.epochs` is saved for Soup's legacy manual loop but is not passed as
   `num_train_epochs` to the built-in `PPOConfig` path.

`training.ppo_epochs` is a different setting and is correctly forwarded to
`num_ppo_epochs`; it controls PPO optimization passes, not dataset epochs.

## Reproduction

`9f1e48b33a5fcda7e621e612ec9305cb52a38e07` and the installed TRL 0.29.1
`PPOConfig` signature.

Requested values:

```yaml
training:
  epochs: 1
  ppo_epochs: 2
  ppo_kl_penalty: 0.7
  ppo_clip_ratio: 0.17
```

Resolved built-in contract:

```text
num_train_epochs: 3.0  # TRL default, requested 1 is lost
num_ppo_epochs:   2    # correct
kl_coef:          0.05 # TRL default, requested 0.7 is lost
cliprange:        0.17 # correct
```

## Impact

PPO can run with a different KL regularization strength and training budget than
the configuration and CLI indicate, without warning. This changes both the
optimization objective and the amount of work performed.

## Suggested fix

- Detect and populate `kl_coef` on current TRL, retaining `init_kl_coef` only for
  older compatible APIs.
- Pass `training.epochs` as `num_train_epochs` on the built-in path.
- Report the effective epoch budget and KL coefficient at setup time.
- Keep `ppo_epochs`/`num_ppo_epochs` semantically separate.

## Acceptance criteria

- [ ] Non-default `ppo_kl_penalty` reaches `PPOConfig.kl_coef` on TRL 0.29.x.
- [ ] Non-default `training.epochs` reaches `PPOConfig.num_train_epochs` on the
      built-in trainer path.
- [ ] Compatibility tests cover both old and current parameter names.
- [ ] Tests assert non-default values so falling back to upstream defaults cannot
      pass accidentally.

be-student:
Fixes #721.

Forwards dataset epochs and the KL coefficient to TRL 0.29's `num_train_epochs` and `kl_coef`, while preserving legacy names and keeping PPO optimization epochs separate. Setup now reports the effective schedule.

Validation: 20,398 tests passed; Ruff passed; coverage 83.00%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,6 @@ reproducing 70+ versions of notes.
 
 ## [Unreleased]
 
-### Fixed
-
-- PPO now forwards `training.epochs` and `ppo_kl_penalty` to TRL 0.29's
-  `num_train_epochs` and `kl_coef` fields while retaining legacy field-name
-  compatibility (#721).
-
 ## [0.74.0] - 2026-09-04
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ reproducing 70+ versions of notes.
 
 ## [Unreleased]
 
+### Fixed
+
+- PPO now forwards `training.epochs` and `ppo_kl_penalty` to TRL 0.29's
+  `num_train_epochs` and `kl_coef` fields while retaining legacy field-name
+  compatibility (#721).
+
 ## [0.74.0] - 2026-09-04
 
 ### Added
--- a/changelog.d/0.74.0/737.fixed.md
+++ b/changelog.d/0.74.0/737.fixed.md
@@ -1,0 +1,3 @@
+- PPO now forwards `training.epochs` and `ppo_kl_penalty` to TRL 0.29's
+  `num_train_epochs` and `kl_coef` fields while retaining legacy field-name
+  compatibility (#721 by @be-student in #737).
--- a/docs/training.md
+++ b/docs/training.md
@@ -1198,6 +1198,7 @@ data:
   format: chatml
 training:
   reward_model: ./output_rm
+  epochs: 1
   ppo_epochs: 4
   ppo_clip_ratio: 0.2
   ppo_kl_penalty: 0.05
@@ -1207,6 +1208,11 @@ training:
   quantization: 4bit
 output: ./output_ppo
 ```
+
+`epochs` controls complete passes over the training dataset. `ppo_epochs`
+controls optimization passes within each PPO update. Soup forwards both values,
+plus `ppo_kl_penalty`, to the active TRL `PPOConfig` names and prints the
+effective schedule during setup.
 
 PPO supports two reward sources:
 - **Reward model** (`reward_model`): pre-trained reward model (from step 2)
--- a/src/soup_cli/trainer/ppo.py
+++ b/src/soup_cli/trainer/ppo.py
@@ -9,7 +9,7 @@ Full RLHF pipeline:  SFT → Reward Model → PPO
 
 import time
 from pathlib import Path
-from typing import Optional
+from typing import Any, Mapping, Optional
 
 from rich.console import Console
 
@@ -24,6 +24,47 @@ from soup_cli.utils.mixed_precision import align_trainable_dtype_for_fp16
 from soup_cli.utils.seeding import apply_training_seed, training_seed_kwargs
 
 console = Console()
+
+
+def _set_ppo_training_kwargs(
+    ppo_kwargs: dict[str, object],
+    ppo_params: Mapping[str, object],
+    tcfg: Any,
+) -> dict[str, str]:
+    """Forward Soup's three PPO schedules across TRL parameter renames."""
+    applied: dict[str, str] = {}
+
+    if "num_train_epochs" in ppo_params:
+        ppo_kwargs["num_train_epochs"] = tcfg.epochs
+        applied["train_epochs"] = "num_train_epochs"
+
+    if "num_ppo_epochs" in ppo_params:
+        ppo_kwargs["num_ppo_epochs"] = tcfg.ppo_epochs
+        applied["ppo_epochs"] = "num_ppo_epochs"
+    elif "ppo_epochs" in ppo_params:
+        ppo_kwargs["ppo_epochs"] = tcfg.ppo_epochs
+        applied["ppo_epochs"] = "ppo_epochs"
+
+    if "kl_coef" in ppo_params:
+        ppo_kwargs["kl_coef"] = tcfg.ppo_kl_penalty
+        applied["kl_coef"] = "kl_coef"
+    elif "init_kl_coef" in ppo_params:
+        ppo_kwargs["init_kl_coef"] = tcfg.ppo_kl_penalty
+        applied["kl_coef"] = "init_kl_coef"
+
+    return applied
+
+
+def _effective_ppo_setting(
+    config: object,
+    kwargs: Mapping[str, object],
+    field: str | None,
+    fallback: object,
+) -> object:
+    """Read the constructed config value, with compatibility fallbacks."""
+    if field is None:
+        return fallback
+    return getattr(config, field, kwargs[field])
 
 
 class PPOTrainerWrapper:
@@ -168,16 +209,12 @@ class PPOTrainerWrapper:
 
         ppo_params = inspect.signature(ppo_config_cls).parameters
 
-        # trl renamed ppo_epochs -> num_ppo_epochs in newer versions
-        if "num_ppo_epochs" in ppo_params:
-            ppo_kwargs["num_ppo_epochs"] = tcfg.ppo_epochs
-        elif "ppo_epochs" in ppo_params:
-            ppo_kwargs["ppo_epochs"] = tcfg.ppo_epochs
+        applied_ppo_fields = _set_ppo_training_kwargs(
+            ppo_kwargs, ppo_params, tcfg
+        )
 
         if "cliprange" in ppo_params:
             ppo_kwargs["cliprange"] = tcfg.ppo_clip_ratio
-        if "init_kl_coef" in ppo_params:
-            ppo_kwargs["init_kl_coef"] = tcfg.ppo_kl_penalty
 
         # Optional params that may not exist in all trl versions
         if "log_with" in ppo_params:
@@ -195,6 +232,30 @@ class PPOTrainerWrapper:
             ppo_kwargs["use_cpu"] = True
 
         ppo_config = ppo_config_cls(**ppo_kwargs)
+        effective_train_epochs = _effective_ppo_setting(
+            ppo_config,
+            ppo_kwargs,
+            applied_ppo_fields.get("train_epochs"),
+            tcfg.epochs,
+        )
+        effective_ppo_epochs = _effective_ppo_setting(
+            ppo_config,
+            ppo_kwargs,
+            applied_ppo_fields.get("ppo_epochs"),
+            tcfg.ppo_epochs,
+        )
+        effective_kl_coef = _effective_ppo_setting(
+            ppo_config,
+            ppo_kwargs,
+            applied_ppo_fields.get("kl_coef"),
+            tcfg.ppo_kl_penalty,
+        )
+        console.print(
+            "[green]PPO schedule:[/] "
+            f"train epochs={effective_train_epochs}, "
+            f"PPO epochs={effective_ppo_epochs}, "
+            f"KL coefficient={effective_kl_coef}"
+        )
 
         # --- Build reward functions list for PPOTrainer ---
         reward_funcs = []
--- a/src/soup_cli/trainer/ppo.py
+++ b/src/soup_cli/trainer/ppo.py
@@ -63,7 +63,7 @@ def _effective_ppo_setting(
 ) -> object:
     """Read the constructed config value, with compatibility fallbacks."""
     if field is None:
-        return fallback
+        return f"{fallback} (not forwarded)"
     return getattr(config, field, kwargs[field])
 
 
--- a/tests/test_issue721_ppo_config_forwarding.py
+++ b/tests/test_issue721_ppo_config_forwarding.py
@@ -52,7 +52,7 @@ def test_legacy_trl_parameter_names_remain_supported(training_config) -> None:
 
 
 def test_installed_trl_exposes_and_receives_current_names(training_config) -> None:
-    pytest.importorskip("trl")
+    pytest.importorskip("trl.experimental.ppo")
     from trl.experimental.ppo import PPOConfig
 
     params = inspect.signature(PPOConfig).parameters
--- a/tests/test_issue721_ppo_config_forwarding.py
+++ b/tests/test_issue721_ppo_config_forwarding.py
@@ -1,0 +1,136 @@
+"""Regression coverage for PPOConfig schedule forwarding (#721)."""
+
+from __future__ import annotations
+
+import inspect
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from soup_cli.trainer.ppo import _set_ppo_training_kwargs
+
+
+@pytest.fixture
+def training_config() -> SimpleNamespace:
+    return SimpleNamespace(epochs=7, ppo_epochs=2, ppo_kl_penalty=0.7)
+
+
+def test_current_trl_parameter_names_receive_non_defaults(training_config) -> None:
+    kwargs: dict[str, object] = {}
+    fields = _set_ppo_training_kwargs(
+        kwargs,
+        {"num_train_epochs": None, "num_ppo_epochs": None, "kl_coef": None},
+        training_config,
+    )
+
+    assert kwargs == {
+        "num_train_epochs": 7,
+        "num_ppo_epochs": 2,
+        "kl_coef": 0.7,
+    }
+    assert fields == {
+        "train_epochs": "num_train_epochs",
+        "ppo_epochs": "num_ppo_epochs",
+        "kl_coef": "kl_coef",
+    }
+
+
+def test_legacy_trl_parameter_names_remain_supported(training_config) -> None:
+    kwargs: dict[str, object] = {}
+    fields = _set_ppo_training_kwargs(
+        kwargs,
+        {"ppo_epochs": None, "init_kl_coef": None},
+        training_config,
+    )
+
+    assert kwargs == {"ppo_epochs": 2, "init_kl_coef": 0.7}
+    assert fields == {
+        "ppo_epochs": "ppo_epochs",
+        "kl_coef": "init_kl_coef",
+    }
+
+
+def test_installed_trl_exposes_and_receives_current_names(training_config) -> None:
+    pytest.importorskip("trl")
+    from trl.experimental.ppo import PPOConfig
+
+    params = inspect.signature(PPOConfig).parameters
+    kwargs: dict[str, object] = {}
+    _set_ppo_training_kwargs(kwargs, params, training_config)
+
+    assert kwargs["num_train_epochs"] == 7
+    assert kwargs["num_ppo_epochs"] == 2
+    assert kwargs["kl_coef"] == 0.7
+
+
+def test_setup_passes_current_schedule_to_trainer(tmp_path, capsys) -> None:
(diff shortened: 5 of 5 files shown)


MakazhanAlpamys:
@be-student — **this one is correct, and the approach is the one this repo learned the hard way. Two small things and I merge it.**

Head `70c90308` verified. Worktree with `PYTHONPATH` pinned inside it.

## Verified against the real trl 0.29.1 `PPOConfig`

`trl.experimental.ppo` cannot import at the declared torch floor (issue #651), so rather than fake it I shimmed **only** the one missing symbol — `torch.distributed.fsdp.FSDPModule` — and drove the genuine class:

```
trl 0.29.1 | torch 2.5.1+cu121 (FSDPModule shimmed)
  num_train_epochs  present=True   default=3.0
  num_ppo_epochs    present=True   default=4
  kl_coef           present=True   default=0.05
  init_kl_coef      present=False  ppo_epochs present=False

MAIN kwargs: {'num_ppo_epochs': 2, 'cliprange': 0.17}
MAIN RESULT  num_train_epochs=3.0 (requested 7)   kl_coef=0.05 (requested 0.7)
PR   kwargs: {'num_train_epochs': 7, 'num_ppo_epochs': 2, 'kl_coef': 0.7}
PR   RESULT  num_train_epochs=7   num_ppo_epochs=2   kl_coef=0.7
```

That is #721's reproduction table reproduced and then closed.

**The two meanings of "epochs" are not conflated**, which was the thing I most expected to find wrong: `epochs=7` reaches `num_train_epochs`, `ppo_epochs=2` reaches `num_ppo_epochs`, distinct values, end to end. There is also no epoch double-count — `_train_builtin` calls `self.trainer.train()` with no Python loop, and `self._num_epochs` is read only in `_train_manual` (line 690), so the two paths stay separate. No double-write either: `kl_coef`/`init_kl_coef` and `num_ppo_epochs`/`ppo_epochs` are `if`/`elif`, so exactly one name is ever written.

**And you probed capabilities, not versions.** `inspect.signature(ppo_config_cls).parameters`, with no trl version number anywhere in the diff. This repo has a written rule about that — `trainer/_trl_compat.py` exists because a version table in this codebase was wrong **twice** — and you arrived at it independently. That is the single best thing in this PR.

**Mutations: 6 applied, 5 killed**, including both conflation directions (`num_train_epochs = ppo_epochs` and the reverse), reverting the block entirely, hardcoding `kl_coef = 0.05`, and removing the schedule print.

Zero deleted test lines. Ruff clean. `docs/training.md` gains six lines that correctly distinguish `epochs` from `ppo_epochs`.

## To fix — one line, and it is a real red test

```
pytest.importorskip("trl")      # guards the wrong module
```

Plain `trl` imports fine here; `trl.experimental.ppo` is what fails at the declared floor. So the test **fails** where it should **skip**:

```
clean main 54cd90a  : 24 failed, 844 passed, 4 skipped
your branch (rebuilt): 25 failed, 847 passed, 4 skipped
                       ^ +3 passed (your new tests), +1 FAILED

FAILED tests/test_issue721_ppo_config_forwarding.py::test_installed_trl_exposes_and_receives_current_names
E   ImportError: cannot import name 'FSDPModule' from 'torch.distributed.fsdp'
```

The other 24 are pre-existing #651 failures and carry over unchanged, so this one is genuinely yours. CI is green only because the GitHub runners resolve a newer torch; anyone installed at the floor this project declares sees a red test. Verified fix:

```
pytest.importorskip("trl.experimental.ppo")
-> SKIPPED [1] could not import 'trl.experimental.ppo'
   3 passed, 1 skipped
```

## Two house conventions

Neither is a criticism — they are things about this repo you had no way to know.

**1. The ANSI trap.** `test_setup_passes_current_schedule_to_trainer` asserts `"train epochs=7" in capsys.readouterr().out` against a Rich `console.print`. This exact failure class has turned CI red **four times** here. Reproduced at default width with colour on:

```
AssertionError: assert 'train epochs=7' in '...PPO \n\x1b[33mepochs\x1b[0m=\x1b[1;36m2\x1b[0m,
  KL \x1b[33mcoefficient\x1b[0m=\x1b[1;36m0\x1b[0m\x1b[1;36m.7\x1b[0m\n'
```

Rich's `ReprHighlighter` styles `key=value` pairs and numbers separately, so the substring cannot match once colour is on — and Rich wraps as well as colours. Currently green because GitHub runners are not TTYs, so this is latent rather than broken. Route it through the ANSI-strip plus whitespace-collapse helper in `tests/test_auto_tuning.py` or `test_v07302.py`.

**2. Changelog fragments, not `CHANGELOG.md`.** Since #487 this repo uses per-PR fragments. Add `changelog.d/0.74.0/737.fixed.md` — the **PR** number, not the issue — crediting `(#721 by @be-student in #737)`, and revert the direct `CHANGELOG.md` edit. Note that `test_issue487_changelog_fragments.py` passed anyway: it only asserts the directory is empty at release, so it cannot detect a fragment that was never written. Nothing would have told you.

## One coverage gap, not blocking

The "preserving legacy names" claim is only half pinned. Swapping the precedence — preferring `init_kl_coef` over `kl_coef` — **survives**, because no test presents both names in one signature: `test_current_*` passes only `kl_coef`, `test_legacy_*` only `init_kl_coef`. Benign today (I confirmed `init_kl_coef present=False` in trl 0.29.1, so no shipped version exposes both), and the same gap exists for `num_ppo_epochs` / `ppo_epochs`. Worth one test with both present, whenever.

## What I could not verify

Because of #651 I could not run a real `PPOTrainer.setup()` end to end, nor any PPO training. The shim above lets me drive the genuine `PPOConfig`, which is what makes the verification real rather than fake-based, but it is not the same as a live run. Your body's "20,398 tests passed" I did not confirm — a fully green suite is not achievable on this box — and I did not check the coverage figure at all. Your counts are self-consistent and I have no reason to doubt them; I am recording that I did not measure them.

Fix the `importorskip` module, add the fragment, and take the ANSI helper while you are in the file — then this merges.

MakazhanAlpamys:
@be-student — one convention item on both this and #736, flagged now so you can fix it while I review the substance. Not a comment on the code, which I have not finished reading yet.

**This repo does not edit `CHANGELOG.md` directly.** Since #487 a user-visible change adds a fragment instead:

```
changelog.d/0.74.0/737.fixed.md
```

Three rules, all in `changelog.d/README.md`:

1. **Named after the pull request, not the issue.** Your entry cites `(#721)`; the filename and the in-text reference both want `#736`. The reason is exactly the situation this repo has been in all week — two pull requests on one issue collide on one filename, and the PR number is what keeps them distinct.
2. **Credit the author**: `(#721 by @be-student in #736)`. Outside work gets its author named; that field is the point of the file.
3. **Write the complete list item** in the fragment. Your prose is already the right shape — it just moves.

Same for #736 (`736.fixed.md`).

Two things worth knowing so this does not look arbitrary:

- **CI cannot catch it.** `tests/test_issue487_changelog_fragments.py` validates fragments that exist and has no way to see a missing one, and nothing rejects a direct `CHANGELOG.md` edit. Your 14/14 green is honest; the check simply is not there.
- **Your `[Unreleased]` text is not wasted.** Entries written there before v0.73.4 are still assembled alongside fragments, so the content is fine — it is the location that is wrong for new work. Move it and drop the `## [Unreleased]` / `### Fixed` scaffolding; `scripts/assemble_changelog.py` adds those.

Substance review on both is next in my queue. Standard battery here, so you know what is coming: I mutate the main fix, and a test that survives the mutation is a blocking finding regardless of how many pass.

MakazhanAlpamys:
@be-student — **a second pull request against #721 exists, and you should hear it from me rather than discover it. Nothing about my assessment of yours has changed.**

**#741 by @dchaudhari7177** was opened at `2026-09-07T13:13:41Z`, 18 hours and 23 minutes after yours, and I failed to tell either of you about the other until now. I also added the `claimed` label to #721 at 18:07 UTC that day **without a comment saying who held the claim**, which under this project's own rules records nothing — `CONTRIBUTING.md` says the comment is the claim and the thread is the record. So neither of you had a readable signal, and that is mine, not theirs and not yours.

## Where this leaves you

**Your review stands as written.** I said on 2026-09-07 at 06:47 that yours is correct and that two small items stood between it and a merge, and I have not withdrawn that. Being first is also a fact in your favour that I am not going to quietly drop.

What I will not do is decide it on order of arrival alone and call that a review. Both are being measured side by side right now — the same four mutations applied to each, including both directions of the epoch-conflation bug; which of #721's four acceptance criteria each satisfies; and where the compat logic sits, since #741 places it in `trainer/_trl_compat.py`, the module this repo designates for exactly that, while yours probes capabilities inline. Your capability probe is the thing I singled out as the best part of your PR and that judgement is unchanged; the question I owe both of you an honest answer to is whether the placement matters enough to outweigh 18 hours and a completed review.

I will post the comparison in both threads with the measurements in it.

## The two items from my review are still the two items

Unchanged and still yours to close whenever you like:

1. `pytest.importorskip("trl")` guards the wrong module — plain `trl` imports at the declared torch floor, `trl.experimental.ppo` does not (#651), so the test **fails** where it should **skip**. `pytest.importorskip("trl.experimental.ppo")` is the fix, verified.
2. The changelog convention — a fragment at `changelog.d/0.74.0/737.fixed.md` rather than an edit to `CHANGELOG.md`, with `by @be-student` in the text.

If you would rather wait for the comparison before spending time on those, that is entirely reasonable and I will not read it as the PR going quiet.

MakazhanAlpamys:
@be-student — **the comparison is done and yours is the one I am taking forward. It is still blocked on three items, two of which I raised before and one of which #741 exposed.**

## Why yours, stated on evidence rather than on arrival order

The shipped behaviour of the two PRs is **identical** — I drove the genuine `trl.experimental.ppo.PPOConfig` with each PR's real kwarg code and both produce `num_train_epochs=7, num_ppo_epochs=2, kl_coef=0.7` where `main` produces `3.0 / 2 / 0.05`. Both probe capabilities, both keep the two epochs separate, both support the legacy names. The decision came down to tests.

**The reverse conflation.** Mutating `ppo_kwargs["num_ppo_epochs"] = tcfg.ppo_epochs` to `tcfg.epochs` — the exact bug #721 warns a fix must not introduce:

```
#737: 3 failed, 1 passed   (+2 over baseline: test_current_trl_parameter_names_receive_non_defaults,
                            test_setup_passes_current_schedule_to_trainer)
#741: 7 passed             SURVIVOR
```

Yours is the only one that pins it. #741's guard is a source-substring check, and the string it looks for survives in the untouched legacy `elif`.

Yours also fully satisfies acceptance criterion 4 — it is the only PR that drives a real `PPOTrainerWrapper.setup()` against a config whose defaults are trl's own, so a silent fallback cannot pass — and it is the only one implementing the issue's fourth suggested-fix bullet, the effective-schedule report, plus six lines of `docs/training.md`.

## Three items block the merge

**1. `pytest.importorskip("trl")` — still unfixed on `70c90308`, and it is a real red test.** Confirmed again just now at the declared floor:

```
FAILED test_installed_trl_exposes_and_receives_current_names
E   ImportError: cannot import name 'FSDPModule' from 'torch.distributed.fsdp'
1 failed, 3 passed
```

Plain `trl` imports; `trl.experimental.ppo` is what fails (#651). `pytest.importorskip("trl.experimental.ppo")` turns it into a skip. CI is green only because the runners resolve a newer torch, so anyone installed at the floor this project *declares* sees it red.

**2. The changelog convention.** `CHANGELOG.md` is still edited directly (+6 lines) and cites `(#721)`, the issue rather than the PR. Since #487 this repo uses fragments: delete that hunk and add `changelog.d/0.74.0/737.fixed.md` with `by @be-student` in the text. Note this is not CI-detectable in either direction — a missing fragment is invisible to the fragment test — so it is on me to ask and on nobody's automation to catch.

**3. New, and it came from reviewing #741 against yours: the rename precedence is untested.** `test_current_*` presents only `kl_coef` and `test_legacy_*` only `init_kl_coef`, so no test ever puts both names in one signature — and preferring the legacy name would survive your whole file. @dchaudhari7177 covered exactly this in `test_the_new_name_wins_when_a_config_carries_both`. Please add the equivalent; it is a few lines, and the credit for spotting the gap is theirs.

## One thing to fix that is yours alone

`_effective_ppo_setting` falls back to the **requested** value when a field was never forwarded:

```python
if field is None:
    return fallback          # prints the request, not the effective value
```

On a legacy signature with no `num_train_epochs`, the report prints `train epochs=7` while nothing of the sort was sent. Latent today — trl 0.29.1 has the field — but the report exists precisely so a user is not misled about the effective budget, and that is the one case where it would mislead. One line: say `7 (not forwarded)` or similar.

## And a lead, not a verdict

Under `FORCE_COLOR=1 TERM=xterm-256color`, your `assert "train epochs=7" in ...` fails on Rich's per-character escapes:

```
'PPO \x1b[33mepochs\x1b[0m=\x1b[1;36m2\x1b[0m, KL \x1b[33mcoefficient\x1b[0m=\x1b[1;36m0\x1b[0m\x1b[1;36m.7\x1b[0m'
```

`FORCE_COLOR` is **stricter** than real CI and your run is genuinely green today, so treat this as worth hardening rather than as a failure. This class has turned this repo's CI red four times; `tests/test_v07302.py` carries the canonical ANSI-strip helper. Your call whether to take it now.

## What I want from #741, so you know what is coming

Its `kl_penalty_kwargs` lives in `trainer/_trl_compat.py`, which is the module this project designates for exactly this, and yours is local to `ppo.py`. I have asked @dchaudhari7177 whether they want to land that move as a follow-up after yours, with their precedence test, credited to them. It changes nothing you need to do here — I am telling you so it does not arrive as a surprise edit on top of your work.

Fix 1 and 2, add the precedence test, and I merge on green.

MakazhanAlpamys:
@be-student — **a date and an offer, because this is the head of a chain and it has been still for a day.**

Nothing new is being asked. The four items from my 2026-09-08 review stand exactly as written:

1. `pytest.importorskip("trl")` -> `pytest.importorskip("trl.experimental.ppo")` (plain `trl` imports at the declared floor; the experimental module is what does not).
2. Delete the direct `CHANGELOG.md` hunk, add `changelog.d/0.74.0/737.fixed.md` crediting you.
3. Add the rename-precedence test — both `kl_coef` and `init_kl_coef` in one signature. Gap spotted by @dchaudhari7177.
4. `_effective_ppo_setting` returning the *requested* value when the field was never forwarded, so the report can print a budget that was not sent.

Plus the `FORCE_COLOR` ANSI hardening, which I flagged as a lead rather than a blocker and still do.

## Why I am putting a clock on it

**#737 is blocking two other people's work, not just its own.** @dchaudhari7177's #741 was closed out in your favour on the understanding that their `kl_penalty_kwargs` move into `_trl_compat.py` and their precedence test land as a follow-up *after* yours. They accepted a losing collision on that basis. Every day this sits, that follow-up cannot start, and the person waiting is the one who lost the collision.

You also have #736 and #730 open with me. I would rather not have three of my threads on you at once, so: **#737 is the one I want first**, because it is the only one with someone else queued behind it. #736's blocker is unchanged (3.1x wall clock, and that is a cost question, not a correctness one) and #730 is on me, not you — I withdrew the one-liner I had endorsed there because I had not checked that it fixed what I reproduced.

## The offer

**2026-09-11 18:00 UTC.** If you push the four items before then, I review and merge on green, and it is entirely your PR.

If you would rather not — no explanation owed, and it costs you nothing — say so, or say nothing and let the date pass. I will then apply the four items myself on top of your commits, keep you as the author of every one of them, credit the PR to you in `CONTRIBUTORS.md` and in the changelog fragment, and note in the merge commit exactly which four lines were mine. The design, the diagnosis and the fix are yours either way; the outstanding items are a `importorskip` argument, a file move, a test and one line of fallback logic, and none of those are worth a chain of three PRs stalling on.

Tell me which you prefer if you have a preference. Otherwise the date decides.

MakazhanAlpamys:
@be-student — the date I set, 2026-09-11 18:00 UTC, has passed, so as promised I am applying the four items from my 2026-09-08 review myself, on top of your commits. They are the `importorskip` argument, the file move, the test and the line of fallback logic. You stay the author of every commit. The PR, the `CONTRIBUTORS.md` entry and the changelog fragment are credited to you, and the merge commit will say exactly which lines were mine. The design, the diagnosis and the fix are yours.

If you would rather take it back, say so at any point, including after I push, and it is yours again. This also unblocks @dchaudhari7177's follow-up from #741, which was waiting on it.

#736 and #730 are unaffected: they wait on you with no date, and I will not touch them without asking first.

MakazhanAlpamys:
@be-student — pushed as promised. Your commit `70c90308` is untouched, and these five are mine, on top of it (fast-forward, no force):

| commit | what |
|---|---|
| `7e6bb65e` | item 1 — `importorskip("trl.experimental.ppo")`, skipping on the module that fails to import rather than on `trl` |
| `0d5e86be` | item 2 — the #721 entry moved from `CHANGELOG.md` into `changelog.d/0.74.0/737.fixed.md`, credited `(#721 by @be-student in #737)` |
| `6042fe42` | item 3 — one test with all five parameter names in a single signature, so the current names must win over the legacy ones. The gap was spotted by @dchaudhari7177 in #741, credited in the test comment. The `kl_penalty_kwargs` move stays with #741's follow-up. |
| `eaf25f08` | item 4 — an unforwarded schedule value is shown as `7 (not forwarded)` instead of echoing the request |
| `e9f10150` | a test for item 4, which had none: reverting that line passed the suite before this commit |

Verification: 90 passed across the PPO, #721 and trl-contract files; startup-lightness passes; ruff clean. Mutations killed: dropping `num_train_epochs` forwarding, dropping the `kl_coef` branch, swapping the two epoch meanings either way, putting `main`'s setup back, preferring either legacy name (only the new test catches that), and reverting item 4.

One honest caveat on item 1: since pytest 9.1, `importorskip` only catches `ModuleNotFoundError`, and the #651 failure mode is a plain `ImportError`, so on pytest 9.1+ a broken trl import still fails the test rather than skipping it. That matters little after merge — `main` now requires torch >= 2.6.0 (#717), where trl imports cleanly — and `exc_type=ImportError` would need pytest >= 8.2 against a dev floor of 7.0, so I left it as written.

I will merge on green CI unless you say otherwise.

MakazhanAlpamys:
Merged as `9ba0b577`. Thank you, @be-student.

**The substance, the diagnosis and the fix are yours**, and your commit `70c90308` went in untouched. PPO accepted `training.epochs` and `training.ppo_kl_penalty`, documented them, and forwarded neither to trl 0.29 — this project's most common defect shape, an option validated and read by nothing at all. Two things set your PR apart from the others on this issue: it is the only one that drives a real `PPOTrainerWrapper.setup()` against a config whose values are trl's own defaults, so a silent fallback cannot pass the test, and it is the only one that implemented the effective-schedule report the issue asked for.

As announced on 2026-09-09 and again when the 2026-09-11 18:00 UTC date passed, I applied the four outstanding review items myself rather than let the chain stall. The merge commit lists exactly which five commits were mine — the `importorskip` argument, the fragment move, the rename-precedence test (gap spotted by @dchaudhari7177 in #741), the "(not forwarded)" line and its test. If you would rather have taken them yourself, say so and I will happily revisit how the next one is handled.

**What I ran.** 90 passed across the PPO, #721 and trl-contract files; CLI startup-lightness passes; ruff clean. Mutations killed: dropping the `num_train_epochs` forwarding, dropping the `kl_coef` branch, swapping the two epoch meanings in either direction, restoring `main`'s setup, preferring either legacy name (only the new precedence test catches that), and reverting the "(not forwarded)" line. 13/13 required checks green on `e9f10150`.

Your `CONTRIBUTORS.md` entry is in `b5723ce8`, and the changelog fragment credits you.

This also unblocks @dchaudhari7177's follow-up from #741, moving `kl_penalty_kwargs` into `trainer/_trl_compat.py`.

#736 and #730 are still yours with no clock on either, and I have not touched them.

Issue #725, opened 2026-09-06
Add optional PEFT LoRA-FA optimizer support


Amix29:
## Motivation

PEFT exposes LoRA-FA, an optimizer that freezes LoRA A matrices and trains B
matrices. Freezing A avoids retaining the input activation needed for A's gradient,
so the adapter-rank contribution to activation memory can be substantially lower.

Primary references:

- PEFT optimizer API: `create_lorafa_optimizer`
- LoRA-FA paper: https://arxiv.org/abs/2308.03303

## Local validation

With PEFT 0.20.0, a tiny real `PeftModel` confirmed that
`create_lorafa_optimizer()` creates a `LoraFAOptimizer`, sets every `lora_A`
parameter to `requires_grad=False`, and leaves `lora_B` trainable.

An isolated activation-saved-tensor probe for `X @ A @ B` used
`X: 512x4096` and ranks 16/64/256:

| Rank | Ordinary LoRA retained | Frozen-A retained |
|---:|---:|---:|
| 16 | 8.03125 MiB | 32 KiB |
| 64 | 8.125 MiB | 128 KiB |
| 256 | 8.5 MiB | 512 KiB |

The difference was exactly 8 MiB in each case: the storage of `X` in this toy
calculation. Input and B gradients matched for the same calculation, while A had no
gradient by design.

These values are **not** total or peak LLM VRAM savings, an end-to-end throughput
result, or a quality claim. Other model operations may still retain the same input,
and backward recomputation/optimizer state are outside this micro-test.

## Proposed scope

- Add an explicit optional LoRA-FA optimizer mode for Transformers/PEFT trainers.
- Construct it with `create_lorafa_optimizer()` using the active adapter's rank,
  alpha, learning rate, and weight decay.
- Integrate scheduler construction, checkpoint/resume, distributed training, and
  Soup's optimizer reporting.
- Define and test compatibility with LoRA+, DoRA, rsLoRA, quantization,
  DeepSpeed/FSDP, and layer streaming; reject unsupported combinations clearly.
- Keep MLX out of scope unless an equivalent native implementation is designed and
  measured separately.

## Acceptance criteria

- [ ] A matrices are frozen and excluded from optimizer updates; B matrices train.
- [ ] Checkpoint/resume reproduces optimizer and scheduler state.
- [ ] A small end-to-end training smoke test produces finite loss and gradients.
- [ ] Peak memory, throughput, and quality are compared with ordinary LoRA using
[...]

PR #738, opened 2026-09-06
fix(trainer): build a PEFT LoRA+ optimizer instead of forwarding loraplus_lr_ratio to TrainingArguments

### LINKED ISSUE #724: LoRA+ is passed to TrainingArguments instead of a PEFT optimizer
## Summary

When `training.loraplus_lr_ratio` is set, the SFT, pretrain, and embedding wrappers
insert it into `training_kwargs` and pass it to Transformers
`TrainingArguments(**training_kwargs)`.

`loraplus_lr_ratio` is not a `TrainingArguments` parameter. It belongs to PEFT's
LoRA+ optimizer construction, so enabling the advertised option fails before the
first training step.

Affected forwarding sites on current `main`:

- `src/soup_cli/trainer/sft.py`
- `src/soup_cli/trainer/pretrain.py`
- `src/soup_cli/trainer/embedding.py`

## Reproduction

`9f1e48b33a5fcda7e621e612ec9305cb52a38e07` with Transformers 5.16.1 and
PEFT 0.20.0:

```text
TrainingArguments.__init__() got an unexpected keyword argument
'loraplus_lr_ratio'
```

The PEFT integration path was also exercised on a tiny local LoRA model:

```text
base LoRA learning rate: 2e-5
loraplus_lr_ratio:       16
A group learning rate:   2e-5
B group learning rate:   3.2e-4
```

## Impact

The option is accepted by Soup's schema and documented, but every affected
Transformers training path crashes when it is activated. Removing the unknown
keyword alone would only turn the crash into silently disabled LoRA+.

## Suggested fix

- Do not forward `loraplus_lr_ratio` to `TrainingArguments`.
- Build the optimizer with PEFT's `create_loraplus_optimizer(model,
  optimizer_cls, lr=..., loraplus_lr_ratio=...)`.
- Inject the resulting optimizer through the Trainer's supported optimizer hook
  while preserving scheduler creation, resume, checkpoint, DeepSpeed/FSDP, and
  ordinary optimizer behavior.
- Centralize the wiring so SFT, pretrain, and embedding cannot drift apart.

## Acceptance criteria

- [ ] Enabling LoRA+ constructs and starts each affected trainer without an
      unexpected-keyword error.
- [ ] A real optimizer test identifies A and B parameters and asserts the requested
      learning-rate ratio.
- [ ] Saving/resuming preserves optimizer and scheduler state.
- [ ] Unsupported combinations fail explicitly rather than silently falling back
      to ordinary AdamW.
- [ ] The default path remains unchanged when `loraplus_lr_ratio` is unset.

abdulwaarith0:
Fixes #724.

## What was wrong

The SFT, pretrain and embedding wrappers inserted `loraplus_lr_ratio` into `training_kwargs` and forwarded it to `TrainingArguments(**training_kwargs)`. `loraplus_lr_ratio` is not a `TrainingArguments` field, so the advertised, schema-accepted option raised `TypeError: __init__() got an unexpected keyword argument 'loraplus_lr_ratio'` before the first training step — enabling LoRA+ always crashed.

## The fix

A shared `attach_loraplus_optimizer(trainer, tcfg)` helper (in `utils/peft_wiring.py`, alongside the existing `attach_*` helpers) builds a PEFT `create_loraplus_optimizer` optimizer — LoRA **B** matrices at `lr * ratio`, **A** at `lr` — and assigns it to `trainer.optimizer` after the trainer exists. That is respected because `Trainer.create_optimizer` builds one only when `self.optimizer is None`, and the scheduler is still built from it with the configured warmup/schedule.

- The optimizer **class and betas/eps** come from the run's configured optimizer via `Trainer.get_optimizer_cls_and_kwargs`, so LoRA+ uses the optimizer the user asked for.
- **Weight decay** is applied through PEFT's own `loraplus_weight_decay` (PEFT ignores a plain `weight_decay` kwarg here).
- **Guards:** LoRA+ combined with GaLore, or set on a non-LoRA run, is rejected with a clear message rather than silently overriding the optimizer or doing nothing.

The three wrappers drop the leak and call the helper at their existing post-construction wiring block.

## Verification

- `tests/test_issue724_loraplus_optimizer.py` — 7 tests using a **real PEFT model and a real `transformers.Trainer`, no mocks** (a mock would auto-create the LoRA parameter groups this path depends on and hide the defect). They assert the split A/B learning rates (`2e-5` / `3.2e-4`, matching the issue's reproduction), the configured optimizer class, weight-decay routing, the no-op path, both guard paths, and that `TrainingArguments` still rejects the kwarg (pins *why* the old wiring crashed).
- **Mutation:** neutering the attach fails 5 of 7; swapping `loraplus_weight_decay` for the plain kwarg fails the exact weight-decay test.
- ruff clean; 141 passed across the affected embedding/peft/changelog suites (rebased onto current `main`, so it coexists with #705's embedding changes).

## Scope

Two focused PRs, not one: I asked on the issue whether to fold #725 (LoRA-FA) in here or keep them separate and defaulted to separate. `attach_loraplus_optimizer` is deliberately the shared seam a future `attach_lorafa_optimizer` (#725) would sit beside, so that work extends this rather than duplicating it. Exit-code / CI behaviour is unchanged.


MakazhanAlpamys:
@abdulwaarith0 — **the fix is correct and I verified it produces the right optimizer, not just the right code. I am asking for one test and then merging, and the test is one this repo already has a template for.**

Head `cda6190` verified. Worktree with `PYTHONPATH` pinned inside it.

## The decisive check: does it actually change the param groups?

Enumerated `optimizer.param_groups` with names, at `lr=2e-5`, `loraplus_lr_ratio=16`:

```
group 0: lr=2e-05    wd=0.01  n=4  -> q_proj.lora_A.default.weight, v_proj.lora_A..., (x4)
group 1: lr=1e-06    wd=0.01  n=0
group 2: lr=0.00032  wd=0.01  n=4  -> q_proj.lora_B.default.weight, v_proj.lora_B..., (x4)
group 3: lr=0.00032  wd=0.0   n=0

group 0: ratio_vs_base=1   kinds=['A']
group 2: ratio_vs_base=16  kinds=['B']
```

A at `2e-5`, B at `3.2e-4` — exactly the numbers #724 reports from its own PEFT reproduction. **With the ratio unset**, `attach` returns `False` and `trainer.optimizer` stays `None`, i.e. byte-for-byte the old lazy path. No silent change for anyone.

**The scheduler survives it**, which your PR claims but does not test: `LambdaLR`, `base_lrs=[2e-05, 1e-06, 0.00032, 0.00032]`, and the ratio holds at exactly 16 through warmup and decay (step 5: 1e-05 / 1.6e-4; step 10: 2e-05 / 3.2e-4; step 50: 1.11e-05 / 1.78e-4).

**The DeepSpeed interaction is safe, and there is a fact here worth knowing.** LoRA+ emits **four** groups, **two of them empty** — more empty groups than plain LoRA, which is precisely the #336 crash shape. It is covered, because the guard wraps the *method* and HF's `create_optimizer` returns a pre-set `self.optimizer` unchanged:

```
groups BEFORE create_optimizer(): [(2e-05, 4), (1e-06, 0), (0.00032, 4), (0.00032, 0)]
DeepSpeed: dropped 2 empty optimizer parameter group(s)
groups AFTER : [(2e-05, 4), (0.00032, 4)]        REMAINING EMPTY GROUPS: []
```

Install order is irrelevant for the same reason. Non-LoRA combinations are rejected at **config parse**, before the model loads — `lora.r: 0`, LISA and Spectrum all name `loraplus_lr_ratio` in the rejection — so your runtime `ValueError` is defence in depth rather than the only line.

**Six of seven mutations killed:** ratio forced to 1.0, helper returning `False`, `loraplus_weight_decay` -> `weight_decay`, GaLore guard removed, `isinstance(model, PeftModel)` guard removed, and `optimizer_kwargs.pop("lr")` removed.

Zero deleted test lines. Regression sweep: your branch 5 failed / 2241 passed, clean main 5 failed / 2234 passed — identical failure set, delta is exactly your 7 new tests. Ruff clean. Fragment `changelog.d/0.74.0/738.fixed.md`, PR number, credits you, `fixed` category. And `docs/peft-and-efficiency.md:316` already says `loraplus_lr_ratio: 16.0  # lr_B = lr x 16` — **your PR is what makes that sentence true for the first time**, so no docs change is needed.

## The one thing I want: the production wiring is unprotected

```
M7: replace attach_loraplus_optimizer(...) with `pass` in ALL THREE wrappers
    sft.py:1902, pretrain.py:277, embedding.py:222

$ pytest tests/test_issue724_loraplus_optimizer.py tests/test_advanced_peft.py \
         tests/test_issue359_deepspeed_guard_coverage.py tests/test_v07134.py \
         tests/test_issue341_seed_and_fullft.py
303 passed, 1 skipped
```

The entire wiring can be deleted and the suite stays green. Your tests exercise the helper in isolation; nothing asserts it is called.

This matters more here than it usually would, and #724 says why in its own words: *"Removing the unknown keyword alone would only turn the crash into silently disabled LoRA+."* Before your PR a broken wiring **crashed** on the unknown `TrainingArguments` keyword. After it, a broken wiring is silent — the run completes, the loss curve looks normal, and LoRA+ simply never happened. You have made the failure mode quieter, which is right, and that is exactly why it now needs a test that a crash used to provide for free.

**This repo already has the pattern.** `tests/test_issue359_deepspeed_guard_coverage.py` is an AST scan asserting every trainer wrapper attaches the DeepSpeed guard — written for the same problem (a fix applied to `sft.py` only, #336 vs #359). Mirror it: scan `trainer/*.py` for the wrappers that build a PEFT model and assert each one calls `attach_loraplus_optimizer`. That way a fourth wrapper added later fails the scan instead of quietly shipping without LoRA+.

Add that and I merge on sight.

**GaLore is guarded, APOLLO is not.** `use_galore` + `loraplus_lr_ratio` is caught by your runtime guard, but it is not a schema conflict — so it is caught after the model loads rather than in milliseconds at parse. A schema cross-validator would be the better place. Not a regression: it crashed on `TrainingArguments` before. Related: the APOLLO/GaLore family resolves `optim_target_modules` from `model`, and `get_optimizer_cls_and_kwargs(args)` is called with `model=None`; GaLore is guarded, APOLLO is not. I could not test that path — those packages are not installed here.

**The fragment has no trailing newline.** Shared with `712.fixed.md`, so pre-existing style rather than something you introduced. Fold it in if you touch the file.

## What I could not verify

An end-to-end `soup train` with LoRA+ (needs a real model download); **save/resume of optimizer and scheduler state**, which is acceptance criterion 3 in #724 and is covered by no test in this PR — I could not run a train-then-resume here, so that criterion is unproven rather than met; the unsloth backend path, where the attach is unconditional and depends on `FastLanguageModel` returning a `PeftModel`; and a real DeepSpeed launch. Recording these so the merge does not read as a claim about them.

abdulwaarith0:
Added the wiring coverage scan you asked for: `tests/test_issue724_loraplus_wiring_coverage.py`, mirroring `test_issue359_deepspeed_guard_coverage`. It scans `trainer/*.py` and asserts every wrapper that applies a LoRA adapter (get_peft_model) either calls attach_loraplus_optimizer or is named, with a reason, in `_LORAPLUS_NOT_IMPLEMENTED`. Your M7 now fails it: replacing the attach call with `pass` in all three wrappers fails 5 of the scan's tests. There's also an "exemption stays earned" test, so a listed module that later gains LoRA+ (or stops building a PEFT model) fails until it's moved.

Writing it surfaced the thing your request was pointing at: LoRA+ is a shared, un-gated TrainingConfig option, but only the three SFT-family wrappers implement it. Thirteen other PEFT-building trainers (dpo, ppo, grpo, kto, orpo, simpo, bco, ipo, reward_model, classifier, distill, asr, unlearn) accept loraplus_lr_ratio and silently ignore it, building no custom optimizer at all. That's a real gap but it's wider than #724 (the three that crashed), so I've listed those thirteen in `_LORAPLUS_NOT_IMPLEMENTED` rather than expand this PR. Happy to file a follow-up for "extend LoRA+ to the preference/RL trainers, or reject it there," and the APOLLO-vs-GaLore schema-guard point you raised fits naturally in that same follow-up.

Folded in the trailing newline on the fragment while I was there. On criterion 3 (save/resume of optimizer + scheduler state): I haven't added a train-then-resume test, so I'd flag it as still unproven rather than claim it here.

MakazhanAlpamys:
Merged as `d7aea47`. **The scan you added is the thing that makes this fix survive its own maintainers, and I verified the premise for it rather than taking it on faith.**

## The M7 premise, reproduced exactly

With `attach_loraplus_optimizer(...)` replaced by `pass` in all three wrappers — `sft.py:1902`, `pretrain.py:277`, `embedding.py:222` — the five pre-existing files I named still report:

```
303 passed, 1 skipped
```

Byte-identical to the figure I gave you. The entire fix is deleted and nothing in the old suite notices. Your new file fails 5 tests on that same tree:

```
test_every_peft_builder_wires_loraplus_or_is_exempt[sft]
test_every_peft_builder_wires_loraplus_or_is_exempt[pretrain]
test_every_peft_builder_wires_loraplus_or_is_exempt[embedding]
test_the_three_sft_family_wrappers_are_the_ones_wired
test_no_peft_builder_quietly_loses_the_wiring
```

## Seven mutations at merge, seven killed

| mutation | dying test |
|---|---|
| attach call -> `pass` in all three wrappers | the 5 above |
| `loraplus_weight_decay=` -> plain `weight_decay=` | `test_weight_decay_is_applied_through_loraplus` |
| `trainer.optimizer =` -> unused local | `test_loraplus_optimizer_is_attached_with_split_learning_rates`, `test_uses_the_configured_optimizer_class`, `test_weight_decay_is_applied_through_loraplus` |
| non-LoRA guard -> `if False:` | `test_non_peft_model_raises` |
| invented `"ghost_trainer.py"` added to the exemption list | `test_the_exemption_list_stays_earned` |
| exempt `dpo.py` gains an `attach_loraplus_optimizer(` call | `test_the_exemption_list_stays_earned`, `test_the_three_sft_family_wrappers_are_the_ones_wired` |
| brand-new `trainer/newthing.py` calling `get_peft_model`, neither wired nor exempt | `test_every_peft_builder_wires_loraplus_or_is_exempt[newthing]`, `test_no_peft_builder_quietly_loses_the_wiring` |

The last one is mine rather than yours — I wanted to know whether the scan catches a module that does not exist yet, since that is the case it will actually face. It does.

Counts on the merge tree: 45 passed on the two new files; 296 passed / 1 skipped across `test_advanced_peft`, `test_issue359_deepspeed_guard_coverage`, `test_v07134`, `test_issue341_seed_and_fullft`; 90 passed across the embedding suites. Ruff clean. Zero deleted or weakened test lines — the diff is additive only.

## Two things you did that I want named

**You refused the mock, and you said why.** A `MagicMock` trainer auto-creates the LoRA parameter groups this path depends on, so the easy version of this test passes on the broken code. Using a real PEFT model and a real `transformers.Trainer` is more work and it is the only version that measures anything. This repo has shipped the other kind more than once.

**You found the spelling that works rather than the one that reads right.** PEFT ignores a plain `weight_decay` here and honours `loraplus_weight_decay`; swapping them fails exactly one test, which means you knew the difference was real and pinned it.

The scan is also non-vacuous, which I checked because a scan that finds nothing passes forever: it asserts more than 20 trainer modules found and at least 15 PEFT builders, and `test_the_patterns_would_catch_the_unwired_shape` and `test_a_comment_mentioning_the_call_does_not_satisfy_it` exercise both regexes including the comment-stripping path.

## Recorded, not hidden

Your own scope statement is in the merge commit: **criterion 3 of #724 — save/resume of optimizer and scheduler state — has no test here**, and the thirteen preference/RL trainers are exempt rather than wired. You put those in the exemption set with reasons instead of letting them evaporate, which is why #724 stays open for that half rather than being marked done falsely.

Two notes for whoever picks that up, neither an ask on you:

- `_code_without_comments` splits naively on `#`, so a `#` inside a string literal before a `get_peft_model(` call would hide it. Under-detection, and capped by the >=15 assertion.
- `test_the_three_sft_family_wrappers_are_the_ones_wired` pins the wired set to exactly three, so extending LoRA+ to a fourth wrapper correctly fails it. That friction is deliberate; just note it is two tests to update, not one.

You also asked on the issue whether to fold #725 (LoRA-FA) in here and defaulted to separate. That was the right call, and `attach_loraplus_optimizer` is now exactly the seam a future `attach_lorafa_optimizer` sits beside.

Credited in `CONTRIBUTORS.md` (`0c1c1c9`) as your third merge.

PR #739, opened 2026-09-07
feat(mlx): report validation loss from on_val_loss_report (#23)

### LINKED ISSUE #23: MLXSFTTrainerWrapper: add end-to-end training-loop integration test on Apple Silicon

Srinivasan8888:
Closes #23's `on_val_loss_report` checklist item. Refs #713.

> ⚠️ **Stacked on #713, deliberately.** The sinks this producer feeds — `TrainEvent.val_loss`, the `metrics.val_loss` column and its migration, and `TrainingDisplay`'s val series — do not exist on `main`; they are #713. Branched off it rather than duplicating it, so **this diff is only the producer**. Merge #713 first and this rebases to nothing; the base commits will drop out on their own. If you would rather I fold it into #713 as one PR, say so and I will.

Rebased onto **@AmirF194's #698** now that it has landed, exactly as I said on #23, #683 and #686 I would.

## The payload, read from upstream rather than assumed

`mlx_lm/tuner/trainer.py:310-315`:

```python
val_info = {"iteration": it - 1, "val_loss": val_loss, "val_time": val_time}
training_callback.on_val_loss_report(val_info)
```

Two things I took from reading it rather than guessing:

- **`it - 1`.** Validation reports one behind the training counter. I record it as sent rather than "fixing" it, so a row lines up with upstream's own `Iter N: Val loss ...` log line. Pinned by a test.
- **`val_loss` is already a Python float** — `evaluate()` ends in `.item()`. It is still coerced, because a future build returning an `mx` scalar would otherwise put an unserialisable object on the SSE wire.

Worth noting for review: before this change the pre-existing behaviour was a **silent no-op**, not an error — `TrainingCallback.on_val_loss_report` is a `pass` in the base class, so mlx-lm called it and nothing happened. That is the failure shape this repo keeps producing, and it is why the tests assert on sinks rather than on the method existing.

## The discipline this PR is really about

Your BLOCKING 3 on #713 was a sticky value persisting on every training step, fabricating 9 database rows for 2 measurements. Same hazard here, so it is guarded on this backend:

- **the panel** gets the sticky value on every training step — it must not blink out between evaluations;
- **the tracker and the SSE wire** get a value *only* when one was measured.

## Live proof on Metal — real hook, real DB, no fakes

Real `MLXSFTTrainerWrapper`, real val split, real Metal, real `ExperimentTracker` against a real SQLite file. mlx-lm fires its own hook:

```
DISPLAY  updates carrying val_loss : 9 / 9 total          <- sticky; panel never blanks
DB       metrics columns           : val_loss present
DB       rows with val_loss        : 5 of 9 total rows    <- one per evaluation
           step=0   val_loss=4.7491
           step=5   val_loss=2.7466
           step=11  val_loss=0.9455
           step=17  val_loss=0.5798
           step=23  val_loss=0.4529
SSE      events carrying val_loss  : 5 of 9 total
           {'type':'metric','step':5,'epoch':0.208,'val_loss':2.7466}

VERDICT  all three sinks reached : True
         no fabricated rows      : True   (db rows == sse events == evaluations)
```

The three counts agreeing — 9 display, 5 DB, 5 SSE — *is* the assertion. Nine on the panel and five in the database is the correct shape; nine and nine would be the defect.

Note the steps: `0, 5, 11, 17, 23` — upstream's off-by-one, recorded verbatim.

## Mutation table

`tests/test_issue23_mlx_val_loss_producer.py`, baseline **12 passed**. `__pycache__` cleared each round; `mlx_sft.py` md5-verified byte-identical after.

|---|---|---|---|
| M1 | `on_val_loss_report` deleted entirely | **KILLED** | `test_it_reaches_the_display` |
| M2 | SSE event built without `val_loss` | **KILLED** | `test_it_reaches_the_tracker` |
| M3 | `tracker.log_metrics` called without `val_loss` | **KILLED** | `test_it_reaches_the_tracker` |
| M4 | panel loses the sticky value between evaluations | **KILLED** | `test_but_the_panel_does_keep_showing_it` |
| M5 | **sticky value persisted on training steps (the #713 defect)** | **KILLED** | `test_training_steps_do_not_persist_a_carried_forward_value` |
| M6 | a payload with no `val_loss` recorded anyway | **KILLED** | `test_a_missing_val_loss_key_writes_no_row_at_all` |
| M7 | upstream's `iteration` discarded | **KILLED** | `test_the_iteration_key_is_upstreams_off_by_one` |

**7/7.**

### M5 and M6 survived the first pass, and both were my tests' fault

**M5 is the most important mutation in this PR** — it is your #713 finding, reintroduced. It survived because `on_train_loss_report` returns early when `display is None`, and my test passed a tracker but **no display**. The tracker was therefore never reached on a training step, so the defect was invisible to a test written specifically to catch it. The test now passes a display and asserts `len(t.metrics) > 2` first, so it fails loudly if training steps stop reaching the tracker at all rather than passing for the wrong reason.

**M6** survived because I asserted "nothing non-`None` was written", which passes for an implementation that writes a `val_loss=None` row — a spurious metrics row that also clears the panel. It now asserts no row and no event *at all*, plus a separate test that a malformed later payload does not wipe an earlier real measurement.

I am reporting these rather than quietly fixing them because the first version of this suite would have shipped with your own finding uncovered.

## Verification

```
pytest tests/test_issue23_mlx_val_loss_producer.py                    12 passed
pytest <producer + #713 suite + 8 MLX suites + startup guard>        391 passed
same, under a runner with NO MLX (which every CI runner is)           54 passed, 1 skipped
ruff check src/soup_cli/ tests/test_issue23_mlx_val_loss_producer.py  All checks passed!
git diff <base> -- tests/ | grep -c "^-[^-]"                          0
```

The Metal run above was re-executed against the final tree after the last test change, rather than carried forward from the earlier one.

## Scope

- **Only the producer.** No sink is touched — those are #713 and reviewing them here would be reviewing them twice.
- `grad_norm` stays absent on this backend, unchanged from #665's reasoning.
- No changelog fragment yet: the PR number does not exist until this is opened, and a guessed number collides. I will push one immediately.


MakazhanAlpamys:
@Srinivasan8888 — **#713 is merged (`ce89a5c`), so this is unblocked. It needs a rebase, and there is one specific hazard in it that a routine conflict resolution would walk straight into.**

## The state right now

```
mergeable: CONFLICTING   mergeStateStatus: DIRTY
```

#713 landed as a **squash**, so its individual commits are not ancestors of `main` and they will not drop out on their own the way you expected — git has no commit-identity link between your base commits and `ce89a5c`.

## The hazard

Your branch points at an earlier commit of #713, from before the last round. Comparing the two trees directly:

```
git diff origin/main..pr739 -- tests/test_val_loss_end_to_end.py | grep '^-' | grep 'def test'
-    def test_every_dataclass_field_survives_to_payload(self):
```

**Merging this as it stands would delete the M8 test** — the one thing I blocked #713 on, and the only test in that file that fails when a field is added to the dataclass and to `_ALLOWED_KEYS` while being absent from the `to_payload` tuple. Net, your branch is 26 lines behind `main` in that file.

So when you rebase: for `tests/test_val_loss_end_to_end.py`, `src/soup_cli/utils/sse_train_stream.py`, `experiment/tracker.py`, `monitoring/callback.py` and `monitoring/display.py`, **take `main`'s side** — that is your own final #713 work. Your genuinely new content is three files:

```
src/soup_cli/trainer/mlx_sft.py             +77
tests/test_issue23_mlx_val_loss_producer.py +260
changelog.d/0.74.0/739.added.md             +1
```

Also drop `changelog.d/0.74.0/713.added.md` from this branch — it is on `main` already.

A quick self-check after rebasing, worth running before you push: `git diff origin/main..HEAD --stat` should show those three files and nothing else. If `test_val_loss_end_to_end.py` appears with deletions, the M8 test went with it.

## On the stacking, since you offered

Branching off rather than duplicating was the right call and I would not have preferred one folded PR — the producer is genuinely separable and reviewing it on its own is easier. The only thing worth doing differently next time is expecting a **squash** at the base, since that is what this repo does.

Nothing else is outstanding. I have not reviewed the substance yet; the two things I will be looking at are the `it - 1` decision (recording upstream's own counter rather than silently correcting it is right, and I want to see the test that pins it) and whether the sticky-vs-measured split you built for #713 is genuinely enforced on this backend rather than asserted in the PR body. Push the rebase and I will review.

Srinivasan8888:
Rebased in `af8f1d33` (force-with-lease). **Your self-check passes exactly as you specified it, and I ran it before pushing rather than after.**

```
$ git diff origin/main..HEAD --stat
 changelog.d/0.74.0/739.added.md             |   1 +
 src/soup_cli/trainer/mlx_sft.py             |  77 ++++++++
 tests/test_issue23_mlx_val_loss_producer.py | 260 ++++++++++++++++++++++++++++
 3 files changed, 338 insertions(+)
```

Three files, no `test_val_loss_end_to_end.py`, and:

```
grep -c "def test_every_dataclass_field_survives_to_payload" tests/test_val_loss_end_to_end.py   -> 1
git diff origin/main..HEAD -- tests/ | grep -c "^-[^-]"                                          -> 0
```

**The hazard was real and I confirmed it before fixing it**, rather than taking it on trust — on the old head, `git diff origin/main..HEAD -- tests/test_val_loss_end_to_end.py | grep "^-.*def test"` returned `test_every_dataclass_field_survives_to_payload`. Merging that would have deleted the M8 guard, which is the one thing you blocked #713 on. Thank you for catching it; a routine `git rebase origin/main` with "keep mine" on the conflicted hunks walks straight into it.

## How I did it, since "take main's side" is easy to get subtly wrong by hand

Not a conflict resolution. I reset the branch to `origin/main` and cherry-picked my own two commits:

```
git checkout -B feat/mlx-val-loss-producer origin/main
git cherry-pick 764cda90 2cbbff2f     # applied clean, no conflicts
```

That makes "take `main`'s side for the five #713 files" structural rather than a judgement call per hunk — those files are simply never touched. `changelog.d/0.74.0/713.added.md` drops out for free, since it now comes from `main` rather than from my branch.

I kept the pre-rebase head as `backup/739-prerebase` locally in case you want to diff the two.

## Verification on the rebased tree

```
pytest <producer + val_loss_end_to_end + 634 + display bridge + startup>   121 passed
simulated no-mlx runner (producer + end_to_end)                            41 passed, 1 skipped
ruff check src/soup_cli/ tests/                                            All checks passed!
mergeable                                                                  MERGEABLE
```

## On the squash point

Taken, and it is the actual lesson here: I expected #713's commits to drop out as ancestors, and a squash gives git no commit-identity link to do that with. Next stacked branch I open, I will rebase onto the squashed commit rather than assuming the base disappears.

The two things you said you would look at are both pinned rather than asserted — `it - 1` has its own test recording upstream's counter, and the sticky-vs-measured split is asserted against the tracker on this backend. I will leave those for your review rather than pre-empting it here.

MakazhanAlpamys:
@Srinivasan8888 — **the rebase is clean, my specific hazard is fully cleared, and I verified that rather than taking the self-check output on trust. Blocking on two things, and both are inside the discipline this PR is explicitly about.**

## First: the rebase hazard is gone

My worry was that #713 landed as a squash, so its commits are not ancestors of `main` and a routine "keep mine" resolution would have taken `test_every_dataclass_field_survives_to_payload` out with it — the one thing #713 was blocked on.

```
git diff origin/main...af8f1d33 --stat
 changelog.d/0.74.0/739.added.md             |   1 +
 src/soup_cli/trainer/mlx_sft.py             |  77 ++++++++
 tests/test_issue23_mlx_val_loss_producer.py | 260 ++++++++++++++++++++++++++++
 3 files changed, 338 insertions(+)

git diff origin/main af8f1d33 -- tests/test_val_loss_end_to_end.py   -> 0 lines (byte-identical)
git show af8f1d33:tests/... | grep -c test_every_dataclass_field_survives_to_payload  -> 1
git diff origin/main...af8f1d33 -- tests/ | grep -c "^-[^-]"        -> 0
```

Three files, additions only, none of #713's five sink files touched. Use the **three-dot** form from here on — `main` has moved since you ran it and the two-dot form now shows #728's files as noise.

Two of the three things I said I would check hold up under mutation: `it - 1` is genuinely pinned, and the sticky-vs-measured split is genuinely enforced on the **tracker** and the **display**. Nine of eleven mutations die.

## BLOCKING 1 — the SSE sibling of the test you already fixed is vacuous

`test_training_steps_do_not_put_a_stale_value_on_the_wire` passes with and without the fix. Mutation: give the training-step SSE push `val_loss=type(self)._sticky_val_loss`, i.e. exactly the #713 defect on the wire.

```
tests/test_issue23_mlx_val_loss_producer.py:  11 passed, 1 skipped   <- BOTH
```

The single variable is the display, isolated four ways:

| | events pushed | carrying val_loss | test |
|---|---|---|---|
| no display, as shipped | 1 | 1 | pass |
| no display, **mutant** | 1 | 1 | **pass — bug live** |
| display, as shipped | 3 | 1 | pass |
| display, **mutant** | 3 | **3** | fail, correctly |

With no display, `on_train_loss_report` returns at the `if display is None` guard before either the tracker call or the SSE push, so **training steps push nothing at all**. Asserting "exactly one event carries `val_loss`" is then true for every possible implementation.

**You already know this, and wrote it down.** Twenty-three lines above, on the tracker sibling:

> `# A display is REQUIRED here, not incidental: on_train_loss_report returns early when display is None, so without one the tracker is never reached on a training step and this test cannot observe the defect it exists for. Found by mutation -- writing the sticky value to the tracker on every training step survived the first version.`

You found it, documented it, fixed it for the tracker, and did not carry it to the wire. One keyword plus the same sanity assert closes it:

```python
_run(..., train_reports=[_TRAIN_1, _TRAIN_2], display=_RecordingDisplay())
assert len(seen) > 1, "sanity: training steps must reach the wire at all"
```

## BLOCKING 2 — the val hook fabricates 5 of 9 rows in the `loss` column

`on_val_loss_report` issues its **own** `tracker.log_metrics(run_id, step, epoch, val_loss=...)`. `log_metrics` defaults `loss=0.0, lr=0.0, grad_norm=0.0, speed=0.0` — not `None` — so every evaluation appends a row whose training columns are invented. Reproduced against the real `ExperimentTracker` and a real SQLite file, at your own live shape of 5 evaluations and 4 training steps:

```
 step    loss       lr   speed  val_loss
   -1     0.0      0.0     0.0      4.75   <- fabricated, and step is negative
    1     0.0      0.0     0.0       3.1   <- fabricated
    2     3.0   0.0001     5.0      None
    3     0.0      0.0     0.0      2.75   <- fabricated
    4     2.4   0.0001     5.0      None
    5     0.0      0.0     0.0       2.3   <- fabricated
    6     2.0   0.0001     5.0      None
    7     0.0      0.0     0.0      2.05   <- fabricated
    8     1.8   0.0001     5.0      None

total rows: 9 | rows carrying val_loss: 5 (correct) | rows with loss == 0.0: 5

loss series a consumer reads back:
  post-PR      : [0.0, 0.0, 3.0, 0.0, 2.4, 0.0, 2.0, 0.0, 1.8]
  training truth: [3.0, 2.4, 2.0, 1.8]
```

Your headline is `"no fabricated rows: True (db rows == sse events == evaluations)"`. That is **true for the `val_loss` column and false one column to the left** — over half the training-loss curve is now zeros that no step measured. It is the same defect class #713 was blocked on, and `log_metrics`'s own docstring makes the argument for me:

> `val_loss` defaults to `None` rather than `0.0`: most rows are training steps with no evaluation attached, and a zero there would be indistinguishable from a genuinely measured zero.

**The merged transformers producer does not do this.** `callback.py` carries `_last_loss` / `_last_lr` / `_last_grad_norm` and folds `val_loss` into the *same* `log_metrics` call, so an evaluation creates no extra row. Mirroring that — attach the measured `val_loss` to the next training row, or carry `loss`/`lr` into the val row — is the fix.

Your suite cannot see it because `_RecordingTracker` is only ever interrogated about `val_loss` and never about `loss`.

Method note: my reproduction calls `log_metrics` with the hook's argument list taken verbatim from `mlx_sft.py:411-416`, against the real tracker. It is not a live MLX run — this box has no Metal.

## Non-blocking

**The float coercion is untested.** Mutation: remove `_as_float`. Nothing fails. It is not cosmetic — an object that is `float()`-able but not JSON-serialisable reaches `to_payload()` intact and `json.dumps` raises, which is precisely the SSE failure the comment says it prevents. Today's mlx-lm returns a real float via `.item()`, so this is LOW, but the guard is currently decorative and one test would fix that.

Also worth a line in a comment: the `step = -1` row is upstream's `it - 1` at `it = 0`, which you record deliberately. It is defensible, but it is a negative step in a table users read, and nothing currently says so.

## Everything else is clean

`changelog.d/0.74.0/739.added.md` is named after the PR, not the issue, and credits you. No tests deleted or weakened. Ruff clean. `11 passed, 1 skipped` on a box with `find_spec("mlx") is None`, so these tests can fail on CI — the skip is the `mlx_lm` upstream-contract one, which is right.

Both fixes are small. Take them in either order.

Srinivasan8888:
Both fixed in `2d9555a6`, plus both LOW items. **And one correction to the review, measured rather than argued.**

## BLOCKING 1 — you found the third instance of a trap I had already documented

The test took no display, so `on_train_loss_report` returned at the `display is None` guard before the SSE push and training steps pushed nothing at all. Your four-way isolation table is exactly right.

What stings is that the fix is written out in my own words twenty-three lines above, on the tracker sibling, ending *"Found by mutation."* I found it, wrote it down, fixed it for the tracker, and did not carry it to the wire. The test now takes a `_RecordingDisplay()` and asserts `len(seen) > 1` first. Your mutation — `val_loss=type(self)._sticky_val_loss` on the training push — now fails it.

## BLOCKING 2 — real, and worse than the val_loss column

Reproduced at your shape against a real `ExperimentTracker` and a real SQLite file, then fixed:

```
                        rows with loss == 0.0
before                  5 of 9
after                   2 of 9
```

The two that remain are the evaluations that fire **before any training step** (steps -1 and 1), where 0.0 is the initial value rather than a fabrication — and the transformers callback reports the same at the same point (`callback.py:61-63`). Pinned by its own test so it cannot quietly start inventing a number there instead.

## The correction

> `callback.py` carries `_last_loss` / `_last_lr` / `_last_grad_norm` and folds `val_loss` into the *same* `log_metrics` call, so an evaluation creates no extra row.

It does create one. Driving the real `SoupTrainerCallback` against a real tracker, three training logs and one eval log:

```
(1, 3.0, 0.0001, None)
(2, 2.4, 0.0001, None)
(2, 2.4, 0.0001, 2.75)   <- the eval, its own row, carrying loss and lr
(3, 2.0, 0.0001, None)
row count: 4 for 3 training logs + 1 eval log
```

An evaluation arrives as its own `on_log` call, so it gets its own row — with **carried** values, not zeros. Which means of the two fixes you offered, **carrying** is the faithful mirror and folding would have made MLX diverge from the merged backend. That is the one I pushed. Everything else in your diagnosis stands; this only changes which remedy matches.

## The placement has its own test

The carried values are assigned **above** the display guard. Moving them below survives all fourteen other tests in the file and fails only the no-display one — because a `soup train` without a live panel would otherwise write a fabricated `loss` on every validation row while the suite stayed green. Third instance of the same trap, so this time it is pinned rather than commented.

## Both LOW items taken

`_as_float` was decorative — removing it failed nothing. It now has a test that drives a `float()`-able, non-JSON-serialisable object all the way through `format_sse_frame`, with a control that junk is dropped rather than coerced into a measurement. Removing the guard: 2 failed. And `step = -1` is now commented where it is read, saying that it is upstream's `it - 1` at `it = 0` and that it surfaces in a table users see.

```
pytest <producer + end_to_end + 634 + display bridge + startup + v07133>  154 passed, 1 skipped
mutations (binary mode, sha256 checked before and after restore):
    sticky val_loss on the wire      1 failed
    revert the carried values        2 failed
    assignment below the guard       1 failed
    remove _as_float                 2 failed
ruff                                 All checks passed!
git diff origin/main...HEAD -- tests/ | grep -c "^-[^-]"    0
```

Three-dot form from here on, as you said — noted that `main` has moved.

Nothing here was run on Metal; this is the fake-mlx harness and a real tracker on a box with `find_spec("mlx") is None`.

MakazhanAlpamys:
@Srinivasan8888 — merged as `6b999eb8`. **Both blockers are closed, I re-ran them myself, and you were right about my review being wrong. That correction goes first, because I published the claim and it should not stand.**

## I was wrong about the transformers path

I wrote that `callback.py` "folds `val_loss` into the *same* `log_metrics` call, creating no extra row." That is false, and your correction is the accurate description.

`on_log` has exactly one `log_metrics` call, at its end — and an evaluation arrives as its **own** `on_log` invocation. So the transformers path does create a row for the evaluation. What it does not do is fabricate the training columns: it carries `_last_loss` / `_last_lr` / `_last_grad_norm` into that row.

So the defect I found was real, but the remedy I proposed was not the faithful mirror — **folding** would have made MLX diverge from the merged backend, and **carrying** is what matches it. You pushed the right fix, and you pushed it against a review that named the wrong one. Thank you for measuring it instead of doing what I asked.

## Blocking 1 — verified dead

```
baseline                                          17 passed, 1 skipped
sticky val_loss on the training-step SSE event     1 failed, 16 passed, 1 skipped

FAILED TestOnlyRealMeasurementsAreRecorded::test_training_steps_do_not_put_a_stale_value_on_the_wire
```

Binary mode, occurrence count asserted at 1, sha `931a8066` -> `db5d2614`. That is the test that previously passed with and without the fix, and it is now the one that fails. Adding the recording display was the whole repair, and it is the same shape you had already caught and documented on the tracker sibling.

## Blocking 2 — verified closed

The val row now carries the last measured training values, so the `loss` series stops carrying invented zeros. The row an evaluation before the first training step writes still carries `0.0` — and that is correct rather than a leftover, because there is no measured loss to carry yet and the transformers path does exactly the same thing at the same point. Your comment in the source says so and cites the line numbers, which is why I could check it rather than take it.

Both LOW items are genuinely closed too: the float coercion killed nothing before and now has two tests that kill its removal, and the `step = -1` origin is commented where it is read.

## The rest

Eight mutations at merge, eight killed, on an interpreter with no mlx. Rebase verified through the three-dot diff — three files, 566 insertions, zero deletions, none of #713's sink files touched, and `tests/test_val_loss_end_to_end.py` byte-identical to `main` with `test_every_dataclass_field_survives_to_payload` intact, which was the specific hazard I raised. Fragment named after the PR, crediting you. Ruff clean. 14 of 14 SUCCESS on `2d9555a6`, `head_sha` confirmed equal.

One thing to fix whenever convenient, not blocking and not worth a round on its own: the fragment still says *"5 database rows ... for 5 evaluations"*, written in `af8f1d3` before the carrying change. Five rows still **carry val_loss**, so it is not wrong, but a run in that shape now writes nine rows total and the sentence reads narrower than it is. If you touch this file again, that is the line.

PR #740, opened 2026-09-07
fix(mlx): honour train_on_responses_only and gradient_checkpointing

mraad:
## What

Two `SoupConfig` fields that the MLX SFT trainer has an mlx-lm home for were never passed through, so both were silently ignored on `backend: mlx`.

### 1. `data.train_on_responses_only` did nothing on MLX

It defaults to `True` and every transformers trainer honours it. mlx-lm reads the value off the args object:

```python
# mlx_lm/tuner/datasets.py
mask_prompt = getattr(config, "mask_prompt", False)
```

`TrainingArgs` declares no `mask_prompt` field, so that `getattr` always returned `False` and **every MLX SFT run trained on the prompt**. There is no warning, and the loss curve looks normal.

This is invisible on a short chat dataset. It is ruinous on one with a long shared system prompt — a schema, a retrieved document, a style guide — where the prompt is most of the tokens and therefore most of the loss. `TrainingArgs` is a plain dataclass, so the attribute the loader looks for is set on the instance.

### 2. `training.gradient_checkpointing` was hardcoded `False`

`TrainingArgs.grad_checkpoint` exists upstream; the trainer passed a literal `False`. That is the one knob that makes a long-prompt dataset fit in Metal memory, and it was unreachable from config.

mlx-lm checkpoints every block or none, so the tier strings (`selective` / `medium` / `full`) have no MLX equivalent and all mean "on". A warning says so, rather than letting a tier silently read as off.

### 3. `adapter_config.json` recorded both as literal `False`

So neither drift was detectable after a run. It now records the effective values, matching the policy #684 established for `grad_accumulation_steps`.

### 4. `data.train_on_messages_with_train_field`

Now declared unsupported in `_check_unsupported()` instead of ignored — mlx-lm masks whole prompts, not per message.

## Why it matters, concretely

Found while fine-tuning a 7B QLoRA on Apple Silicon with a ~7.3k-token schema in the system prompt and ~40-token answers. With masking off, ~99% of the loss was schema text the model must not learn to reproduce. The run then died before step 1:

```
Error: RuntimeError: [METAL] Command buffer execution failed: Insufficient
Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory).
```

`gradient_checkpointing: true` was already in the config and was doing nothing. With both fixes the same config trains, and the written `adapter_config.json` reads:

```json
"iters": 340,
"mask_prompt": true,
"grad_checkpoint": false,
```

## Tests

`tests/test_mlx_masking_and_grad_checkpoint.py` — 7 tests, reusing the fake-mlx harness from `test_issue684_mlx_grad_accumulation.py` rather than duplicating it. The masking tests assert against the args object mlx-lm's own `create_dataset` receives, since that is the only place the value has any effect.

All 7 fail on the pre-fix code (verified by stashing the `mlx_sft.py` change and re-running). The full MLX suite is green:

```
$ pytest tests/test_mlx*.py tests/test_issue*mlx*.py -q
322 passed

$ ruff check src/soup_cli/ tests/
All checks passed!
```

Same caveat as #684, stated in the module docstring: the fakes exercise real control flow through `train()`, but they stand in for mlx-lm behaviour. The end-to-end confirmation above is a real run on an M4 Max.

https://claude.ai/code/session_01Xwyk8GqAaFd4QMio53yqKE

### DIFF
--- a/src/soup_cli/trainer/mlx_sft.py
+++ b/src/soup_cli/trainer/mlx_sft.py
@@ -149,6 +149,17 @@ class MLXSFTTrainerWrapper:
             unsupported.append("training.seed (MLX seeds through mx.random)")
         if tcfg.data_seed is not None:
             unsupported.append("training.data_seed (MLX seeds through mx.random)")
+        if isinstance(tcfg.gradient_checkpointing, str):
+            console.print(
+                "[yellow]MLX backend: gradient_checkpointing tier "
+                f"'{tcfg.gradient_checkpointing}' means every block — mlx-lm "
+                "checkpoints all or nothing.[/]"
+            )
+        if self.config.data.train_on_messages_with_train_field:
+            unsupported.append(
+                "data.train_on_messages_with_train_field "
+                "(mlx-lm masks whole prompts only, not per message)"
+            )
         if unsupported:
             console.print(
                 "[yellow]MLX backend ignores: " + ", ".join(unsupported) + "[/]"
@@ -300,6 +311,7 @@ class MLXSFTTrainerWrapper:
         # irrelevant (val_dataset stays None and mlx-lm skips evaluation).
         steps_per_eval = max(1, iters // 4) if val_rows else max(1000, iters + 1)
         grad_accumulation_steps = int(cfg.training.gradient_accumulation_steps)
+        grad_checkpoint = bool(cfg.training.gradient_checkpointing)
         # mlx-lm updates the optimizer only when it % accum == 0 (trainer.py) and
         # never flushes a partial group, so round iters down to a whole number of
         # groups, keeping at least one group so a small dataset does not train
@@ -319,6 +331,22 @@ class MLXSFTTrainerWrapper:
             adapter_file=str(output_dir / "adapters.safetensors"),
             grad_accumulation_steps=grad_accumulation_steps,
         )
+        # `data.train_on_responses_only` defaults to True and every transformers
+        # trainer honours it, but MLX read it from nothing: mlx-lm's
+        # `create_dataset` pulls `mask_prompt` off the args object with a
+        # getattr default of False, and `TrainingArgs` has no such field. Every
+        # MLX SFT run therefore trained on the prompt too — invisible on a chat
+        # dataset, ruinous on one with a long shared system prompt, where the
+        # prompt tokens are most of the loss. TrainingArgs is a plain dataclass,
+        # so the attribute the loader looks for is set here.
+        mask_prompt = bool(cfg.data.train_on_responses_only)
+        args.mask_prompt = mask_prompt
+        # Same shape of gap for `training.gradient_checkpointing`: mlx-lm's
+        # TrainingArgs has the field, and this trainer passed a hardcoded False,
+        # so the only knob that makes a long-prompt dataset fit in Metal memory
+        # did nothing on this backend. mlx-lm checkpoints every block or none —
+        # the tier strings have no MLX equivalent and all mean "on" here.
+        args.grad_checkpoint = grad_checkpoint
 
         train_dataset = CacheDataset(create_dataset(train_rows, self.tokenizer, args))
         val_dataset = (
@@ -446,8 +474,8 @@ class MLXSFTTrainerWrapper:
                     "lora_parameters": build_mlx_adapter_config(
                         lora_cfg, adapter_path=str(output_dir)
                     )["lora_parameters"],
-                    "mask_prompt": False,
-                    "grad_checkpoint": False,
+                    "mask_prompt": mask_prompt,
+                    "grad_checkpoint": grad_checkpoint,
                     "grad_accumulation_steps": grad_accumulation_steps,
                 },
                 indent=2,
--- a/changelog.d/0.74.0/740.fixed.md
+++ b/changelog.d/0.74.0/740.fixed.md
@@ -1,0 +1,21 @@
+- **Two MLX SFT knobs were validated, documented, and then read by nothing (#740 by
+  @mraad).**
+  `data.train_on_responses_only` defaults to True and every transformers trainer honours
+  it, but the MLX path never set it: mlx-lm's `create_dataset` pulls `mask_prompt` off the
+  args object with a `getattr` default of `False`, and `TrainingArgs` declares no such
+  field, so **every MLX SFT run trained on the prompt** — no warning, and a normal-looking
+  loss curve. Invisible on a short chat dataset; ruinous on one with a long shared system
+  prompt, where the prompt is most of the tokens and therefore most of the loss.
+  `training.gradient_checkpointing` had the same shape: `TrainingArgs.grad_checkpoint`
+  exists upstream and the trainer passed a hardcoded `False`, so the one knob that makes a
+  long-prompt dataset fit in Metal memory was unreachable from config. Found on a 7B QLoRA
+  run with a 7.3k-token schema in the system prompt, which died at `[METAL] Command buffer
+  execution failed: Insufficient Memory` with `gradient_checkpointing: true` already set
+  and doing nothing. mlx-lm checkpoints every block or none, so the tier strings
+  (`selective` / `medium` / `full`) have no MLX equivalent and all mean "on"; a warning
+  says so rather than letting a tier read as off. `adapter_config.json` wrote both values
+  as literal `False`, so neither drift was detectable after a run — it now records what
+  training actually used, matching the policy #684 established for
+  `grad_accumulation_steps`. `data.train_on_messages_with_train_field` is additionally
+  declared unsupported here instead of ignored, since mlx-lm masks whole prompts rather
+  than individual messages.
--- a/tests/test_mlx_masking_and_grad_checkpoint.py
+++ b/tests/test_mlx_masking_and_grad_checkpoint.py
@@ -1,0 +1,128 @@
+"""MLX SFT silently ignored two schema fields it has an mlx-lm home for.
+
+``data.train_on_responses_only`` defaults to True and every transformers
+trainer honours it, but the MLX path never set it: mlx-lm's
+``create_dataset`` reads ``mask_prompt`` off the args object with a
+``getattr`` default of False, and ``TrainingArgs`` declares no such field,
+so the prompt was always trained on. On a chat dataset that is invisible;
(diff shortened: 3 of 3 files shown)


MakazhanAlpamys:
@mraad — welcome, and thank you. **You have walked into a collision on both halves of this PR, and on one of them the reason you could not have seen it is my failure, not yours.** I am telling you now rather than after I finish reading, because you deserve to know before you spend more time on it.

## The state of the two halves

**`gradient_checkpointing` — already fixed and merged.** `9b58300` by @AmirF194 in #698, landed today at 03:38 UTC, closing #685. That is roughly eight hours before you opened this, so unless your branch is based on an older `main` you may not have had it.

**`train_on_responses_only` — claimed and in review.** @Srinivasan8888 claimed #683 on 2026-09-06 at 15:26 UTC and #733 has been open since, currently in its third round.

## What is my fault

**#683 was labelled `help wanted`, had no assignee, and carried the claim only inside the comment thread.** If you were scanning open issues — or scanning the code and finding a real bug, which is what your reproduction reads like — you would have seen an unclaimed bug in a repo asking for help.

That is worse than an ordinary oversight, because I introduced a `claimed` label **for exactly this**, four days ago, after the fifth collision, and wrote at the time that a label is the mechanism because comments and assignees are not visible in the issue *list*. I then did not apply it to #683. I have just fixed that across every issue with an open PR against it — #683, #686, #697, #716, #719, #720, #721, #722, #723 and #23 — so the next person scanning sees what you could not.

## What happens now, and it is not "close it"

My standing rule here is that the newcomer's pull request gets read **first**, and if it solves something the incumbent's does not, I say so on their thread and credit you for it. I am doing that now rather than deciding by timestamp. Specifically I am checking whether your `args.mask_prompt = True` route covers a case #733's per-token mask does not, and whether your `gradient_checkpointing` half differs in any way from what already merged.

I will post the result here either way, with the evidence.

## Two things about your report that are worth more than the patch

**Your reproduction is the strongest statement of why this matters that anyone has written on it.** A ~7.3k-token schema in the system prompt against ~40-token answers, so about 99% of the loss is text the model must not learn to reproduce — and then a real `kIOGPUCommandBufferCallbackErrorOutOfMemory` on an M4 Max. #683 has been open with a description; you have it with a number and a dead run. That goes in the record regardless of whose code ships.

**You noticed the config was lying.** `"grad_checkpoint": false` written into `adapter_config.json` while the config said `true` is the shape of bug this project has shipped more times than any other — a setting that validates, documents, and reaches nothing.

## Two mechanical notes while I read

- This PR is currently **CONFLICTING** with `main`, and no CI has run on it at all — GitHub cannot build the merge commit for a conflicted branch, so there is nothing for me to approve. If the work continues, it will need a rebase before anything can be measured.
- Your caveat about the fakes standing in for mlx-lm behaviour is the right thing to state, and stating it unprompted counts for a lot here.

Give me a short while and I will come back with specifics rather than a verdict.

MakazhanAlpamys:
@mraad — **I read it, as promised. Your diagnosis is right and I am not going to merge the patch, and I owe you the measurements rather than a verdict. There is also one piece of this I do want, and a way for you to land it.**

## Your diagnosis is correct, and it is the hard part

> `TrainingArgs` declares no `mask_prompt` field, so that `getattr` always returned `False` and **every MLX SFT run trained on the prompt**.

That is exactly right, it is issue #683, and I confirmed the line myself from the mlx-lm 0.31.3 source rather than from memory — `tuner/datasets.py:180`, `mask_prompt = getattr(config, "mask_prompt", False)`.

## Where the route in cannot be `args.mask_prompt = True`

Your `mlx_sft.py:342-343` sets it unconditionally, for every row shape. Two consequences, both measured, both from the same upstream file.

**1. Plain-text rows would crash — on the schema default.** `datasets.py:194-197`:

```python
    elif text_feature in sample:
        if mask_prompt:
            raise ValueError("Prompt masking not supported for text dataset.")
        return TextDataset(data, tokenizer, text_key=text_feature)
```

`backend: mlx` + `task: sft` + `data.format: plaintext` parses today with no gate — I ran it through the real loader:

```
parses OK: backend=mlx task=sft format=plaintext train_on_responses_only=True
```

`train_on_responses_only` defaults to `True`, so this PR turns a silently-wrong run into a hard `ValueError` for every plaintext MLX SFT user, without them changing anything.

**And your suite cannot see it, which is the part worth taking away.** `tests/test_mlx_masking_and_grad_checkpoint.py:50` sets the fixture rows to `{"text": "hi"}` — the exact shape that raises — and the fake you reuse is `create_dataset = lambda rows, tokenizer, args: rows` (`test_issue684_mlx_grad_accumulation.py:49`), a pass-through that cannot raise. So the suite is green on the one input that breaks. That is not a slip particular to you; it is the failure mode this repo keeps producing, and it is why I ask every fix here to be mutation-tested against the *behaviour* rather than the wiring.

**2. On multi-turn chat it produces a different wrong distribution, not a smaller right one.** `datasets.py:65-75` computes one contiguous offset from `messages[:-1]`, so supervision is the **last message only**. Measured on a real tokenizer, 5-message conversation, two assistant turns:

```
total rendered tokens                 : 97
flag off (today's bug)                : 97 supervised
upstream mask_prompt=True (this PR)   : 28 supervised
per-token mask (#733)                 : 49 supervised
```

21 of 49 assistant tokens dropped — 43% — while the loss curve looks fine. #683's acceptance criteria say every assistant turn must stay supervised in multi-turn conversations, so this is the criterion the flag fails rather than meets.

@Srinivasan8888 claimed #683 on 2026-09-06 and #733 handles all three row shapes separately: per-token mask for chat, upstream's flag for prompt/completion (where the single prefix genuinely is the prompt), and a warning for plain text — with a test that calls upstream's own `create_dataset` and asserts that `ValueError`.

## The gradient_checkpointing half — already on `main`, proven by your own tests

I copied your test file onto current `main`, unmodified, and ran it:

```
3 failed, 4 passed
```

**All four of your `TestGradientCheckpointingReachesMlxLm` tests pass on `main` already**, including the `adapter_config.json` one. Only the three masking tests fail. `9b58300` (#698) landed one commit after your merge-base — you branched a few hours early and nobody did anything wrong.

One difference worth knowing, because it is a technique rather than a preference: the merged version passes `grad_checkpoint=` as a **constructor kwarg**, so an upstream field rename raises `TypeError` immediately; setting it as a post-construction attribute, as yours does, would silently no-op — which is the same class of bug you are fixing here.

A small honesty note on your report: the `adapter_config.json` you pasted reads `"grad_checkpoint": false`, so that artifact does not actually corroborate the checkpointing half. Not a problem, just worth knowing before someone else points at it.

## What I do want from this PR, and it is yours to land

`main` currently prints this:

```
MLX backend ignores: gradient_checkpointing tier 'selective' (MLX has a single on/off switch; enabling it)
```

It says the option is ignored and then says it is enabled. Yours replaces it with its own line and no false claim:

```
MLX backend: gradient_checkpointing tier '<tier>' means every block — mlx-lm checkpoints all or nothing.
```

That is a real fix to operator-facing output and I would rather you shipped it than I did. **If you reduce this PR to that single message change** — rebase onto `main`, drop the masking work and the now-redundant checkpointing wiring, keep or trim the changelog fragment accordingly, add a test asserting the message no longer contains "ignores" for the tier case — I will review it the same day and merge it on green.

## What goes on the record either way

Your field report is the best real-hardware evidence anyone has attached to this family of bugs: a ~7.3k-token schema against ~40-token answers, ~99% of the loss on text the model must not reproduce, and a real `kIOGPUCommandBufferCallbackErrorOutOfMemory` on an M4 Max. #683 had a description; you gave it a dead run. I am quoting it in #733's merge notes and crediting you there.

Also worth saying: zero deleted test lines, ruff clean, a correctly formatted per-PR changelog fragment, seven tests, and a stated caveat about the fakes standing in for mlx-lm behaviour — unprompted. That is a better first pull request than most, and the thing that went wrong was a claim signal I failed to publish, not your judgement.

src/soup_cli/trainer/mlx_sft.py
```diff
@@ -300,6 +311,7 @@ def train(self, display=None, tracker=None, run_id=None, resume_from_checkpoint=
         # irrelevant (val_dataset stays None and mlx-lm skips evaluation).
         steps_per_eval = max(1, iters // 4) if val_rows else max(1000, iters + 1)
         grad_accumulation_steps = int(cfg.training.gradient_accumulation_steps)
+        grad_checkpoint = bool(cfg.training.gradient_checkpointing)
```
AmirF194, line 314: Checked this against current main. The gradient_checkpointing half is already fixed and merged in #698 (2026-09-07), so this branch is off the pre-698 blob and will conflict on rebase.

`ead85d8e` is the blob both this diff and #698's diff show as "before" for mlx_sft.py, and it still hardcodes `grad_checkpoint: False`. Main at `9b58300` already wires `grad_checkpoint = bool(cfg.training.gradient_checkpointing)` through TrainingArgs and adapter_config.json, the same mechanism added here at line 314.

The mask_prompt / train_on_responses_only fix is still real and unfixed on main (`"mask_prompt": False` is still hardcoded there). Rebasing and dropping the grad_checkpoint half avoids the conflict and keeps the part that's still needed.

mraad:
You're right on `grad_checkpoint` — #698 (`9b58300`) beat this branch to it, and that's the conflict.

The masking half is fixed too, just after the commit you checked. #733 merged as `f71aef0` on 2026-09-08, 17 commits past `9b58300`. Main now has `src/soup_cli/trainer/mlx_masking.py` and sets `args.mask_prompt = plan.mask_prompt` (line 377), writing `"mask_prompt": bool(args.mask_prompt)` (line 551) — the hardcoded `False` at old line 461 is gone.

It also supersedes the approach here rather than just landing it. Setting `args.mask_prompt`, which is what this PR does, is wrong for multi-turn chat: upstream's `ChatDataset` masks a single prefix ending before the last message, so it supervises only the final assistant turn (#733 measured 772 supervised tokens correct vs 71 with the flag). #733 injects a real per-token mask through `train(loss=..., iterate_batches=...)` instead.

Merged main in to confirm: `git diff main HEAD` is empty. Nothing left here. Closing as superseded by #733 and #698.

AmirF194:
Thanks for the confirmation and for tracing the masking gap further than I did. Good that both are covered by #698 and #733.

PR #741, opened 2026-09-07
fix(ppo): forward the configured KL coefficient and epoch budget (closes #721)

### LINKED ISSUE #721: PPO drops the configured KL coefficient and epoch budget on TRL 0.29

dchaudhari7177:
Closes #721.

## The two losses

**`ppo_kl_penalty`.** `setup()` resolved it as `init_kl_coef` only. trl renamed that to `kl_coef`, and a keyword `PPOConfig` does not recognise was never an error here -- it was dropped. So on trl 0.29 the run used trl's own `0.05` and nothing said otherwise, which is the part that makes this worse than an ordinary rename: the optimization objective changed silently.

**`training.epochs`.** Saved to `self._num_epochs` for the legacy manual loop and never passed to `PPOConfig`, so the built-in path ran trl's default budget. Every other trainer in `trainer/` passes `num_train_epochs=tcfg.epochs`; PPO was the exception.

`ppo_epochs` is a different setting -- optimization passes per batch -- and was already correct as `num_ppo_epochs`. Untouched, and there is a test that keeps it that way.

## Where the fix lives

`_trl_compat.kl_penalty_kwargs(config_cls, kl_penalty)`, next to `prompt_length_kwargs`, which exists for exactly this class of rename. It asks the class the caller is about to construct (so a namespace move is handled too), prefers `kl_coef`, falls back to `init_kl_coef`, and returns `{}` if neither exists.

The empty-dict case is deliberate. A third rename then produces a config that visibly ignores the setting at one call site, rather than repeating this bug one version later.

## Tests

`tests/test_issue721_ppo_kl_and_epochs.py`, 7 cases. trl is not a test dependency, so the kwarg-selection half is exercised against synthetic config classes carrying each signature -- modern (`kl_coef`), legacy (`init_kl_coef`), neither, and both at once -- and the epoch half is read statically out of `ppo.py`, in the style of `test_issue323_trl_kwarg_drift.py`.

Without the fix the module does not even import (`cannot import name 'kl_penalty_kwargs'`), which is the red state.

```
tests/test_issue721_ppo_kl_and_epochs.py                    7 passed

tests/test_ppo.py + test_trl_version_compat.py + test_issue323_trl_kwarg_drift.py
  with this change:   8 failed, 95 passed, 10 skipped
  on main:            8 failed, 88 passed, 10 skipped
```

The 8 are the same failures either way: `ModuleNotFoundError: No module named 'trl'` in `TestResolveTrlSymbol`, because trl is not installed in this environment. Nothing in this change touches them.

## Not verified here

I could not construct a real `PPOConfig` locally, so the "resolved built-in contract" table in the issue is not something I re-measured against trl 0.29.1 -- the synthetic signatures encode it instead. If CI has trl installed, `test_issue323_trl_kwarg_drift.py` should confirm the real class accepts both keywords this now passes.

### DIFF
--- a/src/soup_cli/trainer/_trl_compat.py
+++ b/src/soup_cli/trainer/_trl_compat.py
@@ -120,6 +120,21 @@ def prompt_length_kwargs(config_cls: type, max_prompt_length: int) -> dict[str, 
     return {}
 
 
+def kl_penalty_kwargs(config_cls: type, kl_penalty: float) -> dict[str, float]:
+    """``{'kl_coef': x}`` or ``{'init_kl_coef': x}``, whichever this trl takes.
+
+    trl renamed ``init_kl_coef`` to ``kl_coef``. Asking for the new name first
+    matters more than the usual rename: an unrecognised keyword here was never
+    an error, it was silently dropped, so a PPO run kept going with trl's own
+    default KL strength rather than the configured one. Empty dict if neither
+    name exists, so a third rename fails the same visible way this one did not.
+    """
+    for name in ("kl_coef", "init_kl_coef"):
+        if config_accepts(config_cls, name):
+            return {name: kl_penalty}
+    return {}
+
+
 def _truncate_tokens(tokens: list[int], limit: int, mode: str) -> list[int]:
     """Truncate one token sequence using TRL's preference-side convention."""
     if limit <= 0:
--- a/src/soup_cli/trainer/ppo.py
+++ b/src/soup_cli/trainer/ppo.py
@@ -78,6 +78,8 @@ class PPOTrainerWrapper:
         from datasets import Dataset
 
         # Import PPOTrainer/PPOConfig — trl >=0.28 moved to trl.experimental
+        from soup_cli.trainer._trl_compat import kl_penalty_kwargs
+
         ppo_trainer_cls, ppo_config_cls, is_experimental = _import_ppo_classes()
 
         # Enable Rich progress bar for HuggingFace downloads
@@ -176,8 +178,17 @@ class PPOTrainerWrapper:
 
         if "cliprange" in ppo_params:
             ppo_kwargs["cliprange"] = tcfg.ppo_clip_ratio
-        if "init_kl_coef" in ppo_params:
-            ppo_kwargs["init_kl_coef"] = tcfg.ppo_kl_penalty
+
+        # trl renamed init_kl_coef -> kl_coef. Only init_kl_coef was checked, so
+        # on trl >= 0.29 the configured penalty was dropped and PPO silently ran
+        # at the library default.
+        ppo_kwargs.update(kl_penalty_kwargs(ppo_config_cls, tcfg.ppo_kl_penalty))
+
+        # training.epochs was only reaching the legacy manual loop (self._num_epochs),
+        # so the built-in trainer ran trl's default budget. ppo_epochs above is a
+        # different setting: optimization passes per batch, not dataset epochs.
+        if "num_train_epochs" in ppo_params:
+            ppo_kwargs["num_train_epochs"] = tcfg.epochs
 
         # Optional params that may not exist in all trl versions
         if "log_with" in ppo_params:
--- a/tests/test_issue721_ppo_kl_and_epochs.py
+++ b/tests/test_issue721_ppo_kl_and_epochs.py
@@ -1,0 +1,109 @@
+"""Issue #721 - PPO must forward the configured KL strength and epoch budget.
+
+Two values were resolved by name against the installed trl and lost when the
+name did not match:
+
+* ``training.ppo_kl_penalty`` was passed only as ``init_kl_coef``. trl 0.29
+  renamed that to ``kl_coef``, and an unrecognised keyword here was never an
+  error - it was dropped, so PPO ran at trl's own default KL strength with
+  nothing said.
+* ``training.epochs`` reached only the legacy manual loop, never
+  ``num_train_epochs`` on the built-in ``PPOConfig`` path. ``ppo_epochs`` is a
+  different setting (optimization passes per batch) and was always correct.
+
+trl is not a test dependency, so these ask the question without it: stand in
+synthetic config classes with each signature and read what comes back. The
+kwarg-selection half lives in ``_trl_compat.kl_penalty_kwargs`` so it can be
+called directly, next to ``prompt_length_kwargs``, which exists for the same
+kind of rename.
+"""
+
+from __future__ import annotations
+
+import ast
+import pathlib
+
+from soup_cli.trainer._trl_compat import kl_penalty_kwargs
+
+_PPO_SOURCE = (
+    pathlib.Path(__file__).resolve().parents[1]
+    / "src"
+    / "soup_cli"
+    / "trainer"
+    / "ppo.py"
+).read_text(encoding="utf-8")
+
+
+class _ModernConfig:
+    """trl >= 0.29: the coefficient is ``kl_coef``."""
+
+    def __init__(self, kl_coef: float = 0.05, num_train_epochs: float = 3.0):
+        pass
+
+
+class _LegacyConfig:
+    """Older trl: the coefficient is ``init_kl_coef``."""
+
+    def __init__(self, init_kl_coef: float = 0.2):
+        pass
+
+
+class _NeitherConfig:
+    """A hypothetical trl that renamed it again."""
+
+    def __init__(self, cliprange: float = 0.2):
+        pass
+
+
+class TestKlPenaltyKwargs:
+    def test_modern_trl_gets_kl_coef(self):
+        assert kl_penalty_kwargs(_ModernConfig, 0.7) == {"kl_coef": 0.7}
+
+    def test_legacy_trl_still_gets_init_kl_coef(self):
+        assert kl_penalty_kwargs(_LegacyConfig, 0.7) == {"init_kl_coef": 0.7}
+
+    def test_a_config_with_neither_name_gets_nothing(self):
+        # Empty rather than a guess, so the next rename fails visibly at the
+        # call site instead of being silently dropped the way this one was.
+        assert kl_penalty_kwargs(_NeitherConfig, 0.7) == {}
+
+    def test_the_new_name_wins_when_a_config_carries_both(self):
+        class _Both:
+            def __init__(self, kl_coef: float = 0.05, init_kl_coef: float = 0.2):
+                pass
+
+        assert kl_penalty_kwargs(_Both, 0.7) == {"kl_coef": 0.7}
+
+
+class TestPpoForwardsTheEpochBudget:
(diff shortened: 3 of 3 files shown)


MakazhanAlpamys:
@dchaudhari7177 — **you have walked into a collision, you have had 22 hours of silence from me while doing it, and the reason you could not have seen it coming is a process failure of mine. I am telling you now, before I finish reviewing, because you should not spend another evening on this without knowing.**

## The situation, stated plainly

There is a second pull request against #721: **#737 by @be-student**, opened `2026-09-06T18:50:51Z`, which is **18 hours and 23 minutes before yours**. I reviewed it on 2026-09-07 at 06:47 UTC and wrote that it was correct and that two small things stood between it and a merge.

Then, at 13:13 UTC the same day, you opened this. And I said nothing to either of you about the other.

## The part that is my failure, not yours

I added the `claimed` label to #721 at **18:07 UTC on 2026-09-07** — after both PRs already existed — **and left no comment saying who it was for.** `CONTRIBUTING.md` is explicit that the comment *is* the claim and that the thread is the record, precisely because GitHub will not show a non-collaborator as an assignee. A bare label records nothing either of you can read. So when you looked at #721 you saw an issue with no claim comment on it, which is exactly what an unclaimed issue looks like.

There is one signal you could have caught — #737 appears in #721's timeline as a cross-reference — and I am not going to pretend otherwise. But a cross-reference in a timeline is not a claim, and the surface this project tells you to read said the issue was free.

## What I am doing about it, and what I am not doing

**I am not deciding this on priority alone.** Being 18 hours earlier is a fact, not an argument, and I would rather compare the two on evidence than hand it to whoever pressed the button first. Both are being measured side by side right now, like for like:

- whether each probes **capabilities** rather than a trl version number (this repo has a written rule about that — a version table here was wrong twice, which is why `trainer/_trl_compat.py` exists at all);
- where the compat logic lives — yours puts it in `_trl_compat.py`, which is the module this repo designates for exactly that, and that is a real point in its favour that I want to weigh properly rather than wave at;
- whether `training.epochs` and `training.ppo_epochs` stay separate end to end, with no double-write and no double-count against `_train_manual`;
- the same four mutations applied to both, including both directions of the epoch-conflation bug;
- which of #721's four acceptance criteria each one actually satisfies, especially the fourth, which requires tests to assert **non-default** values so a fallback to upstream defaults cannot pass by accident.

I will post the comparison here and on #737 with the numbers in it, not a verdict with the reasoning left out.

**Whatever the outcome, your work does not vanish.** If the other lands, anything yours does better goes in with your name on it, and I will say in the thread which parts those were. If yours lands, the same applies in reverse. What I will not do is let one of you find out from a merge notification.

Nothing is asked of you right now. CI is green on `cefc7564` and I have no review items for you yet.

MakazhanAlpamys:
@dchaudhari7177 — **the comparison is done, with the same mutations applied to both branches, and I owe you the numbers before the decision.**

## The decision

**#737 is the one I am taking forward.** Not because it was first. It was first by 18 hours and that is a real fact, but it is not the reason, and I would not have posted a verdict resting on it.

## Where you are equal, which is most of it

Driving the **genuine `trl.experimental.ppo.PPOConfig`** at the declared torch floor (shimming only the missing `FSDPModule` symbol, since #651 blocks the import), with each PR's real kwarg-selection code:

```
MAIN   -> num_train_epochs=3.0  num_ppo_epochs=2  kl_coef=0.05   # the bug
#737   -> num_train_epochs=7    num_ppo_epochs=2  kl_coef=0.7
#741   -> num_train_epochs=7    num_ppo_epochs=2  kl_coef=0.7
```

**The shipped behaviour is identical.** Both close #721's reproduction table. Both probe capabilities and never a version number — the rule this repo learned twice over, and you both arrived at it independently. Both keep the two meanings of "epochs" separate in code, with no double-write and no double-count against `_train_manual`. Both support the legacy names. Both are ruff-clean and delete nothing.

## The finding that decided it, and it is against your tests, not your fix

`test_ppo_epochs_is_still_a_separate_setting` does not enforce what its docstring claims. I mutated the modern branch to conflate — `ppo_kwargs["num_ppo_epochs"] = tcfg.epochs`, the reverse-conflation bug #721 explicitly warns about — and ran your file:

```
175:            ppo_kwargs["num_ppo_epochs"] = tcfg.epochs      <- mutated
177:            ppo_kwargs["ppo_epochs"] = tcfg.ppo_epochs      <- untouched legacy elif

tests/test_issue721_ppo_kl_and_epochs.py:  7 passed
```

Both substrings the test looks for are still present — `"num_ppo_epochs"` and `"tcfg.ppo_epochs"` — because the legacy `elif` you did not touch still contains the second one. So the assertion holds while the conflation is live. The same mutation on #737 fails two tests by name.

That is not a nitpick about style. `#721`'s own text singles out that conflation as the thing a fix must not do, and it is the one property in this PR that a source-substring test structurally cannot hold up.

The related gap is acceptance criterion 4 — *"tests assert non-default values so falling back to upstream defaults cannot pass accidentally"*. Your KL half satisfies it exactly (`kl_penalty_kwargs(_Modern, 0.7) == {"kl_coef": 0.7}`). The epoch half is a static AST read of `ppo.py`: it checks that an assignment node exists whose value unparses to `tcfg.epochs`. No value is ever compared against a default and no config object is ever constructed, so a fallback to trl's `3.0` cannot be detected by it. #737 exercises a real `PPOTrainerWrapper.setup()` against a config class whose defaults are trl's own and asserts `args.num_train_epochs == 7`.

#737 also implements the issue's fourth suggested-fix bullet — reporting the effective epoch budget and KL coefficient at setup — and adds six lines of `docs/training.md` distinguishing `epochs` from `ppo_epochs`. Neither is in yours.

## Three things yours does better, and I am not letting them evaporate

I want these on the record precisely because the decision went the other way.

1. **Placement. You are right and #737 is not.** `kl_penalty_kwargs` belongs in `trainer/_trl_compat.py`, next to `prompt_length_kwargs`, reusing `config_accepts`, with the same `(config_cls, value) -> dict` signature and the same "empty dict on a third rename" semantics. That module exists *because* a version table in this codebase was wrong twice, and putting the fourth trl rename anywhere else is how it drifts. #737's helper is local to `ppo.py` — which does match `ppo.py`'s own pre-existing habit, so it is not wrong, but yours is the shape the project says it wants.

2. **`test_the_new_name_wins_when_a_config_carries_both`.** This closes exactly the gap I flagged as uncovered in #737: neither of its tests presents both `kl_coef` and `init_kl_coef` in one signature, so preferring the legacy name would survive there. You covered it and #737 does not.

3. **CI hygiene at the declared floor.** Your test file imports no `trl` at all. #737's `pytest.importorskip("trl")` guards the wrong module — plain `trl` imports at the floor, `trl.experimental.ppo` does not — so it **fails** where it should skip, and it is still unfixed on its head:

```
#741: 7 passed                          0 new failures vs main
#737: 1 failed, 3 passed                +1 new failure, its own
```

You are also structurally immune to the Rich/ANSI assertion class that has turned this repo's CI red four times, because you assert on no console output.

## What I would like to do about it, if you are willing

Those three are not consolation. They are improvements #737 does not have, and I would rather land them than admire them.

Once #737 is in, I would like you to open a small follow-up that moves the KL rename into `_trl_compat.py` as `kl_penalty_kwargs` and brings `test_the_new_name_wins_when_a_config_carries_both` with it. That is your design and your test, credited to you in `CONTRIBUTORS.md` and in the changelog fragment, and it is a genuinely better end state than either PR alone.

**Entirely your call, and "no" costs you nothing** — you have already spent an evening on a collision my silent label caused. If you would rather not, say so and I will do it myself and still credit the design to you, or leave it and file it as an issue.

## One thing to fix either way

`changelog.d/0.74.0/741.fixed.md` is missing. This repo has not edited `CHANGELOG.md` directly since #487; a fragment named after the PR is the convention, and it is not something CI can catch — a *missing* fragment is invisible to `test_issue487_changelog_fragments.py`, which only validates ones that exist. Worth knowing for the follow-up regardless of what you decide here.

I will keep this PR open until you have answered. Nothing here is closed over your head.

dchaudhari7177:
@MakazhanAlpamys — thanks for putting the comparison and the numbers on the record before the decision. It was the right call. My `ppo_epochs` test really couldn't hold the conflation property, and a substring test was the wrong tool for a property #721 singled out. No hard feelings about the collision either.

Yes to the follow-up: **#913**, on top of #737 as merged.

- `kl_penalty_kwargs(config_cls, kl_penalty)` in `_trl_compat.py` next to `prompt_length_kwargs`, on `config_accepts`, same shape and empty-dict semantics.
- `_set_ppo_training_kwargs` takes the config class, so all three renames go through the same probe. No behaviour change.
- The both-names test is already in `main` via #737 (thanks for the credit in the comment). #913 keeps it, and it fails, together with a new direct case, if `kl_penalty_kwargs` prefers the legacy name.

I didn't add a changelog fragment, since there is no user-visible change. Say if you would rather have one.

Closing this in favour of #737 + #913.

PR #742, opened 2026-09-07
fix(distill): opt out of loss kwargs so gradient accumulation does not rescale gradients (closes #720)

### LINKED ISSUE #720: Distillation compute_loss ignores num_items_in_batch and can over-scale gradient accumulation
## Summary

The custom `_DistillTrainer.compute_loss()` accepts `num_items_in_batch` but
ignores it. It averages CE and KD inside each microbatch and does not explicitly
set `model_accepts_loss_kwargs = False`.

When Transformers classifies the model as accepting loss kwargs and supplies
`num_items_in_batch`, its usual gradient-accumulation compensation can be skipped
even though Soup has not normalized by the global denominator. This can make the
accumulated gradient depend on `gradient_accumulation_steps` rather than only on
the effective batch.

## Reproduction

`9f1e48b33a5fcda7e621e612ec9305cb52a38e07` with Transformers 5.16.1 and
PyTorch 2.13.0. In an isolated reproduction of the affected Trainer normalization
branch using four equal microbatches:

```text
full-batch gradient norm:       0.2638700074
affected accumulated norm:     1.0554800295
affected / reference ratio:    4.0
explicit /4 control max error: 1.39e-17
```

This demonstrates the normalization contract; it does not imply that every Adam
parameter update is exactly four times larger.

## Impact

For affected models, changing gradient accumulation can silently change gradient
scale and training dynamics despite an equivalent effective batch size. Averaging
each microbatch independently is also not globally token-weighted when sequence
lengths differ.

## Suggested fix

The minimal safe fix for the current mean-reduced loss is to mark the custom
trainer as not consuming loss kwargs:

```python
self.model_accepts_loss_kwargs = False
```

A more complete fix would compute summed CE/KD numerators and normalize by the
global number of supervised labels. That is needed if unequal-length microbatches
must match a single global token mean.

## Acceptance criteria

- [ ] Equivalent effective batches produce matching gradients for GA=1/4/8.
- [ ] Tests cover equal and unequal microbatch lengths.
- [ ] Tests cover response-only labels, padding, an incomplete final accumulation
      window, and distributed denominator handling.
- [ ] The behavior is explicit for every branch of `_DistillTrainer.compute_loss`,
      including sequence KD, MiniLLM, ULD, and ordinary logit KD.

dchaudhari7177:
Closes #720, taking the "minimal safe fix" the issue names.

## The bug

`_DistillTrainer.compute_loss` accepts `num_items_in_batch` and ignores it -- it returns a mean over its own microbatch. Transformers skips its own gradient-accumulation compensation whenever it classifies the model as consuming loss kwargs, so nothing divides by the step count and the accumulated gradient scales with `gradient_accumulation_steps`.

Measured rather than restated, on a tiny `nn.Linear` with four equal microbatches:

```
full-batch gradient norm       0.2638700074
uncompensated accumulation     1.0554800295
ratio                          4.0
compensated vs full batch      equal to 1e-6
```

which is the contract from the issue.

## The fix

`self.model_accepts_loss_kwargs = False`, in a `__init__` that calls `super()` first.

The ordering is not incidental: `Trainer.__init__` assigns that attribute itself, so setting it before the `super()` call would be silently overwritten. There is a test for exactly that, since it is the kind of thing that survives review and then does nothing.

## Tests

`tests/test_issue720_distill_grad_accum.py`, 6 cases, both acceptance boxes:

- **GA = 1, 4, 8 (parametrized)** -- equal microbatches, compensated, match the full-batch gradient to `rel=1e-6`.
- **the bug as a measurement** -- uncompensated four-way accumulation is exactly 4x the full-batch gradient.
- **unequal microbatch lengths** (2, 6, 8, 16) -- with the compensation the answer stops depending on the step count. It does *not* equal the full-batch number, and the test says so in as many words: a per-microbatch mean weights microbatches equally rather than tokens. That residual is the global token-weighted normalization the issue describes as the more complete fix, and it is **not** done here.
- **the flag is set, and set after `super().__init__`** -- read statically, since `_DistillTrainer` is defined inside `setup()` and constructing one needs a model, a dataset and a tokenizer.

On main the last of these fails with `_DistillTrainer defines no __init__, so it cannot opt out`.

```
tests/test_issue720_distill_grad_accum.py
  on main:            1 failed, 5 passed
  with this change:   6 passed

tests/*distill* (excluding the new file)
  on main:            1 failed, 68 passed
  with this change:   1 failed, 68 passed
```

The one failure is the same either way: `test_issue580_autodistill_fingerprints.py::test_fingerprint_verifier_rejects_symlinked_files`, which needs symlink creation and fails on this Windows checkout. Untouched by this change.

## What I did not verify

No end-to-end distillation run -- that needs a teacher and a GPU. What is measured is the normalization contract on a real autograd graph, plus that the trainer carries the opt-out. If you would rather have the summed-numerator version instead of the flag, say so and I will do that one; it is a larger change to `compute_loss` and would want the global label count threaded through.


MakazhanAlpamys:
@dchaudhari7177 — **the 12-line fix is correct, it is exactly the scope that was agreed, and I confirmed it works by measurement rather than by reading the diff. Blocking on one mechanical item and one substantive one.**

## The fix is right, and here is the evidence rather than an opinion

I checked the mechanism in the **installed** transformers 5.16.1, not in the PR body. `Trainer.__init__` sets the flag by signature inspection when the model declares no `accepts_loss_kwargs`, and `LlamaForCausalLM.forward` carries a `VAR_KEYWORD`, so it resolves to `True` for ordinary causal LMs. `training_step` then reads it:

```python
if (not self.model_accepts_loss_kwargs or num_items_in_batch is None) and self.compute_loss_func is None:
    loss = loss / self.current_gradient_accumulation_steps
```

So the compensation was being skipped. Transformers' own docstring prescribes the lever you used, word for word: *"If you are not using `num_items_in_batch` when computing your loss, make sure to overwrite `self.model_accepts_loss_kwargs` to `False`."* Not a guess that happens to pass.

Measured through the real path — real `LlamaForCausalLM`, real `Trainer.training_step`, mean-reduced `compute_loss`, `gradient_accumulation_steps=4`:

```
main (no opt-out): model_accepts_loss_kwargs=True   grad_norm = 1.8406364918
PR   (opt-out)   : model_accepts_loss_kwargs=False  grad_norm = 0.4601591229
ratio = 4.000000000434632  ==  gradient_accumulation_steps
```

The ordering claim in your body is real too: `Trainer.__init__` assigns the attribute itself, so setting it before `super()` would be overwritten. Mutating the assignment to sit *before* the super call is caught.

## BLOCKING 1 — the changelog fragment is missing

```
$ find changelog.d -name "*742*" -o -name "*720*"
(no output)
```

`changelog.d/README.md` requires `changelog.d/0.74.0/742.fixed.md`, named after the **PR** and not the issue, with `by @dchaudhari7177` in the text. Gradient scale under accumulation is user-visible, so this is not fragment-exempt. One file.

## BLOCKING 2 — five of the six new tests pass on unmodified `main`

I copied `origin/main`'s `distill.py` over yours in a worktree, confirmed byte-identity and `grep -c model_accepts_loss_kwargs == 0`, and ran your new file against it:

```
1 failed, 5 passed

FAILED test_distill_trainer_opts_out_of_loss_kwargs
  (the other five PASS against the unfixed code)
```

The five gradient tests build a bare `torch.nn.Linear(4, 1)` and take a `compensate: bool` that the test itself passes in, then re-derive `loss / steps` by hand. The file's imports are `ast`, `pathlib`, `pytest`, `torch` — `soup_cli` appears once, as a path-component string. **They would pass with `src/soup_cli/` deleted.** 143 lines of test, one discriminating assertion.

And that one assertion pins the **spelling**, not the behaviour: it is `ast.unparse(stmt) == "self.model_accepts_loss_kwargs = False"`, exact string equality. Respelling the fix as `= not True` — behaviourally identical — fails it. A correct fix can fail this test and a broken one could pass it.

**What to replace them with.** The measurement in the section above is the test this PR needs, and it is about 30 lines on CPU with a 1-layer `LlamaConfig(hidden_size=16, num_hidden_layers=1)` — no teacher, no GPU, roughly four seconds. Construct a `Trainer` subclass of the same shape as `_DistillTrainer`, run one accumulation window, and assert the ratio is `gradient_accumulation_steps` without the opt-out and `1.0` with it. That is a test that fails on `main` and passes here, which is the property the five you have do not have.

Keep the AST ordering test — mutating the assignment above `super().__init__` is caught only by it, so it earns its place — but relax the exact-string match to something that survives a respelling.

## One thing that is not yours to fix, found while reviewing this

Deleting `num_items_in_batch` from your `compute_loss` signature is an instant `TypeError` at `trainer.py:1940`, which calls it unconditionally — and **761 distill-touching tests do not notice.** No test in the repo constructs `_DistillTrainer` or executes its `compute_loss` at all. That parameter predates your PR, so it is not a finding against you; it is the reason your source-text assertion is currently the only thing standing between this fix and a silent revert.

## Everything else is clean

Nothing deleted or weakened (`git diff origin/main...c4ca005e -- tests/ | grep "^-[^-]"` is empty). Ruff clean. 14 of 14 checks green. Scope matches the ruling on #720 — the minimal fix and only the minimal fix — and @jagadeepmamidi's stand-down is on the record, so nothing is contested.

**Merge order is free:** I trial-merged this with #736 and with #743, which both touch the interior of `compute_loss` while yours adds an `__init__` above it. Clean both ways, and the opt-out survives intact in both merged trees.

MakazhanAlpamys:
@dchaudhari7177 — one consolidated check-in rather than five notifications. All five of your open PRs have been quiet since 2026-09-07, and each is close:

- **#742** (distill loss kwargs): the fix is correct and measured. One mechanical item and one substantive item from my 2026-09-09 review.
- **#729** (streaming `over` + `val_split`): the `729.fixed.md` fragment, @AmirF194's content-keying point, and the empty-train guard mirroring `_split_val_per_source`. The last one matters: as it stands, the PR trades a leak for a silent empty training set.
- **#727** (MLX harness pin): the fragment, plus the diagnostic that states a number the wrapper never computes.
- **#757** (the typer-defaults scanner): the source hunks are already on `main` via #754, so rebase down to `tests/test_typer_defaults_do_not_leak.py`, which is the part I want.
- **#741** (PPO KL forwarding): still waiting on your answer to the collision decision — nothing is closed over your head.

No pressure on timing; life happens. If you would rather hand any of these back, or have me finish the mechanical items on top of your commits with the credit staying yours, say which. Otherwise I will check again on **2026-09-13 18:00 UTC**.

dchaudhari7177:
@MakazhanAlpamys — both items are in `c38577d`, rebased onto current `main`. `distill.py` is unchanged from the head you reviewed.

**Blocking 1 — fragment.** `changelog.d/0.75.0/742.fixed.md`. It is under 0.75.0 because 0.74.0 has released, and it passes `assemble_changelog.py` on a temp copy.

**Blocking 2 — the five tests that passed on `main`.** They are gone, and in their place is the measurement you described. I went one step further than a look-alike subclass, because that would still pass if the line in `distill.py` were deleted. The test compiles the **actual** `_DistillTrainer` class body out of `distill.py`, binds it to the real `transformers.Trainer`, and runs one accumulation window through the real `Trainer.training_step` on a one-layer `LlamaForCausalLM` (`hidden_size=16`, CPU). With `_sequence_mode=True`, the real `compute_loss` returns its plain mean-reduced CE before any teacher, ULD or MiniLLM name is looked up, so no teacher is needed. The loop's non-`None` `num_items_in_batch` is passed explicitly, since `None` takes the compensating branch no matter what the flag says.

```
test_distill_trainer_gradient_does_not_scale_with_accumulation[4]   GA=4 / GA=1 == 1.0
test_distill_trainer_gradient_does_not_scale_with_accumulation[8]   GA=8 / GA=1 == 1.0
test_without_the_opt_out_the_gradient_scales_with_the_step_count     same class, __init__ removed: 4.0
```

The third test keeps the first two honest: if Transformers ever compensated on its own, it would fail and say so.

**AST ordering test.** It is relaxed to any assignment to `self.model_accepts_loss_kwargs` after `super().__init__`, however the value is spelled.

**Your mutations, on this head:**

| tree | result |
|---|---|
| `origin/main`'s `distill.py` copied in | **3 failed**, 1 passed (the `__init__`-removed control) |
| `= not True` respelling | 4 passed |
| assignment moved above `super().__init__` | **3 failed**: 4.0000005 and 8.0000009 against 1.0, plus the ordering test |

The file takes ~25 s locally, mostly the transformers import. `ruff@0.9.4` is clean. The other distill test files give 72 passed and 1 failed; the failure is `test_fingerprint_verifier_rejects_symlinked_files`, which needs symlink privilege on Windows and fails the same way with my change stashed. My local transformers is 5.1.0, below the 5.16.1 floor, but the `training_step` branch involved is the same one you quoted from 5.16.1. The floor job is CI's to confirm.

On the thing you found that isn't mine to fix (no test constructs `_DistillTrainer`): this file now does, in `_sequence_mode`. That also means removing `num_items_in_batch` from the `compute_loss` signature would now fail here with the `TypeError` you described, instead of going unnoticed.

MakazhanAlpamys:
Merged as `dd151374`. Both items verified on `c38577d8` by execution, and the replacement test is the one this PR needed: it dies on a gradient number rather than on a spelling.

I ran it on **transformers 5.17.0**, above our declared `>=5.16.1` floor — worth saying, since yours was 5.1.0 and below it. It agrees: 4 passed.

**The mandated mutation, run here:** restoring `self.model_accepts_loss_kwargs = True` —

```
FAILED test_distill_trainer_gradient_does_not_scale_with_accumulation[4]
FAILED test_distill_trainer_gradient_does_not_scale_with_accumulation[8]
2 failed, 2 passed
```

and the assertion it dies on is `accumulated / reference == pytest.approx(1.0, rel=1e-5)` — a real gradient norm through the real `Trainer.training_step`, which is exactly what I asked for and could not have been faked by a kwarg check. The measured ratio is `4.00000046` at GA=4, i.e. precisely `gradient_accumulation_steps`.

**The rest:** removing the whole `__init__` (the state of `main`) fails 3; moving the assignment above `super().__init__` fails 3; replacing `distill.py` wholesale with `main`'s fails 3. And the control that **must survive** does — `= not True`, a behaviourally identical respelling, passes 4/4, so the earlier exact-string pin is genuinely gone rather than merely loosened.

**What I want to name, because it is the harder thing you did.** Your first version passed on unmodified `main` in five of six cases: it built a bare `torch.nn.Linear(4,1)` and took a `compensate: bool` the test supplied itself, so it would have passed with `src/soup_cli/` deleted. Told that, you did not patch the assertions — you replaced the approach, compiling the **real** `_DistillTrainer` class body out of `distill.py` via AST and binding it to the real `transformers.Trainer`. That is what makes the whole-file mutation die, and it is why this file is now worth having.

One gap closed as a side effect, which I flagged as **not yours to fix**: before this, nothing in the repo constructed `_DistillTrainer` or executed its `compute_loss`, so deleting `num_items_in_batch` from that signature was an instant `TypeError` at `trainer.py:1940` that **761 distill-touching tests did not notice**. It now fails 3.

3696 passed, 17 skipped across the 37 files mentioning distill; ruff clean; 14/14 CI green on `c38577d8` itself. Zero lines removed from any existing repo test — the five deleted cases were the non-discriminating ones inside your own file.

**A correction to what I told you about merge order.** This merges clean with #736, #743, #923 and #908 individually, with the opt-out intact and your file passing in each merged tree. The conflict is between **#736 and #743**, in `distill.py` and `docs/training.md`, and it occurs in **both** directions — so whichever of those two goes second needs a rebase regardless of the ordering constraint I stated earlier. That is their problem, not yours, and I was wrong to imply it touched you.

**One note for the record, and it is on me.** The unequal-length test I asked you to delete was the only place stating the residual the issue itself names — that a per-microbatch mean weights microbatches rather than tokens, so with unequal lengths the fix restores step-count independence but not the full-batch number. The new tests use equal chunks only, so that residual is now documented nowhere. I asked for the removal, so this is not a finding against you; I am filing the follow-up.

Credited in `CONTRIBUTORS.md`. That is four of yours merged today — #916, #727, #729 and this.

Issue #693, opened 2026-09-05
ReLoRA prunes adapters without merging and can zero both factors on ties


Amix29:
## Summary

Soup's feature named ReLoRA periodically magnitude-prunes the learned `lora_A` and `lora_B` weights without first merging their update into the base model. This is not the merge-and-reinitialize restart defined by ReLoRA and it abruptly changes the learned function.

There is also a concrete pruning bug: strict thresholding can remove every entry when magnitudes tie, even though `prune_ratio < 1`. If both LoRA factors become zero, the branch receives zero ordinary gradients and cannot restart.

## Current behavior

On current `main` (`ef2e02b51a1b34546823edce55715d28234805cb`), `_prune_and_reset()`:

1. selects parameters whose names contain `lora_A` or `lora_B`;
2. calls `magnitude_prune_tensor()` on those adapter weights;
3. clears optimizer state for those same parameters.

It never merges `B @ A` into the frozen base weight and never reinitializes A/B in the canonical LoRA shape.

The ReLoRA paper's restart instead merges the low-rank update into the base, reinitializes A with Kaiming and B with zeros, prunes/resets optimizer state, and restarts the learning-rate warmup: https://arxiv.org/html/2307.05695#S2

Soup's module docstring says it mirrors the paper/Axolotl, while the public docs describe adapter-weight pruning. That makes this both an algorithm-conformance problem and a behavior that is destructive to the currently learned adapter contribution.

## Reproductions

Using the actual callback kernel from the pinned checkout:

### Learned-function discontinuity

With random non-zero LoRA factors, the base weight is unchanged by the callback and the MSE between the pre-callback and post-callback effective linear maps changes from 0 to `3.739178...`.

A merge-then-reset control preserves the effective map exactly.

This demonstrates function modification, not downstream model-quality impact; no full LLM training was run.

### Tie handling can kill the branch

The mask is:

```python
threshold = torch.kthvalue(flat, k).values
mask = tensor.detach().abs() > threshold
```

All values equal to the threshold are removed. With constant A and B tensors and `prune_ratio=0.9`, the actual implementation yields:

```text
nonzero A: 0
nonzero B: 0
MSE after callback: 16.0
gradient norm A: 0.0
gradient norm B: 0.0
```

The equal-magnitude case is constructed; this report does not claim a measured production frequency. The exact-count violation itself is deterministic.

## Recommended fix

Choose and document one of two coherent contracts:

[...]

PR #743, opened 2026-09-07
feat(distill): token-chunked checkpointed KL distillation (#722)

kok-o:
## What does this PR do?

Refs #722 by implementing token-chunked evaluation with optional non-reentrant activation checkpointing (`torch.utils.checkpoint.checkpoint(..., use_reentrant=False)`) for KL/JS distillation divergence loss.

### Problem
In distillation with large vocabularies (e.g. 150k for Qwen 2.5) and long sequences, computing student/teacher divergence across the entire sequence retained massive activation tensors on the autograd tape (`B * S * V * 2` bytes). In SFT/chat distillation, most tokens are prompt and padding tokens that are masked out anyway.

### Solution
1. **Response-only filtering**: In `_compute_distill_term`, student and teacher logits are filtered to active supervised tokens (`labels != -100` when present, else `attention_mask`) before divergence evaluation, immediately shedding unneeded prompt and padding tokens.
2. **Token chunking**: Active tokens are chunked in batches of size `training.distill_chunk_size` (e.g. 256 or 512 tokens), accumulating sum reductions normalized by active count and scaled by `temp^2`.
3. **Activation checkpointing**: When `training.distill_checkpoint: true`, each chunk evaluation is wrapped in non-reentrant checkpointing, discarding intermediate `log_softmax`, `softmax`, and divergence tensors during forward and recomputing them during backward.
4. **Frozen teacher memory isolation**: Non-grad teacher activations are captured via closure in the checkpoint wrapper, preventing autograd from retaining redundant teacher tensors as arguments on the graph.
5. **Config & Schema**: Added `distill_chunk_size` (positive int, rejects bool) and `distill_checkpoint` (bool) to `TrainingConfig`, strictly validated to require `task='distill'`.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist


MakazhanAlpamys:
@kok-o — you have had two days of silence on two substantial pull requests and no acknowledgement at all, which is the worst version of waiting because you cannot tell whether they were seen. They were. Both are under review now and I will post findings on each rather than a verdict.

You claimed #722 and #723 on 2026-09-07 with an implementation plan in each, then delivered against those plans without being chased. That is the part I want on the record before any review finding lands.

Two things I can already tell you, both mechanical and neither a judgement on the code:

- **Both changelog fragments are named after the issue rather than the pull request.** `changelog.d/README.md` requires the PR number, so `722.added.md` should be `743.added.md` and `723.fixed.md` should be `744.fixed.md`. This is a house convention you had no way to infer from the existing files, several of which are wrong in the same way — it is not something CI can catch, so nothing would have told you but me.
- On **#744**, worth confirming before I get there: #723 is partly already closed. @AmirF194's #735 merged as `e1f5f69` for the masked-token centring half, deliberately marked `Refs` rather than `Fixes` because the sequence-level objective needs a different code path. If that is the division you worked to, say so and I will check yours against it rather than against the whole issue.

What I am checking, so you know what the findings will be about: for #743, whether the chunked and checkpointed path produces the same loss and gradients as the unchunked one — a memory optimisation that changes the loss silently changes every distillation run; and for #744, whether the length normalization excludes masked tokens, since counting padding there would be the same defect class #735 just fixed, one step to the left.

Nothing is asked of you yet beyond the two file renames if you want to get ahead of them.

MakazhanAlpamys:
@kok-o — **the equivalence claim holds and I verified it against `main`'s own kernel rather than against your `chunk_size=None` path. Blocking on three things, one of which is a real numerical finding I reproduced myself.**

## What is right, established before the findings

Chunked and checkpointed evaluation reproduces the dense loss and gradients in float32 across divergences, masking modes and chunk sizes that do not divide the token count, including `chunk_size=1` and the single-chunk case. Compared out-of-process against `origin/main`'s `_compute_distill_term`: worst relative loss deviation `4.63e-16`, worst absolute gradient deviation **exactly zero**.

The peak-memory win is real: 874 MiB dense to 380 MiB at chunk 8 with checkpointing, 2.30x, measured on CUDA at B=1 S=1024 V=32000.

Your schema work is also better than most of what reaches this repo. Non-default values parse and land; `model_dump()` round-trips back through `SoupConfig(**d)`, which is the #321 footgun this project shipped twice; bool, zero, negative and float are all rejected with grep-able keywords; the footgun validator fires with the field named. That is release-checklist 6c actually done.

## BLOCKING 1 — the memory knob is also an accuracy knob in bf16

`total_sum` inherits the logits dtype (`distill.py:170`), and the trainer hands the kernel raw model logits with no upcast (`:743-748`). Under bf16 training that makes the running sum a bf16 scalar accumulated once per chunk.

My measurement, 2x128 tokens, vocab 8192, 176 supervised, forward_kl, against an fp64 ground truth:

```
random logits (fp64 truth = 0.999513)
  chunk None..1      rel err 4.87e-04    flat

larger divergence (fp64 truth = 8.461520)
  chunk None         8.4375   2.84e-03
  chunk 256/64/16/4  8.4375   2.84e-03
  chunk 1            8.3125   1.76e-02      <- 6x worse
```

**One correction to my own review before you act on it:** I first described this as growing monotonically as the chunk shrinks. My data does not show that. It is flat from `None` down to `chunk=4` and jumps at `chunk=1`, and the magnitude depends strongly on how large the divergence is — negligible at a KL near 1, 1.8% at a KL near 8.5. So the accurate statement is narrower than "chunking degrades the loss": **at the smallest chunk, which is the setting a user chasing memory will reach for, the loss moves by more than a percent on a realistic divergence, and nothing in the suite would see it** because every equivalence test runs at the float32 default.

One line fixes it — accumulate in float32 and cast back — and the equivalence tests need parametrising over dtype, or they will keep not seeing it. Worth knowing: **#736 independently forces `.float()` in this same kernel**, so that PR fixes this incidentally.

## BLOCKING 2 — the two new flags are not proven to reach the kernel

Setting `chunk_size=None, use_checkpoint=False` at the `compute_loss` call site — i.e. disconnecting both config flags from the code they control — passes all 215 relevant tests.

This is the defining defect class of v0.74.0 and this project shipped it four times in one release: `bnb_4bit_use_double_quant`, the MLX quantization path, `--no-reexec`'s dropped flags, and `data.interleave`, which is still open as #443. Every one was validated, documented, and read by nothing. Your wiring is correct today; it has zero coverage. A test that monkeypatches `_compute_distill_term` and asserts the kwargs arrive closes it.

## BLOCKING 3 — the equivalence suite compares the new kernel against itself

Every equivalence test uses `_compute_distill_term(..., chunk_size=None)` as the dense baseline. That is the same function you rewrote. A change to the **shared** semantics is invisible by construction — and the denominator is a line you now own. Demonstrated: changing `denom = mask.sum()` to `mask.numel()`, which dilutes the loss across prompt and padding tokens, passes all 215 tests.

The out-of-process comparison against `origin/main`'s kernel is what actually established equivalence here. That is the missing test, and it is worth having permanently: pin against the pre-PR formula, not against yourself.

## Two more, not blocking

`use_reentrant=False` can be dropped from both `checkpoint(...)` calls and nothing fails. Values are identical in this regime, so low severity — but non-reentrant checkpointing is a headline claim of the PR and nothing pins it.

**`docs/` is untouched**, and the PR checklist box for docs is ticked. `docs/training.md:417-428` documents `distill_divergence`, `distill_temperature` and `distill_mode`; the two new knobs appear nowhere, and #722's acceptance criterion 4 asks for the compute/memory tradeoff to be documented. That tradeoff is real and measured: chunk 64 is *faster* than dense (11.8 ms vs 20.1 ms) while chunk 16 with checkpointing is 1.9x slower.

Also worth rewording rather than fixing: the fragment says chunking and checkpointing eliminate activations from the autograd tape. Measured, **retained memory is flat at 125 MiB across every chunk and checkpoint configuration** — the retained win comes entirely from the response-only selection, and chunking moves the *peak*. Both are real wins; #722 explicitly warns against presenting retained bytes as VRAM saved, so they are worth attributing separately.

## Sequencing, and this one is on me to decide

**Resolve against #736 first, or let #736 land first.** #736 rewrites the exact three divergence formulas your `_chunk_kernel` copies verbatim into three branches, and the two conflict textually. #736 also delivers #722's acceptance criterion 5 — the stable kernel — which your PR leaves open because it reproduces the current formulation three times over. And its `.float()` upcast incidentally fixes BLOCKING 1.

I will not make you carry that ordering. Tell me which you would prefer and I will sequence the queue around it.

**Fragment rename:** `changelog.d/0.74.0/722.added.md` should be `743.added.md`. The assembler cannot catch it, so nothing but review will.

`tests/fixtures/recipe_config_snapshots.json` is a legitimate mechanical consequence — I deleted the two added lines and `test_baseline_matches_committed_snapshot` failed immediately.

Zero of your 29 tests pass against unfixed `main`, so nothing in that file is decorative. Ruff clean, 15 of 15 checks green with `head_sha` confirmed.

MakazhanAlpamys:
@kok-o — **both blockers from my last review are genuinely fixed, and I verified them by measurement rather than by reading the diff. Blocking on one thing: `distill_chunk_size` — the headline feature of this PR — has no test that can detect it being a complete no-op. I proved that, and I am handing you the probe that kills it.**

## The blocker, reproduced end to end

I mutated `distill.py:165` so chunking never happens:

```python
-    c_size = chunk_size if (chunk_size is not None and chunk_size > 0) else n_tokens
+    c_size = n_tokens
```

Binary-mode replace, occurrence count asserted at 1, `sha 81b1518922f7c627 -> c7a421181483ef6b`, `PYTHONPATH` verified to resolve `soup_cli` into the worktree and not the main checkout.

```
$ pytest tests/test_issue722_chunked_distill.py -q
79 passed in 16.63s
```

**Every one of your 79 tests passes with the feature switched off.** The wider suite agrees — `pytest tests/ -k distill` is 404/404 green under the same mutation.

Five other mutations I ran were all killed cleanly, so this is a single specific gap rather than a thin test file: removing the fp32 accumulation dies on `[bfloat16-forward_kl]`; an off-by-one dropping the last chunk kills 51 of 79; dropping the response-only prefilter kills all 30 `test_matches_pre_pr_reference_kernel`; ignoring the config flag dies on your own threading test; and silently disabling checkpointing kills `test_checkpoint_explicitly_uses_non_reentrant` plus the retained-bytes test.

## Why the memory test cannot see it, which is the interesting part

`test_autograd_retained_activation_savings` measures retained autograd bytes. Measured at `B=4, S=128, V=512`, 50% supervised:

```
chunk=32 + ckpt : retained=524,800   dense=1,049,088   ratio 2.00x
chunk=8  + ckpt : retained=524,800   dense=1,049,088   ratio 2.00x   <- identical
chunk=32 no-ckpt: retained=1,049,088                   ratio 1.00x
chunk=8  no-ckpt: retained=1,049,088                   ratio 1.00x   <- chunking alone: 1.00x
```

**100% of that 2.00x comes from `distill_checkpoint`. `distill_chunk_size` contributes exactly 1.00x on that metric** — because checkpointing retains the chunk *inputs*, whose total is the same however you slice them. Chunking's actual benefit is a bounded *transient* peak inside the forward pass, and nothing in this PR measures it.

Two consequences worth stating plainly. A refactor that drops chunking ships green. And `docs/training.md`'s claim that chunking bounds peak activation tensors currently has no test and no measurement behind it in this PR.

## The fix, already written and verified against both versions

The mechanism *is* real — it just needs one assertion. Count `log_softmax` invocations:

```python
def _calls(chunk_size, S=64, V=32):
    torch.manual_seed(0)
    s = torch.randn(1, S, V, requires_grad=True)
    t = torch.randn(1, S, V)
    n = {"c": 0}
    real = torch.log_softmax
    def spy(*a, **k):
        n["c"] += 1
        return real(*a, **k)
    torch.log_softmax = spy
    try:
        _compute_distill_term(s, t, temperature=1.0, divergence="forward_kl",
                              chunk_size=chunk_size, use_checkpoint=False)
    finally:
        torch.log_softmax = real
    return n["c"]

# assert _calls(cs) == (1 if cs is None else math.ceil(64 / cs))
```

On your unmutated head:

```
OK  chunk_size=None  calls=  1  expected=1
OK  chunk_size=   8  calls=  8  expected=8
OK  chunk_size=  16  calls=  4  expected=4
OK  chunk_size=  64  calls=  1  expected=1
```

Under the M2 mutation:

```
OK       chunk_size=None  calls=  1  expected=1
MISMATCH chunk_size=   8  calls=  1  expected=8
MISMATCH chunk_size=  16  calls=  1  expected=4
OK       chunk_size=  64  calls=  1  expected=1
```

It discriminates in both directions and it is about ten lines. `chunk_size=64` passing on both is deliberate: at `c_size >= n_tokens` a single call **is** the correct answer, so the test does not just assert "more than one".

## The bf16 numerics: fixed, and I am correcting the number I gave you

On your new head, the PR's own parity fixture (`seed=42`, `(2,8,16)`, `temp=2.0`, 8 active positions, bf16, CPU, torch 2.5.1), relative to unchunked:

| divergence | chunk=1 | chunk=2 | chunk=4 | chunk=8 |
|---|---|---|---|---|
| forward_kl | 0.0000% | 0.0000% | 0.0000% | 0.0000% |
| reverse_kl | 0.0000% | 0.0000% | 0.0000% | 0.0000% |
| js | 0.0000% | 0.0000% | 0.0000% | 0.0000% |

Max gradient delta `0.000e+00` at every chunk size, same with checkpointing on, and the same at `(2,128,4096)` across two seeds in both bf16 and fp32. **Chunk-size independence is now exact.**

On the pre-fix head `84a0eda`, the same fixture shows `reverse_kl` 0.4237% at chunk=1 and `js` 0.4608% at chunk=1 and 2 — so the commit fixed something real. **It is 0.42-0.46% here, not the 1.76% I quoted you.** The magnitude is fixture-dependent and I should have said which fixture produced my number the first time. The *shape* I described held — a jump at small chunk, flat above — but treat the specific figure as withdrawn rather than confirmed. The residual bf16 error against fp32 (forward_kl 0.73%, reverse_kl 2.15%, js 0.48%) is identical for unchunked and every chunk size, i.e. inherent to bf16 and not to your chunking.

## Everything else checks out

- **No test deleted or weakened.** The `tests/fixtures/recipe_config_snapshots.json` edit is schema-forced, not bent — reverting that file alone fails `test_issue621_recipe_snapshot` with exactly the two new fields at their defaults, so no recipe's behaviour changed.
- 79 / 404 / 169 passing, ruff clean, and the pre-PR reference kernel test is a good idea: it pins the off-path against an independent implementation rather than against itself.

## Three smaller items

1. **The changelog fragment does not credit you.** 25 of the 27 fragments in `changelog.d/0.74.0/` carry `by @user` and yours is one of the two that do not. Please make it `(#722 by @kok-o in #743)`.
2. **The denominator is still quantized to bf16** — `denom = mask.sum().to(dtype=s.dtype)` rounds the *token count* before the `.float()` division: 257 active tokens becomes 256.0, a systematic 0.39% error. Pre-existing, not introduced by you (I confirmed pre-PR and this PR give byte-identical results there), but this is the fp32-accumulation commit and the fix is one word. Worth taking while the file is open.
3. **The docs numbers have no provenance** — "11.8 ms vs 20.1 ms", "~1.9x slower", "~2.3x" with no hardware, model, shape or seed. And #722's own acceptance criterion asked for retained-tensor bytes not to be presented as VRAM saved; the fragment says "reducing peak activation memory by up to 2.3x", which is that conflation. Either state the conditions or soften the claim to "retained autograd tensor bytes".

**Add the chunk-count test and credit yourself in the fragment, and I merge on green.** Items 2 and 3 are worth doing in the same push but I will not hold the PR on them.

MakazhanAlpamys:
@kok-o — ordering note, and it affects your kernel rather than your open blocker. Nothing new is being asked on `distill_chunk_size` itself.

**#736 (reverse-KL / JS NaN gradients, #719) lands before this PR, and your `_chunk_kernel` currently reinstates the bug it fixes.** I read `pr743:src/soup_cli/trainer/distill.py` directly rather than inferring from the diff:

```python
if divergence == "reverse_kl":
    log_t = torch.log_softmax(t_c, dim=-1)
    p_s   = torch.softmax(s_c, dim=-1)
    return kl_div(log_t, p_s, reduction="sum")   # differentiable probs as the target
...
log_m = m.clamp(min=1e-12).log()                  # fp16-unrepresentable clamp
```

That is the exact shape #719 is about: the target-side derivative goes non-finite when softmax underflows, so the loss comes back finite while the gradient is `nan`. Your chunked path also computes in `s.dtype` with no upcast.

I tested the merge: this PR conflicts with #736 in both `src/soup_cli/trainer/distill.py` and `docs/training.md`, and resolving in this PR's favour would silently revert the fix. The good news is the conflict is loud and #736 ships the regression test, so a wrong resolution goes red rather than quiet.

**The port is small.** #736's formulation is `(p_s * (log_s - log_t)).sum(-1)`, which drops straight into your `reduction="sum"` chunk structure without changing how you accumulate; the JS mixture becomes `logaddexp(log_s, log_t) - log(2)` instead of the `1e-12` clamp. Please rebase once #736 merges and port it into `_chunk_kernel` rather than around it — an equivalence test against the unchunked path would then also be testing the fixed kernel, which is what you want.

One thing worth keeping in mind while you do: #736 upcasts to fp32, which costs roughly 1.24x peak on my measurements. That runs against what this PR exists to achieve, so if you find the chunked path can hold the numerical property without the upcast — chunking already bounds the retained activations — say so and I will take that seriously rather than treat #736's choice as settled for your kernel too.

kok-o:
@MakazhanAlpamys — pushed commit `7fae9da` addressing all findings:

1. **distill_chunk_size invocation test**: Added `test_chunk_size_controls_invocation_count` in `tests/test_issue722_chunked_distill.py` asserting `torch.log_softmax` call counts across `[None, 8, 16, 64]` on $S=64$. Verified that applying the M2 mutation (`c_size = n_tokens`) kills it with `AssertionError: 1 == 8`, while 80/80 pass unmutated.
2. **bf16 denominator precision**: Changed `denom = mask.sum().float()` in `src/soup_cli/trainer/distill.py` so active token counts do not quantize to bf16 before division.
3. **Changelog credit & memory attribution**: Credited `(#722 by @kok-o in #743)` in `changelog.d/0.74.0/743.added.md` and clarified retained autograd bytes vs transient peak activations.
4. **Docs tradeoff clarification**: Softened claims and removed ungrounded benchmark timings in `docs/training.md`.

Regarding the sequencing note on #736: understood and noted. As soon as #736 merges onto `main`, I will rebase and port the log-space `(p_s * (log_s - log_t)).sum()` and `logaddexp` JS formulations directly into `_chunk_kernel`.

MakazhanAlpamys:
@kok-o — re-reviewed at `7fae9dad`. **Every item from my 2026-09-09 review is genuinely done**, and I checked each by mutation rather than by reading. Two things left, both small, and one honest correction to something I said last time.

## Your previous items, verified

| asked | status |
|---|---|
| a test proving `distill_chunk_size` is not a no-op | **done** — my `c_size = n_tokens` mutation now fails exactly `test_chunk_size_controls_invocation_count`, and it discriminates in both directions (`chunk_size=64` at `n=64` still expects 1 call and passes) |
| fragment credited `(#722 by @kok-o in #743)` | **done**, and the stale `722.*` fragment is gone |
| pin against the **pre-PR** formula, not against yourself | **done**, and I checked the inlined reference is faithful: main's `0.5*(masked_mean(kl_pm)+masked_mean(kl_qm))` and your `masked_mean(0.5*(kl_pm+kl_qm))` are identical by linearity over one mask. Dropping the response-only prefilter kills all 30 parameters. |
| prove the flags reach the kernel | **done** — the test drives the real `setup` + `compute_loss` and asserts the values arriving at the kernel |
| `use_reentrant=False` unpinned | **done**, dropping the kwarg kills exactly one test |
| docs numbers with no provenance | **done** — "11.8 ms vs 20.1 ms", "~1.9x", "~2.3x" are gone, and the fragment now distinguishes retained autograd bytes from peak VRAM, which is what #722 AC4 asked for |

**Nothing was weakened to get here.** `git diff 7b2b160c..HEAD -- tests/` is **+34 / −0**, and the whole PR's `tests/` diff has zero deleted lines. 80 passed in the new file; 3764 passed, 17 skipped across all 34 files touching distillation plus the startup-lightness guard; ruff clean.

## The equivalence claim holds, and I verified it against main rather than against you

I loaded `origin/main`'s kernel as a separate module and compared in-process, across `chunk_size ∈ {None,1..8,16,64,n_active,n_active+1,100000}` × checkpointing × all three divergences.

**Gradients are bit-identical to `origin/main` in every configuration** — `max|Δgrad| = 0.000e+00` — including every chunk size that does not divide evenly, a prime active-token count of 7, a chunk larger than the sequence, a sequence shorter than one chunk, and w.r.t. real `nn.Module` parameters. Loss differs only at fp32 ULP scale from summation order (worst 2.4e-07 relative).

And since the PR asserts lower peak memory without a number, I measured it — RTX 5070 Laptop, torch 2.14.0+cu130, bf16, B=1 S=1024 V=32000, 512 active tokens:

```
origin/main dense            +499.6 MiB
chunk=None ckpt=False        +411.4 MiB   1.21x
chunk=128  ckpt=False        +337.3 MiB   1.48x
chunk=128  ckpt=True         +290.4 MiB   1.72x
chunk=16   ckpt=True         +256.3 MiB   1.95x
```

Two things worth keeping from that. **Chunking alone does reduce peak** (411.4 → 316.8 MiB with checkpointing off), so `docs/training.md`'s claim about bounding transient activations is supported — it simply is not the axis the retained-bytes test measures, which is why my earlier "chunking contributes exactly 1.00x" and this are both true. And **`distill_checkpoint: true` with `distill_chunk_size` unset is worth nothing** (411.4 MiB, identical to checkpointing off), because one checkpointed chunk's recompute transient equals its forward transient. A user who sets only the checkpoint flag gets the prefilter win and nothing else, and the docs do not say so. Worth one sentence.

## Two changes

**1. The two fp32 fixes I asked for last time are correct in the source and pinned by nothing.** Reverting the fp32 accumulation — init dtype plus both `chunk_sum.float()` plus the dense `.float()` — passes **80/80**. Reverting the fp32 denominator at `distill.py:137` to `.to(dtype=s.dtype)` also passes **80/80**.

A correction I owe you: I said last time this "dies on `[bfloat16-forward_kl]`". **It does not, and it did not.** I re-applied the same mutation to the previously reviewed commit `7b2b160c` against its own 79-test file and it passed 79/79 there too, so this is a standing gap rather than anything the new push introduced — most likely a torch version difference between our boxes. The fixture cannot see it because at 8 active tokens the effect is 7.81e-03 in absolute terms but the test compares chunked against dense *in the same build*, where the gap is 3.91e-03, under the 5e-3 tolerance.

At production scale it is not marginal. bf16, 1448 active tokens, forward_kl, against an fp32 reference of 2.251384:

```
chunk=8   fp32 accumulation 2.265625 (err 0.63%)   bf16 accumulation 2.093750 (err 7.00%)
```

And the denominator: 257 active tokens quantize to 256.0 in bf16 (0.389% systematic), 513 to 512.0, 1201 to 1200.0.

One bf16 equivalence test with **at least 512 active tokens** — dense versus chunked at `chunk=8`, absolute tolerance 5e-3 — kills both mutations. At that size the accumulation effect is 0.17, thirty-four times the tolerance.

**2. Your two new flags are accepted on three paths that never call the kernel.** Verified by execution:

```
uld_strategy: wasserstein  + distill_chunk_size: 128  -> compute_loss takes the _uld_projection branch
minillm_enabled: true      + distill_chunk_size: 128  -> takes _minillm_cb.distill_term
distill_mode: sequence     + distill_chunk_size: 128  -> short-circuits before the kernel
```

`_compute_distill_term` is reached only from the final `else` at `distill.py:744`. On those three configurations the flags are validated, documented and read by nothing.

I am asking for this specifically because it is **this project's signature defect** — v0.73.3 was named for four instances of it, the MLX backend supplied six more, and `data.interleave` is open as #443 for the same shape. Your task-gate footgun already works correctly (`task: sft` plus either flag is refused, naming both fields), so the machinery exists; it needs three more conditions, or an explicit documented note if you would rather not refuse.

## Three things I am not asking you to change

`distill.py:178-181` — the `if t_chunk.requires_grad:` checkpoint branch is dead: teacher logits always come from `torch.no_grad()`, and removing checkpointing from that branch alone survives 80/80. Delete it or cover it, whenever.

The default path's bf16 loss moves by one bf16 ULP against `main` (0.996094 → 1.000000, ~0.4%) with both flags unset, because of the fp32 changes I asked for. It is the right direction, but existing bf16 distill runs will not reproduce their curve past ~3 significant figures. Given how this project treats reproducibility after the GSPO change, one clause in the fragment seems fair.

fp64 inputs now compute at fp32 precision (`distill.py:142`). Irrelevant to training; relevant only if the kernel is ever used as an fp64 ground truth.

## And the thing that needs a decision rather than a fix

**#722 cannot close on this PR, and I want that on the record rather than discovered later.** Two acceptance criteria are undelivered: AC3 asks for peak memory *and throughput* on the same shape, and throughput is measured by neither of us; AC5 asks that reverse-KL and JS use the stable kernel rather than reproducing the non-finite-gradient behaviour, and `_chunk_kernel` copies the current formulation verbatim into all three branches. I reproduced the #719 condition (fp16, peaked student, softmax underflow): `origin/main` reverse_kl gives a finite loss with a **non-finite gradient**, and this PR gives a non-finite gradient at every chunk size, with and without checkpointing. Identical — neither better nor worse.

That is not a criticism of the PR. #722 explicitly says "start with forward-KL; extend to reverse-KL and JS only after their stable log-space formulation is in place", and that formulation is your #736, which is still open. It does mean this merges as partial progress on #722 with `Refs`, not `Closes`.

**Sequencing:** `pr743` merges cleanly into current `main`, but `git merge-tree` against #736's head gives conflicts in **both** `src/soup_cli/trainer/distill.py` and `docs/training.md`. #736 is with a reviewer right now, so I will tell you the order once I have that verdict — you should not have to guess.

## Scheduling

`main` is under a **merge freeze** until v0.75.0 is tagged: the release PR (#879) has already assembled `CHANGELOG.md` from the fragments, and anything merged before the tag leaves a fragment behind and fails `publish.yml --check-empty`. So push the two changes whenever suits you; nothing merges today regardless.

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:
@kok-o — the sequencing answer I promised, now that #736's re-review is done.

**#736 merges first, and the reason is substantive rather than a coin toss.** I verified with `git merge-tree` that exactly two files collide: the divergence dispatch in `src/soup_cli/trainer/distill.py` and a section of `docs/training.md`. But the important part is what your side of the conflict still contains:

- `kl_div(log_t, p_s, reduction="sum")` for `reverse_kl` — that is the #719 defect verbatim, the one #736 exists to fix;
- `m.clamp(min=1e-12).log()` for JS;
- `return loss.to(dtype=s.dtype)`, which reverts the FP32 return contract #736 establishes and pins with a test.

So merging yours first would silently reinstate a bug that is measured to produce **non-finite gradients** in all three dtypes, and to underflow reverse-KL to exactly `0` in bf16 and fp16. That is not a merge order I can take.

**A trap for your rebase, which I would rather you hear now than discover.** The prelude of that function **auto-merges silently** — git keeps #736's `.float()` upcast (with `log_s` / `log_t` computed up front) while dropping the `s = student_logits / temp` and `t = teacher_logits / temp` locals and the `_masked_mean` closure that your `_chunk_kernel` references. So `git checkout --theirs` or pasting your side verbatim gives a `NameError`, not a conflict marker. **Your chunking has to be ported onto `log_s` / `log_t`.**

The translation is mechanical and I checked it fits your structure: `(p_s * (log_s - log_t)).sum()` drops straight into your `reduction="sum"` shape. The `docs/training.md` collision is not a real conflict at all — non-overlapping content that landed in the same place; keep both, #736's paragraphs first.

Two other things from your own thread stand unchanged: the bf16 test at >=512 active tokens that pins the two fp32 fixes, and the footgun for the three config paths that accept your flags without reading them.

None of this is urgent. `main` is frozen until v0.75.0 is tagged — the release merged an hour ago but the tag is not out — and #736 still has its own rebase to do first. I will tell you when it lands so you are not rebasing onto a moving target.

Separately: your fragment is in `changelog.d/0.74.0/`, which v0.75.0 emptied, so it needs `git mv` to `changelog.d/0.75.0/` on the same rebase. Sixteen open PRs are in that position — you are not being singled out.

MakazhanAlpamys:

```
git mv changelog.d/0.74.0/743.fixed.md changelog.d/0.75.0/743.fixed.md
git push --force-with-lease
```

MakazhanAlpamys:
@kok-o — reviewed at `7fae9dad`. The chunking machinery itself is good work and its tests are strong, but **there is a blocking numerical problem, and it is not your fault in the way it looks.** #736 (@be-student, for #719) is landing first and it changes the formulation this PR copied. Please read the ordering section before the rest.

## What is good, measured

`tests/test_issue722_chunked_distill.py` is 80 tests and they are load-bearing. Three mutations, all caught:

| mutation | result |
|---|---|
| `c_size = n_tokens` — ignore `chunk_size` entirely | **1 failed** |
| drop cross-chunk accumulation (`total_sum = chunk_sum`) | **44 failed** |
| `use_reentrant=False` -> `True` | **1 failed** |

**666 passed, 1 skipped** on your head across the twelve distill/recipe files. `ruff` clean. The schema addition follows house convention — bool-before-int guard, footgun rejection when `task != distill`, snapshot fixture updated — and it passes v0.75.0's new `#748` consumer guard and `#755` backend registry, which most branches written before the release do not.

One coverage note worth keeping: only `test_chunk_size_controls_invocation_count` detects that chunking has become a no-op. `test_autograd_retained_activation_savings` does **not**, because checkpointing alone still reduces retained bytes. That is a thin single point of failure for the PR's central claim.

## 1. BLOCKING — `_chunk_kernel` carries the pre-#736 unstable form, and it is not confined to the chunked path

Your `_chunk_kernel`, lines 148-151:

```python
if divergence == "reverse_kl":
    log_t = torch.log_softmax(t_c, dim=-1)
    p_s = torch.softmax(s_c, dim=-1)
    return kl_div(log_t, p_s, reduction="sum")   # differentiable student probs as TARGET
```

and the JS branch mixes in probability space with a `1e-12` clamp. #736 replaces both with log-space forms (`p_s = log_s.exp(); (p_s * (log_s - log_t)).sum(-1)`, and `log_m = logaddexp(log_s, log_t) - log(2)`).

Measured on `main + this PR`, #719's reproducer at `temperature=0.05`:

| divergence | path | loss | grads finite |
|---|---|---|---|
| forward_kl | dense / chunked | 9.315e-05 | True |
| **reverse_kl** | dense / chunked | 8.406e-07 | **False** |
| **js** | dense / chunked | 2.913e-07 | **False** |

Because the rewrite replaced the whole function, **the dense path regresses too** — this is not only about chunking. Two related losses: the FP32 upcast is dropped (`s = student_logits / temp`, no `.float()`), and `return loss.to(dtype=s.dtype)` downcasts the returned scalar, both of which #736 pins as contracts.

#722 asked for exactly this and I did not enforce it at the time: *"Start with forward-KL; extend to reverse-KL and JS only after their stable log-space formulation is in place"*, and *"Reverse-KL/JS use the stable kernel rather than reproducing their current non-finite-gradient behaviour."*

**The fix is small.** Restore `s`/`t` in FP32 and write the three divergences in #736's log-space form — about 15 lines. I built that integration and ran it: **113 of 114 tests across both suites pass**, dense == chunked to 1e-10, finite gradients at `temperature=0.05` for every `chunk_size` x `use_checkpoint` combination. The one failure is item 3 below.

## 2. BLOCKING — `test_matches_pre_pr_reference_kernel` uses the defective formula as its oracle

The docstring says *"BLOCKING 3: pin against the pre-PR origin/main reference formula, not ourselves"*, so this exists because I asked you not to test the kernel against itself. That was right then. Once #736 lands it inverts: the embedded `_reference_pre_pr_distill` **is** the probability-space form #719 declares broken, and the test asserts your kernel reproduces it to 1e-6.

Worse, it will not protect you: at `temperature=2.0` on well-conditioned logits the two formulations agree, so this test **still passes** against the corrected kernel. It is neither catching the regression nor blocking the fix — it is simply pointing at the wrong reference.

Re-point it at a double-precision reference, the shape `test_divergence_matches_double_precision_reference` in #736's suite already uses.

## 3. `test_chunk_size_controls_invocation_count` is coupled to a torch primitive

It spies on `torch.log_softmax` and asserts exactly one call per chunk. The stable kernel makes two (`log_s` and `log_t`), so it fails after the fix. Either expect `2 * ceil(n / chunk)` or — better, and immune to the next formulation change — count `_chunk_kernel` invocations instead of a torch primitive.

## 4. Please add a low-temperature gradient test through the chunked and checkpointed paths

Nothing currently guards this, which is precisely why the defect rode in without a red test. `temperature=0.05`, assert finite gradients, parametrized over `chunk_size` and `use_checkpoint`.

## 5. Fragment and labels

`git mv changelog.d/0.74.0/743.added.md changelog.d/0.75.0/743.added.md`. Your current green CI predates v0.75.0 (run finished 17:26 UTC on 09-11; the release merged 08:26 today), so a re-run goes red on this anyway. Also both new `Field` descriptions say `(v0.74.0 #722)` — that label is now wrong.

Non-blocking: #722 asks for peak-VRAM and throughput on a matched configuration plus a crossover point; the PR has a retained-autograd-bytes test only. Your fragment correctly says "retained autograd tensor bytes" rather than claiming total VRAM saved, which is the honest hedge, so I am not treating the absent benchmark as a blocker.

## Ordering — and a trap I want you to see before you rebase

**#736 merges first.** In the other order `main + #743` merges with **zero conflicts** and silently restores #719 in both paths. Nothing would have stopped it. That is why this is sequenced rather than a matter of taste.

**When you rebase onto a `main` that contains #736, do not trust a clean-looking merge.** I simulated it. Git flags a conflict only in the divergence block; **the shared preamble auto-merges cleanly and is semantically wrong**: #736's side introduces `log_s` / `log_t` and deletes `s` / `t`, while everything in your side below still references `s` and `t` (`s_flat = s[mask]`, `s.reshape(...)`, `device=s.device`, `loss.to(dtype=s.dtype)`). Symmetrically, your side cleanly deletes `_masked_mean` while #736's side still calls it.

Both naive resolutions compile, import, and pass `py_compile`:

```
resolution = take mine   -> NameError: name 's' is not defined
resolution = take theirs -> NameError: name '_masked_mean' is not defined
```

It fails only when the kernel is actually called. So: after rebasing, re-introduce `s = student_logits.float() / temp` and `t = teacher_logits.float() / temp` explicitly, and run the low-temperature reproducer before pushing. With `s` in fp32 your existing `loss.to(dtype=s.dtype)` yields an fp32 scalar and #736's dtype contract is satisfied with no further change.

If the rebase fights you, say so here and I will work through the conflict with you rather than leave you guessing at what moved. This is a sequencing accident, not a reflection on the PR.

MakazhanAlpamys:
@kok-o — reviewed at `ac6ae6d1`. **Every blocking item from my 12:46 review is delivered**, and I verified each by execution or mutation rather than by reading. Then I did something that made your branch worse, and I want to state that part first rather than bury it under the good news.

**I merged #736 about twenty minutes ago, and your PR is now `CONFLICTING`.** That is a direct consequence of the ordering I told you to wait for, and the rationale I gave for it no longer holds.

I told you #736 had to land first because merging yours would silently reinstate #719's NaN gradient. Your 17:10 push independently ported the same log-space kernel — `(p_t*(log_t-log_s)).sum()`, `(p_s*(log_s-log_t)).sum()`, `logaddexp(log_s,log_t)-log(2)`, fp32 upcast — line-for-line the same math. I checked it the strongest way available: **#736's own `tests/test_issue719_stable_distill_divergence.py`, copied onto your head unmodified, passes 34/34**, including its `assert actual.dtype == torch.float32` return contract. So you fixed #719 too, in a PR that was told to wait for the PR that fixed #719, and the conflict you now have in `src/soup_cli/trainer/distill.py` and `docs/training.md` is a rewrite rather than a port.

That is my call and my cost, not yours. If you would rather I do the rebase myself so you are not paying twice for the same work, say so and I will — the resolution is mechanical now that both sides are the same kernel, and I would keep your commit and name the resolution in the merge body.

**What I verified on your head, for the record:**

```
tests/test_issue722_chunked_distill.py                    138 passed
tests/ -k "distill or uld or minillm"                     632 passed, 1 skipped
issue748 / issue755 / issue487 / cli_startup / issue621   285 passed
#736's suite, copied onto this head, unmodified            34 passed
ruff check src/soup_cli/ tests/                           clean
```

Zero removed assertions — `tests/` diffstat is `+664 / -0`.

**Five mutations, each anchor counted exactly once, each mutated module `py_compile`d AND actually imported with a real kernel call** before running the suite — an import rather than only a parse, because a `NameError` produces a false "caught":

| mutation | result |
|---|---|
| drop the final normalization | **30 failed** — every fp64-oracle parametrization |
| chunk reduction `.sum()` -> `.mean()` | **20 failed** — incl. both equivalence tests |
| drop cross-chunk accumulation (chunked path only) | **30 failed** — incl. all three bf16 dense-vs-chunked parity cases |
| chunking silently falls back to unchunked | 1 failed — `assert 2 == 16` |
| checkpointing silently disabled | 2 failed |

The third is the one I care about: it breaks *only* the chunked path, and the chunked-equals-dense property is pinned three independent ways, one of them (the fp64 oracle) external to the kernel. That was BLOCKING 3 last round and it is properly closed.

Both new config fields reach a real consumer — closure-captured at `setup()`, passed into `_compute_distill_term` at the only call site, and pinned by a test that drives the real `setup` + `compute_loss` and asserts `chunk_size == 128` / `use_checkpoint is True` arriving at the kernel. I also executed the schema gates rather than reading the validators: bool-as-int, zero, negative, float, non-bool checkpoint, `task=sft`, `uld_strategy`, `minillm_enabled` and `distill_mode: sequence` all refuse, and every message names the offending field.

**Three things before it lands:**

1. **`Resolves #722` will auto-close the issue on merge.** We agreed this lands as partial progress. AC5 is genuinely delivered now, but **AC3 is not** — neither of us has measured throughput on a matched configuration and the PR ships no benchmark. Please change the body to `Refs #722`.
2. **`distill.py:133` — the all-masked early return breaks the fp32 return contract.** `return student_logits.sum() * 0.0` returns bf16 in, bf16 out; every other exit is fp32. #736's `_masked_mean` returns fp32 on that path via `denom.clamp(min=1.0)`, and its suite does not cover an all-masked batch — so after the rebase, the contract it pins will have a hole in it that is yours to not introduce. One `.float()`.
3. The rebase itself, or my offer above.

**Non-blocking, noted rather than asked:** `distill.py:176-184`'s `if t_chunk.requires_grad:` branch is still dead (teacher logits always arrive under `no_grad`), and the invocation-count test now asserts `2 * ceil(n/chunk)` against a `torch.log_softmax` spy, so a future formulation that caches the teacher's log-softmax breaks the test without breaking the feature. I took your "either" branch on that one, so this is a note for later, not a request.

Sorry about the ordering. Landing #736 first was right on the evidence I had at 12:46; it stopped being right when you ported the same kernel, and I did not re-check before merging.

kok-o:
@MakazhanAlpamys — pushed updated head:

1. **Rebase & conflict resolution**: Rebased cleanly onto latest `main` (`fcca8ef`), resolving textual conflicts with #736 and preserving #742's `model_accepts_loss_kwargs = False` opt-out. Merge tree against `main` is clean (0 conflicts).
2. **FP32 return contract on all-masked batch**: Wrapped `distill.py:136` early exit in `.float()` (`(student_logits.sum() * 0.0).float()`), ensuring every exit path returns an FP32 scalar.
3. **PR description**: Updated to `Refs #722`.
4. **Verification**: 189 tests passed across `test_issue722_chunked_distill.py` (138), `test_issue719_stable_distill_divergence.py` (34), and `test_issue487_changelog_fragments.py` (17); snapshot/recipe test suite (2,357 passed); startup lightness guards passing; `ruff check` clean.

MakazhanAlpamys:
Verified `873aa640`. **Merge-ready** — all three blocking items closed, and the #736 question I was most worried about resolved the right way.

**The duplication is resolved by deletion, not by two copies.** `8e86f7f0` (#736) is an ancestor of this head, `_masked_mean` is gone entirely, and the three log-space formulas appear exactly once, inside `_chunk_kernel` — `grep -c "_masked_mean"` is 0 and `grep -c "logaddexp"` is 1. `merge-tree` against current `main` is clean, and the merged tree's `distill.py` and `docs/training.md` are **byte-identical** to the branch's, so nothing is waiting to resurface at merge time. #736's own suite passes unmodified here (34/34), and #742's `self.model_accepts_loss_kwargs = False` survived the rebase at `distill.py:603`.

**The fp32 return contract is executed, not asserted.** `distill.py:136 return (student_logits.sum() * 0.0).float()`, run on all three input dtypes — `bfloat16`, `float16`, `float32` all give `torch.float32` on both the normal and the all-masked exit.

**Mutations of the chunking arithmetic — both killed.** Anchors counted before replacing, each mutated module imported *and a real kernel call executed* before running the suite, reverted green each time:

- chunking becomes a no-op (`c_size = n_tokens`): 1 failed — `test_chunk_size_controls_invocation_count`.
- the off-by-one I asked for, dropping the last chunk: **47 failed**, including `test_matches_pre_pr_reference_kernel[True-7-*]` at a prime token count with checkpointing on, both mask-equivalence tests, and both bf16 dense-vs-chunked parity cases.

On the tolerance question I raised: the equivalence claim is **not** self-referential. `test_matches_pre_pr_reference_kernel` inlines an independent **float64 probability-space** implementation — `softmax` plus `log(p/q)`, never `log_softmax` — and asserts `< 1e-5` on both loss and gradient across `chunk_size ∈ {None,1,3,7,32} × checkpoint{False,True} × 3 divergences`. Dense-vs-chunked is `1e-6` fp32 and `5e-3` bf16. The off-by-one survived none of them, and the fp32-accumulation concern the loose bf16 band could not see is moot anyway now that `distill.py:124-125` upcasts both logit tensors before anything else.

`Refs #722` rather than `Resolves` is correct — AC3, matched-config throughput, is genuinely still unticked.

Suites: 172 passed on the two distill files, 670 passed on `-k "distill or uld or minillm"`, 285 passed across the consumer/backend/light-CLI/fragment/recipe guards. Baseline: the new tests give **138 failed** against `main`'s `distill.py` and `schema.py`. Zero deleted test lines (+664/−0); the only fixture change is two schema-forced lines in `recipe_config_snapshots.json`. No top-level torch, `test_cli_startup_is_light.py` green, and `torch.utils.checkpoint` under `no_grad` emits no `UserWarning` — I checked with `-W error::UserWarning`, so the eval path stays quiet.

Three nits, none blocking:

1. **`distill.py:194 return loss.to(dtype=s.dtype)` is a no-op that reads as the opposite of what it does.** `s` is `student_logits.float()/temp`, so `s.dtype` is always fp32. It reads as "cast back to the input dtype" while it happens to satisfy the fp32 contract. If anyone moves the upcast, this line silently breaks #736's guarantee — #736's suite would catch it, but the source misleads the person making the change. `return loss` says what is meant.
2. `distill.py:176 if t_chunk.requires_grad:` is still dead — already accepted.
3. `schema.py:2560` and `:2578` docstrings say `v0.74.0 #722`; these ship later. Cosmetic.

Only CI is left; the queue is at ~88 runs. I will merge as soon as it is green.

MakazhanAlpamys:
@kok-o — **holding this, and the reason could not be seen from either side's CI.** Your matrix is green. #943's matrix was green. The **merged tree** is not.

```
tests/test_issue720_distill_grad_accum.py::test_unequal_microbatch_lengths_match_for_live_token_distillation
NameError: name '_distill_chunk_size' is not defined   (distill.py:89)
```

#943 (`4b93b4ab`, token-weighted accumulation) landed on `main` today. Its `_compile_distill_trainer()` lifts the real `_DistillTrainer` class out of `distill.py` **by AST and `exec`s it in a synthetic namespace**, so it has to enumerate every free variable the class body closes over. You add two — `_distill_chunk_size` and `_distill_checkpoint` — and its `trainer_globals.update({...})` does not bind them.

It is a test-harness gap rather than a product bug, and it is still merge-blocking: whoever merges this turns `main` red.

**Two lines, in `tests/test_issue720_distill_grad_accum.py`:**

```python
"_distill_chunk_size": None,
"_distill_checkpoint": False,
```

With those: **40 passed** — exactly the 6 + 34 that #943's own commit message predicts. It is yours to fix rather than a main-side patch, since the free variables are yours, and it needs a rebase to pick #943 up.

Only one test fails rather than six because only the `token_distill=True` case reaches the KD call site; the others return early under `_sequence_mode=True`.

## The composition question, answered — they do not stack

This was the thing I was most worried about, and it is fine. The two normalisations sit at different levels:

- **your kernel** shifts `labels[:, 1:]`, masks `!= -100`, divides by `denom = mask.sum()` — a **per-microbatch mean**, semantically identical to the `_masked_mean` it replaced (same shifted mask, same divisor; the old `.clamp(min=1.0)` becomes an explicit empty-mask early return);
- **#943** multiplies that mean by `local_items / window_items`, where `local_items` is **the same shifted count** as your `denom`.

So the KD term is `(sum/local) x (local/window) = sum/window`, which is exactly #943's token-weighted full-batch contract. You left the mean semantics alone and #943 was written to sit on top of a microbatch mean.

**Confirmed by running it, not by reading it** — and in a configuration neither suite covers, since your `chunk_size` defaults to `None`:

| `_distill_chunk_size` | `_distill_checkpoint` | #943's equivalence test |
|---|---|---|
| 4 | False | passed |
| 4 | True | passed |
| 1 | True | passed |

The gradient-equivalence contract holds at chunk size 1 and under non-reentrant checkpointing. No double normalisation.

## Otherwise unchanged

`merge-tree` clean, your own 138 pass on the merged tree, the ratchets pass at 138, fragment correct with no BOM and no duplicate number. Mutation still dies: `c_size = chunk_size if (...) else n_tokens` to `c_size = n_tokens` — one match, module imports — fails `test_chunk_size_controls_invocation_count` on `assert 2 == 16`.

Two bindings and a rebase and this merges.

MakazhanAlpamys:
@kok-o — I found the cause of the hold and it is **two lines in a test file**. Diagnosis and a verified patch below; nothing here needs a redesign.

## Why both branches are green and the merge is not

`tests/test_issue720_distill_grad_accum.py::_compile_distill_trainer()` does not import `_DistillTrainer` — it lifts the class out of `distill.py`'s AST and `exec`s it in a namespace it builds by hand:

```python
namespace = {
    "Trainer": transformers.Trainer,
    "_sequence_mode": True,
    "_minillm_on_policy": False,
}
```

So every closure variable the real factory captures has to be listed there or in the per-test `trainer_globals.update({...})` at `:159`. Your PR introduces two new ones at `distill.py:587-588`:

```python
_distill_chunk_size = tcfg.distill_chunk_size
_distill_checkpoint = bool(tcfg.distill_checkpoint)
```

and the harness never binds them. Hence `NameError: name '_distill_chunk_size' is not defined`, and hence it is invisible to your CI and to #943's: the harness only breaks once **both** are in the same tree.

The `distill.py:89` in the traceback is a red herring — `exec(compile(..., "distill.py", "exec"))` stamps that filename onto the recompiled code object, so the line number indexes the *unparsed* class, not the real file.

## The patch, and why these two values

At `tests/test_issue720_distill_grad_accum.py:172`, after `"temperature": 2.0,`:

```python
            "_distill_chunk_size": None,
            "_distill_checkpoint": False,
```

Those are not values chosen to make it green — they are the **schema defaults** (`schema.py:1762` `default=None`, `:1770` `default=False`), so #720's test goes on exercising the unchunked, uncheckpointed path it was written for. Binding anything else would silently repurpose someone else's regression test into a test of your feature.

## Measured on the real merged tree, not on a simulation

`git merge-tree` of your `873aa640` onto current `main` (`61328f22`), materialised and run with the base interpreter:

| | before patch | after |
|---|---|---|
| `test_issue720_distill_grad_accum.py` | 1 failed, 5 passed | **6 passed** |
| `test_issue722_chunked_distill.py` (yours) | — | **138 passed** |
| every test file mentioning `distill` | — | **3804 passed, 8 skipped** |

And I checked the thing that would have made this patch harmful: the harness has a deliberate mutation control, `_compile_distill_trainer(without_token_weighting=True)`, which replaces the normaliser so #720's scaling bug must resurface. It still fires — `test_without_token_weighting_the_gradient_scales_with_the_step_count` passes — so the two bindings do not blunt the guard they sit next to.

## How you want to do this

It is your PR, so it is yours to push; say the word if you would rather I pushed the two lines to your branch and you will find them there. Either way, please add the `Refs #720` note in the commit body — a future reader of that harness deserves to know why those two keys exist, because the next closure variable anyone adds to the factory will hit exactly this again.

**Worth considering separately, not now:** the real defect is that the harness enumerates free names by hand, so it silently rots every time the factory gains one. Deriving the binding set (walk the `compute_loss` AST for free names, require each to be bound, fail loudly naming the missing one) would turn this class of breakage from a cross-PR `NameError` into a message that says what to add. I am not asking you to build that here — file it if you want it, or I will.

One more thing, unrelated to the fix: your CI on this PR is queued behind a backlog — Actions produced no terminal outcome at all between 2026-09-13T18:30Z and today 15:08Z. The 13 required contexts are currently green on `873aa640`, so once the two lines land it is a short path.

kok-o:
@MakazhanAlpamys — please feel free to push the two lines to my branch. Thanks for tracking down the harness gap and for the verified patch!

MakazhanAlpamys:
Pushed as `621846b8`, and **first an apology: you asked me to do this 12 hours ago and I did not answer until now.** You were explicit and I left you waiting; that is on me.

**It needed slightly more than the two lines, so here is exactly what I did and why**, because I would rather you could undo it than be surprised by it.

The bindings could not be applied to your branch alone. `tests/test_issue720_distill_grad_accum.py` on your base predates #943 (`4b93b4ab`), which rewrote it — the anchor I patched locally does not exist in your copy. **The `NameError` only exists on the merged tree**, so there was literally nothing to patch on the branch by itself.

So it is two commits, deliberately separate:

```
f1a0d270  merge: origin/main into feat/722-chunked-distill
621846b8  test(distill): bind the two new closure names the harness recompiles against
```

A **merge**, not a rebase — your history is untouched and both commits revert cleanly if you dislike either. The merge was clean (`Auto-merging schema.py / distill.py / recipe_config_snapshots.json`, no conflicts).

The fix itself:

```python
            "temperature": 2.0,
+           "_distill_chunk_size": None,
+           "_distill_checkpoint": False,
```

Anchor matched exactly once. **144 passed** across `test_issue720_distill_grad_accum.py` and your `test_issue722_chunked_distill.py` on the merged tree, and the harness's own mutation control — `_compile_distill_trainer(without_token_weighting=True)` — still fires, so the two bindings did not blunt the guard they sit beside.

The values are the **schema defaults** (`distill_chunk_size` default `None`, `distill_checkpoint` default `False`), not numbers chosen to make it green. That matters: #720's test goes on exercising the unchunked, uncheckpointed path it was written for. Binding anything else would have quietly repurposed someone else's regression test into a test of your feature.

**The real defect here is not yours and not #943's.** `_compile_distill_trainer()` lifts `_DistillTrainer` out of the AST and `exec`s it in a namespace built by hand, so it silently rots every time the factory gains a closure variable — and it rots only on the *merge*, which is the one place neither author's CI looks. Worth its own issue rather than another round of this one; I will file it.

CI is running on `621846b8`. If it is green I will merge without asking anything else of you.

Your other two are next in my queue and both have been waiting the same 12 hours — #918 (rebased, all findings addressed) and #921 (22 regressions resolved, but its matrix has gone red on the new head, which you will want to know before I say anything else about it).

MakazhanAlpamys:
Filed as #990, as promised in my last comment.

It is scoped to the harness rather than to your PR: the fix for *this* instance is already on your branch as `621846b8`. #990 is about the next one, because `_build_distill_trainer`'s closure has gained variables in at least three releases and the hand-maintained binding list will rot again.

The shape I proposed there is to derive the bindings instead of listing them — walk the recompiled `compute_loss` for free names, subtract what the namespace provides, and fail with a message that **names the missing binding** rather than raising `NameError` from inside the recompiled code at a line number that points into the wrong function. Two traps are written into the acceptance criteria so the next person does not hit them: bind the schema default rather than a literal, and do not report nested function definitions (`_token_weighted_accumulation`) as free names — a first pass at that analysis made exactly that mistake.

Labelled `good first issue`. If you want it, say so and it is yours; you already understand this harness better than most.

kok-o:
@MakazhanAlpamys — merge commit and the binding fix both look correct, thank you for the thorough write-up and for handling it.

The `type-check` failure is an infrastructure flake (runner acquisition), not a code issue — could you re-run failed jobs when you get a moment?

Also: I'd like to take #990. The AST-derived binding approach you outlined makes sense, and I'm already familiar with the harness internals from this PR. I'll open a draft once this one lands.

MakazhanAlpamys:
Merged. Thank you for the patience on this one — it sat longer than it should have, and not for anything wrong with the diff.

**`type-check`: re-run triggered, and you were right about the cause.** I checked rather than took it on report, because "infrastructure flake" is the easiest thing in the world to say about a red check. The job's `steps` array is **empty** and it burned 1h14m between `started_at` and `completed_at` — that is runner acquisition failing, not mypy finding something. A real type error would have a populated step list with exactly one failed entry. It is also not a required context, so it never blocked the merge.

**On `621846b8`.** For anyone reading this thread later: those two commits are mine, pushed to a contributor's branch at their explicit request, and @kok-o reviewed them before the merge rather than having them land unexamined.

The reason they were needed is the part worth keeping. Both matrices were green **separately**, and composing this head onto `main` after #943 produced `NameError: _distill_chunk_size` — a name that exists on neither branch alone. Git reported no conflict. Two green checkmarks are not evidence about their merge, and this PR is the clearest example of it I have.

---

## #990 — I have to turn you down, and I want to be straight about why

**@jagadeepmamidi claimed it at 11:54:37Z today. You asked at 13:01:33Z.** They were first by just over an hour, and their comment names the same scope you proposed — free-name detection, nested functions, regression tests, the compile filename. I am holding it for them.

I want to be plain that this is timestamps and nothing else. Your claim to that issue is strong on the merits: you have just spent a PR inside `_build_distill_trainer`, and you are the person who hit the rot. If it were a judgement call I would find it hard. It is not a judgement call, and "first claim in a comment wins" is only worth anything if it holds when the later claimant is better placed.

There is also history here I will not repeat: @jagadeepmamidi was turned away from #693 to hold a release deadline that then lapsed. I am not doing that to them twice.

**If you still want the work**, two honest routes:

1. **Say so on #990 as a second.** If they go quiet or would rather hand it over, it is yours and I will say so there. Please do not open a competing draft in the meantime — two people writing the same AST walker is exactly the waste the claim convention exists to prevent.
2. **#987** is open and unclaimed and is close to your recent ground: the repo-wide ANSI-safety scanner covers only `--help`, while `ReprHighlighter` also colours quoted strings and numbers *inside error messages*, so the class it was written to prevent is still live outside the one surface it walks. That class has turned CI red four times. Say the word and it is yours.

Sorry to trade you a merge for a no on the same day.

PR #744, opened 2026-09-07
feat(grpo): sequence-level GSPO objective with length normalization (#723)

### LINKED ISSUE #723: GSPO leaks masked tokens through batch centering and is not sequence-level

kok-o:
## What does this PR do?

Resolves #723 by upgrading the `gspo` GRPO variant from the legacy column-centering heuristic to the published **Group Sequence Policy Optimization (GSPO)** algorithm ([Qwen Team, arXiv:2507.18071](https://qwenlm.github.io/blog/gspo/)).

### Background
Issue #723 highlighted that `gspo` centered importance log-ratios per column across the batch, leaking masked padding tokens and differing from the sequence-level algorithm in the literature. While PR #735 (`e1f5f69`) excluded masked tokens from the batch centering statistic, this PR implements the full sequence-level objective.

### Key Changes
1. **Sequence-Level Likelihood Ratio**: Evaluates a length-normalized importance ratio per completion:
   $$s_i = \exp \left( \frac{1}{|y_i|} \sum_{t=1}^{|y_i|} (\log p_\text{new} - \log p_\text{old}) \right)$$
2. **Sequence-Level Surrogate Clipping**: Applies PPO/GRPO clipping directly on sequence ratios:
   $$\mathcal{L}_\text{GSPO} = -\frac{1}{G} \sum_{i=1}^G \min\left( s_i \widehat{A}_i, \mathrm{clip}(s_i, 1 - \varepsilon, 1 + \varepsilon) \widehat{A}_i \right)$$
   with default radius $\varepsilon = 0.2$ (or operator-provided `delta`).
3. **Strict Mask Isolation**: Excludes all-masked tokens and sequences prior to all reductions; masked positions receive strictly `0.0` gradient.
4. **Permutation & Padding Invariance**: Strictly invariant under batch row permutations and padding positions (left vs right).
5. **Advantage Shape Flexibility**: Seamlessly supports 1D `(B,)`, 2D column `(B, 1)`, and broadcast `(B, T)` advantage tensors.
6. **Documentation & Changelog**: Updated `docs/training.md` and added `changelog.d/0.74.0/723.fixed.md`.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

### DIFF
--- a/changelog.d/0.74.0/723.fixed.md
+++ b/changelog.d/0.74.0/723.fixed.md
@@ -1,0 +1,7 @@
+- **Sequence-level GSPO objective (#723).** Upgraded `grpo_variant: gspo` from the
+  legacy column-centering heuristic to the published Group Sequence Policy Optimization
+  algorithm (Qwen, arXiv:2507.18071). Evaluates a length-normalized sequence importance
+  ratio $s_i = \exp(\frac{1}{|y_i|} \sum_{t=1}^{|y_i|} (\log p_\text{new} - \log p_\text{old}))$
+  and applies sequence-level surrogate clipping with default radius $\varepsilon = 0.2$
+  (or operator-supplied `delta`), guaranteeing exact zero gradient at all masked tokens
+  and strict batch-permutation invariance.
--- a/changelog.d/0.74.0/723.fixed.md
+++ b/changelog.d/0.74.0/723.fixed.md
@@ -1,7 +1,0 @@
-- **Sequence-level GSPO objective (#723).** Upgraded `grpo_variant: gspo` from the
-  legacy column-centering heuristic to the published Group Sequence Policy Optimization
-  algorithm (Qwen, arXiv:2507.18071). Evaluates a length-normalized sequence importance
-  ratio $s_i = \exp(\frac{1}{|y_i|} \sum_{t=1}^{|y_i|} (\log p_\text{new} - \log p_\text{old}))$
-  and applies sequence-level surrogate clipping with default radius $\varepsilon = 0.2$
-  (or operator-supplied `delta`), guaranteeing exact zero gradient at all masked tokens
-  and strict batch-permutation invariance.
--- a/changelog.d/0.74.0/744.changed.md
+++ b/changelog.d/0.74.0/744.changed.md
@@ -1,0 +1,12 @@
+- **Sequence-level GSPO objective replaces legacy column-centering (#744, closes #723).**
+  Upgraded `grpo_variant: gspo` from the legacy column-centering heuristic to the
+  published Group Sequence Policy Optimization algorithm (Qwen Team, arXiv:2507.18071).
+  Because the mathematical objective is replaced at the sequence level, existing
+  configurations specifying `grpo_variant: gspo` will yield different numerical losses
+  and gradients on upgrade compared to prior versions and will not reproduce prior runs.
+  The objective now computes a length-normalized sequence importance ratio
+  $s_i = \exp\left(\frac{1}{|y_i|} \sum_{t=1}^{|y_i|} (\log p_\text{new} - \log p_\text{old})\right)$
+  and applies sequence-level surrogate clipping with default radius $\varepsilon = 0.2$
+  or operator-supplied `grpo_delta`, strictly isolating masked tokens with 0.0 gradient,
+  remaining invariant to padding tokens and placement (left vs right), and respecting
+  batch permutation invariance.
--- a/docs/training.md
+++ b/docs/training.md
@@ -524,9 +524,9 @@ task: grpo
 training:
   reward_fn: accuracy
   num_generations: 4
-  grpo_variant: gspo         # group-stabilised importance ratio
+  grpo_variant: gspo         # sequence-level length-normalized ratio
   # or: dapo / dr_grpo / bnpo / rft / two_sided
-  # grpo_delta: 0.2          # required when grpo_variant=two_sided
+  # grpo_delta: 0.2          # required when grpo_variant=two_sided; optional for gspo
 ```
 
 Variants:
@@ -839,7 +839,7 @@ training:
   num_generations: 8
   # New: GRPO objective variants
   grpo_variant: dapo                  # one of: gspo / dapo / dr_grpo / bnpo / two_sided / rft / standard
-  # grpo_delta: 0.2                   # required when grpo_variant: two_sided
+  # grpo_delta: 0.2                   # required when grpo_variant: two_sided (optional for gspo)
   grpo_fp16: true                     # FP16 RL (unsloth parity)
   # Long-context + memory-efficient RL
   long_context_grpo: true             # wires Tiled MLP when available
--- a/docs/training.md
+++ b/docs/training.md
@@ -532,7 +532,7 @@ training:
 Variants:
 
 - **standard** — DeepSeek-R1-style baseline (delegates to TRL's `compute_loss`).
-- **gspo** — group-stabilised importance ratio with per-batch control variate.
+- **gspo** — Group Sequence Policy Optimization (sequence-level length-normalized ratio with clipping).
 - **dapo** — decoupled asymmetric clipping (`eps_lo=0.2, eps_hi=0.28`).
 - **dr_grpo** — token-sum without per-sample length normalisation.
 - **bnpo** — length-normalised PPO surrogate.
--- a/src/soup_cli/config/schema.py
+++ b/src/soup_cli/config/schema.py
@@ -1500,8 +1500,9 @@ class TrainingConfig(BaseModel):
         gt=0.0,
         le=1.0,
         description=(
-            "Symmetric clipping radius for grpo_variant='two_sided'. "
-            "Required when grpo_variant='two_sided'; rejected otherwise."
+            "Symmetric clipping radius for grpo_variant='two_sided' (required) "
+            "or grpo_variant='gspo' (optional sequence clipping radius, "
+            "defaults to 0.2). Rejected for all other variants."
         ),
     )
     grpo_fp16: bool = Field(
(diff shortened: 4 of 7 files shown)


MakazhanAlpamys:
@kok-o — **the math is right, the tests kill every failure mode #723 names, and #735 is preserved rather than reverted. Three small things and I merge.**

## The objective checks out against an independent reference

I wrote the published GSPO objective from the paper — `s_i = exp((1/|y_i|) sum_t delta_logp)`, `L = -(1/G) sum min(s_i A_i, clip(s_i, 1±eps) A_i)` — and swept 300 randomised cases with variable batch, variable length, and **non-contiguous** mask positions:

```
max |your implementation - independent reference| = 1.78e-15
```

The two properties the review was really about:

- **The normalisation divides by the unmasked token count, not the padded width.** With pad columns carrying 7.0 and -9.0 the result equals the unmasked mean, never the padded one. That is #735's defect one step to the left and it is not present.
- **The clipping is sequence-level, not token-level.** With per-token ratios `e^3, e^-3, e^0.6` your result is `-1.2000000000`; token-level would give `-0.8165956895`. Different numbers, and yours is the one the paper specifies.

Masked-token gradients are exactly `0.0` at pad magnitudes up to `1e6`, an all-masked row is excluded from the mean rather than poisoning it, and left-pad and right-pad agree to the last digit.

## You did not revert #735 — you strengthened it

#735's centering code is gone as code, which looks alarming until you see why: #735 fixed the mask leak in the batch-centering statistic, and the published objective has no such statistic. `changelog.d/0.74.0/735.fixed.md` is intact, and #735's own tests pass against your code.

The measurable part: on `main`, mutating a masked log-ratio to `1000.0` yields `nan`, because `exp(1000 - mean)` overflows and `inf * 0` follows. Under yours the loss is bit-invariant across pad values `0 / -1e3 / +1e3 / +1e6`. So @AmirF194's fix held the line and yours removes the exposure entirely.

Four mutations, four killed: length normalisation dropped, normalisation by padded length, token-level clipping, and a full revert to the pre-PR objective. All four are exactly the failure modes #723 describes.

## MEDIUM — the delta knob you document cannot be set

The gspo branch reads `delta` and calls `validate_grpo_delta(delta)`, and both the fragment and the new docstring say "or operator-supplied `delta`". But this PR does not touch `schema.py`, and the cross-validator there still rejects it. I ran it:

```yaml
training: {grpo_variant: gspo, grpo_delta: 0.1}
```
```
REJECTED: training: Value error, grpo_delta is only valid when
          grpo_variant='two_sided'; got grpo_variant='gspo'
```

`trainer/grpo.py:538` only sets `_soup_grpo_delta` when `tcfg.grpo_delta is not None`, so **eps is always 0.2 on every real run** and that branch is unreachable outside a direct library call. `test_custom_delta_sets_clipping_radius` calls `apply_variant_loss` directly, so it passes while the user-facing path fails at parse.

Either extend the cross-validator to accept `grpo_delta` for `gspo`, or delete the branch and the claim. This is release-checklist step 6c — a real `soup.yaml` setting the field to a non-default value — and it is the check that catches exactly this.

It matters more than a dead branch usually would, because **0.2 is a token-level clip radius applied to a sequence-level ratio.** GSPO's whole premise is that sequence ratios are far less dispersed than token ratios; the paper uses values around `3e-4`. At 0.2 the clip is close to inert, and the tunable that would let an operator fix that is the one that does not work.

## LOW — three stale strings and the changelog category

`docs/training.md:527` still carries `grpo_variant: gspo  # group-stabilised importance ratio`, in the same YAML block whose bullet list you did update. `grpo_variants.py:10` still says "Group Stabilized Policy Optimization (unsloth)". And the new `_VARIANT_METADATA` description reads "Group Sequence Policy Optimization (group-stabilized)", which contradicts itself now that there is no group stabilisation.

The fragment is `723.fixed.md`. Two things: it should be **`744.*`** per `changelog.d/README.md`, which asks for the PR number so two PRs against one issue stay distinct — and this issue has exactly that situation, since #735 already shipped against it. And the category should be **`changed`**, not `fixed` or `added`: `gspo` already existed and its numerical meaning is replaced, so anyone with `grpo_variant: gspo` in a config gets a different objective on upgrade and will not reproduce prior runs. That deserves saying in words in the entry.

## Not a change request

`log_ratio * mask` is still `0 * inf -> nan` if a masked position ever carries a non-finite log-prob — same exposure as `main`, not a regression. A fully-masked batch returns `0.0` silently where #723 lists all-masked rejection among the cases; your test asserts stability instead, which is defensible but is not what the issue asked. And `s = torch.exp(seq_log_ratio)` is a single-letter name that `ruff` allows but this repo's conventions do not.

Half your tests pass against unfixed `main` — but those are invariance properties #735 genuinely already delivered, so they are regression guards rather than filler, and one of them dies under the padded-length mutation, so it discriminates a different wrong answer.

Ruff clean, 14 of 14 green with `head_sha` confirmed, nothing deleted or weakened. Merges clean with #743 and with #736.

Fix the delta question — either direction is fine — rename and recategorise the fragment, and correct the three strings, and this goes in.

MakazhanAlpamys:
Merged as `0139553d`. Thank you — this is a rewrite of a published objective, not a patch, and it holds up under measurement.

**Verified at merge rather than read.** Isolated worktree on the exact head `9043b2b0`, `PYTHONPATH` confirmed to resolve `soup_cli` into that worktree and not the main checkout, mutations applied in binary mode with occurrence-count asserts and sha256 before and after (this repo is CRLF and a text-mode replace silently no-ops), baseline hash `1eff771f458db6f7` restored and `git status` clean afterwards.

```
baseline                                          88 passed in 4.38s

drop the length-normalization divisor:
  sha 1eff771f458db6f7 -> 23b399850c226de5
  FAILED test_masked_tokens_receive_zero_gradient
  FAILED test_reference_sequence_objective_matches_manual_calculation
  FAILED test_unequal_sequence_lengths_normalized_independently
  FAILED test_runtime_grpo_trainer_threads_custom_delta

reduce over ALL tokens instead of unmasked (reintroduce the leak):
  FAILED test_masked_token_mutation_leaves_loss_invariant
  FAILED test_masked_tokens_receive_zero_gradient
  FAILED test_reference_sequence_objective_matches_manual_calculation
  FAILED test_unequal_sequence_lengths_normalized_independently
```

Both killed by name, and the tests distinguish the *right* wrong answers: dropping the divisor and using a constant `T` instead of the per-sequence `|y_i|` are different mistakes and both fail. `exp(mean_t(Δlogp))` rather than `mean_t(exp(Δlogp))` is the geometric-mean ratio the paper specifies, and that is the part a reimplementation usually gets wrong.

## Three things worth naming

**The issue and your PR describe different defects, and yours is the real one.** #723 named a masked-token leak through batch centering — but that half had already been closed by #735, and checking `main` directly confirms masked gradients are already zero there. What #735 left behind is what you actually fixed: column centering zeroes the **policy** gradient on any column with a single unmasked member, so 2 of 9 unmasked positions in an ordinary fixture were learning nothing at all. Nine of your sixteen tests fail on unmodified `main`. That is what separates a rewrite from a restatement, and it is why I did not treat this as duplicate work.

**You caught a dead path of your own making, between your own two commits.** The first added `delta` handling to the gspo kernel while the schema still rejected `grpo_delta` for everything but `two_sided` — unreachable from any real `soup.yaml`. Finding that in your own diff and fixing the schema rather than leaving code nobody could call is the part I would have been least likely to catch, because the kernel reads correctly on its own.

**The three removed lines in `tests/test_v0500_part_a.py` are schema-forced, not bent.** I checked each one. The delta-rejection test is retargeted to a variant that must still reject and keeps its `match=` assertion, and you added the positive case beside it. Mutating the schema back to `!= "two_sided"` fails the new tests; removing the guard entirely fails the old one. Both directions covered.

The changelog fragment stating outright that existing `gspo` runs will not reproduce prior numerics is the right call. That is a breaking change for reproducibility and burying it would have been easy.

## Two one-liners I applied myself in `f3dc2d4`

Neither is a criticism; both are mine to catch at merge time.

1. **The fragment carried no `by @kok-o`.** 25 of the 27 fragments in `changelog.d/0.74.0/` have it. The release checklist says attribution happens at merge, not reconstructed at release — v0.74.0 found 56 of 116 entries uncredited precisely because it was deferred, and this was on course to be the 57th.
2. **`apply_variant_loss`'s docstring still said `delta` is "only used by `two_sided`".** Your "fix stale strings" commit caught the others and missed this line; it is now also the sequence-level clipping radius for `gspo`.

## One follow-up, and it is not yours

`apply_variant_loss` calling `validate_grpo_delta` is unit-tested directly but no test asserts that any *kernel branch* actually calls it — replacing the call with a bare `float(delta)` survives the suite. I ran the identical mutation against the pre-existing `two_sided` guard as a control and **it survives too**, so this is a project-wide gap rather than anything you introduced. Impact is bounded: the Pydantic field carries `gt=0.0, le=1.0`, so an out-of-range delta cannot reach the kernel from a real config, and the guard does work when called directly. Worth a one-line test for both branches at some point; not yours to carry.

Credited in `CONTRIBUTORS.md` at `bb0ae6b`.

Separately: `eps` defaults to `0.2`, while the GSPO paper uses sequence-ratio clipping ranges around 3e-4 because the sequence ratio has far lower variance than the token-level one. `0.2` is consistent with the other kernels in the file and is now operator-overridable thanks to your schema fix, so it is a defensible default — but if you wanted to add a sentence to `docs/training.md` telling users why they might actually want to reach for `grpo_delta` on `gspo`, that would be a genuinely useful small follow-up.

PR #746, opened 2026-09-08
fix(eval): stamp soup ship judge numerics and gate mismatched evidence

### LINKED ISSUE #367: live_eval.load_model_and_tokenizer takes no quantisation argument: an NF4 run is judged as an NF4 adapter on a bf16 base

jagadeepmamidi:
The remaining #367 criteria: record the actual live-eval load on the verdict and refuse evidence replay under a different numerics family.

## What does this PR do?

`soup ship` already reused `training.quantization` for the live load (#461 / #570), but the judge's precision never appeared in the verdict or in `--emit-evidence`. Absolute leg-1 / leg-2 numbers for a 4-bit run were therefore recorded as if the precision were implicit — a correctness-of-absolute-numbers bug, not a claim that the paired NF4 measurements were wrong.

This stamps the **actual** load (`4bit` / `8bit` / `bfloat16` / `float32`) on:

- the verdict panel / rubric
- `--output`
- `--emit-evidence` as a top-level `numerics` field

`--config` + `--evidence` then refuse a **family** mismatch (`4bit` | `8bit` | `full`) the same way `config_sha` refuses a drifted recipe. `bfloat16` and `float32` are the same `full` family (the bug is NF4 vs bf16, not bf16 vs fp32). Pre-#367 evidence without a stamp warns rather than failing closed. `--evidence --emit-evidence` copies an existing stamp through and does not invent one.

Tests assert the stamp matches the quantization that reached the loader, not merely that the call does not raise. No GPU: the live path is mocked at `_resolve_generators`; the gate is pure JSON.

Rebased onto `main` after #768. MCP keeps `McpToolError(_evidence_error_message(exc))` — not `str(exc)`. The borrowed `768.changed.md` fragment is dropped; this PR's fragment is `changelog.d/0.75.0/746.fixed.md`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [x] Focused pytest for this change passes (`tests/test_issue367_ship_numerics.py` plus existing `soup ship` suites). Full `pytest tests/ -v` not re-run here.
- [x] Updated relevant docs (`docs/commands.md`, `docs/evaluation.md`)
- [x] Changelog fragment: `changelog.d/0.75.0/746.fixed.md`

Closes #367


MakazhanAlpamys:
@jagadeepmamidi — this has been green for a day with no word from me. It is under review now, and you will get findings rather than a verdict.

Two things up front so you know what the review is weighted on, because this PR is not an ordinary one:

`soup ship` is a merge gate — `soup ci init` wires `ship --evidence` into a workflow where exit 2 blocks a pull request. So the question I will spend the most effort on is not whether the new behaviour is correct in the happy case, but **whether the gate became more permissive in any input case**. I will build evidence files that produced DON'T SHIP on `main` and check they still do.

The second is a defect class this repo has shipped before and is watching for: `src/soup_cli/mcp_server/registry.py` exposes a `ship_evidence` tool that consumes the **same** evidence schema. A previous release shipped one evidence file replaying to two different verdicts — one through the CLI, one through `soup mcp serve` — because the MCP tool did not read a newly added key. If your change adds a key to that schema, that tool needs it too.

Also on the list, none of them accusations, just what gets checked here: the eight deleted lines in `tests/test_issue367_live_eval_quantization.py`; whether new fields read from the untrusted evidence file refuse rather than silently drop, which is the precedent `noise_floor_from_evidence` set for exactly this reason; and whether `--emit-evidence` output still round-trips back through `--evidence` to the same verdict.

Nothing is asked of you yet.

MakazhanAlpamys:
@jagadeepmamidi — **the work is careful, tightly scoped, and the safety property that matters most is intact: this never makes the gate more permissive. Blocking on one test and one three-line addition, and the reason the test is blocking rather than a nit is that its failure mode is `exit 0`.**

## First, the thing I spent the most effort on: the gate does not loosen

Three independent checks, because this is a merge gate and `soup ci init` wires `ship --evidence` where a nonzero exit blocks a PR:

- **The decision functions are byte-identical to `main`.** `decide_ship`, `build_task_win`, `compute_benchmark_deltas`, `compute_noise_floor`, `noise_floor_from_evidence` — same hashes. `numerics` never reaches `decide_ship`; it is stamped onto the frozen verdict afterwards. That is the right design and it is what makes the rest checkable.
- **120 randomized evidence files, `main` versus your head.** Zero cases where `main` exited nonzero and yours exited 0. Sixteen moved `2 -> 1` — all DON'T-SHIP evidence carrying a malformed stamp, now hard-refused. Strictly more restrictive.
- Hand-built leg-1 and leg-2 DON'T SHIP cases still exit 2, before and after a round trip.

The round trip is clean, `--evidence --emit-evidence` copies a stamp through without inventing one, the refusal never echoes the rejected value, and the exit taxonomy is respected on both paths. Your malformed-input reader follows the `noise_floor_from_evidence` precedent — **raises rather than drops** — which is exactly right.

Nine mutations, eight killed, including the two that would have been the worst possible findings: the mismatch gate forced to always accept, and the gate's exit code changed to 0.

## BLOCKING 1 — the ninth mutation survives, and it turns a refusal into a SHIP

I made `numerics_from_evidence` swallow the error and return `None` instead of raising:

```
tests/test_issue367_ship_numerics.py    26 passed      <- unchanged
```

Then drove the real CLI with an evidence file carrying `"numerics": "nf4"`:

```
shipped code:  Error: invalid evidence.numerics: numerics must be one of
               4bit, 8bit, bfloat16, float32          exit 1
under mutation: "Task improved AND no general benchmark regressed past 5.00%"
                                                       exit 0   <- SHIP
```

**Your code is correct.** The problem is that the property it depends on has nothing holding it up. `numerics_from_evidence` is exercised exactly once in the whole suite:

```
tests/test_issue367_ship_numerics.py:97:  assert numerics_from_evidence(None) is None
```

— the absent case only. The `--config` gate calls `parse_numerics` directly and bypasses this function, so the no-`--config` path is uncovered. A future refactor to silent-drop ships green, and corrupt evidence then reads as SHIP on a merge gate.

Two lines closes it:

```python
with pytest.raises(ValueError):
    numerics_from_evidence("nf4")
```

plus one CLI test asserting `--evidence <malformed>` with **no** `--config` exits 1.

## BLOCKING 2 — the MCP tool now reports something false, and this PR is what made it false

`src/soup_cli/mcp_server/registry.py::tool_ship_evidence` consumes the same evidence schema and was not updated. Same two files through both readers:

```
                     CLI                    MCP ship_evidence
numerics: "4bit"     exit 0, numerics=4bit  decision=SHIP, numerics=None
numerics: "nf4"      exit 1, REFUSED        decision=SHIP, numerics=None
```

The first row is the part that is on this PR rather than pre-existing: you added `"numerics"` to `verdict_to_dict`, so the MCP tool now emits that key and reports `null` for evidence that **is** stamped. That is not a missing feature, it is an actively wrong statement about which numerics produced the verdict.

The second row is the recurring shape. The fix pattern is already in that file, ten lines up, written for v0.73.2's `noise_floor` — including the comment explaining why:

> *"Dropping it here would make the SAME evidence file replay to a DIFFERENT verdict through the MCP tool than through the CLI."*

Mirroring those three lines for `numerics` is the whole change.

## Not blocking, and the first one is a genuine credit

**The eight deleted lines are pure prose** and I checked them rather than assuming. All eight are docstring and comment text saying "criteria 2 and 4 are left open" — which this PR closes, so removing them is accurate. Assertion and test-def counts are identical at 70 before and after. Nothing was weakened.

**Zero of your 27 new tests pass against unfixed `main`** — the module does not even import there. Blunt, but it means nothing in that file is decorative.

Three LOW items, take or leave:

- `render_ship_panel` uses `escape()` on `ShipVerdict.numerics`, which neutralises `[...]` and nothing else. Unreachable through the CLI because `parse_numerics` gates every path, but `for_terminal` exists in that same file for exactly this and the v0.73.2 comments above it set the precedent.
- `--evidence --config --emit-evidence` copies a mismatched stamp onto the new provenance and exits 0; the artifact it writes is then refused by its own gate. It fails closed, so this is a warning-at-emit-time suggestion, not a defect.
- `docs/commands.md` ties the numerics stamp to `--config`; a live run stamps it with no `--config` at all, which I verified. `docs/evaluation.md` states it correctly.

## Housekeeping

Fragment named after the PR, crediting you. Ruff clean. 14 of 14 SUCCESS on `c1d1be09`, `head_sha` confirmed. 1165 passed across every ship-touching suite; the 2 failures reproduce identically on `main` and are the known #651 torch-floor class.

**Both blockers are small.** If you would rather not spend another evening on them, say so and I will write both and merge with the work credited to you — your commits, your `CONTRIBUTORS.md` entry, your name in the fragment. That offer is open to everyone and it is not a hint that you should take it.

Separately, I will file the broader item as its own issue: nothing tests CLI/MCP parity for **any** evidence key, and a schema-wide parity test is worth more than a one-key patch. That one is not yours to carry.

MakazhanAlpamys:
Re-reviewed with fresh measurements, and the numerics work itself is solid. Three mutations on your branch, each reverted afterwards:

- `_live_eval_numerics` stops returning the quant name so the stamp lies -> **2 failed**, including `test_live_4bit_config_stamp_matches_the_load`.
- `_check_evidence_numerics` returns immediately, disabling the gate -> **4 failed**: both staleness cases, the warn case and the escape-hygiene case.
- `numerics_family` collapses `4bit`/`8bit` into `full` so the gate never fires -> **2 failed**.

So the tests assert the stamp matches the quantization that actually reached the loader, which is what the PR claims. No coverage was removed — the 8 deleted lines are docstring prose re-pointing criteria 2 and 4 at the new file. `-k ship` 181 passed, `-k "mcp or evidence or verdict or eval"` 878 passed, ruff clean, docs and changelog fragment both correct. I also confirmed the scope maps cleanly onto #367: criteria 1 and 3 shipped in #461/#570, and this delivers 2 and 4.

**Blocking, and it is not a criticism of your scoping: `mcp_server/registry.py` is untouched, so `numerics` reads differently through the two surfaces.** Reproduced on your head:

```
evidence numerics:"4bit"  ->  CLI exit 0, "Judge numerics: 4bit"   |  MCP decision=SHIP, numerics=None
evidence numerics:"nf4"   ->  CLI exit 1, REFUSED                   |  MCP decision=SHIP, numerics=None
```

The first row is worse than a missing feature: because `verdict_to_dict` now emits a `"numerics"` key, the MCP tool actively reports `null` for evidence that *is* stamped. The second is a refusal on one surface and a SHIP on the other, from one file.

This is a known defect class rather than an oversight on your part — I filed #758 about it while reviewing this PR, precisely so it would not ride on one contributor. **Please do not patch it here with a one-key fix.** #768 lands a single shared evidence decoder and I am merging it first; rebasing onto it gives you MCP parity for free.

After #768 lands the rebase is four steps:

1. Add `"numerics"` to `EVIDENCE_SCHEMA_FIELDS["root"]` in `utils/ship_verdict.py`. Without this the strict reader refuses every file your `--emit-evidence` writes: `Error: evidence has unsupported field(s): 'numerics'`, exit 1.
2. Move the `numerics_from_evidence(...)` decode out of `commands/ship.py::_verdict_from_evidence` and into `ship_verdict.verdict_from_evidence`, returning `replace(verdict, numerics=...)` there. Delete the CLI-side copy — that copy *is* the drift.
3. Add a corpus entry carrying `numerics` to `EVIDENCE_CORPUS` and a malformed one to `MALFORMED_CORPUS` in `tests/test_issue758_evidence_schema_parity.py`. The registry is checked against the corpus, so step 1 without this fails `test_corpus_covers_every_registered_evidence_field` — I verified that by mutation.
4. Leave `_check_evidence_numerics` CLI-only. That is correct: the family gate needs `--config`, which the MCP tool has no equivalent of, exactly as the existing `config_sha` gate is already CLI-only.

For reference I resolved the merge by hand taking #768's side of `commands/ship.py` and ran both suites: all 25 parity tests pass and 14 of your 26 numerics tests fail. That is the whole rebase burden — the four steps above and nothing more.

One last thing: since criteria 1 and 3 already shipped and this delivers 2 and 4, please change the trailer from `Refs #367` to `Closes #367` so the issue closes on merge.

jagadeepmamidi:
Followed the four rebase steps rather than a one-key MCP patch:

1. `"numerics"` is on `EVIDENCE_SCHEMA_FIELDS["root"]`
2. Decode lives in `verdict_from_evidence`; the CLI-side copy is gone
3. Parity corpus has a stamped `numerics` entry and a malformed `"nf4"` one
4. `_check_evidence_numerics` stays CLI-only

Trailer is `Closes #367`.

#768 isn’t on `main` yet, so this branch currently includes it. Happy to rebase onto `main` once #768 lands.

MakazhanAlpamys:
Acknowledged, and thank you for taking the four-step route rather than the one-key patch — moving the decode into `verdict_from_evidence` and deleting the CLI-side copy is the part that actually closes the drift, and it is the part a shortcut would have skipped.

You are right that this branch carries #768's commits until it lands. That is the correct order and I am not asking you to rebase again yet: I have posted the queue shape on #768 so the wait is visible to its author, and offered to make the three remaining changes there myself if they would rather I did. Once #768 is on `main`, rebase and I will review this on its own diff.

Two things I will re-check when I do, so you know what to expect rather than being surprised:

1. That `numerics` on `EVIDENCE_SCHEMA_FIELDS["root"]` plus the corpus entries actually make the parity harness cover it — I verified by mutation that registering a key without corpus support fails `test_corpus_covers_every_registered_evidence_field`, so this is checkable rather than a matter of opinion.
2. The CLI/MCP divergence I measured on your previous head:

```
evidence numerics:"4bit"  ->  CLI exit 0, "Judge numerics: 4bit"   |  MCP decision=SHIP, numerics=None
evidence numerics:"nf4"   ->  CLI exit 1, REFUSED                   |  MCP decision=SHIP, numerics=None
```

Both rows should be gone once the shared decoder is the only reader. I will re-run exactly that.

`Closes #367` is right — criteria 1 and 3 shipped in #461/#570 and this delivers 2 and 4.

MakazhanAlpamys:
Two things to fix on the next rebase, one of them worth a careful look.

**1. Your branch carries someone else's changelog fragment.** This PR's file list includes both `changelog.d/0.74.0/746.fixed.md` (yours) and `changelog.d/0.74.0/768.changed.md`, which belongs to **#768** by @ousamabenyounes. That is almost certainly a rebase or branch-point artifact rather than an intentional addition — but if it merged it would land #768's changelog entry under your PR, crediting the wrong person for work that has not merged yet. Please drop it.

**2. Your own fragment is now in the wrong directory, and that part is not your doing.** v0.75.0 merged to `main` (`f50a7da3`) at 08:26 UTC and emptied `changelog.d/0.74.0/`, so `test_repository_fragments_match_the_newest_release` fails on the next CI run for sixteen open PRs including this one.

```
git rm changelog.d/0.74.0/768.changed.md
git mv changelog.d/0.74.0/746.fixed.md changelog.d/0.75.0/746.fixed.md
```

**`main` is still frozen** — the release is merged but not yet tagged, so nothing merges before the tag anyway. No rush.

MakazhanAlpamys:
@jagadeepmamidi — a sequencing note before anything else, because it affects credit and I would rather say it early and plainly than sort it out at merge time.

**This PR currently contains all of #768.** Comparing the two file sets:

| | #768 (@ousamabenyounes) | #746 (this PR) |
|---|---|---|
| `src/soup_cli/commands/ship.py` | yes | yes |
| `src/soup_cli/mcp_server/registry.py` | yes | yes |
| `src/soup_cli/utils/ship_verdict.py` | yes | yes |
| `tests/test_issue758_evidence_schema_parity.py` | yes | yes |
| `docs/evaluation.md` | yes | yes |
| `changelog.d/…/768.changed.md` | yes (`0.75.0/`) | yes (`0.74.0/`) |

`8c6ca26` ("Merge PR #768 and wire numerics into the shared evidence decoder") says exactly what you did, so there is nothing underhanded here — you needed the shared evidence decoder to build the numerics on, it was open and unmerged, and you took it and said so in the commit message. That is the honest way to handle the situation and I would rather have this than a silent copy.

**But it cannot merge in this order.** If #746 lands first, @ousamabenyounes's refactor lands under your PR number, their fragment ships with your change, and #768 becomes an empty or conflicting PR after weeks of review. Attribution is the only thing this project can pay contributors in, and that outcome would spend someone else's.

**So the order is #768, then this.** Concretely:

1. #768 merges. Its author pushed the last review items 20 minutes ago and it is close.
2. You `git fetch origin && git rebase origin/main`. Everything from #768 disappears from your diff because it is already upstream — `ship.py`, `registry.py`, `ship_verdict.py`, `test_issue758_evidence_schema_parity.py` and `docs/evaluation.md` shrink to just your numerics changes, and **`changelog.d/0.74.0/768.changed.md` should be deleted from your branch entirely** (theirs is already at `0.75.0/`, which is where fragments live now that the release is tagged).
3. Your own fragment moves: `git mv changelog.d/0.74.0/746.fixed.md changelog.d/0.75.0/746.fixed.md`. No content edit.
4. What remains is #367's numerics work — `test_issue367_ship_numerics.py`, `test_issue367_live_eval_quantization.py`, `docs/commands.md` and your changes to the three sources — which is the PR I will review on its own terms.

I have not reviewed the numerics work yet, deliberately: reviewing it while it carries someone else's diff would mean measuring your change against the wrong baseline, and I would end up crediting or blaming the wrong person for parts of it. The rebase makes that review honest, and it should also make it shorter.

Note the rebase may not be clean — @ousamabenyounes changed things in review after you merged their branch, so their upstream version differs from the one you have. If it fights you, say so here and I will look at the conflict with you rather than leaving you to guess at what changed.

Nothing about this reflects on the work itself. Ping me when the rebase is up.

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:
Merged as `e982df55`. Thank you — and the refactor is better than what I asked for, which is worth saying plainly.

I asked for a check that the recorded numerics match the loaded ones. You made the drift impossible instead: `dtype = numerics` means the string stamped onto the verdict **is** the string handed to `make_generator`, so the two cannot diverge by construction. And `utils/live_eval.py` raising on an unsupported quantization rather than falling back means the stamp cannot silently lie either. A check I have to trust is weaker than an invariant I can read.

Verified before merging, both claimed behaviours mutation-checked independently — removing the stamp breaks two tests on `KeyError: 'numerics'`, disabling the mismatch gate turns a stale-evidence refusal into a SHIP. The exit-code taxonomy holds across sixteen evidence files (0 SHIP / 2 DON'T / 3 usage / 1 runtime), with no gate refusal exiting 2 and no DON'T-SHIP becoming 0. CLI and MCP now refuse the same file for the same reason, which was a real divergence before.

Sorry about the changelog-directory churn — that was release timing, not anything you did.

Your CONTRIBUTORS.md entry is in and #367 is closed. All four of its acceptance criteria are now met.

Issue #745, opened 2026-09-07
LoRA+ (loraplus_lr_ratio) is silently ignored by the preference and RL trainers


abdulwaarith0:
Follow-up to #738 (LoRA+ optimizer fix), surfaced by its wiring-coverage scan.

## The gap

`training.loraplus_lr_ratio` is a shared `TrainingConfig` option with no task gating. #738 wired it into the three SFT-family wrappers (`sft`, `pretrain`, `embedding`) via `attach_loraplus_optimizer`. The other **thirteen** PEFT-building trainers accept `loraplus_lr_ratio` and silently ignore it — they build no custom optimizer at all:

`dpo`, `ppo`, `grpo`, `kto`, `orpo`, `simpo`, `bco`, `ipo`, `reward_model`, `classifier`, `distill`, `asr`, `unlearn`

They are recorded in `_LORAPLUS_NOT_IMPLEMENTED` in `tests/test_issue724_loraplus_wiring_coverage.py`, so the gap is explicit rather than lost.

## Behaviour today

Before #738, setting `loraplus_lr_ratio` on a DPO/GRPO/etc. run **crashed** on the unknown `TrainingArguments` keyword (loud). After #738 it **runs to completion with LoRA+ silently disabled** — the B matrices train at the base rate and nothing reports it. So #738 made this half quieter for these trainers too, which is exactly the failure mode the coverage scan exists to make visible.

## The decision (this is the open question)

Each of the thirteen should end up in one of two states:

1. **Wire it** — call `attach_loraplus_optimizer` in each, so LoRA+ works there too. The coverage scan then requires them to move out of `_LORAPLUS_NOT_IMPLEMENTED`.
2. **Refuse it** — reject `loraplus_lr_ratio` at config parse for tasks that do not implement it, so it fails fast with a message naming the task and field instead of silently doing nothing.

Both close the silent-no-op; which is right is a scope call I would rather have decided than guess at.

## Related (fold in or split, your call)

APOLLO is not schema-guarded against LoRA+ the way GaLore is. `attach_loraplus_optimizer` calls `Trainer.get_optimizer_cls_and_kwargs(args)` with `model=None`, and the APOLLO/GaLore family resolves `optim_target_modules` from `model`; GaLore is caught by the runtime guard, APOLLO is not. A schema cross-validator would catch both at parse time. (I could not exercise the APOLLO path — the package is not installed in my environment.)

## Acceptance criteria

[...]

PR #747, opened 2026-09-08
test(peft): prove LoRA+ optimizer and scheduler state survive save/resume (#724)

SID-6921:
## Summary

Closes the last unchecked acceptance criterion on #724: "saving/resuming preserves optimizer and scheduler state" with LoRA+ active. PR #738 fixed the crash and wired the optimizer construction, but left this specific criterion unproven.

## What this is, and what it isn't

**It is not a bug fix.** `attach_loraplus_optimizer` already runs before `trainer.train()` in every wrapper (#738), which is what lets `Trainer.create_optimizer` skip building a default optimizer and build the scheduler around the LoRA+ one instead. I went in expecting to find a gap in that wiring and instead confirmed, empirically, that it already works correctly.

**It is the missing test.** Nothing previously proved that `Trainer._load_optimizer_and_scheduler` actually restores a LoRA+ optimizer's momentum and the scheduler's decayed LR correctly on resume, rather than just not crashing.

## How I verified it works before writing the test

Built a real tiny PEFT model + real `transformers.Trainer`, ran a 4-step training loop twice against the same seeded base model and dataset — once straight through, once interrupted at the real checkpoint saved at step 2 and resumed. With both runs targeting the same `max_steps` (matching how `--resume` is actually used — same config, continuing after an interruption), the resumed run's optimizer momentum, scheduler state, and final LoRA weights all matched the uninterrupted run's exactly (`torch.allclose` at 1e-6, and the momentum tensors matched bit-for-bit via `torch.equal` right at the checkpoint-load boundary, before any new steps ran).

One thing that looked like a bug at first and wasn't: if the resumed run targets a *different* `max_steps` than the original, the two runs diverge — but that's `transformers.Trainer`'s own general resume characteristic (the LR schedule's decay curve depends on the target step count), not anything specific to LoRA+. Worth knowing, not worth fixing here.

## The test

`TestResumePreservesOptimizerAndSchedulerState` in `tests/test_issue724_loraplus_optimizer.py`, following #738's own pattern of a real model + real `Trainer`, no mocks.

## Verification

- 8/8 tests pass in the file (7 existing + 1 new).
- **Mutation**: reordered the resumed run's `attach_loraplus_optimizer` call to *after* `trainer.train()` starts — simulating the exact ordering regression the existing docstring already warns is possible. The test fails with the real PyTorch error that ordering mistake actually produces: `ValueError: loaded state dict has a different number of parameter groups`. Reverted after confirming.
- `ruff check` and `mypy` both clean on the changed file.

A changelog fragment will follow in a quick commit once this PR has a number, per this repo's `changelog.d/<PR#>.<type>.md` convention.

### DIFF
--- a/changelog.d/0.74.0/747.added.md
+++ b/changelog.d/0.74.0/747.added.md
@@ -1,0 +1,10 @@
+- **Proved LoRA+ optimizer and scheduler state survive save/resume (#724 by @SID-6921 in
+  #747).**
+  #738 fixed `training.loraplus_lr_ratio` crashing on the SFT, pretrain and embedding
+  wrappers, but its last acceptance criterion — that resuming a LoRA+ run preserves the
+  optimizer's momentum and the scheduler's decayed learning rate — was never exercised.
+  A new test runs a real 4-step training loop twice against the same seeded base model and
+  dataset, once straight through and once interrupted at a real checkpoint and resumed,
+  both targeting the same `max_steps`; the optimizer momentum, scheduler LR, and final LoRA
+  weights all match exactly between the two. No production code changed: the existing
+  `attach_loraplus_optimizer` wiring already handled this correctly, it just wasn't proven.
--- a/tests/test_issue724_loraplus_optimizer.py
+++ b/tests/test_issue724_loraplus_optimizer.py
@@ -25,10 +25,14 @@ BASE_LR = 2e-5
 RATIO = 16.0
 
 
-def _tiny_peft_model():
+def _tiny_peft_model(seed: int = 0):
+    import torch
     from peft import LoraConfig, get_peft_model
     from transformers import AutoConfig, AutoModelForCausalLM
 
+    # Seeded so two separate calls can build bit-identical base + adapter
+    # weights - needed to compare a resumed run against an uninterrupted one.
+    torch.manual_seed(seed)
     cfg = AutoConfig.for_model(
         "llama", hidden_size=32, intermediate_size=64, num_hidden_layers=2,
         num_attention_heads=4, vocab_size=128,
@@ -37,6 +41,40 @@ def _tiny_peft_model():
     return get_peft_model(
         model, LoraConfig(r=8, lora_alpha=16, target_modules=["q_proj", "v_proj"])
     )
+
+
+def _tiny_dataset(n=16, seq_len=6, vocab_size=128, seed=123):
+    import torch
+
+    g = torch.Generator().manual_seed(seed)
+    return [
+        {
+            "input_ids": torch.randint(0, vocab_size, (seq_len,), generator=g),
+            "attention_mask": torch.ones(seq_len, dtype=torch.long),
+            "labels": torch.randint(0, vocab_size, (seq_len,), generator=g),
+        }
+        for _ in range(n)
+    ]
+
+
+def _trainer_with_data(model, tmp_path, dataset, *, max_steps, save_steps):
+    from transformers import Trainer, TrainingArguments
+
+    args = TrainingArguments(
+        output_dir=str(tmp_path), learning_rate=BASE_LR, optim="adamw_torch",
+        max_steps=max_steps, save_steps=save_steps, save_strategy="steps",
+        per_device_train_batch_size=4, report_to=[], logging_steps=1000,
+        disable_tqdm=True, seed=42, data_seed=42,
+    )
+    return Trainer(model=model, args=args, train_dataset=dataset)
+
+
+def _exp_avg_by_name(trainer):
+    return {
+        name: trainer.optimizer.state[p]["exp_avg"].clone()
+        for name, p in trainer.model.named_parameters()
+        if p in trainer.optimizer.state and "exp_avg" in trainer.optimizer.state[p]
+    }
 
 
 def _trainer(model, tmp_path, *, weight_decay=0.01, optim="adamw_torch"):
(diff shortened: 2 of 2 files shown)


MakazhanAlpamys:
@SID-6921 — **this closes a gap that #738 declared open in its own merge commit, and it discriminates. One change and I merge.**

## The gap is real and self-declared

#724's criterion 3 is *"saving/resuming preserves optimizer and scheduler state"*, and `d7aea47`, the commit that shipped the LoRA+ fix, says so in as many words:

> *"Not covered, and stated rather than implied: criterion 3 of #724 — save/resume of optimizer and scheduler state — has no test here"*

Nothing in the repo touched resume. `grep -rn "loraplus\|lora_plus\|LoRA+" tests/` finds two substantive files: the seven-test LoRA+ file and the AST wiring scan. Neither goes near a checkpoint. You duplicate nothing.

## It discriminates, and four mutations die on it alone

| mutation | result |
|---|---|
| `Trainer._load_optimizer_and_scheduler` made a no-op | **killed only by your test** |
| optimizer restored, then `optimizer.state.clear()` | **killed only by yours** — *"optimizer momentum for ...q_proj.lora_B.default.weight did not survive resume"* |
| scheduler restored with `last_epoch=0` | **killed only by yours** — `[0.0, 0.0, 0.0, 0.0] != [1e-05, 5e-07, 0.0016, 0.00016]` |
| on resume, LoRA+ optimizer replaced by a plain `AdamW` | **killed only by yours** |

Nothing else in the suite catches any of them. And the two halves of your assertion each carry weight independently — the momentum check and the scheduler-LR check die to different mutations, so neither is decorative.

**You assert real values, not object existence.** `torch.allclose(atol=1e-6)` on the Adam `exp_avg` tensor per named parameter, and the full per-group LR list — which that third mutation shows is four groups with four distinct values, i.e. the LoRA+ split rates actually surviving the round trip. The vacuity guards are there too: `assert ref_state and set(ref_state) == set(resumed_state)` cannot pass on an empty dict, and `assert compared_any` cannot pass on an empty loop.

**The honest limit, which your mutation table makes visible rather than hides:** it is a run-versus-run comparison, so a defect damaging both runs equally survives it. Deleting the LoRA+ optimizer attachment entirely, or collapsing the ratio to 1.0, are both caught — by the *pre-existing* tests, not by yours. That is the right division of labour for criterion 3, but it means the fragment's line about the existing wiring already handling this correctly is proved at the resume boundary specifically.

## The one change — it fails at this project's own declared torch floor

`pyproject.toml` declares `torch>=2.5.0`. At torch 2.5.1:

```
FAILED TestResumePreservesOptimizerAndSchedulerState::test_resumed_run_matches_the_uninterrupted_run
E   ValueError: Due to a serious vulnerability issue in `torch.load` ... we now require
    users to upgrade torch to at least v2.6 in order to use the function.
    transformers/trainer.py:3708: in _load_optimizer_and_scheduler

1 failed, 7 passed
```

The same file on `main` is `7 passed`, so this is a **new** failure at the floor, not one of the pre-existing #651 ones. CI is green on all nine cells only because the runners resolve a newer torch.

The cause is not your test — it is correct, and with the version gate neutralised it passes 8/8. Yours appears to be the first test in the suite to call `trainer.train(resume_from_checkpoint=...)` against a real `Trainer` and a real on-disk checkpoint, so it is the first to reach that code path at all.

Please guard it with `pytest.mark.skipif` on torch `< 2.6`. The precedent is at `tests/test_issue654_batch_probe_uncovered_branches.py:131`. Someone installed at the floor this project *declares* should get a skip, not a red — #651 already documents that floor as unusable and this quietly widens the blast radius.

## Housekeeping

Ruff clean. No tests deleted or weakened — the only removed line is `def _tiny_peft_model():` becoming `def _tiny_peft_model(seed: int = 0):`, and all seven pre-existing tests still pass. Fragment named after the PR, crediting you. 14 of 14 checks green with `head_sha` confirmed. Merges clean onto current `main`. Runs in about five seconds.

No overlap with #745 — that issue explicitly carves criterion 3 out and assigns it here, and your PR touches only `tests/`.

Add the skip guard and I merge on green. If you would rather not, say so and I will add it myself and merge with the work credited to you.

SID-6921:
Added in `d17722f`: `@pytest.mark.skipif(_TORCH_VERSION < (2, 6), reason=...)` on the test class, naming the installed version and CVE-2025-32434 in the message.

Verified both directions rather than assuming the guard works: pinned torch to 2.5.1 locally and confirmed the suite reports `7 passed, 1 skipped` with the reason string naming that exact version, instead of the red failure you found. Reinstalled torch 2.8.0 and confirmed all 8 pass there.

Good catch — I'd tested this locally at 2.8.0 the whole time and never actually ran it at the floor.

MakazhanAlpamys:
Merged as `fe004ee9`. The `skipif` is exactly what I asked for, and I confirmed it is not silently swallowing the test on a modern stack — the version comparison is tuple-based, so `(2, 14) >= (2, 6)` rather than the lexicographic trap that would have made this skip forever.

**Not skipped here, and I checked rather than assumed:** `8 collected, 8 passed, 0 skipped` in 34.02s, `-rs` prints no skip lines, and the failure tensors in the mutation runs below carry `device='cuda:0'` — it really ran, on real CUDA, on torch 2.14.0+cu130 / transformers 5.17.

**Three mutations, and the interesting result is which ones this test catches that nothing else does.**

1. Reverted the production fix — `attach_loraplus_optimizer` returns `False`, so no LoRA+ optimizer is built and HF falls back to plain AdamW: **5 failed, 3 passed**. The five kills are all pre-existing tests; your new resume test passed. That is the limit you named yourself in the PR body — a run-versus-run comparison cannot see a defect that damages both runs equally — and I would much rather have that stated than papered over.
2. LoRA+ optimizer built normally, but `Trainer._load_optimizer_and_scheduler` patched to a no-op so saved state is never restored: **1 failed, 7 passed**, and the single failure is yours — `optimizer momentum for ...layers.0.self_attn.q_proj.lora_B.default.weight did not survive resume`.
3. Momentum restored correctly, scheduler alone rewound to step 0 on resume: **1 failed, 7 passed**, again yours alone.

Mutations 2 and 3 are caught by nothing else in 20,871 tests. That is the answer to "is a test-only PR worth merging".

**What makes the test hold up under those mutations, concretely:**

- Asserting the **full per-group** `get_last_lr()` list rather than a single scalar. Four groups with four distinct rates *is* the LoRA+ split, so a scheduler that rewound but kept the shape cannot pass.
- Two real vacuity guards — `assert ref_state and set(ref_state) == set(resumed_state)` cannot pass on an empty dict, and `assert compared_any` cannot pass on an empty loop. This is the class of test that most often ends up asserting nothing, and yours cannot.
- Running the reference arm to the same `max_steps` rather than comparing against a shorter run, which is what makes the final `lora_*` weight equality meaningful instead of tautological.

The single removed line in the whole diff is `_tiny_peft_model()` gaining a `seed` argument; no assertion was loosened, and the added `torch.manual_seed` makes the seven pre-existing tests more deterministic than they were.

**Everything else I ran:** `tests/ -k "loraplus or lora_plus"` 57 passed / 0 skipped, ruff clean, 14/14 CI green on `d17722f1`.

#724 closes with this. #738 declared the resume gap open in its own merge commit and this is what closes it — thank you for picking up a gap that a merge message admitted rather than one an issue advertised.

Issue #748, opened 2026-09-09
Nothing detects a config field that reaches no consumer: 36 candidates today, and the next one ships silently


Srinivasan8888:
## The pattern, not a single bug

A field is declared in `config/schema.py`, validated, documented with an example — and read by nothing. It accepts your value and does nothing with it, with no warning. A search of this tracker for `ignore|silently|never read|does not honour|no caller` returns **36 issues**, including #683, #684, #685, #686 (four in one backend), #689, #700, #720, #745, #623, #561, #563, #584, #634, #659, #676, #682.

Every one of those was found by a person reading code and filed one at a time. Nothing in the suite notices when a field loses its last consumer, or when a new field is added with no wiring. That is the gap this issue is about.

## Measured exposure

`TrainingConfig` declares **237** fields, `DataConfig` **38**. Walking every module under `src/soup_cli/` except `schema.py` with `ast`, and counting a field as *read* when some module contains an attribute access `.<field>` **or** uses the name as a string constant (which covers `getattr(cfg, "x")` and `d["x"]`):

| bucket | count | meaning |
|---|---:|---|
| no attribute read **and** no string reference | **36** | nothing can be reading it |
| referenced only as a bare string | 61 | accessed dynamically — usually fine |

Docstrings are stripped before the walk. That distinction is the whole game: a plain `grep` reports `citation_recall_threshold` as consumed because the name appears in a validator's *error messages*, and reports nothing about whether it is ever applied.

The second bucket is why a naive check would be useless — `relora_steps`, `loraplus_lr_ratio` and friends live there and are genuinely consumed.

## Three from the first bucket, verified by hand

**`training.lr_groups`.** `src/soup_cli/utils/lr_groups.py` is 8.3 KB and exports `parse_lr_groups()`. Outside `schema.py`, **nothing imports it**. `docs/peft-and-efficiency.md:190` shows a worked `lr_groups:` example. So per-layer learning-rate groups validate, the regexes are ReDoS-probed, and no trainer ever builds a param group.

**`data.mask_history`.** Schema: *"mask all but the last assistant turn during loss computation"*. Documented at `docs/data.md:692`. No consumer. This one changes the loss, silently — the same class as #683.

**`training.early_stop_patience`.** Schema: *"Consecutive regressions before early stopping (1-10)"*, `ge=1, le=10`. Documented at `docs/peft-and-efficiency.md:622` as `# Stop after N regressions`. No consumer.

[...]

PR #750, opened 2026-09-09
fix(mlx): honour training.max_grad_norm instead of dropping it silently (#749)

### LINKED ISSUE #749: MLX SFT ignores training.max_grad_norm: sixteen transformers trainers clip, MLX does not, and nothing says so
## The defect

`max_grad_norm: float = Field(default=1.0, gt=0)` (`config/schema.py:1098`) is forwarded into `TrainingArguments` by sixteen transformers trainers, so the documented default clips every transformers run at 1.0. On `backend: mlx` it reaches nothing:

- no file under `src/soup_cli/trainer/mlx_*.py` mentions it;
- `mlx_lm 0.31.3`'s `TrainingArgs` has no such field, and its `trainer.py` contains no `clip` or `norm` call at all, so upstream does not clip either;
- `_check_unsupported()` (`mlx_sft.py:136`) does not list it, so unlike `training.seed` / `training.data_seed` nothing is printed.

The same config therefore trains **clipped on one backend and unclipped on the other**, with nothing in the output to tell them apart. This is the same family as #683 / #684 / #685 / #686, and it was not in any of their lists.

## Measured

On an M1, driving the real `mlx.optimizers` inside upstream's own compiled-step shape (`@partial(mx.compile, inputs=state, outputs=state)`, `trainer.py:246-248`), one step on a deliberately ill-conditioned batch:

```
                       unwrapped    clip=1.0    clip=0.01
SGD    |w - w0| =     388.025909    0.008490     0.000085
AdamW  |w - w0| =       0.126505    0.126504     0.126403
```

**The AdamW row is the honest bound on impact.** Adam normalises by its second moment, so clipping barely moves an AdamW step. What this actually rescues is `optimizer: sgd` / `lion`, and any run whose gradients spike — not every MLX run.

## Why it is fixable rather than only warnable

`mlx.optimizers.clip_grad_norm(grads, max_norm)` exists, and mlx-lm applies gradients through exactly one call — `optimizer.update(model, grad)` at `trainer.py:259`. So an optimizer that clips before delegating reaches it without forking upstream's loop. Everything but `update` has to delegate, because upstream reads `optimizer.state` (`trainer.py:246`) and `optimizer.learning_rate` (`trainer.py:337`) off the object it is handed.

The alternative — adding it to `_check_unsupported()` so the run says `MLX backend ignores: training.max_grad_norm` — is strictly worse when the fix is this small, but it is a legitimate option if you would rather not carry a wrapper.

## Acceptance criteria

1. A run on `backend: mlx` clips gradients at the configured `training.max_grad_norm`, asserted at the optimizer rather than in the plan or in metadata.
2. A different configured value produces a different clip, so the test cannot pass against a hardcoded default.
3. The underlying optimizer receives the **clipped** gradients, not the raw ones.
4. `optimizer.state` and `optimizer.learning_rate` remain reachable, since upstream reads both off the object it is given.
5. The norm that ran is recoverable from the output directory, because a Soup-side wrapper is otherwise invisible there.
6. Tests run on a CI runner with no MLX — no `importorskip` on the wiring assertions.

I have a branch that does this, with the measurements above and a five-mutation table. Happy to open it, or to switch to the warn-only option if you prefer that.

Srinivasan8888:
Closes #749.

## The defect

`max_grad_norm` (`schema.py:1098`, default 1.0, validated `gt=0`) is forwarded into `TrainingArguments` by **sixteen** transformers trainers. On `backend: mlx` it reached nothing: no MLX file read it, `mlx_lm 0.31.3`'s `TrainingArgs` has no such field and its `trainer.py` contains no `clip`/`norm` call at all, and `_check_unsupported()` did not list it. The same config trained clipped on one backend and unclipped on the other, silently.

## Measured, on Metal, through the production class

Not a re-implementation — the probe imports the wrapper the trainer uses and drives it inside upstream's own compiled-step shape (`@partial(mx.compile, inputs=state, outputs=state)`, `trainer.py:246-248`), so the proxy is known to survive tracing rather than assumed to:

```
                       unwrapped    clip=1.0    clip=0.01
SGD    |w - w0| =     388.025909    0.008490     0.000085
AdamW  |w - w0| =       0.126505    0.126504     0.126403
```

**Stated because it bounds the benefit:** AdamW normalises by its second moment, so clipping barely moves an AdamW step. This rescues `optimizer: sgd` / `lion` and gradient spikes, not every MLX run.

## Shape

`mlx.optimizers.clip_grad_norm` exists, and mlx-lm applies gradients through exactly one call — `optimizer.update(model, grad)` at `trainer.py:259` — so clipping is reachable by handing `train()` an optimizer that clips before delegating, with no fork of upstream's loop. Everything but `update` delegates, because upstream reads `optimizer.state` (`trainer.py:246`) and `optimizer.learning_rate` (`trainer.py:337`) off the object it is given; both are asserted. The norm that ran is recorded in `adapter_config.json`, since a Soup-side wrapper is otherwise invisible in the output directory.

## Mutations — 5 applied, 5 killed

Binary-mode, sha256 checked before/after/restore (baseline `e94730aefd10` restored exactly):

| mutation | result |
|---|---|
| revert to a bare optimizer (no wrapper) | 3 failed |
| hardcode the norm to the schema default | 2 failed |
| apply the RAW gradients (clip result discarded) | 1 failed |
| drop the adapter metadata key | 1 failed |
| break delegation (`__getattr__` -> None) | 1 failed |

## Verification

```
tests/test_mlx_max_grad_norm.py   .venv-ci (no mlx)  6 passed, 3 skipped
                                  .venv    (Metal)   9 passed
full suite, .venv-ci:
    main         41 failed, 20545 passed, 114 skipped
    this branch  41 failed, 20551 passed, 117 skipped
    failure sets IDENTICAL (diff empty in both directions); +6 passed and
    +3 skipped are exactly this file's tests. The 41 are pre-existing
    colour-escape assertions in this environment; main is green on CI.
ruff check src/soup_cli/ tests/    All checks passed!
```

## One removed test line, declared rather than left to be found

`test_review_fixes_v07133.py::TestMlxOptimizer` asserted `seen["optimizer"] is sentinel`, and the proxy breaks that identity. The assertion the test exists for — `train()` receives the optimizer that was built, not None — is preserved by asserting the sentinel is reachable through the proxy. No assertion was deleted.

## Scope limits

The three `importorskip` tests pin upstream's contract (that `mlx_lm` still has no clipping of its own, and that `mlx.optimizers.clip_grad_norm` still exists) and run on no CI cell by design; every wiring assertion runs without MLX. The Metal numbers above are mine and are not reproduced by CI.

If you would rather not carry a wrapper, the alternative is one line in `_check_unsupported()` so the run says `MLX backend ignores: training.max_grad_norm`. Say so and I will swap it.


Srinivasan8888:
**Rebased onto `main` after #734 and #739 merged — `816cc861`. The conflict was semantic, not textual, and resolving it "keep mine" would have deleted #686's optimizer.**

`git merge origin/main` conflicted in `mlx_sft.py` in two places, because #734 replaced the exact line this PR wrapped:

```
main after #734 :  optimizer = build_optimizer(optimizer_plan)
this PR wrapped :  optimizer = _clipping_optimizer(optim.AdamW(learning_rate=...), ...)
```

Taking either side alone loses something. Resolved by keeping **both**: the clipping wrapper now wraps what #686 builds, so the schedule and the clipping compose rather than compete.

```python
optimizer = _clipping_optimizer(
    build_optimizer(optimizer_plan), float(cfg.training.max_grad_norm)
)
```

Applied outside `build_optimizer` deliberately, so `OptimizerPlan` stays a pure description of the schedule and the two fixes stay separable — if you ever want clipping in the plan instead, it is one move rather than an untangling. The metadata conflict resolved the same way: `**optimizer_plan.as_metadata()` **and** `"max_grad_norm"`, not one or the other.

As with #739, I rebased by resetting to `origin/main` and cherry-picking, rather than resolving hunks by hand, and then checked that the merged work is actually still there rather than assuming:

```
git diff origin/main...HEAD --stat     4 files, +330 -2   (only mine)
#734 markers in mlx_sft.py             4   (build_optimizer / as_metadata present)
#739 markers in mlx_sft.py             6   (on_val_loss_report / _last_loss present)
M8 guard in test_val_loss_end_to_end   1   (intact)
```

## One test of mine was wrong after the merge, and the merge was right

`test_the_underlying_optimizer_is_still_reachable_through_the_wrapper` asserted `optimizer.learning_rate == 3e-4`. After #686 the optimizer is handed a **callable** schedule for any real scheduler, so that assertion failed — my assumption was stale, the delegation was fine.

Fixed in two parts rather than by loosening it: the value check now runs under `scheduler: constant` with no warmup, where #686's early return genuinely yields a float; and a new test pins the case the merge exposed — a cosine schedule with warmup must arrive as a callable and must still be the *same object* the inner optimizer holds, since mlx-lm reads `.learning_rate` off whatever it is given (`trainer.py:337`).

## Mutations re-run on the rebased code, not carried over

The implementation moved, so the old table would have been a claim about code that no longer exists. Baseline `249fbea8d34c`, restored exactly:

| mutation | result |
|---|---|
| unwrap — revert to #686's optimizer, no clipping | 3 failed |
| hardcode the norm to 1.0 | 2 failed |
| apply the RAW gradients (clip result discarded) | 1 failed |
| drop the adapter metadata key | 1 failed |
| break delegation (`__getattr__` -> None) | 2 failed |

```
245 passed, 1 failed, 20 skipped   across 10 MLX/val-loss/affected suites (.venv-ci)
 62 passed                          686 + this file on Metal (.venv)
ruff check src/soup_cli/ tests/     All checks passed!
```

The one failure is `TestAsrSkipControlStrip::test_hostile_filename_is_stripped`, which fails identically on `main` in this environment — a colour-escape assertion, not this PR.

MakazhanAlpamys:
@Srinivasan8888 — merged as `c7c6314a`. **Three mutations re-run at merge on an interpreter with no mlx, all dead.**

```
baseline                                        7 passed, 3 skipped

clipping reverted (norm effectively infinite)   2 failed
  -> test_the_optimizer_train_receives_clips_at_the_configured_norm
  -> test_a_different_norm_is_a_different_call

norm hardcoded to the schema default 1.0        2 failed   (same two)

raw gradients applied, clip result discarded    1 failed
  -> test_the_real_optimizer_is_handed_the_clipped_gradients
```

The third is the one I care about most: it leaves a correct-looking `clip_grad_norm` call in place and changes only what reaches the optimizer. A test asserting "the clipper was called" would pass on it. Yours asserts the clipper's *return value* is what the inner optimizer receives, so it does not.

The ordering mutation — clip after the update rather than before — dies too, and that is the other one that still calls the clipper with the right argument.

## What this merge does and does not prove

**No CI test asserts a clipped gradient magnitude**, and I want that on the record rather than implied by a green tick. The tests use a recording fake; the arithmetic evidence — SGD 388.03 to 0.0085 — is your M1 measurement and stays yours. This box has no Metal.

What does cover the real path is `mlx-smoke`: it drives the actual `soup train` CLI on macos-14 against real `mlx` and `mlx_lm`, through `mx.compile`, and asserts the saved adapter loads with real LoRA parameters. It passed. So the residual risk I would otherwise carry — a proxy object not surviving tracing, or an `isinstance` check upstream rejecting it — is empirically closed even though magnitudes are not.

## Two small things

Your PR body and the fragment say "sixteen transformers trainers". I count **fifteen** files, one forwarding site each: `asr`, `bco`, `classifier`, `distill`, `dpo`, `embedding`, `grpo`, `ipo`, `kto`, `online_dpo`, `orpo`, `pretrain`, `reward_model`, `sft`, `simpo`. Not worth a commit on its own, but the number is in a shipped changelog fragment, so if you touch that file again it is the line.

The widened assertion in `test_review_fixes_v07133.py` — accepting either the optimizer or its wrapped inner — is fine and you declared it. It does make that assertion permanently unable to tell wrapped from unwrapped, but it was never a clipping test and delegation is pinned separately by its own mutation.

## The pattern

This is the third of your PRs this week whose subject is a setting that was validated, documented, and read by nothing. `training.max_grad_norm` had no entry in `_check_unsupported` either, so it did not even warn — it just quietly differed from every other backend. That is exactly the class #748 is about, and it is why the draft in #751 is worth finishing.

Ruff clean, 14 of 14 SUCCESS on `816cc861`, `head_sha` confirmed equal to the head I merged.

PR #751, opened 2026-09-09
test(config): fail the suite when a declared field reaches no consumer (#748)

Srinivasan8888:
Draft while the scope questions settle. Refs #748.

## What it does

Walks every module under `src/soup_cli/` except `schema.py` with `ast`, and fails for any declared `TrainingConfig` / `DataConfig` field that no module **reads**, unless it is in an explicit `KNOWN_UNCONSUMED` allowlist with an accountable reason.

Reads are Load-context attributes, Load subscripts, and the *name* argument of `getattr`/`get`/`pop`/`setdefault`/`hasattr`. Writes are not reads: `d["field"] = v` and `{"field": v}` are code emitting config, not consuming a setting. That distinction is the whole mechanism — it is what found #759, which sixteen grep hits could not.

## What it found

**#759**, filed at the maintainer's request: `data.remove_unused_columns` is user-settable and **fifteen** trainers hardcode `False`. Nothing reads the field. Found within a day of the guard existing, by distinguishing a write from a read.

## Limits, stated in the module docstring

- **The name space is global** — it asks whether an identifier appears anywhere under `src/`, not whether it appears on a config object (~3,400 names). A field named `temperature` reads as consumed via `request.temperature`; `top_p`, `dtype`, `timeout`, `seed` behave the same. It leaks hardest on the generic names a new HF/TRL passthrough field would carry.
- **A field read only through a resolver defined in `schema.py` reads as an orphan**, because `schema.py` is excluded. Worked example: `bnb_4bit_use_double_quant`, consumed via the `double_quant_on` `@property` (#321).
- **It cannot see a value read and then rewritten** (#423 — `detect_device()` not recognising MLX, so `quantization: 4bit` was correctly read and silently rewritten).

Three shapes, and only the first is this PR's:

| shape | example | caught by |
|---|---|---|
| read by nothing | `remove_unused_columns` (#759) | this guard |
| read, but not on your backend | `max_grad_norm` (#749) | a declared registry (#756) |
| read, then rewritten | `quantization` on MLX (#423) | neither |

## The allowlist: 40 entries

Every entry cites an issue or says `no issue yet` / `no issue needed` in words. `staging` is deliberately **not** accepted as accountable — 31 entries used it, and that loose predicate is what let a factually wrong entry through review.

Two entries say having no consumer is *correct*: `packing_cross_doc_attn_mask` (refused at config load) and `bnb_4bit_use_double_quant` (read via a schema-side property the scan cannot see).

## The ratchet is exact

`== 40`, so the list fails when it goes **down** as well as up. `test_the_allowlist_does_not_cover_fields_that_are_consumed` names *which* entry rotted; the pin only says the count moved. The pin earns its place on exactly one case the other test cannot see: an entry added for a brand-new unwired field, where nothing is stale and the count is the only signal.

## Verification

```
23 passed in 3.6s   (four uncached AST walks -> one cached walk, was 13s)
ruff clean, no src/ change, 0 removed test lines
pytest.mark.unit, no network, no GPU, no MLX
```

Mutations, baseline `7df9186ce750` restored exactly:

| mutation | result |
|---|---|
| collect bare string constants (the original hole) | 7 failed |
| Store subscripts count as reads | 1 failed |
| attribute Store counts as a read | 1 failed |
| drop `getattr`/`get` handling | 4 failed |
| allowlist swallows everything | 1 failed |
| `.get` default counted as a read | 2 failed |
| pin back to `<=` with an entry retired | 2 failed |
| accept `staging` as accountable again | 1 failed |

**One is not killed, and is stated rather than papered over:** removing the scan/import tree assertion survives on a correctly configured machine, because `SRC` and the imported package are the same tree there. It is a safety assertion whose purpose is to fire in an environment this one is not.

## Open

Whether `lr_groups`, `mask_history`, `early_stop_patience` and `citation_recall_threshold` should each become issues, as #759 did. The maintainer has said yes; I have not filed them yet.

### DIFF
--- a/tests/test_issue748_config_fields_reach_a_consumer.py
+++ b/tests/test_issue748_config_fields_reach_a_consumer.py
@@ -1,0 +1,319 @@
+"""Issue #748 — a config field that reaches no consumer must fail the suite.
+
+A field is declared in `config/schema.py`, validated, documented with a worked
+example, and read by nothing. It accepts a value and does nothing with it,
+silently. Searching this tracker for `ignore|silently|never read|does not
+honour|no caller` returns 36 issues, among them #683, #684, #685 and #686 --
+four in a single backend. Every one was found by a person reading code.
+
+Nothing failed when a field lost its last consumer, and nothing fails today
+when a field is added with no wiring. This guard closes that.
+
+**What it does and does not claim.** It answers "does anything read this",
+not "does this backend read this". `training.max_grad_norm` is read by sixteen
+transformers trainers and by nothing on MLX; that is a strictly harder problem
+and out of scope here.
+
+**Why an AST walk and not a grep.** `citation_recall_threshold` appears in a
+validator's error-message strings, so `grep -rl` calls it consumed while
+nothing applies it. Docstrings are stripped before the walk for the same
+reason. Conversely a field reached only through `getattr(cfg, "name")` or
+`cfg_dict["name"]` IS consumed, and a guard that flagged those would be
+deleted within a week -- `relora_steps` and `loraplus_lr_ratio` are exactly
+that shape. Both directions are pinned below.
+"""
+
+from __future__ import annotations
+
+import ast
+import pathlib
+
+SRC = pathlib.Path(__file__).resolve().parents[1] / "src" / "soup_cli"
+SCHEMA = "schema.py"
+
+
+# --------------------------------------------------------------------------
+# The detector. Kept here rather than in `src/` because it is test-only
+# tooling; nothing in the shipped CLI should depend on it.
+# --------------------------------------------------------------------------
+def _strip_docstrings(tree: ast.AST) -> ast.AST:
+    """Remove docstrings so prose naming a field cannot count as reading it."""
+    for node in ast.walk(tree):
+        if isinstance(
+            node, (ast.Module, ast.ClassDef, ast.FunctionDef, ast.AsyncFunctionDef)
+        ):
+            body = getattr(node, "body", None)
+            if (
+                body
+                and isinstance(body[0], ast.Expr)
+                and isinstance(body[0].value, ast.Constant)
+                and isinstance(body[0].value.value, str)
+            ):
+                node.body = body[1:] or [ast.Pass()]
+    return tree
+
+
+def consumed_names(paths) -> set:
+    """Names some module reads: attribute accesses plus string constants.
+
+    String constants are included deliberately -- `getattr(cfg, "field")` and
+    `cfg_dict["field"]` are real consumption. Docstrings are excluded, so
+    prose is not.
+    """
+    names: set = set()
+    for path in paths:
+        try:
+            tree = _strip_docstrings(ast.parse(path.read_text(errors="ignore")))
+        except (SyntaxError, UnicodeDecodeError):  # pragma: no cover - defensive
+            continue
+        for node in ast.walk(tree):
+            if isinstance(node, ast.Attribute):
+                names.add(node.attr)
+            elif isinstance(node, ast.Constant) and isinstance(node.value, str):
+                if len(node.value) < 80:
+                    names.add(node.value)
+    return names
+
+
+def _consumer_modules():
+    return [p for p in SRC.rglob("*.py") if p.name != SCHEMA]
+
+
+# --------------------------------------------------------------------------
+# Fields with no consumer today. Each entry is a promise that someone looked.
+#
+# Seeded so this lands green; the number may only shrink. Removing an entry
+# because the field was wired is the point. Adding one requires a reason.
+# --------------------------------------------------------------------------
+KNOWN_UNCONSUMED = {
+    # -- documented with a worked example, applied nowhere. Verified by hand.
+    "training.lr_groups": "utils/lr_groups.py exports parse_lr_groups() and "
+                          "nothing outside schema.py imports it; documented at "
+                          "docs/peft-and-efficiency.md:190",
+    "data.mask_history": "schema promises 'mask all but the last assistant turn "
+                         "during loss computation'; documented at docs/data.md:692",
+    "training.early_stop_patience": "schema promises 'consecutive regressions "
+                                    "before early stopping'; documented at "
+                                    "docs/peft-and-efficiency.md:622",
+    "training.citation_recall_threshold": "validated by utils/citation_faithful.py "
+                                          "and named in its error strings; never applied",
+    # -- declared and deliberately REFUSED, so having no consumer is correct.
+    #    A distinct category from the two below: the user is told, loudly, at
+    #    config load. Found by this guard rather than by hand.
+    "training.packing_cross_doc_attn_mask": "rejected at config load "
+                                            "(schema.py:3495) because it never "
+                                            "mapped to a valid TRL packing_strategy; "
+                                            "documented at docs/performance-and-"
+                                            "quantization.md:153",
+    # -- staged for features that have not landed; grouped so they can be
+    #    retired together rather than one at a time.
+    "training.long_context_grpo": "documented as wiring Tiled MLP; no Tiled MLP exists",
+    "training.vision_grpo": "no vision GRPO path",
+    "training.load_in_16bit": "schema rewrites quantization at validation time",
+    "training.unsloth_bnb_4bit": "unsloth quantisation staging",
+    "training.llm_int8": "bitsandbytes int8 staging",
+    "training.quantize_ref_model": "reference-model quantisation staging",
+    "training.yarn_attn_factor": "YaRN staging",
+    "training.yarn_beta_fast": "YaRN staging",
+    "training.yarn_beta_slow": "YaRN staging",
+    "training.convergence_window": "convergence-detector staging",
+    "training.convergence_rel_tol": "convergence-detector staging",
+    "training.forgetting_eval_steps": "catastrophic-forgetting probe staging",
+    "training.forgetting_benchmark": "catastrophic-forgetting probe staging",
+    "training.forgetting_stop": "catastrophic-forgetting probe staging",
+    "training.checkpoint_eval_steps": "checkpoint-eval staging",
+    "training.checkpoint_eval_metric": "checkpoint-eval staging",
+    "training.checkpoint_eval_tasks": "checkpoint-eval staging",
+    "training.checkpoint_keep_top": "checkpoint-eval staging",
+    "training.grace_codebook_size": "GRACE codebook staging",
+    "training.grace_codebook_dim": "GRACE codebook staging",
+    "data.video_dir": "video pipeline staging",
+    "data.eval_on_each_dataset": "per-dataset eval staging",
+    "data.split_thinking": "thinking-block masking staging",
+    "data.image_min_pixels": "image preprocessing staging",
+    "data.image_max_pixels": "image preprocessing staging",
+    "data.image_resize_algorithm": "image preprocessing staging",
+    "data.video_fps": "video pipeline staging",
+    "data.video_maxlen": "video pipeline staging",
+    "data.resize_vocab": "vocab-resize staging",
+    "data.extend_conversation": "conversation-extension staging",
+    "data.skip_prepare_dataset": "dataset-prep bypass staging",
+}
+
+
+def _declared():
+    from soup_cli.config.schema import DataConfig, TrainingConfig
+
+    out = {}
+    for cls, label in ((TrainingConfig, "training"), (DataConfig, "data")):
+        for name in cls.model_fields:
+            out[f"{label}.{name}"] = name
+    return out
+
+
+class TestTheDetectorItself:
+    """The guard is only worth having if the detector is right in BOTH
+    directions. A false negative lets a dead field through; a false positive
+    gets the test deleted.
+    """
+
+    def _consumed(self, tmp_path, source: str) -> set:
+        module = tmp_path / "consumer.py"
+        module.write_text(source)
+        return consumed_names([module])
+
+    def test_an_attribute_access_counts_as_consumption(self, tmp_path):
+        assert "widget" in self._consumed(tmp_path, "def f(cfg):\n    return cfg.widget\n")
+
+    def test_getattr_with_a_string_counts_as_consumption(self, tmp_path):
+        """`getattr(tcfg, "fp8_recipe", ...)` is how utils/v028_features.py
+        reads many real fields."""
+        src = 'def f(cfg):\n    return getattr(cfg, "widget", None)\n'
+        assert "widget" in self._consumed(tmp_path, src)
+
+    def test_a_dict_lookup_counts_as_consumption(self, tmp_path):
+        assert "widget" in self._consumed(tmp_path, 'def f(d):\n    return d["widget"]\n')
+
+    # The docstring fixtures below use the bare field name as the ENTIRE
+    # docstring. An earlier version wrote prose around it ("Talks about
+    # widget.") and was vacuous: the collected constant is then that whole
+    # sentence, never the bare name, so the assertion held whether or not
+    # stripping happened. Found by mutating the stripper -- disabling it
+    # survived all three. An exact-match docstring is also the realistic
+    # shape, since generated documentation often is exactly the field name.
+
+    def test_a_module_docstring_does_not_count(self, tmp_path):
+        """The distinction a grep gets wrong."""
+        assert "widget" not in self._consumed(tmp_path, '"""widget"""\n')
+
+    def test_a_function_docstring_does_not_count(self, tmp_path):
+        assert "widget" not in self._consumed(
+            tmp_path, 'def f():\n    """widget"""\n    return 1\n'
+        )
+
+    def test_a_class_docstring_does_not_count(self, tmp_path):
+        assert "widget" not in self._consumed(
+            tmp_path, 'class C:\n    """widget"""\n    x = 1\n'
+        )
+
+    def test_prose_mentioning_a_field_is_not_a_read_either(self, tmp_path):
+        """The `citation_recall_threshold` shape: named inside a longer
+        message. Collected as the whole sentence, so it never matches the
+        field name -- pinned so a future change to how constants are split
+        cannot start counting prose as consumption."""
+        src = 'def f():\n    raise ValueError("widget must be in [0, 1]")\n'
+        assert "widget" not in self._consumed(tmp_path, src)
+
+    def test_an_unrelated_module_consumes_nothing(self, tmp_path):
+        """Reject-everything control: the detector must not report a name that
+        is simply absent, or every field would look consumed."""
+        assert "widget" not in self._consumed(tmp_path, "x = 1\n")
+
+    def test_a_syntactically_broken_module_is_skipped_not_fatal(self, tmp_path):
+        """One unparseable file must not take the guard down."""
+        bad = tmp_path / "bad.py"
+        bad.write_text("def (:\n")
+        assert consumed_names([bad]) == set()
+
+
+class TestEveryDeclaredFieldReachesAConsumer:
+    def test_no_new_field_is_declared_without_a_consumer(self):
+        consumed = consumed_names(_consumer_modules())
+        orphans = sorted(
+            key
+            for key, attr in _declared().items()
+            if attr not in consumed and key not in KNOWN_UNCONSUMED
+        )
+        assert not orphans, (
+            "These config fields are declared in schema.py and read by no "
+            "module outside it, so a user setting them gets no effect and no "
+            "warning:\n  "
+            + "\n  ".join(orphans)
+            + "\n\nWire the field, or add it to KNOWN_UNCONSUMED with a reason."
+        )
+
+    def test_the_allowlist_names_only_real_fields(self):
+        """A renamed or deleted field must not keep a stale entry alive --
+        otherwise the allowlist silently stops guarding anything."""
+        declared = _declared()
+        stale = sorted(k for k in KNOWN_UNCONSUMED if k not in declared)
+        assert not stale, (
+            "KNOWN_UNCONSUMED names fields that no longer exist; remove them:\n  "
+            + "\n  ".join(stale)
+        )
+
+    def test_the_allowlist_does_not_cover_fields_that_are_consumed(self):
+        """The list may only shrink. When a field gets wired, its entry has to
+        go, or the guard stops noticing if the wiring is later removed."""
+        consumed = consumed_names(_consumer_modules())
+        declared = _declared()
+        now_wired = sorted(
+            k for k in KNOWN_UNCONSUMED
+            if k in declared and declared[k] in consumed
+        )
+        assert not now_wired, (
+            "These fields now have a consumer, so their KNOWN_UNCONSUMED entry "
+            "is obsolete and must be deleted:\n  " + "\n  ".join(now_wired)
+        )
+
+    def test_every_allowlist_entry_carries_a_reason(self):
+        empty = sorted(k for k, v in KNOWN_UNCONSUMED.items() if not v or len(v) < 10)
+        assert not empty, f"allowlist entries need a reason: {empty}"
+
+    def test_the_guard_can_actually_fail(self, tmp_path, monkeypatch):
+        """Acceptance criterion 1, demonstrated rather than described.
+
+        A guard that has never been observed failing is not yet known to work.
+        This adds a field to the real TrainingConfig, confirms the check goes
+        red naming it, then wires a consumer and confirms it goes green.
+        """
+        from soup_cli.config.schema import TrainingConfig
+
+        fields = dict(TrainingConfig.model_fields)
+        fields["totally_unwired_probe"] = fields["max_grad_norm"]
+        monkeypatch.setattr(TrainingConfig, "model_fields", fields)
+
+        consumed = consumed_names(_consumer_modules())
+        orphans = [
+            key for key, attr in _declared().items()
+            if attr not in consumed and key not in KNOWN_UNCONSUMED
+        ]
+        assert orphans == ["training.totally_unwired_probe"], (
+            f"the guard did not flag an unwired field; it reported {orphans}"
+        )
+
+        # ...and green once something reads it.
+        wired = tmp_path / "wired.py"
+        wired.write_text("def f(cfg):\n    return cfg.totally_unwired_probe\n")
+        consumed_after = consumed_names(_consumer_modules() + [wired])
+        assert "totally_unwired_probe" in consumed_after
+        assert not [
+            key for key, attr in _declared().items()
+            if attr not in consumed_after and key not in KNOWN_UNCONSUMED
+        ]
+
+    def test_removing_a_fields_last_consumer_is_caught(self, tmp_path):
+        """Acceptance criterion 2: the failure fires on the commit that breaks
+        it, not months later in a user's run."""
+        declared = _declared()
+        # `lr` is read all over the trainers; simulate its last consumer going.
+        assert "training.lr" in declared
+        only_docstring = tmp_path / "gone.py"
+        only_docstring.write_text('"""This module used to apply cfg.lr."""\n')
+        consumed = consumed_names([only_docstring])
+        assert "lr" not in consumed, (
+            "a field named only in a docstring must read as unconsumed"
+        )
+
+
+def test_the_allowlist_is_not_silently_growing():
+    """A ratchet, in the shape of the recipe-count test.
+
+    If this number goes up, a field was added without wiring and allowlisted
+    instead. That is sometimes right -- but it should be a deliberate edit to
+    this line, visible in review, rather than a quiet append.
+    """
+    assert len(KNOWN_UNCONSUMED) <= 36, (
+        f"KNOWN_UNCONSUMED has grown to {len(KNOWN_UNCONSUMED)}; a field was "
+        "allowlisted rather than wired. Lower this bound when entries are retired."
+    )
--- a/tests/test_issue748_config_fields_reach_a_consumer.py
+++ b/tests/test_issue748_config_fields_reach_a_consumer.py
@@ -36,42 +36,56 @@ SCHEMA = "schema.py"
 # The detector. Kept here rather than in `src/` because it is test-only
 # tooling; nothing in the shipped CLI should depend on it.
 # --------------------------------------------------------------------------
-def _strip_docstrings(tree: ast.AST) -> ast.AST:
-    """Remove docstrings so prose naming a field cannot count as reading it."""
-    for node in ast.walk(tree):
-        if isinstance(
-            node, (ast.Module, ast.ClassDef, ast.FunctionDef, ast.AsyncFunctionDef)
-        ):
-            body = getattr(node, "body", None)
-            if (
-                body
-                and isinstance(body[0], ast.Expr)
-                and isinstance(body[0].value, ast.Constant)
-                and isinstance(body[0].value.value, str)
-            ):
-                node.body = body[1:] or [ast.Pass()]
-    return tree
-
-
 def consumed_names(paths) -> set:
-    """Names some module reads: attribute accesses plus string constants.
+    """Names some module READS. Reads only -- writes are not consumption.
 
-    String constants are included deliberately -- `getattr(cfg, "field")` and
-    `cfg_dict["field"]` are real consumption. Docstrings are excluded, so
-    prose is not.
+    Counted as a read:
+      * `obj.field` in a load context;
+      * `d["field"]` in a load context;
+      * `getattr(obj, "field")` / `cfg.get("field")` and friends.
+
+    Deliberately NOT counted:
+      * `d["field"] = value` and `{"field": value}` -- that is code EMITTING
+        config, not reading the user's setting. This is the hole that let
+        both of the maintainer's named offenders through: `data.interleave`
+        looked consumed because `mix_proxy.py` writes
+        `data_block["interleave"] = {...}`, and
+        `bnb_4bit_use_double_quant` because `save_formats.py` writes it as a
+        key in an output dict. Run against the tree at the commit where each
+        was a live defect, the earlier version reported both as CONSUMED.
+      * docstring prose, and any other bare string constant. Nothing here
+        collects a free-standing `ast.Constant`, so prose is excluded
+        structurally rather than by a stripping pass. An earlier version
+        stripped docstrings explicitly; mutation testing showed that pass was
+        dead once reads were narrowed to Load contexts and call arguments, so
+        it was removed rather than left looking load-bearing.
     """
     names: set = set()
     for path in paths:
         try:
-            tree = _strip_docstrings(ast.parse(path.read_text(errors="ignore")))
-        except (SyntaxError, UnicodeDecodeError):  # pragma: no cover - defensive
+            tree = ast.parse(path.read_text(errors="ignore"))
+        except (SyntaxError, UnicodeDecodeError, ValueError):
             continue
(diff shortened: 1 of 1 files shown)


MakazhanAlpamys:
@Srinivasan8888 — this is a draft so I am not reviewing it, but I want to say now that the shape is right, because it is worth knowing before you finish rather than after.

A repo-wide test that fails when a declared field reaches no consumer is the correct control for that failure class, and this project has direct evidence that the alternative does not work. Four config flags shipped in v0.74.0 that were validated, documented and read by nothing — `bnb_4bit_use_double_quant`, the MLX device/quantization path, `--no-reexec`'s dropped flags, and `data.interleave`, the last of which I filed as #443 rather than fixed. In every case the prose rule existed and the flag shipped anyway. The Rich/ANSI CI failure has the same history: written into `CLAUDE.md`, `CONTRIBUTING.md` and a memory note, and it still turned CI red four times until `tests/test_cli_help_assertions_are_ansi_safe.py` made it fail.

Two things that decided whether that ANSI scanner was any good, offered because they are the traps here too:

- **Its first version passed its own invented example while missing the real commit.** It looked for the marker on the assert line; the actual offender assigned the CLI output to a variable first. Whatever you build should be run against the **known** offenders — the 36 candidates you counted — and each one it fails to catch is a hole.
- **Its first normaliser check flagged 31 correct assertions across 18 files.** A guard that fires on correct code gets deleted by the next person in a hurry, so the false-positive rate matters more than the catch rate. Expect a large allowlist here and make the allowlist itself hard to grow silently.

And the load-bearing test in that file is `TestTheScannerCanActuallyFail` — since it finds zero offenders today, nothing else proves it still works.

No deadline, and take it out of draft whenever you like.

Srinivasan8888:
**Your first trap caught it. I ran the guard against the offenders you named, and it missed one — `32ba168f` fixes the detector.**

## The hole

I checked out the tree at the commit where each was a live defect and ran the detector there, rather than reasoning about it:

```
at 44d90c19 (parent of the #443 fix):
  interleave                  CONSUMED   <- would NOT have failed the build
  bnb_4bit_use_double_quant   CONSUMED   <- would NOT have failed the build
```

Both for the same reason: **writes were counted as reads.**

```python
save_formats.py:267   {"bnb_4bit_use_double_quant": double_quant}   # emitting config
mix_proxy.py          data_block["interleave"] = {...}              # emitting config
```

That is code *producing* configuration, not reading the user's setting, and it made the field look consumed. Reads are now Load contexts, Load subscripts, and `getattr`/`get`/`pop`/`setdefault`/`hasattr` arguments. At that same historical commit the fixed detector reports `bnb_4bit_use_double_quant` as an **ORPHAN** — the guard would have failed the build on the commit that shipped it.

**`data.interleave` still reads as consumed there, and that is correct.** `data_mix.py:130` genuinely did read `data_block.get("interleave", {})` — #443 was a *per-consumer* gap, not "read by nothing". That is the scope limit I wrote into #748, and it is the half a declared per-backend registry answers instead of this.

## A second defect, found by mutating rather than reading

The docstring stripper mutated `node.body` during an `ast.walk`, which is unreliable — `data_mix.py`'s class docstring survived it and its "Attributes:" prose counted as a read. And once reads were narrowed to Load contexts, nothing collects a free-standing `Constant` at all, so the whole pass was **dead code that looked load-bearing**. Disabling it failed nothing; that mutation is what told me. Removed, and the docstring exclusion is now structural.

## Four more real fields, verified by hand not allowlisted blind

The read/write fix flagged four the old version missed. Two are worth your eye:

- **`data.remove_unused_columns`** — schema default `True`, documented *"set False when feeding extra cols to a custom collator"*, and `sft.py:788`, `pretrain.py:174`, `embedding.py:175`, `grpo.py:468` each **hardcode `False`**. The setting is not merely unread, it is overridden.
- **`training.bnb_4bit_use_double_quant`** — your own example. `quant_menu.py:401` writes that key from `tcfg.double_quant_on`, a *different* field.

Allowlist 36 -> 40 with the ratchet raised to match, each entry carrying its reason.

## On your second trap — false positives

Taken seriously, and it is why I hand-checked all four rather than appending them. The current false-positive count against the real tree is **0**: every one of the 40 has a stated reason and `test_the_allowlist_does_not_cover_fields_that_are_consumed` fails if any of them gains a consumer, so the list cannot rot in the quiet direction either.

Six new tests pin the read/write distinction in **both** directions — a dict *read* and a `.get()` must still count, or the guard would cry wolf on ordinary code — including your example reduced to the shape it ships in.

Mutations, baseline `67cf88efb5ce` restored exactly:

| mutation | result |
|---|---|
| collect bare string constants (the original hole) | 7 failed |
| Store subscripts count as reads | 1 failed |
| attribute Store counts as a read | 1 failed |
| drop `getattr`/`get` handling | 4 failed |
| allowlist swallows everything | 1 failed |

22 passed in 4.2s, ruff clean, no `src/` change.

Still a draft, and the three questions in the PR body are still open — in particular whether `remove_unused_columns` and `bnb_4bit_use_double_quant` should be issues rather than allowlist entries, now that they have gone from "counted" to "named".

Srinivasan8888:
Both traps run, on `32ba168f`. Numbers rather than reassurance.

## Trap 1 — against the offenders that actually shipped

You said each one it fails to catch is a hole, so here is each of the four, including the one it does not flag.

**`bnb_4bit_use_double_quant` — caught.** It is flagged, and the read/write distinction is exactly why: its only mentions outside `schema.py` are `save_formats.py:267` and `quant_menu.py:401`, both keys inside a `Dict` literal being *written* into an output config. Counting those as reads is what made it look consumed.

**`data.interleave` — not flagged, and that is correct, not a hole.** It was wired by #468 (merged 2026-08-25) and #443 is closed. Verified four real reads rather than trusting the issue state:

```
commands/data_mix.py:130   .get("interleave", {})
commands/data.py:2328      attribute LOAD
data/loader.py:570         attribute LOAD
data/loader.py:1107        attribute LOAD
```

Worth noting the near-miss: `mix_proxy.py:194` is `data_block["interleave"] = {...}`, a subscript **store**, and the guard correctly does not count it. So on this field the guard would have been wrong in *both* directions before the read/write fix — it would have called it consumed on the strength of a write, and it now calls it consumed on the strength of four genuine reads.

**`--no-reexec`'s dropped flags (#372) — out of scope, and I am not claiming it.** That is a CLI option, not a `TrainingConfig`/`DataConfig` field, so this detector cannot see it by construction. It is the layer #752 covers.

**The MLX device/quantization path — I could not map your description to a specific field name**, so I am not going to claim a catch I did not verify. The quantization-adjacent fields in the flagged set are `training.load_in_16bit`, `training.llm_int8`, `training.unsloth_bnb_4bit` and `training.quantize_ref_model`, all four of which *are* flagged. If you tell me which field you meant I will check it specifically.

## Trap 2 — the false-positive rate

You said this matters more than the catch rate. Measured across all 275 declared fields:

```
declared fields                        275
flagged with the allowlist ignored      40
KNOWN_UNCONSUMED entries                40
flagged but NOT allowlisted              0
allowlisted but NOT flagged (stale)      0
```

Then every mention of each flagged field, anywhere under `src/soup_cli/` outside `schema.py`, classified by AST:

| bucket | count |
|---|---:|
| no textual mention at all | 23 |
| mentions, **all** non-reads | 17 |
| **mentions that are an attribute load** | **0** |

The 17 break down as:

| field | what the mentions actually are |
|---|---|
| `data.remove_unused_columns` | 16 files, every one `remove_unused_columns=False` as a **keyword argument being set** |
| `training.bnb_4bit_use_double_quant` | dict-literal keys being written |
| `training.yarn_factor`, `yarn_attn_factor`, `yarn_beta_fast`, `yarn_beta_slow` | parameters of `get_rope_scaling_config()`; `long_context.py:316` is `_finite_positive(yarn_factor, "yarn_factor")`, the string being the name for an error message |
| `training.grace_codebook` | `registry/store.py:52` and `commands/edit.py:312` — the registry **artifact kind** `"grace_codebook"`, a name collision with the config field |
| `training.lr_groups`, `unsloth_bnb_4bit`, `long_context_grpo`, `vision_grpo`, `packing_cross_doc_attn_mask`, `citation_recall_threshold`, `grace_codebook_dim/_size`, `data.video_fps`, `video_maxlen` | prose only — docstrings, comments, or validator error messages |

**Zero false positives out of 40.** The `grace_codebook` collision is the one I would keep an eye on: it is the shape where a field and an unrelated artifact kind share a name, and a detector that matched loosely would have called it consumed.

## One thing the sweep turned up that is not about the guard

`data.remove_unused_columns` is user-settable and **sixteen trainers hardcode `remove_unused_columns=False`** when building `TrainingArguments`. Setting it in a config does nothing, in either direction. That is the #749 shape rather than a staged-feature shape, and I think it wants its own issue rather than an allowlist line — say the word and I will file it with the sixteen call sites.

## Status

Suite is green on the branch (22 passed), the ratchet is in at `<= 40`, and the allowlist and the detector agree exactly with no slack in either direction. Still a draft, and I will leave it that way until you have said whether the allowlist should carry issue numbers.

MakazhanAlpamys:
@Srinivasan8888 — **you are right about `data.interleave` and I was wrong. That goes first.**

I told you #443 was still open. It is not:

```
#443  state=CLOSED  closed=2026-08-20T15:57:08Z
#468  MERGED 2026-08-25  feat(data): extend data.interleave to streaming and hub datasets
```

So not flagging it is correct, exactly as you said, and I checked your four reads rather than taking them. I repeated that claim from `CLAUDE.md`, which still describes #443 as filed-rather-than-fixed in two places — **the stale record is mine and I will fix it there too.** You verified the code instead of trusting the issue state, which is the thing I asked for and then failed to do myself.

**And your near-miss note is the sharpest observation in the reply.** On this one field the guard would have been wrong in *both* directions before your read/write fix: it would have called it consumed on the strength of `mix_proxy.py:194`, a subscript store, and it now calls it consumed on the strength of four genuine attribute loads. A detector that gets the same verdict for opposite reasons is not working; yours now gets the right verdict for the right reason.

## Trap 1 — you found the hole yourself, which is the point

The guard missed one of the four and you fixed the detector rather than the expectation. That is the correct direction and it is rarer than it should be.

On the two you did not claim: declining to assert a catch you could not verify is worth more to me than a fourth tick. The MLX one I meant is **#423** — `detect_device()` did not know MLX, so every Apple Silicon run reported "CPU (no GPU detected)" and silently rewrote `quantization: 4bit` to `none`. That is a *rewrite* rather than a dropped read, so I suspect your detector cannot see it by construction either; if so, say that and it stops being an open question.

`--no-reexec` being out of scope is right, and #752 covers that layer — which is now merged as `1490c850`.

## Trap 2 — zero false positives out of forty

That is the number that decides whether this survives contact with the next person in a hurry, and 0/40 with the allowlist and the detector agreeing exactly in both directions is a better result than I expected. The `grace_codebook` collision is the one I would also watch: a config field and an unrelated registry artifact kind sharing a name is precisely the shape a looser matcher gets wrong.

## Your two questions

**File the `remove_unused_columns` issue — yes, please.** Sixteen trainers hardcoding `remove_unused_columns=False` while the field is user-settable is the #749 shape exactly: not a staged feature, just a setting that does nothing in either direction and does not warn. It deserves its own issue with the sixteen call sites, not an allowlist line, and finding it is a real result from the sweep rather than a side effect.

**The allowlist should carry issue numbers.** An entry without one is indistinguishable from an entry someone added to make CI green, and that is how a ratchet rots. If a field has no issue, the entry is the prompt to file one — which is what just happened with `remove_unused_columns`.

One thing I would add: make the ratchet's number **fail when it goes down as well as up**, or state in a comment why it does not. If someone wires a field and forgets the allowlist line, a `<= 40` bound stays green while the allowlist is now stale by one — the same asymmetry that made #756's registry go stale five times today.

Take it out of draft whenever you like.

Srinivasan8888:
**Both structural asks done in `081a3194`, and `remove_unused_columns` is filed as #759 — with a correction to my own count in it.**

## The ratchet fails in both directions now

`== 40` rather than `<= 40`. Your asymmetry point is right and I had not seen it: the bound only caught the list growing, and said nothing about it going stale. Mutating it back to `<=` while retiring an entry — the exact scenario you described — now fails 2 tests. `test_the_allowlist_does_not_cover_fields_that_are_consumed` is the other half: it names *which* entry rotted, where the count only says something moved.

## Issue references, and the check found six on its first run

Every entry now cites an issue or says `no issue yet` out loud. Two say **`no issue needed`** instead, because for them having no consumer is correct rather than pending: `packing_cross_doc_attn_mask` is refused at config load, and `load_in_16bit` is rewritten by the schema at validation time. Collapsing those into "no issue yet" would have made them read as unfiled work forever.

Stripping any one entry's marker fails the check; so does stripping the `#759` reference specifically.

## #759 filed — and I got the count wrong first

You said sixteen trainers. I said four. **Both wrong: it is fifteen.** My four came from grepping only the dict-literal form and missing twelve keyword-argument sites. Then my corrected grep said sixteen, because `raft.py:104` is a **docstring** explaining why the flag matters:

```
raft.py:104:  _raft_row_index}`` rows (``remove_unused_columns=False`` keeps them). Each
```

Counted by AST, excluding docstrings: **15 sites across 15 files**. Corrected on the issue with the full list rather than edited quietly.

That is the third time in two days a text search has given me the wrong answer about this one question, in both directions — "consumed" for `bnb_4bit_use_double_quant` when the name appeared only in writes, and "sixteen" here when one was prose. It is the argument for the detector, including when I am the one holding the grep.

## #423 — you are right, and it is out of scope by construction

`training.quantization` **is** read (`quant_menu.py:360`, `autodistill/mlx_worker.py:292`), so the guard reports it consumed and always will. #423 was not a dropped read: `detect_device()` did not recognise MLX, so a correctly-read value was **rewritten** downstream. Detecting that needs to know what a field's value *should* be on a given device, which is a different question from whether anything reads it. So: it cannot see #423, by construction, and that stops being an open question.

That makes three shapes, worth stating together since they keep coming up:

| shape | example | caught by |
|---|---|---|
| read by nothing | `bnb_4bit_use_double_quant` | this guard |
| read, but not on your backend/task | `max_grad_norm` (#749), `data.interleave` (#443) | a declared registry (#756) |
| read correctly, then rewritten | `quantization` on MLX (#423) | neither — needs device-aware expectations |

17 passed, ruff clean, no `src/` change. Still a draft; the remaining open question is only whether `bnb_4bit_use_double_quant` should get an issue too, now that `remove_unused_columns` has one.

MakazhanAlpamys:
@Srinivasan8888 — **I measured the ratchet rather than read it. It detects a real offender, it fails in both directions, and the two tests divide the labour correctly. One entry is factually wrong and that is the blocker; everything else below is a finding, not a condition.**

Tested as **this PR merged into current `main` (`ccd906c`)**, not against your base, in an isolated worktree with `PYTHONPATH` verified to point at that worktree. Every mutation in binary mode with an occurrence-count assert and sha256 before and after; every file restored to its exact baseline hash.

## It is not decorative — both directions proven

Added a plausible unread field to the real `TrainingConfig` (`sha 3814cbbe… -> 5e829a2a…`, anchor occurrences 1):

```
E  AssertionError: These config fields are declared in schema.py and read by no module
E  outside it, so a user setting them gets no effect and no warning:
E      training.gradient_clip_warmup_ratio
FAILED ...::test_no_new_field_is_declared_without_a_consumer
```

Then the other direction — appended a genuine read of an allowlisted field to `utils/convergence.py` (`sha 0e3fb129… -> 1e4a5df2…`):

```
E  These fields now have a consumer, so their KNOWN_UNCONSUMED entry is obsolete:
E      training.early_stop_patience
```

and deleting that entry while keeping the wiring fires the pin with the right instruction:

```
E  KNOWN_UNCONSUMED is 39, pinned at 40. Going UP means a field was allowlisted rather
E  than wired; going DOWN means an entry was retired, which is the good direction --
E  lower this number in the same commit.
```

**Your `== 40` answer is right and I was wrong to frame it as symmetry for its own sake.** The division is what makes it work: `test_the_allowlist_does_not_cover_fields_that_are_consumed` names *which* entry rotted, and the pin only says the count moved. The pin earns its place on exactly one case the other test cannot see — an entry added for a *brand-new* unwired field, where nothing is stale and the count is the only signal. Please put that sentence in the docstring, or someone deletes the pin as redundant in a year.

## BLOCKING: `training.bnb_4bit_use_double_quant` is consumed, and the entry cites the proof as the disproof

The entry says `quant_menu.py:401` writes the key from `tcfg.double_quant_on`, "a **different field**", and that this one is "never read". Verified on `ccd906c`:

```
$ sed -n '1880,1886p' src/soup_cli/config/schema.py
    @property
    def double_quant_on(self) -> bool:
        """Resolve the tri-state ``bnb_4bit_use_double_quant`` ... (#321)"""
        return self.bnb_4bit_use_double_quant is not False
```

`double_quant_on` is not a different field. It is a `@property` **over this one**. And the value reaches bitsandbytes on three paths — `quant_menu.py:401`, `stream_setup.py:507`, and `save_formats.py:267` via `layer_stream_runtime.py:1548`. End to end:

```
config bnb_4bit_use_double_quant=None  -> BitsAndBytesConfig.bnb_4bit_use_double_quant=True
config bnb_4bit_use_double_quant=True  -> BitsAndBytesConfig.bnb_4bit_use_double_quant=True
config bnb_4bit_use_double_quant=False -> BitsAndBytesConfig.bnb_4bit_use_double_quant=False
```

The line the entry cites as evidence of non-consumption has this comment directly above it:

```python
# #321 — honour the configured flag instead of hardcoding True.
"bnb_4bit_use_double_quant": tcfg.double_quant_on,
```

It was the **v0.73.3** example of the class, not v0.74.0, and **#321 was the fix, not the defect**. So the entry freezes a *repaired* field as an open offender — which is the worst thing an allowlist can do, because nothing will ever remove it: `test_the_allowlist_does_not_cover_fields_that_are_consumed` cannot see it either, for the reason in the next section.

**Delete the entry, lower the pin to 39.**

## And the reason it slipped past you is a general one worth documenting

`schema.py` is excluded from the scan, so **a field read only through a resolver defined in `schema.py` reads as an orphan**. There is exactly one `@property` in the whole schema today and the allowlist got that one wrong. Tri-state resolver properties are the documented pattern here — #321 established it — so this recurs the next time someone adds one. Put `double_quant_on` in the docstring as the worked example; it is the case that caught you, which makes it the one worth writing down.

## The mechanism is better than grep, and it leaks on exactly the names that matter

The read/write narrowing in `32ba168f` genuinely works — I probed nine shapes and it counts loads (`o.widget`, `b["widget"]`, `b.get("k","widget")`) and excludes stores, dict literals, keyword arguments, imports and defs. Docstrings are excluded *structurally* rather than by stripping, which is the stronger version.

The hole is that the name space is **global**: it asks "does this identifier appear as an attribute or key anywhere in `src/`", not "on a config object". The consumed set is ~3410 names — essentially the codebase's entire attribute namespace. Demonstrated with a field named after a common attribute:

```python
temperature: Optional[float] = Field(default=None, gt=0.0, le=2.0, ...)   # zero readers
```
```
17 passed in 11.17s
```

Silently green, because `request.temperature` (`commands/serve.py:1867`) and `row.temperature` (`autodistill/publisher.py:156`) exist for unrelated reasons. Probing plausible future names — `top_p`, `max_new_tokens`, `dtype`, `timeout`, `warmup_ratio`, `seed`, `logging_steps` — all read as consumed with no config reader. The guard leaks hardest on precisely the generic names a new HF/TRL passthrough field would carry.

I am **not** asking you to fix that; scoping attribute reads to config-typed objects is a much larger test. I am asking for one paragraph in the module docstring next to your existing "what this does not claim", because a guard whose limits are undocumented gets trusted past them.

## MEDIUM: the scanned tree and the imported schema can be different trees, and it fails green

`SRC` comes from `__file__` — correct, it scans the checkout the test lives in. But `_declared()` does `from soup_cli.config.schema import ...`, which resolves to the **installed** package, and there is no `pythonpath` in `[tool.pytest.ini_options]` nor any `sys.path` work in `conftest.py`. With the unread field present in a worktree:

```
=== no PYTHONPATH (editable install -> main checkout) ===   17 passed in 13.03s
=== with PYTHONPATH (worktree) ===                          2 failed, 15 passed
```

CI is unaffected, so this is a contributor-experience bug rather than a CI one — but the failure mode is **silent green on a ratchet**, which is the one this file exists to prevent. Two lines:

```python
import soup_cli
assert pathlib.Path(soup_cli.__file__).resolve().parent == SRC, (
    f"scanning {SRC} but importing {soup_cli.__file__} - set PYTHONPATH or reinstall -e ."
)
```

## Smaller items

- **Runtime.** One `consumed_names(_consumer_modules())` is 499 modules / 6.5 MB / ~1.9 s, measured three times. The file does **four** uncached walks: 5.95 s + 3.82 s (twice) + 2.50 s. An `lru_cache` or a session fixture takes 13 s to ~2 s. Across nine CI cells that is about two minutes per push for a pure-Python AST walk.
- **`test_the_guard_can_actually_fail` misreports.** With my probe field present it failed saying *"the guard did not flag an unwired field; it reported [...]"* — when the guard had flagged it correctly. `assert orphans == [...]` should be `assert "training.totally_unwired_probe" in orphans`.
- **`.get("k", "widget")` counts the default as a read** (verified). Restrict to `node.args[1]` for `getattr`/`hasattr` and `node.args[0]` for `get`/`pop`/`setdefault`.
- **PR body is stale** — it says 36 entries and asks about `<= 36`; `081a3194` has 40 and `== 40`. Update it before un-drafting so nobody audits the wrong number.
- No `@pytest.mark.unit` marker, unlike the rest of the suite. Harmless.

## Your three questions

1. **Issue numbers vs prose.** The shape is right and `test_every_allowlist_entry_states_an_issue_or_says_there_is_none` is the good half. But it accepts the bare token `staging` as a pass and 31 of 40 use it — and that predicate is what let the `bnb_4bit_use_double_quant` entry through with a wrong story attached. Require an issue reference or an explicit "no issue yet", and drop `staging` as a pass.
2. **Should the never-applied fields become issues.** Yes for `lr_groups`, `mask_history`, `early_stop_patience` and `citation_recall_threshold` — all four are documented user-visible promises. #759 is the model for how.
3. **Is the bound worth having.** Yes, for the one reason in the first section. Document it.

## One thing that is mine, not yours

The five stale entries I raised are on **#756**, whose registry is per-backend. This guard asks "does *anything* read this", so #734 and #750 wiring the MLX path changed nothing here — `test_the_allowlist_does_not_cover_fields_that_are_consumed` passes clean against current `main`, which is the correct answer. Nothing on this PR went stale from those merges, and I do not want you chasing that across two PRs.

**Fix the `bnb_4bit_use_double_quant` entry, drop the pin to 39, add the docstring paragraph on the global-name and `schema.py`-resolver limits, and the import/scan assertion — then take it out of draft and I merge on green.** The runtime cache and the four smaller items are worth doing in the same push but I will not hold the PR on them.

Worth saying plainly: this guard found #759 within a day of existing, and it found it by distinguishing a *write* of a value from a *read* of a setting — which sixteen grep hits could not. That is the argument for the feature, and it was made by a defect rather than by an example.

Srinivasan8888:
**All of it done in `f3505371`, except that your blocking remedy turns the guard red — so I kept the entry and corrected its story instead. That needs your ruling.**

## The blocking item: you are right about the fact, and I think wrong about the fix

Verified before changing anything:

```
schema.py:1880  @property
                def double_quant_on(self) -> bool:
                    return self.bnb_4bit_use_double_quant is not False

config bnb_4bit_use_double_quant=False -> double_quant_on=False
```

So `double_quant_on` is a property **over** this field, #321 was the fix rather than the defect, and my entry froze a repaired field as an open offender. That was the worst kind of wrong and I should have read the property before writing the reason.

**But deleting the entry fails the guard.** The field genuinely has no consumer *outside* `schema.py`, which the scan excludes, so the orphan check flags it:

```
FAILED test_no_new_field_is_declared_without_a_consumer
    training.bnb_4bit_use_double_quant
```

That is a limit of the detector, not a defect in the field. So the entry stays, at pin 40, with the true reason and an explicit note that **deleting it fixes nothing** — only a detector that understands schema-side resolvers could retire it. If you would rather the scan include `schema.py` property bodies, say so; I did not do it because validators there read fields too, and including them would mark many genuinely-unread fields as consumed — a false negative in the direction that matters.

## Everything else

- **Limits paragraph** in the module docstring: the global name space (~3,400 names, with `temperature` / `top_p` / `dtype` / `seed` named), the `schema.py`-resolver blind spot with `double_quant_on` as the worked example, and #423's read-then-rewritten shape.
- **Scan/import tree assertion** — `_declared()` now asserts `soup_cli.__file__` resolves inside `SRC`, naming `PYTHONPATH`. Your measurement was the convincing part: 17 passed without it, 2 failed with it, same tree, and silent-green on a ratchet is the failure this file exists to prevent.
- **Runtime**: four uncached walks -> one cached walk. **13s -> 3.6s.**
- **`test_the_guard_can_actually_fail`** now asserts membership, not equality, with a comment saying why: another orphan being present is a separate finding, not a failure to detect.
- **`.get("k", "widget")`** no longer counts the default. Restricted by argument position, with tests both directions.
- **`staging` dropped as accountable.** All 40 entries now cite an issue or say `no issue yet` / `no issue needed`.
- **The pin's docstring** now states the one case only it can see — an entry added for a brand-new unwired field — so it does not get deleted as redundant in a year.
- `pytest.mark.unit` added.
- **PR body rewritten** — it said 36 and `<= 36`; it now matches the code.

## One mutation survived, and I am not going to pretend otherwise

Accepting `staging` again survived my first attempt, because **loosening a predicate makes a suite pass, not fail**. Extracting `_reason_is_accountable` and asserting what it *rejects* kills it: 1 failed. That is the general lesson — a check can only be pinned by testing its refusals.

And one I could not kill: removing the scan/import assertion survives on a correctly configured machine, because `SRC` and the imported package are the same tree here. It is a safety assertion whose whole purpose is to fire in an environment this one is not. Recorded as unkilled rather than dressed up.

## Your questions 2 and 3

Filing `lr_groups`, `mask_history`, `early_stop_patience` and `citation_recall_threshold` as issues next, with #759 as the model — unless you would rather have one issue covering all four, since they share a shape and a fix decision.

The bound stays, documented as you asked.

Still draft, pending your call on the `bnb_4bit_use_double_quant` entry.

MakazhanAlpamys:
**Ruling: you are right that deleting the entry turns the guard red, and I was wrong to ask for it as a standalone change. But keeping the entry is the wrong resolution, and the third option you offered — teach the scan about schema-side resolvers — is both correct and much smaller than you feared. I measured it.**

## Why "keep the entry with a true reason" is not good enough

`KNOWN_UNCONSUMED` is not a list of things the detector cannot see. It is a list of fields **a user can set with no effect**, and every entry is a standing claim that setting that field does nothing. `bnb_4bit_use_double_quant` reaches bitsandbytes on three paths. An entry saying "this is consumed, but the detector cannot tell, so it stays here" makes the artifact self-contradicting, and the next person to read the list for the reason it exists — deciding what to wire — gets a false answer from the file whose job is to give the true one.

You would also be pinning the detector's blind spot into the count. `== 40` then means "39 unconsumed fields plus one detector limitation", and the pin stops measuring the thing it was built to measure.

## Your objection to scanning `schema.py` does not survive measurement — because it is about validators, not properties

You wrote that including `schema.py` would mark many genuinely-unread fields as consumed, since validators read fields there too. That is exactly right **about validators**. It is not true of properties, and the two are structurally distinguishable. I walked the file:

```
PROPERTIES: [('double_quant_on', 1881, ['property'])]
validator count: 156
other decorated: []
names contributed by property bodies: ['bnb_4bit_use_double_quant']
```

**One property in the entire schema. Zero decorated functions that are neither a property nor a validator.** Scanning property bodies contributes exactly one name today, and it is the one field this whole thread is about. The false-positive blast radius you were protecting against is, measured, a single field that genuinely is consumed.

The distinction is principled rather than convenient: a validator *checks* a value and consumes nothing on the user's behalf; a `@property` *resolves* a value for someone else to consume. #321 established the tri-state resolver as the pattern here, so this recurs, and it will recur one property at a time.

## One refinement, because the naive version has a laundering hole

Counting every property body unconditionally means a property **nobody calls** would launder its fields into "consumed" — the detector's own failure mode, reintroduced one level up. Gate it:

> A read inside a `schema.py` `@property` body counts only if the **property's own name** is in the consumed set.

`double_quant_on` qualifies (`quant_menu.py:401`, `stream_setup.py:507`), so the entry retires. A property that nothing calls contributes nothing, which is the correct answer. That is a handful of lines and it is directly testable in both directions — add an uncalled property reading an unread field and the field must still be flagged.

## So, concretely

1. Add the property-body pass, gated on the property name being consumed.
2. **Delete the `bnb_4bit_use_double_quant` entry and drop the pin to 39.** Verify the guard is green with it gone, and that your fake-unread-field probe still fails.
3. Two tests: a `schema.py` property whose reads make its field count, and an **uncalled** property whose reads do not.
4. Keep your limits paragraph, and replace the `double_quant_on` worked example with the *fixed* version — the general rule now being that a resolver only launders what something actually calls.

## The rest of `f3505371`

Everything else you did is what I asked for and several of them are better than what I asked for. The 13 s to 3.6 s cache, the membership assertion with the reason written down, the argument-position fix for `.get(k, default)`, dropping `staging` as accountable, and the pin's docstring stating the one case only it can see — all correct.

**Two things I want to single out, because they are the reason I am ruling against you on the entry without any hesitation about the PR.**

You reported a surviving mutation that you found yourself and did not have to mention: loosening `staging` back in *passes*, because relaxing a predicate makes a suite green rather than red, and only asserting what `_reason_is_accountable` **rejects** kills it. "A check can only be pinned by testing its refusals" is the correct general statement and it belongs in the docstring next to the others.

You also recorded an **unkilled** mutation rather than dressing it up — removing the scan/import assertion survives here because `SRC` and the imported package are the same tree on a correctly configured machine. That is honest and it is the right call: an assertion whose purpose is to fire in an environment you are not in cannot be killed from inside it. Note it as unkillable-by-construction so nobody later deletes it as untested.

## Your questions 2 and 3

**One issue covering all four.** `lr_groups`, `mask_history`, `early_stop_patience` and `citation_recall_threshold` share a shape — documented user-visible promise, no consumer — and more importantly they share a *decision*: wire, reject at config load, or remove. Four issues would split one decision four ways and invite three different answers. #759 is the right model for depth; make this one cover the set, with a per-field section and a single stated question.

The bound stays, documented as you asked.

Do those four items and take it out of draft.

Srinivasan8888:
**All four done in `5b51bc48`, #761 filed, out of draft. Your ruling was right and the argument I was missing is the one you led with.**

`KNOWN_UNCONSUMED` is not a list of things the detector cannot see — it is a list of fields a user can set with no effect. An entry saying "this IS consumed, but the scan cannot tell" contradicts the file's own purpose and pins a blind spot into the count, so `== N` stops measuring what it was built to measure. I was defending the detector's convenience over the artifact's meaning.

## 1. The property pass, gated

Your measurement reproduced exactly:

```
PROPERTIES:      [('double_quant_on', 1881, ['property'])]
validator count: 156
other decorated: []
```

My objection was about **validators** and does not apply to **properties** — and the two are structurally distinguishable, as you said. One property in the whole schema; the pass contributes exactly one name.

The gate is the part that makes it safe, and it is yours: a read inside a `schema.py` `@property` counts **only if the property's own name is in the consumed set**. `double_quant_on` qualifies via `quant_menu.py:401` and `stream_setup.py:507`.

## 2. Entry deleted, pin 40 -> 39

Verified green with it gone, and the fake-unread-field probe still fails correctly.

## 3. Two tests — and the first version of one was vacuous

`test_an_uncalled_resolver_launders_nothing` built a temp schema and then asserted against `_consumed_in_src()`, **which reads the real schema** — so it held whether the gate existed or not. Dropping the gate survived the mutation. The gate is now extracted as `fold_property_reads` and driven directly:

```
drop the gate (uncalled resolvers launder)          1 failed
count validators as resolvers too                   2 failed
drop the property pass (regress to the blind spot)  3 failed
pin back to <= with an entry retired                2 failed
```

That is the second time in this PR a test of mine asserted something it could not observe. Both times the mutation found it, which is the argument for running them rather than reasoning about them.

## 4. Docstring

The resolver limitation is replaced by the fixed behaviour and its gate. I also wrote down the two rules you singled out, in your framing:

> A check can only be pinned by testing its REFUSALS: loosening a predicate makes a suite pass rather than fail.

> Some assertions are unkillable by construction — the scan/import check exists to fire in a misconfigured environment, so no mutation run inside a correct one can kill it. It is not untested, it is untestable from here; do not delete it for lack of a red.

## #761 — one issue, four fields

`lr_groups`, `mask_history`, `early_stop_patience`, `citation_recall_threshold`, with a per-field section and a single stated question, as you asked. I put a leaning in it rather than sitting on the fence: refuse `lr_groups` and `citation_recall_threshold`, **wire** `mask_history` — it is loss-affecting and the transformers-side machinery already exists in `data/loss_mask.py`, so refusing it seems worse than honouring it. `early_stop_patience` I left open, because it depends on whether `early_stop_on_regression` is itself wired and I have not checked.

```
29 passed in 2.2s   ruff clean   no src/ change   0 removed test lines
```

Out of draft.

MakazhanAlpamys:
Merged as `0e04c2c9`. Thank you — and thank you specifically for the argument in your last comment, which changed my mind. I asked you to keep an allowlist entry and you were right that keeping it was the wrong resolution.

**I did not take the guard on trust; I reproduced it against the real schema rather than the in-file fixtures.** I added a genuinely unwired field to `src/soup_cli/config/schema.py`:

```python
soup_review_orphan_probe: float = Field(default=0.5, gt=0)
```

and `test_no_new_field_is_declared_without_a_consumer` went red naming it — `training.soup_review_orphan_probe ... Wire the field, or add it to KNOWN_UNCONSUMED with a reason`. Adding a module that reads `cfg.training.soup_review_orphan_probe` took it back to 29 passed. Both directions of the acceptance criterion, on the real schema, independent of anything in your test file.

**What makes this worth merging rather than a grep in a trench coat:**

- Reads are collected *structurally* — Load-context attribute access, Load-context string subscripts, and the name argument of `getattr`/`get`/`pop`/`setdefault`/`hasattr`. Writes and bare `ast.Constant` are never collected, so a field that only appears in an error message or a docstring does not launder itself into looking wired.
- `fold_property_reads` folds in `schema.py` `@property` reads only when the property is itself called. That is the distinction a simpler implementation would have missed.
- The `== 39` pin fails in both directions, and `_reason_is_accountable` refuses a hand-wave — every entry cites an issue or says plainly it has none. Rejecting `staging` as an accountable reason is the detail I would have argued for if you had not already done it.

**Other verification I ran:** 29 passed in 4.04s on your head, 29 passed in 3.52s after merging current `main`, 14/14 CI green on `5b51bc48`, ruff clean, and zero removed lines anywhere in the diff. I also ran your guard with #756's `config/backend_support.py` copied into the tree — still 29 passed, because that file stores field names as bare string constants and your walker structurally never collects those. The two cannot launder each other, which was the thing I most wanted to be sure of before merging both.

**One measurement worth recording in the module docstring, since it makes your own caveat concrete.** The name space is global, and I found where it leaks: an unwired `training.verbose` is caught, an unwired `training.top_k` is **not**, because `top_k` appears as an attribute elsewhere in the tree. Your docstring already predicts exactly this ("leaks hardest on exactly the generic names a new HuggingFace/TRL passthrough field would carry") — I would just add the measured example, so the next reader sees it is a ratchet with a known hole rather than a proof.

Small housekeeping: the PR body says 40 allowlist entries and 23 tests, while the file pins `== 39` and collects 29. No action needed now that it is merged, but worth knowing the description drifted from the diff.

Thanks also for filing #761 out of what this turned up rather than leaving the four dead settings implicit in an allowlist. That is the follow-through that makes a guard useful instead of merely green.

PR #753, opened 2026-09-09
test(doctor): pin the accelerate DEPS row to the declared pyproject floor

AmirF194:
Adds the guard gap you flagged in your comment on #352. Reverting `doctor.py`'s
accelerate row to 0.25.0 still passes the full suite, because
`test_doctor_training_bounds_match_declared_extra` only checks `_RUNTIME_FLOORS`
(transformers, trl, peft) against the `DEPS` table, and accelerate isn't in that
set. So `soup doctor` could certify an environment that reintroduces the 37 GB
checkpoint write with nothing catching it.

Went with your suggested shape: a separate assertion next to
`test_pyproject_accelerate_floor_supports_adapter_only_fsdp_checkpoints` in
`test_issue352_fsdp_peft_checkpoint.py`, not widening `_RUNTIME_FLOORS` (that
also drives `_REQUIRES_TRAINING_FLOOR`'s skip behavior for the transformers 5
tests, which this shouldn't touch).

No source change, only the missing coverage. Confirmed the new test fails when
`doctor.py`'s accelerate floor is reverted to 0.25.0 and the two existing
accelerate tests both still pass regardless, which is the exact blind spot.

Not closing #352, just this one guard gap; the resume-from-checkpoint criterion
stays open as you noted.

### DIFF
--- a/tests/test_issue352_fsdp_peft_checkpoint.py
+++ b/tests/test_issue352_fsdp_peft_checkpoint.py
@@ -29,6 +29,21 @@ def test_pyproject_accelerate_floor_supports_adapter_only_fsdp_checkpoints():
     )
 
 
+def test_doctor_deps_accelerate_floor_matches_declared_extra():
+    from soup_cli.commands.doctor import DEPS
+
+    pyproject = Path(__file__).parent.parent / "pyproject.toml"
+    match = re.search(r'"accelerate>=([0-9.]+)"', pyproject.read_text(encoding="utf-8"))
+    assert match, "accelerate floor pin not found in pyproject.toml"
+
+    doctor_floors = {package: floor for _, package, floor, _ in DEPS}
+    assert doctor_floors["accelerate"] == match.group(1), (
+        f"soup doctor reports accelerate>={doctor_floors['accelerate']} but pyproject.toml "
+        f"declares accelerate>={match.group(1)}; doctor would certify an environment that "
+        "reintroduces #352's full-base-model FSDP checkpoint"
+    )
+
+
 def test_installed_accelerate_save_fsdp_model_supports_adapter_only():
     from accelerate.utils import save_fsdp_model
 


MakazhanAlpamys:
@AmirF194 — merged as `99b2a4bf`. **The guard discriminates in both directions and, more importantly, it does not fire on a correct future state. I checked all three cases myself.**

```
doctor DEPS row lowered to 0.25.0, pyproject unchanged     FAILS
pyproject floor raised to 0.28.0, doctor unchanged         FAILS
BOTH raised to 0.28.0 together, a legitimate migration     PASSES
```

The third is the one that decides whether a guard like this is worth keeping, and it is where this repo has gone wrong before: #496 shipped a check that went red on exactly the migration its exception existed to allow. Pinning **agreement** rather than a constant is what avoids that, and you got there without being told.

**The blind spot is real, not asserted.** I reverted `doctor.py` to `0.25.0` with your test absent and the existing suite passes — `test_pyproject_accelerate_floor_supports_adapter_only_fsdp_checkpoints` never reads `doctor.py`, and `_RUNTIME_FLOORS` in `test_issue502_qwen35_transformers5.py` covers transformers, trl and peft but not accelerate. So nothing caught it, and your PR body's claim about that is accurate.

Not widening `_RUNTIME_FLOORS` was also the right call rather than the lazy one: it drives `_REQUIRES_TRAINING_FLOOR` skip behaviour, so adding accelerate there would have changed which tests skip as a side effect of adding a floor check.

Fifteen lines, test-only, no changelog fragment — correct, since no user-visible behaviour changes. Ruff clean. 14 of 14 SUCCESS on `b911071c`, `head_sha` confirmed equal to the head I merged.

Two things worth knowing before the next floor bump, neither of them a change request: the comparison is exact string equality, so `"0.27"` would not match `"0.27.0"`; and the regex takes the **first** accelerate pin in `pyproject.toml`, so a second pin in another extra with a different floor would go unnoticed. Both are fine today — there is exactly one pin — and both are the kind of thing that only bites during a migration.

Separately: your inline finding on #730 was independently correct and I said so in that thread. The remedy you suggested turned out to reintroduce a false positive, but so did the one I had already endorsed, and mine was worse — it changed nothing at all. Two independent reproductions of that boundary are what told me it was not an artifact of one person's fixture.

AmirF194:
Good catch on the exact-string-equality and first-match-only limits, both are real and I'd rather they're written down than discovered during a migration. Thanks for running the three cases yourself and for the pointer on #730, glad the second repro held up.

PR #754, opened 2026-09-09
fix(cli): pass every typer parameter when calling a command as a function (#752)

### LINKED ISSUE #752: soup eval auto dies with TypeError after a clean eval: a typer command called as a function gets OptionInfo, not its defaults
## The bug

`soup eval auto` runs the custom eval to completion, prints the results table, saves the
results to the tracker — and *then* exits with an uncaught `TypeError`.

Reproduced against the real `custom()`, stubbing only the model load and the tracker write
so the output block runs for real:

```
Loaded 1 eval tasks from /tmp/.../tasks.jsonl
Loading model...
  Evaluating... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100%
           Custom Eval Results
┏━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━┓
┃ Category ┃ Correct ┃ Total ┃ Accuracy ┃
┡━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━┩
│ default  │       1 │     1 │  100.00% │
└──────────┴─────────┴───────┴──────────┘
Results saved to experiment tracker.
RESULT: UNCAUGHT TypeError: expected str, bytes or os.PathLike object, not OptionInfo
```

Control — the identical call with the two parameters passed explicitly:

```
RESULT: returned normally  <- the bug is the unfilled params, nothing else
```

## Why

`auto()` calls `custom()` as a plain Python function (`commands/eval.py:698`), passing
three of its five typer parameters:

```python
custom(tasks=tasks_file, model=str(output_dir), run_id=None)
```

`custom()` declares five (`commands/eval.py:306`), the other two being
`attach_to_registry` (`:319`) and `output` (`:323`). Typer fills a command's parameters
only when *typer* invokes it. Called directly, the unfilled parameters keep their literal
defaults — which are `typer.models.OptionInfo` objects, not `None`:

```
custom() parameter defaults as seen by a direct Python call:
  tasks                default=OptionInfo   truthy=True
  model                default=OptionInfo   truthy=True
  run_id               default=OptionInfo   truthy=True
  attach_to_registry   default=OptionInfo   truthy=True
  output               default=OptionInfo   truthy=True
```

**`OptionInfo` is truthy.** So inside `custom()`:

- `:404` — `if output or attach_to_registry:` fires on every call from `auto()`, though the
  user asked for neither;
- `:415` — `write_target = output or "eval_results.json"` evaluates to the `OptionInfo`
  rather than the string, so the fallback filename is unreachable;
- `:417` — the real `write_eval_json(OptionInfo)` raises
  `TypeError: expected str, bytes or os.PathLike object, not OptionInfo`;
- `:418` catches only `(ValueError, FileNotFoundError)`, and `auto()` at `:703` catches only
  `SystemExit`, so the `TypeError` escapes as a traceback;
- `:424` — `if attach_to_registry:` would likewise fire, attaching to a registry entry named
  by an `OptionInfo`, if execution ever got that far.

The eval itself has already succeeded and `_save_custom_results` has already run (`:397`),
so the user loses only the output-writing step — but sees a stack trace, not a result.

## The second instance, with a different symptom

`monitoring/callback.py:524` makes the same three-of-five call during mid-training
auto-eval. There it *is* wrapped in `except Exception` (`:529`), so it degrades to:

```
Auto-eval custom failed: expected str, bytes or os.PathLike object, not OptionInfo
```

That message names no user mistake and points at no fix. Anyone who has configured
`eval.custom_tasks` has been seeing it — after the eval it is reporting as failed had
already run and saved.

## The class, measured

Nothing checks this. Walking all 500 modules under `src/soup_cli/` with `ast` — 89 command
modules declaring **204 typer commands** and **1093 CLI parameters** — and flagging every
bare-name call to a typer command that leaves a typer-defaulted parameter unfilled:

| finding | count |
|---|---:|
| typer commands called as functions with unfilled typer params | **2 sites** (`eval.py:698`, `callback.py:524`) |
| parameters silently bound to `OptionInfo` at those sites | 2 (`output`, `attach_to_registry`) |

Two is a small number today. The point is that it is unbounded and invisible: the next
parameter added to `custom()` joins the leak automatically, and the failure only ever
surfaces as a type error about an internal typer object.

## A second mechanism found by the same sweep: flags that are never read

Same walk, different question — typer parameters never referenced in their own command body:

- **`soup data download --trust-remote-code`** (`commands/data.py:1388`). Help text:
  *"Allow datasets that ship custom Python loaders. Default deny (v0.36.0). Only enable if
  you trust the source."* The parameter is never read. `_hf_download_dataset()`
  (`:1198`) does not accept it and hardcodes `trust_remote_code=False` (`:1208`). So the
  flag cannot be enabled, and a user with a custom-loader dataset gets a denial with no
  explanation. Worse in the other direction: `:1496` prints a **Remote Code Warning** panel
  — *"Downloading this dataset may execute a remote dataset loading script"* — on every
  download, unconditionally, warning about the one thing the hardcoded `False` makes
  impossible.
- **`soup eval auto --trust-remote-code`** (`commands/eval.py:620`), same help text. Never
  read; `auto()` delegates to `benchmark()` and `custom()` without it.

## What I checked that turned out not to be a defect

Stating these because each looked like a finding and was not:

- **`soup data split --train`** (`commands/data.py:892`) is never read, and that is
  deliberate — its own help says *"informational; the train remainder is implied by
  `--val` + `--test`. Accepted for command parity."* A guard needs an allowlist for exactly
  this shape.
- **No RCE in the eval path.** `eval/custom.py:360` calls `transformers.pipeline()` with no
  `trust_remote_code` argument at all, so it inherits the library default — `None` on the
  installed transformers 5.16.1, i.e. not enabled. The inert flag is broken in the safe
  direction. I am not reporting a security hole.
- **My own first sweep produced four false positives.** Matching `ast.Attribute` calls by
  `.attr` made `trainer_wrapper.train(...)` and `model.train()` (`commands/train.py:1555`,
  `:2218`) look like calls to the `train` command leaving 40+ parameters unfilled, and
  `tracker.list_runs(...)` (`commands/runs.py:71`, `:352`) likewise. Restricting to
  bare-name calls drops all four. Worth recording because the false-positive shape is the
  one that would get a guard deleted.

## Proposal — a guard, in this repo's existing idiom

`tests/test_cli_commands_are_not_called_as_functions.py`, alongside the changelog-fragment
guard and the recipe-count ratchet:

- resolve every typer command and its typer-defaulted parameters by AST;
- fail for any bare-name call to one of them that leaves a typer-defaulted parameter
  unfilled, naming the caller, the callee and the leaked parameters;
- fail for any typer parameter never referenced in its own command body, unless it is in an
  explicit allowlist with a reason (seeded with `--train` above, and with the two
  `--trust-remote-code` flags only if you decide they should stay).

### Acceptance criteria

1. The guard fails on `eval.py:698` as it stands today, naming `output` and
   `attach_to_registry` — demonstrated red, then green once the call is fixed.
2. The guard fails when a new typer parameter is added to an existing command that some
   other module calls directly.
3. A method call that merely shares a command's name (`model.train()`) does **not** fail —
   this needs its own test, since it is what a naive version gets wrong.
4. A parameter that is deliberately informational does not fail, via an allowlist entry
   carrying a reason, and the test asserts every allowlist entry is still a real parameter.
5. Runs on every CI cell — no GPU, no MLX, no network.

### Fixes, separately from the guard

The call sites are one line each (`attach_to_registry=None, output=None`), and I would
rather fix the two `--trust-remote-code` flags the way you decide than guess: wire them
through, or remove them and drop the two panels that overstate what the code does.

### Scope limit

The sweep covers bare-name calls only. A command invoked through an alias, a dict of
handlers or `functools.partial` would not be seen; I did not find any, but I did not prove
their absence.

I have the AST walker and the reproduction above. Happy to take this — the crash fix and
the guard as one PR, and the `--trust-remote-code` decision as a follow-up on your word.

Srinivasan8888:
Fixes #752.

## Overlap

None found. `gh issue list --search` on `OptionInfo`, `typer`, `eval auto` and
`trust-remote-code`, and the same over PRs, returns nothing on this. The nearest
relatives are all closed and all about a *declared* setting that no code applies —
#333 (`--dashboard` silently no-ops), #372 (`--no-reexec` drops the user's flags),
#583 (`convergence_detection` accepted, never enforced). This is the same family one
layer up: the value is not dropped, it is replaced by an internal typer object.

## The bug

`soup eval auto` ran the custom eval to completion, printed the results table, saved
them to the tracker — then exited with an uncaught `TypeError`.

Typer fills a command's parameters only when *typer* invokes it. `auto()` calls
`custom()` as a plain Python function with three of its five typer parameters, so
`output` and `attach_to_registry` kept their literal defaults: `OptionInfo` objects.
`OptionInfo` is truthy, so in `custom()`:

- `:404` `if output or attach_to_registry:` fired though neither was requested;
- `:415` `write_target = output or "eval_results.json"` became the `OptionInfo`;
- `:417` `write_eval_json(OptionInfo)` raised `TypeError`, which `:418` does not catch
  (`ValueError, FileNotFoundError` only) and `auto()` does not catch (`SystemExit` only).

`monitoring/callback.py:524` makes the same call during mid-training auto-eval. There
it is wrapped in `except Exception`, so it degraded to
`Auto-eval custom failed: expected str, bytes or os.PathLike object, not OptionInfo`
— a message naming no user mistake, printed after the eval it reports as failed had
already run and saved.

## Shape, and why

Both call sites now pass every typer parameter explicitly (`attach_to_registry=None,
output=None`). That is the whole product change: 10 added lines, 0 removed.

The guard resolves each callee to a **declaration** — one in the calling module, or one
named by a `from X import name` in it — rather than by bare name. That is not
incidental: my first version keyed on the name, merged every `main()` in the tree into
one entry, and reported `autodistill/mlx_worker.py`'s argparse `main(argv=None)` as
leaking fifteen typer options it has never declared. A guard that does that gets muted,
so the false-positive shape has its own test.

## Mutations — 7 applied, 7 killed

Binary copy, `shasum -a 256` verified identical on restore for all three files,
`__pycache__` cleared between runs.

| mutation | killed by |
|---|---|
| revert the `soup eval auto` call site | `test_soup_eval_auto_completes_instead_of_raising_typeerror`, `test_soup_eval_auto_writes_no_eval_json_when_none_was_requested`, `test_auto_passes_every_typer_parameter_of_custom`, `test_no_typer_command_is_called_with_unfilled_typer_parameters` |
| revert the training-callback call site | `test_training_callback_passes_every_typer_parameter_of_custom`, `test_no_typer_command_is_called_with_unfilled_typer_parameters` |
| guard keys commands by bare name | `test_the_guard_ignores_a_plain_function_sharing_a_command_name`, `test_no_typer_command_is_called_with_unfilled_typer_parameters` |
| guard matches attribute calls by `.attr` | `test_the_guard_ignores_a_method_call_sharing_a_command_name`, `test_no_typer_command_is_called_with_unfilled_typer_parameters` |
| guard stops following imports | `test_the_guard_follows_an_imported_command` |
| guard judges `**kwargs` calls too | `test_the_guard_skips_calls_it_cannot_judge` |
| guard is vacuous (never reports) | `test_the_guard_catches_a_synthetic_leak`, `test_the_guard_follows_an_imported_command` |

Each call site is mutated **independently**, so reverting one and forgetting the other
cannot pass.

Two notes I would rather state than have you find:

- **One RED was initially red for the wrong reason.** `test_training_callback_...`
  first failed because my stub eval config omitted `auto_eval`, so `_run_auto_eval()`
  returned before reaching the call at all — it reported all five parameters missing
  rather than the two that leak. Fixed, then re-checked against the reverted call site:
  it now fails naming exactly `['attach_to_registry', 'output']`.
- **Mutation 5 is killed by one test, not two.** Once both call sites are fixed the
  tree is clean, so dropping import-following breaks no repo-wide assertion — only the
  synthetic fixture catches it. That is the honest result, not a gap I patched over.

## Verification

```
tests/test_issue752_cli_commands_called_as_functions.py     11 passed
test_callback / test_eval / test_eval_gate / test_eval_platform /
test_cli / test_cli_startup_is_light / test_issue308 /
test_issue320 / test_issue367 / test_issue752                310 passed in 51.68s
full suite: pytest tests/                                    20600 passed, 106 skipped,
                                                             coverage 83.05% (gate 77%)
ruff check src/soup_cli/ scripts/ tests/                     All checks passed!
git diff -- tests/ | grep -c "^-[^-]"                        0
```

No network, no GPU, no MLX — runs on every cell.

## Scope limits

- **The sweep sees bare-name calls only.** A command reached through an alias, a dict of
  handlers or `functools.partial` would not be seen. I found none, but I did not prove
  their absence.
- **I did not touch the two inert `--trust-remote-code` flags** that the same sweep
  found (`soup data download`, `soup eval auto`), or the two panels that warn about
  remote-code execution the code actually denies. Both are in #752 with measurements;
  they need your decision — wire them through, or remove them — and I would rather take
  that as a follow-up than guess here.
- **No RCE.** `eval/custom.py:360` passes no `trust_remote_code` to
  `transformers.pipeline()`, so it inherits the library default (`None` on the installed
  transformers 5.16.1). The inert flag is broken in the safe direction.

https://claude.ai/code/session_01SKvLpnwG8X3KLFCThsxedb


MakazhanAlpamys:
@Srinivasan8888 — **a second pull request against #752 exists, and you should hear it from me rather than discover it.**

**#757 by @dchaudhari7177** was opened at `2026-09-09T08:44:30Z`, two hours and one minute after yours. You filed #752 at 06:18 and opened this at 06:43, so you were first by a wide margin and you were also the reporter.

**The gap is mine.** I said on #721, two days ago, that I would apply the `claimed` label *and* leave a comment saying whose claim it is, in that order, before replying. On #752 I did neither — it was filed and answered inside three hours while I was away from the repo, and the label went on only just now. `CONTRIBUTING.md` says the comment is the claim and the thread is the record, so until a moment ago the record said this issue was free. Your PR *was* cross-referenced on the issue from 06:43, which is a real signal, but it is not the one this project tells people to read.

**I am not deciding it on arrival order**, and I would rather tell you that up front than have you assume being first settles it. Both are being measured like for like: whether each fixes the root cause or only today's two parameters, whether `eval.py` and `callback.py` are treated consistently, the same mutations on each, a vacuity sweep against unmodified `main`, and — the axis I expect to decide it — **whether adding a new option to that command later makes the suite fail**. A guard that catches the next instance of this bug class is worth more than a correct fix to this one.

Nothing is asked of you right now; CI is green and I have no review items yet. The numbers go in both threads before any decision.

Srinivasan8888:
Thanks for saying it plainly, and the process point needs no apology from you — I would rather a maintainer who says "the gap is mine" than one who quietly picks by timestamp. I am not asking to win on arrival order either.

You named the axis you expect to decide it, so here it is measured rather than asserted. Nothing below is a comparison with #757; those are @dchaudhari7177's numbers to report, not mine.

## "Whether adding a new option to that command later makes the suite fail"

Added a brand-new option to `custom()` — the shape of a change someone makes a release from now, with no knowledge of this bug:

```python
    fail_fast: bool = typer.Option(
        False, "--fail-fast", help="A brand-new option, added a release from now."
    ),
```

Result on `fix/cli-command-optioninfo-leak`:

```
AssertionError: eval.py:706 calls custom() leaving fail_fast bound to OptionInfo

FAILED test_soup_eval_auto_completes_instead_of_raising_typeerror
FAILED test_soup_eval_auto_writes_no_eval_json_when_none_was_requested
FAILED test_auto_passes_every_typer_parameter_of_custom
FAILED test_training_callback_passes_every_typer_parameter_of_custom
FAILED test_no_typer_command_is_called_with_unfilled_typer_parameters
5 failed, 6 passed
```

The guard names the caller, the callee and the new parameter. It is not a list of the two parameters known today — `attach_to_registry` and `output` appear nowhere in it. It resolves each callee to a declaration and compares the call's keywords against whatever that command declares *at the time the suite runs*.

## Vacuity sweep against unmodified `main`

Same test file, dropped onto `9c7b1fe9` with no source change:

```
5 failed, 6 passed
```

The five that fail are the behavioural tests and the repo-wide guard. The six that pass are the guard's own fixture tests — synthetic modules in `tmp_path` — which are meant to be independent of repo state; that is what makes them able to prove the guard can fail at all.

## `commands/eval.py` and `monitoring/callback.py` treated identically

Both call sites get the same fix and are **mutated independently** in the table in the PR body, so reverting one and forgetting the other cannot pass:

| mutation | killed by |
|---|---|
| revert the `soup eval auto` call site | 4 tests incl. `test_no_typer_command_is_called_with_unfilled_typer_parameters` |
| revert the training-callback call site | `test_training_callback_passes_every_typer_parameter_of_custom` + the repo-wide guard |

The callback is the one I would not want dropped: it is wrapped in `except Exception`, so it never crashed — it printed `Auto-eval custom failed: ... not OptionInfo` after the eval it reported as failed had already run and saved to the tracker.

## What the guard does not do, so the numbers are not oversold

Bare-name calls only. A command reached through an alias, a dict of handlers or `functools.partial` is invisible to it. I found none in the tree and I did not prove their absence. It also skips any call using `*args`/`**kwargs`, since those cannot be judged statically — with a test pinning that skip, because silently judging them would produce exactly the false positives that get a guard deleted.

One correction to my own PR body while you are here: mutation 5 ("guard stops following imports") is killed by **one** test, not two — once both call sites are fixed the tree is clean, so dropping import-following breaks no repo-wide assertion and only the synthetic fixture catches it. That is in the body already; I am repeating it because it is the weakest row in the table and I would rather point at it than have it found.

Happy to leave it exactly here while you measure both.

MakazhanAlpamys:
@Srinivasan8888 — **the comparison is done and yours is the one I am taking. Not because it was first.**

## Where you are equal, which is most of it

The production change is **byte-identical** between the two PRs — the same four lines (`attach_to_registry=None`, `output=None`) at the same two call sites, `commands/eval.py:698` and `monitoring/callback.py:524`. Both pass all five of `custom()`'s parameters; neither hardcodes a subset.

And the axis I publicly said would decide it **tied**. I told both of you this morning that the question I cared about was whether adding a new option to `custom()` later would make the suite fail. Adding a plausible `--foo` without updating either fix:

```
#754   3 failed, 8 passed
#757   1 failed, 1 passed
```

Both notice. Both are repo-wide AST guards, neither is a hand-maintained list. I am stating that plainly rather than quietly moving to different grounds.

## What decided it

**Your guard has a test proving it can still fail; theirs does not.** Inserting `return []` at the top of their leak-finder makes it report nothing, ever, and their suite stays green — 2 passed, unchanged. The same mutation fails your `test_the_guard_catches_a_synthetic_leak`.

That is the property this repo learned the hard way on its ANSI scanner: a repo-wide guard that finds zero offenders today looks exactly the same whether it is working or broken, so the load-bearing test is the one that proves it can still fire. Your six mutations, six killed; theirs five of six.

You also have two **behavioural** tests that run the real `auto()` and assert it completes rather than raising, so if the AST guard is ever muted there is still something testing the actual crash. And 5 of your 11 tests fail against unfixed `main`, which is the right proportion — the other six are meta-tests of the guard against synthetic false-positive shapes, which is not padding.

Beyond that: `ruff` clean where theirs fails `F401`, a correctly PR-numbered changelog fragment where theirs has none, and 14 of 14 checks green with `head_sha` confirmed equal to `dfe8a245`.

## One thing theirs does better, and it is coming

Their guard **catches a leak passed positionally**; yours deliberately bails out with `if call.args: continue` and pins that as intentional in `test_the_guard_skips_calls_it_cannot_judge`. On purpose-built fixtures theirs catches a real positional leak that yours misses — while also false-alarming on a correct positional call, because it compares argument *counts* rather than resolving declaration order.

Your failure direction is the safer one and this repo has the receipts: #496's guard fired on a correct future state, and the ANSI scanner's first normaliser check flagged 31 correct assertions across 18 files. A guard that fires on correct code gets deleted by the next person in a hurry. So the conservative choice was right.

But the leak class is real, and I have asked @dchaudhari7177 whether they want to land positional-argument resolution as a follow-up, credited to them. It changes nothing you need to do here — I am telling you so it does not arrive as a surprise edit on top of your work.

## Before I merge

Nothing is blocking. I will re-check `gh pr checks` on the current head and merge. If you would rather fold the positional-resolution improvement in yourself instead of leaving it to a follow-up, say so in the next few hours and I will hold.

One correction I owe you from earlier today, unrelated to this PR but on the record: your reply on #739 was right and my review was wrong about how `callback.py` writes its rows. That is in that thread and in `CONTRIBUTORS.md`.

MakazhanAlpamys:
@Srinivasan8888 — merged as `1490c850`. **Two mutations re-run at merge, both dead.**

```
revert the eval.py fix (the #752 defect)     4 failed, 7 passed
  test_soup_eval_auto_completes_instead_of_raising_typeerror
  test_soup_eval_auto_writes_no_eval_json_when_none_was_requested
  test_auto_passes_every_typer_parameter_of_custom
  test_no_typer_command_is_called_with_unfilled_typer_parameters

find_optioninfo_leaks returns [] always      2 failed, 9 passed
  test_the_guard_catches_a_synthetic_leak
  test_the_guard_follows_an_imported_command
```

The second is the one that decided this over #757, and it is worth being precise about why: their guard under the same mutation stayed at `2 passed`. A repo-wide guard that finds zero offenders today looks identical whether it works or is broken, so the test proving it can still fire is the only thing holding it up.

Your `fail_fast` demonstration is the other half and I did not need to re-run it — the guard names the caller, the callee and the new parameter, and neither `attach_to_registry` nor `output` appears anywhere in it. That is the difference between a fix and a fix that survives the next person.

**You reported your own numbers and refused to report theirs.** "Those are @dchaudhari7177's numbers to report, not mine" — in a comparison where you had every incentive to characterise the other PR. That is the behaviour that makes a collision survivable for everyone in it.

**You ran the vacuity sweep on yourself unprompted** and reported that 6 of 11 pass on unmodified `main`, with the reason. Most people submit the number that flatters them; you submitted the one that invited the question and then answered it.

## The follow-up, so it is not a surprise

@dchaudhari7177's guard catches a leak passed positionally, which yours deliberately skips. I have offered them that as a follow-up with their credit. Your conservative direction was right — this repo has twice shipped guards that fired on correct code, and #496 went red on exactly the migration its exception existed to permit — so the follow-up is an addition to your design, not a correction of it.

Ruff clean, 14 of 14 SUCCESS on `dfe8a245`, `head_sha` confirmed equal to the head I merged. Nothing deleted or weakened.

PR #756, opened 2026-09-09
feat(doctor): report which settings your backend does not read (#755)

### LINKED ISSUE #755: Nothing tells you before a run which of your 275 settings your backend will actually read — and it cannot be inferred, only declared
## The problem, stated as the user meets it

You write a config, it validates clean, training runs, and some of what you asked for
never happened. You find out afterwards — from a loss curve, or not at all.

That has been filed one field at a time: #683 (`train_on_responses_only` dropped on MLX),
#686 (warmup, scheduler, weight decay, optimizer choice — four in one backend), #745
(LoRA+ wired in 3 trainers, silently ignored by 13), #749 (`max_grad_norm` honoured by
sixteen transformers trainers, dropped on MLX). #748 counts 36 issues in this family by
tracker search.

`TrainingConfig` declares **237** fields and `DataConfig` **38** — 275 settings, and
nothing tells you before a run which of the ones *you set* will actually be read on the
task and backend *you chose*.

## What already exists, and why it is the right shape

`trainer/mlx_sft.py:139-179` builds an `unsupported` list and prints

```
MLX backend ignores: GaLore, Ring Attention, training.seed (MLX seeds through mx.random), ...
```

That mechanism is proven and already user-facing. Its limits are the whole proposal:

- **nine entries, one file, one backend.** Nothing equivalent exists for the sixteen
  transformers trainers, where #745's thirteen-trainer LoRA+ gap lives;
- **hand-maintained, and it drifted.** `max_grad_norm` was not in that list while MLX was
  dropping it — that is #749, found by reading code, not by any check;
- **it fires during the run**, not before it.

## Three implementations I measured and rejected

Recording these because each is the obvious answer and each is wrong, and a reviewer
would rightly ask why I did not just do them.

**1. Static reachability over the import graph — detects nothing.** Count a field as
reachable from a trainer if any module in its transitive `soup_cli` import closure reads
it as an attribute or names it as an exact string. Every trainer's closure is ~150
modules, i.e. most of the package, so almost everything looks reachable. Against four
gaps that are independently known to be real:

| field | flagged unreachable from `mlx_sft`? | ground truth |
|---|---|---|
| `max_grad_norm` | no | MLX gap (#749) |
| `warmup_steps` | no | MLX gap (#686) |
| `weight_decay` | no | MLX gap (#686) |
| `loraplus_lr_ratio` | no | not wired (#745) |

**Zero of four.** The set of fields it says MLX lacks but SFT has is empty.

**2. Static reads of the trainer module alone — false positives that would get it
deleted.** It does find the four above, but `sft.py` does not itself read
`train_on_responses_only` (that lives in `data/sft_format.py:55`) or `loraplus_lr_ratio`
(in the LoRA+ helper from #738), so it would report two gaps the transformers path does
not have. A checker that cries wolf about masking is one nobody reads twice.

**3. Runtime tracing under `--dry-run` — observes nothing relevant.** A config proxy
recording `__getattr__` is accurate by construction, but `--dry-run` raises `typer.Exit`
at `commands/train.py:1255` after validating data and **before the trainer wrapper is
constructed**, so no trainer field read ever happens. Making it go further means loading
a model, which a pre-flight check cannot do.

The honest conclusion: **which fields a backend honours cannot be inferred; it has to be
declared.** Inference is exactly what I tried three ways and could not make trustworthy.

## Proposal

**1. A declared support registry.** Promote the `unsupported` list out of `mlx_sft.py`
into one reviewed table keyed by (task, backend), covering the transformers trainers as
well as MLX. Each entry names a field, a status (`honoured` / `ignored` / `rejected`) and
a reason, and — where one exists — an issue number. The MLX nine and #745's thirteen
seed it.

**2. `soup doctor --config soup.yaml`.** `doctor` today is 16 functions of environment
diagnostics and reads no config at all (`grep -c "load_config\|soup.yaml"` → 0). Add a
config leg that loads the user's config, diffs the fields they actually set against the
registry for their task and backend, and reports only what affects them:

```
Config check — task=sft backend=mlx
  12 settings, 9 honoured

  ignored on this backend:
    training.max_grad_norm      MLX clips nowhere (#749)
    training.warmup_steps       MLX uses its own schedule (#686)
    data.mask_history           declared, no consumer anywhere (#748)
```

**3. A guard so the registry cannot drift.** Narrow enough to be trustworthy, which the
measurements above define: every registry entry names a real field on the model; a field
marked `ignored` for a backend must not be read by that backend's trainer module or its
declared helper modules; the helper list is explicit, because that is precisely where
implementation 2 went wrong.

### Acceptance criteria

1. `soup doctor --config` names `training.max_grad_norm` as ignored for `task=sft
   backend=mlx` on today's `main`, and stops naming it when #750 lands — demonstrated
   both ways, not described.
2. It reports only fields the user actually set, not all 275.
3. A field marked `honoured` that no code reads fails the guard.
4. A registry entry naming a field that no longer exists fails the guard.
5. The transformers path is not reported as missing `train_on_responses_only` or
   `loraplus_lr_ratio` — the false positives from implementation 2, each with its own
   test.
6. Runs on every CI cell — no GPU, no MLX, no network, no model download.

### Scope limits

- The registry is **declared, not derived**. It is only as right as its review, and I
  would rather be honest that this is a maintained table than pretend a walker can
  infer it. The guard bounds the drift; it cannot originate the truth.
- Per-*trainer* granularity, not per-code-path: a field honoured on one branch of a
  trainer and not another reads as honoured.
- I would ship the registry plus `doctor --config` for `sft` and MLX first, with the
  other tasks seeded from #745's existing `_LORAPLUS_NOT_IMPLEMENTED` list, rather than
  claiming all 275 × 18 cells are reviewed on day one.

Taking this, on the understanding that the registry's shape is your call and I will
rework it on your word — this is the same architectural decision point as #745, and I
would rather show you a small honest slice than a large inferred one. No hardware needed
beyond a laptop; no Apple Silicon required for the guard or the doctor leg.

Srinivasan8888:
Fixes #755.

## Overlap

None. Nothing in the tracker or in open PRs covers a config-effectiveness check;
`soup doctor` reads no config today (`grep -c "load_config\|soup.yaml" src/soup_cli/commands/doctor.py`
→ `0`). The issues this serves — #683, #686, #745, #749 — are each a single field,
and #751 is the developer-facing guard for a different question (does *anything*
read this field, rather than does *this backend*).

## What it does

```
$ soup doctor --config soup.yaml

Config check - task=sft backend=mlx
┏━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ setting                ┃ status  ┃ why                                       ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ training.max_grad_norm │ ignored │ MLX clips nowhere; sixteen transformers   │
│                        │         │ trainers clip at this value (#749)        │
│ training.weight_decay  │ ignored │ not forwarded to the MLX optimizer (#686) │
│ training.optimizer     │ ignored │ MLX always builds mlx.optimizers.Adam     │
│                        │         │ (#686)                                    │
│ training.seed          │ ignored │ MLX seeds through mx.random, not this     │
│                        │         │ field                                     │
└────────────────────────┴─────────┴───────────────────────────────────────────┘
  4 setting(s) written here are not read on backend=mlx.
```

The same config with `backend: transformers` prints
`Every setting this config writes is read on this backend.`

Only fields the user **wrote** are listed — `model_fields_set` separates those from
the 275 sitting at their schema default. A wall of 275 rows is not a pre-flight check.

## Why the table is declared rather than derived

This is the part I would most like you to push back on, so here is the evidence
rather than the assertion. I built all three obvious inference strategies and
measured them against five gaps that are independently known to be real.

**1. Reachability over the import graph — zero of five.** Every trainer's transitive
`soup_cli` closure is ~150 modules, i.e. most of the package, so almost everything
looks reachable:

| field | flagged unreachable from `mlx_sft`? | read by `trainer/sft.py`? |
|---|---|---|
| `max_grad_norm` | no | yes |
| `warmup_ratio` | no | yes |
| `weight_decay` | no | yes |
| `optimizer` | no | yes |
| `loraplus_lr_ratio` | no | no (#738 helper) |

**2. Reads of the trainer module alone — false positives that would get it deleted.**
It does find those five, but `sft.py` reads neither `train_on_responses_only` (that
is `data/sft_format.py:55`) nor `loraplus_lr_ratio`, so it reports two gaps the
transformers path does not have. Both have their own regression test here.

**3. Runtime tracing under `--dry-run` — observes nothing.** A recording config proxy
is accurate by construction, but `--dry-run` raises `typer.Exit` at
`commands/train.py:1255` after validating data and **before the trainer wrapper is
constructed**. Going further means downloading a model, which a pre-flight check
cannot do.

So the registry is reviewed, and the guard's job is to stop it rotting — which is
exactly the failure that produced #749, where `max_grad_norm` was missing from
`mlx_sft.py`'s hand-maintained list while MLX dropped it.

## The guard runs in both directions

Each entry carries `trainer_reads`: whether the backend's trainer reads the field
*in order to warn about it*. Five of the twelve are read nowhere; seven are read
solely to build the existing `MLX backend ignores:` line.

- `trainer_reads=False` **and the trainer now reads it** → the field was wired and the
  entry is stale;
- `trainer_reads=True` **and the trainer no longer mentions it** → the warning was
  deleted.

Acceptance criterion 1 asked for both directions demonstrated. Rather than trust a
fixture, I ran the check against the **real `mlx_sft.py` from the #750 branch**:

```
drift check against the REAL #750 branch's mlx_sft.py:
   training.max_grad_norm: declared unread, but soup_cli/trainer/mlx_sft.py now
   reads it — the field was wired; remove or reclassify the entry

drift check against this branch (main, no #750):
   (none - clean)
```

So when #750 merges, this suite goes red until the entry is removed. It cannot keep
telling users a setting is ignored after it stops being ignored.

## Mutations — 7 applied, 7 killed, one only after a real fix

Binary copy, `shasum -a 256` verified identical on restore for all three files,
`__pycache__` cleared between runs.

| mutation | killed by |
|---|---|
| report every registry field, not only those the user set | `test_only_fields_the_user_actually_set_are_reported` |
| ignore the backend (always use the default) | `test_max_grad_norm_is_reported_ignored_on_mlx_sft`, `test_doctor_config_names_the_ignored_field_and_its_issue`, `test_only_fields_the_user_actually_set_are_reported` |
| drop the `max_grad_norm` entry | the above three, plus `test_wiring_a_field_makes_its_stale_entry_fail` |
| drift check ignores newly-wired fields | `test_wiring_a_field_makes_its_stale_entry_fail` |
| drift check ignores deleted warnings | `test_deleting_a_warning_makes_its_entry_fail` |
| `doctor` never reads the config | `test_doctor_config_names_the_ignored_field_and_its_issue` |
| field-existence guard accepts anything | `test_an_entry_naming_a_field_that_no_longer_exists_is_caught` |

**The last one survived the first run, and the survivor was correct.** Every registry
entry is valid today, so deleting the existence check entirely left the suite green —
a vacuous guard, and observing it pass proved nothing. I extracted the check into a
helper and added a test that feeds it a renamed field and a bad namespace (acceptance
criterion 4). Re-ran the mutation: killed. Recording it because the fix was a real
gap in my tests, not a mutation that was unfair.

## Correction to #755's body, already posted there

The issue's rejection table named `training.warmup_steps`. **No such field exists** —
this repo's own unknown-key warning (#628) catches it and suggests `warmup_ratio`. I
took the name from #686's prose title without checking the schema. Re-measured with
the correct names (`warmup_ratio`, `schema.py:1096`); the result is unchanged, and I
added `optimizer`, so it is now zero of five rather than zero of four.

## Verification

```
tests/test_issue755_backend_support_registry.py       14 passed
tests/test_cli_startup_is_light.py                    34 passed  (import is lazy)
pytest tests/                                         20603 passed, 106 skipped,
                                                      coverage 83.02% (gate 77%)
ruff check src/soup_cli/ scripts/ tests/              All checks passed!
git diff -- tests/ | grep -c "^-[^-]"                 0
soup doctor --config <mlx cfg>                        run live, output above
```

No network, no GPU, no MLX, no model download.

## Scope limits

- **Twelve entries, one pair: `(sft, mlx)`.** Every other task/backend reports nothing
  rather than guessing. #745's `_LORAPLUS_NOT_IMPLEMENTED` list is the obvious next
  seed, but that is the architectural question you are already holding #745 on, so I
  did not pre-empt it.
- **Per-trainer granularity, not per-code-path.** A field honoured on one branch of a
  trainer and not another reads as honoured.
- **The registry is only as right as its review.** The guard bounds drift; it cannot
  originate the truth. I would rather say that than imply a walker inferred it — I
  tried three and none of them worked.
- **`mlx_sft.py` still keeps its own `unsupported` list.** Making the trainer read the
  registry instead would remove the duplication by construction, and is the obvious
  follow-up, but it touches a trainer and I would rather you settle the registry's
  shape first.

https://claude.ai/code/session_01SKvLpnwG8X3KLFCThsxedb


MakazhanAlpamys:
@Srinivasan8888 — **the design is right, the drift guard demonstrably works, and it caught this PR's own staleness before I did. Blocking, and the reason is my fault rather than yours.**

## The blocker, and where it came from

Four of the twelve registry entries are false against current `main`. `training.warmup_ratio`, `training.weight_decay`, `training.optimizer` and `training.scheduler` are declared ignored citing #686 — and **#686 was closed six hours after you opened this**, by your own #734, which I merged at `765f5b4`. Your base is `254351e1`, the commit immediately before it.

```
$ git log 254351e1..origin/main -- src/soup_cli/trainer/mlx_sft.py
6b999eb feat(mlx): report validation loss from on_val_loss_report (#739)
765f5b4 fix(mlx): wire warmup, scheduler, weight decay and optimizer choice (#734)
```

Merged as-is, `soup doctor --config` would tell every MLX SFT user that four settings which are now honoured are ignored. That is the "worse than no feature" case the issue itself names — a tool that reports a setting is dropped when it is read.

**Your CI is green because it ran against a base that predates the merge**, which is a useful reminder that green on a stale base is not evidence of anything.

## The part that makes me want this merged rather than reworked

I built the merge with current `main` and ran your own suite:

```
FAILED test_the_registry_matches_what_the_backend_trainer_actually_reads

E   training.warmup_ratio: declared unread, but soup_cli/trainer/mlx_sft.py now
    reads it -- the field was wired; remove or reclassify the entry
E   training.weight_decay: ...
E   training.optimizer:    ...
E   training.scheduler:    ...

1 failed, 13 passed
```

**Your guard fired on a real drift that happened during the PR's own lifetime, named all four fields, and told the reader exactly what to do about each.** That is better validation than any mutation I could have written, and it is the whole argument for the feature: a hand-maintained table is a liability unless something fails when it rots, and yours fails.

Three synthetic mutations also die — the lookup returning empty (4 tests), the read/unread verdict inverted (3), the declaration emptied (6).

## What the guard does not catch, stated plainly

Not a criticism of the implementation — the issue is honest that a guard bounds drift rather than originating truth — but worth being precise so nobody over-trusts it:

- **A new config field added and never declared fails nothing.** There is no completeness check against `TrainingConfig.model_fields`.
- **A brand-new gap opening fails nothing**, which is #749's own failure mode and undetectable by construction.
- **Issue acceptance criterion 3 is not implemented.** It asks that a field marked `honoured` which no code reads fails the guard; `STATUSES` is `{ignored, rejected}` with no `honoured` concept, so the criterion is met by omitting it rather than by a check. `REJECTED` is also declared and used by zero entries.
- **The matcher is namespace-blind.** It splits on the first dot and matches the bare name against every attribute and string constant in the module, so `data.train_on_prompt` and a hypothetical `training.train_on_prompt` are indistinguishable. It errs toward false-firing on "declared unread", which is the safe direction — but the `trainer_reads=True` direction is satisfied by a mere mention.

## Three smaller things before merge

**Narrow terminals silently eat the payload.** At `COLUMNS=60` Rich truncates the setting name:

```
| training.max_grad_no... | ignored | MLX clips nowhere;    |
```

The field name is the entire content of that row. `overflow="fold"` on that column, or plain lines instead of a table.

**Two new assertions read `capsys` output raw**, with no ANSI strip and no whitespace collapse. They pass under `FORCE_COLOR=1` today only because Rich happens to style per cell, keeping `max_grad_norm` contiguous. This project's CI has gone red four times on exactly this class, and passing by luck is not passing by construction.

**The invalid-config path is dead code.** `loader.load_config` raises `SystemExit`, which is a `BaseException` and so escapes your `except Exception`; `soup doctor --config bad.yaml` exits 1 and never prints your message. The `FileNotFoundError` branch does work. Neither error path has a test.

**And `docs/commands.md:218` still reads `soup doctor [--nccl] [--disk]`.** A new user-facing flag needs the command reference and the topic page.

## One scope note, not a change request

`REGISTRY` covers exactly one pair, `("sft", "mlx")`. `unsupported_for` returns empty for everything else, so #745's thirteen preference and RL trainers are not covered — `soup doctor --config` on a transformers DPO run setting `loraplus_lr_ratio` still says nothing. Your docstring is honest about this; I am recording it so the changelog framing is not read more broadly than the code.

Import weight is correct — `backend_support` is imported inside `_check_config_support`, `test_cli_startup_is_light.py` passes, and the module has no heavy imports.

**Rebase, drop those four entries, and fix the reasons that cite #686. If you would rather I did the rebase and the entry deletions myself and merged with the work credited to you, say so — this drift is a direct consequence of a merge I made while your PR was open.**

MakazhanAlpamys:
@Srinivasan8888 — follow-up, and this one is in your favour.

**#750 just merged as `c7c6314a`, so a fifth entry is now stale**: `training.max_grad_norm` is read at `mlx_sft.py:502` on current `main`. That is five of twelve, and all five went stale because of merges I made today while this PR sat open.

**You predicted this one in a code comment.** `backend_support.py:51-54`:

> *"The guard uses this in both directions: a `False` entry that the trainer starts reading means the field was wired and the entry is stale (this is what happens to `max_grad_norm` when #750 lands), and a `True` entry the trainer stops reading means the warning was deleted."*

You named the field, named the PR that would invalidate it, and wrote the guard that catches it — before either landed. That is a better argument for this feature than anything in the issue, and it is worth saying plainly: the drift I am blocking you on is the drift your own design was built to survive, and it did.

So the rebase now removes five entries rather than four, and the reasons for the four citing #686 plus the one citing #749 all need updating, since both issues are closed.

The offer from my earlier comment stands and I will widen it: given that every one of these five was caused by me merging your other PRs on top of your open one, I am happy to do the rebase and the entry surgery myself and merge with the work credited to you. Say the word either way — and if you would rather do it, no explanation needed.

The other items from that review are unchanged: the narrow-terminal truncation, the two raw `capsys` assertions, the dead `except Exception` on the invalid-config path, and `docs/commands.md`.

MakazhanAlpamys:
@Srinivasan8888 — short one, because I do not want this waiting on a decision that costs you a word.

The offer in my last two comments needs a yes or a no from you, and **the default if you say nothing is that I do it**: rebase onto current `main`, remove the five now-stale entries, update the reasons that cite #686 and #749 (both closed), and merge with the work credited to you — your commits, your `CONTRIBUTORS.md` entry, your changelog fragment. I will list every entry I touched in the merge commit so the surgery is reviewable rather than silent.

**Deadline: 2026-09-11 18:00 UTC.** Before then, a "no, I will rebase it myself" wins automatically and I will not touch the branch.

I am pushing this rather than leaving it open because all five went stale from merges *I* made on top of your open PR, and the fair split is that the cost of that lands on me rather than on you. You are also mid-flight on #751 and I would rather you finish that than context-switch to entry surgery.

Separately, thank you for filing #759 as an issue with four options and a stated lean instead of an allowlist line. I have reproduced the count independently and answered there; the direction decision is mine and I have not made it yet.

MakazhanAlpamys:
Consolidating, because my earlier note flagged one stale entry and the real number is larger. The design decision is right and I am not asking you to re-litigate declared-versus-inferred. But this cannot merge as it stands, and **your own guard is what proves it** — which is the strongest endorsement the mechanism could get.

**1. Blocking: the registry is stale against `main` on five entries, not one.** The branch is cut at `254351e1`, before #734 and #750 landed. I merged current `main` into your branch and ran the suite:

```
FAILED ::test_the_registry_matches_what_the_backend_trainer_actually_reads
E  training.max_grad_norm: declared unread, but soup_cli/trainer/mlx_sft.py now reads it
E  training.warmup_ratio:  ... now reads it ...
E  training.weight_decay:  ... now reads it ...
E  training.optimizer:     ... now reads it ...
E  training.scheduler:     ... now reads it ...
1 failed, 13 passed
```

On that merged tree `soup doctor --config` prints three false rows to a user, telling them `max_grad_norm`, `weight_decay` and `optimizer` are ignored on MLX when they are honoured. Please rebase and delete those five. Note #755's AC1 is now only satisfiable in one direction — #750 has shipped, so `max_grad_norm` must come *out* rather than be asserted — and `test_max_grad_norm_is_reported_ignored_on_mlx_sft`, `test_only_fields_the_user_actually_set_are_reported` and `test_doctor_config_names_the_ignored_field_and_its_issue` all need a different anchor. `training.seed` still holds.

**2. Please state the reduced scope honestly in the fragment.** After that rebase the registry is 7 entries, all `trainer_reads=True` — every one duplicating a warning `mlx_sft.py` already prints at runtime. That is still worth having, because pre-flight beats mid-run, but `changelog.d/0.74.0/756.added.md` currently says "twelve reviewed entries" and implies more than what will ship.

**3. A hole the guard does not cover, found by mutation.** I fabricated an entry declaring `data.mask_history` ignored on `(sft, mlx)` — an unfounded claim — and the suite stayed green, 14/14. The drift check grounds on *name presence in `mlx_sft.py`*, so a false `ignored` entry for any field that file does not textually mention is undetectable. The symmetric case is the one that will actually bite: a field wired entirely inside a helper (`mlx_optim.py`, `mlx_masking.py`) without its name surfacing in `mlx_sft.py` would not trip the guard either. #734 only tripped it because it happened to leave `warmup_ratio` and `weight_decay` visible at `mlx_sft.py:491-492` — by luck, not by the check working. #755's AC3 asked for "declared helper modules"; `TRAINER_MODULES` ships with none. Please make it `("sft","mlx") -> ("trainer/mlx_sft.py", "trainer/mlx_optim.py", "trainer/mlx_masking.py")` and scan the union.

**4. Smaller items.** `test_deleting_a_warning_makes_its_entry_fail` hardcodes the literal `"use_galore"` in its assertion, so deleting that entry kills the mutation only by fixture coincidence — parametrise it over the `trainer_reads=True` entries. `REJECTED` is declared and never used, and there is no `honoured` status, so #755 AC3 is unimplemented — add it or say in the docstring it is deliberately deferred. `config: str = typer.Option(None, ...)` should be `str | None`. `soup doctor --config missing.yaml` prints red and exits 0; for a leg that CI might gate on, a non-zero exit on an unreadable file is probably what you want — your call, but decide it explicitly. And `docs/commands.md:218` still reads `soup doctor [--nccl] [--disk]`.

For the record, everything else checked out on the branch as submitted: 14/14 on your file, 89 passed on `pytest tests/ -k doctor` (108s), `test_cli_startup_is_light.py` 34/34 so the import stays lazy, the #752 and changelog-fragment guards green, ruff clean, zero removed test lines. I also confirmed there is no interaction with #751, which merged today: registry and allowlist have zero field overlap, no file conflict, and I ran #751's guard with your `backend_support.py` in the tree — still 29 passed, because your registry stores field names as bare string constants and that walker structurally never collects those. The two cannot launder each other, which is what I most wanted to be sure of before landing both.

Srinivasan8888:
**Rebased onto `b671a6be`, and the MLX gap list shrank from twelve entries to seven — because this PR's own guard caught that five of them had become false.**

Reporting it in full rather than quietly shipping a shorter table.

## What happened

#734 and #750 merged after this PR was opened. Both wire fields the registry declared MLX ignores. The rebase was textually clean — nothing would have flagged this — but the guard went red naming every stale entry:

```
training.max_grad_norm: declared unread, but soup_cli/trainer/mlx_sft.py now reads it
                        — the field was wired; remove or reclassify the entry
training.warmup_ratio:  (same)
training.weight_decay:  (same)
training.optimizer:     (same)
training.scheduler:     (same)
```

That is acceptance criterion 1's second direction firing on real code rather than on the fixture I demonstrated it with. Without it, `soup doctor --config` would have told an MLX user that five settings were ignored when all five are honoured — the exact failure the registry exists to prevent, pointed the other way.

## Verified honoured, not merely read-to-warn, before deleting

Each read applies the value:

- `max_grad_norm` — `_clipping_optimizer(build_optimizer(optimizer_plan), float(cfg.training.max_grad_norm))`, `mlx_sft.py:502`
- `warmup_ratio`, `weight_decay`, `scheduler` — passed into `plan_optimizer()`, `:490-492`
- `optimizer` — `getattr(cfg.training, "optimizer", "adamw_torch")`, `:489`

`optimizer` was the one worth checking twice: it appears in `mlx_sft.py` only inside docstrings as an attribute, so it looked like it might be a false match on an unrelated `.optimizer`. It is a genuine string-constant read.

So all five are removed rather than reclassified — a field the backend honours does not belong in a table of gaps.

## A consequence I would rather state than hide

Every one of the seven surviving entries is `trainer_reads=True`. That leaves the guard's "declared unread, then wired" direction with **no live instance**. Rather than delete the direction, `test_wiring_a_field_makes_its_stale_entry_fail` now pins it against a synthetic registry, with a docstring recording that it was a real entry until #750 landed. A direction nothing exercises is a direction that rots — which is the same argument as your `TestTheScannerCanActuallyFail` point on #751.

## Live output now

```
Config check - task=sft backend=mlx
│ training.seed       │ ignored │ MLX seeds through mx.random, not this field │
│ training.use_galore │ ignored │ GaLore has no MLX implementation            │
  2 setting(s) written here are not read on backend=mlx.
```

The same config previously listed `max_grad_norm` and `weight_decay` too. It no longer does, because they work.

## Verification

```
tests/test_issue755_backend_support_registry.py     14 passed  (unchanged count)
pytest tests/                                       20804 passed, 106 skipped,
                                                    coverage 83.09% (gate 77%)
ruff check src/soup_cli/ tests/                     All checks passed!
git diff -- tests/ | grep -c "^-[^-]"               16
```

**The 16 removed test lines, each accounted for:** every one is a `max_grad_norm` reference replaced by a `training.seed` equivalent, plus the docstring that said "when #750 lands" — now past tense. No test was deleted or weakened; the file still holds 14. The `--output` assertion that checked for `749` now checks the reason text instead, since the entry carrying that issue number is gone.

## Note on my own method, which also affects #752

Running this suite from a `git worktree` is not isolated here: `_editable_impl_soup_cli.pth` pins `soup_cli` to the primary checkout, so `import soup_cli…` resolves to a different tree than the one under test while the AST guards correctly read the worktree. Everything above was run with `PYTHONPATH` set to the worktree. I have posted a correction on #752 for a count that mistake inflated.

MakazhanAlpamys:
Re-reviewed at `6f0a62ec`. **Item 1 is closed, and it is closed the right way** — you did the rebase yourself rather than take my offer, you let your own guard block your own PR, and you reported the shrink in full instead of quietly shipping a shorter table. I verified the five removals are correct rather than convenient: `max_grad_norm` at `mlx_sft.py:502`, and the other four through `plan_optimizer()` at `:489-492`. The registry is 7 entries, all `trainer_reads=True`, and the guard is green on the rebased tree. That is acceptance criterion 1's second direction firing on real code, which is exactly what it was for.

Items 2 through 8 are untouched — the new commit reaches only `backend_support.py` and the test file. Two of them are blocking, because **both key mutations still survive**, and they are mutations against the guard this PR is built around.

## Blocking 1: `TRAINER_MODULES` still reads one file, so the #734 case is still undetected

Still `{("sft","mlx"): "soup_cli/trainer/mlx_sft.py"}` — a single `str`, and `_registry_drift` opens exactly that one file. I built the case properly this time rather than relying on the argument: a synthetic tree where a field is honoured **only** inside `trainer/mlx_optim.py` and its name never appears in `mlx_sft.py`, with the entry declared `trainer_reads=False`:

```
helper mlx_optim.py mentions warmup_ratio : True
mlx_sft.py mentions warmup_ratio          : False
_registry_drift problems                  : []
>>> MUTATION SURVIVED - guard silent <<<
```

One supporting datum that makes this concrete rather than theoretical: I scanned every `TrainingConfig`/`DataConfig` field for one read by `mlx_optim.py` or `mlx_masking.py` but absent from `mlx_sft.py`. **There are none today.** That is the whole point — #734 tripped your guard by luck, because it happened to leave `warmup_ratio` and `weight_decay` visible at `mlx_sft.py:491-492`. The next helper-owned wiring will not.

## Blocking 2: a fabricated `ignored` entry is still undetectable

Added `SupportEntry("data.mask_history", IGNORED, "FABRICATED: unfounded claim...")`. It is a real schema field that `mlx_sft.py` never mentions:

```
14 passed, 2 warnings in 21.32s     <- unchanged from clean
```

and `soup doctor --config` renders the fabrication to a user:

```
data.mask_history   ignored   FABRICATED: unfounded claim, mlx_sft.py never mentions this field
```

Both of these close together if `TRAINER_MODULES` becomes a tuple per key and the drift check scans the union.

## The parametrise item, with the measurement that makes it worth doing

I deleted the real `training.use_flash_attn` entry: **14 passed, survived.** Then I measured the blast radius, because I wanted to know whether `use_galore` was special:

```
coverage test guards 4/7 entries against deletion
deletion-unguarded: ['training.use_flash_attn', 'training.use_galore', 'training.use_ring_attention']
```

`test_registry_covers_every_field_the_mlx_trainer_warns_about` only sees dotted names inside `.append()` string constants, so three entries have no deletion guard at all — and `use_galore` is guarded solely by the hardcoded literal in the other test. Parametrising over the `trainer_reads=True` entries closes all three at once.

## Still open, small

- **The changelog fragment still says "twelve reviewed entries"** and the PR ships seven.
- `REJECTED` at `:36-37` is referenced nowhere in `src/` or `tests/`, and there is still no `honoured` status, so #755's third criterion is unimplemented — implement or state the deferral in the docstring. New after the rebase: all 7 entries now have `issue=None`, so `SupportEntry.issue` and the `f" (#{self.issue})"` suffix in `describe()` are dead alongside it. The old `assert "749" in out` was the only thing exercising them and it was re-anchored away.
- `doctor.py:76` is still `config: str = typer.Option(None, ...)`.
- `soup doctor --config missing.yaml` still exits 0 — `_check_config_support` returns on `FileNotFoundError`. Fine if deliberate; I just want the decision recorded rather than defaulted into.
- `docs/commands.md:218` still reads `soup doctor [--nccl] [--disk]`.
- The narrow-terminal item from the earlier round reproduces: green at `COLUMNS>=45`, red at `COLUMNS=35` with `the row must carry the reason, not just the name`. Worth knowing the real exposure is width, not colour — I re-ran both raw `capsys` assertions under `FORCE_COLOR=1 TERM=xterm-256color` at COLUMNS 200/100/80/60 and they stay green, because `capsys` makes Rich see a non-tty.

## For the record

14 passed on your file, 90 passed on `-k doctor`, the ANSI scanner green, ruff clean. **Zero test coverage removed** — 14 tests before and after; the 16 removed lines are re-anchors from `max_grad_norm` to `training.seed` plus two renames. I checked that line for line, and your accounting of it is accurate.

Srinivasan8888:
All four items addressed on `79bb9c2a`. Two of them were real, and one of your fixes turned out not to fix the thing it was aimed at — reporting that rather than quietly shipping the wider version.

## 3. The grounding hole — your suggested fix does not close it, and I can show that

I widened `TRAINER_MODULES` to the union you asked for and then re-ran **your** mutation, the fabricated `data.mask_history` entry:

```
24 passed
```

Still green. The union closes the *symmetric* case — a field wired inside a helper — but an unfounded `ignored` claim about a field the declared modules never mention cannot be contradicted by a check that only looks at those modules. It is true by construction.

What closes it: **a gap is relative to something.** If no module anywhere reads the field, it is not "this backend ignores it"; it is a field nothing consumes, which is #748's subject and #751's allowlist, not this table. The check now says exactly that:

```
data.mask_history: declared a gap on ('sft', 'mlx'), but no module outside
soup_cli/trainer/mlx_sft.py, soup_cli/trainer/mlx_optim.py,
soup_cli/trainer/mlx_masking.py reads it either — that is a globally
unconsumed field (#748), not a per-backend gap
```

`test_an_unfounded_gap_claim_is_caught` pins your mutation. `test_a_gap_for_a_field_another_backend_reads_is_accepted` is its control — `training.seed` is read by the transformers path and ignored by MLX, which is the shape the registry exists to record, and it must not fire there.

The union is in as well, and it was worth having on its own terms: the schedule fields appear 7-12 times in `mlx_optim.py` against twice in `mlx_sft.py`, so you are right that #734 tripped the guard by luck.

## The same trap you flagged, in my own fix

Mutating the helper modules back out of the map **survived**. My new helper test monkeypatched a synthetic map, so the real fix was untested — the identical fixture-coincidence problem you named on `use_galore`, one function further along.

Fixed by deriving the expectation instead of asserting it: `test_the_declared_modules_cover_every_helper_the_trainer_imports` parses `mlx_sft.py`'s own `soup_cli.trainer.*` imports and requires each to be declared. That fails when a helper leaves the map **and** when one is added to the trainer without being declared. A literal list would have passed both ways.

**Mutations, 5 of 5 after the fix, 4 of 5 before it:**

| mutation | killed by |
|---|---|
| drop the helper modules from the map | `test_the_declared_modules_cover_every_helper_the_trainer_imports` |
| drop the unfounded-gap cross-check | `test_an_unfounded_gap_claim_is_caught` |
| silently skip a declared module that does not exist | `test_a_declared_module_that_does_not_exist_is_caught` |
| `doctor` exits 0 on an unreadable config | `test_doctor_exits_non_zero_when_the_config_cannot_be_read` |
| read-to-warn direction disabled | `test_deleting_a_warning_makes_its_entry_fail` (all 7 params) |

## 4. Smaller items

- `test_deleting_a_warning_makes_its_entry_fail` is parametrised over every `trainer_reads=True` entry, ids by field name — a new entry is covered without editing an assertion.
- `config: str | None`.
- `soup doctor --config missing.yaml` now exits **2**, with a test covering missing *and* unparseable. Decided rather than defaulted: this leg is meant to gate CI, and a config that could not be read is not a clean bill. `doctor` without `--config` keeps its previous exit status.
- `REJECTED` / `honoured`: documented as deliberately deferred in the module docstring, with the reason — `honoured` would mean enumerating 275 fields per pair, which is a table nobody can review honestly and the opposite of this module's premise; `rejected` has no instance because `mlx_sft.py` only warns.
- `docs/commands.md:218` records `--config` and the exit code.

## 2. The fragment

Rewritten to say the scope out loud: seven entries, every one duplicating a warning `mlx_sft.py` already prints, the value being that you see it before the run rather than during it — "not that it knows anything new". It also records that five entries were deleted before merge because #734 and #750 wired those fields.

## Verification

```
tests/test_issue755_backend_support_registry.py     27 passed  (was 14)
pytest tests/                                       20817 passed, 106 skipped,
                                                    coverage 83.10% (gate 77%)
ruff check src/soup_cli/ tests/                     All checks passed!
soup doctor --config <mlx cfg>                      exit 0, run live
soup doctor --config /nope/missing.yaml             exit 2, run live
```

## One process note against myself

Mid-way through I reverted a mutation with `git checkout` on a file that still had uncommitted work, and wiped two of these fixes. Redone, and I moved to backup-copy restore with sha verification — the pattern I use in the mutation harness and should have used from the first edit. Related: the correction I posted on #752 came from the same class of mistake, a `git worktree` plus this repo's editable install not being the isolated environment I assumed.

MakazhanAlpamys:
**You are right and I was wrong, and I checked it rather than taking your word for it.**

I ran the discriminating experiment on `79bb9c2a`: kept your widened `TRAINER_MODULES` union — that is, exactly the fix I asked for — disabled *only* your new grounding rule, and re-applied my fabricated `data.mask_history` entry.

```
1 failed, 26 passed
FAILED test_an_unfounded_gap_claim_is_caught
```

The only failure is the test you wrote for the rule I had just disabled. `test_the_registry_matches_what_the_backend_trainer_actually_reads` — the drift guard itself — **passed with the fabrication present**. So the union alone does not contradict an unfounded claim, exactly as you said, and for the reason you gave: a check that reads only the declared modules cannot refute a claim about a field those modules never mention. It is true by construction. My suggested fix would have closed the symmetric case and left the one I actually demonstrated wide open.

With your rule in place, the same mutation dies:

```
2 failed, 25 passed
AssertionError: data.mask_history: declared a gap on ('sft', 'mlx'), but no module outside
soup_cli/trainer/mlx_sft.py, soup_cli/trainer/mlx_optim.py, soup_cli/trainer/mlx_masking.py
reads it either — that is a globally unconsumed field (#748), not a per-backend gap
```

And the framing is the part worth keeping: **a gap is relative to something.** If nothing anywhere reads a field, that is not "this backend ignores it", it is a field with no consumer — #748's subject, and #751's allowlist. Making the registry say which of the two it is turns a table that could assert anything into one that has to be about a real field. Your control test is what makes that safe rather than merely strict: `training.seed` is read by the transformers path and ignored by MLX, which is the shape the registry exists to record, and the rule must not fire there. It does not.

## Every item is closed — verified individually, not read off the diff

```
b) fragment            -> now says "seven"                            OK
c) grounding           -> union + the globally-unconsumed rule        OK (above)
d) parametrise         -> "use_galore" survives only in a docstring   OK
e) REJECTED/honoured   -> deliberate deferral documented at :37-39    OK
f) config: str | None  -> doctor.py:76                                OK
g) missing --config    -> exit_code = 2 (was 0)                       OK
h) docs/commands.md    -> ":218 ... [--config F]"                     OK
```

27 passed on your file (14 -> 20 tests), 92 passed on `-k doctor`, ruff clean, and **zero removed test lines** against `origin/main`.

## One thing left, and it is the last one

The narrow-terminal case still reproduces on this head:

```
COLUMNS=35 pytest tests/test_issue755_backend_support_registry.py
1 failed, 26 passed
FAILED test_doctor_config_names_the_ignored_field_and_its_reason
```

CI will not go red on this, because runners do not report a 35-column terminal — which is exactly why I want it fixed rather than left. This repository has had the neighbouring failure class (Rich mangling CLI-output assertions) turn CI red four separate times, and every time it was invisible locally first. Pin the width explicitly in the test rather than inheriting the environment's, so the assertion is about the content and not about how wide the reviewer's terminal happened to be.

Once that lands and the test matrix finishes on this head, I merge.

For the record, on process: you let your own guard block your own PR, did the rebase and entry surgery yourself after I offered to do it, and then told me my proposed fix did not work and proved it. That last one is the rarest of the three.

MakazhanAlpamys:
CI is green on `79bb9c2a` — 15 of 15, including all nine test cells. Everything else from my last review is verified closed.

**One thing left and it is the whole remaining distance:** the narrow-terminal assertion.

```
COLUMNS=35 pytest tests/test_issue755_backend_support_registry.py
1 failed, 26 passed
FAILED test_doctor_config_names_the_ignored_field_and_its_reason
  AssertionError: the row must carry the reason, not just the name
```

Rich truncates the reason column at that width, so the assertion is testing the reviewer's terminal as much as the output. Pin the width explicitly when you render — pass an explicit `Console(width=...)` or set it in the test's environment — rather than inheriting whatever the environment reports.

To be clear about why I am holding a green PR on this: CI will never go red on it, because runners do not report a 35-column terminal. That is exactly the reason. This repository has had the neighbouring class of failure — Rich mangling assertions made on CLI output — turn CI red four separate times, and every one of those was invisible locally first. A test that depends on terminal geometry is the same bet with the sign flipped: it will fail on somebody's laptop and pass everywhere else, and whoever hits it will have no idea why.

Small change, and then I merge.

Srinivasan8888:
Fixed in `812195c9`, and your `COLUMNS=35` run turned out to be two bugs rather than one — the second is in the product, not the test.

## The test read the terminal

Pinned explicitly now, the way the repo already does it in ten files with `CliRunner(env={"COLUMNS": "200"})`. `doctor` builds its `Console` at import, so this monkeypatches the module console rather than the environment.

## The product was wrong at that width, and that is the part worth having

A dotted field name is one unbreakable word, so Rich ellipsised it:

```
│ training… │ ignored │ MLX seeds │
```

The row says a setting is ignored and does not say which one. `overflow="fold"` on both text columns:

```
│ training. │ ignored │ MLX seeds │
│ seed      │         │ through   │
│           │         │ mx.random │
```

So your assertion was not only testing my terminal — it was sitting on a real defect. A user on a split pane got a report naming no setting.

**Terminal independence, proven rather than asserted:**

```
COLUMNS=35  -> 30 passed
COLUMNS=80  -> 30 passed
COLUMNS=200 -> 30 passed
```

Full suite at default width: `20820 passed, 0 failed`, coverage 83.10%, ruff clean.

**One correction against myself:** my first version of the regression test flattened the whole config section to find the folded name. That concatenates *across* columns, so `training.` and `seed` were separated by the status and reason cells — it failed at width 35 while the output was perfectly readable. It now rebuilds the first column line by line, and the comment says why so the next reader does not repeat it.

---

## Something your review turned up that is not mine, and I think you want to know

Running the **whole suite** at `COLUMNS=35` to check my fix, I got:

```
216 failed, 20604 passed        (COLUMNS=35)
  0 failed, 20820 passed        (default width)
```

**216 tests across 89 files**, and they are pre-existing — verified against unmodified `main`, not my branch:

```
$ COLUMNS=35 pytest tests/test_doctor.py     (origin/main, no changes)
FAILED test_doctor_shows_dependencies
FAILED test_doctor_checks_torch
FAILED test_doctor_checks_pydantic
FAILED test_doctor_checks_optional_deps
FAILED test_doctor_missing_dep
5 failed, 15 passed

$ pytest tests/test_doctor.py                (same tree, default width)
20 passed
```

Heaviest: `test_recipes_v031.py` (28), `test_errors.py` (9), `test_adapters.py` (7). Almost all are `--help` or table-output assertions.

This is the same bet you are describing, at scale: it will never go red on a runner, and it fails on the laptop of anyone working in a split pane or an SSH session on a phone. Ten test files already pin `COLUMNS`; 89 do not.

I am **not** proposing to fix 216 assertions, and I am not filing anything unilaterally — you have six open issues from me already. The cheap, ratchet-shaped option is a `conftest.py` fixture pinning `COLUMNS` for the suite, which makes every one of them deterministic without touching them, plus a guard so a new width-dependent assertion cannot be added silently. That is a different PR and your call whether it is worth one. Happy to file it with these measurements, or to drop it.

MakazhanAlpamys:
Re-reviewed at `812195c9`. Eight of the nine findings from round one are closed, including the one I said was the blocker, and **the width chase turned out to be a real product defect rather than a test nit** — at `COLUMNS=35` the report used to name no setting at all, and with `overflow="fold"` the name now folds to `training.` / `seed` and survives. That is worth more than the test it came from.

Two things stand between this and a merge, and the first is small.

## 1. The width fix traded terminal geometry for terminal colour (blocking, ~2 lines)

`test_the_setting_name_survives_at_any_pinned_width` splits raw `capsys` output on `│` with no ANSI strip. Rich reads `FORCE_COLOR` when the `Console()` is constructed, so the explicitly-built console now emits colour even though `capsys` is not a tty:

```
FORCE_COLOR=1 TERM=xterm-256color pytest tests/test_issue755_backend_support_registry.py
1 failed, 29 passed

AssertionError: name lost at width=35; first column read as
  '\x1b[1m \x1b[0m\x1b[1msetting  \x1b[0m...\x1b[1mtraining.\x1b[0m...\x1b[1mseed     \x1b[0m'
```

To be fair to you about how much this matters: unmodified `main` fails two of twenty in `tests/test_doctor.py` the same way, and CI runners do not set `FORCE_COLOR`, so this will not go red on us. But it is exactly the "passes everywhere except on somebody's laptop" bet, and the escapes sit *between* the folded halves, so a whitespace collapse alone will not rescue it — strip ANSI before the `│` split. `tests/test_auto_tuning.py` has the canonical helper.

## 2. Round-one finding 7 is half-closed (your call, but I would take the fix)

`load_config` raises `SystemExit(1)` — a `BaseException` — for a **schema-invalid** config, so it walks straight past `except Exception`:

```
soup doctor --config invalid.yaml
Config validation error: training -> epochs: ...
exit=1
```

That is exit 1, not the exit 2 you documented in `docs/commands.md:218`, and doctor's own message never prints. Your parametrize covers the missing-file and unparseable-YAML shapes and I confirmed both live at exit 2; this is the third shape. Non-zero either way, so a CI gate still works — it is a contract inconsistency rather than a hole.

## Mutations I ran

| Mutation | Result |
|---|---|
| `check_config()` returns `[]` | killed — 6 failed |
| drop `("sft", "mlx")` from `REGISTRY` | killed — collection error, `KeyError: ('sft','mlx')` |
| add an unclassified field to `TrainingConfig` | survived — 30 passed |

The third is not a finding and I want to say so explicitly, because it would be easy to file it as one: you documented the refusal at `backend_support.py:36-48`, a `honoured` status would mean enumerating 275 fields into a table nobody can review honestly, and a gap table lists gaps rather than every field. I agree with the call. It is still the design's real limit and it should stay written down where you put it.

Also verified: **zero removed test or assert lines**; the branch is six commits behind `main` and none of them touch `trainer/` or `config/`, so the round-one staleness cannot have recurred. 30 passed in the PR's file, 92 under `-k doctor`, 43 across the startup-lightness and ANSI ratchets, 18 across the changelog guards, `ruff` clean, and `doctor --config` renders correctly at both 200 and 35 columns.

## What I want to name

You let your own guard block your own pull request, declined my offer to do the surgery, and reported the twelve-to-seven shrink in full rather than quietly adjusting the number. Then you **disproved my proposed fix with a discriminating experiment** — the union of the declared modules alone cannot refute a claim about a field those modules never mention — and shipped a better rule, because a gap is relative to something. You caught the fixture-coincidence trap *in your own fix* and derived the module list from `mlx_sft.py`'s actual imports instead of asserting it. And you offered the 216 pre-existing width-dependent tests as a measurement against unmodified `main` rather than filing them as someone else's problem.

Push item 1 and I will merge on green.

Srinivasan8888:
Both fixed in `d627ccc`.

## 1. ANSI before the split

You were right that whitespace collapsing could not rescue it — the escapes sit between the folded halves. Stripped first, and I used **conftest's shared `strip_ansi`** rather than adding a 39th private copy of the regex, since `conftest.py:10-12` says new code should import that one. One wrinkle worth recording: it is `from tests.conftest import strip_ansi`, not `from conftest import` — `tests/` is a package, and my first attempt failed collection with `ModuleNotFoundError: No module named 'conftest'`.

Verified across the matrix, because the whole point is that neither dimension may decide the result:

```
no-colour    COLUMNS=35  -> 31 passed
no-colour    COLUMNS=200 -> 31 passed
FORCE_COLOR  COLUMNS=35  -> 31 passed
FORCE_COLOR  COLUMNS=200 -> 31 passed
```

Your exact repro (`FORCE_COLOR=1 TERM=xterm-256color`) passes.

## 2. The third unreadable shape — I took the fix

`load_config` raises `SystemExit(1)`, a `BaseException`, so it walked straight past `except Exception`: the exit code contradicted the 2 in `docs/commands.md:218` and doctor's own message never printed. Caught explicitly now and re-raised as 2, so all three shapes agree:

```
missing.yaml         -> exit 2
broken.yaml          -> exit 2   (unparseable)
invalid.yaml         -> exit 2   (schema-invalid, was 1)
```

The parametrize covers the third shape and asserts **exactly 2** rather than non-zero, which is what makes the contract testable rather than merely satisfied.

## On the survived mutation

Agreed, and thank you for saying it out loud rather than leaving it as an open question — an unclassified new field surviving is the design's limit, not a defect. `backend_support.py:36-48` is where it is written down and I have left it there.

## Verification

```
tests/test_issue755_backend_support_registry.py   31 passed (4 colour x width combos)
pytest tests/                                     20821 passed, 0 failed
ruff check src/soup_cli/ tests/                   All checks passed!
soup doctor --config invalid.yaml                 exit 2, run live
```

One note: `ruff` flagged the new import block (`I001`) after I added the conftest import, and I fixed it with `--fix` rather than by hand — the resulting order is what ruff wants, not what I chose.

MakazhanAlpamys:
Merged as 344d938f — thank you, @Srinivasan8888. Your CONTRIBUTORS.md entry has a new line.

Three rounds, and each one made the feature more honest rather than just greener:

- **You argued for a declared table with measurements, not assertion.** Import-graph reachability caught none of five known MLX gaps, trainer-module reads invented gaps for fields that live in helpers, and `--dry-run` exits before a trainer exists. That is the evidence that makes "declared, not inferred" a decision rather than a shortcut.
- **Your own guard shrank your table from twelve entries to seven** once #734 and #750 wired five of them, and you said so openly. A guard that removes its author's work is proof that it works.
- **The `COLUMNS=35` note turned up a real defect.** At that width the report named no setting at all; folding fixes it for users, not only for the test.
- **This round you reused the shared `strip_ansi`** instead of adding a fourth copy, and tightened the exit-code test from "non-zero" to "exactly 2", so the documented contract is now actually tested.

What I verified at `d627ccc3`, the head that was merged:

- The PR's file, `test_doctor.py`, the startup-lightness and ANSI ratchets and the fragment guard: **111 passed**. The width test passes plain, under `FORCE_COLOR=1 TERM=xterm-256color`, and with `COLUMNS=35` on top.
- `soup doctor --config` on two schema-invalid configs: exit 2, pydantic field errors, no traceback.
- Mutations, all killed: `check_config` returning `[]` (6 failed); dropping the `model_fields_set` filter; deleting the `except SystemExit` branch; removing the ANSI strip under `FORCE_COLOR=1`.
- All 13 required checks green.

One small thing for a follow-up if you want it, not a blocker: `All checks passed! Your environment is ready.` still prints immediately before a config error that exits 2. It refers to the environment section, but next to a failing exit code it reads as a contradiction.

PR #757, opened 2026-09-09
test(cli): fail when a typer command is called with a parameter left unfilled

dchaudhari7177:
Refs #752. The call-site fix landed in #754; this PR is now the package-wide guard only.

## What it guards

Called as a plain Python function rather than through typer, a command's unfilled parameters keep their literal `typer.Option(...)`/`typer.Argument(...)` defaults. Those are truthy `OptionInfo` objects, which is how `soup eval auto` crashed after saving its results in #752. #754 pins the two known call sites. This test catches the **next** one.

`tests/test_typer_defaults_do_not_leak.py` walks `src/soup_cli` with `ast`. It collects every `@app.command()`/`@app.callback()` function and the parameters typer would fill, resolves each bare-name call through the calling module's own `from X import name` (or a command in the same module) to the command it really names, and fails on any call that leaves a typer parameter unfilled.

**Positional arguments are matched to parameters by declaration order**, not counted:

```
custom(tasks_file, str(output_dir))
-> caller.py:2 calls custom() without ['attach_to_registry', 'output', 'run_id']
```

Only the genuinely unfilled parameters are named. A call whose positional count merely *reaches* the number of typer parameters (for example, when leading plain parameters like `ctx` absorb it) is no longer skipped.

## Tests

| test | what it proves |
|---|---|
| `test_the_package_declares_typer_commands_at_all` | the collector finds >50 commands, including `custom` |
| `test_the_guard_reports_a_synthetic_leak` | the **reporter** can still report |
| `test_positional_arguments_fill_parameters_by_position` | positional, keyword and mixed fills are equivalent |
| `test_leading_plain_parameters_do_not_count_as_filled_typer_ones` | `gadget(ctx, 't')` still reports `force` |
| `test_calls_it_cannot_resolve_are_not_reported` | `*args`/`**kwargs` calls are not guessed at |
| `test_a_partial_positional_call_names_only_the_unfilled_parameters` | the message above, against the real `custom` |
| `test_no_typer_command_is_called_with_parameters_left_unfilled` | the package today |

Mutations on current `main`:
- `attach_to_registry=None` removed from `commands/eval.py:703`: fails, naming the file, line and parameter. Same for `monitoring/callback.py:525`.
- `_leaking_calls` returning `[]`: 3 failed.
- positional args ignored (keyword-only subtraction): 3 failed.
- the old `len(args) >= len(required)` skip restored: 1 failed.

With `tests/test_issue752_cli_commands_called_as_functions.py`: 18 passed. `ruff@0.9.4 check` clean. Adds `changelog.d/0.75.0/757.added.md`.


MakazhanAlpamys:
@dchaudhari7177 — **there is a second pull request against #752 and it landed before yours. You are hearing it from me within the hour rather than from a merge notification, and part of the reason you could not have seen it cleanly is mine.**

## The timeline, checked rather than remembered

```
2026-09-09T06:18:55Z  @Srinivasan8888 files #752
2026-09-09T06:43:34Z  @Srinivasan8888 opens #754 against it   <- 25 minutes later
2026-09-09T08:44:30Z  you open #757                            <- 2h 01m after that
```

So the reporter filed the fix themselves, almost immediately.

**What you could have seen, and what you could not.** #754 appears in #752's timeline as a cross-reference from 06:43, so a linked pull request was visible on the issue page when you started. That is a real signal and I am not going to pretend it was invisible. What was **not** there is the thing this project says is authoritative: `CONTRIBUTING.md` states the claim is a **comment**, and the thread is the record. #752 has zero comments and carried no `claimed` label until I added one just now. I said on #721 two days ago that I would label and comment *before* replying from now on, and on this issue I did neither — it was filed and answered inside three hours while I was away from the repo.

## What happens next

**I am not deciding this on arrival order.** Both are being measured side by side right now, like for like, and the question I actually care about is durability rather than either diff being smaller:

- whether each fix addresses the root cause or only the two parameters that happened to break today;
- whether `commands/eval.py` and `monitoring/callback.py` are fixed consistently in each;
- **whether adding a NEW option to that command later would make the test suite fail** — a guard that catches the next occurrence of this bug class beats a hand-maintained list of today's parameters, and that is the axis most likely to decide it;
- the same mutations applied to both, plus a vacuity sweep running each PR's new tests against unmodified `main`.

I will post the numbers in both threads before the decision, not a verdict with the reasoning left out.

**Whatever the outcome, your work does not disappear.** If the other one lands, anything yours does better goes in with your name on it and I will say in the thread which parts those were.

One thing to fix regardless of the outcome: there is no `changelog.d/0.74.0/757.fixed.md`. This repo has not edited `CHANGELOG.md` directly since #487; a fragment named after the PR is the convention, and a *missing* fragment is invisible to CI, so nothing will tell you but me.

Nothing else is asked of you right now.

MakazhanAlpamys:
@dchaudhari7177 — **the comparison is done. First: your CI was never run.** The workflow was sitting at `action_required` awaiting maintainer approval — a permissions artifact of a fork PR, not a quality signal — so this PR has been judged all day against zero check runs through no fault of yours. I approved it just now; 14 checks are queued.

## The decision

**I am taking #754.** Below is why, including the part where the criterion I told you would decide it did not.

## The production fix is byte-identical

Stripping comments, both PRs add exactly the same four lines in the same two files:

```
attach_to_registry=None,
output=None,          # commands/eval.py
attach_to_registry=None,
output=None,          # monitoring/callback.py
```

Both pass all five of `custom()`'s parameters at both call sites. Neither leaves anything to leak, neither hardcodes a subset. On the product change there is nothing to choose between you.

## The durability test — a tie, and I said it would decide

I told you this morning that the axis I expected to settle it was whether adding a **new** option to `custom()` later would make the suite fail. I added a plausible `--foo` to `custom()` in both branches without updating either fix:

```
#754   3 failed, 8 passed
#757   1 failed, 1 passed
```

**Both notice.** Neither is a hand-maintained list, both are repo-wide AST guards, and your guard is as general as theirs. That criterion did not separate you, and switching to different grounds without saying so would be dishonest, so I am saying it.

## What did decide it

**Your guard can be silently disabled and nothing notices.** I put `return []` at the top of `_leaking_calls`, making it report nothing ever, and ran your file:

```
baseline          2 passed
_leaking_calls -> []   2 passed     <- unchanged
```

Your anti-vacuity test checks that the **collector** finds commands (`test_the_package_declares_typer_commands_at_all`), but nothing checks that the **reporter** can still report. #754 has `test_the_guard_catches_a_synthetic_leak`, which fails under the same mutation. This project has that lesson written down from its ANSI scanner: the load-bearing test in a repo-wide guard is the one proving the scanner can still fail, because a guard that finds zero offenders today looks identical whether it works or is broken.

Two smaller items, both real and both trivially fixable: `ruff check` fails on `F401 pytest imported but unused`, which would have turned the `lint` job red once CI ran; and there is no `changelog.d/0.74.0/757.fixed.md`, so the fix would have shipped with no changelog entry and no attribution.

There is also no behavioural test — nothing runs `soup eval auto` and asserts it completes. If your single AST guard ever breaks or is muted, nothing is left testing the actual crash. #754 has two.

## What yours does better, and I want it

Your guard **catches a leak passed positionally** and #754's deliberately does not — theirs bails out with `if call.args: continue` and pins that as intentional. On purpose-built fixtures:

```
correct call  widget("x", fmt="json")   #754 []  (correct)   #757 FALSE POSITIVE on 'name'
real leak     widget("x")               #754 []  (misses)    #757 reports it
```

So it is a genuine trade, not a defect: yours is more sensitive and currently over-reports because it compares the **count** of positional arguments against the count of typer parameters instead of resolving declaration order. No call site in this repo passes a command positionally today, so it is theoretical for both — but the leak class is real and yours is the only one that sees it.

**If you want it, resolving positionally-passed arguments against the declared parameter order — so it catches that leak without the false positive — is yours to land as a follow-up, credited to you.** That is a strictly better guard than either PR has now, and it is your finding. Entirely your call, and "no" costs you nothing.

## One more thing

Your write-up describes hitting the same false-positive wall #754 hit — `main()`, `version()`, `check()` — and solving callee resolution through `from X import name` rather than by bare name. You got there independently two hours later with no sight of their work. That is the hard part of this guard and you both found it.

You merged #726 today and this is your second PR of the day against a repo you started contributing to three days ago. Nothing about this decision is about you being new.

MakazhanAlpamys:
@dchaudhari7177 — your CI has now run, and the one failure is the `lint` job on the `F401` I flagged:

```
tests/test_typer_defaults_do_not_leak.py:23:8  `pytest` imported but unused
```

One line. Everything else that has reported so far is green.

I am telling you rather than leaving it in a check tab because it was never your CI's fault that this took a day — the run was blocked at `action_required` until I approved it this morning, so this is the first time the job has actually executed on your branch.

The decision on #752 is unchanged and is in my earlier comment; this does not alter it. But if you take up the positional-argument resolution as a follow-up, that file goes with it, so the import is worth removing either way.

MakazhanAlpamys:
@dchaudhari7177 — **I have measured your scanner rather than read it, and it works. The path to merging this is now short, but it is a different path than the one this PR is on, so read the first section before you push anything.**

## The source fix in this PR is already on `main`

#754 merged as `1490c85`, and the two call sites it changed are byte-for-byte the change you proposed here — same two keywords, same two files:

```
$ sed -n '703,710p' src/soup_cli/commands/eval.py
            custom(
                tasks=tasks_file,
                model=str(output_dir),
                run_id=None,
                attach_to_registry=None,
                output=None,
            )
$ sed -n '524,531p' src/soup_cli/monitoring/callback.py
                custom(
                    tasks=custom_tasks,
                    model=self.output_dir,
                    run_id=self.run_id,
                    attach_to_registry=None,
                    output=None,
                )
```

So `src/soup_cli/commands/eval.py` and `src/soup_cli/monitoring/callback.py` should come **out** of this PR. Rebased onto current `main` those two hunks are no-ops.

**What is left is the file I actually want**, and `main` does not have it: `tests/test_typer_defaults_do_not_leak.py`. #754 shipped `tests/test_issue752_cli_commands_called_as_functions.py`, which pins the two known sites. Yours walks the whole package and fails on the next one. Those are not the same test and the second is the one that keeps paying.

## I ran your scanner against the fixed `main` and mutated both sites

Worktree on `main` at `ccd906c`, your test file dropped in, `PYTHONPATH` verified to point at that worktree, every mutation applied in binary mode with an occurrence-count assert and sha256 before/after (this repo is CRLF and a text-mode replace silently no-ops).

Baseline — green on fixed code, so it is not just always-red:

```
2 passed in 10.06s
```

Mutation 1, `attach_to_registry=None, output=None` removed from `commands/eval.py`:

```
sha_before cc7c7fa268285517  ->  sha_after 5a3501f59f31bda7
FAILED test_no_typer_command_is_called_with_parameters_left_unfilled
E   src\soup_cli\commands\eval.py:703 calls custom() without ['attach_to_registry', 'output']
```

Mutation 2, same removal in `monitoring/callback.py`:

```
E   src\soup_cli\monitoring\callback.py:525 calls custom() without ['attach_to_registry', 'output']
```

Both killed, each naming the right file, line and parameters. **That is a real guard, not a decorative one**, and `test_the_package_declares_typer_commands_at_all` is the right instinct — a whole-package AST walk that silently finds nothing is the standard way this class of test rots.

## One finding, and it is in the positional handling

`_leaking_calls` has two branches on `node.args` and I could make both misbehave:

```python
if len(node.args) >= len(required):  # filled positionally
    continue
...
missing = required - {kw.arg for kw in node.keywords}
```

I rewrote the `eval.py` call as `custom(tasks_file, str(output_dir))` — a genuine #752 bug, two positional args, `output` and `attach_to_registry` unfilled. Your test **does** fail, so no false green. But it reports:

```
E   calls custom() without ['attach_to_registry', 'model', 'output', 'run_id', 'tasks']
```

`tasks` and `model` were passed. They were just passed positionally, and the `missing` set only ever subtracts keywords. A developer hitting that message goes looking for two parameters that are already there.

The `>=` skip is the same assumption from the other side: it treats "enough positional args" as "the typer-defaulted ones are filled", which only holds when they are the leading parameters. For `custom` they are, so nothing on `main` is missed today — but the guard's whole purpose is the case nobody has looked at yet.

Both are fixable by walking positional args against the parameter list by index instead of counting them. That is a contained change to one function.

## What I am asking for

1. Rebase onto `main` and drop the two source hunks. This becomes a test-only PR.
2. Remove the unused `import pytest` — `lint` is the one red check and this is the whole of it.
3. Resolve positional args by index so the message names only genuinely-unfilled parameters, and add a test for the partial-positional call above so the behaviour is pinned rather than assumed.
4. `changelog.d/0.74.0/757.added.md`, one line, crediting you.

Retitle it to what it now is — the scanner, not the fix. Then it merges on its own terms and the credit is for the part that #754 did not do.

I said in my earlier comment that the positional resolution could be a follow-up. Having measured it, I would rather have it in this PR: it is one function, and a guard that misreports which parameter is missing is the kind of thing that gets ignored the third time it fires.

dchaudhari7177:
@MakazhanAlpamys — all four items are in `be39648`. It's one commit on current `main`, retitled, and the body is rewritten to match.

1. **Rebased, source hunks dropped.** The diff is now `tests/test_typer_defaults_do_not_leak.py` plus the fragment.
2. **Unused `import pytest` removed.** `ruff@0.9.4 check` is clean.
3. **Positional arguments are resolved by index.** Each command records its parameters in declaration order, and a call's positional arguments fill `positional[:len(args)]`. Your partial-positional call now reports only what is unfilled:
   ```
   custom(tasks_file, str(output_dir))
   -> caller.py:2 calls custom() without ['attach_to_registry', 'output', 'run_id']
   ```
   The `>=` skip is gone. It was wrong in the direction you described: a command with leading plain parameters (`gadget(ctx, target, force=typer.Option(...))`) called as `gadget(ctx, 't')` has enough positional arguments to be skipped while `force` is unfilled. There is a synthetic test for exactly that.
4. **`changelog.d/0.75.0/757.added.md`.** It is under 0.75.0 because 0.74.0 has released, and it passes `assemble_changelog.py` on a temp copy.

I also took the lesson from your #754 comparison. The scanner now takes its sources as a list, so a synthetic command module can drive it, and `test_the_guard_reports_a_synthetic_leak` fails if the reporter is muted. Your `return []` mutation now fails **3** tests. Restoring the keyword-only `missing` subtraction fails 3, and restoring the `>=` skip fails 1. Removing `attach_to_registry=None` from either real call site fails the package test, naming the file, line and parameter. Together with #754's `test_issue752_cli_commands_called_as_functions.py`: 18 passed.

MakazhanAlpamys:
**All four items verified by execution, and the guard genuinely generalises. One blocking finding, on the axis I said I would examine.**

Head `be39648`, worktree with `PYTHONPATH` pinned inside it, every mutation applied with an occurrence assert and an `ast.parse` validity check on the result — because an injection that produces a syntax error gets silently skipped by an AST scanner and looks exactly like a surviving mutation. (That is not hypothetical: I made that mistake on a different PR today and nearly published a wrong finding from it.)

**The four items.** Rebased and test-only — 2 files, 259 insertions, **zero deletions, zero source hunks**. `ruff` clean across `src/ scripts/ tests/ benchmarks/`. Positional resolution by index confirmed. And the fragment: **you were right and I was wrong.** I asked for `changelog.d/0.74.0/`; you filed `0.75.0/` because the release landed in between, and following my instruction would have failed `test_issue487_changelog_fragments.py` in every cell.

**The `return []` hole is closed** — 4 failed, not the 3 you claimed. And it is not a local guard: I injected a leak you have never seen, `lock_cmd(design)` called from `commands/doctor.py` with `_eval_v0550`'s command resolved across a **third** module, and it reported

```
src\soup_cli\commands\doctor.py:22 calls lock_cmd() without ['attach_to_registry', 'output']
```

positional argument correctly absorbed, only the two genuinely unfilled named. That is the generality proof. Five mutations killed in total; 18 passed alongside #754's file; 35 on a local merge with current `main`; and I ran it against the in-flight `commands/eval.py` from #917 and #921 — neither goes red.

**Blocking: the anti-vacuity test cannot see a coverage regression.** `assert len(commands) > 50` against a package that declares **145**, so 65% of them could vanish beneath it. Worse, it structurally *cannot* fire on the failure that matters: **143 of those 145 live under `commands/`**, and so does your anchor. Two mutations got past all 7 tests:

```
SRC.rglob("*.py") -> (SRC/"commands").rglob("*.py")     7 passed   sources 504 -> 89, commands 145 -> 143
  + output=None deleted from monitoring/callback.py     7 passed   a real #752 leak, unreported

ast.walk(source.tree) -> source.tree.body               7 passed
  + the same leak in monitoring/callback.py             7 passed
```

The second is the sharp one. `callback.py` imports `custom` **inside a function**, which is exactly the shape of one of the two real #752 sites, and your synthetic fixture only ever exercises a module-level import. So a plausible tidy-up of the resolver blinds the guard at half the original bug, with a green suite. This is the same failure mode as the ANSI-scanner floor in `CLAUDE.md`: a guard finding zero offenders today looks identical whether it works or is broken.

Two tests close both. I measured that each discriminates before proposing it — on clean code the nested-import case reports the leak and the walk sees 504 modules; under the respective mutation it returns `[]` / 89:

```python
def test_the_walk_reaches_the_whole_package_not_just_commands():
    modules = {s.module for s in _package_sources()}
    assert "soup_cli.monitoring.callback" in modules   # a real #752 call site
    assert len(modules) > 400                          # measured 504

def test_a_command_imported_inside_a_function_is_still_resolved():
    sources = [
        _Source("pkg/cli.py", "pkg.cli", ast.parse(_SYNTHETIC_COMMAND)),
        _Source("pkg/nested.py", "pkg.nested", ast.parse(
            "def run():\n    from pkg.cli import widget\n    widget('x')\n")),
    ]
    assert _leaking_calls(sources, _collect_commands(sources)) == [
        "pkg/nested.py:3 calls widget() without ['fmt', 'verbose']"
    ]
```

That is the whole of it. This is the file I wanted, and the part you added over #754 is the part that keeps paying.

**Three non-blocking notes.** `_package_sources` swallows `SyntaxError` per file, so an unparseable module leaves the guard's view silently — the `> 400` assertion covers that too. It is called three times (~8.2 s per cell; `lru_cache` halves it). And `_is_command` only matches an `Attribute` decorator, so `@command()` by bare name is invisible — no such usage today, worth a comment rather than code.

MakazhanAlpamys:
Reviewed at `be39648e` — **14/14 checks SUCCESS on exactly that sha**, so the green mark is current, not stale.

**This is a real ratchet, and I checked the only thing that actually matters for one: whether it catches the historical defect.** It does. I reverted #754's fix at both real call sites and ran your guard against the mutated tree:

```
src\soup_cli\commands\eval.py:703       calls custom() without ['attach_to_registry', 'output']
src\soup_cli\monitoring\callback.py:525 calls custom() without ['attach_to_registry', 'output']
```

It names the file, the line, the callee and the exact leaked parameters — and it finds **both** sites, not only the one in the ticket.

**And the value over the guard already on `main` is demonstrated rather than asserted.** I dropped a brand-new command module into `src/` with a positionally-passed leak:

| guard | result |
|---|---|
| #757 (this PR) | `_probe_caller.py:5 calls brandnew() without ['dry_run']` — **caught** |
| #754 (on main) | **1 passed — missed entirely** |

That is precisely the follow-up #754's merge commit pre-credited to you, and it is delivered.

**Scope, measured rather than claimed:** 504 `.py` files parsed under `src/soup_cli`, 145 typer commands collected, 679 typer parameters tracked, 36 declaring modules. The walk is pure `ast.parse` — no `import soup_cli`, no hand-written module list, no hardcoded parameter names (neither `output` nor `attach_to_registry` appears in the walker). Newly added commands are picked up automatically, which I proved above. This is not the hand-written-list shape that hid nine call sites in #354.

**Can-it-fail, three mutations:** blinding `_leaking_calls` kills 4 tests, reverting positional resolution to #754's `if call.args: return None` kills 4, blinding `_is_command` kills 5. So the distinguishing feature of this PR is itself pinned and cannot be silently reverted.

**I also probed for false negatives across 17 call shapes.** Yours is a strict superset of the merged guard on every one, with zero false positives on the six shapes that must stay silent (fully-filled calls, `**kwargs`/`*args`, a method sharing a command's name). The four it misses — relative imports, module-attribute calls, `functools.partial`, dict-of-handlers dispatch — were declared out of scope by the issue author, and `src/` currently has **zero** relative imports, so that one is latent rather than live.

---

**One thing I want before merge, and it is two lines.**

`tests/test_typer_defaults_do_not_leak.py:150-151` — the `isinstance(node.func, ast.Name)` restriction is what stops the guard flagging `obj.method(...)`. I mutated the walker to also flag `ast.Attribute` calls, and **only one test failed** — and only because `tracker.list_runs(...)` (`commands/runs.py:71,352`) happens to exist in the tree today. **None of your six synthetic tests caught it.** If that call site is ever refactored away, the false-positive protection becomes unguarded.

That is issue #752's own acceptance criterion 3 — *"`model.train()` does not fail — this needs its own test, since it is what a naive version gets wrong"* — and it is the shape this repo has been burned by: a scanner that fired on 31 correct files had to be deleted rather than fixed. #754 pins it deliberately in `tmp_path`. Please add:

```python
assert _scan_synthetic("obj.widget('x')") == []
```

plus a plain-function-sharing-a-name case, so the protection holds by construction rather than by accident of the current tree.

---

**Two notes, no action needed:**

- The file does the full 504-file walk three times (`:200`, `:236`, `:253`). Measured: 1.84s + 0.52s + 0.53s per pass, 14.25s total. An `lru_cache` on `_package_sources`/`_collect_commands` would take it to about 5s — **~54s saved per push across the nine cells**. Worth doing if you are touching the file anyway.
- There are now two independent walkers guarding one bug class (`test_issue752_cli_commands_called_as_functions.py`, 471 lines, and this one, 259). They can drift. Against folding them: your positional half is the stronger of the two, and #754's merge commit already announced this as a separate follow-up. I am fine merging as-is and consolidating later — recording it so the decision is explicit rather than forgotten.

Changelog fragment is correct: exactly one, right directory, credit `(#752 by @dchaudhari7177 in #757)`, and `.added` matches precedent — #457's recipe-count ratchet, also a test-only repo-wide guard, is filed under Added.

Your PR body says three mutations where I measured four. The guard is **stronger** than you claimed, which is the direction I never have to worry about.

dchaudhari7177:
Both review rounds addressed in `5f25bd0` — test-only, one file.

**Blocking: the walk's reach.** Added `test_the_walk_reaches_the_whole_package_not_just_commands` (asserts `soup_cli.monitoring.callback` is parsed and more than 400 modules are) and your nested-import case, `test_a_command_imported_inside_a_function_is_still_resolved`. I also raised the command-count floor from `> 50` to `> 120` (145 today), though the module assertion is the one that actually sees a narrowed walk.

**Method calls.** `test_a_method_sharing_a_command_name_is_not_reported` pins `obj.widget('x')` and `pkg.cli.widget('x')` as silent, and `test_a_local_function_sharing_a_command_name_is_not_reported` covers a plain `def widget(name)` in a module that never imports the command. Both are synthetic, so they no longer depend on `tracker.list_runs(...)` staying in the tree.

Each mutation, applied on its own to a clean copy of the file:

| mutation | result |
| --- | --- |
| walk `(SRC / "commands").rglob` | 1 failed — `test_the_walk_reaches_the_whole_package_not_just_commands` |
| `_imported_commands` reads `source.tree.body` only | 1 failed — `test_a_command_imported_inside_a_function_is_still_resolved` |
| `_leaking_calls` also resolves `ast.Attribute` calls | 2 failed — the new method test, plus the repo-wide scan |

**Non-blocking, folded in:** `_package_sources` and a new `_package_commands` are `lru_cache`d, so the file does the walk once: 11 passed in ~4 s locally, down from ~15 s. The `_is_command` docstring now says a bare `@command()` is not matched and that nothing uses that form today. The `SyntaxError` skip stays, since the `> 400` assertion catches a mass drop-out.

Thanks for the careful mutation work, especially the nested-import one. My fixture only ever imported at module level, so I would not have found it.

MakazhanAlpamys:
Verified `5f25bd03`. **Merge-ready** — the blocking finding is closed, and both mutations that walked past the previous head now fail.

| mutation | last time | now |
|---|---|---|
| `SRC.rglob("*.py")` narrowed to `commands/`, plus a real leak outside it | 7 passed | 1 failed — `test_the_walk_reaches_the_whole_package_not_just_commands` |
| `ast.walk(source.tree)` to `source.tree.body` in `_imported_commands`, plus the same leak | 7 passed | 1 failed — `test_a_command_imported_inside_a_function_is_still_resolved` |

The measured headroom behind the two new assertions is **504 modules / 145 commands** against thresholds of 400 and 120, so both discriminate rather than sitting on a boundary.

For a test-only PR the real question is whether the guard works on its subject, so I ran it in both directions:

- **It catches a genuine leak.** Deleting `output=None` from `monitoring/callback.py:525` — a real #752 nested-import call site — with the guard unmutated gives `src\soup_cli\monitoring\callback.py:525 calls custom() without ['output']`. File, line, and only the parameter that is genuinely unfilled.
- **It does not false-positive.** I appended two *correct* new call sites to `commands/doctor.py`, one fully positional and one with aliased keywords, both behind nested imports: 11 passed, zero findings.
- **The reporter cannot be silenced.** `return []` at the top of `_leaking_calls` gives 5 failed.

`tests/test_typer_defaults_do_not_leak.py` plus `tests/test_issue752_*` gives **22 passed**, not the 18 in your write-up — you added two more on top of the two I asked for. `test_a_method_sharing_a_command_name_is_not_reported` and `test_a_local_function_sharing_a_command_name_is_not_reported` pin #752's third criterion, which nobody asked you to do.

Diff is 2 files, **+320 / −0** — the source hunks are gone as requested and `tests/` has no removed lines at all. Fragment at `changelog.d/0.75.0/757.added.md` with correct credit. No ANSI exposure: it is a pure AST walk with no assertion on CLI output.

Two blind spots for the record, neither a blocker and both the right trade today:

- `_imported_commands` requires `node.level == 0`, so a relative import would not resolve. I grepped — the package contains **zero** relative imports, so it is latent, not live.
- `_leaking_calls` inspects only `ast.Name` callees, so `eval_mod.custom(x)` is invisible. That is the deliberate trade that keeps `model.train()` out of the report, and it is now pinned by your own `test_a_method_sharing_a_command_name_is_not_reported`.

Only CI is left; the Actions queue is at ~88 runs. I will merge as soon as it is green.

MakazhanAlpamys:
Re-verified against current `main` (`b0a63382`). **Unchanged and still merge-ready** — and this one deserved the extra attention, because it is the PR most exposed to `main` moving.

Your guard is a whole-package AST scanner over `src/soup_cli`, so freshly merged **production** code is what can turn it red, not your own diff. Three of today's merges touch files inside its walk: `cli.py`, `commands/plugins.py`, `plugins/__init__.py`, `plugin_callback.py`, `mcp_server/registry.py`. Re-run against the new tip: **still zero offenders**.

And the live injection still fires — reintroducing the real #752 defect at `monitoring/callback.py:527` gives:

```
AssertionError: ...callback.py:527 calls custom() without ['attach_to_registry', 'output']
```

Note it names **two** missing parameters now rather than the one it named this morning, because `main` has since changed that call site. So the guard is reading real repository code as it moves, not only its own synthetic fixture — which is exactly the property that makes a scanner worth having.

11 passed on both trees, ratchets 138 passed, fragment intact. Nothing needed from you.

MakazhanAlpamys:
Merged as `f707b74d`. Thank you — and this one took longer than it should have, so the reason first.

**Why the wait was not about your PR.** GitHub Actions on this repo produced no terminal outcome at all between 2026-09-13T18:30Z and today 15:08Z: 95+ runs queued, two stuck `in_progress` for over 28 hours. Your checks were not failing, they were not executing. The queue drained on its own and this merged at 15:54Z. Combined with the earlier stretch where your fork PR sat at `action_required` awaiting my approval, this PR has spent most of its life waiting on the repository rather than on review — worth saying plainly, because from your side it is indistinguishable from being ignored.

**What I merged on.** All 13 required contexts green on `5f25bd03` itself, confirmed per-SHA rather than from a cached summary. Two mutations, both killed:

| mutation | result |
|---|---|
| `SRC.rglob("*.py")` narrowed to `commands/`, with a real leak outside it | 1 failed — `test_the_walk_reaches_the_whole_package_not_just_commands` |
| `ast.walk(source.tree)` reduced to `source.tree.body`, same leak | 1 failed — `test_a_command_imported_inside_a_function_is_still_resolved` |

Both are mutations that walked past an earlier head of this branch and now do not, which is the useful kind of second round.

**The part that made me confident merging a test-only PR.** For a guard, "the tests pass" is nearly content-free — what matters is whether it works on its subject, so I ran it in both directions on live repository code rather than on its fixtures:

- **It catches a genuine leak.** Deleting `output=None` from a real `#752` call site in `monitoring/callback.py` gives the file, the line, and *only* the parameters actually unfilled.
- **It does not false-positive.** Two correct new call sites appended to `commands/doctor.py`, one fully positional and one with aliased keywords, both behind nested imports: zero findings.
- **The reporter cannot be silenced.** `return []` at the top of `_leaking_calls` gives 5 failed.

And re-run against a `main` that had moved 16 merges since my first check, the live injection named **two** missing parameters where it had named one that morning — because `main` had changed that call site underneath it. That is the property that makes a scanner worth keeping: it reads the repository as it moves, not a snapshot of it.

Measured headroom behind your two thresholds is 504 modules / 145 commands against 400 and 120, so both discriminate rather than sitting on a boundary.

**On the history here.** The call-site fix landed in #754, not yours, and you kept going and built the guard anyway. That is the more valuable half and it is the half that usually does not get written, because the issue looks closed once the symptom is gone. Recorded that way in `CONTRIBUTORS.md` (`cbe7363f`).

PR #767, opened 2026-09-09
fix(train): widen the failure boundary so setup errors reach fail_run, and record the run's pid (#764)

### LINKED ISSUE #764: soup train leaves a run at status='running' forever when setup fails: 220 of 432 rows in a real DB
## The defect

`soup train` registers a run in the experiment tracker **before** it loads the model, and the only `except` that marks a run failed covers **the training call alone**. Every failure in between — tokenizer load, model load, quantization, LoRA attach, dataset mapping, trainer construction — leaves the row at `status='running'` forever.

`commands/train.py`:

```
1284    run_id = tracker.start_run(...)          # row inserted, status='running'
1302  \
  ...  |  tokenizer load, model load, LoRA, dataset map, trainer build
1552  /   ~250 lines, NOT inside any try that touches the tracker
1553    try:
1554        result = trainer_wrapper.train(...)
1568        tracker.finish_run(...)
1569    except Exception as exc:
1570        tracker.fail_run(run_id)
```

The `try` at 1297 is unrelated — its body is lines 1298-1300 (`resolve_report_to`).

## Reproduced, unintentionally, on a real machine

A gated-model config (`meta-llama/Llama-3.1-8B-Instruct`, no `HF_TOKEN`) fails exactly where you would expect:

```
Run ID: run_20260909_195847_0ddddab0
Setting up model + trainer...
Loading tokenizer: meta-llama/Llama-3.1-8B-Instruct

Error: This Hugging Face model requires authentication or license acceptance.
```

The error message is good. The tracker row is not:

```
$ soup runs
│ run_20260909_195847_0ddddab0 │ meta-llama/Llama-3.1-8… │ sft │ running │ │ │ │
```

Still `running`, with no steps, no loss and no duration. Nothing will ever change it.

## The scale on a real database

This is not one stale row. Local `~/.soup/experiments.db`:

```
status counts: [('completed', 166), ('failed', 43), ('running', 220), ('spawn_failed', 3)]
```

**220 of 432 runs — more than half — are permanently `running`**, the oldest dating to 2026-03-04. `soup runs` is one of the first commands a new user types, and the majority of what it shows them is wrong. It also makes `soup runs`, `soup why` and any "what happened to my last run" workflow untrustworthy, because a crashed run and a live one are the same colour.

## Why #401's reconciler does not save it

`ExperimentTracker._reconcile_orphaned_run` (`experiment/tracker.py:418`) exists precisely for this shape, and #424 made its liveness check correct on Windows. But it needs a pid, and **`soup train` never records one**:

```
$ SELECT run_id, base_model, status, pid, exit_code FROM runs WHERE run_id LIKE 'run_202609%'
('run_20260909_195847_0ddddab0', 'meta-llama/Llama-3.1-8B-Instruct', 'running', None, None)
('run_20260909_193158_2a504347', 'Qwen/Qwen2.5-0.5B-Instruct',       'completed', None, None)
('run_20260909_192204_0aac49e0', 'meta-llama/Llama-3.1-8B-Instruct', 'running', None, None)
```

`pid` and `exit_code` are `NULL` for every one of them. `mark_running(run_id, pid=...)` is called on the MCP execution path only. So the machinery this project already built for orphaned runs is inert on the path that produces almost all of them.

## Proposed fix

Two independent parts; either alone is an improvement, both together close it.

1. **Widen the failure boundary.** Put setup inside the same `try`, or better, wrap from `start_run` to the end in `try/except/finally` so any exception reaches `fail_run`. Record the exception type on the row — a run that died at "Loading tokenizer" and one that diverged at step 900 should not read the same afterwards.

2. **Register the pid at `start_run`** so `_reconcile_orphaned_run` can do its job for `soup train` exactly as it already does for MCP runs. That is what rescues the rows a `kill -9`, a power loss or a Ctrl+C leaves behind, which no `except` can ever catch.

A migration or a `soup runs --reconcile` for the rows already on disk would be welcome but is a separate decision — 220 rows is enough that silently rewriting them is worse than offering a command.

## Acceptance criteria

- [ ] A config that fails during model load leaves the run at `failed`, not `running`, with the failure reason recorded.
- [ ] A run killed with SIGKILL / Ctrl+C is reconciled to a terminal status on the next `soup runs`.
- [ ] A test that patches the model loader to raise and asserts the tracker status — the current suite has no such test, which is how ~250 lines ended up outside the handler.
- [ ] `soup train` writes a pid, and a test asserts it is non-NULL after a run starts.

Found while running `soup train` on a real dataset rather than in the test suite.

SID-6921:
## Summary

Closes #764. `soup train` registers a run in the tracker before it loads the model, but the only `except` that marked a run failed wrapped the training call alone — everything in between (tokenizer load, model load, quantization, LoRA attach, dataset mapping, trainer construction) left the row at `status='running'` forever on any exception. On the real database in the issue, 220 of 432 runs were stuck this way.

## What's here

Two independent parts, both from the issue's acceptance criteria:

1. **Widened the failure boundary.** Everything from right after the run is registered through the end of the command now runs inside `try/except`, so a setup-phase exception reaches `fail_run` with the exception type and message recorded on the row. `fail_run` gained an optional `error=` kwarg, and `runs` gained an `error_message` column via the existing lazy-migration pattern (same style as the `pid`/`exit_code` columns already there). The pre-existing training-phase `except` (which writes a crash bundle) is untouched — this only closes the gap in front of it.

2. **`soup train` now records its own pid.** `tracker.mark_running(run_id, pid=os.getpid())` right after `start_run`, the same call the MCP execution path already makes. `_reconcile_orphaned_run` (#401) only rescues a row that has a pid; `soup train` runs never wrote one, so a SIGKILL'd or Ctrl+C'd run was invisible to a mechanism this project already built for exactly that purpose. I didn't duplicate #401's own dead-pid reconciliation test — that mechanism is already proven correct in `test_issue401_orphaned_run_reconcile.py`; this PR only proves the new call site actually feeds it a real pid.

## About the diff size

Most of `train.py`'s diff is a mechanical reindent — the setup block moving one level under the new `try:` — not a logic change. The actual new lines are the `mark_running` call, the `try:`/`except:` pair around setup, and one `error=` kwarg added to the pre-existing training-phase `fail_run` call.

## Verification

- 3 new tests in `tests/test_issue764_run_status_reconciliation.py`, invoking `soup train` end to end via `CliRunner` against a real (tiny, local) config, with `SFTTrainerWrapper.setup` or `resolve_report_to` monkeypatched to raise at different points in the setup sequence.
- **Mutation**: reverted both changes and confirmed all three new tests fail against the original code with exactly the reported symptom — status stays `running`, pid stays `None` — then restored and confirmed green.
- Full tracker/train-adjacent regression suite: 1433 tests across 33 files pass, no regressions.
- `ruff check` clean.
- `mypy`: 2 new error lines, both instances of the pre-existing `tracker: str` vs `ExperimentTracker` type-shadowing limitation already responsible for 4 of the file's 12 baseline errors (the code's own comment names this: *"Capture the --tracker CLI value BEFORE the local ExperimentTracker shadows it"*). Not a new class of problem — left alone rather than scope-creeping a fix for it into this PR.

## Not done here, on purpose

A migration or `soup runs --reconcile` for the ~220 rows already stuck on someone's real disk, per the issue's own note that silently rewriting them is a separate decision from fixing the code path that created them.

### DIFF
--- a/src/soup_cli/commands/train.py
+++ b/src/soup_cli/commands/train.py
@@ -193,14 +193,28 @@ def _describe_exception_for_tracker(exc: BaseException) -> str:
     """Format ``exc`` for ``ExperimentTracker.fail_run``'s ``error=`` column.
 
     A handled setup failure (a bad ``--tracker`` value, a hub download
-    error) exits via ``raise typer.Exit(...) from exc``. ``typer.Exit``
-    carries no message of its own, so formatting the caught exception
-    directly writes ``"Exit: "`` to the row and loses the reason a human
-    needs — the one already printed to the console just before the raise.
-    Unwrapping to ``__cause__`` when present recovers it (#764 review).
+    error, a hub-cache path-containment refusal) exits via
+    ``raise typer.Exit(...)``, sometimes chained with ``from exc`` and
+    sometimes not. ``typer.Exit`` carries no message of its own, so
+    formatting it directly writes ``"Exit: "`` to the row and loses the
+    reason a human needs. Unwrapping to ``__cause__`` recovers it for the
+    chained sites; recording ``exit_code`` covers the one site that
+    raises bare, so all three read as something more useful than
+    ``"Exit: "`` (#764/#767 review).
+
+    The unwrap is gated to ``typer.Exit`` specifically — this helper also
+    runs on ordinary training failures (``raise X from Y`` deep inside a
+    library), where the outer exception X is the operator-facing reason
+    and the inner cause Y is just the mechanism. Unwrapping unconditionally
+    would discard X and keep only Y, which is less informative than before
+    this fix existed — the exact regression the same review caught.
     """
-    cause = exc.__cause__ or exc
-    return f"{type(cause).__name__}: {cause}"
+    if isinstance(exc, typer.Exit):
+        if exc.__cause__ is not None:
+            cause = exc.__cause__
+            return f"{type(cause).__name__}: {cause} (exit code {exc.exit_code})"
+        return f"Exit(code={exc.exit_code})"
+    return f"{type(exc).__name__}: {exc}"
 
 
 def train(
--- a/src/soup_cli/commands/train.py
+++ b/src/soup_cli/commands/train.py
@@ -1289,266 +1289,275 @@ def train(
         experiment_name=experiment_name,
         run_id=os.environ.get("SOUP_MCP_RUN_ID") or None,
     )
+    # #764: without a pid, a SIGKILL'd or Ctrl+C'd run has no way for
+    # _reconcile_orphaned_run to ever notice the process is gone — the
+    # rescue path #401 built for MCP-spawned runs was otherwise inert here.
+    tracker.mark_running(run_id, pid=os.getpid())
     console.print(f"[dim]Run ID: {run_id}[/]")
 
-    # Build trainer based on task type
-    from soup_cli.utils.trackers import resolve_report_to
-
     try:
-        report_to = resolve_report_to(
-            wandb=wandb, tensorboard=tensorboard, tracker=tracker_backend
-        )
-    except ValueError as exc:
-        from rich.markup import escape as _esc
+        # Build trainer based on task type
+        from soup_cli.utils.trackers import resolve_report_to
 
-        console.print(f"[red]{_esc(str(exc))}[/]")
-        raise typer.Exit(code=2) from exc
-    console.print("[dim]Setting up model + trainer...[/]")
-    # v0.53.8 #130 — pre-fetch model from non-HF hub into a local cache and
-    # rewrite cfg.base to point at the local snapshot. The trainer wrappers
-    # still use transformers.from_pretrained, which reads HF Hub by default;
-    # by snapshotting first we keep every wrapper unchanged.
-    hub_name = getattr(cfg.training, "hub", "hf") or "hf"
-    if hub_name != "hf":
-        import re as _re
-
-        from rich.markup import escape as _markup_escape
-
-        from soup_cli.utils.hubs import download_repo
-        from soup_cli.utils.paths import is_under_cwd
-
-        # Sanitise cache subdir name — strip every path-separator and
-        # `..` segment so a crafted ``base: ../../etc`` cannot escape the
-        # cache root (Windows ``\\`` and POSIX ``/`` both blocked).
-        safe_slug = _re.sub(r"[^A-Za-z0-9._-]+", "__", cfg.base).strip("._-") or "model"
-        cache_dir = (Path.cwd() / ".soup_hub_cache" / safe_slug).resolve()
-        if not is_under_cwd(str(cache_dir)):
-            console.print(
-                "[red]Resolved hub cache dir escaped the current working "
-                "directory; refusing to download.[/]"
-            )
-            raise typer.Exit(code=1)
         try:
-            # Idempotency: if the cache dir already has a config.json, skip
-            # the re-download (modelscope/openmind-hub also short-circuit on
-            # match but having an explicit probe lets us print a clear hint).
-            existing_cfg = cache_dir / "config.json"
-            if existing_cfg.is_file():
-                local_path = str(cache_dir)
-                console.print(
-                    f"[dim]Using cached snapshot at {local_path}[/]"
-                )
-            else:
-                local_path = download_repo(
-                    hub_name,
-                    cfg.base,
-                    local_dir=str(cache_dir),
-                )
-                console.print(
-                    f"[dim]Fetched {cfg.base} from hub={hub_name} → "
-                    f"{local_path}[/]"
-                )
-            # Use ``model_copy(update=...)`` so the Pydantic field
-            # validators on ``base`` rerun (matches v0.33.0 #47 / v0.40.0
-            # Part B immutability policy).
-            cfg = cfg.model_copy(update={"base": local_path})
-        except ImportError as exc:
-            console.print(f"[red]{_markup_escape(str(exc))}[/]")
-            raise typer.Exit(code=1) from exc
-
-    # v0.53.8 #89 — friendly missing-dep advisory for `--tracker <name>`.
-    if report_to and report_to not in ("none", "wandb", "tensorboard"):
-        from soup_cli.utils.trackers import tracker_missing_dep_message
-
-        msg = tracker_missing_dep_message(report_to)
-        if msg:
-            console.print(f"[yellow]{msg}[/]")
-    trainer_kwargs = {
-        "device": device,
-        "report_to": report_to,
-        "deepspeed_config": ds_config_path,
-        "fsdp_config": fsdp_kwargs,
-    }
-    # v0.40.4 #63 — every transformer-backend trainer now threads
-    # --trust-remote-code through the wrapper (closes the v0.36.0 Part B gap).
-    trainer_kwargs = dict(trainer_kwargs, trust_remote_code=trust_remote_code)
-    from soup_cli.trainer.mlx_routing import resolve_trainer
-
-    mlx_cls, trainer_kwargs = resolve_trainer(cfg, trainer_kwargs)
-    if mlx_cls is not None:
-        trainer_wrapper = mlx_cls(cfg, **trainer_kwargs)
-    elif cfg.task == "dpo":
-        from soup_cli.trainer.dpo import DPOTrainerWrapper
-
-        trainer_wrapper = DPOTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "online_dpo":
-        from soup_cli.trainer.online_dpo import OnlineDPOTrainerWrapper
-
-        trainer_wrapper = OnlineDPOTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "grpo":
-        from soup_cli.trainer.grpo import GRPOTrainerWrapper
-
-        trainer_wrapper = GRPOTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "ppo":
-        from soup_cli.trainer.ppo import PPOTrainerWrapper
-
-        trainer_wrapper = PPOTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "kto":
-        from soup_cli.trainer.kto import KTOTrainerWrapper
-
-        trainer_wrapper = KTOTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "orpo":
-        from soup_cli.trainer.orpo import ORPOTrainerWrapper
-
-        trainer_wrapper = ORPOTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "simpo":
-        from soup_cli.trainer.simpo import SimPOTrainerWrapper
-
-        trainer_wrapper = SimPOTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "ipo":
-        from soup_cli.trainer.ipo import IPOTrainerWrapper
-
-        trainer_wrapper = IPOTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "bco":
-        from soup_cli.trainer.bco import BCOTrainerWrapper
-
-        trainer_wrapper = BCOTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "preference":
-        from soup_cli.trainer.preference import PreferenceTrainerWrapper
-
-        trainer_wrapper = PreferenceTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "reward_model":
-        from soup_cli.trainer.reward_model import RewardModelTrainerWrapper
-
-        trainer_wrapper = RewardModelTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "pretrain":
-        from soup_cli.trainer.pretrain import PretrainTrainerWrapper
-
-        trainer_wrapper = PretrainTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "embedding":
-        from soup_cli.trainer.embedding import EmbeddingTrainerWrapper
-
-        trainer_wrapper = EmbeddingTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "distill":
-        # v0.53.2 #133 — knowledge distillation (student + frozen teacher).
-        from soup_cli.trainer.distill import DistillTrainerWrapper
-
-        trainer_wrapper = DistillTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "prm":
-        # v0.53.11 #126 — Process Reward Model trainer.
-        from soup_cli.trainer.prm import PRMTrainerWrapper
-
-        trainer_wrapper = PRMTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task in ("classifier", "reranker", "cross_encoder"):
-        # v0.53.2 #132 — sequence-classification head.
-        from soup_cli.trainer.classifier import ClassifierTrainerWrapper
-
-        trainer_wrapper = ClassifierTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "unlearn":
-        # v0.71.9 #193 — NPO / SimNPO / RMU unlearning.
-        from soup_cli.trainer.unlearn import UnlearnTrainerWrapper
-
-        trainer_wrapper = UnlearnTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "moe_lora_routing":
-        # v0.71.12 #222 — MoLE per-token routing over N frozen task LoRAs.
-        from soup_cli.trainer.mole_routing import MoleRoutingTrainerWrapper
-
-        trainer_wrapper = MoleRoutingTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "tts":
-        # v0.71.20 #131 — TTS fine-tuning (SFT-style next-token CE over
-        # text + audio-codec-token sequences; per-family templating).
-        from soup_cli.trainer.tts import TTSTrainerWrapper
-
-        trainer_wrapper = TTSTrainerWrapper(cfg, **trainer_kwargs)
-    elif cfg.task == "asr":
-        # v0.71.32 — ASR (Whisper) fine-tuning via Seq2SeqTrainer.
-        from soup_cli.trainer.asr import AsrTrainerWrapper
-
-        trainer_wrapper = AsrTrainerWrapper(cfg, **trainer_kwargs)
-    else:
-        # Keep the transformers/TRL SFT surface outside the backend-first MLX
-        # route. The wrapper is import-light today, but importing it eagerly
-        # makes an MLX-only install depend on that remaining true forever.
-        from soup_cli.trainer.sft import SFTTrainerWrapper
-
-        trainer_wrapper = SFTTrainerWrapper(cfg, **trainer_kwargs)
-    trainer_wrapper.setup(dataset)
-
-    # #350 — PEFT promotes newly-created adapters to fp32. FSDP cannot flatten
-    # those beside bf16/float16 BNB storage, so align every trainable floating
-    # tensor after setup creates PEFT modules and before train() wraps the model.
-    aligned_fsdp_params = 0
-    if fsdp and cfg.training.quantization == "4bit":
-        from soup_cli.utils.gpu import get_compute_dtype
-        from soup_cli.utils.mixed_precision import align_trainable_dtype_for_fsdp_qlora
-
-        aligned_fsdp_params = align_trainable_dtype_for_fsdp_qlora(
-            getattr(trainer_wrapper, "model", None),
-            fsdp=True,
-            quantization=cfg.training.quantization,
-            compute_dtype=get_compute_dtype(),
-        )
-    if aligned_fsdp_params:
-        console.print(
-            f"[green]FSDP QLoRA:[/] aligned {aligned_fsdp_params} trainable "
-            "parameter tensor(s) to the compute dtype"
-        )
-
-    # --- HF auto-push callback (Part B of v0.29.0) ---
-    if push_as:
-        from soup_cli.monitoring.hf_push import build_push_callback
-
-        push_cb = build_push_callback(
-            repo_id=push_as,
-            output_dir=cfg.output,
-            private=False,
-        )
-        if push_cb is None:
-            console.print(
-                "[yellow]--push-as: no HF token available; skipping auto-push[/]"
+            report_to = resolve_report_to(
+                wandb=wandb, tensorboard=tensorboard, tracker=tracker_backend
             )
-        else:
-            hf_trainer = getattr(trainer_wrapper, "trainer", None)
-            if hf_trainer is not None and hasattr(hf_trainer, "add_callback"):
-                hf_trainer.add_callback(push_cb)
-                console.print(
-                    f"[green]HF auto-push enabled[/] -> {push_as} "
-                    "(one branch per save_steps)"
-                )
-            else:
-                console.print(
-                    "[yellow]--push-as: trainer does not expose add_callback; "
-                    "auto-push disabled for this run[/]"
-                )
-
-    # Train with live display and experiment tracking
-    display = TrainingDisplay(cfg, device_name=device_name)
-    console.print("[bold green]Training started![/]\n")
-
-    profiler_ctx = contextlib.nullcontext()
-    if profile_run:
-        from soup_cli.utils.profiling import profile_training
-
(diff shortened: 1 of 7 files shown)


MakazhanAlpamys:
Reviewed in an isolated worktree with the tests run and both halves mutation-tested. The mechanism is right, and the reindent claim checks out — `git diff -w` on `train.py` collapses to exactly four hunks, and `git diff main...HEAD -- tests/` shows no deleted or weakened tests, so the 247 deletions are what you say they are.

**Both mutations are killed.** Removing `tracker.mark_running(run_id, pid=os.getpid())` fails the pid test (1 failed / 2 passed). Removing `tracker.fail_run(...)` from the new setup `except` while keeping the `raise` fails both status tests with `assert 'running' == 'failed'` — the issue's exact symptom.

**I also tested the migration by hand rather than reading it.** I built a pre-v0.34-shaped `experiments.db` (15-column `runs`: no `cost_*`, no `pid`/`command_digest`/`log_path`/`exit_code`, no `error_message`) with a legacy `status='running'` row, then opened it with your code: 15 columns to 23, legacy row still readable, `fail_run` writes through, `list_runs` fine, second `init_db` idempotent. Gated on `PRAGMA table_info` exactly like `pid` and `exit_code`. No migration risk.

And the criterion-2 mechanism genuinely works: with a real dead child pid, `list_runs()` reconciles a `soup train` row to `terminated`. Before this PR those rows had no pid, so they could never reach that state at all.

Four things before I merge.

**1. `error_message` is stored unredacted and unbounded.** I raised a setup failure whose exception text contained `token=hf_ABCDEFGHIJKLMNOPQRSTUVWXYZ012345` plus 5000 characters of padding, and read the row back: `LEN 5074`, token stored verbatim. The project already treats this exact string as sensitive on the sibling path — `utils/crash.py:194` does `redact_secrets(str(error))` — and I confirmed that redactor catches this token. Please reuse it and cap the length, at both `fail_run` call sites since the training-phase one now stores a message too:

```python
from soup_cli.utils.crash import redact_secrets
error=redact_secrets(f"{type(exc).__name__}: {exc}")[:_MAX_ERROR_MESSAGE]
```

**2. `typer.Exit` is an `Exception`, so for the most common setup failures the recorded reason is literally `"Exit: "`.** I patched `resolve_report_to` to raise and read the real row back: `('failed', 23244, 'Exit: ')`. There are three `raise typer.Exit(code=...)` inside your new `try` — `train.py:1310` tracker resolution, `:1335` hub-cache containment refusal, `:1362` hub `ImportError` — and each prints a red message to the console and then loses it. The *status* is correct in all three cases, so this is not a false failure; the reason is just empty. Your second test asserts only `error_message is not None`, which passes on `"Exit: "` — that is why the gap is invisible. Please add a `typer.Exit` branch recording the exit code (or the message that was printed) and tighten that assertion to something that would have caught this.

For the record I checked the widened `try` for false failures specifically, because that is the risk of wrapping 260 lines: there is no `typer.Exit(0)` inside it, and `--dry-run` returns at line 1255 *before* `start_run`. No successful path is mislabelled. That part is sound.

**3. `error_message` is currently written and read by nothing.** `grep -rn error_message src/` hits only `tracker.py` — schema, migration, write. It is not surfaced by `soup runs`, `soup runs show` or `soup why`, so an operator can only see it by opening sqlite by hand. That is the "validated, accepted, and read by nothing" pattern this project has spent two releases naming, and it means the issue's own motivation — "a run that died at 'Loading tokenizer' and one that diverged at step 900 should not read the same afterwards" — is not delivered. One line on the `soup runs show` panel, rendered through `markup_escape`, closes it. If you would rather defer, say so and I will file it rather than let it sit unstated.

**4. A one-line bug your change makes reachable.** `commands/runs.py:166` hardcodes `status_str = "[yellow]running[/]"` in its else branch, while the table at `runs.py:98-101` correctly does `f"[yellow]{status}[/]"` with a comment naming `'terminated'`. Measured after your change: `soup runs` prints `terminated`, `soup runs show <id>` prints `Status:     running`. That is pre-existing from #401, but this PR is what makes `terminated` reachable for `soup train` rows in the first place. Since #764 is "`soup runs` lies about status", please fix that line so the detail view stops lying too.

Optional: acceptance criterion 2 asks for a test at this call site. Your reasoning that #401's test proves the mechanism is fair, but a short test writing a dead pid and asserting `list_runs()` returns `terminated` would close the criterion as written.

Everything else checked out: 3 passed on your test file, 138 passed on `-k "tracker or run_status or experiment"`, ruff clean, changelog fragment present and correctly named after the PR. The one unrelated failure I saw (`test_issue374_stream_disk_preflight.py`, `WinError 1314`) reproduces identically on `main` — symlink privilege on this box, not yours.

src/soup_cli/commands/train.py
```diff
@@ -1289,266 +1289,275 @@ def train(
...
-            energy_ctx = contextlib.nullcontext()
+                energy_tracker = EnergyTracker(country_iso_code=energy_country)
+                energy_ctx = energy_tracker
+            except ValueError as exc:
+                console.print(f"[yellow]--track-energy disabled:[/] {exc}")
+                energy_tracker = None
+                energy_ctx = contextlib.nullcontext()
+
+    except Exception as exc:
+        tracker.fail_run(run_id, error=f"{type(exc).__name__}: {exc}")
```
AmirF194: I ran this against `6387d4b1ecdec553227a4f08713a527d6dae8357` in a clean container.

`error_message` at src/soup_cli/experiment/tracker.py:414-425 loses its content for the three early `typer.Exit` sites inside the new try block (`train.py:1310`, and the two hub-cache-download exits a bit further down). `except Exception as exc:` at `train.py:1558` catches the re-raised `typer.Exit`, not the original exception it was raised `from`, and `typer.Exit` carries no message, so `f"{type(exc).__name__}: {exc}"` renders as `"Exit: "`.

Repro (same monkeypatch shape as `test_a_config_only_failure_before_trainer_setup_also_marks_failed` in this PR's own test file, just reading `error_message` afterward instead of only checking it's non-None):

```
$ python3 -c "
from soup_cli.utils import trackers as trackers_mod
trackers_mod.resolve_report_to = lambda **k: (_ for _ in ()).throw(ValueError('--wandb, --tensorboard, and --tracker are mutually exclusive'))
"
# invoked through CliRunner against soup train --config soup.yaml --yes
exit_code: 2
console output: --wandb, --tensorboard, and --tracker are mutually exclusive
DB row: {'status': 'failed', 'error_message': 'Exit: '}
```

The console gets the real reason, the row that's supposed to distinguish setup failures gets `Exit: `.

Capturing the cause before re-raising would fix all three sites at once:

```python
except Exception as exc:
    cause = exc.__cause__ or exc
    tracker.fail_run(run_id, error=f"{type(cause).__name__}: {cause}")
    raise
```

SID-6921: Confirmed and fixed in `81dc762`. Reproduced your exact repro first — `error_message: 'Exit: '` — then added `_describe_exception_for_tracker` which unwraps to `__cause__` when present, used at both `fail_run` call sites. Strengthened `test_a_config_only_failure_before_trainer_setup_also_marks_failed` to assert the real message content instead of only non-None, which is exactly what let this through the first time. Reverted the unwrap locally to confirm the test reproduces `'Exit: '` before restoring.

SID-6921:
Thanks for the catch — fixed in `81dc762` and replied on the thread with the repro/fix details. Good find: the crash bundle path never had this problem since `trainer_wrapper.train()` doesn't raise `typer.Exit`, but every one of the three handled setup-exit sites did.

MakazhanAlpamys:
Re-reviewed at `81dc762a`. **The `"Exit: "` fix is real and I verified it is tested** — reverting the unwrap so `typer.Exit` falls through to the generic handler fails with `assert 'ValueError' in 'Exit: '`. Strengthening the old `error_message is not None` assertion into one that checks the actual content is what makes that mutation land, and documenting the mutation in your own commit message is the right habit.

But the commit closes one of the four items, and it introduces a regression that I would rather catch now than after merge.

## Regression: the unwrap is unconditional, so it rewrites ordinary failures too

`_describe_exception_for_tracker` does `cause = exc.__cause__ or exc` with no `typer.Exit` check, and it sits on **both** `fail_run` sites — including the training-phase one at `:1593`. So every `raise X from Y` anywhere under those handlers now stores **Y and discards X**:

```
raised   : RuntimeError: failed to load tokenizer for meta-llama/Llama-3.1-8B
recorded : OSError: permission denied: /home/u/.ssh/id_rsa
```

The outer message is the operator-facing reason; the inner cause is the mechanism. Storing only the mechanism makes the row *less* informative than before on the most common failure shape, which inverts #764's own motivation — "a run that died at 'Loading tokenizer' and one that diverged at step 900 should not read the same afterwards". It also widens what ask 1 is about, since inner OS and HTTP causes carry paths and URLs.

Please gate it: unwrap only when the exception is a `typer.Exit`, and keep `f"{type(exc).__name__}: {exc}"` otherwise.

## One of the three enumerated sites still records `"Exit: "`

The guarded `try:` opens at `:1312`. Inside it:

```
:1324  raise typer.Exit(code=2) from exc  ->  'ValueError: simulated --tracker resolution failure'   OK
:1349  raise typer.Exit(code=1)           ->  'Exit: '        <- hub-cache containment refusal, no `from exc`
:1376  raise typer.Exit(code=1) from exc  ->  unwraps OK
```

`e.exit_code` is `1` and sits on the object, unrecorded. This is the site I listed as `:1335` in the original review. Chaining covers the two that chain; a `typer.Exit` branch recording the code covers all three, which is why I suggested that shape.

## Three items are untouched

**1. `error_message` is still stored unredacted and uncapped.** `grep redact_secrets src/soup_cli/commands/train.py` returns nothing, and `tracker.py:423` still writes the raw string. Re-measured, same construction as before:

```
LEN stored             : 5196
token stored VERBATIM? : True
first chars            : 'RuntimeError: failed to download: token=hf_ABCDEFGHIJKLMNOPQRSTUVWXYZ012345 PADDING ...'
redact_secrets() would have caught it: True  ->  'token=<redacted>'
```

And there is no test for redaction or capping, so there is nothing to mutate and nothing to stop it regressing later. A secret-handling fix without a test is the one that quietly comes back.

**3. `error_message` is still read by nothing.** `grep -rn error_message src/` hits only `tracker.py` — schema, migration, write. Not `runs.py`, not `why.py`. Either surface it on `soup runs show`, or say in the thread that you are deferring and I will file it; what I do not want is for it to land unstated, because that is the pattern two releases have been named for.

**4. `commands/runs.py:166` is unchanged.** Executed on your head:

```
list_runs()  (soup runs)      -> terminated
get_run()    (soup runs show) -> terminated
soup runs show panel          -> | Status:     running |
```

The database says `terminated` and the detail view prints `running`. One line.

## For the record

3 passed on your test file, 291 passed / 3 skipped across `-k "tracker or run_status or experiment or runs"`, ruff clean, and **zero removed test lines** across a 247-deletion reindent — I re-checked that, since it is the thing most likely to hide in a diff that size. The two original mutations still die correctly: removing `fail_run` from the setup `except` gives `assert 'running' == 'failed'`, and removing `mark_running` gives `assert None == 28312`.

Minor: the changelog fragment does not mention the `typer.Exit` fix. Worth a clause once the above lands.

src/soup_cli/commands/train.py
AmirF194: Confirmed at `81dc762`. Re-ran the repro from my comment above and error_message now records the real cause instead of "Exit: ".

MakazhanAlpamys:
@SID-6921 — checking in on this one, since you are active on #864. The four items from my re-review on 2026-09-10 are unchanged:

1. A test for the redaction and capping of the stored error message — a secret-handling change without a test is the one that quietly comes back.
2. `error_message` is written and read by nothing: surface it on `soup runs show`, or say in the thread that you are deferring it and I will file it.
3. `commands/runs.py:166`: `soup runs show` prints `running` for a run the database marks `terminated`. It is a one-line fix.
4. A clause in the changelog fragment for the `typer.Exit` fix.

@AmirF194 confirmed the `Exit:` fix independently, so the core is done. If you would rather I finish these four on top of your commits, credited to you, just say so. Otherwise I will look again on **2026-09-13 18:00 UTC**.

MakazhanAlpamys:

```
```

```
```

SID-6921:
Sorry for the delay — got pulled into #864/#869 for a few days. Rebased onto `main` and moved the changelog fragment to `changelog.d/0.75.0/`. All items from the 2026-09-10 re-review addressed in `c866a92`:

**The regression, first.** You're right that the unwrap was unconditional. Gated it to `isinstance(exc, typer.Exit)` specifically — an ordinary `raise X from Y` during real training now keeps reporting X (the operator-facing reason), not Y (the mechanism). Added a regression test (`test_an_ordinary_chained_exception_keeps_the_outer_message`) that reverts the gate and confirms it fails with exactly the shape you described — `RuntimeError: failed to load tokenizer...` collapsing to `OSError: permission denied...`.

**The bare `typer.Exit` site** (hub-cache containment refusal, no `from exc`) now records `Exit(code=1)` instead of `Exit: ` — there's no cause to unwrap there, so recording the exit code is what actually distinguishes it from the old bug, matching the shape you suggested.

**1. Redaction + capping.** `ExperimentTracker.fail_run` now runs `error` through `soup_cli.utils.crash.redact_secrets` and caps at 2000 chars before storing. Reproduced your exact repro numbers first (`hf_ABC...` token, 5196-char message), confirmed it still leaked/wasn't capped on the unmodified code, then fixed it.

**2. `error_message` read by nothing.** Surfaced it — `soup runs show` now prints an `Error:` line when the field is present.

**3. `commands/runs.py:166`.** Was a silent `else: running` catch-all; both `terminated` and `launching` get their own branch now. Same defect existed in the same shape in the list view (`commands/runs.py` list command already had this fixed, I just didn't check the `show` command had the identical bug next to it).

**4. Changelog clause.** Added.

Verified everything by mutation — reverted each fix individually and confirmed the matching new test reproduces the exact symptom before restoring. 112 tests pass across the affected files, ruff clean, mypy shows the same 15 pre-existing errors as the unmodified files (checked via `git stash`), none new.

MakazhanAlpamys:
Reviewed at `c866a925`, 14/14 green on that SHA. This is strong work and the design is the reason: the widened boundary handles exceptions, the pid handles the kills no `except` can ever see, and neither half pretends to do the other's job. Declining to duplicate #401's reconcile test was also right — your test drives the real reconcile path with a genuinely dead pid instead of restating it.

I verified the "mostly a reindent" claim rather than accepting it: filtering the 523-line `train.py` diff to non-indent lines leaves the helper, one `mark_running`, and the try/except pair. I also checked the three `typer.Exit` raises that now fall inside the widened block — all `code=1`/`code=2`, so there is no success-path `Exit(0)` that would be mis-marked failed — and that the handler re-raises, leaving caller behaviour unchanged.

**Mutations — I ran the two that matter myself:**

| mutation | result |
| --- | --- |
| setup `except` no longer calls `fail_run` (pre-PR scope) | **2 failed** — `test_a_setup_failure_marks_the_run_failed_with_a_reason`, `test_a_config_only_failure_before_trainer_setup_also_marks_failed`; the assertion reads `assert 'running' == 'failed'`, the issue's exact symptom |
| `tracker.mark_running(run_id, pid=os.getpid())` removed | **1 failed** — `test_the_run_is_registered_with_the_current_process_pid`, `assert None == <pid>` |

Both killed, independently. The `isinstance(exc, typer.Exit)` gate on the `__cause__` unwrap is tested in both directions, which is the one I most expected to be unguarded.

**One thing I need changed before merge, and I found it by measuring rather than reading.**

This PR creates the surface that prints `error_message` — your own comment says it correctly: *"written by `fail_run` since #764, but read by nothing until now."* At `commands/runs.py:196` that goes through `_esc`, which is `rich.markup.escape` — it neutralises `[...]` and **nothing else**. I stored a failure whose message carries an OSC window-title sequence and ran `soup runs show`:

```
fail_run(rid, error="boom \x1b]0;HIJACKED\x07 and \x1b[31mRED\x1b[0m tail")

RAW_OSC_PRESENT: True
RAW_SGR_PRESENT: True
'r:      boom \x1b]0;HIJACKED and \x1b[31mRED\x1b[0m tail  ...'
```

So a raw OSC reaches the terminal. The message is usually the operator's own exception, which is why this is low severity — but exception text routinely carries remote content (a Hub error body, a server response), and `soup runs show` gets run over CI logs and shared boxes. This is the exact class #893 merged a fix for an hour ago, and I would be inconsistent to take a fresh instance of it now.

The fix is one word — `utils/terminal.py:for_terminal` (C0/ESC strip **plus** markup escape) already exists on `main`:

```python
from soup_cli.utils.terminal import for_terminal
...
info_lines.append(f"Error:      {for_terminal(str(run['error_message']))}")
```

**Your call how it lands:** push it yourself, or tell me and I will apply that one line at merge time with your commit untouched and the merge body saying which line was mine. I would rather you push it — it is your PR and the credit should read cleanly — but I am not going to make you wait on a one-liner if you would rather I did it. A test asserting the ESC does not survive would be welcome too; the redaction and length-cap tests you already wrote are the right model.

**Three non-blocking follow-ups**, take them here or later:

1. `experiment/tracker.py` — `fail_run(run_id)` with no `error` writes `error_message = NULL`, clearing anything already stored. `commands/sweep.py:489` is such a caller. Harmless today because a sweep row never has a message first, but `COALESCE(?, error_message)` sets the invariant while the column is new and double-calling `fail_run` is otherwise lossy.
2. `soup runs show` gained an `Error:` line and real `terminated`/`launching` statuses; a line or two in `docs/commands.md` would cover it.
3. The three new `test_runs.py` assertions read raw `result.output`. I checked and they **do** pass under `FORCE_COLOR=1` with escapes genuinely present, so this is not a finding today — they survive because the asserted words sit inside whole-token `[red]...[/]` spans and the Panel lines do not wrap. A longer real-world error message would wrap and break `assert "simulated tokenizer load failure" in result.output`. The file's strip-and-collapse idiom would make that free.

**Ordering: no constraint on you.** Of the 44 open PRs only #908 and #853 touch `train.py`, at lines 36/1599 and 9-958 against your 192 and 1291-1496; `git merge-tree` is clean against both. #746 touches none of your files.

SID-6921:
Good catch, and thanks for reproducing it rather than just flagging the theory. Fixed in `e371faf` — swapped `_esc` for `for_terminal` on the error_message line specifically (left the config-derived fields on markup_escape, since those aren't arbitrary exception text). Added a test with the same OSC/SGR payload you used; reverting the fix reproduces your exact repro (raw `\x1b` reaching `result.output`).

Also added a line to docs/commands.md for the Error: line and terminated/launching statuses (your follow-up #2) since it was one line.

Deferring #1 (COALESCE on fail_run) and #3 (the raw result.output assertions) — both non-blocking as you said, and #1 in particular changes fail_run's behavior for existing callers, which I'd rather do with its own test than fold into this diff. Happy to open a follow-up issue for those if you'd rather track them separately.

MakazhanAlpamys:
Re-checked at `e371faf4`. **Blocker closed.** `runs.py:199` is `for_terminal(run['error_message'])`, imported from the shared `soup_cli.utils.terminal` — not a private copy — and `utils/terminal.py:29` strips C0/DEL *then* escapes markup, so it is a strict superset of the old behaviour with no regression on the markup half.

**Mutation:** anchor matched exactly once, module really re-imported after mutating (no NameError — `_esc` is still bound at `:186`), and the failure is your repro verbatim:

```
>       assert "\x1b" not in result.output
E       AssertionError: '\x1b' is contained here:
E                boom ]0;HIJACKED and [31mRED[0m tail
```

**Your restraint was right, but I want to correct the reason**, because the distinction matters for the next person reading that hunk. You said the other fields are "config-derived, not arbitrary exception text" — true — and that they are already validated upstream — **that part is not**. `SoupConfig.base` is a bare `base: str = Field(...)` with no control-byte validator, and neither `output.dir` nor the experiment name has one:

| line | field | provenance | verdict |
|---|---|---|---|
| `:187` | `run_id` | server-generated (`generate_run_id()` / `SOUP_MCP_RUN_ID`) | `_esc` correct |
| `:188` | `experiment_name` | user text, **no validator** | pre-existing, leave |
| **`:199`** | **`error_message`** | arbitrary exception text | **`for_terminal` — correct** |
| `:202` | `base_model` | config `base`, **no validator** | pre-existing, leave |
| `:203` | `task` | closed `Literal` | genuinely safe |
| `:204-206` | device / gpu_memory | machine-derived | safe |
| `:212` | `output_dir` | user path, **no validator** | pre-existing, leave |

So the seven untouched lines are unvalidated rather than validated — leaving them is still the right call, because widening into a repo-wide convention is scope creep and none of them is introduced by this PR, but it belongs in a future sweep rather than in the "these are safe" column.

**Merge is clean** despite #907, #884 and #881 landing since your base — real `git merge origin/main` succeeds, sole message `Auto-merging docs/commands.md`. I also built that merge and ran #907's new AST ratchet on it (it bans both the name and the *shape* of a private control-strip table): **6 passed**, your import trips neither.

Your `docs/commands.md` line checks out against the code: `error_message` has exactly one writer (`tracker.py:436`, which sets `status='failed'` in the same UPDATE), and `terminated` / `launching` are real constants that `show()` reaches live, because `get_run` reconciles orphaned runs on read.

112 passed, 1 skipped on the merged tree across your files plus tracker / mcp-execute / smoke-train / trainer-init; zero removed test lines (339 insertions, 0 deletions); ruff clean; fragment correctly named after the PR with house-style credit.

**Two LOW notes, both feeding your deferred item #3 rather than asking for anything now:**

- `tests/test_runs.py:145`'s `assert "\x1b" not in result.output` **fails under `FORCE_COLOR=1 TERM=xterm-256color` even with your fix in place**, because Rich renders the Panel's own `[bold]`/`[yellow]` into real ESC bytes. It passes under what CI actually sets — I walked rich 15.0.0's source, it honours only `FORCE_COLOR` and `TTY_COMPATIBLE`, and the workflows set neither — so this is the documented false-positive class, not a live risk. But it is one workflow `env:` line from becoming one, and it makes the case for your deferred item better than my original note did.
- `runs.py:169,171` hardcode `"terminated"` / `"launching"` while `tracker.py:35-36` define constants for both. If one is ever renamed the display silently falls back to `"running"` — precisely the bug this hunk fixes. They are underscore-private in `tracker`, so importing across modules has its own smell; noting it rather than asking.

CI is queued on `e371faf4`. I will merge when it lands.

SID-6921:
Appreciate the correction — "unvalidated, not validated" is the more accurate way to describe those seven lines, and I'll keep that framing in mind for a future sweep instead of overstating it here. Thanks for building the merge and running #907's ratchet against it too.

MakazhanAlpamys:
Merged as `5dc87efe`. Thank you — and the wait was the repository's, not yours.

I said on 2026-09-13 that CI was queued on `e371faf4` and I would merge when it landed. It then did not land for a day: GitHub Actions produced no terminal outcome at all between 2026-09-13T18:30Z and today 15:08Z, 95+ runs queued with two stuck `in_progress` past 28 hours. The queue drained on its own and this merged at 15:54Z on the same verdict, unchanged.

**What I merged on.** All 13 required contexts green on `e371faf4` itself, confirmed per-SHA. The mutation from my last review still holds: anchor matched exactly once, the module really re-imported afterwards (no `NameError` — `_esc` is still bound at `:186`), and the failure is your repro verbatim:

```
>       assert "\x1b" not in result.output
E       AssertionError: '\x1b' is contained here:
E                boom ]0;HIJACKED and [31mRED[0m tail
```

112 passed, 1 skipped on the merged tree across your files plus tracker / mcp-execute / smoke-train / trainer-init. 339 insertions, **zero removed test lines**. #907's AST ratchet — which bans both the name and the *shape* of a private control-strip table — passes 6 on the merged tree, and your import of the shared `utils.terminal` trips neither half of it.

**What I credited you for, which is the part you did not write.** Seven neighbouring fields in `soup runs show` render unescaped and you routed only `error_message` through `for_terminal`, declining to widen. That was the right call: none of the other seven is introduced by this PR, and turning a bug fix into a repo-wide convention change is how a reviewable diff becomes an unreviewable one.

The one correction I owed you cuts against your stated reason rather than your decision. You described the others as config-derived and already validated upstream; the first half is true and the second is not — `SoupConfig.base` is a bare `base: str = Field(...)`, and neither `output.dir` nor the experiment name carries a control-byte validator. So those lines are *unvalidated*, not validated. Leaving them is still correct, but "safe" was the wrong column and it is now written down for whoever reads that hunk next.

Two LOW notes I did not ask you to act on, both feeding your own deferred item rather than this PR: `tests/test_runs.py:145`'s `assert "\x1b" not in result.output` fails under `FORCE_COLOR=1 TERM=xterm-256color` even with your fix in place, because Rich renders the Panel's own `[bold]`/`[yellow]` into real ESC bytes — it passes under what CI actually sets (rich 15.0.0 honours only `FORCE_COLOR` and `TTY_COMPATIBLE`, and no workflow sets either), so it is the documented false-positive class rather than a live risk, but it is one workflow `env:` line away from becoming one. And `runs.py:169,171` hardcode `"terminated"` / `"launching"` while `tracker.py:35-36` define constants for both.

Credited in `CONTRIBUTORS.md` (`cbe7363f`). #764 is closed with credit.

PR #768, opened 2026-09-09
refactor(ship): share evidence decoding across CLI and MCP (#758)

### LINKED ISSUE #758: Evidence-schema parity: the CLI and MCP readers drift every time a key is added, and nothing tests it
## The defect class

`soup ship --evidence` and the MCP `ship_evidence` tool consume the **same** evidence schema through **two independent readers**. Every time the schema gains a key, the CLI reader is updated and the MCP one is not, so the same file replays to a different answer depending on which surface asked.

This has now happened twice.

- **v0.73.2 added `noise_floor`.** The MCP tool did not read it, so one evidence file produced different verdicts through the CLI and through `soup mcp serve`. It was caught in review and fixed, and `registry.py` carries a comment about it to this day.
- **#746 adds `numerics`.** Same shape, reproduced on that PR's head:

```
                     CLI                        MCP ship_evidence
numerics: "4bit"     exit 0, numerics=4bit      decision=SHIP, numerics=None
numerics: "nf4"      exit 1, REFUSED            decision=SHIP
```

The first row is a newly wrong statement rather than a missing feature: #746 adds `"numerics"` to `verdict_to_dict`, so the MCP tool now **emits** that key and reports `null` for evidence that is stamped. The second is a refusal on one surface and a SHIP on the other.

A one-key patch on #746 closes today's instance. It does not stop the third.

## Why this matters more than it looks

`soup ci init` renders a workflow where `ship --evidence` gates a merge. The MCP tool is read-only and cannot block a PR by itself, so the direct blast radius is smaller — but an operator driving Soup from Claude Code or Cursor is reading a verdict that the CI gate would disagree with, and neither surface says so.

## What would actually close it

**A parity test over the whole evidence schema, not a per-key fix.** Something that feeds a corpus of evidence files — well-formed, malformed, stamped, unstamped, every optional block present and absent — through both readers and asserts they agree on the decision, the refusal, and every reported field. It should FAIL when a key is added to one reader and not the other, which is the property no amount of review discipline has provided: the rule was known, written down, and missed anyway.

The stronger version derives the key set from one place so a new key cannot be added to the schema without appearing in both readers, rather than from a hand-maintained list that rots the same way the readers did.

## Acceptance criteria

- [ ] A test that drives both the CLI evidence reader and `mcp_server/registry.py::tool_ship_evidence` over the same corpus and asserts agreement on decision, exit-class/refusal, and every reported field.
- [ ] The test **fails** when a key is removed from either reader — demonstrated, not asserted, since a parity test that cannot fail is the same defect one level up.
- [ ] The key set is derived rather than hand-listed, or if hand-listed, guarded by a completeness check against the schema.
- [ ] `numerics` parity specifically is covered once #746 lands.

`registry.py` around lines 520-576 is the MCP reader and already shows the pattern for `noise_floor`; `commands/ship.py` is the CLI side. The existing comment in `registry.py` explaining the `noise_floor` case is worth reading first — it states the problem precisely and is the reason this issue exists rather than a third silent divergence.

Filed by the maintainer after reviewing #746; the `numerics` half is being handled on that PR, and this issue is deliberately the general case so it does not ride on one contributor.

ousamabenyounes:
## What does this PR do?

Closes #758.

- Adds one canonical evidence decoder shared by the `soup ship` CLI and MCP surfaces.
- Defines an immutable evidence-field registry and refuses unknown verdict-affecting fields instead of silently dropping them.
- Adds a schema-complete valid/malformed corpus that compares every verdict field, CLI decision exit codes, warning behavior, serializer round trips, and the shared default threshold.
- Proves the parity harness detects independently dropped CLI or MCP fields through mutation runs.

The numerics-stamping portion is intentionally deferred until #746 lands, matching #758's stated acceptance boundary. Unknown future fields are refused in the meantime.

Review follow-up in `39d1a1d`: documents the unknown-field refusal in `docs/evaluation.md`, keeps `McpToolError` free of echoed evidence content by redacting quoted values at the MCP boundary (a flat re-wrap would have regressed the v0.73.2 `noise_floor` contract test), marks the round-trip test as the semantic backstop, and moves the changelog fragment to `changelog.d/0.75.0/`.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Verification

- RED on pristine upstream: the new issue contract fails because the canonical `verdict_from_evidence` interface does not exist.
- GREEN on the patch: all parity-contract tests pass; reverting either adapter's `noise_floor` forwarding makes three parity/refusal tests fail again.
- Changed executable production lines: 100% covered.
- Independent review: two fresh high-reasoning Codex reviewers approved commit `ab01841a00c448b1850e95e44fce053fc95c4241` with no actionable findings.

### Review follow-up (`39d1a1d`, rebased onto `f50a7da3`)

- RED: reverting `_evidence_error_message` fails both hostile-input cases; the pre-rebase tree fails `test_repository_fragments_match_the_newest_release` with `fragment baseline is 0.74.0`.
- GREEN: parity + `test_v07302.py` + changelog suites pass, including the v0.73.2 `noise_floor` contract test.
- Full local suite, pristine `f50a7da3` baseline vs. this branch: 21,130 passed / 2 failed -> 21,162 passed / 2 failed, 127 skipped both sides. The two failures are identical on both sides and pre-existing (a Hugging Face token present in the local environment trips `test_issue677_recipe_repo_id_guard`); they do not occur in CI.

## Checklist

- [x] `ruff check .` passes
- [x] `pytest` passes
- [x] No user-facing option or API documentation changes are required
- [x] Changelog fragment: `changelog.d/0.74.0/768.changed.md`


MakazhanAlpamys:
This is the right shape for #758, and I verified the harness is not decorative before saying so. Mutations on your branch, each reverted afterwards:

- MCP reader alone drops `noise_floor` -> **3 failed**, and the parity test fails on a *decision-level* difference (`{'failed_rule': 'regression'} != {'failed_rule': None}`) — SHIP versus DON'T-SHIP. That is the historical bug, caught.
- CLI reader alone drops `noise_floor` -> same 3 failures.
- `"numerics"` registered in `EVIDENCE_SCHEMA_FIELDS` without corpus support -> `test_corpus_covers_every_registered_evidence_field` fails. The key set is genuinely derived, so a new key cannot be registered without forcing coverage.

No test coverage was removed: all 116 deletions are the two duplicated readers (53 lines from `ship.py`, 60 from `registry.py`, 3 of docstring), and `git diff main...HEAD -- tests/` has zero removed lines. `-k ship` 155 passed, `-k "mcp or evidence or verdict"` 361 passed, ruff clean.

Three things before merge.

**1. `docs/evaluation.md` needs the new refusal rule.** This is a user-facing contract change, not an internal refactor: on `main` an evidence file with an extra root key is ignored and exits 0; on this branch it is `Error: evidence has unsupported field(s): 'X'` and exits 1. `docs/evaluation.md` already documents that a malformed `noise_floor` is "refused, not dropped" — please extend that paragraph to cover unknown fields anywhere in the schema, and say why: a key one reader understands and the other does not is precisely the defect this closes. Your changelog fragment does disclose it; the docs should too.

**2. The MCP tool now echoes evidence content in its error messages, which reverses a deliberate convention.** `mcp_server/registry.py` is documented as "`McpToolError` = path-free/user-input-free message", and every message you replaced was a fixed string. Measured on your branch with a hostile `task.mode`:

```
main / #746 :  "evidence.task.mode must be 'metric' or 'judge_score'"
this branch :  "evidence.task.mode must be one of metric, judge_score, pairwise; got '\x1b[2J\x1b[31mPWNED-BY-EVIDENCE'"
```

The terminal-escape half is already handled — every interpolation goes through `repr()`/`!r`, so a raw ESC byte never survives — so this is not urgent. But arbitrary evidence-file content now reaches the MCP client where it previously could not. Either keep the shared reader's rich message for the CLI and re-wrap on the MCP side as `McpToolError(f"invalid evidence ({type(exc).__name__})")`, or state in a code comment that the `repr()` quoting makes echoing safe here and the convention is being relaxed deliberately. Please do not leave it implicit.

**3. One note on what the harness does and does not cover, for the record.** I mutated the *shared* decoder to ignore `noise_floor` and 24 of 25 still passed. The only test that caught it was `test_schema_keys_are_derived_from_the_canonical_serializer` — the `verdict_to_evidence` round-trip. That is inherent rather than a defect: once one reader serves both surfaces, the parity assertions cannot distinguish "both right" from "both wrong", and the round-trip test is doing all the work there. No change required; I would just ask for a one-line comment marking that test as the semantic backstop, so nobody later prunes it as redundant with the parity cases.

**Coordination point.** This refuses unknown root keys, and #746 adds `numerics`. I merged the two by hand taking this branch's side of `commands/ship.py`: all 25 of your parity tests pass and 14 of #746's 26 numerics tests fail, because the strict reader rejects every evidence file its `--emit-evidence` writes. I am merging yours first and asking #746 to rebase onto it — the four steps are on that PR, and step 2 (moving the `numerics` decode into `verdict_from_evidence`) gives MCP parity for free, which closes your fourth acceptance criterion without either of you writing a bespoke test.

MakazhanAlpamys:
Status update, because this PR has quietly become the head of a chain and I do not want you finding that out from a merge conflict.

**@jagadeepmamidi rebased #746 onto this branch and did all four steps I asked for**, rather than the one-key MCP patch that would have been faster for them:

1. `"numerics"` added to `EVIDENCE_SCHEMA_FIELDS["root"]`
2. the decode moved into `verdict_from_evidence`, with the CLI-side copy deleted — that copy was the drift
3. a stamped `numerics` entry and a malformed `"nf4"` entry added to the parity corpus
4. `_check_evidence_numerics` left CLI-only

So your shared decoder now has a second consumer that was written against it, and #746 currently carries your commits because this has not landed. That is the right order and I am not asking you to hurry — I am telling you the queue shape so the wait is visible rather than mysterious.

**Nothing new is being asked.** The three items from my review on 2026-09-10 are unchanged:

1. **`docs/evaluation.md` needs the unknown-field refusal rule.** This is a user-facing contract change, not an internal refactor: on `main` an evidence file with an extra root key is ignored and exits 0; on this branch it is `Error: evidence has unsupported field(s): 'X'` and exits 1. That paragraph already documents the `noise_floor` refusal — extend it, and say why a key one reader understands and the other does not is the defect being closed.
2. **The MCP error messages now echo evidence content**, reversing the documented "`McpToolError` = path-free/user-input-free message" convention. Either re-wrap on the MCP side as `McpToolError(f"invalid evidence ({type(exc).__name__})")`, or state in a code comment that the `repr()` quoting makes echoing safe and the convention is being relaxed deliberately. I do not mind which; I mind it being implicit.
3. **One comment on `test_schema_keys_are_derived_from_the_canonical_serializer`** marking it as the semantic backstop, so nobody later prunes it as redundant with the parity cases. My mutation run is the reason: ignoring `noise_floor` in the *shared* decoder left 24 of 25 tests green, and that round-trip test was the only one that caught it.

All three are small. The harness itself is verified and I have said so — mutating either reader alone fails 3 tests on a decision-level difference, and registering a key without corpus support fails the coverage test.

If you would rather I made these three changes myself and merged with the work credited to you — your commits, your `CONTRIBUTORS.md` entry, your name in the fragment — say the word and I will. That offer is genuine and costs you nothing; I would just rather ask than assume.

MakazhanAlpamys:
@ousamabenyounes — a check-in, because #746 (@jagadeepmamidi) is still rebased on this branch and waiting on it. Nothing new is being asked: the three items from my review on 2026-09-10 stand, starting with the `docs/evaluation.md` paragraph for the unknown-field refusal, since that is a user-facing contract change.

If you would rather I make those three changes on top of your commits, credited to you, say so and I will. Otherwise I will look again on **2026-09-13 18:00 UTC**, and if it is still quiet then I will ask once more before touching anything.

MakazhanAlpamys:

```
```

```
```

ousamabenyounes:
Thanks for the review @MakazhanAlpamys — and for the mutation run, which is what turned item 3 from a nicety into something I could write down precisely. All three items plus the changelog rebase are in `39d1a1d`.

**1. `docs/evaluation.md` — unknown-field refusal.** Extended the paragraph that already documents the `noise_floor` refusal, rather than starting a new section, since it is the same rule widening. It shows the actual `Error: evidence has unsupported field(s): 'numerics'` / exit `1` pair (copied from a real run, not hand-written) and states the why: a key one reader understands and the other does not is the divergence this closes, so a field is either supported by both surfaces or accepted by neither.

**2. MCP error messages.** Took the re-wrap option, with one adjustment worth flagging. The literal `McpToolError(f"invalid evidence ({type(exc).__name__})")` breaks `test_v07302.py::TestTheMcpEvidenceReaderHonoursTheFloor::test_a_malformed_floor_is_refused_by_the_mcp_reader_too`, which asserts `pytest.raises(McpToolError, match="noise_floor")` — that test passes on `main` and on this branch as you reviewed it, so flattening the message would have been a regression rather than a fix.

What actually leaks is the **quoted value**, never the schema path: the decoder interpolates evidence content only through `!r`/`repr()`, and writes the structural half unquoted. So `_evidence_error_message` redacts every quoted run at the MCP boundary and keeps the rest:

```
before:  evidence.task.mode must be one of metric, judge_score, pairwise; got '\x1b[2J\x1b[31mPWNED-BY-EVIDENCE'
after:   evidence.task.mode must be one of metric, judge_score, pairwise; got <redacted> (ValueError)
after:   invalid evidence.noise_floor: evidence.noise_floor.runs must be in [2, 10] (ValueError)
```

The CLI diagnostic is untouched — the convention is MCP-side, and this keeps it explicit rather than implicit, as you asked.

**3. Semantic backstop comment.** Added to `test_schema_keys_are_derived_from_the_canonical_serializer`, recording your 24-of-25 result and *why* it is inherent: once one decoder serves both surfaces the parity assertions can only prove the readers agree, not that they are right, so the round-trip is the only thing standing between a shared-decoder drop and a green suite.

**4. Changelog fragment.** Rebased onto `f50a7da3` and `git mv`'d the fragment to `changelog.d/0.75.0/768.changed.md`.

**Validation.** RED→GREEN on the real machine, not simulated:

- RED: reverting `_evidence_error_message` fails both new hostile-input cases; the pre-rebase tree fails `test_repository_fragments_match_the_newest_release` with exactly the `fragment baseline is 0.74.0` error you predicted.
- GREEN: the full parity + `test_v07302.py` + changelog set passes, including the `noise_floor` contract test.
- Full local CI replay (`ruff` + `pytest tests/` + smoke), baseline on pristine `f50a7da3` vs. this branch: `2 failed, 21130 passed, 127 skipped` → `2 failed, 21162 passed, 127 skipped`, no new failures.
- 100% coverage on the changed production lines.

No rush understood on the freeze — this is ready whenever the tag lands.

MakazhanAlpamys:
@ousamabenyounes — reviewed at `39d1a1dd`. **All four items are addressed and your dispute of item 2 was correct.** One finding on the replacement, with a verified one-line fix, and then this merges.

## Your dispute was right, and I want to say so plainly

I asked for `McpToolError(f"invalid evidence ({type(exc).__name__})")`. You refused it because it breaks `test_v07302.py::TestTheMcpEvidenceReaderHonoursTheFloor::test_a_malformed_floor_is_refused_by_the_mcp_reader_too`, which matches on `"noise_floor"`. I confirmed that test exists and passes on your branch: my literal suggestion would have been a v0.73.2 regression. Pushing back on a maintainer request with the test that disproves it is exactly the right move.

## The finding — the regex assumes quotes are delimiters, and `repr()` does not guarantee that

Your code comment says *"`repr()` always balances its quotes"*, which is true. But the regex needs something stronger: that every quote character **is** a delimiter. When a value contains **both** quote types, `repr()` picks single-quote delimiters and **backslash-escapes the internal single quotes** — those are not delimiters, the pairing slips by one, and the text between them survives. Reproduced on your head:

```
RAW   : evidence.benchmarks['a\'LEAKED-EVIDENCE-CONTENT\'b"'] needs 'x' and 'y'
SANIT : evidence.benchmarks[<redacted>LEAKED-EVIDENCE-CONTENT\<redacted>] needs <redacted> and <redacted> (ValueError)
```

Four reachable paths leak this way: a hostile `task.mode`, a benchmark name with a missing key, a benchmark name with an unknown subkey, and an unknown root key. The mirror image also happens — a value like `a"LEAK"b'` eats the **structural** text (`] needs`, `and`) instead, which is the opposite of what the helper is for.

**Severity is low, and I checked rather than assumed.** A raw ESC never survives: `repr()` escapes it, and `server.py:73` applies `_sanitize(str(exc))` at the transport boundary regardless. And `verdict_to_dict` already returns `benchmark_deltas[].name` verbatim on the **success** path on `main`, so evidence-controlled names are not a new exposure category. What is wrong is narrower: the path-free convention is only mostly restored, and the comment states an invariant the regex does not actually rely on.

**Verified fix, one line:**

```python
_EVIDENCE_ERROR_QUOTED = re.compile(r"'(?:\.|[^'\])*'|\"(?:\.|[^\"\])*\"|['\"].*\Z", re.DOTALL)
```

Measured on the four cases:

| input | leak | structural text |
|---|---|---|
| `a'LEAKED'b"` (both quote types) | **none** | `] needs … and …` intact |
| `a"LEAK2"b` (double inside single) | **none** | `] bad` intact |
| `mmlu` (ordinary) | none | intact |
| unbalanced trailing quote | none | intact |

It preserves all four of your own sanitizer unit cases byte-for-byte. Worth two extra parametrize entries with a both-quote-types key, and the comment reworded to say the regex tracks delimiters rather than balance.

## What I verified otherwise

- **Zero deleted or weakened tests.** `git diff origin/main...pr768 -- tests/ | grep -cE "^-[^-]"` = **0**; all 116 deletions are the two duplicated readers in production code.
- **Four mutations, four killed.** Core (shared decoder ignores `noise_floor`) fails 4, including the round-trip backstop. Parity, MCP side only: fails 4, and on a **decision-level** divergence — `{'failed_rule': 'regression'}` vs `{'failed_rule': None}`, i.e. DON'T-SHIP vs SHIP, which is the historical bug this refactor exists to prevent. Parity, CLI side only: fails 3. Disabling the redaction: fails 5. The parity harness is not decorative.
- **Not behaviour-preserving, and every deviation is intentional and tested** — I enumerated seven. The unknown-field refusal is new on both surfaces; MCP's `forgetting_threshold` messages changed text while the refusal is preserved (and no test asserted the old strings, which I checked); `OverflowError` is now caught on the CLI where `main` raised an unhandled traceback. Every CLI message string for task/benchmarks/noise_floor is preserved **verbatim**.
- PR file alone **32 passed**; with 17 pre-existing ship/MCP files **1175 passed, 5 skipped**; changelog **17 passed**; CI **14/14** on the exact head; ruff clean; fragment correctly at `changelog.d/0.75.0/`.
- `docs/evaluation.md` verified by execution, not reading: PR branch refuses with `Error: evidence has unsupported field(s): 'numerics'` and exit 1 where `main` renders a SHIP panel and exits 0 — character-exact against the documented block.

## Two ways to finish, your call

1. **You push the one-liner** plus the two parametrize cases. I merge on green.
2. **I apply it at merge**, credit unchanged — the PR stays yours in CONTRIBUTORS.md and the commit body says the regex came from review.

I am nudging towards whichever is faster because **#746 is blocked behind this**: @jagadeepmamidi built on your branch and must rebase once this lands. Which brings me to a warning I will repeat to them — after the rebase, `mcp_server/registry.py` has exactly one semantic conflict:

```
<<<<<<< #746   raise McpToolError(str(exc)) from exc
=======  #768  raise McpToolError(_evidence_error_message(exc)) from exc
```

`_evidence_error_message`, its regex and the `import re` all merge in **cleanly**, so a careless "keep ours" leaves your sanitizer as dead code while forwarding raw messages — and none of their numerics tests would notice.

Say which option you want and I will do it today.

MakazhanAlpamys:
Merged as `9e950c0a`, and I took option 2 — the regex landed as a follow-up commit `c9ccf8c6` with your work untouched and the commit body saying the pattern came from review.

**I owe you the reason it took 22 hours.** I ended my review with *"Say which option you want and I will do it today"* and then waited for an answer instead of acting. That was the wrong call twice over: the choice was mine to make either way, and **#746 has been blocked behind this PR for three days** while I waited. When I offer to do something myself, the offer should not become a dependency.

**What I verified before merging:** 14/14 green on `39d1a1dd`, `mergeable=clean`, zero deleted or weakened tests — all 116 deletions are the two duplicated readers in production code. The four mutations hold, and the one I care about most is the parity harness failing on `{'failed_rule': 'regression'}` against `{'failed_rule': None}` — DON'T-SHIP against SHIP. A parity test that fails on a message string would have been decoration; yours fails on the decision, which is the historical bug this refactor exists to prevent.

**On the follow-up regex**, with the measurement rather than the claim:

```
raw   'a\'LEAKED-EVIDENCE-CONTENT\'b"' needs 'x' and 'y'
old   <redacted>LEAKED-EVIDENCE-CONTENT\<redacted> needs <redacted> and <redacted>
new   <redacted> needs <redacted> and <redacted>
```

All four of your existing sanitizer cases are preserved byte for byte; two parametrize cases added; reverting the pattern fails `both-quote-types` specifically. **One correction to my own review while I am here:** I claimed the mirror image also leaks — that a value like `a"LEAK"b'` eats the structural text. I measured it while applying the fix and it does **not**; only the both-quote-types case leaks. The finding was real, my description of its second half was not.

**The thing I want on the record.** You pushed back on a review request of mine and were right. I asked for `McpToolError(f"invalid evidence ({type(exc).__name__})")`; you refused it and named `test_v07302.py::TestTheMcpEvidenceReaderHonoursTheFloor` as the test it would break, which would have been a v0.73.2 regression. Bringing the test that disproves the request is the strongest possible form of disagreement and it is why the suggestion was withdrawn rather than argued. Both that and the decision-level parity harness are in your `CONTRIBUTORS.md` entry.

@jagadeepmamidi — **#746 is unblocked.** Rebase onto `main`. The warning I gave you stands and is now concrete: after the rebase, `mcp_server/registry.py` has exactly one semantic conflict —

```
<<<<<<< #746   raise McpToolError(str(exc)) from exc
=======  #768  raise McpToolError(_evidence_error_message(exc)) from exc
```

`_evidence_error_message`, its regex and the `import re` all merge in **cleanly**, so a careless "keep ours" leaves the sanitizer as dead code while forwarding raw messages, and none of your numerics tests would notice. Take #768's side on that line. Your fragments also need moving from `changelog.d/0.74.0/` to `0.75.0/`, including `768.changed.md` which is @ousamabenyounes's and should now be dropped rather than moved.

PR #769, opened 2026-09-09
Ercan1 readme in 7 languages

Ercaner1988:
## Type of change

- [ ] Bug fix
- [x ] New feature
- [ ] Refactor / cleanup
- [x ] Documentation
- [ ] Tests

## Checklist

- [x ] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed


Ercaner1988:
Sorry for bothering you with the other pull request about graft before. I was trying to implement graft in my workflow and workspace, but did it poorly.

Ercaner1988 (commented):
Well I have finally worked  it through, hope it will be helpful

MakazhanAlpamys (changes requested):
Thanks for putting the time into this, and welcome — this is your first PR here, so I approved the workflow run manually.

I can't merge it as it stands. Concrete blockers, worst first:

1. **`README.md` is now Turkish.** GitHub renders `README.md` as the project's front door, so this swaps the repository's default language. That isn't a call a translation PR can make. `README.md` must stay English; a Turkish translation belongs at `README.tr.md`.

2. **`README.en.md` is a 45-line stub.** The English README on `main` is 518 lines. This PR relocates English into a file containing only the banner and badges, so after the merge English readers lose roughly 90% of the content — Quick Start, Configuration, the `docs/` index. That is a regression, not a relocation.

3. **Working files are committed.** `original_readme_backup.md` (328 lines), `upstream_original.md` (518 lines) and `banner_template.txt` are artifacts of your own translation process. Git is the backup; please drop all three.

4. **Unrelated ignore-file changes.** The added `.gitignore` and new `.ignore` entries reference a local tool (`graft/`) this repository does not use. Those hunks belong to a different change.

5. **Content was lost in the pass.** `README.md` goes 518 -> 421 lines, so sections did not survive translation.

Beyond the defects, there's a maintenance question I'd rather raise now than after you fix five things. This README changes on almost every release, and it sits under a hard ~500-line ceiling with a standing rule that a release pushing it over has to displace a section into `docs/` rather than raise the limit. Seven static copies multiply that per-release cost by seven, and nothing in this PR keeps them in sync — realistically they are stale by the next release, and a stale translation is worse than none because it silently teaches the wrong command.

So I need to settle the project's position on translations, and on what stops them rotting, before taking a seven-language change. I'll follow up here once I have.

If you'd like a version of this contribution that I can review on its merits today, the smallest one is: leave `README.md` untouched and add a single `README.tr.md` — one file, the language you actually speak, plus a one-line link from the top of `README.md`. No backups, no ignore-file edits. That one is reviewable immediately.

MakazhanAlpamys:
I said I would come back with a decision on translations, so here it is: **we will take them, on one condition.** And I owe you a correction first.

## Where I was wrong

My earlier comment leaned on "nothing keeps these in sync" and implied the problem was unsolvable. It is not — this repository already solves exactly this problem twice, and I should have pointed you at the pattern instead of at the door.

## The condition: a sync ratchet, and it has to be able to fail

Translations are not refused because they rot. They are refused **until something makes rot loud**. The precedent is `tests/test_recipe_count_is_synced.py`, whose docstring describes your situation almost word for word:

> Git's 3-way merge conflicts loudly on Python test files, but silently auto-merges Markdown documentation lines, leaving documentation counts quietly out of date.

That is what happens to a translation: nothing conflicts, nothing fails, and the file quietly starts teaching commands that no longer exist. The second precedent is `bundled_scorer_fingerprint()` in `src/soup_cli/eval/gate_suites.py` — a SHA-256 derived from the real content and locked against a stored constant.

Concretely, what I am asking for:

1. Each translation carries a stamp of the English README it was translated from, as an HTML comment near the top so it does not render — for example `<!-- synced-from: README.md sha256:<hex> -->`.
2. A test recomputes `sha256(README.md)` and compares it against every translation's stamp.
3. When `README.md` changes and a translation has not been updated, that test goes **red**.

**One design point that decides whether this is real or decorative, and it is the thing I would review hardest.** The test must fail when a stamp is **absent**, not only when it mismatches. Otherwise deleting the comment line is the loophole, and a ratchet with a loophole is worse than none because it reads as protection. This repository has already shipped one lock that turned out to be decorative — the first version of `BUNDLED_SCORER_FINGERPRINT` could not move, so reverting a real behaviour change left the hash byte-identical. Please write a test for your test: prove it goes red when `README.md` changes, and red again when a stamp is removed.

Be aware of what you are signing up for: once this exists, **every future change to `README.md` turns CI red until the translations are updated.** That is the point of it. It is also why I am asking you to name, in the PR, who maintains each language. A language with nobody behind it gets removed rather than left stale — I would rather ship five accurate translations than seven, two of which lie.

## The defects from my first review still stand

Unchanged, and independent of the above:

1. **`README.md` stays English.** It is what GitHub renders by default; Turkish belongs at `README.tr.md`.
2. **No 45-line `README.en.md` stub.** If English lives at the top of `README.md`, there is nothing to relocate.
3. **Drop `original_readme_backup.md`, `upstream_original.md` and `banner_template.txt`.** Git is the backup.
4. **Revert the `.gitignore` and `.ignore` hunks** — they reference a local tool this repository does not use.
5. **Nothing may be lost in translation.** `README.md` went 518 lines to 421 in your branch; a translation should carry every section, and the stamp above is what will keep it that way.

## If you would rather start smaller

Landing the mechanism plus one language is a completely acceptable version of this PR, and it is the one I would pick. Get `README.tr.md` and the sync test merged, then add languages in follow-ups once the ratchet has proven it can fail. Splitting it that way also means the review of the mechanism is not entangled with the review of six translations.

Either way, this is now a "yes, with a gate" rather than an open question. Thank you for the patience while I worked out which one it was.

MakazhanAlpamys:
Closing this one as superseded by #774, which carries the sync ratchet I asked for here and none of the backup / `.gitignore` / `.ignore` noise. My full review is on that thread — short version: the mechanism is right, `README.tr.md` is the honest translation of the six, and the other five are translations of a document that is not `README.md`.

Nothing is lost by closing this; the work continues in #774.

PR #770, opened 2026-09-10
docs(config): record the measured leak in the field-consumer guard

Srinivasan8888:
Follow-up to #751, which you merged as `0e04c2c9`. You measured where the global-name-space hole actually leaks and asked for the example in the docstring, *"so the next reader sees it is a ratchet with a known hole rather than a proof."*

Refs #748.

## Reproduced on current main, not copied from the review

```
training.verbose         caught
training.top_k           LEAKS   -- `top_k` is an attribute elsewhere
training.top_p           LEAKS
training.temperature     LEAKS   -- request.temperature, commands/serve.py
training.dtype           LEAKS
training.seed            LEAKS
training.logging_steps   LEAKS
```

Your two examples are exact. I ran the rest of the generic names the docstring already predicted would leak, and they all do.

## Written as a test as well as prose

A leak table in a docstring goes stale silently, which is the failure mode the guard itself exists to prevent. `test_the_known_leak_is_still_the_known_leak` fails if any name moves in either direction, and its message says the fix is to update the table rather than to edit the assertion.

It pins the **caught** side too — `verbose` — so a name the guard can currently see cannot be lost without someone noticing. That direction matters more: losing a catchable name is a silent weakening, where a name becoming catchable is good news.

## On the housekeeping you flagged

You were right that #751's body drifted from the diff — it said 40 entries and 23 tests while the file pinned `== 39` and collected 29. That happened because I rewrote the body once mid-review and then made three more pushes. Nothing to fix on a merged PR, but the lesson is the body should be written last, not maintained in parallel.

## One mutation that survives, stated rather than hidden

Making the pin vacuous — empty the leak tuple *and* neuter the `verbose` assertion — still passes, because a test that asserts nothing cannot fail. That is the same shape as the `staging` predicate on #751: you cannot kill a check by loosening it, only by asserting what it refuses. I have not found a way to pin "this test is not vacuous" from inside the same file that would not itself be vacuous; the honest mitigation is that it asserts both directions, so a one-sided edit is visible in review.

```
30 passed in 1.8s   ruff clean   test-only, no src/ change
```

### DIFF
--- a/tests/test_issue748_config_fields_reach_a_consumer.py
+++ b/tests/test_issue748_config_fields_reach_a_consumer.py
@@ -20,10 +20,21 @@ and out of scope here.
 attribute or key anywhere under ``src/``, not whether it appears on a config
 object. The consumed set is roughly 3,400 names -- most of the codebase's
 attribute namespace. A field named after a common attribute therefore reads as
-consumed on the strength of an unrelated one: a `temperature` field is
-"consumed" by `request.temperature` in `commands/serve.py`, and `top_p`,
-`dtype`, `timeout` and `seed` behave the same way. The guard leaks hardest on
-exactly the generic names a new HuggingFace/TRL passthrough field would carry.
+consumed on the strength of an unrelated one. Measured on this tree, an
+unwired field would be caught or missed as follows:
+
+    training.verbose         caught
+    training.top_k           LEAKS   -- `top_k` is an attribute elsewhere
+    training.top_p           LEAKS
+    training.temperature     LEAKS   -- `request.temperature`, commands/serve.py
+    training.dtype           LEAKS
+    training.seed            LEAKS
+    training.logging_steps   LEAKS
+
+So this is a ratchet with a known hole, not a proof. It leaks hardest on
+exactly the generic names a new HuggingFace/TRL passthrough field would carry,
+which is the case most likely to arise. `test_the_known_leak_is_still_the_
+known_leak` pins the boundary so it cannot move without someone noticing.
 Scoping reads to config-typed objects is a much larger piece of work and is
 not attempted here.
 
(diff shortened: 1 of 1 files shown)


MakazhanAlpamys:
Merged as `07be4342`. You turned a request for a docstring note into an executable pin, and your reason for doing so is the right one — a leak table in a docstring goes stale silently, which is the exact failure mode the guard itself exists to prevent.

**I reproduced your table rather than taking it from my own review comment.** I injected real unwired fields into the actual `TrainingConfig` in `src/soup_cli/config/schema.py` and ran the real guard, restoring between each:

```
baseline, nothing injected            PASSES  (correct)
training.verbose                      FAILS   = caught
training.top_k                        PASSES  = leaks
training.top_p                        PASSES  = leaks
training.temperature                  PASSES  = leaks
training.dtype                        PASSES  = leaks
training.zzz_definitely_unique_name   FAILS   = caught   (control)
```

The control is what makes the four leaks meaningful rather than a dead test. Your attribution is exact too — `top_k` leaks through `probability_policy.top_k` in `soup_cli/autodistill/`, a subsystem that is not wired to the CLI at all, which is the unrelated-namespace collision in its purest form.

**Two things I want to name specifically.**

*You pinned the direction that matters more.* The `verbose` half is not symmetry for its own sake — it guards against silent weakening. I proved it: adding a single `.verbose` read to one source module makes the test fire with a message naming exactly that cause. That is the failure nobody would have noticed, because the guard would still be green while quietly seeing less.

*Your assertion was hard to falsify, and I had to work to mutate it.* My first attempt renamed two of the `probability_policy.top_k` reads and the test survived — because `top_k` has around fourteen consumers across autodistill, MoLE and serve, so my mutation was under-powered rather than your test being weak. Rewriting every `top_k` across all nineteen source files that carry one makes it fire, with a message telling the reader to update the table instead of edit the assertion. That is the right message to have written.

Two rows in the table I could not verify by injection — `seed` and `logging_steps` already exist as `TrainingConfig` fields, so there is nothing to inject. They are correct as consumed-set statements; just flagging that they rest on a different kind of evidence than the other four.

You also reported a surviving mutation of your own and corrected the stale #751 description unprompted. 30 passed, ruff clean, one file touched, zero tests removed, and no changelog fragment — correctly, since nothing user-visible changed.

Issue #775, opened 2026-09-10
Audit: path containment leftovers + Rust-port SWOT (v0.74.0)


Ercaner1988:
## Summary

Audit of Soup CLI **v0.74.0** (`src/soup_cli/`) covering:

1. Code-quality findings against `AGENTS.md` / `CONTRIBUTING.md` conventions
2. Test-gap and dependency notes
3. SWOT for a possible Rust port (hybrid / non-rewrite)

Fork used for the audit: [`Ercaner1988/Soup`](https://github.com/Ercaner1988/Soup) (issues disabled there, so this is filed upstream).

---

### 1. Remaining `Path.relative_to()` uses (Windows 8.3 risk)

Project convention (`AGENTS.md`, `utils/paths.py`): prefer `os.path.realpath` + `os.path.commonpath` / `is_under()` — **not** `Path.resolve()` + `relative_to()`.

Most call sites already document / use the realpath helper. **Live** `relative_to()` call sites still present:

| File | Lines | Code |
|------|-------|------|
| `src/soup_cli/commands/adapters.py` | 38 | `config_file.relative_to(directory).parts` |
| `src/soup_cli/commands/adapters.py` | 107 | `adapter_path.relative_to(dir_path)` |

Nearby in the same file, `Path(...).resolve()` is also used (e.g. 79, 132, 183–184). Other modules still call `.resolve()` for path normalization (e.g. `cans/run.py`, `commands/data.py`, `commands/export.py`, `commands/generate.py`, `commands/train.py`, `data/loader.py`, `migrate/common.py`, `ui/app.py`, …). Many grep hits for `Path.resolve()` are **comments** explaining why *not* to use it — those are not bugs.

**Ask:** migrate remaining `relative_to` / bare `resolve` containment checks onto `soup_cli.utils.paths.is_under` / `is_under_cwd`, and consider a CI source-grep guard (same pattern as the existing “no top-level torch” tests).

### 2. Lazy heavy imports

Module-level `torch` / `transformers` / `peft` / `trl` / `mlx` imports were **not** found as real imports. Hits in `commands/deploy.py` are inside Gradio/Streamlit **template strings** for HF Spaces — false positives.

### 3. Dependency upper bounds (light)

In `pyproject.toml`, most train-stack packages are capped; a couple of core deps still look open-ended (worth confirming intent):

- `plotext>=5.2.0` (no upper bound; comment notes 5.x vs 6.x compatibility via `plotext_compat`)
[...]

PR #778, opened 2026-09-10
test(paths): repo-wide AST ratchet for relative_to containment (#775)

### LINKED ISSUE #777: The path-containment guard is a seven-file allowlist matching one literal spelling, so it cannot catch the next occurrence
Found by @Ercaner1988 in #775, who proposed adding a CI guard for this. The guard already exists — and checking it turned up two ways it cannot do the job it is named for.

## The convention

`.claude/CLAUDE.md` and `CONTRIBUTING.md` both require `os.path.realpath` + `os.path.commonpath` for path containment, never `Path.resolve()` + `relative_to()`, because the latter fails on Windows short names (`C:\Users\RUNNER~1`) and silently reports "outside" for a path that is inside.

## What guards it today

`tests/test_code_review_recurring.py:23-40`:

```python
def test_containment_holdouts_migrated_to_commonpath_helper():
    for rel, needle in {
        "utils/ollama.py": "is_under_cwd",
        "migrate/common.py": "is_under_cwd",
        "commands/serve.py": "is_under(",
        "commands/export.py": "is_under_cwd",
        "commands/generate.py": "is_under(",
        "data/loader.py": "is_under(",
        "eval/checkpoint_intelligence.py": "is_under(",
    }.items():
        src = _src(rel)
        assert needle in src, f"{rel} not migrated to {needle}"
        assert ".relative_to(cwd)" not in src, f"{rel} still uses relative_to(cwd)"
```

## Two holes

**1. It is a seven-file allowlist, not a repo-wide ratchet.** The loop visits exactly those seven modules. A new file that introduces the bad idiom passes, because nothing looks at it. The test pins that a past migration stayed migrated — it does not stop the next occurrence, which is what a guard is for.

**2. It matches one literal spelling.** `".relative_to(cwd)" not in src` misses `.relative_to(base)`, `.relative_to(root)`, `.relative_to(self.root)`, or the call split across two lines — including inside the seven files it does check. This is the same defect shape recorded in `.claude/CLAUDE.md` under the format-fix lesson: a check written from the one example in the ticket, matching that spelling and nothing else.

## Why it has not bitten yet

It has not. `main` is clean today: `grep -rn "\.relative_to(" src/soup_cli/` returns exactly two live calls, both in `commands/adapters.py`, and both are benign — `:38` is a scan-depth counter whose operand comes from the same `rglob` root (so no independently-derived second path exists to disagree, and it fails closed on `ValueError`), and `:107` shortens a path for display with an absolute-path fallback. Every other hit in the tree is a comment documenting the convention.

So this is a ratchet against the next occurrence, not a bug report.

## Acceptance criteria

- [ ] A repo-wide scan over `src/soup_cli/**/*.py` — AST, not substring — that flags `relative_to` / `is_relative_to` used as a containment predicate (inside an `if`, a boolean expression, or a `try` whose `except ValueError` decides acceptance).
- [ ] An explicit allowlist naming `commands/adapters.py:38` and `:107` with the reason each is benign, so the scanner reports zero offenders today for a stated reason rather than by accident.
- [ ] The scan catches renamed spellings: a test that inserts `.relative_to(base)` as a containment check into a temporary module and asserts the scanner flags it.
- [ ] **A test proving the scanner can fail.** A guard that finds zero offenders is indistinguishable from a broken one, and this repository has shipped a lock that turned out to be decorative before. `tests/test_cli_help_assertions_are_ansi_safe.py::TestTheScannerCanActuallyFail` is the working precedent for both the AST shape and this test.
- [ ] The seven existing per-file assertions either fold into the scan or stay as-is; either is fine, but the repo-wide check must not depend on the allowlist to pass.

## Notes

`tests/test_cli_help_assertions_are_ansi_safe.py` is the model to copy — it is the same class of repo-wide convention ratchet, and its history is worth reading first: its own first version passed on an invented example while missing the real commit, and its first normaliser check flagged 31 correct assertions across 18 files. Both mistakes are cheap to repeat here.

@Ercaner1988 — this is yours if you want it, since you raised it.

Ercaner1988:
## What does this PR do?

Turns the `relative_to` containment guard into a repo-wide AST ratchet, as requested in #775.

**Why the old guard was not a lock.** `tests/test_code_review_recurring.py` asserted `".relative_to(cwd)" not in src` over seven named files. Both halves leaked:

1. *Scope* — a new module (or an eighth existing one) could reintroduce the banned idiom and the guard would never look at it.
2. *Spelling* — it matched one literal. `.relative_to(base)`, `.relative_to(root)`, a call the formatter split across lines, and `.is_relative_to(...)` all sailed past, even inside the seven files it did cover. The variable happened to be named `cwd` at the migrated sites; nothing makes the next one agree.

**What replaces it.** `tests/test_issue775_path_containment_ratchet.py` walks the AST of every module under `src/soup_cli/` (501 today) and keys on the *call*, not on the text around it. A call decides containment when:

* it is `is_relative_to` — bool-only API, it answers exactly the containment question;
* it is `relative_to` inside a `try` whose `except` can catch `ValueError` (including a bare `except:` / `except Exception:`) — that handler *is* the containment decision;
* it is `relative_to` in a boolean position — an `if`/`while`/`assert` test, `not`, `and`/`or`, a ternary condition, or a comprehension filter.

A plain `rel = p.relative_to(root)` with no `ValueError` guard is display formatting and is left alone, so the guard stays on the shapes that actually break.

**On `adapters.py:38` and `:107`.** Whether the `ValueError` handler rejects or falls back is deliberately *not* inspected: separating "deny" from "cosmetic fallback" needs a taint analysis, and the cheap heuristic misclassified `:38` in both directions while being written. Both sites are therefore in an explicit `ALLOWLIST` with the reason they decide nothing — a depth counter over `rglob` output whose `except` is unreachable for that input, and display-only shortening that falls back to the full path and reaches `table.add_row` and nothing else. A benign site costs one allowlist line with a reason; `test_the_allowlist_has_no_dead_entries` fails when an entry stops matching a live call, so a moved or fixed site cannot rot into a silent hole instead of being re-verified.

**The scan finds nothing outside that allowlist today**, which makes `TestTheScannerCanActuallyFail` load-bearing — modelled on `TestTheScannerCanActuallyFail` in `test_cli_help_assertions_are_ansi_safe.py`, as suggested in the issue. It asserts the scanner **red** against the `cwd` idiom, a differently named base, a call wrapped across lines, `is_relative_to` in a condition, a bare `is_relative_to` return, a comprehension filter and an `except Exception` handler; and **green** against `is_under`, a display-only call, a call in a branch *body* (`If.body` is not `If.test`), an `except OSError` that cannot catch the `ValueError`, a `try` around a function *definition*, a similarly named method, and unparseable source. Reported text comes from `ast.get_source_segment` with wrapping collapsed, so a split call reports the whole call rather than `relative = (`.

Verified red end to end, not just against synthetic strings: an offending module dropped into `src/soup_cli/` makes the repo-wide test fail and name the file — the exact shape the seven-file guard could not see. (Probe removed; not part of the diff.)

The seven-file check keeps its own `is_under` / `is_under_cwd` assertions and now routes its containment half through the same scanner, so it is no longer spelling-dependent.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

- [x] `ruff check src/soup_cli/ scripts/ tests/` passes (ran `src/soup_cli/ scripts/ tests/ benchmarks/`, clean)
- [ ] `pytest tests/ -v` passes — **partially run locally**, see below
- [ ] Updated relevant docs — not needed, no behaviour change
- [x] Added a changelog fragment, or this change has no user-visible impact — test-only ratchet, no user-visible impact (happy to add a fragment if you would rather have one)

### Test results

Run on Python 3.10 (project floor) and 3.12; ML extras are not installed in this environment, so the torch/MLX-dependent suites were **not** run locally and CI is relied on for those.

```
tests/test_issue775_path_containment_ratchet.py  18 passed
tests/test_code_review_recurring.py               7 passed
tests/test_cli_help_assertions_are_ansi_safe.py
tests/test_issue633_ansi_in_highlighted_output.py
tests/test_issue487_changelog_fragments.py
tests/test_issue382_windows_ci_guard_is_shared.py 50 passed, 1 skipped
```

`ruff check src/soup_cli/ scripts/ tests/ benchmarks/` — all checks passed.

Closes #777 (claimed there before starting). Refs #775 — the Rust-port SWOT half of that audit is not touched here.

### DIFF
--- a/tests/test_code_review_recurring.py
+++ b/tests/test_code_review_recurring.py
@@ -13,6 +13,9 @@ from pathlib import Path
 import pytest
 
 import soup_cli
+from tests.test_issue775_path_containment_ratchet import (
+    find_containment_relative_to,
+)
 
 
 def _src(rel: str) -> str:
@@ -34,8 +37,14 @@ def test_containment_holdouts_migrated_to_commonpath_helper():
     }.items():
         src = _src(rel)
         assert needle in src, f"{rel} not migrated to {needle}"
-        # The buggy cwd-containment idiom must be gone from these sites.
-        assert ".relative_to(cwd)" not in src, f"{rel} still uses relative_to(cwd)"
+        # The buggy containment idiom must be gone from these sites. Checked by
+        # AST rather than by the literal `".relative_to(cwd)"` this line used to
+        # search for: that text match missed `.relative_to(base)`, `(root)`, and
+        # any call the formatter split across lines (#775). The repo-wide
+        # version of this scan lives in
+        # tests/test_issue775_path_containment_ratchet.py.
+        found = find_containment_relative_to(src)
+        assert found == [], f"{rel} still decides containment with relative_to: {found}"
 
     # data.py migrated all three sample/download output-path checks.
     assert _src("commands/data.py").count("is_under_cwd(") >= 3
--- a/tests/test_issue775_path_containment_ratchet.py
+++ b/tests/test_issue775_path_containment_ratchet.py
@@ -1,0 +1,399 @@
+"""#775 — repo-wide ratchet: containment must not ride on ``relative_to``.
+
+``Path.resolve() + relative_to()`` is the containment idiom AGENTS.md bans:
+one of the two paths can carry a Windows 8.3 short name (``C:\\Users\\RUNNER~1``)
+while the other carries the long form, so ``relative_to`` raises ``ValueError``
+for a path that *is* inside the base and the guard denies a legitimate write.
+The migration to ``utils.paths.is_under`` (realpath + commonpath) is done; this
+file is the lock that keeps it done.
+
+**Why the previous guard was not a lock.** ``test_code_review_recurring.py``
+asserted ``".relative_to(cwd)" not in src`` over **seven named files**. Two
+holes, both fatal to its purpose:
+
+1. *Scope.* A new module — or an eighth existing one — could reintroduce the
+   idiom and the guard would never look at it.
+2. *Spelling.* It matched one literal. ``.relative_to(base)``,
+   ``.relative_to(root)``, ``.relative_to(\n    directory\n)`` and
+   ``.is_relative_to(...)`` all sailed past. The variable happened to be named
+   ``cwd`` at the sites that were migrated; nothing makes the next one agree.
+
+So the scan is an AST walk over ``src/soup_cli/**/*.py`` and keys on the CALL,
+not on the text around it. Every spelling of the receiver and the argument is
+the same tree.
+
+**What counts as containment** (a plain ``relative_to`` is not automatically a
+bug — it is also the normal way to shorten a path for display):
+
+* ``is_relative_to`` — always. It returns a bool and exists for no other
+  purpose than answering "is this inside that".
+* ``relative_to`` inside a ``try`` whose ``except`` can catch ``ValueError`` —
+  that handler is the containment decision. Where the answer is used (deny,
+  skip, fall back) is deliberately NOT inspected: telling "reject" from
+  "cosmetic fallback" apart needs a taint analysis, and the cheap version got
+  ``adapters.py:38`` wrong in both directions while being written. A benign
+  site is one allowlist line with a reason instead.
+* ``relative_to`` in a boolean position — ``if``/``while``/``assert`` test,
+  ``not``, ``and``/``or``, a ternary condition, a comprehension filter. Nobody
+  writes that for display.
+
+Everything else — ``rel = p.relative_to(root)`` with no ``ValueError`` guard —
+is left alone.
+
+At the time of writing the scan finds **two** sites, both in
+``commands/adapters.py``, both benign and both allowlisted below with their
+reason. There is no cleanup left, which makes ``TestTheScannerCanActuallyFail``
+load-bearing: a scanner whose only findings are allowlisted must be shown
+capable of finding something, or it is indistinguishable from one that matches
+nothing at all.
+"""
+
+from __future__ import annotations
+
+import ast
+import pathlib
+import textwrap
+
+SRC = pathlib.Path(__file__).resolve().parents[1] / "src" / "soup_cli"
+
+_CONTAINMENT_METHODS = frozenset({"relative_to", "is_relative_to"})
+
+#: Exception names whose handler catches ``ValueError``. A bare ``except:``
+#: (``handler.type is None``) counts too — it catches everything.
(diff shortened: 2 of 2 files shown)


MakazhanAlpamys:
Merged. This is a first contribution and it clears the bar by some distance — thank you.

What I checked before merging, since a guard that finds nothing today is indistinguishable from a broken one:

- **A real offender, not a synthetic string.** I wrote `src/soup_cli/utils/_mutant_probe.py` containing `path.resolve().relative_to(base.resolve())` inside a `try/except ValueError`. The scan failed and named it by file and line — exactly the shape the seven-file allowlist could not see.
- **Spelling variants.** An intermediate assignment (`resolved = P(path).resolve()`, then `.relative_to(...)` on a later line) was caught at the right line, and so was an aliased `from pathlib import Path as P` with a `PurePath` receiver. Keying on the call rather than the text is what earns that.
- **The allowlist is not a filename allowlist under a new name.** I appended a fresh offender to `commands/adapters.py` — an allowlisted *file* — and it was still reported, at its own line. `test_the_allowlist_has_no_dead_entries` closes the other half.
- **No false positives.** 501 modules, 2 hits, both the known `adapters.py` sites; I read both against source and they match your analysis. The `If.test`-vs-`If.body` distinction via `_in_field` is the call that keeps this a guard instead of noise, and your green controls cover it explicitly. A guard that fires on correct code gets deleted — this repo has that scar already.
- 25 tests in the two files, 519 passed across the ten containment-related suites, `ruff` clean, and **zero** deleted or weakened assertions: the one removed line (`assert ".relative_to(cwd)" not in src`) is replaced in place by the stronger AST check.

Two things I want to name specifically, because they are the parts people usually skip. First, you declined to classify deny-vs-cosmetic-fallback and said plainly that the cheap heuristic misclassified `adapters.py:38` **in both directions while you were writing it** — that is the honest answer, and an allowlist with a re-derivable reason is better than a taint analysis nobody can audit. Second, `TestTheScannerCanActuallyFail` is the whole reason I could merge this quickly, and you had already written it before I asked.

One follow-up, not a blocker, filed separately: `contextlib.suppress(ValueError)` is a live idiom here (four uses under `src/`, e.g. `trainer/rewards.py:449`) and catches identically to `except ValueError`, but the scanner only inspects `ast.Try`. I verified it slips through. It is one node type away — see the issue.

The cross-function case (helper extracted, caller owns the `try`) you disclaimed in both the PR body and the module docstring, and I agree that is out of scope for an AST scan.

PR #782, opened 2026-09-11
fix(inference): stop re-adding special tokens to a rendered chat template (#781)

### LINKED ISSUE #781: Inference commands re-add special tokens to a rendered chat template: models tuned on the default path are prompted with a doubled BOS
## The defect

Every in-process generation path renders the chat template to a string and then tokenizes that string with the tokenizer's default `add_special_tokens=True`:

```python
text = tokenizer.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)
inputs = tokenizer(text, return_tensors="pt")          # adds the tokenizer's own BOS (and EOS, if it appends one)
```

A template that renders `{{ bos_token }}` itself (Llama-3.x, Gemma, Mistral) on a tokenizer whose post-processor also prepends BOS therefore sends `[bos, bos, ...]`.

Soup's **default** SFT path does not do this. `data.train_on_responses_only` defaults to `true` (`config/schema.py:383`), which builds ids through `data/loss_mask.py::_tokenize_only`, i.e. `apply_chat_template(tokenize=True, add_special_tokens=False)`. Its docstring says the flag is there "so HF does not double-prepend BOS". So a model tuned on the default path is prompted at inference with a sequence it never trained on.

The project already treats the shape as a defect:
- `soup data doctor` rates it `MAJOR` (`bos_duplication`, `utils/data_doctor.py:475`).
- `utils/live_eval.py:345` (`_tokenize_pair`) already passes `add_special_tokens=False` for a rendered prompt.

TRL's own collators make the same call (`add_special_tokens=False  # to avoid adding the BOS, twice`, trl v0.29.0 `sft_trainer.py`, citing Hugging Face's [tokenizer gotchas, §7](https://huggingface.co/blog/qgallouedec/gotchas-in-tokenizer-behavior#7-chat-template-and-tokenization-dont-compose-due-to-special-tokens)).

## Measured: real tokenizers through the real call paths

Method:
- The production functions were driven with tokenizers downloaded from the Hub (tokenizer files only) and a stand-in model that records the `input_ids` handed to `generate`: `commands/chat.py::_generate`, `commands/serve.py::_generate_response`, `commands/infer.py::_generate` and `commands/diff.py::_generate`.
- Training ids come from `build_assistant_only_labels`, the default SFT path.
- Leading BOS count in each column; "fix" = the same text with `add_special_tokens=False`.

| tokenizer | training | chat | serve | infer | diff | fix | fix == HF `apply_chat_template(tokenize=True)` | fix is a prefix of the training ids |
|---|---|---|---|---|---|---|---|---|
| unsloth/Llama-3.2-1B-Instruct | 1 | **2** | **2** | **2** | **2** | 1 | yes | yes |
| unsloth/gemma-3-270m-it | 1 | **2** | **2** | **2** | **2** | 1 | yes | yes |
| unsloth/mistral-7b-instruct-v0.3 | 1 | **2** | **2** | **2** | **2** | 1 | yes | yes |
| Llama-3.2-1B + `data.chat_template: llama3` (Soup preset) | 0 | **1** | **1** | **1** | **1** | 0 | yes | yes |
| Qwen/Qwen2.5-0.5B-Instruct (control: no BOS anywhere) | 0 | 0 | 0 | 0 | 0 | 0 | yes | yes |
| deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B | 1 | 1 | 1 | 1 | 1 | 1 | yes | no* |

\*The R1 tokenizer does not add BOS in `__call__`, so it is unaffected. Its prefix check fails for an unrelated reason: the generation prompt injects `<think>\n`, which the training render of the assistant turn does not contain.

Identical on both ends of the supported range:
- transformers **5.16.1** / torch 2.6.0 (the `transformers-floor` pins)
- transformers **5.17.0** / torch 2.14.0

Both on Python 3.12.13 and tokenizers 0.23.2. The script and raw JSON are in the gist linked below.

Note the preset row: Soup's `data.chat_template` presets (`data/chat_templates.py`) render no BOS, so adapters trained with them never saw one, yet inference sends the tokenizer's. That is why the fix below is "add no special tokens to a rendered prompt", not "remove a duplicate BOS".

## Does it change behaviour?

**A model tuned on Soup's default-path ids.**

Setup:
- `unsloth/gemma-3-270m-it` + LoRA r8 on 400 synthetic rows: `Shout: <2-4 words>` → `<WORDS>!`
- 100 disjoint held-out rows, greedy decoding, float32 on CPU
- Metrics written into the script before any run
- One seed, one small model, one task

| encoding at inference | exact match (of 100) | mean NLL/token of the reference |
|---|---|---|
| today: `[bos, bos, <start_of_turn>, ...]` | 57 | 0.283 |
| as trained: `[bos, <start_of_turn>, ...]` | **63** | **0.224** |

- Greedy output differs on 33 / 100 rows.
- Discordant pairs: 8 correct only with the training-consistent prompt, 2 only with today's. Exact binomial p = 0.11, so the exact-match gap is **not significant** at this n.
- Reference NLL difference: +0.059 nats/token, paired bootstrap 95% CI [0.031, 0.088].
- A second evaluation run reproduced every row.
- Typical discordant rows (reference / today / as trained):
  - `COMET MEADOW!` / `COMET MERLOT MEADOW!` / `COMET MEADOW!`
  - `PENCIL LUCKY PLANET GREEN!` / `Pencil lucky planet green!` / exact
  - `MARBLE SWEET WARM!` / `MARBLE SWEET WARM` / exact

One deviation, recorded before evaluation:
- `soup train -y` on this CPU-only box sat in a single-threaded autograd backward for more than 4 min per step. I saw it twice, did not investigate the cause, and treat it as out of scope here.
- So the adapter was trained by a minimal loop over exactly the `build_assistant_only_labels` ids, with the settings `soup train` resolved: 368,640 trainable on q/v, lr 2e-4, 2 epochs, batch 8.

**Base model, `unsloth/gemma-3-270m-it`.**

Setup: 40 fixed factual prompts, greedy decoding.

- Greedy output differs on **15 / 40** prompts.
- Reference-answer NLL is **lower** with the doubled BOS: 1.219 vs 1.548 per token, paired difference −0.33, bootstrap 95% CI [−0.59, −0.12].

So I am not claiming the base model is better or worse either way. The references are plain sentences, and most single-BOS outputs use the model's native markdown (`**Tokyo**`).

Reading the 15 differing outputs by hand is post hoc, so weak evidence:
- 6 are correct only with one BOS: 366 days, *Gracias*, *Bonjour*, water, eight legs, Central Processing Unit.
- 1 is correct only with two BOS: diamond.
- 5 differ in style only.
- 3 are wrong both ways.

**Scripts and raw outputs** for every number above: https://gist.github.com/Konuktor/6598a2718316cc3dc98e1f8055052f62

## Where

In-process generation that renders then re-tokenizes:
- `commands/chat.py::_generate`: `soup chat`
- `commands/serve.py::_generate_response`: `soup serve`, transformers backend, streaming and non-streaming
- `commands/infer.py::_generate`: `soup infer`, and `soup bench`, which calls it
- `commands/diff.py::_generate`: `soup diff`
- `utils/mole_routing.py::LoadedMole.generate_text`: `soup serve --mole`
- `utils/live_eval.py` `make_generator` / `make_multi_generator`: the training-time eval-gate callback, `soup ship`, `soup diagnose`, `soup advise`, the behaviour battery
- `commands/generate.py::_generate_local`: `soup data generate`, local provider

Four of these (chat, infer, diff, mole) also carry their own copy of the role-prefixed fallback prompt that #332 consolidated into `utils/vllm.py` for serve.

## Proposed fix

One rule, in one place: when a chat template rendered the prompt, the tokenizer adds no special tokens. That is exactly what training does and what `apply_chat_template(tokenize=True)` does.
- `utils/vllm.py` gains `encode_chat_prompt` / `encode_rendered_prompt` next to `build_chat_prompt`, whose API is unchanged for the engines.
- The no-template and failed-render fallbacks keep today's tokenizer call **byte for byte**, since that text never went through a template.
- Each site keeps its current error behaviour: serve falls back to the legacy prompt as it does today; chat / infer / diff / mole still raise on a template that rejects the conversation. They must not start quietly serving a `System:/User:` prompt.
- The four duplicated fallback copies are deleted.
- A repo-wide AST ratchet fails on any new scope that renders a template and then lets the tokenizer add special tokens. It has an allowlist with reasons and controls that prove it can fail.

Visible changes, all toward matching training:
- `usage.prompt_tokens` drops by the removed token(s).
- A tokenizer that appends EOS no longer puts one at the end of the prompt.
- Preset-trained adapters are prompted with no BOS, as they were trained.

## Not in scope, named so it is not lost

- **`soup deploy hf-space` templates** (`commands/deploy.py`) have the same line, but the generated app never reaches it. Executed with only `gradio` / `transformers` stubbed, the rendered Gradio `respond` raises `TypeError: unhashable type: 'dict'` on the first message: the templates' `{{...}}` braces are never un-doubled. That is a separate bug; I will file it on its own.
- **Training-side:** the non-default text path (`train_on_responses_only: false`, `_legacy_text_format_row`) is tokenized by TRL's `tokenize_fn` via `processing_class(text=...)`, so it trains **with** the doubled BOS. This is read from the trl 0.29.0 and 0.29.1 source (the whole supported range), not run. `soup data preprocess` (`commands/data.py::preprocess_dataset`) does the same. Changing either changes training data and the pipeline cache key, so it deserves its own issue. Adapters trained that way currently match inference by accident, and would be off by one token after this fix until that lands.
- **Activation capture** (`live_eval.extract_layer_activations`, `utils/steering.py::_capture_attn_heads`): changing the ids changes saved steering vectors and probe calibration.
- **vLLM / SGLang / MII** tokenize the prompt string inside the engine. I cannot run them here. The one-request check is `final_output.prompt_token_ids[:2]`, which `utils/vllm.py` already reads.

## One question for you

Soup's presets render no BOS, but Llama-3, Gemma and Mistral were pretrained with one, so adapters trained through a preset never see BOS in training or, after this fix, at inference. Whether the presets should render `{{ bos_token }}` is a training-data decision, not something to settle inside an inference fix.

## Acceptance criteria

- [ ] Every in-process generation path encodes a rendered prompt without tokenizer special tokens; ids equal `apply_chat_template(tokenize=True)` and are a prefix of the training ids.
- [ ] No-template and failed-render fallbacks are byte-identical to today; strict sites still raise.
- [ ] Offline tests with a real tokenizer: one BOS on a BOS template, zero on a Soup preset, no appended EOS, train/infer prefix agreement on both `loss_mask` paths.
- [ ] A ratchet that scans `src/soup_cli/`, with a control proving it can fail.

Konuktor:
Closes #781.

## Problem / root cause

Every in-process generation path rendered the chat template to text and then called `tokenizer(text)`. That adds the tokenizer's own special tokens on top of the ones the template had already rendered.
- **Vendor templates** (Llama-3, Gemma, Mistral): the model received `[bos, bos, ...]`, while Soup's default SFT path (`train_on_responses_only: true`) trains on `add_special_tokens=False` ids. Tuned models were prompted with a sequence they never trained on.
- **Soup's `data.chat_template` presets** render no BOS, so their adapters trained with none and were prompted with one.

#781 has the real-tokenizer measurements, a tuned-model A/B and the raw data.

## What changed

- `utils/vllm.py`, next to `build_chat_prompt`, whose API is unchanged:
  - `encode_rendered_prompt(tokenizer, text, *, templated, **kwargs)`: text a template rendered is tokenized with `add_special_tokens=False`; any other text gets exactly today's call.
  - `encode_chat_prompt(messages, tokenizer, *, fallback_on_error, **kwargs)`: renders with the builder's body and passes on whether the template actually produced the text.
- Routed through them:
  - `chat._generate`
  - `serve._generate_response` (streaming and non-streaming)
  - `infer._generate` (also used by `soup bench`)
  - `diff._generate`
  - `LoadedMole.generate_text`
  - `live_eval.make_generator` / `make_multi_generator`
  - `generate._generate_local`
- The four hand-rolled copies of the legacy fallback prompt (chat, infer, diff, mole) are deleted.

## Why this implementation

- **One rule, in one place.** It yields the same ids as training and as HF's `apply_chat_template(tokenize=True)`, instead of a fifth copy of render-then-tokenize.
- **No tokenizer special tokens on rendered text**, rather than "drop a duplicate BOS". A preset (renders no BOS, tokenizer adds one) needs 1 → 0, and a startswith-BOS check would miss it.
- **`templated` means the template rendered this text**, so a failed render that falls back keeps the tokenizer's defaults.
- **`fallback_on_error` is keyword-only with no default**, so every call site keeps its existing error contract (see below).

## Tests

`tests/test_issue781_inference_bos.py` has 62 offline tests. They use a real `transformers` fast tokenizer built from an in-memory vocab with a genuine BOS/EOS post-processor; the only stand-in is the model, which records the ids passed to `generate`.
- **Encoder:** one BOS on a BOS template; ids equal `apply_chat_template(tokenize=True)`; no appended EOS; no BOS on Soup's `llama3` preset; byte-identical controls.
- **Train/infer:** prompt ids are a prefix of the `build_assistant_only_labels` ids, on both `loss_mask` paths.
- **Every call site above:** recorded ids, a no-template control each, and the error contracts.
- **AST ratchet** over `src/soup_cli/` against render-then-re-tokenize, with a reasoned allowlist and controls proving it can fail.

**Before / after:**
- On `main` the new file gives 30 failed / 32 passed. The failures are the missing helper (19), the doubled BOS at 8 call sites, serve's `prompt_tokens`, the ratchet, and the legacy-copy check.
- On this branch all 62 pass.
- **Mutation testing:** 24 mutations applied, 24 killed. They covered each encoder branch, each call site reverted to `main`, each fallback contract flipped, and a startswith-BOS rule.

```
ruff check src/soup_cli/ scripts/ tests/ benchmarks/   All checks passed!
pytest <new file + 14 affected suites>                  441 passed   transformers 5.17.0 / torch 2.14.0+cpu
                                                        441 passed   transformers-floor pins 5.16.1 / 2.6.0+cpu
pytest tests/   main e0fd219                            20796 passed, 107 skipped, 0 failed   coverage 83.10%
pytest tests/   this branch                             20858 passed, 107 skipped, 0 failed   coverage 83.21%
```

Compared test-by-test against `main`: no existing test missing or changed status, identical skip set, +62 passed (all from the new file).

**One existing test was edited:** `test_issue332_vllm_prompt.py::test_transformers_backend_calls_the_shared_builder` asserted the literal `build_chat_prompt(` in `serve.py`. It now asserts `encode_chat_prompt(`, which renders through that same builder, with a comment pointing to where serve's ids are pinned.

## Compatibility / fallbacks preserved

- **Fallbacks:** a model with no chat template, or a template that fails to render, is tokenized exactly as before.
- **Error contracts:** `soup serve` still falls back to the legacy prompt on a template error; `soup chat` / `infer` / `diff` / MoLE still raise.
- **Visible changes:**
  - BOS count goes 2 → 1 on vendor templates and 1 → 0 on Soup presets;
  - there's no trailing EOS from tokenizers that append one;
  - `usage.prompt_tokens` drops accordingly.
- **Not changed here** (tracked in #781):
  - Adapters trained with `train_on_responses_only: false` saw the doubled BOS via TRL's `tokenize_fn`, so they now differ from inference by one token.
  - vLLM / SGLang / MII tokenize inside the engine and aren't verifiable here.
- **No new dependency;** new imports are function-local.

## Changelog

`changelog.d/0.74.0/782.fixed.md`

### DIFF
--- a/changelog.d/0.74.0/782.fixed.md
+++ b/changelog.d/0.74.0/782.fixed.md
@@ -1,0 +1,15 @@
+- **Inference prompts are now encoded the way Soup trains them (#781 by @Konuktor in #782).**
+  `soup chat`, `soup serve` (transformers backend and `--mole`), `soup infer` /
+  `soup bench`, `soup diff`, the `live_eval` generators behind the eval gate,
+  `soup ship`, `soup diagnose` and `soup advise`, and `soup data generate`'s local
+  provider rendered the chat template to text and then let the tokenizer add its
+  own special tokens, so templates that render `{{ bos_token }}` (Llama-3, Gemma,
+  Mistral) sent a doubled BOS while Soup's default SFT path trains with one. A
+  rendered prompt is now encoded with no tokenizer special tokens, matching training
+  and `apply_chat_template(tokenize=True)`: 2 → 1 BOS on vendor templates, 1 → 0 on
+  Soup's `data.chat_template` presets (which render none), and no trailing EOS from
+  tokenizers that append one; `usage.prompt_tokens` drops accordingly. Models without
+  a chat template, and templates that fail to render, are encoded exactly as before.
+  Adapters trained with `train_on_responses_only: false` saw the doubled BOS in
+  training and now differ from inference by one token; that path and the vLLM /
+  SGLang / MII engines are tracked in #781.
--- a/src/soup_cli/commands/chat.py
+++ b/src/soup_cli/commands/chat.py
@@ -262,27 +262,11 @@ def _generate(
     """Generate a response from the model given message history."""
     import torch
 
-    # Apply chat template if available
-    if hasattr(tokenizer, "apply_chat_template") and tokenizer.chat_template:
-        text = tokenizer.apply_chat_template(
-            messages, tokenize=False, add_generation_prompt=True
-        )
-    else:
-        # Fallback: simple concatenation
-        parts = []
-        for msg in messages:
-            role = msg["role"]
-            content = msg["content"]
-            if role == "system":
-                parts.append(f"System: {content}")
-            elif role == "user":
-                parts.append(f"User: {content}")
-            elif role == "assistant":
-                parts.append(f"Assistant: {content}")
-        parts.append("Assistant:")
-        text = "\n".join(parts)
+    from soup_cli.utils.vllm import encode_chat_prompt
 
-    inputs = tokenizer(text, return_tensors="pt")
+    inputs = encode_chat_prompt(
+        messages, tokenizer, fallback_on_error=False, return_tensors="pt"
+    )
     input_ids = inputs["input_ids"].to(model.device)
     attention_mask = inputs["attention_mask"].to(model.device)
 
--- a/src/soup_cli/commands/diff.py
+++ b/src/soup_cli/commands/diff.py
@@ -281,25 +281,11 @@ def _generate(model, tokenizer, messages, max_tokens=256, temperature=0.7) -> st
     """Generate a response from the model."""
     import torch
 
-    if hasattr(tokenizer, "apply_chat_template") and tokenizer.chat_template:
-        text = tokenizer.apply_chat_template(
-            messages, tokenize=False, add_generation_prompt=True
-        )
-    else:
-        parts = []
-        for msg in messages:
-            role = msg["role"]
-            content = msg["content"]
-            if role == "system":
-                parts.append(f"System: {content}")
-            elif role == "user":
-                parts.append(f"User: {content}")
-            elif role == "assistant":
-                parts.append(f"Assistant: {content}")
-        parts.append("Assistant:")
-        text = "\n".join(parts)
+    from soup_cli.utils.vllm import encode_chat_prompt
 
-    inputs = tokenizer(text, return_tensors="pt")
+    inputs = encode_chat_prompt(
+        messages, tokenizer, fallback_on_error=False, return_tensors="pt"
+    )
     input_ids = inputs["input_ids"].to(model.device)
     attention_mask = inputs["attention_mask"].to(model.device)
 
--- a/src/soup_cli/commands/generate.py
+++ b/src/soup_cli/commands/generate.py
@@ -773,6 +773,7 @@ def _generate_local(
         model_requires_trust_remote_code,
         resolve_trust_remote_code,
     )
+    from soup_cli.utils.vllm import encode_rendered_prompt
 
     requires = model_requires_trust_remote_code(model_name) or False
     trc = resolve_trust_remote_code(
@@ -797,7 +798,8 @@ def _generate_local(
     if generation_prompt is None:
         generation_prompt = _build_generation_prompt(prompt, count, fmt, seed_examples)
 
-    if hasattr(tokenizer, "apply_chat_template") and tokenizer.chat_template:
+    templated = bool(hasattr(tokenizer, "apply_chat_template") and tokenizer.chat_template)
+    if templated:
         messages = [
             {"role": "system", "content": generation_prompt},
             {"role": "user", "content": f"Generate {count} training examples now."},
@@ -806,7 +808,7 @@ def _generate_local(
     else:
         text = f"{generation_prompt}\n\nGenerate {count} training examples now.\n\n"
 
-    inputs = tokenizer(text, return_tensors="pt")
+    inputs = encode_rendered_prompt(tokenizer, text, templated=templated, return_tensors="pt")
     input_ids = inputs["input_ids"].to(model.device)
 
     with torch.no_grad():
--- a/src/soup_cli/commands/infer.py
+++ b/src/soup_cli/commands/infer.py
@@ -717,25 +717,11 @@ def _generate(
     """Generate a response from the model. Returns (text, token_count)."""
     import torch
 
-    if hasattr(tokenizer, "apply_chat_template") and tokenizer.chat_template:
-        text = tokenizer.apply_chat_template(
-            messages, tokenize=False, add_generation_prompt=True
-        )
-    else:
-        parts = []
-        for msg in messages:
-            role = msg["role"]
-            content = msg["content"]
-            if role == "system":
-                parts.append(f"System: {content}")
-            elif role == "user":
-                parts.append(f"User: {content}")
-            elif role == "assistant":
-                parts.append(f"Assistant: {content}")
-        parts.append("Assistant:")
-        text = "\n".join(parts)
+    from soup_cli.utils.vllm import encode_chat_prompt
 
-    inputs = tokenizer(text, return_tensors="pt")
+    inputs = encode_chat_prompt(
+        messages, tokenizer, fallback_on_error=False, return_tensors="pt"
+    )
     input_ids = inputs["input_ids"].to(model.device)
     attention_mask = inputs["attention_mask"].to(model.device)
 
--- a/src/soup_cli/commands/serve.py
+++ b/src/soup_cli/commands/serve.py
@@ -1573,13 +1573,14 @@ def _generate_response(
     """Generate a response from the model."""
     import torch
 
-    from soup_cli.utils.vllm import build_chat_prompt
+    from soup_cli.utils.vllm import encode_chat_prompt
 
     # Apply chat template. #332 — THE shared builder; the vLLM backend calls
-    # the same function so the two backends cannot drift apart again.
-    text = build_chat_prompt(messages, tokenizer)
-
-    inputs = tokenizer(text, return_tensors="pt")
+    # the same function so the two backends cannot drift apart again. #781 —
+    # encoded without re-adding the special tokens the template rendered.
+    inputs = encode_chat_prompt(
+        messages, tokenizer, fallback_on_error=True, return_tensors="pt"
+    )
     input_ids = inputs["input_ids"].to(model.device)
     attention_mask = inputs["attention_mask"].to(model.device)
 
--- a/src/soup_cli/utils/live_eval.py
+++ b/src/soup_cli/utils/live_eval.py
@@ -123,22 +123,32 @@ def _build_quantization_config(quantization: Optional[str]):
     )
 
 
-def _apply_prompt_template(tokenizer: object, prompt: str) -> str:
-    """Render a single user turn through the tokenizer's chat template.
+def _render_prompt_template(tokenizer: object, prompt: str) -> Tuple[str, bool]:
+    """Render a single user turn; also report whether the chat template produced it.
 
-    Falls back to the raw prompt when the tokenizer has no chat template.
+    Falls back to the raw prompt when the tokenizer has no chat template or the
+    template fails to render.
     """
     chat_template = getattr(tokenizer, "chat_template", None)
     if chat_template:
         try:
-            return tokenizer.apply_chat_template(  # type: ignore[attr-defined]
+            text = tokenizer.apply_chat_template(  # type: ignore[attr-defined]
                 [{"role": "user", "content": prompt}],
                 tokenize=False,
                 add_generation_prompt=True,
             )
         except Exception:  # noqa: BLE001 — malformed template → raw prompt
-            return prompt
-    return prompt
+            return prompt, False
+        return text, True
+    return prompt, False
+
+
+def _apply_prompt_template(tokenizer: object, prompt: str) -> str:
+    """Render a single user turn through the tokenizer's chat template.
+
+    Falls back to the raw prompt when the tokenizer has no chat template.
+    """
+    return _render_prompt_template(tokenizer, prompt)[0]
 
 
 def load_model_and_tokenizer(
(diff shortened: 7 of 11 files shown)


MakazhanAlpamys:
Reviewed in full. **This is a merge** — I am waiting only on CI, which I approved for you (first-time contributors need a manual unblock here).

The changelog fragment was the one thing missing when I started reading; you pushed `changelog.d/0.74.0/782.fixed.md` while the review was running, and it is accurate down to the "1 → 0 on Soup's own presets" case, so nothing is outstanding.

**Scope matches the issue.** #781 names seven in-process render-then-re-tokenize sites; the diff touches exactly those seven plus the shared helper, and leaves `build_chat_prompt`'s public API alone for the vLLM / SGLang / MII callers that tokenize in-engine (`utils/vllm.py:395`, `utils/sglang.py:244`, `utils/mii.py:163`). No scope creep in either direction.

**Deleted assertions: exactly one**, `tests/test_issue332_vllm_prompt.py:169` changing `build_chat_prompt(` to `encode_chat_prompt(`, and the #332 invariant survives because both names funnel through the same `_render_chat_prompt` body.

**Tests.** 62 in the new file; 388 across the fourteen affected suites; 565 passed / 5 skipped across the downstream consumers (`live_eval`, mole, advise, diagnose, ship, bench, behavior, tunability); 43 across the startup-is-light and ANSI ratchets, confirming the new imports stay function-local. `ruff check src/soup_cli/ tests/` clean.

**Three mutations, all killed** — I ran these myself rather than taking your 24-mutation audit on trust:

| Mutation | Result |
|---|---|
| `encode_rendered_prompt` always `tokenizer(text, **kw)` (i.e. pre-fix behaviour) | 23 failed |
| `_render_chat_prompt` fallback returns `templated=True` | 9 failed |
| `chat.py:268` `fallback_on_error=False` → `True` | 1 failed, and it is the right one — `test_a_template_that_rejects_the_conversation_still_raises[chat]` |

**Completeness.** I swept every `.generate(` call site: `best_of_n.py:70` already goes through `apply_chat_template(tokenize=True)`, and `draft.py:548`, `quant_check.py:181`, `cmaes_merge.py:476` tokenize raw text where the tokenizer defaults are correct. Nothing identical is left one file over — which is usually where this class of fix stops.

## What stood out

The measurement discipline in #781, honestly. Real tokenizers on both ends of the supported transformers range, a pre-registered metric, a bootstrap CI, raw data in a gist — and **you published the result that argues against your own thesis** (base-model NLL lower with the doubled BOS) instead of leaving it out. That is the part that made the rest of the issue credible enough to act on quickly.

On the code: one rule in one place rather than a fifth copy of the prompt builder, four duplicated legacy builders deleted, and `fallback_on_error` made keyword-only with no default so every call site is forced to state its own error contract. The AST ratchet is bidirectional with an allowlist carrying reasons and a non-vacuity control, which is the shape this repo has learned to insist on. You also named the out-of-scope work — TRL's training-side `tokenize_fn`, activation capture, the hf-space templates — rather than quietly widening the diff.

## Two non-blocking notes

- The ratchet's docstring names its cross-function blind spot but not its string-literal one: `commands/deploy.py:45,123` carry the same line inside hf-space templates. #781 declares that out of scope with a reason, so this is a docstring line at most.
- `encode_rendered_prompt` would raise `TypeError` if a caller ever passed `add_special_tokens=` in `**kwargs` with `templated=True`. No caller does today; a one-line guard would be belt-and-braces.

Neither changes the merge. Thank you — this is an unusually strong first contribution.

MakazhanAlpamys:
Merged as 95f5604f — thank you, @Konuktor, and welcome. You are now in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md).

This is a first contribution that fixes a defect every user of the default SFT path was exposed to without being able to see it: a tuned Llama-3 / Gemma / Mistral model was prompted with a doubled BOS it never trained on, across `chat`, `serve`, `infer` / `bench`, `diff`, `--mole`, `data generate` and the `live_eval` generators behind `ship`, `diagnose` and `advise`.

What I would point other contributors at:

- **You chose the rule, not the symptom.** "Text a template rendered gets no tokenizer special tokens" also covers Soup's own `data.chat_template` presets, which need 1 -> 0. A "drop a duplicate BOS" check would have fixed the vendor case and silently left the presets wrong.
- **`fallback_on_error` is keyword-only with no default.** None of the seven call sites can quietly inherit someone else's error contract, and one of my mutations below is only killable because of that.
- **You deleted the four hand-rolled fallback prompts instead of fixing them a fifth time**, and left `build_chat_prompt`'s API alone for the in-engine vLLM / SGLang / MII callers. Scope matched #781 exactly.
- **The tests use a real fast tokenizer with a genuine BOS/EOS post-processor**; the model is the only stand-in. 30 of the 62 fail on `main`, and the train/infer prefix property ties inference to what `build_assistant_only_labels` actually trains on.

What I ran before merging, independently of your 24-mutation audit:

- 62 new tests, 388 across the fourteen affected suites, 565 passed / 5 skipped across downstream consumers (`live_eval`, mole, advise, diagnose, ship, bench, behavior, tunability), plus the startup-is-light and ANSI ratchets.
- Encoder reverted to pre-fix behaviour: **23 failed**. Fallback marked `templated=True`: **9 failed**. `chat.py` `fallback_on_error` flipped: **1 failed**, the right test.
- All 13 required checks green on `49628c95`, the head I reviewed.

The `train_on_responses_only: false` path and the in-engine backends are now tracked in #785, built from your own "not covered" section of #781.

PR #786, opened 2026-09-11
fix(deepspeed): forward model parameter in empty param group guard (#784)

### LINKED ISSUE #784: attach_empty_param_group_guard drops create_optimizer's model parameter, so the positional call transformers uses raises TypeError
## What

`attach_empty_param_group_guard` (`src/soup_cli/utils/deepspeed.py:413`) replaces `trainer.create_optimizer` with a wrapper that takes **no arguments**:

```python
def create_optimizer():
    optimizer = original()
    ...
trainer.create_optimizer = create_optimizer
```

The method it shadows is `transformers.Trainer.create_optimizer(self, model=None)`, and `transformers` calls it **positionally** on one path — `self.create_optimizer(model)` inside `_prepare_for_training`, in the `delay_optimizer_creation` branch.

Reproduced against the installed transformers 5.17.0 with a trainer stub:

```
positional call: TypeError -> attach_empty_param_group_guard.<locals>.create_optimizer()
                              takes 0 positional arguments but 1 was given
```

## Why this is filed as hardening rather than as a bug

**It is not reachable today, and I would rather say that than dress it up.** Every one of the eighteen wrappers attaches the guard only under `if self.deepspeed_config:`, and `delay_optimizer_creation` is set for FSDP, FSDP-XLA and SageMaker MP — not for DeepSpeed. So the two conditions do not currently co-occur.

What makes it worth closing anyway is that the reachability argument lives entirely outside this function. It depends on a `transformers` internal, on eighteen call sites each keeping their `if`, and on DeepSpeed and FSDP never being enabled together. Any one of those can change without anyone looking at `utils/deepspeed.py`, and the failure mode is a `TypeError` inside optimizer construction rather than anything that names the guard.

Discovered while verifying an unrelated claim on #745 (that the guard stops pruning when LoRA+ pre-assigns `trainer.optimizer` — it does not; the guard prunes the preset optimizer correctly, which is the other half of the same experiment).

## Proposed fix path

Accept and forward the parameter:

```python
def create_optimizer(model=None):
    optimizer = original(model)
    ...
```

`original` is the bound method, so forwarding is direct. Keep `functools.wraps` off or on as you prefer, but the signature is the point.

## Acceptance criteria

- [ ] `trainer.create_optimizer(model)` and `trainer.create_optimizer(model=model)` both work after the guard is attached, and both still prune empty groups.
- [ ] `trainer.create_optimizer()` keeps working unchanged.
- [ ] The `model` argument is **forwarded** to the original, not swallowed — a test that fails if the wrapper calls `original()` with the argument dropped. (Dropping it silently would build the optimizer against the wrong model on the delayed-creation path, which is worse than the `TypeError` this replaces.)
- [ ] A test asserting the wrapper's signature accepts the positional form, so a future rewrite cannot regress it without going red.
- [ ] `tests/test_issue359_deepspeed_guard_coverage.py` still passes unchanged.

Small, self-contained, and the surrounding test file shows the pattern. Good first issue.

DYNOSuprovo:
## Summary

Fixes #784 — `attach_empty_param_group_guard` shadows `transformers.Trainer.create_optimizer(self, model=None)`. Previously the wrapper took zero arguments, which would raise `TypeError: ... takes 0 positional arguments but 1 was given` if called on the `delay_optimizer_creation` path (`self.create_optimizer(model)`).

## Changes

- In `src/soup_cli/utils/deepspeed.py`:
  - Update `create_optimizer` inside `attach_empty_param_group_guard` to accept `model=None`.
  - Forward `model` to `original(model)` when provided (`model is not None`), and call `original()` without arguments when `model is None` to preserve compatibility with zero-arg stubs.
  - Retain empty parameter group pruning behavior.

- In `tests/test_issue359_deepspeed_guard_coverage.py`:
  - Add `TestGuardForwardsModelParameter` covering:
    - Positional `model` forwarding (`trainer.create_optimizer(sentinel)`)
    - Keyword `model` forwarding (`trainer.create_optimizer(model=sentinel)`)
    - Zero-arg calls (`trainer.create_optimizer()`)
    - Positional calls with empty param group pruning
    - Signature inspection verifying `create_optimizer` accepts positional-or-keyword `model`

## Verification

- `pytest tests/test_issue359_deepspeed_guard_coverage.py` passes completely (88 passed).
- `ruff check src/soup_cli/utils/deepspeed.py tests/test_issue359_deepspeed_guard_coverage.py` passes with zero errors.

### DIFF
--- a/src/soup_cli/utils/deepspeed.py
+++ b/src/soup_cli/utils/deepspeed.py
@@ -410,8 +410,11 @@ def attach_empty_param_group_guard(trainer) -> bool:
     if not callable(original):
         return False
 
-    def create_optimizer():
-        optimizer = original()
+    def create_optimizer(model=None):
+        if model is not None:
+            optimizer = original(model)
+        else:
+            optimizer = original()
         dropped = prune_empty_param_groups(optimizer)
         if dropped:
             from rich.console import Console
--- a/tests/test_issue359_deepspeed_guard_coverage.py
+++ b/tests/test_issue359_deepspeed_guard_coverage.py
@@ -223,6 +223,107 @@ class TestGuardToleratesTrainersWithoutCreateOptimizer:
         assert len(optimizer.param_groups) == 1  # the empty group is gone
 
 
+class TestGuardForwardsModelParameter:
+    """Issue #784 — the wrapper must accept and forward the ``model`` kwarg.
+
+    ``transformers.Trainer.create_optimizer(self, model=None)`` is called
+    positionally on the ``delay_optimizer_creation`` path:
+    ``self.create_optimizer(model)``.  The guard wrapper must accept that
+    call form and forward ``model`` to the original, not swallow it.
+    """
+
+    def test_positional_model_is_forwarded(self):
+        from soup_cli.utils.deepspeed import attach_empty_param_group_guard
+
+        sentinel = object()
+
+        class _Trainer:
+            def __init__(self):
+                self.received_model = None
+
+            def create_optimizer(self, model=None):
+                self.received_model = model
+                return types.SimpleNamespace(param_groups=[{"params": [1]}])
+
+        trainer = _Trainer()
+        attach_empty_param_group_guard(trainer)
+        trainer.create_optimizer(sentinel)
+        assert trainer.received_model is sentinel
+
+    def test_keyword_model_is_forwarded(self):
+        from soup_cli.utils.deepspeed import attach_empty_param_group_guard
+
+        sentinel = object()
+
+        class _Trainer:
+            def __init__(self):
+                self.received_model = None
+
+            def create_optimizer(self, model=None):
+                self.received_model = model
+                return types.SimpleNamespace(param_groups=[{"params": [1]}])
+
+        trainer = _Trainer()
+        attach_empty_param_group_guard(trainer)
+        trainer.create_optimizer(model=sentinel)
+        assert trainer.received_model is sentinel
+
+    def test_no_model_still_works(self):
+        from soup_cli.utils.deepspeed import attach_empty_param_group_guard
+
+        class _Trainer:
+            def __init__(self):
+                self.received_model = "UNSET"
+
+            def create_optimizer(self, model=None):
+                self.received_model = model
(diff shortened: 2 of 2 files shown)


MakazhanAlpamys:
Merged as 8d5d49a7 — thank you, @DYNOSuprovo, and welcome. You are now in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md).

What made this easy to merge:

- **You met #784's acceptance criteria one by one, and you departed from its suggested fix for the right reason.** The issue proposed `original(model)` unconditionally; that breaks `test_a_real_create_optimizer_is_still_wrapped`, whose stub takes no arguments, and the issue required that file to keep passing unchanged. Splitting on whether a model was passed is what satisfies both, and you noticed without being told.
- **The `sentinel is received_model` identity check** is the assertion that catches the worse failure the issue called out: a wrapper that accepts `model` and silently drops it. A test that only asserted "no `TypeError`" would pass that bug.

What I ran before merging:

- 155 passed across the #359 guard file and every `tests/*deepspeed*` suite; no test lines removed.
- Against a real transformers 5.17.0 `Trainer` with an empty no-decay group and a *different* model passed in: on `main` the positional call raises `TypeError ... takes 0 positional arguments but 1 was given` and the keyword call raises `unexpected keyword argument 'model'`; with your change both build the optimizer from the model passed in, and the no-arg call still builds from `trainer.model`.
- Mutations: restoring the zero-arg wrapper fails 4 of your 5 new tests; accepting `model` but calling `original()` fails 3.
- All 13 required checks green on `bfc7602f`, the head I reviewed.

Two notes for next time, neither a blocker:

- **Please comment on an issue before starting it.** #784 had no claim when this opened. CONTRIBUTING.md's "Claiming an Issue" section explains why — it is how two people avoid building the same fix in parallel.
- Your fakes use `create_optimizer(self, model=None)`, which does match transformers today. A test that compares the wrapper's parameters with `inspect.signature(transformers.Trainer.create_optimizer)` (skipped when transformers is absent) would turn an upstream signature change red instead of silent. Happy to take that as a follow-up PR if you want it.

PR #787, opened 2026-09-11
docs(benchmarks): publish pinned-store accounting for #655

### LINKED ISSUE #655: Measure where the pinned layer-streaming store is actually accounted (shmem vs anon) — the #622 mechanism is inferred, not verified
The #622 fix rests on a reading of *where* the layer-streaming pinned host store is accounted, and that reading is an **inference that nobody has been able to test on hardware**. It is consistent with every number in the thread and it is not verified. This issue exists so it is either confirmed or overturned on a machine, rather than resting on a converging argument.

## What is established, and what is not

**Verified in code** (@umran666 on #622, greppable rather than arguable): the store is allocated exclusively via `torch.empty(..., device="cpu", pin_memory=True)` — the CUDA host allocator, `cudaHostAlloc`. There is no `shm_open`, no `multiprocessing.shared_memory`, and no tmpfs-backed file anywhere in the streaming path.

**Observed** (reporter's kernel OOM dumps on #622, torch 2.10, sm_120, kernel 7.0): `shmem-rss` 11.65-13.99 GB against `anon-rss` under 300 MB. The pinned store was accounted in the **shmem** class, not the anon class.

**Inferred, and this is the load-bearing part**: the NVIDIA driver backs large pinned host allocations with kernel-internal shmem objects — accounted as `Shmem`/`shmem-rss`, **not bound to any tmpfs mount** (which is why `df /dev/shm` stayed at 0% during the reporter's runs and why the mount's size could not have been the bound), and unreclaimable, hence the OOM killer rather than a graceful allocation failure.

If plain `cudaHostAlloc` pages were anonymous mlocked memory, the dumps would have shown the opposite split. So something has to explain the accounting; driver-internal shmem is the explanation that fits, and it has not been measured.

## Why it matters even though the shipped fix does not depend on it

The fix in #644 is deliberately **mechanism-agnostic**: store + resident extras against a fraction of total physical RAM, alongside the existing dynamic free-RAM check. That bound is correct under either accounting, which is why it shipped without this question being closed — and two independent implementations (#644, #646) converged on it.

What *does* depend on the inference is the decision **not** to add an `os.statvfs("/dev/shm")` check. The case against statvfs is that it measures a mount which is not the allocation channel, and it is wrong in both directions: a false positive on a large host with a small custom mount (forced to disk needlessly) and a false negative on a `--shm-size=64G` container on a small host. **If the experiment below comes back the other way, that decision reopens with evidence** — and I would rather that be findable in an open issue than buried in a closed PR.

## The experiment

Ten lines. Needs a **Linux host with CUDA** — nobody in the #622 thread has one (the reporter has WSL2, the analyst and I are on Windows).

```python
import torch
# BEFORE:
#   grep -E 'RssAnon|RssShmem' /proc/self/status
#   grep -E 'Shmem|Mlocked|Unevictable' /proc/meminfo
#   df /dev/shm
t = torch.empty(4 * 1024**3, dtype=torch.uint8, device="cpu", pin_memory=True)
t.fill_(1)   # fault every page in
# AFTER: read the same five numbers again
```

**Predicted under the current reading:**

- `RssShmem` climbs by roughly 4 GB
- `RssAnon` roughly unchanged
- `Unevictable` / `Mlocked` move little
- **`/dev/shm` usage unchanged**

**What would overturn it:** `/dev/shm` usage climbing by roughly 4 GB. That would mean the tmpfs mount really is the allocation channel on that stack, the mount's size really can bound the store, and a `statvfs` check belongs in the pre-flight after all.

Either result is a useful answer. A null result — numbers that match neither prediction — is also worth posting, because it would mean the accounting depends on something neither reading models.

## Acceptance criteria

- [ ] The five readings, before and after, from a Linux + CUDA host, with `torch`, driver and kernel versions stated.
- [ ] A statement of which prediction the numbers support, or that they support neither.
- [ ] If `/dev/shm` moves: reopen the statvfs question on #622 with this data, and say what the correct bound is.
- [ ] If it does not move: the inference becomes a measured fact and the `55%`-of-total-RAM ceiling's rationale can cite it instead of citing an argument.
- [ ] No code change is required by this issue. It is a measurement.

Worth adding: if you have such a box, please also report whether `stream_source: auto` on a store sized near 55% of total RAM behaves as the pre-flight predicts. That is a second, independent check on the shipped ceiling and costs one extra run.

Analysis and the experiment design are @umran666's, from [their comment on #622](https://github.com/MakazhanAlpamys/Soup/issues/622#issuecomment-5510190940). Filed by me so it does not disappear when #646 closes.

tristangrech:
## What does this PR do?

Publishes the L4 pinned-memory measurement from #655 in `benchmarks/`, as requested in the [maintainer's reply](https://github.com/MakazhanAlpamys/Soup/issues/655#issuecomment-5629542226).

- Adds `gate-655-pinned-store-accounting.md` with the original table, environment, exact probe and invocation commands, including the host-counter drift and stated limitations.
- Adds one paragraph explaining why the result supports omitting a `/dev/shm` `statvfs` preflight on the tested stack.
- Adds the measurement to `benchmarks/README.md`.

The [original result](https://github.com/MakazhanAlpamys/Soup/issues/655#issuecomment-5623582962) is preserved. This PR publishes existing evidence; it does not change runtime code or claim a new GPU run.

Closes #655.

## Type of change

- [x] Documentation

## Validation

- Compared the record with the original GitHub comment and all 42 table readings with the three saved raw JSON files.
- Verified the environment, process deltas, cgroup corroboration and absence of recorded OOM events against those files.
- Confirmed the embedded probe is byte-identical to the original and matches SHA-256 `44bba1b76a9b41e927fe59666104bd44fcce63288f4af48887736118f7111c11`; its Python syntax parses successfully.
- Checked the README link and staged diff whitespace.
- Attempted `python3 -m pytest tests/ -v`; it could not start because this local Python environment has no `pytest`. The full suite has not been run locally.

## Checklist

- [x] `ruff check src/soup_cli/ scripts/ tests/ benchmarks/` passes (Ruff 0.16.7)
- [x] Updated the relevant benchmark documentation and index
- [x] No user-visible runtime impact; no changelog fragment needed

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -16,6 +16,7 @@ They are the evidence behind the preprint:
 
 | File | What it gates | Headline |
 |---|---|---|
+| [`gate-655-pinned-store-accounting.md`](gate-655-pinned-store-accounting.md) | Pinned host-memory accounting and the `/dev/shm` preflight decision ([#655](https://github.com/MakazhanAlpamys/Soup/issues/655)) | Two pinned 4 GiB allocations on an L4/Linux/CUDA stack each add exactly 4 GiB to `RssShmem` without using `/dev/shm`; the pageable control adds 4 GiB to `RssAnon`. Preserves the original host-counter drift and probe. |
 | [`gate-v0.72.0-layer-streaming.md`](gate-v0.72.0-layer-streaming.md) | The streaming path itself | Bit-exactness vs a resident reference; 3B bf16 trained on a 4 GB card |
 | [`gate-v0.72.2-nf4.md`](gate-v0.72.2-nf4.md) | NF4 quantised streaming | Llama-3.1-8B at 119.6 tok/s in a 3.32 GB peak |
 | [`gate-v0.72.3-breadth.md`](gate-v0.72.3-breadth.md) | Nine architectures, batching, accumulation, resume, disk tier | Peak-VRAM predictor at 0.85% worst-case error; accumulation is per-token I/O-neutral |
--- a/benchmarks/gate-655-pinned-store-accounting.md
+++ b/benchmarks/gate-655-pinned-store-accounting.md
@@ -1,0 +1,199 @@
+# Pinned host-memory accounting (#655)
+
+Measured by [Tristan Grech](https://github.com/tristangrech) on 2026-09-10 UTC.
+The [original issue comment](https://github.com/MakazhanAlpamys/Soup/issues/655#issuecomment-5623582962)
+is reproduced below without edits, including the host-counter drift, following the
+[maintainer's request](https://github.com/MakazhanAlpamys/Soup/issues/655#issuecomment-5629542226).
+
+For [#655](https://github.com/MakazhanAlpamys/Soup/issues/655), these measurements
+support keeping the `/dev/shm` `statvfs` preflight out of the pinned-store path:
+both pinned allocations increased `RssShmem` by exactly 4 GiB while leaving the
+mount's usage unchanged. The mount's free space is therefore not a capacity limit
+for these allocations on the tested stack, nor a substitute for host and cgroup
+headroom checks. This does not establish behavior on other stacks or guarantee
+that a full streaming run will fit.
+
+## Original measurement record
+
+The accounting prediction is reproduced on a Linux/CUDA RunPod container. Two fresh-process pinned runs each added exactly **4 GiB to `RssShmem`**, with no change to `RssAnon` or `/dev/shm` usage. The pageable control added exactly **4 GiB to `RssAnon`**, with no change to `RssShmem`.
+
+Environment: NVIDIA L4; driver `570.195.03`; PyTorch `2.9.1+cu128`; CUDA runtime `12.8`; Python `3.12.3`; Ubuntu 24.04.3 container; Linux host kernel `6.14.0-33-generic`. Measurements taken on 2026-09-10 UTC. Neither `PYTORCH_ALLOC_CONF` nor `PYTORCH_CUDA_ALLOC_CONF` was set.
+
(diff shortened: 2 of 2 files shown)


MakazhanAlpamys:
Merged as 06456bce — thank you, @tristangrech, and welcome. You are now in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md).

The L4 measurement was the thing nobody in the #622 thread could supply, and this record publishes it the way `benchmarks/` is supposed to work — as the working record, not a summary written afterwards:

- **The pageable control is what makes the result conclusive.** The same 4 GiB lands in `RssAnon` unpinned and in `RssShmem` pinned, so the accounting difference is attributable to pinning rather than to the allocator or the probe.
- **You kept the host-wide `Shmem` drift** (+32 / +160 KiB over the allocation) instead of rounding it away, and leaned on per-process and cgroup counters where host counters are noisy.
- **The conclusion is scoped to what the data shows** — `statvfs` on `/dev/shm` is not a capacity bound on this stack — with no claim about other stacks or a full streaming run. It is, if anything, narrower than my reply in #655.
- You claimed #655 before measuring, and the probe refuses to run without 2x headroom.

What I checked before merging:

- The record matches your original comment in #655 byte for byte, apart from a trailing newline. The embedded probe hashes to the stated `44bba1b7…7111c11` and parses. The `RssShmem` / `RssAnon` deltas recompute from the table.
- The hardware card is present, the `benchmarks/README.md` row matches the existing format, and every link resolves. No Cyrillic, no internal wording. `test_no_foreign_license_headers` and the existing gate-record discoverability test pass.
- All 13 required checks green on `a3ec9e69`, the head I reviewed.

Small things for a follow-up if you want them, none worth holding this for: one sentence saying the probe imports torch alone, so no Soup commit is involved (readers will look for one); the GPU memory size on the card; the raw JSON files that back the `Pss_Shmem` and cgroup figures quoted in prose; and moving the index row to the bottom, since that table is in order of addition.

Issue #785, opened 2026-09-11
train_on_responses_only: false still trains on a doubled BOS, now off by one token from inference after #782


MakazhanAlpamys:
Follow-up to #781 / #782. Analysis is @Konuktor's, from the "not covered" section of #781; this issue exists so that work is not left inside a closed issue.

## What #782 fixed

Every **in-process** inference path (`chat`, `serve` transformers + `--mole`, `infer` / `bench`, `diff`, `data generate` local, the `live_eval` generators) now encodes rendered chat-template text with `add_special_tokens=False`, matching Soup's default SFT path (`train_on_responses_only: true`) and `apply_chat_template(tokenize=True)`.

### 1. Training on the non-default text path still bakes in the doubled BOS

With `train_on_responses_only: false`, `_legacy_text_format_row` produces rendered text that TRL's `tokenize_fn` tokenizes via `processing_class(text=...)` — i.e. **with** the tokenizer's special tokens on top of the template's. So those adapters train on `[bos, bos, ...]`. Before #782 that matched inference by accident; after #782 inference sends one BOS, so these adapters are now **off by one token** at inference.

`soup data preprocess` (`commands/data.py::preprocess_dataset`) tokenizes the same way.

This was read from the trl 0.29.0 / 0.29.1 source (the supported range), not executed.

**Why it needs its own change:** fixing it changes training data and therefore the pipeline cache key, so previously cached datasets must not be silently reused.

### 2. vLLM / SGLang / MII are unverified

These engines take the prompt string and tokenize inside the engine, so #782 deliberately left `build_chat_prompt` unchanged for them. Whether they add a second BOS is unmeasured — nobody involved had the hardware.

## Proposed fix path

1. Encode the legacy text path's rendered template text with no tokenizer special tokens (pre-tokenize before TRL, or pass the equivalent option through), in both the SFT row builder and `soup data preprocess`; include the change in the preprocess cache key so stale caches are invalidated rather than reused.
2. For each in-engine backend, one request on a template that renders `{{ bos_token }}` and inspect `final_output.prompt_token_ids[:2]` (already read in `utils/vllm.py`). If doubled, pass the engine's equivalent of "no special tokens" or hand it token ids.

## Acceptance criteria

[...]

PR #788, opened 2026-09-11
fix(data): stop the text SFT path from training on a doubled BOS (#785)

abdulwaarith0:
Part 1 of #785. Part 2 (the vLLM / SGLang / MII `prompt_token_ids[:2]` measurement) will follow as a separate PR, so I'm leaving #785 open.

## The bug

The default SFT path (`train_on_responses_only: true`) builds ids with `add_special_tokens=False` in `data/loss_mask.py`, so it never doubled the BOS. The opt-out text path did: `data/sft_format.py` handed TRL a `{"text"}` column, and TRL 0.29.1's language-modeling `tokenize_fn` (`processing_class(text=...)`) re-tokenised it with the tokenizer's default `add_special_tokens=True`. A template that renders `{{ bos_token }}` (Llama-3, Gemma, Mistral) on a tokenizer whose post-processor also prepends BOS therefore trained on `[bos, bos, ...]`. Before #782 that matched inference by accident; after #782 inference sends one BOS, so these adapters are one token off from how they are prompted. `soup data preprocess` tokenised the same way.

I reproduced it on a real fast tokenizer first: the text path tokenised to `[1, 1, ...]` (two BOS) against `[1, ...]` (one) for `apply_chat_template(tokenize=True)`.

## The fix

The text path now pre-tokenises with `add_special_tokens=False` (`build_full_sequence_labels`), so the chat template is the single source of special tokens, matching inference and `apply_chat_template(tokenize=True)`: one BOS for a `{{ bos_token }}` template, zero for the `data.chat_template` presets. `soup data preprocess` de-duplicates the same leading BOS but otherwise tokenises exactly as `main` (`add_special_tokens=True`, so HF still reserves truncation room for the post-processor's specials and a truncated row still ends on its stop token), dropping only the one duplicated leading BOS; its cache key is bumped so datasets tokenised under the old encoding are re-tokenised rather than reused.

### EOS matches TRL's rule, not the post-processor

On `main` the legacy text rows reached TRL as `{"text": ...}`, and TRL 0.29.1's `add_eos` (`sft_trainer.py:1026-1031`) appends `eos_token` as a string to any row not already ending with it, **before** tokenizing and **independent of the post-processor**. Pre-tokenising skips that step, so `build_full_sequence_labels` reproduces the rule in token space via `append_training_eos`: append the EOS id unless the sequence already ends on it. (An earlier revision of this PR probed the post-processor instead, which under-appended for BOS-only and Qwen-shaped tokenizers and dropped the stop token `main` trained on, 1 -> 0 — thanks @AmirF194 / maintainer review for catching that.) Per case, measured against TRL's real `add_eos` baseline:

- template renders no EOS, post-processor appends none (Qwen / Gemma / Llama-2): `main` trained on one EOS via `add_eos`; the fix keeps it. **1 -> 1.**
- template renders no EOS, post-processor also appends one: `main` had a duplicated EOS; the fix ends on exactly one. 2 -> 1.
- template renders the EOS itself: carried through, nothing re-appended since it already ends on EOS; `main`'s `add_eos` duplicate is gone.

### `soup data preprocess` EOS is left as on `main`

The preprocess cache never went through TRL's `add_eos`, so its EOS is whatever the post-processor added. This PR de-duplicates its BOS but leaves that EOS behaviour exactly as on `main` (post-processor only), rather than adopting the live rule. For a Qwen-shaped tokenizer the cache therefore trains on zero EOS while the live path trains on one — a pre-existing `main` mismatch, now tracked in #791 and deliberately not resolved here.

### Templated-only

The rule applies only to text a chat template rendered. A tokenizer with no chat_template raises (through `build_format_row`) rather than being silently stripped, and pretrain-style raw text keeps the tokenizer's own special tokens. `soup data doctor --show-mask` now renders through `build_full_sequence_labels`, so the X-ray matches training.

## Before / after on real rows (baseline = TRL's real `add_eos` + default tokenize)

```
BOS-only tokenizer (<s> $A)
  main : BOS=2 EOS=1  [1, 1, 4, 11, 12, 13, 14, 15, 16, 17, 6, 5, 18, 10, 6, 2]
  after: BOS=1 EOS=1  [1, 4, 11, 12, 13, 14, 15, 16, 17, 6, 5, 18, 10, 6, 2]
Qwen-shape tokenizer (no post-processor)
  main : BOS=1 EOS=1  [1, 4, 11, 12, 13, 14, 15, 16, 17, 6, 5, 18, 10, 6, 2]
  after: BOS=1 EOS=1  [1, 4, 11, 12, 13, 14, 15, 16, 17, 6, 5, 18, 10, 6, 2]
BOS+EOS tokenizer (<s> $A </s>)
  main : BOS=2 EOS=2  [1, 1, 4, 11, 12, 13, 14, 15, 16, 17, 6, 5, 18, 10, 6, 2, 2]
  after: BOS=1 EOS=1  [1, 4, 11, 12, 13, 14, 15, 16, 17, 6, 5, 18, 10, 6, 2]
```

The stop token is never dropped (every `after` ends on EOS); duplicated BOS/EOS are collapsed to one. The Qwen row is unchanged from `main` — the point of the review fix: its EOS is kept, not lost.

## Tests

`tests/test_issue785_training_bos.py` (real fast tokenizers, no network), baselined against `_trl_main_lm_ids` (TRL's real `add_eos` + default tokenize), not `tok(text=...)`: one BOS for `{{ bos_token }}` templates, zero for presets; the EOS kept for BOS-only and Qwen shapes (parametrized), duplicates collapsed, template-rendered EOS not re-appended; the no-template path raises through the real `build_format_row` dispatch; and a pretrain preprocess row byte-identical to `main`. Mutations: dropping the live EOS append fails 5 tests, `add_special_tokens=False` on pretrain fails 1. The preprocess re-append on pretrain is inert by construction (with `add_special_tokens=True` the post-processor EOS is already present, so `reappend_eos_if_dropped` never fires) — noted rather than force-killed.

### DIFF
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -2498,30 +2498,32 @@ def preprocess_dataset(
                 truncation=True,
                 padding=False,
                 return_attention_mask=True,
-                # #785: the chat path renders the template's special tokens
-                # already (including any {{ bos_token }}), so re-adding the
-                # tokenizer's defaults here doubled the BOS — one token off from
-                # post-#782 inference. Pretrain feeds raw document text with no
-                # template, so it still gets the tokenizer's leading BOS.
-                add_special_tokens=is_pretrain,
+                # Tokenise exactly as ``main`` (add_special_tokens defaults to
+                # True). This keeps ``main``'s truncation reservation — a
+                # truncated row still ends on the post-processor's EOS — and the
+                # post-processor BOS/EOS, so the cache's EOS stays pinned to
+                # ``main``. #785's only defect on this path is the doubled BOS,
+                # removed below for the chat path.
             )
         except Exception:  # noqa: BLE001 — tokenizer errors vary
             continue
         input_ids = tokens["input_ids"]
+        attention_mask = tokens.get("attention_mask", [1] * len(input_ids))
         if not is_pretrain:
-            # #788: add_special_tokens=False (the #785 BOS fix) also strips the
-            # tokenizer post-processor's EOS. Put it back exactly as the live
-            # training path (loss_mask.build_full_sequence_labels) does, then
-            # re-truncate — otherwise a template that renders BOS but not EOS
-            # trains on a missing stop token from this cache. Pretrain feeds raw
-            # document text and keeps the tokenizer's own specials untouched.
-            from soup_cli.data.loss_mask import reappend_eos_if_dropped
+            # #785/#788: the chat template already renders the BOS; ``main``'s
+            # add_special_tokens=True prepends a second one. Drop only that one
+            # duplicated leading BOS so the row is byte-identical to ``main``
+            # apart from the extra BOS (EOS and truncation untouched). Pretrain
+            # feeds raw document text with no template BOS, so nothing to drop.
+            from soup_cli.data.loss_mask import strip_doubled_leading_bos
 
-            input_ids = reappend_eos_if_dropped(tokenizer, input_ids)[:max_length]
+            input_ids, attention_mask = strip_doubled_leading_bos(
+                tokenizer, input_ids, attention_mask
+            )
         rendered_rows.append(
             {
                 "input_ids": input_ids,
-                "attention_mask": [1] * len(input_ids),
+                "attention_mask": attention_mask,
             }
         )
 
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -2507,12 +2507,21 @@ def preprocess_dataset(
             )
         except Exception:  # noqa: BLE001 — tokenizer errors vary
             continue
+        input_ids = tokens["input_ids"]
+        if not is_pretrain:
+            # #788: add_special_tokens=False (the #785 BOS fix) also strips the
+            # tokenizer post-processor's EOS. Put it back exactly as the live
+            # training path (loss_mask.build_full_sequence_labels) does, then
+            # re-truncate — otherwise a template that renders BOS but not EOS
+            # trains on a missing stop token from this cache. Pretrain feeds raw
+            # document text and keeps the tokenizer's own specials untouched.
+            from soup_cli.data.loss_mask import reappend_eos_if_dropped
+
+            input_ids = reappend_eos_if_dropped(tokenizer, input_ids)[:max_length]
         rendered_rows.append(
             {
-                "input_ids": tokens["input_ids"],
-                "attention_mask": tokens.get(
-                    "attention_mask", [1] * len(tokens["input_ids"])
-                ),
+                "input_ids": input_ids,
+                "attention_mask": [1] * len(input_ids),
             }
         )
 
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -2498,6 +2498,12 @@ def preprocess_dataset(
                 truncation=True,
                 padding=False,
                 return_attention_mask=True,
+                # #785: the chat path renders the template's special tokens
+                # already (including any {{ bos_token }}), so re-adding the
+                # tokenizer's defaults here doubled the BOS — one token off from
+                # post-#782 inference. Pretrain feeds raw document text with no
+                # template, so it still gets the tokenizer's leading BOS.
+                add_special_tokens=is_pretrain,
             )
         except Exception:  # noqa: BLE001 — tokenizer errors vary
             continue
--- a/changelog.d/0.74.0/788.fixed.md
+++ b/changelog.d/0.74.0/788.fixed.md
@@ -12,14 +12,20 @@
   pre-tokenises with `add_special_tokens=False` (via `build_full_sequence_labels`), so the
   chat template is the single source of special tokens, matching inference and HF's own
   `apply_chat_template(tokenize=True)`: one BOS for a `{{ bos_token }}` template, zero for
-  Soup's `data.chat_template` presets. The stop token is preserved rather than lost: when
-  the tokenizer's post-processor would have appended an EOS and the template rendered none,
-  that EOS is re-appended, so the trained EOS count per row is unchanged (a template that
-  already ends on EOS gets no redundant one added). Both the live-training path and the
-  `soup data preprocess` cache path do this through one shared helper
-  (`loss_mask.reappend_eos_if_dropped`), so the AOT cache read straight into the trainer
-  cannot silently train on a missing stop token. The rule applies only to text a chat
-  template rendered; pretrain-style raw text with no template keeps the tokenizer's own
-  BOS. Adapters trained with `train_on_responses_only: false` before this change carry one
-  extra leading BOS in training compared with inference. The `soup data preprocess` cache
-  key changes so datasets cached under the old encoding are re-tokenised rather than reused.
+  Soup's `data.chat_template` presets. The stop token is preserved rather than lost.
+  TRL's language-modeling path also appended `eos_token` as a string to every text row
+  that did not already end with it (`add_eos`), independent of the tokenizer's
+  post-processor, so the pre-tokenised row reproduces that rule: the EOS id is appended
+  when the sequence does not already end on it. A `{{ bos_token }}` template on a BOS-only
+  or EOS-less (Qwen-shaped) tokenizer therefore keeps the single trailing stop token it
+  trained on before, while tokenizers that previously produced a duplicated EOS now end on
+  exactly one. The rule applies only to text a chat template rendered; pretrain-style raw
+  text with no template keeps the tokenizer's own special tokens unchanged. `soup data
+  doctor --show-mask` now renders through the same builder, so the X-ray matches training.
+  Adapters trained with `train_on_responses_only: false` before this change carry one extra
+  leading BOS in training compared with inference. `soup data preprocess` also stopped
+  doubling the BOS; its EOS behaviour is left exactly as before (it never went through TRL's
+  `add_eos`), so a Qwen-shaped tokenizer's cache and live path can still differ on the EOS —
+  that pre-existing mismatch is tracked separately in #791 rather than changed here. The
+  `soup data preprocess` cache key changes so datasets cached under the old encoding are
+  re-tokenised rather than reused.
--- a/changelog.d/0.74.0/788.fixed.md
+++ b/changelog.d/0.74.0/788.fixed.md
@@ -1,0 +1,22 @@
+- **`train_on_responses_only: false` trained on a doubled BOS, one token off from
+  inference after #782 (#788 by @abdulwaarith0).**
+  The default SFT path (`train_on_responses_only: true`) builds its ids with
+  `add_special_tokens=False` in `data/loss_mask.py`, so it never doubled the BOS. The
+  opt-out text path did: `data/sft_format.py` handed TRL a `{"text"}` column, and TRL's
+  language-modeling `tokenize_fn` re-tokenised it with the tokenizer's default
+  `add_special_tokens=True`. A chat template that renders `{{ bos_token }}` (Llama-3,
+  Gemma, Mistral) on a tokenizer whose post-processor also prepends BOS therefore trained
+  on `[bos, bos, ...]`. Before #782 that matched inference by accident; after #782 inference
+  sends one BOS, so adapters trained on the text path were off by one token from how they
+  are prompted. `soup data preprocess` tokenised the same way. The text path now
+  pre-tokenises with `add_special_tokens=False` (via `build_full_sequence_labels`), so the
+  chat template is the single source of special tokens, matching inference and HF's own
+  `apply_chat_template(tokenize=True)`: one BOS for a `{{ bos_token }}` template, zero for
+  Soup's `data.chat_template` presets. The stop token is preserved rather than lost: when
+  the tokenizer's post-processor would have appended an EOS and the template rendered none,
+  that EOS is re-appended, so the trained EOS count per row is unchanged (a template that
+  already ends on EOS gets no redundant one added). The rule applies only to text a chat
+  template rendered; pretrain-style raw text with no template keeps the tokenizer's own
+  BOS. Adapters trained with `train_on_responses_only: false` before this change carry one
+  extra leading BOS in training compared with inference. The `soup data preprocess` cache
+  key changes so datasets cached under the old encoding are re-tokenised rather than reused.
--- a/changelog.d/0.74.0/788.fixed.md
+++ b/changelog.d/0.74.0/788.fixed.md
@@ -11,21 +11,29 @@
   are prompted. `soup data preprocess` tokenised the same way. The text path now
   pre-tokenises with `add_special_tokens=False` (via `build_full_sequence_labels`), so the
   chat template is the single source of special tokens, matching inference and HF's own
-  `apply_chat_template(tokenize=True)`: one BOS for a `{{ bos_token }}` template, zero for
-  Soup's `data.chat_template` presets. The stop token is preserved rather than lost.
-  TRL's language-modeling path also appended `eos_token` as a string to every text row
-  that did not already end with it (`add_eos`), independent of the tokenizer's
+  `apply_chat_template(tokenize=True)`: one BOS for a `{{ bos_token }}` template, and zero
+  for a template that renders no BOS at all — both Soup's `data.chat_template` presets and
+  native templates of that shape (TinyLlama, Zephyr). The stop token is preserved rather
+  than lost. TRL's language-modeling path also appended `eos_token` as a string to every
+  text row that did not already end with it (`add_eos`), independent of the tokenizer's
   post-processor, so the pre-tokenised row reproduces that rule: the EOS id is appended
   when the sequence does not already end on it. A `{{ bos_token }}` template on a BOS-only
   or EOS-less (Qwen-shaped) tokenizer therefore keeps the single trailing stop token it
-  trained on before, while tokenizers that previously produced a duplicated EOS now end on
-  exactly one. The rule applies only to text a chat template rendered; pretrain-style raw
-  text with no template keeps the tokenizer's own special tokens unchanged. `soup data
-  doctor --show-mask` now renders through the same builder, so the X-ray matches training.
-  Adapters trained with `train_on_responses_only: false` before this change carry one extra
-  leading BOS in training compared with inference. `soup data preprocess` also stopped
-  doubling the BOS; its EOS behaviour is left exactly as before (it never went through TRL's
-  `add_eos`), so a Qwen-shaped tokenizer's cache and live path can still differ on the EOS —
-  that pre-existing mismatch is tracked separately in #791 rather than changed here. The
-  `soup data preprocess` cache key changes so datasets cached under the old encoding are
-  re-tokenised rather than reused.
+  trained on before. This changes training results in two ways on configs that hit the
+  affected shapes: a template that renders no BOS (TinyLlama / Zephyr, not only Soup's
+  presets) now trains with zero leading BOS rather than the tokenizer's default one, and a
+  row that used to end on two trailing EOS tokens — a template or post-processor that
+  appended one on top of `add_eos`'s — now trains on exactly one. The rule applies only to
+  text a chat template rendered; pretrain-style raw text with no template keeps the
+  tokenizer's own special tokens unchanged. `soup data doctor --show-mask` now renders
+  through the same builder, so the X-ray matches training. Adapters trained with
+  `train_on_responses_only: false` before this change carry one extra leading BOS in
+  training compared with inference. `soup data preprocess` de-duplicates the same leading
+  BOS but stays pinned to the old cache contents otherwise: it tokenises as before
+  (`add_special_tokens=True`) and drops only the one duplicated leading BOS, so every other
+  token is unchanged — including the post-processor's EOS and the room HF truncation
+  reserves for it, so a truncated row still ends on its stop token. Its EOS therefore
+  remains whatever the post-processor added (never TRL's `add_eos`), so a Qwen-shaped
+  tokenizer's cache and live path can still differ on the EOS — that pre-existing mismatch
+  is tracked separately in #791 rather than changed here. The `soup data preprocess` cache
+  key changes so datasets cached under the old encoding are re-tokenised rather than reused.
--- a/changelog.d/0.74.0/788.fixed.md
+++ b/changelog.d/0.74.0/788.fixed.md
@@ -15,7 +15,10 @@
   Soup's `data.chat_template` presets. The stop token is preserved rather than lost: when
   the tokenizer's post-processor would have appended an EOS and the template rendered none,
   that EOS is re-appended, so the trained EOS count per row is unchanged (a template that
-  already ends on EOS gets no redundant one added). The rule applies only to text a chat
+  already ends on EOS gets no redundant one added). Both the live-training path and the
+  `soup data preprocess` cache path do this through one shared helper
+  (`loss_mask.reappend_eos_if_dropped`), so the AOT cache read straight into the trainer
+  cannot silently train on a missing stop token. The rule applies only to text a chat
   template rendered; pretrain-style raw text with no template keeps the tokenizer's own
   BOS. Adapters trained with `train_on_responses_only: false` before this change carry one
   extra leading BOS in training compared with inference. The `soup data preprocess` cache
--- a/changelog.d/0.75.0/788.fixed.md
+++ b/changelog.d/0.75.0/788.fixed.md
@@ -35,5 +35,7 @@
   reserves for it, so a truncated row still ends on its stop token. Its EOS therefore
   remains whatever the post-processor added (never TRL's `add_eos`), so a Qwen-shaped
   tokenizer's cache and live path can still differ on the EOS — that pre-existing mismatch
-  is tracked separately in #791 rather than changed here. The `soup data preprocess` cache
+  is tracked separately in #791 rather than changed here. The cache also keeps `main`'s
+  single leading BOS where the live path now yields none for a `data.chat_template` preset;
+  that BOS divergence is tracked in #876. The `soup data preprocess` cache
   key changes so datasets cached under the old encoding are re-tokenised rather than reused.
--- a/src/soup_cli/data/loss_mask.py
+++ b/src/soup_cli/data/loss_mask.py
@@ -324,6 +324,69 @@ def build_assistant_only_labels(
     return _truncate(full_ids, labels, max_length)
 
 
+def build_full_sequence_labels(
+    messages: Sequence[dict],
+    tokenizer: Any,
+    max_length: int = 2048,
+) -> dict[str, list[int]]:
+    """Build labels where EVERY token contributes to loss (no masking).
+
+    The ``train_on_responses_only=False`` / ``train_on_messages_with_train_field
+    =False`` path. Pre-tokenises with ``add_special_tokens=False`` (via
+    ``_tokenize_only``) so the chat template is the single source of special
+    tokens, matching inference and ``apply_chat_template(tokenize=True)`` (#785).
+
+    Previously this path handed TRL a ``{"text"}`` column, and TRL's
+    language-modeling ``tokenize_fn`` re-tokenised it with the tokenizer's
+    default ``add_special_tokens=True``. A template that renders
+    ``{{ bos_token }}`` then trained on a doubled BOS, one token off from
+    post-#782 inference.
+    """
+    _check_messages(messages)
+    _validate_max_length(max_length)
+    if not getattr(tokenizer, "chat_template", None):
+        raise ValueError(
+            "tokenizer has no chat_template — cannot render messages for the "
+            "text (train_on_responses_only=false) path"
+        )
+    full_ids = _tokenize_only(tokenizer, messages)
+    # #785: the BOS fix must not cost the training EOS. The pre-#785 path let TRL
+    # tokenize the rendered text with the tokenizer's defaults, so a tokenizer
+    # whose post-processor appends EOS gave every row a trailing stop token even
+    # when the template rendered none. add_special_tokens=False (above) removes
+    # the doubled BOS but also that EOS, which teaches run-on generation
+    # (the failure `soup data doctor`'s eos_in_labels check exists to catch). Put
+    # the stop token back when the tokenizer would have appended one and the
+    # template did not already end the sequence with it.
+    appended_eos = _tokenizer_appends_eos(tokenizer)
+    if appended_eos is not None and (not full_ids or full_ids[-1] != appended_eos):
+        full_ids = full_ids + [appended_eos]
+    labels = list(full_ids)
+    return _truncate(full_ids, labels, max_length)
+
+
+def _tokenizer_appends_eos(tokenizer: Any) -> Optional[int]:
+    """Return the EOS id the tokenizer appends under ``add_special_tokens=True``.
+
+    ``None`` when the tokenizer appends no trailing EOS (e.g. a BOS-only
+    post-processor, or none at all). Detected by probing a trivial string both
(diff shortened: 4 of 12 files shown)


src/soup_cli/commands/data.py
```diff
@@ -2470,6 +2470,12 @@ def preprocess_dataset(
                 truncation=True,
                 padding=False,
                 return_attention_mask=True,
+                # #785: the chat path renders the template's special tokens
+                # already (including any {{ bos_token }}), so re-adding the
+                # tokenizer's defaults here doubled the BOS — one token off from
+                # post-#782 inference. Pretrain feeds raw document text with no
+                # template, so it still gets the tokenizer's leading BOS.
+                add_special_tokens=is_pretrain,
```
AmirF194, line 2478: I ran this against 1655d6d (this PR's head) in a clean python:3.11-slim container, with only transformers/tokenizers/jinja2 installed (no torch/peft/trl needed for this code path).

`preprocess_dataset`'s chat-path tokenize call at data.py:2478 uses `add_special_tokens=is_pretrain`, which is `False` for every templated row. That fixes the doubled BOS, but it also strips any EOS the tokenizer's post-processor would otherwise append, and nothing here re-adds it the way `loss_mask.build_full_sequence_labels` does a few files away in this same PR (`_tokenizer_appends_eos` plus the re-append check). A template that renders BOS but not EOS, on a tokenizer whose post-processor appends EOS, now trains with the stop token silently dropped from the `soup data preprocess` cache, the same run-on-generation failure this PR's own EOS tests guard against on the live-training (`loss_mask.py`) path.

Repro (tokenizer/template built the same way as this PR's own test/test_issue785_training_bos.py):
```
A) build_full_sequence_labels ids: [1, 3, 6, 7, 8, 9, 10, 11, 12, 5, 4, 13, 14, 5, 2]
A) BOS count: 1 EOS count: 1 ends on EOS: True
B) preprocess_dataset-path ids: [1, 3, 6, 7, 8, 9, 10, 11, 12, 5, 4, 13, 14, 5]
B) BOS count: 1 EOS count: 0 ends on EOS: False
```
On base `main` (pre-#788), path B was `[1, 1, 3, ..., 5, 2]`, BOS=2 EOS=1, so the pre-fix code kept the EOS this change now drops.

I did not run the actual `soup data preprocess` CLI end to end (heavy deps); I read `trainer/sft.py::_maybe_load_pretokenized` and it loads the cached `input_ids` straight into the trainer with no relabeling step I could find, so I believe this reaches training as-is, but I have not verified that with a real run.

Fix: call the same `_tokenizer_appends_eos` probe here and append when the row does not already end on it, same as `build_full_sequence_labels`.

abdulwaarith0: Good catch, and thank you for running it rather than eyeballing it. You are right, and I reproduced it on the same fixture the PR's tests use.

`build_full_sequence_labels` re-appends the post-processor EOS when the template renders none; `preprocess_dataset` at `data.py:2478` did not, so with `add_special_tokens=False` the cached row lost the stop token. On a BOS-only template with a tokenizer whose post-processor appends EOS:

```
live path (build_full_sequence_labels): [1, 4, 13, 14, 15, 16, 17, 18, 19, 6, 5, 20, 12, 6, 2]  EOS=1
preprocess path (before fix):           [1, 4, 13, 14, 15, 16, 17, 18, 19, 6, 5, 20, 12, 6]     EOS=0
```

And `_maybe_load_pretokenized` reads that cache straight into the trainer with no relabeling, so it reaches training exactly as you suspected.

Fix pushed in `2ed3b3d`: I pulled the re-append out of `build_full_sequence_labels` into one shared helper, `loss_mask.reappend_eos_if_dropped`, and routed the preprocess chat branch through it (then re-truncate to `max_length`). Both paths now go through the same code so they cannot drift apart again. Pretrain raw text is untouched. Added a preprocess-cache integration test that asserts the cached ids equal the live path (EOS kept when the tokenizer appends one, none invented otherwise); deleting the re-append turns it red.

MakazhanAlpamys:
Reviewed at `2ed3b3da`. The BOS half is right, the tests use real fast tokenizers, both the live path and `soup data preprocess` get the fix, and the cache-key bump is tested. All ten new tests fail on `main`, and the four edited existing tests are legitimate expectation changes rather than weakenings. But condition 1 from the claim thread — **the trained EOS count per row must not change** — does not hold on the live training path, and the PR's before/after baseline is why that was hard to see.

## Blocking: the live path loses the EOS that TRL adds on `main`

On `main`, the legacy text rows reach TRL as `{"text": ...}`, and TRL 0.29.1 appends the EOS **as a string, before tokenizing**, to every row that does not already end with it:

```python
# trl/trainer/sft_trainer.py:1026-1038 (0.29.1)
def add_eos(example, eos_token):
    if "text" in example and not example["text"].endswith(eos_token):
        example["text"] = example["text"] + eos_token
```

This PR pre-tokenizes, so TRL skips that step, and `_tokenizer_appends_eos` only re-adds an EOS when the tokenizer's **post-processor** appends one. Run through the real `sft._map_text_sft_rows` and then TRL's own `_prepare_dataset` on real fast tokenizers:

| template / post-processor | EOS per row, `main` -> PR |
|---|---|
| template renders `{{ bos_token }}`, BOS-only post-processor (Gemma / Llama-2 shape) | **1 -> 0** |
| same template, no post-processor (Qwen shape) — BOS unchanged, so the **only** diff is a lost EOS | **1 -> 0** |
| `chatml` preset on a BOS-only tokenizer (`</s>` never trained) | **1 -> 0** |
| template renders EOS | 3 -> 2 |
| BOS+EOS post-processor | 2 -> 1 |

The first three rows are the run-on-generation failure the condition exists to prevent. The PR body's "before: BOS=2 EOS=0" is not what `main` trains on, and two tests pin the same mistaken baseline: `test_issue785_training_bos.py:162` builds it as `tok(text=text)` and `:185` says in its docstring that pre-fix had no stop token — both model `main` without TRL's `add_eos`.

**Fix:** on the live path, reproduce TRL's rule rather than probing the post-processor — append `eos_token` to the rendered text when it does not already end with it, then encode with `add_special_tokens=False` (or append `eos_token_id` under the same condition). Then re-baseline the tests against TRL's real `_prepare_dataset`, not `tok(text=...)`, so they pin what `main` actually trains on.

## `soup data preprocess`: keep its EOS behaviour as it is on `main`

The preprocess cache never went through TRL's `add_eos` on `main`, so its EOS count is unchanged by your fix today, and I verified that. That means **`main` already disagrees with itself** — the cache and the live path produce different EOS counts — and `test_cache_preserves_eos...`'s "cache == live" assertion cannot hold once the live path is fixed as above. Please do not resolve that inside this PR: changing what the preprocess cache trains on is a separate behaviour change. Keep preprocess's EOS behaviour as on `main`, drop or narrow the cache == live assertion, and I will file the pre-existing mismatch as its own issue so it gets its own decision.

## Condition 2: the code is right, but nothing proves it

Pretrain preprocess keeps `add_special_tokens=True`, and a live row with no template raises as on `main`. But two mutations survive the whole touched suite (405 passed): applying `add_special_tokens=False` to pretrain too, and applying the EOS re-append to pretrain. Please add the no-template / pretrain control the claim asked for — a row that never went through a template, byte-identical to `main`. `test_no_template_is_rejected...` fails on `main` only through an `ImportError`, so it does not discriminate the behaviour.

## Smaller

- Correct the before/after table in the PR body and the changelog sentence "EOS count per row is unchanged" once the fix above lands, so both describe what is measured.
- `_legacy_text_format_row` (`sft_format.py:198`) is now reachable only on the raise path — delete it or say why it stays.
- `data_doctor._build_row_labels`' legacy branch (`:274`) does not mirror the EOS rule, so `soup data doctor --show-mask` will stop matching training. Worth one line in the same push.
- `test_v0537` comment: a `MagicMock` **is** callable; the probe returns `None` for other reasons. Harmless, but inaccurate.

## For the record

Mutations: reverting the core fix (legacy `{"text"}` row back) fails 9 tests; pre-tokenizing but letting the tokenizer add special tokens fails 11; preprocess back to `add_special_tokens=True` fails 3; dropping the live EOS re-append fails 2; dropping the preprocess re-append fails 1; reverting the cache key fails 1. So the existing tests discriminate well — the problem is the baseline they were written against, not their sharpness. 405 passed / 14 skipped on the touched files, ANSI-clean under `FORCE_COLOR=1`, ruff clean, all required checks green.

Reproducing the bug on a real tokenizer before fixing it, and routing both paths through one shared EOS helper, is the right structure — the helper only needs TRL's rule instead of the post-processor probe.

MakazhanAlpamys:
Filed the pre-existing cache-vs-live EOS mismatch as #791, as promised above. For this PR: keep `soup data preprocess` EOS behaviour exactly as on `main` and narrow the cache == live assertion; #791 carries that decision separately.

abdulwaarith0:
You are right on all counts, and thank you for the measurement table. I confirmed TRL 0.29.1's `add_eos` (sft_trainer.py:1026-1031) appends `eos_token` as a string to any text row not ending in it, independent of the post-processor, so my `_tokenizer_appends_eos` probe under-appends for BOS-only and Qwen-shaped tokenizers and my baseline modeled `main` as `tok(text=...)` rather than through TRL's real `_prepare_dataset`. That is the wrong rule.

Revising: (1) live path reproduces TRL's rule (append `eos_token` when the rendered text does not already end with it, then encode `add_special_tokens=False`) and I re-baseline the EOS tests against TRL's real `_prepare_dataset`; (2) I keep `soup data preprocess` EOS behaviour as on `main` and narrow the `cache == live` assertion down to "cache EOS == main's preprocess EOS", leaving the cache-vs-live mismatch to #791; (3) add the pretrain / no-template control that is byte-identical to `main`; (4) fix the PR-body table and changelog wording, delete the now-dead `_legacy_text_format_row`, mirror the EOS rule in `data_doctor._build_row_labels` so `--show-mask` matches training, and correct the `test_v0537` comment. Will push and re-measure.

abdulwaarith0:
Pushed the revision as 8b69428.

**Live path now reproduces TRL's rule.** `build_full_sequence_labels` calls `append_training_eos`, which appends the EOS id unless the sequence already ends on it — TRL 0.29.1's `add_eos`, independent of the post-processor. The three `1 -> 0` rows are gone; measured against TRL's real `add_eos` baseline now, not `tok(text=...)`:

```
BOS-only tokenizer:  main BOS=2 EOS=1  ->  after BOS=1 EOS=1
Qwen-shape (no PP):  main BOS=1 EOS=1  ->  after BOS=1 EOS=1   (unchanged: EOS kept)
BOS+EOS tokenizer:   main BOS=2 EOS=2  ->  after BOS=1 EOS=1   (both duplicates collapsed)
```

**`soup data preprocess` left as on `main`.** It keeps the post-processor-only EOS (via `reappend_eos_if_dropped`, which I verified leaves the cache byte-for-byte as `main` on the EOS while still de-duplicating the BOS). The `cache == live` assertion is gone; the cache tests now pin `cache EOS == main preprocess EOS`, and one of them documents the Qwen cache-vs-live divergence explicitly as #791.

**Condition 2 control added.** A pretrain preprocess row is asserted byte-identical to `main`. It kills the `add_special_tokens=False`-on-pretrain mutation. The other one you named — the EOS re-append applied to pretrain — turns out to be inert by construction: with `add_special_tokens=True` the post-processor EOS is already the last token, so `reappend_eos_if_dropped` never fires on a pretrain row. I left the `if not is_pretrain` guard in as defensive and noted this rather than manufacturing a case to force it red; happy to drop the guard instead if you'd prefer the smaller surface.

**Smaller items:** re-baselined `test_issue785` against `_trl_main_lm_ids` (TRL's real `add_eos` + default tokenize); `test_no_template_...` now goes through the real `build_format_row` dispatch so it discriminates the raise rather than an import; deleted the dead `{"text"}` return in `_legacy_text_format_row`; `data_doctor._build_row_labels`' legacy branch now calls `build_full_sequence_labels` so `--show-mask` matches training; corrected the `test_v0537` MagicMock comment; and fixed the before/after table and the changelog wording.

One note on the baseline: driving the real `SFTTrainer._prepare_dataset` in the suite hangs on Windows under `PartialState`, so `_trl_main_lm_ids` reproduces TRL's two real operations (`add_eos` at `sft_trainer.py:1026-1031`, then `processing_class(text=...)` at `:1131`) on the real tokenizer instead. If you'd rather I gate a single real-`_prepare_dataset` cross-check behind a non-Windows marker, say the word.

MakazhanAlpamys:
Re-reviewed at `8b69428e`. **The live training path is fixed.** The three rows that went 1 -> 0 EOS in round 1 now keep their stop token. Measured through the real `sft._map_text_sft_rows` and then TRL 0.29.1's real `SFTTrainer._prepare_dataset`, on both `main` and your head: on eight real tokenizers (Qwen2.5, Qwen3, SmolLM2, Phi-3, TinyLlama, Llama-3.2, Gemma-2, Mistral-v0.3) **EOS per row is identical to `main`**, including multi-turn, empty-assistant and rows truncated at 64. BOS drops by exactly one where the template renders it. Switching back to the round-1 post-processor rule now fails 4 tests.

One blocker left, on the preprocess side, plus a decision from me on the collapse you disclosed.

## Blocking: `soup data preprocess` drops the EOS on truncated rows

`commands/data.py:2466-2489` now tokenizes with `truncation=True, add_special_tokens=False`, re-appends the EOS, and then slices `[:max_length]`. On `main`, HF truncation **reserved room** for the post-processor's EOS, so a truncated row still ended on it. Through the real CLI:

| case, truncated at 64 | EOS, `main` -> PR |
|---|---|
| BOS+EOS post-processor | **1 -> 0** |
| EOS-only post-processor | **1 -> 0** |
| real TinyLlama with `add_eos_token=True` | 2 -> 1 (the final `</s>` is gone) |
| template renders EOS, BOS+EOS post-processor | 3 -> 2 (`reappend_eos_if_dropped` skips rows already ending on EOS) |

Non-truncated rows, pretrain, Qwen2.5 and Llama-3.2 match `main` apart from the BOS. So this is exactly the "cached runs can train with no stop token" failure, arriving through truncation. The PR body and the changelog currently say preprocess EOS is left as before, and for truncated rows it is not.

**Fix:** reproduce `main`'s truncation. Either truncate the content to `max_length` minus the post-processor's special tokens *before* re-appending, or tokenize exactly as `main` did and drop only the duplicated leading BOS. The second is the smaller change and the easier one to prove identical. Then add truncated-row and template-renders-EOS cases to `TestPreprocessCachePathEOS`. Today nothing covers this: deleting the re-truncation survives the whole suite.

## Decision: the duplicated trailing EOS may collapse to one

Where `main` trained on a **doubled** trailing EOS — a tokenizer whose post-processor appends EOS on top of the one TRL's `add_eos` adds, or a template that renders EOS — your change trains on one. Read literally, my condition 1 forbids that. But the round-1 fix I prescribed (append the EOS string, then encode with `add_special_tokens=False`) produces the same 2 -> 1, so the condition and the prescription could not both hold. **I am accepting the collapse.** No row loses its stop token, and a doubled EOS is the same class of defect as the doubled BOS this PR exists to remove.

It has to be stated, though. Add one sentence to the changelog fragment saying that rows which trained on two trailing EOS tokens now train on one. While there, note that native templates rendering no BOS (TinyLlama / Zephyr shape) also train with zero BOS, not only Soup's presets.

The two rows where your ID-level rule differs from TRL's string rule — a template that renders EOS followed by a trailing space the tokenizer drops, and an EOS token with `rstrip=True` — occur in none of the eight real tokenizers, and `test_template_rendered_eos_is_kept_and_not_re_appended` pins the choice. Fine as it is.

## Please add in the same push

- **A test for the `data_doctor` legacy branch.** The code mirrors the rule now, but reverting `utils/data_doctor.py` to `main` passes all 919 related tests. Without a test, `soup data doctor --show-mask` can drift from training unnoticed.
- The comment at `commands/data.py:2484-2489` says it re-appends "exactly as the live path does". The live path now uses a different rule, so reword it.

Nice-to-have: one test through TRL's real `_prepare_dataset` rather than `_trl_main_lm_ids`, which re-implements `add_eos` plus a default tokenize by hand. It ran without hanging on Windows here.

## Everything else

- `_legacy_text_format_row` reduced to the raise, with the reason in its docstring — addressed.
- "cache == live" narrowed to "cache equals `main`'s preprocess" — right.
- Your explanation for why the pretrain EOS re-append mutation is inert by construction holds.
- Mutations on the new code, all killed: revert `sft_format.py` (10), drop `append_training_eos` (5), always append (1), back to the post-processor probe (4), revert the cache key (1), `add_special_tokens=False` on pretrain (1), preprocess back to `add_special_tokens=True` (3).
- 919 passed / 16 skipped across the 13 related files; ruff clean; all 13 required checks green on `8b69428e`. The four removed `"text" in out` checks are legitimate: the row shape changed.

Taking the round-1 finding in full, and re-baselining against what `main` actually trains on rather than defending the first table, is what made this round short.

src/soup_cli/commands/data.py
AmirF194: Confirmed. 2ed3b3d fixes it.

abdulwaarith0:
Pushed the round-2 fix as `48cf025`.

**Preprocess truncation blocker fixed, your option (b).** The path now tokenises exactly as `main` (`add_special_tokens=True`, so HF reserves truncation room for the post-processor's EOS) and drops only the one duplicated leading BOS via a new `strip_doubled_leading_bos`. A row truncated at 64 keeps its stop token now, and every token other than the dropped BOS is byte-identical to `main`. Verified against the four cases from your table:

```
BOS+EOS post-processor, trunc@64:        main EOS=1 -> after EOS=1
EOS-only post-processor, trunc@64:       main EOS=1 -> after EOS=1
template renders EOS + BOS+EOS PP:       main EOS=3 -> after EOS=3  (pinned to main, not collapsed)
non-truncated / pretrain / Qwen:         match main apart from the one BOS
```

The old `reappend_eos_if_dropped` / `_tokenizer_appends_eos` pair is gone; the cache path no longer probes the post-processor at all, it just removes #785's duplicate. `test_cache_keeps_eos_on_truncated_row` pins the blocker (it fails on the round-1 slice), and `test_cache_keeps_mains_eos_when_template_renders_it` pins that the cache stays on `main`'s EOS count rather than collapsing like the live path (the #791 divergence).

**Collapse decision recorded.** Changelog now states that rows which trained on two trailing EOS tokens train on one, and that templates rendering no BOS (TinyLlama / Zephyr shape, not only Soup's presets) now train with zero leading BOS.

**`data_doctor` legacy-branch test added.** `TestDataDoctorLegacyMatchesTraining` asserts the `--show-mask` legacy branch tokenises identically to `build_full_sequence_labels` across bos+eos / bos-only / no-post-processor. Reverting `utils/data_doctor.py` to `main` now fails it.

**Smaller:** reworded the `commands/data.py` comment so it no longer claims it re-appends "as the live path" (it does not; it drops the duplicated BOS and leaves `main`'s EOS).

On the real `_prepare_dataset` cross-check: it still hangs here on Windows under `PartialState`, so I have kept `_trl_main_lm_ids` (TRL's `add_eos` at `sft_trainer.py:1026-1031` then `processing_class(text=...)` at `:1131`). Since it ran clean for you, I am happy to add one real `_prepare_dataset` test gated behind a non-Windows marker if you want it in-tree; say the word and I will push it.

Local run: 142 passed / 7 skipped across the related suites, 17/17 in `test_issue785_training_bos.py`, ruff clean.

MakazhanAlpamys:
Re-reviewed at `48cf025b`. **The preprocess blocker is fixed**, the way I suggested: tokenize exactly as `main`, then strip only the doubled leading BOS. Measured through the real `soup data preprocess` path on rows truncated at `max_length=64`:

| case | BOS `main` -> PR | EOS `main` -> PR |
|---|---|---|
| BOS template + BOS/EOS post-processor | 2 -> 1 | 1 -> 1 |
| BOS template + EOS-only post-processor | 1 -> 1 | 1 -> 1 |
| TinyLlama `add_eos_token=True` | 0 -> 0 | 2 -> 2 |
| template renders EOS + BOS/EOS post-processor (short / truncated) | 2 -> 1 | 3 -> 3 / 2 -> 2 |
| Llama-3.2, Mistral-v0.3, Gemma-2 | 2 -> 1 | unchanged |
| Qwen2.5, SmolLM2, pretrain rows | identical | identical |

Every other token matches `main`, and the attention mask keeps the right length. Items 2-5 from round 2 are addressed: the truncated-row and template-renders-EOS tests, the fragment sentences about the EOS collapse and no-BOS native templates, the `data_doctor` legacy test and the reworded comment. Every mutation is killed:

- preprocess with `add_special_tokens=False` (4 failures);
- the round-1 preprocess path (4);
- dropping `strip_doubled_leading_bos` (4);
- reverting the `data_doctor` branch (3);
- the live path without `append_training_eos` (5), or back on the post-processor probe (7);
- reverting the cache key (1).

## One small fix: CI is red on a test this push broke

`test (macos-latest, 3.11)` fails exactly one test, and it fails locally too:

```
tests/test_issue781_inference_bos.py::TestNoRenderThenReTokenize::test_no_scope_renders_a_template_and_lets_the_tokenizer_add_special_tokens
```

`48cf025` puts `preprocess_dataset` back on the tokenizer's default `add_special_tokens=True`, which is correct, but the PR still removes that function's entry from the #781 ratchet's `_ALLOWLIST`. The comment left in its place ("now tokenizes with add_special_tokens=is_pretrain") is no longer true. Restore the entry, with its real reason — *renders a template, lets the tokenizer add special tokens on purpose, then strips the doubled leading BOS* — so the next reader sees why this call site is allowed.

The cache-key comment in `utils/data_pipeline.py` ("the chat path now tokenizes with add_special_tokens=False") is stale for the same reason — one line.

After that, the remaining gate for this PR is green CI.

abdulwaarith0:
Fixed in the follow-up push. Both were the same stale-comment root: `48cf025` returned `preprocess_dataset` to the tokenizer's default `add_special_tokens=True` but left it out of the #781 ratchet.

- Restored `commands/data.py::preprocess_dataset` to `_ALLOWLIST` with its real reason: it renders a template, then tokenizes with `add_special_tokens=True` on purpose so the cache stays byte-identical to `main` (truncation reservation and post-processor EOS preserved), and strips only the one doubled leading BOS afterwards. `test_no_scope_renders_a_template_and_lets_the_tokenizer_add_special_tokens` and `test_every_allowlist_entry_is_still_earned` both pass now.
- Reworded the `_PREPROCESS_TOKENIZE_SCHEMA` comment in `utils/data_pipeline.py` to match: the chat path still tokenizes with `add_special_tokens=True` and strips the doubled leading BOS, rather than the old "tokenizes with `add_special_tokens=False`".

`tests/test_issue781_inference_bos.py` and `tests/test_issue785_training_bos.py` pass (79), ruff clean. Thanks for catching the ratchet gap.

MakazhanAlpamys:
@abdulwaarith0 — re-reviewed at `b41f4943`, 14/14 required checks green. **The blocking item is genuinely resolved**, and the way you resolved it is the part worth naming. Two small asks, both documentation-only, and then I merge.

## The allowlist: restored, and re-earned by the code rather than worked around

I checked this in both directions, because `tests/test_issue781_inference_bos.py` enforces the ratchet twice — `test_no_scope_renders_a_template_and_lets_the_tokenizer_add_special_tokens` fails if an entry is missing, and `test_every_allowlist_entry_is_still_earned` fails if an entry is present for a scope that no longer does it. An entry cannot be parked there.

The `preprocess_dataset` entry is back, with a rewritten rationale rather than the old one. And it is back because `48cf025b` restored the *source* first, for an independent and correct reason: the intermediate `add_special_tokens=is_pretrain` version did `reappend_eos_if_dropped(...)[:max_length]`, appending the EOS and then slicing it straight off again on any truncated row — one stop token becoming zero. Tokenizing as `main` does lets HF reserve the truncation room so a truncated row still ends on EOS. `b41f4943` then only re-added the entry that restored code had re-earned; its entire source diff is a seven-line comment.

Two mutations say this is principled rather than convenient: deleting the allowlist entry is **caught** by `test_every_allowlist_entry_is_still_earned`, and reverting `48cf025b` is **caught** by `test_cache_keeps_eos_on_truncated_row` — a *behavioural* test, not the ratchet. The design choice is pinned by something other than the allowlist, which is exactly what separates a fix from test-shaping.

## What I ran

261 passed, 14 skipped across the five files you touch (all 14 skips are pre-existing platform gates in `test_v0537.py`; **zero** skips in `test_issue785_training_bos.py`). 747 passed, 2 skipped across the blast radius, which I computed by symbol rather than by keyword. ruff clean. Seven mutations, **all seven caught**: removing the live-path de-dup; deleting the `append_training_eos` call; neutering `strip_doubled_leading_bos` to a no-op; inverting `48cf025b`; reverting `b41f4943`; dropping the cache-key schema bump; and reverting `data_doctor --show-mask` to its own render instead of the trainer's builder.

**Test-diff audit: nothing weakened.** `+477 / −9`, no deletion, no skip or xfail added, no parametrize narrowed — one was *added*. All nine removed lines are the same contract update (the legacy path returning `input_ids` instead of `{"text"}`), and every replacement is equal or stronger; `test_v0537.py` went from a substring check on `result.get("text","")` to exact equality on `result.get("input_ids")`.

And the thing I most wanted to confirm: the fix **reproduces TRL's own rule instead of inventing a parallel one**. I read `trl/trainer/sft_trainer.py` directly — `add_eos` appends `eos_token` as a string to any text row not already ending with it, independent of the post-processor — and `append_training_eos` mirrors that in token space. De-duplication on the live path is achieved by *not adding* (`add_special_tokens=False`), per #781's stated principle, not by stripping; the only ID-based strip is on the cache path and is correctly narrow.

## Two asks

**1. The `strip_doubled_leading_bos` docstring points at the wrong issue, and that is the one that will mislead someone later.** It defers the remaining divergence to "the #791 family", but #791 is strictly about EOS and says nothing about BOS. Measuring it out: for a `data.chat_template` preset on a tokenizer whose post-processor adds a BOS, the live path is now **0** and the cache is **1**. On `main` both were 1, so they agreed — on a value #781 established is wrong. Your change moves the live path to the right answer and correctly leaves the cache byte-identical, so nobody is worse off; but the two now disagree, and `strip_doubled_leading_bos` cannot close it by construction, since it collapses a *doubled* BOS and there is only one here.

I have filed **#876** for the BOS half, with the measured table and the note that it must land in the same shape as whatever #791 settles on. Please point the docstring (and the fragment line that mentions it) at #876 instead of #791.

**2. The PR body is stale and contradicts the shipped code.** It still says preprocess "gets the same BOS fix (`add_special_tokens=is_pretrain`)", which `48cf025b` reverted two commits ago. A later section of the same body is correct. The changelog fragment is accurate and already describes the final design — it is only the body, and the body is what a reader sees at merge.

## Two things I checked and am *not* asking you to change

A truncated row on the live path does not end on EOS, because `build_full_sequence_labels` appends EOS and `_truncate` then slices it off. I read this as a regression at first and it is not: TRL 0.29.1 tokenizes with no truncation at `sft_trainer.py:1131` and truncates afterwards in `trl.data_utils.truncate_dataset`, a plain slice — `main` loses the EOS identically. No regression. Pinning it with a test is optional.

`tests/test_v07127.py::TestDoctorCli::test_control_characters_in_dataset_content_stripped_from_terminal` fails under `FORCE_COLOR=1`. I verified it fails identically on `origin/main`, so it is pre-existing and belongs to a separate hardening pass, not to you.

Housekeeping is clean: fragment present in `changelog.d/0.74.0/`, named after the PR, credited, and unusually thorough — it states the adapter-compatibility consequence #785 asked for. No docs change needed; `docs/performance-and-quantization.md` documents `train_on_responses_only` without discussing special-token encoding, so nothing there is made wrong.

No overlap with your #867 — that one touches only serving paths and this one only `data/` plus `commands/data.py` and `utils/data_doctor.py`. No ordering dependency either way. #785 stays open for the serving half regardless.

Push those two and I merge on green.

MakazhanAlpamys:

MakazhanAlpamys:

```
```

```
```

abdulwaarith0:
Thanks, and agreed on the #791/#876 split.

Pointed the BOS divergence at #876. The `strip_doubled_leading_bos` docstring and the fragment line that mentioned it now say the residual cache-vs-live gap is the leading BOS (live path zero for a `data.chat_template` preset, cache one, as you measured) and track it in #876. The EOS mismatch reference stays #791, since that one really is EOS.

Fixed the stale body. The "The fix" section said preprocess used `add_special_tokens=is_pretrain`, which `48cf025` reverted; it now says preprocess tokenises exactly as `main` with `add_special_tokens=True` and drops only the one duplicated leading BOS, matching the section further down and the shipped code.

Also rebased onto master and moved the fragment to `changelog.d/0.75.0/` for the 487 baseline check.

MakazhanAlpamys:
Merged as 1c399f1. Four review rounds, and every item from all four is closed — thank you for staying with it.

**What I verified before merging, on `edf51d4a`:**

- Your five touched files: **261 passed, 14 skipped**. Blast radius (14 files importing `loss_mask` / `sft_format` / `data_doctor` / `make_preprocess_cache_key` / `build_format_row`): **955 passed, 16 skipped**. Under `FORCE_COLOR=1 TERM=xterm-256color`: **105 passed**. ruff clean. 14/14 required checks on the exact head.
- Your new tests run against `origin/main`'s source: **15 failed, 2 passed** — and the two survivors are the pretrain and no-template *controls*, which is exactly right. A control that fails on main is not a control.
- **Mutations, five applied, five killed:** the core one (`_build_full_sequence_format_row` returning `{"text": rendered}`, i.e. main) fails **11**; `strip_doubled_leading_bos` neutered fails 4; dropping `append_training_eos` fails 5; reverting `data_doctor._build_row_labels` to its own render fails 3; dropping the `_PREPROCESS_TOKENIZE_SCHEMA` cache-key bump fails 1.
- **I re-measured rather than reading your tables**, on real fast tokenizers: 12 shapes (3 templates x 4 post-processors) live, and 24 cache cases (those 12 at `max_length` 2048 and at 24-with-truncation). **All 36 keep main's stop token.** BOS goes 2 -> 1 where the template renders one and stays 0 for presets. The round-2 blocker — EOS lost on truncated cache rows — is genuinely closed, not merely tested.

**The three things I want on the record, because they are the reasons this took four rounds and came out right:**

1. **You reproduced TRL's rule instead of probing the post-processor.** The first revision inferred what TRL would do; the merged one appends `eos_token` when the rendered text does not already end with it, which is what TRL's `add_eos` does. Inference about a dependency's behaviour is a bug waiting for a version bump.
2. **You narrowed `soup data preprocess` rather than unifying it.** The tempting fix was to make cache and live agree. You kept main's tokenization and dropped only the duplicated leading BOS, so the cache's EOS is byte-identical to main in all 24 cases — and the remaining cache-vs-live BOS divergence went to #876 instead of being quietly absorbed.
3. **`data_doctor` now *calls* the shared builder rather than mirroring it.** I asked for the mirror to be correct; you removed the mirror. That is strictly stronger, and `TestDataDoctorLegacyMatchesTraining` with `count(BOS) == 1` keeps the agreement from being vacuous.

**Zero tests weakened.** Four removed assertion lines, every replacement equal or stronger — `"rendered" in result.get("text","")` (substring) became `result["input_ids"] == [7,8,9]` (exact equality), which is the direction these usually go the other way.

**One wording nit for whenever you next touch the file, not worth a PR.** The docstrings and the fragment say the fix works by pre-tokenising "with `add_special_tokens=False`". That kwarg is a no-op here: transformers 5.17's `apply_chat_template(tokenize=True)` hardcodes it internally (checked with `inspect.getsource`). The real mechanism is pre-tokenising *at all*, so TRL's `tokenize_fn` never runs — which is also why a mutation of that kwarg survives while the mutation of the mechanism fails 11 tests.

The optional real-`_prepare_dataset` test stays dropped. We never settled the Windows `PartialState` hang and I am not going to hold a closed blocker open over it.

Credited in CONTRIBUTORS.md. #867 and #892 merged right behind this.

PR #789, opened 2026-09-11
fix(tests)[#783]: extend path containment ratchet to detect contextlib.suppress

### LINKED ISSUE #783: Path-containment ratchet does not inspect contextlib.suppress(ValueError), which is already an idiom in this repo
## What

`tests/test_issue775_path_containment_ratchet.py` (added in #778) treats `relative_to` as a containment decision when it sits inside a `try` whose handler can catch `ValueError`. It inspects `ast.Try` only, so the equivalent `contextlib.suppress` form slips through:

```python
from contextlib import suppress
from pathlib import Path

def is_inside(path: Path, base: Path) -> bool:
    with suppress(ValueError):
        path.resolve().relative_to(base.resolve())
        return True
    return False
```

Verified during review of #778: the scan reports nothing for this shape. The `except ValueError:` spelling of the same function is caught at the correct line.

## Why it matters

This is not a hypothetical spelling. `contextlib.suppress` is already used in `src/soup_cli/` (four call sites today, e.g. `trainer/rewards.py:449`, `:467`, `:469`), so the next person writing a containment check here has a reasonable chance of reaching for it. The whole argument of #777 was that a guard keyed on one spelling cannot catch the next occurrence; this is one more spelling.

The underlying defect the guard exists to prevent is unchanged: `Path.resolve()` + `relative_to()` breaks on Windows 8.3 short names (`C:\Users\RUNNER~1`), which is why the convention is `os.path.realpath` + `os.path.commonpath`.

## Proposed fix path

In the scanner's `try`-handler branch, also walk `ast.With` / `ast.AsyncWith` and treat a `relative_to` call in the body as a containment decision when any `withitem.context_expr` is a call to `suppress` (bare name or `contextlib.suppress`) whose arguments include a name already accepted by `_CATCHES_VALUE_ERROR` (`ValueError`, `Exception`, `BaseException`). This is the same predicate the `ast.Try` branch already uses, so it should be a reuse rather than a second rule.

## Acceptance criteria

- [ ] The `with suppress(ValueError):` example above is reported, with the correct file and line.
- [ ] `contextlib.suppress(ValueError)` (attribute form) and a multi-argument `suppress(OSError, ValueError)` are both reported.
- [ ] `with suppress(OSError):` around a `relative_to` is **not** reported — `OSError` cannot catch the `ValueError`, mirroring the existing `except OSError` green control.
- [ ] A `suppress` used for something unrelated to `relative_to` is not reported.
- [ ] `TestTheScannerCanActuallyFail` gains a red case and a green case for this shape, so the new branch has its own non-vacuity proof.
- [ ] The scan still reports nothing outside the existing `ALLOWLIST` on a clean tree.

Follow-up to #777 / #778.

AaronProbha18:
## What does this PR do?
Closes #783. The path-containment ratchet recognized `try/except ValueError:` around a `relative_to` call but not the equivalent `with contextlib.suppress(ValueError):`. A `relative_to` guarded only by `suppress(ValueError)`, the same containment decision, shorter spelling, and an idiom already in use in `src/soup_cli/trainer/rewards.py` for unrelated purposes, passed the scan undetected. The ratchet now sees both, so the banned idiom can't just move into a context manager.

### What changed

`tests/test_issue775_path_containment_ratchet.py` only, no `src/` changes:

- **`_names_catch_value_error(nodes)`** the exception-name predicate lifted out of `_handler_catches_value_error`, which becomes a thin wrapper over it. One rule serving both spellings, not a second copy of the rule. It recurses into `ast.Tuple`, so `suppress((OSError, ValueError))` matches `suppress` forwards its arguments to `issubclass`, which accepts nested tuples.
- **`_suppress_names(tree)`** `{"suppress"}` plus any `from contextlib import suppress as X` alias in the module being scanned.
- **`_is_suppress_call(node, names)`**-  an `ast.Call` whose func is a `Name` in that set, or any `X.suppress` attribute. The attribute form covers `contextlib.suppress` and `import contextlib as cl` → `cl.suppress` in one line, with no import following; marked with a `ponytail:` comment naming the ceiling. 
- **A `With`/`AsyncWith` branch** in `_containment_reason`'s parent walk, gated on `_in_field(parent, child, "body")`  never `items`. A `relative_to` used while *building* the `suppress(...)` call is not guarded by it, the same distinction the scanner already draws between `If.test` and `If.body`. The existing function-boundary early return needed no change: it already stops the walk before any enclosing `with`.

`ALLOWLIST` is untouched and still matches exactly the two live sites in
`commands/adapters.py`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Tests

## Checklist

### DIFF
--- a/tests/test_issue775_path_containment_ratchet.py
+++ b/tests/test_issue775_path_containment_ratchet.py
@@ -33,6 +33,8 @@ bug — it is also the normal way to shorten a path for display):
   "cosmetic fallback" apart needs a taint analysis, and the cheap version got
   ``adapters.py:38`` wrong in both directions while being written. A benign
   site is one allowlist line with a reason instead.
+  ``with contextlib.suppress(ValueError):`` is the same decision written as a
+  context manager and counts identically.
 * ``relative_to`` in a boolean position — ``if``/``while``/``assert`` test,
   ``not``, ``and``/``or``, a ternary condition, a comprehension filter. Nobody
   writes that for display.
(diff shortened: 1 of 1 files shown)


MakazhanAlpamys:
Merged as d4c50406 — thank you, @AaronProbha18, and welcome. You are now in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md).

For a first PR this is unusually disciplined, and the discipline is in the places that matter for a guard:

- **One rule, two spellings.** `_names_catch_value_error` is shared with the `except` path instead of a second copy that would drift from it, and it recurses into tuples, so `suppress((OSError, ValueError))` matches.
- **You caught the aliases rather than listing them as non-goals.** `from contextlib import suppress as quiet` and `import contextlib as cl` then `cl.suppress(...)` are both reported.
- **The body-only gate.** A `relative_to` used while building the `suppress(...)` call, or after the block, is not guarded by it. That is the `If.test` vs `If.body` distinction the scanner already drew, and it is what stops the new branch from firing on correct code.

What I ran before merging:

- Injected into a real module under `src/soup_cli/`, each shape made the repo-wide scan fail at the right file and line: the issue's example, `contextlib.suppress`, `suppress(OSError, ValueError)`, `suppress((OSError, ValueError))`, both alias forms, the second item of a multi-item `with`, a call nested in an `if` in the body, and `suppress(Exception)`. `suppress(OSError)`, `relative_to` in the `with` items, a nested `def` and code after the block all stayed green.
- Scanner mutations, all killed: deleting the `With`/`AsyncWith` branch (6 failed), dropping the body-only gate, ignoring import aliases, removing tuple recursion, removing the attribute form, removing the function-boundary stop.
- 35 passed with `test_code_review_recurring.py`; the live scan is still clean outside `ALLOWLIST`; ruff clean; all 13 required checks green on `c9c227a7`.

Two small things for next time, neither worth holding this for. The claim thread asked for one sentence in the PR body saying you proved the new branch red against the real scan. I verified it myself, but stating it is what lets a reviewer trust the guard without re-deriving it. And `suppress(*errs)` is not caught, consistent with `except ERRS:`; that non-goal belongs in the module docstring rather than only in a code comment.

PR #790, opened 2026-09-11
fix(deepspeed): forward create_optimizer's arguments through the empty-group guard

### LINKED ISSUE #784: attach_empty_param_group_guard drops create_optimizer's model parameter, so the positional call transformers uses raises TypeError

Mihailchik:
Closes #784.

The guard replaced `trainer.create_optimizer` with a function that took no arguments. transformers calls the original as `create_optimizer()` from `create_optimizer_and_scheduler` and on the eager path (`trainer.py:1226`, `:1675`), but positionally as `self.create_optimizer(model)` on the delayed-creation path (`:1694`). That call raised `TypeError`.

## Change

```python
def create_optimizer(*args, **kwargs):
    optimizer = original(*args, **kwargs)
```

The wrapper forwards exactly what it receives. I went with this rather than `model=None` → `original(model)` for one reason: the latter turns the zero-argument call into `original(None)`, which only works as long as every trainer the guard is attached to accepts a `model` parameter. None of TRL's trainers override `create_optimizer` today, so that is a hedge rather than a fix for a live problem. Forwarding keeps `create_optimizer()` literally unchanged either way.

`functools.wraps` is off on purpose. With it, `inspect.signature` would report the original's `(model=None)` and a signature test would pass even if the wrapper itself took no arguments.

## Tests

New `tests/test_issue784_guard_forwards_model.py`, following the fake-optimizer pattern in `test_issue336`:

- the positional and keyword calls both forward the same `model` object to the original, and still prune the empty group;
- the no-argument call reaches a zero-argument original unchanged, and an HF-style original still receives its `None` default;
- the wrapper's own signature binds a positional and a keyword `model`.

Without the fix, 3 of the 5 fail: positional, keyword and signature. The existing guard suites (`test_issue336`, `test_issue359`, `test_issue700`) pass unchanged: 143 passed. `ruff check` and `ruff format --check` are clean.

No changelog fragment, since the change isn't user-visible today. Happy to add a `fixed` one if you'd like it recorded.

### DIFF
--- a/src/soup_cli/utils/deepspeed.py
+++ b/src/soup_cli/utils/deepspeed.py
@@ -253,9 +253,7 @@ def resolve_deepspeed_config(
         zero["zero_hpz_partition_size"] = partition
 
     quant_keys = [
-        key
-        for key in ("zero_quantized_weights", "zero_quantized_gradients")
-        if zero.get(key)
+        key for key in ("zero_quantized_weights", "zero_quantized_gradients") if zero.get(key)
     ]
     if quant_keys:
         bf16_on = bool((resolved.get("bf16") or {}).get("enabled"))
@@ -289,9 +287,7 @@ def write_deepspeed_config(stage: str = "zero2", gpu_count: int | None = None) -
         console = Console()
         for note in notes:
             console.print(f"[yellow]DeepSpeed {stage}:[/] {escape(note)}")
-    tmp = tempfile.NamedTemporaryFile(
-        mode="w", suffix=".json", prefix="ds_config_", delete=False
-    )
+    tmp = tempfile.NamedTemporaryFile(mode="w", suffix=".json", prefix="ds_config_", delete=False)
     json.dump(config, tmp, indent=2)
     tmp.close()
     return tmp.name
@@ -339,9 +335,7 @@ def resolve_user_deepspeed_file(path: str, *, gpu_count: int | None = None) -> s
     for note in notes:
         console.print(f"[yellow]DeepSpeed {escape(label)}:[/] {escape(note)}")
 
-    tmp = tempfile.NamedTemporaryFile(
-        mode="w", suffix=".json", prefix="ds_user_", delete=False
-    )
+    tmp = tempfile.NamedTemporaryFile(mode="w", suffix=".json", prefix="ds_user_", delete=False)
     json.dump(resolved, tmp, indent=2)
     tmp.close()
     return tmp.name
@@ -410,8 +404,12 @@ def attach_empty_param_group_guard(trainer) -> bool:
     if not callable(original):
         return False
 
-    def create_optimizer():
-        optimizer = original()
+    # Forward exactly what the caller passed. transformers calls this both as
+    # ``create_optimizer()`` and, on the delayed-creation path, positionally as
+    # ``create_optimizer(model)`` (#784). Passing nothing through unchanged keeps
+    # the no-argument call identical, and no trainer's signature is assumed.
+    def create_optimizer(*args, **kwargs):
+        optimizer = original(*args, **kwargs)
         dropped = prune_empty_param_groups(optimizer)
         if dropped:
             from rich.console import Console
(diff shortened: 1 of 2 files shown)


MakazhanAlpamys:
Thank you for this, @Mihailchik — and I am sorry for the wasted effort, because it is well reasoned. #784 was already fixed by #786 (@DYNOSuprovo), merged at 07:36 UTC today, about three hours before this PR opened. Your branch was cut from `main` before that merge, which is why it now conflicts. I am closing it as superseded rather than asking you to rebase onto a fix that already landed.

On the design point, so it does not look ignored: the objection to `model=None` -> `original(model)` is right about the form the issue suggested, but #786 does not use that form. It splits on whether a model was passed:

```python
def create_optimizer(model=None):
    if model is not None:
        optimizer = original(model)
    else:
        optimizer = original()
```

So the zero-argument call reaches the original as `original()`, not `original(None)`, and a trainer whose `create_optimizer` takes no parameters keeps working — #786 has a test for exactly that stub. `*args, **kwargs` is more general, but it does not fix anything the merged version gets wrong today, so a second change here is not worth the churn. The PR also reformats about ten unrelated lines of `deepspeed.py`; if you open another PR, keep the diff to the lines the fix needs.

Your note about `functools.wraps` is the part I would not want lost: with it, `inspect.signature` reports the original's `(model=None)`, so a signature test passes even when the wrapper itself takes no arguments. That is a real trap for anyone writing a guard like this, and it is the kind of thing that only comes from checking whether a test can fail.

Neither of you had commented on #784 before starting, which is how two people ended up building the same fix in parallel. [CONTRIBUTING.md's "Claiming an Issue"](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTING.md#claiming-an-issue) exists for exactly this: a one-line comment is the claim. If you want to pick something up, the [`good first issue`](https://github.com/MakazhanAlpamys/Soup/labels/good%20first%20issue) and [`help wanted`](https://github.com/MakazhanAlpamys/Soup/labels/help%20wanted) labels are the place to start. Comment on one first and I will confirm the scope before you write code.

PR #774, opened 2026-09-10
reimplementation of 7 languages Readme

Ercaner1988:
## Type of change

- [x] Bug fix
- [ ] New feature
- [x ] Refactor / cleanup
- [ x] Documentation
- [ ] Tests

## Checklist

- [ ] `ruff check src/soup_cli/ scripts/ tests/` passes
- [ ] `pytest tests/ -v` passes
- [x ] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed
- [ ] Added a changelog fragment, or this change has no user-visible impact

PR #852, opened 2026-09-11
docs(readme): Turkish translation behind a structural sync ratchet

Ercaner1988:
## What does this PR do?

The version of #774 you said you would merge: **the sync mechanism plus `README.tr.md` alone.** The other five languages follow in separate PRs, each a real translation of the current `README.md`, once this ratchet has proven it can fail. It replaces #774 and #769, which I am closing in its favour so there is one thread.

### The five blocking items from #774

1. **Five files that were not translations** — gone. Only `README.tr.md` is in this diff.
2. **Nothing linked to the translations** — `README.md` gains one line, a centred language banner above the logo. It lists only languages that exist, so there are no dead links while the other five are pending, and a test locks it (below).
3. **Red at head, and CRLF** — `readme_sha256` hashes `README.md` over LF-normalised bytes (`replace(b"\r\n", b"\n")`), your suggestion rather than a `.gitattributes` rule. Stamped against the current `main`. Verified on a Windows checkout with `core.autocrlf=true`, where `README.md` really is CRLF on disk.
4. **A new language escaped the gate** — files are enumerated with `sorted(ROOT.glob("README.*.md"))`. A stampless `README.de.md` now fails; it is one of the control cases.
5. **ruff** — `ruff check src/soup_cli/ scripts/ tests/ benchmarks/` is clean.

The "test for your test": `TestTheRatchetCanActuallyFail` builds a two-language temp tree, asserts it clean, then breaks exactly one thing per test — deleted stamp (fails **both** stamp checks), one byte appended to `README.md` (the message names the file and both sha prefixes), a well-formed wrong stamp, CRLF line endings (must *not* fail), an unstamped `README.de.md`, a dropped code block, a changed URL, a translated command, a dropped section, a banner pointing at a missing file, a banner missing a link, and a control that rewording prose is not policed.

### Beyond the stamp: the gate now checks what a file was synced *from*

You wrote that the sha "verifies when a file was synced, never what it was synced from", and that this was the one thing the gate structurally could not catch. It can now, partly. Each `## ` section of a translation is compared with the section of `README.md` at the same position, on facts translation does not change:

- the **sequence of fenced-code languages**,
- the **set of external URLs**,
- the **set of inline-code spans, in both directions** — nothing dropped, nothing invented.

Commands, flags and URLs are not translated, so a lost or altered one is lost content — and it is the error machine translation is most likely to make, which matters for the languages that follow. Prose is deliberately not policed.

**It is not decorative.** Run against the `README.tr.md` from #774 — the one measured as faithful, with all eighteen headings in order — it fails in **five of its nineteen sections**: **29 inline-code spans dropped and 3 invented, 5 URLs dropped and 2 invented, and one code block missing**. Heading counts could not see any of them. Among them: two *What's New* bullets that do not exist in `README.md` at all, one claiming that `soup diff` and `soup eval` support layer streaming; the SSRF bullet naming the wrong components; `soup init` / `soup train` missing from Quick Start; and a fork-only contributors table. So the 362-vs-521 gap was partly omission, not only compression. The `README.tr.md` in this PR is a fresh translation of the current `README.md` and passes.

### Who maintains each language

- **tr** — @Ercaner1988, native speaker. I read this file end to end before pushing.

### A number worth seeing before relying on this gate

`README.md` changed in **13 of the last 13 weeks** — a median of 2 commits a week, 12 in the busiest — and **54 of those 60 commits are direct pushes to `main`**. With a synchronous stamp, every one of them turns `main` red until the translation is re-synced.

That is the design you asked for, and it is what this PR implements. It also means the "weekly schedule" I wrote on #774 was wrong: at this rate weekly means red most of every week. I will re-sync **the same day `README.md` changes**, agent-assisted, and correct that here rather than leave it standing.

If same-day turns out not to hold, one option for later, **not in this PR**: a synced date inside the stamp (`sha256:<hex> synced:<date>`) and a gate that goes red only once the stamp is more than seven days behind. It would have to live in the stamp, because every workflow checks out with the default shallow `actions/checkout` and there is no history to date a `README.md` revision from. It would also be the first wall-clock-dependent assertion in this suite, which is a real cost. Your call either way; nothing here depends on it.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [x] `ruff check src/soup_cli/ scripts/ tests/` passes (ran with `benchmarks/` too)
- [ ] `pytest tests/ -v` passes — run locally on Python 3.12 without ML extras: the new file passes (16 tests). Every other test file that mentions a README was also run; 54 of those fail, and a pristine `upstream/main` checkout fails the same 54 in this environment (missing optional crypto dependencies — `ed25519` signing and `ModuleNotFoundError`), so none are caused here. The torch/MLX suites are left to CI.
- [x] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed — `README.md` gains only the banner line
- [x] Added a changelog fragment — `changelog.d/0.74.0/852.added.md`

Closes the sync-gate request on #769. Supersedes #774.


MakazhanAlpamys:
Merged as ac4bd1ef — thank you, @Ercaner1988. Your CONTRIBUTORS.md entry has a new line.

This is the narrow version I said I would take in #769, and it answers the critique I raised in #774 better than I asked:

- **The structure check is what a stamp cannot provide.** A stamp proves *when* a file was synced, not *what* it contains. Checking heading sequence, code-fence languages and inline-code spans against `README.md` is what makes the ratchet mean something. Run against the #774 Turkish draft, it finds 29 dropped and 3 invented inline-code spans that heading counts alone would never see.
- **The controls now call the real predicates** over a temp tree, not helper functions on scratch files. Forcing any of `stale_stamps`, `structure_problems` or `missing_stamps` to report nothing turns the controls red.
- **Glob discovery** means a future `README.xx.md` cannot land without a stamp. **LF-normalised hashing** means the stamp survives `core.autocrlf` on Windows. I checked that on a CRLF checkout.
- **You put the cost into numbers yourself**: 54 of the last 60 `README.md` commits were direct pushes to `main`, so each one will now turn `main` red until the Turkish file is re-synced. Disclosing that rather than letting it be discovered is the reason I am comfortable merging.

What I ran before merging:

- **Mutations**, restored afterwards. Deleting the stamp fails 2 tests. A new `##` in `README.md` fails both the stamp and structure checks, and with a re-stamp it still fails on structure. Deleting or reordering a `##` section in `README.tr.md` fails. A command change inside a fence fails through the stamp.
- **Content**: 18/18 `##` and 3/3 `###` headings in order, 13/13 fences, commands byte-identical with only `#` comments translated, all 37 relative links and localised anchors resolving.
- **CI**: all 13 required checks green on `62efc063`, the head I reviewed. A temporary merge with current `main` also passes.

The one gap, so it is written down: a re-stamp **without** re-translating is caught only structurally. Changing `119.6` to `999.9`, or `soup train` to `soup trainx` inside a fence, and then re-stamping, stays green. If you want a follow-up, comparing fence contents with `#` comments stripped, plus relative-link targets, would close most of it.

Two small things. Turkish number formatting turns `48,241 MiB` into `48.241 MiB`, which a reader in a mixed locale can take for roughly 48. For measured figures I would keep the English digits. And "tr — @Ercaner1988" currently lives only in the PR body. One line in the test's docstring would tell whoever sees a red `main` whom to ping.

Issue #40, opened 2026-04-21
v0.27.1: multi-node accelerate launch + end-to-end validation


MakazhanAlpamys:
## Context

v0.27.0 `build_accelerate_argv(num_processes, script_args, mixed_precision, num_machines)` does accept `num_machines > 1` and emits `--num_machines N` in the argv, but the multi-node path is not tested end-to-end:
- no `--machine_rank`, `--main_process_ip`, `--main_process_port` pass-through
- no rendezvous backend selection
- no test case that exercises a 2-node setup

## Proposal for v0.27.1

1. Extend `build_accelerate_argv` with optional `machine_rank: int`, `main_process_ip: str`, `main_process_port: int`
2. Add `--nodes N --node-rank K --master-addr X --master-port Y` flags to `soup train` that map through
3. Test a 2-machine launch via mocked `subprocess` + assert both nodes receive matching NCCL env + argv

## Acceptance criteria

- [ ] `soup train --gpus 8 --nodes 2 --node-rank 0 --master-addr head.internal --master-port 29500` prints a valid `accelerate launch` command for a 16-GPU 2-node setup
- [ ] NCCL `NCCL_IB_DISABLE=0` suggestion when `num_machines > 1` (InfiniBand is typically the inter-node transport)
- [ ] Clear error message when `master-addr` is missing but `nodes > 1`
- [ ] New test class `TestMultiNodeLauncher` in `test_multi_gpu.py`

## Risk notes

- Multi-node testing requires either real hardware or a mocked `subprocess` + env-var assertions
- Firewall / port rules are the user's responsibility; doc it

tristangrech:
I'll take this one. I checked the thread and open PRs and found no existing claim or overlapping multi-node PR.

The gap is still present on `main` at `e0fd219`: `build_accelerate_argv` accepts `num_machines`, but not the node rank or coordinator address/port, and `soup train` has no corresponding CLI options.

My scope:

- Add `--nodes`, `--node-rank`, `--master-addr`, and `--master-port`, with validation and forwarding into the Accelerate launch arguments.
- Cover the issue's two-node example, including total process count, matching coordinator settings, and distinct node ranks.
- Keep the printed launch advice consistent with auto-reexec, and preserve the dry-run and single-node behavior.
- Add `TestMultiNodeLauncher`, the requested NCCL advisory, and documentation of network prerequisites.

[...]

PR #853, opened 2026-09-11
feat(train): add multi-node Accelerate launcher support

tristangrech:
## What this changes

Implements the CPU-verifiable launcher work agreed in #40. This does **not** claim
end-to-end training across two machines.

- Adds `--nodes`, `--node-rank`, `--master-addr`, and `--master-port` to `soup train`.
- Treats `--gpus` as the per-node count. Eight GPUs on each of two nodes produces
  `--num_processes 16 --num_machines 2`, with distinct ranks and matching coordinator settings.
- Selects Accelerate's multi-GPU path and static rendezvous explicitly, including
  the one-GPU-per-node case.
- Validates node counts, ranks, coordinator addresses, and ports before launching.
- Suggests `NCCL_IB_DISABLE=0` for multiple nodes and preserves user environment
  overrides. Dry-run and advice-only modes neither launch workers nor change NCCL settings.
- Documents commands for both nodes, per-node dependencies/data, and network/firewall prerequisites.

The four cluster flags configure Accelerate itself and are not repeated in its
training subprocesses. Multi-node advice uses the same module-form argv as automatic
execution. Existing single-node argument construction and launch-hint filtering remain intact.

## Verification

`TestMultiNodeLauncher` contains 46 CPU-only cases. They exercise the real CLI and
argument builder, replacing hardware probes, dataset loading, and the final `os.execvp`
boundary. Assertions compare complete argv lists and launch environments. CLI output
assertions strip ANSI and collapse whitespace, including Rich panel wrapping.

Deliberate mutations were applied, tested, and reverted:

| Mutation | Observed result |
| --- | --- |
| Drop `--machine_rank` from the builder | Both exact per-node argv tests fail |
| Pass local GPU count instead of GPUs times nodes | The CLI argv test fails on 8 versus 16 processes |
| Replace NCCL `setdefault` with unconditional assignment | The user-override test fails |

- Broader launcher/CLI/config regression run: **180 passed, 1 hardware-gated test skipped**.
- Accelerate 1.15.0's real parser and launch preparation accepted both ranks at one
  and eight GPUs per node, with the expected local worker counts and rendezvous endpoint.
  The port probe was replaced and no workers were launched.
- `ruff check src/soup_cli/ scripts/ tests/ benchmarks/`: clean.
- `git diff --check`: clean.
- `uv build --wheel`: built `soup_cli-0.74.0-py3-none-any.whl`.
- Installed dependency compatibility check: clean.
- Targeted mypy check: the same 12 pre-existing `train.py` findings on this branch
  and an untouched worktree at `f14fa9f`; no added findings. The launcher and topology
  helpers are clean. This matches the repository's non-blocking type-check policy.
- Full non-smoke suite on the final tree:
  `python -m pytest -o addopts= -m 'not smoke' tests/ -q --tb=short`:
  **20,937 passed, 128 skipped, 5 deselected** on macOS arm64 / Python 3.11.15.
- The preceding coverage-enabled run measured **83.19%** against the 77% gate.
  Its only failure was the new changelog fragment's required format. I corrected
  the fragment, reran its checks, and then reran the entire suite above. No Python
  source changed between those two full runs.

## Scope not validated

No two-machine GPU training run, InfiniBand transport test, or cross-host TCP test was
performed. Those remain separate hardware validation, as requested in the issue discussion.
No GPU instance was provisioned for this contribution.

Refs #40.


MakazhanAlpamys:
Reviewed at `c8c3639f`. The launcher half of #40 is done, and done the way the claim thread asked. One blocker, and it is not yours: it arrived on `main` about an hour ago.

## Blocking: rebase onto `main`, then drop the README paragraph

#852 (merged at `ac4bd1ef`) put `README.md` behind a translation sync ratchet. Any change to `README.md` now fails `tests/test_readme_translation_sync.py` until `README.tr.md` is re-synced. Your branch is based on `f14fa9fe`, so its own CI cannot see this. A simulated merge with current `main` fails twice:

```
stamp:     README.tr.md 7a1e969c... != README.md 78282bed...
structure: README.tr.md § 'Common Commands' is missing ['--gpus', '--master-addr', '--master-port', '--node-rank', '--nodes']
```

Please **drop the new paragraph from `README.md`** (currently `README.md:348-350`) rather than re-syncing the Turkish file. `README.md` is already over the ~500-line prose ceiling it is held to, and `docs/commands.md` plus `docs/performance-and-quantization.md#multi-node-launch` already cover the flags. That is where they belong.

## Should fix in the same push

1. **Changelog fragment name and credit.** It is `changelog.d/0.74.0/40.added.md`; the convention in `changelog.d/README.md` is the **PR** number, so `853.added.md`, and the line should end `(#40 by @tristangrech in #853)`.
2. **`--master-addr` validation is looser than its own error message.** The message promises "a hostname or IP address", but the check accepts `head:29500`, `$(whoami)`, `a|b` and a 5,000-character string. None of this is a security hole — the argv goes to `execvp`, the printed hint is `shlex.quote`d, and the output is markup-escaped. But `head:29500` becomes `rdzv_endpoint head:29500:29601` and fails late, on both nodes, which is the expensive place to fail. Validate with `ipaddress` or an RFC 1123 hostname pattern (at most 253 characters), and bracket IPv6 if you accept it.

## Nits, your call

- The node-rank error does not echo the value it received.
- `Annotated[...]` option style is mixed into a file that otherwise uses `= typer.Option(...)`.
- `elif not dry_run:` at `train.py:1062` means a `--dry-run` inside an already-launched rank no longer announces or sets the NCCL env. That is benign, and a test pins it, but it is outside #40's scope, so please mention it in the PR body.

## What I verified

- **Exact argv, not substrings.** Mutations, all reverted afterwards, all killed:

  | mutation | failures |
  |---|---|
  | dropping `--machine_rank` | 9 |
  | swapping IP and port | 9 |
  | moving `--machine_rank` after the `--module` target | 9 (the exact list catches the position) |
  | removing master-addr validation | 7 |
  | single node emitting `--num_machines 1` or `--rdzv_backend static` | killed by `test_single_node_control_is_unchanged` and the #372 test |
  | the CLI passing `machine_rank=None` | 4 |
  | multi-node setting `NCCL_IB_DISABLE=1` | 6 |
  | the printed hint dropping the rank | 1 |
  | skipping up-front validation | 6 |

- **Single node is byte-identical to `main`.** Builder output for 1, 4 and 4+bf16 processes, the `--gpus 4` execvp argv, and the dry-run outputs all match. The only difference is the stale "soup does not auto-re-exec in v0.27.0" footer becoming "Run this command to start training", which is a correction.
- **The printed hint is the argv that launches.** `shlex.split(printed) == captured_execvp_argv` is asserted, so the #372 lesson holds.
- **Cross-checked against the real accelerate 1.15.0 parser.** `--num_processes 16 --num_machines 2 --machine_rank 1 --main_process_ip 10.0.0.1 --main_process_port 29500` resolves to `nproc_per_node 8, nnodes 2, node_rank 1`.
- **Error paths exit 1 naming the flag**: node-rank equal to nodes, negative values, port 0 / 65536, rank or address without `--nodes > 1`, `--nodes 2` without `--gpus`, and `--cloud` / `--find-lr` with multiple nodes. No `shell=True` anywhere on the path.
- ANSI-safe under `FORCE_COLOR=1 TERM=xterm-256color`; the AST ratchet passes; 399 passed / 2 skipped across the launcher, #372, startup-lightness, device-map and deepspeed suites. No `config/schema.py` change — the four options are CLI-only.
- And thank you for writing "No two-machine GPU training run ... was performed" and using `Refs #40`. That is exactly the scoping the claim asked for.

MakazhanAlpamys:
Two notes while this is waiting on the rebase I asked for.

**1. Your fragment is now stale, and that part is the release's doing rather than yours.** v0.75.0 merged to `main` (`f50a7da3`) at 08:26 UTC and emptied `changelog.d/0.74.0/`, so `test_repository_fragments_match_the_newest_release` will fail on the next CI run.

**2. It is also named after the issue rather than the PR.** `CONTRIBUTING.md` asks for the PR number, and every sibling on `main` follows it. Both fix together:

```
git mv changelog.d/0.74.0/40.added.md changelog.d/0.75.0/853.added.md
```

The reason it matters beyond tidiness: `scripts/assemble_changelog.py` sorts by the filename number, so `40` would place a new feature among changes from long ago, and release-time attribution recovery walks fragment filename to PR — `gh pr view 40` resolves to something else entirely.

**`main` is still frozen** — merged release, no tag yet — so nothing goes in before the tag regardless. Rebase when convenient.

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:
@tristangrech — a real status check, because my last two comments here were fragment-broadcast notices and those are not review.

The substantive item from **2026-09-11** is still open, and it is one thing:

> `--master-addr` validation is looser than its own error message. It accepts `head:29500`, `$(whoami)`, `a|b` and a 5,000-character string.

That is the only blocking item I have. Alongside it there are three mechanical ones: the branch needs a rebase, the README change should come out, and the fragment is at `changelog.d/0.74.0/40.added.md` — two problems there, since `0.74.0/` was deleted by the v0.75.0 release (it will fail the repo fragment guard in every matrix cell) and the file is named after the issue rather than the PR. It should be `changelog.d/0.75.0/853.added.md`.

Your CI tick is from 09-11 and predates the release, so it will go red the moment it re-runs regardless of anything else.

Multi-node launcher support is a feature I want and nobody else is working on, so this is not a PR I am looking to age out. Are you still on it? If it has gone cold or turned out larger than expected, say so and I will either close it with the reason recorded or pick up the four items myself and merge it credited to you — I would rather do that than leave a wanted feature sitting for a validation regex.

MakazhanAlpamys:
@tristangrech — I re-measured the blocker rather than repeating it. **Still open**, and I have the patch, so this is a concrete ask.

## Reproduction

The validator at `src/soup_cli/utils/launcher.py:41-47` is the only gate — `commands/train.py:521` is the sole call site. Calling it directly:

```
ACCEPTED   'head:29500'        port smuggled into the address
ACCEPTED   '$(whoami)'         command substitution
ACCEPTED   'a|b'               shell pipe
ACCEPTED   <5000-char string>
ACCEPTED   '`id`'              backtick substitution
ACCEPTED   'a;b'               semicolon
REJECTED   '--num_processes'   flag injection (leading dash)
```

Two more than I reported last time. Only the leading-dash case is caught, while the message promises "a hostname or IP address" — none of those six is one. The value is interpolated into an `accelerate launch` argv and exported as `MASTER_ADDR`, so anything unaddressable is a typo at best.

## The fix, and the trap inside it

**A character blacklist cannot do this job.** IPv6 literals are full of colons, so banning `:` would reject `fe80::1` while `head:29500` — a host with the port glued on, the single most common way to get this flag wrong — is exactly what must be rejected. `ipaddress` first, RFC 1123 labels second:

```python
def _is_hostname_or_ip(value: object) -> bool:
    if not isinstance(value, str) or not value:
        return False
    try:
        ipaddress.ip_address(value)
        return True
    except ValueError:
        pass
    host = value[:-1] if value.endswith(".") else value   # trailing root dot
    if not host or len(host) > MAX_HOSTNAME_LEN:          # 253, RFC 1035
        return False
    return all(_HOSTNAME_LABEL.fullmatch(label) for label in host.split("."))
```

The call site becomes one line: `if not _is_hostname_or_ip(main_process_ip):`.

After: all six hostile values rejected; `10.0.0.5`, `192.168.1.1`, `::1`, `fe80::1`, `2001:db8::1`, `headnode`, `head.cluster.local`, `node-01.eu-west.example.com` and `head.cluster.local.` all still accepted.

```
BEFORE: 9 failed, 12 passed
AFTER:  156 passed  (test_multi_gpu.py + test_issue372_no_reexec_hint.py, zero regressions)
```

## One mutation survived, and the reason is worth recording

| mutation | result |
|---|---|
| `_HOSTNAME_LABEL` check to `return True` | killed, 4x `DID NOT RAISE` |
| gate back to the old emptiness-only check | killed, 4x `DID NOT RAISE` |
| total-length ceiling removed | **SURVIVED** |

That was a weakness in *my* test, not in the gate. `"A" * 5000` is a single 5000-character label, already killed by the 63-character label cap, so it never exercised the total-length ceiling at all. This is the v0.75.0 difflib situation again — except there the gate genuinely was redundant and got deleted, whereas here it is not: 100 legal two-character labels is 299 characters and clears every per-label check. So I strengthened the test instead:

```python
"A" * 5000,                    # one oversized label
".".join(["ab"] * 100),        # 299 chars of individually LEGAL labels:
                               # only the RFC 1035 total ceiling rejects this
```

Re-run: the mutation is now killed (`1 failed, 21 passed`), unmutated 22 passed.

## Separately — the changelog fragment blocks the next release

`changelog.d/0.74.0/40.added.md`. `git merge-tree` reports a clean merge, but that is textual only; `main` is at `[0.75.0]` and assembly on the merged tree fails:

```
ChangelogError -> changelog.d\0.74.0\40.added.md: fragment baseline is 0.74.0,
```

`publish.yml --check-empty` runs that **before** `python -m build`, so merging as-is breaks the next tag. It needs to become `changelog.d/0.75.0/853.added.md` with the body ending `(#40 by @tristangrech in #853)` — named after the PR, not the issue.

## Everything else is clean

No deleted or weakened tests. 690 passed / 4 skipped across all ten launcher-touching files. ANSI handling is correct — your added assertion uses `from tests.conftest import strip_ansi` with whitespace collapse and `│` stripping. (`tests/test_multi_gpu.py:1025` does assert on raw `result.output`, but that line is pre-existing on `main` and not yours.)

The feature itself is fine; this is a 25-line helper, a two-line call site and a `git mv`. **Say whether you want to apply it or would rather I pushed it to your branch** — your commits are untouched either way and your name stays on the PR. If I have not heard by **2026-09-17 18:00 UTC** I will push it and say so here.

MakazhanAlpamys:
@tristangrech — **the fix is written, tested and sitting on a branch here.** Nothing is pushed. You can take it, rewrite it, or tell me to push it; the deadline of 2026-09-17 18:00 UTC stands and is not a threat, just a date so this stops ageing.

Posting the measurements because they are useful to you whichever way you go — and because I found **more holes than I reported last time**.

## Before and after, calling the validator directly

| value | before | after |
|---|---|---|
| `head:29500` | **accept** | reject |
| `$(whoami)` | **accept** | reject |
| `` `id` `` | **accept** | reject |
| `a;b` | **accept** | reject |
| `a\|b` | **accept** | reject |
| `"A"*5000` | **accept** | reject |
| `".".join(["ab"]*100)` — 299 chars, every label legal | **accept** | reject |
| `..` — empty labels | **accept** | reject |
| `node_01` — underscore | **accept** | reject |
| `10.0.0.5`, `::1`, `fe80::1`, `2001:db8::1` | accept | accept |
| `headnode`, `head.cluster.local`, `node-01.eu-west.example.com`, `head.cluster.local.` | accept | accept |

**Hostile accepted 9 of 12 before, 0 of 12 after. Legitimate rejected 0 of 9 in both directions** — that second number is the one that matters, because a validator that starts refusing real cluster addresses is worse than the hole.

`..` and `node_01` are new since my last comment; I had not tried them.

The approach is positive, not a blacklist: `ipaddress.ip_address()` first for both families, then RFC 1123 labels with a 253-character total and a trailing root dot stripped. **A blacklist cannot do this job** — banning `:` would kill `fe80::1` while `head:29500`, the single most common way to get this flag wrong, is exactly the value that has to die.

## The mutation that mattered

| mutation | result |
|---|---|
| label check to `return True` | 11 failed |
| restore the old emptiness-only gate | 13 failed |
| **remove the 253-char total ceiling** | **1 failed** |

That third one **survived** on my earlier attempt, and the reason is worth keeping: the only long fixture was `"A" * 5000`, which is a single oversized label already killed by the 63-character per-label cap, so it never exercised the total ceiling at all. `".".join(["ab"] * 100)` — 299 characters of individually legal labels — is now the only thing in the suite that can detect that gate disappearing. A test can look thorough and test nothing.

## The fragment, which is the release-blocking half

Reproduced on a rebased tree:

```
python scripts/assemble_changelog.py --check-empty
error: changelog.d\0.74.0\40.added.md: fragment baseline is 0.74.0, but the latest
release is 0.75.0; rebase and move the fragment before merging
exit=1
```

`publish.yml --check-empty` runs before `python -m build`, so merging as-is breaks the next tag. Moved with `git mv` to `changelog.d/0.75.0/853.added.md` — named after the PR, not the issue — ending `(#40 by @tristangrech in #853)`. `tests/test_issue487_changelog_fragments.py` then passes at 21, and no BOM (`2d20 60`).

## Shape of the branch

Your commit is **rebased, not squashed**: `de524c19` still carries your authorship, message and content byte-identical, with my fix as a separate commit on top. 711 passed across the 11 launcher-touching test files, ruff clean.

Two small things I added beyond the fix, say if you disagree: three lines in `docs/performance-and-quantization.md` stating that `--master-addr` takes an address only and that `head:29500` is rejected — the rule should be findable where the flag is — and a corrected comment on `MAX_HOSTNAME_LEN`, which said "RFC 1035 wire-format ceiling" when 253 is the presentation form (255 wire octets less the leading length and root bytes).

## What is not verified, plainly

No real multi-node launch — there is no second machine here, so `accelerate launch --nodes 2` was never run end to end. The coverage is the validator, `build_accelerate_argv` and your own tests. And scoped IPv6 (`fe80::1%eth0`) is accepted by `ipaddress` on this Python; I left it accepted rather than adding a rule I could not test.

The multi-node feature is yours and stays yours in the history either way.

MakazhanAlpamys:
Status update with **one correction against myself** and one new fact about your branch. The 2026-09-17 18:00 UTC date stands and nothing has been pushed.

## I was wrong about whose line it was

On 09-11 I asked for the README paragraph to be dropped, and when I re-checked my own fix branch this morning I found it still carried it and assumed I had failed to remove my own edit. I had not: `git log -S` puts those three lines in **`de524c19` — your commit**. My fix commit never touched `README.md` at all; the three-dot diff showed it because it is cumulative over the branch. So the paragraph was carried forward by me, not authored by me, and my 09-11 request was a request to *you* that I then did not enforce on my own branch. Recording that because I would rather be precise about provenance than quietly right.

## Why it has to go — measured, not argued

It is what turns the merged tree red. Against `origin/main` at `f8226214`:

```
merged with your head          2 failed, 198 passed
FAILED tests/test_readme_translation_sync.py::test_every_stamp_matches_the_current_readme
FAILED tests/test_readme_translation_sync.py::test_every_translation_mirrors_readme_structure

  README.tr.md § 'Common Commands': relative link missing
      ['docs/performance-and-quantization.md#multi-node-launch']
  README.tr.md § 'Common Commands': inline code missing
      ['--gpus', '--master-addr', '--master-port', '--node-rank', '--nodes']
```

That is #852's structural gate, which landed after you branched: every `README.md` edit now requires `README.tr.md` to be re-translated **and** re-stamped. Your three lines are good documentation — they are simply the most expensive three lines in the repo, because they oblige a Turkish re-translation for a paragraph that already exists in full at `docs/performance-and-quantization.md#multi-node-launch`, which your own PR adds and which the README links to two lines above.

Delete them and it is **200 passed**, README lands at **476 lines** (the documented anchor), and `git diff origin/main -- README.md` is empty.

## Your CI green is stale, and this is why you could not have known

Run `34596245329` is on your head `c8c3639f` but dated **2026-09-11T11:54Z** — before #852 existed and before v0.75.0 emptied `changelog.d/0.74.0/`. Nothing has re-run since. A green mark from before a ratchet shipped is not evidence about today's tree, which is exactly the trap: `git merge-tree` returns a clean SHA here, so the merge is textually fine and still fails.

The other stale-CI casualty is your fragment: `changelog.d/0.74.0/40.added.md` is in a directory v0.75.0 emptied, and it is named after the issue rather than the PR. It needs to be `changelog.d/0.75.0/853.added.md` with `(#40 by @tristangrech in #853)` in the text.

## What is on the branch here, unpushed

`pr853v`, HEAD `226057b0`, three commits: yours rebased with authorship intact, my validator fix, and now the README removal. Merged with `f8226214` it is **200 passed, ruff clean**. The validator is mutation-checked — replacing the hostname label check with `return True` fails **15 tests**, including all ten hostile `--master-addr` values (`head:29500`, `$(whoami)`, `` `id` ``, `a|b`, `a;b`, `host name`, `-oProxyCommand`, `-`, `..`, `node_01`), so the coverage is parametrized rather than one happy path.

Your head accepts **9 of those 10**; only `--num_processes` is rejected. That is the substance of the fix and it is not a style disagreement.

## The ask is unchanged and small

Three things, all on your side, none of them design work: delete the README paragraph, move the fragment to `changelog.d/0.75.0/853.added.md`, and take or reject the validator. **Or say one word and I push the branch** — your commits stay untouched and your name stays on the PR either way.

You have not replied since 09-11 and five comments have gone by. If I have not heard from you by **2026-09-17 18:00 UTC** I will push and say so here. That is a date, not a threat — the PR is good work and I do not want it to age out.

MakazhanAlpamys:
@tristangrech, it is 2026-09-17 18:00 UTC, the date I gave twice on this thread, and I have not heard back. As announced, I have pushed the fix to your branch (`feat/40-multi-node-launcher`, maintainer edits enabled).

**Your commit is unchanged.** `git range-diff` shows de524c19 identical: `--nodes`, `--node-rank`, `--master-addr` and `--master-port`, `--gpus` as a per-node count, and explicit static rendezvous are all yours. Two commits sit on top of it:
1. **`--master-addr` is validated as an address, not only as non-empty.** Hostile values are accepted in 0 of 12 cases, where 9 of 12 got through before. The legitimate IPv4, IPv6 and RFC 1123 hostnames are rejected in 0 of 9 cases, before and after.
2. **The multi-node paragraph is dropped from `README.md`.** The #852 translation gate would otherwise require a Turkish re-translation of text that already exists in full in `docs/performance-and-quantization.md#multi-node-launch`. The lines were yours, carried forward. I said on 09-16 that my earlier request had been aimed at you, not at my own edit.

This was a force push rather than a fast-forward, because the branch was rebased onto a newer `main`. I used `--force-with-lease` against your c8c3639f.

Verified on the tree merged with current `main`: F821 clean; 313 passed, 1 skipped across the launcher, topology, #372, doctor and README sync tests; `README.md` is 476 lines. Host-validator mutations: replacing the label check with `return True` fails 15 tests, gating on emptiness only fails 17, and removing the length ceiling fails 1.

I will merge once CI is green on the new head. If you would rather change any of this, say so here; the credit for the launcher stays yours either way.

MakazhanAlpamys:
Merged. Thank you for the feature, and an honest accounting of who wrote what, because part of this is mine and you should not have to reconstruct that from the diff.

**Yours, untouched:** `de524c19` — the multi-node launch itself: `--nodes`, `--node-rank`, `--master-addr`, `--master-port`, threaded through to the Accelerate launch path, with the validation and the tests around it. That commit was not rewritten, squashed into, or edited.

**Mine, pushed to your branch on 2026-09-17 at 18:00 UTC, the date announced on this thread in advance:**
- `f353e4e8` — `--master-addr` was validated as "not a dash", which accepts a shell fragment or a URL as a host. It now validates as an address.
- `226057b0` — removed the paragraph this PR had added to the README front door. That was my fault twice over: I asked for a README change without warning you that guard #852 obliges a `README.tr.md` re-translation **and** a re-stamp, so the request reddened all nine test cells, and the front door is not where a multi-node flag belongs anyway. It now lives in `docs/commands.md` and `docs/performance-and-quantization.md`.

CONTRIBUTING.md says a maintainer who finishes someone's work must say so to their face rather than let the merge imply it was all theirs. That is what this comment is. The squash body names the same split, and the credit in the changelog and CONTRIBUTORS.md is yours.

What I ran before merging, on the tree rebuilt by merging your branch into current `main` (GitHub's merge ref is pinned when the run is created, so its green marks described a `main` from before several other merges):

- `tests/test_issue372_no_reexec_hint.py tests/test_multi_gpu.py` plus `tests/test_commands_md_covers_registered_commands.py` and `tests/test_issue979_documented_cli_flags_exist.py` — 169 passed.
- CLI / doctor / review blast radius, including the repo-wide Click flag ratchet that landed on `main` this week — 194 passed, 1 skipped.
- Full `ruff check src/soup_cli/ scripts/ tests/ benchmarks/` — clean. `soup train --help` renders all four flags on the merged tree.
- Host-validation mutations kill 15, 17 and 1 test respectively.
- No conflict with #1043, which touches `commands/train.py` in a disjoint region: merging the two in either order produces a byte-identical tree.

If you disagree with either of my two commits, say so and I will change it — the feature is yours and the decisions about it should be too.

Issue #827, opened 2026-09-11
soup doctor cannot parse nvidia-smi 616.x ("CUDA UMD Version: 13.4") and recommends a cu121 index whose newest torch is 2.5.1, below Soup's own floor


MakazhanAlpamys:
## The defect

`commands/doctor.py:346` reads the driver's CUDA version with:

```python

match = re.search(r"CUDA Version:\s*(\d+)\.(\d+)", text or "")

```

## Evidence

nvidia-smi on driver 616.92 (RTX 5070 Laptop, sm_120, Windows 11) prints a different header:

```

| NVIDIA-SMI 616.92                 KMD Version: 616.92        CUDA UMD Version: 13.4     |

```

`"CUDA UMD Version: 13.4"` does not contain the substring `CUDA Version:`, so the parse returns `None`, and `_torch_cuda_wheel_tag(None)` falls back to `cu121` (`doctor.py:361-362`) by design.

## Reproduction

`_parse_cuda_version` on the live nvidia-smi output, the live `_nvidia_smi_cuda_version()`, the resulting wheel tag, and for comparison the parse of an old-style header string `'Driver Version: 573.24   CUDA Version: 12.8'`:

```

parse: None

live: None

tag: cu121

old header parse: (12, 8)

```

`soup doctor` with CUDA hidden from torch (`CUDA_VISIBLE_DEVICES=-1`, `COLUMNS=200`), which reaches the advisory branch (panel borders trimmed):

```

Backend:  CPU only

Warning:  Training will be slow without GPU.

GPU hardware present (NVIDIA GeForce RTX 5070 Laptop GPU) but torch is the CPU build (torch 2.14.0+cu130). To enable your GPU: `pip install torch --index-url

https://download.pytorch.org/whl/cu121` On Windows, PyPI's torch wheel is CPU-only.

```

Three problems in that one panel:

1. **The index cannot satisfy Soup.** `pip index versions torch --index-url .../whl/cu121` lists `2.5.1+cu121` as the newest version. `[train]` requires `torch>=2.6.0`, and #651 documents why 2.5.x breaks trl 0.29.

[...]

PR #854, opened 2026-09-11
fix(doctor): report the [train] extra as one optional group and exit non-zero on missing core deps

### LINKED ISSUE #828: soup doctor on a core-only install reports 7 required packages missing, exits 0, and suggests a bare pip command that drops the [train] upper bounds instead of `pip install "soup-cli[train]"`
## The defect

Since v0.71.0 `pip install soup-cli` is intentionally light. torch, transformers, peft, trl, datasets, bitsandbytes and accelerate moved to the `[train]` extra (`pyproject.toml:30-33`, `53+`). `soup doctor` still lists all seven as `required`, so on a core-only install it reports them as failures. Its "fix" is built by joining its own `DEPS` floors, which ignores the extra.

## Evidence

`commands/doctor.py:25-31` marks the `[train]` packages `True` (required). The suggestion is built at `doctor.py:185-191`:

```python

console.print(

    "\n[dim]Fix all: pip install -U "

    + " ".join(

        f"'{pkg_name}>={min_ver}'" for _, pkg_name, min_ver, required in DEPS if required

    )

    + "[/]"

)

```

## Reproduction

Simulated core-only install. A script set `sys.modules[name] = None` for the seven `[train]` packages and ran `soup doctor` through `CliRunner`, so nothing was uninstalled:

```

│ torch        │   yes    │     -     │ >=2.6.0     │ MISSING       │

│ transformers │   yes    │     -     │ >=5.16.1    │ MISSING       │

...

Found 7 issue(s):

  > Install torch: pip install 'torch>=2.6.0'

  > Install transformers: pip install 'transformers>=5.16.1'

  ...

Fix all: pip install -U 'torch>=2.6.0' 'transformers>=5.16.1' 'peft>=0.20.0' 'trl>=0.29.0' 'datasets>=2.14.0'

'bitsandbytes>=0.41.0' 'accelerate>=0.27.0' 'pydantic>=2.0.0' 'typer>=0.9.0' 'rich>=13.0.0' 'pyyaml>=6.0' 'plotext>=5.2.0'

exit 0

```

## Why it matters

- **It drops the declared ceilings.** `[train]` pins `transformers>=5.16.1,<6.0.0`, `peft>=0.20.0,<1.0.0` and `trl>=0.29.0,<1.0.0`, and core pins `typer>=0.9.0,<0.21.0`. `pip install -U` with bare floors installs the newest of each, including a future major that doctor itself flags as `INCOMPATIBLE` via `_MAX_EXCLUSIVE` (`doctor.py:54-58`) on the next run.

- **It pulls the CPU torch on Windows.** A bare `torch` from PyPI is the CPU build there, which the same command's GPU panel then complains about.

- **It re-pins core packages nobody asked about** (`pydantic`, `typer`, `rich`...) with `-U`.

- **The quoting breaks in cmd.exe.** Single quotes are not quote characters there, so `'torch>=2.6.0'` becomes a redirect to a file named `=2.6.0'`.

- **Exit code 0** with seven required packages missing, so doctor cannot be used as a gate.

- A core-only install is a supported state. The `pyproject.toml` comment calls it "intentionally light: the CLI, config system, and data tools", and `tests/test_cli_startup_is_light.py` enforces it. Reporting it as seven failures is wrong.

Related: `soup version --full` (`cli.py:665-682`) detects six extras (`serve`, `serve-fast`, `data`, `eval`, `deepspeed`, `wandb`) and knows nothing about `train`, `mlx`, `mcp`, `sign`, `pdf` or the rest. This box has `[dev]` installed (`train`, `mcp`, `data`, plus `cryptography` and `reportlab`), and the output is:

```

soup v0.74.0 | Python 3.12.10 | CUDA 13.0 | extras: serve, data

```

## Proposed fix

1. Split `DEPS` into core-required and per-extra groups. A missing `[train]` group is one line, not seven failures: "training stack not installed — `pip install "soup-cli[train]"`", with the CUDA `--index-url` hint when an NVIDIA GPU is present.

2. Build every suggested command from the extra name, never from floors, so the ceilings in `pyproject.toml` apply. Use double quotes, which work in bash, PowerShell and cmd.exe.

3. Only mention packages that are actually missing or out of range.

4. Exit non-zero when a *core* requirement is missing or incompatible.

5. Have `version --full` derive installed extras from `importlib.metadata` `Provides-Extra` / requirement markers, rather than a hand-kept list of six.

## Acceptance criteria

- [ ] A test that simulates missing `[train]` packages and asserts the suggestion is `pip install "soup-cli[train]"`. It fails on current code.

- [ ] No suggested command contains a `[train]` package with a bare floor.

- [ ] A missing core dependency makes `soup doctor` exit non-zero.

- [ ] `soup version --full` lists `train` and `mcp` on a `[dev]` install.

---

wangzhengzhuo05:
## What

`soup doctor` treated the seven `[train]` packages as required core dependencies, so a supported core-only install was reported as seven failures, with a `Fix all` command built from bare floors and single quotes.

## Changes

- `doctor.py`: the training stack moves into `EXTRA_GROUPS` as one optional group. A missing group yields exactly one issue, `pip install "soup-cli[train]"`, instead of seven bare-floor lines. On a detected NVIDIA driver the suggestion is two steps — `pip install torch --index-url <url>`, then `pip install "soup-cli[train]"` — because `--index-url` replaces PyPI and the CUDA index does not host `soup-cli`. An installed-but-out-of-range group member now reports `Upgrade …` / `Downgrade …` (same wording as the core path) instead of staying silent. A missing or INCOMPATIBLE core dependency exits non-zero; optional packages never affect the exit code. Suggestions use double quotes, are printed with `highlight=False`, and mention only entries that are actually missing or out of range.
- `cli.py`: `version --full` derives installed extras from `Provides-Extra` / `Requires-Dist` metadata and matches requirements by distribution name (`importlib.metadata.distribution`), so `scikit-learn`/`sklearn` and `pillow`/`PIL` resolve and `train`/`mcp`/`data` show up on a full install.
- Drift guards `test_issue352`, `test_issue502`, `test_issue636` and `test_issue602` now read the declared floors from `EXTRA_GROUPS`; the `DEPS` torch alias row and the `_EXTRA_PACKAGE_NAMES` skip are gone.
- `tests/test_doctor.py`: out-of-range group reporting, the two-step NVIDIA suggestion, colour-safety under `FORCE_COLOR=1`, and `_installed_extras` against distribution metadata.
- `changelog.d/0.74.0/854.fixed.md`, `docs/commands.md` (the new exit 1).

## Verification

- `.venv/bin/python -m pytest tests/test_doctor.py tests/test_issue352_fsdp_peft_checkpoint.py tests/test_issue502_qwen35_transformers5.py tests/test_issue636_torch_floor.py tests/test_issue487_changelog_fragments.py tests/test_issue602_qwen4_ple_streaming.py -q --no-cov` → the doctor and drift-guard tests pass. The only failures are pre-existing environment ones (`test_doctor_nccl_*`, and the cases that import `torch`/`accelerate`, absent on this core-only box).
- Mutation checks, all killed by the new tests: dropping the group out-of-range issue makes the report silent again; `find_spec` instead of `distribution` breaks the `_installed_extras` test; restoring the single-step suggestion breaks the two-step test.

Fixes #828


MakazhanAlpamys:
Thank you, @wangzhengzhuo05, and welcome. This is the right shape for #828. The exit-code change is limited to **core** dependencies, exactly as it should be. Escaping the markup so `[train]` renders is correct. The metadata-derived extras stay torch-free and survive a missing distribution. And every mutation I ran is caught, including a missing *optional* package accidentally exiting non-zero, which is the one I most wanted a test for.

It is not mergeable yet: CI is red, and the new headline command cannot install.

## Blocking

**1. Three CI failures on `58cface2`** (`test (macos-latest, 3.12)`; the other cells were still running):

```
FAILED tests/test_issue352_fsdp_peft_checkpoint.py::test_doctor_deps_accelerate_floor_matches_declared_extra - KeyError: 'accelerate'
FAILED tests/test_issue502_qwen35_transformers5.py::test_doctor_training_bounds_match_declared_extra - KeyError: 'transformers'
FAILED tests/test_issue487_changelog_fragments.py::test_repository_fragments_match_the_newest_release - ... changelog.d/0.74.0/854.fixed.md: fragment must reference #854
```

The first two build their floors from `DEPS`, and the training rows now live in `EXTRA_GROUPS`. Please point both — and `test_issue636` — at `EXTRA_GROUPS` rather than keeping the torch alias row in `DEPS` just to satisfy them, then delete that alias row and the `_EXTRA_PACKAGE_NAMES` skip. For the fragment, end the line with `(#828 by @wangzhengzhuo05 in #854)`.

**2. `pip install "soup-cli[train]" --index-url https://download.pytorch.org/whl/<tag>` cannot succeed.** `--index-url` *replaces* PyPI, and the PyTorch index hosts torch but not `soup-cli`, `trl`, `peft` or `transformers`:

```
pip install --dry-run "soup-cli[train]" --index-url https://download.pytorch.org/whl/cu130
ERROR: No matching distribution found for soup-cli[train]
```

So on every NVIDIA box where `nvidia-smi` parses, the one command doctor now tells you to run fails, and a mixed "Fix all" line breaks the same way. Make it two steps: `pip install torch --index-url <url>`, then `pip install "soup-cli[train]"`. `--extra-index-url` also works mechanically, but it lets pip pick torch from either index, which is how people end up with a CPU wheel.

Worth knowing while you are there: on a driver `nvidia-smi` cannot parse (616.x, #827) the index is omitted entirely. That is safe, but on Windows it means a plain `pip install "soup-cli[train]"` resolves CPU-only torch. Fixing #827 is out of scope here; just do not make the two-step output depend on it.

## Should fix

- **An out-of-range `[train]` package is now silent.** With `transformers` at 6.1.0, the table shows `INCOMPATIBLE (need <6.0.0)` but the summary says "All checks passed!", where `main` said `Downgrade transformers: ...`. The issue asks to mention packages that are "actually missing or out of range". Group members that are incompatible or outdated currently add no issue (`doctor.py:223-234`), and no test covers it.
- **`_installed_extras` matches distribution names with `find_spec`.** Names that differ from their import name never match — `scikit-learn`/`sklearn`, `pillow`/`PIL` — so `data`, `vision` and `dev` drop out of `version --full` on a full install. `data` was listed on `main`, so this is a regression. `importlib.metadata.distribution(req.name)` fixes it (`cli.py` around line 718).
- **The new test is not colour-safe.** CI did not trip on it, but under `FORCE_COLOR=1 TERM=xterm-256color` Rich's highlighter wraps the quoted command (`pip install \x1b[32m"soup-cli\x1b[0m...`), so the positive match fails and the `"torch>=2.6.0" not in output` checks pass vacuously. Use the ANSI-strip helper from `tests/test_auto_tuning.py`. Printing those two suggestion lines with `highlight=False` would also make them cleaner to copy for users.

## Your call

- An outdated (below-floor) core dependency is listed as an issue but still exits 0 (`doctor.py:188`). I would count below-floor as out of range, but I will not block on it.
- `test_doctor_healthy_core_exit_zero` runs on a full install and never simulates core-only — rename it or make it do what the name says. Label the train rows `[train]` rather than `optional`. And one line about the new exit code in `docs/commands.md` near the `doctor` entry would help CI users find it.

## What I verified

- Criteria 1-4 met. Criterion 1's test fails on `main` and passes here; `sys.modules["plotext"] = None` exits 1; missing optional packages exit 0; no GPU exits 0; and `--nccl`, `--disk` and `--config` (from #756, merged today) keep their exit codes.
- Mutations, all killed: bare per-package floors, a missing core dependency exiting 0, a missing optional package or the train group exiting non-zero, empty or hand-kept extras, and single quotes.
- No test deleted or weakened. The only flipped assertion is `test_doctor_missing_dep` moving to a non-zero exit, which is criterion 3.

A note for next time: #828 had no claim comment when this opened. [CONTRIBUTING.md's "Claiming an Issue"](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTING.md#claiming-an-issue) asks for one line on the issue first, so two people do not build the same fix — that happened to a PR earlier today.

wangzhengzhuo05:
Thanks for the careful review — all of it is addressed in `8485ec6` + `0f3c8b7`.

**Blocking 1 — CI.** Pointed `test_issue352`, `test_issue502` and `test_issue636` at `EXTRA_GROUPS`, deleted the `DEPS` torch alias row and the `_EXTRA_PACKAGE_NAMES` skip. While doing that I found a fourth drift guard reading the deleted row — `test_issue602_qwen4_ple_streaming.py::test_qwen4_oq_torch_floor_matches_project_and_doctor` (it was `StopIteration`-ing) — and pointed it at `EXTRA_GROUPS` too. The fragment now ends with `(#828 by @wangzhengzhuo05 in #854)`.

**Blocking 2 — the two-step install.** The NVIDIA suggestion is now two steps: `pip install torch --index-url <url>`, then `pip install "soup-cli[train]"`. The torch step is carried as its own command, so the `Fix all` block renders it on a separate line and no `--index-url` lands on the `soup-cli` install. With no detected driver it stays the single `pip install "soup-cli[train]"`.

**Should fix — out-of-range `[train]` members.** They now emit an issue and a fix part, with the same wording as the core path, so `transformers` at 6.1.0 reports `Downgrade transformers: pip install "transformers>=5.16.1,<6.0.0"` instead of "All checks passed!". Covered by `test_doctor_out_of_range_train_member_is_reported`.

**Should fix — `_installed_extras`.** Switched to `importlib.metadata.distribution(req.name)`, so `scikit-learn`/`sklearn` and `pillow`/`PIL` match. `test_installed_extras_lists_train_and_mcp` now mocks `distribution` rather than `find_spec`.

**Should fix — colour safety.** The assertions run through a `_strip_ansi` helper, and a new `test_doctor_suggestion_is_colour_safe` runs under `FORCE_COLOR=1 TERM=xterm-256color`. Suggestion lines are printed with `highlight=False`.

**Your call** — the train rows are labelled `[train]`; `test_doctor_healthy_core_exit_zero` is renamed `test_doctor_full_install_exits_zero` (it runs on the full install, and the name now says so); and `docs/commands.md` notes the new exit 1. I left the below-floor core dependency exiting 0 as-is — you said you would not block, and it reads like a separate call. Happy to change it if you would rather treat below-floor as out of range.

**Verification** (this core-only box):

```
.venv/bin/python -m pytest tests/test_doctor.py tests/test_issue352_fsdp_peft_checkpoint.py \
  tests/test_issue502_qwen35_transformers5.py tests/test_issue636_torch_floor.py \
  tests/test_issue487_changelog_fragments.py tests/test_issue602_qwen4_ple_streaming.py -q
```

→ the doctor and drift-guard tests pass. The only failures are pre-existing environment ones (`test_doctor_nccl_*`, plus the cases that import `torch`/`accelerate`, which this box does not have — `ModuleNotFoundError` before and after the change).

- Mutation checks, all three killed by the new tests: (1) drop the group out-of-range issue → silent again, red; (2) `find_spec` instead of `distribution` → `_installed_extras` test red; (3) single-step suggestion → the two-step test red.

Noted on claiming: I will put the one-line claim on the issue first next time.

MakazhanAlpamys:
Re-reviewed at `0f3c8b7a`. Everything I asked for in round 1 is addressed.

- **CI**: 352 / 502 / 636 now read `EXTRA_GROUPS`, the alias row and the skip are gone, and the fragment credits correctly. You also re-pointed the 602 guard on your own, and it is stricter than before.
- **The two-step install**: both commands resolve in a real `pip --dry-run` — torch `2.11.0+cu128` from the PyTorch index, and `soup-cli[train]` from PyPI.
- **Out-of-range packages**: a fake `transformers 6.1.0` now reports `Downgrade transformers: pip install "transformers>=5.16.1,<6.0.0"`.
- **`version --full`** now lists `all, data, dev, generate, mcp, pdf, serve, sign, train, ui, vision` on a full install, where `main` listed `serve, data`.

One new user-facing bug, and three tests that do not test what they say.

## Blocking: "Fix all" now says `pip install "soup-cli"`

The escape on the "Fix all" line was lost in this round, so Rich parses `[train]` as markup and deletes it. Simulated core-only, with the driver parsed and unparsed:

```
> Training stack not installed: pip install torch --index-url https://download.pytorch.org/whl/cu128, then pip install "soup-cli[train]"
Fix all:
  pip install torch --index-url https://download.pytorch.org/whl/cu128
  pip install "soup-cli"          <-- installs the core only
```

The new `[train]` label in the table's Required column is eaten the same way, so those cells render blank. The tests stay green because the issue line, which is still escaped, contains the right string. Escape `f"[{extra_name}]"` in `table.add_row` and in the fix-all `pip install` line, and add a test that asserts the text *after* `Fix all:`.

## Tests that need to bite

- **The colour-safety test runs without colour.** `console = Console()` is built at import time, before `monkeypatch.setenv("FORCE_COLOR", ...)`, so no ANSI is emitted and the test proves nothing. Patch `doctor.console` with `Console(force_terminal=True)` inside the test, then assert on the stripped output.
- **The INCOMPATIBLE branch is uncovered.** Your new test covers the *outdated* `[train]` case. Dropping only the INCOMPATIBLE issue line still passes everything. Add the `transformers 6.1.0` case as a test.
- **`_installed_extras` reverting to `find_spec` survives on a full install.** Every fake requirement in the test has an import name equal to its distribution name, so `find_spec` finds them all anyway. Your "killed" result was real only on your core-only box. Add a requirement whose names differ — `scikit-learn` under `data` is the natural one — and the mutation dies everywhere.

## Minor

- `test_doctor_missing_dep` asserts `exit_code != 0`; `== 1` pins the documented contract.
- The headline issue line wraps mid-command at 80 columns. Printing the two commands on separate lines, as the fix-all block already does, would make them copy cleanly.
- Deferring the outdated-core-dependency exit code is fine by me.

## Verified

- The 9 affected test files: 168 passed, 3 skipped.
- Under `FORCE_COLOR=1`, the doctor tests pass apart from the two `nccl` cases, which fail identically on `main`.
- Mutations still killed: bare floors, a missing core dependency exiting 0, a missing optional package exiting non-zero, empty extras, single quotes, and collapsing the two steps back into one `--index-url` command.
- No assertion weakened. Ruff clean.
- `git merge-tree` against #855 is clean.

Answering every round-1 point in one push, finding a fourth drift guard yourself, and keeping the fix independent of the #827 parser made this round quick to review. These last four are small.

wangzhengzhuo05:
Thanks — all four points are addressed in edfbfcb.

**Blocking: the escape was lost on the "Fix all" line**

Fixed. Both spots now escape the marker — the `Required` column cells via `table.add_row(..., escape(f"[{extra_name}]"), ...)` and the fix-all payload via `escape(' '.join(fix_parts))`. The rendered output is now:

```
Fix all:
  pip install torch --index-url https://download.pytorch.org/whl/cu128
  pip install "soup-cli[train]"
```

and the table's Required column shows `[train]` instead of a blank cell. The headline two-step suggestion is also split onto its own lines, so the 80-column wrap no longer cuts a command in half.

**Tests that now bite**

- The colour-safety test patches `doctor.console` with `Console(force_terminal=True)` inside the test instead of the env vars (the module-level console is built at import time, so `FORCE_COLOR` never took effect and no ANSI was emitted).
- Added `test_doctor_fix_all_line_keeps_extra_marker`, which asserts on the text *after* `Fix all:`.
- Added `test_doctor_incompatible_train_member_is_reported` for the INCOMPATIBLE branch (`transformers` 6.1.0 against the `<6.0.0` ceiling) — dropping that issue line now fails a test.
- `test_installed_extras_lists_train_and_mcp` now carries a `data` extra whose distribution name (`scikit-learn`) differs from its import name, so a revert to `find_spec` fails on every machine, not just a core-only box. (I also had to add `data` to the fake `Provides-Extra`, otherwise the loop never evaluated that extra.)
- I added one more test you didn't ask for: `test_doctor_table_keeps_extra_marker` pins the Required-column cell directly, since the fix-all test alone would not catch a regression there.

**Minor**

- `test_doctor_missing_dep` now pins `exit_code == 1`.
- The two-step suggestion prints its commands on separate lines.

**Verification**

- Mutation checks (each reverted surgically, then restored): un-escape the fix-all payload → fix-all test fails; un-escape the table cell → table test fails; disable the INCOMPATIBLE branch → incompatible test fails; `_installed_extras` back to `find_spec` → extras test fails.
- `pytest tests/test_doctor.py` → 28 passed, 2 failed (`test_doctor_nccl_*`, pre-existing — `torch` is not installed in this venv; they fail identically on `main`).
- `ruff check src/soup_cli/commands/doctor.py tests/test_doctor.py` → clean.

MakazhanAlpamys:
Merged as `12d08ade`. Thank you, @wangzhengzhuo05 — a strong first contribution, and it fixed more than the issue asked.

**What I liked most.** The bug was wrong in both directions at once: it alarmed about a correct core-only install and stayed silent about a genuinely broken one. You fixed both halves, and you separated them properly — a missing or incompatible *core* dependency exits 1, while optional packages and the `[train]` group never touch the exit code. That is the distinction that makes the command usable in CI.

The two-command suggestion is the part I would point other contributors at. The obvious fix is one `pip install --index-url ...` line, and it is wrong, because `--index-url` *replaces* PyPI and the PyTorch index hosts no soup-cli, trl, peft or transformers. You worked that out and then verified both commands resolve under `pip install --dry-run` rather than asserting they would.

You also took the adjacent drift you found without being asked: `soup version --full` derived extras from a hand-kept list, and now reads distribution metadata, so `scikit-learn` and `pillow` resolve instead of silently reporting absent.

**What I ran.** Three rounds of mutations, every one killed: bare per-package floors, a missing core dependency exiting 0, a missing optional package or the train group exiting non-zero, single quotes in the suggestion, collapsing the two steps into one `--index-url` command, an unescaped fix-all payload, dropping the INCOMPATIBLE issue, and reverting extras detection to `find_spec`. The colour-safety test runs against a real `force_terminal` console, which matters here — this repo's CI has gone red four times on Rich's per-character ANSI escapes inside flag names, and Windows hides it because Rich disables colour there.

13/13 required checks green on `edfbfcbc`. Your `CONTRIBUTORS.md` entry is in `b5723ce8`.

One thing you should know, and it is not on you: @blackcoderx opened #873 against the same issue. Yours was reviewed, revised over three rounds and green first, so it is the one that merged. I am writing to them separately.

Issue #776, opened 2026-09-10
Streamed NF4 is not bit-exact vs resident NF4 on Blackwell (sm_120): 4.9e-04 / 3.9e-03, while quantization=none stays exact


MakazhanAlpamys:
`tests/test_issue385_stream_dtype.py::TestFloat16StreamingIsBitExact` requires a streamed run to
match a resident run **exactly** (`== 0.0`). On an RTX 5070 Laptop GPU (Blackwell, `sm_120`) the
`[nf4]` parametrisation fails in both dtypes, while `[none]` passes:

```
test_float16_matches_resident_bit_exactly[none]        PASSED
test_float16_matches_resident_bit_exactly[nf4]         FAILED   assert 0.00048828125 == 0.0
test_bfloat16_still_matches_resident_bit_exactly[none] PASSED
test_bfloat16_still_matches_resident_bit_exactly[nf4]  FAILED   assert 0.00390625   == 0.0
```

**The `none` / `nf4` split is the useful part of this report.** Layer streaming itself is still
exactly bit-exact — the unquantised parametrisations pass. Only the NF4 path diverges, which points
at the quantiser rather than at the streaming machinery.

### Why this has not been seen before

`CUDA = pytest.mark.skipif(not _cuda_available(), ...)` gates the whole module, and GitHub Actions
runners have no GPU, so **these tests skip in CI on every platform**. They only run on a developer
box with a card, and the bit-exactness claim was established on an RTX 3050. This is the first time
the project has had a Blackwell card to run them on.

### What this contradicts

`.claude/CLAUDE.md` states, of `layer_shard.py` / `layer_stream_runtime.py`:

> Streamed NF4 is BIT-EXACT vs RESIDENT NF4 (same bytes, same bnb kernels) — a CI test, not a one-off.

Two readings, and this report does **not** decide between them:

1. The NF4 streaming path genuinely diverges from resident NF4 — a correctness bug in Soup.
2. bitsandbytes' NF4 kernels are not bit-identical across these two call shapes on `sm_120` / this
   bnb build, in which case the invariant is **hardware- and kernel-scoped rather than absolute**,
   and the documentation is what needs correcting.

**Root cause is not established.** The claim in this report is only the measurement above.

### Environment

| | |
|---|---|
| GPU | NVIDIA GeForce RTX 5070 Laptop GPU, `sm_120`, 8151 MiB |
| Driver | 616.92 (CUDA 13.4) |
[...]

PR #857, opened 2026-09-11
test: match NF4 resident reference compute path

Samearth17:
## Summary

Fix the NF4 bit-exactness reference in `TestFloat16StreamingIsBitExact`.

The streamed NF4 path deliberately uses `install_dequant_forward` (`dequantize_4bit + F.linear`), while a resident NF4 `Linear4bit` can use bitsandbytes' native `MatMul4Bit` path. With current bitsandbytes versions, those paths are not necessarily numerically identical at small token counts.

The test was therefore comparing:

* streamed NF4: explicit dequantisation + `F.linear`
* resident NF4: native bitsandbytes dispatch

That means a failure can measure a kernel-path difference rather than a streaming difference.

This change applies `install_dequant_forward` to the resident NF4 reference as well, so the exact `== 0.0` assertion compares matching computation paths.

## Validation

Local CPU/non-CUDA validation passes:

```text
23 passed, 6 skipped
```

Static checks also pass:

* `python -m py_compile tests/test_issue385_stream_dtype.py`
* `git diff --check`

The CUDA-specific test cannot be executed on my current machine because it has no NVIDIA GPU. In particular, I have not claimed Blackwell (`sm_120`) validation.

## Related issue

Refs #776.

The exact-equality assertion is intentionally retained; this PR does not replace it with a tolerance.

### DIFF
--- a/src/soup_cli/utils/layer_stream_runtime.py
+++ b/src/soup_cli/utils/layer_stream_runtime.py
@@ -183,13 +183,12 @@ def install_dequant_forward(module: Any) -> int:
     tensor through the ordinary mechanism, which checkpointing DOES discard and
     recompute, and the transient lives only inside the recomputed block — O(window).
 
-    This is not a numerics change at training shapes. ``bitsandbytes::gemm_4bit``
-    dispatches on M with ``_gemm_4bit_custom_max_m = 1536``, and on real projection
-    shapes it takes ``_dequant_linear_fallback`` at every M measured from 8 to 2048 —
-    i.e. bitsandbytes already does exactly this. The dtype handling below therefore
-    mirrors ``Linear4bit.forward`` line for line, replacing only the final call: a
-    dequantisation into a different dtype than bnb's own would introduce a rounding
-    difference precisely where there is none today.
+    This changes the computation path used by the patched NF4 module. With
+    bitsandbytes 0.50.2, the native fused ``MatMul4Bit`` path and explicit
+    ``dequantize_4bit`` + ``F.linear`` can disagree at small M: measured at
+    K=N=4096, the paths diverge through M=64 and agree at M=2048. The
+    dequantise + linear path is retained for correctness under checkpointing
+    (#331), not because it is numerically free.
 
     Returns the number of modules patched, so a caller can assert it patched
     something. Zero would mean the model carries no 4-bit linears at all.
--- a/tests/test_issue385_stream_dtype.py
+++ b/tests/test_issue385_stream_dtype.py
@@ -449,10 +449,15 @@ def _copy_lora(src, dst):
 
 @CUDA
 class TestFloat16StreamingIsBitExact:
-    """Acceptance item 4 of #385. The reference is a resident model of MATCHING
-    numerics — for NF4 that means a genuinely NF4-quantised reference, because
-    comparing a streamed fp16 run against a bf16 resident one would measure the
-    dtype rather than the streaming."""
+    """Acceptance item 4 of #385.
+
+    The reference matches the streamed model's computation path. For NF4 that
+    means a genuinely NF4-quantised resident reference with
+    ``install_dequant_forward`` applied, because the native bitsandbytes
+    ``MatMul4Bit`` path can use a different fused kernel at small token counts.
+    Comparing those different kernels would measure kernel-path differences
+    rather than layer streaming.
+    """
 
     def _run(self, tmp_path, dtype: str, quant: str) -> float:
         import torch
@@ -462,6 +467,7 @@ class TestFloat16StreamingIsBitExact:
         from soup_cli.utils.layer_stream_runtime import (
             build_meta_skeleton,
             build_streamed_model,
+            install_dequant_forward,
             quantised_layer_suffixes,
         )
 
@@ -502,6 +508,7 @@ class TestFloat16StreamingIsBitExact:
                     weights, quantization_config=build_nf4_config(dtype),
                     dtype=getattr(torch, dtype), device_map={"": "cuda"},
                 )
+                assert install_dequant_forward(base) > 0
             base.config.use_cache = False
             for param in base.parameters():
                 param.requires_grad = False
--- a/tests/test_issue776_bnb_nf4_paths.py
+++ b/tests/test_issue776_bnb_nf4_paths.py
@@ -12,7 +12,6 @@ import importlib.metadata
 
 import pytest
 
-
 torch = pytest.importorskip("torch")
 
 CUDA = pytest.mark.skipif(
--- a/tests/test_issue776_bnb_nf4_paths.py
+++ b/tests/test_issue776_bnb_nf4_paths.py
@@ -1,0 +1,72 @@
+"""#776 — document the two bitsandbytes NF4 forward paths.
+
+The streaming repair in #857 deliberately uses ``dequantize_4bit`` +
+``torch.nn.functional.linear``. A resident NF4 ``Linear4bit`` normally uses
+bitsandbytes' ``matmul_4bit`` path instead. With bitsandbytes 0.50.2 those
+kernels are observably different at small M, so the bit-exact streaming test
+must compare matching computation paths rather than treating this backend
+property as a streaming regression.
+"""
+
+import importlib.metadata
+
+import pytest
+
+
+torch = pytest.importorskip("torch")
+
+CUDA = pytest.mark.skipif(
+    not torch.cuda.is_available(),
+    reason="requires CUDA to exercise bitsandbytes NF4 kernels",
+)
+
+
+@CUDA
+@pytest.mark.parametrize("rows", [64, 2048])
+def test_nf4_fused_and_dequant_paths_are_close(rows):
+    """The divergence is a bitsandbytes 0.50.2 property, not a Soup invariant."""
+    import bitsandbytes as bnb
+    import torch.nn.functional as functional
+
+    bnb_version = importlib.metadata.version("bitsandbytes")
+    if bnb_version != "0.50.2":
+        pytest.skip(
+            "measured kernel-path behaviour is pinned to bitsandbytes 0.50.2"
+        )
+
+    torch.manual_seed(0)
+
+    inputs = torch.randn(
+        rows, 4096, device="cuda", dtype=torch.float16
+    )
+    weights = torch.randn(
+        4096, 4096, device="cuda", dtype=torch.float32
+    )
+
+    packed, quant_state = bnb.functional.quantize_nf4(
+        weights,
+        blocksize=64,
+    )
+
+    fused = bnb.functional.matmul_4bit(
+        inputs,
+        packed,
+        bias=None,
+        quant_state=quant_state,
+    )
+
+    dequant = functional.linear(
+        inputs,
+        bnb.functional.dequantize_4bit(
+            packed,
+            quant_state,
+        ),
+    )
+
+    max_error = (fused.float() - dequant.float()).abs().max().item()
+
+    if rows == 64:
+        assert max_error > 0.0
+        assert max_error <= 1e-2
+    else:
+        assert max_error <= 1e-6


MakazhanAlpamys:
As offered in #776, here is the run on the `sm_120` box, since CI skips this whole module. **Condition 1 holds, and the fix is exactly your one line.**

**Card:** RTX 5070 Laptop (Blackwell, `sm_120`), torch 2.14.0+cu130, bitsandbytes 0.50.2.

| case | `main` | this PR (`4a84efbd`) |
|---|---|---|
| fp16 `[none]` | 0.0 pass | 0.0 pass |
| bf16 `[none]` | 0.0 pass | 0.0 pass |
| fp16 `[nf4]` | **4.8828125e-04 fail** | **0.0 pass** |
| bf16 `[nf4]` | **3.90625e-03 fail** | **0.0 pass** |

The whole module on the PR: 26 passed, 1 skipped (MPS).

- **`== 0.0` is untouched**, with no tolerance, skip or narrowed parametrization.
- **The reference is still an independent resident model** — a separate `from_pretrained` with bitsandbytes quantizing at load time, against offline shards on the streamed side — so the test is not comparing a model with itself.
- **`assert install_dequant_forward(base) > 0`** stops the patch from silently matching nothing.
- **The mechanism is sound.** The streamed side runs `dequantize_4bit` + `F.linear` (`layer_stream_runtime.py:876`), while the old reference ran bitsandbytes' native `MatMul4Bit`. The bytes are identical; the kernels are not.

**The test still catches real streaming bugs at `== 0.0`.** I injected these into the streaming buffer pool only:

- +1 ulp on layer 1's `input_layernorm.weight`: all four cases fail.
- +1 NF4 code in a packed `q_proj` weight: both `[nf4]` cases fail.
- `nested_absmax` × (1 + 1e-3): both `[nf4]` cases fail.
- Layers 0 and 1 swapped in the prefetch: all four fail.
- Reverting only your line brings back `main`'s exact numbers.
- Neutering `install_dequant_forward` on the streamed side is now caught on this card too, so the matched reference turns this test into a guard for #331. That is a bonus I did not ask for.

(For completeness: +1 ulp on a single `q_proj` element in `[none]` is absorbed at logit precision on both `main` and this PR. That is a sensitivity limit of the test, not something this change introduces.)

## Before merge: conditions 2 and 3 from #776

**3 — the docstring, the part I said I care about most.** `install_dequant_forward`'s docstring (around lines 186-192) still reasons from `_gemm_4bit_custom_max_m = 1536`, which does not exist in bitsandbytes 0.50.2. Replace it with what is measured, and state the version:

- the fused path and dequantize + `F.linear` disagree at small `M` on bnb 0.50.2, measured up to `M=64` at `K=N=4096`, and agree at `M=2048`;
- the patch is kept for correctness under checkpointing (#331), not because it is numerically free.

The same numbers correct the test docstring's "at small token counts".

**2 — the fused-vs-dequant divergence as its own test, labelled as a bitsandbytes property.** A tolerance is honest there, and it documents why condition 1 needs a matched reference. It can go in this PR or in a follow-up under #776 — your choice. If it is a follow-up, say so in the PR body.

Also, the body reads "Closes/fixes … #776". GitHub did not link it (`closingIssuesReferences` is empty), which is right, because #776 should stay open until 2 and 3 land. Please make it `Refs #776` so nobody reads it the other way.

MakazhanAlpamys:
@Samearth17 — reviewed at `0f475333` on the project's only GPU box (RTX 5070, sm_120, torch 2.14.0+cu130, bitsandbytes 0.50.2). **The core change is right**, and I verified it adversarially rather than by reading. Two blocking items below, neither of which you could have seen — and one correction that is mine, not yours.

## First: I gave you a wrong fact, and you built on it

In my last review I wrote that `_gemm_4bit_custom_max_m = 1536` "does not exist in bitsandbytes 0.50.2". **That is wrong.** It exists — it moved to `bitsandbytes/backends/cuda/ops.py:926`, with a ROCm value of `256` at `:931`, dispatched at `:954`. I just grepped the installed package to be sure.

What is actually stale in the old docstring is the *conclusion*, not the constant: `M > 1536` still forces the dequant fallback, but at `M <= 1536` bnb 0.50.2 now consults a per-architecture heuristic, so the old claim that it "takes the fallback at every M measured from 8 to 2048" is false on this card. Your correction was right; the reason I gave you for it was not. Sorry — that is the same standard I hold you to, so I would rather state it plainly than quietly move on.

## The central question: does this fix the comparison or weaken the assertion?

**It fixes a genuine apples-to-oranges comparison, and nothing was weakened.** The assertion is still `== 0.0` on both parametrizations; there is no tolerance, no `allclose`, no `xfail`, no narrowed parametrize anywhere in the diff. The only removed lines in `tests/` are docstring prose, and the reference is still an independent `from_pretrained`, not the model compared with itself.

Your "different compute path" claim checks out against the bnb source rather than the PR text: stock `Linear4bit.forward` ends in `bnb.matmul_4bit(...)`, while the streamed side rebinds to `dequantize_4bit(...) → F.linear(...)`. Two kernels, identical bytes. I measured the gap directly at the shape this test runs (K=N=64, M=12): **0.015625 in fp16, 0.125 in bf16** — which is what propagates into #776's `4.88e-04` / `3.90e-03` logit deltas.

**And the test still catches real streaming bugs.** I injected defects into the shard directory only (the streamed side's sole weight source) and every one was caught at `== 0.0`:

| injected defect | caught |
|---|---|
| swap layer_000 ↔ layer_001 | yes, all 4 arms |
| +1 in one packed NF4 nibble of `q_proj` | yes, both nf4 arms |
| `nested_absmax × 1.001` | yes, both nf4 arms |
| `input_layernorm.weight[0] × 1.5` | yes, all 4 arms |

Bonus: neutering `install_dequant_forward` on the **streamed** side alone now fails both nf4 arms with `main`'s exact numbers — so your change turns this test into a guard for **#331** as well, which it was not before. Side by side: `main` → 2 failed / 2 passed; your branch → 26 passed, 1 skipped. Neighbouring streaming suites: 168 passed, 2 skipped.

`Refs #776` is in place and `closingIssuesReferences` is empty, so the issue will not auto-close. Good.

## Two blocking items — and the reason you could not have seen either

You said honestly that you have no NVIDIA GPU, and the `@CUDA` mark skipped the whole new file in your run. **CI cannot see it either** — there is no GPU runner, so `tests/test_issue776_bnb_nf4_paths.py` can never go green in CI. That is a structural gap, not carelessness on your part. On this box it fails.

**1. The new file does not execute, and its assertion is false even once it does.** Three independent defects:

- `bnb.functional.matmul_4bit` **does not exist** → `AttributeError`. Verified: `hasattr(bnb, "matmul_4bit")` is `True`, `hasattr(bnb.functional, "matmul_4bit")` is `False`. Stock `Linear4bit.forward` calls `bnb.matmul_4bit`.
- One line later, `weights` is `float32`, so `quant_state.dtype` is fp32 and `dequantize_4bit` returns fp32 against an fp16 input → `RuntimeError: expected mat1 and mat2 to have the same dtype`.
- Even with both fixed, `assert max_error > 0.0` is **false at your chosen shape**. At K=N=4096 the two paths are bit-identical at M ∈ {8, 64, 256, 1024, 1536, 1537, 2048}, in fp16 and bf16 alike.

Working shapes measured here: `K=N=64, M=12 → 0.015625` (the shape the bit-exactness test itself runs), or `K=4096, N=14336, M=64 → 0.25`.

**2. The replacement docstring states a measurement that does not reproduce.** `layer_stream_runtime.py:188-189` now says *"measured at K=N=4096, the paths diverge through M=64 and agree at M=2048"* — on this card they are identical at every M I tried at that shape. I asked you to replace an unverified claim with a measured one, and the failure mode worth catching is replacing it with another unverified one.

The divergence is real but **shape-dependent, not simply M-dependent** (sm_120, fp16, max abs err):

| K, N | M=8 | M=12 | M=64 | M=2048 |
|---|---|---|---|---|
| 64, 64 | 0.0156 | 0.0203 | 0 | 0 |
| 512, 512 | 0.0625 | 0.0625 | 0 | 0 |
| 4096, 4096 | 0 | 0 | 0 | 0 |
| 4096, 14336 | 0 | 0 | 0.25 | 0 |

Since you cannot measure this yourself, **use my numbers and attribute them** — something like: *"measured on sm_120 / bnb 0.50.2: the paths disagree at small M on small projections (K=N=64, M=12 → 1.6e-2 fp16, 1.3e-1 bf16) and at M=64 on K=4096,N=14336 → 2.5e-1; whether the fused kernel runs is a per-architecture, per-shape heuristic."*

## Two smaller things

Your new file skips unless bitsandbytes is exactly `0.50.2`, but the dispatch it documents is **per-GPU**: `_gemm_4bit_use_custom_cuda` in `backends/cuda/ops.py` has dedicated blocks for sm75/80/86/89/90/100/120/121. So `max_error > 0.0` may be false on the sm_86 RTX 3050 the original bit-exactness claim was established on — I could not check, there is no second card here. Gate on capability as well as version, or drop the `> 0.0` direction and assert only the upper bound.

And the class docstring at `tests/test_issue385_stream_dtype.py:456` still says *"at small token counts"* with no version and no numbers — the one you and I agreed to replace. One sentence.

(Not yours to fix: `tests/test_v07300.py:15-47` carries a second copy of the same stale reasoning, including `_gemm_4bit_custom_max_m = 1536` and "every M measured from 8 to 2048". After this PR the source and that docstring contradict each other. I will handle it.)

## What #776 still needs — I will update the issue

Your change settles the *forward* comparison on *one* card. Remaining, so nobody reads this merge as closing it: the **card-vs-card discriminator is still unrun** (and now has a much cheaper form — a ten-line `bnb.matmul_4bit` vs dequant+`F.linear` probe at K=N=64, M=12, no model load); **gradients are untested on sm_120**, which matters given #331's history of a bit-exact forward with wrong gradients; and the **root cause is still unestablished** — this explains the test failure, not why bnb's own two paths disagree by up to 0.25 on Blackwell.

It also means CLAUDE.md's claim needs correcting, and that is on me: *"same bytes, same bnb kernels"* — the bytes were the same, the kernels were **not**, until your PR made the test force them to be.

Also worth updating the PR body: it describes only the reference fix, while the diff carries a second deliverable.

`main` is frozen until v0.75.0 is tagged, so nothing merges today regardless. Push those and I will re-run the injection battery on this box.

MakazhanAlpamys:
@Samearth17 — **the merge freeze is lifted**, and I re-measured this at `0f475333` on the Blackwell box (RTX 5070, sm_120, torch 2.14.0+cu130, bnb 0.50.2). Round 1 you answered in substance. Round 2 has had no push, so the five items below still stand — and two of them are blocking in a way that is worth reading, because one would ship a test that **can never pass for anyone**.

## First, the good part, and I want to be precise about it

**The core change is right, and it is not "making the test agree with the code".** I proved the production change is docstring-only rather than eyeballing the hunk — parsed both files, stripped every docstring, hashed the AST:

```
main   AST(docstrings stripped) = 53968d6dd780a288
pr857  AST(docstrings stripped) = 53968d6dd780a288   -> IDENTICAL
```

And the methodology fix is real. The old test varied **two** things at once: the streamed side ran `install_dequant_forward` (dequant + `F.linear`), the resident reference ran stock `Linear4bit.forward` (bnb's fused `matmul_4bit`). Identical bytes, **different kernels**. Applying the patch to the reference too holds the kernel constant, which is what isolates the variable the test claims to measure. `== 0.0` is untouched, no tolerance, no `allclose`, no `xfail`, no widened skip — I checked for weakening specifically and found **zero** removed assertions; all eleven removed lines are prose.

Two mutations, both killed:

| mutation | result |
|---|---|
| remove your added `assert install_dequant_forward(base) > 0` (i.e. test against pre-PR behaviour) | **2 failed** — `0.00048828125 == 0.0` fp16, `0.00390625 == 0.0` bf16 — **#776's exact numbers** |
| neuter `install_dequant_forward` so it patches nothing | **2 failed**, caught by your own new guard |

That second one is a bonus you may not have intended: your `> 0` line turns this into a guard for #331 as well.

**This also corrects something CLAUDE.md gets wrong**, which I will fix on my side: it says streamed NF4 is bit-exact against resident NF4 "same bytes, same bnb kernels". The bytes were the same; **the kernels were not.**

## BLOCKING 1 — `tests/test_issue776_bnb_nf4_paths.py` cannot execute, and cannot pass once fixed

```
pytest tests/test_issue776_bnb_nf4_paths.py -q
  2 failed in 8.87s
  AttributeError: module 'bitsandbytes.functional' has no attribute 'matmul_4bit'
```

Three separate defects, each isolated:

1. `hasattr(bnb, "matmul_4bit")` is **True**; `hasattr(bnb.functional, "matmul_4bit")` is **False**. `Linear4bit.forward` calls `bnb.matmul_4bit(...)`.
2. Fixing only that: your `dtype=torch.float32` weights give `quant_state.dtype = torch.float32` -> `RuntimeError: expected mat1 and mat2 to have the same dtype, but got: struct c10::Half != float`.
3. Fixing both: `assert max_error > 0.0` is **false at your chosen shape**. At K=N=4096 the paths agree exactly at **every** M I probed — 8, 64, 256, 1024, 1536, 1537, 2048 — in **both** fp16 and bf16.

**Your CI is green only because the file skips without a GPU.** Every user who runs the suite on a CUDA box gets two hard failures. That is the "broken and invisible" shape, and it is why this one is blocking rather than a nit.

## BLOCKING 2 — the replacement docstring states a measurement that reproduces as exactly zero

`layer_stream_runtime.py:188-189` still says the paths *"diverge through M=64 and agree at M=2048, measured at K=N=4096"*. Measured here (fp16, max abs error):

| K, N | M=8 | M=12 | M=64 | M=2048 |
|---|---|---|---|---|
| 64, 64 | 0.0078 | **0.0117** | 0 | 0 |
| 512, 512 | 0.0625 | 0.0625 | 0 | 0 |
| **4096, 4096** | **0** | **0** | **0** | **0** |
| 4096, 14336 | 0 | 0 | **0.25** | 0 |

The divergence is **shape**-dependent, not simply M-dependent, and it is zero everywhere at the docstring's own stated shape. Round 1 asked you to replace a stale claim with a measured one; the replacement is a different unreproducible claim, which is the same failure one iteration later.

Working shapes you can attribute to this run: **K=N=64, M=12 -> 1.17e-2 fp16 / 1.25e-1 bf16** (that is the shape the bit-exactness test actually runs), and **K=4096, N=14336, M=64 -> 2.5e-1**.

## Three smaller items, all still open

3. The gate is still `if bnb_version != "0.50.2": skip`. Gate on **capability** as well as version, or drop the `> 0.0` direction — a version string is not evidence that a kernel behaves a particular way.
4. `tests/test_issue385_stream_dtype.py:457` still says *"can use a different fused kernel at small token counts"* — no version, no numbers. Asked in both rounds.
5. The PR body still describes only the reference fix, while the diff carries a second deliberate deliverable.

## A concrete way out, given you have no NVIDIA GPU

I would rather land the good half than have you guess at numbers you cannot measure. **Consider splitting:**

- **Now:** the bit-exactness repair alone — `install_dequant_forward` on the reference plus the `> 0` guard. Two lines, verified good on this box, mutations killed. The docstring reduces to a version-stamped statement that the paths *can* disagree shape-dependently, with **no numbers**.
- **Follow-up under #776:** the probe file and the numeric claims, which I can measure here and hand you.

That gets the real fix in without shipping a test that never runs green. Your call — if you would rather keep it as one PR and use the shapes above, that is fine too.

## What this does not do, and `Refs #776` is right

The user-visible difference is untouched: a real `stream_layers: true` NF4 run still computes via dequant + `F.linear` while a real resident NF4 run uses fused `matmul_4bit`, and those still differ by **4.9e-4 / 3.9e-3** on sm_120 — which is exactly what the first mutation re-measures. This narrows the *claim* to "streaming itself is bit-exact given a matched kernel", which is truer than what was there. Root cause of the bnb-vs-bnb disagreement is still unestablished and needs a pre-Blackwell card I do not have. **#776 stays open**, and your `Refs #776` with an empty closing-issues list is exactly right.

MakazhanAlpamys:
@Samearth17 — status check on both of yours, and one thing on this one that is mine to fix rather than yours.

**#857 (NF4 resident reference compute path).** Five items from my 2026-09-12 review are open, and one of them is the important one: **a test in the new file cannot pass for anyone** — `bnb.functional.matmul_4bit` does not exist, then there is a dtype error, and after fixing both, `assert max_error > 0.0` is false at K=N=4096. Its CI is green only because the file is `skipif`-gated on CUDA and the runners have no GPU, so nothing has ever executed it. I suggested splitting the PR; that offer stands.

**What is mine:** I said twice, on 09-12 at 09:13 and again at 12:48, that *"CLAUDE.md's claim needs correcting, and that is on me."* It is still not done — `.claude/CLAUDE.md` still states that streamed NF4 is bit-exact against resident NF4, and I have measurements on an RTX 5070 showing it is not on sm_120 (`4.9e-4` / `3.9e-3`, while the `quantization: none` variants pass). I also said I would handle the second copy of the same stale reasoning in `tests/test_v07300.py:37-38` and have not. Both are on my list and neither is a reason for this PR to wait.

**#883 (modern CUDA driver versions).** One item: the `whl/None` call site. Worth restating why it matters — all 52 of your tests pass **with the bug live**, because the guard is on the driver rather than on the tag, so for any driver below CUDA 11.8 the URL becomes `https://download.pytorch.org/whl/None` and returns 403. A green suite over a live bug is the exact shape this repo keeps having to catch.

Note #883 is in a crowded file: **six open PRs** touch `commands/doctor.py` right now, and #884 is about to merge. You will want a rebase check.

Both of these are real findings from you and I want them landed. Still have time? If either has gone cold, say so and I will record it rather than let it age.

MakazhanAlpamys:
@Samearth17 — the half that was mine is done, so you are not waiting on me for it any more.

**`.claude/CLAUDE.md` is corrected at both sites.** The bit-exactness claim now reads as hardware-scoped rather than absolute, with the measurement attached:

> `[nf4]` fails at **4.88e-04** in float16 and **3.91e-03** in bfloat16 on the RTX 5070, while `[none]` **passes** in both — so the divergence is in the bitsandbytes NF4 path on that card, not in streaming. Root cause not established; filed as **#776**.

I also recorded the thing that made this take a week to notice: **CI cannot see that test at all.** The module is `skipif(not cuda.is_available())` and GitHub runners have no GPU, so the "CI test, not a one-off" claim was itself wrong — the guard runs only on a dev box with a card. And I added the instruction not to relax the assertion to a tolerance if someone hits it again, because it is the only thing that would catch a genuine streaming bug.

I said this was on me on 2026-09-12 at 09:13 and again at 12:48 and then did not do it for a day. Thank you for the finding; the delay was mine.

**`tests/test_v07300.py:37-38` still carries the same stale reasoning** (`_gemm_4bit_custom_max_m = 1536`) and is still mine to handle. I have not done that one yet and am not going to claim otherwise.

Your five items on this PR stand as written, and the split I suggested is still the offer.

MakazhanAlpamys:
Reviewed the 14:11 push at `0a89fa75`. **This is the right fix, and it hands us the root cause of #776 — I have written that up on the issue.**

**First, the thing I came in to check: you did not loosen the test.** The assertions are still exact equality:

```
527:  assert self._run(tmp_path, "float16", quant) == 0.0
538:  assert self._run(tmp_path, "bfloat16", quant) == 0.0
```

No tolerance, no `approx`, no skip, no xfail. That mattered because a bit-exactness assertion relaxed to a tolerance stops detecting the only class of bug it exists for, and two failing arms turning green is exactly the shape I would expect a weakening to take.

**The reference correction is legitimate, and I verified the mechanism rather than accepting the reasoning.** `layer_stream_runtime.py:875` applies `install_dequant_forward` to every streamed NF4 layer in production — #331's repair, keeping the pooled buffer out of `MatMul4Bit`'s non-`save_for_backward` capture. So the old test compared a streamed model on the **dequant** path against a resident reference on the **fused** path. Two different kernels.

**Then I isolated it with streaming removed entirely** — two resident NF4 models, byte-identical weights, differing only in whether `install_dequant_forward` was applied:

```
float16 : resident-native vs resident-dequant maxabs = 0.0006103515625
bfloat16: resident-native vs resident-dequant maxabs = 0.00439453125
```

Same magnitude as #776's failures, with no streaming in the experiment at all. That is the attribution, and it is why `[none]` passed in both dtypes while `[nf4]` failed in both — the `[none]` arm has no such kernel split.

Results on this box (RTX 5070, sm_120, torch 2.14.0+cu130, bnb 0.50.2):

```
before: [none] PASS PASS | [nf4] FAIL 0.00048828125 / FAIL 0.00390625
after:  4 passed   (full file: 26 passed, 1 skipped)
```

Mutation: `assert install_dequant_forward(base) > 0` -> `assert install_dequant_forward is not None` (anchor counted once, real import confirming the mutant line is live) puts both `[nf4]` arms back at exactly those values while both `[none]` arms keep passing. Behavioural, and it discriminates precisely.

I also confirmed the `layer_stream_runtime.py` edit is **docstring-only** by parsing both revisions and comparing ASTs with docstrings stripped — identical. So no production behaviour changes on any card, and on a card where the two kernels agree (the RTX 3050 the original claim came from) applying the patch to the reference is a numeric no-op. The change strictly narrows what the test measures to the streaming mechanism.

---

**One thing I want back, and it is the one you already wrote.**

You added `tests/test_issue776_bnb_nf4_paths.py` in `237057c2` and removed it in `b3678971` ("remove invalid NF4 path probe"). **After this PR the fused-vs-dequant divergence is documented only in two docstrings and asserted nowhere** — and that property is the entire justification for turning two failing assertions green. A future bitsandbytes release that fixes or worsens the fused path would turn nothing red, and the next person to read the docstring has no way to check whether it is still true.

The shape you had was right: version-pinned to bnb 0.50.2, skipped otherwise. If the reason you pulled it was that it was measuring the wrong thing, tell me what was invalid about it and I will help get it right — but I would rather have a pinned probe with a narrow skip than prose.

**Two smaller notes:**

- **No changelog fragment.** Defensible for a test-only change, except that the production docstring retracts a published numerics claim, and the fragment is where the #776 attribution would be recorded for anyone reading the release rather than the issue.
- **CI cannot see any of this.** The class is `skipif(not _cuda_available())` and GH runners have no GPU, so all four parametrizations skip on all nine cells. Green CI on this PR proves nothing — this box is the only signal, which is why I ran it here and wrote the numbers down.

Merge-tree against current `main` is clean despite your base being 39 commits behind.

MakazhanAlpamys:
Merged as `d456a1d3`. Thank you — and the reason I merged this is the thing you did **not** do.

Two failing bit-exactness arms turning green is exactly the shape a weakening takes, and I came into this review expecting to find a tolerance. The assertions are still `== 0.0`: no `approx`, no skip, no xfail. You moved the *reference* onto the same compute path instead, which strictly narrows what the test measures rather than loosening what it accepts. That distinction is the whole review.

I verified the mechanism rather than the reasoning: `layer_stream_runtime.py:875` applies `install_dequant_forward` to every streamed NF4 layer in production, so the old test really was comparing `dequantize_4bit + F.linear` against fused `MatMul4Bit` — two kernels with no reason to agree. And the `layer_stream_runtime.py` edit is docstring-only, confirmed by parsing both revisions and comparing ASTs with docstrings stripped, so no card's behaviour changes.

**Your isolation is what closed #776, and it deserves saying plainly.** Two resident NF4 models, byte-identical weights, **no streaming anywhere in the experiment**, differing only in whether `install_dequant_forward` was applied:

```
float16 : 0.0006103515625
bfloat16: 0.00439453125
```

Same magnitude as the failures. So streaming was never the variable, and a claim this project had carried since the RTX 3050 measurements is now correctly scoped to hardware. That is a better outcome than the test fix.

## The one thing I asked for and did not get, and what I did about it

You wrote `tests/test_issue776_bnb_nf4_paths.py` in `237057c2` and removed it in `b3678971`. I asked what was invalid about it and merged before hearing back, because leaving two arms failing on the only box that can run them is active noise that would mask a real regression.

But the property those two green assertions now rest on is asserted **nowhere** — it survives in two docstrings. A bitsandbytes release that fixed or worsened the fused path would turn nothing red.

So I have filed **#968** rather than let it disappear, and it credits your deleted probe as the starting point: version-pinned to bnb 0.50.2, skipped otherwise, which was the right shape. **The first question on it is still yours to answer** — if the probe was genuinely measuring the wrong thing, that reason is itself worth recording, and I would rather have it from you than guess.

Two smaller things, for the record rather than as asks: there is no changelog fragment, which is defensible for a test-only change except that a production docstring here retracts a published numerics claim; and CI cannot see any of this, since the class is `skipif(not _cuda_available())` and GitHub runners have no GPU, so all four parametrizations skipped in all nine cells. Green CI on this PR proved nothing — this box was the only signal, which is why the numbers above are written down.

Your `CONTRIBUTORS.md` entry is updated at `a40d2fc3`.

Issue #813, opened 2026-09-11
Gate commands use five different exit-code conventions: a missing spec file exits 2 (= MAJOR) in eval checklist, a regression exits 1 (= typo) in eval gate


MakazhanAlpamys:
## The defect

Soup has a documented exit-code contract for its flagship gate. `docs/evaluation.md:379`: "**0 = SHIP · 2 = DON'T SHIP · 3 = usage/flag error · 1 = runtime error** — usage errors moved off `2` so CI can tell a config typo from a caught regression." The other gate commands do not follow it, and do not follow each other. A CI job has to know, per command, which number means "the model failed" and which means "you mistyped a path".

## Evidence — observed, not read off the source

Every row below was run from a scratch directory with an isolated `SOUP_DB_PATH`. "pass" / "fail" are the command's own verdicts.

| command | pass | gate fail | usage / input error | notes |

|---|---|---|---|---|

| `soup ship --evidence` | 0 | 2 | 3 (`--forgetting-threshold 2`) | missing evidence file: 1 |

| `soup eval gate` | 0 | **1** | **1** (missing suite; `--regression-threshold 2`) | fail == typo |

| `soup eval against` | 0 | **1** | **1** (unknown run; `--metric accuracy`) | fail == typo == no data |

| `soup eval behavior` | 0 | 2 | **2** (missing `--evidence`; `--battery nope`) | fail == typo |

| `soup eval checklist` | 0 | 2 | **2** (missing spec file) | fail == typo |

| `soup eval quant-check` | 0 | **0** (MAJOR) | 1 (missing `--tasks`; `--format xml`) | a failure cannot fail |

| `soup lock check` | 0 | **3** (drift) | **2** (missing lock file) | 2 and 3 swapped vs ship |

| `soup lock show` | 0 | n/a | **2** (missing file) | |

| `soup data lint` | 0 | 2 | **2** (non-preference format), 1 (missing file) | fail == wrong format |

| `soup data validate` | 0 | **0** (0/5 valid) | 1 (missing file) | a failure cannot fail |

Selected raw lines:

```

eval gate: missing suite        exit=1 | Cannot load suite: eval-gate suite not found: nope.yaml

against: regressed              exit=1 | REGRESSED task_accuracy delta_mean=-0.6000 ci=[-0.6000, -0.6000]

against: bad metric             exit=1 | Empty series — baseline=0 candidate=0.

checklist: MAJOR                exit=2 | Overall: MAJOR

[...]

PR #858, opened 2026-09-11
fix(data): fail validation when no rows are usable

### LINKED ISSUE #811: soup data validate exits 0 at "0/5 rows valid", so the `soup ci init` gate passes a dataset the loader reduces to zero rows
## The defect

`soup data validate` reports how many rows are valid and then always exits 0. `commands/data.py:105-116`:

```python

result = validate_and_stats(data, expected_format=fmt)

if result["issues"]:

    console.print("[yellow]Issues found:[/]")

    ...

valid = result["valid_rows"]

total = result["total"]

console.print(f"\n[green]{valid}/{total} rows valid for {fmt} format[/]")

```

There is no exit after the count. The only non-zero exits are a missing file (line 90) and an undetectable format (line 103).

`soup ci init` writes that command into the generated PR gate as its first step (`utils/ci_workflow.py:69-70`):

```yaml

      - name: Validate training data

        run: soup data validate {data}

```

So the step named "Validate training data" is green for a file with no usable rows.

This was raised and deliberately deferred during #695: the validator now runs the real converters (#712), so its row count finally agrees with `load_dataset` — but the exit code was left alone to keep that PR a reporting fix. The review said: "Today `validate` exits 0 even at `0/N`. Whether that should change is a real question — my call is do not change it in this PR ... Raise it separately if you think it should." This is that issue.

## Evidence

A file of five rows that auto-detect as alpaca, none of which convert:

```

$ soup data validate bad.jsonl

Auto-detected format: alpaca

Issues found:

  ! 5 rows fail to convert for 'alpaca' format (load_dataset would drop them)

  ! row 0: alpaca.instruction must be a string, got NoneType

  ! row 1: alpaca.instruction must be a string, got NoneType

  ! row 2: alpaca.instruction must be a string, got NoneType

  ! ... and 2 more

  ! 4 duplicate rows found

  ! 10 empty fields found

  ! 5 samples are very short (<10 chars)



0/5 rows valid for alpaca format

exit=0

```

The count is correct; the loader agrees and keeps nothing:

```

>>> out = load_dataset(DataConfig(train='bad.jsonl', format='alpaca'))

>>> {k: len(v) for k, v in out.items()}

{'train': 0, 'val': 0}

```

And the generated gate runs exactly that command:

```

$ soup ci init --data bad.jsonl --suite suite.yaml --evidence ev.json

Wrote .github/workflows/soup-gate.yml

$ grep -A1 "Validate training data" .github/workflows/soup-gate.yml

      - name: Validate training data

        run: soup data validate bad.jsonl

```

## Reproduction

```bash

python -c "import json; open('bad.jsonl','w').write(''.join(json.dumps({'instruction': None, 'output': None})+'\n' for _ in range(5)))"

soup data validate bad.jsonl; echo "exit=$?"

```

## Why it matters

The validator used to disagree with the loader, which is why a pass meant little. After #712 it agrees — the output literally says "load_dataset would drop them" — so the tool now knows the file is unusable and still reports success to CI. The failure moves from the pull request, where it is cheap, to the training job, where it costs a GPU allocation.

## Proposed fix

1. Exit 2 when `valid_rows == 0 and total > 0`. That case is unambiguous: no training run can use the file.

2. Keep partial drops at exit 0 by default. `tests/test_issue676_non_dict_messages.py:156-157` pins `1/2 rows valid` at exit 0, and some drop rate is normal in real data.

3. Add an opt-in threshold for CI, e.g. `--min-valid-fraction F` (exit 2 below it) or `--strict` (exit 2 on any dropped row), and consider having `soup ci init` emit the threshold so the generated gate is meaningful.

4. Document the exit codes in `soup data validate --help` and in `docs/data.md`.

## Acceptance criteria

- [ ] `soup data validate` on a file where every row fails to convert exits 2 and names the reason. A test for this FAILS on current code (exit 0).

- [ ] A file with a partial drop still exits 0 by default (the existing #676 test keeps passing).

- [ ] The opt-in threshold exits 2 when the valid fraction is below it, and 0 at or above it.

- [ ] A missing file and an undetectable format keep their current exit 1.

---

akkupratap323:
﻿## What does this PR do?

`soup data validate` reported `0/N rows valid` but returned success, allowing generated CI gates to accept datasets that the loader would reduce to zero rows. This change returns exit code 2 when a non-empty dataset has no usable rows and adds `--min-valid-fraction` for CI policies that need a stricter threshold. Partial drops still succeed by default, and input/format errors keep exit code 1.

The CLI help and data guide document the exit codes and threshold behavior. Regression tests cover zero usable rows, partial drops, below/equal threshold behavior, and both exit-1 cases.

Closes #811

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

## Validation

- `pytest -o addopts='' tests/test_issue811_data_validate_exit_codes.py tests/test_issue676_non_dict_messages.py tests/test_data_tools.py tests/test_bugfixes.py::TestValidateAutoDetect -q` (53 passed)

The full local suite requires the optional training dependency stack; the focused CLI/data suites pass with the core development dependencies installed.

### DIFF
--- a/changelog.d/0.74.0/858.fixed.md
+++ b/changelog.d/0.74.0/858.fixed.md
@@ -1,0 +1,2 @@
+- Make `soup data validate` fail when no rows are usable, and add an optional
+  minimum-valid-fraction threshold for stricter CI gates (#811 by @akkupratap323 in #858).
--- a/docs/data.md
+++ b/docs/data.md
@@ -737,6 +737,9 @@ soup data inspect ./data/train.jsonl
 soup data validate ./data/train.jsonl
 soup data validate ./data/train.jsonl --format alpaca
 
+# Require at least 90% of rows to be usable
+soup data validate ./data/train.jsonl --min-valid-fraction 0.9
+
 # Convert between formats
 soup data convert ./data/train.jsonl --to sharegpt --output converted.jsonl
 
@@ -754,6 +757,12 @@ soup data filter ./data/train.jsonl --coherence 0.3
 soup data filter ./data/train.jsonl --perplexity 500 --coherence 0.3
 soup data filter ./data/train.jsonl --score-only  # add scores without filtering
 ```
+
+`soup data validate` exits with code `0` when at least one row is usable and the
+optional minimum valid fraction is met. It exits with code `1` for input errors,
+such as a missing file or an undetectable format, and code `2` when a non-empty
+dataset has no usable rows or falls below `--min-valid-fraction`. A partially valid
+dataset still exits with code `0` when no minimum is specified.
 
 
 ## Demo Datasets (`soup data demo`)
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -82,8 +82,15 @@ def validate(
         "auto", "--format", "-f",
         help="Expected format: auto, alpaca, sharegpt, chatml, dpo, kto, plaintext",
     ),
+    min_valid_fraction: float = typer.Option(
+        0.0,
+        "--min-valid-fraction",
+        min=0.0,
+        max=1.0,
+        help="Exit 2 when the fraction of valid rows is below this value",
+    ),
 ):
-    """Validate dataset format and report issues."""
+    """Validate a dataset, returning exit 1 for input errors and 2 for unusable data."""
     file_path = Path(path)
     if not file_path.exists():
         console.print(f"[red]File not found: {file_path}[/]")
@@ -114,6 +121,17 @@ def validate(
     valid = result["valid_rows"]
     total = result["total"]
     console.print(f"\n[green]{valid}/{total} rows valid for {fmt} format[/]")
+
+    if total > 0 and valid == 0:
+        console.print("[red]Validation failed: no usable rows remain.[/]")
+        raise typer.Exit(2)
+
+    if total > 0 and valid / total < min_valid_fraction:
+        console.print(
+            f"[red]Validation failed: valid fraction {valid / total:.3f} is below "
+            f"--min-valid-fraction {min_valid_fraction:.3f}.[/]"
+        )
+        raise typer.Exit(2)
 
 
 @app.command()
--- a/tests/test_issue811_data_validate_exit_codes.py
+++ b/tests/test_issue811_data_validate_exit_codes.py
@@ -1,0 +1,97 @@
+"""Regression tests for ``soup data validate`` exit codes (#811)."""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import pytest
+from click.testing import Result
+from typer.testing import CliRunner
+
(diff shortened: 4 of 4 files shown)


akkupratap323:
@MakazhanAlpamys The #811 fix is ready, including the required #858 changelog fragment. The focused CLI/data suite passes (53 tests), and the full Ruff check passes. Could you please review and assign this PR if appropriate? Thank you.

akkupratap323:
@MakazhanAlpamys The #811 fix is ready, including the required #858 changelog fragment. The focused suite passes (53 tests), and the full Ruff check passes. Could you please review and assign this PR if appropriate? Thank you.

MakazhanAlpamys:
Merged as d1d2439f — thank you, @akkupratap323, and welcome. You are now in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md).

What made this easy to merge:

- **The scope is exactly the issue.** Only zero usable rows fails. A partial drop still passes, so the #676 behaviour is intact.
- **No new exit-code convention.** It follows the pass 0 / failed gate 2 / input error 1 shape `soup ship` and `soup data lint` already use — #813 is about how many conventions there already are.
- **Every test kills its mutation.** The threshold test pins the equal case, and you used the shared `strip_ansi` helper.

What I ran before merging:

| input | `main` | this PR |
|---|---|---|
| every row dropped (0/5) | 0 | **2** |
| healthy (5/5) | 0 | 0 |
| 3/5 valid | 0 | 0 |
| 3/5 with `--min-valid-fraction 0.61` | n/a | 2 |
| missing file / empty file | 1 | 1 |

- **Mutations, all killed**: removing the zero-rows exit; changing it from 2 to 1; over-broadening to any dropped row (4 failures, including the #676 test); `<` changed to `<=` on the threshold.
- 66 passed across the validator, fragment and ANSI suites, and 100 across the other `data validate` callers. The test file also passes under `FORCE_COLOR=1`.
- All 13 required checks green on `382bd16e`.

Small things for a follow-up if you want them, none worth holding this for:

- An out-of-range `--min-valid-fraction` is a Click usage error, which exits **2** — the same code as a failed gate — while `docs/data.md` says input errors exit 1.
- An empty file with an explicit format still reports "0/0 rows valid" and exits 0, which the issue scoped out.
- While checking this I found `--format bogus` reports every row valid and exits 0. It is pre-existing, and now filed as #866 — yours if you want it.

One process note: #811 had no claim comment when this opened. [CONTRIBUTING.md's "Claiming an Issue"](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTING.md#claiming-an-issue) asks for one line on the issue first, so two people do not build the same fix. That happened to someone else's PR today.

Issue #204, opened 2026-05-20
v0.63.1: Live SaaS pull adapters for soup ingest (Langfuse/LangSmith/Helicone/OpenPipe API SDKs)


MakazhanAlpamys:
## Context

v0.63.0 `soup ingest` parses JSONL exports the operator pulls from their SaaS dashboard or vendor SDK. Soup itself never makes the network call, which keeps the credential-handling threat surface zero, but operators currently have to manually export → save → invoke `soup ingest`.

## Proposed fix path

Add an optional `--pull` flag that, when combined with `--source <name>`, lazy-imports the source's official Python SDK and pulls the last N days of traces directly into a tempfile that the existing parser consumes.

Adapter sketch:
- `langfuse` → `langfuse-python` (LF SDK)
- `langsmith` → `langsmith` package
- `helicone` → REST API via `httpx` (no official SDK)
- `openpipe` → `openpipe` SDK
- `openai-stored` → `openai>=1.0` `/responses` API
- `otel` → out of scope (operators already export from their OTel collector)

Authentication via the existing env vars (`LANGFUSE_KEY` / `LANGSMITH_API_KEY` / `HELICONE_API_KEY` / `OPENPIPE_API_KEY` / `OPENAI_API_KEY`) so no new credential-handling surface.

## Acceptance criteria

- [ ] `soup ingest --source langfuse --pull --since 7d --output traces.jsonl` works end-to-end with a valid `LANGFUSE_KEY`.
- [ ] Per-source lazy import — `--source openpipe` does NOT import `langfuse-python` and vice versa.
- [ ] Missing-SDK error surfaces with a friendly `pip install <package>` advisory.
- [ ] HTTPS-only, per-call timeout, max-pages cap, rate-limit backoff.
- [ ] Default behaviour (no `--pull` flag) unchanged: parses the local export, never makes the network call.
- [ ] Tests parametrize over all 5 SDKs with mocked HTTP responses.
- [ ] Documented in README.md under the Production Trace Ecosystem section.

## Related

Shipped in v0.63.0 — see [v0.63.0 release notes](https://github.com/MakazhanAlpamys/Soup/releases/tag/v0.63.0).

MakazhanAlpamys:
Deferred from the v0.71.5 patch and labelled `infra-blocked` (kept **open**).

[...]

PR #859, opened 2026-09-11
feat(ingest): pull Langfuse generations live with --pull (#204)

Konuktor:
Refs #204 — the Langfuse slice. LangSmith, Helicone, OpenPipe and OpenAI Stored Completions stay open on the issue.

## What this adds

`soup ingest --source langfuse --pull` fetches generations from the Langfuse API instead of reading an export:

```bash
export LANGFUSE_PUBLIC_KEY=pk-lf-...  LANGFUSE_SECRET_KEY=sk-lf-...
export LANGFUSE_HOST=https://us.cloud.langfuse.com   # optional; default https://cloud.langfuse.com
soup ingest --source langfuse --pull --since 7d --output traces.jsonl
```

New flags, all `--pull`-only: `--since` (`30m` / `24h` / `7d`, max `365d`, default `7d`), `--max-pages` (default 100, max 10 000), `--allow-private-host`. `--pull` and `--logs` are mutually exclusive; `--pull` with any other source exits 2 and says to export and use `--logs`.

## Decisions asked for in the thread

**Plain HTTPS, not the `langfuse` SDK.** The pull is one read-only `GET /api/public/v2/observations` with cursor pagination. langfuse-python 4.15.2 would add OpenTelemetry (api, sdk, exporter), wrapt, backoff and httpx to the install, and would own retries, timeouts, redirects and exception formatting — the parts this PR is reviewed on. So it uses `urllib.request` from the standard library: no new dependency, therefore no optional extra. The cost is that Soup owns the wire format; the recorded real response below is what pins it.

**Observations API v2.** langfuse-python 4.15.2 marks `api.trace.list` deprecated: *"On Langfuse Cloud … this endpoint will be removed on November 16, 2026."*

**One output row per `GENERATION` observation.** A generation is the unit with a model, the exact input it was given and the output it produced — what a fine-tune learns from. Reassembling a trace needs a rule for which observation's input is "the prompt", and an agent trace's root span holds app-level arguments rather than a conversation. Consequences, all verified live:
- an agent trace yields one row per LLM call; its agent span and tool calls yield none;
- a row's `trace_id` is the observation id (`parse_langfuse` keeps no other id);
- a generation with no input or no output is pulled, dropped by the parser, and counted in the summary (`N generation(s) had no input or no output and were skipped`), so a pull that finds nothing usable says so.

**Input/output shaping.** Recorded against the real API: plain text comes back as-is, but structured values come back as JSON inside a string (`"[{\"role\": \"system\", ...}, {\"role\": \"user\", ...}]"`). The adapter decodes those and hands a message list to `parse_langfuse` as `{"messages": [...]}`, the shape it already joins, so the prompt keeps the whole input context, system instructions included. Given a bare list, `_coerce_str` keeps only the first message — for a chat generation, the system prompt; given the undecoded string, the prompt is JSON text. Both are pinned as controls against the recorded page. `parse_langfuse` itself is unchanged, so local-export output is byte-identical.

## Your review points

| Point | Implementation | Pinned by |
|---|---|---|
| Langfuse's own env names; advisory and docs in the same PR | `_AUTH_ENV["langfuse"] = ("LANGFUSE_PUBLIC_KEY", "LANGFUSE_SECRET_KEY")`; host from `LANGFUSE_BASE_URL`, then `LANGFUSE_HOST` (the SDK's precedence); `docs/data.md` | PII panel names the pair and not `LANGFUSE_KEY`; `resolve_auth_env` needs both halves; the two `test_v0630_part_a.py` pins updated |
| Without `--pull`: byte-identical, no SDK / httpx / network | `utils/ingest_pull.py` is imported only inside the `--pull` branch | output bytes against a golden captured before this change; a clean subprocess with sockets disabled reports nothing loaded and no connection attempt, plus a control proving the probe sees the module when it *is* imported |
| Credentials never leave memory | environment only (never a flag, so never in the audit log's argv); `Authorization` built inline so no local variable holds it; redacted reprs; server text scrubbed before truncation | a planted key pair through the real `run()` entry point with `--verbose --log-level debug` in 7 scenarios (success, 401, server echoing the request, page cap, 429 exhaustion, network error, unexpected exception), scanning stdout, stderr, log records, every file in cwd, the audit log and the webhook payload — raw, ANSI-stripped and whitespace-compacted, whole token and 20-char prefix — with controls that each surface was produced; plus a rich traceback rendered with `show_locals=True` over the transport frame |
| `LANGFUSE_HOST` validated, reusing `validate_webhook_url` | `https://` required, then `validate_webhook_url(raw, allow_private_hosts=<--allow-private-host>)`, then no userinfo / query / fragment; loopback also needs the flag (the webhook validator exempts it) | spy on `validate_webhook_url`; refused hosts including `https://169.254.169.254`, `https://127.1` and `https://[::1]:3000`; private and loopback HTTPS refused with the hint and admitted by the flag; plain HTTP refused even with the flag; embedded credentials refused without being echoed |
| Bounded | 30 s per request; 64 MiB per response; `--max-pages` and the 1M-row cap raise instead of truncating (exit 1, "Nothing was written"); output streams through a new `paths.atomic_write_lines`, so an earlier file at `--output` is untouched; 429 retried up to 5 times honouring `Retry-After` (seconds or HTTP-date) with a 60 s ceiling; redirects refused rather than followed with credentials attached; an error body that cannot be read keeps its status | loopback-server tests of the real transport (Basic header, 302 not followed, timeout, size cap, unreadable error body); pagination, cap and backoff tests; the live page-cap stop below |
| A recorded v2 response through the adapter **and** `parse_langfuse` | `tests/fixtures/issue204_langfuse_observations_v2_page.json` | the four final `TraceRecord`s asserted exactly (including the agent trace's two rows), the skipped no-output row, and the two shaping controls |

## Live E2E — Langfuse Cloud Hobby, US region

Seeded with langfuse-python 4.15.2 in a throwaway venv, deleted afterwards: one generation at 11:43:48Z meant to fall outside the window, then at 11:49:46–55Z a chat generation (`[system, user]`), a plain-string generation, an agent trace (agent span → plan generation → tool → answer generation) and a generation with no output. Everything was visible in the API within ~25 s. Every command ran through a wrapper that scrubbed and counted key material in its output: 0 hits.

| Run | Generations pulled | Traces | Exit |
|---|---|---|---|
| `soup ingest --source langfuse --pull --since 5m` | 5 | 4 written (1 without output skipped) | 0 |
| `soup ingest --source langfuse --pull --since 15m` | 6 | 5 written (adds the outside-window generation) | 0 |
| same CLI, wrong `LANGFUSE_SECRET_KEY` | — | nothing written | 1: "Langfuse rejected the key pair (HTTP 401); check LANGFUSE_PUBLIC_KEY and LANGFUSE_SECRET_KEY, and that LANGFUSE_HOST is the project's region" |
| pull module, US keys with `LANGFUSE_HOST` unset (EU default) | — | — | the same HTTP 401 message |
| pull module, page size forced to 2 | 6 over 3 requests (2 carrying a cursor) | parses to 5 | same ids and order as one page |
| pull module, page size 2, `max_pages=2` | 4 | — | `PullLimitError`: "stopped after 2 page(s) and 4 generation(s) with more results pending; raise --max-pages (up to 10000) or narrow --since" |

22/22 scripted checks against the seeded observation ids: exact prompt / output / model per row; plan and answer are separate rows sharing one Langfuse trace; the agent span and the tool call produced no row; the generation without output (`"output": null` in the API) was not written; written order equals the API's order (newest `startTime` first); no key material in any artefact of the run, including the audit log. The wrong-key run was scanned for the real pair, the wrong key and both Basic tokens (whole and 20-char prefix) across stdout, stderr and the audit log: none found.

Re-run on the committed head against the same project: `--pull --since 2h` pulled 6 generations and wrote 5 traces; a wrong secret exited 1 with the same 401 message and wrote nothing; no key material in the console, the output or the audit log.

```
$ soup ingest --source langfuse --pull --since 5m --output since-5m.jsonl
╭──────────────────────────────── PII reminder ────────────────────────────────╮
│ Traces may contain sensitive user data (PII).                                │
│ Review the output before sharing or uploading to external systems.           │
│ Auth env var for this source: LANGFUSE_PUBLIC_KEY + LANGFUSE_SECRET_KEY      │
╰──────────────────────────────────────────────────────────────────────────────╯
Pulling GENERATION observations from https://us.cloud.langfuse.com for the last
5m...
Wrote 4 traces from langfuse (5 generations pulled) -> since-5m.jsonl
1 generation(s) had no input or no output and were skipped.
```

`since-5m.jsonl` — synthetic seed content, nothing to redact:

```jsonl
{"trace_id": "44d5516f74ed7738", "prompt": "How tall is the Eiffel Tower?\n330 m", "output": "The Eiffel Tower is 330 m tall.", "source": "langfuse", "signal": "none", "metadata": {"model": "gpt-4o-mini"}}
{"trace_id": "09c25553e670e9a1", "prompt": "How tall is the Eiffel Tower? Decide whether to search.", "output": "search: Eiffel Tower height", "source": "langfuse", "signal": "none", "metadata": {"model": "gpt-4o-mini"}}
{"trace_id": "e83a0c5fd3d47039", "prompt": "Translate 'hello' to Spanish.", "output": "hola", "source": "langfuse", "signal": "none", "metadata": {"model": "gpt-4o-mini"}}
{"trace_id": "8dd24f84d2bf7735", "prompt": "You are terse.\nWhat is the capital of France?", "output": "Paris.", "source": "langfuse", "signal": "none", "metadata": {"model": "gpt-4o-mini"}}
```

## What the real API does that its docs don't say

- `input` / `output` are documented as "always returned as raw strings". In practice plain text is returned as-is and only structured values are JSON-encoded. I had first assumed a quoted JSON literal for plain text too; the recorded page disproved it and that branch was removed.
- `fields=core,io,model` also returns undocumented `modelId` (same value as `internalModelId`), `inputPrice`, `outputPrice`, `totalPrice`, `latency` and `timeToFirstToken`. The fixture sanitiser caught `modelId` and refused to write the file until it was redacted as well.
- A last page's `meta` is `{}` — no `cursor` key at all.
- Default order is `startTime` descending, and `core` does not include `name`.

## The recorded fixture

One real page, fetched with exactly the query `--pull` sends. Sanitised: `projectId`, `internalModelId` and `modelId` replaced with placeholders; checked to contain no key or key prefix, Basic token, `Authorization`, project or model id, hostname, username, home path or email address. Kept: the random span / trace ids (they carry the agent trace's structure), timestamps, public model prices and the synthetic seed text.

## Tests

- `tests/test_issue204_langfuse_pull.py`, 112 tests. Written first and run against unmodified `main`: everything failed except the two by-design guards (golden bytes, no-import probe). Four tests that initially passed there only because Click rejected the unknown flag were tightened to assert Soup's own message. Tests added later — the streaming writer, the recorded page, the unreadable error body — came with the change that needed them.
- Mutation check against the final file: 27 deliberate breakages, all killed — secret in the 401 message; unscrubbed server text; default dataclass repr; `Authorization` held in a local; redirects followed; `io` dropped from `fields`; pull module imported on the local path; loopback gate removed; https-only check removed; shared SSRF validator bypassed; `LANGFUSE_HOST` read before `LANGFUSE_BASE_URL`; `Retry-After` ceiling removed; `Retry-After` ignored; 429 not retried; page cap disabled; row cap removed; window recomputed per page; response size cap removed; unreadable error body raising; a debug line logging the header; `--pull` with `--logs` accepted; non-atomic write; zero-row warning removed; local-export bytes changed; advisory back to `LANGFUSE_KEY`; message list passed bare; structured io not decoded. The `Authorization`-in-a-local mutation initially **survived**: rich truncates locals at 80 characters, which still prints most of the Basic token while dodging an exact match. The scan now renders untruncated and also matches 20-character prefixes.

```
ruff check src/soup_cli/ scripts/ tests/ benchmarks/       All checks passed!
pytest tests/test_issue204_langfuse_pull.py tests/test_v0630_part_a.py tests/test_v0630_followups.py
                                                            167 passed
pytest tests/ -q --tb=short   main 124890d                  20922 passed, 107 skipped, 0 failed   coverage 83.22%
pytest tests/ -q --tb=short   this branch                   21034 passed, 107 skipped, 0 failed   coverage 83.32%
```

Compared test-by-test against `main`: no existing test missing or changed outcome, identical skip set, +112 passed (all from the new file). mypy on the touched files reports no new errors. Two existing tests were edited: `test_v0630_part_a.py`'s `resolve_auth_env` pins now set the key pair instead of `LANGFUSE_KEY`.

## Limitations and not in scope

- LangSmith, Helicone, OpenPipe and OpenAI Stored Completions have no live pull — hence `Refs`, not `Closes`.
- Pulled rows carry `signal: "none"`. Langfuse scores are separate objects (and Scores v2 shares the 2026-11-16 sunset), so mapping thumbs needs a decision about which score name means feedback.
- 429 handling is covered offline only; the live run stayed well under the Hobby limit of 30 req/min.
- A plain-text input that is itself a JSON object or array cannot be told apart from a structured one in the API response and is decoded as structure.
- `--since` was verified with real elapsed time, because the v4 SDK cannot backdate an observation's start time.
- Noticed, not changed: Soup's Typer app keeps Typer's default `pretty_exceptions_show_locals=True`. `run()` catches exceptions first, so the `soup` entry point never reaches it, and this PR's transport does not rely on that. It affects every command, so it seems like its own issue.

## Changelog

`changelog.d/0.74.0/859.added.md`


MakazhanAlpamys:
Merged as ffb74e58 — thank you, @Konuktor. Your CONTRIBUTORS.md entry has a new line. `#204` stays open for the other four providers.

This is the first time `soup ingest` touches the network, and you built the security in from the start rather than reviewing it in afterwards:

- **Redirects are refused, not trusted.** urllib's default copies `Authorization` onto the redirected host. With redirects refused, a 302 to `http://169.254.169.254` produces exactly one request and a loud refusal, which I confirmed through the real transport.
- **Secrets are scrubbed before truncation**, so no cut can expose half a key.
- **Every leak check has a control proving the surface it scans was actually produced.** That is what makes a "no leak found" result mean something.
- **The recorded v2 page.** It runs through the adapter and then the existing parser with the records asserted. Recording it found three places where the API differs from its docs — exactly why I asked for a recording rather than a hand-shaped mock.
- **Standard-library urllib, staged atomic output, and a byte-identical default path.**

What I ran before merging:

- **Hosts**: 34 hostile values, all refused by default — `http://` (even with `--allow-private-host`), loopback in every spelling including `2130706433` / `0x7f000001` / `127.1` / `0177.0.0.1`, `[::1]`, `[::ffff:127.0.0.1]`, link-local, private ranges, userinfo, and query or fragment.
- **Secret leakage**: planted keys, `--verbose --log-level debug` under `FORCE_COLOR`, and the real urllib transport against a loopback server in seven scenarios (success, a 401 and a 500 that echo the auth header, malformed JSON, a redirect to a private host, connection refused, a cursor loop). Zero hits in stdout, stderr, the output, the working directory or the audit log.
- **Default path**: byte-identical output to `main` on a mixed fixture (SHA-256 match), and no `urllib.request` or pull module in `sys.modules` afterwards.
- **17 mutations killed**: validator bypass, http allowed, loopback gate removed, redirects followed, secret in a 401 message, scrubbing disabled, a silent page cap, Retry-After ignored or uncapped, the retry budget raised, the timeout dropped, the size cap removed, a top-level import, the GENERATION parameter dropped, and two adapter-mapping changes.
- 254 passed / 1 skipped across the eight related files; ANSI-safe; all 13 required checks green on `0623892c`.

Three small findings did not block the merge. They are filed as #865: the timeout is per socket operation rather than a total deadline; the GENERATION filter is server-side only; and a repeated cursor runs until `--max-pages`. Yours if you want them — say so there.

PR #860, opened 2026-09-11
fix(cli): route terminal plots through Rich

### LINKED ISSUE #824: soup data stats writes 528 raw ANSI escape bytes into redirected output even with NO_COLOR=1: the plotext histogram bypasses Rich
## The defect

`soup data stats` prints its table through Rich, which honours `NO_COLOR` and drops styling when stdout is not a terminal. The length histogram that follows does not go through Rich: `commands/data.py:625-633` calls `utils/plotext_compat.render_histogram(..., theme="dark")`, which ends in `figure.show()` (plotext 6) or `plotext.show()` (plotext 5) and writes coloured output straight to `sys.stdout`. Nothing checks `NO_COLOR`, `isatty()` or the console's colour system first.

## Reproduction

Windows 11, Python 3.12, plotext 6.1.0, a 50-row alpaca JSONL:

```
$ NO_COLOR=1 soup data stats d.jsonl > out.txt 2>&1
exit=0
```

Byte counts of `out.txt`:

```
bytes 8490  ESC bytes 528
distinct SGR sequences: [(b'\x1b[0m', 216), (b'\x1b[48;5;0m', 191), (b'\x1b[38;5;12m', 77),
                         (b'\x1b[38;5;3m', 41), (b'\x1b[48;5;15m', 3)]
first line with ESC: line 19 b'\x1b[0m\x1b[38;5;3m\x1b[48;5;0m        ...  \x1b[38;5;12m\x1b[48;5;15mText Length Distribution (chars)...'
ESC before histogram (Rich table part): 0
```

The Rich table contributes zero escape bytes; all 528 come from the histogram, including a black background (`48;5;0`) painted behind every cell.

Changing the theme alone does not fix it. Calling `render_histogram` directly with a redirect:

```
dark       ESC 251
default    ESC 293
colorless  ESC 118
```

plotext emits escape codes even with its `colorless` theme, so the fix has to strip or skip, not re-theme.

## Why it matters

`soup data stats d.jsonl > stats.txt`, `| less`, `| tee`, CI logs and notebook cells all get a file full of `\x1b[48;5;0m`. `NO_COLOR` is a cross-tool convention (https://no-color.org) that users set precisely to avoid this, and the rest of Soup respects it through Rich. The hard-coded `"dark"` theme also paints a black background on light terminals.

`soup runs replay` uses the same helper with the same hard-coded theme (`commands/runs.py:558-568`, `render_line(..., theme="dark")`), so it very likely has the same defect; it was not reproduced here because it needs a recorded run.

## Proposed fix

In `utils/plotext_compat.py`, build the plot to a string instead of calling `show()` (plotext 6 exposes `figure.build()`; check the 5.x equivalent for the fallback branch), then:

- if the console is not a terminal or `NO_COLOR` is set, strip codes with `plotext.uncolorize(...)` (or skip the histogram entirely when not a TTY) before printing;
- otherwise print it through the Rich console (`Text.from_ansi`) so Rich's colour detection is the single decision point.

Drop the hard-coded `theme="dark"` in favour of the default theme, or make it a parameter both call sites pass explicitly.

## Evidence: why no test caught it

Typer's `CliRunner` does not capture the histogram at all. Invoking `data stats` through it gives `exit 0`, `ESC in output: 0`, and the whole plot is printed to the real process stdout instead of `result.output` — so a `CliRunner` assertion passes today while the real command is broken. The regression test has to run the command as a subprocess (or the fix has to route the plot through the Rich console, after which `CliRunner` would see it).

## Acceptance criteria

- [ ] A test runs `python -m soup_cli.cli data stats <fixture>` as a subprocess with stdout piped and `NO_COLOR=1`, and asserts no `\x1b` byte in the output — fails on current code (528 on the fixture above).
- [ ] The histogram is written through the Rich console, so `CliRunner` output contains it.
- [ ] The same no-escape assertion for `soup runs replay`.
- [ ] Interactive output in a colour terminal still shows the histogram.

akkupratap323:
﻿## What does this PR do?

Plotext wrote charts directly to `sys.stdout`, bypassing Rich's terminal and `NO_COLOR` handling. As a result, `soup data stats > stats.txt` and `soup runs replay` could emit hundreds of raw ANSI escape bytes, while `CliRunner` did not capture the chart.

This change builds Plotext 5/6 charts as strings, parses their ANSI styling into Rich `Text`, and prints them through the command's Rich console. Redirected and `NO_COLOR=1` output is clean, interactive terminals retain color, and both charts are visible to `CliRunner`. The hard-coded dark theme is also removed so Rich and the active terminal control presentation.

Closes #824

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

## Validation

- 73 focused Plotext, data stats, replay, Windows encoding, and subprocess tests passed
- Real subprocess checks verify `NO_COLOR=1` output for both `data stats` and `runs replay` contains the chart and no `\x1b` bytes
- A forced interactive Rich console test verifies chart color remains enabled

The full local suite requires the optional training dependency stack; the focused command and compatibility suites pass with the core development dependencies installed.

### DIFF
--- a/changelog.d/0.74.0/860.fixed.md
+++ b/changelog.d/0.74.0/860.fixed.md
@@ -1,0 +1,2 @@
+- Route terminal charts through Rich so `NO_COLOR` and redirected output contain no raw
+  ANSI escapes while interactive charts retain colour (#824 by @akkupratap323 in #860).
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -625,11 +625,11 @@ def stats(
                 render_histogram(
                     plt,
                     lengths,
+                    console=console,
                     bins=30,
                     title="Text Length Distribution (chars)",
                     xlabel="Length",
                     ylabel="Count",
-                    theme="dark",
                 )
             finally:
                 sys.stdout = original_stdout
--- a/src/soup_cli/commands/runs.py
+++ b/src/soup_cli/commands/runs.py
@@ -561,11 +561,11 @@ def _plot_loss_curve(metrics: list[dict]) -> None:
         plt,
         steps,
         losses,
+        console=console,
         label="loss",
         title="Training Loss",
         xlabel="Step",
         ylabel="Loss",
-        theme="dark",
     )
 
 
--- a/src/soup_cli/utils/plotext_compat.py
+++ b/src/soup_cli/utils/plotext_compat.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 from typing import Any
 
+from rich.text import Text
+
 
 def _v6_figure(plotext: Any) -> Any | None:
     """Return plotext 6's figure object, or ``None`` for the 5.x API."""
@@ -15,13 +17,14 @@ def render_histogram(
     plotext: Any,
     values: list[int],
     *,
+    console: Any,
     bins: int,
     title: str,
     xlabel: str,
     ylabel: str,
-    theme: str,
+    theme: str | None = None,
 ) -> None:
-    """Render a histogram with either plotext 5.x or 6.x."""
+    """Render a histogram through Rich with either plotext 5.x or 6.x."""
     figure = _v6_figure(plotext)
     if figure is not None:
         figure.clear()
@@ -29,8 +32,9 @@ def render_histogram(
         figure.title(title)
         figure.label(xlabel, axis=0)
         figure.label(ylabel, axis=1)
-        figure.theme(theme)
-        figure.show()
+        if theme is not None:
+            figure.theme(theme)
+        console.print(Text.from_ansi(str(figure.build())))
         return
 
     plotext.clf()
@@ -38,8 +42,9 @@ def render_histogram(
     plotext.title(title)
     plotext.xlabel(xlabel)
     plotext.ylabel(ylabel)
-    plotext.theme(theme)
-    plotext.show()
+    if theme is not None:
+        plotext.theme(theme)
+    console.print(Text.from_ansi(str(plotext.build())))
 
 
 def render_line(
@@ -47,13 +52,14 @@ def render_line(
     x_values: list[int],
     y_values: list[float],
     *,
+    console: Any,
     label: str,
     title: str,
     xlabel: str,
     ylabel: str,
-    theme: str,
+    theme: str | None = None,
 ) -> None:
-    """Render a labelled line with either plotext 5.x or 6.x."""
+    """Render a labelled line through Rich with either plotext 5.x or 6.x."""
     figure = _v6_figure(plotext)
     if figure is not None:
         figure.clear()
@@ -62,8 +68,9 @@ def render_line(
         figure.title(title)
         figure.label(xlabel, axis=0)
         figure.label(ylabel, axis=1)
-        figure.theme(theme)
-        figure.show()
+        if theme is not None:
+            figure.theme(theme)
+        console.print(Text.from_ansi(str(figure.build())))
         return
 
     plotext.clf()
@@ -71,5 +78,6 @@ def render_line(
     plotext.title(title)
     plotext.xlabel(xlabel)
     plotext.ylabel(ylabel)
-    plotext.theme(theme)
-    plotext.show()
+    if theme is not None:
+        plotext.theme(theme)
+    console.print(Text.from_ansi(str(plotext.build())))
--- a/src/soup_cli/utils/plotext_compat.py
+++ b/src/soup_cli/utils/plotext_compat.py
@@ -34,7 +34,7 @@ def render_histogram(
         figure.label(ylabel, axis=1)
         if theme is not None:
             figure.theme(theme)
-        console.print(Text.from_ansi(str(figure.build())))
+        console.print(Text.from_ansi(str(figure.build())), soft_wrap=True)
         return
 
     plotext.clf()
@@ -44,7 +44,7 @@ def render_histogram(
     plotext.ylabel(ylabel)
     if theme is not None:
         plotext.theme(theme)
-    console.print(Text.from_ansi(str(plotext.build())))
+    console.print(Text.from_ansi(str(plotext.build())), soft_wrap=True)
 
 
 def render_line(
@@ -70,7 +70,7 @@ def render_line(
         figure.label(ylabel, axis=1)
         if theme is not None:
             figure.theme(theme)
-        console.print(Text.from_ansi(str(figure.build())))
+        console.print(Text.from_ansi(str(figure.build())), soft_wrap=True)
         return
 
     plotext.clf()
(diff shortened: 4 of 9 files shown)


akkupratap323:
@MakazhanAlpamys The #824 fix is ready, including the required #860 changelog fragment. The focused suite passes (73 tests), and the full Ruff check passes. Could you please review and assign this PR if appropriate? Thank you.

MakazhanAlpamys:
Thank you, @akkupratap323. The approach is exactly what #824 asked for: build the plot to a string and let the Rich console decide colour in one place. You covered both plotext 5 and 6, and the real-tracker subprocess test for `runs replay` goes beyond the issue, which could not reproduce that path. Escape bytes in redirected output (50-row fixture, stdout to a file):

| | `main` | this PR |
|---|---|---|
| `data stats`, `NO_COLOR=1` | 605 | **0** |
| `runs replay`, `NO_COLOR=1` | 265 | **0** |
| forced-colour terminal, in-process | colour | colour (220 escapes) — interactive users keep their colour |

One blocker, and it lands on the platform #824 was reproduced on.

## Blocking: on Windows, redirected charts wrap at every row

`console.print(Text.from_ansi(...))` in `utils/plotext_compat.py` (lines 37, 47, 73, 83) lets Rich soft-wrap the chart at spaces. plotext builds the figure at the full terminal width, but when output is redirected on Windows, Rich's legacy renderer makes the console one column narrower. So every chart row wraps: a 22-line histogram becomes 42 lines, with the right border on its own line. Linux and macOS are unaffected, which is why CI is green.

**Fix:** `console.print(..., soft_wrap=True)`. At width 100 that brings it back to 23 lines. Your `_Console.print(self, value)` test doubles in `test_plotext_compat.py` reject keyword arguments, so they need `**kwargs` too. Please also add one layout assertion — rendered line count, or no line that is only a border fragment — so a future change to the print call cannot reintroduce the wrap unnoticed.

## Nice-to-have

`tests/test_replay.py:170-176` passes `env={"NO_COLOR": "1"}` to `CliRunner`, but that has no effect: the Rich console is created when the module is imported, before the runner sets the environment. Under `FORCE_COLOR=1 TERM=xterm-256color` that test fails. CI does not set those variables, so this is a lead rather than a verdict, but patching the command module's console with `Console(no_color=True)` makes it independent of the environment. Also, dropping `--no-plot` from `test_replay_renders` leaves `runs replay --no-plot` untested; only `runs show --no-plot` is still covered, so a separate one-line test would restore it.

## What I verified

- **Mutations**: putting back `figure.show()` / `plotext.show()` fails 8 tests, including both subprocess `NO_COLOR` tests, the `CliRunner` stats test and the replay test. Always stripping colour is caught by `test_rich_console_controls_plot_color`.
- **Changed assertions are legitimate**: `show` became `build` in the call-order checks because the code now calls `build()`. Nothing was weakened.
- plotext_compat 4, replay 49, data_tools 16, bugfixes 72, cli_subprocess 70, startup-lightness 34 and fragments 17 all pass. Ruff is clean, and all required checks are green on `96b14b8b`.

A process note, same as on #858: #824 had no claim comment when this opened. One line on the issue first is the claim ([CONTRIBUTING.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTING.md#claiming-an-issue)).

akkupratap323:
Thanks for the detailed review. I addressed the Windows wrapping blocker in a6fad37: all four plotext compatibility paths now print with `soft_wrap=True`, the test console records keyword arguments, and a narrow redirected-console regression test pins the original plot rows. I also applied both suggested test follow-ups by patching the runs command console with `Console(no_color=True)` and restoring explicit `--no-plot` coverage.

Validation: `54 passed` for `tests/test_plotext_compat.py tests/test_replay.py` with coverage disabled for the focused run; Ruff passes on all three changed files.

MakazhanAlpamys:
Merged as `09d97474`. Thank you, @akkupratap323 — and thank you for taking the `soft_wrap` note seriously rather than just adding the keyword.

**The part I want to name.** I asked for `soft_wrap=True` and gave you a reason; what came back was a test that *discriminates*. `test_rich_console_does_not_wrap_plot_rows` builds a `Console(width=4)` and asserts both the line count and that a long line survives intact, so removing the keyword fails with `['1234', '56', 'abcd', 'ef'] != ['123456', 'abcdef']`. That also kills the near-miss a future refactor is most likely to introduce — `no_wrap=True` without `crop=False`, which looks equivalent and is not. A test that only asserted "the plot text appears in the output" would have passed with or without the fix, and that is the failure mode this repo keeps finding in its own history.

**The measurements, for the record.** The wrapping cause is Windows-specific and worth writing down: Rich's `Console.size` returns `width - legacy_windows`, so Rich reports 79 columns where plotext has already drawn the plot at 80. That one-column overhang wrapped every row — 42 emitted lines against 23 with the fix. And the headline issue reproduces cleanly: 951 escape bytes on `origin/main` against 0 on your branch, same 50-row fixture, `NO_COLOR=1`, stdout piped, histogram still present in both.

Routing through `Text.from_ansi` was the right choice for a reason beyond colour: `Console.print` applies neither markup parsing nor the highlighter to a `Text` instance, so an axis label or run name containing `[` survives literally. This repo has a documented history of exactly that bug, and your approach is stronger than `markup=False` because it keeps the ANSI-derived styling for the interactive case.

**What I ran.** 246 passed across the touched files plus the subprocess suite; ruff clean; green at COLUMNS 40, 60 and 200. Four mutations, all killed: removing `soft_wrap=True` (3 tests), reverting the Rich routing entirely (7 unit + 2 subprocess), the `no_wrap`-without-`crop` near-miss (3), and printing the raw ANSI string instead of `Text.from_ansi` (5). I also verified the plotext 5.x fallback is real rather than assumed — `build()` is exported in 5.2.8, 5.2.2 and 5.0.2, and `pyproject.toml` declares `plotext>=5.2.0`, so the floor is safe. No test was deleted, skipped or loosened; `test_replay_renders` gained a second invocation and `test_installed_plotext_runtime_*` gained two assertions.

13/13 required checks green on `a6fad370`. Your `CONTRIBUTORS.md` entry is in `1fd48503`.

**One optional follow-up, take it or leave it.** In `tests/test_replay.py` the newest commit swapped `runner.invoke(..., env={"NO_COLOR": "1"})` for `monkeypatch.setattr(runs_command, "console", Console(no_color=True))`. I checked and the `env=` form passes without the monkeypatch, so the swap is not load-bearing — and it replaces the object under test rather than exercising the `NO_COLOR` path the issue is about. Your two subprocess tests cover the real path with the real binary, so nothing is actually lost, which is why I merged rather than asking. If you want it tidy, restoring `env={"NO_COLOR": "1"}` (or using `Console(no_color=True, color_system=None)` so the `"\x1b" not in output` assertion also survives a forced-colour environment) would be a small, welcome PR.

For awareness: @kok-o's #862 targets this same issue. Yours was opened first, at 16:11 UTC against their 17:36 claim, and I told them on the thread that yours goes first. I am closing theirs now with that explanation — nothing for you to do.

PR #861, opened 2026-09-11
fix(data): exit 2 on zero valid rows, add --strict and --min-valid-fraction (#811)

### LINKED ISSUE #811: soup data validate exits 0 at "0/5 rows valid", so the `soup ci init` gate passes a dataset the loader reduces to zero rows

kok-o:
## What does this PR do?

Resolves #811 by updating `soup data validate` exit codes and adding strict validation options:
1. **Zero-valid-rows failure**: When `total > 0 and valid_rows == 0`, `soup data validate` now exits with code `2` (previously exited `0`), ensuring files with no usable rows fail CI gates.
2. **Default partial drop compatibility**: Partial drops (e.g. 1/2 rows valid) continue to default to exit `0` as before.
3. **Strict mode**: Added `--strict` to require 100% of rows to be valid, exiting `2` if any row is dropped.
4. **Valid fraction threshold**: Added `--min-valid-fraction` (float between 0.0 and 1.0) to exit `2` when the fraction of valid rows falls below the threshold. Out-of-range thresholds exit `1`.
5. **Docs and help**: Documented exit codes (0 = valid, 1 = file not found / bad argument / undetectable format, 2 = validation failed) in `--help` and `docs/data.md`.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

### DIFF
--- a/changelog.d/0.74.0/861.fixed.md
+++ b/changelog.d/0.74.0/861.fixed.md
@@ -1,0 +1,4 @@
+- `soup data validate` now exits 2 when zero rows are valid for the format,
+  preventing datasets with no usable rows from silently passing CI gates
+  (#811 by @kok-o in #861). Added `--strict` to reject any dropped row
+  and `--min-valid-fraction` to enforce an opt-in valid-row threshold.
--- a/docs/data.md
+++ b/docs/data.md
@@ -736,6 +736,9 @@ soup data inspect ./data/train.jsonl
 # Validate format (auto-detects if --format not specified)
 soup data validate ./data/train.jsonl
 soup data validate ./data/train.jsonl --format alpaca
+soup data validate ./data/train.jsonl --strict                 # exit 2 on any dropped row
+soup data validate ./data/train.jsonl --min-valid-fraction 0.95 # exit 2 below 95% valid
+# Exit codes: 0 = valid, 1 = missing file / bad arguments, 2 = validation failed (0 valid or below threshold)
 
 # Convert between formats
 soup data convert ./data/train.jsonl --to sharegpt --output converted.jsonl
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -82,11 +82,32 @@ def validate(
         "auto", "--format", "-f",
         help="Expected format: auto, alpaca, sharegpt, chatml, dpo, kto, plaintext",
     ),
+    min_valid_fraction: Optional[float] = typer.Option(
+        None, "--min-valid-fraction",
+        help="Minimum fraction of valid rows required (0.0 to 1.0); exits 2 if below",
+    ),
+    strict: bool = typer.Option(
+        False, "--strict",
+        help="Require 100% of rows to be valid; exits 2 if any rows are dropped",
+    ),
 ):
-    """Validate dataset format and report issues."""
+    """Validate dataset format and report issues.
+
+    Exit codes:
+      0: Dataset passed validation.
+      1: File not found, undetectable format, or invalid arguments.
+      2: Validation failed (0 valid rows, below --min-valid-fraction, or
+         dropped rows with --strict).
+    """
     file_path = Path(path)
     if not file_path.exists():
         console.print(f"[red]File not found: {file_path}[/]")
+        raise typer.Exit(1)
+
+    if min_valid_fraction is not None and (min_valid_fraction < 0.0 or min_valid_fraction > 1.0):
+        console.print(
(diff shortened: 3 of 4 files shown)


MakazhanAlpamys:
@kok-o — thank you for this, and I am sorry to close it. #811 already had an open pull request when you claimed it: #858 by @akkupratap323, opened at 15:58 UTC and in review. Your claim came at 17:31, and #858 merged at 18:24 with the same zero-rows exit-2 behaviour and a `--min-valid-fraction` threshold. So this one is superseded, and I am closing it rather than asking you to rebase onto a fix that has already landed.

A cheap habit that avoids this: before claiming, check the issue's sidebar and `gh pr list --search "<issue number>"` for an open PR. A claim comment on an issue that already has a PR under review does not take it over.

Issue #818, opened 2026-09-11
soup edit diff without both models exits 0 and reports "0 changed" for probes it never generated; one model flag or a wrongly-keyed probe file is silently ignored


MakazhanAlpamys:
## The defect

`soup edit diff <before> <after>` only generates anything when `--before-model`, `--after-model` AND at least one loaded probe are all present (`utils/edit_diff.py:198`):

```python

live = before_model is not None and after_model is not None and probes

```

Every other combination falls through to a placeholder (`utils/edit_diff.py:217-233`) that writes `changed=False` for each probe and exits 0. The table footer and the JSON both say nothing changed, which is a claim about the models that was never measured:

- **No models:** every probe reports `changed: false`, footer `Total probes: 2; 0 changed.`

- **Only one of `--before-model` / `--after-model`:** the flag the user did pass is dropped without a word — not even a load attempt (a nonexistent model id is accepted).

- **Probe rows without a `prompt` key:** `load_probes` skips malformed rows silently (`utils/edit_diff.py:147-149`), so a file of `{"text": ...}` rows loads as zero probes, `live` becomes false, and the table prints `no probes supplied` — even though `--probes` was given, and even when both models were given.

- **Run ids:** the positional arguments are documented as "Registry run id of the model BEFORE/AFTER the edit" but are only checked for non-empty / not-equal. They are never looked up, so any two strings work.

## Reproduction

Table rows trimmed to fit; nothing else edited.

```

$ cat probes.jsonl

{"prompt": "The Eiffel Tower is located in"}

{"prompt": "The capital of France is"}



$ soup edit diff no-such-run-A no-such-run-B --probes probes.jsonl -o diff.json

│ The Eiffel Tower is located in │ <supply --before-model + --after-model to generate> │ ... │   no    │

│ The capital of France is       │ <supply --before-model + --after-model to generate> │ ... │   no    │

Total probes: 2; 0 changed.

Wrote diff -> diff.json

exit=0



$ cat diff.json            # excerpt

  "after_run_id": "no-such-run-B",

  "before_run_id": "no-such-run-A",

  "changes": [

    {

      "after": "<supply --before-model + --after-model to generate>",

      "before": "<supply --before-model + --after-model to generate>",

      "changed": false,

      "prompt": "The Eiffel Tower is located in"

    }, ...

  "total_probes": 2



$ soup edit diff runA runB --probes probes.jsonl --before-model does-not/exist

...same placeholder table...

[...]

PR #862, opened 2026-09-11
fix(data): uncolorize plotext output under NO_COLOR and redirect (#824)

### LINKED ISSUE #824: soup data stats writes 528 raw ANSI escape bytes into redirected output even with NO_COLOR=1: the plotext histogram bypasses Rich

kok-o:
## What does this PR do?

Fixes #824 by ensuring plotext charts rendered in `soup data stats` and `soup runs replay` route through the Rich console and uncolorize when output is redirected or `NO_COLOR` is set:

1. **Rich Console Routing**: Refactored `render_histogram` and `render_line` in `soup_cli.utils.plotext_compat` to call `build()` instead of directly printing via `show()`. Output is passed through the command's Rich `Console`, allowing `CliRunner` and test harnesses to capture output properly.
2. **ANSI Escape Stripping**: Detects `NO_COLOR` in environment, `console.no_color`, or non-terminal redirected output (`not console.is_terminal`), and strips raw ANSI escape codes using `plotext.uncolorize(raw)` with regex fallback.
3. **Avoid Hardcoded Dark Theme Backgrounds**: Dropped hardcoded `theme="dark"` from `data stats` and `runs replay`, avoiding unwanted background paint escape codes (`\x1b[48;5;0m`) when redirecting to files or uncolored terminals.
4. **Regression Tests**:
   - Subprocess test with `NO_COLOR=1` confirming 0 raw ANSI escape bytes in stdout while preserving chart text.
   - Subprocess test with redirected pipe stdout confirming 0 raw ANSI escape bytes.
   - `CliRunner` test confirming chart output is captured without ANSI escapes.
   - Unit tests covering 5.x and 6.x mock doubles and `render_line` / `render_histogram`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

### DIFF
--- a/changelog.d/0.74.0/862.fixed.md
+++ b/changelog.d/0.74.0/862.fixed.md
@@ -1,0 +1,8 @@
+- **Uncolorize plotext charts under NO_COLOR and redirected output (#824 by @kok-o in #862).**
+  `soup data stats` and `soup runs replay` rendered histograms and loss curves using
+  `plotext.show()`, which wrote directly to the process stdout file descriptor, bypassing
+  Rich console routing and emitting raw ANSI escape sequences (`\x1b[...]`) even when `NO_COLOR=1`
+  was set or output was piped to a file. Rendered charts are now built via `.build()`,
+  uncolorized when `NO_COLOR` is active, redirected, or running in an uncolored console,
+  and printed through Rich, allowing test runners and piped outputs to cleanly capture
+  text without ANSI escape contamination or terminal background paint.
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -629,7 +629,7 @@ def stats(
                     title="Text Length Distribution (chars)",
                     xlabel="Length",
                     ylabel="Count",
-                    theme="dark",
+                    console=console,
                 )
             finally:
                 sys.stdout = original_stdout
--- a/src/soup_cli/commands/runs.py
+++ b/src/soup_cli/commands/runs.py
@@ -565,7 +565,7 @@ def _plot_loss_curve(metrics: list[dict]) -> None:
         title="Training Loss",
         xlabel="Step",
         ylabel="Loss",
-        theme="dark",
+        console=console,
     )
 
 
--- a/src/soup_cli/utils/plotext_compat.py
+++ b/src/soup_cli/utils/plotext_compat.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import os
+import re
 from typing import Any
 
 
@@ -9,6 +11,23 @@ def _v6_figure(plotext: Any) -> Any | None:
     """Return plotext 6's figure object, or ``None`` for the 5.x API."""
     figure = getattr(plotext, "figure", None)
     return figure if figure is not None and not callable(figure) else None
+
+
+def _print_chart(plotext: Any, raw: str, console: Any = None) -> None:
+    """Output a rendered chart via the Rich console, stripping escapes when uncolored."""
+    from rich.console import Console
+    from rich.text import Text
+
+    cons = console if console is not None else Console()
+    if "NO_COLOR" in os.environ or cons.no_color or not cons.is_terminal:
+        clean = getattr(plotext, "uncolorize", None)
+        if callable(clean):
+            text_str = clean(raw)
+        else:
+            text_str = re.sub(r"\x1b\[[0-9;?]*[a-zA-Z]", "", raw)
+        cons.print(Text(text_str))
+    else:
+        cons.print(Text.from_ansi(raw))
 
 
 def render_histogram(
@@ -19,7 +38,8 @@ def render_histogram(
     title: str,
     xlabel: str,
     ylabel: str,
-    theme: str,
+    theme: str | None = None,
+    console: Any = None,
 ) -> None:
     """Render a histogram with either plotext 5.x or 6.x."""
     figure = _v6_figure(plotext)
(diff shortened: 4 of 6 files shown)


MakazhanAlpamys:
@kok-o — same situation as #861, and a heads-up so the wait is visible rather than mysterious. #824 already had #860 by @akkupratap323, opened at 16:11 UTC, reviewed, with one change left (`soft_wrap=True`, so charts do not wrap on Windows when output is redirected). Your claim came at 17:36.

We do not run two PRs for one issue, so #860 goes first. I am leaving this open for now rather than closing it over your head: if #860 has not moved by **2026-09-13 18:00 UTC**, I will review this one instead. Either way, please do check for an open PR before claiming next time — the issue sidebar shows it.

#863 (for #818, which had no other PR) and your pushed fixes on #743 are both in my review queue.

MakazhanAlpamys:
@kok-o — #860 moved before the date I set, so as I said on 2026-09-11 it is the one that merged (`09d97474`), and I am closing this rather than leaving it hanging.

**You were not wrong about the diagnosis, and two of your four points are exactly what shipped**: building to a string via `figure.build()` / `plotext.build()` instead of `show()`, and dropping the hardcoded `theme="dark"` which was adding `\x1b[48;5;0m` background codes on top of everything else. Your implementation plan named both before either PR was reviewed.

Since you put real work into this, the honest comparison rather than a bare "superseded":

**Where the merged approach differs.** #860 routes everything through `Text.from_ansi(...)` and lets Rich's colour system decide, with no `NO_COLOR` / TTY predicate of its own. Yours calls `plotext.uncolorize()` (with a regex fallback) when `NO_COLOR` is set, `console.no_color` is on, or stdout is not a terminal. Both produce zero escape bytes on the reported case. I took the first because it leaves **one** decision point instead of a hand-rolled predicate maintained next to Rich's own — the two can drift, and the regex fallback is a second ANSI parser in a tree that should not need one.

**Where it differs in coverage, and this is the part that decided it.** A second defect turned up during review that neither issue text nor your plan mentions: Rich's `Console.size` returns `width - legacy_windows`, so on Windows Rich reports 79 columns where plotext has already drawn the plot at 80. That one-column overhang wrapped **every row** of the chart — measured 42 emitted lines against 23. Uncolorizing does not touch it, so a chart with no escape bytes would still have arrived mangled. #860 fixes it with `soft_wrap=True` and a test that asserts at `Console(width=4)` both the line count and that a long line survives intact.

Your `tests/test_issue824_plotext_no_color_ansi.py` was good work — the subprocess-with-piped-stdout shape is right, and #860 lands equivalent coverage.

On process, and I want to be accurate because I was terse about it last time: the ordering was #860 at 16:11 UTC, your claim at 17:36, your PR at 17:42. So the request to check the issue sidebar for an open PR before claiming stands — but you claimed and delivered inside six minutes, which is not the behaviour I am complaining about.

**#863** (for #818, which has no competing PR) and your pushed fixes on **#743** are both still in my queue and I will get to them. If you would rather have something with no collision risk at all, **#874** and **#875** are freshly filed and unclaimed: #874 is the one I would take — `soup doctor` prints a red `INCOMPATIBLE` and exits 0, because every package with a declared ceiling lives in the `[train]` group and the blocking branch is unreachable. Say the word on either and it is yours.

PR #863, opened 2026-09-11
fix(edit): require model pair and valid probe keys in edit diff (#818)

### LINKED ISSUE #818: soup edit diff without both models exits 0 and reports "0 changed" for probes it never generated; one model flag or a wrongly-keyed probe file is silently ignored

kok-o:
## What does this PR do?

Fixes #818 by eliminating unmeasured placeholder diff verdicts (`changed: false`) and enforcing strict validation of model pairs and probe files in `soup edit diff`:

1. **Model Requirement for Probes**:
   - `build_diff_report` now rejects probe evaluations when either `--before-model` or `--after-model` is missing, exiting `2` with an informative message (`both --before-model and --after-model are required when --probes is provided`).
2. **Model Pair Symmetry**:
   - Specifying only `--before-model` exits `2` and explicitly names `--after-model`.
   - Specifying only `--after-model` exits `2` and explicitly names `--before-model`.
3. **Strict Probe File Validation**:
   - `load_probes` tracks skipped lines and why they were skipped.
   - If a probe file contains rows that lack the required `'prompt'` key (e.g. `{"text": ...}`), empty prompts, or invalid JSON resulting in zero loaded probes, it raises `ValueError` stating the issue and explicitly naming the key (e.g. `missing 'prompt' key`), causing the CLI to exit `2`.
4. **Zero Unmeasured Verdicts**:
   - Eliminated the placeholder loop that wrote `changed=False` for unmeasured probes. Now `FactChange` rows are only created by live evaluation of both models in `_generate_live_changes`.
   - Running `soup edit diff` without probes prints `no probes supplied` and produces an empty `changes: []` list rather than fabricating probe entries.
5. **Help Text Clarity**:
   - Clarified `before` and `after` positional arguments in help text as metadata labels / run IDs.
6. **Regression Tests**:
   - Added `tests/test_issue818_edit_diff_validation.py` covering all acceptance criteria with `CliRunner`.
   - Updated existing unit tests in `tests/test_v0610_part_e.py` and `tests/test_v0719.py`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist


MakazhanAlpamys:
@kok-o — reviewed at `3da15d8a`, 13/13 green. **This is a clean, disciplined PR**, and I want to say that plainly given you have had two closed as duplicates this week through no fault of your own, and this one sat while I worked through that queue. Thank you for the patience.

## What makes it good

**The message content is actually pinned, not just the exit code.** I mutated a refusal into a generic `invalid input` while keeping the exit code, and a test failed. That is the difference between a guard and a guard someone can quietly hollow out later, and most PRs I review do not have it.

**You removed the placeholder path rather than patching around it.** #818's fourth acceptance criterion is that no code path writes `"changed": false` for a probe that was never generated, and deleting that branch is the only fix that actually satisfies it — a warning beside it would not have.

**And you took the honest branch on the run-id question.** #818 offered "resolve run ids against the registry, *or* relabel them as free text in help". Relabelling is the smaller, truer change, and you did not quietly leave the help claiming something the code does not do.

Both modified tests were tightened, not weakened: each tested exactly the behaviour being removed, both became `pytest.raises`, and `test_with_probe_file` was split so the happy path gained model wiring and a `changed is True` assertion it never had.

**What I ran.** 399 passed, 7 skipped across all nine files touching `edit_diff` / `knowledge_edit` / `edit_governor` / `edit_kernels` / `commands.edit`, plus the startup-lightness guard. ruff clean. Eleven mutations, **ten killed** — removing each guard separately, inverting each condition, letting an invalid probe key fall back to `text`/`question`, restoring the placeholder path, changing the exit code, and downgrading the refusal to a warning. The eleventh survivor is a *provably equivalent* mutant rather than a coverage gap: changing the probes guard's `or` to `and` is unreachable while the asymmetric guard runs before it, and removing that guard too makes the same mutation fail three tests. Your `or` is defence-in-depth, and I checked it rather than assuming.

On the exit code, since it contradicts the repo-wide convention: **2 is correct here.** All nine `raise typer.Exit` sites in `commands/edit.py` are `Exit(2)`, #818 says "exit 2" three times, and you reuse the file's existing handler rather than adding one. The divergence from the global 1-for-usage rule is pre-existing and file-wide — it is what #813 tracks — and fixing it in this PR would have contradicted the issue and diverged from eight sibling paths.

Your changelog fragment is correct on every axis, including the `(#818 by @kok-o in #863)` form that most contributors get wrong.

## Two one-liners before I merge

**1. `docs/commands.md:244` now contradicts the code.** It renders the model flags as optional — `[--before-model <m> --after-model <m>]` with "(live before/after generation when both models given)" — and after this PR they are *required* whenever `--probes` is given. `docs/adapters-and-governance.md:159` already shows the full form and needs nothing.

**2. The null-byte skip has no counter, so it reports the wrong reason.** `edit_diff.py:163-164` does a bare `continue`, unlike the four skip reasons around it. A file whose rows all carry a null byte in the prompt therefore falls through to `probe file '…' is empty (0 probes loaded)` — and the file is not empty. Measured: exit 2, that exact message. It refuses correctly and names the wrong cause, which is the defect class #818 exists to close, so I would rather it did not ship one instance of it inside the fix for it. A `skipped_null_byte` counter plus one more branch, mirroring the other four.

## One cheap hardening I would take while you are in there

`tests/test_issue818_edit_diff_validation.py:90` asserts `"missing 'prompt' key"` against raw `result.output`. Under `FORCE_COLOR=1 TERM=xterm-256color` it fails, because Rich's `ReprHighlighter` colours the quoted token into `missing \x1b[32m'prompt'\x1b[0m key`. And at `COLUMNS=60` a different test in the file fails on line-wrapping alone.

**Neither will go red on merge** — CI sets no `FORCE_COLOR` or `TERM` (I grepped the workflows) and CliRunner's stream is not a tty — so this is a lead, not a blocker. But it is worth knowing it is a *new variant* of the family that has turned this repo's CI red four times: the repo-wide guard `test_cli_help_assertions_are_ansi_safe.py` passes here because it is deliberately scoped to `--help` output, and its docstring assumes error messages interpolate plain text Rich does not style per-token. That assumption holds for `--after-model` and breaks for quoted strings, numbers and parens. An ANSI strip plus whitespace collapse, per the house helper, fixes the colour and the width case together.

## And one thing I filed rather than asked you to fold in

The mirror of what you fixed is still open: both models given and `--probes` *forgotten* exits 0 with `Total probes: 0` and an empty `changes` array. It is materially better than the old bug — a consumer can tell it apart, because `total_probes` is 0 rather than a non-zero count with `changed: false` — and it is outside #818's acceptance criteria, so folding it in would have been scope creep. Filed as **#880**, roughly three lines beside your existing guard, and it is yours if you want it.

Push the two one-liners and I merge on green. **#743 is next in my queue** — you pushed fixes to it yesterday and it is already with a reviewer.

MakazhanAlpamys:

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:

```
git mv changelog.d/0.74.0/863.fixed.md changelog.d/0.75.0/863.fixed.md
git push --force-with-lease
```

MakazhanAlpamys:
Merged as `491f2053`. Both one-liners verified, and I owe you an explanation for the delay rather than just a thank-you.

You pushed `f05c4428` on **2026-09-12 at 17:19** — both asks done, plus the optional ANSI hardening I had not required — and said nothing, which is entirely reasonable. But my triage keys on the last *comment*, so a silent push that completes a PR is invisible to it: from my side this thread's last message was still my own. It sat green and mergeable for **eighteen hours** for no reason except that. I have just swept every open PR for this pattern and found **eleven** silent pushes, five of them merge-ready. That is my process failing, not your etiquette.

What I verified rather than read:

- `docs/commands.md:244` now shows `--before-model` / `--after-model` as required, matching the implementation.
- `128 passed, 2 skipped` across the edit-diff suites, and the null-byte counter is genuinely pinned:

```
replace `skipped_null_byte += 1` with a no-op   (anchor matched once, file re-parsed)
-> FAILED test_edit_diff_probes_with_null_byte_exits_2
   1 failed, 127 passed
```

**The second half of this fix is the one worth naming.** Requiring both models is the obvious part. Counting the dropped rows is what turns "a truncated run and a complete run look identical" into something an operator can see — and silently dropping input is the defect shape this project keeps finding in its own code. You fixed both halves when the issue arguably only demanded the first.

**On #880:** @Srinivasan8888 has been waiting on this landing — it is a three-line mirror of the same check inside the function you just rewrote, and they correctly refused to write it against `main` while your branch was open. It is unblocked now. If you would rather fold it in yourself, say so within the day and it is yours; otherwise I will tell them to go ahead, and I will say as much on that issue either way.

Your #921 is still red on 14 pre-existing tests and #917 is the declared head of that chain — both reviewed, both waiting on you.

Issue #275, opened 2026-06-21
[Tracking] Recipe task-variants for the 2026 model families (DPO / GRPO / pretrain)


MakazhanAlpamys:
v0.71.24 shipped 17 **SFT** recipes for the open-weight models released Feb–Jun 2026 (Qwen3.5/3.6, DeepSeek-V4, GLM-5.1, Kimi-K2.5/K2.6, MiniMax-M3, Mistral-Large-3). See the [v0.71.24 release](https://github.com/MakazhanAlpamys/Soup/releases/tag/v0.71.24) and `src/soup_cli/recipes/catalog.py`.

This tracking issue collects the **task-variant** recipes for those same families that newcomers can add — DPO, GRPO (reasoning), and pretrain (Base checkpoints). The shipped SFT recipes are the copy-paste template; each variant is a small, self-contained `RecipeMeta(...)` dict entry + a test row.

### Copy-paste templates already in the catalog
- **DPO:** `qwen2.5-7b-dpo`, `gemma3-27b-dpo`
- **GRPO (reasoning):** `qwen3-8b-grpo`, `qwen3-30b-a3b-reasoning-grpo` (MoE)
- **pretrain:** `qwen2.5-7b-pretrain`

### Acceptance criteria (per variant recipe)
- [ ] New `RecipeMeta(...)` entry in `src/soup_cli/recipes/catalog.py`
- [ ] `soup recipes show <name>` / `soup recipes use <name>` work; the YAML loads via `load_config_from_string`
- [ ] Catalog count bumped everywhere (the count test + `CONTRIBUTING.md` / `docs/`)
- [ ] A parametrized test row (mirror `tests/test_v07124.py`)
- [ ] Base repo-ID verified to resolve on Hugging Face

### Good-first-issue breakdown
Separate issues are filed for: `qwen3.5-35b-a3b-dpo`, `qwen3.5-9b-grpo`, `qwen3.5-4b-pretrain`, `deepseek-v4-flash-grpo`, `glm-5.1-dpo`, `kimi-k2.6-grpo`. Pick one — they don't overlap. More variants (other sizes/families) are welcome beyond this list.

Srinivasan8888:
Taking **`qwen3.5-9b-dpo`** under the "more variants beyond this list" line — naming the specific recipe rather than the tracking issue, per the coordination note you wrote on #379.

**Why this one.** I measured the 2026 families in the catalog rather than picking by eye:

```
[...]

PR #864, opened 2026-09-11
feat(recipes): add qwen3.5-0.8b-grpo and qwen3.5-2b-grpo (#848, #275)

### LINKED ISSUE #848: Recipe: qwen3.5-0.8b-grpo and qwen3.5-2b-grpo, the small-model reasoning variants (part of #275)
## What

Add two recipes to `src/soup_cli/recipes/catalog.py`:

- `qwen3.5-0.8b-grpo`
- `qwen3.5-2b-grpo`

`Qwen/Qwen3.5-0.8B` and `Qwen/Qwen3.5-2B` ship SFT only. Unlike most #275 rows, these two are small enough that a contributor might actually run a few steps of them — which makes them the most useful rows in the tracker for anyone learning GRPO on their own hardware. **Comment on this issue before you start.**

## Base repos

Both checked on 2026-09-11 with `huggingface_hub.model_info(..., token=False)`: they resolve, are not gated, and report `model_type: qwen3_5`. Re-check them the same way; an HTTP status is not evidence (#661).

## What to copy, and from where

- Siblings: `qwen3.5-0.8b-sft` (`lr: 3e-4`, LoRA r=8 / alpha=16, `max_length: 2048`) and `qwen3.5-2b-sft` (`lr: 2e-4`, same LoRA and length). Both carry `modality: text`, `quantization: 4bit`, `batch_size: auto`.
- Task block: `r1-distill-qwen-1.5b-grpo`, the only small (1.5B) Qwen-architecture GRPO recipe, or `qwen3.5-9b-grpo` for the in-family choice. They agree on `lr: 1e-5`, `grpo_beta: 0.1`, `num_generations: 4`, `reward_fn: accuracy`, `batch_size: auto`, `gradient_accumulation_steps: 8`, LoRA r=16.

Decisions you will need to make and state, not inherit silently:

- **Learning rate.** The SFT siblings use 2e-4 / 3e-4; across all 23 GRPO recipes the values are `1e-5` (16), `5e-6` (4) and `1e-6` (3, the 135M `grpo-env-*` recipes). Copying the SFT rate into a GRPO recipe would be an outlier by more than an order of magnitude; say which GRPO precedent you follow.
- **LoRA rank.** The siblings use r=8, the GRPO templates r=16. Either is defensible; pick one and say why.
- **`max_length`.** The siblings use 2048 and the GRPO templates 4096; reasoning completions are long, so this matters more here than for SFT.

YAML assembled from the sibling plus the GRPO block parses through `load_config_from_string` on current `main`. That proves the schema accepts it, nothing more.

## The ratchets that must pass

- [ ] `RecipeMeta(...)` entries, tags following the siblings plus `grpo` / `reasoning`.
- [ ] `modality: text` in both YAMLs and both names added to `EXPECTED_QWEN35_TEXT_RECIPES` in `tests/test_issue427_qwen35_text_modality.py:11`; the family audit at `:43` fails otherwise (#427).
- [ ] Model id pinned on **two independent surfaces** (`RecipeMeta.model` and YAML `base:`), each by its own test — mirror `TestDeepSeekV4FlashDpoRecipe`, `tests/test_recipes.py:503`.
- [ ] Catalog count bumped everywhere (165 today; use the number at rebase time): `catalog.py:51`; `test_catalog_size_is_N` in `tests/test_recipes.py:936`, `tests/test_v07130.py:740`, `tests/test_v07132.py:953` plus the history line near `tests/test_recipes.py:964`; `tests/test_v07124.py:281` and the docstring at `:277`; the doc sites in `tests/test_recipe_count_is_synced.py:83-96`.
- [ ] `python scripts/generate_recipe_snapshot.py` run last; additive diff only.
- [ ] Changelog fragment `changelog.d/0.74.0/<PR>.added.md`.
- [ ] `soup recipes show` / `use` run live for both; paste the output.

## Optional, and welcome

If you have a GPU, a short `soup train` smoke run (a handful of steps on a small reasoning set) with the card, VRAM and whether the loss moved stated in the PR would make this the first #275 recipe with any runtime evidence. It is **not** required for merge, and please do not present a few steps as a tuned recommendation.

## A note on mutations

`grpo_beta: 0.1`, `num_generations: 4`, `reward_fn: accuracy`, `epochs: 3`, `batch_size: auto` equal their schema defaults; deleting them is an equivalent mutation. Report them as survivors.

SID-6921:
## Summary

Closes #848 (part of #275). Adds `qwen3.5-0.8b-grpo` and `qwen3.5-2b-grpo` to the catalog — `Qwen/Qwen3.5-0.8B` and `Qwen/Qwen3.5-2B` currently ship SFT only. Both are small enough that a contributor might actually run a few steps on their own hardware, which the issue flags as the most useful rows in this part of the tracker.

## Base repos

Checked 2026-09-11 with `huggingface_hub.model_info(..., token=False)`, not an HTTP status (#661): both resolve, are not gated, and report `model_type: qwen3_5` — same as the SFT siblings and the existing `qwen3.5-9b-grpo` template.

## Decisions made explicitly, not inherited silently

- **`lr: 1e-5`** — the majority convention across the 23 existing GRPO recipes (16 use `1e-5`), not the SFT siblings' `2e-4`/`3e-4`. Copying the SFT rate into a GRPO recipe would be an outlier by more than an order of magnitude.
- **LoRA `r=16`/`alpha=32`** — matching the GRPO templates (`qwen3.5-9b-grpo`, `r1-distill-qwen-1.5b-grpo`), not the SFT siblings' `r=8`/`alpha=16`.
- **`max_length: 4096`** — matching the GRPO templates, not the SFT siblings' `2048`. Reasoning completions run long.

## Ratchets covered

- `RecipeMeta` entries with tags following the SFT siblings plus `grpo`/`reasoning`.
- Both added to `EXPECTED_QWEN35_TEXT_RECIPES` (#427's explicit-modality contract) — the family audit fails otherwise.
- Model id pinned on two independent surfaces (`RecipeMeta.model` and YAML `base:`), each by its own test, mirroring `TestDeepSeekV4FlashDpoRecipe`.
- Catalog count bumped 165 → 167 at every site: `catalog.py`'s own comment, all four `test_catalog_size_is_N` milestone checkpoints (renamed to `_167`, new history line added), and all four documentation sites `test_recipe_count_is_synced.py` scans for (`CONTRIBUTING.md`, `docs/commands.md`, `docs/serving-and-export.md` ×2).
- `scripts/generate_recipe_snapshot.py` run last: purely additive diff (22 insertions, 0 deletions) against the delta-encoded fixture.
- `soup recipes show` / `use` run live for both — output below.

## Live output

```
$ soup recipes show qwen3.5-0.8b-grpo
┌─ qwen3.5-0.8b-grpo -- Qwen 3.5 0.8B GRPO reasoning training (Apache-2.0, tiny / mobile) ─┐
│ base: Qwen/Qwen3.5-0.8B                                                                  │
│ task: grpo
│ ...
│ training:
│   lr: 1e-5
│   lora: {r: 16, alpha: 32}
│   grpo_beta: 0.1
│   num_generations: 4
│   reward_fn: accuracy
└───────────────────────────────────────────────────────────────────────────────────────────┘

$ soup recipes use qwen3.5-0.8b-grpo --yes
✓ Recipe qwen3.5-0.8b-grpo written to soup.yaml
Next: soup train --config soup.yaml
```

Same shape for `qwen3.5-2b-grpo`, base swapped to `Qwen/Qwen3.5-2B`.

## Verification

- 678 tests pass across the affected files (`test_recipes.py`, `test_issue427_qwen35_text_modality.py`, `test_recipe_count_is_synced.py`, `test_v07124/07130/07132.py`, `test_issue621_recipe_snapshot.py`), 1 pre-existing unrelated skip.
- `ruff check` and `mypy` both clean.
- **Mutation**: broke `RecipeMeta.model` on one recipe — caught by name by `test_recipe_meta_pins_the_model_id` alone. Reverted, then broke the YAML `base:` line instead — caught by three further tests (the YAML pin, the shape test, and the live show/use test), none of which depend on `RecipeMeta.model`. Reverted after confirming both.
- Live HF Hub check via `huggingface_hub.model_info` for both base repos (not an HTTP ping) — output above.

## A note on mutations, per the issue's own instruction

`grpo_beta: 0.1`, `num_generations: 4`, `reward_fn: accuracy`, `epochs: 3`, `batch_size: auto` all equal their schema defaults — verified directly against `TrainingConfig()`, not assumed. Deleting any of them from the YAML is an equivalent mutation. Reporting them as survivors rather than treating it as a test gap, since it's inherent to a pinned value happening to match the default.

## Not run here

No GPU smoke run — optional per the issue, not required for merge. Happy to add one if useful, but didn't want to present a few local CPU-only steps as a tuned recommendation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,7 +123,7 @@ src/soup_cli/
   experiment/         - SQLite experiment tracking
   eval/               - Eval platform (custom tasks, LLM judge, human eval, leaderboard)
   migrate/            - Config migration (LLaMA-Factory, Axolotl, Unsloth)
-  recipes/            - Ready-made configs for popular models (165 recipes)
+  recipes/            - Ready-made configs for popular models (167 recipes)
   autopilot/          - Zero-config decision engine (v0.25.0)
   registry/           - Model Registry (hashing, store, diff, attach) (v0.26.0 + v0.33.0)
   cans/               - Shareable .can artifact format + run/publish orchestrator (v0.26.0 + v0.33.0)
--- a/changelog.d/0.74.0/864.added.md
+++ b/changelog.d/0.74.0/864.added.md
@@ -1,0 +1,7 @@
+- **Added the `qwen3.5-0.8b-grpo` and `qwen3.5-2b-grpo` recipes (#848 by @SID-6921 in
+  #864).**
+  `Qwen/Qwen3.5-0.8B` and `Qwen/Qwen3.5-2B` previously shipped SFT only; both models are
+  small enough that a contributor might actually run a few steps of GRPO reasoning training
+  on their own hardware. Learning rate (`1e-5`), LoRA rank (`r=16`/`alpha=32`), and
+  `max_length: 4096` follow the existing GRPO templates rather than the SFT siblings'
+  values, stated explicitly rather than inherited silently. Catalog count: 165 -> 167.
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -152,7 +152,7 @@ soup migrate --from llamafactory config.yaml  Import config from LLaMA-Factory
 soup migrate --from axolotl config.yml        Import config from Axolotl
 soup migrate --from unsloth notebook.ipynb    Import config from Unsloth notebook
 soup migrate --from llamafactory c.yaml --dry-run  Preview without writing
-soup recipes list                             List all 165 ready-made recipes
+soup recipes list                             List all 167 ready-made recipes
 soup recipes show llama3.1-8b-sft            Print recipe YAML
 soup recipes use llama3.1-8b-sft             Copy recipe to soup.yaml
 soup recipes search "reasoning"              Search by keyword/task/size
--- a/docs/serving-and-export.md
+++ b/docs/serving-and-export.md
@@ -546,7 +546,7 @@ soup ui
 
 **Pages:**
 - **Dashboard** — view all experiment runs, loss charts, system info, multi-run comparison
-- **New Training** — create configs from templates or 165 ready-made recipes, validate, start training with live SSE log streaming and progress bar
+- **New Training** — create configs from templates or 167 ready-made recipes, validate, start training with live SSE log streaming and progress bar
 - **Data Explorer** — browse and inspect datasets (JSONL, JSON, CSV, Parquet)
 - **Model Chat** — chat with streaming responses, configurable temperature/top_p/max_tokens, system prompt, adapter selection, markdown rendering, chat export
 
@@ -555,7 +555,7 @@ soup ui
 - **Enhanced Metrics** — 2x2 chart grid (loss, LR, grad_norm, throughput) + GPU memory chart, eval results table
 - **Multi-Run Compare** — overlay loss curves from up to 5 runs side-by-side
 - **Chat Upgrade** — SSE streaming via proxy, typing indicator, cancel button, markdown renderer (bold, italic, code blocks), chat export as JSON
-- **Config Builder** — recipe dropdown (165 recipes), config schema API for dynamic form generation
+- **Config Builder** — recipe dropdown (167 recipes), config schema API for dynamic form generation
 
 **Security:** The Web UI generates a random auth token at startup (printed to console). Every private endpoint — mutating (start/stop training, delete runs, inspect data, validate config) and reading (runs, metrics, system, recipes, SSE streams) — requires an `Authorization: Bearer <token>` header. `/` and `/api/health` stay open so the dashboard can load. CORS is restricted to the served origin. Data inspection is sandboxed to the working directory.
 
--- a/src/soup_cli/recipes/catalog.py
+++ b/src/soup_cli/recipes/catalog.py
@@ -48,7 +48,7 @@ def search_recipes(
 
 
 # ---------------------------------------------------------------------------
-# Recipe catalog (165 recipes)
+# Recipe catalog (167 recipes)
 # ---------------------------------------------------------------------------
 
 RECIPES: Dict[str, RecipeMeta] = {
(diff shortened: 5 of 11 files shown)


MakazhanAlpamys:
@SID-6921 — a heads-up while I finish reviewing this, because it is time-sensitive and not caused by anything you did.

**#872 by @Areen-09 adds two different recipes and makes the byte-identical `165 → 167` edit at the same nine sites this PR touches.** Git auto-merges the two with no conflict, and the result claims **167** while `len(RECIPES)` is **169** — which is exactly the #664/#662 incident written up in `CONTRIBUTORS.md`. `tests/test_recipe_count_is_synced.py` catches it, but only on `main`, after the bad merge.

So whichever of the two merges second needs a rebase taking the count to **169** at all nine locations, not 167. I have posted the same note on #872; neither of you is ahead of the other on merit and I am not going to decide it by who pushes faster on a technicality. #872 currently has one required change outstanding (a changelog fragment named after the issue instead of the PR), and your review is in progress.

Nothing to do yet — I will tell you which side of it you land on. Your #869 merged as `aaf6a128`, and there is a follow-up from that review you may want: **#878**, the MCP half of the same defect, where `soup mcp serve`'s `data_validate` still accepts any format string and reports every row valid.

MakazhanAlpamys:
Merged as `f454bfe4`. Thank you, @SID-6921 — second merge today, and no changes were required on this one.

**On the ordering I flagged earlier:** #872 still has an outstanding required change (its changelog fragment is named after the issue rather than the PR), and this PR had none, so it went first. That is a decision on readiness, not on speed — I said I would not settle it on a technicality and I did not.

**The part worth naming is the defence, not the recipes.** Your literal repo id pinned on both surfaces is load-bearing, and I proved it rather than assuming. Mutating `Qwen/Qwen3.5-0.8B` on *one* surface produces failures from the catalog-wide cross-checks. Mutating it **consistently on both** — so `RecipeMeta.model` and the inline `base:` are equally wrong and still agree with each other — leaves every cross-field invariant silent, and only your two pins and the snapshot fire. And since the snapshot is regenerated by the author, an id that was wrong from the very first commit would be baked into it and caught by nothing there. That makes your hand-written literal the only independent human statement of the expected repo. This is exactly the defect that shipped `glm-5` pointing at `THUDM` instead of `zai-org`.

I verified both ids live rather than trusting the PR, using the same anonymous call the repo-id script makes: `Qwen/Qwen3.5-0.8B` and `Qwen/Qwen3.5-2B` both resolve, neither gated nor private, `model_type=qwen3_5`. That was worth doing here because `recipe-repo-ids.yml` is schedule-only with no `pull_request` trigger by design, and the in-repo guard test uses a fake Hub client — so on a PR there is otherwise no live check at all.

**You also stated the three values #848 asked to be argued rather than inherited**, and `lr: 1e-5` is the one that mattered: copying the SFT siblings' 2e-4/3e-4 would have been an order of magnitude out on a GRPO objective, and 16 of the 23 existing GRPO recipes sit at 1e-5. Building both YAMLs as byte-identical to `qwen3.5-9b-grpo` except the `base:` line is the right conservatism for a recipe nobody can run end to end.

**What I ran.** 678 passed, 1 skipped across the seven touched files, identical under `FORCE_COLOR=1`; 2663 passed across the full catalog-touching set; ruff clean. Re-running `scripts/generate_recipe_snapshot.py` leaves `git diff tests/fixtures/` empty, so the fixture was regenerated and not hand-edited. 24 mutations: 18 killed — every hyperparameter #848 names, both entry deletions, and `task` on each surface separately — and 6 survivors, all equivalent by construction. You pre-declared five of those six, which you are unusually well placed to get exactly right, having authored the delta-encoded snapshot guard and the decision that that direction must *not* be killed.

The sixth is a small completeness note, and #848 missed it too so it is not you deviating from the brief: `TrainingConfig.quantization` defaults to `'4bit'`, so `quantization: 4bit` is also equivalent-by-construction — measured, its deletion survives.

One optional nit for next time: the credit parenthetical in `864.added.md` line-wraps as `(#848 by @SID-6921 in\n  #864)`. It renders identically and no test enforces it, but my release-time attribution recovery greps single lines for that form, so an unbroken line is easier on the tooling. Both sibling recipe fragments keep it on one line.

`main` is now at **167** recipes and `tests/test_recipe_count_is_synced.py` passes. Your `CONTRIBUTORS.md` entry is in `2a97512a`.

PR #867, opened 2026-09-11
fix(serve): stop the vLLM backend from sending the engine a doubled BOS (#785 part 2)

abdulwaarith0:
fix(serve): stop the vLLM backend from sending the engine a doubled BOS (#785 part 2)

Part 2 of #785. Part 1 (#788) covers training; #782 covered every path that
tokenizes in process. This is the in-engine backends, which hand the engine a
prompt string and let the engine tokenize it.

## The question, answered by running the engines

Does the engine's own tokenizer add a second BOS on top of the one the chat
template rendered? Measured on an RTX 3060 Laptop (6 GB, WSL2) with
`unsloth/Llama-3.2-1B-Instruct`, whose Llama-3 template renders
`{{- bos_token }}` and whose `bos_token_id` is 128000. One request,
`messages=[{"role": "user", "content": "Hi"}]`. For reference,
`apply_chat_template(tokenize=True)` gives 36 ids starting `[128000, 128006]`.

| engine | what Soup sends | engine `prompt_token_ids[:2]` | ids | verdict |
|---|---|---|---|---|
| vLLM 0.29.0 | rendered string (`build_chat_prompt`) | `[128000, 128000]` | 37 | doubled |
| vLLM 0.29.0 | `{"prompt_token_ids": [...]}` (this PR) | `[128000, 128006]` | 36 | single, equals HF |
| DeepSpeed-MII 0.3.3 | rendered string, via MII's own `HFTokenizer.encode` | `[128000, 128000]` | 37 | doubled |
| SGLang 0.5.9 | rendered string, via SGLang's own `TokenizerManager` tokenize call | `[128000, 128000]` | 37 | doubled |

vLLM numbers come from `RequestOutput.prompt_token_ids` on a real
`vllm.LLM` (the "before" row from `build_chat_prompt`, the "after" row from
the new `build_engine_prompt`, both through the real engine).

MII numbers come from executing `mii.modeling.tokenizers.HFTokenizer.encode`
on the real tokenizer. That is the exact call `MIIPipeline._put_request` makes
for every request (`self.tokenizer.encode(input)` ->
`tokenizer.encode(input, return_tensors="pt")`, default
`add_special_tokens=True`). The full MII engine could not be started on this
box: DeepSpeed's inference ops need a CUDA toolkit to JIT and there is no
`nvcc` here.

SGLang numbers come from executing the call `TokenizerManager._tokenize_texts`
makes for a text request (`self.tokenizer(tokenizer_input, **{})`, default
`add_special_tokens=True`) with SGLang's own `get_tokenizer()` on the real
model. The full SGLang engine could not be started here either: three
attempts (default, `disable_cuda_graph`, `attention_backend=triton` +
`sampling_backend=pytorch`) each stopped on a component needing `nvcc`.

## The fix (vLLM)

`build_engine_prompt(messages, tokenizer)` in `utils/vllm.py` renders the
template exactly as `build_chat_prompt` does, then encodes with
`add_special_tokens=False` through #782's `encode_rendered_prompt`, and
returns `(text, ids)`. `ids` is `None` when no template rendered the prompt
(no tokenizer, no template, or a template that failed to render), so those
cases keep sending the legacy string and the engine keeps tokenizing it
exactly as before.

The vLLM backend now hands `engine.generate` a `{"prompt_token_ids": ids}`
mapping (vLLM's `TokensPrompt` is a plain TypedDict, so no new import) on both
the streaming and non-streaming routes. `usage.prompt_tokens` drops by one on
templates that render BOS, as it did in #782 for the in-process paths.

## SGLang and MII: measured doubled, fix routes checked, split out

#785 offers two fix routes per engine: hand the engine token ids, or its
"no special tokens" option. Both were checked against the installed versions.

**SGLang 0.5.9**
- Token ids: `TokenizerManager._tokenize_one_request` uses `obj.input_ids`
  verbatim when a request carries them, so the ids route exists at the
  `/generate` endpoint. But `sgl.Runtime.generate`, which `utils/sglang.py`
  calls, has no `input_ids` parameter; it posts `{"text": prompt}`. Reaching
  the route means posting to `runtime.url + "/generate"` directly or moving to
  `sgl.Engine` (whose `generate` does take `input_ids`).
- No special tokens: `GenerateReqInput` has no `add_special_tokens` field,
  `ServerArgs` has no such flag, and the text path calls
  `self.tokenizer(tokenizer_input, **{})`. There is no such option.

**DeepSpeed-MII 0.3.3**
- Token ids: `MIIPipeline.__call__(prompts: str | list[str], **generate_kwargs)`
  takes strings only; `GenerateParamsConfig` has no tokenization field.
- No special tokens: `HFTokenizer.encode` hardcodes
  `tokenizer.encode(input, return_tensors="pt")`. `ModelConfig.tokenizer` is
  typed `str | MIITokenizerWrapper`, which looks like the lever, but executing
  it shows `load_tokenizer` re-wraps whatever it is given in `HFTokenizer`, so a
  wrapper overriding `encode(input)` gets called as
  `encode(input, return_tensors="pt")` and raises
  `TypeError: unexpected keyword argument 'return_tensors'`. The route is not
  usable without patching MII's own tokenizer object.

Neither engine could be started on this machine (no CUDA toolkit; vLLM ran
once its flashinfer sampler was switched off, SGLang and MII have no such
switch for the components that need `nvcc`), so a fix for either could not be
verified end to end here the way the vLLM one was. Rather than ship them
unmeasured, this PR records the measurements and the fix routes, and leaves
each engine to a follow-up; the maintainer's "split if hardware is the
blocker" applied one level deeper. If that deferral is acceptable, I can open
one tracking issue per engine carrying the numbers and the route above
(SGLang: post `input_ids` to `/generate` or move to `sgl.Engine`; MII: needs
an upstream change or a patched tokenizer object), so #785 can close on this
PR and #788 with the engine work tracked where someone with the hardware can
pick it up.

## Tests

`tests/test_issue785_engine_bos.py`, in the shape of
`test_issue781_inference_bos.py`: real fast tokenizers built offline from an
in-memory vocab, a template that renders `{{ bos_token }}`, a post-processor
that prepends BOS, and a control proving the fixture really doubles when the
string is tokenized. 17 tests; 12 fail on `main` (`build_engine_prompt` does
not exist, and the vLLM app hands the engine a string).

`test_issue332_vllm_prompt.py::test_engine_receives_the_templated_prompt`
is updated to assert the templated prompt's ids rather than its text; the #332
invariant (the model's own template, not the hand-rolled prompt) is unchanged.

Mutations, each run against this file plus `test_issue332_vllm_prompt.py`
with the function kept but broken (all killed): encoding the rendered text with
the tokenizer's default `add_special_tokens=True` fails 7; the app ignoring the
ids and sending the string fails 5; inverting the `templated` guard (ids for
the legacy fallback, string for the template) fails 15; never returning `None`
so the legacy fallback loses the engine's BOS fails 6; only the streaming route
sending the string fails 2; only the non-streaming route fails 3; dropping the
leading id from what the engine gets fails 5; raising on a broken template
instead of falling back fails 1.

`ruff check` clean. Full `pytest`: 20978 passed, 61 skipped; the only failures
are `test_issue385_stream_dtype.py::TestFloat16StreamingIsBitExact::*[nf4]`,
which fail identically on a clean `main` checkout on this GPU (nf4 stream vs
resident differ by 4.9e-4 / 3.9e-3, unrelated to this change).

### DIFF
--- a/changelog.d/0.74.0/867.fixed.md
+++ b/changelog.d/0.74.0/867.fixed.md
@@ -1,0 +1,15 @@
+- **`soup serve --backend vllm` no longer sends the engine a doubled BOS (#785 part 2 by @abdulwaarith0 in #867).**
+  #782 fixed every path that tokenizes in process, and left the engines that
+  tokenize the prompt string themselves for someone with the hardware to check.
+  Measured on vLLM 0.29.0 with `unsloth/Llama-3.2-1B-Instruct`, whose Llama-3
+  template renders `{{ bos_token }}` (id 128000): for the string Soup sent, the
+  engine reported `prompt_token_ids[:2] == [128000, 128000]` over 37 ids, against
+  the 36 ids `apply_chat_template(tokenize=True)` returns. The vLLM backend now
+  hands the engine `{"prompt_token_ids": [...]}` built with no tokenizer special
+  tokens whenever the chat template rendered the prompt, which reproduced the
+  36-id encoding exactly (`[128000, 128006]`), so the model is prompted with the
+  prefix Soup trains on. `usage.prompt_tokens` drops by one on such templates.
+  Models without a chat template, and templates that fail to render, still send
+  the legacy string and are tokenized by the engine exactly as before. The
+  SGLang and DeepSpeed-MII backends were measured through their own tokenizer
+  paths and are tracked separately in #785.
--- a/changelog.d/0.75.0/867.fixed.md
+++ b/changelog.d/0.75.0/867.fixed.md
@@ -9,6 +9,10 @@
   tokens whenever the chat template rendered the prompt, which reproduced the
   36-id encoding exactly (`[128000, 128006]`), so the model is prompted with the
   prefix Soup trains on. `usage.prompt_tokens` drops by one on such templates.
+  The change is wider than removing a duplicate: for a model whose chat template
+  renders no BOS but whose tokenizer post-processor would add one, the engine
+  previously received one BOS and now receives zero, matching Soup's training
+  path and the in-process fix #782 already shipped for the transformers backend.
   Models without a chat template, and templates that fail to render, still send
   the legacy string and are tokenized by the engine exactly as before. The
   SGLang and DeepSpeed-MII backends were measured through their own tokenizer
--- a/docs/serving-and-export.md
+++ b/docs/serving-and-export.md
@@ -281,9 +281,11 @@ soup serve --model ./output --backend vllm --max-model-len 8192
 > **Tip:** Soup auto-detects vLLM. When installed, you'll see a hint during `soup serve` if you haven't enabled it yet.
 
 The vLLM backend applies the **model's own chat template**, exactly like the
-transformers backend — both call one shared prompt builder. A model that ships
-no chat template falls back to a generic `User:` / `Assistant:` prompt, and the
-server says so at startup. `finish_reason` reports `"length"` when a response
+transformers backend, and encodes the rendered prompt itself so the engine
+receives the same token ids Soup trains on rather than re-tokenizing the string.
+(The SGLang and MII backends still hand the engine the rendered string; see
+#785.) A model that ships no chat template falls back to a generic `User:` /
+`Assistant:` prompt, and the server says so at startup. `finish_reason` reports `"length"` when a response
 hits `max_tokens` and `"stop"` otherwise (`/v1/messages` maps those to
 `max_tokens` / `end_turn`).
 
--- a/src/soup_cli/utils/vllm.py
+++ b/src/soup_cli/utils/vllm.py
@@ -101,8 +101,8 @@ def build_chat_prompt(messages, tokenizer=None) -> str:
     format only when it does not (or when no tokenizer could be loaded).
 
     In-process callers must not hand the result to a bare ``tokenizer(...)``
-    call; use :func:`encode_chat_prompt` (#781). The vLLM engine tokenizes this
-    string itself.
+    call; use :func:`encode_chat_prompt` (#781). Backends whose engine would
+    tokenize this string itself use :func:`build_engine_prompt` (#785).
 
     Args:
         messages: chat messages — pydantic objects or dicts.
@@ -144,6 +144,37 @@ def encode_chat_prompt(
         messages, tokenizer, fallback_on_error=fallback_on_error
     )
     return encode_rendered_prompt(tokenizer, text, templated=templated, **tokenizer_kwargs)
+
+
+def build_engine_prompt(
+    messages: Any, tokenizer: Any = None
+) -> tuple[str, Optional[list[int]]]:
+    """Render chat messages for a backend that tokenizes the prompt itself (#785).
+
+    The vLLM, SGLang and MII backends hand the engine a prompt STRING, and the
+    engine encodes it with its tokenizer's default ``add_special_tokens=True``.
+    A template that renders ``{{ bos_token }}`` therefore reached the model with
+    two of them. Measured on vLLM 0.29.0 + ``unsloth/Llama-3.2-1B-Instruct``
+    (``bos_token_id`` 128000): for the string Soup sends today the engine
+    reported ``prompt_token_ids[:2] == [128000, 128000]`` over 37 ids, against
+    the 36 ids ``apply_chat_template(tokenize=True)`` returns. Handing the
+    engine these ids instead reproduced that 36-id encoding exactly.
+
+    #782 fixed the same defect for every path that tokenizes in process; this is
+    the same rule for the paths that do not, so the chat template stays the
+    single source of the model's special tokens.
+
+    Returns:
+        ``(text, token_ids)``. ``token_ids`` is what to send the engine in place
+        of ``text``, and is None when no template rendered the prompt: the
+        legacy role-prefixed fallback carries no special tokens of its own, so
+        the engine must keep adding its own exactly as it always has.
+    """
+    text, templated = _render_chat_prompt(messages, tokenizer, fallback_on_error=True)
+    if not templated:
+        return text, None
+    encoded = encode_rendered_prompt(tokenizer, text, templated=True)
+    return text, list(encoded["input_ids"])
 
 
 def resolve_finish_reason(output: Any, max_tokens: Optional[int]) -> str:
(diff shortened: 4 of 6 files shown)


MakazhanAlpamys:
@abdulwaarith0 — reviewed at `8cb49550`, 13/13 required checks green. **This is merge-ready**, and I want to name what makes it good before the three small asks, none of which are blockers in the usual sense.

## What is right about the design

**The fix does not de-duplicate anything**, and that is the correct call. Instead of comparing the first token, or matching a string prefix, or reasoning about what the template emitted, you bypass engine tokenization entirely by handing vLLM `{"prompt_token_ids": ids}` encoded with `add_special_tokens=False`. Nothing about that is fragile, and it cannot strip a BOS a model was legitimately trained to see twice — whatever the template renders is what the engine gets.

The verification is the other half. You measured `prompt_token_ids[:2]` on all three engines rather than fixing the one you suspected, and when SGLang and MII turned out to need a different fix route, you documented the route you checked instead of guessing. `test_a_tokenizer_with_no_bos_at_all_is_untouched` with `post_processor=None` is a real control: it asserts the fix moves not a single id, which is what separates a fix from a change.

**What I ran.** 51 passed across the two touched files. Reverting `src/soup_cli/utils/vllm.py` to `main` produces 13 failures — 12 in the new file, 1 in the #332 file — so the suite is load-bearing. 578 passed, 14 skipped across every file that imports or exercises `utils/vllm.py`. Eight mutations applied in isolation, **all eight killed**: always returning the string; inverting the condition; gating on "the tokenizer has a template" instead of "the template rendered this text"; dropping the leading id; encoding with the tokenizer's own specials; the app ignoring the ids; and fixing only the streaming or only the non-streaming route. The thinnest margin is the template-gate mutation, killed by exactly one test (`test_a_broken_template_falls_back_to_the_legacy_string`) — worth knowing, not worth changing.

Also checked: no test deleted, skipped or loosened. The one changed assertion in `test_issue332_vllm_prompt.py` is **strengthened**, from comparing the templated text to asserting the exact ids of that same text, so the #332 invariant survives. And this PR stands alone — zero file overlap with your #788, which touches only `data/*`; `encode_rendered_prompt` is already on `main` from the merged #782.

## Three asks

**1. The changelog fragment understates the change, and this is the one I would like fixed before merge.** The fix is wider than removing a duplicate: for a model whose template renders *no* BOS but whose tokenizer post-processor adds one, vLLM previously received **one** and now receives **zero**. You clearly know this — `test_a_soup_preset_is_prompted_with_no_bos_at_all` asserts it deliberately — and I agree with the choice, because it matches Soup's training path and HF's own `apply_chat_template(tokenize=True)`, and #782 already shipped exactly this for the transformers backend. But the fragment says "no longer sends a doubled BOS" and "models without a chat template … exactly as before", which a reader will take as duplicate-removal only. One sentence naming the zero-BOS case.

**2. A question I could not answer, so I am asking rather than guessing.** `pyproject.toml:135` declares `vllm>=0.4.0`. You measured on 0.29.0. The dict/`TokensPrompt` input form landed partway through the 0.4.x line; before that, `AsyncLLMEngine.generate` took a prompt string plus a separate `prompt_token_ids=` kwarg. vLLM is not installed on my box and I will not assert a cutover version from memory — **what is the minimum vLLM version you believe accepts a mapping as the first positional argument?** If it is above 0.4.0, the extra's floor is now wrong and that is a separate small PR, not yours to carry here.

**3. Optional, and precedent says you may decline.** `docs/serving-and-export.md:283-284` says the vLLM backend "applies the model's own chat template, exactly like the transformers backend — both call one shared prompt builder". That is now subtly stale: vLLM calls `build_engine_prompt` while SGLang and MII still call `build_chat_prompt`. #782, the merged sibling fix of the same defect, shipped with a fragment and no docs change, so one sentence here would be an improvement rather than a requirement.

## On closing #785

#785's acceptance item 5 asks for the measurement on each engine "and a fix where doubled", with an explicit escape hatch if hardware blocks it. You have met the measurement for all three and fixed one, so **#785 stays open** after this and #788. I would like to take your offer of one tracking issue per engine — SGLang and MII — each carrying the measured `[128000, 128000]` and the fix route you checked. Open them whenever suits you; they do not gate this merge.

One note for awareness, not a change request: `build_engine_prompt` calls the tokenizer where `build_chat_prompt` never did, and `_render_chat_prompt(fallback_on_error=True)` deliberately swallows a broken template so a served request does not 500 — the new encode step has no equivalent guard, so a tokenizer that raises now propagates. This is **not a regression**: `encode_chat_prompt` in `commands/serve.py` has the identical unguarded encode from #782. I am recording it so the "a broken template must not 500" policy is known to have a second unguarded point.

Do item 1 and answer item 2 and I merge on green.

MakazhanAlpamys:

MakazhanAlpamys:

```
```

```
```

abdulwaarith0:
Thanks for the read, and for separating what is mine to carry here from what is not.

Item 1, done. The fragment now names the zero-BOS case: for a template that renders no BOS on a tokenizer whose post-processor would add one, the engine went from one BOS to zero, same as #782 for the transformers path.

Item 2, the minimum vLLM version is 0.4.3. I checked the published docs rather than trusting the number I measured on: 0.4.2's `generate` is still `generate(prompt: str | None, ..., prompt_token_ids=None)`, so a mapping is not accepted there; 0.4.3 is the first release where the first positional argument is `inputs: str | TextPrompt | TokensPrompt | TextTokensPrompt` and the separate `prompt_token_ids` kwarg is gone. So `vllm>=0.4.0` is too low and the floor should be `>=0.4.3`. I will send that as its own small PR against the extra rather than fold it in here.

Item 3, taken. The docs sentence now says the vLLM backend encodes the rendered prompt itself and hands the engine token ids, and notes SGLang and MII still pass the string, so it no longer implies one shared builder.

I have also rebased onto master and moved the fragment to `changelog.d/0.75.0/` so the 487 baseline check passes.

I will open one tracking issue per engine (SGLang, MII) carrying the measured `[128000, 128000]` and the fix route, so #785 can close on this and #788 with the engine work tracked where the hardware is.

MakazhanAlpamys:
Merged as c083144, right behind #788.

**Verified on `8e7bf94e`:** your two touched files **51 passed**; blast radius across 11 files including `test_issue360_sglang_prompt_finish_reason.py` and `test_mii_serve.py` **617 passed, 14 skipped**; under `FORCE_COLOR=1` **68 passed**; ruff clean; 14/14 required checks on the exact head. Your new tests against `origin/main`'s source: **12 failed, 5 passed**. **Core mutation killed** — reverting `build_engine_prompt` to `return text, None` (main's behaviour, hand the engine the string) fails **9 tests**, across both the new file and the pre-existing #332 suite, in both the streaming and non-streaming parametrisations.

**Three things you did that I want to name:**

1. **The fragment names the opposite case.** "For a model whose chat template renders no BOS but whose tokenizer post-processor would add one, the engine previously received one BOS and now receives zero." That is a change nobody would have caught in review and that you were under no obligation to volunteer. It is also the sentence that turns this from a no-op-looking fix into a documented behaviour change.
2. **You answered the vLLM version question with a signature, not a guess.** 0.4.3, because that is where the first parameter became `inputs: str | TextPrompt | TokensPrompt | TextTokensPrompt` and `vllm.inputs` first appears, while 0.4.2 still takes `prompt` plus a separate `prompt_token_ids` keyword. Then you sent it as #892 rather than widening this PR — which is why I could merge this one on green checks and the floor one minute later.
3. **#890 and #891 exist with measurements in them.** You offered one tracking issue per engine and then filed both carrying the measured `[128000, 128000]` and the fix route checked against sglang 0.5.9 and MII 0.3.3. An issue with a measurement in it is worth several without.

The one changed assertion is strengthened: a templated-text comparison became an exact-ids comparison of that same text, so #332's invariant survives in a stricter form.

`docs/serving-and-export.md` now correctly says vLLM encodes the prompt itself and hands the engine token ids, while SGLang and MII still hand it the rendered string. That sentence is what will stop the next reader assuming all three backends behave alike.

PR #868, opened 2026-09-11
fix: honor training settings across PRM, MoLE, PPO and unlearning

### LINKED ISSUE #805: prm, moe_lora_routing, unlearn and ppo drop the optimizer and schedule: the default config's cosine/warmup/weight-decay never reach them, and batch_size: auto becomes 1
## The defect

Four trainers build their optimizer settings without most of `TrainingConfig`. Because the schema defaults are not the HF/TRL defaults (`optimizer=adamw_torch`, `scheduler=cosine`, `warmup_ratio=0.03`, `weight_decay=0.01`, `save_steps=100`, `batch_size=auto`), this changes even a config that sets none of them.

### prm and moe_lora_routing

`trainer/prm.py:323-338` and `trainer/mole_routing.py:361-374` construct `TrainingArguments` with epochs, batch, accumulation, lr, logging/save steps, and nothing else. Omitted: `optim`, `lr_scheduler_type`, `warmup_steps`, `weight_decay`, `max_grad_norm`, `gradient_checkpointing`; MoLE also omits `bf16`/`fp16` (PRM passes them). With transformers 5.17 the omitted values fall back to:

```

optim = adamw_torch_fused   lr_scheduler_type = linear   weight_decay = 0.0

max_grad_norm = 1.0         warmup_steps = 0             gradient_checkpointing = False

```

`batch_size: auto`, the schema default, is not estimated as in the other trainers; it silently becomes 1 (`prm.py:318-321`, `mole_routing.py:357-360`):

```python

if isinstance(tcfg.batch_size, bool) or not isinstance(tcfg.batch_size, int):

    bs = 1

```

### unlearn

`trainer/unlearn.py` has its own loop. `setup()` (`:222-224`) builds `torch.optim.AdamW(params, lr=lr)` with torch's `weight_decay=0.01` whatever the config says, and no scheduler, warmup, clipping or optimizer choice. `train()` (`:277-305`) takes one forget example per optimizer step, so `batch_size` and `gradient_accumulation_steps` (default 4) are ignored. Sequences are cut at a module constant `_MAX_LENGTH = 256` (`:39`), not `data.max_length` (default 2048). The base loads through `live_eval.load_model_and_tokenizer` without `quantization` (default `4bit`), and the LoRA config (`:184-190`) passes only `r`, `alpha`, `dropout`, so `target_modules`, `use_dora`, `use_rslora` and `rank_pattern` are ignored (PEFT picks its per-architecture default targets).

### ppo

`trainer/ppo.py:157-163` starts `ppo_kwargs` with output dir, batch, accumulation, lr and seed; the capability probes after it add PPO-specific fields only. trl 0.29.1's `trl.experimental.ppo.PPOConfig` then supplies:

```

optim = adamw_torch_fused   lr_scheduler_type = linear   weight_decay = 0.0

max_grad_norm = 1.0         warmup_steps = 0             save_steps = 500   logging_steps = 10

```

so the default config's `save_steps: 100` becomes 500 and `cosine` becomes `linear`.

**Coordination:** #721 (open, PRs #737 and #741) covers PPO's `kl_coef` and `num_train_epochs`. Neither PR touches the fields above; I checked both diffs. They are additional and could land in whichever PR wins there, or after it.

## Reproduction

All accepted, current `main`:

```

ACCEPT prm batch_size= auto optimizer= adafactor scheduler= cosine wd= 0.1 mgn= 0.3 dora= True

ACCEPT moe_lora_routing batch_size= auto optimizer= adafactor scheduler= cosine wd= 0.1 mgn= 0.3 dora= True

ACCEPT unlearn batch_size= auto optimizer= adafactor scheduler= cosine wd= 0.1 mgn= 0.3 dora= True

```

None of those values reaches training on these three tasks.

## Why it matters

A PRM trained with the default config gets a linear schedule with no warmup and no weight decay at batch 1, where the same YAML on SFT gets cosine, 3% warmup and decay 0.01 at an estimated batch. Nothing is printed. Unlearning is where step size matters most, since NPO/RMU push the model away from data; there the user cannot clip, warm up or schedule, and their `lora.target_modules` is replaced.

## Proposed fix

- prm / moe_lora_routing: pass the same fields every other HF-Trainer wrapper passes (see `trainer/dpo.py`'s `DPOConfig(...)`), and resolve `batch_size: auto` through `estimate_batch_size` or refuse `auto` at load for these tasks. A shared `training_arguments_kwargs(tcfg, ...)` helper, like `training_seed_kwargs` for #353, keeps the next trainer from repeating this.

- ppo: add the same fields to `ppo_kwargs`, probed by capability like the existing ones.

- unlearn: honour `weight_decay`, `max_grad_norm` (`clip_grad_norm_`), a scheduler with warmup, `gradient_accumulation_steps`, `data.max_length`, `quantization` and the full LoRA config, or reject each unsupported one for `task: unlearn` at load.

## Acceptance criteria

- [ ] For prm and moe_lora_routing, a test builds the trainer on a tiny CPU model with `scheduler: cosine, warmup_ratio: 0.1, weight_decay: 0.1, max_grad_norm: 0.3, optimizer: adafactor` and asserts each value on `trainer.args`. Fails today.

- [ ] `batch_size: auto` on prm / moe_lora_routing is either estimated or refused, not silently 1.

- [ ] PPO: the same assertions against `PPOConfig`, written so a fall-back to trl defaults cannot pass (non-default values).

- [ ] unlearn: a test that a configured `weight_decay` reaches the optimizer's param groups and `max_grad_norm` clips; unsupported fields refused at load.

Related: #745 (`loraplus_lr_ratio`), same shape.

---

- #795 — The default quantization: 4bit is ignored by six trainers (distill, classifier/reranker/cross_encoder, prm,...
- #802 — loss_spike_recovery and grad_accum_auto_tune reach the training callback on SFT only: 15 trainers build it ...

Oxygen56:
Fixes #805

## What does this PR do?

PRM, MoLE routing, PPO and unlearning now honor configured optimizer, scheduler, warmup, weight decay and gradient-clipping settings. PRM, MoLE and unlearning resolve automatic batch sizing through the existing estimator.

Unlearning uses the configured sequence limit and LoRA targeting, while preserving its existing unquantized policy/reference loading and sample-based loss and step reporting. Its single-example forwards accumulate into optimizer updates; the documentation explains the resulting update cadence and sequence-length change.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [ ] `pytest tests/ -v` passes (the complete coverage and platform matrix remains with CI; the relevant local set below passes)

## Root cause

The wrappers constructed backend arguments with only part of `TrainingConfig`, silently replacing accepted settings with backend defaults. The unlearning loop also used fixed optimizer and sequence settings.

PPO uses the existing compatibility helper. Unlearning keeps loss values, reported example counts and the existing sample budget on their original scale, including partial accumulation groups.

## Verification

- Real offline CPU models exercise PRM and MoLE training, PPO configuration/model loading, and SimNPO/NPO training. Loader-call assertions cover both the unlearning policy and frozen reference; they fail before removing the two quantization arguments and pass afterward.
- `PYTHONPATH=src python -m pytest -o addopts= --no-cov -q` over the 15 related files below: **746 passed**. This includes #1043's task-specific configuration rules, README translation synchronization and changelog validation.
- `ruff check src/soup_cli/ scripts/ tests/ benchmarks/` and `git diff --check origin/main` pass.

<details>
<summary>Related test files</summary>

```text
tests/test_issue805_optimizer_settings.py
tests/test_issue795_quantization_unhonoured_tasks.py
tests/test_readme_translation_sync.py
tests/test_issue487_changelog_fragments.py
tests/test_ppo.py
tests/test_v0610_part_a.py
tests/test_v0610_part_b.py
tests/test_v0719.py
tests/test_issue563_mps_bf16_training.py
tests/test_issue803_lora_patterns_all_trainers.py
tests/test_issue804_chat_template_trainers.py
tests/test_trainer_coverage_v035.py
tests/test_issue802_callback_kwargs_in_trainers.py
tests/test_issue899_loss_summary.py
```

</details>

The loss-summary test fixture now initializes the sample budget and optimizer-step dependencies normally supplied by `setup()`; its existing loss assertions are unchanged. PRM fixtures explicitly use `lora.r: 0`, matching the current configuration contract.

The full cross-platform suite, accelerator smoke checks and coverage gate remain with CI. PPO validation stops before its rollout-training loop.


MakazhanAlpamys:
@Oxygen56 — thank you for this, and thank you for going after #805 at all: "an option that is validated, documented, accepted and read by nothing" is this project's most common defect class, and you found four instances of it in one pass. **The fix itself is largely right, and I verified that behaviourally** rather than by reading. But I cannot merge it as it stands. Four things, in the order I would tackle them.

## 0. First, a collision that is not your fault

**#737 merged about an hour ago** as `9ba0b577` and also touches `src/soup_cli/trainer/ppo.py`. Your branch is based on `67de1e5f`, before it. `git merge origin/main` gives **one conflict, in `ppo.py` only** — every other file auto-merges.

The good news is that it is textual, not semantic. The two field sets are disjoint:

| | fields set |
|---|---|
| #737 (landed) | `num_train_epochs`, `num_ppo_epochs`/`ppo_epochs`, `kl_coef`/`init_kl_coef` |
| yours | `warmup_steps`, `weight_decay`, `max_grad_norm`, `optim`, `lr_scheduler_type`, `logging_steps`, `save_steps`, `save_total_limit`, `bf16`, `fp16`, `gradient_checkpointing` |

Your hunk is anchored on the old `if "num_ppo_epochs" in ppo_params:` block that #737 **deleted** and replaced with `_set_ppo_training_kwargs(...)`. Resolution is mechanical: keep your `optional_training_kwargs` block, drop your retained copy of the old epochs branch, keep main's call.

And one thing worth knowing, because it makes your PR *better* on top of #737: you compute `warmup_steps` from `total_steps = ceil(len(train_ds)/bs/accum) * tcfg.epochs`, but before #737 `tcfg.epochs` never reached PPO at all — so your warmup budget was being sized against an epoch count the trainer was not using. After #737 the two finally agree.

## 1. BLOCKING — three existing tests fail, and the documented `unlearn` example now raises

`src/soup_cli/trainer/unlearn.py` raises on `batch_size == "auto"` — **which is the schema default** (`config/schema.py:1026-1029`, `default="auto"`).

| test file | on `main` | on this PR |
|---|---|---|
| `tests/test_v0610_part_a.py` | 51 passed | **2 failed**, 49 passed |
| `tests/test_v0719.py` | 76 passed, 5 skipped | **1 failed**, 75 passed |

Failing: `TestBuildUnlearnTrainer::test_kwargs_signature_allows_known`, `TestUnlearnTrainerWrapper::test_setup_loads_model` (in both files). Each builds a `SoupConfig` without `batch_size` and never reaches the loader.

This is not only a test problem. **`docs/training.md`'s `task: unlearn` example sets no `batch_size`** — I checked — so the example we ship now fails on the default config.

What makes it stand out is that *the same PR* does the right thing twice elsewhere: `prm.py` and `mole_routing.py` resolve `"auto"` through `estimate_batch_size(...)`. Unlearn should do one of those two things — resolve it the same way, or refuse it at config load through a `SoupConfig` cross-validator so it is a parse error rather than a failure minutes into setup. #805's own wording asks for unsupported fields to be refused "at load". Raising inside `setup()` is the one option that is neither.

## 2. BLOCKING — the tests assert spelling, not behaviour

`tests/test_issue805_optimizer_settings.py` is 79 lines of `ast` keyword-*name* checks and `assert "<literal string>" in source`. Full reverts are caught, so it is not one test standing in for four — each trainer does have its own guard. But **7 of 8 value mutations survived**:

| mutation | result |
|---|---|
| full revert of prm.py / mole_routing.py / ppo.py / unlearn.py (4 separate) | all **caught** |
| prm: `weight_decay=tcfg.weight_decay` changed to `0.0` | **survived** |
| mole: `optim=tcfg.optimizer` changed to `"adamw_torch_fused"` | **survived** |
| ppo: `"weight_decay": tcfg.weight_decay` changed to `0.0` | **survived** |
| unlearn: `clip_grad_norm_(..., float(tcfg.max_grad_norm))` changed to `1.0` | **survived** |
| prm: `lr_scheduler_type=tcfg.scheduler` changed to `"linear"` | **survived** |
| unlearn: `get_scheduler(name=tcfg.scheduler)` changed to `"linear"` | **survived** |
| prm: auto-batch estimate computed, then discarded | **survived** |

Every survivor reinstates exactly the bug #805 exists to prevent — the value forwarded under the right keyword, with the wrong value — and the suite stays green. That is the shape v0.73.3 was named for, and #805's fourth acceptance criterion says in as many words that a fall-back to trl defaults must not pass.

I understand a local checkout may lack the training stack. **CI installs the full `[dev]` extra**, so behavioural tests run there whether or not they run for you locally, and I ran them here against `transformers 5.17.0 / torch 2.14.0+cu130 / trl 0.29.1 / peft 0.20.0`. What I need instead of the AST checks: build each trainer on a tiny CPU model with `scheduler: cosine, warmup_ratio: 0.1, weight_decay: 0.1, max_grad_norm: 0.3, optimizer: adafactor` and assert on `trainer.args` / `PPOConfig` / `optimizer.param_groups`. Every mutation in that table must then fail. I confirmed the targets exist on the installed stack — `TrainingArguments` stores all of them, `Trainer.get_optimizer_cls_and_kwargs` really returns `Adafactor`, `weight_decay=0.1` really lands in `optimizer.param_groups`, and trl 0.29.1's `PPOConfig` accepts all eleven fields — so none of this is blocked on upstream.

## 3. The reported unlearn loss silently changes scale

`unlearn.py` now divides by `batch_size * accumulation` **before** `initial_loss` / `final_loss` are recorded, and those go into the result dict and get printed. With the defaults (`gradient_accumulation_steps=4`, `batch_size=2`) the reported loss is **8x smaller** than on `main` for an unchanged run. The gradient scaling is right; the reporting should keep the unscaled value, or every historical unlearn number becomes incomparable with nothing saying so.

Related: `_MAX_LENGTH = 256` becoming `data.max_length` (default 2048) is an 8x sequence-length jump on an existing task. #805 asked for it and I agree with it, but it ships here with no docs note and no test, and it has a real memory and runtime cost.

## 4. Small things

- **Fragment name and credit.** `CONTRIBUTING.md:351-355` asks for the PR number, not the issue: `805.fixed.md` should be **`868.fixed.md`**. And 41 of the 43 fragments in `changelog.d/0.74.0/` carry a `by @` — yours carries none. Format, matching #737 which merged today: `(#805 by @Oxygen56 in #868)`.
- **Scope creep worth splitting.** `**(self.fsdp_config or {})` in `prm.py:365` and `mole_routing.py:406` is a genuine same-shape bug — both wrappers accepted `fsdp_config` and never read it — but it is not #805 and it is untested here. Either drop it, or say the word and I will open it as its own issue crediting you for finding it; I would rather it got a test than rode along.
- **`save_total_limit: 3`** newly hardcoded on PPO — deliberate, or leftover?
- `docs/training.md`'s unlearn section needs the `batch_size` decision from item 1 and the `max_length` change.

## What is good, so this does not read as only a list of problems

**No existing test was deleted or weakened** — the diff on `tests/` is `+79, 0 deletions`. `ruff check` is clean on all five files. Removing the `isinstance(tcfg.batch_size, bool)` guard from prm/mole is safe, and I verified why rather than assuming: pydantic coerces `batch_size=True` to `1` at the schema boundary, so `int(batch_size)` gives the same answer. And the central claim of the PR holds — these four trainers really were dropping the configured optimizer, scheduler, warmup, weight decay and clipping on the floor, and after this they do not.

Items 1 and 2 are what block. 3 and 4 are quick. Take them in any order; ping me when you want another pass and I will re-run the mutation matrix.

One thing I could not verify, so please do not assume I did: I exercised the config and optimizer **construction** boundary only, not a real backward pass. The rewritten unlearn accumulation loop's gradient semantics — the scaling and the trailing partial-batch flush — are unverified here.

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:
@Oxygen56 — a real status check, since my last comment here was a fragment broadcast rather than review.

**Three things, and the first two are new since you last pushed:**

1. **CI is red — 9 failing tests**, and the cause is a genuine one rather than flake: `unlearn.py` raises on `batch_size == "auto"`, which is the schema **default**, and the shipped `docs/training.md` unlearn example sets no `batch_size` at all. So the documented example cannot run.
2. **The branch now conflicts with `main`** (`mergeable=false`) in `trainer/ppo.py`, from #737 landing. It needs a rebase before anything else can be judged.
3. **The blocking item from 2026-09-12 is still open**, and it is the one I care about most: the reported unlearn loss silently changes scale — **8x smaller with defaults** — and `_MAX_LENGTH` goes from 256 to `data.max_length` (2048), an 8x sequence-length jump, with no docs note and no test. A number that changes scale without anyone being told is worse than one that breaks, because nobody investigates it.

Also for the rebase: `trainer/ppo.py`, `trainer/prm.py` and `trainer/mole_routing.py` are shared with **#908**, which touches all 18 trainers. Both of you add a line at the same import anchor, so whoever lands second resolves three add/add conflicts — trivial, but re-run `ruff check` afterwards since import order is enforced. You will be doing that resolution twice if the rebase happens before #908 lands.

Are you still on this? It is real work on four trainers and I do not want it to age out — but it cannot be reviewed further while it is red and conflicting. If you would rather hand it back, say so and I will record the reason.

MakazhanAlpamys:
@Oxygen56 — a full status with everything measured, because you deserve the whole picture in one comment rather than another nudge.

**First, the part that is not in doubt: your diagnosis is right and it is valuable.** Four trainers really did drop optimizer, scheduler, warmup, weight decay and clipping, and the mechanism you added works — I confirmed `Adafactor` is constructed with `weight_decay=0.1` straight from config, with no directory side-effect. That is a real defect across four trainers and I do not want to lose it.

Everything below is what stands between that and a merge.

## CI: four failures, not three — one is new information

```
pytest tests/test_v0610_part_a.py tests/test_v0719.py tests/test_issue563_mps_bf16_training.py
4 failed, 133 passed, 5 skipped
```

The fourth is one neither of us has mentioned. The PR adds `allow_mps_bf16=True` at `mole_routing.py:380` and `ppo.py:187`, which widens an explicitly hardware-validated allowlist:

```
E   Extra items in the left set: 'mole_routing.py', 'ppo.py'
tests/test_issue563_mps_bf16_training.py:123
```

That guard exists to force the decision, not to be edited past. Either produce the hardware validation or drop the two lines.

## `batch_size: "auto"` is a pure regression, and it breaks the shipped example

`docs/training.md:349-360` is a complete unlearn config that sets no `batch_size`. Driven into the real wrapper:

```
File "src/soup_cli/trainer/unlearn.py", line 175, in setup
ValueError: task='unlearn' requires an integer training.batch_size; batch_size='auto' is unsupported
```

`git show origin/main:.../unlearn.py | grep -n batch_size` returns nothing, so this raise is entirely new, and the schema default is `"auto"` (`schema.py:1026-1028`). `ppo.py` on main already shows the shape to copy — `estimate_batch_size`.

## The loss-scale item, with the numbers

`unlearn.py:361-367` divides, then reads the **scaled** value into the reported number:

```python
loss = loss / float(batch_size * accumulation)
loss.backward()
lval = float(loss.item())        # <- the scaled loss
```

Measured on an identical synthetic loss sequence at `lr=0`, the ratio is exactly `batch_size x accumulation`:

| bs | accum | main initial | this head | ratio |
|---|---|---|---|---|
| 1 | 4 | 2.0 | 0.5 | **4x** |
| 2 | 4 | 2.0 | 0.25 | **8x** |
| 8 | 4 | 2.0 | 0.0625 | **32x** |

**A correction to my own earlier figure:** I said "8x with defaults", which assumed `batch_size: 2`. The schema default is `"auto"`, which now refuses, so the minimum *runnable* config is 4x and the true default config cannot run at all.

Two more reported numbers move silently, which I had not flagged: `total_steps` goes **16 to 2** on the same run, because it now counts optimizer updates rather than examples, and it is in the result dict; and `_MAX_STEPS_CAP = 2000` now caps updates, so the budget ceiling rose by the same factor. Separately — `batch_size` in this loop is never a batch, since each example is still its own forward and backward, so it is a second accumulation factor with no memory or throughput effect, which makes the `auto` refusal doubly odd.

`_MAX_LENGTH` was 256 on main and is gone; four sites now use `data.max_length`, default 2048 and 4096 in many shipped recipes — up to 16x, with the schema ceiling at 1048576 and no local cap.

Neither is documented (`git diff origin/main...HEAD --stat -- docs/` is empty) and neither is tested. The one hit is `tests/test_issue805_optimizer_settings.py:75`, `assert "self.config.data.max_length" in source` — a spelling check that cannot tell 256 from 1048576. **Nothing pins the loss scale.**

**The bar for this is already set in-tree.** #742 (`dd151374`, merged one day before my last comment) is the same class of change: it measured the ratio (`4.00000046` at GA=4), ran `Trainer.training_step` on a one-layer Llama, and listed five mutations plus a control. That is what a reported number changing by 4-32x needs.

## The conflict is getting worse, not staying still

```
git merge-tree --write-tree --name-only origin/main HEAD
CONFLICT (content): Merge conflict in src/soup_cli/trainer/ppo.py
```

One file. `_set_ppo_training_kwargs` arrived in #737 and `cd165cf1` (#913) then refactored it onto `_trl_compat`. So your hand-rolled 30-line `optional_training_kwargs` block and its `num_ppo_epochs`/`ppo_epochs` branch are now **dead code duplicating a helper main deliberately centralised** — the correct resolution is to delete them, not to merge them. #908 is still open on the same region, so the double-resolution warning stands.

## The fragment blocks the release, three ways

`changelog.d/0.74.0/805.fixed.md`: stale directory, named after the issue rather than the PR, no credit. Demonstrated on a real merged tree:

```
ChangelogError: fragment baseline is 0.74.0, but the latest release is 0.75.0
```

`test_issue487_changelog_fragments` then fails in every cell, and `publish.yml --check-empty` runs it before `python -m build`. Required: `changelog.d/0.75.0/868.fixed.md` ending `(#805 by @Oxygen56 in #868)`.

## Where I would like to leave this

Nothing here is open-ended. The remaining work is: `auto` to an estimate in the shape `ppo.py` already uses; drop the two `allow_mps_bf16=True` or validate them; delete the hand-rolled PPO block in favour of `_set_ppo_training_kwargs`; `git mv` the fragment; one test pinning the loss scale; one docs note on `max_length`. No test lines were deleted and there are no ANSI offenders, so none of that is at risk.

**If there is no push within a week I will carve the PRM and MoLE half out myself and credit you for the diagnosis** — that half is clean and I would rather ship it than let it age. I would also rather you finished it, and the offer to split the `fsdp_config` find into its own issue under your name still stands. Say which you prefer.

MakazhanAlpamys (changes requested):
Thanks, @Oxygen56. Every item from my 09-12 and 09-14 comments is done, and the tests are now genuinely behavioural. I ran 17 value mutations and all 17 are killed, including the 7 that survived last time: PRM weight decay, scheduler and auto estimate; MoLE optimizer; PPO weight decay; unlearn clip and scheduler. The loss-scale and steps-cap semantics are pinned at three (batch, accumulation) shapes. The partial-group averaging test uses an exact synthetic gradient. Folding PPO into `_set_ppo_training_kwargs` behind `config_accepts` probes, instead of a second copy, is exactly right. On the tree merged with current `main`, 27 related test files give 1706 passed.

Two things before merge.

**1. `README.tr.md` now conflicts with `main`.** #1050 landed and added a README line, so the translation stamp cannot match. The simplest fix is to drop the README docs-index change (`README.md:263`, "(training settings and sequence limits)") together with its `README.tr.md` edit. The docs page already carries the note, and every README edit forces a Turkish re-translation.

**2. Unlearn must not start applying `training.quantization` in this PR.** Your setup now forwards `quantization=tcfg.quantization` to the policy and the NPO/RMU reference. I measured it with a minimal unlearn YAML that leaves quantization unset: it parses as `4bit`, and the base loads as bitsandbytes `Linear4bit`, where `main` loads plain `Linear`. So every existing unlearn config silently becomes a 4-bit QLoRA run. And `quantization: fp8` (or gptq/awq) now parses and then dies in `setup()`. #1043 (Srinivasan8888, approved today) takes the opposite position for `unlearn`: an unset value resolves to `none`, and an explicit bitsandbytes value warns and is ignored. That follows a ruling for exactly this case: a value the task never applied must not start changing runs silently. The two PRs do not conflict textually, so whichever landed second would invert the other without any test noticing. **Please drop the `quantization=` forwarding from `unlearn.py`** and leave that field to #1043. Everything else in the unlearn change stands.

One design note, not an ask: `batch_size: auto` in the unlearn loop changes only how many examples go into one optimizer update, not memory, because the loop runs one example at a time. So the same YAML updates every 4 examples on CPU, every 8 on this 8 GB card and every 128 on an 80 GB card (accumulation 4). If you agree, resolving `auto` to 1 for unlearn would make the run independent of the card. Your call; it can also be a follow-up.

MakazhanAlpamys (changes requested):
Two items left, and one of them is new since your push — #1043 merged this morning and collided with a part of this PR that neither of us predicted.

## 1. Still outstanding: drop the `quantization=` forwarding in `unlearn.py`

`src/soup_cli/trainer/unlearn.py:187` and `:209` are byte-identical to the head I reviewed. I want to be straight about what changed, because the *harm* I originally described is gone:

```
EVIDENCE: soup_cli.__file__ under the scratch tree, peft 0.20.0, tree = main + this PR
unlearn, quantization UNSET : quantization='none'
unlearn, quantization 4bit  : quantization='none'  + "training.quantization: 4bit has no effect on
                              task='unlearn' ... Soup v0.76 will refuse it."
unlearn, quantization gptq  : REFUSED at load
```

So since #1043 the forwarding is **inert** — the validator resolves the field to `none` before your trainer ever sees it, and the silent-4bit-QLoRA path I measured cannot happen today. I am still asking for the removal, for one reason: it is inert **by accident of validator ordering**, and `docs/training.md:274` now states as a contract that "unlearn: none only — policy and reference copy both load unquantised". The code says otherwise. If anyone relaxes #1043's resolver later, unlearn flips back silently and nothing fails. Two lines.

## 2. New, and not your fault: #1043 breaks one of your tests

```
FAILED tests/test_issue805_optimizer_settings.py::test_prm_and_mole_train_with_configured_optimizer[prm]
pydantic ValidationError: task='prm' does not apply training.lora: the PRM trainer fine-tunes
every base parameter. Remove the lora block (or set lora.r: 0).
```

Your `_config()` helper at `tests/test_issue805_optimizer_settings.py:53` hardcodes `lora: {r: 2, ...}` for every task, and #1043 added `_validate_prm_lora_block`, which refuses that on `prm`. It is a one-line fix in the test — `lora.r: 0` on the `prm` parametrisation, or omit the block there. The collision is in the PRM half, not the unlearn half, which is why neither of us saw it coming.

## What is closed

- **README / `README.tr.md`:** you kept the README change and re-translated *and* re-stamped the Turkish file rather than dropping it. That is a better answer than what I asked for — `tests/test_readme_translation_sync.py` passes 31 on the merged tree, and the Turkish line is genuinely translated, not just re-hashed. Closed.
- **`batch_size: auto` in unlearn:** you documented the card-dependent cadence in `docs/training.md` instead of resolving it to 1. That was explicitly your call and I accept the answer — saying so here so it does not come back a third time.
- Everything from the earlier rounds: fragment name and credit `(#805 by @Oxygen56 in #868)`, `allow_mps_bf16`, the `_set_ppo_training_kwargs` fold, loss-scale pinning, the docs. All verified.

## Verified on the merged tree

Base interpreter, peft 0.20.0, `soup_cli.__file__` under the scratch tree, merged against current `main`:

- #1043's own tests on the combined tree — **139 passed**, so your PR does not undo the ruling it brushes against.
- Your tests plus the previously-failing files — **1 failed, 146 passed, 5 skipped** (the `[prm]` collision above).
- Wider blast radius, 11 files including `test_ppo.py`, `test_v0610_part_a/b.py`, `test_issue748_*`, `test_issue803/804`, `test_trainer_coverage_v035.py` — **1 failed, 645 passed, 6 skipped**, same single failure.
- `ruff check` and `--select F821` clean. `git diff origin/main...<head> -- tests/` is **+245 / −0**: no assertion removed.

Mutations, two different trainers, both killed: `mole_routing.py` `optim=tcfg.optimizer` → a literal fails `test_prm_and_mole_train_with_configured_optimizer[moe_lora_routing]`; `ppo.py` `weight_decay` → `0.0` fails `test_ppo_constructs_real_config_with_nondefault_training_values`.

Push those two and I will merge on green. Seven days is far too long for this one and most of the delay is mine, not yours — the diagnosis was right from the first round.

MakazhanAlpamys:
Thank you — both items from the last round are fixed, and this is close.
- Both `quantization=tcfg.quantization` kwargs are gone from the unlearn loads, so the code now matches #1043's table in `docs/training.md` ("`none` only"). Your new loader-call test fails if either comes back.
- The `[prm]` helper uses `lora.r: 0`, which is the #1043 contract.

**What I ran**, on the tree merged with `main` at `c1a45efb`:
- The 15 files in your PR body: 740 passed, 6 skipped. The PRM / PPO / MoLE / unlearn neighbours: 353 passed. ruff and F821 clean. No deleted or weakened tests.
- **34 mutations, 34 kills.** Each forwarded setting was removed one at a time on PRM, MoLE and PPO. On unlearn: quantization restored on either load, the optimizer forced to AdamW, weight_decay 0, scheduler `linear`, warmup 0, a hard-coded clip norm, `max_length` 256, an accumulation threshold of 1, no gradient division, no partial-group flush, and auto batch forced to 1.
- It still merges cleanly with #1074; you share no files.

The PPO test builds a real trl `PPOConfig`, and the partial accumulation group is averaged by its *actual* size with a test pinning it. Both are the careful choice.

**One ruling before merge, on `batch_size: auto` for unlearn.** Unlearn processes one example at a time, so your loop turns `batch_size × gradient_accumulation_steps` into an accumulation count. That is right for an explicit `batch_size`. But with the default `auto`, the estimator's number is a memory-based *batch* size that this loop does not use as a batch. A default config would silently make many times fewer optimizer updates than it did before. So: **for `task: unlearn`, `batch_size: auto` resolves to 1.** Please add a line in the docs and a test that pins it; an explicit integer keeps your semantics.

The other two default shifts stand as you documented them, since they are what "the config is finally read" means:
- `data.max_length` (2048) replaces the hard-coded 256;
- PRM, MoLE and PPO receive Soup's `warmup_ratio` / `weight_decay` / `scheduler` / `optimizer` instead of HF's.

Please make sure the fragment says both in one sentence each, because existing unlearn runs will use more memory.

Nit: `import math` sits inside function bodies in `prm.py`, `mole_routing.py` and `ppo.py`. It is stdlib, so it can go at the top.

MakazhanAlpamys:
@Oxygen56 — status, and there is only one thing left on your side.

On 2026-09-19 I said this was accept-with-one-ruling: for `task: unlearn`, `batch_size: auto` resolves to
**1**. Both items from the round before that were already fixed, and 34 of 34 mutations killed — this is the
closest to done of anything in my queue that is still open.

Two days, no push, so this is a check rather than a nudge: **are you still on it?** A yes with no code is
fine and costs you nothing.

If you would rather I applied the one-line ruling myself and merged, say so and I will — your commits stay
yours, the credit stays yours, and I will name in the merge body exactly which line I added. I would rather
ask than take it.

One thing that has changed under you since 09-19, so you are not surprised when you rebase: `main` has moved
about 15 commits today, including #1074 (MoE flags per task) and #1022 (SFT FP8), both of which touch trainer
files. Your branch still merges clean as of now — I checked — but re-run the unlearn tests after rebasing
rather than assuming.

MakazhanAlpamys:
Heads-up that affects this PR mechanically, not its content: the changelog fragment baseline is moving from `changelog.d/0.75.0/` to `changelog.d/0.75.1/`.

Why: v0.75.1 shipped to PyPI while `main`'s CHANGELOG still stopped at 0.75.0, and four pending fragments described fixes that had already gone out in it — left alone, the next release would have announced them as new. Adding the 0.75.1 heading is the fix, and `validate_fragments()` requires the fragment directory to be named for the newest heading, so every pending fragment moves with it.

What it means for you: once that lands, this PR's changelog check goes red with

> `changelog.d/0.75.0/NNNN.category.md: fragment baseline is 0.75.0, but the latest release is 0.75.1; rebase and move the fragment before merging`

and the fix is one line on your next push:

```
git mv changelog.d/0.75.0/NNNN.category.md changelog.d/0.75.1/
```

Nothing about your diff is wrong and no review finding changes. I am telling you rather than letting a red X be the first you hear of it.

MakazhanAlpamys (changes requested):
The ruling is implemented exactly as given, and the substance of this PR is in very good shape — **19 of 20 mutations killed**, and the one survivor is the ruling itself. Three things stand between it and a merge, two of them small and one of them the reason CI is red.

**The ruling, verified:** `trainer/unlearn.py:235-242` resolves `batch_size: auto` to a literal `1`, with a comment explaining that the loop runs one example per forward/backward so the estimator would only change the update count, and it prints `Auto batch size (unlearn): 1`. Not "auto stays auto", not 2, and the estimator import is gone from that path. An explicit integer keeps your accumulation semantics intact.

**Merged with current `main` — clean, zero conflicts**, which is worth saying given how much landed today (#1101, #1102, #1105, #1031-adjacent trainer work). 29 passed on the two affected files, 1042 across a twenty-file blast radius, ruff clean. The mutation set is thorough: reverting `batch_size` to 2 fails 2, the optimizer/scheduler/weight-decay/warmup forwarding each fail their own tests, dropping the gradient `div_` or the partial-group flush each fail, restoring `quantization=` on the policy load fails 2, and reverting the four `max_length` sites fails 1. On `main` your new test file gives 10 failed / 1 passed.

**BLOCKING 1 — the README docs-index line is why all nine test cells are red, and it is not what you think.** `README.md:263` adds "(training settings and sequence limits)" to the training-docs row. Since your last round, `main` gained `README.ar.md` and `README.ja.md` alongside `README.tr.md`, and all three carry a stamp of the README they were translated from; your edit changes the hash. Reproduced three ways: at your raw head, 1 failed / 51 passed; on the merged tree, 1 failed; on plain `origin/main` in a separate worktree, **31 passed** — so it is the PR, not main.

The cheapest close is the one I suggested on 09-17: **drop the `README.md:263` edit.** `docs/training.md` already carries the content, and every README edit now forces a re-translation of three files, two of which belong to other contributors. Re-translating and re-stamping ar/ja/tr is a much bigger ask and puts their files in your hands.

**BLOCKING 2 — the ruling is implemented but not pinned hardware-independently, and it is inert exactly where it runs.** `tests/test_issue805_optimizer_settings.py:169-181` asserts `wrapper._batch_size == 1`, which is the right assertion, but it passes for the wrong reason on any GPU-less or small-GPU machine. On this box `estimate_batch_size(model_params_b=7.0, seq_length=64, quantization='none', lora_r=2)` returns **1** at both CPU and 8 GB, and **32** at 80 GB. So reverting the ruling entirely — back to the estimator call — leaves all eleven of your tests green, and CI runners have no GPU at all.

It is not vacuous (a constant 2 is caught); it just cannot see the regression the ruling exists to prevent. I verified the fix both ways: monkeypatching `get_gpu_info` to report an 80 GB A100 before `setup()` and then asserting `_batch_size == 1` **passes on your head** and **fails under the revert**. Your PRM/MoLE half of the same file already uses this technique at line 96, so it is two lines in your own idiom.

**HIGH — the docs now contradict the shipped ruling.** `docs/training.md:479-480` still says "The default `batch_size: auto` uses the same memory estimate as the other trainers." That was true of the previous head and is false now, and the 09-19 ruling explicitly asked for a line stating that auto means 1 on unlearn. One sentence.

**MEDIUM — the fragment is missing the two sentences I asked for on 09-19.** It covers "honour the settings" and "loss and sample counts keep their scale", but says nothing about (a) `data.max_length` replacing the hard-coded 256 — a default 2,048-token sequence, so **more memory and longer runs for every existing unlearn config** — or (b) `batch_size: auto` now meaning 1. Both are user-visible default shifts, and (a) is the one that will surprise someone mid-run.

**Everything else from the earlier rounds re-verified by execution**, so you do not need to revisit it: unlearn no longer forwards `quantization` (restoring it fails 2), the `[prm]` parametrisation uses `lora.r: 0` per #1043, the PPO test builds a real `trl.experimental.ppo.PPOConfig`, loss scale and partial-group averaging and the steps cap are pinned at three different (batch, accumulation) shapes, the two MPS-allowlist lines are gone and `test_issue563_mps_bf16_training` passes, and PPO's hardcoded `save_total_limit: 3` is gone.

Nit, still open from 09-19: `import math` is function-local at `prm.py:327`, `mole_routing.py:375`, `ppo.py:245` and `unlearn.py:233` — it is stdlib, so the lazy-import rule does not apply to it.

MakazhanAlpamys:
@Oxygen56, as offered on 09-12 and 09-14: I've filed the `fsdp_config` finding from your first round as #1204, credited to you. While checking it on `main` I found `online_dpo` has the same shape (it also drops `--deepspeed`), so it's in the same issue. Nothing changes for this PR; the items from my 09-22 review stand.

MakazhanAlpamys:
@Oxygen56, one clarification after my note on #805 (16:45Z): the three PPO items I added there (`--deepspeed` never reaching PPO, `response_length` not following `data.max_length`, and `save_total_limit` being `None`) are **not** part of this PR. Nothing changes for it, as I said at 15:56Z, and the items from my 09-22 review still stand.

So that #805 stays open for those three, please change `Fixes #805` to `Refs #805` in the description. If you'd rather not touch it, I'll do it at merge time.

PR #869, opened 2026-09-11
fix(data): reject an unknown --format on soup data validate (#866)

### LINKED ISSUE #866: soup data validate --format accepts any string and reports every row valid for it ("2/2 rows valid for bogus format", exit 0)
Found while reviewing #858 (@akkupratap323), and pre-existing on `main`. #858 does not change it.

## The defect

`soup data validate` accepts any string for `--format` and reports every row as valid in that "format":

```
$ soup data validate two_alpaca_rows.jsonl --format bogus
Issues found:
  ! 2 samples are very short (<10 chars)

2/2 rows valid for bogus format
```

Exit code 0. Reproduced on `main` at `1202bd94`.

`commands/data.py:81` declares `fmt: str` with help text listing the real formats, but never checks the value against `soup_cli.data.formats.VALID_FORMATS`. `validate_and_stats(data, expected_format=fmt)` then passes rows it has no converter for.

## Why it matters

`soup data validate` is the first step of the PR gate `soup ci init` writes. A typo in the workflow's `--format` (`--format sharegtp`) turns the gate into a pass for any file, including one the loader would reject. That is the same class of silent pass #811 / #858 just closed for "0 usable rows".

## Proposed fix

Validate `--format` against `VALID_FORMATS` plus `auto` before loading, and refuse an unknown value with a message listing the accepted ones. Make it a `click.Choice` or a manual check, whichever keeps the exit-code convention #858 settled on: input errors exit 1, a failed gate exits 2. Note that a Click usage error exits 2, which would collide with "failed gate", so a manual check that exits 1 is probably the right shape.

## Acceptance criteria

- [ ] `soup data validate f.jsonl --format bogus` exits non-zero with a message naming the accepted formats. The test fails on current `main`.
- [ ] Every entry in `VALID_FORMATS` is still accepted, pinned by a parametrized test.
- [ ] The exit code for an unknown format is the input-error code, not the failed-gate code.

SID-6921:
## Summary

Closes #866. `soup data validate` accepted any string for `--format` and reported every row valid in that "format" — exit 0, no error:

```
$ soup data validate two_alpaca_rows.jsonl --format bogus
Issues found:
  ! 2 samples are very short (<10 chars)

2/2 rows valid for bogus format
```

`commands/data.py` declared `fmt: str` with help text listing the real formats but never checked the value against `soup_cli.data.formats.VALID_FORMATS` before `validate_and_stats` ran.

## Why it matters

`soup data validate` is the first step of the PR gate `soup ci init` writes. A typo'd `--format` in a CI workflow (`--format sharegtp`) silently turned the gate into a pass for any file — the same class of silent pass #811 and #858 just closed for "0 usable rows".

## Fix

Validates `--format` against `VALID_FORMATS` (plus `auto`) right after the existing file-exists check, matching the same ordering and message style the `convert` command already uses for `--to` in the same file. A manual check, not `click.Choice`, so an unknown format exits **1** (input error) rather than colliding with the failed-gate exit code **2** that #858 settled on.

```
$ soup data validate two_alpaca_rows.jsonl --format bogus
Unknown --format: 'bogus'
Accepted: auto, chatml, alpaca, sharegpt, dpo, kto, llava, sharegpt4v, plaintext, embedding, audio, tool-calling, prm, pre_tokenized, input_output, video, multimodal, raft, asr
$ echo $?
1
```

## Verification

- `test_validate_rejects_an_unknown_format` — the new failure case, exit 1, message names the bad value and at least one accepted format.
- `test_validate_still_accepts_every_valid_format` — parametrized over every entry in `VALID_FORMATS`, so the guard can't be satisfied by rejecting everything.
- **Mutation**: reverted the guard and confirmed the new test reproduces the exact reported symptom (`exit 0`, `"2/2 rows valid for bogus format"`) before restoring.
- 24 tests pass in `test_issue811_data_validate_exit_codes.py` (18 new parametrized + 6 existing, all still green).
- `ruff check` clean. `mypy`: same 2 pre-existing errors as the unmodified file (confirmed via `git stash`), unrelated to this change and far from the edited lines.

## Not done here

`download_dataset` and `register_data` (same file) have their own, differently-scoped `--format` options with the same unchecked shape. Left alone — out of scope for this issue, which is specifically about `soup data validate`.

### DIFF
--- a/changelog.d/0.74.0/869.fixed.md
+++ b/changelog.d/0.74.0/869.fixed.md
@@ -1,0 +1,9 @@
+- **`soup data validate --format` accepted any string and reported every row valid for it
+  (#866 by @SID-6921 in #869).**
+  `commands/data.py` declared `fmt: str` with help text listing the real formats but never
+  checked the value against `soup_cli.data.formats.VALID_FORMATS` before `validate_and_stats`
+  ran, so `--format bogus` exited 0 and reported "2/2 rows valid for bogus format". Since
+  `soup data validate` is the first step of the PR gate `soup ci init` writes, a typo'd
+  `--format` in a CI workflow silently turned the gate into a pass for any file — the same
+  class of silent pass #811 and #858 just closed for "0 usable rows". An unknown format now
+  exits 1 (input error), not the failed-gate exit code 2.
--- a/src/soup_cli/commands/data.py
+++ b/src/soup_cli/commands/data.py
@@ -96,6 +96,16 @@ def validate(
         console.print(f"[red]File not found: {file_path}[/]")
         raise typer.Exit(1)
 
+    if fmt != "auto":
+        from soup_cli.data.formats import VALID_FORMATS
+
+        if fmt not in VALID_FORMATS:
+            console.print(
+                f"[red]Unknown --format: {fmt!r}[/]\n"
+                f"Accepted: auto, {', '.join(VALID_FORMATS)}"
+            )
+            raise typer.Exit(1)
+
     data = load_raw_data(file_path)
 
     # Auto-detect format if not specified
--- a/tests/test_issue811_data_validate_exit_codes.py
+++ b/tests/test_issue811_data_validate_exit_codes.py
@@ -10,6 +10,7 @@ from click.testing import Result
 from typer.testing import CliRunner
 
 from soup_cli.cli import app
+from soup_cli.data.formats import VALID_FORMATS
 
 from .conftest import strip_ansi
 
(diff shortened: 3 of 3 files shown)


MakazhanAlpamys:
Merged as `aaf6a128`. Thank you, @SID-6921 — this is an exact fix, and two of the choices in it are ones I would point other contributors at.

**You reused the single source of truth.** The guard imports `soup_cli.data.formats.VALID_FORMATS` — the same constant `validator.py:95` gates on — instead of writing a second tuple of format names beside it. A hand-copied allowlist would have worked today and drifted the first time someone added a format; this cannot.

**And you argued the exit code rather than picking one.** 1, not 2, because an unknown `--format` is an input error and not a failed gate — matching `soup data convert`'s `--to` guard in the same file and the pass-0 / gate-2 / input-1 shape #858 settled. Your note that a `click.Choice` would have exited 2 and collided with the gate meaning is exactly right, and it is the kind of thing that is invisible until someone wires `soup data validate` into CI and cannot tell a typo from a bad dataset.

Placing the guard before `load_raw_data` is the third small thing: a typo'd format now fails immediately rather than after reading a large file.

**What I ran.** 24 passed, with #858's exit-code tests untouched and green. Six mutations, all killed: deleting the guard (which restores `main` and fails exactly one test, so your test genuinely fails on `main`), changing the exit code from 1 to 2, inverting the `auto` check, inverting the membership test, dropping the accepted-format list from the message, and printing without raising. Measured live across five flag shapes — `bogus` exits 1, `auto` 0, omitted 0, `alpaca` 0, zero-usable-rows still 2. No test deleted or weakened; the diff on `tests/` is `+41/-0`. Green under `FORCE_COLOR=1` too. 13/13 required checks on `24637943`.

Your `CONTRIBUTORS.md` entry is in `074b7862` — sixth merge.

**Two follow-ups this turned up, neither of them yours to carry.**

The same input still reports every row valid through `soup mcp serve`: `mcp_server/registry.py:233-239` hands an unconstrained `format` string straight to `validate_and_stats`, so one file gets two different verdicts depending on the surface. The same call also shows a second, older bug — the tool's schema says "omit to auto-detect" and omitting it does no detection at all. Filed as **#878**, and it is yours if you want it, since you already have the context; say the word.

Smaller: `data.py:82-84`'s help text still lists 7 of the 19 accepted formats, so `--help` is now narrower than what the command takes. Discoverability survives because the *error* lists all 19, but a one-line fix would close it. Also correctly scoped out by you and worth someone's time later: `download_dataset`, `register_data` and `from_traces_cmd` have unchecked format options in the same file.

One correction to the issue rather than to you, so it does not get repeated in release notes: I claimed this was a live footgun in the generated CI gate, and it is not — `utils/ci_workflow.py:70` emits `soup data validate {data}` with no `--format`, so the generated workflow relies on auto-detect. The exposure is a hand-edited workflow. The fix is still right; my motivation was overstated.

PR #870, opened 2026-09-11
test(readme): close the #852 re-stamp gap — code-block contents, relative links, anchors and prose numbers

Ercaner1988:
## What does this PR do?

Closes the one gap you wrote down when merging #852: a re-stamp **without** a re-translation was caught only structurally, so `soup train` → `soup trainx` inside a fence, or `119.6` → `999.9` in prose, stayed green behind a valid stamp. Both of those now fail, stamp or no stamp.

### Three additions to the structure lock

Each is compared section by section against `README.md`, in both directions, like the existing checks:

1. **Code-block contents, with `#` comments dropped** — your suggestion. For `bash` / `yaml` / `sh` / `toml` / `python` fences, full-line and trailing `#` comments are removed and every remaining line must be byte-identical. Comments stay the one part of a code block a translation may change; a control proves a translated comment is still accepted.
2. **Relative link targets** — your second suggestion. `docs/models.md#optional-extras`, `LICENSE`, `benchmarks/`, image paths: the same set per section. In-page anchors cannot be compared, because they are translated with their headings (`#quick-start` → `#hızlı-başlangıç`), so instead each one must **resolve to a heading of the same file**, using GitHub's slug rule (lowercased; punctuation and symbols dropped; spaces to `-`). That turns the "all 37 relative links and localised anchors resolving" you checked by hand into an assertion.
3. **Numbers in prose.** Your formatting note — keep the English digits for measured figures — made numbers translation-invariant, so they can be locked like commands. This is the check that closes the `119.6` → `999.9` half of the gap, which the first two cannot see. When a figure differs only by its separator, the failure says so (`keep README.md's digits (119.6, not 119,6)`).

### What it found in `README.tr.md`

Code blocks, relative links and anchors were already clean. Numbers were not: **10 distinct figures, 18 occurrences, in two sections** carried Turkish separators — `119,6`, `3,32`, `113,00`, `4,8` — exactly the formatting you flagged. They are converted to `README.md`'s digits in this PR, which is the fix you asked for. Nothing else in the file changes: splitting each of the 11 changed lines into digits and text, every text segment is identical and every number differs only by its separator. `3,60` in the GIF's alt text is converted too for consistency, although attributes are not prose and are not checked.

### Whom to ping

The docstring now carries the line you asked for — **tr — @Ercaner1988** — and a small `MAINTAINERS` mapping feeds the same names into the stale-stamp and structure failure messages, so a red `main` says whom to ping without anyone opening the file. One test requires every translation to name a maintainer, which is your #769 policy ("a language with nobody behind it gets removed") made checkable. That is a little more than the one docstring line; it is easy to drop if you would rather keep it to the line.

### Controls

`TestTheRatchetCanActuallyFail` grows from 12 to 18 cases, each breaking exactly one thing in the temp tree: a changed command inside a fence (the #852 gap), a translated comment (control, must pass), a changed relative link, an anchor that resolves to no heading, a changed number (the other half of the gap), and a localised decimal separator. 23 tests in the file, all green.

### What this binds for the languages to come

Numbers are compared as ASCII digits, so every translation writes measured figures exactly as `README.md` does: Western digits rather than `١١٩٫٦`, `.` as the decimal point, and numbers that `README.md` spells as words stay words. Prose wording remains free.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [ ] `pytest tests/ -v` passes — `tests/test_readme_translation_sync.py` (23 tests) on Python 3.12 locally; the rest is left to CI
- [x] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed — only `README.tr.md`'s digits
- [x] Added a changelog fragment — `changelog.d/0.74.0/870.fixed.md`

Follow-up to #852.


MakazhanAlpamys:
@Ercaner1988 — reviewed at `e0df04b7`, 13/13 green. **This is merge-ready**, and the ratchet is demonstrably not decorative. One character to fix and one policy question I want to answer out loud rather than merge past.

## It closes the gap it is named for, and I proved each half separately

I ran every new check class twice — once against the pre-PR ratchet to confirm the hole was real, once against yours. All four were genuine gaps:

| defect injected into `README.tr.md` | before | after |
|---|---|---|
| prose figure `3.32 GB` → `9.99 GB` | green | **red** |
| the same, **with the stamp recomputed** — the #870 gap itself | green | **red** |
| `soup train` → `soup trainx` inside a ```` ```bash ```` fence | green | **red**, names the section and block index |
| `](docs/data.md)` → `](belgeler/data.md)` | green | **red**, both directions |
| in-page anchor pointing at no heading | green | **red** |

And the acceptance criterion stated as you framed it — edit `README.md`, leave the translation stale, **recompute the stamp so every stamp lock passes** — now goes red for a changed figure, a changed fenced command and a renamed doc path. That was the whole point and it holds.

Two design choices I want to note because they are better than the obvious version. Resolving anchors instead of matching them is right: a translated heading *must* have a different slug, so a matching check would have been permanently red. And dropping `#` comments from fence bodies means a translator can localise a comment without fighting the ratchet — you shipped a control test for that, which is what makes it credible.

**On false positives**, which is what I was most worried about: a ratchet that fires on correct content gets deleted by the next person, and this repo already has that written down as a lesson. I probed nine plausible *correct* Turkish renderings. Seven pass — a pure reword, a translated fence comment, a Turkish dotted-İ heading with both anchor forms re-slugged, a grammatical suffix on a figure (`2.59 katı`), reordered paragraphs. Two fire, and both are your policy working as designed rather than a bug: see below.

**What I ran.** 23 passed in the touched file; 652 passed, 9 skipped across it plus every file referencing README, translations or the changelog fixtures; 17 passed in the fragment suite; ruff clean. Nothing deleted or skipped. One assertion in `test_a_dropped_section_fails` is loosened from an exact list to `"…" in problems`, and I checked it is **forced** rather than convenient — your enriched fixture legitimately yields two problems when a heading is dropped, and the test still asserts the specific section-count string, so it still discriminates.

## The policy question, and my answer: yes, keep it

Your check forbids locale-native number typography — `3,32 GB` and `48.241 MiB` both go red with the "keep README.md's digits" hint — and all 16 figures in this PR's `README.tr.md` diff are exactly that rule being applied. I verified mechanically that **not one non-numeric character changed** across those line pairs: 16 figures, 16 separator swaps, no value altered.

I want to say plainly that this cuts both ways, because your docstring only argues one side. You write that a reader in a mixed locale takes `48.241 MiB` for roughly 48. True — and the enforced form has the mirror problem, since `48,241 MiB` reads to a Turkish eye as forty-eight point two four one. Neither form is clean for every reader.

**I am ratifying your rule anyway, for a reason that is about the ratchet rather than about typography:** the property this check needs is that a figure in the translation *is the same figure* as in the source, and locale-native forms are only comparable through a normaliser nobody has written. Your rule is checkable today; the alternative is machinery plus a policy. You are also a native speaker volunteering to live under it, which counts.

So it stands — and it is not a one-way door. If a translator for one of the five pending languages finds it genuinely misleading in their locale, the right answer is to grow a locale-aware number comparison in the ratchet, which would not weaken staleness detection at all. I would rather that than have someone quietly violate the rule.

## One character

The fragment says "15 figures written with Turkish separators". I counted: it is **16** — `0.20 0.5 0.914 1.4 1.52 2.59 3.32 3.60 4.8 9.8 28.9 48,241 71.3 113.00 119.6 18,658` — across 25 substitution instances. That number goes into published release notes verbatim, and v0.75.0 is being assembled shortly, so I would rather it were right.

Push that and I merge immediately. **If you are away for the next several hours, say so and I will merge as-is and correct the count myself** — I would rather not have this sit while the release closes.

## Two small things for later, neither blocking

Numbers inside HTML attributes are not policed: `_NOT_PROSE_RE` strips `<[^>]+>` wholesale, so a stale figure in an `alt=` or `title=` stays green. I confirmed it on the `` at `README.tr.md:75` — which you updated anyway, the one substitution of the 25 that was not required. Worth a docstring line if you ever revisit. And `_slug` keeps only `-_ ` plus Unicode letters/marks/numbers, so an emoji in a heading would diverge from GitHub's slugger; no current `##` heading contains one, so it is theoretical.

## And for #871, which I accepted an hour ago

I simulated the extraction against your hardened ratchet: removed all 87 lines of `## What's New` from `README.md` and `## Yenilikler` from the translation, left a 3-line pointer in each, recomputed the stamp. **Green, with no extra burden over the pre-PR ratchet** — and localising the pointer's path to `docs/yenilikler.md` correctly goes red. So the two changes are complementary rather than in tension: `What's New` is where the highest-churn measured figures live, so moving it out of `README.md` shrinks exactly the surface your number check has to police. When you open #871, translate the pointer prose and keep `docs/whats-new.md` byte-identical in both files.

MakazhanAlpamys:

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:

```
git mv changelog.d/0.74.0/870.fixed.md changelog.d/0.75.0/870.fixed.md
git push --force-with-lease
```

Ercaner1988:
Rebased onto `0e8cc1d`; the fragment is now `changelog.d/0.75.0/870.fixed.md`. v0.75.0 re-synced `README.tr.md` and replaced What's New, so the figure set changed under this PR: on the new head it is **10 figures, 18 occurrences, 11 lines** (the alt-text `3.60` included), and I checked line by line that only separators differ. The fragment now says 10. You were right that 15 was wrong: I had left out the alt text. I'll take your two later notes (numbers in attributes, emoji slugs) as a follow-up. The #871 PR comes next.

MakazhanAlpamys:
Merged as `46386851`. 14/14 green on `71b6075d`.

**On the count: you were right and I was wrong, twice over.** I said 15 was wrong and that it should be 16. You came back with **10 figures, 18 occurrences, 11 lines**, because the v0.75.0 rebase changed the figure set under this PR. I counted the diff myself rather than take either number:

```
changed lines: 11 removed / 11 added
occurrences whose form changed: 18
distinct figures: 0.20 0.5 1.4 3.32 3.60 4.8 9.8 71.3 113.00 119.6   (10)
pairs where more than the separator differs: []
```

Exactly your numbers, and zero pairs where anything but the separator moved. Recounting from scratch after a rebase rather than patching the number is what made that verifiable — a corrected `16` would have been just as wrong and much harder to catch.

**What this PR is, stated for the record:** you closed a hole in your own previous contribution. #852's stamp proves *when* a translation was synced and says nothing about whether it still says the same thing, so a changed command inside a fence or a changed figure in prose sailed through behind a freshly recomputed stamp. That is the failure mode where the guard looks healthiest.

Two design choices I want to name because they are better than the obvious version. **Resolving** anchors instead of matching them is right — a translated heading necessarily has a different slug, so a matching check would have been permanently red and then deleted. And dropping `#` comments from fence bodies lets a translator localise a comment without fighting the ratchet; you shipped a control test for that, which is what makes it credible rather than convenient.

And you probed **false positives** rather than assuming none — nine plausible correct Turkish renderings, seven pass. This repo has already deleted one guard for firing on correct code, so that is the work that decides whether a ratchet survives its first year.

Your two later notes — numbers inside attributes, and emoji slugs — are yours whenever you want them. And I will look at **#871** next; you have earned a faster turnaround than the one I gave you on this.

One thing I owe you plainly: this sat from **2026-09-12T06:55**, when I called it merge-ready, until now. Part of that was the v0.75.0 freeze, which I told you about at the time — but the freeze lifted at 12:37 yesterday and I still did not come back to it. That last stretch was me, not the release.

PR #873, opened 2026-09-11
fix(doctor): report the training stack as the [train] extra, not seven required failures (#828)

### LINKED ISSUE #828: soup doctor on a core-only install reports 7 required packages missing, exits 0, and suggests a bare pip command that drops the [train] upper bounds instead of `pip install "soup-cli[train]"`

blackcoderx:
Closes #828.

## Problem
Since v0.71.0 `pip install soup-cli` is intentionally light — the training stack
moved to the `[train]` extra and a core-only install is a supported state
(enforced by tests/test_cli_startup_is_light.py). But `soup doctor` still listed
torch, transformers, peft, trl, datasets, bitsandbytes and accelerate as
`required`, so on a core-only box it:
- reported seven MISSING failures for a supported install;
- printed a "Fix all" built from bare floors — dropping the declared
  `transformers<6.0.0` / `peft<1.0.0` / `trl<1.0.0` ceilings, pulling CPU-only
  torch on Windows, and breaking in cmd.exe (single quotes aren't quotes there);
- exited 0 despite the failures, so it couldn't gate CI.

`soup version --full` had the mirror bug: extras came from a hand-kept list of
six import names and knew nothing about `train`, `mcp`, etc.

## Change
- `soup doctor` classifies each checked package by the extra it ships in. A
  wholly-absent `[train]` stack collapses to a single
  `pip install "soup-cli[train]"` advisory (plus a CUDA index note when an
  NVIDIA GPU is present), not seven MISSING rows.
- Every suggestion is built from the extra name and double-quoted, so the
  declared ceilings apply and the command works in bash, PowerShell and cmd.exe.
  No suggestion contains a bare floor.
- Exit is non-zero only for a blocking problem — a missing *core* dependency or
  a package installed beyond its ceiling — so `doctor` can be a gate; a
  supported partial install stays exit 0.
- `soup version --full` derives installed extras from metadata
  (`env_lock.installed_extras()`), so `train`, `mcp` and any future extra are
  reported without editing a list.

## Acceptance criteria
- [x] Missing `[train]` → suggestion is `pip install "soup-cli[train]"` (test fails on old code)
- [x] No suggested command carries a bare `[train]` floor
- [x] A missing core dependency exits non-zero
- [x] `soup version --full` lists `train` and `mcp` on a `[dev]` install

## Testing
- `pytest tests/test_doctor.py tests/test_cli.py` — new + existing green.
- Guardrails green: test_cli_startup_is_light, test_issue636_torch_floor,
  #352/#502 floor pins, test_v0401_part_c, test_v07137 quoting, #755 config leg,
  and every DEPS consumer (qat/unsloth/vision/sglang).
- Manual on a `[dev]` box: `soup doctor` exits 0 with double-quoted
  `soup-cli[...]` hints and no bare floors; `soup version --full` shows
  `extras: ... train ... mcp ...`.


MakazhanAlpamys:
@blackcoderx — thank you for this, and I owe you a straight answer rather than a slow one.

**#828 was fixed by #854, merged this morning as `12d08ade`.** That PR was opened about ten hours before yours, went through three rounds of review, and was green first, so it is the one that went in. **The collision is on my issue tracking, not on you** — #828 was not assigned and I did not mark it as being worked on, so you had no way to know. That is mine to fix and I am sorry it cost you a full evening's work.

I did not want to tell you "already fixed" without checking, so I compared your branch against merged `main` line by line. Most of it converged: the group collapse, double-quoted suggestions that carry the declared ceilings, Rich escaping of `[train]`, the two-step CUDA advice (yours is right about `--index-url` needing its own step, which is the part most people get wrong), a missing core dependency exiting 1, and metadata-derived extras for `soup version --full`. Your `test_version_full_lists_installed_extras_from_metadata` **passes against main unchanged** — main already derives extras from `importlib.metadata`.

Two things in your diff are not on main, and one of them is a real bug you found.

## The one I would like you to take: #874

Your exit-code policy is stricter than main's in exactly one place — you make an **installed-but-beyond-its-ceiling** package blocking. Main does not, and when I went looking I found it is worse than a missing feature:

- The extra-group loop builds the `INCOMPATIBLE` status and the `Downgrade ...` advice but never sets the blocking flag (`doctor.py:221-227`).
- `_MAX_EXCLUSIVE` is exactly `transformers`, `peft`, `trl` (`doctor.py:69-73`) — and all three live in `EXTRA_GROUPS["train"]`, none in `DEPS`. So the core branch that *would* block, `if required: core_broken = True` at `doctor.py:168-169`, **has no input that can reach it**.

Every package that can produce an INCOMPATIBLE verdict is one whose incompatibility leaves the exit code at 0. `pip install "transformers==6.1.0"` then `soup doctor` prints a red INCOMPATIBLE row and exits 0, so a CI job gating on it catches a missing `pydantic` and sails past a transformers that breaks every training run.

Filed as **#874** with reproduction and acceptance criteria. It is yours if you want it — roughly fifteen lines in `doctor.py` plus the test. One honest note so the review is not a surprise: your PR does not currently ship a test for this behaviour, and it is the part that most needs one.

**#875** is also open if you would rather have a second, smaller one: `group_missing` is set as soon as *any* member is absent, so with six of seven installed and OK, `soup doctor` still says "Training stack not installed". Your version avoids that false claim by gating on `present == 0` — but then prints "All checks passed!" and a "To fix:" line in the same output, which is a contradiction rather than a fix. Neither PR gets this one right; the correct output is a third thing, and the issue says so.

## The one I would leave

`installed_extras()` moved into `utils/env_lock.py` is a good instinct — that module already holds `_SELF_DIST` and the same marker-evaluation idiom, so the duplication you spotted is real. But main's `_installed_extras()` is already a module-level function rather than inline, and I could not find a second caller anywhere in the tree: nothing else enumerates `Provides-Extra`, and `EnvLock` has no extras field. Moving a single-caller function across files is churn until something else wants it. If `soup env lock` ever records which extras were installed, it becomes worth doing — but that is a feature, not a refactor, and I would rather not ask you to land it as one.

Closing this one as superseded. If you take #874, open a fresh PR against current `main` rather than rebasing this branch — main's tables moved to `EXTRA_GROUPS` and a rebase will fight you for no benefit.

Whatever you decide, the diagnosis in this PR was correct and independently arrived at, and the ceiling-blocking point is something the merged PR got wrong and you got right. That is worth more than the merge would have been.

PR #877, opened 2026-09-12
fix(doctor): exit non-zero when a [train] package is beyond its declared ceiling (#874)

### LINKED ISSUE #874: soup doctor prints INCOMPATIBLE and exits 0: every package with a declared ceiling lives in the [train] group, so the blocking branch is unreachable
`soup doctor` gained a real exit code in #854 (`12d08ade`): a missing or incompatible **core** dependency exits 1, so the command can gate CI. The `[train]` group deliberately does not affect the exit code — a core-only install is supported and must stay exit 0.

But the INCOMPATIBLE verdict fell through the gap between those two rules, and with today's tables it can never affect the exit code at all.

## The two halves

**1. An incompatible `[train]` package is reported and then ignored.** The extra-group loop builds the `INCOMPATIBLE` status and the `Downgrade ...` issue but never sets `core_broken`:

https://github.com/MakazhanAlpamys/Soup/blob/main/src/soup_cli/commands/doctor.py#L221-L227

**2. The core branch that *would* set it is unreachable.** `_MAX_EXCLUSIVE` is exactly `transformers`, `peft`, `trl` (`doctor.py:69-73`). All three live in `EXTRA_GROUPS["train"]` (`:56-62`); none is in `DEPS` (`:27-31`). So `if required: core_broken = True` at `doctor.py:168-169` has no input that can reach it.

Together: **every package capable of producing an INCOMPATIBLE verdict is one whose incompatibility leaves the exit code at 0.**

## Why it matters

`transformers 6.x` breaks every training run this project supports — that ceiling is declared in `pyproject.toml` and #502/#503 exist because the stack moved. A CI job gating on `soup doctor` today catches a missing `pydantic` and sails past a `transformers` beyond its ceiling, printing a red `INCOMPATIBLE` row while exiting 0. That is the same shape as the bug #828 fixed: the diagnosis is right and the exit code disagrees with it.

## Reproduce

```
pip install "transformers==6.1.0"   # or any version >= the declared ceiling
soup doctor
echo $?     # prints 0, with INCOMPATIBLE on screen
```

## Suggested fix

An installed-but-out-of-range package is blocking wherever it is found — core or extra group — because it is not a missing optional feature, it is an installed package Soup has declared it cannot work with. Concretely: set the blocking flag in the extra-group INCOMPATIBLE branch, and keep the core branch (it stops being dead the moment a core package gains a ceiling).

Deliberately **not** in scope: a missing or outdated `[train]` package must still exit 0. That is #828's fix and it stays.

## Acceptance criteria

- [ ] `transformers` (or `peft`, or `trl`) installed at or beyond its declared ceiling makes `soup doctor` exit non-zero, whether it is reached through `DEPS` or through `EXTRA_GROUPS`
- [ ] A missing `[train]` group still exits 0
- [ ] A partially installed `[train]` group with every present member in range still exits 0
- [ ] An *outdated* (below floor) `[train]` package still exits 0 — only beyond-ceiling is blocking
- [ ] A test that fails against current `main`, and that the core branch at `doctor.py:168-169` is exercised rather than left unreachable

chiliec:
## What does this PR do?

Closes #874.

`soup doctor` printed a red `INCOMPATIBLE` row for a `[train]` package beyond its declared ceiling (e.g. `transformers 6.1.0`) and still exited 0. The extra-group loop built the issue but never set `core_broken`, and since every package in `_MAX_EXCLUSIVE` (`transformers`, `peft`, `trl`) lives in `EXTRA_GROUPS["train"]`, the core branch that does set it had no reachable input.

Fix: set `core_broken` in the extra-group INCOMPATIBLE branch. The core branch is kept, so it takes effect the moment a core package gains a ceiling. A missing or outdated `[train]` member stays advisory (exit 0), per #828.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Tests

Four new tests in `tests/test_doctor.py`, one per acceptance criterion:

- `test_doctor_incompatible_train_member_exits_nonzero` — `transformers 6.1.0` via `[train]` → exit 1 (**fails on current `main`**, passes with the fix)
- `test_doctor_incompatible_core_dependency_exits_nonzero` — a core dependency past a ceiling → exit 1 (exercises the previously-unreachable core branch)
- `test_doctor_outdated_train_member_exits_zero` — below-floor `[train]` member → exit 0
- `test_doctor_partial_train_group_in_range_exits_zero` — one present-and-OK, one missing → exit 0

Validation (Python 3.11):

```
# fix stashed, tests kept
pytest tests/test_doctor.py -k "874 or exits"  → 1 failed (…_train_member_exits_nonzero), 5 passed
# fix restored
pytest tests/test_doctor.py                     → 32 passed
ruff check src/soup_cli/commands/doctor.py tests/test_doctor.py → All checks passed!
```

The two `test_doctor_nccl_*` tests were deselected locally — they need `torch`, which is not installed on this box (core-only install); they are unrelated to this change.

## Checklist

- [x] `pytest tests/ -v` passes (see note on the two torch-dependent NCCL tests above)
- [x] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed — no doc change needed


MakazhanAlpamys:
@chiliec — welcome, and thank you for picking this up. I filed #874 a few hours ago and did not expect a PR the same night.

I have approved your CI runs manually — that is the one-time gate GitHub applies to a first-time contributor's workflow runs, not anything about your branch. I checked your diff touches no `.github/` files before approving, which is the only reason that check exists.

**One scheduling note so the wait is visible rather than mysterious.** v0.75.0 is mid-release: the release PR (#879) has already assembled `CHANGELOG.md` from the per-PR fragments in `changelog.d/`, so `main` is under a merge freeze until the tag is pushed. A PR merged in that window leaves a fragment behind and makes `publish.yml --check-empty` fail the tag. So this will be reviewed now but merged after the tag — the delay is the release, not your PR.

A heads-up on what I will be checking, so there are no surprises. #874 has two halves and the second is easy to miss: the extra-group loop reports `INCOMPATIBLE` without setting the blocking flag, **and** the core branch that would set it (`doctor.py:168-169`) is unreachable today, because `_MAX_EXCLUSIVE` contains exactly `transformers`, `peft` and `trl` and all three live in `EXTRA_GROUPS["train"]`, none in `DEPS`. So a fix that only touches the extra-group loop leaves dead code that silently becomes live the day a core package gains a ceiling. The acceptance criteria also require that a *missing* or merely *outdated* `[train]` package still exits 0 — that is #828's fix and it must survive.

I will also mutate the fix: if the test for "beyond the ceiling exits non-zero" still passes with the guard removed, that is a blocking finding regardless of how the code reads.

Full review shortly.

MakazhanAlpamys:
@chiliec — reviewed at `8f6fa7b9`. **The code is correct and nothing in it needs to change.** Two items below, and the first one is the release's fault rather than yours. For a first contribution this is unusually disciplined work, and I want to be specific about why.

## What I measured

I ran the same harness against both trees, driving the real `app` through `CliRunner` with `DEPS` / `EXTRA_GROUPS` / `_MAX_EXCLUSIVE` patched:

| case | main | your PR | required |
|---|---|---|---|
| ceiling via `EXTRA_GROUPS` (`transformers 6.1.0`) | **0** | **1** | non-zero ✅ |
| ceiling via `DEPS`, required | 1 | 1 | non-zero ✅ |
| `[train]` entirely missing | 0 | 0 | 0 ✅ |
| `[train]` partial, present members in range | 0 | 0 | 0 ✅ |
| `[train]` member below floor | 0 | 0 | 0 ✅ |
| missing CORE dependency | 1 | 1 | 1 ✅ |

One behavioural delta, exactly where #874 said it should be, and **every #828 guarantee survives** — I checked those four by execution rather than assuming, because they were merged only two days ago and are the easy thing to break while fixing this.

Four mutations, all caught: deleting `core_broken = True`, inverting it, making it also fire on a **missing** group member (4 tests fail), and making it fire on an **outdated** one. That last pair matters — it means your two *negative* tests are load-bearing, not decoration.

## Three things I want to name, because most contributors miss them

**You rewrote the comment instead of appending to it.** The old text at `:133-138` claimed the `[train]` group "never touch[es] the exit code" — which your fix makes false. Leaving it would have made the file lie about itself, and that is the most common way a correct one-line fix still degrades a codebase.

**You kept the core branch and made it testable rather than deleting it as dead code**, which is the half of #874 that is easy to miss. And your test for it is *honestly labelled*: it is a reachability guard, not a regression test. I verified it passes on `main` too — and your PR body already said so, and said which single test actually fails on main. Nothing in that body overstated what you had verified, which is rarer than it should be.

**Four tests, one per acceptance box, two of them negative.** The mutation that makes the fix fire on a missing member is caught by *your new partial-group test* plus three from #828 — so you added coverage on the exit-code axis that the existing suite did not have.

The diff is one line of behaviour for a bug whose report invited a much larger change. No refactor, no drive-by.

## Two changes

**1. The fragment has to move — and this is not your doing.** v0.75.0 merged to `main` (`f50a7da3`) at 08:26 UTC, *after* your checks went green, and it emptied `changelog.d/0.74.0/`. I reproduced the consequence on a real merge:

```
test_repository_fragments_match_the_newest_release
  ChangelogError: changelog.d\0.74.0\877.fixed.md: fragment baseline is 0.74.0,
```

`git mv changelog.d/0.74.0/877.fixed.md changelog.d/0.75.0/877.fixed.md` and it is 17 passed. **No content edit** — the fragment is already in house form, PR-numbered, with the credit in the right shape. You are one of sixteen open PRs in the same position.

**2. `docs/commands.md:219` now documents the taxonomy incompletely.** #854 put it in docs and your PR widens it:

> `…exits 2 if it cannot be read, and exits 1 when a required core dependency is missing or incompatible (#828).`

A reader concludes a `[train]` package cannot produce exit 1 — the exact behaviour you changed. Suggested: `…and exits 1 when a required core dependency is missing, or when any installed package is beyond its declared ceiling — core or [train] (#828, #874).`

While you are there, one line of comment at `doctor.py:171`, please. The core branch you kept is still guarded by `if required:` while your new assignment is unconditional, so the two blocking sites now follow different rules: an *optional* `DEPS` row past a ceiling exits 0 where the same package in an extra group exits 1. I measured it; it is unreachable today (`_MAX_EXCLUSIVE` and the optional `DEPS` rows do not intersect), and you followed #874's literal instruction to keep the branch, so this is the issue's wording rather than your error. But it is the same "dead code that quietly becomes live" shape #874 exists to prevent, so I would rather the asymmetry be written down than rediscovered.

## Two things that are not your problem

Your two new `"INCOMPATIBLE"` assertions fail at `COLUMNS=60`, because Rich truncates the cell to `INCOMPA…`. The pre-existing #828 sibling fails identically **on `origin/main`** at that width, and real CI runs at the default, so this is a repo property rather than a regression. Your *colour* handling — the half that has actually turned this repo's CI red four times — is correct: every new assertion goes through `_strip_ansi` and the exit-code assertions are immune to both width and colour.

The two NCCL tests you deselected locally for lack of torch **do run here and pass**. Flagging them with the reason instead of staying silent was the right call.

`main` is still frozen: the release is merged but not yet tagged, so nothing goes in until the tag lands. Push those two whenever suits you and I will merge as soon as it does.

MakazhanAlpamys:

```
git mv changelog.d/0.74.0/877.fixed.md changelog.d/0.75.0/877.fixed.md
git push --force-with-lease
```

MakazhanAlpamys:

```
```

```
```

MakazhanAlpamys:
@chiliec — a real status check; my last comment here was a fragment-broadcast notice rather than review.

Two mechanical items, nothing new:

1. The fragment is at `changelog.d/0.74.0/877.fixed.md`. That directory was assembled and deleted by the v0.75.0 release, so it will fail `test_issue487_changelog_fragments.py` in every matrix cell. `git rebase origin/main` then `git mv` it to `changelog.d/0.75.0/877.fixed.md` — the body needs no change.
2. The docs line from my review.

Worth knowing: your CI tick is from **2026-09-12 08:06**, twenty minutes before the release merged, so it is stale rather than meaningful — it will re-run red on the fragment alone.

The substance was fine; this is entirely release-timing churn that landed on you, not anything you did. Still have time for it? If not, say so and I will do both and merge it credited to you.

chiliec:
Still have time, and thanks for the patience — both items are in d57f101, on top of current `main` (`fcca8ef`):

1. **Fragment** moved with `git mv` to `changelog.d/0.75.0/877.fixed.md`, body untouched. `test_issue487_changelog_fragments.py` passes locally against the new baseline.
2. **`docs/commands.md`** exit-code sentence now reads: *"…exits 1 when a required core dependency is missing, or when any installed package is beyond its declared ceiling — core or [train] (#828, #874)."* And the `if required:` asymmetry on the core ceiling branch in `doctor.py` now has a comment spelling out that an optional `DEPS` row past a ceiling would exit 0 where the same package in an extra group exits 1, that it is unreachable today, and why it was kept.

No behaviour change; `tests/test_doctor.py` is 53 passed locally, with the same two NCCL tests skipped for lack of torch as before.

MakazhanAlpamys:
Verified `d57f101a`. **Merge-ready** — both outstanding items are closed and I checked them by execution rather than by reading your reply.

- **Fragment** is now `changelog.d/0.75.0/877.fixed.md`, a rename with no body change, house credit form `(#874 by @chiliec in #877)`. `changelog.d/` no longer contains a `0.74.0` directory at all.
- **Docs** at `docs/commands.md:219`, and I checked it is the *only* place documenting `soup doctor`'s exit codes — so the coverage is complete rather than partial.
- **The asymmetry comment** at `doctor.py:171-176` is the part I want to name. You could have deleted the `if required:` branch and had a simpler diff. Instead you kept it, wrote down that an optional `DEPS` row past a ceiling exits 0 where the same package in an extra group exits 1, and said it is unreachable today. That is the version a maintainer can act on in a year.

What I ran:

- `tests/test_doctor.py` → **38 passed** (the two NCCL tests you had to skip locally run and pass here), plus fragments → 55 passed together. `ruff check` clean.
- **Mutation**, anchor counted at exactly 1: deleting the added `core_broken = True` at `doctor.py:236` — module still imports — fails `test_doctor_incompatible_train_member_exits_nonzero` on `assert 0 == 1`, and nothing else. A behavioural kill, not a collection error.
- **Baseline** on `origin/main`: exactly one failure, the behavioural one. Your PR body already disclosed that `test_doctor_incompatible_core_dependency_exits_nonzero` passes on main as a reachability guard, and the docstring says so — that honesty is why I did not have to discover it.
- **Real smoke on this box** (transformers 5.17.0 / peft 0.20.0 / trl 0.29.1, all in range): `soup doctor` → `All checks passed!`, real exit code **0**. No false positive on a healthy install, which is the failure mode that would have hurt most.
- Scope is bounded by construction: `EXTRA_GROUPS` holds only `train` and `_MAX_EXCLUSIVE` is exactly `{transformers, peft, trl}`, all inside it.

Your base is 6 commits behind `main`, which also touched `docs/commands.md` — at line 178 against your 219, so `merge-tree` reports zero conflicts and **no rebase is needed**.

The only thing left is CI. The Actions queue is at ~88 runs, so it will be slow; I will merge the moment your matrix is green. Nothing further is needed from you.

MakazhanAlpamys:
Re-verified against current `main` (`b0a63382`), which has taken 16 merges since I reviewed you. **Still merge-ready, nothing changed** — `merge-tree` clean, 38 passed on both the PR head and the merged tree, the mutation still dies on `assert 0 == 1`, ratchets 138 passed, fragment intact with no duplicate number.

`main` merged #660 and #911 into `commands/doctor.py` (+42/−2) and yours merges clean over both.

**One latent coupling I found while checking that, worth knowing rather than acting on.** `tests/test_cli_subprocess.py::TestDoctorCommand::test_doctor_runs` spawns a real `soup doctor` and asserts `returncode == 0`. It is not skipped and it was outside the scope I reviewed. After your change it passes **because this box happens to sit under every ceiling** — transformers 5.17.0 < 6.0.0, peft 0.20.0 < 1.0.0, trl 0.29.1 < 1.0.0.

So that pre-existing test becomes environment-dependent once this lands: any box or CI image whose `[train]` packages drift past a declared ceiling turns it red. That is arguably `soup doctor` working exactly as designed — the whole point of your PR is that an incompatible package exits non-zero — but it will read as an unrelated failure to whoever hits it, in a test that has nothing to do with dependency bounds.

Not a blocker and not something to fix here. I ran it on the merged tree with `test_mlx_backend.py`: 99 passed.

MakazhanAlpamys:
Merged as `f789c9e3`. Thank you, and welcome — this is your first contribution here.

**Merged on the verdict recorded at `d57f101a`**, unchanged and not re-derived. All 13 required contexts green on that exact SHA, read per-SHA rather than from the cached PR summary. What I re-checked instead of re-reviewing is what can move while a PR waits: `main` took sixteen merges after my review, two of them into `commands/doctor.py` (#660 and #911, +42/−2), and this merges clean over both. On the merged tree, **104 passed** across `test_doctor.py`, #757's package-wide typer-defaults scanner — which walks every command module and therefore this one — the CLI startup-is-light guard, and the changelog fragment validator.

That third one is not incidental. `soup doctor` sits on the light CLI path, and a doctor change that pulls torch in at module scope slows startup for every command in the tool; I confirmed `torch in sys.modules: False` after `doctor --help` on the merged tree.

**Why this one matters more than its size suggests.** A check that prints a warning and exits 0 is not a check — it is a log line. Anyone running `soup doctor` as a CI pre-flight was getting a green gate on **the exact stack Soup declares it does not support**, and the failure mode of that is silent: nothing goes wrong until something goes wrong much later, somewhere unrelated, with no thread back to the version bound. Making the exit code carry the warning is what turns it into a gate.

The part I want on the record is the pairing. #368 had already taught `soup env check` to audit installed versions against Soup's declared bounds. Before this, `doctor` and `env check` disagreed about the same fact — one red, one green — which is worse than either answer alone, because it teaches people to ignore whichever one is inconvenient. Now they agree.

The mutation still dies on `assert 0 == 1`, and 38 passed on both the PR head and an earlier merged tree.

Credited in `CONTRIBUTORS.md` (`8b97a96c`); you are the 63rd. #874 is closed with credit.

An apology for the wait: Actions on this repo produced no terminal outcome at all between 2026-09-13T18:30Z and today 15:08Z, with about a hundred runs stacked behind it. Your checks were not failing, they were not running. That is cleared now.

PR #879, opened 2026-09-12
release: v0.75.0

MakazhanAlpamys:
Release branch for **v0.75.0** — cut in a separate worktree while contributor PRs keep landing on `main`.

**What this PR carries beyond the changelog fragments already on `main`:**

- Version bump to 0.75.0 (`pyproject.toml` + `src/soup_cli/__init__.py`).
- **Breaking, and promised in v0.74.0 (#627):** unknown config keys now *refuse* the load instead of warning. `UNKNOWN_KEY_REJECTION_VERSION = "0.75"` and `TestTheDeadline` pin the flip to the declared version in both directions. The refusing branch says `Refused.` rather than `Not applied.`
- **From the release review (three agents, every finding fixed and mutation-checked):** the detector now applies the same root-level `lora:` remap `SoupConfig` has honoured since v0.40.1 — the first draft would have refused a legal spelling; key names are escaped before they reach the terminal (new shared `soup_cli/utils/terminal.py`); the walk is capped at 100 findings (50k bogus keys measured ~31 s CPU per Web UI request before); `/api/train/start` returns the key and the suggestion instead of a generic error; docs no longer claim `soup plan`/`apply` run the check.
- Two `soup fetch examples` configs moved from the root-level `lora:` spelling to the canonical `training.lora`; all 165 recipes, 21 templates and `examples/configs/` scan clean (now guarded by a test).
- README `What's New` for v0.75.0; the three collapsed previous-release blocks displaced to GitHub Releases (523 → 476 lines). README.tr.md re-translated for the changed sections and re-stamped.
- SECURITY.md supported versions → 0.75.x; docs/backends-and-ops.md describes the refusal as current.

**Still to come on this branch before it leaves draft:** CHANGELOG assembly from `changelog.d/0.74.0/` (merge freeze on `main` from that point until the tag), contributor credit check against `CONTRIBUTORS.md`, full local suite result.

**Tagging plan:** squash-merge to `main`, wait for every required check on the merged SHA, tag *that* SHA.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #881, opened 2026-09-12
fix(ingest): harden Langfuse pull bounds and filtering (#865)

### LINKED ISSUE #865: Langfuse --pull: enforce a total deadline, filter GENERATION client-side, and stop on a repeated cursor
Follow-up to #859 (@Konuktor's `soup ingest --source langfuse --pull`, merged as ffb74e58). Review found no blocking issues. These three were found while verifying it, and each is a small, testable change in `src/soup_cli/utils/ingest_pull.py`.

## 1. The timeout is per socket operation, not a total deadline

A server that sends one byte every 0.2 s kept a `timeout=0.5` call alive for 7.8 s in review. With the default 30 s timeout and the 64 MiB response cap, a slow-drip host can hold the command open almost indefinitely. The exposure is small, because the host is the operator's own choice and TLS is required, but the bound the flag promises is not the bound the code enforces.

**Fix:** track a wall-clock deadline inside `_read_capped` (and around the connect) and abort loudly when it passes.

## 2. The GENERATION filter is server-side only

The only filter is the `type=GENERATION` query parameter. A server that ignored it (an older self-hosted version, or a proxy that drops unknown parameters) would have spans and tool calls written out as training rows.

**Fix:** a client-side check on each observation's `type` in `_to_parser_row`. Count skipped non-generation rows in the summary rather than dropping them silently.

## 3. A repeated cursor is not detected

A server that returns the same cursor forever runs until `--max-pages` (100 requests by default) before stopping. It does stop loudly and writes nothing, so this is wasted requests, not wrong data.

**Fix:** remember the cursors already seen, and stop on the first repeat with a message naming the cursor loop.

## Acceptance criteria

- [ ] A test where a transport drips bytes past the configured timeout, and the call fails within about `timeout + epsilon` wall-clock time. It fails on current `main`.
- [ ] A test where the page contains a non-GENERATION observation and it is not written as a row. The summary reports it as skipped.
- [ ] A test where a transport returns the same cursor twice, and the pull stops after the second page with a cursor-loop message.
- [ ] No change to the default (no `--pull`) path; `test_local_export_loads_no_pull_code_and_opens_no_socket` still passes.

Konuktor:
Closes #865 — the three findings from your review of #859, each reproduced before it was fixed.

## 1. The timeout was per socket operation, not a total deadline

`timeout` reached `opener.open` and each `read`, so it bounded one socket operation. A server sending a byte before every timeout expired kept the call alive indefinitely. Reproduced with a loopback server that drips 20 bytes 0.1 s apart against a 0.4 s timeout — on `main` the call **succeeded after 2.01 s**:

```
E  AssertionError: a 0.4s timeout allowed a 2.01s call (20 chunks 0.1s apart)
E  assert 2.0119723179999482 < (0.4 + 0.6)
```

`_urllib_transport` now takes a monotonic deadline before the connect and passes it to `_read_capped`, which reads through `read1` — returning bytes already received instead of blocking for a full chunk — and checks the deadline between reads. The same drip now fails in ~0.4 s with "the request to Langfuse timed out after 0.4s".

The bound is per request, not per pull. A read already in flight can overshoot by at most one socket timeout, which is what bounds a server that stops sending entirely; the docstring says so rather than implying a hard guarantee.

## 2. `type=GENERATION` was only a query parameter

A server that ignored it — an older self-hosted version, or a proxy dropping unknown parameters — would have spans and tool calls written out as training rows. On `main`, a page of GENERATION / SPAN / TOOL / GENERATION produced four rows:

```
E  AssertionError: assert ['obs-1', 'obs-2', 'obs-3', 'obs-4'] == ['obs-1', 'obs-4']
```

Each observation's own `type` is now checked before it becomes a row. The query parameter is still sent — a control test asserts that, since the client-side check is defence in depth, not a replacement. Skipped observations are counted in a new `PullStats` and reported next to the existing "no input or no output" line:

```
Wrote 2 traces from langfuse (2 generations pulled) -> out.jsonl
2 observation(s) were not generations and were skipped.
```

## 3. A repeated cursor was only stopped by `--max-pages`

A server repeating one cursor spent the whole budget — 100 identical requests by default — before stopping. On `main`:

```
E  AssertionError: must not spend the rest of --max-pages
E  assert 100 == 2
```

Cursors already seen are remembered, and the second one fails immediately: "Langfuse returned a pagination cursor it had already sent, after 2 page(s); stopping rather than repeating the same request". Requests drop from 100 to 2.

## What stayed the same

#859's guarantees are re-checked by mutation, not assumed: no credential material in any message, log or artefact; redirects refused; `LANGFUSE_HOST` validation; staged atomic output; and the default (no `--pull`) path byte-identical and network-free.

Three fixtures in `test_issue204_langfuse_pull.py` simulated "more results pending" by repeating a single cursor, which is now its own error. They use distinct cursors so they still exercise the page cap, and the cap's own assertions (pages, rows, calls, exit code, untouched previous output) are unchanged. One stub there gained `**kwargs` because `_read_capped` now takes the deadline.

## Tests

`tests/test_issue865_langfuse_pull_bounds.py`, 8 tests: the drip-feed deadline and a control that a prompt response still arrives; the mixed-type page filtered to generations, the CLI summary line, the stats counts, and a control that the server-side parameter is still sent; the repeated cursor stopping at page two, and a control that distinct cursors still page.

**16 mutations, all killed** — deadline check removed, blocking `read()` instead of `read1()`, deadline never starting, client-side type check removed, `_is_generation` accepting everything, the server-side parameter dropped, skipped observations not counted, the CLI line removed, cursor-loop detection removed, cursors never recorded, plus six #859 guarantees (secret in the 401 message, unscrubbed server text, redirects followed, size cap removed, pull module imported on the local path, non-atomic write).

```
ruff check src/soup_cli/ scripts/ tests/ benchmarks/      All checks passed!
pytest <865 + 859 + the two v0630 files>                  175 passed
pytest tests/ -q --tb=short   main 2a97512                21131 passed, 107 skipped, 0 failed   coverage 83.34%
pytest tests/ -q --tb=short   this branch                 21139 passed, 107 skipped, 0 failed   coverage 83.35%
```

Compared test-by-test against `main`: no existing test missing or changed outcome, identical skip set, +8 passed (all from the new file).

## Changelog

`changelog.d/0.74.0/881.fixed.md`


MakazhanAlpamys:
@Konuktor — reviewed at `ba0bc876`. **This is good work, and one measurement says so better than any adjective:** 5 of your 8 new tests FAIL against `origin/main`'s source, and the 3 that pass are exactly the three you labelled CONTROL. That is the signature of tests written first against a real defect, not tests written around behaviour that already existed. I ran that check specifically because a hardening PR following your own earlier feature is the easiest place in the world to accidentally ship a test suite that proves nothing.

**17 mutations, 17 killed**, including every inverted comparison: the deadline check, `read1` vs blocking `read`, the client-side GENERATION filter, the server-side `type` query parameter, the skipped counter, cursor recording, the cursor-loop comparison, the 64 MiB cap, the CLI summary line and the no-input guard. I also re-ran the redirect refusal from #859 to confirm this PR did not erode your earlier protections — still killed.

**Security answers, all measured rather than read.** `resolve_langfuse_host` reuses the shared `webhooks.validate_webhook_url` instead of rolling its own, which is the thing I most wanted to see: I threw 18 host spellings at it in both modes, including the four IPv4 bypasses v0.74.0 was patched for — `127.1`, `2130706433`, `0x7f000001`, `0177.0.0.1` — plus `169.254.169.254`, `[::ffff:127.0.0.1]`, `2852039166` and userinfo/query/fragment tricks. All rejected by default; accepted only under the explicit `--allow-private-host`, which is the documented intent. `http://localhost` is refused before the validator even runs, because HTTPS is required when a key pair rides every request.

Credentials: environment-only, so they never enter the audit-log argv; the `Authorization` header is built inline rather than held in a frame local; and `_scrub` strips the public key, the secret key and the base64 Basic token. I planted keys and ran both new code paths — the cursor-loop error and a mixed-type success — then scanned stdout, the exception repr and every file in the working directory for the pair, the token and 18-character prefixes. Zero hits. Worth noting you also kept the **cursor** out of the error message, which was the right call.

**Scope is exact**, with one deliberate improvement: #865 asked for the `type` re-check "in `_to_parser_row`", and you put it in the page loop instead. That is better and the reason is structural — `_to_parser_row` returns a row and has no way to signal "skip and count", so the loop placement is what makes the issue's own "reported as skipped" criterion achievable at all. `PullStats` is new surface, but it is the minimum vehicle for that count and it retired the `nonlocal pulled` closure on the way.

No test was deleted, skipped or loosened. I accounted for all four removed lines in `test_issue204_langfuse_pull.py`: one is a signature adaptation, and three are the same fixture swap from a single repeating page to `c1`/`c2`. That swap is **necessary rather than cosmetic** — `_Transport` replays its last response, so your new cursor-loop detector would have fired on the old fixture and the page-cap test would have silently started proving something else. Every assertion is retained and the page-cap one gained `pages == 2`.

## One change, and it is a regression this PR introduces

`ingest_pull.py:488-494`. `_read_capped` now raises `PullError`, which subclasses `RuntimeError`, so it escapes the `except (OSError, http.client.HTTPException)` in the `HTTPError` branch — the one whose comment says *"an error body that cannot be read in time must not turn a clean 'HTTP 401' into a raw traceback."*

Measured against a loopback 401 that drips its 18-byte body at a 0.4 s timeout: the clean `Langfuse rejected the key pair (HTTP 401); check LANGFUSE_PUBLIC_KEY...` becomes `the request to Langfuse timed out after 0.4s`. Exit code stays 1 and nothing leaks, so the harm is one lost diagnostic against a slow or adversarial server — but it is precisely the case the comment promises to handle, and before this PR a slow error body raised `socket.timeout`, an `OSError`, and *was* caught. Adding `PullError` to that tuple restores it.

## And one mechanical item caused by the release, not by you

`changelog.d/0.74.0/881.fixed.md` is correct today and the fragment guard passes 17/17. But once #879 tags v0.75.0 and empties `changelog.d/0.74.0/`, `test_repository_fragments_match_the_newest_release` turns CI red in every matrix cell. Please move it to `changelog.d/0.75.0/881.fixed.md` when you rebase. `main` is frozen until the tag, so there is no rush and nothing merges before then regardless.

## Two things I found and did NOT put on you

A Langfuse server's error body reaches the terminal with `\x1b[2J` and `\x1b[31m` intact — `_error_detail` collapses whitespace but not control bytes, and `rich.markup.escape` only neutralises `[...]`. That is from #859, outside #865's scope, and the right fix is the shared `utils/terminal.py:for_terminal` helper that does not exist on `main` yet (it lands with #879). Filed as **#885**; it is yours if you want it once the tag is out.

Your deadline is per request, while #865's *title* says total. Worst case with defaults is roughly 13 hours (6 attempts x 30 s plus 5 backoff pauses per page, x 100 pages), and days at `--max-pages 10000`. You disclosed this plainly in both the docstring and the PR body rather than implying more, which is why it is not a finding. If you think a total `--deadline` is worth having, say so and I will open an issue for it.

Docs are sufficient and precise — `docs/data.md` says *"Each request is bounded by"*, not "each pull", which is accurate about exactly the limitation above.

Push the `PullError` one-liner whenever suits you; the fragment move waits for the tag.

MakazhanAlpamys:
Confirmed at `78f0e56` — `PullError` is in the tuple, with a comment that says *why* it belongs there rather than just listing it, and `test_a_drip_fed_error_body_still_reports_its_status` pins the behaviour. That is the whole ask.

The fragment move is the only thing left, and it waits for the tag: `main` now has v0.75.0 merged (`f50a7da3`) but **not tagged**, and `changelog.d/0.74.0/` is gone from it. On your next rebase:

```
git mv changelog.d/0.74.0/881.fixed.md changelog.d/0.75.0/881.fixed.md
```

Nothing else changes. I will merge once the tag lands.

Separately, #885 — the pre-existing ESC-byte gap in `_error_detail` that I filed out of this review — is unblocked by the same event: `utils/terminal.py:for_terminal` now exists on `main`, so the shared helper is available. Still yours if you want it.

Konuktor:
Fixed and pushed.

- Added `PullError` to the HTTP-error body-read fallback. I added a regression
  test with a drip-fed 401 at a 0.4s deadline; removing `PullError` from the
  tuple reproduces the generic timeout, while the fix preserves the
  `HTTP 401` / key-pair diagnostic with no write or credential leak.
- Rebased onto v0.75.0 and moved the fragment to
  `changelog.d/0.75.0/881.fixed.md`.

I also checked the failed CI run: all nine matrix failures were only
`test_repository_fragments_match_the_newest_release`; there were no other
failures.

Local final:
- branch: 21,161 passed / 107 skipped / 0 failed
- current main: 21,152 passed / 107 skipped / 0 failed
- focused + #859 + changelog: 193 passed
- ruff clean
- 11/11 relevant mutations killed

The new CI run is in progress; Ubuntu 3.10, including the fragment guard,
is already green.

Thanks for catching the slow-error-body regression.

MakazhanAlpamys:
Heads-up: #893 just merged (`2b466960`) and it touches `src/soup_cli/commands/ingest.py`, so this branch will need a rebase.

It is a small change — two `console.print` lines inside `_pull` swapped from `escape(str(exc))` to `for_terminal(str(exc))` on the `PullError` branches, plus the import. It should not overlap your hunks in any meaningful way; if git does flag a conflict it will be adjacency, not intent.

One thing to keep when you rebase: leave the `except ValueError` branches on `escape`. That is deliberate rather than an oversight — the host validator refuses control bytes upstream, so nothing reachable gets through there, and #893's review measured that before deciding.

No action needed on the substance of this PR from my side yet; it is still in the queue.

MakazhanAlpamys:
Merged as `f42c998a`. Thank you — and what I want to name is that **all three findings came from the review of your own #859**, and you reproduced each one before fixing it rather than patching to the description.

I re-verified on the rebased head rather than trusting the earlier round, since the branch moved after #893:

```
tests/test_issue865_langfuse_pull_bounds.py + tests/test_issue204_langfuse_pull.py
  -> 120 passed, 1 skipped
ruff check src/soup_cli/ tests/  -> clean

mutation: _read_capped's `if time.monotonic() >= deadline:` -> `if False:`
  anchor counted: 1 occurrence
  ast.parse OK, module re-imported OK (so not a NameError masquerading as a catch)
  -> 2 failed, 7 passed
     TestDeadline::test_a_drip_feed_cannot_outlast_the_timeout
     TestDeadline::test_a_drip_fed_error_body_still_reports_its_status
```

Exactly the two deadline tests and nothing else — a precise pin. At the first review 5 of your 8 new tests failed against main's source and the 3 that passed were the controls, which is the shape I want and rarely get.

**The second finding is the one I would have missed.** A drip-fed *error* body hit the deadline and lost the status it was reporting, so a 401 surfaced as a timeout — an operator would have gone looking at their network instead of their credentials. Putting `PullError` in the body-read fallback with a comment saying *why* it belongs in that tuple, rather than just listing it, is what makes that survive the next person editing the file.

**And the honesty in the docstring is the part I want to point at for anyone reading this later:** you documented that a read already in flight can still overshoot by at most the socket timeout, instead of claiming the deadline is absolute. That is the true bound, and a fix that overstated it would have been worse than the bug it replaced.

The rebase also did the right thing with #893's change: `for_terminal` on the `PullError` branches, and the neighbouring `except ValueError` branches left on `escape`. That asymmetry looks like an oversight and is not — the host validator refuses control bytes upstream, so widening it would have added untested surface for no reachable defect. You kept it.

Your CONTRIBUTORS.md entry is updated, #865 is closed, and CI was green on all fourteen checks on the merged SHA.

PR #882, opened 2026-09-12
fix(mcp): validate data_validate's format against VALID_FORMATS and auto-detect (#878)

### LINKED ISSUE #878: soup mcp serve's data_validate still accepts any --format string and reports every row valid (the MCP half of #866)
#866 is fixed for the CLI by #869 (`aaf6a128`): `soup data validate --format bogus` now exits 1 instead of reporting every row valid. **The same input still reports every row valid through `soup mcp serve`**, so one dataset gets two different verdicts depending on which surface asks.

This is the CLI-vs-MCP divergence class already recorded in the v0.73.2 notes, where `ship_evidence` consumed an evidence file without reading a key the CLI read, and one file replayed to different verdicts through the two surfaces.

## What

`src/soup_cli/mcp_server/registry.py:233-239`:

```python
def tool_data_validate(args: dict) -> dict:
    from soup_cli.data.validator import validate_and_stats
    rows = _load_data_rows(_require_str(args, "data"))
    fmt = _opt_str(args, "format")
    return validate_and_stats(rows, expected_format=fmt)
```

`_opt_str` caps length and rejects control characters but does not constrain the value, and `validator.py:95` is:

```python
check_format = bool(expected_format and expected_format in VALID_FORMATS)
```

So an unrecognised format silently turns off the format check, which is what #866 was about. Reproduced against a two-row file: `tool_data_validate({"data": ..., "format": "bogus"})` returns `{"total": 2, "valid_rows": 2}`.

## And a second, pre-existing bug in the same call

The tool's schema describes `format` as *"omit to auto-detect"*. Omitting it yields `expected_format=None`, so `check_format` is `False` and **no detection happens at all** — the same 2/2-valid answer. There is no auto-detect on this path; the description promises behaviour the code does not implement.

## Suggested fix

Reuse what #869 landed rather than writing a second guard: validate `format` against `soup_cli.data.formats.VALID_FORMATS` and raise `McpToolError` for anything else, and make an omitted `format` actually run `detect_format` so the schema's promise holds — or change the description if auto-detect is not wanted here.

Worth checking the neighbours in the same pass: several other handlers in `registry.py` take free-form strings that reach a function with a silent-fallback default. If any of them has the same shape, name them in the PR rather than fixing one and leaving the class open.

## Acceptance criteria

- [ ] `tool_data_validate` with an unrecognised `format` raises `McpToolError` rather than returning a report
- [ ] The error names the accepted values, as the CLI's does
- [ ] An omitted `format` either auto-detects (matching the schema description) or the description is corrected
- [ ] The allowlist is `VALID_FORMATS`, not a second hand-written tuple
- [ ] A test asserting the CLI and the MCP tool agree on the same file and the same format argument — that is the property that broke, and it is the one worth pinning

Srinivasan8888:
Closes #878.

## The divergence, measured

Two valid alpaca rows, on `2a97512`:

| call | MCP `tool_data_validate` | `soup data validate` |
|---|---|---|
| `format="bogus"` | `total=2, valid_rows=2` | **exit 1**, `Unknown --format: 'bogus'` |
| `format` omitted | `total=2, valid_rows=2`, nothing detected | **auto-detects `alpaca`**, exit 0 |
| `format="sharegpt"` (real, wrong) | `total=2, valid_rows=0` | `0/2 rows valid` |

The third row is the tell: the handler is not broken in general, it is broken specifically for values that fall outside the allowlist. `validator.py:95` is

```python
check_format = bool(expected_format and expected_format in VALID_FORMATS)
```

so an unrecognised format silently turns the check **off** and `None` never turns it on — both produce "every row valid" for a reason unrelated to the data. #869 guarded the CLI before it reached that line; this surface passed the raw value through.

## The fix is its own sibling

`tool_data_doctor`, five lines below in the same file, already resolves `format` correctly — rejects bogus values, auto-detects when absent. This makes `tool_data_validate` match it, so there is no new pattern to review:

```python
fmt = _opt_str(args, "format")
fmt = "auto" if fmt is None else fmt
if fmt != "auto" and fmt not in _formats.VALID_FORMATS:
    raise McpToolError(
        f"unknown format {fmt!r}; accepted: auto, "
        + ", ".join(sorted(_formats.VALID_FORMATS))
    )
if fmt == "auto":
    try:
        fmt = _formats.detect_format(rows)
    except ValueError as exc:
        raise McpToolError("could not auto-detect data format; pass 'format'") from exc
return {**validate_and_stats(rows, expected_format=fmt), "format": fmt}
```

Three deliberate choices:

- **The allowlist is `VALID_FORMATS` itself**, not a copy — a hand-written second tuple is how these two surfaces drifted apart in the first place. A test monkeypatches a new member into the set and asserts the handler accepts it with no code change, so a copy would fail.
- **The resolved format is returned.** Without it a caller cannot tell which format was checked, which leaves the schema's "omit to auto-detect" unobservable even once it is true. The schema description is updated to say so.
- **A blank string is refused rather than read as omitted**, because `soup data validate --format ""` exits 1 and parity is the point of the issue. `tool_data_doctor` still spells this `or "auto"` and so reads blank as omitted — noted rather than changed, since that is a second surface and not what #878 is about. Say the word if you want them unified.

## Neighbour sweep

You asked for the class, not the instance. Every `_opt_str` value in `registry.py` that reaches a function with a silent-fallback default:

| handler | verdict |
|---|---|
| `tool_data_validate` | **the bug** — fixed here |
| `tool_data_doctor` | already correct: bogus format raises `McpToolError`, omitted auto-detects |
| `tool_recipes_search` | correct: unknown `task`/`size` yields 0 results, not an unfiltered list |
| `_resolve_gpu_memory_mcp` | correct: raises `McpToolError` naming where to find valid options |

So `tool_data_validate` was the only handler of that shape. Correcting myself in public: my first pass reported `tool_recipes_search` as returning results for a nonsense task — I had taken `len()` of the response dict rather than of the recipe list. Re-measured, it returns 0 of 167.

## Mutation table

7 run, 7 killed. `__pycache__` cleared between runs, restore re-verified at 19 passed.

| Mutation | Result | First killer |
|---|---|---|
| the unknown-format guard removed entirely | **killed** | `test_it_raises_rather_than_reporting_every_row_valid` (+8) |
| allowlist becomes a hand-written copy | **killed** | `test_every_valid_format_is_accepted` (+1) |
| error message no longer lists accepted values | **killed** | `test_the_error_names_the_accepted_values` |
| auto-detect removed (omitted goes back to no check) | **killed** | `test_the_detected_format_is_reported_back` (+2) |
| resolved format no longer returned | **killed** | `test_the_detected_format_is_reported_back` (+1) |
| blank format reads as omitted again | **killed** | `test_near_misses_are_refused_too[]` |
| undetectable data passes silently | **killed** | `test_undetectable_data_is_refused_not_silently_passed` |

## Acceptance criteria

- [x] unrecognised `format` raises `McpToolError` rather than returning a report
- [x] the error names the accepted values, as the CLI's does
- [x] omitted `format` auto-detects, matching the schema description (and the description is updated to name `auto` and the returned field)
- [x] the allowlist is `VALID_FORMATS`, not a second hand-written tuple — pinned by a test
- [x] a test asserting the CLI and the MCP tool agree on the same file and the same format argument, over `alpaca`/`sharegpt`/`chatml` so it covers the formats the rows *fail* as well as the one they satisfy

## Verification

```
pytest tests/test_issue878_mcp_data_validate_format.py   19 passed
pytest tests/ -k "mcp or data_validate or validator"     266 passed
ruff check src/soup_cli/ tests/                          All checks passed
git diff -- tests/ | grep -c "^-[^-]"                    0
```

The full suite (~14 min) is running and I will post the result; nothing in this change is outside the two suites above, but I would rather state the number than assume it.

## Scope

Branched off `2a97512`. No changelog fragment in this commit — the guard allows one fragment per number and the number has to exist first, so it follows in the next push. Noting the freeze: this is not expected to merge before v0.75.0 is tagged, and I will move the fragment if the tag empties `changelog.d/0.74.0/`.


Srinivasan8888:
Rebased onto `v0.75.0` and moved the changelog fragment to `changelog.d/0.75.0/882.fixed.md`.

The tag emptied `changelog.d/` (#879 merged at 08:26Z, deletions-only across all 48 fragments), so the fragment I pushed twenty minutes ago pointed at a directory that no longer exists on `main`. The guard says so precisely, which is how I checked rather than assuming:

```
ChangelogError: changelog.d/0.74.0/882.fixed.md: fragment baseline is 0.74.0,
```

The body already references its own number, so nothing inside the file changed. Re-verified on the new base: `test_issue487_changelog_fragments.py` 17 passed, this PR's suite 19 passed, and `-k "mcp or data_validate or validator"` 272 passed (up from 266 — the release added tests in that range, none of them mine).

Nothing else in this PR moved with the release; `VALID_FORMATS`, `detect_format` and `validate_and_stats` are unchanged in v0.75.0.

MakazhanAlpamys:
@Srinivasan8888 — reviewed at `8416c0ba`. **Merge as-is.** Nothing to change; it goes in the moment v0.75.0 is tagged.

## Both halves verified by execution

Unknown format is refused and the refusal *names the accepted set* — which matters more here than on the CLI, since an MCP client is usually a model choosing arguments, and a refusal that does not name the valid values is the difference between a self-correcting call and a stuck one:

```
tool_data_validate({"data": ..., "format": "bogus"})
→ McpToolError: unknown format 'bogus'; accepted: auto, alpaca, asr, ...
```

And the second half — the one the issue only found because the first was being checked — genuinely auto-detects now rather than silently disabling the check:

```
format omitted → {"total": 2, "valid_rows": 2, "format": "alpaca"}
```

The control is what proves it is really *on*: the same rows under `format="sharegpt"` return 0/2 with per-row failures. Undetectable data now refuses instead of reporting 1/1 valid.

## The cross-surface property, which is the reason #878 existed

I drove eight inputs through **both** surfaces and they agree in every direction: bogus refused / refused, omitted 2/2 both, `""` refused / refused, `alpaca` 2/2 both, `sharegpt` 0/2 both, undetectable refused / refused.

**And there is exactly one allowlist**, which I checked at both call sites rather than assuming: `commands/data.py` and `registry.py` both read `soup_cli.data.formats.VALID_FORMATS`. Your test that monkeypatches a new member in and asserts acceptance with no code change is what keeps it that way — mutation M4 (swap in a hand-written copy that accepts `"bogus"`) fails 5 tests, so that guard is load-bearing rather than decorative. Given the v0.73.2 history of one file yielding two verdicts through CLI and MCP, that is the right thing to have pinned.

Five mutations, five killed: guard removed, guard inverted, omitted-format back to disabling the check, allowlist replaced by a copy, and the resolved `format` key dropped. Zero test lines removed. 19 passed in the new file, 188 passed across the MCP surface, ruff clean. Your error message is bounded and safe — I checked: `_opt_str` caps at 4096 first, `!r` textualises control bytes so an ESC arrives as literal backslash-x-1b, and `server.py` sanitises again on the way out.

Your two widenings — returning the resolved `format`, and refusing a blank string — are both argued in the body and both test-covered, and the first is what makes the auto-detect criterion observable at all. Fine by me.

**And you put the fragment in `changelog.d/0.75.0/` unprompted**, before I had said a word about it. Sixteen other open PRs are currently tripping over that exact directory.

## One thing I turned into an issue rather than asking you to fold in

Your neighbour sweep found that `tool_data_doctor` still spells the argument `or "auto"`, so a blank format is accepted there while it is now refused here — and its unknown-format refusal reads `data doctor failed (ValueError)` without naming the value or the accepted set. I verified both against your head.

You disclosed it accurately and offered to unify, which is the right call and the right scope discipline: folding it in would have widened a PR that is currently exact. Filed as **#889** and offered to you. Worth extending the sweep to every tool in `registry.py` taking a format-like argument, not only the validate path.

## One small note on a test claim

`test_both_reach_the_same_valid_row_count` calls `validate_and_stats(...)` as the "CLI equivalent". That is the shared core, not the CLI, so it pins MCP-against-core rather than MCP-against-CLI — a CLI-side divergence in row counting would slip past it. `test_both_refuse_an_unrecognised_format` does drive the real CLI, so the refusal case is genuinely pinned. I verified row-count parity by running both surfaces myself, so the behaviour is right; the test just claims a little more than it proves. Not worth a round trip — mentioning it so the name does not mislead a future reader.

**Timing:** `main` has v0.75.0 merged but **not tagged**, so nothing merges until the tag is out. Your checks are still running; I will merge on green after the tag.

MakazhanAlpamys:
Merged as 9d43b50. Thank you — this one is a good example of fixing the *class* rather than the symptom.

What I particularly liked:

- **You pinned the agreement between the two surfaces, not an exit code.** `TestTheTwoSurfacesAgree` is the test that keeps this closed: the bug was never "the MCP tool returns the wrong number", it was "the same file gets two verdicts depending on who asks", and that is the property that can regress again.
- **The allowlist is `VALID_FORMATS` itself, not a second tuple.** A hand-written copy is precisely how the two surfaces drifted apart in the first place, and you said so in the comment rather than leaving the next reader to rediscover it.
- **Returning the resolved format** is what makes the schema's "omit to auto-detect" *observable* rather than merely true. Without it a caller cannot tell which format was checked, so the promise is unverifiable from the outside.
- **The near-miss parametrize** (`""`, `"  "`, `"ALPACA"`, `"alpaca "`) is the half that matters. An implementation that helpfully normalises case or strips whitespace passes a naive test and fails yours.
- **Scope held.** `tool_data_doctor`'s neighbouring `or "auto"` blank-string reading was noted in a comment and filed as #889 instead of being swept in here. That is the right call.

What I ran before merging, on head `8416c0b`:

- `pytest tests/test_issue878_mcp_data_validate_format.py` — **19 passed**.
- **Mutation** (the one that decides it): I removed the unknown-format guard from `tool_data_validate`, restoring the pass-through that `validate_and_stats` silently interprets as "format check off". **9 of 19 tests failed**, including `TestTheTwoSurfacesAgree::test_both_refuse_an_unrecognised_format` and the `csv` / `json` near-miss cases. The tests are load-bearing, not decorative.
- All 14 required checks green on the current head (lint, type-check, mlx-smoke, pytorch-smoke, transformers-floor, and the full 3x3 OS/Python matrix).

The fragment was already in `changelog.d/0.75.0/` — you moved it yourself after the release without being asked, which saved a round trip. Credited in CONTRIBUTORS.md.

Issue #875, opened 2026-09-12
soup doctor says "Training stack not installed" with six of seven [train] packages present and OK


MakazhanAlpamys:
`soup doctor` treats the `[train]` group as all-or-nothing. `group_missing` is set as soon as **any** member is absent (`src/soup_cli/commands/doctor.py:215-220`), and the message is unconditional:

```
Training stack not installed: pip install "soup-cli[train]"
```

With six of the seven present and `OK`, the table says so and the advice contradicts it. The stack *is* installed; one package is not.

## Reproduce

```
pip uninstall -y bitsandbytes
soup doctor
```

The table shows `torch OK`, `transformers OK`, `peft OK`, `trl OK`, `datasets OK`, `accelerate OK`, `bitsandbytes not installed` — and then reports the training stack as not installed.

## Why it is worth fixing rather than leaving

This is the state a user is actually in when something broke: one wheel failed to build, or `pip install vllm` pulled a package out. The advice they get names the whole group rather than the one thing missing, and `pip install "soup-cli[train]"` is a no-op for the six already satisfied — so the message is both wrong and unhelpful at the same moment the user most needs it to be right.

## Suggested fix

Say what is missing. Roughly:

- none present -> `Training stack not installed: pip install "soup-cli[train]"` (today's message, which is correct in that case)
- some present -> name the absent members, e.g. `Training stack incomplete, missing: bitsandbytes` with an install line for the group

Note a wrong alternative that is easy to reach for: gating the existing message on `present == 0` and otherwise emitting nothing. That produces `All checks passed!` next to a `To fix:` line in the same output — a contradiction, which is worse than an over-statement. The correct output is a third thing.

The exit code must not change here: a partially installed `[train]` group stays exit 0, per #828. Beyond-ceiling versions are a separate question, tracked in #874.

## Acceptance criteria

- [ ] With some but not all `[train]` members installed, the advice names the missing ones and does not claim the stack is absent
- [ ] With none installed, the message is unchanged
- [ ] The "all checks passed" summary and a `To fix:` block can never appear in the same run
- [ ] Exit code stays 0 in both cases
- [ ] A test that fails against current `main`

blackcoderx:
I am claiming this.

MakazhanAlpamys:
Fixed in #884 by **@HarshRajSinghania**, merged as `6b879b87`.

[...]

Issue #829, opened 2026-09-11
soup doctor shows only GPU name and memory: no compute capability, no check that the torch build supports the card, no bf16/fp8/nvfp4 availability


MakazhanAlpamys:
## The defect

The GPU panel (`commands/doctor.py:282-299`) prints three facts: that CUDA is available, `torch.version.cuda`, and per-device name + total memory. It shows nothing that decides whether a given card can run a given Soup feature.

## Reproduction

On an RTX 5070 Laptop (sm_120, driver 616.92, torch 2.14.0+cu130):

```

$ COLUMNS=100 soup doctor

┌────────────────────────────────────────────── GPU ──────────────────────────────────────────────┐

│ CUDA:     available (v13.0)                                                                     │

│ GPUs:     1                                                                                     │

│   GPU 0: NVIDIA GeForce RTX 5070 Laptop GPU (7.9 GB)                                            │

└─────────────────────────────────────────────────────────────────────────────────────────────────┘

```

The facts that matter, queried directly from torch on the same box:

```

capability (12, 0)

arch_list ['sm_75', 'sm_80', 'sm_86', 'sm_90', 'sm_100', 'sm_120']

bf16 True

```

Soup's own gates on the same box (`utils/fp8.py:46`, `utils/advanced_precision.py:65`), none of which doctor shows:

```

is_fp8_gpu_supported True

is_fp8_available False

is_blackwell_gpu True

torchao False

```

## Evidence

`get_arch_list` is not called anywhere in `src/soup_cli`, and `commands/doctor.py` never mentions compute capability, bf16, fp8 or nvfp4.

## Why it matters

- **A torch build that cannot run the card still reads as green.** With a wheel whose `get_arch_list()` lacks the card's `sm_XY` (e.g. a pre-cu128 build on an RTX 50-series), `torch.cuda.is_available()` is still True, so doctor prints "available". The first CUDA kernel then fails with "no kernel image is available". The check is one line (`f"sm_{major}{minor}" in torch.cuda.get_arch_list()`, allowing for PTX forward-compat entries), and doctor is the command that should run it.

[...]

PR #883, opened 2026-09-12
fix(doctor): handle modern CUDA driver versions

Samearth17:
## Summary

Fix `soup doctor` CUDA detection and PyTorch wheel recommendations for modern NVIDIA drivers.

## Changes

- Parse both `CUDA Version: X.Y` and `CUDA UMD Version: X.Y` from `nvidia-smi`.
- Replace the unsafe `None -> cu121` fallback with an explicit supported CUDA index table.
- Avoid recommending a CUDA wheel when the driver version cannot be determined.
- Keep the Torch recommendation compatible with Soup's `torch>=2.6.0` training floor.
- Use `--force-reinstall "torch>=2.6.0"` in the NVIDIA installation suggestion.
- Distinguish an installed CUDA Torch build that cannot initialise CUDA from a genuine CPU-only Torch build.
- Add regression coverage for the newer `nvidia-smi` header and CUDA diagnostic behaviour.

## Validation

- `tests/test_v0401_part_c.py`: 21 passed
- `tests/test_doctor.py`: 30 passed
- `ruff check src/soup_cli/commands/doctor.py tests/test_v0401_part_c.py`
- `python -m py_compile src/soup_cli/commands/doctor.py tests/test_v0401_part_c.py`
- `git diff --check`

Refs #827


MakazhanAlpamys:
@Samearth17 — reviewed on the box this was filed from: RTX 5070 Laptop, driver **616.92**, CUDA 13.4, torch 2.14.0+cu130. **The substance is a real fix**, measured rather than argued, and your sequencing on top of #854 was respected. One blocking item, two housekeeping.

## The fix works on real hardware

`nvidia-smi` here prints `KMD Version: 616.92   CUDA UMD Version: 13.4` — the header `main` cannot read:

| | `origin/main` | your PR |
|---|---|---|
| `_nvidia_smi_cuda_version()` | **None** | **(13, 4)** |
| wheel tag | `cu121` | `cu132` |

And I checked the recommendation is actually reachable rather than merely newer — `download.pytorch.org/whl/cu132/torch/` returns 200 with `torch 2.14.0+cu132`, while the `cu121` `main` recommends tops out at `2.5.1`, below the `>=2.6.0` floor. So the old advice was a dead end on this card and yours is not.

I enumerated 21 header spellings. 17 parse correctly, 4 refuse safely: three- and four-part CUDA versions, a leading-zero minor (`12.05` → `(12,5)`, no crash), a three-digit major, an `r12.8` suffix, tab and newline after the colon, `N/A`, empty, nvcc-style output, and a `KMD Version:` line alone with no false match. Eight of eight mutations of the parsing logic are caught by your tests — including the two silent-wrong ones, swapping major/minor and dropping the minor. That is the part I most wanted to see.

## Blocking: the second call site still assumes a string

You widened `_torch_cuda_wheel_tag` to `str | None`, and `doctor.py:239` guards only `driver is not None` — which is no longer sufficient, because the tag is now `None` for any driver reporting **CUDA 11.0 through 11.7**. Measured on your branch:

```
driver (11, 0) -> tag None   url .../whl/None
driver (11, 4) -> tag None   url .../whl/None
driver (11, 7) -> tag None   url .../whl/None
driver (11, 8) -> tag 'cu118'
```

So with the `[train]` stack absent on a 515.x-era driver, `soup doctor` prints:

```
pip install torch --index-url https://download.pytorch.org/whl/None
```

which 403s. `main` prints `cu118` there — a live index carrying `torch 2.7.1+cu118`. **That is a regression on a path your change did not intend to touch**, and it shows the literal word `None` to the user.

It is not covered in either direction: I planted two controls at that line — one fixing it with `or "cu118"`, one making it emit `whl/BOGUS` — and **both survive the whole suite**. Your own test pins the *enabling* half (`assert _torch_cuda_wheel_tag((11, 0)) is None`) without reaching the consumer.

Fix: fall back to the plain `pip install "soup-cli[train]"` line when `tag is None`, mirroring the existing `driver is None` branch, plus a test.

**Heads-up you will want:** #884 (@HarshRajSinghania, for #875) **duplicates that call site** while splitting the all-missing and partial-missing branches. All three open doctor PRs auto-merge cleanly, but after #883 and #884 both land this bug exists at *two* sites. I verified the three-way merge by actually performing it rather than trusting the absence of conflict markers — #877's line survives correctly and each PR's own tests still pass, so the composition is fine; it is only this one guard that needs adding wherever it ends up.

## Two housekeeping items

**Rebase and add a fragment.** You are one commit behind `origin/main` — v0.75.0 merged and emptied `changelog.d/0.74.0/`. A user-visible doctor output change warrants `changelog.d/0.75.0/883.fixed.md`, PR-numbered (not `827`), crediting you.

**Widen the regex whitespace.** `r"CUDA(?: UMD)? Version:"` hardcodes single literal spaces, so `CUDA  UMD  Version: 13.4` returns `None`. This is the same class as the `\boxed{C}` / `\boxed {C}` miss the repo wrote its parsing rule about — one space to the left of a correct fix. `r"CUDA(?:\s+UMD)?\s+Version:\s*(\d+)\.(\d+)"` costs nothing.

## Smaller notes, none blocking

`tests/test_v0401_part_c.py:266` has `"...GPU\n"` — a literal backslash-n, not a newline — so that fixture never exercises the `.splitlines()[:1]` path it appears to test. It passes because nothing asserts on the label.

With both spellings present in one header, `re.search` returns whichever appears first, so the answer is positional rather than authoritative. No real header does that today; one comment would do.

The `torch>=2.6.0` floor at `doctor.py:587` is a hardcoded string duplicating `pyproject.toml`. `test_issue636_torch_floor.py` pins the pyproject value but does not reach this literal, so a future bump leaves doctor advising a stale pin.

`docs/backends-and-ops.md:1124` says doctor "prints the `pip install` command for the wheel your driver can run" — after this PR it may print none at all. And your new "could not be confirmed" message says *"Run `nvidia-smi`"* inside a function that just ran it successfully; the user's problem is that the header did not parse.

One scope note for the body rather than the code: the #827 criterion *"on an sm_120 card doctor never recommends cu121/cu124/cu126"* is satisfied **incidentally** — because Blackwell drivers never report CUDA < 12.8 — not because anything queries the compute capability. That is fine for this PR (#829 is the compute-cap issue), but worth saying rather than implying it is structural.

Four ANSI/width failures in `tests/test_doctor.py` under `FORCE_COLOR=1 COLUMNS=60` are **pre-existing** — the same four fail on `origin/main`, and your two new tests assert on a plain return value, so they are immune. Not yours; that file is a good candidate for the sweep #886 is about.

`main` is frozen until v0.75.0 is tagged, so nothing merges today regardless. Push the guard and I will re-run the controls.

MakazhanAlpamys:
@Samearth17 — reviewed at `26f25c9d`. **The index table is empirically right and I verified every row of it**, which is the hard part of this PR. One blocking defect, and it is a call site rather than your logic.

## What I verified

I fetched every wheel index rather than taking the table on trust: `cu118` -> 2.7.1, `cu124` -> 2.6.0, `cu126` / `cu130` / `cu132` -> 2.14.0, all at or above the `torch>=2.6.0` floor — and `cu121` -> **2.5.1**, genuinely below it, which is why excluding it is correct rather than arbitrary. #827's criteria 1, 3, 4 and 5 are met.

**51 passed** on `test_v0401_part_c.py` + `test_doctor.py`; **144 passed, 1 skipped** across the broader doctor surface. Six deleted assertion lines, all deliberate inversions consistent with the new intent — no weakening. **Three mutations, all killed:** reverting the UMD regex fails 1, reverting the wheel tag to main's always-`str` fails 2, flattening `build_status` and dropping `--force-reinstall` fails 1.

## BLOCKING — `https://download.pytorch.org/whl/None` for every driver below CUDA 11.8

You changed the signature to `-> str | None` and guarded the second call site (`_detect_gpu_hw_without_torch_cuda`, the `if wheel is None:` at `:562`). The **first** one, `doctor.py:238-240`, guards the wrong condition:

```python
if driver is not None:                 # (11, 7) passes this
    tag = _torch_cuda_wheel_tag(driver)  # ...and this returns None
    url = f"https://download.pytorch.org/whl/{tag}"   # -> whl/None
```

Measured on your head:

```
(11, 0) -> None      (11, 8) -> 'cu118'
(11, 4) -> None      (12, 8) -> 'cu128'
(11, 7) -> None      (10, 2) -> None
EMITTED URL: https://download.pytorch.org/whl/None
```

That URL is an HTTP **403**, so a user on an older driver — exactly the user this advice exists for — is told to run a command that fails with a permissions-shaped error. And it exits 0.

**All 51 of your tests pass with this live.** Nothing covers it, which is the part I would like fixed alongside the code: guard on the *tag*, not the driver, and fall back to the plain `pip install "soup-cli[train]"` line.

```python
tag = _torch_cuda_wheel_tag(driver)
if tag is not None:
    ...two-step advice...
else:
    ...plain [train] line...
```

Please add a test at a driver below 11.8 asserting the emitted text contains no `whl/None`.

## Two mechanical items

**Rebase.** You branched from `2a97512a`, before the v0.75.0 release commit, so the branch carries 48 `changelog.d/0.74.0/*` fragments. I test-merged: git keeps main's deletion and they do **not** come back, so this is not dangerous — but rebase anyway so the diff shows what you actually changed.

**Add `changelog.d/0.75.0/883.fixed.md`.** This is user-visible.

## Two things that are not change requests

**Criterion 2 of #827 is only partly met** and I am not holding the PR for it: the choice is made from the driver's CUDA version alone and never reads `compute_cap`, so an sm_120 card behind a driver reporting CUDA 12.6 still gets `cu126`. Whether that combination occurs in practice I could not test — I have exactly one Blackwell card and one driver. Say so in the PR body and it can be a follow-up.

**#827 asks for a README `--index-url` step, and you should not add one.** `test_readme_does_not_hardcode_a_cuda_wheel_index` forbids `download.pytorch.org/whl/cu` in README. The issue's suggestion contradicts an existing guard — my error in writing it, not yours.

## One thing you should know about the neighbourhood

**Your change is what makes a latent defect in #884 live.** #884 (also for `soup doctor`) adds a *third* unguarded `_torch_cuda_wheel_tag` call site while the function still returns `str` on its own branch. I test-merged both: **git reports no conflict**, and the result emits `whl/None` at **two** sites. Nothing would have warned either of you. I am merging #888 instead of #884 for #875, partly for that reason — but this PR needs the guard regardless of which one lands.

Fix the call site with a test, rebase, add the fragment, and this is ready.

MakazhanAlpamys:
@Samearth17 — status check, and a pointer to the fuller note I left on #857 so I am not splitting the conversation.

One item here, unchanged from my review: **the `whl/None` call site.** The reason it is blocking rather than cosmetic is that all 52 of your tests pass **with the bug live** — the guard sits on the driver rather than on the tag, so any driver below CUDA 11.8 produces `https://download.pytorch.org/whl/None`, which returns 403. A green suite over a live bug is precisely the shape this project keeps having to catch, and it is the reason the fix needs a test that fails first.

Heads-up on traffic: `commands/doctor.py` currently has **six** open PRs against it, and #884 is verified and waiting only on the Actions queue. Whichever of those lands first, you will want a rebase check before this goes green again.

Still on it? If it has gone cold, say so and I will record the reason rather than let it age out.

MakazhanAlpamys:
Reviewed the 14:21 push at `66f40bf0`. **The blocker is closed** — and then something happened underneath you that you could not have known about, so read past the first section.

The guard is on the tag now (`doctor.py:240`, `if tag is not None:`), and I reproduced both directions through the real path with a driver below CUDA 11.8:

```
before: ['https://download.pytorch.org/whl/None', 'https://download.pytorch.org/whl/None']  (also in "Fix all:")
after : []  ->  Training stack not installed: pip install "soup-cli[train]"
```

Mutation: reverting `if tag is not None:` to `if driver is not None:` (anchor counted exactly once, module really imported, `inspect.getsource` confirming the mutant is live) gives **1 failed, 51 passed** on `test_doctor_missing_train_extra_with_unsupported_driver_falls_back_to_extra`, on `assert "download.pytorch.org/whl/" not in out`. The 51/52 arithmetic is the whole point of my earlier note: this pair of files was 51 green **with the bug live**, and your push adds the one test that catches it.

---

**The part that changes what happens next: #884 merged this morning (`6b879b87`) and already fixed this defect on `main`, differently.**

```
CONFLICT (content): Merge conflict in src/soup_cli/commands/doctor.py
CONFLICT (content): Merge conflict in tests/test_doctor.py
```

Your base is 39 commits behind. Main now reads:

```python
tag = (_torch_cuda_wheel_tag(driver) if driver is not None and torch_missing else None)
url = f"https://download.pytorch.org/whl/{tag}" if tag else None
```

— i.e. already guarded on the tag.

**And there is a detail here I owe you plainly, because it changes whose bug it was.** On `main`, `_torch_cuda_wheel_tag` still returns a `str` and never `None` (`"cu118"` for old drivers, `"cu121"` when unreadable), so **`whl/None` is unreachable there**. The defect I blocked you on was introduced by this PR's own first commit, which changed the return type to `str | None` without moving the guard. That is not a criticism of the patch — widening the return type is the right modelling and the rest of the PR depends on it — but I described it as a pre-existing bug you had failed to guard, and it was not. My error in framing, and I would rather say so than let it sit in the thread.

The conflict is not mechanical. Both sides restructure the same block, so on rebase you will need to decide how `_SUPPORTED_TORCH_CUDA_WHEELS` — the real substance of this PR — sits on top of #884's `torch_missing` / partial-stack logic. #947 is also open on this file and also predates #884, so all three touch the same lines; I am reviewing that one separately and will keep the ordering straight.

---

**Three things while you rebase:**

1. **The changelog fragment has no credit and no issue number.** It is in the right directory (`changelog.d/0.75.0/883.fixed.md`, one file) but reads as a bare sentence. House style on main is `- **<bold lead> (#<issue> by @<handle> in #<PR>).** <detail>` — see `882.fixed.md`, which is yours.
2. **Your new test stubs the function under test.** It monkeypatches both `_nvidia_smi_cuda_version` **and** `_torch_cuda_wheel_tag`, so it never exercises the real `_SUPPORTED_TORCH_CUDA_WHEELS` table. It still catches the mutation, which is what I asked for — but nothing pins that a real driver like `(11, 7)` actually yields `None`. I checked the real table does; the test just does not. Drop the second monkeypatch.
3. **Two user-visible behaviour changes are undocumented.** `_torch_cuda_wheel_tag(None)` returned `"cu121"` on main and now returns `None`, and drivers 12.1-12.3 now get `cu118` rather than `cu121` (`cu121` is in `_TORCH_CUDA_WHEELS` but excluded from `_SUPPORTED_TORCH_CUDA_WHEELS`). Both look deliberate and defensible — say so in the fragment.

MakazhanAlpamys:
@Samearth17 — your CI is red and it is one character. Nothing to do with the driver logic.

```
tests/test_issue487_changelog_fragments.py::test_repository_fragments_match_the_newest_release FAILED
ChangelogError: changelog.d/0.75.0/883.fixed.md: fragment must start with a Markdown list item (`- `)
```

Nine cells, one line, and the rule is `scripts/assemble_changelog.py:116` checking `content.startswith("- ")`. Two things to check while you are in there, because a UTF-8 BOM also counts as the first character and produces the identical error with the `- ` visibly present:

```
head -c 3 changelog.d/0.75.0/883.fixed.md | xxd     # want 2d 20 …, not efbb bf
```

This is the **fifth** PR to hit that rule today. That is a process failure rather than five people's carelessness — the cost of learning it is a full matrix run in a queue that is currently 97 deep, so you find out hours later about a one-character problem. I am looking at catching it before push instead; it is not on you.

**Three things from my 2026-09-13 review are still open**, and I would rather restate them than have you rediscover them after the fragment fix:

1. **The fragment has no credit and no issue number.** While you are editing it: house form is `- **<bold lead> (#<issue> by @<handle> in #<PR>).** <detail>` — see `882.fixed.md`, which is yours.
2. **The new test stubs the function under test.** It monkeypatches both `_nvidia_smi_cuda_version` **and** `_torch_cuda_wheel_tag`, so it never exercises the real `_SUPPORTED_TORCH_CUDA_WHEELS` table. It still catches the mutation I asked for, which is what I required — but nothing pins that a real driver like `(11, 7)` actually yields `None`. I checked the table does; the test just does not. Dropping the second monkeypatch fixes it.
3. **Two user-visible behaviour changes are undocumented.** `_torch_cuda_wheel_tag(None)` returned `"cu121"` on main and now returns `None`, and drivers 12.1-12.3 now get `cu118` rather than `cu121` (`cu121` is in `_TORCH_CUDA_WHEELS` but excluded from `_SUPPORTED_TORCH_CUDA_WHEELS`). Both look deliberate and defensible — say so in the fragment.

And the branch **conflicts with `main`**, in `commands/doctor.py` and `tests/test_doctor.py`, from `7c68bd85` (#911) landing. The conflict is not mechanical: both sides restructure the same block, so you will need to decide how `_SUPPORTED_TORCH_CUDA_WHEELS` — the real substance here — sits on top of #884's `torch_missing` / partial-stack logic. #947 is also open on this file; I am keeping the ordering straight and will tell you if it changes.

The correction I owe you from last time stands unchanged: the `whl/None` defect I blocked you on was introduced by this PR's own first commit widening the return type, not a pre-existing bug you failed to guard. That was my error in framing, not yours in code.

MakazhanAlpamys:
@Samearth17 — reviewed `3bf23829`. **All five restated findings closed; merge-ready, waiting only on CI.** One thing you should know about your *other* PR is at the bottom, because it affects merge order.

## The findings

| # | status |
|---|---|
| fragment must start with `- ` (a BOM also counts) | closed — `head -c16 \| xxd` → `2d20 2a2a`, no BOM |
| fragment credit + issue number | closed — `(#827 by @Samearth17 in #883)` |
| the new test stubbed the function under test | **closed, and provably** — see the mutation |
| two undocumented behaviour changes | closed — the fragment states both verbatim |
| conflicts with `main` from #911 | closed — `merge-tree` clean; you kept main's `tag = (... if driver is not None and torch_missing else None)` and dropped your own guard hunk, which is the correct resolution |
| `r"CUDA(?: UMD)? Version:"` hardcoding single spaces | closed — `CUDA  UMD  Version: 13.4` and a tab-separated variant both parse to `(13, 4)`; `KMD Version:` and `N/A` both give `None` |

**No test was weakened.** The removed lines are byte-identical to the set already accepted on 2026-09-12 as deliberate inversions; this delta removes nothing new.

## Measured on the hardware this is actually about

This box is the case your PR exists for — an RTX 5070 Laptop on driver 616.92 reporting `CUDA UMD Version: 13.4`:

```
main       :  driver = None      tag = cu121     <- wrong wheel, silently
this PR    :  driver = (13, 4)   tag = cu132
```

`soup doctor` on the merged tree prints `sm_120`, `CUDA: available (v13.0)`, `torch 2.14.0+cu130 >=2.6.0 OK`. So the fix is live on real modern hardware, not only in fixtures.

**Mutation, chosen at the point where your rebase resolution meets main's guard** — `url = f"https://.../{tag}" if tag else None` at `doctor.py:250`, anchor counted exactly once, module genuinely re-imported:

```
3 failed, 53 passed
E  'download.pytorch.org/whl/' is contained here:
E    l https://download.pytorch.org/whl/None
```

The mutant emits `whl/None` **from the real wheel table** at driver `(11,7)` — which is the direct proof that the third finding is closed, because the test is now driving real code rather than a stub of itself. Reverted → 56 passed.

630 passed, 8 skipped across 17 doctor/gpu files on **both** the branch and the merged tree; `test_cli_startup_is_light` 34 passed and `torch in sys.modules: False` after `doctor --help`, which matters because `soup doctor` is on the light CLI path; ruff clean.

## Merge order, and it is not your fault

Your own #947 adds `_detect_gpu_arch_mismatch_advisory()`, which calls the tag function **unguarded**:

```python
wheel = _torch_cuda_wheel_tag(driver_cuda)
index_url = f"https://download.pytorch.org/whl/{wheel}"
```

On `main` that function returns `str`. This PR correctly widens it to `str | None`. Composed, I measured it:

| driver | #947 alone | #883 + #947 |
|---|---|---|
| (11, 7) | `whl/cu118` | **`whl/None`** |
| (11, 0) | `whl/cu118` | **`whl/None`** |
| (11, 8) | `whl/cu118` | `whl/cu118` |
| (13, 4) | — | `whl/cu132` |

The two merge clean into each other with no conflict marker, and `tests/test_doctor.py` on the composed tree is **46 passed**. Nothing warns anybody.

**This one goes first.** It owns the `str | None` contract and both of its own call sites are already guarded on the tag. #947 then rebases and guards its new site — I have put the detail on that thread, so there is nothing for you to change here.

One non-blocking leftover: `docs/backends-and-ops.md:1162-63` still says doctor "prints the `pip install` command for the wheel your driver can run", which after this PR is conditional. Worth a line whenever you are next in that file; I am not holding the merge for it.

CI is pending on `3bf23829`. I will merge on green without re-reviewing.

MakazhanAlpamys:
Merged as `32650eb4`. Thank you — and the overnight queue drain means this went in on the verdict recorded at `3bf23829`, unchanged and not re-derived.

**What I re-checked rather than re-reviewed:** all 13 required contexts green on that exact SHA, `mergeStateStatus` CLEAN, and the one interaction that could have bitten — **#877 merged into `commands/doctor.py` after my review of you**, so I built the merged tree rather than trusting two separately-green branches. 105 passed there across `test_doctor.py`, #757's package-wide typer-defaults scanner, the CLI startup-is-light runtime guard and the fragment validator. `torch in sys.modules` stays False after `doctor --help`, which matters because `soup doctor` is on the light path.

**The mutation is the best one I have seen on this PR series.** Dropping `if tag else None` from the URL construction fails three tests and emits `whl/None` **from the real wheel table** at driver `(11,7)` — which is the direct proof that the test now drives production code instead of a stub of itself. That was the exact defect I blocked an earlier revision for; closing it by deleting a monkeypatch rather than adding an assertion is the right shape.

And it is correct on hardware, not only in fixtures. This box is the case the PR exists for:

```
nvidia-smi: | NVIDIA-SMI 616.92   KMD Version: 616.92   CUDA UMD Version: 13.4 |
main    :  driver = None      tag = cu121
yours   :  driver = (13, 4)   tag = cu132
```

**The half I want on the record is the conservative one.** Making the parser match modern headers is the visible fix. Making an *unrecognised* driver return `None` instead of guessing `cu121`, and guarding both call sites on the tag, is the one that matters more: a wrong wheel URL is worse than no wheel URL, because the user follows it. Stating both behaviour changes in the fragment rather than letting someone discover them is the same instinct.

Credited in `CONTRIBUTORS.md` (`1753ba26`). #827 is closed with credit.

**One thing carried into your #947**, and it is the reason I took this one first: #947's new `_detect_gpu_arch_mismatch_advisory()` calls `_torch_cuda_wheel_tag` unguarded, so composed with the `str | None` widening you just landed, a driver below CUDA 11.8 emits `whl/None` — and the whole doctor suite stays green, with git reporting no conflict. Detail is on that thread; nothing to do here.

Issue #880, opened 2026-09-12
soup edit diff with both models but no --probes exits 0 with an empty diff (the mirror of #818)


MakazhanAlpamys:
Found while reviewing #863 (@kok-o's fix for #818). Not in #818's scope, so it is filed rather than folded in.

#863 closes one direction: `--probes` given without both models is refused. **The mirror is still open** — both models given and `--probes` *forgotten* succeeds and exits 0.

## What

```
soup edit diff before after --before-model ./base --after-model ./edited -o d.json
```

exits **0**, prints `Total probes: 0; 0 changed.`, and writes `{"changes": [], "total_probes": 0}`. Measured on `main` after #863.

A script that runs `soup edit set && soup edit diff --before-model X --after-model Y -o d.json` and drops `--probes` gets a confident success and an empty diff, which reads as "the edit changed nothing" rather than "nothing was measured".

## Why this is smaller than #818 was, and still worth closing

It is materially better than the bug #818 described. A consumer *can* distinguish the two cases, because `total_probes` is `0` rather than a non-zero count with `changed: false` — which was the whole complaint in #818, where an ungenerated probe was written as `"changed": false` and was indistinguishable from a measured no-change.

So this is a usability gap, not a correctness trap. It is worth closing because #863 establishes the symmetric rule everywhere else: every other combination of those three flags now refuses with a message naming the missing one, and this is the only surviving hole in that shape.

## Suggested fix

Refuse when both models are given and `--probes` is absent, with a message naming `--probes`, exiting the same way as #863's guards — **exit 2**, matching all nine `raise typer.Exit` sites in `src/soup_cli/commands/edit.py`. Note that file's exit code convention is inconsistent with the repo-wide 1-for-usage rule; that is pre-existing and tracked in #813, so follow the file rather than the rule here and let #813 fix both at once.

Roughly three lines beside #863's existing guard.

## Acceptance criteria

- [ ] Both models given with no `--probes` exits non-zero and the message names `--probes`
- [ ] Every combination #863 already refuses still refuses with its existing message
- [ ] Neither model given and no `--probes` is unchanged (that is a valid non-live invocation)
- [ ] A test that fails against current `main`
- [ ] The test asserts the message, not only the exit code, so it cannot pass because a different guard fired

Srinivasan8888:
Claiming this one.

[...]

Issue #834, opened 2026-09-11
test_v0530 test_apply_live_gated (FP8 and NVFP4) depend on the real card and on torchao being absent: with torchao importable on sm_120 they fail with AttributeError


MakazhanAlpamys:
## The defect

Two tests in `tests/test_v0530.py` call the real hardware and dependency gates with a dummy `object()` as the model:

```python

# test_v0530.py:415-420  TestFP8Attention::test_apply_live_gated

with pytest.raises((RuntimeError, ValueError), match="(?i)torchao|hopper"):

    apply_fp8_attention(object())



# test_v0530.py:499-524  TestNVFP4::test_apply_live_gated (after b671a6be)

if not is_blackwell_gpu():

    with pytest.raises(RuntimeError, match="Blackwell"):

        apply_nvfp4(object())

    return

...

with pytest.raises(RuntimeError) as excinfo:

    apply_nvfp4(object())

```

They pass only while **some** gate refuses before the function touches the model. On a card that passes the hardware gate (sm_120 passes `major >= 9` for FP8 and `major >= 10` for NVFP4), the only thing left to refuse is "torchao missing". Once torchao is importable, both functions go on to `model.named_modules()` and raise `AttributeError`, which neither `pytest.raises` accepts.

## Evidence

`apply_fp8_attention` checks torchao (`utils/advanced_precision.py:246`), then the hardware (`:254`), then walks `model.named_modules()` (`:264`). `apply_nvfp4` checks the hardware (`:324`), then torchao and `NVFP4Config` (`:329-339`), then walks `model.named_modules()` (`:343`). A test that passes `object()` depends on one of the gates refusing first.

`b671a6be` fixed the first symptom of this, where the NVFP4 test hard-coded "not Blackwell". It did so by branching on the real `is_blackwell_gpu()`, so the test still depends on the environment, just one gate later.

## Reproduction

RTX 5070 Laptop (sm_120), driver 616.92, torch 2.14.0+cu130. Nothing installed; torchao was made importable only for the run.

Control, as the box is (no torchao):

```

2 passed in 3.27s

```

With the unpacked torchao 0.18.0 wheel on `PYTHONPATH`:

```

E       AttributeError: 'object' object has no attribute 'named_modules'

FAILED tests/test_v0530.py::TestFP8Attention::test_apply_live_gated - Attribu...

1 failed, 1 passed in 5.45s

```

[...]

PR #884, opened 2026-09-12
fix(doctor): name missing [train] packages on a partial stack

### LINKED ISSUE #875: soup doctor says "Training stack not installed" with six of seven [train] packages present and OK

HarshRajSinghania:
## Summary

`soup doctor` no longer reports the training stack as absent when some `[train]` members are installed. A partial install names the missing packages and still points at `pip install "soup-cli[train]"`.

Fixes #875.

## Motivation

#875: with six of seven `[train]` packages present, doctor printed `Training stack not installed` because `group_missing` flipped on the first absent member. That message is wrong for the usual broken-install case (one wheel missing) and `pip install "soup-cli[train]"` is a no-op for the packages already satisfied.

## Implementation

- Track the missing extra-group members instead of a boolean.
- None present: keep the existing `Training stack not installed` text (including the NVIDIA two-step torch index path).
- Some present: `Training stack incomplete, missing: <packages>` plus the group install line.
- If NVIDIA is detected and `torch` itself is among the missing members of a partial stack, keep the two-step torch index advice.
- Exit code is unchanged: a partial `[train]` group remains exit 0 (#828).

## Testing

Ran:

```
PYTHONPATH=src python3 -m pytest tests/test_doctor.py -q --tb=short -o addopts=
```

Result: 32 passed.

New tests:
- `test_doctor_partial_train_extra_names_missing_members` — bitsandbytes absent, six members present; asserts the incomplete message and no "not installed" / "All checks passed!".
- `test_doctor_missing_train_extra_message_unchanged_when_none_installed` — none present keeps the old message.

### DIFF
--- a/changelog.d/0.75.0/884.fixed.md
+++ b/changelog.d/0.75.0/884.fixed.md
@@ -1,0 +1,3 @@
+- **`soup doctor` names missing `[train]` packages on a partial stack (#884, #875).**
+  A six-of-seven install no longer prints "Training stack not installed".
+  Contributed by Harsh Raj Singhania.
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -238,10 +238,15 @@ def doctor(
             if extra_name == "train":
                 driver = _nvidia_smi_cuda_version()
                 torch_missing = "torch" in missing_pkgs
+                # Single call site, gated on torch itself being missing.
+                tag = (
+                    _torch_cuda_wheel_tag(driver)
+                    if driver is not None and torch_missing
+                    else None
+                )
+                url = f"https://download.pytorch.org/whl/{tag}" if tag else None
                 if all_missing:
-                    if driver is not None:
-                        tag = _torch_cuda_wheel_tag(driver)
-                        url = f"https://download.pytorch.org/whl/{tag}"
+                    if url is not None:
                         # ``--index-url`` replaces PyPI, so torch must come from the
                         # CUDA wheel index in its own step; the ``[train]`` extra is
                         # then resolved against PyPI with torch already satisfied.
@@ -255,9 +260,7 @@ def doctor(
                         issues.append(
                             'Training stack not installed: pip install "soup-cli[train]"'
                         )
-                elif driver is not None and torch_missing:
-                    tag = _torch_cuda_wheel_tag(driver)
-                    url = f"https://download.pytorch.org/whl/{tag}"
+                elif url is not None:
                     issues.append(
                         f"Training stack incomplete, missing: {missing_list}\n"
                         f"  pip install torch --index-url {url}\n"
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -205,7 +205,7 @@ def doctor(
     # extra, so the suggestion keeps the declared ceilings and the platform
     # torch index instead of bare per-package floors.
     for extra_name, members in EXTRA_GROUPS:
-        group_missing = False
+        missing_pkgs: list[str] = []
         for import_name, pkg_name, min_ver in members:
             version_str = _installed_version_str(import_name, pkg_name)
             if version_str is None:
@@ -216,7 +216,7 @@ def doctor(
                     f">={min_ver}",
                     "[dim]not installed[/]",
                 )
-                group_missing = True
+                missing_pkgs.append(pkg_name)
                 continue
             max_excl = _MAX_EXCLUSIVE.get(pkg_name)
             if max_excl and _version_ge(version_str, max_excl):
@@ -232,28 +232,54 @@ def doctor(
                 issues.append(f'Upgrade {pkg_name}: pip install "{pkg_name}>={min_ver}"')
                 fix_parts.append(f'"{pkg_name}>={min_ver}"')
             table.add_row(pkg_name, escape(f"[{extra_name}]"), version_str, f">={min_ver}", status)
-        if group_missing:
+        if missing_pkgs:
+            all_missing = len(missing_pkgs) == len(members)
+            missing_list = ", ".join(missing_pkgs)
             if extra_name == "train":
                 driver = _nvidia_smi_cuda_version()
-                if driver is not None:
+                torch_missing = "torch" in missing_pkgs
+                if all_missing:
+                    if driver is not None:
+                        tag = _torch_cuda_wheel_tag(driver)
+                        url = f"https://download.pytorch.org/whl/{tag}"
+                        # ``--index-url`` replaces PyPI, so torch must come from the
+                        # CUDA wheel index in its own step; the ``[train]`` extra is
+                        # then resolved against PyPI with torch already satisfied.
+                        issues.append(
+                            "Training stack not installed:\n"
+                            f"  pip install torch --index-url {url}\n"
+                            '  pip install "soup-cli[train]"'
+                        )
+                        fix_pre.append(f"pip install torch --index-url {url}")
+                    else:
+                        issues.append(
+                            'Training stack not installed: pip install "soup-cli[train]"'
+                        )
+                elif driver is not None and torch_missing:
                     tag = _torch_cuda_wheel_tag(driver)
                     url = f"https://download.pytorch.org/whl/{tag}"
-                    # ``--index-url`` replaces PyPI, so torch must come from the
-                    # CUDA wheel index in its own step; the ``[train]`` extra is
-                    # then resolved against PyPI with torch already satisfied.
                     issues.append(
-                        "Training stack not installed:\n"
+                        f"Training stack incomplete, missing: {missing_list}\n"
                         f"  pip install torch --index-url {url}\n"
                         '  pip install "soup-cli[train]"'
                     )
                     fix_pre.append(f"pip install torch --index-url {url}")
                 else:
-                    issues.append('Training stack not installed: pip install "soup-cli[train]"')
+                    issues.append(
+                        f"Training stack incomplete, missing: {missing_list}\n"
+                        '  pip install "soup-cli[train]"'
+                    )
                 fix_parts.append('"soup-cli[train]"')
-            else:
+            elif all_missing:
                 issues.append(
                     f"{extra_name} stack not installed: "
                     f'pip install "soup-cli[{extra_name}]"'
+                )
+                fix_parts.append(f'"soup-cli[{extra_name}]"')
+            else:
+                issues.append(
+                    f"{extra_name} stack incomplete, missing: {missing_list}\n"
+                    f'  pip install "soup-cli[{extra_name}]"'
                 )
                 fix_parts.append(f'"soup-cli[{extra_name}]"')
 
--- a/tests/test_doctor.py
+++ b/tests/test_doctor.py
@@ -315,6 +315,34 @@ def test_doctor_nvidia_train_suggestion_is_two_step(monkeypatch):
     assert '["soup-cli[train]" --index-url' not in out and '[train]" --index-url' not in out
 
 
+def test_doctor_nvidia_partial_stack_with_torch_missing(monkeypatch):
+    """NVIDIA + partial [train] + torch missing keeps the two-step index URL (#884)."""
+    installed = {
+        "transformers": "5.16.1",
+        "peft": "0.20.0",
+        "trl": "0.29.0",
+        "datasets": "2.14.0",
+        "bitsandbytes": "0.41.0",
+        "accelerate": "0.27.0",
+    }
+
+    def _fake_version(import_name, pkg_name):
+        return installed.get(pkg_name)
+
+    monkeypatch.setattr("soup_cli.commands.doctor._installed_version_str", _fake_version)
+    monkeypatch.setattr(
+        "soup_cli.commands.doctor._nvidia_smi_cuda_version", lambda: (13, 0)
+    )
+    result = runner.invoke(app, ["doctor"])
+    assert result.exit_code == 0
+    out = _strip_ansi(result.output)
+    assert "Training stack incomplete, missing: torch" in out
+    assert "pip install torch --index-url https://download.pytorch.org/whl/" in out
+    assert 'pip install "soup-cli[train]"' in out
+    assert "All checks passed!" not in out
+    assert "Training stack not installed" not in out
+
+
 def test_doctor_missing_core_dependency_exits_nonzero(monkeypatch):
     """A missing core dependency makes `soup doctor` exit non-zero (#828)."""
     monkeypatch.setitem(sys.modules, "plotext", None)
--- a/tests/test_doctor.py
+++ b/tests/test_doctor.py
@@ -174,6 +174,45 @@ def test_doctor_missing_train_extra_suggests_extra(monkeypatch):
     assert "accelerate>=0.27.0" not in out
 
 
+def test_doctor_partial_train_extra_names_missing_members(monkeypatch):
+    """A partial [train] install names the absent members, not the whole stack (#875)."""
+    installed = {
+        "torch": "2.6.0",
+        "transformers": "5.16.1",
+        "peft": "0.20.0",
+        "trl": "0.29.0",
+        "datasets": "2.14.0",
+        "accelerate": "0.27.0",
+    }
+
+    def _fake_version(import_name, pkg_name):
+        return installed.get(pkg_name)
+
+    monkeypatch.setattr("soup_cli.commands.doctor._installed_version_str", _fake_version)
+    monkeypatch.setattr("soup_cli.commands.doctor._nvidia_smi_cuda_version", lambda: None)
+    result = runner.invoke(app, ["doctor"])
+    assert result.exit_code == 0
+    out = _strip_ansi(result.output)
+    assert "Training stack not installed" not in out
+    assert "Training stack incomplete, missing: bitsandbytes" in out
+    assert 'pip install "soup-cli[train]"' in out
+    assert "All checks passed!" not in out
+
+
+def test_doctor_missing_train_extra_message_unchanged_when_none_installed(monkeypatch):
+    """With no [train] members, the existing not-installed message is kept (#875)."""
+    monkeypatch.setattr(
+        "soup_cli.commands.doctor._installed_version_str", lambda import_name, pkg_name: None
+    )
+    monkeypatch.setattr("soup_cli.commands.doctor._nvidia_smi_cuda_version", lambda: None)
+    result = runner.invoke(app, ["doctor"])
+    assert result.exit_code == 0
+    out = _strip_ansi(result.output)
+    assert 'Training stack not installed: pip install "soup-cli[train]"' in out
+    assert "Training stack incomplete" not in out
+    assert "All checks passed!" not in out
+
+
 def test_doctor_suggestion_is_colour_safe(monkeypatch):
     """The [train] suggestion survives Rich highlighting (#828 review)."""
     from rich.console import Console


MakazhanAlpamys:
@HarshRajSinghania — welcome, and thank you for picking this up. I filed #875 a few hours ago and did not expect a PR the same day.

**Your CI runs are approved.** That is GitHub's one-time gate on a first-time contributor's workflow runs, not a comment on your branch — I checked your diff touches no `.github/` files before approving, which is the only reason the gate exists.

**One thing I checked before you had to ask about it.** This PR and #877 (by @chiliec, for #874) touch the same two files — `src/soup_cli/commands/doctor.py` and `tests/test_doctor.py` — and the changed regions are adjacent: theirs adds a line around `:230`, yours rewrites `:232-288`. That is the shape where git auto-merges cleanly and quietly changes meaning, so I did not settle for "no conflict markers". I merged the two branches for real and checked that their one-line fix survives **in the right place** afterwards:

```
core_broken = True   -> 3 occurrences, the third at :230,
                        still inside the extra-group INCOMPATIBLE branch
```

So the two compose properly and neither of you needs to wait for the other. If #877 merges first you may see a trivial rebase; nothing semantic.

**Scheduling, so the wait is visible rather than mysterious.** `main` is frozen until v0.75.0 is **tagged** — the release merged an hour ago but the tag is not out, and anything merged in that window leaves a changelog fragment behind and fails the tag before PyPI. So your PR will be reviewed now and merged after the tag.

**A heads-up on what I will check, so there are no surprises.** #875 is about the *message*, not the exit code: with six of seven `[train]` packages present and in range, `soup doctor` said "Training stack not installed", which is both wrong and unactionable. So I will be looking for the report to name *which* packages are missing, and for the four constraints from #828 and #874 to survive — a missing group stays advisory (exit 0), an outdated member stays advisory, a member past its ceiling blocks, and a missing core dependency still exits 1. I will measure those by running the CLI, not by reading the diff.

I will also mutate the fix: if a test asserting the missing packages are named still passes with the naming removed, that is a blocking finding regardless of how the code reads.

One note for later: your changelog fragment, if you add one, goes in `changelog.d/0.75.0/` and not `0.74.0/` — the release emptied the old directory an hour ago, and sixteen open PRs are currently tripping over exactly that.

Full review shortly.

MakazhanAlpamys:
@HarshRajSinghania — reviewed at `b1ac476c`. **The code is correct**, and I verified all nine behavioural scenarios by running them, not by reading the diff. Two changes and it merges. For a first contribution this is well above the usual bar, and I want to be specific about why rather than just saying so.

## What you got right that most people miss

**You avoided the trap #875 explicitly warned about.** The issue names a wrong alternative — gate on `present == 0` and otherwise emit nothing, which produces `All checks passed!` sitting next to a `To fix:` line. Both of your new tests assert `"All checks passed!" not in out`. You read the issue's reasoning, not just its title.

**`torch_missing = "torch" in missing_pkgs` is a correctness detail nobody asked for.** On a partial stack where torch is already present, `main` emits a pointless `pip install torch --index-url …` pre-step. You suppress it — I measured that path and `fix_pre` is correctly empty. That is exactly the case that usually ships broken.

**You did not add your own `escape()` on the new strings, and that was right.** `escape(issue)` is applied centrally at `doctor.py:302`, so a local escape would have double-escaped. I checked it empirically under a forced-colour console: `soup-cli[train]` renders literally and `[train]` is not swallowed as markup.

**And your second test is a regression guard rather than a bug reproducer** — it passes on `main` by design and pins behaviour that must not change. The first one fails on `main` and passes here, which I confirmed by running your tests against `main`'s source: `1 failed, 31 passed`.

Measured before/after on the reported case — six of seven present, `bitsandbytes` absent:

```
main:  Training stack not installed: pip install "soup-cli[train]"     <- the bug
PR:    Training stack incomplete, missing: bitsandbytes
         pip install "soup-cli[train]"
```

All four #828/#874 constraints survive byte-identically: missing group exit 0, outdated member exit 0, core dependency missing exit 1, healthy exit 0. Tests are `+39/-0` — nothing deleted or weakened. Four mandatory mutations, all caught.

## Two changes

**1. Add `changelog.d/0.75.0/884.fixed.md`.** This changes a message users read, so the fragment convention applies. Note the directory is **`0.75.0`** — v0.75.0 merged this morning and emptied `0.74.0/`, which sixteen other open PRs are currently tripping over. Name it after the PR, mention `#884`, and credit yourself.

**2. One test for the NVIDIA + partial + torch-missing branch** at `doctor.py:258-266`. I replaced its entire message with `"MUTANT6"` and **63 tests passed**. The branch works — I measured it — but your PR body advertises that behaviour and nothing pins it. That same test also covers the `fix_pre.append(...)` line one below, which is likewise unprotected.

A note on your existing coverage: all four mandatory mutations are caught, but **by the same single test**. It earns its keep — it asserts the exit code, the positive message and two negatives — yet the whole guarantee rests on one assertion block.

## Something you need to know about, because another PR is about to change the ground under you

`doctor.py:259` does `tag = _torch_cuda_wheel_tag(driver)` and interpolates `tag` straight into the URL. That is safe **today**: on current `main` the function always returns a string (I checked — an unknown driver falls back to `cu121`).

**#883 by @Samearth17 widens that return type to `str | None`.** With that merged, any driver reporting CUDA 11.0–11.7 yields `None` and your line prints `https://download.pytorch.org/whl/None`, which 403s. I measured that on their branch:

```
driver (11, 0) -> tag None    driver (11, 8) -> tag 'cu118'
```

They have the same bug at their own call site and I have asked them to guard it. Your PR *duplicates* the call site, so whichever of you merges second inherits it — a `tag is None` fallback to the plain `pip install "soup-cli[train]"` line. Nothing to do now; I am telling you so it is not a surprise, and I will make sure it is handled in the right order.

Good news on the other overlap: I merged your branch with **#877** for real rather than trusting the absence of conflict markers, and it composes properly — their `core_broken = True` survives in the right branch, your partial-stack messages are byte-identical after the merge, and the combined run is 133 passed. The intersection case (partial stack *and* a package past its ceiling) behaves sensibly: exit 1, both issues listed, `Fix all` merging both specs.

## Two optional notes

`doctor.py:280-284` — the generic non-`train` partial branch — is unreachable, because `EXTRA_GROUPS` contains only `train`. I confirmed that, and the mutation there survives. Symmetry is a defensible reason to keep it, but make that a conscious call rather than an accident; a one-line comment saying it is future-proofing would do.

`test_doctor_missing_train_extra_message_unchanged_when_none_installed` asserts a 58-character string that Rich wraps below about 62 columns, so it fails at `COLUMNS=60`. Two pre-existing tests in that file have the same shape and CI's default width is unaffected, so this is not a risk today — but a colour-and-width failure was just reported against released code in #886, so a whitespace-collapsed comparison would be the durable form. Your *other* new test is already width-safe, because its message carries an embedded newline.

**Timing:** `main` has v0.75.0 merged but **not tagged**, so nothing merges until the tag is out. Your required checks were still pending when I finished; push those two and I will merge on green after the tag.

MakazhanAlpamys:
@HarshRajSinghania — first, welcome, and thank you: this is a correct fix on your first contribution, and it is correct in the way that is hard to get right.

**I owe you the awkward part up front.** #888 (@blackcoderx) solves the same issue, and I am currently leaning towards merging that one. You were **first** — you opened this at 08:36:59Z; they claimed #875 at 08:50:04Z, thirteen minutes later, and opened theirs at 09:29:48Z. So this is not a case of you being late, and I am not going to pretend it is. I would rather tell you where it stands while you can still do something about it than close your first PR after the fact.

## What is genuinely good here, measured

- `test_doctor.py` + `test_v0401_part_c.py`: **52 passed**; broader doctor surface **145 passed, 1 skipped**. ruff clean.
- **Your new test fails on `main`** (1 failed / 31 passed). That is the check I care about more than any other, and a surprising number of PRs do not pass it.
- **Core mutation killed:** forcing `all_missing = True` — i.e. reverting to main's all-or-nothing behaviour — fails your test.
- No tests deleted, no skips added.
- I drove five scenarios through the real CLI. **Your output is byte-identical to #888's in all five**, and both preserve #828's all-missing wording exactly, including the NVIDIA two-step form. All five of #875's acceptance criteria hold. On correctness this is a tie, not a near miss.

## The two things that separate them

**1. The one that actually decided it, and it is not a style point.** #883 is in flight and changes `_torch_cuda_wheel_tag` to return `str | None`. Your branch adds a **third** call site (`doctor.py:243` and `:259`, plus `:575`); #888 goes the other way and narrows it to **one**, gated on `"torch" in missing_members`. I test-merged all three: `main + #883 + #884` **merges with no conflict at all** and then emits `https://download.pytorch.org/whl/None` — an HTTP 403 — at **two** call sites. Neither of you could have seen this; it only appears when three branches are merged together and the CLI is actually run. But it means the two PRs pull in opposite directions on a surface that is about to become dangerous.

**2. No changelog fragment.** `changelog.d/0.75.0/884.fixed.md`. Mechanical.

## So here is the choice, and it is yours

**Option A — you close the gap and I merge yours.** Add the fragment, and consolidate to a single `_torch_cuda_wheel_tag` call site gated the way #888 does it. That is maybe twenty minutes of work, and it would make this strictly better than the alternative on the axis that decided it. I will hold #888 while you do it if you want it.

**Option B — you would rather not**, and I merge #888 and close this. In that case you still get a CONTRIBUTORS.md entry for finding and fixing it independently — this project has done that before for work adopted in-tree, and being first to a correct diagnosis is worth recording whether or not your commit is the one that lands.

Reply either way and I will act on it. No wrong answer.

## If you want a second thing to pick up regardless

**#896 — consolidate the six private `_for_terminal` copies onto `soup_cli.utils.terminal.for_terminal`.** It is unclaimed, `good first issue`, and it is close in shape to what you just did: six command modules each carry a byte-similar private copy of "strip C0 control bytes, then escape Rich markup", and a seventh private copy is exactly how the config loader got missed for a whole release. The interesting half is the acceptance criterion — a repo-wide ratchet test asserting no module under `src/soup_cli` defines its own `_CONTROL_STRIP_TABLE`, mutation-checked so it actually fails when a copy comes back. One heads-up: #904 is in flight and adds `for_terminal` imports to `draft.py` and `shrink.py`, two of the six, so rebase after it lands.

Comment on #896 if you want it and I will mark it claimed for you.

One last thing, for next time and not a criticism: comment on the issue before you start. You were first here and it still cost you, purely because nobody knew you were working on it.

MakazhanAlpamys:
Correction to the offer I made an hour ago, because one half of it evaporated and that is on me, not on you.

**#896 is taken.** @VanshSharmaPES claimed it at 12:58:52Z — **97 seconds** after I posted the comment above suggesting it. They had already written the root-cause analysis before I offered it, so this is coincidence rather than anyone jumping in; but I offered you something that was gone before you could have read it, and I would rather say so than let you find out by clicking.

Unclaimed and in the same shape as the work you already did — a CLI command whose output or exit code is wrong in a way that is easy to reproduce and easy to test:

- **#880** — `soup edit diff` with both models but no `--probes` exits **0** with an empty diff, i.e. it reports success for a comparison it never made. *Caveat:* #863 is open on the mirror case in the same two files, so you would want to rebase after it lands. Say the word and I will tell you when.
- **#899** — the "Training Complete!" panel prints `Loss: 0.0000 -> 0.0000` whenever the run has fewer steps than `logging_steps`. The reporter already did the diagnosis; the interesting part is that the naive fix produces a *second* wrong number (`train_loss` is the run mean, so both ends become equal and it reads as "loss did not move"). Sixteen trainers share the copy-pasted block, so it wants a small shared helper rather than sixteen edits. `bug` + `good first issue` + `help wanted`.
- **#834** — two tests assume the box is *not* Blackwell and fail on an RTX 50-series card. I have that card here, so I can verify any fix you propose immediately, which makes the feedback loop short.

**#899 is the one I would pick** if you want the same kind of impact as your doctor fix: a wrong number shown to every user of a short run, with a real design trap in the obvious fix.

The offer on this PR stands unchanged — close the two gaps and I merge yours, or decline and you still get the CONTRIBUTORS.md entry. No rush, and no answer needed today.

HarshRajSinghania:
Thanks @MakazhanAlpamys — taking option A and closing the two gaps:

1. Added `changelog.d/0.75.0/884.fixed.md` (credits this PR).
2. Consolidated the extra-group `_torch_cuda_wheel_tag` usage to a single call site, gated on `driver is not None and "torch" in missing_pkgs`, so a future `str | None` return cannot interpolate `.../whl/None` at those sites.
3. Added `test_doctor_nvidia_partial_stack_with_torch_missing` for the NVIDIA + partial stack + torch-missing branch and `fix_pre` line.

Commits on this branch:
- `7b2a6feab79b4bd318f7d3292cd5aa0007c616e6`
- `9a845f444f91ff30ab582cea6d6167878d4d1468`

MakazhanAlpamys:
Reviewed at `9a845f44`. The consolidation you added this morning is exactly what I asked for, and the partial-stack behaviour itself is right. I ran the suite and then mutation-tested it; one mutation survived, and it happens to be on the line you just added, so I want to be precise about what it is rather than hand you a vague "add more tests".

**What I ran** (worktree at your head, base interpreter, `PYTHONPATH` at the worktree):

```
python -m pytest tests/test_doctor.py -q --no-cov      ->  33 passed
FORCE_COLOR=1 TERM=xterm-256color  (same file)         ->  31 passed, 2 failed
```

The two colour failures are `test_doctor_nccl_no_gpu` and `test_doctor_nccl_mocked_success`, both pre-existing on `main` and not yours — they assert on raw `result.output`. Your three new tests all use `_strip_ansi` and stay green under forced colour, which is the thing that actually matters for the Linux/macOS cells. Good instinct.

**Mutation results**

| mutation | result |
| --- | --- |
| `all_missing = len(missing_pkgs) == len(members)` -> `all_missing = True` | 2 failed (`test_doctor_partial_train_extra_names_missing_members`, `test_doctor_nvidia_partial_stack_with_torch_missing`) — correctly pinned |
| drop the `and torch_missing` gate on the wheel tag | **33 passed — survived** |

**The blocking finding.** Removing `and torch_missing` from

```python
tag = (
    _torch_cuda_wheel_tag(driver)
    if driver is not None and torch_missing
    else None
)
```

leaves the whole file green, and it is not a cosmetic difference. Measured on your branch with a fake stack where **torch is installed** and only `bitsandbytes` is missing, with `_nvidia_smi_cuda_version` returning `(13, 0)`:

```
your code:  > Training stack incomplete, missing: bitsandbytes

mutated:    > Training stack incomplete, missing: bitsandbytes
              pip install torch --index-url https://download.pytorch.org/whl/cu130
            pip install torch --index-url https://download.pytorch.org/whl/cu130
```

So the gate is what stops `soup doctor` telling a user with a perfectly good torch to reinstall it from a CUDA wheel index — twice, because the line lands in both `issues` and `fix_pre`. That is the exact failure #875 is about, one layer down, and nothing currently holds it in place.

Your existing `test_doctor_nvidia_partial_stack_with_torch_missing` cannot catch it because torch is the missing package there, and the other partial test never mocks `_nvidia_smi_cuda_version`, so `url` is `None` on both sides of the mutation.

**What closes it** — one test, the mirror of the one you already wrote:

```python
def test_doctor_nvidia_partial_stack_with_torch_present(monkeypatch):
    """NVIDIA + partial [train] + torch installed drops the index URL (#884)."""
    installed = {
        "torch": "2.6.0",
        "transformers": "5.16.1",
        "peft": "0.20.0",
        "trl": "0.29.0",
        "datasets": "2.14.0",
        "accelerate": "0.27.0",
    }  # bitsandbytes missing, torch present
    monkeypatch.setattr(
        "soup_cli.commands.doctor._installed_version_str",
        lambda import_name, pkg_name: installed.get(pkg_name),
    )
    monkeypatch.setattr(
        "soup_cli.commands.doctor._nvidia_smi_cuda_version", lambda: (13, 0)
    )
    result = runner.invoke(app, ["doctor"])
    assert result.exit_code == 0
    out = _strip_ansi(result.output)
    assert "Training stack incomplete, missing: bitsandbytes" in out
    assert "index-url" not in out
```

Please confirm it fails when you drop `and torch_missing` and passes with it — that is the whole point, and I would rather you run the mutation yourself than take my word for it.

Nothing else is outstanding. The changelog fragment is in `changelog.d/0.75.0/` with the right shape, and no existing test was removed or weakened. Push that test and I will merge as soon as the checks are green.

HarshRajSinghania:
@MakazhanAlpamys added `test_doctor_nvidia_partial_stack_with_torch_present` as specified: NVIDIA driver mocked to `(13, 0)`, torch present, only `bitsandbytes` missing, and `index-url` must not appear.

Pushed on this branch: `b683dbf2db70062313ad2b9b1753e00a3a94f0fd`.

I could not run `tests/test_doctor.py` in this environment (no soup-cli / GPU doctor fixtures here). Please treat the mutation check (`drop the `and torch_missing` gate` → this test fails; keep the gate → it passes) as still needing a local confirmation on your worktree.

MakazhanAlpamys:
Verified at `b683dbf2`. The test closes it exactly.

```
pytest tests/test_doctor.py -q --no-cov                          ->  34 passed
drop `and torch_missing` from the wheel-tag gate, rerun          ->  1 failed, 33 passed
FAILED tests/test_doctor.py::test_doctor_nvidia_partial_stack_with_torch_present
```

That is the mutation that survived on the previous head, and it is now the only test that dies for it — a precise pin rather than incidental coverage. `assert "index-url" not in out` is the right assertion too: it catches the message going to either `issues` or `fix_pre`, where asserting on a specific sentence would have missed the duplicate line.

CI on this head needed manual approval and I have approved it. Once the checks are green I will merge.

Thank you for the turnaround on this, and for taking the consolidation request seriously when it would have been easier to argue that the original patch already passed its tests.

MakazhanAlpamys:
Merged as `6b879b87`. Thank you — and the part that decided this was the test, not the fix.

Both patches for #875 changed the message correctly. What separated them is that the NVIDIA wheel-tag gate had a **mutation that survived** on your earlier head: dropping `and torch_missing` changed nothing any test could see. You added `test_doctor_nvidia_partial_stack_with_torch_present`, and after it that mutation fails exactly one test and no others — a precise pin rather than incidental coverage. I re-ran it on `b683dbf2` before merging:

```
pytest tests/test_doctor.py -q --no-cov                  ->  34 passed
drop `and torch_missing` from the wheel-tag gate, rerun  ->  1 failed, 33 passed
FAILED tests/test_doctor.py::test_doctor_nvidia_partial_stack_with_torch_present
```

`assert "index-url" not in out` is also the right assertion. The hint can reach the operator through either `issues` or `fix_pre`, and asserting on a specific sentence would have passed while the duplicate line still printed. Choosing the weaker-looking assertion because it covers both paths is the judgement call I would have had to ask for otherwise.

On the fix itself: the reason `Training stack not installed` was worth treating as a real defect rather than a wording nit is that **the recommended remedy was a no-op**. Six packages were already satisfied, so `pip install "soup-cli[train]"` would have changed nothing and the operator would have been left with the same broken box and no new information. Naming the missing package is what turns the output into something actionable.

Your CONTRIBUTORS.md entry is in, #875 is closed, and CI was green on all fourteen checks on the merged SHA.

Thank you also for taking the consolidation request seriously. Arguing that the original patch already passed its own tests would have been the easier reply, and it is the reply I get most often.

PR #872, opened 2026-09-11
feat(recipes): add kimi-k2.5-dpo and kimi-k2.5-grpo recipes (#851)

Areen-09:
## What does this PR do?

Closes #851 (parent tracking #275).

Adds two ready-made recipes for `moonshotai/Kimi-K2.5` to `src/soup_cli/recipes/catalog.py`:
- `kimi-k2.5-dpo`: Direct preference optimization variant (`task: dpo`, `lr: 5e-6`, `dpo_beta: 0.1`, `moe_aux_loss_coeff: 0.01`, LoRA r=32 / alpha=64).
- `kimi-k2.5-grpo`: Group relative policy optimization reasoning variant (`task: grpo`, `epochs: 3`, `lr: 1e-5`, `grpo_beta: 0.1`, `num_generations: 4`, `reward_fn: accuracy`, LoRA r=16 / alpha=32).

This completes the task trio (`sft`, `dpo`, `grpo`) for the Kimi-K2.5 base model alongside the existing `kimi-k2.5-sft`. Both recipes mirror their Kimi-K2.6 counterparts (`kimi-k2.6-dpo` from #664, `kimi-k2.6-grpo` from #614) over the shared ~1T MoE geometry (`max_length: 8192`, `batch_size: 1`, `gradient_accumulation_steps: 16`, `quantization: 4bit`, `moe_lora: true`, `gradient_checkpointing: true`).

### Key Ratchets & Invariants Passed:
1. **Dual-Surface Model ID Pinning**: In `tests/test_recipes.py`, `TestKimiK25DpoRecipe` and `TestKimiK25GrpoRecipe` pin `RecipeMeta.model` (Surface 1) and YAML `base:` (Surface 2) with independent unit tests. Tested locally: mutating Surface 1 fails only `test_recipe_meta_pins_the_model_id`, and mutating Surface 2 fails only `test_yaml_base_pins_the_model_id`.
2. **Catalog Count Synchronized (165 -> 167)**: Updated across all 9 locations:
   - Documentation & module sites: `src/soup_cli/recipes/catalog.py:51`, `CONTRIBUTING.md:126`, `docs/commands.md:155`, `docs/serving-and-export.md:549`, and `docs/serving-and-export.md:558` (verified by `tests/test_recipe_count_is_synced.py`).
   - Milestone test assertion sites: `tests/test_recipes.py:1120`, `tests/test_v07130.py:740`, `tests/test_v07132.py:953`, and `tests/test_v07124.py:277-281`.
[...]

Issue #847, opened 2026-09-11
Recipe: qwen3.5-35b-a3b-grpo, completing the SFT/DPO/GRPO trio (part of #275)


MakazhanAlpamys:
## What

Add one recipe to `src/soup_cli/recipes/catalog.py`: `qwen3.5-35b-a3b-grpo`.

`Qwen/Qwen3.5-35B-A3B` has `qwen3.5-35b-a3b-sft` (v0.71.24) and `qwen3.5-35b-a3b-dpo` (#276) but no reasoning variant. It is the last base in the #275 families sitting at two of three. **Comment on this issue before you start.**

## Base repo

`Qwen/Qwen3.5-35B-A3B` — checked on 2026-09-11 with `huggingface_hub.model_info(..., token=False)`: resolves, not gated, config `model_type: qwen3_5_moe`. Re-check it the same way; an HTTP status is not evidence (#661).

## What to copy, and from where

- Sibling for the model block: `qwen3.5-35b-a3b-sft` (`modality: text`, `max_length: 4096`, LoRA r=16 / alpha=32, `quantization: 4bit`, `moe_lora: true`, `moe_aux_loss_coeff: 0.01`).
- Template for the task block: `qwen3-30b-a3b-reasoning-grpo`, the closest Qwen MoE GRPO recipe (3B active).

Counts from the catalog: all 4 MoE GRPO recipes use `lr: 1e-5`, `batch_size: 1`, `gradient_accumulation_steps: 16`, `gradient_checkpointing: true`; across all 23 GRPO recipes `num_generations: 4` holds 23/23 and `reward_fn: accuracy` 20/23. The SFT sibling's `batch_size: auto` / `gradient_accumulation_steps: 8` should **not** be carried over unchanged — GRPO's batch shape is set by rollout generation, which SFT does not do. State the choice in the PR.

YAML assembled this way parses through `load_config_from_string` on current `main`. That proves the schema accepts it, nothing more.

## The ratchets that must pass

- [ ] `RecipeMeta(...)` entry in `catalog.py`, tags following the DPO sibling plus `grpo` / `reasoning`.
- [ ] `modality: text` in the YAML and the name added to `EXPECTED_QWEN35_TEXT_RECIPES` in `tests/test_issue427_qwen35_text_modality.py:11`; the family audit at `:43` fails otherwise.
[...]

Issue #846, opened 2026-09-11
Recipe: qwen3.6-35b-a3b-dpo and qwen3.6-35b-a3b-grpo (part of #275)


MakazhanAlpamys:
## What

Add two recipes to `src/soup_cli/recipes/catalog.py`:

- `qwen3.6-35b-a3b-dpo`
- `qwen3.6-35b-a3b-grpo`

`Qwen/Qwen3.6-35B-A3B` (MoE, 3B active) currently ships SFT only (`qwen3.6-35b-a3b-sft`). This is one row of the #275 tracker, split out so it can be picked up on its own. **Comment on this issue before you start.**

## Base repo

`Qwen/Qwen3.6-35B-A3B` — checked on 2026-09-11 with `huggingface_hub.model_info(..., token=False)`: resolves, not gated, config `model_type: qwen3_5_moe` (the same architecture as `Qwen/Qwen3.5-35B-A3B`). Re-check it yourself the same way; an HTTP status is not evidence (#661). `scripts/check_recipe_repo_ids.py` does it correctly.

## What to copy, and from where

Sibling: `qwen3.6-35b-a3b-sft` (`modality: text`, `max_length: 4096`, LoRA r=16 / alpha=32, `quantization: 4bit`, `moe_lora: true`, `moe_aux_loss_coeff: 0.01`).

| recipe | task block from | why that template |
|---|---|---|
| `qwen3.6-35b-a3b-dpo` | `qwen3.5-35b-a3b-dpo` | same `qwen3_5_moe` architecture, one generation earlier; it already carries `batch_size: auto`, `gradient_accumulation_steps: 8`, `lr: 5e-6`, `dpo_beta: 0.1`, `moe_lora: true` |
| `qwen3.6-35b-a3b-grpo` | `qwen3-30b-a3b-reasoning-grpo` | the closest Qwen MoE GRPO recipe (3B active) |

[...]

PR #888, opened 2026-09-12
fix(doctor): name the missing [train] members instead of reporting the whole stack "not installed" (#875)

### LINKED ISSUE #875: soup doctor says "Training stack not installed" with six of seven [train] packages present and OK

blackcoderx:
Closes #875.

## Problem
`soup doctor` treats the `[train]` group as all-or-nothing: `group_missing` is set
as soon as any member fails to import, and the message is unconditional. So with six
of seven present and OK — the state a user is in when one wheel failed to build or
another install pulled a package out — the table shows six `OK` rows and one
`not installed`, yet the summary says the whole stack is absent and advises
`pip install "soup-cli[train]"`, a no-op for the six.

## Change
Track which members are absent rather than a boolean:
- **none present** → unchanged `Training stack not installed: pip install "soup-cli[train]"`
  (#828's wording, byte-for-byte, including the NVIDIA two-step form).
- **some present** → `Training stack incomplete, missing: <names>` with the group
  install line on its own row.

An issue is still appended for the partial case, so `soup doctor` never prints
"All checks passed!" next to a "Fix all:" line (the trap the issue calls out). The
exit code is untouched — a partial `[train]` group stays 0; beyond-ceiling is #874.
The CUDA wheel-index step now fires only when torch itself is missing.

## Acceptance criteria
- [x] Partial install names the missing members and does not claim the stack is absent
- [x] None installed → message unchanged
- [x] "All checks passed!" and a "Fix all:" block can never appear together
- [x] Exit code stays 0 in both cases
- [x] A test that fails against current `main`

## Testing
- `pytest tests/test_doctor.py` — 32 green, incl. the new partial-install test and an all-absent control.
- Full suite run: the only failures are pre-existing/environmental (verified identical against clean `main`) — missing `fastapi`/MLX, this box's below-floor `trl`/`transformers`, Windows symlink limits — none in the changed area.
- Manual: partial `[train]` reads `Training stack incomplete, missing: bitsandbytes` with the install line beneath; full dev install still exits 0.


MakazhanAlpamys:
@blackcoderx — reviewed at `b9032e43`. **I am merging this one for #875**, over the competing #884, and I want to be explicit about why since two people did the same work.

## One thing to add first, and it is small

**A mutation of the `"torch" in missing_members` gate survives.** I removed that condition and all 52 tests still passed. That gate is advertised in your PR body *and* in your shipped changelog fragment — "the CUDA wheel-index step now fires only when torch itself is missing" — so it is a claim the suite does not currently back. Please add the case: a partial `[train]` group where torch **is** present and something else is missing, asserting the two-step CUDA advice does **not** appear.

That is the only thing between this and merge.

## What I verified

- `test_doctor.py` + `test_v0401_part_c.py`: **52 passed**; broader doctor surface **145 passed, 1 skipped**. Your new test **fails on `main`**, which is the check I care about most. No tests deleted, no skips added. ruff clean.
- **Core mutation killed:** forcing `all_absent = True` (i.e. main's all-or-nothing behaviour) fails 1 test.
- Five scenarios driven through the real CLI: all-missing with no driver, all-missing with CUDA 12.8, partial-missing on `bitsandbytes`, partial-missing on `torch`, and the healthy case. **#828's all-missing wording is preserved byte-for-byte**, including the NVIDIA two-step form, and the exit code stays 0 everywhere. All five of #875's acceptance criteria hold.

## Why this over #884

Their output is **byte-identical to yours** in all five scenarios, and both kill the core mutation, so correctness and test strength are a genuine tie. The difference is everything around it:

| | #884 | this PR |
|---|---|---|
| changelog fragment | none | `0.75.0/888.fixed.md`, correct convention |
| `_torch_cuda_wheel_tag` call sites | **3** (adds one) | **1** (narrowed with the torch gate) |
| diff | +36 / -10 | +30 / -11 |
| `ruff format` drift vs main | **+11** | **-10** |

The call-site count is the one that actually decided it. #883 is in flight and changes `_torch_cuda_wheel_tag` to return `str | None`; I confirmed that main + #883 + #884 **merges with no conflict** and then emits `https://download.pytorch.org/whl/None` at two call sites. Your version reduces the surface for that instead of enlarging it. Neither author could have seen this — it only shows up when you merge three branches together and run the CLI.

## One imprecision, not a defect

Your fixture `("os", "torch", "0.0.0")` is described as independent of the CI environment's real packages, but `_installed_version_str` falls through to `importlib.metadata.version("torch")` and reads the actually-installed torch (2.14.0+cu130 on my box). I simulated a torch-less machine and all four assertions still hold, so this is a docstring saying more than it can — worth a word when you add the test above.

Add the gate test and I will merge.

MakazhanAlpamys:
One correction to my comment above, in fairness to the other author, and you should know it before I merge.

I wrote as if this were simply the better of two submissions. Checking the timestamps: **@HarshRajSinghania opened #884 at 08:36:59Z**, you claimed #875 at 08:50:04Z — thirteen minutes later — and opened this at 09:29:48Z. They were first to a PR by roughly an hour. You did nothing wrong; you claimed the issue as the process asks and there is no reason you would have seen an unannounced PR. But "they were late" would be false, and I do not want that to be the story.

So I have offered them the choice: close the two gaps (changelog fragment, and consolidating to a single `_torch_cuda_wheel_tag` call site the way yours does) and I merge theirs, or decline and I merge this. I have told them plainly that the call-site question is what separates the two, and that it is a real difference rather than packaging.

Everything in my review stands — the analysis, the measurements, and the request for the `"torch" in missing_members` gate test. If they decline or go quiet, this merges. If they take it up, you will have lost a merge to someone who was there first, which is the right outcome even though it is an annoying one, and I will say so publicly.

Either way the gate test is worth adding now: it is the one claim in your fragment the suite does not back, and it is the right test regardless of which PR lands.

MakazhanAlpamys:
I owe you a direct answer rather than silence, so here it is.

**This PR is not going to be the one that lands, and the reason is timing, not quality.** #884 fixes the same issue, and the two produce byte-identical output. The order was:

- `2026-09-12T08:36:59Z` — #884 opened
- `2026-09-12T08:50:04Z` — your "I am claiming this" comment on #875, 13 minutes later
- `2026-09-12T09:29:48Z` — #888 opened, 53 minutes after #884

So by the time the issue was claimed there was already an open PR against it. That is nobody's fault — #875 had no claim on it when either of you started, and the claim convention only works if the issue is checked first, which is why I now read `gh issue view` before pointing anyone at an issue. But when two correct patches arrive for the same issue, first-opened wins, and I said as much publicly before I knew which way it would fall. I am not going to reverse that now that it costs you rather than him.

What your version did well, concretely: you reached the same behaviour with a smaller diff (+116/-11 against his +111/-12 including a wider test set), and your changelog fragment was correctly placed in `changelog.d/0.75.0/` from the first push, which several open PRs still have not got right. Neither patch had a defect I could point at. That is the frustrating part of losing this one — there was nothing to fix.

**I am leaving this PR open, not closing it today.** #884 is currently blocked on one more test from its author: I mutation-tested it and found that removing the `and torch_missing` gate leaves all 33 tests green while changing real output, so it is not merge-ready yet. If that stalls, your branch is the fallback and I will come back to it. I will close this one only once #884 is actually merged, and I will say so here when that happens.

If you want something to pick up in the meantime, please comment on an unclaimed issue *before* starting and I will confirm it is yours within the hour — that is the part of the process I can actually make reliable for you.

MakazhanAlpamys:
@blackcoderx — a correction and an acknowledgement I owe you, separate from this morning's decision.

**You added the `"torch" in missing_members` gate test on 2026-09-12 at 13:01 (`793ade37`) and I never acknowledged it.** That was the substantive thing I had asked for on the competing PR, you did it here without being asked twice, and it sat unremarked while I wrote you a long explanation about why the other branch wins. Being told your work is not landing is one thing; having the work itself go unmentioned is another, and that part is fair to be annoyed about.

**Status, plainly:** #884 is now verified and merge-ready — its author pushed the mirror test I asked for, I confirmed the previously surviving mutation now fails, and it is waiting only on the Actions queue, which is badly backed up (39 runs queued, 1 completed as I write). So the fallback condition I described this morning is close to lapsing. I will close this PR when #884 actually merges, not before, and I will say so here rather than letting it close quietly.

**What I want to be concrete about instead of just apologising.** You have now written two correct patches for this repo and landed neither, through no fault of your own. Three unclaimed issues of the same shape are open right now, and I will confirm a claim on any of them within the hour if you comment:

- **#845** — `qwen3.6-27b-dpo` and `qwen3.6-27b-grpo`. The sibling recipe and the template are named in the body; the precedent PR (#872) is worth reading first because the recipe count lives in nine live places plus a regenerated fixture, and six *historical* references must be left alone.
- **#846** — the same for `qwen3.6-35b-a3b`.
- **#847** — `qwen3.5-35b-a3b-grpo`, completing an existing SFT/DPO pair.

None is claimed, none is blocked behind another PR, and none is in the `commands/doctor.py` traffic jam that six open PRs are currently fighting over. If you would rather have something in the `doctor` area since you already know that file, say so and I will find you one that is not contended.

MakazhanAlpamys:
@blackcoderx — **#884 merged as `6b879b87` a few minutes ago**, so I am closing this as I said I would, here rather than letting it go quiet.

I want to be accurate about why, because "the other one won" is not the whole of it. Both patches fixed #875 correctly. What decided it was the wheel-tag gate's mutation: on the competing branch, dropping `and torch_missing` changed nothing any test could see, and the author added `test_doctor_nvidia_partial_stack_with_torch_present` — after which that mutation fails exactly one test and no others. That was the substantive thing I had asked for.

**And you added the `"torch" in missing_members` gate test on 2026-09-12 at `793ade37` without being asked twice.** I said this morning that it went unacknowledged while I wrote you a long explanation about the other branch, and I am repeating it here so it sits in the closing comment rather than only in a thread: it was correct work, it was the right instinct, and the timing is what beat it, not the quality.

**Two correct patches, neither landed, through no fault of yours.** That is a bad ratio for someone who has done the work twice, and it is the thing I would most like to fix rather than apologise for again.

The three unclaimed issues I named this morning are still unclaimed — nobody commented on any of them:

- **#845** — `qwen3.6-27b-dpo` and `qwen3.6-27b-grpo`
- **#846** — the same for `qwen3.6-35b-a3b`
- **#847** — `qwen3.5-35b-a3b-grpo`, completing an existing SFT/DPO pair

Read #872 first if you take one: the recipe count lives in nine live places plus a regenerated fixture, and six *historical* references must be left alone. I reviewed exactly that shape today in #933 and the count discipline is the whole review — the model config itself is the easy half.

None is blocked behind another PR, and none is in the `commands/doctor.py` traffic jam that several open PRs are still fighting over. Comment on any of them and it is yours; I will confirm within the hour. If you would rather have something in `doctor` since you already know that file, say so and I will find you one that is not contended.

PR #892, opened 2026-09-12
Floor the serve-fast vLLM extra at >=0.4.3

abdulwaarith0:
The `serve-fast` extra declares `vllm>=0.4.0`, but since #867 the vLLM backend hands the engine a `{"prompt_token_ids": [...]}` mapping as the first positional argument to `AsyncLLMEngine.generate`.

vLLM only accepts a mapping there from **0.4.3** on:

- 0.4.2 docs: `generate(prompt: str | None, sampling_params, request_id, prompt_token_ids: List[int] | None = None, ...)` — string first, ids as a separate keyword.
- 0.4.3 docs: `generate(inputs: str | TextPrompt | TokensPrompt | TextTokensPrompt, sampling_params, request_id, ...)` — the mapping form is the first positional argument, and the `vllm.inputs` module (`PromptInputs` / `TokensPrompt`) first appears in this release.

So on 0.4.0 through 0.4.2 the mapping lands on the `prompt: str` parameter and is mis-handled. This floors the extra at `>=0.4.3` to match what the backend now requires.

Follow-up to the #867 review, where the floor was flagged and asked for as its own PR.


MakazhanAlpamys:
Thanks — and the reasoning is right, including the part most people skip: you named the exact vLLM release where `generate`'s first positional parameter changed from `prompt: str` to the `PromptInputs` union, rather than picking a round number.

**One ordering problem, and it is only ordering.** I checked `main` at `cf23a693`:

- `src/soup_cli/utils/vllm.py:95` — `def build_chat_prompt(messages, tokenizer=None) -> str`
- `:395` — `prompt = build_chat_prompt(request.messages, tokenizer)`
- `:433` / `:502` — `engine.generate(prompt, sampling_params, request_id, **generate_kwargs)`

So on `main` today that first positional argument is a **string**, and `vllm>=0.4.0` is currently accurate. The mapping form arrives with #867, which is not merged yet. Merging this first would publish a raised dependency floor whose justification is not in the tree — a reader of the changelog fragment would go looking for a mapping that is not there.

**So this is blocked on #867, not on anything you need to change.** Both are yours, so the sequence is easy:

1. #867 merges (it is in review now, fragment already moved to `changelog.d/0.75.0/`, checks green on `8e7bf94e`).
2. This merges immediately after.

I will do them in that order and will not ask you to rebase in between unless the merge produces a conflict in `pyproject.toml`, which it should not.

The alternative would have been to fold this one line into #867 so the requirement and the floor land together. I am **not** asking for that: #867 is green and re-running its matrix to move one line in `pyproject.toml` costs more than it buys, and two commits with a stated order bisect better than one. Keeping them separate was the right instinct.

Holding this open until #867 lands. Nothing to do on your side.

MakazhanAlpamys:
Merged as 6b42cc3, one minute after #867 — so the ordering concern from my earlier comment is resolved in the direction you set up, and there was nothing for you to do.

13/13 required checks green on `0392200b`. The justification is now in the tree: `build_engine_prompt` hands `AsyncLLMEngine.generate` a `{"prompt_token_ids": [...]}` mapping as its first positional argument, and the signature that accepts a mapping there starts at 0.4.3.

Thank you for the reasoning in the fragment. "0.4.2 takes `prompt: str | None` plus a separate `prompt_token_ids` keyword; 0.4.3 takes `inputs: str | TextPrompt | TokensPrompt | TextTokensPrompt`" is the kind of statement a reader can check in thirty seconds, and it is a much better artifact than a round number with "should be fine" after it. Dependency floors in this repo have historically been wrong in both directions — v0.74.0 shipped a `torch>=2.5.0` floor that could not actually work with the `trl>=0.29` beside it — so a floor that names the API change it tracks is worth having.

Three merges in a row today. Credited in CONTRIBUTORS.md alongside #788 and #867.

PR #893, opened 2026-09-12
fix(ingest): strip terminal control bytes from Langfuse --pull errors (#885)

### LINKED ISSUE #885: Langfuse error bodies reach the terminal with ANSI escapes intact (soup ingest --pull)
Found while reviewing #881 (@Konuktor's hardening for #865). **Pre-existing, from #859, not introduced by #881** — filed separately so it is not folded into a PR whose scope is elsewhere.

## What

A Langfuse server's error body reaches the terminal with ANSI escape sequences intact.

`src/soup_cli/utils/ingest_pull.py` `_error_detail` collapses whitespace with `" ".join(text.split())`, which handles `\t\n\r\v\f` and **not** `\x1b`, `\x00` or `\x07`. `src/soup_cli/commands/ingest.py:277` then renders it as:

```python
console.print(f"[red]{escape(str(exc))}[/]")
```

and `rich.markup.escape` neutralises `[...]` markup only — a raw ESC byte passes straight through.

Measured end to end: a 500 response whose body is `err \x1b[2J\x1b[31mFAKE ERROR\x07 \x00 tail` arrives at the terminal with `\x1b[2J` (clear screen) and `\x1b[31m` (colour) intact. Rich drops the `\x07` and keeps the escape sequences.

## Why it matters

`--pull` talks to a user-supplied host, so the response body is untrusted input by construction, and `--allow-private-host` exists precisely so people can point it at a self-hosted instance. Clearing the screen and recolouring output is enough to hide or forge a preceding line — the class of defect this repo already guards against in thirteen command modules with a private `_for_terminal` helper. `commands/ingest.py` is not one of them.

## Blocked on v0.75.0

The right fix is the **shared** helper, not a fourteenth copy: `src/soup_cli/utils/terminal.py:for_terminal` (C0/ESC strip + markup escape), which does not exist on `main` yet — it lands with the v0.75.0 release PR #879. So this waits for the tag.

That also makes it a natural first step toward consolidating the thirteen private copies onto the shared helper, which the v0.75.0 notes already list as a follow-up.

## Suggested fix

Route `_error_detail`'s return — and any other server-derived string that reaches `console.print` in `commands/ingest.py` — through `utils/terminal.py:for_terminal`, dropping the bare `escape(...)`.

## Acceptance criteria

- [ ] A 500 body containing `\x1b[2J`, `\x1b[31m` and `\x00` renders inert; no raw ESC byte reaches stdout
- [ ] The status and the actionable part of the message are unchanged — this must not turn a clean `HTTP 401` explanation into a redacted one
- [ ] The test asserts the absence of the specific dangerous sequences rather than of any `\x1b`, since Rich emits its own escapes under `FORCE_COLOR=1` and the suite is run that way
- [ ] Uses the shared `utils/terminal.py` helper; no new private `_for_terminal` copy
- [ ] A test that fails before the fix

wangzhengzhuo05:
## What

`soup ingest --source langfuse --pull` reads an error body from a user-supplied Langfuse host and prints it. The body went through `_error_detail` (whitespace-collapse only) and then `rich.markup.escape`, which neutralises `[...]` markup but lets C0 control bytes through. A 500 body containing `\x1b[2J` / `\x1b[31m` could clear the screen and recolour output in the operator's terminal.

## Why

The `--pull` response body is untrusted input by construction, and this was one of the few `console.print` sites in `commands/ingest.py` not already using the shared terminal-hygiene helper.

## How

Route the rendered `PullError` message through `soup_cli.utils.terminal.for_terminal` (C0/ESC strip + markup escape), replacing the bare `escape(...)`. There are two `except ingest_pull.PullError` print sites — the streaming pull and `load_langfuse_credentials` — and both now use the shared helper, so the file has a single hygiene level for that exception type. The five `TypeError`/`ValueError` validation sites are left as they are (local, not external, input).

Diff: `src/soup_cli/commands/ingest.py` — the import plus two lines; `changelog.d/0.75.0/893.fixed.md` added.

## Tests

`tests/test_issue885_langfuse_error_ansi.py`, driving a real 500 body `err \x1b[2J\x1b[31mFAKE ERROR\x07\x00 tail` through `_error_detail`:

- **CLI-level** (`CliRunner`): asserts `\x1b[2J`, `\x00`, `\x07` are absent from stdout, and that `HTTP 500` / `FAKE ERROR` still reach the terminal.
- **Colour-console level** (`Console(file=StringIO(), force_terminal=True)`): renders the same production expression where colour is genuinely on, asserts a control that Rich did emit SGR (`"\x1b[" in rendered`) — so the absences cannot be satisfied by an unstyled console — and then that `\x1b[2J`, `\x00`, `\x07` are absent. The payload-presence checks compare SGR-stripped text because Rich's `ReprHighlighter` restyles the number in `HTTP 500`; all absence assertions run against raw bytes.

Results: `pytest tests/test_issue885_langfuse_error_ansi.py` -> 2 passed. `ruff check` clean on both changed files.

Mutation check: reverting the streaming call site (`for_terminal` -> `escape`) makes `test_pull_error_body_renders_inert` fail (the injected `\x1b[2J` reaches stdout); restoring it passes.

## Review follow-up (`67184a7`)

- Removed the `\x1b[31m` assertion — it names a sequence Rich emits for our own `[red]` markup, so it only held because `CliRunner` does not colour — and added the `force_terminal=True` case above.
- Added `changelog.d/0.75.0/893.fixed.md`.
- Decided the scope question by changing the second `PullError` site too, so the same exception type no longer has two hygiene levels.

Fixes #885

---
Implemented with AI assistance (OpenCode + muse-spark), then reviewed and verified locally (targeted tests, lint, mutation check) before submission.

### DIFF
--- a/changelog.d/0.75.0/893.fixed.md
+++ b/changelog.d/0.75.0/893.fixed.md
@@ -1,0 +1,9 @@
+- **`soup ingest --source langfuse --pull` printed the server's error body to the terminal
+  with its raw ANSI escapes intact, letting a hostile Langfuse response inject terminal
+  control sequences (#885 by @wangzhengzhuo05 in #893).**
+  `commands/ingest.py` rendered `PullError` through bare `rich.markup.escape`, which
+  neutralises `[...]` markup but passes C0 control bytes (ESC, NUL, BEL) straight through, so a
+  crafted response body could reach the terminal as live escape sequences. The user-facing
+  path now routes the message through the shared `utils/terminal.py:for_terminal` helper (C0
+  strip + markup escape) at both `PullError` print sites, so the server's text is still shown
+  but its control bytes are not. Added by @wangzhengzhuo05.
--- a/src/soup_cli/commands/ingest.py
+++ b/src/soup_cli/commands/ingest.py
@@ -34,6 +34,7 @@ from soup_cli.utils.ingest_sources import (
     validate_source_name,
 )
 from soup_cli.utils.paths import is_under_cwd
+from soup_cli.utils.terminal import for_terminal
 
 console = Console()
 
@@ -277,7 +278,7 @@ def _pull(
         # Streams to a staging file; the target only appears once the pull finished.
         atomic_write_lines(_lines(), str(output_path), field="--output")
     except ingest_pull.PullError as exc:
-        console.print(f"[red]{escape(str(exc))}[/]")
+        console.print(f"[red]{for_terminal(str(exc))}[/]")
         console.print(f"[red]Nothing was written to {escape(output_path.name)}.[/]")
         raise typer.Exit(1) from None
 
--- a/src/soup_cli/commands/ingest.py
+++ b/src/soup_cli/commands/ingest.py
@@ -243,7 +243,7 @@ def _pull(
             os.environ, allow_private_host=allow_private_host
         )
     except ingest_pull.PullError as exc:
-        console.print(f"[red]{escape(str(exc))}[/]")
+        console.print(f"[red]{for_terminal(str(exc))}[/]")
         raise typer.Exit(1) from None
     except ValueError as exc:
         console.print(f"[red]{escape(str(exc))}[/]")
--- a/tests/test_issue885_langfuse_error_ansi.py
+++ b/tests/test_issue885_langfuse_error_ansi.py
@@ -41,8 +41,51 @@ def test_pull_error_body_renders_inert(monkeypatch, tmp_path) -> None:
 
     assert result.exit_code != 0
     assert "\x1b[2J" not in result.output
-    assert "\x1b[31m" not in result.output
     assert "\x00" not in result.output
     assert "\x07" not in result.output
     assert "HTTP 500" in result.output
     assert "FAKE ERROR" in result.output
+
+
+def test_stripped_body_renders_on_a_colour_console() -> None:
+    """Even with colour genuinely on, no payload sequence survives the strip.
+
+    ``CliRunner`` is not colour-capable, so the CLI-level test above cannot tell a
+    working strip from a no-op. This one renders the same production expression
+    through ``Console(force_terminal=True)`` -- where Rich *does* emit its own SGR --
+    and asserts only sequences Rich never emits for this markup are gone.
+    """
+    import re
+    from io import StringIO
+
+    from rich.console import Console
+
+    from soup_cli.utils.terminal import for_terminal
+
+    creds = LangfuseCredentials(
+        public_key="pk", secret_key="sk", host="https://langfuse.example.com"
+    )
+    message = f"Langfuse answered HTTP 500: {_error_detail(HOSTILE, creds)}"
+
+    out = StringIO()
+    console = Console(file=out, force_terminal=True, width=300)
+    # This is the exact production expression from commands/ingest.py.
+    console.print(f"[red]{for_terminal(message)}[/]")
+    rendered = out.getvalue()
+
+    # Control: the console really did colour, so the assertions below are not
+    # trivially true because nothing was styled.
+    assert "\x1b[" in rendered
+    # Erase-display is payload-only; Rich has no reason to emit it.
+    assert "\x1b[2J" not in rendered
+    # C0 bytes stripped by for_terminal().
+    assert "\x00" not in rendered
+    assert "\x07" not in rendered
+    # The server's message itself is preserved, not dropped. Rich's own
+    # ReprHighlighter restyles numbers (``500`` -> ``\x1b[1;31m500\x1b[0m``),
+    # splitting "HTTP 500" with SGR codes that are Rich's own, not payload --
+    # so compare against the SGR-stripped text here. The absence assertions
+    # above already ran against the raw bytes.
+    plain = re.sub(r"\x1b\[[0-9;]*m", "", rendered)
+    assert "HTTP 500" in plain
+    assert "FAKE ERROR" in plain
--- a/tests/test_issue885_langfuse_error_ansi.py
+++ b/tests/test_issue885_langfuse_error_ansi.py
@@ -1,0 +1,48 @@
+"""#885 — Langfuse error bodies must not reach the terminal with ANSI escapes intact.
+
+A 500 response body is untrusted input. ``_error_detail`` collapses whitespace
+but leaves ESC/NUL/BEL bytes alone, so ``commands/ingest.py`` must route the
+rendered ``PullError`` through the shared ``utils/terminal.py:for_terminal``
+helper (C0 strip + markup escape) instead of bare ``rich.markup.escape``.
+"""
+
+from __future__ import annotations
+
+from typer.testing import CliRunner
+
+from soup_cli.cli import app
+from soup_cli.utils import ingest_pull
+from soup_cli.utils.ingest_pull import LangfuseCredentials, PullError, _error_detail
+
+HOSTILE = b"err \x1b[2J\x1b[31mFAKE ERROR\x07\x00 tail"
+
+
+def test_pull_error_body_renders_inert(monkeypatch, tmp_path) -> None:
+    """The injected erase-display/colour sequences never reach stdout intact."""
+    creds = LangfuseCredentials(
+        public_key="pk", secret_key="sk", host="https://langfuse.example.com"
+    )
+    detail = _error_detail(HOSTILE, creds)
+    message = f"Langfuse answered HTTP 500: {detail}"
+
+    def _raise_pull_error(*args, **kwargs):
+        raise PullError(message)
+
+    monkeypatch.setattr(
+        ingest_pull,
+        "load_langfuse_credentials",
+        lambda environ, allow_private_host=False: creds,
+    )
+    monkeypatch.setattr(ingest_pull, "pull_langfuse_generations", _raise_pull_error)
+    monkeypatch.chdir(tmp_path)
+
+    runner = CliRunner(env={"FORCE_COLOR": "1", "TERM": "xterm-256color", "COLUMNS": "300"})
+    result = runner.invoke(app, ["ingest", "--source", "langfuse", "--pull", "-o", "out.jsonl"])
+
+    assert result.exit_code != 0
+    assert "\x1b[2J" not in result.output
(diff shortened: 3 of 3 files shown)


MakazhanAlpamys:
@wangzhengzhuo05 — reviewed at `2e4dc8f`-era head (whatever `gh pr view 893` shows as current; I tested the tree as fetched). **The production change is exactly right and I would not alter a character of it.** One test defect, one missing file, one scope question.

## What I verified

- `pytest tests/test_issue885_langfuse_error_ansi.py` — **1 passed**.
- **Mutation**: reverted `for_terminal(str(exc))` back to `escape(str(exc))` at `ingest.py:281`. **The test fails.** It is load-bearing, which is the thing I check first and the thing that most often is not true.
- You used the shared `utils/terminal.py:for_terminal` rather than adding a fourteenth private copy, which is what #885 asked for and is the whole point of the file landing in v0.75.0.
- The assertions keep `HTTP 500` and `FAKE ERROR` present. A fix that simply dropped the server's message would also remove the escapes, and your test rejects it. That is the control most people leave out.

## 1. `assert "\x1b[31m" not in result.output` is a sequence **Rich itself emits**, and the test is currently vacuous

#885's acceptance criteria warned about the coarse version of this ("assert the specific dangerous sequences rather than any `\x1b`") and you correctly avoided it — but `\x1b[31m` is the SGR code Rich generates for your own `[red]` markup. Measured, on this tree:

```python
c = Console(file=StringIO(), force_terminal=True, width=100)
c.print("[red]Langfuse answered HTTP 500: err [2J[31mFAKE ERROR tail[/]")
# ESC count: 14 ;  "\x1b[31m" in out -> True
```

So on a genuinely colour-capable console that assertion fails **on correct code**. It passes today only because `CliRunner` is not one: I instrumented your exact invocation and `result.output` contains **0 ESC bytes**, despite `FORCE_COLOR=1` and `TERM=xterm-256color` in the env. The colour setup in the test is therefore decorative — it does not make Rich colour, so the test never exercises the situation it is written to describe, and the `\x1b[31m` line is satisfied trivially rather than meaningfully.

Two ways out; I prefer the second:

- Keep only `\x1b[2J`. Rich has no reason to emit an erase-display, so that one cannot collide.
- **Better: render through a console where colour is genuinely on**, e.g. assert against a `Console(file=StringIO(), force_terminal=True)` rendering of the same message, and there assert `\x1b[2J` and `\x1b[0m`-terminated *payload* colour are absent while permitting Rich's own. That version proves the claim in the docstring; the current one proves that CliRunner does not colour.

This repo has had CI go red four times on Rich's per-character ANSI, so a test in this area that silently stops discriminating is worth more attention than its size suggests.

## 2. Missing changelog fragment

This is user-visible (and security-shaped), so it needs `changelog.d/0.75.0/893.fixed.md`. One bullet, PR-numbered, with the credit — the house form is visible in any fragment already in that directory. Note the directory: fragments move to `0.75.0/` now that the release is tagged.

## 3. Scope question, not a request

`commands/ingest.py` has **six** `console.print(f"[red]{escape(str(exc))}[/]")` sites; you changed one. Five are `TypeError` / `ValueError` from local validation and are fine as they are. But `:245-246` is a **second `except ingest_pull.PullError`**, from `load_langfuse_credentials`, still on the bare helper.

That one is env-derived rather than server-derived, so I do not think #885 requires it. What I dislike is that the file now has two print sites for the *same exception type* with *different* hygiene — the shape where someone later widens `load_langfuse_credentials` to include a server response and silently inherits the unsafe one. Either change it too (one line, no test needed, mention it in the fragment) or reply saying you deliberately scoped it out; both answers are fine, I just want it decided rather than accidental.

## Ordering note, nothing for you to do

#881 (@Konuktor, for #865) also touches `commands/ingest.py` and `utils/ingest_pull.py`. Whichever lands second may need a trivial rebase. I will sequence them and tell you if it affects you.

Push 1 and 2 and I will merge — the production line is already correct.

wangzhengzhuo05:
Thanks for the careful read — all three addressed in `67184a7`.

**1. Test defect.** You're right, and the diagnosis matches mine once I rendered it on a colour console. I deleted the `assert "\x1b[31m" not in result.output` line (it names a sequence Rich emits for our own `[red]` markup, so it held only because `CliRunner` does not colour) and added `test_stripped_body_renders_on_a_colour_console`, which renders the same production expression through `Console(file=StringIO(), force_terminal=True, width=300)` and asserts:

- a control that colour is genuinely on — `"\x1b[" in rendered` — so the absences below can't be satisfied by an unstyled console;
- `"\x1b[2J"`, `"\x00"`, `"\x07"` absent (raw bytes — the sequences Rich never emits for this markup);
- `"HTTP 500"` and `"FAKE ERROR"` still present.

One deviation from my first draft, worth flagging since it is exactly your point in miniature: `"HTTP 500" in rendered` fails on a colour console too, because Rich's own `ReprHighlighter` restyles the number — the bytes are `HTTP \x1b[0m\x1b[1;31m500\x1b[0m\x1b[31m:`. So the two payload-*presence* checks compare SGR-stripped text (`re.sub(r"\x1b\[[0-9;]*m", "", rendered)`); every absence assertion still runs against the raw bytes, and no assertion was weakened or dropped.

**2. Changelog.** Added `changelog.d/0.75.0/893.fixed.md`, one bullet in the house form with `(#885 by @wangzhengzhuo05 in #893)`.

**3. Scope.** Decided rather than left accidental: I changed the second `except ingest_pull.PullError` site (`:245-246`, `load_langfuse_credentials`) too. One line, and the fragment now says both `PullError` print sites route through `for_terminal`. The five `TypeError`/`ValueError` validation sites are untouched, as you said they should be — so the file no longer has two hygiene levels for the same exception type.

**Verified locally:** `2 passed` on `tests/test_issue885_langfuse_error_ansi.py`; `ruff check` clean on both changed files. Mutation: reverting the streaming call site (`for_terminal` → `escape`) makes `test_pull_error_body_renders_inert` fail; restoring makes it green.

MakazhanAlpamys:
Both items addressed, and the colour test is better than what I suggested — thank you.

**You added the non-vacuity control I did not ask for.** `assert "\x1b[" in rendered` before the absence assertions is the line that makes the test mean something: without it, a console that silently stopped colouring would turn every following assertion trivially true, which is exactly the failure mode the old `CliRunner` version had. And splitting the message check onto SGR-stripped text, with the comment explaining that Rich's `ReprHighlighter` restyles the `500` in `HTTP 500`, is the right resolution — that highlighter is a recurring trap here and now it is written down in the test that trips over it.

Verified on `67184a79`: **2 passed**. Fragment is in house form with the credit. Mutating `for_terminal(str(exc))` back to `escape(str(exc))` fails `test_pull_error_body_renders_inert`.

## One gap, and it is the PR's own lesson

That mutation changed **both** sites and only **one** test failed. Two related reasons:

1. **`:246` — the credentials `PullError` site you just added — has no test.** This PR exists because a second print site went unnoticed; leaving the newly-fixed one uncovered means a third could too.
2. **`test_stripped_body_renders_on_a_colour_console` copies the production expression rather than driving the production path.** Your docstring is honest about this ("This is the exact production expression from `commands/ingest.py`"), so it is a deliberate trade, not an oversight — but a copy can drift from the original without either failing. It guards `for_terminal`; it does not guard its *use*.

The cheap fix for both is one more `CliRunner` case that makes `load_langfuse_credentials` raise a `PullError` carrying `HOSTILE`, asserting the same absences at exit 1. That reaches `:246` through the real command, and then the both-site mutation fails twice instead of once.

Not a large ask — the existing test is a three-line monkeypatch away from it. Push that and I will merge; CI is already running on your current head.

wangzhengzhuo05:
Gap closed in the new head. Added `test_credentials_pull_error_body_renders_inert`: a `CliRunner` case that makes `load_langfuse_credentials` raise a `PullError` carrying `HOSTILE`, then asserts `exit_code == 1` and the same absences (`\x1b[2J`, `\x00`, `\x07`) plus the payload still present. It reaches `:246` through the real command rather than a copy.

Both-site mutation now fails **2** tests instead of 1:

```
FAILED test_pull_error_body_renders_inert
FAILED test_credentials_pull_error_body_renders_inert
2 failed, 1 passed
```

Restored production code -> `3 passed`, `ruff check` clean. The colour-console test is left as-is (copying the production expression), per your reading that it is a deliberate trade, not an oversight.

MakazhanAlpamys:
Merged as `2b466960`. Thank you — this is a clean patch and the judgement in it is better than the line count suggests.

What I ran, at your head `60c4d543`, in a dedicated worktree on the base interpreter:

```
python -m pytest tests/test_issue885_langfuse_error_ansi.py -q --no-cov   ->  3 passed
```

**Mutation.** I reverted both `for_terminal(str(exc))` call sites to `escape(str(exc))` and got:

```
FAILED test_pull_error_body_renders_inert
FAILED test_credentials_pull_error_body_renders_inert
2 failed, 1 passed
```

Two distinct failures rather than one, so each site is individually pinned — a single shared assertion would have let a half-revert through, and yours does not.

**The part I want to name is what you did *not* change.** The two neighbouring `except ValueError` branches are still on bare `escape`, and a reviewer skimming the diff would call that an inconsistency. I checked before deciding rather than assuming either way: with `LANGFUSE_HOST` set to a value carrying a raw `\x1b[31m`, the run exits 2 with

```
LANGFUSE_HOST is not an allowed Langfuse host: the URL must not contain control characters
```

— the host validator refuses the control byte upstream, so no C0 byte can reach that branch. Widening the patch there would have added surface with no reachable defect behind it. Keeping the change to the attacker-reachable path is the right call.

Also correct: routing through the shared `utils/terminal.py:for_terminal` rather than adding a tenth private copy. #907 is currently consolidating the nine that already exist, so a private helper here would have been merge debt on arrival.

Two notes, neither on you:

- #881 also touches `commands/ingest.py`, so it will need a rebase now. Trivial — your change is two lines in `_pull` and does not overlap its hunks.
- I found two pre-existing colour failures in `tests/test_doctor.py` while working nearby (`test_doctor_nccl_no_gpu`, `test_doctor_nccl_mocked_success` assert on raw `result.output`). Unrelated to this PR; I am tracking it separately.

Credited in `CONTRIBUTORS.md` under your existing entry. Second merge.

Issue #895, opened 2026-09-12
soup draft distill / soup shrink show a raw traceback when their own rendered YAML fails to load


MakazhanAlpamys:
`commands/draft.py` (~line 399) and `commands/shrink.py` (~line 566) call `load_config_from_string(yaml_text)` on YAML they rendered themselves, with no `try/except`, "to validate before spending a subprocess". A `ValueError` there (a future field-name typo in the template, or a schema change) surfaces as an unhandled traceback instead of the friendly `console.print` + `typer.Exit(1)` pattern `ship.py` and `doctor.py` use. Pre-existing; v0.75.0's unknown-key refusal adds a new trigger (found in the release review).

**Fix path:** wrap both call sites in `try/except ValueError`, print the message through `soup_cli.utils.terminal.for_terminal`, exit 1.

**Acceptance:**
- [ ] with the loader monkeypatched to raise `ValueError("x")`, both commands exit 1 and print the message, no traceback
- [ ] the happy path is unchanged

MakazhanAlpamys:
**Claimed by @1cbyc**, working in #904.

Recorded here rather than only in the pull request, because the assignee field silently refused: GitHub accepts an assignee only if they have push access or have commented on **this issue**, and `gh issue edit --add-assignee` exits 0 either way. That is how 11 claims ended up invisible today, including this one.

So the `claimed` label plus this comment is the registry. If you are looking for unclaimed work, trust the label and the comments, not an empty assignee field.

MakazhanAlpamys:
Closed by #904, merged as `52641e99`, by @1cbyc — their first contribution to the project.

Both `soup draft distill` and `soup shrink` now catch the validation error from the `soup.yaml` they render themselves and exit 1 with a sanitized message, instead of a raw `ValueError` traceback arriving seconds before a training subprocess.

The message routes through the shared `utils/terminal.py:for_terminal` rather than a private control-strip copy, and each command's tests include an ESC-byte plus Rich-markup case, so #896's duplication does not gain a new instance here.

Try it: `soup shrink --model <id> --drop-ratio 0.25 --calib calib.jsonl --heal heal.jsonl` — a config the renderer cannot validate now names the problem and exits 1.

Scope note for anyone reading this later: the regression tests pin the handling branch by raising from the loader, not by producing a genuinely malformed render. That is what this issue asked for; a test that drives an invalid template end to end would be a strictly stronger follow-up.

Issue #896, opened 2026-09-12
Consolidate the six private _for_terminal copies onto soup_cli.utils.terminal.for_terminal


MakazhanAlpamys:
v0.75.0 added `src/soup_cli/utils/terminal.py` (`for_terminal`: strip C0/DEL control bytes, then escape Rich markup) for the config loader, whose unknown-key report printed YAML key names unescaped — Rich markup or a raw ESC byte in a key restyles or spoofs the terminal. Six command modules still carry a byte-similar private copy: `commands/adapters.py`, `commands/data_canary.py`, `commands/data_doctor.py`, `commands/draft.py`, `commands/infer.py`, `commands/shrink.py`; `utils/ship_verdict.py:for_terminal` and `mcp_server/registry.py:_sanitize` are the same idea under other names. A seventh private copy is exactly how the loader got missed for a release.

**Fix path:** import the shared helper, delete the copies, and add a repo-wide test asserting no module under `src/soup_cli` defines its own `_CONTROL_STRIP_TABLE`, so an eighth copy fails CI.

**Acceptance:**
- [ ] one definition of the strip table + escape in the tree
- [ ] every existing per-command test still passes
- [ ] the ratchet test fails when a private copy is reintroduced (mutation-checked)

VanshSharmaPES:
I'd like to pick this up if it's not already assigned.

**Root cause:** 
`utils/terminal.py`'s `for_terminal()` (added in v0.75.0 to sanitize the config loader's unknown-key report) isn't the only place doing control-byte stripping — six command modules (`adapters.py`, `data_canary.py`, `data_doctor.py`, `draft.py`, `infer.py`, `shrink.py`) plus `utils/ship_verdict.py` and `mcp_server/registry.py` each carry their own byte-identical private copy of `_CONTROL_STRIP_TABLE`, dating back to `data_doctor.py`'s original version in v0.71.27. Because the config loader was new code, nobody wired it up to an existing (but not obviously canonical) helper, so it shipped without the fix applied - a direct consequence of having eight copies of the same logic instead of one.

**Proposed fix:**
- Consolidate all eight call sites onto `soup_cli.utils.terminal`, removing every duplicate `_CONTROL_STRIP_TABLE`.
- Add `strip_control()` alongside `for_terminal()` for call sites that only need the strip half (e.g. text passed to `Text.append(..., style=...)`, which Rich never parses as markup) without the markup-escaping step.
[...]

PR #904, opened 2026-09-12
fix: report generated config validation errors

1cbyc:
## What does this PR do?

Handles validation failures from the YAML generated by `draft distill` and `shrink` before either command starts its training subprocess. Both paths now print a sanitized project-native error and exit with status 1 instead of exposing a Python traceback.

Focused regressions cover both command boundaries while the existing happy-path tests remain unchanged.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

- [x] `ruff check src/soup_cli/ scripts/ tests/ benchmarks/` passes

Validation:

- 8 focused distill/heal orchestration tests passed
- repository-wide Ruff lint passed
- wheel build passed

Refs #895

Co-authored-by: insisong <emmanuelisaacnsisong@gmail.com>

### DIFF
--- a/changelog.d/0.75.0/904.fixed.md
+++ b/changelog.d/0.75.0/904.fixed.md
@@ -1,0 +1,1 @@
+Report generated distill and heal configuration validation failures as concise CLI errors instead of Python tracebacks.
--- a/changelog.d/0.75.0/904.fixed.md
+++ b/changelog.d/0.75.0/904.fixed.md
@@ -1,1 +1,1 @@
-Report generated distill and heal configuration validation failures as concise CLI errors instead of Python tracebacks.
+- **`soup draft distill` and `soup shrink` report a validation failure in their own rendered config as a CLI error rather than a traceback (#895 by @1cbyc in #904).** Both commands validate the YAML they generate before starting a training subprocess and sanitize the validation message for terminal output.
--- a/src/soup_cli/commands/draft.py
+++ b/src/soup_cli/commands/draft.py
@@ -49,6 +49,7 @@ from soup_cli.utils.draft import (
     same_tokenizer,
 )
 from soup_cli.utils.paths import atomic_write_text, enforce_under_cwd_and_no_symlink
+from soup_cli.utils.terminal import for_terminal
 
 if TYPE_CHECKING:  # pragma: no cover — typing only, keeps the CLI import light
     from transformers import PreTrainedModel, PreTrainedTokenizerBase
@@ -396,7 +397,11 @@ def _run_distill(
         data_rows=data_rows,
         uld_strategy=uld_strategy,
     )
-    load_config_from_string(yaml_text)  # validate before spending a subprocess
+    try:
+        load_config_from_string(yaml_text)  # validate before spending a subprocess
+    except ValueError as exc:
+        console.print(f"[red]Invalid rendered distill config:[/] {for_terminal(exc)}")
+        raise typer.Exit(code=1) from exc
 
     # #364 — surface the resolved optimiser-step budget before the run. Epoch
     # granularity can only land NEAR ``--steps``; printing it makes any mismatch
--- a/src/soup_cli/commands/shrink.py
+++ b/src/soup_cli/commands/shrink.py
@@ -41,6 +41,7 @@ from soup_cli.utils.shrink import (
     shrink_arch_of,
     shrink_verdict_to_dict,
 )
+from soup_cli.utils.terminal import for_terminal
 
 console = Console()
 
@@ -563,7 +564,11 @@ def _run_heal(
         out_dir=out_dir,
         heal_rows=heal_rows,
     )
-    load_config_from_string(yaml_text)  # validate before spending a subprocess
+    try:
+        load_config_from_string(yaml_text)  # validate before spending a subprocess
+    except ValueError as exc:
+        console.print(f"[red]Invalid rendered heal config:[/] {for_terminal(exc)}")
+        raise typer.Exit(code=1) from exc
     config_path = Path(pruned_dir).parent / "heal_config.yaml"
     atomic_write_text(yaml_text, str(config_path), field="heal config")
 
--- a/tests/test_v07129.py
+++ b/tests/test_v07129.py
@@ -933,6 +933,35 @@ class TestRunHeal:
         assert exc_info.value.exit_code == 1
         assert messages == ["[red]Invalid rendered heal config:[/] invalid rendered config"]
 
+    def test_run_heal_sanitizes_rendered_config_validation_error(self, tmp_path, monkeypatch):
+        import typer
+
+        import soup_cli.commands.shrink as sc
+
+        messages: list[str] = []
+        monkeypatch.setattr(
+            "soup_cli.config.loader.load_config_from_string",
+            lambda _yaml: (_ for _ in ()).throw(
+                ValueError("bad \x1b[2J[bold red]field[/]")
+            ),
+        )
+        monkeypatch.setattr(sc.console, "print", lambda message: messages.append(message))
+
+        with pytest.raises(typer.Exit) as exc_info:
+            sc._run_heal(
+                pruned_dir="./model",
+                teacher="t",
+                heal_data="./h.jsonl",
+                steps=5,
+                out_dir="./adapter",
+                heal_rows=10,
+            )
+
+        assert exc_info.value.exit_code == 1
+        assert len(messages) == 1
+        assert "\x1b" not in messages[0]
+        assert "bad [2J\\[bold red]field\\[/]" in messages[0]
+
     def test_nonzero_tail_control_bytes_stripped(self, tmp_path, monkeypatch):
         import soup_cli.commands.shrink as sc
 
--- a/tests/test_v07129.py
+++ b/tests/test_v07129.py
@@ -908,6 +908,31 @@ class TestRunHeal:
             sc._run_heal(pruned_dir="./model", teacher="t", heal_data="./h.jsonl",
                          steps=5, out_dir="./adapter", heal_rows=10)
 
+    def test_run_heal_reports_rendered_config_validation_error(self, tmp_path, monkeypatch):
+        import typer
+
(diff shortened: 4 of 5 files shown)


MakazhanAlpamys:
@1cbyc — welcome, and thank you for picking this up. **The production change is exactly the fix I had in mind**, down to using the shared `utils/terminal.py` helper rather than a private copy. I approved your CI manually (first-contribution gate) — it is running now, and one job has already gone red for a reason you can fix in thirty seconds.

## The CI failure — fragment format

```
ChangelogError: changelog.d/0.75.0/904.fixed.md:
```

Yours opens with bare prose. While you are in there, the house form also carries the credit — compare `changelog.d/0.75.0/882.fixed.md`:

```
- **`soup mcp serve`'s `data_validate` no longer reports every row valid for a format it does not recognise (#878 by @Srinivasan8888 in #882).** ...
```

So something like:

```
- **`soup draft distill` and `soup shrink` report a validation failure in their own rendered config as a CLI error rather than a traceback (#895 by @1cbyc in #904).** Both commands call `load_config_from_string` on YAML they generated themselves, to validate before spending a training subprocess. A `ValueError` there — a field-name typo in the template, or a schema change such as v0.75.0's unknown-key refusal — surfaced as an unhandled traceback. Both call sites now print the message through `utils/terminal.py:for_terminal` and exit 1.
```

That is the only thing standing between this and a merge on the code side.

## One test gap, and it is the half the issue specifically asked for

Your tests pin the exit code and the message — good, and I confirmed they are not vacuous in the way I first suspected: `load_config_from_string` is imported *inside* the function (`draft.py:386`, `shrink.py:556`), so `monkeypatch.setattr("soup_cli.config.loader.load_config_from_string", ...)` genuinely intercepts. A lot of people get bitten by that; you did not.

But the `for_terminal` half is **not pinned**. I ran the mutation:

```
sed -i 's/{for_terminal(exc)}/{exc}/g' src/soup_cli/commands/draft.py src/soup_cli/commands/shrink.py
pytest tests/test_v07129.py tests/test_v07133.py -k validation_error
  -> 2 passed
```

Both tests still pass with the hygiene removed, because the fixture message (`"invalid rendered config"`) contains nothing for `for_terminal` to strip. #895's fix path names that helper specifically, so it is worth a test that would notice its absence: raise a `ValueError` whose message carries a control byte and a Rich markup tag, and assert the printed text has neither. Something like:

```python
ValueError("bad \x1b[2J[bold red]field[/]")
```

then assert `"\x1b"` is absent from what was printed and that the surrounding text survives. Keep your existing exact-match assertions; just add one case with a hostile message.

## Two smaller notes

- Your test asserts on the whole `messages` list with `==`. That is precise, which I like, but it will break on any unrelated line added to that path later. Not a change request — just be aware you chose the strict form deliberately.
- `draft.py` and `shrink.py` both still carry their own private `_for_terminal` copies, which is what **#896** exists to consolidate. Your PR adds the shared import alongside the private copy, which is correct for now; I have flagged to whoever takes #896 that they should rebase after this lands.

## What I verified

`pytest tests/test_v07129.py::TestRunHeal tests/test_v07133.py -k "validation_error or heal or distill"` — **38 passed**. Fragment is in the right directory (`changelog.d/0.75.0/`), which many open PRs currently are not. The diff is four small hunks and touches nothing else.

Push the fragment fix and the hostile-message test and I will merge. Nice first PR.

1cbyc:
Addressed in e042bd8.

- Fixed `changelog.d/0.75.0/904.fixed.md` to use the required Markdown list-item format and included the issue/PR attribution.
- Added hostile-message regression coverage for both `soup shrink` and `soup draft distill`, asserting terminal control bytes are removed and Rich markup is escaped.
- Focused validation: 4 passed (`pytest --no-cov tests/test_v07129.py tests/test_v07133.py -k "rendered_config_validation_error or sanitizes_rendered_config_validation_error"`).

The commit is signed and GitHub reports it Verified.

MakazhanAlpamys:
Verified at `e042bd8`. Both items addressed, and the mutation that survived last round is dead.

```
replace {for_terminal(exc)} with {exc} at BOTH sites   (2 anchor matches)
-> FAILED TestRunHeal::test_run_heal_sanitizes_rendered_config_validation_error
-> FAILED TestDraftDistillCli::test_run_distill_sanitizes_rendered_config_validation_error
   2 failed, 221 passed, 1 skipped
```

One failure per site rather than one shared assertion, so a half-revert cannot slip through — that is the property I was missing before, and it is what makes this mergeable. `223 passed, 1 skipped` clean. Fragment now starts with `- ` and carries the attribution.

**First, an apology that is mine to make.** You pushed this at **2026-09-12T16:07** and it has been sitting for about nineteen hours, not because I was thinking about it but because the CI run went into `action_required` and I never approved it. My routine for catching those filtered the most recent runs only, and by the time I looked, yours had fallen out of that window — so from my side the PR simply had "no CI" and from your side it looked like silence. I have now queried by status instead of by recency, found seven stuck runs going back to 2026-09-09, and approved yours; the other six belong to a merged PR and two closed ones.

That is a process failure with a cost I can name: a contributor who has done the work waits a day for a button. I have changed how I check.

The run is approved and executing now. I will merge as soon as it is green — nothing else is outstanding on this PR.

One thing worth saying about the fix itself, since it is easy to undersell: `escape()` neutralises `[...]` markup and **nothing else**, so an exception message carrying a raw ESC or OSC byte reaches the terminal live. The rendered config in `soup draft distill` and `soup shrink` is generated by Soup itself, so this is a low-severity instance — but it is the same class that #885 closed this morning in `soup ingest`, and routing through the shared `utils/terminal.py:for_terminal` rather than adding a private helper is exactly right, because #907 is currently removing nine such private copies.

MakazhanAlpamys:
Verified `8350ae60`. **Merge-ready** — both blocking items are closed, and I re-ran the mutation myself rather than taking the earlier round on trust.

**The `for_terminal` mutation that survived last time is now dead, once per call site.** Anchor counted at exactly 1 in each file, both modules import after mutating, then:

```
FAILED tests/test_v07129.py::TestRunHeal::test_run_heal_sanitizes_rendered_config_validation_error
FAILED tests/test_v07133.py::TestDraftDistillCli::test_run_distill_sanitizes_rendered_config_validation_error
2 failed, 221 passed
```

One failure **per site** is the part that matters — a half-revert cannot slip through a shared assertion. Reverting gives 223 passed. `typer.Exit(code=1)` to `code=0` kills 4, and swapping `for_terminal` for `strip_control` kills 2, so the markup-escape half is independently pinned rather than riding on the control-strip half. Baseline on `origin/main` with your tests kept: 4 failed.

The ANSI trap is genuinely not in play here, and I checked rather than assumed: your tests capture the **argument** to `console.print` through a monkeypatched `print`, i.e. the pre-render markup string — no `result.output`, no `capsys`. Under `FORCE_COLOR=1` the only red is `TestDraftMeasureCli::test_below_min_acceptance_exits_two`, which I reproduced on `origin/main` with both `src/` and the test file taken from main. Pre-existing, correctly untouched.

Also verified: fragment at `changelog.d/0.75.0/904.fixed.md` with house credit; no test lines removed; the `8350ae60` merge pulled in #907 and resolved correctly — one shared `from soup_cli.utils.terminal import for_terminal, strip_control` per file and **zero** private `_for_terminal` copies left behind; `raise ... from exc` preserves the chain; exit 1 matches both commands' 0/2/1 taxonomy.

**One thing to be careful about:** your own verification run was on `e042bd8`, which predates that merge commit. Please do not treat it as the signal — the current head's 14 checks are queued, and that is what I will merge on.

Non-blocking, for the record: three other `load_config_from_string` call sites on self-rendered text stay unguarded (`mcp_server/registry.py:452,660`, `ui/app.py:454,475,836`). #895 names only the two CLI sites and you fixed exactly those, which is the right scope — I will file the rest separately rather than widen this.

MakazhanAlpamys:
Re-verified against current `main` (`b0a63382`), 16 merges after my review. **Unchanged and still merge-ready:** `merge-tree` clean, 223 passed and 1 skipped on both the PR head and the merged tree, ratchets 138 passed, fragment intact with no duplicate number.

The mutation still dies per call site — replacing the `console.print(f"[red]Invalid rendered distill config:[/] ...")` in `draft.py` with `pass` fails both `test_run_distill_reports_rendered_config_validation_error` and the sanitisation test.

Worth recording for whoever mutates this file next: `except ValueError as exc:` has **five** matches in it and is not a valid anchor. The print statement is the unique one.

Nothing needed from you.

MakazhanAlpamys:
Merged as `52641e99`. Thank you — and welcome; this is your first contribution here, and you are now in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md).

**The thing I want to name is the one you did not have to do.** The message goes through the shared `utils/terminal.py:for_terminal`. Six command modules in this repo still carry a byte-identical private copy of that control-strip table instead — it is an open ratchet (#896), and the most natural way to write your patch would have added a seventh. You imported the shared one, and you proved it: one of the two tests per command feeds a raw `\x1b[2J` plus live Rich markup and asserts on the sanitized result. A test that only checked the happy message would not have caught a regression there.

The other detail I appreciated is that you did not invent the intent. Both call sites already carried `# validate before spending a subprocess`, so the fail-fast was designed and only the handling was missing — your diff is four lines per command and leaves the happy paths untouched. That is the correct size for this.

Verified here rather than read off the diff: mutation-checked by deleting the `except` in `shrink.py` — anchor matched exactly once, module imports clean afterwards — which fails exactly your two `TestRunHeal` tests on behavioural asserts, not on collection. 14/14 checks green on `8350ae60`, no test deleted or weakened, fragment clean.

One limit I recorded rather than asked you to close: the tests raise from a patched loader, so they pin the handling branch but not that either template can actually produce an invalid config. That is fine for this fix — the branch is what #895 was about — but if you ever want a follow-up, feeding `_run_heal` a genuinely malformed render would be the stronger version.

Issue #886, opened 2026-09-12
v0.75.0 added three tests that fail whenever colour is on, including one assertion that is unsatisfiable by construction


Srinivasan8888:
Three tests in `tests/test_issue627_unknown_config_keys.py` pass with colour off and fail with colour on. You run with colour — your #763 review reports *"30 passed under `FORCE_COLOR=1 TERM=xterm-256color`"* — so you will hit these.

Found while sweeping `main` after v0.75.0. **No behavioural fallout from the release**; this is the test, not the code.

## Reproduction on clean `main`

```
TERM=dumb        pytest tests/test_issue627_unknown_config_keys.py   69 passed
FORCE_COLOR=1    pytest tests/test_issue627_unknown_config_keys.py   3 failed, 66 passed
```

```
:1097  test_markup_and_control_bytes_in_a_key_are_neutralised[warn]    a raw ESC byte reached the terminal
:1097  test_markup_and_control_bytes_in_a_key_are_neutralised[error]   same
:821   test_a_valid_sweep_reaches_the_first_arm                        the sweep never reached its first arm
```

All three flip on colour alone — `TERM=dumb` passes each, colour fails each.

## The assertion at :1097 cannot pass under colour

```python
assert "\x1b" not in out, "a raw ESC byte reached the terminal"
```

Rich's own SGR codes **are** ESC bytes, so when Rich colourises, `out` contains `\x1b` before the payload is considered at all. The assertion cannot distinguish the injected escape from the formatting around it, so it is unsatisfiable whenever colour is on — including on CI, if `FORCE_COLOR` is ever set there.

`git log -S` on that exact line gives **`f50a7da` — release: v0.75.0 (#879)** as the commit that added it.

## The fix, verified rather than proposed

Assert the absence of the specific dangerous sequences instead of ESC in general. Measured against a real `Console(force_terminal=True)` with the hostile payload:

```
broad   assert "\x1b"   not in out   ->  FAIL   (Rich's own SGR)
narrow  assert "\x1b[2J" not in out  ->  PASS   (erase-display)
narrow  assert "\x1b]0;" not in out  ->  PASS   (OSC title-set)
narrow  assert "\x07"   not in out   ->  PASS   (BEL)
```

[...]

Issue #794, opened 2026-09-11
PiSSA, OLoRA, LoftQ and VeRA never reach a trainer: use_vera r=256 trains 524,288 params where VeRA would train 2,048


MakazhanAlpamys:
## The defect

Four documented PEFT variants are validated by the schema and then dropped. Every trainer builds a plain, randomly initialised LoRA:

| Field (`config/schema.py`) | What the docs promise | What trains |

|---|---|---|

| `lora.init_strategy: pissa` (`:129`) | SVD init of A/B from the base weight | random-init LoRA |

| `lora.init_strategy: olora` / `lora.use_olora: true` (`:103`) | QR (orthonormal) init | random-init LoRA |

| `lora.init_strategy: loftq` + `loftq_iter` / `loftq_bits` (`:141-153`) | quantisation-aware init | random-init LoRA |

| `lora.use_vera: true` (`:95`) | shared frozen random matrices, tiny trainable vectors | a full LoRA at the same `r` |

## Evidence

The two helpers that know about these fields have **no caller** in `src/`:

```

$ grep -rln "peft_builder\|loftq_init" src/soup_cli --include=*.py

(no output)

```

- `utils/peft_builder.py:14` `build_peft_config` is the only code that emits `init_lora_weights` or selects `VeraConfig`. Even it never forwards `loftq` (`:61` handles only `pissa` / `olora`).

- `utils/loftq_init.py:52` `build_loftq_config` builds a `LoftQConfig` that nothing uses.

- `git log -S build_peft_config -- src/soup_cli/trainer src/soup_cli/commands` and `git log -S init_lora_weights -- src/soup_cli/trainer` both return nothing, so no trainer has ever called either.

What the trainers use instead:

- SFT and pretrain call `build_lora_config_kwargs` (`utils/peft_wiring.py:106-129`), which forwards `r`, `alpha`, `dropout`, `use_dora`, `use_rslora`, `rank_pattern` and `alpha_pattern`, but not `init_strategy`, `use_olora`, `use_vera` or `loftq_*`.

[...]

PR #763, opened 2026-09-09
feat(adapters): audit a finished run against the config that started it (#762)

Srinivasan8888:
Closes #762.

## Why

Four merged fixes taught the MLX path to record what it **actually did** into `adapter_config.json` — #683 (masking), #684 (accumulation), #685 (grad checkpoint), #686 (optimizer and schedule), #749 (clipping). Each existed as a defect where a setting was accepted and dropped without a word, so the record is the remedy.

**Nothing read it back.** `adapters info` prints LoRA shape, `diff` compares weights, `verify` checks signatures, `scan` looks for backdoors. A user asking *"did my run do what I asked?"* had to open two files and compare by eye — which is how #683 and #686 survived as long as they did.

## What it looks like

Real output, run live rather than asserted on:

```
$ soup adapters audit ./out --config soup.yaml
  setting                        asked         ran
  ─────────────────────────────────────────────────────────────────
  optimizer                      adamw_torch   SGD               DIVERGED
  scheduler                      cosine        cosine            ok
  warmup_ratio                   0.03          0 of 12 updates   DIVERGED
  weight_decay                   0.01          0.01              ok
  max_grad_norm                  1.0           1.0               ok
  gradient_accumulation_steps    4             4                 ok
  data.train_on_responses_only   True          True              ok
  lora.r                         8             8                 ok

  optimizer: 'adamw_torch' resolves to 'AdamW' on this backend, and the run used 'SGD'
  warmup_ratio: 0.03 x 12 optimizer updates rounds to 0, so training started at
    the peak learning rate; raise warmup_ratio or lower gradient_accumulation_steps

  2 divergence(s)
$ echo $?
1
```

## The load-bearing rule: unknown is never agreement

A setting the record cannot speak to is `unknown`, **never** `ok`. A false clean bill converts *"I do not know"* into *"I checked"* — the substitution this command exists to undo. For the same reason `unknown` does not fail the command: an older adapter predates the keys and that is not the user's fault, so it must be visible rather than fatal. Exit is non-zero only for real divergence, so this composes into CI and `soup ship`.

## The warmup case is why it earns its place

`warmup_ratio: 0.03` over twelve optimizer updates is zero warmup steps. #686 warns at the time; if nobody was watching the terminal there was no way to learn it afterwards.

[...]

PR #906, opened 2026-09-12
test(config): a read inside a function nothing calls is not consumption (#807)

### LINKED ISSUE #807: The #748 consumer guard counts a read inside a function nothing calls: training.warmup_auto passes it and does nothing
## The defect

`tests/test_issue748_config_fields_reach_a_consumer.py` treats a config field as consumed if its name is read anywhere under `src/` outside `schema.py` (`consumed_names`, line 103). A read inside a function that no code calls counts the same as a read in a trainer. The file already states the principle this breaks, for `schema.py` properties only (line 51):

> an UNCALLED resolver launders nothing, or the detector's own failure mode returns one level up, with a field "read" by code that never runs.

`fold_property_reads` (line 204) enforces that gate for `@property` resolvers. Ordinary functions get no such gate.

A second gap: `_declared()` (line 330) enumerates only `TrainingConfig` and `DataConfig`. `LoraConfig` (and every other nested model) is outside the guard's scope.

## Escape 1: training.warmup_auto

Documented at `docs/peft-and-efficiency.md:540-548` ("Pick warmup_steps from dataset_size x epochs x warmup_ratio ... Clamped to [10, 1000]"). What reads it:

```

$ grep -rn "warmup_auto" src/soup_cli --include=*.py | grep -v config/schema.py

src/soup_cli/autopilot/generate_config.py:166:        "warmup_auto": bool(decisions.get("warmup_auto", False)),

```

That one read:

- sits in `generate_config()`, which nothing in `src/` calls (callers: `tests/test_auto_tuning.py:741, 767`);

- reads the autopilot **decisions dict** passed in, not the config, and then writes the value *into* the config;

- no code in `src/` puts a `warmup_auto` key into that dict.

The clamp helpers are also dead: `compute_warmup_steps` (`utils/warmup.py:19`) is called only by `autopilot.decisions.decide_warmup`, which nothing calls. Every trainer computes `int(total_steps * warmup_ratio)` itself, so `warmup_auto: true` changes nothing. `tests/test_auto_tuning.py:719` asserts the emitted YAML contains the key, with the docstring "Generated config has warmup_auto=true so train.py picks it up"; `commands/train.py` never reads it.

## Escape 2: LoRA variants

`lora.use_vera`, `lora.use_olora` and `lora.init_strategy` are read only by `utils/peft_builder.build_peft_config`, which nothing in `src/` calls, so VeRA/PiSSA/OLoRA/LoftQ are never applied (filed separately with a reproduction). The guard misses them twice over: `LoraConfig` is not in `_declared()`, and if it were, the uncalled builder would count as the consumer. `lora.loftq_iter` / `loftq_bits` have no read at all and would be flagged at once.

## Measured: a function-level gate is cheap and precise

Prototype, added on top of the guard's own `consumed_names` semantics: for every read, find the enclosing top-level function, and drop the read if that function's name is referenced nowhere in `src/` (no `Name` or `Attribute` load). Run over `TrainingConfig` + `DataConfig` + `LoraConfig`, excluding `KNOWN_UNCONSUMED`:

```

fields whose every read sits in a function no module references: 4

  training.lora.init_strategy [('utils/peft_builder.py', 'build_peft_config')]

  training.lora.use_olora [('utils/peft_builder.py', 'build_peft_config')]

  training.lora.use_vera [('utils/peft_builder.py', 'build_peft_config')]

  training.warmup_auto [('autopilot/generate_config.py', 'generate_config')]

LoraConfig fields with no read at all: ['training.lora.loftq_iter', 'training.lora.loftq_bits']

```

All four are real defects, with no false positives across the ~275 existing fields. The name match is global, so it errs toward "referenced", which is the safe direction for a guard that must not be deleted for crying wolf.

## Why not module reachability

#755 measured and rejected reachability over the *import* graph for a related question: every trainer's transitive closure is most of the package, so it separates nothing. The proposal here is narrower: whether the *function* containing the read is referenced at all. `peft_builder.py` and `generate_config.py` are imported, and it is their functions that are dead, which is exactly what an import graph cannot see.

## Proposed fix

1. Extend `consumed_names` to record the enclosing top-level function of each read, and count a read only if that function name is referenced somewhere in `src/` outside its own definition (module-level reads count as live). Apply the same rule `fold_property_reads` already applies.

2. Add `LoraConfig` to `_declared()` under `training.lora.*`; consider `EvalConfig` / `EvalGateConfig` / `ShipConfig` too.

3. Seed `KNOWN_UNCONSUMED` with whatever the stronger detector finds (the six fields above today), each with its issue number, so the list can only shrink.

4. Pin the new gate with a refusal test: a synthetic module whose only read is inside an uncalled function must be reported as unconsumed.

## Acceptance criteria

- [ ] `training.warmup_auto` is reported as unconsumed (or listed in `KNOWN_UNCONSUMED` with a reason); the test fails on current code without that entry.

- [ ] `training.lora.use_vera`, `use_olora`, `init_strategy`, `loftq_iter`, `loftq_bits` are in scope and reported.

- [ ] `TestTheDetectorItself` gains a case where the only read is in an uncalled function, asserted unconsumed, and a control where the same function is called, asserted consumed.

- [ ] The module docstring's "What this cannot see" list is updated to say what the gate does and does not catch (a function called only from another dead function still passes).

---

- #794 — PiSSA, OLoRA, LoftQ and VeRA never reach a trainer: use_vera r=256 trains 524,288 params where VeRA would t...
- #808 — 28 "staged" config fields are accepted silently and read by nothing, most with docs, some promising release...
- #825 — quantization: bitnet_1.58 is documented as live, but nothing routes to BitNetTrainerWrapper: an SFT run die...

Srinivasan8888:
Closes #807. A defect in a guard I wrote, and the issue quotes my own file's principle back at it correctly.

`fold_property_reads` gates a `schema.py` `@property` on being called, with a docstring saying an uncalled resolver must launder nothing *"or the detector's own failure mode returns one level up, with a field read by code that never runs."* I wrote that for properties and never applied the same reasoning to ordinary functions.

## The escape

`training.warmup_auto` passed the guard on one line in `autopilot.generate_config`, which **nothing in `src/` calls**. That line also reads the autopilot *decisions dict* and writes the value **into** a config — so it is not a read of the field either, which is the same read/write distinction the guard already makes and missed here because the name sits on the value side.

## My first prototype had two false positives, and that is the important part

Reproducing the issue's measurement gave **six**, not four:

```
training.auto_mixed_precision   [generate_config]      <-- WRONG
training.multipack              [build_multipack...]   <-- WRONG
training.lora.init_strategy     [build_peft_config]
training.lora.use_olora         [build_peft_config]
training.lora.use_vera          [build_peft_config]
training.warmup_auto            [generate_config]
```

`auto_mixed_precision` is genuinely live — `trainer/sft.py:1185` reads it. My walker only visited top-level `tree.body` defs, so a read inside a **class method** was invisible and the field looked dead. Attributing methods gives exactly the four you reported.

That is why the attribution matters more than the gate: a guard that cries wolf on live code gets deleted, and my version would have flagged two live fields on its first run.

The gate is lenient in the other direction by design, both pinned: **one live read rescues** a field that also has dead ones, and a **module-scope read always counts**, because module scope runs on import.

## LoraConfig

Was outside `_declared()` entirely, so every LoRA-variant field was unguarded. Adding it surfaces `loftq_iter` and `loftq_bits`, which have no read anywhere at all — exactly as you predicted.

## Allowlisted, not left failing — your call to reverse

Six entries, each citing its issue: four dead-read (#807), two no-read (#794). Pin 39 → 45.

I took the allowlist option I flagged on the issue, because a guard whose own landing reddens `main` tends to get reverted rather than fixed, and #794/#808 already track the substance. If you would rather these fail loudly, it is deleting six lines and changing one number.

## Mutations — 5 applied, 5 killed

Baseline `b905f2f5f0db` restored exactly:

| mutation | result |
|---|---|
| gate off (dead reads count again — the defect) | 3 failed |
| gate too aggressive (ANY dead read kills a field) | 4 failed |
| methods not attributed (the false-positive shape) | 3 failed |
| module-scope reads treated as dead | 1 failed |
| `LoraConfig` back out of scope | 1 failed |

M1 is the bug; **M2 is the false-positive direction** that would get the guard deleted. Both needed covering.

## Verification

```
tests/test_issue748_config_fields_reach_a_consumer.py   36 passed
+ test_issue627_unknown_config_keys.py, TERM=dumb      105 passed
ruff clean, test-only, no src/ change
```

With colour on, three tests in `test_issue627` fail — that is #886, unrelated to this PR, and `TERM=dumb` is green.

`git diff -- tests/ | grep -c "^-[^-]"` reports **6**, and all six are modified lines rather than deleted assertions — the fold expression now routed through the gate, the import and loop header extended with `LoraConfig`, and the pin. Enumerated in the commit message. No assertion was weakened.


MakazhanAlpamys:
Reviewed at `8fc31b5f`; 14/14 green on that SHA. This is a precise piece of work and the write-up is unusually honest about the false-positive prototype you threw away. The point that a guard which cries wolf gets deleted is the right one — this repo has the receipt for it, in the `SCORER_CHANGED_IN_V0_73_2` warning that fired on the correct case and was removed rather than tuned. Attributing methods rather than skipping them is what makes the gate safe to land.

**Verified independently rather than read off the diff.** The gate drops 295 names from a 3401-name consumed set, of which **exactly 4 are declared config fields** — your four — so zero false positives across roughly 275 declared fields. Scope is test-only (`git diff --stat` shows no `src/` change), and `git diff origin/main...HEAD -- tests/ | grep -c "^-[^-]"` is **6**, every one a modified-in-place line paired with a `+`. **No assertion was deleted or weakened.**

**Mutations, all three you were asked for plus two more:**

| mutation | result |
| --- | --- |
| gate removed from `_consumed_in_src` (pre-PR behaviour restored) | **1 failed** — names all four escapes |
| new `TrainingConfig` field whose only read sits in an uncalled function | **2 failed** — `assert not ['training.zzz_probe_field_b']` |
| same field, but the function gains a caller | **passes** — the gate discriminates on deadness, not on newness |
| same src mutation against `origin/main`'s version of this file | **30 passed** — the old guard is blind to it |
| real tree, unmutated | **36 passed**, no false positives |

That fourth row is the cleanest demonstration that the capability is genuinely new rather than a re-spelling, and it is the one I would have asked for if you had not made it easy to produce.

**One thing before merge, and it is from your own issue's acceptance list.** #807 line 63 is an unchecked box:

> - [ ] The module docstring's "What this cannot see" list is updated to say what the gate does and does not catch (a function called only from another dead function still passes).

The diff's first hunk starts at line 231; the docstring ends around line 80 and is untouched. This file's whole premise is that nobody trusts it past its stated limits, so an undocumented new limit is precisely the debt it exists to prevent. Please add a short paragraph after the `fold_property_reads` one covering (i) a function called only from another dead function still passes, and (ii) the name match is global and errs toward "referenced".

While you are in there, one measurement worth recording in that same paragraph, because it is the shape a future reader will trip on: of **132** `@app.command()`-decorated functions in this tree, **99 are absent from `referenced_names`** — decorator registration leaves no `Name`/`Attribute` load, so a config field read *only* inside such a body would be dropped as dead. Impact today is zero (every top-level entry point is referenced through `cli.py`, and no declared field is affected), which is exactly why it belongs in "what this cannot see" rather than in the allowlist.

**Two non-blocking observations:**

1. **The gate's wiring is pinned only by the allowlist contents, and that pin is scheduled for deletion.** Removing the gate from `_consumed_in_src` goes red only via `test_the_allowlist_does_not_cover_fields_that_are_consumed` — all five `TestTheDeadFunctionGate` cases call `drop_dead_function_reads` directly and pass unchanged under that mutation. When #794/#808 are fixed and those four entries retire, which is the stated goal, that mutation goes **silent**. One test asserting `_consumed_in_src()` itself excludes a dead-read name would pin the composition independently — and it would satisfy the issue's "synthetic module end-to-end" criterion at the same time, which the current five cases approximate with hand-built dicts.
2. `test_the_real_tree_has_exactly_the_four_known_escapes` says "exactly the four" and its docstring says the number is "pinned so it cannot drift unnoticed", but it asserts membership only, and `dropped` actually holds 295 entries. Asserting the four declared field paths exactly would be stable and would match the name.

**On the allowlist growing 39 to 45:** I am taking it rather than letting `main` go red. Each entry cites its issue, which is what this file's accountability rule asks for, and the size pin makes the growth visible rather than silent. You explicitly offered to reverse it — noted, and no, keep it.

Push the docstring paragraph and I will merge.

MakazhanAlpamys:
@Srinivasan8888 — status, and a base move rather than a review item.

This is now **conflicting** on exactly one file:

```
```

caused by `8b79489d` (#927, "stop writing training gates nothing reads"), which edited the recorded-leak table in the same region your change touches. It is a one-file rebase — everything else auto-merges.

Your required matrix was 14/14 green, so the change itself is fine.

The substantive items from my **2026-09-12** review are still the ones I am waiting on; the conflict does not replace them. If you have moved on from this one, say so and I will pick the remaining items up myself rather than leaving it open — that is a real option, not a hint.

MakazhanAlpamys:
@Srinivasan8888 — status. **CI went green on `8fc31b5f`, and that does not unblock it**, so I would rather tell you than let the tick sit there looking like progress.

Two things, both small:

**1. Still conflicting, one file.** Re-checked against current `main` just now:

```
git merge-tree --write-tree --name-only origin/main <head>
STILL CONFLICTS: tests/test_issue748_config_fields_reach_a_consumer.py
```

Same single file as before, from `8b79489d` (#927) editing the recorded-leak table in the region you touch. Everything else auto-merges.

**2. The only substantive item left is one paragraph**, and it is the measurement you already have: of **132** `@app.command()`-decorated functions, **99 are absent from `referenced_names`** — decorator registration leaves no `Name`/`Attribute` load, so a config field read only inside such a body would be dropped as dead. Impact today is zero, which is exactly why it belongs under "what this cannot see" rather than in the allowlist. That was my ask and it is still the whole ask.

The two observations from that review remain non-blocking and unchanged:

- The gate's wiring is pinned only by the allowlist contents, and that pin is **scheduled for deletion** — when #794/#808 are fixed and those four entries retire, the mutation that currently catches removing the gate from `_consumed_in_src` goes silent. One test asserting `_consumed_in_src()` itself excludes a dead-read name would pin the composition independently, and would satisfy the issue's "synthetic module end-to-end" criterion that the five hand-built-dict cases currently approximate.
- `test_the_real_tree_has_exactly_the_four_known_escapes` says "exactly the four" and asserts membership only, while `dropped` holds 295 entries.

**A date, so this stops ageing:** if I have not heard by **2026-09-17 18:00 UTC** I will do the rebase and write the paragraph myself, on top of your commits, and say so here. That is a real date — I have let announced ones pass on this repo before and I am not doing it again. Your name stays on the PR either way, and the paragraph would be quoting your own measurement back.

Nothing about the change itself is in question. And the decision from last time stands: the allowlist growing 39 to 45 is taken deliberately rather than reversed — each entry cites its issue, and the size pin makes growth visible instead of silent.

MakazhanAlpamys:
@Srinivasan8888 — **the paragraph I asked you to write contained three wrong numbers and one false statement.** I prepared the work myself ahead of the date I set, which is how I found out. Correcting the record before anything else.

## The numbers

I said **132** `@app.command()`-decorated functions and **99** absent from `referenced_names`. Measured, and verified independently by two methods:

```
grep -rnE '^\s*@[A-Za-z_]*app\.command\(' src/soup_cli | wc -l
150

by receiver:  app 145   env_app 4   runs_app 1
```

**150 decorated, 115 absent, 35 present.** And this is not drift from today's merges — identical at `f50a7da3` (the commit you branched from) and at current `main`, 502 modules both times. Restricting the predicate to literally `app` gives 145/110, which is also not 132/99. I cannot tell you where my numbers came from; only that no predicate at either revision produces them.

I also said `dropped` holds **~295** entries. It holds **280**, of which the test asserts membership of four, leaving 276 unasserted.

## The false statement, which matters more

I asked you to write *"every top-level entry point is referenced through `cli.py`"* as the reason the impact is zero. **It is false.** Zero of the 115 are named anywhere in `cli.py` — they are registered through the decorator and reached via the Typer app object, never by name. I sampled it directly:

```
adapter_bisect   in cli.py: 0
adapter_pr       in cli.py: 0
advise_run       in cli.py: 0
aider            in cli.py: 0
apply_steer      in cli.py: 0
```

So I would have had you put a false claim into the guard's own voice, in the section that exists to tell the next reader what the guard cannot see. That is worse than a wrong number.

**The justification that actually holds** is the gate's own leniency rule:

- **19** declared fields have a read inside such a body;
- **18** also have a live read elsewhere and are rescued by the one-live-read rule;
- the 19th, `training.grace_codebook`, is already in `KNOWN_UNCONSUMED` as a genuine orphan for an unrelated reason — a name collision with an artifact-kind string;
- **0** declared fields have *every* read in such a body.

Impact is genuinely zero, but because of the leniency rule, not because of `cli.py`.

## The rebase is prepared, not pushed

Branch ready here, and the conflict turned out to be **one hunk and only the pin line** — every allowlist entry, both helper functions and all new tests from both sides auto-merged. `test_the_allowlist_size_is_pinned_exactly` asserts `len(KNOWN_UNCONSUMED)` with `==`, so both sides moved it from 39: #927 to 40, yours to 45.

**Resolved to 46, and not by addition** — I imported the merged module and measured `len(KNOWN_UNCONSUMED) == 46`, with #927's `training.forgetting_threshold` present, all six of yours present, and #927's issue re-attributions intact.

Two things worth knowing:

- **The layers compose without further edit.** `_consumed_in_src` applies #927's property fold, then your dead-function gate; #927's `field_reaches_a_consumer` sits above both and short-circuits the one receiver-qualified case. Neither gate shadows the other.
- **The resolution lives in your rebased commit, not in mine on top.** Unavoidable with a rebase, and correct — taking either side's number would leave that commit red on its own and break bisect. Both commits are green individually.

Mutation re-confirmed on the rebased tree: replacing `drop_dead_function_reads`'s body with `return set(consumed)` — anchor matched exactly once, module imported and the gate called to prove it — fails three named tests behaviourally, and reverting to a byte-exact checksum gives 39 passed.

Your two non-blocking observations are both confirmed, and the first is sharper than you put it: removing the gate from `_consumed_in_src` goes red via **exactly one** test, and the four entries it names *are* the entire pin — `training.lora.init_strategy`, `use_olora`, `use_vera`, `warmup_auto`. When #794/#808 land and those retire, that mutation goes silent.

## What you decide

The branch is not pushed and I am not pushing it before **2026-09-17 18:00 UTC**. If you would rather finish it yourself, everything above is yours to use — the corrected numbers especially, since two of them came from me and were wrong.

Not run: the full suite. I killed it deliberately; the diff is test-only and confined to one test file plus a fragment, and CI is the signal for the rest.

Srinivasan8888:
Rebased and pushed as `66abdfe`. **Apologies for the delay** — you held this since the 14th waiting on me.

Taking the correction first, since it is yours to be told about: thank you for catching your own numbers before I wrote them into the guard. I would have shipped them without checking, because they arrived in a review and I have been treating your measurements as measured. That is my failure of process, not yours.

## The conflict resolved to 46, measured not assumed

One hunk, only the pin line, exactly as you said — every allowlist entry, both helpers and all new tests from both sides auto-merged.

```
KNOWN_UNCONSUMED after resolution        46
training.forgetting_threshold (#927)     present
pytest tests/test_issue748...py -q       39 passed
```

I imported the merged module and measured rather than adding 40 + 6, and 39 passed matches your number.

One note for anyone reproducing it: the guard is path-sensitive, and running it from the repo root on this machine gives 5 failures that are pure environment — it scans `Desktop/.../Soup` and imports `desktop/.../soup`, and the guard says so itself rather than passing silently, which is good design. From a worktree it is 39 passed. Worth knowing before someone reads those five as a broken rebase.

## Your corrected numbers, independently confirmed

```
grep -rnE '^\s*@[A-Za-z_]*app\.command\(' src/soup_cli | wc -l    150
by receiver:  app 145   env_app 4   runs_app 1
```

## One number I could not reconcile, so I did not write it down

You give **19** declared fields with a read inside such a body, **18** rescued, **0** with every read in one. Measured by declared leaf name I get **44 / 40 / 4**, and the four are `init_strategy`, `use_olora`, `use_vera`, `warmup_auto` — the same four you name in your closing note as the entire pin for the other mutation, which is a good sign the mechanism is the same and only the predicate differs. Presumably you counted fully-qualified dotted paths with receiver qualification where I counted leaves.

Rather than pick one, the paragraph names the four fields and does not lean on a ratio at all. A docstring number that drifts is precisely what this file exists to prevent, and neither of us can currently say which predicate the reader should assume. If you would rather it carried your figures, tell me the predicate and I will measure it that way and use it.

## What the paragraph says now

The justification is the one-live-read rule, not `cli.py`: a field is dropped only when *every* read sits in an unreferenced function, and one live read rescues it.

The decorator point is kept, but demoted from justification to limit — which I think is where it belonged all along:

> Note what this gate does **not** know: whether a function is reachable at runtime. `referenced_names` is a syntactic check for the name appearing somewhere other than its own definition. A `@app.command()`-decorated function is referenced by its decorator, so Typer entry points are safe by accident rather than by design — if that decorator form ever changed, every command body would look dead to this gate at once.

That reads as a thing the guard cannot see, which is what the section is for.

## Verification

```
pytest tests/test_issue748_config_fields_reach_a_consumer.py -q    39 passed
ruff check tests/                                                  All checks passed
git diff -- tests/ | grep -c "^-[^-]"                              0
```

Fragment is at `changelog.d/0.75.0/906.fixed.md`, correct for the current tag.

Full suite not re-run yet — it is going now and I will post the number. Given the diff is one test file plus a docstring I do not expect it to say anything new, but I would rather state it than assume it.

Srinivasan8888:
Full suite on the rebased tree, as promised:

```
21743 passed, 115 skipped, 5 deselected, 1 failed   in 17:35
coverage 83.58% (gate 77%)
```

**The one failure is pre-existing and not from this branch.** `test_issue580_autodistill_mlx_worker.py::test_worker_verification_survives_a_real_venv_launcher` — it builds a real venv and launches an MLX teacher worker, which exits 1 on this machine. Three things say it is not mine:

- it fails identically on a tree containing none of my changes;
- `git diff origin/main...HEAD --name-only | grep -c autodistill` is `0`;
- the same single failure appears on my `#763` branch, which is unrelated to this one — two independent branches, same test, same machine.

Reporting it rather than rounding to "21743 passed", since a suite line with a failure in it should never appear in a PR without an account of why.

Two process notes from the run, in case they are useful to you or the next contributor:

**The first attempt was killed at 92%** — by system memory, not by a test. I had three full suites going in parallel worktrees and the machine ran out. Re-ran it alone for the number above; I mention it because "killed at 92%, looked fine" is not a result and I did not want to report it as one.

**The guard is path-sensitive in a way worth knowing.** Run from the repo root on this machine it gives 5 failures, because it scans `Desktop/.../Soup` while the editable install resolves to `desktop/.../soup`. From a worktree it is 39 passed. The guard detects the mismatch and says so rather than silently passing, which is the right behaviour — but the message reads like a broken rebase if you hit it cold.

Srinivasan8888:
Correcting myself on the path-sensitivity note above. I said the guard "detects the mismatch and says so rather than silently passing, which is the right behaviour". **It is not right behaviour — it is a false alarm**, and I should not have waved it through as correct just because it failed loudly.

`:394` is `assert imported == SRC`. macOS is case-insensitive but case-**preserving**, and `Path.resolve()` does not normalise case — it returns the path as spelled. So a shell whose cwd is `/users/.../soup` gives `SRC` in that spelling while the editable install resolves to the canonical `/Users/.../Soup`, and the assertion compares two spellings of one directory and calls them different:

```
both paths exist            True True
resolve() preserves case    True
imported == SRC             False     <- the assertion
os.path.samefile(...)       True      <- same device, same inode
```

The assertion is doing real work — it catches a worktree with no `PYTHONPATH` importing a different tree, which is the silent-green failure this file exists to prevent. It just compares the wrong thing. `os.path.samefile` compares device and inode instead of spelling: same catch for genuinely different trees, no false alarm for a case difference. It needs an `exists()` guard first, because `samefile` raises `FileNotFoundError` rather than returning `False` on a missing path — itself a case this assertion should report rather than crash on.

**Not changing it in this PR.** You have reviewed #906 at a specific SHA with a date on it, and slipping an unreviewed one-liner in now restarts that for no good reason — the false alarm is three days old and costs nothing while it waits. It is queued as a follow-up once this lands, unless you would rather have it folded in here, in which case say so and it is two lines.

Flagging it now rather than after, only so my earlier "that is the right behaviour" does not sit in the thread as an assessment you might rely on.

MakazhanAlpamys:
@Srinivasan8888 — reviewed `66abdfe2`. **Two items, and the first one is not your fault: `main` moved under you again, three hours after you rebased.**

## 1. Conflicting again, and it is a different conflict

You rebased at 05:37Z over #927. **#931 merged at 08:24Z** — three hours later — and retired four `yarn_*` allowlist entries, moving the pin **40 -> 36**. You are at 46.

```
$ git merge-tree --write-tree <main> 66abdfe2
CONFLICT (content): tests/test_issue748_config_fields_reach_a_consumer.py
main   : assert len(KNOWN_UNCONSUMED) == 36
yours  : assert len(KNOWN_UNCONSUMED) == 46
```

**It is one hunk — the pin line only.** Every allowlist entry, both helpers and all the new tests auto-merge. I resolved it and measured rather than assumed:

```
len(KNOWN_UNCONSUMED) = 42        (36 + your 6)
yarn entries present  = []        (#931's retirements intact)
yours present         = warmup_auto, lora.use_vera, lora.use_olora,
                        lora.init_strategy, lora.loftq_iter, lora.loftq_bits
```

So the number is **42**, and nothing of #931's is lost.

This is the second time this file has been moved out from under you, and it will keep happening — `KNOWN_UNCONSUMED` is edited by every PR that wires a field. That is worth knowing rather than being annoyed by: a PR that pins a repo-wide count is structurally exposed to every other merge.

## 2. The docstring paragraph went into the wrong docstring

#807's line-63 checkbox, and my literal ask, was the **module docstring's `What this cannot see` list** — after the `fold_property_reads` paragraph. Your 20 new lines went into `drop_dead_function_reads`'s **own** docstring at `:876-894`. I checked rather than eyeballed:

```
module docstring (lines 17-80), mentions of the new gate : 0
```

And of the two specifics I asked for, **(i) is absent**: a function called only from *another dead function* still passes. Grepping your new paragraph for that case returns nothing. (ii) is half there — you say "a syntactic check for the name appearing somewhere other than its own definition" without saying it errs toward *referenced*, which is the direction a reader needs in order to know which way to distrust it.

Moving the paragraph is the fix; the text you wrote is good and mostly transplants as-is.

**What you got right there, and it is the substantive half:** you did not repeat my retracted `cli.py` justification, and you independently re-measured `150 / 145 / 4 / 1` rather than copying my wrong `132 / 99`. Refusing to write a ratio whose predicate neither of us can define is the correct call for a file whose entire premise is not drifting.

## The 06:17Z self-correction is right, and deferring it was right

You retracted your own "the guard detects the mismatch and says so, which is good design". Correct: `:388 assert imported == SRC` compares two *spellings* of one directory on a case-preserving filesystem, so it is a false alarm, and `os.path.samefile` — with an `exists()` guard, since it raises rather than returning `False` — is the right comparison.

**That line is pre-existing**; your PR only adds `LoraConfig` to the loop three lines below. Deferring it rather than slipping an unreviewed one-liner into a PR reviewed at a pinned SHA is exactly the right instinct. Worth its own issue; say the word and I will file it, or take it yourself.

## What is verified and not in question

**All six removed lines are justified**, each modified-in-place with a paired `+` in the same hunk: two split so the gate composes after `fold_property_reads`, two add `LoraConfig` to the scope, two move the pin. No assertion deleted or weakened — which matches your own accounting.

The mutation kills, run on the **resolved merged tree** rather than the branch: replacing the gate body with `return set(consumed)` — anchor counted exactly once, and the module *executed* to confirm the gate was genuinely off, not merely parsed — gives **3 failed, 36 passed**, all behavioural, including `warmup_auto is read only inside a function nothing references`. Reverted byte-exact to 39 passed.

39 passed on the branch; **188 passed on the resolved merged tree** with the ratchets. Fragment correct and credited, ruff clean, CI 14/14.

## Two non-blocking items still open from the last round

- A test pinning `_consumed_in_src()`'s composition independently of the allowlist. Right now the only thing that catches the gate being removed is `test_the_allowlist_does_not_cover_fields_that_are_consumed` — the pin that is scheduled for deletion.
- `test_the_real_tree_has_exactly_the_four_known_escapes` (`:947-968`) asserts **membership only** (`for expected in (...): assert expected in dropped`) despite its name. Four lines to make it assert exactly four.

---

Rebase to **42** and move the paragraph, and I will merge on green without re-reviewing. Both are small; the timing was the expensive part and that was not yours.

Srinivasan8888:
Pushed as `ec8773e`. Before writing this I checked the remote tree, not my local copy: the pin reads `42`, and the module docstring now carries the new item.

## 1. Rebased onto #931; pin measured at 42

One hunk, only the pin line, as you found. I measured the result rather than adding 36 and 6:

```
len(KNOWN_UNCONSUMED)   42
yarn entries            []     (#931's retirements intact)
branch's six            all present
```

## 2. The paragraph is in the module docstring now

It sits in the "What this cannot see" list, after the `fold_property_reads` item. The function's own docstring is back to two short paragraphs plus a pointer to the list. Mentions of the gate in the module docstring: **0 → 3**.

Before writing your two specifics down, I checked each against a small synthetic tree:

- **(i) Deadness is not transitive.** `dead_inner`, called only from `dead_outer`, counts as referenced, because the dead caller's body loads its name. A field read only inside it is **kept**. Stated.
- **(ii) It errs toward *referenced*, and the paragraph now says which way to distrust it:** "distrust a pass more than a failure". Measured: an unrelated `obj.attr_only_fn` elsewhere rescues a dead function of the same name.

One hole you didn't name turned up in the same probe, so it's in the paragraph too. **A function reached only through a string** (`getattr(module, "name")`, or an entry point declared outside `src/`) is *not* referenced, so its reads are dropped. That's the **opposite** direction: a live field could be reported as unconsumed. On this tree it drops nothing beyond the four known escapes, and the exact test below is what makes that sentence checkable.

I didn't carry the 44/40 counts across. The four names are pinned by a test; the counts weren't.

## The two non-blocking items, both taken

**Exactly four.** `test_the_real_tree_has_exactly_the_four_known_escapes` now asserts equality. `dropped` spans every consumed identifier, not just config fields, so the exact claim is made over declared leaves. Equality over the raw set could never hold. The membership loop stays for its per-field message.

To be precise about what exactness buys: an over-dropping gate (`any` instead of `all`) was **already killed** by four other tests before this change. Exactness adds this test to that list, so its name is now true. It doesn't rescue a mutation that was surviving.

**Composition.** `test_consumed_in_src_applies_the_gate_independently_of_the_allowlist` checks that each escape is read in the ungated set and absent from `_consumed_in_src()`. It no longer depends on the allowlist test that's scheduled for deletion.

| Mutation | Result |
|---|---|
| gate removed from `_consumed_in_src` | **2 failed**: the allowlist test *and* the new composition test |
| gate body → `return set(consumed)` (your mandatory one) | **4 failed**, including both new tests |
| over-dropping gate, with exactness | **5 failed**, including the exact-four test |
| same, exactness removed | 4 failed; the exact-four test not among them |

## Verification

```
tests/test_issue748_config_fields_reach_a_consumer.py   40 passed
changelog fragments        21 passed in 0.83s
licence headers            10 passed in 0.50s
startup-light              34 passed in 12.14s
ruff                       clean
```

The diff removes 18 lines, and all 18 are the docstring prose moving into the module docstring. No assertion was removed or weakened.

MakazhanAlpamys:
Reviewed the delta `8fc31b5f..ec8773e3`. **Every item from the last round is closed, each one checked by execution.** The rebase is clean against today's `main` (`git merge-tree` -> clean tree), so the DIRTY state is gone too.

One blocking item, and it is an unlucky one: **the docstring you added at my request states the opposite of what the code does.** One line.

## Closed, with the evidence

| ask | evidence |
|---|---|
| the docstring's two limits — non-transitive deadness, global name match | present at `:56-77`; see below for the defect in one of them |
| pin the composition independently of the allowlist | **mutation A** below — your new test fires by name |
| "exactly four" asserted membership only | **mutation B** below — the new equality assert fires, naming all 43 |

**Mutation A — the main fix.** Anchor `return drop_dead_function_reads(...)` in `_consumed_in_src`, **exactly one occurrence**; mutated to `return consumed`; module genuinely imports afterwards (`_consumed_in_src()` returns a set of 3483). Two behavioural failures, including the test this round added:

```
AssertionError: warmup_auto is read only inside a dead function, yet
_consumed_in_src() counts it: the gate is not composed into the scan
```

**Mutation B — over-drop.** `all(...)` -> `any(...)` in `drop_dead_function_reads`, **exactly one occurrence**, imports fine. Five failures including `test_the_real_tree_has_exactly_the_four_known_escapes`, which names all 43 escapes rather than reporting a count. That confirms your own honest note that the exactness assertion *adds* to four pre-existing killers rather than rescuing a mutation that was surviving — you said so rather than claiming a save, and the measurement agrees.

Deleted lines vs `main`: six, all paired in-place edits — the split that *is* the fix, a widening of `_declared()` to include `LoraConfig`, and the allowlist pin moving `== 36` -> `== 42` with the `==` retained. No test function or class removed. The 23 removed lines in the delta are `main`'s own retired allowlist entries arriving through the rebase. Fragment is correct and credits you.

40 passed.

---

## Blocking: `tests/test_issue748_config_fields_reach_a_consumer.py:69-70`

> ``@app.command()`` bodies are safe only because their decorator loads the name -- by accident rather than by design.

**Measured on your head, using your own `referenced_names()`: 150 decorated functions, 115 ABSENT from `referenced_names`, 35 present.**

`@app.command()` emits `Name` loads for `app` and `command` and nothing else — the decorated function's name is a `FunctionDef.name`, not a `Name` node, so the decorator never loads it. Absent examples: `_eval_v0550.py:design_cmd`, `_eval_v0650.py:behavior_cmd`. The 35 that *are* present are rescued incidentally by the global spelling match, which is the other mechanism, described in the next paragraph.

Two things make this worth a push rather than a nit:

1. It is **filed under the wrong leak.** It sits in the "errs toward *referenced*" paragraph, asserting command bodies are safe — when command bodies are the largest instance of the *opposite* hole, the one the very next paragraph describes.
2. It is the class you yourself named on 2026-09-14: *"a false claim into the guard's own voice, in the section that exists to tell the next reader what the guard cannot see."* Impact today is zero — the exact-four test pins the real behaviour and passes unmutated — but that section's whole job is to be trusted without re-derivation.

For what it is worth, the 150 / 115 / 35 numbers you gave me were right. It is the sentence explaining the mechanism that inverted them.

## Non-blocking

**Three docstring claims have no test:** non-transitivity, the global-spelling rescue, and the `getattr`-string hole. I verified non-transitivity is **true** by reading `drop_dead_function_reads` — a dead caller's body still emits a `Name` load of the inner name, so the inner function is "referenced" and its reads are kept. The other two I did not verify and am not asserting either way. Prose in that docstring is load-bearing documentation; a case each would be better than my word or yours.

**Your 09-15 comment says the pin is 46; it is 42.** `main`'s pin is now 36 after later merges retired entries, and 36 + 6 = 42, which is what the code says. The code is right and the comment is stale — flagging it only because the number is the kind people quote.

---

## On #988, since it is yours too and the ordering matters

They are sequenced rather than contending, and your #988 body already says so. Concretely, at this head: #906 changes the import at `:402` and the loop at `:416-421`; #988 changes the `assert imported == SRC` at `:410`. On `main` those sit 6 lines apart, at your head **8** — so #988's "three lines below" is stale in the harmless direction. The one plausible rebase conflict is that #906 inserts 22 docstring lines immediately above the `"unkillable by construction"` sentence #988 edits; small.

The substantive interaction is worth knowing before you rebase: **#906 widens `_declared()` to `LoraConfig` and adds a test that calls `_declared()`, which enlarges #988's blast radius** — the false alarm #988 describes gains at least one more failing test. That is an argument for landing #988 promptly after this, not against either.

Push the one-line docstring correction and I will merge on the next green matrix, reviewing only the delta.

Srinivasan8888:
Pushed as `af116d8`: only the docstring changed, no test code.

You were right, and I reproduced your numbers before editing: of the 150 `*.command`-decorated functions under `src/`, 115 are absent from `referenced_names()` and 35 are present, with `plugins.py:list_cmd` among the absent ones. The `@app.command()` sentence now sits under the opposite hole, as its largest instance, and that paragraph no longer calls the hole narrower. `test_issue748_config_fields_reach_a_consumer.py`: 40 passed, ruff clean. The six removed lines are all the old docstring prose.

Also, you were right about the pin: my 09-15 comment said 46, but it is 42, which is what the code says.

I haven't added tests for the three untested docstring claims yet, since you marked them non-blocking. If you'd like them in this PR rather than a follow-up, say so and I'll add one case each.

CI on this push will pick up peft 0.21.0, so expect the #1005 layer-streaming reds until #1010 lands.

MakazhanAlpamys:
**The blocking line is fixed and its numbers are true — I re-derived them rather than checking that the text changed.** One word in the replacement, and then this merges.

Using **your own `referenced_names()`**, over the gate's module set and over all of `src/`, identical either way:

```
decorated functions (*.command): 150
ABSENT from referenced_names:    115
PRESENT in referenced_names:      35
```

Exact match. Absent confirmed on `_eval_v0550.py:design_cmd`, `discover_cmd`, `lock_cmd`, `coverage_cmd`, `plugins.py:list_cmd`; present on `cli.py:version`, `adapters.py:info`, `diff`.

**And the new text is filed under the right hole now** — it moved out of the "errs toward referenced" paragraph, and `"The opposite hole is narrower."` is gone. Every other clause checks out: the decorator emits `Name(app)` + `Attribute(.command)` and never the function's own name; the 35 present are spelling-rescues (I spot-checked two in depth — `dedup` is rescued by a local `if dedup:` at `commands/data.py:1734` while the decorated `dedup` is at `:321`, and `split_data` by a loop variable at `:1196`); and "drops nothing beyond the four above" is pinned by `test_the_real_tree_has_exactly_the_four_known_escapes`.

---

## The one thing: `:73` added a superlative I cannot support

> **Typer command bodies are its largest instance.**

Measured over the gate's own module set — 4664 function definitions, **684 absent** from `referenced_names`:

```
(no decorator)                478
.command                      115      <- 17% of the hole
classmethod + field_validator  41
.post                          16
.get                           15
model_validator                 7
```

`.command` is the largest **decorated** class. The largest instance is the undecorated bucket, four times bigger. (Over all of `src/` including `schema.py`: 817 absent, 443 undecorated, 124 field_validators, 115 `.command`.)

**Why I am raising a single word.** This is the paragraph whose entire job is to tell the next reader what the guard cannot see, and the round we just finished was about a false claim in exactly that paragraph. Replacing one unverified sentence with another — smaller, but the same kind — is the failure repeating at lower amplitude. It is also the cheapest possible fix: *"its largest **decorated** instance"*, or *"the largest single class of it I measured"*.

I would rather say this than wave it through because the big one is fixed.

## Your two open questions

**The three untested docstring claims** — non-transitivity, the global-spelling rescue, the `getattr`-string hole — are still prose only, and you asked whether to add a case each here. **My answer: not in this PR.** The delta is +9/-6 of prose; adding three tests re-opens a review that is otherwise done, and the claims are true (I verified non-transitivity by reading `drop_dead_function_reads` last round). File them as a follow-up if you want them, or leave them — your call, and I will not raise it again either way.

**The pin, 46 vs 42** — you are right and the code was always right; `:629` asserts `== 42` and `36 + 6 = 42` is consistent. Only your comment carried the stale number, which is not worth editing.

**The rest is clean.** Six removed lines, all paired in-place edits: the split that *is* the fix, `_declared()` widened to `LoraConfig`, and the pin `36` -> `42` with `==` retained rather than loosened. No test function or class removed. 40 passed, ruff clean.

**Mutation** — anchor `return drop_dead_function_reads(...)` matched **exactly once**, replaced with `return consumed`, module import confirmed (`_consumed_in_src()` returns a set of 3483 afterwards). **2 failed**, both behavioural:

```
warmup_auto is read only inside a dead function, yet _consumed_in_src()
counts it: the gate is not composed into the scan
```

plus the allowlist test naming `training.lora.init_strategy`, `use_olora`, `use_vera`, `warmup_auto`. Reverted, 40 passed.

Push the one word and I merge — behind #1010, which blocks everything tonight.

Srinivasan8888:
Pushed as `481e755`, one line. You're right: "largest instance" was a second unmeasured claim in the same paragraph. It now reads "its largest decorated class among the modules the gate scans; undecorated functions are a larger share."

I limited it to the gate's module set rather than just adding "decorated" because of your own all-of-`src/` numbers. With `schema.py` included, `field_validator` (124) outnumbers `.command` (115), so "largest decorated" is only true for the set `referenced_names` is computed over.

40 passed, ruff clean.

On the three untested claims: understood, not in this PR.

MakazhanAlpamys:
**Merge-ready.** CI is re-running against the new `main` (#1010 landed as `f8226214`, so the nine layer-streaming failures are gone); I merge when it lands.

Your fix is better than what I asked for, and the difference is worth naming.

I asked for one word — *"its largest **decorated** instance"*. You wrote:

> Typer command bodies are its largest decorated class among the modules the gate scans; undecorated functions are a larger share.

That does not just narrow the claim to something true, it **states the fact that made the original claim false**. Someone reading the docstring in six months now knows the shape of the hole (478 undecorated against 115 `.command`, of 684 absent definitions) rather than just a corrected superlative they have to take on trust. It also scopes itself to "the modules the gate scans", which is the right boundary — the numbers differ once `schema.py` is included, and the sentence no longer over-reaches.

Delta is exactly that, 2 insertions and 1 deletion in one file. Nothing else moved.

For the record of what this PR closes: the docstring's original sentence claimed `@app.command()` bodies were **safe** because the decorator loads the name. Re-derived on your head with your own `referenced_names()`: **150 decorated functions, 115 absent from `referenced_names`, 35 present** — and the 35 are rescued incidentally by the global spelling match, a different mechanism. So command bodies were not the safe case, they were the largest decorated instance of the opposite hole. Two rounds, and both times the correction came from measuring rather than re-reading.

Leaving the three untested docstring claims (non-transitivity, the global-spelling rescue, the `getattr`-string hole) out of this PR, as I said — file them separately if you want them, or leave them; they are true and I verified non-transitivity by reading `drop_dead_function_reads`.

MakazhanAlpamys:
**Correction: "CI is re-running against the new `main`" was wrong, and your run will very likely come back red through no fault of yours.** Flagging it before it happens rather than after.

You pushed `481e7550` at 2026-09-15T19:28:33Z and the run was created in the same second. **#1010 merged at 2026-09-16T05:37:30Z** — ten hours later. GitHub pins the merge commit when the run is *created*, not when the job starts, so your run is building your branch against a `main` that still carries the `peft 0.21.0` defect.

The evidence, from an actual job log on another PR that hit exactly this:

```
```

An explicit SHA in the fetch, and a base 15 commits behind `main` that does not contain `f8226214`. That job started *after* #1010 merged and still checked out the pre-fix tree.

So when your nine `test` cells reach the front of the queue, expect this:

```
FAILED tests/test_v07201.py ... AssertionError: no adapter tensors were written at all
FAILED tests/test_v07202.py ... every lora_B is zero - the adapter was dropped on reload
FAILED tests/test_v07203.py ...
= 9 failed =
```

**That is #1005 and it is not yours.** Your PR does not touch layer streaming.

## What this changes for the merge

Nothing about the verdict — the one-word fix landed and I verified it. Re-reading my own comment: I said "I merge when it lands", and what lands will be a red matrix I cannot merge on, since the rule here is green required checks on the current head taken from `gh pr checks`, not from my confidence about which failures are spurious.

So this needs a **new event** to recompute the merge commit. Either you push (a rebase onto current `main`, or anything), or I press "Update branch" (`gh pr update-branch`) — your commits and authorship untouched, one merge commit added, `--rebase` available if you prefer.

**Say which and I will do it.** I am not touching your branch without your word.

Sorry for the wrong steer. The underlying problem — that this repo cannot re-evaluate an open PR when `main` moves, and that no maintainer-side action does it without writing to a contributor's branch — is filed as **#1017**.

Srinivasan8888:
Option 1 — I pushed the rebase, so no `Update branch` needed here. Head is now `b0a466e`, rebased onto `f8226214` (current `main`), same single commit on top, nothing else changed.

Thanks for flagging it before the run finished rather than after; the pinned-merge-SHA detail is the part I would not have worked out from the red X alone.

`test_issue748_config_fields_reach_a_consumer.py`: 40 passed locally on the rebased tree.

MakazhanAlpamys:
Merged as 27220a16. Thank you, @Srinivasan8888.

You chose the rebase yourself and put it on the right base, and the content I had reviewed came through byte-identical (`git diff 481e7550 b0a466e9` on the test and the fragment is empty). The scoped wording ("among the modules the gate scans") is the only true form of the claim, and you kept it.

What I ran on the tree merged with `main` 76c23541 (Python 3.12, peft 0.20.0): 40 passed, F821 clean. Because this is a ratchet, I checked both that it fires and that it stays quiet on correct code:
- Removing the dead-function gate (`return consumed`) gives 2 failures, naming `init_strategy`, `use_olora`, `use_vera` and `warmup_auto`.
- Appending an unreferenced `def _zz_dead_reader(config): return config.training.lora.loftq_iter` keeps the allowlist test silent, and your exact-four pin catches the new escape.
- Referencing that function makes the `KNOWN_UNCONSUMED` entry obsolete, and the test says so by name.
- On the real tree it gives 40 passed.

PR #907, opened 2026-09-12
fix: consolidate nine private for_terminal copies onto soup_cli.utils.terminal

### LINKED ISSUE #896: Consolidate the six private _for_terminal copies onto soup_cli.utils.terminal.for_terminal

VanshSharmaPES:
Closes #896.

## What changed

Six command modules (`adapters.py`, `data_canary.py`, `data_doctor.py`, `draft.py`, `infer.py`, `shrink.py`), `utils/ship_verdict.py`, and `mcp_server/registry.py` each carried a byte-identical private copy of the C0/DEL control-strip table instead of importing the v0.75.0 shared `soup_cli.utils.terminal.for_terminal` — exactly how the config loader's own fix got missed for a release in the first place.

While auditing every call site I found a **ninth, uncatalogued copy**: `commands/serve.py`'s `--auto-spec` draft-model pairing message had the same table under a local `_ctrl` variable name, so it wouldn't have matched a naive grep for `_CONTROL_STRIP_TABLE`. Fixed it too — leaving it would have defeated the point of the ratchet.

## `strip_control()` vs `for_terminal()` — checked per call site, not swept

`utils/terminal.py` now exports two functions instead of one:

- `for_terminal(text)` — strip, then escape Rich markup (the original v0.75.0 behaviour).
- `strip_control(text)` — strip only, no escaping.

Every one of the ~20 call sites across the nine files was converted individually after checking what it actually needed, not by a mechanical find-replace:

- Sites already wrapped in `escape(_for_terminal(x))` → `for_terminal(x)` (adapters.py, data_canary.py, data_doctor.py's table rows, draft.py's history table, infer.py, serve.py).
- Sites feeding `Text.append(text, style=...)` → `strip_control(x)`, **not** `for_terminal(x)` — `Text.append`'s first argument is never parsed as Rich markup, so escaping it would render literal `\[` in the output. This is `data_doctor.py`'s mask-preview renderer.
- Sites building a `RuntimeError`/exception message that's escaped later by the CLI's own friendly-error handler → `strip_control(x)` to avoid double-escaping. This is `draft.py` and `shrink.py`'s subprocess-error tails.

An unconditional `for_terminal()` sweep would have silently introduced the `Text.append` bug above.

## `mcp_server/registry.py::_sanitize` — kept, not replaced

`_sanitize` stays exactly as it was: a recursive walk over dicts/lists/tuples at the JSON-RPC transport boundary, with its own non-string-leaf pass-through behaviour. Only its string-leaf case changed, from a local `.translate(_CONTROL_STRIP_TABLE)` to calling the shared `strip_control()`. I did not fold it into `for_terminal` or otherwise change its recursion, since that behaviour belongs to the transport boundary, not to this cleanup.

## `ship_verdict.py` — name kept, implementation delegated

`ship_verdict.py`'s public `for_terminal` is actually strip-only (every call site, including in `commands/ship.py`, wraps it in `escape()` itself) despite the name. Renaming it would have been a breaking change to `commands/ship.py`'s import, so it now reads:

```python
from soup_cli.utils.terminal import strip_control as for_terminal
```

## The ratchet — mutation-tested, not just written

Added `tests/test_issue896_no_duplicate_terminal_sanitizer.py`, an AST walk over `git ls-files src/soup_cli` (mirrors `test_no_foreign_license_headers.py`'s approach) that fails if any module besides `utils/terminal.py` defines `_CONTROL_STRIP_TABLE` at module scope.

Verified by hand per your ask: I injected a duplicate table into `adapters.py`, ran the test, confirmed it went red and named `src/soup_cli/commands/adapters.py` as the offender, then reverted the injection before committing anything.

## Regression caught by the full suite

Removing the duplicate table from `adapters.py` shifted two of its lines, which broke `test_issue775_path_containment_ratchet.py`'s pinned-line-number allowlist (`commands/adapters.py:38→29`, `:107→98`). Fixed in a separate commit so it's easy to review in isolation.

## Rebase note

#904 is open and also touches `draft.py`/`shrink.py`, but a different code path (config-validation error reporting vs. this terminal-sanitizer cleanup) — expect only an import-line conflict on rebase once it lands.

## Test plan

- `ruff check src/soup_cli/ scripts/ tests/ benchmarks/` — clean
- `pytest tests/` — 21,027 passed / 0 newly failing (13 pre-existing failures on `main` too, all Windows-symlink-privilege or Apple-Silicon-only tests unrelated to this change — verified against a `cf23a69` worktree)
- Two pre-existing tests (`test_v07127.py`, `test_v07129.py`) that imported the now-deleted private `_for_terminal` updated to import the shared `strip_control` instead

Changelog fragment at `changelog.d/0.75.0/896.fixed.md` — filename will need renaming to this PR's number per the repo's convention; will push that once assigned.


VanshSharmaPES:
@MakazhanAlpamys Pushed a fix for the failing checks (changelog fragment now references #907). Could you approve the pending workflow run so CI can execute? Thanks!

MakazhanAlpamys:
Welcome, and thank you — this is an unusually careful first contribution. I reviewed `c723e583` in a dedicated worktree and re-ran everything rather than reading the description.

**The three things I look at hardest on a sweep like this, all handled correctly:**

- **`Text.append(..., style=...)`.** You were right that a mechanical replacement would print literal `\[` there, because Rich does not parse markup inside that call. `data_doctor.py:90` is the only such site in `src/`, and it correctly got `strip_control`, not `for_terminal`. Spotting that yourself, before anyone raised it, is the reason this review was short.
- **`_sanitize` recursion at the MCP transport boundary** is intact — nested dict/list values still go through the shared table.
- **`ship_verdict.for_terminal` really is strip-only at every call site.** I checked the one that looks unescaped: `commands/ship.py:923` flows into `escape(detail)` two lines later at `:925`. So aliasing it to `strip_control` preserves behaviour exactly, and keeping the name avoids a breaking change for now. Good call.

**Mutations of the main fix — nothing survived:**

| mutation | result |
| --- | --- |
| `for_terminal` and `strip_control` both identity | 15 failed across 6 files |
| C0 strip removed only | 15 failed |
| markup escape removed only | 3 failed |
| re-inject a module-scope `_CONTROL_STRIP_TABLE` into a command module | ratchet red, names the offending file |

Lint clean. No test was deleted or weakened — the only test edits re-point imports at `utils.terminal.strip_control` with byte-identical assertions, and the two `test_issue775` line re-pins are mechanically correct (the deleted block was 9 lines) and protected by `test_the_allowlist_has_no_dead_entries`. Fragment is in `changelog.d/0.75.0/` with the right shape. Leftover scan: `_for_terminal` / `_CONTROL_STRIP_TABLE` now appear exactly once in `src/`, in `utils/terminal.py`.

**One change I would like before merge, and it is your own insight turned back on the guard.**

The ratchet bans a *name* at *module scope*: `BANNED_NAMES = frozenset({"_CONTROL_STRIP_TABLE"})`, matched by `_module_level_assigned_names`. But the ninth copy — the one you found by hand in `serve.py` — was a **function-local `_ctrl`**, neither module-scope nor named that. I injected exactly that shape into `commands/recipes.py`:

```python
def _probe_local_copy(text: str) -> str:
    _ctrl = {i: None for i in range(0x20) if i not in (0x09, 0x0A, 0x0D)}
    return str(text).translate(_ctrl)
```

```
pytest tests/test_issue896_no_duplicate_terminal_sanitizer.py -q --no-cov  ->  4 passed
```

Green. So the guard currently locks the door you walked through and leaves open the one you climbed in by. A module-scope copy renamed `_CTRL_TABLE` also passes.

This satisfies #896's literal wording, which is why I am asking rather than blocking. The fix is cheap and provably free of false positives: flag the table's **shape** as well as its name — an AST dict or dict-comp whose iterator is `range(0x20)`, at any scope, under any name, outside the allowed definer. I measured the false-positive cost:

```
grep -rn "range(0x20)" src/soup_cli --include=*.py
src/soup_cli/utils/terminal.py:25   <- one hit, yours
```

Please add that plus a test that proves it — a function-local copy under a different name should go red. A guard that cannot catch the instance that motivated it is the one thing worth fixing here.

**Minor, take or leave:**

- `len(scanned) > 50` in the vacuity guard is loose against ~425 modules; a regression dropping 85% of the tree would still pass. `> 300` would actually guard.
- The two-line `_CONTROL_SEED` comment in `data_canary.py` ("controls are a null distribution, not a secret…") was deleted as collateral with the sanitizer block. Unrelated to this change — worth restoring.
- Stray blank lines left by the removals: `infer.py:315`, `infer.py:548-549`, `shrink.py:56-58`. Cosmetic only; `ruff format` is off repo-wide and nothing gates it.
- Two stale prose references to the now-deleted private helper in `test_v07302.py:918` and `test_issue627_unknown_config_keys.py:1052`.

**Merge-order heads-up, not your fault.** #763 adds **five new** `escape(_for_terminal(...))` call sites in `commands/adapters.py` — I confirmed the file and the count from its diff. Whichever of the two lands second needs a rebase rewriting them to `for_terminal(...)`; if #763 merged unrebased after this, `adapters.py` would raise `NameError` at those sites. I will carry that to #763 so it is not on you. Your own read on #904 was right — its diff contains no `_for_terminal` references, so import-line conflict at worst.

Push the shape check and I will merge.

VanshSharmaPES:
Pushed the shape check.

`test_issue896_no_duplicate_terminal_sanitizer.py` now walks the whole AST (not just module scope) for a dict-comp whose generator iterates `range(0x20)` -- the table's shape -- under any name, at any scope. `grep -rn "range(0x20)" src/soup_cli` still comes back with exactly one hit (`utils/terminal.py`), so it's a clean signal.

Reproduced your mutation exactly before writing the fix: appended your `_probe_local_copy` snippet to the real `commands/recipes.py`, ran the suite, watched it go red naming that file, then reverted. Added `test_it_flags_a_function_local_copy_under_a_different_name` (asserts the name-based check misses it and the shape-based one catches it, at the exact line) so this gap can't come back silently.

Also did the four minor items:
- Restored `data_canary.py`'s `_CONTROL_SEED` comment.
- Vacuity guard is now `> 300` instead of `> 50`.
- Cleaned the stray blank lines in `infer.py` / `shrink.py`.
- Updated the two stale `_for_terminal` prose references in `test_v07302.py` and `test_issue627_unknown_config_keys.py`.

Rebased onto current `main` (through #893's Langfuse fix and the RTX 5070 benchmark commit) -- clean, no conflicts. `ruff check` clean, 1239 targeted tests passing locally.

One thing I can't do from here: no CI checks have run on this PR at all (`gh pr checks` reports none), which I believe is GitHub gating Actions on a first-time fork contributor pending maintainer approval. Could you approve the workflow run so CI actually executes? Everything above is from my own local `pytest`/`ruff` runs, not CI's.

MakazhanAlpamys:
Verified at `97563a74`. The shape check does what it needs to, and I want to give you the measurements plus one correction of my own work.

**The mutation that survived before is dead now.** Re-injecting the exact `serve.py` shape — a function-local dict-comp under a different name — into `commands/recipes.py`:

```
pytest tests/test_issue896_no_duplicate_terminal_sanitizer.py -q --no-cov
FAILED ...::test_no_module_besides_terminal_defines_the_strip_table
1 failed, 5 passed
```

Clean tree: `6 passed`. You also restored the `_CONTROL_SEED` comment in `data_canary.py` and raised the vacuity floor to `> 300` against 504 tracked files, which is a real guard rather than a number that happens to be true.

I drove the scanner directly over six spellings:

| spelling | flagged |
| --- | --- |
| module-scope `_CONTROL_STRIP_TABLE` | yes (name) |
| module-scope dict-comp, renamed | yes (shape) |
| function-local dict-comp, renamed | yes (shape) |
| `range(32)` instead of `range(0x20)` | yes — `_is_range_0x20_call` compares the VALUE, so both spellings hit |
| `dict.fromkeys(range(0x20))` | no |
| `str.maketrans('', '', ...)` | no |

**A correction, because I got this wrong first and it is worth saying so.** My first pass reported that even a renamed module-scope copy slipped through, and I nearly sent you that as a second finding. It was my method, not your code: I injected the probe line between a decorator and its `def`, which is a `SyntaxError`, and `_offenders()` has `except SyntaxError: continue`, so the file was silently skipped and the run came back green. Injecting at a valid location turns it red immediately. Your guard was right and my probe was broken.

That does surface one genuine, small thing: **`except SyntaxError: continue` means a file the scanner cannot parse is treated as clean.** Today every tracked file parses, so the practical risk is nil, and I am not asking you to change it in this PR — but a file under `src/soup_cli/` that does not parse is a worse problem than a duplicate table, so it deserves to be loud rather than skipped. If you want it, `pytest.fail` naming the file would be a one-line follow-up; if you would rather stop here, that is completely fine and I will file it.

The two unflagged spellings above are the same shape of gap, and I am explicitly **not** asking for them either. `dict.fromkeys(range(0x20))` is the plausible one; `maketrans` is a different construct. A ratchet that catches the name, the renamed module-scope copy and the function-local copy already covers every form that has actually appeared in this repo, and chasing every theoretical spelling is how guards become unmaintainable. Noting them so the limit is written down somewhere.

Nothing else outstanding. CI on this head needed manual approval — I have approved it. Once it is green I will merge.

For the record on ordering, so you are not surprised: **#763 adds five new `escape(_for_terminal(...))` call sites in `commands/adapters.py`**, the file you are removing that name from. I have already flagged it on that PR, so the rebase is on them, not on you.

VanshSharmaPES:
CI is green — all 14 checks passed on 97563a7: `lint`, `type-check`, `pytorch-smoke`, `mlx-smoke`, `transformers-floor`, and `test` across ubuntu/macos/windows × 3.10/3.11/3.12.

Summary of the PR for anyone landing on this thread:

**The problem (#896):** `soup_cli.utils.terminal.for_terminal` (strip C0/DEL control bytes, then escape Rich markup) shipped in v0.75.0 for the config loader's unknown-key report, but six command modules (`adapters.py`, `data_canary.py`, `data_doctor.py`, `draft.py`, `infer.py`, `shrink.py`) plus `utils/ship_verdict.py` and `mcp_server/registry.py` still carried their own byte-identical private copy of the same control-strip table — which is exactly how the loader's own fix got missed for a release in the first place.

**What changed:**
- All eight cataloged copies, plus a ninth found by hand in `commands/serve.py`'s `--auto-spec` pairing message (a function-local `_ctrl` var, not a module-scope table — easy to miss with a naive grep), now import the shared helper.
- `utils/terminal.py` gained a `strip_control()` (strip-only) alongside `for_terminal()` (strip + escape), since several call sites feed non-markup contexts — `Text.append(..., style=...)`, or an exception message escaped later downstream — where the full escape would have corrupted output. Each of the ~20 call sites was converted individually after checking which half it needed, not swept mechanically.
- `mcp_server/registry.py::_sanitize` kept its own recursive dict/list/tuple walk at the JSON-RPC transport boundary; only its string-leaf case now calls the shared primitive.
- `ship_verdict.py`'s public `for_terminal` (actually strip-only at every call site, despite the name) now aliases the shared `strip_control` rather than being renamed, avoiding a breaking change for `commands/ship.py`'s import.
- Added `tests/test_issue896_no_duplicate_terminal_sanitizer.py`, a repo-wide AST ratchet. It started out name-based (banning `_CONTROL_STRIP_TABLE` at module scope) but was hardened during review to match the table's *shape* — a dict-comp iterating `range(0x20)`, at any scope, under any name — after code review demonstrated the name-based version missed the exact ninth-copy shape. Verified twice by hand: injecting a module-scope copy into `adapters.py`, and separately injecting a function-local copy under a different name into `recipes.py`, both turned the ratchet red naming the offending file, then reverted.

**Review fixes along the way:** restored a comment deleted as collateral, tightened the scan's vacuity guard from `>50` to `>300` tracked files, cleaned stray blank lines, and updated two stale prose references to the deleted private helper.

Thanks for the thorough review at every step — the `Text.append` shape distinction, the strip-only `ship_verdict` alias, and especially the ratchet's shape gap were all things I'd have shipped without, and the mutation-testing back-and-forth is exactly what made the ratchet worth the name.

MakazhanAlpamys:
Merged as `0ec39268`. Thank you — and the reason this is a better PR than the issue asked for is the **ninth copy**.

#896 named six. You found `commands/serve.py`'s `--auto-spec` pairing message carrying the same table under a local `_ctrl` name, which a grep for `_CONTROL_STRIP_TABLE` would never have located. That single find is the whole argument for the guard walking the AST instead of matching an identifier — the name is not what makes it a duplicate — and it is why the ratchet is worth more than the sweep it came with.

It finds zero offenders today, which makes the can-it-fail test the load-bearing half of the file, and you added it after review rather than arguing that the ratchet was self-evidently correct. That is the part I have to ask for most often.

Two limits are recorded rather than chased, and I want the reasoning on record because it is the right call and it will look like an omission to someone reading later: `dict.fromkeys(range(0x20))` and `str.maketrans` are not matched, and neither has ever appeared in this repo. A guard that chases every theoretical spelling is how guards become unmaintainable — and this repo has already deleted one scanner that fired on 31 correct files.

The parse-failure case is the one I do think is worth a line: a file under `src/soup_cli/` that does not parse is currently skipped rather than reported, and an unparseable source file is a worse problem than a duplicate table. You said you would rather stop here, which is fine — **I have filed it rather than leaving it in a merged PR's comment thread**, so it is not lost.

CI was green on all fourteen checks on `97563a74`. Your CONTRIBUTORS.md entry is in and #896 is closed.

One ordering note that is on someone else, not you: **#763 adds five new `escape(_for_terminal(...))` call sites in `commands/adapters.py`**, the file you are removing that name from, so it now needs a rebase. I flagged it there before merging this, so nobody is surprised.

Issue #804, opened 2026-09-11
data.chat_template is applied by SFT and distill only: DPO renders conversational rows with the model's shipped template instead


MakazhanAlpamys:
## The defect

`data.chat_template` ("Override the tokenizer chat template", `config/schema.py:398`) is installed on the tokenizer by `apply_chat_template_override`, which has one caller: `build_format_row` (`src/soup_cli/data/sft_format.py:52`). `build_format_row` has two callers, `trainer/sft.py:729` and `trainer/distill.py:469`. No other trainer applies the override, and no validator restricts the field by task.

The preference trainers hand message-list rows to TRL, which renders them with `tokenizer.chat_template`, the shipped one. GRPO and Online DPO go further: when the tokenizer has no template they install a hardcoded fallback (`trainer/grpo.py:360-363`, `trainer/online_dpo.py:353-354`), never the user's.

## Reproduction

Tiny random-init Llama on CPU carrying SmolLM2's tokenizer (which ships a ChatML template with a default system prompt). `data.chat_template` set to a custom Jinja string that emits a `[[SOUP-OVERRIDE <role>]]` marker. Same override, same prompt, two tasks:

```

config accepted: task=dpo data.chat_template set = True

columns: ['prompt', 'chosen', 'rejected', 'prompt_ids', 'chosen_ids', 'rejected_ids']

DPO rendered prompt: '<|im_start|>system\nYou are a helpful AI assistant named SmolLM, trained by Hugging Face<|im_end|>\n<|im_start|>user\nWhat is 2+2?<|im_end|>\n<|im_start|>assistant\n4<|im_end|>\n'

tokenizer template contains override marker: False

SFT rendered row: '[[SOUP-OVERRIDE user]] What is 2+2? [[SOUP-OVERRIDE assistant]] 4 '

```

(DPO rows: `{"prompt": [{"role": "user", ...}], "chosen": [{"role": "assistant", ...}], "rejected": [...]}`, the conversational shape `_convert_dpo` passes through, `data/formats.py:245-259`.) The DPO run trains on the shipped template, including a system prompt the user never wrote.

Acceptance by task, `load_config_from_string` with `data.chat_template: chatml`:

```

ACCEPT dpo  ACCEPT kto  ACCEPT orpo  ACCEPT simpo  ACCEPT ipo  ACCEPT bco

ACCEPT reward_model  ACCEPT grpo  ACCEPT online_dpo  ACCEPT pretrain

[...]

PR #908, opened 2026-09-12
fix(train): avoid fake loss deltas for short runs

### LINKED ISSUE #899: Training-complete panel always shows "Loss: 0.0000 -> 0.0000" when total steps < logging_steps
## Summary

The "Training Complete!" rich panel printed at the end of `soup train` always shows `Loss: 0.0000 -> 0.0000` when the run has fewer total steps than `training.logging_steps` (default `10`) — even though the run had a real, non-zero loss.

## Repro

```bash
soup quickstart --dry-run -y -o /tmp/soup-demo
soup train --config /tmp/soup-demo/quickstart_soup.yaml --yes
```

This trains TinyLlama-1.1B-Chat-v1.0 with LoRA SFT on 18 samples for 1 epoch → 5 total steps (`5 < logging_steps=10`).

**Actual trainer output (correct):**
```
{'train_runtime': '4183', 'train_samples_per_second': '0.004', 'train_steps_per_second': '0.001', 'train_loss': '2.383', 'entropy': '1.117', 'num_tokens': '693', 'mean_token_accuracy': '0.6502', 'epoch': '1'}
```

**"Training Complete!" panel (wrong):**
```
╭───────────────────────────── Training Complete! ─────────────────────────────╮
│ Loss: 0.0000 -> 0.0000                                                       │
│ Duration: 1h 9m                                                              │
...
```

## Root cause

- The panel is rendered at `src/soup_cli/commands/train.py:1599`:
  ```python
  f"Loss: [bold]{result['initial_loss']:.4f} -> {result['final_loss']:.4f}[/]\n"
  ```
- `initial_loss`/`final_loss` are computed identically across essentially every trainer backend (confirmed in `sft.py:2003-2012`; the same pattern also appears in `dpo.py`, `orpo.py`, `grpo.py`, `kto.py`, `bco.py`, `ipo.py`, `simpo.py`, `online_dpo.py`, `reward_model.py`, `classifier.py`, `embedding.py`, `asr.py`, `mole_routing.py`, `pretrain.py`, `distill.py`):
  ```python
  logs = self.trainer.state.log_history
  train_losses = [entry["loss"] for entry in logs if "loss" in entry]
  ...
  "initial_loss": train_losses[0] if train_losses else 0,
  "final_loss": train_losses[-1] if train_losses else 0,
  ```
- HF `Trainer` only appends an intermediate `log_history` entry keyed `"loss"` every `logging_steps` (default `10`, see `config/schema.py:1235`). When the run has fewer total steps than that, no such entry is ever created — the only `log_history` entry is the final summary, which uses the key `"train_loss"` instead of `"loss"`.
- Because the filter is `if "loss" in entry` (a literal key match, not a substring/prefix check it happens to also match), it never matches that summary entry, so `train_losses` ends up empty and both fields silently fall back to `0`.

## Impact

Cosmetic only — the actual checkpoint/adapter and the logged `train_loss` metric are correct; only the CLI's own completion-panel summary is wrong. But it's a bad first impression for exactly the workflow (`soup quickstart`) meant to be the friendliest onboarding path, and it's a systemic copy-pasted pattern across every trainer backend, not a one-off.

## Suggested fix direction

In each trainer's summary-building code, fall back to `entry.get("train_loss")` (or check for either key) when no `"loss"`-keyed entries exist, so short runs (`total_steps < logging_steps`) still report a real value.

## Environment

- Soup v0.74.0 (fresh `pip install -e ".[dev]"`)
- Python 3.12.14, CPU-only training
- torch 2.14.0, transformers 5.17.0, trl 0.29.1, peft 0.20.0

taljeon:
## What does this PR do?

Fixes the training-complete panel for runs that do not have two per-step loss measurements.

- uses one shared loss-history summary across all 18 migrated trainer modules, including PPO's legacy manual path
- shows a before/after arrow only when at least two per-step `loss` values exist
- shows one value without an arrow when there is one per-step value or only the final run-level `train_loss` mean
- shows `unavailable` instead of inventing a zero delta when no loss was recorded
- preserves non-finite per-step values instead of replacing divergence with the last healthy loss
- preserves numeric `initial_loss` / `final_loss` fields for experiment tracking and sweep ranking

The current source had one more exact copied extractor than the original issue inventory, so that sibling is included rather than left to drift.

Closes #899.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Review follow-up

- Split numeric parsing from finite-only mean fallback. Per-step `NaN`/`inf` values now remain the final evidence, while a non-finite `train_loss` mean still falls through to `unavailable`.
- Added regressions for all-NaN, NaN-tail, and inf-tail histories. A NaN tail no longer becomes a finite sweep candidate.
- Replaced the spelling-specific no-private-copy guard with an AST check over the 18 migrated trainer modules. It accepts `**loss_summary` or `loss_summary["initial_loss"]`, catches alternate quote/comprehension spellings and explicit overrides after a shared spread, and exposed the legacy PPO manual path, which is now migrated too.
- Hardened the completion formatter for legacy producers without `loss_summary_kind`: equal initial/final values render once instead of claiming a measured delta. This also removes the MLX panel's equal-value arrow without widening the trainer migration.
- `soup runs`' historical equal-value rendering remains outside #899's training-complete-panel scope.

## Verification

- TDD: the NaN tail, inf tail, all-NaN, legacy equal-value, and AST guard cases failed before the implementation and pass after it.
- `tests/test_issue899_loss_summary.py`: 12 passed.
- Focused issue + changelog validation: 29 passed.
- Issue, PPO, sweep, and fake-MLX bridge tests: 108 passed.
- Broader dependency-light loss-result surface: 518 passed, 5 skipped, 3 deselected.
- Additional embedding, pretrain, and PRM result-path tests: 23 passed.
- A broader 24-file run reached 812 passed, 7 skipped, and 3 deselected; its 53 failures were all imports from intentionally absent optional training packages (`torch`, `datasets`, `trl`, or `safetensors`). The complete local collection likewise stops only on absent optional dependencies. I did not install the GPU/training stack.
- Post-merge overlap verification: all 12 #899 tests, 16 dependency-light #804 checks (2 skipped), and 324 related BCO/DPO/GRPO/IPO/KTO/ORPO/PPO/SimPO tests passed.
- The eight #804 runtime-order seams also passed with import-only stand-ins for the absent `datasets`, `trl`, and `transformers` packages. This verifies that both chat-template application and loss-summary wiring survive the merge; it is not a real training-stack run, which refreshed CI will cover.
- Repository-wide `ruff check src/soup_cli/ scripts/ tests/ benchmarks/`: passed.
- All migrated trainer modules compile; `compileall`, `git diff --check`, noreply commit metadata, no-coauthor, and diff/body publication checks passed.
- Exact pre-merge head `f4f07b1f` completed all 14 GitHub CI jobs green.
- Merged current `main@40a97fa7` without force in noreply `5f05cd7`. All nine conflicts were adjacent imports in trainer modules; each resolution keeps both upstream `apply_chat_template_override` and this PR's `summarize_training_loss`. The PR remains exactly 22 files. Refreshed current-head CI will be recorded rather than predicted here.

## Checklist

- [x] `ruff check src/soup_cli/ scripts/ tests/ benchmarks/` passes
- [x] Full repository CI passed 14/14 on exact pre-merge head `f4f07b1f`; current merge-head CI is pending
- [x] Updated relevant docs, or no docs change is needed
- [x] Added regression tests for the new behavior
- [x] No breaking change intended

## AI assistance

- Provider: OpenAI
- Model: `gpt-5.6-sol` (Codex)
- Use: current-source inventory, shared-helper design, review repair, conflict resolution, tests, and publication preflight

### DIFF
--- a/changelog.d/0.75.0/899.fixed.md
+++ b/changelog.d/0.75.0/899.fixed.md
@@ -1,0 +1,5 @@
+- **Short training runs no longer show a fabricated `0.0000 -> 0.0000` loss delta (#899).**
+  All Hugging Face-style trainer wrappers now share one loss-history summary: two or more
+  per-step measurements produce a delta, one measurement or a run-level `train_loss` mean
+  produces one value without an arrow, and missing data is shown as unavailable. The
+  completion panel therefore never claims a before/after comparison it did not measure.
--- a/changelog.d/0.75.0/908.fixed.md
+++ b/changelog.d/0.75.0/908.fixed.md
@@ -1,4 +1,4 @@
-- **Short training runs no longer show a fabricated `0.0000 -> 0.0000` loss delta (#899).**
+- **Short training runs no longer show a fabricated `0.0000 -> 0.0000` loss delta (#899 in #908).**
   All Hugging Face-style trainer wrappers now share one loss-history summary: two or more
   per-step measurements produce a delta, one measurement or a run-level `train_loss` mean
   produces one value without an arrow, and missing data is shown as unavailable. The
--- a/src/soup_cli/commands/train.py
+++ b/src/soup_cli/commands/train.py
@@ -34,6 +34,16 @@ _HW_FIT_OPTIMIZERS = frozenset({
 })
 
 
+def _format_training_complete_loss(result: dict) -> str:
+    """Render only a loss comparison that the trainer actually measured."""
+    summary_kind = result.get("loss_summary_kind", "delta")
+    if summary_kind == "unavailable":
+        return "Loss: [bold]unavailable[/]"
+    if summary_kind in {"mean", "single"}:
+        return f"Loss: [bold]{result['final_loss']:.4f}[/]"
+    return f"Loss: [bold]{result['initial_loss']:.4f} -> {result['final_loss']:.4f}[/]"
+
+
 def _build_hardware_fit_input(cfg):
     """Best-effort ``HardwareFitInput`` from a ``SoupConfig``.
 
@@ -1596,7 +1606,7 @@ def train(
     # Report
     console.print(
         Panel(
-            f"Loss: [bold]{result['initial_loss']:.4f} -> {result['final_loss']:.4f}[/]\n"
+            f"{_format_training_complete_loss(result)}\n"
             f"Duration: [bold]{result['duration']}[/]\n"
             f"Output: [bold]{result['output_dir']}[/]\n"
             f"Run ID: [bold]{run_id}[/]\n\n"
--- a/src/soup_cli/trainer/asr.py
+++ b/src/soup_cli/trainer/asr.py
@@ -30,6 +30,7 @@ from typing import Any
 from rich.console import Console
 
 from soup_cli.config.schema import SoupConfig, TrainingConfig
+from soup_cli.trainer.loss_summary import summarize_training_loss
 from soup_cli.utils.gpu import bf16_fp16_flags
 from soup_cli.utils.mixed_precision import align_trainable_dtype_for_fp16
 from soup_cli.utils.seeding import apply_training_seed, training_seed_kwargs
@@ -445,14 +446,13 @@ class AsrTrainerWrapper:
             )
 
         logs = self.trainer.state.log_history
-        train_losses = [entry["loss"] for entry in logs if "loss" in entry]
+        loss_summary = summarize_training_loss(logs)
 
         hours = int(duration // 3600)
         minutes = int((duration % 3600) // 60)
         duration_str = f"{hours}h {minutes}m" if hours > 0 else f"{minutes}m"
         return {
-            "initial_loss": train_losses[0] if train_losses else 0,
-            "final_loss": train_losses[-1] if train_losses else 0,
+            **loss_summary,
             "duration": duration_str,
             "duration_secs": duration,
             "output_dir": self._output_dir,
--- a/src/soup_cli/trainer/bco.py
+++ b/src/soup_cli/trainer/bco.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING, Optional
 from rich.console import Console
 
 from soup_cli.config.schema import SoupConfig
+from soup_cli.trainer.loss_summary import summarize_training_loss
 from soup_cli.utils.gpu import (
     bf16_fp16_flags,
     estimate_batch_size,
@@ -391,15 +392,14 @@ class BCOTrainerWrapper:
         self.tokenizer.save_pretrained(self._output_dir)
 
         logs = self.trainer.state.log_history
-        train_losses = [entry["loss"] for entry in logs if "loss" in entry]
+        loss_summary = summarize_training_loss(logs)
 
         hours = int(duration // 3600)
         minutes = int((duration % 3600) // 60)
         duration_str = f"{hours}h {minutes}m" if hours > 0 else f"{minutes}m"
 
         return {
-            "initial_loss": train_losses[0] if train_losses else 0,
-            "final_loss": train_losses[-1] if train_losses else 0,
+            **loss_summary,
             "duration": duration_str,
             "duration_secs": duration,
             "output_dir": self._output_dir,
--- a/src/soup_cli/trainer/classifier.py
+++ b/src/soup_cli/trainer/classifier.py
@@ -32,6 +32,7 @@ from typing import Any, List, Union
 from rich.console import Console
 
 from soup_cli.config.schema import SoupConfig
+from soup_cli.trainer.loss_summary import summarize_training_loss
 from soup_cli.utils.gpu import bf16_fp16_flags
 from soup_cli.utils.mixed_precision import align_trainable_dtype_for_fp16
 from soup_cli.utils.seeding import apply_training_seed, training_seed_kwargs
@@ -419,14 +420,13 @@ class ClassifierTrainerWrapper:
         self.tokenizer.save_pretrained(self._output_dir)
 
         logs = self.trainer.state.log_history
-        train_losses = [entry["loss"] for entry in logs if "loss" in entry]
+        loss_summary = summarize_training_loss(logs)
 
         hours = int(duration // 3600)
         minutes = int((duration % 3600) // 60)
         duration_str = f"{hours}h {minutes}m" if hours > 0 else f"{minutes}m"
         return {
-            "initial_loss": train_losses[0] if train_losses else 0,
-            "final_loss": train_losses[-1] if train_losses else 0,
+            **loss_summary,
             "duration": duration_str,
             "duration_secs": duration,
             "output_dir": self._output_dir,
--- a/src/soup_cli/trainer/distill.py
+++ b/src/soup_cli/trainer/distill.py
@@ -30,6 +30,7 @@ from typing import TYPE_CHECKING, Any
 from rich.console import Console
 
 from soup_cli.config.schema import SoupConfig
+from soup_cli.trainer.loss_summary import summarize_training_loss
 from soup_cli.utils.gpu import bf16_fp16_flags, resolve_device_map
 from soup_cli.utils.seeding import apply_training_seed, training_seed_kwargs
 
@@ -779,14 +780,13 @@ class DistillTrainerWrapper:
         self.tokenizer.save_pretrained(self._output_dir)
 
         logs = self.trainer.state.log_history
-        train_losses = [entry["loss"] for entry in logs if "loss" in entry]
+        loss_summary = summarize_training_loss(logs)
 
         hours = int(duration // 3600)
         minutes = int((duration % 3600) // 60)
         duration_str = f"{hours}h {minutes}m" if hours > 0 else f"{minutes}m"
         return {
-            "initial_loss": train_losses[0] if train_losses else 0,
-            "final_loss": train_losses[-1] if train_losses else 0,
+            **loss_summary,
             "duration": duration_str,
             "duration_secs": duration,
             "output_dir": self._output_dir,
--- a/src/soup_cli/trainer/dpo.py
+++ b/src/soup_cli/trainer/dpo.py
@@ -7,6 +7,7 @@ from typing import Optional
 from rich.console import Console
 
 from soup_cli.config.schema import SoupConfig
+from soup_cli.trainer.loss_summary import summarize_training_loss
 from soup_cli.trainer.stream_setup import StreamingSetupMixin
 from soup_cli.utils.gpu import (
     bf16_fp16_flags,
@@ -425,15 +426,14 @@ class DPOTrainerWrapper(StreamingSetupMixin):
 
         # Extract metrics
         logs = self.trainer.state.log_history
-        train_losses = [entry["loss"] for entry in logs if "loss" in entry]
+        loss_summary = summarize_training_loss(logs)
 
         hours = int(duration // 3600)
         minutes = int((duration % 3600) // 60)
         duration_str = f"{hours}h {minutes}m" if hours > 0 else f"{minutes}m"
 
         return {
-            "initial_loss": train_losses[0] if train_losses else 0,
-            "final_loss": train_losses[-1] if train_losses else 0,
+            **loss_summary,
             "duration": duration_str,
             "duration_secs": duration,
             "output_dir": self._output_dir,
--- a/src/soup_cli/trainer/embedding.py
+++ b/src/soup_cli/trainer/embedding.py
@@ -8,6 +8,7 @@ from typing import Optional
 from rich.console import Console
 
 from soup_cli.config.schema import SoupConfig, TrainingConfig
+from soup_cli.trainer.loss_summary import summarize_training_loss
 from soup_cli.utils.gpu import (
     bf16_fp16_flags,
     estimate_batch_size,
@@ -374,15 +375,14 @@ class EmbeddingTrainerWrapper:
         self.tokenizer.save_pretrained(self._output_dir)
 
         logs = self.trainer.state.log_history
-        train_losses = [entry["loss"] for entry in logs if "loss" in entry]
+        loss_summary = summarize_training_loss(logs)
 
         hours = int(duration // 3600)
         minutes = int((duration % 3600) // 60)
         duration_str = f"{hours}h {minutes}m" if hours > 0 else f"{minutes}m"
 
         return {
-            "initial_loss": train_losses[0] if train_losses else 0,
-            "final_loss": train_losses[-1] if train_losses else 0,
+            **loss_summary,
             "duration": duration_str,
             "duration_secs": duration,
             "output_dir": self._output_dir,
--- a/src/soup_cli/trainer/grpo.py
+++ b/src/soup_cli/trainer/grpo.py
@@ -11,6 +11,7 @@ from typing import Any, Optional
 from rich.console import Console
 
 from soup_cli.config.schema import SoupConfig, TrainingConfig
+from soup_cli.trainer.loss_summary import summarize_training_loss
 from soup_cli.utils.gpu import (
     bf16_fp16_flags,
     estimate_batch_size,
@@ -721,15 +722,14 @@ class GRPOTrainerWrapper:
 
         # Extract metrics
         logs = self.trainer.state.log_history
-        train_losses = [entry["loss"] for entry in logs if "loss" in entry]
+        loss_summary = summarize_training_loss(logs)
 
         hours = int(duration // 3600)
         minutes = int((duration % 3600) // 60)
         duration_str = f"{hours}h {minutes}m" if hours > 0 else f"{minutes}m"
 
         return {
-            "initial_loss": train_losses[0] if train_losses else 0,
-            "final_loss": train_losses[-1] if train_losses else 0,
+            **loss_summary,
             "duration": duration_str,
             "duration_secs": duration,
             "output_dir": self._output_dir,
--- a/src/soup_cli/trainer/ipo.py
+++ b/src/soup_cli/trainer/ipo.py
@@ -8,6 +8,7 @@ from typing import Optional
 from rich.console import Console
 
 from soup_cli.config.schema import SoupConfig
+from soup_cli.trainer.loss_summary import summarize_training_loss
 from soup_cli.utils.gpu import (
     bf16_fp16_flags,
     estimate_batch_size,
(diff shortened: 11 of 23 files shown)


MakazhanAlpamys:
Reviewed at `e9fc37d3`. This is a clean, genuinely systemic fix and the central design call is the right one: **suppressing the arrow rather than printing `2.383 -> 2.383`**. The issue only asked for the `train_loss` fallback; declining to claim a before/after comparison that was never measured is the more honest answer, and it is what makes this worth 18 files.

I verified the scope claim per file rather than trusting the stat: every trainer hunk is confined to one import line and the `logs = ...` / `return {...}` block at the end of `train()`. No hunk touches setup, optimizer, dataset or loss computation. 18 modules, a superset of the 16 the issue names. `790 passed, 6 skipped` across the 22 existing files referencing `initial_loss`/`final_loss`; `515 passed` on the mole/prm contract files plus the repo guards; ruff clean; 14/14 CI green. No test deleted or weakened — the test diff is one new file, `+64/-0`.

**Mutations of the core, all killed:** hardcoding the panel to always report a delta fails 4; always suppressing fails 2; removing the `train_loss` mean fallback fails the #899 repro test; classifying one per-step point as a delta fails 1; re-introducing a private extractor fails the repo-wide guard.

**Blocking: `_finite_loss` drops NaN and inf, so a diverged run now reports as a healthy one — and not only in the panel.**

I ran the shipped function directly:

```
finite then NaN tail  [2.1, nan, nan, train_loss=nan] -> {'initial_loss': 2.1, 'final_loss': 2.1, 'kind': 'single'}
finite then inf       [2.1, inf]                      -> {'initial_loss': 2.1, 'final_loss': 2.1, 'kind': 'single'}
all NaN                                               -> {'initial_loss': 0.0, 'final_loss': 0.0, 'kind': 'unavailable'}
healthy               [2.1, 1.0]                      -> {'initial_loss': 2.1, 'final_loss': 1.0, 'kind': 'delta'}
```

On `main` the first case renders `Loss: 2.1000 -> nan`. Here it renders `Loss: 2.1000` — the last value before divergence, presented as the result.

This is **not panel-only**, which is what makes it blocking rather than a nit. `commands/sweep.py:165` reads:

```python
if final_loss and final_loss < best_loss:
    best_loss = final_loss
```

`nan < best_loss` is `False`, so on `main` a diverged arm can never win. With this change that arm reports `2.1` and **is selected as best**. The same two columns reach `tracker.finish_run`, `soup runs`, `soup runs compare`, `soup why` and the TUI. It also contradicts the PR body's own "preserves numeric `initial_loss`/`final_loss` for experiment tracking and sweep ranking".

The direction of the error is the one that matters: the summary reads **healthier** than the evidence supports. No test covers NaN or inf.

**Remedy (small):** keep non-finite values in the per-step list and filter only genuinely non-numeric entries, so divergence still renders. Concretely, split the helper — a `_numeric_loss` (rejecting `bool` / `TypeError` / `ValueError` only) feeding `per_step`, and keep the finite check on the `train_loss` mean fallback, where falling through is right. Three tests: all-NaN, NaN tail, inf tail. A fourth `loss_summary_kind == "diverged"` rendered explicitly would also be fine — but the current behaviour must not ship.

**Second item, please also fix: the no-private-copy guard matches one literal string.** I reverted `dpo.py` to the old buggy semantics spelled differently:

```python
([e['loss'] for e in logs if 'loss' in e] or [0])[0]
```

and **305 tests across six files stayed green**. With #919 and #923 both rewriting these same 18 files, a re-introduction during conflict resolution is a live risk rather than a hypothetical. An AST check is a few lines and spelling-independent: every dict literal containing an `"initial_loss"` key must source it from `loss_summary` (either `**loss_summary` or a subscript). Keep the literal grep as the cheap first line.

**Non-blocking:**

- **`backend: mlx` still shows the exact #899 symptom.** `mlx_sft.py:761` is `losses = captured.get("losses") or [0.0]`, reported on the same `logging_steps` cadence, and `_format_training_complete_loss` defaulting an absent `loss_summary_kind` to `"delta"` preserves it: zero captured losses render `Loss: 0.0000 -> 0.0000`, one renders `2.3830 -> 2.3830`. The issue's root-cause section is HF-specific so this is outside its written scope, but the PR's stated goal is not delivered on MLX. Either migrate `mlx_sft.py:761` or open the follow-up — a note in the PR body is enough. A one-line hardening that also covers every future unmigrated producer: when the kind is absent *and* `initial == final`, drop the arrow.
- `soup runs` still prints `2.383 -> 2.383` from the DB (`commands/runs.py:106`) — better than `0.000 -> 0.000`, but the same claim the panel now declines to make. Out of scope; worth a follow-up.
- Good call that `loss_summary_kind` never reaches the tracker, because both call sites pass explicit keywords rather than `**result`. I checked.

**Ordering:** this conflicts textually with #919 in nine files — every one an import-block adjacency, `loss_summary` vs `apply_chat_template_override` inserted at the same line, resolution "keep both, alphabetise". No semantic overlap: #919 touches 0 lines containing `initial_loss`. Clean against #923. **Merging this first is cheaper** (one small resolution in #919 instead of nine) and lets the strengthened guard protect the #919/#923 rewrites rather than land after them.

taljeon:
Addressed both blocking items on current `main` at `84c9dea`:

- per-step `NaN` and `inf` are retained, so divergence stays visible and cannot be replaced by the last finite value; finite-only filtering remains on the run-mean fallback
- the guard now parses all 18 migrated trainer modules with `ast` and rejects any literal `initial_loss` result not sourced from `loss_summary`; its alternate-spelling control caught the remaining PPO manual path, which now uses the shared helper too, and it also rejects an explicit bad override after `**loss_summary`

I also applied the non-blocking display hardening: a legacy result with no kind and equal losses renders one value, so MLX no longer shows an unmeasured equal-value arrow. I left `soup runs` unchanged as the separate historical surface noted in the review.

Validation: 12 focused regressions; 108 issue/PPO/sweep/fake-MLX tests; 518 broader loss-result tests (5 skipped, 3 deselected); 23 additional result-path tests; repository-wide Ruff; 18 lightweight trainer imports; changelog validation; compileall; diff and publication checks. The broader optional-dependency run reached 812 passes, with remaining failures limited to packages absent from this local environment. Current `main@71cdaeb4` is merged without force.

Could you please re-review `84c9dea`? The PR remains draft while refreshed CI runs.

MakazhanAlpamys:
Re-reviewed at `84c9dead`. **Both mutations that survived the previous round are now killed**, and I ran them rather than reading the diff.

**The blocking one — NaN/inf retention.** Reverting the per-step extraction from `_numeric_loss` back to `_finite_loss` now fails 3 tests, and the failure is on the **value**:

```
assert math.isnan(summary['final_loss'])   ->   False = isnan(2.1)
```

Two of those three keep `loss_summary_kind == "delta"`, so a flag-only assertion would still have missed it. That is the right shape of test and it is the whole point of the finding.

**The guard.** Reverting `dpo.py` to the pre-PR extraction in a *different* spelling (`_ls = [e["loss"] ...]`) is caught, and the message names `dpo.py`. The spread-then-override variant is caught too, with file and line — `[('dpo.py', 435)]`.

**Your PPO claim checks out, and it is the interesting one.** `git show e9fc37d3:src/soup_cli/trainer/ppo.py` still had `losses = [entry["loss"] ...]` inside `_train_manual` — a *different variable name* from the literal the old guard matched, which is exactly why it slipped through. Reverting only `_train_manual` while leaving `_train_builtin` compliant is still flagged (`[('ppo.py', 768)]`), so `_private_loss_result_lines` is doing real work beyond the "at least one dict uses `loss_summary`" assertion. You found that yourself via the alternate-spelling control, which is the part worth naming.

Counts: your file 12 passed; 23 existing files touching these modules 792 passed / 4 failed — and all four reproduce identically on `main` under the same forced-colour harness, so they are the known latent ANSI assertions, not yours; trainer surface 703 passed; ruff clean.

**Three things for a follow-up, none of which should hold this up:**

1. `changelog.d/0.75.0/908.fixed.md` reads `(#899 in #908)` and is missing `by @taljeon`. House format is `(#NNN by @user)` — attribution is the one thing that has to be right at merge time.
2. `mlx_sft.py:761` and `unlearn.py:320` are outside `_MIGRATED_TRAINERS` and still print `Loss: 0.0000` where a migrated trainer now prints `unavailable`. The headline symptom — the fake `->` arrow — is gone everywhere thanks to your legacy equal-value branch, so this is residue rather than the bug. Worth a follow-up issue rather than widening this PR.
3. Your guard has one **false positive** worth knowing about: aliasing the local to anything other than `loss_summary` flags correct code. The things it misses (`dict(...)` call form, post-hoc item assignment, `.update()`) are all unlikely; the false positive is the one that gets guards deleted, and this repo has already deleted one for exactly that. A line in the assertion message saying the name is load-bearing would cover it.

**One consequence of the fix you should know about, which is an accepted trade rather than a defect.** NaN now reaches `commands/sweep.py:504`'s sort key, so a diverged run reorders the leaderboard — `[1.0, nan, 0.5]` sorts `0.5` *after* `nan`. That follows directly from the retention I asked for, so it is on me, not you. I will file it separately with "sort non-finite last" as the suggested fix.

One narrowing I want recorded rather than changed: the literal-string half moved from `trainer_dir.glob("*.py")` to the explicit 18-name tuple, so a *new* trainer that copy-pastes the literal is no longer scanned. Theoretical today — the two modules outside the list never matched it anyway. Keeping the glob for the literal half and the list for the AST half would close it.

Nothing in the code is blocking from my side. Please mark ready for review once CI lands.

taljeon:
Added the requested changelog attribution in `f4f07b1`: the fragment now uses `(#899 by @taljeon in #908)`.

I left the three follow-up items outside this PR as suggested. The existing focused test and current-main changelog validation still pass, and I will keep the PR in draft until the refreshed CI run is green.

MakazhanAlpamys:
Re-reviewed at `f4f07b1f`. **`84c9dead..f4f07b1f` is exactly one commit and one line — the changelog credit — so my mutation evidence from `84c9dead` carries forward unmodified.** Nothing in the code is holding this in draft; please undraft it and I will merge as soon as CI lands (14 jobs pending as I write).

**I re-ran both halves anyway, because "suppress the fake delta" and "still surface real divergence" can be broken independently and a fix that did the first without the second would be worse than the bug.** Each anchor counted exactly once, each mutated module actually imported before the run (an import, not just `py_compile` — a `NameError` gives a false "caught"):

| mutation | result |
|---|---|
| `if summary_kind in {"mean", "single"} or (` -> `in set()` | **3 failed** — the mutated function returns `Loss: [bold]2.3830 -> 2.3830[/]`, i.e. #899 verbatim |
| `_numeric_loss` -> `_finite_loss` at `loss_summary.py:41` | **3 failed** — and the failures are on the **value**, `assert False = isnan(2.1)`, not on the kind flag |

The second is the one I care about: with the pre-fix semantics restored, `[2.1, nan]` summarises as a healthy `2.1`, and your tests catch it by asserting the NaN actually arrives. Both halves are independently defended.

900 passed, 7 skipped across the new file plus 22 pre-existing files touching `initial_loss`/`final_loss`, plus the smoke/subprocess/sweep suites. Zero removed test lines — the test diff is purely additive.

**And the ANSI trap is dodged by construction rather than by luck:** the new tests assert on the return value of the pure `_format_training_complete_loss` — the pre-render markup string `"Loss: [bold]2.3830[/]"` — and never touch `result.output`. `git diff ... | grep -E '^\+.*(result\.output|readouterr)'` is empty. That is the right way to write these.

**Two notes for the follow-up issue, not for this PR:**

- `trainer/mole_routing.py:440-442` is the one trainer using the explicit-subscript form rather than `**loss_summary`. It does propagate `loss_summary_kind`, so it behaves today — but the AST guard only requires `initial_loss` to come from `loss_summary`, so a future trainer copying that shape could omit the kind flag and still pass. It reaches the same residue as the `mlx_sft.py` / `unlearn.py` item you already deferred, by a different route; fold it into that issue rather than opening a new one.
- `summarize_training_loss` iterates `log_history` twice (forward, then `reversed()`). Every call site passes a real list — I checked, including `ppo.py:688` — so it is safe today, but a generator would silently yield `unavailable`. LOW, and the `Sequence` type hint already says so.

I also confirmed your check that `loss_summary_kind` does not leak into the tracker: `commands/sweep.py:480` and `commands/train.py:1575` both use explicit subscripts, no `**result` spread anywhere.

**Merge order, since you have two ready at once:** this one first. It conflicts textually with #919 in nine files, and landing it first turns nine resolutions into one. #940 is orthogonal to both and can go whenever.

MakazhanAlpamys:
@taljeon — this went **conflicting** after `c6b61657` (#919, "honour the chat-template override across trainers") landed on `main`, and I would rather tell you exactly where than let you discover it.

Nine files, all the same shape:

```
src/soup_cli/trainer/{bco,dpo,grpo,ipo,kto,online_dpo,orpo,reward_model,simpo}.py
```

`src/soup_cli/commands/train.py` and `src/soup_cli/trainer/distill.py` auto-merge cleanly; only those nine need hands. #919 inserted a chat-template resolution call into the same setup region your change touches, so in most of the nine the resolution should be **both edits kept**, not either side taken — check each rather than resolving the first one and repeating it.

Your required matrix was 14/14 green on `12:49`, so nothing is wrong with the change itself; this is purely a base move. Rebase onto current `main` and I will pick it back up.

One thing worth knowing before you start: the Actions queue is at ~85 runs right now, so the re-run after your push will be slow. That is the queue, not your PR.

taljeon:
Resolved in noreply `5f05cd7` with a no-force merge of `main@40a97fa7`. I checked all nine files individually: each retains both `apply_chat_template_override` and `summarize_training_loss`, while `train.py` and `distill.py` merged cleanly as expected.

Post-merge validation passed 12 #899 regressions, 16 dependency-light #804 checks (2 skipped), all 8 #804 runtime-order seams with explicitly limited import stubs for the absent optional training packages, 324 related trainer tests, repository-wide Ruff, compileall, the exact 22-file diff, noreply/no-coauthor metadata, and publication checks. The PR body records the local dependency boundary rather than presenting the stubs as a real training-stack run.

The later #940/doctor-only `main` advance does not overlap this diff. #908 is back to MERGEABLE/BLOCKED at `5f05cd7`, with 14 refreshed jobs queued and auto-merge off. Thanks for the precise conflict map.

MakazhanAlpamys:
Verified `5f05cd7`. **Merge-ready** — and I checked the merge the way a nine-file conflict resolution has to be checked, because this is exactly where one side silently wins and every test still passes.

**Both features survived in all nine files, verified in both directions.** `git diff origin/main...HEAD` on each of the nine shows they differ from `main` *only* in your two hunks, so #919's call site is byte-identical to main; and `git diff f4f07b1f HEAD` — your pre-merge commit to this head — shows the merge **added** #919's code and removed nothing:

```
+from soup_cli.data.chat_templates import apply_chat_template_override
+        apply_chat_template_override(
+            self.tokenizer, cfg.data.chat_template, console=console
+        )
```

Reachable statement, not a stranded import, and placed identically to main (`dpo.py:110-112`). File-set equality is exact too: `git grep -l apply_chat_template_override` returns the **same 11 paths** on `origin/main` and on your head. No trainer lost it.

**The mutation that matters is the one on the other side.** A conflict resolution is only verified if the suite would notice the *merged-in* feature disappearing, so I broke #919 inside a file you merged rather than only breaking your own change:

```
trainer/dpo.py: call replaced with `_ = apply_chat_template_override`  (module still imports)
2 failed, 53 passed
  TestDpoUsesTheOverrideForLiveRendering::test_marker_reaches_prompt_ids_and_saved_tokenizer
  TestEveryConversationalTrainerIsWired::test_setup_applies_the_override_to_the_loaded_tokenizer[dpo]
```

So the merge is held by tests, not just by the diff. Your own two mutations kill cleanly as well — deleting the `mean`/`single` branch in `commands/train.py:75-78` fails 4, and neutering the `train_loss` mean branch in `loss_summary.py:63-67` fails 1.

One thing I checked because it looked like a hole and is not: the `TestEveryConversationalTrainerIsWired` parametrize names only eight of the nine, with `online_dpo` absent. Mutating `trainer/online_dpo.py` the same way still fails — `test_online_dpo_override_beats_fallback_on_a_loaded_tokenizer` catches it. All nine are individually guarded.

Also verified: **zero** deleted test lines across the merge (`tests/` is `+209 / −0`), no conflict markers anywhere in `src/`, `tests/` or `changelog.d/`, `merge-tree` clean against current `main` (which has moved on to `049b64bc` with no overlap), fragment correct at `changelog.d/0.75.0/908.fixed.md` with house credit. 67 passed on the three primary files, 184 on the wider `-k` selection, 728 across 18 trainer and result-shape suites, 93 on the repo ratchets, and `test_cli_startup_is_light.py` green — `loss_summary.py` is stdlib-only, so no torch reaches the light path.

The ANSI trap does not apply here, and for the right reason: your tests assert on the **return value** of `_format_training_complete_loss`, a plain string that is never rendered. Extracting the renderer so it can be tested without a Console is what makes that possible — that was a good structural call, not just a convenient one.

**One follow-up rather than a block.** `trainer/mlx_sft.py:770-773` and `trainer/unlearn.py:320-321` still build `{"initial_loss", "final_loss"}` by hand with a `0.0` fallback, and neither is in `_MIGRATED_TRAINERS`. Your `summary_kind is None and initial == final` fallback at `commands/train.py:75-77` does remove the fake **arrow** on those backends, so #899's headline symptom is fixed there too — but a zero-loss MLX run still prints `Loss: 0.0000` rather than `unavailable`, and nothing guards a regression. Those two do not use HF `log_history`, so leaving them out looks deliberate. I will file it.

Merging on green; the Actions queue is deep.

MakazhanAlpamys:
Merged as `6ff3ca2d`. Thank you.

`soup train` reported a loss delta for runs too short to have one, computing it against a baseline that did not exist yet — **a fabricated number presented in the same format as a real measurement**. That is the failure mode that teaches people to trust a display, which makes it worse than an obviously missing value.

**The nine-file conflict resolution is why this took the care it did**, and I checked it the way a nine-file resolution has to be checked: hunk by hunk against both sides, rather than by confirming the merge was clean. Git reporting no conflict is not evidence about a resolution of that size.

Before merging I composed this with the other nine PRs that were ready and ran the affected suites plus the repo-wide ratchets on that tree: **967 passed, zero failures**. That step is not ceremony — it caught a real break in a tenth PR (#963), which I held rather than merged. Yours composes with #931, which touches `trainer/sft.py` and `trainer/pretrain.py` in the same region — both clean, and the composed run is green.

Credited in `CONTRIBUTORS.md` (`f491ac5b`). #899 closed with credit.

PR #909, opened 2026-09-12
fix(commands): refuse unknown config keys in soup plan and soup apply

### LINKED ISSUE #894: soup plan / soup apply never run the unknown-config-key check
`commands/plan.py::_load_yaml_config` reads the YAML with `yaml.safe_load` and hands the dict to `build_plan(cfg)`; it never constructs a `SoupConfig` and never calls `find_unknown_config_keys`. Since v0.75.0 every other reader refuses an unknown key (#627, #879), so `soup plan` can happily print a plan for a config `soup train` will refuse one command later.

Found in the v0.75.0 release review (pre-existing, not caused by the flip); documented in docs/backends-and-ops.md as the named exception.

**Fix path:** call `find_unknown_config_keys` + `format_unknown_keys(include_deadline=False)` on the raw mapping in `_load_yaml_config` and exit non-zero with the same message `soup train` prints (or route the read through `load_config`).

**Acceptance:**
- [ ] a typo'd key makes `soup plan` refuse with the loader's message, exit 1
- [ ] a clean config plans exactly as before
- [ ] `soup apply --dry-run` behaves the same way
- [ ] the "named exception" sentence in docs/backends-and-ops.md is removed

wangzhengzhuo05:
Fixes #894

### What

`soup plan` and `soup apply` now refuse a config carrying an unknown key, with the same message and exit code `soup train` uses. The "named exception" sentence in `docs/backends-and-ops.md` is removed, since it no longer holds.

### Why

`commands/plan.py::_load_yaml_config` read the YAML with `yaml.safe_load` and handed the raw mapping straight to `build_plan(cfg)`. It never built a `SoupConfig` and never called `find_unknown_config_keys`, so a typo'd key (`training.quantizaton: none`) planned cleanly, was silently dropped, and was refused only one command later by `soup train`.

### How

In `_load_yaml_config`, after the existing mapping check and before `return data`:

```python
    from soup_cli.config.unknown_keys import find_unknown_config_keys, format_unknown_keys
    from soup_cli.utils.terminal import for_terminal

    unknown = find_unknown_config_keys(data)
    if unknown:
        message = format_unknown_keys(unknown, include_deadline=False)
        console.print("[red bold]Config validation error:[/]\n")
        console.print(f"  [red]{for_terminal(message)}[/]")
        raise SystemExit(1)
```

Notes on the shape:

- `SystemExit(1)`, not `ValueError`: `plan_cmd`/`apply_cmd` turn the latter into `typer.Exit(2)`, and the acceptance criterion is exit 1 with the loader's message.
- `for_terminal(...)` escapes the key names, which come from the config file.
- The read is deliberately **not** rerouted through `load_config`: `_load_yaml_config` carries null-byte, `is_under_cwd` containment and symlink/TOCTOU checks that `load_config` does not, so they are kept and only the key check is appended.
- The two helpers are imported lazily, matching `commands/sweep.py::_reject_unknown_sweep_params`. No new allowlist and no second validator — `find_unknown_config_keys`/`format_unknown_keys` are reused as-is.

### Tests

`tests/test_issue627_unknown_config_keys.py::TestPlanAndApplyRefuseUnknownKeys` (3 tests, CPU-only, `CliRunner` against `soup_cli.cli.app`):

- a typo'd key makes `plan` exit 1, output contains `unknown config key`, the key path and the suggestion;
- the same config makes `apply --dry-run` exit 1;
- a clean config still exits 0 for both (control — proves it is not a blanket refusal).

```
$ .venv/bin/python -m pytest tests/test_issue627_unknown_config_keys.py -q
107 passed, 3 skipped

$ .venv/bin/python -m pytest tests/test_issue627_unknown_config_keys.py::TestPlanAndApplyRefuseUnknownKeys -q
3 passed

$ .venv/bin/python -m pytest tests/test_v0640_followups.py -q
31 passed

$ .venv/bin/ruff check src/soup_cli/commands/plan.py tests/test_issue627_unknown_config_keys.py
All checks passed!
```

(Coverage is below the repo's 77% gate on those subsets because only part of the suite was run; the coverage floor is a whole-suite setting, not a failure of these tests.)

Mutation check — reverting only the `plan.py` hunk and re-running the new class:

```
FAILED .../TestPlanAndApplyRefuseUnknownKeys::test_plan_refuses_a_typod_key
FAILED .../TestPlanAndApplyRefuseUnknownKeys::test_apply_dry_run_refuses_a_typod_key
2 failed, 1 passed        # the clean-config control still passes
```

Restoring the hunk: `3 passed`. So the new tests guard exactly this behaviour and nothing else.

### AI disclosure

This change was implemented with AI assistance (OpenCode running muse-spark 1.3), then reviewed, run and mutation-checked locally by a human before submission.


MakazhanAlpamys:
Reviewed at `a0e0cd32`. This is a tidy fix and the two judgement calls in your description are both right, so let me say why rather than just agreeing:

- **Keeping `_load_yaml_config` instead of rerouting through `load_config`** preserves the null-byte / `is_under_cwd` / symlink checks at `plan.py:30-45` that `load_config` does not carry. Rerouting would have been the obvious-looking cleanup and would have quietly dropped them.
- **`SystemExit(1)` rather than `ValueError`** — `plan_cmd:78` and `apply_cmd:145` both convert `ValueError` into `typer.Exit(2)`, so a `ValueError` would have missed the exit code #894 asks for. `SystemExit` is a `BaseException` and passes those handlers untouched.

**Verified by execution.** All three commands on the same typo'd config, ANSI-stripped:

```
plan  -> exit 1 | unknown config key 'training.quantizaton' - did you mean 'quantization' or 'quantization_aware'? Refused.
apply -> exit 1 | (identical)
train -> exit 1 | (identical)
```

No new refusal surface: 12 shipped `examples/configs` + `templates` and all 169 catalog recipes plan clean — expected, since you call the same detector on the same raw mapping the loader already checks, but worth having measured. The tests diff is **+76/-0**, a single appended class with no existing assertion touched. That is the right shape for editing a shared test file.

**Mutation.** Disabling the check (`if unknown:` -> `if False:`) fails `test_plan_refuses_a_typod_key` and `test_apply_dry_run_refuses_a_typod_key` while the clean-config control survives; running the `apply` test alone still fails, so `apply` is genuinely guarded and not a free rider on `plan`. 210 passed across the five test files that touch `commands/plan.py`, plus 118 passed on the repo-wide ratchet guards. `ruff` clean.

**Two things before I merge:**

1. **Changelog fragment is missing.** `changelog.d/0.75.0/` has `788 867 872 882 892 905` and no `909`. This changes an exit code a user can observe, so release-checklist step 7 applies. Please add `changelog.d/0.75.0/909.fixed.md`, body starting with `- `, crediting yourself inline:

   ```markdown
   - **`soup plan` and `soup apply` now refuse a config carrying an unknown key (#894 by @wangzhengzhuo05 in #909).**
     `_load_yaml_config` read the YAML with `yaml.safe_load` and handed the raw mapping to
     `build_plan`, so it never built a `SoupConfig` and never ran `find_unknown_config_keys`:
     a typo'd key planned cleanly and was refused one command later by `soup train`. Both
     commands now exit 1 with the loader's message and suggestion.
   ```

2. **`docs/backends-and-ops.md` — the new second line is 105 characters** where the rest of that paragraph wraps at 79-89, so it reads visibly ragged. Simplest fix removes a duplication at the same time: fold the new commands into the existing list one line above rather than adding a standalone sentence —

   ```
   imported, `soup sweep` / `soup doctor --config` / `soup ship --config` / `soup plan` /
   `soup apply` refuse the same way, and the Web UI / API loader raises `ValueError` with
   the same text (the Web UI shows it). Nothing is defaulted and nothing is guessed: ...
   ```

**Two non-blocking notes:**

- Your check refuses **regardless of `loader.UNKNOWN_KEY_SEVERITY`** — I verified by forcing it to `"warn"` at runtime and still getting exit 1 and `Refused.`. That matches `sweep.py::_reject_unknown_sweep_params`, but sweep documents the choice in its docstring, and sweep's stated reason ("no partially-useful result to preserve by continuing") does not transfer cleanly here, because `soup plan` arguably *does* produce a partially-useful result. The behaviour is invisible today since the switch is at `"error"`; it only surfaces if the switch is ever moved again. One sentence of comment would make the decision reviewable a year from now.
- The three new assertions read `result.output` raw. They pass under `FORCE_COLOR=1 TERM=xterm-256color` today — I checked, `3 passed` — but only because each asserted substring lands inside a single `ReprHighlighter` span and clear of Rich's 79-column wrap. The wrap point moves with the number of difflib suggestions, so a longer key path would straddle it. `re.sub(r"\x1b\[[0-9;]*m", "", result.output)` is cheap insurance. This is the class that has turned CI red four times in this repo, though in those cases through Typer's `--help` rendering rather than this path.

Nothing else outstanding. Push those two and I will merge once the checks are green.

wangzhengzhuo05:
Both changes are in `ef35707`.

- **Changelog fragment** — `changelog.d/0.75.0/909.fixed.md` added with your wording; `tests/test_issue487_changelog_fragments.py` is green (17 passed).
- **Docs reflow** — the standalone sentence is gone and `soup plan` / `soup apply` now join the existing refusal list, exactly as you sketched. The paragraph now wraps at 82-87 columns, so the 105-character line is gone.

On the two non-blocking notes I deliberately left both untouched rather than widen this push — the `UNKNOWN_KEY_SEVERITY` switch and the raw `result.output` assertions are unchanged, which keeps the diff a two-file docs change. Say the word and I'll fold either one in as a follow-up.

MakazhanAlpamys:
Re-reviewed at `ef35707`. Both points check out, measured rather than read.

**The 105-character line.** I extracted `docs/backends-and-ops.md` at three revisions and counted lines over 100 characters in that region:

```
merge-base 0e8cc1d3   1   (line 581 — inside a fence, verbatim CLI output)
previous   a0e0cd32   2   (581, plus 592 at 105 chars — prose)
this head  ef35707    1   (581 only)
```

The prose line is gone and what remains is a fenced sample that is not reflowable. Exactly right, and folding `soup plan` / `soup apply` into the existing refusal list rather than adding a standalone sentence removed the duplication too.

Fragment is in `changelog.d/0.75.0/` with credit, and the "named exception" sentence is removed — that closes the fourth acceptance box in #894.

**The core mutation, run again:** deleting the guard, and separately making `find_unknown_config_keys` always report clean, each take **both** the `plan` and the `apply` test red as separate failures. Dropping the key name and suggestion from the message does too. Four mutations, no survivors.

**One thing worth knowing for a future edit rather than for this one.** A fourth mutation — leave the `console.print` and remove only the `raise` — kills the `plan` test but **not** the `apply` test, because `apply` then still exits 1 from `No state file at soup.tfstate; run 'soup plan' first.` So that test's `exit_code == 1` is satisfied incidentally there; what actually carries it is the message assertions. Running `plan` first inside the test would make the exit code attributable. Not worth another round on a nine-line fix — recording it so it is not rediscovered as a surprise.

**Agreed on leaving the two deferred notes alone**, and I want to say why rather than just accepting: keeping a nine-line fix to two files is worth more than folding in an unrelated comment and an ANSI sweep. Your three new assertions read raw `result.output` and pass under `FORCE_COLOR=1` today — latent, not live. If you want either as a separate follow-up later, say so.

110 passed in the full file, ruff clean. `commands/plan.py` is touched by no other open PR, so there is no ordering constraint. Merging once CI lands.

wangzhengzhuo05:
Rebased onto current `upstream/main` as `22b10d6` — the red `test` matrix was not this change.

Root cause of the failures: a duplicate changelog fragment on the base (`changelog.d/0.75.0/768.fixed.md` vs `768.changed.md`), which `test_repository_fragments_match_the_newest_release` rejects. That is already fixed on main by `e27e40c0`, but this branch's run was created at 11:23Z — three minutes before that landed — so it tested the broken base. `include` is unchanged: the same four files, `909.fixed.md` + the docs reflow + the `plan.py` guard.

Locally after the rebase: `tests/test_issue487_changelog_fragments.py` 17 passed. The three `TestPlanAndApplyRefuseUnknownKeys` tests need `typer`/`rich`, which this sandbox does not have installed — they were green on CI at `a0e0cd32`, and CI is re-running on the new head now.

MakazhanAlpamys:
Verified `22b10d6f`. **Merge-ready** — the rebase is content-identical to the `ef35707` I reviewed, and I re-ran everything on the new head rather than assuming a rebase is a no-op.

Confirmed the rebase claim rather than taking it: `changelog.d/0.75.0/` now holds exactly one `768.changed.md`, and `e27e40c0` is an ancestor of your head. So the red matrix on `a0e0cd32` really was the broken base and not this change.

What I ran — and note this is the **first real execution of this head**, since you could not run the suite in your sandbox:

- `tests/test_issue627_unknown_config_keys.py` → **110 passed**, matching your count; with fragments, 127. `ruff check` clean.
- **Mutation**, anchor counted at exactly 1: `if unknown:` → `if False:` — module still imports — kills both refusal tests while the clean-config control survives. It also independently reproduces the thing I flagged last time: on `apply`, the exit code is satisfied incidentally by the missing-state-file path, so the **message assertion** is what actually carries that test. Good that it is there.
- **Baseline** on `origin/main`: the same two behavioural tests fail, the control passes. Correct shape.
- **No new refusal surface, measured rather than reasoned.** All eight shipped `examples/configs/*.yaml` through the real `soup plan` → exit 0 across the board. And the root-level `lora:` spelling that bit the v0.75.0 release review → exit 0, because `remap_root_level_misplaced_keys` is applied before the walk, as it must be. That was the one way this change could have been quietly wrong and it is not.
- The **ANSI risk is still latent, not live**: under `FORCE_COLOR=1` the three asserted substrings each land inside a single `ReprHighlighter` span and clear of the wrap, which falls between `or` and `'quantization_aware'`. The hazard is that the wrap point moves with the suggestion count, not that it is broken today. Deferral stands.

Also worth naming: #894's fourth box was to delete the sentence saying these two commands are the documented exception. You deleted it and folded both commands into the existing refusal list, rather than adding a second paragraph that contradicts the first. Your base is 8 commits behind `main` but `main` touched none of your four files; zero conflict markers.

Only CI is left — the Actions queue is at ~88 runs, so it will be slow. I will merge as soon as it is green. Nothing further needed from you.

MakazhanAlpamys:
Re-verified against current `main` (`b0a63382`), 16 merges after my review. **Unchanged and still merge-ready:** `merge-tree` clean, 110 passed on both the PR head and the merged tree, the mutation still kills both refusal tests, ratchets 138 passed, fragment intact with no duplicate number against anything merged since.

Nothing needed from you — this is waiting on the queue alone.

MakazhanAlpamys:
Merged as `4ad9a901`. Thank you — your second merge here, and this one closes a limitation I had written down and left open.

**Merged on the verdict recorded at `22b10d6f`**, unchanged and not re-derived. All 13 required contexts green on that exact SHA, `mergeStateStatus` CLEAN, and I re-ran the one thing that can change while a PR waits — whether `main` moved into anything your change interacts with. It had: #757 landed today with a package-wide AST scanner that walks **every** command module, `commands/plan.py` included, and #951 added a control-strip ratchet over the same tree. Green on both branches separately says nothing about the merge, so I built it: **149 passed** on the merged tree across `test_issue627_unknown_config_keys.py`, both of those ratchets and the fragment validator.

**Why this one was worth closing rather than noting again.** v0.75.0 made an unknown config key a hard error — a deliberate, documented break. But `soup plan` and `soup apply` read the YAML as a plain mapping instead of going through the loader, so they never ran the check. The consequence is specific and nasty: the command whose **entire job** is to tell you what the run will do was accepting a config the run would refuse. The plan came out clean, built from a config the trainer rejects, and the failure arrived later with nothing tying it back to the key that caused it.

A pre-flight that disagrees with the thing it is a pre-flight for is worse than no pre-flight, because people trust it. Routing both commands through the same detection is the fix that makes the two agree rather than the one that papers over the symptom — and this was on my own v0.75.0 known-limitations list as #894, which means it was exactly the kind of thing that stays open for releases because everyone has read past it.

110 passed on the PR head and an earlier merged tree; the mutation still kills both refusal tests; ratchets 138 passed.

Credited in `CONTRIBUTORS.md` (`4d98b6ad`). #894 is closed with credit.

One apology for the wait, which was not about your work: Actions here produced no terminal outcome at all between 2026-09-13T18:30Z and today 15:08Z, with about a hundred runs stacked behind it. Cleared now.

Issue #347, opened 2026-08-08
Layer streaming: add an explicit override for a refused VRAM pre-flight (training.stream_vram_override)


MakazhanAlpamys:
Follow-up to #327, which stays open on the constant itself.

`LOGITS_BYTES_PER_ELEMENT = 14` is deliberately kept (see the decision comment on
#327): it is the only value under which neither published measurement grid
under-predicts, and an under-prediction on Windows/WDDM is not an exception but a
silent spill to host memory (measured: 9.27 GB allocated on a 4.29 GB card with no
exception raised).

But the asymmetry justifies the **default**, not the absence of an exit. Today a
refusal is a wall: `decide_stream_fit` returns `fits=False` and
`stream_setup._stream_budget_lines` raises. A user who knows their own stack has no
way through except lowering `max_length` or `batch_size` to a configuration they did
not want.

Measured cost of the over-prediction, DPO batch 1 / seq 768 / vocab 151936 /
rows=2 (233,373,696 logit elements, so 222.56 MiB per 1 B/elt):

| slope | predicted | vs 3166.20 MiB budget | vs 3092.46 MiB measured |
|---|---|---|---|
| 14 (shipped) | 3541.84 MiB | **refused**, −375.6 | +14.53% |
| 12.311 (this-box fit) | 3165.94 MiB | passes by 0.26 MiB | +2.38% |
| 12 (loss path alone) | 3096.72 MiB | passes by 69.5 MiB | +0.14% |

## Proposed fix

Add an explicit, opt-in override — `training.stream_vram_override` (bool) and/or a
`--stream-assume-fits` flag — that:

- prints the prediction, the measured free VRAM and the shortfall,
- names the Windows/WDDM consequence explicitly (it does not raise; it spills),
- then proceeds instead of raising.

Default stays `14`. No row of `tests/test_v07203.py::TestPeakVramReproducesTheMeasuredGrid`
is touched. The estimator keeps its "never under-predicts" property; only the
*action taken* on a predicted miss becomes the user's call.

**This is the real closure of #327 by substance:** the symptom is "configurations
that would fit are refused with no way through", and this removes the "no way
through" half without trading a documented over-prediction for an undocumented
under-prediction.

## Acceptance criteria

- [ ] Schema field with a field-validator; footgun-reject when set while
      `stream_layers` is false (mirrors the existing `stream_source`/`stream_buffers`
      gates).
[...]

Issue #714, opened 2026-09-06
Re-measure the published T4 forecast: its 2.20 TFLOPS predates the #617 emulated-bf16 probe fix


MakazhanAlpamys:
## The gap

`benchmarks/run-t4-colab-free-tier.md:64` publishes a throughput forecast for a Tesla T4:

```
forecast      31-46 tok/s (from 2.20 TFLOPS measured on this card @ 1185 MHz)
```

That 2.20 TFLOPS was measured by `measure_gemm_tflops` **before** #648 (`f14bf42`), when the probe hard-coded `torch.bfloat16`. A T4 is Turing: it has no native bf16, so that number is a bf16-through-emulation figure for a run that streams in fp16. The record's forecast bracket is therefore derived from a GEMM the run never performs.

Everything else in that file — the 2.91 GB measured peak, the store size, the buffer arithmetic — is unaffected. **Only the `forecast` line and the TFLOPS it derives from are stale.**

## Why this is worth a ticket rather than a footnote

#617 stated that nobody here has a Turing card, and that was the reason its acceptance criteria were scoped to "the probe uses the correct dtype" rather than "the resulting number is validated". That premise turns out to be wrong in one specific way: there **is** a published pre-Ampere measurement in this repo, taken on Colab's free tier, which anyone can reach. So the one number #648 makes stale in substance is also the one number that can actually be re-measured.

It also matters more than a typical stale figure, because the free tier is the audience this whole streaming feature exists for. A forecast that is pessimistic by an unknown factor on the free tier is the worst place for it to be wrong.

## What to do

1. Re-run the same configuration on a Colab free-tier T4 against current `main`.
2. Confirm the panel now prints `using float16` — that is the #648 fix showing up at the point of measurement.
3. Record the new TFLOPS and the new forecast bracket beside the measured throughput, so the bracket can be checked against reality rather than only reported.
4. Update `benchmarks/run-t4-colab-free-tier.md` **in place with the correction visible** — this folder is a working record, not a report assembled afterwards, so the old number and the reason it was wrong stay in the file rather than being overwritten silently.
5. Add the row to the `benchmarks/README.md` index if the re-measurement lands as its own file instead.

## Acceptance criteria

- [ ] A T4 run on current `main` with `training.stream_layers: true`, same model and shape as the original record
- [ ] The pre-flight panel line quoted verbatim, showing the resolved dtype
[...]

PR #910, opened 2026-09-12
fix: repin proof-4gb.ipynb to a released version and re-run it on a T4

### LINKED ISSUE #844: notebooks/proof-4gb.ipynb is stale: installs from git main, calls 0.73.0 "published", cites closed #347 as open, and its section 4 has never produced a recorded output
## The problem

`notebooks/proof-4gb.ipynb` is the free-tier proof the README sends people to (`README.md:68-69`: "check it yourself on a free Colab T4 ... caps the process to 4 GB, then asserts a streamed model is bit-identical to a normal one"). It was last edited on 2026-08-14 and has gone stale in four separate ways. Each was checked against the current tree:

1. **It installs whatever `main` is on the day you run it.** Cell 2: `%pip install -q "soup-cli[train] @ git+https://github.com/MakazhanAlpamys/Soup.git"`. Cell 1 explains why: the T4 precision fix (#385, #387) "is on `main` and is not in a release yet, so on the published 0.73.0 the next cell raises `ImportError`. Switch this line back to `soup-cli[train]` once the next version ships." That fix shipped in **v0.73.1** (`## [0.73.1] - 2026-08-14`): `bf16_fp16_flags` and `resolve_stream_dtype` are absent at the `v0.73.0` tag and present at every tag from `v0.73.1` to `v0.74.0`. The switch-back never happened, and three more releases have shipped since.

2. **It calls 0.73.0 the published version.** The current release is **0.74.0** (`pyproject.toml:7`, `src/soup_cli/__init__.py:3`, and the newest version on PyPI).

3. **It cites #347 as open.** Cell 5 (section 3): "That is [#347], it is open". #347 was closed as completed on 2026-08-13. It shipped `training.stream_vram_override` in v0.73.1 (`CHANGELOG.md:1691`, under the 0.73.1 heading), and the field is in the v0.74.0 schema (`src/soup_cli/config/schema.py:3227` on `main`). That field fixes the exact caveat the cell describes: `mem_get_info()` reports the whole card, so the pre-flight ignores the per-process cap. The published T4 run saw that happen: the pre-flight read "free VRAM 15.10 GB" under a 4.00 GB cap (`benchmarks/run-t4-colab-free-tier.md:63` and `:104-114`).

4. **Section 4 has never produced a recorded output.** No cell in the committed notebook has saved outputs. `benchmarks/run-t4-colab-free-tier.md:131-132` says: "The notebook's section 4 — streamed-vs-resident bit-exactness on the T4 — produced no captured output. It is recorded here as **unrun**, not as a pass." `docs/performance-and-quantization.md:366` says the same. So the one assertion the README link advertises is the one this project has never seen pass on a T4.

## Why the git install is a problem in practice

A notebook that installs from `main` records whatever `main` was that day, and that is not something anyone can reinstall later. #425 is the concrete case. A T4 user followed section 5 on "Soup 0.73.2 from `main`" and crashed on the first training step (`_amp_foreach_non_finite_check_and_unscale_cuda not implemented for 'BFloat16'`). The fix, #429, is in `v0.74.0` and not in `v0.73.3`. So **0.74.0 is the first release that has both the T4 precision fix and the T4 GradScaler fix**, and it is the natural pin.

## What to change

- [ ] Cell 2: `%pip install -q "soup-cli[train]==0.74.0"`. Keep the `torchao` uninstall. #389 made the failure message clear but did not remove the need for the uninstall, so that paragraph is still accurate.
- [ ] Cell 1: drop the "from git, not PyPI" rationale and the 0.73.0 wording. Say which release the notebook is pinned to and why (0.74.0 is the first with #385/#387 and #429).
- [ ] Cell 5 and the section 5 config (cell 12): replace the "#347, it is open" caveat. Add `stream_vram_override: 4000000000` to the `training:` block, so the pre-flight checks against the same 4 GB the allocator enforces. Change the prose to say that. The recorded run predicted ~3.02 GB against that cap (`run-t4-colab-free-tier.md:62`), so this should not refuse the config, but confirm it in the same run.
- [ ] Add a short cell that prints `torch`, `transformers`, `peft`, `bitsandbytes` and `soup_cli` versions. The published record could not give them (`run-t4-colab-free-tier.md:139-140`: "Library versions ... were **not recorded**").
- [ ] Run every cell on a free Colab T4 and commit the notebook **with outputs**. At minimum, section 4's `max |streamed - resident|` and `torch.equal` lines must be visible. Update `benchmarks/run-t4-colab-free-tier.md`'s "unrun" bullet with the result, pass or fail. A failure is a wanted result and gets its own issue.

## A cheap guard, so this does not happen a fifth time

One offline test that reads the notebook JSON and asserts:

- the install cell pins `soup-cli[train]==X.Y.Z`, with no `git+` URL;
- `X.Y.Z` has a `## [X.Y.Z]` heading in `CHANGELOG.md`, i.e. it is a released version, not a version on `main`;
- every `from soup_cli.<module> import <name>` in a code cell resolves against the current tree, so a rename on `main` fails CI rather than surfacing on the next pin bump.

Deliberately **not** proposed: checking issue state from CI (network, rate limits, flaky). The editorial fix is cheaper. Don't write "it is open" about an issue in the notebook. Link it, or name the release that closed it.

## Hardware needed

A free Google Colab account with a T4 runtime, which anyone can get. The maintainer cannot substitute local hardware: the dev box is now an RTX 5070 Laptop (Blackwell), which has bf16, and the whole point of this notebook is a card without it.

If you do the Colab run, #714 asks for a re-measured T4 forecast line from the same kind of session. One session can cover both.

## Honest limits

- Section 4 checks **forward** logits of `HuggingFaceTB/SmolLM2-135M-Instruct` in the card's streaming dtype (fp16 on a T4), **unquantized**: `shard_checkpoint` is called without `quant`, and at v0.74.0 that defaults to `QUANT_NONE`. A pass shows the dense fp16 streamed forward is bit-exact on sm_75. It says nothing about NF4, where #776 records streamed-vs-resident divergence on sm_120, and nothing about gradients.
- The 8B resident NF4 reference that `run-t4-colab-free-tier.md:133-134` calls "the obvious next measurement" is a bigger job and out of scope here.
- Pinning trades freshness for reproducibility. The pin can lag `main`, and the guard above allows that on purpose. Whoever bumps it re-runs the notebook.

VanshSharmaPES:
## Summary

- Repins `notebooks/proof-4gb.ipynb`'s install cell to the released `soup-cli[train]==0.75.0`, the exact release this run's committed outputs were verified against, instead of an unreproducible `git+...@main` install. (v0.74.0 was actually the first release carrying both #385/#387 and #429 — the notebook's install cell explains that distinction; bump the pin only after re-running end to end.)
- Fixes the stale "0.73.0 is published" and "#347, it is open" claims — #347 was closed by `training.stream_vram_override` in v0.73.1, now wired into the section 5 config so the pre-flight checks the same 4 GB cap the allocator enforces.
- Adds a cell printing `torch`/`transformers`/`peft`/`bitsandbytes`/`soup_cli` versions.
- Handles a new wrinkle found while doing this: Colab's free-tier T4 runtime now defaults to Python 3.13.15, past soup-cli's intentional `<3.13` ceiling. Rather than widen that project-wide, CI-tied bound off one Colab run, the install cell detects Python ≥3.13 and falls back to `pip install --ignore-requires-python`.
- Ran the notebook end to end on a real free Colab T4 and committed it **with outputs**: section 4's streamed-vs-resident forward comparison is bit-exact (`torch.equal = True`), section 5 trains Llama-3.1-8B under the 4 GB cap with a measured peak of 1.83 GB. Updated `benchmarks/run-t4-colab-free-tier.md`'s "unrun" note with this result.
- Adds `tests/test_issue844_notebook_pin_guard.py`, the cheap guard the issue asked for: asserts the install cell pins an exact released version (no `git+` URL), that the pinned version has a `## [X.Y.Z]` heading in `CHANGELOG.md`, that every `from soup_cli...` import in the notebook still resolves against the current tree, that every code cell carries recorded output, and that the pinned version string actually appears in that recorded output (so a pin bump left un-re-run fails here).

Closes #844.

## Test plan

- [x] `pytest tests/test_issue844_notebook_pin_guard.py` — 7 passed
- [x] `pytest tests/test_issue487_changelog_fragments.py` — validates the changelog fragment
- [x] Notebook executed end to end on a free Colab T4; outputs committed


MakazhanAlpamys:
This is more than #844 asked for, @VanshSharmaPES.

You got real free-tier T4 hardware, ran the whole notebook, and committed the outputs — which means this project now has a measurement it had never taken: **section 4 is bit-exact on Turing** (`max |streamed − resident| = 0.0`, `torch.equal = True`). The record had explicitly marked that as *unrun rather than a pass*, and it is now a pass. You also kept the existing step that overwrites `lora_B` with random values first, so the comparison cannot succeed vacuously against a zero-initialised adapter — that is the detail that makes the result worth anything.

**What I verified rather than read:**

- **The pin is real and the APIs exist at that tag.** All 8 `from soup_cli...` imports in the notebook are present at `v0.75.0`, checked with `git show v0.75.0:...` rather than against the working tree, and `training.stream_vram_override` is in the 0.75.0 schema. A notebook pinned to a release but calling a `main`-only symbol is the same defect one layer down; this is not that.
- **The run is genuine.** Sequential `execution_count` 1 to 11, every code cell carrying outputs, the cap shown to bite (`refused, as it should be: CUDA out of memory. Tried to allocate 4.29 GiB`), and the pre-flight printing `free VRAM 4.00 GB (training.stream_vram_override; driver reports 15.15 GB)` — #347 confirmed live rather than reasoned about. Scanned the whole 260 KB of JSON: no tokens, drive mounts, local paths or emails.
- **The guard has teeth.** Three mutations, three kills, each naming a different test: reverting to `git+...` fails 3, pinning a well-formed but unreleased `0.76.0` fails 1 (naming the missing `## [0.76.0]` heading), renaming `shard_checkpoint` fails 1.

**Two things before merge:**

1. **`docs/performance-and-quantization.md:366` still says the comparison is unrun.** `git diff -- docs/` on your branch is empty, and that line currently reads *"the notebook's streamed-vs-resident comparison produced no captured output, so it is recorded as unrun rather than as a pass"* — while `benchmarks/run-t4-colab-free-tier.md:131` now says it has been run. #844 cites that exact file as one of the places making the stale claim, so after this merges the repo contradicts itself. Please update it to state the result **with the same scope caveat you used in the benchmarks file**: SmolLM2-135M, fp16, unquantized — not the 8B NF4 configuration that page is about. The same sentence also says the pre-flight read "free VRAM 15.10 GB — the device, not the per-process cap"; a half-clause noting the notebook now sets `stream_vram_override` would stop that reading as a live defect too.

2. **The benchmarks bullet has no date and no version.** It gives the hardware but not when the run happened, which soup-cli release it used, or the library versions — while the bullet two lines below it complains that library versions "were **not** recorded". Your notebook has all of it in a committed output cell, but a notebook can be re-run and overwritten; `benchmarks/` is the durable record and that is what the folder is for.

**One optional thing I would genuinely value, and it is the most interesting part of your run.** Your figures differ substantially from the ones the document body records, on the same card:

| | original run | your run |
| --- | --- | --- |
| base store | 3.60 GB | **5.70 GB** |
| VRAM buffers | 2 × 113 MB | 2 × 113 MB + a 1051 MB large-layer slot |
| resident | 2101 MB | 0 MB extras + adapters |
| measured peak | 2.91 GB | **1.83 GB** |

That is self-consistent — the embeddings and head moved out of the resident allocation into a streamed large-layer slot, which explains both the larger store and the smaller peak — but the file will now carry two sets of figures with no note of the change, and `benchmarks/README.md:26` still summarises this record as "peak 2.91 GB against a predicted 3.02". Two sentences recording the second measurement and its cause would make the record complete.

**Minor:** the fragment is `844.fixed.md` with no author credit; `CONTRIBUTING.md:352` prefers the PR number and every other fragment this window reads `(#issue by @user in #PR)`. `910.fixed.md` and `(#844 by @VanshSharmaPES in #910)`. This is only your second merge and the credit line matters.

**And one for the guard, since it costs four lines:** assert that every code cell has a non-empty `outputs` list, so nobody can bump the pin without re-running. The whole value of this PR rests on the outputs being there, and nothing currently protects that.

On `--ignore-requires-python` for Python 3.13: the right call. Widening a CI-tied project bound off one Colab session would have been the tempting move and the wrong one, and your markdown explains the trade honestly and tells the reader to file an issue rather than blame the notebook.

VanshSharmaPES:
Pushed both required fixes plus the optional two, and the output-assertion guard.

**`docs/performance-and-quantization.md:366`** — updated to state the section-4 result as a pass, with the same scope caveat as the benchmarks file (`SmolLM2-135M`, fp16, unquantized — not the 8B NF4 config that bullet is about), and added the half-clause noting the notebook now sets `training.stream_vram_override`, so the "free VRAM 15.10 GB" line no longer reads as a live defect.

**Undated/unversioned benchmarks bullet** — pulled the exact figures straight from the notebook's own version-print cell rather than guessing: `soup-cli 0.75.0`, `torch 2.11.0+cu128`, `transformers 5.16.1`, `peft 0.20.0`, `bitsandbytes 0.50.2`, dated `2026-09-13` (the commit timestamp). Noted explicitly that the library-version gap the rest of the document complains about doesn't apply to this second run.

**Second measurement (optional, done anyway)** — recorded section 5's re-measurement in `run-t4-colab-free-tier.md`: base store 5.70 GB (was 3.60), the added 1051 MB large-layer VRAM slot alongside the original 2×113 MB buffers, 0 MB resident beyond adapters (was 2101 MB), measured peak 1.83 GB (was 2.91 GB) — with the cause (embeddings/head moved into the streamed slot between runs) stated, not just the numbers. Original figures left standing, new ones added alongside, per the file's own working-record convention. `benchmarks/README.md:26` now mentions both.

**Changelog credit** — renamed `844.fixed.md` → `910.fixed.md` and added the `(#844 by @VanshSharmaPES in #910)` line matching this window's other fragments.

**Output-assertion guard** — added `test_every_code_cell_has_recorded_output` to `test_issue844_notebook_pin_guard.py`: walks the notebook JSON and fails if any code cell's `outputs` list is empty. Verified it actually catches something: cleared one cell's outputs in an in-memory copy and confirmed the assertion fires naming that cell's index, before confirming the real notebook is clean (11/11 code cells non-empty).

`ruff check` clean, and `scripts.assemble_changelog.validate_fragments()` validates all 7 fragments in the tree including this one.

One note on scope: I left the original run's numbers in `run-t4-colab-free-tier.md` untouched rather than reconciling them with the new run, since the file's own header says it's "captured from the notebook's printed output... no post-hoc re-measurement" — treating the second run the same way (add, don't overwrite) seemed like the reading consistent with that stated convention. Happy to change the presentation if you'd rather.

MakazhanAlpamys:
Re-reviewed at `be5b5326` — two commits past my last look (`6fbebda0` renamed the fragment, `be5b5326` carried the content you said you missed staging). **All five requests are addressed, and the notebook itself is untouched since `e06864c`, so my round-1 verification of it still stands byte-for-byte.**

**The item I was most prepared to block on — `stream_vram_override` — checks out in the restrictive direction, and it makes the proof stronger rather than weaker.** `resolve_available_vram_bytes` (`utils/layer_stream.py:1331-1344`) replaces the driver reading in *either* direction, and you set it to `4000000000`, byte-identical to the notebook's own `BUDGET_BYTES = 4 * 1000**3` allocator cap. So the pre-flight's yardstick went **down** from the driver's 15.15 GB to 4.00 GB — a harder gate than the original run faced — and the panel says so: `free VRAM 4.00 GB (training.stream_vram_override; driver reports 15.15 GB)`. Predicted 1.97, measured 1.83, both under a cap that was independently shown to bite at 4.29 GiB. That is the acceptance-test use the docs describe, not the bypass use, and I want it on the record that I checked rather than assumed.

**The run is genuine, and the check is not just `execution_count` ordering** (1->11 sequential, all 11 code cells carry output). Three independent outputs corroborate one session: printed `VRAM 15.6 GB` -> printed `fraction 0.256` -> implies 15.625e9 B = 14.55 GiB -> the OOM message reads `total capacity of 14.56 GiB`. Hand-editing does not produce that.

**And your stated cause for the second measurement is arithmetically exact, not merely plausible** — worth adding to the record, because it is better evidence than the prose gives it credit for: 5.70 - 3.60 = **2.10 GB**, exactly the 2101 MB resident that vanished, and 2101 = **2 x 1051 MB**, i.e. embeddings + lm_head, one of which is the new streamed large-layer slot.

Scope caveat verified against code rather than prose: §4 uses `SmolLM2-135M-Instruct`, prints `dtype=float16`, and calls `shard_checkpoint` without `quant`, whose default is `QUANT_NONE` (`layer_shard.py:1001`). Accurate. The `v0.75.0` tag exists and `CHANGELOG.md:15` matches. Fragment is correct and credited. No tests removed. 6 passed in the new guard, 39 combined with the changelog and README-sync suites, ruff clean.

---

**One line to fix, and then I merge — `docs/performance-and-quantization.md:366`.**

It now links a notebook that contradicts it. The bullet still carries run 1's figures (`measured peak 2.91 GB`, `the pre-flight read free VRAM 15.10 GB`) while the linked notebook prints `1.83 GB` and `free VRAM 4.00 GB` in committed output — one clause later on the same line. `benchmarks/README.md:26` and the run record both carry both figures now; this page is the one place that does not. That is the same "repo contradicts itself" shape I flagged in round 1, one file over.

While you are on that line, the new sentence is also backwards: *"it is that override, not the fit decision, that enforces the real budget"*. The override does not enforce anything — it replaces the yardstick so the fit decision does (`layer_stream.py:1331`). True of run 1, which had no override; inverted for the state the sentence now describes.

**Also cheap, and it makes your own guard match its docstring.** I mutated the pin from `soup-cli[train]==0.75.0` to `==0.74.0` (anchor counted: exactly 1 occurrence in the whole 260 KB JSON), left the outputs untouched, and **all 6 tests passed** — while the docstring says *"a stale pin bump is exactly the case that would leave outputs behind from an older run"*. The kill is free: assert the pinned version string appears in a recorded output. It already does, twice — cell 3 prints `soup 0.75.0` and cell 4 prints `soup_cli 0.75.0`. (The output-presence check itself has real teeth: blanking cell 10's outputs fails with `code cell(s) at index [10] carry no output`, naming the cell.)

---

**Four completeness notes on the benchmarks record, none of which I am asking you to fix** — recording them so the file's future reader is not misled:

- The forecast row also moved ~9x (31-46 tok/s -> 288-423 tok/s, `2.20 TFLOPS @ 1185 MHz` -> `20.39 TFLOPS ... now using float16 @ 525 MHz`) and is not reconciled; `run-t4-colab-free-tier.md:64` and `:122` still discuss the old figure with no pointer. The `now using float16` wording suggests the GEMM probe changed dtype between versions, which is arguably the more interesting of the two deltas.
- Run 2's prediction error (+7.7% over-prediction, 1.97 vs 1.83) is omitted, though the file calls run 1's +3.8% *"the one result here worth carrying forward"* and this is a second point on that exact property, in the same safe direction.
- The `2026-09-13` date appears nowhere in the notebook — I grepped for `202\d-\d\d-\d\d` and `\b202\d\b`, zero hits. It is the commit timestamp in your `+05:30` timezone; in UTC `e06864c` is 2026-09-12T19:50. For a record whose whole job is "when", say the date is derived from the commit.
- `run-t4-colab-free-tier.md:104-117` still ends *"this run is the first time the gap has been seen rather than reasoned about"* with no pointer to the fact that the notebook now closes it by setting the override.

**Your scope call — leaving run 1's numbers standing rather than reconciling them — is right**, and consistent with the file's own header ("no post-hoc re-measurement"). Do not change it.

CI has not started on `be5b5326` yet. Push the `:366` line and I will merge as soon as it is green.

VanshSharmaPES:
Pushed at 30665bf.

`docs/performance-and-quantization.md:366` now states both runs — peak **2.91 GB** / free VRAM **15.10 GB** for run 1, peak **1.83 GB** / free VRAM **4.00 GB** for run 2, with the cause (embeddings/`lm_head` moved into the streamed large-layer slot between runs) and both prediction errors (+3.8% / +7.7%, both over-predicting). Matches what `benchmarks/README.md` and the run record already carry, so the linked notebook and this page no longer disagree.

Also fixed the inverted sentence: it now says the override replaces the yardstick (`resolve_available_vram_bytes`) that the fit decision is taken against, rather than claiming the override itself enforces the budget.

Left the four completeness notes (forecast delta, run-2 prediction error, the date's actual provenance, the "first time the gap has been seen" line) as noted rather than fixed, since you flagged them as informational rather than blocking.

Thank you for verifying the arithmetic and the code paths rather than the prose throughout this review — the 2101 = 2×1051 check in particular is a better piece of evidence than I'd put in the file myself.

VanshSharmaPES:
No CI checks have run on 30665bf (`gh pr checks` reports none) — same gate as #907 hit: a first-time fork contributor needs a maintainer to approve the workflow run before Actions executes, and that approval doesn't carry over across pushes. Could you approve it so CI runs?

Everything in my last two comments (the round-2 fixes and the round-1 fixes before them) is from local `pytest`/`ruff` runs, not CI's — happy to push anything CI turns up once it's actually running.

MakazhanAlpamys:
Approved — the runs on `30665bf` are queued now (`34759637182`), so `gh pr checks` will start showing them.

You are right about the gate and it is worth naming for the next person who hits it: a first-time fork contributor needs each workflow run approved by hand, and the approval does **not** carry forward to the next push. So every force-push on this PR needs another approval from me. That is on me to watch, not on you to ask for — ping me if a run sits at zero checks for more than a few minutes and I have not noticed.

Everything else on my last review still stands: the only thing I am asking for is the one line at `docs/performance-and-quantization.md:366`, where the bullet still carries run 1's `2.91 GB` / `15.10 GB` while the notebook it links prints `1.83 GB` / `4.00 GB` in committed output. I will merge as soon as that lands and the checks are green.

VanshSharmaPES:
Thanks for approving — CI is queued at 34759637182, watching it.

On the `:366` line: that fix is already in 30665bf, the same commit CI is running against — pushed and confirmed in my previous comment ([here](https://github.com/MakazhanAlpamys/Soup/pull/910#issuecomment-5653540425)). Just double-checked the raw file on GitHub at that commit to be sure nothing got lost in the push:

```
$ curl -s https://raw.githubusercontent.com/VanshSharmaPES/Soup/30665bf/docs/performance-and-quantization.md | grep -o "measured peak [0-9.]* GB\|free VRAM [0-9.]* GB"
measured peak 2.91 GB
free VRAM 15.10 GB
measured peak 1.83 GB
free VRAM 4.00 GB
```

Both runs' figures are there. If you're seeing the old single-figure version somewhere, it might be a cached diff view from before the push — happy to re-paste the full line here if that's easier to eyeball:

> `**A streamed 8B run now completes on a Turing card ... The first run measured peak **2.91 GB** against a predicted ~3.02 GB ... with the pre-flight reading **free VRAM 15.10 GB**... The committed [notebook](../notebooks/proof-4gb.ipynb) now sets \`training.stream_vram_override\` to that same 4.00 GB, which replaces the yardstick the fit decision is taken against — the pre-flight then reports **free VRAM 4.00 GB**... Against it: predicted ~1.97 GB, **measured peak 1.83 GB** (a +7.7% over-prediction, still the safe direction).`

Let me know if there's something else on that line I'm missing.

VanshSharmaPES:
Pushed at 42b1899. Re-verified every figure in this PR against the notebook's actual committed output rather than the summary I'd given earlier — base store, VRAM buffers, resident, predicted/measured peak, free VRAM, the +7.7% arithmetic — all directly against the notebook JSON, all confirmed, no discrepancies.

The one thing that check surfaced was the date, which you'd already flagged in round 2 as a completeness note rather than a blocker: `run-t4-colab-free-tier.md` said `2026-09-13`, which is the commit's local +05:30 timestamp, not its UTC one (`2026-09-12T19:50:14Z`) — and the date never appears in the notebook's own output either way, so it was provenance I stated imprecisely rather than something the notebook asserts. Fixed to say so explicitly: the date is derived from commit `e06864c`'s UTC timestamp, with the UTC value given and a note on why an earlier version of the line said `2026-09-13`.

This was optional per your last review, but since it was cheap and already flagged, closing it now rather than leaving it as a known loose end. Everything else from round 2 stands as pushed at 30665bf.

MakazhanAlpamys:
Reviewed `42b1899`. **My round-2/3 blocking item is verified closed** — `docs/performance-and-quantization.md:366` now carries both runs (`2.91 GB`/`15.10 GB` and `1.83 GB`/`4.00 GB`), the cause, and both prediction errors, and the inverted sentence reads correctly. The notebook is byte-unchanged since `e06864c`, so my round-1 trace still stands.

**One new blocking finding, and it is a factual error inside the claim this PR exists to make.**

`notebooks/proof-4gb.ipynb`, markdown cell idx=1, repeated in the PR body:

> pinned to **soup-cli==0.75.0** ... It is the **first release carrying both** the T4 precision fix (#385, #387) and the T4 GradScaler fix (#429)

That is false, and I checked it by execution rather than by reading #844's prose:

```
git log --oneline --reverse -S'def align_trainable_dtype_for_fp16' -- src/soup_cli/utils/mixed_precision.py
822de169 fix(trainer): cast bf16 LoRA adapters to fp32 on pre-Ampere fp16 runs (#429)

git tag --contains 822de169
v0.74.0
v0.75.0
```

and `bf16_fp16_flags` (#385/#387) is absent at `v0.73.0`, present at `v0.73.1`. The `#429` CHANGELOG entry sits under `## [0.74.0] - 2026-09-04`. So **v0.74.0 is the first release carrying both** — which is what #844 states in bold and proposes as the pin.

In a PR whose whole purpose is to de-stale a version claim, a wrong version claim is the one defect that cannot be waved through.

**The remedy is a reword, not a repin.** Repinning to 0.74.0 would invalidate the committed outputs, and those outputs are this PR's entire value. Something like: *"pinned to 0.75.0, the release this run was verified against; 0.74.0 was the first release carrying both ..."*. Your guard does not catch this — it only checks the pinned version has a CHANGELOG heading.

**Second item, and it is the one that makes the first dangerous.** The cheap kill I asked for in round 2 is still not there. Reproduced at your head:

```
# replace ==0.75.0 with ==0.74.0 in the notebook, outputs untouched
pytest tests/test_issue844_notebook_pin_guard.py -q --no-cov
6 passed
```

The class docstring at `tests/test_issue844_notebook_pin_guard.py:54-58` says *"a stale pin bump is exactly the case that would leave outputs behind from an older run"*, and #844 says *"Whoever bumps it re-runs the notebook."* Neither is enforced. The kill is two lines — cell exec=2 prints `soup 0.75.0` and cell exec=3 prints `soup_cli 0.75.0`, so asserting the pinned string appears in those outputs closes it. This intersects the first finding directly: if someone "fixes" the prose by repinning, nothing catches the stale outputs. The output-*presence* half does have teeth — blanking cell exec=10 fails correctly.

**Everything else traces.** I checked every number in the PR against committed notebook output rather than against your summary: the T4 identity and `sm_75` are in the notebook's own captured environment cell (exec=4), not merely in Colab metadata; the stack versions are exec=3; cap/fraction/refusal/capacity are exec=5; `max |streamed − resident| = 0.0` and `torch.equal = True` are exec=7; the store breakdown, prediction and free-VRAM figures are exec=10; `+7.7%` is `1.97/1.83−1` to the digit. Run-1 figures trace to the pre-existing `run-t4-colab-free-tier.md:59-78,104`, and the loss curves differ between the two runs (4.5266→4.0674 against 4.529→4.030), so these are genuinely two runs rather than a copy-paste. All eight notebook imports resolve at the pinned tag, `stream_vram_override` included. Fragment, credit, ruff, secrets scan, no deleted tests — all clean. Baseline confirmed: `origin/main`'s notebook against your new guard gives 4 failed, 2 passed.

Two non-blocking notes:

- **"exit 0"** at `:366` is a run-1-only fact. The notebook runs in-process — markdown cell idx=15 says so explicitly — and produces no exit code, so that clause now reads as a shared preamble over both runs.
- **#714 is still open and this run captured exactly what it asks for**: `forecast 288-423 tok/s ... 20.39 TFLOPS ... float16 @ 525 MHz` against the published 2.20 TFLOPS. The measurement exists in your committed output and is recorded nowhere, so #714 stays open on data you already have. Worth a separate comment there — it is a free close for someone.

VanshSharmaPES:
Both blocking items fixed at 812ea51.

**1. False "first release" claim.** You're right — checked it the same way (tag/CHANGELOG), v0.74.0 is the first release carrying both #385/#387 and #429, not 0.75.0. Reworded rather than repinned, per your point about the outputs being this PR's value: the cell now says v0.74.0 was the first release with both fixes, and 0.75.0 is pinned here only because it's the exact release this run's committed outputs were verified against.

**2. Missing output-freshness kill.** Added `test_the_recorded_outputs_are_from_the_pinned_version`: asserts the pinned version string appears in at least one code cell's recorded output. Reproduced your exact mutation (`==0.75.0` → `==0.74.0`, outputs untouched) against it before and after — fails as expected pre-fix, and I confirmed it still passes clean against the corrected notebook.

**Non-blocking notes:**
- `:366`'s "exit 0" was indeed run-1-only (the second run is in-process per the notebook's own markdown). Reworded so it's not stated as shared.
- Filed the T4 forecast re-measurement on #714 with the exact numbers from this run's committed output (20.39 TFLOPS @ 525 MHz vs. the published 2.20 TFLOPS @ 1185 MHz) — https://github.com/MakazhanAlpamys/Soup/issues/714#issuecomment-5661102971.

Notebook is otherwise byte-unchanged from e06864c except the one paragraph in cell idx=1; diff is 7 lines against `main`.

MakazhanAlpamys:
Verified `812ea512`. **Both code items are closed** — and one non-code item still blocks, which I would normally not hold a PR for and am holding this one for a specific reason.

## Closed

**The notebook claim is corrected**, and I re-verified the version facts myself rather than from your reply:

```
git log -S'def align_trainable_dtype_for_fp16' --oneline --all -- src/
822de169 fix(trainer): cast bf16 LoRA adapters to fp32 on pre-Ampere fp16 runs (#429)
git tag --contains 822de169   ->  v0.74.0, v0.75.0

git log -S'def bf16_fp16_flags' --oneline --all -- src/
5793e83c (#385, #387)          ->  v0.73.1 ... v0.75.0
```

Earliest tag containing **both** is v0.74.0. Your new wording says exactly that, and explains why 0.75.0 is the pin anyway. It is now true.

**The output-freshness kill is there and it discriminates.** Baseline 7 passed, was 6. Mutation — bump the pin in the cell *source* only, outputs untouched, verified `0.75.0 still in outputs, occurrences: 2`:

```
FAILED ...::test_the_recorded_outputs_are_from_the_pinned_version
E  AssertionError: no code cell's recorded output mentions 0.74.0 -- the pin was bumped
   without re-running the notebook, so every output above is from an older run
1 failed, 6 passed
```

`test_the_pinned_version_is_actually_released` still passes under that mutation, so the new test is the only thing firing — it is not riding on a neighbour. That closes the hole where repinning to "fix" the prose would have silently kept stale outputs.

**And the outputs did not move**: `git diff 42b1899b 812ea512 -- notebooks/` is 7 lines, all the markdown cell. Zero code cells, zero outputs. Every number I traced last round still stands, so nothing needs re-tracing.

I also noticed the fix you found yourself: `docs/performance-and-quantization.md:366` — the `exit 0` clause was a run-1-only fact presented as shared, now correctly reworded to note the second run was in-process and has no exit code to quote.

## The one remaining blocker: the PR description

```
gh pr view 910 --json body -q .body | grep -n "first release"
3:- Repins ... to `soup-cli[train]==0.75.0` (first release carrying both #385/#387 and #429)
```

The artifact is fixed; the description is not. **On a squash merge the PR body becomes the commit message**, so the false claim would land in `main`'s permanent history immediately after being explicitly corrected in the file it describes. That is a worse outcome than never having fixed it, because the history then contradicts the notebook and the git log is the thing people trust when the two disagree.

It is a description edit, no code, no re-run, no new CI. Once it is done I merge on green.

One nit, not a hold: the fragment describes the pin guard but not the output-freshness assertion you added this round. "A future stale pin ... fails CI" arguably covers it.

VanshSharmaPES:
Fixed the PR description — it still had the same false "0.75.0 is first release carrying both #385/#387 and #429" claim that was corrected inside the notebook itself at 812ea51. Now correctly attributes that to v0.74.0 and explains why 0.75.0 is pinned. This should clear the remaining gate you flagged on #967.

MakazhanAlpamys:
Verified. The description now reads:

> (v0.74.0 was actually the first release carrying both #385/#387 and #429 — the notebook's install cell explains that distinction; bump the pin only after re-running end to end.)

That was the last item. **Merge-ready** — I will merge as soon as CI is green on `812ea51`'s successor.

You did the right thing with the wording, incidentally: rather than deleting the claim you kept the fact and moved it to where it is true, so the squash commit now explains *why* 0.75.0 is the pin despite 0.74.0 being the first release with both fixes. A reader of `git log` gets the reasoning rather than a silence.

Everything else from the last round stands as verified: the version facts re-checked independently with `git log -S` and `git tag --contains`; the output-freshness kill present and discriminating (bump the pin without re-running → `no code cell's recorded output mentions 0.74.0`, and the sibling test still passes, so the new one is the only thing firing); and the notebook outputs untouched, so nothing I traced needs re-tracing.

**#967 unblocks the moment this lands** — it cites this notebook and I confirmed the figures are not on `main` yet (`20.39`, `288-423`, `525 MHz` all return 0 hits there). Your ordering note in its commit message was right.

One thing from #967's review that belongs on your plate rather than in that thread: **#714 asked for the new bracket alongside a *measured* tok/s** so over- or under-prediction is visible, and that half is genuinely unobtainable from this run — the notebook deliberately quotes none, and cell 16's trainer metrics render as an `HTML object` with no text capture. You have already stated that plainly in the record, which is the right answer. I will open the validation half as its own issue so #714 does not close on a half-met criterion.

MakazhanAlpamys:
**Re-reviewed at `812ea512`. Merge-ready — waiting on CI, nothing else.**

## The correction you made is itself correct, and I checked it rather than reading it

This is the part I refuse to take from a commit message, because the whole point of the commit is that the previous sentence was a confident false claim. Established from `git tag --contains` and the CHANGELOG:

```
#385/#387 fix   -> first tag v0.73.1
#429 fix        -> first tag v0.74.0   (merge-base --is-ancestor v0.73.3 -> NO)
```

Intersection is **v0.74.0**, which is what the cell now says. The new sentence also keeps the reason the pin is nevertheless `0.75.0` — that it is the release these committed outputs were produced against — which is the honest framing and the one that stops the next reader "fixing" the pin downward.

## The new test closes a gap that genuinely existed

`test_the_recorded_outputs_are_from_the_pinned_version` is not redundant with the six checks already there, and the proof is that the six stay green under the exact defect it catches. Mutation, your own reproduction:

```
pin 0.75.0 -> 0.74.0, outputs untouched
1 failed, 6 passed
AssertionError: no code cell's recorded output mentions 0.74.0 -- the pin was
bumped without re-running the notebook
```

Anchor `soup-cli[train]==0.75.0` matched exactly once; notebook still parsed as JSON afterwards and the guard module imported through a real `exec_module`; the failure is a behavioural assert at line 131, not a collection error. `test_the_pinned_version_is_actually_released` passes under that mutation — 0.74.0 *is* a real release — which is exactly the hole.

**What makes the guard stronger than it looks**: the install cell's recorded `soup 0.75.0` comes from `print("soup", soup_cli.__version__)`, i.e. the *installed* package, not a pip echo of the source string. So it binds the recorded run to the installed release rather than to the text of the cell.

Control mutation on the pre-existing guard (strip one cell's outputs) also dies behaviourally: `1 failed, 6 passed`, `assert not [2]`.

## Outputs are untouched

I checked this specifically, because a commit that touches a notebook whose entire value is its recorded run is the place to check it:

```
42b1899b   cells=19 code=11 code-without-output=0 outputs=41 output-json-chars=29827
812ea512   cells=19 code=11 code-without-output=0 outputs=41 output-json-chars=29827
```

Byte-identical. The commit adds a freshness requirement, it does not strip anything.

## Dropping `exit 0` is a strengthening, not a loss

The docs bullet now says the second run was in-process and has no exit code to quote. Checking the benchmark header, run **1** was also captured from printed notebook output — so the old `exit 0` was unsupported for *both* runs, not just the new one. `128 of 128` non-zero tensors is present in both the notebook's cell 17 output and `run-t4-colab-free-tier.md:79`.

## Merge state

`merge-tree` clean against `origin/main`, real merge on a scratch branch clean. On the merged tree, the four repo ratchets plus your guard file: **104 passed**, 0 failed. Ruff clean. `tests/` is `+178 -0` across the whole PR — not one assertion removed anywhere. Fragment starts `- ` with no BOM (`2d 20 52`) and credits you.

1. **A residual gap in the freshness guard, which I am naming rather than asking you to close.** It requires the pinned version in *at least one* cell's output. Since the install cell prints the installed `__version__`, a *partial* re-run — install cell re-executed, later sections left over from an older session — would still pass. The guard proves "this pin was installed in some recorded session", not "every output is from that session". Tightening it (execution-count monotonicity, or requiring the match in the install cell specifically) is more machinery than the problem is worth.
2. The fragment still describes only the repin and the guard file, not this round's attribution correction or the freshness test. Cosmetic, and neither is user-visible behaviour.

One thing I cannot verify and neither can anyone else: that the Colab session behind these outputs was a real T4. The numbers are internally consistent and T4-shaped, and the benchmark record says so in its own header — that is inherent to the artifact, not something this commit changed.

Merging as soon as the 14 checks are green on `812ea512`.

MakazhanAlpamys:
Merged as `0183ef9d`. Thank you.

`proof-4gb.ipynb` pinned an unreleased version, so anyone running the notebook got a **different tool than the one the proof was measured with** — which for a notebook that exists as evidence is not a pinning nit, it is the thing being claimed.

**You repinned it and re-ran it on a real T4**, rather than repinning and assuming the result held. That distinction is the entire value of the change, and it is the step most people skip because the diff looks the same either way.

The correction you made mid-review was itself correct, and I verified it rather than accepting it.

Credited in `CONTRIBUTORS.md` (`f491ac5b`). #844 closed with credit.

Your other PR #967 is green and reviewed but **still marked as a draft**, so the merge is refused — details on that thread. Queue note: the backlog that stalled everything yesterday is gone — 3 live runs this morning against 104 at its worst — so turnaround is short again.

PR #911, opened 2026-09-13
fix(cli): report MLX-unread CUDA-only fields and scope the all-clear

### LINKED ISSUE #903: soup doctor --config: the all-clear overclaims — CUDA-only kernels (use_liger, use_cut_ce, neftune_alpha, ...) are unread on MLX but not in the declared table
Found by the v0.75.0 post-release smoke (checklist step 6f: every message must name the actual problem).

**Repro** (v0.75.0):
```yaml
base: Qwen/Qwen2.5-0.5B-Instruct
task: sft
backend: mlx
data: {train: ./t.jsonl}
training: {epochs: 1, use_liger: true, neftune_alpha: 5}
```
```
$ soup doctor --config soup.yaml
Config check - task=sft backend=mlx
  Every setting this config writes is read on this backend.
```
Neither `use_liger` nor `neftune_alpha` appears anywhere in `trainer/mlx_sft.py`, `mlx_optim.py` or `mlx_masking.py` (grep: 0 hits), so both are accepted and dropped — the exact "validated, documented, read by nothing" pattern #755/#756 was built to report. The declared `_MLX_SFT` table in `src/soup_cli/config/backend_support.py` has seven entries (`seed`, `data_seed`, `use_galore`, `use_ring_attention`, `use_flash_attn`, `train_on_messages_with_train_field`, `train_on_prompt`) and the all-clear message in `commands/doctor.py:358` is worded as a universal claim.

Two halves:
1. **Wording (small, do first):** the all-clear must state what was checked — e.g. `None of the 7 settings known to be unread on sft/mlx is set in this config.` — not "every setting is read". A message that fires on the correct case as a stronger claim than the evidence supports is the class the release notes already warn about (known limitation 2).
2. **Table (the real work):** add the CUDA/Triton-only training fields to `_MLX_SFT`, each with a reason: `use_liger`, `use_cut_ce`, `kernel_auto_compose`, `neftune_alpha`, `use_mod`, `use_longlora`, `moe_lora`, `activation_offloading`, `quantization_aware`, `packing_cross_doc_attn_mask`, `use_fsdp2_compile` — verify each with a grep over the three MLX modules before adding, because the #756 guard fails the suite in BOTH directions (an entry for a field the trainer reads is rejected, and it is not a proof for fields left out).

**Acceptance:**
- [ ] the repro above lists `use_liger` and `neftune_alpha` as unread with a reason
- [ ] the all-clear names the number of declared gaps checked
- [ ] `tests/test_issue755_backend_support_registry.py` still passes in both directions

feiiiiii5:
## Summary

Fixes #903 — both halves:

1. **Wording:** the `soup doctor --config` all-clear no longer claims universality. It now states what was actually checked: `None of the N setting(s) known to be unread on task=<task> backend=<backend> is set in this config.`
2. **Table:** adds the two repro fields to the declared `_MLX_SFT` table — `training.use_liger` (Liger fused kernels have no MLX implementation) and `training.neftune_alpha` (NEFT noise is applied on the transformers training path, not MLX) — plus matching warning branches in `trainer/mlx_sft.py::_check_unsupported` so the trainer itself reports them.

## Motivation

The v0.75.0 smoke repro (`task=sft backend=mlx` with `use_liger: true, neftune_alpha: 5`) printed "Every setting this config writes is read on this backend." while both fields are accepted and silently dropped — the exact "validated, documented, read by nothing" pattern #755/#756 exists to report. Each added entry was verified by grep to have zero hits in the three MLX modules before adding.

## Changes

- `src/soup_cli/config/backend_support.py`: 2 new `SupportEntry`s (`IGNORED`, `trainer_reads=True`), matching the surrounding entry style.
- `src/soup_cli/trainer/mlx_sft.py`: 2 new `_check_unsupported` branches; the dotted field name is inside the message string (the #756 guard AST-matches on it).
- `src/soup_cli/commands/doctor.py`: all-clear rewritten as a dynamic count via `unsupported_for(task, backend)`.
- `tests/test_issue755_backend_support_registry.py`: 2 new tests — repro fields reported via `check_config` (fails pre-fix), and the all-clear names its evidence count (fails pre-fix on the old wording).

## Testing

- New regression `test_new_mlx_gaps_use_liger_and_neftune_alpha_are_reported`: **passes** locally.
- Registry guard suite (`test_issue755_*`, both directions): **29 passed** locally in a minimal harness (no GPU/MLX/network/model download needed).
- Doctor end-to-end tests (5 pre-existing + 1 new): **not runnable in my minimal harness** — `doctor()` exits 1 because the training stack (torch/transformers/...) is not installed there; I verified the same 5 fail identically on unmodified `main` in that harness, so this is a pre-existing environment gap, not a regression. The new all-clear line was observed rendering correctly in the captured output (`None of the 9 setting(s) known to be unread on task=sft backend=mlx is set in this config.`). Full matrix left to CI.
- `tests/test_issue755_backend_support_registry.py` bidirectional guards still pass, per the issue's acceptance criterion.

## Notes for reviewers

Scope kept to the two repro fields: each table entry must be grep-verified against the MLX modules, and unverified entries would trip the #756 guard in the wrong direction. Happy to extend the table field-by-field in follow-ups.

Signed-off-by: fei <204683769+feiiiiii5@users.noreply.github.com>

### DIFF
--- a/src/soup_cli/commands/doctor.py
+++ b/src/soup_cli/commands/doctor.py
@@ -322,7 +322,11 @@ def _check_config_support(config_path: str) -> None:
     pre-flight check, and the fields sitting at their schema default are not
     what anyone came here to ask about.
     """
-    from soup_cli.config.backend_support import DEFAULT_BACKEND, check_config
+    from soup_cli.config.backend_support import (
+        DEFAULT_BACKEND,
+        check_config,
+        unsupported_for,
+    )
 
     try:
         from soup_cli.config.loader import load_config
@@ -354,8 +358,10 @@ def _check_config_support(config_path: str) -> None:
         f"backend=[bold]{backend}[/]"
     )
     if not gaps:
+        known = unsupported_for(cfg.task, backend)
         console.print(
-            "  [green]Every setting this config writes is read on this backend.[/]"
+            f"  [green]None of the {len(known)} setting(s) known to be unread "
+            f"on task={cfg.task} backend={backend} is set in this config.[/]"
         )
         return
 
--- a/src/soup_cli/config/backend_support.py
+++ b/src/soup_cli/config/backend_support.py
@@ -118,6 +118,18 @@ _MLX_SFT: tuple[SupportEntry, ...] = (
         "MLX masks the prompt or supervises the whole sequence; no per-field switch",
         trainer_reads=True,
     ),
+    SupportEntry(
+        "training.use_liger",
+        IGNORED,
+        "Liger fused kernels have no MLX implementation",
+        trainer_reads=True,
+    ),
+    SupportEntry(
+        "training.neftune_alpha",
+        IGNORED,
+        "NEFT noise is applied on the transformers training path, not MLX",
+        trainer_reads=True,
+    ),
 )
 
 
--- a/src/soup_cli/trainer/mlx_sft.py
+++ b/src/soup_cli/trainer/mlx_sft.py
@@ -209,6 +209,15 @@ class MLXSFTTrainerWrapper:
             unsupported.append("Ring Attention")
         if tcfg.use_flash_attn:
             unsupported.append("FlashAttention (MLX has its own attention kernels)")
+        if tcfg.use_liger:
+            unsupported.append(
+                "training.use_liger (Liger fused kernels have no MLX implementation)"
+            )
+        if tcfg.neftune_alpha is not None:
+            unsupported.append(
+                "training.neftune_alpha (NEFT noise is applied on the "
+                "transformers training path, not MLX)"
+            )
         # #353's fourth criterion. #381 threaded training.seed through every
         # transformers task wrapper; MLX has its own RNG (mx.random) and reads
         # neither field, so a seeded MLX run is silently unseeded. `is not None`
--- a/tests/test_issue755_backend_support_registry.py
+++ b/tests/test_issue755_backend_support_registry.py
@@ -185,6 +185,20 @@ def test_a_setting_the_backend_ignores_is_reported(config_at):
     assert "training.seed" in reported
 
 
+def test_new_mlx_gaps_use_liger_and_neftune_alpha_are_reported(config_at):
+    """#903: accepted-but-dropped fields must be reported, not all-cleared."""
+    from soup_cli.config.backend_support import check_config
+    from soup_cli.config.loader import load_config
+
+    cfg = load_config(
+        config_at(
+            "sft", "mlx", "  use_liger: true\n              neftune_alpha: 5"
+        )
+    )
+    reported = {e.field for e in check_config(cfg)}
+    assert {"training.use_liger", "training.neftune_alpha"} <= reported
+
+
 def test_only_fields_the_user_actually_set_are_reported(config_at):
     """Not all 275 — and not the other MLX gaps the user never touched."""
     from soup_cli.config.backend_support import check_config
@@ -290,6 +304,23 @@ def test_doctor_exits_non_zero_when_the_config_cannot_be_read(
     # 2 specifically, not merely non-zero: docs/commands.md:218 documents it,
     # and all three unreadable shapes must agree.
     assert excinfo.value.exit_code == 2
+
+
+def test_all_clear_states_what_was_checked(config_at, capsys, monkeypatch):
+    """#903: the all-clear must name its evidence, not claim a universal."""
+    from rich.console import Console
+
+    import soup_cli.commands.doctor as doctor_module
+    from soup_cli.commands.doctor import doctor
+    from soup_cli.config.backend_support import unsupported_for
+
+    monkeypatch.setattr(doctor_module, "console", Console(width=200))
+    path = config_at("sft", "mlx")
+    doctor(nccl=False, disk=False, config=path)
+
+    out = strip_ansi(capsys.readouterr().out)
+    n = len(unsupported_for("sft", "mlx"))
+    assert f"None of the {n} setting(s) known to be unread" in out
 
 
 @pytest.mark.parametrize("width", [35, 60, 200])


MakazhanAlpamys:
Reviewed at `62868127`. Verified against the issue's exact repro, and this does both halves properly.

On `main` that config prints `Every setting this config writes is read on this backend.` With this PR it prints the two-row table with reasons, then `2 setting(s) written here are not read on backend=mlx.` A clean mlx config prints `None of the 9 setting(s) known to be unread on task=sft backend=mlx is set in this config.`

I confirmed your grep claim independently rather than taking it: `use_liger` and `neftune_alpha` have **0 hits** across `mlx_sft.py`, `mlx_optim.py` and `mlx_masking.py` on `main`. I also checked the schema defaults (`use_liger=False`, `neftune_alpha=None`), so the trainer branches stay off for an unset config and there is no false positive.

**Mutations, all killed:** dropping the `use_liger` entry fails 2 (two independent tests); printing the all-clear unconditionally fails 4; deleting the `neftune_alpha` trainer branch fails the bidirectional drift guard — which is the one I most wanted to see fire, because it proves the registry and the trainer cannot drift apart in either direction; hardcoding the count fails 1; restoring the old wording fails 1. 35/35 pass, and also under `FORCE_COLOR=1` — thank you for routing through `conftest.strip_ansi` instead of growing a tenth private copy.

**Blocking: there is no changelog fragment.** `changelog.d/0.75.0/` has `788, 867, 872, 882, 892, 905` and no `911`. `test_issue487_changelog_fragments.py` only checks that *existing* fragments match the newest release heading, so CI stays green without one — but the release checklist requires it and #487 exists precisely so nothing edits `CHANGELOG.md` directly. Suggested body:

```markdown
- **`soup doctor --config` no longer overclaims, and reports two more MLX gaps (#903 by @feiiiiii5 in #911).**
  The all-clear now names its evidence -- `None of the N setting(s) known to be unread on
  task=<task> backend=<backend> is set in this config.` -- instead of claiming every setting
  is read. `training.use_liger` and `training.neftune_alpha` are declared unread on
  `task=sft backend=mlx`; both were accepted and silently dropped.
```

**One design question I would like your view on before merge.** For an unreviewed pair the message is now:

```
None of the 0 setting(s) known to be unread on task=sft backend=transformers is set in this config.
```

Strictly truer than `main`, but it is still a green pass built on zero evidence — which is the class #903 is about, and `backend_support.py`'s own docstring says unreviewed pairs should "report nothing rather than guessing". Would you add a `not known` branch saying the pair has no reviewed table and nothing was checked, with a test pinning that wording? It is one branch and it closes the same hole the issue opened.

**Two small ones:**

- **`use_liger` is not purely "ignored".** `commands/train.py:1193` calls `validate_liger_config`, which errors whenever `device != "cuda"` and exits 1 — so on a real Apple Silicon box `soup train` refuses the run before MLX ever sees the field. `neftune_alpha` is the pure case; `use_liger` is not. `IGNORED` is still defensible under this module's definitions (MLX genuinely never reads it, and `doctor --config` does not run that validation), but a clause in the reason string would make the report match what the user will actually experience.
- The `"  use_liger: true\n              neftune_alpha: 5"` indentation **works** — I reproduced the fixture and parsed it, `model_fields_set == ['neftune_alpha', 'use_liger']` — but the 14 spaces only survive because of the fixture's own literal indentation plus `dedent`. Any future change there breaks it silently into invalid YAML rather than failing loudly. `"  use_liger: true\n  neftune_alpha: 5"` works too and cannot break that way.

**On scope, and this is the part worth deciding rather than drifting into.** The issue names 11 fields; this adds 2. I checked the other nine myself — `use_cut_ce`, `kernel_auto_compose`, `use_mod`, `use_longlora`, `moe_lora`, `activation_offloading`, `quantization_aware`, `packing_cross_doc_attn_mask`, `use_fsdp2_compile` — all real `TrainingConfig` fields, all 0 hits in the three MLX modules, so all nine would pass the bidirectional guard today. The narrowing is a choice, not a limit, and I am happy to take the two-field version now. But **the all-clear now advertises a count**, and a count is only as honest as the table is complete — "None of the 9" reads stronger than 9-of-20 warrants. Please open the follow-up for the remaining nine so the number means what it sounds like.

No conflicts with the other five open PRs touching `commands/doctor.py` (#660, #877, #883, #884, #888) — all in different functions, `merge-tree` clean against each. CI on your head was still partly pending when I reviewed; I will merge once it is green.

MakazhanAlpamys:
Merged as `7c68bd85`. Thank you — and the half worth naming is the wording, not the table.

Adding two entries to `_MLX_SFT` is the obvious patch and it would have closed the repro. It would also have left the command saying `Every setting this config writes is read on this backend.` the next time someone set the tenth unread field. You fixed the sentence instead, so the guarantee can no longer be read wider than the evidence behind it — that is the actual defect #903 describes, and it is the part most patches would have skipped.

Two more things I want on the record:

- **You grep-checked each entry for zero MLX hits before declaring it unread.** A declared table is only worth anything if entries are added on evidence rather than on plausibility, and that is the discipline that keeps it worth anything.
- **You declared the narrowing instead of hiding it.** 2 of the issue's 11 fields, said plainly in the PR body. I checked the other nine myself and they would all pass the same bidirectional guard today, so the choice was defensible — but it was reviewable only because you stated it.

What I ran before merging:

- The issue's exact repro in both directions. On `main` that config prints the all-clear; on your head it prints the two-row table then `2 setting(s) written here are not read on backend=mlx.` A clean mlx config prints the scoped `None of the 9 ...` line.
- `merge-tree` against each of the five other open PRs touching `commands/doctor.py` (#660, #877, #883, #884, #888) — all clean, all in different functions.
- Full required matrix on `62868127`: 15/15 green.

I have opened #958 for the remaining nine fields so the advertised count means what it sounds like, and credited it to this PR. The two smaller notes from my review — the `use_liger` reason string, and the fixture's 14-space indentation surviving only by `dedent` — I am not holding the merge for; both are recorded there as notes.

Your entry is in `CONTRIBUTORS.md` at `40a97fa7`. Welcome.

PR #914, opened 2026-09-13
fix(eval): connect against gate to stored eval results

### LINKED ISSUE #812: soup eval against cannot read any score Soup's own eval commands write, and the pre-push hook reports that as "regression" (candidate improved 0.80 -> 0.90)
## The defect

`soup eval against` compares two runs from the experiment tracker, and `soup eval gate-install` wires it into a pre-push hook. The two ends were never connected: the metric names `against` accepts and the benchmark names Soup's eval commands write are disjoint sets.

**What `against` accepts.** `--metric` defaults to `task_accuracy`; `decide_regression` accepts only `task_accuracy`, `refusal_rate`, `format_validity`, `p95_latency_ms` (`utils/eval_gate_hook.py:221-253`).

**Where it looks.** `ExperimentTracker.get_metric_series` (`experiment/tracker.py:484-520`) reads the per-step `metrics` table, and when that is empty falls back to `eval_results` rows whose `benchmark` column equals the metric name (the #164 bridge).

## Evidence

**What Soup writes to `eval_results`.** Every `save_eval_result` call in the codebase:

| writer | `benchmark` value |

|---|---|

| `commands/eval.py:283` (aider polyglot) | `aider_polyglot` |

| `commands/eval.py:589` (`eval judge`) | `judge:<judge_model>` |

| `commands/eval.py:1036` (`eval benchmark`) | the lm-eval task name, e.g. `mmlu` |

| `commands/eval.py:1082` (`eval custom`) | `custom` |

None of them is ever `task_accuracy` or any other allowed name, so the #164 fallback cannot fire for data Soup produced. The #164 issue text assumed "the `eval_results` table where `task_accuracy` / `refusal_rate` / etc. actually live"; no code writes those names.

**How the hook reads the result.** `utils/eval_gate_hook.py:299-304` maps ANY non-zero exit to a regression:

```bash

soup eval against "$BASELINE_RUN_ID" --candidate "$CANDIDATE_RUN_ID" \

    --suite "$GATE_SUITE" --json-only \

    || {

        echo "[soup] pre-push gate blocked: regression vs $BASELINE_RUN_ID" >&2

        exit 1

    }

```

`against` exits 1 for an empty series (`commands/_eval_v0550.py:324-329`), for a regression (363), and for bad input (287, 295, 322, 342), so "no data" and "regressed" are the same signal.

## Reproduction

Isolated DB (`SOUP_DB_PATH`), two runs, eval results written through the real tracker API exactly as `soup eval custom --run-id` and `soup eval benchmark --run-id` write them. The candidate is better on both:

```python

t = ExperimentTracker()

b = t.start_run({"base": "base-model", "task": "sft"}, "cpu", "cpu", {}, run_id="run_base")

c = t.start_run({"base": "base-model", "task": "sft"}, "cpu", "cpu", {}, run_id="run_cand")

t.save_eval_result("out/base", "custom", 0.80, {"total": 50, "correct": 40, "category_scores": {}}, run_id=b)

t.save_eval_result("out/cand", "custom", 0.90, {"total": 50, "correct": 45, "category_scores": {}}, run_id=c)

t.save_eval_result("out/base", "mmlu", 0.61, {}, run_id=b)

t.save_eval_result("out/cand", "mmlu", 0.64, {}, run_id=c)

```

```

$ soup eval against run_base --candidate run_cand --suite evals/locked.json

Empty series — baseline=0 candidate=0.

exit=1



$ soup eval against run_base --candidate run_cand --metric custom

Regression check failed: unknown metric 'custom'; allowed: format_validity, p95_latency_ms, refusal_rate, task_accuracy

exit=1



$ soup eval gate-install --baseline run_base --suite evals/locked.json

Installed pre-push gate → .git/hooks/pre-push

$ SOUP_CANDIDATE_RUN_ID=run_cand bash .git/hooks/pre-push

Empty series — baseline=0 candidate=0.

[soup] pre-push gate blocked: regression vs run_base

hook exit=1

```

A typo is indistinguishable too: `--metric accuracy` prints `Empty series — baseline=0 candidate=0.` and exits 1, because the series lookup runs before the metric name is validated.

## Why it matters

The gate can only be satisfied by hand-inserting rows named `task_accuracy`. Any team that installs it and uses Soup's own eval commands gets every push blocked with the word "regression" — for a model that got better. The only way through is `git push --no-verify`, which disables the gate entirely.

A related weakness: `eval_results` holds ONE aggregate score per eval invocation, so even with matching names the "paired-bootstrap CI" runs over n=1. With a hand-written `task_accuracy` row per run the output is `ci=[-0.0100, -0.0100]` — a point, not an interval.

## Proposed fix

1. Accept the benchmark names Soup writes: allow `--metric custom`, `mmlu`, `judge:<m>`, etc. with a declared direction (higher-is-better), or map `task_accuracy` to `custom` explicitly and document it.

2. Validate `--metric` before querying, so a typo is a usage error rather than an empty series.

3. Give "no data" its own exit code and teach the hook to say "gate could not run: no eval results for run X" instead of "regression". Whether that blocks the push is a policy choice; lying about the reason is not.

4. When the series has fewer than two points, say the CI is degenerate, or compare against per-item scores (the eval commands already have them in `details`).

## Acceptance criteria

- [ ] A test that writes `custom` results for two runs via `save_eval_result`, runs `soup eval against` with a documented metric, and asserts a verdict (not "Empty series"). It FAILS on current code.

- [ ] `--metric accuracy` (unknown) exits with a usage error naming the allowed metrics, before touching the DB.

- [ ] The generated hook distinguishes "regression" from "no data" in its message; a test runs the rendered hook against an empty series and asserts it does not print "regression".

---

Amix29:
## Summary
- resolve Soup-produced custom, Aider Polyglot, judge, and namespaced lm-eval results before comparison
- validate and bound metric names before opening the experiment database, including rejecting reserved training columns in the benchmark namespace
- report unavailable comparisons honestly while leaving exit-code taxonomy to #921
- mark one-sample aggregate comparisons as having no confidence interval

## Tests
- 160 focused and compatibility tests passed
- full suite: 21,280 passed, 106 skipped, 1 unrelated existing Rich-width control failure in tests/test_issue204_langfuse_pull.py; coverage 83.41%

Closes #812


MakazhanAlpamys:
Reviewed at `6cbda9ad`. **The data-resolution half of this is exactly right and I could not break it** — seven mutations, zero survivors. I am requesting changes on sequencing, for one reason, plus one real bug.

**What I confirmed works.** I reproduced #812's three scenarios on `main` and on your branch against the same database. On `main`: default gives `Empty series ... exit=1`, `--metric custom` and `--metric judge:gpt-4o` give `unknown metric`, and the hook reports `regression` for a candidate that improved 0.80 to 0.90. On your branch all of those resolve and the hook exits 0. I also generated the rows with the **real command** rather than a fixture — `soup eval custom --tasks ev.jsonl --model <cached gpt2> --run-id ...` twice, writing `('run_base','custom',...)` / `('run_cand','custom',...)` — and `against` with no `--metric` picked them up as `"source_metric": "custom"`. That is the end-to-end path the issue says never existed.

**Mutations: 7 applied, 0 survivors.** Dropping the `custom` fallback, accepting any metric name, reverting the unavailable exit, disabling the `judge:`/`benchmark:` namespaces, forcing `ci_degenerate = False`, moving validation after `ExperimentTracker()`, and reverting the invalid-suite exit each fail a test. The single existing assertion you changed in `test_code_review_high.py` still pins "blocks", so it is a re-specification rather than a weakening — I checked that specifically.

**The blocker is #921, and neither of you can fix it alone.** @kok-o's open PR for #813 rewrites the same two files and the same `set +e / $? / set -e` hook block with an **incompatible** taxonomy:

| branch | this PR | #921 |
| --- | --- | --- |
| regression | 1 | 2 |
| invalid `--suite` | 2 | 3 |
| empty / unavailable | 3 ("unavailable") | 3 ("usage error" — same code, different meaning) |
| `AttributeError` | 2 | 1 |

Whichever lands second has to re-decide what all three codes mean for `soup eval against`. And #921's numbering is the one matching what the project already documents for `soup ship` — `0=SHIP / 2=DON'T / 3=usage / 1=runtime`, with usage deliberately moved **off** 2 in v0.71.38. This PR puts usage on 2 and keeps regression on 1.

**My suggestion, which unblocks you now:** split the exit-code renumbering out and keep the part that fixes #812 — `resolve_gate_metric`, the `custom` fallback, the `judge:`/`benchmark:` namespaces, the degenerate-CI reporting, and the honest "comparison unavailable" *message*, with the distinct exit code left to #921. That reduces the collision to a docs conflict and makes this mergeable without waiting on another author.

**One real bug, small but in the issue's own failure class.** `benchmark:` forwards its suffix straight into `get_metric_series`, which reads the per-step `metrics` table **before** falling back to `eval_results`. So `--metric benchmark:loss` reads training loss and treats it as higher-is-better. Measured: baseline loss 1.0/0.9/0.8 against candidate 2.0/1.9/1.8 — a strictly worse model —

```
OK  benchmark:loss  delta_mean=+1.0000  ci=[+1.0000,+1.0000]  exit=0
```

On `main` `--metric loss` was rejected as unknown, so this is new. A gate that passes a worse model is exactly what #812 is about. Rejecting the reserved columns (`loss`, `val_loss`, `lr`, `grad_norm`, `speed`, `gpu_mem`, `epoch`) in `resolve_gate_metric`, or documenting `benchmark:` as `eval_results`-only, closes it.

**Test hygiene, not a blocker but measured in both directions.** Under `FORCE_COLOR=1 TERM=xterm-256color` six of the new tests fail: five because `json.loads(result.output)` meets Rich's `ReprHighlighter` colouring the `--json-only` payload, and one on `assert "benchmark:<name>" in result.output` because the highlighter bolds the angle brackets (`benchmark:<name\x1b[1m>\x1b[0m`). I checked whether this is live — no workflow sets `FORCE_COLOR`, and under a simulated `GITHUB_ACTIONS=true CI=true` env all 7 pass — so **latent, not live**. But it is the class that has reddened this repo's CI four times, and `benchmark:<name>` is a live counterexample to the assumption in `test_cli_help_assertions_are_ansi_safe.py`'s own docstring that flags inside error messages are never styled per-token. Please route both through the strip helper in `tests/test_auto_tuning.py`.

**Smaller:** `"aider_polyglot"` and `"custom"` are re-typed as literals rather than imported from `aider_polyglot.TASK_NAME` and the writer at `commands/eval.py:1084` — a rename would drift the gate silently, which is #812 recurring; an unknown metric name is echoed at unbounded length (the namespaced branch caps at 256, the generic one does not — control bytes are safe, I checked all four cases); and the fragment could use `(#812 by @Amix29 in #914)`.

7 passed + 1 skipped in the new file, 188 passed / 7 skipped across the five files touching the changed modules, ruff clean. The docs match the code by execution on every claim — subject to the taxonomy question above, which would invalidate that paragraph.

Amix29:
Addressed the review at `d72c7e0`:

- removed this PR’s competing exit-code taxonomy; existing non-zero semantics remain, and #921 can define the shared taxonomy
- kept the unavailable-data diagnostic honest and made the hook failure label generic rather than falsely calling every failure a regression
- rejected `benchmark:loss`, `val_loss`, `lr`, `grad_norm`, `speed`, `gpu_mem`, and `epoch` before tracker access
- bounded all metric names at 256 characters before echoing errors
- reused the canonical custom/Aider task constants, including the custom-result writer
- routed output/JSON assertions through the shared ANSI-strip helper and verified the suite under `FORCE_COLOR=1`
- fixed changelog attribution

Validation: `196 passed` across the affected compatibility suites; `16 passed` under forced Rich color; full repository Ruff scope and `git diff --check` pass.

MakazhanAlpamys:
Re-reviewed at `917bad7a`. (Heads-up: the SHA in your summary, `d72c7e0`, does not resolve in the repo — `git cat-file -t` cannot find it. I reviewed the actual head.)

**The previous finding is closed, and I verified it by execution on both heads** with an isolated `SOUP_DB_PATH` — baseline loss `1.0/0.9/0.8`, candidate `2.0/1.9/1.8`, i.e. a strictly worse model:

```
6cbda9ad  exit 0   {"metric":"benchmark:loss","regressed":false,"delta_mean":1.0,...}
917bad7a  exit 1   Invalid --metric: benchmark:loss is a reserved training metric
```

**The refusal really is ahead of the database**, proven two ways: a counting spy on `ExperimentTracker.__init__` records **zero** constructions on that path, and a mutation that injects a tracker construction as the first statement of `against_cmd` takes 9 tests red.

All seven reserved names refuse. I probed the obvious evasions — `benchmark:LOSS`, `benchmark: loss`, `benchmark:loss ` — and they clear the reserved check but then resolve to an empty series and exit 1, so there is no bypass. `judge:loss` is still accepted, correctly: different namespace, different table key. The 256-char bound refuses at 257 and the message does not echo the input.

Five mutations, no survivors: dropping the reserved guard (7 failed), dropping the bound, dropping the `custom` fallback, relabelling the unavailable diagnostic as a regression, and opening the DB first. Every assertion routes through `strip_ansi` and the file is green under `FORCE_COLOR=1`. The rendered pre-push hook passes `bash -n`, and its failure branch now says `comparison failed vs $BASELINE_RUN_ID` rather than calling every failure a regression.

**Splitting out the exit-code taxonomy was the right call and I want to be explicit that it was your judgement, not my instruction** — I offered it as one option among several. Reusing `eval/custom.py::TASK_NAME` and `aider_polyglot.TASK_NAME` rather than retyping the literals is the other thing worth naming: the writer at `_save_custom_results` and the reader can no longer drift, which is #812 recurring in miniature.

**Two small things, neither blocking:**

1. `_eval_v0550.py:332` still raises `typer.Exit(2)` on the legacy-tracker `AttributeError` branch, while the docstring two screens above now says the command exits `0` when there is no regression and `1` otherwise. Since the point of this round was to leave the taxonomy to #813/#921, either make it `1` or add "(the legacy-tracker advisory exits 2)" to the docstring.
2. Neither `docs/evaluation.md` nor `--metric --help` mentions that training column names are reserved, so a user who types `benchmark:loss` meets the refusal with no prior warning. One clause — "training columns (`loss`, `val_loss`, `lr`, ...) are not valid benchmark names" — would cover it.

**Ordering, mine to manage:** this and #921 both touch `_eval_v0550.py` and `eval_gate_hook.py`. Yours is now the narrower change, so it lands first and #921 re-derives against it — I will tell @kok-o that their diff needs re-deriving rather than re-applying, since your branches are collapsed where theirs expects the old shape.

MakazhanAlpamys:
Merged as `f77fdbe5`. Thank you — reading the stored eval results rather than recomputing is what turns the gate into a comparison instead of a second measurement, and a second measurement is exactly how a gate ends up disagreeing with the numbers already on record for the same model.

Verified before merging: the regression fails on `origin/main` for the real reason rather than a missing symbol, and reverting the lookup fails a behavioural assertion.

**Ordering, which was mine to manage:** this landed first because it is the narrower change, so #921 now needs re-deriving against it rather than re-applying — both touch `_eval_v0550.py` and `eval_gate_hook.py`, and your branches are collapsed where that PR expects the old shape. I have told @kok-o.

Your CONTRIBUTORS.md entry is in and #812 is closed.

PR #915, opened 2026-09-13
fix(ui): cap YAML request bodies

### LINKED ISSUE #897: Web UI YAML-accepting endpoints have no request-size cap
`ui/app.py` `/api/config/validate`, `/api/train/start` and `/api/config/from-form` accept an arbitrary-length YAML string (`TrainRequest.config_yaml: str` with no `max_length`, no body-size middleware). v0.75.0 bounded the unknown-key WALK (it stops at 100 findings; 50,000 bogus keys measured ~31 s of CPU before), but YAML parsing and Pydantic validation still scale with the body. Authenticated only (Bearer), so this is a token holder stalling a shared `--public` deployment, not an anonymous DoS. The MCP side already caps reads at `_MAX_JSON_BYTES = 16 MiB`. Found in the v0.75.0 security review.

**Fix path:** a `max_length` on the request models (1 MiB is generous — the largest shipped recipe is under 4 KiB) and/or a body-size middleware returning 413 before parsing.

**Acceptance:**
- [ ] a 2 MiB YAML body is refused with 413 before `yaml.safe_load` runs
- [ ] every shipped recipe still validates through `/api/config/validate`

Amix29:
Closes #897

## Summary
- reject request bodies over 1 MiB on the three Web UI YAML-entry endpoints with HTTP 413
- check both Content-Length and the actual ASGI body bytes before FastAPI JSON or Soup YAML parsing
- keep all 169 shipped recipes valid through /api/config/validate
- document the Web UI request-size contract

## Tests
- .venv/bin/ruff check src/soup_cli tests (pass)
- focused issue tests: 173 passed
- Web UI compatibility tests: 425 passed
- FastAPI 0.104.0 compatibility probe: oversized request returned 413
- full suite: 21,445 passed, 106 skipped, 5 deselected, 1 failed; coverage 83.41 percent

The sole full-suite failure is tests/test_issue204_langfuse_pull.py::test_control_the_leak_scan_sees_a_token_rich_highlighted_and_folded. The current Rich renderer uses 80 columns despite Console(width=24), so neither planted token wraps. The same isolated failure was reproduced while qualifying #812; it is unrelated to the Web UI files changed here.


MakazhanAlpamys:
Reviewed at `20b45a3d`. Approving — this closes #897 properly and the guard is enforced where it has to be.

**I re-enumerated the YAML endpoints from `ui/app.py` rather than trusting the issue.** Of the 7 POST routes, exactly three reach `load_config_from_string` (`/api/config/validate`, `/api/train/start`, `/api/config/from-form`); the other four take no YAML. Your set is exactly those three.

**Measured, not inferred:**

- The issue's named attack — 50,000 bogus keys under one section, 1,038,966 B of YAML — is refused in **0.03 s**, against the ~31 s of CPU the issue documents.
- Largest admitted body 1,048,568 B costs **3.01 s**; one step over returns 413 in **0.01 s**. `received > max` admits exactly 1 MiB, which is the intended reading.
- The middleware sits **inside** CORS (`[CORSMiddleware, _YamlRequestBodyLimitMiddleware]`), so a 413 still carries `access-control-allow-origin`.
- I probed an 8-level x 30-alias YAML bomb at 911 bytes specifically, because a byte cap does **not** bound alias expansion: 200 in 0.02 s. pyyaml shares aliases by reference and #627's 100-finding cap stops the walk, so there is no expansion hole behind the cap. Worth stating since it was the first thing I looked for.

**Mutations, all killed:** cap to infinity fails 3; **moving the check after the downstream app fails 4** — that is the one that matters, since a cap applied after `yaml.safe_load` would be decorative; removing each of the three endpoints from the allowlist fails 1/2/1 respectively; deleting the byte-count check fails 1. 173 passed in the new file, 332 across the existing UI and code-review suites, ruff clean, and all **169** shipped recipes still validate through `/api/config/validate`.

**One mutation survived, and it is not a correctness problem but it is a documentation problem.** Deleting the `Content-Length` fast path entirely leaves all 173 tests green. The byte loop still enforces the cap, so the worst case is one extra 1 MiB read — but the changelog and docs both claim the server "checks both `Content-Length` and the bytes actually received", and nothing pins that. An inverted comparison there *would* be caught by the 169-recipe test; silent deletion would not.

Two small tests close it, and they are cheap:

```python
# 1. Content-Length-only rejection: receive() must never be called
async def _receive():
    raise AssertionError("body was read despite an oversized Content-Length")
# scope with ("content-length", b"2097152") -> 413

# 2. the exact boundary, so a future > -> >= flip is visible
#    1 MiB body -> 200 ; 1 MiB + 1 -> 413
```

**Three notes, none blocking:**

- `_YAML_REQUEST_PATHS` is a hand-maintained allowlist with no guard, so a future YAML endpoint gets no cap silently. A ratchet asserting that every route whose handler reaches `load_config_from_string` is in the set would keep this closed.
- **Residual cost is 3.01 s of CPU per authenticated request at the cap.** Far better than ~31 s and within the issue's acceptance, but worth saying in the changelog rather than implying the class is fully closed.
- An unauthenticated oversized POST gets 413 rather than 401, because the limiter runs outside the route's `Depends(_verify_token)`. That is the right order for DoS — it also bounds FastAPI's pre-existing full-body read for anonymous callers — and it only confirms path existence the SPA already names. Deliberate, not a finding, but you should know it is the behaviour.

**Out of scope but adjacent, and I will file it rather than ask you to widen this PR:** `/api/chat/send` accepts a 3 MiB body and returns 200, forwarding it upstream, and `/api/data/inspect` reads a full 3 MiB body before rejecting the path with 400. Neither is YAML so neither is #897.

The fragment needs `by @Amix29`, and the diff drops the trailing blank line of `docs/serving-and-export.md` — unrelated whitespace.

Push the two tests and I will merge; if you would rather I take them at merge with the work credited to you, say so.

Amix29:
Addressed the review at `2f76231`:

- added a Content-Length fast-path test whose `receive()` raises if the body is read
- pinned the exact 1 MiB boundary (`limit` accepted, `limit + 1` rejected)
- added `by @Amix29` to the changelog fragment
- preserved the documentation trailing newline

Validation after the changes: 193 targeted tests passed; full Ruff scope passed. Rebased onto current `origin/main`.

MakazhanAlpamys:
Verified at `2f76231`. Both tests kill their mutations — that is the whole of it.

```
remove the Content-Length fast path entirely
-> FAILED test_oversized_content_length_rejects_without_reading_body    (1 failed, 175 passed)

flip `received > max` to `received >= max`
-> FAILED test_body_cap_exact_boundary[12345-204]                       (1 failed, 175 passed)
```

The first is the one I asked for and it is written the right way: a `receive()` that raises if called, so the test proves the body was **not read**, rather than merely that a 413 came back. The boundary test now pins the comparison in both directions, so a future `>` to `>=` flip is visible instead of silently shifting the accepted size by one byte.

Fragment carries `by @Amix29`, and the docs trailing newline is restored.

Merging as soon as the checks report — they are still queued on this head.

The three notes I left as non-blocking stand as non-blocking, and I am not asking for them here: the hand-maintained `_YAML_REQUEST_PATHS` allowlist has no ratchet, the residual cost at the cap is 3.01 s of CPU per authenticated request, and an unauthenticated oversized POST gets 413 rather than 401 (correct ordering for DoS, but worth knowing). I have filed the `/api/chat/send` and `/api/data/inspect` body-size gap separately — neither is YAML, so neither was #897's.

MakazhanAlpamys:
Cross-reference and an ordering decision, so you hear it from me rather than from a conflict.

**#948 (`fix(ui): cap /api/chat/send and /api/data/inspect request bodies`) implements the same mechanism as this PR**, and the two conflict:

```
#915 class: 69 lines   #948 class: 69 lines   identical lines: 54 (78%)
git merge-tree pr915rv pr948rv -> CONFLICT in src/soup_cli/ui/app.py
```

`_reject`, the `content-length` header loop including its comment byte-for-byte, the buffering loop and `receive_buffered` are the same in both. The only structural difference is that #948 looks the limit up in a per-route table instead of testing membership of a single path set with one scalar limit.

**My call: #948's table becomes the single limiter.** That is not a quality judgement between the two — yours came first and #948 is visibly derived from it — it is that a table is the strict generalisation, and #939's own acceptance criteria say so: *"Reuse `_YamlRequestBodyLimitMiddleware`'s mechanism rather than adding a second, differently-shaped limiter. If the limits should differ per route, make that a table rather than a second code path."*

Concretely, what I would like to land is one `_RequestBodySizeLimitMiddleware` whose `_BODY_SIZE_LIMITS` carries all five routes: your three YAML paths (`/api/config/validate`, `/api/train/start`, `/api/config/from-form`) at 1 MiB, plus `/api/chat/send` at 1 MiB and `/api/data/inspect` at 8 KiB — with `_YamlRequestBodyLimitMiddleware` deleted.

**Whichever of us does that rework, the three YAML routes must stay capped.** I measured on #948's branch: with only its middleware installed, all three of your routes accept a 3 MiB body. So merging #948 alone would leave 3 of the 5 body-taking routes uncapped — your PR is not redundant, it is the other half of the table.

You have two options and I am fine with either:

1. **You fold first.** Rework this PR into the table shape (or just add the per-route dict to your existing class), covering all five routes. Then #948 becomes a no-op rework on their side and I merge yours.
2. **They land first.** I merge #948, and you rebase onto it by adding three entries to `_BODY_SIZE_LIMITS` and deleting your class — a much smaller diff than what you have now.

Say which you prefer and I will hold the other one. If neither of you claims it within the day I will pick option 2, since #948's shape is the one that survives either way.

One thing that is on neither of you: #939 also asked for a ratchet asserting that every POST route reaching a body-consuming handler appears in the table. Nobody has attempted it, and with two hand-maintained tables it matters more, not less. I will file it as a follow-up rather than push it into either PR.

Amix29:
Option 2 works for me. Please land #948 first; once it is on `main`, I will rebase #915 onto it, add the three YAML routes to the shared `_BODY_SIZE_LIMITS`, remove the now-duplicate YAML-specific middleware, and preserve the existing Content-Length/boundary coverage for those routes.

MakazhanAlpamys:
Agreed, option 2 it is. Thank you for taking that without argument — you had the mechanism first and the rework lands on you, which is the wrong way round on merit and the right way round on diff size.

So the sequence is: I merge #948 once its CI is green and the two items I raised there are settled, then you rebase this onto it — three entries added to `_BODY_SIZE_LIMITS` and `_YamlRequestBodyLimitMiddleware` deleted.

Two things to carry into that rebase:

**Keep your limits, do not inherit theirs by accident.** #948's table is `{/api/chat/send: 1 MiB, /api/data/inspect: 8 KiB}`. Your three YAML routes want 1 MiB, not 8 KiB — a training config that large is unusual but a `soup.yaml` with a long `unfrozen_parameters` list is not absurd, and 8 KiB would reject it with a 413 that reads like a bug.

**The boundary test is yours to keep too.** #948 pins `n == limit -> 422` and `n == limit + 1 -> 413`; if your three routes land in the same table with a different limit, they need their own boundary cases, otherwise the table has entries nothing exercises.

I will file the route-coverage ratchet (#939's "every POST route reaching a body-consuming handler appears in the table") as a separate issue once the consolidation lands, since it only makes sense against one table. If you want it, say so and it is yours.

MakazhanAlpamys:
#948 merged (`7e90de6b`), so this is unblocked — and it now conflicts, as expected:

```
CONFLICT (content): Merge conflict in src/soup_cli/ui/app.py
```

Just the one file. What the rebase needs, per what we agreed:

- add the three YAML routes to `_BODY_SIZE_LIMITS`;
- delete `_YamlRequestBodyLimitMiddleware` entirely — the streamed-byte path in #948 already covers the chunked and lying-`Content-Length` cases your middleware was carrying separately;
- keep `tests/test_issue897_ui_yaml_body_cap.py`. Those tests are the point of this PR and they should pass unchanged against the shared mechanism. If any of them needs its expectation loosened to go green, stop and tell me rather than adjusting it — that would mean the shared cap does not actually cover a case yours did, which is a finding, not a test problem.

Worth stealing from #948 while you are in there: its before-the-parse proof is a *pair* — the same malformed JSON at 64 bytes and oversized, asserting 422 and 413. If your three routes get the same pair, the YAML cap's ordering is demonstrated rather than assumed, the same way the JSON one now is.

Your current 14 green checks are against the pre-#948 base, so they will need re-running regardless. No rush from my side; ping me when it is pushed.

Amix29:
Update pushed in 010de78 and rebased after #948. The three YAML endpoints now use the shared route-specific BODY_SIZE_LIMITS table, the superseded YAML-only middleware and route set are removed, and registration stays on the shared request-body middleware. Existing boundary, Content-Length, streamed-body, and catalog-recipe expectations are preserved. I also added a before-parse proof for all three routes: malformed small JSON reaches parsing and returns 422, while the same oversized request is rejected first with 413. Validation: 252 targeted tests passed across #897, the shared #939 cap, the full UI suite, and changelog validation; Ruff is clean on touched Python files. GitHub reports the branch mergeable. Ready for re-review.

MakazhanAlpamys:
@Amix29 — reviewed `010de781`. **Both things I asked for are genuinely done**, and one item remains: the three new table rows are pinned from above but not from below. It is one parametrized test and I have proved it kills the mutation.

## The fold is real, not cosmetic

`_YamlRequestBodyLimitMiddleware` and `_YAML_REQUEST_PATHS` are gone — `git grep` at your head returns zero hits — and `app.py` went from **85 lines of new mechanism to 14**, because the YAML routes are now three entries in the table #948 already owns:

```python
_MAX_YAML_REQUEST_BYTES = 1024 * 1024
_BODY_SIZE_LIMITS = {
    "/api/chat/send": _MAX_CHAT_SEND_BODY_BYTES,
    "/api/data/inspect": _MAX_DATA_INSPECT_BODY_BYTES,
    "/api/config/validate": _MAX_YAML_REQUEST_BYTES,
    "/api/train/start": _MAX_YAML_REQUEST_BYTES,
    "/api/config/from-form": _MAX_YAML_REQUEST_BYTES,
}
```

You also rewrote the docstring paragraph that referenced the old class rather than leaving it behind, which is the part that usually rots.

**No test was weakened.** Against the base there are zero removed test lines; against the previously-verified head every removal is the class rename. The `(b"12345", 204), (b"123456", 413)` boundary, the `receive()`-raises fast path and `assert called is False` are all intact.

The security properties hold under measurement rather than argument: the Content-Length fast path rejects with **zero** `receive()` calls and the streamed path rejects the moment `received > max_body_size`, so memory is bounded at cap + one chunk, entirely above `await self._app(...)`. Coverage was re-enumerated rather than trusted — `load_config_from_string` appears at exactly three call sites, all three are in the table, and of the seven POST routes the remaining four are either capped by #948 or declare no body. Every body-taking POST route now sits in one table, which is the whole argument for the fold.

## The one item: nothing detects the routes being capped *lower* than documented

`ui/app.py:31`, anchor counted exactly once, module genuinely re-imported afterwards:

```
_MAX_YAML_REQUEST_BYTES = 1024 * 1024   ->   8 * 1024
306 passed
```

The entire UI suite is green with your three routes silently dropped to 8 KiB. That is the accident I flagged in the option-2 handoff — *"keep your limits, do not inherit theirs by accident; 8 KiB would reject it with a 413 that reads like a bug"* — and it is still reachable, because every oversized fixture in the file is ≥ 1 MiB and the boundary test builds its own middleware with `limits={"/api/config/validate": 5}`, so it pins the `>` operator rather than the configured ceiling.

Measured both directions, three routes each, with a body under 1 MiB and far over 8 KiB that is deliberately unparseable:

| | `/api/config/validate` | `/api/train/start` | `/api/config/from-form` |
|---|---|---|---|
| real code (1 MiB) | **422** reached parser | **422** | **422** |
| mutated (8 KiB) | **413** capped below ceiling | **413** | **413** |

422 is the correct answer: the body got through to the parser and failed there. 413 means the route silently enforces a limit its documentation does not claim.

Three parametrized cases close it:

```python
@pytest.mark.parametrize(
    "endpoint",
    ["/api/config/validate", "/api/train/start", "/api/config/from-form"],
)
def test_yaml_routes_keep_the_1mib_ceiling_not_the_8kib_one(ui_client, endpoint):
    client, auth = ui_client
    body = b'{"unterminated":"' + (b"x" * (900 * 1024))   # under 1 MiB, far over 8 KiB
    r = client.post(endpoint, content=body,
                    headers={**auth, "Content-Type": "application/json"})
    assert r.status_code == 422
```

Unmutated it passes; with the mutation re-applied it fails `assert 413 == 422` on all three. This is the same request I made when the table was #948's alone — *"they need their own boundary cases, otherwise the table has entries nothing exercises"* — so it is the standing bar rather than a new one.

## Two corrections I owe you, both in your favour

**The trailing newline: we were both wrong, and your form is the right one.** I said the docs trailing newline was restored and you said you had preserved it; in fact the earlier blob already ended `).\n` where its base ended `).\n\n`, and your head still does. But `end-of-file-fixer` is an active pre-commit hook and normalises `).\n\n` to exactly your `).\n`, and every other `docs/*.md` ends with a single newline. **Keep it — I withdraw the objection.**

**`from typing import Mapping`** is #948's line, not yours. Not your finding.

## Nit, not blocking

The fragment says "all 169 shipped recipes remain valid". The live count on the merged tree is **171** — your number was right when I measured it for you and `main` has since added two. The test is derived from `RECIPES.items()`, so only the published prose is stale.

---

`git merge-tree` against current `main` is clean. 1084 passed, 15 skipped across the 26 files touching `ui/app.py` on **both** the branch and the merged tree; 1105 with the fragment validator; `ruff check` clean. Fragment named after the PR with house-style credit.

CI on this head has not run yet, so it needs another push regardless — which makes the test above cheap to fold in. Push it and I will merge on the next green matrix without re-reviewing. Queue note: the Actions backlog is cleared (104 live runs down to ~57), so turnaround is much shorter than it was earlier today.

Amix29:
Addressed the final review item and pushed 975ddd3 after rebasing on current main at 4d98b6a. Added the parametrized 900 KiB malformed-JSON test for all three YAML routes; each reaches parsing and returns 422 under the documented 1 MiB ceiling. Mutation proof: changing _MAX_YAML_REQUEST_BYTES to 8 * 1024 makes exactly those three cases fail with 413 != 422; restoring the constant returns them green. I also updated the non-blocking changelog count from 169 to the current 171, confirmed from len(RECIPES). Validation: 255 passed across #897, the shared #939 body cap, the full UI suite, and changelog validation; full Ruff is clean. Ready for the next green matrix.

MakazhanAlpamys:
Merged as `ca06b2c0`. Thank you — that is four of yours in two days, and this one closed a limitation I had written into the v0.75.0 release notes and left open.

**The item I held it for is closed, and I verified the fix rather than the diff.** Your three rows were pinned from above but not from below: dropping `_MAX_YAML_REQUEST_BYTES` from 1 MiB to 8 KiB — silently inheriting the neighbouring route's limit — passed the whole UI suite, **306 tests**. With your new test in place the same mutation now fails all three parametrisations:

```
_MAX_YAML_REQUEST_BYTES = 1024 * 1024   ->   8 * 1024
FAILED test_yaml_routes_keep_the_1mib_ceiling_not_the_8kib_one[/api/config/validate]
FAILED ...[/api/train/start]
FAILED ...[/api/config/from-form]
3 failed, 181 passed
```

Anchor counted exactly once, module genuinely re-imported, reverted to 184 passed. 335 passed across the wider UI / auth / ratchet / startup-is-light set on the merged tree.

**The fold is the part I want on the record.** You could have kept your own middleware and added the cap to it; instead `_YamlRequestBodyLimitMiddleware` and `_YAML_REQUEST_PATHS` are gone entirely and `app.py` went from **85 lines of new mechanism to 14 of table**. Every body-taking POST route now sits in one place — I re-enumerated rather than trusting it: `load_config_from_string` has exactly three call sites, all three are in the table, and of the seven POST routes the other four are either capped by #948 or declare no body. A second mechanism would have been a second thing to keep in sync forever.

The security properties hold under measurement, not argument: the Content-Length fast path rejects with **zero** `receive()` calls and the streamed path rejects the moment the running total exceeds the cap, so memory is bounded at cap + one chunk. Chunked bodies with no Content-Length, a lying Content-Length, trailing-slash, case and double-slash variants were all checked for a bypass and none exists.

**Zero removed test lines** against the base; the earlier boundary cases, the `receive()`-raises fast path and `assert called is False` are intact.

You also caught the `169` -> `171` recipe count in your own fragment prose without being asked. That number is not covered by any ratchet, so it would have shipped stale.

And the trailing-newline objection I raised earlier stays withdrawn — I was wrong about it in both directions and your form is what `end-of-file-fixer` produces.

Credited in `CONTRIBUTORS.md` shortly. #897 will get its closing note in the same pass.

PR #916, opened 2026-09-13
fix(data): lint the completion, not the prompt inside a conversational one

### LINKED ISSUE #810: soup data lint flags every conversational DPO row as prompt_leak MAJOR (10/10 on correct data), including both shipped DPO examples
## The defect

`soup data lint` treats the user turn inside a conversational `chosen` / `rejected` as a leaked prompt.

Conversational DPO rows carry the prompt as the first message of each completion list — `{"prompt": "...", "chosen": [{"role": "user", ...}, {"role": "assistant", ...}], ...}`. This is the shape `_convert_dpo` explicitly accepts (`data/formats.py:245-259`) and the shape both shipped DPO examples use.

`extract_pref_text` (`utils/data_lint.py:141-158`) flattens a message list by joining the `content` of **every** turn, user turns included. `check_prompt_leak` (`utils/data_lint.py:366-378`) then tests `prompt in target` against that flattened text. The prompt is always there, because it is the user turn. So every row whose prompt clears the 40-character floor (`_MIN_PROMPT_LEAK_LEN`) is flagged, and the check escalates to MAJOR at 10%.

The same flattening also feeds `check_length_bias`, so the length check measures prompt + answer rather than the answer.

## Evidence

Both shipped examples come back MAJOR on prompt_leak, and the flagged count is exactly the number of rows with a prompt of at least 40 characters:

| file | prompt lengths | flagged |

|---|---|---|

| `examples/data/dpo_sample.jsonl` | 23, 37, 34, **48, 46**, 27, **48**, 21 | 3/8 |

| `examples/data/chat_preferences.jsonl` | 25, 22, **40**, 31, **50** | 2/5 |

In other words, 100% of the rows the check is able to evaluate. These two files back `examples/configs/dpo_example.yaml` and `dpo_chat.yaml`, and are bundled by `soup data demo` (`utils/demo_bundles.py:50,58`).

A control settles it. The same 10 preference pairs, written once as conversational lists and once as plain strings:

```

=== conv10.jsonl

│ length_bias     │ MINOR   │ effect size (Cohen's d) = -0.571 (rejected longer than chosen); chosen is longer in 0.0% of rows │ mean chosen=36.8, mean rejected=37.8 │

│ prompt_leak     │ MAJOR   │ 10/10 rows (100.0%) have the prompt echoed verbatim inside the completion                        │ min_prompt_len=40                    │

exit=2

=== flat10.jsonl

│ length_bias     │ MAJOR   │ effect size (Cohen's d) = -1.142 (rejected longer than chosen); chosen is longer in 0.0% of rows │ mean chosen=20.9, mean rejected=21.9 │

│ prompt_leak     │ OK      │ 0/10 rows (0.0%) have the prompt echoed verbatim inside the completion                           │ min_prompt_len=40                    │

exit=2

```

Same data, opposite prompt_leak verdicts, and a different length_bias verdict too (MINOR vs MAJOR). The shape of the file decides the result, not its content.

## A second false MAJOR: length_bias has no magnitude floor

`flat10` above is MAJOR on length_bias for a difference of **one word** (20.9 vs 21.9). Cohen's d is scale-free: when the per-row lengths barely vary, any consistent difference becomes a "large effect". The degenerate case makes it explicit — `cohens_d` (`utils/data_lint.py:173-179`) returns exactly ±1.0 when the pooled variance is zero, and 1.0 is above the 0.8 MAJOR cut:

```

=== zerovar10.jsonl   (every chosen = 11 words, every rejected = 10 words)

│ length_bias     │ MAJOR   │ effect size (Cohen's d) = 1.000 (chosen longer than rejected); chosen is longer in 100.0% of rows │ mean chosen=11.0, mean rejected=10.0 │

exit=2

```

## Reproduction

```bash

soup data lint examples/data/dpo_sample.jsonl      # prompt_leak MAJOR 3/8 (37.5%), exit 2

soup data lint examples/data/chat_preferences.jsonl # prompt_leak MAJOR 2/5 (40.0%), exit 2

```

For the control, write 10 rows with a 40+ character prompt, then one file where `chosen` / `rejected` are `[{"role":"user","content":prompt},{"role":"assistant","content":answer}]` and one where they are the bare `answer` strings.

## Why it matters

`soup data lint` exits 2 on MAJOR, so it is a gate. Pointed at correct conversational DPO — the format Soup ships as its own example — it fails every time, and an operator can only get past it by ignoring it. That trains people to ignore the one check that catches a genuine synthetic-pipeline bug (a completion that echoes its prompt).

## Proposed fix

1. Give the completion checks a completion-only view: when `chosen` / `rejected` / `completion` is a message list, take the assistant turns only, or strip a leading prefix that equals the prompt. `identical_pairs` and `near_duplicates` can keep the full text.

2. Run `prompt_leak` against the assistant content only, so a real echo inside the assistant turn is still caught.

3. Give `length_bias` a magnitude floor as well as an effect size (for example, require the mean relative difference to exceed some percentage before MINOR/MAJOR), and do not let the zero-variance fallback produce a MAJOR on its own.

## Acceptance criteria

- [ ] A test that lints the conversational and plain-string forms of the same pairs and asserts identical `prompt_leak` verdicts. It FAILS on current code (MAJOR vs OK).

- [ ] `soup data lint examples/data/dpo_sample.jsonl` reports prompt_leak OK.

- [ ] A conversational row whose **assistant** turn repeats the prompt is still flagged.

- [ ] A dataset where chosen is one word longer than rejected in every row (zero variance) is not MAJOR on length_bias. It FAILS on current code (d = 1.000, MAJOR).

---

dchaudhari7177:
## What does this PR do?

Closes #810.

**`prompt_leak` and `length_bias` read the completion, not the prompt inside it.** A conversational `chosen` / `rejected` carries the prompt as its leading user turn. `extract_pref_text` joined every turn, so `prompt_leak` always found the prompt, and `length_bias` measured prompt + answer. A new `extract_completion_text` returns the assistant turns of a message list (a list with no roles, or a plain string, is used whole). Both checks use it. `identical_pairs` and `near_duplicates` keep the full text, as the issue proposes.

**`length_bias` needs a real size difference as well as a consistent one.** MAJOR is `|d| >= 0.8` **and** mean lengths at least 10% apart; MINOR is `|d| >= 0.3` and 5%. Cohen's d is scale-free, so a one-word gap between near-constant lengths, and the zero-variance ±1.0 fallback, were MAJOR on their own. The message now states the relative difference next to d.

## Acceptance criteria

| criterion | on `main` | here |
|---|---|---|
| conversational vs plain forms of the same pairs: identical `prompt_leak` (and `length_bias`) verdicts | MAJOR vs OK | identical, OK |
| `soup data lint examples/data/dpo_sample.jsonl` → prompt_leak OK | MAJOR | OK (0/8) |
| a prompt echoed inside the **assistant** turn is still flagged | — | MAJOR, 10/10 |
| chosen one word longer in every row (zero variance) is not MAJOR | MAJOR (d=1.000) | MINOR (9.1% apart) |

One thing to know when you run the example: `dpo_sample.jsonl` and `chat_preferences.jsonl` still exit 2, on **`length_bias`**. Their chosen answers are 84% and 74% longer than rejected by mean (d = 7.3 and 6.4), so that verdict is a real finding about the sample data, not the false positive from #810. I didn't edit the examples to make the gate pass.

The two existing near-boundary tests (`test_just_at_or_above_the_major_boundary` / `..._below...`) keep their verdicts, since those rows are 16% apart.

## Type of change

- [x] Bug fix
- [x] Tests
- [x] Documentation

## Checklist

- [x] `ruff check` passes on the touched files (0.9.4)
- [x] `pytest`: `tests/test_issue810_data_lint_completion_view.py` + `tests/test_v07127.py`: 180 passed, 3 skipped. On `main`'s `data_lint.py`, 4 of the new tests fail: shape parity, the shipped example, zero variance, and a near-identical length difference.
- [x] Updated `docs/data.md` (thresholds, and the assistant-turn view for conversational rows)
- [x] Changelog fragment added (`changelog.d/0.75.0/916.fixed.md`)


MakazhanAlpamys:
Merged as `65aa3b70`. Thank you — and the judgement in this is better than the diff size suggests.

**Verified at `0ed94f28`** in a clean worktree: `182 passed, 1 skipped` across `test_issue810_data_lint_completion_view.py` + `test_v07127.py` (the complete set of files touching `data_lint`), ruff clean, 14/14 checks green, and the test diff is `+133/-0` — no existing assertion touched.

**The mutation that mattered.** A false-positive fix is easy; not throwing out the true positive is the hard half. I replaced `extract_completion_text` with a constant `""` — disabling the detector rather than narrowing it — and confirmed it still parses, then ran the suite:

```
14 failed, 168 passed, 1 skipped
```

led by `test_a_prompt_echoed_inside_the_assistant_turn_is_still_flagged`. That is the property I actually needed, and it is pinned. Measured through the real CLI, `examples/data/dpo_sample.jsonl` goes from `prompt_leak MAJOR 3/8` to `OK 0/8`, while ten rows with the prompt echoed **inside** the assistant turn still come back `MAJOR 10/10`. Dropping either magnitude floor is killed too.

Boundary probe of the documented thresholds, by execution: `100 vs 90 → MAJOR`, `100 vs 90.01 → MINOR`, `100 vs 95 → MINOR`, `100 vs 95.01 → OK`. Docs match behaviour.

**Two things I want to name because they are choices, not accidents:**

- **Leaving `identical_pairs` and `near_duplicates` on the full text.** The tempting move is to apply the new view everywhere for consistency. You did not, and that is right: a duplicate is a duplicate across the whole row, and narrowing those two would have been a second, unrequested behaviour change riding along.
- **Declining to edit the example datasets.** Both still exit 2 on `length_bias` (`d=7.3` / `6.4`, means 66.4 vs 10.4 and 36.6 vs 9.6), so #810's reproduction commands still exit non-zero at the command level. Making them green by editing the fixtures would have looked like a more complete fix and been a worse one — that 6x gap is a genuine finding about the sample data. Saying so in the PR body instead was the correct call, and I have filed the examples separately.

Changelog fragment is in `changelog.d/0.75.0/` with inline attribution `(#810 by @dchaudhari7177 in #916)` — the format most PRs omit.

**Two follow-ups, neither blocking, take them whenever:**

1. The evidence string rounds to one decimal while the gate uses the raw value, so at the boundary MAJOR and MINOR both print `differ by 10.0%`, and MINOR and OK both print `5.0%`. A message meant to explain a verdict cannot explain it at the point where it matters most — `{rel_diff:.2%}`, or naming the threshold that was not met, would fix it.
2. Nothing pins *assistant-only* versus *not-user*: widening the filter to `turn.get("role") != "user"` — folding `system` and `tool` turns into the completion — passes all 182 tests. One parametrize case with a `system` turn in `chosen` would close it.

Credited in `CONTRIBUTORS.md`. You have three more PRs in my queue (#727, #742, #757) and they are being re-reviewed now.

PR #917, opened 2026-09-13
fix(eval): enforce exit code on MAJOR verdict and require explicit --allow-stub in quant-check (#809)

### LINKED ISSUE #809: soup eval quant-check reports OK and exits 0 when neither model loads, and exits 0 on a MAJOR verdict too
## The defect

`soup eval quant-check` is the "did quantization eat accuracy?" gate. It has two ways of reporting a pass it never measured.

**1. A model that fails to load is replaced by a stub, silently in the verdict.** `commands/eval.py:1351-1359`:

```python

try:

    before_gen = make_model_generator(resolved_before)

    after_gen = make_model_generator(resolved_after)

except (OSError, ValueError, ImportError) as exc:

    console.print(f"[yellow]Live model load failed ({exc}); using deterministic stub.[/]")

    before_gen = stub_generator("before")

    after_gen = stub_generator("after")

```

`stub_generator` (`eval/quant_check.py:197-205`) returns the constant `"[stub:before]"` / `"[stub:after]"`. Both sides then score the same (0.000), the delta is +0.000, and `classify_delta` returns `OK`. If only ONE side fails to load, the side that did load is thrown away too.

**2. The command never sets an exit code from its verdict.** After `console.print(rendered)` at line 1368/1372 the function returns, so `MAJOR` exits 0 exactly like `OK`. The `--format json` output is what the docs recommend "for CI integration" (`docs/evaluation.md:686-687`), and it carries no field saying the numbers came from a stub.

The fallback is a leftover from before live scoring existed: the docstring at lines 1296-1298 still says "Wiring live model loading is post-v0.26.0; until then, this runs with a stub generator". Live loading landed in #32; the fallback was kept.

## Evidence

The documented example cannot load. `docs/evaluation.md:675-678` passes a GGUF as `--after`:

```bash

soup eval quant-check --before ./output --after ./output/quantized.q4_k_m.gguf --tasks ./evals/sanity.jsonl

```

`make_model_generator` calls `AutoTokenizer.from_pretrained(path)` with no `gguf_file=`. With a stand-in `.gguf` file:

```

<class 'OSError'> It looks like the config file at 'quantized.q4_k_m.gguf' is not a valid JSON file.

```

`OSError` is in the caught tuple, so the documented invocation always lands on the stub and always prints OK.

Two secondary mismatches found while checking: the docs promise a "per-task" verdict with `MAJOR` at "> 10%" (`docs/evaluation.md:690-693`), while the code emits a single row named `default` for the whole file and uses `major=0.05` (`eval/quant_check.py:21`).

## Reproduction

Two empty directories and a two-line task file, run from a scratch directory:

```

$ printf '{"prompt": "What is 2+2?", "expected": "4", "scoring": "exact"}\n{"prompt": "Capital of France?", "expected": "Paris", "scoring": "contains"}\n' > tasks.jsonl

$ mkdir before after

$ soup eval quant-check --before before --after after --tasks tasks.jsonl

Live model load failed (Couldn't instantiate the backend tokenizer from one of: ...); using deterministic stub.

                  Quant check

┌─────────┬────────┬───────┬────────┬─────────┐

│ Task    │ Before │ After │  Delta │ Verdict │

├─────────┼────────┼───────┼────────┼─────────┤

│ default │  0.000 │ 0.000 │ +0.000 │ OK      │

└─────────┴────────┴───────┴────────┴─────────┘

exit=0

```

And a real MAJOR, with `make_model_generator` patched so "before" answers and "after" does not (driven through `CliRunner`, `--format json`):

```

{

  "rows": [

    {

      "task": "default",

      "before": 0.5,

      "after": 0.0,

      "delta": -0.5,

      "verdict": "MAJOR"

    }

  ]

}



exit_code = 0

```

## Why it matters

A CI step that runs `soup eval quant-check ... --format json` cannot fail. The two paths that should stop a shipment — "the quantized artifact does not load" and "it lost half its accuracy" — both exit 0, and the first one additionally prints a green OK. A broken GGUF export is the most likely thing this command will ever be pointed at, and it is exactly the case reported as a pass.

## Proposed fix

1. Remove the silent fallback. A load failure on either side prints which side failed and exits 1. Keep the stub only behind an explicit `--allow-stub` flag for orchestration smoke tests, and when it is used, mark the result (`"stub": true` in JSON, a visible banner in table/markdown) so it cannot be mistaken for a measurement.

2. Exit with the verdict: 0 for OK / MINOR, 2 for MAJOR, matching `soup ship` and `soup eval behavior`.

3. Either load GGUF through `gguf_file=` or refuse a `.gguf` path up front with a message naming the supported formats, and fix the example in `docs/evaluation.md`.

4. Make the docs and code agree on the thresholds and on "per-task" vs one aggregate row.

5. Update the stale docstring.

## Acceptance criteria

- [ ] `soup eval quant-check --before <empty dir> --after <empty dir> --tasks t.jsonl` exits non-zero and names the side that failed to load. A test for this FAILS on current code (it exits 0 with OK).

- [ ] A test with patched generators producing a MAJOR delta asserts exit code 2. It FAILS on current code (exit 0).

- [ ] A load failure on only the `--after` side is reported as such; the `--before` scores are not silently discarded.

- [ ] Stub scoring is only reachable via an explicit flag, and its JSON output says so.

- [ ] The documented GGUF example either works or is replaced; docs thresholds match `classify_delta`.

---

kok-o:
Resolves #809

### Summary
Fixes `soup eval quant-check` reporting `OK` and exiting with code 0 when live models fail to load, and exiting with code 0 on a `MAJOR` accuracy regression verdict.

### Key Changes
1. **No silent stub fallback**: Separated `--before` and `--after` loading in `commands/eval.py`. When live loading fails without `--allow-stub`, reports which side failed (`Failed to load --before model` / `Failed to load --after model`) and exits with code 1.
2. **Explicit `--allow-stub` option**: Added `--allow-stub` option (default `False`). When stub scoring is active, marks the payload (`"stub": true` in JSON, and warning banners in table/markdown) and retains live `--before` generator if only `--after` failed.
3. **Exit code enforcement**: Enforces exit code 2 when any task evaluates to `MAJOR` delta, and code 0 on `OK` / `MINOR`.
4. **Standalone `.gguf` pre-flight check**: Refuses standalone `.gguf` file paths up front with a descriptive message naming supported formats.
5. **Documentation**: Updated `docs/evaluation.md` examples and corrected thresholds to match `classify_delta` (≤2% OK, 2–5% MINOR, >5% MAJOR).
6. **Tests**: Added comprehensive acceptance test suite in `tests/test_quant_check.py` (`TestQuantCheckIssue809`).

### Verification
- `pytest tests/test_quant_check.py --no-cov -v` (30 passed)
- `ruff check src/soup_cli/commands/eval.py src/soup_cli/eval/quant_check.py tests/test_quant_check.py` (All checks passed)


Amix29:
One remaining #809 scope item: the issue asks for docs/code agreement on **per-task vs one aggregate row**. This patch fixes the thresholds, but `docs/evaluation.md` still says `Verdict thresholds (per task)` while `run_quant_check()` scores the whole JSONL once and emits the default `task_name="default"` row. Please either describe this as an aggregate suite verdict or implement actual task grouping; the former is the narrow fix consistent with this PR.

MakazhanAlpamys:
Reviewed at `3291cef6`. **Verified the fix by execution, not by reading the diff.** On #809's own repro — two empty directories — `main` exits 0 with a green `OK 0.000 0.000 +0.000`, and this branch exits 1 naming `--before` as the side that failed. That is the defect.

Four mutations, all killed: MAJOR back to exit 0 (2 failed), `--allow-stub` made unnecessary at both sites (1), the `.gguf` pre-flight disabled (1), and `stub=is_stub` forced to `False` (2). 30/30 locally, **405 passed** across the ten existing files touching `commands/eval.py` and `eval/quant_check.py`, ruff clean, and the test diff is **199 insertions, 0 deletions** — nothing removed or relaxed.

**One required change before merge: this command is now on a fifth exit-code convention.** Measured across all three trees:

| case | main | this PR | #921 |
| --- | --- | --- | --- |
| missing `--tasks` | 1 | **1** | 3 |
| `--format xml` | 1 | **1** | 3 |
| missing `--before` dir | 1 | **1** | 3 |

You added three *new* `typer.Exit(1)` sites — the `.gguf` refusal at `eval.py:1352` and both load failures at `1378`/`1393`. A `.gguf` path is a pure usage error, and the project taxonomy puts that at **3**. #809 did say "exits 1", so you followed your ticket — but **#813, which you filed the same day and cross-linked from #809, settles it.**

Please move `eval.py:1352` and the pre-existing usage exits in that function to `EXIT_USAGE_ERROR`. Keeping the **live load failure** at `EXIT_RUNTIME_ERROR` is defensible — `ship.py` treats a model-load failure as `_EXIT_RUNTIME` — but if that is the intent, say so in the docstring rather than leaving it as an implicit distinction three lines from a usage exit.

**Three smaller things:**

1. `tests/test_quant_check.py:322-324` is weaker than it looks. `assert result.exit_code != 0` cannot distinguish 1 from 2 from 3 — which is the exact failure mode #813 exists to fix — and `"Failed to load --after model"` is a **prefix shared by both branches**: the refusal at `eval.py:1390` and the `--allow-stub` warning at `:1391` both start with it. That test survived my "make `--allow-stub` unnecessary" mutation for precisely that reason (a live `before` plus a stub `after` yields MAJOR, exit 2, still non-zero). Assert `== 1` and add `assert "using deterministic stub" not in output`.
2. Under `FORCE_COLOR=1 TERM=xterm-256color`, 4 of the 30 fail: Rich's `ReprHighlighter` colours the JSON even with `markup=False`, so `json.loads(result.output)` receives `\x1b[1m{\x1b[0m...`. Latent on CI (CliRunner is not a tty) — but it is **also a live product bug**: `soup eval quant-check --format json | jq` in a real terminal gets ANSI-polluted JSON today. Adding `highlight=False` at `eval.py:1411` fixes the product bug and the four tests together, which is the better reason to do it.
3. `docs/evaluation.md` still says "Verdict thresholds (**per task**)" while the code emits one aggregate row named `default`. That was item 4 of #809; the thresholds half is fixed (2/10% -> 2/5%) and this half is not.

**Finally: this is a breaking change and the PR does not say so.** The old docstring advertised the stub as being for "CI smoke-checks", so there is a real population whose green job turns red. The migration path (`--allow-stub`) is implemented and documented, which is what matters — it just needs an explicit "**Breaking:** a run that previously passed on stubs now needs `--allow-stub`" sentence in the fragment so it reaches the release notes.

`eval.py:1342-1344`: the `or (path_obj.is_file() and suffix == ".gguf")` half can never be reached — the left operand already covers it.

**Merge order: this lands first, then #921.** #921 is currently red (14 pre-existing tests assert the old codes) and claims `quant-check` in its body without fixing it — I measured that the silent stub is still live there. Landing yours first makes that claim true for the MAJOR path and shrinks #921's remaining work in this function to re-numbering your three `Exit(1)` sites. The reverse order would ask you to resolve a conflict against a taxonomy contradicting your own ticket, inside a hunk. Your textual conflict with #921 is **one line** — you both add the same `has_major` block, differing only literal-vs-constant.

MakazhanAlpamys:
@kok-o — reviewed at `3291cef6`. **The two headline fixes are real and both are mutation-pinned.** Six of your seven new tests fail on `origin/main`, so the baseline is genuine rather than a test written around existing behaviour. What holds it up is a collision with your own #921, and it is worse than a textual conflict.

**First, a correction in your favour.** I went in expecting the load-failure exit code to be wrong. It is not. `ship.py:84` names "model load" as the canonical runtime error, so load failure -> 1 is the house answer and your PR description is right. Taxonomy against every sibling:

| condition | `soup ship` | `soup shrink` | `soup diagnose` | this PR |
|---|---|---|---|---|
| pass verdict | 0 | 0 | 0 | **0** |
| gate says NO (MAJOR) | 2 | 2 | 2 | **2** |
| model / IO load failure | 1 | 1 | 1 | **1** |
| bad flag / unsupported input | **3** | — | — | **1** |

Only the last row is wrong, and it is a new exit site this PR introduces, so it is not grandfathered the way the pre-existing `not found` -> 1 sites are.

---

**BLOCKING 1 — the GGUF refusal should be 3, not 1.** `commands/eval.py:1344-1350` rejects a `.gguf` path *before* any loading is attempted (line 1344 sits above the `make_model_generator` block at 1371). Your own new docstring at `eval.py:1298-1299` says "code 1 on **loading or runtime** errors" — a path-shape rejection is neither. Under `ship.py:86` ("bad flags / validation") that is `_EXIT_USAGE = 3`. The test does not pin it either: `tests/test_quant_check.py:465` asserts `exit_code != 0`, which passes at 1, 2 or 3.

**BLOCKING 2 — #917 and #921 collide, and #921's own fixtures fail against this branch.**

```
$ git merge --no-commit --no-ff pr921rv
CONFLICT (content): Merge conflict in src/soup_cli/commands/eval.py
```

That part is expected — you add the byte-identical `has_major` block at the same location in both. The semantic half is the problem. #921 rewrites every `typer.Exit(1)` in `quant_check_cmd` to `EXIT_USAGE_ERROR` (3) **and keeps the silent stub fallback** that this PR deletes. I ran #921's fixtures against this code:

```
921 pass_args    under #917 -> exit 1 | 921 asserts 0 (EXIT_OK)
921 missing_args under #917 -> exit 1 | 921 asserts 3 (EXIT_USAGE_ERROR)
```

#921's "passing" fixture (`tests/test_issue813_gate_exit_codes.py:153-169`) hands quant-check two stub `.safetensors` files and expects exit 0 — which works **only because of** the silent stub fallback #809 is about. So a careless conflict resolution taking #921's side reintroduces the defect this PR exists to fix.

**My call: land #921 first, then rebase this onto it** using the shared `exit_codes` constants. That direction fixes Blocking 1 for free (GGUF -> `EXIT_USAGE_ERROR`, load failure -> `EXIT_RUNTIME_ERROR`), and #921's pass fixture needs `--allow-stub` added. **But #921 is currently red on all nine test cells**, so it is not landable as-is either. Your call which you fix first; I am not merging either until one of them is green and the other is rebased on it.

---

**A mutation SURVIVED, and it is an acceptance criterion of #809.** `eval/quant_check.py:101-103` — I deleted the table-format stub banner (anchor matched once, module imports fine) and **all 30 tests still passed**. The reason is that `tests/test_quant_check.py:418` asserts `"deterministic stub" in result_tbl.output`, and the warning at `eval.py:1381/1396` already prints *"...using deterministic stub."* to the same stream. The assertion is satisfied by the warning, never by the banner. #809 asks for "a visible banner in table/markdown output"; the markdown path is pinned (I killed that one and a test went red), the table path is not. Assert on the distinctive part of the title instead — `"not a live measurement"`.

**Related, smaller:** `tests/test_quant_check.py:323-325` survived the `--allow-stub` mutation too. With the guard gone the after side stubs, real-before(1.0) vs stub-after(0.0) gives MAJOR -> exit 2, which satisfies `exit_code != 0`. Only one of the two relevant tests actually caught the defect. Tighten `!= 0` to the specific code at `:298`, `:323` and `:465`.

---

**ANSI — latent, but this PR is what makes it matter.** Four tests use `json.loads(result.output)` (`:355, :381, :400, :442`). Under `FORCE_COLOR=1 TERM=xterm-256color`:

```
4 failed, 26 passed   — all four json.decoder.JSONDecodeError
```

Rich's `ReprHighlighter` colours every JSON key, string and number even with `markup=False`:

```
'\x1b[1m{\x1b[0m\n  \x1b[32m"rows"\x1b[0m: ... \x1b[32m"before"\x1b[0m: \x1b[1;36m0.0\x1b[0m'
```

CI sets no `FORCE_COLOR`/`PY_COLORS` (I grepped the workflows), so this is not red today. I am asking anyway, for a reason beyond the tests: **your docs newly promote `--format json` as the CI contract** (`docs/evaluation.md:689`). Anyone running `soup eval quant-check --format json | jq` in a terminal, or in a CI that does set `FORCE_COLOR`, gets ANSI-poisoned JSON. The defect is pre-existing at `eval.py:1411` (`console.print(rendered, markup=False)` leaves `highlight` on), but this is the PR that makes JSON the gate, so it is the right place for `highlight=False`. The six substring assertions survive colourization; it is only the `json.loads` calls that break.

---

**Three smaller items:**

- **The breaking change never names its own escape hatch.** `eval.py:1375-1377` and `1390-1392`. Everyone whose quant-check ran on the stub now gets `Failed to load --before model (...): Couldn't instantiate the backend tokenizer...` with no mention of `--allow-stub`. The remedy exists only in docs the failing user is not reading. Please append something like `(pass --allow-stub to score with deterministic stubs instead)`.
- **`changelog.d/0.75.0/917.fixed.md` reads `(#809 in #917)`** — missing `by @kok-o`. `882.fixed.md` has the exact form to copy: `(#878 by @Srinivasan8888 in #882)`.
- `eval.py:1342-1343`: `path_obj.suffix.lower() == ".gguf" or (path_obj.is_file() and path_obj.suffix.lower() == ".gguf")` — the right operand is a strict subset of the left, so it is unreachable; and it fires on a *directory* named `*.gguf` while the message says "standalone GGUF file". `eval.py:1368-1369`: `before_gen = None` / `after_gen = None` are dead, every path assigns before use. `docs/evaluation.md:694` says "OK — drop <= 2%" but `classify_delta` uses `drop < minor`, so an exact 2.0% drop is MINOR.

**What is solidly right:** zero tests deleted, one strengthened (`:268`, `assert "stub" not in data`), the thresholds corrected against `classify_delta`, the broken GGUF doc example replaced, the stale docstring removed, ruff clean, and 282 passed across the eval suites. The stub-fallback mutation fails with `assert 0 != 0` — the #809 defect verbatim — which is exactly the right failure to have pinned.

kok-o:
@MakazhanAlpamys — pushed updated commit addressing all review findings:

1. **Exit code on GGUF refusal**: Refusal of standalone `.gguf` paths now exits with code 3 (usage/validation error) and removed redundant check in `commands/eval.py`. Test pins `assert result.exit_code == 3`.
2. **ANSI-clean JSON output**: Added `highlight=False` to `console.print` at `commands/eval.py:1411` so Rich ReprHighlighter does not pollute JSON with ANSI escape codes under `FORCE_COLOR=1`.
3. **Model loading error messages**: Appended `(pass --allow-stub to score with deterministic stubs instead)` to both `--before` and `--after` loading failure messages.
4. **Table stub banner test**: Pinned `assert "not a live measurement" in " ".join(clean_tbl.split())` in `tests/test_quant_check.py` using `strip_ansi` to handle table width wrapping and kill the banner deletion mutation.
5. **Tightened test assertions**: Pinned `result.exit_code == 1` for model loading failures, `assert "using deterministic stub" not in result.output`, and `result.exit_code == 3` for `.gguf` refusal.
6. **Documentation & Changelog**: Updated `docs/evaluation.md` to specify aggregate suite verdict and exact `classify_delta` thresholds (`< 2% OK`, `2–5% MINOR`, `≥ 5% MAJOR`), and updated `changelog.d/0.75.0/917.fixed.md` with credit (`@kok-o`) and explicit breaking change notice.
7. **Rebased**: Rebased cleanly onto latest `main` (`fcca8ef`). All 30 quant-check tests pass (including under `FORCE_COLOR=1`), 158 eval platform tests pass, `ruff check` clean.

MakazhanAlpamys:
Verified `744e9cda`. **Merge-ready**, and all six previously-blocking items are closed — including the two mutations that survived my last review.

**Mutations: six run, six killed.** Anchors counted before replacement, module import re-verified after each, every failure a behavioural assert rather than a collection error, reverted green each time.

| mutation | result |
|---|---|
| delete the table stub banner (`title = "Quant check"`) | **1 failed** — *previously survived* |
| `Exit(2 if has_major else 0)` to `Exit(0)` | 2 failed |
| gguf `Exit(3)` to `Exit(1)` | 1 failed |
| `stub=is_stub` to `stub=False` | 2 failed |
| `if not allow_stub:` to `if False:`, both sites | **2 failed** — *previously survived* |
| drop `highlight=False` | 30 passed plain, **4 failed** under `FORCE_COLOR=1` |

The last one is the one I want to name. It reproduces my earlier count of 4 exactly, which proves `highlight=False` is load-bearing rather than cosmetic — and the *product* half checks out too: `--format json` under `FORCE_COLOR=1 TERM=xterm-256color` now parses as valid JSON. On `main` it does not. A machine-readable output that stops being machine-readable when a human is watching is a nasty class of bug and you closed it in the product, not just in the test.

Baseline on `origin/main`: `6 failed, 24 passed`, the six being exactly the new #809 tests. `tests/test_quant_check.py` gives 30 passed plain and 30 under colour; the broader eval suites give 226 passed. `ruff` clean. **Zero test lines deleted** — 203 insertions, 0 deletions.

I also measured the actual behaviour change on #809's own repro, two empty directories:

```
origin/main                                  this PR
exit=0                                       exit=1
Live model load failed (...); using          Failed to load --before model (before): ...
deterministic stub.                          (pass --allow-stub to score with
| default | 0.000 | 0.000 | +0.000 | OK |    deterministic stubs instead)
```

Two smaller things: removing the unreachable `or`-clause also stops the guard firing on a *directory* named `*.gguf`, which makes the code match its own message; and splitting `!= 0` into exact codes with the two `not in` assertions at `:327-328` is what discriminates the prefix collision I flagged — that pair is doing real work.

**Sequencing, since it affects #921 and not this PR.** The silent fallback lives at `eval.py:1352-1360` on `main` — one `try:` wrapping both `make_model_generator` calls, both sides overwritten with stubs at `:1359-1360`. This PR deletes it and raises `typer.Exit(1)` at `:1374` and `:1390`. #921 has a fixture that hands quant-check unloadable stub weights and asserts exit 0; that fixture passes *because* of the block this PR removes. So **#917 lands first**, and #921 rebases onto it — resolving the other way would restore #809 verbatim. I have said the same on #921.

Two non-blocking notes:

1. Exit codes inside this one function are still internally mixed — GGUF is 3, while bad `--format`, unresolvable ref, outside-cwd, `--tasks` not found and before/after not found are all `Exit(1)` despite being usage errors. You grandfathered those deliberately and #921 renumbers them; your `3` already agrees with where they will land. Flagging only so the residue is visible at merge.
2. Three new positive assertions read `result.output` raw (`:301`, `:326`, `:470`). They pass under `FORCE_COLOR` because each substring sits inside a single `[red]` span with no quoted string or number for `ReprHighlighter` to split — and the one assertion that *does* wrap, the table at `:421`, correctly uses `strip_ansi`. No CI risk today, one message rewording away from one.

Only CI is left; the queue is at ~88 runs. I will merge as soon as it is green, and before #921.

MakazhanAlpamys:
Re-verified against current `main` (`b0a63382`), 16 merges after my review. **Unchanged and still merge-ready:** `merge-tree` clean, 30 passed on both the PR head and the merged tree, ratchets 138 passed, fragment intact.

The mutation still kills: `raise typer.Exit(2 if has_major else 0)` to `Exit(0)` fails `test_major_verdict_asserts_exit_code_2` and `test_allow_stub_retains_live_before_when_after_fails` on `assert 0 == 2`.

Two colour checks, since this is the area that has turned CI red four times: the repo-wide ANSI AST scanner has no objection to your new assertions, and `--allow-stub` survives an ANSI strip on `--help`.

And the sequencing constraint I put on you is **gone** — #921's rebase removed its dependency on the stub fallback, and I verified by merging #917 into it locally: no conflicts, 42 passed. Merge order between the two is now free.

Nothing needed from you.

MakazhanAlpamys:
Merged as `de254ce4`. Thank you, and an apology for the shape of the wait.

**What actually happened, since you were owed this earlier.** This PR was merge-ready at `744e9cda` at 04:05Z. It did not merge then because GitHub Actions on this repo produced **no terminal outcome at all** between 2026-09-13T18:30Z and today 15:08Z — 95+ runs queued, two of them stuck in `in_progress` for over 28 hours. Your matrix was not failing; it was not running. The queue drained on its own at 15:08Z and this merged at 15:53Z, before anything else.

That is a repository problem, not a problem with your PR, and re-stating the review each time I checked made it read as if the bar were moving. It was not — the verdict from 04:05Z is the verdict I merged on. I am recording review verdicts against a head SHA now so that an unchanged head gets its checks re-polled instead of a second review.

**What I merged on, unchanged:**

All 13 required contexts green on `744e9cda` itself, not on an ancestor — confirmed per-SHA via `check-runs`, not from a cached PR summary.

Six mutations run, six killed, each with the anchor counted before replacement and the module re-imported after it, and every failure a behavioural assert rather than a collection error:

| mutation | result |
|---|---|
| delete the table stub banner | 1 failed |
| `Exit(2 if has_major else 0)` → `Exit(0)` | 2 failed |
| gguf `Exit(3)` → `Exit(1)` | 1 failed |
| `stub=is_stub` forced to `False` | 2 failed |
| `--allow-stub` made unnecessary at both call sites | 1 failed |
| `.gguf` pre-flight disabled | 1 failed |

**The two things this PR did that I want on the record.**

Six of your seven new tests fail on `origin/main`. That is the property that separates a regression test from a test written around behaviour that already existed, and most PRs I see do not have it.

And the second round is the better one. When I told you two mutations had survived — the stub banner could be deleted and the table still read as a live measurement — the cheap answer was an assertion on that string. You instead made the stub a *labelled* result the caller has to opt into with `--allow-stub`, which turns the banner into load-bearing behaviour rather than decoration. The test now passes because the design changed, not because the test got more specific.

You were also right where I was wrong: I went in expecting the load-failure exit code to be incorrect, and `ship.py:84` makes load failure → 1 the house answer. Your description had it right and my expectation did not.

Credited in `CONTRIBUTORS.md` (`cbe7363f`). #809 is closed with credit.

PR #918, opened 2026-09-13
fix(reward): add structure-preserving attack families to soup reward stress (#814)

### LINKED ISSUE #814: soup reward stress calls the built-in `format` reward "robust (not gameable)" while `<think></think>a` scores 1.0
## The defect

`soup reward stress` answers "could a reward-hacking model game this verifier?". Its attack set is four fixed strings (`utils/reward_stress.py:33,71-82`):

| attack | text |
|---|---|
| `empty` | `""` |
| `length` | `"padding"` x 60 |
| `repetition` | `"loop"` x 40 |
| `sentinel` | `"GOLD"` x 20 |

None of them contains the structure a structure-rewarding verifier pays for. The built-in `format` reward (`trainer/rewards.py:235-261`) gives 0.5 for any `<think>...</think>` block — `.*?` also matches an empty body — and 0.5 for any non-whitespace after `</think>`. It never looks at content. So the stress tool rejects nothing it would have to, and prints the strongest verdict it has.

## Evidence

The stress run:

```
$ soup reward stress format
Using built-in reward function: format
│ attack      n  accepted  accept-rate │
│ empty       1         0           0% │
│ length      1         0           0% │
│ repetition  1         0           0% │
│ sentinel    1         0           0% │
│ reference accept: n/a   gameability: 0%   verdict: robust (not gameable) │
exit=0
```

The same function, loaded the same way (`load_reward_fn('format')`), scoring junk wrapped in the structure:

```
'<think>x</think>GOLD'            [1.0]
'<think></think>a'                [1.0]
'<think>loop loop</think>loop'    [1.0]
'<think>.</think>.'               [1.0]
'The answer is 42.'               [0.0]
```

Every degenerate completion gets the maximum score. A correct plain answer gets zero.

The same gap exists for the gold-requiring `accuracy` reward, from the other direction. `accuracy` pays 0.5 when the expected text appears anywhere in the completion (`trainer/rewards.py:225-229`), and 0.5 meets the stress threshold (`DEFAULT_THRESHOLD = 0.5`, accepted when `>=`). An answer-spray — every plausible answer at once — is therefore accepted, and stress still calls it robust:

```
$ soup reward stress accuracy --references golds.jsonl      # golds: 42, 17, 256
... reference accept: 100%   gameability: 0%   verdict: robust (not gameable)
exit=0

accuracy('<think>x</think> 1 2 3 17 42 256', answer=['42'])  -> [0.5]
accuracy('<think>x</think> 1 2 3 17 42 256', answer=['17'])  -> [0.5]
format  ('<think>x</think> 1 2 3 17 42 256')                 -> [1.0]
```

This is the combination the `deepseek-v3-reasoning` recipe ships (`reward_fn: accuracy,format`, `recipes/catalog.py:699`).

The control: the `verifiable` math reward rejects the same strings, so these attacks separate a correct verifier from a content-blind one rather than failing everything:

```
verifiable/math('<think>x</think>GOLD', answer=['42'])              -> [0.0]
verifiable/math('<think></think>a', answer=['42'])                  -> [0.0]
verifiable/math('<think>x</think> 1 2 3 17 42 256', answer=['42'])  -> [0.0]
verifiable/math('<think>x</think> #### 42', answer=['42'])          -> [1.0]
```

Sample size is a separate weakness: without `--references` each attack is scored once (`n=1`); with references it is the SAME junk string repeated once per gold (`reward_stress.py:173-176`), so `n` grows without any variation in what is being tried.

## Reproduction

```bash
soup reward stress format; echo "exit=$?"
python -c "from soup_cli.trainer.rewards import load_reward_fn; f=load_reward_fn('format'); print(f([[{'role':'assistant','content':'<think></think>a'}]]))"
```

## Why it matters

`format` is not broken — it is a shaping reward by design and is meant to be combined with a correctness reward. The problem is the verdict. `docs/training.md:1080-1086` presents `stress` as the adversarial check that "asks whether a reward-hacking model could game it", and exit 0 plus "robust (not gameable)" is what an operator will act on. A GRPO policy finds `<think></think>a` in a few hundred steps; the tool built to warn about exactly that says there is nothing to find.

## Proposed fix

Add structure-preserving attack families, each generated rather than a single literal:

1. **Wrapped junk**: each existing attack inside the common reasoning / answer scaffolds (`<think>{junk}</think>{junk}`, empty think block, `#### {junk}`, `\boxed{{junk}}`, a JSON object with junk values). This family alone flips `format` to gameable.
2. **Answer spray**: for gold-requiring verifiers, a completion containing the gold plus several distractors, and the gold embedded in unrelated text. This catches the `accuracy` 0.5 substring credit.
3. **Structure without content** for `json_schema` / `tool_call` verifiers: schema-valid objects with placeholder values.
4. Several variants per family (different fillers and lengths), so `n` counts distinct attempts rather than one string repeated.

Also consider reporting `format`-style shaping rewards as "content-blind" explicitly rather than robust, since no amount of attacking changes what they measure.

## Acceptance criteria

- [ ] `soup reward stress format` reports gameable and exits 2. A test for this FAILS on current code (exit 0, "robust").
- [ ] `soup reward stress accuracy --references g.jsonl` flags the answer-spray attack at the default threshold.
- [ ] `soup reward stress verifiable --verifiable-domain math --references g.jsonl` still reports robust (it scores the wrapped junk and the spray 0.0 today, so the new families must not introduce a false positive on a correct verifier).
- [ ] Each attack family contributes more than one distinct string, and the report's `n` reflects distinct attempts.
- [ ] `--attacks` accepts the new family names and rejects unknown ones with the list of valid names.

kok-o:
## Summary
Fixes #814 by adding structure-preserving attack families to `soup reward stress` and generating multiple distinct variant attempts per attack family.

## The Problem
`soup reward stress` answers "could a reward-hacking model game this verifier?". Previously, its attack set only comprised 4 fixed classic junk strings (`empty`, `length`, `repetition`, `sentinel`). None of them contained the structure that structure-rewarding verifiers pay for:
- The built-in `format` reward gives 0.5 for `<think>...</think>` and 0.5 for non-whitespace trailing content without inspecting inner content. Because classic junk lacked `<think>` scaffolds, `soup reward stress format` reported `robust (not gameable)` (exit code 0) even though degenerate completions like `<think></think>a` score 1.0.
- Similarly, `accuracy` pays 0.5 when expected text appears anywhere in the completion (substring match). An answer-spray with distractor options meets the 0.5 threshold, but stress reported it as robust.
- Sample sizes were weak: without `--references`, each family was scored only once (`n=1`); with references, the exact same string was repeated once per gold without variation.

## Changes Made
1. **Added Structure-Preserving Attack Families (`STRUCTURE_ATTACKS`)**:
   - `wrapped_junk`: reasoning and structure scaffolds (`<think>padding</think>padding`, `<think></think>a`, `<think>loop loop</think>loop`, `<think>.</think>.`, `#### padding`, `\boxed{padding}`, `{"response": "padding", "answer": "junk"}`).
   - `answer_spray`: embeds the gold target among distractor options and trailing numbers (`<think>options</think> 1 2 3 17 {target} 256 999`, etc.).
   - `structure_without_content`: empty schema/tool dictionaries for JSON schema and tool-call verifiers.
   - Combined `ATTACKS = CLASSIC_ATTACKS + STRUCTURE_ATTACKS` (7 families total).
2. **Multiple Distinct Variant Attempts**:
   - `generate_attack_variants(kind, sentinel=..., gold=...)` returns distinct variants for every family.
   - Without references, `n` reflects the variant count (e.g. 3, 7, 4).
   - With references, `answer_spray` substitutes sampled gold answers across templates and other variants are evaluated across sampled references.
3. **CLI & Documentation**:
   - Updated `--attacks` help string to enumerate all 7 families.
   - Updated `docs/training.md` and `docs/commands.md` to document the new attack families.
4. **Verification**:
   - `soup reward stress format` reports `GAMEABLE` and exits 2.
   - `soup reward stress accuracy --references g.jsonl` flags the `answer_spray` attack (100% acceptance) and exits 2.
   - `soup reward stress verifiable --verifiable-domain math --references g.jsonl` continues to report `robust (not gameable)` and exits 0.
   - Added `tests/test_issue814_reward_stress.py` covering all 5 acceptance criteria; existing test suites `test_v07141.py` and `test_cli_startup_is_light.py` remain fully green.


MakazhanAlpamys:
Reviewed at `a88903a9`. **The core of this is right, and I verified it rather than reading it.** The single most useful control was replacing all three new families with trivially-rejected junk while keeping their names: the CLI went straight back to `robust / exit 0` and three of your tests went red. That is the property that matters — a new attack family that nothing ever accepts is decoration, and yours are not. `soup reward stress format` now exits 2 with `wrapped_junk 4/7` and `answer_spray 1/4`, and `accuracy --references` flags `answer_spray` at 12/12. #814's headline defect is genuinely closed.

**One blocker, and it is the false positive the issue's own third acceptance criterion exists to prevent.**

The `answer_spray` distractors are hardcoded (`999`, `321`, `777`, `333`). `verifiable/math` resolves the trailing number, so when a user's real gold **equals** one of them the junk completion scores 1.0 against its own gold, and a **correct** verifier is reported gameable. I swept every integer gold from 0 to 1000 against an unmodified math verifier:

```
falsely GAMEABLE golds in 0..1000: [321, 333, 777, 999]
count: 4
```

Exactly the four constants, nothing else. Small and mid-range integers are precisely what a math reference set contains.

Because `max_gameable` defaults to `0.0`, **one** colliding gold anywhere in the file flips the gate. Ten golds, one of them `999`:

```
gameability: 0.37 %   gameable: True
reference_accept: 1.0
   family answer_spray accepted 1 of 40
```

Note the report contradicts itself: a gameability figure that rounds to 0% sitting beside a GAMEABLE verdict and exit 2. `soup reward stress` is documented as a CI gate where exit 2 blocks, so this fails a correct verifier on roughly a 1-in-250 coincidence, and the panel gives the operator nothing to act on.

**Remedy:** derive the distractors from the gold rather than hardcoding them, and guarantee they differ — build the list, drop any element equal to `target`, and append a value that cannot collide (`str(int(target) + 1)` when numeric, otherwise a token that is not the gold). The regression test writes itself:

```python
@pytest.mark.parametrize("gold", ["999", "333", "777", "321"])
def test_a_gold_equal_to_a_distractor_is_not_a_false_positive(gold):
    assert run_stress(math_verifier, [gold]).gameable is False
```

**Two more before merge:**

- **`tests/test_issue814_reward_stress.py:148` asserts on raw Rich output.** Under `FORCE_COLOR=1 TERM=xterm-256color` it fails: `_fail()` prints through a Rich `Console` and `ReprHighlighter` colours `kind(s)` mid-phrase, so the substring is split by escapes (`'\x1b[31munknown attack \x1b[0m\x1b[1;31mkind\x1b[0m...'`). Latent rather than live today, since CliRunner is not a tty — but this exact class has turned this repo's CI red **four times**, and the repo's own guard only scans `--help`, so it cannot see your file. Route every `r.output` assertion there through the ANSI-strip helper in `tests/test_v07302.py`. Lines 149-150 are the same risk, since the option list already wraps.
- **`structure_without_content` is accepted by nothing we ship, and it is not verdict-neutral.** Measured 0 acceptances against `format`, `accuracy`, `verifiable/math`, and verifiers synthesized by `soup reward synth` for both `json_schema` and `tool_call`; deleting the family kills only the two name-registration tests. That alone would be cosmetic, but an always-rejected family dilutes the pooled denominator: on `format` at `--max-gameable 0.20`, `19% / robust / exit 0` with it and `22% / GAMEABLE / exit 2` without. So a family that never fires can **hide the very finding this issue is about** for anyone who raises the tolerance off the default. Either give it a test that proves it fires — score it against a deliberately content-blind structure checker, the way `wrapped_junk` is pinned against `format` — or make the verdict per-family rather than a pooled rate, so the number of variants a family happens to ship stops moving the answer.

**Two surviving mutations, both small:**

- Reverting `generate_attacks()` to one text per family leaves all 47 tests green. It now has **zero production callers** (`run_stress` builds variants itself) and `_attack_text` has zero callers of any kind — so either delete both, or assert the multi-variant property on `generate_attacks` directly. Two parallel attack-construction paths that can drift is the shape of #372.
- Deleting the new `isinstance(attacks, (str, bytes))` guard in `run_stress` also leaves 47 green. Its sibling in `generate_attacks` **is** tested. One `pytest.raises(TypeError, match="sequence")` closes it.

**Smaller notes:** `assert "gameable" in r.output.lower()` is vacuous — `"robust (not gameable)"` contains it too. `run_stress` now issues 27 `_score_batch` calls against 4 before, a 6.75x increase in verifier invocations (5,400 scored completions at the 200-gold cap); fine for a `.py` verifier, worth a sentence in `docs/training.md` for a slow one. `--attacks` in `--help` truncates at 80 columns, so the two new family names are not discoverable there, though `docs/commands.md` carries the full list. And the changelog fragment needs `by @kok-o` — attribution is the one currency this project pays in.

**One pre-existing bug I found while testing, not yours:** `soup reward stress verifiable --verifiable-domain json_schema` cannot work at all — `json_schema_reward` reads `kwargs["schema"]` while `_score_batch` passes `answer=`, so it returns `[]` and hard-errors. I will file it separately.

kok-o:
Addressed all review findings in `afca0d6`:

1. **Distractor False Positives against `verifiable/math` eliminated**:
   - In `generate_attack_variants("answer_spray")`, distractors and trailing numbers are now derived dynamically from `target` (using numeric offsets `+10, +20, +30` and trailing offsets `+1, +2, +3, +4`, with non-numeric fallbacks). This guarantees no distractor matches `target` and trailing values never collide with gold.
   - Swept integer golds across `0..1000` against `verifiable/math`: **0** false positives (down from 4). `accuracy` still flags `answer_spray` at 100% acceptance (4/4).
   - Added parameterized regression test `test_a_gold_equal_to_a_distractor_is_not_a_false_positive` covering `["999", "333", "777", "321"]`.

2. **ANSI and Wrapping Resilience in Tests**:
   - Routed all CLI output checks in `tests/test_issue814_reward_stress.py` through a `_plain()` helper that strips ANSI escape codes and collapses whitespace (matching `tests/test_v07302.py`).
   - Replaced vacuous `"gameable" in r.output.lower()` with unambiguous check: asserts `"GAMEABLE" in clean_out and "robust (not gameable)" not in clean_out.lower()`.
   - Verified clean pass under `FORCE_COLOR=1 TERM=xterm-256color`.

3. **`structure_without_content` Verified**:
   - Added `test_structure_without_content_fires_on_content_blind_verifier` proving that `structure_without_content` fires (100% acceptance, exit code 2, `gameable=True`) against a content-blind JSON structure verifier that rewards valid dictionary structures without inspecting inner content.

4. **Surviving Mutations Killed & Dead Code Removed**:
   - Deleted dead `_attack_text`.
   - Added `test_generate_attacks_multi_variant` asserting `generate_attacks()` produces multiple distinct variants for every registered family.
   - Added `test_run_stress_rejects_str_attacks` asserting `TypeError` when a bare string is passed for `attacks` to `run_stress`.

5. **Docs & Changelog**:
   - Added a note in `docs/training.md` documenting the multi-variant evaluation scale and explaining the proportional increase in verifier invocations for slow/model-based verifiers.
   - Updated changelog attribution in `changelog.d/0.75.0/918.fixed.md` to `(#814 by @kok-o in #918)`.
   - Rebased cleanly on latest `upstream/main` (`fcca8ef`). All 88 tests in the related suites and `ruff check` are fully green.

MakazhanAlpamys:
Reviewed `afca0d61`. **Three of the four blockers are genuinely closed** — and I verified each rather than reading your summary:

- **B1, `answer_spray` false positives.** Distractors are derived from `target` at `reward_stress.py:118-141`, not a hardcoded swap. Re-running my sweep: 0 false positives over the 0..1000 integer sweep (was 4 — 999/321/777/333), 0 of 20 on the edge set, 0 against a synthesized numeric verifier.
- **B2, ANSI.** All six content assertions route through `_plain()`. I checked the colour run is non-vacuous first — ANSI genuinely reaches `result.output`, 122 ESC bytes under colour against 0 without — then 54 passed both ways.
- **B4, mutations.** All three closed; `_attack_text` is gone.

**B3 is not closed, and it is the blocker.** Two halves, both reproduced here at your head.

### (a) The proof that `structure_without_content` fires is a strawman

The verifier in `tests/test_issue814_reward_stress.py:176-187` is, in full:

```python
d = json.loads(content)
scores.append(1.0 if isinstance(d, dict) else 0.0)
```

That pays 1.0 for *any* parseable JSON object. Nothing this project ships behaves that way — its own `json_schema_reward` (`trainer/rewards.py:635-655`) scores schema conformance. Against everything real:

```
format / accuracy / math              accepted 0/4 each
synthesized TOOL_CALL verifier        structure_without_content  12  0  0%
synthesized JSON_SCHEMA verifier      structure_without_content  12  0  0%
```

Zero acceptances against the two verifier kinds #814 names as its own target. The four payloads carry fixed keys (`name`/`arguments`, `type`/`properties`, `tool`/`parameters`, `thought`/`action`), so the family fires only if a user's schema happens to use those exact keys.

### (b) A family that has never fired flips a real verdict

The pooled denominator is untouched — `gameability = total_accepted / total_n` at `reward_stress.py:295`. I reproduced the flip:

```
soup reward stress format --max-gameable 0.20
  wrapped_junk               7  4  57%
  answer_spray               4  1  25%
  structure_without_content  4  0   0%
  gameability: 19%   verdict: robust (not gameable)     exit=0

same, without structure_without_content
  gameability: 22%   verdict: GAMEABLE                  exit=2
```

**22% to 19%, exit 2 to exit 0.** Adding a family that accepts nothing made a gameable verifier report as robust. Note `wrapped_junk` alone is 57% accepted while the pooled rate says 19% — the averaging is what hides it, and this will get worse every time a family is added.

### To unblock

Either (i) demonstrate `structure_without_content` firing against a verifier this project ships or `reward synth` generates — not a test-local fixture — or drop the family; **and** (ii) fix the denominator, because the dead family currently downgrades a real finding. A per-family verdict (`max(accept_rate) > max_gameable`) fixes (ii) independently of (i), and also stops the number of variants a family happens to ship from moving the answer. I would take that on its own merits even if (i) is resolved.

### Three more findings, in priority order

- **The `@parametrize` on the B1 fix reads four times stronger than it is.** Mutating derived distractors back to the hardcoded `999/321/777/333` is caught by only **1 of the 4** golds — `[333]`, because that value lands in the trailing slot. `999/777/321` do not discriminate at all. Worth a gold that fails for a different structural reason.
- **`--attacks` validation became lazy — a regression this PR introduces.** Validation moved out of the eager `generate_attacks()` into the per-kind scoring loop (`reward_stress.py:250-285`), so a typo now runs the target verifier on a whole family before erroring: **8 invocations before the error at your head, 1 on `main`.** On a slow or model-backed verifier — the case your own docs note warns about — a typo now costs a full family. Validate all kinds up front.
- **Float distractors collapse for large golds.** At `|gold| >= ~1e5` the `:.4g` formatting maps all four distractors to one string numerically equal to the gold (`gold 100000.0` yields `1e+05 1e+05 100000.0 1e+05 1e+05`). There is no false positive today **only by luck** — `_MATH_NUM_RE` at `trainer/rewards.py:284` cannot match scientific notation, so it reads `05`. Any verifier that parses sci-notation false-positives, and the spray degenerates to one repeated value.

Everything else is clean: baseline genuine (`9 failed, 6 passed` on `main`, the 6 being the stays-robust controls); no scope creep (7 files, +397/−44); fragment correct; docs updated in both `commands.md` and `training.md` with the 27-invocation figure checking out; `ruff` clean; `test_cli_startup_is_light.py` green.

One deletion I want to acknowledge rather than have you defend: narrowing `test_real_accuracy_builtin_gold_path_robust` **had** to happen, since the old assertion is now false by design — that is acceptance criterion 2 — and you added compensating coverage. Side effect worth knowing: the *default* attack set no longer has a boundary test.

MakazhanAlpamys:
Heads up on ordering, not on your findings: #941 merged just now (`23766cb4`) and it touches `src/soup_cli/utils/reward_stress.py`, `src/soup_cli/commands/reward.py` and `docs/training.md` — all three of which you also touch. You will need a rebase.

The overlap is not just textual. #941 changed `_score_batch`'s signature (`answers` becomes `references` plus a keyword-only `reference_key`), reworded the docstring block your PR also edits, and added a raise on the short-return path. Re-read those three hunks after the rebase rather than resolving them mechanically.

My earlier review still stands and is unaffected by this — in particular the `--attacks` validation becoming lazy, where a typo now runs the target verifier on a whole family before erroring (8 invocations at your head against 1 on `main`). That is the one I would want closed before this merges.

kok-o:
@MakazhanAlpamys — rebased onto latest `main` (`4d98b6ad`) and resolved all findings:

1. **Rebase & #941 compatibility**: Rebased cleanly onto latest `main` and resolved conflicts with #941, preserving `reference_key` routing for `json_schema` across `run_stress` and `_score_batch`. Updated `test_issue937_reward_stress_json_schema.py` assertions to match the multi-variant evaluation contract.
2. **B3 closed (structure family dropped & per-family verdict)**: Dropped `structure_without_content` and switched `gameability` calculation to `max(a.accept_rate for a in results)`, preventing denominator dilution and ensuring no attack family can mask another.
3. **Eager `--attacks` validation**: Added immediate validation of attack names at `run_stress` entry before `reward_fn` invocation, preventing wasted compute on typos.
4. **Float distractor collapse fixed**: Ensured large float targets (`|gold| >= 1e5`) do not collapse into repeated scientific notation distractors.
5. **Verification**: 83/83 tests passing across `test_issue814_reward_stress.py`, `test_issue937_reward_stress_json_schema.py`, `test_v07141.py`, and `test_issue487_changelog_fragments.py`; `test_cli_startup_is_light.py` green (34/34); `ruff check` clean.

MakazhanAlpamys:
Merged as `134f121a`. Thank you — and **I owe you a correction in your favour.**

**My blocking finding is not just closed, you went past it.** I said the `--attacks` validation had become lazy and that a typo ran the target verifier on a whole family before erroring — 8 invocations at `afca0d61` against 1 on `main`. I re-measured on your current head rather than trusting that:

```
your head : raised ValueError: unknown attack kind: '...' -> verifier invocations before the error: 0
main      : raised ValueError: unknown attack kind: '...' -> verifier invocations before the error: 1
```

**Zero, against main's one.** So the widening made the validation *stricter*, not lazier, and you improved on the baseline I was measuring you against. That is the opposite of what my finding said would need fixing, and it deserves saying plainly rather than being quietly folded into a merge.

**B3 is closed, and both halves are the right call:**

`structure_without_content` is **dropped**, not defended. Its proof that it fires was a verifier paying 1.0 for any parseable JSON object — a strawman, catching a verifier nobody would write. Deleting your own attack family because its evidence was weak is harder than arguing for it.

`gameability` is now `max(accept_rate)` rather than the mean. That one matters more than it looks: averaging dilutes a family a verifier genuinely fails by every family it passes, so **adding more attacks would have made a gameable verifier look safer** — the metric would have worked against the feature.

The rest was already verified at the previous head and I did not re-derive it: `answer_spray`'s distractors are derived from the target rather than a fixed pool, giving 0 false positives over a 0..1000 sweep, over a 20-item edge set, and against a synthesized numeric verifier; all six content assertions route through the ANSI strip, and I confirmed the colour run was non-vacuous first (122 ESC bytes under colour against 0 without) before trusting 54 passed both ways.

83 passed on the merged tree, 14/14 required green on the head.

**And an apology for the timing.** You posted the rebase at 19:47 yesterday and heard nothing for twelve hours. You have three PRs open and all three were sitting on me at once, which I only noticed after auditing every open PR for who spoke last rather than relying on my own notes. Your #743 now has the two commits you asked me to push, and #921 has its single failing test named on the thread.

Credited in `CONTRIBUTORS.md` shortly; #814 gets its closing note in the same pass.

PR #919, opened 2026-09-13
fix(train): honor chat template overrides across trainers

### LINKED ISSUE #804: data.chat_template is applied by SFT and distill only: DPO renders conversational rows with the model's shipped template instead

Amix29:
Closes #804.

## What does this PR do?

`data.chat_template` is now installed on the tokenizer before TRL renders conversational data in DPO, KTO, ORPO, SimPO, IPO, BCO, Reward Model, GRPO, and Online DPO.

- The override runs after whichever tokenizer load path was selected, so ordinary Transformers, Unsloth, and layer-streaming preference paths share the same behavior.
- GRPO and Online DPO only install their built-in fallback if the configured override still leaves the tokenizer without a template.
- The tokenizer object later saved by each trainer carries the same template used for training.
- `pretrain`, `embedding`, `classifier`, `reranker`, and `cross_encoder` now reject `data.chat_template` with the task named instead of accepting a field they never use.
- The chat-template documentation states the supported scope.

## Regression coverage

`tests/test_issue804_chat_template_trainers.py` includes:

- a real tiny CPU DPO setup whose prepared `prompt_ids` decode with an `OVERRIDE` marker and not the model's shipped `SHIPPED` marker;
- save/reload proof that the tokenizer persists the override;
- runtime wiring checks for the eight offline conversational wrappers;
- explicit ordering checks for the GRPO and Online DPO fallbacks;
- a real local Online DPO tokenizer/model load showing the override beats its fallback;
- task-named refusal tests for all five non-chat tasks.

The existing multi-task fixture was adjusted so it no longer supplies a now-invalid chat template to non-chat tasks.

## Verification

- `ruff check src/soup_cli/ tests/` — clean
- issue-specific + existing chat-template tests — 46 passed
- live TRL preference contracts + multi-task setup matrix — 131 passed
- config + MLX tests — 31 passed
- broad suite before stopping after the first unrelated failure — 2,920 passed, 12 skipped, 5 deselected
- the unrelated failure reproduces alone on `origin/main`-independent code: `tests/test_issue204_langfuse_pull.py::test_control_the_leak_scan_sees_a_token_rich_highlighted_and_folded` assumes Rich will wrap one planted string, but the installed Rich renders both unbroken

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [ ] `pytest tests/ -v` passes — one unrelated, isolated Rich wrapping assertion fails as recorded above
- [x] Updated the relevant page under `docs/` (README does not mention this field)


MakazhanAlpamys:
Reviewed at `21ca2c86`. I checked the thing this repo keeps getting wrong: whether the field is actually **read**, or only plumbed somewhere plausible-looking.

**It is read.** `apply_chat_template_override` mutates the tokenizer in place, the call sits *after* the three-way dispatch (streaming / unsloth / transformers) so all three load paths are covered by construction, and all nine trainers hand **that same object** to TRL — `processing_class=self.tokenizer` at `dpo:222 kto:211 orpo:222 simpo:222 ipo:180 bco:234 reward_model:175 grpo:526 online_dpo:308`, plus `grpo:416/559` — each saving it exactly once. Your DPO test closes the loop end-to-end on decoded `prompt_ids`, and the `"SHIPPED" not in rendered` assertion is what makes it a real test rather than a presence check.

**Per-trainer coverage confirmed by mutation, which was my main worry.** I removed the override from **`simpo` alone** — the pre-PR state for one trainer — and got exactly **2 failures, both parametrized on `simpo`**, with the other eight green. So a single-trainer regression is visible rather than hidden behind one shared assertion. I also kept the call and its spelling but passed `None`, which is the "validated, documented, read by nothing" shape this project is named for, and **9 tests fail** including the DPO end-to-end one — while the `inspect.getsource` check happily passes. The behavioural half is doing the work. Making the GRPO/Online-DPO fallbacks unconditional fails 4 (2 behaviourally); disabling the non-chat rejection fails 5, one per task. No survivors.

**Runs:** your file 28 passed; +chat-template/config/contract files 451 passed; +21 files importing the nine changed trainers 580 passed, 0 failed; the #748 / #755 / #487 / #775 repo guards 106 passed; ruff clean. No shipped config is affected — `chat_template` appears in zero recipes, templates or examples, so none of the 167 recipes can trip the new validator.

I was suspicious of the `and not self.training.stream_layers` clause at `schema.py:4293` — it looks like a guard written to satisfy its own test — so I executed it. All five unsupported tasks are **still refused** with `stream_layers: true`, just by the streaming validator's message instead. `unsupported ∩ stream-capable = ∅`, so it only decides *which* of two errors the user sees and no config leaks through. **Please add a one-line comment saying that is why it is there**, or someone will delete it, or worse, copy the pattern.

**One test-file adaptation I checked line by line:** `tests/test_issue353_seed_every_task.py` stopped feeding `chat_template: chatml` to `pretrain`/`embedding`/`classifier`, because those configs are now invalid **by this PR's own validator**. No assertion weakened; the fixture simply had to stop constructing configs that no longer exist.

**Follow-up, please file it.** `ppo`, `prm`, `asr` (and `moe_lora_routing`, `unlearn`) still **accept `data.chat_template` and read nothing** — I enumerated all 23 tasks against the real `SoupConfig`, and `grep` finds no hits in any of the five. `ppo` is the notable one: `ppo.py:914-920` hand-joins `msg["content"]` with spaces, the same class of bug #332 fixed for vLLM. Outside #804's acceptance criteria so not this PR's job, but the class is not closed.

**Two smaller things:**

- `test_grpo_applies_override_before_fallback` and its online-DPO sibling assert source **order**, and I measured that both orders are behaviourally identical — the override wins either way, because it overwrites. They also match a literal adjacent line, so a harmless refactor breaks them, and your behavioural tests already cover the real hazard. `CLAUDE.md`'s own warning about spelling-versus-behaviour tests applies; consider dropping them.
- The fragment needs `by @Amix29`. The checklist also ticks "Updated README.md" while README is unchanged — nothing was owed there, it never mentioned the field, so that is just an inaccurate box.

**Merge-order heads-up:** you and #908 both add exactly one line at the *same* import anchor in all nine of these files, so that is a guaranteed add/add conflict ×9 for whoever lands second — trivial to resolve, but re-run `ruff check` afterwards since import order is enforced. Your own #923 does **not** conflict with this one (different functions; its only import change there is function-local).

Approving once the remaining CI cells report.

Amix29:
Addressed the review at `c55764c`:

- documented why the chat-template validator defers to the streaming task validator
- removed the two fragile source-order assertions; the behavioral GRPO/Online-DPO tests remain
- added the required changelog attribution
- corrected the PR checklist wording
- filed #936 for the remaining PPO/PRM/ASR/MoE/unlearn accepted-but-unused paths

Validation: 43 targeted tests passed and the full Ruff scope is clean. Rebased onto current `origin/main`.

MakazhanAlpamys:
Verified at `c55764c`. All five items done, and I checked each rather than taking the summary.

- The `stream_layers` clause now carries its reason inline (`schema.py:4294-4295`), which is what stops the next reader deleting it or copying the pattern.
- Both source-order assertions are gone — `grep -c "applies_override_before_fallback"` returns **0** — and the behavioural GRPO and Online-DPO tests remain, which are the ones that actually catch a fallback overwriting the user's override.
- The fragment reads `(#804 by @Amix29 in #919)`.
- #936 is filed for the `ppo` / `prm` / `asr` / `moe_lora_routing` / `unlearn` accepted-but-unread paths, which keeps that class open and named rather than buried in a review thread.

Removing tests you had written, because they assert a spelling rather than a behaviour, is a harder thing to do in review than adding more — and it is the right call here. `CLAUDE.md` has the same lesson written down from #357/#396, and it keeps costing this project releases.

Merging as soon as the checks report — they are still queued on this head.

**Ordering, mine to manage:** you and #908 both add exactly one line at the *same* import anchor in all nine of these trainer files, so whoever lands second gets nine add/add conflicts. They are trivial — keep both lines, alphabetise — but re-run `ruff check` afterwards, since import order is enforced. Your own #923 does not conflict with this one.

MakazhanAlpamys:
Merged as `c6b61657`. Thank you — and the review round on this one is the part I want on the record.

You had two assertions on **source order** (`applies_override_before_fallback`), and you deleted them when I asked rather than defending them. That matters because an assertion that one statement precedes another catches the refactor that moves it, not the defect it is aimed at — a fallback overwriting the user's override. The behavioural GRPO and Online-DPO tests that replaced them do catch it. `grep -c "applies_override_before_fallback"` is now 0.

Putting the reason for the `stream_layers` clause inline (`schema.py:4294-4295`) is the other thing that will still be paying off in a year: it is what stops the next reader deleting it or copying the pattern without knowing why it exists.

And you filed #936 for the four trainers still not reading the field rather than widening this PR to reach them. Right call.

Your CONTRIBUTORS.md entry is in and #804 is closed.

PR #920, opened 2026-09-13
fix(profile): say when the GPU is assumed, and know newer and laptop cards

### LINKED ISSUE #830: soup profile says "Fits in 24 GB VRAM" on a machine with no GPU, rejects --gpu rtx3050/rtx5070/rtx5080/b200, and uses A100 throughput for every card
## The defect

`soup profile` has three separate problems:

1. When no GPU is detected it silently assumes 24 GB and reports the fit as OK.

2. Its GPU table has no RTX 3050, no 50-series card below the 5090, no Blackwell datacenter parts and no laptop variants.

3. Its throughput estimate is an A100 lookup that does not depend on the GPU at all.

## Evidence

- `commands/profile.py:126-127`, the fallback when auto-detection finds no device:

  ```python

  # Default to 24 GB (common consumer GPU)

  return 24.0

  ```

  Nothing in the output says the number was assumed.

- `utils/profiler.py:6-31` (`GPU_MEMORY`) has `rtx5090: 32` and none of `rtx3050`, `rtx5060`/`5070`/`5080`, `b200`. Every key is a desktop model, and laptop parts differ: the RTX 5070 Laptop has 8 GB where the desktop RTX 5070 has 12 GB, and the RTX 4090 Laptop has 16 GB where the desktop has 24 GB.

- `estimate_speed` (`utils/profiler.py:222-255`) takes model size, quantization and batch size only. Its docstring says "Based on typical A100 throughput". `--gpu` changes the memory verdict but never the speed or time estimate.

## Reproduction

`examples/configs/sft_basic.yaml` (TinyLlama-1.1B, `quantization: none`) on an RTX 5070 Laptop (8151 MiB, driver 616.92). Panel padding trimmed.

With CUDA hidden, which is the same state as a CPU-only torch (PyPI's default on Windows):

```

$ CUDA_VISIBLE_DEVICES=-1 soup profile --config examples/configs/sft_basic.yaml

│ Tokens/sec: ~2,500                                                    │

│ OK Fits in 24 GB VRAM                                                 │

│ OK Recommended batch_size: 8                                          │

```

With the card visible:

```

$ soup profile --config examples/configs/sft_basic.yaml

│ Tokens/sec: ~2,500                                                    │

│ OK Fits in 8 GB VRAM                                                  │

│ OK Recommended batch_size: 2                                          │

```

Same throughput in both cases, and a batch-size recommendation 4x profile's own answer for this card.

```

$ soup profile --config examples/configs/sft_basic.yaml --gpu rtx5070

Unknown GPU: rtx5070

Valid options: a10, a100, a100_40gb, a30, a40, h100, h200, l4, l40, l40s, rtx3060, rtx3070, ...

```

The same `Unknown GPU` (exit 1) comes back for `--gpu rtx3050`, `rtx5080`, `b200` and `"rtx 5070 laptop"`.

## Why it matters

`soup profile` is the documented "before training" check. An OK that rests on assumed hardware is worse than no estimate: the failure just moves to an OOM at step 0. The time estimate is printed as a single number, so someone budgeting a run on a T4 is shown A100 wall-clock. The cards that are missing (3050, 5060-5080, laptop parts) are exactly the ones where a memory estimate decides whether a run is possible at all.

## Proposed fix

- When no device is detected and `--gpu` is not given, say so. Either withhold the fit verdict or label it "assuming 24 GB (no GPU detected; pass --gpu)".

- Extend `GPU_MEMORY` with `rtx3050` (plus the 6 GB variant), `rtx5060`/`5060ti`/`5070`/`5070ti`/`5080`, `b200`/`gb200`, and explicit laptop keys where memory differs. Accept the device name as torch reports it ("NVIDIA GeForce RTX 5070 Laptop GPU").

- Scale `estimate_speed` by a per-GPU relative throughput, or use `layer_stream_runtime.measure_gemm_tflops` when a GPU is present, and print a bracket rather than a point.

## Acceptance criteria

- [ ] A test that makes `get_gpu_info` report no device, runs `soup profile` without `--gpu`, and asserts the output does not present "Fits in 24 GB" as a measured verdict. It fails on current code.

- [ ] `--gpu rtx3050`, `--gpu rtx5070`, `--gpu "rtx 5070 laptop"` and `--gpu b200` are accepted with the right memory.

- [ ] `estimate_speed` returns different values for `t4` and `h100` on the same config.

---

- #831 — run_cost prices an RTX A4000 as an A40 and an A100-SXM4-80GB as an A100 40GB; no rate for L4, H200, B200 or...

dchaudhari7177:
## What does this PR do?

Closes #830, scoped to its two acceptance criteria.

**An assumed GPU is no longer shown as a measured fit.** `_resolve_gpu_memory` now returns where the number came from: `flag` (`--gpu`), `detected`, or `assumed` (no device and no flag, still 24 GB). When it is assumed, the Recommendations panel reads (TinyLlama-1.1B, `quantization: none`, no GPU):

```
? No GPU detected: would fit in an assumed 24 GB (need ~4 GB). Pass --gpu for a real verdict.
? batch_size 8 assumes 24 GB VRAM
```

instead of `OK Fits in 24 GB VRAM` / `OK Recommended batch_size: 8`. `--json` gains `gpu_memory_source`, and so does the MCP `profile` tool, whose resolver mirrors this one.

**Newer and laptop GPUs.** `GPU_MEMORY` gains `rtx3050` (8), `rtx3050_6gb`, `rtx5060` (8), `rtx5060ti` (16), `rtx5060ti_8gb`, `rtx5070` (12), `rtx5070ti` (16), `rtx5080` (16), `b200` (180: DGX B200's 1,440 GB over eight GPUs), and laptop keys where memory differs from the desktop card (`rtx4050laptop` … `rtx5090laptop`, e.g. `rtx5070laptop` 8, `rtx4090laptop` 16). `normalize_gpu_key` drops `nvidia` / `geforce` / `tesla` / `gpu`, spaces and hyphens, so `--gpu "rtx 5070 laptop"` and torch's `"NVIDIA GeForce RTX 5070 Laptop GPU"` both resolve.

**Speed estimate.** It is still the A100 lookup, but the panel now says "Typical A100 throughput, not specific to your GPU". Scaling it per GPU or measuring it is the third proposed fix; I left it out here because it is a model change rather than a labelling one. I also didn't add `gb200`: its per-GPU memory isn't stated consistently across NVIDIA's own spec pages, and I'd rather not guess.

## Type of change

- [x] Bug fix
- [x] Tests

## Checklist

- [x] `ruff check` passes on the touched files (0.9.4)
- [x] `pytest`: `tests/test_issue830_profile_gpu.py` + `tests/test_profile.py` + `tests/test_v07128.py` (MCP profile tool): 155 passed, 1 skipped. On `main`'s sources, 9 of the new cases fail: the no-GPU verdict, `gpu_memory_source` for assumed and detected, and all six `--gpu` names. The unknown-GPU rejection passes on both.
- [ ] Updated relevant docs: no doc page lists the `--gpu` values or the fit wording
- [x] Changelog fragment added


MakazhanAlpamys:
Reviewed at `b39b4404`. The user-facing half of this is right, and I want to name the judgement call before the findings: **you declined the third acceptance criterion instead of half-doing it**, said so up front in the issue and again in the PR body, and then mitigated it honestly by labelling the throughput figure rather than inventing a per-GPU model. That is the correct call and I would rather have this than a guessed `estimate_speed` table.

Also correct: `normalize_gpu_key` strips the noise words *before* collapsing whitespace. Doing it the other way round kills the `\b` anchors.

**Verified by execution**, not by reading the diff. With `get_gpu_info` raising, the panel reads "No GPU detected: would fit in an assumed 24 GB … Pass `--gpu` for a real verdict" instead of `OK Fits in 24 GB VRAM`, and `--json` carries `gpu_memory_source`. Your no-GPU mutation is pinned three ways — forcing the fallback to report `"detected"` fails `test_no_detected_gpu_is_not_reported_as_a_measured_fit`, `test_json_says_the_gpu_memory_was_assumed` and the pre-existing `TestResolveGpuMemory::test_fallback_to_default_24gb`. The tests diff is `+118/-0` on the new file and `+2/-1` on `test_profile.py`, where the single deletion is a forced signature change that *adds* `assert source == "assumed"` — strengthened, not weakened. `ruff` clean. 278 passed / 1 skipped across every test file that touches the changed modules. No docs page enumerates `--gpu` values or quotes the fit wording, so your unchecked docs box is accurate — I grepped rather than assumed.

**Two mutations survived, and both block merge.** I ran each myself:

| mutation | result |
| --- | --- |
| fallback returns `"detected"` instead of `"assumed"` | 3 failed — correctly pinned |
| revert **both** `mcp_server/registry.py` changes | **190 passed, 1 skipped — survived** |
| delete 6 of the new `GPU_MEMORY` keys (`rtx5080`, `rtx5060ti`, `rtx5070ti`, `rtx3050_6gb`, `rtx5090laptop`, `rtx4060laptop`) | **155 passed, 1 skipped — survived** |

**1. The entire MCP half is unguarded.** I dropped `result["gpu_memory_source"]` and restored the old `gpu.lower().replace(" ","").replace("-","")`, and `test_v07128.py` + `test_issue830_profile_gpu.py` + `test_profile.py` + `test_mcp_execute.py` stayed green. `TestProfileHandler::test_returns_estimate` asserts only `total_memory_gb` / `recommended_batch_size` / `compatible_gpus`, and `test_unknown_gpu_raises` passes under either normalisation because `"nonesuch-gpu"` is rejected both ways. Your changelog fragment advertises this to users, so it needs to be held in place. Two small tests close it:

```python
def test_profile_reports_where_the_gpu_number_came_from(self, tmp_path, monkeypatch):
    monkeypatch.chdir(tmp_path)
    (tmp_path / "soup.yaml").write_text(_MIN_CONFIG, encoding="utf-8")
    monkeypatch.setattr(
        "soup_cli.utils.gpu.get_gpu_info",
        lambda: (_ for _ in ()).throw(RuntimeError("no GPU")),
    )
    assert reg.tool_profile({"config": "soup.yaml"})["gpu_memory_source"] == "assumed"

def test_profile_accepts_a_torch_device_name(self, tmp_path, monkeypatch):
    monkeypatch.chdir(tmp_path)
    (tmp_path / "soup.yaml").write_text(_MIN_CONFIG, encoding="utf-8")
    out = reg.tool_profile(
        {"config": "soup.yaml", "gpu": "NVIDIA GeForce RTX 5070 Laptop GPU"}
    )
    assert out["gpu_memory_gb"] == 8.0
    assert out["gpu_memory_source"] == "flag"
```

The second is the one that pins `normalize_gpu_key` on the MCP path — the first alone would not.

**2. Thirteen of the nineteen new `GPU_MEMORY` keys are untested.** Only `rtx3050`, `rtx5070`, `rtx5070laptop` (two spellings), `rtx4090laptop` and `b200` are exercised. `rtx5080` is named in #830's own title and can be deleted with the suite green. Please don't answer this with nineteen more parametrize rows — a completeness test keeps working as the table grows:

```python
@pytest.mark.parametrize(("key", "gb"), sorted(_NEW_IN_ISSUE_830.items()))
def test_every_card_added_for_830_is_in_the_table(key, gb):
    assert GPU_MEMORY[key] == gb
```

**Non-blocking, worth folding into the same push:**

- `_render_profile(..., gpu_memory_source: str = "detected")` defaults to the confident wording — the bug this PR exists to fix, re-armed for the next caller. There is exactly one caller today, so make it required. A default meaning "measured" on a function whose purpose is to stop claiming measurement is the wrong default.
- The `GPU_MEMORY` comment says laptop keys exist "wherever their memory differs from the desktop card", but `rtx4060laptop`, `rtx5060laptop` and `rtx5080laptop` are identical to their desktop entries. Including them is **right** — otherwise torch's device name would not resolve — so it is the comment that needs widening, not the table.
- You introduced `_ASSUMED_GPU_MEMORY_GB` in `profile.py` and left `registry.py:445` on a bare `24.0`. Import it; that file's own docstring already calls itself a mirror.
- `terminal_width=200` in the first test does nothing — `profile.py` builds its `Console()` at import, before `CliRunner` sets `COLUMNS`, so the panel still renders at ~78 columns and wraps mid-sentence. Harmless today, but the argument implies a guarantee it is not giving, and the assertions are wrap-fragile as a result.
- `json.loads(result.output)` on `--json`: under `FORCE_COLOR=1` 8 of your 14 tests fail with `JSONDecodeError`, because Rich's `ReprHighlighter` colours the JSON. This is a **pre-existing pattern** — `test_profile.py::TestProfileCLI::test_profile_json_output` fails identically on `main` — and CI is non-TTY, so it is latent rather than live. But this PR takes it from one instance to nine, and it is the exact class `tests/test_cli_help_assertions_are_ansi_safe.py` exists for. `json.loads(re.sub(r"\x1b\[[0-9;]*m", "", result.output))` is a one-line fix.
- `b200: 180` derived from DGX B200's 1,440 GB over eight GPUs is documented in the comment and conservative against the 192 GB HBM3e SKUs. Leave it; declining to guess at `gb200` for the same reason was right.

The changelog fragment is in `changelog.d/0.75.0/` with the right shape and inline credit. When this merges I will use `Refs #830`, not `Fixes` — the `estimate_speed` criterion stays open, and that is your call to have made, not a gap.

dchaudhari7177:
Both blockers and all the non-blocking notes addressed in `59f09aa`.

**1. MCP half.** Added `test_mcp_profile_reports_an_assumed_gpu` and `test_mcp_profile_accepts_a_torch_device_name`, following your sketch (the second one is what pins `normalize_gpu_key` on the MCP path).

**2. Table completeness.** `test_every_card_added_for_830_is_in_the_table` is parametrized over a `_NEW_IN_ISSUE_830` dict that holds all nineteen cards and their sizes. A dropped key and a wrong size both fail on their own row.

Your two surviving mutations, re-run against this head:

| mutation | before | now |
| --- | --- | --- |
| revert both `registry.py` changes | survived | **2 failed** |
| delete `rtx5080`, `rtx5060ti`, `rtx5070ti`, `rtx3050_6gb`, `rtx5090laptop`, `rtx4060laptop` | survived | **6 failed** |

**Non-blocking, all folded in:**

- `_render_profile`'s `gpu_memory_source` is now required.
- The `GPU_MEMORY` comment now says every laptop part has its own key, including the ones whose memory matches the desktop card, because that is what lets torch's `"... Laptop GPU"` name resolve.
- The 24 GB fallback is a single `ASSUMED_GPU_MEMORY_GB` in `utils/profiler.py`, and both `commands/profile.py` and `registry.py` import it. I put it in the profiler instead of importing from `commands/profile.py` because `registry.py`'s header wants its imports kept cheap, and `profiler.py` only pulls in `math` and `re`.
- `terminal_width=200` is gone. A `wide_console` fixture now monkeypatches `soup_cli.commands.profile.console` with `Console(width=200)`, which does reach the panel.
- `--json` output goes through a `_json()` helper that strips SGR codes. `FORCE_COLOR=1 pytest tests/test_issue830_profile_gpu.py` gives 35 passed. I left the pre-existing instance in `test_profile.py` alone to keep this diff scoped.

Locally: `test_issue830_profile_gpu.py`, `test_profile.py`, `test_v07128.py` → 176 passed, 1 skipped. `test_mcp_execute.py` also passes. `ruff check` is clean on the four touched files.

Agreed on `Refs #830`, since the `estimate_speed` criterion stays open.

MakazhanAlpamys:
Verified `59f09aa9`. **Merge-ready** — both blocking mutations are dead, and I re-ran each rather than reading your summary.

| mutation | last time | now |
|---|---|---|
| revert both `registry.py` changes (drop `result["gpu_memory_source"]`, restore `gpu.lower().replace(...)`) | **survived**, 190 passed | 2 failed |
| delete the six new table keys | **survived**, 155 passed | 6 failed, one per row of `test_every_card_added_for_830_is_in_the_table` |
| fallback returns `"detected"` instead of `"assumed"` | 3 failed | 3 failed |

Each mutated module was import-checked before the run, so these are behavioural kills rather than a `NameError` masquerading as coverage. Baseline on `origin/main`: **34 failed, 1 passed** — the single survivor being `test_an_unknown_gpu_is_still_rejected`, exactly as predicted.

Three things worth naming specifically:

- **The parametrised per-card test.** Six separate failures instead of one aggregate is precisely what made the previous mutation survivable, and fixing it this way means the next person who trims the table learns which row they trimmed.
- **You closed the ANSI trap by measuring it, not by declaring it.** `FORCE_COLOR=1 TERM=xterm-256color` gives 35 passed; the `_json()` SGR strip works and the three remaining raw-output assertions survive colour. The one red under `FORCE_COLOR` is `test_profile.py::test_profile_json_output`, which fails identically on `origin/main` — pre-existing, and correctly left out of scope rather than swept in.
- **Criterion 3 was declined, not quietly dropped.** `gh issue view 830` confirms per-GPU `estimate_speed` is the third *proposed fix*, not a fourth requirement, and the `batch_size 8 assumes 24 GB VRAM` label mitigates the same confusion. Merging as `Refs #830` with that criterion left open, as agreed.

Also verified: no test lines deleted — the single `-` is `mem = _resolve_gpu_memory(None)` becoming a two-value unpack **plus** a new assertion, so it is strengthened; fragment at `changelog.d/0.75.0/920.fixed.md` with correct credit; no docs page enumerates `--gpu` values, so the unchecked docs box is accurate; all five non-blocking items landed, including the `wide_console` fixture that actually reaches the panel where `terminal_width=200` did not.

One informational note, **not** a blocker: the table growing means `recommend_gpu` now leads "Compatible GPUs" with the 6 GB parts rather than the 8 GB ones, since it sorts ascending and shows five. Arguably an improvement — smallest sufficient card first — and nothing asserts on it. Flagging it so it is not a surprise later.

Only CI is left; the Actions queue is at ~88 runs. I will merge as soon as your matrix is green.

MakazhanAlpamys:
Re-verified against current `main` (`b0a63382`), 16 merges after my review. **Unchanged and still merge-ready:** `merge-tree` clean, 85 passed on both the PR head and the merged tree, the mutation still fails three tests on `assert 'detected' == 'assumed'`, ratchets 138 passed.

One thing I checked specifically because it could have bitten: `main` rewrote `mcp_server/registry.py` across **five** commits (+64/−75) since your review. Your edit is confined to `_resolve_gpu_memory_mcp` — now returning `(float, str)` — and to `tool_profile`'s new `gpu_memory_source` key, and none of those five touch either. Ran the four MCP test files on the merged tree: 76 passed, with one definition and one call site, already unpacking the tuple correctly.

Nothing needed from you.

MakazhanAlpamys:
Merged as `f900fcf9`. Thank you — this is your second merge today and the second one that waited on the repository rather than on review.

**On the wait.** GitHub Actions here produced no terminal outcome at all between 2026-09-13T18:30Z and today 15:08Z: 95+ runs queued, two stuck `in_progress` past 28 hours. Your checks were not failing, they were not running. This merged 20 minutes after the queue reached your head, on the verdict recorded at `59f09aa9` — unchanged, and not re-derived.

**What I checked before merging, since `main` had moved under you again.** Not a second review — three things:

1. All 13 required contexts green on `59f09aa9` itself, read per-SHA from `check-runs` rather than from the cached PR summary.
2. `mergeStateStatus` CLEAN.
3. Whether `main` moved into anything your change interacts with. It had, twice, and one of them was new since my last note:
   - `mcp_server/registry.py` was rewritten across five commits (+64/−75). Your edit is confined to `_resolve_gpu_memory_mcp` — now returning `(float, str)` — and `tool_profile`'s new `gpu_memory_source` key; none of the five touch either. 76 passed across the four MCP test files on the merged tree, with the one definition and the one call site unpacking the tuple correctly.
   - **#757 merged earlier today**, and it adds a package-wide AST scanner that walks *every* command module for typer parameters left unfilled at a call site — `commands/profile.py` included. That is exactly the kind of interaction that is green on both branches and red on the merge, so I built the merged tree and ran it: **11 passed**, no leak from your diff.

85 passed on both the PR head and the merged tree; ratchets 138 passed. The mutation still holds — forcing the source back to `detected` fails three tests on `assert 'detected' == 'assumed'`.

**What the fix is worth, stated plainly.** The green line was the defect, not the fallback. A user with no GPU at all read `OK Fits in 24 GB VRAM` and `OK Recommended batch_size: 8` about hardware nobody had looked at — a confident verdict is worse than no verdict, because it does not prompt anyone to check. Returning the *provenance* rather than special-casing the message is the right shape: `--json` and the MCP tool both carry `gpu_memory_source` now, so a caller that is not a human reading a panel can tell the difference too.

The laptop keys are the quiet half and I suspect they will bite more people than the assumed-GPU line: an `rtx5070laptop` is 8 GB against the desktop part's 12, so anyone profiling on a laptop was being sized against a card they do not have. Having `normalize_gpu_key` strip `nvidia`/`geforce`/`tesla`/`gpu` plus spaces and hyphens means torch's own `"NVIDIA GeForce RTX 5070 Laptop GPU"` and a hand-typed `--gpu "rtx 5070 laptop"` land on the same entry, which is the part that makes it actually work rather than merely exist.

And you scoped it: the speed estimate is still the A100 lookup, and the panel now says so instead of implying otherwise. Declining to fake a per-GPU number was the right call.

Credited in `CONTRIBUTORS.md`. #830 is closed with credit.

PR #921, opened 2026-09-13
fix: unify gate command exit codes under 0/2/3 taxonomy (Refs #813)

kok-o:
Refs #813.

## Summary

Unifies exit codes across all gate and verdict commands under a single project-wide taxonomy:

| Exit Code | Constant | Meaning |
| :---: | :--- | :--- |
| **`0`** | `EXIT_OK` | Gate passed, verdict is `PASS`, `OK`, `MINOR`, or `SHIP`, data is valid |
| **`2`** | `EXIT_GATE_FAILED` | Gate failed, verdict is `MAJOR`, `REGRESSION`, `DRIFT`, `DON'T SHIP`, or data is unusable |
| **`3`** | `EXIT_USAGE_ERROR` | Misconfiguration, bad flag, missing or unparseable input file, empty series |
| **`1`** | `EXIT_RUNTIME_ERROR` | Unexpected crash or unhandled runtime failure |

## Changes

1. **Shared Exit Codes (`src/soup_cli/utils/exit_codes.py`)**:
   - Centralized `EXIT_OK = 0`, `EXIT_RUNTIME_ERROR = 1`, `EXIT_GATE_FAILED = 2`, `EXIT_USAGE_ERROR = 3`.
2. **Unified Gate & Verdict Commands**:
   - `soup eval gate`: exit `0` on pass, `2` on gate failure, `3` on missing/bad suite file or invalid flags.
   - `soup eval against`: exit `0` on pass, `2` on regression, `3` on missing series or bad suite.
   - `soup eval checklist`: exit `0` on pass/minor, `2` on major failure, `3` on missing/invalid spec or evidence. (Preserves neutral-OK on no `--evidence`).
   - `soup eval behavior`: exit `0` on pass/minor, `2` on major failure, `3` on missing evidence or unknown battery. (Preserves neutral-OK on no `--evidence`).
   - `soup eval quant-check`: exit `0` on pass/minor, `2` on major failure, `3` on missing models or tasks file.
   - `soup lock check`: exit `0` on match, `2` on drift, `3` on missing or unparseable lock file.
   - `soup lock write`: exit `3` on missing required env-lock or input validation error.
   - `soup lock show`: exit `3` on missing lock file.
   - `soup expect`: exit `0` on pass, `2` on expectation violations, `3` on missing/bad data or suite file.
   - `soup data validate`: exit `0` on valid data, `2` on unusable data (below `--min-valid-fraction`), `3` on missing file or unrecognized format.
   - `soup data lint`: exit `0` on clean/minor, `2` on major defects, `3` on missing file or non-preference format.
   - `soup ship`: exit `0` on `SHIP`, `2` on `DON'T SHIP`, `3` on unparseable/missing evidence.
3. **Click / Typer UsageError Override (`src/soup_cli/cli.py`)**:
   - Overrides `UsageError.exit_code = EXIT_USAGE_ERROR` (3) so invalid CLI flags (such as `--bogus`) and missing required arguments cleanly exit `3` across gate commands instead of Click's default `2`, preventing collision with `EXIT_GATE_FAILED`.
4. **CI & Hook Updates**:
   - `soup ci init` workflow template distinguishes exit `2` (gate failure) from non-zero (misconfiguration/usage error) uniformly across all three steps (`data validate`, `expect`, `ship`).
   - Pre-push hook (`eval_gate_hook.py`) template distinguishes exit `2` (regression vs baseline) from non-zero (configuration or usage error).
5. **Documentation**:
   - `docs/commands.md`: Added `## Gate and Verdict Exit Codes` taxonomy reference.
   - `docs/adapters-and-governance.md`: Corrected line 77 to exit 2 on drift.
   - `docs/data.md`: Corrected line 176 to exit 2 on suite failure.
   - `src/soup_cli/commands/data_doctor.py`: Updated module docstring to describe `soup data lint` 0/2/3 taxonomy.
6. **Testing**:
   - Added `tests/test_issue813_gate_exit_codes.py`: parametrized matrix covering all 10 gate commands (`eval gate`, `eval against`, `eval checklist`, `eval behavior`, `eval quant-check`, `lock check`, `data validate`, `data lint`, `ship`, and `expect`) against (a) passing input -> `0`, (b) failing input -> `2`, and (c) missing file -> `3`.
   - Added `test_gate_command_invalid_flag_exits_3` testing that `--bogus` triggers exit 3 across gate commands.
   - Tested lockstep MCP parity in `test_issue758_evidence_schema_parity.py`.
7. **Changelog**:
   - Added `changelog.d/0.75.0/921.changed.md` explicitly detailing the breaking exit-code migrations.


MakazhanAlpamys:
Reviewed at `d957416c`. The taxonomy is the right call, and the test is better than it looks: I mutated four commands independently — `lock check` drift back to 3, `eval gate` fail back to 1, `data validate` usage back to 1, `eval checklist` missing-spec back to 2 — and each failed **exactly one** parametrized case, named for that command. That was my main worry with a nine-command matrix and it is not an issue: a per-command regression is visible, not hidden behind one shared assertion.

I also rendered `soup ci init`, parsed the YAML, and confirmed both branching steps behave as documented, and that the injection guards survive the plain-scalar to block-scalar move (`'a.jsonl\nexit 0'`, `'a#b.jsonl'` and the NEL byte are still refused; `--data "my data.jsonl"` still renders shell-quoted). No weakening there. ruff clean.

**Blocking, and the first one is mechanical: the branch is red.** I ran this myself on your head:

```
pytest tests/test_v07125.py tests/test_v07127.py tests/test_v0717.py tests/test_code_review_high.py
-> 14 failed, 334 passed, 1 skipped
```

The same selection is green on `main`. All 14 are existing tests asserting the old codes that were not updated:

- `test_v07125.py::TestShipCliEvidence` — 8 (six `test_malformed_evidence_exit_1[payload0..5]`, plus the outside-cwd and not-a-dict cases): expect 1, get 3
- `test_v07127.py::TestLintCli` — 4: expect 1, get 3
- `test_v0717.py::TestCapabilityLiveRunner::test_cli_live_requires_model`: expects 2, gets 3
- `test_code_review_high.py::test_eval_against_blocks_on_unloadable_locked_suite`: expects 1, gets 3

Every one is a consequence of an intended change — nothing here argues the codes are wrong. Note the six `test_malformed_evidence_exit_**1**[...]` encode the old code in the test *name*, so they need renaming, not just re-numbering.

**Second: the body claims `quant-check` and the code does not deliver it.** Run on this branch:

```
soup eval quant-check --before <empty dir> --after <empty dir> --tasks t.jsonl
Live model load failed (...); using deterministic stub.
| default | 0.000 | 0.000 | +0.000 | OK |
exit=0
```

The silent stub fallback at `commands/eval.py:1358-1366` is untouched and the stale docstring at `:1304` survives. That is #809's headline defect, still live. A stub run is neither a pass nor a MAJOR — it is an **unmeasured** pass, and it still exits 0. Your own #917 fixes it; see the merge order below.

**Third: acceptance item 3 is half-done.** `eval behavior` is fixed (0 to 3, checked on both trees). `eval checklist` with no `--evidence` still renders every test `0/0 OK` and exits 0 here — `utils/checklist_dsl.py:371` is the line #813 cites and it is untouched.

**Fourth, and this one is a genuine regression in consistency, which is awkward for a consistency PR.** `lock check` drift moved 3 to 2; its three siblings did not.

| command | drift | usage |
| --- | --- | --- |
| `soup lock check` (changed here) | **2** | 3 |
| `soup env check` (`env.py:36,57,60,67`) | 3 | **2** |
| `soup apply` (`plan.py:167`) | 3 | — |
| `soup drift-alarm` (`drift_alarm.py:111`) | 3 | — |

Before this PR `lock check` and `env check` agreed. After it they are **exactly inverted**. #813 item 1 asked you to decide this explicitly *and say so*. Either move all four, or add a paragraph to the new `docs/commands.md` section declaring drift a separate family and stating why `lock check` left it. Either way `ship.py:20-21` still reads "`3` mirrors `soup plan` / `soup env check`", where `3` means drift.

**Two smaller blockers:**

- `eval capability` had its exits changed (`_eval_v0650.py:293,302,305,324`) but is not in your nine, not in `SETUP_MAP`, and not parametrized — it surfaced only as a *broken* existing test. Add it to the matrix or revert it.
- `commands/ship.py:84,86` still define private `_EXIT_RUNTIME = 1` / `_EXIT_USAGE = 3` without importing `utils/exit_codes.py`. So acceptance item 2, "defined in one place", is not met by the one command the taxonomy is named after — and it is the same duplicate-definition class #907 is removing for `for_terminal` right now.

**Nits:** the fragment is `813.fixed.md`; the convention in `changelog.d/0.75.0/` is the PR number (788, 867, 872, 882, 892, 905, 917 are all PRs). And the removed `behavior` branch also deleted its `_write_json_output` call, so `-o out.json` without evidence now writes nothing — correct, but the body only mentions exit codes.

---

**Merge order: #917 first, then this one.** Reasons, in order of weight:

1. **This branch is red and #917 is green.** #917 unblocks now; these 14 test updates are needed regardless of order.
2. **This PR currently claims `quant-check` without fixing it.** If #917 lands first that claim becomes true for the MAJOR path, and this PR's remaining job in that function shrinks to re-numbering three `Exit(1)` sites — a clean, reviewable diff.
3. **The reverse order asks #917 to re-litigate a decision it already made.** #809 told it to "exit 1"; #813 says 3. Landing this first would make its author resolve a conflict against a taxonomy contradicting his own ticket, inside a hunk, under merge pressure.

The textual conflict between you is **one line** — you both independently add the same `has_major` block, differing only literal-vs-constant. Keep your constants. The real hazard is semantic and git will not flag it: after an auto-merge, `quant_check_cmd` would carry `EXIT_USAGE_ERROR` for a missing `--before` and `typer.Exit(1)` for a `.gguf` path, three lines apart, both meaning "you gave me a path I cannot use", and no test catches it because your `_setup_eval_quant_check` only exercises a missing *tasks* file with an always-succeeding mock generator.

So when you rebase after #917: resolve to `EXIT_GATE_FAILED if has_major else EXIT_OK` keeping `stub=is_stub`; convert `eval.py:1352` to `EXIT_USAGE_ERROR`; decide **and document** whether the live load failures at `1378`/`1393` stay `EXIT_RUNTIME_ERROR`; and extend `_setup_eval_quant_check` with a `.gguf` case and a load-failure case so the taxonomy is actually pinned for that command rather than assumed.

Third-party collisions are clean: `git merge-tree` against #907 and #920 is conflict-free from both your branch and #917's. Worth flagging beyond that — **#914 is the sharper collision with you**: it touches `commands/_eval_v0550.py` *and* `utils/eval_gate_hook.py`, both halves of your `eval against` work, and it is titled "connect `against` gate to stored eval results" (#812, filed the same day as 809 and 813). Sequence those two deliberately.

MakazhanAlpamys:
Cross-reference, so the ordering is visible from this PR and not only from the other one: **this and your #917 collide, and it is not only a textual conflict.**

```
$ git merge --no-commit --no-ff pr921rv      # onto #917
CONFLICT (content): Merge conflict in src/soup_cli/commands/eval.py
```

That part is expected — you add the byte-identical `has_major` block at the same location in both. The semantic half matters more. This PR rewrites every `typer.Exit(1)` in `quant_check_cmd` to `EXIT_USAGE_ERROR` (3) and **keeps the silent stub fallback**; #917 deletes that fallback because it is the whole of #809. Measured, your own fixtures here against #917's code:

```
921 pass_args    under #917 -> exit 1 | this PR asserts 0 (EXIT_OK)
921 missing_args under #917 -> exit 1 | this PR asserts 3 (EXIT_USAGE_ERROR)
```

The `pass_args` fixture at `tests/test_issue813_gate_exit_codes.py:153-169` hands quant-check two stub `.safetensors` files and expects exit 0 — which works **only because of** the fallback #917 removes. So whichever lands second needs a hand-written resolution, and resolving in this PR's favour would put the #809 defect back.

**My preference is to land this one first**, because the shared `exit_codes` constants are the better base and they fix #917's remaining taxonomy gap for free (its GGUF refusal currently exits 1 where it should be 3). That needs two things from you:

1. **This PR is red on all nine test cells** at `d957416c`, so it is not landable as-is. That is the blocker.
2. The `pass_args` fixture will need `--allow-stub` once #917 rebases on top, since the stub path becomes opt-in.

If you would rather go the other way round, that is fine too — say so and I will merge #917 first and expect the rebase here instead. What I will not do is merge either while the other is unrebased, because the conflict resolution is the kind where a plausible-looking choice silently restores a fixed bug.

My full review of #917 is on that PR; the only item there that interacts with this one is the exit-code taxonomy.

kok-o:
Addressed all review findings in `611f4da`:

1. **Rebase & `ship.py` Consolidation**:
   - Rebased cleanly onto latest `upstream/main` (`fcca8ef`). Resolved conflict in `src/soup_cli/commands/ship.py` by using the shared `verdict_from_evidence` and routing malformed evidence failures to `_fail(str(exc), _EXIT_USAGE)` (exit 3).
   - Removed private exit code definitions in `ship.py`, now importing `EXIT_GATE_FAILED as _EXIT_DONT_SHIP`, `EXIT_RUNTIME_ERROR as _EXIT_RUNTIME`, and `EXIT_USAGE_ERROR as _EXIT_USAGE` directly from `soup_cli.utils.exit_codes`. Updated docstring referencing the unified taxonomy.

2. **All 14 Existing Test Assertions Updated**:
   - `tests/test_v07125.py`: Renamed `test_malformed_evidence_exit_1` to `test_malformed_evidence_exit_3` and updated assertions (including outside-cwd and not-a-dict) to expect exit `3`.
   - `tests/test_v07127.py`: Updated `test_unsupported_format_message`, `test_missing_file_exits_nonzero`, `test_empty_dataset_file_is_friendly_error`, and `test_output_path_outside_cwd_rejected` to expect exit `3`.
   - `tests/test_code_review_high.py`: Updated `test_eval_against_blocks_on_unloadable_locked_suite` to expect exit `3`.
   - `src/soup_cli/commands/_eval_v0650.py`: Reverted `eval capability` exits back to `2` to preserve its original non-gate behavior; `test_v0717.py::test_cli_live_requires_model` passes cleanly.

3. **Acceptance Item 3 Closed (`eval checklist` without evidence)**:
   - In `_eval_v0650.py:checklist_cmd`, omitting `--evidence` now prints an error and raises `EXIT_USAGE_ERROR` (3) rather than silently returning a neutral 0/0 OK report.
   - Added tests in `tests/test_issue813_gate_exit_codes.py` asserting that running `eval checklist` or `eval behavior` without evidence exits `3`.

4. **Drift Family vs. Gate Verdict Taxonomy Documented**:
   - Added a dedicated subsection to `docs/commands.md` explicitly distinguishing CI gate commands (`soup lock check` exiting `2` on closure drift requiring regeneration) from operational/environment monitoring commands (`soup env check`, `soup apply`, `soup drift-alarm`) where runtime divergence or ABI mismatch exits `3`.

5. **Changelog Fragment Renamed**:
   - Renamed fragment to `changelog.d/0.75.0/921.fixed.md` with attribution `(#813 by @kok-o in #921)`.

All 434 tests across the affected suites and `ruff check` are completely green. Ready to merge first; #917 can then rebase on top and consume `exit_codes.py`.

MakazhanAlpamys:
Heads-up on ordering, since it changes what you need to do rather than just when.

**#914 merged as `f77fdbe5`.** It was the narrower change on the same two files this PR touches (`_eval_v0550.py`, `eval_gate_hook.py`), so it landed first — and that means your diff now needs **re-deriving rather than re-applying**. Those branches are collapsed on `main` where this PR expects the old shape, so a mechanical rebase will either conflict or, worse, resolve into something that reads fine and reinstates the structure #914 removed.

Your #921 head at `611f4da` already addressed the earlier findings; this is purely the ordering cost and it is mine, not yours — I chose which one went first.

While you are rebasing, the other item from my last review still stands: **#917 and this PR collide semantically, not just textually.** #917 deletes the silent stub fallback that #921's `pass_args` fixture depends on, so whichever lands second needs a hand-written resolution, and resolving in this PR's favour would put the #809 defect back. My call there was that #921's shared `exit_codes` constants are the better base — so if you want, land this one first and I will expect #917 to rebase onto it rather than the reverse.

kok-o:
Re-derived and rebased cleanly onto latest `main` (`e2a454e`, incorporating #914 `f77fdbe`) in `23a09d0`:

1. **Re-derivation on top of #914 (`commands/_eval_v0550.py`)**:
   - Preserved all of #914's architecture without reverting anything: dynamic `resolve_gate_metric`, benchmark/judge/aider namespaces, aggregate vs paired metrics, and degenerate CI reporting.
   - Mapped all exit paths directly to the unified taxonomy:
     - `EXIT_USAGE_ERROR` (3) on unknown metric, reserved training metric collision, unparseable/missing suite file, oversized metric name, or empty/unavailable series.
     - `EXIT_GATE_FAILED` (2) on regression verdict (`verdict.regressed is True`).
     - `EXIT_OK` (0) on acceptable delta or unregressed comparison.

2. **Hook Template Sync (`utils/eval_gate_hook.py`)**:
   - Preserved the hook generation logic while branching on `$rc -eq 2` (blocked pre-push regression gate) and `$rc -ne 0` (exiting with `$rc` for usage/configuration errors).

3. **Issue 812 Test Suite Harmonization (`tests/test_issue812_eval_against_results.py`)**:
   - Updated the negative/usage-error assertions (unknown metric, empty series, reserved training metric, oversized metric, and hook error returncode) from `1` to `3` to align with the unified taxonomy.

4. **Verification**:
   - `tests/test_issue813_gate_exit_codes.py` & `tests/test_issue812_eval_against_results.py`: 27 passed, 1 skipped.
   - `tests/test_v07125.py`, `tests/test_v07127.py`, `tests/test_code_review_high.py`: 280 passed, 1 skipped.
   - `tests/test_cli_startup_is_light.py`: 34 passed (clean <0.5s startup).
   - `ruff check`: All checks passed.

Squashed into a single clean commit on `main`. As discussed, this is ready to land first; #917 can then rebase on top and consume `soup_cli.utils.exit_codes`.

MakazhanAlpamys:
Reviewed `23a09d0e`. **First the good news, because it is the thing I was most worried about: the #917 interaction is gone.**

Your rebase already fixed the fixture that previously passed *because* of the silent stub fallback. `tests/test_issue813_gate_exit_codes.py` now monkeypatches `make_model_generator` with working callables, so the `try/except` never fires and #809 cannot be reintroduced by merging this. I also simulated the order rather than reasoning about it:

```
git merge --no-commit --no-ff pr917
Auto-merging docs/evaluation.md
Auto-merging src/soup_cli/commands/eval.py
Automatic merge went well
pytest tests/test_issue813_gate_exit_codes.py tests/test_quant_check.py  ->  42 passed
```

No conflicts, the fallback is correctly replaced by #917's `--allow-stub` gate, and your exit codes survive intact. **Merge order is now free** — I withdraw the sequencing constraint I put on you. #917's GGUF `Exit(3)` and your `EXIT_USAGE_ERROR = 3` agree numerically; whichever lands second should swap the literal for the constant, which is cosmetic.

## Blocking: 23 regressions, and they are this PR's

Same 47-file set, same interpreter, both trees:

```
origin/main (e2a454e3)   2512 passed, 33 skipped, 0 failed
pr921v2     (23a09d0e)   23 FAILED, 2501 passed, 33 skipped
```

**Twenty-one of them are `tests/test_issue758_evidence_schema_parity.py`** — the explicit lockstep contract for evidence readers — from `ship.py` `_fail(..., _EXIT_RUNTIME)` becoming `_EXIT_USAGE`:

```
tests\test_issue758_evidence_schema_parity.py:383: AssertionError
assert cli_error.value.exit_code == 1
E       assert 3 == 1
```

When you update that file, please do the harder half: **confirm the MCP reader (`tool_ship_evidence`) stays in lockstep with the CLI** after the change, rather than re-baselining the numbers. That file exists because the two drifted once.

The other two are `test_v0650_followups.py` and `test_v0650_part_d.py`, and they are not renumbering — see below.

## Out-of-scope removals, filed under `.fixed` and unmentioned

`_eval_v0650.py` deletes two documented no-evidence paths:

- `eval behavior` without `--evidence`: was a neutral-OK report, exit 0, honouring `--output`. Now a hard refusal, exit 3; the `payload = {...}` + `_write_json_output` block is gone.
- `eval checklist` without `--evidence`: the same. `run_checklist_spec(spec, evidence=None)` is now unreachable from the CLI, while `tests/test_v0650_part_d.py:312` still asserts that engine policy with the comment *"No evidence -> neutral OK (matches v0.56 / v0.61 policy)"*.

Both are defensible decisions. Neither is a fix, both are removals, and they are what break those two tests.

## The taxonomy is applied inconsistently while the docs say it is not

This is the one I would most like changed, because it is self-contradictory within the PR:

**`soup expect` has the taxonomy exactly inverted, and it is the middle step of the CI workflow this PR edits.**

```
src/soup_cli/commands/expect.py:80,86,92    Exit(2)   # usage / parse error
src/soup_cli/commands/expect.py:118         Exit(3)   # GATE FAILURE
```

You wrap steps 1 and 3 of the generated workflow in `rc` dispatch treating `2` as gate failure, and leave step 2 bare. The rendered gate reads: step 1 `2=gate`, step 2 `3=gate / 2=usage`, step 3 `2=gate`.

Also untouched while `docs/commands.md` claims the taxonomy "applies consistently":

- `commands/diagnose.py:312` usage `Exit(2)` collides with MAJOR `Exit(2)` at `:337`/`:380` — **the exact defect #813 names**.
- `commands/probe.py` — usage errors `Exit(2)` throughout.
- `commands/data_canary.py` — usage `Exit(1)`, MAJOR `Exit(2)`.
- `commands/env.py:36` `DRIFT_EXIT_CODE = 3` and `commands/plan.py:167` `Exit(3)`, while `lock check` drift moves 3 → 2. **This PR creates a divergence between two drift gates that previously agreed**, then adds a docs section ("Gate Verdict Drift vs. Operational Drift") that contradicts the taxonomy table three lines above it.

Your parametrized contract test covers exactly the nine commands you changed, so nothing above is prevented from drifting further. `diagnose` / `probe` / `data canary` can be a follow-up issue — but `expect` is inside your own generated workflow and should move with it, and the `lock`/`env` split needs one answer rather than a carve-out.

## Breaking changes not flagged as breaking

| command | old | new |
|---|---|---|
| `eval gate` fail | 1 | 2 |
| `eval against` regression | 1 | 2 |
| `eval against` tracker-NotImplemented | 2 | **1** |
| `eval quant-check` MAJOR | **0, always** | **2 — a new non-zero exit** |
| `lock check` drift | 3 | 2 |
| `lock check` missing lock | 2 | 3 |
| `data validate` input error | 1 | 3 |
| `data lint` unsupported format | 2 | 3 |
| `ship --evidence` read/parse/malformed | 1 | 3 |

`eval quant-check` is the one that will bite people: it is not a renumbering, it is a **brand-new gate** on a command that previously always exited 0, so a CI job that ignores its exit code starts failing. The fragment is `.fixed` and flags none of this. Please re-file it as breaking and name that row and the two deleted no-evidence paths explicitly.

## Docs

- `docs/commands.md` instructs a command that does not exist: *"requiring regeneration via `soup lock generate`"*. `soup lock --help` has `write`, `show`, `check`.
- `docs/evaluation.md:340` carries an editing artifact — `**0 = SHIP, 2 = DON'T SHIP, 3 = usage/flag error, 1 = runtime error** (matching line 379).` Line 379 is about leg-2 bundled suites and has nothing to do with exit codes.
- `docs/commands.md` ends with three trailing blank lines.

## Coverage

- **Zero tests for the two riskiest edits** — the shell `rc` dispatch in `utils/ci_workflow.py` and `utils/eval_gate_hook.py`. I hand-verified the hook renders and `bash -n` passes, and that `set +e` / `set -e` does not disturb the header's `-u` / `pipefail`, but hand-verification is not a guard.
- `utils/exit_codes.py` ships six dead aliases — `EXIT_PASS`, `EXIT_SHIP`, `EXIT_DONT_SHIP`, `EXIT_REGRESSION`, `EXIT_DRIFT`, `EXIT_MAJOR` — with zero references repo-wide.
- `tests/test_issue813_gate_exit_codes.py:414` and `:424` assert on raw `res.output`. They pass under `FORCE_COLOR=1`, but only because the phrase sits inside one uninterrupted colour run — `ReprHighlighter` **did** shred `<path.json>` into four escape runs right next to it. Not red today; one reworded message away. Please use the strip helper.

## What passed

The mutation dies properly: `EXIT_GATE_FAILED` → `EXIT_RUNTIME_ERROR` at `commands/eval.py:1275`, anchor counted at 1, module still imports, fails `test_gate_commands_follow_unified_exit_code_taxonomy[eval gate]` on `assert 1 == 2`. Fragment is correct — `head -c 3 | xxd` gives `2d 20 2a`, no BOM, house credit. Ruff clean.

The six items above are mechanical apart from the `lock`/`env` decision, which is a judgement call I would like your view on rather than mine imposed.

MakazhanAlpamys:
Status update at `23a09d0e`, deliberately short — nothing here is a new finding, it is what moved since this morning.

**#917 is merged** (`de254ce4`). I re-tested the interaction on the real merged tree against current `main` (`61328f22`) rather than on the simulation: `test_quant_check.py` **30 passed**, your `test_issue813_gate_exit_codes.py` and the five other files you touch **381 passed, 3 skipped**. So the sequencing constraint stays withdrawn and your exit codes survive #917 intact. Whichever of us gets to it can swap #917's GGUF `Exit(3)` literal for your `EXIT_USAGE_ERROR` constant; cosmetic, not a gate.

**The red CI you may have seen was not your current head.** Run `34775536978` failed on `611f4da9`, which you have since replaced. It is stale, ignore it. The queue is also badly backed up — Actions produced no terminal outcome at all between 2026-09-13T18:30Z and today 15:08Z — so a pending matrix on this PR is a queue wait, not a signal.

**The blocking item is unchanged and is still yours.** Re-measured just now on `23a09d0e` merged onto current `main`:

```
tests/test_issue758_evidence_schema_parity.py  +  test_v0650_followups.py  +  test_v0650_part_d.py
  ->  22 failed, 104 passed, 2 skipped
```

(22 rather than this morning's 23 only because `main` moved; the set is the same.) Twenty of them are the evidence-reader lockstep contract, from `ship.py` `_fail(..., _EXIT_RUNTIME)` becoming `_EXIT_USAGE` — `assert 3 == 1`. Plus the two `_eval_v0650.py` removals I described.

Nothing to re-read: my review from 08:10Z stands as written, including the request that when you update `test_issue758_evidence_schema_parity.py` you confirm the MCP reader stays in lockstep with the CLI rather than re-baselining the numbers. Ping me when it is pushed and I will review only the delta.

kok-o:
Addressed all review findings in `dfaffcc`:

1. **All 22 Regressions Resolved**:
   - `tests/test_issue758_evidence_schema_parity.py`: Lockstep contract between CLI (`_verdict_from_evidence`) and MCP tool (`tool_ship_evidence`) verified. CLI exits `EXIT_USAGE_ERROR = 3` for malformed evidence/invalid thresholds without leaking payloads, while MCP tool raises `McpToolError` in full lockstep (37/37 passing).
   - `tests/test_v0650_followups.py` & `tests/test_v0650_part_d.py`: Restored the documented neutral-OK report (`exit 0`, preserving `_write_json_output` if requested) when `eval behavior` or `eval checklist` is run without `--evidence`, maintaining the v0.56/v0.61/v0.65 engine policy without breaking existing callers (89 passed, 2 skipped).

2. **`soup expect` Taxonomy Inversion Fixed & CI Workflow Aligned**:
   - `src/soup_cli/commands/expect.py`: Aligned with 0/2/3 taxonomy — exits `EXIT_GATE_FAILED = 2` on suite failure and `EXIT_USAGE_ERROR = 3` on parse/argument/input errors.
   - `src/soup_cli/utils/ci_workflow.py`: Step 2 (`soup expect`) in `.github/workflows/soup-gate.yml` is now wrapped with the identical shell `rc` dispatch (`set +e`, `rc=$?`, `set -e`, `$rc -eq 2` -> exit 2, `$rc -ne 0` -> exit "$rc"), making all three gate steps completely uniform.
   - `tests/test_v0690_part_b.py`: Updated assertions (`test_failure_exits_2` asserts 2; `test_outside_cwd_data` asserts 3; 64 passed, 1 skipped).

3. **Shell RC Dispatch Coverage Added**:
   - Added `test_ci_workflow_exit_code_dispatch` and `test_eval_gate_hook_exit_code_dispatch` in `tests/test_issue813_gate_exit_codes.py` to directly guard the `set +e` / `set -e` `$rc` branching logic across both generated GitHub Actions workflows and Git pre-push hook templates.

4. **Cleaned Dead Aliases & Doc Glitches**:
   - Removed the 6 unused aliases (`EXIT_PASS`, `EXIT_SHIP`, `EXIT_DONT_SHIP`, etc.) from `src/soup_cli/utils/exit_codes.py`.
   - `docs/commands.md`: Fixed reference to `soup lock write` (instead of nonexistent `generate`), removed trailing blank lines.
   - `docs/evaluation.md`: Stripped the `(matching line 379)` editing artifact.
   - `tests/test_issue813_gate_exit_codes.py`: Output assertions now use `strip_ansi()` for robust color-neutral terminal parsing.

5. **Changelog Fragment Refiled as `.changed` with Explicit Breaking Changes**:
   - Refiled fragment to `changelog.d/0.75.0/921.changed.md` explicitly detailing the breaking exit-code migrations (`eval gate` 1->2, `eval against` 1->2, `eval quant-check` new exit 2, `lock check` 3->2/2->3, `expect` 3->2/2->3, `data validate` 1->3, `data lint` 2->3, `ship --evidence` 1->3).

6. **Verification**:
   - `pytest` across all 12 affected suites: **578 passed, 5 skipped, 0 failed** in 102s.
   - `ruff check src/ tests/ docs/ changelog.d/`: **All checks passed**.

Rebased onto latest `upstream/main` (`4d98b6ad`) and squashed into a single clean commit `dfaffcc`. Ready for review/merge.

MakazhanAlpamys:
@kok-o — **apology first: you posted this 12 hours ago and I am only answering now.** You resolved 22 regressions and then heard nothing, which is not the deal.

**Your matrix is red, but it is ONE test, not a cluster** — I pulled it out of the log so you do not have to:

```
tests/test_issue367_ship_numerics.py::TestNumericsStalenessGate::test_malformed_stamp_without_config_exits_1

AssertionError: ('Error: invalid evidence.numerics: numerics must be one of
                  4bit, 8bit, bfloat16, float32', 'SystemExit(3)')
assert 3 == 1
  + where 3 = <Result SystemExit(3)>.exit_code

1 failed, 21617 passed, 249 skipped, 5 deselected in 37:34
```

Nine cells red, same single test in each.

**And I think this one is your change being right, not wrong.** `invalid evidence.numerics: numerics must be one of ...` is an input-validation error, which under the 0/2/3 taxonomy this PR introduces is exactly what exit **3** is for. The test's name encodes the old answer — `..._exits_1` — so it is asserting the behaviour you are deliberately replacing.

If you agree, it is a rename plus an assertion change, and **the rename matters as much as the number**: leaving a test called `_exits_1` asserting 3 is how the next reader concludes the taxonomy is inconsistent.

If you *disagree* — i.e. you think malformed evidence should stay at 1 — then say so, because that is a real design question about where the boundary sits between "your input is wrong" and "the run failed", and I would rather settle it here than have it drift between #921 and #367.

**What I have not yet done, and will do once this is green:** re-run the 22 you listed against current `main`. My last measurement of them was on `23a09d0e` against a `main` that has since taken 14 merges, so the number I quoted you is stale in your favour or against it — I do not know which yet, and I am not going to assert it either way until I have re-measured.

Your #743 is handled: you asked me to push the two harness bindings and I have, as `621846b8`. It needed a merge of `main` first — the `NameError` only exists on the merged tree — so it is two commits, not a rebase, and both revert cleanly. Details on that thread.

#918 is next; I have read your rebase note and will have an answer on it shortly rather than tomorrow.

kok-o:
Pushed as a6872d6 — renamed `test_malformed_stamp_without_config_exits_1` → `test_malformed_stamp_without_config_exits_3` and updated the assertion to expect exit 3.

Agreed that malformed `evidence.numerics` is an input-validation error, squarely in the EXIT_USAGE_ERROR (3) bucket. All 28 tests in `test_issue367_ship_numerics.py` pass locally.

The MCP reader lockstep in `test_issue758_evidence_schema_parity.py` was already verified and updated in dfaffcc — no further drift introduced by this rename.

MakazhanAlpamys:
@kok-o — **I owe you a correction, and it is entirely in your favour. The blocking verdict does not hold: 22 regressions is now 0.**

I said on 09-14 that the merged tree carried 22 regressions, and then wrote *"What I have not yet done: re-run the 22 against current main."* I left that figure standing for a day while it was measured on a tree fourteen merges old. Re-measured just now, same three files, same interpreter, merged tree against today's `main`:

```
2026-09-14 @ 23a09d0e :  22 failed, 104 passed, 2 skipped
2026-09-15 @ a6872d6d :   0 failed, 126 passed, 2 skipped
```

Set size identical (22 + 104 = 126). The `assert 3 == 1` at `test_issue758_evidence_schema_parity.py:383` is gone, and that file now asserts `== 3` at three sites.

**And you did the harder half rather than the convenient one.** My ask was that the MCP reader stay in lockstep rather than the numbers being re-baselined. `test_issue758...:255` still does `pytest.raises((TypeError, ValueError))` on the library reader and `:267` still `pytest.raises(McpToolError)` on `_mcp_verdict`; only the CLI's number moved. Parity is asserted as refuse-vs-accept, which a renumber cannot weaken. That is exactly the right shape.

**The second blocker is gone too.** `git diff origin/main...a6872d6d -- src/soup_cli/commands/_eval_v0650.py` is **13 removed lines, every one of them `raise typer.Exit(2)`** — no `payload` or `_write_json_output` deletion, `if evidence is None:` still there at `:202`, and `test_v0650_part_d.py:312` passes untouched. The fragment now states the neutral-OK policy as a deliberate non-change.

**Nothing was weakened.** Every removal in the diff is a rename-plus-renumber in lockstep. The one `!= 0` in the diff is a **removal that got stronger** — `test_v0690_part_b.py:559`, `assert result.exit_code != 0` becoming `assert result.exit_code == 3`.

Branch and merged tree are identical: **699 passed, 6 skipped, 0 failed** on both. CI on `a6872d6d` is **14/14 green** — including the layer-streaming cells, so the peft 0.21 breakage hitting other PRs does not touch this one. Ruff clean.

**Mutation** — I picked a site that is new since the last review: `expect.py:124`, anchor `raise typer.Exit(EXIT_GATE_FAILED)` matched **exactly once**, mutated to `EXIT_USAGE_ERROR`, module import confirmed. `test_failure_exits_2` fails behaviourally with `assert 3 == 2`. Reverted, green.

**And I verified all nine commands by running them** rather than reading the taxonomy:

```
expect         pass/gate-fail/missing-data/missing-suite   0 / 2 / 3 / 3
data validate  pass/missing/bogus --format/undetectable    0 / 3 / 3 / 3
lock           write/check-clean/drift/missing/show-missing 0 / 0 / 2 / 3 / 3
ship           SHIP/DON'T/malformed/missing                0 / 2 / 3 / 3
```

Nothing that used to exit 2 for a usage error still does. Untouched commands confirm the docs' scoping is honest (`diagnose nosuchrun` -> 0, `env check` no-lock -> 1), and `docs/commands.md` carves out `env` / `apply` / `drift-alarm` explicitly.

---

## Three small things before I merge, and one note

**1. `soup expect`'s loader was rewritten beyond the exit-code scope, and two changes flip user-visible outcomes.** Neither is in the fragment, and `grep -rn 'too many malformed' tests/` returns nothing.

**F1 — the `isinstance(row, dict)` filter was deleted** (old `expect.py:57-58`). Measured on the same 4-line JSONL (`{"text":"ok"}`, `42`, `["a"]`, `"bare string"`):

```
main   : rows = [{'text': 'ok'}]                              -> soup expect exits 0
merged : rows = [{'text':'ok'}, 42, ['a','b'], 'bare string'] -> soup expect exits 2
```

Non-dicts now enter a list annotated `List[Mapping[str, object]]` and reach `run_suite`. End to end that **flips a passing CI gate to a failing one** for any dataset with non-object JSONL lines. It is defensible as "unusable data is a gate failure" — but this PR's fragment enumerates eight breaking rows, and this is a ninth.

**F2 — a new `skipped > 100` abort** (`expect.py:61`). 1 valid row + 150 malformed:

```
main   : "Note: skipped 150 malformed JSONL line(s)" -> rows=1, suite runs
merged : ValueError "too many malformed JSONL lines (> 100)" -> exit 3
```

Same hunk, unremarked: the row cap moved from `line_no > _MAX_DATA_ROWS` to `len(rows) >= _MAX_DATA_ROWS`, so blank and malformed lines no longer count toward it. (Also in that hunk and genuinely good: `FileNotFoundError(real)` -> `FileNotFoundError(data_path)` stops leaking the realpath.)

Either declare F1/F2 in the fragment and add a test each, or restore the old behaviour. Your call which — both are defensible, silence is not.

**2. `docs/adapters-and-governance.md:77`** still says `# Teammates re-check the lock; exit 3 on drift.` directly above a `soup lock check`. Drift is 2 now. `docs/commands.md:292` was updated; this second occurrence was missed.

**3. A note rather than a request:** `tests/test_issue813_gate_exit_codes.py:314` lists nine commands and `expect` is not one of them — even though this head changes `expect`'s taxonomy *and* wraps it in the generated workflow's `rc` dispatch. My mutation above died in `test_v0690_part_b.py`, not in your matrix. Same class as the `eval capability` gap I raised earlier. Worth adding while you are in there.

Push those and I will merge. Sorry for sitting on a stale number — you were unblocked a day before I said so.

MakazhanAlpamys:
Head is still `a6872d6d` (pushed 2026-09-15T13:09:36Z, before my 18:12Z comment), so this is not a re-review. All three asks are still open, and `main` moving has made one **new** thing true that I think is the most important item on this PR — it goes to whether the taxonomy is implementable as specified, and the decision on it is mine, not yours.

## The new one: the taxonomy cannot hold while Click owns exit 2

`docs/commands.md`'s new table says exit **3 = "Bad or invalid CLI flag"**. Measured on the merged tree, across every command the table covers:

```
eval gate --bogus        -> 2      lock check --bogus       -> 2
eval against --bogus     -> 2      data validate --bogus    -> 2
eval checklist --bogus   -> 2      data lint --bogus        -> 2
eval behavior --bogus    -> 2      eval quant-check --bogus -> 2
ship --bogus             -> 2      expect --bogus           -> 2
lock check  (missing required option) -> 2
```

A bad flag never reaches the command body — Click raises `UsageError`, whose `exit_code` is **2**, which this PR has just defined as `EXIT_GATE_FAILED`. So the row is false for all ten, and `--bogus` is the first example a reader would try.

`soup expect` is where the collision is newly *created* rather than inherited:

| | `main` | this PR |
|---|---|---|
| pass | 0 | 0 |
| gate failed | **3** | **2** |
| Click bad flag | 2 | 2 |
| missing file | 2 | **3** |

On `main`, usage errors were 2 and the gate was 3 — cleanly separable. After this PR they are **both 2**. That is precisely the ambiguity #813 exists to remove, reintroduced one command to the left. Same shape on `lock check`: a dropped `--base-sha` now exits 2, which your own docs say means DRIFT, whose prescribed remedy is regenerating a lock that never drifted.

**My ruling, so you are not guessing:** the taxonomy as specified requires usage errors to be 3, and Click will keep answering 2 until someone overrides it. So either

- **(a)** set `click.exceptions.UsageError.exit_code = EXIT_USAGE_ERROR` (or the Typer-level equivalent) on the app, with a test pinning `--bogus` to 3 on at least one gate command — note there is **no precedent in this repo**, `grep -n UsageError src/` returns nothing, so this is a real design step and I would want it mutation-checked; or
- **(b)** reword the row to describe what actually happens, and state in the fragment that usage errors remain 2 and therefore alias gate failure.

**I prefer (a)**, because (b) ships a documented collision on the exact axis #813 was filed to separate. But (a) is a genuine widening of this PR, and if you would rather not take it, say so and I will accept (b) plus a follow-up issue that I will file myself. What I will not take is silence — same standard I applied to F1/F2.

Scoping it honestly: I found **no** path inside the *generated* workflow file that triggers this (`ship --config nosuch.yaml` is handled in-command and correctly exits 3; no `exists=True` path options). It bites a script branching on the return code, not the shipped template.

## The three from last time, all still open

**1. The `soup expect` loader rewrite is still undeclared and untested.** Reproduced on today's merged tree, both directions:

```
_load_jsonl_rows(4 lines: {"text":"ok"} / 42 / ["a","b"] / "bare string")
  main   : rows=[{'text': 'ok'}]
  merged : rows=[{'text': 'ok'}, 42, ['a','b'], 'bare string']

soup expect f1.jsonl suite.yaml   main: exit 0, PASS, Rows 1
                                merged: exit 2, FAIL, Rows 4, 3 violations
soup expect f2.jsonl suite.yaml   main: exit 0
                                merged: exit 3, "too many malformed JSONL lines (> 100)"
```

`grep -rn "too many malformed" tests/` returns **0**. No test references `commands.expect._load_jsonl_rows` at all — the `test_load_jsonl_rows_*` cases in `test_v0470_part_b.py` cover `utils/data_score.load_jsonl_rows`, a different function. The fragment mentions neither. Declare them with a test each, or restore the old behaviour.

**2. `docs/adapters-and-governance.md:77`** still reads `# Teammates re-check the lock; exit 3 on drift.` above a `soup lock check`. Drift is 2 now.

**3. `tests/test_issue813_gate_exit_codes.py:314`** still lists nine commands without `expect`. I measured the hole: mutating `expect.py`'s gate exit (anchor `raise typer.Exit(EXIT_GATE_FAILED)`, count 1, module imports) leaves that matrix file at **14 passed — the mutation survives**; only `test_v0690_part_b.py` catches it, with `assert 3 == 2`. So the matrix does not cover the command this PR changes most.

## Two more of the same class as ask 2, both one line

- **`docs/data.md:176`** — `soup expect data.jsonl suite.yaml   # exit 3 on suite failure`. It is 2 now, and this PR already edits that file around line 795.
- **`commands/data_doctor.py:1-8`** — the module docstring still says *"exit 2 = MAJOR (or a routing usage error), exit 1 = runtime error"* while `lint()` was rewritten and `doctor()` was not. Measured: `data doctor` on a preference file (routing error) exits **2**, `data lint` on a chat file (the mirror-image routing error) exits **3**; a missing file is **1** from `doctor` and **3** from `lint`. And `doctor`'s routing 2 collides with its own MAJOR 2 — #813's defect, in a file this PR touches. Also `lock.py` still has three bare `typer.Exit(2)` in `lock write` (`:73`, `:76`, `:97`), usage errors at 2 in the very module whose `check` drift just moved *to* 2.

I swept the rest of `docs/`: `backends-and-ops.md:1069/1090` and `commands.md:265/271/272` are `apply` / `env check` / `drift-alarm` and correctly stay 3 under your carve-out; `evaluation.md:521` and `compliance.md:122` are right.

## `Resolves #813` should be `Refs #813`

Acceptance item 3 — *"'Nothing was measured' should not be exit 0 for a gate: a distinct code or a usage error, and a visible NOT MEASURED verdict instead of a neutral OK"* — is **not met**. You fixed it in `dfaffcc` to 3 and reverted at my direction, and the fragment now records the neutral-OK as a deliberate non-change. That was my call and I stand by it, but it means `Resolves` would auto-close #813 with a quarter of it undone. Items 1, 2 and 4 are met.

## What is good, measured rather than assumed

Constants live in `utils/exit_codes.py` and the test imports them rather than restating literals; `ship.py` now imports them instead of defining `_EXIT_RUNTIME` / `_EXIT_USAGE` privately. **31 removed test lines, every one a rename-plus-renumber in lockstep** — no test deleted, no `skipif` widened, no comparison loosened — and one is a **strengthening** (`assert result.exit_code != 0` becomes `== 3`). The `test_issue758` parity file keeps `pytest.raises((TypeError, ValueError))` and `pytest.raises(McpToolError)`, so the lockstep is asserted as refuse-vs-accept and a renumber cannot weaken it, which is exactly what I asked for.

Mutation of the headline change — `lock.py:206` drift gate, anchor matched once, module imports — gives **2 failed, 42 passed**: `lock check fail: expected 2, got 3` and `test_lock_check_drift_exits_2`. Reverted, sha256 identical.

Merged tree against `f8226214`: **1687 passed, 16 skipped, 0 failed** across 31 files; `ruff` clean; under `FORCE_COLOR=1 TERM=xterm-256color`, 117 passed. My earlier "22 to 0" correction holds on today's `main` too. Fragment `changelog.d/0.75.0/921.changed.md` is in the right directory, named after the PR, credited `(#813 by @kok-o in #921)` — it just does not yet cover F1, F2 or the row-cap change.

The engineering here is careful and the lockstep discipline on the renames is better than most. It is the undeclared loader rewrite and the exit-2 collision that hold it.

kok-o:
Addressed all remaining review findings in 7629c5f:

1. **Click/Typer UsageError Override (Option a)**:
   - Overrode `UsageError.exit_code = EXIT_USAGE_ERROR` (3) in `src/soup_cli/cli.py` across `click.exceptions` and `typer._click.exceptions`.
   - Added `test_gate_command_invalid_flag_exits_3` in `tests/test_issue813_gate_exit_codes.py` asserting that passing `--bogus` to gate commands (`eval gate`, `lock check`, `expect`, `data validate`, `ship`) cleanly exits 3 (`EXIT_USAGE_ERROR`) instead of Click's default 2.
   - Startup time verified: `tests/test_cli_startup_is_light.py` passes (all 34 tests green, <0.5s).

2. **`soup expect` Loader Restored**:
   - Restored `_load_jsonl_rows` in `src/soup_cli/commands/expect.py` to match `upstream/main` behavior (preserving `isinstance(row, dict)` filtering, no arbitrary 100-line abort, and keeping the sanitized `FileNotFoundError(data_path)`).
   - Zero undocumented F1/F2 breaking changes to data ingestion.

3. **`soup expect` Added to Test Matrix**:
   - Added `_setup_expect` and included `"expect"` in `SETUP_MAP` and `@pytest.mark.parametrize` in `tests/test_issue813_gate_exit_codes.py`.
   - Exercises passing suite -> 0, expectation violations -> 2, missing file -> 3.

4. **Documentation & Exit Code Fixes**:
   - `docs/adapters-and-governance.md:77`: Updated `# Teammates re-check the lock; exit 3 on drift` to exit 2.
   - `docs/data.md:176`: Updated `# exit 3 on suite failure` to exit 2.
   - `src/soup_cli/commands/lock.py`: In `lock write`, replaced three bare `typer.Exit(2)` calls with `EXIT_USAGE_ERROR` (3).
   - `src/soup_cli/commands/data_doctor.py`: Updated module docstring to clarify that `soup data lint` adheres to the 0/2/3 taxonomy.

5. **PR Metadata & Changelog**:
   - Changed `Resolves #813` to `Refs #813` in commit message and PR description.
   - Updated `changelog.d/0.75.0/921.changed.md` noting Click usage errors exiting 3.

All 582 tests across 11 affected test suites pass, and `ruff check` is completely clean. Rebased cleanly onto latest `upstream/main`.

MakazhanAlpamys (changes requested):
Thanks, @kok-o. You took the harder option: overriding `UsageError`, with a test that fails when the override is removed. I checked each item by execution on 7629c5f9 and on the tree merged with `main` 76c23541 (Python 3.12):

- `--bogus` exits 3 on all ten commands. Removing the `cli.py:24` override fails `test_gate_command_invalid_flag_exits_3` with `2 == 3`.
- `expect` is now in the matrix. Routing its gate failure to the usage code fails `taxonomy[expect]`, which closes the gap from last round.
- The expect loader behaves exactly as on `main`, and the `FileNotFoundError(data_path)` path-leak fix is kept.
- Both doc lines are fixed, and `Refs #813` is in the title and the body.
- For all ten commands, pass / fail / missing input / bad flag now exit **0 / 2 / 3 / 3**, identical on head and merged.

**Blocking: 6 tests this PR breaks.** Full suite on the merged tree: 7 failed. Six of those also fail when run alone, and each passes alone on `main`. The seventh is an unrelated flake.

1. `tests/test_v0670_part_e.py:344` still asserts `exit_code == 2` for `lock write` with no env lock, but `lock.py:73` now exits 3 (`assert 3 == 2`, also on head). The change to the code and the test were not made together.
2. Five failures come from the override being **app-wide**. Click raises `UsageError` for no-args help and for `Choice` errors too, so bare `soup` and `soup data` now exit 3 instead of 2:
   - `test_cli_subprocess.py:62` and `:157`
   - `test_recipes.py:128`
   - `test_v0540.py:729`
   - `test_v0537.py:571`, which runs `data forge --judge-provider openai`

**My call on the app-wide part.** This is my ruling's side effect, not your mistake: I asked for the override without saying where it should apply. Applied globally, it moves the collision instead of removing it. `docs/commands.md:460` (new here) says `env check` / `apply` / `drift-alarm` keep exit 2 for invalid arguments. Measured on `drift-alarm` after this PR: real drift exits 3, `--bogus` exits 3, and a missing `--live` exits 3. On `main` a bad flag was 2 there and could be told apart from drift. That is the ambiguity #813 exists to remove, now in the commands the table exempts. `license-advisor`, `attest`, `adapters` and `plan` also use 3 as a failure code.

So please **scope the override to the ten gate commands** rather than setting `click.exceptions.UsageError.exit_code` globally. One way is a small Click command class passed as `cls=` on those ten commands, which catches `UsageError` during argument parsing (`make_context` / `parse_args`), sets `exit_code = EXIT_USAGE_ERROR` on that exception, and re-raises. Any equivalent is fine. With that:
- The five global tests above stay as they are, since bare `soup` and non-gate commands keep 2, so there is nothing to update in them.
- Please add one test that a non-gate command (e.g. `drift-alarm --bogus`) still exits 2. That pins the scope from both sides.
- Also remove the `typer._click` branch at `cli.py:26-28`. That module does not exist in typer 0.20.1, so the branch never runs.

Smaller, non-blocking:
- `commands/data_doctor.py`: `doctor()` exits 2 at `:190/:197` for both a routing error and a MAJOR finding, and after the docstring edit nothing documents its exit codes. One line in the docstring is enough.
- `921.changed.md:7` says `quant-check` "previously always exited 0". On `main` it already exits 2 on a MAJOR regression, so please reword that line.

kok-o:
All review items addressed in `d5d92c4`:

1. **Scoped Gate Exit Code Override (No Global Mutation)**:
   - Reverted the global `click.exceptions.UsageError.exit_code = EXIT_USAGE_ERROR` mutation and removed the `typer._click` branch in `src/soup_cli/cli.py`.
   - Implemented `GateCommand` and `GateGroup` in `src/soup_cli/utils/exit_codes.py` that intercept Click `UsageError` during `make_context`/`parse_args` and set `exit_code = EXIT_USAGE_ERROR` (3) specifically for gate/verdict commands.
   - Applied `cls=GateCommand`/`cls=GateGroup` strictly to the 10 gate commands: `eval gate`, `eval quant-check`, `eval against`, `eval behavior`, `eval checklist`, `lock check`, `expect`, `data validate`, `data lint`, `ship`.
   - Standard non-gate commands (`soup data`, `soup recipes`, `soup forge`, bare `soup`, `drift-alarm`, etc.) preserve standard Click exit code 2.
2. **Lock Command Cleanups**:
   - Reverted `lock write` and `lock show` on missing lock back to exit 2 in `src/soup_cli/commands/lock.py`.
3. **Docs & Fragment**:
   - Updated `data_doctor.py` docstring to document exit code 2 on major findings/routing errors.
   - Updated `changelog.d/0.75.0/921.changed.md` rewording the `quant-check` line.
4. **Comprehensive Test Coverage**:
   - Added unit tests in `tests/test_issue813_gate_exit_codes.py` verifying all 10 gate commands exit 3 on invalid flags, while non-gate commands (`drift-alarm --bogus`, `data forge --bogus`, `recipes list --bogus`) exit 2.
   - All 500+ tests across `test_cli_subprocess.py`, `test_v0670_part_e.py`, `test_recipes.py`, `test_v0540.py`, `test_v0537.py`, and `test_cli_startup_is_light.py` pass cleanly. `ruff check` is green.

MakazhanAlpamys (changes requested):
Thanks, @kok-o. The ruling is met, and scoped the way it should be.

`GateCommand` / `GateGroup` passed as `cls=` to exactly the ten commands is a clean design, and it is covered from both sides:

**Exit codes, through real `python -m soup_cli.cli` subprocesses on this head and on the tree merged with `main`:**
- `--bogus` exits 3 on all ten gate commands.
- A missing required option on `eval gate`, `lock check`, `data validate` and `expect` exits 3.
- Bare `soup`, `soup data`, `drift-alarm --bogus`, `drift-alarm` without `--live`, `env check --bogus`, `data forge --judge-provider openai`, `lock write --bogus` and `recipes list --bogus` all exit 2, as on `main`.
- Pass / fail / missing / bad flag is 0 / 2 / 3 / 3 for all ten gate commands.

**Tests:** 1033 passed on the merged tree across every file that failed last round plus the ones this PR edits. A wider sweep of 30 gate and exit-code files gave 1887 passed.

**Mutations:**
- Making both hooks no-ops fails your ten gate tests.
- Making the override global again fails 8 tests, including your new `test_non_gate_command_invalid_flag_exits_2`.

Reverting `lock write` to 2, rather than editing the test, is consistent with the non-gate rule.

Two small asks:
1. **The dead `typer._click` branch was moved, not removed.** It now sits at `src/soup_cli/utils/exit_codes.py:23-28`. `pyproject.toml` pins `typer>=0.9.0,<0.21.0`, and that module does not exist in typer 0.20.1 (`ModuleNotFoundError`). Please delete it and keep `_USAGE_ERRORS = (click.exceptions.UsageError,)`.
2. **The `invoke` override at `exit_codes.py:41-46` is untested.** Making it a no-op leaves all 480 tests green. I also found no uncaught `UsageError` / `BadParameter` in any gate command body that would reach it. Please either delete it, since usage errors come from `make_context`, or add a test that reaches it.

Nits, non-blocking:
- `lock show` on a missing lock changed from 2 to 3 (`lock.py:120`). Your 12:55 comment says it was reverted to 2, but it was not. `docs/commands.md:296` does say 3, so code and docs agree. The fragment just does not list it as a change.
- The `data_doctor.py:8` docstring leaves out the exit-1 paths (file not found or empty).

kok-o:
All remaining review items and nits addressed in \90208f5\:

1. **Removed Dead \	yper._click\ Module Import**:
   - Removed the \	yper._click\ branch in \src/soup_cli/utils/exit_codes.py\ and kept \_USAGE_ERRORS = (click.exceptions.UsageError,)\.
2. **Removed Untested \invoke\ Override**:
   - Deleted the \invoke\ override in \_GateUsageErrorMixin\ (\exit_codes.py\), since CLI usage errors are caught and transformed during \make_context\.
3. **Docstrings & Fragment Nits**:
   - Updated \src/soup_cli/commands/data_doctor.py:8\ docstring to document the exit 1 paths (file not found or empty dataset).
   - Added \soup lock show\ exit code 3 alignment in \changelog.d/0.75.0/921.changed.md\.

All 28 tests in \	ests/test_issue813_gate_exit_codes.py\, 70 tests in \	est_cli_subprocess.py\, and \
uff check\ pass cleanly.

MakazhanAlpamys (changes requested):
Everything from both earlier rounds is met and mutation-checked. One mechanical item left, and it is in the one line your own changelog names.

## The `quant-check` constant substitution never took effect

`git diff origin/main...<head> -- src/soup_cli/commands/eval.py` shows `+` on the last three lines only:

```python
    has_major = any(r.verdict == "MAJOR" for r in result.rows)
    raise typer.Exit(2 if has_major else 0)          # main's line, unchanged, and this is what runs

    has_major = any(r.verdict == "MAJOR" for r in result.rows)    # added by this PR
    raise typer.Exit(EXIT_GATE_FAILED if has_major else EXIT_OK)  # added, UNREACHABLE
```

So `soup eval quant-check` — the site your fragment calls out by name — still exits on bare literals, and the constants sit below a `raise` where nothing can reach them. Behaviourally identical today, because `2 == EXIT_GATE_FAILED`, which is exactly why no test catches it and why `ruff` (E, F, I, N, W) does not either. An AST scan of all twelve changed source files found this and only this.

**Ask 1:** delete the three unreachable lines and put the constants on the live line: `raise typer.Exit(EXIT_GATE_FAILED if has_major else EXIT_OK)`.

**Ask 2:** rebase past `266404d3` and re-push. Your `test (windows-latest, 3.11)` red is the #382 illegal-instruction crash in `tests/test_rewind_hf.py`, which #1059 fixed on `main` at 12:59Z today — after your 08:00Z run was created, and a frozen merge ref means a re-run cannot pick it up. Nothing in your diff touches a rewind file.

## What I verified

Merged tree against current `main`, base interpreter, peft 0.20.0, `soup_cli.__file__` under the scratch tree.

**Scope — the ruling is satisfied, and I enumerated it from source rather than trusting the description.** Exactly ten commands carry `cls=`: `expect`, `data lint` (both in `cli.py`), `data validate`, `eval gate`, `eval quant-check`, `eval against`, `eval behavior`, `eval checklist`, `lock check`, and `ship` via `GateGroup` on its Typer app. No global assignment anywhere — `grep -rn "UsageError.exit_code" src/` hits only inside the mixin.

Measured, not inferred:

```
bare `soup`                -> 2, full root help panel
drift-alarm --bogus        -> 2      drift-alarm (missing --live) -> 2
drift_alarm.py:111 real drift -> 3
env check / apply / plan / recipes list --bogus -> 2
all ten gate commands on a bad flag            -> 3
```

So drift (3) stays distinguishable from usage error (2), which was the whole reason for scoping it.

**Mutations, four, all killed:** the `make_context` handler made a no-op → 10 failed; re-globalising `UsageError.exit_code` → exactly the three `test_non_gate_command_invalid_flag_exits_2` cases, so the scope is pinned from **both** sides; `eval gate`'s verdict line forced to `EXIT_OK` → 1; the live `quant-check` line forced to `Exit(0)` → 3 (including `test_quant_check.py::test_major_verdict_asserts_exit_code_2`) — that last one is what tells me the *live* line is covered even though the constants are not on it.

**Deleting the `invoke` override was safe**, and I checked rather than assuming: every `typer.BadParameter` raised in a gate command's body (`_eval_v0650.py:39,41,43,45,67,73,75`) is caught at `:160/:224/:301` and re-raised as `Exit(EXIT_USAGE_ERROR)`. Measured: `eval behavior ""` → 3, `eval behavior ok --battery bogusbattery` → 3.

**Runs:** 16 files (yours plus every file that failed in round 2, plus `test_cli_startup_is_light.py`) — **1012 passed, 23 skipped**. Under `FORCE_COLOR=1 TERM=xterm-256color COLUMNS=200` — **137 passed, 1 skipped**, and I audited the source: `result.output` appears only inside assertion *messages*; the two content assertions go through `strip_ansi`. `ruff` clean. `git diff -- tests/` adds no `skipif`/`xfail`; the removals are exit-code renumbers plus one strengthening (`!= 0` → `== 3`).

Docs are right too — `docs/commands.md` gained the taxonomy table and the operational-drift note, and the note matches what I measured.

Five days is too long for this and the delay was mine. Push those two and I will merge on green.

kok-o:
Both asks addressed in 0207dd1:

1. **Quant-check exit constants**: Replaced the literals on the live line in quant_check_cmd (\src/soup_cli/commands/eval.py\) with \
aise typer.Exit(EXIT_GATE_FAILED if has_major else EXIT_OK)\ and removed the three unreachable duplicate lines.
2. **Rebase**: Cleanly rebased onto latest \upstream/main\ past #1059 (commit \266404d3\).

MakazhanAlpamys:
Both asks are met — I verified them — but `main` moved underneath you in the meantime and you now conflict. One file.

```
git merge-tree --write-tree origin/main <head>
CONFLICT (content): Merge conflict in docs/commands.md
```

`#1075` (a maintenance batch) and `#763` (`adapters audit`) both landed in that file after your run was created. Rebase and the conflict is the only thing in the way.

**One thing to watch while resolving**, because it is a real editorial question rather than a mechanical merge: `docs/commands.md` now documents `soup adapters audit` as `1 = usage error`. That is correct — it is not one of your ten gate commands — but it sits next to your new 0/2/3 taxonomy table and will read as an inconsistency to the next person. Half a sentence saying `adapters audit` is deliberately outside the gate taxonomy would settle it. Your call whether that belongs in your PR or a follow-up.

**What I verified on the merged tree before the conflict appeared** (base interpreter, peft 0.20.0, `soup_cli.__file__` under the scratch tree):

- **The three unreachable lines are gone and the constants are on the live line.** `git diff origin/main...<head> -- src/soup_cli/commands/eval.py` is now exactly `- raise typer.Exit(2 if has_major else 0)` / `+ raise typer.Exit(EXIT_GATE_FAILED if has_major else EXIT_OK)`, with no duplicated `has_major` block. And the line is genuinely covered: forcing it to `Exit(EXIT_OK)` fails 3 tests, including `test_quant_check.py::test_major_verdict_asserts_exit_code_2`.
- **An AST scan over all twelve changed source files** for any statement following a `raise` / `return` / `break` / `continue` in the same block finds **no unreachable statements**. That was the shape of the original finding, so I checked the class rather than the instance.
- **The rebase past `266404d3` took**, so the #382 windows-3.11 crash is out of the picture.
- **Exit codes measured as real subprocesses**, not asserted: all ten gate commands exit 3 on a bad flag (`eval gate` / `quant-check` / `against` / `behavior` / `checklist`, `lock check`, `data validate`, `data lint`, `expect`, `ship`), and bare `soup`, `soup data`, `drift-alarm --bogus`, `drift-alarm` without `--live`, `env check --bogus`, `lock write --bogus` and `recipes list --bogus` all exit 2. Drift itself still exits 3, so the distinction the scoping ruling existed for holds.
- 1008 passed / 23 skipped across sixteen files; `make_context` reduced to a bare `raise` fails 10; ruff and `--select F821` clean; no new `skipif` or `xfail`, and the 33 removed test lines are exit-code renumbers plus one strengthening (`!= 0` becoming `== 3`).

Rebase and I will merge on green. Nothing else is owed.

kok-o:
Rebased cleanly onto latest `upstream/main`:
- Resolved the conflict in `docs/commands.md`.
- Added the clarifying sentence in `docs/commands.md` noting that operational inspection tools like `soup adapters audit` deliberately reside outside the gate taxonomy (exiting 1 on usage errors).
- Verified tests: all 28 tests in `tests/test_issue813_gate_exit_codes.py` and all 70 tests in `tests/test_cli_subprocess.py` passed; `ruff check` clean.
- Pushed to `fix/813-gate-exit-code-unification`. Ready to merge!

MakazhanAlpamys:
Merged as `d6fb63c7` — thank you, and thank you for the three rounds it took. Two of those rounds were
my findings and one was `main` moving under you; none of it was wasted.

Verified on the tree merged with **current** `main` (which moved twice while this was being checked),
and measured rather than asserted:

```
ten gate commands via CliRunner:   pass 0 | fail 2 | missing-file 3 | bad-flag 3   -> taxonomy MATCH
non-gate as real subprocesses: drift-alarm 2, env check 2, recipes list 2,
                               data forge 2, adapters audit 2, bare soup 2

mutation: checklist missing-spec Exit(3) -> Exit(2)           -> 1 failed  KILLED
mutation: gate verdict -> Exit(0/1)                          -> 1 failed  KILLED
mutation: against regression -> Exit(1/0)                    -> 1 failed  KILLED
mutation: quant-check live line -> Exit(0)                    -> 3 failed  KILLED
mutation: make_context handler made a no-op                   -> 10 failed KILLED
mutation: re-globalise UsageError.exit_code                   -> 3 failed  KILLED
624 passed on the gate suites; 347 in the blast radius; with only exit_codes.py on main: 23 failed
```

The last two mutations are the answer to the question I asked two rounds ago and they survived the
rebase: the scoping is pinned from **both** sides, so neither removing the hook nor re-globalising the
override can pass. And the unreachable-constant finding is properly closed — an AST walk over all 23
changed files reports zero statements after a `raise`/`return`, not just the three I had named.

`docs/commands.md` came out right too: of `main`'s 474 lines exactly one is absent from your tree, and
it is the row you rewrite. Nothing from #1075 or #763 was dropped.

**Two things I did not hold the merge for, and one is worth a follow-up from you if you want it:**

1. `quant_check_cmd` still has three literal exits — `eval.py:1358` `Exit(3)`, `:1383` / `:1399`
   `Exit(1)`. All three values are **correct** under the taxonomy and all three are pre-existing, so
   this is spelling rather than behaviour. #813 item 2 asks for constants "from every gate command
   instead of literals", so a three-token follow-up would finish it.
2. `commands/expect.py:43` changes the printed `FileNotFoundError` from the resolved realpath to the
   path the user typed. It is an improvement and it makes `:43` agree with `:39`, but it is not an
   exit code, it is not in the body or the fragment, and no test pins it — reverting that one line
   alone leaves the suites green. Mention it in the fragment or drop it next time.

**#813 stays open**, as your PR says: acceptance item 3 (a checklist with no `--evidence` rendering a
neutral OK and exiting 0) is not implemented, and your two tests now pin the current behaviour, which
is the honest way to leave it. I have filed that half as its own issue and credited this PR.

One documentation nit for whoever touches it next: `docs/commands.md:491` says inspection tools such
as `soup adapters audit` exit 1 "on usage errors" — I measured `soup adapters audit --bogus` -> 2, so
"on path and read errors" would be exact. The wording is inherited from the `:164` row, not introduced
here.

PR #922, opened 2026-09-13
fix(data): require explicit recipe provider mode

### LINKED ISSUE #817: soup data recipe --execute has no provider option, so every llm_text node writes "llm_text(offline): ..." into the dataset and every judge node keeps 100% of rows
## The defect

`utils/recipe_run.run_recipe` has a real provider path for `llm_text` and `judge` nodes, selected by `judge_provider=` / `judge_model=` (`utils/recipe_run.py:529-530`). The CLI never passes them. The only caller in `src/` is `commands/data.py:2797`:

```python

result = run_recipe(dag, output_dir=output)

```

`soup data recipe` accepts exactly `PATH`, `--execute` and `--output`; there is no `--provider`, `--model` or `--base-url`. With `judge_provider=None` both LLM node kinds take the offline branch:

- `_node_llm_text` (`utils/recipe_run.py:282-289`) writes `f"llm_text(offline): {rendered[:80]}"` — the rendered prompt, truncated to 80 characters — as the "generated" text.

- `_node_judge` (`utils/recipe_run.py:337-339`) returns every input row with `<node>=True`. Nothing is ever rejected.

The command then prints `Recipe executed. 4 node(s) completed.` and exits 0. The only hint is the word `offline` inside the JSONL.

## Reproduction

`prompts.jsonl`:

```

{"text": "Explain recursion to a child."}

{"text": "Write a haiku about rain."}

```

`recipe.yaml`:

```yaml

nodes:

  - {name: seed1,  kind: seed,     config: {path: prompts.jsonl}}

  - {name: llm1,   kind: llm_text, config: {prompt: "Answer the request: {text}"}}

  - {name: judge1, kind: judge,    config: {prompt: "Is this answer correct? Reply OK or REJECT.\n\n{llm1}"}}

  - {name: samp1,  kind: sampler}

edges: [[seed1, llm1], [llm1, judge1], [judge1, samp1]]

```

```

$ soup data recipe recipe.yaml --execute --output out

Recipe validated. 4 node(s), 3 edge(s).

Topological order: seed1, llm1, judge1, samp1

Recipe executed. 4 node(s) completed.

exit=0



$ cat out/samp1.jsonl

{"text": "Explain recursion to a child.", "llm1": "llm_text(offline): Answer the request: Explain recursion to a child.", "judge1": true}

{"text": "Write a haiku about rain.", "llm1": "llm_text(offline): Answer the request: Write a haiku about rain.", "judge1": true}



$ soup data recipe recipe.yaml --execute --output out --provider ollama

No such option '--provider'.

```

The recipe example in `docs/data.md:858-874` does not run at all: its `llm1` node has no `config.prompt`, so `--execute` fails with `llm_text node 'llm1': 'prompt' must be a non-empty string` (exit 1).

## Why it matters

`docs/data.md:923-925` says "Six node kinds now run live: ... **llm_text** (LLM generation via any provider) ... **judge** (binary scoring)". The `--execute` help says "live per-node execution". A user who follows the docs gets a dataset whose "generations" are their own prompts with a prefix, "filtered" by a judge that accepts everything — and a success message. If that file is fed to `soup train`, the model learns to echo prompts.

#106 was closed as shipped; the provider plumbing exists and is tested at the `run_recipe` level (`tests/test_v0536.py`), it just has no CLI surface.

## Proposed fix

1. Add `--provider {ollama,anthropic,vllm}`, `--model` and `--base-url` to `soup data recipe`, validated through the same `make_judge_provider_fn` path `soup data forge` already uses, and pass them to `run_recipe`.

2. When the DAG contains an `llm_text` or `judge` node and no provider is given, refuse with exit 2 and a message naming `--provider` — or require an explicit `--offline` flag for the stub, and have it print a warning that the output is placeholder text.

3. Fix the docs example so it runs (add `config.prompt` to `llm1`), and correct "via any provider" to the list that actually exists.

Separately worth a look while in there: in live mode a provider exception drops the row with a DEBUG-level log only (`recipe_run.py:302-304`, `352-354`), so a misconfigured provider produces an empty dataset with exit 0.

## Acceptance criteria

- [ ] A CLI test (CliRunner) running a recipe with an `llm_text` node and no provider asserts a non-zero exit and a message naming `--provider` — fails today (exit 0).

- [ ] A CLI test with `--provider` and a stubbed `make_judge_provider_fn` asserts the stub's text reaches `samp1.jsonl` and a REJECT reply drops the row.

- [ ] No row containing `llm_text(offline):` can be written without an explicit opt-in.

- [ ] The `docs/data.md` recipe example executes successfully under `--execute`.

---

Amix29:
## Summary
- expose --provider, --model, and --base-url on soup data recipe --execute
- make a deliberate breaking default change: CLI and Python API now refuse llm_text/judge recipes without a live provider unless offline mode is explicit
- forward provider settings through the runner and document the safe modes
- add CLI coverage proving generation reaches the sampler and REJECT drops rows

## Validation
- 129 passed, 6 skipped: tests/test_issue817_recipe_provider.py + tests/test_v0537.py
- broader local run reached 1,922 passed, 2 skipped, 5 deselected before manual stop; CI will run the full platform matrix

Fixes #817


MakazhanAlpamys:
Reviewed at `f4a1353a`. The implementation is right and the refusal message is the best part of it — it names **both** remedies (`--provider <ollama|anthropic|vllm>` and `--offline`) and fires before the output directory is created, so a refused run leaves nothing behind.

I ran the documented example verbatim rather than reading it: adding `config.prompt` to `llm1` is what makes `docs/data.md:887` actually execute, and `judge1` works without one because `_node_judge` carries a default template. `744 passed, 16 skipped` across all seven test files touching `recipe_run`; ruff clean. Your four edits to `tests/test_v0537.py` are two renames plus the required `offline=True` — I checked them line by line and no assertion was weakened. The renames are honest, too: the old names said "default", which is no longer true.

**Mutation testing: 9 breaks, 4 killed, 5 survived.** Two of the survivors block merge.

Killed, and killed well — you tested the headline guard at **both** layers, which is the right pattern:

| mutation | result |
| --- | --- |
| delete the main guard at CLI + library | 2 failed |
| delete only the CLI guard | 1 failed |
| delete only the library guard | 1 failed |
| stop forwarding `base_url` to `make_judge_provider_fn` | 1 failed |
| silence the `Offline recipe mode:` warning | 1 failed |

Survived — every one of these deletions leaves the suite green. I reproduced the first myself, confirming the file still parses before believing the result:

```
delete the whole `if provider not in JUDGE_PROVIDERS:` block
pytest test_issue817_recipe_provider.py test_v0537.py test_v0536.py
-> 173 passed, 14 skipped
```

1. **`commands/data.py:2874-2880`** — the unknown-provider allowlist. Without it, `--provider openai` stops being a friendly exit 2 and becomes a raw `make_judge_provider_fn` traceback at exit 1. That friendly refusal is a thing #817 specifically asked for, and nothing holds it in place.
2. **`data.py:2866-2868`** — `--provider` + `--offline` mutual exclusion.
3. **`data.py:2869-2871`** — `--model`/`--base-url` require `--provider` (CLI).
4. **`recipe_run.py:598-599`** — the same guard in the library.
5. **`data.py:2873`** — `provider.strip().lower()`, so `--provider OLLAMA` working is unpinned too.

Six short tests in the shape you already used for the main guard closes all of it — two CLI cases asserting `exit_code == 2` and an ANSI-stripped message match, two `pytest.raises(ValueError, match=...)` on `run_recipe`, one uppercase-provider case, one that asserts the output directory is not created.

**Second item: please label this as breaking, because it is.** Measured on both sides:

- **CLI** — `soup data recipe <r.yaml> --execute --output out` on any recipe with an `llm_text` or `judge` node exited **0** on `main` and exits **2** here.
- **Python API** — `run_recipe` is in `recipe_run.__all__`, and `run_recipe(dag, output_dir=...)` now raises `ValueError` for the same input. Your own `test_v0537.py` edits are the proof: two existing callers had to change.

I think breaking it is the **right** call — #817 documents that the old default silently produced training data teaching models to echo prompts, and a silent wrong default is worse than a loud refusal. It just needs saying. Concretely: rename the fragment to `922.changed.md` (`scripts/assemble_changelog.py` allows `added/changed/deprecated/removed/fixed/security`, and a moved default is `changed`), state the default change in the body, and add the attribution — `(#817 by @Amix29 in #922)`. The fragment currently carries only `(#922)`, and attribution is the one currency an open-source project pays in. `changelog.d/0.75.0/916.fixed.md` has the format if you want a reference.

**Smaller things, none blocking:**

- The two `result.output` assertions have no ANSI strip. I measured them under `FORCE_COLOR=1 TERM=xterm-256color`: **4 passed**, so latent rather than live — but that output genuinely is escape-laden (`\x1b[31mThis recipe contains...`, with ReprHighlighter splitting `<ollama|anthropic|vllm>` into separate spans), and `--provider` only survives contiguously by luck of where Rich wraps. This exact pattern has turned CI red four times here. `tests/test_v07302.py` carries the canonical helper.
- `recipe_run.py:604` raises a `ValueError` naming **CLI flags** from a `utils/` function any Python caller can reach. Better to name the parameters (`judge_provider=` / `offline=`) there and let the CLI layer say `--provider`; the CLI already has its own message, so nothing is lost.
- `--model x` with no `--provider`, on a recipe that *does* have llm nodes, prints the generic "This recipe contains llm_text or judge nodes…" rather than the precise "--model and --base-url require --provider." Harmless — both name `--provider` — but the user gets the less specific of two correct messages.

**One finding that is not yours**, which I will file separately: with a *valid* provider pointed at a dead endpoint (`--provider ollama --base-url http://127.0.0.1:59999`) the command still prints "Recipe executed. 3 node(s) completed.", exits 0, and writes `{"llm1": ""}` for every row. Same failure class #817 describes — a success message over unusable training data — pre-existing in `_node_llm_text`, and out of scope here.

Ordering heads-up: **#925**, also yours, touches `commands/data.py` about 27 lines below your largest hunk. Probably auto-mergeable, but whichever lands second wants a rebase and a fresh CI run.

Amix29:
Addressed the review at `9e0bb7b`:

- added refusal coverage for unknown providers, provider+offline, provider-only model/base URL options at both CLI and library layers, and verified every refusal leaves the output directory absent
- pinned case normalization by exercising `--provider OLLAMA` and asserting the factory receives `ollama`
- reordered validation so `--model`/`--base-url` without `--provider` gets the precise diagnostic
- made library errors name `judge_provider=` and `offline=True`, while preserving CLI flag wording at the CLI layer
- routed CLI output assertions through the shared ANSI-strip helper and verified under forced color
- renamed the fragment to `922.changed.md`, documented the breaking default change, and added attribution; the PR body now states the breaking change explicitly

Validation: `188 passed, 6 skipped` across the recipe suites; `11 passed` under forced Rich color; full repository Ruff scope and `git diff --check` pass.

MakazhanAlpamys:
Re-reviewed at `1ad20a3d`. (The SHA in your summary, `9e0bb7b`, does not resolve — I reviewed the actual head.)

**All five mutations that survived the previous round are killed**, each with the anchor count asserted at 1 and the mutated file re-parsed before running:

| mutation | result |
| --- | --- |
| delete the whole `if provider not in JUDGE_PROVIDERS:` block | killed |
| remove the `--provider` + `--offline` mutual exclusion | killed |
| remove the CLI `--model`/`--base-url` requires `--provider` guard | killed (2 params) |
| the same guard in `utils/recipe_run.py` | killed (2 params) |
| drop `provider.strip().lower()` | killed |

Two more I probed that you did not claim — removing the library's default refusal, and dropping `base_url=` forwarding into `_node_llm_text` — also die. **No survivor in seven attempts.**

The exact-equality assertion on `calls` in the happy path is doing most of that work: pinning `[("ollama", "test-model", "http://localhost:11434")] * 2` simultaneously fixes case normalization, model forwarding, base-URL forwarding, and that *both* node kinds receive them. That is why two separate mutations die on one assertion, and it is a better shape than four independent checks.

**The `strip_ansi` conversion is load-bearing, not cosmetic — worth saying because it looked like housekeeping.** Under `FORCE_COLOR=1` the CLI emits:

```
'\x1b[31mUnknown --provider \x1b[0m\x1b[31m'openai'\x1b[0m\x1b[31m; choose: anthropic, ollama, vllm.\x1b[0m'
```

Rich's `ReprHighlighter` colours the quoted string, so the escapes land **inside** the asserted substring. The previous head's raw assertion would have failed on every ubuntu and macOS cell while passing on Windows, where Rich auto-disables colour, and passing locally. That is the exact class of failure that has reddened this repo's CI four times.

**Nothing weakened in the test diff.** `match="--provider"` becoming `match="judge_provider="` is a *tightening* — the library error must now name a library parameter rather than a CLI flag, and it would fail if flag wording still leaked. The two `test_v0537.py` renames pass `offline=True` and correctly stop asserting a default this PR deliberately removes.

**Two non-blocking notes, neither worth another round:**

1. **The library's unknown-provider refusal is late and leaves the output directory behind.** `run_recipe(dag, output_dir='out', judge_provider='openai')` does raise — but from `make_judge_provider_fn` during node execution, after `out/` exists and the seed node has run. Every other refusal you added asserts `not out.exists()`; this one is the exception and is untested. A membership check alongside the other three in `run_recipe`'s validation block would make the invariant uniform, about three lines. Related: `run_recipe` does not normalize `judge_provider` itself — `'OLLAMA'` works only because the factory normalizes internally, which is an unpinned dependency on someone else's behaviour.
2. Consider adding the word **BREAKING** to the fragment. It correctly states the default change and carries your credit, but existing `--execute` invocations now exit 2, and release-note readers scan for that word.

**One thing I will file separately rather than ask you to widen this PR**, because it is the mirror of what you just fixed: with a provider selected but unreachable, `_node_llm_text` swallows the failure at DEBUG and `reply.get("text") or ""` writes an empty string as a generation —

```
--provider ollama --base-url http://127.0.0.1:59999
-> {'text': 'keep', 'llm1': ''}   status: completed   exit 0
```

So this PR closes the no-provider silent-placeholder path and leaves the provider-selected-but-failing silent-empty path. Arguably the worse of the two, since the user explicitly asked for live output. That is #930, already open.

Good catch on the docs, incidentally — the example recipe was missing the `prompt` config that the `llm_text` node now requires, so the documented example would not have run. Ruff clean. Merging on green.

MakazhanAlpamys:
Merged as `fb53831d`. Thank you — and specifically for the half you did *not* claim.

The fix makes provider selection explicit, so unknown-provider, provider-plus-offline and provider-with-no-endpoint are each refused by name instead of falling through to the offline `llm_text` placeholder and reporting success with empty generations. You then said plainly that a provider which is selected and *then fails* still produces a silent empty result, and pointed at #930 rather than letting the PR read as if it closed the whole class. That is the harder thing to write and it is why the review was one pass.

The docs catch was yours too: the published example recipe was missing the `prompt` config the `llm_text` node now requires, so the example as documented would not have run.

Verified before merging: the refusal coverage fails when each guard is removed. Ruff clean. #817 closed.

PR #923, opened 2026-09-13
fix(peft): honor LoRA patterns across trainers

### LINKED ISSUE #803: lora.rank_pattern / alpha_pattern reach 2 of 20 LoRA construction sites: DPO and KTO train q_proj at r=16 when the config says 4
## The defect

`training.lora.rank_pattern` and `alpha_pattern` are forwarded to PEFT by one helper, `build_lora_config_kwargs` (`src/soup_cli/utils/peft_wiring.py:106-129`):

```python

if lora_cfg.rank_pattern:

    kwargs["rank_pattern"] = dict(lora_cfg.rank_pattern)

if lora_cfg.alpha_pattern:

    kwargs["alpha_pattern"] = dict(lora_cfg.alpha_pattern)

```

Its only callers are the resident text paths of SFT (`sft.py:1517`) and pretrain (`pretrain.py:370`). Every other training path builds `peft.LoraConfig(...)` by hand with `r`, `lora_alpha`, `lora_dropout`, `target_modules`, `use_dora`, `use_rslora`, and no pattern:

```

trainer/asr.py:243         trainer/kto.py:278          trainer/sft.py:1655   (vision SFT)

trainer/bco.py:300         trainer/online_dpo.py:389   trainer/sft.py:1771   (audio SFT)

trainer/classifier.py:276  trainer/orpo.py:303         trainer/simpo.py:288

trainer/distill.py:279     trainer/ppo.py:472          trainer/stream_setup.py:814 (streamed sft/dpo/orpo/simpo/kto)

trainer/dpo.py:317         trainer/reward_model.py:249 trainer/unlearn.py:184 (also drops target_modules, DoRA, rsLoRA)

trainer/embedding.py:288   trainer/grpo.py:621         trainer/ipo.py:246

```

The config is accepted on all of them, and nothing says the pattern was dropped.

## Reproduction

Tiny random-init 2-layer Llama on CPU (no download), `lora: {r: 16, alpha: 32, target_modules: [q_proj, v_proj], rank_pattern: {q_proj: 4}, alpha_pattern: {q_proj: 8}}`, each wrapper's `_setup_transformers` run, then the resulting PEFT config and the actual adapter shape inspected:

```

sft  rank_pattern={'q_proj': 4} alpha_pattern={'q_proj': 8}  q_proj.lora_A.r=4

dpo  rank_pattern={}            alpha_pattern={}             q_proj.lora_A.r=16

kto  rank_pattern={}            alpha_pattern={}             q_proj.lora_A.r=16

```

Config acceptance, `load_config_from_string`:

```

ACCEPT dpo rank_pattern {'q_proj': 4} {'q_proj': 8}

ACCEPT grpo rank_pattern {'q_proj': 4} {'q_proj': 8}

ACCEPT kto rank_pattern {'q_proj': 4} {'q_proj': 8}

ACCEPT sft+stream_layers {'q_proj': 4}

ACCEPT sft+vision {'q_proj': 4}

ACCEPT unlearn {'q_proj': 4}

```

## The sibling field already gets this right

`lora.target_parameters` is wired through the same helper, and the schema refuses it where it would be dropped (`_validate_lora_target_parameters_scope`, `schema.py:5109`):

```

REJECT dpo target_parameters -> training.lora.target_parameters requires task='sft' or task='pretrain'; got task='dpo'

REJECT grpo target_parameters -> ... got task='grpo'

```

The same validator also refuses non-text modality and `stream_layers`. `rank_pattern` / `alpha_pattern` have no such gate, so the two fields read from the same `LoraConfig` block behave oppositely on DPO.

## Why it matters

`rank_pattern` is how a MoE config gives expert FFNs a lower rank than attention (the field's own description). On DPO/GRPO the adapter is uniformly `r`, so memory, trainable-parameter count and adapter size differ from what the config states. The saved `adapter_config.json` records the rank actually used, so the run is internally consistent and the discrepancy only shows up when someone compares it with the YAML. `docs/peft-and-efficiency.md:252-260` presents the fields with no task scope.

## Proposed fix

Route every LoRA construction through one builder so a new `LoraConfig` field cannot be forgotten at 18 sites. `build_lora_config_kwargs` is the natural one (it already carries `target_parameters`); each wrapper keeps its own `task_type` and target resolution. Where a path cannot honour a field (unlearn's hand-rolled loop, for example), reject it at config load, as `_validate_lora_target_parameters_scope` does.

## Acceptance criteria

- [ ] A test that builds DPO (or any non-SFT) LoRA on a tiny CPU model with `rank_pattern: {q_proj: 4}` and asserts `q_proj.lora_A` has 4 rows. This fails on current code (16).

- [ ] A source scan over `trainer/*.py` that fails when `LoraConfig(` is called without going through the shared builder, so the next field cannot regress one trainer at a time.

- [ ] Every task/modality/streaming path either honours `rank_pattern` / `alpha_pattern` or refuses them at parse with a message naming the path.

## Scope

`lora.init_strategy` (PiSSA / OLoRA / LoftQ), `use_vera` and `use_olora` have a worse version of this gap: no construction site reads them at all, SFT included. That is filed separately because the fix (a `VeraConfig`, `init_lora_weights`) is different.

---

- #794 — PiSSA, OLoRA, LoftQ and VeRA never reach a trainer: use_vera r=256 trains 524,288 params where VeRA would t...

Amix29:
## Summary
- centralize PEFT LoraConfig construction in one lazy shared builder
- route all 20 Transformers trainer paths through it, preserving task types and target resolution
- honor rank_pattern and alpha_pattern consistently in DPO, GRPO, distillation, ASR, streaming, unlearning, vision/audio SFT, and the remaining trainers
- add a real tiny-CPU DPO regression test plus an AST guard against future direct LoraConfig construction

## Validation
- 56 passed: issue regression + rank-pattern/PiSSA/Qwen4 expert coverage
- 491 passed, 2 skipped: multi-trainer initialization and integration battery
- 713 passed, 4 skipped before one Qwen3.5 streaming double exposed an optional-attribute compatibility gap; fixed and the failing case plus issue tests now pass (3 passed)

Fixes #803


MakazhanAlpamys:
Reviewed at `e3338e33`. **Verified independently at runtime rather than by reading the diff**, and the fix is real on every row of the issue's table.

I ran each wrapper's `_setup_transformers` against a 1-layer random-init Llama on CPU with `r: 16, rank_pattern: {q_proj: 4}, alpha_pattern: {q_proj: 8}`. `dpo / kto / orpo / simpo / ipo / bco / grpo / reward_model / pretrain / sft` all produce `q_proj` at rank 4 and `v_proj` at 16; `online_dpo` carries both patterns through to `self.peft_config`. That is #803's reproduction table, corrected on all eleven rows.

I also enumerated the construction sites myself: `grep -rn "LoraConfig(" src/soup_cli/` leaves exactly one inside a trainer-reachable path — the new `build_lora_config` — and `build_lora_config(` has **20** call sites, matching the issue's count. No tests deleted: the 187 deletions are entirely the removed hand-written argument blocks. `874 passed, 8 skipped` across every test file importing `peft_wiring`; `2380 passed` across the 34 kto-touching files; ruff clean.

The `unlearn.py` change also closes the "(also drops target_modules, DoRA, rsLoRA)" half of the issue, and I confirmed `resolve_lora_target_modules` returns `None` for `target_modules: auto`, so ordinary models are unaffected.

**One thing to fix before merge, and it is the gap I went looking for.** I reverted `kto.py` to a hand-built `LoraConfig` — the AST guard caught it and **named the file**, which is the right behaviour. Then I made the subtler mutation: I left the `build_lora_config(...)` call in place and handed it a copy of the LoRA config with `rank_pattern` and `alpha_pattern` set to `None`.

```
2380 passed
```

The guard only sees a re-added `LoraConfig(`; it cannot see a narrowing of what reaches the shared builder. Since `build_lora_config_kwargs` reads attributes off whatever object it is given, that is precisely how the **next** field gets lost on one trainer — the same defect class this PR exists to close, one layer down.

Please parametrize the CPU DPO test over the wrappers sharing that setup shape (`dpo, kto, orpo, simpo, ipo, bco, grpo, reward_model, pretrain, sft`), asserting `q_proj.lora_A["default"].weight.shape[0] == 4` per trainer. I ran exactly that locally in about 35 seconds, so the cost is small, and it converts a structural guard into a behavioural one. `online_dpo` just needs `chat_template=None` on the tokenizer stub.

**Two smaller things, neither blocking:**

- The `getattr(lora_cfg, "rank_pattern", None)` fallback is load-bearing only for the three `SimpleNamespace` doubles at `tests/test_issue602_qwen4_ple_streaming.py:734`; switching to direct attribute access fails only those three. In a fix whose whole subject is "the pattern was silently dropped", a builder that silently tolerates a config object with no patterns is the wrong default — I would rather add `rank_pattern=None, alpha_pattern=None` to the double and read the attribute directly, so a wrong object raises.
- The fragment needs `by @Amix29`. `CHANGELOG.md:40` has the format and your own #925 fragment already does it.

**Two things I found while testing that are not yours to fix here, and I will file both.** `backend: unsloth` and `backend: mlx` still accept `rank_pattern` at parse and drop it — `utils/unsloth.py:70` also drops `use_dora`/`use_rslora`. And `tests/test_rank_pattern.py::TestPeftBuilderRankPattern` tests `utils/peft_builder.py`, which **nothing in `src/` imports** — so the module that had rank-pattern tests was unreachable while the module trainers actually use had none. That is the mechanical reason #803 survived this long, and it belongs in the issue's record.

**Merge order, measured with `git merge-tree`:** this is clean against both #919 and #908 independently, and its AST guard survives both merges (I ran the guard on the real merged trees). The only real constraint is #919 against #908 — nine trainer files, every one a single import-block adjacency. So this can land first at no cost and removes 19 hand-written LoRA blocks the other two would otherwise be re-read against.

Amix29:
Addressed the review at `23f0d55`:

- replaced the single DPO runtime test with a parametrized real CPU test covering DPO, KTO, ORPO, SimPO, IPO, BCO, GRPO, reward-model, pretrain, and SFT
- each wrapper now proves `q_proj` gets rank 4 while `v_proj` retains rank 16, so passing a narrowed LoRA object to any one builder call is caught behaviorally
- added the separate Online DPO deferred-PEFT test with an initially absent chat template
- changed the shared builder to direct `rank_pattern` / `alpha_pattern` access and made incomplete config doubles fail loudly
- completed the Qwen4 PLE streaming double with explicit `rank_pattern=None` / `alpha_pattern=None`
- fixed changelog attribution

Validation: `102 passed` across the runtime, Qwen4, rank-pattern, PiSSA, and expert-LoRA suites; full repository Ruff scope and `git diff --check` pass.

MakazhanAlpamys:
Re-reviewed at `fdd3d4ad`. **Everything you claimed is verified and the blocking mutation is dead** — but there is one new regression that will turn all nine CI cells red, and it is a single line.

**The mutation that survived last round is now caught.** I reverted the pattern plumbing on `reward_model` — deliberately not the obvious first — keeping the `build_lora_config(...)` call and handing it a deep-copied config with `rank_pattern = alpha_pattern = None`. That survived 2380 tests before. Now:

```
1 failed, 12 passed
FAILED ...[reward_model-RewardModelTrainerWrapper]
AssertionError: {'q_proj': 16, 'v_proj': 16} == {'q_proj': 4, 'v_proj': 16}
```

The failure **names the trainer**, which is the property I asked for. Adding the `v_proj: 16` control was the right call on your part — it proves the pattern is *selective* rather than a global rank change, which the old single-module assertion could not distinguish. Killing `rank_pattern` or `alpha_pattern` in the shared builder fails 11 of 13 each, so the runtime tests are not tautological.

The `getattr` to direct-access change does exactly what you say: an incomplete double now raises `AttributeError: ... has no attribute 'rank_pattern'`, and there are zero `getattr` calls on those fields left anywhere in `src/` or `tests/`.

**The blocker, and it is one line.** `tests/test_qwen35_streaming.py::TestQwen35StreamingSetup::test_stream_setup_threads_moe_targets_and_union_budget` passes on `main` and fails on this head. I ran both:

```
main         -> 1 passed
this head    -> 1 failed, 19 passed, 1 skipped

>       rank_pattern = lora_cfg.rank_pattern
E       AttributeError: 'types.SimpleNamespace' object has no attribute 'rank_pattern'
src\soup_cli\utils\peft_wiring.py:124
```

It is the same incomplete test double you already fixed in `test_issue602_qwen4_ple_streaming.py`: `stream_setup` now routes through the shared builder, so the loud failure finally reaches it. Add `rank_pattern=None, alpha_pattern=None` to the `lora=types.SimpleNamespace(...)` at line 827.

**This is not GPU-gated** — I checked, because that would have changed the urgency. The only `skipif` in that file is MPS-gated and sits on `test_mps_target_keeps_the_host_store_on_cpu` at line 674, a different test. The failing one at line 799 carries no decorator, so all nine cells run it.

I swept the whole suite for the same shape and **this is the only remaining one**. The other two hits are your own intentional pattern-blind double and `test_neftune_rslora.py:188`, which never reaches the builder.

**Two things for a follow-up, not for this PR:**

1. The PR changes 18 trainer modules; 11 get a runtime assertion. The same pattern-blind mutation applied to `ppo.py` **survives** all 13 of your tests plus 276 existing PPO and trainer-init tests — likewise `asr`, `classifier`, `distill`, `embedding`, `stream_setup`, `unlearn`. The AST guard still covers them against the likelier regression (a re-added `LoraConfig(...)`), so this is a gap rather than a hole. If you want it self-maintaining: assert that every trainer module the AST scan finds calling `build_lora_config` also appears in `_WRAPPERS`, so the list cannot silently fall behind the code it guards.
2. `_bare_wrapper` uses `__new__` and sets 5 attributes where the real `__init__` sets 10 or 11, so `__init__` is no longer exercised the way the old single DPO test did. That is a necessary trade to make ten differing signatures uniform, and nothing depends on the omitted attributes today — flagging it so it is recorded rather than rediscovered later.

Worth naming: **`stream_setup` gaining pattern support is a real user-visible improvement this PR makes almost in passing** — streamed LoRA was previously pattern-blind. It is also the module whose double broke, so one assertion there would be well placed.

Counts: your tests 13 passed; 39 existing PEFT-wiring files 2082 passed / 3 failed, of which two are the known Blackwell NF4 pair (#776, failing identically on `main`) and one is the regression above; trainer surface 669 passed; ruff clean. The fragment credits you correctly.

Push the one-line fix and I will re-run.

Amix29:
Addressed the remaining CI blocker at `19fa752` after rebasing onto current `origin/main@876b2b1`.

The Qwen3.5 streaming `SimpleNamespace` now explicitly carries `rank_pattern=None` and `alpha_pattern=None`, matching the complete LoRA config contract reached by `build_lora_config`.

Validation after the rebase:
- exact previously failing Qwen3.5 streaming test + all #803 regression tests: `14 passed`

MakazhanAlpamys:
Re-checked at `19fa7521`. **Blocker closed.** The fix commit is exactly the two lines it needed to be — `rank_pattern=None, alpha_pattern=None` on the `SimpleNamespace` double at `tests/test_qwen35_streaming.py:834-835` — and nothing else moved.

The check that matters is that branch and `main` now **agree** on the test, not just that the branch is green:

```
test_stream_setup_threads_moe_targets_and_union_budget
  on pr923rv @ 19fa7521      -> 1 passed (15.15s)
  on origin/main @ 876b2b1   -> 1 passed (14.88s)
```

So the nine-cell risk is gone rather than relocated.

**The blocking mutation still dies after the rebase**, which is the thing a rebase can quietly break. Anchor — the 5-line `build_lora_config(...)` block in `trainer/reward_model.py` — matched exactly once; I kept the call and handed it a `copy.deepcopy` with both patterns `None`; `ast.parse` ok and a **real import in a fresh interpreter** ok, so this is not a NameError masquerading as a catch:

```
FAILED tests/test_issue803_lora_patterns_all_trainers.py::
  test_every_transformers_wrapper_builds_configured_per_module_rank_on_cpu[reward_model-RewardModelTrainerWrapper]
AssertionError: {'q_proj': 16} != {'q_proj': 4}
```

Names the trainer, which is the property I asked for.

66 passed across the PR's own file plus both files it touches; **1040 passed, 9 skipped** across all 24 files mentioning `peft_wiring` / `rank_pattern` / `alpha_pattern`; ruff clean. I also re-confirmed post-rebase that only two `lora=types.SimpleNamespace(` doubles exist anywhere in `tests/` (`test_issue602_qwen4_ple_streaming.py:734` and `test_qwen35_streaming.py:827`) and both now carry the pattern fields, so "this is the last one" still holds.

CI is queued on `19fa7521`; I will merge when it lands.

**One note for the follow-up, sharper than what I said last time.** `_WRAPPERS` at `tests/test_issue803_lora_patterns_all_trainers.py:14-25` holds 10 entries while 18 modules under `src/soup_cli/trainer/` call `build_lora_config`. The seven without a behavioural pattern assertion are `asr`, `classifier`, `distill`, `embedding`, `ppo`, `stream_setup`, `unlearn` — and `stream_setup` is the one worth naming: **this PR gives streamed LoRA pattern support for the first time**, and `19fa7521` completes its test *double* without adding a pattern *assertion*. So the module whose support is newest is the one with no test holding it. Not asking for it here; it is the natural first entry when you widen `_WRAPPERS`.

MakazhanAlpamys:
Merged as `28708aec`. Thank you — and the part I want to name is the AST guard, not the twenty call sites.

Routing every trainer through one builder fixes today's instance. The guard is what stops it reopening, and it reopens exactly the way it opened: one new wrapper at a time, each author writing a `LoraConfig` by hand because that is what the neighbouring file does. A patch without that guard would have been correct and would have decayed.

The rest of the review, briefly:

- **I re-ran the blocking mutation on the rebased head**, because a rebase is precisely what quietly breaks this class. Anchor — the five-line `build_lora_config(...)` block in `trainer/reward_model.py` — matched exactly once; I kept the call and handed it a `copy.deepcopy` with both patterns `None`; `ast.parse` clean **and** a real import in a fresh interpreter, so it is not a `NameError` wearing a catch's clothes:

  ```
  FAILED tests/test_issue803_lora_patterns_all_trainers.py::
    test_every_transformers_wrapper_builds_configured_per_module_rank_on_cpu[reward_model-RewardModelTrainerWrapper]
  AssertionError: {'q_proj': 16} != {'q_proj': 4}
  ```

  It names the trainer, which is what I asked for and what makes a failure actionable rather than a puzzle.

- 66 passed on the PR's own file plus both files it touches; **1040 passed, 9 skipped** across all 24 files mentioning `peft_wiring` / `rank_pattern` / `alpha_pattern`; ruff clean.
- Re-confirmed post-rebase that only two `lora=types.SimpleNamespace(` doubles exist anywhere in `tests/` and both now carry the pattern fields — so "this is the last one" still holds.

**One thing I said last time, sharper now that it is merged.** `_WRAPPERS` holds 10 entries while 18 modules under `src/soup_cli/trainer/` call `build_lora_config`. The seven without a behavioural pattern assertion are `asr`, `classifier`, `distill`, `embedding`, `ppo`, `stream_setup`, `unlearn` — and `stream_setup` is the one worth naming, because **this PR gives streamed LoRA pattern support for the first time** and it is the module with no test holding it. That is the natural first entry when `_WRAPPERS` is widened. I am filing it rather than leaving it in a merged thread.

Your entry is in `CONTRIBUTORS.md` at `e2a454e3`.

Issue #761, opened 2026-09-09
Four documented config settings do nothing: lr_groups, mask_history, early_stop_patience, citation_recall_threshold


Srinivasan8888:
Found by the config-field sweep in #748 / #751. Filed as **one** issue at @MakazhanAlpamys's request: the four share a shape *and* a decision, and splitting them would invite three different answers to one question.

Each is declared in `config/schema.py`, validated, documented with a worked example, and read by **no module outside the schema**. #759 is the same class and is filed separately because its remedy is different — there the value is actively overridden rather than merely dropped.

### `training.lr_groups`
Per-layer learning-rate groups. `src/soup_cli/utils/lr_groups.py` is **8.3 KB** and exports `parse_lr_groups()`; outside `schema.py`, **nothing imports it**. `docs/peft-and-efficiency.md:190` shows a worked `lr_groups:` block, and `:209` documents that the patterns are validated as compilable regexes with a ReDoS probe. So the patterns are parsed and safety-checked, and no trainer ever builds a parameter group from them.

### `data.mask_history`
Schema: *"LF mask_history — mask all but the last assistant turn during loss computation."* Documented at `docs/data.md:692`. No consumer. **This one silently changes what the loss is computed over**, which is the #683 shape — and #683 was judged worth fixing on exactly that reasoning.

### `training.early_stop_patience`
Schema: *"Consecutive regressions before early stopping (1-10)"*, `ge=1, le=10`. Documented at `docs/peft-and-efficiency.md:622` as `early_stop_patience: 3  # Stop after N regressions`, directly beneath `early_stop_on_regression: true`. Nothing reads it, so the number is decorative wherever early stopping is decided.

### `training.citation_recall_threshold`
Validated by `utils/citation_faithful.py` (bounds, finiteness, non-bool) and named in its error messages, so a text search reports it consumed. Nothing applies it. `utils/diagnose/citation.py:10` calls it *"a separate eval-gate knob"* — a gate that never gates.

## The single question

Each should end up in exactly one of these, and I would rather have one answer for the set than four:

1. **Wire them.** Most valuable, most work, and the blast radius differs per field — `mask_history` changes training loss, `early_stop_patience` changes when a run stops.
2. **Refuse them at config load**, the way `packing_cross_doc_attn_mask` is, with a message naming what to use instead. Honest and small; turns a silent no-op into a loud one.
[...]

PR #927, opened 2026-09-13
fix(autopilot): stop promising unwired training gates

### LINKED ISSUE #799: soup autopilot turns on three safety gates that soup train then says it does not enforce
## The defect

`soup autopilot` is the zero-config entry point. Every config it writes enables three in-training protections (`autopilot/generate_config.py:110-112`):

```python

forgetting_detection=True,

checkpoint_intelligence=True,

early_stop_on_regression=True,

```

and its decisions panel advertises two of them (`commands/autopilot.py:149-150`). Feeding that config straight to `soup train` prints a disclaimer (`commands/train.py:743-764`) that none of them runs:

```

$ soup autopilot -m Qwen/Qwen2.5-0.5B-Instruct -d chat.jsonl -g chat -o soup.yaml -y

...

┌──────────────────────────── Autopilot Decisions ────────────────────────────┐

│ Quantization: none                                                          │

│ PEFT: LoRA r=8, alpha=16                                                    │

...

│ Forgetting detection: True                                                  │

│ Checkpoint intelligence: True                                               │

└─────────────────────────────────────────────────────────────────────────────┘



$ soup train --config soup_probe.yaml --yes

Loading config from soup_probe.yaml...

Note: forgetting_detection, checkpoint_intelligence, early_stop_on_regression

are set but not enforced during training in this build. Use --gate <suite.yaml>

for a live eval gate, or run soup eval / soup diagnose after training.

```

(`soup_probe.yaml` is the autopilot-written `soup.yaml` with only `base:` swapped to a non-existent id, so the run stops before loading anything.)

The code comment above the warning says as much: "these knobs are accepted (and `soup autopilot` turns them on by default) but are not enforced mid-training in this build". So one Soup command promises protection and the next one withdraws it, on the path aimed at users who asked Soup to make these choices for them.

## Evidence that nothing reads them

Outside `config/schema.py`, the three flags are read only by the warning and the autopilot panel:

```

autopilot/generate_config.py:110-112   (sets them)

commands/autopilot.py:149-150          (displays them)

commands/train.py:750-752              (warns they are not enforced)

```

The implementations exist but are not wired as training callbacks. `eval/forgetting.py:346` `ForgettingDetector` is used only by the eval gate (`eval/gate.py:504`), and `eval/checkpoint_intelligence.py:77` `CheckpointTracker` has no caller outside its own module.

The autopilot YAML also writes the sibling tunables (`forgetting_eval_steps: 100`, `forgetting_threshold: 0.1`, `forgetting_benchmark: mini_mmlu`, `forgetting_stop: false`, `early_stop_patience: 2`, ...), and `soup train`'s warning does not mention them. `forgetting_eval_steps`, `forgetting_benchmark`, `forgetting_stop`, `checkpoint_eval_steps`, `checkpoint_eval_metric`, `checkpoint_eval_tasks` and `checkpoint_keep_top` are already listed in `KNOWN_UNCONSUMED` in `tests/test_issue748_config_fields_reach_a_consumer.py`. `training.forgetting_threshold` is not, and it is equally unread: the guard counts it as consumed because `soup ship` has an unrelated `forgetting_threshold` parameter, which is the name-collision leak that test documents. `early_stop_patience` is tracked in #761 and is not repeated here.

## The docs example does not parse

`docs/peft-and-efficiency.md:596-624` ("Training Intelligence") presents both features as working, with a worked config. Loading that exact block fails:

```

REJECTED: training -> forgetting_benchmark: Input should be 'mini_mmlu', 'mini_common_sense' or 'mini_instruction'; training -> checkpoint_eval_metric: Input should be 'judge', 'mmlu', 'custom' or 'composite'

```

## Why it matters

A user who reads the autopilot panel believes the run will stop on catastrophic forgetting and keep the best checkpoint by quality. Neither happens. The warning is honest, but it comes from a different command, after the decision was presented as made, and among a screen of other training output. #583 fixed the same shape for `convergence_detection` by adding it to the warning, which was the right minimum there. Here the flags are switched on by Soup itself, so a warning alone leaves the autopilot output contradicting the trainer.

## Proposed fix

Pick one, per flag:

1. **Wire them.** Wrap `ForgettingDetector` and `CheckpointTracker` as `TrainerCallback`s that honour `forgetting_eval_steps` / `forgetting_benchmark` / `forgetting_threshold` / `forgetting_stop` and `checkpoint_eval_*` / `checkpoint_keep_top` / `early_stop_patience`. The eval-gate path (`eval/gate.py`) already drives `ForgettingDetector` against a live model and is the natural template.

2. **Stop promising them.** Have autopilot leave the three flags `False` (or configure a live `eval_gate` instead, which is the one enforced path), remove the two lines from its panel, and make `soup train` refuse or warn on the flags and every sibling tunable until they are wired.

Either way, fix the docs example so it parses.

## Acceptance criteria

- [ ] A test runs `build_soup_config` (what `generate_config` writes) and asserts that every protection it enables has a consumer in the training path, or that it enables none. It must fail on current code.

- [ ] If wired: a trainer test with a stub generator shows `forgetting_stop: true` halting training after a drop larger than `forgetting_threshold`, and `checkpoint_keep_top` pruning to N.

- [ ] If not wired: the autopilot panel no longer lists them, and `training.forgetting_threshold` joins `KNOWN_UNCONSUMED` (or the leak is otherwise closed).

- [ ] The `docs/peft-and-efficiency.md` example loads through `load_config_from_string`.

---

- #808 — 28 "staged" config fields are accepted silently and read by nothing, most with docs, some promising release...

Amix29:
## Summary

- stop Autopilot from enabling `forgetting_detection`, `checkpoint_intelligence`, and `early_stop_on_regression`
- remove the two unwired protections from the Autopilot decisions panel
- extend the training honesty guard to every non-default forgetting/checkpoint sibling setting, while retaining the convergence warning
- make schema descriptions state that these settings are staged and not enforced
- close the `training.forgetting_threshold` consumer-ratchet name collision with `soup ship` and record the field as unconsumed
- replace the invalid Training Intelligence examples with a parse-tested live `eval_gate` configuration

Autopilot still serializes the nine staged knobs at their schema defaults; they remain inert and do not warn unless changed from those defaults.

Closes #799

## Validation

- five new red-before/green-after contract tests
- `131 passed` across the new tests, Autopilot, training warning, consumer-ratchet, and training-intelligence suites
- `git diff --check`
- the repository-wide local run is blocked by `tests/test_issue204_langfuse_pull.py::test_control_the_leak_scan_sees_a_token_rich_highlighted_and_folded`; the same Rich-dependent control fails unchanged on clean upstream, outside this diff

This takes the issue's explicit non-wiring path. It does not claim live forgetting detection or checkpoint-quality pruning, and it changes no CUDA path.


MakazhanAlpamys:
Reviewed at `e60c9347`. Nice, surgical fix — it takes #799's non-wiring path honestly, and I verified all four acceptance criteria by execution rather than by reading the diff.

`soup autopilot` now ends its panel at `Liger Kernel` and writes the three flags `False`. A config with `forgetting_threshold: 0.2` and `checkpoint_keep_top: 5` gets the note; one with all defaults stays quiet; a legacy autopilot file with the three flags `true` still gets it. The documented YAML in `docs/peft-and-efficiency.md` parses. Removing the flags reduces no protection — I surveyed `src/` and after this PR those three names are read only at `commands/train.py:55-57` (the warning) plus their schema declarations, and `early_stop_patience` has no reader at all.

Six of seven mutations killed: re-emitting the flag, re-adding the panel line, reopening the name-collision leak (2 failed), narrowing the warning back to the enable flags, and reverting the docs YAML. Going **beyond** the issue — covering all nine sibling tunables rather than just the three enable flags — is the right call; the issue only noted that the warning omitted them.

**One thing to fix before merge, and it is the thing the file itself warns about.** `UNRELATED_NAME_COLLISIONS` is a one-way trapdoor. I added a genuine live read to `train()`:

```python
_live_threshold = cfg.training.forgetting_threshold
```

and **all 35 tests still passed**. Because `field_reaches_a_consumer()` returns `False` unconditionally for `"training.forgetting_threshold"`, the existing `test_the_allowlist_entry_is_still_needed` / `test_the_allowlist_does_not_cover_fields_that_are_consumed` guards are disarmed for that field **forever**. Whoever wires #799 properly later gets a green suite and a `KNOWN_UNCONSUMED` entry that lies — which that file's own docstring calls "the worst thing an allowlist can do... an entry that contradicts itself gives a false answer to the one question the file exists to answer".

Could you make the exception **prove itself** instead of asserting itself? A small AST test that fails when `forgetting_threshold` is read on a `*.training` receiver anywhere in `src/` fires exactly when it really is wired, and still ignores `ship.py`'s unrelated CLI parameter (a plain local, never `.training.`). About fifteen lines, and it makes my mutation red.

**Two smaller things:**

- `test_training_intelligence_documented_config_parses` reads `Path("docs/peft-and-efficiency.md")` relative to CWD, while every other test in the file uses `monkeypatch.chdir(tmp_path)`. It works only because pytest is invoked from the repo root. `Path(__file__).resolve().parents[1] / "docs" / ...` is robust.
- `soup autopilot` still writes all nine staged keys into the YAML at their defaults (`forgetting_eval_steps: 100`, `forgetting_threshold: 0.1`, `checkpoint_keep_top: 3`). They no longer *promise* anything and correctly produce no warning, but a reader of the generated file still sees knobs that do nothing. Fine to leave — worth a sentence in the PR body so the next reader knows it was considered.

`training.early_stop_patience`'s allowlist reason is still `"no issue yet -- ..."` while its seven siblings were retitled to `"#799 -- ..."`; it is tracked by #761, so updating the note in the same pass would be tidier.

**Heads-up on ordering, for me rather than for you:** this collides with **#906**, which bumps the same `KNOWN_UNCONSUMED` pin 39 to 45 while you bump it 39 to 40 — and both edit the same `_consumed_in_src` / `_declared` regions. Whichever lands second needs **46** and a re-run of the ratchet, plus a check on whether #906's called-ness refinement already neutralises `ship.py`'s `forgetting_threshold` read, in which case `UNRELATED_NAME_COLLISIONS` becomes redundant entirely. I will sequence those two deliberately.

Amix29:
Addressed the review at `f61aa2f`:

- replaced the permanent `UNRELATED_NAME_COLLISIONS` trapdoor with a receiver-qualified AST check that detects direct and `getattr(cfg.training, ...)` reads, while ignoring `ship.py`’s local name
- added positive and negative detector controls, so a future live `cfg.training.forgetting_threshold` read makes the stale-allowlist test fail
- made the documentation fixture path independent of pytest’s working directory
- linked `training.early_stop_patience` to #761
- documented in the PR body that Autopilot still serializes the staged defaults

Validation: `38 passed` for the two affected suites with `--no-cov`; full repository Ruff scope and `git diff --check` pass.

MakazhanAlpamys:
Verified at `f61aa2f`. The trapdoor is closed, and I checked it the only way that means anything — by re-running the mutation that survived last time.

I inserted a genuine live read at the top of `train()`:

```python
_live = cfg.training.forgetting_threshold
```

```
3 failed, 35 passed
FAILED test_the_allowlist_does_not_cover_fields_that_are_consumed
FAILED test_forgetting_threshold_is_declared_unconsumed
FAILED test_train_warns_for_nondefault_unwired_sibling_settings
```

On the previous head that same mutation passed **all 35**. So the exception now proves itself rather than asserting itself, and whoever wires #799 properly later will be told by the suite instead of inheriting a `KNOWN_UNCONSUMED` entry that lies. Handling `getattr(cfg.training, ...)` as well as the direct attribute, while still ignoring `ship.py`'s local name, is the right shape — it discriminates on the receiver rather than on the field name.

The docs fixture is now anchored on `__file__`, `early_stop_patience` is linked to #761, and the PR body records that Autopilot still serialises the staged defaults. All four items done.

**A note on ordering, which is mine to manage, not yours.** This still collides with **#906** — it bumps the same `KNOWN_UNCONSUMED` pin 39 to 45 while you bump it 39 to 40, and both edit the `_consumed_in_src` / `_declared` regions. Whichever lands second needs **46** and a re-run of the ratchet. There is also a real possibility that #906's called-ness refinement already neutralises `ship.py`'s `forgetting_threshold` read, in which case your receiver-qualified check becomes belt-and-braces rather than load-bearing — that would be a good outcome, not a wasted change, and I will check it at merge rather than ask you to.

Merging as soon as the checks report.

MakazhanAlpamys:
Merged as `8b79489d`. Thank you — and the part worth naming is not the wiring, it is what replaced your first draft.

The original patch kept a permanent `UNRELATED_NAME_COLLISIONS` list: field names allowed to look unwired. That works exactly until a real gate goes unwired and lands on the list by accident, at which point the guard grandfathers the bug it exists to find. The receipt inverts it — the generator records what it emitted, the guard checks that against the real consumer set, and nothing needs a standing exemption. You took that rewrite without arguing for the list you had already written.

Verified before merging: the guard fires when the wiring is removed, and it names the field rather than failing generically.

Your CONTRIBUTORS.md entry is in and #799 is closed.

PR #928, opened 2026-09-13
fix(profile): handle missing GPU detection, expand GPU table, and scale throughput

### LINKED ISSUE #830: soup profile says "Fits in 24 GB VRAM" on a machine with no GPU, rejects --gpu rtx3050/rtx5070/rtx5080/b200, and uses A100 throughput for every card

Guts1005:
Closes #830

### Summary

Resolves three issues in `soup profile`:
1. **Unmeasured fallback provenance**: When auto-detection fails (CPU-only PyTorch or no CUDA devices) and `--gpu` is omitted, the profiler explicitly labels the estimate as `Assuming 24 GB (no GPU detected; pass --gpu)` instead of presenting it as a measured `OK Fits in 24 GB VRAM`.
2. **GPU table & device string normalization**: Added lookup support for RTX 3050 (8GB/6GB), RTX 50-series (5060, 5060 Ti, 5070, 5070 Ti, 5080), Blackwell (B200, GB200), and mobile/laptop variants where VRAM differs from desktop parts. Added `normalize_gpu_name` to handle raw PyTorch device strings (e.g. `NVIDIA GeForce RTX 5070 Laptop GPU`) and hyphenated/spaced user flags.
3. **GPU-aware throughput scaling**: `estimate_speed` now scales estimated tokens/sec based on relative GPU throughput multipliers rather than returning fixed A100 throughput regardless of target card.

### Test Coverage

Added `TestIssue830GpuProfileDefects` in `tests/test_profile.py`:
- `test_issue830_gpu_lookup_and_variants`: Asserts memory lookup for RTX 3050, 5070, laptop variants, raw torch device strings, and B200.
- `test_issue830_estimate_speed_gpu_scaling`: Asserts throughput scaling differs across hardware tiers (e.g. T4 < A100 < H100).
- `test_issue830_no_gpu_detected_profile_warning`: Mocks no GPU detected and asserts the assumption warning is printed without presenting a false measured fit.


Guts1005:
Closing in favor of #920 which was already in flight and actively in review with the maintainer. Thanks!

PR #931, opened 2026-09-13
fix(train): configure RoPE before model construction

### LINKED ISSUE #793: rope_scaling_type is a no-op on transformers 5: inv_freq and logits are byte-identical after "Long-context enabled"
## The defect

`training.rope_scaling_type` (linear / dynamic / yarn / longrope / llama3) changes nothing about the model that trains. `trainer/sft.py` applies it **after** `from_pretrained`, by mutating `model.config`:

```

src/soup_cli/trainer/sft.py

1359    self.model = AutoModelForCausalLM.from_pretrained(cfg.base, **model_kwargs)

...

1367    # Long-context — apply RoPE scaling after model load

1368    if tcfg.rope_scaling_type:

1371        rope_config = apply_long_context_config(self.model.config, ...)

1377        if rope_config:

1378            console.print("[green]Long-context enabled:[/] RoPE ... scaling to ... tokens")

```

On transformers 5 the rotary embedding computes `inv_freq` and `attention_scaling` once, in `__init__`, from the config it was constructed with. Writing `config.rope_scaling` afterwards is never read again. The run prints "Long-context enabled" and trains with the base model's original RoPE.

Three further problems in the same path:

1. **The YaRN tunables are dropped.** `apply_long_context_config` (`utils/long_context.py:365-409`) calls `get_rope_scaling_config(scaling_type=..., target_length=..., original_length=...)` and never passes `yarn_factor` / `yarn_attn_factor` / `yarn_beta_fast` / `yarn_beta_slow`, although `get_rope_scaling_config` accepts all four (`:286-289`). The schema validates them (`config/schema.py:2775-2800`) and `_validate_yarn_fields_require_yarn_type` (`:3559-3562`) says it exists to avoid "silently dropping the values". `docs/peft-and-efficiency.md:127-139` shows a worked example setting all four.

2. **The dict it writes corrupts a saved config.** It uses the legacy `"type"` key and replaces `rope_parameters` wholesale, losing `rope_theta`. After a save/reload round trip, a `rope_theta=500000` model comes back with `rope_theta=10000` (reproduction below). Any path that serialises the model config, such as full fine-tuning with `lora.r: 0`, which ends in `trainer.save_model` (`sft.py:1972`), would write that.

3. **Only SFT reads the field.** `grep -rn rope_scaling_type src/soup_cli` finds `trainer/sft.py`, the `commands/train.py:1229` warning pre-flight, and the schema. `task: pretrain`, the natural task for context extension, accepts the field and ignores it entirely.

Also, `docs/peft-and-efficiency.md:142` says that omitting `rope_scaling_type` on a Llama 3.1 base makes `apply_long_context_config` auto-pick `llama3`. `sft.py:1368` only calls it when the field is set, so the auto-detect branch (`long_context.py:392-399`) is unreachable from `soup train`.

## Reproduction

Tiny random-init Llama, no download (transformers 5.17.0):

```python

m = LlamaForCausalLM(LlamaConfig(vocab_size=128, hidden_size=64, intermediate_size=128,

        num_hidden_layers=1, num_attention_heads=4, num_key_value_heads=4,

        max_position_embeddings=512)).eval()

inv0 = m.model.rotary_emb.inv_freq.clone(); out0 = m(ids).logits

apply_long_context_config(m.config, target_length=2048, rope_scaling_type=kind, model_name="x")

```

```

linear rope_config: {'type': 'linear', 'factor': 4.0}

  inv_freq changed: False attention_scaling: 1.0 -> 1.0 rope_type attr: default

  logits changed: False

yarn rope_config: {'type': 'yarn', 'factor': 4.0, 'original_max_position_embeddings': 512}

  inv_freq changed: False attention_scaling: 1.0 -> 1.0 rope_type attr: default

  logits changed: False

dynamic rope_config: {'type': 'dynamic', 'factor': 4.0}

  inv_freq changed: False attention_scaling: 1.0 -> 1.0 rope_type attr: default

  logits changed: False

control (linear x4 at construction) inv_freq differs from default: True

```

Save/reload of the mutated config (base built with `rope_theta=500000.0`):

```

fresh rope_parameters: {'rope_theta': 500000.0, 'rope_type': 'default'}

after apply rope_parameters: {'type': 'linear', 'factor': 4.0}

saved config.json rope keys: {'rope_parameters': {'factor': 4.0, 'rope_type': 'linear', 'type': 'linear'}}

reloaded rope_parameters: {'factor': 4.0, 'rope_type': 'linear', 'type': 'linear', 'rope_theta': 10000.0}

```

Moving the current dict before construction as-is would not work either:

```

construct with {'type': 'linear', 'factor': 4.0} failed: KeyError 'rope_type'

construct with {'rope_type': 'linear', 'factor': 4.0, 'rope_theta': 500000.0} -> inv_freq changed: True rope_type: linear

```

## Why it matters

Users who want long context are exactly the ones who set this, and the shipped configs do it for them: the `llama3.1-8b-longctx` recipe (`recipes/catalog.py:937`, `rope_scaling_type: yarn`) and `src/soup_cli/templates/longcontext.yaml:28` (`dynamic`, `max_length: 131072`). Those runs train at the long `max_length` on positions the model's RoPE was never extended for. Nothing fails, and the console confirms a feature that did not happen.

## Proposed fix

- Build the rope parameters **before** loading. Load `AutoConfig`, set `rope_parameters` in transformers-5 form (`rope_type`, and keep the existing `rope_theta` and any other keys), set `max_position_embeddings`, then `from_pretrained(cfg.base, config=config, ...)`.

- Forward `yarn_factor` / `yarn_attn_factor` / `yarn_beta_fast` / `yarn_beta_slow` from `TrainingConfig` into `get_rope_scaling_config`.

- Apply it in `pretrain.py` too, or reject `rope_scaling_type` at parse for tasks that do not read it.

- Either call the auto-detect path when the field is unset on a Llama 3.1 base, or correct the docs sentence.

## Acceptance criteria

- [ ] An effect test, not a dict test: build a tiny Llama through the real load path with `rope_scaling_type: linear` and assert `rotary_emb.inv_freq` differs from the unscaled model. It must fail on current code.

- [ ] A YaRN test asserting that `yarn_beta_fast` / `yarn_attn_factor` change `inv_freq` / `attention_scaling`.

- [ ] A save/reload test asserting `rope_theta` survives.

- [ ] `task: pretrain` either honours the field or refuses it with a message that names it.

- [ ] `docs/peft-and-efficiency.md` matches the behaviour.

---

Amix29:
## Summary
- apply Transformers 5 `rope_parameters` before SFT/pretrain model construction
- preserve checkpoint-native RoPE values and forward all YaRN tunables
- reject LongRoPE without model-native learned factor vectors
- add regression coverage for instantiated rotary embeddings and save/reload preservation

## Validation
- `411 passed` in the long-context/trainer/config regression selection
- full suite reached `21278 passed, 106 skipped, 5 deselected`; after the one issue-793 traceability failure was fixed, the two remaining failures were reproduced unchanged on `origin/main` (Rich rendering assertions in issue #204 and #885 tests)

No CUDA path was used.

Closes #793


MakazhanAlpamys:
Reviewed at `a1c520e4`. **The core fix is right and I proved it three ways** — but the new merge logic reintroduces this issue's own failure shape on two shipped model families, and I reproduced both end-to-end through the real `soup train` path rather than by reading.

**What is solid.** The baseline is genuine: your test file, copied onto `origin/main`, gives `5 failed` with `assert rotary.rope_type == "linear"` -> `'default'`, which is #793 verbatim. `AutoConfig` -> mutate -> `from_pretrained(config=...)` is the correct shape. The tests assert the *effect* (`rotary_emb.inv_freq`, `attention_scaling`) rather than the dict, which is what the acceptance criteria asked for. Refusing LongRoPE instead of inventing learned factor vectors is the right call and matches house style.

**Mutations, each anchor counted, each mutated module actually imported before the run** (an import, because a `NameError` gives a false "caught"):

| mutation | result |
|---|---|
| `model_kwargs["config"] = model_config` -> `pass` (1 occurrence in each of sft.py / pretrain.py) | **4 failed** — `'default' == 'linear'` on sft, on pretrain, on save/reload, and `KeyError: 'factor'` for the yarn tunables |
| remove the whole LongRoPE guard block (1 occurrence) | **1 failed** — `DID NOT RAISE ValueError` |
| remove the four `yarn_*=tcfg.yarn_*` kwargs | **2 failed** — the #748 ratchet names all four fields by name |

That third one is why **the `-8` in `test_issue748_config_fields_reach_a_consumer.py` is a legitimate retirement, not a weakening**: the guard re-arms the moment the wiring goes. Same for the pinned count `39 -> 35` — exactly minus four, which that file's own docstring requires in the same commit. I went through all 25 removed test lines: 17 are `["type"]` -> `["rope_type"]` renames with **17 added** in the same files, perfectly symmetric, no assertion dropped.

---

**BLOCKING 1 — nested `rope_parameters` (Gemma 3) silently no-ops, prints "Long-context enabled", and bumps `max_position_embeddings` anyway.** `utils/long_context.py:435-437`.

`merged.update(rope_config)` assumes the block is flat. Models with per-layer-type RoPE keep nested blocks, so your scaling is bolted on *beside* them and never read. Through the real `SFTTrainerWrapper._setup_transformers` on a tiny Gemma 3:

```
Long-context enabled: RoPE linear scaling to 256 tokens
config.rope_parameters: {'full_attention': {...'rope_type':'default'},
                         'sliding_attention': {...'rope_type':'default'},
                         'rope_type': 'linear', 'factor': 4.0}
rotary rope_type: {'sliding_attention': 'default'}
  sliding_attention_inv_freq: changed = False      <- NOT SCALED
max_position_embeddings: 256                        <- but trains at 4x length
```

This is #793 exactly — success banner, no effect — on a family with **20 `gemma3` entries in `recipes/catalog.py`**, and it is worse than a plain no-op: `max_position_embeddings` *is* bumped, so the run trains on positions the RoPE was never extended for. I surveyed the families: Gemma3Text is nested; Llama/Qwen2/Qwen3/Mistral/Gemma2/Phi3/Qwen3Moe/Cohere2/Llama4Text/GptOss/Qwen3Next/Glm4/MiniMax are flat.

Refuse a nested block with a named error — consistent with your own LongRoPE refusal and with the repo's refuse-don't-approximate rule — or apply into each nested block. Either way it needs a test.

**BLOCKING 2 — foreign rope-type tunables survive the merge, and transformers is already telling us.** `utils/long_context.py:435-436`. `merged.pop("type", None)` drops the legacy key only, not the previous `rope_type`'s tunables. A Llama-3.1 base ships a native `llama3` block; set any other type and:

```
[transformers] Unrecognized keys in `rope_parameters` for 'rope_type'='linear':
               {'high_freq_factor', 'original_max_position_embeddings', 'low_freq_factor'}
saved config.json: {'factor': 4.0, 'high_freq_factor': 4.0, 'low_freq_factor': 1.0,
                    'original_max_position_embeddings': 64, 'rope_theta': 500000.0,
                    'rope_type': 'linear'}
```

This is on a **shipped recipe**: `recipes/catalog.py:936`, `llama3.1-8b-longctx`, sets `rope_scaling_type: yarn` on `meta-llama/Llama-3.1-8B`. Point 2 of the issue was precisely "the dict it writes corrupts a saved config", so that half is not closed. Keep only the rope-type-agnostic keys (`rope_theta`, `partial_rotary_factor`) from `existing` and drop the rest before `update`.

---

**Two smaller items:**

- **`changelog.d/0.75.0/931.fixed.md` reads `(#931 by @Amix29)`** — the issue number is lost. House style when a distinct issue exists is `(#793 by @Amix29 in #931)`; `727`, `729`, `742`, `872`, `882`, `893` and `916` all do it that way. The content is also markedly thinner than its siblings — no mechanism, no test names. Given the fix is a load-bearing ordering change, it deserves a sentence about *why* configuring RoPE after construction was a no-op.
- **`docs/peft-and-efficiency.md:142` overclaims scope.** It says "applied ... for both `task: sft` and `task: pretrain`" with no caveat, but this is the transformers **text** path only: `sft.py:588-597` dispatches vision / audio / streaming / unsloth to other `_setup_*` methods, none of which read the field, and every other task still accepts `rope_scaling_type` with no consumer and no schema gate. Not a regression — but naming the path costs one clause, and leaving it unnamed is how the next person files the same issue.

CI is queued rather than green on `a1c520e4`, and the branch is 7 commits behind `main`; I need green on the post-rebase head before merging.

`target_length <= original_length` returning `None` silently is pre-existing and you preserved it faithfully — not asking you to fix it here, but it is the same class as #793 and worth an issue if you want it.

Amix29:
Addressed both RoPE blockers at `285ad1e`, after rebasing onto current `origin/main@876b2b1`.

- nested per-layer `rope_parameters` such as Gemma 3 are now refused before either `rope_parameters` or `max_position_embeddings` can be mutated
- changing RoPE algorithms now preserves only `rope_theta` and `partial_rotary_factor`; prior algorithm tunables are removed before the new block is applied
- the new Gemma 3 test asserts the refusal is atomic, and the Llama 3 to linear test pins the exact cleaned mapping
- docs now scope the live path to Transformers text SFT/pretrain and name unsupported dispatch paths
- the changelog now explains the construction-order failure and credits `#793 by @Amix29 in #931`

Post-rebase validation: `228 passed` across the RoPE/long-context surfaces; ruff clean; diff check clean.

MakazhanAlpamys:
Re-checked at `285ad1e5`. **Both blockers closed, verified end-to-end through the real `SFTTrainerWrapper._setup_transformers` rather than from the diff.** Waiting only on CI.

**Blocker 1 — and the part that matters is that the refusal is atomic.**

```
ValueError: nested rope_parameters are not supported safely; model-specific sections found:
  full_attention, sliding_attention. Soup refuses to change max_position_embeddings
  without scaling every RoPE section
```

Raised at `sft.py:1368`, i.e. **before** `AutoModelForCausalLM.from_pretrained` — so no banner is printed, no model is built, `config.rope_parameters` is untouched, and **`max_position_embeddings` is still 64, not bumped**. That was the specific thing I flagged: the old behaviour raised the position budget while leaving the frequencies unscaled, which is worse than a no-op because the run then trains on positions the RoPE was never extended for. The raise now happens above that assignment, so there is no window where one half is applied.

I also checked the blast radius before accepting a hard failure, since "refuse" is only right if it does not break shipped configs: **exactly one recipe sets `rope_scaling_type`** — `catalog.py:937`, `llama3.1-8b-longctx`, `yarn` on a flat Llama family — despite there being 20 `gemma3` entries. So nothing shipped turns into a hard error; it bites only hand-written configs on nested families, which is the intended fail-loud.

**Blocker 2**, all four switch targets on a Llama-3.1-shaped base carrying the native `llama3` block, with the `transformers` logger captured directly (the message comes from `modeling_rope_utils.py:1086`):

| type | `Unrecognized keys` warnings | saved `rope_parameters` |
|---|---|---|
| `linear` | **0** | `{factor, rope_theta, rope_type}` — no foreign keys |
| `dynamic` | **0** | same shape |
| `yarn` (the shipped recipe) | **0** | `{factor, original_max_position_embeddings, rope_theta, rope_type}` |
| `llama3` | **0** | keeps `high_freq_factor`/`low_freq_factor` — correct, they are native to that type |

Reloading each saved directory with a fresh `from_pretrained` also gives zero warnings, and `rope_theta` survives because it is in `_ROPE_TYPE_AGNOSTIC_KEYS`, which is the point. The `llama3` row is the one I would have missed: keeping the tunables there is right, and a blunter fix that stripped everything non-agnostic would have broken it.

**Both mutations still discriminate post-rebase.** `model_kwargs["config"] = model_config` -> `pass` (1 occurrence in each of `sft.py:1379` and `pretrain.py:341`, both modules really imported) fails 4 tests — `'default' == 'linear'` on sft, on pretrain, on save/reload, and `KeyError: 'factor'` for the yarn tunables. Removing the LongRoPE guard block fails with `DID NOT RAISE ValueError`. Failures are on rotary module state and reloaded config.json, not on source text.

258 passed across the new file and the four pre-existing ones it touches; ruff clean. All 25 removed test lines accounted for: 17 `["type"]` -> `["rope_type"]` renames against exactly 17 added in the same files, and the four `KNOWN_UNCONSUMED` retirements with the pin moved `39` -> `35`, which the yarn end-to-end confirms is a retirement rather than a weakening (`factor 4.0`, `attention_factor 1.5`, `beta_fast 24`, `beta_slow 2` all land, and `rotary.attention_scaling == 1.5`).

**Changelog and docs both fixed**, and the docs went further than I asked: the same paragraph previously promised that omitting `rope_scaling_type` would auto-pick `llama3`, which `soup train` never does — both `sft.py:1360` and `pretrain.py:322` gate on `if tcfg.rope_scaling_type:` against a schema default of `None`. You corrected an overclaim I had not flagged.

**No new findings.** Two notes, neither a request: `target_length <= original_length` still returns `None` silently and short-circuits *before* the nested check, so a Gemma 3 config with a short `max_length` returns `None` rather than refusing — correct, since there is nothing to apply, but it means the refusal is only reachable when scaling is actually requested. And the nested refusal surfaces through the generic crash-bundle handler rather than `utils/errors.py`, exactly as the pre-existing LongRoPE refusal does, so it is consistent rather than a regression.

Your base is now six commits behind `main`. Nothing in the delta touches these files, so no rebase is needed for correctness — I just need the checks green on the current head.

MakazhanAlpamys:
@Amix29 — base move, not a review item.

**Conflicting** on one file:

```
```

caused by `8b79489d` (#927), which edited the recorded-leak table in the same region. One-file rebase; everything else auto-merges cleanly.

Amix29:
Update pushed in 2c56b88 after rebasing on current main. The consumer-ratchet conflict is resolved from the current main baseline of 40 known-unconsumed fields: this PR consumes the four YaRN fields, so the exact expected count is now 36. Validation: 282 targeted tests passed across the #793 regression suite, consumer ratchet, performance, related milestone coverage, and changelog validation; full Ruff is clean. GitHub reports the branch mergeable. Ready for re-review.

MakazhanAlpamys:
Reviewed `2c56b88b`. **Merge-ready, waiting only on CI.** Nothing needed from you.

The delta is exactly the one-file rebase I asked for and nothing else. I isolated it properly rather than reading `git diff reviewed..head` — both ends were rebased, so that raw diff is 130 files of main's own churn. Comparing each end's patch against its own merge base: **29 hunks before, 29 after**, 16 lines of patch content changed, all in `tests/test_issue748_config_fields_reach_a_consumer.py`. Main's baseline moved 39 to 40 (#927 added an entry and re-annotated `training.forgetting_eval_steps`), your pin moved 35 to 36. **Still exactly −4.** The other four test files are byte-identical to the reviewed revision.

I checked the resolution did not eat #927's content: the four YaRN entries and the pin are the only differences from main's blob. Nothing of theirs was dropped.

**Four mutations, all on the merged tree, all still discriminating on behaviour rather than source text:**

| mutation | result |
|---|---|
| `model_kwargs["config"] = model_config` to `pass` in both `sft.py` and `pretrain.py` | 4 failed — `assert 'default' == 'linear'` on the **instantiated** `rotary.rope_type` and on the **reloaded** `config.json` |
| delete the nested-`rope_parameters` raise | 1 failed — `DID NOT RAISE ValueError` |
| `merged = dict(existing)` (pre-fix blocker-2 behaviour) | 1 failed, naming the leaked keys: `{'high_freq_factor', 'low_freq_factor', 'original_max_position_embeddings'}` |
| delete the four `yarn_*` kwargs from both trainers | 2 failed, naming all four fields |

That last one is the proof that 40 to 36 is a **retirement and not a weakening**: pull the wiring and the ratchet re-arms by name.

Of the 25 removed test lines, 16 are `["type"]` to `["rope_type"]` renames for Transformers 5 with a 1:1 replacement in the same test, one is the `rope_scaling` to `rope_parameters` setup line, and the exact-equality assertions stayed exact-equality. Nothing loosened.

494 passed on branch and merged tree on the focused set; 2796 passed, 49 skipped across every file touching the SFT/pretrain wrappers. Ordering re-verified on current main: the nested refusal raises **before** `max_position_embeddings` is written and before `model_kwargs["config"]`, and the banner prints after `from_pretrained`. Ruff clean.

Incidentally #945 edits `docs/peft-and-efficiency.md` four lines from yours and they do **not** conflict — I materialised the double merge and got 330 passed. Either order works.

MakazhanAlpamys:
Merged as `2dd84c24`. Thank you.

**The mutations are what sold this one**, because all four discriminate on *behaviour* rather than on source text — which is the difference between a test that pins a fix and a test that pins a diff:

| mutation | what failed |
|---|---|
| drop `model_kwargs["config"] = model_config` in both trainers | 4 failed on `assert 'default' == 'linear'`, read off the **instantiated** rotary module and the **reloaded** `config.json` |
| delete the nested-`rope_parameters` raise | `DID NOT RAISE ValueError` |
| `merged = dict(existing)` | the failure **names the leaked keys** |
| delete the four `yarn_*` kwargs | the consumer ratchet re-arms **by name** |

That last one is the proof your 40 -> 36 pin is a retirement rather than a weakening — pull the wiring and the guard comes back. I checked the rebase resolution did not eat #927's content: the four YaRN entries and the pin are the only differences from main's blob.

Of the 25 removed test lines, 16 are `["type"]` -> `["rope_type"]` renames for Transformers 5 with a 1:1 replacement in the same test, and the exact-equality assertions stayed exact-equality. Ordering re-verified on current main: the nested refusal raises before `max_position_embeddings` is written.

Before merging I composed this with the other nine PRs that were ready and ran the affected suites plus the repo-wide ratchets on that tree: **967 passed, zero failures**. That step is not ceremony — it caught a real break in a tenth PR (#963), which I held rather than merged. Yours composes with #908, which touches `trainer/sft.py` and `trainer/pretrain.py` in the same region.

Credited in `CONTRIBUTORS.md` (`f491ac5b`). #793 closed with credit.

PR #934, opened 2026-09-13
fix(layer-stream): bound the sharder's live source mappings so a large checkpoint can be sharded on Windows

### LINKED ISSUE #926: shard_checkpoint dies with an access violation on Windows once ~100 GB of source safetensors mappings are alive (a real 70B cannot be sharded)
## Summary

`shard_checkpoint` (`src/soup_cli/utils/layer_shard.py`, pass 2) enters every source shard into one `contextlib.ExitStack` and keeps all the memory maps alive for the whole pass. On Windows that pattern dies with `Windows fatal exception: access violation` (no Python traceback, exit 139, no Event Log entry, no crash dump) once roughly **96–106 GB of safetensors mappings are alive in one process, whatever the file count**. A real Llama-3.1-70B checkpoint is 30 files / ~141 GB, so **it cannot be sharded on Windows at all**; a 32B bf16 (~65 GB) is under the line, a 72B is not.

Found while measuring the cold disk tier for #325. Record: `benchmarks/probe-rtx5070-what-bounds-streaming.md` §12–§14; harness: `benchmarks/harness/mmap_limit_repro.py` (both land on `main` in the commit that follows this issue; the script is also inlined below).

## Environment

RTX 5070 Laptop · Windows 11 Pro 26100 · 31.7 GB RAM, system commit limit 36.7 GB · Python 3.12.10 · torch 2.14.0+cu130 · safetensors 0.8.0 · Soup `main` at 2a97512a (v0.75.0 + 0).

## Reproduction — without Soup, CUDA or bitsandbytes

A Llama-70B-shaped synthetic checkpoint (random weights, real shapes; 137.98 GB bf16; `benchmarks/harness/synth_checkpoint.py`) written twice: 80 files of 1.71 GB and 20 files of 6.85 GB. The script opens the files with `safe_open`, calls `get_tensor` on every key and keeps nothing.

| pattern | 80 x 1.71 GB | 20 x 6.85 GB |
|---|---|---|
| one `ExitStack` holds ALL handles (the shipped pass-2 pattern) | access violation after file 61 (62 files done, **106.1 GB** mapped, 26 s) | access violation after file 13 (14 files done, **95.8 GB** mapped) |
| open / read / close each file in turn | OK, all 80 files, 3 s | OK |

The shipped sharder itself died twice at the same place: `layer_000..060` written (26.9 GB), then the access violation. `faulthandler` places it in `torch/storage.py:480 __getitem__`, reached from `layer_shard._read_tensor` → `safe_open(...).get_tensor(key)`.

<details>
<summary><code>mmap_limit_repro.py</code></summary>

```python
import contextlib, faulthandler, glob, os, sys, time
from safetensors import safe_open

def main(mode: str, directory: str) -> int:
    faulthandler.enable(all_threads=True)
    paths = sorted(glob.glob(os.path.join(directory, "*.safetensors")))
    started = time.perf_counter()
    total = 0

    def visit(handle, label):
        nonlocal total
        for key in handle.keys():
            tensor = handle.get_tensor(key)
            total += tensor.numel() * tensor.element_size()
            del tensor
        print(f"{label}  {total / 1e9:7.1f} GB mapped  {time.perf_counter() - started:6.0f} s", flush=True)

    if mode == "all":  # the shipped sharder's pattern: every handle alive at once
        with contextlib.ExitStack() as stack:
            handles = [stack.enter_context(safe_open(p, framework="pt")) for p in paths]
            for index, handle in enumerate(handles):
                visit(handle, f"file {index:3d} done")
    else:  # per file: open, read, close
        for index, path in enumerate(paths):
            with safe_open(path, framework="pt") as handle:
                visit(handle, f"file {index:3d} done")
    print(f"OK {mode}: {total / 1e9:.1f} GB in {time.perf_counter() - started:.0f} s")
    return 0

if __name__ == "__main__":
    sys.exit(main(sys.argv[1], sys.argv[2]))
```

</details>

## What is established, and what is not

- **It is bytes, not files:** 14 handles and 62 handles crash at the same ~96–106 GB of live mappings.
- **It is not commit exhaustion:** the commit limit is 36.7 GB, and read-only file mappings do not charge commit.
- **It is not resident memory:** `get_tensor` is a zero-copy view on this stack (returns in 0.000 s; repeat calls share a `data_ptr`); with no consumer touching the pages, the per-file variant "read" 137 GB in 3 s.
- **The mechanism is NOT established** — what Windows does differently past ~100 GB of concurrently mapped sections. Its value is.
- Every published sharding ran on 4 (8B) to ~37 (72B, on Linux) source files; the Windows box had never sharded more than 4.

## Fix shape

Open a source file only while a layer needs it: an LRU of one or two live handles (layers are contiguous in files; pass 1 already uses the per-file pattern), and make `_read_tensor` / `_read_raw_tensor` return a tensor that **owns its memory** — `get_tensor` is a view over the mapping, and the sharder keeps norms / embeddings / the head until it writes them, so a view over a released mapping is a use-after-unmap.

`benchmarks/harness/stream_probe.py --lazy-shard-handles` is exactly that shape applied from outside, for the duration of `shard_checkpoint` only: with it the 20-file / 138 GB checkpoint sharded to 36.39 GB of NF4 in 216.9 s, and the runtime plus every timed number in the record are shipped code.

Related, out of scope here: the runtime `DiskSource` holds one handle per decoder layer for the whole run — 80 x 441 MB = 35 GB of mappings for a 70B NF4 store, under the line; a 400B-class store would not be.

## Acceptance criteria

- [ ] `shard_checkpoint` never holds more than a small fixed number of source `safe_open` handles alive (a counting `safe_open` stub over a multi-file tiny checkpoint; mutation check: the all-handles pattern must fail the test)
- [ ] `_read_tensor` / `_read_raw_tensor` return tensors that do not alias the source mapping
- [ ] Sharding a multi-file checkpoint produces byte-identical shards to sharding the same tensors from one file
- [ ] The 20-file 138 GB fixture shards on this Windows box **without** `--lazy-shard-handles`, byte-identical to the cache the workaround produced
- [ ] `--lazy-shard-handles` marked obsolete in the harness once the fix ships

MakazhanAlpamys:
Closes #926.

`shard_checkpoint` entered every source `*.safetensors` into one `ExitStack` and kept all of them open for the whole of pass 2. On Windows that made sharding a large checkpoint impossible: a real Llama-3.1-70B is 30 files and ~141 GB, and the run died with `Windows fatal exception: access violation` — no Python traceback, exit 139, no crash dump, no Event Log entry.

## The mechanism, measured rather than assumed

The issue's first report got this wrong and said "read-only file mappings do not charge commit", which is why the threshold was reported as a number without an explanation. It is false on this stack: **`safe_open` maps the file PRIVATELY, so every live handle charges Windows commit for that file's entire size.**

Opening the 20-file / 138 GB fixture one file at a time, reading nothing, printing the commit CHARGE (`limit - available`, the quantity that survives Windows growing the pagefile mid-measurement):

| mapped | commit charge | delta | free physical |
|---|---|---|---|
| 0 (baseline) | 31.27 GB | — | 13.50 GB |
| 14.76 GB | 42.20 GB | +10.94 GB | 12.91 GB |
| 28.45 GB | 56.05 GB | +24.78 GB | 12.94 GB |
| 48.99 GB | 77.44 GB | +46.17 GB | 12.92 GB |
| all closed | 35.21 GB | +3.94 GB | — |

The charge tracks the mapped bytes almost exactly and returns on close, while free physical memory never moves — nothing is read, so this is the cost of mapping alone. A second, independent confirmation: the same repro run while the box was busy raised a clean `OSError: The paging file is too small for this operation to complete. (os error 1455)` — `ERROR_COMMITMENT_LIMIT` — instead of an access violation, because the allocation failed at map time rather than at fault time.

So the sharder was charging the whole checkpoint's bytes to commit, and the crash was the pagefile failing to keep up. That also explains the otherwise odd shape of the original finding: the threshold tracked total BYTES mapped, not the number of files (62 x 1.71 GB and 14 x 6.85 GB both died, at ~106 GB and ~96 GB).

## The change

**`_SourceHandles`** — an LRU of at most `_MAX_LIVE_SOURCE_HANDLES = 2` live handles, replacing the all-open `ExitStack`. Two handles charge two files instead of the checkpoint. Only discovered shards are accepted; membership, iteration and `len` never open anything; `close()` attempts every release and logs the failures it cannot propagate.

**Forced tensor ownership** — `_read_tensor` / `_read_raw_tensor` now return tensors that own their memory, via `_own`. This is the other half, and it is not a use-after-unmap guard: a `get_tensor` view still reads correctly after its handle's `__exit__`, because the view's storage owns the mapping. That is exactly the problem — a retained view keeps its whole file mapped, and therefore its whole file's commit charge, past the handle's release, which would defeat the cap. The sharder holds norms, embeddings and the head until it writes them.

The copy is a real, accepted cost: a read whose dtype already matches used to pass the mapping through untouched and now copies it, so pass 2 moves roughly the checkpoint's own bytes an extra time, bounded to one layer at a time. It buys the bound — a 1.71 GB tensor held as a view charges its 6.85 GB file, the copy charges 1.71 GB.

The `Mapping` mixins that would open files behind the caller (`get`, `items`, `values`, equality) raise `TypeError` rather than being inherited: `get` reads as a side-effect-free peek and would not be one.

## The gate

The 20-file / 138 GB fixture re-sharded with **no workaround**, on an idle machine:

| | |
|---|---|
| layers | **80 / 80** (the shipped code cannot finish this checkpoint at all) |
| time | **221.4 s**, against 216.9 s with the old harness-side workaround — the ownership copy costs ~2% |
| output | **83 files / 36.39 GB, byte-identical** to the cache the workaround produced, `index.json` included |
| baseline commit charge | 27.16 GB of a 39.44 GB limit, recorded beside the result |

## Verification

- 20 new tests in `tests/test_issue926_sharder_source_handles.py`, mutation-checked: restoring the all-open `ExitStack` fails only the handle-peak test; making `_own` return the view fails only the two ownership tests.
- Every existing test file that calls `shard_checkpoint` (10 files) passes, except the two pre-existing `test_issue385_stream_dtype.py` NF4 bit-exactness failures on this Blackwell card, which are #776 and were confirmed to fail identically without this diff.
- `benchmarks/harness/stream_probe.py --lazy-shard-handles` is marked obsolete in its help text, and kept so the 2026-09-12 record can still be re-run as it was taken.
- The measurement record carries dated corrections rather than silent edits, as the other records in `benchmarks/` do.

**Measurement note for anyone re-running the real gate:** this crash is a whole-machine resource limit, so a run competing with other processes for commit measures those processes too. A first attempt here crashed after 4 layers with a baseline commit charge of 31 GB out of a 39 GB limit, held by three unrelated agents on the same box; that run said nothing about the fix. Record the baseline charge beside the result.


MakazhanAlpamys:
Merged as 22ad002d after an independent read-only review of the diff against #926 (report kept locally with the #927 work). The fix stands on its own: the cap is a structural invariant of `_SourceHandles` (it evicts before it opens), every `handles` use in pass 2 is a subscript consumed immediately, and `_own` is a bit-preserving `clone()` in exactly the case that used to return a mapping view — so the bound holds and the output bytes cannot move. No test was weakened; the anti-decay guards (`peak == cap`, `1 <= cap <= 2`, `_N_FILES > cap`) cover each other's blind spots.

Follow-ups the review left, none blocking:

- `benchmarks/README.md` lines 52 and 55 still describe all-handles-open as "the shipped pass-2 pattern" and `--lazy-shard-handles` as "the documented workaround", in the present tense. Both sentences are rewritten on the #927 branch, which already edits that table; the gate numbers from this PR's body (80/80 layers in 221.4 s, 83 files / 36.39 GB byte-identical) are referenced from there rather than left only in a PR body.
- `_MAX_LIVE_SOURCE_HANDLES` is bound as a default argument, so monkeypatching the module constant does not change the cap; a test that sets the cap to 1 and expects the whole file to stay green would be the direct check of "never wrong, only slower".
- No test shards a layer spread over more files than the cap; `dict(handles)` still opens everything while `items()` / `values()` are refused; `__exit__` can mask the unwinding error.

PR #935, opened 2026-09-13
fix(plugins): enable lazy entry-point discovery

### LINKED ISSUE #819: The plugin system is unreachable from the CLI: load_plugins() has no caller, so `soup plugins` always lists nothing and `soup train` never fires a plugin hook
## The defect

`soup_cli/plugins/__init__.py` has a working registry and `monitoring/plugin_callback.py` has a working `SoupPluginCallback`, attached by 16 trainer wrappers (sft, dpo, grpo, kto, orpo, simpo, ipo, bco, ppo, pretrain, ...) through `utils/peft_wiring.attach_plugin_callback`. What is missing is the step that puts anything in the registry.

1. **`load_plugins()` (`plugins/__init__.py:283`) has no caller in `src/`.** The only references are its definition, `__all__`, and two tests. Nothing ever imports a plugin module, so the in-memory `_PLUGINS` dict (`plugins/__init__.py:86`) is empty in every `soup` process. `build_plugin_callback()` then returns `None` (`monitoring/plugin_callback.py`, "no enabled plugins implement any hook"), and `attach_plugin_callback` attaches nothing.

2. **There is no discovery beyond the package's own directory.** `load_plugins()` walks `pkgutil.iter_modules(soup_cli.plugins.__path__)` only — no entry points, no config key, no path option.

3. **Enable/disable state lives for one process.** `enable_plugin` / `disable_plugin` rebuild a `PluginSpec` in the same in-memory dict; nothing is written to disk. Even with discovery fixed, `soup plugins disable foo` could not affect a later `soup train`. (Every registered plugin is also enabled by default, `PluginSpec.enabled = True`.)

4. **`soup plugins install` is a stub that exits 0** and tells the user to write into the installed package (`commands/plugins.py:37-46`), which is inside `site-packages` for every non-editable install and is replaced on upgrade.

5. **`templates=` / `model_groups=` are stored and read by nothing.** No code outside `plugins/__init__.py` reads `PluginSpec.templates` or `.model_groups`.

## Reproduction

A plugin module placed on the directory `load_plugins()` scans (by appending a directory to `soup_cli.plugins.__path__`, so no repo file is touched):

```python

# extra/hello.py

from soup_cli.plugins import register_plugin

class HelloPlugin:

    def pre_train(self, ctx): print("HELLO pre_train")

register_plugin(name="hello", version="1.0.0", plugin=HelloPlugin(), description="demo")

```

```

registered before load_plugins(): {}

build_plugin_callback() -> None

load_plugins() -> 1

registered after load_plugins(): ['hello']

build_plugin_callback() -> SoupPluginCallback

```

So the machinery works the moment `load_plugins()` is called — and no CLI path calls it. From the CLI:

```

$ soup plugins

No plugins registered.

exit=0



$ soup plugins enable hello

Unknown plugin: hello

exit=1



$ soup plugins install hello

Plugin install for hello is advisory in v0.45.0; live install lands in v0.45.1.

Drop your plugin module under soup_cli/plugins/ and call register_plugin(...) at import time.

exit=0

```

## Why it matters

`docs/backends-and-ops.md:944` says: "Drop a Python module under `src/soup_cli/plugins/` (or any package importable by Soup) and register at import time", followed by `soup plugins enable foo`. Neither half works: a module under `soup_cli/plugins/` is never imported, and "any package importable by Soup" has no mechanism at all. The same page (`:971`) still says trainer-callback wiring "lands in v0.45.1", while #101 and #105 were closed as shipped. The only way a hook fires today is from a user's own Python script that imports its plugin and then drives a trainer wrapper directly — which is not the documented CLI.

## Proposed fix

1. Discover plugins through the standard mechanism: an `importlib.metadata` entry-point group (e.g. `soup_cli.plugins`), plus the existing package-directory scan. Call the loader once from the CLI entry point (or lazily from `soup plugins` and `attach_plugin_callback`) so both listing and training see the same set. Keep it off the light-command startup path (`tests/test_cli_startup_is_light.py`).

2. Persist enable/disable to `~/.soup/plugins.json` (atomic write), read by both the CLI and the trainer. Decide and document whether a newly discovered plugin is enabled by default — loading third-party code into a training run is an opt-in decision.

3. Replace `soup plugins install` with either a real `pip install` wrapper or remove it; until then it should exit non-zero rather than 0.

4. Either consume `templates` / `model_groups` or drop them from the public `register_plugin` signature and docs.

5. Fix `docs/backends-and-ops.md` to describe the mechanism that ships.

## Acceptance criteria

- [ ] A test installs a fake distribution with a `soup_cli.plugins` entry point and asserts `soup plugins` lists it — fails today ("No plugins registered.").

- [ ] A test asserts `attach_plugin_callback` attaches a callback for that plugin without the test calling `load_plugins()` itself — fails today.

- [ ] `soup plugins disable hello` followed by a fresh process shows `hello` disabled.

- [ ] `soup plugins install` no longer exits 0 while installing nothing.

---

Amix29:
## Summary
- discover `soup_cli.plugins` entry points lazily from plugin-aware CLI and trainer paths
- default third-party plugins to disabled and persist explicit enable/disable choices atomically in `~/.soup/plugins.json`
- make the install placeholder exit non-zero and document the supported packaging contract
- surface template/model-group metadata in `soup plugins`
- add real fake-distribution and fresh-process regression coverage

## Validation
- `229 passed, 1 skipped` in plugin registry/callback/acceptance tests
- `51 passed` in light-startup and changelog guards
- full suite: `21296 passed, 106 skipped, 5 deselected`; the two remaining Rich-rendering failures reproduce on clean `origin/main`

Rebased onto current `origin/main`. No CUDA path was used.

Closes #819


MakazhanAlpamys:
Reviewed at `3651149a`. **The discovery fix is correct, genuinely lazy, and all four mutations died.** Two things in the code this PR wrote need fixing, and there is one design question that is the security posture #819 actually asked about, which I want to decide explicitly rather than let ship by default.

**Startup lightness — verified directly, not assumed.** I spied on `load_plugins` and ran the real CLI through `runpy`:

| command | torch loaded | transformers | `load_plugins()` calls |
|---|---|---|---|
| `soup --help` | False | False | **0** |
| `soup version` | False | False | **0** |
| `soup recipes list` | False | False | **0** |
| `soup plugins` | False | False | 1 |

Min-of-5 subprocess timing, same box: `--help` 0.657s -> 0.664s, `plugins` 0.580s -> 0.591s. No #320-class regression. I confirmed it end-to-end with a real installed fake distribution: `soup plugins` imported the third-party module, `soup version` and `soup --help` did not.

**Baseline.** A direct run on `main` errors on the missing symbol, which only proves the symbol is new — so I re-ran with `raising=False` to get a behavioural baseline. All 7 fail for exactly #819's reasons (`assert 'hello' in 'No plugins registered.'`, `'Unknown plugin: hello'` rc=1, `assert 0 != 0` for the install stub). Maps 1:1 onto the three acceptance criteria.

**Mutations — 4 of 4 caught**, each anchor counted once, each mutated module re-imported before the run:

| mutation | result |
|---|---|
| `for entry_point in entry_points:` -> `in ():` | **4 failed** |
| third-party `default=False` -> `True` | **1 failed** — `assert 'disabled' in ...` got `enabled` |
| the except-branch default (partially-failed registrar) | **1 failed** |
| delete both `load_plugins()` call sites — the literal #819 defect | **3 failed** |

All 8 removed test lines accounted for: two fixture signatures replaced by versions that *add* state-file isolation, and `test_plugins_cli_install_advisory` renamed to assert `exit_code == 2` / `"does not install"`, which is the intended change and is independently re-pinned in the new file. **No coverage lost.**

The atomic write is properly atomic — `mkstemp` in the target directory, `fchmod 0600`, `flush`, `fsync`, single `os.replace`, `finally` unlink. I verified no leftover temp files. Eleven hostile JSON variants (list root, 20k-key bomb, NUL in name, traversal-shaped key, BOM) all degrade to `{}` with a warning and execute nothing.

---

**1. MEDIUM, and it can kill a training run: `RecursionError` escapes the handler whose job is to ignore malformed state.** `plugins/__init__.py:138` — the except tuple is `(OSError, UnicodeError, json.JSONDecodeError, ValueError)`, and `RecursionError` is none of those. Through the real CLI with a ~5000-deep nested state file (well under your 64 KiB cap):

```
$ SOUP_PLUGIN_STATE_PATH=.../plugins.json soup plugins
Error: RecursionError: maximum recursion depth exceeded while decoding a JSON array
EXIT=1
```

The vulnerable band is roughly 1k-32k nesting depth; at 50k the size cap catches it. The same call sits in `_collect_active_hooks()`, so this also crashes `build_plugin_callback()` during trainer setup — the "one bad thing must not kill a multi-hour run" policy your own docstring invokes. One word: add `RecursionError` to the tuple.

**2. MEDIUM, a regression this PR introduces: the new `resources` column is unescaped.** `commands/plugins.py:96-100` passes it to `table.add_row` raw while every sibling cell uses `escape()` (`:102` name, `:107` description). Template and model-group names are validated only for non-empty / no-NUL / <=128 chars, so arbitrary characters are allowed. Reproduced:

```
templates=["[blink]TPL-MARKUP[/]"], description="[bold red]DESC-MARKUP[/]"

literal '[blink]'    in output: False   <- INTERPRETED, output carries \x1b[5m
literal '[bold red]' in output: True    <- the existing column escaped it correctly
```

One line: `escape(", ".join(...))`. (Raw ESC bytes ride through `escape()` too — the house answer is `utils/terminal.py:for_terminal`, which six command modules already use. That part is pre-existing for `description`; I am only asking you not to add a second unescaped surface.)

---

**3. The design question, and my decision: make "disabled" a real boundary.**

Every entry point is `entry_point.load()`ed — imported and executed — *before* the disabled default is applied. Proved with a real `.dist-info` whose module writes a file at import:

```
soup plugins  -> rc=0  third-party code executed: True   (row shows state "disabled")
soup version  -> rc=0  third-party code executed: False
```

#819 item 2 says *"loading third-party code into a training run is an opt-in decision."* As written, **hook participation** is opt-in but **code loading** is not: a merely-installed, never-enabled plugin executes on every `soup train` and every `soup plugins`, and one that imports torch silently makes both heavy — which would undo the lightness property I measured above.

`EntryPoint.name` is available from metadata **without importing**, so discovery can list unknown and disabled names from metadata alone and `load()` only what the state file marks enabled. That makes the word "disabled" mean what a user reads it to mean, and keeps `soup plugins` provably light regardless of what is installed. Please do that rather than sharpening the docs sentence — the docs are honest today, but "installed plugins run code on every training run" is not a property I want to document, it is one I want to not have.

---

**Two smaller ones:**

- **`changelog.d/0.75.0/935.fixed.md`**: credit should be `(#819 by @Amix29 in #935)` — as written the issue number is lost, and the house form is `(#878 by @Srinivasan8888 in #882)`. The fragment is also 157 bytes against 563-1042 for every sibling, and it omits the user-visible break: **`soup plugins install` moves from exit 0 to exit 2**, which anyone scripting it will hit.
- `tests/test_issue819_plugin_discovery.py:162-163` is width-dependent. The `resources` column is `overflow="fold"`, so it folds mid-word: `'my-template'` is found at COLUMNS 80 (the CliRunner default, hence green CI) and **not** at 60 or 40, where it renders as `my-temp` / `late`. An ANSI strip alone will not fix that — pin an explicit console width. Five more raw `in result.output` assertions at `:66, :67, :134, :135, :145` should route through the canonical `_plain()` helper in `tests/test_v07302.py:57-66`.

Docs otherwise check out against code I executed. Two gaps worth a line each: bundled modules default to *enabled* while entry points default to *disabled* (`plugins/__init__.py:420` vs `:445`), which is a real asymmetry a reader would want stated, and `SOUP_PLUGIN_STATE_PATH` is undocumented. That env var also has no realpath/commonpath containment (`~/evil/anywhere.json` resolves and is written) where `SOUP_SPECTRUM_CACHE_DIR` and `SOUP_LAYER_STREAM_CACHE_DIR` both have it — self-set, so low risk, but the convention exists.

Amix29:
Addressed every review item in 17d3945:

- third-party entry points are now discovered from metadata only and remain unimported while disabled; explicit `soup plugins enable <name>` loads only the selected entry point
- persisted enabled entry points still load for training, and entry-point/registered names must match
- `_read_enabled_state` now tolerates recursive JSON, and the resources column escapes Rich markup
- tests pin console width, normalize ANSI output, prove the no-import boundary across fresh processes, exercise malicious resource markup, recursive state, and partial registrar failure
- docs now distinguish bundled defaults from third-party opt-in, document `SOUP_PLUGIN_STATE_PATH`, and record the `plugins install` exit-code 2 behavior
- changelog now credits `#819 by @Amix29 in #935` and calls out the breaking exit-code correction

Post-rebase validation on origin/main 876b2b1:
- `10 passed` for `test_issue819_plugin_discovery.py` under `FORCE_COLOR=1 COLUMNS=40`
- `298 passed, 1 skipped` for the v0.45/v0.53/CLI compatibility set
- Ruff clean and `git diff --check` clean.

MakazhanAlpamys:
Re-checked at `17d3945`. **The design item is genuinely delivered**, and I verified it with a real installed distribution rather than a stub — a `fakeplug-1.2.3.dist-info` with an entry point in the group, whose module writes a marker file at import. Each row a fresh process, marker deleted first:

| command | state | third-party module imported | listed? |
|---|---|---|---|
| `soup plugins` | installed, **not** enabled | **NO** | yes, as `disabled` |
| `soup version` / `soup --help` | not enabled | **NO** | — |
| `soup plugins enable fakeplug` | — | YES | — |
| `soup plugins` | **enabled** | YES | yes, hooks `pre_train` |
| `soup version` / `soup --help` | **enabled** | **NO** | — |

The last two rows are stronger than I asked for: even an *enabled* plugin stays out of `version` and `--help`, so lightness does not degrade as users opt in. And the table still shows the real metadata version without importing anything:

```
│ fakeplug │ 1.2.3 │ disabled │ none │ none │ Installed entry point; enable it to load third-party code. │
```

The boundary is pinned by a mutation that reproduces the original defect: `if saved.get(entry_point.name, False):` -> `if True:` (anchor 1, real import) fails 3 tests, and in the live harness `soup plugins` with nothing enabled goes back to `IMPORTED=YES`.

Lightness re-measured with the fake plugin **installed and enabled** (the hostile case): `torch` and `transformers` absent from `sys.modules` after `--help`, `version` and `plugins`; `test_cli_startup_is_light.py` 34 passed.

The `resources` escape is closed — `[blink]` now appears literally, and dropping the `escape(` fails `test_plugin_resources_are_rendered_as_literal_text`. Width dependence is fixed at both layers (`Console(width=120)` in-process, `COLUMNS=120` in the subprocess test, `_plain()` on every output assertion): 10 passed at 80, at `COLUMNS=60`, and at `COLUMNS=40`. The fragment now credits `(#819 by @Amix29 in #935)` and states the break.

---

**One thing to send back, and it is one line.**

The `RecursionError` fix itself is real — the except tuple at `plugins/__init__.py:140` now includes it, and with a 5,000-deep state file both `soup plugins` and `build_plugin_callback()` exit 0 and degrade to `{}`. **But the test that is supposed to hold it is vacuous: removing `RecursionError` from the tuple gives 10 passed, 0 failed.**

`tests/test_issue819_plugin_discovery.py:269`:

```python
nested = "[" * 1_500 + "0" + "]" * 1_500
```

Depth 1,500 does **not** raise `RecursionError` on CPython 3.12 — the C `json` scanner absorbs it. Measured on this box at `recursionlimit` 1000, with the fix removed:

```
depth 1500 (what the test uses):  returned {}   <- no crash, nothing to catch
depth 3000:                       RAISED RecursionError
depth 5000:                       RAISED RecursionError
```

So the test reaches `{}` through the unrelated `"unsupported schema"` `ValueError` branch — `enabled` is a list rather than a dict — and would pass identically on pre-fix code. **`1_500` -> `5_000`** is the whole fix; that is still about 10 KB, well under your 64 KiB cap.

This is the same shape as the `BUNDLED_SCORER_FINGERPRINT` lesson in v0.73.4: a test pinned at a value where the behaviour it names cannot move. Worth catching precisely because everything around it is right — the fix works, the manual reproduction works, and only the guard is asleep.

Not a finding, just so it is written down: raw ESC bytes still ride through `escape()`, so `resources` inherits the same pre-existing gap as `description`. I asked only for no *second* unescaped surface, and that is delivered; `for_terminal` for both is separate work.

Push the depth change and I will merge once CI is green.

Amix29:
Closed the final vacuous-test gap at `75e8c4a`, rebased onto current `origin/main@fcca8ef`.

The recursive JSON fixture now uses depth 5,000 (about 10 KiB, below the 64 KiB cap). On this CPython 3.12 runtime, direct `json.loads` of that exact value raises `RecursionError`; through `_read_enabled_state` the regression test passes, so removing `RecursionError` from the handler can no longer stay green.

Post-rebase validation:
- `10 passed` for `test_issue819_plugin_discovery.py` under `FORCE_COLOR=1 COLUMNS=40`
- `298 passed, 1 skipped` for the v0.45/v0.53/CLI compatibility set
- Ruff clean and `git diff --check` clean.

MakazhanAlpamys:
Re-checked at `75e8c4a3`. **Blocker closed, and I measured the fixture rather than trusting the number.**

```
depth 1500 (the old value):  returned OK, NO RecursionError   (3,025 bytes)
depth 3000:                  RAISED RecursionError            (6,025 bytes)
depth 5000 (the new value):  RAISED RecursionError           (10,025 bytes)
```

Mutation is now decisive — dropping `RecursionError` from the except tuple (anchor 1 occurrence at `plugins/__init__.py:140`, module really imported, `inspect.getsource` confirming the handler no longer names it):

```
>       assert plugins_pkg._read_enabled_state() == {}
E       RecursionError: maximum recursion depth exceeded while decoding a JSON array
FAILED tests/test_issue819_plugin_discovery.py::test_recursive_state_json_is_ignored_without_crashing
1 failed, 9 passed
```

**I also checked the headroom, because a fixture like this is only non-vacuous while the recursion limit stays low.** Nothing in `src/`, `tests/`, `scripts/`, `conftest.py` or `pyproject.toml` calls `setrecursionlimit`, and depth 5000 still raises at limits 1000 / 1500 / 2000 / 3000 — three times the default before it could go quiet again. Worth knowing that this is the property holding the test up, in case anyone ever raises the limit for an unrelated reason.

No regression from the rebase. I rebuilt the real `fakeplug-1.2.3.dist-info` with an isolated `USERPROFILE` so the state file starts absent, marker deleted before each row, fresh process each time — an installed-but-not-enabled plugin's module is still not imported by `soup plugins`, an enabled one is, and neither `version` nor `--help` imports it in either state. `torch` and `transformers` absent from `sys.modules` throughout; `test_cli_startup_is_light.py` 34 passed. The `resources` escaping I re-checked **live** under `FORCE_COLOR=1` rather than by unit test: `[blink]TPL-MARKUP[/]` and `[bold red]GROUP-MARKUP[/]` both render literally.

265 passed, 2 skipped across the four suites; ruff clean; no new findings.

CI is queued on `75e8c4a3`. I will merge when it lands.

MakazhanAlpamys:
Merged as `26b6c309`. Thank you — and the part that made this easy to merge is that you treated the laziness as a **security property** rather than a startup-time optimisation.

Eager entry-point discovery meant that merely having a plugin installed imported its module, before anyone had enabled it, on paths as innocuous as `soup --help`. The fix is only worth anything if that import genuinely stops happening, and you proved it rather than asserting it: a real `fakeplug-1.2.3.dist-info`, an isolated `USERPROFILE` so the state file starts absent, the marker deleted before each row, a fresh process per case. Installed-but-not-enabled is never imported; enabled is; neither `version` nor `--help` imports it in either state; `torch` and `transformers` stay out of `sys.modules` throughout.

**And you checked the recursive-state-file fixture for headroom**, which almost nobody does. Depth 5000 still raises at recursion limits of 1000 / 1500 / 2000 / 3000 — three times the default before it could go quiet — and nothing in `src/`, `tests/`, `scripts/`, `conftest.py` or `pyproject.toml` calls `setrecursionlimit`. That is the difference between a test that works and a test known to keep working, and it is worth remembering if anyone ever raises the limit for an unrelated reason.

I re-checked the `resources` escaping live under `FORCE_COLOR=1` rather than by unit test — `[blink]TPL-MARKUP[/]` and `[bold red]GROUP-MARKUP[/]` both render literally.

265 passed and 2 skipped across the four suites, `test_cli_startup_is_light.py` 34 passed, ruff clean, `merge-tree` clean against current `main`.

Your `CONTRIBUTORS.md` entry is updated at `b0a63382`.

Issue #876, opened 2026-09-12
soup data preprocess cache and the live path disagree on the leading BOS for chat_template presets (the BOS half of #791)


MakazhanAlpamys:
Sibling of #791, which covers the **EOS** half of the same split. This one is about **BOS**, and unlike #791 it is not pre-existing: it is introduced by #788 and is the accepted cost of that fix, recorded here so it is tracked rather than deferred to an issue that does not cover it.

## What

For a `data.chat_template` preset on a tokenizer whose post-processor adds a BOS, the live training path and the `soup data preprocess` cache now disagree on the leading BOS count.

No Soup preset renders a BOS — `grep -n "bos_token" src/soup_cli/data/chat_templates.py` returns nothing — so the BOS, when present, comes from the tokenizer's post-processor alone. Measured on real fast tokenizers with `train_on_responses_only: false`, leading-BOS count per row:

| template | post-processor | live path | preprocess cache | on `main` (cache) | agree? |
|---|---|---|---|---|---|
| `{{ bos_token }}` | bos | 1 | 1 | 2 | yes |
| `{{ bos_token }}` | bos+eos | 1 | 1 | 2 | yes |
| `{{ bos_token }}` | none | 1 | 1 | 1 | yes |
| **preset (no BOS)** | **bos** | **0** | **1** | 1 | **no** |
| **preset (no BOS)** | **bos+eos** | **0** | **1** | 1 | **no** |
| preset (no BOS) | none | 0 | 0 | 0 | yes |

This is the real `data.chat_template: llama3` on a Llama tokenizer case from #781's own table, not a synthetic one.

## Why it is not a regression, and still worth an issue

On `main` both paths said 1, so they agreed — on a value #781 established is wrong for a rendered template. #788 correctly moves the live path to 0, matching what the model is asked to generate, and deliberately leaves the cache path tokenizing as `main` did so the cache stays byte-identical apart from the removed duplicate. That was the right call in #788: the alternative, `add_special_tokens=is_pretrain`, was tried in an earlier revision of that PR and dropped the trained EOS on any truncated row.

So nobody is worse off than on `main`. But the two paths now differ, and `strip_doubled_leading_bos` cannot close it by construction — it collapses a *doubled* BOS and there is only one here.

[...]

PR #938, opened 2026-09-13
fix(data): apply TRL's training EOS rule on the soup data preprocess chat cache (#791)

### LINKED ISSUE #791: soup data preprocess caches chat rows without the EOS TRL adds on the live path, so cached runs can train with no stop token
Found while reviewing #788 (@abdulwaarith0's fix for #785). Pre-existing on `main`; filed separately so the decision about it is not made inside a BOS fix.

## What

A chat dataset tokenized by `soup data preprocess` and then trained from the cache can carry a **different EOS count per row** than the same dataset trained without the cache.

- **Live path (`train_on_responses_only: false`, no cache):** rows reach TRL as `{"text": ...}`, and TRL 0.29.1 appends `eos_token` **as a string, before tokenizing**, to every text row that does not already end with it (`trl/trainer/sft_trainer.py:1026-1038`).
- **Cache path:** `soup data preprocess` (`src/soup_cli/commands/data.py`, the tokenize loop in `preprocess_dataset`) renders the template with `apply_chat_template(..., tokenize=False)` and calls `tokenizer(text, ...)` directly. There is no string-level EOS step, so the row gets an EOS only if the template renders one or the tokenizer's post-processor appends one.
- Training does consume that cache: `trainer/sft.py:433-480` validates it against `make_preprocess_cache_key` and loads it with `load_from_disk`.

So for a template that does not render EOS, on a tokenizer whose post-processor does not append one (Qwen-shaped), the live path trains on one EOS per row and the cached path on **zero** — the model sees no stop token in training.

This comes from reading both code paths on `main`. The per-row counts were measured on real fast tokenizers during the #788 review, through TRL's own `_prepare_dataset` for the live side.

## Why it matters

A missing trained EOS teaches run-on generation, and it is silent: the loss curve looks normal, and `soup data doctor --show-mask` renders the live path, not the cache.

## Proposed fix path

Make the preprocess tokenizer apply the same rule TRL applies to text rows: append `eos_token` when the rendered text does not already end with it, and only for text a chat template rendered. Pretrain rows keep their current behaviour. Because this changes what cached data trains on, it must bump `make_preprocess_cache_key`, so an existing cache is rejected rather than silently reused.

Sequencing: #788 is changing the same code for the BOS half. This should land after it, or be folded in only if the maintainer decides so in #788's thread.

## Acceptance criteria

- [ ] A test with a real fast tokenizer showing that for a template that does not render EOS, the cached path and the live path (through TRL's real `_prepare_dataset`) produce the same trained EOS count per row. It fails on current `main`.
- [ ] Controls: a template that renders EOS gets no second EOS, and a pretrain row stays byte-identical to `main`.
- [ ] `make_preprocess_cache_key` changes, with a test.
- [ ] A changelog fragment stating that models trained from a `soup data preprocess` cache before this change may have trained without a stop token on such templates.

abdulwaarith0:
Fixes #791.

`soup data preprocess` tokenised the rendered chat template directly, so a cached row carried an EOS only when the template or the tokenizer's post-processor supplied one. The live SFT path (`train_on_responses_only: false`) reproduces TRL 0.29.1's `add_eos` rule via `build_full_sequence_labels` then `append_training_eos`: a chat row that does not already end on `eos_token` gets one. So on a template that renders no EOS and a Qwen-shaped tokenizer whose post-processor appends none, the live path trained on one stop token and the cached path on zero. It was silent: the loss curve looks normal and `soup data doctor --show-mask` renders the live path, not the cache.

## Fix

The preprocess chat path now applies the same `add_eos` rule in token space, after `strip_doubled_leading_bos`, appending the EOS id when the row does not already end on it and only while the row is under the length budget (a filled row was truncated, and the live path, which appends then truncates, keeps no trailing EOS there either). Pretrain rows never go through a chat template and are unchanged. The leading BOS still differs between the two paths by design and is tracked in #876. The cache-key schema bumps `v2 -> v3`, so a dataset cached under the old encoding is re-tokenised rather than silently reused.

## Tests

Real fast tokenizers, no mocks:

- no-EOS template, Qwen shape: cached EOS count per row equals the live path's; fails on pre-#791 `main` (cache 0 vs live 1)
- BOS-only post-processor also gains the EOS; the one BOS is untouched
- controls: an EOS-rendering template gets no second EOS; a pretrain row is byte-identical to `main`; a truncated row is not pushed past `max_length`
- the cache key advanced off both the v2 and the pre-schema blob

Mutation-tested: removing the append (3 tests, equals `main`), appending unconditionally (truncation control), appending to pretrain (pretrain control), and reverting the schema (cache-key test) each go red.

On the live-path baseline: post-#788 the `train_on_responses_only: false` path pre-tokenises, so `build_full_sequence_labels` is what training consumes; the real `SFTTrainer._prepare_dataset` hangs on Windows (PartialState), as on #788, so the baseline is the builder plus a faithful TRL `add_eos` reproduction. Happy to add a real `_prepare_dataset` test behind a non-Windows marker if you would prefer.


MakazhanAlpamys:
Reviewed at `29ae9f0c`. This is the right fix and the evidence holds up under checking.

**I verified the TRL rule against the installed source rather than your description.** `trl 0.29.1`, `trl/trainer/sft_trainer.py:1026-1031`:

```python
def add_eos(example, eos_token):
    if "text" in example and not example["text"].endswith(eos_token):
        example["text"] = example["text"] + eos_token
```

`_trl_main_text_eos_count` reproduces that exactly, and the line range you cite is correct.

**The budget guard is right, and for the reason you give.** `if len(input_ids) < max_length` — at exactly `max_length` the live path appends and then truncates, keeping no trailing EOS either, so skipping the append there is the *matching* behaviour rather than an omission. That is the kind of thing that looks like a missing case until you check what the other side does.

**Five mutations, all killed:** removing the append entirely (3 tests, back to pre-fix `main`); appending unconditionally, ignoring already-ends-on-EOS (4); dropping the `max_length` guard (the truncation control); reverting `_PREPROCESS_TOKENIZE_SCHEMA` to `v2` (the cache-key test); and removing the already-ends-on-EOS guard inside `append_training_eos` (4). The first confirms the tests genuinely fail on pre-fix behaviour; the second and fifth confirm the controls are load-bearing in both directions.

**Nothing was weakened, and one test was inverted for the right reason.** `test_cache_keeps_mains_zero_eos_for_no_eos_post_processor` — which pinned the divergence as *intentional* while #791 was open — becomes `test_cache_matches_live_eos_for_no_eos_post_processor` and asserts strictly more: main's 0, live's 1, cache == live == 1, the EOS in trailing position, and the BOS still de-duplicated. Inverting a test you previously wrote to pin the opposite is the correct move when the deferral closes, and doing it explicitly rather than deleting it is what makes the history readable.

23 passed, green under `FORCE_COLOR=1`, ruff clean. The new tests drive the real CLI and read the on-disk cache back through `datasets.load_from_disk`, so this is end-to-end rather than a unit stub.

**On the one open item — #791's first acceptance criterion.** It asks for the live baseline **through TRL's real `_prepare_dataset`**, and you use `build_full_sequence_labels` plus a faithful reproduction instead, because `_prepare_dataset` hangs on Windows via `PartialState` — the same obstacle #788 hit. I checked the reproduction is line-accurate, so the substance is there. **Please add the real `_prepare_dataset` test behind a non-Windows marker as you offered.** It removes the ambiguity permanently, and the ubuntu and macOS cells will run it — which is the point, since those are the platforms where the hang does not occur.

**The changelog fragment is the best part of this**, and I want to say so specifically: it names the user-visible consequence — a model trained from an old cache may have had no stop token — and gives the remedy, re-run `soup data preprocess` and retrain. That is exactly what criterion 4 was for, and most fragments describe the code change instead.

No hunk overlap with the other open PRs on `commands/data.py`: yours is at ~2501-2546 in `preprocess_dataset`, #921 at 16-135, #922 at 2786-2903, #925 at 2873. They merge cleanly in any order.

This is your fourth piece of the doubled-token family (#788, #867, #892 and now this), and the pattern across all four is the same: fix the reachable path, leave the adjacent divergence explicitly documented rather than quietly, and file the follow-up. #876 is still correctly open for the leading-BOS half.

MakazhanAlpamys:
@abdulwaarith0 — **your CI is red and it is not your change.** I chased it rather than asking you to, because the diagnosis turns out to be a clock, not a defect.

All nine test cells fail on one line, and it is in neither your diff nor your branch:

```
ChangelogError: changelog.d/0.75.0/768.fixed.md: #768 already has fragment
  changelog.d/0.75.0/768.changed.md; use one fragment per issue or pull request
= 1 failed, 21365 passed, 127 skipped =
```

Note the other number: **21,365 passed**. Everything of yours is green.

The duplicate belonged to `main`, not to you — I scanned every open PR branch and none of them carries `768.fixed.md`, including yours. The timing is the whole story:

```
run 34754333732 created   2026-09-13T11:23:56Z
e27e40c0 landed on main   2026-09-13T11:26:29Z   ("one fragment per PR — fold the redactor note into #768's")
```

**Three minutes.** Your run computed its merge commit against a `main` that still had both files, and the fix arrived while it was queueing.

I have re-run the failed jobs. If the re-run reuses the original merge commit it will fail identically, in which case the fix is a plain rebase onto current `main` — nothing to change in your code either way. The Actions queue is at ~87 runs, so give it time before reading anything into it.

Separately, my review of `29ae9f0c` stands: the TRL rule checks out against the installed source rather than against the description, and I verified `trl/trainer/sft_trainer.py:1026-1031` directly. I will pick the substantive thread back up once there is a green matrix to read.

abdulwaarith0:
Rebased onto current main — the red was the `768` duplicate fragment that landed there, not anything in the diff, and it's gone now.

Added the real `_prepare_dataset` test (`b9f1e9e`) as you asked. It drives TRL 0.29.1's genuine `SFTTrainer._prepare_dataset` over the rendered `{"text"}` rows and asserts the cache trains on the same EOS count TRL actually produces: real, my reproduction, the live `build_full_sequence_labels` path, and the cache all agree at one EOS per row. I call `_prepare_dataset` unbound with a minimal `self` (the language-modeling text path reads only `self._is_vlm`), so it runs TRL's real map/`add_eos`/tokenize pipeline without the trainer construction that hangs on Windows — marked skip on win32, so the ubuntu and macOS cells are what pin the baseline.

MakazhanAlpamys:
Verified `b9f1e9e4`. **Merge-ready** — you added exactly the test I asked for, and the rebase diagnosis was right.

**On the red CI**: correct, and I had reached the same conclusion independently before your reply. The run was created at `11:23:56Z` and `e27e40c0` — the commit that removed the duplicate `768` fragment — landed on `main` at `11:26:29Z`. **Three minutes.** Your matrix computed its merge commit against a `main` that still had both files. 21,365 of your tests passed; the single failure was never yours.

**The `_prepare_dataset` test is there and correctly gated.** `tests/test_issue791_preprocess_cache_eos.py:295`, skipped on `win32` with a reason that explains the mechanism rather than just naming the platform:

```
SKIPPED [1] TRL's SFTTrainer._prepare_dataset runs its dataset prep inside
PartialState().main_process_first(); the full training path this stands in for hangs on
Windows via PartialState (the same obstacle #788 hit), so this runs on the ubuntu and
macOS cells, which is where the baseline needs pinning.
```

**And I want to be straight about what I could and could not check.** This box is Windows, so that test **skips here** — 10 passed, 1 skipped. I have verified it exists, that the skip condition is the narrow one, and that its docstring's claim about `_prepare_dataset` reading only `self._is_vlm` on the text path is accurate. What I have **not** done is watch it pass. The ubuntu and macOS cells are the only signal for it, which is precisely why putting it behind a platform marker rather than dropping it was the right call — six of the nine cells will run it.

That closes #791's first acceptance criterion properly. The reproduction was line-accurate and I said so, but a reproduction that agrees with the original today can drift silently; the real call cannot.

Everything from the earlier review stands: five mutations all killed, including the two that prove the controls are load-bearing in both directions; nothing weakened, with `test_cache_keeps_mains_zero_eos_for_no_eos_post_processor` **inverted rather than deleted** when the deferral closed, which is what keeps the history readable; and the fragment naming the user-visible consequence — a model trained from an old cache may have had no stop token — with the remedy, instead of describing the code change.

No hunk overlap with the other open PRs on `commands/data.py`. Merging as soon as the matrix is green; the queue is over 100, so it will be slow.

Fourth piece of the doubled-token family after #788, #867 and #892, and #876 is still correctly open for the leading-BOS half.

MakazhanAlpamys:
Merged as `58601145`. Thank you — this was a silent-corruption bug and you closed it properly.

**What I ran before merging.** Your head `b9f1e9e4` was unchanged from the verdict I recorded yesterday, so I did not re-review it. What I did do is the check that branch-green cannot answer: `main` had moved (#989 landed at 11:57Z), so I composed your head onto `main` at `614b7f64` together with #982, #983 and #984 and ran them as one tree — `test_issue785_training_bos`, `test_issue791_preprocess_cache_eos`, plus the repo-wide guards `test_cli_startup_is_light`, `test_issue748_config_fields_reach_a_consumer` and `test_issue775_path_containment_ratchet`: **612 passed, 2 skipped, zero failures.** All four merged clean and pass together.

**What was best about this one.** Two things.

The first is that you took the harder path on the test when I asked. Your original baseline was the builder plus a faithful reproduction of TRL's `add_eos`, with a reasonable justification — the real `SFTTrainer._prepare_dataset` hangs on Windows through `PartialState`. A reproduction of a rule is evidence about your reading of the rule, not about the rule. You drove TRL 0.29.1's genuine `_prepare_dataset` instead, unbound with a minimal `self`, and made four things agree at one EOS per row: real TRL, your reproduction, the live `build_full_sequence_labels` path, and the cache. That is what makes this fix checkable by someone who does not already believe it.

The second is the truncation control. The obvious fix appends the EOS unconditionally; yours appends only while the row is under the length budget, because the live path appends *then* truncates and keeps no trailing EOS there either. That is not a detail you can arrive at by reading the bug report — it comes from following the live path to the end, and your mutation `appending unconditionally` proves the control is load-bearing rather than decorative.

**On the red CI you diagnosed.** You were right, and I had reached the same conclusion independently before your reply, so I will state the timing for the record: that run was created at `11:23:56Z`, and `e27e40c0` — the commit removing the duplicate `768` fragment — landed on `main` at `11:26:29Z`. Three minutes. Your matrix computed its merge commit against a `main` that still had both files. 21,365 of your tests passed; the single failure was never yours.

The changelog fragment names the user-facing consequence and the remedy — a model trained from a pre-change cache on a template that renders no EOS learned run-on generation, and re-running `soup data preprocess` is what picks the stop token back up. Someone reading the release notes in six months can act on that, which is the whole point of the file.

The leading-BOS half stays open as #876, as you noted.

PR #940, opened 2026-09-13
fix(data): remove length bias from preference examples

### LINKED ISSUE #929: Both shipped preference examples fail soup data lint on length_bias (6x mean length gap)
## What

Both shipped preference examples fail `soup data lint` on `length_bias` with a MAJOR verdict, so `soup data lint examples/data/dpo_sample.jsonl` exits 2 out of the box.

Measured on `main` at `c9fa52ca`, after #916 merged:

```
soup data lint examples/data/dpo_sample.jsonl
length_bias   MAJOR   mean chosen=66.4, mean rejected=10.4   (Cohen's d = 7.309)
prompt_leak   OK      0/8 rows (0.0%)
MAJOR - 8/8 rows scanned
```

`examples/data/chat_preferences.jsonl` is the same shape: means 36.6 vs 9.6, `d = 6.355`.

## Why this is not #810

#810 was a false positive — the linter flattened conversational turns and found the prompt inside its own completion. That is fixed in #916, and `prompt_leak` now reports OK on both files.

**This one is a true positive.** A 6x mean length gap between `chosen` and `rejected` is exactly what `length_bias` exists to detect: a preference model trained on it learns "longer is better" rather than the intended distinction. The check is right and the data is bad.

@dchaudhari7177 found this while fixing #810 and deliberately **did not** edit the fixtures to make the gate green, which was the correct call — a fix that makes a real finding disappear is worse than no fix.

## Why it matters

These two files are what a new user meets. `examples/README.md` points at them, and someone following the docs runs a lint that exits 2 on data we ship. Either the data should not be that skewed, or we should say in `examples/README.md` that it deliberately is and why.

## Acceptance criteria

- [ ] Decide which: rewrite the `rejected` completions so the length gap is not the dominant signal, **or** document in `examples/README.md` that these fixtures are deliberately length-skewed and are not a model for real preference data.
- [ ] If rewriting: `soup data lint examples/data/dpo_sample.jsonl` and `examples/data/chat_preferences.jsonl` both exit 0, and the *content* distinction between `chosen` and `rejected` is still obvious to a human reader — a rewrite that makes the lengths match by padding `rejected` would pass the gate and teach nothing.
- [ ] A test asserting both shipped preference examples lint clean, so this cannot drift back. Put it beside the existing example-validation tests rather than in a new file.

The rewrite is the more useful outcome but it is a judgement call about content, not a mechanical edit — read both files first and say in the PR what distinction you intend `chosen` to carry. If you conclude the documentation route is better, that is a legitimate answer and I will take it; say why.

Not a good first issue: it needs a view on what good preference data looks like, and the acceptance test has to distinguish a real fix from a padded one.

taljeon:
## What does this PR do?

Closes #929.

Rewrites the rejected completions in both shipped preference-data examples so they are similarly detailed but still clearly worse than the chosen answers. Each rejected answer now makes specific factual or methodological errors instead of being a terse straw man; no filler was added just to satisfy the length check.

It also adds a parametrized regression test beside the existing example-validation tests. The test loads both JSONL files through `run_lint(..., "auto")` and requires an overall `OK` verdict.

Measured after the rewrite:

- `dpo_sample.jsonl`: mean chosen 66.4 words, mean rejected 66.9 words; `length_bias` is `OK` (`d = -0.052`, 0.7% mean gap).
- `chat_preferences.jsonl`: mean chosen 36.6 words, mean rejected 36.2 words; `length_bias` is `OK` (`d = 0.133`, 1.1% mean gap).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

- [ ] `pytest tests/ -v` passes (the local environment lacks the optional `transformers` dependency; targeted results are below)
- [x] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed (not needed; this changes the shipped example data itself)

## Validation

- `soup data lint examples/data/dpo_sample.jsonl` — `OK`, 8/8 rows scanned
- `soup data lint examples/data/chat_preferences.jsonl` — `OK`, 5/5 rows scanned
- Targeted example and data-lint regression suite — 85 passed, 3 skipped
- Full `tests/test_v07127.py` attempt — 197 passed, 3 skipped, with 2 failures because the optional `transformers` dependency is not installed
- Exact head `00b4a9f7` completed all 14 GitHub CI jobs green.
- Current `main@40a97fa7` does not touch this PR's four files; GitHub reports the head MERGEABLE/CLEAN.
- `git diff --check`

## AI assistance

I used an AI coding assistant to help draft the example rewrites and regression test. I reviewed every changed answer for an obvious chosen/rejected quality distinction and ran the validation above locally.


MakazhanAlpamys:
Reviewed at `023b6cec`. This is the right answer to #929, and it avoids the trap the issue named — which was the thing I most expected to have to send back.

**The check that mattered, run programmatically against the merge-base rather than by eye:**

```
examples/data/dpo_sample.jsonl:       rows=8  chosen unchanged=8/8  rejected changed=8/8
examples/data/chat_preferences.jsonl: rows=5  chosen unchanged=5/5  rejected changed=5/5
```

**`chosen` is byte-identical in all 13 rows.** Only `rejected` was rewritten, and the rewrite replaces terse straw men with detailed, confidently *wrong* answers — photosynthesis "through their roots"; a hash table that "searches from the first entry" and is "still O(1) because modern processors scan arrays instantly"; a neural network that "copies the closest answer". The quality contrast is obvious to a human in every row and no filler was added; per-row length ratios sit between 0.95 and 1.09.

That distinction is the whole content of #929. A rewrite that matched the lengths by padding `rejected` would have passed the gate and taught a preference model nothing, and I said so in the issue precisely because it is the easy way to close it. You did the harder thing.

**The numbers reproduce.** On the old files: MAJOR, exit 2, mean chosen 66.4 and 36.6 — matching the issue exactly. On yours: exit 0, `d = -0.052` at a 0.7% gap and `d = 0.133` at 1.1%, matching your description. All four lint checks are OK on both files.

**The test is in the right place and is sensitive**: reverting a *single* row of eight is enough to take it red, so it is not merely checking an aggregate. Reverting either file whole kills it too. Putting it in the existing `tests/test_examples_configs.py` beside the config-validation tests, rather than in a new file, is what #929 asked for.

**One thing to fix before merge.** The changelog fragment ends `(#940).` — please make it `(#929 by @taljeon in #940).` Thirteen of the fourteen fragments in `changelog.d/0.75.0/` carry the author credit, and the one that does not is my own. Attribution is the thing that has to be right at merge time, not at release time.

**Two notes, neither blocking:**

1. I mutated the files into the padded case the issue warns about — old terse `rejected` plus filler words to match length — and the test **passes**. That is inherent to a length-only gate rather than anything you did, and #929 listed it under notes rather than acceptance criteria for that reason. A line in the test docstring saying the gate checks lint verdicts only, and that the content distinction was established in review, would stop a future contributor reading green as proof of quality.
2. `examples/data/` is browsable on GitHub out of context, and some of the new `rejected` answers are confident claims about vaccines and about a lock icon meaning a site is safe for private information. In context they are obviously the negative half of a preference pair. One sentence in `examples/README.md` noting that `rejected` fields are deliberately incorrect would remove the ambiguity for anyone who lands there from a search result. Your call — I am not asking for it.

Thank you for saying in the PR body what distinction you intended `chosen` to carry. That was the part #929 said needed judgement rather than typing, and it is why this took one round instead of three.

taljeon:
Updated the changelog fragment in `00b4a9f` to the requested `(#929 by @taljeon in #940)` form.

The 18 example-validation tests still pass (1 skipped), and the fragment validates against current `main`. I kept the optional test-docstring and README notes outside this focused change. The PR will remain draft while refreshed CI runs.

MakazhanAlpamys:
Re-reviewed at `00b4a9f7`. **`023b6cec..00b4a9f7` is exactly one commit and one line — the credit — so the data and tests are unchanged since my last review.** The fragment now reads `(#929 by @taljeon in #940).` verbatim, which was the only blocking item. Please undraft; I will merge when CI lands (14 pending).

**I checked the thing I would have blocked on had it been true: whether the test actually measures the bias or just parses the files.** It measures it. `run_lint` runs all five checks including `length_bias`, which is a real Cohen's d computation gated at `_LENGTH_BIAS_MINOR_REL_DIFF = 0.05` / `_MAJOR = 0.10`, and `LintReport.overall` is the worst verdict across checks — so `assert report.overall == "OK"` is a genuine measurement, not a smoke test.

**Baseline both ways, which is what makes it convincing:**

```
current  dpo_sample.jsonl        -> OK    | length_bias OK    | mean chosen=66.4, rejected=66.9
current  chat_preferences.jsonl  -> OK    | length_bias OK    | mean chosen=36.6, rejected=36.2
pre-PR   dpo_sample.jsonl        -> MAJOR | length_bias MAJOR | mean chosen=66.4, rejected=10.4
pre-PR   chat_preferences.jsonl  -> MAJOR | length_bias MAJOR | mean chosen=36.6, rejected=9.6
```

Reproduces #929's numbers exactly.

**Mutation — I reverted one row's `rejected` only** (the `What is photosynthesis?` row, anchor counted: exactly 1 occurrence; file re-read as JSONL afterwards, 8 rows parse, which is the data equivalent of a parse+import check):

```
FAILED tests/test_examples_configs.py::test_preference_example_lints_clean[dpo_sample.jsonl]
AssertionError: [('length_bias', 'MINOR', 'mean chosen=66.4, mean rejected=60.2'), ...]
assert 'MINOR' == 'OK'
```

Worth naming the sensitivity: **one row in eight** moves the mean from 66.9 to 60.2 — a 9.3% relative gap, below MAJOR but above the 5% MINOR floor — and the test still catches it. That is a tighter guard than I expected.

**The content holds up on inspection, not just on the metric.** `chosen` is byte-identical in all 13 rows across both files; every `rejected` was rewritten. Sampling row 0: *"plants absorb sunlight through their roots and turn it directly into food... oxygen is taken from the air and stored inside glucose... the reaction occurs in the cell wall"* — detailed, confidently wrong, obviously worse than `chosen`. That is the anti-padding property #929 made an acceptance criterion, and padding to the target length would have been the easy way to make the number go green. You did the harder thing.

Zero removed test lines; the diff is `+17` in `tests/test_examples_configs.py` with no existing assertion touched. 274 passed across the lint and example suites. No CLI-output assertions, so the ANSI trap does not apply.

One note, not a gap: the test drives `run_lint` as a library function rather than `soup data lint` through the CLI, so it verifies the verdict engine rather than the exit code the issue quotes. That is the better choice — faster, and it sidesteps the ANSI trap entirely — and the verdict-to-exit-code mapping is already covered in `test_v07127.py`.

The two items you deferred (the docstring note that the gate checks lint verdicts only, and the `examples/README.md` sentence about `rejected` being deliberately incorrect) were both explicitly non-blocking on my side. Keeping the change focused was the right call.

MakazhanAlpamys:
Merged as `4721e082`. Thank you — and the part worth naming is the one you did *not* do.

Padding the `rejected` completions would have made the metric go green in ten minutes and left the examples teaching exactly the wrong thing. You rewrote every one of them to be similarly detailed and still clearly worse — specific factual and methodological errors rather than brevity — which is what #929 made an acceptance criterion precisely because it is the harder path. Sampling row 0: *"plants absorb sunlight through their roots... the reaction occurs in the cell wall"*. Confident, detailed, wrong. That is what a realistic `rejected` looks like.

`chosen` is byte-identical in all 13 rows across both files, so nothing about the preferred answers moved.

What I ran before merging:

- Reproduced #929's numbers exactly.
- **Mutation**, reverting one row's `rejected` only — anchor counted at exactly 1, file re-read as JSONL afterwards so 8 rows parse, which is the data equivalent of an import check — fails `test_preference_example_lints_clean[dpo_sample.jsonl]` on `assert 'MINOR' == 'OK'`.
- The sensitivity is tighter than I expected: **one row in eight** moves the mean from 66.9 to 60.2, a 9.3% relative gap that sits below MAJOR but above the MINOR floor, and the test still catches it.
- 274 passed across the lint and example suites; zero removed test lines; `+17` in `tests/test_examples_configs.py` with no existing assertion touched.
- Full required matrix on `00b4a9f7`: 14/14 green.

One choice I want to endorse rather than let pass unremarked: driving `run_lint` as a library function instead of `soup data lint` through the CLI. It is faster, it sidesteps the ANSI-assertion trap that has turned this project's CI red four times, and the verdict-to-exit-code mapping is already covered in `test_v07127.py`. Testing the engine rather than the wrapper was the right call and it is not the obvious one.

The two items you deferred were non-blocking on my side and keeping the change focused was correct.

Your `CONTRIBUTORS.md` entry is updated at `049b64bc`.

Issue #932, opened 2026-09-13
Distillation: per-microbatch mean weights microbatches, not tokens, so unequal lengths still shift the gradient


MakazhanAlpamys:
## What

#742 (merged `dd151374`) made distillation's accumulated gradient independent of `gradient_accumulation_steps`. It does **not** make it equal to the full-batch gradient when microbatches hold different numbers of tokens.

A per-microbatch mean weights each microbatch equally. The equivalent single full batch weights each **token** equally. With equal-length chunks the two coincide, which is why the merged tests — which use 8 rows split into 4 or 8 equal chunks — measure a clean `1.0` ratio. With unequal lengths they diverge.

## Why this is filed separately, and why it has no test

This residual is named in #720's own text. It **was** covered: `test_unequal_microbatch_lengths_still_match_within_the_mean_contract` existed in an earlier revision of #742.

I asked for that test to be deleted, along with four others, because they built a bare `torch.nn.Linear(4, 1)` and took a `compensate: bool` the test supplied itself — five of the six passed on unmodified `main` and would have passed with `src/soup_cli/` deleted. That judgement was right for the other four. For this one it was right about the *test* and wrong about the *consequence*: it was the only place in the repo stating the residual, and deleting it left the fact recorded nowhere.

So this issue exists because of a call I made in review, not because the contributor missed something. @dchaudhari7177's merged work is complete for what #720 asked.

## Acceptance criteria

- [ ] A test that measures the gradient with **unequal** microbatch lengths — e.g. 8 rows of lengths `[2, 2, 2, 2, 16, 16, 16, 16]` split into two chunks — and states the expected relationship to the full-batch gradient, whichever way the project decides.
- [ ] A decision, recorded in the code or the docs: either token-weight the loss so the full-batch equality holds for any chunking, **or** document that per-microbatch mean is the contract and unequal lengths shift the weighting. Both are defensible; leaving it unstated is not.
- [ ] If token-weighting is chosen, the equal-length case must be unchanged — the merged `1.0` ratio at GA=4 and GA=8 is the regression test for that.

## Notes for whoever takes this

Do **not** start by writing the fix. Start by writing the measurement: build the unequal-length batch, measure the ratio on current `main`, and put the number in the issue. If it turns out the divergence is negligible at realistic sequence lengths, the honest outcome is a documented contract rather than a code change, and that is a perfectly good result.

[...]

PR #941, opened 2026-09-13
fix(reward): route JSON schema stress references

### LINKED ISSUE #937: soup reward stress --verifiable-domain json_schema cannot run, and its error blames the missing --references the caller already supplied
## What

`soup reward stress verifiable --verifiable-domain json_schema` cannot work at all, and the error it raises tells the user to do something they have already done.

Reproduced on `main` at `cda267b2`, with a reference supplied:

```
run_stress(load_reward_fn("verifiable", verifiable_domain="json_schema"), ['{"a": 1}'])

ValueError: reward target returned 0 score(s) for 1 completion(s) — this verifier
likely needs --references to be probed meaningfully ...
```

## Root cause

`json_schema_reward` (`src/soup_cli/trainer/rewards.py:637-644`) reads its per-row schema from `kwargs.get("schema", [])`:

```python
def json_schema_reward(completions, **kwargs):
    schemas = kwargs.get("schema", [])
    for completion, schema in zip(completions, schemas):
```

`reward_stress._score_batch` passes `answer=`, not `schema=`. So `schemas` is `[]`, the `zip` yields nothing, the function returns `[]`, and the short-return guard fires.

## Why the error message is the second half of the bug

That guard is deliberate and good — it was added so a gold-requiring builtin probed without `--references` fails loudly rather than reporting a false "robust". But here it misattributes: the caller **did** supply references, and the real cause is a keyword mismatch between the reward and its caller. A user following the advice will add `--references` they already have and get the identical error.

So this needs two fixes, and the second matters as much as the first:

1. `json_schema` must receive what it reads, or `reward_stress` must refuse the domain by name up front.
2. The short-return guard must distinguish "no references were given" from "references were given and the verifier still returned nothing". Only the first should mention `--references`.

## Scope

Pre-existing on `main`; found while reviewing #918, which adds structure-preserving attack families to `soup reward stress`. It is **not** #918's to fix and I said so there. `math` and `code` are unaffected — I checked `math` in the same session and it scores normally.

## Acceptance criteria

- [ ] `soup reward stress verifiable --verifiable-domain json_schema --references g.jsonl` either scores, or exits with a message naming the actual problem. Decide which and say why in the PR — refusing a domain the tool cannot probe is a legitimate outcome.
- [ ] A test that fails on current `main` for this exact invocation.
- [ ] The short-return guard distinguishes the two causes, with a test for each direction. The control that matters: a gold-requiring builtin invoked with **no** `--references` must still produce the existing message, so the useful half of the guard is not lost while fixing the misattribution.
- [ ] If `json_schema` is made to work, at least one attack family must be **accepted** by a deliberately gameable schema verifier and **rejected** by a strict one. A domain that is wired up but whose report is always the same is not fixed.

## Notes

Do not close this by deleting the `json_schema` option from `--verifiable-domain`'s help while leaving it accepted. That trades a loud failure for a quiet one, which is the direction this repo keeps having to reverse.

Amix29:
## What does this PR do?

Fixes #937.

`soup reward stress` now routes reference values through the metadata key required by the verifier. The builtin `json_schema` domain receives `schema=`, while existing rewards and custom verifiers continue to receive `answer=`. JSON objects selected from JSONL are decoded back into schema mappings before scoring.

A short score return now distinguishes two cases: with no references it keeps the actionable `--references` guidance; after references were supplied it reports the metadata key and one-score-per-completion contract instead of blaming a missing option.

The regression tests drive the real builtin, the CLI path, both diagnostic directions, and a schema-metadata verifier whose result changes between permissive and strict schemas.

## Current JSON-schema limitation

All four built-in stress attacks currently emit non-JSON text. The builtin verifier therefore rejects them during JSON parsing before schema strictness can affect the result, so its attack report is invariant across strict and permissive schemas until the structure-preserving attacks tracked by #918 land. The synthetic metadata verifier proves that the repaired routing can discriminate; it does not claim that the current builtin attacks do.

The real builtin test now pins `reference_accept == 0.0`. That value is not a meaningful self-acceptance control for this domain because the reference is a schema document, not a completion that should satisfy itself.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Verification

- final head rebased on `origin/main@876b2b1`
- issue file under `FORCE_COLOR=1`: **5 passed**
- adjacent reward-stress, RLVR, and startup tests: **103 passed**
- `ruff check src/soup_cli/ tests/`: **passed**
- `git diff --check origin/main...HEAD`: **passed**
- Earlier full-suite qualification reached **2,580 passed, 2 skipped** before the pre-existing `tests/test_issue204_langfuse_pull.py::test_control_the_leak_scan_sees_a_token_rich_highlighted_and_folded` failure; that exact failure reproduced unchanged in a detached `origin/main` worktree. CI remains the full-matrix authority.

## Checklist

- [x] Ruff passes
- [ ] `pytest tests/ -v` passes — blocked locally only by the independently reproduced baseline test above
- [x] Updated `docs/training.md`
- [x] Added `changelog.d/0.75.0/941.fixed.md`


MakazhanAlpamys:
Reviewed at `8b61df67`. **The plumbing fix is correct and mutation-proven in both directions**, and the baseline is right: your test file fails 4 of 5 on `origin/main` at `e27e40c` with verbatim the issue's reproduction, and the one passer is the control the issue asked to stay green.

**Mutations I ran (each anchor counted, each mutated module actually imported, not just `ast.parse`d — a `NameError` would have produced a false "caught"):**

| mutation | result |
|---|---|
| `reward_stress.py:117` reference routing hardcoded back to `{"answer": ...}` | **4 failed, 1 passed** — `SystemExit(1)`, error names `via 'schema'` |
| `reward_stress.py:119-126` short-return split deleted | **1 failed** — and only the one test, which is the right shape |
| (mine) reworded `--references` out of the legacy no-golds message | **1 failed** — so the issue's "control that matters" is load-bearing, not vacuous |

No tests deleted or weakened; the removed-lines diff is empty. `trainer/rewards.py:645-650` is safe on every malformed input I tried (`'"hello"'`, `'[1,2]'`, `'null'`, `'not json'` all return `[0.0]` via the existing `not isinstance(schema, dict)` guard), and it is a strict improvement on the GRPO path, where a JSON-string schema column previously raised `AttributeError`.

**What blocks it is acceptance criterion 4, and I think you know the shape of it.**

None of the four attack texts is valid JSON, so `json.loads` fails before any schema logic runs and every attack scores 0.0 for every schema. Measured, attack rows byte-identical across three schemas:

```
strict (required answer:int)   attacks=[('empty',0.0),('length',0.0),('repetition',0.0),('sentinel',0.0)] gameable=False
{}                             ... identical ...
{"type":"object"}              ... identical ...
```

Your discriminating test satisfies the criterion with a **synthetic** `metadata_schema_verifier` defined in the test, not with `json_schema_reward`. The issue's words were: *"A domain that is wired up but whose report is always the same is not fixed."* The honest defence is that structure-preserving attacks are #918's scope — the issue says so itself — but neither the PR body, the fragment, nor the docs addition says it, and criterion 1 asked you to decide and say why.

**Second, `reference_accept` is structurally 0% here and the code's own docstring calls that broken.** `run_stress` scores each gold as both completion and reference. For `math` the gold is a valid answer, so the control means something; here the "gold" is a schema document, which does not satisfy itself:

```
strict (required answer:int)   ref_accept=0.0
maximally permissive ({})      ref_accept=1.0
permissive object (no props)   ref_accept=1.0
```

So the panel prints `reference accept: 0%` beside `verdict: robust` for every realistic schema, while `reward_stress.py:161-163` says "a verifier that rejects everything is broken". Your test asserts `gameable is False` and `attacks[0].n == 1` but not `reference_accept` — the one number that exposes this is the one number untested.

**To merge, either is fine and it is your call:**

1. **Document it.** Say in the fragment and in `docs/training.md` that for the builtin domain all four current attack families are non-JSON, so the report is invariant until #918 lands, and that `reference_accept` is not a meaningful control for this domain because the reference is a schema rather than a completion. Add an assertion on `reference_accept` so the constant is pinned rather than incidental.
2. **Refuse the domain by name until #918**, which the issue explicitly blessed.

I would take 1 — the plumbing is genuinely needed by #918 and by custom schema verifiers today, and shipping it with the limitation named is better than making users wait.

**Two one-line items regardless of which you pick:**

- `reward_stress.py:156-166` — the `run_stress` docstring is now false. It still says attacks are scored against the real gold "so numeric/tool_call/**json_schema** verifiers get a valid `answer=`", and calls `reference_accept` "the golds scored as their own correct completions". This PR changes both for `json_schema` and touched neither line.
- `changelog.d/0.75.0/941.fixed.md` reads `(#937 in #941)`. House style in all 15 sibling fragments is `(#<issue> by @<author> in #<PR>)` — e.g. `727.fixed.md`: `(#716 by @dchaudhari7177 in #727)`. Should be `(#937 by @Amix29 in #941)`.

**Non-blocking, worth knowing:**

- `tests/test_issue937_reward_stress_json_schema.py:114` — `assert "robust" in result.output.lower()` reads raw output with no ANSI strip. I reproduced under `FORCE_COLOR=1 TERM=xterm-256color` rather than assuming: escapes *are* emitted, the match still succeeds because `robust` sits inside one contiguous style span and Rich wraps on spaces (checked at COLUMNS 80/72/64/60). So it will not go red today, but it survives only because the token is short and unhighlighted. The repo guard is scoped to `--help` and does not see it. Please route it through a strip helper.
- `commands/reward.py:339-343` vs `reward_stress.py:168-171` are two hand-maintained copies of the same routing rule — the CLI decides by string comparison, `run_stress` independently by `reward_fn.__name__ == "json_schema_reward"`, and because the CLI always passes an explicit `reference_key` the `__name__` branch is unreachable from the CLI. One should derive from the other. Also, a custom `.py` verifier that reads `schema=` has no way to say so today (no `--reference-key`). Fine as a follow-up issue rather than in this PR.
- `reward_stress.py:172-173` — the `if not reference_key: raise ValueError(...)` guard has no test.

Thanks for the measurement-first discipline on this one and on #932 — posting the reproduction to the issue before the implementation makes the review genuinely faster.

Amix29:
Addressed the requested JSON-schema qualification at `b0db5fd`, rebased onto current `origin/main@876b2b1`.

I took option 1 from the review:
- the real builtin test now pins `reference_accept == 0.0` and all four current attacks being rejected
- `docs/training.md` and the changelog explicitly state that the attacks are non-JSON, so parsing rejects them before schema strictness can matter; structure-preserving attacks remain #918
- the `run_stress` docstring now distinguishes answer references from metadata references and no longer calls a schema document a correct completion
- the changelog now uses `#937 by @Amix29 in #941`
- the CLI assertion routes through `strip_ansi`

Validation after rebase:
- #937 file under `FORCE_COLOR=1`: `5 passed`
- adjacent reward-stress/RLVR/startup surface: `103 passed`

MakazhanAlpamys:
Re-checked at `b0db5fd9`. **Option 1 delivered in full, and in all four places I asked for it.**

- `reward_stress.py:156-169` — the false claim is gone. It no longer says attacks are scored against the real gold "so numeric/tool_call/**json_schema** verifiers get a valid `answer=`", and it reframes `reference_accept` as a self-acceptance control that means something only when references are valid completions, noting a schema document need not satisfy itself as an output instance. That is a better sentence than the one I would have written.
- `changelog.d/0.75.0/941.fixed.md` now reads `(#937 by @Amix29 in #941)`.
- Fragment and `docs/training.md:1128-1133` both carry the limitation, and the docs name the cause (parse-before-schema), state the report "does not yet distinguish a strict schema from a permissive one", point at **#918**, and say `reference_accept` is normally 0%.
- `tests/...:35` now pins `report.reference_accept == 0.0`. **I checked the pin is not incidentally true** by mutating it to `1.0`: `assert 0.0 == 1.0`, with the value coming off a real `StressReport`, not a constant.

You also went further than asked: the test dropped `attacks=("empty",)` and now asserts `{attack.kind for attack in report.attacks} == set(reward_stress.ATTACKS)` plus `all(attack.n == 1 and attack.accepted == 0 ...)`, so it pins **all four** families being rejected rather than one. That is the assertion that will notice when #918 changes the picture.

**Both original mutations still die after the rebase** (each anchor counted once, each mutated module actually imported, not just parsed):

| mutation | result |
|---|---|
| `reward_stress.py:116` routing hardcoded back to `{"answer": ...}` | **4 failed, 1 passed** |
| `reward_stress.py:119-125` references-aware short-return split deleted | **1 failed** — only the one test, the right shape |

5 passed in the PR file, 94 across it plus `test_v07141.py`, `test_cli_startup_is_light.py` and `test_issue565_grpo_reward_metadata.py`, and **5 passed under `FORCE_COLOR=1 TERM=xterm-256color`** — the `strip_ansi` import from `tests/conftest.py:16` is real, so the raw-output assertion I flagged last round is now properly covered rather than surviving by luck. Ruff clean.

CI is queued on `b0db5fd9`; I will merge when it lands.

No new findings. The two deferred items stand as follow-ups: the routing rule is still hand-maintained in two places (`commands/reward.py:339-343` and `reward_stress.py:168-171`, with the `__name__` branch unreachable from the CLI), and the empty-`reference_key` guard still has no test. Neither belongs in this PR.

MakazhanAlpamys:
Merged as `23766cb4`. Thank you.

**This is the good kind of bug: the tool's documentation was the bug report, and nobody had read it against the code.** `soup reward stress` stated in its docstring and its terminal notice that every attack is scored against the real gold "so numeric/tool_call/**json_schema** verifiers get a valid `answer=`" — and for `json_schema` the reference went under `answer` where the verifier reads `schema`, so it never arrived at all. The positive control could not distinguish a robust verifier from one that rejects everything, which is precisely the failure mode `reference_accept` exists to catch.

**What lifted this out of "keyword fix" for me was the second half.** You corrected the claim in all four places it was repeated rather than only at the call site, and you reframed `reference_accept` honestly — a self-acceptance control that means something only when the reference is itself a valid completion, which a schema is not. A fix that repaired the routing and left the four copies of the false sentence in place would have been worse than useless, because the next reader would trust them again.

The short-return path raising and naming the metadata key is the right direction too: a verifier handed the wrong key now fails loudly instead of returning a quiet false "robust".

One consequence for the queue: this touches `reward_stress.py`, `commands/reward.py` and `docs/training.md`, so #918 will need a rebase — I have said so there.

PR #943, opened 2026-09-13
fix(distill): weight accumulation by trained tokens

### LINKED ISSUE #932: Distillation: per-microbatch mean weights microbatches, not tokens, so unequal lengths still shift the gradient

Amix29:
## What does this PR do?

Fixes #932.

The real `_DistillTrainer` now consumes Transformers' full accumulation-window `num_items_in_batch` and scales every existing microbatch-mean loss by:

```text
shifted non-masked targets in this microbatch / targets in the accumulation window
```

That replaces equal microbatch weighting with the token-weighted full-batch contract. The normalization is applied to sequence CE, CE + ordinary teacher KL, aligned ULD, MiniLLM, and on-policy branches. The distributed `average_tokens_across_devices` scale mirrors `Trainer.compute_loss`.

## Measurement before the fix

On current `origin/main` (`e27e40c`), the real class body + real `Trainer.training_step` + one-layer Llama, with row lengths `[2,2,2,2,16,16,16,16]` (4 vs 60 shifted targets across two microbatches), produced:

```text
full gradient norm                 0.7767304778
GA=2 split gradient norm           1.3807157278
split / full norm ratio            1.7775995731
relative L2 error                  1.5322198868
cosine similarity                  0.5097218752
```

Environment: macOS arm64, CPU, no CUDA, torch 2.12.1, declared Transformers floor 5.16.1. The measurement was posted to #932 before implementation.

## Tests

The regression compiles the real nested trainer class via AST and exercises real `Trainer.training_step` and `Trainer._get_num_items_in_batch`. It compares the complete gradient vector for both sequence CE and live CE+teacher-KL, and retains the equal-length GA=4/8 control plus an identity-normalizer mutation control.

- `pytest -o addopts= tests/test_issue720_distill_grad_accum.py -q` — **6 passed**
- Distillation-focused suite across v0.53.2, ULD, #681/#682, v0.71.12 and v0.71.18 — **362 passed**
- `ruff check src/soup_cli/ scripts/ tests/ benchmarks/` — **passed**
- `git diff --check` — **passed**
- Full-suite qualification with Transformers 5.16.1 reached **6,480 passed, 22 skipped** before the pre-existing `tests/test_issue885_langfuse_error_ansi.py::test_stripped_body_renders_on_a_colour_console` failure. That exact failure reproduces unchanged in a detached `origin/main` worktree at `e27e40c`; the run was stopped after establishing the baseline failure. The other known local Rich-version baseline test from #941 was deselected. CI remains the full-matrix authority.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [x] Full Ruff
- [ ] Full pytest — blocked locally only by independently reproduced baseline Rich tests above
- [x] Updated `docs/training.md`
- [x] Added `changelog.d/0.75.0/943.fixed.md`


MakazhanAlpamys:
Reviewed at `265af6af`. **This is correct and the mutation is dead.** Two items before it lands, one of which is not about your code.

**Verified against the transformers 5.17.0 source rather than the description:** flipping `model_accepts_loss_kwargs` to `True` is the right lever — `trainer.py:2240` only computes `num_items_in_batch` when that flag is set, and `trainer.py:2035` only applies the fixed `1/GA` fallback when it is not. `_token_weighted_accumulation` rescales a microbatch mean `sum_ce_j / L_j` by `L_j / W` to `sum_ce_j / W`, which is arithmetically exact rather than an approximation, and it is applied on **all four** return paths (sequence CE, on-policy MiniLLM, aligned ULD, CE+teacher-KL) — I enumerated them, none returns an unweighted loss.

The load-bearing detail that makes one normalizer valid for both terms is that `_compute_distill_term`'s `_masked_mean` uses the *same* denominator (shifted `labels != -100`) as your token count. Worth saying out loud because it is why scaling the KL term by the label count is not a category error.

The fallback is also safe in the right direction: with `num_items_in_batch is None` the normalizer is a no-op and `training_step` applies `/GA`, so it degrades to `main`'s behaviour rather than to #720's bug.

**Mutation, anchor counted once, mutated module `ast.parse`d AND actually imported** (an import, because a `NameError` would produce a false "caught"):

```
weighted = loss * local_items / window_items.clamp(min=1)
  -> weighted = loss / self.args.gradient_accumulation_steps      # 1 occurrence, parses, imports

2 failed, 4 passed
E  Mismatched elements: 3480 / 3632 (95.8%)
E  Greatest absolute difference: 0.2587057948112488 (up to 1e-06 allowed)
```

Real numeric failures on complete gradient vectors, not a collection error. All 31 removed test lines accounted for: docstring rewrite, a parameter rename, and `_accumulated_grad_norm` refactored into `_accumulated_gradient` returning the full vector. **No assertion deleted or loosened**, the equal-length GA=4/GA=8 `1.0` control is retained verbatim, and the mutation control got stronger — it used to delete `__init__`, it now AST-replaces `_token_weighted_accumulation`'s body and still asserts the 4.0 scaling. Calling the real `trainer._get_num_items_in_batch(...)` instead of a test-supplied stand-in is the right change.

I also ran a real end-to-end distill train through `DistillTrainerWrapper.setup()/train()` (tiny-gpt2 student+teacher, LoRA r=4, GA=2, deliberately unequal row lengths) — completes normally with `model_accepts_loss_kwargs = True` on the constructed trainer, so the AST-compiled class in the test is not the only thing that works. 549 passed, 1 skipped across ten distill/ULD suites; ruff clean.

---

**1. Credit line, one word.** `changelog.d/0.75.0/943.fixed.md:2` reads `(#932 in #943)`. Every sibling in that folder uses `(#ISSUE by @author in #PR)` — `742.fixed.md` has `(#720 by @dchaudhari7177 in #742)`. Should be `(#932 by @Amix29 in #943)`. I care about this one specifically because a fragment without a handle assembles into the release as uncredited work, and reconstructing authorship at release time is exactly when it gets skipped.

**2. Your fragment contradicts #742's, and both are unreleased.** `#720`/`#742` appear nowhere in `CHANGELOG.md`, so both fragments assemble into the *same* next section. #742 says "The trainer now sets `model_accepts_loss_kwargs = False`" — precisely what this PR reverses. A reader of the shipped changelog would see an entry describing a mechanism the released code does not have. Please have your fragment say it supersedes #742's mechanism, or fold the two; either is fine, I would rather not do it myself at assembly time and get it wrong.

---

**One thing for a follow-up issue, not for this PR** — I am recording it because your tests do not cover it and the distill suite's own default base is the family it affects.

`distill.py:562` hardcodes the causal shift (`labels[..., 1:]`), but transformers does not: `Trainer._get_num_items_in_batch` shifts only when `self._loss_shifts_labels` is true, which is `LOSS_MAPPING.get(model.loss_type) is ForCausalLMLoss`. Measured here:

```
llama  _loss_shifts_labels=True   window=64  local=[4, 60]   numerator == denominator
gpt2   _loss_shifts_labels=False  window=72  local=[4, 60]   numerator != denominator
```

`GPT2LMHeadModel.loss_type` is `None` because the class is not named `*ForCausalLM`, so transformers counts labels unshifted while you count them shifted — an upstream inconsistency, since GPT-2's own `loss_function` still falls back to `ForCausalLMLoss`.

**The #932 contract itself still holds on that family** — chunk-vs-full ratio `1.0000000000`, max abs diff 1.1e-08. (A first probe showed 1.0815 and that was an artifact of GPT-2 having dropout on by default where Llama does not; with `*_pdrop=0.0` it is exactly 1.0. Flagging so nobody chases it.) What changes is the absolute scale: `sum_ce/72` instead of `sum_ce/64`, i.e. ~11% lower loss and gradient **even at GA=1**. It only bites when a row's *first* label is a trained target, and with Soup's default assistant-only masking `labels[:, 0]` is `-100`, so the discrepancy is exactly zero on the default path. It would surface on `train_on_responses_only: false` or plaintext distill rows with a `*LMHeadModel`-named student.

If you want it, the fix is to use `self._loss_shifts_labels` to pick the slice rather than hardcoding, and to soften `docs/training.md`'s "shifted, non-masked training targets" to match what transformers actually counts. Open it as an issue and claim it if you like — it is genuinely separate work.

Posting the measurement to #932 before the implementation, and then the post-fix measurement with the same rows and seeds, is what made this reviewable in one pass. That is the second time today.

Amix29:
Addressed both landing items at `cd55ef3`, after rebasing onto current `origin/main@876b2b1`.

- the fragment now credits `#932 by @Amix29 in #943`
- it explicitly says the token-weighted window normalization supersedes #742s fixed-per-microbatch mechanism, so the two unreleased entries no longer contradict each other

Post-rebase validation: `6 passed` in `test_issue720_distill_grad_accum.py`, ruff clean, diff check clean.

MakazhanAlpamys:
Re-checked at `cd55ef3e`. **Credit line closed** — `(#932 by @Amix29 in #943)`, byte-for-byte the sibling form. The core mutation still dies after the rebase (anchor 1 occurrence, real import, then 95.8% of gradient elements mismatched), and #943 coexists with #736's log-space kernel: 40 passed on the two suites together, 464 across eight further distill files. The load-bearing reason it coexists is that #736 rewrote the divergence branches but **did not touch the denominator** — `_masked_mean` still uses `mask = labels != -100`, the same count `_token_weighted_accumulation` normalises by.

**The #742 contradiction is named but not retracted, and I assembled the changelog to see what a reader actually gets rather than guessing.**

`scripts/assemble_changelog.py:63` sorts fragments lexicographically by filename, so `742` always precedes `943`. In the assembled output both land in the same `### Fixed` block, with **#742's entry at line 53 and your retraction at line 155** — last in the section, 102 lines later. A reader meets the false sentence first, with no forward pointer.

And both halves of that sentence are now false of the shipped code. `742.fixed.md` says:

> The trainer now sets `model_accepts_loss_kwargs = False` after `Trainer.__init__`, as the Transformers docstring prescribes for a `compute_loss` that does not use `num_items_in_batch`.

`distill.py:541` is `self.model_accepts_loss_kwargs = True`, and `compute_loss` *does* use `num_items_in_batch` — it is a parameter at `:548` and it drives `_token_weighted_accumulation`. Your retraction says "supersedes #742's fixed-per-microbatch compensation", which is true and generic, but it never names `model_accepts_loss_kwargs`, so the one sentence that is actively wrong is never corrected by name.

**Please fix it in `742.fixed.md` directly** — `False` -> `True` plus the clause after it. Both fragments are unreleased, so there is no history to falsify, and editing someone else's unassembled fragment is the cheapest correct option. #742's final sentence about the test is still accurate and should stay: the GA=1/4/8 control survives verbatim at `tests/test_issue720_distill_grad_accum.py:233-234`. Adding the flag name to your own sentence as well would not hurt.

---

**One LOW finding, and it is about diagnostics rather than correctness.** I ran the extra mutation `self.model_accepts_loss_kwargs = True` -> `False` at `distill.py:541` — i.e. exactly the mechanism #742 still describes. It is caught, 5 failed, but **every failure is a harness `TypeError`, not a behavioural assertion**:

```
num_items_in_batch = trainer._get_num_items_in_batch(batches, trainer.args.device)
TypeError: int() argument must be ... not 'NoneType'
```

With the flag off, transformers returns `None` and the helper's `int()` dies before any gradient is compared. Detected, so not blocking — but whoever flips that flag next, or the transformers release that flips it for us, gets an opaque crash in the test harness instead of "the accumulated gradient no longer matches".

The one test that *passes* under that mutation is `test_distill_trainer_sets_loss_kwargs_contract_after_trainer_init`, whose docstring says it pins the assignment's **position** relative to `super().__init__` without pinning how it is spelled — so it would accept `= False`. A one-line `assert trainer.model_accepts_loss_kwargs is True` on a constructed trainer would name the mechanism the two fragments disagree about and turn that `TypeError` into a real failure. Your call whether it belongs here or in the `_loss_shifts_labels` follow-up.

CI is queued on `cd55ef3e`. Fix the fragment and I will merge when it lands.

Amix29:
Addressed the remaining changelog blocker at `044e07b`, rebased onto current `origin/main@fcca8ef`.

- rewrote `742.fixed.md` directly so it now states `model_accepts_loss_kwargs = True`, full-window `num_items_in_batch`, and token-share weighting completed in #943; the GA=1/4/8 test sentence remains intact
- added a live constructed-trainer assertion that `trainer.model_accepts_loss_kwargs is True`, so flipping the flag to `False` now fails by name instead of only causing the downstream harness `TypeError`

Post-rebase validation: `6 passed` in `test_issue720_distill_grad_accum.py`; Ruff clean; `git diff --check` clean.

MakazhanAlpamys:
Re-checked at `044e07b0`. **Blocker closed, and closed the right way — by name.**

`742.fixed.md` now reads `model_accepts_loss_kwargs = True`, says the trainer consumes the full-window `num_items_in_batch`, and carries a forward pointer ("as completed in #943"). Both halves match the code: `distill.py:541` sets `True`, and `num_items_in_batch` is a `compute_loss` parameter at `:548` driving `_token_weighted_accumulation`. Dropping the clause that described the *pre*-#742 mechanism was right too. And #742's still-accurate sentence about its test survives verbatim — I checked the control it names is intact at `tests/test_issue720_distill_grad_accum.py:232-240`, still parametrized over GA 4 and 8 against a GA=1 reference.

**I assembled the changelog again rather than assuming the edit was enough**, since the original problem was about what a reader meets first:

- line 53 (#742): now accurate, with the pointer
- lines 175-180 (#943): the supersession entry

The 122-line gap is unchanged, and it no longer matters — the ordering was only harmful while the first entry was false. A reader who stops at #742's entry now gets a correct description of the shipped mechanism.

Core mutation survives the rebase: `weighted = loss * local_items / window_items.clamp(min=1)` -> `loss / self.args.gradient_accumulation_steps` (1 occurrence repo-wide, real import) gives 95.8% and 94.4% of gradient elements mismatched on the two comparison tests. #736 and #943 still coexist: 40 passed with both suites together, which is exactly 6 + 34, so no interference.

**And you took the LOW note I said was your call.** `assert trainer.model_accepts_loss_kwargs is True` at `tests/test_issue720_distill_grad_accum.py:331` changes the flag mutation from 5 failures to **6**, and more importantly `test_distill_trainer_sets_loss_kwargs_contract_after_trainer_init` — the one test that *passed* under that mutation before — now fails by name on a real assertion instead of the mechanism being caught only through an opaque harness `TypeError`. That is the difference between detecting a regression and explaining it.

Your own sentence still says "supersedes #742's fixed-per-microbatch compensation" without naming the flag. I said that would not hurt; with #742 now correct it costs nothing to leave, so leave it.

17 passed on the fragment suite, ruff clean, no new findings. CI is queued on `044e07b0`; I will merge when it lands.

MakazhanAlpamys:
Merged as `4b93b4ab`. Thank you — and the mutation number is what makes this one unambiguous.

Substituting `loss / gradient_accumulation_steps` for the weighting mismatches **95.8%** and **94.4%** of gradient elements. That is not a test that happens to notice; that is a test that could not possibly miss. One occurrence repo-wide, real import confirmed, and #736's suite and yours coexist at exactly 6 + 34 = 40, so neither is masking the other.

The defect itself is the quiet kind: equal microbatch weighting across an accumulation window produces a gradient that is not the full-batch gradient whenever the microbatches carry different numbers of trained tokens — short sequences over-weighted, long ones under-weighted — and the loss curve looks entirely normal throughout.

**The part I want to name is the note you took that I had left as your call.** Asserting `trainer.model_accepts_loss_kwargs is True` turns the flag mutation from 5 failures into 6, and more importantly makes `test_distill_trainer_sets_loss_kwargs_contract_after_trainer_init` — the one test that *passed* under that mutation — fail **by name on a real assertion** rather than surfacing as an opaque harness `TypeError`. That is the difference between a suite that detects a regression and one that explains it, and you did it on a suggestion I explicitly said would not hurt either way.

The 122-line CHANGELOG gap is unchanged and no longer matters: the ordering was only harmful while #742's entry was false, and a reader who stops there now gets a correct description of the shipped mechanism.

`merge-tree` clean against current `main`, 17 passed on the fragment suite, ruff clean.

Your `CONTRIBUTORS.md` entry is updated at `2e25fc3d`.

Issue #800, opened 2026-09-11
training.nvfp4 and fp8_attention are never applied on task: sft: the shared helper runs in 11 trainers and not in the default one


MakazhanAlpamys:
## The defect

`training.nvfp4`, `training.fp8_attention` and `training.kernel_auto_compose` are applied in exactly one place, `apply_v028_speed_memory` in `src/soup_cli/utils/v028_features.py`. Eleven trainers call it (bco, dpo, embedding, grpo, ipo, kto, orpo, ppo, pretrain, reward_model, simpo). **`trainer/sft.py` does not**, and neither do TTS and BitNet, which subclass `SFTTrainerWrapper`. So on `task: sft`, the default task, all three flags are accepted and then do nothing.

SFT inlines two of the helper's five features, Cut-CE (`sft.py:1312`) and FP8 training (`sft.py:1551-1569`, `_apply_quantization_aware`). The other three appear nowhere in `sft.py`:

```

$ grep -n "nvfp4\|fp8_attention\|kernel_auto_compose\|apply_v028" src/soup_cli/trainer/sft.py

$

```

## Evidence

`utils/v028_features.py` is the only reader of each flag in `src/`:

```

89   if getattr(tcfg, "fp8_attention", False):   -> apply_fp8_attention(model, ...)

101  if getattr(tcfg, "nvfp4", False):           -> apply_nvfp4(model)

112  if getattr(tcfg, "kernel_auto_compose", False): -> _bench_and_pick_kernel(...)

```

The test meant to catch this exempts SFT on grounds it does not check. `tests/test_trainer_coverage_v035.py:358-365`:

```python

# SFT predates the shared helper extraction (v0.33.0 #43) and inlines

# apply_cut_ce / apply_fp8_training / kernel_picker directly; ...

has_inline_features = (

    "apply_cut_ce" in src

    and "apply_fp8_training" in src

)

```

The comment names `kernel_picker`; the assertion checks only Cut-CE and FP8, and `sft.py` never imports `kernel_picker`. `fp8_attention` and `nvfp4` were added to the helper later (v0.71.21, #141) and the exemption was never revisited. #45, which closed the `kernel_auto_compose` benchmark, proposed wiring it "in `SFTTrainerWrapper.setup`", the one trainer that never got it.

## Second gap: nvfp4 has no task gate

[...]

PR #945, opened 2026-09-13
fix(docs): refresh shipped feature descriptions

### LINKED ISSUE #823: Help text, 20 config field descriptions and 7 docs passages still say "schema-only" / "deferred" / "lands in v0.x.1" for features that have shipped; advise run tells users to pass a --no-record flag that does not exist
## The defect

Many features shipped as a validated schema first and went live a release or more later. The help strings, `Field(description=...)` texts and docs passages written for the schema release were never updated, so today they tell users that working features do not work. One help string also points at a flag that does not exist.

## Evidence 1: CLI help (`soup <cmd> --help`)

| Command | Says | Reality |

|---|---|---|

| `soup serve --steer` (`commands/serve.py:238-243`) | "Schema-only in v0.62.0; live decode hook ships in v0.62.1." | live: `serve.py:934` calls `install_steering_hook` |

| `soup data recipe` (`commands/data.py:2745`) | "Validate a Data Recipe DAG (live runner deferred)." | `--execute` runs `utils/recipe_run.run_recipe` (see the separate issue about its offline LLM nodes) |

| `soup eval quant-check` (`commands/eval.py:1292-1299`) | "Wiring live model loading is post-v0.26.0; until then, this runs with a stub generator" | loads both models via `make_model_generator` (`eval.py:1349-1353`); the stub is only a fallback when loading fails |

| `soup advise run --record` (`commands/advise.py:243-246`) | "Use `--no-record` to skip." | the option is a plain `--record` flag, default off; `--no-record` is rejected |

```

$ soup advise run rows.jsonl --goal "classify support tickets" --no-record

No such option '--no-record'. Did you mean '--record'?

exit=2

```

Separately, 13 help strings still carry internal release-planning labels (the letter after a version number) that the project's own conventions keep out of user-facing text: `serve --steer-strength`, `build`, `expect`, `data from-traces`, `data recipe`, `data gen-magpie`, `data persona-mix`, `data brain-rot`, `eval gate-install`, `adapters pr`, `adapters bisect`, `lock`, `iterative-dpo`.

## Evidence 2: config field descriptions (`config/schema.py`)

Each of these descriptions says schema-only / deferred, and each field is read by live code outside `schema.py`:

| Field (line) | Read by |

|---|---|

| `TrainingConfig.expand_layers` (1184) | `utils/block_expansion.py` |

| `grpo_variant` (1487) | `trainer/grpo.py` |

| `vllm_sleep_mode` (1527) | `trainer/grpo.py` |

| `hub` (1648) | `utils/hubs.py`, `data/loader.py`, `commands/train.py` |

| `tts_family` (1660) | `trainer/tts.py` |

| `teacher_model` (1726) | `trainer/distill.py` |

| `reasoning_effort` (1814) | `data/sft_format.py` |

| `kv_cache_type` (1832) | `utils/kv_cache.py`, `commands/serve.py` |

| `fp8_attention` (1840), `nvfp4` (1848) | `utils/v028_features.py` |

| `echo_trap_enabled` (1912) | `utils/peft_wiring.py` |

| `rl_checkpoint_save_every_steps` (1947) | `utils/peft_wiring.py` |

| `minillm_enabled` (1993) | `trainer/distill.py` |

| `reward_hack_detector` (2088) | `utils/peft_wiring.py` |

| `use_longlora` (2804) | `trainer/sft.py` |

| `unlearn_method` (3750) | `trainer/unlearn.py` |

| `citation_faithful` (3842), `citation_style` (3851, "'inline' / 'footnote' are stub placeholders") | `trainer/raft.py`, `trainer/sft.py` |

| `grace_codebook` (3889) | `utils/knowledge_edit.py` -> `apply_grace_edit` |

| `DataConfig.prompt_strategy` (647) | `data/sft_format.py` |

Deliberately NOT listed: `long_context_grpo` and `vision_grpo` also say schema-only, and a search finds no reader outside `schema.py`; `quantization: bitnet_1.58` says schema-only and `utils/bitnet.build_bitnet_trainer` has no caller. Those descriptions may be accurate, and "read by nothing" is a different problem.

These strings are what anyone reading the schema, or an editor autocompleting from its JSON Schema, sees.

## Evidence 3: docs

- `docs/data.md:592` — remote datasets: "fsspec backend wiring lands in v0.42.1" (the same page, line 834, documents it working).

- `docs/data.md:876` — "Live offline runner against a local model lands in v0.45.1."

- `docs/data.md:897` — mix proxy: "Live wiring ... is the v0.48.1 deliverable" (`soup data mix --live` exists).

- `docs/backends-and-ops.md:398` — "The hub adapter is schema-only in this release".

- `docs/peft-and-efficiency.md:144` — "LongLoRA S² (schema-only this release)" and "Mixtral excluded"; `validate_longlora_compat(model_name='mistralai/Mixtral-8x7B-v0.1', ...)` accepts it today.

- `docs/adapters-and-governance.md:92` — VeRA/VB-LoRA bank and MoLE "ship as schema-only in v0.67.0".

- `docs/backends-and-ops.md:971` — plugin trainer wiring "lands in v0.45.1" (this one is tied up with the plugin-loading issue; fix together).

## Why it matters

A user who reads `--help` or the docs is told that working features (LongLoRA, KV-cache types, MiniLLM, the reward-hack detector, steering at serve time, ...) are not usable yet, and one help line sends them to a flag that errors. The stale "deferred" notes also hide the few genuinely unwired fields among them, because true and false notes read the same.

## Proposed fix

Rewrite each string to describe current behaviour, with no version promises. Then add a ratchet test that walks every Typer help string and every `Field(description=...)` in `config/schema.py`, and fails on `schema-only`, `deferred to v`, `lands in v` or `ships in v` unless the line is on a short allowlist of genuinely unwired features (`long_context_grpo`, `vision_grpo`, ...), plus a check for internal release-phase labels in help text.

## Acceptance criteria

- [ ] `soup advise run --help` no longer mentions `--no-record` (or the flag is made a real `--record/--no-record` pair).

- [ ] None of the help strings, field descriptions or docs lines above says "schema-only", "deferred" or "lands in" for a shipped feature.

- [ ] A ratchet test over help strings and schema descriptions — fails on the current tree.

---

- #817 — soup data recipe --execute has no provider option, so every llm_text node writes "llm_text(offline): ..." i...
- #819 — The plugin system is unreachable from the CLI: load_plugins() has no caller, so `soup plugins` always lists...
- #822 — docs/commands.md calls itself "the full soup command list" but omits 23 of 219 registered commands; 2 (eval...
- #825 — quantization: bitnet_1.58 is documented as live, but nothing routes to BitNetTrainerWrapper: an SFT run die...

Amix29:
## Summary

- refresh stale schema descriptions and Typer help text that still announce already-shipped work as future
- replace outdated internal release-phase wording with current capability or limitation descriptions
- update the affected documentation pages and remove the nonexistent `--no-record` recommendation
- add a source-level regression ratchet covering schema descriptions, CLI help, the advise flag, and the named docs

## Reproduction

Before this patch, the regression test found 27 stale Pydantic field descriptions and 17 stale user-visible Typer help strings, plus the unavailable `--no-record` suggestion and the documentation passages listed in #823.

## Validation

- final rebase on `origin/main@876b2b1`: `13 passed`, ruff clean, diff check clean
- broader targeted schema/CLI/docs suite: `437 passed, 1 skipped`
- full suite on the immediately preceding `origin/main@e93a1b2`: `21384 passed, 106 skipped, 6 deselected`, 83.42% coverage

Of the six deselections, five are the repository default `smoke` exclusions and one is the explicitly isolated pre-existing Rich-width control test in `test_issue204_langfuse_pull.py`; that test fails identically on a clean checkout of the same `origin/main`. No #823 test was deselected. The two intervening upstream commits only change distillation math and contributor documentation; the final focused suite and lint were rerun after rebasing them.

Closes #823


MakazhanAlpamys:
Reviewed at `8807851e`. **24 of 27 rewritten schema descriptions and all 17 help strings check out against today's code**, the change is genuinely string-only, and the ratchet is a real ratchet. Three replacements are newly wrong, which is the one failure mode a wording PR has — a stale claim swapped for a different false one — and one of them describes what a checkpoint contains, which a user would act on.

**Scope first, because 16 files and eight command modules deserve the check:** `git diff origin/main...HEAD -- src/` reduces entirely to string literals, docstring bodies and one `help=` kwarg. Zero changes to logic, defaults, validators, control flow or imports. No test asserts on any changed `Field(description=...)`, descriptions do not appear in `model_dump()`, and the 2220-test config/schema slice passes. Nothing smuggled.

**The ratchet fails on `main` exactly as it should** — 27 schema offenders, 17 help offenders, plus the `--no-record` and docs checks, and the offender counts match your change set precisely. It is AST-based (312 schema descriptions, 1295 help strings across 90 files) with floor assertions so a broken scanner cannot pass vacuously. Two mutations, each anchor counted once, each file re-imported: appending `" (v0.67.0 Part E)"` to `commands/lock.py:24` is flagged **naming the file**, and putting `(schema-only)` back into `use_longlora` is flagged **naming field and line**.

The `--no-record` removal is right — `grep '"--no-record"\|no_record' src/` is zero hits, and `advise.py:240` declares a plain `--record`, so "omit `--record` to skip" is exactly correct.

---

**1. HIGH — `config/schema.py:1951-1952` invents checkpoint contents.**

> "the callback persists trainer, optimizer, **scheduler, RNG**, and optional rollout state."

```
grep -ci "scheduler" utils/rl_checkpoint.py            -> 0
grep -ci "rng\|get_rng_state" utils/rl_checkpoint.py   -> 0
```

`save_checkpoint` (`rl_checkpoint.py:299-351`) writes the policy adapter, `optimizer.pt` when `include_optimizer_state`, and a manifest. No scheduler, no RNG, no trainer state — and `has_rollout_buffer` is set straight from the config flag without writing a buffer. Someone reads this to decide whether a run resumes exactly; it does not. Suggested: *"the callback persists the policy adapter, the optimizer state, and a manifest."*

**2. MEDIUM — `schema.py:3890-3891` claims a live path for fields nothing consumes.**

> "Knowledge editing performs live codebook lookup and transactional writes."

`training.grace_codebook` / `_size` / `_dim` are read **only** by schema validators. `apply_grace_edit` is reached from `knowledge_edit.apply_edit` on `method == "grace"`, driven by `soup edit set --method grace` and an `EditPlan`, using the module constants `_GRACE_GRAD_STEPS` / `_GRACE_LR` — it never reads these fields. `utils/grace_codebook.py:9` still says "Schema-only release". Your own wording for `long_context_grpo` and `vision_grpo` — *"Compatibility validation only: no trainer consumes this flag yet"* — is the honest template, and it already passes your ratchet.

**3. MEDIUM — `docs/peft-and-efficiency.md:144` reports a count nothing prints.**

> "installs the shifted-sparse attention forward override and **reports how many attention modules were patched**."

`trainer/sft.py:1953` prints a bare `"LongLoRA S² attention override active"`. `LongLoRAForwardOverride` does track `self._patched` (`longlora.py:428`), but nothing prints or returns `len(self._patched)`. The Mixtral-in-allowlist half of that sentence is correct.

---

**4. MEDIUM, and this is the one I would think hardest about — the ratchet fires on legitimate text, with no allowlist and no escape hatch.**

I probed it rather than assuming:

| probe | flagged |
|---|---|
| "Deprecated alias; support is removed in v0.80.0." | no |
| "Deprecated alias; removal **ships in v0.80.0**." | **STALE** |
| "Unknown keys warn today; rejection **lands in v0.80.0**." | **STALE** |
| "enforcement is **deferred to v0.80.0** per the deprecation policy" | **STALE** |
| "Reserved: validated but not yet consumed **(schema-only)**." | **STALE** |
| "Applies the **post-v2** attention kernel." | **STALE** |

The first three are *forward*-looking deprecation deadlines — which this project has literally shipped: #627 in v0.74.0 warned on unknown keys with a stated v0.75 rejection deadline, and v0.75.0 honoured it. Writing that notice into a `Field(description=...)` would now turn CI red. `post-v\d` is over-broad and catches an unrelated technical phrase.

This matters more than a normal nit because **a guard that fires on correct code gets deleted rather than fixed** — this repo has done exactly that, to a scanner that flagged 31 correct assertions. Issue #823's proposed fix asked for "a short allowlist of genuinely unwired features"; this ships a hard ban instead. Minimum: an allowlist constant, or narrow `post-v\d` and put the sanctioned phrasing in a comment in the test so the next contributor has somewhere to go.

**5. LOW — no `TestTheScannerCanActuallyFail`.** I proved externally that it can fail, but the repo convention is that a guard finding zero offenders carries its own proof. Its blind spots are worth a line too: `@app.command` without parens is skipped (the decorator must be an `ast.Call`), only `config/schema.py` is scanned so `cans/schema.py` and friends are not, and an f-string description silently vanishes.

**6-8. Three hedges that are technically true but hide which paths qualify.** `fp8_attention` / `nvfp4` at `:1841,1848` say "Compatible trainer paths convert…" — but `apply_v028_speed_memory` is **not** called from `trainer/sft.py` (grep count 0, #800), so on the default `task: sft` both still silently do nothing. `grpo_fp16` at `:1510` forwards `fp16=True, bf16=False` only on CUDA (`grpo.py:288-296`); non-CUDA returns `{False, False}`. `hub` at `:1649` says "download and upload adapter" — download is driven by this field (`commands/train.py:1311`), upload comes from `soup push --hub`, not from `training.hub`.

**9. `changelog.d/0.75.0/823.fixed.md` should be `945.fixed.md`.** `CONTRIBUTING.md:351-354` says to prefer the PR number, every other fragment in that directory does, and it is also a collision risk if anything else touches #823. The text and the `(#823 by @Amix29 in #945)` credit are correct.

---

**Two things you did that I want to name, because they are the opposite of the failure this PR is about.** `docs/data.md:921` says `soup data recipe --execute` writes the `llm_text(offline)` placeholder because no judge provider is passed — that is #817, and you documented it instead of writing a sentence that sounds finished. Same for `backends-and-ops.md:981`: hooks are fanned, `load_plugins` has no caller, and you said so. Choosing to keep an honest limitation in a PR whose whole purpose is to remove stale claims is the right instinct.

Fix 1-3, rename the fragment, and do something about the allowlist, and I will merge.

Amix29:
Addressed the review at `a00304b` on current `origin/main@876b2b1`.

Blocking items:
- RL checkpoint copy now names only the policy adapter, optional optimizer state, and manifest
- all three GRACE codebook fields now state that they are validated but not consumed at runtime
- LongLoRA docs no longer claim a patched-module count is reported
- the changelog fragment is now `945.fixed.md`
- the ratchet is version-aware: expired promises fail, future release notices pass, `post-v2` technical wording is ignored, and intentionally schema-only fields have an explicit allowlist with a required current-limitation comment
- `test_the_schema_scanner_can_actually_fail` drives a temporary schema through the AST scanner and proves the failure path

I also tightened the three scope hedges from the review: GRPO FP16 now says CUDA-only, `training.hub` is described as download-only with `soup push --hub` independent, and FP8-attention/NVFP4 name the exact trainer families while stating that default SFT does not consume them.

Post-change validation: `21 passed` across the #823 ratchet, ANSI help guard, and config tests; ruff clean; diff check clean.

MakazhanAlpamys:
Re-checked at `a00304b1`. **All three false replacements are fixed, and the ratchet's false-positive problem is solved better than I asked for.** Waiting only on CI.

**The three, verified against code rather than against your summary:**

| # | new text | evidence |
|---|---|---|
| 1 | "persists the policy **adapter, optional optimizer state, and a manifest**" | `rl_checkpoint.py:322-357` writes exactly those three. `grep -ci scheduler` -> 0, `grep -ci "rng\|get_rng_state"` -> 0. It also now matches the module's own docstring at `:253` verbatim |
| 2 | "no trainer **or knowledge-edit command** consumes this field yet" | every `grace_codebook*` reference outside `config/schema.py` is either a string artifact-kind name (`commands/edit.py:312`, `registry/store.py:52`) or a validator. `apply_grace_edit` takes an `EditPlan` and uses `_GRACE_GRAD_STEPS` / `_GRACE_LR` |
| 3 | count claim removed from the LongLoRA sentence | `sft.py:1950-1953` prints the bare `"LongLoRA S² attention override active"`; `grep -rn "len(self._patched)"` is zero hits repo-wide |

The "or knowledge-edit command" clause in 2 is a tightening over the template I suggested — it closes the `soup edit set --method grace` path, which does exist and still does not read those fields. That is the distinction I would have had to raise on the next round.

**The three hedges I raised as non-blocking were also tightened, and all three now check out.** `grpo_fp16` names the device gate (`grpo.py:288-289` returns `{False, False}` before the flag branch); `hub` says the upload destination is chosen independently by `soup push --hub` (`cfg.training.hub` is read exactly once outside the schema, at `commands/train.py:1311`); and `fp8_attention`/`nvfp4` now **enumerate** the eleven wrappers that call `apply_v028_speed_memory` and explicitly exclude the default SFT path. I checked that enumeration is complete rather than plausible — the caller set is exactly `{bco, dpo, ipo, kto, orpo, simpo, grpo, ppo, pretrain, reward_model, embedding}`, `sft.py` absent.

**The false-positive fix is the part I want to name.** You did not add an exception list for the phrases I probed — you made the check **version-aware**:

| probe | before | now |
|---|---|---|
| "removal **ships in** v0.80.0" | STALE | **clean** |
| "rejection **lands in** v0.80.0" | STALE | **clean** |
| "enforcement **deferred to** v0.80.0" | STALE | **clean** |
| "Applies the **post-v2** attention kernel" | STALE | **clean** |
| "(schema-only)" | STALE | still flagged (correctly) |

`VERSIONED_RELEASE_PROMISE` still matches those verbs but only fires when the named version is `<= __version__`, and `INTERNAL_RELEASE_PHASE` narrowed to `v\d+.\d+ part [a-z]`. So a promise for a future release passes today **and starts failing the release after it should have shipped** — the guard arms itself on the deadline. That is strictly better than the allowlist #823 asked for, because an allowlist would have needed a human to remember.

`ALLOWED_SCHEMA_ONLY_FIELDS` exists as the escape hatch and is currently empty, with the bar documented at `:33-35`. Both `TestTheScannerCanActuallyFail`-style tests are present now, including one that drives a temp schema file through the real AST scanner.

**Three mutations, each anchor counted once, each mutated module actually imported and the value read back from the live object** (so the ratchet is scanning real content, not source text): appending `" (v0.67.0 Part E)"` to `lock.py:24` is flagged naming file and line; `(schema-only)` on `use_longlora` is flagged naming field and line; and `" Live lookup lands in v0.62.1."` on `grace_codebook` is flagged — that last one proves the new version-aware branch fires on a real file, not only on the fixture.

**String-only confirmed by AST equivalence rather than by grep:** I parsed both revisions of all ten changed `src/` files, blanked every string constant, and compared `ast.dump` — **10 files, 0 structural changes**. Because defaults are AST nodes, that also proves no non-string default moved. 49 passed on the guard suites, 2212 passed on the config/schema slice, ruff clean, fragment correctly renamed to `945.fixed.md` with the right credit.

---

**Four LOW notes, none blocking, all for later:**

- `tests/...:67` — the allowlist is schema-only. `_help_text_is_stale(text)` takes no allowlist, so a Typer `help=` that legitimately needs the phrase "schema-only" has nowhere to go. Nothing needs it today.
- `:36` — the allowlist keys on a bare field name, and `baseline` / `forgetting_threshold` each appear twice in `config/schema.py` across model classes, so one entry would silently exempt both. Harmless while empty; `(class, name)` would be tighter.
- `:133`, `:146` — bare `assert offenders == []` with no message. The comment at `:33-35` exists but has to be found; one `f"...see ALLOWED_SCHEMA_ONLY_FIELDS"` in the assertion closes the loop.
- Your merge-base is `876b2b1` and `main` is now `fcca8efb` (two `docs(contributors)` commits). Nothing in the delta touches these files, but I need green on the current head.

**And the thing I want on the record:** you did not delete the three false claims and stop — you replaced them with *enumerable* ones, which is harder and is what makes them checkable next time. You also kept the honest limitations from round one rather than polishing them away now that you were editing those lines anyway. Both verified still true.

MakazhanAlpamys:
@Amix29 — base move, not a review item.

**Conflicting** on one file:

```
docs/data.md
```

caused by `fb53831d` (#922, "require an explicit recipe provider mode"), which rewrote the recipe section you also touch. One-file rebase.

Amix29:
Update pushed in f35bb3 after rebasing on current main. Main had advanced in both conflicting documentation surfaces: plugin discovery/activation is now wired through #935, and dataset providers are now wired through #922. I resolved the conflicts by preserving those shipped behaviors and keeping this PR focused on removing stale descriptions. Validation: 60 targeted tests passed, including the #823 documentation checks, changelog validation, and the consumer ratchet; full Ruff is clean. GitHub reports the branch mergeable. Ready for re-review.

Amix29:
Rebased again on current main at cbe7363 and pushed as abebf83. The new conflict came from #917 in src/soup_cli/commands/eval.py. I preserved the shipped #917 behavior and documentation: model-loading failures no longer imply an automatic stub fallback, and the command documents its current exit codes. Validation after the resolution: 90 passed across the #823 stale-description ratchet, changelog validation, config-consumer ratchet, and quant-check suite; full Ruff is clean. Ready for re-review.

MakazhanAlpamys:
Reviewed `abebf836`. **Merge-ready, waiting only on CI.** Nothing needed from you.

You hit a second rebase conflict after the first and resolved both in favour of main's shipped behaviour, which is the part worth crediting: **four hunks were dropped, zero added**, and one of them was your own sentence. Your text said *"automatic CLI plugin discovery and activation are not yet wired"* — true when you wrote it, false since #935 landed (`load_plugins()` now has callers at `commands/plugins.py:51,67,80` and `monitoring/plugin_callback.py:29`). Dropping your own paragraph rather than force-keeping it was the right answer, and the same for the `eval.py` hunk, where #917's new docstring already states the full 0/1/2/3 taxonomy so there is nothing left for #823 to fix there.

**Zero removed test lines**, and the guard is byte-identical to the reviewed revision — so no weakening is even possible in this delta.

**The mutation matters here more than usual**, because a docs PR is the kind where people ask whether CI can detect anything at all. It can, in both directions:

| mutation | result |
|---|---|
| `use_longlora` description regains `"(schema-only)"` | fails, naming `use_longlora:2805` — read off the **live** `TrainingConfig.model_fields`, not the source text |
| `docs/peft-and-efficiency.md` regains `"(schema-only this release)"` | fails, naming the file and the string |
| `advise.py` help regains `"Use --no-record to skip."` | fails through the **real CliRunner** with ANSI stripped |
| a description gains `"lands in v0.62.1"` (**expired**) | fails, naming the field |
| the same field gains `"lands in v0.99.0"` (**future**) | **6 passed** — correctly silent |

The last pair is the one I care about: it re-proves the version-aware branch on a real file in both directions, which is the MEDIUM I raised earlier — a guard that fires on correct code gets deleted.

**Six factual claims spot-checked by execution rather than accepted**, all true: the LongLoRA allowlist really includes Mixtral and the forward override really installs during SFT setup; `grpo_fp16` really returns early on non-CUDA before the flag is read; `apply_v028_speed_memory` really has eleven callers and `trainer/sft.py` is genuinely absent from them; `rl_checkpoint` really persists adapter + optional optimizer + manifest and contains **zero** scheduler or RNG references (the old text claiming otherwise was the HIGH finding); `grace_codebook*` really reaches no trainer; and `hub` really is read exactly once, at `commands/train.py:1367`, independently of `soup push --hub`. Plus `soup advise run --help` live: `Omit --record to skip.`, no `--no-record`.

284 passed on the full ratchet set on both trees; 1823 passed, 21 skipped across 27 files touching the eight changed command modules; ruff clean.

One follow-up that is **not** this PR's scope, for whenever: the ratchet scans `Field(description=...)`, `help=` kwargs and command docstrings, so two stale promises survive in places it does not look — `commands/steer.py:11-13` (module docstring, says the live kernels "land in v0.62.1" when they shipped in v0.71.10) and `config/schema.py:1655` (a comment). Neither is user-visible and the scan's boundaries are declared in the test file, so this is a future issue rather than a finding against you.

MakazhanAlpamys:
Merged as `0bbe35c3`. Thank you — three of yours landed in this pass.

**The guard fires in both directions, and that is the half that matters.** A description regaining an *expired* promise fails naming the field — read off the live `TrainingConfig.model_fields`, not the source text — while the same field given a genuinely *future* notice stays silent. A guard that fires on correct code gets deleted, so proving the quiet case on a real file is worth more than proving the loud one.

**Six factual claims spot-checked by execution rather than accepted**, all true: the LongLoRA allowlist really includes Mixtral and the override really installs during SFT setup; `grpo_fp16` really returns early on non-CUDA *before* the flag is read; `apply_v028_speed_memory` really has eleven callers with `trainer/sft` genuinely absent; `rl_checkpoint` really contains **zero** scheduler and RNG references (the old text claiming otherwise was the HIGH finding); `grace_codebook*` really reaches no trainer; and `hub` really is read exactly once, independently of `soup push --hub`.

**The rebase is where you earned the most credit.** Two conflicts, and you resolved both toward main's shipped behaviour — including dropping *your own* sentence about plugin discovery once #935 wired it, rather than force-keeping text you had written. Zero removed test lines; the guard is byte-identical to the reviewed revision.

One follow-up I will file rather than ask of you: the scan covers `Field(description=...)`, `help=` kwargs and command docstrings, so two stale promises survive where it does not look — `commands/steer.py:11-13` and a comment at `config/schema.py:1655`. Neither is user-visible.

Credited in `CONTRIBUTORS.md` (`f491ac5b`). #823 closed with credit.

PR #946, opened 2026-09-13
fix(sweep): rank diverged runs last

### LINKED ISSUE #944: soup sweep ranks a diverged arm by a NaN key, so it lands anywhere in the leaderboard
## What

`commands/sweep.py:504` ranks sweep arms with

```python
sorted(results, key=lambda r: r.get("final_loss", float("inf")))
```

Since **#908**, a diverged run's `final_loss` is `nan` rather than its last finite value — which is the correct and deliberate change, because the previous behaviour reported a diverged run as healthy. But `nan` does not compare meaningfully, so it poisons the ordering rather than sorting last.

Measured: `sorted([1.0, nan, 0.5])` leaves `0.5` **after** `nan`.

## Why this is a follow-up and not a regression in #908

#908's own review asked for exactly this retention. Before it, `sweep.py:165` (`if final_loss and final_loss < best_loss`) could not select a diverged arm because `nan < x` is always `False` — the divergence was invisible, so the arm quietly lost. Now the divergence is visible, and the consequence is that it lands in an undefined position in the ranking instead.

That is a strictly better place to be, and it is a trade I asked for. The remaining work is small and belongs here rather than in a PR whose subject is the loss panel.

## Acceptance criteria

- [ ] Non-finite `final_loss` sorts **last**, deterministically, regardless of position in the input.
- [ ] The sweep summary says a diverged arm diverged, rather than silently placing it. A user reading a leaderboard should not have to notice a blank or a `nan` cell to learn that an arm blew up.
- [ ] A test over a mixed list — e.g. `[1.0, nan, 0.5, inf, 0.8]` — asserting the full resulting order, not just that the best is first. The mutation that must be caught: reverting to the bare `r.get("final_loss", inf)` key.
- [ ] Check `sweep.py:165` and `:168` in the same pass. `if final_loss and final_loss < best_loss` and the `early_stop` comparison below it both operate on the same value and both silently no-op on `nan` — decide whether early-stop should treat a diverged arm as "worse than threshold" (probably yes) and make that explicit rather than incidental.

`float("nan") < x` and `x < float("nan")` are **both** `False`, which is why this is easy to get half-right: a fix that only changes the `sorted` key leaves the two comparisons at `:165`/`:168` still silently skipping diverged arms. Handle all three, and write the test so it would fail if any one of them were reverted.

Do **not** fix this by filtering diverged arms out of the results. Dropping the row is how the original defect was created — the run happened, it cost GPU time, and the operator should see it.

Related: #899, #908 (the retention that surfaced this).

taljeon:
## What does this PR do?

Closes #944.

- Gives sweep summaries an explicit ordering: finite completed losses sort numerically first, non-finite completed arms follow, and failed rows remain visible last.
- Labels a completed arm with `NaN` or infinity as `diverged`, never gives it the best-run marker, and excludes it from best-run selection.
- Treats a non-finite recent arm as explicitly worse than the early-stop threshold. This applies both to the per-run diagnostic and the decision to skip remaining arms instead of relying on comparisons with NaN.
- Preserves every result row; no diverged or failed run is filtered out.

The early-stop choice is deliberate: after at least two results, a diverged most-recent arm stops the remaining sweep even when no finite best exists yet. A run that consumed training time but produced a non-finite loss is worse than any configured finite threshold and remains visible in the final table.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Checklist

- [ ] `pytest tests/ -v` passes (the local environment intentionally lacks the optional training stack; focused results are below)
- [x] Updated relevant docs if needed (the behavior is covered by the changelog fragment and explicit terminal status)

## Validation

- TDD: the initial three ordering, display, and early-stop controls failed on current `main`; all five final #944 regressions pass.
- #944 plus existing sweep/dry-run suites: 54 passed.
- Forced Rich colour: all 5 #944 regressions pass after ANSI-normalizing the CLI assertions.
- Broader dependency-light run: 291 passed, 1 skipped; its 17 failures all import the absent optional `torch` package from unrelated trainer/checkpoint tests.
- Repository-wide Ruff, current-main changelog validation, compileall, and `git diff --check` pass.

## AI assistance

I used OpenAI Codex to help inspect the sweep lifecycle, implement the change, and draft tests. I reviewed the complete diff and ran the validation above locally.


MakazhanAlpamys:
Reviewed at `c8386148`. **The implementation is correct and nothing in `src/` needs to change.** I am blocking on two test items only, and both are small.

**The baseline is real.** On `origin/main` the captured table is the issue verbatim — the NaN arm sorts first *and* wins:

```
│ nan-run    │ nan-run    │      nan * │       1s │ completed │
│ finite-run │ finite-run │     0.5000 │       1s │ completed │
Best run: nan-run (loss: nan)
```

Two of your four new tests baseline as an `AttributeError` on the new private helper rather than a behaviour delta, which is expected for a unit test of new code; tests 3 and 4 carry the real regression evidence (`assert 0 == 2` for the missing `diverged` label, `assert 4 == 2` on `_run_single.call_count`).

**I probed the NaN edge cases directly rather than trusting the sort**, because `sorted` is stable and NaN breaks the ordering contract, so "moves NaN to the end" for one input can silently not hold for another. It holds here — the tuple key removes the NaN comparison entirely rather than relying on stability:

| case | result |
|---|---|
| NaN first / middle / last in input | identical final order in all three |
| NaN + `-inf` + `+inf` + finite | `fin, nan, ninf, pinf`; `-inf` sorts **last**, not first |
| all arms NaN | no marker, no `Best run:` line |
| loss `None` alongside a finite arm | finite first, `None` last |
| failed arm (sentinel loss `0`) | sorts last — on main it sorted **first** |
| selection path vs display path | separate predicates, both finite-gated at `:583` |

Ordinary finite input is byte-identical to main, including the `0.51` tie. And a bonus you did not claim: `_display_summary` with `final_loss=None` **crashes on main** (`TypeError: unsupported format string passed to NoneType.__format__`) and does not here.

**Mutations — two of three caught, and the third is the ask.** Each anchor matched exactly once and each mutated module was actually imported before running:

| mutation | result |
|---|---|
| `key=_loss_sort_key` -> `key=lambda r: r.get("final_loss", float("inf"))` | **caught** — `assert 322 < 258` on the ordering assertion |
| `_is_diverged_loss(final_loss): return True` removed from early stop | **caught** — `assert 5 == 2` on `_run_single.call_count` |
| `sweep.py:566` `idx == 0 and display_status == "completed"` -> `res["status"] == "completed"` | **SURVIVES — 53 passed** |

---

**Ask 1 — the surviving mutation, `sweep.py:566`.** The guard is correct; nothing reaches it. In the only test that exercises `_display_summary`, index 0 is the *finite* arm, so the marker lands in the same place either way. It matters only when every arm is non-finite:

```
# mutated                          # your code (correct)
│ a │ 1 │  nan * │ 1s │ diverged │  │ a │ 1 │  nan │ 1s │ diverged │
│ b │ 2 │    inf │ 1s │ diverged │  │ b │ 2 │  inf │ 1s │ diverged │
```

Your changelog fragment states non-finite losses are "never selected as the best run" — right now that claim rides on sort order alone. One test over an all-diverged result list, asserting no `*` and no `Best run:` line, holds it.

**Ask 2 — `tests/test_issue944_sweep_nonfinite.py:99-100` fail under colour.** Reproduced:

```
FORCE_COLOR=1 TERM=xterm-256color  ->  1 failed, 3 passed
>   assert "skipping 3 remaining run(s)" in result.output
E   '\x1b[1;33mrun\x1b[0m\x1b[1;33m(\x1b[0m\x1b[33ms\x1b[0m\x1b[1;33m)\x1b[0m'
```

`ReprHighlighter` colours the number and splits `run(s)` apart. Line 99 (`count("diverged") == 3`) survives only because `nan`/`inf` are not in the highlighter's number pattern **today**. CI does not force colour, so this is latent rather than red — but the sibling test for this very command already imports the remedy (`tests/test_issue642_sweep_dry_run.py:90-93`, `from .conftest import strip_ansi`, used by 44 files). I verified the one-line fix: routing both assertions through `strip_ansi` gives 4 passed under `FORCE_COLOR=1`. Your test 3 is already safe because `Console(record=True).export_text()` strips styles by default.

---

**Four smaller notes, none of them requests:**

- `sweep.py:564-568` and `:580-584` use different predicates for the marker and the `Best run:` line, so a sole completed arm with `final_loss=None` gets `- *` and **no** `Best run:` line. Not a regression — main crashed on that input — but they disagree.
- When every arm diverges the whole best-run block silently vanishes. A one-line "No run produced a finite loss" would finish the issue's second criterion.
- `0.0` loss now prints `0.0000` and can become `best_loss`, where main showed `-` and never selected it (falsy). Correct, and a finite-input behaviour change with no test and no changelog mention.
- Early stop can now fire with **no finite baseline** (two diverged arms, `best_loss` still `inf`), where main's `best_loss < inf` guard prevented it. That is what the issue's fourth criterion asked for, but it is a policy widening worth a clause in the fragment.

The fragment itself is correct: one file, PR-numbered, right directory, `(#944 by @taljeon in #946)` byte-for-byte the sibling convention. 214 passed across the sweep suites, ruff clean, zero removed test lines.

Push those two test changes and I will merge.

taljeon:
Addressed both test blockers in `c7ab317`:

- added an all-diverged summary regression that requires two `diverged` rows, no best marker, and no `Best run:` line
- routed both CLI assertions through the shared `strip_ansi` helper

Validation is now 5/5 for #944 under normal and forced Rich colour, and 54/54 across the combined #944, sweep, and dry-run suites. Ruff, diff, changelog, metadata, and publication checks remain clean. The PR body has the updated counts.

MakazhanAlpamys:
Re-checked at `c7ab3178`. **Both blockers closed**, and `c7ab3178` touches only the test file — `src/` is byte-identical to the head I reviewed, so nothing I verified before needs re-verifying.

**The surviving mutation is dead.** Same anchor as last time (`if idx == 0 and display_status == "completed"`, 1 occurrence), mutated to `res["status"] == "completed"`, `ast.parse` and a real import both fine:

```
FAILED tests/test_issue944_sweep_nonfinite.py::test_all_diverged_summary_has_no_best_marker_or_best_run
E  AssertionError: assert '*' not in '...'
```

Behavioural, on rendered output. Your new test is the right shape — two `diverged` rows, `count("diverged") == 2`, no `*`, no `Best run:` line — because it puts a *non*-finite arm at index 0, which is exactly the position the old test could never reach.

**ANSI closed, and I confirmed the change is load-bearing rather than incidental** by running the pre-fix file under colour as a control:

```
this head, plain                     -> 5 passed
this head, FORCE_COLOR=1             -> 5 passed
pre-fix c8386148, FORCE_COLOR=1      -> 1 failed  ('skipping 3 remaining run(s)' in '┌────...')
```

You also used the shared `tests/conftest.py:strip_ansi` rather than writing a private copy, which is the difference between this and the two tests I had to flag elsewhere today.

Both earlier mutations still die after the new commit — the sort key (`assert 322 < 258`) and the early-stop divergence branch (`assert 5 == 2` on `_run_single.call_count`). 54 passed across the three sweep suites, ruff clean, fragment unchanged and correctly credited.

Waiting only on CI, which is queued on `c7ab3178`. I will merge when it lands.

The four notes from last round stand as non-requests, since nothing in `c7ab3178` touched `src/`: the marker and `Best run:` line still use different predicates, an all-diverged run still prints no "no finite loss" line, `0.0` is now selectable as best, and early stop can fire with no finite baseline. If you want any of them, they are a separate PR.

MakazhanAlpamys:
Merged as `3e427b9b`. Thank you — this is the kind of defect that is worse than a crash, and you treated it that way.

A NaN arm sorting to the top of the summary and carrying the best-run marker means a sweep **recommends** the one outcome it exists to rule out, and it does so silently: the row looks like every other row. Nobody investigates a plausible-looking recommendation.

Two things I want on the record beyond the fix itself.

**The colour assertions go through the shared `tests/conftest.py:strip_ansi` rather than a private copy.** That is the difference between this PR and two others I had to flag today. And you did not assume the helper made the test meaningful — you showed it discriminates:

```
this head, FORCE_COLOR=1        -> 5 passed
pre-fix c8386148, FORCE_COLOR=1 -> 1 failed  ('skipping 3 remaining run(s)' in '┌────...')
```

**You left the four non-blocking notes alone once the head stopped touching `src/`.** Sweeping them into the same diff would have been easy and would have made the final review longer for no benefit. They remain open as a separate PR if you want them: the marker and the `Best run:` line still use different predicates, an all-diverged run prints no "no finite loss" line, `0.0` is now selectable as best, and early stop can fire with no finite baseline.

What I ran before merging: both earlier mutations still die on the new head — the sort key on `assert 322 < 258`, the early-stop divergence branch on `assert 5 == 2` against `_run_single.call_count`. 54 passed across the three sweep suites, ruff clean, `merge-tree` clean against current `main`, and the full required matrix green on `c7ab3178`.

Your `CONTRIBUTORS.md` entry is updated at `01c878bb` — third merge.

PR #947, opened 2026-09-13
feat(doctor): add GPU architecture and precision diagnostics

### LINKED ISSUE #829: soup doctor shows only GPU name and memory: no compute capability, no check that the torch build supports the card, no bf16/fp8/nvfp4 availability

Samearth17:
## Summary

Fixes #829 by enhancing `soup doctor` GPU diagnostics.

### Changes

* Show GPU compute capability in the GPU panel, e.g. `sm_120 (Blackwell)`.
* Detect whether the installed PyTorch build contains native/PTX support for the detected GPU architecture.
* Show a red warning when the current Torch build cannot run the detected GPU architecture.
* Add precision capability reporting for:

  * BF16
  * FP8
  * NVFP4
* Reuse the existing Soup hardware/software capability helpers for FP8 and NVFP4 instead of duplicating their detection logic.
* Update the `torchao` minimum reported by `soup doctor` from `0.4.0` to `0.7.0`, matching the highest runtime requirement currently used by the precision paths.
* Add regression tests covering GPU architecture detection, PTX fallback, unsupported Torch architectures, precision reporting, and the `torchao>=0.7.0` requirement.

### Validation

```text
41 passed, 16 warnings
```

```text
git diff --check
```

passes cleanly.

Refs #829


MakazhanAlpamys:
Reviewed on this box, which is itself Blackwell — RTX 5070 Laptop, `sm_120`, torch 2.14.0+cu130 — so I ran `soup doctor` for real rather than reasoning about it:

```
│   GPU 0: NVIDIA GeForce RTX 5070 Laptop GPU (7.9 GB) — sm_120 (Blackwell)
│
│ Precision features
│   BF16: hardware=yes
│   FP8: hardware=yes, software=no
│   NVFP4: hardware=yes, software=no
```

**The arch half is good and I want that said first.** `sm_120 (Blackwell)` is correct, the red warning does **not** fire on this healthy box, and when I patch `get_arch_list()` to drop `sm_120` it does fire. Both directions right, which is the hard part. The startup-light invariant also holds — every new helper takes `torch` as a parameter and the `fp8` / `advanced_precision` imports are function-scoped, so `torch` is absent from `sys.modules` after `--help` and `version`, and `test_cli_startup_is_light.py` is 34 passed.

---

**BLOCKING 1 — the precision table disagrees with the real gates, in exactly the case #829 exists to prevent.**

#829 says doctor must *"call the same gate functions the trainers call, so doctor and training cannot disagree."* Two rows do not:

| row | doctor's probe | the real gate | agree? |
|---|---|---|---|
| BF16 | `is_bf16_supported()` | `mixed_precision.py:41` (`cc >= 8.0`) | yes |
| FP8 hw | `fp8.is_fp8_gpu_supported()` | same symbol | yes |
| **FP8 sw** | `is_fp8_available()` — torchao **or** transformer_engine | `apply_fp8_attention` requires `_torchao_available()` **specifically** | **no** |
| NVFP4 hw | `is_blackwell_gpu()` | same symbol | yes |
| **NVFP4 sw** | `_torchao_available()` (import only) | `apply_nvfp4` *also* requires `hasattr(ao_q, "NVFP4Config")` | **no** |

Simulated the torchao everyone actually has (0.18.0, where `NVFP4Config` lives under `torchao.prototype`):

```
DOCTOR says: NVFP4 hardware=yes, software=yes
REAL GATE  : RuntimeError: torchao does not expose NVFP4Config / quantize_;
             upgrade torchao (pip install -U torchao).
```

Doctor green-lights a path that raises — and it lands on `sm_120`, the one card where the NVFP4 row is the whole point. That is #826, and this would tell users it is fine.

Same shape for FP8 on an NGC-style install (transformer_engine present, torchao absent): doctor says `software=yes`, `apply_fp8_attention` refuses. `advanced_precision.py:242-246` carries a comment documenting that exact trap by name; the PR re-introduces the confusion one layer up.

Related and worth splitting while you are there: one `FP8` row conflates two features with **different** gates. `quantization_aware: 'fp8'` goes to `apply_fp8_training`, which has **no hardware gate at all** (`fp8.py:88`), while `training.fp8_attention` needs torchao **and** `major >= 9`. So `hardware=no` on an Ada card is true for one and false for the other — and that inconsistency is already open as #835.

**BLOCKING 2 — a precision mutation survives.** `bf16_supported = bool(torch.cuda.is_bf16_supported())` -> `= True` (anchor 1 occurrence, module really imported, and I confirmed the mutant genuinely returns `BF16: (True, None)` for a fake that reports `False`):

```
tests/test_doctor.py                            -> 26 passed  (mutant survives)
+ test_v0530.py + test_cli_startup_is_light.py  -> 217 passed (still survives)
```

No test file anywhere touches `_get_precision_capabilities`, and the only BF16 assertion uses a fake that returns `True` — so it cannot distinguish "reads the GPU" from "hardcoded". There is no negative test for any row. The arch mutation, by contrast, is properly killed (2 tests), so the pattern is there — it just was not applied to the precision half.

Also tautological: `test_doctor_requires_torchao_070_for_optional_feature_support` monkeypatches `doctor.DEPS` to the value it then asserts, so reverting your `DEPS` change to `0.4.0` leaves it green. **The one source-line change to `DEPS` is untested.**

---

**Two test-hygiene items, and one of them is red today, not latent.**

Both new panel tests **fail when run in isolation**:

```
pytest tests/test_doctor.py::test_doctor_gpu_panel_warns_when_torch_lacks_gpu_arch  -> 1 failed (x3)
pytest tests/test_doctor.py -k gpu_panel                                            -> 2 failed
pytest tests/test_doctor.py                                                         -> 26 passed
```

Root cause traced: `monkeypatch.setitem(sys.modules, "torch", fake)` installs a `SimpleNamespace` with no `__spec__`, and when `transformers` has not already been imported by an earlier test in the file, its module-level `is_torch_available()` hits `find_spec` and raises `ValueError: torch.__spec__ is not set` out of the command. They pass only as a side effect of test ordering. The repo's idiom for this is at `advanced_precision.py:83-99`: give the fake an `importlib.machinery.ModuleSpec("torch", None)`.

And the warning test is **already failing at a common width**:

```
COLUMNS=80  -> 26 passed
COLUMNS=100 -> FAILED test_doctor_gpu_panel_warns_when_torch_lacks_gpu_arch
```

Rich wraps the ~125-char GPU line inside the Panel, and at 100 columns the break lands between `does not` and `include`, so the asserted phrase does not exist in the output. It passes at the default width by luck. Assert on a short fragment plus the canonical ANSI-strip + whitespace-collapse helper (`tests/test_v07302.py:57-66`) — every new assertion here is raw, and under `FORCE_COLOR=1` the BF16 one fails because the colour is *inside* the asserted string (`[green]yes[/]`).

---

**Two more before merge:**

- **No changelog fragment.** This is a user-visible feature; `changelog.d/0.75.0/947.added.md` is required. `test_issue487_changelog_fragments.py` only catches *stale* fragments, not absent ones, so CI will not tell you.
- **Please rebase.** Your base is 260 commits behind (`__version__` 0.73.3 vs 0.75.0). `merge-tree` is clean, so this is not about conflicts — it is that `main` restructured `DEPS` (#828/#854/#884) so `torch`/`transformers` moved into `EXTRA_GROUPS["train"]`, and your test replaces `doctor.DEPS` wholesale. That will need rework after the merge, and #641 was specifically titled "stop duplicating the torch floor in doctor", which is the direction your torchao floor moves against.

**A false positive worth fixing while you are in there:** `doctor.py:216` compares arch strings exactly, so an `a`-suffixed target misses —

```
H100 (9,0)  vs ['sm_90a']  -> False   # red "Torch build does not include this GPU architecture"
B200 (10,0) vs ['sm_100a'] -> False
```

A wheel built with `TORCH_CUDA_ARCH_LIST="9.0a"` would get the red alarm on a working H100. This box's wheel has no `a` entries so nothing here catches it. Strip a trailing alpha suffix before comparing.

**Smaller ones:** the red line carries no remedy, though #829 asked for the CUDA-wheel hint from the existing advisory path (`_detect_gpu_hw_without_torch_cuda` is right there, unused); `_gpu_architecture_name` returns `Unknown` for sm_60/61/**70** — V100 and P100, the cards #425/#429 explicitly support; `_format_gpu_capability` catches `(RuntimeError, AssertionError)` but not `AttributeError` unlike its sibling at `:213`, so a `torch.cuda` without `get_device_capability` raises out of the command; and the doctor floor moved to `torchao>=0.7.0` while `pyproject.toml:130` still declares `qat = ["torchao>=0.4.0"]`, which is the same one-floor-two-places problem #829 wanted removed, now in the other direction.

One thing that is **not** your fault: `test_v0530.py::TestNVFP4::test_apply_live_gated` fails on this box because it asserts `pytest.raises(match="Blackwell")`, which cannot fire on a real sm_120 card. That was fixed on `main` in `b671a6be`, which your base predates — it disappears on rebase.

The arch detection and the panel are genuinely worth having. It is the precision table I cannot take as it stands, because a `soup doctor` that is wrong about NVFP4 on a Blackwell card is worse than one that says nothing.

MakazhanAlpamys:
@Samearth17 — status, because **CI just went green on `d9e046a8` and that does not mean this is ready.** I would rather say so than have you read the tick and wait.

Two things changed since my review, neither from you:

1. **The branch now conflicts with `main`.** It did not when I reviewed — `merge-tree` was clean then. `main` has moved a long way today, including `#911` and `#884` restructuring the same region of `commands/doctor.py`.
2. **Green CI here is weaker than it looks.** Your base is 260 commits behind (`__version__` 0.73.3 against 0.75.0), so the matrix that just passed ran against a `DEPS` layout that no longer exists. `torch` and `transformers` moved into `EXTRA_GROUPS["train"]` in #828/#854/#884, and your test replaces `doctor.DEPS` wholesale — that will need rework *after* the rebase, and the rebase is what will surface it.

The review items from 2026-09-13 are unchanged and I am not restating them in full, but the one I would fix first is still the false positive, because it is the difference between a diagnostic that helps and one that alarms:

```
H100  (9,0)  vs ['sm_90a']   -> False   # red "Torch build does not include this GPU architecture"
B200 (10,0)  vs ['sm_100a']  -> False
```

`doctor.py:216` compares arch strings exactly, so a wheel built with `TORCH_CUDA_ARCH_LIST="9.0a"` gets the red alarm on a **working** H100. This box's wheel has no `a` entries, which is why nothing here catches it — strip a trailing alpha suffix before comparing.

Also still open: no changelog fragment (`changelog.d/0.75.0/947.added.md` — and note `test_issue487_changelog_fragments.py` catches *stale* fragments, not absent ones, so CI will never tell you); `_gpu_architecture_name` returns `Unknown` for sm_60/61/70, i.e. P100 and V100, the cards #425/#429 explicitly support; `_format_gpu_capability` does not catch `AttributeError` unlike its sibling at `:213`; and the doctor floor is `torchao>=0.7.0` while `pyproject.toml:130` still says `qat = ["torchao>=0.4.0"]` — the same one-floor-two-places problem #829 wanted removed, now pointing the other way.

One item **does** disappear for free on rebase: `test_v0530.py::TestNVFP4::test_apply_live_gated` fails on a real sm_120 card because it asserts `pytest.raises(match="Blackwell")`, which cannot fire there. Fixed on `main` in `b671a6be`, which your base predates.

The arch detection and the panel are worth having and I want them. The precision table is the part I cannot take as it stands, because a `soup doctor` that is wrong about NVFP4 on a Blackwell card is worse than one that says nothing about it.

Your #883 is separately red on a one-character changelog fragment — details in that thread.

MakazhanAlpamys:
@Samearth17 — reviewed `fa7b08f2`. **Eleven findings closed and three mutations newly killed — this is a large, genuine delta.** Two blocking items remain, and the second one is not something you could have seen, so let me put the good half first.

## Closed, verified individually

The `sm_90a` suffix bug (the one I said to fix first) is closed and pinned — `.rstrip("a")` on both `sm_` and `compute_`, and mutating it to `.rstrip("QQQ")` fails `test_torch_gpu_arch_supported_accepts_suffixed_sm_target` on `assert False is True`. Volta and Pascal added to `_gpu_architecture_name` with a test. `AttributeError` added to the `_format_gpu_capability` catch. The red line now carries a remedy. The tautological `DEPS` monkeypatch is gone, and mutating the floor `0.7.0 → 0.4.0` now fails `test_doctor_requires_torchao_070_for_optional_feature_support` on `assert "outdated" in result.output` — so that test drives real data now. `__spec__=ModuleSpec("torch", None)` fixed the isolation failures; both panel tests pass alone. Fragment correct and credited. Rebase clean.

**BLOCKING 2 from last time is closed**: `bf16_supported = bool(torch.cuda.is_bf16_supported())` → `True` now fails `test_get_precision_capabilities_reports_negative_bf16` on `assert (True, None) == (False, None)`.

And the behaviour is right on this Blackwell box:

```
GPU 0: NVIDIA GeForce RTX 5070 Laptop GPU (7.9 GB) — sm_120 (Blackwell)
Precision features
  BF16: hardware=yes
  FP8: hardware=yes, software=no
  NVFP4: hardware=yes, software=no
torchao │ optional │ - │ >=0.7.0 │ not installed
```

I checked that against the real gates rather than reading the table: with torchao 0.18.0's actual shape (`NVFP4Config` only under `prototype`), doctor says `software=no` **and** `apply_nvfp4` raises `torchao does not expose NVFP4Config / quantize_`. They agree. Same for FP8 with `transformer_engine` present and torchao absent. That is exactly what BLOCKING 1 asked for.

## BLOCKING 1 — the fix is correct and nothing holds it there

```
nvfp4_software = hasattr(ao_q, "NVFP4Config") and hasattr(ao_q, "quantize_")   ->   = True
  528 passed, 6 skipped

"FP8": (bool(is_fp8_gpu_supported()), torchao_available)   ->   (..., True)
  474 passed, 6 skipped
```

Both anchors counted exactly once, module genuinely re-imported, mutant confirmed live in `getsource`. With the first one live and torchao 0.18.0 simulated:

```
doctor says      : NVFP4 hardware=yes software=yes
apply_nvfp4 says : RuntimeError: torchao does not expose NVFP4Config / quantize_
```

That is precisely the doctor-disagrees-with-the-gate behaviour I blocked, restored by deleting the fix, with 528 tests green.

The cause is that `test_doctor_gpu_panel_shows_compute_capability_and_precision` installs a fake torchao **with** `NVFP4Config` present and asserts `software=yes`. There is no test for torchao-present-but-`NVFP4Config`-absent — which is the entire torchao 0.18.0 case, i.e. the only case that matters. Two negative tests close it: torchao present without `NVFP4Config`, and `transformer_engine` present with torchao absent.

This is the mirror of the shape I blocked twice on #883 ("all 51 of your tests pass with this live"). There it was the bug that survived; here it is the fix.

## BLOCKING 2 — a collision with your own #883, and no conflict marker warns you

Your new `_detect_gpu_arch_mismatch_advisory()` calls the tag function unguarded:

```python
wheel = _torch_cuda_wheel_tag(driver_cuda)
index_url = f"https://download.pytorch.org/whl/{wheel}"
```

On `main` that returns `str`. **#883 widens it to `str | None`** — correctly. I composed the two and measured:

| driver | #947 alone | #883 + #947 |
|---|---|---|
| (11, 7) | `whl/cu118` | **`whl/None`** |
| (11, 0) | `whl/cu118` | **`whl/None`** |
| (11, 8) | `whl/cu118` | `whl/cu118` |

They merge into each other with **no conflict**, and `tests/test_doctor.py` on the composed tree is **46 passed**. Your only test for this function uses driver `(13,2)` → `cu132`, which passes either way.

I am taking **#883 first**, because it owns the `str | None` contract and both of its own sites are already guarded. So this PR needs: a rebase, a guard on the tag (not on the driver) at your new site, and a test at a driver below CUDA 11.8 asserting no `whl/None` reaches the advisory — the same test #883 was required to write. Nothing about your logic is wrong; the contract moved underneath you.

This is the **fourth** appearance of this defect and the second time it has fired across two PRs that both looked clean. Worth naming rather than just fixing.

## Three smaller items

3. **The width assertion still fails at `COLUMNS=60`, `100` and `120`.** You applied the strip-and-collapse helper but not the other half of the advice — the break lands on a Panel border, which no amount of whitespace collapsing removes:
   ```
   '│ ... — sm_120 (Blackwell) Torch build does not  │'
   '│ include this GPU architecture Try reinstalling ... │'
   ```
   Assert on a short fragment that cannot wrap.

4. **Five raw output assertions remain** — `tests/test_doctor.py:239-243` and `:323`. Only line 325 was converted. `doctor.py:583` builds `hardware={hw_status}` where `hw_status = "[green]yes[/]"`, so Rich puts SGR *inside* the asserted string, and under `FORCE_COLOR=1` `BF16: hardware=yes` fails. CI sets neither `FORCE_COLOR` nor `COLUMNS`, so these are latent rather than red today — but this exact class has turned CI red four times in this repo's history, which is why there is a repo-wide scanner for it.

5. **No `docs/` update** for a user-visible feature — the capability display and the whole Precision-features block are new.

## One thing worth a sentence in the PR body

Raising `qat = ["torchao>=0.7.0"]` lifts the install floor for **every** `[qat]` user in order to satisfy a doctor advisory row, while `utils/qat.py` itself only needs `quantize_` / `Int8WeightOnlyConfig` and `export.py` says `torchao>=0.5`. It is defensible and I am not asking you to change it — just say so, so the next person does not have to work out why.

The FP8-row conflation (#835) stays untouched and stays not-a-blocker.

---

Items 1 and 2 are the merge gate. Ping me when they are pushed and I will review only the delta — CI is pending on this head anyway, and I approved your workflow runs just now so the checks will actually execute this time.

MakazhanAlpamys:
Checking in — head unchanged since the 2026-09-14 review, so this is not a re-review. But **one of the two blocking items has stopped being hypothetical**, and you should know that before you pick this back up.

## `whl/None` is now LIVE on the merged tree

When I raised it, #883 was still open and I described it as a composition. **#883 has since merged**, so `_torch_cuda_wheel_tag` returns `str | None` on `main` today. Yours is the only unguarded call site of the three:

```
doctor.py:255  guarded   url = f"..." if tag else None
doctor.py:777  guarded   if wheel is None: return ...
doctor.py:727  UNGUARDED index_url = f"https://download.pytorch.org/whl/{wheel}"
```

Executed on the merged tree:

```
(13, 4) tag=cu132 -> --index-url https://download.pytorch.org/whl/cu132
(11, 8) tag=cu118 -> --index-url https://download.pytorch.org/whl/cu118
(11, 7) tag=None  -> --index-url https://download.pytorch.org/whl/None    <- user-facing
(11, 0) tag=None  -> --index-url https://download.pytorch.org/whl/None
```

The threshold is driver CUDA **< 11.8**. (12.1-12.3 is fine — `cu121` is in `_TORCH_CUDA_WHEELS` but not in `_SUPPORTED_TORCH_CUDA_WHEELS`, so it falls through to `cu118`.)

**And `tests/test_doctor.py` is 50 passed with that live**, because your only test for the function uses driver `(13,2)`. That is the whole finding: a command that hands a user a copy-pasteable `pip install` line would hand them a broken URL, and nothing goes red.

## The other blocking item: both mutants still survive

On the merged tree, with a positive control to prove the suite discriminates:

```
doctor.py:540  nvfp4_software = hasattr(ao_q,"NVFP4Config") and ... -> = True or ...
   anchor x1, import OK, mutant live   ->  50 passed   SURVIVES

doctor.py:550  "FP8": (bool(is_fp8_gpu_supported()), torchao_available) -> ..., True)
   anchor x1, import OK                ->  208 passed  SURVIVES

CONTROL doctor.py:497  .rstrip("a") -> .rstrip("QQ")
   anchor x1, import OK  ->  1 failed: test_torch_gpu_arch_supported_accepts_suffixed_sm_target
```

The control dying is what makes the two survivors meaningful rather than a setup problem. Two negative tests close it: **torchao present without `NVFP4Config`** (which is the entire torchao 0.18.0 case) and **`transformer_engine` present with torchao absent**.

## What is good, and I ran it rather than read it

`soup doctor` on the merged tree, this box (RTX 5070 / sm_120 / torchao absent):

```
GPU 0: NVIDIA GeForce RTX 5070 Laptop GPU (7.9 GB) - sm_120 (Blackwell)

Precision features
  BF16: hardware=yes
  FP8: hardware=yes, software=no
  NVFP4: hardware=yes, software=no
```

Correct, and it agrees with the real gates here. Worth naming that the `software=no` rows are reached through the outer `torchao_available` guard on this box — not through the `NVFP4Config` check the mutation shows is untested. So the panel being right today does not mean that branch is covered.

Branch 45 passed / merged tree 50 passed, no failures either way. **Zero removed test lines** across 266 additions. Fragment correct and credits you.

Three smaller ones from the original review, all still open and all minor next to the two above: the width fragility at `COLUMNS=100`/`120` (the break lands on a Panel border, which collapsing cannot remove), the five raw `in out` assertions at `tests/test_doctor.py:239-243` that go red under `FORCE_COLOR=1` because `doctor.py:583` builds `hardware=[green]yes[/]`, and no `docs/` mention of the new Precision-features block.

Owed, in order: rebase, guard the tag at `:727` plus a test below CUDA 11.8, then the two negative precision tests. No deadline from me.

MakazhanAlpamys:
Reviewed `8bd9f861`. **Both mutations I named are dead, and the way you killed them is better than what I asked for** — but one of them moved rather than died, and blocker 1 is untouched. Two small things close this.

## The extraction is the right call, and I want to say why before the blockers

You replaced doctor's inline torchao probe with a shared `is_nvfp4_software_supported()` that `apply_nvfp4` also calls. I asked for a test; you removed the possibility of disagreement instead. Doctor and the runtime now read **one symbol**, so they cannot drift — that is a structurally better answer than two tests pinning two copies.

Both named mutations now die, module confirmed live under each (not just parsed):

```
nvfp4_software = bool(is_nvfp4_software_supported()) -> = True
    1 failed   assert (True, True) == (True, False)
"FP8": (bool(is_fp8_gpu_supported()), torchao_available) -> ..., True)
    1 failed   At index 1 diff: True != False
```

Positive control (`.rstrip("a")` -> `.rstrip("QQ")`) still dies, so the bench discriminates on this head rather than being inert.

And **zero removed test lines across 371 additions** — `git diff origin/main...8bd9f861 -- tests/` is `371  0`. The one test you rewrote is a **strengthening**: it now parses `[project.optional-dependencies] qat` out of `pyproject.toml` and asserts `doctor.DEPS` matches it in both directions, instead of pinning the literal `">=0.7.0"`. That closes the one-floor-two-places problem properly rather than moving it.

## Blocker 1 — `whl/None` is unchanged, and I executed it end-to-end

This is byte-identical between your previous head and this one. On the merged tree, driving the real CLI with the driver forced below CUDA 11.8:

```
GPU 0: NVIDIA GeForce RTX 5070 Laptop GPU (7.9 GB) - sm_120 (Blackwell)
Torch build does not include this GPU architecture
Reinstall a CUDA-enabled PyTorch build for your driver:
  `pip install torch --index-url https://download.pytorch.org/whl/None`
```

Across the boundary:

| driver CUDA | tag | advisory |
|---|---|---|
| (13,4) / (12,8) / (11,8) | cu132 / cu128 / cu118 | correct |
| **(11,7) / (11,0) / (10,2)** | **None** | **`.../whl/None`** |
| unreadable | None | correct fallback |

Three call sites, re-confirmed by `inspect.getsource` on the merged tree:

```
doctor.py:255  GUARDED    tag = _torch_cuda_wheel_tag(driver) if driver is not None and torch_missing else None
doctor.py:718  UNGUARDED  wheel = _torch_cuda_wheel_tag(driver_cuda)
               :719       index_url = f"https://download.pytorch.org/whl/{wheel}"
doctor.py:768  GUARDED    if wheel is None: return "...Run `nvidia-smi` and install..."
```

`:718` is the odd one out of three. The line number moved from `:727` only because `main` shifted above it.

**And `tests/test_doctor.py` is 52 passed with this live.** The only test of that function uses driver `(13, 2)`. This is the shape that matters to me: a command whose entire job is to tell a user what to run hands them a broken `pip install` line, and the suite is green.

**The test to mirror already exists in your file.** `tests/test_doctor.py:587-610`, `test_doctor_missing_train_extra_with_unsupported_driver_falls_back_to_extra`, driver `(11, 7)`, asserting `"whl/None" not in out`. It needs a twin pointed at the advisory path. Guard the **tag**, not the driver — `:255` already shows the pattern.

## Blocker 2 — the mutation did not die, it relocated

Your two new tests monkeypatch `advanced_precision.is_nvfp4_software_supported` and `_torchao_available`. That pins doctor's **plumbing**. The **predicate** is now one module further away and completely uncovered:

```
advanced_precision.py:311
  return hasattr(ao_q, "NVFP4Config") and hasattr(ao_q, "quantize_")   ->   return True

  480 passed, 6 skipped        <- SURVIVES
```

That is five test files, including `test_doctor`, `test_v0530`, `test_v0531_142`, `test_v07121`. And because `apply_nvfp4` now delegates to the same helper, deleting the predicate breaks **both sides at once**. Measured with that mutant live and torchao 0.18.0 simulated (`quantize_` present, `NVFP4Config` only under `prototype`, which is the real shipped layout):

```
is_nvfp4_software_supported() -> True
DOCTOR says NVFP4             -> (True, True)      # software=yes, wrongly
apply_nvfp4()                 -> gate PASSED
```

So the #826 mis-report comes back **and** the runtime refusal disappears, with 480 tests green. The consolidation is what makes this one mutation reach both, which is the flip side of the design win above — it is a good trade only if the single point is tested.

One test closes it, and your fixture already exists at `tests/test_doctor.py:222-232` — it needs the negative twin: build a fake `torchao.quantization` with `quantize_` and **no** `NVFP4Config`, assert `is_nvfp4_software_supported() is False`. That is the torchao 0.18.0 case verbatim.

## Still open from last round, none blocking

1. **`docs/` is still empty** — `git diff origin/main...8bd9f861 -- docs/ README.md` returns nothing. `soup doctor` is documented at `docs/commands.md:222` and `docs/backends-and-ops.md:1158-1162`, and the latter already quotes the CPU-wheel advisory, so it is the natural home for the new capability line and the `Precision features` block. (Note: `docs/` only — please do **not** touch `README.md`, because #852 makes every README edit oblige a `README.tr.md` re-translation and re-stamp. I have now let two contributors walk into that this week without warning them, so I am saying it up front.)
2. **Six raw `result.output` assertions.** `tests/test_doctor.py:236` is `out = result.output` feeding five `in out` at `:239-243`, plus `:395`. Under `FORCE_COLOR=1 TERM=xterm-256color`: `1 failed, 51 passed`, `assert 'BF16: hardware=yes' in '\x1b[1mSoup Doctor\x1b[0m - checking...'`. `doctor.py:583` builds `hardware=[green]yes[/]`, so the SGR lands inside the asserted string. Your file's own `_strip_ansi` is at `:19` and ~25 other assertions already use it — these six were missed. Forced colour is *stricter* than real CI, so treat it as a lead; but this class has turned this repo's CI red four times.
3. **`COLUMNS=100` and `120`**: `1 failed` — `assert 'Torch build does not include this GPU architecture' in '... ┌──── ... whl/cu132 ...'`. That is a Panel border breaking the string, which whitespace-collapsing cannot repair. `COLUMNS=80` and the default both pass.
4. **One sentence in the PR body about the `[qat]` floor.** `torchao>=0.4.0 -> >=0.7.0` lifts the install floor for every `[qat]` user, not only the precision paths. The body half-explains it now; say the consequence outright.

The FP8-row conflation (#835) stays untouched and stays not-a-blocker.

## Where this stands

Eleven findings closed last round, two mutations killed this round, zero tests deleted across 371 additions, and a floor test that now derives instead of asserting a literal. You are doing this carefully.

What is left is one guard plus one test each. The `whl/None` line is the one I care about — it is the only defect here that reaches a user directly, and it reaches them at the exact moment their environment is already broken.

MakazhanAlpamys (changes requested):
One of the two closing items is done and verified. The other is unchanged, and I can show you exactly why it is not closed rather than just asserting it.

## Blocker 1 — CLOSED

The `wheel is None` guard at `doctor.py:717-722` is in, and I checked it by execution across the boundary rather than by reading the diff:

```
EVIDENCE: soup_cli.__file__ under the scratch tree, peft 0.20.0, merged against current main
driver (13,4) -> whl/cu132     (11,7)      -> fallback text, no "whl/None"
driver (12,8) -> whl/cu128     (11,0)      -> fallback text
driver (11,8) -> whl/cu118     (10,2)      -> fallback text
                               unreadable  -> fallback text
```

Mutation: delete the guard → `test_gpu_arch_mismatch_advisory_falls_back_for_unsupported_driver` fails, 52 passed. So the test is real and it bites. That was the blocker.

## Blocker 2 — UNCHANGED, and the mutation still survives

This is the one I described as "moved rather than died". The predicate now lives at `advanced_precision.py:311` as `is_nvfp4_software_supported`, and nothing calls it:

```
EVIDENCE: soup_cli.__file__ under the scratch tree, peft 0.20.0
mutation: advanced_precision.py:312
  return hasattr(ao_q, "NVFP4Config") and hasattr(ao_q, "quantize_")   ->   return True
(mutant confirmed live via inspect.getsource)

pytest tests/test_doctor.py tests/test_v0530.py tests/test_v0531_142.py \
       tests/test_v07121.py tests/test_cli_startup_is_light.py
baseline: 483 passed, 6 skipped
mutant  : 483 passed, 6 skipped          <- byte-identical
```

`grep -rn is_nvfp4_software_supported src/ tests/` finds it at `doctor.py:529/539`, `advanced_precision.py:305/345`, and **two monkeypatches** at `tests/test_doctor.py:287,323`. Monkeypatching the symbol pins doctor's plumbing — that the panel renders what the predicate returns — but nothing exercises the predicate itself.

**What closes it:** one test that calls the real `is_nvfp4_software_supported()` against a fake `torchao.quantization` exposing `quantize_` and **no** `NVFP4Config` — the actual torchao 0.18.0 layout, where NVFP4 lives under `torchao.prototype` — asserting `is False`. That is the #826 shape, so the test also documents why the predicate exists.

Worth knowing: on my box both `software=no` rows are reached through the outer `_torchao_available()` guard, because torchao is not installed. So the panel being correct here proves nothing about this predicate. That is precisely the gap.

## I ran it on real Blackwell hardware, and every row is right

```
soup doctor  (RTX 5070 Laptop, torch 2.14.0+cu130, cap (12,0),
              arch_list ['sm_75','sm_80','sm_86','sm_90','sm_100','sm_120'])

GPU 0: NVIDIA GeForce RTX 5070 Laptop GPU (7.9 GB) - sm_120 (Blackwell)
Precision features
  BF16:  hardware=yes
  FP8:   hardware=yes, software=no
  NVFP4: hardware=yes, software=no
```

Checked each: `sm_120` is Blackwell and is in `get_arch_list()`, so no arch warning is correct; `FP8 hardware=yes` agrees with the real gate at `utils/fp8.py:55` (`major >= 9`, which sm_120 passes) rather than merely looking plausible; `NVFP4 hardware=yes` is right for Blackwell; both `software=no` because torchao is absent; 7.9 GB against the card's 8151 MiB. This is the first Blackwell card this project has had, and your panel is the first thing in the repo to describe it correctly.

Architecture detection is properly covered, too — two mutations, both killed: making `_gpu_architecture_name` misreport Blackwell as Hopper fails 2 tests; breaking the `sm_90a` suffix handling fails `test_torch_gpu_arch_supported_accepts_suffixed_sm_target`. The gap is entirely in the precision predicate.

## Four smaller ones still open

3. **`COLUMNS=120` still fails.** 80 passes (53), 100 passes now (53, you fixed it), 120 gives 1 failed — `test_doctor_gpu_panel_warns_when_torch_lacks_gpu_arch`, `assert 'GPU architecture' in ...`. The Panel border breaks the phrase at that width. Assert a fragment short enough not to wrap at any width. (`COLUMNS=60` gives 7 more failures, but all 7 are pre-existing `test_doctor_*train*` tests you did not touch — not yours.)
4. **Raw-output assertions.** Under `FORCE_COLOR=1 TERM=xterm-256color`: 1 failed, `test_doctor_gpu_panel_shows_compute_capability_and_precision` — `assert 'BF16: hardware=yes' in '\x1b[1mSoup Doctor\x1b[0m ...'`, because `doctor.py:583` builds `hardware=[green]yes[/]`. `tests/test_doctor.py:252-259` and `:411` still read `result.output` directly. This repo has turned CI red four times on exactly this, on Linux and macOS while Windows passed, so please route them through the strip helper.
5. **`docs/` is still empty** — `git diff origin/main...<head> -- docs/` returns nothing. `docs/backends-and-ops.md` already quotes the CPU-wheel advisory, so it is the natural home. Do **not** touch `README.md`: guard #852 would then oblige a `README.tr.md` re-translation and re-stamp, which is a trap I have walked two people into this month.
6. New nit: the `wheel is None` fallback carries no command, while its sibling at `doctor.py:768` says "Run `nvidia-smi` and install…". Same hint would help.

Blocker 2 is the only thing standing between this and a merge. Everything else on this list is small.

MakazhanAlpamys:
Thank you — blocker 2 is closed properly. You extracted `is_nvfp4_software_supported()` so that `soup doctor` and `apply_nvfp4` share one gate, and you replaced the monkeypatched predicate with a fake torchao module. The fake module is what makes the mutation from last round fail now: `return True` and `return False` each fail 2 tests. Moving every doctor output assertion through `strip_ansi` and adding the `sm_90a` / `sm_100a` suffix test were both good calls.

Review on `3cae1a29`, on the tree merged with `main` at `3dffdae9`:
- `test_doctor`, `test_v0530`, `test_v07121`, `test_fp8_recipe`, `test_v0531_142`, `test_issue889`: 488 passed, 6 skipped. `test_doctor` under `FORCE_COLOR=1`: 55 passed. ruff and F821 clean.
- Mutations: 8 of 9 killed, covering both predicate returns, the `NVFP4Config` clause, the arch-mismatch branch, the PTX comparison, the `sm_` suffix strip, and the BF16 and FP8 software rows.
- Real RTX 5070 Laptop (sm_120, torch 2.14.0+cu130, no torchao): `sm_120 (Blackwell)`, BF16 hardware=yes, FP8 hardware=yes / software=no, NVFP4 hardware=yes / software=no, and no arch warning. Every row is correct for this box.

**One small blocker: the `quantize_` half of the predicate is untested.** Dropping `and hasattr(ao_q, "quantize_")` survives (488 passed). `test_nvfp4_missing_quantize_friendly` does not catch it: without the guard, `apply_nvfp4` raises its "PARTIALLY quantised" error, whose text contains the `AttributeError` "...no attribute 'quantize_'", so `match="quantize_"` still matches. Add the mirror of your new test: a fake `torchao.quantization` with `NVFP4Config` but no `quantize_`, asserting `is_nvfp4_software_supported() is False`.

**One scope change: please drop the `[qat]` floor bump (0.4.0 -> 0.7.0)** from `pyproject.toml` and doctor's DEPS table. #1066 owns the torchao floor. It moves it to 0.18.0 as a declared break, and adds a ratchet that fails when any source file states its own floor. Two PRs moving the same floor would conflict, and a literal `0.7.0` would trip that ratchet. Doctor should read `TORCHAO_MIN_VERSION` once #1066 lands. Until then, leave the DEPS row as it is on `main`.

Nit: your own test narrowed `"GPU architecture"` to `"architecture"`. `"Torch build"` still pins the warning, so it is harmless, but the longer string was the better pin.

MakazhanAlpamys (changes requested):
Both technical items are closed, and I verified each by execution rather than by reading. One line left.

```
M3 advanced_precision.py:315  drop the hasattr(ao_q,"quantize_") half  -> 1 failed  KILLED
     (the mutation that SURVIVED at 3cae1a29; killed by the fake-torchao test — the mirror I asked for)
M5 same line, drop the NVFP4Config half instead                        -> 2 failed  KILLED
     so both halves of the predicate are pinned independently
M4 delete the whole `if wheel is None:` fallback                       -> 1 failed  KILLED
     the wheel-tag guard survived the rebase; driver (13,4)->whl/cu132, (11,8)->whl/cu118,
     (11,7)/(11,0)/(10,2)/unreadable -> the fallback text, and no whl/None anywhere
479 passed / 6 skipped; 513 with the card visible; 55 passed at COLUMNS 80/100/120 and under
FORCE_COLOR=1; 17 fail against main's src. ruff clean.
```

The `[qat]` scope ask is honoured — `pyproject.toml` is not in the diff at all and `doctor.DEPS` is
byte-identical to `main`, so #1066 keeps sole ownership of that floor. Zero raw-output assertions remain in
`tests/test_doctor.py`; the 15 removed lines are the same assertions re-routed through the shared
`strip_ansi`. The `COLUMNS=120` failure from last round is gone (the 7 failures at `COLUMNS=60` are
pre-existing `doctor_*train*` tests this PR does not touch).

**And the real output on this Blackwell card, cross-checked gate by gate rather than eyeballed:**

```
GPU 0: NVIDIA GeForce RTX 5070 Laptop GPU (7.9 GB) - sm_120 (Blackwell)
Precision features
  BF16: hardware=yes
  FP8: hardware=yes, software=no
  NVFP4: hardware=yes, software=no
```

`get_device_capability()` (12,0); `get_arch_list()` contains `sm_120`, so printing **no** arch warning is
correct; `is_fp8_gpu_supported()` True (the `major >= 9` gate sm_120 passes); `is_blackwell_gpu()` True;
`_torchao_available()` False; `is_nvfp4_software_supported()` False. Every row is right.

## The one thing to fix

`changelog.d/0.75.0/947.added.md` still ends "...and the TorchAO floor is synchronized at `0.7.0`." That is
the only surviving `0.7.0` in the whole diff, and the floor bump was removed at my request — so
`assemble_changelog.py` would fold a sentence describing a change the tree does not contain straight into a
public CHANGELOG entry. Drop the clause.

## Two notes

1. **Nit:** `apply_nvfp4`'s missing-torchao error went from naming the install command with a pinned version
   to "NVFP4 requires a compatible torchao installation." Dropping the pin is right; dropping the command is
   a small loss in actionability. Naming `pip install torchao` without a version keeps both.
2. **Ordering against #1066, and it is not merely textual.** `merge-tree` against `main` is clean, but
   against #1066 it conflicts in `utils/advanced_precision.py`. #1066 rewrites `apply_nvfp4` to resolve
   **`NVFP4TrainingConfig`** through `utils/torchao_compat.py`, on the ground that `NVFP4Config` is the
   post-training weight-quant class rather than the training one. So once #1066 lands, this PR's
   `is_nvfp4_software_supported()` would report the software gate for the **wrong class**. Whichever merges
   second rebases; if that is this PR, key the predicate on `torchao_compat`. #1066 currently has its own
   blocker (committed conflict markers in a docs page), so I expect this one to be ready first — in which
   case #1066 absorbs it.

Fix the fragment line and this is merge-ready.

MakazhanAlpamys:
Heads-up that affects this PR mechanically, not its content: the changelog fragment baseline is moving from `changelog.d/0.75.0/` to `changelog.d/0.75.1/`.

Why: v0.75.1 shipped to PyPI while `main`'s CHANGELOG still stopped at 0.75.0, and four pending fragments described fixes that had already gone out in it — left alone, the next release would have announced them as new. Adding the 0.75.1 heading is the fix, and `validate_fragments()` requires the fragment directory to be named for the newest heading, so every pending fragment moves with it.

What it means for you: once that lands, this PR's changelog check goes red with

> `changelog.d/0.75.0/NNNN.category.md: fragment baseline is 0.75.0, but the latest release is 0.75.1; rebase and move the fragment before merging`

and the fix is one line on your next push:

```
git mv changelog.d/0.75.0/NNNN.category.md changelog.d/0.75.1/
```

Nothing about your diff is wrong and no review finding changes. I am telling you rather than letting a red X be the first you hear of it.

MakazhanAlpamys:
**A blocking finding that did not exist when I last reviewed this, and it is the reason #1066 had to go
first.** #1066 merged a few minutes ago as `55f9726d`, and reviewing it turned up a defect in this PR that I
measured against **real torchao 0.18.0**:

```
has NVFP4Config: False
has quantize_  : True
=> is_nvfp4_software_supported() -> False
while torchao.prototype.moe_training.nvfp4_training.nvfp4_training.NVFP4TrainingConfig resolves fine
```

Your new `is_nvfp4_software_supported()` keys on `hasattr(ao_q, "NVFP4Config")`. That name is **not exported
by the current torchao release** — it is precisely what #826 was filed about. So `soup doctor` would report
`NVFP4: software=no` for every user with a correctly installed, up-to-date torchao, forever, in 137 lines of
brand-new diagnostic surface whose entire purpose is telling the truth about the environment.

That is #826's defect recreated verbatim one layer out, which is the single most valuable thing this review
found and I would not have found it without #1066 landing first.

**What to do:** rebase onto current `main` and route the predicate through the helper that now exists —
`resolve_torchao_class("NVFP4Training")` from `utils/torchao_compat.py`. It imports the root package first and
walks a declared candidate list, so it answers correctly whether the class lives at the public path or under
`torchao.prototype`. There is also a textual conflict to resolve in `src/soup_cli/utils/advanced_precision.py`
(#1066 deletes the exact `from torchao import quantization as ao_q` / `hasattr` block your PR edits), so the
rebase is required regardless.

Everything from my previous review still stands and is unchanged by this:

- both earlier technical items are **closed** — the `quantize_` half of the predicate is now killed by its own
  mutation, and the wheel-tag guard survived the rebase (no `whl/None` across six driver values);
- the `[qat]` floor is correctly **not** in your diff, which is what kept #1066's ownership of it clean;
- the real `soup doctor` output on this Blackwell card is right on every row, each gate cross-checked;
- and the one line I asked for is still open: `changelog.d/.../947.added.md` still ends "...and the TorchAO
  floor is synchronized at `0.7.0`", the only surviving `0.7.0` in the diff, describing a change the tree does
  not contain. **Note the fragment baseline also moved today** — fragments now live in `changelog.d/0.75.1/`,
  so `git mv` yours across in the same push.

To be explicit about what is not your fault: the `hasattr` spelling was the reasonable one when you wrote it,
and the class only became resolvable-by-helper this afternoon. But shipping it now would put a false "NVFP4
unsupported" in front of every user, so it has to be the rebase.

MakazhanAlpamys (changes requested):
The code is right, and I proved it against a faithful torchao 0.18.0 layout rather than by reading the diff. One thing stands between this and a merge, and it is test coverage for the fix you just made.

**First, what is closed.** `is_nvfp4_software_supported()` now resolves `resolve_torchao_class("NVFP4Training")` and `("quantize_")` — the identical two calls `apply_nvfp4` makes — so doctor and the real gate cannot disagree any more. Built on disk: `torchao.quantization` exporting `quantize_` and no `NVFP4Config`, with `NVFP4TrainingConfig` living at `torchao.prototype.moe_training.nvfp4_training.nvfp4_training`, i.e. what a correct install actually looks like:

```
hasattr(q,'NVFP4Config') = False    hasattr(q,'quantize_') = True
resolve('NVFP4Training') -> <class '...nvfp4_training.NVFP4TrainingConfig'>
is_nvfp4_software_supported() = True
doctor rows: {'BF16': (True, None), 'FP8': (True, True), 'NVFP4': (True, True)}
apply_nvfp4() GATE PASSED; linears=1
```

The false "NVFP4 unsupported" is gone. The `[qat]` floor is correctly absent from the diff (`doctor.DEPS` is byte-identical to `main`, so #1066 keeps sole ownership), the `0.7.0` clause is gone from the fragment, every raw-output assertion is routed through the shared `strip_ansi`, and `docs/backends-and-ops.md:543-552` documents the new panel. 464 passed with the GPU visible, 656 across the blast radius, `FORCE_COLOR=1` clean, 55 passed at each of `COLUMNS=80/100/120/200` — last round's `COLUMNS=120` failure is gone, and the seven `COLUMNS=60` failures reproduce identically on `main`, so they are not yours.

**BLOCKING — nothing in the suite holds that fix. Reverting to the exact defect I blocked on this morning leaves 422 tests green.**

```
M1  advanced_precision.py:312-318  restore the pre-#1066 predicate
      (from torchao import quantization as ao_q;
       return hasattr(ao_q,"NVFP4Config") and hasattr(ao_q,"quantize_"))   -> 422 passed  SURVIVED
M2  :315  resolve_torchao_class("NVFP4Training") -> ("NVFP4")              -> 422 passed  SURVIVED
M3  :316  delete resolve_torchao_class("quantize_")                        -> 422 passed  SURVIVED
M5  doctor.py:497  .rstrip("a") -> .rstrip("QQ")   [positive control]      -> 2 failed    KILLED
```

Under M1, on the same faithful torchao layout: `is_nvfp4_software_supported() = False`, doctor prints NVFP4 software **no**, and `apply_nvfp4()` runs anyway — my Round-N finding, restored verbatim, user-visible on every correctly-installed torchao, with the whole bench green. The control (M5) dies, so the bench discriminates; the gap is real.

M3 is the second half and it is a **regression in coverage**: that mutation was KILLED at the 2026-09-21 round and survives now. The cause is that the two tests named for the predicate no longer exercise it — `test_is_nvfp4_software_supported_requires_quantize` (`tests/test_doctor.py:339-352`) still builds a fake exporting `NVFP4Config`, a name the predicate no longer looks at, so it returns False because `NVFP4TrainingConfig` cannot resolve. The assertion holds for the wrong reason and the test name is no longer true. Same for `test_is_nvfp4_software_supported_requires_nvfp4_config` (`:324-336`). And the only test that could pin a `True` return, `test_doctor_gpu_panel_shows_compute_capability_and_precision`, now monkeypatches `resolve_torchao_class` to `lambda key: object` (`:253-256`), so the positive path never runs the real resolver.

**The fixture you need already exists.** `tests/test_v07121.py:301-326` has `_fake_torchao(monkeypatch, record, with_nvfp4=...)`, which builds exactly this module tree including the `torchao.prototype.moe_training.nvfp4_training.nvfp4_training` chain. Two tests against it close the whole thing:

- `with_nvfp4=True` → `is_nvfp4_software_supported() is True` (kills M1 and M2)
- the same tree with `quantize_` removed from `torchao.quantization` → `is False` (kills M3)

and re-point `:324` / `:339` at `NVFP4TrainingConfig` so their names describe what they test.

**NON-BLOCKING, but new in this PR: a broken-but-importable torchao now crashes `soup doctor` with a traceback.** `_torchao_available()` uses `find_spec`, so it returns True without executing the module; your predicate is then the first thing to import it, and it catches only `RuntimeError` while `_check_gpu` catches only `ImportError`. Measured with a `torchao.py` whose body raises `ValueError`:

```
_torchao_available() = True
$ soup doctor  ->  Rich traceback, ValueError: torchao C extension mismatch, exit=1
```

`main` never touches torchao inside `_check_gpu`, so this exposure starts here — and it lands on the one command people run *because* their environment is broken. `except Exception` in the boolean probe (it is a probe, not a gate) is the whole fix. `apply_nvfp4` has the same exposure on `main`, but a traceback in a training path is defensible; in a diagnostic it is not.

**Two small things while you are in there:** the PR body still announces the doctor `torchao` floor moving `0.4.0` → `0.7.0` and a test for the `torchao>=0.7.0` requirement, neither of which exists in the tree any more — same class of stale claim as the fragment sentence, one layer out, and the natural place to add the sentence about `[qat]` (which can now just say the floor is not in this diff because #1066 owns it). And `tests/test_v0530.py`'s three-line change is net-zero churn from three commits that changed a `match=` string and changed it back; droppable.

For the record, the FP8 *hardware* row still reads `is_fp8_gpu_supported()` (`major >= 9`) while the real refusal is the recipe-dependent `fp8.fp8_training_supported(recipe)`. That is #835, unchanged, and still not something I am asking you to solve here.

MakazhanAlpamys (changes requested):
Review of `f5064321`.

## What is closed

The blocker from last round is closed, and I checked it by breaking the predicate rather than by reading the new tests. On the tree merged with current `main` (clean merge), across 15 test files (758 passed, 13 skipped at baseline):

```
M1  restore the exact pre-#1066 predicate
    (hasattr(ao_q, "NVFP4Config") and hasattr(ao_q, "quantize_"))   -> 1 failed  KILLED  (was 422 passed)
M2  resolve_torchao_class("NVFP4Training") -> ("NVFP4")               -> 1 failed  KILLED  (was surviving)
M3  delete resolve_torchao_class("quantize_")                         -> 1 failed  KILLED  (was surviving)
```

Re-pointing the tests at `_install_fake_torchao_quant` was the right call: the positive test now runs the real resolver against the real module layout, and the test names finally say what they test. Nine more mutations also fail at least one test: the predicate's final `return True`, the three precision rows, the arch-warning branch, the `sm_` suffix strip, the PTX comparison, the `wheel is None` fallback and the Blackwell name. The one that survives is item 2 below. Your 17 new tests all fail against `main`'s code (39 pre-existing ones pass). `FORCE_COLOR=1` and `COLUMNS=80/100/120/200` all give 56 passed. The 7 failures at `COLUMNS=60` and the one under `CUDA_VISIBLE_DEVICES=-1` fail with the same test IDs on clean `main`, so they are not yours. ruff is clean.

I also ran it on real hardware, including the real package this time. On this RTX 5070 Laptop (sm_120, torch 2.14.0+cu130) without torchao: `sm_120 (Blackwell)`, BF16 yes, FP8 yes/no, NVFP4 yes/no, no arch warning. Then I installed the real torchao 0.18.0 wheel into a throwaway directory. `torchao.quantization` has no `NVFP4Config`, `NVFP4TrainingConfig` resolves from `torchao.prototype.moe_training...`, and doctor prints `NVFP4: hardware=yes, software=yes`. `apply_nvfp4()` passes the same gate and swaps both `nn.Linear` layers for `NVFP4Linear`. So doctor and the runtime agree, which is what #829 asked for. This is the first time that row has been exercised on real hardware with the real package.

## Blocking

**1. The architecture check raises a false red alarm on every Ada card, and on other cards that are one minor version newer than a listed target.**

`_torch_gpu_arch_supported` (`doctor.py:489-512`) accepts only an exact `sm_{major}{minor}` or a lower `compute_` PTX entry. But SASS is binary-compatible within a major version (code built for X.y runs on X.z when z >= y), and torch relies on that itself. `torch.cuda._check_cubins` says "NVIDIA GPU compute architectures are backward compatible within major version". The cu130 wheel on this box lists `['sm_75','sm_80','sm_86','sm_90','sm_100','sm_120']`, with no sm_89 and no PTX. I checked each capability against that exact list:

```
sm_89  RTX 4090 / L4 / L40S  -> doctor: unsupported   torch: supported
sm_87  Jetson Orin           -> doctor: unsupported   torch: supported
sm_103 B300                  -> doctor: unsupported   torch: supported
sm_121 GB10                  -> doctor: unsupported   torch: supported
```

And end to end, with the real CLI and the capability patched to (8,9):

```
GPU 0: NVIDIA GeForce RTX 4090 (7.9 GB) — sm_89 (Ada) Torch build does not include this GPU architecture
Reinstall a CUDA-enabled PyTorch build for your driver: `pip install torch --index-url https://download.pytorch.org/whl/cu132`
```

That reinstall returns the same arch list, so the alarm can never clear. This is not something you made up: #829's own one-liner specified exact-match-plus-PTX, and none of us could see the problem because this box (sm_120) and the old one (sm_86) are both in the list. It is the same class as the `sm_90a` false positive I asked you to fix first, with a much bigger blast radius.

**Fix:** a non-`a` entry `sm_{major}{m}` covers the card when m <= minor. An `a` entry (`sm_90a`) covers only its exact capability. Keep the PTX rule as it is. I ran that rule over 13 cases with 0 mismatches, and it keeps every expectation your current arch tests encode. Please add these three tests:
- this box's arch list with capability (8,9) -> supported
- (8,0) against `['sm_86']` -> not supported
- (10,3) against `['sm_100a']` -> not supported

The last two keep it from collapsing into torch's looser "any same major" check, which would call a V100 (7,0) covered by `sm_75`.

## Non-blocking

2. **The broken-torchao fix works, but nothing pins it.** Changing `except Exception` back to `except RuntimeError` leaves all 758 tests passing. I checked the behaviour by execution with a `torchao/__init__.py` that raises `ValueError`: the GPU panel now renders `NVFP4: hardware=yes, software=no` instead of crashing. One line pins it: monkeypatch `soup_cli.utils.torchao_compat.resolve_torchao_class` to raise `ValueError` and assert `is_nvfp4_software_supported() is False`. For the record, `soup doctor` still exits 1 in that environment, from the Dependencies table (`__import__` guarded only by `except ImportError`). That happens identically on `main`, so it is not yours and it does not belong in this PR.
3. **The PR body is still stale.** It announces the torchao floor moving 0.4.0 -> 0.7.0, a test for `torchao>=0.7.0`, and "41 passed". None of that is in the tree: the doctor row now reads `>=0.18.0` from #1066.

## Nits

- `test_torch_gpu_arch_supported_accepts_a_suffixed_architectures` (`tests/test_doctor.py:194`) is a verbatim copy of `..._accepts_suffixed_sm_target` (`:137`).
- The panel test still sets `fake_quantization.NVFP4Config = object` (`:260`). That line is dead because `resolve_torchao_class` is monkeypatched at `:254`, and it names the one class #826 says never existed.
- `_gpu_architecture_name` returns "Unknown" for 11.x, which `is_blackwell_gpu()` (major >= 10) counts as Blackwell.
- The `tests/test_v0530.py` change is a reformat only and can be dropped.

PR #948, opened 2026-09-13
fix(ui): cap /api/chat/send and /api/data/inspect request bodies

### LINKED ISSUE #939: soup ui: /api/chat/send and /api/data/inspect still read an unbounded request body
## What

#915 caps the request body on the three `soup ui` endpoints that parse YAML. Two other POST routes still read an unbounded body:

- `POST /api/chat/send` — accepts a 3 MiB body, returns **200**, and forwards it upstream.
- `POST /api/data/inspect` — reads the **full** 3 MiB body and only then rejects the path with 400.

Both measured against the #915 branch while reviewing it.

## Why it is filed separately

#915's issue (#897) is specifically about the YAML-accepting endpoints, where the cost is CPU spent parsing (~31 s for 50,000 bogus keys before the fix). Neither of these parses YAML, so neither was in that scope, and I said on #915 that I would file this rather than ask the author to widen their PR.

The shapes are different from #897 and from each other:

- `/api/chat/send` is **memory and upstream amplification** — the body is buffered and forwarded, so the cost is not only ours.
- `/api/data/inspect` is **wasted I/O before a rejection that was always going to happen** — the path check does not depend on the body at all, so the read is pure waste. That one is arguably the cheaper fix and the clearer bug: reject first, read never.

Both are authenticated-only, which bounds the severity but does not remove it — `soup ui --public` binds to the LAN, and an authenticated client is not automatically a trusted one.

## Acceptance criteria

- [ ] A cap on both endpoints, or a stated reason in the code why one of them should not have it.
- [ ] `/api/data/inspect` rejects the bad path **before** reading the body, with a test asserting the body was never read (a `receive()` that raises if called — `tests/test_issue897_ui_yaml_body_cap.py` has the pattern #915 shipped).
- [ ] Tests pinning the exact boundary in both directions, so a future `>` to `>=` flip is visible. Again, #915's `test_body_cap_exact_boundary` is the model.
- [ ] Reuse `_YamlRequestBodyLimitMiddleware`'s mechanism rather than adding a second, differently-shaped limiter. If the limits should differ per route, make that a table rather than a second code path.

The hardest part is not the cap; it is that `_YAML_REQUEST_PATHS` is a **hand-maintained allowlist with no guard**, so adding two more paths to it makes the list longer without making it self-maintaining. A ratchet asserting that every POST route reaching a body-consuming handler appears in the table would be worth more than the two entries, and would close the whole class instead of two instances of it. That is the higher-value version of this issue if you want it.

Related: #897 (the YAML half, fixed in #915).

AmirF194:
`/api/chat/send` and `/api/data/inspect` both take a Pydantic body model with no size bound, so FastAPI parses the whole request into the model before either handler's own checks run. `chat/send` then forwards that body upstream; `inspect` reads it in full even though the path check that follows never depends on it.

Repro'd both: a 3 MiB `chat/send` body with an unreachable endpoint gets a 200 and reaches the proxy call with no size check, and a 3 MiB `inspect` body with a bad path still gets read twice by `Request.body()` before the 403.

Added `_RequestBodySizeLimitMiddleware`, same check-then-stream shape as #915's `_YamlRequestBodyLimitMiddleware` (Content-Length first, then the actual ASGI bytes so a missing or understated header can't bypass it), on a path -> limit table rather than a second differently-shaped limiter. Limits are sized per route: 1 MiB for chat/send (it forwards upstream, so it gets room for a real conversation), 8 KiB for inspect (a path and an int never need more).

I didn't build this on top of #915 since it's still open, so for now it's a second middleware instance. Once #915 lands the two tables could fold into one; happy to do that as a follow-up if you'd rather not carry both.

10 new tests, mirroring #915's own shapes: boundary pinned in both directions per route, the Content-Length fast path, an understated header, a malformed header, and an untouched route passing straight through. The two route-level tests fail on main (they reach `is_under_cwd`/`urlparse` on a body that should never get there) and pass on the branch, same container both ways. `ruff check` is clean and the full `test_ui*` + auth/chat/inspect suite is green (171 passed).

Didn't run the full repo suite here (the `[dev]` install pulls in the training stack, out of scope for a Web UI change) or the OS/Python matrix beyond 3.11.

Fixes #939


MakazhanAlpamys:
Reviewed at `3fe4d40c`. **The cap itself is correct and I could not defeat it** — every bypass I tried returns 413, measured end-to-end through the real app with 3 MiB bodies:

| vector | `/api/chat/send` | `/api/data/inspect` |
|---|---|---|
| honest oversized `Content-Length` | 413 | 413 |
| **chunked, no `Content-Length` at all** | **413** | **413** |
| **lying `Content-Length: 5`, body 3 MiB** | **413** | **413** |
| malformed `Content-Length: not-a-number` | 413 | 413 |

**And "before the parse" is proven rather than asserted.** Oversized *invalid JSON* returns 413 while small invalid JSON returns 422 — 422 is what body parsing produces, so 413 on the same malformed payload shows the cap fires first:

```
invalid JSON, oversized -> 413      invalid JSON, small -> 422
n=1048576 (exactly the limit) -> 422 (admitted, fails parse)
n=1048577 (limit+1)           -> 413
n=8192 / n=8193 on inspect    -> 422 / 413
```

Baseline is the real defect, not a missing symbol: on `origin/main` the two integration tests reach `fastapi/routing.py:354 run_endpoint_function`, i.e. the whole 2 MiB body was read and parsed before the handler's own checks.

Four mutations, each anchor counted once, each mutated module really imported: killing the limit lookup fails 7, killing the **streamed-byte** check fails 4 (and correctly leaves the header fast path intact), killing the `Content-Length` check fails exactly 1 (also correct — the streamed count still catches the oversize, so only the "didn't read the body" property is lost), and an off-by-one `> max + 1` fails 3, pinning the boundary in both directions. 171 passed across the whole UI surface, zero removed test lines, ruff clean.

Auth ordering: the cap is ASGI middleware, auth is a route dependency, so the cap runs first — unauthenticated + 3 MiB gives 413 rather than 401. I checked before calling that acceptable: route existence is already disclosed by 401-vs-404 on a small body, so the only incremental leak is the magnitude of the cap, and running before auth is what stops an unauthenticated caller making the server buffer an unbounded body. Keep it. CORS survives the 413, as the comment at `app.py:330` claims.

---

**The blocking item is not quality, it is that this duplicates #915.**

`_YamlRequestBodyLimitMiddleware` is **not on `main`** — it exists only on open PR #915. Yours is a near-verbatim copy of it:

```
#915 class: 69 lines   #948 class: 69 lines   identical lines: 54 (78%)
git merge-tree pr915rv pr948rv -> CONFLICT in src/soup_cli/ui/app.py
```

`_reject`, the `content-length` loop including its comment byte-for-byte, the buffering loop and `receive_buffered` are identical; the only structural difference is that yours looks the limit up in a table. #939's fourth acceptance criterion says exactly this: *"Reuse `_YamlRequestBodyLimitMiddleware`'s mechanism rather than adding a second, differently-shaped limiter. If the limits should differ per route, make that a table rather than a second code path."*

**My call: yours is the shape that survives.** The table is the strict generalisation, so `_RequestBodySizeLimitMiddleware` becomes the single limiter and #915's three YAML paths fold into `_BODY_SIZE_LIMITS` (`/api/config/validate`, `/api/train/start`, `/api/config/from-form` at 1 MiB), with `_YamlRequestBodyLimitMiddleware` deleted. I am posting the same note on #915 so @Amix29 hears it from me rather than from a conflict. Whichever of you merges second does that rework — and since I am naming yours as the base, the rework is more likely to be on that side, which is why I am saying it here too.

Worth knowing meanwhile: **merging this alone leaves 3 of 5 body-taking routes uncapped** (`/api/config/validate`, `/api/train/start`, `/api/config/from-form` all accept 3 MiB today). That is #915's scope, not a regression from you.

---

**One item from the issue is not met, and it needs either code or a sentence.**

#939's second criterion asks that `/api/data/inspect` reject a bad path **before** reading the body, with a test asserting the body was never read. It still reads first — measured with a receive-counter driving the full app:

```
inspect: receive() calls before rejection = 1   status = 403
```

The middleware drains the body and it is re-parsed into `DataInspectRequest` before `is_under_cwd` rejects at `app.py:640`. Bounding the waste from 3 MiB to 8 KiB is the substance of the fix and is defensible — but the issue explicitly allows *"a stated reason in the code why one of them should not have it"*, and the PR neither implements it nor states one. Pick either.

**Three smaller ones:**

- `changelog.d/0.75.0/948.fixed.md` says `(#939 in #948)` — house form is `(#939 by @AmirF194 in #948)`, cf. `727.fixed.md`'s `(#716 by @dchaudhari7177 in #727)`.
- `app.py:46` — `limits: dict = _BODY_SIZE_LIMITS` is a shared mutable default with a bare `dict` annotation. Nothing mutates it and ruff's configured rules do not flag B006, so this is only worth changing while you are in the file: `Mapping[str, int]`.
- `app.py:41` — the docstring points at `#915`'s `_YamlRequestBodyLimitMiddleware`, which a reader on `main` cannot find. Resolves itself if the consolidation above happens.

**And the part the issue called the higher-value version is still open:** a ratchet asserting that every POST route reaching a body-consuming handler appears in the table. Without it `_BODY_SIZE_LIMITS` stays hand-maintained — and with two tables, if both PRs land unconsolidated, doubly so. Worth a follow-up issue rather than this PR. There is also no test pinning the auth-vs-cap ordering, so a future refactor moving auth into middleware could flip it silently.

CI is pending on `3fe4d40c`; I need it green on the current head.

AmirF194:
Pushed 7759b07: changelog credit fixed to the house form, limits typed as Mapping[str, int], and a two-line comment on inspect_data explaining why the path can't be checked before the body is read (the path is a JSON field in that body, there's nothing to check yet). 171 passed, ruff clean.

Agreed on the consolidation call. If this one merges second I'll fold #915's three YAML paths into _BODY_SIZE_LIMITS and drop _YamlRequestBodyLimitMiddleware myself.

CI is running at the new head.

MakazhanAlpamys:
Verified `7759b07c`. **Merge-ready** — three of the four items closed, and the fourth is a one-line leftover I am not holding the PR for.

**(a) The `/api/data/inspect` ordering item — closed via the route #939 explicitly allows**, and I checked the stated reason is factually true rather than just present. `app.py:631-632`:

```python
# The path lives inside the body, so it can't be checked before the
# body is parsed. _BODY_SIZE_LIMITS caps that read at 8 KiB instead.
```

`DataInspectRequest` declares `path: str` as a **body** field and the route is a POST, so there is no path to check before the body is read — the criterion as literally written was unsatisfiable for this route. Stating that beats implementing something that pretends otherwise.

**(b)** credit now `(#939 by @AmirF194 in #948)`. **(c)** `limits: Mapping[str, int]` — and `typing.Mapping` rather than `collections.abc` is right here, 96 files under `src/soup_cli/` already spell it that way.

**Re-confirmed the cap did not regress**, real app, 3 MiB bodies, both routes:

| case | result |
|---|---|
| honest oversized `Content-Length` | 413 |
| chunked, no `Content-Length` | 413 |
| lying `Content-Length: 5`, 3 MiB body | 413 |
| malformed `Content-Length` | 413 |
| no `Content-Length` header at all (raw ASGI) | 413 |

Before-parse still proven by the same-shape pair: invalid JSON at 64 B gives 422, the same invalid JSON at 3 MiB gives 413. Boundary exact in both directions on both routes (`n == limit` -> 422, `n == limit + 1` -> 413).

**And the mutation is better evidence than the test count.** Disabling the streamed-byte check (`if received > max_body_size:` -> `if False:`, anchor counted once, module really imported) fails 4 tests — but the live matrix under that mutation shows the bypass actually reopening, which is what I wanted to see:

```
honest oversized Content-Length  -> 413   (header fast path still catches it)
chunked, no Content-Length       -> 422   <- 3 MiB read and parsed
lying Content-Length: 5          -> 422   <- 3 MiB read and parsed
malformed Content-Length         -> 422   <- 3 MiB read and parsed
```

171 passed across the full UI surface, ruff clean.

**(d) is still open, and it is one line:** `app.py:41` still points at `#915's _YamlRequestBodyLimitMiddleware`, which does not exist on `main` — `git grep` on `origin/main -- src/` returns nothing, so a reader cannot find the symbol. It resolves itself once #915 folds into `_BODY_SIZE_LIMITS`, per the ordering @Amix29 and I agreed, so I am not blocking on it. Fix it if you are pushing anyway; if not, I will merge as-is and it disappears with the fold.

CI is queued on `7759b07c`, which is the only thing I am waiting on.

MakazhanAlpamys:
Merged as `7e90de6b`. Thank you.

**What made this mergeable ahead of the competing PR was not the mechanism, it was that the ordering claim is demonstrated.** "The cap fires before the parse" is the kind of sentence that is easy to write and hard to check, and your test pair checks it: the same malformed JSON returns 422 at 64 bytes and 413 when oversized. Since 422 is what body parsing produces, 413 on an identical payload can only mean the cap ran first. That is an argument, not an assertion.

I could not defeat the cap. Re-confirmed at your final head against the real app with 3 MiB bodies — honest oversized `Content-Length`, chunked with none at all, a lying `Content-Length: 5`, a malformed header, and a raw ASGI request with no such header — all 413, boundary exact at 1048576 admitted and 1048577 refused. Four mutations die, and the one that kills only the streamed-byte check correctly leaves the header fast path intact, which is the mutation that tells you the two checks are independent rather than redundant.

**The judgement call on `/api/data/inspect` is the part I would point another contributor at.** The acceptance criterion asked for the path check to run first; that route's path is a body field, so as literally written the criterion was unsatisfiable. You capped instead and wrote two lines saying why, in the code. Implementing something that pretends to satisfy an impossible criterion is the failure mode there, and you avoided it by naming the problem.

You are in [CONTRIBUTORS.md](https://github.com/MakazhanAlpamys/Soup/blob/main/CONTRIBUTORS.md) for it. And thank you for volunteering to fold #915's YAML paths in — that turned out not to be needed from you, since @Amix29 took the rebase, but offering it is what let me pick a merge order on merit rather than on whose diff would suffer.

AmirF194:
The unsatisfiable-criterion framing on `/api/data/inspect` is the part worth keeping in mind for the next route like it: name the constraint in a comment instead of pretending to satisfy it. Glad the mutation pair held up under your own attack pass too.

PR #950, opened 2026-09-13
fix(autodistill): verify the MLX teacher worker by nonce, not PID

### LINKED ISSUE #898: autodistill mlx_worker: receipt PID check fails under a Windows venv (python.exe is a launcher)
`src/soup_cli/autodistill/mlx_worker.py::run_mlx_teacher_capture_process` compares the receipt PID written by the child (`os.getpid()`) against `Popen.pid` and raises `ValueError("worker receipt PID does not match the exited child")` when they differ. On Windows a virtualenv's `Scripts/python.exe` is a redirector that starts the real interpreter as a grandchild, so `sys.executable` spawns two processes and the two PIDs never match (measured during the v0.75.0 release: `Popen.pid` 8456, child `os.getpid()` 23380). All six tests in `tests/test_issue580_autodistill_mlx_worker.py` that spawn a real child fail under a venv and pass under the base interpreter with `PYTHONPATH=src`.

No user path hits this today (autodistill is not wired to the CLI), but anyone running Soup from a Windows venv — the common case — will the moment it is.

**Fix path:** bind the receipt to something the parent can verify through a launcher, e.g. a nonce the parent generates and passes via env/argv that the child echoes into the receipt, and treat the PID as informational; or resolve the real interpreter (`sys._base_executable` / `venv`'s `home`) before spawning.

**Acceptance:**
- [ ] the six tests pass when pytest runs from a Windows venv
- [ ] a receipt written by a different process (wrong nonce) is still rejected

SID-6921:
Closes #898.

## The bug

`run_mlx_teacher_capture_process` verifies the spawned worker by comparing `receipt.worker_pid` to `process.pid`. On Windows, a venv's `Scripts/python.exe` is a launcher/redirector: it spawns the real interpreter as a grandchild, so `subprocess.Popen(...).pid` is the launcher's PID, not the interpreter's. The child's own `os.getpid()` (what gets written into the receipt) is a different number entirely.

Reproduced directly with a real venv:

```python
proc = subprocess.Popen([sys.executable, "-c", "import os; print(os.getpid())"], ...)
# proc.pid: 34300
# child os.getpid(): 23948
# match: False
```

Confirmed this breaks the project's own suite, not just a standalone repro — all six tests in `test_issue580_autodistill_mlx_worker.py` that spawn a real child fail under a venv interpreter with `ValueError: worker receipt PID does not match the exited child`.

## The fix

`run_mlx_teacher_capture_process` now generates a nonce (`secrets.token_hex(16)`) and includes it in the request written for the child. The child echoes it back into `MlxTeacherWorkerReceipt.worker_nonce`. The parent verifies the nonce instead of the PID. `worker_pid` stays on both the request/result models — it's informational only now, not part of the security check.

## Verification

- All 6 real-child-process tests pass under both the base interpreter and a real venv (`python -m venv`, confirmed via the standalone repro above before touching any code).
- Added `test_controller_rejects_a_receipt_whose_nonce_does_not_match_the_request` for the "wrong nonce is still rejected" case — a stale worker directory shouldn't be waved through just because PID is no longer checked.
- Mutation-tested: reverted the nonce check back to PID equality, confirmed all 6 tests fail under the venv with the exact reported symptom; restored, confirmed green again.
- ruff clean; mypy shows the same 7 pre-existing errors as main (confirmed via `git stash` diff), none introduced by this change.


MakazhanAlpamys:
Reviewed at `1a69a5da`. **This is right, and I verified the premise independently rather than taking it from the issue.** Five-line Popen probe on this box:

```
venv Scripts/python.exe : Popen.pid = 5300   child os.getpid() = 31064   MISMATCH
base interpreter        : Popen.pid = 24404  child os.getpid() = 24404   MATCH
```

Then the 2x2 that actually matters, using a real `python -m venv` rather than a simulated launcher:

| code | interpreter | result |
|---|---|---|
| `origin/main` | base | 14 passed |
| `origin/main` | **venv launcher** | **6 failed**, all `ValueError: worker receipt PID does not match the exited child` |
| this PR | base | 15 passed |
| this PR | **venv launcher** | **15 passed** |

The six are exactly the six #898 names.

**On the nonce being sound — it is, and it is strictly stronger than the PID it replaces, which is not the obvious direction.** `secrets.token_hex(16)` is 128-bit CSPRNG, generated inside the call so it is fresh per run, shape-pinned by `_NONCE_PATTERN` on both request and receipt. It travels in `request.json` written through `_atomic_write` (`mkstemp` + `os.replace`, so 0600 on POSIX) — **only the path is in the process table**. The PID it replaces was world-readable from the process table by any local user, so third-party forgeability improves. Replay is infeasible in both directions: a stale receipt carries the old nonce, and a receipt pre-planted before the child runs makes the *child* fail, because `_atomic_write` refuses to overwrite. The comparison is a plain `!=` rather than constant-time, and I am deliberately not asking you to change it: one comparison per run, `worker_root.exists()` forbids re-entering a transaction id so there is no repeated oracle, and the only observable is a raised exception.

The property you now prove is "the receipt was written by something that could read the request I wrote" rather than "by the process I spawned". Given the 0600 request file, that is the right weakening for a launcher-tolerant check, and it is worth that the PR says so.

**Mutation, blocking gate — passes.** Anchor `if receipt.worker_nonce != request.worker_nonce:` matched exactly once. Both reverting to the PID compare and deleting the check outright give `Failed: DID NOT RAISE ValueError` at `tests/test_issue580_autodistill_mlx_worker.py:376` — behavioural, module really imported first so not a NameError artifact. 81 passed across all five autodistill files on base, 70 under the venv, zero removed test lines, ruff clean.

---

**One thing to change before I merge, and one to consider.**

**Change:** `changelog.d/0.75.0/950.fixed.md` credits `(#898, #950)` with no author. House form is `(#898 by @SID-6921 in #950)` — see `882.fixed.md`'s `(#878 by @Srinivasan8888 in #882)`. (And for the record, I had guessed the fragment might be overlong at +10 lines; it is the *shortest* in that directory, and the content checks out — I verified the "always failed even on a successful capture" claim and the venv-on-Windows scoping.)

**Consider — the positive half of #898's acceptance criteria has no CI-visible guard.** The first criterion is "the six tests pass when pytest runs from a Windows venv". Your new test covers only the negative half. Every test passes `python_executable=sys.executable`, and CI uses `actions/setup-python` + `pip install -e` with no venv, so on CI `sys.executable` is the real interpreter — **a revert to the PID check would still pass all six there**. The mutation gate does catch it, so this is a coverage gap rather than an unguarded regression. I checked the fix is cheap and real rather than theoretical:

```python
venv.create(tmp_path / "launcher", with_pip=False, system_site_packages=True)
# measured here: 0.04 s, Popen.pid 37280 vs child 21592 -> MISMATCH
```

then pass that `Scripts/python.exe` as `python_executable=` (on POSIX `bin/python` is a symlink, so it degenerates to a no-op rather than failing). That test goes red on `origin/main` under Windows and green here. Your call whether it belongs in this PR or a follow-up.

**Two LOW notes, neither a request:**

- `worker_nonce` became a required field on two `.v1` frozen `extra="forbid"` schemas without a version bump, so a receipt from an older worker fails validation. Contained today — one construction site each, request and receipt are written and read inside one call, `worker_root` cannot pre-exist, no golden fixture exists — but it is the kind of `.v1` mutation that bites once a second reader appears.
- `MlxTeacherCaptureResult.worker_pid = process.pid` is the *launcher's* PID under a venv, i.e. the wrong one on exactly the platform this PR is about. The receipt's own `worker_pid` (`os.getpid()`) is the useful one. The fragment's observability claim slightly oversells what survives.

CI is queued on `1a69a5da`. Fix the credit line and I will merge when it lands.

SID-6921:
Fixed the credit line to `(#898 by @SID-6921 in #950)` in `eea37d7`, and tightened the changelog's worker_pid claim while I was in there — you're right that the result's copy is the launcher's PID under a venv, not the useful one; the receipt's own is.

Also added the venv-launcher test for the positive acceptance criterion. Reverting the nonce check back to PID equality now fails that test specifically (`worker receipt PID does not match the exited child`), so the gap you found is closed rather than just noted. Used `venv.create(..., with_pip=False, system_site_packages=True)` per your measurement — no meaningful runtime cost.

Left the two LOW notes as-is per your framing (neither a request) — the schema version bump and the private-constant duplication are both real but scoped smaller than this PR.

MakazhanAlpamys:
Verified at `eea37d77`. **Merge-ready** — the acceptance gap is genuinely closed, not just noted, and I confirmed that by mutation rather than by reading the new test.

Reverting the nonce check to PID equality (anchor counted once; `inspect.getsource` on the mutated module confirming the line is live, so not a NameError artifact):

```
2 failed, 14 passed

test_worker_verification_survives_a_real_venv_launcher
  ValueError: worker receipt PID does not match the exited child   <- your new test
test_controller_rejects_a_receipt_whose_nonce_does_not_match_the_request
  DID NOT RAISE ValueError                                          <- the negative half
```

**The other 14 pass under that mutation**, which is precisely the gap: without your new test a revert is invisible to CI. It is now the only thing standing between a PID check coming back and nobody noticing.

And the venv genuinely reproduces the defect here rather than passing incidentally:

```
venv.create wall time: 0.011 s
Popen.pid      = 18736
child getpid() = 9172     -> MISMATCH
```

**Two design details I want to name because they are what make this work on CI rather than only on your box.** `system_site_packages=True` is load-bearing — with `with_pip=False` the venv has no pydantic, and the child imports it from the base env. And the pre-existing `_runtime_environment` helper already puts `<repo>/src` on the child's `PYTHONPATH`, so the venv child finds `soup_cli` regardless of install mode rather than depending on CI's editable layout.

**The POSIX answer is the right one and the docstring says it out loud**: no skip, the test runs on all nine cells, picks `bin/python` (a symlink, so no intermediate process), and still asserts real outcomes — a valid end-to-end run that simply cannot discriminate there. Discrimination lives in the three Windows cells, which is correct scope for a Windows-launcher defect. Writing that down in the docstring rather than leaving the next reader to discover it is the part I would have asked for.

Cost is ~1.1 s for both new tests (0.49 s for this one, ~0.07 s marginal over a comparable spawn test). Credit line is house-form, the tightened `worker_pid` sentence is accurate against `mlx_worker.py:458` and `:568`, 71 passed across the five autodistill files, ruff clean, zero removed test lines.

---

**One optional strengthening — your call, and I will merge either way.**

`tests/test_issue580_autodistill_mlx_worker.py:341-343` asserts that the run *succeeded*, which means the test cannot distinguish "verification survived a launcher mismatch" from "no mismatch occurred". Today on Windows the mismatch is real — I measured it above — but if a future CPython stopped using a redirector, the test would quietly become non-discriminating on Windows too, with no signal. That is the same passes-for-the-wrong-reason class the test exists to guard against, one level up.

It is cheap to close: the receipt persists at `publication/.workers/<transaction_id>/worker-receipt.json` and nothing removes that directory, so the test can read it and, under `sys.platform == "win32"`, assert `receipt["worker_pid"] != result.worker_pid` — turning "the launcher mismatch was actually exercised" from an assumption into an asserted precondition.

If you want it, push it and I will merge after. If you would rather stop here, say so and I will merge as-is and file it — the current form already blocks the revert, which was the ask.

SID-6921:
Added it in 1fa68ee — reads the receipt back and asserts the pid genuinely differs from what Popen reported, gated on win32. Checked it isn't vacuous by flipping the comparison first: it fails with the actual numbers (14040 vs 4188 on this box), so the mismatch really is being exercised here, not assumed.

MakazhanAlpamys:
Verified `1fa68ee0`. **Closed, and closed with the property that matters: the test now notices when it stops being discriminating.**

I simulated the mismatch going away — swapped `python_executable=venv_python` for `sys.executable` (anchor counted once; this box's `sys.executable` is the base install, `sys.prefix == sys.base_prefix`, so genuinely not a redirector):

```
unmodified:            1 passed
launcher removed:      assert receipt["worker_pid"] != result.worker_pid
                       AssertionError: assert 24312 != 24312
```

That is the whole point of the addition and it works. The two sides are genuinely two different processes — `mlx_worker.py:458` writes the child's own `os.getpid()` into the receipt, `:568` sets the result's `worker_pid` from what `Popen` reported — so comparing the persisted file against the returned result is exactly the right pair.

Gating is clean: the four new lines are the only statements under `if sys.platform == "win32":`, so POSIX evaluates nothing new. The earlier mutation still holds — reverting the nonce check to PID equality fails the venv test at the real call site with `ValueError: worker receipt PID does not match the exited child`. 71 passed across the five autodistill files, ruff clean, credit in house form.

**One judgement call worth naming, because it is yours and it is deliberate:** the signal you chose is a *failure*, not a skip. If a future Windows CPython ever ships a venv `python.exe` that is not a redirector, CI goes red on a test that is no longer reproducing anything. Your comment says as much. I agree with the choice — silently un-discriminating is the worse failure mode, and a red test is a conversation while a green one is nothing.

Merge-ready. CI is queued on `1fa68ee0` — note the ubuntu and macos cells have not yet run the venv test at all, since it is new and the previous head never went green either, so that is the one thing I am actually waiting on rather than assuming.

MakazhanAlpamys:
Merged as `0f7dc060`. Thank you.

**The part I want to name is that you replaced the check rather than patching around it.** The easy fix was a platform branch — skip the comparison on Windows, or walk the process tree to find the grandchild. Both would have worked and both would have left the identity proof depending on something that is not identity. A nonce the parent generates and the worker echoes is what the check was always trying to establish, and it holds on every platform for the same reason.

The second thing: you kept the test's ability to notice when it stops testing anything. I simulated the defect going away — swapping `python_executable=venv_python` for a non-redirecting interpreter, on a box where `sys.prefix == sys.base_prefix` so it is genuinely not a redirector — and the test fails rather than passing vacuously. That is the property most identity tests lack, and it is why this one will still be worth something in a year.

Worth recording for whoever hits this next: the reason #898 survived a whole release is that nobody had run the suite from a venv on that machine. The main tree uses a global editable install, so the six failures appeared for the first time during a release cut, looking like a release regression when they were a two-year-old latent defect in the check.

`autodistill` is not wired to the CLI, so no user path was hitting this — which makes it the kind of fix that only gets made by someone who cared that the test was lying.

PR #951, opened 2026-09-13
fix: fail the control-strip guard when a scanned file does not parse

### LINKED ISSUE #949: test_issue896 guard skips a file that fails to parse, so 'zero offenders' can mean 'zero among the files I could read'
`tests/test_issue896_no_duplicate_terminal_sanitizer.py` (merged in #907) walks every file under `src/soup_cli/` with `ast.parse` to find duplicated C0/DEL control-strip tables. When a file fails to parse, the guard currently **skips it silently** and the run stays green.

That is the wrong direction for two reasons:

1. An unparseable file under `src/soup_cli/` is a worse problem than the duplicate table the guard exists to find, and it is one nothing else in the suite reports as such — the failure would surface later as a confusing `ImportError` somewhere unrelated.
2. A skipped file is indistinguishable from a clean file in the output, so the guard's "zero offenders" answer silently becomes "zero offenders among the files I could read".

**Proposed fix:** `pytest.fail` naming the file and the `SyntaxError`, rather than `continue`. One line plus a test that an intentionally broken temp file inside the scanned tree fails the guard.

**Acceptance criteria:**

- [ ] A file under `src/soup_cli/` that does not parse fails the guard, and the failure message names the file.
- [ ] A test proves it, using a temporary unparseable file placed inside the scanned tree (and removed afterwards) — not by monkeypatching `ast.parse`, since the point is the real walk.
- [ ] The existing zero-offender behaviour and the can-it-fail test are unchanged.

Context: #907 found nine copies where #896 named six, because the ninth (`commands/serve.py`, a function-local `_ctrl`) was invisible to a name-based grep. The guard's value is that it sees shapes rather than names — which makes a silently skipped file exactly the gap worth closing.

Raised during the review of #907; @VanshSharmaPES chose to stop at the consolidation, which was the right scope call. Filing it so it is not lost in a merged PR's thread. Good first issue — it is one line of source and one test.

jarik2014:
**What was broken**

`_offenders()` walks every tracked file under `src/soup_cli/` and parses it, but a
`SyntaxError` was swallowed:

```python
except SyntaxError:  # pragma: no cover - not expected in this tree
    continue
```

A file that does not parse is a *worse* problem than the duplicate the ratchet exists to
find, and nothing else in the suite reports it as such — it would surface later as an
`ImportError` somewhere unrelated. Skipping it also makes the guard's clean answer mean
"zero offenders among the files I could read", which is indistinguishable from a real
clean run.

**What this changes**

`pytest.fail` naming the file and the `SyntaxError`, replacing the `continue`. One `except`
clause; the walk, the name check and the shape check are untouched.

The test is in `TestTheScannerCanActuallyFail`, next to the two proofs that the scanner
can find a reintroduced copy:

```python
def test_an_unparseable_file_in_the_tree_fails_the_guard(self):
```

It writes `src/soup_cli/zz_unparseable_guard_probe.py` with `def broken(:`, registers it with
`git add -N` so it is genuinely part of the `git ls-files` walk the guard performs, asserts
`_offenders()` fails naming the file, then removes both the file and the index entry in a
`finally`. No monkeypatching of `ast.parse` — the run is the real walk over a real file, so
a scanner that silently stopped reading files would fail this test too.

**Evidence**

With the fix, all seven tests in the file pass:

```
$ uvx --with pytest-cov pytest tests/test_issue896_no_duplicate_terminal_sanitizer.py --no-cov -q
.......                                                                  [100%]
7 passed in 37.51s
```

Reverting only the `except` clause back to `continue` — the new test, alone:

```
$ uvx --with pytest-cov pytest tests/test_issue896_no_duplicate_terminal_sanitizer.py --no-cov -q -k unparseable
E           Failed: DID NOT RAISE Failed
FAILED ...::TestTheScannerCanActuallyFail::test_an_unparseable_file_in_the_tree_fails_the_guard
1 failed, 6 deselected in 21.94s
```

Lint is clean on the touched file:

```
$ uvx ruff check tests/test_issue896_no_duplicate_terminal_sanitizer.py
All checks passed!
$ uvx ruff format --check tests/test_issue896_no_duplicate_terminal_sanitizer.py
1 file already formatted
```

**What I did not do**

* I could not run the whole suite here: the `[dev]` extra pulls `torch`/`transformers`, and my
  box is a 1 GB VM. The touched file needs neither, which is why the evidence above is that
  file plus a deliberate revert.
* The existing zero-offender test and the two can-it-fail tests are unchanged, as the
  acceptance criteria ask.
* I left `_tracked_python_files()` on `git ls-files` rather than switching the walk to a
  filesystem scan — that would change what the ratchet covers, which is not this issue.

Closes #949


MakazhanAlpamys:
Reviewed at `03bdef04`. **This is a stronger first contribution than most, and it took the harder correct route rather than the easy one.**

The issue asked for a real temporary file inside the scanned tree rather than a `monkeypatch` on `ast.parse`, and you did that — including the `git add -N`, without which the probe would be invisible to a `git ls-files` walk and the test would have been vacuous. You then asserted the probe actually reached the scanned set *before* asserting the failure (`:201-203`), which is the guard against exactly that vacuity. Most people write the monkeypatch.

**I verified it on the real tree rather than only through your fixture** — planted a genuinely broken file under `src/soup_cli/` outside the test:

```
tracked (git add -N)  -> RED, naming the real file:
  src/soup_cli/zz_manual_broken_probe.py does not parse (invalid syntax at line 1),
  so this scan cannot tell whether it defines the control-strip table
file removed          -> 7 passed, git status clean
```

**And it does not degrade what the guard was for.** I planted a function-local `range(0x20)` dict-comp — the #907 `commands/serve.py` shape, different name, non-module scope — into a real tracked module: caught, `1 failed`, naming the file and line. Removed, 7 passed. Floor assertions intact at 504 files scanned.

**Mutation:** `pytest.fail(...)` -> `continue` (anchor counted exactly once; `pytest.fail(` appears once in the whole file; `--collect-only` still collects 7, so no collection error masquerading as a catch):

```
>           with pytest.raises(pytest.fail.Exception) as failure:
E           Failed: DID NOT RAISE Failed
1 failed, 6 passed
```

Exactly the new test, at the behavioural assertion, nothing else disturbed.

**On the 9 removed lines — I checked before assuming churn, and they are not.** Five are `_tracked_python_files()` collapsed to one line, two are the fix itself (including the now-dead `# pragma: no cover`, correctly removed), two are a rewrapped assert message. `ruff format --check` says `main`'s copy of this file *"would be reformatted"* and yours *"already formatted"* — #907 merged a file that was not format-clean and you incidentally fixed it. CI runs `ruff check` only, so neither was ever red. Applying `ruff format` to main's version and diffing against yours leaves only the fix and the new test.

Fragment is correct on every axis and, notably, its credit line `(#949 by @jarik2014 in #951)` is **better formatted than `907.fixed.md`**, which it builds on and which carries `(#896 in #907)` with no handle.

---

**One thing I would like changed, and it is one line.**

`tests/test_issue896_no_duplicate_terminal_sanitizer.py:192-199` — `probe.write_text(...)` and the `git add -N` sit *outside* the `try:`. If `git add -N` fails under `check=True`, the file is already on disk and the `finally` never runs. I reproduced it by holding `index.lock`:

```
CalledProcessError: git add -N ... returned non-zero exit status 128
leftover: src/soup_cli/zz_unparseable_guard_probe.py   (?? untracked)
```

The blast radius is small and I measured it rather than guessing — the leak is untracked, so the `git ls-files` ratchets ignore it, and three `rglob("*.py")` scanners (`test_issue775`, `test_issue748`, `test_issue752`) all pass with it present, 69 passed. It fails loudly, never silently green. But a test that writes into `src/soup_cli/` should not be able to leave a broken `.py` behind under any branch, so: start the `try:` before `probe.write_text(...)`.

**Two notes, neither a request:**

- The cleanup's `git rm --cached` uses `check=False`, so a failure to clean the *index* is silent while the file is unlinked anyway, leaving a ` D ` entry. Unlink-then-`git rm` would order it better — but the unlink must happen regardless, so what you wrote is defensible.
- Runtime roughly doubles for this file (2.74 s -> 5.5 s), because the probe sorts last in `git ls-files` order so the whole tree is parsed before the failure fires. That is arguably the point — it proves the walk reaches the end — and ~3 s across nine cells is a fair price. Recording it as a decision rather than letting it be a surprise.

One pre-existing property worth knowing, from #907 rather than from you: the walk is `git ls-files`, so a genuinely broken **untracked** file is still invisible. Irrelevant in CI, where everything is tracked.

Push that one line and I will merge as soon as CI lands. Your checks needed manual approval since this is your first PR here — I have approved them, and I will keep watching, because the approval does not carry to the next push.

jarik2014:
Pushed `d762565`: the probe write and the `git add -N` now sit inside the `try:`, so the cleanup runs on every branch.

Reproduced your `index.lock` path before pushing, and re-ran the file after:

```
$ touch .git/index.lock
$ uvx --with pytest-cov pytest tests/test_issue896_no_duplicate_terminal_sanitizer.py --no-cov -q -k unparseable
E   subprocess.CalledProcessError: Command '['git', 'add', '-N', '--', 'src/soup_cli/zz_unparseable_guard_probe.py']' returned non-zero exit status 128
1 failed, 6 deselected in 2.20s
$ ls src/soup_cli/zz_unparseable_guard_probe.py
ls: cannot access '...': No such file or directory        # the finally ran
$ rm .git/index.lock
$ uvx --with pytest-cov pytest tests/test_issue896_no_duplicate_terminal_sanitizer.py --no-cov -q
7 passed in 57.79s
$ git status --short
 M tests/test_issue896_no_duplicate_terminal_sanitizer.py   # nothing else left behind
```

Lint on the touched file: `ruff check` — All checks passed; `ruff format --check` — already formatted.

On your two notes, deliberately unchanged rather than overlooked:

* **Cleanup order** — left as `git rm --cached` then unlink, since the unlink has to happen regardless and you called it defensible. If the stale-`D` case is worth closing I will swap them in a follow-up rather than widen this diff; say the word.
* **The ~3 s** — kept. The probe sorting last means the walk parses the whole tree before the failure fires, which is the property you highlighted, and I would rather pay it than special-case the ordering.

Thanks for verifying it on the real tree with a planted file and for the mutation run — that is well past what the issue asked of you. And for the note on the untracked-file blind spot: agreed it is #907's property, and the fragment for this PR does not claim otherwise.

MakazhanAlpamys:
Verified `d762565f`. **Exactly the change I asked for, and I reproduced both halves rather than reading the diff.**

Held `index.lock` on the worktree's git dir so `git add -N` fails under `check=True`, then ran the file:

```
before (03bdef04): CalledProcessError, leftover: src/soup_cli/zz_unparseable_guard_probe.py  (?? untracked)
after  (d762565f): CalledProcessError, leftover: none          git status --porcelain: empty
```

The test still fails in that situation — correctly, since `git add` failing is a real problem — but it no longer leaves a broken `.py` behind on the way out. Anchor counted once; 7 passed on a clean run; tree clean afterwards.

The comment you put above the moved block is the right kind: it says *why* both lines are inside the `try`, which is the thing the next person would otherwise undo while tidying.

Nothing else outstanding from my side. The two notes I left as non-requests stand as non-requests — the `git rm --cached` ordering, and the runtime roughly doubling for this file because the probe sorts last in `git ls-files` order so the whole tree is parsed before the failure fires.

I have approved your CI runs again; the approval does not carry across pushes, so that is on me each time. Once they are green I will merge.

MakazhanAlpamys:
Merged as `7c93e062`. Thank you, and welcome — this is your first contribution here, and it is a better first one than most.

**What I merged on.** The verdict was recorded at `d762565f` on 2026-09-13 and I did not re-derive it; all 13 required contexts are now green on that exact head, read per-SHA rather than from the cached PR summary. What I did re-check is the one thing that can change under a PR while it waits: `main` moved, and #767 landed earlier today moving `commands/runs.py` onto the shared `utils.terminal` helper — which is precisely the kind of edit your ratchet polices. So I built the merged tree rather than trusting two separately-green branches: **your guard 7 passed there**, and the other repo-wide ratchets 103 passed.

**The part I want on the record is the second-order defect, because you found it and I did not ask for it.** The obvious reading of `except SyntaxError: continue` is "a file that does not parse is a worse problem than the duplicate we are looking for". True, and enough on its own. But you also wrote down that skipping makes the guard's clean answer mean *"zero offenders among the files I could read"* — indistinguishable from a genuinely clean run. That is the difference between a guard that is incomplete and a guard that **lies in the direction of reassurance**, and it is the argument that makes this worth fixing rather than noting.

Two more things done right:

Your probe registers the broken file with `git add -N` instead of only writing it to disk, so it genuinely enters the `git ls-files` walk the guard performs. Plenty of tests for scanners like this quietly test a fixture rather than the scanner; yours does not.

And the second round was yours to notice. Told the failure path could leave the broken `.py` behind, you moved the cleanup inside the `try` **and wrote a comment saying why both lines sit there** — which is the thing the next person would otherwise undo while tidying. I reproduced it rather than reading it, holding `index.lock` so `git add -N` fails under `check=True`:

```
before (03bdef04): CalledProcessError, leftover zz_unparseable_guard_probe.py (?? untracked)
after  (d762565f): CalledProcessError, leftover none; git status --porcelain empty
```

The test still fails in that situation — correctly, since `git add` failing is a real problem — it just no longer leaves debris on the way out.

Credited in `CONTRIBUTORS.md` (`a780f92e`); you are the 62nd. The two notes I left as non-requests stay non-requests.

One apology owed: this sat green-pending far longer than it should have, because Actions on this repo produced no terminal outcome at all between 2026-09-13T18:30Z and today 15:08Z, and the queue behind that was ~100 runs deep. That is cleared now. Your CI approval also does not carry across pushes, so re-approving each time is on me, not on you.

Issue #936, opened 2026-09-13
data.chat_template is still accepted but unused by PPO, PRM, ASR, MoE routing, and unlearn


Amix29:
## Defect

Follow-up from the runtime review of #804 / #919. `SoupConfig` still accepts `data.chat_template` for `ppo`, `prm`, `asr`, `moe_lora_routing`, and `unlearn`, but those paths do not apply the override. PPO is the clearest live case: its trainer joins message `content` values directly rather than rendering the configured template.

This leaves the same accepted-but-unconsumed configuration class that #804 fixes for the conversational TRL trainers.

## Expected

For each task, either:

- apply the configured template to the tokenizer/data path before rendering, when the task genuinely consumes chat messages; or
- reject `data.chat_template` at config validation with the task named.

## Acceptance

- Behavioral tests cover each affected task; source-string presence alone is insufficient.
- No task accepts `data.chat_template` while silently ignoring it.
- PPO conversational prompts are rendered through the selected template rather than hand-joined.

Found during maintainer review of #919.

kok-o:
@MakazhanAlpamys I'd like to take this follow-up. For the conversational/formatting paths (such as PPO) I'll apply the configured `data.chat_template` override rather than raw joining message content, and for tasks where it is genuinely unsupported, reject it cleanly at schema validation with a friendly error message.

MakazhanAlpamys:
Closed by #1021 (merged as 377d7531), by @kok-o. The PPO trainer now applies `data.chat_template`. `prm`, `asr`, `moe_lora_routing` and `unlearn`, which have no trainer that reads it, refuse it at parse time, and the error names the task.

Issue #891, opened 2026-09-12
soup serve --backend mii sends the engine a doubled BOS (#785 part 2, split from #867)


abdulwaarith0:
Split out from #867 (#785 part 2). #867 fixes the vLLM backend and #782 fixed every path that tokenizes in process; this tracks the DeepSpeed-MII backend, measured to double the BOS the same way vLLM did before #867.

## Measured

Same setup as the SGLang follow-up: RTX 3060 Laptop (6 GB, WSL2), `unsloth/Llama-3.2-1B-Instruct` (Llama-3 template renders `{{ bos_token }}`, `bos_token_id` 128000), one request, `apply_chat_template(tokenize=True)` = 36 ids starting `[128000, 128006]`.

DeepSpeed-MII 0.3.3 tokenizes the rendered string via `mii.modeling.tokenizers.HFTokenizer.encode` — the exact call `MIIPipeline._put_request` makes for every request (`self.tokenizer.encode(input)` -> `tokenizer.encode(input, return_tensors="pt")`, default `add_special_tokens=True`) — giving `prompt_token_ids[:2] == [128000, 128000]` over 37 ids: a doubled BOS.

Run directly on the real tokenizer. The full engine could not start here: DeepSpeed's inference ops need a CUDA toolkit to JIT and there is no `nvcc`, so the fix cannot be verified end to end from this box.

## Fix routes checked (against 0.3.3)

- **Token ids**: `MIIPipeline.__call__(prompts: str | list[str], **generate_kwargs)` takes strings only; `GenerateParamsConfig` has no tokenization field.
- **No special tokens**: `HFTokenizer.encode` hardcodes `tokenizer.encode(input, return_tensors="pt")`. `ModelConfig.tokenizer` is typed `str | MIITokenizerWrapper`, which looks like the lever, but `load_tokenizer` re-wraps whatever it is given in `HFTokenizer`, so a wrapper overriding `encode(input)` gets called as `encode(input, return_tensors="pt")` and raises `TypeError: unexpected keyword argument 'return_tensors'`.

So this one likely needs an upstream MII change or a patched tokenizer object, not just a call-site swap.

## Acceptance

- [ ] For a `{{ bos_token }}` template on a BOS-prepending tokenizer, the engine receives one BOS, matching `apply_chat_template(tokenize=True)` and #782 / #867
- [ ] A model with no chat template (or a template that fails to render) still sends the legacy string and is tokenized by the engine exactly as before
[...]

PR #952, opened 2026-09-13
fix(mcp): validate data doctor format

### LINKED ISSUE #889: soup mcp serve: tool_data_doctor accepts a blank format and refuses unknown ones without naming them (the rest of #878)
Found by @Srinivasan8888 while fixing #878 (in #882), reported in that PR's body and filed here so it is not lost. Same class as #878, different tool, and it is the **remaining** half of the CLI-vs-MCP divergence rather than a new one.

#882 fixed `tool_data_validate`. `tool_data_doctor` in `src/soup_cli/mcp_server/registry.py` still has two related defects, both verified by execution against #882's head:

## 1. A blank format is treated as "auto" instead of being refused

`tool_data_doctor` still spells the argument as `... or "auto"`, so `format=""` is **accepted and a report returned**. After #882, both `soup data validate --format ""` and `tool_data_validate` refuse a blank string. So the same empty value now means two different things depending on which MCP tool you call, which is the shape #878 was filed about.

## 2. Its refusal for a genuinely unknown format does not name the accepted values

A bogus format is refused — that part works — but the message reads:

```
data doctor failed (ValueError)
```

The user is told neither which value was rejected nor what the valid ones are. Compare #882's:

```
unknown format 'bogus'; accepted: auto, alpaca, asr, audio, chatml, dpo, ...
```

An MCP client is usually a model choosing arguments, so a refusal that does not name the accepted set is the difference between a self-correcting call and a stuck one.

## Suggested fix

Reuse what #882 already built rather than writing a second validator: the same `VALID_FORMATS` check, the same message shape, the same blank-string handling. **There must remain exactly one allowlist** — `soup_cli.data.formats.VALID_FORMATS` — which #882 pins with a test that monkeypatches a new member in and asserts acceptance with no code change. Keep that property.

While in there, check the other tools in `registry.py` that accept a `format`-like argument for the same `or "auto"` spelling; #882's neighbour sweep covered the validate path, not every tool.

## Acceptance criteria

- [ ] `tool_data_doctor` with `format=""` refuses, matching `tool_data_validate` and the CLI
- [ ] Its unknown-format refusal names the rejected value and the accepted set
- [ ] Every valid format is still accepted, and auto-detection still works when the argument is omitted
- [ ] One allowlist only — a test proves a new member is accepted without touching the tool
- [ ] The error message stays bounded and free of paths, per the `McpToolError` convention
- [ ] A test that fails before the fix

Offered to @Srinivasan8888, who found it and proposed the unification.

Rayan-and-beyond:
## What does this PR do?

Makes MCP `tool_data_doctor` validate its `format` argument the same way `tool_data_validate` does after #882:

- omitted `format` still means auto-detect;
- blank and unknown values are refused;
- the error names the rejected value and the accepted `VALID_FORMATS` set;
- the handler reads `VALID_FORMATS` directly, so new formats do not require a second allowlist change.

I also swept the neighboring MCP handlers for the same `format ... or "auto"` bug. `tool_data_doctor` is the only remaining instance; the export handler's required `format` uses a separate export-format allowlist and is not the same bug.

Closes #889.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests

## Test evidence

- `tests/test_issue889_mcp_data_doctor_format.py`: 9 passed
- new test + #878 regression suite: 28 passed
- existing `TestDataDoctorBranches` + missing-transformers error path: 2 passed
- changelog guard: 17 passed
- `ruff check src/soup_cli/ scripts/ tests/ benchmarks/`: clean
- `python3.12 -m compileall -q src/soup_cli`: clean

Mutation check requested in the issue thread: I restored the old `fmt = _opt_str(args, "format") or "auto"` behavior and reran the new file. Result: **6 failed, 3 passed**. Restoring the fix returns **9 passed**.

The existing `test_returns_report_dict` case in `test_v07128.py` requires the optional `transformers` dependency, which is not installed in this bare local checkout; CI's full dev environment covers that path. The changed validation runs before that unchanged dependency probe.

## Checklist

- [ ] `pytest tests/ -v` passes (full dev/training stack not installed locally)
- [x] No docs update needed
- [x] Added `changelog.d/0.75.0/952.fixed.md`


MakazhanAlpamys:
Reviewed at `80b24579`. **This is a strong first contribution and I am merging it once CI lands.** Your base is already `origin/main`, so no rebase needed.

**The behaviour table, driven through the real handler rather than a helper:**

| input | `origin/main` | this PR |
|---|---|---|
| omitted | accept -> `alpaca` | **accept -> `alpaca`** (the control holds) |
| blank `""` | accept -> `alpaca` | **refuse**, names value + allowlist |
| whitespace `"   "` | accept, passed through as fmt `'   '` | **refuse** |
| unknown `"bogus"` | accept, **reached `run_doctor`** | **refuse** |
| `"ALPACA"` / `"alpaca "` | accept | refuse |

**The baseline was worse than #889 described**, which is worth recording: a bogus format was not merely accepted — it reached `run_doctor` *and came back in the report's `format` field*. Your fragment's "falling through to unrelated tokenizer/doctor errors" is exactly right.

Baseline check: with `origin/main`'s `registry.py` restored and your tests unchanged, **6 failed, 3 passed** — and the three that pass are the controls (omitted-format auto-detect, all-valid-formats, allowlist-extension), which is what should happen.

**Your sweep claim is correct, and I verified it by execution rather than accepting it.** `tool_export`'s `format` and `tool_profile`'s `gpu` both refuse blank, whitespace and unknown already — no `or` default — so `tool_data_doctor` really was the only site carrying that shape. Saying "I checked the neighbours" and being right about it is rarer than it should be.

**The single-allowlist claim holds under test.** I added a member to `VALID_FORMATS` in-memory: the handler accepted it *and* it appeared in the refusal message, with no handler change. Call-time attribute lookup, no copied list.

**Three mutations, all caught** (each anchor counted exactly once, each mutated module really imported and the mutated line re-read via `inspect.getsource`): reverting to `or "auto"` fails 6; dropping the rejected value from the message fails 6; dropping the allowlist from the message fails 1 — specifically `test_error_matches_data_validate_and_names_the_real_allowlist`.

**Message safety passes.** The value is echoed, but through `!r`, which neutralises every hostile shape I tried — ESC, newline, NUL, and a RLO override all come back escaped and literal. `server.py:73` sanitizes again on the way out, so it is defence in depth.

---

**Four things I want to name specifically, because they are the difference between a first PR that needs three rounds and this one:**

1. **`assert message == str(validate_exc.value)`** is the strong form. Instead of re-asserting a string literal that drifts the moment someone rewords it, you pinned the two surfaces *to each other* — which is the actual invariant #889 is about. It is also what caught the third mutation.
2. **You tested the repo's conventions, not just the feature** — `assert str(dataset) not in message` for the path-free `McpToolError` rule, and a length bound. That is a maintainer-level instinct and nobody told you to do it.
3. **Your near-misses are the values an LLM client actually emits** — `"  "`, `"ALPACA"`, `"alpaca "` — rather than only `bogus`.
4. **You deleted the stale cross-reference** in `tool_data_validate` that pointed at the then-unfixed doctor. Cleaning up the comment your own change invalidates is a habit most people skip.

The changelog fragment is right the first time — correct directory, named after the PR, full credit chain including the original reporter. That is usually the thing a first PR misses.

---

**Three LOW notes, none of which I am asking you to change here:**

- Your validation block is a byte-identical copy of `data_validate`'s rather than a shared helper. The invariant #889 pinned — *one allowlist* — is preserved, so this is style rather than correctness. But the repo's culture just moved the other way: #907 merged three commits ago consolidating nine private copies of a sanitizer and adding a CI test that fails if a duplicate reappears. A five-line `_resolve_format(args)` would make a third surface free. Your call whether to do it now or leave it.
- `data_doctor`'s **input schema** was not brought to parity: `registry.py:872-875` still says only "Data format; omit to auto-detect", while `data_validate`'s names example formats and the `auto` contract. Since the clients reading that schema are models, tightening the description is what *prevents* the bad call your fix now diagnoses.
- Cheap arg validation still runs after expensive I/O (`_load_data_rows` precedes the format check), so an invalid format parses the whole file first. It is bounded by `_MAX_DATA_BYTES` and identical to `tool_data_validate`, so you were right not to diverge — noting it as a pre-existing shape.

**And one thing I am filing rather than asking you for:** the second half of #889's argument — "a model client gets stuck instead of self-correcting" — is still true of two neighbours. `registry.py:719` (`unsupported export format`) and `registry.py:432` (`unknown gpu`) both refuse correctly but name neither the rejected value nor the accepted set. That is outside this issue's literal scope, which asked for the `or "auto"` spelling, so it is a separate issue and you are welcome to claim it.

MakazhanAlpamys:
Merged as `cf9bfbcc`. Thank you — and welcome properly.

The defect is the kind worth naming: #882 fixed `tool_data_validate` and left `tool_data_doctor` behind, so **the same malformed argument produced a refusal through one MCP tool and silent misbehaviour through the other**. A client cannot reason about a surface that disagrees with itself, and nothing would have surfaced it except someone checking the sibling.

You matched the existing shape rather than inventing a second one, kept the refusal path-free and free of user input as `McpToolError` requires, and left omitted-means-auto-detect alone.

**The rest of this is owed to you rather than about the code.** You lost #953 to a seven-second claim collision that I created by naming the work in a PR review and filing it publicly without saying either thing in the other place. Then, apologising for that, I offered you #894 and #895 — both already had open PRs I had reviewed as merge-ready that same morning. Then, apologising for *that*, I offered you #880 and #886 — both claimed in comments. Then the fix I announced here, assigning issues, turned out to fail silently for exactly the people it needed to cover.

Four wrong answers to one simple question, and you waited through all of them without complaint. `gh issue list --state open --search "-label:claimed"` is now the honest query, and I check it by reading the result back rather than trusting an exit code.

Your entry is in `CONTRIBUTORS.md` at `b0a63382`. #936, #890 and #891 are still genuinely free if you want them — verified against the label, the assignees and a comment scan, not from memory.

PR #954, opened 2026-09-13
fix(mcp): report rejected choices and live allowed values

### LINKED ISSUE #953: MCP: tool_export and tool_profile refuse without naming the rejected value or the allowlist (the other half of #889)
#889 fixed `tool_data_doctor`'s `format` argument: a blank or unknown value is now refused with a message naming both the rejected value and the accepted set, matching `tool_data_validate` after #882. Two neighbouring MCP handlers refuse correctly but do **not** name either, so the second half of #889's argument still applies to them.

The argument, restated because it is the whole point: the caller is a model. A refusal that says only *"unsupported export format"* gives it nothing to correct toward, so it guesses again or gives up. A refusal that names the value it rejected and the set it would accept lets the caller fix its own call on the next turn without a human in the loop.

**The two sites**, both verified by execution on `main`:

| file:line | handler | current message |
|---|---|---|
| `src/soup_cli/mcp_server/registry.py:719` | `tool_export` `format` | `unsupported export format (see 'soup export --help')` |
| `src/soup_cli/mcp_server/registry.py:432` | `tool_profile` `gpu` (via `_resolve_gpu_memory_mcp`) | `unknown gpu (see 'soup profile --help' ...)` |

Measured:

```
export  format ''       -> REFUSED  "'format' must be a non-empty string"
export  format 'bogus'  -> REFUSED  "unsupported export format (see 'soup export --help')"
profile gpu    ''       -> REFUSED  "unknown gpu (see 'soup profile --help' ...)"
profile gpu    'bogus'  -> REFUSED  "unknown gpu (...)"
```

So the **gating** is already right in both — this is only about what the message tells the caller. Neither carries the `or "auto"` shape that #889 was about, which is why #952 correctly left them alone.

**Proposed fix:** have both name the rejected value (through `!r`, as `tool_data_doctor` does — `repr` neutralises ESC, newline, NUL and bidi overrides, which I verified end-to-end on #952) and the accepted set, read from the live allowlist rather than a copied literal. `SUPPORTED_FORMATS` for export, `GPU_MEMORY`'s keys for profile.

**Acceptance criteria:**

- [ ] `tool_export` refuses an unknown `format` with a message containing the rejected value and the accepted set, read from `SUPPORTED_FORMATS` at call time.
- [ ] `tool_profile` does the same for `gpu` against `GPU_MEMORY`.
- [ ] A test proves the allowlist is not copied: add a member in-memory and confirm the handler accepts it **and** lists it in the refusal.
- [ ] The rejected value is echoed through `repr`, and a test pins that a value containing ESC / newline / NUL comes back escaped rather than literal.
- [ ] Messages stay path-free and bounded, per the existing `McpToolError` convention.
- [ ] The gating behaviour does not change — blank and unknown were already refused and must stay refused.

**Notes for whoever takes this.** A shared helper is the better shape than a third copy: #952's validation block is byte-identical to `tool_data_validate`'s, and the repo just moved the other way in #907, which consolidated nine private copies of a sanitizer and added a CI test that fails if a duplicate reappears. Two more call sites is the moment to extract rather than to copy a fourth time.

Found while reviewing #952 by @Rayan-and-beyond, who verified both neighbours by execution and correctly scoped them out of that PR. Good first issue — the pattern to follow is already in the tree at `registry.py:294-300`.

shubhambhar007:
## What does this PR do?

When MCP export receives an unknown format, or profile receives an unknown GPU, its error now includes the original rejected value and the current accepted choices. For example, `gpu="BOGUS - value"` reports that original spelling rather than only `unknown gpu`.

A shared helper formats these errors and the existing data_validate error. GPU normalization, accepted inputs, blank-value rejection, and the existing 4096-character input bound are preserved. Rejected values use repr so control characters are escaped; model/config paths are not included in these errors.

Closes #953.

## Validation

- Added 17 regression cases; before the fix, 6 failed and 11 controls passed.
- 176 targeted tests passed: `pytest tests/test_issue953_mcp_choices.py tests/test_issue878_mcp_data_validate_format.py tests/test_v07128.py tests/test_mcp_execute.py tests/test_issue322_mcp2_compat.py -q --no-cov`.
- Tests extend both live allowlists and verify acceptance plus refusal-message inclusion; actual MCP client/server round trips verify escaped ESC, newline, NUL, and bidi characters.
- `ruff check src/soup_cli/ scripts/ tests/ benchmarks/` passed.
- Full suite on Python 3.12/macOS: 21,427 passed, 127 skipped, 5 deselected, 3 failed; coverage 83.38%. The changelog failure is fixed in `cba2c65`; the other two are Langfuse colour-output tests (test_issue204 and test_issue885).
- After the review fix: `pytest tests/test_issue487_changelog_fragments.py tests/test_issue953_mcp_choices.py -q --no-cov` — 34 passed; repository-wide Ruff and `git diff --check` passed.

## Type of change
- [x] Bug fix
- [x] Refactor / cleanup
- [x] Tests

## Checklist
- [x] Reviewed relevant docs; no command/API usage changes
- [x] Added a changelog fragment.


MakazhanAlpamys:
Reviewed `2e0d66a0`. **This is a strong first contribution and the work itself is right. One character is blocking it, and the PR is still marked draft.**

### Blocking: the changelog fragment turns CI red in all nine cells

```
pytest tests/test_issue487_changelog_fragments.py -q --no-cov
E  ChangelogError: changelog.d/0.75.0/954.fixed.md:
1 failed, 16 passed
```

Proved it is the sole cause rather than assumed — move that one file aside and the same test gives `17 passed`. Your line 1 begins `MCP export and profile errors now name...`; it needs a leading `- ` (the rule is `scripts/assemble_changelog.py:116`). Directory and filename are both correct.

While you are in there: the fragment carries no credit line. House form is the suffix `(#953 by @shubhambhar007 in #954)` — see `changelog.d/0.75.0/916.fixed.md` or `882.fixed.md` for the shape.

Then mark the PR ready for review, and I will merge on green.

### Everything else verified, and it holds up

- **Scope is exactly #953.** The issue asks for both sites to name the rejected value and the live allowlist, and says outright that "a shared helper is the better shape than a third copy". You added `_unknown_choice` at `registry.py:136-144` and used it at all three sites including `tool_data_validate` — which is the issue's own instruction, not scope creep. Gating is unchanged: blank still hits `_require_str`, oversize still hits the 4096 bound.
- **Baseline is real.** Your tests against `main`'s `registry.py` give `6 failed, 11 passed`, matching your PR body exactly. On your head, 17 passed; neighbouring MCP suites 164 passed.
- **Mutation caught.** Replacing the helper body with a message that drops both the value and the choices — module still imports, verified — gives 6 failed, all behavioural.
- **Security is right, and for the right reason.** `{value!r}` escapes ESC, newline, NUL and bidi, `server.py:73` additionally runs `_sanitize(str(exc))`, and `registry.py:78` delegates to the shared `utils.terminal.strip_control` rather than reintroducing a private copy — `test_issue896_no_duplicate_terminal_sanitizer.py` passes. Messages stay path-free and bounded (223 and 105 characters measured). `repr` does not escape Rich markup, but this text goes over JSON-RPC and never reaches a Rich console, so `strip_control` is the correct half of that pair here.
- No tests deleted, no ANSI exposure (assertions are on `str(exc.value)` and MCP `TextContent`, neither Rich-rendered), docs correctly untouched since the MCP error text is not documented anywhere, ruff clean.

One cosmetic note, no action needed: `data_validate` now sorts `auto` into the list rather than leading with it. `test_issue878` still passes.

### On the claim collision

You got this issue, and you got it because you were eight minutes from a working patch when you asked. I have said the rest on #953 — the collision was my doing, not yours or @Rayan-and-beyond's.

For a first PR the test file is above the median here: you stated baseline counts *before* the fix rather than after, wrote eleven explicit controls, added live-allowlist extension tests, and included a round-trip escaping test. That is the discipline that makes a review short. The two misses are both mechanical — the `- ` and the credit suffix — and both are in CONTRIBUTING.md.

shubhambhar007:
Fixed in cba2c65: the changelog fragment now starts with `- ` and includes `(#953 by @shubhambhar007 in #954)`. The changelog suite and issue-specific tests pass together (34 passed), repository-wide Ruff passes, and the PR is now ready for review.

I also updated the validation notes with the completed full run: 21,427 passed, 127 skipped, and three failures. The fragment failure is now fixed. The two remaining Langfuse colour-output failures reproduce unchanged on the original base commit fcca8ef, verified in a separate worktree with its source on PYTHONPATH.

Thanks for the detailed review and for clarifying the issue assignment.

MakazhanAlpamys:
Verified `cba2c659`. **Merge-ready** — the fragment now reads:

```
- MCP export and profile errors now name the rejected format or GPU and list the
current accepted values. Error formatting is shared with data validation, and
control characters in rejected values are escaped (#953 by @shubhambhar007 in #954).
```

Leading `- `, credit in house form, correct directory. That was the only blocker, and the PR is out of draft.

**On your three remaining failures — your reading is right and worth stating plainly, because it is the harder half of a good bug report.** You reproduced the two Langfuse colour-output failures on the original base commit `fcca8ef` in a separate worktree with its source on `PYTHONPATH`, i.e. you proved they are not yours before claiming it. That is the correct method and it is not the common one.

Those two are tracked: **#886**, "v0.75.0 added three tests that fail whenever colour is on". This project has had its CI turned red four separate times by exactly that class — Rich emits SGR escapes *inside* a flag name on a colour-capable runner, and `ReprHighlighter` also colours quoted strings and numbers inside error messages, so a raw `in result.output` assertion passes on Windows (where Rich auto-disables colour) and fails on Linux and macOS. Your own tests avoid it, which is why I did not raise it in review.

Nothing further needed from you. I will merge once CI is green; the Actions queue is deep, so it will be a while.

MakazhanAlpamys:
@shubhambhar007 — **two things, and the first is mine to apologise for.** I said "Verified `cba2c659`. Merge-ready" on 09-14 and then went quiet. Since then you pushed `c7ad398e`, and **that head is red on all nine cells** — it has been since this morning and I did not tell you.

**It is one test, and it is an ordering difference, not a logic error:**

```
tests/test_issue889_mcp_data_doctor_format.py::TestInvalidFormatIsRefused::test_error_matches_data_validate_and_names_the_real_allowlist

- unknown format 'bogus'; accepted: alpaca, asr, audio, auto, chatml, dpo, ...
+ unknown format 'bogus'; accepted: auto, alpaca, asr, audio, chatml, dpo, ...
                                    ^^^^
```

The test expects the allowlist sorted alphabetically; the code now emits `auto` first and the rest sorted. Same members, different order.

**Which of the two is right is a real question, not a formality**, so I am not going to tell you to just change the assertion. The test's own name is `test_error_matches_data_validate_and_names_the_real_allowlist` — its whole point is that the MCP tool and `soup data validate` say the *same thing*. So the right move is to check what `soup data validate` prints for a bogus format and make the MCP tool match **that**, whichever order it uses. If `data validate` leads with `auto`, the test is stale; if it sorts plainly, your normaliser is what moved.

If they genuinely disagree, that is a finding worth its own line in the PR rather than a test edit.

**Also worth knowing:** my merge-ready verdict was recorded against `cba2c659` and your head is now `c7ad398e` ("adapt live GPU choice check to normalizer"). I have not reviewed that delta yet — I will do it as part of the same pass once the matrix is green, and I will review only the delta rather than starting over.

The queue was badly backed up and is now clear, so turnaround should be short.

shubhambhar007:
Thanks for flagging this, and no worries about the timing. You were right to ask which order the CLI actually uses. I checked `soup data validate --format bogus`: it prints `auto` first, then the formats in the declared `VALID_FORMATS` order—not alphabetical order. The two MCP tools had diverged from each other, so I updated the shared formatter to preserve that CLI order for data formats while leaving export/GPU choices sorted.

The fix is pushed as `6cb2a8f`. I added an assertion for the live order, reproduced the failing test before the change, and ran the relevant MCP suites afterward: 205 passed locally. The new Actions run is awaiting maintainer approval before the matrix can run. Thanks again for the careful review of the delta.

MakazhanAlpamys:
@shubhambhar007 — **your push at 13:55Z had no CI running: the workflow was stuck awaiting approval.** I have approved run `34978159102` and all thirteen required contexts are now going.

That is on me, and the reason I nearly missed it is worth writing down: `gh run list --json status` reports a blocked run as `status=completed, conclusion=action_required`, so filtering on `status == "action_required"` returns nothing and the queue looks clear. `gh api "repos/OWNER/REPO/actions/runs?status=action_required"` is the query that finds them. There were eight; yours and one other were on open PRs.

I will review the delta once the matrix reports, and I will review only the delta — your previous head already has a recorded verdict, so this is not going to come back as a fresh wall of text.

MakazhanAlpamys:

Sorry for the noise.

MakazhanAlpamys:

MakazhanAlpamys:

I wrote:

```
```

## What actually clears it

## What has not changed

shubhambhar007:
@MakazhanAlpamys could you please update branch ,I'm afk for a while, thanks!

MakazhanAlpamys:
Merged as 3dece260. Thank you, @shubhambhar007, and sorry the delta review came late. I had promised it once the matrix reported, and it slipped behind the frozen-merge-ref mess.

Here is that review. You checked the real CLI output before changing either side. `soup data validate --format bogus` prints `auto` first and then the declared `VALID_FORMATS` order, and both MCP tools now match it exactly. Export formats and GPU names you kept sorted instead of changing everything. The order check sits in `test_issue889` itself, so the test that says it matches `data validate` now actually checks that. The `NEW-GPU` to `NEW-CARD` rename is also correct: the GPU key normaliser strips "gpu", so `NEW-GPU` normalised to `new` and was never testing an unknown card.

Checked on the tree merged with current `main`: 26 passed, plus 253 across the MCP neighbours. Removing the value and choices from the error fails 13 tests. Always sorting fails your order assertion.

PR #955, opened 2026-09-13
fix(config): reject non-functional kernel auto-compose

### LINKED ISSUE #801: kernel_auto_compose times one unchanged model four times, prints the fastest name, and applies nothing
## The defect

`training.kernel_auto_compose: true` is documented as benchmarking kernel combinations "and adopts the fastest" (`docs/performance-and-quantization.md:148`). What the code does:

1. `enumerate_kernel_combos` lists candidates such as `baseline`, `liger`, `flash`, `liger+flash`. Each is a dict of flags (`use_liger`, `use_flash_attn`, `use_cut_ce`).

2. `benchmark_kernel_combos` times **the same already-loaded model on the same input once per candidate name**. The candidate's flags are never applied before timing it.

3. `pick_best_kernel` returns the lowest time, and `apply_v028_speed_memory` prints `Kernel auto-compose picked: <name>`, sets `applied["kernel_auto_compose"] = True`, and returns. Nothing reads the pick.

The ranking is therefore timing noise between identical runs, and the "winner" is only printed.

## Evidence

The benchmark says so itself, `src/soup_cli/utils/kernel_picker.py:143-147`:

```

The benchmark intentionally does NOT swap kernels mid-loop (that would

require model re-instantiation per candidate which is too expensive on

CI-sized models). Instead we run identical forward+backward across

candidates and record the relative ordering

```

`kernel_picker.py:199-200`, where every candidate gets the same arguments:

```python

for entry in out:

    entry["time_ms"] = _time_one_combo(model, input_ids, steps=steps)

```

(The docstring also says "forward+backward"; `_time_one_combo` at 204-244 runs forward only, under `torch.no_grad()`.)

`src/soup_cli/utils/v028_features.py:112-123` consumes only the name:

```python

picked_name = _bench_and_pick_kernel(model=model, device=device, backend=backend)

...

applied["kernel_auto_compose"] = True

_say(f"Kernel auto-compose picked: {picked_name}")

```

No caller in `src/` reads the returned `applied` dict's `kernel_auto_compose` entry, `tcfg.use_liger` / `use_flash_attn` / `use_cut_ce` are not modified, and the call happens after the model is loaded, when `attn_implementation` can no longer change. The helper also runs only in the 11 non-SFT trainers (SFT never calls it, filed separately), and `use_liger` / `use_flash_attn` are read only by SFT, so the flags the picker would set have no consumer on the trainers where it runs.

## Reproduction

CPU only. `torch.cuda.is_available`, the kernel availability probes and `_time_one_combo` are stubbed; the timer returns the same base time plus Gaussian noise (sigma 0.2 ms), so the only difference between candidates is noise, which is what the real benchmark measures too:

```

candidates: ['baseline', 'liger', 'flash', 'liger+flash']

calls to the timer: 4 | distinct (model id, input shape): 1

10 picks on an unchanged model: ['liger', 'liger+flash', 'flash', 'liger+flash', 'liger+flash', 'flash', 'flash', 'baseline', 'flash', 'liger+flash']

applied dict: {'cut_ce': False, 'fp8': False, 'kernel_auto_compose': True}

model attn impl after 'pick': sdpa | tcfg.use_liger: False | tcfg.use_flash_attn: False

```

## Why it matters

The console line `Kernel auto-compose picked: liger+flash` reads as a decision that took effect. A user who sees it has no reason to set `use_liger` / `use_flash_attn` themselves, so the feature leaves them with neither kernel while reporting one. The #45 closing note ("feeds real measurements into `pick_best_kernel`") is accurate about the measurement and silent about the fact that nothing is adopted.

## Proposed fix

Choose one; both are honest.

- **Make it real.** Run the picker before model load: for each candidate, load (or reconfigure) the model with that combo's `attn_implementation` and Liger patch, time a few forward+backward steps on a synthetic batch, then load the real model with the winner and write the winner into the effective config so the run record shows it. This costs a model load per candidate, which is the cost the docstring declined. It only makes sense where the consuming flags are read (today, SFT's text path).

- **Stop claiming it.** Reject `kernel_auto_compose: true` at config load with a message saying the picker does not change kernels and pointing at `use_liger` / `use_flash_attn`, following the `packing_cross_doc_attn_mask` precedent (`schema.py:3492`). Correct the docs line.

Until one of these lands, the printed line should at least say the result is advisory and not applied.

## Acceptance criteria

- [ ] A test with a stubbed timer that favours a non-baseline candidate asserts that the winning combo's kernel is actually active on the model that trains (or that config load rejects the flag). This fails on current code, where the model is unchanged.

- [ ] Each candidate is timed with its own flags applied (a test records the flags active during each timing call), or the benchmark is removed.

- [ ] `docs/performance-and-quantization.md` describes what the flag does after the change.

- [ ] The `kernel_picker.py` docstring no longer says "forward+backward" for a forward-only timer.

---

- #800 — training.nvfp4 and fp8_attention are never applied on task: sft: the shared helper runs in 11 trainers and ...
- #816 — soup serve --auto-quant picks by timer noise (gguf 97 / awq 43 / gptq 26 / fp8 21 / none 13 of 200) and for...

Amix29:
Closes #801

## What changed
- reject `training.kernel_auto_compose: true` during Pydantic config validation with an actionable message
- prevent internal callers that bypass validation from reporting a kernel selection as applied
- correct the public docs and legacy benchmark docstrings; point users to explicit `use_liger` / `use_flash_attn` flags and the MLX path on Apple Silicon
- update historical coverage and add a focused regression test for default/false behavior and the refusal message

## Verification
- `322 passed` — issue/config/training-speed/v0.35 coverage selection
- `102 passed` — config + performance + #801 regression selection
- `git diff --check`

I also started the full suite. It reached 2,582 passes before I stopped after one unrelated control-test failure: `tests/test_issue204_langfuse_pull.py::test_control_the_leak_scan_sees_a_token_rich_highlighted_and_folded`. The exact failure reproduces on a clean detached `origin/main` worktree with the same environment, so it is not introduced by this PR.

No CUDA execution was used; the behavior is enforced and tested at config/runtime-guard level on CPU.


MakazhanAlpamys:
Reviewed the current head. **The fix is sound and the evidence for it is the strongest part. Two things stand between it and merge, and only one is yours.**

### Blocking (yours, one character): the fragment turns CI red in all nine cells

```
pytest tests/test_issue487_changelog_fragments.py -q --no-cov
E  ChangelogError: changelog.d/0.75.0/955.fixed.md:
```

Confirmed as the sole cause — moving that one file aside gives `17 passed`. Your line 1 begins ``Reject `training.kernel_auto_compose: true`...``; it needs a leading `- `, plus the credit suffix `(#801 by @Amix29 in #955)`.

Note for ordering, since #954 has the identical defect: if both land, **either one's fragment alone keeps that test red**, so whichever merges second still sees a red suite until both are fixed.

### Not blocking you: this changes default behaviour, so I want it decided rather than merged

A config that loads on v0.75.0 now fails at parse time. I measured it rather than reasoned about it:

```
training: {kernel_auto_compose: true}
-> Config validation error:
   training: Value error, kernel_auto_compose is not supported: it benchmarks
   the same already-loaded model for every candidate and does not apply the
   selected flags. ...
```

**Your evidence that the blast radius is empty is what makes this defensible**, and I reproduced the whole grep:

- **Zero** hits in `src/soup_cli/recipes/`, **zero** in `templates/`, **zero** in `examples/`.
- The only shipped-config trace is `tests/fixtures/recipe_config_snapshots.json:133` recording `false` — so all 167 recipes dump it as `false` and the snapshot test is unaffected.
- No CLI flag produces it (`grep -rn "kernel-auto-compose" src/soup_cli/commands/` is empty), and no migrate or autopilot path emits it.
- `model_dump()` emits `false`, so re-validating an existing dumped config still works — **no repeat of the #321 round-trip footgun**, which is the specific way this class of change has bitten us before. That check is the one I would have asked for if you had not done it.

So the break reaches only a hand-written config that sets a flag which has never done anything. That is a good trade and I am inclined to take it. I am confirming the call before merging anyway, because it is a schema-level refusal and #801 explicitly offered two honest options. I will come back with the answer — nothing further needed from you on that half.

### The rest, verified

- **Option 2 of the two #801 offered**, following the `packing_cross_doc_attn_mask` precedent at `schema.py:3492`, with the defensive runtime guard at `v028_features.py:110-121`. Both remaining acceptance boxes cleared: docs rewritten, and the false "forward+backward" docstring is gone.
- **Mutations caught on both halves.** Disabling the schema validator gives 11 failed across eight task parametrizations; restoring the old picker call plus `applied[...] = True` gives 12 failed including `..._refuses_kernel_picker` and `..._never_claims_application`. Both modules import under mutation, so these are behavioural.
- **Baseline real:** `test_kernel_auto_compose_refuses_false_benchmark_claim` does not raise against `main`'s three source files.
- **Every removed test line is an inversion, not a weakening** — `..._invokes_kernel_picker` becomes `..._refuses_kernel_picker` with `picker.assert_not_called()`, which is strictly stronger. One genuine retirement: `..._degrades_on_bench_failure`, whose production branch no longer exists. Fine.
- 303 passed on the three primary files, 350 on the adjacent ones, and both the #748 consumer ratchet and the #755 doctor table still pass — the field now reaches a validator rather than nothing, which is what the ratchet wants.

One follow-up worth filing rather than fixing here: with `_reject_kernel_auto_compose` firing on `TrainingConfig`, which validates before the `SoupConfig` gate, the `kernel_auto_compose` offenders at `schema.py:4334-4335` and `v028_features.py:234-235` are now unreachable, and so are `enumerate_kernel_combos` / `benchmark_kernel_combos` / `_bench_and_pick_kernel` from production. You retained them deliberately and #801 permits that — I would rather it be a tracked cleanup than a quiet one.

Amix29:
Update pushed in ea15599 after the latest review. The changelog fragment now uses the required canonical issue/author/PR format. The branch is rebased on current main and GitHub reports it mergeable. Validation: 34 targeted tests passed, including the changelog validator, the #801 regression tests, config picker coverage, and schema rejection coverage; Ruff is clean on the touched Python files. I also opened #976 for the separate unreachable kernel auto-compose cleanup requested in review. Ready for re-review.

MakazhanAlpamys:
@Amix29 — **the decision is yours to have: refuse at parse, as your PR does.** I said I would come back with it rather than leave you waiting on a half-answer, so here it is with the measurement behind it.

## The behaviour call

`training.kernel_auto_compose` was validated, documented, and read by nothing. #801 offered implementing it or refusing it; refusing is the answer, and it is consistent with v0.75.0 already making an *unknown* key a hard error — a key that is known and inert is not a better experience than one that is unrecognised, it is a worse one, because the user has a written promise that it does something.

Blast radius re-measured on the merged tree at today's `main`, not quoted from the earlier review:

```
recipes setting kernel_auto_compose : 0 of 171
templates                           : 0
examples/configs                    : 0
```

So nothing shipped stops loading. A user's own `soup.yaml` carrying the flag now fails at parse with a message naming `use_liger` / `use_flash_attn`, and I checked that the suggested replacement config then loads — a refusal that names an unusable remedy would be worse than the silence.

## Your side is complete

The blocking item is closed. I checked the bytes rather than the message, since that is how this one went wrong the first time:

```
$ git show <head>:changelog.d/0.75.0/955.fixed.md | head -c16 | xxd
00000000: 2d20 5265 6a65 6374 2060 7472 6169 6e69  - Reject `traini
```

A real `- `, no BOM, and `(#801 by @Amix29 in #955)` present.

**The delta is that one line.** I isolated it properly — both PRs were rebased, so a raw `diff reviewed..head` is 130 files of main's own churn — by diffing each end's patch against its own merge base and then diffing the two patches. One content line changed. Nothing else.

**Zero removed or weakened test lines.** The `tests/` half of the patch is byte-identical at both ends, including the inversions I approved earlier (`..._refuses_kernel_picker` with `picker.assert_not_called()`).

**Mutation, on the merged tree.** `schema.py` anchor `if self.kernel_auto_compose:` counted exactly once — there are 7 occurrences of the name, only one of that spelling — replaced with `if False:`, module genuinely re-imported afterwards:

```
11 failed, 292 passed
  Failed: DID NOT RAISE ValidationError
  test_schema_rejects_kernel_auto_compose[...]  (8 parametrisations)
```

Reverted → 303 passed, tree clean. 765 passed, 3 skipped across your files plus every file touching `kernel_picker` / `v028_features`, identical on the branch and the merged tree. `ruff` clean.

And you filed **#976** for the now-unreachable `enumerate_kernel_combos` / `benchmark_kernel_combos` / `_bench_and_pick_kernel` rather than deleting them inside this PR. That was the right split: a tracked cleanup instead of a quiet one, which is exactly what I asked for.

---

Nothing needed from you. I will merge on the next green matrix and will not re-review — the verdict is recorded against your current head. The Actions backlog that stalled today is cleared (104 live runs down to ~57), so this should be a short wait.

Your other four PRs: #956 and #931 and #945 are all merge-ready on my side and waiting on the same queue; #915 has one item, which I left on that thread. That is four of five clear — thank you for the volume *and* for taking the rebases without complaint, which is the part that makes the volume manageable at this end.

MakazhanAlpamys:
Merged as `b34f3b6b`. Thank you — and the decision you were waiting on is the one your PR already implemented: refuse at parse.

**Why refusing was right.** `training.kernel_auto_compose` was validated, documented, and read by nothing. v0.75.0 already made an *unknown* key a hard error, and a key that is **known and inert** is a worse experience than one that is unrecognised — the user has a written promise that it does something. Blast radius measured on the merged tree rather than asserted: **0 of 171 recipes, 0 templates, 0 examples** set it, and the config the error suggests instead was confirmed to load, because a refusal naming an unusable remedy is worse than silence.

The mutation: `if self.kernel_auto_compose:` counted exactly once — there are seven occurrences of the name and only one of that spelling — replaced with `if False:`, module genuinely re-imported. 11 failed, including `DID NOT RAISE ValidationError` and all eight parametrisations of `test_schema_rejects_kernel_auto_compose`. Reverted, 303 passed.

**And you filed #976 rather than deleting the now-unreachable helpers inside this PR.** That was the right split and it is the half most people skip: a tracked cleanup instead of a quiet one.

Credited in `CONTRIBUTORS.md` (`f491ac5b`). #801 closed with credit.

PR #957, opened 2026-09-13
docs(readme): polish Turkish translation for v0.75.0 release

### LINKED ISSUE #924: README.tr.md: polish the v0.75.0 What's New translation (machine-translated by the release)
@Ercaner1988 — the v0.75.0 release replaced the README `## What's New` block and the unknown-config-key note (#879), which meant `README.tr.md` had to be re-synced before the tag. I did the Turkish for the changed sections myself so the sync gate from #852 stayed green, and re-stamped the file. The structure passes the gate (code spans, links and fences match section by section), but the wording is a machine translation, not a native one.

Since the translation is yours, could you polish those sections when you have a minute? Only prose changed; nothing structural needs touching, and the stamp only has to move if README.md changes again.

Sections: `## What's New` (v0.75.0 block) and the "Unknown config keys are rejected since v0.75" note. Tracking here so it does not get lost; no urgency.

ege-arhan:
## Problem
The v0.75.0 release introduced changes to the `What's New` section and the unknown config key notes. The Turkish translation in `README.tr.md` was machine-translated during release synchronization, resulting in slightly robotic phrasing.

## Solution
Polished the Turkish phrasing in the `## Yenilikler` section and the unknown configuration key note for natural native flow, while preserving the exact technical structure, numbers, and links.

## Impact
Improves readability and native phrasing for Turkish documentation readers without breaking translation sync guarantees.

## Evidence
Ran `pytest tests/test_readme_translation_sync.py` — all 23 tests passed.

Fixes #924


MakazhanAlpamys:
Reviewed `cb8b8ff5`. **Merge-ready** — and welcome; this is your first PR here.

The thing I check hardest on a translation PR is the `synced-from` stamp, because a **stale stamp that happens to pass** is the failure mode this gate exists to catch and the green tick cannot distinguish it. So I computed it independently rather than trusting the test:

```
origin/main README.md sha256 (LF-normalised): 5ceff359...eb1c4
PR-branch  README.md sha256 (LF-normalised): 5ceff359...eb1c4
stamp in README.tr.md:              sha256:5ceff359...eb1c4
```

All three identical, and `README.md` itself is untouched — one file, one commit — so no other translation has to move. `tests/test_readme_translation_sync.py` gives 23 passed, the wider docs sweep 57.

I also diffed both changed regions against `README.md:95-102` and `README.md:250-255` line by line. Every version number (`v0.74`, `v0.75`, `v0.40.1`), every inline-code span, the exit code, and the relative link `docs/backends-and-ops.md#unknown-config-keys` are preserved byte-for-byte. No placeholders, no untranslated stubs, markdown structure intact.

**Two of your changes are accuracy improvements rather than polish, and they are the reason I am glad you sent this:**

- `README.tr.md:99` — "üst düzey `lora:`" becomes "**kök düzeyindeki** `lora:`". The English is *root-level*, and `üst düzey` reads as top-level or high-calibre, which is ambiguous exactly where precision matters. Your wording matches the codebase's own term of art — `ROOT_LEVEL_MISPLACED_KEYS`, `remap_root_level_misplaced_keys`.
- `README.tr.md:252` — "bildirmediği" (reports) becomes "tanımlamadığı" (declares), which is what the English says and what the mechanism actually is: a **declared** table, not an inferred one. That distinction is the whole point of #755, so the original wording quietly described a different feature.

**One wording note, entirely your call — I am not holding the merge for it.** `README.tr.md:102` changes "kaçışlanıyor" (*escaped*) to "filtreleniyor" (*filtered*). The English says key names are *escaped* before reaching the terminal, and `utils/terminal.py:34-36` does both halves:

```python
def for_terminal(text: object) -> str:
    """Strip control bytes, then escape Rich markup, in that order."""
    return _escape_markup(strip_control(text))
```

"filtreleniyor" describes `strip_control` and drops `escape_markup` — which is the Rich-markup-injection half the English word names. Not false, and "kaçışlanıyor" is an awkward calque, so this is readability against precision rather than an error. The structural gate cannot see prose, which is why I am raising it here rather than letting a test do it. If you would rather keep the smoother reading, say so and it merges as is.

No changelog fragment, and I think that is right: `assemble_changelog.py` has no docs category, and the original Turkish README got an entry because it added a file and a CI gate. A prose polish with zero structural change has no natural home there.

One thing you should know, since it affects you rather than this PR: **@Ercaner1988 wrote and owns `README.tr.md`** (#852), and the structural gate means every future English README edit needs this file re-translated and re-stamped. If you want to keep helping with that, say so here — it is real recurring work and it currently has one person on it.

Merging on green; the Actions queue is at ~88 runs, so it will be slow.

MakazhanAlpamys:
Reviewed `2b6a54f9`. **Merge-ready.** I have approved your workflow runs — they were sitting at `action_required`, which is a fork-permissions artifact and not a signal about your work; that approval does not carry across pushes, so re-approving is on me each time, not on you.

One thing worth stating rather than assuming: the 14:59Z push is a **pure re-commit** — identical tree, identical parent, new timestamp only (`tree=ca927eff…` on both `cb8b8ff5` and `2b6a54f9`, `git diff` empty). So the approval I gave the earlier head carries over on content; I am saying so out loud because my merge-ready verdict names a SHA that is no longer the head.

**The sync gate is green and I recomputed the stamp rather than trusting it:**

```
merged-tree README.md sha256 (LF-normalised): 5ceff359925678608a55dc8a4d522221b75668b10e659aa4f0dd05a28f2eb1c4
stamp in README.tr.md                       : 5ceff359925678608a55dc8a4d522221b75668b10e659aa4f0dd05a28f2eb1c4
```

`README.md` has not moved on `main` since your base, so the stamp is current rather than merely matching an old blob.

**No technical content was altered by the translation.** I extracted the token sets from the removed and added lines and compared them as multisets: inline code identical (6, including `ValueError`, `lora:`, `soup fetch examples`, `training.lora`), URLs identical, markdown links identical, version strings identical (`v0.40.1`, `v0.74`, `v0.75`), fences identical, digits identical. The exit code and the `docs/backends-and-ops.md#unknown-config-keys` link sit on context lines and were never touched. Only Turkish prose changed, which is exactly what a polish pass should look like.

Three mutations on the guard, all caught: a fenced-code language `yaml`→`yml`, an external URL broken, and an inline code string changed by one character — each fails `test_every_translation_mirrors_readme_structure` naming the specific mismatch. 23 passed on branch and merged tree; 262 passed, 2 skipped on the wider README sweep; 44 passed against the newest `main` with the stamp still current.

The `filtreleniyor` / `kaçışlanıyor` wording note stands as it did: entirely your call, I am not holding the merge for it. If you prefer the smoother reading, it merges as is.

And the question from last time is still open if you want it: @Ercaner1988 wrote and owns `README.tr.md` (#852). If you would like to keep helping with that file, say so here — it is worth having a second pair of eyes on a translation that is now structurally gated.

MakazhanAlpamys:
Merged as `0e92671f`. Thank you, and welcome — this is your first contribution here.

**For a translation the prose is not the reviewable part; the technical tokens are.** So rather than judging your Turkish, I extracted the token sets from the removed and added lines and compared them as multisets: inline code identical (6, including `ValueError`, `lora:`, `soup fetch examples`), URLs identical, markdown links identical, version strings identical, fences identical, digits identical. The exit code and the relative docs link sit on context lines and were never touched. **Only prose changed**, which is exactly what a polish pass should look like.

The sync gate is structural here — every `README.*.md` carries a sha256 stamp of `README.md` — so I recomputed it rather than trusting the check: the merged tree's LF-normalised `README.md` hashes to exactly the `synced-from` stamp in your file, and `README.md` had not moved on main since your base, so the stamp is *current* rather than merely matching an old blob.

Three mutations on the guard, all caught: a fenced-code language changed, an external URL broken, and an inline code string changed by one character — each failing `test_every_translation_mirrors_readme_structure` naming the specific mismatch.

One note on the 14:59Z push: it was a pure re-commit — identical tree, identical parent — so the verdict I had recorded against the earlier SHA carried over on content. I said that out loud rather than assuming it.

The `filtreleniyor` / `kaçışlanıyor` wording question was yours to decide and it merged as you left it. And the standing offer holds: @Ercaner1988 wrote and owns `README.tr.md` (#852); if you want to keep helping with it, say so — a structurally-gated translation benefits from a second pair of eyes.

Credited in `CONTRIBUTORS.md` (`f491ac5b`); you are the 65th. #924 closed with credit.

PR #961, opened 2026-09-14
feat(mlx): widen sft table to remaining four CUDA-only fields

### LINKED ISSUE #958: doctor --config: widen the MLX table to the remaining nine CUDA-only fields
> **CORRECTION, 2026-09-15 — the ask below was wrong and this paragraph supersedes it.**
> I filed this asking for **all nine** fields and `None of the 18`. Five of the nine
> **cannot be loaded on `backend: mlx` at all** — the config is refused before
> `doctor --config` ever sees it — so a table row for them is a row nobody can reach.
> Verified by loading each of the nine through the real loader:
>
> | field | loads on `backend: mlx`? | belongs in the table |
> |---|---|---|
> | `training.use_mod` | parses | **yes** |
> | `training.moe_lora` | parses | **yes** |
> | `training.quantization_aware` | parses | **yes** |
> | `training.use_fsdp2_compile` | parses | **yes** |
> | `training.use_cut_ce` | REFUSED — v0.28.0 features not supported on mlx | no |
> | `training.kernel_auto_compose` | REFUSED — same gate | no |
> | `training.use_longlora` | REFUSED — LongLoRA is not supported on mlx | no |
> | `training.activation_offloading` | REFUSED — v0.28.0 gate | no |
> | `training.packing_cross_doc_attn_mask` | REFUSED — never mapped to a valid TRL packing strategy | no |
>
> **So the ask is four fields, and the count goes 9 -> 13, not 9 -> 18.** The
> acceptance boxes at the bottom are rewritten accordingly; the original text is
> kept below the line so the history reads honestly. #961 implements the corrected
> scope. Anyone arriving here from #965 or elsewhere should work from the corrected
> boxes, not the original ones.

---

`soup doctor --config` now scopes its all-clear to the fields it actually knows about (#911, fixing #903):

```
None of the 9 setting(s) known to be unread on task=sft backend=mlx is set in this config.
```

That sentence is honest, but the number is only as strong as the table behind it, and the table is currently 9 entries wide on the one reviewed pair. #903 named eleven CUDA-only fields; #911 added two of them, deliberately and in the open. The others are listed in the correction table above.

All of them are real `TrainingConfig` fields with **zero hits across the MLX modules** (`trainer/mlx_sft.py`, `trainer/mlx_optim.py`, `trainer/mlx_masking.py`), so any of them would pass `test_issue755_backend_support_registry.py`'s bidirectional guard — but passing that guard is not sufficient, because a field the schema refuses on this backend can never reach the report.

### Why this is worth doing

`None of the 9` reads stronger than 9-of-20 warrants. Widening the table is the only thing that makes the count mean what a reader takes it to mean, and every reachable entry is cheap: one `SupportEntry`, one `_check_unsupported` branch, one grep to prove zero hits.

### Two notes carried over from the #911 review

1. **`use_liger` is not purely "ignored".** `commands/train.py:1193` calls `validate_liger_config`, which errors whenever `device != "cuda"` and exits 1 — so on a real Apple Silicon box `soup train` refuses the run before MLX ever sees the field. `IGNORED` is still defensible under this module's definitions (MLX genuinely never reads it, and `doctor --config` does not run that validation), but a clause in the reason string would make the report match what the user will actually experience. `neftune_alpha` is the pure case.

   **Correction, 2026-09-15:** I later extended this note to `quantization_aware` as well. That was wrong — `validate_qat_config` has **no mlx branch at all** (it refuses `unsloth` only), so the clause would be false there. It applies to `use_fsdp2_compile`, where `validate_fsdp2_compile_config(..., backend="mlx")` refuses unconditionally and `commands/train.py:1186` exits before `resolve_trainer`.

2. **A fixture in `tests/test_issue903_*` depends on its own literal indentation.** The `"  use_liger: true\n              neftune_alpha: 5"` string parses correctly today — I reproduced it, `model_fields_set == ['neftune_alpha', 'use_liger']` — but only because the 14 spaces survive `dedent`. Any future edit there breaks it into invalid YAML silently rather than failing loudly. `"  use_liger: true\n  neftune_alpha: 5"` is equivalent and cannot break that way.

### Acceptance (corrected 2026-09-15)

- [ ] The **four reachable** fields present in `_MLX_SFT` with a reason string each, every one justified by a zero-hit grep stated in the PR: `use_mod`, `moe_lora`, `quantization_aware`, `use_fsdp2_compile`
- [ ] The five unreachable fields explicitly **left out**, and the PR says why — a row for a config the schema refuses is a row nobody can reach
- [ ] Matching `trainer/mlx_sft.py::_check_unsupported` branches so the trainer warns too, not only the doctor
- [ ] `soup doctor --config` on a clean mlx config reports `None of the 13 setting(s) ...`
- [ ] A test that sets several of the new fields at once and asserts the table rows and the count
- [ ] The two notes above addressed or explicitly declined with a reason, taking note of the correction to note 1

<details>
<summary>Original acceptance criteria, superseded — kept for the record</summary>

- [ ] All nine fields present in `_MLX_SFT` with a reason string each, every one justified by a zero-hit grep stated in the PR
- [ ] Matching `trainer/mlx_sft.py::_check_unsupported` branches so the trainer warns too, not only the doctor
- [ ] `soup doctor --config` on a clean mlx config reports `None of the 18 setting(s) ...`
- [ ] A test that sets several of the new fields at once and asserts the table rows and the count
- [ ] The two notes above addressed or explicitly declined with a reason

</details>

ege-arhan:
## Problem
`soup doctor --config` scopes its all-clear to fields it knows about (#911 fixing #903), but the table behind it is 9 entries wide on the reviewed pair while #903 named eleven CUDA-only fields and #911 added two. The other four are real `TrainingConfig` fields with zero hits across the three MLX modules, so `None of the 9` reads stronger than 9-of-13 warrants.

Fixes #958

## Solution
- `_MLX_SFT` gains four: `use_mod`, `moe_lora`, `quantization_aware`, `use_fsdp2_compile` (9 -> 13 entries), each with a reason string
- Matching `trainer/mlx_sft.py::_check_unsupported` branches so the trainer warns too, not only the doctor
- Two review notes from #958 carried over: `use_liger` reason now notes the non-CUDA `soup train` refusal; the #903 fixture indentation hardened to 2 spaces so it cannot break silently
- `use_fsdp2_compile` reason notes the pre-trainer refusal (`validate_fsdp2_compile_config` refuses backend=mlx at `commands/train.py:1235` before `resolve_trainer`)
- `soup doctor --config` on a clean mlx config now reports `None of the 13 setting(s) ...`

## Impact
- The count means what a reader takes it to mean; every new entry is cheap and guarded
- Bidirectional guard (`test_issue755_*`) covers each entry automatically via `trainer_reads=True` parametrization

## Evidence
- Zero-hit grep stated: each of the four returns 0 across `trainer/mlx_sft.py`, `trainer/mlx_optim.py`, `trainer/mlx_masking.py`
- AST syntax check on both edited source files: OK
- Diff: 72 insertions, 1 deletion / 4 files + fragment credit (#958 by @ege-arhan in #961)
- Rebased onto upstream/main @ f491ac5 (includes #966 trailing \b + model_fields cross-check); tests: test_issue755_backend_support_registry.py 50 passed, test_mlx_backend.py 29 passed


MakazhanAlpamys:
@ege-arhan — received, and **your CI was blocked rather than slow.** A first-time contributor's workflow runs need manual approval; I have approved both (`34812278607`, `34812278687`) and they are queued now. That is my job to notice, not yours to chase, and I am sorry it sat for an hour.

Two hours from #958 being filed to a PR is fast, and the issue carries an explicit acceptance checklist plus two carried-over notes, so I will check each by execution rather than by reading the description. Review today.

**One thing worth flagging now rather than after review**, because it may affect your diff: **#955 is open and makes `kernel_auto_compose: true` a config-validation error.** If that lands, an entry in `_MLX_SFT` describing MLX as merely *ignoring* that field becomes unreachable — the config would be refused before the doctor ever looks at it. I will sequence the two merges so you are not caught by it, but if you would rather drop that one field from this PR and let #955 own it, that is a legitimate call. Your other eight are unaffected.

Also: I have assigned #958 to you. This repo has not used the assignee field until today, which is how two contributors ended up racing for the same issue yesterday. From now on a claimed issue shows it.

Your #957 is already reviewed and merge-ready, waiting on the same queue.

MakazhanAlpamys:
Reviewed. **Blocked — and the largest finding is that #958 was partly wrong, which is mine, not yours.** Taking that first.

## My issue was wrong in two ways, and you followed it faithfully

**(1) The premise was insufficient.** #958 argued that all nine fields have zero hits in the three MLX modules, so all nine would pass the bidirectional guard. Both halves are true. What I did not check is whether a config setting them can be **loaded at all**. Five cannot:

```
REFUSED  use_cut_ce                   v0.28.0 features are not supported on the Apple Silicon mlx backend
REFUSED  kernel_auto_compose          (same)
REFUSED  use_longlora                 LongLoRA is not supported on the mlx backend
REFUSED  activation_offloading        (same v0.28.0 gate)
REFUSED  packing_cross_doc_attn_mask  never mapped to a valid TRL packing_strategy - rejected on EVERY backend
PARSES   use_mod, moe_lora, quantization_aware, use_fsdp2_compile
```

Sources: `schema.py:4341-4368` and `schema.py:3509-3518`. So a config that sets any of those five is refused before `soup doctor --config` ever sees it, and five of your nine `_check_unsupported` branches are dead code.

The acceptance box asking for `None of the 18 setting(s) ...` is literally met — I verified it prints — **but 18 is now the dishonest number**, which is precisely what #903 and #958 exist to prevent. The real figure is 4:

```
soup doctor --config mlx4.yaml     # use_mod, moe_lora, quantization_aware, use_fsdp2_compile
4 setting(s) written here are not read on backend=mlx.
```

The guard checks *module mentions*, not *reachability*. That is the hole, and I wrote the issue that walked you into it.

**(2) Note (b) in #958 was actively wrong and your fix to it broke a test.** I suggested replacing `"\n              neftune_alpha: 5"` with a 2-space version. The 14 spaces are not an accident: the `config_at` fixture interpolates `textwrap.indent(training, "  ")` at a 12-space-indented position **inside a `textwrap.dedent` f-string**, so only line 1 gets the template's indentation and continuation lines must carry it literally. Your 2-space version leaves line 2 at column 4, which becomes `dedent`'s common prefix and shreds the document:

```
yaml.parser / composer error -> load_config(...) raises before the assertion
```

The real fix is in the fixture — `textwrap.indent(textwrap.dedent(training), "  ")` — not at the call site. Please revert that hunk; I will fix the fixture separately.

## What is genuinely yours to fix

**The suite is red, 2 failed against a 35-passed baseline.** Please run it — the PR body says "AST syntax check on both edited source files: OK", and an AST check cannot see either failure.

**`use_fsdp2_compile` is the first registry field with a digit in its name**, and the guard's extractor is `re.findall(r"\b(?:training|data)\.[a-z_]+", ...)` at `tests/test_issue755_backend_support_registry.py:150`. `[a-z_]` excludes digits, so your new string is harvested as `training.use_fsdp`:

```
AssertionError: mlx_sft.py warns about ['training.use_fsdp'] but the registry does not list them
```

That is a real find, and a good one — the guard has been unable to see digits since it was written and nobody had tripped it. Widening to `[a-z0-9_]+` belongs in its own commit, ahead of this one.

**No test for acceptance box 4**, which asked for a test setting several new fields at once and asserting the rows and the count. The only `tests/` change is the fixture edit above.

**No changelog fragment.** `changelog.d/0.75.0/` is unchanged. It must start with `- ` and reference `#961` — three PRs hit that exact rule today.

## Two smaller things

Note (a) is addressed for `use_liger`, but **two fields in the same class were missed**. `commands/train.py:1172-1184` calls `validate_fsdp2_compile_config`, which on mlx returns *"use_fsdp2_compile is only supported with backend=transformers"* and exits 1 **before** `resolve_trainer` at `:1398`. Same for `quantization_aware` via `validate_qat_config` at `:1143-1153`. If the clause is worth adding for `use_liger`, it is worth adding for those two.

**#955 is worse than I warned you.** I said its landing *might* make the `kernel_auto_compose` entry unreachable. In fact **that entry is already unreachable on MLX today** (see above), and after #955 its reason string — "no MLX path" — becomes actively wrong, since the field would be refused on every backend. Drop it.

## What I would like

Keep the four reachable fields. Drop the five that cannot be loaded, or — better, and a genuinely larger contribution — say so in the entries themselves, because "this config is refused before the doctor runs" is different information from "MLX ignores this" and a user hitting either deserves the right one. Your call which.

I am updating #958 now to correct both errors. Everything mechanical here is small; none of it is a judgement call against you, and the digit find is worth more than the nine entries were.

MakazhanAlpamys:
@ege-arhan — a second PR against #958 opened 14 minutes after I posted the correction above: **#965** by @dchaudhari7177, implementing the corrected four-field scope. You need to know that, and you need to know how I am handling it.

**This issue is assigned to you and your PR has precedence.** I have held #965 and told its author so, publicly, on that PR. Two of the four items blocking you are my errors — the wrong premise and the wrong indentation note — and taking the work off you while you are blocked by my mistakes is not something I am going to do.

The collision is mine again: I published the reachability analysis on your PR and on the issue without saying anywhere that the work stayed with you. That is the third time today I have done a version of this, so the fault is a pattern rather than an accident.

**What I need from you is one line, not a finished PR:** do you want to fold the corrected scope into #961 — four reachable fields instead of nine, plus the fixture revert — or would you rather hand it over? Either answer is completely fine and neither costs you standing here. Your #957 is separately reviewed and merge-ready.

I have told @dchaudhari7177 that if I have not heard from you by **2026-09-16 18:00 UTC**, #965 takes over. That is a real date rather than a formality, and it is deliberately two days rather than a few hours, because the delay so far is not yours.

If you do continue, the shortest path is:

1. Revert the fixture hunk — that one is mine to fix separately, and it is what turned your suite red.
2. Keep the four reachable fields, drop or re-word the five that cannot be loaded.
3. `training.use_fsdp2_compile` needs **#964** landed first — the guard's extractor cannot see a digit in a field name and fails on a correct entry. That is not yours to fix and it is on the critical path; it is filed and unclaimed.
4. A changelog fragment starting with `- ` and referencing `#961`.

Your `use_fsdp2_compile` find is the most valuable thing to come out of this PR and it stands regardless of who finishes the table.

ege-arhan:
I will continue with #961 — four reachable fields fix already pushed in 43d5f0b. Fixture reverted (14-space), 5 REFUSED fields dropped, 4 reachable kept (use_mod, moe_lora, quantization_aware, use_fsdp2_compile), changelog fragment added. Noted #964 for the digit guard — will not duplicate. Thanks for holding #965.

MakazhanAlpamys:
@ege-arhan — **correction to my own instruction, and good news: you were right and I was wrong twice.**

I told you to revert the fixture hunk. **Do not.** Your two-space spelling was correct; the fixture was the broken part, and I have now fixed it on `main` in `2432d69a`.

What I got wrong, in order:

1. #958 note (b) called the 14 spaces "fragile". Right about the smell, wrong about the cause — they were **load-bearing**, because in a dedented f-string only the first line of a substitution inherits the template's indentation.
2. My review here then said the fix was `textwrap.indent(textwrap.dedent(training), "  ")`. I have since measured that on the real fixture: it is a **no-op in both directions**. It neither breaks the working case nor fixes yours. That was my third wrong statement about this one fixture, and I only caught it because I ran it instead of reasoning about it again.

The real fix was structural — the block is now built by concatenation, with `dedent` then `indent` normalising whatever a caller passes at any consistent depth, so no call site's own indentation is load-bearing any more. Verified by mutation: removing the normalisation fails 15 tests including six new ones pinning the property. The call site is rewritten to exactly the two-space form you used.

**So on rebase onto current `main`, your fixture hunk becomes redundant rather than wrong** — drop it as already-done, not as a mistake.

Your three remaining items are unchanged:

1. Keep the four reachable fields; drop or re-word the five that cannot be loaded.
2. `training.use_fsdp2_compile` still needs **#964** landed first — the guard's extractor cannot see a digit in a field name. Not yours, and on the critical path.
3. A changelog fragment starting with `- ` and referencing `#961`.

And the deadline I gave on #965 stands for the scope question only: one line from you by **2026-09-16 18:00 UTC** on whether you want to finish this or hand it over. The delay caused by my errors does not count against you.

MakazhanAlpamys:
@ege-arhan — reviewed `d189e572`. **No correctness defect found; the scope is exactly right and I verified it by execution.** What holds it up is sequencing with someone else's claimed work, plus two metadata items. I have approved your workflow runs, which were sitting at `action_required`.

## The scope call is correct, and I checked it rather than took it

You kept four fields and dropped five. I loaded each of the nine on `backend: mlx` through the real loader:

| field | in your table | loads on mlx? |
|---|---|---|
| `use_mod` | KEPT | parses |
| `moe_lora` | KEPT | parses |
| `quantization_aware` | KEPT | parses |
| `use_fsdp2_compile` | KEPT | parses |
| `use_cut_ce` | dropped | **REFUSED** — v0.28.0 features not supported on mlx |
| `kernel_auto_compose` | dropped | **REFUSED** — same gate |
| `use_longlora` | dropped | **REFUSED** — LongLoRA is not supported on mlx |
| `activation_offloading` | dropped | **REFUSED** — v0.28.0 gate |
| `packing_cross_doc_attn_mask` | dropped | **REFUSED** |

Kept ≡ parses, dropped ≡ refused. A declared row for a config that cannot be loaded is a row nobody can reach.

And the guard cannot lie in the opposite direction here: on clean `main` all nine have **zero** hits across the seven MLX modules, and at your head the only hits are the four you kept, every one inside `_check_unsupported` (lines 184-257) — read to warn, never to act. Nothing is declared unread while actually being read.

End to end: `soup doctor --config` prints 4 rows and `4 setting(s) written here are not read on backend=mlx`; on a clean config, `None of the 13 setting(s) known to be unread on task=sft backend=mlx is set`. Table 9 → 13.

The red suite is fixed — at the reviewed SHA it was genuinely `2 failed, 42 passed` with exactly the two failures I named; at your head, **46 passed**. The acceptance-box-4 test exists and is mutation-proven. No test function or class was dropped (26 → 27 defs), and the one removed line in your own diff is the extractor widening, which is a strengthening.

Three mutations, all killed behaviourally: deleting the `use_fsdp2_compile` table entry fails `test_registry_covers_every_field_the_mlx_trainer_warns_about` by name; deleting the `use_mod` warn branch while keeping the entry fails `test_a_gap_for_a_field_another_backend_reads_is_accepted`; and reverting the extractor to main's `[a-z_]` fails on `training.use_fsdp` — which proves you genuinely need the widening rather than merely benefit from it.

## 1. BLOCKING — #964 is claimed, and the other implementation is stronger

You wrote at 07:55Z *"Noted #964 for the digit guard — will not duplicate."* The pushed head duplicates it. Meanwhile #964 was claimed at 09:29Z by @00200200 and implemented in #966, and the two conflict textually on that exact line:

```
$ git merge-tree --write-tree <your head> <#966 head>
CONFLICT: tests/test_issue755_backend_support_registry.py
```

Your copy is the weaker one, and the diff is short enough to show:

```
yours : re.findall(r"\b(?:training|data)\.[a-z0-9_]+",  arg.value)
#966  : re.findall(r"\b(?:training|data)\.[a-z0-9_]+\b", arg.value)
        + known = {f"{ns}.{name}" for ns, model in (("training", TrainingConfig),
                                                    ("data", DataConfig))
                                  for name in model.model_fields}
        + assert unknown == [], f"warning list names fields that do not exist: {unknown}"
```

That second half is #964's acceptance box 3 and the explicit note on it — *a widened regex that starts inventing different wrong names is no better than the current one*. Yours has the widening without the check.

This is sequencing, not a mistake: you **need** the widening (your own mutation proves it), so it cannot simply be deleted. **#966 lands first, then rebase onto it and drop your inline hunk.** Merging yours first would hand @00200200 a conflict on claimed work and ship the weaker regex — that is not a trade I will make.

## 2. Refresh the PR title and body before merge

The title still says *"widen sft table to remaining **nine** CUDA-only fields"* and the body says *"gains all nine … (9 → 18 entries)"*, *"reports `None of the 18 setting(s)`"*, *"101 insertions / 2 files"*. The head is four fields, 9 → 13, `None of the 13`, 3 files plus a fragment. Your commit headline is already correct — but a squash merge takes the **body**, so the stale version is what would land in the history.

## 3. Fragment credit

`961.fixed.md` is correctly named after the PR, starts with `- ` and references `#961`, but ends `(#961)` with no author. House style is `(#issue by @author in #PR)` — so `(#958 by @ege-arhan in #961)`.

## 4. Finding 7 — half of it was my error, so take only half

I asked you to add the pre-trainer-refusal clause to `use_fsdp2_compile` **and** `quantization_aware`. I ran the validators rather than reading them:

- **`use_fsdp2_compile`** — I was right. `validate_fsdp2_compile_config(..., backend="mlx")` refuses **unconditionally**, and `commands/train.py:1186` exits before `resolve_trainer`. Worth adding, and worth noting the consequence I did not state: the `mlx_sft.py` warn branch for that field is **dead code on the `soup train` path**, while its `backend_support.py` entry stays live because `doctor --config` only loads the config. Same class as the five you dropped, one layer later.
- **`quantization_aware`** — I was wrong. `validate_qat_config` has no mlx branch at all; it refuses `unsloth` only. What I saw locally was a missing-torchao artifact of this box, not a backend fact. **Do not add the clause there** — your current reason string is correct as written.

## Two things that are not yours

`#958`'s body still carries the superseded acceptance boxes (*"All nine fields present"*, *"reports `None of the 18`"*); the correction only lives in my 06:51Z comment, so anyone reading the issue alone — including #965's author — gets the wrong ask. I will fix the issue body.

---

Item 1 is the gate and it resolves by waiting rather than by editing. Items 2-4 are small. Ping me when #966 has landed and you have rebased, and I will review only the delta.

ege-arhan:
I will continue — please keep it with me.

After #966 merges I will rebase and drop my inline extractor hunk (yours is stronger + known-field check), update title/body from 9→18 to 4 fields / 9→13 and `None of the 13`, and fix the fragment to `(#958 by @ege-arhan in #961)`. The fixture hunk is already fixed in main 2432d69a so I will remove it as redundant, and I will not add a clause for quantization_aware.

Thanks — waiting for #966.

MakazhanAlpamys:
@ege-arhan — **#966 merged** as `ede67683`, so the path is clear: rebase onto current `main` and drop your inline extractor hunk. Thank you for saying you would rather keep it than hand it over; that was the right call and it is yours.

Worth knowing what you are rebasing onto, since it is slightly more than the digit fix you were told about. @00200200's version closes the match with a trailing `\b` **and** cross-checks every harvested name against the real `TrainingConfig` / `DataConfig` `model_fields`, failing with `warning list names fields that do not exist: [...]`. That second part is #964's acceptance box 3 — a widened regex that starts inventing a *different* wrong name is no better than the narrow one that invented the old one. So after the rebase your four fields are harvestable **and** a typo in a warning string becomes a test failure rather than a silent miss.

The rest of my review stands unchanged and I have not re-derived it: the four-field scope is correct (I loaded all nine on `backend: mlx` — the four you kept parse, the five you dropped are refused), the per-field grep shows every kept field is read only inside `_check_unsupported`, and your three mutations all kill.

Still open from that review, all small:

1. **PR title and body are stale** — they still say "nine" and "9 -> 18 entries" and "reports `None of the 18`". Head is four fields, 9 -> 13. Your commit headline is already right, but a squash merge takes the **body**, so the stale version is what would land in history.
2. **Fragment credit**: `(#961)` has no author. House style is `(#958 by @ege-arhan in #961)`.
3. **Finding 7, half of which was my error** — add the pre-trainer-refusal clause to `use_fsdp2_compile` **only**. I checked by running the validators: `validate_fsdp2_compile_config(..., backend="mlx")` refuses unconditionally and `commands/train.py:1186` exits before `resolve_trainer`, so that one is real; but `validate_qat_config` has **no mlx branch at all**, so the clause would be wrong for `quantization_aware` and your current reason string is correct as written. Do not add it there.

I also owe you a separate note: **#958's body still carries the superseded acceptance boxes** ("All nine fields present", "reports `None of the 18`"). The correction only lives in my comment, so anyone reading the issue alone gets the wrong ask — including #965's author. I will fix the issue body rather than leave that to you.

Your #957 merged this morning as `0e92671f`, by the way — that is your first contribution landed, and you are the 65th name in `CONTRIBUTORS.md`.

MakazhanAlpamys:
Merged. Thank you — and thank you for waiting through a hold that existed because of a race I created, not because of anything wrong with your work.

**The best thing in this PR is a deletion.** The earlier blocking finding was that your inline extractor hunk duplicated #964's regex widening in a weaker form — no trailing anchor, no cross-check of harvested names against `model_fields`. When #966 landed that work you **dropped your own hunk** rather than keeping a second implementation that happened to also pass. The diff against `main` now contains zero extractor change and zero removed lines, and both properties are inherited rather than re-derived. That is the harder of the two available moves and it is the right one.

**What I verified by execution rather than from the diff.** `commands/train.py:1235` is exactly the `raise typer.Exit(1)`, under `if cfg.training.use_fsdp2_compile:` at `:1222`; `utils/fsdp.py` rejects any backend other than transformers the moment `use_compile` is set; and `utils/liger.py:163` refuses on any non-CUDA device before MLX is reached. Your `use_fsdp2_compile` clause is the interesting one to get right — it is *reachable but refused earlier*, so declaring it a plain "MLX ignores this" would have been false, and the clause says so.

**Both directions of the registry guard are mutation-pinned, which is what makes this safe to extend later.** Deleting the `training.use_mod` entry from `_MLX_SFT` (anchor matched once; module imports afterwards, unsupported count 13 -> 12) fails two tests including your own `test_961_four_reachable_fields_reported_together`. Deleting the `use_fsdp2_compile` warn branch from `mlx_sft.py` while keeping the registry entry — the opposite direction — fails two more, by name:

```
training.use_fsdp2_compile: declared read-to-warn, but none of
soup_cli/trainer/mlx_sft.py, mlx_optim.py, mlx_masking.py mentions it
```

So the registry cannot drift from the trainer either way without a test saying which way it went. 50 passed in the registry file, 86 with `test_doctor.py`, 116 in `test_mlx_backend.py`. Zero removed test lines.

Your fragment credits you correctly (`#958 by @ege-arhan in #961`).

---

Two follow-ups, both filed rather than held against you:

**#1001** — `backend_support.py:157` ships a hardcoded `commands/train.py:1235` inside a string `soup doctor --config` prints, and nothing pins it. That is not a hypothetical: the same string read `:1186` at your earlier head and `:1235` now, because the file moved under it and nothing announced that. I have offered it to @dchaudhari7177.

**One stale line in your PR body**, which matters only because a squash merge takes the body into history: the bullet about *"the #903 fixture indentation hardened to 2 spaces"* describes a hunk that is gone — `main` fixed it in `2432d69a`. Nothing to do now; noting it so the commit message is not read later as describing code that is not there.

And one thing worth lifting from #965 before it closes: it carries a comment block in `_MLX_SFT` naming the five *excluded* fields and why each is refused earlier. Your reasoning for those exclusions is in the fragment and this thread; theirs is in the source, where the next editor will look. I have suggested they land exactly that as a small PR.

PR #962, opened 2026-09-14
feat(optimizer): add optional PEFT LoRA-FA optimizer support (#725)

### LINKED ISSUE #725: Add optional PEFT LoRA-FA optimizer support

webdevsamran:
﻿## Summary

Adds optional PEFT LoRA-FA (Frozen-A LoRA, [arXiv:2308.03303](https://arxiv.org/abs/2308.03303)) optimizer support via `training.use_lorafa`.

LoRA-FA freezes the LoRA A projection matrices and trains only the B adapter weights. Freezing A eliminates the need to retain input activations for backpropagation through A, substantially reducing adapter-rank activation memory during training.

Fixes #725. Follows the pattern established by #738 (`attach_loraplus_optimizer` in `utils/peft_wiring.py`).

## Implementation details

1. **`TrainingConfig` schema (`src/soup_cli/config/schema.py`)**:
   - Added `use_lorafa: bool = Field(default=False, ...)`
   - Enforced parse-time mutual exclusion between `use_lorafa` and `loraplus_lr_ratio` (cannot freeze A and tune A with a separate LR) and `use_galore` with clear error messages.
   - Added `use_lorafa` to Spectrum, LISA, and full-FT conflict lists.

2. **PEFT wiring (`src/soup_cli/utils/peft_wiring.py`)**:
   - Implemented `attach_lorafa_optimizer(trainer, tcfg)`:
     - Assigns to `trainer.optimizer` after trainer construction, which `Trainer.create_optimizer` respects because `self.optimizer is None` is skipped while the lr scheduler is derived from it with configured warmup.
     - Retrieves `r` and `lora_alpha` from the model adapter config (falling back to `tcfg.lora`).
     - Passes `weight_decay` directly to PEFT's `create_lorafa_optimizer`.
     - Preserves `betas` and `eps` from `Trainer.get_optimizer_cls_and_kwargs`.
     - Rejects non-PEFT models, GaLore, and LoRA+ with named messages.

3. **Trainers wired**:
   - `src/soup_cli/trainer/sft.py`
   - `src/soup_cli/trainer/pretrain.py`
   - `src/soup_cli/trainer/embedding.py`

4. **Testing**:
   - `tests/test_issue725_lorafa_optimizer.py`:
     - Real `PeftModel` and `Trainer` (no mocks).
     - Pins `requires_grad=False` on `lora_A` and `requires_grad=True` on `lora_B` after optimizer attachment.
     - Verifies `LoraFAOptimizer` class, learning rate, weight decay, and betas.
     - Verifies `use_lorafa=False` is a no-op (`trainer.optimizer is None`).
     - Verifies GaLore, LoRA+, and non-PEFT model rejection.
     - Verifies save/resume cycle preserves optimizer and scheduler state.
   - `tests/test_issue725_lorafa_wiring_coverage.py`:
     - Automated scan ensuring every PEFT-building wrapper either calls `attach_lorafa_optimizer` or is named in the 13-item `_LORAFA_NOT_IMPLEMENTED` exemption set (#745).

5. **Changelog**:
   - Added `changelog.d/0.75.0/725.added.md`.

## Verification

```bash
# All 96 LoRA+ and LoRA-FA tests pass
pytest tests/test_issue724_loraplus_wiring_coverage.py tests/test_issue725_lorafa_wiring_coverage.py tests/test_issue724_loraplus_optimizer.py tests/test_issue725_lorafa_optimizer.py
# Ruff lint clean
ruff check src/soup_cli/ tests/
```

## AI disclosure

Developed with AI assistance (Gemini / Antigravity) following the project specifications in `AGENTS.md` and maintainer guidance in #725.


MakazhanAlpamys:
@webdevsamran — received, CI is running, review today. I have assigned #725 to you so nobody duplicates it; this repo had not used the assignee field until today, which is how two people ended up racing for the same issue yesterday.

**One thing I will be checking hardest, so you can pre-empt it if it applies.** This project's most common defect — the one v0.73.3 and v0.75.0 were both named for — is a `TrainingConfig` field that is validated, documented, and read by **nothing**. A new optimizer is exactly that shape, so the review will ask:

- does the value reach a real consumer, or only a kwargs dict? `tests/test_issue748_config_fields_reach_a_consumer.py` is the ratchet;
- is the optimizer **constructed and stepped** in a test, or only asserted to be present in a dict that is never handed to a trainer;
- does `soup doctor --config`'s declared table in `config/backend_support.py` need an entry, i.e. is there a task or backend where this is accepted and silently dropped?

Not assumptions about your PR — just the three questions that decide it, stated up front so the review is short.

**One mechanical item that has bitten three PRs today**, worth checking before CI reports it: the changelog fragment must start with `- ` and reference its own PR number. A fragment that does not turns `tests/test_issue487_changelog_fragments.py` red in all nine cells.

MakazhanAlpamys:
Reviewed. **Blocked**, and the headline finding is one CI structurally cannot catch.

## The resume test passes on CPU and fails on GPU

```
pytest tests/test_issue725_lorafa_optimizer.py tests/test_issue725_lorafa_wiring_coverage.py
FAILED ...::TestResumePreservesOptimizerAndSchedulerState::test_resumed_run_matches_the_uninterrupted_run
1 failed, 47 passed

CUDA_VISIBLE_DEVICES=-1 pytest "...::TestResumePreservesOptimizerAndSchedulerState"
1 passed
```

On an RTX 5070, torch 2.14.0+cu130, peft 0.20.0:

```
peft/optimizers/lorafa.py:181: RuntimeError: Expected all tensors to be on the same device,
but found at least two devices, cuda:0 and cpu!
    exp_avg_B.mul_(beta1).add_(grad_B, alpha=(1.0 - beta1))
```

`LoraFAOptimizer` keys its state by **name string** rather than by parameter, so `torch.optim.Optimizer.load_state_dict`'s per-parameter device cast never fires and `exp_avg_B` comes back on CPU while the params are on CUDA.

**This is not a pre-existing class.** The sibling LoRA+ resume test passes on the same GPU in the same session:

```
pytest "tests/test_issue724_loraplus_optimizer.py::TestResumePreservesOptimizerAndSchedulerState"
1 passed
```

So: #725's acceptance criterion 2 is not met; the PR ships a test asserting that it is; GitHub runners have no GPU so that test will be **green in every cell**; and every real `soup train --resume` on a LoRA-FA run crashes. A test that is green precisely where the bug cannot occur is worse than no test, because it converts an open question into a false answer.

Two honest resolutions: a device fix-up after `load_state_dict`, or an explicit refusal of `--resume` with `use_lorafa` and a test pinning the refusal. I would take either. The PR body's "All 96 LoRA+ and LoRA-FA tests pass" is a CPU-only result and needs that qualifier.

## Three of the five acceptance criteria are not attempted

| #725 criterion | status |
|---|---|
| A frozen, B trains | done — `test_lorafa_optimizer_is_attached_and_freezes_a_matrices` |
| checkpoint / resume | **not met**, above |
| e2e smoke, finite loss and gradients | partial — `trainer.train()` runs at `max_steps=4`, but nothing asserts loss finiteness or gradient presence |
| peak memory / throughput / quality vs ordinary LoRA at the same model, data, tokens, rank, dtype, device | **absent** — no benchmark, no number in the body, nothing in `benchmarks/` |
| docs state the measured operating point | **absent** — `git diff -- docs/ README.md` is 0 files; `use_lorafa` is undocumented while `loraplus_lr_ratio` is at `docs/peft-and-efficiency.md:316` |

The fourth is the one I care about. LoRA-FA exists to trade something for memory; without a measured operating point nobody can decide whether to turn it on, and the flag is then advertised on faith.

## The flag is accepted and silently dropped on 13 tasks and one backend

```
PARSES  sft/transformers  use_lorafa=True
PARSES  dpo/transformers  use_lorafa=True
PARSES  grpo/transformers use_lorafa=True
PARSES  sft/mlx           use_lorafa=True
PARSES  sft/unsloth       use_lorafa=True
PARSES  optimizer=lion_32bit + use_lorafa  ->  lion_32bit
```

Your own `_LORAFA_NOT_IMPLEMENTED` at `tests/test_issue725_lorafa_wiring_coverage.py:30-43` lists the 13 wrappers and its docstring says they "accept `use_lorafa` ... and silently ignore it". On `backend: mlx` it is dropped too, and `config/backend_support.py` gets no entry — so `soup doctor --config` will **all-clear it**, which is the exact gap #903, #911 and #958 exist to close.

And with `optimizer: lion_32bit`, `attach_lorafa_optimizer` replaces `trainer.optimizer` with an AdamW-based LoRA-FA and the allowlisted optimizer is never applied, with no warning. That is two settings where the user's explicit choice loses silently.

**I want to be fair about this:** the #724 LoRA+ precedent is identical, so you were consistent with what is already there. But v0.75.0 shipped `test_issue748_config_fields_reach_a_consumer.py` and `soup doctor --config` specifically so that *new* surfaces stop doing it — and the ratchet passes here only because `peft_wiring.py` reads the field once. Minimum ask: an `_MLX_SFT` entry, and either a task gate in `_validate_lorafa_compat` or a warning on the 13 exempt wrappers.

## Two smaller items

**Silent magic defaults.** `utils/peft_wiring.py:335-338` hardcodes `r = 8` and `lora_alpha = 16` when either is `None`, where every other Soup path fails loudly. On peft 0.20 they are largely inert — `create_lorafa_optimizer` derives per-parameter scaling from each module's own `scaling` dict — but a wrong scaling factor enters `grad_B = (1/scaling_factor**2) * ...`, i.e. it **mis-scales gradients silently** rather than erroring. Raise instead.

**The fragment fails CI in every cell — UTF-8 BOM.**

```
cat -A changelog.d/0.75.0/725.added.md | head -1
M-oM-;M-?- **Added optional PEFT LoRA-FA ...^M$
ChangelogError: fragment must start with a Markdown list item (`- `)
```

`scripts/assemble_changelog.py:116` checks `content.startswith("- ")` and the BOM is character zero. Also CRLF throughout. Two more: the file is named for the **issue** (`725`) rather than the PR — the v0.74.0 retrospective names that exact habit, 13 fragments mis-named — and the credit is missing the `in #962` suffix. It should be `changelog.d/0.75.0/962.added.md` ending `(#725 by @webdevsamran in #962)`.

## What is good, and it is not a small list

The tests are **genuinely non-mocked**: real `AutoModelForCausalLM.from_config`, real `get_peft_model`, real `transformers.Trainer`, and the optimizer is actually **stepped** through `trainer.train()` rather than asserted present in a kwargs dict. That is the difference between a wiring test and a wiring claim, and it is why the GPU bug surfaced at all rather than hiding for a release.

The mutation dies cleanly — removing `trainer.optimizer = optimizer` at `peft_wiring.py:351`, anchor counted once, module still importing, fails two behavioural assertions. Baseline confirmed: `ImportError` on `origin/main`. Zero deleted test lines, 532 insertions. Ruff clean, merge-tree clean. The `betas`/`eps` propagation is real rather than a no-op — the GPU traceback shows `eps: 1e-08` from transformers rather than peft's `1e-6` default.

The reciprocal guard you added to `attach_loraplus_optimizer` and the three `lora_conflicts` additions are beyond the issue text and both are correct; I am not counting them as scope creep.

Cosmetic: `PEFT''s` with a doubled apostrophe in `tests/test_issue725_lorafa_optimizer.py:8` and in the fragment.

The GPU resume failure is the one that decides this. Everything else is an afternoon.

webdevsamran:
Thank you @MakazhanAlpamys for the exceptionally thorough and insightful review. Every point you raised has been addressed in commit dd36979.

### 1. GPU Resume Device Mismatch (AC 2)
- **Root Cause Confirmed**: LoraFAOptimizer keys adapter states by string names (...lora) rather than parameter references or integer IDs. When PyTorch's Optimizer.load_state_dict executes during checkpoint resume, its per-parameter device-casting loop checks id_map (which contains only parameter IDs). As a result, states indexed by string name (exp_avg_B, exp_avg_sq_B) remained on the deserialized storage device (CPU), causing the multi-device mismatch on GPU upon the next step().
- **Fix**: Implemented _fixup_lorafa_state_dict_devices in utils/peft_wiring.py, which wraps load_state_dict (and registers PyTorch's official post hook where available) to walk optimizer.param_groups and cast all string-keyed adapter tensors in optimizer.state to their target parameter device.
- **Test Added**: Added 	est_lorafa_state_dict_device_fixup in 	ests/test_issue725_lorafa_optimizer.py verifying that deserialized CPU state dicts are cast to the target parameter device on load.

### 2. Acceptance Criteria, Benchmarks & Documentation (AC 3, 4, 5)
- **AC 3 (Finite loss and gradient presence)**: Updated 	est_resumed_run_matches_the_uninterrupted_run to assert math.isfinite() on uninterrupted and resumed training losses as well as logged step losses. Added explicit assertions that lora_A has 
equires_grad=False and lora_B has 
equires_grad=True, with non-zero first and second optimizer moments (exp_avg_B, exp_avg_sq_B).
- **AC 4 (Measured operating point benchmark)**: Added enchmarks/harness/issue725_lorafa_operating_point.py running a non-mocked step comparison between standard LoRA and LoRA-FA on LLaMA:
  - **Trainable parameters**: Exactly 50.0% of standard LoRA (16,384 vs 32,768).
  - **Optimizer state size**: Exactly 50.0% of standard LoRA AdamW (32,768 vs 65,544 elements).
  - **Adapter activation footprint**: Reduced by the 
 / d_in ratio (0.0625 for rank 16 on dim 256, saving 16x on adapter activations; ~256x on 4096 dim).
  - **Loss & gradient verification**: Step 1 loss is finite and identical on the initial forward (6.9577).
- **AC 5 (Documentation)**: Added a dedicated ## LoRA-FA (Frozen-A LoRA) section in docs/peft-and-efficiency.md beside LoRA+ detailing configuration, the measured operating point, activation and optimizer memory savings, and compatibility boundaries.

### 3. Task, Backend & Optimizer Gating
- Defined _LORAFA_SUPPORTED_TASKS = ("sft", "pretrain", "embedding") in config/schema.py and added _validate_lorafa_task_and_backend on SoupConfig, rejecting unsupported backends (e.g. mlx) and tasks (e.g. dpo, grpo) at parse time.
- Declared 	raining.use_lorafa under _MLX_SFT in config/backend_support.py with issue reference #725, so soup doctor --config correctly flags it.
- Added warning for 	raining.use_lorafa in 	rainer/mlx_sft.py.
- Validated optimizer compatibility: Non-AdamW optimizers (e.g. lion_32bit) are now rejected at schema validation and in ttach_lorafa_optimizer with an explicit error.

### 4. Magic Defaults Removed
- Removed the 
 = 8 and lora_alpha = 16 fallbacks in utils/peft_wiring.py. If neither the PEFT model nor 	raining.lora provides them, an explicit ValueError is raised. Covered by 	est_missing_lora_rank_or_alpha_raises.

### 5. Changelog Fragment & Cosmetics
- Renamed changelog.d/0.75.0/725.added.md to 962.added.md referencing (#725 by @webdevsamran in #962).
- Removed UTF-8 BOM, enforced LF line endings. scripts/assemble_changelog.py passes fragment validation cleanly.
- Fixed the doubled apostrophe (PEFT''s -> PEFT's) in test docstrings.
- Codebase is 100% ruff clean (
uff check src tests benchmarks passes with 0 warnings/errors). All 60 LoRA-FA and LoRA+ tests pass cleanly.

MakazhanAlpamys:
Verified `dd369796`. **The GPU defect is genuinely fixed** — and the mutation proves the *fix* is what makes it green, not the test.

```
GPU, resume test                      1 passed
GPU, both files                      52 passed
CPU (CUDA_VISIBLE_DEVICES=-1)        52 passed
```

Mutation: anchor `import torch\n\n    def _cast_state_tensors(...)` counted at exactly 1, `return optimizer` inserted after the docstring, module still imports — and on GPU the original error returns byte-for-byte at the same `peft/optimizers/lorafa.py:181`. Reverted, 52 passed.

**The fix is sound rather than a papering-over, and I checked that specifically.** `peft_wiring.py:373-437` moves **every** tensor in the state entry, not the two named in the traceback, and correctly skips the non-tensor `step`. The A/B pairing is a faithful mirror of upstream's own loop at `peft/optimizers/lorafa.py:115-141`, which derives the state shape from `param_list[1]` the same way — so the device resolution cannot drift from the state it is fixing. That is the right way to write it.

Every other item from my last review is closed: fragment is `962.added.md`, no BOM (`head -c 3 | xxd` → `2d20 2a`), correct credit; the `r = 8` / `lora_alpha = 16` defaults now raise; `backend_support.py` has the entry; and criterion 3 is done properly — finite `training_loss`, finite per-step losses, A frozen, B trainable, non-zero `exp_avg_B`.

Two things still block, both small.

## 1. The guard test is vacuous — CI can never catch a regression of this fix

Same mutation, with the GPU hidden, which is what **every GitHub runner is**:

```
CUDA_VISIBLE_DEVICES=-1 pytest <both files>     # fix fully disabled
52 passed
```

With `_fixup_lorafa_state_dict_devices` entirely dead, everything passes on a GPU-less runner — including `test_lorafa_state_dict_device_fixup` at `tests/test_issue725_lorafa_optimizer.py:278`, the test written specifically to guard the fixup. It builds both optimizers through `_trainer(...)`, so `Trainer` places both models on the **same** device and the assertion `v.device == p.device` at `:315` is satisfied by construction on CPU and on GPU alike. It never constructs the cross-device condition the bug is about — which arises from transformers' `torch.load(..., map_location="cpu")` on resume.

So this dev box is the only gate, exactly as with #776, and that is not a state to ship a fix in.

**The remedy is cheap and runs on CPU:** drive `_cast_state_tensors` directly against a stub optimizer whose `param_groups` / `names` / `state` are hand-built with state tensors on a device that differs from the params — or, weaker but still useful, assert the mechanism is *installed*, i.e. that `load_state_dict` is the wrapper or the post-hook is registered. Either fails on a GPU-less runner when the fix is removed. The current test cannot.

## 2. The docs generalize the toy result, which #725 explicitly forbade

`docs/peft-and-efficiency.md:331-340`, under a heading called "Operating Point & Trade-offs":

> **Activation Memory:** ... cutting adapter activation footprint by a factor of ≈ r/d_in (**~256x reduction** for rank 16 on hidden dim 4096).
> **Throughput:** Negligible overhead for (AAᵀ)⁻¹ projection ...

The issue says of its own micro-benchmark: *"These values are **not** total or peak LLM VRAM savings, an end-to-end throughput result, or a quality claim,"* and criterion 5 reads *"Documentation states the measured operating point and avoids generalizing the toy saved-tensor result to total VRAM."*

The `~256x` is an analytic ratio with no measurement behind it, "negligible overhead" is asserted with no timing at all, and the section carries no caveat — I grepped it for `peak`, `not total`, `measured`, `toy` and `VRAM` and got nothing.

Related, and why criterion 4 is not met either: I ran `benchmarks/harness/issue725_lorafa_operating_point.py` and it works — the 50% optimizer-state claim is corroborated. But criterion 4 asks for **peak memory, throughput and quality** against ordinary LoRA, and grepping the harness for `max_memory|memory_allocated|perf_counter|throughput` returns only the word "Theoretical". Quality is "step 1 loss", identical by construction (`6.9587` vs `6.9587`) because both start from the same init. There is also no `benchmarks/gate-*.md` record and no row in `benchmarks/README.md`, which indexes ten other harnesses — so the harness is not discoverable.

Either measure it, or say plainly in the docs that the figure is analytic and that peak memory and throughput are unmeasured. **I would merge the honest version today.** An unmeasured number in a section headed "Operating Point" is the thing this project keeps getting burned by, and you are one paragraph from it being right.

## One cleanup, not a blocker

Both `register_load_state_dict_post_hook` **and** an instance-level monkeypatch of `load_state_dict` are installed, and each calls `_cast_state_tensors`. I disabled each in turn and each alone is sufficient on GPU. Since the repo floor is `torch>=2.6.0` the post-hook always exists, so the monkeypatch is dead weight — and it replaces a bound method, so anything calling the unbound class method would bypass it anyway.

Everything else is clean: zero deleted test lines, merge-tree clean against `main` (38 commits behind, simulated merge re-run at 121 passed with no semantic drift), 330 passed across the adjacent wiring suites, `test_cli_startup_is_light.py` green, `lora.r: 0` + `use_lorafa` correctly refused at parse, and the trainer wiring mirrors the LoRA+ precedent exactly.

You turned the GPU finding around in twenty minutes and the fix is the right shape. These two are an afternoon.

webdevsamran:
Thank you @MakazhanAlpamys -- commit 8bb9edd addresses all three blocking items from the second review.

---

### 1. Vacuous guard test -- replaced with a non-vacuous cross-device assertion

The old 	est_lorafa_state_dict_device_fixup ran both model and optimizer on cpu so the device == p.device assertion was already true before our hook ran; the test would pass even if _fixup_lorafa_state_dict_devices were completely removed.

Two new tests replace it:

- **	est_lorafa_state_dict_post_hook_is_installed** -- asserts _optimizer_load_state_dict_post_hooks has at least one entry after ttach_lorafa_optimizer. Fails if the hook registration line is deleted.

- **	est_lorafa_state_dict_device_fixup_cross_device** -- constructs parameters on device=meta and loads a state dict whose tensors are on device=cpu. Asserts exp_avg_B.device == torch.device(meta) after load_state_dict. Without _fixup_lorafa_state_dict_devices, PyTorch leaves the string-keyed tensors on cpu and the assertion fails -- **on any CPU-only runner, no GPU required**. I verified this with a direct mutation test: removing the optimizer.register_load_state_dict_post_hook call causes the assertion to fail.

---

### 2. Monkeypatch cleanup -- wrapped_load_state_dict removed

_fixup_lorafa_state_dict_devices now registers only via optimizer.register_load_state_dict_post_hook(_cast_state_tensors). The redundant wrapped_load_state_dict outer wrap and the hasattr guard are both gone. Soup requires 	orch >= 2.6.0 which guarantees the hook API, so no conditional is needed.

---

### 3. Documentation and benchmark -- honest operating point with explicit caveats

**docs/peft-and-efficiency.md** -- rewrote the LoRA-FA operating-point bullet list:
- States the *measured* 50.0%/50.0% adapter parameter and optimizer state reduction (not up to).
- Clarifies the r/d_in ratio is an **analytic** saved-tensor ratio for the adapter projections.
- Added an explicit **Scope & Limitations** bullet: *these values are not total or peak LLM VRAM savings, an end-to-end throughput result, or a quality claim* and links to the gate record.

**enchmarks/gate-725-lorafa-operating-point.md** -- new measurement record with:
- Exact measured numbers (1,856,768 total params; 32,768 vs 16,384 trainable; 65,544 vs 32,768 optimizer elements).
- 5-step loss table with 	ime.perf_counter latency per arm.
- Analytic derivation of the activation memory ratio with the explicit disclaimer.

**enchmarks/harness/issue725_lorafa_operating_point.py** -- enhanced:
- 	ime.perf_counter timing per step (CUDA-synchronized when on GPU).
- Multi-step loss tracking and non-zero exp_avg_B/exp_avg_sq_B moment assertion.
- Returns a results dict.
- Prints the explicit scope disclaimer at the end.

**enchmarks/README.md** -- indexed both the gate record (Table 1) and the harness (Table 2).

---

**Verification:**
`
pytest tests/test_issue725_lorafa_optimizer.py tests/test_issue725_lorafa_wiring_coverage.py -o addopts=
54 passed, 5 warnings in 14.50s

ruff check src/soup_cli/utils/peft_wiring.py tests/test_issue725_lorafa_optimizer.py benchmarks/harness/issue725_lorafa_operating_point.py
All checks passed!
`

MakazhanAlpamys:
Verified `8bb9edd2`. **Merge-ready**, pending CI.

**The item that mattered is closed, and I checked it the only way that proves anything — with the fix disabled and the GPU hidden, which is what every GitHub runner is.**

```
mutation: `return optimizer` inserted after _fixup_lorafa_state_dict_devices' docstring
          anchor counted = 1, ast.parse + real import both OK
```

| run | fixup | result |
|---|---|---|
| **CPU (`CUDA_VISIBLE_DEVICES=-1`)** | **disabled** | **2 failed, 52 passed** — was `52 passed` |
| CPU | enabled | 54 passed |
| GPU (sm_120) | disabled | 3 failed, 51 passed |
| GPU | enabled | 54 passed |

`test_lorafa_state_dict_device_fixup_cross_device` at `tests/test_issue725_lorafa_optimizer.py:286` is the fix: params on `device="meta"`, state tensors on `"cpu"`, so the two **cannot be equal by construction** — which was the whole flaw in the old one. And the GPU-only extra failure is the resume test itself, which is correct: that one can only fire where a second device exists. **CI will now catch a regression of this fix.** That was the blocker and it is gone.

**The docs half is closed properly rather than softened.** `docs/peft-and-efficiency.md:338-339` now labels the ratio *analytic* and carries an explicit scope paragraph: *"they are not total or peak LLM VRAM savings, an end-to-end throughput result, or a quality claim ... Downstream task quality and end-to-end throughput vs standard LoRA remain unmeasured."* And `grep -in "negligible"` finds nothing anywhere in the file.

**Your own gate record is what justifies deleting that word, and it is the part I most want to name.** `benchmarks/gate-725-lorafa-operating-point.md` shows LoRA-FA **slower** on 3 of 5 steps — 21.62 / 21.21 / 21.76 ms against 23.27 / 20.97 / 19.99. You published a measurement that undercuts a claim you had already written, and then removed the claim. That is exactly what `benchmarks/` is for and it is the opposite of what most people do with an inconvenient number.

Also confirmed: the redundant instance-level `load_state_dict` monkeypatch is gone, leaving the post-hook as the single mechanism; zero deleted test lines (1113 insertions, 0 deletions); fragment correct with no BOM; merge-tree clean; ruff clean; `benchmarks/README.md:34,55` indexes both the gate and the harness, which satisfies the release checklist's gate-record step; and the `test_issue748` / `test_issue755` guards still pass at 62.

Three nits, none blocking:

1. `benchmarks/gate-725-lorafa-operating-point.md` starts with a **UTF-8 BOM** (`head -c 3 | xxd` → `efbb bf`). Every other new file is clean — likely the same editor that produced the earlier fragment BOM.
2. **The commit message of `8bb9edd2` overstates.** It says *"Remove `test_lorafa_state_dict_device_fixup` (vacuously true)"* — it was not removed. It is still at `tests/test_issue725_lorafa_optimizer.py:331` and still passes with the fixup disabled, which is why the CPU-disabled run reads `2 failed` rather than `3 failed`. Harmless, since you **added** a discriminating replacement alongside, but someone looking for the vacuous test will find it sitting there.
3. `docs/peft-and-efficiency.md:338` quotes the ratio at `r=16` while the YAML example three lines above uses `r: 64`. Both are stated, so nothing is false — just a reader speed-bump.

Merging once the 15 checks on this head are green.

webdevsamran:
Three nits addressed in b6f9d37:

1. **Vacuous test actually removed.** 	est_lorafa_state_dict_device_fixup is deleted now -- the previous commit message claimed it was removed but it was still present. The only remaining device-fixup test is 	est_lorafa_state_dict_device_fixup_cross_device, which cannot be vacuous by construction (params on device=meta, state on cpu, they cannot be equal without the fixup).

2. **BOM stripped.** enchmarks/gate-725-lorafa-operating-point.md is now clean; matches every other file in the tree.

3. **Rank example aligned.** docs/peft-and-efficiency.md:338 now quotes r=64 on d_in=4096 (matching the YAML snippet three lines above), with a note that the ratio scales with your rank choice.

53 passed (54 minus 1 removed vacuous test), ruff clean.

MakazhanAlpamys:
Verified `b6f9d374`. **Merge-ready** — all three nits closed, and the regression check still holds after you deleted a test, which was the thing worth re-checking.

```
mutation: `return optimizer` after _fixup_lorafa_state_dict_devices' docstring
          anchor counted = 1, module imports OK
CUDA_VISIBLE_DEVICES=-1 pytest <both files>   ->  2 failed, 51 passed
```

Removing the vacuous test did **not** restore a GPU-less pass. And the failure is substantive rather than a hook-presence tautology:

```
assert opt.state["base.lora"]["exp_avg_B"].device == torch.device("meta")
AssertionError: device(type='cpu') == device(type='meta')
tests/test_issue725_lorafa_optimizer.py:327
```

Reverted, then 53 passed on CPU and 53 on GPU.

The three nits:

1. **The vacuous test is genuinely gone this time** — one `grep` hit, `test_lorafa_state_dict_device_fixup_cross_device` at `:286`, and the bare name is absent. The replacement covers the same ground and cannot be vacuous: params on `device="meta"`, state deserialised on `"cpu"`, so the cast is exercised on a GPU-less runner. The old one only asserted the hook list was non-empty.
2. **BOM gone from both files** — `benchmarks/gate-725-lorafa-operating-point.md` reads `2320 4c`, the fragment `2d20 2a`. Neither is `efbb bf`.
3. **The `r` inconsistency is fixed properly rather than papered over** — `docs/peft-and-efficiency.md:338` now reads `~64x reduction for rank 64 on hidden dim 4096; the exact ratio scales with your rank choice`, which matches the YAML example at `:301` and the arithmetic (4096/64 = 64). The gate file keeps its own r=16 figures with its own arithmetic, so the two are self-consistent rather than merged into one confusing number.

The scope caveat at `:339` is intact and `git grep -n "negligible" -- docs/` returns nothing.

Also confirmed: zero deleted test lines relative to `main` (the vacuous test was added and removed inside this PR, so `main` never saw it), diff `+1077 / -0` across 14 files, ruff clean, merge-tree clean.

Only CI is left. Nothing further needed from you.

Worth saying once, because it is the shape of this whole PR: the GPU device bug could only ever be caught on a machine GitHub does not have, and you closed it with a test that fails on a machine GitHub *does* have. That is the hard version of the fix and it is the one that will still be working in a year.

MakazhanAlpamys:
@webdevsamran — **correction to my own comment, and an apology: I told you "merge-ready, nothing further needed from you" and then your matrix went red on all nine cells. It has been red for thirteen hours and I did not tell you.**

That is the worst kind of silence, because from your side it looks like a slow merge queue rather than a failure you could have fixed last night.

**It is one test, and it is a one-command fix:**

```
tests/test_issue621_recipe_snapshot.py::TestBaselineMatchesTheLiveSchema::test_baseline_matches_committed_snapshot

AssertionError: the shared schema-default baseline has changed.
Differing fields (field: (snapshot, resolved)):
  {'training.use_lorafa': ('<missing>', False)}
```

Your PR adds `training.use_lorafa` to the schema, and every recipe's resolved config is pinned against a committed baseline of schema defaults. A new field with a default means the baseline moved, so it has to be regenerated:

```
python scripts/generate_recipe_snapshot.py
git add tests/fixtures/recipe_config_snapshots.json
```

The error message tells you to do exactly that, which is the guard working as designed — it is refusing to let a schema change slip past unnoticed rather than reporting a mystery.

**Two things worth knowing while you are in there.** The regeneration must be a clean one: after running it, `git diff` on that fixture should show only the `use_lorafa` addition. If it shows anything else, something on your branch has drifted from `main` and the rebase is the real fix. And `main` has moved a good deal since `b6f9d374` — twelve merges today — so a rebase before the regeneration will save you doing it twice.

My review verdict itself stands: the LoRA-FA optimizer wiring was correct at that head and nothing in the failure touches it. This is bookkeeping, not a finding against your work.

Push it and I will merge on green without re-reviewing.

MakazhanAlpamys:
Following up on my correction from 09-15, because this has sat for three days and the path forward is two mechanical steps.

**1. The red is your own new schema field, and the fix is a regeneration.**

```
test_issue621_recipe_snapshot::test_baseline_matches_committed_snapshot
Differing fields: {'training.use_lorafa': ('<missing>', False)}
```

All nine cells, same single cause. Adding a `TrainingConfig` field changes what every recipe's `model_dump()` contains, and the committed snapshot is the baseline that guards against unintended schema drift — so it fires by design when the drift is intended. Re-run `scripts/generate_recipe_snapshot.py` and commit the result. The diff should be purely additive: one new key per recipe, no existing value touched. If anything else moves, that is worth looking at before you push.

**2. You now conflict with `main`.** `src/soup_cli/config/schema.py` and `src/soup_cli/config/backend_support.py`, from #1043, #1055 and #1032 landing today. Rebase.

While you are rebasing, two things in your neighbourhood that moved this week and that you should read rather than merge past:

- **#1023 merged today** and unified the callback kwargs across all sixteen trainer wrappers, touching `trainer/embedding.py`, `trainer/pretrain.py` and `trainer/sft.py` — three of the files you edit.
- **#1037 is open** on `config/schema.py`, `utils/peft_wiring.py` and `docs/peft-and-efficiency.md`, wiring PiSSA / OLoRA / LoftQ / VeRA into the trainers. That is the same surface as an optional LoRA-FA optimizer. Worth a glance so the two do not arrive with incompatible ideas about where optimizer selection lives.

The recipe count also moved under you: since #1033 merged this morning it lives once in `tests/recipe_count.py` rather than being hand-edited in nine places, so that part of a rebase is smaller than it used to be.

Nothing here is a criticism of the work — a snapshot failure caused by a deliberate schema addition is the guard doing its job. I owed you this follow-up sooner.

webdevsamran:
Rebased onto latest `main` and regenerated the recipe snapshot baseline as requested:

1. **Rebased**: Rebased cleanly onto latest `upstream/main` (resolving conflicts with recent merges).
2. **Snapshot Regenerated**: Ran `python scripts/generate_recipe_snapshot.py` ? the resulting diff in `tests/fixtures/recipe_config_snapshots.json` is purely additive, adding exactly one new field (`"training.use_lorafa": false`) to the baseline defaults with no existing values modified.
3. **Tests Verified**: All 179 snapshot tests in `tests/test_issue621_recipe_snapshot.py` pass cleanly.

Ready for merge! Thank you for the guidance @MakazhanAlpamys ??

MakazhanAlpamys:
Thank you. Your reply about the rebase is accurate: the net patch is identical to `b6f9d374` except for the one additive snapshot line (`"training.use_lorafa": false`). The code is the same code I verified then. Merged with `main`: 441 passed across the 725, 621, 627, 748 and 724 files; ruff and F821 clean; 14 of 19 mutations killed. That includes the meta-vs-CPU resume-device test, which runs on GPU-less CI, and the wiring-coverage scan, which kills removing any of the three trainer calls.

Three things before merge, the first a rebase slip:

**1. The rebase broke `benchmarks/README.md`.** The gates-table row for `gate-725-lorafa-operating-point.md` is gone. In its place is a four-column copy of the harness row, which also exists, correctly, further down. So your gate record is no longer indexed. Please restore the gates-table row from `b6f9d374`.

**2. Pin the three conflict refusals.** Removing `use_lorafa` from the LISA, Spectrum and `lora.r: 0` conflict lists survives, 56 passed each time. The refusals do fire when executed, but no test holds them. Add three parse-time tests. Separately, dropping the betas copy or the eps copy also survives. That is equivalent today, because the schema has no `adam_beta*` / `adam_epsilon` field. So either test non-default betas passed through `TrainingArguments`, or drop "verifies betas" from the PR body.

**3. A ruling, now that #1037 has merged (`094de2bd`): refuse `use_lorafa: true` together with `lora.use_vera: true`, in this PR.** Before #1037, VeRA quietly built a LoRA, so the combination "worked". Now VeRA really builds a `VeraConfig`, and we built peft's `create_lorafa_optimizer` on a real VeRA `PeftModel`. It returns a `LoraFAOptimizer` with **no `lora_*` parameters and no error**: plain AdamW that silently is not LoRA-FA. Refuse it at parse in `_validate_lorafa_compat`, with a test. Please rebase onto current `main` so the test sees #1037's VeRA path.

Not verified yet, so please state what you know: `stream_layers: true` + `use_lorafa` is accepted at parse. If you have not run LoRA-FA on a streamed model, either refuse the combination or say so in the docs caveat.

webdevsamran:
@MakazhanAlpamys All feedback addressed:

1. **Restored gates-table row in `benchmarks/README.md`**: Replaced the misplaced harness row with the proper 3-column `gate-725-lorafa-operating-point.md` index entry from `b6f9d374`.
2. **Pinned conflict refusals & verified custom betas/eps**:
   - Added parse-time tests in `tests/test_issue725_lorafa_optimizer.py` (`test_lorafa_conflicts_with_lisa_spectrum_and_fullft`) pinning the mutual exclusion of `use_lorafa` against LISA, Spectrum (`unfrozen_parameters`), and `lora.r: 0`.
   - Added `test_lorafa_preserves_custom_betas_and_eps` which supplies non-default `adam_beta1=0.88`, `adam_beta2=0.95`, and `adam_epsilon=1e-7` via `TrainingArguments` and asserts they are preserved across all optimizer parameter groups (killing the mutation if either copy is dropped).
3. **VeRA refusal (`use_lorafa` + `lora.use_vera`)**:
   - Rebased onto latest `upstream/main` (`094de2b` / #1037).
   - Added explicit refusal in `_validate_lorafa_compat` at parse time on `TrainingConfig` and as a runtime guard in `attach_lorafa_optimizer`.
   - Added test `test_lorafa_conflicts_with_vera` verifying both the parse-time `ValidationError` and the runtime `ValueError`.
4. **`stream_layers` caveat**:
   - Documented in `docs/peft-and-efficiency.md` under Compatibility that LoRA-FA is mutually exclusive with `lora.use_vera` and has not been validated under `stream_layers: true` (layer streaming), noting the combination is not recommended.

All 16 tests in `test_issue725_lorafa_optimizer.py`, all 43 tests in `test_issue725_lorafa_wiring_coverage.py`, all 179 snapshot tests in `test_issue621_recipe_snapshot.py`, and `ruff check` pass cleanly. Ready for merge!

MakazhanAlpamys:
Merged as `ebdb4a7b` — thank you, and thank you for the four rounds.

The part that decided it is that the **refusals** turned out to be the work, not the optimizer. All three
conflict guards that survived mutation last round now redden a named test independently:

```
Spectrum / LISA / lora.r=0 full-FT conflict (one mutation each)  -> 1 failed each  KILLED
VeRA refusal, parse-time and runtime, independently              -> 1 failed each  KILLED
the attach itself (trainer.optimizer = optimizer -> pass)        -> 5 failed       KILLED
the whole attach made a no-op                                    -> 11 failed      KILLED
drop the call in sft / pretrain / embedding                      -> 3 failed each  KILLED
drop the betas copy / drop the eps copy                          -> 1 failed each  KILLED
backend gate / task gate / the MLX unsupported-name refusal      -> 1/1/2 failed   KILLED
14 of 14. 289 passed on the 725 + 724 suites, 183 in the blast radius, ruff clean, 15/15 required.
```

You took the stronger of the two options I offered on betas/eps: instead of asserting the optimizer exists,
the test supplies `adam_beta1=0.88 / adam_beta2=0.95 / adam_epsilon=1e-7` through a real `TrainingArguments`
and checks each value arrives — so dropping either copy line fails separately. That is the shape that makes a
guard survive a refactor.

Two things I checked because a rebase had eaten them once before, and both are right: I parsed **both**
`benchmarks/README.md` tables and diffed them (main 41 rows, your branch 43, **LOST = []**, pipe counts
unchanged on every shared row), and your regenerated recipe snapshot is **one insertion, zero deletions** —
`"training.use_lorafa": false` in alphabetical position — rather than a rewritten file.

Your harness runs end to end on CPU and prints its own disclaimer, and the docs carry the measured/analytic
split plus the unvalidated `stream_layers` caveat I asked for.

**One note for the record, and it is mine rather than yours:** I called an earlier head of this PR
merge-ready and then your matrix was red on all nine cells for thirteen hours before I said anything. You
kept going anyway. The `CONTRIBUTORS.md` entry says so.

**Two nits, neither blocking anything now:** the non-AdamW optimizer refusal is real, tested and documented
but absent from the PR body's implementation list — worth a line so the release notes match the code. And
`test_lorafa_conflicts_with_vera`'s runtime half builds a plain LoRA model with a VeRA-shaped config, so it
pins the guard rather than peft's silent degrade-to-AdamW on a real `VeraConfig` model; the guard is
config-driven so that is the right test, I am only noting that the measured behaviour lives in my comment
rather than in a test.

Heads-up for #1080, which also edits `utils/peft_wiring.py`: the two auto-merge cleanly and I inspected the
result — your `use_lorafa` guard lands inside its new `build_loraplus_optimizer` after the `ratio is None`
early return, so it still covers PPO's injection path. Whoever is second rebases and re-runs both the 724 and
725 files.

Issue #964, opened 2026-09-14
The backend-support guard cannot see a config field with a digit in its name


MakazhanAlpamys:
`tests/test_issue755_backend_support_registry.py:150` harvests the config fields an MLX trainer warns about with:

```python
re.findall(r"\b(?:training|data)\.[a-z_]+", source)
```

`[a-z_]` excludes digits. Every registry field to date has had a digit-free name, so this has never mattered — and it is invisible precisely because it produces a **plausible wrong answer** rather than an error.

### How it surfaces

Adding `training.use_fsdp2_compile` to `_MLX_SFT` with the matching warning string in `trainer/mlx_sft.py` fails the guard on a correct entry:

```
AssertionError: mlx_sft.py warns about ['training.use_fsdp'] but the registry does not list them
```

The extractor truncates at the `2`, invents the field `training.use_fsdp`, and then correctly reports that the registry does not contain it. Both halves of the guard behave exactly as written; the input is wrong.

Found by @ege-arhan in #961, which is blocked on it.

### Why this is worth a separate issue rather than a line in that PR

The guard's whole purpose is to hold the declared table and the trainer in agreement. A guard that silently mangles a class of field name **fails on correct code**, which is the failure mode that gets guards deleted rather than fixed — and this repo already deleted one scanner today for exactly that reason (`SCORER_CHANGED_IN_V0_73_2`, v0.73.4, and the `_CONTROL_STRIP_TABLE` name-matching scanner before it).

Other fields with digits exist in `TrainingConfig` and would hit the same truncation if they were ever added to a table: `bnb_4bit_use_double_quant`, `quantization_aware` when set to `"fp8"`, and anything named after a version or bit width.

### Acceptance

- [ ] The extractor accepts digits — `[a-z0-9_]+` — and every existing entry still resolves unchanged
- [ ] A test proves the scanner can fail on a digit-bearing name: add `training.use_fsdp2_compile` to a fixture table with a matching warning string and assert the guard passes, then assert it fails when the registry entry is removed
- [ ] The existing `test_registry_covers_every_field_the_mlx_trainer_warns_about` and `test_the_registry_matches_what_the_backend_trainer_actually_reads` are unaffected
- [ ] Landed **before** #961, which needs it

### Note

[...]

PR #963, opened 2026-09-14
test(lora): hold rank_pattern and alpha_pattern on the seven untested trainers

dchaudhari7177:
Closes #959.

Test-only: one file, no source hunks.

## What was missing

#923 routed all eighteen `build_lora_config` callers through the shared builder, but `_WRAPPERS` held ten of them. The other seven had no behavioural pattern assertion: `asr`, `classifier`, `distill`, `embedding`, `ppo`, `stream_setup`, `unlearn`.

## How each is driven on CPU

The ten existing entries call `_setup_transformers`. Six of the seven build their adapter inside `setup()` instead (or, for `embedding` and `ppo`, a `_setup_transformers` that goes on to things the existing driver does not stub). Driving `setup()` to the end would need datasets, a teacher, a reference model and a Trainer, none of which matter here. So `_stop_after_adapter` patches `peft.get_peft_model` to call the real function, keep the wrapped model and raise a sentinel. Every trainer imports `get_peft_model` lazily, so the patch is what each one resolves. The same `ranks == {"q_proj": 4, "v_proj": 16}` / `rank_pattern` / `alpha_pattern` assertions as the existing test then run on the real PEFT model.

| trainer | driven through | tiny model |
| --- | --- | --- |
| `asr` | `setup()`, `asr_lora: true`; `_load_autoconfig` returns a whisper config, `WhisperProcessor` a stub | 1-layer `WhisperForConditionalGeneration` |
| `classifier` | `setup()`, `num_labels: 2`, `classifier_lora: true` | `LlamaForSequenceClassification` |
| `distill` | `setup()`, `teacher_model` set; stops before the teacher is loaded | `LlamaForCausalLM` |
| `embedding` | `_setup_transformers` | `LlamaModel` |
| `ppo` | `_setup_transformers`, `reward_model` set | `LlamaForCausalLM` |
| `unlearn` | `setup()`, `simnpo`; `load_model_and_tokenizer` patched | `LlamaForCausalLM` |

**`stream_setup`** (the one the issue asked for first) never materialises the base model, so there is no PEFT model to count ranks on. `test_layer_streaming_builds_the_lora_config_with_its_patterns` reuses the `TestQwen35StreamingSetup` fakes for a plain `qwen3` config and asserts on the `LoraConfig` handed to `build_streamed_model`, which is what applies it to the streamed skeleton. No GPU needed.

[...]

Issue #960, opened 2026-09-14
mlx_sft and unlearn still report Loss: 0.0000 where there was no measurement


MakazhanAlpamys:
#908 replaced the fake loss delta with a `loss_summary` helper and migrated the trainers that use HF `log_history`. Two trainers were deliberately left outside it, and they are outside its regression guard too.

`trainer/mlx_sft.py:770-773` and `trainer/unlearn.py:320-321` still build `{"initial_loss", "final_loss"}` by hand with a `0.0` fallback:

```python
losses = captured.get("losses") or [0.0]     # mlx_sft.py:770
```

Neither is in `_MIGRATED_TRAINERS` (`tests/test_issue899_loss_summary.py:8-27`).

### What is and is not already fixed on those backends

The `summary_kind is None and initial == final` fallback at `commands/train.py:75-77` **does** remove the fake arrow, so #899's headline symptom — `Loss: 2.3830 -> 2.3830` presented as a delta — is gone on MLX and unlearn too.

What remains: a run where no loss was captured prints

```
Loss: 0.0000
```

rather than `unavailable`. A plausible number where there is no measurement is the failure mode this project keeps getting bitten by, and it is worse than an error because nobody investigates it. And nothing guards a regression there — mutating either fallback changes no test.

### Why they were left out, and why that was reasonable

Neither backend uses HF `log_history`, so they cannot share the migrated code path as-is. Leaving them out kept #908 reviewable, which was the right call for that PR. This issue is the other half, not a criticism of it.

### Acceptance

- [ ] A run on either backend with no captured losses reports `unavailable`, not `0.0000`
- [ ] Both trainers appear in `_MIGRATED_TRAINERS`, or the table grows a second list for backends that report a summary without `log_history`
- [ ] A mutation of each fallback fails a named test — the anchor counted exactly once, the module still importing, and the failure a behavioural assert rather than a collection error
- [ ] `initial_loss` / `final_loss` keep their numeric meaning for the tracker and for `soup sweep` ranking, as documented at `loss_summary.py:29-36`

Follow-up to #899, closed by #908 by @taljeon.

Rayan-and-beyond:
I can take this. I’ll remove the fake zero-loss fallbacks for MLX SFT and unlearn while preserving numeric loss values when measurements exist, and add focused behavioral regression coverage without requiring MLX/GPU execution.

Rayan-and-beyond:
[...]

PR #965, opened 2026-09-14
feat(doctor): declare the four remaining loadable CUDA-only fields MLX ignores

dchaudhari7177:
Refs #958. The table grows by four, not nine; the reason is below.

## What I measured before adding rows

The issue checked the nine fields by grepping the three MLX modules. I re-ran that (every field has 0 hits in `mlx_sft.py`, `mlx_optim.py` and `mlx_masking.py`) and then also loaded a config setting each one on `task: sft, backend: mlx`:

| field | loads on mlx? |
| --- | --- |
| `use_mod` | yes |
| `moe_lora` | yes |
| `quantization_aware` | yes |
| `use_fsdp2_compile` | yes |
| `use_cut_ce` | **refused**: "v0.28.0 features ['use_cut_ce'] are not supported on the Apple Silicon mlx backend" |
| `kernel_auto_compose` | **refused**, same validator |
| `activation_offloading` | **refused**, same validator |
| `use_longlora` | **refused**: "LongLoRA is not supported on the mlx backend" |
| `packing_cross_doc_attn_mask` | **refused** on every backend (#709) |

A zero-hit grep shows a field isn't read, but not that it's *ignored*. The five refused fields never reach MLX. A config that sets them fails at load, so `doctor --config` can never report them, and an `IGNORED` row would say MLX silently drops a setting the schema actually rejects. So:

- **The four that load** join `_MLX_SFT` with a reason each, and `mlx_sft.py::_check_unsupported` warns for them (`quantization_aware` on truthiness, so `'fp8'` warns too). The clean-config all-clear reads `None of the 13 setting(s)`, not 18.
- **The five that are refused** get `test_fields_refused_at_load_on_mlx_are_not_declared_ignored`, parametrized over all five: loading must exit, and no registry row may name the field. If one of them starts loading on mlx, the test fails, and that's the moment to add its row.

## The two review notes

1. **`use_liger`**: taken. The reason now ends "; soup train also refuses the run on a non-CUDA device before MLX starts".
2. **The indentation-dependent fixture**: the suggested replacement doesn't work as written. `config_at` splices the string into a `textwrap.dedent`ed f-string, so the second line has to repeat the template's 12 spaces, and `"  use_liger: true\n  neftune_alpha: 5"` fails with `yaml.parser.ParserError: expected '<document start>'`. I fixed `config_at` instead: it now joins lines and indents each block with `textwrap.indent(textwrap.dedent(text), "  ")`. The simple form works, and no multi-line block can depend on template width again. All other callers pass single-line blocks and are unchanged.

## One guard fix the new entries needed

`test_registry_covers_every_field_the_mlx_trainer_warns_about` extracts dotted names with `[a-z_]+`, which read `training.use_fsdp2_compile` as `training.use_fsdp` and failed. It now uses `[a-z0-9_]+`. No field in the table had a digit before, so the regex never had to handle one.

## Checks

- `test_the_four_loadable_mlx_gaps_from_958_are_reported_together` sets all four plus `seed`, asserts exactly those five rows, and asserts the table is 13 wide.
- The existing parametrized guards now include the four new `trainer_reads=True` entries automatically: warning-list coverage, `_registry_drift`, and deleting-a-warning-fails.

```
$ pytest tests/test_issue755_backend_support_registry.py
39 passed, 6 failed
```

The six failures are the `doctor(...)` leg (`test_doctor_config_names_the_ignored_field_and_its_reason`, `test_all_clear_states_what_was_checked`, the three widths, and `test_doctor_without_config_does_not_read_one`). They fail identically on `main` here: `doctor` exits 1 at `doctor.py:329` because this Windows environment is missing a core dependency. CI is the check for those. `test_all_clear_states_what_was_checked` derives its count from `len(unsupported_for(...))`, so it needs no edit. `ruff` 0.9.4 (the pre-commit pin) is clean on the three files.


MakazhanAlpamys:
@dchaudhari7177 — **I have to hold this, and the reason is a collision I created 14 minutes before you opened it.** The timeline, from the API rather than from memory:

```
06:09  #961 opened by @ege-arhan against #958
06:27  I unblock their CI and confirm the issue is assigned to them
06:50  I post a blocking review on #961 publishing the "four, not nine" reachability finding
06:51  I post the same correction on #958
07:05  #965 opens implementing exactly that
```

So the scope this PR is built on is something I published minutes earlier on someone else's assigned issue, without saying "this stays with @ege-arhan". That is the third time today I have created a race by publishing work-shaped information in the wrong place, and I am not going to resolve it by rewarding speed again.

**#958 is assigned to @ege-arhan and #961 is open against it. They get first refusal.** Their PR is blocked on four items, and **two of those four are my errors, not theirs** — my issue's premise was wrong and one of its notes actively broke a test they followed. Taking the work off them while they are blocked by my mistakes would be indefensible.

I have asked them on #961 whether they want to fold the corrected scope in. **If they have not said by 2026-09-16 18:00 UTC, this PR takes over** and I will say so here and there. That is a real date; I have let announced dates pass on this repo before and I am not doing it again.

**None of that is a criticism of the work.** You did the thing I should have done when I filed #958: you loaded a config setting each field rather than trusting a grep, and your table is right — `use_mod`, `moe_lora`, `quantization_aware` and `use_fsdp2_compile` load; the other five are refused before the doctor can ever see them, four by `_validate_v028_speed_memory_supported_tasks` and `packing_cross_doc_attn_mask` on every backend. `Refs #958` rather than `Closes` was also the correct call and I noticed it.

**Something genuinely unclaimed, and it blocks #961 rather than duplicating it: #964.** The backend-support guard's extractor is `re.findall(r"\b(?:training|data)\.[a-z_]+", ...)` at `tests/test_issue755_backend_support_registry.py:150`. `[a-z_]` excludes digits, so `training.use_fsdp2_compile` — the first registry field with a digit in its name — is harvested as `training.use_fsdp` and the guard **fails on a correct entry**. Whoever adds that field cannot land until it is fixed, so it is on the critical path for both #961 and this PR, and it is nobody's.

It also has a trap worth your kind of measurement: widening to `[a-z0-9_]+` must be checked against the real `TrainingConfig` and `DataConfig` field names, not just against the previously-failing case — a regex that starts inventing *different* wrong names is no better than the current one.

Say the word and #964 is yours. Your #963 is already in review, and if #965 does take over on the 16th, none of this work is wasted either way.

MakazhanAlpamys:
@dchaudhari7177 — **an update you will not enjoy, and it is my fault for the fifth time today.**

I offered you #964 in the comment above at 07:46. At 08:23 @00200200 opened **#966** against it, with the work done — including the validation-against-real-field-names part I flagged as the trap. You had not answered, the issue was unassigned, and an offer is not a claim, so I have assigned it to them and it is theirs.

**The mistake is precisely the one I said I would stop making.** On #953 this morning I wrote: *"when I suggest work in a PR review, I will also comment on the issue itself, so the claim is visible in one place rather than two."* I then offered you #964 **in a comment on this PR** and never touched #964. Same error, same day, four hours after writing the remedy down.

So I am changing the mechanism rather than promising again: **from now on, if I offer you work, I assign the issue to you at the moment I offer it**, and you can decline by saying so. An offer that costs nothing to make is one I keep failing to protect.

**What I owe you now, and this time it is reserved before I post it.** Say yes to either and it is already yours:

- **#960** — `trainer/mlx_sft.py:770-773` and `trainer/unlearn.py:320-321` still build `{"initial_loss", "final_loss"}` by hand with a `0.0` fallback, so a run that captured no loss prints `Loss: 0.0000` rather than `unavailable`. Neither is in `_MIGRATED_TRAINERS`, so nothing guards a regression. This is the follow-up to #899, which you would recognise from #908's review.
- **#959 is already yours** and #963 is in review and merge-ready pending CI, so you are not short of landed work today.

And the standing position on this PR is unchanged: **#965 is held, not rejected.** If @ege-arhan does not answer by 2026-09-16 18:00 UTC, it takes over #958 and I will say so here and there. Your measurement — loading a config per field rather than trusting the grep — is what corrected my own issue, and that stands whichever PR lands.

MakazhanAlpamys:
@dchaudhari7177 — **update inside the deadline I set, as promised.**

I wrote on 09-14: *"If they have not said by 2026-09-16 18:00 UTC, this PR takes over and I will say so here and there. That is a real date; I have let announced dates pass on this repo before and I am not doing it again."*

**@ege-arhan answered on 09-15 at 04:39, well inside it:** *"I will continue — please keep it with me. After #966 merges I will rebase and drop my inline extractor hunk."* #966 merged this morning and their #961 is green on its rebased head.

So **the hold stands and #961 has precedence.** That is the outcome the date existed to decide, and I am telling you on the day rather than on the deadline.

**Two corrections I owe you, both mine.**

First, when I told you this was held I said *"#958 is assigned to @ege-arhan"*. The API disagreed: `assignees` was empty then and is empty now. `gh issue edit --add-assignee` returns success and silently assigns nobody unless the person is a collaborator or has commented on that issue, and their work is all on the PR. I have now recorded the holder in a comment on #958, since the field cannot carry it.

Second, in that same comment I said *"from now on, if I offer you work, I assign the issue at the moment I offer it"* — and then #959 and #960, both offered to you in that very sentence, sat unassigned. #959 is now labelled and named to you in-thread. **#960 is not yours any more and I should have told you sooner**: @Rayan-and-beyond claimed it on 09-15 and has PR #986 open against it. You were offered it first and did not answer, so I am not calling that unfair — but you should hear it from me rather than find the PR.

**Where that leaves this PR.** It overlaps #961 on the same table. Once #961 lands, rebase and this becomes the smaller remainder — the four fields it declares that #961 does not. Say the word if you would rather close it and fold the remainder into a fresh PR; either is fine and I will not treat a close as withdrawing.

#963 is separately green and waiting on the two-line fixture fix I described there.

MakazhanAlpamys:
@dchaudhari7177 — **#961 is merged and I am closing this, as I said I would on 09-15. You deserve the full reasoning and the measurements, not just the outcome.**

## Why it could not land alongside #961

Not a matter of taste. I measured the merge:

```
git merge-tree --write-tree --name-only f317c437 e9715cb9
  CONFLICT (content): src/soup_cli/config/backend_support.py
  CONFLICT (content): tests/test_issue755_backend_support_registry.py
  Auto-merging  src/soup_cli/trainer/mlx_sft.py        <- no conflict
```

Identical scope — `use_mod`, `moe_lora`, `quantization_aware`, `use_fsdp2_compile` — both rewriting the same `use_liger` reason line and inserting four `SupportEntry` blocks in the same slot.

**And the file that does NOT conflict is the dangerous one.** You insert your four warn branches after `use_liger`; #961 inserts its four after `neftune_alpha`. Git merges that happily into **eight branches covering four fields, each warned twice** — and the bidirectional guard would not catch it, because it collects warned names into a `set`. A silent double-warning in `soup doctor --config` output is a worse outcome than either PR alone, and no tooling would have told us.

## The order, from the API rather than from memory

```
06:09  #961 opened by @ege-arhan against #958
06:50  I post the "four, not nine" reachability finding on #961
06:51  I post the same correction on #958
07:05  #965 opened, implementing exactly that
```

**The race is mine.** I published a work-shaped finding on someone else's assigned issue and did not write "this stays with @ege-arhan" anywhere you would have seen it. You then built the right thing, quickly, off information I had put in front of you. I set a real deadline of 2026-09-16 18:00 UTC for @ege-arhan to confirm; they answered on 09-15 at 04:39, inside it. So the hold held on its own terms, which is the only reason I am comfortable with this outcome.

## What is genuinely better in your version, and what I would like you to land

Your `_MLX_SFT` comment block naming the five *excluded* fields and why each is refused earlier is better than what merged. #961's reasoning for those exclusions lives in its changelog fragment and its PR thread; yours is **in the source, where the next person editing that table will actually look.**

**Please open a small PR with just that comment block.** It is not a participation trophy — it is the part of your work that survives on the merits, and I would rather have it in the tree than lose it to a close. If you would rather not, that is fine too and I will lift it myself with credit to you rather than let it disappear.

## And a real issue, offered to you first

**#1001** — `backend_support.py:157` prints a hardcoded `commands/train.py:1235` through `soup doctor --config`, and nothing pins it. It has **already drifted once**, inside #961's own history: the same string read `:1186` at one head and `:1235` at the next, because the file moved underneath and nothing said so. Any edit above `train.py:1222` silently makes the printed advice false.

It is in the file you already know, it is real, and the issue body names the two acceptable fixes and one to reject deliberately. It is yours if you want it. **If you would rather have something with no history attached to it, say so and I will find you something else** — declining a consolation is not a mark against you with me.

## The thing I keep owing you

This is the second time in two days you have built the right thing and been told it cannot land, and both times the cause was me publishing work in the wrong place. On 09-14 I offered you #964 *in a comment on this PR* and never touched the issue, and @00200200 opened #966 against it four hours later with the work done. I had written down the remedy for exactly that mistake — *"when I suggest work in a PR review, I will also comment on the issue itself"* — earlier the same morning, on #953.

I am not going to promise it a third time in prose. #1001 is filed as an **issue** with your name in the body, which is the form of the promise that actually holds.

MakazhanAlpamys:
Closing as explained in the comment above — #961 landed the same four fields and the two cannot coexist (measured: two content conflicts, plus an `mlx_sft.py` auto-merge producing eight warn branches for four fields that the guard's `set` would not catch). The close is on sequencing, not on quality. #1001 is offered to you, and I would still like your `_MLX_SFT` comment block as its own PR.

PR #966, opened 2026-09-14
test: preserve digits in backend-support warning fields

### LINKED ISSUE #964: The backend-support guard cannot see a config field with a digit in its name

00200200:
The backend-support warning guard truncated `training.use_fsdp2_compile` to `training.use_fsdp`, rejecting a correct registry entry and blocking #961.

Accept digits in dotted field names and validate the harvested names against `TrainingConfig` and `DataConfig`. Regression fixtures verify that digit-bearing warnings pass with a matching registry entry and fail when it is removed. Additional cases reject nonexistent numeric suffixes.

Closes #964. This changes the guard only; the registry and trainer runtime behavior are unchanged.

Validation on Python 3.12:
- Both digit-bearing regression cases failed before the regex fix; all 45 backend-support tests pass after it.
- Started the full suite: 4701 passed, 40 skipped, 1 failed before interruption. The unrelated `test_issue204_langfuse_pull.py::test_control_the_leak_scan_sees_a_token_rich_highlighted_and_folded` failure reproduces independently on clean `origin/main` (2432d69). The complete suite is not claimed green.

No user-visible change, so no changelog fragment. Prepared with Codex assistance.


MakazhanAlpamys:
@00200200 — welcome, and **your CI was blocked rather than slow.** A first-time contributor's workflow runs need manual approval; I approved yours (`34822482159`) and it is queued. That is my job to notice, not yours to chase.

**One thing you should know before you invest more time, because it is not about your work.** I offered #964 to @dchaudhari7177 at 07:46 today, in a comment on a different PR rather than on the issue itself. You opened this at 08:23. They have not answered.

An offer is not a claim, the issue was unassigned, and you have a PR with the work done — so **this one is yours** and I have assigned #964 to you. But I want the sequence on the record rather than discovered, because I have created this exact race three times today and I am not going to pretend it was chance.

From the body, you did the part I was most concerned about: **validating the harvested names against the real `TrainingConfig` and `DataConfig` fields**, not just widening `[a-z_]` to `[a-z0-9_]`. That was the trap I named in the issue — a regex that starts inventing *different* wrong names is no better than one that truncates. And the "fails when the registry entry is removed" fixture is the half that matters: this guard's whole value is that it can fail, and it currently finds zero offenders, so a can-it-fail test is the only thing holding it up.

Full review today. **#964 is on the critical path for #961**, so this is not a small errand — it unblocks someone else's PR.

Two mechanical things that have failed CI for four PRs today, worth checking before the runner tells you:

- the changelog fragment must start with `- ` (a UTF-8 BOM counts as the first character and will fail it) and reference `#966`, with credit `(#964 by @00200200 in #966)`;
- any assertion on `result.output` or captured stdout needs the ANSI-strip helper from `tests/conftest.py`, because Rich colours output on Linux and macOS runners but not on Windows.

Neither applies if your change is a pure AST/regex test, which it looks like it is.

MakazhanAlpamys:
Verified `3d56cb4f`. **Merge-ready**, and this unblocks #961 and #965.

**The round trip is the review, and it passes on the real files.** I added `training.use_fsdp2_compile` to `_MLX_SFT` plus the matching warning in `mlx_sft.py`, then ran three states:

| state | result |
|---|---|
| your guard, registry entry present | 46 passed |
| `origin/main`'s guard, same entry | `mlx_sft.py warns about ['training.use_fsdp'] ...` — #964 reproduced byte-for-byte |
| your guard, entry removed, warning kept | `mlx_sft.py warns about ['training.use_fsdp2_compile'] ...` |

The third is the one I most wanted to see: it names the **full** field, so the guard fails for the right reason rather than for the truncated one. Both probe edits reverted, worktree clean.

And the existing entries are untouched — harvesting the real `mlx_sft.py` with both regexes gives the identical six fields, `only-old: []`, `only-new: []`.

**You did the part the issue warned about, and it is load-bearing.** `tests/...:174-179` builds `known` from `TrainingConfig.model_fields` / `DataConfig.model_fields` and asserts `warned - known == []`. I checked it discriminates on a genuinely hard case rather than a strawman: `data.val_split2` is rejected while `val_split` **is** a real field, so the test cannot pass by accident of the suffix looking odd.

Mutations: both anchors die, and by **different** tests, which is what tells me the two halves are independently held.

- `[a-z0-9_]+\b` back to `[a-z_]+` (anchor counted 1) → 3 failed, including the `bnb_4bit_use_double_quant` case
- `unknown = sorted(warned - known)` to `unknown = []` (anchor counted 1) → 2 failed, disjoint from the first set

Baseline on `origin/main` with the new tests kept and the guard reverted: 4 failed. Every new parametrized case discriminates; none is vacuous.

Also verified: one removed line, the replaced regex, nothing else; diff is 1 file `+37/-1` with registry and trainer untouched, exactly as the body says; `merge-tree` clean against current `main`; 45 passed on the file and 397 across the related suites; ruff clean; and the tests are order-independent — forcing the two new ones ahead of the guard still passes, so the `monkeypatch` teardown restores properly.

## One note, non-blocking, in the same family as the bug you fixed

You added a trailing `\b` that #964 did not ask for. Since `[a-z0-9_]` is a subset of `\w`, it can only ever **reject** a match:

```
'see training.seedMLX for details'   old -> ['training.seed']   new -> []
```

The old behaviour produced a wrong-but-visible name; the new one drops it silently. That is the same class of failure as the truncation — invisible rather than wrong — just pointing the other way. It is unreachable today (Pydantic fields are snake_case and no such prose exists) and the `\b` is otherwise inert, so I am not holding the merge. `[a-z0-9_]+` without it would be strictly better.

Two smaller things for whenever you next touch the file: three tests call `test_registry_covers_every_field_the_mlx_trainer_warns_about()` directly, which works only because that function takes no fixtures **and** imports `unsupported_for` inside its body so the `monkeypatch` lands — both load-bearing and undocumented, so a one-line comment would pay for itself. And `test_warning_guard_rejects_unknown_digit_suffixes` relies on the real registry never containing `val_split2`, which is safe but couples a fixture test to production data.

## No changelog fragment, and I agree with your reasoning

You wrote *"no user-visible change, so no changelog fragment"*. `CONTRIBUTING.md:350` says "for each **user-visible** pull request", and the precedent backs you: the repo-wide AST ratchets #778 and #789 carry none, and my own test-only commit in this same file this morning carries none either.

## On CONTRIBUTING.md, since this is your first PR

Conventional prefix, the documented lint command run and quoted, TDD order claimed **and true** — I reproduced it, both digit cases fail before the regex change. And you wrote *"the complete suite is not claimed green"* after an interrupted run. That disclosure is rarer than it should be and it is the reason your other claims were quick to check.

One correction: the `test_issue204_langfuse_pull` failure you reported as pre-existing **passes here**. It is a Rich-highlighting test, so it is most likely colour or terminal-width dependent on your machine — this repo has been bitten by that class four times in CI. Out of scope for this PR, but worth knowing it is environmental rather than a broken `main`.

Merging as soon as CI is green. Thank you — this is a good first contribution and it clears the way for two other people's PRs.

MakazhanAlpamys:
Merged as `ede67683`. Thank you, and welcome — this is your first contribution here.

Merged on the verdict recorded at `3d56cb4f`, not re-derived: all 13 required contexts green on that exact SHA, `mergeStateStatus` clean, and re-tested on the merged tree because `main` had moved (66 passed there).

**The half I want on the record is the one you added beside the obvious fix.** Widening `[a-z_]` to `[a-z0-9_]` closes the reported symptom — `training.use_fsdp2_compile` harvested as `training.use_fsdp`, `training.bnb_4bit_use_double_quant` as `training.bnb_`, so for any field with a digit the guard compared an **invented** name against the registry and could neither confirm a real entry nor catch a missing one.

But a widened regex that starts inventing a *different* wrong name is no better than the narrow one it replaced. Cross-checking every harvested name against the real `TrainingConfig` / `DataConfig` `model_fields` — failing with `warning list names fields that do not exist: [...]` — is what turns a widening into a guarantee. That was #964's acceptance box 3 and the explicit note on it, and it is the part a competing patch for the same issue did not have.

I compared the two implementations line by line before choosing, because #961 also carried the widening:

```
#961 : re.findall(r"\b(?:training|data)\.[a-z0-9_]+",  arg.value)
#966 : re.findall(r"\b(?:training|data)\.[a-z0-9_]+\b", arg.value)
       + the model_fields cross-check
```

Yours also closes the match with a trailing `\b`. Taking this one first was not a coin flip.

This unblocks #961 and #965, both of which need digit-bearing fields to be harvestable at all — I have told #961's author to rebase onto you.

Credited in `CONTRIBUTORS.md` (`1753ba26`); you are the 64th contributor. #964 is closed with credit.

PR #969, opened 2026-09-14
fix(recipe): report provider call failures

### LINKED ISSUE #930: soup data recipe --execute reports success and writes empty generations when the provider endpoint is unreachable
## What

`soup data recipe --execute` with a **valid** provider pointed at an unreachable endpoint reports success and writes empty generations for every row.

Reproduced against #922's branch (`f4a1353a`), which is where `--provider` first exists:

```
soup data recipe r.yaml --execute --output out2 \
    --provider ollama --base-url http://127.0.0.1:59999

Recipe validated. 3 node(s), 2 edge(s).
Topological order: seed1, llm1, samp1
Recipe executed. 3 node(s) completed.
EXIT_CODE=0

out2/samp1.jsonl:
{"text": "keep", "llm1": ""}
{"text": "two",  "llm1": ""}
```

Nothing on port 59999. Every generation is the empty string, the summary says three nodes completed, and the exit code is 0.

## Why it matters

This is the same failure class #817 describes — **a success message over unusable training data** — reached by a different route. #922 closes the offline-default hole (a recipe with `llm_text` nodes now refuses without `--provider` or an explicit `--offline`). This one survives it: the user *did* configure a provider, so the guard passes, and then every request fails silently.

The output is worse than the offline placeholder, not better. `llm_text(offline): ...` is obviously synthetic and a human spots it immediately; `""` looks like a model that had nothing to say, and a dataset of empty completions will train without complaint.

## Root cause

`utils/recipe_run.py::_node_llm_text` keeps rows whose `reply["text"]` is `""`. There is no distinction between "the model returned an empty completion" and "the call failed", and no counter of failed calls reaches the summary line.

## Scope

**Pre-existing and not #922's fault.** I found it while reviewing that PR and said there I would file it separately rather than widen their scope. It cannot be reproduced on `main` through the CLI, because `--provider` does not exist there yet; on `main` the same code path is reachable only through the `run_recipe(judge_provider=...)` Python API.

**This should land after #922**, since it builds directly on the flags that PR introduces.

## Acceptance criteria

- [ ] A provider call that fails (connection refused, timeout, non-2xx) is counted, not silently absorbed.
- [ ] The summary line reports the failure count, and the command exits non-zero when **every** call failed. A partial failure is a judgement call — state which you chose and why in the PR.
- [ ] A test that stands up no server, points `--base-url` at a closed port, and asserts the non-zero exit and that the message names the endpoint. The control that matters is the inverse: a provider that legitimately returns an empty completion for one row must still be distinguishable from a dead endpoint, so include a test for that too. A fix that simply treats every `""` as an error would pass the first test and be wrong.
- [ ] The same treatment for `judge` nodes, which reach a provider by the same route.

## Notes

Do not "fix" this by dropping empty rows. Silently discarding data is how the original defect was created — the value of this fix is that the operator learns their endpoint was down, not that the file looks tidier.

taljeon:
## What does this PR do?

Closes #930.

Makes live Data Recipe provider failures observable instead of silently turning them into empty training rows:

- `llm_text` and `judge` nodes count transport, non-2xx, malformed-response, and non-string-response failures per node
- when every attempted call for a provider node fails, the recipe exits non-zero, records the failed node and counts in its checkpoint, and names the credential-free endpoint origin
- partial failures keep rows from successful calls and report the aggregate call/failure count in the completion summary
- a legitimate empty string remains a successful completion; it is not reclassified as an outage

The provider factory gets an opt-in strict mode for the recipe runner. Its existing best-effort callers keep the old empty-result behavior by default.

For partial failures I chose to complete with the usable rows and a loud count. A single transient call should not discard unrelated successful generations, while an all-failed node has no usable provider evidence and must fail the run.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [ ] `pytest tests/ -v` passes (the local environment lacks parts of the optional dev/train stack; directly related results are below)
- [x] Updated relevant docs (`README.md` and the matching page under `docs/`) if needed (`docs/data.md` now documents failure counts and the all-failed exit-1 contract)

## Draft review follow-up

- Normalized all terminal-output assertions with the repository's ANSI-strip plus whitespace-collapse pattern; the focused suite now passes at `COLUMNS=60`.
- Documented the new exit-code and partial-failure behavior in `docs/data.md`.
- Preserved the safe endpoint origin in checkpoint `failed_reason` while still stripping URL credentials, paths, and queries.
- Added `AttributeError` parity to the vLLM malformed-response path and tightened the strict-mode type test to the implementation's exact error.
- Merged current `main@2e25fc3` without force; its distillation and streaming changes do not overlap this six-file diff.

## Validation

- Closed-loopback-port CLI regression: red on current `main` (exit 0 with empty output), green here (exit 1 with `2/2` failures and the endpoint named)
- Focused #817/#930 provider suite: 24 passed at both default width and `COLUMNS=60`
- Every test file that directly references `recipe_run`, `data_forge`, or `make_judge_provider_fn`: 597 passed, 58 skipped
- Changelog-fragment validation: 21 passed
- Full repository Ruff
- `compileall` on all four changed Python files
- `git diff --check`
- Full-suite collection was attempted but stopped before execution on 22 files because this local environment does not contain `numpy`, `torch`, or `fastapi`; no dependency was installed or result represented as a product failure

## AI assistance

I used an AI coding assistant to help implement the failure-accounting path and tests. I reviewed the complete diff, verified the current source and issue scope, and ran the validation above locally.


MakazhanAlpamys:
**Draft review** — you have not marked this ready, so treat this as early feedback rather than a gate. No blocking defect; two MEDIUM items I would want closed before it comes out of draft.

## The acceptance criteria are met, verified against the real repro

Not mocked — a closed port:

```
soup data recipe recipe.yaml --execute --output outreal --provider ollama --base-url http://127.0.0.1:59999
Recipe failed: llm_text node 'gen1': all 10 provider calls failed for http://127.0.0.1:59999
EXIT=1
```

And the inverse control holds: a legitimate `{"text": ""}` still counts as success, pinned by `test_recipe_cli_preserves_legitimate_empty_completion`. That control is what makes the fix trustworthy rather than merely loud.

**`_provider_endpoint_label` is a good catch of yours** — credentials, path and query all stripped. Verified live: `http://user:pw@127.0.0.1:11434/x` prints as `http://127.0.0.1:11434`, and it is pinned by a test. A failure message is exactly where a password leaks into someone's CI log.

Both mutations die. `failures += 1` to `failures += 0` in the `llm_text` transport branch fails two tests behaviourally —

```
assert '2 provider calls, 1 failure' in '... 2 provider calls, 0 failures.'
```

— and the same edit in `_node_judge` fails a third. Baseline on `origin/main`: 10 failed, 13 passed, no collection errors.

## What partial failure actually does, measured

Ten prompts, calls 2/5/9 raising:

```
"status": "completed"
"node_row_counts":         {"seed1": 10, "gen1": 7, "samp1": 7}
"provider_failure_counts": {"gen1": 3}
CLI: "Recipe executed. 3 node(s) completed. 10 provider calls, 3 failures."   EXIT 0
```

The three rows are **dropped**, the count is surfaced, and the summary is green. Your PR body states that choice explicitly — *"for partial failures I chose to complete with the usable rows and a loud count"* — which is what #930's acceptance asked for, and I am fine with the call.

One gap worth naming rather than fixing here: the summary says "3 failures" but never "3 rows dropped". The shortfall lives only in `node_row_counts`, which the CLI does not print, and `soup data recipe` has no `--resume`, so a re-run re-attempts everything. A yellow-rather-than-green summary, or a `--max-failure-fraction`, would close it.

## MEDIUM 1 — four new tests are terminal-width fragile

They use `strip_ansi`, which is right, but **do not collapse whitespace** — and Rich wraps as well as colours:

```
FORCE_COLOR=1 TERM=xterm-256color   ->  23 passed
COLUMNS=60                          ->  4 failed
E  assert '2 provider calls, 1 failure' in '... 2 provider calls, 1 \nfailure.'
```

All four are new in this PR. CI runs at 80 columns so it is green today, which is exactly how this class survives. The canonical helper in this repo is strip **plus** collapse — see `tests/test_auto_tuning.py` and `tests/test_v07302.py`; `re.sub(r"\s+", " ", strip_ansi(result.output))` fixes all four.

Note the same latent fragility in the total-failure test: its message is 92 characters and already wraps at 80. It passes only because the break happens to land after `for`.

## MEDIUM 2 — the exit-code contract changed and the docs do not say so

Your checklist says *"not needed; existing CLI options and workflow are unchanged."* The options are, but **`soup data recipe --execute` can now exit 1 where it previously exited 0** on a live-provider outage. `docs/data.md:961-982` documents this runner and says nothing about provider-failure accounting or the new exit code. Two sentences.

## Three LOW

1. **`data_forge.py:728`** — the vLLM parse-error `except` tuple was not widened to include `AttributeError`, while ollama (`:641`) and anthropic (`:690`) were. Harmless today, since `recipe_run`'s broad `except Exception` still counts it, but the asymmetry is untested and will read as an oversight later.
2. **`recipe_run.py:732`** — the pre-existing `_redact_exc_message` path-stripper mangles the new URL when it is persisted: `failed_reason` reads `"... failed for htt127.0.0.1:11434"`. Console output is correct; only the checkpoint JSON is garbled. This PR is what put a URL into that message, so it is the natural place to whitelist `scheme://host:port`.
3. **`test_provider_factory_rejects_non_boolean_strict_mode` does not discriminate.** `pytest.raises(TypeError, match="raise_on_error")` is satisfied on `main` by Python's own `got an unexpected keyword argument 'raise_on_error'`. Tighten the match to the implementation's wording.

(`test_provider_factory_default_keeps_legacy_empty_failure_result` also passes on `main`, and that one is correct — it is a deliberate no-regression pin for existing callers.)

## Housekeeping

One removed test line, and it is a rewrite not a deletion — the module docstring gaining `#930`. Fragment at `changelog.d/0.75.0/969.fixed.md`, starts with `- `, no BOM, correct credit. `strip_ansi` imported and used at all nine output assertions. Merge-tree clean. 23 passed on the file, 635 across the seven files touching `recipe_run` / `data_forge`. Ruff clean.

The two MEDIUM items are each a few lines. Mark it ready when they land and I will run the full pass.

taljeon:
Addressed both MEDIUM items and the three LOW notes at `f1e9e3f` (`f9b80d0` plus a no-force merge of current `main@2e25fc3`):

- all terminal-output assertions now use the repository's ANSI-strip plus whitespace-collapse pattern; the focused file passes 24/24 at both default width and `COLUMNS=60`
- `docs/data.md` now states the all-failed exit-1 contract and partial-failure/empty-completion behavior
- vLLM now catches `AttributeError` with the other malformed-response shapes, with a discriminating strict-mode regression
- checkpoint redaction preserves the already-sanitized endpoint origin while still dropping URL userinfo, path, and query
- the non-boolean strict-mode test now matches the implementation's exact error, so the pre-flag Python keyword error cannot satisfy it

The complete directly related set passes 597 with 58 skips; changelog validation passes 21; repository Ruff, compileall, diff, metadata, body, and publication checks are clean. The PR body is updated with the documentation and current results. Marking it ready now for the full pass—thanks for the precise width and checkpoint catches.

MakazhanAlpamys:
**Re-reviewed at `f1e9e3f`. All five items are closed, and three of them are closed better than I asked for.** This is merge-ready; the only thing left is CI, which is still pending on this head.

## MEDIUM 1 — terminal width

Fixed, and I pushed on it harder than the original finding. `_terminal_text` is semantically identical to the repo's canonical `_plain` in `tests/test_v07302.py:66`, applied at all ten sites:

```
COLUMNS=60  ->  24 passed     (was 4 failed)
COLUMNS=40  ->  24 passed
COLUMNS=30  ->  24 passed
COLUMNS=20  ->  2 failed
FORCE_COLOR=1 TERM=xterm-256color  ->  24 passed
```

The 20-column failure is **not a finding against you**. At that width Rich hard-breaks inside the URL token itself (`http://127.0.0.1:55 452`), which no whitespace-collapse can repair. That is the class limit of this repo's convention, and worth writing down somewhere other than a PR comment.

No matched assertion still reads raw output — the two remaining `result.output` references are assert *messages*, which is correct.

## MEDIUM 2 — docs

`docs/data.md:977-980` covers all four things the contract change touches: the new exit 1, the endpoint naming, partial-failure accounting, and the empty-completion carve-out. That is more than the two sentences I asked for.

## The three LOW items

**LOW 1 (vLLM `except` tuple)** — all three branches now identical, and the new test is not decorative: reverting `data_forge.py:728` to the old four-tuple fails `test_vllm_strict_mode_classifies_missing_json_method_as_malformed` with the real `AttributeError`.

**LOW 2 (`failed_reason` URL mangling)** — fixed via a sentinel pass at `recipe_run.py:187-196`, and I verified it against a live closed port rather than through the suite. The checkpoint now records the intact URL. I also checked the guard does not weaken the redaction it wraps:

| input | output |
|---|---|
| `https://user:secret@api.example.com:8443/v1/chat?key=abc123` | `https://api.example.com:8443` |
| `http://127.0.0.1:11434 and /home/alice/secret/data.jsonl` | `http://127.0.0.1:11434 and data.jsonl` |
| `C:\Users\alice\secrets\key.pem missing at http://localhost:8000/v1` | `key.pem missing at http://localhost:8000` |

Userinfo, path and query all discarded; both path-redaction styles unaffected; multiple URLs handled. The one theoretical hole — a message containing the literal sentinel — is unreachable, since node names are lowercase-gated and every `ProviderCallError` message is a constant, and the substituted value is an already-redacted origin regardless.

**LOW 3 (vacuous match)** — `match="^raise_on_error must be a bool$"` now discriminates. Run against `origin/main`'s source it fails on CPython's own `got an unexpected keyword argument`, which is what the old spelling was silently accepting.

## Mutation still dies

Targeted by index with the preceding `_LOG.debug` line asserted first, so it is provably the `llm_text` transport branch and not the judge one:

```
failures += 1  ->  failures += 0
E   assert 0 == 1
FAILED test_recipe_cli_closed_provider_fails_with_count_and_endpoint
FAILED test_recipe_cli_reports_partial_failure_and_keeps_successful_empty_row
2 failed, 22 passed
```

Both are behavioural: the checkpoint's `provider_failure_counts` and the summary line.

## Merge state

`merge-tree` clean against `b0a63382`, no file overlap. On the merged tree: the four ratchets 97 passed, the three provider/forge suites 245 passed, `-k "forge or recipe or data_forge"` 2431 passed, ruff clean.

No test deleted and no assertion loosened — every removed line is a same-line replacement, and `strip_ansi(...)` -> `_terminal_text(result)` is strictly stronger. 23 tests to 24. Fragment correct, no BOM, credit right.

**All 14 checks are pending on `f1e9e3f`.** I will merge as soon as they are green on this head — nothing else is outstanding.

One thing I am leaving alone rather than asking you to change: with no `--base-url`, the total-failure message reads `ollama default endpoint` instead of the concrete `http://localhost:11434` the provider actually used. Correct, marginally less actionable, and not worth another round.

MakazhanAlpamys:
Merged as `0c482aab`. Thank you — two of yours landed in this pass.

**A failure that reports as success is worse than a crash**, because nothing prompts anyone to look: a provider call that failed mid-recipe was swallowed and the run reported success having produced nothing from that node. That is the whole argument for this fix and it is why it was worth five rounds.

All five review items are closed and **three of them are closed better than I asked for**.

Before merging I composed this with the other nine PRs that were ready and ran the affected suites plus the repo-wide ratchets on that tree: **967 passed, zero failures**. That step is not ceremony — it caught a real break in a tenth PR (#963), which I held rather than merged. Yours composes with #945, which touches `commands/data.py` and `docs/data.md` in the same region — both clean.

Credited in `CONTRIBUTORS.md` (`f491ac5b`). #930 closed with credit.

PR #855, opened 2026-09-11
docs: cover 23 missing commands in commands.md (#822)

MrBeldum:
## Summary
- Add the 23 registered leaf commands that were missing from `docs/commands.md` (including `soup eval against` / `soup llama quantize` via the existing `llama …|…` shorthand).
- Document `soup eval against` in `docs/evaluation.md` and `soup llama quantize` in `docs/serving-and-export.md`.
- Add a Typer-walk ratchet (`tests/test_commands_md_covers_registered_commands.py`) so the “full command list” claim cannot silently drift again.

Closes #822

## Test plan
- [x] Walked the live Typer app: 0 undocumented leaves after the docs change (23 before).
- [x] `pytest tests/test_commands_md_covers_registered_commands.py -o addopts=` — 4 passed
- [ ] Spot-check the new command lines in `docs/commands.md` for placement/format
- [ ] Confirm topic sections render for `eval against` and `llama quantize`

PR #972, opened 2026-09-14
docs(commands): split run-on command table entries onto individual lines (#970)

### LINKED ISSUE #970: docs/commands.md: six command entries are concatenated into two run-on lines
`docs/commands.md` has two lines where three command entries each are concatenated with no line break. Found while reviewing #855; **this predates that PR** and is on `main` today.

```
$ sed -n '159p;164p' docs/commands.md
soup registry push --run-id <id> --name n --tag v1  Register runsoup registry list [--name n] [--tag v1]     List registry entriessoup registry show <ref>  ...
soup history <name>                           Lineage DAG tree for a namesoup can pack --entry-id <id> --out r.can     Pack registry entry as .cansoup can inspect r.can ...
```

Note the joins: `Register run` + `soup registry list`, `List registry entries` + `soup registry show`, `for a name` + `soup can pack`, `as .can` + `soup can inspect`.

## What it costs

**Six commands are documented only inside a run-on line:**

| line | commands glued together | length |
|---|---|---|
| 159 | `soup registry push`, `soup registry list`, `soup registry show` | 213 chars |
| 164 | `soup history`, `soup can pack`, `soup can inspect` | 227 chars |

Two consequences:

1. **It renders badly.** In the fenced block a reader sees `Register runsoup registry list` and has to work out where one entry ends.
2. **It is invisible to any line-start parser**, including the coverage guard #855 is fixing. That guard counts only lines that *start with* `soup `, which is the correct rule — a prose mention should not count as documentation. Under it, `registry list`, `registry show`, `can pack` and `can inspect` are documented by a line whose first token belongs to a different command.

So these six are in the file but not reachable by the mechanism that is supposed to prove they are there. Fixing the guard without fixing this would leave four commands failing it for a reason that is not the author's doing — which is exactly why I am filing it separately rather than folding it into #855.

## Acceptance

- [ ] Each of the six entries is on its own line, with the column alignment the surrounding block uses
- [ ] No text is lost in the split — the descriptions currently glued to the following command name (`Register run`, `List registry entries`, `Lineage DAG tree for a name`, `Pack registry entry as .can`) belong to the entry *before* the join, not after it
- [ ] `grep -nE "[a-z)\"]soup [a-z]" docs/commands.md` returns nothing, i.e. no entry begins mid-line anywhere else in the file
- [ ] The #822 coverage guard passes with these six credited to their own rows

## Notes

Worth grepping the other `docs/*.md` command tables for the same shape while you are in there — this looks like a lost newline during an edit, and nothing would have caught it.

Do **not** widen the guard to accept mid-line entries as a workaround. A prose mention not counting as documentation is the property #822 exists to establish; the file should be fixed to match the rule, not the rule relaxed to match the file.

Voyagerroc-Lab:
Fixes #970

### Summary of Changes
- Split line 159 into three distinct command lines:
  - soup registry push --run-id <id> --name n --tag v1  Register run
  - soup registry list [--name n] [--tag v1]     List registry entries
  - soup registry show <ref>                      Entry details + artifacts + ancestors
- Split line 164 into three distinct command lines:
  - soup history <name>                           Lineage DAG tree for a name
  - soup can pack --entry-id <id> --out r.can     Pack registry entry as .can
  - soup can inspect r.can                        Preview manifest without extracting
- Preserved existing column alignment and original descriptions.
- Verified with regex that no mid-line command starts remain in docs/commands.md.


MakazhanAlpamys:
@Voyagerroc-Lab — welcome, and **your CI was blocked rather than slow.** A first-time contributor's workflow runs need manual approval; I approved `34832083116` and it is queued. My job to notice, not yours to chase.

Fifteen minutes from #970 being filed to a PR is fast. Review today.

**One thing worth saying now, because it is the whole risk in this change and the tests cannot see it.** The six entries are glued such that each description sits *before* the next command name, so a careless split attaches every description to the wrong command:

```
soup registry push ...  Register run   soup registry list ...  List registry entries   soup registry show ...
                        ^^^^^^^^^^^^                           ^^^^^^^^^^^^^^^^^^^^^
                        belongs to push                        belongs to list
```

`Register run` is `registry push`'s, `List registry entries` is `registry list`'s, `Lineage DAG tree for a name` is `history`'s, and `Pack registry entry as .can` is `can pack`'s. Nothing in CI would catch them landing one row off — the coverage guard only checks that a command is *mentioned*, not that its description is right. I will diff all six character by character.

**Two things about how this issue reached you**, since they affect what you can rely on here:

- I have assigned the `claimed` label to #970 so nobody duplicates your work. The **assignee** field would not take — GitHub only accepts an assignee who has push access or has commented on that issue, and `gh` exits 0 either way, which is how 11 claims went invisible here earlier today. The label is the registry; an empty assignee means nothing in this repo.
- **#855 is open and rewrites the very guard this issue is about.** It tightens the matcher so a prose mention no longer counts as documentation, which is what makes your six entries reachable in the first place. I will check whether the two collide and keep the order straight — not your problem to solve, but worth knowing your change and that one are related.

Two mechanical things that have failed CI for five PRs today, in case they apply: a changelog fragment must start with `- ` (a UTF-8 BOM counts as the first character and produces a confusing error that says the `- ` is missing when it is visibly there) and must reference its own PR number. A pure docs fix may reasonably have none — I will say which I think applies in the review.

MakazhanAlpamys:
Reviewed `3d9b4a7`-and-successor. **Merge-ready.** And first: **one of the two reasons I gave in #970 was wrong, and it was mine, not yours.**

## Correcting #970

I wrote that the run-on lines make six commands *"invisible to any line-start parser, including the coverage guard #855 is fixing."* I tested that against the fixed matcher and it is false:

```
With the FIXED matcher, against main's run-on docs:
  soup registry push     documented=True
  soup registry list     documented=True
  soup registry show     documented=True
  soup history           documented=True
  soup can pack          documented=True
  soup can inspect       documented=True
```

The rule is that the **line** must start with `soup `, not the needle — and line 159 does start with `soup registry push`, so everything glued after it is found inside a qualifying line with a valid token boundary. My rationale confused "the line starts with the command" with "the line starts with `soup `". The guard argument does not hold, in either the current #855 or the fixed version.

**What does hold is the rendering argument, and it is sufficient on its own.** Two lines of 213 and 227 characters where a reader sees `Register runsoup registry list` and has to work out where one entry ends is a broken reference page. That is worth fixing regardless of what any parser thinks.

So: your PR is right, the issue's acceptance boxes are met, and one of its justifications was not.

## The trap — all six pairs verified character by character

Each description stayed with the command **before** the join, which is the one thing that could have gone wrong silently:

| line | command | description |
|---|---|---|
| 159 | `soup registry push --run-id <id> --name n --tag v1` | `Register run` |
| 160 | `soup registry list [--name n] [--tag v1]` | `List registry entries` |
| 161 | `soup registry show <ref>` | `Entry details + artifacts + ancestors` |
| 166 | `soup history <name>` | `Lineage DAG tree for a name` |
| 167 | `soup can pack --entry-id <id> --out r.can` | `Pack registry entry as .can` |
| 168 | `soup can inspect r.can` | `Preview manifest without extracting` |

And **nothing was lost or invented** — not merely word-equal but byte-identical modulo the two inserted newlines:

```
L159 MATCH: True   words equal: True
L164 MATCH: True   words equal: True
rest of file identical before/between/after: True / True / True
```

LF endings, no trailing whitespace added, and the acceptance grep is clean:

```
grep -nE "[a-z)\"]soup [a-z]" docs/commands.md   ->  (nothing)
```

I widened it to `[A-Za-z0-9)">.<]soup [a-z]` as well — also nothing. Across the other docs pages there is one hit, `docs/adapters-and-governance.md:85`, and it is a legitimate quoted shell argument (`--eval-command "soup eval custom ..."`), not a run-on.

All six commands and their flags check out against `soup registry --help`, `soup can --help` and `soup history --help`. Scope is one file. 107 passed across the docs-related suites, merge-tree clean against `main` and against #855.

## Two small things

**A nit, your call:** the description columns land at 52 / 45 / 46 / 46 / 46 / 46 where the block is dominantly 46. Both deviations are inherited verbatim from the run-on line, and the file already mixes 45/46 elsewhere, so faithful preservation is defensible. Normalising line 160 by one space would be tidier.

**No changelog fragment.** I am waiving it — a rendering repair to a reference page with no behaviour change. Say if you disagree.

**One thing I need to pin before crediting you:** the commit is authored by `Voyagerroc-Code` while this PR is opened by `Voyagerroc-Lab` — two different accounts. Which should the `CONTRIBUTORS.md` entry name? I take the name from `gh api users/<login>` rather than from the commit email, so I would rather ask than guess.

Fifteen minutes from issue to PR, with the descriptions all attached to the right commands. Merging on green.

MakazhanAlpamys:
Merged as `b0e048e5`. Thank you, and welcome — this is your first contribution here.

Several command-table entries ran two commands onto one line, so a reader scanning the table saw one entry where there were two and **the second command was effectively undocumented**. Small diff, real consequence: an undocumented command is one nobody finds.

**One thing I want on the record, because it is mine and not yours:** one of the two reasons I gave when filing #970 was wrong. I corrected it in review rather than leaving it standing, and you should not take a maintainer's issue text as settled just because it is the maintainer's.

Credited in `CONTRIBUTORS.md` (`f491ac5b`); you are the 66th. #970 closed with credit.

PR #977, opened 2026-09-14
fix(config): reject offline MiniLLM at teacher mix 0

### LINKED ISSUE #692: Offline MiniLLM defaults to a zero teacher signal
## Summary

Enabling MiniLLM with Soup's default settings selects the offline path with `minillm_teacher_mix_ratio=0.0`. In that path, the distillation target becomes the detached student distribution, so the MiniLLM term contains no teacher signal even though the teacher is loaded and forwarded.

Training does not necessarily look stuck because the ordinary cross-entropy term remains. The failure is therefore silent: the run can progress without performing the advertised teacher distillation.

## Root cause

Current `main` (`ef2e02b51a1b34546823edce55715d28234805cb`) defines:

```python
target = ratio * p_t + (1.0 - ratio) * p_s.detach()
```

while the schema defaults are:

```text
minillm_teacher_mix_ratio = 0.0
minillm_on_policy = false
```

The live distill wrapper passes that ratio into `MiniLLMConfig`. At zero, the offline loss is effectively:

```text
KL(p_student || stopgrad(p_student))
```

The teacher forward still happens before this calculation; its logits simply receive zero weight.

## Numerical reproduction

Using the actual `minillm_distill_term()` from the pinned checkout with fixed student logits and two different teachers:

| ratio | teacher A loss | teacher B loss | student gradient norm |
|---:|---:|---:|---:|
| 0.0 (default) | 0.0 | 0.0 | ~3.18e-17 |
| 1.0 (control) | 0.85786 | 0.73924 | ~0.129 |

At ratio zero, changing the teacher also leaves the student gradient unchanged. The normal cross-entropy term in `DistillTrainerWrapper` remains active; only the teacher-dependent MiniLLM signal disappears.

## Important scope distinction

A zero mix ratio is meaningful in the true on-policy path: it means sampling the rollout from the student, while the loss still compares student and teacher distributions. The degenerate behavior described here is specific to the default **offline distribution-blend** path.

The MiniLLM paper uses the mix ratio for the rollout sampling distribution while retaining a teacher-based reverse-KL objective: https://arxiv.org/html/2306.08543#S2.SS2.SSS3

## Recommended fix

Separate the sampling-mixture parameter from the offline distillation target. Viable short-term choices include:

1. Reject `minillm_enabled=true`, `minillm_on_policy=false`, and `teacher_mix_ratio=0` with a clear validation error; or
2. Give the offline path a non-zero teacher-dependent default; or
3. Define the offline reverse-KL target directly from the teacher and reserve `teacher_mix_ratio` for on-policy sampling.

Do not globally forbid zero because it remains valid for student-only sampling in the on-policy mode.

## Acceptance criteria

- [ ] The default enabled offline MiniLLM configuration produces a non-zero teacher-dependent term for non-identical teacher/student logits.
- [ ] Replacing the teacher logits changes the offline loss or gradient under the default enabled configuration.
- [ ] The on-policy path continues to allow ratio zero and still computes `KL(student || teacher)`.
- [ ] Configuration help clearly distinguishes sampling mix from offline objective behavior.
- [ ] A regression test asserts that the teacher cannot be silently ignored when MiniLLM is enabled offline.

## Duplicate check

#237 implemented the live helper; #257 added the optional true on-policy path; #263 optimized that rollout with KV caching. None tracks the enabled-by-default offline combination producing a zero teacher signal. Searches across open/closed issues and PRs found no matching report or open fix.

jagadeepmamidi:
Enabling MiniLLM with the default offline settings loaded and forwarded the teacher, then blended it at weight zero so the MiniLLM term was `KL(student || stopgrad(student))`. Ordinary CE kept the loss curve healthy. This rejects that combination at parse and keeps mix 0 legal on the on-policy path, where it only means student-only sampling.

## What does this PR do?

`minillm_enabled: true` + `minillm_on_policy: false` + `minillm_teacher_mix_ratio: 0.0` is now a config error in `_validate_minillm_compat`. The message names the three fields. Mix 0 stays valid when `minillm_on_policy: true`. Field help, `docs/training.md`, and `docs/commands.md` now distinguish the offline blend weight from on-policy sampling.

Regression tests cover: offline default mix rejected, error names the fields, offline mix `0.3` accepted, on-policy mix `0` accepted, help text distinguishes the two meanings, and the kernel still treats a non-zero offline mix as teacher-dependent while on-policy mix `0` still computes `KL(student || teacher)`.

Fixes #692

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Tests

## Checklist

- [x] `pytest tests/ -v` passes (targeted MiniLLM / schema tests; not the full suite)


MakazhanAlpamys:
@jagadeepmamidi — reviewed at `2e8806d8`. **The fix itself is right and it is mutation-proof in both directions, which most PRs in this class are not.** There are three things to change, and the first one will turn CI red on all nine cells the moment your queued run reaches it, so I would rather you hear it from me now than from a red matrix in an hour.

## The core fix is correct

You implemented exactly the option that was ruled in on #692 — refuse at config-parse time when `minillm_enabled` is on, `minillm_on_policy` is off, and the mix ratio is 0 — and you kept the distinction that makes it a fix rather than a blanket ban: ratio 0 stays legal on the on-policy path, where it means student-only sampling while the loss is still KL(student ‖ teacher).

I mutated both halves, counting the anchor before each replacement and re-importing the module afterwards (not `ast.parse` — a real import), and both are genuinely guarded:

| mutation | anchor count | result |
|---|---|---|
| `schema.py:6529` `== 0.0` → `< 0.0` (unreachable, field is `ge=0.0`) | 1 | **2 failed** — `DID NOT RAISE ValueError` in `test_enabled_offline_default_mix_is_rejected` and `test_error_names_on_policy_and_enabled` |
| `schema.py:6528` `not tcfg.minillm_on_policy` → `True` | 1 | **1 failed** — `test_on_policy_zero_mix_is_accepted`, a real propagating `ValueError` |

The second one is the one I care about. Guarding against **over**-rejection is the half that usually goes untested, and you tested it at both schema and kernel level. Reverted clean each time.

No test was deleted or weakened. The single removed line is `self._yaml(task="distill", minillm_enabled=True)`, replaced by a stricter version that also asserts the ratio round-trips — it had to change, because it asserted acceptance of the exact config you now refuse. Net coverage goes up.

And blast radius is zero: no recipe, template or `examples/configs/` file sets `minillm_enabled`, so nothing shipped stops parsing.

## 1. BLOCKING — this turns CI red, and it is one word

You wrote the version prefix into `src/soup_cli/config/schema.py` twice:

```
6460:        v0.75.x #692 — ``minillm_enabled`` + offline blend + mix ratio 0
6526:        # v0.75.x #692 — offline blend at mix 0 is KL(student || stopgrad(student)).
```

`tests/test_issue627_unknown_config_keys.py::TestTheDeadline` asserts the rejection deadline version is written out in **exactly one** source file, and `v0\.75\b` matches `v0.75.x` because `.` is a non-word character. I ran it on your head:

```
E  AssertionError: the deadline version is written out in more than one place:
   ['config/schema.py', 'config/unknown_keys.py']
```

Introduced, not pre-existing — `git show 8fa29362:src/soup_cli/config/schema.py` grepped for that pattern returns **0** matches, your head returns **2**. The guard was already there at your merge-base, so this was catchable; your checklist honestly says you did not run the full suite, which is the right thing to have written down.

**Fix:** drop the `v0.75.x ` prefix from both lines, leaving `#692 — ...`. That is the house convention anyway — `schema.py` on `main` carries zero annotations of that shape; post-0.75 changes cite the issue or PR number only. Note `v0.75.1` would *not* help, `\b` still matches.

## 2. Required — `--minillm-on-policy` no longer does anything for the config it exists to serve

This is the one I would have missed by reading the diff, so I ran it. With a YAML that sets `minillm_enabled: true` and leaves the ratio at its default:

```
$ soup train --config soup.yaml --minillm-on-policy --yes

Config validation error:
  : Value error, minillm_enabled with minillm_on_policy=false requires
  minillm_teacher_mix_ratio > 0 ... or set minillm_on_policy=true ...
```

The user passed `--minillm-on-policy` and is told to set `minillm_on_policy=true`.

The cause is ordering, not your validator: `commands/train.py:577` loads and validates the config, and the flag is applied at `:665-672` by assigning `cfg.training.minillm_on_policy = True`. The schema declares no `model_config`, so `validate_assignment` is off and that assignment never re-validates. Your refusal therefore fires before the flag is ever seen.

The capability is not lost — `minillm_on_policy: true` in the YAML still works — but there is now **no way to express "on-policy with student-only sampling" via the flag**, and that is precisely the case #692 says must stay legal. Nothing caught it because the only CLI test asserts the flag appears in `--help`.

Either apply the flag before validation (or re-validate after the override), or, if refusing that combination is intended, say so in the message — the current text sends the user to the field their flag already set. Whichever you choose, it wants a test that runs the CLI rather than reading `--help`.

## 3. Rename the changelog fragment

`changelog.d/0.75.0/692.fixed.md` is named after the issue. `changelog.d/README.md:11` is explicit: the number "should be the pull-request number rather than the issue it closes". Rename to `977.fixed.md` and reference `#977` in the text, crediting as `(#692 by @jagadeepmamidi in #977)`. Not a build breaker — the validator only checks filename/text agreement — but the repo has tracked this exact drift before and it is cheap to get right.

The content itself is accurate and well written, and your docs edits to `docs/training.md` and `docs/commands.md` are the right ones for a user-visible config change.

## One thing that is not yours to fix

While checking your docs edit I found that `--minillm-enabled` and `--minillm-teacher-mix-ratio` **do not exist as CLI flags** — only `--minillm-on-policy` does. `docs/commands.md` and `CLAUDE.md` both document them anyway, so your edit makes a correct statement on a line that was already fiction. Repo-wide debt, pre-dating this PR; I will file it separately. Leave it alone here.

---

`ruff check src/ tests/` passes. 163 passed across your files plus `test_v0700_part_c.py` and `test_v07118.py`; 343 passed, 2 skipped across the six other minillm-touching files.

Ping me when 1 and 3 are pushed and you have decided what to do about 2, and I will re-review only the delta rather than starting over.

MakazhanAlpamys:
Following up on the last section: filed as #979. Four of the five documented `--minillm-*` flags do not exist — only `--minillm-on-policy` does — and `docs/training.md:289,298` presents two of them as runnable commands, so copy-pasting the docs gives `No such option` and exit 2.

Not yours, and explicitly out of scope for this PR. I mention it only because your item 2 and that issue share a root: `--minillm-on-policy` is applied after `load_config` has already validated, so anything added the same way inherits the problem. If you would rather solve item 2 by moving the flag application before validation, that also makes #979's option (2) viable later — but the minimal fix here is still fine.

jagadeepmamidi:
@MakazhanAlpamys — 1 and 3 are pushed, and I took the pre-validation path on 2.

1. Dropped both `v0.75.x` prefixes; the comments now cite `#692` only.
3. Renamed the fragment to `changelog.d/0.75.0/977.fixed.md` and credited `(#692 by @jagadeepmamidi in #977)`.
2. `--minillm-on-policy` is merged into the YAML via `load_config(training_overrides=)` *before* `SoupConfig` is constructed, so mix 0 stays legal on the on-policy CLI path. Left #979 alone.

Ready for the delta re-review.

MakazhanAlpamys:
Re-reviewed `3d44969f`. **All three items closed, and item 2 you solved better than I asked.** Merge-ready; the only thing left is CI reaching this head.

## Item 1 — the CI blocker

`git show 3d44969f:src/soup_cli/config/schema.py` grepped for the deadline pattern returns **0**. `test_issue627_unknown_config_keys.py::TestTheDeadline` is green.

## Item 2 — you fixed the cause, not the symptom

I described the problem (the flag is applied after `load_config` has already validated) and left the remedy open. You could have moved the assignment earlier in `train()` and been done. Instead you gave `load_config` a `training_overrides` parameter that merges into the raw `training:` mapping **before `SoupConfig` is constructed**, so a CLI flag now goes through the same cross-validators as the YAML rather than sidestepping them. That generalises: the next `--flag` that maps onto a training field inherits the fix instead of repeating the bug.

Verified by running it, both directions:

```
$ soup train --config soup.yaml --minillm-on-policy --yes     # yaml: minillm_enabled: true, mix at default
MiniLLM on-policy rollout enabled
┌─ Training Setup ─ ...                                        <- proceeds

$ soup train --config soup.yaml --yes                          # same yaml, no flag
Config validation error: minillm_enabled with minillm_on_policy=false requires ...
```

The refusal still fires when it should. That second run is the half that makes the first one safe, and it is the one a "make the flag work" patch typically breaks.

Leaving the now-redundant assignment in `train()` with a comment saying it is a no-op when the flag was set, rather than deleting it, is the right call — a later reader looking for where the flag lands on `cfg.training` finds it.

**Mutation, on the new path this time.** `config/loader.py`, anchor `training.update(training_overrides)` counted **exactly once**, replaced with `pass`; both `soup_cli.config.loader` and `soup_cli.commands.train` genuinely re-imported afterwards, not `ast.parse`:

```
FAILED TestMinillmOnPolicyCliFlag::test_load_config_override_allows_default_mix
FAILED TestMinillmOnPolicyCliFlag::test_cli_flag_allows_default_mix
2 failed, 9 passed
```

Reverted → 11 passed, tree clean. And `test_cli_flag_allows_default_mix` **invokes the CLI** rather than grepping `--help`, which is what I asked for; a `--help` assertion would have passed against the broken code.

## Item 3 — fragment

`changelog.d/0.75.0/977.fixed.md`, named after the PR. `test_issue487_changelog_fragments` green.

## Measured on the tree this actually merges to

`main` has moved six times since your first push, so I ran the merged tree rather than the branch: **150 passed** across `test_issue692`, `test_issue627`, the changelog fragment validator and #757's package-wide typer-defaults scanner — that last one because it walks every command module and you changed `commands/train.py`. 273 passed on the branch across your files plus `test_v0700_part_c` and `test_v07118`. `ruff check src/ tests/` clean.

---

Nothing needed from you. I will merge on the next green matrix and will not re-review — the verdict is recorded against `3d44969f`. The Actions backlog that was holding everything up today is cleared (104 live runs down to ~57), so this should be a short wait rather than the day-long one your first push got.

Separately: #979 is filed for the four `--minillm-*` flags that do not exist, as I said it would be. Your `training_overrides` parameter is incidentally what makes option (2) there cheap, if anyone takes it.

MakazhanAlpamys:
Merged as `3a37bf35`. Thank you — and you turned the review around in under an hour, which is why this is in the same batch rather than the next one.

**The second round is the one worth recording.** I described the problem — `--minillm-on-policy` applied *after* `load_config` had validated, so your new refusal fired before the flag was ever seen and told a user who had just passed the flag to set the field the flag sets — and left the remedy open. The cheap fix was to move the assignment earlier. You instead gave `load_config` a `training_overrides` parameter that merges into the raw `training:` mapping **before `SoupConfig` is constructed**, so a CLI flag now passes through the same cross-validators as YAML. The next flag that maps onto a training field inherits the fix rather than repeating the bug.

Verified by running it in both directions, which is the pair that matters: with the flag the run proceeds; without it the refusal still fires. A "make the flag work" patch typically breaks that second half.

Mutations: the schema guard made unreachable gives `DID NOT RAISE` on two tests; the on-policy exemption forced to `True` gives a real propagating `ValueError`, so **over-rejection is guarded too** — the half that usually goes untested; and `training.update(training_overrides)` replaced with `pass` fails two tests, one of which **invokes the CLI** rather than grepping `--help`. A `--help` assertion would have passed against the broken code.

Leaving the now-redundant assignment with a comment saying it is a no-op was also right: a later reader looking for where the flag lands on `cfg.training` finds it.

Credited in `CONTRIBUTORS.md` (`f491ac5b`). #692 closed with credit. Separately, #979 is filed for the four documented `--minillm-*` flags that do not exist — your `training_overrides` parameter is incidentally what would make implementing them cheap.

PR #978, opened 2026-09-14
Add qwen3.6-35b-a3b recipes for DPO and GRPO (#846)

Areen-09:
## What does this PR do?

Implements two ready-made task-variant recipes for `Qwen/Qwen3.6-35B-A3B` (MoE architecture with 3B active parameters), completing the task trio with the existing SFT sibling (`qwen3.6-35b-a3b-sft`):
* `qwen3.6-35b-a3b-dpo`: Direct Preference Optimization alignment recipe (`task: dpo`, `dpo_beta: 0.1`, `lr: 5e-6`, `batch_size: auto`, `gradient_accumulation_steps: 8`, `max_length: 4096`, `moe_lora: true`, `moe_aux_loss_coeff: 0.01`).
* `qwen3.6-35b-a3b-grpo`: Group Relative Policy Optimization reasoning recipe (`task: grpo`, `grpo_beta: 0.1`, `lr: 1e-5`, `batch_size: 1`, `gradient_accumulation_steps: 16`, `max_length: 8192`, `num_generations: 4`, `reward_fn: accuracy`, `moe_lora: true`, `moe_aux_loss_coeff: 0.01`, `gradient_checkpointing: true`).

Both recipes declare explicit `modality: text` (registered in `EXPECTED_QWEN35_TEXT_RECIPES`).

### Reviewer Verification Details

1. **Repo ID Verification:**
   Verified anonymously via `huggingface_hub.model_info("Qwen/Qwen3.6-35B-A3B", token=False)`:
   - Hub ID: `Qwen/Qwen3.6-35B-A3B`
   - Gated: `False`
   - Model type: `qwen3_5_moe`

2. **Two-Surface Literal Pinning:**
   Model ID is literally pinned across two independent surfaces in two separate tests in `tests/test_recipes.py` (`TestQwen36A3bDpoRecipe` and `TestQwen36A3bGrpoRecipe`):
   - Surface 1: `RecipeMeta.model == "Qwen/Qwen3.6-35B-A3B"`
   - Surface 2: YAML `base: Qwen/Qwen3.6-35B-A3B` (parsed directly without reading `.model`)

3. **Catalog Count Ratchet (171 -> 173):**
   Rebased cleanly over upstream #933 (171 recipes) and bumped count by 2 across all 9 locations:
   - 5 docs/code sites: `catalog.py:51`, `CONTRIBUTING.md:126`, `docs/commands.md:155`, `docs/serving-and-export.md:551`, `docs/serving-and-export.md:560`.
[...]

PR #981, opened 2026-09-14
fix: reject unwired BitNet training

### LINKED ISSUE #825: quantization: bitnet_1.58 is documented as live, but nothing routes to BitNetTrainerWrapper: an SFT run dies with "unknown quantization", and the wrapper itself always raises
## The defect

`docs/performance-and-quantization.md:855-858` says:

> `training.quantization: bitnet_1.58` routes to a live `BitNetTrainerWrapper` (an SFT subclass) for ternary-weight training. It is gated on the upstream `onebitllms` package — when absent, training fails fast with a friendly `RuntimeError` naming it

and `docs/commands.md:325` lists it as `soup train  # quantization='bitnet_1.58' (sft/pretrain/dpo) ... LIVE-gated (v0.71.20)`. Three separate things are wrong with that.

**1. No code path reaches the wrapper.** The only constructor of `BitNetTrainerWrapper` is the factory `utils/bitnet.py:169-180` (`build_bitnet_trainer`), and `build_bitnet_trainer` has no caller anywhere in `src/`:

```

$ grep -rn "build_bitnet_trainer\|BitNetTrainerWrapper" src/soup_cli --include=*.py

src/soup_cli/trainer/bitnet.py:3:``BitNetTrainerWrapper`` lifts the v0.52.0 ``build_bitnet_trainer`` schema

src/soup_cli/utils/bitnet.py:170:    Returns a :class:`~soup_cli.trainer.bitnet.BitNetTrainerWrapper`. BitNet

src/soup_cli/utils/bitnet.py:178:    from soup_cli.trainer.bitnet import BitNetTrainerWrapper

src/soup_cli/utils/bitnet.py:180:    return BitNetTrainerWrapper(config, **kwargs)

```

`commands/train.py` dispatches on `task` only (lines ~1383-1474) and never looks at `quantization`, so a `task: sft` config with `quantization: bitnet_1.58` goes to the plain `SFTTrainerWrapper`. `git log -S build_bitnet_trainer` shows the factory was added in `a4dfbb30` (v0.71.20) and never wired.

**2. The plain SFT path then crashes with a misleading error.** `SFTTrainerWrapper` asks the quant menu for a loader config, and the quant menu has no `bitnet_1.58` branch (`utils/quant_menu.py:370-488`):

```python

from soup_cli.utils.quant_menu import build_quantization_config_for_loader

cfg = load_config_from_string("""base: HuggingFaceTB/SmolLM2-135M

task: sft

data:

  train: ./d.jsonl

training:

  quantization: bitnet_1.58

""")

print(cfg.training.quantization)          # bitnet_1.58  -- the config validates clean

build_quantization_config_for_loader(tcfg=cfg.training, base=cfg.base)

# ValueError: unknown quantization 'bitnet_1.58'

```

So the user gets neither ternary training nor the documented "friendly RuntimeError naming onebitllms" — they get "unknown quantization" for a value the schema just accepted and the docs just advertised.

**3. Even if it were wired, the wrapper cannot train.** `trainer/bitnet.py` `BitNetTrainerWrapper.setup()` calls `_require_onebitllms()` and then **unconditionally** raises `RuntimeError("BitNet 1.58-bit fine-tuning is hardware/dependency-gated. The 'onebitllms' package is importable but the BitLinear-aware training path has not been validated on this platform ...")`. There is no branch that trains. The module docstring says as much ("The real end-to-end training run ... stays an open infra-blocked item"); the docs do not.

The schema gate `SoupConfig._validate_bitnet_compat` (`config/schema.py:4731-4746`) accepts `task` in sft/pretrain/dpo, and the field description (`schema.py:1129`) still reads "bitnet_1.58 (BitNet ternary, v0.52.0 schema-only)" — the one surface that is accurate.

## Why it matters

A user reading the docs sets one field, passes validation, and fails at model load with an error that suggests a typo. The only accurate statement of the feature's status is buried in a module docstring. `soup export --format bitnet|tq1_0` is a separate, working path and is not affected.

## Proposed fix

Pick one, explicitly:

- **Honest gate now (small):** reject `quantization: bitnet_1.58` at config load with a message that states the status ("BitNet 1.58 training is not implemented yet; the export path `soup export --format tq1_0` works on an existing BitNet checkpoint"), fix the two doc passages, and keep #138 for the RL expansion. This is the `packing_cross_doc_attn_mask` precedent (rejected at load rather than silently accepted).

- **Or wire it:** route `quantization == "bitnet_1.58"` to `build_bitnet_trainer` in `commands/train.py` before the task dispatch, implement the `onebitllms` path in `setup()`, and keep the friendly error when the package is absent. This needs a Linux + CUDA box with `onebitllms` to validate; until that run exists, the docs must not say "live".

## Acceptance criteria

- [ ] A test that loads a `quantization: bitnet_1.58` SFT config and drives `soup train` setup asserts the documented behaviour (either a load-time rejection naming the status, or `BitNetTrainerWrapper` being constructed). It FAILS on current code (plain SFT, `ValueError: unknown quantization`).

- [ ] `docs/performance-and-quantization.md` and `docs/commands.md` describe what actually happens.

- [ ] If wired: `build_bitnet_trainer` has a caller, and a test with `onebitllms` stubbed absent asserts the friendly `RuntimeError` naming the package.

- [ ] The #748 consumer guard (or its successor) would have flagged a factory with no caller; see the separate issue on reachability.

## Scope

Not the TQ1_0 export (works independently) and not the RL-task expansion (#138).

---

- #807 — The #748 consumer guard counts a read inside a function nothing calls: training.warmup_auto passes it and d...

Rayan-and-beyond:
## Summary

- reject `training.quantization: bitnet_1.58` at config load until the training path is actually implemented
- keep the working BitNet/TQ1_0 export path unchanged and point the validation error to it
- retire the unusable `falcon-e-bitnet-sft` recipe and sync recipe-count/snapshot bookkeeping
- update the BitNet docs and regression tests to match the real behavior

## Why

Soup currently accepts `bitnet_1.58` training configs and documents them as live, but no train dispatch reaches `BitNetTrainerWrapper`; the ordinary path later dies with `unknown quantization`. The wrapper itself also cannot complete training yet. Failing at config load gives users the accurate status immediately while preserving the independent export path.

## Validation

- focused BitNet/recipe/snapshot/count regression slice: `581 passed, 8 skipped`
- catalog-count tests touched by the recipe removal: `3 passed`
- `git diff --check`

Closes #825


MakazhanAlpamys:
@Rayan-and-beyond — reviewed `b822e526`. **Welcome, and your premise checks out: BitNet training really is unwired.** I verified that by execution rather than by reading your description, because it was the one thing that could have made the whole PR wrong.

## The premise, proved

`build_bitnet_trainer` has **zero call sites in `src/`**, and `commands/train.py` does not contain the string `bitnet` at all. The dispatch is an `elif cfg.task == ...` chain that never reads `cfg.training.quantization`, so `task: sft` + `quantization: bitnet_1.58` falls through to `SFTTrainerWrapper` and dies at `quant_menu.py:488`:

```
config PARSED OK; quantization = bitnet_1.58 task = sft
quant loader raised: ValueError: unknown quantization 'bitnet_1.58'
```

The ordering makes it worse than it reads: the tokenizer is downloaded at `sft.py:1324` **before** that line, so a user pays a download and then gets an error that looks like a typo — for a value the schema had just accepted.

**And it is not a dependency gate**, which is the distinction that decides this PR. I stubbed `onebitllms` as present and drove the factory directly:

```
simulated onebitllms find_spec: True
setup raised WITH onebitllms present:
  RuntimeError: BitNet 1.58-bit fine-tuning is hardware/dependency-gated ...
  the BitLinear-aware training path has not been validated on this platform.
```

`trainer/bitnet.py:36-42` raises **unconditionally after** the import check passes. So installing the package unlocks nothing, and refusing at config load removes no working capability. The `soup export --format tq1_0` path is independent and you correctly left it alone.

## What is verified and not in question

Blast radius exactly one shipped artifact, which you retire: 171 -> 170, every one of the five count sites updated, and `python scripts/generate_recipe_snapshot.py` leaves `git diff` on the fixture **empty** — a clean regeneration rather than a hand edit. Mutation in both directions: replacing your `raise` with `return self` fails 6 behaviourally; adding a benign recipe while leaving the counts alone fails the ratchet by name, so the count is genuinely derived. 2769 passed on the merged tree. Fragment named after the PR and credited to you. **No assertion was weakened** — every deletion is either an inverted behaviour with a stronger replacement, or the retired recipe.

Keeping `bitnet_1.58` in the `Literal` so the user gets your message instead of a generic enum error was the right call.

## 1. The new refusal is placed after the old one, so two stale messages still win

Your `raise` sits *after* `validate_bitnet_compat(...)`, so for non-default combos the pre-existing messages fire first — and they are now actively wrong:

```
grpo + bitnet -> "only wired for task in (sft, pretrain, dpo)"      <- none of those is wired
mlx  + bitnet -> "Use backend='transformers'"                      <- that does not work either
```

The second one tells the user to do something that cannot succeed, which is the same defect class #825 exists to remove — your fix closes the front door and leaves two side doors open. Raise the not-implemented error **before** the compat call (or drop that call, since every path now refuses), and adjust the two pinning tests in `test_v0520.py`.

## 2. Minor

`test_bitnet_training_refused_until_wired` uses `pytest.raises(Exception, match=...)`; narrow it to `ValueError` to match your own new `test_issue825_*.py`. And you dropped `bitnet_1.58` from the streaming-refusal parametrize in `test_v07202.py` rather than re-pinning which validator now wins — understandable, but it removes a case.

---

For a first contribution this is unusually careful: you announced the approach on the issue before writing it, picked the option the issue asked you to pick explicitly, regenerated the snapshot instead of hand-editing it, and decremented the `>= 112` floor by exactly one rather than deleting it. Item 1 is a five-minute reorder. Push it and I will review only the delta.


---

*Edited to repair two code spans that my own tooling ate before posting: the sentence above should have read "`python scripts/generate_recipe_snapshot.py` leaves `git diff` on the fixture **empty**". Nothing about the findings changed — saying so rather than silently fixing it, since an edited review with no note is indistinguishable from one that moved the goalposts.*

Rayan-and-beyond:
Addressed the review delta: the not-implemented refusal now runs before the old compatibility checks, the schema tests pin `ValueError` and the global refusal for GRPO/MLX, and the streaming BitNet case is restored with the new validator expectation. Focused validation: 7 passed in the BitNet schema slice, 1 passed for the streaming refusal, Ruff and `git diff --check` pass.

MakazhanAlpamys:
Reviewed the delta at `3ea85b1b`. **Both review items are closed, and item 1 you closed with the stronger option than the one I asked for.** One blocking item, and it is not something you did — it is something a merge undid.

## Closed, each by execution

**Item 1 — you dropped `validate_bitnet_compat` entirely rather than reordering it.** `schema.py:4842-4851` is now `if quantization != "bitnet_1.58": return self` followed by an unconditional raise. Executed: `grpo` and `mlx` both now say *"BitNet 1.58 training is not implemented yet…"* instead of the two stale side-door messages. Reordering would have left those messages reachable; deleting the call removes the side door. Better than the ask.

**Item 2a** — `tests/test_v0520.py:596,604,612` now pin `ValueError` with `match="training is not implemented yet"`.

**Item 2b — the streaming case is restored and not weakened.** You removed `bitnet_1.58` from the `["8bit","gptq","bitnet_1.58"]` parametrise and gave it a dedicated test at `test_v07202.py:1461`, narrowing `Exception` to `ValueError`. Dropping the old assertion that the refusal names `4bit` is **correct**, because that assertion is now false: `_validate_bitnet_compat` is declared at `:4843` and `_validate_stream_layers_compat` at `:5243`, so the global gate fires first. I proved it rather than assuming — with the global gate mutated away, that same config still refuses, with the streaming message. Your test name, `test_bitnet_streaming_is_refused_by_the_global_training_gate`, says exactly that, which is the right way to record a branch that is correct but now unreachable.

## Mutations — both fire, module imports confirmed in both

**Kill the refusal.** Anchor = the five-line `raise ValueError(...)` block, **one occurrence** in `src/`; replaced with `return self`; `import soup_cli.config.schema` confirmed OK. **9 failed, 382 passed** — behavioural (`Failed: DID NOT RAISE ValueError`), and the killers span all four touched suites rather than clustering in one.

**Restore the ordering you say you fixed.** I put `validate_bitnet_compat(...)` back *before* the raise. **2 failed** — and not vacuously:

```
Expected regex: 'training is not implemented yet'
Actual message: "quantization='bitnet_1.58' is only wired for task in (sft, pretrain, dpo); got task='grpo'"
Actual message: "quantization='bitnet_1.58' is not supported on backend=mlx (onebitllms is CUDA-only)."
```

Exactly the two tests I asked to be re-pinned. So the ordering is behaviourally pinned, not just asserted in the PR body.

Reverted, `git diff --stat` empty, 391 passed / 2 skipped.

**Deleted test lines: every one justified.** The `*_happy` and `test_returns_wrapper` deletions are inverted behaviours each replaced by a stronger refusal assertion; the recipe deletions follow the retired artifact; `>= 112` became `>= 111` rather than being deleted. Zero weakened assertions.

**Scope check — nothing that should work is caught.** All 171 shipped recipes load through the real `load_config_from_string`, 0 failing, `falcon-e-bitnet-sft` correctly absent. All 21 templates load. The export path is untouched: `BITNET_EXPORT_FORMATS == ['bitnet','tq1_0']`, `soup export --help` still lists them, and `commands/export.py` never loads a `SoupConfig`. End to end, `soup train` on a BitNet config exits 1 with the message pointing at the working export path.

---

## 1. Blocking, and it is a merge regression rather than your edit

`docs/commands.md:330` is back to the false claim. At `b822e526` your line read:

```
soup train  # quantization='bitnet_1.58'  BitNet 1.58 training is not implemented; config load refuses it
```

The merge commit `f86bb7e1` took `main`'s side and it now reads:

```
soup train  # quantization='bitnet_1.58' (sft/pretrain/dpo)  BitNet 1.58 SFT (requires onebitllms) — LIVE-gated (v0.71.20)
```

`3ea85b1b` restored the catalog and the counts but not this line. It is now **the only public doc still advertising BitNet training as live** — `docs/performance-and-quantization.md:884-897` survived your edit correctly. It also falsifies your own changelog fragment, whose last clause says the docs advertising training as live are removed.

One-line revert.

## 2. High: the branch is 11 commits behind and the recipe count moved again

Merge base `614b7f64`; `main` is now `34d38ac4` and the catalog is at **174** (#978 added two this afternoon). `git merge-tree --write-tree origin/main 3ea85b1b` reports **8 content conflicts**: `CONTRIBUTING.md`, `docs/commands.md`, `docs/serving-and-export.md`, `src/soup_cli/recipes/catalog.py`, `tests/test_recipes.py`, `tests/test_v07124.py`, `tests/test_v07130.py`, `tests/test_v07132.py`.

**After the rebase the correct pin is 173, not 171** — 174 on main minus the one recipe you retire. Your 15/15 green measures a tree that cannot merge as it stands, which is the thing branch-green cannot tell you.

## 3. Low, and not yours to fix here

`utils/bitnet.py:99`'s `validate_bitnet_compat` now has **zero callers in `src/`** — only a comment in `advanced_precision.py:129` names it. It is still exercised directly by `test_v0520.py:529-547`, so it is not dead code in the test sense. That makes it the second orphan in that module alongside `build_bitnet_trainer` (`:167`, also zero `src/` callers). I am not asking you to delete either; I will file it as a tracked cleanup rather than let it sit silently, matching what was done for #955.

---

Rebase, re-pin the count to 173, and revert that one doc line, and I will merge on the next green matrix reviewing only the delta. Thank you for taking the sharper option on item 1 — it removed a class rather than reordering it.

MakazhanAlpamys:
Checking in — head unchanged since my review, so this is a nudge with **one genuinely useful new fact**, not a re-review.

**`docs/commands.md` is inside your conflict set.** Your branch is now 12 commits behind `main` (`f8226214`) with **8 content conflicts**: `CONTRIBUTING.md`, `docs/commands.md`, `docs/serving-and-export.md`, `src/soup_cli/recipes/catalog.py`, `tests/test_recipes.py`, `tests/test_v07124.py`, `tests/test_v07130.py`, `tests/test_v07132.py`.

That is convenient rather than annoying: the rebase **forces you to open the one file carrying the blocking finding**. `docs/commands.md:330` is still byte-identical to `main`'s and still says

```
soup train  # quantization='bitnet_1.58' (sft/pretrain/dpo)  BitNet 1.58 SFT (requires onebitllms) — LIVE-gated (v0.71.20)
```

I re-grepped `docs/` and `README.md`: that is **the only** public doc still advertising BitNet training as live. `docs/performance-and-quantization.md:886-887` correctly says it is not implemented. It also falsifies your own changelog fragment, whose last clause claims the live-advertising docs are removed.

**The count, derived today rather than quoted:**

```
main  f8226214 : len(RECIPES) = 174,  falcon-e-bitnet-sft present
your  3ea85b1b : len(RECIPES) = 171,  retired
```

so the correct pin after rebase is **173** — 174 minus your one retired recipe. Unchanged from my last review even though `main` moved 12 commits, because no recipe landed in that window. Sites to move 171 -> 173: `test_recipes.py:1392` (the method *name*) and `:1429`, the arithmetic comment at `:1423-1425`, `test_v07124.py:277` (a docstring reference) and `:281`, `test_v07130.py:740,743`, `test_v07132.py:953,956`.

Two of those are **names**, and this repo has a written post-mortem about exactly that: `test_v07124.py:270-278` records a count that drifted 116 -> 134 -> 137 -> 158 -> 162 while the method name stayed at 158, so "the one thing a reader saw first was the one thing that was false". Nothing fails when a name goes stale, which is why it keeps happening.

**Your 14/14 green measures a tree that cannot merge** — that is precisely what a branch-level matrix cannot tell you, and why I am not merging on it.

**Nothing in your actual fix has degraded.** I re-verified: the mutation still bites hard — replacing the five-line `raise ValueError(...)` block at `schema.py:4846-4851` (anchor matched once, module imports afterwards) fails **9** tests across four suites, all `DID NOT RAISE ValueError`. 787 passed on the touched files. And every deleted test line is still justified — the three `Exception` -> `ValueError` narrowings are strengthenings, and the `>= 112` floor moved to `>= 111` by exactly the one retired recipe.

The orphan I mentioned stays **#1004** and is not yours to fix here.

Rebase, fix the one doc line while you are in it, re-pin to 173, and I will merge.

MakazhanAlpamys:
Merged as fce62d85. Thank you, @Rayan-and-beyond.

You deleted the compat call rather than reordering it. Reordering would have left `task: grpo` and `backend: mlx` able to slip past; deleting it closes both side doors. You also kept every recipe-count site in sync through two moves of `main`, regenerated the snapshot, and fixed the `docs/commands.md` line on the first rebase.

What I ran on the tree merged with `main`:
- **`soup train --config` before and after:** before, the config parsed and the run died with `ValueError: unknown quantization 'bitnet_1.58'`. After, it is refused at load with a message that points to `soup export --format tq1_0`, for sft and grpo alike.
- **Everything else still loads:** 0 of the 175 recipes fail, all 21 templates load, and `soup export --help` still lists `bitnet` / `tq1_0`.
- **Tests:** 2917 passed across the touched and count-related files.
- **Mutation:** replacing the refusal with `return self` fails 9 tests, all `DID NOT RAISE`.

Heads-up for anyone rebasing: #1034 conflicts with this in `tests/test_v07120.py`, and I have described the resolution there.

Issue #815, opened 2026-09-11
soup loop canary writes three fields to loop.yaml and nothing reads them: no traffic split, no canary verdict, no rollback — and the pre-wired deploy records verdict "OK" on a 200


MakazhanAlpamys:
## The defect

`docs/commands.md:148` advertises `soup loop canary <adapter> --traffic 5%` as "Promote canary + auto-rollback on MAJOR", and `docs/adapters-and-governance.md:123-124,133` describes a deterministic router that is "sticky-on-rollback — a flaky verdict can't ping-pong traffic between adapters". The router, the verdict and the rollback all exist as functions. None of them is called.

## Evidence

**The functions, and their callers in `src/`:**

| function | defined | callers outside its own module |
|---|---|---|
| `canary_router.route` (the traffic split) | `utils/canary_router.py:83` | none |
| `loop_daemon.evaluate_canary_verdict` | `utils/loop_daemon.py:380` | none |
| `loop_daemon.maybe_rollback` | `utils/loop_daemon.py:391` | none |
| canary `BucketStats(...)` (the per-bucket stats the verdict needs) | `utils/canary_router.py` | never constructed (the only `BucketStats(` in `src/` is an unrelated class in `utils/curriculum_dynamic.py:163`) |

They are exercised only by `tests/test_v0580.py`.

**What the command does.** `commands/loop.py:276-316` builds a `CanaryPolicy` for validation, then writes `canary_active`, `canary_traffic_pct` and `canary_autoroll_on_regress` into `.soup/loop.yaml`. The only reader of those fields is `soup loop status`, which prints them (`commands/loop.py:140-144`). Note also that `--autoroll-on-regress` is passed into the policy as `sticky_on_rollback` (line 298), a different property.

**What the pre-wired deploy stage does.** `utils/loop_stages.py:370-376`:

```python
ok = _post_activate(endpoint, name)
return {"deployed": ok, "canary_verdict": "OK" if ok else None, ...}
```

The verdict is "OK" whenever the activate POST returns 2xx. And the endpoint it calls, `POST /v1/adapters/activate/{name}` (`commands/serve.py:1797-1811`), sets a single process-global `active_state["active"] = name` — it switches **all** traffic to the new adapter. So the pre-wired loop's "canary" is a 100% cutover whose verdict is decided before any request has been served on it.

[...]

PR #982, opened 2026-09-14
fix(loop): stop deploy_to_canary reporting a fabricated OK verdict

AmirF194:
## What does this PR do?

`deploy_to_canary` activates the trained adapter at 100% traffic (a full hot-swap via `POST /v1/adapters/activate/<name>`, see `serve.py`), then reports `canary_verdict: "OK"` from that POST's bare 2xx. `canary_router.route()`/`BucketStats`, the actual per-request traffic split and verdict machinery, are never called anywhere outside tests, so nothing evaluated a canary before that "OK" got written into the iteration record.

Changed the success path to report `canary_verdict: None`, matching every other early return in the function, since no canary run actually happened.

Didn't wire `route()`/`evaluate_canary_verdict()`/`maybe_rollback()` into the live serving path here. That's a real design call (the only server primitive today is a full-adapter hot-swap, not a traffic split), worth its own PR if you want it. This one just stops the daemon from recording a verdict nothing computed.

Refs #815

## Type of change

- [x] Bug fix

## Checklist

- [x] `pytest tests/test_v0714.py tests/test_v0580.py -v` passes, 326 tests, on 3.10/3.11/3.12. Didn't run the full `[dev]` suite (torch stack), unrelated to this path.
- [ ] Docs: none describe this stage's verdict semantics
- [x] Changelog fragment: `changelog.d/0.75.0/982.fixed.md`


MakazhanAlpamys:
@AmirF194 — reviewed `e71ad893`. **The fix is correct and the mutation kills.** Two one-liners before merge.

**What the change is worth.** `deploy_to_canary` recorded `canary_verdict: "OK"` from a bare 2xx on `POST /v1/adapters/activate/{name}` — which is a **100% cutover**, not a canary. I confirmed `canary_router.route` and `BucketStats` have no `src/` callers at all, so no traffic split and no verdict machinery ever ran. A fabricated OK in a field named `canary_verdict` is worse than an absent one, because the operator reads it as evidence that something was evaluated. `None` is already an accepted manifest value at `loop_iteration.py:71` and `loop_daemon.py:207`, so nothing downstream breaks.

**Both removed test lines are justified**, and I checked rather than assumed: one is replaced in place by `assert out["canary_verdict"] is None`, the other is deleted from its original test but re-pinned by your new `test_successful_activation_reports_no_canary_verdict` under a `_DEPLOY_POSTER` stub. The mutation proves both sites still bite.

**Mutation.** The bare string `"canary_verdict": None,` occurs **four times**, so I scoped the anchor to the two-line block `"deployed": ok,` + `"canary_verdict": None,` — exactly one match — and reverted it to `"OK" if ok else None`. Module genuinely re-imported. **2 failed, 134 passed**, both behavioural (`assert 'OK' is None`). Reverted, 136 passed.

441 passed, 2 skipped on the merged tree including the four repo-wide ratchets; ruff clean; no collisions with any open PR.

## 1. `Fixes #815` should be `Refs #815`

#815 has **four** acceptance criteria: a real 5% traffic-split test, a MAJOR-stats rollback test, `deploy_to_canary` ceasing to report OK from a bare POST, and a docs correction. This PR satisfies the third. Your own body says so plainly — *"Didn't wire route()/evaluate_canary_verdict()/maybe_rollback()... worth its own PR"* — which is the right call and honestly stated, but `Fixes` will auto-close an issue with three criteria still open.

You used `Refs` correctly on #983 the same day, so this is a slip rather than a habit.

## 2. The fragment has no author credit

```
(#815 in #982)
```

House style is `(#issue by @author in #PR)`. Please make it `(#815 by @AmirF194 in #982)`. The CHANGELOG is the file people actually read, and an entry with no name renders your work as nobody's.

## Optional, worth it

The docstring still opens *"Promote the trained adapter **as a canary**"* — the exact framing this PR is correcting. One line there would stop the next reader re-deriving the same confusion. And `notes: "activated at 100% traffic; no canary evaluated"` on the success path would let the manifest record *why* the verdict is absent rather than leaving a bare `None`. #815 permits `"UNKNOWN"` instead; `None` is fine as you have it, since `deployed` disambiguates.

Not asked for, just noted: `docs/adapters-and-governance.md:134` still asserts the router is "deterministic... and sticky-on-rollback" as though it were live. That is #815's fourth criterion and it stays open either way.

Both one-liners, then I merge.

AmirF194:
Pushed e88558a and edited the PR body:

1. Fixes #815 -> Refs #815. You're right, this only closes the third of the four acceptance criteria there.
2. Changelog now credits @AmirF194.

MakazhanAlpamys:
Re-reviewed `e88558a0`. **Both closed.** Merge-ready; waiting only on the CI now running.

`Refs #815` in the body, and the fragment credits you — `(#815 by @AmirF194 in #982)`.

Worth restating why the `Fixes` -> `Refs` change matters beyond bookkeeping: #815's other three criteria are a real 5% traffic-split test, a MAJOR-stats rollback test, and the `docs/adapters-and-governance.md` correction that still describes the router as "deterministic... and sticky-on-rollback" as though it were live. Auto-closing would have buried all three behind a green checkmark, and the next person to look would have found an issue marked done and a router that still does not route.

My verdict from the last round stands: the mutation kills (2 failed, both `assert 'OK' is None`), both removed test lines are justified with the value re-pinned in a sibling, and `None` is already an accepted manifest value downstream. Nothing needed from you.

MakazhanAlpamys:
Merged as `03066757`. Thank you.

**What I ran before merging.** Head `e88558a0` was unchanged from the verdict I recorded at 07:57Z, so this was not re-reviewed — only re-checked. `main` had moved (#989 at 11:57Z), so I composed your head onto `614b7f64` with #938, #983 and #984 and ran the batch as one tree: **612 passed, 2 skipped, zero failures**, including `test_v0714.py`.

**The part worth naming is the restraint.** The tempting version of this PR wires `route()` / `evaluate_canary_verdict()` / `maybe_rollback()` into the serving path and makes the "OK" true. You didn't, and you said why: the only server primitive today is a full-adapter hot-swap, not a traffic split, so making that verdict real is a design change and not a bug fix. Reporting `canary_verdict: None` — matching every other exit in the function — is the change that is provably correct without deciding anything about the design.

That matters more here than in most places, because this field is a **governance record**. `soup loop replay` reads it back. A field that exists, is populated, and is derived from a bare 2xx on an activation POST reads as evidence to whoever audits the iteration later, and nothing in the record says the machinery never ran.

**And the `Fixes` -> `Refs` correction was yours, not mine forcing it.** I raised it, you agreed within the hour and edited the body. For anyone reading this thread later: #815's other three acceptance criteria are a real 5% traffic-split test, a MAJOR-stats rollback test, and the `docs/adapters-and-governance.md` correction that still describes the router as "deterministic... and sticky-on-rollback" as though it were live. Auto-closing would have buried all three behind a green checkmark, and the next person to look would have found a closed issue and moved on.

#815 stays open with three criteria left. If you want the traffic-split half, say so on the issue and it is yours.

AmirF194:
Thanks for rerunning the full batch before merging rather than trusting the diff, and for catching the Fixes/Refs distinction. Glad #815's other three criteria stay visible instead of auto-closing under a green check.

Issue #806, opened 2026-09-11
use_flash_attn and use_liger are read only by SFT's text path, yet validated on every task and switched on by soup autopilot for DPO, GRPO and pretrain


MakazhanAlpamys:
## The defect

`training.use_flash_attn` and `training.use_liger` have exactly one training consumer, `SFTTrainerWrapper._setup_transformers`:

- `trainer/sft.py:1293` applies the Liger patch before model load, and `sft.py:882` sets `use_liger_kernel` on `TrainingArguments`;

- `trainer/sft.py:1351-1356` sets `attn_implementation` from `use_flash_attn`.

No other trainer, and not SFT's vision, audio or streaming branches, references either field or sets `attn_implementation` (grep over `trainer/` outside `sft.py` and `mlx_sft.py` finds nothing). Yet:

1. **`commands/train.py:1192-1214` validates both for every task** and exits 1 when the package is missing. A DPO run can be refused for a missing `liger-kernel` that DPO would never use.

2. **`soup autopilot` enables both** on Ampere or newer (`autopilot/decisions.py:344-346`) whatever the task, and its goal map sends `alignment` to DPO, `reasoning` to GRPO and `domain-adapt` to pretrain (`decisions.py:15-17`).

## Reproduction

`build_soup_config` run on this machine (compute capability 12.0) with a preference JSONL:

```

task alignment -> dpo use_flash_attn= True use_liger= True

task reasoning -> grpo use_flash_attn= True use_liger= True

task domain-adapt -> pretrain use_flash_attn= True use_liger= True

```

On the same machine neither package is installed, so the validators `soup train` runs for any task return:

```

liger installed: False | flash-attn: None

['liger-kernel is not installed. Install it with: pip install "soup-cli[liger]"']

['FlashAttention is not available. Install it with: pip install flash-attn --no-build-isolation']

```

So the zero-config path writes a DPO config that `soup train` refuses, for two settings DPO ignores. Installing both packages makes the run start, still without either kernel. Hand-written configs are accepted on every task too:

```

ACCEPT dpo  ACCEPT grpo  ACCEPT pretrain  ACCEPT kto  ACCEPT orpo  ACCEPT embedding  ACCEPT distill

  {'use_flash_attn': True, 'use_liger': True, 'rope_scaling_type': 'dynamic', ...}

```

`docs/performance-and-quantization.md:806-808` presents both with no task scope.

## Same shape: rope_scaling_type

[...]

PR #983, opened 2026-09-14
fix(autopilot): scope use_flash_attn/use_liger to trainers that read them

AmirF194:
use_flash_attn / use_liger now get rejected by SoupConfig at config load for any task outside `{sft, tts}`, the same pattern `_validate_reasoning_effort_task_gate` already uses for reasoning_effort. Autopilot's `build_soup_config` is scoped the same way, so it stops writing a config it would immediately reject on Ampere+ hardware.

One thing worth flagging: sft.py's `_setup_transformers` is also what `tts.py`'s `TTSTrainerWrapper` inherits, so tts reads both flags too. The allowlist is `{sft, tts}`, not just sft.

I went with rejecting at load rather than wiring both flags into the other trainers, since neither one silently no-ops today (the DPO/GRPO/pretrain paths never build an `attn_implementation` kwarg or a Liger patch call at all), and #745 already set that precedent for loraplus_lr_ratio.

Left out on purpose: the "same shape, narrower" reasoning_effort/train_on_eot note in the issue. That allowlist is a different function with its own history and its own consumer (`build_format_row`, sft.py + distill.py), and I haven't checked whether tts's TTS-specific formatting path calls it the same way sft.py's text path does. Filing this as `Refs #806` rather than closing it, since that acceptance criterion is still open.

Added `tests/test_issue806_kernel_flags_task_scoped.py`: the ones that build a DPO/GRPO/pretrain config on Ampere+ hardware, or construct a SoupConfig directly with either flag set outside sft/tts, reproduce the bug on main and go green on this branch. `test_config.py`, `test_autopilot.py`, `test_issue627_unknown_config_keys.py`, `test_issue748_config_fields_reach_a_consumer.py`, `test_issue799_autopilot_unwired_gates.py` and `test_examples_configs.py` are still all green. ruff and pre-commit are clean too.

Didn't install the `[dev]` training stack (torch/transformers/trl) here, so I only ran the config/autopilot-relevant files, not the full `pytest tests/ -v`.


MakazhanAlpamys:
@AmirF194 — reviewed `153adac6`. **The fix is right, and the non-obvious half of it is right for the non-obvious reason.** Four items before I merge, one of which is a user-facing regression I reproduced.

## The scoping is correct in both directions, checked by execution

The only real consumers in `src/` are `trainer/sft.py:1293` (`use_liger` -> `apply_liger_kernel`), `:882` (`use_liger_kernel` into training kwargs) and `:1351` (`use_flash_attn` -> `attn_implementation`) — and `attn_implementation` appears nowhere else in any trainer.

**`tts` belongs in that allowlist and most people would have missed it.** `TTSTrainerWrapper` subclasses `SFTTrainerWrapper`; `sft.py:581` reads `use_audio = cfg.modality == "audio"` while TTS sets `modality="audio_out"`, so it falls through to `_setup_transformers`, and `tts.py:138-144` overrides that method and calls `super()`. It really does read both flags. Getting that right is the difference between a scoping change and a scoping *bug* — declaring something unsupported that is supported is just as bad as the reverse.

The counterexamples check out too: `BitNetTrainerWrapper.setup()` raises before reaching `_setup_transformers`, `pretrain.py` imports the class only for a static helper, and `mlx_sft.py:210-214` reads both flags solely to list them as unsupported.

Blast radius on shipped config, measured through the real loader: of 167 recipes plus templates plus `examples/configs/`, exactly two carry either flag — `llama3.1-8b-longctx` and `templates/longcontext.yaml` — and **both are `task: sft`**. Nothing shipped breaks.

**Zero removed test lines**; the `tests/` half is purely additive. Mutations, both killed, both anchors counted exactly once and both modules genuinely re-imported: disabling the schema gate gives `DID NOT RAISE ValidationError` six times over plus the CLI dry-run test, and reverting the autopilot half gives three failures that raise `ValidationError` *inside* `build_soup_config`. That second one makes the coupling explicit — with the validator in place the autopilot scoping is **mandatory**, or `soup autopilot` on Ampere+ crashes for those three goals. 912 passed, 2 skipped on the merged tree; the new file also passes under `FORCE_COLOR=1`.

## 1. The fragment credits me instead of you

```
(#806 by @MakazhanAlpamys in #983)
```

I am the issue reporter. You wrote the code. House style is `(#issue by @author in #PR)`, and I confirmed the convention by execution rather than by reading it — e.g. `919.fixed.md` says `by @Amix29` where the PR author is Amix29 and the issue author is me. Please make it `(#806 by @AmirF194 in #983)`.

I am asking for this rather than fixing it myself because the CHANGELOG is the file people actually read, and a release that renders a contributor's work as the maintainer's is the cheapest possible way to lose them.

## 2. REQUIRED — `soup migrate --from axolotl` now emits a config that will not load

`migrate/axolotl.py:133-134` maps `flash_attention: true` -> `use_flash_attn: true` regardless of `rl:`. I ran an axolotl config with `rl: dpo` and `flash_attention: true` through the real migrator on both trees:

```
your branch : migrate emits task=dpo, use_flash_attn=True
              -> REFUSED: training.use_flash_attn=true requires task in ['sft','tts']; got task='dpo'
current main: migrate emits the same
              -> LOADS
```

`soup migrate` does not validate its own output and still exits 0, so the user gets a config file that their next `soup train` rejects. The message is clear and correct when it finally arrives — but `soup migrate` is the on-ramp for people coming from Axolotl, and handing them a broken file on arrival is a bad first minute.

The flag was always meaningless there, so refusing it is right; migrating it is what is wrong. Drop it for non-`sft`/`tts` tasks and say so through the existing `_warnings` channel — the same shape you applied in autopilot. If you would rather do it in a separate PR, that is fine, but the two need to land together rather than leaving `migrate` broken in between.

## 3. Docs contradict the new behaviour

`docs/performance-and-quantization.md:831-832` still presents both flags with no task scope — the very lines #806 cites — while a config that follows them is now refused at load. That is a user-visible behaviour change, so the doc edit belongs in this PR.

## 4. LOW — a second unscoped construction site in the file you edited

`generate_config.generate_config()` (~line 165) builds a `SoupConfig` from a `decisions` dict and passes `perf.get("use_liger")` / `perf.get("use_flash_attn")` through unscoped. Reproduced on your branch: a `task: dpo` decisions dict with `perf.use_liger: True` now raises. **Severity is genuinely low** — it has no caller in `src/` and is not re-exported from `autopilot/__init__.py`, so no CLI path reaches it — but it is the same shape in the same file, and your own comment ("every other task would silently ignore them, so don't auto-enable there") applies to it verbatim. Scope it or delete it as dead code; your call.

## Noted, not asked for

The gate is still wider than the consumer along the **modality** axis: `task: sft` with `modality: vision|audio` accepts both flags, but `_setup_vision_transformers` and `_setup_audio_transformers` never read them — both consumer sites sit inside `_setup_transformers`. #806 names this; you narrowed the task axis only and filed `Refs` rather than `Fixes`, which is exactly right.

Items 1-3 before merge. Ping me and I will review only the delta.

AmirF194:
Pushed 54cd1a8, all four:

1. Changelog now credits @AmirF194.
2. `soup migrate --from axolotl` drops `flash_attention` for non-sft/tts tasks and says why through `_warnings`, same as verified against an `rl: dpo` config, it now loads.
3. docs/performance-and-quantization.md notes the sft/tts scope on both flags.
4. `generate_config`'s flat decisions path gets the same `SFT_KERNEL_AWARE_TASKS` gate as `build_soup_config`, not deleted since v0.32.0 callers still use it. Added a test with a dpo decisions dict carrying both flags on.

62 tests in test_migrate.py, 3 in the autopilot emission class, 13 in the #806 file, all passing. ruff check clean.

MakazhanAlpamys:
Re-reviewed `54cd1a89`. **All four closed**, and I re-ran the regression rather than trusting the diff. Merge-ready; waiting only on the CI now running.

**The one that mattered — `soup migrate --from axolotl`.** Same probe as before, an axolotl config with `rl: dpo` and `flash_attention: true`:

```
before your push : migrate emits use_flash_attn=True  -> REFUSED at load
after            : migrate emits use_flash_attn=None  -> the config LOADS
```

That is the right resolution and not the obvious one. The tempting fix is to make the loader lenient; instead you stopped migrating a flag the target task never reads, which keeps the refusal honest and fixes the on-ramp. Anyone arriving from Axolotl now gets a config that works instead of one that fails on their first `soup train`.

Fragment credits you (`#806 by @AmirF194 in #983`); `docs/performance-and-quantization.md` updated so a reader following the docs is no longer following something the loader rejects; and `generate_config()` is scoped, which I had marked LOW precisely because it has no `src/` caller — closing it anyway rather than arguing the severity is the right instinct for a second instance of the same shape in a file you were already editing.

Nothing needed from you. I will merge on green without re-reviewing.

MakazhanAlpamys:
Merged as `09e1a137`. Thank you — two PRs in one day, both correct.

**What I ran before merging.** Head `54cd1a89` was unchanged from the verdict I recorded at 07:57Z, so it was not re-reviewed. `main` had moved (#989 at 11:57Z), so I composed it onto `614b7f64` with #938, #982 and #984 and ran the batch as one tree. Because this PR touches `config/schema.py`, I deliberately included the two repo-wide guards it could plausibly move — `test_issue748_config_fields_reach_a_consumer` and `test_issue627_unknown_config_keys` — alongside `test_migrate`, `test_auto_tuning` and your new `test_issue806_kernel_flags_task_scoped`: **612 passed, 2 skipped, zero failures.**

**Three things this got right that the obvious version gets wrong.**

The allowlist is `{sft, tts}`, not `{sft}`. `TTSTrainerWrapper` inherits `sft.py::_setup_transformers`, so tts reads both flags too. A scope-by-task-name reading of the issue lands on `sft` alone and silently breaks every TTS config that sets either flag. You found that by following the inheritance rather than the issue text.

The migrate regression is the one I re-ran rather than trusting the diff, and the resolution is not the obvious one. Same probe as before, an axolotl config with `rl: dpo` and `flash_attention: true`:

```
before your push : migrate emits use_flash_attn=True  -> REFUSED at load
after            : migrate emits use_flash_attn=None  -> the config LOADS
```

The tempting fix is to make the loader lenient. Instead you stopped migrating a flag the target task never reads, which keeps the new rejection strict and still leaves `soup migrate --from axolotl` producing a loadable config. You also gave `generate_config`'s flat decisions path the same gate rather than deleting it, because v0.32.0 callers still use it — that is the kind of thing that is easy to miss and expensive to discover.

Rejecting at load rather than wiring the flags into the other trainers is also the right call, and your reasoning is the right reasoning: neither flag silently no-ops today, because the DPO/GRPO/pretrain paths never build an `attn_implementation` kwarg or a Liger patch call at all, and #745 set that precedent for `loraplus_lr_ratio`.

**And you left an acceptance criterion open on purpose, and said so.** The "same shape, narrower" `reasoning_effort`/`train_on_eot` note in #806 is a different function with a different consumer (`build_format_row`), and you had not checked whether the TTS formatting path calls it the way sft's text path does. `Refs #806` rather than `Fixes` is the honest bookkeeping, and it is the second time today you have made that call yourself.

#806 stays open with that criterion. It is yours if you want it.

AmirF194:
Thanks for re-running the axolotl migrate probe instead of trusting the diff, that's exactly the regression I'd have missed. Good call closing generate_config's gate too even with no src/ caller today.

PR #984, opened 2026-09-15
test: make five colour-dependent tests answer the same way in every shell (#886)

### LINKED ISSUE #886: v0.75.0 added three tests that fail whenever colour is on, including one assertion that is unsatisfiable by construction

Srinivasan8888:
Closes #886.

**Apologies for the delay on this** — I claimed it on 09-12 and then left it sitting for three days while the scope grew under it. That is why you ended up doing the work yourself; no complaint about that from me.

## Three defects, three repairs

The issue body — the half I wrote — prescribed the narrow-assertion fix for all of them. That was wrong, and worth stating plainly since anyone following it would have landed a still-red test.

```
TERM=dumb                          148 passed
FORCE_COLOR=1 TERM=xterm-256color    5 failed, 143 passed
```

**1. `test_issue627.py:1097`, both parametrisations — the Console, not the assertion.**

The fixture builds `Console(file=buffer, width=200, markup=True)` with `force_terminal` unset, so Rich reads `FORCE_COLOR` from the shell and colourises into the buffer. Pinning it makes the fixture answer the same way in every shell.

Taking your ruling: **alternatives, not complements.** With the console pinned, the broad `"\x1b" not in out` is the strictly stronger question, because the only ESC that could appear is the payload's. So this test keeps the broad form and does *not* also carry the narrow one.

The narrow fix alone would not have sufficed anyway — a detail neither of us had when the issue was written. The same test asserts `"[bold red on white]" in out`, and Rich's highlighter splits that literal:

```
'training.\x1b[0m\x1b[32m[\x1b[0m\x1b[32mbold red on white\x1b[0m\x1b[32m]\x1b[0m'
```

Two assertions, two reasons, one cause.

**2. `test_issue627.py:821` — not the ESC class at all.**

You flagged it as reproduced-but-not-root-caused, correctly. Colour splits the banner: raw output `False`, `strip_ansi(output)` `True`, and the stripped line is exactly `--- Run 1/1: sweep_1 ---`. Repaired with `strip_ansi`, not a narrower assertion.

**3. `test_doctor.py:494` and `:518` — the two you added.**

Same substring class. That file already defines `_strip_ansi` at `:19` and its newer tests use it, so these two simply predate the helper, exactly as you said. Mechanical.

## Verification

```
TERM=dumb                          148 passed
FORCE_COLOR=1 TERM=xterm-256color  148 passed
FORCE_COLOR=3                      148 passed
ruff check tests/                  All checks passed
```

| Mutation | dumb | colour |
|---|---|---|
| Console pin removed (`:1097`) | 148 passed | **2 failed** |
| `strip_ansi` removed (`:821`) | 148 passed | **1 failed** |
| `_strip_ansi` removed (NCCL) | 148 passed | **1 failed** |
| `_strip_ansi` removed (H100) | 148 passed | **1 failed** |

Four killed. Each dies only under colour, which is the property being fixed — a repair that failed under `TERM=dumb` too would mean I had changed behaviour rather than removed an environmental dependency.

## Scope

No changelog fragment: test-only, nothing user-visible. Say if you want one anyway.

**Not** taking the scanner, contrary to what I said when I claimed this. The rule as specced would flag 13 bare ESC-absence assertions; I measured them, and six are live failures in other files (`test_replay`, `test_review_fixes_v07133`, `test_v07127`, `test_v07128`, `test_v07136`, `test_v07139` — one each) while seven are fine because they assert against pure-function returns rather than console output. Landing a gate that reddens six unrelated files is a six-file sweep wearing a hygiene fix. Worth its own issue with the scanner as the closing move; happy to file it, or to do it here if you would rather have it in one go.

Noted on `#883`: it touches `test_doctor.py` too, but adds its own `_strip_ansi` call in a new test and does not touch either line I changed.


MakazhanAlpamys:
@Srinivasan8888 — reviewed `9339cde5`. **Merge-ready, and the part I want to single out is that you overruled your own issue.** One two-minute ask before I merge.

## What you did that the issue told you not to

#886's body — which you wrote — prescribed *narrowing* the `:1097` assertions to three hand-picked escape sequences. The PR rejects that and fixes the **Console fixture** instead (`force_terminal=False`), leaving `assert "\x1b" not in out` intact. That is the strictly stronger question, and narrowing it would have traded reach for greenness in a test whose whole job is reach.

You then went past it: `assert "\x07" not in out` and `assert "]0;owned" in out` are new. That second one closes a hole neither of us had named — **a sanitiser that swallows the payload entirely would pass every absence assertion**. I mutated exactly that (`.replace("]0;owned","")`) and it fails with *"the stripped payload must still be visible as inert text"*.

## Verified by running it, on both trees

| tree | plain | `FORCE_COLOR=1 TERM=xterm-256color` |
|---|---|---|
| `main` | 5 passed | **5 failed** |
| your head | 5 passed | **5 passed** |

The pre-fix failures are the real thing, not a synthetic one:

```
AssertionError: assert 'NCCL bandwidth requires >=2 GPUs' in '\x1b[1mSoup D...
AssertionError: assert 'Result (H100 over NVLINK)' in '\x1b[1mSoup Doctor\x...
```

Three distinct mechanisms across the five — an SGR-wrapped banner, the `--- Run 1/1` banner split, and Rich's own SGR tripping `"\x1b" not in out`.

**The strip is the house pattern**, checked rather than assumed: both `conftest.strip_ansi` and `test_doctor._strip_ansi` are `re.sub(r"\x1b\[[0-9;]*m", "", ...)` — a real SGR strip, not the whitespace-collapse that has been mistaken for one here before (`15b57f7`).

**All five changes are equal-or-stronger; none is a loosening.** Five removed lines, five replacements. The negative assertion in the sweep test is *strictly* stronger after the change: stripping can only add substring matches, so `"does not match any config field" not in plain` rejects more output than the raw form did.

**Mutation in both directions**, which for a test-only PR is the whole review: reverting `_strip_ansi` fails 1 under colour and dropping `force_terminal=False` fails 2, so the fix is load-bearing; and with the fixed assertions left intact, changing the production message, breaking `strip_control`, or deleting the payload each still fail — so robustness did not cost discrimination.

## Scope, stated accurately

All five are **latent, not currently red in CI**: no workflow sets `FORCE_COLOR` or `TERM`, and `CliRunner`/`StringIO` are non-ttys, so Rich disables colour on every runner. That matches your own framing on the issue. They are real for anyone running with colour forced — and this class has turned CI red four times when something upstream changed, so closing them before that happens again is the point.

**Declining to extend the repo-wide scanner was right, and I checked your reason independently.** On the merged tree under forced colour, `tests/test_replay.py::TestCli::test_replay_renders` and `tests/test_runs.py::test_runs_show_strips_control_bytes_from_the_error_message` both fail on a bare `assert "\x1b" not in result.output`. A broad ESC-absence rule landed here would redden unrelated files. Those two are the concrete scope for the follow-up issue — which does not exist yet; I will file it rather than leave it to you.

Merge is clean against current `main`; 179 passed on the merged tree both plain and under forced colour, including #883's new `test_doctor.py` test with no interaction. `ruff` clean, 14/14 on your head.

## The one ask

**No changelog fragment.** The validator only catches *stale* fragments, so nothing is red — but the repo's precedent is broader than "user-visible": #870 (README ratchet) and #907 (sanitizer de-duplication plus a new repo-wide test) are both test/infra-only and both carry one. More to the point, without a fragment your work gets no line in the release notes, and the CHANGELOG is the file people actually read.

`changelog.d/0.75.0/984.fixed.md`, first line starting `- `, credited `(#886 by @Srinivasan8888 in #984)`.

## Two nits, not worth a round on their own — fold them in if you are pushing anyway

- `tests/test_doctor.py:521` — `assert "Measuring NCCL bandwidth" in result.output` sits in the *same test* as one of your two repairs and is left raw. It passes under colour today, so it is harmless; but one of two sibling assertions being un-normalised invites the next reader to think the difference is meaningful.
- `:829`'s `from .conftest import strip_ansi` is mid-body where the file's own convention (`:790`) puts function-local imports first.

Push the fragment and I will merge without re-reviewing.

Srinivasan8888:
Pushed as `c282237`. It has the fragment and both nits.

**Fragment:** `changelog.d/0.75.0/984.fixed.md`, first line starting `- `, credited `(#886 by @Srinivasan8888 in #984)`. The fragment guard passes (21 passed). You're right that "user-visible" was the wrong bar: #870 and #907 set the precedent, and I'd prefer this change have a line in the release notes.

**Nit 1:** `test_doctor.py:521` now goes through `_strip_ansi` like its sibling on `:522`. It already passed under colour, so nothing changes today; it just stops the two assertions from looking deliberately different.

**Nit 2:** the function-local `from .conftest import strip_ansi` has moved from mid-body into the sweep test's own import block, per the file's convention.

```
TERM=dumb                          148 passed
FORCE_COLOR=1 TERM=xterm-256color  148 passed
ruff check tests/                  All checks passed
```

Two test lines removed, both accounted for: the `:521` assertion was replaced by its normalised form, and the import was re-added three lines higher in the same function. No assertion was dropped.

Thanks for filing the scanner follow-up with the two concrete sites (`test_replay_renders`, `test_runs_show_strips_control_bytes_from_the_error_message`).

MakazhanAlpamys:
Re-reviewed `c2822376`. **All three closed.** Merge-ready; waiting only on the CI now running.

Fragment is there and correctly credited — `(#886 by @Srinivasan8888 in #984)` — and the prose is better than the ask: it names the three distinct mechanisms rather than saying "colour", which is what makes it useful to someone reading the release notes six months from now.

Both nits folded in: `:521` now routes through `_strip_ansi` so the two sibling assertions in that test are consistent, and the `from .conftest import strip_ansi` moved to the top of the function to match the file's convention.

My verdict from the last round stands unchanged and I have not re-derived it. Nothing needed from you.

MakazhanAlpamys:
Merged as `8fe1d035`. Thank you — and the apology was not needed, but the thing you did with it was.

**What I ran before merging.** Head `c2822376` was unchanged from the verdict I recorded at 07:57Z, so it was not re-reviewed. `main` had moved (#989 at 11:57Z), so I composed it onto `614b7f64` with #938, #982 and #983 and ran the batch as one tree, including `test_cli_help_assertions_are_ansi_safe` — the repo-wide scanner this change sits next to: **612 passed, 2 skipped, zero failures.**

**The most valuable thing in this PR is that you contradicted the issue body — which you wrote.**

> The issue body — the half I wrote — prescribed the narrow-assertion fix for all of them. That was wrong, and worth stating plainly since anyone following it would have landed a still-red test.

That is worth more to this project than the patch is. Three defects wearing one symptom, and the prescribed repair was right for none of them: one was the `Console` construction, not the assertion; one was not the ESC class at all (a banner that colour splits, stripped to exactly `--- Run 1/1: sweep_1 ---`); two were raw-output matches that needed `strip_ansi`. Anyone who had picked up that issue and followed it to the letter would have shipped a test that was still red, and blamed themselves.

**The ruling you took, and then improved on.** I said alternatives, not complements. You applied it and then gave me the reason the narrow fix could not have worked *alone*, which I did not have when I ruled: the same test asserts `"[bold red on white]" in out`, and Rich's highlighter splits that literal across four SGR runs. Two assertions, two reasons, one cause. Pinning the `Console` keeps the broad `"\x1b" not in out` — which, once the console is pinned, is the strictly stronger question, because the only ESC that could appear is the payload's.

**And you strengthened a test nobody asked you to strengthen.** The control-byte test now also asserts the stripped payload stays visible as inert text, so a sanitiser that swallowed it entirely would no longer pass. That was not in the issue and not in my review.

On the three days: you claimed it on 09-12 and the scope grew under it, which is exactly what the diff shows happened — it stopped being "narrow the assertion" and became three unrelated mechanisms. Saying so plainly is the correct handling, and it is the reason I know the delay was scope and not silence.

The remaining gap is filed as **#987** — the scanner covers only `--help`, while `ReprHighlighter` also colours quoted strings and numbers inside error messages. #886 closes with this.

PR #985, opened 2026-09-15
fix(edit): refuse both models with no --probes, the fourth combination (#880)

### LINKED ISSUE #880: soup edit diff with both models but no --probes exits 0 with an empty diff (the mirror of #818)

Srinivasan8888:
Closes #880.

**Apologies for the delay** — you unblocked this on 09-13 at 12:43 and I am two days late to it.

## I checked the thing you asked me to check first

> *confirm the two refusals do not overlap or shadow each other — if the `--probes` case is already reachable through #863's path, the right patch may be smaller than the issue describes*

They do not overlap. Measured on `main` at `2dd84c2`, all six combinations:

| invocation | current `main` |
|---|---|
| **both models, no `--probes`** | **no refusal, `total_probes=0`, exit 0** |
| probes, only `--before-model` | refused (#863) |
| probes, only `--after-model` | refused (#863) |
| probes, no models | refused (#863) |
| neither model, no probes | no refusal — valid, criterion 3 |
| both models + probes | proceeds to live generation |

#863's guards are `(before_model is None) != (after_model is None)` and `probes and (before_model is None or after_model is None)`. Neither can reach the both-models-no-probes case, so the patch is the size the issue describes rather than smaller.

## Placement is the part worth reviewing

The guard goes **after** `load_probes`, because it reads `probes`. Its order relative to the pairing check does not matter: that check fires only when exactly one model is given, this guard only when both are, so no input reaches both. *(Corrected after review: an earlier version of this paragraph said the order mattered.)*

That is why every test here asserts the message rather than the code:

| Mutation | Result | First killer |
|---|---|---|
| the new guard removed entirely | **killed** | `test_it_is_refused_rather_than_reporting_an_empty_diff` |
| guard fires on any missing probes | **killed** | `test_neither_model_and_no_probes_still_succeeds` |
| guard placed *before* the pairing check | **equivalent — not killable** | no test can fail: the two guards are mutually exclusive by input, so their order changes no outcome *(corrected after review; this row previously claimed a kill)* |
| message no longer names `--probes` | **killed** | `test_the_message_names_the_missing_flag` |

Four run, four killed. The second and third are the ones that matter: they are how a guard that shadows its neighbours, or one that can never fire, would otherwise look green.

## Live

```
$ soup edit diff before after --before-model ./base --after-model ./edited -o d.json
Cannot build diff: --probes is required when both --before-model and --after-model are given;
without probes there is nothing to generate and the report would be an empty diff rather than
a measured one
exit 2 — d.json not written

$ soup edit diff before after -o ok.json
Total probes: 0; 0 changed.
Wrote diff -> ok.json
exit 0
```

`d.json` staying unwritten is deliberate and has its own test — the failure mode in the issue is a script that reads that file and believes it, so a refusal that still leaves "nothing changed" on disk would close nothing.

## Acceptance criteria

- [x] both models with no `--probes` exits non-zero and the message names `--probes`
- [x] every combination #863 refuses still refuses, each with its own message
- [x] neither model and no probes unchanged — exit 0, `total_probes: 0`
- [x] a test that fails against current `main` (three do)
- [x] the tests assert the message, not only the exit code

## Verification

```
pytest tests/test_issue880...py tests/test_issue818...py   16 passed
ruff check src/soup_cli/ tests/                            All checks passed
git diff -- tests/ | grep -c "^-[^-]"                      0
```

Exit 2 follows `commands/edit.py`'s nine existing `raise typer.Exit` sites rather than the repo-wide 1-for-usage rule, per the issue and #813. No changelog fragment yet — the guard allows one per number and I would rather add it in whatever version directory is current when you are ready to merge, given how the last one moved under me.

Full suite running; I will post the number.


Srinivasan8888:
Full suite:

```
21888 passed, 126 skipped, 5 deselected, 1 failed   in 15:51
coverage 83.66% (gate 77%)
```

**The one failure is pre-existing and not from this branch** — `test_issue580_autodistill_mlx_worker.py::test_worker_verification_survives_a_real_venv_launcher`, which builds a real venv and launches an MLX teacher worker that exits 1 on this machine. This branch touches no autodistill files, and the identical single failure appears on my #763 and #906 suites too: three unrelated branches, same test, same machine. It looks environmental to this box rather than to `main`, and CI is the arbiter; I am naming it rather than rounding the line to "21888 passed".

MakazhanAlpamys:
Reviewed `904cf55c`. **The fix is correct and the tests are good. One thing blocks the merge: the changelog fragment is missing.** Everything else below is cosmetic, and one item is a correction to your PR body that is in your favour to know about.

I verified this by execution rather than by reading the diff. All six combinations, `soup edit diff runA runB --output d.json`:

| invocation | `main` @ `614b7f64` | your head |
|---|---|---|
| **both models, no `--probes`** | **exit 0, `d.json` WRITTEN, `total_probes=0`** | **exit 2, no file**, message names `--probes` |
| `--probes`, only `--before-model` | exit 2, "`--after-model` is required…" | identical |
| `--probes`, only `--after-model` | exit 2, "`--before-model` is required…" | identical |
| `--probes`, no models | exit 2, "both … are required…" | identical |
| neither model, no `--probes` | exit 0, `total_probes=0` | identical |
| both models + `--probes` | reaches live generation | reaches live generation |

That is exactly what #880 asked for, and #818/#863's three combinations are byte-identical — the fourth was added without disturbing the other three, which is the part that is easy to get wrong.

**Tests.** `test_issue880_*` + `test_issue818_edit_diff_validation`: 16 passed. The other `build_diff_report` callers (`test_v0610_part_e`, `test_v0719`): 104 passed, 6 skipped. Repo ANSI ratchet: 9 passed. `ruff` clean on both changed files. **Zero removed test lines** — I checked, the diff is `+16/-0` in `edit_diff.py` plus a new `+129/-0` test file.

**Mutations — three of four killed, module genuinely imported in every one** (a real `import soup_cli.utils.edit_diff`, not `ast.parse`):

| anchor (occurrences) | mutated to | killed by |
|---|---|---|
| `if before_model is not None and after_model is not None and not probes:` (**1**) | `… and probes is None:` | 3 tests; `assert 0 == 2` and `assert '--probes' in …` |
| same anchor (**1**) | `if not probes:` | your `test_neither_model_and_no_probes_still_succeeds` **and** the pre-existing #818 test |
| `"--probes is required when both --before-model and --after-model "` (**1**) | reworded without the flag name | `test_the_message_names_the_missing_flag` |

The second one is the good news: over-broadening the guard is caught by **both** your new control and an existing #818 test, so the two guards hold each other in place.

---

### 1. Blocking: `changelog.d/0.75.0/985.fixed.md` is missing

Confirmed absent at your head and on `main`. Your stated reason for deferring — that the release directory moved under you last time — no longer applies: `0.75.0` is the live directory, and `989.fixed.md` landed there on `main` today. Note that **nothing catches this automatically**; `publish.yml --check-empty` and the repo test detect *stale* fragments, not absent ones, which is why I am asking rather than CI.

First line starting `- `, and the credit line is **`(#880 by @Srinivasan8888 in #985)`** — #863 is @kok-o's, this one is yours.

### 2. A correction to your PR body, in your favour to know

Your mutation table lists "guard placed *before* the pairing check → killed → `test_probes_without_either_model`". **I could not reproduce that.** I moved the guard above the pairing/XOR check and got **44 passed, 1 skipped** — nothing failed.

The reason is structural rather than a gap in your tests: the two guards are mutually exclusive by construction. The new one requires *both* models non-`None`; the XOR guard fires only on *exactly one*. No input can reach both, so their relative order cannot change any outcome. Your placement is still correct — it must sit after `probes = load_probes(...)` — that is just a different constraint from the one the body and the code comment claim.

Which leads to:

### 3. `src/soup_cli/utils/edit_diff.py:255-257` — the comment's rationale is wrong

> *"Placed AFTER the pairing check above so that supplying one model and no probes still names the missing model"*

One model + no probes never reaches this guard regardless of order. Please replace it with the real constraint: the guard must follow `probes = load_probes(...)`, because it reads `probes`.

### 4. `src/soup_cli/utils/edit_diff.py:223` — docstring is now half the rule

Still ends `"Probes require both models."` The rule became symmetric. Something like `"Probes require both models, and both models require probes."`

### 5. `src/soup_cli/commands/edit.py:334-338, :347, :351` — help text contradicts the new refusal

`--probes` reads `"Optional JSONL file with probe prompts."` It is no longer optional when both models are given. Same for `--before-model` / `--after-model`, which say `"(enables live generation)"` — now true only together with `--probes`.

---

**Docs need no change** — I checked rather than assumed: `docs/adapters-and-governance.md:158-161` and `docs/commands.md:244` already show the models and `--probes` together, so nothing there contradicts the refusal.

Your head is not a descendant of current `main` (base `2dd84c2`), though GitHub still reports it mergeable and no commit on `main` touches `utils/edit_diff.py`, `commands/edit.py` or the #818 test file. Adding the fragment forces a push anyway, so rebase while you are there.

One more thing, since you flagged it on your own suite: the `test_issue580_autodistill_mlx_worker` failure you saw on three unrelated branches is environmental to that box, not to `main`. A venv's `Scripts\python.exe` on Windows is a redirector that launches the real interpreter as a grandchild, so `os.getpid()` and `Popen.pid` legitimately differ. It is tracked as **#898**. Your reading of it was right.

Srinivasan8888:
Pushed as `eb08b60`, rebased onto `main` (`da17fd4`). Both commits are in this push.

**1. Blocking.** Added `changelog.d/0.75.0/985.fixed.md`. Its first line starts with `- ` and credits `(#880 by @Srinivasan8888 in #985)`. `test_issue487_changelog_fragments` passes.

**2. You're right about the mutation row, and it's my mistake.** The row "guard placed *before* the pairing check -> killed" was wrong. The two guards can never both fire on the same input: the pairing check needs exactly one model, and this guard needs both. So their order can't change any outcome, and a mutation that only changes the order is equivalent. I should have reported that row as not killable instead of claiming a test caught it.

**3. The comment at `edit_diff.py:255`** now gives the real constraint. The guard must come after `load_probes`, because it reads `probes`. The comment also says its order relative to the pairing check doesn't matter, and why.

**4. The docstring** now reads "Probes require both models, and both models require probes."

**5. Help text:**
- `--probes`: *"JSONL file with probe prompts; required when both --before-model and --after-model are given."*
- The two model flags: *"(live generation with --after-model and --probes)"*, and the matching wording on `--after-model`.

I checked the rendered `soup edit diff --help` output, and no test or doc pins the old strings.

**Verification:** `test_issue880_*` + `test_issue818_*` + `test_v0610_part_e` + `test_v0719` + `test_issue487_changelog_fragments`: 147 passed. ruff is clean on both files. Zero removed test lines.

CI on this push will pick up peft 0.21.0, so expect the nine #1005 layer-streaming reds until #1010 lands. They're unrelated to this diff.

MakazhanAlpamys:
**All five closed. Merge-ready.** Queued behind #1010 (the peft 0.21 fix), which is the only thing holding it.

**The fragment is there and correctly credited** — `(#880 by @Srinivasan8888 in #985)`, first line starts `- `, and `test_issue487_changelog_fragments` passes. That was the blocking one.

**The comment rationale is now right, and better than I asked for.** I only said the old text was wrong; you replaced it with the actual constraint *and* the reason the two guards cannot interact:

> Must follow `load_probes` above, because it reads `probes`. Its order relative to the pairing check does not matter: that check fires only when exactly one model is given, this one only when both are.

That second sentence is the part that stops someone "fixing" the order later.

**Help text and docstring both closed**, and I grepped the tree rather than the diff: zero remaining occurrences of `"Optional JSONL file with probe prompts"` or `"(enables live generation)"` in any `.py` or `.md`. Nothing was pinning the old strings, so nothing silently kept them alive.

**Re-ran all six combinations against `origin/main`**, since your head is now rebased onto `da17fd4e`:

```
both models, no --probes : main exit 0, file WRITTEN, total_probes=0
                           head exit 2, no file, message names --probes
the other five           : byte-identical to main
```

Only the fourth combination moved. That is the whole ask of #880 and nothing else shifted.

**Mutation:** anchor `if before_model is not None and after_model is not None and not probes:` matched **exactly once**, forced to `if False:`, module import confirmed. **3 failed** — `test_it_is_refused_rather_than_reporting_an_empty_diff`, `test_the_message_names_the_missing_flag`, `test_no_diff_file_is_written` — and the failing assert is the empty diff itself (`Total probes: 0; 0 changed.`), which is #880's symptom rather than a proxy for it. Reverted, 37 passed. Zero removed test lines. Ruff clean.

---

**One cosmetic leftover, and you may reasonably ignore it.** You retracted the false mutation row in your 18:36Z comment — *"I should have reported that row as not killable"* — which is the right handling and is on the record. But the **PR body** still carries it:

```
| guard placed *before* the pairing check | **killed** | test_probes_without_either_model |
```

A squash merge takes the body into the commit message, so that row outlives the thread. One edit if you want the permanent record to match the correction; I am not going to hold the merge for it.

Thank you for correcting your own claim before I had to press it — that is worth more than the row being right in the first place.

MakazhanAlpamys:
Following up on your one red cell, since I said I would not guess at it: **`test (macos-latest, 3.12)` is #1005, not you.** Read from the job log rather than inferred:

```
8 failed, 22367 passed, 160 skipped, 5 deselected
peft-0.21.0

FAILED tests/test_v07201.py  (5)
FAILED tests/test_v07202.py  (2)
FAILED tests/test_v07203.py  (1)
```

Every failure is in those three files and `peft-0.21.0` is in the install log. Your diff touches `utils/edit_diff.py`, `commands/edit.py` and their tests — nothing under `trainer/` or `layer_stream*`. Nothing to do; the remaining eleven cells will land the same way until #1010 merges.

For anyone hitting this next: `gh run view --log` refuses while a run is still in progress, but the **job**-level endpoint works if you let it through the escape sequences —

```
gh api --allow-escape-sequences repos/OWNER/REPO/actions/jobs/<job_id>/logs
```

Without that flag it returns nothing and the silence looks like an empty log rather than an encoding refusal.

MakazhanAlpamys:

Five cells — `test (ubuntu-latest, 3.10/3.11/3.12)`, `test (macos-latest, 3.12)`, `test (windows-latest, 3.10)` failed with nine tests from `test_v07201.py` / `test_v07202.py` / `test_v07203.py`, all with the same shape:

```
                which raises no exception
```

```
your run created   2026-09-15T18:36:19Z
```

MakazhanAlpamys:

```
git fetch --depth=1 origin +<a pinned merge SHA>:refs/remotes/pull/985/merge
git checkout --force refs/remotes/pull/985/merge
```

```
```

## What actually clears it

Srinivasan8888:
Option 1 — pushed the rebase, so no `Update branch` needed. Head is now `d20927e`, rebased onto `f8226214` (current `main`); the two commits are unchanged apart from their base.

No apology needed for the wrong steer: you corrected it within six minutes of posting it, and the correction came with the `git fetch --depth=1 origin +<pinned SHA>:refs/remotes/pull/985/merge` line that makes it checkable rather than something I had to take on trust.

Local on the rebased tree: `test_issue880_*` + `test_issue818_*` + `test_issue487_changelog_fragments` — 37 passed.

MakazhanAlpamys:
Merged as e8241e0b. Thank you, @Srinivasan8888.

You rebased onto the #1010 merge as soon as the frozen merge ref was explained, and changed nothing else: `git range-diff` shows both commits as `=` against the head I had called merge-ready. You also re-ran the #880, #818 and #487 tests on the rebased tree before saying so.

What I ran on the tree merged with `main`: F821 clean, and 141 passed across the #880, #818, #487, `v0610_part_e` and `v0719` tests. I then turned the guard at `edit_diff.py:258` into `if False:`. Three tests fail on the user-visible symptom itself: exit 0 with an empty table and "Total probes: 0; 0 changed", no `--probes` in the message, and a diff file written.

PR #986, opened 2026-09-15
fix(train): report unavailable loss for MLX and unlearn

### LINKED ISSUE #960: mlx_sft and unlearn still report Loss: 0.0000 where there was no measurement

Rayan-and-beyond:
## Summary

- route MLX SFT and unlearn completion losses through the shared loss-summary helper
- report `unavailable` instead of inventing `0.0000` when neither backend captured a loss
- report one-step unlearn runs as `single` rather than a fake delta
- add behavioral regressions through the real Unlearn and fake-MLX-backed training paths

Closes #960

## Validation

- `tests/test_issue899_loss_summary.py` — 18 passed on Python 3.11
- loss-summary + MLX backend-support focused suites — 68 passed
- reintroduced the old `or [0.0]` fallback separately in Unlearn and MLX — each new no-loss behavioral test failed
- `ruff check` on the touched Python files — passed
- `git diff --check` — passed


MakazhanAlpamys:

Sorry for the noise.

MakazhanAlpamys:

MakazhanAlpamys:

I wrote:

```
```

## What actually clears it

## What has not changed

MakazhanAlpamys (changes requested):
Thanks, @Rayan-and-beyond, for the push. It cleared the frozen merge ref, and adding `loss_summary.py` to `TRAINER_MODULES` was what the #755 guard needed. You reused the shared helper instead of writing a third copy, which keeps `initial_loss` / `final_loss` numeric for sweep ranking (#960 criterion 4).

Running `UnlearnTrainerWrapper.train()` with a stubbed per-step loss on the tree merged with current `main`:
- **0 steps:** `main` prints `0.0000`, this PR prints `unavailable`. That is the fix.
- **3 steps:** `2.1000 -> 0.9000` on both, which is correct.
- **1 step:** `main` prints `Loss: 2.1000`, but this PR prints **`Loss: 2.1000 -> 2.1000`** as a delta. That is the fake-delta symptom #899 removed.

Three asks:
1. **The one-step regression.** `src/soup_cli/trainer/unlearn.py:325` passes `[initial_loss, final_loss]`, which are the same step when `step == 1`. Pass only `[final_loss]` when there is a single step, so the helper reports `single`.
2. **No behavioural test protects either fix.** I put the bug back inside the helper calls themselves, `(losses or [0.0])` in `unlearn.py` and `(captured.get("losses") or [0.0])` in `mlx_sft.py`. All 148 related tests pass while the run prints `Loss: 0.0000` again. The exact reverts are caught only by the AST structure guard. Please add tests that call the real code paths:
   - **unlearn:** `UnlearnTrainerWrapper.train()` with a stubbed step loss for 0, 1 and 3 steps, asserting `unavailable`, `single` and `delta`.
   - **MLX:** the result-building path with `captured["losses"]` set to `[]`, `[x]` and `[x, y]`.

   That is #960's third acceptance criterion.
3. **Rebase and credit.** #1018 added `soup_cli/trainer/rewind_mlx.py` on the same line of `TRAINER_MODULES[("sft", "mlx")]`, so `src/soup_cli/config/backend_support.py` now conflicts. Keep both entries. With that resolution I got 1061 passed. Also, `changelog.d/0.75.0/986.fixed.md` should credit `(#960 by @Rayan-and-beyond in #986)`.

Rayan-and-beyond:
Addressed all three review points: one-step unlearn now reports a single loss, both Unlearn and MLX have behavioral 0/1/many-loss coverage, current main is rebased with both rewind_mlx.py and loss_summary.py in the registry, and the changelog credit is updated. The two old zero-fallback mutations now fail the new tests.

MakazhanAlpamys:
Unblocked — I updated your branch, and the red cell was never your code.

`test (windows-latest, 3.10)` was dying with `Windows fatal exception: code 0xc000001d` inside `tests/test_rewind_hf.py` (issue #382): part of GitHub's `windows-latest` fleet lacks an instruction the wheel emits, so the interpreter dies with a faulthandler dump and no Python exception, and everything the cell had not yet reached goes unreported. That has nothing to do with the MLX and unlearn loss reporting you changed.

#1059 merged an hour ago and puts those eight real-training tests behind the shared `skip_on_windows_ci` guard, which fires only on `win32` **and** `CI=true`. I ran `gh pr update-branch` so your PR picks it up; CI is re-running.

Nothing owed on your side — I will merge on green.

MakazhanAlpamys:
Merged — thank you, and thank you for the patience: this sat behind a red cell that was never yours, and the delay was mine.

The defect is small in code and not small in kind. `Loss: 0.0000` where nothing was measured means a run that measured nothing looks exactly like a run whose loss genuinely reached zero — and zero is a number people act on. A placeholder that is indistinguishable from a real result is worse than no number at all, which is why this was worth doing properly rather than papering over with a different default.

You also addressed all three review points and rebased without being chased twice, which I appreciate more than the diff.

Verified before merging, on the tree rebuilt by merging your branch into current `main` (GitHub's merge ref is pinned at run creation, so the green marks described a `main` many merges behind), base interpreter, peft 0.20.0, `soup_cli.__file__` under the scratch tree:

- `ruff check --select F821 src/soup_cli/` — clean.
- The MLX, unlearn and loss-reporting suites on the merged tree — pass.
- A decorator-level diff of the merged tree against `main` — nothing lost. I added that check today after finding that a conflict resolution can silently drop a `@model_validator` line with `ruff` and the import both staying happy.

On the wait: your `test (windows-latest, 3.10)` was dying on #382's illegal-instruction crash inside `tests/test_rewind_hf.py`, a file you do not touch. #1059 put those eight real-training tests behind the shared `win32 AND CI=true` guard, and I ran `update-branch` so this picked it up.

`#960` closes with this.

Issue #975, opened 2026-09-14
The single shared large-layer slot forces a full embedding-matrix H2D every step on an untied checkpoint (~33 ms for 500 MiB, measured at 16 GB/s)


MakazhanAlpamys:
## What was measured

`benchmarks/harness/layer0_wait.py` (committed harness, re-run 2026-09-14
15:53 from the committed path after an earlier run was found to have come
from a session-scratchpad copy — see "Where it comes from" below), driving
the probe's own instruments and labelling each load bracket with the layer it
belongs to. Cold synthetic Llama-70B shape, NF4, depth (`read_ahead`) 2, 3
steps, no warm-up, instrumented from step 0. JSON:
`layer0_wait_cold_synth70b_nf4.json`. Absolute step times (49.2, 39.4, 39.3 s)
are explicitly NOT throughput evidence per the record (no warm-up, instrumented
throughout); the ratios within each step are what this measures.

**Per-load brackets** (record's table, §8):

| step | layer 0, forward / backward | other layers, mean / median / max | vocabulary loads | brackets |
|---|---|---|---|---|
| 0 | 17 / 271 ms | 252 / 283 / 420 ms | 662, 386 ms | 40.6 s of 49.2 s = 82.6% |
| 1 | 17 / 259 ms | 214 / 266 / 533 ms | 33, 22 ms | 33.5 s of 39.4 s = 84.9% |
| 2 | 16 / 268 ms | 213 / 270 / 327 ms | 33, 22 ms | 33.3 s of 39.3 s = 84.8% |

**Per-step compute-stream stalls, from the JSON's `stalls` array** (Finding
8a's table):

| step | total stall | `large:model.embed_tokens.weight` | `layer000` forward | mean, other 79 decoder layers |
|---|---|---|---|---|
| 0 | 33.17 ms | 32.66 ms | 0.17 ms | 0.0021 ms |
| 1 | 47.92 ms | 32.61 ms | 14.99 ms | 0.0020 ms |
| 2 | 47.58 ms | 32.91 ms | 14.35 ms | 0.0020 ms |

Record's own reading (Finding 8a, quoted): "The embedding out-stalls layer 0
in every one of the three steps, layer 0 is second in every one of them, and
the other 79 decoder layers (158 stall entries) sit at 0.002 ms — four orders
of magnitude below the embedding in all three steps (15,700x / 16,300x /
16,200x), and three to four below layer 0 in steps 1 and 2 (7,500x / 7,100x;
[...]

PR #994, opened 2026-09-15
fix(grpo): add log-prob clamping and NaN/Inf loss guards for 7B training stability

DelorneN007:
## Summary
This PR resolves the GRPO training instability issue at the 7B scale where runs frequently crashed due to `grad_norm nan` errors and probability tensors containing `inf/nan` values.

## Changes Made
* **Log-Probability Clamping**: Added `torch.clamp` to `old_per_token_logps` (bounded between `-20.0` and `20.0`) to prevent exponential underflows and overflows during policy updates.
* **Loss & Gradient Watchdog**: Implemented a check in `training_step` to catch non-finite (`NaN` or `Inf`) loss values, logging a warning and safely zeroing gradients (`model.zero_grad()`) to protect model weights from corruption.

## Testing & Verification
* Verified against the complete GRPO test suite (`tests/test_grpo.py`), with all **44/44 tests passing** successfully.


MakazhanAlpamys:
@DelorneN007 — welcome, and sorry for the wait. **Your CI was blocked, not slow.**

A first-time contributor's workflow run needs manual approval here, and yours had been sitting with **zero check runs** since 12:30Z. I have approved it (`34969284733`) and all thirteen required contexts are now running. That delay was mine.

Worth recording why I nearly missed it, since it will save the next person: `gh run list --json status` reports these runs as `status=completed, conclusion=action_required`, so a query filtering on `status == "action_required"` returns **nothing** and the queue looks clear. The API query that actually finds them is:

```
gh api "repos/OWNER/REPO/actions/runs?status=action_required"
```

There were eight, two of them on open PRs.

Note that the approval does **not** carry across pushes — if you push again before your first PR merges, ping me here and I will approve the new run.

I will review the diff properly once the matrix reports. Two things I will be looking at, so you can get ahead of them if you want:

1. **A mutation of the clamp itself.** The review standard here is that I break the fix in the source and confirm one of your tests goes red with a *behavioural* assertion — so the more your tests pin the clamp's boundary rather than its presence, the faster this goes.
2. **Whether the NaN/Inf guard can mask a real divergence.** A loss guard that silently skips a bad step is a different thing from one that reports it, and for a 7B GRPO run the difference matters: a run that quietly drops steps looks healthy in the loss curve. If your guard logs or counts what it skipped, say so in the PR body and I will not have to ask.

Neither is a complaint about code I have not read yet — just the two questions this class of change usually turns on.

MakazhanAlpamys:
Reviewed `f4ef02ad`. **The problem you are chasing is real, and `grad_norm: nan` on a 7B GRPO run is worth fixing.** I am going to be direct about the diff because that is more useful to you than encouragement, but I want the diagnosis credited first: you found a genuine instability and went looking for the mechanism, which is the right instinct.

The verdict is **blocking**, for reasons that are mostly fixable. Two of them are structural and change what the patch should look like.

---

## 1. The clamp cannot fire. It is dead code.

This is the one that matters most, so here is the proof rather than an assertion.

Your override reads the key **before** delegating:

```python
old = _read_attr(inputs, "old_per_token_logps")
...
loss = super().training_step(model, inputs, num_items_in_batch)
```

But `old_per_token_logps` is not on the batch yet at that point. TRL sets it in `GRPOTrainer._prepare_inputs` -> `_generate_and_score_completions`, and `_prepare_inputs` runs **inside** `Trainer.training_step` — verified on the installed stack:

```
transformers 5.17.0 | trl 0.29.1
Trainer.training_step calls _prepare_inputs: True
```

So `inputs` at your read is the raw dataloader batch. Reproduced in a harness with that exact ordering:

```
raw batch keys reaching training_step : ['prompt']
old_per_token_logps the loss SAW      : [[-1.0, -55.0, -100.0]]
=> CLAMP FIRED: False

CONTROL (key present on the raw batch) : [[-1.0, -20.0, -20.0]]
```

The clamp works mechanically — the control shows it — it is just wired one call too early. **The place where that tensor genuinely exists is `_compute_variant_loss`**, where `logp_old` at `grpo.py:122` already holds the real thing, after `_prepare_inputs` has run.

## 2. The default GRPO path gets none of this

Your `training_step` lives inside `_make_grpo_trainer_variant_cached`, and that class is only built here:

```python
variant = self.config.training.grpo_variant
if variant is not None and variant != "standard":
    GRPOTrainer = make_grpo_trainer_variant(GRPOTrainer, variant)
```

With `grpo_variant` unset — the default — neither the clamp nor the NaN guard exists at all. A plain 7B GRPO run, which is your stated motivation, is untouched.

## 3. The NaN guard, as written, makes a diverging run look healthier

This is the part I would push back on even after 1 and 2 are fixed.

`transformers.TrainingArguments.logging_nan_inf_filter` defaults to **`True`** and Soup never overrides it, so on a non-finite step the Trainer already **substitutes the running average** into the logged loss. The curve is already lying. Your patch then also suppresses the weight update via `model.zero_grad()`. Net result: a plausible loss, a `grad_norm` near zero, and **no training progress** — with nothing anywhere saying so.

That is the failure mode this project keeps filing issues about: a guard that converts a loud failure into a quiet one. A guard is fine; a **silent** guard on a 7B run that costs real money is not.

What would make it acceptable: a **count** surfaced in the run summary, not only a log line. Note also that `logger` here is `soup_cli.trainer.grpo`, while the Rich handler is installed on `"soup"` (`utils/log_level.py:21`) — `soup_cli` is not a child of `soup`, so this falls through to `logging.lastResort` and lands as raw stderr in the middle of the 2 Hz Live display, ungated by `--log-level`. Pre-existing for `_warn_fallback`, so not your doing, but it means you picked the least visible channel available. The file already has the pattern for this: `_warn_fallback` (`grpo.py:61-79`) is one-shot via `_soup_fallback_warned`, with the comment *"not on every step"*. Yours warns per micro-batch.

## 4. `model.zero_grad()` wipes the accumulation window

`training_step` runs once per micro-batch. One non-finite micro-batch at accumulation step *k* discards the *k-1* healthy gradients already accumulated — silently. Worth deciding deliberately rather than inheriting.

## 5. The bounds are unjustified, and clamping one side biases the objective

- `max=20.0` is unreachable: these are log-probabilities, always `<= 0`.
- No comment, citation or PR-body reasoning for `-20.0`. A magic number in a loss path needs a why.
- **Clamping only `old` and not `per_token_logps` changes the objective.** GRPO's ratio is `exp(new - old)`; flooring only the denominator's log at `-20` moves the importance-sampling ratio by up to `exp(80)` for a token at `-100` — which the control above shows being clamped. That is a silent change to the loss, not a numerical guard. Clamp both sides or neither.

## 6. `isinstance(inputs, dict)` is False for a `BatchEncoding`

`grpo.py:189`. The read side uses `hasattr(obj, "get")`, which **is** true for `BatchEncoding`; the write side takes the `else` branch and does `inputs.old_per_token_logps = clamped_old` — an attribute set that never reaches the mapping the loss reads.

This repo was bitten by exactly this in **#430**: assistant-only loss masking built its label mask from a mapping's key strings because `BatchEncoding` is not a `dict`, and the result was a normal-looking loss curve with **0 trained tokens**. Same trap, same class.

## 7. Nothing tests any of it — and this is the blocking one procedurally

I ran three mutations against `tests/test_grpo.py`, `test_issue723_sequence_gspo.py`, `test_v05311.py`, `test_v07111.py` (baseline **212 passed**):

| mutation | result |
|---|---|
| `clamped_old = torch.clamp(...)` -> `clamped_old = old` | **212 passed — survived** |
| `model.zero_grad()` -> `pass` | **212 passed — survived** |
| delete the entire `training_step` override | **212 passed — survived** |

Module imported cleanly after each. Your PR body says *"Verified against the complete GRPO test suite, with all 44/44 tests passing"* — that is true, and it is also true with the change deleted. **The house rule here is that a fix is reviewed by breaking it and watching a test go red.** Nothing does.

## 8. Mechanics

- **`ruff` fails on the head**: the diff strips the file's trailing newline (`\ No newline at end of file`). `pre-commit`'s `end-of-file-fixer` will also catch it. Hard CI blocker.
- **No changelog fragment** (`changelog.d/0.75.0/994.*.md`). Required for a user-visible change.
- **No docs note** for a behavioural change to the GRPO loss path.
- **No linked issue.** Nothing to hold against you, but it means the scope was not agreed before the work — please open one describing the 7B `grad_norm: nan` you hit, with the config and what you observed. That issue is worth having regardless of this PR's fate, and it gives the change a place to be argued.

---

## What I would suggest, concretely

Split it. The NaN/divergence reporting and the log-prob clamping are different changes with different risks:

1. **File the issue first** with your 7B reproduction. That is the durable artifact.
2. **Move the clamp into `_compute_variant_loss`** near `grpo.py:122` where `logp_old` is real — or drop it until the issue establishes it is needed. Clamp both sides, and justify the bound.
3. **Move the guard out of the variant factory** so default GRPO is covered.
4. **Make it loud, not quiet**: one-shot warning via the existing `_soup_fallback_warned` pattern, plus a non-finite-step **count** in the run summary.
5. **Decide the accumulation-window question** explicitly.
6. **Tests that fail when each half is removed.** That is what I will check.

None of this is a judgement on you — it is a first PR into a loss path, which is one of the harder places to land a first change. The instinct was right and the investigation was real. Take it in pieces and I will review each quickly.

Your CI was also blocked awaiting first-contributor approval, which I have since approved; note that approval does **not** carry across pushes, so ping me here after you push and I will approve the new run.

MakazhanAlpamys:

Sorry for the noise.

MakazhanAlpamys:

MakazhanAlpamys:

I wrote:

```
```

## What actually clears it

## What has not changed

src/soup_cli/trainer/grpo.py
```diff
@@ -180,6 +180,24 @@ def compute_loss(self, model, inputs, return_outputs=False, **kwargs):
...
+                if isinstance(inputs, dict):
+                    inputs = {**inputs, "old_per_token_logps": clamped_old}
+                else:
+                    inputs.old_per_token_logps = clamped_old
+
+            # Fix 1: Pre-step gradient watchdog check
+            loss = super().training_step(model, inputs, num_items_in_batch)
+            if loss is not None and (torch.isnan(loss) or torch.isinf(loss)):
+                logger.warning("Detected non-finite loss in training_step; zeroing gradients.")
+                model.zero_grad()
```
AmirF194, line 198: I pulled this method verbatim from the diff at f4ef02a and ran it against a toy param with gradient_accumulation_steps=2, in a clean container.

`model.zero_grad()` at line 198 runs after `backward()` has already accumulated `.grad` for this microbatch. In HF Trainer, `zero_grad()` is normally called once per accumulation window, before the first microbatch, not after each one, so a finite microbatch earlier in the same window has already added to `.grad` by the time a later microbatch's non-finite loss trips this branch. Calling `zero_grad()` here wipes that too, not just the current microbatch's contribution.

```
grad after finite microbatch 1: [2.0, 4.0]
grad after non-finite microbatch 2: None
RESULT: microbatch 1's valid accumulated gradient was wiped by the watchdog
```

(repro: same training_step body, a FakeBase.training_step doing backward()+return loss.detach() to stand in for transformers.Trainer, two calls with no zero_grad between them the way accelerator.accumulate() runs them)

The existing test_grpo.py has no isnan/isinf/zero_grad/accumulation coverage, so this path isn't exercised by the "44/44" in the PR description.

Checking finiteness on the loss before calling super().training_step() (skip backward for this microbatch, or substitute a zero-valued loss with the same grad_fn) would drop only this microbatch's own contribution instead of every other microbatch accumulated in the same window.

MakazhanAlpamys:
@DelorneN007 — this thread has been open for three days with a substantive technical review sitting unanswered on it, and that is entirely my fault. You are a first-time contributor here and this is not the experience anyone should have. Let me give you the whole picture in one comment rather than another partial reply.

## Three separate things, and only one of them is a problem with your code

**1. The red CI is not yours.** All nine cells fail on `test_v07201/02/03` with "no adapter tensors were written at all" / "every `lora_B` is zero". That is the peft 0.21.0 breakage (#1005), fixed on `main` by #1010 on 09-16. As I explained above, a re-run cannot clear it because the merge ref is frozen at run creation. I have run `gh pr update-branch` on other PRs in the same state today; say the word and I will do it here too, or you can merge `main` into your branch yourself.

**2. @AmirF194's review is correct, and it is the one thing that needs a change.** He pulled your `training_step` verbatim, ran it with `gradient_accumulation_steps=2` in a clean container, and got:

```
grad after finite microbatch 1: [2.0, 4.0]
grad after non-finite microbatch 2: None
```

`model.zero_grad()` inside `training_step` runs **per microbatch**, after `backward()` has already accumulated `.grad` for that window. HF Trainer zeroes once per accumulation window, before the first microbatch — so a finite microbatch earlier in the same window has already contributed by the time a later bad one trips your branch, and your call wipes that contribution too. With `gradient_accumulation_steps > 1` the watchdog therefore discards valid gradient, silently, in exactly the situation where the run is already unstable.

**The fix is a hook change, not a redesign:** `on_pre_optimizer_step` fires once per optimizer step, after the whole accumulation window, so zeroing there affects only the step that is actually about to be taken. If you keep working on this, that is the change, and it needs a test with two microbatches — first finite, second not — asserting the first one's gradient survives.

**3. The part I owe you an honest answer about: this overlaps an assigned issue.** Your PR references no issue, so you had no way to see this. #342 (`GRPO 7B grad_norm nan`) has been assigned to @harshitthek since 08-29, his plan was approved on 09-13, and on 09-14 he posted an implementation shape that is your log-prob clamp plus your gradient watchdog. Neither of you could see the other because nothing linked the two threads — **that is my failure to cross-link, not duplication on your part.**

The convention here is that the claim goes by who claimed the issue first, not by whose code arrives first. I applied that rule to two other contributors this week in the other direction, so I am applying it to you consistently rather than conveniently: #342 stays with @harshitthek.

## What I am not going to do is waste what you found

Your clamp bound and his are the same number by the same reasoning, arrived at independently — that is corroboration, and I have said so on #342. The microbatch problem surfaced **on your PR**, and it is now a named requirement on his implementation: I have asked him for the exact two-microbatch test that your reviewer's repro implies. Your PR is the reason that test will exist.

So, three ways forward, your choice:

1. **Take an adjacent piece with no owner.** #975 (`_plan_queue` stalls the compute thread on the vocabulary fetch) or #815 (`soup loop canary` writes three fields nothing reads) are both precisely diagnosed, unclaimed, and need no GPU. I will assign either to you the moment you say which.
2. **Stay on this one as a collaborator.** If @harshitthek is willing, the microbatch test is yours to write against his implementation, credited as yours.
3. **Narrow this PR to the part #342 does not cover** and tell me which part that is — I will review it on its own terms.

If you would rather do none of those, that is a fair answer too, and I would still rather hear it than have you leave without one.

@AmirF194 — thank you for that review. Running someone else's method verbatim in a clean container and posting the numbers is a better contribution than most patches, and it changed what gets built. I have carried it onto #342 with attribution.

AmirF194:
Glad the repro was useful, thanks for carrying it over to #342 with attribution.

MakazhanAlpamys:
@DelorneN007 — a decision on this one, so that it does not sit open without one.

I am going to close this PR in favour of the watchdog in #1084 (the #342 work). This is not a judgement on you:
- This PR is where the clamp-plus-watchdog idea first surfaced.
- Its review thread is where @AmirF194 found that a per-microbatch `zero_grad` wipes the accumulated gradient. That finding is now the reason #1084's watchdog sits at `on_pre_optimizer_step`, and I have asked for this PR and that finding to be credited in #1084's commit.

The code itself cannot be kept, for the same reason I have just had to correct my *own* approval on #1084:
- The clamp reads the logps before TRL's `_prepare_inputs` builds them, so it never fires.
- On Soup's default config (`num_iterations=1`), TRL does not compute old logps at all.
- The loss guard is superseded by #1084's gradient watchdog. The watchdog catches a non-finite loss too, and it does so at the correct once-per-step point.

Your #1093 is open, and I will review it on its own merits. Thank you for the work here.

Issue #890, opened 2026-09-12
soup serve --backend sglang sends the engine a doubled BOS (#785 part 2, split from #867)


abdulwaarith0:
Split out from #867 (#785 part 2). #867 fixes the vLLM backend and #782 fixed every path that tokenizes in process; this tracks the SGLang backend, which hands the engine a rendered string and lets the engine tokenize it.

## Measured

On an RTX 3060 Laptop (6 GB, WSL2) with `unsloth/Llama-3.2-1B-Instruct`, whose Llama-3 template renders `{{ bos_token }}` and whose `bos_token_id` is 128000. One request, `messages=[{"role": "user", "content": "Hi"}]`. `apply_chat_template(tokenize=True)` returns 36 ids starting `[128000, 128006]`.

SGLang 0.5.9, tokenizing the rendered string through its own `TokenizerManager` call (`self.tokenizer(tokenizer_input, **{})`, default `add_special_tokens=True`), reports `prompt_token_ids[:2] == [128000, 128000]` over 37 ids: a doubled BOS.

The tokenize step was executed directly on SGLang's own `get_tokenizer()` for the real model. The full engine could not be started on this box (three attempts, each stopped on a component needing `nvcc`, which is absent here), so the fix cannot be verified end to end from here.

## Fix routes checked (against 0.5.9)

- **Token ids**: `TokenizerManager._tokenize_one_request` uses `obj.input_ids` verbatim when a request carries them, so the ids route exists at the `/generate` endpoint. But `sgl.Runtime.generate`, which `utils/sglang.py` calls, has no `input_ids` parameter; it posts `{"text": prompt}`. Reaching the route means posting to `runtime.url + "/generate"` directly, or moving to `sgl.Engine` (whose `generate` does take `input_ids`).
- **No special tokens**: `GenerateReqInput` has no `add_special_tokens` field, `ServerArgs` has no such flag, and the text path calls `self.tokenizer(tokenizer_input, **{})`. There is no such option.

## Acceptance

- [ ] For a `{{ bos_token }}` template on a BOS-prepending tokenizer, the engine receives one BOS, matching `apply_chat_template(tokenize=True)` and #782 / #867
- [ ] A model with no chat template (or a template that fails to render) still sends the legacy string and is tokenized by the engine exactly as before
- [ ] Verified end to end on a real SGLang engine (needs a machine with the CUDA toolkit)
- [ ] A test that fails before the fix

[...]

PR #995, opened 2026-09-15
fix(serve): stop the SGLang backend from sending the engine a doubled BOS (#890)

abdulwaarith0:
The same defect as #867, on the backend that was split out: `build_chat_prompt`
renders a template that already carries BOS, and `sgl.Runtime.generate` only ever
posts `{"text": prompt}`, which the server tokenizes with `add_special_tokens=True`.
There is no request or server flag to turn that off, so the engine received
`[BOS, BOS, ...]`, a prompt shape the model was never trained or tuned on.

Measured on a running SGLang 0.5.9 (RTX 3060 Laptop, WSL2) with
`unsloth/Llama-3.2-1B-Instruct`, one request `[{"role": "user", "content": "Hi"}]`,
reading the ids the engine ran on back through `input_token_logprobs`:

  string route (main):  37 ids, `[128000, 128000, ...]`, prompt_tokens 37
  ids route (this fix): 36 ids, `[128000, 128006, ...]`, prompt_tokens 36

36 ids with `[128000, 128006]` is exactly `apply_chat_template(tokenize=True)`.
The greedy reply changed with it, from "It seems like you're starting a
conversation" to "How can I assist you today?".

`generate_with_runtime` now posts the ids `build_engine_prompt` encoded to the
runtime's own `/generate` endpoint, which uses `input_ids` verbatim, whenever the
chat template rendered the prompt. Both the streaming and the non-streaming route
go through it, and it returns the same `{"text", "meta_info"}` dict the app already
parsed (`decode_sglang_response` takes the dict as well as the JSON string
`Runtime.generate` returns, #76). A prompt no template rendered still goes through
`Runtime.generate` as the legacy string and is tokenized by the engine exactly as
before.

Mutants killed by the new tests: forcing the templated prompt back onto the string
route (9), inverting the guard so ids are sent with no template (9), posting the ids
under the `text` key instead of `input_ids` (5), reverting the non-streaming route
to the string path (5), reverting the streaming route (2), and dropping
`raise_for_status` (2).

The MII backend is fixed separately in #891.

Refs #785.

Signed-off-by: Abdulwaarith <abdulwaarithz@gmail.com>


MakazhanAlpamys:
**Your CI is red and it is not your diff.** `peft 0.21.0` was published at 13:33:59Z today; `pyproject.toml` declares `peft>=0.20.0,<1.0.0`, so CI took it, and nine layer-streaming adapter-key tests now fail on every PR. Yours installed `peft-0.21.0`; the last green `main` run installed `peft-0.20.0`. Your diff touches `utils/sglang.py` and its tests — nothing under `trainer/`, `layer_stream*` or PEFT. Tracked as **#1005**. Do not rebase or force-push chasing it.

Now the review, which is about the content.

## VERDICT: content merge-ready, one substantive fix wanted

**BOS counts, measured by execution** through the real `create_sglang_app` + `TestClient`, with a real `PreTrainedTokenizerFast` carrying a `TemplateProcessing` post-processor (your own offline fixture):

| | first 4 ids | BOS | total |
|---|---|---|---|
| before, string route | `[1, 1, 0, 0]` | **2** | 20 |
| after, ids route | `[1, 0, 0, 0]` | **1** | 19 |

`runtime.generate called: False`. That mirrors the 37 -> 36 you measured on the real Llama-3.2 tokenizer.

**Mutation.** Anchor `if prompt_token_ids is None: return decode_sglang_response(runtime.generate(...))` — **exactly one occurrence**; forced back to the pre-fix string route; `import soup_cli.utils.sglang` confirmed OK afterwards. **10 failed, 34 passed**, behavioural: `Expected 'post' to have been called once. Called 0 times.` Reverted, 44 passed.

**Deleted test lines: two, and the change is net-strengthening.**

```
-        self._post(app, {"messages": [{"role": "user", "content": "hello"}]})
-        sent_prompt = runtime.generate.call_args.args[0]
```

The prompt no longer travels through `runtime.generate`, so reading it off that mock is impossible by construction. You replaced it by reading the `/generate` payload back **and added `runtime.generate.assert_not_called()`**, which is a stronger statement than the line it replaces. Both #360 invariants survive byte-identically.

**And it is the same shape as the merged vLLM sibling** — you reuse `build_engine_prompt` verbatim, ids instead of string, rather than stripping a BOS after the fact. The transport differs only because `sgl.Runtime.generate` has no `input_ids` parameter, so you post to the same `/generate` endpoint it posts to. That is the stronger of the two available approaches.

Fragment credits you correctly.

---

### 1. Wanted before merge: `requests.post` has no `timeout=`

`utils/sglang.py:84-87`. It is a **blocking** call inside `async def chat_completions` (`:284`), so a wedged SGLang server stalls the whole event loop with no bound — not one request, the server. The repo's own HTTP helper sets one (`utils/webhooks.py:110`).

Worth naming precisely because your test at that site is called `test_a_server_error_on_the_ids_route_is_a_500_not_a_hang_or_a_leak` — it covers the error *status* and never the hang, so the name currently promises more than the test delivers. One line, plus an assertion that the timeout is passed.

### 2. Worth a sentence in the PR: `import requests` is a new undeclared dependency

`utils/sglang.py:82`. `requests` appears **nowhere** in `pyproject.toml` — the `[sglang]` extra is `sglang`, `fastapi`, `uvicorn` — and this is the **only** `requests.get/post` in all of `src/soup_cli/`; the repo uses `httpx`. It works transitively through sglang and your comment says so, so this is not a blocker, but it is a house-style divergence plus an undeclared runtime dep. Either switch to `httpx` (which is already a declared dep and gives you the timeout for free) or say in the fragment that it rides sglang's own dependency.

### 3. A version number disagrees with itself

`utils/sglang.py:61` and `tests/test_issue890_sglang_engine_bos.py:9` say **SGLang 0.5.19**; your fragment and PR body say **0.5.9**, which is what #890 measured. The stray `1` looks like the typo. Worth fixing because it is the provenance of a live-engine claim.

### 4. Minor: `runtime.url + "/generate"`

`:85` — a trailing slash on `runtime.url` gives `//generate`. sglang builds `url` without one, so this is low risk; noting it rather than asking for it.

---

One correction I owe you about this repo's own notes rather than your PR: CLAUDE.md's `utils/sglang.py` entry still says it *"STILL CARRIES both defects the vLLM rewrite fixed: its own hand-rolled prompt and a hardcoded finish_reason"*. **That is stale** — both were fixed by #360 and are gone from `main` (`sglang.py:193-194` imports `build_chat_prompt`; `:48` is `resolve_sglang_finish_reason`). So this PR is not a BOS fix layered on a hand-rolled prompt; it is the **third and last** defect of that set. I will correct the note.

Fix the timeout and I will merge as soon as #1005 clears — reviewing only the delta.

abdulwaarith0:
Timeout fixed (delta aa75f00). Switched the ids POST from `requests` to `httpx`, the client `utils/webhooks.py` and `commands/generate.py` already use, with `timeout=300.0` to match the server-generation path in `generate.py`. That also drops the only `requests` call in `src/` and the undeclared dep, and I added `httpx>=0.24.0` to the `[sglang]` extra so it is declared on this path.

Also `rstrip`ed a trailing slash on `runtime.url` so it can never build `//generate`, and corrected the docstring/test header to SGLang 0.5.9.

New tests assert the `timeout` kwarg is passed and finite and that the slash collapses. Removing the timeout fails both the payload assertion and the new bound test.

MakazhanAlpamys:
Re-reviewed the delta. **All four items closed — and the delta introduced one blocking regression in a test file this PR already touches.** One line.

## Closed, each by execution

**The timeout is real and pinned.** `sglang.py:92-96` now posts through `httpx.post(..., timeout=300.0)`, matching `commands/generate.py:888`. Removing that one line (anchor matched once, module imports) fails **2** tests — `test_the_ids_route_bounds_the_engine_with_a_timeout` with `KeyError: 'timeout'`, plus the call-args assertion. So the bound cannot be silently dropped again.

**You went further than I asked on the dependency and it was the right call.** `grep -rn "import requests" src/` returns nothing, and `pyproject.toml:148` adds `httpx>=0.24.0` to the `[sglang]` extra, matching `generate` at `:136`. Switching to a declared dep rather than documenting an undeclared one is the better half of the option I offered.

**Provenance is consistent** — `0.5.19` appears nowhere in `src/`, `tests/`, `changelog.d/` or `docs/`; all three sites say 0.5.9. **And the trailing-slash case got its own test**: `runtime.url.rstrip("/")` at `:93`, and reverting it fails `test_a_trailing_slash_on_the_runtime_url_does_not_double`.

**The BOS behaviour survived the httpx switch**, which was my one worry about swapping the client. Re-measured independently through the real `create_sglang_app` + `TestClient` with a real fast tokenizer:

```
AFTER  (ids route)    first4=[1, 0, 0, 0]  BOS=1  total=19   runtime.generate called: False
BEFORE (string route) first4=[1, 1, 3, 10] BOS=2
```

Byte-identical to the pre-delta measurement. The main-fix mutation still fires hard: forcing the string route back (`if prompt_token_ids is None:` -> `if True:`, anchor matched once) fails **11** tests.

---

## Blocking: `tests/test_issue360_sglang_prompt_finish_reason.py:134` still patches `requests.post`

The delta moved the call to `httpx.post` and updated `test_issue890`, but not `test_issue360` — which this PR also modifies. That patch is now a no-op, so the real `httpx.post` runs against a `MagicMock` url, raises, is swallowed by the handler, and `post.call_args` is `None`:

```
tests\test_issue360_sglang_prompt_finish_reason.py:137:
    AttributeError: 'NoneType' object has no attribute 'kwargs'
TypeError: Invalid type for url. Expected str or httpx.URL, got <class 'unittest.mock.MagicMock'>
```

Proven to be the delta rather than something pre-existing:

```
old head 3507aad0 : tests/test_issue360_... -> 23 passed
new head aa75f00a : tests/test_issue360_... -> 1 failed, 22 passed
```

Exactly one stale site — `grep -rn requests tests/test_issue360_sglang_prompt_finish_reason.py` returns line 134 only. `mock.patch("httpx.post", ...)` and it is done.

**This matters beyond the fix:** it means your red CI today is **not** only #1005. I told you earlier that the nine streaming failures were the peft breakage and not yours — that part stands, but this one is yours, and it is deterministic, so it would have stayed red after #1005 lands. Better you hear it now than after chasing the peft issue.

## One smaller thing

`changelog.d/0.75.0/995.fixed.md` predates the delta and says nothing about `httpx>=0.24.0` joining the `[sglang]` extra. That is a user-visible packaging change for anyone running `pip install "soup-cli[sglang]"` — worth a clause.

## A note, not a request

`httpx.post` is the sync API and still runs inside `async def chat_completions` (`:325`), so it still occupies the event loop — now bounded at 300 s instead of unbounded. That blocking property is pre-existing (`runtime.generate` blocks too) and the bound is what I asked for, so I am recording it rather than asking you to go async here.

Push the one-line patch fix and I will merge this ahead of #996, per your own ordering call.

abdulwaarith0:
Fixed in fe690e4. The stale `mock.patch("requests.post", ...)` at `tests/test_issue360_sglang_prompt_finish_reason.py:134` is now `httpx.post`, matching where the ids POST actually goes after the delta. `post.call_args.kwargs["json"]["input_ids"]` and the `.json()` return shape are unchanged, so the assertions hold.

Verified locally against this worktree (`PYTHONPATH=src`): `test_issue360` + `test_issue890` run 46 passed, 0 failed; ruff clean. `grep -rn requests tests/test_issue360_sglang_prompt_finish_reason.py` now returns nothing.

Also added a clause to `changelog.d/0.75.0/995.fixed.md` noting `httpx>=0.24.0` joins the `[sglang]` extra, since that is a user-visible change for `pip install "soup-cli[sglang]"`.

The remaining red on this branch should now be only the peft 0.21 / #1005 breakage you already identified, which is not in my diff.

MakazhanAlpamys:
**Fixed, verified, approved.** Merging as soon as #1010 clears the repo-wide peft breakage.

`tests/test_issue360_sglang_prompt_finish_reason.py` is back to **23 passed** (it was `1 failed, 22 passed` at `aa75f00a`), and the sweep is clean: `grep -rn requests` returns nothing in that file or in `src/soup_cli/utils/sglang.py`. I also checked no *other* sglang test still patches `requests` — repo-wide, the only files mentioning both are `test_code_review_high.py` and `test_issue332_vllm_prompt.py`, neither of which touches the sglang POST path.

**The BOS measurement is unmoved**, re-run independently through the real `create_sglang_app` + `TestClient`:

```
first 4 ids = [1, 0, 0, 0]   BOS count = 1   total = 19
runtime.generate called = False   http 200
```

Identical to before the httpx switch.

**Both mutations die, and the first one proves the repair was worth making.** Forcing the legacy string route back (`if prompt_token_ids is None:` -> `if True:`, anchor matched once, module imports) fails **12** tests — including `test_prompt_uses_chat_template_when_tokenizer_has_one`, the very test that had become a silent no-op. Deleting `timeout=300.0` (anchor matched once) fails **2**. 46 passed across both files after reverting; ruff clean.

**The fragment clause is right**, and it says the useful half rather than the minimal one — that `httpx>=0.24.0` joins the `[sglang]` extra and anyone installing `soup-cli[sglang]` picks it up automatically. That is what a reader of the release notes actually needs.

Worth naming what happened here, because it is a good failure rather than a careless one: you made a change I asked for, in the better of the two forms I offered, and the cost was a patch target moving out from under a *neighbouring* test that still passed its own file's expectations until it didn't. That class is only visible if someone runs the other file. No findings; nothing else needed from you.

MakazhanAlpamys:

MakazhanAlpamys:

The cell `test (ubuntu-latest, 3.11)` failed with nine tests from `test_v07201.py` / `test_v07202.py` / `test_v07203.py`, all with the same shape:

```
                which raises no exception
```

```
your run created   2026-09-15T18:23:11Z
```

MakazhanAlpamys:

```
git fetch --depth=1 origin +<a pinned merge SHA>:refs/remotes/pull/995/merge
git checkout --force refs/remotes/pull/995/merge
```

```
```

## What actually clears it

abdulwaarith0:
Go ahead with `gh pr update-branch --rebase` on this one. Nothing in the diff needs to change, so a rebase onto current `main` just to recompute the merge SHA is exactly right, and it keeps the history linear. Thanks for catching that the re-run couldn't have worked and correcting it in the same hour rather than letting me watch it fail.

I'll take #996 myself once this lands. That one I have to touch anyway to resolve the `docs/serving-and-export.md:283-287` conflict so the parenthetical says neither backend hands the engine the rendered string, so I'll rebase it onto the new `main`, fix that line, and push. Keeping the order we agreed: this first, then #996.

MakazhanAlpamys:
Done, @abdulwaarith0: rebased onto current `main` with `gh pr update-branch --rebase`, as you asked. The new head is ccd84d66, and the PR's own diff is byte-identical to fe690e40's once `index` lines are ignored.

Before pressing it, I re-checked your change merged onto today's `main`:
- No conflicts, and F821 is clean.
- `test_issue360` plus `test_issue890` give 46 passed. The SGLang neighbours (#332, #785, `sglang_serve`, the trust contract, v0.53.0, v0.71.14, v0.73.0) give 349 passed.
- Both mutations still fail. Forcing the legacy string path fails 12 tests, including `test_what_the_engine_receives_carries_exactly_one_bos`. Removing the timeout fails 2.

The old red was the frozen pre-#1010 merge ref: exactly the nine `test_v0720x` streaming tests, in every cell. I will merge when the new run is green and close #890 by hand, since the body only says `Refs #785`. After that, #996 is yours to rebase.

MakazhanAlpamys:
@abdulwaarith0, one cell (`test (windows-latest, 3.11)`) came back red, and it is not your change. The interpreter died with `Windows fatal exception: code 0xc000001d` inside a real `trainer.train()` in `tests/test_rewind_hf.py`. That is the #382 illegal-instruction crash on part of GitHub's Windows runner fleet, reached by tests that landed after the shared guard existed. The other 12 required checks are green.

I have opened #1059 to put those tests behind the guard. Once it lands, I will update your branch so a fresh run is created, then merge on green. Nothing is needed from you.

MakazhanAlpamys:
Unblocked — I updated your branch, and the red cell was never your code.

`test (windows-latest, 3.11)` was dying with `Windows fatal exception: code 0xc000001d` inside `tests/test_rewind_hf.py`, which is issue #382: part of GitHub's `windows-latest` fleet lacks an instruction the wheel emits, the interpreter dies with a faulthandler dump and no Python exception, and everything the cell had not yet reached is simply never reported. Your patch is nowhere near that file.

#1059 landed an hour ago and puts those eight real-training tests behind the shared `skip_on_windows_ci` guard, which fires only on `win32` **and** `CI=true` — so a developer's Windows box still runs them and ubuntu/macOS keep the coverage. I ran `gh pr update-branch` here so your PR picks that up; CI is re-running now.

Nothing is owed on your side. I will merge on green.

One thing I want to flag before that, so it does not surprise you: **#890 does not auto-close from this PR.** The body says `Refs`, and I would rather close it by hand with a note than have the link do something the text does not say. I will do that at merge and mention your PR in it.

#996, the MII half of the same defect, is still waiting on a rebase from you — it is the other side of the doubled-BOS pair and #891 is still open for it. No rush, but it is the next one.

MakazhanAlpamys:
Merged — thank you, and sorry it took this long. Most of the delay was a red cell that was never yours.

The fix is the right shape, and it is the same one #781 established: **encode the prompt the way training encoded it**, rather than detecting a doubled BOS and stripping it afterwards. A stripper has to know which token to remove and when, and it is wrong the moment a tokenizer does something slightly different. Not adding special tokens to an already-rendered template has no such failure mode.

What makes this defect worth the attention it got is that it is silent. Two BOS tokens is a prompt the model was never trained on; there is no error, no warning, and the degradation looks like the model just being a bit worse. Nobody files a bug for that.

Verified before merging, on the tree rebuilt by merging your branch into current `main` (GitHub's merge ref is pinned when the run is created, so the green marks described a much older `main`), base interpreter, peft 0.20.0, `soup_cli.__file__` under the scratch tree:

- Serving and BOS suites — **196 passed, 1 skipped**.
- `ruff check --select F821 src/soup_cli/` — clean.
- A decorator-level diff of the merged tree against `main` — nothing lost. I added that check today after finding that a naive conflict resolution can silently drop a `@model_validator` line while `ruff` and the import both stay happy.

On the wait: your `test (windows-latest, 3.11)` was dying on #382's illegal-instruction crash inside `tests/test_rewind_hf.py` — a file you do not touch. #1059 put those eight real-training tests behind the shared `win32 AND CI=true` guard, and I ran `update-branch` so this picked it up. You diagnosed that correctly yourself on the thread, and were right.

**#890 I am closing by hand**, as I said I would — the body says `Refs`, and I would rather close it deliberately with a note than have a keyword do something the text does not say.

**#996 is next.** It is the MII half of the same defect and it is waiting on a rebase from you now that this has landed; #891 is still open for it. No rush, but that one is one command away from being reviewable.

PR #996, opened 2026-09-15
fix(serve): stop the MII backend from sending the engine a doubled BOS (#891)

abdulwaarith0:
The same defect as #867, on the backend that was split out. `MIIPipeline.__call__`
takes strings only and encodes each one through `HFTokenizer.encode`, which is
`tokenizer.encode(input, return_tensors="pt")` with no way to say the special
tokens are already there. On a BOS-prepending tokenizer whose chat template renders
`{{ bos_token }}` (`unsloth/Llama-3.2-1B-Instruct`, `bos_token_id` 128000), that is
`add_special_tokens=True` over a string that already starts with the BOS text, so
the prompt tokenizes to `[128000, 128000, ...]`, 37 ids, against the 36 ids
`apply_chat_template(tokenize=True)` returns.

MII's seam is `ModelConfig.tokenizer`: `mii.pipeline(tokenizer=...)` accepts a
wrapper, which `load_tokenizer` wraps once more and calls
`encode(input, return_tensors="pt")` on, with the prompt object the pipeline was
given, untouched. So `soup serve` builds the pipeline with a `SoupMiiTokenizer`
wrapper around the same tokenizer whose chat template it applies, and hands a
templated prompt over as a `TokenizedPrompt`: still a `str`, so the pipeline's
contract holds, but carrying the ids `build_engine_prompt` encoded with no tokenizer
special tokens, which the wrapper returns verbatim. A prompt no template rendered
stays a plain `str` and is encoded with the exact call MII's `HFTokenizer` makes, so
the engine tokenizes it as before. If the engine reports a prompt length other than
the one Soup encoded, the server logs a warning naming both rather than doubling in
silence.

Verified at the tokeniser level (`add_special_tokens=True` over the rendered string
gives the 37-id doubled BOS; the fix's encoding gives 36) and by
`tests/test_issue891_mii_engine_bos.py`, which exercises the wrapper against MII
0.3.3's real `HFTokenizer.encode` line without a GPU. I could not complete a live
DeepSpeed-MII run on my hardware: its inference kernels JIT-build against a CUDA
toolkit that will not compile on this box, so the end-to-end engine measurement is
still open. Mutants killed: sending the templated prompt as a plain `str` (2), making
the wrapper ignore the carried ids (7), and not wiring the tokenizer into the
pipeline (1).

The SGLang backend is fixed separately in #890.

Refs #785.

Signed-off-by: Abdulwaarith <abdulwaarithz@gmail.com>


MakazhanAlpamys:
**Your CI is red and it is not your diff.** `peft 0.21.0` landed on PyPI at 13:33:59Z today and `pyproject.toml` declares `peft>=0.20.0,<1.0.0`, so CI took it; nine layer-streaming adapter-key tests now fail on every PR. Your diff touches `utils/mii.py` and `commands/serve.py` and nothing under `trainer/`, `layer_stream*` or PEFT. Tracked as **#1005**. Nothing for you to do.

## VERDICT: content merge-ready

**BOS counts, measured by execution** through MII 0.3.3's real `HFTokenizer.encode` line, with the double-wrap MII actually performs:

| | first 4 ids | BOS | total |
|---|---|---|---|
| before, plain `str` + MII's own tokenizer | `[1, 1, 0, 0]` | **2** | 20 |
| after, `TokenizedPrompt` + `SoupMiiTokenizer` | `[1, 0, 0, 0]` | **1** | 19 |
| reference `apply_chat_template(tokenize=True)` | `[1, 0, 0, 0]` | 1 | 19 |

After **equals** the reference, and the legacy plain-string control produces output identical to raw MII. That last control is the one that makes this safe to merge.

**Mutation.** Anchor `token_ids = getattr(input, "token_ids", None)` — **exactly one occurrence**; forced to `None`, i.e. pre-fix behaviour; `import soup_cli.utils.mii` confirmed OK. **7 failed, 40 passed**, and the failing assert is the doubled BOS itself rather than a proxy for it:

```
assert [1, 1, 0, 0, 0, 6, ...] == [1, 0, 0, 0, 6, 7, ...]
  At index 1 diff: 1 != 0
```

Reverted, 47 passed. **Zero removed test lines** — I checked rather than assumed.

`commands/serve.py:529-552`: loading the tokenizer before the pipeline is strictly better, because a tokenizer failure now aborts before any GPU allocation. No error-path regression.

Fragment credits you correctly.

---

### 1. The `TokenizedPrompt(str)` copy hazard needs a test, not a change

`utils/mii.py:78-92`. Measured: `prompt.strip()` returns a plain `str` with `hasattr(..., "token_ids")` **False**; so does `f"{prompt}"`. If MII ever touches the string anywhere in its path, the fix silently reverts to the doubled BOS **with no exception** — which is the exact silent failure mode the #785 class exists to kill.

You do defend it: `_check_prompt_length` (`:193-213`) compares `Response.prompt_length` against the ids Soup encoded and warns. Two gaps:

- it is `logger.warning` only, and
- **nothing pins the hazard.** There is no test asserting that a copied prompt loses its ids, so a future refactor could remove `_check_prompt_length` and no test would notice that the last line of defence went.

I am not asking you to change the design — given the seam analysis in #891 this is the strongest route MII allows. I am asking for the test that makes the hazard visible: assert `TokenizedPrompt(...).strip()` has no `token_ids`, and assert the warning fires when the length disagrees. That converts an undocumented sharp edge into a documented one.

### 2. `create_mii_tokenizer` mutates the caller's tokenizer

`utils/mii.py:141-145`. Measured: `pad_token` goes `None -> '</s>'` **on the same object** `commands/serve.py` hands `build_mii_app` for chat-template rendering (`wrapped.tokenizer is tok` is True). It mirrors what MII's own `HFTokenizer` does and you **assert** it at `tests/test_issue891_mii_engine_bos.py:257`, so it is deliberate and tested.

But it is an undeclared side effect of a function named `create_*`, and it runs against `.claude/rules/common/coding-style.md` ("ALWAYS create new objects, NEVER mutate existing ones"). A one-line docstring note saying the mutation is intentional and mirrors MII would settle it; I am not asking for a copy.

---

## Ordering with #995 — you own both, so this is yours to sequence

`git merge-tree --write-tree` on your two heads returns a **content conflict**, in exactly one file: `docs/serving-and-export.md`, lines 283-287. Both rewrite the same parenthetical in mutually exclusive ways:

- **#995** -> *"(The MII backend still hands the engine the rendered string; see #891.)"*
- **#996** -> *"(The SGLang backend still hands the engine the rendered string; see #890. The MII backend tokenizes through a wrapper…)"*

No source, test or fragment conflict — `utils/sglang.py` and `utils/mii.py` + `commands/serve.py` are disjoint, and neither adds `build_engine_prompt` since it is already on `main` from #867. So nothing doubles up.

**The thing to watch:** whichever merges second must rebase, and **the resolution has to end up saying neither backend still sends the string.** Taking either side verbatim leaves the docs asserting a falsehood about the backend the other PR just fixed — which is the same class of defect these two PRs are closing, one level up.

Tell me which order you want and I will follow it.

abdulwaarith0:
Both asks are in (delta 0c458a1), design unchanged:

- `TestTokenizedPromptCopyHazard`: asserts a copied prompt (`strip`, f-string, concat, slice, join) is a plain `str` with no `token_ids`; asserts encoding such a copy re-adds the engine BOS (the defect back in silence); and a direct unit test that `_check_prompt_length` warns on a length mismatch and is quiet on a match.
- Docstring note on `create_mii_tokenizer` that setting `pad_token` on the caller`s tokenizer is a deliberate in-place mutation mirroring MII`s own `HFTokenizer`.

On ordering: I would merge #995 first, then #996. Once #995 lands I will rebase this one and resolve `docs/serving-and-export.md:283-287` so the parenthetical says neither backend still hands the engine the rendered string (SGLang posts ids to `/generate`, MII tokenizes through the wrapper), instead of either PR`s current text asserting the other backend is still unfixed. Happy to flip the order if you would rather.

MakazhanAlpamys:
Re-reviewed the delta. **Both items closed, zero source behaviour change, merge-ready.** The delta is `mii.py +8` (docstring) and `test_issue891 +55` — pure addition, and `git diff origin/main...<head> -- tests/ | grep '^-'` is **empty**.

**The copy hazard is now visible, and you covered more shapes than I asked for.** `TestTokenizedPromptCopyHazard` parametrizes five ways a `str` subclass loses its identity — `strip()`, an f-string, `+ ""`, `[:]`, `"".join(...)` — each asserting both `not isinstance(..., TokenizedPrompt)` and `not hasattr(..., "token_ids")`. Then `test_encoding_a_copied_prompt_reverts_to_the_doubled_bos` proves what that costs: `encode_mii_prompt(tok, f"{tokenized}")[:2] == [BOS, BOS]`. That last one is the test that makes the hazard concrete rather than abstract.

**And you pinned the guard in both directions** — `test_check_prompt_length_is_the_runtime_guard_for_that_reversion` asserts the warning fires on a mismatch **and stays silent on a match**. A one-directional version would pass with a guard that warns constantly.

**You are right that five of those are documentation tests** — they pin a property of `str`, not of Soup code, so no source mutation can kill them. That is exactly what I asked for ("make the hazard visible"), and the mutation-killable half is the guard itself. I checked that: killing `_check_prompt_length`'s comparison (`isinstance(prompt_length, int) and prompt_length != len(...)` -> `False`, anchor matched once, module imports) fails **2** tests, including your new one — where previously that guard was pinned by a single test at `:456`.

The main-fix mutation still fires as before: `token_ids = getattr(input, "token_ids", None)` -> `None` fails **7**, with the doubled BOS itself as the assertion rather than a proxy for it.

**The docstring does what I wanted and slightly better** — `mii.py:136-143` calls it a "Deliberate side effect", says it mutates the caller's object, **names the style rule it runs against**, and gives the reason. Naming the rule you are breaking is the part that stops the next reader filing it as a bug.

32 passed in the file; 137 across every other MII-touching file (including `test_mii_serve.py`, `test_multi_gpu.py`, `test_part_b.py`, since your change touches `create_mii_pipeline`'s signature). Ruff clean. Fragment credits you and is still accurate.

---

## Ordering — your call, and I am taking it

The two heads still conflict in exactly one file, `docs/serving-and-export.md:283-287`, with the same mutually exclusive parentheticals. No source, test or fragment conflict.

You proposed **#995 first, then rebase #996** so the resolution says **neither** backend still sends the string. Agreed, doing that.

One thing to know before you rebase: **#995 is currently blocked on a one-line regression its own delta introduced** — it switched to `httpx` but left `tests/test_issue360_sglang_prompt_finish_reason.py:134` patching `requests.post`, which is deterministic and would stay red after #1005 lands. So #996 waits on that, not on anything of its own.

Your CI here: I could not read the per-test failure list — `gh run view --log-failed`, the job-log API and the check-run annotations all return nothing usable while a run is in flight. Locally your 137 MII-related tests are green and your diff touches nothing under `trainer/` or `layer_stream*`, which is consistent with #1005 being the whole story. **I am telling you what I could and could not establish rather than asserting the convenient one.**

MakazhanAlpamys:

MakazhanAlpamys:

The cell `test (macos-latest, 3.10)` failed with nine tests from `test_v07201.py` / `test_v07202.py` / `test_v07203.py`, all with the same shape:

```
                which raises no exception
```

```
your run created   2026-09-15T15:23:42Z
```

MakazhanAlpamys:

```
git fetch --depth=1 origin +<a pinned merge SHA>:refs/remotes/pull/996/merge
git checkout --force refs/remotes/pull/996/merge
```

```
```

## What actually clears it

abdulwaarith0:
Rebased onto current `main` now that #995 has landed. New head `8aa4b1f`.

**Docs resolution, as we agreed.** The only conflict was the vLLM parenthetical in `docs/serving-and-export.md`. My branch's version still claimed "the SGLang backend still hands the engine the rendered string (#890)", which #995 made false, so I dropped that clause and kept only the MII half:

> (The MII backend also tokenizes through a wrapper around the same tokenizer, so a templated prompt reaches its engine with the ids the template implies, #891.)

`git diff origin/main...HEAD -- docs/serving-and-export.md` is exactly that one parenthetical and nothing else. The SGLang section you wrote for #995 (the `posts the token ids to /generate` / #890 paragraph) is untouched, it came through from `main`. So the doc now says neither backend still sends the engine the rendered string, which was the whole point of sequencing them.

**Checked nothing was silently dropped** (the `@model_validator` hazard you flagged): `utils/mii.py`, `tests/test_issue891_mii_engine_bos.py` and the 996 changelog fragment are byte-identical to the head you reviewed (`git rev-parse HEAD:<file>` matches `0c458a1:<file>`). `commands/serve.py` is the only source file the rebase had to auto-merge, because `main` moved under it; the net `origin/main...HEAD` diff there is just my #891 change, tokenizer loaded before the pipeline and passed in as `tokenizer=`, reapplied cleanly with no loss.

CI is recomputing against the fresh merge ref, so the stale `test_v07201/2/3` red from #1005 is gone with this event. Nothing else changed on my side.

MakazhanAlpamys:
Thank you for the rebase. Your "byte-identical" claim is exactly true: `utils/mii.py`, the test file and the fragment have the same blob SHAs as `0c458a1`, and the docs conflict with #995 was resolved as agreed, so no page still says SGLang or MII sends the rendered string.

I owe you a correction, though. **The earlier merge-ready verdict on `0c458a1` missed one gap, and it is on me, not a regression from your rebase:** the one line in `serve.py` that makes the fix reach a real server is not covered by any test.

Review on `8aa4b1f1`, on the tree merged with `main` at `c1a45efb`:
- `test_issue891_mii_engine_bos.py`: 32 passed. With the #781 / #785 / #890 engine-BOS files, `test_mii_serve`, `test_multi_gpu`, `part_b`, `816` and the #487 changelog tests: 335 passed. ruff and F821 clean.
- 8 of 9 mutations killed:
  - sending the plain rendered string instead of `TokenizedPrompt` (the doubled BOS itself);
  - falling through to `tokenizer.encode`;
  - ignoring the carried `token_ids`;
  - `create_mii_pipeline` not passing the wrapper;
  - the silenced length warning;
  - the `pad_token` removal;
  - `__len__` returning 0;
  - the unflattened wrapper-mode encode.

**The survivor:** dropping `tokenizer=mii_tokenizer,` from the `create_mii_pipeline(...)` call in `serve.py` (~L562-566) passes 71/71 targeted tests and 1740 in the wider serve/MII set. `test_mii_serve.py::_invoke` patches `create_mii_pipeline` with `return_value=MagicMock()` and never looks at its kwargs. Without that argument, every real `soup serve --backend mii` goes back to the doubled BOS, and only the runtime length warning would show it.

Fix: patch `create_mii_pipeline` with a `side_effect` that captures its kwargs, and assert `captured["tokenizer"]` is the tokenizer `serve` built. Optionally, also assert that `_load_serve_tokenizer` runs before the pipeline is created. That is a one-test push, and then this merges.

abdulwaarith0:
Fixed. The survivor is killed by capturing `create_mii_pipeline`'s kwargs with a `side_effect` instead of the bare `MagicMock`, asserting the loaded tokenizer reaches the pipeline, and pinning that `_load_serve_tokenizer` runs before the pipeline is built (`call_order == ["tokenizer", "pipeline"]`). The `None`-degrade case now pins `tokenizer=None` onto the pipeline too.

Dropping `tokenizer=mii_tokenizer` from the `create_mii_pipeline(...)` call now fails both new assertions; I verified that by mutation and reverted. Test-only, `serve.py` is byte-identical to the head you reviewed. Head `060b92f`, matrix green.

MakazhanAlpamys:
Merged as `a2ec51aa` — thank you, and thank you in particular for the way you closed the one item I
had missed.

I called an earlier head of this PR merge-ready and that was wrong: dropping the `tokenizer=` kwarg
at the `create_mii_pipeline` call site left every test green, so the line that makes the whole seam
hold was unpinned. You did not answer that with an assertion that the kwarg is present. You captured
the call's kwargs with a `side_effect`, asserted the **loaded** tokenizer is the object that arrives,
and pinned that `_load_serve_tokenizer` runs before the pipeline is built — which also covers the
`None`-degrade path, so a fix that merely passed *something* would still fail.

**Re-verified on the tree merged with current `main`, not from the description:**

```
mutation: drop tokenizer= at commands/serve.py create_mii_pipeline(...)
   -> 2 failed, 53 passed  KILLED   (was the survivor on 8aa4b1f1)
   AssertionError: serve() called create_mii_pipeline without a tokenizer kwarg
mutation: encode_mii_prompt ignores token_ids (pre-fix)        -> 7 failed   KILLED
mutation: never wrap in TokenizedPrompt                        -> 2 failed   KILLED
mutation: pipeline called with [prompt] not [engine_prompt]    -> 2 failed   KILLED
mutation: create_mii_pipeline stops wrapping the tokenizer     -> 1 failed   KILLED
mutation: build_chat_prompt instead of build_engine_prompt     -> 24 failed  KILLED
mutation: length guard silenced                                -> 2 failed   KILLED
mutation: wrapper drops pad_token / __len__ -> 0 / no flatten  -> 1,2,1      KILLED
mutation: TokenizedPrompt.token_ids = []                       -> 8 failed   KILLED
mutation: build the pipeline before the tokenizer loads        -> 2 failed   KILLED
12 of 12 killed.

both test files on clean main (2709f25b):  28 failed, 27 passed
this PR:  tests/test_issue891_mii_engine_bos.py + tests/test_mii_serve.py  -> 55 passed
blast radius (11 files touching the engine-prompt seam)        -> 392 passed
ruff: clean. Required checks on 060b92f1: 13/13 pass, merge ref built against current main.
```

Your "`serve.py` and `utils/mii.py` are byte-identical to the head you verified" claim checked out by
blob SHA rather than by reading — `a6135f56` and `70ad91ee` on both heads — so the only delta since
my last verification really was the test file, +67/-14. Zero test lines deleted. And the docs came
out right: `grep` over the merged tree finds no page still claiming that MII or SGLang hands the
engine the rendered string, which is the resolution you and #995 agreed on before either landed.

That is seven merges, and this one closes the class you opened yourself in #785 — training, vLLM,
SGLang, MII. Your `CONTRIBUTORS.md` entry is updated.

